PR #6307, opened 2017-11-14
Reorder windows path list

msarahan:
Having Library/usr/bin and Library/mingw-w64/bin ahead of Library/bin bit me pretty hard today with curl.  There was a git4win curl version shadowing ours, and that git4win one was busted.

### DIFF
--- a/conda/cli/activate.py
+++ b/conda/cli/activate.py
@@ -66,10 +66,10 @@ def prefix_from_arg(arg, shell):
 def _get_prefix_paths(prefix):
     if on_win:
         yield prefix.rstrip("\\")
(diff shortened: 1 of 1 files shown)


msarahan:
corresponding constructor PR at https://github.com/conda/constructor/pull/141

kalefranz:
@mingwandroid Can you think of any issues that this reordering could create?

msarahan:
@mingwandroid and I discussed this on flowdock.  He is concerned that this will break msys2 stuff.  I have tried to bundle a more minimal git4win package, and we'll try that before we muck with this.  Closing for now.

PR #6313, opened 2017-11-15
Make environments.txt record non-named environments

### LINKED ISSUE #5537: Environment missing from list when created with `conda create -p`
### Environment

-   conda version: `4.3.14` though I've seen this issue in previous and newer versions as well. All installed through `Anaconda` installers. Tested upto `4.3.21`
-   Python version: `2.7` `3.5` `3.6`. I mean these as the main conda python version
-   Operating System: `Win7`, `Win8`, `Win10`.

### Details

-   Describe what you were trying to get done.

Create and manage a conda environment created in a custom path.

-   What commands did you run to trigger this issue? If you can provide a
    [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve)
    this will help us understand the issue.

`conda create -p c:\python27 python=2.7`

-   If there was a crash, please include the traceback here.

When running:

```
conda create -p c:\python27 python=2.7
conda env list
```

This returns the following list
```
# conda environments:
#
root                  *  C:\python36
```

But when running
```
conda create -n py35 python=3.5
conda env list
```

This returns
```
# conda environments:
#
py35                     C:\python36\envs\py35
root                  *  C:\python36
```

If I run `activate c:\python27` the correct environment activates with no issue.

Inside of `\.conda\environments.txt` it has
```
c:\python27
C:\python36\envs\py35
```

### LINKED ISSUE #6088: flags --name and --prefix should pick up env's .condarc file
Right now probably picking up the `.condarc` of the active env, which is wrong when `--name` or `--prefix` flags are used for CIUR operations.

### LINKED ISSUE #6303: Ensure environments.txt file is maintained in conda 4.4

kalefranz:
fix #6088
resolve #6303
resolve #5537

This PR fixes an issue were using `-p` or `-n` in `create`, `install`, `update`, and `remove` commands fails to pick up configuration for the target prefix.  It also makes sure we're maintaining the `envs/environments.txt` file.  Only environments that are not 'named' environments are recorded in environments.txt.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -190,6 +190,22 @@ class Context(Configuration):
     def __init__(self, search_path=None, argparse_args=None):
         if search_path is None:
             search_path = SEARCH_PATH
+
+        if argparse_args:
+            # This block of code sets CONDA_PREFIX based on '-n' and '-p' flags, so that
+            # configuration can be properly loaded from those locations
+
+            func_name = ('func' in argparse_args and argparse_args.func or '').rsplit('.', 1)[-1]
+            if func_name in ('create', 'install', 'update', 'remove', 'uninstall', 'upgrade'):
+                if 'prefix' in argparse_args and argparse_args.prefix:
+                    os.environ['CONDA_PREFIX'] = argparse_args.prefix
+                elif 'name' in argparse_args and argparse_args.name:
+                    # Currently, usage of the '-n' flag is inefficient, with all configuration
+                    # files being loaded/re-loaded at least three times.
+                    target_prefix = determine_target_prefix(context, argparse_args)
+                    if target_prefix != context.root_prefix:
+                        os.environ['CONDA_PREFIX'] = determine_target_prefix(context, argparse_args)
+
         super(Context, self).__init__(search_path=search_path, app_name=APP_NAME,
                                       argparse_args=argparse_args)
 
@@ -428,7 +444,6 @@ class Context(Configuration):
     def target_prefix(self):
         # used for the prefix that is the target of the command currently being executed
         # different from the active prefix, which is sometimes given by -p or -n command line flags
-        from ..core.envs_manager import determine_target_prefix
         return determine_target_prefix(self)
 
     @memoizedproperty
@@ -883,7 +898,7 @@ def locate_prefix_by_name(name, envs_dirs=None):
         return context.root_prefix
     if envs_dirs is None:
         envs_dirs = context.envs_dirs
-    for envs_dir in concatv(envs_dirs, (os.getcwd(),)):
+    for envs_dir in envs_dirs:
         if not isdir(envs_dir):
             continue
         prefix = join(envs_dir, name)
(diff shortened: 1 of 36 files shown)


kalefranz:
I don't know why these TravisCI builds are hanging around, and showing the whole PR as failing.  I'll be working on figuring that out...

nehaljwani:
I toggled this button:
screenshot from 2017-11-29 21-15-33

kalefranz:
Oh perfect

conda/core/envs_manager.py
```diff
@@ -299,155 +154,42 @@ def unregister_env(self, location):
...
+            idx = environments_txt_lines.index(location)
+            del environments_txt_lines[idx]
+        except ValueError:
+            # remove_location was not in list.  No problem, just move on.
+            pass
+
+        real_prefixes = tuple(p for p in environments_txt_lines if self.is_conda_environment(p))
+        if self.is_writable:
+            with open(self.catalog_file, 'w') as fh:
+                fh.write('\n'.join(real_prefixes))
```
nehaljwani: A new line is missing here. This should be
```
fh.write('\n'.join(real_prefixes) + '\n')
```

nehaljwani: If the new line is missing at the end, then the next prefix will get appended on the previous line, causing something like this:
```
/tmp/test1/tmp/test2
```
And a consecutive call to `conda env list` will remove this line itself, since it might not be a valid folder

conda/base/context.py
```diff
@@ -187,6 +187,28 @@ class Context(Configuration):
...
+            search_path = SEARCH_PATH
+
+        if argparse_args:
+            # This block of code sets CONDA_PREFIX based on '-n' and '-p' flags, so that
+            # configuration can be properly loaded from those locations
+            func_name = ('func' in argparse_args and argparse_args.func or '').rsplit('.', 1)[-1]
+            if func_name in ('create', 'install', 'update', 'remove', 'uninstall', 'upgrade'):
+                if 'prefix' in argparse_args and argparse_args.prefix:
+                    os.environ['CONDA_PREFIX'] = argparse_args.prefix
+                elif 'name' in argparse_args and argparse_args.name:
```
nehaljwani: I see repeated use of:
```
if 'x' in argparse_args and argparse_args.x
```

There is absolutely nothing wrong with it. One could argue that it is more articulate. 

However, I would like your opinion on doing:
```
argparse_args_d = vars(argparse_args)
```
... and then using it as:
```
if argparse_args_d.get('x', None):
```

conda/core/envs_manager.py
nehaljwani: Perhaps a test case could cover this scenario too?

conda/base/context.py
kalefranz: Would be fine too.  I handled both cases in https://github.com/conda/conda/blob/d432501ea747202c8b1f6b6d32d131df59fdb4c6/conda/common/configuration.py#L767-L782

It would be creating an extra dict instance that's very soon thrown away though inside the `super` call I think.

kalefranz: That `Namespace` thing that's part of argparse is just so dang weird.

kalefranz:
Looks like tests are good to go now.

conda/gateways/disk/create.py
```diff
@@ -315,28 +315,13 @@ def create_package_cache_directory(pkgs_dir):
     # returns False if package cache directory cannot be created
     try:
         log.trace("creating package cache directory '%s'", pkgs_dir)
-        mkdir_p(pkgs_dir)
-        touch(join(pkgs_dir, 'urls'))
-        touch(join(pkgs_dir, PACKAGE_CACHE_MAGIC_FILE))
+        sudo_safe = expand(pkgs_dir).startswith(expand('~'))
```
nehaljwani: What if a user has defined a package cache directory using `pkgs_dirs` in ~/.condarc to some place other than his `~`? Could be a shared location among users of the same primary group.

conda/gateways/disk/__init__.py
```diff
@@ -63,3 +64,18 @@ def mkdir_p(path):
...
+
+
+def mkdir_p_sudo_safe(path):
+    if isdir(path):
+        return
+    base_dir = dirname(path)
+    if not isdir(base_dir):
+        mkdir_p_sudo_safe(base_dir)
+    log.trace('making directory %s', path)
+    mkdir(path)
```
nehaljwani: What if `mkdir()` throws an exception? Partition ran out of inodes or something ...

nehaljwani:
Empty environments are not being reported by `conda env list`?
```
[root@cc14dc945ce7 conda]# conda create -yp /tmp/test9                                                                                                                                                             
Solving environment: done

## Package Plan ##

  environment location: /tmp/test9


Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use:
# > source activate /tmp/test9
#
# To deactivate an active environment, use:
# > source deactivate
#

[root@cc14dc945ce7 conda]# tail -1 ~/.conda/environments.txt 
/tmp/test9
[root@cc14dc945ce7 conda]# conda env list
# conda environments:
#
test                     /opt/conda/envs/test
base                  *  /opt/conda

```

conda/gateways/disk/create.py
kalefranz: Not ideal, but this is all current behavior, even in 4.3. The config for `pkgs_dirs` and `envs_dirs` *is* independent though. What’ll happen if a user defines a custom envs dir is that, to test permissions, a `pkgs` directory will be created along side it, then an empty `urls.txt` file in that directory. Like I said, not ideal. But itshow conda 4.3 does it, so the behavior the user sees at least doesn’t change.

nehaljwani: No, I meant `sudo_safe` would be false, if the `pkgs_dir` is a custom path.

kalefranz: I’m not sure right now how to improve on it. I wanted to add the sudo_safe stuff to to help people who shouldn’t be using sudo at all. If people are sophisticated enough to be customizing directories like that, then for now I’m ok hoping they’re also sophisticated enough to dig themselves out of a sudo hole if they manage to get into one.

nehaljwani: Fair enough.

PR #6345, opened 2017-11-27
Replace Travis-CI with CircleCI

kalefranz:
This PR replaces Travis-CI with CircleCI, to help spread the load out among CI services and projects under the conda org.

In the transition, we've dropped a couple of test targets.  Once merged, macOS will no longer be tested.  I don't think our macOS tests on Travis-CI have ever given us additional information beyond what we got from the Linux tests, so I think that's an acceptable reduction in the matrix.  Also, this drops the python 3.4 test target.  With the latest miniconda's (4.3.27+), I think we've declared that we're no longer supporting python 3.4 in the root environment.  And, for the purposes of the way code is written in this project, there isn't much if any difference between python 3.4 - 3.6+.


kalefranz:
The Travis-CI test (marked as failure here) will go away once this PR is merged.  I believe that TravisCI configuration is coming from the `.travis.yml` file that's in the mainline 4.4.x branch right now.

kalefranz:
The Appveyor failure was a single test, with a pexpect timeout.  It didn't fail in previous tests.  I think it's a one-off.  This is the second time I've seen that failure though.  If it becomes more common, we're going to have to address it.  I don't think we should address it here in this PR.

kalefranz:
We're now using custom Docker images as the test base.

The Docker images used here are
  - condatest/linux-64-python-3.6
    https://github.com/conda/conda-docker/blob/master/condatest/linux-64-python3.6/Dockerfile
  - condatest/linux-64-python-2.7
    https://github.com/conda/conda-docker/blob/master/condatest/linux-64-python2.7/Dockerfile

PR #6347, opened 2017-11-27
revert features to 4.3.x behavior, removing key-value features for now

### LINKED ISSUE #6304: Revert interfaces for key-value features

kalefranz:
resolve #6304

This PR reverts key-value features incorporated into 4.4, back to the 4.3 behavior.  After lengthy discussion, the core team decided that features are still too confusing.  We came to the realization that we are likely conflating several different desired behaviors and capabilities, and forcing them all into a single concept.

So in 4.4, the core team decided to step back from key-value features, reverting to 4.3 behavior, and take another stab in the future to try to get it right.

In our internal discussions, we came up with a list of user stories that describe the behaviors and capabilities we've been trying to accomplish with "features."  Those are recorded at https://github.com/conda/conda/issues/6348


kalefranz:
I don't know why these TravisCI builds are hanging around, and showing the whole PR as failing.  I'll be working on figuring that out...  It might be something in the TravisCI web ui that needs to be deactivated.  But in that case, we'd probably have to backport the CircleCI config to 4.3.x, which I wasn't planning on doing (but we still can).

kalefranz:
Reviewer: Please merge if this PR looks good to you.

conda/core/repodata.py
```diff
@@ -591,7 +585,7 @@ def maybe_decompress(filename, resp_content):
         raise CondaIndexError("Invalid index file: {0}: {1}".format(join_url(url, filename), e))
 
 
-def make_feature_record(feature_name, feature_value):
+def make_feature_record_key_value(feature_name, feature_value):
```
mcg1969: It's not clear why we should even keep this here; I know we could possibly bring it back, but the easiest way to do that would be to effectively revert this entire PR

conda/models/index_record.py
```diff
@@ -326,11 +251,11 @@ class IndexJsonRecord(BasePackageRef):
     # track_features is being depracated and replaced with provides_features
     # NOTE: it's important that track_features comes before provides_features here
     track_features = _LegacyFeaturesField(required=False, default=(), default_in_dump=False)
-    provides_features = ProvidesFeaturesField(required=False, default=frozendict(),
-                                              default_in_dump=False)
+    # provides_features = ProvidesFeaturesField(required=False, default=frozendict(),
```
mcg1969: Same comment here. Why not just delete tehse

conda/models/match_spec.py
```diff
@@ -346,8 +342,8 @@ def _build_components(**kwargs):
         def _make(field_name, value):
             if field_name not in IndexRecord.__fields__:
                 raise CondaValueError('Cannot match on field %s' % (field_name,))
-            elif field_name == 'track_features':
-                field_name = 'provides_features'
+            # elif field_name == 'track_features':
```
mcg1969: Delete here too

kalefranz:
@mcg1969 I agree with all of your comments.  For the commented out code, the second commit in this PR was meant to remove all of it.  You found a few instances that I'd missed though.  I've made all changes indicated in your comments.

PR #6364, opened 2017-11-29
resolve #5417 ensure unlink order is correctly sorted

### LINKED ISSUE #5417: Respecting topological sorting for environment removal
Not sure if this is implemented in newer versions of `conda`, but it would be nice if environment removal performed a topological sort and uninstalled packages based on this sorting. Namely only removing packages that are not required for anything else first. For the most part this doesn't matter, but when `pre-unlink` scripts are involved it does.

kalefranz:
resolve #5417

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1,15 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from enum import Enum
 from genericpath import exists
 from logging import DEBUG, getLogger
 from os.path import join
 
-from enum import Enum
-
 from .index import get_reduced_index
 from .link import PrefixSetup, UnlinkLinkTransaction
-from .linked_data import PrefixData, linked_data
+from .linked_data import PrefixData
 from .._vendor.boltons.setutils import IndexedSet
 from ..base.context import context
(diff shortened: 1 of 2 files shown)


kalefranz:
Reviewer: Please merge if this PR looks good to you.

kalefranz:
The single test failure is a one-off 502 error.

conda/core/solve.py
```diff
@@ -397,6 +400,9 @@ def solve_for_diff(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NULL,
         # TODO: force_reinstall might not yet be fully implemented in :meth:`solve_final_state`,
         #       at least as described in the docstring.
 
+        unlink_precs = IndexedSet(reversed(sorted(unlink_precs,
```
nehaljwani: With this, can the following comment be removed?:
```
         # TODO: get the sort order correct for unlink_records
```

kalefranz: Yes, I'll remove it in that follow-on PR though that I spoke with you about.  Would rather not run CI again for this PR.

jakirkham:
Thanks. Really excited to see this. :)

conda/core/solve.py
jakirkham: What is the follow-on PR?

kalefranz: #6394

PR #6369, opened 2017-11-29
fix #6366 shell activation message

### LINKED ISSUE #6366: Better error message for when `conda activate` isn't a command
Tell users how to get `conda activate` as a command in 4.4 through a better error message.

kalefranz:
fix #6366

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -199,6 +199,7 @@ class CommandNotFoundError(CondaError):
             'render',
             'skeleton',
         }
+        # TODO: Point users to a page at conda-docs, which explains this context in more detail
         if command in activate_commands:
             from .base.context import context
             builder = ["Your shell has not been properly configured to use 'conda %(command)s'."]
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -221,7 +221,7 @@ class CommandNotFoundError(CondaError):
                 $ echo "conda activate" >> ~/%(config_file)s
 
             Previous to conda 4.4, the recommended way to activate conda was to modify PATH in
-            your ~/.%(config_file)s file.  You should manually remove the line that looks like
+            your ~/%(config_file)s file.  You should manually remove the line that looks like
 
                 export PATH="%(root_prefix)s/bin:$PATH"
 
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -6,7 +6,7 @@ import json
 from logging import getLogger
 import os
 import sys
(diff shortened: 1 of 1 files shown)


conda/exceptions.py
```diff
@@ -195,13 +199,46 @@ def __init__(self, command):
...
+            put conda's base (root) environment on PATH.  To do so, run
+
+                $ conda activate
+
+            in your terminal, or to put the base environment on PATH permanently, run
+
+                $ echo "conda activate" >> ~/%(config_file)s
+
+            Previous to conda 4.4, the recommended way to activate conda was to modify PATH in
+            your ~/.%(config_file)s file.  You should manually remove the line that looks like
```
nehaljwani: Extra `.` before `%(config_file)`

conda/exceptions.py
```diff
@@ -195,13 +199,46 @@ def __init__(self, command):
             'render',
             'skeleton',
         }
-        if command in build_commands:
+        if command in activate_commands:
+            from .base.context import context
+            builder = ["Your shell has not been properly configured to use 'conda %(command)s'."]
+            builder.append(dals("""
+            If your shell is Bash or a Bourne variant, enable conda for the current user with
```
nehaljwani: The message displayed below tells the user to modify their dotrc files. In some scenarios (like a shared user/service account), the dev might not want to modify the common user's dotrc file. In that scenario, one can just do 
```
. %(root_prefix)s/etc/profile.d/conda.sh
```
Should this be told to the user as part of the message before telling them that the conda function can always be made available by putting that line in their dotrc file, or should we let them infer it from the message?

kalefranz: I'm still struggling with everything here, specifically the right way to communicate it all.  There are *so many* different possibilities and scenarios.  Makes it difficult to explain everything in a linear way.

I think this PR is an improvement to what we had before, and we should merge it in prep for 4.4.0.  Also, in the very near term, I need to write up documentation that goes in conda-docs explaining how activation and deactivation are fundamentally shell manipulations, how every shell is different, what it even means to be a shell manipulation rather than a "normal" conda command, etc. etc.

Once we have that URL, I'll add it to this error message.

nehaljwani: Fair enough, I have pushed a patch over your branch to add a TODO. Hope you don't mind! :sweat_smile:

kalefranz: Don’t mind direct pushes like that at all. That’s why you have write access to my fork!

PR #6370, opened 2017-11-30
resolve #6194 warn when conda is outdated

### LINKED ISSUE #6194: conda only updates itself in the root environment
I was listening to a presentation about conda a couple of days ago where it was mentioned that conda aggressively auto-updates itself. Aggressive auto-updates can be problematic, because they can result in the quick spread of a buggy release that is no longer able to update itself, which has happened to me and co-workers in the past.  But in general they are a good thing, because they can help make sure that the installed user base is (a) not using outdated versions where problems were fixed long ago, and (b) are using a similar set of versions so that we don't have to reason about the effects of all possible version combinations.  

However, the main benefits of aggressively auto-updating only accrue if all or nearly all installations auto-update, which does not appear to be the case.  Right now, it looks to me like auto-updates happen only when people regularly use the root environment.  In my own workflows, I haven't installed anything in the root environment in many months; I always work on one of a number of more specialized environments instead.  I didn't even realize this until recently, when I ran into a bug fixed several months ago in conda, and which I expected not to apply to me.  I eventually fixed it by switching to the root environment and running ``conda update conda``, which was fine, but it was not at all clear that this would be needed.  I imagine other users who don't use the root environment may encounter similar issues.  I don't think conda is installed in any environment other than root, but maybe the other environments could warn on a ``conda install`` command if conda itself is more than a month or so out of date?

kalefranz:
resolve #6194

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -5,10 +5,14 @@ from enum import Enum
 from genericpath import exists
 from logging import DEBUG, getLogger
 from os.path import join
+import sys
+from textwrap import dedent
 
 from .index import get_reduced_index
 from .link import PrefixSetup, UnlinkLinkTransaction
 from .linked_data import PrefixData
+from .repodata import query_all
+from .. import __version__ as CONDA_VERSION
 from .._vendor.boltons.setutils import IndexedSet
 from ..base.context import context
 from ..common.compat import iteritems, itervalues, odict, string_types, text_type
@@ -22,6 +26,7 @@ from ..models.channel import Channel
 from ..models.dag import PrefixDag
 from ..models.dist import Dist
 from ..models.match_spec import MatchSpec
+from ..models.version import VersionOrder
 from ..resolve import Resolve, dashlist
 
 try:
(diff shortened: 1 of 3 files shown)


kalefranz:
The single failing appveyor test is a pexpect timeout.  Nothing to see here.

conda/core/solve.py
```diff
@@ -450,7 +455,27 @@ def solve_for_transaction(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NU
...
+
+        conda_newer_spec = MatchSpec('conda >%s' % CONDA_VERSION)
+        if not any(conda_newer_spec.match(prec) for prec in link_precs):
+            conda_newer_records = sorted(query_all(self.channels, self.subdirs, conda_newer_spec),
+                                         key=lambda x: VersionOrder(x.version))
+            if conda_newer_records:
+                latest_version = conda_newer_records[-1].version
+                sys.stderr.write(dedent("""
+
+                ==> WARNING: A newer version of conda exists. <==
```
nehaljwani: Can this instead be:
```
==> INFO: A newer version of conda is available!  <==
```

nehaljwani: Warning is could be considered something in negative sense.

kalefranz: I think it is negative.  I think we WANT people to actively be upgrading, and feel "uncomfortable" if they're on a non-latest version.  Pretty tired of people reporting errors that have already been fixed :(

nehaljwani: Will this warning be thrown on each operation involving solver even if I have pinned conda?

kalefranz: Yes, it will.  pip has basically the exact same behavior.  Are you concerned with being too annoying?

kalefranz: I really don't mind being annoying like this when people are using outdated versions of conda.  Lots of software does it.

In fact, I'd rather *just* be annoying, and change the default setting for `auto_update_conda` to `false`.  Crystal didn't want to do that though.

nehaljwani: In the scenario when the base environment is controlled by an admin and user is not allowed to modify it, I think it can be annoying. I think this warrants a discussion with a wider audience. Personally, I am okay with it being displayed always, but I'd still vote for the log level to be INFO instead of warning. Pip also allows users to control this behaviour with:
```
[global]
disable-pip-version-check = True
```
Maybe we could have something similar for Conda too?

kalefranz:
How about we start with a compromise? Keep “WARNING” in the message, and add the configuration parameter?

I think it’s easier to back it off to INFO, then scale it up to WARNING. If we get feedback indicating that we should back it off, I’d be open to it then.

nehaljwani:
As long as there is a configuration option to silence the warning, I am okay with it.

kalefranz:
Added a `notify_outdated_conda` configuration parameter.

PR #6391, opened 2017-12-04
canonicalize package names when comparing with pip

minrk:
because pip canonicalizes names, so `conda install jupyter_client` will show up as `jupyter-client` installed with pip, resulting in inaccurate `conda env export`.

It would be nice if conda itself did similar (or the same!) canonicalization, but it may be too late for that now.

### DIFF
--- a/conda_env/pip_util.py
+++ b/conda_env/pip_util.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, print_function
 import json
 import os
 from os.path import isfile, join
+import re
 import subprocess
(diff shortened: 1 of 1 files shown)


kalefranz:
LGTM if tests pass.  Will keep watching...

kalefranz:
Actually, looks like that appveyor run is just hung.  No idea why.  Backporting this PR to 4.4.x now.

kalefranz:
targeting 4.4.x via #6438

PR #6410, opened 2017-12-07
speed up the creation of repository

nouiz:
I'm working to create a clone on the local disk while the origin is on the network file system.

Here is a few optimization. In particular, the first commit remove many operation on the remove file system and speed up the operation by doing bigger read/write. The new size is the same as the cp Linux command line.


kalefranz:
Very cool.  I'll re-target this back to the 4.4.x branch.

kalefranz:
targeting 4.4.x via #6421

PR #6162, opened 2017-10-13
Allow redirecting the installation to another directory.

rainwoodman:
The key is to differentiate between prefix and dest. I overrode target_prefix to allow this. Most places we want to use dest, except the path that the replacement happens shall be prefix.

This PR serves as a demonstration of idea and needs to be carefully reviewed by an expert.

Issue #6104, opened 2017-10-08
Feature Request: relocated conda environments


rainwoodman:
I propose adding an argument `--dest-dir` to `conda create`.

The argument specifies an alternative location of the environment other than the intended `prefix`.

Here is an example:
```
conda create -n env -p /opt/my-python-application --dest-dir=$TMP/buildroot/opt/my-python-applicatoin python=3.6
```
This will write the the conda environment to $TMP/buildroot/my-python-application, but when these files are copied to `/opt/my-python-application` it will run as is.

One potential use-case is to create linux distribution packages (e.g. RPM) of entire conda environments; can simplify for example to create Rolls of certain computing environments for RocksClusters.

The package builders usually require the installation script to have such a feature like autotools' `DESTDIR`, because the assumption is that the builder does not otherwise have access to the intended target directory of installation. For example, rpmbuild prefixes make install with DESTDIR to allow non-root building and automatic packaging of files from the install-root. (see https://fedoraproject.org/wiki/How_to_create_an_RPM_package#.25install_section )

In addition, we have another use case for deployment of conda environments on HPC systems, but it may be too specialized to be mentioned here.

I suppose this is a trivial change -- if hopefully it doesn't get into the way of the path-fixers.

rainwoodman:
I am poking with a proposed patch.

kalefranz:
I’m not opposed to this, at least at the moment, but I’m also not sure it’ll be a trivial change. I guess we’ll find out!

kalefranz:
To give you some initial guidance, if it’s not already the case in the master branch, all prefix determination on what the target environment is for any run of conda should be coming from `context.target_prefix`. This is what we have to deconvolve, probably into a new property `context.dest_dir`. Under most cases they would be the same value, except when using a `—dest-dir` CLI flag. 

Then we have to find every place that `context.target_prefix is used, and also every place that value gets passed through to subordinate function/method calls, and make sure we don’t have to change the value to be the new `context.dest_dir`.

rainwoodman:
Thanks. I've put up some demonstration of idea in #6162 along these lines. My initial audit is that all but in the function for updating prefix shall be changed to dest_dir. We shall probably move the discussion over there and close this issue.

rainwoodman:
closing since the feature is already in.

PR #6413, opened 2017-12-08
target prefix override

kalefranz:
This PR does what I think is the goal of #6162 (and detailed in #6104) by @rainwoodman.  I ended up needing it myself for building some minimal docker containers.

Usage remains undocumented right now.  Easiest way to use is to set and environment variable, e.g.

    export CONDA_TARGET_PREFIX_OVERRIDE=/usr/local

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -180,6 +180,8 @@ class Context(Configuration):
     update_dependencies = PrimitiveParameter(False, aliases=('update_deps',))
     _verbosity = PrimitiveParameter(0, aliases=('verbose', 'verbosity'), element_type=int)
 
+    target_prefix_override = PrimitiveParameter('')
+
     # conda_build
     bld_path = PrimitiveParameter('')
     anaconda_upload = PrimitiveParameter(None, aliases=('binstar_upload',),
@@ -576,6 +578,7 @@ class Context(Configuration):
             'subdirs',
 # https://conda.io/docs/config.html#disable-updating-of-dependencies-update-dependencies # NOQA
 # I don't think this documentation is correct any longer. # NOQA
+            'target_prefix_override',  # used to override prefix rewriting, for e.g. building docker containers or RPMs  # NOQA
             'update_dependencies',
         )
         return tuple(p for p in super(Context, self).list_parameters()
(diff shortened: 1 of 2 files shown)


conda/base/context.py
```diff
@@ -871,7 +874,11 @@ def _get_user_agent(context_platform):
     libc_family, libc_ver = linux_get_libc_version()
     if context_platform == 'linux':
         from .._vendor.distro import linux_distribution
-        distinfo = linux_distribution(full_distribution_name=False)
+        try:
+            distinfo = linux_distribution(full_distribution_name=False)
+        except Exception as e:
+            log.debug('%r', e, exc_info=True)
+            distinfo = ('Linux', 'unknown')
```
kalefranz: This part has nothing to do with the PR.  Just got snuck in.  The `linux_distribution` function from the vendored `distro` can fail (it also makes a subprocess call to `lsb_release`, which I didn't know).  Not all that happy with how it works.  But oh well.

conda/core/path_actions.py
```diff
@@ -404,7 +404,9 @@ def verify(self):
 
         try:
             log.trace("rewriting prefixes in %s", self.target_full_path)
-            update_prefix(self.intermediate_path, self.target_prefix, self.prefix_placeholder,
+            update_prefix(self.intermediate_path,
+                          context.target_prefix_override or self.target_prefix,
+                          self.prefix_placeholder,
```
goanpeca: Maybe we could use named args (no necessarily kwargs) so that this function (and other functions throughout the codebase are  easier to read/understand on first read?

```python
update_prefix(
    prefix=self.intermediate_path,
    new_prefix=context.target_prefix_override or self.target_prefix, 
    placeholder=self.prefix_placeholder,
    mode=self.file_mode,
)
```

Just wondering what you think about this in general @kalefranz. Not a request for change :-p, but a request for comment ;-)

conda/base/context.py
goanpeca: Ok!

For some reason I usually prefer to do things like

```python
distinfo = ('Linux', 'unknown')
try:
    distinfo = linux_distribution(full_distribution_name=False)
except Exception as e:
    log.debug('%r', e, exc_info=True)
```

I know they have the same meaning :-p

conda/core/path_actions.py
kalefranz: I like that idea. I’ll try to start doing that.

conda/base/context.py
kalefranz: On this style issue, can we agree to disagree?

I feel like the way it is here is more idiomatic for try/except. Try, and then on failure, here’s how to recover. Rather than the default value case.

kalefranz:
@goanpeca let’s go ahead and merge this if you don’t mind

rainwoodman:
I figured that the meaning of the variable is that the environment will run properly after the entire tree is moved/copied to CONDA_TARGET_PREFIX_OVERRIDE. Perhaps keep this in mind when adding the documentation.

Issue #6403, opened 2017-12-06
Can't install conda development version


nouiz:
I want to modify conda to better suite our need. For this I followed the instruction at:

https://github.com/conda/conda#contributing

But this give me this error:

"""
Traceback (most recent call last):
  File "/home/bastienf/miniconda2/bin/conda", line 11, in <module>
    load_entry_point('conda', 'console_scripts', 'conda')() 
  File "/home/bastienf/miniconda2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 570, in load_entry_point 
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/bastienf/miniconda2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2750, in load_entry_point
    raise ImportError("Entry point %r not found" % ((group, name),))
ImportError: Entry point ('console_scripts', 'conda') not found
"""

Are those instruction up to date? I did this on an installation of miniconda2 version 4.3.30

sdvillal:
I'm also curious regarding what is best practice to have a conda development environment. I personally looked at several examples in the conda repo, and ended up using *utils/functions.sh* to get me there (```install_conda_dev``` is most useful).

sdvillal:
Just found also this: https://github.com/conda/conda/issues/6160.

nouiz:
Thanks for pointing to the other issue. Using "python -m conda ..." allow me to use my cloned version of conda.

PR #6421, opened 2017-12-08
reduce redundant sys call during file copying

kalefranz:
supersedes #6410
target is 4.4.x


kalefranz:
Single error in `main test py36` is 502 BAD GATEWAY.

nouiz:
That kind of error is probably a temporary error. Can you try to just restart that test run?

kalefranz:
Definitely just temporary. Was just waiting in appveyor to finish before merging.

nouiz:
thanks for the quick review/merge. Do you have time to rapidly check this issues from me? https://github.com/conda/conda/issues/6403

Mostly, I'm not able to have a good development environment for conda, so it make it harder for me to work on it.

Issue #6430, opened 2017-12-11
conda 4.4.0rc2 AssertionError on 'conda update --all'


kalefranz:
"command": "/home/nicolas/miniconda3/bin/conda update --all"

```
Traceback (most recent call last):
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/exceptions.py\", line 683, in __call__
    return func(*args, **kwargs)
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/cli/main.py\", line 78, in _main
    exit_code = do_call(args, p)
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/cli/conda_argparse.py\", line 75, in do_call
    exit_code = getattr(module, func_name)(args, parser)
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/cli/main_update.py\", line 14, in execute
    install(args, parser, 'update')
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/cli/install.py\", line 222, in install
    force_reinstall=context.force,
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/core/solve.py\", line 441, in solve_for_transaction
    force_remove, force_reinstall)
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/core/solve.py\", line 391, in solve_for_diff
    final_precs = self.solve_final_state(deps_modifier, prune, ignore_pinned, force_remove)
  File \"/home/nicolas/miniconda3/lib/python3.6/site-packages/conda/core/solve.py\", line 216, in solve_final_state
    assert len(matches_for_spec) == 1
AssertionError

```

kalefranz:
We need more information here.  It's possible that more than one installed package matched the spec.  It's possible that no installed package matched the spec.

PR #6435, opened 2017-12-11
add debug exception for #6430

kalefranz:
This PR does not _close_ #6430, but it adds an exception with additional information for the next time anybody runs across it.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -12,7 +12,8 @@ from .index import get_reduced_index
 from .link import PrefixSetup, UnlinkLinkTransaction
 from .linked_data import PrefixData
 from .repodata import query_all
-from .. import __version__ as CONDA_VERSION
+from .. import CondaError, __version__ as CONDA_VERSION
+from .._vendor.auxlib.ish import dals
 from .._vendor.boltons.setutils import IndexedSet
 from ..base.context import context
 from ..common.compat import iteritems, itervalues, odict, string_types, text_type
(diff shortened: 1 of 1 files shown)


conda/core/solve.py
```diff
@@ -219,7 +220,14 @@ def solve_final_state(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NULL,
...
             if matches_for_spec:
-                assert len(matches_for_spec) == 1
+                if len(matches_for_spec) != 1:
+                    raise CondaError(dals("""
+                    Conda encountered an error with your environment.  Please report an issue
+                    at https://github.com/conda/conda/issues/new.  In your report, please include
+                    the output of 'conda info' and 'conda list' for the active environment, along
+                    with the command you invoked that resulted in this error.
+                      matches_for_spec: %s
+                    """) % matches_for_spec)
```
goanpeca: Looks good, but don't we have almost the exact same text somewhere else? Does it make sense to reuse it instead of rewriting?

kalefranz: We do, except it's only displayed for exceptions that don't subclass `CondaError`.  This in particular is a `CondaError`, which means the full stack trace won't be shown (which is what the intention is here).

This exception is also meant to be temporary.  It'll be removed as soon as we understand the root cause of the issue.

Note that I pulled #6430 from our auto-reports.  There was no user who created an issue for us to interact with.

goanpeca: Got it!

PR #6437, opened 2017-12-11
add information to displayed error report

kalefranz:
To be compliant with GDPR (according to Mathew Lodge), for the error upload report, we need to indicate what information is being uploaded, why, and how that information will be used.  This PR should bring us into GDPR compliance.


kalefranz:
The first commit in this PR is currently showing as the commit for #6436.  That PR should be merged first.

PR #6443, opened 2017-12-13
fix imports for conda-build

kalefranz:


kalefranz:
Going to go ahead an merge this one.  It's just a rearrangement of code, and only needed until 4.4.0 is tagged.

PR #6446, opened 2017-12-13
fix #6419 set PS1 in python code

### LINKED ISSUE #6419: conda 4.4 weird prompt addition in zsh
Before conda 4.4 my prompt with zsh would look like this, when I'm in my conda env `stable`:

screenshot 2017-12-08 13 03 34

After the update to conda 4.4 (using the blog instructions), I have this:

screenshot 2017-12-08 13 03 43

Note the addition of '\[\]' that was added after the environment ID.

How can I get rid of that?

Using zsh 5.4.2 on macOS 10.12.6

kalefranz:
fix #6419

This PR slightly reworks how PS1 is set for Bourne shells.  Basically setting PS1 through our python code, and not doing logic at the shell level.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -51,7 +51,7 @@ class Activator(object):
             self.shift_args = 0
 
             self.unset_var_tmpl = 'unset %s'
-            self.set_var_tmpl = 'export %s="%s"'
+            self.set_var_tmpl = "export %s='%s'"
             self.run_script_tmpl = '. "%s"'
 
         elif shell == 'csh':
@@ -73,7 +73,7 @@ class Activator(object):
             self.shift_args = 0
 
             self.unset_var_tmpl = 'del $%s'
-            self.set_var_tmpl = '$%s = "%s"'
+            self.set_var_tmpl = "$%s = '%s'"
             self.run_script_tmpl = 'source "%s"'
 
         elif shell == 'cmd.exe':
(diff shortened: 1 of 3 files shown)


nehaljwani:
I am a bit skeptical about this patch. Not sure what's happening or it is me doing something wrong, but with rc2, I have:
screenshot from 2017-12-15 16-26-42
And with this patch, I have:
screenshot from 2017-12-15 16-27-09

nehaljwani:
Okay, this is weird:
```zsh
/tmp # ❯❯❯ echo $PS1
${SSH_TTY:+"%F{9}%n%f%F{7}@%f%F{3}%m%f "}%F{4}${_prompt_sorin_pwd}%(!. %B%F{1}#%f%b.)${editor_info[keymap]} 

/tmp # ❯❯❯ env | grep -i PS1
PS1=[\u@\h \W]\$ 
```

kalefranz:
Yeah this isn’t easy to get right. So many edge cases.

I think in general the strategy is correct. When possible we should be doing everything at the python level, including configuring the prompt. 

It might take some more iteration on the exact details. 

I don’t think it makes anything worse for zsh, since zsh wasn’t quite right to begin with. It also worked on zsh for me. 

If this is the strategy we want to take, I think the most important part is that we don’t regress for bash, dash/ash, etc. There are supposed to be tests for this, but I still don’t completely trust them yet. So I guess I’d like for you to independent test bash and dash/ash.

kalefranz:
Ok, I pushed an update, which does *not* `export PS1`.  If somebody else has that variable exported somewhere, that's fine.  But it's not going to be conda that does it.

PR #6458, opened 2017-12-14
add timing instrumentation

kalefranz:
This PR has a companion PR pointed at 4.4.x

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -9,13 +9,15 @@ import logging
 from logging import CRITICAL, Formatter, NOTSET, StreamHandler, WARN, getLogger
 import os
 from os import chdir, getcwd
-from os.path import join, isfile
+from os.path import dirname, isdir, isfile, join
 import sys
 from time import time
 
 from .compat import StringIO
 from .path import expand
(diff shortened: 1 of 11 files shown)


conda/common/io.py
```diff
@@ -156,3 +164,70 @@ def attach_stderr_handler(level=WARN, logger_name=None, propagate=False, formatt
...
+
+class time_recorder(ContextDecorator):  # pragma: no cover
+    start_time = None
+    record_file = expand(join('~', '.conda', 'instrumentation-record.csv'))
+
+    def __init__(self, entry_name):
+        self.entry_name = entry_name
+
+    def __enter__(self):
+        enabled = os.environ.get('CONDA_INSTRUMENTATION_ENABLED')
```
kalefranz: Using a hard-coded environment variable here rather than putting it in `context` because `common` can't import from `conda.base`, and thus we don't have access to `context`.

kalefranz:
@msarahan I fixed the smoketest issues.  Let's go ahead and merge this now.

kalefranz:
Approved by Mike, so long as we went back to doing the conda-build smoke test.  Which we have.  So I'm going to go ahead and merge this.

PR #6463, opened 2017-12-15
csh wrapper

kalefranz:
Basic wrapper for csh.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -175,7 +175,7 @@ class Activator(object):
         elif command == 'activate' and len(remainder_args) > 1:
             from .exceptions import ArgumentError
             raise ArgumentError('activate does not accept more than one argument:\n'
-                                + str(sys.argv))
+                                + str(remainder_args))
         elif command != 'activate' and remainder_args:
             from .exceptions import ArgumentError
             raise ArgumentError('%s does not accept arguments\nremainder_args: %s'
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -49,6 +49,7 @@ class Activator(object):
             self.script_extension = '.sh'
             self.tempfile_extension = None  # write instructions to stdout rather than a temp file
             self.shift_args = 0
+            self.command_join = '\n'
 
             self.unset_var_tmpl = 'unset %s'
             self.set_var_tmpl = "export %s='%s'"
@@ -60,6 +61,7 @@ class Activator(object):
             self.script_extension = '.csh'
             self.tempfile_extension = None  # write instructions to stdout rather than a temp file
             self.shift_args = 0
+            self.command_join = ';\n'
 
             self.unset_var_tmpl = 'unset %s'
             self.set_var_tmpl = 'setenv %s "%s"'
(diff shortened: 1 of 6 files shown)


kalefranz:
Let's get https://github.com/conda/conda/pull/6446 merged first

kalefranz:
Right now, I believe csh is mostly working, even though the integration test doesn't pass for it.  Integration test passes for tcsh.  Because of that integration test issue, I don't yet think we're done with the csh/tcsh story.  But this PR is a good start that lays out the framework.

shell/etc/csh/login.d/conda.csh
```diff
@@ -0,0 +1,43 @@
+
+# Recommended way to make the conda command available in csh is
+#   $ sudo ln -s <CONDA_ROOT>/etc/csh/login.d/conda.sh /etc/csh/login.d/conda.sh
```
nehaljwani:
```
s/conda.sh/conda.csh
```
I am not sure, but shouldn't this be `/etc/profile.d/conda.csh` ? That's where `/etc/csh.login` sources stuff from.

kalefranz: Really? Must be different on every distribution. That sucks. I think I looked at csh on Ubuntu. Guess I’ll check a couple different distros and see if I can tease out what’s most common

conda/activate.py
```diff
@@ -165,8 +173,9 @@ class Help(CondaError):  # NOQA
             from .exceptions import ArgumentError
             raise ArgumentError("invalid command '%s'" % command)
         elif command == 'activate' and len(remainder_args) > 1:
+            print(sys.argv, file=sys.stderr)
```
nehaljwani: Is this deliberate or a debug statement?

kalefranz: yeah there's a problem here in this block

shell/etc/csh/login.d/conda.csh
kalefranz: Ok, on Ubuntu, `csh` doesn't pre-populate any files in `/etc`, but `tcsh` does.  Maybe that's the difference I'm seeing...

kalefranz:
```
root@47cf55bec20a:/# apt-get install csh
...
root@47cf55bec20a:/# echo "echo hello world" > /etc/profile.d/hello.csh
root@47cf55bec20a:/# . !$
. /etc/profile.d/hello.csh
hello world
root@47cf55bec20a:/# csh
#
```

I guess that didn't work.

nehaljwani:
This functionality for csh is not 100% complete yet, but I'm still going to go ahead and merge this.
The part missing here is that `$prompt` is not exported by csh (at least in my case), so it is not available to python. So after activating an environment, only the modifier is displayed and on deactivation, nothing is displayed.
```
[nwani@centos7-nwani ~]$ sudo docker exec -it 519cb4b3763f /bin/csh
[root@519cb4b3763f /build_scripts]# source /opt/conda/etc/profile.d/conda.csh
[root@519cb4b3763f /build_scripts]# conda activate root
(base) echo $prompt
(base)
(base) conda deactivate

echo $prompt

set prompt="[%n@%m %c]%#"
[root@519cb4b3763f /build_scripts]#
```

kalefranz:
Agreed. The prompt modifier is not yet complete. At least we have a start though.

PR #6090, opened 2017-10-06
add activate/deactivate scripts for powershell

HyundongHwang:
I wrote this PR to clean up issue #1519. 
In this case, one bat script can not be used together with window terminal and powershell terminal.  Because there is no general way to move the environment variables set in bat script to the powershell terminal. Therefore, I solved the problem by creating and using a new ps script for the powershell terminal.

Issue #626, opened 2014-03-28
Powershell activate and deactivate


asmeurer:
See https://github.com/Liquidmantis/PSCondaEnvs.

rossant:
Will #913 work in Powershell? If not, would it be worth just adding the [ps scripts](https://github.com/Liquidmantis/PSCondaEnvs) to the distribution?

Ivoz:
Is there anything bad about just including ps scripts while you figure out #913 ? I thought conda was supposed to be advertised as more featureful than virtualenv :P

liujimj:
Just to chime in, I also wasted about an hour of time figuring out why my environments weren't activating.

Anaconda envs are advertised as full-featured replacements for venv, so I would expect Powershell support.

msarahan:
If @LiquidMantis or anyone else wants to submit the ps scripts that @rossant mentioned, I'll merge them.    It would be nice to not include another license file with Conda.  I the meantime, we are reviewing #913 and its newer derivatives and hope to merge it soon.

Liquidmantis:
I'm fine with them being added, and was going to submit a PR now, but it looks like conda has been reworked and the old bat files have been removed anyway so I'm not sure how to proceed.

msarahan:
Things are in a little bit of flux right now.  The activate scripts were moved out of conda, into conda-env.  This has unfortunately made the activate scripts harder to test, since they effectively depend on two packages.  We will be discussing this after the Anaconda 2.4.1 release, and hopefully coming up with a better arrangement.  If you put a PR in against conda-env, we'll be happy to consider it, and we'll take care of any moves necessary thereafter.

masaeedu:
@msarahan It looks like the Anaconda 2.5 release is out now. Has there been any progress on this? It would greatly improve the Windows UX if this could be fixed. :pray:

vmuriart:
@Liquidmantis Seems like the old `bat` files have been added back. I just tested your scripts from `pscondaenvs` and they work with `conda 4.0.5`. 

#913 was closed in favor or #1727, but it didn't address the powershell issue.
@msarahan would you still accept the pr if @Liquidmantis or anyone else submits them?

msarahan:
Sorry for the delay here.  Absolutely.

[...]

PR #6471, opened 2017-12-15
[WIP] conda 4.4 powershell wrapper

kalefranz:


kalefranz:
https://github.com/conda/conda/issues/626

mbargull:
Some code regarding prompt support can be found in the previous PR #6090 by @HyundongHwang.

mbargull:
Prompt code is also in https://github.com/BCSharp/PSCondaEnvs/blob/052698cf442320ac635eac2391dcb7f0766a49e0/activate.ps1#L129-L133

kalefranz:
Yeah as soon as I'm able to finally wrap up https://github.com/conda/conda/pull/6518 I was going to see if @BCSharp was interested in contributing the official powershell wrapper that makes use of the new framework.

My biggest unknown at the moment is the "right" way to initialize powershell and bootstrap in the conda-specific code.  E.g. for bash we have option (1) modify `~/.bashrc` with something like `. /path/to/conda_base/etc/conda/conda.sh` (or the `eval` that'll fall out of #6518), or option (2) adding a file `/etc/profile.d/conda.sh` that does something similar.

njalerikson:
It would appear that the "proper" way in PowerShell is to add the conda.ps1 into the $profile script. The only issue I see with that is that the user will need to modify their ExecutionPolicy before Powershell will (1) even read their own $profile script and (2) also read the conda.ps1 script.

Given the ExecutionPolices, I think that you would want to get the conda.ps1 script signed (whatever that means and however one would do that).

zooba:
I would recommend doing it this way:
* rename to `conda.psm1` (that makes it a module)
* add `Export-ModuleMember -Function conda` at the end of the file
* sign the .psm1 with your usual code signing certificate
* install/copy it to either `%UserProfile%\Documents\WindowsPowerShell\Modules\Conda` or `%ProgramFiles%\WindowsPowerShell\Modules\Conda` (for both 32 and 64-bit variants, I believe)

PowerShell will scan the files in these directories and automatically import them when the user invokes the exported command, so you shouldn't have to configure anything other than the installation.

Also, for "good" PowerShell style, I'd recommend:
* rename the function to `Invoke-Conda`
* add `New-Alias conda Invoke-Conda -Force` and `Export-ModuleMember -Alias conda` at the end of the module
* define as many parameters as feasible ([this page](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-6) should help) and avoid using `$args`
* depending on how committed you are here, bringing the commands up a level and having (e.g.) `Install-CondaPackage` etc. commands that simply turn into `Invoke-Conda install ...` would make this feel like an impressively native experience for PS users (though it may also make a nice project for "someone else"). Or dynamic parameter sets (see the link above) would also be neat.

kalefranz:
Thanks @zooba that's exactly what we needed.

kalefranz:
https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-powershell-core-on-macos-and-linux?view=powershell-6#paths

* `$PSHOME` is `/opt/microsoft/powershell/6.0.0/`
* User modules will be read from `~/.local/share/powershell/Modules`
* Shared modules will be read from `/usr/local/share/powershell/Modules`
* Default modules will be read from `$PSHOME/Modules`

kalefranz:
Update on where we are with shell wrappers in current master (4.6.x).

First, all the shell wrapper files have moved slightly.  They're now at https://github.com/conda/conda/blob/710d30a9cc2e2417c00c0fcdc188eeff2216fb2f/conda/shell.  We currently have five varieties:

* generic posix shell (ash, bash, dash, zsh, busybox, etc) via `conda.sh`
* the cmd.exe prompt via `conda.bat` (and lots of other helper bat files)
* tcsh (though I don't think csh works currently) via `conda.csh`
* fish via `conda.fish`
* xonsh via `conda.xsh`

In 4.6, there is a new `conda init` command specifically to help users set up shell wrappers correctly.  The logic and implementation for `conda init` is in https://github.com/conda/conda/blob/710d30a9cc2e2417c00c0fcdc188eeff2216fb2f/conda/core/initialize.py. The approach we've taken in `conda init` for setting up shells is borrowed from https://direnv.net/.  The bash case, for example, adds a block of code to `~/.bashrc` / `~/.bash_profile` that is basically `eval "$(/path/to/bin/conda shell.bash hook)"`.  The other shells work the same way, except for `cmd.exe`, where we jump through a bunch of different hoops.  On Windows, we've put in code to let `conda init` start making registry modifications, if that's helpful here.  Right now, the two modifications we're making are (1) making the `conda` command available to all cmd.exe shells (not just the "Anaconda Prompt") by setting `HKEY_CURRENT_USER\Software\Microsoft\Command Processor\AutoRun` and (2) enabling `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` as the python.org python3.6 installers do.

The general strategy of keeping the bulk of the `conda activate` logic in python (specifically [conda/activate.py](https://github.com/conda/conda/blob/710d30a9cc2e2417c00c0fcdc188eeff2216fb2f/conda/activate.py)) and making the shell parts as thin of wrapper as possible has proven so far to be stable and successful.  Powershell is already built into the conda/activate.py file, although it may need some tweaking once it's actually being used.  All shell activation tests are located in https://github.com/conda/conda/blob/710d30a9cc2e2417c00c0fcdc188eeff2216fb2f/tests/test_activate.py.  We probably won't need any specific unit tests for powershell since they should already be written for the most part.  We'll need to add an integration test or two in https://github.com/conda/conda/blob/710d30a9cc2e2417c00c0fcdc188eeff2216fb2f/tests/test_activate.py#L1255-L1256.

kalefranz:
This PR was superseded by https://github.com/conda/conda/pull/7794

PR #5912, opened 2017-09-04
Fixes for Fish shell completion code

luc-j-bourhis:
It is great conda ships with support for the Fish shell but the command-line completion code had several issues: this pull request is a proposition to fix them. I tested it works on MacOS Sierra and Ubuntu Xenial.

PR #6475, opened 2017-12-17
fish update for conda 4.4

kalefranz:
This PR makes progress on bringing the fish wrapper in line with the other 4.4 shell wrappers.  I'm not a fish expert, so I'm sure it's probably not perfect yet.

One particular problem with the integration test... I can't get the `pexpect` library to collect stdout from a fish terminal.  Not sure exactly what's going on there.  With manual testing though, this PR at least appears to have basic functionality.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -94,12 +94,12 @@ class Activator(object):
             self.run_script_tmpl = '@CALL "%s"'
 
         elif shell == 'fish':
-            self.pathsep_join = ' '.join
+            self.pathsep_join = '" "'.join
             self.path_conversion = native_path_to_unix
             self.script_extension = '.fish'
             self.tempfile_extension = None  # write instructions to stdout rather than a temp file
             self.shift_args = 0
-            self.command_join = '\n'
+            self.command_join = ';\n'
 
             self.unset_var_tmpl = 'set -e %s'
             self.export_var_tmpl = 'set -gx %s "%s"'
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -1,9 +1,3 @@
-#
-# Conda environment activate / deactivate functions for fish shell v2.2.0+.
-#
-# Ivan Smirnov (C) 2015
-#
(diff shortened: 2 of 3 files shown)


RSully:
Any reason for `source [CONDA_INSTALL_LOCATION]/etc/fish/conf.d/conda.fish` instead of using the `conda` command to find the path? As long as users set their path first I liked that approach. I haven't used 4.4 though, so I'm not sure if there is a technical limitation to doing that now.

kalefranz:
> Any reason for `source [CONDA_INSTALL_LOCATION]/etc/fish/conf.d/conda.fish` instead of using the `conda` command to find the path like previously?

Previous version assumes the conda root location is already on `PATH`, which this doesn't.  Also, `conda info --root` will probably add several hundred milliseconds to your shell startup time, so for something like that, it's best just to hard-code it.

Good explanation, even though it's focused on bash the concepts are the same, is in: https://www.anaconda.com/blog/developer-blog/how-to-get-ready-for-the-release-of-conda-4-4/

jaimergp:
LFTM! Very snappy as well :)

I would try to integrate the changes listed in [PR 5912](https://github.com/conda/conda/pull/5912/commits), because the completions I coded back then are painfully slow.

kalefranz:
Thanks @jaimergp for the review!  I'm going to go ahead and merge this PR as-is then.  And we'll also try to bring #5912 in for 4.4.0 also.

jaimergp:
I had to add `set _CONDA_EXE ~/.local/anaconda/bin/conda` to my `etc/fish/conf.d/conda.fish`, but I guess that my installation process was faulty. A proper installation should add that line automatically, right?

kalefranz:
Using the script in dev mode, maybe.  Most users shouldn't have to set that.  Except... I forgot something really important to ensure that's the case.

https://github.com/conda/conda/blob/4.4.x/utils/functions.sh#L199-L201

When conda is packaged up, and then under normal installs, that should be templated in correctly at the top of the file.  I forgot to add that though for fish.  I'll make sure I do that when I bring #5912 in.

PR #6479, opened 2017-12-17
move shell inside conda directory

kalefranz:
This is a not insignificant structural change to the repository. It's one we should make before releasing 4.4.0 final.

The reason is, I've recently been contemplating the addition of a `conda initialize` command, which could aid the transition from the old `export PATH="..."` modification (e.g. for Bourne shells) to the new method of sourcing the `conda.sh` file.  It could also programmatically be much more comprehensive, in terms of asking users which shells they wanted to enable, and if they wanted conda to attempt to modify the shell's initialization files.  It would also be helpful in the circumstance where we wanted to make the conda executable and base environment again pip-installable.

For now, this structural change to the repository is only an anticipation of and hedge for future needs.  The only cost is the possibility of twice-duplicating the contents of the scripts contained in `conda/shell` within actual conda package.


jakirkham:
How does this impact this content being installed?

kalefranz:
It shouldn't at all. In the `4.4.x` and `4.5.x` releases, the `utils/functions.sh` script was modified appropriately within this PR.

Are you running into a problem though?

jakirkham:
Right, was just learning about this. Thanks for confirming. :)

Actually got macOS to build locally. Windows IDK we will see what CI says.

kalefranz:
It's a shame we don't yet have https://github.com/conda/conda/pull/6518 merged and 4.6 ready to go.  All this stuff is so much simpler.  At least from the build perspective.  #6518 is just a large volume of changes though, so it's going to take a while for someone to get around to reviewing it.

PR #6489, opened 2017-12-19
remove dead code

kalefranz:
This PR just removes some unused, dead code, and also adds a couple tests.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -6,7 +6,7 @@ from logging import DEBUG, getLogger
 
 from .base.constants import DEFAULTS_CHANNEL_NAME, MAX_CHANNEL_PRIORITY
 from .base.context import context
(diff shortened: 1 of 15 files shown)


conda/resolve.py
```diff
@@ -386,22 +386,6 @@ def ms_depends(self, dist):
             self.ms_depends_[dist] = deps
         return deps
 
-    def depends_on(self, spec, target):
```
mcg1969: We really don't use this? I could have sworn we have some sort of `conda search` command that can find packages that depend on another.

mcg1969:
Merge conflicts

conda/resolve.py
kalefranz: I couldn't find anywhere that we do.  Not any more anyway.  `conda search` probably doesn't use resolve at all any longer, so maybe that's what happened.

PR #6495, opened 2017-12-19
skip an unnecessary environments.txt rewrite

kalefranz:


conda/core/envs_manager.py
```diff
@@ -39,6 +39,7 @@ def _clean_environments_txt(environments_txt_file, remove_location=None):
         return ()
 
     environments_txt_lines = list(yield_lines(environments_txt_file))
+    original_environments_txt_hash = hash(tuple(environments_txt_file))
```
mcg1969: Should these be sorted?

kalefranz: Thanks for looking at this PR.  I don't think it matters that they're sorted.  What matters is that it has the same order as `hash(real_prefixes)` in line 52.  However, I think I should add a test that ensures this is all working correctly.

huguesv:
I patched my local copy with this change. I can confirm that this fixes the issue for me.

kalefranz:
Thanks @huguesv!

conda/core/envs_manager.py
mcg1969: But does order matter in environments.txt? If not, and the goal is to eliminate unecessary rewrites, then it would seem to me that we should be doing a comparison that matters.

Also, I know a hash is very, very unlikely to let false equivalencies pass through, but why take a chance? What benefit is there to avoiding a full comparison?

PR #6502, opened 2017-12-20
fixes for fish code completion

kalefranz:
supersedes #5912 to target 4.4.x

Also adds templating of `_CONDA_EXE` and `_CONDA_ROOT` for the installed versions of `conda.fish`.

### DIFF
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -87,11 +87,13 @@ end
 
 # Faster but less tested (?)
 function __fish_conda_commands
-  command echo -e "activate\ndeactivate" ;and ls --color=none (conda info --root)/bin/conda-* | sed -r 's/^.*conda-([a-z]+)/\1/'
+  string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda/cli/main_*.py
+  echo activate
+  echo deactivate
 end
 
 function __fish_conda_envs
-  command echo root ;and ls -1 --color=none (conda info --root)/envs/
+  basename -a $_CONDA_ROOT/envs/*/
 end
 
 function __fish_conda_packages
--- a/utils/functions.sh
+++ b/utils/functions.sh
@@ -198,7 +198,9 @@ install_conda_shell_scripts() {
 
     mkdir -p "$prefix/etc/fish/conf.d/"
     rm -f "$prefix/etc/fish/conf.d/conda.fish"
-    cp "$src_dir/conda/shell/etc/fish/conf.d/conda.fish" "$prefix/etc/fish/conf.d/conda.fish"
+    echo "set _CONDA_EXE \"$conda_exe\"" > "$prefix/etc/fish/conf.d/conda.fish"
+    echo "set _CONDA_ROOT \"$prefix\"" >> "$prefix/etc/fish/conf.d/conda.fish"
+    cat "$src_dir/conda/shell/etc/fish/conf.d/conda.fish" >> "$prefix/etc/fish/conf.d/conda.fish"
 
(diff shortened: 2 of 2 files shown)


conda/shell/etc/fish/conf.d/conda.fish
```diff
@@ -87,11 +87,13 @@ end
...
 function __fish_conda_commands
-  command echo -e "activate\ndeactivate" ;and ls --color=none (conda info --root)/bin/conda-* | sed -r 's/^.*conda-([a-z]+)/\1/'
+  string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda/cli/main_*.py
+  echo activate
+  echo deactivate
 end
 
 function __fish_conda_envs
-  command echo root ;and ls -1 --color=none (conda info --root)/envs/
+  basename -a $_CONDA_ROOT/envs/*/
```
kalefranz: Better here might be using the output of `conda info --envs`.  Not sure how to do that though.  Maybe something like

    command conda info --envs | awk '{print $1}'

?

conda/shell/etc/fish/conf.d/conda.fish
```diff
@@ -87,11 +87,13 @@ end
 
 # Faster but less tested (?)
 function __fish_conda_commands
-  command echo -e "activate\ndeactivate" ;and ls --color=none (conda info --root)/bin/conda-* | sed -r 's/^.*conda-([a-z]+)/\1/'
+  string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda/cli/main_*.py
+  echo activate
+  echo deactivate
```
kalefranz: Maybe should also add here

    string replace -r '.*conda-([a-z]+)$' '$1' $_CONDA_ROOT/bin/conda-*

or something like that.  All of the commands that come with conda-build, for example, have an executable associated with them.  `bin/conda-build` and `bin/conda-skeleton`, for example.

kalefranz:
CC @jaimergp

conda/shell/etc/fish/conf.d/conda.fish
jaimergp: I used `conda info --envs | awk 'NR>2{print $1}' ` at some point but switched to the `envs` listing due to performance issues (but only philosophically maybe, because `conda info --root` is equally slow). Maybe in 4.4 now that's not a problem and that way we can list additional envs not present in the default directory.

conda/shell/etc/fish/conf.d/conda.fish
jaimergp: That's what we were using in the *slow* implementation, the `conda-*` binaries. I guess we have no reason to stop doing that.

conda/shell/etc/fish/conf.d/conda.fish
kalefranz: OK, maybe we should leave it just to this PR for 4.4.0, and then you all can play with this again after you get a feel for things with 4.4.

kalefranz:
@jaimergp Are you ok going with this for 4.4.0, and then iterating on other optimizations in future releases?

kalefranz:
I cancelled the appveyor build for this PR.  Don't think it's necessary.

kalefranz:
The conda-build failure has nothing to do with this PR.

jaimergp:
> @jaimergp Are you ok going with this for 4.4.0, and then iterating on other optimizations in future releases?

@kalefranz, sure! We can iterate later. What do you think, @luc-j-bourhis ?

luc-j-bourhis:
Without this pull request, it does not work on MacOS and with only the commit fixing this, it is still too slow to be usable, so yes, definitively, I reckon to get this out asap, and tweak further later.

kalefranz:
Conda 4.4.0 is in canary right now if you want to play with it. Will probably be released to defaults today or tomorrow, barring anything critical being raised anywhere.

PR #6524, opened 2017-12-22
resolve #6523 don't prune with --update-all

### LINKED ISSUE #6523: 'conda update --all' causes "'python' is not recognized as an internal or external command"
repro steps:
1. fresh install of latest version of miniconda
2. `conda update --all` and yes to all questions
3. restart 'Anaconda Prompt'

and you will see
```
'python' is not recognized as an internal or external command,
operable program or batch file.
```

kalefranz:
resolve #6523

Until we get https://github.com/conda/constructor/issues/138 resolved and also until it has some time to settle in, we're just going to have to be more conservative with what `--update-all` is allowed to do.  If users want prune behavior, there's always the `--prune` flag.


kalefranz:
Appveyor test failures unrelated to this PR.

PR #6526, opened 2017-12-22
fix #6520 --use-local flag

### LINKED ISSUE #6520: --use-local flag not working in 4.4.0
For some time I've been doing this:
```
conda build --old-build-string --python 2.7
conda install taxcalc=0.0.0 --use-local --yes
```
But that `conda install` command now fails under conda 4.4.0 and 4.4.1.

I have been able to work around this problem by doing this:
```
conda build --old-build-string --python 2.7
conda install --offline ~/anaconda/conda-bld/osx-64/taxcalc-0.0.0-py27_0.tar.bz2
```
Why does my old code fail under conda 4.4.0 and conda 4.4.1?

The failure is that it says it cannot find the package, but it is where it has always been built:
`~/anaconda/conda-bld/osx-64/taxcalc-0.0.0-py27_0.tar.bz2`

kalefranz:
fix #6520

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -160,6 +160,7 @@ class Context(Configuration):
     migrated_custom_channels = MapParameter(string_types)  # TODO: also take a list of strings
     override_channels_enabled = PrimitiveParameter(True)
     show_channel_urls = PrimitiveParameter(None, element_type=(bool, NoneType))
+    use_local = PrimitiveParameter(False)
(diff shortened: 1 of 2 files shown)


kalefranz:
Appveyor test failures unrelated to this PR.

kalefranz:
Appveyor test failures unrelated to this PR.

Issue #6546, opened 2017-12-23
I got: Solving environment: failed


dacliem:
After my Conda environment activated, I issue this command:
`conda update conda`

And I received an error message:

> Solving environment: failed

matches_for_spec:
    - defaults::certifi-2017.11.5-py36hb8ac631_0
    - defaults::h5py-2.7.1-py36he54a1c3_0
    - defaults::hdf5-1.10.1-h98b8871_1
    - defaults::icc_rt-2017.0.4-h97af966_0
    - defaults::intel-openmp-2018.0.0-hd92c6cd_8
    - defaults::keras-2.0.8-py36h65e7a35_0
    - defaults::libprotobuf-3.4.1-h3dba5dd_0
    - defaults::mkl-2018.0.1-h2108138_4
    - defaults::numpy-1.12.1-py36hf30b8aa_1
    - defaults::pip-9.0.1-py36h226ae91_4
    - defaults::protobuf-3.4.1-py36h07fa351_0
    - defaults::python-3.6.3-h3b118a2_4
    - defaults::pyyaml-3.12-py36h1d1928f_1
    - defaults::scipy-1.0.0-py36h1260518_0
    - defaults::setuptools-36.5.0-py36h65f9e6e_0
    - defaults::six-1.11.0-py36h4db2310_1
    - defaults::tensorflow-1.1.0-np112py36_0
    - defaults::vc-14-h2379b0c_2
    - defaults::vs2015_runtime-14.0.25420-0
    - defaults::werkzeug-0.12.2-py36h866a736_0
    - defaults::wheel-0.30.0-py36h6c3ec14_1
    - defaults::wincertstore-0.2-py36h7fe50ca_0
    - defaults::yaml-0.1.7-hc54c509_2
    - defaults::zlib-1.2.11-h8395fce_2

=====
My `conda info` output:

     active environment : tensorflow
    active env location : D:\Anaconda3\envs\tensorflow
            shell level : 2
       user config file : C:\Users\liemld\.condarc
 populated config files :
          conda version : 4.4.3
    conda-build version : 3.0.27
         python version : 3.6.3.final.0
       base environment : D:\Anaconda3  (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/win-64
                          https://repo.continuum.io/pkgs/main/noarch
                          https://repo.continuum.io/pkgs/free/win-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/win-64
[...]

PR #6551, opened 2017-12-23
add additional information to the #6546 exception

kalefranz:
#6546


kalefranz:
After #6565, we should probably just remove this exception altogether.  But still, with `MatchSpec('')` being equivalent to `*`, I think it's best to leave the exception for the time being.

PR #6518, opened 2017-12-22
conda initialize

kalefranz:
toward #6507

PR #6552, opened 2017-12-23
fix #6549 'conda activate' has sys.path problem with cwd

### LINKED ISSUE #6549: source activate <env> does not work from command line in a folder containing io.py
I just noticed (after quite a bit of head scratching) that I cannot activate a python environment using source activate from a folder containing a file called io.py:

```
lsb_release -a    
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial
```

To reproduce w/ conda 4.4.0:

```
mkdir tmp
cd tmp
echo raise > io.py
source activate <my_env_name_here>
```

Result:

```
[nicholasc@nicholasc-ubuntu]~/$mkdir tmp
[nicholasc@nicholasc-ubuntu]~/$cd tmp
[nicholasc@nicholasc-ubuntu]~/tmp$echo raise > io.py
[nicholasc@nicholasc-ubuntu]~/tmp$source activate mtrain
  File "/home/nicholasc/anaconda2/bin/conda", line 7, in <module>
    from conda.cli import main
  File "/home/nicholasc/anaconda2/lib/python2.7/site-packages/conda/__init__.py", line 10, in <module>
    from .common.compat import iteritems, text_type
  File "/home/nicholasc/anaconda2/lib/python2.7/site-packages/conda/common/compat.py", line 108, in <module>
    from io import open as io_open  # NOQA
  File "/home/nicholasc/tmp/io.py", line 1, in <module>
    raise
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

Similarly, conda --version (called from inside ~/tmp, after the echo above) results in:

```
[nicholasc@nicholasc-ubuntu]~/tmp/$conda --version
  File "/home/nicholasc/anaconda2/bin/conda", line 7, in <module>
    from conda.cli import main
  File "/home/nicholasc/anaconda2/lib/python2.7/site-packages/conda/__init__.py", line 10, in <module>
    from .common.compat import iteritems, text_type
  File "/home/nicholasc/anaconda2/lib/python2.7/site-packages/conda/common/compat.py", line 108, in <module>
    from io import open as io_open  # NOQA
  File "/home/nicholasc/tmp/io.py", line 1, in <module>
    raise
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

Same behavior on on macOS (10.12.6), conda version 4.3.30

kalefranz:
fix #6549 

---

The solution proposed here is unconventional.  To be honest, I've never seen another python application do this (although I guess I haven't seen source code for a lot of python *applications*).  I've definitely never seen another python *library* do this.  And since conda is used both as an application and as a library, this PR is conservative in isolating the change only to the internal `shell.*` conda commands used for the shell wrappers.

### DIFF
--- a/conda/__init__.py
+++ b/conda/__init__.py
@@ -5,17 +5,12 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import sys
 
-# before any more imports, leave cwd out of sys.path for internal 'conda shell.*' commands
+# Before any more imports, leave cwd out of sys.path for internal 'conda shell.*' commands.
 # see https://github.com/conda/conda/issues/6549
-if len(sys.argv) > 1 and sys.argv[1].startswith('shell.'):
-    try:
-        # The standard first entry in sys.path is an empty string, and os.path.abspath('')
-        # expands to os.getcwd()
-        idx = sys.path.index('')
-    except ValueError:
-        pass
-    else:
-        del sys.path[idx]
+if len(sys.argv) > 1 and sys.argv[1].startswith('shell.') and sys.path and sys.path[0] == '':
+    # The standard first entry in sys.path is an empty string,
+    # and os.path.abspath('') expands to os.getcwd().
+    del sys.path[0]
 
 import os  # NOQA
 from os.path import dirname  # NOQA
--- a/conda/__init__.py
+++ b/conda/__init__.py
@@ -3,12 +3,25 @@
 """OS-agnostic, system-level binary package manager."""
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os
-from os.path import dirname
 import sys
 
-from ._vendor.auxlib.packaging import get_version
-from .common.compat import iteritems, text_type
+# before any more imports, leave cwd out of sys.path for internal 'conda shell.*' commands
+# see https://github.com/conda/conda/issues/6549
+if len(sys.argv) > 1 and sys.argv[1].startswith('shell.'):
+    try:
+        # The standard first entry in sys.path is an empty string, and os.path.abspath('')
+        # expands to os.getcwd()
+        idx = sys.path.index('')
+    except ValueError:
+        pass
+    else:
+        del sys.path[idx]
+
+import os  # NOQA
+from os.path import dirname  # NOQA
+
+from ._vendor.auxlib.packaging import get_version  # NOQA
+from .common.compat import text_type  # NOQA
 
 __all__ = (
     "__name__", "__version__", "__author__", "__email__", "__license__", "__summary__", "__url__",
@@ -59,6 +72,7 @@ class CondaError(Exception):
             raise
 
     def dump_map(self):
+        from .common.compat import iteritems
         result = dict((k, v) for k, v in iteritems(vars(self)) if not k.startswith('_'))
         result.update(exception_type=text_type(type(self)),
                       exception_name=self.__class__.__name__,
--- a/conda/core/repodata.py
+++ b/conda/core/repodata.py
@@ -17,12 +17,12 @@ from textwrap import dedent
 from time import time
 import warnings
 
-from .. import CondaError, iteritems
+from .. import CondaError
 from .._vendor.auxlib.ish import dals
 from .._vendor.auxlib.logz import stringify
 from ..base.constants import CONDA_HOMEPAGE_URL
 from ..base.context import context
-from ..common.compat import (ensure_binary, ensure_text_type, ensure_unicode, text_type,
+from ..common.compat import (ensure_binary, ensure_text_type, ensure_unicode, iteritems, text_type,
                              with_metaclass)
 from ..common.url import join_url, maybe_unquote
 from ..core.package_cache import PackageCache


mcg1969:
I'm not sure I should be alone in reviewing this one. Maybe pull Michael S and/or Ray in here?

kalefranz:
Reference for `sys.path`: https://stackoverflow.com/a/38403654

mingwandroid:
IMHO this is a massive security issue in Python, here's my PoC:

put this in your current directory as `getpass.py`
```
def getpass():
    import imp, sys
    f, pathname, desc = imp.find_module('getpass', sys.path[1:])
    module = imp.load_module('real_getpass', f, pathname, desc)
    f.close()
    result = module.getpass()
    print('You entered {}\n', result)
    return result
```

Now launch python and do:
```
import getpass
getpass.getpass()
```

I've captured your password by managing to get a file into your CWD.

mingwandroid:
Here's Perl's take on this: https://www.effectiveperlprogramming.com/2017/01/v5-26-removes-dot-from-inc/

Note, this broke quite a lot of things so I had to pass the configure option to disable this!

Anyway, Kale's fix LGTM as it's the best we can do for now. I'll probably file this PoC over a bugs.python.org, but does anyone know if this has been brought up before? Maybe on Python Ideas mailing list? I'd be surprised if it hasn't been discussed to death already.

kalefranz:
Yeah the perl case is even less risky, because perl had `.` last.  Python in general puts it first!

kalefranz:
I cancelled the Appveyor build here.  Have too much other stuff pending.  This can be merged now.

mingwandroid:
Upstream, 14 year old bug report about this: https://bugs.python.org/issue946373

> This is too problematic to change the behavior. Closing as "Won't fix", the consensus of the comments.

Despite this I believe adding an argument to the Python interpreter to prevent the addition of '.' to the front of `sys.path` is the best approach and would not hack it in conda. I'd appreciate you opinion here @pitrou

Less relevant (due to the embedding focus) bug report: https://bugs.python.org/issue5753

Less relevant (due to the `/tmp` focus which strikes me as a red-herring) bug report from 5 years ago: https://bugs.python.org/issue16202

conda/__init__.py
```diff
@@ -3,12 +3,20 @@
...
-import os
-from os.path import dirname
 import sys
 
-from ._vendor.auxlib.packaging import get_version
-from .common.compat import iteritems, text_type
+# Before any more imports, leave cwd out of sys.path for internal 'conda shell.*' commands.
+# see https://github.com/conda/conda/issues/6549
+if len(sys.argv) > 1 and sys.argv[1].startswith('shell.') and sys.path and sys.path[0] == '':
+    # The standard first entry in sys.path is an empty string,
```
nicain: For reference in my PR review

msarahan:
I think this PR is fine.  Another way this problem manifests itself is when sitting in a source folder, python will find appropriately named folders with ``__init__.py`` files in them, and happily use them instead of vendored dependencies.  This bit me really hard with virtualenv and its vendoring: https://github.com/pypa/virtualenv/pull/1094

In general, if you're executing code in a folder that has python files in it (not just data or whatever) you might have a bad time.  Everyone is advised to create packages, and perhaps install them in development mode as necessary.

mingwandroid:
Merging this mean that `sys.path` gets modified globally as soon as anything does `import conda`? That's a huge change in behaviour and will cause failures, could we at least not put it back at the end of the file?

mbargull:
@nicain:
> My suggestion, since conda is also a library, is to push the fix (the del of sys.path[0]) all the way down close to the entry points

That wouldn't help since `conda/__init__.py` and `conda/cli/__init__.py` will be loaded before `conda.cli.main:main` is executed. Hence, to guard as much imports as possible, the fix must be in `conda/__init__.py` and as early as possible.

@mingwandroid: `sys.path` will only be modified conditionally (esp. only if `sys.argv[1].startswith("shell.")`), so it's not as bad as it could be.

It think this is the right workaround if you'd want to do this inside the `conda` Python package. But another possibility would be to use custom entry points, i.e., replace
```python
import re
import sys

from conda.cli import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
with something like
```python
import sys

# Before any more imports, leave cwd out of sys.path.
# see https://github.com/conda/conda/issues/6549
if sys.path and sys.path[0] == '':
    # The standard first entry in sys.path is an empty string,
    # and os.path.abspath('') expands to os.getcwd().
    del sys.path[0]

import re

from conda.cli import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
. This would add the workaround as early as possible, I believe.

mingwandroid:
It would be modified conditionally basically all of the time though, that test was only added incase of very unexpected sys.path values.

For something someone imports changing sys like this is not something I'm comfortable with. Putting it in the entry points is better.

mbargull:
Oh yeah, I'm *very much* in favor of modifying the entry points. IMHO, such workarounds should be done by a launcher but if at all possible not inside an "arbitrary" library.

If we want to be "idealistic" -- or "evil", depending on the POV --, we could also say f*** that legacy behavior and let `conda-build` (via an opt-in or opt-out?) always create scripts with that workaround 😉.
😇 ¯\\\_(ツ)\_/¯ 😈
At least for `conda` itself I think it would be the best fix, though.

nicain:
Thanks @mbargull your suggestion is exactly what I was thinking.

> such workarounds should be done by a launcher but if at all possible not inside an "arbitrary" library.

+1

When I said:

> push the fix (the del of sys.path[0]) all the way down

What I meant was "all the way up" :)

Out of curiosity, is there precedence in other projects for this?  I was under the impression that setuptools builds the entry-point code (easy_install does this?), so how would you make that modification?

mbargull:
> is there precedence in other projects for this?

don't know.

> I was under the impression that setuptools builds the entry-point code

It does (if they are specified in `setup.py`), but `conda-build` overwrites those if they are also given in `meta.yaml`. In either case, with `setuptools` or `conda-build`, you can, of course, always create your own scripts instead of using the generated ones.

kalefranz:
`__init__.py` is the common entry point for the "generated" entry points that go in `bin`/`Scripts` as well as the `python -m conda` entry point provided by `__main__.py`.

The location of the change in this PR is as far up in the call stack as possible without creating duplicate code or forcing us to write custom entry points (which I also don't want do).  The structure of entry points have a huge WTF history with stupid stuff setuptools has done in the past, and I'd rather keep the WTFs to a minimum, even if justified.

mbargull:
I feel the changes from this PR are (by designed/necessity) incomplete as they only fix the `(source|conda) (activate|deactivate)` case but nothing more, e.g., with those changes,
```bash
echo raise >io.py
conda activate
conda install foo
```
will successfully activate but fail to install.
Regarding the entry point for `python -m conda`, I don't think this is much relevant for the user as they are unlikely to not use `conda` instead -- and if `python -m conda` is used in development/CI systems, it doesn't really matter as there you could/should easily avoid thinks like `touch io.py && conda ...`.
I understand your reasoning (esp. "setuptools/pkg_resources shenanigans" 😑...), but I still think the modified console script I proposed is the more appropriate and consistent fix.

kalefranz:
If we do bump the fix up to the entry point, let's do it as part of https://github.com/conda/conda/pull/6518.  Notice in the conda recipe there I've removed entry points declarations from `meta.yaml`, and e.g. in build.sh

https://github.com/conda/conda/pull/6518/files#diff-4d5085f0c17bf23334372effd03a4317

`python setup.py install` creates a "neutered" entry point where only `conda init` and `conda info` work (and this is what we'll use in sdists and wheels as well), and then `conda init --install` finishes out the entry points and shell wrapper installation in the prefix.

mbargull:
I haven't looked into #6518 yet, but I think it is very reasonable to first figure `conda init` out and then apply the fix for this issue (or incorporate it in #6518, as you suggested)!

kalefranz:
https://bugs.python.org/issue33053

travishathaway:
This pull request is no longer relevant; therefore, it is being closed.

PR #6570, opened 2017-12-24
resolve #6569 make conda more conservative writing to home directories

### LINKED ISSUE #6569: conda 4.4.3 unconditionally attempts to write to HOME, even when not necessary
Steps to reproduce:

```
root@922395dccc83:~# sudo -u jenkins /opt/conda/bin/conda install -q -y conda


Package plan for installation in environment /opt/conda:

The following packages will be UPDATED:

    conda:   4.3.31-py36_0        --> 4.4.3-py36_0        
    pycosat: 0.6.2-py36h1a0ea17_1 --> 0.6.3-py36h0a5515d_0

root@922395dccc83:~# sudo -u jenkins /opt/conda/bin/conda install -q -y numpy
Solving environment: ...working... failed

NotWritableError: The current user does not have write permissions to a required path.
  path: /root/.conda/pkgs/urls.txt
  uid: 1014
  gid: 1014

If you feel that permissions on this path are set incorrectly, you can manually
change them by executing

  $ sudo chown 1014:1014 /root/.conda/pkgs/urls.txt

```

In older versions of Conda, it would not attempt to write to HOME when installing packages.

The end user here can easily solve the problem by passing `-H` to `sudo` (thus causing HOME to be set correctly in this case).

kalefranz:
resolve #6569

The goal of this PR is to only *write* to these magic file locations when we know we're actually going to be writing to that "location" (either `pkgs` dir or `envs` dir).

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -32,6 +32,10 @@ log = getLogger(__name__)
 _FORMATTER = Formatter("%(levelname)s %(name)s:%(funcName)s(%(lineno)d): %(message)s")
 
 
+def dashlist(iterable, indent=2):
+    return ''.join('\n' + ' ' * indent + '- ' + str(x) for x in iterable)
+
+
 class ContextDecorator(object):
     """Base class for a context manager class (implementing __enter__() and __exit__()) that also
     makes it a decorator.
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1145,20 +1145,28 @@ def determine_target_prefix(ctx, args=None):
 
 
 def _first_writable_envs_dir():
-    from ..gateways.disk.update import touch
+    # Calling this function will *create* an envs directory if one does not already
+    # exist. Any caller should intend to *use* that directory for *writing*, not just reading.
     for envs_dir in context.envs_dirs:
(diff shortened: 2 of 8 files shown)


kalefranz:
Not ready yet.  I need more time with this.

kalefranz:
Ok I'm ready to call this one "Ready"

conda/common/io.py
```diff
@@ -32,6 +32,10 @@
 _FORMATTER = Formatter("%(levelname)s %(name)s:%(funcName)s(%(lineno)d): %(message)s")
 
 
+def dashlist(iterable, indent=2):
+    return ''.join('\n' + ' ' * indent + '- ' + str(x) for x in iterable)
```
kalefranz: The point in putting this here now is to start transitioning to using it from here.  We've always been importing it from `conda.resolve` before, which is so wrong.

kalefranz:
rebased to clear merge conflict

kalefranz:
This PR will help with, although not fully address, #7109.  The point of this PR is to not write files (and therefore attach permissions to them) when doing so can be problematic for multiuser installs.

PR #6585, opened 2017-12-27
fix #6577, #6580 single quote in PS1

### LINKED ISSUE #6577: Updated to conda 4.4.4, can't activate any environment
I just updated to 4.4.4, and now activation fails on a bash syntax error at the PS1 configuration, even for a freshly created env:
```bash
conda create -n foo -y -q  python
````

then:

```bash
alpamayo[junk]> conda activate foo
bash: syntax error near unexpected token `('
alpamayo[junk]> source activate foo
bash: eval: line 58: syntax error near unexpected token `('
bash: eval: line 58: `PS1='(foo) \[\033[0;32m\]$(__git_ps1 '(%s)')\[\033[1;31m\]\[\033[1;34m\]\h[\[\033[1;36m\]\W\[\033[1;34m\]]\[\033[1;32m\]> \[\033[0m\]''
````

Some conda info:

```bash
alpamayo[junk]> conda --version
conda 4.4.4
alpamayo[junk]> conda info -a

     active environment : None
            shell level : 0
       user config file : /Users/fperez/.condarc
 populated config files : /Users/fperez/.condarc
          conda version : 4.4.4
       base environment : /Users/fperez/usr/conda  (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/osx-64
                          https://repo.continuum.io/pkgs/free/osx-64
                          https://repo.continuum.io/pkgs/r/osx-64
                          https://repo.continuum.io/pkgs/pro/osx-64
          package cache : /Users/fperez/usr/conda/pkgs
                          /Users/fperez/.conda/pkgs
       envs directories : /Users/fperez/usr/conda/envs
                          /Users/fperez/.conda/envs
               platform : osx-64
             user-agent : conda/4.4.4 requests/2.18.4 CPython/3.6.4 Darwin/16.7.0 OSX/10.12.6
                UID:GID : 33212:20665
             netrc file : None
           offline mode : False
```

Any tips? thanks!

### LINKED ISSUE #6580: activate script breaks prompt on 4.4.0-4.4.4
I have the following bash prompt:
```
$ echo $PS1
\u@\h:\w$(chg prompt '[\[\e[0;32m\]{\[\e[0;31m\]{status|modified}}{branch}\[\e[0m\]{{update}}] ' 2>/dev/null)$(__git_ps1)$
```
Up to version 4.3.27, conda works fine with that. Starting from 4.4.0, it breaks the prompt when I source the `activate` script:
```
$ echo $PS1
(dask36) \u@\h:\w$(chg prompt [[e[0
```

kalefranz:
fix #6577
fix #6580

Both issues are caused by having single-quotes in the PS1 string, and also trying to re-set that string using single-qutoes, as `PS1='<CONTENTS>'`, where `<CONTENTS>` itself then contains single-quotes.

It appears the best solution is https://stackoverflow.com/a/1250279.  It seems to work, based on my testing.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -426,6 +426,10 @@ class Activator(object):
             current_prompt_modifier = os.environ.get('CONDA_PROMPT_MODIFIER')
             if current_prompt_modifier:
                 ps1 = re.sub(re.escape(current_prompt_modifier), r'', ps1)
+            # Because we're using single-quotes to set shell variables, we need to handle the
+            # proper escaping of single quotes that are already part of the string.
+            # Best solution appears to be https://stackoverflow.com/a/1250279
+            ps1 = ps1.replace("'", "'\"'\"'")
             set_vars.update({
                 'PS1': conda_prompt_modifier + ps1,
             })


mbargull:
Looks like the most appropriate fix to me. Would it make sense to use a more general solution and turn the `self.*_tmpl` attributes into functions, i.e., instead of
```python
    self.set_var_tmpl = "%s='%s'"
...
    ps1 = ps1.replace("'", "'\"'\"'")
...
    yield self.set_var_tmpl % (key, value)
```
use
```python
    def format_set_var(key, value):
        return  "%s='%s'" % (key, value.replace("'", "'\"'\"'"))
    self.format_set_var = format_set_var
...
    yield self.format_set_var(key, value)
```
? To keep the code succinct for cases that don't need special escaping, you could then use something like
```python
def format(self, tmpl, *args):
    return tmpl % args
...
    self.format_unset_var = partial(self.format, "unset %s")
...
    yield self.format_unset_var(key, value)
```

kalefranz:
@mbargull Might make sense to go a step further even, and make `Activator` an abstract base class, with implementations for each shell.  Rather than that crazy branching `if/else` statement in the `__init__`.  I'll create a new issue and make it a TODO item.

kalefranz:
I'd rather not do all of that in this PR.  We need to get a fix out, and that change is going to throw around a lot more code, which means more development time and more code review time.

mbargull:
> make `Activator` an abstract base class

That's also what I had in mind but I didn't want to be even more intrusive :wink:.

> We need to get a fix out

Agreed, and can confirm your fix works as intended!

PR #6587, opened 2017-12-27
fix #6581 git-bash activation

### LINKED ISSUE #6581: Conda 4.4.4 breaks activation from Git Bash on Windows
Running 

    source ~/Anaconda3/Scripts/activate

gives

    bash: /c/ci/conda_1514155543394/_h_env/etc/profile.d/conda.sh: No such file or directory

kalefranz:
fix #6581

### DIFF
--- a/conda.recipe/bld.bat
+++ b/conda.recipe/bld.bat
@@ -1,6 +1,7 @@
 :: Delegate to the Unixy script. We need to translate the key path variables
 :: to be Unix-y rather than Windows-y, though.
 set "saved_recipe_dir=%RECIPE_DIR%"
+set "WIN_PREFIX=%PREFIX%"
 FOR /F "delims=" %%i IN ('cygpath.exe -u -p "%PATH%"') DO set "PATH_OVERRIDE=%%i"
 FOR /F "delims=" %%i IN ('cygpath.exe -u "%LIBRARY_PREFIX%"') DO set "LIBRARY_PREFIX=%%i"
 FOR /F "delims=" %%i IN ('cygpath.exe -u "%PREFIX%"') DO set "PREFIX=%%i"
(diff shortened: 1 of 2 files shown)


petebachant:
Think it would make sense to try to add a test to AppVeyor to catch this in the future? If so, let me know where it might go and I can try to work on a PR.

kalefranz:
I think it would be great.  I'm working on #6518 right now, which when done will get rid of a lot of the bash code in this `functions.sh` file that I just changed.

Honestly, I don't know how to write this test just yet, because to do it correctly and completely, we need to use conda-build to build conda, install that package, and then run the test from there.  Which I guess points to maybe putting something in the `test` block of the recipe.
https://github.com/conda/conda/blob/master/conda.recipe/meta.yaml#L46-L59

mingwandroid:
You could remove a heck of a lot of duplication here by making `cygpath` a shell function on non-Windows that returns what it is passed!

kalefranz:
If I was an average unix user, and saw that somewhere, I might feel like I was being violated by Windows shenanigans haha

kalefranz:
#6518 Should end up cleaning a lot of this up.  And probably allow us to get rid of most of that `functions.sh` file altogether.

kalefranz:
@mingwandroid You asked in another PR how to set up a dev environment.  With #6518, it'll probably be something like

1. install a miniconda somewhere
2. clone the conda repo and cd into it
3.     eval `miniconda/bin/python -m conda initialize --dev`

nehaljwani:
@kalefranz We should make sure that it goes into the README :)
I remember @msarahan asking about it too!

kalefranz:
Yup! We need to have better developer docs, once #6518 is ready.

PR #6594, opened 2017-12-28
Rename instances of 'root' to 'base'

petebachant:


conda/cli/conda_argparse.py
```diff
@@ -267,9 +267,9 @@ def configure_parser_info(sub_parsers):
         help="Display information about packages.",
     )
     p.add_argument(
-        '--root',
+        '--base',
         action='store_true',
-        help='Display root environment path.',
+        help='Display base environment path.',
     )
     p.add_argument(
```
kalefranz: We need to keep root here for backward compatibility.  So, keep `--base` here as you've modified it.  But then also add

```python
p.add_argument(
    '--root',
    action='store_true',
    help=SUPRESS,
    dest='base',
)
```

kalefranz:
~line 284 if `conda/cli/main_info.py` needs to be changed from `args.root` to `args.base`

kalefranz:
Oh, and the target of the PR should be the `4.4.x` branch instead of master.  You should be able to click `Edit` button by the title of the PR to switch it.

kalefranz:
Ok I guess retargeting didn't work.  Probably need to open a new PR to fix this.  I can do it if you want, but you're obviously fully capable... Fixing it looks something like

* git checkout 4.4.x
* git checkout -b root2base-2
* git cherry-pick 84567ac

kalefranz:
I cancelled the circle ci jobs for your last push.  Hope you don't mind.  Thought you were working on this now and would be closing the PR anyway.  I can restart them if you want.

PR #6608, opened 2017-12-29
fix #6599 space in path to base prefix

### LINKED ISSUE #6599: Anaconda Prompt Error
_From @armstrang on December 24, 2017 6:29_

untitlgged

i can't use The anaconda Prompt

_Copied from original issue: ContinuumIO/anaconda-issues#7904_

kalefranz:
fix #6599

### DIFF
--- a/conda.recipe/bld.bat
+++ b/conda.recipe/bld.bat
@@ -1,6 +1,7 @@
 :: Delegate to the Unixy script. We need to translate the key path variables
 :: to be Unix-y rather than Windows-y, though.
 set "saved_recipe_dir=%RECIPE_DIR%"
+set "WIN_PREFIX=%PREFIX%"
 FOR /F "delims=" %%i IN ('cygpath.exe -u -p "%PATH%"') DO set "PATH_OVERRIDE=%%i"
 FOR /F "delims=" %%i IN ('cygpath.exe -u "%LIBRARY_PREFIX%"') DO set "LIBRARY_PREFIX=%%i"
 FOR /F "delims=" %%i IN ('cygpath.exe -u "%PREFIX%"') DO set "PREFIX=%%i"
--- a/conda/shell/Library/bin/conda.bat
+++ b/conda/shell/Library/bin/conda.bat
@@ -1,12 +1,8 @@
-@REM @SET _CONDA_EXE="%~dp0..\..\Scripts\conda.exe"
-
-@IF NOT "%_CONDA_EXE%" == "" GOTO skip_conda_exe_dev
-    @IF "%CONDA_PYTHON_EXE%" == "" (
-        @SET "_CONDA_EXE=python "%~dp0..\..\bin\conda""
-    ) ELSE (
-        @SET "_CONDA_EXE="%CONDA_PYTHON_EXE%" "%~dp0..\..\bin\conda""
-    )
-:skip_conda_exe_dev
+@IF EXIST "%~dp0..\..\Scripts\conda.exe" (
+    @SET "_CONDA_EXE="%~dp0..\..\Scripts\conda.exe""
+) ELSE (
+    @SET "_CONDA_EXE=python "%~dp0..\..\bin\conda""
+)
 
 @FOR /F "delims=" %%i IN ('@python -c "import ctypes; print(ctypes.cdll.kernel32.GetACP())"') DO @SET "PYTHONIOENCODING=%%i"
 @chcp %PYTHONIOENCODING% > NUL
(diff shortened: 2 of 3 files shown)


mbargull:
Carrying over from #6599:

----

The `%_CONDA_EXE%` doesn't need to be quoted as I said before since that is already done in
https://github.com/conda/conda/blob/4.4.4/conda/shell/Library/bin/conda.bat#L5-L7.
I currently wonder how `CMD` interprets
```bat
"_CONDA_EXE="%CONDA_PYTHON_EXE%" "%~dp0..\..\bin\conda""
```
guess I'll just try it out in a VM in a minute..

EDIT: seems to work as intended, so no change in `conda.bat` needed, I guess.

----

Maybe the problem is in https://github.com/conda/conda/blob/4.4.4/utils/functions.sh#L195:
```bash
echo "@SET \"_CONDA_EXE=$win_conda_exe\"" > "$prefix/Library/bin/conda.bat"
```
would
```bash
echo "@SET \"_CONDA_EXE=\"$win_conda_exe\"\"" > "$prefix/Library/bin/conda.bat"
```
work? I'm not sure whether `$win_conda_exe` from
```bash
local win_conda_exe="$(cygpath --windows "$conda_exe")"
```
include quotes..

----

kalefranz:
From https://github.com/conda/conda/issues/6599#issuecomment-354458263 and https://github.com/conda/conda/issues/6599#issuecomment-354458454

---

The `$win_conda_exe` stuff has changed in https://github.com/conda/conda/pull/6587.  Go ahead and look through that, looking for the same thing...

---

Actually, this one I've been through before.  I *know* we've settled on

    SET "ENV_VAR=some value with spaces"

as the correct way to quote for `cmd.exe` syntax.

kalefranz:
If I'm wrong on that last comment, then we also need to modify https://github.com/conda/conda/blob/4.4.4/conda/activate.py#L93

mbargull:
Yes, I know we agreed to
```bat
SET "ENV_VAR=some value with spaces"
```
and use it then with
```bat
CALL "%ENV_VAR%"
```
and I think this is usually the right thing to do. Hence `activate.py` should be fine, I guess.
With `_CONDA_EXE` there is a special handling though (which I forgot in my initial comments to #6599) to include quotes in the variable itself as it might contain separate arguments, e.g., from
```bat
@SET "_CONDA_EXE="%CONDA_PYTHON_EXE%" "%~dp0..\..\bin\conda""
```

mbargull:
But maybe we can just remove the special handling in https://github.com/conda/conda/blob/4.4.4/conda/shell/Library/bin/conda.bat#L3-L9 and always assume we have `conda.exe`? For which occasions is that code even used? From the label `:skip_conda_exe_dev` I'd guess only for some development mode?

mbargull:
If we can remove that special handling, we can treat `_CONDA_EXE` the "normal" way, i.e., just like with the changes from this PR as `"%_CONDA_EXE%"`.

kalefranz:
Here's a reference: https://ss64.com/nt/set.html

> ## Values with Spaces - using Double Quotes
> There is no need to add quotation marks when assigning a value that includes spaces
> 
>     SET _variable=one two three
> 
> For special characters like & you can surround the entire expression with quotation marks.
> The variable contents will not include the surrounding quotes:
> 
>     SET "_variable=one & two"
> 
> If you place quotation marks around the value, then those quotes will be stored:
> 
>     SET _variable="one & two"

kalefranz:
> For which occasions is that code even used?

Yeah that block only gets triggered in a dev mode where we're using relative paths in the shell wrappers.  Basically so that you can clone the repo, then run `py.test tests/test_activate.py` without having to hard-code any paths or change the contents of files.  When we do a build, that code in `functions.sh` hard-codes the needed paths, and then conda's prefix rewriting (ideally) handles things correctly when installing the package into an environment.  I say ideally, because #6587.

kalefranz:
Ok, just force-pushed to switch things up here.  The first two commits right now are from #6587, so we don't create merge issues.  The third commit is the new one.

Because cmd.exe doesn't have the concept of `source`, and because we're using PATH tricks to put `conda.bat` in front of `conda.exe`, relative paths in `conda.bat` will probably work ok, and we don't even have to worry about writing absolute paths.  Trying that out in the third commit on this PR.

kalefranz:
I think this is safe to include in 4.4.5.  After I build 4.4.5 and post it to canary, I'll do further testing on windows to make sure we've fixed all the issues there -- before we release 4.4.5 to `defaults`.

PR #6609, opened 2017-12-29
fix #6607 use subprocess rather than execv for conda command extensions

### LINKED ISSUE #6607: conda env giving strange non-blocking behavior on Windows
Using conda 4.4.4, this is happening in both the Anaconda prompt and Git Bash. To reproduce, simply run `conda env -h`. The prompt will return and it seems like conda is run in the background, producing output a few seconds later. Strangely enough, this doesn't happen when running `conda-env -h`.

kalefranz:
fix #6607


kalefranz:
This PR precisely reverts back to code used in 4.3.x.

nehaljwani:
I found more info on this at https://bugs.python.org/issue19124

PR #6614, opened 2017-12-30
recipe test improvements and associated bug fixes

kalefranz:
There are three single-line patches (bug fixes) in this PR that could potentially be their own PRs.  They've been called out in comments added to this PR.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,8 +1,8 @@
-{% set setup_py_data = load_setup_py_data() %}
+{% set on_win = SUBDIR in ('win-64', 'win-32') %}
 
 package:
   name: conda
-  version: "{{ setup_py_data.version }}"
+  version: "{{ GIT_DESCRIBE_TAG }}"
 
 source:
   path: ../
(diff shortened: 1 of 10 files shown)


conda.recipe/meta.yaml
```diff
@@ -44,19 +44,45 @@ requirements:
...
+    - FOR /F "delims=" %%i IN ('python -c "import sys; print(sys.version_info[0])"') DO set "PYTHON_MAJOR_VERSION=%%i"
+    - set TEST_PLATFORM=win
+    - FOR /F "delims=" %%i IN ('python -c "import random as r; print(r.randint(0,4294967296))"') DO set "PYTHONHASHSEED=%%i"
+    - set
+    - conda info
+    - py.test tests -m "not integration and not installed" -vv
+    - conda create -y -p .\built-conda-test-env
+    - conda activate .\built-conda-test-env
+    - echo %CONDA_PREFIX%
+    - IF NOT "%CONDA_PREFIX%"=="%CD%\built-conda-test-env" EXIT /B 1
```
kalefranz: For whatever reason, conda-build doesn't seem to want to run the above on Windows.  It doesn't have a problem running the below on macOS though.

conda/cli/main_info.py
```diff
@@ -226,7 +226,7 @@ def get_env_vars_str(info_dict):
     env_vars = info_dict.get('env_vars', {})
     for key in sorted(env_vars):
         value = wrap(env_vars[key])
-        first_line = value[0]
+        first_line = value[0] if len(value) else ""
```
kalefranz: This actually fixes a bug.  Maybe it should be a separate PR?

conda/cli/conda_argparse.py
```diff
@@ -134,6 +134,7 @@ def error(self, message):
                             from ..exceptions import CommandNotFoundError
                             raise CommandNotFoundError(cmd)
                         args = [find_executable('conda-' + cmd)]
+                        args.extend(sys.argv[2:])
```
kalefranz: This fixes a bug that's a show-stopper for 4.4.5.  Meaning 4.4.5 will never be released to defaults.

utils/functions.sh
```diff
@@ -162,7 +162,7 @@ install_conda_shell_scripts() {
     local src_dir=${2:-${SRC_DIR:-$PWD}}
 
     if [ -n "$ON_WIN" ]; then
-        local conda_exe="$WIN_PREFIX\$BIN_DIR\conda$EXE_EXT"
+        local conda_exe="$WIN_PREFIX\\$BIN_DIR\\conda$EXE_EXT"
```
kalefranz: Also fixes a bug in the 4.4.5 release.

conda/cli/conda_argparse.py
kalefranz: Just an error/omission in copying over the code from 4.3.x.

PR #6651, opened 2018-01-04
resolve #6650 add upgrade message for unicode errors in python 2

### LINKED ISSUE #6650: Suppress full stack trace for UnicodeEncodeError and UnicodeDecodeError in python 2

kalefranz:
resolve #6650

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -663,7 +663,7 @@ class EncodingError(CondaError):
         A unicode encoding or decoding error has occurred.
         Python 2 is the interpreter under which conda is running in your base environment.
         Replacing your base environment with one having Python 3 may help resolve this issue.
-        If you still have a need for python 2 environments, consider using the 'conda create'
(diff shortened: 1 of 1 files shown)


conda/exceptions.py
```diff
@@ -656,6 +656,20 @@ def __init__(self, invalid_spec):
...
 
 
+class EncodingError(CondaError):
+
+    def __init__(self, caused_by, **kwargs):
+        message = dals("""
+        A unicode encoding or decoding error has occurred.
+        Python 2 is the interpreter under which conda is running in your base environment.
+        Replacing your base environment with one built on Python 3 may help resolve this issue.
+
```
mingwandroid: Maybe add something here along the lines of "you may be able to create a new conda environment containing Python 2 (depending on the other software you need to run)"?

kalefranz: Yeah I like that.  I'll add it.  Or something about the best practices way of interacting with python2 is with non-base environments.

PR #6653, opened 2018-01-04
fix #6624 can't start new thread

### LINKED ISSUE #6624: RuntimeError: can't start new thread
```
error: RuntimeError("generator didn't stop after throw()",)
command: /home/mholekev/miniconda3/bin/conda install --yes numpy scipy matplotlib jupyter git pandas
user_agent: conda/4.4.4 requests/2.18.4 CPython/3.6.3 Linux/2.6.32-696.10.3.el6.x86_64 centos/6.7 glibc/2.12
_messageid: -9223372036842075552
_messagetime: 1514766990662 / 2017-12-31 18:36:30

  File "/home/mholekev/miniconda3/lib/python3.6/site-packages/conda/common/io.py", line 399, in backdown_thread_pool
    yield ThreadPoolExecutor(max_workers)
  File "/home/mholekev/miniconda3/lib/python3.6/site-packages/conda/core/index.py", line 214, in get_reduced_index
    new_records = query_all(spec)
  File "/home/mholekev/miniconda3/lib/python3.6/site-packages/conda/core/index.py", line 187, in query_all
    return tuple(concat(future.result() for future in as_completed(futures)))
  File "/home/mholekev/miniconda3/lib/python3.6/site-packages/conda/core/index.py", line 187, in <genexpr>
    return tuple(concat(future.result() for future in as_completed(futures)))
  File "/home/mholekev/miniconda3/lib/python3.6/concurrent/futures/_base.py", line 217, in as_completed
    fs = set(fs)
  File "/home/mholekev/miniconda3/lib/python3.6/site-packages/conda/core/index.py", line 186, in <genexpr>
    futures = (executor.submit(sd.query, spec) for sd in subdir_datas)
  File "/home/mholekev/miniconda3/lib/python3.6/concurrent/futures/thread.py", line 123, in submit
    self._adjust_thread_count()
  File "/home/mholekev/miniconda3/lib/python3.6/concurrent/futures/thread.py", line 142, in _adjust_thread_count
    t.start()
  File "/home/mholekev/miniconda3/lib/python3.6/threading.py", line 846, in start
    _start_new_thread(self._bootstrap, ())
RuntimeError: can't start new thread

```

kalefranz:
fix #6624 

This PR removes the `backdown_thread_pool` context manager, and replaces it with `ThreadLimitedThreadPoolExecutor`.  The original purpose of `backdown_thread_pool` was to prevent `RuntimeError`s as shown in #6624, but the strategy failed to do so.  After looking more at `concurrent.futures.ThreadPoolExecutor`, it seems the exception will only be thrown when a new thread is created, and that only happens after submitting the first task.  So `backdown_thread_pool` was wholly ineffective at its intended purpose.

This PR has some ugly parts, things I complain about in comments.  Specifically import and usage of "protected" attributes (names starting with `_`) from `concurrent.futures`.  That seems unavoidable though.  The only solution there would be to vendor our own `concurrent.futures`.  I'm indifferent on whether we want to do that or not.

I'd like to have two reviewers on this PR.  Probably @mingwandroid since he's already taken a first pass at it, and also @goanpeca.

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -395,16 +395,43 @@ def backdown_thread_pool(max_workers=10):
     """Tries to create an executor with max_workers, but will back down ultimately to a single
     thread of the OS decides you can't have more than one.
     """
+    from concurrent.futures import _base  # These "_" imports are gross, but I don't think there's an alternative  # NOQA
+    from concurrent.futures.thread import _WorkItem
+
+    class CondaThreadPoolExecutor(ThreadPoolExecutor):
+
+        def submit(self, fn, *args, **kwargs):
+            with self._shutdown_lock:
+                if self._shutdown:
+                    raise RuntimeError('cannot schedule new futures after shutdown')
+
+                f = _base.Future()
+                w = _WorkItem(f, fn, args, kwargs)
+
+                self._work_queue.put(w)
+                try:
+                    self._adjust_thread_count()
+                except RuntimeError:
+                    # RuntimeError: can't start new thread
+                    # See https://github.com/conda/conda/issues/6624
+                    if len(self._threads) > 0:
+                        # It's ok to not be able to start new threads if we already have at least
+                        # one thread alive.
+                        pass
+                    else:
+                        raise
+                return f
+
     try:
-        yield ThreadPoolExecutor(max_workers)
+        yield CondaThreadPoolExecutor(max_workers)
     except RuntimeError as e:  # pragma: no cover
         # RuntimeError is thrown if number of threads are limited by OS
         log.debug(repr(e))
         try:
-            yield ThreadPoolExecutor(floor(max_workers / 2))
+            yield CondaThreadPoolExecutor(floor(max_workers / 2))
         except RuntimeError as e:
             log.debug(repr(e))
-            yield ThreadPoolExecutor(1)
+            yield CondaThreadPoolExecutor(1)
 
 
 class ContextDecorator(object):
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -2,7 +2,8 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from collections import defaultdict
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, _base, as_completed
+from concurrent.futures.thread import _WorkItem
 from contextlib import contextmanager
 from enum import Enum
 from errno import EPIPE, ESHUTDOWN
@@ -11,7 +12,6 @@ from itertools import cycle
 import json
 import logging
 from logging import CRITICAL, Formatter, NOTSET, StreamHandler, WARN, getLogger
-from math import floor
 import os
 from os.path import dirname, isdir, isfile, join
 import signal
@@ -34,7 +34,6 @@ _FORMATTER = Formatter("%(levelname)s %(name)s:%(funcName)s(%(lineno)d): %(messa
 
 class CaptureTarget(Enum):
     """Constants used for contextmanager captured.
-
     Used similarily like the constants PIPE, STDOUT for stdlib's subprocess.Popen.
     """
     STRING = -1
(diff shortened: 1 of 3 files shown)


conda/common/io.py
```diff
@@ -395,16 +395,43 @@ def backdown_thread_pool(max_workers=10):
     """Tries to create an executor with max_workers, but will back down ultimately to a single
     thread of the OS decides you can't have more than one.
     """
+    from concurrent.futures import _base  # These "_" imports are gross, but I don't think there's an alternative  # NOQA
+    from concurrent.futures.thread import _WorkItem
```
kalefranz: Need to make sure these imports are ok for python 3.4 and python 3.5.  They obviously are for python 2.7 and 3.6.

kalefranz: Confirmed ok for both py34 and py35.

conda/common/io.py
```diff
@@ -395,16 +395,43 @@ def backdown_thread_pool(max_workers=10):
...
+
+    class CondaThreadPoolExecutor(ThreadPoolExecutor):
+
+        def submit(self, fn, *args, **kwargs):
```
kalefranz: This is an exact reimplementation of the `submit()` method on the parent class, except I've added `try/except` around `self._adjust_thread_count()`.  Seems to me like this is the technically sound and correct solution.  I don't have any independent references though, and I'm frustrated it was so dirty to implement.

mingwandroid: Maybe if you included a link to the exact lines you copied on github it'd make it easy to see if it's changed indicating you need to update your modified version?

kalefranz:
BTW, at the moment tests are running, but they've all passed previously.  I force-pushed only to correct some cosmetic things in docstrings.

conda/common/io.py
```diff
@@ -390,21 +389,48 @@ def close(self):
...
+                if len(self._threads) > 0:
+                    # It's ok to not be able to start new threads if we already have at least
+                    # one thread alive.
+                    pass
+                else:
+                    raise
+            return f
+
+
+as_completed = as_completed  # anchored here for import by other modules
```
goanpeca: Why do we need this here, if it is imported anyway on the top?

conda/common/io.py
```diff
@@ -390,21 +389,48 @@ def close(self):
...
+            w = _WorkItem(f, fn, args, kwargs)
+
+            self._work_queue.put(w)
+            try:
+                self._adjust_thread_count()
+            except RuntimeError:
+                # RuntimeError: can't start new thread
+                # See https://github.com/conda/conda/issues/6624
+                if len(self._threads) > 0:
+                    # It's ok to not be able to start new threads if we already have at least
```
goanpeca: Ok so this looks like it will prevent a lot of issues, but maybe we can be honest that it will still occur (less likely though). 

Maybe we need to add some python minor/patch version check and add some warning if it is not using the version for which this is tested. Or we could vendor (as@kalefranz  suggest) it if it is small enough. I am just worried we inadvertedly fail due to api changes? 

Maybe we should vendor this as part of this PR already and avoid future issues?

conda/common/io.py
kalefranz: It's for flake8 so that we don't have an unused import.

conda/common/io.py
kalefranz: > Maybe we should vendor this as part of this PR already and avoid future issues?

Much prefer that to some type of version checking.  I'll vendor it in this PR if @mingwandroid agrees.

kalefranz: > Ok so this looks like it will prevent a lot of issues, but maybe we can be honest that it will still occur (less likely though).

Yeah, if *no* threads can be created at all, then we have further work to do.  I don't want to handle that case though until we actually see evidence that it exists.

kalefranz: I've looked at code for concurrent.futures in all the various locations.  I'd rather not vendor right now.  We'll of course test conda with python 3.7 before it's released, and other future versions.  Actually, what I *will* do though is add a `run_constrained` requirement for python to conda's recipe.

kalefranz:
After this PR is approved and merged, I think there's some refactoring in order for the `conda/common/io.py` module.  It's starting to be a collection of a lot of unrelated stuff.

conda/common/io.py
goanpeca: ahhh right... you can just do 

```
as_completed
```

for that, no ?

conda/common/io.py
goanpeca: Ok, makes sense

conda/common/io.py
kalefranz: No, because then PyCharm complains.

image

goanpeca: Ok ;-)

PR #6669, opened 2018-01-05
fix #6667 aliased posix commands

### LINKED ISSUE #6667: Source activation for anaconda python environments no longer working (Mac)
The source activate command in conda is suddenly no longer working for me. When I type `source activate py3` in my Mac terminal, I get this error `-bash: _conda_activate: command not found`.

I recently created a new conda environment for TabPy (Tableau python server); I'm not sure if that changed anything. I have checked that the conda environments I'm trying to activate exist by using `conda info --envs`, here is that output:

```
# conda environments:
#
base                  *  /Users/applemacbook/anaconda
Tableau-Python-Server     /Users/applemacbook/anaconda/envs/Tableau-Python 
Server
py3                      /Users/applemacbook/anaconda/envs/py3
```

I already have the following in my bash profile from my initial installation of anaconda awhile ago `export PATH="/Users/applemacbook/anaconda/bin:$PATH"`.

Here is some additional information: The output of `which conda` is `/Users/applemacbook/anaconda/bin/conda`. The output of `type source` is `source is a shell builtin`. The output of `which activate` is `/Users/applemacbook/anaconda/bin/activate`. My conda version is conda 4.4.6

kalefranz:
fix #6667

This PR fixes #6667 caused by aliased posix commands.  It also dumps the complex and buggy logic to support the error message about `activate` needing to be sourced rather than executed.  With the `conda activate` command out in the wild and functioning well, there isn't as much need for that message any longer.


mbargull:
I just tried out some more stupid `alias`es:
Turns out, the user could also alias `if`, `then`, `fi`, `case`, `[`, `local`, `return` 😆

conda/shell/bin/activate
```diff
@@ -1,65 +1,2 @@
...
-if ! _conda_script_is_sourced; then
-    (
-      >&2 echo "Error: activate must be sourced. Run 'source activate envname'"
-      >&2 echo "instead of 'activate envname'."
-    )
-    exit 1
-fi
-
-_conda_set_vars
-. "$_CONDA_ROOT/etc/profile.d/conda.sh" || return $?
```
mbargull: I really like this change due to its simplifications! Since `_CONDA_ROOT` is added on installation, this is really as minimal as it gets 😄.
I'd like to suggest to remove the `chmod +x` commands on `activate`/`deactivate` in `utils/functions.sh`. That way the user should get at least a `command not found` message when they try to run them without `source` and know they've done something wrong.

kalefranz: I guess `source` uses the `PATH` env var to find its scripts?

mbargull: I believe so.

kalefranz:
> Turns out, the user could also alias `if`, `then`, `fi`, `case`, `[`, `local`, `return`

Probably `shift` too?  Who knows what else...

I think I'll cover `local`, `return`, and `case` at least.  Let's see how far that gets us.  If we have more problems, we'll just do the whole thing...

mbargull:
Ah yes, `shift` too!

kalefranz:
... and `else`, and `in`, and ... omg

kalefranz:
And *this* is why we keep the actual shell code and shenanigans to a minimum!

mbargull:
Damn, I also tried `shift`, `else`, `in` before but just didn't include those in my list somehow.
Though, aliasing `in` should be fine since it only gets used in a `case` command but isn't one of its own.

> And this is why we keep the actual shell code and shenanigans to a minimum!

Yep, really glad you were able to move so much to the Python side!

mbargull:
Wait, what?? You can `alias case=...` but it's not possible to use `\case`??!! Oh my....

kalefranz:
Yeah it looks that way

kalefranz:
@mbargull Does this PR look good to you now?

utils/functions.sh
```diff
@@ -182,13 +182,15 @@ install_conda_shell_scripts() {
         echo "#!/bin/sh" > "$prefix/$BIN_DIR/activate"
         echo "_CONDA_ROOT=\"\$(cygpath '$WIN_PREFIX')\"" >> "$prefix/$BIN_DIR/activate"
         cat "$src_dir/conda/shell/bin/activate" >> "$prefix/$BIN_DIR/activate"
-        chmod +x "$prefix/$BIN_DIR/activate"
+        # chmod +x "$prefix/$BIN_DIR/activate"
+        # we no longer set activate as executable, to ensure that it is being sourced
```
mbargull: The same changes should also be applied to the non-Windows case below

kalefranz: Done!

kalefranz:
rebased...

mingwandroid:
Let's see how this goes. I like the cleanup.

PR #6710, opened 2018-01-10
unexpected token in conda.bat

### LINKED ISSUE #6704: Installation with spaces// Anaconda prompt stop responding after updating conda.
_From @Usama23 on January 10, 2018 12:29_

OS: Windows
Anaconda2/3- 5.0.1
Anaconda Navigator 1.6.12

**Steps to Reproduce:**

1: Download and install anaconda with single space (single User/ All User).
2: From anaconda prompt update conda to latest version (4.4.7) .
3: Open another anaconda prompt.

**Actual Result:**

A weird statement appear on the start of prompt and then nothing works on that newly open prompt.

``` 
> was unexpected at this time.
C:\Users\tester>@IF NOT ""=="" @chcp  > NUL
```
screenshot_767
screenshot_768
screenshot_769

_Copied from original issue: ContinuumIO/anaconda#1175_

kalefranz:
fix #6704

@mbargull Could I get your eyes on this one if you have a moment?  Don't want to mess this one up again if I can help it.


mbargull:
The existing code looks fine and but the changes are wrong..
No idea what caused #6704, sorry.

kalefranz:
What the heck I know I tested this.  And now testing again, I'm getting a different result.  So based on this, yeah, I think the original is correct and something else is happening.

```
(base) c:\>echo %CONDA_SHLVL%
1

(base) c:\>echo "%CONDA_SHLVL%"
"1"

(base) c:\>IF %CONDA_SHLVL%=="1" echo "yes!"

(base) c:\>IF "%CONDA_SHLVL%"=="1" echo "yes!"
"yes!"

```

kalefranz:
I think there's a solution [here](https://ss64.com/nt/if.html) with `IF DEFINED`

mbargull:
When I (am forced to) write `.bat` files from scratch, I usually remember `defined` -- but so many times I forget that one...
So, I think it is good to use `defined` here, but I'm inclined to believe something else in #6704 might not be right as I still don't see why the original code failed..

Issue #6096, opened 2017-10-07
ValidationError: invalid value for timestamp


jennirinker:
MODERATOR EDIT:  This issue relates to an error that looks something like

    ValidationError: Invalid value 1505856869.685 for timestamp

The issue relates to using conda 4.3.28, and then downgrading (presumably via conda-forge) to a previous version of conda, like 4.3.27.

# SOLUTION #

### To fix on macOS:

    # in the command below, set PATH_TO_ENVIRONMENT yourself
    sed -i '' -E 's|("timestamp": [0-9]+)\.|\1|' /PATH_TO_ENVIRONMENT/conda-meta/*.json

### To fix on Linux:

    # in the command below, set PATH_TO_ENVIRONMENT yourself
    sed -i -E 's|("timestamp": [0-9]+)\.|\1|' /PATH_TO_ENVIRONMENT/conda-meta/*.json

### To fix on Windows:

Open notepad, and copy the contents below to `c:\fix_timestamps.py`

```python
PATH_TO_ENVIRONMENT="c:\\ProgramData\\Anaconda3"  # <-- fill this in yourself
                                                  # backslashes must be doubled

from glob import glob
import json
import os
for path in glob(os.path.join(PATH_TO_ENVIRONMENT, 'conda-meta', '*.json')):
    with open(path) as fh:
        content = json.load(fh)
    if 'timestamp' in content:
        old_timestamp = content['timestamp']
        content['timestamp'] = int(old_timestamp)
        if old_timestamp != content['timestamp']:
            with open(path, 'w') as fh:
               fh.write(json.dumps(content, indent=2, sort_keys=True, separators=(',', ': ')))
```

Also, change the path in the variable `PATH_TO_ENVIRONMENT` to point to the conda environment you want to fix.  Then run the script with `python c:\fix_timestamps.py`.

----

EDITED: I realized this is not a pyqt issue, it's a conda issue

### steps to reproduce

1. Install 64-bit miniconda on windows 10 (version 4.3.27)  
2. conda update to `4.3.28-py36h9daa44c_0`  
3. conda install -c anaconda spyder  
4. conda config --add channels conda-forge  

Now any calls to conda results in the print-out below.

### conda info

This prints any time I try to use a conda command.

```conda install -c dsdale24 pyqt5
An unexpected error has occurred.
Please consider posting the following information to the
conda GitHub issue tracker at:

    https://github.com/conda/conda/issues



Current conda install:

               platform : win-64
          conda version : 4.3.27
[...]

PR #5531, opened 2017-06-15
resolve #5530 add ability for users to choose to post unexpected errors to core maintainers

kalefranz:
resolve #5530

This PR also adds a new configuration parameter, `report_errors`.  The default value is `None`, which means conda will prompt and ask if the user wants to send the error report or not.  With

    conda config --set report_errors true

there won't be any prompt, and the error will automatically be sent.  This is equivalent to an opt-in always.  While

    conda config --set report_errors false

is equivalent to "always opt out."

PR #6720, opened 2018-01-11
issue template

kalefranz:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,59 @@
...
+
+This is an issue tracker for conda -- the package manager.  File feature requests
+for conda here, as well as bug reports about something conda has messed up.
+
+If your issue is a bug report or feature request for:
+  * a specific conda package from Anaconda:
+      ==> file at https://github.com/ContinuumIO/anaconda-issues
+  * a specific conda package from conda-forge:
+      ==> file at the corresponding feedstock under https://github.com/conda-forge
+  * repo.continuum.io:
```
mbargull: Will your "everyday casual user" always know that `repo.continuum.io` encompasses, e.g., "conda packages from the defaults channels" (... given unchanged `defaults`, ofc...)? `conda info` or at least `conda config --show` should make that obvious, but you know, people being unobservant/in a rush/etc. 😉 

[otherwise, good idea to use a template -- structured issues FTW!]

kalefranz: Top bullet is for packages from the `defaults` channel.  I meant this one to be more service problems.  I'll make that top bullet clearer.

kalefranz:
This has been reviewed by Crystal. Going to go ahead and merge.

.github/ISSUE_TEMPLATE.md
mbargull: Haha, me talking about "unobservant" and at the same time overlooking the *very first* bullet point >.<' -- but I also like your further clarifications, thanks!

BTW: Oh and, congrats about shipping the automatic issue reporting with `4.4.x` -- by the looks of the more recent issues you opened yourself, it seems that system works quite nicely!

kalefranz: The automatic reporting has been *so awesome*.  First, I think we implemented it well.  It's heavily opt-in by default; we don't ever send anything without explicit opt-in.  You actually have to type `y` or `yes`, not just hit enter (like with proceeding on `conda install`).  So it doesn't have a slimy feel.

But then, we're getting plenty of reports.  Total count for 4.4.7 is at 514 right now, and that's after we've been picking off the most noisy reports with every patch release.  It's an early warning system that probably would have made https://github.com/conda/conda/issues/6096 mostly a non-issue.  I would have known about the issue almost immediately after the release (probably in canary), and been able to pull it immediately to limit the damage.

mbargull: > heavily opt-in by default [...]  we implemented it well [...] it doesn't have a slimy feel.

Yeah, it looks like you directly hit the sweet spot, well done!

> we're getting plenty of reports

That "information flood" was what I immediately wondered about when I first saw #5531 -- but then again in the majority of cases, generated reports that directly include `conda info` and tracebacks should be much faster categorized/reproduced/fixed.

mbargull: Happy for you that it's working so well! 😄

PR #6722, opened 2018-01-12
conda.bat: fix #6713 allow parenthesis in prefix path

### LINKED ISSUE #6713: activate does not work.
when i perform the activate commad line, there are hints, but not work!
此时不应有 \Anaconda2\Library\bin\..\..\Scripts\conda.exe""。

C:\Users\username>    @SET "_CONDA_EXE="C:\Program Files (x86)\Anaconda2\Library\bin\..\..\Scripts\conda.exe""

"此时不应有 " it means: here they do not should eixst.

besides, my OS is Win10, 64bit.

mbargull:
Fixes #6713.
I can't say I know what's exactly going on, but there seems to be a problem with using parenthesis expressions and quoted arguments to `set` that include quotes and `)`, see the following:
```bat
@echo off

set X=")1"

(
  set X=")2"
)

( set X=")3" )

set "X=")4""

(
  set "X=")5""
)

( set "X=")6"" )
```
which fails for case 5 and 6:
```
5"" was unexpected at this time.
```
. I have no idea about the exact semantics for parenthesis expressions...
If we run into trouble with those expressions again, we could also just resort to replacing some
```bat
if condition (
  ...
) else (
  ...
)
```
with
```bat
if not condition goto :condition_else
  ...
goto :condition_end
:condition_else
  ...
:condition_end
```
.

### DIFF
--- a/conda/shell/Library/bin/conda.bat
+++ b/conda/shell/Library/bin/conda.bat
@@ -1,7 +1,7 @@
 @IF EXIST "%~dp0..\..\Scripts\conda.exe" (
-    @SET "_CONDA_EXE="%~dp0..\..\Scripts\conda.exe""
+    @SET _CONDA_EXE="%~dp0..\..\Scripts\conda.exe"
 ) ELSE (
-    @SET "_CONDA_EXE=python "%~dp0..\..\bin\conda""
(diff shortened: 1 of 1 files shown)


kalefranz:
Thanks for this Marcel.  Looks good to me to go with for now until we have more problems.

PR #6727, opened 2018-01-13
fix aggressive_update_packages availability

mbargull:
This is a followup to #6392 and makes `aggressive_update_packages` available under its designated name as it was previously only accessible via
```
_aggressive_update_packages:
  - ...
```
. Additionally, `aggressive_update_packages` wasn't allow to be empty before, which is also fixed.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -9,9 +9,10 @@ from os.path import (abspath, basename, dirname, expanduser, isdir, isfile, join
 from platform import machine
 import sys
 
-from .constants import (APP_NAME, DEFAULTS_CHANNEL_NAME, DEFAULT_CHANNELS, DEFAULT_CHANNEL_ALIAS,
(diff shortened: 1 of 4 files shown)


conda/base/context.py
```diff
@@ -120,7 +120,9 @@ class Context(Configuration):
     non_admin_enabled = PrimitiveParameter(True)
 
     # Safety & Security
-    _aggressive_update_packages = SequenceParameter(string_types)
+    _aggressive_update_packages = SequenceParameter(string_types,
+                                                    ('ca-certificates', 'certifi', 'openssl'),
```
kalefranz: We should add these as constants in `conda/base/constants.py`, and then use that value as the default here.  Probably just call it `DEFAULT_AGGRESSIVE_UPDATE_PACKAGES`.

mbargull: definitely! +1, done

PR #6735, opened 2018-01-15
issue template improvements

mbargull:
Some improvements to `.github/ISSUE_TEMPLATE.md` I like to suggest after seeing the first uses of the issue template.
Changes are in single single commits so can be (mostly) cherry-picked if only a subset is used. Changes are:
* Add `Environment Information` section: so that `conda info` etc. are not part of `Expected Behavior` secion.
* `Steps to Reproduce` can be part of `Current Behavior`: mainly to hint at/"indirectly suggest" to rather put output from commands there.
* Directly suggest to include console output between tick-enclosed lines: to hopefully make the user put tracebacks and such there for better formatting.
* Reducing subsection depth: gives better visual separation, esp. if `Steps to Reproduce` is moved inside `Current Behavior`.
* Put `conda (info|config|list)` output in `<details>` HTML tags: makes them collapsible which is esp. useful for `conda list` output, e.g., with long package lists from `anaconda` users. `conda (info|config)` outputs are uncollapsed on default whereas `conda list` is collapsed.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -45,22 +45,28 @@ If your issue is a bug report or feature request for:
 
 
 ## Environment Information
-##### `conda info`
+<details open><summary><code>`conda info`</code></summary><p>
 <!-- between the ticks below, paste the output of 'conda info' -->
-```
 
 ```
 
+```
+</p></details>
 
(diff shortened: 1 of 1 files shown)


mbargull:
For preview purposes, I used that template for #6736 :wink:.

PR #6741, opened 2018-01-17
resolve #6733 make empty env vars sequence-safe for sequence parameters

### LINKED ISSUE #6733: Empty sequences for configuration parameters cannot be expressed in environment variables
**I'm submitting a...**
  - [ ] bug report
  - [x] feature request

### Current Behavior

For a configuration parameter that accepts sequences, an empty value in the corresponding environment variable gets interpreted as an one-element sequence containing an empty string, i.e., `[""]`. Thus there is no way to specify an empty sequence via an environment variable.

### Steps to Reproduce

```bash
$ export CONDA_CHANNELS=
$ conda config --show channels
channels:
  - 
$ export CONDA_CHANNELS=defaults,
$ conda config --show channels
channels:
  - defaults
  - 
$ export CONDA_CHANNELS=defaults,,
$ conda config --show channels
channels:
  - defaults
  - 
```

### Expected Behavior

One possibility to allow empty sequences would be to allow and ignore trailing separator, i.e., interpret
  * `,` as `[]`,
  *  `,,` as `[""]`,
  * `,,,` as `["", ""]`,
  * `defaults,` as `["defaults"]`,
  * `defaults,,` as `["defaults", ""]`.

The commands given above would then yield:
```bash
$ export CONDA_CHANNELS=
$ conda config --show channels
channels: []
$ export CONDA_CHANNELS=defaults,
$ conda config --show channels
channels:
  - defaults
$ export CONDA_CHANNELS=defaults,,
$ conda config --show channels
channels:
  - defaults
  - 
```

##### `conda info`

```
     active environment : base
    active env location : /home/maba/code/conda/conda-4.4
            shell level : 1
       user config file : /home/maba/.condarc
 populated config files : /home/maba/code/conda/conda-4.4/.condarc
          conda version : 4.4.7
       base environment : /home/maba/code/conda/conda-4.4  (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/linux-64
                          https://repo.continuum.io/pkgs/free/linux-64
                          https://repo.continuum.io/pkgs/r/linux-64
                          https://repo.continuum.io/pkgs/pro/linux-64
          package cache : /home/maba/code/conda/conda-4.4/pkgs
                          /home/maba/.conda/pkgs
       envs directories : /home/maba/code/conda/conda-4.4/envs
                          /home/maba/.conda/envs
               platform : linux-64
             user-agent : conda/4.4.7 requests/2.18.4 CPython/3.6.4 Linux/4.14.12-1-MANJARO manjaro/17.1-dev glibc/2.26
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

##### `conda config --show-sources`

(not needed)

##### `conda list --show-channel-urls`

(not needed)

kalefranz:
resolve #6733 

@mbargull Let's see how this works...  I don't even know if there's a situation where we'd actually *want* an empty string, instead of just an empty sequence.

### DIFF
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -206,6 +206,7 @@ class EnvRawParameter(RawParameter):
 
     def value(self, parameter_obj):
         if hasattr(parameter_obj, 'string_delimiter'):
+            assert isinstance(self._raw_value, string_types)
             string_delimiter = getattr(parameter_obj, 'string_delimiter')
             # TODO: add stripping of !important, !top, and !bottom
             return tuple(v for v in (
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -210,9 +210,9 @@ class EnvRawParameter(RawParameter):
             # TODO: add stripping of !important, !top, and !bottom
             raw_value = self._raw_value
             if string_delimiter in raw_value:
-                value = raw_value.split(string_delimiter)
+                value = [v for v in raw_value.split(string_delimiter) if v]
             else:
-                value = [raw_value]
+                value = [raw_value] if raw_value else []
             return tuple(v.strip() for v in value)
         else:
             return self.__important_split_value[0].strip()
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -208,12 +208,9 @@ class EnvRawParameter(RawParameter):
         if hasattr(parameter_obj, 'string_delimiter'):
             string_delimiter = getattr(parameter_obj, 'string_delimiter')
             # TODO: add stripping of !important, !top, and !bottom
-            raw_value = self._raw_value
-            if string_delimiter in raw_value:
-                value = [v for v in raw_value.split(string_delimiter) if v]
-            else:
-                value = [raw_value] if raw_value else []
-            return tuple(v.strip() for v in value)
+            return tuple(v for v in (
+                vv.strip() for vv in self._raw_value.split(string_delimiter)
+            ) if v)
         else:
             return self.__important_split_value[0].strip()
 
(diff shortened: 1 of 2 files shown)


conda/common/configuration.py
```diff
@@ -210,9 +210,9 @@ def value(self, parameter_obj):
             # TODO: add stripping of !important, !top, and !bottom
             raw_value = self._raw_value
             if string_delimiter in raw_value:
```
mbargull: `self._raw_value` is always a `str`, right?
`str.split` always returns a list of strings so that you can simplify the whole if-else-block to
```python
value = [v for v in self._raw_value.split(string_delimiter) if v]
```
. If you want to allow empty strings anywhere in the sequence, you could use a trailing separator to distinguish `""`->`[]` from `","`->`[""]` with
```python
value = self._raw_value.split(string_delimiter)  # str.split always returns a list with length >=1
value = value if value[-1] else value[:-1]
```

kalefranz: > self._raw_value is always a str, right?

I guess in this case that's safe to assume.  I feel like I'd want to add an `assert` in there though just to be safe.

> If you want to allow empty strings anywhere in the sequence

Can you think of a use-case that we currently have, or one that we could potentially have, where you'd ever want an empty string?  I can't, at the moment.

mbargull: The currently available `SequenceParameter`s in `base/context.py` are lists of packages (specs), features, paths (/subdirs), or channels.
So, no real use case there. Hence this would be more of a "possible forward-compatibility" decision.

kalefranz: I say for now that we don't support empty strings.  I can see it just as easily causing us problems in the future, as it could saving us from a future problem.

mbargull: That's fine by me. My mind is oscillating between 30/70, 50/50, and 70/30 on whether the one or the other possibility would cause more potential issues. Currently I'm slightly more leaning toward removing empty strings, just like you did :wink:. So, be fast and merge before I change my mind again and again 😝

kalefranz:
Unit test added!

PR #6757, opened 2018-01-19
fix several lgtm.com flags

kalefranz:
https://lgtm.com/projects/g/conda/conda/alerts/?mode=tree&severity=error

I don't think this covers everything, but it's a huge swath.  Eventually we can probably turn this on for all PRs as part of the pass/fail CI.

### DIFF
--- a/conda/_vendor/auxlib/entity.py
+++ b/conda/_vendor/auxlib/entity.py
@@ -544,12 +544,12 @@ class DateField(Field):
 class EnumField(Field):
 
     def __init__(self, enum_class, default=NULL, required=True, validation=None,
-                 in_dump=True, default_in_dump=True, nullable=False, immutable=False):
+                 in_dump=True, default_in_dump=True, nullable=False, immutable=False, aliases=()):
         if not issubclass(enum_class, Enum):
             raise ValidationError(None, msg="enum_class must be an instance of Enum")
         self._type = enum_class
         super(EnumField, self).__init__(default, required, validation,
-                                        in_dump, default_in_dump, nullable, immutable)
+                                        in_dump, default_in_dump, nullable, immutable, aliases)
 
     def box(self, instance, instance_type, val):
         if val is None:
@@ -573,10 +573,10 @@ class ListField(Field):
     _type = tuple
 
     def __init__(self, element_type, default=NULL, required=True, validation=None,
-                 in_dump=True, default_in_dump=True, nullable=False, immutable=False):
+                 in_dump=True, default_in_dump=True, nullable=False, immutable=False, aliases=()):
         self._element_type = element_type
         super(ListField, self).__init__(default, required, validation,
-                                        in_dump, default_in_dump, nullable, immutable)
+                                        in_dump, default_in_dump, nullable, immutable, aliases)
 
     def box(self, instance, instance_type, val):
         if val is None:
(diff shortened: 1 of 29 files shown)


conda/_vendor/auxlib/entity.py
```diff
@@ -573,10 +573,10 @@ class ListField(Field):
     _type = tuple
 
     def __init__(self, element_type, default=NULL, required=True, validation=None,
-                 in_dump=True, default_in_dump=True, nullable=False, immutable=False):
+                 in_dump=True, default_in_dump=True, nullable=False, immutable=False, aliases=()):
         self._element_type = element_type
         super(ListField, self).__init__(default, required, validation,
```
goanpeca: I think we should be using explicit named arguments

conda/cli/main_list.py
```diff
@@ -50,15 +50,18 @@ def list_packages(prefix, installed, regex=None, format='human',
...
-                    and schannel != DEFAULTS_CHANNEL_NAME):
-                disp += '  %s' % schannel
-            result.append(disp)
+            if info is None:
+                result.append('%-25s %-15s %15s' % tuple(dist.quad[:3]))
+            else:
+                features = set(info.get('features') or ())
+                disp = '%(name)-25s %(version)-15s %(build)15s' % info  # NOQA lgtm [py/percent-format/wrong-arguments]
+                disp += '  %s' % disp_features(features)
+                schannel = info.get('schannel')
```
goanpeca: what does `schannel` stands for?

conda/common/url.py
```diff
@@ -7,11 +7,11 @@
 import re
 import socket
 
+from .compat import input, on_win
```
goanpeca: This is overloading Python's `input` right?

kalefranz: Yes. And that's exactly what ltgm was suggesting we to here.
https://lgtm.com/rules/3980089/

conda/cli/main_list.py
kalefranz: I think "short channel".  I don't know.  It predates me.  It's now more commonly `channel.canonical_name`.

conda/_vendor/auxlib/entity.py
kalefranz: What do you mean?

conda/cli/main_list.py
kalefranz: I don't like these names here.  This is an old, crusty piece of the code that could use a makeover.  However, the only reason there's a code change here is because of the indent.  The makeover is probably beyond the scope of this PR.  We can do it in a future PR though.

conda/cli/install.py
```diff
@@ -122,7 +122,7 @@ def get_revision(arg, json=False):
     try:
         return int(arg)
     except ValueError:
-        CondaValueError("expected revision number, not: '%s'" % arg, json)
+        raise CondaValueError("expected revision number, not: '%s'" % arg, json)
```
kalefranz: I think this is one of the only actual errors that lgtm.com discovered.

conda/install.py
```diff
@@ -32,7 +32,7 @@
 from .core.linked_data import is_linked, linked, linked_data  # NOQA
 is_linked, linked, linked_data = is_linked, linked, linked_data
 from .core.package_cache import rm_fetched  # NOQA
-rm_fetched = rm_fetched
+rm_fetched = rm_fetched  # lgtm [py/redundant-assignment]
```
goanpeca: I mean

```
from .core.linked_data import is_linked, linked, linked_data  # NOQA
is_linked, linked, linked_data = is_linked, linked, linked_data   
from .core.package_cache import rm_fetched  # NOQA
```

vs

```
from .core.linked_data import is_linked, linked, linked_data  # NOQA
from .core.package_cache import rm_fetched  # NOQA 
is_linked, linked, linked_data = is_linked, linked, linked_data
```

conda/_vendor/auxlib/entity.py
goanpeca:
```
 super(ListField, self).__init__(default, required, validation,
```

vs

```
 super(ListField, self).__init__(default, required=required, validation=validation,
```

?

mbargull: If @goanpeca means keyword-only arguments, then I don't know, they look ugly when you have to support Python 2 and don't have https://www.python.org/dev/peps/pep-3102/ ...

PR #6760, opened 2018-01-20
Preserve status in fish shell

Zaharid:
The fish_prompt function typically does things like changing the color
of the prompt based on the exit status of the last command. This
status is however lost in the function defined by conda, since the
status of the last operation is always zero.

This fixes the issue by defining an identity function that returns the
last status (thereby setting the $status variable) before the original
fish_prompt function is called.

### DIFF
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -39,10 +39,17 @@ else
     function __fish_prompt_orig
     end
 end
+
+function return_last_status
+  return $argv
+end
+
 function fish_prompt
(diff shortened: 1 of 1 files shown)


kalefranz:
Is there an existing issue number that this PR is meant to address?

Zaharid:
Not that I know of.

kalefranz:
@jaimergp Do you have any comments here?

jaimergp:
I didn't even know I was losing this info in my fish theme (bobthefish, btw) because I was using conda before fish...

I have tested the new changes and looks good to me:
fish_exit

PR #6719, opened 2018-01-11
fix #6718 broken 'conda install --rev'

kalefranz:
fix #6718
resolve #6230 
fix #6290

Issue #6269, opened 2017-11-06
Bug in conda dependency solver


lparsons:
Attempting to install `trinity==2.4.0` results in `bowtie=2.2.5=0` being installed. This is not the lastest bowtie2 build (there is a build 2) and it doesn't work. Also, attempting to install `bowtie==2.2.5` results in the broken build 0 install. However, first installing `bowtie=2.2.5_py36_2` does work and subsequently installing `trinity==2.4.0` into that envrionment results in `bowtie2 2.3.3.1-py36pl5.22.0_0` being installed. Then everything works.

From a discussion on Gitter with @mbargull :
> Sounds like Conda's solver didn't prioritize its objectives (maximize req. pkg. version, maximize req. pkg. build number, minimize number of packages) correctly. The cause seems to be the dependency on pythonfor the newer builds which requires more packages to be installed. If you explicitly add python to the list of packages of your conda install/create command, you'll see it will then pick up the newer build. Interestingly conda 4.4.0rc1 behaves correctly.

mbargull:
This is about the `bowtie2=2.2.5` package from the `bioconda` channel.
The difference between `bowtie2=2.2.5=0` and `bowtie2=2.2.5=2` are mainly the added dependencies, notably `python`.

See the following logs which show the described behavior:
<details><summary><code>
(ins)(root) [miniconda3]$ cat .condarc
</code></summary><p>

```
default_channels:
  - https://repo.continuum.io/pkgs/free
```
</p></details>

<details><summary><code>
(ins)(root) [miniconda3]$ conda info
</code></summary><p>

```
Current conda install:

               platform : linux-64
          conda version : 4.3.30
       conda is private : False
      conda-env version : 4.3.30
    conda-build version : 3.0.19
         python version : 3.5.4.final.0
       requests version : 2.14.2
       root environment : /home/maba/code/miniconda3  (writable)
    default environment : /home/maba/code/miniconda3
       envs directories : /home/maba/code/miniconda3/envs
                          /home/maba/.conda/envs
          package cache : /home/maba/code/miniconda3/pkgs
                          /home/maba/.conda/pkgs
           channel URLs : https://repo.continuum.io/pkgs/free/linux-64
                          https://repo.continuum.io/pkgs/free/noarch
            config file : /home/maba/code/miniconda3/.condarc
             netrc file : None
           offline mode : False
[...]

PR #6766, opened 2018-01-23
Adjust the feature score assigned to packages not installed

### LINKED ISSUE #6765: Unintentional build downgrades in mixed feature settings
Under certain limited circumstances involving features, `conda` unexpectedly downgrades to an earlier build of a package. The conditions are as follows:

- The current version of a package `foo` has multiple builds with distinct build numbers.
- Newer builds depend on a second package `bar`, while older builds do not.
- Some of the valid choices for the `bar` dependency have features, and some do not.

When those conditions are met, the command
```
conda create -n errtest foo
```
will select an older build of `foo`, without the `bar` dependency, instead of the latest build. (Contact me or @msarahan for a specific test case.)

Currently, this is not occurring with conda 4.4. The reasons for this are not fully clear, because the bug has been located and exists in 4.3 and 4.4. Bug fix coming.

mcg1969:
Fixes #6765 . Built off of 4.3.x, but it should to port to 4.4.x with no changes.

The bug is this: consider the first three optimization passes:

1. Maximize the channel priority and version of the requested packages
2. Minimize the number of features required
3. Maximize the number of installed packages that utilize the features
4. Maximize the build number of the requested packages

We deliberately placed 4 after 2 and 3 in response to previous issues; for instance, if there are fewer builds of a featured package than a non-featured package, we need to be able to reach back and grab the featured build if requested. Now, per #6765, however, consider this scenario:

- The current version of a package `foo` has multiple builds with distinct build numbers.
- Newer builds depend on a second package `bar`, while older builds do not.
- Some of the valid choices for the `bar` dependency have features, and some do not.

The problem arises in the implementation 3. In effect, we were giving the same score to a package that is _not installed_ as we were to packages that are _installed with active features_. Therefore, if `foo` is requested, `conda` actually preferred _downgrading_ to an older build, if doing so allowed it to remove `bar` altogether.

To fix this, we're actually giving the SAT variable corresponding to "don't install `bar`" the same score as a package with no features. This will give `conda` a slight preference for pulling in unnecessary packages. However, `get_reduced_index` ensures that most unnecessary packages will not even be considered, so the likelihood of this occurring is small. And it's also a somewhat benign effect, compared to the downgrading behavior itself.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -628,7 +628,14 @@ class Resolve(object):
         for name, group in iteritems(self.groups):
             nf = [len(self.features(dist)) for dist in group]
             maxf = max(nf)
-            eq.update({dist.full_name: maxf-fc for dist, fc in zip(group, nf) if fc < maxf})
+            if maxf > 0:
+                eq.update({dist.full_name: maxf-fc for dist, fc in zip(group, nf) if maxf > fc})
+                # This entry causes conda to weight the absence of a package the same
+                # as if a non-featured version were instaslled. If this were not here,
+                # conda will sometimes select an earlier build of a requested package
+                # if doing so can eliminate this package as a dependency.
+                # https://github.com/conda/conda/issues/6765
+                eq['!' + self.push_MatchSpec(C, name)] = maxf
             total += maxf
         return eq, total
 
--- a/tests/test_resolve.py
+++ b/tests/test_resolve.py
@@ -766,6 +766,28 @@ def test_install_package_with_feature():
     r.install(['mypackage','feature 1.0'])
 
 
+def test_unintentional_feature_downgrade():
+    # See https://github.com/conda/conda/issues/6765
+    # With the bug in place, this bad build of scipy
+    # will be selected for install instead of a later
+    # build of scipy 0.11.0.
+    good_rec = index[Dist('scipy-0.11.0-np17py33_3.tar.bz2')]
+    bad_deps = tuple(d for d in good_rec.depends
+                     if not d.startswith('numpy'))
+    bad_rec = IndexRecord.from_objects(good_rec,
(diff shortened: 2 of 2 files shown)


mcg1969:
It would be fair to ask whether or not this PR is a good opportunity to insert a significant amount of documentation about each of these more esoteric functions in the resolve object.

mingwandroid:
Would it be possible to have meaningful debug information printed out at each pass where it gives human readable package names? For me, that would make investigating issues with the solver a much less scary proposition.

mcg1969:
It’s important to note that we actually don’t have an explicit list of selected packages until the very end. If we print things after each pass it will just be the gigantic list of SAT clauses. 

And doing queries of the current SAT clause set to determine the current set of candidate recipes is destructive. So we’d have to create a new code path to do that when debugging. I think that could be quite powerful and interesting though.

kalefranz:
> It would be fair to ask whether or not this PR is a good opportunity to insert a significant amount of documentation about each of these more esoteric functions in the resolve object.

Any PR is a good opportunity to add more documentation, especially for resolve.py code.

kalefranz:
LGTM. Did you want to include any additional comments @mcg1969 ?

mbargull:
Very nice! This will directly fix the source of these issues: https://github.com/conda/conda/issues/6269, https://github.com/bioconda/bioconda-recipes/issues/6637, https://github.com/bioconda/bioconda-recipes/issues/6163, https://github.com/ContinuumIO/anaconda-issues/issues/7314.

kalefranz:
Ok I'm going to merge.  @mcg1969 Can add more comments later (please do!).

kalefranz:
Once we get https://github.com/conda/conda/pull/6719 merged, I'll release 4.3.33 with the two changes.

PR #6768, opened 2018-01-23
Add workaround for system Python usage by lsb_release

mingwandroid:
See: https://github.com/ContinuumIO/anaconda-issues/issues/8244


kalefranz:
targeting 4.4.x via #6769

PR #6782, opened 2018-01-24
fix #5884 allow --insecure CLI flag without contradicting meaning of ssl_verify

### LINKED ISSUE #5884: ssl_verify functionality doesn't match description
Was experimenting to see if I could use a cert directory on Windows 7:
```
> conda config --set ssl_verify C:\Users\HarmerS\certs
```

However it seems like the description of `ssl_verify` doesn't agree with the implementation.
```
> conda config --describe
[...]
ssl_verify (basestring, bool)
  default: true
  aliases: insecure, verify_ssl
  Conda verifies SSL certificates for HTTPS requests, just like a web
  browser. By default, SSL verification is enabled, and conda operations
  will fail if a required url's certificate cannot be verified. Setting
  ssl_verify to False disables certification verificaiton. The value for
  ssl_verify can also be (1) a path to a CA bundle file, or (2) a path
  to a directory containing certificates of trusted CA.
[...]

> conda install -c conda-forge pydicom

CustomValidationError: Parameter ssl_verify = 'C:\\Users\\HarmerS\\certs' declared in <<merged>> is invalid.
ssl_verify value 'C:\Users\HarmerS\certs' must be a boolean or a path to a certificate bundle.
```

```
> conda --version
conda 4.3.25
```
Via Miniconda 4.3.21 [installer](https://repo.continuum.io/miniconda/Miniconda2-latest-Windows-x86.exe) Python 2.7 32-bit + `conda update conda` on Windows 7 64-bit.

I would prefer to have the feature of using a cert directory (much easier to manage) so I raised the issue here rather than under conda-docs.

kalefranz:
resolve #5884

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -95,9 +95,10 @@ def default_python_validation(value):
 
 def ssl_verify_validation(value):
     if isinstance(value, string_types):
-        if not isfile(value):
-            return ("ssl_verify value '%s' must be a boolean or a path to a "
-                    "certificate bundle." % value)
+        if not isfile(value) and not isdir(value):
+            return ("ssl_verify value '%s' must be a boolean, a path to a "
+                    "certificate bundle file, or a path to a directory containing "
+                    "certificates of trusted CAs." % value)
     return True
 
 
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -147,7 +147,7 @@ class Context(Configuration):
 
     # remote connection details
     ssl_verify = PrimitiveParameter(True, element_type=string_types + (bool,),
-                                    aliases=('insecure', 'verify_ssl',),
+                                    aliases=('verify_ssl',),
                                     validation=ssl_verify_validation)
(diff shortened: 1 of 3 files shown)


mcg1969:
While the PR itself seems reasonable I don't see how this resolves #5884.

kalefranz:
I guess it doesn't.  I just came across the inconsistency when I was looking into the issue.  We should probably just close #5884 directly.

kalefranz:
Actually, I didn't understand what the issue was in #5884.  I finally do.  What's being called out is this

```
CustomValidationError: Parameter ssl_verify = 'C:\\Users\\HarmerS\\certs' declared in <<merged>> is invalid.
ssl_verify value 'C:\Users\HarmerS\certs' must be a boolean or a path to a certificate bundle.
```

That I can definitely fix.

kalefranz:
So we actually *do* fix #5884 in this PR now.

cjmsantos:
Resolved

PR #6790, opened 2018-01-25
fix #6457 env var cleanup

### LINKED ISSUE #6457: Sanitize env vars to remove secrets (e.g credentials for PROXY)
For operations like `conda info -a` and error report uploads.

kalefranz:
fix #6457

### DIFF
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -195,25 +195,29 @@ def get_info_dict(system=False):
         info_dict['UID'] = os.geteuid()
         info_dict['GID'] = os.getegid()
 
-    evars = {
+    env_var_keys = {
         'CIO_TEST',
         'REQUESTS_CA_BUNDLE',
         'SSL_CERT_FILE',
     }
 
     # add all relevant env vars, e.g. startswith('CONDA') or endswith('PATH')
-    evars.update(v for v in os.environ if v.upper().startswith('CONDA'))
-    evars.update(v for v in os.environ if v.upper().startswith('PYTHON'))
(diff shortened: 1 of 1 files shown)


conda/cli/main_info.py
```diff
@@ -195,25 +195,29 @@ def get_info_dict(system=False):
...
+    env_var_keys = {
         'CIO_TEST',
         'REQUESTS_CA_BUNDLE',
         'SSL_CERT_FILE',
     }
 
     # add all relevant env vars, e.g. startswith('CONDA') or endswith('PATH')
-    evars.update(v for v in os.environ if v.upper().startswith('CONDA'))
-    evars.update(v for v in os.environ if v.upper().startswith('PYTHON'))
-    evars.update(v for v in os.environ if v.upper().endswith('PROXY'))
```
mcg1969: It looks like you've _removed_ PROXY variables from the upload. But I'm not sure we want to do that, because that may be a genuine clue to some download issues. I'd rather we try and scrub them of any identifying information, even if it's just reduced to a simple boolean indicating "this variable is set".

kalefranz: Check out line 213...

ahmadia: I think it's definitely a good idea not to send PROXY variables or proxy variables, since they can unintentionally disclose passwords in a few environments. I don't think there's a problem indicating if they are set.

mcg1969: Great, thanks @ahmadia. It looks like that's what he's doing so I'm good

PR #6794, opened 2018-01-25
resolve #6791 conda search --envs

kalefranz:
resolve #6791 

TODO after merging this PR: Rename `PackageCache` to `PackageCacheData`

PR #6807, opened 2018-01-28
add back support for search --reverse-dependency

### LINKED ISSUE #6670: conda search --reverse-dependency not working on conda 4.4
The `--reverse-dependency` argument does not seem to effect the results of `conda search` in conda 4.4:

```
~$ conda search --reverse-dependency networkx
Loading channels: done
Name                       Version                   Build  Channel        
networkx                   1.7                      py26_0  defaults       
networkx                   1.7                      py27_0  defaults       
networkx                   1.7                      py33_0  defaults       
networkx                   1.8                      py26_0  defaults       
networkx                   1.8                      py27_0  defaults
...

~$ conda search --reverse-dependency networkx
Fetching package metadata ...........
anaconda                     1.6.0                np16py26_0  defaults        
                             1.6.0                np16py27_0  defaults        
                             1.6.0                np17py26_0  defaults        
                             1.6.0                np17py27_0  defaults        
                             1.6.0                np17py33_0  defaults        
                             1.6.1                np16py26_0  defaults        
                             1.6.1                np16py27_0  defaults        
                             1.6.1                np17py26_0  defaults        
                             1.6.1                np17py27_0  defaults        
                             1.6.1                np17py33_0  defaults
...
```

jjhelmus:
Add back support for the --reverse-dependency argument in conda search.

closes #6670

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -918,9 +918,8 @@ def configure_parser_search(sub_parsers):
     p.add_argument(
         "--reverse-dependency",
         action="store_true",
-        help="Perform a reverse dependency search. When using this flag, the --full-name "
-             "flag is recommended. Use 'conda info package' to see the dependencies of a "
-             "package.",
+        help="Perform a reverse dependency search. Use 'conda info package' to see the "
+             "dependencies of a package.",
     )
     add_parser_offline(p)
     add_parser_channels(p)
--- a/conda/core/repodata.py
+++ b/conda/core/repodata.py
@@ -75,6 +75,24 @@ class SubdirData(object):
     _cache_ = {}
 
     @staticmethod
+    def reverse_query_all(channels, subdirs, package_ref_spec):
+        from .index import check_whitelist  # TODO: fix in-line import
+        channel_urls = all_channel_urls(channels, subdirs=subdirs)
+        check_whitelist(channel_urls)
+        with ThreadLimitedThreadPoolExecutor() as executor:
+            futures = (executor.submit(
+                SubdirData(Channel(url)).reverse_query, package_ref_spec
+            ) for url in channel_urls)
+            return tuple(concat(future.result() for future in as_completed(futures)))
+
+    def reverse_query(self, package_ref_spec):
+        if not self._loaded:
+            self.load()
+        for prec in self._package_records:
+            if any(package_ref_spec.match(MatchSpec(drec)) for drec in prec.depends):
+                yield prec
+
+    @staticmethod
     def query_all(channels, subdirs, package_ref_or_match_spec):
         from .index import check_whitelist  # TODO: fix in-line import
         channel_urls = all_channel_urls(channels, subdirs=subdirs)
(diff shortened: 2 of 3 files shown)


conda/cli/conda_argparse.py
```diff
@@ -918,9 +918,8 @@ def configure_parser_search(sub_parsers):
     p.add_argument(
         "--reverse-dependency",
         action="store_true",
-        help="Perform a reverse dependency search. When using this flag, the --full-name "
-             "flag is recommended. Use 'conda info package' to see the dependencies of a "
-             "package.",
+        help="Perform a reverse dependency search. Use 'conda info package' to see the "
```
kalefranz: I'd rather have documentation here say

    'conda search spec --info'

kalefranz:
Still doing some review... Sorry for the single message... Not done yet though.

conda/core/repodata.py
```diff
@@ -74,6 +74,24 @@ def __call__(cls, channel):
...
+            futures = (executor.submit(
+                SubdirData(Channel(url)).reverse_query, package_ref_spec
+            ) for url in channel_urls)
+            return tuple(concat(future.result() for future in as_completed(futures)))
+
+    def reverse_query(self, package_ref_spec):
+        if not self._loaded:
+            self.load()
+        for prec in self._package_records:
+            if any(package_ref_spec.match(MatchSpec(drec)) for drec in prec.depends):
```
kalefranz: You can't match a MatchSpec to a MatchSpec.  We haven't written in the capability to "merge", or determine overlaps, in e.g. version ranges.

kalefranz: Do you want to use `prec.depends` or `prec.combined_depends`?

conda/core/repodata.py
```diff
@@ -74,6 +74,24 @@ def __call__(cls, channel):
...
+        from .index import check_whitelist  # TODO: fix in-line import
+        channel_urls = all_channel_urls(channels, subdirs=subdirs)
+        check_whitelist(channel_urls)
+        with ThreadLimitedThreadPoolExecutor() as executor:
+            futures = (executor.submit(
+                SubdirData(Channel(url)).reverse_query, package_ref_spec
+            ) for url in channel_urls)
+            return tuple(concat(future.result() for future in as_completed(futures)))
+
+    def reverse_query(self, package_ref_spec):
```
kalefranz: Let's call `package_ref_spec ` instead `package_ref_or_match_spec` for consistency.  `query()` is intended to be an API-level function, so we should get this right here.

kalefranz: Let's call `package_ref_spec ` instead `package_ref_or_match_spec` for consistency.  `query()` is intended to be an API-level function, so we should get this right here.

conda/core/repodata.py
kalefranz: > You can't match a MatchSpec to a MatchSpec.

Actually, that might be exactly the reason this got dropped from 4.4.  I think it's harder to implement than it seems on the surface.  It's going to take a lot of work to get `package_ref_spec.match(MatchSpec(drec))` right.

kalefranz: Or... For now we could lock this down to just `package_ref`, and not `package_ref_or_match_spec`.  But then we'd have to change that API in the future if we do end up adding the matchspec compare/merge capability.

nehaljwani:
Friendly bump! This feature is really needed.

kalefranz:
I've created #6888 that builds on this PR and should get us a lot closer.  It also builds on #6794, which did have some conflicts to resolve.

PR #6825, opened 2018-02-01
fix broken Anaconda Prompt for users with spaces in paths

kalefranz:


mbargull:
At first glance, this looks okay as a workaround (didn't fire up a VM to test this, though).
So a limitation would be if a user installs `console_short` into a non-base environment, so that they get a
```
%windir%\system32\cmd.exe "/K" "C:\Users\includes space\Miniconda3\Scripts\activate.bat" "C:\Users\includes space\Miniconda3\envs\some-env"
```
shortcut. The shortcut would then launch the base environment instead. But that's an edge case (which is currently broken either way) that is far less important than the one that you want to fix.

Are there any mechanisms  in place that let's you test this with AppVeyor?

*Very minor* comments:
* The `::` "comments" can be a bit finicky sometimes but should be fine in the way used here, I think. As with so many other things, it's [crazy](http://www.robvanderwoude.com/comments.php) [madness](https://stackoverflow.com/a/12408045) 😆 
* `_c1=` could be unset right after line 10.

kalefranz:
I think

    %windir%\system32\cmd.exe "/K" "C:\Users\includes space\Miniconda3\Scripts\activate.bat" "C:\Users\includes space\Miniconda3\envs\some-env"

is going to be broken in all cases anyway right now.

So what we're doing here is attempting to fix conda 4.4 users who have a bad shortcut created, and then fix menuinst for future shortcuts with https://github.com/ContinuumIO/menuinst/pull/61

kalefranz:
and my gosh did conda-forge release something with a version `2.0.0~alpha0`?????

mbargull:
> is going to be broken in all cases anyway right now.

Correct.

> and my gosh did conda-forge release something with a version 2.0.0~alpha0?????

Yeah, look like it. Some hours ago I wanted to try out some stuff with `conda search` which horribly choked on that `2.0.0~alpha0` 😆

kalefranz:
Or did we release that?

kalefranz:
Dist(channel='conda-forge', dist_name='pyside2-2.0.0~alpha0-py36_0', name='pyside2', version='2.0.0~alpha0', build_string='py36_0', build_number=0, base_url=None, platform=None)

conda/shell/Scripts/activate.bat
```diff
@@ -1 +1,25 @@
...
+@REM   %windir%\system32\cmd.exe "/K" "C:\Users\builder\Miniconda3\Scripts\activate.bat" "C:\Users\builder\Miniconda3"
+@REM Rather than the correct
+@REM    %windir%\system32\cmd.exe /K ""C:\Users\builder\Miniconda3\Scripts\activate.bat" "C:\Users\builder\Miniconda3""
+@REM this solution taken from https://stackoverflow.com/a/31359867
+@set "_c1=%1"
+@set _c1f=%_c1:~0,1%
+@set _c1l=%_c1:~-1%
+@set _c1f=%_c1f:"=+%
+@set _c1l=%_c1l:"=+%
+@set _c1=
```
jjhelmus: Any reason not to use more descriptive variable names here?

conda/shell/Scripts/activate.bat
```diff
@@ -1 +1,25 @@
...
+
+@if "%_c1f%"=="+" if NOT "%_c1l%"=="+" (
+    @CALL "%~dp0..\Library\bin\conda.bat" activate
```
jjhelmus: This will activate the root/base environment correct?  Is that the intended behavior?  Do we want to echo any type of warning message about dropping the command line parameters?

kalefranz: The point of this is to give conda 4.4 the behavior conda 4.3 and earlier had.  If you had a directory `C:\Users\build er\Miniconda3`, and were to do something like

    %windir%\system32\cmd.exe "/K" "C:\Users\build er\Miniconda3\Scripts\activate.bat" "C:\Users\build er\Miniconda3"

and also put `echo %*` at the top of `activate.bat`, you get 

    "C:\Users\build

as output.  Which is total crap.  But conda 4.3 and earlier just swallowed that, and activated the root environment.  That's all we're trying to accomplish here.

conda/shell/Scripts/activate.bat
kalefranz: ok variable names changed

jjhelmus:
LGTM, will merge when CI are done

PR #6841, opened 2018-02-05
JSONDecodeError was added in Python 3.5

opoplawski:


kalefranz:
targeting 4.4.x via #6848

PR #6853, opened 2018-02-07
fix #6833 improve feature mismatch metric

### LINKED ISSUE #6833: Regression in conda 4.3.33 regarding build features
It seems that `conda 4.3.33` installs packages that have a build feature when installing any package that has the same build feature.

From a fresh miniconda, with conda pinned at `4.3.32`:

```sh
scorlay@mbp ~]$ conda install cryptopp -c QuantStack -c conda-forge
Fetching package metadata ...............
Solving package specifications: .

Package plan for installation in environment /Users/scorlay/miniconda3:

    cling-patches: 1-0           QuantStack
    cryptopp:      5.6.5-cling_0 QuantStack [cling]
```

With `conda 4.3.33`

```sh
[scorlay@mbp ~]$ conda install cryptopp -c QuantStack -c conda-forge
Fetching package metadata ...............
Solving package specifications: .

Package plan for installation in environment /Users/scorlay/miniconda3:

    clangdev:      5.0.0-cling_1 QuantStack  [cling]
    cling-patches: 1-0           QuantStack 
    cryptopp:      5.6.5-cling_0 QuantStack  [cling]
    icu:           58.2-0        conda-forge
    libiconv:      1.15-0        conda-forge
    libxml2:       2.9.7-0       conda-forge
    llvmdev:       5.0.0-cling_1 QuantStack  [cling]
```

The only package tracking the `cling` build feature is `cling-patches`.

Notice in the second case that `clangdev` gets installed while it is not a dependency.

mcg1969:
Attempts to fix #6833. Will copy the relevant text from there

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -566,13 +566,21 @@ class Resolve(object):
         return sorted(dists, key=self.version_key)
 
     @staticmethod
-    def ms_to_v(ms):
+    def to_ms_clause_id(ms):
         ms = MatchSpec(ms)
         return '@s@' + ms.spec + ('?' if ms.optional else '')
 
+    @staticmethod
+    def to_notd_ms_clause_id(ms):
+        return '!@s@'+ms.name
+
+    @staticmethod
+    def to_feature_metric_id(dist, feat):
+        return '@fm@%s@%s' % (dist, feat)
+
     def push_MatchSpec(self, C, ms):
         ms = MatchSpec(ms)
-        name = self.ms_to_v(ms)
+        name = self.to_ms_clause_id(ms)
         m = C.from_name(name)
         if m is not None:
             return name
@@ -588,7 +596,7 @@ class Resolve(object):
         if m is None:
             libs = [dist.full_name for dist in libs]
             if ms.optional:
-                libs.append('!@s@'+ms.name)
+                libs.append(self.to_notd_ms_clause_id(ms))
             m = C.Any(libs)
         C.name_var(m, name)
         return name
@@ -601,7 +609,7 @@ class Resolve(object):
             for fkey in group:
                 C.new_var(fkey)
             # Create one variable for the group
-            m = C.new_var(self.ms_to_v(name))
+            m = C.new_var(self.to_ms_clause_id(name))
             # Exactly one of the package variables, OR
             # the negation of the group variable, is true
             C.Require(C.ExactlyOne, group + [C.Not(m)])
@@ -633,12 +641,12 @@ class Resolve(object):
             dist_feats = {dist.full_name: self.features(dist) for dist in group}
             active_feats = set.union(*dist_feats.values()).intersection(self.trackers)
             for feat in active_feats:
-                feat_spec = self.push_MatchSpec(C, '@' + feat)
+                clause_id_for_feature = self.push_MatchSpec(C, '@' + feat)
                 for dist, features in dist_feats.items():
                     if feat not in features:
-                        mname = '@fm@{}@{}'.format(dist, feat)
-                        C.name_var(C.And(dist, feat_spec), mname)
-                        eq[mname] = 1
+                        feature_metric_id = self.to_feature_metric_id(dist, feat)
+                        C.name_var(C.And(dist, clause_id_for_feature), feature_metric_id)
+                        eq[feature_metric_id] = 1
         return eq
 
     def generate_removal_count(self, C, specs):
@@ -938,7 +946,12 @@ class Resolve(object):
             log.debug('Track feature count: %d', obj1)
 
             # Featured packages: minimize number of featureless packages
-            # installed when a featured alternative is feasible
+            # installed when a featured alternative is feasible.
+            # For example, package name foo exists with two built packages. One with
+            # 'track_features: 'feat1', and one with 'track_features': 'feat2'.
+            # The previous "Track features" minimization pass has chosen 'feat1' for the
+            # environment, but not 'feat2'. In this case, the 'feat2' version of foo is
+            # considered "featureless."
             eq_feature_metric = r2.generate_feature_metric(C)
             solution, obj2 = C.minimize(eq_feature_metric, solution)
             log.debug('Package misfeature count: %d', obj2)
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -624,20 +624,22 @@ class Resolve(object):
 
     def generate_feature_metric(self, C):
         eq = {}  # a C.minimize() objective: Dict[varname, coeff]
-        total = 0
+        # Given a pair (dist, feature), assign a "1" score IF:
+        # - The dist is installed
+        # - The dist does NOT require the feature
+        # - At least one package in the group DOES require the feature
+        # - A package that tracks the feature is installed
         for name, group in iteritems(self.groups):
-            nf = [len(self.features(dist)) for dist in group]
-            maxf = max(nf)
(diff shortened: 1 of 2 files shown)


mcg1969:
That said: I have an idea for another approach to the `generate_feature_metric` method; @mbargull you may want to try this. The idea is this. Right now the two feature optimization passes are:

1. Minimize the number of active features activated.
2. Maximize the number of packages installed that contain an active feature.

The problem has been with step 2 there, and how it has been implemented. Indeed, as I've stated it here, it's quite obvious why it's causing the problems @SylvainCorlay is seeing now. But I think a better metric is this:

2. Minimize the number of packages installed with _feature mismatch_. That is: the package _has_ a featured alternative, but it was not installed for other reasons (say, dependency conflicts).

Here is the code I've written to generate this optimization metric. @mbargull it doesn't break the test I created for PR #6766. And I _believe_ this will resolve the behavior we're seeing here, because it's not doing anything to _favor_ the installation of a package you didn't ask for.

mcg1969:
If you're reading this and wondering why conda should be _minimizing_ the number of installed features---note that if you _install_ a package that contains a `track_feature`, then it is "stuck" and won't be minimized away. So really, this is just the way for `conda` to try to ensure you don't get a feature you didn't ask for.

mcg1969:
@mbargull can you try this out on the scenarios that you have been most keenly interested?

mbargull:
> Right now the two feature optimization passes are: \[...\]
>
> 2. Maximize the number of packages installed that contain an active feature.
>
>\[...\] But I think a better metric is this:
>
> 2. Minimize the number of packages installed with feature mismatch. That is: the package has a featured alternative, but it was not installed for other reasons (say, dependency conflicts).

Yes, that makes much more sense, and sounds more in line with the intended use of features (as I came to understand them, at least ;).
I'll try it out in a couple of hours and will also take a look at the code. Thank you for providing an update on this so quickly!

conda/resolve.py
```diff
@@ -624,20 +624,22 @@ def generate_update_count(self, C, specs):
 
     def generate_feature_metric(self, C):
         eq = {}  # a C.minimize() objective: Dict[varname, coeff]
-        total = 0
+        # Given a pair (dist, feature), assign a "1" score IF:
+        # - The dist is installed
+        # - The dist does NOT require the feature
+        # - At least one package in the group DOES require the feature
+        # - A package that tracks the feature is installed
```
mbargull: Thanks for including this comment! That's very helpful to countercheck the code against.

mcg1969:
Thanks so much for the review @mbargull. I am relieved to hear it seems to cover both regimes of difficulty.

mbargull:
Me too. Thanks for this.

kalefranz:
We should ensure this PR also resolves https://github.com/ContinuumIO/anaconda-issues/issues/8471

mbargull:
It does: https://github.com/ContinuumIO/anaconda-issues/issues/8471#issuecomment-364137796

kalefranz:
I've updated this PR with @mbargull's requested code comments, as well as expanding on some variable names to hopefully make the code more self-documenting.

@mcg1969 Are you fine removing the WIP tag and having this merged at this point?

mcg1969:
Well, I do feel like we could benefit from some key tests. Like for instance #8471 ...

kalefranz:
Don't disagree with that.  Who's going to write them?

mbargull:
```python
# in test_resolve.py

def test_surplus_features():
    index = {
        'feature-1.0-0.tar.bz2': IndexRecord(**{
            'name': 'feature',
            'version': '1.0',
            'build': '0',
            'build_number': 0,
            'track_features': 'feature',
        }),
        'package1-1.0-0.tar.bz2': IndexRecord(**{
            'name': 'package1',
            'version': '1.0',
            'build': '0',
            'build_number': 0,
            'features': 'feature',
        }),
        'package2-1.0-0.tar.bz2': IndexRecord(**{
            'name': 'package2',
            'version': '1.0',
            'build': '0',
            'build_number': 0,
            'depends': ['package1'],
            'features': 'feature',
        }),
        'package2-2.0-0.tar.bz2': IndexRecord(**{
            'name': 'package2',
            'version': '2.0',
            'build': '0',
            'build_number': 0,
            'features': 'feature',
        }),
    }
    r = Resolve({Dist(key): value for key, value in iteritems(index)})
    install = r.install(['package2', 'feature'])
    assert 'package1' not in set(d.name for d in install)
```

mbargull:
Oh, just noticed now, that I actually wanted
```diff
-        'package2-2.0-0.tar.bz2': IndexRecord(**{
+        'package2-1.0-1.tar.bz2': IndexRecord(**{
             'name': 'package2',
-            'version': '2.0',
-            'build': '0',
-            'build_number': 0,
+            'version': '1.0',
+            'build': '1',
+            'build_number': 1,
             'features': 'feature',
         }),
```
in that test case, i.e., make the decision dependent on build number and dependency only.
I'm a little surprised that even with the version number increase that test failed without the changes from this PR... @mcg1969, should we use the test as committed or rather like in the diff above? Either one fails without this PR and passes with it.

kalefranz:
No reason not to include both.

kalefranz:
@mcg1969 Are you good to go on this?

kalefranz:
@mcg1969 We're bumping up against the last few hours of what we can sneak into conda for the next Anaconda and matching Miniconda.  If we want this in, we have to be moving on it.

mbargull:
A `Miniconda3 4.3.34` with this would be nice, indeed (esp. since Bioconda/conda-forge still use `4.3.x` and I assume this could likely be the last `Miniconda 4.3.x`, right?).

mcg1969:
Understood. Thanks for the tremendous help on this, both of you.

mingwandroid:
Once AD 5.1 comes out with conda 4.4 it would be weird for us to make a Miniconda with 4.3.

I guess we *could* make an exception here for you guys but what's holding up conda 4.4 adoption?

Pinging @csoja and @nehaljwani.

mbargull:
> Once AD 5.1 comes out with conda 4.4

Ah, okay, wasn't aware/sure of that.

> but what's holding up conda 4.4 adoption?

Good question. Can anyone from @conda-forge/core (`@conda-forge/core`) or @bioconda/core (`@bioconda/core`) comment on why we shouldn't update to 4.4?
IMHO, 4.4's teething troubles have been worked out in the patch releases, so from me there won't be any objections.

BTW: Can anyone confirm whether team mentions work outside of an organization? At least they are never *rendered* on GitHub... Either way, I'll put the word out on Gitter just to be sure.

bgruening:
Nothing against 4.4 from my side. Thanks for all the work you all have put into this.

kalefranz:
I’ve got a PR against conda-build-all that’s still failing tests for some reason. Thought conda 4.4.9 fixed it, but apparently not. That’s the biggest road block right now I think.

kalefranz:
@mbargull if you have the bandwidth, take a look at https://github.com/conda-tools/conda-build-all/pull/96 and see if you can spot what’s going on in those failing tests. 

I thought I had fixed the issue with https://github.com/conda/conda/pull/6808, but guess I was wrong.

mbargull:
Just opened the Travis log for that a couple of seconds ago 😉

isuruf:
No issues with conda=4.4 for conda-forge. We'll just downgrade to conda=4.3, until conda-forge is ready.

Issue #6811, opened 2018-01-29
IsADirectoryError


pvanderlinden:
**I'm submitting a...**
  - [X] bug report

## Current Behavior
When installing certain packages it will break with IsADirectoryError
```
<<HTTPS 200 OK
< Accept-Ranges: bytes
< Content-Disposition: attachment; filename="six-1.11.0-py35h423b573_1.tar.bz2"; filename*=UTF-8''six-1.11.0-py35h423b573_1.tar.bz2
< Content-Type: application/x-tar
< Date: Mon, 29 Jan 2018 11:43:42 GMT
< ETag: "576d38abfcb7ceb3b323b802b1fec665"
< Last-Modified: Sat, 21 Oct 2017 18:55:29 GMT
< Server: AmazonS3
< x-amz-id-2: IGiKj9lk7QKmYRYyZCVaUEM4Ax19T3slC8e1+ZVXTR3gT5Hg9UNd/8CxoS9yDWZt5PdfORpg9mo=
< x-amz-request-id: 8A668C11FD79A607
< Content-Length: 21870
< Elapsed: 00:00.106578

TRACE conda.core.path_actions:execute(1238): extracting /opt/conda/pkgs/six-1.11.0-py35h423b573_1.tar.bz2 => /opt/conda/pkgs/six-1.11.0-py35h423b573_1
TRACE conda.gateways.disk.update:rename(60): renaming /opt/conda/pkgs/six-1.11.0-py35h423b573_1 => /opt/conda/pkgs/six-1.11.0-py35h423b573_1.c~
DEBUG conda.gateways.disk.create:extract_tarball(132): extracting /opt/conda/pkgs/six-1.11.0-py35h423b573_1.tar.bz2
  to /opt/conda/pkgs/six-1.11.0-py35h423b573_1
TRACE conda.gateways.disk.create:write_as_json_to_file(61): writing json to file /opt/conda/pkgs/six-1.11.0-py35h423b573_1/info/repodata_record.json
TRACE conda.gateways.disk.create:write_as_json_to_file(61): writing json to file /opt/conda/pkgs/six-1.11.0-py35h423b573_1/info/repodata_record.json
TRACE conda.gateways.disk.delete:rm_rf(29): rm_rf /opt/conda/pkgs/six-1.11.0-py35h423b573_1.tar.bz2.c~
TRACE conda.gateways.disk.delete:rm_rf(29): rm_rf /opt/conda/pkgs/six-1.11.0-py35h423b573_1.c~
[...]

PR #6867, opened 2018-02-12
Don't attempt copy/remove fallback for rename failures

nehaljwani:
Drop custom copy + remove fallback for rename failures

`os.rename()` returns `EXDEV` for directories and `EINVAL` for files in a docker
container with `overlayfs` or `aufs` storage backend.

xref: https://github.com/moby/moby/issues/25409#issuecomment-238537855

The fallback for this was to use copy + remove, but it assumed that the rename
failed for a file. So, whenever a copy attempt was made on a directory, it
cascaded into an uncaught `EISDIR`.

This patch delegates the copy+remove logic to `shutil.move()` which works for
directories and files and saves us from reinventing the wheel.

### DIFF
--- a/conda/gateways/disk/update.py
+++ b/conda/gateways/disk/update.py
@@ -7,10 +7,11 @@ import os
 from os import rename as os_rename, utime
 from os.path import dirname, isdir
 import re
+from shutil import move
 
 from . import exp_backoff_fn, mkdir_p, mkdir_p_sudo_safe
 from .delete import rm_rf
-from .link import islink, lexists
+from .link import lexists
 from ...common.compat import on_win
 from ...common.path import expand
 from ...exceptions import NotWritableError
@@ -44,15 +45,6 @@ def update_file_in_place_as_binary(file_full_path, callback):
             fh.close()
 
 
-def _copy_then_remove(source_path, destination_path):
-    from .create import copy, create_hard_link_or_copy
-    if islink(source_path):
-        copy(source_path, destination_path)
-    else:
-        create_hard_link_or_copy(source_path, destination_path)
-    rm_rf(source_path)
-
-
 def rename(source_path, destination_path, force=False):
     if lexists(destination_path) and force:
         rm_rf(destination_path)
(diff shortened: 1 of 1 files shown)


conda/gateways/disk/update.py
```diff
@@ -62,6 +62,11 @@ def rename(source_path, destination_path, force=False):
             os_rename(source_path, destination_path)
         except EnvironmentError as e:
             if e.errno in (EINVAL, EXDEV):
+                # https://github.com/conda/conda/issues/6811
+                # Don't attempt fallback for directories, let the caller catch EXDEV, see:
+                # https://github.com/conda/conda/blob/4.4.9/conda/core/path_actions.py#L1243-L1254
+                if e.errno == EXDEV and isdir(source_path):
+                    raise
```
nehaljwani: To be honest, I think we should just get rid of this fallback and add EINVAL to https://github.com/conda/conda/blob/4.4.9/conda/core/path_actions.py#L1246

But I am not sure of rename is being called only from that place or not.

nehaljwani: cc @quasiben

kalefranz: `rename()` is called 7 times through `backoff_rename()` in `conda/core/path_actions.py`. It looks like that's currently the full use of it.

kalefranz: When using `backoff_rename()` in a `reverse()` method, I don't think we have to worry about any special handling for docker filesystem issues.

Use of `backoff_rename()` in `execute()` methods at:

* https://github.com/conda/conda/blob/4.4.9/conda/core/path_actions.py#L974
* https://github.com/conda/conda/blob/4.4.9/conda/core/path_actions.py#L1151
* https://github.com/conda/conda/blob/4.4.9/conda/core/path_actions.py#L1244

kalefranz: > I think we should just get rid of this fallback completely and add EINVAL to 

I'm ok with raising the error handling up to the caller of `rename()`/`backoff_rename()` for now.  @nehaljwani Why don't you go ahead and see what that looks like.  Just make sure to add notes in top-line comments for `rename()` and `backoff_rename()` about the errors future users of those functions might want to think about handling.

kalefranz:
Wasn't there an issue this was meant to address? I thought so, but I'm not finding it...

nehaljwani:
@kalefranz https://github.com/conda/conda/issues/6811

kalefranz:
@nehaljwani Do you think we should backport this to 4.3.x?

nehaljwani:
We could, yes.

PR #6881, opened 2018-02-14
fix #6880 build_number comparison not functional in match_spec

### LINKED ISSUE #6880: build_number comparison not functional in match_spec
Version comparisons work.

    conda search 'python[version="<3"]'

Seems that build_number comparisons do not.

    conda search 'python[build_number="<5"]'

kalefranz:
fix #6880

There's a lot of surface area to cover here for testing the 4.4 MatchSpec implementation.  We haven't gotten it all yet.

### DIFF
--- a/tests/models/test_match_spec.py
+++ b/tests/models/test_match_spec.py
@@ -395,6 +395,24 @@ class MatchSpecTests(TestCase):
         assert c.match(DPkg(dst, track_features='nomkl'))
         assert c.match(DPkg(dst, track_features='blas=nomkl debug'))
 
+    def test_bracket_matches(self):
+        record = {
+            'name': 'numpy',
+            'version': '1.11.0',
+            'build': 'py34_7',
+            'build_number': 7,
+        }
+
+        assert MatchSpec("numpy<2").match(record)
+        assert MatchSpec("numpy[version<2]").match(record)
+        assert not MatchSpec("numpy>2").match(record)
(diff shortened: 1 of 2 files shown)


mbargull:
There isn't a way to express conjunctions or disjunctions with multiple `MatchSpec`s like with version, à la `>=1,<2|>3`, right? I haven't really though about (practical) feasibility in terms of solver clauses and such, just curious.
The use case I had in mind was something like
```
python[version=3.6.3, build_number=">=5"] OR python[version=">=3.6.4"] 
```
, i.e., putting a lower bound on the build number for older versions.

kalefranz:
Nah haven't implemented anything like that yet.  I guess it would just be sort of a compound matchspec.

Also one thing we need to attempt at some point, I guess related, is *merging* and being able to determine if one matchspec fully encapsulates another.  That's the problem we're bumping into in https://github.com/conda/conda/pull/6807

tests/models/test_match_spec.py
```diff
@@ -642,3 +660,47 @@ def test_parse_parens(self):
...
+        assert _parse_spec_str("python[build_number='<3']") == {
+            "name": "python",
+            "build_number": '<3',
+        }
+        assert _parse_spec_str("python[build_number='<=3']") == {
+            "name": "python",
+            "build_number": '<=3',
+        }
+
+        # # these don't work right now, should they?
```
jjhelmus: It would be nice to support these un-quoted but I wouldn't make them a high priority.

PR #6885, opened 2018-02-15
Fix #6816 fish environment autocomplete

### LINKED ISSUE #6816: fish autocompletion of available environments breaks with no environments
**I'm submitting a...**
  - [X] bug report
  - [ ] feature request

## Current Behavior

If a user doesn't have any environments in `$_CONDA_ROOT/envs/`, fish shell complains about having nothing to expand.

### Steps to Reproduce

With no environments in `$_CONDA_ROOT/envs/`:
```
conda activate <tab>
```

## Expected Behavior

Not only should it not puke when there's no environments available, but it should also probably check `$HOME/.conda/envs/` for user environments for completion.

I wrote the following to replace `__fish_conda_envs` in `$_CONDA_ROOT/etc/fish/conf.d`:
```
function __fish_conda_envs
  eval $_CONDA_EXE env list | tail -n +3 | awk '{print $1}'
end
```

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : None
            shell level : 0
       user config file : /home/e330665/.condarc
 populated config files : /opt/anaconda3/.condarc
          conda version : 4.4.8
    conda-build version : 3.1.6
         python version : 3.6.3.final.0
       base environment : /opt/anaconda3  (read only)
           channel URLs : https://repo.continuum.io/pkgs/main/linux-64
                          https://repo.continuum.io/pkgs/free/linux-64
                          https://repo.continuum.io/pkgs/r/linux-64
                          https://repo.continuum.io/pkgs/pro/linux-64
          package cache : /opt/anaconda3/pkgs
                          /home/e330665/.conda/pkgs
       envs directories : /home/e330665/.conda/envs
                          /opt/anaconda3/envs
               platform : linux-64
             user-agent : conda/4.4.8 requests/2.18.4 CPython/3.6.3 Linux/4.14.14-1-ARCH arch/rolling glibc/2.26
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```
Can't see a reason to post this
```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

```
Nor this
```
</p></details>

edwardreed81:
Uses conda env to get a list on available environments.

### DIFF
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -93,7 +93,7 @@ function __fish_conda_commands
 end
 
 function __fish_conda_envs
-  eval $_CONDA_EXE env list | tail -n +3 | awk '{print $1}'
+  conda config --show envs_dirs | awk 'NR>1{print }' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
 end
 
 function __fish_conda_packages
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -93,7 +93,7 @@ function __fish_conda_commands
 end
 
 function __fish_conda_envs
-  basename -a $_CONDA_ROOT/envs/*/
+  eval $_CONDA_EXE env list | tail -n +3 | awk '{print $1}'
 end
 
 function __fish_conda_packages
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -93,7 +93,7 @@ function __fish_conda_commands
 end
 
 function __fish_conda_envs
-  conda config --show envs_dirs | awk 'NR>1{print $2}' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
+  conda config --show envs_dirs | awk 'NR > 1 {print $2}' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
 end
 
 function __fish_conda_packages
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -93,11 +93,11 @@ function __fish_conda_commands
 end
 
 function __fish_conda_envs
-  conda config --show envs_dirs | awk 'NR>1{print }' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
+  conda config --show envs_dirs | awk 'NR>1{print $2}' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
 end
 
 function __fish_conda_packages
-  command conda list | awk 'NR > 2 {print $1}'
+  conda list | awk 'NR > 3 {print $1}'
 end
 
 function __fish_conda_needs_command


kalefranz:
There's a speed trade-off here.  `conda env list` is slow.  @jaimergp I think we've went back and forth on this before.  Are you ok with this change?

jaimergp:
Is there any reason to _not_ use the current approach? Is there any other location where the envs might reside? The difference in speed is noticeable:

```
$> time conda env list #(x3)
0.82user 0.07system 0:00.90elapsed 100%CPU (0avgtext+0avgdata 43552maxresident)k
0inputs+0outputs (0major+56036minor)pagefaults 0swaps

0.84user 0.07system 0:00.91elapsed 100%CPU (0avgtext+0avgdata 43628maxresident)k
0inputs+0outputs (0major+55964minor)pagefaults 0swaps

0.83user 0.08system 0:00.91elapsed 100%CPU (0avgtext+0avgdata 43828maxresident)k
0inputs+0outputs (0major+55974minor)pagefaults 0swaps

$> time conda info --envs

0.43user 0.02system 0:00.46elapsed 100%CPU (0avgtext+0avgdata 43744maxresident)k
0inputs+0outputs (0major+26374minor)pagefaults 0swaps

0.43user 0.03system 0:00.47elapsed 100%CPU (0avgtext+0avgdata 43668maxresident)k
0inputs+0outputs (0major+26393minor)pagefaults 0swaps

0.42user 0.03system 0:00.46elapsed 100%CPU (0avgtext+0avgdata 43708maxresident)k
0inputs+0outputs (0major+26363minor)pagefaults 0swaps

$> time basename -a $CONDA_PREFIX/envs/*/

0.00user 0.00system 0:00.00elapsed ?%CPU (0avgtext+0avgdata 1828maxresident)k
0inputs+0outputs (0major+73minor)pagefaults 0swaps

0.00user 0.00system 0:00.00elapsed ?%CPU (0avgtext+0avgdata 1904maxresident)k
0inputs+0outputs (0major+74minor)pagefaults 0swaps

0.00user 0.00system 0:00.00elapsed 100%CPU (0avgtext+0avgdata 1848maxresident)k
0inputs+0outputs (0major+74minor)pagefaults 0swaps
```

kalefranz:
> Is there any reason to *not* use the current approach?

There is.  Conda has for a long time recorded non-named environments in a `~/.conda/environments.txt` file. So when a user does `conda create -p /some/random/place/on/disk`, conda can keep track of that environment.

Also,

```
$ conda config --show envs_dirs
envs_dirs:
  - /Users/kfranz/miniconda/envs
  - /Users/kfranz/.conda/envs
```

Conda always has at least two envs_dirs available by default, one of them located in the user home directory, because in many deployments the base environments `envs` directory is not user-writable.  The `envs_dirs` parameter is also configurable.

`conda env list` and `conda info --envs` (they should be equivalent) take care of all of these details.

As a compromise, using something like

    conda config --show envs_dirs | tail -n +2 | cut -d' ' -f4

might get you most of the way there, and be performant enough.

edwardreed81:
The original bug was that if a deployment doesn't have any environments in `$_CONDA_ROOT/envs/`, fish complains about no matches in the wildcard expansion:

    fish: No matches for wildcard “/opt/anaconda3/envs/*/”. See `help expand`.

So just getting a list of possible directories where environments can exist doesn't actually fix the bug because you'll still get no matches for the wildcard.

kalefranz:
> So just getting a list of possible directories where environments can exist doesn't actually fix the bug because you'll still get no matches for the wildcard.

Those are the directories you need to list, in a loop, then use `basename` on.

jaimergp:
Ok, so what about this:

`conda config --show envs_dirs | awk 'NR>1{print $2}' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort`

It runs in 0.1s in my system.

edwardreed81:
@kalefranz 

> Those are the directories you need to list, in a loop, then use basename on.

Isn't basename being called on all directories under those directories? That's the `/*/` portion. The problem is that `*` - if there's nothing in there, fish throws an error when it tries to expand it.

@jaimergp 
That works on my system as well.

jaimergp:
@edwardreed81, @kalefranz 

I have edited my `conda.fish` with my proposal and seems fast enough for interactive use. If it works for you too, I am fine with this approach :)

kalefranz:
Seems like a winner to me.

edwardreed81:
I just edited mine as well and it is pretty smooth. Thanks @jaimergp. Should I modify this commit/PR or how do you want to handle it? This is actually my first PR on any project ever, so this has been a nice experience, thanks guys.

kalefranz:
In that case thanks for contributing! This PR is already targeting the 4.4.x branch, so that’s perfect. Just go ahead and modify this PR.

kalefranz:
> we could use this commit to fix that as well.

Of course!

conda/shell/etc/fish/conf.d/conda.fish
```diff
@@ -93,7 +93,7 @@ function __fish_conda_commands
 end
 
 function __fish_conda_envs
-  eval $_CONDA_EXE env list | tail -n +3 | awk '{print $1}'
+  conda config --show envs_dirs | awk 'NR>1{print }' | xargs -IX find X -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort
```
jaimergp: It should be `awk 'NR > 1 {print $2}'`!

edwardreed81: Does the whitespace matter? It doesn't seem to on my system.

jaimergp: It's not the whitespace, but the `$2` part.

jaimergp: The whitespace is only for consistency with the command below (`__fish_conda_packages`) :)

edwardreed81: oh nevermind, I missed the `$2` somehow!

jaimergp:
LGTM!

mbargull:
I'm -1 for the `conda config` + `find` workaround. We could achieve essentially the same speedup by using conditional imports and adjusting the execution order (i.e., avoid unneeded operations) in `conda.cli.main_info`. From preliminary testing, this would make `conda info --envs` about as fast as that workaround.

kalefranz:
Some of these non-core CLI commands--like `search`, `info`, and `list`--have pretty much all their logic in the CLI wrapper layer.  Basically equivalent to putting business logic and direct db access in your view layer.  Anyway I don't at all disagree that there's a lot of refactoring and cleanup to be done here.

kalefranz:
@mbargull I'm inclined to merge this in the near-term.  That's not to say we can't improve it again with further refactors.  The PR as it is improves on what's there right now without creating any functional downside.  So, I guess what I'm saying is that I think we can have both.  First this, then any improvements when we get to them.

mbargull:
Then I'd like to suggest to add a `TODO` comment on this which states we should remove the workaround and work on `main_info` instead at some point -- or open another issue that tracks speeding up `conda info --envs`.

PR #6893, opened 2018-02-15
channel matchspec fixes

kalefranz:
Fixes included in #6799, which is being closed.

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -6,7 +6,7 @@ from collections import Mapping
 from os.path import basename
 import re
 
-from .channel import Channel, MultiChannel
+from .channel import Channel
 from .dist import Dist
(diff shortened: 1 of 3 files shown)


conda/models/match_spec.py
```diff
@@ -778,11 +778,14 @@ def match(self, other):
         if self._re_match:
             return self._re_match(_other_val.canonical_name)
         else:
-            return self._raw_value.canonical_name == _other_val.canonical_name
+            # assert ChannelMatch('pkgs/free').match('defaults') is False
+            # assert ChannelMatch('defaults').match('pkgs/free') is True
```
jjhelmus: Do you want to retain these comments?

kalefranz: Yeah, I do.  Just to help clarify in this section of code the order of channel <=> multi-channel matching.

PR #6913, opened 2018-02-22
Add install instructions for fish as default shell

JarnoRFB:
If fish is the default shell [CONDA_INSTALL_LOCATION]/bin has to be in $PATH because the variable is not inherited from bash. This is make explicit in the install instructions now.


kalefranz:
The advice here isn't correct.  You shouldn't be manually adjusting path anymore for conda.  Yes, you'll probably want to add something like

    source [CONDA_INSTALL_LOCATION]/etc/fish/conf.d/conda.fish
    conda activate base

to `~/.config/config.fish`.  That's probably more in line with what the instructions should say.

kalefranz:
I'm going to close this PR, only because (for some reason) a bunch of other commits got mixed into it that don't belong.  Feel free to open a new PR that incorporates the changes I talked about.

JarnoRFB:
I am afraid I'm not really getting it. I right now have conda 4.3.31. Are there changes that affect the behavior in fish significantly on master right now? Because if I just put 


    source [CONDA_INSTALL_LOCATION]/etc/fish/conf.d/conda.fish

without adding
    
    set -gx PATH /home/jarno/.miniconda/bin $PATH

before in `config.fish`. I get the following error with fish as *default* shell

```
conda: command not found
~/.miniconda/etc/fish/conf.d/conda.fish (line 13): 
                command conda $argv
                        ^
in function “conda”
	called on line 165 of file ~/.miniconda/etc/fish/conf.d/conda.fish
	with parameter list “..activate fish py34”

in command substitution
	called on line 163 of file ~/.miniconda/etc/fish/conf.d/conda.fish

in function “activate”
	called on standard input
	with parameter list “py34”
```


when setting the path explicitly everything works fine. I am on Ubuntu 17.10 by the way.

kalefranz:
Yeah we're about to release conda 4.4.11 and conda 4.5.0.  conda-forge is behind...

JarnoRFB:
Alright, it works now correctly after updating conda. I added a PR with the changed instructions here #6944. Thanks for the support!

PR #6922, opened 2018-02-22
resolve #5610 add PrefixData, SubdirData, and PackageCacheData to conda/api.py

### LINKED ISSUE #5610: Add PrefixData, SubdirData, and PackageCacheData to conda/api.py
https://github.com/conda/conda/commit/69f1b7af939adb783ee284fa952174e8fa447569#diff-aa516f421733d7851d878e08d4bf7464R408

This is used in conda-build at https://github.com/conda/conda-build/blob/master/conda_build/render.py#L210

Please provide an interface (ideally in conda's exports) that I can use to ask Conda to download packages for me.

kalefranz:
resolve #5610

### DIFF
--- a/conda/api.py
+++ b/conda/api.py
@@ -15,6 +15,8 @@ DepsModifier = _DepsModifier
 
 class Solver(object):
     """
+    **Beta**
+
     A high-level API to conda's solving logic. Three public methods are provided to access a
     solution in various forms.
 
@@ -26,6 +28,8 @@ class Solver(object):
 
     def __init__(self, prefix, channels, subdirs=(), specs_to_add=(), specs_to_remove=()):
         """
+        **Beta**
+
         Args:
             prefix (str):
                 The conda prefix / environment location for which the :class:`Solver`
@@ -44,7 +48,10 @@ class Solver(object):
 
     def solve_final_state(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NULL,
                           force_remove=NULL):
-        """Gives the final, solved state of the environment.
+        """
+        **Beta**
+
+        Gives the final, solved state of the environment.
 
         Args:
             deps_modifier (DepsModifier):
@@ -56,6 +63,7 @@ class Solver(object):
                     * ONLY_DEPS
                     * UPDATE_DEPS
                     * UPDATE_DEPS_ONLY_DEPS
+                    * FREEZE_INSTALLED
             prune (bool):
                 If ``True``, the solution will not contain packages that were
                 previously brought into the environment as dependencies but are no longer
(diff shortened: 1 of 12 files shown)


conda/api.py
```diff
@@ -1,5 +1,110 @@
...
+from .core.prefix_data import PrefixData as _PrefixData
+from .core.solve import Solver as _Solver, DepsModifier as _DepsModifier
+from .core.subdir_data import SubdirData as _SubdirData
+from .models.channel import Channel
+
+
+DepsModifier = _DepsModifier
+
+
+class Solver(object):
```
goanpeca: We need a docstring here, that also explains why this `Solver` wrapper is needed, or intended for?

conda/api.py
```diff
@@ -1,5 +1,110 @@
...
+                                             force_remove, force_reinstall)
+
+    def solve_for_transaction(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NULL,
+                              force_remove=NULL, force_reinstall=False):
+        return self._internal.solve_for_transaction(deps_modifier, prune, ignore_pinned,
+                                                    force_remove, force_reinstall)
+
+
+class SubdirData(object):
+
```
goanpeca: We need a docstring here, that also explains why this `Solver` wrapper is needed, or intended for?

conda/api.py
```diff
@@ -1,5 +1,110 @@
...
+    def iter_records(self):
+        return self._internal.iter_records()
+
+    def reload(self):
+        self._internal = self._internal.reload()
+        return self
+
+
+class PackageCacheData(object):
+
```
goanpeca: same here

conda/api.py
kalefranz: Any recommendation on how to do this without duplicating everything that's already in the underlying `Solver` class?  Plenty of nice docstrings here: https://github.com/conda/conda/blob/master/conda/core/solve.py#L52

kalefranz:
If we want this incorporated into conda 4.5, we might have to move on it without docstrings.  Conda 4.5.0 needs to be in canary tomorrow.  If we don't mind bumping it to conda 4.6, then we can add docstrings.

conda/api.py
```diff
@@ -1,5 +1,110 @@
...
+from .models.channel import Channel
+
+
+DepsModifier = _DepsModifier
+
+
+class Solver(object):
+
+    def __init__(self, prefix, channels, subdirs=(), specs_to_add=(), specs_to_remove=()):
+        self._internal = _Solver(prefix, channels, subdirs, specs_to_add, specs_to_remove)
```
goanpeca: This will only work for interactive scripting but we could do at least


```python
    self.solve_final_state.__doc__ = self._internal.solve_final_state.__doc__
```

?

kalefranz:
Ok, no solid examples yet, but pretty solid documentation.  Is that a reasonable compromise to get this into 4.5.0?

kalefranz:
I think we should add "beta" tags to every docstring in `conda/api.py` right now. I'd like @msarahan to start really using this with conda-build to see how it works out.  There might need to be some tweaks once it's in use and we gain further information.

kalefranz:
@msarahan has done a quick review.  We discussed needing to add a way to add individual packages to repodata.  That's beyond the scope of this PR, as it'll take some work to the underlying `SubdirData` class.  That'll be done in a future PR.

kalefranz:
merging based on approval by @goanpeca

PR #6923, opened 2018-02-22
 resolve #6810 add CONDA_EXE environment variable to activate

### LINKED ISSUE #6810: Conda (4.4): 'No such file or directory' in a new environment using non-shell process
**I'm submitting a...** (I don't know whether it is a bug or feature)
  - [x] bug report
  - [x] feature request

### Steps to Reproduce

Create an environment, 'test' using Conda 4.4, open a new tab and directly activate that environment. Now try to invoke conda commands using any non shell program (lua/python/java/....) it will error out saying `conda: command not found` or `No such file or directory` since they can't `exec()` that.

## Expected Behavior

The program invoking conda should not error out, or should have a way to figure out where the conda  program resides, so that it can exec the full `/path/to/conda`. Currently the only thing close to this is `$_CONDA_EXE`, but it is not documented, and also not exported, so not available to child processes and no guarantee whether it will be available in future releases.

The current work around is to first activate the base environment and then activate the non-base environment in the same shell session, which for whatever reason, retains the root environment's bin directory in the PATH.

kalefranz:
resolve #6810

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -244,6 +244,7 @@ class Activator(object):
             new_path = self.pathsep_join(self._add_prefix_to_path(prefix))
             export_vars = {
                 'CONDA_PYTHON_EXE': self.path_conversion(sys.executable),
+                'CONDA_EXE': self.path_conversion(context.conda_exe),
                 'PATH': new_path,
                 'CONDA_PREFIX': prefix,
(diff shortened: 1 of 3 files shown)


kalefranz:
merging based on review by @goanpeca

PR #6924, opened 2018-02-22
Unify location of record types in conda/models/records.py

kalefranz:
Another refactor in preparation for conda 4.5.  This should be reviewed only after #6922 is merged.


kalefranz:
I'll rebase this PR after we get #6922 finalized and merged.

kalefranz:
Just had to rebase

conda/api.py
```diff
@@ -4,11 +4,10 @@
 from .common.constants import NULL
 from .core.package_cache_data import PackageCacheData as _PackageCacheData
 from .core.prefix_data import PrefixData as _PrefixData
-from .core.solve import Solver as _Solver, DepsModifier as _DepsModifier
+from .core.solve import DepsModifier as _DepsModifier, Solver as _Solver
```
goanpeca: Why do we do 

```
from .core.solve import DepsModifier as _DepsModifier
```

and then

```
DepsModifier = _DepsModifier
```

conda/core/package_cache_data.py
```diff
@@ -28,9 +28,9 @@
                                   read_index_json_from_tarball, read_repodata_json)
 from ..gateways.disk.test import file_path_is_writable
 from ..models.dist import Dist
-from ..models.index_record import PackageRecord, PackageRef
+from ..models.records import PackageRecord, PackageRef
 from ..models.match_spec import MatchSpec
-from ..models.package_cache_record import PackageCacheRecord
+from conda.models.records import PackageCacheRecord
```
goanpeca: Why the mixture of import types??

`conda.models.records` vs `..models.records`

conda/core/path_actions.py
```diff
@@ -36,10 +36,9 @@
 from ..history import History
 from ..models.channel import Channel
 from ..models.enums import LinkType, NoarchType, PathType
-from ..models.index_record import Link, PackageRecord, PathDataV1, PathsData
+from ..models.records import Link, PackageRecord, PathDataV1, PathsData
 from ..models.match_spec import MatchSpec
-from ..models.package_cache_record import PackageCacheRecord
-from ..models.prefix_record import PrefixRecord
+from conda.models.records import PackageCacheRecord, PrefixRecord
```
goanpeca: Why not join imports ?

conda/core/prefix_data.py
```diff
@@ -16,9 +16,9 @@
 from ..gateways.disk.delete import rm_rf
 from ..gateways.disk.test import file_path_is_writable
 from ..models.dist import Dist
-from ..models.index_record import PackageRef
+from ..models.records import PackageRef
 from ..models.match_spec import MatchSpec
-from ..models.prefix_record import PrefixRecord
+from conda.models.records import PrefixRecord
```
goanpeca: same here?

conda/models/dist.py
```diff
@@ -21,6 +21,9 @@
                                          'dist_name'))
 
 
+IndexRecord = PackageRecord  # for conda-build backward compat
```
goanpeca: Is conda build accessing this module directly or should we expose it in the `api` module then?

kalefranz: It should be in exports right now.  That's where conda-build should be accessing it from.  I'm not sure I want to expose it yet as something that can be created through the api.  Maybe; maybe not.

conda/core/prefix_data.py
kalefranz: thank, yeah that's a problem

conda/api.py
kalefranz: Yeah we don't need that.

conda/models/dist.py
goanpeca: Yeah exports module, so should we add this renaming there...?

conda/api.py
kalefranz: Actually, did that to avoid the reassignment error thing.

goanpeca: Yeah but I mean why not import it directly as 

`from .core.solve import DepsModifier` ?

goanpeca: @kalefranz?

conda/models/dist.py
goanpeca: @kalefranz ?

conda/api.py
kalefranz: Because then you need a line

    DepsModifier = DepsModifier

conda/models/dist.py
kalefranz: We can't for a while.

https://github.com/conda/conda-build/blob/3.5.1/conda_build/conda_interface.py#L67

kalefranz:
Merging following review by @goanpeca

PR #6929, opened 2018-02-23
fix #6904 logic errors in prefix graph data structure

### LINKED ISSUE #6904: Conda 4.4.x does not prevent removal of conda dependencies from base environment?
Some user tried to remove `pycosat` from within Navigator using conda 4.4.8 
 as reported  on [this issue](https://github.com/ContinuumIO/anaconda-issues/issues/8563)

It seems conda 4.4.x is allowing user to remove conda dependencies like `pycosat` which effectively leaves them in a broken state. I checked with 4.4.9 and 4.4.10 and it seems to be that this is still possible.

I have tested on Mac, @jjhelmus tested on Linux. Need to check on windows

kalefranz:
fix #6904 

- [X] make it work
- [X] make it right
- [ ] make it fast

We probably have some suboptimal duplication of work, but at least this is a step forward.  That said, I doubt Amdahl would particularly have a problem here.


conda/models/prefix_graph.py
```diff
@@ -0,0 +1,387 @@
...
+                # add menuinst as a parent if python is a parent and the node
+                # isn't a parent of menuinst
+                assert python_node is not None
+                menuinst_parents = graph[menuinst_node]
+                for node, parents in iteritems(graph):
+                    if python_node in parents and node not in menuinst_parents:
+                        parents.add(menuinst_node)
+
+            if conda_node:
+                # add conda as a parent if python is a parent and node isn't a parent of conda
```
kalefranz: This comment is wrong.

conda/models/prefix_graph.py
```diff
@@ -0,0 +1,387 @@
...
+        if node not in graph:
+            raise KeyError('node %s does not exist' % node)
+        graph.pop(node)
+        self.spec_matches.pop(node, None)
+
+        for node, edges in iteritems(graph):
+            if node in edges:
+                edges.remove(node)
+
+    def _toposort(self):
```
kalefranz: This toposort is a pretty strict re-implementation of the toposort we already use.  It's reimplementation because we're using `PackageRecord`s as nodes, rather than simple strings.  For reference, the original toposort is at `conda/common/toposort.py`.

conda/models/prefix_graph.py
```diff
@@ -0,0 +1,387 @@
...
+
+    The terminology used for edge direction is "parents" and "children" rather than "successors"
+    and "predecessors". The parent nodes of a record are those records in the graph that
+    match the record's "depends" field.  E.g. NodeA depends on NodeB, then NodeA is a child
+    of NodeB, and NodeB is a parent of NodeA.  Nodes can have zero parents, or more than two
+    parents.
+
+    Most public methods mutate the graph.
+    """
+
```
kalefranz: The potential inefficiencies I referred to before are

1. Probably some unnecessary calls to toposort
2. ad-hoc creation of `inverted_graph` in several methods

conda/models/prefix_graph.py
```diff
@@ -0,0 +1,387 @@
...
+            if conda_node:
+                # add conda as a parent if python is a parent and node isn't a parent of conda
+                conda_parents = graph[conda_node]
+                for node, parents in iteritems(graph):
+                    if (hasattr(node, 'noarch') and node.noarch == NoarchType.python
+                            and node not in conda_parents):
+                        parents.add(conda_node)
+
+
```
kalefranz: This stuff is left over from the old `dag.py` module and should be incorporated later.

tests/models/test_prefix_graph.py
```diff
@@ -0,0 +1,740 @@
+# -*- coding: utf-8 -*-
```
kalefranz: Lots of tests.  GitHub wanted to hide this file.

mbargull:
Regarding
> - [ ] make it fast

Do you have an example that induces heavy computations on the graph data structure? Something with `anaconda`-sized package lists or larger?
I'm only halfway through reviewing this (that is, only the actual code, completely disregarding the tests). Didn't saw major flaws, but also haven't completely comprehended the usage yet. Just saw a bunch of (likely) duplicate computations and thus wondered about possible impact on the practical runtime..

kalefranz:
Yeah some of the methods just require the data structure to be in one direction, and others require it to be inverted.  Without profiling and more thought, I don't know if it's better to keep the equivalent of a doubly linked list--maintain both directions--or to just generate the inverted graph on demand when we need it.  I guess we could write up a series of integration tests that use `--dry-run`.  That'd probably be the easiest way to put load on here.  I don't want to put *that* much data in the repository for unit tests.

mbargull:
> I don't want to put that much data in the repository for unit tests.

Yeah, I agree. I was more thinking about an example to test locally :wink:

PR #6909, opened 2018-02-21
fix #6899 handle Unicode output in activate commands

mbargull:
See: https://github.com/conda/conda/issues/6739#issuecomment-362941051

PR #6951, opened 2018-02-27
set encoding for PY2 stdout/stderr

mbargull:
Followup to #6909. I completely misread
```python
  File "/opt/conda/lib/python3.6/codecs.py", line 377, in write
    self.stream.write(data)
TypeError: string argument expected, got 'bytes'
```
before and though `self.encode` failed instead of `self.stream.write`.
So, easy fix is to apply the IO encoding only for the case that is needed, i.e., `PY2 and isinstance(stream, file)`.
(To make sure everything is in order, I added an "inverted" test that fails for Python 3 and passed for Python 2. Once CI shows exactly that result, I'll update the test accordingly.)

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -129,10 +129,7 @@ class Activator(object):
     def _finalize(self, commands, ext):
         commands = concatv(commands, ('',))  # add terminating newline
         if ext is None:
-            output = self.command_join.join(commands)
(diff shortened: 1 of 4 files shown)


mbargull:
If there won't be a `4.4.12`, I can retarget to `master`.

PR #6956, opened 2018-02-28
MatchSpec merge

kalefranz:
reference: https://github.com/conda/conda-build/pull/2694

### DIFF
--- a/conda/models/channel.py
+++ b/conda/models/channel.py
@@ -368,7 +368,7 @@ def _get_channel_for_name(channel_name):
         # stripping off path threw information away from channel_name (i.e. any potential subname)
         # channel.name *should still be* channel_name
         channel = copy(channel)
-        channel.name = channel_name
+        channel.name = _stripped
         if platform:
             channel.platform = platform
         return channel
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -408,19 +408,12 @@ class MatchSpec(object):
 
     @classmethod
     def merge(cls, match_specs):
-        match_specs = tuple(match_specs)
-        for match_spec in match_specs:
-            assert isinstance(match_spec, cls), match_spec
-
+        match_specs = tuple(cls(s) for s in match_specs)
         grouped = groupby(lambda spec: spec.get_exact_value('name'), match_specs)
-
-        specs_map = {}
-        for name, specs in iteritems(grouped):
-            if len(specs) > 1:
-                specs_map[name] = reduce(lambda x, y: x._merge(y), specs)
-            else:
-                specs_map[name] = specs[0]
-
+        specs_map = {
+            name: reduce(lambda x, y: x._merge(y), specs) if len(specs) > 1 else specs[0]
+            for name, specs in iteritems(grouped)
+        }
         return tuple(itervalues(specs_map))
 
     def _merge(self, other):
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -20,9 +20,9 @@ from ..common.url import is_url, path_to_url, unquote
 from ..exceptions import CondaValueError
 
 try:
-    from cytoolz.itertoolz import concat, groupby
+    from cytoolz.itertoolz import concat, concatv, groupby
 except ImportError:  # pragma: no cover
-    from .._vendor.toolz.itertoolz import concat, groupby  # NOQA
+    from .._vendor.toolz.itertoolz import concat, concatv, groupby  # NOQA
 
 
 class MatchSpecType(type):
@@ -410,11 +410,12 @@ class MatchSpec(object):
     def merge(cls, match_specs):
         match_specs = tuple(cls(s) for s in match_specs)
         grouped = groupby(lambda spec: spec.get_exact_value('name'), match_specs)
+        dont_merge_these = grouped.pop('*', []) + grouped.pop(None, [])
         specs_map = {
             name: reduce(lambda x, y: x._merge(y), specs) if len(specs) > 1 else specs[0]
             for name, specs in iteritems(grouped)
         }
-        return tuple(itervalues(specs_map))
+        return tuple(concatv(itervalues(specs_map), dont_merge_these))
 
     def _merge(self, other):
         if self.optional != other.optional or self.target != other.target:
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from abc import ABCMeta, abstractmethod, abstractproperty
 from collections import Mapping
+from functools import reduce
 from os.path import basename
 import re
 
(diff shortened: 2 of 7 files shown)


tests/models/test_match_spec.py
```diff
@@ -717,3 +717,36 @@ def test_parse_build_number_brackets(self):
...
+
+class MatchSpecMergeTests(TestCase):
+
+    def test_merge_version(self):
+        specs = (MatchSpec('exact'), MatchSpec('exact 1.2.3 1'), MatchSpec('exact >1.0,<2'))
+        merged_specs = MatchSpec.merge(specs)
+        print(merged_specs)
+        assert len(merged_specs) == 1
+        print(merged_specs[0])
+        assert str(merged_specs[0]) == "exact[version='>1.0,<2',build=1]"
```
msarahan: does this merged spec even make sense?  build numbers should only be for an exact match.  Furthermore, an exact version number (one with no relational operators and no *) should always be higher in priority than an inexact version number.

This test loses the two different kinds of behavior:
1. exact has priority over inexact
2. within inexact, there's an order that results in the tightest constraint.  Max of lower bounds, min of upper bounds.

kalefranz: > build numbers

I'm questioning whether that third position in the conda-build style match spec string is build or build number.  I guess it has to be build.  In which case, yes, exact match only.  In conda 4.4, build numbers support comparison operators (see #6880 and #6881).

> does this merged spec even make sense?

I think you're right.  This was just an initial commit to start chatting with @mcg1969 about this.  Very early WIP!

kalefranz: @msarahan Small tweak in the last commit.  This is how @mcg1969 suggested combining these.  Any thoughts?

tests/models/test_match_spec.py
```diff
@@ -717,3 +717,60 @@ def test_parse_build_number_brackets(self):
...
+            'bounded >=1.0,<2.0', 'bounded >=1.5', 'bounded <=1.8',
+        ))
+        merged_specs = MatchSpec.merge(specs)
+        print(merged_specs)
+        assert len(merged_specs) == 2
+
+        exact_spec = next(s for s in merged_specs if s.name == 'exact')
+        bounded_spec = next(s for s in merged_specs if s.name == 'bounded')
+
+        assert str(exact_spec) == "exact[version='1.2.3,>1.0,<2',build=1]"
```
msarahan: This is fine.  It's more constraints than it needs to be.  My code cut them out, but I expect this is safer than my code (i.e. less likely to lose the finer details of when < and <= are not equivalent), and less prone to logic errors.  It's a little confusing looking at it, because so much of the information there is superfluous - you have to think about what bound or pin will take effect, rather than just seeing it.  That said, as long as the performance impact of having more constraints will scale well, I'm happy enough with how this is.

tests/models/test_match_spec.py
```diff
@@ -717,3 +717,60 @@ def test_parse_build_number_brackets(self):
         #     "build_number": '>=3',
         # }
 
+
+class MatchSpecMergeTests(TestCase):
+
+    def test_merge_single_name(self):
+        specs = (MatchSpec('exact'), MatchSpec('exact 1.2.3 1'), MatchSpec('exact >1.0,<2'))
```
mbargull: Can you also add
```python
specs = (MatchSpec('exact 1.2.3 1'), MatchSpec('exact 1.2.3 2'))
with pytest.raises(ValueError):
    MatchSpec.merge(specs)
```
so that we have an "expected failure" test?

conda/models/match_spec.py
```diff
@@ -404,6 +406,36 @@ def fn(self):
         assert val
         return val
 
+    @classmethod
+    def merge(cls, match_specs):
+        match_specs = tuple(cls(s) for s in match_specs)
+        grouped = groupby(lambda spec: spec.get_exact_value('name'), match_specs)
+        specs_map = {
+            name: reduce(lambda x, y: x._merge(y), specs) if len(specs) > 1 else specs[0]
```
mbargull:
```python
name: reduce(lambda x, y: x._merge(y), specs)
```
should suffice as `reduce` on a one-element sequence should return `specs[0]`.

conda/models/match_spec.py
```diff
@@ -404,6 +406,36 @@ def fn(self):
...
+            this_component = self._match_components.get(component_name)
+            that_component = other._match_components.get(component_name)
+            if this_component is None and that_component is None:
+                continue
+            elif this_component is None:
+                final_components[component_name] = that_component
+            elif that_component is None:
+                final_components[component_name] = this_component
+            else:
+                final_components[component_name] = this_component.merge(that_component)
```
mbargull: You could simply this as
```python
final_components = {}
for name, component in iteritems(self.match_components):
    if component is None:
        continue
    final_components[name] = component
for name, component in iteritems(other.match_components):
    if component is None:
        continue
    if name in final_components:
        final_components[name].merge(component)
    else:
        final_components[name] = component
```
or
```python
final_components = {name: comp for name, comp in iteritems(self._match_components) if comp is not None}
for name, component in ((name: comp) for name, comp in iteritems(other._match_components) if comp is not None):
    if name in final_components:
        final_components[name].merge(component)
    else:
        final_components[name] = component
```

conda/models/match_spec.py
```diff
@@ -404,6 +406,36 @@ def fn(self):
         assert val
         return val
 
+    @classmethod
+    def merge(cls, match_specs):
+        match_specs = tuple(cls(s) for s in match_specs)
+        grouped = groupby(lambda spec: spec.get_exact_value('name'), match_specs)
```
mbargull: Didn't check myself, but is the `match_specs` input always sorted by `spec.get_exact_value('name')`? `groupby` groups by consecutive keys only.

mbargull: Ah, disregard that comment, you used `(cy)toolz.itertoolz.groupby` not `itertools.groupby`, so everything is fine!

tests/models/test_match_spec.py
```diff
@@ -717,3 +717,60 @@ def test_parse_build_number_brackets(self):
...
+            'name': 'exact',
+            'version': '1.2.2',
+            'build': '1',
+            'build_number': 1,
+        })
+
+    def test_merge_multiple_name(self):
+        specs = tuple(MatchSpec(s) for s in (
+            'exact', 'exact 1.2.3 1', 'exact >1.0,<2',
+            'bounded >=1.0,<2.0', 'bounded >=1.5', 'bounded <=1.8',
```
mbargull: How about intermixing the specs to make sure order doesn't matter? E.g.:
```python
            'exact', 'exact 1.2.3 1',
            'bounded >=1.0,<2.0', 'bounded >=1.5', 'bounded <=1.8',
            'exact >1.0,<2',
```

conda/models/match_spec.py
kalefranz: I plugged both of these in.  They both have four test failures with the additional tests I added after the most recent commit.

tests/models/test_match_spec.py
kalefranz: done

tests/models/test_match_spec.py
kalefranz: Done.  Plus a bunch more tests added.

conda/models/channel.py
```diff
@@ -368,7 +368,7 @@ def _get_channel_for_name_helper(name):
         # stripping off path threw information away from channel_name (i.e. any potential subname)
         # channel.name *should still be* channel_name
         channel = copy(channel)
-        channel.name = channel_name
+        channel.name = _stripped
```
kalefranz: This one-line fix could be a separate PR. But oh well. It's at least in its own commit.  The commit has a test, which is included in `tests/models/test_channel.py`.

tests/models/test_match_spec.py
kalefranz: Yeah this is what @mcg1969 suggested, exactly because of the logical complexity of reducing to a minimal form.  I think this is the solution we should go with for now, with the goal of implementing the minimal-form reduction in the future.

conda/models/match_spec.py
mbargull: Writing code directly in GitHub is a **bad** habit...
Sorry, forgot the assignment in
```python
final_components[name] = final_components[name].merge(component)
```
Edited the comment above with that fix.

msarahan:
Merging to facilitate testing for 4.5.0rc release.  Thanks for your comments @mbargull - I hope we'll keep addressing them in the release.

PR #6976, opened 2018-03-02
Performance enhancement for Dist objects

groutr:
Remove key function and replace with key attribute.

Since the key is immutable, we don't need a function call to produce it.
Building the key once at object creation avoids a lot of new tuple
objects and function calls.  I also store the hash of the key since
tuples recursively calculate their hash values each time.

In creating a basic data science environment, these changes shave about 1.7s off the total runtime for conda.
```
time python main.py create -n testdev python numpy scipy pandas jupyter --dry-run
dist-key2: 13.32s user 0.20s system 93% cpu 14.521 total

master: 15.33s user 0.21s system 95% cpu 16.223 total
```
The improvement is more noticeable with more packages.
```
time python main.py create -n testdev python numpy scipy pandas jupyter pytables bokeh cython dask datashader holoviews seaborn --dry-run
dist-key2: 30.51s user 0.29s system 96% cpu 32.031 total

master: 35.55s user 0.44s system 96% cpu 37.267 total
```

### DIFF
--- a/conda/models/dist.py
+++ b/conda/models/dist.py
@@ -73,7 +73,6 @@ class Dist(Entity):
                                    platform=platform)
 
         self.__key = (channel, dist_name)
-        self.__key_hash = hash(self.__key)
 
     def to_package_ref(self):
         return PackageRef(
@@ -244,7 +243,7 @@ class Dist(Entity):
         return isinstance(other, self.__class__) and self.__key >= other.__key
 
     def __hash__(self):
-        return self.__key_hash
+        return hash(self.__key)
 
     def __eq__(self, other):
         return isinstance(other, self.__class__) and self.__key == other.__key
--- a/conda/models/dist.py
+++ b/conda/models/dist.py
@@ -72,6 +72,9 @@ class Dist(Entity):
                                    base_url=base_url,
                                    platform=platform)
 
+        self.__key = (channel, dist_name)
+        self.__key_hash = hash(self.__key)
+
     def to_package_ref(self):
         return PackageRef(
             channel=self.channel,
@@ -228,33 +231,26 @@ class Dist(Entity):
                 if self.platform
                 else join_url(self.base_url, filename))
 
-    def __key__(self):
-        return self.channel, self.dist_name
-
     def __lt__(self, other):
-        assert isinstance(other, self.__class__)
-        return self.__key__() < other.__key__()
+        return isinstance(other, self.__class__) and self.__key < other.__key
 
     def __gt__(self, other):
-        assert isinstance(other, self.__class__)
-        return self.__key__() > other.__key__()
+        return isinstance(other, self.__class__) and self.__key > other.__key
 
     def __le__(self, other):
-        assert isinstance(other, self.__class__)
-        return self.__key__() <= other.__key__()
+        return isinstance(other, self.__class__) and self.__key <= other.__key
 
     def __ge__(self, other):
-        assert isinstance(other, self.__class__)
-        return self.__key__() >= other.__key__()
+        return isinstance(other, self.__class__) and self.__key >= other.__key
 
     def __hash__(self):
-        return hash(self.__key__())
+        return self.__key_hash
 
     def __eq__(self, other):
-        return isinstance(other, self.__class__) and self.__key__() == other.__key__()
+        return isinstance(other, self.__class__) and self.__key == other.__key
 
     def __ne__(self, other):
-        return not self.__eq__(other)
+        return not self == other
 
     # ############ conda-build compatibility ################
 


groutr:
Looks like the conda-build tests are passing now except for some missing R package from CRAN (r-cran-nmf)

mbargull:
What's the motive to remove the `assert`s? I'm no big fan of inducing potential silent fails..
@kalefranz: What was the motive that no `assert` is used for `__eq__`?

groutr:
@mbargull Running/compiling Python code in any optimized mode will disable assert statements, therefore asserts should never be used for control flow. Assertions are debug level statements for checking assumptions in a chunk of code.  We shouldn't assume that Python is always running our code in what is effectively a debug mode.  Using assert statements could introduce the silent fail behavior you wanted to avoid.

As for the silent fail potential, a better way would probably be to return NotImplemented if the type check fails.

groutr:
I should also explain for the benefit of all why we can't cache the has of self.__key.  What I observed is that the hash was changing between different instances of Dist objects.  For a given package in the solution, the dist object in the solution had a different hash than the dist object in the index.

In one of the failing conda-build tests, we had this situation.  The first line is the object in the solution and the second line is the corresponding object in the index.  The three fields are the objects hash, the contents of the object's key, and the object's id.
```
                   hash                contents of key                    id
Solution: -8194553981133064073 conda-forge ca-certificates-2018.1.18-0 139627647241296
Index: -8752704017545300206 conda-forge ca-certificates-2018.1.18-0 139627677492168

6310709686116748962 conda-forge certifi-2018.1.18-py36_0 139627647241464
4339292024634228224 conda-forge certifi-2018.1.18-py36_0 139627678240160

8068237682851260134 conda-forge ncurses-5.9-10 139627647241128
7380129252822756334 conda-forge ncurses-5.9-10 139627678464600

806033304550791492 conda-forge openssl-1.0.2n-0 139627647241240
-5172741414867245551 conda-forge openssl-1.0.2n-0 139627678617104

6783883222000868768 conda-forge pip-9.0.1-py36_1 139627647239784
4375778942271413089 conda-forge pip-9.0.1-py36_1 139627683574056

-4066167169479619783 conda-forge python-3.6.4-0 139627647239560
5076903470494095772 conda-forge python-3.6.4-0 139627681752008

8612007355177735106 conda-forge readline-7.0-0 139627647240848
-5141768007290159154 conda-forge readline-7.0-0 139627679659568

-1909622333390710995 conda-forge setuptools-38.5.1-py36_0 139627647239448
2847656242649401031 conda-forge setuptools-38.5.1-py36_0 139627681856480

24015384924622500 conda-forge sqlite-3.20.1-2 139627647240736
3893619110724713884 conda-forge sqlite-3.20.1-2 139627717402296

-504215723807823378 conda-forge tk-8.6.7-0 139627647240568
-8546783321497732552 conda-forge tk-8.6.7-0 139627695720936

1287821086695946909 conda-forge wheel-0.30.0-py36_2 139627647240064
2618240198229800662 conda-forge wheel-0.30.0-py36_2 139627678202848

-6611820260125506438 conda-forge xz-5.2.3-0 139627647240344
-8305524842936266159 conda-forge xz-5.2.3-0 139627680903296

-6725206302785262792 conda-forge zlib-1.2.11-0 139627647240232
-4831967795482785717 conda-forge zlib-1.2.11-0 139627698195872
```

mbargull:
> @mbargull Running/compiling Python code in any optimized mode will disable assert statements, therefore asserts should never be used for control flow. Assertions are debug level statements for checking assumptions in a chunk of code. 

Yeah, I know and I'm also very inclined to rap over people's knuckles if they dare to incorporate `assert` into control flow -- and it's actually one major thing that I dislike about defaults and conventions on how (C)Python is used. I wish `python -OO` were the default, static type checking would become more common, etc. /minor (agreeing) rant over
I assumed (i.e., haven't checked all uses), that those `assert`s were used as they should (i.e., as you depicted, not in control flow and not expected failures) from what I then assumed that the comparisons are never done with non-`Dist` objects. And thus the existing `isinstance` for `__eq__` didn't make sense to me.
(And rather than throwing `NotImplemented`/`TypeError`, I'd make more sense to look review the `Dist` comparison usages -- but I guess, investing time in #6979  and further work makes even more sense :wink:)

----
Thanks for the additional information and investigation regarding the hashes; interesting, I wonder how that happens (something pickled/unpickled? would that make an influence? strange.. ).

groutr:
@mbargull, I'm glad that we agree about assert statements.
I see what you mean by the silent behavior.  I'm in the process of checking the various versions of these comparison methods to see how they affect the test suite.  I can also gather information on if the type check fails (and what types attempt to be compared with Dist objects).

I haven't seen any code that explicitly compares Dist objects, and the fields that are in the key lead me to believe that the comparisons are probably only used for sorting purposes.  What meaning can you ascribe to the comparison of 'conda-forge' < 'defaults' or comparing the string representation of the dist?

At the end of the day, I agree, if Dist is to go away, then more effort on this PR probably isn't warranted.  I'd much rather put the additional effort into #6979.

As for the hashes, I suspect that somewhere an object is using the default `__hash__` which relies on the object's id.

travishathaway:

PR #6988, opened 2018-03-05
circleci: switch conda-build 2.1 tests to 2.1.x

mbargull:
Switches the `conda-build 2.1` tests from the `2.1.17` tag to the `2.1.x` branch so that `2.1` tests work again.
xref: https://github.com/conda/conda-build/pull/2726

PR #6990, opened 2018-03-05
circleci: skip conda-build test r-cran-nmf

mbargull:
Replaces #6988. Didn't notice the existing `conda-build` test skips before...

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -65,6 +65,7 @@ conda_build_test: &conda_build_test
       not xattr
       and not skeleton_pypi
       and not perl-cpan-Moo
+      and not r-cran-nmf
       and not test_expand_globs
       and not test_build_expands_wildcards
       and not numpy_build_run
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -65,6 +65,7 @@ conda_build_test: &conda_build_test
       not xattr
       and not skeleton_pypi
       and not perl-cpan-Moo
+      and not r-cran-nmf
       and not test_expand_globs
       and not test_build_expands_wildcards
       and not numpy_build_run
```
kalefranz: I think this whole `CONDABUILD_SKIP ` block should be removed.  Doesn't it get trampled by what's below in every case it's used anyway?

kalefranz:
What the heck is going on with the conda-build 2.1 tests?

    Exit code 5: No tests were collected

Weird.

circle.yml
mbargull: Yeah, looks that way. I'll remove it.

kalefranz:
So it looks like `and not cran-nmf` is fine, but `and not r-cran-nmf` causes the no tests found situation. I have no idea why.

kalefranz:
superseded by #7065

PR #7000, opened 2018-03-07
Fix #6996: change fish prompt only if changeps1 is true

### LINKED ISSUE #6996: `conda.fish` does not respect `changeps1` config setting
**I'm submitting a...**
  - [x] bug report
  - [ ] feature request

## Current Behavior

The `conda.fish` file does not respect the `changeps1` [config setting](https://conda.io/docs/user-guide/configuration/use-condarc.html?highlight=changeps1#change-command-prompt-changeps1). In particular his function:

https://github.com/conda/conda/blob/a88a7b065ff4c00712286e099e4cebd199ebf6b6/conda/shell/etc/fish/conf.d/conda.fish#L34

seems a bit too happy about always displaying the environment.

This is annoying when your shell theme also display the environment (see e.g. oh-my-fish/theme-agnoster#28).

I'm also a bit puzzle by the lack of prompt changing in the corresponding Bash source file. (Though perhaps that is handled somewhere else?)

## Environment Information
<details open><summary><code>`conda info`</code>(excerpt)</summary><p>

```
     active environment : None
            shell level : 0
          conda version : 4.4.11
    conda-build version : 3.7.0
       base environment : /usr/local/var/pyenv/versions/miniconda3-latest  (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/osx-64
                          https://repo.continuum.io/pkgs/free/osx-64
                          https://repo.continuum.io/pkgs/r/osx-64
                          https://repo.continuum.io/pkgs/pro/osx-64
               platform : osx-64
             user-agent : conda/4.4.11 requests/2.18.4 CPython/3.6.4 Darwin/17.4.0 OSX/10.13.3
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```

teake:
Fixes #6996.

`changeps1` is respected for posix shells, but not yet for fish.

Note that this adds a slight overhead on instantiating a fish shell, because the added conda call is expensive. This could be made lazy if we want shells that start snappy.


teake:
Come to think of it, it is probably cleaner to export an environment variable (e.g. `CONDA_CHANGE_FISHPROMPT`) in `conda.activate.Activator` and then check for this in the fish source file. That way you’d only enter `conda` during an activate, instead of also during sourcing `conda.fish`. You’d expose an additional environment variable however.

I can change the PR, but I would appreciate some feedback first.

njalerikson:
In my opinion the correct thing to do is to add the bulk of the logic into Activator, not `conda.fish`. You would actually be removing almost all of the prompt related functions from `conda.fish`.

This may need some redesign in Activator to allow for the `_update_prompt` function to support fish prompts but this would be a much cleaner solution.

kalefranz:
There's already an environment variable wired up for you to use here.  `CONDA_PROMPT_MODIFIER`

See e.g.

https://github.com/conda/conda/blob/5835aa7bc5761e35d5c975691c74e2d2cd1d9fbd/conda/activate.py#L252

and

https://github.com/conda/conda/blob/5835aa7bc5761e35d5c975691c74e2d2cd1d9fbd/conda/activate.py#L489-L490

teake:
Ok, how's this?

kalefranz:
Well that version doesn't keep the coloring for the prompt then, as was there before.  I would go back to the way you had it before, except instead of using a boolean check on `conda config --show changeps1`, just do a boolean check on the `CONDA_PROMPT_MODIFIER` environment variable.  If it's set and has length >=1, then show the conda addition to the prompt as before, otherwise don't add the conda prompt.

kalefranz:
The green prompt is what fish users had prior to the conda 4.4 release.  There's no need to cause a regression there for fish users, especially if there's no way to get it back.

I just pushed a commit as https://github.com/conda/conda/pull/7047, which was intended to give users more flexibility over their prompt.  However I hadn't been able to actually get it to work.

kalefranz:
So... Given #7047 could be ready to go, that could affect what we do here...

teake:
Yes, #7047 will break this PR in its current state. If the former gets merged I'll update this one.

Btw, I'm not sure #7047 is the way to go. You're introducing quite a bit of machinery into conda that I think would be left better to the shells themselves (and their respective themes). I'd prefer a simple default prompt that can be disabled, plus properly documented exported environment variables (which then can be picked up by the shell or shell theme). Because then I don't have to have my prompt config in two different places (whatever I"m using for theming now + the conda config file). Besides, #7047 won't work for fish.

Ok, that's it for my (unsolicited) 2 cents. I'm sure #7047 will drive adoption of complex conda prompts at a faster rate; I'm just a bit worried it will hamper long-term integration with shell managers and themes.

mbargull:
I disagree apart from one aspect.
> You're introducing quite a bit of machinery into conda that I think would be left better to the shells themselves (and their respective themes). I'd prefer a simple default prompt that can be disabled

#7047 merely allows a little bit of customization of an existing feature. With that PR it's just that you'd be able to change the prompt prefix from the default `(env) ` to, e.g., `[env] ` or color it, nothing more.
> Besides, #7047 won't work for fish.

With the current code, yes, but I don't think there is no reason why it wouldn't generally. We can let `conda shell.fish activate` emit code that sets `fish_prompt`.

> I'm just a bit worried it will hamper long-term integration with shell managers and themes.

Since #7047 only makes a current functionality a bit more customizable, I see no reason for this claim. Also, changing the prompt can be disabled, see https://github.com/conda/conda/pull/7047#issuecomment-373885640:
> In addition to `changeps1`, if the value of `prompt` is None, we treat it as if `changeps1` is `False`.

----
> plus properly documented exported environment variables (which then can be picked up by the shell or shell theme).

With that I agree.

teake:
Sorry for the late reaction, I've unfortunately had to put this on the back burner up until now.

@kalefranz I've partly made the changes you suggested. The whole prompt modifier will now be colored green instead of only the part within the parentheses, but it will now pick up the content of the `env_prompt` config setting as per #7047. I hope that's a reasonable compromise.

Note that you can only set colors in fish via the `set_color` command, so there's no possibility of color-coding via `env_prompt`, alas.

teake:
@kalefranz bump? Do you need anything more from me on this MR?

sdaitzman:
Bumping this because it's still broken 😢

Issue #3915, opened 2016-11-18
Custom deactivate scripts in etc/conda/deactivate.d/ cannot modify the PATH environment variable


dmaxime:
On Windows with conda 4.2.9, I am attempting to set environment-specific environment variables by following [this documentation](http://conda.pydata.org/docs/using/envs.html#windows).

I've set up the following custom activate/deactivate scripts in the environment folder.
In `etc/conda/activate.d/env_vars.bat`:
```
set R_HOME=C:\Local\Programs\R-3.3.2
set R_USER=C:\Users\ENCPB\Documents\
set R_LIBS_USER=C:\Local\Programs\R_libs
set PATH=C:\Local\Programs\R-3.3.2\bin:%PATH%
```

In `etc/conda/deactivate.d/env_vars.bat`:
```
set R_HOME=
set R_USER=
set R_LIBS_USER=
set PATH=%PATH:C:\Local\Programs\R-3.3.2\bin:=%
```

`R_HOME`, `R_USER`, `R_LIBS_USER` are properly set and unset, but not `PATH`. It is properly set but the deactivate script leaves `C:\Local\Programs\R-3.3.2\bin` in the PATH.

In the global Anaconda `Scripts\deactivate.bat` script, the `PATH` environment variable is actually set **after** the custom `etc/conda/deactivate.d/` scripts are run. Any change custom scripts do to `PATH` is ignored.

njalerikson:
Well first of all in the Windows and CMD.EXE world the `PATH` is `;` delimited not `:` delimited.

So in your `etc\conda\activate.d\env_var.bat`:

```
set PATH=C:\Local\Programs\R-3.3.2\bin;%PATH%
```

And in your `etc\conda\deactivate.d\env_var.bat`:

```
set PATH=%PATH:C:\Local\Programs\R-3.3.2\bin;=%
```

dmaxime:
True, sorry for the confusion.. 
I have corrected both `env_var.bat` scripts with your suggestions and the behavior remains identical: 
- `R_HOME`, `R_USER`, `R_LIBS_USER` are properly set and unset
- `C:\Local\Programs\R-3.3.2\bin` is not removed from `PATH` 

I have been able to correct that behavior by making the global Anaconda `Scripts\deactivate.bat` script execute the environment-specific `etc/conda/deactivate.d/` scripts at the very beginning (before the `@SETLOCAL` command. I am not sure, though, if this could trigger any undesirered side-effect. 

[...]

Issue #6919, opened 2018-02-22
csh prompt handling & deactivate broken


jrevans:
**I'm submitting a...**
  - [x] bug report
  - [ ] feature request

## Current Behavior

In a csh shell, after sourcing the '~/miniconda3/etc/profile.d/conda.csh'

When activating the environment the original prompt is not a part of the new prompt.

Neither is it reverted to the original prompt when deactivating.

Additionally, when deactivating the 'CONDA' specific env vars are not removed from the environment.

### Steps to Reproduce

```
$> echo "'$prompt'"
'$> '
$> conda activate test_env
(test_env) echo "'$prompt'"
'(test_env) '
(test_env) conda deactivate
env | grep CONDA
_CONDA_EXE=/home/jrevans/miniconda3/bin/conda
_CONDA_ROOT=/home/jrevans/miniconda3
CONDA_SHLVL=0
CONDA_DEFAULT_ENV=test_env
CONDA_PREFIX=/home/jrevans/miniconda3/envs/test_env
CONDA_PROMPT_MODIFIER=(test_env) 
CONDA_PYTHON_EXE=/home/jrevans/miniconda3/bin/python

```

## Expected Behavior

The prompt needs to be preserved throughout and the conda env vars need to be removed from the environment.

I have attached a version of the '.../etc/profile.d/conda.csh' script that fixes the prompt problem.

The env var problem can be fixed by changing the results of 'conda shell.csh deactivate' to use 'unsetenv' instead of 'unset'

[conda.csh.txt](https://github.com/conda/conda/files/1748844/conda.csh.txt)

njalerikson:
Resolved with #7041

Issue #6877, opened 2018-02-14
handle non-zero exit codes in conda.csh


kalefranz:
See https://github.com/conda/conda/issues/6785#issuecomment-360560807

kalefranz:
Good first issue for someone familiar with csh and/or tcsh.

njalerikson:
resolved via #7042

Issue #6595, opened 2017-12-28
Clean up activate.py using proper OO inheritence


kalefranz:
See

* https://github.com/conda/conda/pull/6585#issuecomment-354178822
* https://github.com/conda/conda/pull/6585#issuecomment-354303897

kalefranz:
Good first issue for someone comfortable with object oriented programming.

PR #7026, opened 2018-03-13
Two phase deactivate

njalerikson:
This is the PR for #3915.

As discussed there this PR solves the `{de,}activate.d` scripts' issue of being unable to change the `$PATH` by introducing a two phase deactivate.

Two phase because upon the first call, if there are `deactivate.d` scripts then the first phase is to source those scripts and then to do a second conda call - `deactivatepost`. The second phase, the deactivate post, will then perform the final unsetting, prompt modifications, and `$PATH` modification. If however there are no `deactivate.d` scripts then there will only be a single deactivate phase.

This is a relatively dumb fix in my opinion, primarily because I've statically embedded the `deactivatepost` command in the Activator class for each shell. I would love to hear any smarter ideas.

As discussed with @kalefranz, this is not an ideal fix but rather a quick fix until other longterm ideas become reality.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -26,12 +26,18 @@ class _Activator(object):
     # Shells should also use 'reactivate' following conda's install, update, and
     #   remove/uninstall commands.
     #
-    # All core logic is in build_activate() or build_deactivate(), and is independent of
-    # shell type.  Each returns a map containing the keys:
-    #   export_vars
-    #   unset_var
-    #   activate_scripts
-    #   deactivate_scripts
+    # All core logic is in the build_*() functions:
+    #   - build_activate()
+    #   - build_deactivate()
+    #   - build_reactivate()
+    #   - build_post()
+    # and is independent of shell type.  Each returns a map optionally containing the keys:
+    #   unset_vars          (tuple of keys to unset)
+    #   export_vars         (dict of key-values to export)
+    #   set_vars            (dict of key-values to set)
+    #   activate_scripts    (tuple of activate scripts)
+    #   deactivate_scripts  (tuple of deactivate scripts)
+    #   post                (bool to include post command)
     #
     # The value of the CONDA_PROMPT_MODIFIER environment variable holds conda's contribution
     #   to the command prompt.
@@ -87,6 +93,10 @@ class _Activator(object):
         return self._finalize(self._yield_commands(self.build_reactivate()),
                               self.tempfile_extension)
 
+    def post(self):
+        return self._finalize(self._yield_commands(self.build_post()),
+                              self.tempfile_extension)
+
     def execute(self):
         # return value meant to be written to stdout
         self._parse_and_set_args(self._raw_arguments)
@@ -96,7 +106,8 @@ class _Activator(object):
         # the first index of arguments MUST be either activate, deactivate, or reactivate
         if arguments is None:
             from .exceptions import ArgumentError
-            raise ArgumentError("'activate', 'deactivate', or 'reactivate' command must be given")
+            raise ArgumentError("'activate', 'deactivate', 'reactivate', or 'post' command must "
+                                "be given")
 
         command = arguments[0]
         arguments = tuple(drop(self.shift_args + 1, arguments))
@@ -107,13 +118,14 @@ class _Activator(object):
 
         if not command:
             from .exceptions import ArgumentError
-            raise ArgumentError("'activate', 'deactivate', or 'reactivate' command must be given")
+            raise ArgumentError("'activate', 'deactivate', 'reactivate', or 'post' command must "
+                                "be given")
         elif help_requested:
             from . import CondaError
             class Help(CondaError):  # NOQA
                 pass
             raise Help("help requested for %s" % command)
-        elif command not in ('activate', 'deactivate', 'reactivate'):
+        elif command not in ('activate', 'deactivate', 'reactivate', 'post'):
             from .exceptions import ArgumentError
             raise ArgumentError("invalid command '%s'" % command)
         elif command == 'activate' and len(remainder_args) > 1:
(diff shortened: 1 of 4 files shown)


conda/activate.py
```diff
@@ -99,6 +105,8 @@ def __init__(self, shell, arguments=None):
             self.export_var_tmpl = '@SET "%s=%s"'
             self.run_script_tmpl = '@CALL "%s"'
 
+            self.deactivatepost_tmpl = """@CALL %_CONDA_EXE% shell.cmd.exe deactivatepost"""
```
mbargull: For `cmd.exe`, I think  we'd need the monstrosity of
https://github.com/conda/conda/blob/4.5.0rc0/conda/shell/Library/bin/conda.bat#L45
https://github.com/conda/conda/blob/4.5.0rc0/conda/shell/Library/bin/conda.bat#L50-L52
(Maybe with a slightly different variable name for `_TEMP_SCRIPT_PATH` to avoid collision?)

conda/activate.py
```diff
@@ -111,6 +119,8 @@ def __init__(self, shell, arguments=None):
             self.export_var_tmpl = 'set -gx %s "%s"'
             self.run_script_tmpl = 'source "%s"'
 
+            self.deactivatepost_tmpl = """eval $_CONDA_EXE shell.fish deactivatepost"""
```
mbargull: For `fish` we need two `eval`s, acc. to https://github.com/conda/conda/blob/4.5.0rc0/conda/shell/etc/fish/conf.d/conda.fish#L92.

conda/activate.py
```diff
@@ -291,6 +308,28 @@ def build_activate(self, env_name_or_prefix):
 
     def build_deactivate(self):
         # query environment
+        try:
```
mbargull: Better avoid a large `try: ... except` block, i.e.,
```python
try:
    old_conda_prefix = self.environ['CONDA_PREFIX']
except KeyError:
    pass
else:
    ...
```
or (more consistent with the other code in this file)
```python
old_conda_prefix = self.environ['CONDA_PREFIX']
if old_conda_prefix is not None:
    ...
```

conda/activate.py
```diff
@@ -147,6 +157,10 @@ def deactivate(self):
         return self._finalize(self._yield_commands(self.build_deactivate()),
                               self.tempfile_extension)
 
+    def deactivatepost(self):
+        return self._finalize(self._yield_commands(self.build_deactivatepost()),
+                              self.tempfile_extension)
+
     def reactivate(self):
```
mbargull: For `reactivate` we also need a two-staged process if we have `deactivate` scripts.

conda/activate.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, shell, arguments=None):
             self.set_var_tmpl = "%s='%s'"
             self.run_script_tmpl = '\\. "%s"'
 
+            self.deactivatepost_tmpl = """eval "$(PS1="$PS1" $_CONDA_EXE shell.posix deactivatepost)" """
```
mbargull: Use `\eval` to guard against potential `alias`es.

conda/activate.py
```diff
@@ -87,6 +91,8 @@ def __init__(self, shell, arguments=None):
             self.export_var_tmpl = "$%s = '%s'"
             self.run_script_tmpl = 'source "%s"'
 
+            self.deactivatepost_tmpl = """!(@(_CONDA_EXE) shell.xonsh deactivatepost)"""
```
mbargull: Looking at https://github.com/conda/conda/blob/4.5.0rc0/conda/shell/conda.xsh#L24-L38, it seems we need some more code for `xonsh` to properly source the output.

mbargull:
https://github.com/conda/conda/blob/4.5.0rc0/tests/test_activate.py will need to be adjusted accordingly once the code is finalized.

conda/activate.py
njalerikson: wow I didn't even process how ugly the `cmd.exe` version was...

njalerikson:
The fixes thus far has addressed the nix shells, I haven't looked further at the win shells yet.

Yes, the activate tests need to be updated as well.

I've also addressed two other CSH issues (#6919, #6877) since I was already touching the CSH script. The fix to #6919 is courtesy of @jonathaneichelberger.

conda/activate.py
```diff
@@ -23,21 +24,31 @@ class Activator(object):
...
+    #   set_vars            (dict of key-values to set)
+    #   export_vars         (dict of key-values to export)
+    #   unset_vars          (tuple of keys to unset)
+    #   unexport_vars       (tuple of keys to un-export)
+    #   activate_scripts    (tuple of activate scripts)
+    #   deactivate_scripts  (tuple of deactivate scripts)
+    #   deactivatepost      (bool to include deactivate post command)
+    #   reactivatepost      (bool to include reactivate post command)
     #
-    # The value of the CONDA_PROMPT_MODIFIER environment variable holds conda's contribution
```
mbargull: The prevalent (and linter-checked, https://github.com/conda/conda/blob/4.5.0rc0/setup.cfg#L34) style in the code base uses a 99 character line limit. Please don't enforce foreign rules and revert all those changes.

conda/activate.py
```diff
@@ -71,10 +87,16 @@ def __init__(self, shell, arguments=None):
             self.command_join = ';\n'
 
             self.unset_var_tmpl = 'unset %s'
-            self.export_var_tmpl = 'setenv %s "%s"'
+            self.unexport_var_tmpl = 'unsetenv %s'
```
mbargull: I see no reason to introduce `unexport_var_tmpl`. With your changes you effectively just replaced `unset_var_tmpl` with that. The `unset_var_tmpl` is only used to unset environment variables. If you think `unsetenv` should be used for `csh`, then just you that in `unset_var_tmpl`.

conda/activate.py
```diff
@@ -58,10 +69,15 @@ def __init__(self, shell, arguments=None):
             self.command_join = '\n'
 
             self.unset_var_tmpl = '\\unset %s'
-            self.export_var_tmpl = "\\export %s='%s'"
+            self.unexport_var_tmpl = self.unset_var_tmpl
             self.set_var_tmpl = "%s='%s'"
+            self.export_var_tmpl = "\\export %s='%s'"
             self.run_script_tmpl = '\\. "%s"'
 
+            self.post_tmpl = dedent("""\
```
mbargull: Don't overly complicate things, just use
```python
self.post_tmpl = '''eval "$(PS1="${PS1}" "${_CONDA_EXE}" shell.posix %spost)"'''
```

conda/activate.py
```diff
@@ -84,9 +106,18 @@ def __init__(self, shell, arguments=None):
...
 
             self.unset_var_tmpl = 'del $%s'
+            self.unexport_var_tmpl = self.unset_var_tmpl
             self.export_var_tmpl = "$%s = '%s'"
+            self.set_var_tmpl = self.export_var_tmpl
             self.run_script_tmpl = 'source "%s"'
 
+            self.post_tmpl = dedent("""\
+                post_pipeline = !(@(_CONDA_EXE) shell.xonsh %spost)
+                post_stdout = _raise_pipeline_error(post_pipeline)
```
mbargull: `_raise_pipeline_error` is a function defined in https://github.com/conda/conda/blob/4.5.0rc0/conda/shell/conda.xsh#L24-L31. We shouldn't assume it's always available.

conda/activate.py
```diff
@@ -84,9 +106,18 @@ def __init__(self, shell, arguments=None):
             self.command_join = '\n'
 
             self.unset_var_tmpl = 'del $%s'
+            self.unexport_var_tmpl = self.unset_var_tmpl
             self.export_var_tmpl = "$%s = '%s'"
+            self.set_var_tmpl = self.export_var_tmpl
             self.run_script_tmpl = 'source "%s"'
 
+            self.post_tmpl = dedent("""\
```
mbargull: I'm not big fan of the line-continuation characters in Python code.
I'd rather just
```python
self.post_tmpl = dedent("""
...
""".strip("\n"))
```
or
```python
self.post_tmpl = "\n".join((
    "line1",
    "line2",
))
```

conda/activate.py
```diff
@@ -96,9 +127,15 @@ def __init__(self, shell, arguments=None):
             self.command_join = '\r\n' if on_win else '\n'
 
             self.unset_var_tmpl = '@SET %s='
+            self.unexport_var_tmpl = self.unset_var_tmpl
             self.export_var_tmpl = '@SET "%s=%s"'
+            self.set_var_tmpl = self.export_var_tmpl
             self.run_script_tmpl = '@CALL "%s"'
 
+            self.post_tmpl = dedent("""\
+                @CALL %_CONDA_EXE% shell.cmd.exe %spost\
```
mbargull: I know this part is WIP, so just a heads-up: use `%%_CONDA_EXE%%`.

conda/activate.py
```diff
@@ -234,15 +308,18 @@ def build_activate(self, env_name_or_prefix):
             #  i.e. step back down
             return self.build_deactivate()
 
-        activate_scripts = self._get_activate_scripts(prefix)
+        assert 0 <= old_conda_shlvl <= max_shlvl
+
+        args = {}
```
mbargull: I'd rather not have these unrelated `args = {}` changes in this PR.
It maybe would make sense to do that as part of https://github.com/conda/conda/issues/6595.

conda/shell/etc/profile.d/conda.csh
```diff
@@ -4,43 +4,49 @@
 # or in ~/.cshrc add the line
 #   source <CONDA_ROOT>/etc/profile.d/conda.csh
 
-# This block should only be for dev work. Under normal installs, _CONDA_EXE will be templated
-# in at the top of this file.
+# This block should only be for dev work. Under normal installs, _CONDA_EXE
+# will be templated in at the top of this file.
```
mbargull: Like with the Python code: Just retain the 99 character lines.

conda/shell/etc/profile.d/conda.csh
```diff
@@ -4,43 +4,49 @@
...
-            $_CONDA_EXE $argv[1-]
-            eval `$_CONDA_EXE shell.csh reactivate`
+            ${_CONDA_EXE:q} ${argv[1-]:q}
+            set ask_conda="`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh reactivate)`" || exit ${status}
+            eval "${ask_conda}"
             rehash
             breaksw
         default:
-            $_CONDA_EXE $argv[1-]
+            ${_CONDA_EXE:q} ${argv[1-]:q}
```
mbargull: I don't know `csh`, so can't really comment on those changes. I guess it makes sense to put that in a separate PR, so that it's easier for a C-shell-knowing person to review.

conda/shell/etc/profile.d/conda.sh
```diff
@@ -9,24 +9,25 @@ _conda_set_vars() {
     elif [ -n "${POSH_VERSION:+x}" ]; then
         _CONDA_SHELL_FLAVOR=posh
     else
-        # default to dash; if we run into a problem here, please raise an issue
+        # default to dash; if we run into a problem here, please raise an
```
mbargull: ditto, character limit

conda/activate.py
njalerikson: Didn't realize conda wasn't using the 78 char convention.

conda/activate.py
njalerikson: While I agree with this I chose to add it as it makes the solution more distinct in what it's actually unsetting - but alright - I'll revert.

conda/activate.py
njalerikson: Well I would also expect any changes to the individual shell scripts to also make mirrored updates to the applicable `post_tmpl`. This is why I consider this implementation to be relatively dumb.

conda/activate.py
njalerikson: I totally agree, I still haven't quite decided what I prefer. Will reconsider this.

conda/activate.py
njalerikson: It was introduced in part because of the unexport idea but also because I know that the prompt support is currently limited to posix and csh. This change was done for future needs in the update prompt function. But since that's being reverted I guess this will be reverted as well.

conda/shell/etc/profile.d/conda.csh
njalerikson: CSH's `:q` variable modifier properly preserves quoting within the variables. I was trying to do this same trick for the `'${2}' ${argv[3-]}` but due to the sub-shelling it doesn't work (CSH is fun...).

As for another PR to improve review, from my experience with Conda (and just looking at how long CSH issues remain unaddressed in Conda) it will not be easy to find another CSH user.

kalefranz:
I know I've been absent the last week or so.  Been working on some other projects.  I'm generally onboard with this approach now, even though I may have hinted that I wasn't in #3915.

I do prefer that we try to keep PRs as single-purpose as possible.  And I don't mind breaking PRs into multiple, that must be reviewed and merged in a certain order, if one builds on another.

kalefranz:
I'm guessing there are several problems with csh/tcsh support.  I'm glad we're finally looking at it again!

kalefranz:
Want to rebase this one again master now and see where we're at?

njalerikson:
Woohoo! A successful rebase!

njalerikson:
If still desired I'll roll back the incremental minor changes I have introduced and bundle them into a separate PR.

njalerikson:
@kalefranz do we want this to be implemented before or after #6518?

njalerikson:
I realized that actually all three functions (`activate`, `deactivate`, and `reactivate`) all had cases where a `post` function was necessary.

So I back tracked a little and instead of splitting each function into two phases (e.g. `deactivate` and `deactivatepost`) I introduced a new function called `post`. The `post` function relies on a new environment variable `$CONDA_POST` which is a `old_prefix:new_prefix` syntax. This then allows the `post` command to only deal with fixing the `$PATH` after the necessary `deactivate.d` scripts have been called (as well as unsetting `$CONDA_POST`).

Yes I am still missing the `post_tmpl` for `cmd.exe` and `powershell`. And yes I am still missing tests.

njalerikson:
I think I have addressed the issues for Nix shells.

@kalefranz Is this still a PR that is of interest? If so then I'll start hacking on the Powershell and Cmd.exe solutions that are still missing.

njalerikson:
Just realized that Powershell isn't actually supported yet. So I guess I'll keep ignoring that one for now.

beenje:
Anything I could do to help this PR move forward?

beenje:
I'd really like to be able to modify the PATH in activate/deactivate script.
@mbargull is it worth me trying to rebase this PR on master and keep only what is needed?

beenje:
Thanks for the reply @msarahan. I'll give it a try!

beenje:
Looking at this PR, I wonder if it wouldn't be easier to save the original PATH in an environment variable during the activate and always restore it during deactivate.

That wouldn't allow to modify the PATH in the deactivate.d scripts. But I think what we really want is for the PATH to always be the same as before the activate after deactivate (even when the PATH was modified during the activate.d scripts).

This idea didn't come up when discussing #3915. Is that a bad idea?

beenje:
This PR is quite old and the PATH issue has been resolved by https://github.com/conda/conda/pull/8464

My guess is that it should be closed.

msarahan:
@njalerikson I'll close this, but I'm happy to reopen if it still needs to be considered.

PR #7041, opened 2018-03-15
Fix csh prompt

njalerikson:
Splitting #7026 

This specifically addresses #6919

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -227,7 +227,7 @@ class Activator(object):
         old_conda_prefix = self.environ.get('CONDA_PREFIX')
         max_shlvl = context.max_shlvl
 
-        if old_conda_prefix == prefix:
+        if old_conda_prefix == prefix and old_conda_shlvl > 0:
             return self.build_reactivate()
         if self.environ.get('CONDA_PREFIX_%s' % (old_conda_shlvl-1)) == prefix:
             # in this case, user is attempting to activate the previous environment,
@@ -291,9 +291,10 @@ class Activator(object):
 
     def build_deactivate(self):
         # query environment
+        old_conda_prefix = self.environ.get('CONDA_PREFIX')
         old_conda_shlvl = int(self.environ.get('CONDA_SHLVL', 0))
-        old_conda_prefix = self.environ.get('CONDA_PREFIX', None)
-        if old_conda_shlvl <= 0 or old_conda_prefix is None:
+        if not old_conda_prefix or old_conda_shlvl < 1:
+            # no active environment, so cannot deactivate; do nothing
             return {
                 'unset_vars': (),
                 'set_vars': {},
@@ -351,7 +352,7 @@ class Activator(object):
 
     def build_reactivate(self):
         conda_prefix = self.environ.get('CONDA_PREFIX')
-        conda_shlvl = int(self.environ.get('CONDA_SHLVL', -1))
+        conda_shlvl = int(self.environ.get('CONDA_SHLVL', 0))
         if not conda_prefix or conda_shlvl < 1:
             # no active environment, so cannot reactivate; do nothing
             return {
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -70,7 +70,7 @@ class Activator(object):
             self.shift_args = 0
             self.command_join = ';\n'
 
-            self.unset_var_tmpl = 'unset %s'
+            self.unset_var_tmpl = 'unsetenv %s'
             self.export_var_tmpl = 'setenv %s "%s"'
             self.set_var_tmpl = "set %s='%s'"
             self.run_script_tmpl = 'source "%s"'
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ record.txt
 Vagrantfile
 docs/_build
 env/
+.pytest_cache/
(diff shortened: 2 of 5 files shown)


kalefranz:
Good changes.  Just need to modify the failing tests a bit.

kalefranz:
https://github.com/conda/conda/blob/deb16ee5dbfa68ae3433f1cc2766999613914b20/tests/test_activate.py#L1130-L1157

Hmm I wonder if we're skipping those in our CircleCI?

kalefranz:
Probably shouldn't be.

```
(base) circleci@31131a2b540b:~/conda-build$ which csh
/bin/csh
(base) circleci@31131a2b540b:~/conda-build$ which tcsh
/usr/bin/tcsh
```

kalefranz:
Ohhh. 🤦‍♂️

https://github.com/conda/conda/blob/deb16ee5dbfa68ae3433f1cc2766999613914b20/tests/test_activate.py#L1149

kalefranz:
Ammend that last commit to remove `.pytest_cache/v/cache/lastfailed`.  And I guess we need to add `.pytest_cache/` to `.gitignore`.

njalerikson:
Grrrr I got the cache file the first time I ran the tests but then I reran the test...grrrr

conda/activate.py
```diff
@@ -291,9 +291,10 @@ def build_activate(self, env_name_or_prefix):
 
     def build_deactivate(self):
         # query environment
+        old_conda_prefix = self.environ.get('CONDA_PREFIX')
         old_conda_shlvl = int(self.environ.get('CONDA_SHLVL', 0))
-        old_conda_prefix = self.environ.get('CONDA_PREFIX', None)
-        if old_conda_shlvl <= 0 or old_conda_prefix is None:
+        if not old_conda_prefix or old_conda_shlvl < 1:
+            # no active environment, so cannot deactivate; do nothing
```
njalerikson: This was done to mirror the other code blocks (for example in build_reactivate) that check for the same thing.

njalerikson:
@kalefranz hm I am unable to replicate the csh error on any machine I have access too. I have also not been able to figure out what csh is actually in use by CircleCI, do you have that info?

I have however found a csh version on Ubuntu which doesn't support argument passing to sourced scripts - this clearly breaks all functionality, and if i remember correctly we ran into this same issue for dash(?) - not sure what to do with this scenario besides ignoring it.

kalefranz:
This is directly from an interactive shell with circleci

```
circleci@1bb3e3f34db1:~$ which csh
/bin/csh
circleci@1bb3e3f34db1:~$ csh --version
$?tcsh: No match.
circleci@1bb3e3f34db1:~$ ls -al /bin/csh
lrwxrwxrwx 1 root root 21 Dec 16 22:22 /bin/csh -> /etc/alternatives/csh
circleci@1bb3e3f34db1:~$ ls -al /etc/alternatives/csh
lrwxrwxrwx 1 root root 12 Dec 16 22:22 /etc/alternatives/csh -> /bin/bsd-csh
circleci@1bb3e3f34db1:~$ ls -al /bin/bsd-csh
-rwxr-xr-x 1 root root 155040 Sep 30  2016 /bin/bsd-csh
```

I'm not sure if that helps or not.  But... If you have docker installed on your machine, just do something like

```
kfranz@0283:~/continuum/conda master ❯ docker run -it --rm condatest/linux-64-python-3.6 /bin/bash
circleci@6c8be23ffb5e:/go$ csh
% pstree
bash───csh───pstree
%
```

You should have everything you need there to interact with csh.  Conda is installed at `/opt/conda`.  I think the current version is 4.3.30, but just install whatever conda you need.

njalerikson:
Ok. I was finally able to fully replicate the issue (the `csh` used in the container was actually the same as the one I found on Ubuntu).

The issue is as I previously mentioned. Pure `csh` doesn't allow argument passing to a sourced script. This is actually explicitly mentioned in the man page:

```csh
# csh man page
    source name
    source -h name

# tcsh man page
    source [-h] name [args ...]
```

This would suggest that there are only two real options:
1. Take a page from LMOD (and other tools out there) and instead of conda being an alias to a source command, make it an alias of an eval command - this would likely result in the elimination of both conda.csh and conda.sh. But if I remember correctly we had this conversation in the past and @kalefranz mentioned not wanting all conda commands to be evaluated.
2. Simply ignore pure CSH and other shells that cannot support necessary argument passing. Implementing a fix that relies on environment variables instead (as my original PR from way back when had) would be a lot of additional work for relatively limited gains.

So I guess this means we should revert to skipping the failing test in order to move on with this PR?

njalerikson:
Any idea why `test_repo` is failing in the `conda-build` tests? As far as I can tell it doesn't appear related to any of the changes introduced here.

njalerikson:
@kalefranz what's the next steps for this? I want to get this finished so I can continue with the other PRs.

kalefranz:
The conda-build tests failures are not related to this PR, and I don't mind merging it with those failures.

njalerikson:
My gut feeling is that we shouldn't care about pure CSH (because I had to deliberately seek it out, so while users may think they are using CSH it's actually TCSH, its almost always just symlinked).

If we do wish to care about pure CSH then that would need to be a whole separate PR/refactor since `source` simply cannot solve that problem.

PR #7042, opened 2018-03-15
Handle non-zero return in CSH

njalerikson:
Splitting #7026

This specifically addresses #6877

#7041 is a prerequisite

### DIFF
--- a/conda/shell/etc/profile.d/conda.csh
+++ b/conda/shell/etc/profile.d/conda.csh
@@ -27,16 +27,19 @@ if ("`alias conda`" == "") then
 else
     switch ( "${1}" )
         case "activate":
-            eval "`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh activate '${2}' ${argv[3-]})`"
+            set ask_conda="`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh activate '${2}' ${argv[3-]})`" || exit ${status}
+            eval "${ask_conda}"
             rehash
             breaksw
         case "deactivate":
-            eval "`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh deactivate '${2}' ${argv[3-]})`"
+            set ask_conda="`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh deactivate '${2}' ${argv[3-]})`" || exit ${status}
+            eval "${ask_conda}"
             rehash
             breaksw
         case "install" | "update" | "uninstall" | "remove":
             $_CONDA_EXE $argv[1-]
-            eval "`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh reactivate)`"
+            set ask_conda="`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh reactivate)`" || exit ${status}
+            eval "${ask_conda}"
             rehash
             breaksw
         default:


conda/shell/etc/profile.d/conda.csh
```diff
@@ -27,16 +27,19 @@ if (`alias conda` == "") then
 else
     switch ( $1 )
         case "activate":
-            eval `$_CONDA_EXE shell.csh activate "$2" $argv[3-]`
+            set ask_conda="`(setenv prompt '${prompt}' ; '${_CONDA_EXE}' shell.csh activate '${2}' ${argv[3-]})`" || exit ${status}
```
kalefranz: Just want to clarify... Is `exit` correct here?  In bash, it's be `return`, because we're running in the current process.  If we used `exit`, we'd be exiting out of the current shell, which isn't what we want.

njalerikson: I do not have any one reference to prove this, rather it's easier to show you. Yay for spotty (T)CSH documentation...

I have included two scripts, one for (T)CSH and one for Bash. The Bash version is just a convenience that mirrors the (T)CSH script. Both scripts accepts one argument, the argument should be `s` (for source) or something else (for execute).

When a (T)CSH script is sourced the correct way to "bubble" out of the script is via `exit`, this will not kill your actual (T)CSH shell. If however the script is being executed, then `exec /bin/true` or `exec /bin/false` should be used (depending on the error code we want). I am not sure in what universe this made sense...

### Untitled.csh ((T)CSH Script)
```csh
./untitled.csh s  # an ok exit for non-nested scripts
./untitled.csh e  # a proper "bubble" out exit
source ./untitled.csh s  # a proper source exit
source ./untitled.csh e  # will kill top level shell
```

#### The Script
```csh
#!/usr/bin/env csh -f

# exit according to user inputs
if ( "${1}" == "s" ) then
    echo '(T)CSH Sourced - running `exit 1`'
    exit 1
else
    # get True value
    if ( -x "/usr/bin/false" ) then
        set FALSE=/usr/bin/false
    else
        set FALSE=/bin/false
    endif

    echo '(T)CSH Executed - running `exec '${FALSE}'`'
    exec ${FALSE}
endif
```

### Untitled.sh (Bash Script)
```bash
./untitled.sh s  # "can only `return' from a function or sourced script" error
./untitled.sh e  # a proper exit
. ./untitled.sh s  # a proper source exit
. ./untitled.sh e  # will kill top level shell
```

#### The Script
```bash
#!/usr/bin/env bash -f

# exit according to user inputs
if [ "${1}" == "s" ]; then
    echo 'BASH Sourced - running `return 1`'
    return 1
else
    echo 'BASH Executed - running `exit 1`'
    exit 1
fi
```

kalefranz: lol cool.  That's way more than I needed.  I just wanted to double check and just be sure you really meant `exit`. I trust that you know (t)csh *way* better than I do.

kalefranz:
Now... This test failure... The one here is an easy fix.  You just need to change `unset` to `unsetenv` in the relevant test.

These unit tests are all *really* locked down.  And that's a good thing!  Makes sure we're absolutely explicit about changes.  And yours here seems just fine to me.  As far as I'm concerned, you especially own the (t)csh stuff.

njalerikson:
whoops, that was not what I wanted to do...

kalefranz:
Haha. You need help fixing this, or can you dig your way out?  Just ask if you want some pointers with git...

njalerikson:
I think I'm digging my way out (pointers would be good though), I clearly don't do this enough...

njalerikson:
Phew

kalefranz:
First, I set my remotes up so that `origin` is the main repo, and `kalefranz` (i.e. my github username) is my fork

```
$ git remote -vv
kalefranz	git@github.com:kalefranz/conda.git (fetch)
kalefranz	git@github.com:kalefranz/conda.git (push)
mcg1969	git@github.com:mcg1969/conda (fetch)
mcg1969	git@github.com:mcg1969/conda (push)
origin	git@github.com:conda/conda.git (fetch)
origin	git@github.com:conda/conda.git (push)
```

Make sure your local `master` is tracking `origin/master`.

    git checkout master
    git branch -u origin/master

And some pieces from my global `~/.gitconfig`

```
[push]
	default = current
[pull]
	default = current
	rebase = true
[branch]
	autosetuprebase = always
[credential]
	helper = osxkeychain
[commit]
	gpgsign = true
[core]
	editor = vim
	excludesfile = /Users/kfranz/.gitignore_global
[color]
	ui = true
[merge]
    # make --no-ff the default behavior
    ff = false
[rerere]
        # this one is important to remember rebasing changes, so you don't have to make them multiple times!
	enabled = true
```

And finally, from where you were with the bit of a mess, I would have done

```
git fetch origin
git checkout csh-return
git reset --hard origin/master
git cherry-pick 09febb09bcc438c2ab71075dcb5ae0386b668740
git push -f
```

kalefranz:
^^ Someday I'll get around to having developer docs with this stuff in it.

njalerikson:
The failing test in `py27` appears to be unrelated to this PR, but rather related to the test's timeout:

https://github.com/conda/conda/blob/27157ddcec0fefbccd168b6b5be7258f03115b05/tests/test_activate.py#L1091-L1092

PR #7046, opened 2018-03-16
Check if original fish prompt functions exist before copying onto them

teake:
Copying fish functions to existing targets yields an error.
So sourcing `conda.fish` twice was not possible.

Besides, if it wouldn't have yielded an error, you'd attach extra copies
of the conda prompt modifier each time you'd source the file.


teake:
@conda/conda-core bump?

teake:
This is getting a bit stale. Who do I bump to get this merged? @kalefranz?

travishathaway:

PR #7049, opened 2018-03-17
resolve #6595 use OO inheritence in activate.py

### LINKED ISSUE #6595: Clean up activate.py using proper OO inheritence

kalefranz:
resolve #6595

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -209,13 +209,10 @@ class _Activator(object):
             }
             deactivate_scripts = ()
 
-        self._update_prompt(set_vars, conda_prompt_modifier)
+        if context.changeps1:
+            self._update_prompt(set_vars, conda_prompt_modifier)
 
-        if on_win and self.shell == 'cmd.exe':
-            import ctypes
-            export_vars.update({
-                "PYTHONIOENCODING": ctypes.cdll.kernel32.GetACP(),
-            })
+        self._build_activate_shell_custom(export_vars)
 
         return {
             'unset_vars': (),
@@ -276,7 +273,8 @@ class _Activator(object):
             }
             activate_scripts = self._get_activate_scripts(new_prefix)
 
-        self._update_prompt(set_vars, conda_prompt_modifier)
+        if context.changeps1:
+            self._update_prompt(set_vars, conda_prompt_modifier)
 
         return {
             'unset_vars': unset_vars,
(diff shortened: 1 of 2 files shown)


conda/activate.py
```diff
@@ -9,15 +9,15 @@
...
+context.__init__()  # On import, context does not include SEARCH_PATH. This line fixes that.
 
 try:
     from cytoolz.itertoolz import concatv, drop
 except ImportError:  # pragma: no cover
     from ._vendor.toolz.itertoolz import concatv, drop  # NOQA
 
 
-class Activator(object):
+class _Activator(object):
```
mbargull: We *could* make this `class _Activator(ABC)` and thus use `@abstractmethod` to automatically check for methods/properties being implemented.
This would add a little bit more boilerplate (which I would be fine with). But we would also need to look out for Python 2 intricacies (e.g., need to use `metaclass=ABCMeta`, the deprecated `@abstractproperty`, ...).
So, although I'd prefer using `ABC`, I'm also not eager to deal with PY2 stuff...

conda/activate.py
```diff
@@ -39,8 +39,21 @@ class Activator(object):
...
+    # The following instance variables must be defined by each implementation.
+    shell = None
+    pathsep_join = None
+    path_conversion = None
+    script_extension = None
+    tempfile_extension = None  # None means write instructions to stdout rather than a temp file
+    shift_args = None
+    command_join = None
+
+    unset_var_tmpl = None
```
mbargull: We could replace those template with methods that return the appropriate formatted strings, like suggested in https://github.com/conda/conda/pull/6585#issuecomment-354178822, which would make the code a little cleaner/better readable.
But that could also be done in a subsequent PR anytime and is not crucial for this one.

conda/activate.py
kalefranz: Yeah that's why I didn't go in that direction, and just renamed `Activator` to `_Activator`.

kalefranz:
@mbargull If you're around, do we look good to go here now?

mbargull:
LGTM. Saw that you already fixed `3.0 conda-build` on `master`.
Any further improvements should profit from these changes and can be put into subsequent separate PRs.

kalefranz:
Thanks @mbargull! I’m getting closer on #6518, and I rebased that PR against this one and built on top. So wanted to be getting this merged.

PR #7060, opened 2018-03-20
add CondaError details to auto-upload reports

kalefranz:


conda/exceptions.py
```diff
@@ -915,6 +915,10 @@ def get_error_report(self, exc_val, exc_tb):
             'traceback': _format_exc(exc_val, exc_tb),
             'conda_info': info_dict,
         }
+
+        if isinstance(exc_val, CondaError):
+            error_report['conda_error_dump'] = exc_val.dump_map()
```
kalefranz: I'm not sure I like the name of this key here...

kalefranz: Maybe `conda_error_details`?

kalefranz: `conda_error_components`

Issue #6353, opened 2017-11-28
`conda --help` does not show `conda activate` and `conda deactivate` info


kalefranz:
### What is the problem?

Currently, the main help page does not show `activate` or `deactivate`. 

### Why does this happen?

This happens because conda itself is not like most commands in that it must be installed into your shell profile for it to be used. When you call conda at the command line, you are actually invoking a shell function that finds and executes the write scripts. In these shell functions, `activate` and `deactivate` are treated differently than the other commands.

### How do we fix it?

To address this, we could potentially create some mock entries for "activate" and "deactivate" that would show up on the help page. Because these commands are intercepted and parsed first by shell functions, we do not have to worry about them being invoked instead of the real `activate` and `deactivate` commands (which are really just shell functions).

kalefranz:
> Is the `source` style supposed to be deprecated at some point? If so when?

Yes.  

1. I'd like to get the `conda init` PR well behind us first.  
2. Then get the Anaconda and Miniconda installers doing whatever the right thing is at that point (i.e. *not* `export PATH`).  
3. Then put formal deprecation notices in `activate` / `deactivate` / `activate.bat` / `deactivate.bat` with all the officially correct and forevermore settled shell setup instructions.

And, if we get all of that landed prior to conda 5.0, then that's probably when we'll actually remove those four files.

kalefranz:
Honestly, should we show both? Idk. We're in a little bit of limbo right now.  We really need this `conda init` stuff finalized.

If you're looking to beef up some help docs though, `conda activate --help` needs some... help

https://github.com/conda/conda/blob/b2915b97abea4cec509c6737eb4ddf30f05769ac/conda/activate.py#L175-L179

derrickoswald:
+1
It is deceitful to not mention two of the main commands (activate & deactivate) in the help screen - especially for new users who rely on what documentation there is.

dmd:
I just ran into this as well. I had a brain-fart and couldn't remember the literally most common thing I type, and then was stunned when the help mentions all sorts of commands I *never* use, but not the thing used most often.

travishathaway:
@dmd, this is annoying and also related to how the command parsing for the application works. I updated the description of this issue today to provide more background on why this happens and a potentially way to fix it.

PR #7070, opened 2018-03-22
Help for activate & deactivate

kenodegard:
This is one way to implement the help processing for activate/deactivate by relying on conda's argparse. Is this a good way to implement this or are there other ways that make more sense?

#6353

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -91,6 +91,7 @@ def main(*args):
     args = tuple(ensure_text_type(s) for s in args)
 
     if len(args) > 1:
+        from ..exceptions import HelpError
         try:
             argv1 = args[1].strip()
             if argv1.startswith('shell.'):
@@ -103,6 +104,8 @@ def main(*args):
             elif argv1 in ('activate', 'deactivate'):
                 from ..exceptions import CommandNotFoundError
                 raise CommandNotFoundError(argv1)
+        except HelpError:
+            args = ["-h"]
         except Exception as e:
             _, exc_val, exc_tb = sys.exc_info()
             init_loggers()
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -173,10 +173,8 @@ class Activator(object):
             from .exceptions import ArgumentError
             raise ArgumentError("'activate', 'deactivate', or 'reactivate' command must be given")
         elif help_requested:
-            from . import CondaError
-            class Help(CondaError):  # NOQA
-                pass
-            raise Help("help requested for %s" % command)
+            from .exceptions import HelpError
+            raise HelpError("help requested for %s" % command)
         elif command not in ('activate', 'deactivate', 'reactivate'):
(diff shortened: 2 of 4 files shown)


conda/cli/main.py
```diff
@@ -91,6 +91,7 @@ def main(*args):
     args = tuple(ensure_text_type(s) for s in args)
 
     if len(args) > 1:
+        from ..exceptions import HelpError
```
kalefranz: I don't think this PR is generally the right direction, and this line illustrates why.  The activate stuff has to be blazing fast, and imports end up costing us a lot.  The `conda/activate.py` file is already set up to support informational help strings appropriately.  We just need to fill out the content of those strings.  For example,

```
$ conda activate -h
help requested for activate
```

So that just needs content.  And for the `activate` and `deactivate` commands missing from `conda --help`, we can do something like we do with the "other commands" message overriding the `print_help()` function.  Just add some extra information in there to output about those two commands.

kalefranz: I should say that I'm also hostile to argparse in general, because of how much work it does to get all set up on every invocation of the program.  It's something we can live with (and just have to live with) for conda commands that are expected to take a bit longer, but for things like activate that potentially affect the startup time of shells (because `conda activate base` will be part of most people's shell setup), we *have* to be fast.

njalerikson: I was expecting you to say something along those lines. I guess my main point in trying to do it this way was to bring content together. There are definitely bits and pieces of conda that feels (or in the case of the old source scripts, felt) fragmented.

What if we were even more reluctant to import `HelpError`, hence it only gets imported inside the except clause (and then use an `isinstance` check)? This way we wont expect any slowdown for successful cases but a slight slowdown for error cases or cases requesting more info.

Adding the help info for activate and deactivate in a unique, separate manner from everything else feels like we're reinventing the wheel and makes the implementation feel fragmented again...

kalefranz: > I was expecting you to say something along those lines.

Well that kind of make me smile 😁

Ok, how about this... We already have help information in the `conda/exceptions.py` module, as you've probably noticed.  `CommandNotFoundError` for example.  I guess the whole exceptions module is really about user messaging, which makes since the help information could be there too.

So only line 106/109 below, 

    except Exception as e:

change the content after that to do type checking on the exception instance `e`.

```python
from ..exceptions import HelpError
if isinstance(e. HelpError):
```

Actually, scratch that...  All we need to do is a local import of `HelpError` in `conda/activate.py`.  And raise that.  Everything else will take care of itself then.  And we can still have the help information in `exceptions.py`.

njalerikson: So you're suggesting that instead of the help information being part of the argparse to make an exception that contains/prints the help info?

Do we want argparse to use this same strategy for printing the help or would there be two separate help implementations - a classic argparse print_help and an exception help?

PR #7079, opened 2018-03-25
resolve #7078 add warnings for deprecated modules

### LINKED ISSUE #7078: Add deprecation warnings for conda 4.6
deprecated modules:

*  `conda.cli.activate`

kalefranz:
resolve #7078

### DIFF
--- a/conda/cli/activate.py
+++ b/conda/cli/activate.py
@@ -218,5 +218,9 @@ def main():
     print(path)
 
 
+print("WARNING: The conda.cli.activate module is deprecated and will be removed in a "
+      "future release.", file=sys.stderr)
+
+
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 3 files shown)


goanpeca:
I have two questions:

1.) Why don't we use the warnings module?
2.) Why not make a helper function that only accepts the message, so we can avoid the `file=sys.stderr` ?

kalefranz:
Because I didn't want to spend the time with it, or over-engineer it.  Feel free to do that if you want though.

I also wrote some pretty decent deprecation library code at one point.  It's at https://github.com/conda/conda/blob/master/conda/_vendor/auxlib/deprecation.py

I haven't looked at it in a while, but maybe we should use that.  Feel free to.  Go for it.

goanpeca:
>I also wrote some pretty decent deprecation library 

Yes, it looks nice :-)

PR #7101, opened 2018-03-28
fish autocompletion for conda env

mdeff:
This PR adds autocompletion for `conda env` to the fish shell.

### DIFF
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -107,8 +107,17 @@ end
 # Faster but less tested (?)
 function __fish_conda_commands
   string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda/cli/main_*.py
+  for f in $_CONDA_ROOT/bin/conda-*
+    if test -x "$f" -a ! -d "$f"
+      string replace -r '^.*/conda-' '' "$f"
+    end
+  end
   echo activate
   echo deactivate
+end
+
+function __fish_conda_env_commands
+  string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda_env/cli/main_*.py
 end
 
 function __fish_conda_envs
@@ -139,6 +148,7 @@ end
 
 # Conda commands
 complete -f -c conda -n '__fish_conda_needs_command' -a '(__fish_conda_commands)'
+complete -f -c conda -n '__fish_conda_using_command env' -a '(__fish_conda_env_commands)'
 
 # Commands that need environment as parameter
 complete -f -c conda -n '__fish_conda_using_command activate' -a '(__fish_conda_envs)'


mdeff:
CC @jaimergp and @kalefranz who worked on #6885 and #6502

conda/shell/etc/fish/conf.d/conda.fish
```diff
@@ -109,6 +109,11 @@ function __fish_conda_commands
   string replace -r '.*_([a-z]+)\.py$' '$1' $_CONDA_ROOT/lib/python*/site-packages/conda/cli/main_*.py
   echo activate
   echo deactivate
+  echo env
```
mbargull: I think instead of `echo env` we should do the more general
```fish
  string replace -r '^.*/conda-' '' "$_CONDA_ROOT"/bin/conda-*
```
or maybe better
```fish
  for f in "$_CONDA_ROOT"/bin/conda-*
    if test -x "$f" -a ! -d "$f"
      string replace -r '^.*/conda-' '' "$f"
    end
  end
```
That way we would also pick up other commands, e.g., from `conda-build`.
NOTE: I have next to no `fish` experience, so I don't know certainly if the code snippet above is fine or not.

mdeff: Good idea! I think the first is fine. We don't have conda directories AFAIK, and we might discover other things to exclude later anyway.

mdeff:
Thanks for your review @mbargull. I've implemented your suggestion.

I too thought about keeping the separation between conda and conda-env by putting these rules in `conda_env/shell/etc/fish/conf.d/conda_env.fish` but thought it was a bit absurd to create this whole hierarchy for some lines of fish shell (plus we'd have to insert the `_CONDA_EXE` and `_CONDA_ROOT` variables in this file at install time). Let me know if you think it's preferable this way and I'll move it.

conda/shell/etc/fish/conf.d/conda.fish
mbargull: I was more thinking about the `test -x` part, i.e., to only pick up executables. The `-d` part is only to a necessary addition.
If one would do something like writing a (of course non-executable) `conda-build.log` file to `bin` or another stupidity, then we can easily exclude that with the second code snippet.
(BTW, `conda` itself doesn't do that executable test yet and thus will ungracefully exit if you were to run it with such `conda conda-build.log` -- I'll submit a PR for this.)

mdeff: Looks strange to me to have non-executable files in `bin` but ok, done. :)

mbargull: Definitely agree to that. But who knows what people might do :wink: -- we can always remove that check later on.
One case where we have non-executables would be on Windows where (due to no shebang support) a single `conda` script is replaced by `conda.exe` and `conda-script.py` (with the former essentially running `python conda-script.py`).
But well, to satisfy the "immensely large" Windows+`fish`+`conda` user base, we'd have to glob for `conda-*(.exe)?` or something :laughing:

jaimergp:
LGTM, but I haven't tested it yet. Do you need me to check it today?

conda/shell/etc/fish/conf.d/conda.fish
mdeff: Hahaha this immensely large user base will certainly submit a PR if there's an issue :laughing:

jaimergp:
Quick update: I have replaced my system's `conda.fish` with this proposal and works perfect.

mdeff:
@mbargull can we merge?

kalefranz:
Thanks for the PR!

PR #7133, opened 2018-04-06
resolve #4274 add option to remove an existing environment with 'conda create'

### LINKED ISSUE #4274: ENH: add --force option to conda env create
Both conda env create and conda create need a --force option that will overwrite the environment if it already exists.

conda could warn the user and prompt for confirmation but must also allow for --yes option to not prompt so conda can be used effectively in automation scripts that won't have human interaction.

kalefranz:
resolves #4274

This PR also introduces one new capability.  The `conda create` command has learned a `--remove-existing` flag, that'll remove an existing conda environment at the target location if one already exists (rather than throwing an error).

### DIFF
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -4,6 +4,7 @@ from os.path import basename, dirname
 import re
 import sys
 
+from conda.common.constants import NULL
 from .._vendor.auxlib.ish import dals
 from ..base.constants import ROOT_ENV_NAME
 from ..base.context import context
@@ -43,8 +44,9 @@ def confirm(message="Proceed", choices=('yes', 'no'), default='yes'):
             return choices[user_choice]
 
 
-def confirm_yn(message="Proceed", default='yes', dry_run_exit=True):
-    if dry_run_exit and context.dry_run:
+def confirm_yn(message="Proceed", default='yes', dry_run=NULL):
+    dry_run = context.dry_run if dry_run is NULL else dry_run
+    if dry_run:
         from ..exceptions import DryRunExit
         raise DryRunExit()
     if context.always_yes:
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -43,8 +43,8 @@ def confirm(message="Proceed", choices=('yes', 'no'), default='yes'):
             return choices[user_choice]
 
 
-def confirm_yn(message="Proceed", default='yes'):
-    if context.dry_run:
+def confirm_yn(message="Proceed", default='yes', dry_run_exit=True):
+    if dry_run_exit and context.dry_run:
         from ..exceptions import DryRunExit
         raise DryRunExit()
     if context.always_yes:
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -8,7 +8,7 @@ from .._vendor.auxlib.ish import dals
 from ..base.constants import ROOT_ENV_NAME
 from ..base.context import context
 from ..common.constants import NULL
-from ..common.io import swallow_broken_pipe, timeout
+from ..common.io import swallow_broken_pipe
 from ..common.path import paths_equal
 from ..common.serialize import json_dump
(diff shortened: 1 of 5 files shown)


conda/cli/common.py
```diff
@@ -4,6 +4,7 @@
 import re
 import sys
 
+from conda.common.constants import NULL
```
kalefranz: TODO: before merge, fix this import

kalefranz:
I'm interested to have feedback on this workflow.  In the original commit, I had a command flag and configuration parameter for `remove_existing`.  In the second commit, I removed the extra flag, because it clashed with the apparent meaning of `always_yes`.  If `always_yes` really does mean "always yes," then we shouldn't have an extra layer of confirmation to get in the way.

conda/cli/main_create.py
```diff
@@ -3,10 +3,36 @@
...
+from ..gateways.disk.test import is_conda_environment
+
+log = getLogger(__name__)
 
 
 def execute(args, parser):
+    if is_conda_environment(context.target_prefix):
+        if paths_equal(context.target_prefix, context.root_prefix):
+            raise CondaValueError("The target prefix is the base prefix. Aborting.")
+        confirm_yn("WARNING: A conda environment already exists at '%s'\n"
```
kalefranz: TODO: Incorporate `conda.common.io.timeout` into the `confirm_yn()` function before this PR is merged.

kalefranz: Done.

goanpeca:
The original issue suggested using --force to provide this behavior. Is force not the right "concept" so the additional new flag can be avoided?

kalefranz:
I originally had a `--remove-existing` flag.  You can see it in the first commit.  After I worked through it, this workflow felt better.

I'll demo it next Wednesday and see what other people think.

goanpeca:
Ok, no I am confused, so now there is no extra flag, and only `force`is used instead?

kalefranz:
>  I see "--remove-existing" in the code still

Oops that's wrong.  No wonder you were confused.  I'll fix that.

There's no `--force` flag that applies here.  I'm pretty close to getting rid of the `--force` flag completely (or hiding it in a deep dark corner).  The problem is that `--force` is overloaded and means different things in different contexts, but often those different contexts all apply during a single conda operation.  So now in master there are for example `--force-reinstall`, `--force-remove`, and `--clobber` flags.

The flow in this PR is that if you do `conda create`, and the environment already exists, we prompt to ask you if you want to delete the environment.  The default answer is 'no' of course.

conda/cli/common.py
```diff
@@ -43,18 +44,22 @@ def confirm(message="Proceed", choices=('yes', 'no'), default='yes'):
...
+def confirm_yn(message="Proceed", default='yes', dry_run=NULL):
+    dry_run = context.dry_run if dry_run is NULL else dry_run
+    if dry_run:
         from ..exceptions import DryRunExit
         raise DryRunExit()
     if context.always_yes:
         return True
     try:
-        choice = confirm(message=message, choices=('yes', 'no'),
+        choice = timeout(30, confirm, message=message, choices=('yes', 'no'),
```
mbargull: The timeout should be configurable and, I guess, infinite on default for now.
It could be helpful to prevent complete stalling in unattended execution. But other times it could be annoying:
Even though `conda >=4.4` has a considerably faster solving phase, there still might be situations where a user isn't actively waiting and the "grab-a-coffee-time" might exceed 30 seconds :wink:.

kalefranz: I'm starting to become sensitive to configuration parameter fatigue.  This is trivial enough that I'd rather not even use the timeout wrapper here then if we think it's something that needs to be configurable.

mbargull: > I'm starting to become sensitive to configuration parameter fatigue.

Understandable. At some point this becomes nightmare fuel :fearful:

PR #7135, opened 2018-04-08
fix #2154 conda remove should complain if requested removals don't exist

### LINKED ISSUE #2154: conda remove doesn't complain if only some of the packages don't exist
If I do `conda remove foo bar` and neither of them are in the current environment then I get an error.  However if I do `conda remove foo bar` and `foo` is present but `bar` is not then `foo` gets removed but there is no error or warning that `bar` wasn't there and was not removed.  I discovered this because I misunderstood the behavior of the `--features` flag, so it also applies to removing features.

I would suggest that a warning is required as a minimum, and perhaps it should be an error to be "symmetric" with a `conda install foo bar` where `foo` exists to be installed but `bar` does not (which just reports an error, and does not install anything).

marcelotrevisani:
fix #2154 
Bug fix about the Issue 2154. With this implementation Conda raises
an exception when the user tries to remove a non-existent package.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -190,12 +190,18 @@ class Solver(object):
             graph = PrefixGraph((index[dist] for dist in solution), itervalues(specs_map))
 
             removed_records = []
+            records_not_found = []
             for spec in specs_to_remove:
                 # If the spec was a track_features spec, then we need to also remove every
                 # package with a feature that matches the track_feature. The
                 # `graph.remove_spec()` method handles that for us.
                 log.trace("using PrefixGraph to remove records for %s", spec)
-                removed_records.extend(graph.remove_spec(spec))
+                spec_remove = graph.remove_spec(spec)
+                removed_records.extend(spec_remove)
+                if not spec_remove:
+                    records_not_found.append(spec.name)
+            if records_not_found:
+                raise PackagesNotFoundError(records_not_found)
 
             for rec in removed_records:
                 # We keep specs (minus the feature part) for the non provides_features packages
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -208,7 +208,9 @@ class Solver(object):
                 if not any(spec.match(rec) for rec in all_removed_records)
             )
             if unmatched_specs_to_remove:
-                raise PackagesNotFoundError(unmatched_specs_to_remove)
+                raise PackagesNotFoundError(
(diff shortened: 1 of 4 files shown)


marcelotrevisani:
@kalefranz , Could you please review that? And it's kinda weird why this is falling in parallel tests, the first time I ran that failed a different test. It seems the parallel tests are kinda a flake (some of them)

kalefranz:
@marcelotrevisani I've added a commit to this PR that I think gets us closer.  What do you think?

conda/core/solve.py
```diff
@@ -195,7 +195,8 @@ def solve_final_state(self, deps_modifier=NULL, prune=NULL, ignore_pinned=NULL,
                 # package with a feature that matches the track_feature. The
                 # `graph.remove_spec()` method handles that for us.
                 log.trace("using PrefixGraph to remove records for %s", spec)
-                removed_records.extend(graph.remove_spec(spec))
+                spec_remove = graph.remove_spec(spec)
```
kalefranz: This line is no longer needed.

tests/test_create.py
```diff
@@ -995,14 +1004,6 @@ def test_update_deps_flag_present(self):
...
             assert package_is_installed(prefix, 'flask')
 
-    # @pytest.mark.skipif(not on_win, reason="shortcuts only relevant on Windows")
-    # def test_shortcut_in_underscore_env_shows_message(self):
-    #     prefix = make_temp_prefix("_" + str(uuid4())[:7])
-    #     with make_temp_env(prefix=prefix):
-    #         stdout, stderr = run_command(Commands.INSTALL, prefix, "console_shortcut")
-    #         assert ("Environment name starts with underscore '_'.  "
-    #                 "Skipping menu installation." in stderr)
-
```
kalefranz: Let's put this one back.  The rest of the commented out test code is fine to be gone.

marcelotrevisani:
Sure! Looks good to me!
:)

tests/test_create.py
marcelotrevisani: Sure. But I just want to know, why these commented code are necessary?

kalefranz:
Looks like I created new problems :-/

marcelotrevisani:
hahaha, sometimes this also happens. Do you want to fix that? I think there is some error in your commit because the test is failing. Just let me know if you need help! :)

kalefranz:
Should be good to go.  I'll double-check that all tests pass in master after merge.

Issue #7092, opened 2018-03-27
Incorrect command to activate the environment, if there is space in environment name.


kalefranz:
_From @RidaZubair on March 27, 2018 11:51_

**OS: Linux
Anaconda Navigator**

**Steps to Reproduce:**

- Install Anaconda.
- Create an environment with space.
- Check the command to activate the environment

**Actual Result:**
inverted commas are missing in command to activate the environment if there are spaces in name.
![image](https://user-images.githubusercontent.com/27444898/37964518-3fcfe6ea-31db-11e8-8720-8a8da9242a6f.png)

_Copied from original issue: ContinuumIO/anaconda#1367_

kalefranz:
_From @csoja on March 27, 2018 14:3_

@kalefranz I didn't think we supported spaces in environment names?  Is that documented?

kalefranz:
You have to quote the name. Standard command line parsing.

kalefranz:
_From @sriaz2 on March 27, 2018 14:58_

@kalefranz "You have to quote the name. Standard command line parsing."
That is what the issue is reporting, the command presented by conda in the terminal to activate environment should contain quotes if there is a space in the environment name. 
<img width="715" alt="screen shot 2018-03-27 at 7 58 14 pm" src="https://user-images.githubusercontent.com/20394675/37975652-39817b0a-31f9-11e8-977c-6d634f73959e.png">

kalefranz:
The correct fix here is to detect spaces in the environment name, and then if detected, show the red box as

    source activate "test space enviroooooonment"

mingwandroid:
Or just don't support this pointless misfeature :-)

kalefranz:
Right.  We could also raise an error on `conda create -n "environment name"` that spaces are not allowed in environment names.

kalefranz:
> inverted commas are missing in command

???????

kalefranz:
Oh, quotation marks.  I've never heard them called inverted commas haha

goanpeca:
This has been fixed on #7168. Closing

PR #7138, opened 2018-04-09
Handle env name with space, Issue#7092

makbigc:
Modify conda/activate.py to accept unquoted env name containing space. #7092

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -116,17 +116,16 @@ class _Activator(object):
         elif command not in ('activate', 'deactivate', 'reactivate'):
             from .exceptions import ArgumentError
             raise ArgumentError("invalid command '%s'" % command)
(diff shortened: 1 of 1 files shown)


mingwandroid:
Why would we want this? It will cause trouble that need fixes all over the ecosystem.

kalefranz:
@makbigc Thanks for your PR! Unfortunately, this isn't the direction we want to take.  It would be incorrect to do this from a technical level, and it would also create inconsistencies all over the conda CLI, not to mention knock-on effects across the whole package ecosystem.

I've added an additional comment in #7092 about the correct way to resolve the issue.  We'd also welcome PRs to add additional documentation to our FAQ https://conda.io/docs/user-guide/troubleshooting.html or explanatory help in other parts of our docs.

Issue #7142, opened 2018-04-10
delete_trash missing for clean


garu57:
**I'm submitting a...**
  - [ ] bug report
  - [ ] feature request

### Steps to Reproduce

```

```

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```

```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```

```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

```

```
</p></details>

garu57:
I think there is another problem with `delete_trash()`, in `conda/cli/main_remove.py` the call looks to be out of place.
It is called _before_ the pages and env remove are done so at the end removed packages and env are left in `pkgs/.trash`.

garu57:
Well, after looking into the code, given windows limitation on locking, anything going into `.trash` for the current run, is something that couldn't be deleted in the first place, so trying to delete it again from there has the same chances to fail again.
This means that `delete_trash()` is actually deleting what was left from previous runs and moving `delete_trash()` from where it is now into `conda/cli/main_remove.py` shouldn't give any real advantage.

What do you think?

garu57:
Submitted PR #7149

garu57:
As a side note running `%CONDA%\python setup.py develop` as shown in github readme copies the scripts and clobbers irrecoverably the installation (anaconda or miniconda). Well, at least i was not able to make `conda` working again without a complete reinstall.

Running `%CONDA%\python setup.py develop -x`, ie without copying the scripts, gets the best of both worlds. When running `python -m conda` from the git repo root, that one is running. Calling `conda` command out of git repo root uses the installed one. This is coherent with the message printed by `setup.py`
```
WARNING: Your current install method for conda only supports conda
as a python library.  You are not installing a conda executable command
or activate/deactivate commands.  If your intention is to install conda
as a standalone application, currently supported install methods include
the Anaconda installer and the miniconda installer.  If you'd still like
for setup.py to create entry points for you, use `utils/setup-testing.py`.
```

Running `%CONDA%\python setup.py develop -u` in the second case, effectively restores the installation to the initial state.

May be this is worth documenting.

kalefranz:
> What do you think?

[...]

PR #7149, opened 2018-04-11
fix for issue 7142

garu57:
Fix for  #7142.
After testing i see that the change in `conda/cli/main_remove.py` is well worth when removing `conda` environments.
When removing an env it is moved entirely in `.trash`. If `delete_trash()` is called _after_ the move, at least the subdirectories (conda-meta DLLs, etc) are emptied and space is recovered.

Instead there is NO WAY in windows to delete dll's and python executables hardlinks that point to the same python that is running `conda`, from `conda` itself. They will be in use and `os.unlink()` will get permission denied error.

I think it should be documented that cleaning the `.trash` after removing env's in windows, can require manual intervention.

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -14,6 +14,7 @@ import sys
 
 from ..base.constants import CONDA_TARBALL_EXTENSION
 from ..base.context import context
+from ..gateways.disk.delete import delete_trash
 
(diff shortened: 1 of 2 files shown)


roachsinai:
how is this going on?

travishathaway:

PR #7155, opened 2018-04-13
add a verification rule to ensure we never clobber system python

kalefranz:
As we make our way to supporting `pip install conda` via #6518, I want to add this check in to make it extra hard for people to use conda to change their system python.

### DIFF
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -4,12 +4,14 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 from collections import defaultdict, namedtuple
 from logging import getLogger
 import os
-from os.path import basename, dirname, isdir, join
+from os.path import basename, dirname, isdir, join, exists
 from subprocess import CalledProcessError
 import sys
 from tempfile import mkdtemp
 from traceback import format_exception_only
 import warnings
+
+import re
 
 from .package_cache_data import PackageCacheData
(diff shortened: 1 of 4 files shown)


mingwandroid:
What about `homebrew` Python installed elsewhere than `/usr/local`? And macOS system Python (not in `/usr`)?

I think refusing to clobber these Pythons is a good idea, but shouldn't we just check the version string and refuse to do anything on any non `Anaconda`/`Miniconda`/`conda-forge` Python?

jjhelmus:
> I think refusing to clobber these Pythons is a good idea, but shouldn't we just check the version string and refuse to do anything on any non Anaconda/Miniconda/conda-forge Python?

Whitelisting versions strings would prevent [random](https://anaconda.org/intel/python) [people's](https://anaconda.org/rpi/python) Python conda packages from working.  I would rather see this controlled via the presence/lack of a file or something else that doesn't require re-compiling code.

kalefranz:
> but shouldn't we just check the version string 

That's actually what I had initially.  Something that searched the version string for content between pipes, i.e. `\|(.*?)\|`, but I backed off of it for some reason.  I searched through my git history, but I guess I never checked that code in.  Guess I need to reimplement it and see if I still agree with the reason I backed off and went with this alternate implementation.

kalefranz:
This seems controversial enough that I don't want to hold up an alpha release to get this in.

PR #7162, opened 2018-04-13
resolve #7061 remove pkgs/pro from defaults

### LINKED ISSUE #7061: Remove pkgs/pro from defaults in conda 4.6
For the release for conda 4.6.0, remove `https://repo.anaconda.com/pkgs/pro` from the default value for `defaults`.

kalefranz:
resolve #7061

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -10,9 +10,9 @@ from platform import machine
 import sys
 
 from .constants import (APP_NAME, DEFAULTS_CHANNEL_NAME, DEFAULT_AGGRESSIVE_UPDATE_PACKAGES,
-                        DEFAULT_CHANNELS, DEFAULT_CHANNEL_ALIAS, ERROR_UPLOAD_URL,
-                        PLATFORM_DIRECTORIES, PREFIX_MAGIC_FILE, PathConflict, ROOT_ENV_NAME,
-                        SEARCH_PATH, SafetyChecks)
+                        DEFAULT_CHANNELS, DEFAULT_CHANNEL_ALIAS, DEFAULT_CUSTOM_CHANNELS,
+                        ERROR_UPLOAD_URL, PLATFORM_DIRECTORIES, PREFIX_MAGIC_FILE, PathConflict,
+                        ROOT_ENV_NAME, SEARCH_PATH, SafetyChecks)
 from .. import __version__ as CONDA_VERSION
 from .._vendor.appdirs import user_data_dir
 from .._vendor.auxlib.collection import frozendict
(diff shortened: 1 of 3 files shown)


conda/base/context.py
```diff
@@ -577,11 +577,14 @@ def custom_multichannels(self):
     @memoizedproperty
     def custom_channels(self):
         from ..models.channel import Channel
+        hard_coded_custom_channels = (Channel.make_simple_channel(self.channel_alias, url, name)
+                                      for name, url in iteritems(DEFAULT_CUSTOM_CHANNELS))
```
mbargull: Did you opt to not use
```python
    _custom_channels = MapParameter(string_types, DEFAULT_CUSTOM_CHANNELS, aliases=('custom_channels',))
```
to avoid a breaking change for existing user configuration that use the default `defaults` but also user-defined `custom_channels`?
It feels a little awkward to hardcode `pkgs/pro` that deeply, given that `defaults` itself isn't hardcoded by that amount.

(Due to its very limited value, do many people even use `custom_channels` at all? I for one was very confused by its purpose when I first encountered it. To me it feels like a "deprecation"-worthy parameter in favor of `custom_multichannels`. But I have no idea how frequently it's used...)

kalefranz: I'm pushing the change as you propose.  It's the right way to do it.

> Due to its very limited value, do many people even use custom_channels at all?

I don't think so.  I also think that people are just now starting to learn about it because our documentation has been (and still is lacking).  I do think there's a distinction between `custom_channels` and `custom_multichannels` though still.  In that a custom channel need not be a single-file multichannel.  That seems awkward to me and I'd be frustrated with being forced into that if I was actually using the feature.

mbargull: When I saw `custom_channels` first, I thought (w/o reading the docs -- genius, I know...) I could just use
```yaml
custom_channels:
  bioconda-broken: https://conda.anaconda.org/bioconda/label/broken
```
but then learned that the key is tied to become the value's suffix and thus only of limited use.
Hence, I needed to use that "single-channel multichannel", i.e.,
```yaml
custom_multichannels:
  bioconda-broken:
    - https://conda.anaconda.org/bioconda/label/broken
```
. To me `custom_channels` feels redundant since (if I didn't miss anything) `custom_multichannels` provides a true superset of its features, e.g., to replicate the example from
`conda config --describe custom_channels`:
```yaml
custom_multichannels:
  conda-forge:
    - https://anaconda-repo.dev/packages/conda-forge
```

mbargull: If we wanted to change this, we'd need an extended deprecation phase, i.e., nothing that should be done in any upcoming `4.x` release, I think. IIUC, that implementation exists since `4.2` or earlier? Definitely quite some time in which it could have been picked up by a few users..

PR #7163, opened 2018-04-13
Remove reference to 'system' channel.

scw:
This channel no longer exists, and looks like an artifact left over in
the doc from a conda 2.x feature that is no longer supported nor
necessary (.condarc specified channels are search by default). This came
up on the conda mailing list as a point of confusion, and is confusing
since the term 'system' is already overloaded with multiple meanings.


goanpeca:
@electronwill do you know if there is any reference to this on the dosc (anaconda/conda?)

electronwill:
I just checked and found no references to a channel called "system" in the conda or Anaconda docs. I've heard about the changes involving the channels "free" and "main", but I don't ever remember hearing anyone mention a channel called "system".

Issue #6338, opened 2017-11-25
activate.d scripts is likely not ran in order.


rainwoodman:
scripts in  /etc/conda/activate.d are likely not ran in order:

```
   if [[ -d "$_CONDA_D" ]]; then
        eval $(find "$_CONDA_D" -iname "*.sh" -exec echo source \'{}\'';' \;)
    fi

```

See here:
https://serverfault.com/questions/181787/find-command-default-sorting-order

If there are interdependency between scripts this will be very badly messed up. -- I am looking into if we can make the new cross-compilation enabled mpich package to fall back to use CXX and CC when the cross-toolchain is not installed (the result should still work on the local system since it is ABI compatable). For this to work, however, it depends on the cross compilation activate.d scripts to run before the mpich one.

mingwandroid:
A PR to conda would be welcome here. We're all probably too busy. You'd need to perform a topological sort. You wouldn't need to do this upon every activation, just when things get installed or removed form an env, caching the result in a script to be run.

rainwoodman:
I am not sure if I understand your proposal...

I was thinking more along the lines of init.d scripts style -- ordering by file names, then run forward in activate and backward in deactivate. This is sufficiently deterministic to implement some depencency. 

However the Linux init.d scripts then eventually it was replaced/superseded with real 'dependency rules' with system.d which is the right thing to do, but too much of an effort. So I don't actually know which route to take.

mingwandroid:
My proposal is to respect the dependencies and run the activation scripts in that order.

rainwoodman:
Your proposal will require conda to be able to tell the owner package of individual files. I don't think it can do this.

I can file a PR to run the files in alphabetical order if you'd like -- not quite there but it will at least make sure the order is deterministic.

mingwandroid:
Conda absolutely can tell the owner for each file except in the case of ambiguity in which case just talking the first is fine.  Search for which_package in conda-build (look like my fork for a much faster version).

mingwandroid:
Please look through the conda source code instead of asking me to do the same.

mingwandroid:
@kalefranz, making sure you are aware of this discussion incase you have any objections to this idea or can provide and assistance to @rainwoodman

rainwoodman:
I will. I was asking just in case if anyone already knew where to look. It's a pretty big code base already.

mingwandroid:
[...]

Issue #5082, opened 2017-04-19
Add a prefix state file


kalefranz:
Add a new json-formatted file to the `conda-meta` directory in a prefix to hold state information. 
 Possible names for the file could be `state.json` or `prefix_state.json`.

Implementation should be through a new `PathAction` that gets added to the 'register' `ActionGroup` in `UnlinkLinkTransaction._prepare()`.

The only contents of the new file right now should be the dependency sort order or all the packages installed in the prefix (i.e. the result of `Resolve().dependency_sort()`.

kalefranz:
Also don't forget: add configuration to lock an environment against mutation operations #6231

kalefranz:
Also related:

  * https://github.com/conda/conda/issues/4929
  * https://github.com/conda/conda/issues/4943

kalefranz:
With this work, may also want to move certain configuration parameters (like `channels`) out of the current configuration context, and into this new state file.  Under this scenario, the `channels` configuration would only be used when first creating environments.  Once environments are created, channels are only pulled from the configuration for that individual environment.

kalefranz:
Configuration parameters that make sense to move to the state file when the environment is first created include:

- channels
- pinned_packages
- default_channels
- aggressive_update_packages ?
- auto_update_conda
- channel_priority
- track_features
- subdirs

kalefranz:
@jjhelmus The `subdirs` component above should handle keeping a "force_win32" env as `win-32`.

aryarm:
Have there been any updates on this issue? I think this would be really helpful to have. I'm hoping it would also help with https://github.com/conda/conda/pull/8776

PR #7176, opened 2018-04-17
Sort activation order of the activation scripts

shadowwalkersb:
Currently, the activation script paths are gathered with `glob.glob` which returns its results in arbitrary order. AFAICT the resulting results are not sorted or manipulated anywhere else. This caused us problems when we added an activation script to a metapackage of ours where we wanted to manipulate the environment variables set by the new compilers, https://github.com/ContinuumIO/anaconda-issues/issues/9172. This PR sorts the resulting list of activation scripts. This way packages can ensure when their activation scripts are run by naming them accordingly. As @msarahan [suggested](https://github.com/ContinuumIO/anaconda-issues/issues/9172#issuecomment-381607827), a script named `zz_patch_flags.sh` will run after compiler scripts that are `activate-*.sh`.

Here, sorting is done in two places, but one of them is redundant. I am submitting both of them, but will remove one upon feedback.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -414,9 +414,9 @@ class _Activator(object):
         ))))
 
     def _get_deactivate_scripts(self, prefix):
-        return self.path_conversion(glob(join(
+        return self.path_conversion(sorted(glob(join(
             prefix, 'etc', 'conda', 'deactivate.d', '*' + self.script_extension
-        )))
+        )), reverse=True))
 
 
 def expand(path):
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -409,9 +409,9 @@ class _Activator(object):
         return "(%s) " % conda_default_env if context.changeps1 else ""
 
     def _get_activate_scripts(self, prefix):
-        return self.path_conversion(glob(join(
+        return self.path_conversion(sorted(glob(join(
             prefix, 'etc', 'conda', 'activate.d', '*' + self.script_extension
-        )))
+        ))))
 
     def _get_deactivate_scripts(self, prefix):
         return self.path_conversion(glob(join(
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -170,7 +170,7 @@ class _Activator(object):
             #  i.e. step back down
             return self.build_deactivate()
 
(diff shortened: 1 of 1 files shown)


shadowwalkersb:
cc @mingwandroid , @msarahan

conda/activate.py
```diff
@@ -409,9 +409,9 @@ def _prompt_modifier(self, conda_default_env):
         return "(%s) " % conda_default_env if context.changeps1 else ""
 
     def _get_activate_scripts(self, prefix):
-        return self.path_conversion(glob(join(
+        return self.path_conversion(sorted(glob(join(
             prefix, 'etc', 'conda', 'activate.d', '*' + self.script_extension
-        )))
+        ))))
```
kalefranz: Let's keep all the logic in this helper method.  So prefer this over the change on line 173.

conda/activate.py
```diff
@@ -409,9 +409,9 @@ def _prompt_modifier(self, conda_default_env):
...
 
     def _get_deactivate_scripts(self, prefix):
         return self.path_conversion(glob(join(
```
kalefranz: We should also do the same thing here.  Except to be technically correct, it should be `sorted()` with a `reverse=True` flag.

kalefranz:
The test failures are a list vs tuple mismatch.  I think if you remove the line 173 change we should be good on those specific test failures.

mingwandroid:
I think sorted alphanumerically is better than random, but I also think topologically sorting first then alphanumerically after is best.

kalefranz:
Yeah I agree we need to sort topologically too. That’ll take additional work. The intent in 4.4 was to sort basically based on an ascii table.

conda/activate.py
shadowwalkersb: Agreed and done.

shadowwalkersb:
Anything else needs to be done here? When can we expect this to go in?

jakirkham:
Is there an issue for using topo sort for this?

kalefranz:
@jakirkham The launch point there is https://github.com/conda/conda/issues/5082

mbargull:
xref: https://github.com/conda/conda/issues/6338#issuecomment-348344786

PR #7178, opened 2018-04-18
Support complete MatchSpec syntax in environment.yml files

### LINKED ISSUE #6385: Channel pins in "environment.yaml" files are not saved to package specs
Channel pins specified in environment files are not respected.  For example,run ```conda env create``` with this environment file:

```yaml
name: channel-not-written-to-user-specs

dependencies:
  - defaults::six
```

If we look at conda-meta/six*.json we can see that the channel pin has not been added to specs. Where we should read ```"requested_spec": "defaults::six"```, we only find ```"requested_spec": "six"```.

This is with conda 4.4.0rc2.

ShannonGreen:
I think this fixes #6385.

I've run the conda_env tests and they seem OK, but I'm not sure if there are other wider impacts of this change.

### DIFF
--- a/conda_env/installers/conda.py
+++ b/conda_env/installers/conda.py
@@ -11,19 +11,8 @@ from conda.models.channel import Channel, prioritize_channels
 def install(prefix, specs, args, env, *_, **kwargs):
     # TODO: support all various ways this happens
     # Including 'nodefaults' in the channels list disables the defaults
-    new_specs = []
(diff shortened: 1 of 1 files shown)


ShannonGreen:
Looks like the 2.1 conda-build tests are broken upstream?

msarahan:
yeah, don't worry about conda-build 2.1.

PR #7181, opened 2018-04-18
[WIP] towards #6610 sudo workflow resolution

### LINKED ISSUE #6610: Workflow problem that causes permissions errors
1. Install miniconda as unprivileged user to `~/miniconda`
2. `sudo su - root`
3. Perform conda operation using the above miniconda install, where package download/extract is required

Result is that `~/miniconda/pkgs` gets populated with files that the unprivileged user doesn't have write access to.

In this particular flow, I think what needs to happen is that, in the process of step 3, the `~/miniconda/pkgs` package cache directory is flipped to read-only for the unprivileged user.

So we're going to be doing *something* with the package cache magic file `~/miniconda/pkgs/urls.txt`, but what exactly?  It's probably going to be something like `chown UID:GID ~/miniconda/pkgs/urls.txt` every time you *write* to that package cache.  And the UID part is simple.  How do we know what the best GID is though?  Do we just use the primary of the current user?  Seems like that would cause problems in certain workflows too.

kalefranz:
resolve #6610

Needs rebased following merge of #7180.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -14,7 +14,6 @@ from .constants import (APP_NAME, DEFAULTS_CHANNEL_NAME, DEFAULT_AGGRESSIVE_UPDA
                         PLATFORM_DIRECTORIES, PREFIX_MAGIC_FILE, PathConflict, ROOT_ENV_NAME,
                         SEARCH_PATH, SafetyChecks)
 from .. import __version__ as CONDA_VERSION
-from .._vendor.appdirs import user_data_dir
 from .._vendor.auxlib.collection import frozendict
 from .._vendor.auxlib.decorators import memoize, memoizedproperty
 from .._vendor.auxlib.ish import dals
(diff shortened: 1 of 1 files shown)


kalefranz:
In addition to the first commit, also need to look at doing a better job with

```python
    @property
    def envs_dirs(self):
        if self.root_writable:
            fixed_dirs = (
                join(self.root_prefix, 'envs'),
                join(self._user_data_dir, 'envs'),
                join('~', '.conda', 'envs'),
            )
        else:
            fixed_dirs = (
                join(self._user_data_dir, 'envs'),
                join(self.root_prefix, 'envs'),
                join('~', '.conda', 'envs'),
            )
        return tuple(IndexedSet(expand(p) for p in concatv(self._envs_dirs, fixed_dirs)))

    @property
    def pkgs_dirs(self):
        if self._pkgs_dirs:
            return tuple(IndexedSet(expand(p) for p in self._pkgs_dirs))
        else:
            cache_dir_name = 'pkgs32' if context.force_32bit else 'pkgs'
            return tuple(IndexedSet(expand(join(p, cache_dir_name)) for p in (
                self.root_prefix,
                self._user_data_dir,
            )))
```

beeankha:
This pull request is being closed in order to reduce the number of outdated/stale PRs on this repo; if this work is still actively in progress, please feel free to re-open!

PR #7195, opened 2018-04-20
resolve #7194 change default value of max_shlvl to 0

### LINKED ISSUE #7126: With conda being a shell function in 4.4+, what's the suggested way to use conda via scripts when an environment isn't explicitly activated by the user beforehand?
From gitter starting https://gitter.im/conda/conda?at=5ac4e56dc574b1aa3e67a0d8

@natefoo:

> With conda being a shell function in 4.4+, what's the suggested way to use conda via scripts when an environment isn't explicitly activated by the user beforehand?

@mbargull:

> You'd have to explicitly activate the environment beforehand for this to work. I.e., run
>
>     . /path/to/etc/profile.d/conda.sh
>     conda activate base
> If you run that outside of your script, then this would put the conda executable on PATH. If you run it inside of the script, then the script can use the shell function. There are some features, e.g., automatic environment reactivation, that the shell function provides. So if you install/remove packages with (de)activation script into the currently activated environment, you'd want to use the shell function, i.e., source conda.sh inside your script.

@natefoo:

> The issue here is that it's potentially really confusing for users. If I'm delivering a script that is supposed to create an env and install all our dependencies in to it, until 4.4 they didn't need to do anything other than the required conda set up (putting the base env's `bin` on `$PATH`). Now they have to activate the base env before they run the script.
> I can't do it in the script because I don't know where conda is installed if they don't activate the base env. But that wasn't the case before.
> If nothing else, an exported env var for the path to conda.sh would be sufficient.
> Because the function stuff is pretty transparent to users, so it's not going to be readily apparent why they can run conda on the command line but if they call my script in the same shell, it can't run conda.

@mbargull:

> For reference `CONDA_EXE` was introduced in conda/conda#6923.

@natefoo:

> `$CONDA_EXE` only gets you part of the way there, too, since you need the shell functions for `activate` to work, and if you don't have the path to `conda.sh`, you don't have the shell functions. I can use `$(dirname $CONDA_EXE)/../etc/profile.d/conda.sh` to get the path, but that's pretty hacky.
> 
> I guess source `$(dirname $CONDA_EXE)/activate <env>` would also work as long as `activate` and `deactivate` stick around, but it's a hack too.
> 
> The shell check in activate explicitly precludes `. $(dirname $CONDA_EXE)/activate ...` from working under dash. =(

### LINKED ISSUE #7194: In conda 4.6, change default value of max_shlvl to 0
Conda 4.4 reworked shell environment activation, and unified sometimes differentiated behavior across shells and platforms.  Prior to conda 4.4, one particular behavior difference was that environments could effectively be "stacked" on unix/posix shells, whereas only a single environment at a time could be active in the cmd.exe shell.  Conda 4.4 unified the behavior, opting to keep effective behavior the same for unix shells, allowing cmd.exe to also have environments "stacked" on PATH.

The ideal situation is that, while a stack of active environments should be available to push onto and pop off of, only a single environment should be able to be active at any given time.

The purpose of this issue is to track introduction of a setting for `max_shlvl` of `0`.  The effect would be to have only one environment active at a time.  The proposal is to also make `0` the default value for `max_shlvl`.  Users are free to configure `max_shlvl` to `2` or more so that environments can be stacked on PATH as in conda 4.4 and 4.5.

kalefranz:
resolve #7194 

also contains: fix #7126 export CONDA_EXE #7206

This PR changes `activate` to the behavior advocated for by @mingwandroid and others. Namely, if a conda environment is currently active, a subsequent `conda activate` will replace on PATH the old environment for the new one.  The PR also adds a `--stack` flag to `conda activate` that supports the previous behavior.

To be clear, part of this PR is also removing the `max_shlvl` configuration parameter.  It's no longer needed.  It was originally introduced to keep users from unwittingly *stacking* many levels of environments beyond two.  Now, with the `--stack` flag, stacking becomes very intentional.  For users who have custom-configured `max_shlvl` in a .condarc, conda will now just ignore the configuration.


mingwandroid:
Looks good apart from an issue with the old script when attempting to follow `CONTRIBUTING.md` and upgrading from the Miniconda installed via `./dev/start.sh`:

```
. /Users/rdonnelly/conda/conda/devenv/bin/activate
'activate', 'deactivate', 'hook', or 'reactivate' command must be given
```

`conda activate` worked here instead but until we get the next `Miniconda` and `Anaconda` installers released we probably need the old stuff to work?

PR #7206, opened 2018-04-24
fix #7126 export CONDA_EXE

### LINKED ISSUE #7126: With conda being a shell function in 4.4+, what's the suggested way to use conda via scripts when an environment isn't explicitly activated by the user beforehand?

kalefranz:
resolve #7126


kalefranz:
incorporated into #7195

PR #7211, opened 2018-04-25
fix #7210 add suppressed --mkdir back to 'conda create'

### LINKED ISSUE #7210: canary: mkdir option removed from conda create on a minor bump
On conda 4.5

```bash
$ conda create -n testmkdir --mkdir
```
It works

On 4.6 on canary
```bash
$ conda create -n testmkdir --mkdir
```
It fails.

Removing an option from the CLI in a minor version sounds like a no no. Could we instead deprecate (even if the option does not do anything?)

kalefranz:
fix #7210


kalefranz:
Passes locally.  Not wasting CI time with this.

PR #7235, opened 2018-05-01
Update log.warn (deprecated) to log.warning

Carreau:
`log.warn` has been [deprecated](https://docs.python.org/3/library/logging.html#logging.Logger.warning) since the 2.x series. Use `log.warning`
instead.

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -48,7 +48,7 @@ def check_prefix(prefix, json=False):
         raise CondaValueError(error, json)
 
     if ' ' in prefix:
-        stderrlog.warn("WARNING: A space was detected in your requested environment path\n"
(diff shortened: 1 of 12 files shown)


mbargull:
I definitely don't oppose this change. When all of those `.warn`s are replaced, then https://github.com/conda/conda/blob/4.6.0a2/conda/gateways/logging.py#L159-L160 can be removed, too.

msarahan:
@Carreau sorry this got neglected.  Is it worth reopening and us pushing it over the line?

Carreau:
> Is it worth reopening and us pushing it over the line?

It's a relatively simple thing that it may make sens to restart from scratch. I'll force push and re-open if I find the time to do so. I'll keep it in the backburner. 

It may also be a nice simple first contribution for someone a little less comfortable than me in the Python ecosystem.

PR #7258, opened 2018-05-05
marker support

kalefranz:


conda/core/solve.py
```diff
@@ -538,6 +538,8 @@ def _prepare(self, prepared_specs):
             self.channels.update(additional_channels)
             reduced_index = get_reduced_index(self.prefix, self.channels,
                                               self.subdirs, prepared_specs)
+            _supplement_index_with_system(reduced_index)
+            import pdb; pdb.set_trace()
```
jjhelmus: looks like this got left in on accident.

jjhelmus:
I like the idea of virtual packages for system markers.  Do you have plans on how other packages could uses these to specify marker specific dependencies?

kalefranz:
The way we do the track_features stuff is with virtual/shadow packages. So this is really just generalizing that concept. 

I’m guessing we’ll probably extend the key-value part of MatchSpec, unless we can find a better way. The “package names” in the code here that end with “@“ I’d like to keep internal and not expose to users. 

Mike and I also discussed today how markers in conda world are themselves just MatchSpecs.

kalefranz:
```
enum34 ; python<3.4
- or -
enum34 if python<3.4
- turns into -
MatchSpec(“enum34”, if=[MatchSpec(python<3.4)])
```

kalefranz:
And now you can see why, if we implement markers as just an extension of MatchSpec, we need to inject the additional platform and system information into the solver index.

seibert:
I'm very interested in this as a way to create a virtual package corresponding to the version of the currently installed NVIDIA driver (assuming it is present).  This is needed to ensure we can set requirements on the cudatoolkit packages, which each require the nvidia driver to be >= some version.

(Right now the user has to force a particular cudatoolkit if their driver is not new enough to support the latest one in the Anaconda repository.)

Do you imagine these sort of "virtual package creators" to be pluggable, or do all of them have to be hardcoded into conda?

kalefranz:
I think the answer is probably hard-coded to start with, until we work out the patterns and what it all should look like. And then from there figure out how to extend it in a more pluggable way if that’s needed.

That said, do you know of any type of “gpuinfo.py” that’s equivalent to this cpuinfo.py package I found?

kalefranz:
cuda detection for markers from @seibert 

https://gist.github.com/seibert/52a204395cdc84eeeaf0ce05464a636b

> I've tried it on Win/Mac/Linux CUDA systems
> feel free to copy-paste this into the conda codebase
> (I wrote it from scratch, so you don't need a release form)
> takes between 100 and 300 ms to run the first time, so you should only call it when you need it
> Also, when you add markers, having an easy way to configure an override (maybe with conda config) will be important for both testing and supporting users when the automatic detection fails due to strange system configuration.
> (let me know if there is an issue or PR I should drop this information into)

travishathaway:

PR #7259, opened 2018-05-05
fix collisions from multichannel canonical name

kalefranz:


travishathaway:

Issue #7253, opened 2018-05-04
problem creating a prefixed environment on a fhgfs filesystem


karl616:
**I'm submitting a...**
  - [x] bug report
  - [ ] feature request

## Current Behavior

Creating a prefixed environment on a fhgfs partition does not work. It generates symbolic links to a temporary folder that doesn't exist. I can install on an ext4 partition without problem. In the example below I used xz, but I have observed this with other packages as well.

### Steps to Reproduce

```
bash Miniconda2-latest-Linux-x86_64.sh
conda create -p /tmp/conda/xz xz
ls -lh /tmp/conda/xz/bin/xz
#-rwxrwxr-x 1 user user 97K May  4 16:42 /tmp/conda/xz/bin/xz
conda create -p /DEEP_fhgfs/tmp/conda/xz xz
#lrwxrwxrwx 1 user user 62 May  4 16:43 /DEEP_fhgfs/tmp/conda/xz/bin/xz -> ../../../../tmp/tmpsl8kMX/23b991ec-39b8-48ac-bc89-79aaaeeef66c
df -Th /tmp /DEEP_fhgfs/tmp
#Filesystem     Type   Size  Used Avail Use% Mounted on
#/dev/md2       ext4   XXX  XXX  XXX   1% /local
#fhgfs_nodev    fhgfs  XXX  XXX  XXX  97% /DEEP_fhgfs

```

## Expected Behavior

The installation should complete similar to how it does on the ext4 partition

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : None
       user config file : /home/karln/.condarc
 populated config files : 
          conda version : 4.4.10
    conda-build version : not installed
         python version : 2.7.14.final.0
       base environment : /home/karln/miniconda2  (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/linux-64
                          https://repo.continuum.io/pkgs/main/noarch
                          https://repo.continuum.io/pkgs/free/linux-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/linux-64
                          https://repo.continuum.io/pkgs/r/noarch
                          https://repo.continuum.io/pkgs/pro/linux-64
                          https://repo.continuum.io/pkgs/pro/noarch
          package cache : /home/karln/miniconda2/pkgs
                          /home/karln/.conda/pkgs
       envs directories : /home/karln/miniconda2/envs
                          /home/karln/.conda/envs
[...]

PR #7268, opened 2018-05-08
path_action.ReplacePrefixLinkAction: if always_copy is set, copy

karl616:
instead of creating a hard link

suggested solution for  #7253 
perhaps it is too naive, but at least this allows me to create functional environments. It relies on always_copy instead of making a test similar to determine_link_type for each file. I cannot judge the impact of this change.


kalefranz:
Closing in lieu of #7269.

Issue #6406, opened 2017-12-06
conda 4.4 hangs after "solving environment"


michaelaye:
I followed instructions from here https://www.anaconda.com/blog/developer-blog/how-to-get-ready-for-the-release-of-conda-4-4/
to get conda 4.4

My conda history: Was using exclusively conda-forge (well, i had conda-forge on top of my channels, and the defaults below.

Executed commands to get conda 4.4:

```bash
$ conda config --add channels conda-canary
$ conda update conda
$ conda config --system --add pinned_packages conda-canary::conda
```

then put this in my .zprofile:

```bash
. /Users/klay6683/miniconda3/etc/profile.d/conda.sh
```
withOUT the recommended "conda activate base" because I like to have my default environment to be NOT root/base. Is that a problem? I thought it was frowned upon doing everday work in root/base? At least I corrupted more than one miniconda install doing so.

I have my iTerm2 profiles set up that each new terminal now executes "conda activate stable" and confirmed that it works correctly, e.g. paths are set correctly.

Then I executed a "conda update --all" in my main environment called "stable" (guess that's that? ;) 
and then I got feedback:
```bash
Solving environment: done (after a minuter or so)
[]
```
and then it's stuck at above box symbol.

How to debug?

michaelaye:
This seems to be caused by my anti-virus program checking (Avira). Deactivating it created an appropriate response time after solving the environment (order of 5 s for approx 290 packages).

matiasandina:
Deactivating Windows Defender also fix issue :)

jayspeidell:
Can confirm that Windows Defender blocks conda. Had same issue, deactivating Windows Defender fixed it.

HangsunKim:
Hi. I tried deactivating the Windows Defender and my other anti-virus program, but still I'm unable to do conda update.

aspiringguru:
Found this thread with same problem.
Have Win 10 home version 1709, gpedit.msc not enabled in windows home. (really? mistake purchase, hating ms right now)

this page provided a batch file to enable gpedit.msc
gpedit.msc now accessible

tried these instructions to disable windows defender.
https://www.windowscentral.com/how-permanently-disable-windows-defender-windows-10
Computer Configuration > Administrative Templates > Windows Components > Windows Defender > double-click Turn off Windows Defender.

exit powershell, new instance of powershell. (not sure if this is required but good practice to open new shell)

conda update conda 

[...]

PR #7275, opened 2018-05-09
Fix twofold tarball extraction, improve progress update

mbargull:
Tarballs are index-less archives, thus a call of `tarfile.TarFile.getmembers()` usually extracts the whole file, such that `tar_file.getmembers() ; tar_file.extractall()` extracts the file twice.
Because of that, users see a "stall" on the progress output at `75 %` for big tarballs when the package extraction starts. That lead them to cancel (CTRL+C) the process. When they re-run that `conda` command the extraction happens silently without any progress indication, i.e., another perceived stall. (This was reported some weeks ago -- need to find that issue again...)

This PR fixes the first part, that is, avoid the twofold extraction and progress bar stall, but not the silent extraction for previously interrupted extractions.
The first commit retained the progress update on a tarball member basis. That fixes the stalled progress output for big packages like with `conda create -yntmp mkl`. But it still lacks progress updates if the files inside the tarball also have large file sizes. A good example for that is `conda create -yntmp bioconda::gatk4`. To fix that case, I just added a wrapper to the file object that updates the progress bar on each `fileobj.read()` call.

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -5,7 +5,7 @@ from errno import EACCES, ELOOP, EPERM
 from io import open
 from logging import getLogger
 import os
-from os import X_OK, access
+from os import X_OK, access, fstat
 from os.path import basename, dirname, isdir, isfile, join, splitext
 from shutil import copyfileobj, copystat
 import sys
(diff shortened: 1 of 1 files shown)


mbargull:
Hm.. Can't seem to find an open issue for this. But at least there is a report in a comment of a closed issue:
@joseph-valles in https://github.com/conda/conda/issues/6406#issuecomment-369463406:
>After reinstalling Anaconda, I may have isolated the problem to this:
>
>1. `source activate base`
>2. `conda install cudatoolkit` The download should hang at 75%. Then press Ctrl-C.
>
>After that, conda should be slower to solve the environment and it will hang after that.
>In fact, conda hangs in a similar way after pressing Ctrl-C while installing *anything*.

That "being slower" and "hanging" is likely that silent extraction that this PR **does not** solve.

PR #7291, opened 2018-05-14
resolve #6845 add option -S / --satisfied-skip-solve to exit early for satisfied specs

### LINKED ISSUE #6845: Feature request: fast conda env update for satisfied dependencies
I would love a mode for `conda env update -f environment.yml` that functions a bit more like `pip install -r requirements.txt` in terms of dong less unneeded work. In particular, I would like it to avoid updating packages that don't need updating, and ideally avoid talking to the index at all if the environment.yml is already satisfied.

I think `update_dependencies: False` does *part* of this, but it doesn't apply to explicitly listed dependencies. In general, there appears to be no available operation to say "Given this environment.yml, make the fewest possible changes to satisfy its specification." My particular goals:

1. `conda env update` with an already satisfied environment.yml should install no packages, make no network requests, and in general complete as fast as possible.
2. if spec is partially satisfied, do not update packages that already satisfy their requirement, unless it is required to satisfy the packages that will be installed

I would like this behavior for `conda install` as well, such that I could call `conda install notebook` and it would only perform any actions if notebook is not up to date.

Are there existing combinations of flags that accomplish some or all of this?

kalefranz:
resolve #6845

```
  -S, --satisfied-skip-solve
                        Exit early and do not run the solver if the requested
                        specs are satisfied. Also skips aggressive updates as
                        configured by 'aggressive_update_packages'. Similar to
                        the default behavior of 'pip install'.
```

This PR also makes `aggressive_update_packages` even more aggressive, by overriding any user-requested specs in history.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -65,8 +65,8 @@ conda_build_test: &conda_build_test
           conda create -n blarg -yq --download-only python=2.7
           conda create -n blarg -yq --download-only python=3.4
           conda create -n blarg -yq --download-only python=3.5
-          conda create -n blarg -yq --download-only python=3.6 setuptools
-          conda create -n blarg -yq --download-only python setuptools
+          conda create -n blarg -yq --download-only python=3.6
+          conda create -n blarg -yq --download-only python setuptools cython
           conda create -n blarg -yq --download-only libpng=1.6.17
 
     - run:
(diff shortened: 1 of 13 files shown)


kalefranz:
Gets a `deprecation-breaking-change` tag because we're changing the signature of the `Solver` methods to disentangle the `DepsModifier` from the `UpdateModifier`.  It was just `DepsModifier` before, but we had conflated concepts that I didn't fully understand until now.  Since the API is still in beta, this change shouldn't be a problem.

Issue #5220, opened 2017-05-04
Test and report on state of conda with proxy servers


kalefranz:
Start by manually setting up a local proxy server. Google "docker image squid proxy".  There are multiple ways to configure these. Choosing a specific port to attach to. Using authentication or not. Using https or not. Etc. Before you even touch stuff with conda code, plink around just with some one-off python code that uses the requests and accessing the internet through the proxy. Make sure it's working and you understand how it's working. 

Then test conda against it. Follow the documentation on conda.io about using a proxy server with conda. Test all the different scenarios that are presented. First just figure out what works and what doesn't with 4.3.latest.  Post the report back here.

From memory, specific features that need tested:

* basic proxy server use, no proxy authentication, all conda traffic going through proxy server
* proxy server that requires a user to type in a password for access
* configuration of conda's proxy use via all three of (1) condarc configs, (2) .netrc files, (3) standard proxy configuration environment variables supported by requests

kalefranz:
Issue referencing proxy configuration environment variables: https://github.com/conda/conda/issues/3481

tyler-thetyrant:
[WIP] Testing info. 
Using squid from [here](https://hub.docker.com/r/poklet/squid/).

```
Current conda install:

                   platform : osx-64
              conda version : 4.3.17
           conda is private : False
          conda-env version : 4.3.17
        conda-build version : 2.1.9
             python version : 2.7.13.final.0
           requests version : 2.13.0
           root environment : /Users/tkim/miniconda3  (writable)
        default environment : /Users/tkim/miniconda3
           envs directories : /Users/tkim/miniconda3/envs
                          /Users/tkim/.conda/envs
              package cache : /Users/tkim/miniconda3/pkgs
                          /Users/tkim/.conda/pkgs
               channel URLs : https://repo.continuum.io/pkgs/free/osx-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/osx-64
                          https://repo.continuum.io/pkgs/r/noarch
                          https://repo.continuum.io/pkgs/pro/osx-64
                          https://repo.continuum.io/pkgs/pro/noarch
                          https://conda.anaconda.org/anaconda-fusion/osx-64
                          https://conda.anaconda.org/anaconda-fusion/noarch
                config file : /Users/tkim/.condarc
               offline mode : False
[...]

PR #7308, opened 2018-05-18
proxy test routine

kalefranz:
Just adds a test script for conda interaction with proxy servers.  Invoke with

    bash ./tests/integration/proxy/proxy-test.sh


kalefranz:
Toward #5220.

PR #7314, opened 2018-05-19
Fix MSYS2 path conversion in the face of bind mounts and /etc/fstab

mingwandroid:
There is a subtle issue that is caused by Cygwin/MSYS2's fstab where it
contains a bind mount /bin => /usr/bin. With cygpath in /Library/usr/bin
(translated to /usr/bin) this bind mount is translated to /bin and thus
/Library/usr/bin is shadowed by /usr/bin. To address this we translate
the prefix root and then just append the individual paths to the result
of that thus preventing any and all fstab-related short-circuiting

### DIFF
--- a/conda/cli/activate.py
+++ b/conda/cli/activate.py
@@ -89,7 +89,8 @@ def binpath_from_arg(arg, shell, going_to_shell=True):
     prefix = prefix_from_arg(arg, shell)
     # convert paths to shell-native paths
     if going_to_shell:
-        return [shelldict['path_to'](path) for path in _get_prefix_paths(prefix)]
+        return [shelldict['path_to'](path, root_prefix=prefix)
+                for path in _get_prefix_paths(prefix)]
     else:
         return [path for path in _get_prefix_paths(prefix)]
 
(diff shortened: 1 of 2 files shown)


kalefranz:
Also just skimming this on my phone, and kinda hard to grep the context. I’ll look at porting it to conda/activate.py later.

mingwandroid:
Urgh, somehow I managed to not commit and then lose the changes to activate.py! Closing until I sort this mess out (I need it to work on 4.5.x branch too though).

kalefranz:
I don’t think conda-build is importing anything from cli/activate.py. The shell level `source activate` which I think is all that conda-build uses, goes through conda/activate.py.

mingwandroid:
OK well, I did fix up both but managed to lose it via a bad `cp` command, will sort this out. Please detail your context issues if you don't grok my initial comment. It's subtle but very important!

kalefranz:
I’m guessing I’ll get it. Just only have a mobile phone right now, and not easy to review and understand code unless everything is immediately obvious :)

PR #7316, opened 2018-05-19
Fix MSYS2 path conversion in the face of bind mounts and /etc/fstab

mingwandroid:
There is a subtle issue that is caused by Cygwin/MSYS2's fstab where it
contains a bind mount /bin => /usr/bin. With cygpath in /Library/usr/bin
(translated to /usr/bin) this bind mount is translated to /bin and thus
/Library/bin (contains native Windows software) is shadowed by
Library/usr/bin (contains MSYS2 software). To address this we translate
the prefix root and then just append the individual paths to the result
of that thus preventing this unfortunate fstab-related short-circuiting.


mingwandroid:
I made this PR mainly to see how CI reacts to it.

Force pushing to a closed branch is to a good thing to do ..

Oh well, here it is again with the missing bits (needed a good bit of rework unfortunately and so now needs a good bit more testing).

Issue #6820, opened 2018-02-01
Consider shell-agnostic activate.d/deactivate.d mechanism


kalefranz:
Currently `activate.d` and `deactivate.d` directories contain files meant to be sourced by the currently-active shell.  E.g. `.sh` files for bash shells; `.bat` files for cmd.exe shells.  In those two files alone there's potential need for duplicated logic.  It's also a common problem that `.sh` files are often naively written for bash, but should in fact be written for posix shell compliance.  These shell-specific files also completely neglect powershell, xonsh, tcsh, fish, and any other future shells we might easily support for conda with a simple shell wrapper addition.

We therefore need a shell-agnostic method for specifying changes to process environment variables during environment activation and deactivation.

kalefranz:
Right now I see three general options

1. Putting .py files in `activate.d`/`deactivate.d` that contain convention-defined function names with convention-defined function signatures that can be called directly within conda's python process.
2. Using subprocess calls to generic executables in `activate.d`/`deactivate.d`, which use one the passed environment variables as inputs, and are expected to output json of some documented form.
3. To hedge against #6819, and the possibility that conda might not have a strict python dependency in the future, a variation option 1 that instead uses lua as an interpreter.  Lua is a <200kB package that has only readline and ncurses as dependencies.  Python is a 27MB package with quite a few more dependencies.  This would mean, however, the lua, along with python, would be hard dependencies of conda.

njalerikson:
Is there ever a use case for these scripts that is activating/deactivating something not related to setting/unsetting environment variables? With the current design more or less anything can be done within those scripts. I can imagine some interesting use cases of activate scripts calling lmod and I do not see a clear way to preserve such information via the first two ideas.

The third idea is interesting. I did just mention lmod so maybe...

kalefranz:
I had to google lmod. https://lmod.readthedocs.io/en/latest/

njalerikson:
After thinking about this and discussing with coworkers I think you should consider the option to integrate LMOD for the use of these `{de,}activate.d` scripts.

Just out of the box it would support a good number of shells, quoting the [TACC page](https://www.tacc.utexas.edu/research-development/tacc-projects/lmod):
> All the popular shells are supported: bash, ksh, csh, tcsh, zsh. LMOD is also available for perl and python.

This list of supported shells is clearly missing Windows support but since Lua is supported on Windows it would just be a matter of investing the time to add Windows abilities to LMOD.

[...]

PR #7320, opened 2018-05-20
resolve #2682 add conda run experimental support

### LINKED ISSUE #2682: conda run command removed in 4.1.0
After upgrading to 4.1.0:

```
$ conda run
```

outputs:

```
Error: Could not locate 'conda-run'
```

Also, when running `conda --help`, the documentation about the run command is no longer present.

kalefranz:
resolve #2682

Proof of concept code for `conda run`, implemented at the python level.  So far, only implemented on unix.  To test on unix

1. check out repo
2. `. dev/start`
3. `conda run python -- -V`
   Will show python version.
4. `mkdir -p "$CONDA_PREFIX/etc/conda/activate.d"`
   `echo "export TESTVAR=123456" > "$CONDA_PREFIX/etc/conda/activate.d/test.sh"`
   `conda run python -- -c "import os; print(os.environ['TESTVAR'])"`
   Will print `123456` to stdout.

I used the command `conda run` here instead of `conda exec` because we historically already "own" `conda run`, and I think there's already a `conda-exec` extension project that we would clobber.

This pretty much implements the old `conda run`, except we don't assume that users want to "run a package," but instead users know the name of the executable they want to run.

We may want to think of ways to extend the capabilities here.  For example,

    conda run /path/to/conda/environment/bin/python -- -V

should also maybe "just work" in addition to

    conda run -p /path/to/conda/environment python -- -V

Or maybe not.

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from argparse import (ArgumentParser as ArgumentParserBase, RawDescriptionHelpFormatter, SUPPRESS,
-                      _CountAction, _HelpAction)
+                      _CountAction, _HelpAction, ZERO_OR_MORE, REMAINDER, ONE_OR_MORE)
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
@@ -964,14 +964,20 @@ def configure_parser_run(sub_parsers):
     )
 
     add_parser_prefix(p)
-    add_parser_json(p)
-
     p.add_argument(
-        'executable_name',
-        action="store",
-        help="Executable name.",
+        "-v", "--verbose",
+        action=NullCountAction,
+        help="Use once for info, twice for debug, three times for trace.",
+        dest="verbosity",
+        default=NULL,
     )
 
+    p.add_argument(
+        'executable_call',
+        nargs=ONE_OR_MORE,
+        help="Executable name, with additional arguments to be passed to the executable "
+             "on invocation.",
+    )
     p.set_defaults(func='.main_run.execute')
 
 
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from argparse import (ArgumentParser as ArgumentParserBase, RawDescriptionHelpFormatter, SUPPRESS,
-                      _CountAction, _HelpAction, ZERO_OR_MORE, REMAINDER, ONE_OR_MORE)
+from argparse import (ArgumentParser as ArgumentParserBase, ONE_OR_MORE,
+                      RawDescriptionHelpFormatter, SUPPRESS, _CountAction, _HelpAction)
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
@@ -140,13 +140,7 @@ class ArgumentParser(ArgumentParserBase):
                             raise CommandNotFoundError(cmd)
                         args = [find_executable('conda-' + cmd)]
                         args.extend(sys.argv[2:])
-                        p = Popen(args)
-                        try:
-                            p.communicate()
-                        except KeyboardInterrupt:
-                            p.wait()
-                        finally:
-                            sys.exit(p.returncode)
+                        _exec(args, os.environ)
 
         super(ArgumentParser, self).error(message)
 
@@ -161,6 +155,24 @@ class ArgumentParser(ArgumentParserBase):
                 builder.append("conda commands available from other packages:")
                 builder.extend('  %s' % cmd for cmd in sorted(other_commands))
                 print('\n'.join(builder))
+
+
+def _exec(executable_args, env_vars):
+    return (_exec_win if on_win else _exec_unix)(executable_args, env_vars)
+
+
+def _exec_win(executable_args, env_vars):
+    p = Popen(executable_args, env=env_vars)
+    try:
+        p.communicate()
+    except KeyboardInterrupt:
+        p.wait()
+    finally:
+        sys.exit(p.returncode)
+
+
+def _exec_unix(executable_args, env_vars):
+    os.execvpe(executable_args[0], executable_args, env_vars)
 
 
 class NullCountAction(_CountAction):
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from argparse import (ArgumentParser as ArgumentParserBase, ONE_OR_MORE,
-                      RawDescriptionHelpFormatter, SUPPRESS, _CountAction, _HelpAction)
+from argparse import (ArgumentParser as ArgumentParserBase, REMAINDER, RawDescriptionHelpFormatter,
+                      SUPPRESS, _CountAction, _HelpAction)
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
@@ -986,7 +986,7 @@ def configure_parser_run(sub_parsers):
 
     p.add_argument(
         'executable_call',
-        nargs=ONE_OR_MORE,
+        nargs=REMAINDER,
         help="Executable name, with additional arguments to be passed to the executable "
              "on invocation.",
     )
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -68,17 +68,17 @@ def _main(*args, **kwargs):
     if len(args) == 1:
         args = args + ('-h',)
 
-    try:
-        idx = args.index("--")
-    except ValueError:
-        extra_args = ()
-    else:
-        extra_args = args[idx+1:] if idx < len(args)-1 else ()
-        args = args[:idx]
-
     p = generate_parser()
-    args = p.parse_args(args[1:])
-    args.extra_args = extra_args
+    args, unknown_args = p.parse_known_args(args[1:])
+    if unknown_args:
+        if args.cmd == "run":
+            args.unknown_args = unknown_args
+        else:
+            from ..exceptions import ArgumentError
+            raise ArgumentError("Unknown arguments for command '%(command)s': %(unknown_args)s",
+                                command=args.cmd, unknown_args=unknown_args)
+    else:
+        args.unknown_args = []
 
     from ..base.context import context
     context.__init__(argparse_args=args)
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -69,16 +69,7 @@ def _main(*args, **kwargs):
         args = args + ('-h',)
 
     p = generate_parser()
-    args, unknown_args = p.parse_known_args(args[1:])
-    if unknown_args:
-        if args.cmd == "run":
-            args.unknown_args = unknown_args
-        else:
-            from ..exceptions import ArgumentError
-            raise ArgumentError("Unknown arguments for command '%(command)s': %(unknown_args)s",
-                                command=args.cmd, unknown_args=unknown_args)
-    else:
-        args.unknown_args = []
+    args = p.parse_args(args[1:])
 
     from ..base.context import context
     context.__init__(argparse_args=args)
--- a/conda/cli/main_run.py
+++ b/conda/cli/main_run.py
@@ -36,8 +36,10 @@ def _get_activated_env_vars_win(env_location):
             tf.write(ensure_binary(
                 "@%CONDA_PYTHON_EXE% -c \"import os, json; print(json.dumps(dict(os.environ)))\""
             ))
+        # TODO: refactor into single function along with code in conda.core.link.run_script
         cmd_builder = [
-            "cmd.exe /K \"",
+            "%s" % os.getenv('COMSPEC', 'cmd.exe'),
+            "/C \"",
             "@SET PROMPT= ",
             "&&",
             "@SET CONDA_CHANGEPS1=false"
@@ -83,4 +85,4 @@ def _get_activated_env_vars_unix(env_location):
 def execute(args, parser):
     from .conda_argparse import _exec
     env_vars = get_activated_env_vars()
-    _exec(args.executable_call + args.unknown_args, env_vars)
+    _exec(args.executable_call, env_vars)
--- a/conda/cli/main_run.py
+++ b/conda/cli/main_run.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import json
 import os
-from os.path import isfile, join
 from subprocess import Popen
 import sys
 from tempfile import NamedTemporaryFile
@@ -94,46 +93,44 @@ def get_activated_env_vars():
     return env_var_map
 
 
-def find_executable(executable_name):
-    target_prefix = context.target_prefix
-    if on_win:
-        executable_path = _find_executable_win(target_prefix, executable_name)
-    else:
-        executable_path = _find_executable_unix(target_prefix, executable_name)
-    if executable_path is None:
-        raise ExecutableNotFound(target_prefix, executable_name)
-    return executable_path
+# def find_executable(executable_name):
+#     target_prefix = context.target_prefix
+#     if on_win:
+#         executable_path = _find_executable_win(target_prefix, executable_name)
+#     else:
+#         executable_path = _find_executable_unix(target_prefix, executable_name)
+#     if executable_path is None:
+#         raise ExecutableNotFound(target_prefix, executable_name)
+#     return executable_path
+#
+#
+# def _find_executable_win(target_prefix, executable_name):
+#     from ..activate import _Activator
+#     pathext = tuple(os.environ["PATHEXT"].split(';'))
+#     if executable_name.endswith(pathext):
+#         for path_dir in _Activator._get_path_dirs(target_prefix):
+#             executable_path = join(path_dir, executable_name)
+#             if isfile(executable_path):
+#                 return executable_path
+#     else:
+#         for path_dir in _Activator._get_path_dirs(target_prefix):
+#             for ext in pathext:
+#                 executable_path = join(path_dir, executable_name + ext)
+#                 if isfile(executable_path):
+#                     return executable_path
+#     return None
+#
+#
+# def _find_executable_unix(target_prefix, executable_name):
+#     executable_path = join(target_prefix, 'bin', executable_name)
+#     if isfile(executable_path) and os.access(executable_path, os.X_OK):
+#         return executable_path
+#     return None
 
 
-def _find_executable_win(target_prefix, executable_name):
-    from ..activate import _Activator
-    pathext = tuple(os.environ["PATHEXT"].split(';'))
-    if executable_name.endswith(pathext):
-        for path_dir in _Activator._get_path_dirs(target_prefix):
-            executable_path = join(path_dir, executable_name)
-            if isfile(executable_path):
-                return executable_path
-    else:
-        for path_dir in _Activator._get_path_dirs(target_prefix):
-            for ext in pathext:
-                executable_path = join(path_dir, executable_name + ext)
-                if isfile(executable_path):
-                    return executable_path
-    return None
-
-
-def _find_executable_unix(target_prefix, executable_name):
-    executable_path = join(target_prefix, 'bin', executable_name)
-    if isfile(executable_path) and os.access(executable_path, os.X_OK):
-        return executable_path
-    return None
-
-
-def _exec_win(executable_path, extra_args=(), env_vars=None):
+def _exec_win(executable_args, env_vars=None):
     env_vars = os.environ.copy() if env_vars is None else env_vars
-    args = [executable_path]
-    args.extend(extra_args)
-    p = Popen(args, env=env_vars)
+    p = Popen(executable_args, env=env_vars)
     try:
         p.communicate()
     except KeyboardInterrupt:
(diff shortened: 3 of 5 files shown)


conda/cli/main_run.py
```diff
@@ -0,0 +1,123 @@
...
+import json
+import os
+from os.path import isfile, join
+
+from ..base.context import context
+from ..common.compat import on_win
+from ..gateways.subprocess import subprocess_call
+
+# OLD conda-run help
+"""
```
kalefranz: This is the old `conda run --help` pulled from the 3.x branch.  The old `conda run` for whatever reason assumed that there would only be a single executable per package, and so you would "run a package."

mcg1969:
I'm not sure I agree with the approach here. Suppose I were to create a shell script to do this; it would likely look like this:
```
#!/bin/bash
_CONDA_ROOT=<whatever>
. "$_CONDA_ROOT/etc/profile.d/conda.sh"
conda activate $1
shift
exec $@
```
The `conda activate` involves a call to `conda shell.posix` and the sourcing of activate scripts. So we have the surrounding bash call, a single python call, and then the final exec.

In this case, you have the python call for `conda run` itself, which then calls a programmatically generated shell script, which in turn makes three more python calls, and then _returns_ for the final exec.

Seems like a better chain of events is this. From _within_ the `conda run` call itself, call the methods invoked by `conda shell.posix hook` and `conda shell.posix activate`. Build a script that includes those outputs as well as the command to be executed itself. Then send _that_ bash call to `execve`. That way, you're back to a single Python call and a single bash call.

Or perhaps we save ourselves the trouble and just ship the above script, and call it `conda-run`?

kalefranz:
As I read it, what you propose will leave a running bash process in the process tree for the process being created. Once the `executable_name` process is up and running, the patent process should be the process in which `conda run` was called. That is, you should be able to write an init.d script using `conda run` where a shell doesn’t end up in the process tree, and you should be able to use `conda run` to make the `executable_name` process PID 1 in a Docker container.

mcg1969:
I don't think that has to be the case. If Python can `execve` the bash script, then that script can in turn `exec` the command to be run.

mcg1969:
I agree that’s a useful requirement (preserving the PID, no extra sustained processes).

kalefranz:
@mcg1969 mentioned we should experiment with using `conda run` as a shebang line.

mcg1969:
This seems to work, but interestingly, just using the exact path to `conda-run` instead of `/usr/bin/env conda-run` does not.
```
#!/usr/bin/env conda-run exactglm python
from __future__ import print_function
import os, sys
print('Prefix: {}'.format(sys.prefix))
print('PID: {}'.format(os.getpid()))
```

mcg1969:
I think the reason why `#!/Users/mgrant/bin/conda-run` doesn't work directly is because script can't serve as a shebang processor. So this works:
```
#!/bin/sh conda-run portfolio python
from __future__ import print_function
import os, sys
print('Prefix: {}'.format(sys.prefix))
print('PID: {}'.format(os.getpid()))
```

mcg1969:
(I added the `from __future__` just in case it wasn't working properly and it dumped me back to system python ;-))

mbargull:
Only took a very quick look at the changes and you discussion yet. Thoughts sofar:
- :+1: for `exec`ing to replace the current process.
- :+1: for @mcg1969's idea to reduce the amount of processes involved, i.e., let the first Python process create and `exec` a script that does the activation and `exec`s the requested program.
  BUT: This can be considered an implementation detail and added later on, not necessarily in this PR.
- :+1: to implement this in Python (for flexibility and to avoid more shell-specific shenanigans).

- :-1: to try to use this in shebangs (at least not without additional help like https://github.com/shlevy/long-shebang).
  Shebangs are far to brittle/inflexible since they are not fit (designed) for many modern uses.
  Extremely helpful source: https://www.in-ulm.de/~mascheck/various/shebang/#results
  @mcg1969 encountered the limitations on macOS. Linux has others (127 char limit, argv[2:] not split, i.e., only one argument possible).
- :+1: on making it dumber as the original `conda run`, i.e., not doing the package related stuff.
- :-1: on not making it even dumber :wink:.
- :-1: on the `--` requirement.
  `nargs=argparse.REMAINDER` exists for exactly those use cases.
  Using that, you could just run `conda run -p env-path prog args`.

Regarding that "even dumber" point:
I would expect `conda run -p env-path prog args` to be equivalent to
```sh
(conda activate env-path && exec prog args)
```
meaning, I wouldn't expect `conda run` to check for `prog` to exist in the target environment, but just run the `prog` that is available once `env-path` is activated. That way you could also run foreign application in that evironment, e.g., `conda run which python`, being usually equivalent to `conda run /usr/bin/which python`, which returns the path of the `python` binary that would get called from within that environment. Apart from being more flexible, this would also avoid doing those existence/executable checks manually (instead, e.g., use `exevpe` if running from Python).
If you disagree and like the "`prog` must be in environment" more, that behavior could be made optional with a flag (something like `--env-only`/`--no-env-only`, depending on the default).

kalefranz:
> for `exec`ing to replace the current process.

You'll notice we're not doing that on Windows.  I tried switching the conda command plugin mechanism we have (i.e. the entry point `bin/conda-build` is the `conda build` command) to `exec` in conda 4.4

https://github.com/conda/conda/blob/4.4.0/conda/cli/conda_argparse.py#L137

and I had to switch it back to subprocess because of unexpected (for me) Windows behavior.

kalefranz:
> on the `--` requirement

Should we do both?  Use `nargs=argparse.REMAINDER`, but also keep `--` for the case when arguments need to be duplicated?

Or...

Maybe we should take all of these arguments away from `conda run`:

```
Output, Prompt, and Flow Control Options:
  --json                Report all output as json. Suitable for using conda
                        programmatically.
  -v, --verbose         Use once for info, twice for debug, three times for
                        trace.
  -q, --quiet           Do not display progress bar.
```

which leaves the only arguments left `-n` and `-p`.  Still could have some use cases where you'd want to send to both though I guess?

mbargull:
> You'll notice we're not doing that on Windows.

Yes, intentionally ignored that one :wink:. A while back I learned about this not working properly on Windows.. (I think while looking at `menuinst`'s `cwp.py`?)

> Should we do both? Use `nargs=argparse.REMAINDER`, but also keep `--` for the case when arguments need to be duplicated?

With `nargs=argparse.REMAINDER` you essentially partition the argument list, i.e., everything before that argument is parsed normally, but all following arguments are appended to the value of that `nargs=REMAINDER` argument without being parsed anymore. You'll have something like
```python
a = ArgumentParser()
a.add_argument('-p')
a.add_argument('exe', nargs=REMAINDER)
a.parse_args(['-p', 'some-path', 'some-exe', '-p', 'some-other-path'])
# returns: {'p': 'some-path', 'exe': ['some-exe', '-p', 'some-other-path']}
```
Hence duplicate arguments should be no problem.
I think having the `--` between the `some-exe` and its arguments is very awkward. I'd rather expect it to be used as `conda run -p env-path -- prog args`. I'm not entirely sure if `argparse` allows the `--` by itself or whether you have to add a dummy argument for this. I'll check.

> Maybe we should take all of these arguments away from `conda run`:

Yeah, maybe. I think `-v` could be useful if we want to output debug info before the `exec`. `-q` I don't know -- I'd expect `conda run` to be quietly launching anyway, so this is likely a no-op. `--json` could output the `-v` info in JSON format -- but the rest of the output is likely non-JSON since it comes from the executed program; not sure how useful this is...

kalefranz:
> let the first Python process create and `exec` a script that does the activation and `exec`s the requested program

If this were the final state, I would agree.  However I see all the `get_activated_env_vars()` as a huge hack that lets us ship this feature while still not having #6820 both (1) done in conda, and (2) broadly deployed throughout the package base.  So the code right now looks closer to the final ideal state for when we can assume #6820.  I could maybe still be convinced to go along with pushing more into a script.  Still thinking about it.

mbargull:
> I could maybe still be convinced to go along with pushing more into a script. Still thinking about it.

IMHO, it's just an implementation level detail whose benefit would likely "just" be faster execution. And we are talking about something in the realm of having a 0.1 or 0.2 second execution time, I believe (haven't tried it out yet). If I were you, I'd just leave it as is for now, I think :wink: 

----
As for
> I'm not entirely sure if `argparse` allows the `--` by itself or whether you have to add a dummy argument for this. I'll check.

It looks like you could use `conda run -p env-path -- prog args` with `nargs=REMAINDER`, but you'll have to remove `--` manually from the argument value, i.e., use some like
```python
if not args.executable_name:
    # complain b/c argument is missing
elif args.executable_name[0] == '--':
    del args.args.executable_name[0]
```

See the last output block from the following:
<details open><summary>

```python
#!/usr/bin/env python

from argparse import *

def error(self, *args):
    raise Exception('error', *args)
ArgumentParser.error = error

parsers = []

p = ArgumentParser()
p.add_argument('-a')
p.add_argument('-b', nargs=ZERO_OR_MORE)
p.add_argument('e', nargs=ONE_OR_MORE)
parsers.append(p)

p = ArgumentParser()
p.add_argument('-a')
p.add_argument('-b', nargs=ZERO_OR_MORE)
p.add_argument('e', nargs=ONE_OR_MORE)
parsers.append(p)

p = ArgumentParser()
p.add_argument('-a')
p.add_argument('-b', nargs=ZERO_OR_MORE)
p.add_argument('-', dest='_', action='store_true', help=SUPPRESS)
p.add_argument('e', nargs=ONE_OR_MORE)
parsers.append(p)

p = ArgumentParser()
p.add_argument('-a')
p.add_argument('-b', nargs=ZERO_OR_MORE)
p.add_argument('e', nargs=REMAINDER)
parsers.append(p)

p = ArgumentParser()
p.add_argument('-a')
p.add_argument('-b', nargs=ZERO_OR_MORE)
p.add_argument('-', dest='_', action='store_true', help=SUPPRESS)
p.add_argument('e', nargs=REMAINDER)
parsers.append(p)

for p in parsers:
    for args in (
        '-a 1 -b 2 3             ',
        '-a 1 -b 2 3         -  E',
        '     -b 2 3    -a 1 -  E',
        '     -b 2 3 -- -a 1 -  E',
        '-a 1 -b 2 3         -- E',
        '     -b 2 3    -a 1 -- E',
        '     -b 2 3 -- -a 1 -- E',
    ):
        try:
            n = p.parse_args(args.split())
            print(args, '|', *('{}={:<10}'.format(a, str(getattr(n, a))) for a in ('a', 'b', 'e')))
        except Exception as e:
            print(args, '|', e)
    print()
```
</summary>

```
-a 1 -b 2 3              | ('error', 'the following arguments are required: e')
-a 1 -b 2 3         -  E | ('error', 'the following arguments are required: e')
     -b 2 3    -a 1 -  E | a=1          b=['2', '3'] e=['-', 'E']
     -b 2 3 -- -a 1 -  E | a=None       b=['2', '3'] e=['-a', '1', '-', 'E']
-a 1 -b 2 3         -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3    -a 1 -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3 -- -a 1 -- E | a=None       b=['2', '3'] e=['-a', '1', '--', 'E']

-a 1 -b 2 3              | ('error', 'the following arguments are required: e')
-a 1 -b 2 3         -  E | ('error', 'the following arguments are required: e')
     -b 2 3    -a 1 -  E | a=1          b=['2', '3'] e=['-', 'E']
     -b 2 3 -- -a 1 -  E | a=None       b=['2', '3'] e=['-a', '1', '-', 'E']
-a 1 -b 2 3         -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3    -a 1 -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3 -- -a 1 -- E | a=None       b=['2', '3'] e=['-a', '1', '--', 'E']

-a 1 -b 2 3              | ('error', 'the following arguments are required: e')
-a 1 -b 2 3         -  E | a=1          b=['2', '3'] e=['E']     
     -b 2 3    -a 1 -  E | a=1          b=['2', '3'] e=['E']     
     -b 2 3 -- -a 1 -  E | a=None       b=['2', '3'] e=['-a', '1', '-', 'E']
-a 1 -b 2 3         -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3    -a 1 -- E | a=1          b=['2', '3'] e=['E']     
     -b 2 3 -- -a 1 -- E | a=None       b=['2', '3'] e=['-a', '1', '--', 'E']

-a 1 -b 2 3              | a=1          b=['2', '3'] e=[]        
-a 1 -b 2 3         -  E | a=1          b=['2', '3', '-', 'E'] e=[]        
     -b 2 3    -a 1 -  E | a=1          b=['2', '3'] e=['-', 'E']
     -b 2 3 -- -a 1 -  E | a=None       b=['2', '3'] e=['--', '-a', '1', '-', 'E']
-a 1 -b 2 3         -- E | a=1          b=['2', '3'] e=['--', 'E']
     -b 2 3    -a 1 -- E | a=1          b=['2', '3'] e=['--', 'E']
     -b 2 3 -- -a 1 -- E | a=None       b=['2', '3'] e=['--', '-a', '1', '--', 'E']

-a 1 -b 2 3              | a=1          b=['2', '3'] e=[]        
-a 1 -b 2 3         -  E | a=1          b=['2', '3'] e=['E']     
     -b 2 3    -a 1 -  E | a=1          b=['2', '3'] e=['E']     
     -b 2 3 -- -a 1 -  E | a=None       b=['2', '3'] e=['--', '-a', '1', '-', 'E']
-a 1 -b 2 3         -- E | a=1          b=['2', '3'] e=['--', 'E']
     -b 2 3    -a 1 -- E | a=1          b=['2', '3'] e=['--', 'E']
     -b 2 3 -- -a 1 -- E | a=None       b=['2', '3'] e=['--', '-a', '1', '--', 'E']
```
</details>

mcg1969:
I think I still need convincing here that the approach being taken for activation is legitimate. @mbargull, have you reviewed with care the specific approach Kale is doing here? 

Specifically: what this is doing is constructing a script that activates the environment and then runs a program. But the program that is being run inside this script is _not_ the one the user specifies. Instead, it's a python one-line that dumps a JSON representation of the resulting environment variables to standout, to be read back in by this process.

_After that_, he sets those environment variables and executes the _requested_ program directly from python.

For one thing, I'm still a little worried that it's really not equivalent to `conda activate <env> && exec $@`. Maybe I'm wrong about that. I certainly do not yet want to make the assumption that the only thing activate scripts are allowed to do is set environment variables; for instance, I have an application where I'd like the activate script to make some filesystem changes. I know that may be frowned upon, but this is not the place to tell me I can't do that---we should have that discussion in a larger forum. (And if I'm outvoted in that larger forum, I'll accept the verdict and find another way!)

Secondly, if we are indeed going through the trouble to construct this "activation plus run a program" script, why not just _put the user's requested program_ in that script instead? I'll point out what I'm talking about in line comments.

conda/cli/main_run.py
```diff
@@ -0,0 +1,162 @@
...
+            ))
+        cmd_builder = [
+            "cmd.exe /K \"",
+            "@SET PROMPT= ",
+            "&&",
+            "@SET CONDA_CHANGEPS1=false"
+            "&&",
+            "@CALL {0} activate \"{1}\"".format(conda_bat, env_location),
+            "&&",
+            "\"{0}\"".format(tf.name),
```
mcg1969: Why can't we put the user's requested program _right here_, or perhaps inside the temporary script you're creating above?

conda/cli/main_run.py
```diff
@@ -0,0 +1,162 @@
...
+    except KeyError:
+        raise CommandNotFoundError("run")
+
+    cmd_builder = [
+        "sh -c \'"
+        "eval \"$(\"{0}\" shell.posix hook)\"".format(conda_exe),
+        "&&",
+        "conda activate \"{0}\"".format(env_location),
+        "&&",
+        "\"$CONDA_PYTHON_EXE\" -c \"import os, json; print(json.dumps(dict(os.environ)))\"",
```
mcg1969: Likewise, for Unix: why not put the user command in this line? We have the environment activated and ready to go. Let's use it!

kalefranz: Haha makes sense.  This code was first pulled from https://github.com/tox-dev/tox-conda/pull/1, and I was planning on porting it back once we had this up and working.  If we do as you suggest, it's not as composable and reusable.  I don't mind doing as you suggest though, after we get this working for the other use case.

kalefranz:
> I'm still a little worried that it's really not equivalent to `conda activate <env> && exec $@`. Maybe I'm wrong about that.

I believe it's exactly equivalent.  Can you describe in what case it wouldn't be?  Maybe I'm missing something...

> for instance, I have an application where I'd like the activate script to make some filesystem changes.

We're still calling activate exactly once, so if you're putting side effects in `activate.d` scripts, those side effects will still apply.

> I know that may be frowned upon, but this is not the place to tell me I can't do that---we should have that discussion in a larger forum.

I do, and we should.  Or...  I'd at least like to understand the use cases where such side effects are desired.  I so far haven't been able to think of any, and I've also asked this exact question to other people.  Am still in search of *any* affirmative situation where someone wanted disk-mutating side effects, whether it's a good design or not.

kalefranz:
> I think I still need convincing here that the approach being taken for activation is legitimate.

@mcg1969, I do believe the approach is legitimate.  Subprocessing to `/bin/sh` or `cmd.exe` is unfortunately but cleanly the requirement of having `activate.d` scripts be `.sh` or `.bat`.  We *do* execute each script, and if they have side effects other than environment variable changes, those side effects *will* be applied.

The optimization to call `exec` from the composed script isn't as clean, and I also don't think it buys us much.  Right now on unix, we fork a python process, subprocess to a generated script to activate, and then fall back to the python process to exec to the target executable.  The alternative you suggest is to fork a python process, exec to a generated script to activate, and then exec again to the target executable.  I would be shocked if that optimization gains more than 50 ms.  The process flow as coded here so far is the process flow we'd ultimately want to get to once we have #6820.

mbargull:
> Subprocessing to `/bin/sh` or `cmd.exe` is unfortunately but cleanly the requirement of having `activate.d` scripts be `.sh` or `.bat`.

Yes, that one is unavoidable currently.

> We *do* execute each script, and if they have side effects other than environment variable changes, those side effects *will* be applied.

Yes, I can't think of any relevant changes that won't be applied. Some changes like function definitions or `set` but not `export`ed variables (activation scripts are `source`d!) won't survive with the current approach -- but these are irrelevant either way since we want to `exec` the target. (IMO, `conda run` really doesn't have to guarantee to be able to execute an equivalent of obscure cases like `(conda activate env && some_shell_func_from_activate_script)` and a `conda run -p /env0 conda activate /env1` doesn't make much sense either.)

> The alternative you suggest is to fork a python process, exec to a generated script to activate, and then exec again to the target executable.

I think we have these options:
```
#1
conda run (python)
|\
| (shell)
| |\
| | conda shell.posix activate (python)
| |/
| |- eval activate scripts (shell, same process)
| |\
| | python -c 'import os, json; print(json.dumps(dict(os.environ)))'
| |/
|/
exec target

#2
python (conda run)
|\
| (shell)
| |\
| | conda shell.posix activate (python)
| |/
| |- eval activate scripts (shell, same process)
| |
| exec python -c 'import os, json; print(json.dumps(dict(os.environ)))'
| |
|/
exec target

#3
conda run (python)
|
|\
| conda shell.posix activate (python)
|/
|- add 'exec target' to shell script (python, same process)
|
exec composed script (shell)
|
|- eval activate scripts (shell, same process)
|
exec target

#4
conda run (python)
|
|- conda shell.posix activate (python, same process)
|
|- add 'exec target' to shell script (python, same process)
|
exec composed script (shell)
|
|- eval activate scripts (shell, same process)
|
exec target

#5
conda run (python)
|\
| (shell)
| |\
| | conda shell.posix activate (python)
| |/
| |- eval activate scripts (shell, same process)
| |
| exec target
|/
exit <-- ugly, waiting process...

#6
conda run (shell) <-- less flexible, e.g. needs to parse argv etc., in shell...
|\
| conda shell.posix activate (python)
|/
|- eval activate scripts (shell, same process)
|
exec target
```
- `#1` is the current state of the PR, if I'm not missing anything.
- `#2` is a very minor, trivial change to avoid another fork (but that one is not important at all).
- `#3` is nicer since we can directly invoke the `python`->`exec shell`->`exec target` chain.
- `#4` improves `#3` so that no fork (and costly `python` call with `conda.context` init) is necessary.
- `#5` has a waiting process :-1:.
- `#6` is the less flexible `conda-run.sh` approach.

IMO, `#1` or `#2` are fine as an initial implementation, `#3` needs some additional changes but would be nice to have and `#4` is my favorite but needs some more (minor?) changes to make `conda.activate` work nicely from within an existing `conda` process.

> I would be shocked if that optimization gains more than 50 ms.

I expect something in the ballpark of 100-150 ms, maybe, which is may be a factor 2-3 more, but is still not too far off considering the overall process would be < 300 ms, I think. And if a user wants to run it in a tight loop, then they're likely doing something wrong anyway..

mbargull:
NB: I disregarded Windows in all of the above again :wink:

conda/cli/main_run.py
```diff
@@ -0,0 +1,86 @@
...
+
+    temp_path = None
+    try:
+        with NamedTemporaryFile('w+b', suffix='.bat', delete=False) as tf:
+            temp_path = tf.name
+            tf.write(ensure_binary(
+                "@%CONDA_PYTHON_EXE% -c \"import os, json; print(json.dumps(dict(os.environ)))\""
+            ))
+        cmd_builder = [
+            "cmd.exe /K \"",
```
mbargull: Instead of `cmd.exe /K`, maybe this could use `/C` instead? Would it make sense to use `COMSPEC`?
If yes to both, maybe the logic in https://github.com/conda/conda/blob/616cf994f71db564047dd420216d2dbcb59010f3/conda/core/link.py#L776-L784 should be put into a function that can be reused here?

conda/cli/conda_argparse.py
```diff
@@ -958,6 +952,47 @@ def configure_parser_remove(sub_parsers, name='remove'):
...
+        "-v", "--verbose",
+        action=NullCountAction,
+        help="Use once for info, twice for debug, three times for trace.",
+        dest="verbosity",
+        default=NULL,
+    )
+
+    p.add_argument(
+        'executable_call',
+        nargs=ONE_OR_MORE,
```
mbargull: You'll have to use `REMAINDER` here. Otherwise, e.g.,
```sh
conda run -v -p /p target -v
```
will be parsed as
```python
{'v': 2, 'p': '/p', 'executable_call': 'target'}
```

conda/cli/main.py
```diff
@@ -69,7 +69,16 @@ def _main(*args, **kwargs):
         args = args + ('-h',)
 
     p = generate_parser()
-    args = p.parse_args(args[1:])
+    args, unknown_args = p.parse_known_args(args[1:])
+    if unknown_args:
+        if args.cmd == "run":
+            args.unknown_args = unknown_args
```
mbargull: If you use `argparse.REMAINDER`, you can leave out the the `args.cmd == "run"`-specific part.

kalefranz:
I've unfortunately really jumbled up this PR with https://github.com/conda/conda/pull/7389.  Will definitely try to avoid that in the future.

kalefranz:
When @mbargull is on site, maybe one of the things we can do is work through all the options in https://github.com/conda/conda/pull/7320#issuecomment-397797870.  For now, I'm going to merge this PR soon (noting the command is marked `[EXPERIMENTAL]` in `conda run --help`), so that we can move forward with the next 4.6.0 alpha release.

kalefranz:
Just waiting on CI now following another rebase.

PR #7322, opened 2018-05-21
Verification should be set to true if safety checks are disabled

SubhamRajgaria:


kalefranz:
Even with safety checks disabled, we'll still need to record the expected sha256 of the path in the prefix, after prefix replacement.  However, if for some reason we're doing extra verification of the sha256 of the path in the package cache for `PrefixReplaceLinkAction`, then that one we can skip for safety checks disabled.

kalefranz:
In general we try very hard to minimize the number of files that require prefix replacement.  We ultimately consider it a bit of an anti-pattern, and a hack-of-last-resort.  For example see comment from @mingwandroid at https://github.com/conda/conda-build/issues/2207#issuecomment-390938134.

If there are particular packages you're working with that you've built yourself or on conda-forge that make use of extensive prefix replacement, I'd definitely recommend employing various techniques to scale that back.

conda/core/path_actions.py
```diff
@@ -13,7 +13,7 @@
 from .prefix_data import PrefixData
 from .._vendor.auxlib.compat import with_metaclass
 from .._vendor.auxlib.ish import dals
-from ..base.constants import CONDA_TARBALL_EXTENSION
+from ..base.constants import CONDA_TARBALL_EXTENSION, SafetyChecks
```
kalefranz: This import surprises me. I am in favor right now of moving the `SafetyChecks.disabled` logic from `conda/core/link.py` to `conda/core/path_actions.py`.  There may be several `PathActions` that have to be adjusted.

PR #7369, opened 2018-06-03
fix windows test failures when symlink not available

kalefranz:
These failures went undiscovered for so long because apparently appveyor is configured with symlinks enabled by default.  The failures showed up in the Anaconda build system, where symlinks aren't enabled by default on Windows workers.


kalefranz:
ping

PR #7385, opened 2018-06-05
fix #5681 conda env create / update when --file does not exist

### LINKED ISSUE #5681: conda env create gives confusing message if environment file does not exist
Try executing `conda env create -f does_not_exist.txt`.  The output I get is:

```
Using Anaconda API: https://api.anaconda.org

SpecNotFound: Can't process without a name
```

That is not super useful or meaningful.  I'd suggest it should say something obvious, like:

```
File not found: does_not_exist.txt
```

goanpeca:
Fixes #5681 

```
$ conda env create -f does_not_exist.txt

EnvironmentFileNotFound: Conda Env Exception: /Users/gpena-castellanos/Continuum/conda/does_not_exist.txt file not found
```

```
$ conda env update -f does_not_exist.txt

EnvironmentFileNotFound: Conda Env Exception: /Users/gpena-castellanos/Continuum/conda/does_not_exist.txt file not found
```

```
$ conda env create

EnvironmentFileNotFound: Conda Env Exception: /Users/gpena-castellanos/Continuum/conda/environment.yml file not found
```

### DIFF
--- a/conda_env/specs/__init__.py
+++ b/conda_env/specs/__init__.py
@@ -2,10 +2,8 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 
-# Standard library imports
 import os
 
-# Local imports
 from .binstar import BinstarSpec
 from .notebook import NotebookSpec
 from .requirements import RequirementsSpec
--- a/conda_env/specs/__init__.py
+++ b/conda_env/specs/__init__.py
@@ -1,11 +1,17 @@
 # -*- coding: utf-8 -*-
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+
+# Standard library imports
+import os
+
+# Local imports
 from .binstar import BinstarSpec
-from .yaml_file import YamlFileSpec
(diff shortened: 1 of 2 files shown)


conda_env/specs/__init__.py
```diff
@@ -16,6 +22,12 @@
 
 
 def detect(**kwargs):
+    # Check file existence if --file was provided
```
goanpeca: This part is probably better located elsewhere, maybe as a separate function, just to better isolate concerns? changing it here made it simpler, otherwise we would need to change in different cli methods like create and update to add a file check before the spec detection

Thoughts?

conda_env/specs/__init__.py
```diff
@@ -1,11 +1,17 @@
 # -*- coding: utf-8 -*-
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+
+# Standard library imports
+import os
+
+# Local imports
```
kalefranz: The separation and import organization is good. Let's leave off the comments for the sections.

goanpeca:
@kalefranz should I backport this to 4.5.x series?

PR #7389, opened 2018-06-06
msys2 path conversion

kalefranz:
Per conversation with @mingwandroid, `cygpath "$(cygpath -w "$PATH")"` is not invariant.  See https://github.com/conda/conda/pull/7316 for additional context. This PR is only a partial fix. A full fix would require substantial hacks and additional logic in `conda.sh`.  It's my opinion at this time that suboptimal decisions made by upstream msys2 have lead to this situation (msys2 should have no `/bin`--just `/usr/bin`--and there'd be no bind mount for us to worry about here), and a full solution is ultimately incumbent on upstream.

NOTE: This PR needs to be rebased against master following merge of https://github.com/conda/conda/pull/7320. The functionality in this PR will most likely be back ported to conda 4.5.x.

The first commit of this PR is https://github.com/conda/conda/pull/7389/commits/4b9e0799012bfa1443ef12806d259c3ea1c72bf6.  If reviewing, start by looking only at that commit.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ init:
 #  - ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 
 install:
-  - "%PYTHON_ROOT%\\Scripts\\conda install -y -c defaults -c conda-forge python=%PYTHON_VERSION% pycosat conda requests ruamel_yaml pytest pytest-cov pytest-timeout mock responses pexpect"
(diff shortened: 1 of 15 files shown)


mingwandroid:
> It's my opinion at this time that suboptimal decisions made by upstream msys2 have lead to this situation (msys2 should have no /bin--just /usr/bin--and there'd be no bind mount for us to worry about here

I know you said your opinion and while I wish we'd made a different choice way back, I do think having bind mounts is a 'user' choice .. here we're just guarding against the ill effects of it WRT conda and our prefixes so us taking extra care is not a case of fixing this 'issue' in the wrong place (just my opinion).

PR #7396, opened 2018-06-07
Rename common.platform to common.os and split among windows, linux and unix utils

### LINKED ISSUE #7411: Split common/platform  module on conda to common/os/<os>
This is a refactor of the code to split utilities that refer to specific operating systems

https://github.com/conda/conda/blob/master/conda/common/platform.py

goanpeca:
Fixes #7411

### DIFF
--- a/conda/common/os/__init__.py
+++ b/conda/common/os/__init__.py
@@ -5,32 +5,15 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from logging import getLogger
 
-from ..os.linux import linux_get_libc_version
-from ..os.unix import get_free_space_on_unix, is_admin_on_unix
-from ..os.windows import get_free_space_on_windows, is_admin_on_windows
 from ..compat import on_win
 
 
+if on_win:
+    from .windows import get_free_space_on_windows as get_free_space
+    from .windows import is_admin_on_windows as is_admin
+else:
+    from .unix import get_free_space_on_unix as get_free_space  # noqa
+    from .unix import is_admin_on_unix as is_admin  # noqa
+
+
 log = getLogger(__name__)
-
-
-linux_get_libc_version
-
-
-def is_admin():
-    func = is_admin_on_windows if on_win else is_admin_on_unix
-    return func()
-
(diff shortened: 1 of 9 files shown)


goanpeca:
Is this what you had in mind for the split @kalefranz  ?

conda/common/os/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+from ..os.linux import linux_get_libc_version
+from ..os.unix import get_free_space_on_unix, is_admin_on_unix
+from ..os.windows import get_free_space_on_windows, is_admin_on_windows
+from ..compat import on_win
+
+
+log = getLogger(__name__)
+
+
+linux_get_libc_version
```
kalefranz: We shouldn't be importing from all platforms all the time here.  Should be more like

```python
if on_win:
    from .windows import is_admin_on_windows as is_admin
else:
    from .unix import is_admin_on_unix as is_admin
```

We don't need the `linux_get_libc_version` import at all in this module I don't think.

conda/common/os/linux.py
```diff
@@ -4,45 +4,22 @@
...
 from genericpath import exists
 from glob import glob
 from logging import getLogger
-import os
 import sys
 
-from .compat import iteritems, on_win
-from .._vendor.auxlib.decorators import memoize
+from ..._vendor.auxlib.decorators import memoize
+from ...compat import iteritems
```
kalefranz: This should just be 

    from .compat import iteritems

conda/common/os/windows.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from logging import getLogger
+
+from ...compat import on_win
```
kalefranz: Don't use `conda/compat.py`.  Use `conda/common/compat.py`.

    from .compat import on_win

PR #7457, opened 2018-06-24
Add 'conda package --list' feature: show installed files

ihnorton:
...belonging to given package.


ihnorton:
Rebased to use correct ToT functions and fix flake8 compliance. I believe the appveyor failure is unrelated.

ihnorton:
Bump. Wanted this again in a new install :/

v4hn:
Hellooo. First time active user of the framework and had to debug various issues with wrongly detected package versions in cmake. This feature would have helped *a lot*. I don't understand why people would not want to have this in the front-end...

travishathaway:
@conda-bot

travishathaway:
Hi @ihnorton,

We are currently undergoing an effort to clean up stale pull requests. We do appreciate your original pull request, but we think it's best to close it. The reason being that the `package` command is not widely used and still today marked as experimental.

In the future, it would be better to write such things as conda plugins rather than trying to add this to the core of conda:

- https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/index.html

PR #7484, opened 2018-07-04
Add conda version to conda-meta/history

nehaljwani:
Also raise CondaHistoryError if a newer (minor/major) version of conda is used
to modify an existing environment which was created by an older one


nehaljwani:
Closing, these changes are already in master.

Issue #7437, opened 2018-06-18
Add conda version check to history entries in 4.3 and forward


kalefranz:
part of #7429 epic

PR #7536, opened 2018-07-13
Backport CondaUpgradeError for conda version downgrades on environments to 4.4.x

nehaljwani:
#7437


nehaljwani:
@kalefranz Only one test is failing in this PR, and I am conflicted about the right way to get around it. Also, the tests for history file don't seem to have run. Why would that be the case?

kalefranz:
I'm going to go ahead and close this PR. At this point, there's no reason to maintain the 4.4.x branch.  May even delete it.  There's no reason to maintain it any longer.  Definitely looking at the 4.3.x PR though still.

Issue #5540, opened 2017-06-16
Error while activate env


ntsd:
I use Conda Python 3.6
and create env for Python 3.5

```
C:\Windows\system32>conda env list
# conda environments:
#
py35                  *  C:\ProgramData\Miniconda3\envs\py35
root                     C:\ProgramData\Miniconda3


C:\Windows\system32>activate py35
Fatal Python error: Py_Initialize: can't initialize sys standard streams
LookupError: unknown encoding: 874

Current thread 0x00002338 (most recent call first):
Fatal Python error: Py_Initialize: can't initialize sys standard streams
LookupError: unknown encoding: 874

Current thread 0x000021d8 (most recent call first):
Fatal Python error: Py_Initialize: can't initialize sys standard streams
LookupError: unknown encoding: 874

Current thread 0x00000754 (most recent call first):

C:\Windows\system32>python --version
Python 3.6.1 :: Anaconda 4.4.0 (64-bit)
```

ntsd:
Fixed just change system local to English(UK)

PR #4558, opened 2017-02-07
activate.bat: Set codepage to CP_ACP

mingwandroid:
@kalefranz, @csoja would like a test Miniconda release containing this change. This is an attempt to fix https://github.com/ContinuumIO/anaconda-issues/issues/1410. If you would prefer not to merge it then I'll make a local patch for conda for @ilanschnell to try out.

PR #7584, opened 2018-07-26
Remove PYTHONIOENCODING on Windows

methane:
Python uses current codepage automatically.
Additionally, Python uses UTF-8 for full unicode support
on console.  See PEP 529.

Theare are no reason to use PYTHONIOENCODING on Windows.


kalefranz:
What about python 2.7, 3.4, and 3.5?

methane:
In all of 2.7, 3.4, 3.5, PYTHONIOENCODING is not needed.  Python uses codepage encoding automatically.

When stdout is console, there are no reason to use it on Windows.
On *nix, it is useful when we use C/POSIX locale but we want UTF-8.  (This situation is cared automatically from Python 3.7).

When stdout is redirected, Python 2 used ascii for sys.stdout.encoding.
So `print u"こんにちは"` can be raise UnicodeEncodeError when stdout is redirected.
This is fixed by Python 3.2.

```
$ python2 -c 'import sys; print(sys.stdout.encoding)'
UTF-8

$ python2 -c 'import sys; print(sys.stdout.encoding)' | cat
None

$ python3 -c 'import sys; print(sys.stdout.encoding)' | cat
UTF-8
```

methane:
I don't know why these hack is added. I found two issues:

* https://github.com/conda/conda/pull/4558
* https://github.com/ContinuumIO/anaconda-issues/issues/1410

But I don't know why PYTHONIOENCODING hack is needed to solve this.

Isn't that issue is solved by installer?
https://github.com/conda/constructor/compare/9a2e991476e4d6c6d2f8f8712f3a960a4dd2a2ef...6f41ba9f3af2c995b1847c7333d8825f48bc5b0b

methane:
Anyway, codepage number is not valid encoding always:

*  #5540
* https://bugs.python.org/issue33865
* https://stackoverflow.com/questions/43306103/error-in-anaconda-when-activate-env
* https://medium.com/@jatupornjirundorn/%E0%B9%80%E0%B8%A1%E0%B8%B7%E0%B9%88%E0%B8%AD%E0%B9%80%E0%B8%9B%E0%B8%B4%E0%B8%94-anaconda-prompt-%E0%B9%81%E0%B8%A5%E0%B9%89%E0%B8%A7%E0%B9%80%E0%B8%88%E0%B8%AD-err-%E0%B9%81%E0%B8%9A%E0%B8%9A%E0%B8%99%E0%B8%B5%E0%B9%89-fatal-python-error-py-initialize-cant-initialize-sys-1186db3b0bc9

mingwandroid:
Thanks for the PR @methane, I need to investigate all of this again, I'll do it ASAP but I have a lot to do first.

msarahan:
Cycling to re-run CI with curernt code before merging.

travishathaway:

PR #7628, opened 2018-08-07
address #7626 featureless package minimization default behavior change

kalefranz:
addressing #7626


kalefranz:
Closing, per discussion in #7626.

PR #7629, opened 2018-08-07
implement win 10 long path reg key for conda init

msarahan:


conda/core/initialize.py
```diff
@@ -1107,6 +1118,29 @@ def init_cmd_exe_registry(target_path, conda_prefix):
         return Result.NO_CHANGE
 
 
+def init_long_path(target_path):
+    win_ver, win_rev = platform.platform().split('-')[1:3]
+    # win10, build 14352 was the first preview release that supported this
```
kalefranz: I just extracted from another PR and merged into master something that might be cleaner here (the new code was in https://github.com/conda/conda/pull/7630; you'll need to rebase).  For `win_rev`, I believe you can use `context.os_distribution_name_version[1]`.  I think on older systems, what you have for `win_ver` might actually be able to be `NT`, so `int('NT')` would 💣.

PR #7641, opened 2018-08-10
fix py37 compat 4.5.x

kalefranz:


mingwandroid:
We need this in the new miniconda 3.7's I guess? When are we going to trail conda 4.6 in a miniconda release? Could we make preview ones?

PR #7645, opened 2018-08-11
PrefixData: replace glob with fnmatch 4.5.x

nehaljwani:
Special characters like '[',']' in prefix path need to be escaped for the glob module, which is annoying. Use fnmatch instead, which is used by glob1 internally anyway.

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from glob import glob
+import fnmatch
+from os import listdir
 from logging import getLogger
 from os.path import isfile, join, lexists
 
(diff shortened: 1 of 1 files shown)


conda/core/prefix_data.py
```diff
@@ -46,8 +47,10 @@ def __init__(self, prefix_path):
 
     def load(self):
         self.__prefix_records = {}
-        for meta_file in glob(join(self.prefix_path, 'conda-meta', '*.json')):
-            self._load_single_record(meta_file)
+        _conda_meta_dir = join(self.prefix_path, 'conda-meta')
+        if lexists(_conda_meta_dir):
+            for meta_file in fnmatch.filter(listdir(_conda_meta_dir), '*.json'):
```
kalefranz: What an awful name for a function in the standard library.

kalefranz: Would normally do this import as `from fnmatch import filter as fnmatch_filter`. Importing each symbol into the global scope for the module avoids one extra lookup.  Not a big deal here though.

PR #7648, opened 2018-08-12
Don't print name, version & size, if unknown 4.5.x

nehaljwani:


kalefranz:
I'm going to just close this PR.  Just merged #7648 in 4.5.x, and the merge up is straight forward.  When that's the case, there's really no need for two PRs.  Just target the earlier version, and the fix will be covered in the merge up.

PR #7649, opened 2018-08-12
PrefixData: replace glob with fnmatch

nehaljwani:
Special characters like '[',']' in prefix path need to be escaped for
the glob module, which is annoying. Use fnmatch instead, which is
used by glob1 internally anyway.


kalefranz:
This was already merged into master via merge up from #7645 targeted at 4.5.x.

PR #7660, opened 2018-08-16
Implement optional strict channel priority

### LINKED ISSUE #7656: "strict" channel preference
It would be nice if conda could handle 3 scenarios when solving packages via an option on the `.condarc`. The options would be:

0: no channel preference, solve for the newest versions (already possible with a config option)
1: "some" channel preference (the current status)
2: "strict" channel preference (not implemented)

Rationale for this request:

At the moment if we use 1 we may get a different solution for two virtually identical environments, for example one with `cartopy` only and another one with `cartopy` and `matplotlib` may yield different soluytions. The reason for that is b/c in the former `matplotlib` is implicit and conda won't prioritize it, on the second one conda will prioritize `matplotlib` along-side with `cartopy`. (@kalefranz please correct me if I am wrong.)

Also, under this scenario one may get a bad mix of packages from different channels that is hard to debug. With option 2 both envs above would have the same solution and an unsolvable env would result in an error rather than a broken mix packages from different channels.

The "strict" option would solve packages for the first priority channel and look for packages on the lower priority only if they **are not present** in the first priority.

It would make an env like this,

```yaml
name: mapping
channels:
  - defaults
dependencies:
  - python=3.6
  - cartopy
  - conda-forge::folium
  - conda-forge::compliance-checker
  - conda-forge::gsw
```

simpler because we could just add `conda-forge` as the lower priority in the list of channels and drop the `conda-forge::` syntax for the other packages. Note that when there are many packages exclusively on the lower priority channel this can be very convenient.

Another scenario would be to allow for third party channel to easily use `defaults` compilers and runtimes without the need to incorporate them (or vendorize, ping @scopatz :wink:) into their channel. Ensuring they are always getting the latest and most stable compiler/runtime while still building their env with the third party channel.

Note that none of the options above would ensure a stable env! However,  option 2 is arguably the only one easy to debug and we expect "unsolvable" envs to happen rather than a bad mix of channels.

jjhelmus:
resolves #7656

Adds a new configuration option, "strict_channel_priority", which determines if packages in lower priority channel are to be considered by the solver if a package with the same name is present in a higher priority channel.  When set to False, the default, these packages are still considered which is consistent with the current behavior.  When True, these packages are not considered for installation.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -174,6 +174,7 @@ class Context(Configuration):
                                         aliases=('channel_alias',),
                                         validation=channel_alias_validation)
     channel_priority = PrimitiveParameter(True)
+    strict_channel_priority = PrimitiveParameter(False)
     _channels = SequenceParameter(string_types, default=(DEFAULTS_CHANNEL_NAME,),
                                   aliases=('channels', 'channel',))  # channel for args.channel
     _custom_channels = MapParameter(string_types, DEFAULT_CUSTOM_CHANNELS,
@@ -716,6 +717,7 @@ class Context(Configuration):
             'aggressive_update_packages',
             'auto_update_conda',
             'channel_priority',
+            'strict_channel_priority',
             'create_default_packages',
             'disallowed_packages',
             'pinned_packages',
@@ -1045,6 +1047,13 @@ class Context(Configuration):
                 """),
             'show_channel_urls': dals("""
                 Show channel URLs when displaying what is going to be downloaded.
+                """),
+            'strict_channel_priority': dals("""
+                When True, packages in lower priority channels are not
+                considered if a package with the same name appears in a higher
+                priority channel. When False, packages from all channels are
+                considered by the solver.  The 'channel_priority' configuration
+                parameter is still applied in either case.
                 """),
             'ssl_verify': dals("""
                 Conda verifies SSL certificates for HTTPS requests, just like a web
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from collections import defaultdict
 from itertools import chain
 from logging import getLogger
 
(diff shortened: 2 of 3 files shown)


jjhelmus:
@kalefranz, I would appreciate your feedback on this.  I'm happy to add tests and polish to this if the approach seems reasonable.

kalefranz:
Awesome you started working on this! A couple initial thoughts without looking at the code line-by-line.

1. The `get_reduced_index()` method in `conda/resolve.py` is much more fine-grained than the more course `get_reduced_index()` function you've added this too. I think it's better to implement strict channel priority in the more fine-grained version.
2. Have you thought through how strict channel priority should work with `track_features`? That seems like a potential complication that one of us needs to think through carefully. It's not at all clear to me yet what the right thing to do here is.
3. I don't think we need a new `strict_channel_priority` config parameter.  Really what we need is for `channel_priority` to be an enum parameter with three options... Maybe call them `strict`, `flexible`, and `disabled`?  Except I don't like the word `flexible`.  But something like that.  The enum class would go in `conda/base/constants.py`. It'd be nice to try to [override the `__new__` method](https://stackoverflow.com/questions/24105268/is-it-possible-to-override-new-in-an-enum-to-parse-strings-to-an-instance) to translate `True` / `False` into `flexible` and `disabled`. Also when you add a new primitive config parameter that's an enum type, you have to add it to https://github.com/conda/conda/blob/3ed26bc6a2fb033f45d6a8e4d5bc8610ec1dc1a1/conda/cli/main_config.py#L348-L351.

kalefranz:
On tests... They will go in https://github.com/conda/conda/blob/3ed26bc6a2fb033f45d6a8e4d5bc8610ec1dc1a1/tests/core/test_solve.py, which is unit tests for the solver.  There are several different data sources already.  Might be worthwhile right now for me to take the time to clean some stuff up for you to use here.  I'll work on that now.

Look at that file for uses of `get_solver_aggregate_1()`.  For the general pattern of how best to construct a two-stage test (e.g. a `create` step followed by `install`, `update`, or `remove`), `test_prune_1()` is probably a good example.

kalefranz:
I've merged https://github.com/conda/conda/pull/7664, so you should be good to go with tests now, and looking around at other tests for examples of how to write them.

kalefranz:
Might also be nice to figure out how to have a "strict, but fallback to flexible mode if strict is unsatisfiable."  Honestly, maybe that should even be the default.  I'm guessing the details of that mode would probably get implemented somewhere around https://github.com/conda/conda/blob/7a7cc094adf1f29829e6ed063fb4098f84facc69/conda/core/solve.py#L346-L360

jjhelmus:
> The get_reduced_index() method in conda/resolve.py is much more fine-grained than the more course get_reduced_index() function you've added this too. I think it's better to implement strict channel priority in the more fine-grained version.

Looking at the `get_reduce_index()` method in conda/resolve.py it is unclear to me how to effectively remove packages which appear in a higher priority channel inside the method.  Once an instance of the Resolve class is created, self.index contains all the packages which are to be considered.  Pruning this index outside of the class insures that the lower priority packages are never considered and the resolution would be faster.

kalefranz:
superseded by #7729

PR #7689, opened 2018-08-24
30% performance improvement on resolve (4.5.x)

fabioz:
I took a quick stab at profiling conda with http://www.pyvmmonitor.com to improve on https://github.com/conda/conda/issues/7239 -- working on 4.5.x instead of master as it's the version I'm using and it should've a working release sooner ;)

For what I could see in the profile, it seems that 80% of the time is in the resolve.

The first really low hanging fruit is that `records.PackageRef.__hash__` is called tons of times, so, just memoizing the result gave a 25% improvement (in my use case from 15.09 seconds to 11.39 seconds).

Doing a profile after that has shown another problem in the `entity.box` -- I'm not really sure how to solve it, but the profiler shows almost 25% of the time is spent creating the entities structure in `prefix_data:133:load_single_record`... (given that there's such a big amount of time spent here it should probably use a lighter data structure for the solve, but given that such changes are controversial, I won't provide a pull request for that as this is better suited for someone with deeper understanding of the code as it should be a bigger change -- also, probably better suited to tackle on master and not on the 4.5.x branch).

Another quick grab was in `models.channel.Channel.canonical_name` which was almost 5% of the execution time (so, memoized that too).

I'm short on time to do more on this front right now, but I think a 30% improvement with those 2 changes is already worthwhile 😄 .

### DIFF
--- a/conda/models/channel.py
+++ b/conda/models/channel.py
@@ -142,6 +142,13 @@ class Channel(object):
 
     @property
     def canonical_name(self):
+        try:
+            return self.__canonical_name
+        except AttributeError:
+            self.__canonical_name = self._compute_canonical_name()
+        return self.__canonical_name
+
+    def _compute_canonical_name(self):
         for multiname, channels in iteritems(context.custom_multichannels):
             for channel in channels:
                 if self.name == channel.name:
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -236,7 +236,11 @@ class PackageRef(BasePackageRef):
                 self.build_number, self.build)
 
     def __hash__(self):
-        return hash(self._pkey)
+        try:
+            return self._hash
+        except AttributeError:
+            self._hash = hash(self._pkey)
+        return self._hash
 
     def __eq__(self, other):
         return self._pkey == other._pkey


mbargull:
Thanks for taking a look at this. Though, there are a few issues which make it hard for me to understand your results:
1. The workload can be very different depending on, e.g.,
  - which setup you have (channel lists, etc.),
  - which command is run,
  - which packages are concerned.
2. It's not clear if the relative times you measured are influenced by the profiler or really correspond to normal execution. Profilers are great pieces of software, but they can significantly influence and skew runtime measurements as well.
3. I couldn't replicate these numbers at all for very light and light workloads, with or without channel repodata updates, so I kind of fear the Profiler skewed your measurements..
4. I can't see what your setup looks like at all and how you generated those numbers.

Plus, the part of the code you are looking at has already been worked on/been improved for the future `4.6` release. It would probably make sense for you to work with the `master` branch for such improvements since otherwise you are likely to do duplicate work that is already in `master`. And working on `master` has been made fairly easy: you can just run `. dev/start` in the repository's root folder and that will automatically create and put you right into a separate Conda installation just for developing `conda`.

I do really appreciate you looking at this and like to encourage you to work on `master` and document your setup so we can work together on this!

kalefranz:
Work on this is definitely appreciated. Also agreed with @mbargull that the master branch is the right place for profiling and performance improvements. Once a feature release flies, we typically wouldn’t backport performance enhancements unless we’ve introduced a major regression somewhere in a patch release.

fabioz:
My setup is a real world use case with 3 channels and more than a 100 deps (not on my work machine to give you the real number right now, but definitely not a  light setup).

Timing was done without the profiler (just used it to find hotspots).

Sorry to hear that you can't backport performance improvements (I'm sure the master is very different, but I can't really use it until there's a stable release..).

So, feel free to close it if it can't be applied to 4.5.

mbargull:
> My setup is a real world use case with 3 channels and more than a 100 deps (not on my work machine to give you the real number right now, but definitely not a light setup).

I'd still be interested in that setup. Please feel free to post it here once you get your hands on that machine again :). I'd definitely want to see if I can replicate it and also whether the `4.6.x` changes from `master` cover that scenario already or if the changes you proposed would even make a difference on that branch.

> Timing was done without the profiler (just used it to find hotspots).

:+1:

fabioz:
I'm adding the environment.yml which is being used for the resolve below (so, a bit more than 100 deps).

`conda env update -f environment.yml` can be used to get timings (http://www.pyvmmonitor.com/ can be used with the open source license to profile conda -- but as previously noted, the final timing should not be done using the profiler).

[environment.yml.txt](https://github.com/conda/conda/files/2323732/environment.yml.txt)

Since this can't be applied to 4.5, can you share when 4.6 is expected to be released?

mbargull:
Which channels did you use?

fabioz:
> Which channels did you use?

Humm, just noted that one of the channels I'm using is internal, so, providing a new version which you can install just with `conda-forge` and `defaults`.

[environment.yml.txt](https://github.com/conda/conda/files/2324569/environment.yml.txt)

You can measure doing `conda env update` (the first time will be pretty slow due to download time, but executing the same command again show the overhead on `resolve`).

mbargull:
I took a look at this and targeted this against `master` in gh-7715.
Caching `Channel.canonical_name` is already in `master` so that only the caching of `__hash__` is ported over.

mbargull:
Regarding the `30 %` figure: I applied this to `4.5.10` which gave
- `14.3 s` unchanged,
- `13.5 s` with `Channel.canonical_name` cached,
- `13.2 s` with `Channel.canonical_name` and `PackageRef.__hash__` cached

So at most a `8 %` less time for the smaller test case from https://github.com/conda/conda/pull/7689#issuecomment-416271341. Although a lot less than `30 %`, this is still significant, so thanks nonetheless, @fabioz.

----
> Since this can't be applied to 4.5, can you share when 4.6 is expected to be released?

I don't know (I'm just a contributor, not affiliated with Anaconda :wink:), but I'd expect this to take at least a few more weeks.

mbargull:
closing in favor of gh-7715.

Issue #7717, opened 2018-08-31
conda search improvement


kalefranz:
@msarahan 
> to be fair, conda search isn't very good with partial name matching

@kalefranz
> It's not?

> ```
> $ conda search *jdk*
> Loading channels: done
> # Name                  Version           Build  Channel
> copy-jdk-configs-cos7-ppc64le             3.3               0  pkgs/main
> java-1.7.0-openjdk-cos6-i686       1.7.0.131      h3a8c334_0  pkgs/main
> java-1.7.0-openjdk-cos6-x86_64       1.7.0.131      h06d78d4_0  pkgs/main
> java-1.7.0-openjdk-cos6-x86_64       1.7.0.131      haf0aaab_0  pkgs/main
> java-1.7.0-openjdk-cos7-ppc64le       1.7.0.171               0  pkgs/main
> java-1.7.0-openjdk-devel-cos6-i686       1.7.0.131      hb747e4b_0  pkgs/main
> java-1.7.0-openjdk-devel-cos6-x86_64       1.7.0.131      hca4a25c_0  pkgs/main
> java-1.7.0-openjdk-devel-cos7-ppc64le       1.7.0.171               0  pkgs/main
> java-1.7.0-openjdk-headless-cos7-ppc64le       1.7.0.171               0  pkgs/main
> openjdk                 8.0.121               0  pkgs/free
> openjdk                 8.0.121               1  pkgs/free
> openjdk                 8.0.152      h393ad39_1  pkgs/main
> openjdk                 8.0.152      he2ab291_0  pkgs/main
> ```

@msarahan 
> right, but I'd bet most people do: `conda search jdk`

> ```
> $ conda search jdk
> Loading channels: done
> 
> PackagesNotFoundError: The following packages are not available from current channels:
> 
>   - jdk
> 
> Current channels:
> 
>   - https://repo.anaconda.com/pkgs/r/osx-64
>   - https://repo.anaconda.com/pkgs/r/noarch
>   - https://repo.anaconda.com/pkgs/main/osx-64
>   - https://repo.anaconda.com/pkgs/main/noarch
> 
> To search for alternate channels that may provide the conda package you're
> looking for, navigate to
> 
>     https://anaconda.org
> 
> and use the search bar at the top of the page.
> ```

@kalefranz

> I see.  At a minimum, we should suggest using wildcards in the PackagesNotFound message for `conda search` (though not for `create`, `install`, `update`, `remove`).

[...]

PR #7722, opened 2018-09-02
conda search improvements

### LINKED ISSUE #7336: conda search should do partial name searches (regression)
The behavior of `conda search` used to be that `conda search X` would show all packages whose name contained the string `X`. There was a `-f` flag to search for full names.

Now it seems you have to explicitly add `*` to either side of the string to do a partial name search. 

I get that search now accepts a general package spec, but adding stars is annoying and unintuitive to beginner users. The most common desired behavior is to search for substrings. This is how the search functionality for every other package manager that I've used works (`pip search`, `apt-cache search`, `yum search`, `brew search`, `fink list`, `cabal list`, I'm sure you could find and try others)

I would suggest making search do a partial match search by default, as long as the search string doesn't include characters that would indicate it's a more general package spec. It can also use the `-f` flag, which was apparently kept but hidden from the help, to override this behavior.

kalefranz:
resolve #7336, #7717

* Adds timestamp information to output of `conda search --info`. If timestamp is not given in repodata, tries falling back to date field.
* Addresses #7336 by trying exact match first, but then falling back to more flexible search by padding the package name with `*` before raising `PackagesNotFoundError`.

### DIFF
--- a/conda/_vendor/boltons/__init__.py
+++ b/conda/_vendor/boltons/__init__.py
@@ -1,2 +1,0 @@
-from __future__ import absolute_import, division, print_function, unicode_literals
-__version__ = "16.5.1"
--- a/conda/_vendor/boltons/setutils.py
+++ b/conda/_vendor/boltons/setutils.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+# source: https://raw.githubusercontent.com/mahmoud/boltons/18.0.0/boltons/setutils.py
+# version: 18.0.0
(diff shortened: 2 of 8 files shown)


calebmarchent:
Is there a way to disable the auto-fallback to substring matching implemented in this change?
I can no longer rely on the output of conda search to determine if a package is known to conda and need to filter the results by hand.

The following now gives lots of results, despite the package not being known to my conda channels:
```
conda search -f 'pyinter' 
```

Issue #7394, opened 2018-06-07
Behavior changes to consider for conda 5.0


kalefranz:
A running list of behavior changes to consider for the next major version bump of conda.

* #6147
* #12399
* #12401
* #12400
* #12403
* #12402

astrojuanlu:
Was any of the behaviors described in the top comment implemented? Or was this closed because the issue didn't make sense anymore?

kenodegard:
@astrojuanlu These ideas are all still valid but deserve to be their own epics, I'll reopen this for now and work to split them up.

kenodegard:
New issues have been created to track the ideas above, closing this epic

PR #7729, opened 2018-09-03
strict channel priority via 'channel_priority' config option or --strict-channel-priority CLI flag

### LINKED ISSUE #7656: "strict" channel preference

kalefranz:
resolves #7656 
supersedes #7660

```
$ conda config --describe channel_priority
# # channel_priority (ChannelPriority)
# #   Accepts values of 'strict', 'flexible', and 'disabled'. The default
# #   value is 'flexible'. With strict channel priority packages in lower
# #   priority channels are not considered if a package with the same name
# #   appears in a higher priority channel. With flexible channel priority,
# #   the solver may reach into lower priority channels to fulfill
# #   dependencies, rather than raising an unsatisfiable error. With channel
# #   priority disabled, package version takes precedence, and the
# #   configured priority of channels is used only to break ties. In
# #   previous versions of conda, this parameter was configured as either
# #   True or  False. True is now an alias to 'flexible'.
# #
# channel_priority: flexible
```

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -565,13 +565,6 @@ class IntegrationTests(TestCase):
             assert stderr == ''
             self.assertIsInstance(stdout, str)
 
-    def test_anaconda_solve_time(self):
-        stdout, stderr = run_command(Commands.CREATE, "/", "-qd anaconda",
-                                     use_exception_handler=True)
-        assert not stderr
-        assert "The following NEW packages will be INSTALLED:" in stdout
-        assert "Dry run. Exiting." in stdout
-
     @pytest.mark.skipif(on_win and context.subdir == "win-32", reason="conda-forge doesn't do win-32")
     def test_strict_channel_priority(self):
         stdout, stderr = run_command(
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -5,6 +5,10 @@ import bz2
 from contextlib import contextmanager
 from datetime import datetime
 from glob import glob
+
+from conda._vendor.toolz.itertoolz import groupby
+from conda.models.channel import Channel
+from conda.resolve import Resolve
 from itertools import chain
 import json
 from json import loads as json_loads
@@ -38,6 +42,7 @@ from conda.common.path import get_bin_directory_short_path, get_python_site_pack
     pyc_path
 from conda.common.serialize import yaml_load, json_dump
 from conda.common.url import path_to_url
+from conda.core.index import get_reduced_index
 from conda.core.prefix_data import PrefixData, get_python_version_for_prefix
 from conda.core.package_cache_data import PackageCacheData
 from conda.core.subdir_data import create_cache_dir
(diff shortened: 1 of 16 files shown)


tests/test_create.py
```diff
@@ -565,13 +565,6 @@ def test_create_empty_env(self):
             assert stderr == ''
             self.assertIsInstance(stdout, str)
 
-    def test_anaconda_solve_time(self):
```
mbargull: Such performance/benchmarking tests are no bad idea, generally. Especially in light of all those index and solver changes for `4.6.x`.
It could make sense to add a `@pytest.mark.` to those and only run them on demand and/or in a separate, e.g., nightly, job. Or, if necessary, even put them into a separate test suite or repository.
I'd probably just use some `@pytest.mark.performance` and/or create a `test_performance.py` and set
```
import pytest
pytestmark = pytest.mark.performance
```
in there.

ocefpaf:
@kalefranz and @mbargull just wanted to let you know that, even though it took me a while to test this, so far it has done wonders to our workflow! Thanks a lot for implementing this!

msarahan:
Glad to hear it!  Hopefully it doesn't have any nasty side effects and we can make it the default someday soon.

kalefranz:
I'll add the default value change to https://github.com/conda/conda/issues/7394

ocefpaf:
> Glad to hear it! Hopefully it doesn't have any nasty side effects and we can make it the default someday soon.

So far everything is working as expected, even the failures :smile: 

I'll make sure to report anything unexpected.

Issue #7529, opened 2018-07-11
consider conda package format version 2


kalefranz:
* https://github.com/conda/conda/issues/4861
* https://blog.bethselamin.de/posts/argh-pm.html
* https://en.wikipedia.org/wiki/Ar_(Unix)
* https://en.wikipedia.org/wiki/Open_Packaging_Conventions (the format used for most Office documents, but also the extensions for Visual Studio and VS Code, Nuget, and all the apps in the Microsoft Store)
* https://en.wikipedia.org/wiki/Deb_(file_format)

kalefranz:
This work has been completed with the .conda package file format.

PR #7798, opened 2018-09-23
(WIP) Add zstandard support

mingwandroid:
For discussion only at present, please rip this to shreds @conda/conda-core 

Conda-build PR at: https://github.com/conda/conda-build/pull/3163

Some results:

mkl, pre-sorting by file extension:

```
-rw-r--r--  1 rdonnelly  staff  161938050 21 Sep 02:27 /opt/conda/pkgs/mkl-2019.0-118.tar.bz2
```

mkl post-sorting and using zstandard: (edited)

```
-rw-r--r--  1 rdonnelly  staff  161570441 22 Sep 16:18 /opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar.bz2
-rw-r--r--  1 rdonnelly  staff  106457667 22 Sep 16:23 /opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar.zst
```

```
rm /opt/conda/conda-bld/osx-64//opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar ; time bzip2 -d -k /opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar.bz2

real    0m20.032s
user    0m18.342s
sys    0m0.758s

rm /opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar ; time zstd -d -k /opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar.zst
/opt/conda/conda-bld/osx-64/mkl-2019.0-118.tar.zst: 581647360 bytes

real    0m3.128s
user    0m1.313s
sys    0m0.607s
```

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -101,7 +101,8 @@ DEFAULT_AGGRESSIVE_UPDATE_PACKAGES = (
 # Maximum priority, reserved for packages we really want to remove
 MAX_CHANNEL_PRIORITY = 10000
 
-CONDA_TARBALL_EXTENSION = '.tar.bz2'
+CONDA_TARBALL_EXTENSIONS = ('.tar.bz2',)
+# CONDA_TARBALL_EXTENSIONS = ('.tar.zst', '.tar.bz2')
 
 UNKNOWN_CHANNEL = "<unknown>"
 
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -14,7 +14,7 @@ from . import common
 from .common import check_non_admin
 from .. import CondaError
 from .._vendor.auxlib.ish import dals
-from ..base.constants import ROOT_ENV_NAME
+from ..base.constants import CONDA_TARBALL_EXTENSIONS, ROOT_ENV_NAME
 from ..base.context import context, locate_prefix_by_name
 from ..common.compat import on_win, text_type
 from ..core.index import calculate_channel_urls, get_index
@@ -167,7 +167,7 @@ def install(args, parser, command='install'):
         'use_local': args.use_local
     }
 
-    num_cp = sum(s.endswith('.tar.bz2') for s in args_packages)
+    num_cp = sum(s.endswith(CONDA_TARBALL_EXTENSIONS) for s in args_packages)
     if num_cp:
         if num_cp == len(args_packages):
             explicit(args_packages, prefix, verbose=not context.quiet)
--- a/conda/common/url.py
+++ b/conda/common/url.py
@@ -12,6 +12,7 @@ from .path import split_filename
 from .._vendor.auxlib.decorators import memoize
 from .._vendor.urllib3.exceptions import LocationParseError
 from .._vendor.urllib3.util.url import Url, parse_url
+from ..base.constants import CONDA_TARBALL_EXTENSIONS
 
 try:  # pragma: py2 no cover
     # Python 3
@@ -189,6 +190,11 @@ def split_anaconda_token(url):
     return cleaned_url.rstrip('/'), token
 
 
+def rreplace(s, old, new, occurrence):
+    li = s.rsplit(old, occurrence)
+    return new.join(li)
+
+
 def split_platform(url, known_subdirs):
     """
 
@@ -201,7 +207,7 @@ def split_platform(url, known_subdirs):
     _platform_match_regex = r'/(%s)/?' % r'|'.join(r'%s' % d for d in known_subdirs)
     _platform_match = re.search(_platform_match_regex, url, re.IGNORECASE)
     platform = _platform_match.groups()[0] if _platform_match else None
-    cleaned_url = url.replace('/' + platform, '', 1) if platform is not None else url
+    cleaned_url = rreplace(url, '/' + platform, '', 1) if platform is not None else url
     return cleaned_url.rstrip('/'), platform
 
 
@@ -214,7 +220,7 @@ def has_platform(url, known_subdirs):
 
 
 def _split_package_filename(url):
-    cleaned_url, package_filename = (url.rsplit('/', 1) if url.endswith(('.tar.bz2', '.json'))
+    cleaned_url, package_filename = (url.rsplit('/', 1) if url.endswith((CONDA_TARBALL_EXTENSIONS + ('.json',)))
                                      else (url, None))
     return cleaned_url, package_filename
 
--- a/conda/core/package_cache_data.py
+++ b/conda/core/package_cache_data.py
@@ -13,7 +13,7 @@ from conda._vendor.auxlib.decorators import memoizemethod
 from .path_actions import CacheUrlAction, ExtractPackageAction
 from .. import CondaError, CondaMultiError, conda_signal_handler
 from .._vendor.auxlib.collection import first
-from ..base.constants import CONDA_TARBALL_EXTENSION, PACKAGE_CACHE_MAGIC_FILE
+from ..base.constants import CONDA_TARBALL_EXTENSIONS, PACKAGE_CACHE_MAGIC_FILE
 from ..base.context import context
 from ..common.compat import (JSONDecodeError, iteritems, itervalues, odict, string_types,
                              text_type, with_metaclass)
@@ -88,7 +88,7 @@ class PackageCacheData(object):
             if islink(full_path):
                 continue
             elif (isdir(full_path) and isfile(join(full_path, 'info', 'index.json'))
-                  or isfile(full_path) and full_path.endswith(CONDA_TARBALL_EXTENSION)):
+                  or isfile(full_path) and full_path.endswith(CONDA_TARBALL_EXTENSIONS)):
                 package_cache_record = self._make_single_record(base_name)
                 if package_cache_record:
                     _package_cache_records[package_cache_record] = package_cache_record
@@ -269,12 +269,12 @@ class PackageCacheData(object):
         return "%s(%s)" % (self.__class__.__name__, ', '.join(args))
 
     def _make_single_record(self, package_filename):
-        if not package_filename.endswith(CONDA_TARBALL_EXTENSION):
-            package_filename += CONDA_TARBALL_EXTENSION
+        if not package_filename.endswith(CONDA_TARBALL_EXTENSIONS):
+            package_filename += CONDA_TARBALL_EXTENSIONS[0]
 
         package_tarball_full_path = join(self.pkgs_dir, package_filename)
         log.trace("adding to package cache %s", package_tarball_full_path)
-        extracted_package_dir = package_tarball_full_path[:-len(CONDA_TARBALL_EXTENSION)]
+        extracted_package_dir = package_tarball_full_path[:-len(CONDA_TARBALL_EXTENSIONS[0])]
 
         # try reading info/repodata_record.json
         try:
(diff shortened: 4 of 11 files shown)


mingwandroid:
I've only tested this as far as building packages goes (and updating the index, I *think*).

Actually using them may be another matter, I may do a full rebuild of miniconda making nothing but tar.zst files, see what happens and fix whatever I run into next.

mbargull:
I'm very much in favor of supporting zstandard as a compression format for packages due to its far superior decompression speed (with a nice extra benefit of higher compression ratio, unlike with `lz4` and friends).
So, a big +1 from me to adding this sooner rather than later. (Plus -- and since this isn't something that touches any other logic in `conda` -- depending on when this gets adapted and battle-tested, I'd might even suggest you to maybe consider to backport this to `conda <4.5` versions to have the possibility to also deploy this to the enterprise realm faster, too.)

One minor issue with adding `python-libarchive-c` as a dependency is that this pulls in the `python-libarchive-c -> libarchive -> libxml2 -> icu` dependency chain. This is due to `libarchive` on default being compiled with XAR support which makes it depend on `libxml2`. Adding `icu` to Miniconda would make it considerably bigger -- in regards to the uncompressed packages, that is; using `.tar.zst` for the packages in Miniconda itself might roughly negate this for the compressed installers, though.

Another issue with using `.tar.zst` files in Miniconda directly is of course that `constructor` has to be taught to handle those archives as well. Likely, `zstd` has to be bundled for this since, unlike `bzip2`, I don't think `zstd` can't be assumed to be preinstalled.

kalefranz:
Another complication here will be that anaconda.org will have to be taught about the new extension.  That could end up being fairly time consuming.

kalefranz:
> I may do a full rebuild of miniconda making nothing but tar.zst files, see what happens and fix whatever I run into next.

There's been a `.tar.bz2` assumption in the ecosystem from the *very* beginning.  I'm afraid this has far-reaching and right now unknowable consequences and impact on code.  It's deep enough that we have to be thoughtful about how we roll it out.  All sorts of downstream tools will need to be adjusted.  Conda-build, constructor, mirror tools, AE Repo4, AE 5, conda-forge tools, etc. etc.

So with that, I'm not sure how feasible this is going to be to backport to 4.5.x or earlier.  I definitely think this effort would be easier to target at the 4.6.x branch (right now master, but we'll cut off 4.6.x soon).  4.6.x removes `Dist`, a major difference from 4.5.x.  All the Dist information is/was tied to and derived from the package filename.  There are far fewer package filename "things to worry about" in 4.6 than in 4.5.

mingwandroid:
I havee that libxml2 links to ICU. We should fix that mistake.

mbargull:
Oh yes, I'm fully aware that there are *tons of* downstream projects that needs to be changed, too. My thought process was rather that it would make sense to push support for this to the user side of things as soon as possible to have them prepared for future `.tar.zst` deployments.
I don't expect you to switch to `.tar.zst` on any official channels *any time soon*. I'd just like to avoid that awkward situation when everything is prepared and battle-tested but you can't start deployment because of an "oh, we still have to support `conda <4.6` users for X years or so; no can do, sorry".
Though, I initially thought backporting would involve only minimal changes but I kind of forgot (or rather repressed? ;P) the `Dist` tie-ins...

> I hate that libxml2 links to ICU. We should fix that mistake.

I don't really know much of how ICU is used in libxml2, but I wonder how many downstream projects depend on/expect ICU-enabled `libxml2` builds. So I'm not sure what the lesser evil is; a `libarchive` build w/o XAR support or `libxml2` w/o ICU...

mingwandroid:
This PR is targeted at the 4.5 branch already.

Do you want it for 4.6 instead?

mingwandroid:
@kalefranz any comments on this? I'm keen to get it merged ASAP.

kalefranz:
I don't think I can support merging this into 4.5.x.  It must be 4.6.x at a minimum, but we have to talk through as a group and with product anything new that's going into 4.6.x at this point (just like we did for PowerShell).

The PR will change when pointed at master because we have less to worry about with `Dist`.

How should conda handle the situation where both a `.tar.bz2` and `.tar.zst` file by the same name exists in repodata?  How should conda handle the situation where both files exist in a package cache?  Especially since extracting both would cause the second to effectively clobber the first.

The file extension of a conda package is very user-facing.  I want to think through what we're doing here very carefully.  If we add yet another file extension a couple years from now, we'd be in a world of three different file extensions being able to be conda packages (though not guaranteed to be conda packages).  For me, a proliferation of file extensions looks sloppy, unplanned, unprofessional, and generally like we really don't know what we're doing. I'm all for adding new compression algorithms, but I also feel strongly we should take this opportunity to introduce one new file extension that will be the conda package file extension from here on out, so that we can abstract and hide details under the cover as our community evolves.  Reference the .deb package format https://en.wikipedia.org/wiki/Deb_(file_format) for a clean and straight-forward way we could accomplish this.

mingwandroid:
This PR addresses the outer format. If internally to the outer format you want to split the metadata from the actual package files in two .tar files, then that's great but it's got nothing to do with this PR.

The issues about metadata do need consideration, at present (at least of 4.5.x branch), the internal conda index is keyed from the package name without the file extension so one will clobber the other. This doesn't technically matter though because the clobbering replaces the data with the exact same data. We have an implicit preference for the package format which is actually used (unpacked) encoded implicitly as the first entry of `CONDA_TARBALL_EXTENSIONS`. The stomping concerns aren't really important, once we switch the preference to `.tar.zst` it'll just stomp the old unpacked `.tar.bz` package with the contents of the `.tar.zst` package, but again, the contents are exactly the same (in the case of a 1:1 swap without version update).

I agree with having one package format, and propose to convert all existing packages to `.tar.zst` packages so we can cut down on our bandwidth and our users install time and storage requirements.

Having support in as early as possible will increase our ability to take full advantage of this new outer format. Is it not worth my while trying to get a different PR into each branch for this?

However, the PR is clearly incomplete and arguably we want to use rust for handling tar, zstd and bz2. If you want to take that stance then great, let's crack on, but please, do not hold this up for some way-in-the-future conda package version #2.

mingwandroid:
@msarahan queried whether we were decided on `zstd` here, but none of this commits us to `.tar.zst`, arguably, before I start on rust we should figure out which format we want. My impression was zstandard is the current frontrunner, but brotli was also impressive. There's no brotli support in libarchive at present though which could present me some troubles but there is a very lightweight pure-rust no-stdlib implementation from dropbox that I think we could use, so ...

libarchive supports:

```
    "bzip2 >=1.0.6,<2.0a0",
    "lz4-c >=1.8.1.2,<1.9.0a0",
    "lzo >=2.10,<3.0a0",
    "xz >=5.2.4,<6.0a0",
    "zlib >=1.2.11,<1.3.0a0",
    "zstd >=1.3.3,<1.3.4.0a0
```

.. should we perform some benchmarking? I can probably use my work on `conda-pack` to get some typical values we can expect from each format.

mingwandroid:
I believe files that changing extensions is a bit dodgy, but that's my opinion. It doesn't indicate to the end user what tools they can use to manipulate them. I know you'll say "use conda", but there are many legitimate usecases for using `tar -cJ` or it is `tar -xf` or `tar -xL`, I don't know, the extension just says `.condapkg`.

kalefranz:
Maybe I should be more clear with what I'm proposing, rather than just pointing to #7529.

Let's for now say the conda package extension is now going to be `.condapkg`. I don't really care what the actual letters are, as long as it's something that (1) can stay consistent across time as a user-facing interface, and (2) actually identifies the file as a conda package rather than something generic that may or may not be a conda package.

The format of the package would be an uncompressed tar archive with three files.

1. A file named `version` that for now just contains a plain text `1`.
2. A file named `info.tar.zst` (or `info.tar.bz2` or whatever other compression type you want) that contains the contents of the `info/` directory.
3. A file named `data.tar.zst` (or whatever else we want to call it) that contains all the components of the installed conda package.

I don't think this is a way far out conda package format v2.  This is something we could implement with basically the same amount of effort that would be required for this zstd work.  We're also then giving us a clean evolution path for future improvements.

mingwandroid:
I would expect you want a manifest file containing the sha256 of the other files too?

This tar.zst stuff for conda was a tiny amount of work (read the PR), ok, not entirely finished, and I believe what you are talking about is a much larger effort. Still, if we can find resources for that soon then that's great.

mingwandroid:
I was never unclear about your .deb proposal BTW, I just don't believe that you should necessarily merge it with this work.

kalefranz:
> I believe files that changing extensions is a bit dodgy

We should probably take queues from other package managers.  Look at the decisions they have made, and try to understand why.

When I was first exploring deb packages, the first thing I did was `tar -xf pkgname.deb`, and it worked.  When I was first exploring whl packages, the first thing I did was `tar -xf pkgname.whl`, and it worked.  I didn't know if it would when I first ran those commands.  It's awesome that it did.  What I'm proposing here would follow the same pattern.

> It doesn't indicate to the end user what tools they can use to manipulate them.

How do users know whether a `.tar.bz2` / `.tar.zst` / `.tar.brotli` file is a conda package or not?  We need to weigh how often a package builder actually gets confused with `tar -xf` with the frequency a user would want to identify a file or url being a conda package or not.

kalefranz:
I just downloaded a ruby gem.

```
$ wget https://rubygems.org/downloads/pony-1.4.gem
$ tar -xf pony-1.4.gem
$ ls -al
drwxr-xr-x    5 kfranz  staff   160B Oct  4 16:42 ./
drwx------+ 107 kfranz  staff   3.3K Oct  4 16:42 ../
-rw-r--r--@   1 kfranz  staff   6.4K Dec 31  1969 data.tar.gz
-rw-r--r--@   1 kfranz  staff   722B Dec 31  1969 metadata.gz
-rw-r--r--@   1 kfranz  staff   9.5K Oct  4 16:42 pony-1.4.gem
```

mingwandroid:
I don't doubt there's a lot of precedent here, I just don't agree with it, but I do not care enough to object as I see the upsides.

mingwandroid:
If we can get resources assigned to tackle your new format then great. @msarahan and I can take care of the `conda-build` side I reckon but if you want I can see if I have time to do it in both `conda` and `conda-build`. I am still keen that we don't preclude `conda 4.5` users do you think that is achievable?

mingwandroid:
I just don't want this to go cold.

kalefranz:
I can do the conda part. I do want to have sign-off from @spara though before we get too far along.

I don’t think this is feasible for 4.5.x wherher we take the archive approach or not. We get ourselves into trouble when we try to introduce new capabilities into old versions of conda. If users want new features, then they’re going to have to use new versions of conda. If users want absolute stability, then we allow them to use pinned versions of conda. You don’t get to have your cake and eat it too. New packages with new capabilities and features means a version of conda that was designed to support that capability.

Why do users want to stay on old versions of conda? It’s because we have a history of breaking things through new features. We don’t solve that problem by backporting new features to old versions. We solve that problem by getting better at QA and spending time thinking carefully through upgrade paths.

msarahan:
Thanks for trailblazing this, @mingwandroid - closing as we now have support for the .conda format which uses zstd internally.

mingwandroid:
I'm so glad we got this done (despite the libarchive and free hassle). Congrats all and congrats to our team for the AD 2019.07 release, it was difficult, by you all did great! Wish I had contributed a bit more recently!

Issue #7716, opened 2018-08-30
Colon in environment name results in misconfigured PATH


Machione:
Installed a fresh version of Miniconda in `/home/ryan/Conda` and created a new environment with a colon (:) in its name. Activating my new environment and running Python results in the base Python installation being called (`/home/ryan/Conda/bin/python`) rather than Python installed in my new environment (`/home/ryan/Conda/envs/.../bin/python`), meaning that any installed Python packages are not importable. Perhaps Conda should not allow colons to be present in environment names? Or alternatively, this limitation should be made clear in documentation and a meaningful warning printed to the user when they try to do so.

## Current Behavior
Environment names containing colons result in the PATH environment variable being incorrectly set upon activation. This is due to colons being used as a separator between paths in PATH. Escaping of colons in PATH is not possible, and therefore no path in PATH should contain a colon.
<http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03>
<https://stackoverflow.com/questions/14661373/how-to-escape-colon-in-path-on-unix>

### Steps to Reproduce
```
$ conda create -n my:env python
```
Completes without error or suggestion that the user may want to reconsider their choice of environment name. All expected packages installed in `/home/ryan/Conda/envs/my:env`.
```
$ source activate my:env
```
Again, completes without complaining. Confirmed successfully working within my:env due to the (my:env) prefix on the prompt.
```
$ which python
```
Returns `/home/ryan/Conda/bin/python`, not the Python installed in the newly created environment which is located in `/home/ryan/Conda/envs/my:env/bin/`.
```
$ echo $PATH
```
Reveals the true problem; namely that this environment variable is set to `/home/ryan/Conda/envs/my:env/bin:/home/ryan/Conda/bin:...`. Notice the additional colon that separates what should be a single path `/home/ryan/Conda/envs/my:env/bin` into `/home/ryan/Conda/envs/my` and `env/bin`, which are illegitimate.

## Expected Behavior
An error thrown to the user upon the attempted creation of an environment containing a colon in its name. The error should describe that such names are invalid.

## Environment Information
<details><summary><code>`conda info`</code></summary><p>

```
     active environment : my:env
    active env location : /home/ryan/Conda/envs/my:env
            shell level : 1
       user config file : /home/ryan/.condarc
 populated config files : 
          conda version : 4.5.4
    conda-build version : not installed
         python version : 3.6.5.final.0
       base environment : /home/ryan/Conda  (writable)
[...]

PR #7801, opened 2018-09-25
Env name with colon is not supported.(#7716)

makbigc:
When creating an environment name with colon, an error is thrown.


makbigc:
I just signed the CLA.

msarahan:
Makes sense to me - I wonder how many people have tried to do this.  Scary stuff!  Thanks @makbigc

Issue #7699, opened 2018-08-26
Solver improvements: 5) explore ways to invoke the SAT solver more incrementally


mbargull:
ref: https://github.com/msoos/cryptominisat/issues/513#issuecomment-415503020

@mcg1969 in https://github.com/msoos/cryptominisat/issues/513#issuecomment-415226906
> One challenge we face here is that while our problems are somewhat incremental, they do involve some backtracking. From a mathematical point of view, our model is actually a multi-objective, pseudoboolean optimization problem. We start with a set of mandatory logical clauses that must be satisfied; then we proceed by performing a sequence of pseudoboolean optimization passes, implemented with bisection. We handle the conversion of the pseudoboolean inequalities to standard logical form using a BDD approach.
> 
> So at first glance we could exploit an incremental approach, but many of our bisection steps are unsatisfiable. When that happens, we have to *roll back* the latest set of incremental clauses, and apply a less strict set of clauses (e.g., we relax the pseudoboolean inequality). The solvers I've seen that implement incremental solver haven't (at least, by my reading) supported the kind of state-saving or rollback capability needed to handle this efficiently—and Picosat certainly doesn't—so we've simply had to feed a sequence of entirely independent SAT problems to the solver.

@msoos in https://github.com/msoos/cryptominisat/issues/513#issuecomment-415503020
> Ahhh... let me address all of your comments here, one by one. First, about rollback:
>
> > Thee solvers I've seen that implement incremental solver haven't (at least, by my reading) supported the kind of state-saving or rollback capability needed to handle this efficiently
> 
> So... there is rollback ;) Even in picoSAT, though probably I shouldn't tell you that :D The trick is the following. You have the core clauses:
> ```
> a V b V c
> -a V b
> ```
> Now you want to add new clauses that you may want to remove:
> ```
> f V g
> -f V e
> ```
> You have to do this:
> ```
> x V f V g
> x V -f V e
> ```
> and then you call the solver with:
> ```
> solve_with_assumptions(x = false)
> ```
> This will effectively force the original clauses to be used (since X is false, so the solver sees only the literals of the clauses you wanted to add). To remove the clauses, you simply add to the solver:
> ```
> add_clause(x = true)
> ```
> It's kinda tricky but cool. Where CMS excels is that it will do the *right thing* when it comes to assumed clauses and variables. Picosat would too from a correctness perspective, but CMS is a lot more scalable.
> 
[...]

Issue #7697, opened 2018-08-26
Solver improvements: 3) store clauses in a more memory efficient way


mbargull:
ref: https://github.com/conda/conda/pull/7676#issuecomment-415810372
> - Decide whether we should/can use `ClauseArray` instead of `ClauseList`
>   - Goal: Reduce memory.
>   - Goal: Avoid costly processing of list of tuple of ints in SAT solver's Python interface.
>   - Requirement: Try to make it faster so it's not *that much* slower than `ClauseList`.
>   - Alternative: See if we can avoid storing clauses at all in the Python part. refs: https://github.com/msoos/cryptominisat/issues/513#issuecomment-415503020 and https://github.com/msoos/cryptominisat/issues/513#issuecomment-415505531

If we implement support for reading clauses as provided by `ClauseArray.as_array()` in Pycosat and CryptoMiniSat (https://github.com/msoos/cryptominisat/pull/515), this can save a good chunk of memory, see https://github.com/conda/conda/pull/7676#issuecomment-414338896 and also make passing the clauses to the SAT solvers faster since they wouldn't have to unpack all those Python tuples.
That speed improvement is currently negated by the fact that adding clauses to a `ClauseArray` instance is slower than adding them to a `ClauseList` instance. If we translate the part of the code where the majority of clauses are added to compiled code, we can avoid that slowness and would even be much faster than when `ClauseList` is used, xref: https://github.com/conda/conda/issues/7696.

Additionally, it makes very much sense to fuse the `ClauseArray` (`ClauseList`) and `SatSolver` classes which have been introduced in https://github.com/conda/conda/pull/7676. That way we could easily use a uniform interface for adding clauses as well as restoring the solver state. So we might even get away with not storing the clauses at all on our part but potentially just let the SAT solvers do all of that for us. Very good suggestions for strategies to rollback the solver state are given by @msoos in https://github.com/msoos/cryptominisat/issues/513#issuecomment-415503020 ff.

mbargull:
Support for `array.array` as and input to `pycryptosat.Solver.add_clauses` is in PR https://github.com/msoos/cryptominisat/pull/519.
It would be good if someone could
- port this over to `pycosat`,
- add support for https://docs.python.org/3/c-api/buffer.html,
- check that I didn't mess anything up in there since I'm not too familiar with Python's C-API.

mbargull:
[...]

PR #7811, opened 2018-09-28
add undocumented sat_solver config parameter

mbargull:
The first two commits are not related to the configuration parameter and are just to move the stored clauses to the `SatSolver` instances (helpful in regards to https://github.com/conda/conda/issues/7697 and https://github.com/conda/conda/issues/7699). With CMS `5.6.5` we can then easily use `conda.common.logic.ClauseArray` without affecting the Pycosat interface that doesn't yet support that.

The main objective for this PR is to get some configuration parameter into the next prerelease versions so that people can easily try out the CryptoMiniSat solver, e.g., via
```sh
conda install -ynbase conda-forge::cryptominisat
export CONDA_TIMED_LOGGING=1
export CONDA_SAT_SOLVER=pycryptosat
conda create -vvdnx anaconda
```
It's obviously not supposed to be a configuration that is meant to stay, thus undocumented. But if it makes it into a prerelease version, we should tell people about it and encourage them to test it out.

(This should work out of the box already, but feel free to apply any changes you deem necessary to this branch. I currently don't have time to look into it any further.)

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -206,6 +206,7 @@ class Context(Configuration):
     # ######################################################
     deps_modifier = PrimitiveParameter(DepsModifier.NOT_SET)
     update_modifier = PrimitiveParameter(UpdateModifier.UPDATE_SPECS)
+    sat_solver = PrimitiveParameter(None, element_type=string_types + (NoneType,))
 
     # no_deps = PrimitiveParameter(NULL, element_type=(type(NULL), bool))  # CLI-only
     # only_deps = PrimitiveParameter(NULL, element_type=(type(NULL), bool))   # CLI-only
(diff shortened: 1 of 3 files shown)


conda/resolve.py
```diff
@@ -27,6 +28,8 @@
 Unsatisfiable = UnsatisfiableError
 ResolvePackageNotFound = ResolvePackageNotFound
 
+get_sat_solver_cls = memoize(get_sat_solver_cls)
```
kalefranz: Nice job using memoize here rather than importing it into conda.common :)

conda/base/context.py
```diff
@@ -206,6 +206,7 @@ class Context(Configuration):
     # ######################################################
     deps_modifier = PrimitiveParameter(DepsModifier.NOT_SET)
     update_modifier = PrimitiveParameter(UpdateModifier.UPDATE_SPECS)
+    sat_solver = PrimitiveParameter(None, element_type=string_types + (NoneType,))
```
kalefranz: If we were to make this more official, I’d probably want to turn it into an enum, following the example for `channel_priority`. It’s fine for now though.

PR #7814, opened 2018-09-28
add progressbar to download process

wolfv:
I was talking about this on the chat a couple of days ago. 

I think it's nice to have a progress bar for downloading the repo data to get some information where time is spent, and also make sure that you're connected to fast internet, server down etc.

this is just an initial attempt with some questions:

I was wondering about the special casing of repo.anaconda.com (I checked and the zipped file is also available for conda-forge).
Also, is there a policy for third party packages such as tqdm?

Otherwise I guess it's feasible to write conda's own progress bar.

Cheers!

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -16,6 +16,7 @@ from os.path import dirname, isdir, join, splitext
 import re
 from time import time
 import warnings
+from tqdm import tqdm
 
 from .. import CondaError
 from .._vendor.auxlib.ish import dals
@@ -414,7 +415,7 @@ def fetch_repodata_remote_request(url, etag, mod_stamp):
     if mod_stamp:
         headers["If-Modified-Since"] = mod_stamp
 
-    if 'repo.anaconda.com' in url:
+    if '.anaconda.com' in url:
         filename = 'repodata.json.bz2'
         headers['Accept-Encoding'] = 'identity'
     else:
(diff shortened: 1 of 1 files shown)


kalefranz:
tqdm is already vendored in and used. So that part isn’t an issue. The master branch, which is conda 4.6, and at the first beta release in `conda-canary` right now, splits out spinners for repodata loading from solve time, so there will be a little more info there.

The problem with this approach is how we would present the progress bar if/when we shard repodata by package name and then recursively download the smaller shards. In that case most of the responses will be 304s and so no content transfer. And you furthermore can’t do a progressbar for the whole procedure because with its recursive nature you don’t know ahead of time where the endpoint is.

We’ve always used repodata.tar.bz2 from repo.continuum.io / repo.anaconda.com. It’s historic, and at one point was a bit political. .tar.bz2 is just all around more inefficient, and so we assume web serves are going to respect request’s default compression gzip compression type ask.

wolfv:
Thanks for the reply!

Is there some docs on what you exactly mean by "shard repodata by package name"? Is that going to be a new protocol which downloads only a diff of the repo data?

msarahan:
@wolfv yes, sorry for the delay.  "sharding" is one approach we're hoping to implement soon to help with how conda is (not) behaving well with increasingly large repodata.  Rather than downloading and then loading complete repositories, we were hoping to shard it by package name to start.  That would allow us to make more requests, but for smaller data.  As @kalefranz points out, the progress bars are not going to be terribly helpful there, because you won't know beforehand how many requests you'll have to make.

PR #7826, opened 2018-10-03
Allow expansion of environment variables

### LINKED ISSUE #7805: Add expansion of environment variables in `channels` config key
As per https://github.com/conda/conda/issues/1156#issuecomment-299913430, can we get expansion of environment variables in the `channels` config key? This would be useful for me, as I can then add the following lines to my `.condarc` file:

```
channels:
  - https://${USERNAME}:${PASSWORD}@my.private.conda.channel
```

teake:
Aims to resolve #7805.

Expansion is only enabled for `custom_channels` at the moment. I will expand this to the other config parameters mentioned in #7805.

@kalefranz Should I update the docs as well in case this lands?


teake:
Ok, I've finished a first iteration. Some observations and thoughts:

- `envs_dirs` and `pkgs_dirs` already were expanded ([here](https://github.com/conda/conda/blob/5e1629ef61a787d50fce25f2dc5df26b6967d9af/conda/base/context.py#L392) and [here](https://github.com/conda/conda/blob/5e1629ef61a787d50fce25f2dc5df26b6967d9af/conda/base/context.py#L397) respectively). Expanding twice shouldn't hurt but isn't very nice. I'm not quite sure what the best way forward is with these two, since you do want to expand things like `~` in paths.
- Also not quite sure if [`conda.common.configuration.Configuration#check_source`](https://github.com/conda/conda/blob/5e1629ef61a787d50fce25f2dc5df26b6967d9af/conda/common/configuration.py#L805) should expand environment variables before calling `typify_data_structure`. Either way, I didn't add it.
- Not all newly expanded config parameters in the context have unit tests for expansion (though all three subclasses of `conda.common.configuration.Parameter` do).

Any feedback is welcome!

teake:
@kalefranz bump?

teake:
This is getting a bit stale, which is fine since this isn't high priority. But I'd appreciate a bit of expectation management, just so it doesn't slip of my radar. @msarahan do you have an idea when I can expect a review? Post-4.6?

msarahan:
I kind of doubt that the one appveyor failure is related, but can you please confirm that?  https://ci.appveyor.com/project/ContinuumAnalyticsFOSS/conda/build/1.0.6898/job/lmwm405bmknsio47#L1043

teake:
Doesn't look related. IIRC there were some Powershell PRs being merged around the time I created this one. I can re-merge the master branch into this one and I think this particular error should go away.

PR #7849, opened 2018-10-12
Allow configuring instrumentation record file location

geremih:
Currently all instrumentation is written to the same file. This makes it harder to look up the records retrospectively and can clobber the same file if you are creating multiple envs.

This is just one possible solution, another one would be to fix the file to be relative to the env (if so, where?).


msarahan:
I think this is fine - the test failures look like old ones that should now be fixed.  You've signed the CLA on other work.  I'm re-running the tests in the hopes they'll magically fixe themselves, but I'll merge either way.  Thanks, and sorry for the delay.

kalefranz:
I think this PR forgets to call `expanduser()` (or `expand()`) on `~`.

geremih:
@msarahan can we get this merged? wondering if it can still make it into 4.6.0

msarahan:
Sorry, we really, really don't want anything to delay 4.6.0 any further.  We're making last-minute fixes and hoping for a release Monday or Tuesday.  This can go in afterwards.

msarahan:
Thanks for your patience.  Merging.

Issue #7870, opened 2018-10-18
conda update conda fail


H3zi:
~ conda update conda
Solving environment: failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/exceptions.py", line 819, in __call__
        return func(*args, **kwargs)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/cli/main.py", line 78, in _main
        exit_code = do_call(args, p)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/cli/conda_argparse.py", line 77, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/cli/main_update.py", line 14, in execute
        install(args, parser, 'update')
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/cli/install.py", line 235, in install
        force_reinstall=context.force,
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/solve.py", line 518, in solve_for_transaction
        force_remove, force_reinstall)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/solve.py", line 451, in solve_for_diff
        final_precs = self.solve_final_state(deps_modifier, prune, ignore_pinned, force_remove)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/solve.py", line 180, in solve_final_state
        index, r = self._prepare(prepared_specs)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/solve.py", line 592, in _prepare
        self.subdirs, prepared_specs)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/index.py", line 215, in get_reduced_index
        new_records = query_all(spec)
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/index.py", line 184, in query_all
        return tuple(concat(future.result() for future in as_completed(futures)))
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/subdir_data.py", line 95, in query
        self.load()
      File "/home/heziz/miniconda/lib/python2.7/site-packages/conda/core/subdir_data.py", line 149, in load
        _internal_state = self._load()
[...]

PR #7878, opened 2018-10-20
FIX: possible fix for #7870

mrakitin:
Credits to @randolf-scholz for the possible quick solution, proposed in https://github.com/conda/conda/issues/7870#issuecomment-430948096.


mingwandroid:
We will fix this this bad metadata, but the PR isn't really harmful, but but I'd like to take this opportunity to mention that the `anaconda` channel that is causing trouble should not be used, ever. You should use `defaults' instead.

travishathaway:

PR #7894, opened 2018-10-22
disable timestamp prioritization 4.6.x

kalefranz:
Behavior remains the same for conda-build solver code path.


msarahan:
CB tests look fine to me.  One is openssl not being pinned and showing a new version now.  The other is a bizarre absent /usr/bin/file executable.

PR #7895, opened 2018-10-23
disable timestamp prioritization 4.5.x

kalefranz:


msarahan:
disabling this causes multiple solutions to show up.  I believe @jjhelmus came up with a better solution in #8012.  Closing this.

PR #7949, opened 2018-11-09
Move docs to conda repo

soapy1:


msarahan:
CI failures unrelated.  Merging.  @soapy1 we'll need to get this set up on readthedocs.  I'll chat with you about that in the office.

nehaljwani:
No preserve history? 😭

msarahan:
That's definitely essential for code changes.  I don't think it's necessary for docs changes.  Nevertheless, if you think it is important, we can add another sprint item for restoring the git history for these files.

hugobuddel:
Where is the documentation copied from?

I'm trying to see where this line comes from:
https://github.com/conda/conda/blob/8bc852ae825595cc00116028f5b0a30533359f4f/docs/source/user-guide/tasks/manage-environments.rst#id15

> This creates the myenv environment in /envs/. This environment uses the same version of Python that you are currently using, because you did not specify a version.

It seems that this is wrong, because if you don't specify Python, you won't get Python at all. So I was wondering whether Python actually was always included in some earlier version of conda (which is behavior that I recall). (See #9982 )

But I can't figure out for which version of conda this line was originally added to the documentation because there is no history.

hugobuddel:
Answering my own questions:

* The documentation is copied from https://github.com/conda/conda-docs
* The last version of conda-docs that still has the documentation: https://github.com/conda/conda-docs/tree/ad5ac6d2610f31c7f7638a0986c142fb5bffacc0
* The commit that added the sentence that I was interested in: https://github.com/conda/conda-docs/commit/247b689f6dfc6ecc7fa92445a56ec924502127ab That was in June 2015, so conda 3 era.

PR #7951, opened 2018-11-10
Allow not more than one " to preceed conda[-_]

### LINKED ISSUE #7950: Conda Init Bug in v4.6.0b1
While exploring #7881 , I found a small bug in the regex pattern matching at: https://github.com/conda/conda/blob/7a93664267488adc5052efef38614fab537da446/conda/core/initialize.py#L1157-L1166

The regex pattern assumes that I would not have quotes in my already existing value in the registry key. For example, if I have the current value set as...

```
"\path\to\custom\startup.bat" & "\path\to\prev_installation\condabin\conda_hook.bat"
```

... upon `conda init`, it will change the value to:
```
"\path\to\new_installation\condabin\conda_hook.bat"
```

Instead of:
```
"\path\to\custom\startup.bat" & "\path\to\new_installation\condabin\conda_hook.bat"
```

nehaljwani:
Fixes #7950

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1157,7 +1157,7 @@ def init_cmd_exe_registry(target_path, conda_prefix):
     hook_path = '"%s"' % join(conda_prefix, 'condabin', 'conda_hook.bat')
     replace_str = "__CONDA_REPLACE_ME_123__"
     new_value = re.sub(
-        r'(\".*?conda[-_]hook\.bat\")',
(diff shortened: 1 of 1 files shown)


nehaljwani:
Fix for failing CI jobs: https://github.com/AnacondaRecipes/repodata-hotfixes/pull/6

nehaljwani:
Looks like now only CB related failures are present.

msarahan:
cycling to re-run CI

msarahan:
I believe the CI failures are not new.  Merging.

PR #7961, opened 2018-11-14
Try to use `mmap` when hashing files

jakirkham:
If that doesn't work due to some unusual filesystems or OSes, fallback to the current strategy. ( https://github.com/conda/conda-build/pull/1764 )

Note: Have not benchmarked this and am unsure if I will have time to. However have found this to be a good solution historically. So am putting this up as a suggestion.

### DIFF
--- a/conda/gateways/disk/read.py
+++ b/conda/gateways/disk/read.py
@@ -76,6 +76,9 @@ def _digest_path(algo, path):
         except OSError:
             for chunk in iter(partial(fh.read, 8192), b''):
                 hasher.update(chunk)
+        except ValueError:
+            # File is empty
+            pass
     return hasher.hexdigest()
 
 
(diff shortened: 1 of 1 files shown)


jakirkham:
AFAICT the test failures are unrelated. Have raised issues ( https://github.com/conda/conda/issues/7973 ),  ( https://github.com/conda/conda/issues/7974 ), and  ( https://github.com/conda/conda/issues/7976 ) about the ones I've seen. There may be more I've missed.

In any event, would be great if someone could review this.

jakirkham:
Would be great to have a review if you have time, @kalefranz. This may be useful for people struggling with the slow verification time. However don't feel comfortable having them try this without some kind of review first.

travishathaway:
Hi @jakirkham,

I believe this pull request has become quite stale; therefore, I will be closing. Please re-open if you feel differently.

Thank you.

jakirkham:
Yeah think so. Thanks for cleaning these up Travis! 🙏

Issue #7952, opened 2018-11-10
Conda Init (v4.6.0b1 ) --system cmd.exe chokes


nehaljwani:
```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "c:\users\wani\downloads\repos\conda\conda\exceptions.py", line 1001, in __call__
        return func(*args, **kwargs)
      File "c:\users\wani\downloads\repos\conda\conda\cli\main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "c:\users\wani\downloads\repos\conda\conda\cli\conda_argparse.py", line 81, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "c:\users\wani\downloads\repos\conda\conda\cli\main_init.py", line 52, in execute
        anaconda_prompt)
      File "c:\users\wani\downloads\repos\conda\conda\core\initialize.py", line 106, in initialize
        run_plan(plan2)
      File "c:\users\wani\downloads\repos\conda\conda\core\initialize.py", line 552, in run_plan
        result = globals()[step['function']](*step.get('args', ()), **step.get('kwargs', {}))
      File "c:\users\wani\downloads\repos\conda\conda\core\initialize.py", line 1196, in init_long_path
        _write_windows_registry(target_path, "1", winreg.REG_DWORD)
      File "c:\users\wani\downloads\repos\conda\conda\core\initialize.py", line 1143, in _write_windows_registry
        winreg.SetValueEx(key, value_name, 0, value_type, value_value)
    ValueError: Could not convert the data to the specified type.

`$ C:\ProgramData\MinicondaX\Scripts\conda-script.py init --system cmd.exe`
```

PR #7989, opened 2018-11-25
Fix long path init/re-init

nehaljwani:
Fixes https://github.com/conda/conda/issues/7952

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1118,7 +1118,9 @@ def _read_windows_registry(target_path):  # pragma: no cover
 
     try:
         value_tuple = winreg.QueryValueEx(key, value_name)
-        value_value = value_tuple[0].strip()
+        value_value = value_tuple[0]
+        if isinstance(value_value, str):
+            value_value = value_value.strip()
         value_type = value_tuple[1]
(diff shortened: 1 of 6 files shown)


nehaljwani:
Fixes https://github.com/conda/conda/issues/7952

conda/core/initialize.py
```diff
@@ -1187,13 +1189,13 @@ def init_long_path(target_path):
     # win10, build 14352 was the first preview release that supported this
     if int(win_ver) >= 10 and int(win_rev) >= 14352:
         prev_value, value_type = _read_windows_registry(target_path)
-        if prev_value != "1":
+        if prev_value != 1:
```
kalefranz: Given the ambiguity here, whey not cast it just to be safe? `int(prev_value)`

kalefranz: Actually, the safer choice is probably `if str(prev_value) != "1":` in case `prev_value` is `None`

nehaljwani: Hmm... reading a registry value of type REG_DWORD should always return an int. But I understand that we wouldn't want a traceback in bad scenarios anyway.

PR #7991, opened 2018-11-26
WIP: verification transaction speedups

msarahan:
* disable per-file sha256 safety checks by default; add extra_safety_checks config option for it instead

This is meant to address user complaints that the verification step takes too long.

As a worst case scenario, I've been trying with our boost package, because it has so many files.  With that package, this PR cuts a 35 sec total time down to 20 sec.

As @kalefranz has mentioned, pip does do these checks, so by disabling this, we're less "safe" than pip: https://github.com/pypa/pip/blob/873662179aebbf5eacdf681078f47bbfe5ee6149/src/pip/_vendor/distlib/database.py#L735

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -127,6 +127,7 @@ class Context(Configuration):
                                                     DEFAULT_AGGRESSIVE_UPDATE_PACKAGES,
                                                     aliases=('aggressive_update_packages',))
     safety_checks = PrimitiveParameter(SafetyChecks.warn)
+    extra_safety_checks = PrimitiveParameter(False)
     path_conflict = PrimitiveParameter(PathConflict.clobber)
 
     pinned_packages = SequenceParameter(string_types, string_delimiter='&')  # TODO: consider a different string delimiter  # NOQA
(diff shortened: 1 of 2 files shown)


msarahan:
There's a little bit of room for potential improvement by cleaning up the PrefixData object.  It looks like we're getting a pretty significant amount of overhead from creation of those objects.  With boost, commenting out all of the PrefixDataV1 instantiations yields another 5 sec or so gain.

It may be beneficial to replace the many PrefixDataV1 instantiations with a dictionary, and perhaps preallocate space on that dictionary.

msarahan:
I experimented with parallelism in two ways:

Using the ThreadPoolExecutor in conda. If there were spare disk bandwidth, this would allow conda to work on multiple reads at once.

Using Dask’s multiprocess-based parallelism. If the problem were compute-bound, this would allow it to go faster.

This work unfortunately is out of scope for 4.6.0, and we'll circle back to it for a future release.

msarahan:
Closing - this went in with #8017 for the 4.5.x branch.

PR #8012, opened 2018-12-04
speed up solver by skipping timestamp minimization when not needed

jjhelmus:
Speed up the solver by skipping the timestamp minimization step when there is a single "simple" solution after the package count minimization step.

It may be possible to skip other minimization steps using similar logic but further testing is needed.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -849,6 +849,19 @@ class Resolve(object):
             constraints = r2.generate_spec_constraints(C, specs)
             return C.sat(constraints, add_if)
 
+        # Return a solution of packages
+        def clean(sol):
+            return [q for q in (C.from_index(s) for s in sol)
+                    if q and q[0] != '!' and '@' not in q]
+
+        # Determine if there are multiple "simple" solutions
+        def has_multiple_simple_solutions(solution):
+            psolution = clean(solution)
+            nclause = tuple(C.Not(C.from_name(q)) for q in psolution)
+            if C.sat((nclause,), True) is None:
+                return False
+            return True
+
         r2 = Resolve(reduced_index, True, True, channels=self.channels)
         C = r2.gen_clauses()
         solution = mysat(specs, True)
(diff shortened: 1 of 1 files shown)


conda/resolve.py
```diff
@@ -849,6 +849,19 @@ def mysat(specs, add_if=False):
             constraints = r2.generate_spec_constraints(C, specs)
             return C.sat(constraints, add_if)
 
+        # Return a solution of packages
+        def clean(sol):
+            return [q for q in (C.from_index(s) for s in sol)
+                    if q and q[0] != '!' and '@' not in q]
+
+        # Determine if there are multiple "simple" solutions
```
soapy1: Can you expand on what a "simple" solution is

jjhelmus: 'simple' or 'obvious' solutions are defined as solution where none of the packages in
the current best solution are used.  I should add this to the comment.

jjhelmus:
Note that this is against the `4.5.x` branch.  

This change can knock some significant time off solve time.  For example the wall time for 
`conda create -nt  --dry-run anaconda` on linux-64 without this change is 270 seconds, with the change is 195 seconds.

conda/resolve.py
```diff
@@ -923,19 +936,18 @@ def mysat(specs, add_if=False):
...
         # Prune unnecessary packages
         eq_c = r2.generate_package_count(C, specm)
         solution, obj7 = C.minimize(eq_c, solution, trymax=True)
         log.debug('Weak dependency count: %d', obj7)
 
-        def clean(sol):
-            return [q for q in (C.from_index(s) for s in sol)
-                    if q and q[0] != '!' and '@' not in q]
+        multiple_solutions = has_multiple_simple_solutions(solution)
+        if multiple_solutions:
```
soapy1: Is there an easy way to show how this speeds up a solve? Just curious

jjhelmus: If `multiple_solutions` is False (i.e. there is a single solution) then the block of code is not executed.  That block, particularly the line `solution, obj6t = C.minimize(eq_t, solution)` line can take considerable time if a large number of packages are being considered.

jjhelmus:
@soapy1 I tried to clarify the logic being applied to conditionally skip the timestep minimization in 816f344.  Does the new flow and documentation help explain this behavior?

soapy1:
@jjhelmus looks great!

msarahan:
Good addition.  Thanks @jjhelmus

PR #8014, opened 2018-12-05
backport skip sha256 verification until multiprocessing can be enabled

nehaljwani:
Brings down env creation time considerably.


mingwandroid:
IMHO this code should be avoided during installs entirely and only run optionally by the user to diagnose problems.

It is the job of operating systems to handle this and they do that already.

nehaljwani:
Unfortunately, this is the wrong backport. Sorry, Ray.

PR #8015, opened 2018-12-05
Single pass pyc creation

### LINKED ISSUE #7621: allow for multi-path PathActions; build .pyc files in one pass per package
from @mingwandroid

> Is it known / expected that installing a few noarch: python packages takes >5 minutes (still running) on Windows?
>
> Looking in process hacker I see it spending all its time firing up new python interpreters to compile .pyc files. One instantiation of python per .pyc to be compiled.
>
> Can we not use a single process here?
>
> 10 minutes now, killing it.

jjhelmus:
Create pyc files in a single pass rather than using multiple calls to py_compile.  

closes #7621

### DIFF
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -27,8 +27,9 @@ from ..common.path import (get_bin_directory_short_path, get_leaf_directories,
 from ..common.url import has_platform, path_to_url, unquote
 from ..exceptions import CondaUpgradeError, CondaVerificationError, PaddingError, SafetyError
 from ..gateways.connection.download import download
-from ..gateways.disk.create import (compile_multiple_pyc, compile_pyc, copy, create_hard_link_or_copy,
-                                    create_link, create_python_entry_point, extract_tarball,
+from ..gateways.disk.create import (compile_multiple_pyc, compile_pyc, copy,
+                                    create_hard_link_or_copy, create_link,
+                                    create_python_entry_point, extract_tarball,
                                     make_menu, mkdir_p, write_as_json_to_file)
 from ..gateways.disk.delete import rm_rf, try_rmdir_all_empty
 from ..gateways.disk.permissions import make_writable
@@ -539,7 +540,7 @@ class CompileMultiPycAction(MultiPathAction):
             noarch_py_file_re = re.compile(r'^site-packages[/\\][^\t\n\r\f\v]+\.py$')
             py_ver = transaction_context['target_python_version']
             py_files = tuple((axn.target_short_path for axn in file_link_actions
-                        if noarch_py_file_re.match(axn.source_short_path)))
+                              if noarch_py_file_re.match(axn.source_short_path)))
             pyc_files = tuple((pyc_path(pf, py_ver) for pf in py_files))
             return (cls(transaction_context, package_info, target_prefix, py_files, pyc_files), )
         else:
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -546,10 +546,8 @@ class CompileMultiPycAction(MultiPathAction):
         target_python_version = self.transaction_context['target_python_version']
         python_short_path = get_python_short_path(target_python_version)
         python_full_path = join(self.target_prefix, win_path_ok(python_short_path))
-        target_prefix = join(self.target_prefix,
-                             self.transaction_context['target_site_packages_short_path'])
         compile_multiple_pyc(python_full_path, self.source_full_paths, self.target_full_paths,
-                             target_prefix, self.transaction_context['target_python_version'])
+                             self.target_prefix, self.transaction_context['target_python_version'])
         self._execute_successful = True
 
     def reverse(self):
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -192,34 +192,6 @@ class CreateInPrefixPathAction(PrefixPathAction):
         return join(prfx, win_path_ok(shrt_pth)) if prfx and shrt_pth else None
 
 
-@with_metaclass(ABCMeta)
-class CreateInPrefixMultiPathAction(PrefixMultiPathAction):
-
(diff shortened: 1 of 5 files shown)


jjhelmus:
@kalefranz We discussed this yesterday, would appreciate your feedback on these changes.

conda/core/path_actions.py
```diff
@@ -490,6 +530,79 @@ def reverse(self):
...
+        python_full_path = join(self.target_prefix, win_path_ok(python_short_path))
+        compile_multiple_pyc(python_full_path, self.source_full_paths, self.target_full_paths)
+        self._execute_successful = True
+
+    def reverse(self):
+        # this removes all pyc files even if they were not created
+        if self._execute_successful:
+            log.trace("reversing pyc creation %s", self.target_full_path)
+            for target_full_path in self.target_full_paths:
+                rm_rf(target_full_path)
```
msarahan: calling rm_rf one by one is probably slow here, but I'm not sure there's a good way around that right now.  If it is isolated as a bottleneck down the line, it may be better to collect up filenames and batch calls to the native shell rm utility.

msarahan:
Seems reasonable to me.  Are the existing tests sufficient?  Is it worthwhile to put a time limit test in that conservatively passes with the new code, but which would fail with the old code?

conda/core/path_actions.py
jjhelmus: Single file rm_rf was used it the original implementation and for non-noarch packages.  I do not expect this loop to be much of a bottleneck.

mingwandroid:
stdin is a pipe on Windows so works much the same, no limit.

jjhelmus:
I think the unicode characters in the prefix name and path were causing issues on Windows.  Hopefully escaping them and using the proper linesep fixes the issue.

jjhelmus:
The windows unit tests are now passing, the CI is timing out.  

An alternative to the special casing in e96463c seems to be encoding stdin using cp1252.  I do not know if this encoding depends on the locale setup of Windows or if it is universal. @mingwandroid would you happen to know?

mingwandroid:
cp1252 is a western encoding. Python has a function to get the filesystem encoding that I think you may need to use instead. on phone so cannot look it up easily ATM.

msarahan:
From Ray on Slack:

> sys.getfilesystemencoding() .. for python 3
May need to use ctypes GetACP() for py2 .. we do this somewhere already I think.
https://vstinner.github.io/painful-history-python-filesystem-encoding.html

msarahan:
sys.getfilesystemencoding is present in py27 also.  I'm trying to just use that instead of hardcoding cp1252.

mingwandroid:
Cool!

msarahan:
latest commit is all wrong, because you have to use the Python that is in the env, not the one that conda is running.  I think it may be best to take a snapshot of the files in the prefix, then run https://docs.python.org/3/library/compileall.html on everything, then remove anything extra that has come along, so that we wind up with only what we intended from this current action.

That should obviate the need to pass explicit file paths on stdin, which should simplify the issue.

mingwandroid:
You can call compile with an  argument for each filename too AFAIK.

jjhelmus:
The filesystem encoding may be different than the encoding used for stdin.  The [documentation for sys.stdin](https://docs.python.org/3/library/sys.html#sys.stdin), which is what `py_compile` is reading from, indicates that when the stream is not interactive, stdin uses a "ANSI code page" encoding.  I think this can be determined from `locale.getpreferredencoding()`, but I'd preferred to interrogate the interpreter itself to determine the encoding.

msarahan:
failing on spaces.  Trying to figure out proper quoting.

jjhelmus:
The scheme used in b3cb857 will fail to compile `.py` files if the paths use characters which cannot be represented in the stdin character encoding.  For example on windows a path may have UTF-8 characters which are not present in the cp1252 code page.  This would be very unusual and the old scheme may have also failed in these cases.

jjhelmus:
Using `compileall` has the potential to compile a lot of unnecessary python files which will then be removed. For a large environment this would happen for each noarch package installed.  Is this a performance concern?

mingwandroid:
You can call compile with a list of files as something, just remove the `-` from Jonathan's command line and add the files..

msarahan:
Yeah, that's why I put in that search for extra pyc files.  I have no idea what the performance implication is without some sort of benchmark.  It's pointed solely at site-packages, which I think dramatically limits potential problems.

Ray, I know that you *can* compile with a list of arguments.  The trouble is that that list of arguments is subject to this hell of encoding nonsense.  I'm trying to find ways to skirt that hell.

mingwandroid:
How about emit to a file as utf-8 a line per file, then write a simple script that reads that back in?

jjhelmus:
> How about emit to a file as utf-8 a line per file, then write a simple script that reads that back in?

`compileall` can take a file containing a list of file as input via the `-i` argument.  The question is where to store said file?

msarahan:
doesn't matter - some tempdir?

conda/gateways/disk/create.py
```diff
@@ -354,19 +348,22 @@ def compile_multiple_pyc(python_exe_full_path, py_full_paths, pyc_full_paths, si
...
+    with tempfile.NamedTemporaryFile('w') as tf:
+        for f in py_full_paths:
+            tf.write(f + "\n")
+        tf.flush()
 
-    # if the python version in the prefix is 3.5+, we have some extra args that make things nicer.
-    #    -j 0 will do the compilation in parallel, with os.cpu_count() cores
-    if int(py_ver[0]) >= 3 and int(py_ver.split('.')[1]) > 5:
-        command.extend(["-j", "0"])
+        command = ["-Wi", "-m", "compileall", "-q", "-i", tf.name]
```
jjhelmus: `-l` might be prudent

msarahan:
Sorry @jjhelmus I'm done for the day on this.  I think we're back where we started, but hopefully we know a bit more now.

PR #8016, opened 2018-12-06
reduce broadening of spec scope during get_reduced_index

msarahan:
get_reduced_index does a recursive collection of entries in repodata.  This
collection can be much smaller if we start by considering as many exact specs
as possible, and limit conda from broadening the search when a more exact match
is already in place.  This commit does so by first sorting specs by the number
of exact specs they in their own dependencies, then by taking that into account
with a pretty simple check for whether build string or version is already
present in the expanded collection, and omitting specs that would broaden the
search space.

A primitive benchmark shows that this cuts the time for 

```date && conda create --dry-run -n test_reduced_index python=3.7 anaconda=5.3.1 && date```

from 

Wed Dec  5 19:42:20 CST 2018
Wed Dec  5 19:45:18 CST 2018

to

Wed Dec  5 19:52:05 CST 2018
Wed Dec  5 19:53:30 CST 2018

### CODE REVIEW

**Comment by msarahan** [comment]
This seems to be a little too aggressive.  It ends up making many envs unresolvable:

```
date && conda create --dry-run -n test_reduced_index python=2.7 anaconda=5.2.0 && date
Wed Dec  5 20:13:42 CST 2018
Solving environment: failed

  - anaconda=5.2.0
  - qt==5.9.5=h02808f3_0

Current channels:

  - https://repo.anaconda.com/pkgs/main/osx-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/free/osx-64
  - https://repo.anaconda.com/pkgs/free/noarch
  - https://repo.anaconda.com/pkgs/r/osx-64
  - https://repo.anaconda.com/pkgs/r/noarch

looking for, navigate to

    https://anaconda.org

```

**Comment by msarahan** [comment]
Anaconda 5.2.0 error is unrelated.  Unfortunately, it appears that is just totally unsatisfiable regardless - qt 5.9.5 must have been removed.

**Comment by jjhelmus** [comment]
I do not think it is safe to prevent match specs from broadening during the creation of the reduced index.  During the creation of the reduced index it is not known if the current set of packages in the index include a satiable set or if the any of the packages in the set are required by the solution.  By preventing broadening, a tight matchspec from a package which will ultimately not be include in the solution has the potential to prevent a looser, and required, matchspec from being considered in a package that will be included in the solution.

**Comment by msarahan** [comment]
I think the key thing here is that we're restricting broadening of indirect stuff.  If you specify anaconda=5.3.1, you should be able to restrict your potential search space to the strict set of any packages associated with anaconda 5.3.1 (all of its exact pins).  We have anaconda 5.3.1 for py27 and py36.  Theoretically, that's about 600 packages.

Similarly, if you have a constraint for ``hdf5 >=1.8,<1.9.0a0``, you should not consider adding in constraints that are just ``hdf5`` - it's just bringing in crap that the solver has to eliminate later.

This PR is not dealing with actual values of ranges.  It is only excluding specs that are less exact (either build string missing, or version missing).  We should definitely test this, and the tests that are failing now need to be understood, but I really think this is safe.  I've gone through it with Kale, too, fwiw.

File: conda/resolve.py [hunk 1]
```diff
@@ -327,29 +327,62 @@ def filter_group(matches):
...
-                            slist.append(ms)
+        seen_pkgs = set()
+        seen_specs = set()
+        for explicit_spec in explicit_spec_list:
+            explicit_packages = set(self.find_matches(explicit_spec)) - seen_pkgs
+            for pkg in explicit_packages:
+                seen_pkgs.add(pkg)
+                if pkg not in reduced_index and self.valid(pkg, filter):
+                    reduced_index[pkg] = self.index[pkg]
+                    dep_specs = set(self.ms_depends(pkg)) - seen_specs
```
jjhelmus: If a spec was added to `seen_specs` but not added to `dep_specs` because of the broadening check for a particular `this_pkg_constraints` is it safe to ignore the spec for all other packages?

I'll try to come up with a example where this would cause a false unsat result.

msarahan: Latest update nests things such that this should be addressed.

conda/resolve.py
```diff
@@ -327,29 +327,67 @@ def filter_group(matches):
...
+            explicit_packages = set(self.find_matches(explicit_spec)) - seen_top_level_pkgs
+            for pkg in explicit_packages:
+                # what we have seen is only relevant within the context of a single package
+                #    that is picked up because of an explicit spec.  We don't want the
+                #    broadening check to apply across packages at the explicit level; only
+                #    at the level of deps below that explicit package.
+                seen_pkgs = {pkg}
+                seen_specs = set()
+                seen_top_level_pkgs.add(pkg)
+                if pkg not in reduced_index and self.valid(pkg, filter):
```
jjhelmus: An indentation could be saved here using:
```
if pkg in reduced_index or not self.valid(pkg, filter):
    continue
```

msarahan: good one.  it is very deeply indented, and this helps.

conda/resolve.py
jjhelmus: This is equivalent to `set(self.ms_depends(pkg))` since seen_specs above

conda/resolve.py
```diff
@@ -327,29 +327,67 @@ def filter_group(matches):
...
+                    while(dep_specs):
+                        ms = dep_specs.pop()
+                        seen_specs.add(ms)
+                        dep_packages = set(self.find_matches(ms)) - seen_pkgs
+                        for dep_pkg in dep_packages:
+                            seen_pkgs.add(dep_pkg)
+                            if dep_pkg not in reduced_index and self.valid(dep_pkg, filter):
+                                reduced_index[dep_pkg] = self.index[dep_pkg]
+
+                            new_specs = set(self.ms_depends(dep_pkg)) - seen_specs
```
jjhelmus: Shouldn't this (and the rest of the block) be inside the if statement?  There is no need to consider a packages depends if it is already in the index or invalid.

jjhelmus: I take that back, this is the correct location for this.

mcg1969:
I'm certainly glad to see this. It seems to me that `get_reduced_index` became less aggressive at some point. This kind of thing makes a lot of sense to me and I _thought_ we used to do something similar.

PR #8024, opened 2018-12-09
try scheduling removal at reboot on win.  Fall back to warning message

msarahan:
Windows thinks the core python stuff is in use while conda is running.  Our method of retrying does nothing more than slow things down on windows.  This avoids the retry stuff for permission errors, and tries to schedule removal at reboot.  Unfortunately, that only works for admin users (elevated prompts, probably) these days (https://marc.durdin.net/2011/09/why-you-should-not-use-movefileex-with-movefile_delay_until_reboot-2/)

so, we give up much sooner and post a warning.

### DIFF
--- a/conda/gateways/disk/delete.py
+++ b/conda/gateways/disk/delete.py
@@ -14,6 +14,11 @@ from .permissions import make_writable, recursive_make_writable
 from ...base.context import context
 from ...common.compat import PY2, on_win, text_type, ensure_binary
 
+if on_win:
+    import win32file
+    import pywintypes
+
+
 log = getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


msarahan:
@jjhelmus please review for inclusion in 4.5.12.  This should be safe, aside from possibly wasting a tiny bit of disk space on windows.  Nothing worse than it already is, though.  It should speed things up on Windows quite a lot, though.

conda/gateways/disk/delete.py
```diff
@@ -102,7 +107,21 @@ def move_path_to_trash(path, preclean=True):
...
+            unlink(path)
+        except (IOError, OSError) as e:
+            if on_win and e.errno == 13:
+                try:
+                    win32file.MoveFileEx(file_or_symlink_path, None,
+                                         win32file.MOVEFILE_DELAY_UNTIL_REBOOT)
+                    log.info("Windows thinks file %s is in use.  We have scheduled it for "
+                             "removal at next reboot." % file_or_symlink_path)
+                except pywintypes.error:
+                    log.info("Windows thinks file %s is in use.  We can't schedule it for removal"
```
nehaljwani: Should this be log.info or log.debug? If it is not an admin prompt, would it fill the end user's shell with this message multiple times, in an exponential decay?

jjhelmus: The `delete_trash` function logs failures to `log.info` so I think this is alright.

msarahan: No, it won't show it multiple times.  The point of this PR is to avoid the exponential decay completely - abort after the first failure.  It will show it once per file that fails to be removed.

msarahan:
Yes, so long as you are avoiding the exponential retry.  The point is that we absolutely know that this will fail on Windows, and we should not be delaying people by waiting the full time for every file only to have it fail exactly like we knew it would.

PR #8025, opened 2018-12-09
add compile_multiple_pyc function

msarahan:
Port #8015 to 4.5.x branch

PR #8027, opened 2018-12-10
Release 4.5.12

jjhelmus:


msarahan:
Windows failures are the pyc compilation failures from #8025 - we are ignoring those.  @jjhelmus let's add an xfail there so that we can run the rest of the win test suite.

PR #8034, opened 2018-12-13
Document allow_conda_downgrades config param

nehaljwani:


conda/base/context.py
```diff
@@ -592,7 +592,6 @@ def list_parameters(self):
             'bld_path',
             'conda_build',
             'croot',
-            'allow_conda_downgrades',
```
msarahan: why is this removed?  I think it has to stay here?

nehaljwani: This is the list of 'unlisted parameters'

msarahan: Of course... Sorry.

PR #8067, opened 2019-01-08
re-enable timestamps (they're reduced now anyway)

msarahan:
Tests removed here are ones that were time-bombed or otherwise ignored and needing attention.  We'll address them if they come up again.

Keeping timestamp optimization is important to avoid warnings about multiple solutions.  I have tested several of the past problems that disabling timestamps solved (anaconda metapackage, r-essentials) and things are working fine now with the changes that went into 4.5.12.


conda/core/path_actions.py
```diff
@@ -300,29 +301,11 @@ def verify(self):
...
-                       self.source_short_path,
-                       reported_sha256,
-                       source_sha256,
-                       )))
-
             try:
                 reported_size_in_bytes = source_path_data.size_in_bytes
             except AttributeError:
                 reported_size_in_bytes = None
+                source_size_in_bytes = 0
```
jjhelmus: I think this needs to one less indentation level.  As it is source_size_in_bytes may be undefined at line 330 if repored_size_in_bytes is 0.

PR #8093, opened 2019-01-14
Added building packages with conda skeleton

rrigdon:


soapy1:
moving to conda-build

PR #8099, opened 2019-01-14
optimize broader constraint reduction with better data structures

msarahan:


jjhelmus:
A frozendict is needed for hashing, rolling back that change.

jjhelmus:
@msarahan LGTM.  I refactored the `_broader` function, I found the new layout easier to follow and it might be a tiny bit faster by avoiding the variable assignment.

conda/resolve.py
```diff
@@ -309,19 +310,19 @@ def _get_strict_channel(self, package_name):
...
-        if matching_specs and (
-                # is there a version constraint defined for any existing spec, but not MS?
-                (any('version' in _ for _ in matching_specs) and 'version' not in ms)
-                # is there a build constraint on any of the specs, but not on ms?
-                or (any('build' in _ for _ in matching_specs) and 'build' not in ms)
-        ):
-            is_broader = True
-        return is_broader
+        if ms.name not in specs_by_name:
+            return False
```
msarahan: I generally prefer to have a single return from any given function, perhaps aside from an immediate bail  condition just before a long function (sparingly).  It's pretty minor here, but especially when returning multiple values, it can get especially annoying to maintain and debug.

https://stackoverflow.com/a/733858

Fine to keep here - not requesting any changes.

PR #8106, opened 2019-01-15
allow usage of conda clean --lock

### LINKED ISSUE #8104: conda clean --lock
Starting in version 4.6.0 `conda clean --lock` no longer works and instead returns:

    ArgumentError: At least one removal target must be given. See 'conda clean --help'

jjhelmus:
This command does nothing but should be retained until the CLI usage is
updated.

closes #8104

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -255,6 +255,11 @@ def _execute(args, parser):
                 verbose=not (context.json or context.quiet))
         one_target_ran = True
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks @jjhelmus !

msarahan:
Stubbing this out seems like the right solution to me.  I would also recommend adding a deprecation notice along the lines of "This flag no longer does anything and will be removed in Conda X.Y.Z."

PR #8117, opened 2019-01-18
get_reduced_index optimizations

msarahan:


conda/resolve.py
```diff
@@ -324,6 +335,7 @@ def _broader(self, ms, specs_by_name):
             return True
         return False
 
+    @memoizemethod
```
jjhelmus: `get_reduced_index` has a cache inside the function based on strict_channel priority and a frozen set of the spec argument.  Does the addition of the `memorizemethod` improve the performance significantly?

msarahan: Good point. Yes, this seems to have a large effect. Try it and see for yourself, though. Maybe the key for the existing cache is overly specific.

conda/resolve.py
```diff
@@ -478,9 +490,8 @@ def filter_group(_specs):
                 while(dep_specs):
                     ms = dep_specs.pop()
                     seen_specs.add(ms)
-                    dep_packages = set(self.find_matches(ms)) - seen_pkgs
+                    dep_packages = set(self.find_matches(ms)) - set(reduced_index2.keys())
```
jjhelmus: This change speeds up the solver considerable but is too aggressive in filter packages whose dependency specification need to be considered.  With this change if a decency package broadens a spec of top level package the dependency package's specs are never considered, even when that package is encountered for a different top level package with a broader spec that the first top level.  Essentially a tight specification can enforce itself on packages with broader specifications.

For example consider the packages `top`, `middle`, and `bottom`.  There are two version of `top`, one version of `middle` and two version of `bottom`.  The packages have the following dependencies:

`top-1.0` : `middle`, `bottom=1.5`
`top-2.0` : `middle`
`middle-3.1` : `bottom`
`bottom-1.5` : --
`bottom-2.5` : --

The preferred solution to `conda install top` would be: `top-2.0`, `middle-3.1` and `bottom-2.5`.  

With this change, if `top-1.0` appears first in the `add_these_precs2` tuple, then `middle-3.1` is added to `reduced_index2` but its dependency on an unrestricted `bottom` package is not added to `dep_specs` because it would broaden the existing specification of `bottom=1.5`. 

In the next iteration of the loop when the dependencies of `top-2.0` are examined, `middle-3.1` is not considered as it already in `reduced_index2`.  Therefore `bottom-2.5` is never added to `reduced_index2` and the solution will include `bottom-1.5` as it is the only version of the `bottom` package in the reduced index.

jjhelmus: I had to run through the loop on paper to find the issue here.  The problem is that a tight dependency constraint from a package which matches an explicit spec will remove from consideration a portion of the dependency tree that may be valid for another package from the explicit spec match which has a looser dependency.  

The following toy example shows how this can result in a reduced index that does not contain the necessary packages for a satisfiability environment.  I'm not sure if it is valid because the troublesome package may be removed from consideration in an earlier step.

Imagine a repository containing three packages, `a`, `b` and `c` with the following packages and dependencies:

```
a-1 : b, c==2  (this package cannot be installed but may still be considered)
a-2 : b
b-1 : c
c-1 : --
```

For the command `conda install a` both `a` packages may be candidates.  If `a-1` is processed first then `b-1` is added to the reduced index, but `c-1` is not because `b's` dependency on `c` would broaden the existing dependency from `a-1`.  When `a-2` is processed `b-1` is already in the reduced index so no additional packages are added.  The final reduced index would be:
`[a-1, b-1, a-2]`
Notice that no `c` packages are included.  The environment therefore cannot be created since there is no solution.  Yet `[a-2, b-1, c-1]` is a perfectly valid, and in fact only, solution.  

If `a-2` is processed first there are no problems, but the order of packages should not effect the reduced index.

msarahan: @jjhelmus good news!  I implemented your idea as a test, and in fact, it still works.  It does so because unsatisfiable solutions are pruned at https://github.com/conda/conda/pull/8117/files#diff-52a8935af92ea699b0cbb5feb84c19a2R374

I think you're on the right track though, and we should think through a few more pathologies before we merge this.  I'm optimistic that we can catch the corner cases and still get the benefit of the more aggressive reduction.

msarahan: OK, after adding a test for your first case, I've convinced myself that that first particular use case is sub-optimal in outcome.

Let me pose a question, though: is this enough of a corner case that we should ignore it for the sake of getting the speedup?  Can we avoid the pitfall of undesirable solutions by convention rather than brute force?

This question needs to be examined in light of the common practices of the packaging ecosystem.  Is it reasonable to say that:

1. newer versions of packages have more constraints than older ones?  (I believe yes; run_exports has made the presence of run constraints much more common)
2. repodata.json is currently sorted in ascending version order.  If we sorted by descending version order, we might further improve behavior here, by utilizing the newest (presumably most correct) behavior wrt constraints.  The problem with the order of interpretation of the reduced index is not so much one of "order should not matter" so much as it is one of "the correct order is used."  I believe the descending version order is correct and would lead to optimal outcomes in the vast majority of cases.

I will look into an inverse sorting method for conda index.

PR #8143, opened 2019-01-25
rpm and debian guide

rrigdon:
Added Sophia P's blog post over RPM and Debian to the installation section of conda


soapy1:
i think something should be added to index too https://github.com/conda/conda/blob/master/docs/source/user-guide/install/index.rst so that users can get to this page more easily

PR #8163, opened 2019-01-29
fix #8157 token leakage in errors and config output

### LINKED ISSUE #8157: CondaHTTPError leaks channel tokens
## Current Behavior
If the channel repo timeout occurs (or something else), the following gets outputted (potentially to a log that can be scraped)
```
CondaHTTPError: HTTP 504 GATEWAY_TIMEOUT for url <https://conda.anaconda.org/t/this-is-a-token/org/noarch/pkg-name.tar.bz2>
```

## Expected Behavior
if `conda info` hides the token, the token should be hidden in the error as well.

## Environment Information

```
     active environment : base
            shell level : 1
 populated config files : /root/.condarc
          conda version : 4.5.12
         python version : 3.7.1.final.0
       base environment : /opt/conda  (writable)
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/pro/linux-64
                          https://repo.anaconda.com/pkgs/pro/noarch
                          https://conda.anaconda.org/t/<TOKEN>/org/linux-64
                          https://conda.anaconda.org/t/<TOKEN>/org/noarch
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/4.5.12 requests/2.21.0 CPython/3.7.1 Linux/4.4.0-93-generic debian/9 glibc/2.24
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

kalefranz:
fix #8157

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ smoketest:
 	$(PYTEST) tests/test_create.py -k test_create_install_update_remove
 
 
-unit: clean
+unit:
 	$(PYTEST) $(ADD_COV) -m "not integration and not installed"
 
 
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from logging import getLogger
 from os.path import basename, dirname
 import re
 import sys
@@ -161,7 +162,7 @@ def disp_features(features):
 
 @swallow_broken_pipe
 def stdout_json(d):
-    print(json_dump(d))
+    getLogger("conda.stdout").info(json_dump(d))
 
 
 def stdout_json_success(success=True, **kwargs):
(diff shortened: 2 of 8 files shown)


msarahan:
Looks like removing memoization has broken one of the tests by showing some output more than once.  Was that intentional?  Show things be shown more than once?

msarahan:
Skipping this for 4.6.2.  We should address it soon, though, because leaking these things in CI logs is pretty bad.

kalefranz:
The two test failures now are because `conda config --get channels` now has the obfuscated token.  I think this is ok, but it's technically an impactful behavior change.  The original design intent of the `conda config --get` command was so that you could do something like

    conda config --get | xargs -n1 conda config

and basically recover the initial config.  At least that's my understanding.  After several years interacting with this issue tracker though, I don't think that pattern is widely used or has much real-world application.

PR #8170, opened 2019-01-30
DO NOT MERGE: small tweaks to make solve environment faster

brentp:
For discussion only.

I know there is work on this elsewhere, but in my profiling, it seems much of the time spent solving environment is outside of the pure SAT stuff. I made a few small changes and get a 10-15% improvement. Perhaps someone more familiar with the code can get more with other small changes.

I added the profiling code, then ran `conda create --dry-run -n conda_forge_r -c conda-forge r-essentials
` 3 times without any other changes. The fastest time reported in profile.txt was 82 seconds. Then made the other changes and the fastest of 3 was 71 seconds. Since this is not far from the improvement reported [here](https://www.anaconda.com/blog/developer-blog/conda-4-6-release/) by using a different solver, I thought it might be worth reporting.

I think most of the improvement comes from putting `hash` inside of the `_hash_key` memoized property.

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -168,6 +168,7 @@ class MatchSpec(object):
     FIELD_NAMES_SET = frozenset(FIELD_NAMES)
     _MATCHER_CACHE = {}
 
+
     def __init__(self, optional=False, target=None, **kwargs):
         self._optional = optional
         self._target = target
@@ -238,7 +239,8 @@ class MatchSpec(object):
         return True
 
     def _match_individual(self, record, field_name, match_component):
-        val = getattr(record, field_name)
+        #val = getattr(record, field_name)
+        val = record.__dict__[field_name]
         try:
             return match_component.match(val)
         except AttributeError:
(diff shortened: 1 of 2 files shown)


conda/models/match_spec.py
```diff
@@ -238,7 +239,8 @@ def match(self, rec):
         return True
 
     def _match_individual(self, record, field_name, match_component):
-        val = getattr(record, field_name)
+        #val = getattr(record, field_name)
+        val = record.__dict__[field_name]
```
kalefranz: Actually this one probably won’t work. All of the record fields are ultimately descriptors, and there can sometimes be a lot of dynamic translation depending on the field.

brentp: yeah. that's the one that is breaking the CI.

brentp:
ok, thanks for having a look. Just wanted to get this up in case it was of use. I also noted that, even though I was in a py3 env, it appeared to be looking at lots and lots of py2 packages.

travishathaway:

PR #8212, opened 2019-02-03
handle missing rename script more gracefully

msarahan:
We've seen some issues where rename_tmp.bat can disappear in the middle of an install, which totally trashes that install.  Hopefully this PR downgrades those errors.  Things may still be in a non-ideal state, but at least this will allow the process to get further and maybe closer to a working state.


jjhelmus:
LGTM

PR #8225, opened 2019-02-05
``Source activate`` to be ``conda activate`` [skip ci]

rrigdon:
…` [skip ci]

Per conda 4.6, updated references to reflect that ``conda activate`` is the new command


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": "${workspaceFolder}/docs/source"
+}
```
kalefranz: Was the inclusion of this file intentional? My guess is `.vscode/` should be added to `.gitignore`.

PR #8229, opened 2019-02-05
activate env when running post-link/unlink scripts

### LINKED ISSUE #7789: Envs need to be activated before post-link scripts run
In Python 3.7, we removed a patch to Python that put Library/bin on PATH.  This was done because putting that on PATH broke the new conda activate stuff somehow.  A side effect of this change is that packages with post-link scripts may fail to run (see https://github.com/ContinuumIO/anaconda-issues/issues/10082 for one example).  At a minimum, we should be activating environments before we run post-link scripts.  We should think about pre-unlink and pre-link, too, I guess.

jjhelmus:
Run post-link and post-unlink scripts in an activated environment. This is
implemented via an temporary script which activates the environment and then
calls the original script.

closes #7789

### DIFF
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -1002,9 +1002,9 @@ def run_script(prefix, prec, action='post-link', env_prefix=None, activate=False
             return False
         if activate:
             conda_bat = env.get("CONDA_BAT", abspath(join(context.root_prefix, 'bin', 'conda')))
-            with tempfile.NamedTemporaryFile(mode='w', prefix=prefix, delete=False) as fh:
-                fh.write('@CALL {0} activate \"{1}\""\n'.format(conda_bat, prefix))
-                fh.write('@CALL "\{0}\"\n'.format(path))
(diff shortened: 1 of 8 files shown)


jjhelmus:
I have a test package that fails to install when not run in an activated environment, just want to see if this passes CI before adding it.

jjhelmus:
Is there a good location to stash the recipe for the `_conda_test_env_activated_when_post_link_executed`. ?

jjhelmus:
Recipe added to `tests/data/_conda_test_env_activated_when_post_link_executed_recipe` directory.

msarahan:
Timeout on appveyor.  Seems to me like a slowdown is expected from this change.  We'll see if we can make it up some other way.  This is the most correct way to do things.  Merging.

kalefranz:
I made the above comments on my phone.  Guess I missed a "submit review" button.  Sorry for getting this to you guys late.

Issue #8238, opened 2019-02-07
conda search --env --info not implemented


spencermathews:
## Current Behavior

The `--envs` option to `conda search` is a nice touch, but it breaks the functionality of the `--info` option. Namely, when used together the detailed information expected from `--info` is not displayed, and the output is identical to using `--envs` alone.

### Steps to Reproduce

```
conda search --envs --info  # output is identical to conda search --envs, i.e. no details
```

## Expected Behavior

Since `conda --info` displays detailed package information (and without a matchspec argument it appears to do so for all packages within my active channels), and since `conda --envs` displays basic information for all packages in all of my environments, it seems reasonable to expect that `conda --envs --info` would then display detailed information about all packages in all of my environments.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : /home/xxxx/anaconda3
            shell level : 1
       user config file : /home/xxxx/.condarc
 populated config files : /home/xxxx/.condarc
          conda version : 4.6.2
    conda-build version : not installed
         python version : 3.7.2.final.0
       base environment : /home/xxxx/anaconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /home/xxxx/anaconda3/pkgs
                          /home/xxxx/.conda/pkgs
       envs directories : /home/xxxx/anaconda3/envs
                          /home/xxxx/.conda/envs
               platform : linux-64
             user-agent : conda/4.6.2 requests/2.21.0 CPython/3.7.2 Linux/4.10.0-38-generic linuxmint/18.3 glibc/2.23
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```
==> /home/xxxx/.condarc <==
ssl_verify: True
channels:
  - defaults
  - conda-forge

```
</p></details>

PR #8240, opened 2019-02-07
Print info for searching envs

soapy1:
Fixes https://github.com/conda/conda/issues/8238


jjhelmus:
I believe the AppVeyor failure is due to a change in behavior of ruamel_yaml.  I moved the 0.15.87 packages to a broken label in the conda-canary channel.

PR #8246, opened 2019-02-09
Solve problem in xonsh activation script on Windows.

melund:
This fixes problem with hook preamble inserted into the xonsh activation script on Windows.

Since paths contain backslashes on windows we need to make the string raw. 

This following will not work due to the `\n` in the path
```
CONDA_EXE = "C:\Users\nemo\anaconda3\Scripts\conda.exe"
```
But this is better:
```
CONDA_EXE = r"C:\Users\nemo\anaconda3\Scripts\conda.exe"
```

The PR also removes the `native_path_to_unix`. Xonsh handles both path with both forward and backslashes, but it doesn't handle the CygWin versions of the Windows paths.


msarahan:
It looks like the tests need to be updated for these raw strings.  I don't see anything wrong otherwise - just that the "correct" value specified by the tests needs to be updated to be a raw string.

You could also maybe just always code forward slashes, then use os.path.normpath to go back to Windows backslashes when necessary? https://pythonconquerstheuniverse.wordpress.com/2008/06/04/gotcha-%E2%80%94-backslashes-in-windows-filenames/

I'm ambivalent, but either way, please look into the failing tests.

melund:
Yes. We really only want to use raw strings (or escape backslashes) for things that are paths. So maybe we could replace `path_conversion` with a function that converts backslashes to forward slashes on Win.

Not sure if `path_conversion` is called an every thing that is a path, but I will just try.

scopatz:
@melund - I have fixed the merge conflicts here.  Can you please give me access to push to your fork? Thanks!

melund:
@scopatz I have given you push rights.. I think ...

scopatz:
OK! I have pushed up the changes (thanks!) but it seems like the tests fail.

scopatz:
Yeah, it does seem like it

scopatz:
Do you want to take another stab at fixing this up @melund?

conda/activate.py
```diff
@@ -943,7 +954,7 @@ def __init__(self, arguments=None):
         super(XonshActivator, self).__init__(arguments)
 
     def _hook_preamble(self):
-        return '$CONDA_EXE = "%s"' % context.conda_exe
+        return '$CONDA_EXE = "%s"' % self.path_conversion(context.conda_exe)
```
melund: Ahh.  `self.path_conversion` goes to the parent class since `XonshActivator.path_conversion()` is a static method.

melund: Or maybe not. More likely just because the ``__init__`` overrides the ``path_conversion`` name.

melund:
@scopatz. I think it is fixed now but the test seems to be failing in some completely unrelated places. I better try to pull from master and see if that fixes it.

melund:
@soapy1 I think this is ready to be merged. The test failures are now unrelated. 

It is a followup to @scopatz's #9319 fix the other day.

melund:
@scopatz Could you help nodge anyone to get this reviewed?

msarahan:
Thanks @melund and @scopatz

PR #8249, opened 2019-02-11
remove experimental featureless_minimization_disabled_feature_flag

kalefranz:
The behavior enabled by this experimental feature flag didn't pan out as hoped, and I think it's now safe to remove from the code base.

xref: #7601


kalefranz:
We should also close https://github.com/conda/conda/issues/7626 with this PR merged.

PR #8251, opened 2019-02-11
clean up dead code

kalefranz:


conda/activate.py
```diff
@@ -98,11 +98,6 @@ def reactivate(self):
     def hook(self, auto_activate_base=None):
         builder = []
         builder.append(self._hook_preamble())
-
-        # here
-        condabin_dir = self.path_conversion(join(context.conda_prefix, "condabin"))
-        self.export_var_tmpl % ("PATH", self.pathsep_join((condabin_dir, )))
```
kalefranz: This apparently snuck in through #7773, and should have been removed before the PR was merged.  It's dead code and has no effect.

msarahan: Could you please elaborate on how it's dead code?  Removing things that appear to be setting PATH is something I think we need to be careful with.

kalefranz: There's no assignment except `condabin_dir`, and none of the methods have side effects. To not be dead code we'd actually need to be doing something with the result of line 104.

msarahan: got it.  Thanks.

Issue #8187, opened 2019-01-30
BUG: conda lib paths for NumPy, vtk, possibly others on Windows with environment.yml that has pip installs


larsoner:
## Current Behavior

As of about 10 days ago, I started hitting a DLL not found error on Windows 10 on AppVeyor (not OSX or Linux on Travis) when using an `environment.yml` file to install NumPy using conda and another package using `pip`, using the [widely used ci-helpers repo](https://github.com/astropy/ci-helpers). This code had been working reliably for a year or two before this. I am mystified because I cannot find changes to `conda`, the packages, or AppVeyor images that could have caused this.

### Steps to Reproduce

I've tried to boil this down to a MWE locally in a Windows 10 VM. In a Windows Git bash terminal (because I don't know how to use `cmd`/`PowerShell` properly):
```Console
$ curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe
$ ./Miniconda3-latest-Windows-x86_64.exe  # just accept all defaults
$ export PATH=/c/Users/tester/Miniconda3/:/c/Users/tester/Miniconda3/Scripts/:$PATH
$ conda env create -n test -f environment.yml
```
<details>
  <summary>Output</summary>

    $ conda env create -n test -f environment.yml
    Solving environment: done
    
    
    ==> WARNING: A newer version of conda exists. <==
      current version: 4.5.12
      latest version: 4.6.1
    
    Please update conda by running
    
        $ conda update -n base -c defaults conda
    
    
    
    Downloading and Extracting Packages
    numpy-1.15.4         | 47 KB     | ############################################################################ | 100%
    mkl_fft-1.0.10       | 135 KB    | ############################################################################ | 100%
    python-3.7.2         | 17.7 MB   | ############################################################################ | 100%
    blas-1.0             | 6 KB      | ############################################################################ | 100%
    numpy-base-1.15.4    | 3.9 MB    | ############################################################################ | 100%
    ca-certificates-2018 | 153 KB    | ############################################################################ | 100%
    mkl_random-1.0.2     | 328 KB    | ############################################################################ | 100%
    icc_rt-2019.0.0      | 9.4 MB    | ############################################################################ | 100%
    mkl-2019.1           | 158.3 MB  | ############################################################################ | 100%
    intel-openmp-2019.1  | 1.7 MB    | ############################################################################ | 100%
    Preparing transaction: done
    Verifying transaction: done
    Executing transaction: done
[...]

PR #8263, opened 2019-02-12
Prepend pip commands with activating env

soapy1:
addresses https://github.com/conda/conda/issues/8187

### DIFF
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -9,7 +9,6 @@ import os
 from os.path import abspath, basename, dirname, isdir, join
 from subprocess import CalledProcessError
 import sys
-import tempfile
 from traceback import format_exception_only
 import warnings
 
@@ -42,7 +41,7 @@ from ..gateways.subprocess import subprocess_call
 from ..models.enums import LinkType
 from ..models.version import VersionOrder
 from ..resolve import MatchSpec
-from ..utils import human_bytes
+from ..utils import human_bytes, wrap_subprocess_call
 
 log = getLogger(__name__)
 
@@ -994,7 +993,6 @@ def run_script(prefix, prec, action='post-link', env_prefix=None, activate=False
             environments.  Future versions of conda may deprecate and ignore pre-link scripts.
             """) % prec.dist_str())
 
-    tmp_prefix = abspath(join(prefix, '.tmp'))
     script_caller = None
     if on_win:
         try:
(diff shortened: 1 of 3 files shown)


msarahan:
Looks good to me.  Once the tests are in order, I think this can go in.  @jjhelmus please take a look, since you wrote the original functionality here.

jjhelmus:
cf: #8229

conda_env/installers/pip.py
```diff
@@ -56,6 +64,8 @@ def _pip_install_via_requirements(prefix, specs, args, *_, **kwargs):
         # So we delete the temporary file in a finally block.
         if requirements is not None and op.isfile(requirements.name):
             os.remove(requirements.name)
+        if script_caller is not None:
+            os.remove(script_caller)
```
jjhelmus: As @kalefranz mentioned in #8229 this should be `rm_rf` not `os.remove` unless I am missing something.  If you want to change the call in line ~1065 of `link.py` that would be great.

conda/core/link.py
```diff
@@ -1003,28 +1001,21 @@ def run_script(prefix, prec, action='post-link', env_prefix=None, activate=False
...
         shell_path = 'sh' if 'bsd' in sys.platform else 'bash'
         if activate:
-            conda_exe = env.get("CONDA_EXE", abspath(join(context.root_prefix, 'bin', 'conda')))
-            with tempfile.NamedTemporaryFile(mode='w', prefix=tmp_prefix, delete=False) as fh:
-                fh.write("eval \"$(\"{0}\" \"shell.posix\" \"hook\")\"\n".format(conda_exe)),
-                fh.write("conda activate \"{0}\"\n".format(prefix)),
-                fh.write("source \"{}\"\n".format(path))
-                script_caller = fh.name
-            command_args = [shell_path, "-x", script_caller]
+            command = ". \"{}\"\n".format(path)
```
jjhelmus: Calling path directly and not souring is correct here.  Thanks you for correcting this mistake from #8229.

conda/utils.py
```diff
@@ -254,3 +257,30 @@ def sys_prefix_unfollowed():
...
         return sys.prefix
     return unfollowed
+
+
+def wrap_subprocess_call(on_win, root_prefix, prefix, command):
+    env = environ.copy()
+    tmp_prefix = abspath(join(prefix, '.tmp'))
+    script_caller = None
+    if on_win:
+        comspec = environ[str('COMSPEC')]
```
jjhelmus: This environment lookup can fail but perhaps allowing the exception to propagate is the best option here.

soapy1:
Looks like the python 2.7 tests are hitting the 1hr limit on appveyor

PR #8265, opened 2019-02-12
Conda pkg v2

msarahan:
Sister PR to https://github.com/conda/conda-build/pull/3334

The general thought process with the new format is to be as non-disruptive as possible.  Conda can still use the old package format, but it will prefer the new one if it sees that those packages are available.  Availability is decided by repodata contents: if a repodata entry has a ``conda_size`` entry, conda will try to download and use the ``.conda`` format file from the repo.  It is assumed that ``.conda`` and ``.tar.bz2`` files live alongside each other in a single folder, and it is also currently assumed that there are not ``.conda`` files that do not have a companion ``.tar.bz2``.

This PR also swaps out md5 for download and file verification to use sha256.  The metadata for packages on the main channel will have this, but on the ``free`` and ``anaconda.org`` channels, this data doesn't exist.  I'll have to walk that back a bit - try sha256 first, but fallback to md5.

* [x] Provide fallback path for repodata without sha256 entries
* [x] Write more tests for behavior with cached packages.  Although conda should prefer the new format for new downloads, it should be able to use any existing .tar.bz2's in the package cache.  We need to test this.


conda/models/records.py
```diff
@@ -245,6 +245,9 @@ class PackageRecord(DictSafeMixin, Entity):
 
     md5 = StringField(default=None, required=False, nullable=True, default_in_dump=False)
     url = StringField(default=None, required=False, nullable=True, default_in_dump=False)
+    sha256 = StringField(default=None, required=False, nullable=True, default_in_dump=False)
+    conda_outer_sha256 = StringField(default=None, required=False, nullable=True,
```
soapy1: why does this not get a `conda_inner_sha256` as well?

conda/models/records.py
```diff
@@ -390,19 +414,48 @@ def is_extracted(self):
...
-            return memoized_md5
+        memo_key = '_memoized_md5'
+        memoized_hash = getattr(self, memo_key, None)
+        if memoized_hash:
+            return memoized_hash
 
         from os.path import isfile
         if isfile(self.package_tarball_full_path):
             from ..gateways.disk.read import compute_md5sum
-            md5sum = compute_md5sum(self.package_tarball_full_path)
```
soapy1: I don't think the `from os.path import isfile` above is needed anymore since it's imported at the beginning

dev/requirements.txt
```diff
@@ -1,7 +1,6 @@
-python=3.6
 pycosat >=0.6.3
 pyopenssl >=16.2.0
 requests >=2.12.4,<3
 ruamel_yaml >=0.11.14
 git
-
+conda-package-handling
```
soapy1: I think this requirement also needs to be added to `utils/requirements.txt` so that docs stuff (read the docs and the gh-pages tool) don't break

msarahan:
There is a fix for the library load problem in https://github.com/conda/conda/pull/8550/commits/42a2df8722d0c7d1152fd469eb80f1dc87477027

conda/models/records.py
msarahan: Seems like a mistake.  Will add it in.

dev/requirements.txt
msarahan: we currently can't.  As we discussed in person, conda-package-handling requires libarchive, and I haven't been able to get that working with the pypi-provided libarchive.  I think there's also a problem with our python-libarchive-c package such that pip doesn't see that it is providing the libarchive package.

Overall, yes, I agree with you, but unfortunately, I'm not sure we can. =(

dhirschfeld:
Does this mean 4.7.0 will support the new conda package format?

Curious, as I'm excited by the possibility of large speedups and am keen to try it out!

msarahan:
Yep. Keep in mind any speedups are only if the package isn't already in the
package cache. It will speed up first time use of a package, but won't have
any impact after that.

dhirschfeld:
A lot of the time (most?) I'm updating my existing environment rather than creating a new environment with existing packages. By definition, that's downloading new/updated packages so I'm hoping this will help with that fairly common (for me) workflow.

PR #8266, opened 2019-02-13
skip scanning folders for contents during reversal of transactions

msarahan:
Just ignore folders for these reversals.  It's a bit messy, but it maintains the fix and it'll be faster.

### DIFF
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -32,7 +32,7 @@ from ..gateways.disk.create import (compile_multiple_pyc, copy,
                                     create_hard_link_or_copy, create_link,
                                     create_python_entry_point, extract_tarball,
                                     make_menu, mkdir_p, write_as_json_to_file)
-from ..gateways.disk.delete import rm_rf, path_is_clean
+from ..gateways.disk.delete import rm_rf
 from ..gateways.disk.permissions import make_writable
 from ..gateways.disk.read import (compute_md5sum, compute_sha256sum, islink, lexists,
                                   read_index_json)
@@ -370,7 +370,7 @@ class LinkPathAction(CreateInPrefixPathAction):
     def reverse(self):
         if self._execute_successful:
             log.trace("reversing link creation %s", self.target_prefix)
-            if not isdir(self.target_full_path) or path_is_clean(self.target_full_path):
+            if not isdir(self.target_full_path):
                 rm_rf(self.target_full_path, clean_empty_parents=True)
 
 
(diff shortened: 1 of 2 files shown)


msarahan:
I hope this change reduces the win py27 test time.  os.walk was massively sped up in python 3.5 (https://www.python.org/dev/peps/pep-0471/), so this is probably only really significant for py27, and especially so for windows.

msarahan:
Nope, doesn't appear to have changed anything with py27.  That's too bad.  I guess it's time to do some profiling to understand why that particular configuration is so much slower.

msarahan:
A few people have complained about dramatically slower runs.  Let's give this a try, and also spend some more time profiling.

kalefranz:
I haven’t looked at the code too closely, but is `rm_rf()` subprocessing now? Subprocessing is horribly slow on Windows. Seems like that might be the root cause.

msarahan:
It tries to, yes.  It's a trade-off between the shell's rd command being much faster than python and the startup cost of subprocessing.  I need to run some benchmarks.  It will probably depend on the number of things being cleaned up.

msarahan:
some interesting benchmarking ideas here: https://bugs.python.org/issue32453

msarahan:
| removal command  | 1k files | 10k files | 100k files | 400k files |
| ------------- | ------------- | ------------- | ------------- |------------- |
| rm -rf (git4win bash)  | 0.259s | 2.475s | 25.396s  | 100.085s  |
| shutil rmtree  | 0.314s | 2.730s  | 23.836s  | 94.399s  |
| subprocess rd /s /q  | 0.251s | 1.758s  | 15.655  | 63.855s  |

Seems like it's not the bottleneck, and definitely shines for large numbers of files.

msarahan:
For anyone wanting to run the benchmark for themselves:

```
for SIZE in 1000 10000 100000
do
    echo "rm -rf"
    time (mkdir dirtest && cd dirtest && seq 1 $SIZE | xargs touch)
    time rm -rf dirtest/
    echo "python shutil rmtree"
    time (mkdir dirtest && cd dirtest && seq 1 $SIZE | xargs touch)
    time python -c 'import shutil; shutil.rmtree("dirtest")'
    echo "python subprocess rd /s /q"
    time (mkdir dirtest && cd dirtest && seq 1 $SIZE | xargs touch)
    time python -c 'import subprocess; subprocess.check_call("rd /s /q dirtest", shell=True)'
done
```

PR #8267, opened 2019-02-13
Add virtual package to solver based on detected CUDA version.

seibert:
Loosely based on #7258, this PR specifically adds support for a virtual package to represent the version of CUDA (if any) supported by the graphics driver.

There are several issues to work out:

* I've added a new PackageType called `VIRTUAL_SYSTEM`. Presumably other system attributes (like glibc, and AVX) exposed as virtual packages in the future will also be of this type.
* I've called the package `cuda`, but I know there was interest in prefixing virtual packages with a special character, making the preferred name something like `+cuda` or `@cuda`.
* To simplify testing (since changing graphics drivers is awkward and can't be virtualized with Docker), the detected CUDA version can be overridden with the environment variable `CONDA_CUDA_OVERRIDE`.  This name probably needs to be something else.
* Virtual packages are reported in `conda info`.  Not sure if the layout is ideal.
* For some reason, I needed to inject the `cuda` virtual package in two different branches of `_prepare`, when it seems like only one should be required.
* This code needs some unit tests.


seibert:
FYI: I've verified the basic operation of the PR manual by building a local variant of the cudatoolkit package that depends on `cuda ==10.0` and using the `CONDA_CUDA_OVERRIDE` environment variable to dial the CUDA version up and down.  (The real constraint should be `cuda >=10.0`, though.)

conda/common/cuda.py
```diff
@@ -0,0 +1,70 @@
...
+        lib_filenames = [
+            'libcuda.dylib', # check library path first
+            '/usr/local/cuda/lib/libcuda.dylib'
+        ]
+    elif system == 'Linux':
+        lib_filenames = [
+            'libcuda.so', # check library path first
+            '/usr/lib64/nvidia/libcuda.so', # Redhat/CentOS/Fedora
+            '/usr/lib/x86_64-linux-gnu/libcuda.so', # Ubuntu
+        ]
```
kkraus14: This is the driver libcuda, and not the cuda runtime. This will return the latest version of CUDA supported as opposed to the version in use.

conda/common/cuda.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Get CUDA version
+    try:
+        cuInit = libcuda.cuInit
+        flags = ctypes.c_uint(0)
+        ret = cuInit(flags)
+        if ret != 0:
+            return None
+    
+        cuDriverGetVersion = libcuda.cuDriverGetVersion
```
kkraus14: Ditto my above comment.

kkraus14:
@seibert this looks like you're detecting the version of cuda supported as opposed to the version of cuda installed. Is that the intention? Given nvcc is tied to the CUDA runtime version I imagine we'd want to detect the runtime version for version pinning of packages.

jjhelmus:
I believe the intention of this `cuda` virtual package is to provide the solver with the version(s) of CUDA that are supported by the driver.  The solver would make use of this information when installing the `cudatoolkit` package and select a compatible version.

kkraus14:
> I believe the intention of this `cuda` virtual package is to provide the solver with the version(s) of CUDA that are supported by the driver. The solver would make use of this information when installing the `cudatoolkit` package and select a compatible version.

Understood, I apologize for jumping the gun 😅.

seibert:
If there is a better way to do the solver unit test, I would appreciate knowing what it is.  I had to add some fake cudatoolkit packages to one of the test indices, and then duplicate some code to avoid having the index cached.

msarahan:
This looks great to me, @seibert.  I think we'll wait to merge it until after the 4.6.12 goes out, but I don't think anything else needs to be done on your side.  I can do the troubleshooting on the os.environ stuff that appears to be tripping things up.

Docs and everything!  Thanks!

seibert:
This error in the Python 2.7 build makes no sense to me: https://circleci.com/gh/conda/conda/16545#tests/containers/0
```
      File "/home/circleci/conda/conda/exceptions.py", line 1003, in __call__
        return func(*args, **kwargs)
      File "/home/circleci/conda/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/home/circleci/conda/conda/cli/conda_argparse.py", line 82, in do_call
      File "/home/circleci/conda/conda/cli/main_info.py", line 314, in execute
        info_dict = get_info_dict(args.system)
      File "/home/circleci/conda/conda/cli/main_info.py", line 134, in get_info_dict
        _supplement_index_with_system(virtual_pkg_index)
      File "/home/circleci/conda/conda/core/index.py", line 139, in _supplement_index_with_system
        cuda_version = context.cuda_version
      File "/home/circleci/conda/conda/_vendor/auxlib/decorators.py", line 268, in new_fget
        cache[inner_attname] = func(self)
      File "/home/circleci/conda/conda/common/decorators.py", line 12, in wrapper
        value = os.environ.get(envvar_name, None)
    AttributeError: 'module' object has no attribute 'environ'

`$ /opt/conda/bin/conda info`
```
It is clearly triggered by the new code, but I don't understand how `os.environ` could not exist....

seibert:
Same problem with Python 2.7 on Windows.  Python 3 on windows also fails, but I don't understand what the failure is from the console output...

msarahan:
Thanks, @seibert.  We can take it from here if you need to get on to other things.

nehaljwani:
The python 2.7 problem is a relative import which `import os` in `decorators.py` is picking up from:
```
/path/to/conda/conda/common/os/__init__.py
```

msarahan:
ouch.  That hurts my head.  Thanks for figuring that out, @nehaljwani

msarahan:
@seibert I have rebased your work on the current master branch.  The flake8 errors are OK to ignore.  I have already fixed them in #8265.  We're waiting to merge this until we have finished up work on miniconda 4.6.x.  We will then branch off 4.6.x and merge your PR as part of what will be 4.7.0.

Thanks again for contributing this.  It will be a very useful improvement.

jakirkham:
> I believe the intention of this `cuda` virtual package is to provide the solver with the version(s) of CUDA that are supported by the driver. The solver would make use of this information when installing the `cudatoolkit` package and select a compatible version.

Have raised issue ( https://github.com/conda/conda/issues/9115 ) about adding the `cuda` virtual package as a requirement of the `cudatoolkit` packages.

Issue #8271, opened 2019-02-13
Add #!override flag to condarc


dhirschfeld:
This would allow a user to e.g. not include channels defined in a system environment or an env to not inherit user channels.

e.g.
```
# file: ~/.condarc
channels: 
  - three 
  - four

# file: $CONDA_PREFIX/.condarc
channels:  #!override
  - five 
  - six
```
...would result in:
```
channels:
  - five
  - six
```
rather than
```
channels:
  - five
  - six
  - three
  - four
```
...as would have been the case without the `#!override`

dhirschfeld:
I don't think this has any security implementation as it only allows a user to *remove* keys, not add them.

PR #8272, opened 2019-02-13
Add C:/ProgramData/conda as a search path on Windows

### LINKED ISSUE #8270: condarc file precedence not respected when merging
In my system `.condarc` I have `defaults` and `conda-forge` and in my user `.condarc` I have my own `dhirschfeld` channel:

image

My understanding ([from this blog post](https://www.anaconda.com/conda-configuration-engine-power-users/)) was that the system `.condarc` should take precedence when merging and that my channel should be *appended* to the channel list. This is the behaviour I'd like, but not what I'm observing in practice:
```
(base) C:\> conda info

     active environment : base
    active env location : C:\Miniconda3
            shell level : 1
       user config file : C:\Users\dhirschfeld\.condarc
 populated config files : C:\Miniconda3\.condarc
                          C:\Users\dhirschfeld\.condarc
          conda version : 4.6.3
         python version : 3.7.1.final.0
       base environment : C:\Miniconda3  (writable)
           channel URLs : https://conda.anaconda.org/dhirschfeld/win-64
                          https://conda.anaconda.org/dhirschfeld/noarch
                          https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
                          https://conda.anaconda.org/conda-forge/win-64
          package cache : C:\Miniconda3\pkgs
                          C:\Users\dhirschfeld\.conda\pkgs
                          C:\Users\dhirschfeld\AppData\Local\conda\conda\pkgs
       envs directories : C:\Miniconda3\envs
                          C:\Users\dhirschfeld\.conda\envs
                          C:\Users\dhirschfeld\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.6.3 requests/2.21.0 CPython/3.7.1 Windows/10 Windows/10.0.17763
          administrator : False
             netrc file : None
           offline mode : False
```

dhirschfeld:
Closes #8270

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -24,25 +24,44 @@ machine_bits = 8 * tuple.__itemsize__
 
 APP_NAME = 'conda'
 
-SEARCH_PATH = (
-    '/etc/conda/.condarc',
-    '/etc/conda/condarc',
-    '/etc/conda/condarc.d/',
-    '/var/lib/conda/.condarc',
-    '/var/lib/conda/condarc',
-    '/var/lib/conda/condarc.d/',
-    '$CONDA_ROOT/.condarc',
-    '$CONDA_ROOT/condarc',
-    '$CONDA_ROOT/condarc.d/',
-    '~/.conda/.condarc',
-    '~/.conda/condarc',
-    '~/.conda/condarc.d/',
-    '~/.condarc',
-    '$CONDA_PREFIX/.condarc',
-    '$CONDA_PREFIX/condarc',
-    '$CONDA_PREFIX/condarc.d/',
-    '$CONDARC',
-)
+if on_win:
+    SEARCH_PATH = (
+        'C:/ProgramData/conda/.condarc',
+        'C:/ProgramData/conda/condarc',
+        'C:/ProgramData/conda/condarc.d',
+        '$CONDA_ROOT/.condarc',
+        '$CONDA_ROOT/condarc',
+        '$CONDA_ROOT/condarc.d/',
+        '~/.conda/.condarc',
+        '~/.conda/condarc',
+        '~/.conda/condarc.d/',
+        '~/.condarc',
+        '$CONDA_PREFIX/.condarc',
+        '$CONDA_PREFIX/condarc',
+        '$CONDA_PREFIX/condarc.d/',
+        '$CONDARC',
+    )
+else:
+    SEARCH_PATH = (
+        '/etc/conda/.condarc',
+        '/etc/conda/condarc',
+        '/etc/conda/condarc.d/',
+        '/var/lib/conda/.condarc',
+        '/var/lib/conda/condarc',
+        '/var/lib/conda/condarc.d/',
+        '$CONDA_ROOT/.condarc',
+        '$CONDA_ROOT/condarc',
+        '$CONDA_ROOT/condarc.d/',
+        '~/.conda/.condarc',
+        '~/.conda/condarc',
+        '~/.conda/condarc.d/',
+        '~/.condarc',
+        '$CONDA_PREFIX/.condarc',
+        '$CONDA_PREFIX/condarc',
+        '$CONDA_PREFIX/condarc.d/',
+        '$CONDARC',
+    )
+
 
 DEFAULT_CHANNEL_ALIAS = 'https://conda.anaconda.org'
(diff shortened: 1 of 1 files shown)


conda/base/constants.py
```diff
@@ -42,7 +43,9 @@
     '$CONDA_PREFIX/condarc',
     '$CONDA_PREFIX/condarc.d/',
     '$CONDARC',
-)
+]
+if on_win:
+    SEARCH_PATH += ['C:/ProgramData/conda/.condarc']
```
kalefranz: Probably best to keep it immutable and list the paths explicitly got each platform. Follow the pattern set by `DEFAULT_CHANNELS` below.

kalefranz: I guess `COMPATIBLE_SHELLS` uses a different pattern. Would still prefer to keep the values immutable.

dhirschfeld:
Had some issues so I just force pushed. Fortunately this change is so simple that it shouldn't create any confusion.

conda/base/constants.py
```diff
@@ -24,25 +24,44 @@
...
+        '~/.conda/condarc',
+        '~/.conda/condarc.d/',
+        '~/.condarc',
+        '$CONDA_PREFIX/.condarc',
+        '$CONDA_PREFIX/condarc',
+        '$CONDA_PREFIX/condarc.d/',
+        '$CONDARC',
+        'C:/ProgramData/conda/.condarc',
+        'C:/ProgramData/conda/condarc',
+        'C:/ProgramData/conda/condarc.d'
```
dhirschfeld: So, to get my system config (`C:/ProgramData/conda`) to take precedence I had to have it at the *bottom* of the `SEARCH_PATH` list and I'm not sure if that's the correct thing to do.

Reading the blog post again, it actually shows the precedence being `ENV --> USER --> SYSTEM`
```
# file: /etc/conda/condarc
channels: 
  - one 
  - two

# file: ~/.condarc
channels: 
  - three 
  - four

# file: $CONDA_PREFIX/.condarc
channels: 
  - five 
  - six
```
results in:
```
channels: 
  - five 
  - six 
  - three 
  - four 
  - one 
  - two
```
i.e. the system config should be the least significant rather than the most?

That being the case it seems `C:/ProgramData/conda` should be at the top of the list and that user-specified channels will take precedence to those specified at the system level.

Is that correct? Is that the desired default behaviour?

dhirschfeld: As a sysadmin I would like user channels to be *appended* to the channel list - i.e for the channels I specify to take precedence in the config (but also able to be overridden by the user on the command line).

I'm fine changing the precedence to put `C:/ProgramData/conda` at the top / least significant position if that's the right thing to do - I think I can get my desired behaviour by using `#!top` in my system config

dhirschfeld:
I've changed the precedence of `ProgramData` so that it follows the design (which is the usual ordering for config files)

dhirschfeld:
This does bring up the question of how, as a system administrator I can change the inheritance so that the channels I specify have the highest priority. The trick of `#!top` works but seems a bit hacky.

Even though it is usual for more specific config to override system/global config the specification of channels is a bit different in that the system config isn't being overridden but is instead being extended.

When extending the channel list there are two ways to do it - either the channels are prepended or appended to those specified in the system config. As it currently stands the channels are prepended and hence take priority over those I specify at the system level. Given this behaviour my only option AFAICS is to use the `#!top` hack.

If instead the channels were appended to the system config I would have the behaviour I'm after and if the [`#!override`](https://github.com/conda/conda/issues/8271) flag were implemented the user could use that to reorder the channels if that was what they wanted. 

Changing the channels to append to the system config and implementing #8271 would make it possible for everyone to get whichever behaviour they wanted

msarahan:
Could you please rebase on master?  We've finally fixed the windows tests, and it'll be easier to reason about the test failures here.

dhirschfeld:
ping! Any chance this could be merged for the next point release? 

This is a pretty straight-forward improvement which doesn't change any behaviour - it simply replaces linux search paths which don't make sense on Windows with the equivalent Windows versions.

PR #8287, opened 2019-02-14
Set the necessary PATH entries in conda.bat

mingwandroid:
We cannot run python.exe (and expect it to work in general) without having all of the environment's path entries on PATH.

### DIFF
--- a/conda/shell/condabin/conda.bat
+++ b/conda/shell/condabin/conda.bat
@@ -5,7 +5,11 @@
 @IF [%1]==[activate]   "%~dp0_conda_activate" %*
 @IF [%1]==[deactivate] "%~dp0_conda_activate" %*
 
+@setlocal
+@for %%B in (%~dp0.) do @set PATH=%%~dpB;%%~dpBLibrary\mingw-w64\bin;%%~dpBLibrary\usr\bin;%%~dpBLibrary\bin;%%~dpBScripts;%%~dpBbin;%PATH%
 @CALL "%CONDA_EXE%" %*
+@endlocal
+
 @IF %errorlevel% NEQ 0 EXIT /B %errorlevel%
 
 @IF [%1]==[install]   "%~dp0_conda_activate" reactivate


mingwandroid:
@conda/conda-core please review this.

conda/shell/condabin/conda.bat
```diff
@@ -5,7 +5,11 @@
 @IF [%1]==[activate]   "%~dp0_conda_activate" %*
 @IF [%1]==[deactivate] "%~dp0_conda_activate" %*
 
+setlocal
+for %%B in (%~dp0.) do set PATH=%%~dpB;%%~dpBLibrary\mingw-w64\bin;%%~dpBLibrary\usr\bin;%%~dpBLibrary\bin;%%~dpBScripts;%%~dpBbin;%PATH%
```
msarahan: Probably need quoting?

```
for %%B in ("%~dp0.") do set "PATH=%%~dpB;%%~dpBLibrary\mingw-w64\bin;%%~dpBLibrary\usr\bin;%%~dpBLibrary\bin;%%~dpBScripts;%%~dpBbin;%PATH%"
```

mingwandroid: Nope.

mingwandroid: Neither need quoting here. BBIAB, going for a little walk.

msarahan:
This looks fine to me.  Do we also need a powershell equivalent?  What about the bash scripts that may get used on windows?

mingwandroid:
I guess: https://github.com/conda/conda/blob/master/conda/shell/condabin/Conda.psm1#L146

soapy1:
Tested for this bug against master using powershell and git bash and it looks like they are unaffected

msarahan:
Thanks for verifying that, @soapy1.  Merging.

conda/shell/condabin/conda.bat
czatom: It needs quoting, otherwise causing problems if the path contains spaces. The loop iterates over each substring separated by space.
Also, in conda_hook.bat there is something very similar:
@FOR %%F in ("%~dp0") do @SET "__condabin_dir=%%~dpF"

mingwandroid: Checking, thanks @czatom.

mingwandroid: Latest version is at  https://github.com/mingwandroid/conda/blob/ray_activation_fixes/conda/shell/condabin/conda.bat#L12-L17

While we are now quoting it prompting me to check the code found a pointless, broken for loop, so thanks again.

PR #8289, opened 2019-02-14
Use active code page for encoding activate script on Windows

### LINKED ISSUE #8241: activate.d scripts not run on windows if prefix contains unicode
activate.d scripts are not run on windows if the path contains unicode characters.  A test for this is to create environments with and without unicode characters with the `_test_set_env_in_activate_d` package in the `jjhelmus` channel.  When activated the `TEST_VAR` variable should be set to "1".  In environments with unicode characters running the activate.d script fails because of an invalid path in the activation script.

jjhelmus:
Allows activate.d scripts to be run on Windows if even if the prefix path contains unicode characters. 
closes #8241

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -72,9 +72,15 @@ class _Activator(object):
         if ext is None:
             return self.command_join.join(commands)
         elif ext:
+            #with open("C:\\Users\jhelmus\jjh_test.bat", 'w+b') as tf:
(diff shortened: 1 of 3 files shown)


msarahan:
looks like appveyor failures are web problems, not code problems.  Merging.

blackgear:
https://github.com/conda/conda/issues/8341

this code caused a Exception on Windows with Unicode support...

PR #8292, opened 2019-02-15
prepend root prefix paths to PATH before running conda commands on win

msarahan:
supersedes #8287 

This depends on a python patch that is in development.  This should not be merged until we can link it to confirmation that those python builds are available.


msarahan:
This looks like it breaks conda run.  I'm going to revert it for now.

PR #8300, opened 2019-02-18
tweak some logic in trash renaming to fix permission errors

msarahan:
The previous code here was incorrect.  It led to only the first file being renamed in-place to .conda_trash - all others were left untouched, which resulted in permissions errors on the next thing to overwrite them.  This was especially a problem with conda-build and subpackages, where a prefix is removed and then created with new deps.

### DIFF
--- a/conda/compat.py
+++ b/conda/compat.py
@@ -58,9 +58,6 @@ class TemporaryDirectory(object):
                     raise
                 _rm_rf(self.name)
             self._closed = True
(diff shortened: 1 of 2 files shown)


msarahan:
It appears that the py27 timeout on the advanced_pip test is because of interactive error prompts that are showing up.  The test is installing py37, but the core conda is py27.  Somehow that's interacting badly and the wrong MSVC runtimes are in play.

PR #8107, opened 2019-01-15
remove doc of --lock argument of conda clean

jjhelmus:
Remove the documentation of the --lock argument for conda clean whose
functionality was removed in #3862.

closes #8104

PR #8310, opened 2019-02-20
Removing --lock from conda clean (not supported)

gabrielcnr:
The `--lock` command line option for `conda build` has no behaviour associated to it anymore.

In fact, in the latest stable release of `conda`, invoking `conda clean --lock` gives you an `ArgumentError` error.

So, in this PR, I'm just removing the `--lock` from the `argparse` options.

```
$ conda clean --lock

ArgumentError: At least one removal target must be given. See 'conda clean --help'.
```

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -229,11 +229,6 @@ def configure_parser_clean(sub_parsers):
         help="Remove index cache.",
     )
     removal_target_options.add_argument(
-        "-l", "--lock",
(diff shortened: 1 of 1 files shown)


msarahan:
related to #8106, #8107 

Thanks for having a look at this.  We need to revisit this topic once we get the next release out.  We're having a lot of trouble on Windows with PATH related stuff, and we need to focus on that.  If we don't get to this soon, please don't hesitate to remind us with a comment here.

PR #8315, opened 2019-02-21
[skip ci] [wip] document mkl dll hell and new env vars

msarahan:
CC @mingwandroid @rrigdon - this needs more work and testing based on the final python builds we come up with in the next day or two.

### DIFF
--- a/docs/source/user-guide/troubleshooting.rst
+++ b/docs/source/user-guide/troubleshooting.rst
@@ -109,9 +109,47 @@ Cause
 You may observe this error cropping up after a conda update. More recent
 versions of conda and more recent builds of Python are more strict about
 requiring activation of environments. We're working on better error messages for
-them, but here's the story for now. Activation on Windows is a must. See more
-information on activation in :ref:`Activating environments <activate-env>`. When you don't activate your environment,
-conda can't find the libraries that it needs.
+them, but here's the story for now. Windows relies on the PATH environment
+variable as the way to locate libraries that are not in the immediate folder,
+and also not in the C:\Windows\System32 folder. Searching for libraries in the
+PATH folders goes from left to right. If you choose to put Anaconda's folders on
+PATH, there are several of them:
+
+  * (install root)
+  * (install root)/Library/mingw-w64/bin
+  * (install root)/Library/usr/bin
+  * (install root)/Library/bin
+  * (install root)/Scripts
+  * (install root)/bin
+  * (install root)/condabin
+
+Early installers for Anaconda put these on PATH. That was ultimately fragile
+because Anaconda isn't the only software on the system. If other software had
+similarly named executables or libraries, and came earlier on PATH, Anaconda
+could break. On the flip side, Anaconda could break other software if Anaconda
+were earlier in the PATH order and shadowed any other executables or libraries.
+To make this easier, we began recommending "activation" instead of modifying
+PATH. Activation is a tool where conda sets your PATH, and also runs any custom
+package scripts which are often used to set additional environment variables
+that are necessary for software to run (e.g. JAVA_HOME). Because activation runs
+only in a local terminal session (as opposed to the permanent PATH entry), it is
+safe to put Anaconda's PATH entries first. That means that Anaconda's libraries
+get higher priority when you're running Anaconda, but Anaconda doesn't interfere
+with other software when you're not running Anaconda.
+
+Anaconda's python interpreter included a patch for a long time that added the
+(install root)/Library/bin folder to that python's PATH. Unfortunately, this
+interfered with reasoning about PATH at all when using that python interpreter.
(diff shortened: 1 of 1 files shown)


docs/source/user-guide/troubleshooting.rst
```diff
@@ -6,17 +6,112 @@ Troubleshooting
...
+     renaming them to add .bak to the filename to effectively hide them. Observe
+     if any other software breaks. Try moving the DLL files alongside the exe of
+     the software that broke. If it works again, you can keep things in the
+     moved state - Anaconda doesn't need MKL in System32, and nothing should. If
+     you identify software that is installing software here, please contact the
+     creators of that software and inform them that their practice of installing
+     MKL to a global location is fragile and is breaking other people's software
+     and wasting a lot of time.  See the list of guilty parties below.
+  2. You may try a special DLL loading mode that Anaconda builds into Python.
+     This changes the DLL search path from system32 first to system32 last.
```
mingwandroid: It's technically not 'system32 last', it's 'system32 from where it appears in PATH`.

docs/source/user-guide/troubleshooting.rst
```diff
@@ -6,17 +6,112 @@ Troubleshooting
...
+
 Cause
------
+~~~~~
+
+You may observe this error cropping up after a conda update. More recent
+versions of conda and more recent builds of python are more strict about
+requiring activation of environments. We're working on better error messages for
+them, but here's the story for now. Activation on Windows is a must. There's
+more information on activation at the activate-env_ page. When you don't activate your environment,
```
mingwandroid: 'Activation on Windows is a must' .. well, people *can* add to PATH on Windows and it works as well as it does on other OSeS, if you exclude the problems of conda 4.6.0..4.6.4. I'd rather we removed all conda 4.6 before 4.6.6 here since that's where the bug actually was all along.

docs/source/user-guide/troubleshooting.rst
msarahan: Thanks, I've adjusted this.

docs/source/user-guide/troubleshooting.rst
msarahan: Have we actually pinpointed the change that made 4.6.0..4.6.4 break with PATH?  I'm not sure we have.  In my view, it is the combination of the removal of Library/bin (in python 3.7) along with some change in conda.

Removing all the older 4.6 releases is pretty dramatic.  While I understand your reasons for wanting to do that, it's something we'll have to think through very carefully.

mingwandroid:
> Have we actually pinpointed the change that made 4.6.0..4.6.4 break with PATH? 

I don't think condas branch strategy is conducive to that. It's the whole condabin instead of numerous paths thing though. Just not appropriate with how we've laid things out on Windows. Maybe ask @kalefranz?

docs/source/user-guide/troubleshooting.rst
```diff
@@ -6,17 +6,114 @@ Troubleshooting
...
+
+If you are not activating your environments, start with doing that. There's more
+info at the activate-env_ page. If you are still stuck, you need to consider
+more drastic measures.
+
+  1. remove any mkl-related files from C:\Windows\System32. We recommend
+     renaming them to add .bak to the filename to effectively hide them. Observe
+     if any other software breaks. Try moving the DLL files alongside the exe of
+     the software that broke. If it works again, you can keep things in the
+     moved state - Anaconda doesn't need MKL in System32, and nothing should. If
```
csoja: "nothing should"  I don't follow what this is trying to say.  Is this no other software should need it there either?  if so, I'd recommend rewording it as "no other software should need it either"

docs/source/user-guide/troubleshooting.rst
```diff
@@ -6,17 +6,114 @@ Troubleshooting
...
+       * python 3.6.8 build 7
+       * python 3.7.2 build 8
+
+     Control environment variables:
+
+       * `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`
+       * `CONDA_DLL_SEARCH_MODIFICATION_DEBUG`
+       * `CONDA_DLL_SEARCH_MODIFICATION_NEVER_ADD_WINDOWS_DIRECTORY`
+       * `CONDA_DLL_SEARCH_MODIFICATION_NEVER_ADD_CWD`
+
```
csoja: I'd recommend specifying an example conda command for updating to that build or later and where they should set those control environment variables and to what.  Reading this, I think I could update python appropriately, but I don't know what to do with the control environment variable information.  (and some people running into this problem may be even more newbie than me)

docs/source/user-guide/troubleshooting.rst
msarahan: I wrote a bigger section on this.  Please take another look.

PR #8324, opened 2019-02-22
Wrap more calls to conda because it may not be on PATH properly

mingwandroid:
And make all MS-DOS commands uppercase BECAUSE THIS WAS THE HEIGHT
OF TECHNOLOGICAL SOPHISTICATION BACK IN THE OLDEN DAYS AND DID NOT
MEAN SHOUTING LIKE IT DOES NOW.


jjhelmus:
Can this be closes now as this work is expanded upon in #8326?

msarahan:
rolled into #8369; closing

PR #8336, opened 2019-02-25
Build docs

soapy1:


circle.yml
```diff
@@ -158,6 +173,17 @@ flake8: &flake8
...
+  working_directory: ~/conda
+  docker:
+    - image: readthedocs/build
+  steps:
+    - checkout
+    - run:
+        command: |
+          pip install -r utils/requirements-docs.txt
+          cd docs
+          make html
```
msarahan: If I'm reading this correctly, you're building the docs, but you're not deploying them anywhere.  Is that correct?  Should we figure out a way to push them somewhere, or perhaps trigger the temporary addition of a branch to RTD, followed by automated removal of that branch once the PR has been merged?

soapy1: Correct. I looked into doing that but RTD doesn't have a public API that allows that currently. Another option could be to expose these docs on the github pages branch?
The current state of the PR will at least try to remove the situation where the docs don't build at all.

msarahan: Yeah, maybe we can bring the github pages view back, and have that be our testbed.  With more than one docs PR at a time, it might get hairy, but it's better than nothing.

msarahan:
Why did the tests run on circle and appveyor here?  It looks like they were manually cancelled.  Is this PR ready, or does it still need some work?

soapy1:
So, I was trying for a bit to build the docs and push them to gh-pages. However, the crux of the problem is that we don't want forks to have access to deploy type keys. So, instead of doing this in a ci in conda, we should use something like a git webhook and make some private service do it. So I think the work in this PR is done and it's ready to be merged baring any other changes anyone might want.

msarahan:
cycling to re-run ci

PR #8350, opened 2019-02-26
Post link last

soapy1:


conda/core/link.py
```diff
@@ -524,14 +524,39 @@ def _verify(cls, prefix_setups, prefix_action_groups):
         return exceptions
 
     @classmethod
-    def _execute(cls, all_action_groups):
+    def _execute(cls, prefix_action_groups):
+        groups = list(interleave(itervalues(prefix_action_groups)))
```
jjhelmus: Can a tuple be used here rather than a list?  Perhaps creating a dictionary with action types would be even better so the actions do not need to be referred to by indices below.

conda/core/link.py
```diff
@@ -238,7 +238,7 @@ def execute(self):
         assert not context.dry_run
 
         try:
-            self._execute(self.prefix_action_groups)
+            self._execute(tuple(concat(interleave(itervalues(self.prefix_action_groups)))))
```
msarahan: I often find it better for debugging purposes to have data prep steps separate from function calls.  I find that it makes tweaking and inspecting what goes into the function easier.  This is not blocking this PR - just something to think about.

msarahan:
Is this covered by existing tests, or should we add something new?

soapy1:
Looks like there are some tests https://github.com/conda/conda/blob/master/tests/test_create.py#L2050-L2072. I think they cover the basic functionality. Do you think a test is needed to do something like ensure order of actions running?

msarahan:
That could be good for peace of mind.  I think you should do something like:

package a: depends on b.  post-link script depends on the presence of a file created by b (errors if not present)
package b: generates a file in $PREFIX in a post-link script

the key being alphabetical order intentionally not working, and things only working if toposort is done correctly.

msarahan:
failure on appveyor is because the test packages are not available for win-32.  Could they be noarch?

soapy1:
Updated packages are kicked the win-32 build

PR #8369, opened 2019-03-03
Ray activation fixes

mingwandroid:
testing


msarahan:
I'm realizing that my latest commits are tackling the output mismatch wrong.  I believe the exception output to stderr has to be in the native format, whatever that may be.  Outputting it in unicode means the leading u on python 2.7, and that's explicitly in the string that we read in for the tests.  Nasty.

msarahan:
closing as superseded by #8406

PR #8374, opened 2019-03-04
abort tarball extraction if unsafe path encountered

msarahan:
Conda's handling of tarball extraction leaves potential for malicious tarballs to do naughty things.  More at

https://bugs.python.org/issue1044#msg55464
https://nvd.nist.gov/vuln/detail/CVE-2007-4559

The most recent development in Python seems to be at https://bugs.python.org/issue21109

This patch is intended to exit out when malcious paths are encountered, but we should study the most recent development in Python mentioned above to double-check.

This code will be rendered obsolete by the switch to libarchive (via conda-package-handling) thanks to its security flags: https://github.com/conda/conda-package-handling/blob/master/conda_package_handling/tarball.py#L68-L77


msarahan:
Note: this code is really a placeholder.  The right answer to this is our libarchive-based tarball handling.  This implementation is here to get 4.6.x into a better state, but it should be replaced by the conda-package-handling code.

conda/gateways/disk/create.py
```diff
@@ -175,6 +175,11 @@ def extract_tarball(tarball_full_path, destination_directory=None, progress_upda
         if progress_update_callback:
             fileobj = ProgressFileWrapper(fileobj, progress_update_callback)
         with tarfile.open(fileobj=fileobj) as tar_file:
+            for member in tar_file.getmembers():
+                if (os.path.isabs(member.name) or
+                        not os.path.realpath(member.name).startswith(os.getcwd())):
+                    sys.exit("tarball {} contains unsafe path: {}.  Erroring out."
```
jjhelmus: Would raising an error be more appropriate here?  That would allow conda's error handling and logging framework to capture and report the error.

kalefranz:
Before merging this PR I'd suggest somebody quantify the performance impact, just so we know the baseline cost.

msarahan:
The performance penalty here is going to suck. We should probably put it behind the safety checks parameter. Maybe we can skip safety checks for signed packages from known channels or something.

Issue #8384, opened 2019-03-07
conda run fails with arguments containing spaces


IRDonch:
## Current Behavior

If one of the arguments passed to `conda run` contains spaces, it's not passed correctly to the program being run.

### Steps to Reproduce

```
>conda run -n base python -c "import sys"
ERROR conda.cli.main_run:execute(26): Subprocess for 'conda run' command failed.  Stderr was:
The system cannot find the path specified.
  File "<string>", line 1
    import
         ^
SyntaxError: invalid syntax
```

## Expected Behavior

There should be no output (since `python -c "import sys"` doesn't produce any).

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3
            shell level : 1
       user config file : C:\Users\[redacted]\.condarc
 populated config files :
          conda version : 4.6.7
    conda-build version : not installed
         python version : 3.7.1.final.0
       base environment : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\pkgs
                          C:\Users\[redacted]\.conda\pkgs
                          C:\Users\[redacted]\AppData\Local\conda\conda\pkgs
       envs directories : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\envs
                          C:\Users\[redacted]\.conda\envs
                          C:\Users\[redacted]\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.6.7 requests/2.21.0 CPython/3.7.1 Windows/10 Windows/10.0.16299
          administrator : False
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```

```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

```
[...]

PR #8389, opened 2019-03-07
Fix spaces in the root path

mingwandroid:


msarahan:
Is this to fix #8384 ?

mingwandroid:
Unfortunately not, no. Without this you get unnecessary (non-existent) entries on PATH if the batch file itself is located in a directory with any spaces in it. Not sure why our tests aren't covering this though, we should probably look to fix that.

mingwandroid:
Though that is fixed (and tested for) in my new PR.

msarahan:
OK.  I think we should just go ahead with your other PR, then?

mingwandroid:
Fine by me either way. I'll keep pushing my changes forward.

msarahan:
superseded by #8406, closing

PR #8400, opened 2019-03-11
implement 'conda init --reverse' to undo rc file changes

msarahan:


conda/core/initialize.py
```diff
@@ -536,6 +552,7 @@ def find_powershell_paths(*exe_names):
                 'kwargs': {
                     'target_path': join(conda_prefix, 'condabin', 'Anaconda Prompt.lnk'),
                     'conda_prefix': conda_prefix,
+                    'reverse': reverse,
```
soapy1: What does it mean to run a `conda init --reverse` in the anaconda prompt? Do you just have a regular cmd line then?

msarahan: right now, nothing.  This is not used to create the anaconda prompt.  @kalefranz planned for it to replace the other stuff, but that's not in place right now.  I'd just ignore anything regarding "anaconda prompt" in the conda codebase for now.

PR #8401, opened 2019-03-11
Insert current conda into sys path

soapy1:


msarahan:
LGTM.  Since this is only for docs, this is not disruptive or terribly controversial.

soapy1:
Please don't merge yet, gotta do some testing with it

PR #8403, opened 2019-03-11
Changed color scheme and thematic elements

rrigdon:
This commit updates the conda site to match the colors and themes of Distro and conda-build.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -1,0 +1,83 @@
+@import url("theme.css");
+
+.wy-nav-content {
+    padding: 1.618em 3.236em;
+    height: 100%;
+    max-width: 1500px;
+    /* max-width: 800px; */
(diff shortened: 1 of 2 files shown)


docs/source/_static/css/custom.css
```diff
@@ -0,0 +1,83 @@
+@import url("theme.css");
```
soapy1: just curious, what is theme.css?

rrigdon: I'm not sure. It was the initial line in the conda-build CSS and I copied that over for conda. I don't know if it's necessary but @msarahan may know since I think he created the original .css in conda-build.

msarahan: whoa, what?  I don't remember anything like that.  I have no idea what I'm doing with CSS.  Maybe I was copying over some other template or rearranging things.

PR #8436, opened 2019-03-21
Added flowchart and information on installing with conda

rrigdon:


docs/source/user-guide/concepts.rst
```diff
@@ -6,6 +6,67 @@ Concepts
...
+    :align: right
+
+.. _installing-with-conda:
+
+Installing with conda
+=====================
+
+Conda install
+-------------
+* When you `conda install <package>` the happy path (with no dependencies) is that conda:
```
soapy1: I think `the happy path` is a little confusing here, that's my bad. I think what it really means is "when you install a package that exists in your channels and has no dependencies"

docs/source/user-guide/concepts.rst
```diff
@@ -6,6 +6,68 @@ Concepts
...
+
+  * channels: list of channels for conda to search for packages
+
+  * default_channels: normally pointing to channels at repo.continuum.io, sets the
+    list of "default channels"
+
+  * allow_other_channels: a boolean value that determines if the user may install
+    packages outside of the channels list
+
+  * channel_alias: sets an alias for a channel. For example, if `channel_alias: https://my.repo`
```
msarahan: Honestly, I'd advise removing  all of these except for ``channels``.  The others are for advanced applications, and they really don't belong in a "concepts" section.  Perhaps referencing the more-advanced section of the docs from here is a better way than showing those options completely here.

docs/source/user-guide/concepts.rst
```diff
@@ -6,6 +6,48 @@ Concepts
...
+    :align: right
+
+.. _installing-with-conda:
+
+Installing with conda
+=====================
+
+Conda install
+-------------
+* When you `conda install` a package that exists in your channels and has no dependencies, conda:
```
jjhelmus: "your" seems a bit odd here as the majority of users will be installing from a channel they are not owners of.  Perhaps "a channel" would be better.

docs/source/user-guide/concepts.rst
```diff
@@ -6,6 +6,48 @@ Concepts
...
+.. _specifying-channels:
+
+Specifying channels
+-------------------
+
+* From the command line use `--channel`
+
+.. code-block:: bash
+
+  $ conda install scipy --channel conda-forge
```
msarahan:
```suggestion
  $ conda install scipy --channel conda-forge
  
  You may specify multiple channels by passing the argument multiple times:
  
.. code-block:: bash

  $ conda install scipy --channel conda-forge --channel bioconda
  
  Priority decreases from left to right - the first argument is higher priority than the second.
```

docs/source/user-guide/concepts.rst
```diff
@@ -6,6 +6,48 @@ Concepts
...
+
+* From the command line use `--override-channels` to only search the specified channel
```
msarahan:
```suggestion
* From the command line use `--override-channels` to only search the specified channel(s), rather than any channels configured in .condarc.  This also ignores conda's default channels.
```

Issue #8311, opened 2019-02-20
Noarch(type='python') is not a valid NoarchType error importing from Artifactory channel


analog-cbarber:
## Current Behavior

Using conda 4.6.2 through current version (4.6.4) I get the following error when I try to install a package from a conda channel on our local artifactory server. This error does not occur when using 4.5:

```
kadenz-0.1.4.dev1    | 11 KB     | ##################################### | 100% 
Preparing transaction: done
Verifying transaction: done
Executing transaction: failed
ERROR conda.core.link:_execute(543): An error occurred while installing package 'https://artifactory.analog.com/artifactory/garage-conda::kadenz-0.1.4.dev1-py_0'.
ValidationError(ValueError("Noarch(type='python') is not a valid NoarchType",),)
Attempting to roll back.

Rolling back transaction: done
```

The entry for the this package from the `repodata.json` on the Artifactory server is:

```
"kadenz-0.1.4.dev1-py_0.tar.bz2" : {
      "build" : "py_0",
      "build_number" : 0,
      "depends" : [ "python >=3.5.*" ],
      "license" : "Apache 2",
      "md5" : "928b777a9aa557e74ba6c139562ce1b0",
      "name" : "kadenz",
      "sha256" : "42145ebce75a6f8b8ec175bad547ba6a73886ed809bdafea7fab1587fe155250",
      "size" : 11228,
      "subdir" : "noarch",
      "timestamp" : 1549393159581,
      "version" : "0.1.4.dev1"
    },
```
Installing the same package from my local `conda-bld` cache using `--use-local` does not fail.
The corresponding entry from my `repodata.json` in my local `conda-bld` is:

```
"kadenz-0.1.4.dev1-py_0.tar.bz2": {
      "build": "py_0",
      "build_number": 0,
      "depends": [
        "python >=3.5.*"
      ],
      "license": "Apache 2",
      "md5": "928b777a9aa557e74ba6c139562ce1b0",
      "name": "kadenz",
      "noarch": "python",
      "sha256": "42145ebce75a6f8b8ec175bad547ba6a73886ed809bdafea7fab1587fe155250",
      "size": 11228,
      "subdir": "noarch",
      "timestamp": 1549393159581,
      "version": "0.1.4.dev1"
    },
```

## Environment Information

<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : foo
    active env location : /Users/cbarber/miniconda3/envs/foo
            shell level : 1
[...]

PR #8438, opened 2019-03-22
Workaround for issue 8311

analog-cbarber:
Since it is clear that Artifactory is no hurry to fix their broken conda repository indexing, I am submitting my proposed workaround for for issue #8311.

I only changed one place in `path_actions.py` that was accessing noarch from the package metadata. It is possible that the two other locations may be needed for other uses cases.


analog-cbarber:
I have signed the CLA. I guess it takes a little while for it to show up in the check?

I think that the remaining failures are in previous pull requests so probably not related to my change.

msarahan:
The CLA is manual.  I'll add you, but it probably won't every show green on this PR.  Just ignore it.

We'll look at the tests as time allows.  Right now, we are completely tied up with preparing for AnacondaCon in 2 weeks or so.  It will take us time to review this, but we should definitely be faster than the Q3/Q4 that artifactory has estimated.

Bezier89:
@msarahan Congrats on a successful AnacondaCon!

Now that it's over and this PR has sat for a couple weeks, think anyone could take a look at it? We're also stuck on conda 4.5 until this is resolved :(

msarahan:
CI fails look unrelated.  Looks fine.  Is there a test that could be added to prevent future regressions?

analog-cbarber:
There should be a way to add a test, but I won't be able to work on this until I get back from vacation in ten days.


On Apr 8, 2019 4:22 PM, Mike Sarahan <notifications@github.com> wrote:
[External]


CI fails look unrelated. Looks fine. Is there a test that could be added to prevent future regressions?

Bezier89:
@msarahan @analog-cbarber 

Created https://github.com/conda/conda/pull/8523/files that adds a simple test.

PR #8451, opened 2019-03-25
Handle read-only FS for environments.txt

spalmrot-tic:
In my Disaster Recovery environment we have read-only home directories because they are replicated from our normal environment.

### DIFF
--- a/conda/core/envs_manager.py
+++ b/conda/core/envs_manager.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import absolute_import, division, print_function, unicode_literals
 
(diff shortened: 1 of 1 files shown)


kalefranz:
You’ll need to sign the CLA at https://docs.conda.io/en/latest/contributing.html before this could be merged.

spalmrot-tic:
Done

Thank you,
Steve

spalmrot-tic:
The main test failures seem completely unrelated to the minor change that I made. I am unsure as to the problem.

msarahan:
Conda-forge changed something on their side.  We need to fix the test.  Thanks for your work here.

PR #8453, opened 2019-03-26
Edge case for RO FS & environments.txt

spalmrot-tic:
I found an edge case related to #8451 where if the environments.txt file didn't exist package installation would fail when the user home directory was on a read-only file system. This patch prevents failure when trying to create the environments.txt or if the file doesn't exist when trying to register the environment.


msarahan:
Confirming that you have signed the CLA (it's a manual process to add you).  This looks fine to me.  Merging.

spalmrot-tic:
Yes, I signed the CLA.

Thank you,
Steve

PR #8457, opened 2019-03-27
Ensure COMSPEC is set

soapy1:


conda/common/constants.py
```diff
@@ -12,3 +12,5 @@
 NULL = NULL
 
 CONDA_TEMP_EXTENSION = '.c~'
+
+COMSPEC_PATH = "C:\\Windows\\system32\\cmd.exe"
```
msarahan: We can't hardcode it like this.  There's a couple of fallbacks in https://github.com/ContinuumIO/anaconda-distribution-installer/pull/11/files#diff-11d98bb00d3a7be9301a2b03fcd8b9c8R251

conda/utils.py
```diff
@@ -286,3 +288,15 @@ def wrap_subprocess_call(on_win, root_prefix, prefix, command):
         command_args = [shell_path, "-x", script_caller]
 
     return script_caller, command_args
+
+
+def ensure_comspec_set():
+    if 'COMSPEC' not in environ:
+        cmd_exe = join(environ['SystemRoot'], 'System32', 'cmd.exe')
```
jjhelmus: Is a guard/default needed if 'SystemRoot' or 'windir' are not in environ?

soapy1: Not sure what other guards should be put in place here other than a hard code to cmd.exe.... I found this article with default windows environment variables https://ss64.com/nt/syntax-variables.html. 
Maybe there should also be a check for `join(environ['SystemRoot'], 'SysWOW64', 'cmd.exe')`

msarahan: I think Jonathan is more referring to using something like .get() on environ, so that we don't crash out with a keyerror.  I think that's a good idea here.

soapy1: ahh, gotcha

conda/utils.py
```diff
@@ -286,3 +288,15 @@ def wrap_subprocess_call(on_win, root_prefix, prefix, command):
...
     return script_caller, command_args
+
+
+def ensure_comspec_set():
+    if 'COMSPEC' not in environ:
+        cmd_exe = join(environ.get('SystemRoot'), 'System32', 'cmd.exe')
+        if not isfile(cmd_exe):
+            cmd_exe = join(environ.get('windir'), 'System32', 'cmd.exe')
+        if not isfile(cmd_exe):
+            log.info("cmd.exe could not be found. "
```
nehaljwani: Should this be .info() or .warn()?

conda/utils.py
```diff
@@ -286,3 +288,15 @@ def wrap_subprocess_call(on_win, root_prefix, prefix, command):
         command_args = [shell_path, "-x", script_caller]
 
     return script_caller, command_args
+
+
+def ensure_comspec_set():
+    if 'COMSPEC' not in environ:
```
msarahan: @soapy1 we need to revisit this, similar to our usage in the installers.  It's not enough that it is set, it needs to be set to cmd.exe.

Issue #8393, opened 2019-03-08
conda deactivate from package deactivate.d script not


tomashek:
## Current Behavior
We have a package that uses conda env hooks for activate and deactivate. The activation script in etc/conda/activate.d adds an directory to the front of the PATH, and the etc/conda/deactivate.d script uses sed to remove it. We are having a problem, though, because the element is still present after deactivation. The confusing part is if I output the PATH prior to exiting the deactivate script, the element is gone. Other env variables get unset properly, but the PATH is problematic.

### Steps to Reproduce
Linux-64 bit
The package exhibiting the behavior is:
intel::impi_rt-2019.3-intel_200.tar.bz2
I have tried using conda 4.5.12 and 4.6.7
```
conda create -n impitest python=3.6 impi_rt -c intel
conda activate impitest (adds $CONDA_PREFIX/bin/libfabric:$CONDA_PREFIX/bin to the PATH)
conda deactivate
```

## Expected Behavior
$CONDA_PREFIX/bin/libfabric:$CONDA_PREFIX/bin should no longer be in the PATH. If a simple "echo $PATH" statement is added to the end of the deactivate script, it shows the PATH with the elements removed, but the user shell PATH still contains those elements.

Looks similar to #3915

tomashek:
Note: I moved impi_rt-2019.3-intel_200.tar.bz2 to intel/label/validation and replaced it with impi_rt-2019.3-intel_201.tar.bz2 that adds only $CONDA_PREFIX/bin/libfabric to the PATH. This minimizes the unwanted items in the PATH since PREFIX/bin is cleared correctly. PREFIX/bin/libfabric still remains a problem.

nickelsey:
this, I believe, is a known problem: see #8070... I am also waiting for a resolution to this.

msarahan:
No, I think #8070 is different.  #8070 was a request that conda deactivation clean up any effects from activate.d scripts (i.e. back up and restore variables).  Conda absolutely will not do that, for reasons discussed in #8070.  This issue is a bug where changes made in deactivate.d don't appear to be "sticking."

@tomashek, thanks for the report.  We'll try to fix it.

jiyub:
The PATH issue was resolved with #8464, but there was another issue present with MANPATH at the same time. MANPATH gets modified by activate.d script (export MANPATH="${I_MPI_ROOT}/share/man:${MANPATH}"), but does not get removed when issuing "conda deactivate".

cchriste:
[...]

PR #8459, opened 2019-03-27
Remove all path variables related to environment being deactivated

soapy1:
fixes https://github.com/conda/conda/issues/8393

Needs some tests too


soapy1:
This is the wrong implementation, closing

Issue #8346, opened 2019-02-26
conda env create Not Working With environment.yml File


cSchubes:
## Current Behavior
I am trying to create a conda environment using an environment.yml file on Windows 10 inside the Anaconda Prompt. When I do so, none of the packages listed in the environnment.yml actually get installed, only the defaults. You can see the command executing below with verbosity set to 2:
```
(base) C:\Users\Me\Documents\Stuff\raven>conda env create --file=./environment_windows.yml -v -v
DEBUG conda.gateways.logging:set_verbosity(150): verbosity set to 2
DEBUG conda.core.solve:solve_final_state(212): solving prefix C:\Users\Carson\Anaconda3\envs\raven
  specs_to_remove: frozenset()
  specs_to_add: frozenset({MatchSpec("certifi=2018.11.29"), MatchSpec("sqlite=3.26.0"), MatchSpec("zlib=1.2.11"), MatchSpec("setuptools=40.8.0"), MatchSpec("python=3.6.8"), MatchSpec("wheel=0.32.3"), MatchSpec("xz=5.2.4"), MatchSpec("ca-certificates=2019.1.23"), MatchSpec("tk=8.6.8"), MatchSpec("pip=19.0.1"), MatchSpec("openssl=1.1.1a")})
  prune: False
Collecting package metadata: ...working... DEBUG conda.core.package_cache_data:_check_writable(252): package cache directory 'C:\Users\Carson\Anaconda3\pkgs' writable: True
DEBUG conda.core.subdir_data:_load(213): Local cache timed out for https://repo.anaconda.com/pkgs/r/noarch at C:\Users\Carson\Anaconda3\pkgs\cache\4ea078d6.json
DEBUG urllib3.util.retry:from_int(210): Converted retries value: 3 -> Retry(total=3, connect=None, read=None, redirect=None, status=None)
DEBUG urllib3.connectionpool:_new_conn(813): Starting new HTTPS connection (1): repo.anaconda.com:443
DEBUG urllib3.connectionpool:_make_request(393): https://repo.anaconda.com:443 "GET /pkgs/r/noarch/repodata.json.bz2 HTTP/1.1" 304 0
DEBUG conda.core.subdir_data:fetch_repodata_remote_request(431):
>>GET /pkgs/r/noarch/repodata.json.bz2 HTTPS
> User-Agent: conda/4.6.7 requests/2.21.0 CPython/3.7.1 Windows/10 Windows/10.0.17134
> Accept: */*
[...]

PR #8435, opened 2019-03-21
Ray activation fixes 4

mingwandroid:
Rebased and some of test_create.py ran OK on macOS, not tested yet beyond that.

PR #8460, opened 2019-03-28
change pip_subprocess from "pip <cmd>" to "python pip-script.py <cmd>" on windows

zdog234:
Hello all,

This is motivated by #8346. Prior to 4.6.7, the generated syntax for pip commands had been `/path/to/venv/python.exe /path/to/venv/Scripts/pip-script.py <cmd>` but was changed to  `pip <cmd>`.

I don't know how this could've been caught -- it seems like `utils.wrap_subprocess` should have kept this from happening -- but this particular change has caused `pip` dependencies in `environment.yml` to not properly install into new environments when running `conda env create -f environment.yml`. It seems like they're instead being installed into the `base` environment.

A bunch of unit tests are failing, but it seems like they're the same as the ones on master.

It's totally possible that this is the wrong fix. I just figured I'd submit this PR as a conversation starter/to get help from people who know what they're doing :)

Also, this probably deserves a unit test, but I haven't taken the time yet to figure out how the test framework works for this project.

### DIFF
--- a/conda_env/pip_util.py
+++ b/conda_env/pip_util.py
@@ -51,6 +51,8 @@ def pip_args(prefix):
         pip_path = join(prefix, 'bin', 'pip')
         py_path = join(prefix, 'bin', 'python')
     if isfile(pip_path) and isfile(py_path):
+        py_path = '"{}"'.format(py_path)
+        pip_path = '"{}"'.format(pip_path)
         ret = [py_path, pip_path]
         return ret
     else:
--- a/conda_env/pip_util.py
+++ b/conda_env/pip_util.py
@@ -10,8 +10,10 @@ from __future__ import absolute_import, print_function
 
 import json
 import os
+from os.path import isfile, join
 import re
 import subprocess
+import sys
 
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks for your efforts here.  Indeed, we've been very confused how the tests didn't catch this.  We've done a lot of work on fixing activation and running things in activated envs.  Unfortunately, it spiraled out of control and has become a much larger body of work in #8435.

If you could please keep an eye out for when we release 4.6.9 into canary - ideally in the next 1-2 days - we'd really appreciate your confirmation that our efforts have addressed the issue for you.

zdog234:
Thanks! I can see that you guys are putting in a lot of work there. The new implementation of `pip_subprocess` looks really neat!

msarahan:
@zdog234 conda 4.6.9 has been released to the conda-canary channel.  Any feedback you have for us would be most appreciated.  I'm closing this PR as I hope this issue is fixed.

zdog234:
Sweet! I'll try to check some time this weekend :)

PR #8464, opened 2019-03-29
Export PATH prior to deactivate

soapy1:
I think PATH can be treated differently on deactivation since deactivation scripts might depend on stuff that exists in it. 
Should resolve https://github.com/conda/conda/issues/8393

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -266,6 +266,9 @@ class _Activator(object):
         self.command = command
 
     def _yield_commands(self, cmds_dict):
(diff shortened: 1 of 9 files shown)


cchriste:
This seems to be doing almost what we want, but the PATH after the deactivate script(s) run gets reset to what was created _after_ the activate scripts(s) have run. I've seen a few comments about this, but no documentation for a resolution (and it's still happening). Any pointers appreciated.

Issue #8213, opened 2019-02-04
Powershell prompt doesn't work with cmder


dhirschfeld:

dennispg:
I submitted a pull request for this. Use `conda config --set changeps1 false` and then in cmder or posh-git or whatever else, configure your prompt however you want by reading from `$Env:CONDA_DEFAULT_ENV`.

musm:
Yeah the prompt is not modified when activating environments.

dennispg:
@dhirschfeld I can confirm the change in my pull request work great for cmder with PowerShell, are you also able to confirm?

I think this can be closed now.

apwebb:
Hello, I may be missing something but I am still having these issues with conda/cmder/powershell/pwsh.
Specifically, I cannot properly activate/deactivate environments, although other basic conda commands (eg. install, list, env list) work fine. I just stay in the base environment.  I have run `conda init powershell`. The value of `changeps1` doesn't seem to matter. Further, importing the conda.psm1 module from shell\condabin throws the following error on usage in pwsh/powershell through cmder, but not native pwsh/powershell, where they work fine. The below error is thrown for any command from the module. 

System: Microsoft Windows 10 Pro 10.0.17763
Cmder: 1.3.11
Powershell: 5.1.17763.316
Powershell Core: 6.2.0
Conda: 4.6.14 (miniconda3 installed via scoop)

```
❯ import-module $HOME\scoop\apps\miniconda3\current\shell\condabin\conda.psm1
❯ genv
The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.
At ~\scoop\apps\miniconda3\current\shell\condabin\conda.psm1:23 char:11
+         & $Env:CONDA_EXE $Env:_CE_M $Env:_CE_CONDA env list | `
+           ~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidOperation: (:) [], RuntimeException
+ FullyQualifiedErrorId : BadExpression
```
FWIW, conda-hook.ps1 appears to be set correctly (only CONDA_EXE is set to a non-empty string and it is the correct path), which makes sense if it all works with native pwsh/powershell

dennispg:
Yes, the step you are missing is executing `conda-hook.ps1`. (That, or setting those environment variables one way or another yourself..)

[...]

PR #8465, opened 2019-03-29
Dont modify Function:prompt when "changeps1" is false in PowerShell

dennispg:
Proposed fixes for #8213

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -169,6 +169,9 @@ class _Activator(object):
             builder.append(fsrc.read())
         if auto_activate_base is None and context.auto_activate_base or auto_activate_base:
             builder.append("conda activate base\n")
+        postamble = self._hook_postamble()
+        if postamble is not None:
+            builder.append(postamble)
         return "\n".join(builder)
 
     def execute(self):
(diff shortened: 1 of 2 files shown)


msarahan:
I confirm that you've signed the CLA.  Thanks for working on this.  We're incredibly busy through next week with AnacondaCon, but we'll look forward to reviewing this with you soon.

msarahan:
@dennispg thanks for this PR.  It looks like a format string needs more arguments now.  Can you please take a look at the tests on circle CI and see if you can address them?  If you need help, please feel free to ask.

dennispg:
So during deactivate there are some env vars that get unset.. After I added this change I noticed that with `changeps1` sometimes some of those vars which would have been set are not. But the deactivate tries to happily remove them anyways, so I added a check to be sure it is set before unsetting.

Now the automated tests seem to have some assumptions made as well about what vars are set and not, and I'm not sure how to make the asserts match up.

dennispg:
Hi @msarahan, I rolled back the checks I put in to check for the env variables before trying to unset.. I think there is still some bug in that process, but I can't figure out the right way to sort out the tests without some assistance.

But it is probably best left to a new and separate issue.. This limited change should serve the purpose of only modifying the global prompt function when actually changing the prompt.

msarahan:
LGTM.  Thanks!

Issue #8192, opened 2019-01-31
[4.6.2] Regression with conda env create -n NAME


dhirschfeld:
When I try the `conda env create -n NAME -f environment.yml` command with a non-existent yml file (i.e. from the wrong directory) I get an inscrutable `SpecNotFound` error:
```
(base) C:\dev\code> conda env create -n py-dev -f quantdev.yml

SpecNotFound: Invalid name, try the format: user/package

(base) C:\dev\code>
```

It seems this problem has cropped up previously (#5681) and was supposed to have been fixed in #7385 so this appears to be a regression?

dhirschfeld:
```
(base) C:\dev\code> conda info

     active environment : base
    active env location : C:\Miniconda3
            shell level : 1
       user config file : C:\Users\dhirschfeld\.condarc
 populated config files : C:\Users\dhirschfeld\.condarc
          conda version : 4.6.2
    conda-build version : 3.17.7
         python version : 3.7.1.final.0
       base environment : C:\Miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
                          http://conda.stanwell.com/conda-forge/win-64
                          http://conda.stanwell.com/conda-forge/noarch
          package cache : C:\Miniconda3\pkgs
                          C:\Users\dhirschfeld\.conda\pkgs
                          C:\Users\dhirschfeld\AppData\Local\conda\conda\pkgs
       envs directories : C:\Miniconda3\envs
                          C:\Users\dhirschfeld\.conda\envs
                          C:\Users\dhirschfeld\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.6.2 requests/2.21.0 CPython/3.7.1 Windows/7 Windows/6.1.7601
          administrator : True
             netrc file : None
           offline mode : False
```

javabrett:
I've spotted some bugs in https://github.com/conda/conda/blob/master/tests/conda_env/test_cli.py which result in regressions such as this not being picked-up.

[...]

PR #8468, opened 2019-04-01
[tests] conda_env create tests confuse -f and -n

javabrett:
When a prefix is passed to `run_env_command` with `Commands.ENV_CREATE` it should be passed to `-n`, not `-f`.  Fixed that and corrected tests calling that helper method.

Noticed while investigating #8192.

### DIFF
--- a/tests/conda_env/test_cli.py
+++ b/tests/conda_env/test_cli.py
@@ -83,7 +83,7 @@ def run_env_command(command, prefix, *arguments):
         arguments[1:1] = ['-n', prefix]
     elif command is Commands.ENV_CREATE: # CREATE
         if prefix:
-            arguments[1:1] = ['-f', prefix]
+            arguments[1:1] = ['-n', prefix]
     elif command is Commands.ENV_REMOVE:  # REMOVE
         arguments[1:1] = ['--yes', '-n', prefix]
     elif command is Commands.ENV_UPDATE:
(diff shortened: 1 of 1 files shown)


tests/conda_env/test_cli.py
```diff
@@ -83,7 +83,7 @@ def run_env_command(command, prefix, *arguments):
         arguments[1:1] = ['-n', prefix]
     elif command is Commands.ENV_CREATE: # CREATE
         if prefix:
-            arguments[1:1] = ['-f', prefix]
+            arguments[1:1] = ['-n', prefix]
```
msarahan: I think we should test ``prefix`` to see if it is path-like (has slashes, leading periods, is an abspath, etc.).  If it is a path, this should be -p.  If it's a name, it should be -n.

javabrett: I hadn't considered `-p`.  It's also possible I'm not across any overloaded meaning of the terms `prefix` and `name`, which are used interchangeably in the code in places.

None of the existing tests make use of `-p`, so there might be some coverage options there.  Hopefully none of the existing tests pass a path-type `-n` - this arg is positional `prefix`, or so it seems.

Existing tests are mostly creating by name and file, but important to test that the name takes priority in the created env I guess.  Just looked wrong that the prefix was passed to file in this one case, and test callers were bent to fit that.

Eventually these tests need to point out issues in
https://github.com/conda/conda/blob/9a303dcb394fe4e2a76cd97dd379ebc9df7e8679/conda_env/specs/__init__.py#L15
... where remote def and name might be passed in a single kwarg without being able to distinguish between them, causing logic issues in that method.

msarahan:
@javabrett has signed the CLA.  Thanks for your contribution!

Issue #8474, opened 2019-04-02
conda init (4.6.9): minor inconvenience with zsh


nehaljwani:
```
➜  ~ sudo docker exec -it play /bin/zsh
(eval):61: = not found
```

If I revert the commit https://github.com/conda/conda/commit/b0c04256981dcbb5378c8177e8fc04eb983dabe3 , the problem goes away.

tyxxzjpdez:
same issue

blmhemu:
+1
It was working fine until I have updated (conda update conda)

Switham007:
Same issue with me, after conda update conda: 
zsh terminal says "...miniconda3/etc/profile.d/conda.sh:61: = not found"

msarahan:
#8475 and #8476 should fix this.

Switham007:
Thanks, that fixed it for me

sandacn:
I've downgrade to 4.6.8. When this release ,  I'll check this whether is fixed.
```bash
conda install conda=4.6.8 
```

PR #8475, opened 2019-04-02
Uncomplicate logic to avoid ':' at end of PATH

nehaljwani:
Also:
- Make conda.sh less non-POSIX complaint by removing the use of '=='
 ```
  In conda/shell/etc/profile.d/conda.sh line 57:
    if [ "${_CE_CONDA+x}" == "condax" ]; then
                          ^-- SC2039: In POSIX sh, == in place of = is undefined.


  In conda/shell/etc/profile.d/conda.sh line 58:
        if [ "${PATH+x}" == "x" ]; then
                         ^-- SC2039: In POSIX sh, == in place of = is undefined.
 ```
- Escape dirname, so that aliases/functions don't surprise the end user

Fixes https://github.com/conda/conda/issues/8474

### DIFF
--- a/conda/shell/etc/profile.d/conda.sh
+++ b/conda/shell/etc/profile.d/conda.sh
@@ -54,15 +54,7 @@ conda() {
 
 if [ -z "${CONDA_SHLVL+x}" ]; then
     \export CONDA_SHLVL=0
-    if [ "${_CE_CONDA+x}" == "condax" ]; then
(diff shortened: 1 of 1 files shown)


nehaljwani:
cc @mingwandroid 
Please check whether this would result into anything unexpected.

msarahan:
py27 is anaconda.org hiccups.  Merging.  Thanks @nehaljwani

PR #8478, opened 2019-04-02
Improve detection of explicit pip specs in environment yml files

duncanmmacleod:
This PR fixes a minor bug in the detection of `pip` as a conda dependency in environment files. Without this patch, `conda env` would warn about `pip` dependencies being used when `pip` itself wasn't listed as a dependency, even if `pip` was listed like this:

```yaml
  - pip=x.y.z
```

The change uses a regular expression (`r"[\s=]"`) to detect the name `pip` in the dependencies list.

### DIFF
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, print_function
 from collections import OrderedDict
 from itertools import chain
 import os
+import re
 
 from conda.base.context import context
 from conda.cli import common  # TODO: this should never have to import form conda.cli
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
CLA has been signed, not sure how to refresh that status check.

conda_env/env.py
```diff
@@ -47,9 +48,10 @@ def validate_keys(data, kwargs):
         print('')
 
     deps = data.get('dependencies', [])
+    depsplit = re.compile(r"[\s=]")
```
msarahan: would it make sense to handle >, <, and possible ~ also in this regex?

msarahan:
Thanks!  CLA is a manual process on our end.  I have confirmed that you signed it.  I had just one small question on the regex, and otherwise this is a good improvement.  Thank you.

conda_env/env.py
duncanmmacleod: (hopefully) implemented in 2b032b6

msarahan:
Thanks!

PR #8481, opened 2019-04-02
init_fish_user: don't assume config file exists

nehaljwani:


conda/core/initialize.py
```diff
@@ -979,8 +979,13 @@ def init_fish_user(target_path, conda_prefix, reverse):
...
     user_rc_path = target_path
 
-    with open(user_rc_path) as fh:
-        rc_content = fh.read()
+    try:
+        with open(user_rc_path) as fh:
+            rc_content = fh.read()
+    except FileNotFoundError:
+        rc_content = ''
+    except:
```
msarahan: This is not really necessary.  Uncaught exceptions will always be raised.  Fine to have it here, though.

Issue #8479, opened 2019-04-02
conda constructor seems to be broken after conda upgrade to 4.6.9


huamichaelchen:
## Current Behavior

conda constructor failed to generate installer.sh after conda upgrade to 4.6.9 

### Steps to Reproduce

```
$ cat construct.yaml 
name: temp-construct
version: 2.5.5
install_in_dependency_order: True

channels:
  - 'https://conda.anaconda.org/conda-forge'
  - 'https://repo.anaconda.com/pkgs/main'
specs:
  - python
  - grpcio
  - pyyaml=5.1
```

```
$ constructor -v --output-dir='installer/' .
platform: linux-64
conda packages download: /root/.conda/constructor/linux-64
Traceback (most recent call last):
  File "/opt/conda/lib/python3.7/site-packages/conda/common/io.py", line 132, in env_vars
    callback(True)
  File "/opt/conda/lib/python3.7/site-packages/conda/base/context.py", line 1110, in reset_context
    context.__init__(search_path, argparse_args)
  File "/opt/conda/lib/python3.7/site-packages/conda/base/context.py", line 285, in __init__
    argparse_args=argparse_args)
  File "/opt/conda/lib/python3.7/site-packages/conda/common/configuration.py", line 762, in __init__
    self._set_search_path(search_path)
  File "/opt/conda/lib/python3.7/site-packages/conda/common/configuration.py", line 767, in _set_search_path
    self._search_path = IndexedSet(search_path)
  File "/opt/conda/lib/python3.7/site-packages/conda/_vendor/boltons/setutils.py", line 96, in __init__
    self.update(other)
  File "/opt/conda/lib/python3.7/site-packages/conda/_vendor/boltons/setutils.py", line 310, in update
    for o in other:
TypeError: 'bool' object is not iterable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/constructor", line 11, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.7/site-packages/constructor/main.py", line 221, in main
    dry_run=opts.dry_run)
  File "/opt/conda/lib/python3.7/site-packages/constructor/main.py", line 113, in main_build
    fcp_main(info, verbose=verbose, dry_run=dry_run)
  File "/opt/conda/lib/python3.7/site-packages/constructor/fcp.py", line 237, in main
    }, conda_reset_context):
  File "/opt/conda/lib/python3.7/contextlib.py", line 112, in __enter__
[...]

PR #8483, opened 2019-04-02
env_vars(): don't assume callback takes bool as param

nehaljwani:
Fixes https://github.com/conda/conda/issues/8479

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -129,7 +129,7 @@ def env_vars(var_map=None, callback=None):
         os.environ[name] = value
     try:
         if callback:
-            callback(True)
+            callback()
         yield
     finally:
         for name, value in iteritems(saved_vars):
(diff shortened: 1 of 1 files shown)


nehaljwani:
@mingwandroid Is this correct?

jjhelmus:
The `stack_content_defaults`, `replace_context` and `replace_context_defaults` function in base/context.py expect a single positional argument the last of which is used for `conda_test_ctxt_mgmt_def_pol`.  Should the `pushing` argument be removed from these or should a partial be used for `conda_test_ctxt_mgmt_def_pol`?

mingwandroid:
No, I'll make a PR now. We'll need new constructors too, sorry.

nehaljwani:
Closing, this PR is indeed wrong. But does bring up the point that internals had been modified, which broke constructor. We should also be running constructor tests in the CI with each commit.

msarahan:
Is there no way to modify the internals to keep the API?

jjhelmus:
Restored the previous API by adding a new `stack_callback` parameter which expects a function with a single Boolean argument.

PR #8499, opened 2019-04-04
Updated import from collections for python>=3.3

duncanmmacleod:
This PR updates an import from `collections` to remove a `DeprecationWarning` when running on python3.7.

### DIFF
--- a/conda/common/compat.py
+++ b/conda/common/compat.py
@@ -131,7 +131,7 @@ if PY3:  # pragma: py2 no cover
     viewvalues = methodcaller("values")
     viewitems = methodcaller("items")
 
-    from collections import Iterable
+    from collections.abc import Iterable
     def isiterable(obj):
         return not isinstance(obj, string_types) and isinstance(obj, Iterable)
 


conda/common/compat.py
```diff
@@ -131,7 +131,7 @@ def iteritems(d, **kw):
     viewvalues = methodcaller("values")
     viewitems = methodcaller("items")
 
-    from collections import Iterable
+    from collections.abc import Iterable
```
msarahan: This has to be wrapped in a try/except to keep py27 compat:

```suggestion
    try:
        from collections.abc import Iterable
    except ImportError:
        from collections import Iterable
```

jjhelmus: This code is inside of a `if PY3` block so there should be no need for py27 compatibility.  `collection.abc` is only needed for compatibility with Python 3.0-3.2.

msarahan: Thanks for expanding the context.  Seems good to go.

jjhelmus:
@duncanmmacleod Thanks for the fix.

PR #8507, opened 2019-04-05
fix compat import showing warning

msarahan:


marireeves:
Is there a way to still install keras within R studio?

marireeves:
Because this is the error I'm getting after updating my Anaconda version today:
> install_keras()
Error: lexical error: invalid char in json text.
                                       WARNING: The conda.compat modul
                     (right here) ------^
>

PR #8508, opened 2019-04-05
Master 8485 2

mingwandroid:
I force pushed badly it seems.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -1,3 +1,4 @@
+
 # -*- coding: utf-8 -*-
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
@@ -7,6 +8,7 @@ from collections import OrderedDict
 from errno import ENOENT
 from glob import glob
 from itertools import chain
+from logging import getLogger
 import os
 from os.path import abspath, basename, dirname, expanduser, expandvars, isdir, join
 import re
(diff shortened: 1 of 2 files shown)


mingwandroid:
@jjhelmus @msarahan please review.

conda/activate.py
```diff
@@ -546,8 +550,13 @@ def index_of_path(paths, test_path):
...
                 first_idx = 0
             else:
-                last_idx = index_of_path(path_list, prefix_dirs[-1])
-                assert last_idx is not None
+                last_idx_idx = len(prefix_dirs) - 1
+                last_idx = None
+                while last_idx is None and last_idx_idx > -1:
+                    last_idx = index_of_path(path_list, prefix_dirs[last_idx_idx])
+                    if last_idx is None:
+                        log.info("Did not find path entry {}".format(prefix_dirs[last_idx_idx]))
```
msarahan: if an entry isn't found here, should we end (``break``) the loop?  What's the difference between last_idx and last_idx_idx?  Can these be named any more descriptively?

mingwandroid: I'm not sure it matters much. It would subtly change the behaviour of a very broken scenario (which I still do not get).

mingwandroid: Ah, sorry no, this change was made specifically because some of our PATH entries were missing for some users so we do not want to break when one isn't found, we keep stepping back the entry we look for until (eventually, in a very broken setup) we hit the first entry. I'll change last_idx_idx to prefix_dir_idx if you prefer?

msarahan: sure, that's a little bit clearer

msarahan:
superseded by #8528

Issue #8505, opened 2019-04-05
DLL error breaks conda > 4.6.8


dhirschfeld:
#### `conda=4.6.10`

In my `base` environment I have `openssl=1.1.1b=hfa6e2cd_2`.
In my `dev`  environment I have `openssl=1.0.2r=hfa6e2cd_0`.
When I activate `dev` it breaks conda which gives SSL (I assume DLL load) errors from `requests`:
```
(dev) C:\> conda search openssl
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Loading channels: failed

CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/main/noarch/repodata.json.bz2>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.

If your current network has https://www.anaconda.com blocked, please file
a support request with your network engineering team.

SSLError(MaxRetryError('HTTPSConnectionPool(host=\'repo.anaconda.com\', port=443): Max retries exceeded with url: /pkgs/main/noarch/repodata.json.bz2 (Caused by SSLError("Can\'t connect to HTTPS URL because the SSL module is not available."))'))
```

This bug can be worked around by running conda from the `base` enviornment and using the `-n dev` argument to install into the `dev` environment without having it in the path

#### `conda=4.6.8`

If I install 4.6.8 then (in a new terminal) conda works again from `dev` with the mis-matched `openssl` versions:
```
(base) C:\> conda install conda=4.6.8 --quiet --yes
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: ...working... done
Solving environment: ...working... done

## Package Plan ##

  environment location: C:\Miniconda3

  added / updated specs:
    - conda=4.6.8


The following packages will be SUPERSEDED by a higher-priority channel:

  conda                      pkgs/main::conda-4.6.11-py37_0 --> conda-forge::conda-4.6.8-py37_0


Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
(base) C:\> conda activate dev
(dev) C:\> conda search openssl
(dev) C:\Windows\system32> conda search openssl
Loading channels: done
# Name                       Version           Build  Channel
openssl                       1.0.1c               0  pkgs/free
<snip>
openssl                       1.1.1b      hfa6e2cd_2  conda-forge
```

### Steps to Reproduce

[...]

PR #8516, opened 2019-04-08
Add xfail Win/PowerShell use-conda-outside-of-activation test

mingwandroid:
Note: Although it *could* be argued that the thing I'm testing here is something we don't really support, we will run into the same problem on Windows with conflicts between Python 2.7 and Python 3 and the OpenSSL module.

The fix is to add PATH to the invocation of CONDA_EXE as well as those of activate and deactivate for PowerShell, the same as we do for cmd.exe


mingwandroid:
This captures the core of the problem reported at https://github.com/conda/conda/issues/8505

PR #8517, opened 2019-04-08
Point conda.bat to condabin instead of version in Scripts for symlink_conda.

katietz:
Long term symlink_conda could be replaced by putting up conda.bat within condabin to PATH

### DIFF
--- a/conda/exports.py
+++ b/conda/exports.py
@@ -324,7 +324,7 @@ def symlink_conda(prefix, root_dir, shell=None):  # pragma: no cover
     if os.path.normcase(os.path.normpath(prefix)) in os.path.normcase(os.path.normpath(root_dir)):
         return
     if on_win:
-        where = 'Scripts'
+        where = 'condabin'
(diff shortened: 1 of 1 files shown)


mingwandroid:
Hey @katietz it's not to do with host or target, rather old conda used to have Scripts folder and new conda doesn't.

katietz:
Well, sorry my bad, I adjusted commit line for it

conda/exports.py
```diff
@@ -324,7 +324,7 @@ def symlink_conda(prefix, root_dir, shell=None):  # pragma: no cover
     if os.path.normcase(os.path.normpath(prefix)) in os.path.normcase(os.path.normpath(root_dir)):
         return
     if on_win:
-        where = 'Scripts'
+        where = 'condabin'
```
msarahan:
```suggestion
        where = 'condabin' if conda_46 else 'Scripts'
        
```

PR #8520, opened 2019-04-08
Master msys2 old activation

mingwandroid:
@katietz, @jjhelmus, @msarahan 

Nothing beyond removal of special casing that's not needed and broken.


jjhelmus:
This looks reasonable to me but would appreciate @msarahan or @katietz thoughts.

PR #8531, opened 2019-04-09
Fix conda init profile modification for powershell

### LINKED ISSUE #8530: conda init keeps modifying powershell profile
Repeated calls to conda init keep modifying profile.ps1 each time.

```
(base) wani@lenovo ~/h/conda ❯❯❯ conda init --all
...
modified      /home/wani/.config/powershell/profile.ps1
...


(base) wani@lenovo ~/h/conda ❯❯❯ conda init --all
...
modified      /home/wani/.config/powershell/profile.ps1
...
```

nehaljwani:
Fixes #8530

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -81,6 +81,11 @@ CONDA_INITIALIZE_RE_BLOCK = (
     r"([\s\S]*?)"
     r"# <<< conda initialize <<<(?:\n|\r\n)?")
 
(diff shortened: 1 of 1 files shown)


msarahan:
Am I reading this right that we just weren't factoring in newlines into the regex?  Instead of making the regex more complicated, should we instead use re.DOTALL as a flag?

I don't really care either way.  A fix is a fix.  It is nice that this new block matches the earlier block for shell stuff.

Issue #8541, opened 2019-04-10
conda env update from file does not upgrade pip packages


jpgill86:
## Current Behavior

`conda env update -f environment.yml` never upgrades existing pip packages.

This issue is likely related to #7774, but the problem described here manifests without changing pip versions, and I'm particularly interested in the use of an environment file.

### Steps to Reproduce

Here's an example that uses two environment files. The first is intended to install an old version of a package using pip. The second is intended to test the upgrade capabilities of `conda env update`.

Contents of `environment-old.yml`:

```
dependencies:
  - pip
  - pip:
    - numpy==1.15  # this version is out-of-date for purpose of example
```

Contents of `environment-new.yml`:

```
dependencies:
  - pip
  - pip:
    - numpy>=1.15  # upgrading to the latest version is expected
```

Run these commands:

```
conda env create -n test -f environment-old.yml
conda env update -n test -f environment-new.yml
```

The output looks like this:

```
(base) C:\Users\Jeffrey\Desktop>conda env create -n test -f environment-old.yml
WARNING: The conda.compat module is deprecated and will be removed in a future release.
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: done
Solving environment: done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Ran pip subprocess with arguments:
['C:\\Users\\Jeffrey\\Anaconda3\\envs\\test\\python.exe', '-m', 'pip', 'install', '-r', 'C:\\Users\\Jeffrey\\Desktop\\condaenv.39lhrqfr.requirements.txt']
Pip subprocess output:
Collecting numpy==1.15 (from -r C:\Users\Jeffrey\Desktop\condaenv.39lhrqfr.requirements.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/8b/8a/5edea6c9759b9c569542ad4da07bba0c03ffe7cfb15d8bbe59b417e99a84/numpy-1.15.0-cp37-none-win_amd64.whl
Installing collected packages: numpy
Successfully installed numpy-1.15.0

#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate


(base) C:\Users\Jeffrey\Desktop>conda env update -n test -f environment-new.yml
WARNING: The conda.compat module is deprecated and will be removed in a future release.
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: done
Solving environment: done
[...]

PR #8542, opened 2019-04-10
Add upgrade flag to pip installer

jpgill86:
This is intended as a fix for #8541.

The `-U` flag ensures that pip-managed packages are always upgraded (within the requirements) when `conda env update` is used.

Adding the flag here causes it to always be used, even for `conda env create`, but in that case the `-U` has no effect because the package is not already installed.

This strikes me as the simplest solution. If using `pip install -U` with `conda env create`, or any other time `_pip_install_via_requirements` is called, were a problem, a more complicated solution could be built in which the `-U` is added only when the call is coming from `conda env update`. However, I don't think that's necessary.

### DIFF
--- a/conda_env/installers/pip.py
+++ b/conda_env/installers/pip.py
@@ -42,7 +42,7 @@ def _pip_install_via_requirements(prefix, specs, args, *_, **kwargs):
         requirements.write('\n'.join(specs))
         requirements.close()
         # pip command line...
-        pip_cmd = ['install', '-r', requirements.name]
+        pip_cmd = ['install', '-U', '-r', requirements.name]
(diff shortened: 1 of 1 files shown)


jpgill86:
I just completed the CLA via [this form](https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement). I'm not sure what I may need to do to rerun the Conda Contributor License Agreement check.

jjhelmus:
> I'm not sure what I may need to do to rerun the Conda Contributor License Agreement check.

The CLA conformation process is manual at the moment.  @msarahan can you verify?

msarahan:
Confirmed that @jpgill86 has signed.

jjhelmus:
Thanks @jpgill86.  I agree this solution seems straightforward.

PR #8548, opened 2019-04-11
rename conda.common.os to conda.common._os to avoid conflict with bui…

msarahan:
…lt in


jjhelmus:
AppVeyor failure is due to a timeout, not failed tests, merging.

PR #8550, opened 2019-04-11
Try exiting with 0 on success from conda.bat

mingwandroid:
Otherwise people are forced to use call to prevent spurious
errorlevels from being propagated and causing failures.

### DIFF
--- a/conda.recipe/bld.bat
+++ b/conda.recipe/bld.bat
@@ -1,6 +1,15 @@
-echo %PKG_VERSION% > conda\.version
-%PYTHON% setup.py install --single-version-externally-managed --record record.txt
+copy "%RECIPE_DIR%\build.sh" .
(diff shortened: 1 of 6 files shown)


jjhelmus:
On Windows the test build of the conda package is failing with:
```
(base) C:\Miniconda36-x64\conda-bld\conda_1555018423191\test_tmp>conda.bat activate base 
2865
2866CommandNotFoundError: No command 'conda conda'.
2867
2868
2869
```

msarahan:
Sadly, I don't think you can avoid putting CALL before conda commands.  Ultimately, it's a .bat script getting run by another .bat script, and the rule there is that you must CALL.

msarahan:
@rrigdon please review the docs portion of this PR (https://github.com/conda/conda/pull/8550/files#diff-bba7262e8d0dd68611c15250f0aee1fdR10)

PR #8564, opened 2019-04-16
Fix activation not removing base env paths on win

### LINKED ISSUE #8394: Better Powershell activate and deactivate
Although conda 4.6 introduced Powershell support, I feel the implementation is still lackluster. I focus mostly on activation and deactivation of environments, which is a fundamental part of the conda workflow. 

Currently, an instance of python runs at startup of any Powershell process querying for the active environment, and its name is printed at the beginning of the prompt. This does not seem optimal because it slows down the loading of the shell even if the user is not interested in actually working with conda. 

The implementation in cmd seems to be better because conda runs only when actively requested by the user (via `conda activate`) and information about the environment disappears when no longer needed (`deactivate`). Shouldn't it behave in the same way in Powershell?

### LINKED ISSUE #8566: Path exploding on Windows due to conda init
## Current Behavior

Every time I issue a conda command in powershell, the windows path gets extended with duplicate entries... eventutually becoming too long to set.

**BUG FIX**
Note the line below:

```
Write-Output Env:PATH is $Env:PATH;
```
It's a debug output that was never removed that's being added to the PATH and thus the system keeps adding to the PATH over and over and over...

### Steps to Reproduce
Do conda init powershell and then using commmands the path keeps growing. It appears to be happening in Conda.psm1:

```
<#
    .SYNOPSIS
        Adds the entries of sys.prefix to PATH and returns the old PATH.

    .EXAMPLE
        $OldPath = Add-Sys-Prefix-To-Path
#>
function Add-Sys-Prefix-To-Path() {
    $OldPath = $Env:PATH;
    if ($Env:_CE_CONDA -ne '' -And $Env:OS -eq 'Windows_NT') {
        # Windows has a different layout for the python exe than other platforms.
        $sysp = Split-Path $Env:CONDA_EXE -Parent;
    } else {
        $sysp = Split-Path $Env:CONDA_EXE -Parent;
        $sysp = Split-Path $sysp -Parent;
    }
    if ($Env:OS -eq 'Windows_NT') {
        $Env:PATH = $sysp + ';' +
                    $sysp + '\Library\mingw-w64\bin;' +
                    $sysp + '\Library\usr\bin;' +
                    $sysp + '\Library\bin;' +
                    $sysp + '\Scripts;' +
                    $sysp + '\bin;' + $Env:PATH;
    } else {
        $Env:PATH = $sysp + '/bin:' + $Env:PATH;
    }
    Write-Output Env:PATH is $Env:PATH;
    return $OldPath;
}

```
## Expected Behavior
Path should not lengthen with duplicates

msarahan:
fixes #8394 
fixes #8566


msarahan:
anaconda.org was out for dinner.  Cycling to re-run.

PR #8570, opened 2019-04-17
fix missing powershell function export

msarahan:
Reported by QA.  This function shows up in an error report when Anaconda Powershell Prompt is opened.


msarahan:
Please consider submitting a separate issue and/or a PR. We are definitely not powershell experts, so any help and improvement are much appreciated.

PR #8578, opened 2019-04-19
WIP: strict channel priority by default

msarahan:
Strict channel priority solves a number of problems that the user community is experiencing.  It can dramatically reduce the solve times, as well as helping prevent creation of bad environments due to bad dependency metadata on old packages.

This PR is to work out the test changes that are necessary from this change.  This change also requires improvement of the UnsatisfiableError reporting.  @jjhelmus is spearheading that separately.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -199,7 +199,7 @@ class Context(Configuration):
     _channel_alias = PrimitiveParameter(DEFAULT_CHANNEL_ALIAS,
                                         aliases=('channel_alias',),
                                         validation=channel_alias_validation)
-    channel_priority = PrimitiveParameter(ChannelPriority.FLEXIBLE)
+    channel_priority = PrimitiveParameter(ChannelPriority.STRICT)
     _channels = SequenceParameter(string_types, default=(DEFAULTS_CHANNEL_NAME,),
                                   aliases=('channels', 'channel',),
                                   expandvars=True)  # channel for args.channel
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -343,7 +343,6 @@ class Resolve(object):
             top_level_sdeps = {top_level_spec.name: set(top_level_pkgs)}
 
             # find all depends specs for in top level packages
-            # find the depends names for each top level packages
             second_level_specs = set()
             top_level_pkg_dep_names = []
             for pkg in top_level_pkgs:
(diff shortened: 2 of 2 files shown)


msarahan:
There's a pretty major wrinkle with this.  When strict channel priority is on, and you give conda an explicit spec that exists only in some lower-priority channel, that lower-priority channel can be hidden completely by strict channel priority.  This breaks a ton of conda's test suite, because a lot of it was written with reference to the free channel, which was current at that time.  Now the main channel is our top priority.  When a test uses ``python=3.5.2`` for example, that test is now failing because 3.5.2 is unsatisfiable.  Strict channel priority has eliminated it from consideration, because the main channel has other python packages.

What this means is that "frozen" env files might be very sensitive to this change.  They might suddenly break with no change on the user's side other than updating conda.

This change may need to force a major version bump.

h-vetinari:
@msarahan 
Has it been considered to be "strict" based on package _version_ and not on just the package?
That way, if main does not have `python=3.5.2`, the package would still be found.

This seems to me to be the "obvious" choice for how strict channel priority should work (actually, down to the build versions etc. if specified in a lock file, for example) - what are the reasons that it should only be per package (name)?

msarahan:
I did experiment with that.  I *think* it might be OK if you go by version if and only if the channel is also explicitly provided.  If you fall back to other channels when a version is not available, that is otherwise a violation of what strict channel priority means, and definitely not desirable.

h-vetinari:
> @msarahan: I did experiment with that. I _think_ it might be OK if you go by version if and only if the channel is also explicitly provided. If you fall back to other channels when a version is not available, that is otherwise a violation of what strict channel priority means, and definitely not desirable.

Weeeeeeell, I get that channel priority is supposed to give you packages built with consistent compiler chains etc., so being strict about that is good. OTOH, if a user is explicitly asking for a version that is not available in a high-priority channel, then the resolver should descend the channels IMO (possibly with a warning). That very same package might for example be python-only, and have no build issues whatsoever.

What this suggestion boils down to: be as strict as possible, but allow resolution for version pins (resp. ranges) that are only available for lower-priority channel.

msarahan:
Sorry, but I disagree.  Having conda error out when constrained versions are not available has been helpful to understand when conda-forge or defaults needs to update their packages.  The fall-through behavior was definitely harmful.  How many people need this behavior to work in order to compensate for the people who get stuck with subtle binary compatibility bugs?

Messaging is key here, and that's why we have #8638 as a key requirement for having strict channel priority by default.

h-vetinari:
@msarahan 
Fair enough, that's a very valid reason, of course...

mbargull:
> This change may need to force a major version bump.

+1

travishathaway:

PR #8579, opened 2019-04-19
remove free from defaults.  Provide restore_free_channel config

msarahan:


nehaljwani:
The CB 3.10 test is fails when `test_no_force_upload` runs followed by `test_conda_py_no_period`. Individually, both pass.

mingwandroid:
Every time I've ran into this is because a previous test messes up the global state. 

Look for things that poke context or os.environ directly. Use with env_vars etc instead.

mingwandroid:
Future Nehal with broken arm says 'huh?'

nehaljwani:
This is problem: https://github.com/conda/conda-build/blob/3.10.1/tests/test_cli.py#L486

msarahan:
We can't retroactively fix the cb test suite. The best use of time is
getting Ray's cb PR in, since he fixed these things. We need not block this
PR for this one test, since we understand it, but we should update conda's
use of conda builds test suite.

mbargull:
Hm, I think adding a separate `restore_free_channel` config param creates more maintenance burden than just adding an additional `defaults_legacy_foobar` "mulitchannel" such that one could just do `CONDA_DEFAULT_CHANNELS=defaults_legacy_foobar conda create -nx pkg=ancient_version`.
(Though, AFAIR, referencing a multichannels from another multichannel hasn't been fixed the last time I checked.)

Issue #8584, opened 2019-04-22
Conda init tcsh does install the right script


petrosb:
## Current Behavior

When I type conda init tcsh (or conda init --all) conda mentions it updates my .tcshrc file. However, when I open a tcsh shell, I get an "Illegal variable name" error, and the path is not updated. I believe the issue is that the .tcshrc created is not actually tcsh compatible. For example, it includes a redirect to /dev/null/ of the form "2>" which might not be an acceptable tcsh command (see here, https://unix.stackexchange.com/questions/197878/inconsistency-of-stderr-redirection-between-tcsh-and-other-shells/197880). Also uses the $(...) syntax, which also seems to be an illegal tcsh command (see here, https://stackoverflow.com/questions/43725152/illegal-variable-name-tcsh-shell-linux). Also see below for a dump of the .tcshrc that is created. Another explanation is that for some reason the command installs an incorrect .tcshrc, intended for some other shell.
Current solution is to delete the .tcshrc, but then "conda activate base" does not work.

### Steps to Reproduce

```
On a Mac:

>>conda init tcsh
no change     /Users/petrosb/anaconda3/condabin/conda
no change     /Users/petrosb/anaconda3/bin/conda
no change     /Users/petrosb/anaconda3/bin/conda-env
no change     /Users/petrosb/anaconda3/bin/activate
no change     /Users/petrosb/anaconda3/bin/deactivate
no change     /Users/petrosb/anaconda3/etc/profile.d/conda.sh
no change     /Users/petrosb/anaconda3/etc/fish/conf.d/conda.fish
no change     /Users/petrosb/anaconda3/shell/condabin/Conda.psm1
no change     /Users/petrosb/anaconda3/shell/condabin/conda-hook.ps1
no change     /Users/petrosb/anaconda3/lib/python3.6/site-packages/xonsh/conda.xsh
no change     /Users/petrosb/anaconda3/etc/profile.d/conda.csh
modified      /Users/petrosb/.tcshrc

==> For changes to take effect, close and re-open your current shell. <==

>> reopening a new shell:
Last login: Mon Apr 22 12:42:13 on ttys005
Illegal variable name.
[ap8w39:~] petrosb% which conda
conda: Command not found.
[ap8w39:~] petrosb% which python
/usr/local/bin/python
[ap8w39:~] petrosb% more .tcshrc                                                
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
[...]

PR #8591, opened 2019-04-23
Make initalization block for tcsh and csh have correct syntax

soapy1:
addresses: https://github.com/conda/conda/issues/8584

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1081,27 +1081,51 @@ def _bashrc_content(conda_prefix, shell):
         }
     else:
         conda_exe = join(conda_prefix, 'bin', 'conda')
-        conda_initialize_content = dals("""
-        # >>> conda initialize >>>
-        # !! Contents within this block are managed by 'conda init' !!
-        __conda_setup="$('%(conda_exe)s' 'shell.%(shell)s' 'hook' 2> /dev/null)"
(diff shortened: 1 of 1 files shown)


jjhelmus:
LGTM, thanks @soapy1

conda/core/initialize.py
```diff
@@ -1081,27 +1081,51 @@ def _bashrc_content(conda_prefix, shell):
...
+            # >>> conda initialize >>>
+            # !! Contents within this block are managed by 'conda init' !!
+            set __conda_setup="`%(conda_exe)s shell.%(shell)s hook >& /dev/null`"
+            if ( $? == 0 ) then
+                eval "$__conda_setup"
+            else
+                if ( -f "%(conda_prefix)s/etc/profile.d/conda.sh" ) then
+                    . "%(conda_prefix)s/etc/profile.d/conda.sh"
+                else
+                    export PATH="%(conda_bin)s:$PATH"
```
olaurino: This would also fail because `export` is not csh, although I am not sure whether, although I am not sure whether this 
 `else` branch is ever executed.

soapy1: ohh, good catch. Should this be `setenv` instead? 
I think this bit will pretty much never get executed. I believe it's there in case users downgrade conda and the newer activation bits are removed.

PR #8248, opened 2019-02-11
resolve #8176 SAT solver choice error handling

kalefranz:
resolve #8176

PR #8602, opened 2019-04-26
review gh-8248 sat solver choice

mbargull:
This is a review for gh-8248.

- moves the SAT solver choice from `conda.common.logic` into `conda.resolve` to avoid importing from `conda.base` into `conda.common.
- adds the choices for enum-based configuration parameters to error messages and replaces `auxlib`'s `TypeCoercionError` with a `CustomValidationError`. With that change we get the following instead of nasty tracebacks:
```
$ conda config --set sat_solver asdf

CustomValidationError: Parameter sat_solver = 'asdf' declared in --set parameter is invalid.
'asdf' is not a valid SatSolverChoice
Valid choices for sat_solver: 'pycosat', 'pycryptosat', 'pysat'

$ CONDA_SAT_SOLVER=asdf conda config --show sat_solver

CustomValidationError: Parameter sat_solver = 'asdf' declared in <<merged>> is invalid.
'asdf' is not a valid SatSolverChoice
Valid choices for sat_solver: 'pycosat', 'pycryptosat', 'pysat'
```

### DIFF
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -310,7 +310,7 @@ def execute_config(args, parser):
     for key, item in args.set:
         key, subkey = key.split('.', 1) if '.' in key else (key, None)
         if key in primitive_parameters:
-            value = context.typify_parameter(key, item)
+            value = context.typify_parameter(key, item, "--set parameter")
             rc_config[key] = value
         elif key in map_parameters:
             argmap = rc_config.setdefault(key, {})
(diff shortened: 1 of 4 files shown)


mbargull:
gh-8248 also changed it so that if the explicitly requested SAT solver interface can not be used, it tries to fallback to the others. The original implementation from gh-7811 would fail in that case to try to be verbose about not being able to run as requested. I did **not** restore the behavior from gh-7811 in this PR, but would like to hear someone else's opinion on it.

mbargull:
> I'm weakly in favor of the fallback

I'm weakly in favor of the slap-in-the-face variant :wink:. But I'm fine with either. Just wanted to make sure both options were considered by a third pair of eyes. Thanks for your input!

PR #8604, opened 2019-04-26
WIP PoC of what sharded repodata might look like

kalefranz:
This should really just be viewed as a guide on what might need to change if conda were to use repodata sharded by package name.  It's not complete and still has multiple problems, especially switching between sharded channels and non-sharded channels.


msarahan:
I think this is superseded by #8651.  I'm closing it, but happy to reopen if I'm wrong.

kalefranz:
Yup #8651 is more current than this PR.

PR #8605, opened 2019-04-26
move token to header instead of in URL

kalefranz:
Actually meant to make this PR against my own master fork.  But now that it's here I guess I'll leave it.  Sorry about that.

This PR moves the token that conda submits to a repo server (when it finds a token) from embedded in a url path to an HTTP header.  Support for this has been built into anaconda.org for a long time.  Putting auth information in a url path is very non-standard, and this PR future-proofs conda for when a future repo might no longer accept that.  Again, this PR is fully compatible now with anaconda.org.

### DIFF
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -96,7 +96,9 @@ class CondaHttpAuth(AuthBase):
     # TODO: make this class thread-safe by adding some of the requests.auth.HTTPDigestAuth() code
 
     def __call__(self, request):
-        request.url = CondaHttpAuth.add_binstar_token(request.url)
+        request.url, token = CondaHttpAuth.add_binstar_token(request.url)
+        if token is not None:
+            request.headers["Authorization"] = "Bearer {0}".format(token)
         self._apply_basic_auth(request)
         request.register_hook('response', self.handle_407)
         return request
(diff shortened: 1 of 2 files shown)


conda/gateways/connection/session.py
```diff
@@ -96,7 +96,9 @@ class CondaHttpAuth(AuthBase):
     # TODO: make this class thread-safe by adding some of the requests.auth.HTTPDigestAuth() code
 
     def __call__(self, request):
-        request.url = CondaHttpAuth.add_binstar_token(request.url)
+        request.url, token = CondaHttpAuth.add_binstar_token(request.url)
+        if token is not None:
+            request.headers["Authorization"] = "Bearer {0}".format(token)
```
soapy1: Is there a reason for having the token prepended with "Bearer"? It seems not super necessary to me

kalefranz:
That’s very much a standard for a “bearer” type token, which is what this is. If you have the token, you get access. It’s also the format anaconda.org accepts. Google for “bearer token” and you’ll find all sorts of references.

conda/gateways/connection/session.py
soapy1: https://jwt.io/introduction/ this article explains that pattern pretty well

kalefranz:
We can’t do this yet. Closing for now.

mcg1969:
@kalefranz would you mind sharing details of why "we can't do this yet"?
cc: @fycyang
It's coming up with a customer discussion

PR #8620, opened 2019-04-30
fast frozen solve

msarahan:
when running conda install, we should try to keep all other deps (those not explicitly specified) frozen.  This should be very fast, relative to having them considered in the solve.

This behavior already existed with the ``--freeze-installed`` or ``--no-update-deps`` CLI arguments, but this puts that behavior as the initial default, with a fallback to the old, full solve if that returns an unsatisfiable result.

### DIFF
--- a/conda/common/compat.py
+++ b/conda/common/compat.py
@@ -11,6 +11,8 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 from itertools import chain
 from operator import methodcaller
 import sys
+from tempfile import mkdtemp
+import warnings as _warnings
 
 on_win = bool(sys.platform == "win32")
 on_mac = bool(sys.platform == "darwin")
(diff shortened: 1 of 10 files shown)


msarahan:
@jjhelmus tested this on an old, somewhat complicated env with a fair number of already installed packages and some more complicated stuff, like dev installs.  The time to install "imagesize" dropped from 90 sec to 30 sec.

conda/common/compat.py
```diff
@@ -294,3 +296,46 @@ def ensure_utf8_encoding(value):
...
+        self.name = mkdtemp(suffix, prefix, dir)
+
+    def __repr__(self):
+        return "<{} {!r}>".format(self.__class__.__name__, self.name)
+
+    def __enter__(self):
+        return self.name
+
+    def cleanup(self, _warn=False, _warnings=_warnings):
+        from ..gateways.disk.delete import rm_rf as _rm_rf
```
mbargull: I think it makes sense to adhere to https://github.com/conda/conda/blob/93db2c14f7b49bce0a00d80a0463470f8a797d50/conda/common/README.rst and not import from outside `conda.common` here to ease later refactoring (e.g., in the advent of a new `libconda`).
Thus, I think it's better to move `TemporaryDirectory` to `conda.gateways.disk` (and use `tempfile.TemporaryDirectory` for Python >=3.2).

msarahan: Thanks, that's good feedback.

PR #8621, opened 2019-05-01
Change developer warnings to use warnings module, for better filtering

epruesse:
The `WARNING` is turning up in client code running `conda info` via the shell, creating the illusion that something is broken in the way that conda is being employed. Using the warning module would be an option, but still require all parts of `conda` itself to catch that warning.

### DIFF
--- a/conda/compat.py
+++ b/conda/compat.py
@@ -65,7 +65,3 @@ class TemporaryDirectory(object):
     def __del__(self):
         # Issue a ResourceWarning if implicit cleanup needed
         self.cleanup(_warn=True)
-
-
-print("WARNING: The conda.compat module is deprecated and will be removed in a future release.",
-      file=sys.stderr)


msarahan:
I think the goal originally was to get rid of that particular module over time.  It seems pretty close to gone for internal usage, but there was some residual stuff.  Keeping the warning in for external usage is a good thing, in my opinion.

https://github.com/conda/conda/pull/8620/commits/e8a29a72db0ab60036195b3c215d9909c2b5f1f2

does some work to clean up one remaining vestige of usage in conda.  Hopefully this is the last of it, but if not, we'll work to continue removing internal usage of this particular module.

Closing, as I don't think this is the right fix, but thank you for your concern here.

msarahan:
PS: that commit is part of #8620

epruesse:
Suggestion for an actual fix:
Use [`warnings.warn(message[, category[, stacklevel]])`](https://docs.python.org/2/library/warnings.html#warnings.warn):

```python
_warnings.warn("The conda.compat module will be removed in a future release.", 
               DeprecationWarning, stacklevel=2)
```

`stacklevel=2` will make the warning appear originating from where the module is imported, so from where the error actually is. 

You then use facilities like then use `warnings.filterwarning('ignore')` or `with warnings.catch_warnings()` or `logging.captureWarnings(true)` to keep things meant for devs from confusing users. A simple way is to just check if `conda` was installed editable (`pip install -e`) and if not, disable warnings wholesale from the main cli code. That way, things calling `conda` don't get spam, while external API callers get a nice deprecation warning (and can silence it for their users). 

TBH, I was expecting you to close the PR. I can't implement the above suggestion for you because it'd take me weeks to get permission to sign the CLA.

msarahan:
That sounds wise.  I'm going to reopen the issue and change the title.

travishathaway:

PR #8625, opened 2019-05-05
Fix path_to_url "string index out of range" error

yuvalreches:


mingwandroid:
Can you explain why it is valid to pass a PATH of length 0?

yuvalreches:
> Can you explain why it is valid to pass a PATH of length 0?

@mingwandroid it is actually not possible :)
The scenario we are facing is having the path to be `/`, length 1, then we hit `IndexError: string index out of range`
This is obviously as `path[1]` doesn't exist, only `path[0]` does exist.

msarahan:
@yuvalreches has signed the CLA.  That should not block this PR.  The Py27 fail is my fault from another PR.  This PR is green.

msarahan:
Thanks @yuvalreches

PR #8639, opened 2019-05-07
separate conda package metadata under new root-level key in repodata

kalefranz:
This PR denormalizes the overlaying/entangled metadata in the original .conda package format work (i.e. `conda_size`, `conda_sha256`, `conda_inner_sha256`, and `conda_outer_sha256` fields).  With this PR, `repodata.json` should now add a new root-level `packages.conda` field containing information for all packages with the `.conda` extension.

This strategy has several advantages, including:

1. All existing versions of conda will simply not see the new packages.
2. New packages are decoupled from old packages, and so, in cases where someone only uploads the new format to a channel skipping the old format, current versions of conda won't crash.
3. Having a metadata entry for each package makes clear exactly which package the metadata refers to and where the source of truth is.  That's especially useful for an dynamic anaconda.org repo situation where users can copy, move, and delete individual packages from channels.

The root key `packages.conda` was chosen over `conda_packages` so that sorting `repodata.json` keeps the `info` key as the top key in the sort.

An example of what repodata.json now looks like is included in this PR as `tests/data/conda_format_repo/win-64/repodata.json`.

```
{
  "info": {
    "subdir": "win-64"
  },
  "packages": {
    "zlib-1.2.11-h62dcd97_3.tar.bz2": {
      "build": "h62dcd97_3",
      "build_number": 3,
      "depends": [
        "vc >=14.1,<15.0a0"
      ],
      "license": "zlib",
      "license_family": "Other",
      "md5": "a46cf10ba0eece37dffcec2d45a1f4ec",
      "name": "zlib",
      "sha256": "10363f6c023d7fb3d11fdb4cc8de59b5ad5c6affdf960210dd95a252a3fced2b",
      "size": 131285,
      "subdir": "win-64",
      "timestamp": 1542815182812,
      "version": "1.2.11"
    }
  },
  "packages.conda": {
    "zlib-1.2.11-h62dcd97_3.conda": {
      "build": "h62dcd97_3",
      "build_number": 3,
      "depends": [
        "vc >=14.1,<15.0a0"
      ],
      "license": "zlib",
      "license_family": "Other",
      "md5": "edad165fc3d25636d4f0a61c42873fbc",
      "name": "zlib",
      "sha256": "2fb5900c4a2ca7e0f509ebc344b3508815d7647c86cfb6721a1690365222e55a",
      "size": 112305,
      "subdir": "win-64",
      "timestamp": 1542815182812,
      "version": "1.2.11"
    }
  },
  "removed": [],
  "repodata_version": 1
}
```


msarahan:
This PR does not currently specify or demonstrate how new packages should look - what the actual data in repodata.json should be.  We need that in tests.  @kalefranz are you going to have time to develop that, or should we take that on?

The more hacky way that I had done things is very much still in the codebase, and this PR should probably remove it.  It's not necessary to keep it because this is a better way that will require much less hacky stuff, but we need the tests for usage of the new file format to reflect this new way of doing things.

kalefranz:
I've updated this PR with a second commit.  In addition to functional code changes, it exercises all the package cache code pretty thoroughly with additions to `tests/core/test_package_cache_data.py`.

I need another hour or so on this though.  The one remaining item on my TODO list here is to audit all uses of `CONDA_TARBALL_EXTENSION` to make sure they are still appropriate.

kalefranz:
OK, pending test success, I'm happy with the state of this PR and welcome review and feedback.

conda/models/match_spec.py
```diff
@@ -177,8 +178,8 @@ def __init__(self, optional=False, target=None, **kwargs):
     @classmethod
     def from_dist_str(cls, dist_str):
         parts = {}
-        if dist_str.endswith(CONDA_TARBALL_EXTENSION):
-            dist_str = dist_str[:-len(CONDA_TARBALL_EXTENSION)]
+        if dist_str.endswith(CONDA_PACKAGE_EXTENSION_V1):
+            dist_str = dist_str[:-len(CONDA_PACKAGE_EXTENSION_V1)]
```
kalefranz: Should only be used in deprecated code paths.  `CONDA_PACKAGE_EXTENSION_V1` should be fine here.

conda/models/match_spec.py
```diff
@@ -258,7 +259,7 @@ def _to_filename_do_not_use(self):
             return fn_field
         vals = tuple(self.get_exact_value(x) for x in ('name', 'version', 'build'))
         if not any(x is None for x in vals):
-            return '%s-%s-%s.tar.bz2' % vals
+            return ('%s-%s-%s' % vals) + CONDA_PACKAGE_EXTENSION_V1
```
kalefranz: The name of this method is `_to_filename_do_not_use()`.

conda/models/match_spec.py
```diff
@@ -570,7 +570,7 @@ def _parse_spec_str(spec_str):
     spec_str = spec_split[0]
 
     # Step 2. done if spec_str is a tarball
-    if spec_str.endswith(CONDA_TARBALL_EXTENSION):
+    if spec_str.endswith(CONDA_PACKAGE_EXTENSION_V1):
```
kalefranz: This needs to be `CONDA_PACKAGE_EXTENSIONS`.  Need to add a test.

kalefranz:
The failing py27 test is `tests/test_api_build.py::test_warning_on_file_clobbering`, and it seems to be failing on other branches as well.

The failing conda-build test is `tests/test_api_build.py::test_warning_on_file_clobbering`.  @msarahan do you have any insight into what got out of sync here?

conda/core/subdir_data.py
```diff
@@ -354,7 +354,17 @@ def _process_raw_repodata_str(self, raw_repodata_str):
...
+        conda_packages = json_obj.get("packages.conda", {})
+        _tar_bz2 = CONDA_PACKAGE_EXTENSION_V1
+        use_these_legacy_keys = set(iterkeys(legacy_packages)) - set(
+            k[:-6] + _tar_bz2 for k in iterkeys(conda_packages)
+        )
+
+        for fn, info in concatv(
+            iteritems(conda_packages),
+            ((k, legacy_packages[k]) for k in use_these_legacy_keys),
+        ):
```
kalefranz: @msarahan if we want to add a feature flag to opt-in to using the new package format (which we can switch to `true` by default in a later conda release) then the only change we have to make to do that is localized right here.

conda/core/package_cache_data.py
```diff
@@ -275,17 +260,15 @@ def _check_writable(self):
         return i_wri
 
     @staticmethod
-    def _clean_tarball_path_and_get_checksums(tarball_path, sha256sum=None, md5sum=None):
+    def _clean_tarball_path_and_get_md5sum(tarball_path, md5sum=None):
```
msarahan: It looks like you are reverting a lot of the work I did to prefer sha256 over md5 where possible.  Is that intentional?  What's your reasoning on that?

kalefranz: The purpose of using md5 in all of these places is only for quick identification, and to have reasonable assurance that we have the "right" artifact, in the context of distinguishing if the package `flask-1.0.2-py_0.tar.bz2` in the package cache came from `free`, `main`, `conda-forge`, or some other channel.  We know pretty close to what it should be by the file name, but because we don't include channel information with the bare tarballs in the package cache, we don't know what channel they come from.  It's also a quick check to ensure that a package hasn't been updated in-place in the remote channel (and therefore needs to be downloaded again).

Matching the md5 to the expected md5 from repodata gives us the assurance that an artifact that has the right name in a package cache is actually the artifact that we want.  To be clear: **Using md5 here is not a security measure.**  Just as we use md5 as a checksum to ensure downloads have not been _inadvertently_ corrupted, we use md5 here for a fast check to make sure the artifact is as we expect (or not).

Neither md5 nor sha256 in this context is sufficient as a security measure to guard against malicious attacks.  The purpose of using md5 here is ONLY as a quick checksum.  And in these quick checksum cases we should prefer md5 to sha256 since the former is indeed quite a bit faster.

When we add cryptographic signatures, we will be making specific and more targeted use of sha256 + size, in combination.

msarahan: Thanks.  That makes sense.  Is it worth keeping at least for the file downloads, though?

kalefranz: Yeah, I don't see any problem with that.  And while we're at it, we might as well add in size to that function too and get rid of one of those TODOs.  Pushing another commit soon with the change.

conda/gateways/connection/download.py
```diff
@@ -89,11 +92,22 @@ def download(url, target_full_path, md5sum=None, progress_update_callback=None):
                 log.debug("%s, trying again" % e)
             raise
 
-        actual_md5sum = digest_builder.hexdigest()
-        if md5sum and actual_md5sum != md5sum:
-            log.debug("MD5 sums mismatch for download: %s (%s != %s), "
-                      "trying again" % (url, digest_builder.hexdigest(), md5sum))
-            raise MD5MismatchError(url, target_full_path, md5sum, actual_md5sum)
+        if md5:
```
msarahan: Is it worthwhile to compute both sha256 and md5? My original thought was to only compute the most secure option available. I guess also computing md5 won't add much time, though.

kalefranz: Using sha256 instead of md5 isn't really a valid security measure.  You need to couple sha256 with cryptographic signatures to actually have "security" in this context.  Without signatures, the best assurance we can have here is that the artifact wasn't inadvertently corrupted by erroneous HTTP over the wire.  It's false to think that md5 is any more "secure" in this use case than sha256.  (Probably why s3 uses md5 for etags, as I mentioned in https://github.com/conda/conda/pull/8651.)  So while it's not any more secure, you also incur the performance penalty of the heavier sha256.

All that said, we'll probably need to start using sha256 and incorporating that into any upcoming package signing implementation, so while we're really not adding any security now, it makes sense to prepare for the future.  I'll make the change.

msarahan:
The 2 test failures are confusing to me.  They're reproducible locally, but only when running that test file as a whole.  If those tests are run individually, or run in a subset (``--lf`` after a failing run), then they pass.  Clearly something is more stateful than it should be.  I have verified that they are not sharing a package cache dir (the test folder paths are very different).  If anyone has any ideas for what else to check, please chime in.

conda/gateways/connection/download.py
mbargull: > Is it worthwhile to compute both sha256 and md5?

Computing both is wasteful because computing SHA256 usually isn't `x` times but rather `0.x` times slower than computing MD5. On my machine it's roughly 50 %.

> Using sha256 instead of md5 isn't really a valid security measure.

Sure.

> So while it's not any more secure, you also incur the performance penalty of the heavier sha256.
> \[...\]
> All that said, we'll probably need to  start using sha256 and incorporating that into any upcoming package signing implementation

For performance reasons you might want to consider SHA512 or SHA512/256 (supported from `openssl 1.1.1` onwards, available in Python via `cryptography >=2.5`). For me ( = not representative, of course), computing SHA512/256 is nearly as performant as MD5:
```
> podman run --rm -it continuumio/miniconda3 sh -lic '                                                                                                                                                                                                                      
conda create -qynopenssl openssl >/dev/null && conda activate openssl
openssl version
for hash in md5 sha256 sha512 sha512-256 ; do openssl speed -seconds 2 -bytes 1048576 -evp "${hash}" 2>/dev/null | tail -1 ; done
'
OpenSSL 1.1.1b  26 Feb 2019
md5             790102.02k
sha256          515857.24k
sha512          767557.63k
sha512-256      770179.07k
```

msarahan: cool!  thanks for the pointer.

Issue #8590, opened 2019-04-23
Fish init does not work


RSully:
## Problem

I previously had Conda setup in fish by simply using:

```
source $CONDA_PATH/etc/fish/conf.d/conda.fish
```

Recently I found out about `conda init` and ran `conda init fish` which put the following in my fish config instead (I removed the above `source`):

```
eval (eval $CONDA_PATH/bin/conda "shell.fish" "hook" $argv)
```

When using the first method, `$CONDA_PATH/condabin/conda` is added to my PATH, which seems wrong. Using the second method, `conda activate` does not work at all, and I get an error that I should run `conda init`.

## Environment Information
<details open><summary><code>`conda info` (using first method)</code></summary><p>

```
     active environment : None
            shell level : 0
       user config file : ~/.condarc
 populated config files : ~/.condarc
          conda version : 4.6.14
    conda-build version : 3.17.8
         python version : 3.6.8.final.0
       base environment : ~/dev/opt/miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/osx-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://conda.anaconda.org/conda-forge/osx-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : ~/dev/opt/miniconda3/pkgs
                          ~/.conda/pkgs
       envs directories : ~/dev/opt/miniconda3/envs
                          ~/.conda/envs
               platform : osx-64
             user-agent : conda/4.6.14 requests/2.21.0 CPython/3.6.8 Darwin/18.5.0 OSX/10.14.4
                UID:GID : 501:20
             netrc file : ~/.netrc
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```
==> ~/.condarc <==
auto_update_conda: False
channel_priority: flexible
channels:
  - defaults
  - conda-forge
show_channel_urls: True
```
</p></details>

RSully:
The documentation for fish is also inconsistent in this regard:

[...]

PR #8645, opened 2019-05-08
Fix fish activation

soapy1:
fixes https://github.com/conda/conda/issues/8590

According to this issue:
https://github.com/fish-shell/fish-shell/issues/3993

you can not just eval a script in fish. The proper way to do this
is apparently to pipe the output to source.

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -973,7 +973,7 @@ def _config_fish_content(conda_prefix):
     conda_initialize_content = dals("""
     # >>> conda initialize >>>
     # !! Contents within this block are managed by 'conda init' !!
-    eval (eval %(conda_exe)s "shell.fish" "hook" $argv)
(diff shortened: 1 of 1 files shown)


RSully:
The output of `conda shell.fish hook` includes:

```fish
function conda --inherit-variable CONDA_EXE
    if [ (count $argv) -lt 1 ]
        eval $CONDA_EXE
    else
        set -l cmd $argv[1]
        set -e argv[1]
        switch $cmd
            case activate deactivate
                eval (eval $CONDA_EXE shell.fish $cmd $argv)
            case install update upgrade remove uninstall
                eval $CONDA_EXE $cmd $argv
                and eval (eval $CONDA_EXE shell.fish reactivate)
            case '*'
                eval $CONDA_EXE $cmd $argv
        end
    end
end
```

Does this need to be changed too to avoid the `eval (eval ...)`s?

PR #8651, opened 2019-05-10
sharded repodata

kalefranz:
This PR implements a strategy for repodata sharding. It assumes two new file types.

The first is a file type `<name>.json` that sits in each subdir alongside `repodata.json`, where `<name>` is either the package `name` or a `track_features` value.  The package metadata contained within `<name>.json` is that associated with `<name>`, and the format is otherwise similar to `repodata.json`.

The second is a `lastmodified.json` file that sits in each subdir alongside `repodata.json`.  The format of `lastmodifed.json` is

```
{
  "lastmodified_version": 1,
  "packages": {
    "<name>": {
      "etag": "<md5>",
      "mod":  <integer_unix_timestamp>
    },
    ...
  }
}
```

In the above, `mod` is a unix timestamp that is converted to a compliant `If-Modified-Since` header by `datetime.utcfromtimestamp(integer_unix_timestamp).strftime("%a, %d %b %Y %H:%M:%S GMT")`.  It is also optional.  In general `etag` should be preferred over `mod`.  For s3-hosted repositories, the standard etag is the md5 hash of the binary content of the object, in this case the md5 hash of the corresponding `<name>.json` file.  While the md5 has is the standard for s3 buckets, the etag need not be the md5 hash in general.  Another server implementation, for example, may choose to use the first eight characters of the sha256 hash. The etag value in `lastmodified.json` should *not* include escaped quotation marks. Conda will add quotation marks to the `If-None-Match` header as appropriate.

This strategy should be straight-forward to implement with `conda index` for statically-hosted channels.

This PR is currently built on top of #8639.  If/when that PR is merged, the diff for this PR will be substantially smaller.  The bulk of the code change for this PR is isolated to `conda/core/subdir_data.py`.

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -195,6 +195,7 @@ class ResponseException(Exception):
         super(ResponseException, self).__init__()
         self.url = url
         self.response_code = response_code
+        self.response = response
 
 
 def load_url_object(cache_path_base, url, channel, cache_cls):
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -25,7 +25,7 @@ from .._vendor.toolz import concat, concatv, take
 from ..base.constants import CONDA_HOMEPAGE_URL, CONDA_PACKAGE_EXTENSION_V1
 from ..base.context import context
 from ..common.compat import (ensure_binary, ensure_text_type, ensure_unicode, iteritems, iterkeys,
-                             string_types, text_type, with_metaclass)
+                             open, string_types, text_type, with_metaclass)
 from ..common.io import ThreadLimitedThreadPoolExecutor, as_completed
 from ..common.url import join_url, maybe_unquote
 from ..core.package_cache_data import PackageCacheData
@@ -49,7 +49,7 @@ except ImportError:  # pragma: no cover
 log = getLogger(__name__)
 stderrlog = getLogger('conda.stderrlog')
 
-REPODATA_PICKLE_VERSION = 28
+INTERNAL_PICKLE_VERSION = 35
 MAX_REPODATA_VERSION = 1
 REPODATA_HEADER_RE = b'"(_etag|_mod|_cache_control)":[ ]?"(.*?[^\\\\])"[,\}\s]'  # NOQA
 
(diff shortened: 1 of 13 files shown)


conda/core/subdir_data.py
```diff
@@ -126,39 +124,59 @@ def __init__(self, channel):
...
+            # 2. If that fails, try repodata.json
+            try:
+                cached_repodata = load_url_object(
+                    self.cache_dir,
+                    join_url(self.url_w_credentials, "repodata.json"),
+                    self.channel,
+                    CachedRepodata
+                )
+            except ResponseException as e:
+                # TODO: Add in litany of error messages here
```
kalefranz: Big TODO here.  All the error messages need to go back in.

conda/core/subdir_data.py
```diff
@@ -126,39 +124,59 @@ def __init__(self, channel):
...
-
     def load(self):
-        _internal_state = self._load()
-        if _internal_state.get("repodata_version", 0) > MAX_REPODATA_VERSION:
-            raise CondaUpgradeError(dals("""
-                The current version of conda is too old to read repodata from
-
-                    %s
+        try:
+            # 1. Try pulling lastmodified.json
```
kalefranz: Consider adding feature flag to gate new strategy, especially to make it easier to benchmark against the current full repodata.json strategy.

conda/core/subdir_data.py
```diff
@@ -126,39 +124,59 @@ def __init__(self, channel):
...
+            if cached_lastmodified:
+                self._names_index = cached_lastmodified._names_index
+                self._track_features_index = cached_lastmodified._track_features_index
+                self._package_records = cached_lastmodified  # implemented with __iter__
+            else:
+                self._package_records = []
+                self._names_index = {}
+                self._track_features_index = {}
+        except ResponseException as e:
+            log.info("falling back to repodata.json for %s due to %r", self.url_w_subdir, e)
```
kalefranz: TODO: Fall back to repodata.json if `lastmodified_version` is too new.

tests/core/test_subdir_data.py
```diff
@@ -2,17 +2,19 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
```
kalefranz: TODO: Write more tests.

msarahan:
I would prefer not to have the new .json files alongside repodata.json.  If we have them there, it raises some not-great questions about what the user-facing index.html should look like.  There will be thousands more entries on that page.  If you exclude them from that index, they'll be more obscure and harder to find.  I do frequently refer to repodata.json from that web view, so I do think it's something that needs to be in that index.html.

This flat structure will also be prohibitive to any effort to combine your package-name sharding idea with my time sharding idea.  Is there any way to have shards live in subfolders somehow?

I'm thinking something like:

```
subdir/shards/packagename/repodata.json
```

and where my time shards might be additional files in that folder like:

```
subdir/shards/packagename/current_repodata.json
subdir/shards/packagename/2018-01-01.json
```

msarahan:
That idea may need to incorporate namespaces somehow?

subdir/shards/namespace/packagename/repodata.json

travishathaway:

PR #8665, opened 2019-05-13
Handle errors from read only filesystems

chrisburr:
I'm currently setting up a multiuser conda installation on a distributed read only filesystem ([CVMFS](https://cernvm.cern.ch/portal/filesystem)). This currently fails to skip the root environment as the error raised is `EROFS` ("Read-only filesystem") instead of `EACCES` or `EPERM`.

My issues were with `create_envs_directory` and `create_package_cache_directory` but I've grep-ed though and added it an a few other places as well.

I've also fixed a minor typo in the tests.

### DIFF
--- a/conda/core/package_cache_data.py
+++ b/conda/core/package_cache_data.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import codecs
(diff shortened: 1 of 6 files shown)


msarahan:
Thanks for this fix.

PR #8666, opened 2019-05-13
Allow setting setgid to fail when making directories

chrisburr:
I'm currently setting up a multiuser installation where the user's home directories are stored on [AFS](https://en.wikipedia.org/wiki/Andrew_File_System). On this system running `chmod 2775` fails with a permission error. This prevents the package and environment directories from being created.

This PR avoid the issue by silently ignoring the error.

### DIFF
--- a/conda/gateways/disk/__init__.py
+++ b/conda/gateways/disk/__init__.py
@@ -94,4 +94,9 @@ def mkdir_p_sudo_safe(path):
     if not on_win:
         # set newly-created directory permissions to 02775
         # https://github.com/conda/conda/issues/6610#issuecomment-354478489
-        os.chmod(path, 0o2775)
+        try:
+            os.chmod(path, 0o2775)
+        except (OSError, IOError):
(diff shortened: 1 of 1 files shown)


msarahan:
I'm not sure how important this error message might be.  "trace" is mostly invisible.  Is it important enough to be a warning?  Or perhaps it would be too much spam to the console as a warning?

chrisburr:
I'm not certain but so far as I can tell this would only be a problem if users all of the following:
 * running conda as `root`
 * have a home directory on a filesystem that doesn't support setting the `setgid` bit
 * are using a multiuser setup
 * installing in their home directory

so I think it's safe for this error to be effectively invisible.

(FYI: I signed the CLA when I made this but I guess it takes a while to update?)

msarahan:
yeah, the CLA is a bit manual on our side.  I confirm that you have signed it.  I'm fine with this as it is.  I hope it helps.  Thanks for the PR.

PR #8670, opened 2019-05-14
Add capability to use custom repodata filename

msarahan:
This relaxes the assumption that conda's index file is always called repodata.json.  Instead, it allows the user to specify an arbitrary filename.  Conda will try to use that one instead.  If it fails to download the file, or if using that file ultimately leads to an unsatisfiable solution, it falls back silently to repodata.json and tries again.

This is used in tandem with https://github.com/conda/conda-build/pull/3502 to provide a reduced view of repodata that has only the most recent packages.  This is cached independently of repodata.json.  If it works, it will be much smaller than repodata.json, and will speed up the rest of the process.


conda/base/context.py
```diff
@@ -1114,6 +1116,11 @@ def description_map(self):
                 A list of features that are tracked by default. An entry here is similar to
                 adding an entry to the create_default_packages list.
                 """),
+            'repodata_fn': dals("""
+                Specify a filename for repodata fetching. The standard is
+                repodata.json, but you may want to specify something else to
```
jjhelmus: The phasing here seems in conflict with the implementation.  The default is to try "current_repodata.json" and if that fail fall back to "repodata.json" but the comment seem to imply that "repodata.json" is the default.

msarahan:
I believe the permission errors on appveyor will be fixed when conda-package-handling 1.1.1 is rolled out.  It's on conda-forge, but our build network had hiccups this evening.  I'll have to build it tomorrow.

conda/cli/install.py
```diff
@@ -265,15 +267,22 @@ def install(args, parser, command='install'):
...
-        channels_urls = tuple(calculate_channel_urls(
-            channel_urls=index_args['channel_urls'],
-            prepend=index_args['prepend'],
-            platform=None,
-            use_local=index_args['use_local'],
-        ))
-        raise PackagesNotFoundError(e._formatted_chains, channels_urls)
+        if args.repodata_fn != 'repodata.json':
+            args.repodata_fn = 'repodata.json'
+            return install(args, parser, command)
```
soapy1: I have 2 questions:
1. Is there a situation when someone might want to fall back on another repodata other than repodata.json? For example, if the first repodata doesn't work, then use a slightly larger view of repodata, then finally fall back on repodata.json.
2. How often is it expected for the package to not be found? In that case solving takes twice as long. Is ensuring that this doesn't happen super often a responsibility of the bit that generates repodata?

jjhelmus: 1. Users may want to use an epoch based repodata if the repodata with the lastest releases does not resolve.  One could imagine a system with multiple layers of fall backs. Each attempt to solve the environment with a particular repodata will have a time cost.  A preferred sparse repodata followed by a single fallback seems like a good initial step before implementing a fully layered approach.
2. Each solve will take time and if a solution is not found in the preferred repodata file that is time wasted.  The preferred repodata file default to a version with just the latest version of all packages which should be small and quick to process and check for a solution. The conda solver prefers these version when installing unbound specification so I expect that solutions will be found regularly via this method.  Basically, simple solves which use the latest version should be very quick (<5 seconds).  More complex solves which require the fallback repodata and second or two overhead to check the "easy" solution should not be a significant overhead.

soapy1: gotcha! For (2) it would be interesting to understand how many solves it needs for most users. Maybe we can look at hits on all the repodata.json's?

conda/core/link.py
```diff
@@ -515,49 +537,93 @@ def _verify(cls, prefix_setups, prefix_action_groups):
...
+                if future.result():
+                    exceptions.extend(future.result())
         return exceptions
 
     @classmethod
     def _execute(cls, all_action_groups):
         # unlink unlink_action_groups and unregister_action_groups
-        unlink_actions = tuple(
-            group for group in all_action_groups if group.type in ("unlink", "unregister"))
+        unlink_actions = tuple(group for group in all_action_groups if group.type == "unlink")
```
soapy1: what happened to the unregister group type?

conda/core/link.py
```diff
@@ -515,49 +537,93 @@ def _verify(cls, prefix_setups, prefix_action_groups):
...
+                                   for axngroup in group)
+                        for future in as_completed(futures):
+                            exc = future.result()
+                            if exc:
+                                log.debug('%r'.encode('utf-8'), exc.errors[0], exc_info=True)
+                                exceptions.append(exc)
+
+                        # Run post-link or post-unlink scripts and registering AFTER link/unlink,
+                        #    because they may depend on files in the prefix.  Additionally, run
+                        #    them serially, just in case order matters (hopefully not)
```
soapy1: If I recall correctly order matters. There was one package I ran into at least that the order that the post link scripts were running in was messed up and it broke the package sometimes.

msarahan: order is preserved, according to the original order in the axngroup.  I'm not certain that this is exactly the same as you used to have.

conda/core/link.py
msarahan: These come in a little later: https://github.com/conda/conda/pull/8670/files#diff-315c016a32ef55653e67795ff2b18b7cR613

PR #8671, opened 2019-05-14
move TemporaryDirectory from conda.common into gateways.disk.create

msarahan:
In response to suggestion at https://github.com/conda/conda/pull/8620#discussion_r283905473


conda/gateways/disk/create.py
```diff
@@ -31,6 +32,53 @@
...
+        Upon exiting the context, the directory and everything contained
+        in it are removed.
+        """
+
+        # Handle mkdtemp raising an exception
+        name = None
+        _closed = False
+
+        def __init__(self, suffix="", prefix='tmp', dir=None):
+            self.name = mkdtemp(suffix, prefix, dir)
```
jjhelmus:
```suggestion
            self.name = tempfilemkdtemp(suffix, prefix, dir)
```

PR #8672, opened 2019-05-15
break out of minimization when bisection has nowhere to go

msarahan:
In debugging conda slowness on my computer, I noticed in the debug logs (-vvv) that it seemed like we had an infinite loop:

```
TRACE conda.common.logic:LinearBound_(708): Eliminating 146/146 terms for bound violation
TRACE conda.common.logic:minimize(865): Bisection attempt: (1,0), (807+0) clauses
TRACE conda.common.logic:minimize(869): Bisection failure, new range=(1,0)
TRACE conda.common.logic:LinearBound_(708): Eliminating 146/146 terms for bound violation
TRACE conda.common.logic:minimize(865): Bisection attempt: (1,0), (807+0) clauses
TRACE conda.common.logic:minimize(869): Bisection failure, new range=(1,0)
```

This PR attempts to break out of the bisection loop if the range is empty.  I think this is especially helpful in the presence of inconsistent environments (my root env is pretty awful)

### DIFF
--- a/conda/common/logic.py
+++ b/conda/common/logic.py
@@ -838,6 +838,8 @@ class Clauses(object):
                 if newsol is None:
                     lo = mid + 1
                     log.trace("Bisection failure, new range=(%d,%d)" % (lo, hi))
+                    if lo > hi:
+                        break
                     # If this was a failure of the first test after peak minimization,
                     # then it means that the peak minimizer is "tight" and we don't need
                     # any further constraints.


conda/common/logic.py
```diff
@@ -838,6 +838,8 @@ def sum_val(sol, odict):
                 if newsol is None:
                     lo = mid + 1
                     log.trace("Bisection failure, new range=(%d,%d)" % (lo, hi))
+                    if lo > hi:
```
mbargull: Huh, this should "never" happen.
Just breaking out of the loop might violate some invariant/expected post-condition, though.
We have to rather fix the cause of this. I'll try to have a look at this this week.
@msarahan, do you have a failing test case for this? That might make debugging considerably faster for me ;).

PR #8682, opened 2019-05-17
Removing free channel from defaults

rrigdon:
This PR adds in the free channel page detailing its removal from defaults, minor grammatical/formatting edits, and the reorganized .condarc doc with free channel info.

### DIFF
--- a/docs/source/user-guide/configuration/free-channel.rst
+++ b/docs/source/user-guide/configuration/free-channel.rst
@@ -1,0 +1,105 @@
+======================
+Using the free channel
+======================
+
+.. contents::
+   :local:
+   :depth: 2
+
+The free channel contains packages created prior to
+September 26, 2017. Prior to conda 4.7, the free
+channel was part of the ``defaults`` channel.
+Read more about the :ref:`defaults channel <default-channels>`.
+
+Removing the ``free`` channel reduced conda's search space
+and hid old software. That old software could have incompatible
+constraint information.
+
+If you still need the content from the ``free`` channel to reproduce
+old environments, you can re-add the channel following the directions below.
+
+.. _free-channel-default:
+
+Adding the free channel to defaults
+===================================
+
+If you want to add the ``free`` channel back into your default list,
+use the command::
+
+   conda config --set restore_free_channel true
+
+The order of the channels is important. Using the above
+command will restore the ``free`` channel in the correct order.
+
+Changing .condarc
+=================
+
+You can also add the ``free`` channel back into your defaults by
+changing the ``.condarc`` file itself.
+
+Add the following to the conda section of your ``.condarc`` file::
+   
+   restore_free_channel: true
+
+Read more about :doc:`use-condarc`.
+
+Package name changes
+====================
+
+Some packages that are available in the ``free`` channel
+have different names in the ``main`` channel.
+ 
+.. list-table::
+   :widths: 25 75
+   :header-rows: 1
+
+   * - Package name in ``free``
+     - Package name in ``main``
+   * - dateutil
+     - python-dateutil
+   * - gcc
+     - gcc_linux-64 and similar
+   * - pil
+     - pillow
+   * - ipython-notebook
+     - now installable via notebook, a metapackage could be created
+   * - Ipython-qtconsole
+     - now installable via qtconsole, a metapackage could be created
+   * - beautiful-soup
+     - beautifulsoup4
+   * - pydot-ng
+     - pydot
+
+
+Troubleshooting
+===============
+
+You may encounter some errors, such as UnsatisfiableError
+or a PackagesNotFoundError.
+
+An example of this error is::
+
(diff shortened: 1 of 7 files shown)


docs/source/user-guide/configuration/free-channel.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Changing .condarc
+=================
+
+You can also add the ``free`` channel back into your defaults by
+changing the ``.condarc`` file itself.
+
+Add the following to the conda section of your ``.condarc`` file::
+   
+   restore_free_channel: true
```
csoja: Does this always work? even if someone adds other channels like conda-forge to their defaults list?

docs/source/user-guide/configuration/free-channel.rst
```diff
@@ -0,0 +1,105 @@
...
+   Use "conda search <package> --info" to see the dependencies for each package.
+
+This can occur if:
+
+- you’re trying to install a package that is only available in
+  ``free`` and not in ``main``.
+- you have older environments in files you want to recreate.
+  If those spec files reference packages that are in ``free``,
+  they will not show up.
+- a package dependency is only available in the ``free`` channel,
```
csoja: This last bullet reads strangely to me.  I understand the intent - but it isn't so much that the package won't work without its dependency - but that conda won't let you install the package because it cannot install the dependency - which the package requires to work.

docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -161,6 +182,12 @@ This is especially useful for air gap and enterprise installations:
     - http://some.custom/channel
     - file:///some/local/directory
 
+As of conda 3.7, the ``free`` channel is no longer part of the
+``defaults`` channel. If you want to add ``free`` back into your
```
csoja: So this section is explaining how to redefine "defaults" in your .condarc file from the standard defaults that come with conda.    The free channel is part of the standard "defaults" - but if someone sets the default channels in .condarc - then only the ones they list are searched by default.  In other words, if someone adds this section into their .condarc - and the free or main channel isn't there, then it isn't in "defaults" - regardless of which version of conda.

docs/source/user-guide/configuration/free-channel.rst
rrigdon: @msarahan  - Can you help with Crystal's question above please?

msarahan: Yes, it's a separate collection of stuff.  If people have set custom default_channels, it might not do the right thing.  I will check that.

docs/source/user-guide/configuration/free-channel.rst
msarahan: That's good clarification.  Thanks!

docs/source/user-guide/configuration/use-condarc.rst
msarahan: Also good clarification.  We should add a reference to the ``default_channels`` config parameter docs.

docs/source/user-guide/configuration/free-channel.rst
rrigdon: Is this clearer: 

> - a package is dependent upon files found only in the ``free``
>   channel. Conda will not let you install the package if it cannot
>   install the dependency, which the package requires to work.

msarahan: I think that's good, but if like Crystal's opinion, too.

docs/source/user-guide/configuration/use-condarc.rst
rrigdon: It seems like it doesn't really make sense to bring up the ``free`` channel in this section then?

@msarahan - what docs should I refer to? https://docs.conda.io/projects/conda/en/latest/configuration.html ?

docs/source/user-guide/configuration/free-channel.rst
csoja: this LGTM

docs/source/user-guide/configuration/use-condarc.rst
msarahan: This one was what I had in mind:

https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html?highlight=default_channels#default-channels-default-channels

docs/source/user-guide/configuration/free-channel.rst
msarahan: sorry, I haven't gotten to this yet.  I'll try to look at it today.

docs/source/user-guide/configuration/use-condarc.rst
rrigdon: That link is for the same section we're editing in. I removed the reference to the ``free`` channel entirely for this part.

msarahan:
I need to look into corner cases with adding free back in, but I guess we can maybe wait until a problem shows up.

Issue #8687, opened 2019-05-19
conda env export fails due to incorrect PyPI spec parsing


martinkou:
## Current Behavior
After upgrading from conda 4.5.x to conda 4.6.x, installing certain packages from PyPI would cause `conda env export` to fail.

e.g. this is seen with installing the `eth-keyfile` package from PyPI.

### Steps to Reproduce
```
$ conda create -n eth-test python==3.6.7
$ conda activate eth-test
$ pip install eth-keyfile==0.5.1
$ conda env export

InvalidVersionSpec: Invalid version '(<1.0.0,>=0.1.0': unable to convert to expression tree: ['(']
```

This is caused by the PyPI spec parser not being able to parse certain requirement strings in the eth-keyfile package. e.g. "eth-utils>=1.0.0-beta.1,<2.0.0". You can see some of the requirement strings here: https://github.com/ethereum/eth-keyfile/blob/v0.5.1/setup.py

In particular, the `parse_specification()` function fails at parsing the spec string above.

```
>>> from conda.common.pkg_formats.python import parse_specification
>>> spec = "eth-utils>=1.0.0-beta.1,<2.0.0"
>>> parse_specification(spec)
PySpec(name='eth-utils', extras=[], constraints='>=1.0.0', marker='', url='')
```

## Expected Behavior
conda 4.5.x was able to print out the environment dependencies correctly:

```
name: eth-test
channels:
  - defaults
dependencies:
  - ca-certificates=2019.1.23=0
  - certifi=2019.3.9=py36_0
  - libedit=3.1.20181209=hc058e9b_0
  - libffi=3.2.1=hd88cf55_4
  - libgcc-ng=8.2.0=hdf63c60_1
  - libstdcxx-ng=8.2.0=hdf63c60_1
  - ncurses=6.1=he6710b0_1
  - openssl=1.1.1b=h7b6447c_1
  - pip=19.1.1=py36_0
  - python=3.6.7=h0371630_0
  - readline=7.0=h7b6447c_5
  - setuptools=41.0.1=py36_0
  - sqlite=3.28.0=h7b6447c_0
  - tk=8.6.8=hbc83047_0
  - wheel=0.33.4=py36_0
  - xz=5.2.4=h14c3975_4
  - zlib=1.2.11=h7b6447c_3
  - pip:
    - cytoolz==0.9.0.1
    - eth-hash==0.2.0
    - eth-keyfile==0.5.1
    - eth-keys==0.2.2
[...]

PR #8688, opened 2019-05-19
Fix for #8687: modified PyPI spec parsing regex to parse hyphenated versions correctly.

martinkou:
This patch allows the correct parsing of required version strings like ">=1.0.0-beta.1,<2.0.0", which appears in PyPI packages like `eth-keyfile==0.5.1` in the wild.

Unit test case included.

### DIFF
--- a/tests/common/pkg_formats/test_python.py
+++ b/tests/common/pkg_formats/test_python.py
@@ -201,7 +201,7 @@ def test_parse_specification():
         '>=2.7,!=3.0.*,!=3.1.*,!=3.2.*':
             PySpec('', [], '>=2.7,!=3.0.*,!=3.1.*,!=3.2.*', '', ''),
         'name>=1.0.0-beta.1,<2.0.0':
-            PySpec('name', [], '>=1.0.0-beta.1,<2.0.0', '', ''),
+            PySpec('name', [], '>=1.0.0.beta.1,<2.0.0', '', ''),
     }
     for req, expected_req in test_reqs.items():
         parsed_req = parse_specification(req)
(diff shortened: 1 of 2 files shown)


msarahan:
Thanks for this.  We have a problem with the new file format.  I'll re-run the tests when we have that fixed, and I'm pretty sure this will be ready to merge.

kalefranz:
As I recall, hyphens are not supported in version strings.  It's all been deliberated and discussed extensively.  I don't remember the reasons (there are issues to look back at in the history), but they are real.

kalefranz:
I guess here if we encounter a hyphen in a python version string we should just change it to a dot.

martinkou:
Would it be sufficient if I just replace the hyphens to dots in the parsing function? Or is a more sophisticated fix needed?

msarahan:
I think that's what @kalefranz is saying, but I'm not sure.  @kalefranz can you please confirm?

kalefranz:
Not quite. I’m saying the version string in the new test should become `>=1.0.0.beta.1,<2.0.0`, and then implement whatever code change is necessary to make that happen.

tests/common/pkg_formats/test_python.py
```diff
@@ -200,6 +200,8 @@ def test_parse_specification():
             PySpec('', [], '>=3,<2', '', ''),
         '>=2.7,!=3.0.*,!=3.1.*,!=3.2.*':
             PySpec('', [], '>=2.7,!=3.0.*,!=3.1.*,!=3.2.*', '', ''),
+        'name>=1.0.0-beta.1,<2.0.0':
```
kalefranz: Keep this ^^ line the same

tests/common/pkg_formats/test_python.py
```diff
@@ -200,6 +200,8 @@ def test_parse_specification():
...
+            PySpec('name', [], '>=1.0.0-beta.1,<2.0.0', '', ''),
```
kalefranz: Change the version spec string here ^^ to `>=1.0.0.beta.1,<2.0.0`

martinkou:
Updated the patch as discussed.

msarahan:
I've rebased your PR on master, which should fix the tests that you had nothing to do with.  I'll watch the CI and fix as necessary until this goes in.  Thanks @martinkou

PR #8689, opened 2019-05-19
fix PrefixRecord creation when file inputs are .conda files

msarahan:
also puts in a DummyExecutor that can be enabled with --debug flag.  This is helpful for debugging, because pdb/ipdb don't always work very well in presence of ThreadPoolExecutor or ProcessPoolExecutor.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -267,17 +267,7 @@ class Solver(object):
 
     @time_recorder(module_name=__name__)
     def _collect_all_metadata(self, ssc):
-        if ssc.specs_from_history_map:
-            # Start with empty specs map for UPDATE_ALL because we're optimizing the update
-            # only for specs the user has requested; it's ok to remove dependencies.
-
-            # However, because of https://github.com/conda/constructor/issues/138, we need
-            # to hard-code keeping conda, conda-build, and anaconda, if they're already in
-            # the environment.
-            for pkg_name in ('anaconda', 'conda', 'conda-build'):
-                if ssc.prefix_data.get(pkg_name, None):
-                    ssc.specs_from_history_map[pkg_name] = MatchSpec(pkg_name)
-        else:
+        if not ssc.specs_from_history_map:
             ssc.specs_map.update(
                 (prec.name, MatchSpec(prec.name)) for prec in ssc.prefix_data.iter_records()
             )
@@ -285,11 +275,25 @@ class Solver(object):
         # add in historically-requested specs
         ssc.specs_map.update(ssc.specs_from_history_map)
 
-        # add in aggressively updated packages
+        for pkg_name in ('anaconda', 'conda', 'conda-build'):
+            if pkg_name not in ssc.specs_map and ssc.prefix_data.get(pkg_name, None):
+                ssc.specs_map[pkg_name] = MatchSpec(pkg_name)
+
         ssc.specs_map.update(
             (prec.name, MatchSpec(prec.name)) for prec in ssc.prefix_data.iter_records()
-            if MatchSpec(prec.name) in context.aggressive_update_packages
+            if prec.subdir == 'pypi'
         )
+        for prec in ssc.prefix_data.iter_records():
+            # first check: add in aggressively updated packages
+            #
+            # second check: add in foreign stuff (e.g. from pip) into the specs
+            #    map. We add it so that it can be left alone more. This is a
+            #    declaration that it is manually installed, much like the
+            #    history map. It may still be replaced if it is in conflict,
+            #    but it is not just an indirect dep that can be pruned.
+            if (MatchSpec(prec.name) in context.aggressive_update_packages or
+                    prec.subdir == 'pypi'):
+                ssc.specs_map.update({prec.name: MatchSpec(prec.name)})
 
         prepared_specs = set(concatv(
             self.specs_to_remove,
@@ -396,8 +400,8 @@ class Solver(object):
 
         # the only things we should consider freezing are things that don't conflict with the new
         #    specs being added.
-        explicit_spec_package_pool = {}
-        if ssc.update_modifier == UpdateModifier.FREEZE_INSTALLED:
+        def get_explicit_spec_package_pool(specs_to_add):
+            explicit_spec_package_pool = {}
             specs_seen = set()
             specs_pool = deque(copy.copy(self.specs_to_add))
             while specs_pool:
@@ -411,6 +415,9 @@ class Solver(object):
                         dep = MatchSpec(dep)
                         if dep not in specs_seen and dep not in specs_pool:
                             specs_pool.append(dep)
+            return explicit_spec_package_pool
+        if ssc.update_modifier == UpdateModifier.FREEZE_INSTALLED:
+            explicit_spec_package_pool = get_explicit_spec_package_pool(self.specs_to_add)
 
         def pkg_in_shared_pool(prec, shared_pool, known_bad_specs, known_good_specs):
             """Determine if a package or its recursive deps conflict with the shared pool"""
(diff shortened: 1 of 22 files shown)


msarahan:
repo's index is fixed up.  Close/reopen to cycle CI.

msarahan:
I think the 2 remaining errors on windows are due to excessively long paths:

```
C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-1\\test_list_with_pip_no_binary0\\tmp4jx90xog\\f40b \xe1\xf1\u56fa\ud55c\xe7\u03b3\u5bb6\xdf\u014d\xea\xf4\u0452\\Lib\\site-packages\\flask\\testsuite\\test_apps\\lib\\python2.5\\site-packages\\site_package\\__init__.py
```

conda/core/solve.py
```diff
@@ -361,9 +370,7 @@ def _find_inconsistent_packages(self, ssc):
             The following packages are causing the inconsistency:"""))
             print(dashlist(inconsistent_precs))
             for prec in inconsistent_precs:
-                # pop and save matching spec in specs_map
-                spec = ssc.specs_map.pop(prec.name, None)
-                ssc.add_back_map[prec.name] = (prec, spec)
+                ssc.add_back_map[prec.name] = (prec, None)
```
soapy1: Why does this have a tuple of `(prec, None)` opposed to just the prec?

msarahan: That looks like a bug to me.  The spec is otherwise not removed from the specs_map, which is probably important.

tests/test_create.py
```diff
@@ -1205,13 +1202,6 @@ def test_install_update_deps_only_deps_flags(self):
             assert package_is_installed(prefix, "flask=0.12.2")
             assert package_is_installed(prefix, "jinja2>2.9")
 
-    @pytest.mark.skipif(on_win, reason="tensorflow package used in test not available on Windows")
```
msarahan: removing this is intentional, as ``--freeze-installed`` is the default behavior now.  However, it's not completely frozen as it was for this test - only frozen as much as possible, but still within the bounds of the explicit constraints in the history.  This test no longer raises the expected UnsatisfiableError.

tests/test_create.py
```diff
@@ -1122,13 +1124,8 @@ def test_install_prune_flag(self):
             assert package_is_installed(prefix, 'python=3')
             run_command(Commands.REMOVE, prefix, "flask")
             assert not package_is_installed(prefix, 'flask')
-            assert package_is_installed(prefix, 'itsdangerous')
```
msarahan: This is removed because the default behavior is pruning as of this PR

tests/core/test_solve.py
```diff
@@ -642,91 +634,43 @@ def test_broken_install():
...
-    # }
-    # with patch.object(History, 'get_requested_specs_map', return_value=spec_map):
-    #     solver = Solver(prefix, (Channel('defaults'),), context.subdirs,
-    #                     specs_to_remove=(MatchSpec("pandas"),))
-    #     solver.index = index
-    #     solver.r = r
-    #     solver._prepared = True
+    # these tests don't work because the specs_map used to determine whether to resolve an
+    #    inconsistency is derived from the history.  Since pandas is in the history, the solver
+    #    will just make it consistent - this injection of inconsistency doesn't work
```
soapy1: Can these tests be removed if they aren't valid anymore

tests/core/test_solve.py
```diff
@@ -249,46 +249,32 @@ def test_prune_1():
     specs_to_remove = MatchSpec("numbapro"),
     with get_solver(specs_to_remove=specs_to_remove, prefix_records=final_state_1,
                     history_specs=specs) as solver:
-        unlink_precs, link_precs = solver.solve_for_diff(prune=False)
```
msarahan: prune is the default behavior.  This argument has no effect.  The prune=True case is shifted above, and remains similar.

tests/core/test_solve.py
```diff
@@ -552,7 +544,7 @@ def test_update_all_1():
 
 
 def test_broken_install():
-    specs = MatchSpec("pandas"), MatchSpec("python=2.7"), MatchSpec("numpy 1.6.*")
+    specs = MatchSpec("pandas=0.11.0=np16py27_1"), MatchSpec("python=2.7")
```
msarahan: We change numpy to be an implicit spec here.  Without doing this, conda will fix the inconsistent spec that we introduce later more aggressively, and our test won't work.

tests/core/test_solve.py
```diff
@@ -861,214 +806,56 @@ def test_conda_downgrade():
...
+                'channel-2::conda-4.3.30-py36h5d9f9f4_0',
+                'channel-4::conda-build-3.12.1-py36_0'
             )
             assert convert_to_dist_str(unlink_precs) == unlink_order
             assert convert_to_dist_str(link_precs) == link_order
     finally:
         sys.prefix = saved_sys_prefix
 
 
-def test_install_uninstall_features_1():
```
soapy1: why do test_install_uninstall_features_1 and test_install_uninstall_features_2 get removed here? I thought they should pretty much be unaffected by the changes in the PR?

msarahan: these tests are unfortunately totally irrelevant.  None of our packages have any features anymore.  We have hotfixed them all away.  These tests broke because of the new, different pruning behavior.

We need to write new tests if we're going to maintain the features functionality.  I'm inclined to rip it out, but we may need to have that be in 5.0.

msarahan:
There's some good discussion relevant to this PR happening in https://github.com/conda/conda/pull/8738#discussion_r289118460

Unfortunately, things got conflated due to my impatience in rebasing other PRs on top of this one.

soapy1:
Looks good. Is it worth it to add a test here for some strange state as a result of something like installing packages with pip. For example, pip install an old version of numpy and then try to conda install a package that relies on a newer version of numpy. I think I would expect that conda should error with  some inconsistent environment error.

conda/core/solve.py
msarahan: This one is more complicated than I thought.  Here's how it works:

With pop:
- inconsistent dep is removed
- none of its specs become part of the solution (and may thus be eliminated)
- package is restored, and its deps may or may not be there

Without pop:
- inconsistent dep is not factored into the solution, but it is entered in as a name spec
- solution goes ahead with any dependencies satisfied
- consistent package will get clobbered by the original inconsistent spec

I'll push up some changes here.

kalefranz:
Speaking of pip installed packages... If `pip_interop_enabled` is on, is the new pruning behavior going to wipe out all of the manageable packages that have been pip-installed? Those won’t be user-requested conda specs in the history file...

kalefranz:
Speaking of pip installed packages... If `pip_interop_enabled` is on, is the new pruning behavior going to wipe out all of the manageable packages that have been pip-installed? Those won’t be user-requested conda specs in the history file...

tests/test_create.py
```diff
@@ -2026,6 +2026,32 @@ def test_conda_pip_interop_compatible_release_operator(self):
                 run_command(Commands.INSTALL, prefix, "-c", "https://repo.anaconda.com/pkgs/free",
                             "agate=1.6", "--dry-run")
 
+    def test_conda_recovery_of_pip_inconsistent_env(self):
+        with make_temp_env("pip=10", "python", "anaconda-client",
+                           use_restricted_unicode=on_win) as prefix:
+            run_command(Commands.CONFIG, prefix, "--set", "pip_interop_enabled", "true")
```
soapy1: will this test only work if pip_interop is enabled?

msarahan:
@kalefranz yes, those are absolutely a concern.  I believe I have them handled in:

https://github.com/conda/conda/pull/8689/files#diff-127bca0abbd39d90577af817c26af378R288
https://github.com/conda/conda/pull/8689/files#diff-127bca0abbd39d90577af817c26af378R453

tests/test_create.py
msarahan: I think so.  If it doesn't then the precs for the pip installed stuff won't be in the collection of prefix_records, if I understand correctly.

PR #8698, opened 2019-05-20
maintain download() interface in conda.exports

kalefranz:


soapy1:
I believe the conda-build test failures are not a result of the changes in this PR

msarahan:
The tests are a mess because the new format stuff is way off.  I'm working on it.  I'm pretty sure this PR is fine, but it would be nice to clarify without the new format mess, just in case.

msarahan:
index is back in shape.  Cycling...

msarahan:

PR #8715, opened 2019-05-22
Ensure username is not in upload report

soapy1:


msarahan:
This looks like a good implementation. One thing I think we might want to try to keep is whether or not the username had:

- Spaces
- Unicode

soapy1:
Makes sense! Pushed up some changes that will add those to the error log

conda/exceptions.py
```diff
@@ -1237,7 +1238,11 @@ def _execute_upload(self, error_report):
             'User-Agent': self.user_agent,
         }
         _timeout = self.http_timeout
+        username = getpass.getuser()
+        error_report['is_unicode'] = True if isinstance(username, bytes) else False
```
msarahan: I think a better way to express this might be "does the username contain characters that are outside of the ascii set?"

https://stackoverflow.com/questions/196345/how-to-check-if-a-string-in-python-is-in-ascii

msarahan:

PR #8723, opened 2019-05-24
Fall back to old file format

### LINKED ISSUE #8697: Option to fall back to old file format
The rollout of the new file format has exposed some problems in conda-package-handling.  Some/all packages are missing symlinks (at least to folders) which makes conda barf about missing content.

We need a way for users to get unstuck if the new format doesn't work for them for whatever reason.

soapy1:
fixes #8697


msarahan:

PR #8727, opened 2019-05-27
JSON-format activation

necaris:
This is to serve the use-case of tools (e.g. editors) that want to know about conda environments, and potentially even activate them for the user. Rather than attempt to parse a shell script, why not simply provide the necessary values directly in a machine-readable format?


jlstevens:
@kalefranz Just a heads up to say this is the feature I discussed with you a while back. I believe this is one of the ideas you suggested and I definitely support this PR!

msarahan:
This is very cool. We have really needed this for a long time. I'm going to spend some proper time reviewing this after I get the 4.7.1 release out. Thanks @necaris !

necaris:
@jlstevens thanks for the ping about it!

@msarahan y'all are very welcome! I'm looking forward to being able to use this functionality in my own editor integration ;-) By the way, I'm not wedded to any of the design decisions made in this PR so happy to change things up as you get time to review & comment.

necaris:
@msarahan rebased this just now but still seeing test failures -- not sure what's causing them, could still use a hand figuring it out.

necaris:
@msarahan is this still of interest?

kalefranz:
Hey Rami, I'm moving back over from the enterprise side back to Conda for a while. Just starting as of yesterday, so need some time to ramp back up. Yes, this is still of interest.

necaris:
@kalefranz thanks! I'll rebase it against current master then, but will probably need a hand figuring out why some tests have continued to fail on CI (including code paths that I never touched 😉 )

kalefranz:
Yeah I'm starting to dig into the tests now actually

necaris:
@kalefranz don't know if this is still on your radar, but I've just rebased it against current master, FYI, so it should be up to date.

mcg1969:
Hey @necaris, I'm taking a fresh look at this

necaris:
@mcg1969 thanks!  I could never quite figure out what was causing the test failures -- would love a hand getting this across the line.

mcg1969:
@kalefranz @chenghlee The test failures are all centered around failure to connect to the CodeCov service. Can we please work around those? The functional tests look good

int19h:
This is incredibly helpful for tooling, thank you!

Just to clarify - after reading through the code for this file, it sounds like almost all this would actually work for _any_ shell (assuming the API client correctly applies "path" and "vars"), despite the name being "posix:json", except for activate/deactivate scripts? And if https://github.com/conda/conda/issues/6820 is resolved, then it would be shell-agnostic even in that respect?

necaris:
@int19h the hope is absolutely that it would work for _any_ shell -- when developing it, I wasn't able to test / check it against Windows so only felt confident naming it `posix:`. I'd be happy to remove that prefix.

int19h:
@necaris I meant that the "path" and "vars" bits in the resulting object should be identical for all shells - at least, so far as I can tell from various activators. The "scripts" part is different, though, because it depends on `shell_extension`, which varies. Thus, so long as scripts remain shell-specific, it would require the same mixin applied to every activator separately.

I can give it a try with other activators on Windows, and report back on how well that works.

Also, a suggestion for the PR: rather than manually combining the mixin with each activator, perhaps treat `:` in the shell name specially in main, and do the combination there? Basically just split it on `:`, and *dynamically* construct a class that combines all the activators named, fetching them one by one from the activator map. Something like:
```py
    shells = argv[1].replace('shell.', '', 1).split(':')
    try:
        activator_bases = [activator_map[shell] for shell in shells]
    except KeyError:
        raise CondaError("%s is not a supported shell." % shell)
    class Activator(*activator_bases):
        pass
    activator = Activator(activator_args)
```

This would allow for arbitrary combinations, and accommodate any future mixings as well. If a particular mixin is not compatible with a particular activator (or with another mixin), it can do a type-check on `self` during construction and raise.

(With such semantics, a symbol like `+` might also make more sense than `:`.)

necaris:
@int19h it sounds to me like your comment is in two parts. To the first part: as far as I can tell you're right that the `path` and `vars` should be consistent regardless of shell, which is really nice :smile: . I think it would be a big change to the intended ergonomics of `conda` for the helper `scripts` to become shell-agnostic, so I don't imagine that will happen any time soon?

To the second part: I like your suggestion, but I'm not sure it's in the scope of these changes. This PR deliberately mimics the current activators as closely as possible, so that it's a minimal behavior change the maintainers can hopefully get on board with soon. 

Unless I've misunderstood your suggestion, it would mean a refactor of how activators are parsed / applied in general, which IMHO would be a cool addition on its own, and perhaps orthogonal to this one?

int19h:
I suggested it as a way to simplify the logic, to avoid having to define all possible combinations of activator+mixin manually for all shells (`JSONCmdExeActivator`, `JSONFishActivator` etc) - which would be necessary to make this feature work consistently on all platforms that are supported by Python tooling that needs this feature to work well with conda. 

My proposal wouldn't require changing the existing activators, or how they're parsed/applied - only the logic for looking up the right activator given its name (which is currently a 4-line `try/except`, and would have to be replaced with something like the snippet above, in the same file that is touched by this PR).

It is, indeed, orthogonal in a sense that it's not really specific to JSON activation - just a generic way to combine activators. But without JSON activation, there wouldn't be any scenario providing motivation to consider such a feature in the first place.  And, conversely, it would simplify the implementation of JSON activation. So, purely from a pragmatic perspective, it might work out better as a single combo change, rather than making a change where activators are combined manually (just for JSON), only to replace that part of it shorty after. The latter would also require the manually defined names for combo JSON activators to be preserved for backwards compatibility later on, which might limit design choices. 

FWIW I don't personally see it as a big deal - just something to consider. If it would require a much longer approval for the change, it's probably not worth it.

necaris:
@int19h Thanks for clarifying! I hadn't understood that the intent was focused on adding the JSON-format capability to existing activator classes.

I'd love to hear from the maintainers on whether that'd improve this PR from their point of view, or anything else I / we could do to make this mergeable. (I've added an implementation based on @int19h's suggestion in https://github.com/necaris/conda/pull/1 so it'd be trivial to merge in if it simplifies matters).

int19h:
I've manually verified that the mixin does the right thing with `CmdExe` and `PowerShell` activators.

int19h:
Ditto for `Fish`.

mcg1969:
Thanks @int19h ! This is really helpful

necaris:
Thanks @int19h ! With that confirmation, I went ahead and merged my version of your suggested code in https://github.com/conda/conda/pull/8727/commits/31a1f3099cfc7ba32680f5059d450a496f0b8117 so that this PR includes that simplifying factor. 

Any comments would be appreciated -- I'm currently getting errors about the metaclass of the dynamic `Activator` needing to match the metaclasses of all its bases :disappointed: must have done something silly somewhere. Would also be happy to revert the commit for now and submit as a separate PR, in the interest of getting this mergeable quicker.

int19h:
My Python-fu is not that strong, but I wonder if it would help to use `class` in a loop to add one mixin at a time. Something like:
```py
activator_cls = activator_map[shell_etc[0]]
for format_mixin in (formatter_map[s] for s in shell_etc[1:]):
    class ActivatorWithMixin(activator_cls, format_mixin):
        pass
    activator_cls = ActivatorWithMixin
```

necaris:
In the interest of keeping the blockers off the critical path for this I've reverted the restructuring in https://github.com/conda/conda/pull/8727/commits/aef6861ed2b3037ec0c2b24e0374168b391f04ce -- I'll bring it back once I've had more of a chance to test things locally.

necaris:
Note that from what I can see, the AppVeyor tests are failing because `dev-init.bat` fails with an "Invalid code page" -- I am assuming that this is completely unrelated to this PR.

necaris:
I've brought back the restructured dynamic activator class in https://github.com/conda/conda/pull/8727/commits/c0894eebe6d5dd39b1fb1709f1c60f58e6ef0c3e with a helper function, and correctly updated tests which were missing before. For good measure I've also rebased against upstream master.

necaris:
Looks like the codecov issued is fixed -- thanks! The only thing that's not passing is AppVeyor, for the same reason as noted above: not IMHO related to this PR at all.

mcg1969:
This is great. We're getting very close. Will see if I can merge without AppVeyor passage, but I don't want to step on any toes on the team.

mcg1969:
OK, I know the problem we're seeing in CI

mcg1969:
boom! thanks folks

necaris:
Awesome! Thanks so much @mcg1969!  And thanks to @int19h !

PR #8737, opened 2019-05-30
Added tutorial to .condarc and cleaned up misc errors

rrigdon:


docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -49,17 +43,34 @@ The ``.condarc`` file can change many parameters, including:
...
   conda config --add channels conda-forge
 
+Alternatively, you can open a text editor such as Notepad
+on Windows, TextEdit on MacOS, or VS Code. Name the new file
+``.condarc`` and save it to your user home directory or root
+directory. To edit the ``.condarc`` file, open it from your
+home or root directory and make edits in the same way you would
+with any other text file. If the ``.condarc`` file is in the root
+environment, it will override any in the home directory.
+
```
msarahan: Should we mention here that the .condarc file is interpreted with standard YAML syntax?

rrigdon: There is a section above that states it so I pulled that sentence down into the Creating and Editing section.

PR #8741, opened 2019-05-31
minimize churn when channels change

msarahan:
This fixes and adds a test to ensure that when users specify a channel for one command, it does not update/change other packages excessively.  This was a regression in 4.7.1.


conda/core/solve.py
```diff
@@ -420,7 +421,8 @@ def _add_specs(self, ssc):
                 elif ssc.update_modifier == UpdateModifier.FREEZE_INSTALLED:
                     if pkg_name in context.aggressive_update_packages:
                         ssc.specs_map[pkg_name] = pkg_name
-                    elif target_prec in explicit_spec_package_pool:
+                    elif (target_prec in explicit_spec_package_pool or
+                          target_prec.name not in explicit_spec_package_pool_names):
```
soapy1: What's the difference between `explicit_spec_package_pool` and `explicit_spec_package_pool_names`? From the naming it seems like they should represent pretty much the same data?

msarahan: These are two different set lookups.  The first is an exact match.  If it finds this match, then freezing things is definitely OK.  The second is an "is this package related to anything I'm doing?" check.  If it's not related, we can also freeze this package, because it shouldn't be affected by what we're trying to do.

soapy1: ahh, gotcha!

conda/resolve.py
```diff
@@ -544,26 +572,14 @@ def filter_group(_specs):
...
+            slist = deque(explicit_specs)
             reduced = False
             while slist:
-                s = slist.pop()
+                s = slist.popleft()
                 top_level_spec = s
                 reduced = filter_group([s])
                 if reduced:
                     slist.append(s)
-                elif reduced is None:
```
msarahan: Removing this early bail-out allows get_reduced_index to serve as a satisfiability check.  Any entries with 0 results are part of a conflict.  This avoids a SAT call, at the cost of doing more filtering in get_reduced_index.  I'm not sure which will be faster in practice.  Based on the test suite, they seem about the same, but this get_reduced_index approach may have more room for optimization / conversion to compiled code.

conda/resolve.py
```diff
@@ -961,41 +977,32 @@ def environment_is_consistent(self, installed):
...
+        reduced_index = self.get_reduced_index(tuple(specs), sort_by_exactness=False)
         r2 = Resolve(reduced_index, True, channels=self.channels)
-        C = r2.gen_clauses()
-        solution = mysat(specs, True)
-        if solution:
+        scp = context.channel_priority == ChannelPriority.STRICT
+        unsat_specs = {s for s in specs if not r2.find_matches_with_strict(s, scp)}
+        if not unsat_specs:
             return ()
-        else:
```
msarahan: see https://github.com/conda/conda/pull/8741/files#r291598332 for why SAT is removed here.

tests/core/test_solve.py
```diff
@@ -461,11 +452,17 @@ def test_only_deps_2():
...
             'channel-1::nose-1.3.0-py27_0',
             'channel-1::numpy-1.7.1-py27_0',
-            # 'channel-1::numba-0.5.0-np17py27_0',
+            # 'channel-1::numba-0.5.0-np17py27_0',  # not in the order because only_deps
         )
         assert convert_to_dist_str(final_state_2) == order
 
+    # fails because numpy=1.5 is in our history as an explicit spec
     specs_to_add = MatchSpec("numba=0.5"),
+    with get_solver(specs_to_add, prefix_records=final_state_1, history_specs=specs) as solver:
```
msarahan: These failures are new behavior in this PR.  We're much more strict about requiring people to explicitly override any past explicit decision.

tests/core/test_solve.py
```diff
@@ -1252,18 +1249,12 @@ def test_fast_update_with_update_modifier_not_set():
...
-            'channel-4::python-2.7.14-h89e7a4a_22',
             'channel-4::sqlite-3.21.0-h1bed415_2',
-            'channel-4::libedit-3.1-heed3624_0',
             'channel-4::openssl-1.0.2l-h077ae2c_5',
-            'channel-4::ncurses-6.0-h9df7e31_2',
         )
         link_order = (
-            'channel-4::ncurses-6.1-hf484d3e_0',
             'channel-4::openssl-1.0.2p-h14c3975_0',
-            'channel-4::libedit-3.1.20170329-h6b74fdf_2',
```
msarahan: This mess is stuff I messed up in earlier 4.7.  This PR restores some sanity here by doing a better job with knowing when to freeze specs.

tests/core/test_solve.py
```diff
@@ -1341,14 +1332,14 @@ def test_pinned_1():
             assert convert_to_dist_str(final_state_2) == order
 
         # incompatible CLI and configured specs
-        specs_to_add = MatchSpec("python=2.7"),
+        specs_to_add = MatchSpec("scikit-learn==0.13"),
```
msarahan: Changed to scikit-learn here because the thing we need to test here is whether an indirect conflict with a configured spec will raise.  If you provide an explicit spec, we override any pins.  This is new behavior in this PR.

Elsewhere in index.json, you'll see that all python 2.6 builds of scikit-learn 0.13 have been removed.  That's for this test.  python is the conflict here, but it has to be indirect.

tests/core/test_solve.py
```diff
@@ -1419,29 +1410,28 @@ def test_pinned_1():
             )
             assert convert_to_dist_str(final_state_5) == order
 
-    # # TODO: re-enable when UPDATE_ALL gets prune behavior again, following completion of https://github.com/conda/constructor/issues/138
```
msarahan: This is just re-enabled because we're now effectively pruning by default.

tests/core/test_solve.py
```diff
@@ -1483,6 +1473,12 @@ def test_no_update_deps_1():  # i.e. FREEZE_DEPS
         assert convert_to_dist_str(final_state_2) == order
 
     specs_to_add = MatchSpec("zope.interface>4.1"),
+    with get_solver(specs_to_add, prefix_records=final_state_1, history_specs=specs) as solver:
```
msarahan: same as https://github.com/conda/conda/pull/8741/files#r291599132

tests/core/test_solve.py
```diff
@@ -1576,6 +1572,23 @@ def test_timestamps_1():
         )
         assert convert_to_dist_str(link_dists) == order
 
+def test_channel_priority_churn_minimized():
```
msarahan: This is a regression test for churn observed with 4.7.1.  If you started with just defaults, did an install with ``-c conda-forge``, you'd update to a lot of packages.  On the next operation, it would go back to defaults.

tests/core/test_solve.py
```diff
@@ -1829,7 +1842,13 @@ def test_freeze_deps_1():
         assert convert_to_dist_str(link_precs) == link_order
 
     # here, the python=3.4 spec can't be satisfied, so it's dropped, and we go back to py27
-    specs_to_add = MatchSpec("bokeh=0.12.5"),
+    with pytest.raises(UnsatisfiableError):
```
msarahan: see https://github.com/conda/conda/pull/8741/files#r291599132

tests/data/index.json
```diff
@@ -17743,78 +17743,6 @@
     ],
     "version": "0.12.1"
   },
-  "scikit-learn-0.13-np15py26_0.tar.bz2": {
```
msarahan: see https://github.com/conda/conda/pull/8741/files#r291600548

conda/__init__.py
```diff
@@ -16,6 +16,24 @@
...
+    # we have openssl 1.0.2 and 1.1.1 filenames here.  If they're not found, it doesn't matter.
+    libraries = ("archive.dll",
+                 "libcrypto-1_1-x64.dll", "libssl-1_1-x64.dll",
+                 "libcrypto-1_1.dll", "libssl-1_1.dll",   # x86
+                 "libeay32.dll", "ssleay32.dll"
+    )
+    for library in libraries:
+        lib_path = join(sys.prefix, "Library", "bin", library)
+        if os.path.exists(lib_path):
+            cdll.LoadLibrary(lib_path)
```
jjhelmus: Should we use a try/except here in case the `LoadLibrary` fails?  If these files were corrupt conda would be in a pickle so failing early might be reasonable behavior.

msarahan: Good point. I think if these files are hosed, conda is not going to be doing much one way or another. Maybe this is a good place for us to provide a better error message for library load failures than people would otherwise see down the line. Libarchive has an especially cryptic message. Easy enough to discern from the traceback, but yeah, who reads those?

conda/cli/install.py
```diff
@@ -242,68 +245,61 @@ def install(args, parser, command='install'):
...
-                update_modifier = UpdateModifier.FREEZE_INSTALLED
-            if isupdate:
-                deps_modifier = context.deps_modifier or DepsModifier.UPDATE_SPECS
+    repodata_fns = args.repodata_fns
+    if not repodata_fns:
+        repodata_fns = ["current_repodata.json", "repodata.json"]
+    elif REPODATA_FN not in repodata_fns:
+        repodata_fns.append(REPODATA_FN)
+
+    for repodata_fn in repodata_fns:
```
soapy1: Why does this check multiple repodatas to begin with opposed to using one and then falling back to regular repodata.json?

msarahan: It doesn't.  Each iteration of this loop runs the install process.  The loop exits after the first one succeeds.  None of them are checked prior to entering this loop.

https://github.com/conda/conda/pull/8741/files/84f93dfeda2143058ad93e3295b352de26372f15#diff-34bb8d01f991a5ffabae5e80b987c83fR283

conda/__init__.py
msarahan: I removed this.  I hoped it might fix win-32, but it didn't.  It's a nasty hack, and we shouldn't have it if it isn't helping.

msarahan:
@nehaljwani I remembered your request a bit after merging this. Sorry I didn't get to  it.  here's the test I added to for you:

```
    def test_create_only_deps_flag(self):
        with make_temp_env("python=2", "flask", "--only-deps") as prefix:
            assert not package_is_installed(prefix, 'flask')
            assert package_is_installed(prefix, 'python')
            if not on_win:
                # sqlite is a dependency of Python on all platforms
                assert package_is_installed(prefix, 'sqlite')
            assert package_is_installed(prefix, 'itsdangerous')

            # test that a later install keeps the --only-deps packages around
            run_command(Commands.INSTALL, prefix, "imagesize")
            assert package_is_installed(prefix, 'itsdangerous')
            assert not package_is_installed(prefix, 'flask')
```

This stuff works like this:
1. --only-deps leaves packages effectively out of the history and not associated with any history spec.  This behavior is something you can emulate pretty well by simply removing the history file.
2. at https://github.com/conda/conda/blob/master/conda/core/solve.py#L492, we have 
```
ipdb> ssc.specs_map
OrderedDict([('python', MatchSpec("pkgs/main/linux-64::python==2.7.16=h9bab390_0")), ('ca-certificates', 'ca-certificates'), ('openssl', 'openssl'), ('certifi', 'certifi')])
```
none of the --only-deps installed stuff is there yet.
3. Everything gets pinned down here: https://github.com/conda/conda/blob/master/conda/core/solve.py#L504 - this is what is enforcing conda's "least change behavior".

This behavior holds only for the "install" command.  I'm adding tests for "update" and "remove" commands, because we need them to not have unexpected side effects.

conda/core/solve.py
```diff
@@ -417,23 +479,37 @@ def _add_specs(self, ssc):
...
+        for s in ssc.pinned_specs:
+            if s.name in explicit_pool:
+                if s.name not in self.specs_to_add_names:
+                    ssc.specs_map[s.name] = s
+                elif explicit_pool[s.name] & self._get_package_pool(ssc, [s])[s.name]:
+                    ssc.specs_map[s.name] = MatchSpec(s, optional=False)
+                    pin_overrides.add(s.name)
+                else:
+                    log.warn("pinned spec %s conflicts with explicit specs.  "
+                             "Overriding pinned spec.", s)
```
yhuai: Can I ask what is the reason to allow user-specified version to override pinned spec? According to the doc (https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html?highlight=pinned#preventing-packages-from-updating-pinning), the pinned file can prevents packages listed in the pinned file from being updated. This seems like a major behavioral change and I am not sure it is documented on the release notes.

We have multiple use cases that rely on pinned file to prevent accidental version change. Is there a workaround to fallback to the old behavior?

yhuai: I created to https://github.com/conda/conda/issues/9016 to discuss this issue.

PR #8750, opened 2019-06-03
Added --force option for deletion of files via rsync (Jira CON-529)

katietz:


msarahan:
Looks right to me.  Test fails are related to the file format, not this PR.  They are looking for .tar.bz2 files, but there are .conda files there instead now.  I'll fix that separately.

PR #8756, opened 2019-06-05
Fix python_api.rst page title

csosborn:
It was referring to nonexistent "conda.api.python_api" module.


msarahan:
Thanks!  good catch.

PR #8759, opened 2019-06-07
conda.models.dist: Add support for conda pkg format v2

nehaljwani:
This is required for constructor to support the new .conda format


msarahan:
nice!  I just hit this myself.  The fix I came up with is in https://github.com/conda/conda/pull/8741/commits/84f93dfeda2143058ad93e3295b352de26372f15

I didn't find it necessary to strip the ext in multiple places.  Just the one from_string, because that's the method that calls the others later, after stripping.  Since my code is more compact, I think we should go with it.  Do you agree?

nehaljwani:
Agreed.

PR #8471, opened 2019-04-01
Fix more JSON encode errors for the _Null data type

mingwandroid:

PR #8761, opened 2019-06-09
#8471 continue: test of keeping old solver-based unsat, vs newer reduced_index

msarahan:
duplicate of #8741 but keeping the old unsat implementation that uses the solver, rather than the rework based on get_reduced_index alone.  I was curious about any speed difference.  It looks like the other PR is faster on the unit tests (52 s vs 57 s), but this PR is faster on the integration tests (672 s vs. 543 s).

Since the integration tests are far fewer, I think the difference there is more significant (averaged out less), and also because the integration tests are a better representation of actual use.  I think we should revert that bit in the other PR, or get the tests passing on this PR and merge it instead.


msarahan:
Apparently not?  I must not have had the packages cached.  Same machine, I now get 528 s with #8741.  Closing this.

PR #8769, opened 2019-06-11
target prefix override for entrypoints

jjhelmus:
Target prefix override applies to entry points as well as replacements
of standard files.


jjhelmus:
This is a follow up to #6413 which added target prefix override to "normal" files.  This adds support for overriding the target prefix to entry points.  This feature is useful for creating snap packages which are run from a path which not available at build time.

PR #8776, opened 2019-06-13
Locking for activate and deactivate scripts [WIP]

johanneskoester:
Once ready, this should solve https://github.com/conda-forge/r-base-feedstock/issues/67.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,6 +12,10 @@ main_test: &main_test
   <<: *defaults
   steps:
     - checkout
+    # TODO remove the next again once the dockerfile is updated
+    - run:
+        name: temporal dependencies
+        command: pip install flufl.lock
     - run:
         name: unit tests
         command: |
--- a/circle.yml
+++ b/circle.yml
@@ -12,10 +12,6 @@ main_test: &main_test
   <<: *defaults
   steps:
     - checkout
-    # TODO remove the next again once the dockerfile is updated
-    - run:
-        name: temporal dependencies
-        command: pip install flufl.lock
     - run:
         name: unit tests
         command: |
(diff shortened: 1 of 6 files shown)


johanneskoester:
Missing dependencies are added to conda-forge now: https://github.com/conda-forge/staged-recipes/pull/8552

kalefranz:
1. What is the performance impact? `conda activate` needs to be VERY fast because it's typically executed for every new terminal a user opens. Anything over about 500 ms is a total non-starter for most users.
2. Not only does this introduce a new external dependency, it's a notoriously fragile namespace dependency. In general I'm a fan of the `flufl.lock` library, but from my recollection it needs to be evaluated for Windows systems, and it should be vendored if moving forward in this direction.

kalefranz:
Is there any way to accomplish the same goal without affecting ALL users, whether they actually need the feature or not. In general activation shouldn't need to have side effects like writing to disk, and therefore there shouldn't be any concurrency issues. The general case is explicitly designed that way. However certain activation scripts do have side effects. Maybe the locking is best handled on a case-by-case basis for those specific packages that actually need it? Seems unfair to impart that cost on all users whether they need to feature or not.

mingwandroid:
Please stop talking about performance until you have correctness addressed.

msarahan:
As I mentioned here:

https://github.com/conda-forge/r-base-feedstock/issues/67#issuecomment-502523856

I think that Ray is correct here.  Where there are activation scripts that execute arbitrary shell scripts, we must default to safety (locking).  We can and should provide a way for packages to indicate that they are safe, though.  Having the JSON activation would go a long way towards safety by default, but there's manual metadata we can add before then.

travishathaway:

PR #8785, opened 2019-06-15
4.5.x

cristoduloc:
not work for windows 7 ultimate 32 bit

https://foram-puturi.ro


jjhelmus:
This appears to be spam with a link to a non-conda related website.  @cristoduloc if this is a real pull request please provide a description of what these changes do.

Issue #8777, opened 2019-06-13
conda-env: no json output?


elibarzilay:
## Current Behavior
`conda env --json` doesn't produce json output, it just ignores it and produces yaml results.

### Steps to Reproduce

    conda env export --json --name base

## Expected Behavior

Expected JSON output.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : /home/nbuser/conda
            shell level : 1
       user config file : /home/nbuser/.condarc
 populated config files : 
          conda version : 4.6.14
    conda-build version : not installed
         python version : 3.7.3.final.0
       base environment : /home/nbuser/conda  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/nbuser/conda/pkgs
                          /home/nbuser/.conda/pkgs
       envs directories : /home/nbuser/conda/envs
                          /home/nbuser/.conda/envs
               platform : linux-64
             user-agent : conda/4.6.14 requests/2.21.0 CPython/3.7.3 Linux/4.15.0-1035-azure ubuntu/18.04.2 glibc/2.27
                UID:GID : 1200:1200
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```
(nothing printed)
```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

```
# packages in environment at /home/nbuser/conda:
#
# Name                    Version                   Build  Channel
_tflow_select             2.3.0                       mkl    defaults
absl-py                   0.7.1                    py37_0    defaults
adal                      1.2.1                    pypi_0    pypi
applicationinsights       0.11.9                   pypi_0    pypi
asn1crypto                0.24.0                   py37_0    defaults
astor                     0.7.1                    py37_0    defaults
attrs                     19.1.0                   py37_1    defaults
...etc...
(cutting off since this output looks wrong too...)
```
</p></details>

msarahan:
fixed by #8790

elibarzilay:
Thanks @msarahan!

PR #8790, opened 2019-06-17
Fix 8777

cjmartian:
This PR adds support for passing `--json` into `conda env export` as requested in https://github.com/conda/conda/issues/8777

### DIFF
--- a/conda_env/cli/main_export.py
+++ b/conda_env/cli/main_export.py
@@ -14,7 +14,6 @@ from .common import get_prefix, stdout_json
 from ..env import from_environment
 from ..exceptions import CondaEnvException
 
-
 description = """
 Export a given environment
 """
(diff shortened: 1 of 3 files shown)


conda_env/cli/main_export.py
```diff
@@ -100,7 +101,7 @@ def execute(args, parser):
         env.add_channels(args.channel)
 
     if args.file is None:
-        print(env.to_yaml())
+        print(json.dumps(env.to_dict())) if args.json else print(env.to_yaml())
```
soapy1: For printing the json there is a function that already exists called `stdout_json`. It will also do a pretty printing of the json, so I think that one should be used. 
The `import json` above can be replaced with `from conda.cli.common import stdout_json`.
And then the `json.dumps(env.to_dict())` can be replaced with `stdout_json(env.to_dict())`

cjmartian: Thank you for that, I made those changes!

msarahan:
Thanks @cjmartian - merging

PR #8791, opened 2019-06-17
Changed api indentation to fix broken page

rrigdon:


rrigdon:
I changed some formatting and it fixed the problem (on my local) where two of the API pages weren't loading. I'm hoping this will fix the live version.

conda/api.py
```diff
@@ -6,8 +6,7 @@
 from .common.constants import NULL
 from .core.package_cache_data import PackageCacheData as _PackageCacheData
 from .core.prefix_data import PrefixData as _PrefixData
-from .core.solve import (DepsModifier as _DepsModifier, Solver as _Solver,
-                         UpdateModifier as _UpdateModifier)
+from .core.solve import (DepsModifier as _DepsModifier, Solver as _Solver, UpdateModifier as _UpdateModifier)
```
msarahan: Your change will break flake8, unfortunately.  An alternate solution would be to break it up across 2 lines:

```suggestion
from .core.solve import DepsModifier as _DepsModifier, Solver as _Solver
from .core.solve import UpdateModifier as _UpdateModifier
```

jjhelmus:
Update to the `environment.yaml` file in #8795

Issue #8733, opened 2019-05-29
Activating/running environments in parallel nondeterministically fails


Bezier89:
## Current Behavior

When using "conda activate" or "conda run" in parallel, there are nondeterministic failures (it will succeed most of the time). This can either be with the same environment or different environments, doesn't matter. The error manifests with a message like this:

```
The process cannot access the file because it is being used by another process.
The system cannot find the file C:\Users\USERNAME\AppData\Local\Temp\conda-26605.tmp.
C:\Users\USERNAME\AppData\Local\Temp\conda-26605.tmp
The process cannot access the file because it is being used by another process.
'python' is not recognized as an internal or external command, operable program or batch file.
```

## Expected Behavior

Should be able to consistently activate/run arbitrarily many environments in parallel.

## Source of bug

https://github.com/conda/conda/blob/38d07e0a2dcfbe4c883af0f246431e1e073fa904/conda/shell/condabin/_conda_activate.bat#L30-L35

What's happening here is that multiple activates end up using the same "unique" file, and one of them deletes it before the other one is done with it. In the best case, `%RANDOM%` is only 15 bits of randomness, which isn't sufficient when you're using conda environments extensively. It's even worse if you hit [this issue](https://devblogs.microsoft.com/oldnewthing/20100617-00/?p=13673), which makes a simple solution such as using %RANDOM% twice not ideal.

Treiblesschorle:
I'm experiencing the same issue.

johnnybabs:
Hi, I'm having problems with my jupyter notebook. I keeps telling me to activate the conda environment , I've done all I've seen to do online including the conda activate myenv into the terminal, uninstalling and reinstalling. I works perfectly when I run it immediately after installation, but after restart my system it gives me the same activate environment message. I use a windows 10, and I'll attach a screen grab of the error message. Any help whatsoever is greatly appreciated.
![Screenshot (57)](https://user-images.githubusercontent.com/56326038/66697285-10bfce80-eccc-11e9-892b-e2f76c96b4c6.png)

Bezier89:
Dupe of https://github.com/conda/conda/issues/8628 (which should both be resolved now)

PR #8793, opened 2019-06-18
Resolve issue #8733

Bezier89:
Not the most elegant solution, but this should resolve issue #8733 using two approaches:

1. Use %RANDOM% twice for a more "unique" file name.
2. Create the file immediately using `TYPE NUL`. This is much faster than letting the conda command create the file, so less likely to have a race condition with the `@IF EXIST "%UNIQUE%" goto :tmpName` line.


Bezier89:
@msarahan Not sure why the check is saying I haven't signed the CLA. I've signed it already twice before.

msarahan:
Just ignore the CLA thing.  You're definitely in there, and I'm not sure why it isn't showing up: https://github.com/conda/conda/blob/master/.cla-signers#L29

Issue #8803, opened 2019-06-20
[4.7.3] Not backward-compatible with internal repos


dhirschfeld:
Having upgraded to 4.7.3 `conda` can no longer pull packages from our internal repo:

![image](https://user-images.githubusercontent.com/881019/59843497-0768ca00-939c-11e9-88d9-6185dacf7b76.png)

Our internal repo has `repodata.json` and `repodata2.json` but no `current_repodata.json`.

It would be nice if the newer conda would fall back to using one of the older formats if `current_repodata.json` didn't exist.

dhirschfeld:
I guess I just need to re-index with a newer conda-build? I'm hesitant to do so as I need the repo to also work for conda 4.6.8 - it's only the dev environment where I'm testing out the latest conda.

dhirschfeld:
Turns out indexing with the latest conda/build had some nasty issues:
```
> conda index
Subdir: linux-64:   0%|                                                                                                                                      | 0/3 [00:00<?, ?it/s]
Gathering repodata:   0%|                                                                                                                                    | 0/9 [00:00<?, ?it/s]
File format::   0%|                                                                                                                                          | 0/2 [00:00<?, ?it/s]
hash & extract packages for linux-64:   0%|                                                                                                                  | 0/3 [00:00<?, ?it/s]Exception in thread QueueManagerThread:
Traceback (most recent call last):
  File "C:\Miniconda3\lib\threading.py", line 917, in _bootstrap_inner
    self.run()
  File "C:\Miniconda3\lib\threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Miniconda3\lib\concurrent\futures\process.py", line 354, in _queue_management_worker
    ready = wait(readers + worker_sentinels)
  File "C:\Miniconda3\lib\multiprocessing\connection.py", line 868, in wait
    ready_handles = _exhaustive_wait(waithandle_to_obj.keys(), timeout)
  File "C:\Miniconda3\lib\multiprocessing\connection.py", line 800, in _exhaustive_wait
    res = _winapi.WaitForMultipleObjects(L, False, timeout)
ValueError: need at most 63 handles, got a sequence of length 65
```
```
> conda list conda*
# packages in environment at C:\Miniconda3:
#
# Name                    Version                   Build  Channel
conda                     4.7.4                    py37_0    https://conda.anaconda.org/conda-canary
conda-build               3.18.2                   py37_0    https://conda.anaconda.org/conda-forge
[...]

PR #8808, opened 2019-06-20
Use less alarming wording in informational message

dhirschfeld:


dhirschfeld:
This might help prevent future false positives such as in https://github.com/conda/conda/issues/8803 but feel free to close if you don't think it's a worthwhile change.

msarahan:
Thanks.  Hopefully this will help people be less concerned.

PR #8820, opened 2019-06-24
handle JSONDecodeError better when package is corrupted at extract time

msarahan:
Handles corrupt packages better, and helps people understand that this is a problem with the upstream source of the package, not their local conda.

Thanks @TomAugspurger for raising this.


TomAugspurger:
👍. This was the final issue we were saw for pandas CI: https://github.com/pandas-dev/pandas/pull/26595

PR #8827, opened 2019-06-26
Fix the checkout step in the Windows dev env setup instruction

twinssbc:
The Windows dev env setup steps are not in correct order.


msarahan:
Thanks! merging

Issue #7171, opened 2018-04-16
cannot create environment in offline mode create conda create -n py35 python=3.5 --offline


hisi91:
**I'm submitting a...**
  - [x] bug report
  - [ ] feature request

## Current Behavior

Preparing transaction: failed when I proceed 

### Steps to Reproduce

```
# conda create -n py35 python=3.5 --offline
Solving environment: done

## Package Plan ##

  environment location: /opt/Anaconda/Anaconda3-5.1.0/envs/py35

  added / updated specs:
    - python=3.5


The following NEW packages will be INSTALLED:

    ca-certificates: 2017.08.26-h1d4fec5_0
    libedit:         3.1-heed3624_0
    libffi:          3.2.1-hd88cf55_4
    libgcc-ng:       7.2.0-h7cc24e2_2
    libstdcxx-ng:    7.2.0-h7a57d05_2
    ncurses:         6.0-h9df7e31_2
    openssl:         1.0.2n-hb7f436b_0
    python:          3.5.5-hc3d631a_1      <unknown>
    readline:        7.0-ha6073c6_4
    sqlite:          3.22.0-h1bed415_0
    tk:              8.6.7-hc745277_3
    xz:              5.2.3-h55aa19d_2
    zlib:            1.2.11-ha838bed_2

Proceed ([y]/n)? y

Preparing transaction: failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/opt/Anaconda/Anaconda3-5.1.0/lib/python3.6/site-packages/conda/exceptions.py", line 789, in __call__
        return func(*args, **kwargs)
      File "/opt/Anaconda/Anaconda3-5.1.0/lib/python3.6/site-packages/conda/cli/main.py", line 78, in _main
        exit_code = do_call(args, p)
      File "/opt/Anaconda/Anaconda3-5.1.0/lib/python3.6/site-packages/conda/cli/conda_argparse.py", line 77, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/opt/Anaconda/Anaconda3-5.1.0/lib/python3.6/site-packages/conda/cli/main_create.py", line 11, in execute
        install(args, parser, 'create')
      File "/opt/Anaconda/Anaconda3-5.1.0/lib/python3.6/site-packages/conda/cli/install.py", line 255, in install
[...]

PR #8837, opened 2019-06-27
Handle None url

mcopes73:
Fixes https://github.com/conda/conda/issues/7171 (@kalefranz)


mcopes73:
Fixed in https://github.com/conda/conda/pull/8868

PR #8874, opened 2019-07-05
Update create-custom-channels.rst

richardjgowers:
I was following this page and got this warning:

`The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.`

So looks like the docs need updating


msarahan:
Thanks @richardjgowers

PR #8881, opened 2019-07-08
move tests to cdn-based channels

forrestwaters:


tests/test_create.py
```diff
@@ -2195,13 +2195,13 @@ def test_anaconda_token_with_private_package(self):
         # Step 2. Now with the token make sure we can see the anyjson package
         try:
             prefix = make_temp_prefix(str(uuid4())[:7])
-            channel_url = "https://conda.anaconda.org/t/zlZvSlMGN7CB/kalefranz"
+            channel_url = "https://conda.anaconda.org/t/co-3f5c68c1-4eec-415a-8131-e1cb93a8d7b3/conda-test"
```
msarahan: I think tokens don't work with the CDN, so we'll have to revert these changes, or perhaps work with @soapy1 to tweak the CDN so that it passes through any token-bearing requests (URL or otherwise) to conda-web.anaconda.org.

PR #8893, opened 2019-07-09
Adding performance and data science concepts

rrigdon:
Need review for accuracy/missing content.


rrigdon:
I found more content that should be added in so this is not yet ready for full review or merging.

rrigdon:
Ready for review.

docs/source/user-guide/concepts/channels.rst
```diff
@@ -23,8 +23,8 @@ For details, see how to :ref:`modify your channel lists <config-channels>`.
 
 We use conda-forge as an example channel.
 `Conda-forge <https://conda-forge.org/>`_ is a community channel
-made up of thousands of contributors. Conda-forge itself is analogous to PyPI
```
msarahan: The comparison to PyPI is important.  Does it need more detail?  What would it take to make it a better part of this section?

docs/source/user-guide/concepts/conda-performance.rst
```diff
@@ -0,0 +1,106 @@
...
+* Is the channel metadata sane?
+* Are channels interacting in bad ways?
+
+
+
+Improving conda performance
+===========================
+
+To address these challenges, you can move packages to archive
+channels and follow the methods below to present a reduced view of
```
msarahan:
```suggestion
channels and follow the methods below to present conda with a smaller, simpler view than
```

docs/source/user-guide/concepts/conda-performance.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Reduce the index
+----------------
+One option for speeding up conda is to reduce the index. It's likely that
+your repodata contains package data that is not used in the solving stage.
+Filtering out unnecessary packages before solving can save time.
+Reducing the index takes specs and reduces the problem size so that the
+SAT doesn’t have as much work to do.
+
+Additionally, it is beneficial to be as specific as possible in your package
```
msarahan: Reducing the index is something that conda does based on the user's input specs.  Making your inputs more specific makes this index reduction more effective, and thus makes the process faster.  Users alone do not reduce the index, but they should understand why it's important to think about specs.

docs/source/user-guide/concepts/data-science.rst
```diff
@@ -0,0 +1,27 @@
...
+Conda for data scientists
+=========================
+
+Conda is useful for any packaging process but it stands out from other
+package and environment management systems through its utility for data
+science.
+
+Conda’s benefits include:
+
+* Facilitating reproducibility of research workflows.
```
msarahan: Remove this, keep the similar line below about "allowing you to provide your environment..."

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+
+Virtual environments
+====================
+
+A virtual environment is a tool that helps to
+keep dependencies required by different projects
+separate by creating isolated Python virtual environments
+for them.
+
+Users can select a Python virtual environment, such as
```
msarahan: Users can create virtual environments using one of several tools...

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+Pipenv or Poetry, or a conda virtual
+environment. Some characteristics of the two:
```
msarahan:
```suggestion
environment. Pipenv and Poetry are based around Python's built-in venv library, whereas conda has its own notion of virtual environments that is lower-level (python itself is a dependency provided in conda environments).  Some other traits are:
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+     -
+     - Manages non-Python dependencies (R, Perl,
+       arbitrary executables)
+   * - **Tracking dependencies**
+     -
+     - Tracks binary dependencies explicitly
+
+|
+
+Why use Python virtual environments
```
msarahan:
```suggestion
Why use venv-based virtual environments
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+===================================
+- You prefer their workflow or spec formats
+- You prefer to use the system Python and libraries
+- You prefer packages that come more directly from the project maintainers
```
msarahan:
```suggestion
- Ryour project maintainers only publish to PyPI, and you prefer packages that come more directly from the project maintainers, rather than someone else providing builds based on the same code
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+
+Why use conda virtual environments?
+===================================
+
+- You want control over binary compatibility choices
+- You want to utilize newer language standards
```
msarahan:
```suggestion
- You want to utilize newer language standards, such as C++ 17
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+Workflow differentiators
+========================
+
+Some questions to consider as you determine your preferred
+workflow and virtual environment: 
+
+- Is your environment shared across multiple code projects?
+- Does your environment live alongside your code or in a separate place?
+- Do your install steps involve installing any external libraries?
+- Do you want to ship your environment as a speckle of some sort
```
msarahan:
```suggestion
- Do you want to ship your environment as an archive of some sort containing the actual files of the environment?
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,85 @@ environment:
...
+  of a big blob of actual stuff?
+- Is a dependency solver a good thing or a bad thing?
```
msarahan: This is cheeky.  The point here is that right now, pip is often carelessly clobbering things.  That makes the newest thing installed work, but breaks older things.  If you use a solver, on the other hand, then you may not even be able to install the new thing because the solver knows that the new and old things are incompatible.  Some people rely on pip's breaking behavior, and there a dependency solver is a "bad thing," where really the correct idea is that solvers are good things, and they make sure we keep our environments all happy and hunky-dory.

docs/source/user-guide/concepts/channels.rst
rrigdon: It can stay in. We have a reference to conda-forge and PyPI on the main landing page but it is probably important for it to be elaborated upon here.

docs/source/user-guide/concepts/environments.rst
rrigdon: I removed that bullet point just to keep it clean.

docs/source/user-guide/concepts/channels.rst
rrigdon: I put the language back in there. Our readers probably understand the importance of that statement better than I did.

docs/source/user-guide/concepts/environments.rst
rrigdon: I removed that bullet point.

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,90 @@ environment:
...
-as the default environment.
\ No newline at end of file
+as the default environment.
+
+Virtual environments
+====================
+
+A virtual environment is a tool that helps to
+keep dependencies required by different projects
+separate by creating isolated Python virtual environments
```
msarahan:
```suggestion
separate by creating isolated spaces for them that contain per-project dependencies.
```

docs/source/user-guide/concepts/environments.rst
```diff
@@ -52,4 +52,90 @@ environment:
...
+       or use apt/yum/brew/etc.
+     - Install system-level libraries as conda dependencies
+   * - **System**
+     - Depend on base system install of Python
+     - Python is independent from system
+   * - **Extending environment**
+     - Extend environment with pip
+     - Extended environment with conda or pip
+   * - **Non-Python dependencies**
+     -
```
msarahan: extra dash here?

msarahan: oh wait, sorry, it's because there is no pip equivalent.

rrigdon: So no changes?

msarahan: Yeah, no changes here.

PR #8895, opened 2019-07-10
add additional debugging info to Resolve.solve

jjhelmus:
Include target and optional information in list of specs which are
solved for.

Include the requested, optional, missing and all other specs in the
debugging information.


jjhelmus:
AppVeyor is timing out not a blocker.

PR #8899, opened 2019-07-10
Hint no update

soapy1:
This will give a message to the user when their update is constrained by something. For example:

ipython is installed by anaconda
```
$ conda update ipython ncurses --quiet                                                                           
Solving environment: ...working... 

Updating ipython is constricted by 

anaconda -> requires ipython==7.2.0=py37h39e3cac_0

If you are sure you want an update of your package either try `conda update --all` or install a specific version of the package you want using `conda install <pkg>=<version>`

done

# All requested packages already installed.
```

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -258,8 +258,7 @@ class Solver(object):
                     pre_packages, post_packages, ssc.index.keys())
                 if len(constrained) > 0:
                     for spec in constrained:
-                        pkg_pool = ssc.r._get_package_pool([i.to_match_spec() for i in ssc.solution_precs])
-                        self.determine_constricting_specs(spec, ssc.solution_precs, pkg_pool[spec.name])
+                        self.determine_constricting_specs(spec, ssc.solution_precs)
 
             # if there were any conflicts, we need to add their orphaned deps back in
             if ssc.add_back_map:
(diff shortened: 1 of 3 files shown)


conda/core/solve.py
```diff
@@ -239,14 +239,21 @@ def solve_final_state(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL
                      context.json):
             ssc = self._collect_all_metadata(ssc)
 
-        with Spinner("Solving environment", not context.verbosity and not context.quiet,
+        with Spinner("Solving environment", context.verbosity and not context.quiet,
```
jjhelmus: `not` should be retained

conda/core/solve.py
```diff
@@ -266,6 +273,53 @@ def solve_final_state(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL
...
+        for pkg in self.specs_to_add:
+            current_version = max(i[1] for i in pre_packages[pkg.name])
+            if current_version == max(i.version for i in index_keys if i.name == pkg.name):
+                continue
+            else:
+                if post_packages == pre_packages:
+                    for pkg_name, matches in post_packages.items():
+                        pkg_match = [m for m in matches if m[1] is not None][0]
+                        print('\n\nTrying to update package {update_pkg} however it seems like '
+                              'you already have this package constrained to \n    {const_pkg} '
```
msarahan: This is a great start.  I think one critical detail is missing, though.  You report a version that is the result of one or more constraints, but the thing we really need most is to understand the constraint itself, and where it comes from.  In order to do this, you'll need to look through all of the specs that factored into the solution, and find which ones have constraints (directly, or perhaps buried deeper in the dependency tree) that affect this package.

conda/core/solve.py
```diff
@@ -268,6 +279,73 @@ def solve_final_state(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL
 
         return ssc.solution_precs
 
+    def determine_constricting_specs(self, spec, solution_precs):
+        highest_version = [VersionOrder(sp.version) for sp in solution_precs
+                           if sp.name == spec.name][0]
+        constricting = []
+        for prec in solution_precs:
```
soapy1: I think I could have made use of `_get_package_pool` here but would have needed to loop through everything again to split out specs with compound version specs. I figure going through once is probably better then doing so twice

msarahan:
Great improvement!  Thank you!

PR #8909, opened 2019-07-11
Fix issue where python version can't be changed in base env with noarch packages

### LINKED ISSUE #8908: Can't downgrade python: "No module named 'distutils.spawn'"
Issue (conda-forge issue: https://github.com/conda-forge/conda-forge.github.io/issues/805)

I'm not sure whether it is a CF or conda issue, but it only occurs after running conda update --all with conda-forge channel added.

Here is a Dockerfile to reproduce:

```
FROM continuumio/miniconda3

RUN conda config --add channels conda-forge && \
    conda config --set channel_priority strict && \
    conda update --all

RUN conda install python=2.7
```

Fails with:

```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
ERROR conda.core.link:_execute(637): An error occurred while installing package 'conda-forge::ipaddress-1.0.22-py_1'.
ModuleNotFoundError("No module named 'distutils.spawn'")
Attempting to roll back.

Rolling back transaction: ...working... done

ModuleNotFoundError("No module named 'distutils.spawn'")
()
```
The failing package varies based on the python version downgraded to, with python=3.6:
```
ERROR conda.core.link:_execute(637): An error occurred while installing package 'conda-forge::tqdm-4.32.2-py_0'.
ModuleNotFoundError("No module named 'distutils.spawn'")
Attempting to roll back.

Rolling back transaction: ...working... done

ModuleNotFoundError("No module named 'distutils.spawn'")
()
```

Tried both with and without conda config --set channel_priority strict, same outcome.

forrestwaters:
Fixes #8908 

When changing python version, compile_multiple_pyc() is run for noarch packages which ultimately calls wrap_subprocess_call() after packages (setuptools) have been unlinked/linked and are no longer importable. This change works because shutil is included in python.

### DIFF
--- a/conda/common/path.py
+++ b/conda/common/path.py
@@ -14,6 +14,7 @@ from .compat import on_win, string_types
 from .. import CondaError
 from .._vendor.auxlib.decorators import memoize
 from .._vendor.toolz import accumulate, concat, take
+from distutils.spawn import find_executable
 
 try:
     # Python 3
@@ -308,7 +309,6 @@ def win_path_to_unix(path, root_prefix=""):
 
 
 def which(executable):
-    from distutils.spawn import find_executable
     return find_executable(executable)
 
 
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -12,7 +12,7 @@ from ._vendor.auxlib.decorators import memoize
 from ._vendor.auxlib.compat import shlex_split_unicode, string_types, Utf8NamedTemporaryFile
 from .common.compat import on_win, isiterable
 
-from .common.path import win_path_to_unix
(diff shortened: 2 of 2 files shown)


msarahan:
Dang, looks like py27 doesn't have that in shutil.

forrestwaters:
Which is probably why it was using distutils.spawn to begin with 😄.. Moving the import outside of the `which` function also seems to fix this since it ends up being imported at some point prior to the deletion of packages..

kszucs:
FYI The same issue occurred when trying to downgrade to py36 from py37.

PR #8910, opened 2019-07-11
make unsat message more readable by grouping

msarahan:


conda/models/version.py
```diff
@@ -637,6 +645,10 @@ def merge(self, other):
                              % (self.raw_value, other.raw_value))
         return self.raw_value
 
+    def union(self, other):
```
soapy1: Looks like this union function is defined a couple times. Can if be extracted into a utils module or something and shared?

msarahan: yep.  I wouldn't bother reading too much into this code right now, just wanted to share the ideas, since they overlap with what you're doing.

conda/core/solve.py
```diff
@@ -627,11 +661,12 @@ def _add_specs(self, ssc):
 
                 spec_set = (python_spec, ) + tuple(self.specs_to_add)
                 if ssc.r.get_conflicting_specs(spec_set):
-                    if self._repodata_fn == REPODATA_FN:
+                    if (self._repodata_fn == REPODATA_FN and
+                            ssc.update_modifier == UpdateModifier.UPDATE_SPECS):
```
jjhelmus: If the user passes a different update_modifier on the command line then this will never be triggered and a conflict list is never generated.

conda/core/solve.py
```diff
@@ -607,12 +626,27 @@ def _add_specs(self, ssc):
...
 
+        if ssc.update_modifier == UpdateModifier.UPDATE_SPECS:
+            for spec in self.specs_to_add:
+                spec = MatchSpec(spec)
+                if (spec.name in pin_overrides and not ssc.ignore_pinned
+                        or spec.name in ssc.specs_from_history_map):
+                    # skip this spec, because it is constrained by pins
+                    continue
+                # the index is sorted, so the first record here gives us what we want.
+                latest_pkg = ssc.r.find_matches(spec)[0]
```
jjhelmus: There is no certainty that the first record here is what will be or even should be installed.  If a particular variant sorts before the variant that will be installed then many specs are un-necessarily neutered.  For example in an environment with openssl 1.1.1 if a package has an openssl 1.1.1 and 1.0.2 variant if the 1.0.2 variant had sorted before the 1.1.1 variant then openssl would neutered for no reason.

conda/cli/install.py
```diff
@@ -295,11 +294,12 @@ def install(args, parser, command='install'):
                 raise PackagesNotFoundError(e._formatted_chains, channels_urls)
 
         except (UnsatisfiableError, SystemExit, SpecsConfigurationConflictError) as e:
-            # Quick solve with frozen env failed.  Try again without that.
-            if isinstall and args.update_modifier == NULL:
+            # Quick solve with frozen env or trimmed repodata failed.  Try again without that.
+            if not hasattr(args, 'update_modifier') or args.update_modifier == NULL:
```
jjhelmus: Removing isinstalled leads to an extra solve when creating an environment with Unsatisfiable specs:

```
$ conda create -n test python=2.7 ipython=7
Solving environment: failed with current_repodata.json, will retry with next repodata source.
Initial quick solve with frozen env failed.  Unfreezing env and trying again.
Solving environment: failed with current_repodata.json, will retry with next repodata source.
Solving environment: failed
Initial quick solve with frozen env failed.  Unfreezing env and trying again.
Solving environment: failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:



Package python conflicts for:
ipython=7 -> prompt_toolkit[version='>=2.0.0'] -> pygments -> setuptools -> certifi[version='>=2016.09'] -> python[version='>=3.6,<3.7.0a0,>=3.7,<3.8.0a0']
python=2.7
```
Since there is no environment to freeze the initial quick solve with a frozen environment is un-necessary and wasteful

conda/core/solve.py
jjhelmus: For example:
```
conda create -n py27 python=2.7
...
conda install -n py27 --freeze-installed ipython=7
Solving environment: failed with current_repodata.json, will retry with next repodata source.

UnsatisfiableError: 
```

conda/core/solve.py
jjhelmus: I'm not following why this section is needed at all.  Is there something in the existing neutering logic that was causing issues?

msarahan: The first record is a statement of the newest version available.  It doesn't say anything about satisfiability.  It only relaxes (trivially found) conflicts such that the newest version _might_ be installable.  I think this is OK.

conda/cli/install.py
msarahan: yep, this needs a fix.

conda/core/solve.py
msarahan: yep, needs a fix.  Thanks.

conda/core/solve.py
jjhelmus: Right, I'm following this now.  Rhis happens before the explicitly request specs are added (around line 680) so there is no chance that they will be neutered.

conda/cli/install.py
msarahan: I think 3acf0a72e should fix this

PR #8921, opened 2019-07-12
use map to ensure order of channels when creating index

msarahan:


jjhelmus:
Circle failure is Codecov timeout, not a blocker.

PR #8938, opened 2019-07-15
add prefix pkg to index if channels do not match

### LINKED ISSUE #8302: Similar packages in pkgs/main and pkgs/free are considered the same package
If `pkgs/main` and `pkgs/free` contain a package with the same subdir, name, version, build_number and build [conda considers](https://github.com/conda/conda/blob/4.6.6/conda/models/records.py#L253-L266) these packages identical and will not respect the sub-channel priority (main > free).  

A good example of this is the blas=1.0=mkl packages for the linux-64 platform.

Additionally packages in `pkgs/free` and `pkg/main` with the [same name, version and build](https://github.com/conda/conda/blob/4.6.6/conda/models/records.py#L272-L273) will be considered by the [solver](https://github.com/conda/conda/blob/4.6.6/conda/resolve.py#L612-L620) to be the same even if they have a different build_number.  This occurs with the blas=1.0=openblas packages for the linux-ppc64le platform which makes them un-installable.

cf: #8301 #8236

jjhelmus:
Add the local package to the index with the a canonical channel set to
the URL if the channel information of the local packages does not match
that of the remote package.

Fixes cases where each channel in a multi-channel has a package with the
same subdir, name, version and build_number.

closes #8302

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -90,14 +90,26 @@ def _supplement_index_with_prefix(index, prefix):
     assert prefix
     for prefix_record in PrefixData(prefix).iter_records():
         if prefix_record in index:
-            # The downloaded repodata takes priority, so we do not overwrite.
-            # We do, however, copy the link information so that the solver (i.e. resolve)
-            # knows this package is installed.
             current_record = index[prefix_record]
-            link = prefix_record.get('link') or EMPTY_LINK
-            index[prefix_record] = PrefixRecord.from_objects(
-                current_record, prefix_record, link=link
-            )
+            if current_record.channel != prefix_record.channel:
+                # If the local packages channel information does not agree with
+                # the channel information in the index then they are most
+                # likely referring to different packages.  This can occur if a
+                # multi-channel changes configuration, e.g. defaults with and
+                # without the free channel. In this case we need to fake the
+                # channel data for the existing package.
+                prefix_channel = prefix_record.channel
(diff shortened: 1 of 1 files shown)


jjhelmus:
I'm not certain this is a proper fix for #8302 and the instances where it is needed are rare.  

One case where this can occur is on `linux-64`:

```
CONDA_DEFAULT_CHANNELS="https://repo.anaconda.com/pkgs/free" conda create -y -n blas blas
CONDA_RESTORE_FREE_CHANNEL=1 conda update -q -n blas blas
```

```
Solving environment: ...working... failed with current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed

  - pkgs/free/linux-64::blas==1.0=mkl

Current channels:

  - https://repo.anaconda.com/pkgs/main/linux-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/free/linux-64
  - https://repo.anaconda.com/pkgs/free/noarch
  - https://repo.anaconda.com/pkgs/r/linux-64
  - https://repo.anaconda.com/pkgs/r/noarch

looking for, navigate to

    https://anaconda.org

```

msarahan:
seems like there's a py27 issue with the ``test_conda_pip_interop_conda_editable_package`` test - the "UNLINK" category is present where it shouldn't be.  No idea off the top of my head, unfortunately.  Let me know if you need help debugging.

PR #8942, opened 2019-07-16
If solving with current_repodata.json don't find conflicts

soapy1:
Instead, fallback to using repodata.json

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -25,7 +25,7 @@ from ..common.compat import iteritems, itervalues, odict, text_type
 from ..common.constants import NULL
 from ..common.io import Spinner, dashlist, time_recorder
 from ..common.path import get_major_minor_version, paths_equal
-from ..exceptions import PackagesNotFoundError, SpecsConfigurationConflictError
+from ..exceptions import PackagesNotFoundError, SpecsConfigurationConflictError, UnsatisfiableError
 from ..history import History
 from ..models.channel import Channel
 from ..models.enums import NoarchType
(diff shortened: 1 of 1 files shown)


conda/core/solve.py
```diff
@@ -626,6 +626,8 @@ def _add_specs(self, ssc):
                     python_spec = MatchSpec(python_spec, version=pinned_version)
 
                 spec_set = (python_spec, ) + tuple(self.specs_to_add)
+                if len(conflict_specs) > 0 and self._repodata_fn == "current_repodata.json":
```
msarahan: More general to say "only find conflicts when we're using repodata.json, as it is the ultimate fallback.

```suggestion
                if len(conflict_specs) > 0 and self._repodata_fn != "repodata.json":
```

conda/core/solve.py
```diff
@@ -626,6 +626,8 @@ def _add_specs(self, ssc):
...
+                    return ssc
                 if ssc.r.get_conflicting_specs(spec_set):
```
msarahan: I'm not totally sure here, but we need to think about how the program flows from here. Do we need to raise an exception to jump out to the outer loop in install.py? Exception handling for flow control is a coffee smell. Is there a better way for us to get back to that outer loop?

soapy1: Ya, that makes sense. How about only running `find_conflicts` if using `repodata.json` and then if any conflicting specs are found using non `repodata.json` then raising an `UnsatsifiableError`. Thats what `find_conflicts` will do so the parent function already knows how to deal with that case

PR #8947, opened 2019-07-16
as of conda 4.6, conda activate should be used for all OS'es

forrestwaters:


conda/cli/install.py
```diff
@@ -84,39 +84,16 @@ def clone(src_arg, dst_prefix, json=False, quiet=False, index_args=None):
 
 def print_activate(env_name_or_prefix):  # pragma: no cover
     if not context.quiet and not context.json:
-        if 'CONDA_SHLVL' in os.environ or os.path.split(os.environ.get('SHELL', ''))[-1] == 'fish':
```
msarahan: let's double-check that we can change it for fish.  I'm not sure what the state there is.

forrestwaters: Did some testing on MacOS and fish seems to work with conda activate as well.

PR #8949, opened 2019-07-17
Preload on windows libcrypto and libssl, if required

katietz:


jjhelmus:
Will this cause attempt to load the OpenSSL 1.1.1 DLLs to be loaded in an environment with OpenSSL 1.0.2 installed?

katietz:
yes, the goal is to load crypto/ssl on Windows before some other module
attempts to load it from different place. and to force the use of 1_1.

msarahan:
I don't think "forcing 1_1" is the right way to look at this.  The correct behavior is to preload whatever matches conda - whatever is in the root env with conda.  We need to be smart enough to figure out which is the correct one to preload.  It may be 1.1, or it may be 1.0.

katietz:
Ok, fair. I will add code to probe for exiting version in Library/bin
folder before attempting to probe or pre-load.

conda/gateways/connection/download.py
```diff
@@ -35,6 +39,42 @@ def download(
...
+        libssl_version = '-1_1'
+        libssl_arch = ''
+        if sys.maxsize > 2**32:
+            libssl_arch = '-x64'
+        so_name = libssl_dllname + libssl_version + libssl_arch
+        libssl_path2 = os.path.join(libbin_path, so_name)
+        # if version 1.1 is not found, try to load 1.0
+        if not exists(libssl_path2 + ".dll"):
+            # print("prior not exits? %s\n" % libssl_path2+".dll")
+            libssl_version = '-1_0'
```
msarahan: unfortunately, I don't think this is sufficient.  The openssl 1.0 DLL names were totally different.  Probably best to look at one of the 1.0.2 package contents, but I remember something like libeay32.dll and libssleay32.dll?

katietz: Sorry, you are right. Just searching for difference of names for x64 and the 32-bit variant

conda/gateways/connection/download.py
```diff
@@ -35,6 +39,45 @@ def download(
...
+
+        libssl_path = find_library(so_name)
+        if not libssl_path:
+            libssl_path = libssl_path2
+        # crypto library might exists ...
+        so_name = libcrypto_dllname + libssl_version + libssl_arch
+        libcrypto_path = find_library(so_name)
+        if not libcrypto_path:
+            libcrypto_path = os.path.join(sys.prefix, 'Library', 'bin', so_name)
+        # print("Attempt to load %s lib\n" % libssl_path)
```
msarahan: please clean up or replace these with log.debug calls

mingwandroid:
This PR causes conda to load any random OpenSSL it finds on PATH, even ones that do not match the architecture. We will be removing it soon.

PR #8950, opened 2019-07-17
fix None in comprehension that needs to be iterable

msarahan:
This was found by:

conda create -n py27env python=2.7
conda install -n py27env rasterio -c conda-forge


jjhelmus:
AppVeyor is timing out, not a blocker.

PR #8961, opened 2019-07-19
strict priority: pick top channel with matching spec

### LINKED ISSUE #8960: filter strict channel priority on spec matches
In repo2docker, we want to support installing frozen environments created with `conda env export`, including build string, etc. Pretty often, packages on conda-forge are found to have too-loose dependency pinning, which results in them being marked as broken, moving them from `main` to the `broken` label. This (rightly) prevents naïve installation of the 'broken' builds, but also (wrongly, according to user intent) prevents installation when the exact build of the package is requested. For this reason, repo2docker adds the conda-forge broken label as a lowest-priority channel. This is not an ideal solution, and we only really want to use the broken channel when exact requested specs can only be satisfied by these semi-revoked packages.

Here is what currently happens with strict channel priority when `pkg=version=build_1` has been marked as broken since the call to `conda env export`:

- package `pkg=version=build_1` is requested
- package `pkg` is found on conda-forge, so no other channels are consulted
- UnsatisfiableError

I think it would suit us very well if strict channel priority picked the first channel with a package *that directly matches the requested spec* rather than only filtering on the name:

- package `pkg=version=build_1` is requested
- package `pkg` is found on conda-forge, but doesn't match the spec, move on
- package `pkg=version=build_1` is found on conda-forge/broken, so only conda-forge/broken is considered
- success

At first, I thought this would be too complicated and single-use-case, but I think it generally results in strict channel priority behaving in a less surprising manner as stricter specs are requested (version only pinning or version + build).

Another option for our specific case is to change conda-forge policy to apply hotfixes much more often rather than moving broken, but I think this would be an improvement to strict channel priority in general.

See more discussion [on repo2docker](https://github.com/jupyter/repo2docker/issues/731)

minrk:
rather than only considering package name when determining top priority channel

This seems to work reliably in my tests, but I would not be surprised to find edge cases

In particular, I am less confident when PackageRecord is passed to `_get_strict_priority` rather than a MatchSpec. I strip the channel from the spec, but I don't know if that's right when this happens.

Not sure how to write a unittest, but this works:

```
conda create --strict-channel-priority --dry-run -n x -c conda-forge -c conda-forge/label/broken zeromq=4.3.2=*_0
```

build 0 of zeromq 4.3.2 has been marked as broken on mac and linux, so this fails as unsatisfiable with conda 4.7.8, but succeeds with this PR, pulling in zeromq from broken and other packages from higher-priority conda-forge.

(actual failure is a TypeError determining conflicts, but it *should* be an UnsatisfiableError):

<details>

```
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/exceptions.py", line 1062, in __call__
        return func(*args, **kwargs)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/cli/conda_argparse.py", line 82, in do_call
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/cli/main_create.py", line 37, in execute
        install(args, parser, 'create')
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/cli/install.py", line 256, in install
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/core/solve.py", line 112, in solve_for_transaction
        force_remove, force_reinstall)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/core/solve.py", line 150, in solve_for_diff
        force_remove)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/core/solve.py", line 262, in solve_final_state
        ssc = self._run_sat(ssc)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/common/io.py", line 88, in decorated
        return f(*args, **kwds)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/core/solve.py", line 780, in _run_sat
        repodata_fn=self._repodata_fn)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/common/io.py", line 88, in decorated
        return f(*args, **kwds)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/resolve.py", line 1183, in solve
        self.find_conflicts(specs, specs_to_add, history_specs)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/resolve.py", line 336, in find_conflicts
        bad_deps = self.build_conflict_map(specs, specs_to_add, history_specs)
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/resolve.py", line 425, in build_conflict_map
        deps = set.union(*[set(self.ms_depends_.get(prec)) for prec in precs])
      File "/Users/minrk/conda/lib/python3.7/site-packages/conda/resolve.py", line 425, in <listcomp>
        deps = set.union(*[set(self.ms_depends_.get(prec)) for prec in precs])
    TypeError: 'NoneType' object is not iterable
```

</details>

closes #8960


minrk:
Looking at the tests and their expectations, there is something unexpected in the strict channel enforcement here, at least relative to what I originally thought, but that may not be undesirable.

As it is, the channel is strictly set for a package the first time the spec for that package shows up in the solve. I was originally thinking of this only applying to user-provided specs (cli or environment.yml), but of course it also applies when specs show up in dependencies, which is what's happening in the tests:

1. b is found in channel-1
2. b depends on `c>=2,<3` which is only in channel 2, so `c=1` in channel 1 is ignored

The result is arguably desirable, but what's presumably not desirable is that the strict channel choice can be resolved differently depending on what spec the package shows up in first, e.g.

- `conda install -c channel-1 -c channel-2 b`:
  - b depends on `c>=2,<3`, which is only in channel-2
  - strict channel choice for c is therefore channel-2
- `conda install -c channel-1 -c channel-2 c b`
  - `c` is specified top-level, and `c=1` is found in channel-1
  - strict channel choice for c is therefore channel-2
  - `b` is unsatisfiable because the required `c=2` is not in channel-1

So the main problem with this approach is that a package could come from different channels depending on which spec is seen first in the solver, which should be an implementation detail. This is caused because the cache key is the package name, rather than the spec (first commit here). Keying on the spec results in candidates from multiple channels showing up, which is perhaps preferable, and results in the same solution no matter where a package shows up.

To be clear, all of this only occurs in the event of a specifically requested spec (in dependency or by the user) not being available at all in the top priority channel. Otherwise, there's no change.

msarahan:
I think this behavior change is fine, but it'll have to wait until 4.8.x.  4.7 should hopefully be almost stable with enough bugs worked out to think about what comes next.

minrk:
Based on comments in https://github.com/conda-forge/openmpi-feedstock/issues/44, this PR now only allows dropping to lower-priority channels in the event of exact matches on version and/or build, not ranges. The result is that all previous tests continue to pass, as does an additional strict priority test which pulls in a package from a second-priority channel due to an exact match.

This means that:

    conda install a b

fails. Even `conda install a b c` fails, but

    conda install a b c==1

succeeds. Notably, because only exact matches are considered, it may be surprising that

    conda install a b c=1

will fail, because `c=1` is a range, not exact.

This totally solves my case with repo2docker and pinning, and solves the openmpi case linked above. It also greatly limits the cases where lower-priority channels can be chosen to only where exact pinning occurs, not any ranges.

conda/resolve.py
```diff
@@ -497,12 +525,13 @@ def get_reduced_index(self, explicit_specs, sort_by_exactness=True):
 
         def filter_group(_specs):
             # all _specs should be for the same package name
-            name = next(iter(_specs)).name
+            spec = next(iter(_specs))
+            name = spec.name
             group = self.groups.get(name, ())
 
             # implement strict channel priority
             if group and strict_channel_priority and name not in cp_filter_applied:
```
minrk: On reading and re-reading, this bit is not being handled in a strictly correct way, I think. Since this pass can occur multiple times per package and the key is the package name, but the filter result depends on the spec, only the first spec that gets here will take effect.

The result is that, in the event we have:

- `a` depends on `c==2`
- `a` is in channel-1
- `c 2.0` is only in channel-2
- `c 1.0` is in channel-1
- channel-1 has higher priority than channel-2

The simple cases:

- `conda install a`
- `conda install c==2`

will get `c 2.0` from channel-2, which is the intended change from this PR. Since it was suggested that *ranges* should not slide to lower-priority channels,

- `conda install 'c>=2'` will fail with unsatisfiable

since the exact-version condition is not triggered. Where it gets a little weird is

- `conda install a c>=2`

which fails with unsatisfiable and slightly weird:

```
Package c conflicts for:
a -> c==2
c[version='>=2']
Note that strict channel priority may have removed packages required for satisfiability.
```

This could be surprising because we have `a->c==2.0`, which picks channel-2 for c when only `a` is installed, but since `c>=2` is evaluated first and channel-priority is selected only once per package, it's picking channel-1 for c, then failing unsatisfiably on the `c==2` requirement. So these two requirements are indeed in conflict, but only because they are resulting in different channel selections, which is not communicated in the error message. I'd be okay with that, were it not the case that the exact same requirements entering the solver in a different order would be satisfiable.

I tried a bit to allow the channel-priority pass to run multiple times, which would have to allow picking channel-1 then later picking channel-2 (removing and then re-adding packages in get_reduced_index), but couldn't get it to work since reduce is expected to only decrease the acceptable packages. The result was channel-1 was picked and the solve failed with no candidate for `c==2.0` before the new channel would have been selected after adding the `c==2.0` spec.

I'm not sure it's possible to get this exactly right, or what exactly right really is.

h-vetinari:
Since 4.8.0rc0 released without this: what's the status on this?

travishathaway:

PR #8968, opened 2019-07-19
Align 'Managing Python' with 'Managing packages'

rogererens:
In 'Managing Python', use the same differentiation between pre-4.6 and newer versions of activating an environment as in 'Managing packages'


rrigdon:
LGTM

rrigdon:
Thanks for your contribution!

Issue #8831, opened 2019-06-26
Conda unable to fix inconsistency. Update fails.


abalter:
## Current Behavior

`conda update conda` reports inconsistency, tries to complete update, but fails.

### Steps to Reproduce

```
(base) balter@spectre:~$ conda update conda -y
Collecting package metadata (current_repodata.json): done
Solving environment: \
The environment is inconsistent, please check the package plan carefully
The following packages are causing the inconsistency:

  - conda-forge/noarch::r-essentials==3.5.1=r351_2000
  - conda-forge/noarch::r-flexdashboard==0.5.1.1=r351h6115d3f_1000
  - conda-forge/noarch::r-gistr==0.4.2=r351h6115d3f_1001
  - conda-forge/noarch::r-kableextra==1.1.0=r351h6115d3f_0
  - conda-forge/noarch::r-knitr==1.23=r351h6115d3f_0
  - conda-forge/noarch::r-rbokeh==0.5.0=r351h6115d3f_1001
  - conda-forge/noarch::r-reprex==0.3.0=r351h6115d3f_0
  - conda-forge/noarch::r-rmarkdown==1.12=r351h6115d3f_0
  - conda-forge/noarch::r-tidyverse==1.2.1=r351h6115d3f_1001
  - conda-forge/noarch::r-tinytex==0.13=r351h6115d3f_0                                                                                                                                     done

## Package Plan ##

  environment location: /home/balter/conda

  added / updated specs:
    - conda


The following NEW packages will be INSTALLED:

  cairo              conda-forge/linux-64::cairo-1.16.0-h18b612c_1001
  expat              conda-forge/linux-64::expat-2.2.5-he1b5a44_1003
  glib               conda-forge/linux-64::glib-2.58.3-h6f030ca_1001
  gsl                conda-forge/linux-64::gsl-2.5-h294904e_0
  harfbuzz           conda-forge/linux-64::harfbuzz-2.4.0-h37c48d4_1
  jedi               conda-forge/linux-64::jedi-0.14.0-py37_0
  pango              conda-forge/linux-64::pango-1.40.14-he7ab937_1005
  parso              conda-forge/noarch::parso-0.5.0-py_0
  pixman             conda-forge/linux-64::pixman-0.38.0-h516909a_1003
  prometheus_client  conda-forge/noarch::prometheus_client-0.7.1-py_0
  r-base             conda-forge/linux-64::r-base-3.5.1-hea2b9be_1007
[...]

PR #8988, opened 2019-07-24
Don't add back orphan_precs that already exist

soapy1:
Addresses https://github.com/conda/conda/issues/8831

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -88,7 +88,10 @@ class PrefixData(object):
         return fn + '.json'
 
     def insert(self, prefix_record):
-        assert prefix_record.name not in self._prefix_records
+        assert prefix_record.name not in self._prefix_records, \
+            "Prefix record insertion error: a record with name %s already exists " \
+            "in the prefix. This is a bug in conda. Please report it at " \
+            "https://github.com/conda/conda/issues" % prefix_record.name
 
         prefix_record_json_path = join(self.prefix_path, 'conda-meta',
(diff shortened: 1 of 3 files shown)


msarahan:
What was the symptom here?  It's a good fix, but I'm not sure how the problem manifested itself?

soapy1:
I was able to reproduce the problem when I was updating my packages while alternating between channels. So something like:
```
<fresh install>
$ conda update conda-build -c conda-forge
$ conda update --all
```

resulting in an inconsistent env. As part of conda's recovery from this it removes that prec, does the solve and then adds it back. Except when it was adding it back, it was also adding all of it's dependencies that have already been solved for in the update solve. So, conda was thinking that it was gong to newly install a bunch of packages instead of updating. This caused the record action to try to create some new prefix records for these thought-to-be-newly-but-actually-not package. So this assert https://github.com/conda/conda/blob/master/conda/core/prefix_data.py#L91 was triggered

msarahan:
Ah!  Good hunting!  Can you add a bit to the end of that assert to describe the bad condition, so that we know which assertion is failing better next time?

``assert prefix_record.name not in self._prefix_records, "Prefix record insertion error: a record with name %s already exists in the prefix.  This is a bug in conda. 
 Please report it at https://github.com/conda/conda/issues" % prefix_record.name``

Aside from that, is there a simple enough reproducer of the failure to add a test?

soapy1:
Can do! I'm still trying to figure out a representative test case. I was thinking of calling solve with a ssc that has some add_specs_back_map

PR #8990, opened 2019-07-24
don't call find_conflicts on initial ffs

forrestwaters:
This cuts the rstudio failure time in half (still takes about 8 minutes almost all of which is spent in find_conflicts) and allows the orange3 install to fail in about 30 seconds..


jjhelmus:
One issue with this change is that no conflict message will be reported if `--freeze-installed` is explicitly passed on the command line.  I do not believe that this occurs often since the option has not been recommended but it is good to keep in mind that `FREEZE_INSTALLED` is not always followed by a second attempt at finding a solution.

PR #9006, opened 2019-07-26
Breadth first search for conflicts

soapy1:


conda/resolve.py
```diff
@@ -409,14 +424,33 @@ def build_conflict_map(self, specs, specs_to_add=None, history_specs=None):
             # records_per_name is a completely arbitrary number here.  It is meant to gather more
             # than just one record, to explore the space of dependencies a bit.  Doing all of them
             #    can be an enormous problem, though.  This is hopefully a good compromise.
-            records_per_name = 7
```
msarahan: removing this limit may slow things down a lot.  Have you done any kind of speed test?  Especially with the anaconda metapackage - the more packages involved, the slower this will go.

conda/resolve.py
```diff
@@ -409,14 +424,33 @@ def build_conflict_map(self, specs, specs_to_add=None, history_specs=None):
...
             spec_order = sorted(sdeps_with_dep.keys(),
-                                key=lambda x: list(g.graph_by_name.keys()).index(x.name))
-            for spec in spec_order:
-                # the DFS approach works well when things are actually in the graph
-                bad_deps.extend(g.depth_first_search_by_name(spec, dep, sdeps[spec]))
+                                key=lambda x: list(gg.graph_by_name.keys()).index(x.name))
 
+            for spec in spec_order:
+                allowed_specs = sdeps[spec]
+                ga = GeneralGraph([_ for allowed_pkgs in allowed_specs.values()
```
msarahan: can we use the graph we already have above, but filter it somehow?  Creation of the graph is pretty expensive.

conda/resolve.py
soapy1: Tried this out with anaconda metapackage, yep slows it down a ton!

PR #9014, opened 2019-07-30
fix setuptools being removed due to conda run_constrains

msarahan:
With the `add_pip_as_python_dependency` config parameter set to False, conda was removing setuptools after a `conda update conda` command, leaving things broken.  This was happening because the run_constrains entry came after the depends entry, and the run_constrains entry made the setuptools constraint optional, leading to its removal.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -371,7 +371,10 @@ class Solver(object):
         # add in historically-requested specs
         ssc.specs_map.update(ssc.specs_from_history_map)
 
-        for pkg_name in ('anaconda', 'conda', 'conda-build',
+        # these are things that we want to keep even if they're not explicitly specified.  This
+        #     is to compensate for older installers not recording these appropriately for them
+        #     to be preserved.
+        for pkg_name in ('anaconda', 'conda', 'conda-build', 'python.app',
                          'console_shortcut', 'powershell_shortcut'):
             if pkg_name not in ssc.specs_map and ssc.prefix_data.get(pkg_name, None):
                 ssc.specs_map[pkg_name] = MatchSpec(pkg_name)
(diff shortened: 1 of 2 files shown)


conda/core/solve.py
```diff
@@ -571,7 +574,7 @@ def _add_specs(self, ssc):
                 if target_prec.is_unmanageable:
                     ssc.specs_map[pkg_name] = target_prec.to_match_spec()
                 elif MatchSpec(pkg_name) in context.aggressive_update_packages:
-                    ssc.specs_map[pkg_name] = pkg_name
+                    ssc.specs_map[pkg_name] = MatchSpec(pkg_name)
```
msarahan: @forrestwaters this is the string/MatchSpec bug you were seeing earlier

conda/core/solve.py
```diff
@@ -642,8 +645,8 @@ def _add_specs(self, ssc):
         elif ssc.update_modifier == UpdateModifier.UPDATE_SPECS:
             for spec in self.specs_to_add:
                 spec = MatchSpec(spec)
-                if (spec.name in pin_overrides and not ssc.ignore_pinned
-                        or spec.name in ssc.specs_from_history_map):
```
msarahan: I'm removing the specs_from_history_map thing here because the specs are explicitly being updated here.  They should not be held back.

conda/models/records.py
```diff
@@ -322,9 +322,9 @@ def is_unmanageable(self):
     def combined_depends(self):
         from .match_spec import MatchSpec
         result = {ms.name: ms for ms in MatchSpec.merge(self.depends)}
-        result.update({ms.name: ms for ms in MatchSpec.merge(
-            MatchSpec(spec, optional=True) for spec in self.constrains or ()
```
msarahan: This was the main bug that this PR addresses.  That `optional=True` is the problem.

conda/core/solve.py
```diff
@@ -371,7 +371,10 @@ def _collect_all_metadata(self, ssc):
         # add in historically-requested specs
         ssc.specs_map.update(ssc.specs_from_history_map)
 
-        for pkg_name in ('anaconda', 'conda', 'conda-build',
+        # these are things that we want to keep even if they're not explicitly specified.  This
+        #     is to compensate for older installers not recording these appropriately for them
+        #     to be preserved.
+        for pkg_name in ('anaconda', 'conda', 'conda-build', 'python.app',
```
msarahan: `python.app` is here because it comes with miniconda on MacOS, but the installers don't seem to record it in the history.  Thus, it is automatically removed on the first conda operation.

jsolbrig:
Should this be patched to 4.7.10?

PR #9015, opened 2019-07-30
Prevent duplicate calls to find_conflicts and add test

forrestwaters:


conda/core/solve.py
```diff
@@ -188,6 +195,9 @@ def solve_final_state(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL
                 for the prefix.
             force_remove (bool):
                 Forces removal of a package without removing packages that depend on it.
+            should_retry_solve (bool):
+                Indicates whether this solve will be retried. This allows us to control
+                find_conflicts in ssc.r.solve
```
msarahan:
```suggestion
                whether to call find_conflicts (slow) in ssc.r.solve
```

conda/resolve.py
```diff
@@ -1178,10 +1178,11 @@ def solve(self, specs, returnall=False, _remove=False, specs_to_add=None, histor
                 raise ResolvePackageNotFound(not_found_packages)
             elif wrong_version_packages:
                 raise UnsatisfiableError([[d] for d in wrong_version_packages], chains=False)
-            if repodata_fn != REPODATA_FN:
-                # bail out until we have the full repodata.
+            if should_retry_solve:
+                # we don't want to call find_conflicts until our last try
```
msarahan:
```suggestion
                # we don't want to call find_conflicts until our last try.
                #    This jumps back out to conda/cli/install.py, where the 
                #    retries happen
```

msarahan:
Looks great.  We'll see how the tests come out, but I don't have any code changes to recommend.

forrestwaters:
Updated with your suggestions. The py27 failure now is a codecov timeout

msarahan:
Windows has annoying permission errors, such that the usual shutil-based file removal doesn't work very well.  You should instead use conda.gateway.disk.delete.rm_rf

PR #9021, opened 2019-07-30
Small refactor of _conda_activate.bat

Bezier89:
This may or may not make https://github.com/conda/conda/issues/8733 less likely depending on how if/else is implemented under the hood, but seems a bit cleaner either way.


msarahan:
Thanks!

PR #9025, opened 2019-07-30
Add support for local version specs in PYPI installed packages.

matthewwardrop:
Currently the PYPI version regex matcher in conda does not support local version identifiers as defined here: https://www.python.org/dev/peps/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions . The local version spec allows you to add something like: `'1.2.1+<githash>'`. In some internal environments (such as ours) local versions are occasionally used, and so this breaks our sharing of conda environments using `conda env export`.

This patch adds '+' to the `PARTIAL_PYPI_SPEC_PATTERN` along with an additional unit test.

I just signed the CLA, but I'm not sure how long it take to propagate :).

### DIFF
--- a/conda/common/pkg_formats/python.py
+++ b/conda/common/pkg_formats/python.py
@@ -48,7 +48,7 @@ PARTIAL_PYPI_SPEC_PATTERN = re.compile(r'''
     \s?
     (\[(?P<extras>.*)\])?
     \s?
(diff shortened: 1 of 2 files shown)


msarahan:
Thanks!  The CLA is a manual thing.  I confirm that you have signed it.  This looks fine to me.  Merging.

Issue #8993, opened 2019-07-25
S3 channels failing with new multi-threaded package metadata retrieval


hajapy:
## Current Behavior

Attempting to create an environment when an s3 channel is provided and boto3 is installed results in errors on linux. 

### Steps to Reproduce

```
conda create -n test -c s3://<redacted> python requests numpy <redacted>
Collecting package metadata (current_repodata.json): failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/exceptions.py", line 1062, in __call__
        return func(*args, **kwargs)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/cli/conda_argparse.py", line 82, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/cli/main_create.py", line 37, in execute
        install(args, parser, 'create')
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/cli/install.py", line 256, in install
        force_reinstall=context.force_reinstall or context.force,
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/core/solve.py", line 112, in solve_for_transaction
        force_remove, force_reinstall)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/core/solve.py", line 150, in solve_for_diff
        force_remove)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/core/solve.py", line 249, in solve_final_state
        ssc = self._collect_all_metadata(ssc)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/common/io.py", line 88, in decorated
        return f(*args, **kwds)
      File "/home/ubuntu/jenkins/workspace/miniconda/lib/python3.6/site-packages/conda/core/solve.py", line 389, in _collect_all_metadata
        index, r = self._prepare(prepared_specs)
[...]

PR #9030, opened 2019-07-31
simplify some conflict checks for speed

msarahan:


conda/core/solve.py
```diff
@@ -518,7 +518,8 @@ def _package_has_updates(self, ssc, spec, installed_pool):
                     has_update = True
                     break
         # let conda determine the latest version by just adding a name spec
-        return MatchSpec(spec.name) if has_update else None
+        return (MatchSpec(spec.name, version=prec.version, build_number=prec.build_number)
```
msarahan: This super-specific version/build number doesn't matter for solving. We used it as a reference point to say "can you install this?  If not, which specs are in the way?" and we then unfreeze them if we can.

conda/core/solve.py
```diff
@@ -653,22 +639,21 @@ def _add_specs(self, ssc):
...
-                    spec_pool = ssc.r._get_package_pool([ms])
-                    if spec_pool.get(spec.name) and latest_pkg not in spec_pool[spec.name]:
-                        # neuter the spec due to a conflict
-                        ssc.specs_map[ms.name] = MatchSpec(ms.name)
+            skip = lambda x: (x.name in pin_overrides and not ssc.ignore_pinned
+                              or x.name in ssc.specs_from_history_map)
+            specs_to_add = [self._package_has_updates(ssc, _, installed_pool)
+                            for _ in self.specs_to_add if not skip(_)]
+            # the index is sorted, so the first record here gives us what we want.
+            conflicts = ssc.r.get_conflicting_specs(specs_to_add +
```
msarahan: condensing this to one call to get_conflicting_specs is a speedup relative to the loop with comparing pools.  That one call is slower than any given pool comparison, but faster overall.

conda/core/subdir_data.py
```diff
@@ -87,9 +87,14 @@ def query_all(package_ref_or_match_spec, channels=None, subdirs=None,
             package_ref_or_match_spec))
 
         # TODO test timing with ProcessPoolExecutor
-        with ThreadLimitedThreadPoolExecutor() as executor:
-            result = tuple(concat(executor.map(subdir_query, channel_urls)))
-        return result
+        if any('s3://' in url for url in channel_urls):
```
msarahan: fixes #8993 maybe

soapy1:
looks good once tests pass!

conda/core/subdir_data.py
hajapy: You could probably allow running s3 checks in parallel if you added `_ = boto3.client('s3')` to `gateways/connection/adapters/s3.py` to allow the first time initialization to happen on main thread. Subsequent calls should be thread-safe. This would require a bit of testing?

msarahan: Please submit a PR if you can.  I'm not very familiar with s3 at all, and don't have confidence in my ability to write meaningful tests.

tests/test_logic.py
```diff
@@ -289,6 +289,7 @@ def test_minimize():
     assert sval == 11
 
 
+@pytest.mark.xfail(reason="Broke this with reworking minimal_unsatisfiable_set.  Not sure how to fix.  minimal_unsatisfiable_subset function is otherwise working well.")
```
soapy1: Can this test be removed if it's never going to pass? Or, should another similar test for the new implementation of this function be written?

tests/core/test_solve.py
```diff
@@ -925,11 +925,12 @@ def test_unfreeze_when_required():
...
-    specs_to_add = MatchSpec("qux"),
-    with get_solver_must_unfreeze(specs_to_add, prefix_records=final_state_1, history_specs=specs) as solver:
-        with pytest.raises(UnsatisfiableError):
-            solver.solve_final_state(update_modifier=UpdateModifier.FREEZE_INSTALLED)
+    # When frozen there is no solution - but conda tries really hard to not freeze things that conflict
+    #    this section of the test broke when we improved the detection of conflicting specs.
+    # specs_to_add = MatchSpec("qux"),
+    # with get_solver_must_unfreeze(specs_to_add, prefix_records=final_state_1, history_specs=specs) as solver:
+    #     with pytest.raises(UnsatisfiableError):
+    #         solver.solve_final_state(update_modifier=UpdateModifier.FREEZE_INSTALLED)
```
soapy1: nice! 💯

tests/test_logic.py
msarahan: Yes, we should do the latter.  Ideally we could adjust this test to behave again.  I'm inclined to postpone this until later, just because we need to get bugfixes out, and this helps.

conda/core/subdir_data.py
hajapy: I gave it a go here: https://github.com/conda/conda/pull/9037

msarahan: Thanks!  I'm going to just revert this blob, as I think your PR is a better approach.

conda/common/logic.py
```diff
@@ -330,9 +330,21 @@ def _assign_no_name(self, vals):
 
     def Convert_(self, x):
         tx = type(x)
+        name = x
         if tx in (tuple, list):
             return tx(map(self.Convert_, x))
-        return self.names.get(x, x)
+        elif tx not in (bool, int) and name not in self.names:
```
mbargull: Hi @msarahan,
I'm working on the preparation part of https://github.com/curiu/notes/issues/2, i.e., cleaning up this module.
I don't understand this change. Can you elaborate on the need for this?
Ideally, `Convert_` would
- leave `True`, `False` and valid `int`s (i.e., `0 < abs(x) <= self.m`) untouched,
- convert registered names into the corresponding valid values,
- error out in case of invalid inputs.

Before this change, it would silently accept wrong inputs and return them as-is which would likely lead to raising some unexpected errors shortly after.
After this change, it silently accepts wrong inputs and adds corresponding new variables for non-mapped names.

Both behaviors, before and after, are wrong, so this change didn't make things much worse ;). It might make things fail more silently, though.
Do you have any test case that lead to this change?

msarahan: I am certain that you have a better handle on this than I did/do. I almost certainly considered silent falling as the improvement I needed without realizing that was what was happening.

For a test case, I recommend just reverting this bit and seeing what failures show up. I unfortunately don't remember any more.

mbargull: > I unfortunately don't remember any more.

Ah, bummer -- but totally understandable ;). Thanks for the prompt answer!

PR #9031, opened 2019-07-31
Added conceptual info on .conda file format

rrigdon:


docs/source/user-guide/concepts/packages.rst
```diff
@@ -29,6 +29,29 @@ package. Directories are not included. Directories are created
...
 
+.conda file format
+==================
+
+
+The .conda file format was introduced in conda 4.7 as a more
+compact, and thus faster, alternative to a tarball. 
+
+The .conda file format consists of an outer, uncompressed
+ZIP-format container, with two inner compressed .tar files.
```
mingwandroid: Could you mention that the inner ones are compressed with zstd or zstandard explicitly here @rrigdon?

msarahan: We should explain that the format inside doesn't matter, as long as libarchive know what to do with it. Zstd is the current implemention, but it is not part of the spec.

docs/source/user-guide/concepts/packages.rst
```diff
@@ -29,6 +29,35 @@ package. Directories are not included. Directories are created
...
+The inner files are compressed using Zstandard (ztsd).
```
msarahan: Paragraph suggestions:

For the .conda format's initial internal compression format support, 
we chose Zstandard (zstd).  The actual compression format used does not
matter, as long as the format is supported by libarchive.  The compression
format may change in the future as more advanced compression algorithms are
developed, and no change to the .conda format is necessary.  Only an updated 
libarchive would be required to add a new compression format to .conda files.

docs/source/user-guide/concepts/packages.rst
```diff
@@ -29,6 +29,35 @@ package. Directories are not included. Directories are created
...
+The inner files are compressed using Zstandard (ztsd). 
+Ztsd is the current implementation but it is not
+part of the spec. This means that the inside format
+does not matter as long as libarchive knows what to
+do with it.
+
+These compressed files can be significantly smaller than their
+bzip2 equivalents. In addition, they decompress much more quickly.
+.conda is the preferred file format to use where available,
+although we continue to support .tar.bzip2 files.
```
msarahan:
```suggestion
although we continue to provide .tar.bz2 files in tandem.
```

docs/source/user-guide/concepts/packages.rst
```diff
@@ -29,6 +29,35 @@ package. Directories are not included. Directories are created
...
+
+Read more about the `introduction of the .conda file format <https://www.anaconda.com/understanding-and-improving-condas-performance/>`_.
+
+.. note::
+
+  In conda 4.7 and later, you cannot use filenames that
```
msarahan:
```suggestion
  In conda 4.7 and later, you cannot use package names that
```

msarahan:
please consider my suggestions and merge after incorporating or rejecting them.

Issue #9034, opened 2019-08-01
Conda doesn't use packages in cache


Bezier89:
Have two similar but not identical explicit spec files. Here's an example with flask=1.0.2 and flask=1.0.3, generated with the following commands:

```
conda create -n flask102 flask=1.0.2
conda list -n flask102 --explicit > flask102.txt
conda create -n flask103 flask=1.0.3
conda list -n flask103 --explicit > flask103.txt
```

The resulting files are [flask102.txt](https://github.com/conda/conda/files/3458397/flask102.txt) and [flask103.txt](https://github.com/conda/conda/files/3458395/flask103.txt). Note that they only differ by a single package. To reproduce, download these two files and run the following commands in a conda installation with a clean cache:

```
conda create -n TestEnv1 --file flask102.txt
conda create -n TestEnv2 --file flask102.txt
conda create -n TestEnv3 --file flask103.txt
```

Notice the difference in behavior when creating TestEnv2 versus TestEnv3. Creating TestEnv1 will, as expected, download all packages into the package cache. Creating TestEnv2 will, as expected, use the existing packages in the package cache and not download any additional packages. TestEnv3 will, rather than downloading a single package (flask=1.0.3) and using the rest from the cache, download every single package over again.

The behavior appears to be that, if even a single package is missing from the cache, ALL packages in the environment will be downloaded rather than just those missing from the cache. You can confirm this by now manually deleting flask 1.0.3 from the package cache and then running this command:

`conda create -n TestEnv4 --file flask103.txt`

and noting that it will again redownload everything, rather than just downloading flask 1.0.3.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : C:\Users\username\AppData\Local\Continuum\miniconda3-debug
            shell level : 1
       user config file : C:\Users\username\.condarc
 populated config files :
          conda version : 4.7.10
    conda-build version : not installed
         python version : 3.7.3.final.0
       virtual packages :
       base environment : C:\Users\username\AppData\Local\Continuum\miniconda3-debug  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
[...]

PR #9039, opened 2019-08-02
Fix return condition in package_cache_data

Bezier89:
This is related to https://github.com/conda/conda/issues/9034. It is the reason why there is different behavior for no packages missing from the cache (where cache_actions is false and the code returns early) versus one or more packages missing from the cache.


msarahan:
Thanks!

PR #9040, opened 2019-08-02
Fix https://github.com/conda/conda/issues/9034

Bezier89:
This fixes https://github.com/conda/conda/issues/9034. The issue is that for @EXPLICIT packages, the md5 is initially None, so conda would end up not using existing packages in the cache.


Bezier89:
The breaking change may have been in https://github.com/conda/conda/commit/8690b55918d1492e9a0c4acda81fe9269620b315

msarahan:
Thanks!

PR #9044, opened 2019-08-03
serialize menu creation

msarahan:
menu creation and removal was happening in parallel with unlinking/linking.  This meant that
menuinst might not have the python executable it needed to do its thing.

This also refactors the loops stuff here to do fewer runs through the same loop.


msarahan:
Sorting out remaining PATH issues on the python-libarchive-c package.  Merging.

PR #9054, opened 2019-08-05
fix incorrect RemoveError when operating on an env that has one of co…

msarahan:
…nda's deps, but is not the conda env


jjhelmus:
Might conflict with #9053 but will be easy to rebase

PR #9072, opened 2019-08-08
using Miniconda3-latest-Linux-x86_64.sh in example

martin-raden:


jjhelmus:
Thanks for the update @martin-raden

Issue #6788, opened 2018-01-25
Provide a conda command to install dependencies from package metadata?


ceball:
Imagine I'm developing 3 pure python packages. In the python world, I might run `pip install -e .` on each of the 3 git checkouts, which would install the 3 packages in 'develop'/'editable' mode, and would also install the dependencies of all 3 into my current environment.

Often, I would prefer to use conda instead of pip - particularly for installing the dependencies of the python packages I'm developing. The python packages I'm developing always have a meta.yaml file in the repository (for building a conda package to release), so I hoped I would be able to use that somehow via `conda develop`. However, `conda develop` does not seem to do what I want, and is currently not recommended (e.g. https://github.com/conda/conda-build/issues/2633#issuecomment-359572215, https://github.com/conda/conda-build/issues/1990#issuecomment-305549938, https://github.com/conda/conda-build/issues/1992).

A suggestion that seems to have come up several times for `conda develop` is that it should create an environment from the meta.yaml file (e.g. https://github.com/conda/conda-build/issues/1677), in addition to installing the package in 'develop' mode. However, I would prefer a new environment not to be created (or for it to be optional), because I want to be able to install multiple packages' sets of dependencies into the same, existing environment. I would also have no problem running `pip install --no-deps -e` myself, after having installed the dependencies using conda, because I am only ever working on pure python packages<sup>1</sup>.

I think what I want has been asked before (#5126), but the resolution there is not what I'm looking for. Unless I'm misunderstanding, I think that issue leads to a command to do something like `conda install --dependencies-only pkg`, where pkg is an existing (built) conda package. I don't want to have to build the conda package and then install dependencies from it because that's a time consuming process - particularly if I am currently editing the dependencies - and the package could anyway fail to be created (e.g. my package tests might fail). And I don't want to use an existing version of pkg from elsewhere because the dependencies may be out of date.

Currently, projects I see tend to have developer instructions like one of the following:

  * `conda install pkg && conda remove --force pkg && pip install --no-deps -e .` (which gets a developer the released package's dependencies, not those specified in the repo's current meta.yaml file), or

  * `conda install dep1 dep2 ... depN && pip install --no-deps -e .` where dep1, dep2, ..., depN are somehow never quite the same as what's in meta.yaml (accidentally, not just because dependencies have moved on since last time), and are different from what developers actually install, and vary again on travis and appveyor too (accidentally, not just because there are genuine platform differences).

[...]

PR #9086, opened 2019-08-12
handle meta.yaml file in environment commands

nicolas-geniteau:
This PR allow `conda env {create/update}` command to handle meta.yaml file. It use `conda_build` to correctly parse Metadata.

Il also add a new argument `--test` to also install test requirements from meta.yaml file.

Feature request: #6788

### DIFF
--- a/conda_env/cli/main_create.py
+++ b/conda_env/cli/main_create.py
@@ -64,6 +64,12 @@ def configure_parser(sub_parsers):
         action='store_true',
         default=False,
     )
+    p.add_argument(
+        '--test',
+        action='store_true',
+        default=False,
+        help='install test requirements from meta.yaml file',
+    )
     add_parser_json(p)
     p.set_defaults(func='.main_create.execute')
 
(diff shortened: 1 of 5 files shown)


msarahan:
This is an interesting idea.  I think it is the same kind of idea as the ``conda debug`` command built into conda-build.  https://docs.conda.io/projects/conda-build/en/latest/user-guide/recipes/debugging.html Have you seen that?  How would you say this differs from that?

If this is indeed different enough that it should be added here, you need to change your recipe rendering to use the conda_build.api.render function.  Only instantiating a MetaData object is not enough to fill in templates and such.  Keep in mind that rendering may return more than one MetaData object, too

nicolas-geniteau:
@msarahan Thanks for your response.

I think `conda debug` command has not exactly the same goal.
In my case, I have a conda package with requirements in my meta.yaml file. From this, I would like to create:
 * a real conda package
 * a dev environment (with test requirements also) with "lock file" (`conda list`)
 * a production environment with "lock file" to create docker image

With this PR, I could do something like:
```
conda env create -n prod -f meta.yaml
conda activate prod
conda list > environment.yaml

conda env create -n dev -f meta.yaml --test
conda activate dev
conda list > environment_dev.yaml
```

Is it possible to do something similar with `conda debug` ? Or maybe my process is wrong (I am totally new with conda) ?

If you think this PR can be usefull, I will update it to use `conda_build.api.render`

> Keep in mind that rendering may return more than one MetaData object, too

I don't know well recipes, in which case can we have more than one MetaData ?

PR #9093, opened 2019-08-13
Add ability to create environment spec from explicit specs in history

forrestwaters:


jjhelmus:
LGTM, AppVeyor is timing out, not a blocker.

PR #9123, opened 2019-08-19
fix ProxyError expected arguments

forrestwaters:
Fixes https://github.com/ContinuumIO/anaconda-issues/issues/11202


jjhelmus:
Timeout on AppVeyor, not a blocker.

PR #9139, opened 2019-08-26
Add bash_completion for the conda command

tartansandal:
Implements the feature requested in #635. 

I'm not sure about integration (filesystem location, init) and where this should be documented.  Any advice most appreciated.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -786,6 +786,16 @@ class PosixActivator(_Activator):
                     result += join(self.export_var_tmpl % (key, value)) + '\n'
         return result
 
+class BashActivator(PosixActivator):
+
+    def __init__(self, arguments=None):
+        super(BashActivator, self).__init__(arguments)
+        self.hook_source_path = join(CONDA_PACKAGE_ROOT, 'shell', 'etc', 'conda_bash.sh')
+
+    def _hook_preamble(self):
+        result = super(BashActivator, self)._hook_preamble()
+        result += join(self.export_var_tmpl % ('_CONDA_ROOT', CONDA_PACKAGE_ROOT)) + '\n'
+        return result
 
 class CshActivator(_Activator):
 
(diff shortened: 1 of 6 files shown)


tartansandal:
I'm thinking this code code could be appended to or sourced from `conda/shell/etc/profile.d/conda.sh`, perhaps with an appropriate guard to ensure $BASH_COMPLETION_VERSINFO is defined.

kalefranz:
> I'm thinking this code code could be appended to or sourced from `conda/shell/etc/profile.d/conda.sh`

This should be wired into the output of `conda shell.bash hook`, but not `conda shell.posix hook`.

tartansandal:
> This should be wired into the output of `conda shell.bash hook`, but not `conda shell.posix hook`.

Ah thanks. Will have a dig around to see how to do that :-) 

As a side note, I plan to close this PR once I get all this sorted, and resubmit a clean squashed PR.  Hopefully figure out why the CI builds are failing too.

tartansandal:
> This should be wired into the output of `conda shell.bash hook`, but not `conda shell.posix hook`.

So subclassing `PosixActivator` from `conda/activate.py` is what you are proposing. Setting `self.hook_source_path` to an alternative file might work, though it is complicated by the need to source he POSIX script as well. This could cause problems with system-wide installs. I'm a little worried about the impact of `/etc/profile.d/`. Hmmm...
I'll pull apart the installer an see if I can come up with a sensible plan.

tartansandal:
Just returned from that rabbit hole.  More complicated than I had hoped.  All I need to do now is write some tests :-)

tartansandal:
This PR has been rebased into https://github.com/conda/conda/pull/9158.

kalefranz:
> So subclassing `PosixActivator` from `conda/activate.py` is what you are proposing.

Yes, exactly. It was always part of the original intent that someday someone would come by and do exactly what you’re doing. By that, I mean the abstractions should be there so that it’s pretty straight forward to add in both bash completions and zsh completions. Apparently you’ve found that’s not the case.

If you’ve closed this PR out of frustration, that’s unfortunate. Might be Tuesday, but I intend to look at where you got hung up and figure out a solution. Conda needs completions for bash, and I’d like to see this move forward.

tartansandal:
I've not abandoned the PR, just rebased it into a single commit. Got all
the tests running so I think this may be complete now. Hopefully that PR
can serve as a guide for someone who wants to implement zsh completion.

Issue #9124, opened 2019-08-19
conda keeps history of python specific version even when python version changes


jjhelmus:
## Current Behavior

The python version in an environment can be changed by explicitly providing a version on the command line (e.g. `conda install -n test python=3.6`).  

When loading the historic specification for future transactions this change is not considered and can result in incompatible historic specifications being add to the solver which can result in all transactions raising an `UnsatisfiableError`.  This did not occur in conda 4.6.14.

### Steps to Reproduce

``` console
$ conda --version
conda 4.7.11.post19+3ac7898ee
$ conda create -n test -q -y psutil=5.6.3=py37h7b6447c_0
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... done

## Package Plan ##

  environment location: /home/jhelmus/workspace/conda/devenv/envs/test

  added / updated specs:
    - psutil==5.6.3=py37h7b6447c_0


The following NEW packages will be INSTALLED:

  _libgcc_mutex      pkgs/main/linux-64::_libgcc_mutex-0.1-main
  ca-certificates    pkgs/main/linux-64::ca-certificates-2019.5.15-1
  libedit            pkgs/main/linux-64::libedit-3.1.20181209-hc058e9b_0
  libffi             pkgs/main/linux-64::libffi-3.2.1-hd88cf55_4
  libgcc-ng          pkgs/main/linux-64::libgcc-ng-9.1.0-hdf63c60_0
  libstdcxx-ng       pkgs/main/linux-64::libstdcxx-ng-9.1.0-hdf63c60_0
  ncurses            pkgs/main/linux-64::ncurses-6.1-he6710b0_1
  openssl            pkgs/main/linux-64::openssl-1.1.1c-h7b6447c_1
  psutil             pkgs/main/linux-64::psutil-5.6.3-py37h7b6447c_0
  python             pkgs/main/linux-64::python-3.7.4-h265db76_1
  readline           pkgs/main/linux-64::readline-7.0-h7b6447c_5
  sqlite             pkgs/main/linux-64::sqlite-3.29.0-h7b6447c_0
  tk                 pkgs/main/linux-64::tk-8.6.8-hbc83047_0
  xz                 pkgs/main/linux-64::xz-5.2.4-h14c3975_4
  zlib               pkgs/main/linux-64::zlib-1.2.11-h7b6447c_3


Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
[...]

PR #9147, opened 2019-08-27
fix neutered specs not being recorded in history, leading to sadness

msarahan:
fixes https://github.com/conda/conda/issues/9124

CC @jjhelmus


jjhelmus:
Looks good except for the failing test in Python 2.7.

Issue #4806, opened 2017-03-04
environment dependent PYTHONSTARTUP file


lewisacidic:
I currently have a minimal set of imports that I use for most interactive sessions in a `.pythonrc` file at $PYTHONSTARTUP, i.e.:

```python
import pandas as pd
import numpy as np
...
```

Which I would like to be environment dependent (as not all envs have numpy etc.).  I have been able to do this by putting this `.pythonrc` in the root of the environment, and setting PYTHONSTARTUP to it in `{ENVROOT}/etc/activate.d/env_vars.sh` (and then unsetting it in `deactivate.d`), but in my opinion, this is unnecessarily clunky and won't port between systems, so could do with specific conda support.   A simple way to set this up would be to have something like a `{ENVROOT}/startup.py` that is run after the script at $PYTHONSTARTUP, allowing the convenience code that typically goes in these files (e.g. https://github.com/whiteinge/dotfiles/blob/master/.pythonrc.py) to be environment independent.  Not sure how the best way to do it would be - thoughts?

I'm afraid I'm currently writing up my PhD so don't have time to dive into the conda code and offer a PR.

Issue #6649, opened 2018-01-04
Allow management of activate.d environment variables in environment.yml files


pgunn:
This is a new feature proposal, to extend the format for the yml files usable to create conda environments, to allow expression of environment variables (with values) to be statically set. I propose:

Another toplevel in that file titled environment, with k=v elements under it being parsed as environment variables to set in the environment. Example:

my_tensorflow_environment.yml
channels:
- defaults
dependencies:
- keras
- tensorflow
- python=3.6
environment:
- KERAS_BACKEND=tensorflow
- AWS_KEY=a32131231231
- AWS_SECRET=a123123123123

When parsing this via "conda env create -f my_tensorflow_environment.yml -n tensor", conda would populate $INSTALL/envs/tensor/etc/conda/activate.d/env_vars.sh and deactivate.d/env_vars.sh accordingly.

The purposes of this proposal are:
1) For people making software that's meant to run in a conda environment, to let the environment.yml fully express what's needed for the environment to be set up correctly so we can keep these instructions out of README files, and
2) To bridge a last bit of gap between containers and conda

If people like this proposal I'd be happy to work on it.

pgunn:
https://github.com/conda/conda/pull/6681 is a proof-of-concept (that would need some cleanup before people consider landing it). @kalefranz is this a reasonable approach?

bpben:
I would love to see this implemented.  Looks like it's failing some of the CI checks? Is there some way I can help with this?

pgunn:
I was hoping to hear back from kalefranz before I spent any more time on it, to get a feel for whether they'd accept it if it passed all the tests and stuff, or if they'd prefer a different approach. If you want to dig into the CI stuff that might be mildly helpful, but I don't know if this proposal is a dead end or not.

StanczakDominik:
I would love to get this implemented - I once asked a question on Stack about the very problem this issue discusses and it I keep getting reminders that this is a thing.

I think the CI stuff may not have been due to the code itself, as it looks like a `hg clone` error:
```
.
----------------------------- Captured stdout call -----------------------------
Warning: failed to download source.  If building, will try again after downloading recipe dependencies.
Error was: 
Command '['hg', 'clone', 'https://bitbucket.org/asmeurer/conda-build-hg-test', '/tmp/cb/popen-gw1/test_recipe_builds_source_hg_j0/hg_cache/__bitbucket.org_asmeurer_conda-build-hg-test']' returned non-zero exit status 255.
```
[...]

PR #9157, opened 2019-08-30
Env vars for environments

soapy1:
resolves: https://github.com/conda/conda/issues/4806 and https://github.com/conda/conda/issues/6649

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -314,11 +314,10 @@ class _Activator(object):
         activate_scripts = self._get_activate_scripts(prefix)
         conda_default_env = self._default_env(prefix)
         conda_prompt_modifier = self._prompt_modifier(prefix, conda_default_env)
-        conda_environment_env_vars_all = self._get_environment_env_vars(prefix)
-        conda_environment_env_vars = OrderedDict({
-            k: v for k, v in conda_environment_env_vars_all.items()
-            if v != CONDA_ENV_VARS_UNSET_VAR
-        })
+        conda_environment_env_vars = self._get_environment_env_vars(prefix)
+        unset_env_vars = [k for k, v in conda_environment_env_vars.items()
+                          if v == CONDA_ENV_VARS_UNSET_VAR]
+        [conda_environment_env_vars.pop(_) for _ in unset_env_vars]
 
         unset_vars = []
         if old_conda_shlvl == 0:
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -677,7 +677,7 @@ class _Activator(object):
 
         # First get env vars from packages
         if exists(pkg_env_var_dir):
-            for pkg_env_var_file in os.listdir(pkg_env_var_dir):
+            for pkg_env_var_file in sorted(os.listdir(pkg_env_var_dir)):
                 with open(join(pkg_env_var_dir, pkg_env_var_file), 'r') as f:
                     env_vars.update(json.loads(f.read(), object_pairs_hook=OrderedDict))
 
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -10,7 +10,7 @@ from glob import glob
 from itertools import chain
 from logging import getLogger
 import os
-from os.path import abspath, basename, dirname, expanduser, expandvars, isdir, join
+from os.path import abspath, basename, dirname, expanduser, expandvars, isdir, join, exists
 import re
 import sys
 from textwrap import dedent
@@ -312,6 +312,7 @@ class _Activator(object):
         activate_scripts = self._get_activate_scripts(prefix)
         conda_default_env = self._default_env(prefix)
         conda_prompt_modifier = self._prompt_modifier(prefix, conda_default_env)
+        conda_environment_env_vars = self._get_environment_env_vars(prefix)
 
         unset_vars = []
         if old_conda_shlvl == 0:
@@ -650,6 +651,18 @@ class _Activator(object):
         return self.path_conversion(sorted(glob(join(
             prefix, 'etc', 'conda', 'deactivate.d', '*' + self.script_extension
         )), reverse=True))
+
+    def _get_environment_env_vars(self, prefix):
+        env_vars = {}
+        env_vars_file = join(prefix, 'etc', 'conda', 'env_vars')
+        if exists(env_vars_file):
+            with open(env_vars_file, 'r') as f:
+                raw_env_vars = f.read()
+            for env_assignment in raw_env_vars.split("\n"):
+                if "=" in env_assignment:
+                    parts = env_assignment.split("=")
+                    env_vars[parts[0].strip()] = parts[1].strip()
+        return env_vars
 
 
 def expand(path):
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -335,24 +335,30 @@ class _Activator(object):
             if stack:
                 new_path = self.pathsep_join(self._add_prefix_to_path(prefix))
                 deactivate_scripts = ()
-                export_vars, unset_vars = self.get_export_unset_vars(odargs=OrderedDict((
+                env_vars_to_export = OrderedDict((
                     ('path', new_path),
                     ('conda_prefix', prefix),
                     ('conda_shlvl', new_conda_shlvl),
                     ('conda_default_env', conda_default_env),
-                    ('conda_prompt_modifier', conda_prompt_modifier))))
+                    ('conda_prompt_modifier', conda_prompt_modifier)))
+                for k, v in conda_environment_env_vars.items():
+                    env_vars_to_export[k] = v
+                export_vars, unset_vars = self.get_export_unset_vars(odargs=env_vars_to_export)
                 export_vars['CONDA_PREFIX_%d' % old_conda_shlvl] = old_conda_prefix
                 export_vars['CONDA_STACKED_%d' % new_conda_shlvl] = 'true'
             else:
                 new_path = self.pathsep_join(
                     self._replace_prefix_in_path(old_conda_prefix, prefix))
                 deactivate_scripts = self._get_deactivate_scripts(old_conda_prefix)
-                export_vars, unset_vars = self.get_export_unset_vars(odargs=OrderedDict((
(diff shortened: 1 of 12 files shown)


tests/test_activate.py
```diff
@@ -71,6 +71,10 @@
     'prompt',
 )
 
+ENV_VARS_FILE = '''ENV_ONE=one
```
soapy1: I think the changes to these tests make sense but I would really appreciate a second set of eyes on them in particular

conda/activate.py
```diff
@@ -651,6 +668,18 @@ def _get_deactivate_scripts(self, prefix):
             prefix, 'etc', 'conda', 'deactivate.d', '*' + self.script_extension
         )), reverse=True))
 
+    def _get_environment_env_vars(self, prefix):
+        env_vars = {}
+        env_vars_file = join(prefix, 'etc', 'conda', 'env_vars')
```
kalefranz: I think this needs to go in `conda-meta/`. That’s the location that’s completely off limits to packages. Otherwise any rando package on anaconda.org could overwrite that location.

soapy1: good call!

msarahan:
``PREFIX_SATE_FILE`` => ``PREFIX_STATE_FILE``?

Otherwise, this looks pretty great.  Perhaps instead of one state file, there could be a folder that collected up JSON files for packages with env vars?  One file per package?  Modifying the state of one file tends to be more fragile than add/remove one file per package.  These could go into conda-meta, or some other new standard location in the env (more user-accessible), perhaps something like etc/conda/env_vars

mcg1969:
I agree with you @msarahan. Each package should be able to contribute to the environment and ensuring consistency in a single file would be a challenge. I definitely believe they need to be in `conda_meta` somewhere just to ensure protection from being clobbered

mcg1969:
Oops didn't mean to close, sorry!

msarahan:
It will be slower, but I think that's a price worth paying for sanity.

soapy1:
That sounds good to me. Will drop a file called `env_vars.json` in `$PREFIX/conda-meta/state`. I think it might be nice to have channel configuration per envs to exist in that folder as well

msarahan:
hmmm... I'm not sure I was clear.  There can be an env_vars folder in that state folder, but you should expect more than one .json file for setting env vars.  Like each package would have one, and they would be placed into that folder.  They would be aggregated at activation time (or perhaps aggregated and cached for faster activation, with cache refresh happening with each install/update/removal operation.)

msarahan:
Also, channel config already lives in .condarc in the root of the env.  Is there a good reason to move it?  Maybe for consolidation?  Duplication and potential inconsistency is something to be very careful with.

mcg1969:
We can't move `.condarc` in my view—it's too important a standard part of conda environment administration.

soapy1:
Right, so I think there are 2 use cases here:
1. package level environment variables that were previously set with activate.d scripts
2. environment variables for a conda environment

To satisfy (1) I think a new directory can be created `<$PREFIX>/etc/conda/env_vars` that has `<package>.json` where each package that has env vars has it's own file

I think this should be separate from environment level configuration of env vars so that users installing packages (hopefully) never messes with environment configuration. This config should exist in `conda-meta`. So to satisfy (2) create a file `<$PREFIX>/conda-meta/state.json`. This also leaves room to add something like https://github.com/conda/conda/issues/5082

msarahan:
I am not proposing moving condarc. I was saying that we should not move channel config out of condarc, nor should we duplicate env-wide channel config outside of condarc.

The state.json file duplicates the info. I see condarc here as the initial value when the env is created. It can be modified afterwards. If it is modified, the env no longer matches the lock file that created it. The lock file can be updated or regenerated if need be.

If you put the channel info in both state.json and condarc, you have to do merging of the lists. That's fine, it's just something to define the behavior of and document it well. I prefer having it in one place with no merge. It's just simpler to me.

Env vars are new. They could go into state.json. I think condarc is nice because users already know to set things there. It would just be a new key. On the other hand, keys that have dict values are rare (conda-build is the only one I can think of), and I'm not sure we have good ways to work with them on the cli right now.

Again, at environment instantiation time, condarc is created from info in the lock file. Beyond that point, it's up for user modification. We should have a way to indicate that an env no longer matches its original lock file, to help users understand when they may need to update or regenerate their lock file. I prefer all of this state to be in condarc, but I'm not going to block on it.

msarahan:
This looks good to me.  I think we should branch off a 4.7.x maintenance branch and merge this to master as the first step towards 4.8.0.

msarahan:
(tests aside) - I think the structure is fine.  I'm OK with the separate state.json file instead of putting env vars into `.condarc`, but we should be very explicit about why we want to do that.  The main reason I can think of is that we want the API for setting env vars to be your new CLI thing.  Is that accurate?  Is the CLI something that people will like equally to editing a YAML file?  I think it's definitely fine for setting one env var at a time, but I wonder if it will annoy people who need to set several env vars at once.

soapy1:
>  I'm OK with the separate state.json file instead of putting env vars into .condarc, but we should be very explicit about why we want to do that

I think going the .condarc will increase the surface area that env vars are touching as well as bring up issues with understanding how merging should work for different levels of condarc's. Though I think that it provides a more seamless experience for users (especially after the initialization of the environment) I think the easier and safer solution is to have another env var file per environment. In order to resolve the ux of introducing a new type of configuration I think the cli should be similar to the current conda config cli. However, in order to enforce that this is an env level specification I think if belongs in the conda env subparser. So, a user should be able to add or remove env vars for their environments using a command of the form conda env vars --set MY_VAR=something -n my_env or conda env vars --unset MY_VAR=something -n my_env respectively, defaulting to the current environment one is not specified.

> Is the CLI something that people will like equally to editing a YAML file

I think it's worth to do a bit of research into this

>  I wonder if it will annoy people who need to set several env vars at once.

With this pr users are able to set env vars in a comma-seperated list `conda env vars --set ONE=something,TWO=somethingelse` though this does not replace the ease of editing several env vars at once in a yaml file

msarahan:
I think that's good enough.  I'm not questioning your choices here, just making sure that you have good reasons for them, so that we can answer people later when they don't like the way we've done it for whatever reason.

msarahan:
Thinking about this more, I think the file that people edit to set these things in is the raw (unlocked) spec file. With that in mind, I definitely think your design here is good. CLI for tweaking, spec file for designing.

conda/base/constants.py
```diff
@@ -228,6 +228,9 @@ def __str__(self):
 PACKAGE_CACHE_MAGIC_FILE = 'urls.txt'
 PREFIX_MAGIC_FILE = join('conda-meta', 'history')
 
+PREFIX_SATE_FILE = join('conda-meta', 'state.json')
+PACKAGE_ENV_VARS_DIR = join('etc', 'conda', 'env_vars')
```
kalefranz: Maybe changes this to `env_vars.d` to make the .d directory pattern explicit.  It'll also then be similar to `activate.d` and `deactivate.d` in the same directory.

conda/core/envs_manager.py
```diff
@@ -128,3 +131,46 @@ def _rewrite_environments_txt(environments_txt_file, prefixes):
...
+
+
+def unset_environment_env_vars(prefix, env_vars):
+    env_state_file = _get_environment_state_file(prefix)
+    current_env_vars = env_state_file.get('env_vars')
+    if current_env_vars:
+        for env_var in env_vars:
+            current_env_vars.pop(env_var)
+        _write_environment_state_file(prefix, env_state_file)
+    return env_state_file['env_vars']
```
kalefranz: Consider moving the changes you've made here to `prefix_data.py`. The `PrefixData` class is what holds all the state for a particular prefix.  The code in `envs_manager.py` is really a level above a single prefix, meant to inspect and interact with all of the prefixes on a given system.

conda_env/cli/main_vars.py
```diff
@@ -0,0 +1,83 @@
...
+from conda.cli import common
+from conda.cli.conda_argparse import add_parser_prefix, add_parser_json
+from conda.core.envs_manager import get_environment_env_vars, set_environment_env_vars, \
+    unset_environment_env_vars
+from conda.base.context import context
+from .common import get_prefix
+
+
+description = """
+Interact with environment varaibles associated with Conda environments
```
kalefranz: typo in varaibles

conda_env/cli/main_vars.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+example = """
+examples:
+    conda env vars --list -n my_env
+    conda env vars --set MY_VAR=something
+    conda env vars --unset MY_VAR
+"""
```
kalefranz: This is looking pretty close to the configuration specific to an environment. Since we're talking about moving channel configuration and a handful of other config parameters to stick to a specific environment, in designing this CLI we should also take into account what the general case will look like.

Seems like `conda env config` is probably the start of the more general case for environment configuration.

Another thing to consider here is that `--flags` are often considered an anti-pattern for what you really intend to be commands.  That is, flags are meant to be optional modifications; if you write out the full list of command examples and every one requires at least one flag, then you've tripped on the anti-pattern.  The current conda CLI does this all over the place, but we need not repeat here as we're introducing new features.

soapy1: sounds good, happy to put this under `conda env config vars`. 
Regarding the flags, conda config currently uses flags for setting/appending/etc. Though this is an antipattern I think another thing that is important to consider is the patterns that users are already familiar with. If it is important that `conda env config ...` matches `conda config` then I think the flags should stay otherwise I'm more than happy to make them sub commands. What do you think?

KeAWang:
@soapy1 is there an example of how to use this? I'd like to specify env vars in a conda environment.yml file

soapy1:
@KeAWang docs are here: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables
I can't speak to if you can use these in an environment.yaml.

PR #9166, opened 2019-09-02
Standardize "conda env list" behavior between platforms

### LINKED ISSUE #9165: conda env list behavior is inconsistent between Linux/macOS and Windows
## Current Behavior

On UNIX-family operating systems (Linux and macOS), `conda env list` generates a list of all environments belonging to the current user. This behavior changes if `conda` is executed with UID 0, in which case `conda` attempts to find all user home directories on the system and list all environments belonging to all users.

This is not consistent with the behavior on Windows. On Windows, `conda env list` always attempts to find all user home directories on the system, and displays all environments belonging to all users. This occurs whether or not the current user is an administrator.

This behavior is not documented in the Conda user guide, which just says that `conda env list` is used "to see a list of all your environments." https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#viewing-a-list-of-your-environments

### Steps to Reproduce

On macOS/Linux, this will list the current user's environments only:

```
conda env list
```

Running this command with `sudo` will list all environments on the system:

```
sudo conda env list
```

On Windows, running this command from either a standard prompt or from an administrative prompts with elevated permissions will list all environments on the system (as long as the executing user can read the environment configuration file for the user that actually owns the environment):

```
conda env list
```

## Expected Behavior

I would expect one of the two following behaviors:

1. `conda env list` only lists the current user's environment on any operating system.
2. `conda env list` only lists the current user's environment, unless it is executed as an administrator. This behavior should be consistent between Windows and macOS. 

The current behavior on Windows (all users) and macOS/Linux (admins) should only be executed for administrators, if at all. In particular, the current Windows code finds all sibling directories of the current user's home folder and tries to check each one of them for an environment configuration file. In environments where Windows home directories are redirected onto a common network share, this may result in `conda env list` firing off a large number of I/O operations as it tries to access all home directories on the network share.

zheaton:
"conda env list" previously behaved differently on Windows and Unix -
on Unix, administrators got a list of all environments on the system,
and regular users saw only their environments. On Windows, however,
all users saw all environments.

This change standardizes both platforms to use the Unix behavior.
Administrators see all environments, and standard users only see
their own environments.

Closes #9165

### DIFF
--- a/conda/core/envs_manager.py
+++ b/conda/core/envs_manager.py
@@ -11,6 +11,7 @@ from os.path import dirname, isdir, isfile, join, normpath
 from .prefix_data import PrefixData
 from ..base.context import context
 from ..common.compat import ensure_text_type, on_win, open
+from ..common._os import is_admin
 from ..common.path import expand
 from ..gateways.disk.read import yield_lines
 from ..gateways.disk.test import is_conda_environment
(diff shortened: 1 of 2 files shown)


zheaton:
I may have botched the new issue/pull request workflow - this issue can be merged with #9165

msarahan:
Thanks, I think this is a nice improvement in consistency.  All of our contributors need to sign our CLA.  Are you able to submit the PDF form at https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement ?

The status here won't change - it's a manual process.  I will post back when I see that you have submitted the CLA.

zheaton:
@msarahan, I signed the CLA on September 2nd. Let me know if you need me to re-sign the agreement for any reason.

msarahan:
Confirmed - sorry about that.  We are behind on the manual step.

Merging.

zheaton:
Thanks!

PR #9190, opened 2019-09-04
Switch for conflict finding

soapy1:
Allows users to set if they want an unsat hint or not

### DIFF
--- a/tests/test_resolve.py
+++ b/tests/test_resolve.py
@@ -354,6 +354,22 @@ def test_unsat_simple():
     assert "a -> c[version='>=1,<2']" in str(excinfo.value)
     assert "b -> c[version='>=2,<3']" in str(excinfo.value)
 
+def test_unsat_simple_dont_find_conflicts():
+    # a and b depend on conflicting versions of c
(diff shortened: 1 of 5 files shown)


tests/test_resolve.py
```diff
@@ -354,6 +354,22 @@ def test_unsat_simple():
...
+def test_unsat_simple_dont_find_conflicts():
+    # a and b depend on conflicting versions of c
+    index = (
+        simple_rec(name='a', depends=['c >=1,<2']),
+        simple_rec(name='b', depends=['c >=2,<3']),
+        simple_rec(name='c', version='1.0'),
+        simple_rec(name='c', version='2.0'),
+    )
+    r = Resolve(OrderedDict((prec, prec) for prec in index))
+    context.unsatisfiable_hints = False
```
msarahan: I don't think changing the context here is doing what you hope it will. You generally have to change env vars and reset the context.  Search the tests for the `env_var()` context manager.

msarahan:
@rrigdon we'll need some docs on this - probably when you get back.

PR #9204, opened 2019-09-05
Add json output to conda env create/update

beenje:
This PR adds json output to the `conda env create/update` commands.

Some things I'm not sure about:
1. I had to `import conda.exports` in `conda_env/cli/main.py`. Otherwise when it is imported in `conda_env/pip_util.py` it resets the context.

2. In my tests, if I keep the same environment name "env-1", each test passes when I run them individually. But they all fail when run together (running `pytest -v tests/conda_env/test_cli.py::IntegrationTests`). I had to use a different env name for each test. The `setUp` and `tearDown` fixtures seem to run properly. At one point I saw that when trying to run pip in the environment `env-1/bin/python` didn't exist. There was only the `env-1/conda-meta` directory. Could some files be deleted by the fixture after the test starts?
Anyway I'm quite sure this is not related to my changes, but an existing issue (in the tests).

### DIFF
--- a/conda_env/cli/common.py
+++ b/conda_env/cli/common.py
@@ -7,6 +7,8 @@ import sys
 
 from conda._vendor.auxlib.entity import EntityEncoder
 from conda.base.context import context
+from conda.cli import install as cli_install
+from conda.cli import common as cli_common
 
 base_env_name = 'base'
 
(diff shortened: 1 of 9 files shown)


msarahan:
Thanks @beenje, merging.

Issue #516, opened 2014-01-21
the use of hard or soft links should be user-definable


asmeurer:
See https://github.com/ContinuumIO/anaconda-issues/issues/65. Some file systems only allow hard links within the same directory. It should be user-configurable.

kalefranz:
Done.

PR #9211, opened 2019-09-05
Update dependencies

rrigdon:
From this conda-docs issue: https://github.com/conda/conda-docs/pull/571

I made the first update to the language around clarifying update_dependencies but additional questions are in the initial PR. I've included it as a comment.


rrigdon:
From initial PR: 

> PR based on discussion in [#516 ](https://github.com/conda/conda-docs/pull/516).
> 
> For the other options in this file, we describe what the various possible settings do, say what the default is, and then provide an example of using a setting other than the default. This updates update_dependencies to match.
> 
> Also @kalefranz , should we restore this note about the command line flags?
> 
> You can enable and disable this option at the command line with the --update-dependencies and --no-update-dependencies flags.

PR #9216, opened 2019-09-06
Mutex metapkgs

rrigdon:
Added info about mutex metapackages and more context about metapackages in general.


rrigdon:
Not sure why that pkg_dirs commit is in here since I put it into this PR: https://github.com/conda/conda/pull/9214

The mutex metapackage section has been reviewed by @msarahan so it just needs a final overview.

PR #9218, opened 2019-09-06
Get conflicting specs speed up

soapy1:


conda/common/logic.py
```diff
@@ -29,6 +29,7 @@
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from functools import lru_cache
```
msarahan: Not used

```suggestion
```

conda/common/logic.py
```diff
@@ -910,19 +911,20 @@ def minimal_unsatisfiable_subset(clauses, sat, explicit_specs):
     <= sat(A)), where A <= B means A is a subset of B and False < True).
 
     """
-    working_set = set(explicit_specs)
+    working_set = set()
     found_conflicts = set()
 
+    if sat(explicit_specs, True) is None:
```
msarahan: This is a smart way to trim the number of SAT solves when there's no conflict.  Good thinking.

soapy1:
conda build tests seem to be adequately sped up for this PR 

https://circleci.com/gh/conda/conda/21154?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```
= 8 failed, 430 passed, 10 skipped, 1 xfailed, 430 warnings in 1016.77 seconds =
```

@Bezier89 this is not quite as fast as your improvements, however it is quite a bit faster than before. Do you think it satisfies the need for conda to be faster?

Bezier89:
@soapy1 Nice! At least with our test suite, speed is roughly the same as my original PR.

PR #9238, opened 2019-09-13
Fixed table

rrigdon:


jjhelmus:
I was not aware of the `list-table` directive in sphinx, happy to have learned about it.   Very nice @rrigdon

Issue #9230, opened 2019-09-11
Windows: All Scripts\*.exe must unset PYTHONHOME


Roffild:
```bat
set PYTHONHOME=123
Scripts\conda.exe

Fatal Python error: initfsencoding: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00003618 (most recent call first):
```
PYTHONHOME should be ignored or deleted when the CLI starts. This will solve the old problem with PYTHONHOME.

The changes that I proposed in the https://github.com/conda/conda/issues/9229 should be taken into account.

mingwandroid:
We absolutely must never mess with `PYTHONHOME` because if we do, we break people who want to use it in our ecosystem for legitimate purposes. If you have it set incorrectly then that is up to you to fix.

Roffild:
I am using PYTHONHOME. In the example, I set PYTHONHOME to a random value. But in reality PYTHONHOME value I have the right.

PYTHONHOME = c:\Anaconda\envs\pycharm\
But Conda package in c:\Anaconda

I have to do ```bat set PYTHONHOME= ``` before upgrading Conda.

> We absolutely must never mess with PYTHONHOME because if we do, we break people who want to use it in our ecosystem for legitimate purposes.

I am from these people. When using the Conda CLI, PYTHONHOME should be ignored. If I need to use Conda with PYTHONHOME, then I will do it in myconda.py

Set the status to Open of this bug.

PR #9239, opened 2019-09-13
Windows: Ignore all PYTHON* environment variables

Roffild:
https://github.com/conda/conda/issues/9230

Conda is a package and environment manager. Conda should not depend on PYTHON* variables.

Very often an error occurs if PYTHONHOME has the wrong path.
```
  File "c:\Anaconda\Scripts\conda-script.py", line 16, in <module>
    from conda.cli import main
```
This patch fixes this error.

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -765,8 +765,8 @@ def make_entry_point(target_path, conda_prefix, module, func):
         original_ep_content = ""
 
     if on_win:
-        # no shebang needed on windows
-        new_ep_content = ""
+        # Ignore all PYTHON* environment variables
+        new_ep_content = "#!python -E"
     else:
         new_ep_content = "#!%s\n" % join(conda_prefix, get_python_short_path())
 


msarahan:
This changes all entry points that conda ever creates - any package that has any entry point won't respect `PYTHON*`

My humble opinion is that the isolation gained here may be worthwhile, but it is also an incredibly dramatic and potentially breaking change that needs to be thought through carefully.

rkagerer:
Does anyone have ownership of this PR and its associated issue, or is it dead?

I tried to install Anaconda and Miniconda on Windows today, and neither worked, due to a lingering PYTHONHOME variable from a previous Python installation.  I kept getting `ModuleNotFoundError: No module named 'conda'` on opening a prompt.  As a new user this was tricky to diagnose.  Maybe I'm ignorant, but I kinda expect some better isolation.  Or a warning from the installer that this pre-existing variable on my machine might trip up the software.

mingwandroid:
It does. We aren't in control of what other python distros do to your environment. I simply cannot make or interpreters ignore these well documented standards. PYTHONPATH is necessary in many cases such as when using PyCharm.

mingwandroid:
And neither should conda ignore them. It is used in my conda development and testing workflows where I use PyCharm.

Roffild:
I use this modification all the time.

PYTHONHOME and PYTHONPATH are part of Python development.
It is quite normal when PYTHONHOME and PYTHONPATH indicate a specific environment in which Conda is most often absent. If PYTHONHOME is installed, then the development environment can pick up this environment automatically (VS Code). I use PYTHONHOME to build my module for Python.

Conda is a tool for setting up the Python environment and should not depend on system settings. It is c:\Miniconda\Scripts\activate.bat that is used to select the desired environment with the installation of the necessary global variables.

As long as Conda is dependent on PYTHONHOME and PYTHONPATH, it is not possible to create or fix an independent Python environment.

mingwandroid:
> PYTHONHOME and PYTHONPATH are part of Python development

It seems to me they are a core part of Python. Here they are described: https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH

This variable is to be used for whatever reason wants to use them (including breaking their ability to import things .. these are sharp tools and should be used with care) and I cannot in good conscience do anything that overrides the documented behavior of Python.

I agree that they are often used as part of Python development (note: I am quite sure this doesn *not* mean developing the python interpreter itself, which could be implied from your wording here, if it were to be used for that purpose it would be documented in the hacking docs instead).

I use them when developing conda itself, they are used by IDEs that someone may care to hack on conda in.I know if I were a potential contributor and I encountered such a hack in conda that prevented my normal dev workflows from working, and I'd spent any significant amount of time trying to figure out why, I'd likely find other projects to hack on.

Even if I could contemplate making such a drastic and surprising change here, it will not be done in the way it's being done here (see https://github.com/conda/conda/pull/9239#issuecomment-531403767).

I *do* support the idea of conda isolation, but it needs to be done comprehensively and *disabled* by default. I also support adding code to conda's early initialization to detect and warn when:

1. These variables are set.
2. The values they are set to are *likely* to cause import failures.

If you can revise this so that it is:

1. Optional based on a conda config var.
2. Works on any OS instead of just Windows.
3. Is documented.

Then I'll be happy to review it again, much more favourably.

Roffild:
The problem with PYTHONHOME and PYTHONPATH is only for Conda, because it is in an isolated place. To create a complete environment, Conda must set or delete PYTHONHOME and PYTHONPATH through "Scripts\activate.bat" or "conda run", and not depend on PYTHONHOME and PYTHONPATH.

If during the development of Conda there is a problem with PYTHONHOME and PYTHONPATH, then you need to divide it into Debug and Release. Conda is an unusual package for Python. Regular packages have no problems with PYTHONHOME and PYTHONPATH and never will.

Now my fix affects only **Windows**. For other OS, nothing will change.

mingwandroid:
Conda does not depend on either PYTHONHOME and PYTHONPATH, either on Windows or on any other OS. Our Python interpreter does however, respect those values and our packages and libraries should  not try to subvert that, not even conda.

You haven't explained how it's different on other OSes. I assert it is not. Also, you haven't addressed any of my suggestions, nor the concerns of @msarahan that this change breaks all of our entry points.

rkagerer:
I really appreciate all you guys chiming in with thoughtful responses after I gave this thread a kick.

It sounds like a warning would be a great stopgap.

The Conda docs landing page claims "With just a few commands, you can set up a totally separate environment to run that different version of Python, while continuing to run your usual version of Python in your normal environment."

It seems to me these environment variables breach (or at least leak into) the isolation guarantees inferred by the statement, and it sounds like that's intentional.  I'm not familiar enough with the Python ecosystem to lend any sense of authority to the conversation; I just assumed when I clicked Anaconda Command Prompt that the particular shell instance launched would come up as an isolated and working environment.  A warning could at least help save some time for others who encounter the same issue.  I'd code it myself but I fear I'm too green in this context to be helpful.

Roffild:
> It seems to me these environment variables breach (or at least leak into) the isolation guarantees inferred by the statement, and it sounds like that's intentional. I'm not familiar enough with the Python ecosystem to lend any sense of authority to the conversation; I just assumed when I clicked Anaconda Command Prompt that the particular shell instance launched would come up as an isolated and working environment. 

Yes, this is a violation of the isolation of the environment.
PYTHONHOME and PYTHONPATH may indicate the main working environment, but should not be accessible to the test environment. The -E flag allows you to do this.

> Conda does not depend on either PYTHONHOME or PYTHONPATH, either on Windows or on any other OS. Our Python interpreter does however, respect those values and our packages and libraries should not try to subvert that, not even conda.

Consider this from a security perspective.
In the test environment of Python is a hacked version of Conda. PYTHONHOME and PYTHONPATH refer to this environment. When launching "conda install", the version installed in the PYTHONHOME environment, and not located in the system folder, will be used. The error "ModuleNotFoundError: No module named 'conda'" will not be because the hacked version of Conda will be used.
This is possible because the -E flag is not used.

> You haven't explained how it's different on other OSes.

For other OSes, the -E flag is also required. But I'm not sure about the additional changes, so my patch is only for Windows.

rkagerer:
In the meantime, I solved this in my own environment by adjusting `C:\ProgramData\Miniconda3\Scripts\activate.bat` to clear the `PYTHONHOME` and `PYTHONPATH` environment variables, and I moved the six new path entries here so they don't pollute my system when I'm not doing conda stuff.  Note on Windows, those changes only apply to the **current** shell session - so my it shouldn't affect anything else, including any non-conda python installations on my machine.  Let me know if you foresee any problems (other than me having to repeat this every time I update your software).  I'd be happy to do some testing and propose it as a separate PR if it seems worthwhile.

```
@REM Added by RK --------------------------------------------------------------------
@echo off
REM Determine the Miniconda installation directory. Get the location of this bat file
REM and strip off the last 21 characters corresponding to "\Scripts\activate.bat".
REM WARNING: Update the 21 value if the location or name of this bat file changes.
SET BASEPATH=%0
SET BASEPATH=%BASEPATH:~0,-21%

REM Append path entries required by Miniconda
SET PATH=%BASEPATH%\Library\usr\bin;%BASEPATH%\Library\bin;%BASEPATH%\Scripts;%PATH%
SET PATH=%BASEPATH%\condabin;%BASEPATH%\Library\mingw-w64\bin;%PATH%

REM Clear global environment variables that break Miniconda
REM Note this only impacts the current terminal session
SET PYTHONPATH=
SET PYTHONHOME=
REM ---------------------------------------------------------------------------------
```

I noticed on Windows, Conda uses an "ugly" patchwork of several BAT files.  Incidentally, on my fresh install, the "Anaconda Prompt" shortcut is broken - it points to a non-existing `activate.bat` in `C:\Program Files (x86)\Microsoft Visual Studio\Shared\Anaconda3_64\Scripts\` even though the rest of the files seem to be there.

The "Anaconda Prompt (Miniconda3)" shortcut, which uses the file edited above, is fine.

Finally, is there a reason all the bat files use the `@` prefix on every line, instead of the arguably cleaner and more idiomatic practice of simply starting the file with `@echo off`?

mingwandroid:
I approve more of doing it in the correct entry points (always a shell script of some sort) as you've done, yes. Still, I'd rather see warnings printed and a global env var queried at least that allows avoiding this stuff.

Roffild:
What usage scenario does Conda use for the PYTHONHOME global variable?
Excluding the installation of packages through hacked Conda.

I am for the use of the -E flag for Conda always.

mingwandroid:
PYTHONPATH and PYTHONHOME and the rest of those variables are used by people investigating problem with conda. They are set in PyCharm so it can find source code, you might want to break into a running conda session with PyCharm and need to have set some of the PYTHON* variables in your debugging workflow.

We can't just disable powerful debugging features so globally. There are various tests in the testsuite that run old *packaged* versions of conda to upgradability and if I ever need to debug into those condas but their entry points are doing this, I will be unable to do that.

Doing this in batch files, I'm more comfortable with. Batch files and shell scripts are the only and correct entry points to conda if you want it to behave correctly, particularly on Windows. I still want it off by default though behind an env. var, such as `CONDA_ISOLATION_ENABLE` which people are free to set if they need to. More importantly, having warnings emitted when conda discovers these variables are set to something other than the default is, IMHO a necessary first step here, and those could and should be implemented in the conda python code.

Cheers.

Roffild:
This patch changes the behavior of ONLY the first command.

Ignore PYTHONHOME and PYTHONPATH:
```bat
Scripts\conda.exe install -h

python -E Scripts\conda-script.py install -h

python -E -m conda install -h
```
Use PYTHONHOME and PYTHONPATH:
```bat
python Scripts\conda-script.py install -h

python -m conda install -h
```
> PYTHONPATH and PYTHONHOME and the rest of those variables are used by people investigating problems with conda (like me). They are set in PyCharm so it can find source code and modules, you might want to break into a running conda session with PyCharm and need to have set some of the PYTHON* variables in that process for your debugging workflow.

This patch does not affect development in PyCharm. By default, PyCharm uses the fourth command to run, not the first.

travishathaway:

PR #9249, opened 2019-09-17
Protect against cases where subcommand help fails

tartansandal:
* fixes some edge cases where attempts to complete --options leaves the cache in a badly broken state.
* handle cases where conda is inactive or only partially setup more gracefully
* workaround obscure bash_completion dynamic loading bug
* ensure completion works even if the base environment is inactive
* always provide completion for core commands

### DIFF
--- a/conda/shell/etc/bash_completion.d/conda
+++ b/conda/shell/etc/bash_completion.d/conda
@@ -42,11 +42,17 @@ function __comp_conda_ensure_root() {
 }
 
 function __comp_conda_commands () {
+    # default core commands
+    echo clean config create help info init install list package
+    echo remove uninstall run search update upgrade
+
+    # check commands from full anaconda install
     for f in $CONDA_SOURCE/cli/main_*.py
     do
         \expr match "$f" '.*_\([a-z]\+\)\.py$'
     done
 
+    # check extra pluggins
     for f in $CONDA_ROOT/bin/conda-*
     do
         if test -x "$f" -a ! -d "$f"
@@ -54,8 +60,9 @@ function __comp_conda_commands () {
             \expr match "$f" '^.*/conda-\(.*\)'
         fi
     done
-    echo activate
-    echo deactivate
+
+    # implied by conda shell function
+    echo activate deactivate
 }
 
 function __comp_conda_env_commands() {
--- a/conda/shell/etc/bash_completion.d/conda
+++ b/conda/shell/etc/bash_completion.d/conda
@@ -31,14 +31,13 @@ function __comp_conda_ensure_root() {
             CONDA_ROOT=$(\dirname "${CONDA_EXE}")
             CONDA_ROOT=$(\dirname "${CONDA_ROOT}")
         fi
-        CONDA_SOURCE=$(
-            cat - <<'            EOF' | sed 's/^ *: //g' | python -
-            : from __future__ import print_function
-            : import os
-            : import conda
-            : print(os.path.dirname(conda.__file__))
-            EOF
-        )
+        \local script="
+        :    from __future__ import print_function
+        :    import os
+        :    import conda
+        :    print(os.path.dirname(conda.__file__))
+        "
+        CONDA_SOURCE=$( python -c "${script//        :    /}" )
     fi
 }
 
@@ -67,17 +66,15 @@ function __comp_conda_env_commands() {
 }
 
 function __comp_conda_envs() {
-    script=$(
-        cat - <<'        EOF' | sed 's/^ *: //g'
-        : from __future__ import print_function;
-        : import json, os, sys;
-        : from os.path import isdir, join;
-        : print('\n'.join(
-        :    d for ed in json.load(sys.stdin)['envs_dirs'] if isdir(ed)
-        :    for d in os.listdir(ed) if isdir(join(ed, d))));
-        EOF
-    )
-    conda config --json --show envs_dirs | python -c "$script"
+    \local script="
+    :    from __future__ import print_function;
+    :    import json, os, sys;
+    :    from os.path import isdir, join;
+    :    print('\n'.join(
+    :       d for ed in json.load(sys.stdin)['envs_dirs'] if isdir(ed)
+    :       for d in os.listdir(ed) if isdir(join(ed, d))));
+    "
+    conda config --json --show envs_dirs | python -c "${script//    :    /}"
 }
 
 function __comp_conda_packages() {
--- a/conda/shell/etc/bash_completion.d/conda
+++ b/conda/shell/etc/bash_completion.d/conda
@@ -37,7 +37,8 @@ function __comp_conda_ensure_root() {
         :    import conda
         :    print(os.path.dirname(conda.__file__))
         "
-        CONDA_SOURCE=$( python -c "${script//        :    /}" )
+        # don't assume an active base environment
+        CONDA_SOURCE=$(conda activate base; python -c "${script//        :    /}")
     fi
 }
 
--- a/conda/shell/etc/bash_completion.d/conda
+++ b/conda/shell/etc/bash_completion.d/conda
@@ -104,6 +104,30 @@ function __comp_conda_cache_dump() {
     done
 }
 
+function __comp_conda_option_lookup() {
+    \local word_list cmd_str cmd_key
+    cmd_str=$1
+
+    # make a key to look up the cached result of the command help
+    # (We should be able to just use $cmd_str, since spaces in an array key are fine,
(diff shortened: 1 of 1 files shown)


imsodin:
I hope it's ok to use this PR for feedback. When using the file from this PR and having a conda env activated, I get the following error on hitting tag:
```
$ conda config --debash: conda config: syntax error in expression (error token is "config")
```
Thanks again for your work on this!

tartansandal:
> I hope it's ok to use this PR for feedback. 

Really appreciate the feedback and this is definitely the best place for it. :-)

The error you are seeing is just the kind of bug that this PR is trying to address. 
I can replicate this issue with the old code, but when I load the new code I get no error. 
Perhaps the new code is not being loaded. Easy way to check. Try running:
```
__comp_conda_cache_dump
```
in you shell. If that function not defined, then the new code has not been loaded. Could you check that for me please?  Let me know how you go and I can help troubleshoot the loading issue.

imsodin:
`__comp_conda_cache_dump` is available and I rechecked that my completion file is identical with the one in your PR branch - I still get that error. Let me know if I can do something to help you narrow this down/reproduce it.

tartansandal:
Thanks for that @imsodin 
Damn. Not as simple as I'd hoped. :-( 

There is a small possibility that you have two completion files, with one being sourced after the other, with the latter masking some of the functions.  If you run the following in your shell:
```
type _comp_conda | head -6
```
you should get
```
comp_conda is a function
_comp_conda () 
{ 
    \local cur prev words cword;
    _init_completion || return;
    __comp_conda_ensure_root 2> /dev/null;
```
The last line in particular. Could you check that please?

If it's not the same, then something is going wrong in the block starting at line 124. Will have to think some more about how we could debug that.

imsodin:
The output I get is exactly what you posted above.

tartansandal:
Thanks for that. In the meantime I think I've tracked down the bug. Or at least I have been able to replicate it.

tartansandal:
Hi @imsodin. I think I've fixed this, or at least avoided the problem. Could you please test the code on the current branch? 

I managed to isolate the bug to a single expression, but it only happens during bash_completion's autoloading of new completions.  If I load the (previous) code manually after starting a new shell, the bug does not show.  Really weird.  Might see if I can generate a replicable case for the bash_completion people.

imsodin:
Just tested and it works indeed now - thanks.

Another thing that I noticed now - I can't say whether it was the case before the latest update too: Many commands do not complete (e.g. `conda sear`, `conda upgr`, ... does not complete, while `conda act`, `conda en` do).

tartansandal:
Yay! I think I deserve another glass of wine ... :-)

Re: your question about completing 'conda se' etc.   We use a heuristic were we look around your installation and try to track down commands. This was translated over to bash from the `fish` completion code. Obviously it's not perfect -- but it could be made better.  The relevant bash function is `__comp_conda_commands` which, for me produces:
```
clean
config
create
help
info
init
install
list
package
pip
remove
run
search
update
build
convert
debug
develop
env
index
inspect
metapackage
render
server
skeleton
verify
activate
deactivate
```
Note that `pip` should not be there, but the rest seem fine.

You are getting much less than that, which suggests the environment variables $CONDA_SOUCE and $CONDA_ROOT may not be getting the best values. For me I get
```
$ echo $CONDA_SOURCE 
/home/kal/GitHub/conda/conda
$ ls -1 $CONDA_SOURCE 
activate.py
api.py
base/
cli/
common/
compat.py
core/
exceptions.py
exceptions.pyc
...
```
and
```
$ echo $CONDA_ROOT 
/home/kal/anaconda3
ls -1 $CONDA_ROOT 
bin/
compiler_compat/
condabin/
conda-meta/
doc/
envs/
etc/
...
```
Are you getting anything like that?  (Off to bed now so responses may be delayed)

imsodin:
> Yay! I think I deserve another glass of wine ... :-)

Definitely ;)

```
$ __comp_conda_commands 

env
activate
deactivate
$ echo $CONDA_SOURCE 

$ echo $CONDA_ROOT 
/home/simon/Linux/miniconda3
```

So the issue seems that I don't have the conda source checked out. However I don't think you can expect the "general user" to have that.

> [...]  (Off to bed now so responses may be delayed)

Good night. And don't worry about responsiveness, daily responses would already be super quick ;)

tartansandal:
Opps. Missed that I was running that from my git checkout. The bash completion picked up that I was in a conda development environment and set `CONDA_SOURCE` appropriately. The following is more appropriate:
```
$ echo $CONDA_SOURCE 
/home/kal/anaconda3/lib/python3.7/site-packages/conda
$ echo $CONDA_ROOT 
/home/kal/anaconda3
```
So I'm thinking your `CONDA_SOURCE` is not being set in a `miniconda` environment. Hmmm maybe `CONDA_EXE is` not being set in `miniconda`
```
$ echo $CONDA_EXE
/home/kal/anaconda3/bin/conda
``` 
Do you get a sensible value for that?

Full disclosure: I'm not a conda expert :-). I only installed it a couple months ago so I could work through a SciKit Learn tutorial and have only used its basic features so far. Have not looked at the `miniconda` layout yet, so your help with this is much appreciated.

imsodin:
Yes, `CONDA_EXE` is set:
```
$ echo $CONDA_EXE
/home/simon/Linux/miniconda3/bin/conda
```
As far as I understand so far anaconda is just miniconda (which is just an installer for conda?) with a lot of packages preinstalled. That is, I am not an expert either, I just started to use it because my team does and I am currently exploring organising/tooling some of the development happening (science :) ). I don't right now have time to dive into it, but I am happy helping to debug.

tartansandal:
Let try the following sequence:
```
$ . ~/.local/share/bash-completion/completions/conda
$ unset CONDA_SOURCE
$ __comp_conda_ensure_root
$ echo $CONDA_SOURCE
/home/kal/anaconda3/lib/python3.7/site-packages/conda
```
you should be getting a similar result with no error output.

If you still just get an empty string for `CONDA_SOURCE`, then I'm thinking miniconda installs conda as an egg or a wheel. If that's the case, we can probably fallback to the core conda subcommands.

imsodin:
The problem still exists:

```
$ __comp_conda_ensure_root 
  File "<string>", line 4, in <module>
$ echo $CONDA_SOURCE

```

Conda source does exist in the miniconda environment:

```
$ ls ~/Linux/miniconda3/lib/python3.7/site-packages/conda
activate.py  compat.py      history.py       __main__.py  resolve.py
api.py       core           __init__.py      misc.py      shell
base         exceptions.py  install.py       models       utils.py
cli          exports.py     instructions.py  plan.py      _vendor
common       gateways       lock.py          __pycache__
```

What "fixes" the issue is install conda inside a running environment:

```
$ __comp_conda_ensure_root 
$ echo $CONDA_SOURCE 
/home/simon/Linux/miniconda3/envs/test/lib/python3.6/site-packages/conda
```

However you don't actually need conda installed in an environment to use the cmd-line tool - so in practice it won't be.

tartansandal:
Ah, I see now. Good catch. I've been assuming the 'base' environment has been activated. This is what 
```
conda init bash
```
would have your `.bashrc` set up. By default this also activates the 'base' environment -- the top level environment used by the `conda` command itself -- but you have to restart your shell or 
```
source bashrc
``` 
to see this activation. Or you can just do
```
conda activate
```
However, depending on this default is obviously not the best way to handle the setup, especially when you can unset the `auto_activate_base` configuration variable.

The newest commit does not make that assumption.

imsodin:
Nice, works fine now. Before I activate any conda env, `$CONDA_SOURCE` is unset, but completion still works (e.g. completing `conda config --de`) and after activating one, it points to the expected path (`/home/simon/Linux/miniconda3/lib/python3.7/site-packages/conda`).

I should have mentioned that I don't automatically activate the base env. Which if I remember correctly is the default in miniconda - and in my opinion that's the only sane default. Automatically enabling a base environment just makes sure people won't encounter and thus learn about environments and then are at a complete loss the first time they encounter some dependency breakage (and lets be real, if you use more than e.g. numpy and matplotlib, you will eventually encounter it) - sorry for the OT rant :)

Thanks again for your work - it's a good experience to for once be on the other (reporting) side of debugging :)

tartansandal:
Excellent! Thanks for all your help with this :-)

tartansandal:
Work on improving this completion script has been moved to https://github.com/tartansandal/conda-bash-completion.

The script and its dependencies can now be installed as a conda package: https://anaconda.org/tartansandal/conda-bash-completion.

See the PR https://github.com/conda/conda/pull/9355 for the discussion that lead to this change.

PR #9266, opened 2019-09-21
Fixup package version in dev builds.

asford:
Write `conda/.version` file using `setup.py --version`, rather than
echoing conda build string, to ensure creation of [pep 440](https://www.python.org/dev/peps/pep-0440/) compliant version
strings. This has no effect for tagged releases, but ensures that
development releases built as conda packages present a version string
compatible with conda-builds's [conda_interface](https://github.com/conda/conda-build/blob/master/conda_build/conda_interface.py#L11).

### DIFF
--- a/conda/_vendor/auxlib/packaging.py
+++ b/conda/_vendor/auxlib/packaging.py
@@ -86,6 +86,12 @@ GIT_DESCRIBE_REGEX = compile(r"(?:[_-a-zA-Z]*)"
                              r"(?P<version>[a-zA-Z0-9.]+)"
                              r"(?:-(?P<post>\d+)-g(?P<hash>[0-9a-f]{7,}))$")
 
+# Regex matching version strings templated from conda, of the form:
+# {{ GIT_DESCRIBE_TAG }}.{{ GIT_BUILD_STR }}
+FILE_VERSION_REGEX = compile(r"(?:[_-a-zA-Z]*)"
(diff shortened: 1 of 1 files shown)


mingwandroid:
What happens here if git is missing?

asford:
I've checked this behavior via a direct download of the conda release from github.

In the immediate recipe the build errors before hitting `build.sh` due to the use of [GIT_DESCRIBE_TAG in meta.yaml](https://github.com/conda/conda/blob/4.7.12/conda.recipe/meta.yaml#L3), so when building directly from the repo this change will have no effect. 

However, `setup.py --version` does not render a valid build string when working from a source tarball. This will cause an issue with building via an external feedstock that may reuse this recipe, and should be addressed before this PR  is merged. 

This problem has been tackled in a few different ways in other projects, the cleanest being something like [setuptools_scm](https://pypi.org/project/setuptools-scm/), which handles extracting a version string from git _or_ from a source archive via [setuptools-scm-git-archive](https://pypi.org/project/setuptools-scm-git-archive/). Changing the versioning library may have unanticipated downstream consequences (like the one being addressed here), so I don't thinking moving to `setuptools_scm` is necessarily the best option. 

I've updated the vendored packaging tool to parse the conda-derived version number generated in `meta.yaml` to produce pep440-compliant strings.

travishathaway:

PR #9267, opened 2019-09-22
Update download link (silent mode)

jpigla:
We are installing Miniconda for MacOS but the linkt to download the lastest MacOS version was pointing to the Linux version.


msarahan:
Good catch, Thanks!

msarahan:
@cla-bot check

Issue #9280, opened 2019-09-25
conda deactivate mangles PATH environment if passed an argument


douglowe:
## Current Behavior

If conda deactivate is passed an argument (such as the name of the environment that you are trying to deactivate), then it will add the base conda bin directory to the PATH environment, as well as throwing up the error about not accepting arguments.

Possibly conda is making other changes to the environment too, however I have only been looking at changes in the PATH environment, so have not seen anything else.

I've tested this on both macOS and linux (ubuntu), and the behaviour is consistent across both these platforms.

### Steps to Reproduce

```
echo $PATH
conda activate [environment]
echo $PATH
conda deactivate [environment]
echo $PATH
conda deactivate
echo $PATH
```

Results:
```
mbessdl2$ echo $PATH
/Users/mbessdl2/anaconda3/condabin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/munki:/opt/X11/bin
mbessdl2$ conda activate sciwork
(sciwork) mbessdl2$ echo $PATH
/Users/mbessdl2/anaconda3/envs/sciwork/bin:/Users/mbessdl2/anaconda3/condabin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/munki:/opt/X11/bin
(sciwork) mbessdl2$ conda deactivate sciwork
deactivate does not accept arguments
remainder_args: ['sciwork']

(sciwork) mbessdl2$ echo $PATH
/Users/mbessdl2/anaconda3/bin:/Users/mbessdl2/anaconda3/envs/sciwork/bin:/Users/mbessdl2/anaconda3/condabin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/munki:/opt/X11/bin
(sciwork) mbessdl2$ conda deactivate
mbessdl2$ echo $PATH
/Users/mbessdl2/anaconda3/bin:/Users/mbessdl2/anaconda3/condabin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/munki:/opt/X11/bin
```

## Expected Behavior
The PATH environment should not be changed if conda deactivate is not successfully run.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : sciwork
    active env location : /Users/mbessdl2/anaconda3/envs/sciwork
            shell level : 1
       user config file : /Users/mbessdl2/.condarc
[...]

PR #9282, opened 2019-09-25
Do not add sys_prefix_path for failed activation/deactivation

soapy1:
Resolves: https://github.com/conda/conda/issues/9280


mingwandroid:
I'm pretty sure we need this @soapy1! This will break Windows (MSYS2, Git for Windows).

Instead, you need to back up and restore I think.

soapy1:
@mingwandroid oooooo, tricky tricky windows! Pushed up another go at a fix

mingwandroid:
LGTM!

PR #9289, opened 2019-09-26
do not memorize PackageRecord.combined_depends

jjhelmus:
Change combined_depends from a memoizedproperty to a property.
Memoization was causing problems with conda-build when building against
a matrix using a conda_build_config.yaml file,
c.f. conda/conda-build#3735

### DIFF
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -21,7 +21,6 @@ from .match_spec import MatchSpec
 from .._vendor.auxlib.entity import (BooleanField, ComposableField, DictSafeMixin, Entity,
                                      EnumField, IntegerField, ListField, NumberField,
                                      StringField)
-from .._vendor.auxlib.decorators import memoizedproperty
 from .._vendor.boltons.timeutils import dt_to_timestamp, isoparse
 from ..base.context import context
 from ..common.compat import isiterable, itervalues, string_types, text_type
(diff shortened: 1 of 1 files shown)


msarahan:
Dang, too bad.  I hope the speed hit isn't too bad.  Better correct than fast, though.  Can we add the reproducer code as a test here?

jjhelmus:
> Can we add the reproducer code as a test here?

I need to spend a bit more time with the reproducer in the conda-build issue to cut it down to something less complicated.  Hopefully this will lead to a better understanding of why memorization cannot be done here because it looks harmless (and beneficial) to me.

jakirkham:
Thanks for working on this @jjhelmus!

When is the next release planned? We are interested in trying out this feature in conda-forge. Should let us use a newer `conda` in `staged-recipes`. 🙂

s22chan:
https://github.com/conda/conda-build/issues/3735 has a simpler recipe where `boto3` is either the latest version or an old version

npanpaliya:
Which next conda version will have this fix?

jakirkham:
Looks like this was tagged as `4.8.0rc0`. On [`conda-canary`]( https://anaconda.org/conda-canary ) there is [a package for it]( https://anaconda.org/conda-canary/conda/files?version=4.8.0rc0 ). Maybe you can give that a go? 😉

npanpaliya:
Thanks @jakirkham . Yes, I verified 4.8.0rc0 and it worked for me.

PR #9303, opened 2019-10-02
Avoid clash-prone OLDPATH variable name in activation script

jooh:
Users can now modify the PATH variable in the env_vars.sh script (#8464), which is useful for managing non-conda binaries. It is all too tempting to store your old path as $OLDPATH in the activation script, and set `PATH="$OLDPATH"` in the deactivation script. But this won't have the intended effect because conda.sh already uses that name. This modest PR proposes a less clash-prone name.

Hopefully this will save someone else a day of total confusion.

### DIFF
--- a/conda/shell/etc/profile.d/conda.sh
+++ b/conda/shell/etc/profile.d/conda.sh
@@ -45,20 +45,20 @@ __conda_activate() {
     \local cmd="$1"
     shift
     \local ask_conda
-    OLDPATH="${PATH}"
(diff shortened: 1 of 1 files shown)


soapy1:
Thanks for your contribution! 
@cla-bot check

cchriste:
This seems to do what we want, but the PATH after the deactivate script(s) run gets reset to what was create _after_ the activate scripts(s) have run. I've seen a few comments about this, but no documentation for a resolution. Any pointers appreciated.

Issue #9312, opened 2019-10-04
conda should retry failed downloads


ChemicalSailor:
**I'm submitting a...**
  - [ ] bug report
  - [x] feature request

## Current Behavior

When creating a new environment and downloading packages if an HTTP error is encountered the environment is not created (as expected). I have to manually retry in order for all packages to get downloaded and the environment created/updated. This is an issue for me because conda sits behind a corporate firewall that scans suspicious downloads (sophos sandstorm) which often initially blocks a download with a 403 forbidden error, but once it's had time to scan the package, retrying the download gives a 200 response.
related: https://github.com/conda/conda/issues/8477

### Steps to Reproduce

If you have a firewall that temporarily blocks packages, you can reproduce this behavior.

## Expected Behavior

conda should have the option to automatically retry failed downloads after a delay, up to a maximum number of tries. Incorporating some functionality of the retrying package would be great!

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
active environment : my-default
active env location : P:\python_envs\my-default
shell level : 2
conda version : 4.6.7
conda-build version : 3.17.6
python version : 3.7.1.final.0
base environment : C:\ProgramData\Anaconda3  (writable)
```
</p></details>

soapy1:
Resolved by https://github.com/conda/conda/pull/9318 Thanks @WillyChen123 !

PR #9318, opened 2019-10-07
Retry download

WillyChen123:
#9312  

``Retry`` class can pass to ``HTTPAdapter`` instead of passing parameter ``max_retries``.
There is an option ``backoff_factor`` in ``Retry`` which can determine the time between two attempts.

### Fix
I add an option ``remote_backoff_factor`` in context and use ``Retry`` class for ``HTTPAdapter`` instead in order to pass the option ``backoff_factor`` in ``Retry`` class.

### Test
Because I cannot reproduce the error "403", I test in the following situation:
1. Run ``conda config --set remote_max_retries 3`` and ``conda config --set remote_backoff_factor 1`` before downloading
2. Execute ``conda install <package>``
3. Before asking ``proceed([y]/N)`` , use ``iptables`` to disable network
4. Enter ``y`` to proceed
5. It waits for about 5 seconds and raised an uncaught exception
    *   Originally, ``conda`` raised an uncaught exception almost immediately
    *   If I use  ``iptables`` to enable network in 5 seconds, it starts downloading

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -201,6 +201,7 @@ class Context(Configuration):
     remote_connect_timeout_secs = PrimitiveParameter(9.15)
     remote_read_timeout_secs = PrimitiveParameter(60.)
     remote_max_retries = PrimitiveParameter(3)
+    remote_backoff_factor = PrimitiveParameter(0)
 
     add_anaconda_token = PrimitiveParameter(True, aliases=('add_binstar_token',))
 
@@ -803,6 +804,7 @@ class Context(Configuration):
             'proxy_servers',
             'remote_connect_timeout_secs',
             'remote_max_retries',
+            'remote_backoff_factor',
             'remote_read_timeout_secs',
             'ssl_verify',
         )),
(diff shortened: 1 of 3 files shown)


soapy1:
Thanks for the contribution, this is awesome!
@cla-bot check

conda/base/context.py
```diff
@@ -201,6 +201,7 @@ class Context(Configuration):
     remote_connect_timeout_secs = PrimitiveParameter(9.15)
     remote_read_timeout_secs = PrimitiveParameter(60.)
     remote_max_retries = PrimitiveParameter(3)
+    remote_backoff_factor = PrimitiveParameter(0)
```
soapy1: Could this default value be changed to 1 so that users can benefit from this change by default.

WillyChen123: This is a great suggestion. 😀
I will change the default value to 1.
( I set it to 0 originally because it is the default value in ``Retry`` class. 😀)
Thank you for giving me suggestion. 😊

PR #9325, opened 2019-10-11
Pass specs as query parameters when fetching repodata

jeremyjliu:
Pass package names as query parameters to repodata GET request. This allows for channels that can do server side repodata filtering to enable that process.

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -434,6 +434,10 @@ class MatchSpec(object):
         return self.conda_build_form()
 
     @property
+    def to_query_param(self):
+        return self.get_exact_value('name')
+
+    @property
     def version(self):
         # in the old MatchSpec object, version was a VersionSpec, not a str
         # so we'll keep that API here
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -370,6 +370,7 @@ class PackageRecord(DictSafeMixin, Entity):
         #              channel_name/subdir:namespace:name-version-build_number-build_string
         return "%s/%s::%s-%s-%s" % (self.channel.name, self.subdir,
                                     self.name, self.version, self.build)
+
     def to_query_param(self):
         return self.name
 
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -370,6 +370,8 @@ class PackageRecord(DictSafeMixin, Entity):
         #              channel_name/subdir:namespace:name-version-build_number-build_string
         return "%s/%s::%s-%s-%s" % (self.channel.name, self.subdir,
                                     self.name, self.version, self.build)
+    def to_query_param(self):
+        return self.name
 
 
 class Md5Field(StringField):
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ miniconda*.exe
 docs/source/_build
 **/.vscode
 conda.tmp/
+*.iml
+*.ipr
+*.iws
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -36,6 +36,7 @@ requirements:
     - setuptools >=31.0.1
   run:
     - python
+    - backports.functools_lru_cache # [py<33]
     - conda-package-handling
     - enum34                       # [py<34]
     - futures >=3.0.0              # [py<34]
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -84,9 +84,10 @@ def pretty_package(prec):
 def print_package_info(packages):
     from ..core.subdir_data import SubdirData
     results = {}
+    specs = [MatchSpec(package) for package in packages]
     for package in packages:
         spec = MatchSpec(package)
-        results[package] = tuple(SubdirData.query_all(spec))
+        results[package] = tuple(SubdirData.query_all(spec, package_ref_or_match_specs=specs))
 
     if context.json:
         stdout_json({package: results[package] for package in packages})
--- a/conda/cli/main_search.py
+++ b/conda/cli/main_search.py
@@ -70,8 +70,9 @@ def execute(args, parser):
         spec_channel = spec.get_exact_value('channel')
         channel_urls = (spec_channel,) if spec_channel else context.channels
 
-        matches = sorted(SubdirData.query_all(spec, channel_urls, subdirs, package_ref_or_match_specs=[spec]),
-                         key=lambda rec: (rec.name, VersionOrder(rec.version), rec.build))
+        matches = sorted(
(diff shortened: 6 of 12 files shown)


conda/models/match_spec.py
```diff
@@ -433,6 +433,10 @@ def strictness(self):
     def spec(self):
         return self.conda_build_form()
 
+    @property
+    def to_query_param(self):
+        return self.get_exact_value('name')
```
jeremyjliu: currently only passes the package name as a query parameter, like in conda-metachannel. Is it worth passing the entire MatchSpec string representation instead? If so, how likely is the string representation to change? General concern is keeping compatibility with those changes.

conda/models/records.py
```diff
@@ -370,6 +370,8 @@ def record_id(self):
         #              channel_name/subdir:namespace:name-version-build_number-build_string
         return "%s/%s::%s-%s-%s" % (self.channel.name, self.subdir,
                                     self.name, self.version, self.build)
+    def to_query_param(self):
+        return self.name
```
jeremyjliu: the method signature of `SubdirData.queryAll` notes `package_ref_or_match_spec`, which I believe references `PackageRef` a legacy representation of `PackageRecord`. So, I believe this is necessary though I don't see any instance in the current code base that passes in `PackageRecord` to `queryAll`.

jeremyjliu:
@conda-core any thoughts on this?

msarahan:
This appears to be a purely additive change, and we think it will not be disruptive.  I'm concerned about the test failing with out-of-memory errors on circle CI.  I have re-run the tests to see if it was a fluke.

msarahan:
Seems happier this time.  It looks like you've introduced some flake8 issues.  Please straighten those out, and then I think this can be merged.

jeremyjliu:
Seems to have OOMed on the rerun, I'll take a look at it + flake8 issues.

msarahan:
That might work...  I think the problem is that your additions are effectively causing cache misses all over the place, and leading the cache to balloon.  I think we may need to try to limit the cache key somehow.

jeremyjliu:
Yup, that is the cause and the change I pushed only tries to ensure the tests don't OOM due to accumulation of the SubdirData cache (and assumes that in the field, the in-memory cache won't balloon too much).

Seems like the behavior we want is:
- in the case of querying a channel that is metachanneled, you'd want each SubdirData entry to keep track of the specs you passed
- otherwise, you'd want fallback to the original cache key

Immediate thoughts:
- global config parameter to turn on/off passing these specs and which cache key to use
- a setting to identify whether a channel is metachanneled or not
- cache key is per spec rather than per set of specs. then we could only fetch the missing specs and merge them together (in the case of the tests, this would be a good amount of redundant merging and would mean more network requests)

jeremyjliu:
@msarahan by "limit the cache key somehow" are you envisioning some sort of cache size limiting? e.g. an LRU cache seems appropriate (possibly in conjunction with having a per-channel setting to avoid extraneous cache entries)

msarahan:
Here's the cache key I was mentioning: https://github.com/conda/conda/pull/9325/files#diff-06975a3580f21bd34a35f056511af1afR67

What I was thinking was to just leave the specs off.  Unfortunately, I don't think that's really an option, because then you'd cache the first request and not any subsequent ones.  I don't think this OOM thing is that big of a deal.  Yes, it bites the tests pretty badly, but that's an extreme case for SubdirData.  In real-world scenarios, since conda is not a long-running process, you'll only have a few SubdirData instances around at any given time, and then they'll be gone after the conda operation completes.

LRU cache would be nice to avoid the hack in the tests, but let's only do it if it's really easy.

jeremyjliu:
> LRU cache would be nice to avoid the hack in the tests, but let's only do it if it's really easy.
I'll take a look, biasing towards simplest solution + ideally no new external deps.

jeremyjliu:
1) @msarahan I added a functools.lru_cache similar to conda-verify, had to adjust Channel's hash function to retain the existing cache key behavior. Note that this does add a backports.functools dependency for backcompat on py2. Let me know what you think of this - if it's too complex, I can take it back out.

2) I am getting an py3 integration test failure that I don't quite understand:
```
________ IntegrationTests.test_conda_pip_interop_conda_editable_package ________
  File "/home/circleci/conda/tests/test_create.py", line 1959, in test_conda_pip_interop_conda_editable_package
    assert "UNLINK" not in json_obj["actions"]
AssertionError: assert 'UNLINK' not in {'FETCH': [{'arch': 'x86_64', 'build': 'py27_0', 'build_number': 0, 'channel': 'https://repo.anaconda.com/pkgs/free/li...ase_url': 'https://repo.anaconda.com/pkgs/main', 'build_number': 0, 'build_string': '0', 'channel': 'pkgs/main', ...}]}
```
Looking closer there appears to be an unlink step for cacertificates, which seems unrelated to the intent of the test. Does not happen locally (mac-osx) or in py2 integration test. Any ideas?

3) In addition, the py2 appveyor/pr build appears to be executing integration tests after the unit tests and then timing out - is the appveyor/pr build supposed to be executing integration tests? (I don't see this behavior on other PRs).

jeremyjliu:
Having trouble replicating 2) locally or on a linux machine (centos 7) to iterate on debugging. Could I get rerun perms so I can rerun the job with ssh?

msarahan:
I don't know how to do that. I will start an ssh job now. If you don't read this in time, let's schedule something.

msarahan:
I realized that in order to give you ssh permission, you need to be a maintainer on the conda repo. Maybe it's easier for you to activate circle CI on your personal account and debug on your fork?

jeremyjliu:
The latter is a great idea, I'll try that now as I'm unable to ssh into the one that you've started for me.

msarahan:
I think not caching metachannels seems right. They are so small that caching doesn't buy much.

Fully supporting metachannels seems like a bit of a rabbit hope, but it's one that I think would be worthwhile. Tied up in there is the notion of per-channel configuration (to indicate here that a channel is a metachannels) but elsewhere for things like pretty-channel proxy settings.

If you'd like to undertake this, maybe it's best for us to start a new issue for discussion, and then build your work here into the larger effort. What do you think?

jeremyjliu:
Sure, I opened https://github.com/conda/conda/issues/9368 for discussion! That being said, what should be the end goal of this current PR?

msarahan:
I think the end goal of this PR hasn't changed, it has just exposed the need to have conda be aware of when a channel is a metachannel and when it is not.  This PR is fine, but it should be updated to use extra information that should be added in another PR for annotating which channels are metachannels.

PR #9327, opened 2019-10-11
Added threads, spec history, and config info

rrigdon:


docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -962,3 +965,56 @@ To remove a key, such as channels, and all of its values:
...
+
+Configuring number of threads
+=============================
+
+You can use your ``.condarc`` file or environment variables to
+add configuration to control the number of threads. You may
+want to do this to tweak conda to better utilize your system.
+If you have a very fast SSD, you might be able to shorten the
+time it takes for conda to create environments and
+install/remove packages.
```
soapy1: I think you should add something like; users will be able to do this by increasing the number of threads.

docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -962,3 +965,56 @@ To remove a key, such as channels, and all of its values:
...
+
+**repodata_threads**
+
+* Default: None
```
soapy1: I think instead of putting just `Default` here it would make more sense to put `Default number of threads` or something. For example, `Default: None` threads really means that the default number of threads is probably 1

docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -962,3 +965,56 @@ To remove a key, such as channels, and all of its values:
...
+You can use your ``.condarc`` file or environment variables to
+add configuration to control the number of threads. You may
+want to do this to tweak conda to better utilize your system.
+If you have a very fast SSD, you might increase the number
+of threads to shorten the time it takes for conda to create
+environments and install/remove packages. 
+
+**repodata_threads**
+
+* Default number of threads: None
```
soapy1:
```suggestion
* Default number of threads: 1
```

PR #9355, opened 2019-10-20
Add infra for testing shell completion

### LINKED ISSUE #9251: Tests and testing infra for tab completion
I'd like to build some tests for bash completion.  The completion functions have become complicated enough now for regressions to creep in while making even small changes. They also depends on the behaviour of `conda` command and its sub-commands, so external changes could lead to regressions as well.

There does not seem to be any existing infrastructure for doing this (at least none that I could find). We could probaly copy this approach  https://github.com/scop/bash-completion/blob/master/doc/testing.txt which is based on `pytest` and `pyexpect`.

tartansandal:
Closes #9251. Implements the hard parts of setting up a completion fixture. Runs some very simple tests against bash completion as proof of concept. Once this and https://github.com/conda/conda/pull/9249 are merged, I can flesh out the `bash` tests and perhaps start on tests for `fish`, or even `zsh`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ markers =
     integration: integration tests that usually require an internet connect
     slow: slow running tests
     installed: tests that assume conda is pre-installed on PATH
+    complete: tests involving shell completion
 
 
 [pycodestyle]
--- a/tests/shell/config/bashrc
+++ b/tests/shell/config/bashrc
@@ -1,0 +1,53 @@
+# bashrc file for bash-completion test suite
+
+# Note that we do some initialization that would be too late to do here in
+# conftest.py.
+
+# Use emacs key bindings
+set -o emacs
+
+# Use bash strict mode
+set -o posix
+
+# Unset `command_not_found_handle' as defined on Debian/Ubuntu, because this
+# troubles and slows down testing
+unset -f command_not_found_handle
+
+export PS2='> '
+
+# For clean test state, avoid sourcing user's ~/.bash_completion
+export BASH_COMPLETION_USER_FILE=/dev/null
+
+# ...and avoid stuff in BASH_COMPLETION_USER_DIR and system install locations
+# overriding in-tree completions. Setting the user dir would otherwise suffice,
+# but simple xspec completions are only installed if a separate one is not
+# found in any completion dirs. Therefore we also point the "system" dirs to
+# locations that should not yield valid completions and helpers paths either.
+export BASH_COMPLETION_USER_DIR=$TESTDIR
+# /var/empty isn't necessarily actually always empty :P
+export BASH_COMPLETION_COMPAT_DIR=/var/empty/bash_completion.d
+export XDG_DATA_DIRS=/var/empty
+
+## Ensure bash completion is available
+# try Linux location
+if [[ -f /usr/share/bash-completion/bash_completion ]]; then
+    . /usr/share/bash-completion/bash_completion
+# try MacOS location
+elif [[ -f /sw/etc/bash_completion ]]; then
+   . /sw/etc/bash_completion
+fi
+
+# >>> conda initialize >>>
+# !! Contents within this block are managed by 'conda init' !!
+__conda_setup="$("$TESTENVDIR/bin/conda" 'shell.bash' 'hook' 2> /dev/null)"
+if [ $? -eq 0 ]; then
+    eval "$__conda_setup"
+else
+    if [ -f "$TESTENVDIR/etc/profile.d/conda.sh" ]; then
+        . "$TESTENVDIR/etc/profile.d/conda.sh"
+    else
+        export PATH="$TESTENVDIR/bin:$PATH"
+    fi
+fi
+unset __conda_setup
+# <<< conda initialize <<<
--- a/tests/shell/config/inputrc
+++ b/tests/shell/config/inputrc
@@ -1,0 +1,13 @@
+# Readline init file for conda bash-completion test suite
+# See: info readline
+
+	# Press TAB once (instead of twice) to auto-complete
+set show-all-if-ambiguous on
+	# No bell.  No ^G in output
+set bell-style none
+	# Don't query user about viewing the number of possible completions
+set completion-query-items -1
+	# Display completions sorted horizontally, not vertically
+set print-completions-horizontally on
+	# Don't use pager when showing completions
+set page-completions off
--- a/tests/shell/conftest.py
+++ b/tests/shell/conftest.py
@@ -1,0 +1,371 @@
+# The following was adapted from the test framework for the bash_completion
+# project: https://github.com/scop/bash-completion/
+
+# TODO:
+# 1. create fixtures for `fish` and `zsh`
+
+import os
+import difflib
+import re
+import shlex
+from typing import Iterable, List, Optional, Union
+
+import pexpect
+import pytest
+
+
+PS1 = "(base) /@"
+MAGIC_MARK = "__MaGiC-maRKz!__"
+
+
+@pytest.fixture(scope="class")
+def bash(request) -> pexpect.spawn:
+
+    logfile = None
+    if os.environ.get("BASHCOMP_TEST_LOGFILE"):
+        logfile = open(os.environ["BASHCOMP_TEST_LOGFILE"], "w")
+
+    testdir = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), os.pardir)
+    )
+    testenvdir = os.path.abspath(
+        os.path.join(testdir, os.pardir, 'devenv')
+    )
+
+    env = os.environ.copy()
+    env.update(
+        dict(
+            TESTDIR=testdir,
+            TESTENVDIR=testenvdir,
+            PS1=PS1,
+            INPUTRC="%s/shell/config/inputrc" % testdir,
+            TERM="dumb",
+            LC_COLLATE="C",  # to match Python's default locale unaware sort
+        )
+    )
+
+    os.chdir(testdir)
+
+    # Start bash
+    bash = pexpect.spawn(
+        "%s --norc" % os.environ.get("BASHCOMP_TEST_BASH", "bash"),
+        maxread=os.environ.get("BASHCOMP_TEST_PEXPECT_MAXREAD", 20000),
+        logfile=logfile,
+        cwd=testdir,
+        env=env,
+        encoding="utf-8",  # TODO? or native or...?
+        # FIXME: Tests shouldn't depend on dimensions, but it's dificult to
+        # expect robustly enough for Bash to wrap lines anywhere (e.g. inside
+        # MAGIC_MARK).  Increase window width to reduce wrapping.
+        dimensions=(24, 160),
+        # TODO? codec_errors="replace",
+    )
+    bash.expect_exact(PS1)
+
+    # Load bashrc which loads bash_completion.sh and runs the conda setup
+    assert_bash_exec(bash, "source '%s/shell/config/bashrc'" % testdir)
+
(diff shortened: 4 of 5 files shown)


msarahan:
@tartansandal thanks for your continuing work on this!  Do you have any ideas on the CI failures happening on circleCI?  

```_________________ ERROR at setup of TestBashCompletion.test_1 __________________
  File "/home/circleci/conda/tests/shell/conftest.py", line 87, in bash
    pytest.fail("Could not load bash completion for conda")
  File "/opt/conda/lib/python3.7/site-packages/_pytest/outcomes.py", line 119, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: Could not load bash completion for conda
_________________ ERROR at setup of TestBashCompletion.test_2 __________________```

**Comment by tartansandal** [comment]
Hi @msarahan :)

> Do you have any ideas on the CI failures happening on circleCI?

I just had a look at the Dockerfile's for the circleCI tests and it looks like we would need to add the  `bash-completion` package to get these test to work.  I can make a PR to add that, but ...

Is it reasonable to assume that the end-user who wants bash completion for `conda` will either already have, or be willing to install, the `bash-completion` package?  

Tab completion is only relevant to users of interactive, so we can ignore the non-interactive `conda` use cases, say in ML pipelines. The `conda init` code gracefully handles any cases where the package is absent or inactive.

In my experience `bash-completion` is a very common (all most de facto) library for building bash completion scripts. I'm struggling to think of any modern commands that do not use it. According to https://repology.org/project/bash-completion/versions there are at least 134 distribution channels, including all the major Linux distros, plus Homebrew and Cygwyn.  The top google search results for bash completion show installation instructions for a variety of environments. Perhaps adding a note to the documentation for version 4.8 is not unreasonable?

Thoughts on this?

**Comment by msarahan** [comment]
> Is it reasonable to assume that the end-user who wants bash completion for conda will either already have, or be willing to install, the bash-completion package?

Yes!  I definitely think so.  Is bash-completion something that we can offer with a conda package, or is it more of a system-package-manager-only thing?  The answer here mostly relies on whether it can exist in a user-only isolated place, rather than being system-wide.

My concern is not at all with needing bash-completion, but only with how conda will behave when it is not present.  What is the fallback behavior?  How do we tell people that they could have a nicer experience if they add this one extra package?

**Comment by tartansandal** [comment]
> Is bash-completion something that we can offer with a conda package, or is it more of a system-package-manager-only thing? The answer here mostly relies on whether it can exist in a user-only isolated place, rather than being system-wide.

Unfortunately there is no simple user install option like `pip`, `npm`, `gem` or `cpan`. Although `brew` must be doing something to allow a non-system-wide install. I don't have access to a Mac at the moment, so figuring that out might be a little tricky. Then again `brew` provides a whole ecosystem for supporting non-system tools, so that might not yield anything useful. On my current Fedora system I have 52 packages that provide `bash-completion` scripts covering 154 additional commands (on top off a base of 653 core commands) -- we would not want a local install to interfere with any of that.

Apologies for the long ramble: I'm thinking a user install might be possible, but would probably be fragile.

> My concern is not at all with needing bash-completion, but only with how conda will behave when it is not present. What is the fallback behavior? How do we tell people that they could have a nicer experience if they add this one extra package?

By default `bash` provides only very basic completion: it will complete the first word on the command line against commands in PATH and subsequent words to filenames in the current directory. Programmable Completion is a feature you can turn on and which allows you to code your own completion functions. The `bash` source code provides some example scripts that you might use copy into, or source from, your `.bashrc` -- at least that was the norm around 2006.  The `bash-completion` package took these examples, made numerous improvements, bundled them up in an extensible way, and provided system-wide installation and integration. Most importantly, it allowed and encouraged other packages to code and take responsibility for the completion functions of their own commands. 

The upshot is, if you don't have `bash-completion` installed, you either have very limited or very out-of-date completion facilities. If you do have `bash-completion` installed you get sophisticated completion for over 650 commands by default, plus specialized completion for many other packages that you have installed. These completion functions are dynamically loaded, so your shell only completes for commands you actually use and does not get unnecessary bloated.

**Comment by tartansandal** [comment]
I did a little more digging and think it might be possible to make a non-intrusive, `conda` installed, version of `bash-completion` that should not interfere with a system wide installation.  I can tweak the shell initialisation so that `conda` installed version only kicks in if there is no system-wide version. Looks like I'm going to have to figure out how to make a conda package now :-)

**Comment by msarahan** [comment]
That sounds great! Let us know if you have any questions or issues.

**Comment by tartansandal** [comment]
The build turned out to be pretty simple. I've uploaded it to https://anaconda.org/tartansandal/bash-completion.  I've skipped running the upstream tests during the build since they require some dependencies not readily available -- the library does not interact with anything other than `bash`, so I'm thinking this is relatively safe.  Hope that's okay.  

I gather that, if that build is acceptable, thats needs to be moved to a more official channel and have the docker test images load that?

I've scratched together an integration, but it depends on adding `$CONDA_ROOT/share` or similar to `XDG_DATA_DIRS` . I can do that in the `conda init bash` hook script, but I'm thinking that that setting is more general and should probably go elsewhere.  Other packages might benefit from that integration.

Moreover, if I assume that environment variable set appropriately, then it makes much more sense to make a separate `conda-bash-completion` package that depends on `bash-completion` and puts the completion code in `$CONDA_ROOT/share/bash-completion/completions/` rather than the legacy loader directory that I've been using. Then we simply make the `conda init bash` hook install those two packages. I think that's a lot more elegant than what we have so far.

Thoughts on this are most welcome :-)

**Comment by tartansandal** [comment]
Okay, full confession. I'm actually relatively new and simplistic user of Anaconda. I only started with it a couple of months ago so I could follow some SciKit tutorials. Now that I've built a conda package and have a better understanding of the ecosystem, if changed my whole approach to integrating bash completion with conda.

I've built an Anaconda package for `bash-completion` with a patch to make relocation into the base (or any other) environment work. This should co-operate with a both system and user installed completion scripts: the set of completion scripts are merged with user scripts preferred over scripts in the conda environment, and conda environment scripts prefered over system installed scripts. This seems the most logical and consistent hierarchy, but please, let me know if you disagree. 

I've also built an Anaconda package for `conda-bash-completion` that has `bash-completion` as a requirement. This simply installs the conda completion script into the environment.  The upstream of that package is a much more appropriate place to develop and test the completion code. I've moved the test code from this PR to that upstream.

The plan is to ask users who want bash completion for conda to simply run:
```
conda install conda-bash-completion
```
This allows for cases where a bash shell is desired, but completion is not, or for cases where bash completion is only desired in certain active environments. It also allows for parallel cooperating installs of various completion scripts. Note that some existing anaconda packages are already installing completion scripts into the environment.

All I need to do now is to tweak the bash init hook to source `$CONDA_PREFIX/etc/profile.d/bash_completion.sh` if it exists. This is POSIX safe, so we can remove all the convoluted special casing for bash.

I've uploaded both of the above packages (for the linux-64) to https://anaconda.org/tartansandal/ for review.  Happy to move upstreams and repos to official repos or channels if you think that's appropriate.

Please let me know if you think this is the right tack to take :-)

**Comment by msarahan** [comment]
@tartansandal I'm very impressed with your work and dedication here.  Your new structure is very clean, and I definitely think that's the right way to do it.

How should I test this?  What behavior do you expect to add, and what behavior might change or break? Right now, I have installed your packages into a docker image with:

`conda install -c tartansandal conda-bash-completion`

My first test was to enter in

`conda <tab>`

This added one further argument - a filename of the only file in the CWD.  That seems unintuitive to me.  I would expect it to complete to the subcommands.  Is this expected?

I get the same behavior if I supply a subcommand:

`conda install <tab>`

**Comment by tartansandal** [comment]
Thanks @msarahan. I think all the ideas are finally coming together. :smile: 

One benefit of this approach is it should make bash completion easily available for older versions of conda.

The only thing missing at the moment is integration with the bash environment. The key is that 
`$CONDA_PREFIX/etc/profile.d/bash_completion.sh` needs to get sourced at an appropriate point. I've made a git repo https://github.com/tartansandal/bash-completion-recipe to track the development of the recipe and to document my thoughts on the integration.  Still a work in progress, but I have some ideas on how to make this "just work".

For the moment, if you
```
source $CONDA_PREFIX/etc/profile.d/bash_completion.sh
```
this should activate the dynamic loader. The next time you try to complete for `conda`, the `conda-bash-completion` code will be loaded and you get completion for commands, options, sub-commands, sub-command options like so:
```
$ conda <tab>
activate     convert      develop      info         list         render       skeleton     verify
build        create       env          init         metapackage  run          uninstall    
clean        deactivate   help         inspect      package      search       update       
config       debug        index        install      remove       server       upgrade      
$ conda

$ conda d<tab>
deactivate  debug       develop
$ conda de

$ conda --<tab>
--help     --version 
$ conda --

$ conda develop --<tab>
 --build_ext    --clean        --help         --name         --no-pth-file  --prefix       --uninstall
$ conda develop --

$ conda env <tab>
create  export  list    remove  update
$ conda env 

$ conda env create --<tab>
--file     --force    --help     --json     --name     --prefix   --quiet    --verbose
$ conda env create --
```
The completions you get depend on the plugins you have installed, with options extracted from help strings.

**Comment by tartansandal** [comment]
I've just uploaded a new version (2.9.5) of the `bash-completion` package which uses an environment activation hook to ensure the library is loaded.  

For users with the default setup, simply installing `conda-bash-completion` to their default environment should be enough to get it working.  This even works for version 4.7 of conda :1st_place_medal: 

For users who chose to not activate the default environment by default, we just need to get them to add
```
source $CONDA_ROOT/etc/conda/activate.d/bash-completion.sh
```
to the end of their `~/.bashrc` file.  I have some other more global ideas that, if accepted, would make this unnecessary.  I'll add some tickets so we can discuss before I start working on any implementation.

There is still more to do:
1. Revert the PR that integrated conda completion directly rather than with packages.
2. Some tweaks to the activation scripts to minimize their impact on activation times.
3. Write more test for `conda-bash-completion`, although I'll be doing that in the upstream project.
4. Modify the docs so more people can start using this :smile:

msarahan:
Great work!  I was going to tell you about that activate.d stuff, but you beat me to it.

> For users who chose to not activate the default environment by default, we just need to get them to add

If people choose not to activate their environment, there's not a whole lot we can (or should) do, in my humble opinion.  Conda's model of environments goes to hell pretty fast once global things are introduced.  Config files are notably confusing as something that isn't separate on a per-env basis.  I'm happy to talk more about your ideas, regardless. Asking people to activate their env is a mostly small ask, though, especially for the simplification we get by saying "if you don't do this, things may not work, and you're on your own."

Right now, conda lacks any notion of add-on packages, like pip's optional requirements. I would like a way to publicize this nice new capability.  I think eventually it should be included by default with miniconda and anaconda, but I'd like to test it in the wild for a while.  I'll have to think on it.  Announcing it on the mailing lists and conda-forge gitter channel is probably a good start.

tartansandal:
Closing this PR since we decided to move the test code to an independent package.

tartansandal:
Might need to hold onto this one for a few more days @msarahan :disappointed: 

It seems that after installing the new packages, I can no longer run `conda build`. I think this has something to do with the environment activation scripts.  Figuring out the issue is proving difficult.

msarahan:
No worries. Point me to your activation script code and I'll take a look.

tartansandal:
Finally tracked it down. Was a bug in the upstream library which will be fixed in version 2.10. In the process of patching so that version 2.9 is safe. :-)

tartansandal:
Newest version bash-completion-2.9-9 on cloud fixes the above issue <phew!> Added a packaging test to check for this issue in future releases. I'm tracking the development of the recipe at  https://github.com/tartansandal/bash-completion-recipe if you want to check that out :smile:

PR #9389, opened 2019-10-30
Use --recursive instead of --archive with rsync for deletions.

afeller08:
`rsync -a` aka `rsync --archive` is an alias for `rsync -rlptgoD`, which in long form are `--recursive --links --perms --time --groups --owner --devices`. When `.empty` is in fact an empty directory created by conda immediately before rsyncing, these two commands are equivalent.

When there is already an empty `.empty` directory before `conda` runs, groups owner and perms are copied over from the existing directory to the target directory that is being emptied, which can cause problems if the `.empty` file was not created by the same user.

If the `.empty` directory is not actually empty, --devices and --links will cause more content to be copied than otherwise would when they are not used, and `--devices` in particular (if running as super user) can cause side effects that would not ordinarily be associated with simply copying files or especially with simply removing them.

### DIFF
--- a/conda/gateways/disk/delete.py
+++ b/conda/gateways/disk/delete.py
@@ -89,7 +89,7 @@ def rmtree(path, *args, **kwargs):
         if rsync and isdir('.empty'):
             try:
                 out = check_output(
-                    [rsync, '-a', '--force', '--delete', join(getcwd(), '.empty') + "/",
+                    [rsync, '--recursive', '--force', '--delete', join(getcwd(), '.empty') + "/",
                      path + "/"],
                     stderr=STDOUT)
             except CalledProcessError:


msarahan:
@afeller08 I apologize for the lack of attention here. We're just swamped. I think this is a good improvement, and I'll look for time in the near future to read more about it and probably merge it. It you have already signed the cla, ignore the bot. We have to manually as you to the list of signers. Our legal department wanted a human in the loop.

afeller08:
Thanks,  @msarahan , I will most likely resubmit this change from a different account in the next few days. My employer has an open source contribution policy that I became aware of after my initial submission.  (I'm waiting for legal approval to sign your CLA in association with that account).

Issue #9178, opened 2018-11-13
Conda tab completion


adrienpacifico:
[Conda doc](https://conda.io/docs/user-guide/configuration/enable-tab-completion.html) state that tab completion is deprecated since Conda 4.4 .

Does this mean that Conda is not supposed to work with tab completion ?

(Fresh install on brand new Mac, tab completion is not working) 

Why is tab completion deprecated ? I tend to expect bash command line programs to always have so sort of tab completion. Can someone explain ?

Best

laggardkernel:
There has been a zsh completion plugin since 2015. 

https://github.com/esc/conda-zsh-completion

imsodin:
@tartansandal I just grabbed your bash completion file - thanks a lot for it!

One issue I have is that when I do not have a conda environment activated yet, on completion I get

```
$ conda acTraceback (most recent call last):
  File "<stdin>", line 3, in <module>
ImportError: No module named conda
```

What I would expect is that it simply doesn't work but doesn't throw an error (muscle memory will still make me hit TAB). Of course an alternative would be for it to determine where conda is installed and still work fine, but I guess that's a lot of work for not much gain.

tartansandal:
Thanks for the heads up @imsodin !
There is a PR https://github.com/conda/conda/pull/9249 that fixes this and some other edges cases. 
You can grab a copy of the patched completion code from https://raw.githubusercontent.com/tartansandal/conda/harden-bash-completion/conda/shell/etc/bash_completion.d/conda

tartansandal:
For those that are interested, you can now install the above completion facility as a conda package: 
```
conda install -c tartansandal conda-bash-completion
```
I've been testing this against conda version 4.7 and it seems pretty solid now.  If you spot any bugs, please raise an issue at https://github.com/tartansandal/conda-bash-completion.

Happy Hacking :smile:

AMR-:
To those who have this issue, consider upgrading and looking at https://github.com/tartansandal/conda-bash-completion

Issue #8191, opened 2019-01-30
feature request: fish tab completion


michaelaye:
The fish implementation (at least on macOS 10.14.3) lacks tab completion of file paths behind `conda install`, which is useful for things like `conda install --file path_to_where_conda_packages.txt_is.txt`
This works nicely in bash and maybe it was working in fish at some point, I don't remember, but in conda 4.5.latest and 4.6.1 it doesn't work;  doing a `conda init` as prescribed for 4.6.1 didn't change anything.

PR #9391, opened 2019-10-30
Remove broken link in tab completion documentation

felker:
I do not know what happened to the issue 415 in the `conda-docs` repo (it might have gotten lost in the migration of documentation to this repository), but the link is broken.

Perhaps a new link to https://github.com/conda/conda/issues/3140 is warranted?

And is the existing Fish autocompletion in https://github.com/conda/conda/blob/master/conda/shell/etc/fish/conf.d/conda.fish missing documentation? Reference: https://github.com/conda/conda/issues/8191

Also, this page will need to be updated if #9158 (rebased version of #9139) is merged for new Bash tab completion in the v4.8.0 release. See https://github.com/conda/conda/issues/9178

### DIFF
--- a/docs/source/user-guide/configuration/enable-tab-completion.rst
+++ b/docs/source/user-guide/configuration/enable-tab-completion.rst
@@ -3,7 +3,7 @@ Enabling tab completion
 =======================
 
 Conda versions up to 4.3 supports tab completion in bash shells via the argcomplete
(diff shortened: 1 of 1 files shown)


rrigdon:
@felker Thanks for your contribution! I'm reviewing the rest of your comments and will try to incorporate those into the documentation as needed.

Issue #8888, opened 2019-07-09
Anaconda Prompt Access Denied: Missing Temp Files


RaufJavid:
Hi, I  have tried to launch Anaconda Prompt but it is giving me Error..Access Denied. Please find the attached ScreenShot as a Proof. I have also try to change TMP and TEMP but could not work.
![Errot](https://user-images.githubusercontent.com/21277031/60873877-83925700-a250-11e9-89c2-d63dd769e544.png)

msarahan:
That's a bug in our space handling.  What version of conda is this?

RaufJavid:
conda version : 4.6.11.PFA
![error2](https://user-images.githubusercontent.com/21277031/60894514-46dc5500-a27c-11e9-9934-c3f0fbe60282.png)

msarahan:
Please update conda and try again

RaufJavid:
Hi,
Respected Sir...I have updated conda using this command...But still, issue:
**conda update -n base -c defaults conda**

RaufJavid:
I installed Anaconda again this morning (Anaconda3-2019.07-Windows-x86_64.exe) and this problem is still present when launching the anaconda prompt. I added the lines:
```
@REM Added by schuelaw, June 9, 2019 after clean install
@SET TMP=C:\CondaTmp
@REM Added by schuelaw, June 9, 2019 after clean install
@SET TEMP=C:\CondaTmp
```
to the file **[Anaconda root]\condabin\_conda_activate.bat** right after the comment:
`@REM This method will not work if %TMP% contains any spaces.
`
and created the folder "**C:\CondaTmp**".

This works perfectly...Issue Resolved.

mingwandroid:
I am re-opening this. We need to use a temp dir without spaces I fear!

mingwandroid:
.. but thank you for the workaround @RaufJavid, I'm sure others will find it helpful.

RaufJavid:
Thanks, @mingwandroid...Yes, the issue is of spaces in the temp dir.

kamosuela:
Yikes, has this been fixed? I ran into this issue trying to update my spyder and matplotlib on admin mode, and it broke everything. Would prefer not having to uninstall/reinstall Anaconda 3 but that might be my workaround for now?

Akkarine:
conda update --prefix C:\Users\DmitryKriv\anaconda3 anaconda:

[...]

PR #9409, opened 2019-11-05
set tmp to shortened path to exclude spaces

forrestwaters:
Fix for https://github.com/conda/conda/issues/8888

### DIFF
--- a/conda/shell/condabin/_conda_activate.bat
+++ b/conda/shell/condabin/_conda_activate.bat
@@ -18,6 +18,7 @@
   FOR %%A IN ("%~dp0.") DO @SET _sysp=%%~dpA
   IF NOT EXIST "!_sysp!\Scripts\conda.exe" @SET "_sysp=!_sysp!..\"
 )
+@FOR %%A in ("%TMP%") do @SET TMP=%%~sA
 @SET _sysp=!_sysp:~0,-1!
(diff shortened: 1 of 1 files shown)


conda/shell/condabin/_conda_activate.bat
```diff
@@ -18,6 +18,7 @@
   FOR %%A IN ("%~dp0.") DO @SET _sysp=%%~dpA
   IF NOT EXIST "!_sysp!\Scripts\conda.exe" @SET "_sysp=!_sysp!..\"
 )
+@FOR %%A in ("%TMP%") do @SET TMP=%%~sA
```
msarahan: Are we sure that this approach will work?  What happens if the original TMP is something like "C:\Users\My User\AppData\tmp" and this shortens it to "C:\Users\MyUser\AppData\tmp" - surely that won't work, right?

forrestwaters: The ~s "Expands path to contain short names only." https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/for

So in my testing with that set, "C:\Users\Test User\AppData\Local\Temp" becomes "C:\Users\TESTUS~1\AppData\Local\Temp"

msarahan: Sounds good.  Thanks!

PR #9410, opened 2019-11-05
Fix some tests

soapy1:


soapy1:
Guess still needs some work to get everything to pass but I think this is enough for this PR

cc @msarahan @jjhelmus for review

PR #9411, opened 2019-11-05
Touch up help for commands that mention CONDA_EXE

brettcannon:
Either a missing period or missing spaces that caused string concatenation where words were being incorrectly conjoined.


msarahan:
Cla confirmed. Thanks!

PR #9416, opened 2019-11-07
WIP: Timeout solve

soapy1:


conda/core/solve.py
```diff
@@ -25,18 +27,34 @@
...
 from ..models.version import VersionOrder
 from ..resolve import Resolve
+from ..common.signals import raise_timeout
+
 
 log = getLogger(__name__)
 
 
+@contextmanager
+def solver_timeout(time):
```
soapy1: unfortunately this approach does not work for windows or python2

msarahan: on the bright side, we don't need to worry about supporting python 2 anymore!

soapy1: could maybe use something like https://dreamix.eu/blog/webothers/timeout-function-in-python-3 instead

travishathaway:

PR #9421, opened 2019-11-08
Revert previously merge "init bash completion" PR

tartansandal:
The previously accepted PR, while it working in many cases, depended on the user having the `bash-completion` library installed on their system.  This is very common on most Linux desktop systems, but less so on MacOS. 

This feature is now provided by the `conda-bash-completion` package, which depends on a specially patched `bash-completion` package.


msarahan:
Thanks @tartansandal !

tartansandal:
Hi @msarahan. It seems the code that this PR reverts has made its way into the 4.8.0 release :scream: 

Since this gets loaded early, the new 'packaged' completion code, with all the bugfixes, never gets called. Be good if this PR could integrated soon. I'm already seeing users with the broken code.

msarahan:
Yep, my bad on not paying attention to that.  A lot of code got merged along with this since the 4.8.0 cut.  We'll get a release cut today, but it will need to go through some "canary" time.

tartansandal:
Thanks for that mike :smile: Really appreciate all your efforts with rolling out new releases, with so many PRs to evaluate and manage.

Issue #8412, opened 2019-03-14
[Windows] Environment has not been activated


hagamuna:
Hi,

I wanted my base python version to always be updated to the latest version on my Windows machine
However in doing so, conda commands will now always return: 
"The system cannot find the path specified"

# Simple reproduction
1. install miniconda
(not Anaconda, this adds a ton of overhead)

Side note: The local user and system %path% variables have more 1048 (or 2048) characters.
(And yes, this path is already reduced by using environment variables)
So conda won't work properly unless we use the 'add to path' checkbox while installing.

2. Restart PC (to ensure path variables work properly)

3. Open a cmd.exe with Administrator

4. Update conda using:
`conda update conda`
Note: This does not seem to update everything, since python version remains 3.7.1, instead of latest 3.7.2

5.  Update all packages, to ensure python and all it's packages are updated in the base environment
`conda update --all`

6. Start python;
`python`

Result:
```
Python 3.7.2 (default, Feb 21 2019, 17:35:59) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32

Warning:
This Python interpreter is in a conda environment, but the environment has
not been activated. Libraries may fail to load.  To activate this environment
please see https://conda.io/activation

Type "help", "copyright", "credits" or "license" for more information.
```

At least the version seems to be correct :)
Anyway you can already see what is wrong, apparently it expects the environment to be activated.

Lets do that:
`activate base`
Result
`The system cannot find the path specified.`

Same goes for deactivation
`conda deactivate`
Result
`The system cannot find the path specified.`

jpgill86:
I'm speculating, but this might be indirectly related to https://github.com/conda/conda/issues/8404#issuecomment-472993777, specifically the first problem described. You could try the hack described [here](https://github.com/conda/conda/issues/8404#issuecomment-472951203) to see if it fixes your problem.

msarahan:
This is going to be pretty hard to debug with the information you've provided.  What does ``where conda`` show on your cmd.exe?

hagamuna:
`where conda`
```
D:\Miniconda3\Library\bin\conda.bat
D:\Miniconda3\Scripts\conda.exe
```

I looked further into it and it seems that
 D:\Miniconda3\Library\bin\conda.bat
internally calls D:\Miniconda3\condabin\conda.bat

I called that second script by hand, which gives the 
`The system cannot find the path specified.`
So it's actually the second script that contains the problem.

[...]

PR #9425, opened 2019-11-11
Added two topics to troubleshooting, clarified activation doc

rrigdon:


rrigdon:
This is related to this issue: https://github.com/conda/conda/issues/8412

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -337,13 +337,35 @@ the environment, and running any activation scripts that the environment may
 contain. These activation scripts are how packages can set arbitrary
 environment variables that may be necessary for their operation.
 
+When `installing Anaconda <http://docs.continuum.io/anaconda/install.html>`_,
+you have the option to “Add Anaconda
+to my PATH environment variable.” This is not recommended because the
+add to PATH option appends Anaconda to path. This means Anaconda’s features
```
mingwandroid: Only kind of correct on Windows, but misses some important details (sorry I should've mentioned them already). On Windows, PATH is composed from two parts, the system PATH and the user PATH. The system PATH always comes first. When you install Anaconda for Just Me, we add it to the user PATH, when you install it for All Users, we add it to the system PATH. In the former case, you can end up with system PATH values taking precedence over our entries. In the latter case you do not (but we don't recommend All User installs).

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -337,13 +337,35 @@ the environment, and running any activation scripts that the environment may
...
+are found last and the installer appending to PATH does not call the
```
mingwandroid: It's not that anaconda's features are found last. To be exact, PATH lookup proceeds as follows:

When the user enters a non-relative and non-fullpath command (e.g. `conda` as opposed to `/opt/conda/bin/conda` and as opposed to `../conda/bin/conda`, then the operating system searches for `conda` on `PATH`. This is literally the dumbest lookup you can imaging. PATH is split into an ordered list at `:` characters on Unix and at `;` characters on Windows. Then the OS appends the program name (e.g. `conda`) to each entry on this list in turn. If a file is found to exist at that location then it is executed and the procedure finishes. If it isn't found, then we move onto the next entry doing the same append-then-check procedure. You are right about activate not happening though and that's more important IMHO. Either we respect activate and promote nothing but activate or we get rid of it entirely. In the middle ground lies nothing but frustration and bug reports.

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -337,18 +337,50 @@ the environment, and running any activation scripts that the environment may
...
+
+.. code-block:: Python
+
+   Warning:
+   This Python interpreter is in a conda environment, but the environment has
+   not been activated. Libraries may fail to load. To activate this environment
+   please see https://conda.io/activation.
+
+If you receive this warning, you need to activate your environment. To do
+so on Windows, run: ``c:\Anaconda3\scripts\activate base`` in
```
mingwandroid: Minor nit, Scripts should have a capital S (Windows has learnt case sensitivity and we should be using it and promoting it I think).

docs/source/user-guide/troubleshooting.rst
```diff
@@ -974,3 +974,88 @@ Solution
...
+   please see https://conda.io/activation
+
+Solution
+--------
+
+If you receive this warning, you need to activate your environment.
+To do so on Windows, use the Anaconda Prompt shortcut in your Windows
+start menu. If you have an existing cmd.exe session that you’d like to
+activate conda in, run:
+``call <your anaconda/miniconda install location>\scripts\activate base``.
```
mingwandroid: Scripts again

docs/source/user-guide/troubleshooting.rst
```diff
@@ -974,3 +974,88 @@ Solution
...
+PATH may have too many entries from 3rd party software adding itself to
+PATH at install time, despite the user not needing to run the software via PATH lookup.
+
+Solution
+--------
+
+Strip PATH to have fewer entries and activate your environment.
+
+If there's some software that needs to be found on PATH (you run it via
+the CLI), we recommend that you create its own batch files to set PATH
```
mingwandroid: > you create its own 

I think this is better:

> you create your own

docs/source/user-guide/troubleshooting.rst
```diff
@@ -974,3 +974,88 @@ Solution
...
+PATH in the system settings.
+
+For example, a new conda prompt batch file that first strips PATH, then
+calls the correct activation procedure could look like:
+
+.. code-block:: Python
+
+   set
+   PATH=”%SystemRoot%\system32;%SystemRoot%;%SystemRoot%\System32\Wbem;%SYSTEMROOT%\System32\WindowsPowerShell\v1.0\;<3rd-party-entries>”
+   call “<miniconda/anaconda root>\Scripts\activate”
```
mingwandroid: Yay, capital already!

PR #9427, opened 2019-11-12
common.logic refactor/cleanup

mbargull:
First bits of https://github.com/curiu/notes/issues/2.

TODO:
- [x] move everything apart from `Clauses` and `minimal_unsatisfiable_subset` to `_logic.py`
  - `_logic.py` can then be drop-in replaced with another implementation
  - `SatSolver` classes will be replaced by `Enum` values in `logic.py` since they won't be exposed directly anymore, i.e., will be "implementation details" of `_logic.py`.
- [x] adjust tests
- [x] remove test code
- [x] final adjustments


mbargull:
This is ready for review.
Test failures all seem unrelated.

When reviewing, I strongly advice to do a per-commit review because of the "large change set due to code movement/refactor" nature of this PR.

All changes have been made with meticulous care, so I'm >95% confident I didn't screw things up here and believe this is fine to merge.

What does this PR do, functionality-wise? Nothing! It really is just cleanup/refactoring.
Everything should as well as before (in terms of speed, memory, correctness).

Why the fuss then? By introducing the additional `conda.common._logic` module, we separate the interface (variable name handling, etc.) from the actual logic, which eases future work on the latter a fair bit. (ref: https://github.com/curiu/notes/issues/2)

conda/common/logic.py
```diff
@@ -888,12 +888,6 @@ def sum_val(sol, odict):
         return bestsol, bestval
 
 
-def evaluate_eq(eq, sol):
```
mbargull: That function is only used in the tests => this commit moves it to where it's used.

conda/common/logic.py
```diff
@@ -778,6 +774,14 @@ def itersolve(self, constraints=None, m=None):
...
 
     def minimize(self, objective, bestsol=None, trymax=False):
+        if isinstance(objective, dict):
+            objective_list = [(v, self.names.get(k, k)) for k, v in objective.items()]
+        else:
+            objective_list = objective
+
+        return self._minimize(objective_list, bestsol=bestsol, trymax=trymax)
+
+    def _minimize(self, objective, bestsol=None, trymax=False):
```
mbargull: The idea here is to separate all "variable name handling" from the rest of the logic.
https://github.com/conda/conda/pull/9427/commits/f6d191f3f1ee9f7348f6b3f5a079646707b8099e has the bulk of those changes.

conda/common/logic.py
```diff
@@ -781,13 +781,39 @@ def minimize(self, objective, bestsol=None, trymax=False):
 
         return self._minimize(objective_list, bestsol=bestsol, trymax=trymax)
 
+    # TODO: test code, remove this!
+    _problem_write_out_condition = None
+    # 1. obtain problematic instance via:
```
mbargull: Don't try to let `conda` solve that command all the way: You will most likely run out of memory!

I used this script during for local testing:
```python
if __name__ == '__main__':
    from conda.gateways.logging import initialize_logging, set_conda_log_level, TRACE
    initialize_logging()
    set_conda_log_level(TRACE)

    from json import load

    from conda.common.logic import Clauses, CryptoMiniSatSolver
    Clauses._problem_write_out_condition = None

    with open('problem.json') as f:
        problem = load(f)

    m = max(max(map(abs, cl)) for cl in problem['clauses'])
    c = Clauses(m=m, sat_solver=CryptoMiniSatSolver)
    c.add_clauses(problem['clauses'])

    newsol = c.minimize(problem['objective'], problem['bestsol'])
```
to re-run a single `minimize` call that has been written out by this code.

conda/common/logic.py
```diff
@@ -262,6 +263,10 @@ def process_solution(self, sat_solution):
         return solution
 
 
+_BIG_NUMBER = maxsize
```
mbargull: We could also use something else here, e.g., `2**31 - 1`.
It just has to be a number that will never be assigned to a SAT variable.
(Using a big number here instead of just `1` and then starting the actual variables from `2` onward is helpful to avoid any of the awkward `1 == True` moments in Python.)

Ideally, we wouldn't have the type ambiguity here, meaning having a `Literal` class whose instances can **either** be a variable number or one of the special values true/false. But that would create to much overhead with CPython.

Still, it is an improvement to have `TRUE`/`FALSE` values explicit rather than `True`/`False` as the latter can lead to ambiguities and hurts comprehension of the code.

conda/common/logic.py
```diff
@@ -313,7 +318,7 @@ def Assign_(self, vals, name=None):
         x = self._assign_no_name(vals)
         if not name:
             return x
-        if isinstance(x, bool):
+        if vals in {TRUE, FALSE}:
```
mbargull: Note that the `x` -> `vals` change is correct here. That was due to an oversight in https://github.com/mbargull/conda/commit/e1d34c277906ff1512f766ab8806d4040a70c184 that changed `vals` -> `x`.
(Either way, they are functionally identical since `x` is a Boolean if and only if `vals` is a Boolean here.)

conda/resolve.py
```diff
@@ -843,7 +843,7 @@ def push_MatchSpec(self, C, spec):
             libs = [fkey for fkey in tgroup if spec.match(fkey)]
         if len(libs) == len(tgroup):
             if spec.optional:
-                m = True
+                m = TRUE
```
mbargull: ..... and this kids is how you waste more than an hour of time...
(Meaning: It was "real fun" when I realized the changes to `logic.py` were erroneous just because I forgot to check the input here.)

conda/common/logic.py
```diff
@@ -275,20 +275,41 @@ class Clauses(object):
...
     def get_clause_count(self):
-        return self._sat_solver.get_clause_count()
+        return self._clauses.get_clause_count()
 
     def as_list(self):
-        return self._sat_solver.as_list()
+        return self._clauses.as_list()
+
+    def _check_variable(self, variable):
+        if 0 < abs(variable) <= self.m:
```
mbargull: These checks are new. They do not hurt performance-wise since they only happen on the interface level but not in the lower level performance critical code paths.

conda/common/logic.py
```diff
@@ -314,61 +330,175 @@ def from_name(self, name):
...
+            return type(x)(map(self._convert, x))
+        if isinstance(x, int):
+            return self._check_literal(x)
+        name = x
+        try:
+            return self.names[name]
+        except KeyError:
+            raise ValueError("Unregistered SAT variable name: {}".format(name))
+
+    def _eval(self, func, args, polarity, name):
```
mbargull: I dropped the `conv=True` parameter here, meaning now we `_convert` all inputs. That parameter has only been used by `LinearBound` which is temporarily removed by this commit from the `Clauses` interface. `LinearBound` is re-added in https://github.com/conda/conda/pull/9427/commits/4f0f7bc3acf8f321a750eac50006319b23faa265.

conda/common/logic.py
```diff
@@ -314,61 +330,175 @@ def from_name(self, name):
...
+        Returned is the list of those solutions.  When the clauses are
+        unsatisfiable, an empty list is returned.
+
+        """
+        if self.unsat:
+            return None
+        if not self.m:
+            return set() if names else []
+        if additional:
+            additional = (tuple(self.names.get(c, c) for c in cc) for cc in additional)
```
mbargull: Do the name handling here, but the actual preprocessing in `_Clauses.sat`.

tests/test_logic.py
```diff
@@ -279,10 +280,10 @@ def test_minimize():
     C = Clauses(15)
     C.Require(C.ExactlyOne, range(1, 6))
     sol = C.sat()
-    C.unsat = True
+    C._clauses.unsat = True
```
mbargull: I did not expose a setter for the `Clauses.m` property since there should not be any need to set from outside it in actual code. Hence we need to use `._clauses.m` here.

conda/common/logic.py
```diff
@@ -217,7 +217,7 @@ def process_solution(self, sat_solution):
         return sat_solution
 
 
-class CryptoMiniSatSolver(SatSolver):
+class _PyCryptoSatSolver(SatSolver):
```
mbargull: PyCryptoSatSolver is more consistent/makes more sense to use here. (In future `cryptominisat` might be an additional, different choice for `context.sat_solver` that would indicate direct usage of `libcryptominisat`.)

conda/common/logic.py
```diff
@@ -263,6 +263,20 @@ def process_solution(self, sat_solution):
...
+_sat_solver_str_to_cls = {
+    "pycosat": _PycoSatSolver,
+    "pycryptosat": _PyCryptoSatSolver,
+    "pysat": _PySatSolver,
+}
+
+_sat_solver_cls_to_str = {cls: string for string, cls in _sat_solver_str_to_cls.items()}
+
+
+PycoSatSolver = "pycosat"
```
mbargull: The plain `str` here may seem a little strange. This is just to decouple the `common.logic` -> `common._logic` dependency a bit more, meaning the former only needs to import `Clauses, FALSE, TRUE` from the latter and nothing more. That is helpful when we drop-in replace `common._logic` by some other implementation.

conda/resolve.py
```diff
@@ -35,31 +35,39 @@
...
 
 
 @memoize
 def _get_sat_solver_cls(sat_solver_choice=SatSolverChoice.PYCOSAT):
-    cls = _sat_solvers[sat_solver_choice]
+    def try_out_solver(sat_solver):
+        c = Clauses(sat_solver=sat_solver)
+        required = {c.new_var(), c.new_var()}
+        c.Require(c.And, *required)
+        solution = set(c.sat())
```
mbargull: We use `Clauses.sat` here instead of `SatSolver.run` because we don't expose the `SatSolver` classes to `resolve` anymore.

conda/common/logic.py
```diff
@@ -364,12 +364,12 @@ def _convert(self, x):
         except KeyError:
             raise ValueError("Unregistered SAT variable name: {}".format(name))
 
-    def _eval(self, func, args, polarity, name):
+    def _eval(self, func, args, no_literal_args, polarity, name):
```
mbargull: In this commit we re-add `LinearBound`. That method has, in contrast to the other ones, not only arguments that are literals (or lists thereof or lists of clauses which are just tuples of literals) but coefficients (i.e., costs/weights). Those coefficients are also integers but do not represent SAT variables. Hence we should not run `_convert` on those.
Previously, `LinearBound` would just omit `_convert` altogether. Now we separate the arguments into `args` with literals and `no_literal_args` for other arguments like coefficients and (non-literal) Booleans.

conda/common/logic.py
```diff
@@ -433,7 +433,18 @@ def ExactlyOne(self, vals, polarity=None, name=None):
             what = self.ExactlyOne_NSQ
         else:
             what = self.ExactlyOne_BDD
-        return self._eval(what, (vals,), polarity, name)
+        return self._eval(what, (vals,), (), polarity, name)
+
+    def LinearBound(self, equation, lo, hi, preprocess=True, polarity=None, name=None):
+        if not isinstance(equation, dict):
+            # in case of duplicate literal -> coefficient mappings, always take the last one
+            equation = {named_lit: coeff for coeff, named_lit in equation}
```
mbargull: We did not really guard against duplicate keys in the `equation`/`objective` inputs. Now we at least detect duplicate strings or duplicate integers for the literals.
If we input `[(c1, 1), (c2, "v1")]` with `clauses.names["v1"] == 1`, we still do nonsense. Maybe we should fix that (easy enough).

conda/common/logic.py
```diff
@@ -920,8 +943,8 @@ def sum_val(sol, odict):
                 log.trace('Beginning sum minimization')
                 objval = sum_val
 
-            odict = {a: c for c, a in objective}
-            bestval = objval(bestsol, odict)
+            objective_dict = {a: c for c, a in zip(coeffs, lits)}
```
mbargull: Renamed this to `objective_dict` since `odict` is used for `OrderedDict` elsewhere in the code base.

conda/common/logic.py
```diff
@@ -785,20 +807,20 @@ def BDD(self, equation, nterms, lo, hi, polarity):
             ret[call_stack_pop()] = ITE(abs(LA), thi, tlo, polarity, add_new_clauses=True)
         return ret[target]
 
-    def LinearBound(self, equation, lo, hi, preprocess, polarity):
+    def LinearBound(self, lits, coeffs, lo, hi, preprocess, polarity):
```
mbargull: We do not check if `len(lits) == len(coeffs)` since the inputs from `Clauses`' `LinearBound` and `minimize` methods come from `dict.keys` + `dict.items`, hence are equal in length.
We could add checks for the length here if we want, though.

conda/common/logic.py
```diff
@@ -29,262 +29,19 @@
...
-_sat_solver_str_to_cls = {
-    "pycosat": _PycoSatSolver,
-    "pycryptosat": _PyCryptoSatSolver,
-    "pysat": _PySatSolver,
-}
-
-_sat_solver_cls_to_str = {cls: string for string, cls in _sat_solver_str_to_cls.items()}
-
+TRUE = TRUE
+FALSE = FALSE
```
mbargull: Those are re-exported for `resolve` and tests.

conda/common/_logic.py
```diff
@@ -749,6 +714,8 @@ def sum_val(sol, objective_dict):
                     lo = mid + 1
                     log.trace("Bisection failure, new range=(%d,%d)" % (lo, hi))
                     if lo > hi:
+                        # FIXME: This is not supposed to happen!
+                        # TODO: Investigate and fix the cause.
```
mbargull: xref: https://github.com/conda/conda/pull/8672/files#r344641428

conda/common/logic.py
```diff
@@ -694,9 +697,9 @@ def LinearBound_(self, equation, lo, hi, preprocess, polarity):
             lo = max([lo, 0])
             hi = min([hi, total])
         if lo > hi:
-            return False
+            return FALSE
         if nterms == 0:
-            res = lo == 0
+            res = TRUE if lo == 0 else FALSE
```
mbargull: This one was a blast too. Took me quite some time to realize I missed that `bool` here..

conda/common/logic.py
msarahan: I can see issues like you describe in a later comment, but I agree with your reasoning here.

conda/common/logic.py
```diff
@@ -621,12 +622,10 @@ def ExactlyOne(self, vals, polarity=None, name=None):
         return self.Eval_(what, (vals,), polarity, name)
 
     def LB_Preprocess_(self, equation):
-        if type(equation) is dict:
```
msarahan: intentional to kill the dict equation possibility?

tests/test_logic.py
```diff
@@ -89,9 +89,15 @@ def my_SOL(ij, sol):
...
+    if type(eq) is not dict:
+        eq = {c: v for v, c in eq if type(c) is not bool}
+    return sum(eq.get(s, 0) for s in sol if type(s) is not bool)
+
+
 def my_EVAL(eq, sol):
-    # evaluate_eq doesn't handle True/False entries
-    return evaluate_eq(eq, sol) + sum(c for c, a in eq if a is True)
+    # _evaluate_eq doesn't handle True/False entries
+    return _evaluate_eq(eq, sol) + sum(c for c, a in eq if a is True)
```
msarahan: Should this be an `is True` => `== TRUE`?

conda/common/logic.py
msarahan: Would an enum buy any clarity here?  I'm not sure, but it seems like it might fit here.

conda/common/logic.py
msarahan: We'll have to be very careful to make sure that people don't import anything else.  Is it worthwhile to mark everything else in that file as _, or is that just dumb overkill?

msarahan: I think it's probably adequate to just have it over in _logic.py.

msarahan:
Looks good.  I had a few tiny questions, but I'm approving.  Test fails look tied to conda run somehow - not related.

Thanks for your meticulous attention to detail here.  This is definitely hard work, and I think it will pay off well when that code is replaced with something faster.

conda/common/logic.py
mbargull: I didn't use an enum due to performance concerns. However thinking about it now, we could do that here: Since we pipe everything that might be a literal through `Clauses._convert`, we could convert a user-facing `common.logic.TRUE` enum to a implementation-detail `common._logic.TRUE`.

conda/common/logic.py
mbargull: I'm not understanding completely. Can you clarify?
Do you mean marking the other members (apart from `Clauses, FALSE, TRUE`) of `_logic` as private via `_`?

conda/common/logic.py
mbargull: Yes, intentional. `LB_Preprocess` is moved to `_logic.Clauses` in a latter commit and that class doesn't use a `dict` anymore. The user-facing part (`logic.Clauses.minimize` and `logic.Clauses.LinearBound`) still allow the `dict` input and convert it accordingly.

tests/test_logic.py
mbargull: Indeed! A latter commit took care of it ;).

mbargull:
Thanks a bunch for the review! I didn't expect someone to have the nerve to look at it that quickly, TBH. Appreciated!

conda/common/logic.py
mbargull: We'd have to do the inverse conversion, i.e., `common._logic.TRUE` -> `common.logic.TRUE`, for the return values, too, of course.
I think I'll add it as a TODO for now, to revisit it later.

conda/common/logic.py
mbargull: Addressed in https://github.com/conda/conda/pull/9427/commits/2aae4ed0822770a9d82e950026e6bf3076b17415.

conda/common/logic.py
mbargull: Added TODO in https://github.com/conda/conda/pull/9427/commits/ff180eeb0d9b42cac5240c9eb824dc7b01de98a9#diff-13e9ff0832c5112fa7e3819fd7872088R38.

PR #9442, opened 2019-11-15
update miniconda osx docs for activation steps

forrestwaters:


mingwandroid:
In this one, we tell people how to avoid writing to their init scripts:
> If you enter “no”, then conda will not modify your shell scripts at all. In order to initialize after the installation process is done, first run source <path to conda>/bin/activate and then run conda init.

The final part of this (running conda init) will change their init scripts and this could be confusing. Unfortunately the alternative is to recommend that for people who *really* cannot modify their init scripts would be to invoke:

```
eval "$(/opt/conda/bin/python -m conda shell.bash hook)"
conda activate base
```

@msarahan, providing a `conda temp-init` might be useful for these people, thinking about it?

mingwandroid:
(suggesting that users `eval` generated scripts is not good security practice)

Issue #8386, opened 2019-03-07
conda run swallows stderr output


IRDonch:
## Current Behavior

If the command executed with `conda run` prints something to stderr, that output is not shown. If it prints a lot, then it freezes.

### Steps to Reproduce

First, create `test.py` with the following contents:
```python
import sys
print('x' * int(sys.argv[1]), file=sys.stderr)
```

Then:
```
>conda run -n base python test.py 1
[nothing is printed]
>conda run -n base python test.py 10000
[nothing is printed _and_ the command freezes]
```

## Expected Behavior

All stderr output from the command should be passed unaltered to `conda`'s own stderr.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3
            shell level : 1
       user config file : C:\Users\[redacted]\.condarc
 populated config files :
          conda version : 4.6.7
    conda-build version : not installed
         python version : 3.7.1.final.0
       base environment : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\pkgs
                          C:\Users\[redacted]\.conda\pkgs
                          C:\Users\[redacted]\AppData\Local\conda\conda\pkgs
       envs directories : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\envs
                          C:\Users\[redacted]\.conda\envs
                          C:\Users\[redacted]\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.6.7 requests/2.21.0 CPython/3.7.1 Windows/10 Windows/10.0.16299
          administrator : False
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```

```
</p></details>

[...]

PR #9448, opened 2019-11-18
additional_tests

gilescope:
Additional tests around #8386


gilescope:
The build failures seem unrelated to the added tests.
By marking the tests as integration, when do they get run? I couldn't find test_run_with_lots_of_std_out in the output now that I've marked them as `@pytest.mark.integration`.

gilescope:
I don't think there's a lot I can do about your build system. 

I'm going to close this until the build system is stable.

Issue #9368, opened 2019-10-24
Support for conda metachannel


jeremyjliu:
Tracking issue for fully supporting conda metachannel (i.e. server side repodata filtering). At a high level, this involves:

1) passing package names / match specs in the network request to fetch repodata
2) caching repodata in-memory accounting for the specs passed
3) caching repodata on disk for non-metachanneled channels (and not caching metachanneled repodata)

Inherent in the approach is a way to distinguish between a metachanneled and a normal conda channel. I can think of a few approaches:

- add a configuration that identifies metachanneled channels, i.e.
```
metachanneled_channels:
    - <channel a>
    - <channel b>
```
- expand the configuration of the existing channels to have optional values, i.e.
```
channels:
    - <channel a>
         - is_metachanneled: true
```
- require metachannel to return an indicator designating the repodata is filtered, i.e. in the repodata.json add a key: `filtered: true`

Opening this issue for discussion on these ideas and more.

jeremyjliu:
A thought that I don't think is immediately necessary: in the case of wanting / needing to support offline mode for metachannels (which I assume is the primary reason to cache repodata on disk), we could theoretically keep a merged repodata cached on disk. This would comprise all the previously fetched repodata results merged into a single entry per metachannel. Offline mode should only utilize previously fetched repodata and package results anyway, so I think this would be sufficient to power that workflow.

jeremyjliu:
@msarahan, wondering if you had any thoughts on this? or are there maintainers who have informed opinions I should tag? thanks!

jeremyjliu:
I like the second option as well, most extendible to future needs.

jeremyjliu:
Having taken a brief look at the existing implementation, it appears that nested config parameters would be non-trivial to add. Does any maintainer @conda/conda-core have any tips on how to add this?

jeremyjliu:
I've sketched out an implementation below, open to feedback! The existing implementation is limited to flat configs only, so this adds abstraction layers to enable nested configs in general. For instance, an example from Nehal:
```
{
   "defaults":{  
      "repodata_gpgcheck":1,
      "gpgkey":"https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc",
      "urls":[  
         "https://repo.anaconda.com/pkgs/main",
         "https://repo.anaconda.com/pkgs/r"
      ],
      "enabled": 1,
      "gpgcheck": 1, // whether each package has to be verified too…
[...]

PR #9449, opened 2019-11-18
Support nested configuration

jeremyjliu:
Partially addresses #9368, allowing for parameters to be nested and to remove the "flat" config file structure limitation.

Broadly, this PR splits up the top level Parameter class into a top level ParameterLoader, Parameter, and LoadedParameter classes. This lets us recursively load, merge, and typify the parameters from a nested configuration.

To properly support the `channels` configuration for metachannel, the next step is we will need to create an ObjectParameter that can support fields of different types - this will be a separate PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ miniconda*.exe
 docs/source/_build
 **/.vscode
 conda.tmp/
+*.iml
+*.ipr
+*.iws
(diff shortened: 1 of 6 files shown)


msarahan:
This looks great!  Thanks for your work here.  In order to make sure that the full test suite gets exercised, please rebase on master.  There's some failing tests associated with stdout/stderr capture, but they're further in in the integration test suite, so you'll get further than the issue that you're hitting here.

jeremyjliu:
Branch rebased on top of master, I'm not sure if the ~5 or so integration tests that fail are related to the changes here - appear to be the same 5 that are broken on master.

msarahan:
Yep, all good.  Thanks for your work on this!

PR #9452, opened 2019-11-20
Transferred Knowledge Base content

rrigdon:
How to suppress the warning message about updating conda
Add default packages to each new environment automatically 
Restoring environment to a previous revision
conda update vs conda install
Installing conda packages offline
List packages that depend on specific packages


mingwandroid:
The versions mentioned to you get updated to are older than the latest, still, unless there's a way to template that in .rst files (I doubt it) then there's not much point worrying about such minor details. Happy to merge.

Issue #9220, opened 2019-09-07
conda env create suppresses pip errors on windows


dsm054:
## Current Behavior

We're suppressing pip errors, making some simple bugs seem mysterious when we're crossing the conda/pip boundary.

When specifying a pip dependency section in an environment file on windows, the batch script created in `conda.utils.wrap_subprocess_call` does not propagate errors in the `python -m pip` call back to the conda process.  These batch files look like

```
@ECHO OFF
@SET PYTHONIOENCODING=utf-8
@SET PYTHONUTF8=1
@FOR /F "tokens=2 delims=:." %%A in ('chcp') do for %%B in (%%A) do set "_CONDA_OLD_CHCP=%%B"
@chcp 65001 > NUL
@CALL "C:\Users\username\Miniconda3\condabin\conda.bat" activate "C:\Users\username\Miniconda3\envs\pip_suppress"
@IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%
@C:\Users\username\Miniconda3\envs\pip_suppress\python.exe -m pip install -U -r C:\Users\username\condaenv.cxpbrg79.requirements.txt
@chcp %_CONDA_OLD_CHCP%>NUL
```
and we see that we reraise if _activation_ fails, but not if pip installation fails.

This means that `conda_env.pip_util.pip_subprocess`'s call of `conda.gateways.subprocess.any_subprocess`  gets a zero return code even though stderr is not empty, and the branch which _should_ show the error output does not get executed [from pip_subprocess]:

```
    stdout, stderr, rc = any_subprocess(run_args, prefix, cwd=cwd)
    print("Ran pip subprocess with arguments:")
    print(run_args)
    print("Pip subprocess output:")
    print(stdout)
    if rc != 0:
        print("Pip subprocess error:")
        print(stderr)
        raise CondaEnvException("Pip failed")
```

and we don't see what the problem is.

### Steps to Reproduce

Start with an environment.yml like

```
name: pip_suppress
dependencies:
  - python=3.7
  - pip
  - pip:
    - requests==99.99.99
```

which should obviously fail noisily as there is no satisfying version of requests.

Running

```
conda env create -v -f environment.yml
```

generates a bunch of output, ending with:

```
Ran pip subprocess with arguments:
[...]

PR #9460, opened 2019-11-22
Fix for #9220 - exist with pip errorlevel if failed

okhoma:
Sorry for bumping in, but this is kind of an important issue for us and it would be great to have it fixed as soon as possible.


msarahan:
@cla-bot check

msarahan:
You never need to apologize for a PR!  Thank you.  We'll get a release with this out soon.

PR #9465, opened 2019-11-26
Support Object configuration parameters

jeremyjliu:
Adds support for Object configuration parameters, with named instance fields of configurable Parameter types.

### DIFF
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -21,6 +21,7 @@ try:
     from collections.abc import Mapping
 except ImportError:
     from collections import Mapping
+import copy
 from glob import glob
 from itertools import chain
 from logging import getLogger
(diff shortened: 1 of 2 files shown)


jeremyjliu:
A comment: I saw a lot of parallels with ObjectParameter and the Entity class, but it would be a large undertaking to merge the Entity and Parameter concepts so I opted to put this together for now.

conda/common/configuration.py
```diff
@@ -709,14 +728,6 @@ def __init__(self, name, value, element_type, key_flag, value_flags, validation=
...
             name, value, key_flag, value_flags, validation)
 
-    def __eq__(self, other):
-        if type(other) is type(self):
-            return self.value == other.value
-        return False
-
-    def __hash__(self):
-        return hash(self.value)
-
```
jjhelmus: Are these needed for comparison elsewhere?

jeremyjliu: remove it bc it is the same as the superclass

jeremyjliu:
@jjhelmus any other thoughts?

jjhelmus:
LGTM

Issue #8628, opened 2019-05-05
conda activate fails on windows 64 bit


tcaduser:
## Current Behavior

I am trying to run multiple tests in parallel in ctest.  Each test runs a batch file to set the environment, so that the right version of python is run for the test.  When the tests are run in parallel, some of the tests fails due to a race condition in the conda tmp files.  If I run the tests in serial, all of my tests pass.

### Steps to Reproduce

Have a .bat script that calls conda activate before invoking the python interpreter:
```
call conda activate python27
call python %*
```

Invoke this script in multiple processes.  For example, I am using "ctest -j4"

All of these error appear in the output of my test when the race condition occurs.
```
The process cannot access the file because it is being used by another process\

The system cannot find the file C:\Users\foo\AppData\Local\Temp\conda-27531.tmp.

C:\Users\foo\AppData\Local\Temp\conda-27531.tmp

The process cannot access the file because it is being used by another process.
```

## Expected Behavior

The tests should run successfully, even when run in parallel invocations of the batch script.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : C:\Users\jsanchez\Miniconda3
            shell level : 1
       user config file : C:\Users\jsanchez\.condarc
 populated config files : 
          conda version : 4.6.14
    conda-build version : not installed
         python version : 3.7.3.final.0
       base environment : C:\Users\foo\Miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\jsanchez\Miniconda3\pkgs
                          C:\Users\jsanchez\.conda\pkgs
                          C:\Users\jsanchez\AppData\Local\conda\conda\pkgs
       envs directories : C:\Users\jsanchez\Miniconda3\envs
                          C:\Users\jsanchez\.conda\envs
                          C:\Users\jsanchez\AppData\Local\conda\conda\envs
               platform : win-64
[...]

PR #9477, opened 2019-12-03
Fix running conda activate in multiple processes #8628

rulerofthehuns:
bugfix for  #8628

The cause of the problem:
_conda_activate.bat uses %RANDOM% env var to create a unique temp file.
That isn't very random, details here: https://devblogs.microsoft.com/oldnewthing/20100617-00/?p=13673
When starting multiple processes at the same time, they will try to create and use the same temp file.

The fix:
Create a random directory instead of a file, and check if the creation failed instead of checking the existence of the directory. Only one of the started processes will succeed, the rest will retry.


msarahan:
Thanks, I confirm that you have signed the CLA and we'll review this pr
soon.

PR #9536, opened 2019-12-19
Added how to set env variables with config API

rrigdon:


docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -603,6 +604,41 @@ We recommend that you:
...
+=============================
+
+If you want to associate environment variables with an environment,
+you can use the config API. This is recommended as an alternative to
+using activate and deactivate scripts since those are an execution of
+arbitrary code that may not be safe.
+
+First, create your environment and activate it::
+
+  Conda create -n test-env
```
soapy1: I think Conda should be lower cased

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -603,6 +604,41 @@ We recommend that you:
...
+
+To check if the environment variable has been set, run
+``echo my_var`` or ``conda env config vars list``.
+
+When you deactivate your environment, you can use those same commands to see that
+the environment variable goes away.
+
+To unset the environment variable, run ``conda env config vars unset my_var -n test-env``.
+
+When you deactivate your environment, you can see that environment variable goes away by rerunning
```
soapy1: This is duplicated from above

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -603,6 +604,41 @@ We recommend that you:
...
+Once you have set an environment variable, you have to reactivate your environment:
+``conda activate test-env``
+
+To check if the environment variable has been set, run
+``echo my_var`` or ``conda env config vars list``.
+
+When you deactivate your environment, you can use those same commands to see that
+the environment variable goes away.
+
+To unset the environment variable, run ``conda env config vars unset my_var -n test-env``.
```
soapy1: This might be a nice place to talk about specifying an environment using the `-n/--name` or `-p/--prefix` flags. Some thing like "You can specify the environment you want to effect using the -n and -p flags". Then the bit that uses that information to unset some vars "``conda env config vars unset my_var -n test-env``."

rrigdon: I added the line in, but I wanted to make sure I put it in the right place and to check if it needs more info.

soapy1: Looks good. In the cmd line help bit it says

```
                        Name of environment.
  -p PATH, --prefix PATH
                        Full path to environment location (i.e. prefix).
```
I think it would be nice to note that `-n` allows users to specify the name of their environment and `-p` allows them to specify the path to an environment. Once that is in, I think this is good to merge!

Issue #9454, opened 2019-11-21
Conda run with nonexistent env crashes


bahaugen:
## Current Behavior

When you use conda run but pass the name of an environment that doesn't exist you get a cryptic error message and are asked if you want to send the error report to the core maintainers. I have tried this on linux and mac and get the same behavior.

### Steps to Reproduce

```
$> conda run -n bad_env python --version
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/exceptions.py", line 1074, in __call__
        return func(*args, **kwargs)
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/cli/conda_argparse.py", line 82, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/cli/main_run.py", line 23, in execute
        args.dev, args.debug_wrapper_scripts, call)
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/utils.py", line 402, in wrap_subprocess_call
        with Utf8NamedTemporaryFile(mode='w', prefix=tmp_prefix, delete=False) as fh:
      File "/Users/bhaugen/miniconda3/lib/python3.7/site-packages/conda/_vendor/auxlib/compat.py", line 80, in Utf8NamedTemporaryFile
        dir=dir, delete=delete)
      File "/Users/bhaugen/miniconda3/lib/python3.7/tempfile.py", line 547, in NamedTemporaryFile
        (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
      File "/Users/bhaugen/miniconda3/lib/python3.7/tempfile.py", line 258, in _mkstemp_inner
        fd = _os.open(file, flags, 0o600)
    FileNotFoundError: [Errno 2] No such file or directory: '/Users/bhaugen/miniconda3/envs/bad_env/.tmpn40ru8f5'

`$ /Users/bhaugen/miniconda3/bin/conda run -n bad_env python --version`

  environment variables:
                 CIO_TEST=<not set>
        CONDA_DEFAULT_ENV=base
[...]

PR #9542, opened 2019-12-23
Check if valid conda environment for conda run

soapy1:
addresses https://github.com/conda/conda/issues/9454

### DIFF
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from logging import getLogger
-from os.path import basename, dirname
+from os.path import basename, dirname, isdir, isfile, join
 import re
 import sys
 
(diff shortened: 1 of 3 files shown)


conda/cli/common.py
```diff
@@ -211,3 +211,10 @@ def check_non_admin():
             The create, install, update, and remove operations have been disabled
             on your system for non-privileged users.
         """))
+
+def is_valid_prefix(prefix):
+    if isdir(prefix):
+        if not isfile(join(prefix, 'conda-meta', 'history')):
```
bkreider: Should this be documented somewhere that conda requires this file at all times?

rrigdon: Forrest and I discussed it and decided this isn't something that the majority of our users would need to know except in some edge cases so to avoid confusing folks, we're not going to add it to docs. Thanks for thinking of the doc impact, though!

soapy1: ya, the existence of that file is an indication that the prefix is indeed a valid conda environment. It's more of an implementation detail I think

bkreider: I really meant documented internally, so someone doesn't change the name of that file in a later implementation.

PR #9560, opened 2020-01-03
Solver messaging content

soapy1:


soapy1:
Got this to the point where is mostly works for small scope of packages. For packages with many dependencies it's far too slow

soapy1:
faster now. It's output is something like

```
$ conda create -n testenv-unsat2 pytest-html=2.0.1 python=2.7 
Solving environment: \ 
Found conflicts! Looking for incompatible packages.
Finding conflicting packages                                                                                                                             
Examining conflicts for pytest-html=2.0.1                                                                                                              / Examining conflicts for python=2.7
classifying bad deps                                                                                                                                   failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:



Package python conflicts for:
pytest-html=2.0.1 -> pytest[version='>=5.0'] -> python[version='>=2.7,<2.8.0a0|>=3.6,<3.7.0a0|>=3.8,<3.9.0a0|>=3.7,<3.8.0a0|>=3.5,<3.6.0a0']
pytest-html=2.0.1 -> pytest[version='>=5.0'] -> attrs[version='>=17.4.0'] -> python[version='>=3.4']
pytest-html=2.0.1 -> python[version='>=3.6']
python=2.7

```

msarahan:
much nicer!  Thanks!

msarahan:
I think a good next step after this PR is merged is to test the constraints present on the right hand side.  If they are satisfiable with all the other constraints, just hide that row.  For instance:

```
pytest-html=2.0.1 -> pytest[version='>=5.0'] -> python[version='>=2.7,<2.8.0a0|>=3.6,<3.7.0a0|>=3.8,<3.9.0a0|>=3.7,<3.8.0a0|>=3.5,<3.6.0a0']
```

should be hidden, because python is a spec, and its bounds are within the bounds specified in that line.

```
pytest-html=2.0.1 -> pytest[version='>=5.0'] -> attrs[version='>=17.4.0'] -> python[version='>=3.4']
pytest-html=2.0.1 -> python[version='>=3.6']
```

should still show up, because their python specs are not compatible with the user provided python spec.

This kind of thing will be impossible in the general case, where you have complicated interrelationships between sets of dependencies, but I think it's still worth trying when we have time.

soapy1:
created an issue for that here https://github.com/conda/conda/issues/9570

There are 2 more things I would like to do with this PR:
* add back tqdm for user feedback on progress
* more testing
might also be nice to do some performance testing

soapy1:
@msarahan I think this is ready for review

conda/base/context.py
```diff
@@ -314,6 +314,7 @@ class Context(Configuration):
     target_prefix_override = ParameterLoader(PrimitiveParameter(''))
 
     unsatisfiable_hints = ParameterLoader(PrimitiveParameter(True))
+    unsatisfiable_hints_check_depth = ParameterLoader(PrimitiveParameter(3))
```
soapy1: maybe this is more useful if set to 1? That way it will only show the first level of incompatibilities

msarahan: Makes sense to me.  I think that's worth doing, and we'll just have to learn from experience on whether that's what people need.

soapy1: 2 is lowest without causing any test breakages. Could change the tests to set this config, or leave the default at 2?

PR #9565, opened 2020-01-07
Added precedence and conflict info

rrigdon:


docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -141,6 +141,26 @@ Conda looks in the following locations for a ``.condarc`` file:
...
+* Dictionaries - merge
+* Primitive - clobber
+
+Precedence
+----------
+
+The precedence by which the conda configuration is built out is shown below.
+Each new arrow takes precedence over the ones before it. For example, config
+files (by parse order) will be superceded by any of the other configuration
+options and environment variables will always take precedence over the other 3.
```
soapy1: I think it would be nice to mention how conda recognizes the config environment variables. They are of the form `CONDA_<name of config parameter>`

PR #9569, opened 2020-01-08
error out faster when unsatisfiable_hints message set to False

msarahan:
This is primarily used for Navigator, to error out faster when a user tries to install rstudio in an env with an incompatible Qt.  The unsatisfiable_hints setting is a condarc setting, and can also be set with CONDA_UNSATISFIABLE_HINTS env var.  It defaults to True.  Setting it to False will enable this fast-error behavior.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -535,7 +535,7 @@ class Resolve(object):
         return pool
 
     @time_recorder(module_name=__name__)
-    def get_reduced_index(self, explicit_specs, sort_by_exactness=True):
+    def get_reduced_index(self, explicit_specs, sort_by_exactness=True, exit_on_conflict=False):
         # TODO: fix this import; this is bad
         from .core.subdir_data import make_feature_record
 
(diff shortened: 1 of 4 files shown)


conda/resolve.py
```diff
@@ -647,16 +647,19 @@ def filter_group(_specs):
...
-                top_level_spec = s
-                reduced = filter_group([s])
-                if reduced:
+                if filter_group([s]):
                     slist.append(s)
+                else:
+                    pruned_to_zero.add(s)
+
+        if pruned_to_zero and exit_on_conflict:
+            return {}
```
jeremyjliu: @msarahan this is a good improvement! a couple questions:
a) what do you think about throwing an exception here, so that you could include the contents of `pruned_to_zero` in the error message? This gives users of unsatisfiable_hints = False a tidbit of information of which packages would be unsatisfiable.
b) why is this not the default behavior regardless of `exit_on_conflict`? it seems that if we've pruned an explicit_spec to 0, it's not likely to show up in the reduced index and the sat solver should fail, right?

PR #9581, opened 2020-01-09
Updated user guide, deleted Overview, minor clean up

rrigdon:
Made user guide more directive and with more info to help readers navigate it.

### DIFF
--- a/docs/source/user-guide/install/index.rst
+++ b/docs/source/user-guide/install/index.rst
@@ -35,10 +35,6 @@ System requirements
 
 * Windows, macOS, or Linux.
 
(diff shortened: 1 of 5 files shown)


docs/source/user-guide/install/index.rst
```diff
@@ -32,9 +33,9 @@ System requirements
 * For Anaconda---Minimum 3 GB disk space to download and install.
 
 
-* Windows, macOS or Linux.
+* Windows, macOS, or Linux.
 
-* Python 2.7, 3.4, 3.5 or 3.6.
```
msarahan: These are only if you install conda from source.  That's really something that only developers working on conda should do.

Issue #9221, opened 2019-09-07
"Collecting package metadata" cannot proceed and never end


owenyoung75:
Hi all, recently updated conda to version 4.7.11, after which, however, 
when I try to install any lib or create a new virtual env, the process stays in the "Collecting package metadata" step, and cannot proceed further at all.

Specifically, when attempt with:
<details open><summary><code>`conda install anylib` or `conda create any_env`</code></summary><p>

Only the following lines appear:

```
/path/to/anaconda/lib/python2.7/site-packages/cryptography/hazmat/primitives/constant_time.py:26: CryptographyDeprecationWarning: Support for your Python version is deprecated. The next version of cryptography will remove support. Please upgrade to a release (2.7.7+) that supports hmac.compare_digest as soon as possible.
  utils.PersistentlyDeprecated2018,
Collecting package metadata (current_repodata.json): \
```
</p></details>
And nothing else proceeds (stuck here at least for hours even for a simple new virtual env).

Don't even know what steps to take for examining the case, so not sure what kind of info I should provide here. Could anyone give some hint?

jancgreen:
I have the exact same problem. Is there a way to just reinstall conda without using conda?

owenyoung75:
> I have the exact same problem. Is there a way to just reinstall conda without using conda?

Still not sure what happens, but eventually I just uninstalled by removing the whole anaconda package, and then reinstalled it. As mentioned in the documentation:
"This may not remove all traces of the configuration files and directories from Anaconda and its programs, but it's fine for me now." See this ref: [https://docs.anaconda.com/anaconda/install/uninstall/](url)

stijn192:
I'm also running into the same problem, I'm running into this with conda 4.7.11 installed on Python 2.7.12.

Furthermore, I've got 2 environments a Python 2.7.12 and a Python 3.5.2, and no matter what I do I only run into the 
`Collecting package metadata (repodata.json): \ |` 
error. 

Has anybody else have been able to solve this?

jancgreen:
It seems like a problem with conda 4.7.11.  I had to go to a saved back up to get the old version of anaconda and conda that I was using.
________________________________

jeffreyrdcs:
I am still having the same problem with conda 4.7.12.  Did this problem ever get fixed?  I cannot downgrade my conda package or change any of my environemnts due to the 
`Collecting package metadata (repodata.json)` issue.

christakahashi:
[...]

PR #9600, opened 2020-01-14
Possible fix for #9221

mmoisse:
Possible fix for #9221

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -98,7 +98,7 @@ class SubdirData(object):
             package_ref_or_match_spec))
 
         # TODO test timing with ProcessPoolExecutor
-        Executor = (DummyExecutor if context.debug or context.repodata_threads == 1
+        Executor = (DummyExecutor if context.debug or context.repodata_threads < 1
                     else partial(ThreadLimitedThreadPoolExecutor,
                                  max_workers=context.repodata_threads))
         with Executor() as executor:


mingwandroid:
Thanks, can you sign the cla?

msarahan:
I confirm that @mmoisse has signed the CLA.

msarahan:
We need to be careful here.  Number of threads can sometimes be None, which equates to the number of CPUs on the system.  I'm not sure, but I think this change effectively disables parallel repodata fetching for anyone who doesn't explicitly set repodata_threads to a value higher than 1.

Conda needs more intelligent behavior here, for sure, but I think the interim workaround is to set config parameters to force single-threaded behavior if you're having trouble.  See https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#configuring-number-of-threads

mmoisse:
The problem for me was that when max_workers='None' it started 0 threads, causing a infinite stall.

msarahan:
I think it's more likely that you have a high core count machine and conda is trying to use too many threads. https://stackoverflow.com/questions/53385479/python-concurrent-futures-threadpoolexecutor-max-workers - note that if you are on a cloud host, the CPU count can often be the underlying number of CPUs on the bare metal machine, not there number of virtual CPUs dedicated to your instance.

I think the right solution here might be to default to some relatively small value, like maybe 10.

mmoisse:
I'm having this issue on my local machine with 8 cores.
If you create a `ThreadLimitedThreadPoolExecutor` without specifying `max_workers` it indeed makes 10 cores.

https://github.com/conda/conda/blob/30df76b7316193afbf5393217f9a1cec0403543b/conda/common/io.py#L532-L535

But if you explicitaly say `max_workers = None` you will get stuck at the following step, never spinning up threads in the next few lines. Although I'm still confused why some people have this issue while other not.

https://github.com/python/cpython/blob/1d3b0aaa54c56282c0a3e7fc396e5b1de8b1974e/Lib/concurrent/futures/thread.py#L192

```
        if num_threads < self._max_workers:
```

A solution could be to set if max_workers is None it is defaulted to 10 ??

https://github.com/conda/conda/blob/30df76b7316193afbf5393217f9a1cec0403543b/conda/common/io.py#L534-L535

mmoisse:
I figured out why it is an issue for some and not for others, it might have to do with the python version (mine is 2.7.11).

In my version of `concurrent/futures/thread.py` it simply sets max_workers to whatever the user supplied. While in the latest version it is being a bit smarter, and defaulting to the number of cores



```
        if max_workers is None:
            # ThreadPoolExecutor is often used to:
            # * CPU bound task which releases GIL
            # * I/O bound task (which releases GIL, of course)
            #
            # We use cpu_count + 4 for both types of tasks.
            # But we limit it to 32 to avoid consuming surprisingly large resource
            # on many core machine.
            max_workers = min(32, (os.cpu_count() or 1) + 4)
        if max_workers <= 0:
            raise ValueError("max_workers must be greater than 0")
```

msarahan:
I strongly recommend you update your python version.  At least to a more recent python 2 build, but ideally to 3.  We dropped official python 2 support at the start of January: https://www.anaconda.com/time-to-upgrade-to-python-3/

I'm going to close this, because I don't think we should add code to support unsupported old versions of python.  Thank you for thinking about it and exploring, though.  This PR serves as a helpful source of information.

fgp:
I just ran into this, and it took me hours to find and fix the problem. My conda base environment is running on python 2 for historical reasons, and I was *completely* unaware that conda dropped support for that until I found this thread just now. After fixing conda to not hang on *every* operation anymore, I just did a `conda update -n base conda`, and still is still no warning about my setup apparently being completely unsupported. WIth completely unsupported meaning that apparently even a major bug won't be fixed.

I do understand the need to discontinue support for python 2, in fact I was surprised conda support that version for that long. But I believe that before conda stops fixing very major bugs such as this one, users still on python2 should be presented with *very* visible warnings every time they invoke conda about their setup being unsupported, for at least a few months, preferably a year.

Leaving conda all of a sudden silently broken on python 2 seems quite hostile to conda users to me.

PR #9615, opened 2020-01-20
Add and utilize a simplified fast unsatisfiable error message

jeremyjliu:
Opening for discussion. This helps the case where unsatisfiable hints may take longer to generate than the actual index reduction or solve itself. Users who then turn off unsatisfiable hints are given no usable error message. This provides a simplified unsatisfiable error to provide at least some information about the underlying unsatisfiability.

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -606,6 +606,27 @@ class PackagesNotFoundError(CondaError):
         )
 
 
+class SimpleUnsatisfiableError(CondaError):
+    """
+    TODO(jeremyliu)
+    """
+
+    def __init__(self, bad_deps):
+
+        msg = '''
+The following required specifications were filtered to 0 in index reduction and thus the requested
+solve is unsatisfiable. If you would like to know which packages conflict ensure that you have
+enabled unsatisfiable hints.
+
+conda config --set unsatisfiable_hints True
+
+Specifications:
+'''
+        for dep in bad_deps:
+            msg += " - %s\n" % dep
+        super(SimpleUnsatisfiableError, self).__init__(msg)
+
+
 class UnsatisfiableError(CondaError):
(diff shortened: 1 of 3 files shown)


jeremyjliu:
Tagging @msarahan kindly, as based on commit history you are the last to touch this section.

remsendos:
......

conda/exceptions.py
```diff
@@ -606,6 +606,27 @@ def __init__(self, packages, channel_urls=()):
...
+    TODO(jeremyliu)
+    """
+
+    def __init__(self, bad_deps):
+
+        msg = '''
+The following required specifications were filtered to 0 in index reduction and thus the requested
+solve is unsatisfiable. If you would like to know which packages conflict ensure that you have
+enabled unsatisfiable hints.
+
```
remsendos: ......

msarahan:
Thanks @jeremyjliu.  If I read this correctly, this is mainly an improvement of the message that is presented when the pruning ends up empty for a package.  Is that right, or is there a functionality change such that the error gets raised sooner?

Also, I think this only takes effect if `unsatisfiable_hints` is set to False.  Is that your intent, or were you hoping for something more broad?  Right now, that setting defaults to True.

jeremyjliu:
There is a functionality change, error gets raised faster.

The intent was for this to only take effect if unsatisfiable_hints is set to False. Though, I would be open to exploring whether this should be surfaced in a broader context?

jeremyjliu:
@msarahan any thoughts on this?

msarahan:
@jeremyjliu I think this is fine.  However, I'm transitioning to a new role, and I'm not able to make decisions for Conda in the interim.  @kalefranz @forrestwaters @jjhelmus are better suited to evaluate this.

Thanks for your patience, and sorry for the delay.

PR #9627, opened 2020-01-23
Strip sysp from os.environ[PATH] (WIP)

mingwandroid:
We also prevent main_pip from excepting in the case of conda run being called
directly (though I am not sure where the other half of the logic around this
feature is!)

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -166,16 +166,21 @@ def main(*args, **kwargs):
     from os import environ, getpid, pathsep, sep
     from psutil import Process
     pp = Process(Process(getpid()).ppid())
+    silent = 'hook' in args
+    silent = True
     if pp and pp.name() == 'conda.exe':
         pp = Process(Process(Process(getpid()).ppid()).ppid())
     if pp:
         cmdline = pp.cmdline()[0]
         if 'conda.bat' in pp.name():
-            print('conda.bat launched me')
+            if not silent:
+                print('conda.bat launched me')
         elif 'charm' in pp.name() or 'code' in pp.name():
-            print('IDE ({}) launched me: {}'.format(pp.name(), cmdline))
+            if not silent:
+                print('IDE ({}) launched me: {}'.format(pp.name(), cmdline))
         else:
-            print('unknowwn ({}) launched me: {}'.format(pp.name(), cmdline))
+            if not silent:
+                print('unknowwn ({}) launched22 me: {}'.format(pp.name(), cmdline))
     if sys.platform == 'win32' and 'CONDA_PREFIX' in environ:
         oep = environ['PATH']
         paths = oep.split(pathsep)
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -149,50 +149,22 @@ def main(*args, **kwargs):
     from ..exceptions import conda_exception_handler
 
     # Here we are detecting (badly) the case when conda.bat has been run but not
-    # to activate or deactivate. conda.bat will have added entries for sys.prefix
+    # to activate or deactivate. condabat will have added entries for sys.prefix
     # to the front of PATH (so that this conda executable can find its DLLs), but
-    # that is problematic for conda run, if executed through conda.bat because
-    # we do not prefix PATH with the new prefixes entries; we swap them in-place
-    # (a-la reactivate) or may even do nothing at all when re-activating the
-    # currently active env.
-    #
-    # I do not want to be messing with PATH here at all though (at least not until
-    # conda has loaded every DLL it could possibly want). We need a new proxy for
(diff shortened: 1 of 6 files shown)


mingwandroid:
PR to start a discussion, hopefully, about how this specific case *should* be handled.

At the crux, we're loose with what conda is sometimes and have to allow for it being Scripts\conda.exe and also condabin\conda.bat.

conda.bat *must* prefix PATH with its entries, so we have a change in behaviour depending on what invoking `conda` will do.

A flaw with this approach is that its changing os.environ['PATH'] but we're not at liberty to do that really. I am only changing it here because I know everything that wants to read PATH will fetch it from the processes env. Those things need to use a proxy for querying this instead so that os.environ['PATH'] remains correct (i.e. contains what it contained at process start-up) [*]

[*] In the case that conda.exe has been run directly without the right entries on PATH, we could thing to actually deliberately prefix `os.environ['PATH']` to correct for this oversight.

cc @msarahan, @kalefranz, @jjhelmus

conda/cli/main.py
```diff
@@ -147,6 +147,41 @@ def main(*args, **kwargs):
...
+    #
+    # I do not want to be messing with PATH here at all though (at least not until
+    # conda has loaded every DLL it could possibly want). We need a new proxy for
+    # os.environ['PATH'] which gets used for any PATH qeuries made by conda and
+    # which strips sysp from the result (which would be emitted to all script files
+    # (that need to set the PATH env var) and all subprocess envs.
+    #
+
+    from os import environ, pathsep, sep
+    if 'CONDA_PREFIX' in environ:
```
kalefranz: Naive question since I'm just jumping back in... Should this be gated for `on_win` only?

mingwandroid: It wouldn't really matter, arguably we could be more uniform and force the sys prefix paths to the front of PATH in the unix scripts too and remove it on all if discovered there. Still, I'd like a better way to know if we've been called from the the wrapper script or now. I'll think about ways. Suggestions welcome!

conda/cli/main.py
```diff
@@ -147,6 +147,41 @@ def main(*args, **kwargs):
...
+        oep = environ['PATH']
+        paths = oep.split(pathsep)
+        # We do not catch the case of CONDA_PREFIX == sys.prefix. That just works.
+        if paths.index(sys.prefix) < paths.index(environ['CONDA_PREFIX']):
```
jjhelmus: Does this need to be protected from cases where sys.prefix and/or CONDA_PREFIX are not in paths?  On unix system these are typically not on PATH.

conda/cli/main.py
```diff
@@ -147,6 +147,41 @@ def main(*args, **kwargs):
...
+    # conda has loaded every DLL it could possibly want). We need a new proxy for
+    # os.environ['PATH'] which gets used for any PATH qeuries made by conda and
+    # which strips sysp from the result (which would be emitted to all script files
+    # (that need to set the PATH env var) and all subprocess envs.
+    #
+
+    from os import environ, pathsep, sep
+    if 'CONDA_PREFIX' in environ:
+        oep = environ['PATH']
+        paths = oep.split(pathsep)
```
jjhelmus: Should this split on a separator rather than whitespace?

mingwandroid: It splits on `os.pathsep`

conda/cli/main.py
```diff
@@ -147,6 +147,58 @@ def main(*args, **kwargs):
...
+    #
+    # I do not want to be messing with PATH here at all though (at least not until
+    # conda has loaded every DLL it could possibly want). We need a new proxy for
+    # os.environ['PATH'] which gets used for any PATH qeuries made by conda and
+    # which strips sysp from the result (which would be emitted to all script files
+    # (that need to set the PATH env var) and all subprocess envs.
+    #
+
+    from os import environ, getpid, pathsep, sep
+    from psutil import Process
```
kalefranz: This is a super heavy import, _right_ at the beginning of every single invocation of conda. We've got to figure out a way to be more frugal here.

kalefranz: Also, psutil isn't currently a hard dependency. I guess the implicit requirement here is that it becomes one. We really need to do better I think.

mingwandroid: Super heavy and it doesn't even work! I am not sure I need it to work either, I think looking at `PATH` and maybe other env vars would be enough to determine whether conda has been run through `condabin\conda.bat` or not. Suggestions are welcome.

Note though that this is still WIP, but it's becoming pressing that I get this solved since I increasingly rely on conda run *during package builds* to test things (I can, to expedite things, comment out these tests though if you want this to remain in the oven for a while yet, it's certainly a bit half-baked at present; I am also interested in your thoughts on monkey-patching os.environ['PATH'] inside conda rather than me modifying the real thing directly as I am doing here).

PR #9695, opened 2020-02-19
base env python not accessible

mfansler:
The previous version of the comment appears to be a hold-over from pre-v4.4, that is it assumes that the **base** env's `bin/` directory remains on `PATH` even after activating the empty env. This should not be the case. That is, users should only expect a Conda-installed Python if they request it to be in the env they are activating.


chenghlee (approved):

mfansler:
Apparently, this has already been resolved.

PR #9699, opened 2020-02-21
Remove fetch repodata reponse logging

jeremyjliu:
Repodata response logging logs all http headers, including authentication headers which should not be logged.

If this logging is diagnostically critical, would also be open to implementing a header whitelist though it would be more work.


jeremyjliu:
@conda-core any input on this one?

PR #9704, opened 2020-02-24
Prefer channel subdir when writing package repodata_record when inconsistent with index.json

### LINKED ISSUE #9696: Spec file install inconsistent behavior relative to solve with noarch packages in the wrong subdir
**Summary**: when a noarch package is erroneously placed into a channel's linux-64 subdirectory, a conda solve can properly install this package but a conda install with spec file cannot. this issue contains a few proposals to solve this issue.

**Reproduction steps**:
- place a noarch package (with index.json:subdir = noarch) into the linux-64 directory of a (local filesystem) channel
- run a `conda create` or `conda install` that installs this package into an env
- run `conda list -n <name of env> > specfile.txt`
- run `conda create -n second-env --file specfile.txt

**Error**:
```
      File "/Volumes/git/conda/conda/exceptions.py", line 1086, in __call__
        return func(*args, **kwargs)
      File "/Volumes/git/conda/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/Volumes/git/conda/conda/cli/conda_argparse.py", line 82, in do_call
      File "/Volumes/git/conda/conda/cli/main_create.py", line 37, in execute
        install(args, parser, 'create')
      File "/Volumes/git/conda/conda/cli/install.py", line 195, in install
        explicit(specs, prefix, verbose=not context.quiet, index_args=index_args)
      File "/Volumes/git/conda/conda/misc.py", line 90, in explicit
        assert not any(spec_pcrec[1] is None for spec_pcrec in specs_pcrecs)
    AssertionError
```

**Root Cause**
It looks like the error arises from conda trying to match the MatchSpec derived from the spec file URL (with subdir=linux-64) with the PackageCacheRecord loaded from disk / saved into <package path>/info/repodata_info.json during the extraction step (with subdir=noarch from the package's index.json).

This means that the user experience is inconsistent, these class of packages can be install with a solve but cannot be installed with a spec file. What could be done to make the experience consistent?

**Proposals**
1) Enforce the invariant at the solve level that a package with subdir 'X' in index.json must come from a channel with subdir 'X'. Thus, solve results and spec file outputs will not contain this class of packages. My initial thought is this could be accomplished with a filter on repodata load that cuts out any packages that don't properly adhere to the invariant.

2) Factor in this edge case when doing the matching between MatchSpec and PackageCacheRecord in the spec file code path. This will likely mean cross referencing the package's index.json and the channel's subdir, deciding to take whichever one matches the input MatchSpec to use as the subdir in the cached repodata_info.json.

jeremyjliu:
Fixes #9696.

When extracting a package, write the repodata_record with the subdir from the channel if it does not match the subdir provided by the package's index.json. This allows for installation with explicit URLs when a package is erroneously placed in the wrong subdir of the channel, which is allowed for a normal solve.


jeremyjliu:
@conda/conda-core any chance of a review of this one?

jeremyjliu:
@conda/conda-core gentle ping on this one (seems like there is more activity recently)

PR #9748, opened 2020-03-11
account for channel in specs

forrestwaters:
This fixes an issue that has been causing a traceback during `breadth_first_search_for_dep_graph()` such as:
```
      File "/Users/fwaters/git/conda/conda/models/match_spec.py", line 766, in union
        return '|'.join(options)
    TypeError: sequence item 0: expected str instance, Channel found
```
Minimal repro:
```
conda install defaults::statistics defaults::python=3.7
```


jjhelmus:
LGTM

Issue #8385, opened 2019-03-07
conda run doesn't forward the exit code


IRDonch:
## Current Behavior

The return code of `conda run` is always 0, no matter what the executed program returns.

### Steps to Reproduce

```
>conda run -n base cmd /c exit 42
ERROR conda.cli.main_run:execute(26): Subprocess for 'conda run' command failed.  Stderr was:
The system cannot find the path specified.


>echo %errorlevel%
0
```
(Note also the strange "The system cannot find the path specified" message from `conda run` - that seems like it shouldn't be there.)

## Expected Behavior

`echo %errorlevel%` should have output "42".

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3
            shell level : 1
       user config file : C:\Users\[redacted]\.condarc
 populated config files :
          conda version : 4.6.7
    conda-build version : not installed
         python version : 3.7.1.final.0
       base environment : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\pkgs
                          C:\Users\[redacted]\.conda\pkgs
                          C:\Users\[redacted]\AppData\Local\conda\conda\pkgs
       envs directories : C:\Users\[redacted]\AppData\Local\Continuum\miniconda3\envs
                          C:\Users\[redacted]\.conda\envs
                          C:\Users\[redacted]\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.6.7 requests/2.21.0 CPython/3.7.1 Windows/10 Windows/10.0.16299
          administrator : False
             netrc file : None
           offline mode : False
```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```

```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

[...]

PR #9665, opened 2020-02-06
set errorlevel when 'conda run' exits (Resolves #9599)

mlline00:
Addresses #9599 by handling the namedtuple returned from conda.gateways.subprocess.subprocess_call.

Also, introduces three new "conda run" unit tests:
- returns an int
- properly returns zero exit codes
- properly returns non-zero exit codes

These three tests fail with conda 4.8.2 and pass with this PR

PR #9765, opened 2020-03-17
Forward return codes when using conda run

aachkar-miovision:
I expect this to fix https://github.com/conda/conda/issues/8385

Should such things not already be covered by a test? If a maintainer can guide me as to where best to add a test case, I'm happy to add one.


aachkar-miovision:
Just noticed https://github.com/conda/conda/pull/9665, which I believe would also fix the issue. Feel free to close this in favour of the more comprehensive PR.

PR #9789, opened 2020-03-26
Set errno on NotWritableError

bfis:
The  `NotWritableError` exception does accept the `errno` argument, but does not pass it to `OSError` (its 2nd parent class). This causes the `errno` attribute to be unset (defaults to `None`).

This causes issues with [`create_package_cache_directory`](https://github.com/conda/conda/blob/master/conda/gateways/disk/create.py#L430) and [`create_envs_directory`](https://github.com/conda/conda/blob/master/conda/gateways/disk/create.py#L446) which expect to catch `OSError` (and use it's `errno`) when using [`touch`](https://github.com/conda/conda/blob/master/conda/gateways/disk/update.py#L91).


bfis:
The CLA was already signed (confirmation mail was received on "Thu, 26 Mar 2020 17:21:32 +0100").

mcg1969:
https://github.com/conda/conda/pull/10040#issuecomment-661384238

jezdez:
@anaconda-issue-bot check

PR #9797, opened 2020-03-30
Enable ppc64 support

beenje:
I built and started to use conda on ppc64 at work: https://gitlab.esss.lu.se/ics-infrastructure/conda4ppc64e6500
It would be nice to have this patch upstream.


kalefranz:
One of the few PRs I'll probably be merging without passing tests. 😜

beenje:
Thanks :-)

PR #9816, opened 2020-04-07
corrected path

Reissner:


docs/source/user-guide/install/rpm-debian.rst
```diff
@@ -11,7 +11,7 @@ To install the RPM on RedHat, CentOS, Fedora distributions, and other RPM-based
    rpm --import https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc
  
    # Add the Anaconda repository
-   cat <<EOF > /etc/yum.repos.d/conda.repo
+   cat <<EOF > /etc/yum/repos.d/conda.repo
```
mcg1969, line 14: The problem is that both are correct—or rather, different operating systems use one or the other: https://serverfault.com/questions/735509/how-to-add-a-repo-to-yum-on-ubuntu
So any change to the documentation would need to reflect both.

kathatherine: Also, I'm sure I messed up that suggestion. It's just for illustration purposes. Don't commit it.

travishathaway:

PR #9835, opened 2020-04-12
resolve #5562 support creating conda environments from web addresses

### LINKED ISSUE #5562: conda env create --url https://raw.githubusercontent.com/.../environment.yml
I'm the maintainer for the datashader package, and I'm trying to streamline the process by which our users can set up an environment containing a bunch of libraries so that they can run our example scripts and notebooks.  These libraries aren't dependencies of the main library, and so they aren't installed with the main `datashader` conda install, but we do want people to be able to get them all easily.  For that reason, we provide an [environment file](https://raw.githubusercontent.com/bokeh/datashader/master/examples/environment.yml), but telling people to use that is complicated -- we either have to tell them to install datashader, then find the environment file wherever it ended up on their platform, or to go to the URL and download it using some platform-specific command or procedure, name it appropriately, then call it from the commandline.  

It would be vastly simpler for us our users if they could simply do the same command on all platforms to get the current environment:

```
conda env create --url https://raw.githubusercontent.com/bokeh/datashader/master/examples/environment.yml
```
 
Would that be feasible?

mrocklin:
Fixes https://github.com/conda/conda/issues/5562

This allows for passing web addresses through to conda env create,
like the following:

    conda env create -f https://domain.com/my-environment.yml

This uses the urllib module to download the file
and then passes things on as normal.

There are a few cases where we need to avoid explicit checks
about if this is a file or if we want to prepend the directory to the
filename.  In these cases we avoid these operations by checking if
the filename starts with "https://" or "http://"

### DIFF
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -13,7 +13,8 @@ from conda.base.context import context
 from conda.cli import common  # TODO: this should never have to import form conda.cli
 from conda.common.serialize import yaml_load_standard
 from conda.core.prefix_data import PrefixData
-from conda.gateways.connection.session import CondaSession
+from conda.gateways.connection.download import download_text
+from conda.gateways.connection.session import CONDA_SESSION_SCHEMES
 from conda.models.enums import PackageType
 from conda.models.match_spec import MatchSpec
 from conda.models.prefix_graph import PrefixGraph
@@ -145,10 +146,9 @@ def from_yaml(yamlstr, **kwargs):
 
 
 def from_file(filename):
-    if any(filename.startswith(prefix) for prefix in ("https://", "http://")):
-        response = CondaSession().get(filename)
-        response.raise_for_status()
-        yamlstr = response.text
+    url_scheme = filename.split("://", 1)[0]
+    if url_scheme in CONDA_SESSION_SCHEMES:
+        yamlstr = download_text(filename)
     elif not os.path.exists(filename):
         raise exceptions.EnvironmentFileNotFound(filename)
     else:
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -8,7 +8,7 @@ from itertools import chain
 import os
 import re
 import json
-import urllib
+from conda._vendor.auxlib._vendor.six.moves import urllib
 
 from conda.base.context import context
 from conda.cli import common  # TODO: this should never have to import form conda.cli
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -8,6 +8,7 @@ from itertools import chain
 import os
 import re
 import json
+import urllib
 
 from conda.base.context import context
 from conda.cli import common  # TODO: this should never have to import form conda.cli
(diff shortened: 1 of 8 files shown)


tests/conda_env/test_env.py
```diff
@@ -77,6 +77,16 @@ def test_with_pip(self):
         assert 'foo' in e.dependencies['pip']
         assert 'baz' in e.dependencies['pip']
 
+    def test_http(self):
+        e = get_simple_environment()
+        f = env.from_file("https://raw.githubusercontent.com/conda/conda/master/tests/conda_env/support/simple.yml")
+        self.assertEqual(e.dependencies, f.dependencies)
+        assert e.dependencies == f.dependencies
```
mrocklin, line 84: Note that this test tries to reach out onto the open internet.  Do you all have suggestions on how best to make this test robust to not having access to the open internet?

jjhelmus:
@cla-bot check

mrocklin:
Hrm, I'm having difficulty associating the test failures to changes in this PR.  Running the failing test on master I'm also seeing it fail locally.

```
tests/test_activate.py::ActivationIntegrationTests::test_activate_deactivate_modify_path_bash
```

mrocklin:
Checking in here.  Is there a good way to motivate feedback on this PR?  

I'd be happy to barter review elsewhere if that would help.

tests/conda_env/test_env.py
angloyna: can we mock the urllib request response with patch?

conda_env/env.py
```diff
@@ -144,11 +145,17 @@ def from_yaml(yamlstr, **kwargs):
 
 
 def from_file(filename):
-    if not os.path.exists(filename):
+    if any(filename.startswith(prefix) for prefix in ("https://", "http://")):
+        try:
+            yamlstr = urllib.request.urlopen(filename).read()
```
kalefranz, line 150: Instead of `urllib`, we should import `from conda.gateways.connection.session import CondaSession`, which is a `requests` `Session`. That hooks into our token and auth management, proxy configuration, etc. So something like

```
response = CondaSession().get(filename)
response.raise_for_status()
yamlstr = response.text
```

tests/conda_env/test_env.py
kalefranz: Reaching out is fine, so long as the test is annotated `@pytest.mark.integration`.  Otherwise, `tests/test_fetch.py` has some examples of `requests` mocking using the `responses` library.

mrocklin: Oh great.  Thanks.  Done.

conda_env/env.py
mrocklin: Ah, that's nice.  Done.

mrocklin:
Tests are failing, but I think in ways that may not be related to this PR.

mrocklin:
> For whatever reason, I've never been able to push directly to contributor branches even though I know "allow edits by maintainers" has been enabled by default for a long time. I also can't seem to find your fork in my list of forks to propose a PR against. :confounded:

:confounded: indeed, that's a tough handicap as a maintainer

mrocklin:
To be clear, this doesn't work?

```
git remote add mrocklin git@github.com:mrocklin/conda.git
git remote update mrocklin
git checkout support-web-environments
git reset --hard mrocklin/support-web-environments
git commit --allow-empty -m "test commit"
git push mrocklin support-web-environments
```

mrocklin:
I'm not sure that I grok this failure

```
________ IntegrationTests.test_conda_pip_interop_conda_editable_package ________
  File "/home/circleci/conda/tests/test_create.py", line 2037, in test_conda_pip_interop_conda_editable_package
    assert len(unlink_dists) == 1
AssertionError: assert 3 == 1
  -3
  +1
```

kalefranz:
The circleci test failures are both `test_conda_pip_interop_conda_editable_package` which is fixed in https://github.com/conda/conda/pull/9877 with 1f2d6012d5539166020879cf96d46b24fbd7de5f.

mrocklin:
Grand.  Is there anything else that I should handle on this PR then, or shall I just wait a bit for upstream to resolve itself.

kalefranz:
Nope, I think we're good now! We've been needing something like this for a while. Always had it with `conda create --file`, but for whatever reason this was never wired up for `conda env create`. Thanks @mrocklin

mrocklin:
Woot!  I appreciate your help in pushing this through Kale.  Also, I think
that this is my first code contribution to Conda.  I'm jazzed to be able to
play a small part.

PR #9842, opened 2020-04-14
Get CI Tests Passing

mlline00:
Fix failing CI tests. Changes are scoped to the test code and CI configurations. The conda codebase is not modified.

Utilize a single PR to demonstrate CI tests pass prior to merge.

Disable the following tests (can be restored if deemed necessary/desirable):
- Linux: conda-build 3.10. Conda-build has a dedicated repo, CI testing should be done there.
- Windows: Python 2.7. The install steps fail and 2.7 is EOL. Disabled instead of working through the noise.

Issues of note:
- Unicode Paths on Windows (no surprise, right). Several tests in test_create.py fall on their face when handling unicode paths on Windows - even under Python 3. In all cases, failure is external to conda, with libarchive frequently appearing in the stack traces.  Address failures using the same approach as Python 2.7; specifically, construct prefixes using a restricted set of Unicode characters in the ascii range.  A future iteration can re-enable the Unicode path tests on Windows.
- activate_deactivate_package. This was historically available on c3i_test2 channel. Not anymore. Modify tests to build and consume the package from within conda source tree.

(Slightly) more detailed information is available in the commit descriptions.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 *.sh text eol=lf
 *.psm1 text eol=lf
 dev/start text eol=lf
+*activate.bat text eol=crlf
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,7 +31,7 @@ init:
 
 install:
   - CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"
-  - 'conda install -y -c conda-canary -c defaults -c conda-forge python=%PYTHON_VERSION% pycosat conda requests ruamel_yaml pytest pytest-cov pytest-timeout mock responses urllib3>1.23 pexpect pywin32 anaconda-client conda-package-handling'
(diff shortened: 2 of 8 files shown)


kalefranz:
@mlline00 Thank you _so_ much for this. Very sincerely thank you.

Courtesy of @mcg1969: https://www.youtube.com/watch?v=51uE0IaXL1U

kalefranz:
In theory I agree about conda-build being a downstream project and therefore conda-build tests should belong in the conda-build repo. For historical reasons and historical coupling between the two projects (i.e. conda-build treating the entire conda code base as "public"), we've needed to test conda against conda-build to make sure a minor refactor somewhere doesn't break conda-build entirely. Hopefully we're beyond that at this point with conda-build following better patterns to isolate its dependencies on conda. We may need to bring the conda-build tests back here someday, but now might be a good opportunity to give it a go without them.

kalefranz:
@mingwandroid Any objection to disabling the appveyor py2.7 tests here?

mingwandroid:
I would rather we kept python 2 ticking along where possible, but am fully on-board with not doing that too!

mlline00:
[yw.](https://me.me/i/youre-welcome-none-df5937d1431748c586c9247964e5cb74)

I'll take a look at the AppVeyor python 2 tests when I have some free cycles.

Issue #9814, opened 2020-04-04
url as channel name breaks splitting channel_subdir_str by slash


croth1:
https://github.com/conda/conda/blob/125413caa9384ae5e9e3e2dfcb4616ec33b7aadb/conda/models/match_spec.py#L186

I have a weird issue where `channel_subdir_str` is an url instead of a simple channel name (https://repo.continuum.io/pkgs/main/osx-64). The split then fails with

```
Traceback (most recent call last):
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/cli/conda_argparse.py", line 82, in do_call
        return getattr(module, func_name)(args, parser)
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/cli/main_install.py", line 20, in execute
        install(args, parser, 'install')
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/cli/install.py", line 250, in install
        index)
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/plan.py", line 289, in revert_actions
        target_state = {MatchSpec.from_dist_str(dist_str) for dist_str in h.get_state(revision)}
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/plan.py", line 289, in <setcomp>
        target_state = {MatchSpec.from_dist_str(dist_str) for dist_str in h.get_state(revision)}
      File "/Users/christian/miniconda3/lib/python3.6/site-packages/conda/models/match_spec.py", line 186, in from_dist_str
        channel_str, subdir = channel_subdir_str.split('/', 2)
    ValueError: too many values to unpack (expected 2)
```

kalefranz:
Steps to reproduce the issue? Also the output of `conda info`, `conda config --show-sources`, and `conda list --show-channel-urls`?

croth1:
I do not understand the problem enough to give a reproducible example from the scratch, but I will collect some hints - maybe it helps you or others to figure out what's going on.

I got the problem during trying to revert a conda environment with `conda install --revision 103`


The revision contains a a full url as channel, which *may*  be the origin of the problem. 
```
[...]

Issue #9812, opened 2020-04-04
Startup time dominated by delete_trash


wolfv:
The startup time of conda (and also mamba) is dominated by a call to `delete_trash` which becomes noticeable slow if one has an environment with many files.

delete_trash walks the entire prefix tree to look for files ending with `.conda-trash`. 

Adding the `conda-bld` folder to the blacklisted folders already helps if one has a few builds in the conda folder.

I have added more description here: https://github.com/QuantStack/mamba/issues/103

wolfv:
Another simple improvement would be to use `endswith` instead of fnmatch:

```
def delete_trash(prefix):
    if not prefix:
        prefix = sys.prefix
    exclude = set(['envs', 'conda-bld'])
    for root, dirs, files in os.walk(prefix, topdown=True):
        dirs[:] = [d for d in dirs if d not in exclude]
        for fn in files:
            if fn.endswith('.conda_trash') or fn.endswith(CONDA_TEMP_EXTENSION):
                filename = join(root, fn)
                try:
                    unlink(filename)
                    remove_empty_parent_paths(filename)
                except (OSError, IOError) as e:
                    log.debug("%r errno %d\nCannot unlink %s.", e, e.errno, filename)
```

wolfv:
But, as outlined in the issue in mamba, the real fix would probably be to have a file in `conda-meta` that's called something like `trash.txt` and that contains the full path to each of the `.conda_trash` files.

kalefranz:
`delete_trash()` is being called at

https://github.com/conda/conda/blob/f4d8ae9f4577008deddf4eb8be52e0abbacf475a/conda/cli/install.py#L140

and I don't understand why we'd need to call it _before_ a solve. It should probably only be called before we actually start to mutate an environment, so at

https://github.com/conda/conda/blob/f4d8ae9f4577008deddf4eb8be52e0abbacf475a/conda/cli/install.py#L346

kalefranz:
Actually, we used to delete trash files only at the end of a mutating operation. It looks like we refactored a lot in https://github.com/conda/conda/pull/8134. My guess of why we might want to delete trash files at the beginning of an operation rather than at the end is for dealing with Windows file system hooks (for virus scanners, etc) that can dramatically slow file removal after Conda has made a bunch of changes to the environment.

[...]

PR #9844, opened 2020-04-14
fix url parsing

cjmartian:
This fixes https://github.com/conda/conda/issues/9814

Instead of `split` we use `rsplit` to correctly parse a url. We also change the max split to 1 instead of 2

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -187,7 +187,7 @@ class MatchSpec(object):
                 channel_str, subdir = channel_subdir_str.rsplit('/', 1)
                 if subdir not in context.known_subdirs:
                     channel_str = channel_subdir_str
-                    subdir = context.subdir
+                    subdir = None
                 parts.update({
                     'channel': channel_str,
                     'subdir': subdir,
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -27,6 +27,7 @@ from ..common.io import dashlist
 from ..common.path import expand, url_to_path, strip_pkg_extension, is_package_file
 from ..common.url import is_url, path_to_url, unquote
 from ..exceptions import CondaValueError, InvalidMatchSpec
+from ..base.context import context
 
 log = getLogger(__name__)
 
@@ -184,6 +185,9 @@ class MatchSpec(object):
             channel_subdir_str, dist_str = dist_str.split("::", 1)
             if '/' in channel_subdir_str:
                 channel_str, subdir = channel_subdir_str.rsplit('/', 1)
+                if subdir not in context.known_subdirs:
+                    channel_str = channel_subdir_str
+                    subdir = context.subdir
                 parts.update({
                     'channel': channel_str,
                     'subdir': subdir,
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -188,10 +188,9 @@ class MatchSpec(object):
                 if subdir not in context.known_subdirs:
                     channel_str = channel_subdir_str
                     subdir = None
-                parts.update({
-                    'channel': channel_str,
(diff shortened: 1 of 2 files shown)


conda/models/match_spec.py
```diff
@@ -183,7 +183,7 @@ def from_dist_str(cls, dist_str):
         if '::' in dist_str:
             channel_subdir_str, dist_str = dist_str.split("::", 1)
             if '/' in channel_subdir_str:
-                channel_str, subdir = channel_subdir_str.split('/', 2)
+                channel_str, subdir = channel_subdir_str.rsplit('/', 1)
```
kalefranz, line 186: `subdir` needs to be checked against `context.known_subdirs`. If `subdir` isn't in `known_subdirs`, then `subdir` should be set to `None` and everything goes into `channel_str`.

kalefranz: Also should include both situations in the unit tests

cjmartian: I'll make those changes, thank you!

tests/models/test_match_spec.py
```diff
@@ -964,3 +964,18 @@ def test_md5_merge_wo_name(self):
         assert str(merged[0]) in str_specs
         assert str(merged[1]) in str_specs
         assert str(merged[0]) != str(merged[1])
+
+
+    def test_dist_str(self):
+        m1 = MatchSpec.from_dist_str("anaconda/{}::python-3.6.6-0.tar.bz2".format(context.subdir))
```
kalefranz, line 970: When using `.format()` and positional arguments, best practice is to enumerate each position. So 

```
"anaconda/{0}::python-3.6.6-0.tar.bz2".format(context.subdir)
```

One reason is that it has a substantial speed advantage, approaching the speed of the older `%` string composition. The performance obviously doesn't matter here, but it can make a difference in a tight loop.

conda/models/match_spec.py
```diff
@@ -183,7 +183,7 @@ def from_dist_str(cls, dist_str):
         if '::' in dist_str:
```
kalefranz, line 183: I guess after line 182 we should start adding in 

```
        if dist_str.endswith(CONDA_PACKAGE_EXTENSION_V2):
            dist_str = dist_str[:-len(CONDA_PACKAGE_EXTENSION_V2)]
```

kalefranz: Although another best practice is to use string slices instead of `.endswith()`, again for performance reasons. See comment at https://github.com/conda/conda/issues/9812#issuecomment-610605261.

conda/models/match_spec.py
```diff
@@ -183,7 +184,10 @@ def from_dist_str(cls, dist_str):
...
             channel_subdir_str, dist_str = dist_str.split("::", 1)
             if '/' in channel_subdir_str:
-                channel_str, subdir = channel_subdir_str.split('/', 2)
+                channel_str, subdir = channel_subdir_str.rsplit('/', 1)
+                if subdir not in context.known_subdirs:
+                    channel_str = channel_subdir_str
+                    subdir = None
                 parts.update({
                     'channel': channel_str,
                     'subdir': subdir,
```
kalefranz, line 193: I'm a little nervous here about adding `parts["subdir"] = None`, and I don't think we need to. Instead of 

```python
parts.update({
    'channel': channel_str,
    'subdir': subdir,
})
```

let's just do

```python
parts["channel"] = channel_str
if subdir:
    parts["subdir"] = subdir
```

kalefranz:
Also now you can rebase against master.

conda/models/match_spec.py
cjmartian: Sounds good, I made that change

PR #9859, opened 2020-04-19
enable trailing '_' to anchor openssl-like versions

kalefranz:
Openssl versions should sort e.g. [1.0.1, 1.0.1a, 1.0.1b] etc., however
this is incompatible with most other version sort conventions and therefore
conda sorts as [1.0.1a, 1.0.1b, 1.0.1]. The VersionOrder docstring
anticipates the need to anchor an openssl-like version sequence at
e.g. 1.0.1, and indicates that appending a dash to the version will
accomplish that goal. However dashes have never been allowed in versions
due to dashes being the delimiters for the legacy 'dist' string.

This commit modifies the anchoring functionality to be provided by
appending an underscore rather than a dash. Therefore a sort order
of [1.0.1_, 1.0.1a, 1.0.1b, 1.0.1] holds.

Note that previous versions of conda prior to this commit will blow up
with VersionOrder("1.0.1_") or MatchSpec("openssl=1.0.1_") being an
invalid version. Therefore this convention cannot show up in repodata
without crashing all current versions of conda. This PR also doesn't
break current versions of conda though, because
MatchSpec("openssl=1.0.1-") was never valid.

### DIFF
--- a/conda/models/version.py
+++ b/conda/models/version.py
@@ -127,6 +127,7 @@ class VersionOrder(object):
          < 0.960923
          < 1.0
          < 1.1dev1    # special case 'dev'
+         < 1.1_       # appended underscore is special case for openssl-like versions
          < 1.1a1
          < 1.1.0dev1  # special case 'dev'
         == 1.1.dev1   # 0 is inserted before string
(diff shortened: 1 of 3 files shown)


jjhelmus:
LGTM

PR #9886, opened 2020-04-27
revert breaking changes to configuration API

kalefranz:
The PRs #9449 and #9465 were submitted in the spirit of moving forward #9368. The purpose of the PRs was to enhance conda's configuration engine with the ability to handle nested types in configuration parameters. However, to accomplish this, there was a major change introduced in the configuration API. We can accomplish nested configuration without breaking the existing API. We can do better.

The composition of config parameters hasn't yet been used in practice, so reverting here to go back for a second try shouldn't break existing functionality, and instead only restore the long-established API.

In the spirit of moving metachannels forward, I think that's an area that we can make some great progress on with some additional discussion.


travishathaway:

PR #9888, opened 2020-04-27
use exec for conda run on unix only

kalefranz:


travishathaway:

PR #9889, opened 2020-04-27
replace listdir and glob with scandir

kalefranz:
This PR dramatically reduces the number of syscalls conda is making when loading up state off of the file system. Should be especially impactful on Windows. The tradeoff is that we incur an additional hard dependency (i.e. `scandir`) when python=2.7 is the base python. Given that python 2.7 is deprecated, it seems like a fair trade off.


conda/cli/install.py
```diff
@@ -6,6 +6,10 @@
 from logging import getLogger
 import os
 from os.path import abspath, basename, exists, isdir, isfile, join
+try:
+    from os import scandir
+except ImportError:
+    from scandir import scandir
```
kalefranz, line 13: Before we merge this PR, we should consider moving all of these imports to `conda.common.compat`.

kalefranz:
Appveyor test failures here are due to timeouts. This PR is mergeable.

conda/cli/install.py
mcg1969: Now that CI seems healthier again do you want to go ahead and do this?

PR #9891, opened 2020-04-27
consolidate source of truth for noarch type to info/link.json

kalefranz:
Previously, the source of truth for detecting noarch type was split between two locations: a noarch field in repodata, and the contents of info/link.json. For installation purposes, we now treat the source of truth as only the latter. Thus, noarch_python package types can in theory be located in any subdir.


travishathaway:

PR #9898, opened 2020-04-28
Minor documentation fix

chenghlee:
Changed en-dashes to `--` for options.


kalefranz:
We have CLA-related stuff to clean up. Right now all the bot is using is https://github.com/conda/clabot-config/blob/master/.clabot. Add your name there if you want. We do have a ticket to go back and re-populate that .cla-signers file though, so it's still relevant.

Don't mind the two red ❌ in this case though. Hit the merge button whenever you're ready.

PR #9902, opened 2020-04-29
audit yaml usage for safe_load vs round_trip_load

kalefranz:
also validated `ruamel.yaml 0.16.10`


conda_env/env.py
```diff
@@ -135,7 +135,7 @@ def from_environment(name, prefix, no_builds=False, ignore_channels=False, from_
 
 def from_yaml(yamlstr, **kwargs):
     """Load and return a ``Environment`` from a given ``yaml string``"""
-    data = yaml_load_standard(yamlstr)
+    data = yaml_safe_load(yamlstr)
```
kalefranz, line 138: This is a substantive change. It is especially necessary following #9835.

mcg1969:
This is passing now with the latest CI.

conda/common/configuration.py
```diff
@@ -435,19 +436,19 @@ def valueflags(self, parameter_obj):
 def load_file_configs(search_path):
     # returns an ordered map of filepath and dict of raw parameter objects
 
-    def _file_yaml_loader(fullpath):
+    def _file_loader(fullpath):
```
mcg1969, line 439: While strictly unnecessary this enabled me to do `find . -name '*.py' -exec grep -q yaml_load {} \; -print` and have it turn up empty, which is nice.

Issue #9590, opened 2020-01-10
conda install error: json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 58 (char 57)


valvesss:
## Current Behavior

Collecting package metadata (current_repodata.json): failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/exceptions.py", line 1078, in __call__
        return func(*args, **kwargs)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/cli/conda_argparse.py", line 82, in do_call
        return getattr(module, func_name)(args, parser)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/cli/main_install.py", line 20, in execute
        install(args, parser, 'install')
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/cli/install.py", line 265, in install
        should_retry_solve=(_should_retry_unfrozen or repodata_fn != repodata_fns[-1]),
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/solve.py", line 117, in solve_for_transaction
        should_retry_solve)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/solve.py", line 158, in solve_for_diff
        force_remove, should_retry_solve)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/solve.py", line 262, in solve_final_state
        ssc = self._collect_all_metadata(ssc)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/common/io.py", line 88, in decorated
        return f(*args, **kwds)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/solve.py", line 415, in _collect_all_metadata
        index, r = self._prepare(prepared_specs)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/solve.py", line 1004, in _prepare
        self.subdirs, prepared_specs, self._repodata_fn)
      File "/tmp/miniconda3/lib/python3.7/site-packages/conda/core/index.py", line 228, in get_reduced_index
        repodata_fn=repodata_fn)
[...]

PR #9906, opened 2020-04-30
Add less opaque error message when repos fail to host data properly

ijustlovemath:
When the conda repos are experiencing high load or internal errors, users of conda will see a really opaque "JSONDecodeError", which is difficult to debug. See #9590 for other users who have seen this error.

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -346,8 +346,11 @@ class SubdirData(object):
         return _pickled_state
 
     def _process_raw_repodata_str(self, raw_repodata_str):
-        json_obj = json.loads(raw_repodata_str or '{}')
-
+        try:
+            json_obj = json.loads(raw_repodata_str or '{}')
+        except json.decoder.JSONDecodeError:
+            log.debug("Conda repository may be experiencing issues, please try again later or use another mirror")
+            raise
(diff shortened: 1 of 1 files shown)


conda/core/subdir_data.py
```diff
@@ -346,8 +346,12 @@ def _check_pickled_valid():
...
 
     def _process_raw_repodata_str(self, raw_repodata_str):
-        json_obj = json.loads(raw_repodata_str or '{}')
-
+        try:
+            json_obj = json.loads(raw_repodata_str or '{}')
+        except json.decoder.JSONDecodeError:
+            log.debug("Conda repository may be experiencing issues, "
+                      + "please try again later or use another mirror")
+            raise
```
kalefranz, line 354: We need to handle this decode error, but I don't think `log.debug()` is sufficient. Most users won't even see that. We need to throw the whole downloaded file away and start again.

ijustlovemath: That's fair. I'm not familiar enough with conda's codebase to make that change; this is more of a bandaid to help those installing over CLI to see what's really going on and how to fix it (which is to wait, unless someone actually wants to implement the change you suggested)

jezdez:
Closing this PR since it seems as if this isn't the right strategy to handle download errors.

PR #9919, opened 2020-05-05
Allow retries for 500 error code.

bnemanich:
I needed to enable retries for a 500 error code. When testing this out, it appeared that retries weren't being done for any error codes, so I added the default codes that urllib3 normally retries on. I also found that if the retry limit had been reached, conda would crash with an unhandled exception. This was solved by adding the `raise_on_status` flag.

This was tested using the following web serving script to return errors:
```
from http.server import BaseHTTPRequestHandler

class ErrorHandler(BaseHTTPRequestHandler):
    
    def do_GET(self):
        self.send_error(500)
        return

if __name__ == '__main__':
    from http.server import HTTPServer
    server = HTTPServer(('localhost', 8090), ErrorHandler)
    print("Starting server, use <Ctrl-C> to stop")
    server.serve_forever()
```

Once the web server was running, the following command was used to test the retry capabilities:
`conda install -c "http://localhost:8090" conda`


jjhelmus:
@cla-bot check

bnemanich:
@jjhelmus Do you have any opinions on this PR?

Issue #9791, opened 2020-03-27
anaconda3/miniconda3 installer for s390x


pradghos:
Hello,

We are looking for Anaconda3 or Miniconda3 installer for s390x platform -
However , It appears that s390x architecture is not supported so far ;

None of the installers are available in the below locations -
    Anaconda Installers
    - `https://repo.anaconda.com/archive/`
    Miniconda Installers 
    - `https://repo.anaconda.com/miniconda/`

**Motivation : -**

Red Hat Enterprise Linux 8 is  already supporting s390x (IBM Z );
So it might be a good timing to get conda support to bring existing AI, Data Science, Machine Learning tools on s390x platform !

JFYI - Travis CI is also supporting s390x environment - which makes development on s390x easier. 
`https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z`

If it is not supported today, Please let us know how do we collaborate and contribute back to conda !

Thank you very much !

chandrureddy:
@pradghos This is very interesting initiative on such powerful platform (LinuxONE) where Data resides closely to the platform in the cases where especially Data Virtualization, Data Science, Machine learning, AI and much more. 

Looking forward to see more progress and support from anaconda team !

kalefranz:
Hello, Anaconda partners with Rocket Software (https://www.rocketsoftware.com) for s390x work. Have you contacted anyone from Rocket?

chandrureddy:
@kalefranz As per my understanding IBM has partnered with RocketSoftware for Anaconda on zOS so far and not on zLinux i.e s390x. If Rocketsoftware willing to contribute for s390x platform too then I guess good idea to have them contribute too.

kalefranz:
Ah sorry my understanding probably has gaps here. So s390x means Linux on z/Architecture? Yes, Rocket is working on zOS on z/Architecture.

Definitely would appreciate some education and clarification here for everyone more familiar with the Windows/Linux/macOS and x86 worlds.

chandrureddy:
@kalefranz you are absolutely right that s390x means Linux on z/Architecture 64-bit.

Base information can be found [here](https://www.ibm.com/in-en/it-infrastructure/z/os)

PR #9933, opened 2020-05-11
Initial code drop to enable support for s390x

pradghos:
Addressing : https://github.com/conda/conda/issues/9791
Initial code drop to enable support for `s390x`;


pradghos:
Please let me know in case of any review comments to address further. Thank you very much !

Issue #4298, opened 2017-01-14
replace path placeholder even if it's stored in a wide string


wojdyr:
Prefix placeholder can be stored in binary files as a wide string (wchar_t).
In such a case it's not replaced by conda.
I came across this problem when building wxWidgets that is using internally wide strings: UCS-2 on Windows and UCS-4 on Unix:
```
007c4f10  2f 00 00 00 6f 00 00 00  70 00 00 00 74 00 00 00  |/...o...p...t...|
007c4f20  2f 00 00 00 63 00 00 00  6f 00 00 00 6e 00 00 00  |/...c...o...n...|
007c4f30  64 00 00 00 61 00 00 00  2f 00 00 00 63 00 00 00  |d...a.../...c...|
007c4f40  6f 00 00 00 6e 00 00 00  64 00 00 00 61 00 00 00  |o...n...d...a...|
007c4f50  2d 00 00 00 62 00 00 00  6c 00 00 00 64 00 00 00  |-...b...l...d...|
007c4f60  2f 00 00 00 77 00 00 00  78 00 00 00 77 00 00 00  |/...w...x...w...|
007c4f70  69 00 00 00 64 00 00 00  67 00 00 00 65 00 00 00  |i...d...g...e...|
```
Perhaps conda could try to replace the placeholder in any of the popular encodings (UTF8, UTF16, UTF32)?

kalefranz:
If you'd like to submit a pull request, the place to focus would be conda/core/portability.py.

hmaarrfk:
Did you ever find a solution to this?

I'm currently building wxWidgets as well, and got into the same troubles.

hmaarrfk:
Would a PR adding a loop to:
https://github.com/conda/conda/blob/master/conda/core/portability.py#L64

Be acceptable?

hmaarrfk:
Unfortunately, it isn't as easy as that, one needs to ensure that the string doesn't accidentally become null terminated at the end.

See for example:

![image](https://user-images.githubusercontent.com/90008/82243947-44fc0d00-98f5-11ea-967c-2bea845f2905.png)

hmaarrfk:
Yeah, the binary replacement path is a little trickier because you need to fixup the regex and string expansion.

wojdyr:
@hmaarrfk  did you see PR #4307 linked above? I didn't follow up on that. But in short, replacing string is trivial, the question is how to minimize the overhead.

jezdez:
[...]

PR #4307, opened 2017-01-15
replace path placeholder stored as wide string

wojdyr:
Fix for #4298.
I made a patch for conda/core/portability.py. I don't know if anything else should be changed (conda-build?). I tested it only on one example, so it's not much tested.

Doing binary_replace() in three variants may affect performance - no idea how
much.

---

The path placeholder may be stored in binaries in different encodings.
This change adds support for UTF16 or UTF32 to the already supported
UTF8 encoding. It should cover wchar_t and wxString (from wxWidgets
library).

Apparently the placeholder was not substituted on Windows,
and this has not changed. I only moved the call to
replace_pyzzer_entry_point_shebang() out of binary_replace()
to avoid calling it more than once.
Removed has_pyzzer_entry_point() - it was redundant as the same check
is done inside of replace_pyzzer_entry_point_shebang().

PR #9946, opened 2020-05-19
UTF 16 and UTF 32 fixes

hmaarrfk:
Fixes: https://github.com/conda/conda/issues/4298

Needs tests for string replacement tests

I tested this locally to install hugin, which uses wide chars.


wojdyr:
It was discussed in #4307 - scanning each file three or five times during installation may not be optimal.

wojdyr:
@mingwandroid I don't quite get your comment, although I only vaguely remember how conda works.
If the path is up to 255 bytes in utf8, it obviously can be longer than 255 in utf32.

mingwandroid:
That's not what I'm talking about here. We can replace 255 bytes. What those bytes contain is irrelevant.

hmaarrfk:
@mingwandroid thanks for chiming in.
I think the key is that the 255 character string, can be stored as more than 255bytes.

Finding that PR from 2017 would have saved me lots of frustration. Not sure why I didn't click the link on the issue.

I don't think there is a way around scanning files multiple times unless conda build pre-scans them for us telling us which files are encoded in which way (binary files may have multiple encodings)

I'm not really a regex master, so if you could help, that would be appreciated it, but the operating system will cache the files, so it shouldn't be a big deal to scan them 5 times.

And for conda to be cross platform, you might need all 5 encodings, not just the 3 from your platform.

hmaarrfk:
Ok, using your tip above to combine a few different regex into one. Lets see if this works. Hopefully this strategy can be accepted in.

hmaarrfk:
Sorry, testing this locally is a little annoying, hopefully it is OK if i test on CIs, sorry for the notification noise.

hmaarrfk:
@wojdyr I don't know if this is accetable. I had to change the tests. But seeing as it is only an internal function, I also changed the the one internal call accordingly as well.

Let me know if that is ok.

hmaarrfk:
I don't understand why there is some branchign logic between text files and non text files

image

if somebody could help explain that then maybe we can work together to get rid of the for loop.

@mingwandroid

hmaarrfk:
Kind ping to move this forward somehow.

It would be greatly appreciated to have this included as a core conda component.

I don't particularly understand why `FileMode` needs to be tested against. This seems to be some integration challenge, but I left it as it for this first version.

The large binary files are only scanned once in this implementation.

mingwandroid:
Hi @hmaarrfk, sorry, we've all been especially busy lately, though I've had to take a bit of a break.

Can you add a test, involving a package which contains files containing each of the prefix encoded in each of the, err, encodings as a separate PR to conda-build that's marked as xfail, I am happy to work with you on a solution for this, but I still don't think we're on the right path yet. I still see an un'for'tunate loop here: https://github.com/conda/conda/pull/9946/files#diff-827c3c3f848db24eacfc18b94aead92aR72

Although nothing to be especially proud about, this recipe might be a useful starting point: 
https://github.com/conda/conda-build/tree/master/tests/test-recipes/metadata/_binary_has_prefix_files

Cheers!

mingwandroid:
FWIW, I believe the design is wrong. These replacement locations should be pre-computed in their entirety by conda-build and conda should dumbly replace X with Y as instructed by conda-build via paths.json. Your current implementation will not, for example, work for a file that contains only one instance of the prefix, encoded in some exotic encoding that conda-build didn't detect in the first place, unless I am mistaken?

hmaarrfk:
Thanks for chiming in @mingwandroid 

> I still see an un'for'tunate loop here: https://github.com/conda/conda/pull/9946/files#diff-827c3c3f848db24eacfc18b94aead92aR72

Right. I don't personally understand the decision to have the original if statement myself.

> dumbly replace X with Y as instructed by conda-build via paths.json

This might very well be the right approach. I totally don't have a full picture of conda-build + conda.

I think i should be able to create a pretty simple conda recipe

```python
for encoding in common_encodings:
    with open(f'file_{encoding}.data', 'wb') as f:
          f.write(os.environ['PREFIX'].encode(encoding))
```

should do the trick.

jezdez (commented):
Thanks @travishathaway for updating the PR.

@hmaarrfk Do you happen to remember if [this comment](https://github.com/conda/conda/pull/9946#issuecomment-643687580) was fully handled since the patch in conda-forge was applied?

dholth (approved):
LGTM. The code will look for any of POPULAR_ENCODINGS-encoded placeholders in binaries.

Issue #9918, opened 2020-05-05
AppVeyor Suitability


mlline00:
The Windows test suite often exceeds AppVeyor's 60 minute build limit.  When the time limit expires, the test runner is killed and the run is failed. These false-negatives add friction to the development process and diminish utility of the testing code.

Hoping to start a discussion on this topic, here are a few paths forward (may very well be others that I don't have visibility into):

1. Ask AppVeyor to relax the build limit for conda.  Even if they would relax the limit for the main conda project, community members will be unable to reliably run the test suite.
2. Make the unit and integration tests faster.  This would be great, but is no small undertaking. AppVeyor's 60 minute build limit is quite restrictive and out of line with other platforms. Why commit to a permanent, on-going battle against exogenous test failures?
3. Transition the Windows tests to a different CI platform. This seems to be the most viable path.

I'm willing to contribute to the effort, but would like some guidance from maintainers before investing too much time.

To move the conversation forward with respect to option #3, a few candidate CI platforms immediately come to mind:

- Circle CI: The Linux tests are currently hosted on Circle CI, so that would be a natural choice for the Windows tests too. Unfortunately, the free tier would limit community users to only one or two runs of the full Windows test suite per week; i.e. Circle IC free tier currently offers 2,500 credits per week. Windows builds consume 40 credits/minute x ~40 minutes/run. It is possible there are more cost effective ways to run the Windows test suite on the Circle CI platform, so don't take that as a final answer.
- Azure Pipelines: Offers solid support for Windows platforms. Other major python projects (such as pandas) utilize Azure Pipelines, so it is reasonable to expect it would serve conda too.
- GitHub CI: I have no experience with this offering, but it deserves a look.

cjmartian:
Hey @mlline00 we are on board for Azure Pipelines.
Thank you for creating this issue!

mlline00:
Appreciate the direction. I'll start working through the migration.

mlline00:
@cjmartian I started a pull request for this. I don't know what will be required in the main conda repository to enable the PR checks, but I think earlier visibility trumps a clean PR in this case.

mlline00:
I marked the PR as ready to review.

Tests are enabled for python 2.7 and 3.7 for win-32 and win-64 platforms.  All tests are run on the vs2017-win2016 image, but could fan out to additional images if desired.

There is one nagging test that I can't resolve. See this [note ](https://github.com/conda/conda/pull/9954#issuecomment-643894267)in the PR.

PR #9954, opened 2020-05-22
Enable Windows CI on Azure Pipelines #9918

mlline00:
PR to track progress toward running Windows test suite on Azure Pipelines.  Opening the PR early to give visibility into the progress. I expect a few iterations before merge.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,62 @@
+# Use default triggering logic
+# trigger: ...
+# pr: ...
+
+pool:
+  vmImage: 'vs2017-win2016'
+strategy:
+  matrix:
+    Python27_32:
+      python.version: '2.7'
+      conda.subdir: 'win-32'
+    Python27_64:
+      python.version: '2.7'
+      conda.subdir: 'win-64'
+    Python37_32:
+      python.version: '3.7'
+      conda.subdir: 'win-32'
+    Python37_64:
+      python.version: '3.7'
+      conda.subdir: 'win-64'
+
+steps:
+- script: |
+    CALL $(CONDA)\scripts\activate.bat
+    set CONDA_SUBDIR=$(conda.subdir)
+    CALL conda create -n ci_base -y python=$(python.version) pycosat conda requests ruamel_yaml pytest pytest-cov pytest-timeout mock responses urllib3 pexpect pywin32 anaconda-client conda-package-handling
+    CALL conda activate ci_base
+    CALL conda install -yq pip conda-build=3.17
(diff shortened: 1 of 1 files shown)


mlline00:
There is one flaky py37 win-32 test:

tests/test_activate.py::ShellWrapperIntegrationTests::test_powershell_PATH_management

Root cause is the infamous SSLError("Can't connect to HTTPS URL because the SSL module is not available") when attempting to connect to https://repo.anaconda.org. I'm not sure what to do about it because the recommended fix to activate the conda environment, but this test executes two "conda deactivate" commands. The test does occasionally pass, but I'm not able to reliably reproduce a passing scenario.

Searching historically, this test was originally xfailed on python 3 because OpenSSL was not on the path (commit fb233e3288cf4f0eb4eb2eeed8f27039b1e97491) , but the xfail was removed a few days later without meaningful commentary (commit 11d66a8ba78c38d079011996ff49f8be1c20586f).   Pure speculation, but AppVeyor is only configured to test win-64 python and the test consistently passes on win-64 platforms, thus the xfail would be inappropriate and have to be removed.

To summarize, I don't see a clear path to resolve this test.

mcg1969:
@mlline00 I actually just got AppVeyor under its timeouts by breaking apart the individual tests into several pieces.

At the same time, I've had great results with other repos (conda-pack, anaconda-project, nb_conda_kernels) using GitHub Actions, an evolution of Azure Pipelines. I think it would be great to standardize on it.

mlline00:
@mcg1969 Great to see the test suite consistently passing again! I'm (largely) indifferent to the CI platform. Went down the Azure Pipelines path based on the feedback in #9918.

This PR makes two material changes to the testing:
- Replaces python 3.6 with 3.7
- Covers 32 and 64-bit platforms for each version of python

Additionally, there was a 64-bit python 2.7 bug uncovered and fixed in conda/common/_logic.py (see comment for d1aecbd).

It won't hurt my feelings if this PR is closed without merging.  Otherwise, let me know how you'd like to proceed.

mcg1969:
@mlline00 would you be willing to cherry pick https://github.com/conda/conda/commit/d1aecbdc98537d36f6de9f664606360a1414aa5c into an independent PR? We should definitely merge that immediately.
I feel like I should defer to the rest of the team on the overall CI effort, although I must admit that if we were to make a change, it's my view we should go to GitHub actions if simply in anticipation of better support and integration from Microsoft on this moving fowrard.

mlline00:
@mcg1969 Cherry picked d1aecbd and submitted as PR #10108

PR #9960, opened 2020-05-28
Change apt-get to apt

ChihweiLHBird:
Since apt command-line was introduced in 2014, apt become a better way for user to install packages because users don't have to worry about shifting between apt-get and apt-cache. Therefore, I would suggest we to make this change in the conda project


ChihweiLHBird:
Just signed Conda Contributor License Agreement after making the pull request.

jjhelmus:
@cla-bot check

jjhelmus:
This change is a good idea, thanks for putting this together @ChihweiLHBird

Issue #9976, opened 2020-06-03
conda sometimes failing to install noarch package entry point scripts


mparry:
###  Current Behaviour
I'm finding that sometimes `conda install` will fail to fully install `noarch` packages on Windows. In particular, the generated entry point `...-script.py` files appear to be missing and the entry points therefore cannot be used. There is no error output from the `conda install` when this happens; it apparently believes it has succeeded.

I can eventually reproduce this if I go back and fore between a couple of versions of a `noarch` package that has one or more entry points. I think this is perhaps somehow related to attempting to delete hard links and some eventual failures in handling of that. (Deleting the hard links fails because another copy is in use; e.g. below, `ep1.exe` cannot be deleted because it is a hard link to `conda.exe`, which is currently being run. This is a general limitation in Windows.)

As I repeat the installs, the amount of 'trash' builds up until it silently fails. For example, with two entry points, e.g. `ep1` and `ep2`, I finally ended up with a `Scripts` directory containing the below (but no `ep1-script.py` or `ep2-script.py`):

```
13/03/2020  18:03     74,752 ep1.exe
13/03/2020  18:03     74,752 ep1.exe.conda_trash
13/03/2020  18:03     74,752 ep1.exe.c~
13/03/2020  18:03     74,752 ep1.exe.c~.conda_trash
13/03/2020  18:03     74,752 ep2.exe
13/03/2020  18:03     74,752 ep2.exe.conda_trash
13/03/2020  18:03     74,752 ep2.exe.c~
13/03/2020  18:03     74,752 ep2.exe.c~.conda_trash
```

As I repeated these upgrades and downgrades, running `conda install --debug` at each step, the process seemed to steadily degrade. I have quoted the relevant sections below. You can diff them to see the full changes in each step but basically there are just more and more errors about failing to delete files.

<details><summary>Upgrade no. 1</summary><p>

```
Preparing transaction: ...working... DEBUG conda.core.package_cache_data:_check_writable(259): package cache directory '
C:\Users\m.parry\AppData\Local\conda\conda\pkgs' writable: True
DEBUG conda.core.link:_get_python_version(827): found in current prefix python version 3.6
done
Verifying transaction: ...working... done
DEBUG conda.common.signals:signal_handler(45): registering handler for SIGABRT
DEBUG conda.common.signals:signal_handler(45): registering handler for SIGINT
DEBUG conda.common.signals:signal_handler(45): registering handler for SIGTERM
[...]

PR #9978, opened 2020-06-04
Don't hardlink Windows entrypoint executables for noarch packages

mparry:
This is the simple fix that I suggested for #9976


jjhelmus:
@cla-bot check

mingwandroid:
The lack of being able to delete hardlinks when another instance has of open is common across all windows file io.

Personally I would prefer hardlinks to optional on windows in condarc..

mparry:
> The lack of being able to delete hardlinks when another instance has of open is common across all windows file io.

Sure, but it only really matters for binaries here, I think, because of the way those files are locked.

> Personally I would prefer hardlinks to optional on windows in condarc..

As described in detail in the associated issue, noarch packages with entrypoints seem to be a bit broken on Windows currently. It doesn't seem very desirable to have to override an option in condarc to make them work reliably?

mingwandroid:
Windows issues with file locking extend beyond hardlinks, stub executables etc (any dll can cause the same problem). They are orthogonal. Conda aggressively uses hardlinks and I don't think we should special case the entry point scripts.

mingwandroid:
Brokenness needs fixing though. I'm not saying otherwise.

mparry:
> Windows issues with file locking extend beyond hardlinks, stub executables etc (any dll can cause the same problem). They are orthogonal. Conda aggressively uses hardlinks and I don't think we should special case the entry point scripts.

The entry point executable is a bit of a special case though, in the sense that it's the only thing that is breaking installation/upgrades of noarch packages.

I don't disagree with your general point, but I think perhaps this PR is a short-term pragmatic solution, and could be replaced later with a more comprehensive change (if desired). I'm happy to update the change to add a comment or a TODO to this effect, if that would help.

travishathaway:

PR #9980, opened 2020-06-05
fix MatchSpec roundtrip parsing

wolfv:
We have an issue in mamba where the MatchSpec looses some information through the representation.

If you parse `libblas=*=*mkl`, the repr of that is `libblas=[build=*mkl]`
When parsing `libblas=[build=*mkl]` one hits a assert because the version is empty. 

I believe it would be better to have `libblas=*[build=*mkl]` and I attached a fix for this.

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -318,7 +318,10 @@ class MatchSpec(metaclass=MatchSpecType):
             elif version[-2:] == ".*":
                 builder.append("=" + version[:-2])
             elif version[-1] == "*":
-                builder.append("=" + version[:-1])
+                if len(version) == 1:
+                    builder.append("=*")
+                else:
+                    builder.append("=" + version[:-1])
             elif version.startswith("=="):
                 builder.append(version)
                 version_exact = True
@@ -532,6 +535,8 @@ def _parse_version_plus_build(v_plus_b):
         ('*', '*')
     """
     parts = re.search(r'((?:.+?)[^><!,|]?)(?:(?<![=!|,<>~])(?:[ =])([^-=,|<>~]+?))?$', v_plus_b)
+
+
     if parts:
         version, build = parts.groups()
         build = build and build.strip()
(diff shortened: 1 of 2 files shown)


jjhelmus:
@cla-bot check

tests/models/test_match_spec.py
```diff
@@ -137,6 +137,13 @@ def test_hash(self):
         assert c != d
         assert hash(c) != hash(d)
 
+    def test_roundtrip(self):
+        a = MatchSpec("numpy=*=*bla")
+        b = MatchSpec(a.__repr__())
+        assert (a == b)
```
mcg1969, line 143: Seems like it is failing here

conda/models/match_spec.py
```diff
@@ -534,6 +537,8 @@ def _parse_version_plus_build(v_plus_b):
         ('*', '*')
     """
     parts = re.search(r'((?:.+?)[^><!,|]?)(?:(?<![=!|,<>~])(?:[ =])([^-=,|<>~]+?))?$', v_plus_b)
+
```
mcg1969, line 540: flake is complaining about the extra blank here :-(

jezdez:
@wolfv Hey Wolf, I'm not sure if this is current or not, could you let me know if this is still an issue?

wolfv:
@jezdez yes, this is still an issue, I just checked.

Also, there is another issue I've been hitting before in micromamba. When requesting a spec from the command line like so `numpy=1.23` it means "loose 1.23" requirement, and is equal to `numpy 1.23.*`.
However, when requesting `numpy=1.23=*123`, the first single-equals is translated into a double equals, e.g. `numpy==1.23=*123`. However, the double-equals signifies a strong pin so that `numpy==1.23.0` which is a lot less flexible.

```
>>> a = MatchSpec("numpy=1.23=*123")
>>> a.__str__()
'numpy==1.23=*123'
>>> a.conda_build_form()
'numpy 1.23 123'
```

PR #9999, opened 2020-06-11
Update prefix_data.py

### LINKED ISSUE #9917: get_python_version_for_prefix in prefix_data.py doesn't handle two digit minor version
The function `get_python_version_for_prefix` silently truncates two digit minor version numbers. For example, if the Python version number is 3.10, it will return "3.1", which can cause some packages to be installed in the incorrect location.

Of course, there isn't a Python 3.10 yet -- I ran into this while experimenting. FWIW, the accepted [PEP 602](https://www.python.org/dev/peps/pep-0602/) suggests that 3.10 will be the version after 3.9 (as opposed to 4.0). 

https://github.com/conda/conda/blob/2cdd7ab052c2b96c088137f52dc40c613e99ce33/conda/core/prefix_data.py#L370-L380

WinstonPais:
fixes #9917

fixed get_python_version_for_prefix to handle two digit minor version


jjhelmus:
@cla-bot check

PR #10002, opened 2020-06-12
Update unix.py

songmeixu:
As indicated by the URL above this line, the 'gid' should not be used to determine whether the user is an admin.  
I found a problem with this when my 'gid' is 0, but I am not an admin. I don't have access to other users\' '\~/.conda' or '\~/.ssh' directories.  
So this modification was proposed.


jjhelmus:
@cla-bot check

PR #10005, opened 2020-06-13
suppress spurious xonsh activation warnings 

scopatz:
And add real warning when xonsh fails to activate.

This fixes / resolves a really weird and nasty error in xonsh activation that would only occur in certain environments due to multiple shells activating at the same time.


scopatz:
@conda-core - I think this is ready for review, and the errors seem unrelated

scopatz:
CC @kalefranz

mcg1969:
CI issues fixed, I believe

PR #10000, opened 2020-06-11
colorama fix

cjmartian:

PR #10069, opened 2020-07-13
Fix code page issue

mcg1969:
Attempt to correct CI failures on Windows.


mcg1969:
@cjmartian can you check out the Windows failures here? They are definitely related to the work on #10000 .

mcg1969:
This now reverts #10000; cc: @chenghlee @mingwandroid

conda/models/enums.py
```diff
@@ -23,7 +23,7 @@ class Arch(Enum):
     aarch64 = 'aarch64'
     ppc64 = 'ppc64'
     ppc64le = 'ppc64le'
-    s390x   = 's390x'
+    s390x = 's390x'
```
mcg1969, line 26: Flake complained

mcg1969:
We're hitting the 60-minute limit on the appveyor runs. The only thing i can think to do is to matrix the individual tasks. I've identified three:

- non integration tests on a dev install
- integration tests on a dev install
- a full conda package build, which includes the non integration tests

Perhaps we can skip the first of these three since they're being done in the conda package build?

mcg1969:
We're _very_ close here. Now just the package build on windows 2.7 is all that is failing

Issue #9982, opened 2020-06-05
Conda environment is activated, however the python console says the environment is not activated


Arpanio:
## Current Behavior

I am following the conda documentation as well as the troubleshooting guide in order to create a new environment on a Windows machine, activate it, and then enter into the python console.

I can see that the environment is active, however, python console gives me the warning that the environment is not active.

Furthermore, trying to import a module (numpy) that I installed in the current conda environment fails as it instead tries to load it from the base conda environment (this is a whole different headache).

### Steps to Reproduce

The following are the exact steps that I have used:

1. Install Anaconda on a Windows machine (version details in section Environment Information below)
2. Download numpy tar file from conda-forge
3. Create a project folder
4. Open Anaconda Prompt (Powershell) as an Admin and cd to project folder
5. Run the conda commands listed below

```
conda create --prefix ./envs --offline
conda activate /path/to/envs
conda install /path/to/numpy/tar/from/conda-forge
python
```

## Expected Behavior

1. The Powershell should reflect that the active conda environment is envs. This behaviour is observed:
<img width="735" alt="screenshot_env_list_remote" src="https://user-images.githubusercontent.com/7851954/83907255-0c658b80-a75d-11ea-8c7e-6c46ae2a10c5.PNG">
2. On running `python`, the python console should start without the "environment not activated" warning. This is **not the case**:
<img width="735" alt="screenshot_env_python_console_remote" src="https://user-images.githubusercontent.com/7851954/83907467-58183500-a75d-11ea-92dc-1650a0709586.PNG">

Plaintext of warning message: 

> This Python interpreter is in a conda environment, but the environment has not been activated. Libraries may fail to load. To activate this environment please see https://conda.io/activation

3. Trying to load numpy fails as it tries to load from `C:\ProgramData\Anaconda` and not from the conda environment I created.
<img width="736" alt="screenshot_env_python_numpy" src="https://user-images.githubusercontent.com/7851954/83910317-438a6b80-a762-11ea-9023-4e7f53aceeb5.PNG">

## Environment Information
1. OS: Windows Server 2016 (v10.0.14393)
2. Anaconda: V2020.02 py37_0
3. Conda: V4.8.2

I have admin access to the Windows machine. The windows machine does not have internet access (corporate environment).
<details open><summary><code>`conda info`</code></summary><p>

```
active environment : D:\conda_project\envs
    active env location : D:\conda_project\envs
            shell level : 2
       user config file : C:\Users\daspana\.condarc
populated config files :
[...]

PR #10089, opened 2020-07-21
Clarify that Python is not by default included

hugobuddel:
The documentation made it appear as if there will always be a version of Python included in the environment.
This might have been the case in the past; currently it is possible to create an environment without Python at al.
See #9982

### DIFF
--- a/docs/source/user-guide/tasks/manage-environments.rst
+++ b/docs/source/user-guide/tasks/manage-environments.rst
@@ -50,8 +50,7 @@ Use the terminal or an Anaconda Prompt for the following steps:
       proceed ([y]/n)?
 
   This creates the myenv environment in ``/envs/``. No
-  packages will be installed in this environment. Python
-  will be the Python of the base environment.
+  packages will be installed in this environment.
 
 3. To create an environment with a specific version of Python:
 
--- a/docs/source/user-guide/tasks/manage-environments.rst
+++ b/docs/source/user-guide/tasks/manage-environments.rst
@@ -49,9 +49,9 @@ Use the terminal or an Anaconda Prompt for the following steps:
 
       proceed ([y]/n)?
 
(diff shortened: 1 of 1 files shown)


hugobuddel:
> We require contributors to sign our Contributor License Agreement, and we don't have one on file for @hugobuddel. In order for us to review and merge your code, please e-sign the PDF at https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement. We then need to manually verify your signature. We will ping the bot to refresh the PR status when we have confirmed your signature.

Signed.

mcg1969:
@cla-bot check

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -49,9 +49,9 @@ Use the terminal or an Anaconda Prompt for the following steps:
 
       proceed ([y]/n)?
 
-  This creates the myenv environment in ``/envs/``. This
-  environment uses the same version of Python that you are
-  currently using because you did not specify a version.
+  This creates the myenv environment in ``/envs/``. No
+  packages will be installed in this environment. Python
```
mcg1969, line 53: I would suggest removing `Python will be the Python of the base environment.` That's not really a good way to look about it. `conda create` doesn't _activate_ an environment, so all that's happening here is that you're staying in the same environment you are currently in. And strictly speaking that may not be the base environment. In fact, the current environment doesn't even need to have Python in it.

hugobuddel: Sentence removed. I kept it to stay in line with what was there. The text is better without.

hugobuddel:
> See the comment above! Thanks for this though!

After resolving the conversation, I realized it might be convention that only the creator of the thread should mark them as resolved. My apologies if that is the case, this is my first conda PR.

I followed your suggestion, so I suppose you're okay with the change though.

PR #10092, opened 2020-07-22
ENH: exporting env takes CONDA_PREFIX into account

pdnm:
Current behaviors of `conda env export` when no environment is activated:
* runs successfully when `--prefix <prefix>` is provided
* raises error when `CONDA_PREFIX` is set.

I think we should either allow or disallow both scenarios for the sake of consistency.
Furthermore, given that `conda list` does run successfully without an activated env when `CONDA_PREFIX` is set,
I made `conda env export` consider `CONDA_PREFIX` in this PR.


jjhelmus:
@cla-bot check

Issue #10083, opened 2020-07-19
Deprecation warning due to invalid escape sequences


tirkarthi:
## Current Behavior

Deprecation warnings are raised due to invalid escape sequences. This can be fixed by using raw strings or escaping the literals. pyupgrade also helps in automatic conversion : https://github.com/asottile/pyupgrade/

### Steps to Reproduce

```
find . -iname '*.py' | grep -Ev 'vendor|example|doc' | xargs -P4 -I{} python3.8 -Wall -m py_compile {}          
./conda/core/subdir_data.py:56: DeprecationWarning: invalid escape sequence \}
  REPODATA_HEADER_RE = b'"(_etag|_mod|_cache_control)":[ ]?"(.*?[^\\\\])"[,\}\s]'  # NOQA
./conda/common/path.py:306: DeprecationWarning: invalid escape sequence \/
  path_re = '(?<![:/^a-zA-Z])([a-zA-Z]:[\/\\\\]+(?:[^:*?"<>|]+[\/\\\\]+)*[^:*?"<>|;\/\\\\]+?(?![a-zA-Z]:))'  # noqa
./conda/gateways/disk/link.py:211: DeprecationWarning: invalid escape sequence \ 
  """
./conda/gateways/disk/link.py:250: DeprecationWarning: invalid escape sequence \*
  """
./conda/gateways/disk/create.py:85: DeprecationWarning: invalid escape sequence \.
  python_entry_point_template = dals("""
./test_data/env_metadata/__init__.py:4: DeprecationWarning: invalid escape sequence \(
  """
```

chenghlee:
Fixed in PR#10083.

PR #10093, opened 2020-07-22
Fix some deprecation warnings for imports and regex: #10083

angloyna:
Removed some deprecation warnings and a py38 incompatible collections import


angloyna:
This ticket should fix issue #10083

PR #10114, opened 2020-07-29
Fix #9589: python 3.8 multiprocessing warning

### LINKED ISSUE #9589: Warning about leaked semaphore under Python 3.8
### Current Behavior
Under Python 3.8, there is a warning printed when running most conda commands:
```
~/miniconda3/lib/python3.8/multiprocessing/resource_tracker.py:216:
UserWarning: resource_tracker: There appear to be 1 leaked semaphore
objects to clean up at shutdown
```

### Steps to Reproduce
Running more or less any conda command will reproduce it for me, e.g. `conda env list`

## Expected Behavior
There should be no warning

## Environment Information
`conda-4.8.1-py38_0` and `python-3.8.1-h359304d_1` on MacOS (Catalina)

angloyna:


mparry:
Ah, I didn't see that you had opened this PR. I have also opened one (#10115) but with a different approach. I did consider dropped the vendored copy of `tqdm` and just importing an installed version, but it seemed like the vendored approach was still desirable - see #2825.

angloyna:
@mparry oh don't worry, i left this in draft because i don't plan on merging. I just wanted to see if the external tqdm would fix the multiprocessing warning for py38. Let me pass your PR along to the team. We've updated the installers to use python 3.8 by default, so I think this should probably be remedied sooner than later.

PR #10134, opened 2020-08-06
add arm64 to non_x86_machines

jjhelmus:
On ARM based macOS systems platform.machines() returns arm64


jjhelmus:
This follows up on #10128

PR #10137, opened 2020-08-07
Fix import error caused by `non_x86_machines` rename

chenghlee:
This import error was originally missed because several PRs were not merged in the order in which they were submitted.

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -187,11 +187,11 @@ def _supplement_index_with_system(index):
 
 
 def get_archspec_name():
-    from conda.base.context import non_x86_linux_machines, _arch_names, _platform_map
+    from conda.base.context import non_x86_machines, _arch_names, _platform_map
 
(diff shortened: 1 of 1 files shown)


conda/core/index.py
```diff
@@ -187,11 +187,11 @@ def _supplement_index_with_system(index):
 
 
 def get_archspec_name():
-    from conda.base.context import non_x86_linux_machines, _arch_names, _platform_map
+    from conda.base.context import non_x86_machines, _arch_names, _platform_map
 
     target_plat, target_arch = context.subdir.split("-")
     # This has to reverse what Context.subdir is doing
-    if target_plat == "linux" and target_arch in non_x86_linux_machines:
+    if target_plat == "linux" and target_arch in non_x86_machines:
```
jjhelmus, line 194:
```suggestion
    if target_arch in non_x86_machines:
```

PR #10186, opened 2020-08-24
Translate Azure Windows CI to GitHub Actions

angloyna:
IT would prefer that we migrate toward using fewer CI vendors, and there has been some consolidating around GitHub Actions. I've translated the azure work into a GitHub Action.


angloyna:
Have a ticket with IT to transfer over the status checks for master from appveyor to the GitHub Action builds

PR #10189, opened 2020-08-25
logz: Swap from pformat to json.dumps when stringifying json response…

mingwandroid:
…s for debug


mingwandroid:
Thanks @angloyna

PR #10196, opened 2020-08-27
Ignore virtual packages when checking environment consistency

ericpre:
In https://github.com/conda-forge/prismatic_split-feedstock/pull/10, I have added `__cuda` as a run dependency to restrict the installation of the GPU build on compatible system, which is working very well! A side effect of this is to make the environment inconsistent because the `__cuda` virtual package is not installed.

For example, on an CUDA capable linux system (there is no cuda package for windows):
```
$ conda install prismatic_cli -c conda-forge
$ conda update --all
Solving environment: / 
The environment is inconsistent, please check the package plan carefully
The following packages are causing the inconsistency:

  - conda-forge/linux-64::prismatic_cli==1.2.1=gpu_h5deb2e2_108
done

# All requested packages already installed.
```

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -911,7 +911,9 @@ class Resolve(object):
         for prec in itervalues(self.index):
             nkey = C.Not(self.to_sat_name(prec))
             for ms in self.ms_depends(prec):
-                C.Require(C.Or, nkey, self.push_MatchSpec(C, ms))
+                # Virtual packages can't be installed, we ignore them
+                if not ms.name.startswith('__'):
+                    C.Require(C.Or, nkey, self.push_MatchSpec(C, ms))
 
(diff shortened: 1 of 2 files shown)


ericpre:
I forgot to mention that the issue is with the warning being wrong, conda doesn't raise any error and there is no inconsistency in the environment. This PR fixes the warning and the approach is similar as the one used in https://github.com/conda/conda/pull/10117.

@jjhelmus, @chenghlee, could you please have a look at this PR?

ericpre:
Anyone to review this PR?

chenghlee:
Closing & re-opening this PR to force checks with new CI system.

ericpre:
Thanks @chenghlee, CI seems to be happy now!

chenghlee (approved):

PR #10208, opened 2020-09-01
Ignore package installations that match specifications

InfiniteChai:
Looking for comment. Interested to optimise conda installation with explicit files so that it will not unlink/link when we have matching specifications already installed (so that reruns of the installation become comparatively cheap).

### DIFF
--- a/conda/misc.py
+++ b/conda/misc.py
@@ -97,10 +97,14 @@ def explicit(specs, prefix, verbose=False, force_extract=True, index_args=None, 
 
         prec = prefix_data.get(pcrec.name, None)
         if prec:
-            precs_to_remove.append(prec)
+            # If we've already got matching specifications, then don't bother re-linking it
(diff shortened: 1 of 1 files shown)


jjhelmus:
@cla-bot check

chenghlee:
@InfiniteChai: Thanks for the PR; this looks like a good addition, but it seems to breaking some behavior with the `__glibc` and `__cuda` virtual packages. I'm working on the solver messaging and behavior at the moment that may fix the issues, and once that's done, I should be take further action on this PR.

chenghlee:
Closing & re-opening PR to trigger new CI checks.

chenghlee (approved):

InfiniteChai:
@chenghlee that's great, thanks!

y2kbugger:
Anyone consider whether any users have workflows that involve hot-patching env and then using conda install to revert?

Issue #9412, opened 2019-11-05
Provide a way to run `conda run` w/o buffering stdout (and output in general)


brettcannon:
Currently `conda run` buffers everything going to stdout (I didn't test stderr). It would be great to be able to get streaming results from the running command rather than having to wait until the application exits.

msarahan:
Yes, this would be a big improvement.  I've looked at ways to "tee" this output, but it ended up being too deep of a rabbit hole at the time.  If you have any recommendations for how to achieve this, please let us know.

mingwandroid:
conda-run also misses any way to handle stdin. IMHO a thread should be launched for all the IO.

brettcannon:
Does `conda run` do anything after the process exits? Could `conda run` use [a `os.exec*()` function](https://docs.python.org/3/library/os.html#os.execv) to substitute the conda process for the Python process in-place so that stdout and friends aren't buffered because Python now owns stdout and friends?

earhart:
FWIW -- as a workaround, I've noticed that using "conda run -n my_env bash -c 'my_actual_command > /dev/tty 2>&1'" seems to work on Linux (and should work fine on any Posix system).

jcmuel:
For those who need something similar to an unbuffered version of "conda run":  The new `--no-capture-output` option (https://github.com/conda/conda/pull/9646) in conda 4.9.0 (see [CHANGELOG](https://github.com/conda/conda/blob/master/CHANGELOG.md)) has a similar effect and solves the problem for me.

    conda run --no-capture-output python3 -u -c $'import time;\nfor x in range(0, 9): print(0); time.sleep(.5)'

kalindkaria:
Thank you for this update! It saved a lot of my time.

kenodegard:
Confirming this is addressed with the `--no-capture-output` argument.

PR #10270, opened 2020-10-08
Enabling "conda run" to display real time output for the subprocess

gabrielcnr:
This enables the command `conda run` to display the live output from the subprocess's stdout and stderr, in real time.

The current behaviour-- which prints the stdout and stderr only after the process has finished --remains unchanged. 

The `conda run` command now can take a flag `--live-stream` which would flip the behaviour to provide the real time output.

I believe this is related to: https://github.com/conda/conda/issues/9412

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -1097,6 +1097,14 @@ def configure_parser_run(sub_parsers):
         help="Executable name, with additional arguments to be passed to the executable "
              "on invocation.",
     )
+
+    p.add_argument(
+        '--live-stream',
+        action="store_true",
+        default=False,
+        help="Display the output for the subprocess stdout and stderr on real time.",
+    )
+
     p.set_defaults(func='.main_run.execute')
 
 
--- a/conda/cli/main_run.py
+++ b/conda/cli/main_run.py
@@ -25,7 +25,7 @@ def execute(args, parser):
                                                        args.dev, args.debug_wrapper_scripts, call)
     env = encode_environment(os.environ.copy())
 
-    response = subprocess_call(command_args, env=env, path=cwd, raise_on_error=False)
+    response = subprocess_call(command_args, env=env, path=cwd, raise_on_error=False, live_stream=args.live_stream)
     if response.rc != 0:
         log = getLogger(__name__)
         log.error("Subprocess for 'conda run {}' command failed.  (See above for error)"
@@ -37,8 +37,9 @@ def execute(args, parser):
             log = getLogger(__name__)
             log.warning('CONDA_TEST_SAVE_TEMPS :: retaining main_run script_caller {}'.format(
                 script_caller))
-    if response.stdout:
-        print(response.stdout, file=sys.stdout)
-    if response.stderr:
-        print(response.stderr, file=sys.stderr)
+    if not args.live_stream:
+        if response.stdout:
+            print(response.stdout, file=sys.stdout)
+        if response.stderr:
+            print(response.stderr, file=sys.stderr)
     return response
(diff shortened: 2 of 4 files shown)


marcelotrevisani:
Hi @chenghlee, could you please take a look on this PR?

mingwandroid (approved):
LGTM. I'd probably say it should be the default to be honest with you. Also, can you handle stdin here? Seems like it should be possible.

gabrielcnr:
I think this should be the default too. I didn't make it, because I wasn't sure if would be desirable to break the current behaviour, so I opted for keeping it backwards compatible. But I guess the impact of the change is not big.
As for the stdin, I think we can address that in a separate PR in the future. What do you think @mingwandroid, @marcelotrevisani ?

marcelotrevisani:
> I think this should be the default too. I didn't make it, because I wasn't sure if would be desirable to break the current behaviour, so I opted for keeping it backwards compatible. But I guess the impact of the change is not big.
> As for the stdin, I think we can address that in a separate PR in the future. What do you think @mingwandroid, @marcelotrevisani ?

I agree that @mingwandroid suggestion can be addressed after as the modification to also support the stdin will be a bit more complicated to implement. And I also agree with @mingwandroid that it should be the default behaviour.

gabrielcnr:
Thanks for the feedback @marcelotrevisani and @mingwandroid. As part of this PR, can you please confirm if I should:

1) make `True` the default option for the `--live-stream` flag
or
2) remove the `--live-stream` flag entirely, but keep the live stream behaviour as the new default behaviour
or
3) nothing

Please let me know and I will happily make the changes. Thank you.

marcelotrevisani:
> Thanks for the feedback @marcelotrevisani and @mingwandroid. As part of this PR, can you please confirm if I should:
> 
> 1. make `True` the default option for the `--live-stream` flag
>    or
> 2. remove the `--live-stream` flag entirely, but keep the live stream behaviour as the new default behaviour
>    or
> 3. nothing
> 
> Please let me know and I will happily make the changes. Thank you.

I believe that is a question for @chenghlee or @mingwandroid 
But anyways we could try to merge it as it is and next release we can let it as a default, but I believe @chenghlee is leading the conda releases so he should know what is the best decision for conda in this case

marcelotrevisani:
just one thing, could you please resolve the conflicts with your PR?

chenghlee:
> > Thanks for the feedback @marcelotrevisani and @mingwandroid. As part of this PR, can you please confirm if I should:
> > 
> > 1. make `True` the default option for the `--live-stream` flag
> >    or
> > 2. remove the `--live-stream` flag entirely, but keep the live stream behaviour as the new default behaviour
> >    or
> > 3. nothing
> > 
> > Please let me know and I will happily make the changes. Thank you.
> 
> I believe that is a question for @chenghlee or @mingwandroid
> But anyways we could try to merge it as it is and next release we can let it as a default, but I believe @chenghlee is leading the conda releases so he should know what is the best decision for conda in this case

For now (i.e., the 4.9.0 release), let's just leave it as is (i.e., option 3 -- do nothing), since this is a new behavior and I want to avoid any surprises for current users.

I agree that that `--live-stream` should be the default, but we should roll that change into the 5.0.0 release, where we're planning on making a bunch of other default behavior changes. 5.0.0 is currently scheduled for late 2020-Q4/early 2020-Q1, and that should give us enough time for users to test out this feature and for us to announce upcoming major changes.

chenghlee:
Once I'll unbreak the CI system, I'll review and merge this PR. Expect that to happen sometime later today.

jonashaag:
Pardon the naive question but what is the point of `--live-stream` when we have `--no-capture-output`?

PR #10295, opened 2020-10-19
Propose clarifcation on merge strategy.

rainwoodman:
This section is a big vague. Could you clarify this section a bit more?

I am not sure if my understanding is correct:
My best guess is that for a list key 'merge' here means extend, and for a dict 'merge' means update. (Per python method names).

It however could also be possible merge here means recursively merge items of a list / values of a dict.

┆Issue is synchronized with this [Jira Task](https://anaconda.atlassian.net/browse/CONDAORG-237)


travishathaway:
Hi @rainwoodman,

We appreciate the suggestion, but we are going to he documentation as-is for now.

To improve the section it would be best to include actual examples and explain it with sentences.

PR #10298, opened 2020-10-20
Remove preload_openssl, it is so so wrong

mingwandroid:


chenghlee (changes requested):
I think you missed another call to preload_openssl() in `conda/gateways/connection/download.py`:
```
def download_text(url):
    if sys.platform == 'win32':
        preload_openssl()
```

Once that gets removed, the tests should pass and the PR should be good to go.

mingwandroid:
I took out the other PR about the autorun, we can deal with that separately

chenghlee (approved):

PR #10306, opened 2020-10-22
Honor --download-only flag for env creation using file explicit spec closes #10299

### LINKED ISSUE #10299: conda create using explicit file does not respect --download-only
On conda 4.8.5:
Using an explicit file (such as retrieved using conda list --explicit), when you use it  with --download only , the prefix is actually created.
Repro:

```
conda activate base
conda list --explicit > /tmp/explicitfile.yml
conda create --prefix /tmp/myprefix --file /tmp/explicitfile.yml --download-only
```

Now you can see the full install has been done in /tmp/myprefix

JJsrc:
Do not run the LinkUnlinkTransaction if --download-only is specified, even if using a file to create environment

┆Issue is synchronized with this [Jira Task](https://anaconda.atlassian.net/browse/CONDAORG-242)

### DIFF
--- a/conda/misc.py
+++ b/conda/misc.py
@@ -14,6 +14,7 @@ import re
 import shutil
 import sys
 
+from conda import CondaExitZero
 from .base.context import context
 from .common.compat import itervalues, on_win, open, scandir
 from .common.path import expand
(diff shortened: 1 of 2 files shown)


jezdez:
@anaconda-issue-bot check

travishathaway:
This pull request has become stale; therefore, we are going to close it. If you are still interested in developing it further, please re-open it.

PR #10313, opened 2020-10-24
Respect PEP440 ~= 'Compatible release clause'

mingwandroid:
.. we expand this to >=X.Y.Z,==X.Y.* as per the PEP

### DIFF
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -99,7 +99,7 @@ spec_pat = re.compile(r'(?P<name>[^=<>!\s]+)'  # package name  # lgtm [py/regex/
                       r'('
                       r'(?P<cc>=[^=]+(=[^=]+)?)'  # conda constraint
                       r'|'
-                      r'(?P<pc>(?:[=!]=|[><]=?).+)'  # new (pip-style) constraint(s)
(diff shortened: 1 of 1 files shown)


mingwandroid:
conda-build has an implementation of this (seems unused) that I resurrected. It gives different results:

`kubernetes >=11.0.0,<11.1` vs `kubernetes >=11.0.0,==11.0.*`

.. which do we want? In general it seems conda-build has changed to defer to conda in this matter and the old code-path is redundant.

chenghlee:
> conda-build has an implementation of this (seems unused) that I resurrected. It gives different results:
> 
> `kubernetes >=11.0.0,<11.1` vs `kubernetes >=11.0.0,==11.0.*`
> 
> .. which do we want? In general it seems conda-build has changed to defer to conda in this matter and the old code-path is redundant.

If we're going for PEP440 compliance, I would say the latter (`kubernetes >=11.0.0,==11.0.*`).

chenghlee (approved):
LGTM

mingwandroid:
Thanks! I'll adjust my conda-build PR, together this fixes https://github.com/conda/conda-build/issues/4108

PR #10350, opened 2020-11-10
Use old filenames for "normal" repodata cache files

chenghlee:
Partially revert changes made in commit ad75abcad, so when the `use_only_tar_bz2` option is set to false (default), conda uses the same cache filenames it did before.  This helps avoid breaking certain applications that made assumptions about the repodata cache filenames.


cjmartian (approved):

mingwandroid:
LGTM

PR #10478, opened 2021-01-26
Support python 3.10 version number

opoplawski:
Just taking a stab here...


opoplawski:
Seems like a proper regexp would be much more appropriate here.

chenghlee (approved):

PR #9037, opened 2019-08-02
Fix boto3 initialization race condition

hajapy:
Fixes https://github.com/conda/conda/issues/8993

PR #10516, opened 2021-02-18
Don't create an unused s3 client at import time.

dan-hook:
The deleted lines could cause issues when using a [credential process provider](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html).  

Specifically, during test discovery in vscode, conda info would run and import this module.  Creating the client would trigger the process provider to run.  This particular process provider would request credentials, but the prompt would not be visible as conda info was run in the background.

There's a number of places that things went wrong in this scenario, put this change appears to be a simple fix.

### DIFF
--- a/conda/gateways/connection/adapters/s3.py
+++ b/conda/gateways/connection/adapters/s3.py
@@ -11,7 +11,6 @@ have_boto3 = have_boto = False
 try:
     import boto3
     have_boto3 = True
-    boto3.client('s3')  # https://github.com/conda/conda/issues/8993
 except ImportError:
     try:
         import boto


cjmartian:
@anaconda-issue-bot check

jezdez (approved):
This seems like a small win.

hajapy:
Did anyone test that conda still works with s3 channels after this change?

The reason it was originally implemented was because boto3 sessions lack thread safety and needs to happen on the main thread before being used in a multi-threaded way. See:
https://github.com/boto/botocore/issues/1246
https://github.com/boto/boto3/pull/2848/files

It seems like this fix addresses a new issue (with a specialized type of credential provider configuration when using a vscode) but regresses on the original issue (which could affect anyone who uses s3 channels with conda).

If the concern is to avoid making this call to initialize a client (thus obtaining credentials) unless actually needed, a thread-safe pattern could be implement whereby if a client hasn't yet been initialized on the main thread, a threading.Lock is used to create it before making boto3.resource('s3') calls.

jezdez:
> Did anyone test that conda still works with s3 channels after this change?
> 
> The reason it was originally implemented was because boto3 sessions lack thread safety and needs to happen on the main thread before being used in a multi-threaded way. See:
> [boto/botocore#1246](https://github.com/boto/botocore/issues/1246)
> https://github.com/boto/boto3/pull/2848/files

Hey @hajapy, thanks for bringing this to my attention, this is useful. Generally speaking it was not clear to me when I reviewed [the old fix](https://github.com/conda/conda/pull/9037) for #8993 that this line was needed for the fix to hold since there wasn't a clear code comment or test to indicate it. That's a regression for sure and needs a better patch.

> It seems like this fix addresses a new issue (with a specialized type of credential provider configuration when using a vscode) but regresses on the original issue (which could affect anyone who uses s3 channels with conda).

Yeah, import-time initialization of the boto3 config mechanism is a UX problem for the majority of Conda users who don't use S3 (independent of which IDE is used). This should not have had a negative side effect of prior fixes of course.

> If the concern is to avoid making this call to initialize a client (thus obtaining credentials) unless actually needed, a thread-safe pattern could be implement whereby if a client hasn't yet been initialized on the main thread, a threading.Lock is used to create it before making boto3.resource('s3') calls.

That's the concern indeed that #10516 tried to address and a lock seems appropriate to make it compatible with your patch. Do you want to take a stab at the proper fix?

By the way, would you mind opening a separate ticket so we can keep track of the regression and do good record keeping?

Thank you again for raising your concerns about this patch.

hajapy:
@jezdez thanks, I've confirmed the original issue has re-appeared in conda 4.10.2 (s3 channels are not usable). I'll create an issue with details. Adding testing for this may be tricky. I'm not sure I have the capacity to contribute another fix at this time.

hajapy:
@jezdez https://github.com/conda/conda/issues/10756

jezdez:
That’s alright, thanks for opening the issue, I’ll work on a fix in the coming days and do a bug fix release.

Oh btw, do you happen to have a S3 based channel available that I can test this with?

hajapy:
> That’s alright, thanks for opening the issue, I’ll work on a fix in the coming days and do a bug fix release. Oh btw, do you happen to have a S3 based channel available that I can test this with?
> […](#)
> Am 29.06.2021 um 19:34 schrieb Andrew Achkar ***@***.***>: ﻿ @jezdez thanks, I've confirmed the original issue has re-appeared in conda 4.10.2 (s3 channels are not usable). I'll create an issue with details. Adding testing for this may be tricky. I'm not sure I have the capacity to contribute another fix at this time. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

No, unfortunately those s3 channels are private.

jezdez:
> > Oh btw, do you happen to have a S3 based channel available that I can test this with?
> 
> No, unfortunately those s3 channels are private.

No problem, thanks for your help in any case :)

PR #10578, opened 2021-03-29
Pilot metdata verification

chenghlee:
Allow conda to verify metadata signatures when the `extra_safety_checks` configuration option is set to `True` and the `conda-content-trust` package is installed in the base environment.


awwad:
Looks good to me. :) 👍

awwad:
Lesson: when reviewing, clarify what you have reviewed.  One of us should have started a clean environment using this PR branch and installed from scratch.  We would have found that conda-content-trust is always treated as not installed....  (I'm puzzled, I would have thought I'd done that, but... I guess I must have used my existing environment.)  Missed a bug introduced in a typo in e670fda.

PR #10609, opened 2021-04-12
Remove expensive deepcopy call from resolve

bnemanich:
Profiling of some of my `conda create --dry-run` runs was showing that around 45% of the time was being spent in the `deepcopy` call within `get_reduced_index` in `resolve.py`. This PR removes the `deepcopy` call and replaces it with a dictionary containing only the values within `specs_by_name` that are different than the values in `specs_by_name_seed`. This has lead to the time it took to call `conda create --dry-run` being cut almost in half.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -709,11 +709,11 @@ class Resolve(object):
                 #    broadening check to apply across packages at the explicit level; only
                 #    at the level of deps below that explicit package.
                 seen_specs = set()
-                specs_by_name = copy.deepcopy(specs_by_name_seed)
+                specs_by_name = dict()
 
                 dep_specs = set(self.ms_depends(pkg))
                 for dep in dep_specs:
-                    specs = specs_by_name.get(dep.name, list())
+                    specs = specs_by_name_seed.get(dep.name, list())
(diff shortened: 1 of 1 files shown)


bnemanich:
@chenghlee What do you think about this?

chenghlee:
@bnemanich this looks like a good suggestion; but I need to do some digging to determine why `deepcopy` was used in the first place and whether this change will break things.

bnemanich:
@chenghlee Have you had anytime to look at this one?

jezdez:
Sorry for the slow response @bnemanich, it took a while to review this and sadly we found a change of using deepcopy in this part of the solver to be too high of a risk since it could have undesirable side-effects in resolving a lazy dependency trees. Conda's solver has lots of implicit behavior when creating solutions and the deepcopy here is to my knowledge needed to force the spec map to be built in-place.

That said, we're ARE working currently on a different project to integrate a [faster solver based on libmamba/libsolv](https://github.com/conda/conda/pull/10881) that will make this type of performance bottleneck moot/ We currently plan to release it as an opt-in feature in Q1/22. It would be great if you'd take a look at that when the time comes. Thanks!

PR #10621, opened 2021-04-15
AV: logs: correct log level for root metadata retrieval

awwad:
conda already having the latest version of root metadata built in is a normal and common case.  It should not be generating a warning, so I dropped it to debug level and reworded it slightly.  (Perhaps info is better?)  It looks to folks like something is wrong when it's not.

Also clarified that the info line on fetching updated root metadata is looking to see if there is a newer root file.  (There will almost always not be.)

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -251,8 +251,8 @@ class SubdirData(object):
                     latest_root_id, latest_root_path = cur_id, cur_path
 
         if latest_root_path is None:
-            log.warn(f"Could not find root metadata in {context.av_data_dir}. "
-                     "Falling back to built-in default.")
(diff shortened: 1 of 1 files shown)


awwad:
I do not like this linter's opinions about line continuation indents.  Hehe.  😅

dbast:
LGTM. Debug is better than warn in that case (I would use info to inform the user about important business decisions, which seems not the case here.)

Even though fstrings are now possible within the conda codebase as we ditched older python versions, we should keep in mind, that for logging fstrings are not always the best option, as the string is always computed, no matter if the log level is activate or not ... the old syntax can be better for expensive statements, see https://docs.python.org/3/howto/logging.html#optimization

dbast (approved):

Issue #10326, opened 2020-10-30
changeps1: False does not work for PowerShell


mika-fischer:
It seems `Conda.ps1m` unconditionally renames the prompt function without checking the `changeps1` setting from `.condarc`: https://github.com/conda/conda/blob/a9f5f25d92d58d41c088e4513610b9c33ea3a683/conda/shell/condabin/Conda.psm1#L268-L275

So if I set `changeps1: False`, then my custom prompt function is renamed, conda does not set it's own and I'm left with `PS>`...

stinos:
This was broken by https://github.com/conda/conda/commit/ca16da40117a759b13e9cd27feeeef05a32b28a7. PR with fix here: https://github.com/conda/conda/pull/10624

PR #8774, opened 2019-06-12
Fix renaming of existing prompt function

ScottEvtuch:
Fixes #8773

PR #10624, opened 2021-04-19
Fix changeps1 implementation for Powershell

stinos:
Since commit ca16da40 using changeps1=false just results in
Add-CondaEnvironmentToPrompt not being called, but still unconditionally
renaming the global prompt function so the end result is the user
ends up with no prompt at all (well, just 'PS>').
Fix this by making everything related to the prompt conditional, but
execute all of that in the module itself (and add arguments to the
module in order to achieve that); see previous issues related to the
prompt function: it should be altered in the global scope and not
in function scope.

Fixes https://github.com/conda/conda/issues/10326, also see https://github.com/conda/conda/pull/8774.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -1055,6 +1055,7 @@ class PowerShellActivator(_Activator):
                 $Env:_CE_CONDA = "conda"
                 $Env:_CONDA_ROOT = "{python_path}{s}conda"
                 $Env:_CONDA_EXE = "{context.conda_exe}"
+                $CondaModuleArgs = @{{ChangePs1 = ${context.changeps1}}}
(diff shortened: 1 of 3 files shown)


dbast:
@stinos Checked your license agreement and it is missing your GitHub username `stinos`.

stinos:
@dbast sorry about that. I admit it wasn't really clear to me what to enter where, I now realize I was just reading the labels in the worng order (i.e. I thought the 'enter email here' applied to the 'github username' field). Didn't find a way to edit so I signed the agreement again, now hopefully correct.

dbast:
@anaconda-issue-bot check

chenghlee (approved):

PR #10631, opened 2021-04-22
fix some issues in conda.sh (injected in .bashrc)

lightmare:
I found some strange behaviour in the shell script that's sourced into `.bashrc`. I may have overlooked or misunderstood some important details, so please triple-check whether the changes make sense.

There were some issues with the function `__conda_activate() `
```sh
    CONDA_INTERNAL_OLDPATH="${PATH}"
    __add_sys_prefix_to_path
    ask_conda="$(PS1="$PS1" "$CONDA_EXE" $_CE_M $_CE_CONDA shell.posix "$cmd" "$@")" || \return $?
    rc=$?
    PATH="${CONDA_INTERNAL_OLDPATH}"
```

1. `rc` is not local, and is useless, because `$?` is always `0` at that point
2. if the command inside `$( ... )` fails for some reason, the function returns without restoring `PATH`

The second problem was also present in function `__conda_reactivate()`.

Function `conda()` properly restored the `PATH` before returning, but the `save; modify; restore` pattern is so wordy and error-prone that I think it's worth replacing all its occurrences with a simple call to a function that runs in a subshell, so that env modification cannot leak by accident.

So that's the first commit.

And once it got less wordy, I was able to see that the function `conda()` doesn't need a special case for 0 arguments (the catch-all `case (*)` handles that just fine), so that's the second commit.

### DIFF
--- a/conda/shell/etc/profile.d/conda.sh
+++ b/conda/shell/etc/profile.d/conda.sh
@@ -24,6 +24,11 @@ __add_sys_prefix_to_path() {
     \export PATH
 }
 
+__conda_exe() (
+    __add_sys_prefix_to_path
+    "$CONDA_EXE" $_CE_M $_CE_CONDA "$@"
+)
+
 __conda_hashr() {
     if [ -n "${ZSH_VERSION:+x}" ]; then
         \rehash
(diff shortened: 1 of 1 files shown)


dbast:
@anaconda-issue-bot check

jezdez:
@lightmare Hey there, I was wondering about this PR being a draft at the moment, do you plan on working on this more?

lightmare:
@jezdez No I don't have anything to add. I probably just wanted to live with the modified script for a while before advancing, and then forgot about it.

jezdez:
@lightmare Got it, thanks for working on this!

jezdez (approved):
Reproduced and fixes the mentioned issues. Thank you @lightmare!

PR #10667, opened 2021-05-14
Add RSA SecurID server support

milljm:
Initial commit to aid dealing with RSA SecurID protected servers.

Requires a third party script to facilitate authentication (the pickling
of authenticated cookie): https://github.com/milljm/rsasecure_login

With everyone's blessing, I will move the 3rd party helper script over to conda-forge.
Or do this first if that is a requirement?

### DIFF
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -11,6 +11,7 @@ from . import (AuthBase, BaseAdapter, HTTPAdapter, Session, _basic_auth_str,
 from .adapters.ftp import FTPAdapter
 from .adapters.localfs import LocalFSAdapter
 from .adapters.s3 import S3Adapter
+from .adapters.rsa import SecureIDAdapter
 from ..anaconda_client import read_binstar_tokens
 from ..._vendor.auxlib.ish import dals
 from ...base.constants import CONDA_HOMEPAGE_URL
@@ -29,6 +30,7 @@ CONDA_SESSION_SCHEMES = frozenset((
     "https",
     "ftp",
     "s3",
+    "rsa",
     "file",
 ))
 
@@ -79,6 +81,7 @@ class CondaSession(Session):
             self.mount("https://", unused_adapter)
             self.mount("ftp://", unused_adapter)
             self.mount("s3://", unused_adapter)
+            self.mount("rsa://", unused_adapter)
 
         else:
             # Configure retries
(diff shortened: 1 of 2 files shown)


dbast:
@anaconda-issue-bot check

milljm:
@dbast I am not sure what the tests are not firing. Anything you can trigger?

milljm:
Please re-test.

milljm:
Conda-Forge companion script: https://github.com/conda-forge/staged-recipes/pull/15016

conda/gateways/connection/session.py
```diff
@@ -91,6 +94,7 @@ def __init__(self):
             self.mount("https://", http_adapter)
             self.mount("ftp://", FTPAdapter())
             self.mount("s3://", S3Adapter())
+            self.mount("rsa://", SecureIDAdapter())
```
chenghlee, line 97: Is `rsa://` a documented/standardized URL scheme component for this server?

milljm: Unfortunately not. I did some basic research on the matter using s3:// as an example. Couldn't figure out how to connect to amazon using that prefix (while using requests). So I figured I might be allowed to do this for servers protected by RSASecurID.

I realize it's pretty hacky! Suggestions?

milljm: Perhaps I should try to figure out if _this_ server is protected by RSA, do all the things...

chenghlee: I think adding support for 2FA and other (HTTP-based) authentication mechanisms to conda is a good idea. But I want to avoid starting (or continuing, I guess) the pattern of requiring to create a custom adapter for every authn scheme, especially for schemes like this that appear to be just slightly modified HTTP requests.

I'm going to bump this out of the 4.10.2 release, so we can maybe discuss some other, more generic ways of handling HTTP-based authn.

milljm: Understood. I can try to dig around Conda's internals and see if I can come up with something on my own as well. I'll gladly post another PR if/when that occurs.

Cheers!

travishathaway:
Hi @milljm,

I just want to let you know that we are currently considering whether to make the `CondaSession` object a plugin hook (i.e. replace it with a different class). This could be an interesting use case for that.

If you are interested in that discussion, please check out the following CEP (conda enhancement proposal). It's still a draft, but if you had any input, that would be very much welcome:

- https://github.com/conda-incubator/ceps/pull/44

PR #10674, opened 2021-05-18
Fix ValueError: invalid literal for int() with base 10: '' in _screen_shape_linux

### LINKED ISSUE #10673: Fix ValueError: invalid literal for int() with base 10: '' in tqdm/utils.py
Valid for conda version > 4.8.3
At line-336 in `_vendors/tqdm/utils.py` (https://github.com/conda/conda/commit/db380cc9a1fbeedb5efab84750d4437f1ab72d45#diff-a2e083f5ce1f74db5268f51073df0c2cdb07b7e2d64f65be3e65f5964e023417), `os.environ` is called which returns an empty string and thus fails with value error when `int()` is called on it. In the previous code, a default value of `1` was used (line-204 in `_vendors/tqdm/_utils.py` ).  I suggest that the line-336 in `_vendors/tqdm/utils.py`  should be changed to `return [int(os.environ.get(i, 1)) - 1 for i in ("COLUMNS", "LINES")]`

### Steps to Reproduce

```

```

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```

```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```

```
</p></details>

<details><summary><code>`conda list --show-channel-urls`</code></summary><p>

```

```
</p></details>

erastogi:
Fixes #10673.

This PR fixes a bug in `tqdm/utils.py`
For conda version > 4.8.3
At https://github.com/conda/conda/commit/db380cc9a1fbeedb5efab84750d4437f1ab72d45#diff-a2e083f5ce1f74db5268f51073df0c2cdb07b7e2d64f65be3e65f5964e023417R336), os.environ is called which returns an empty string and thus fails with value error when int() is called on it.

In the previous code, a default value of 1 was used (line-204 in _vendors/tqdm/_utils.py ).

I suggest that the line-336 in `_vendors/tqdm/utils.py` should be changed to return `[int(os.environ.get(i, 1)) - 1 for i in ("COLUMNS", "LINES")]`

### DIFF
--- a/conda/_vendor/tqdm/utils.py
+++ b/conda/_vendor/tqdm/utils.py
@@ -366,8 +366,10 @@ def _screen_shape_linux(fp):  # pragma: no cover
         except:
             try:
                 return [int(os.environ[i]) - 1 for i in ("COLUMNS", "LINES")]
-            except KeyError, ValueError:
+            except KeyError:
                 return None, None
+            except ValueError:
+                return [0, 0]
 
 
 def _environ_cols_wrapper():  # pragma: no cover
--- a/conda/_vendor/tqdm/utils.py
+++ b/conda/_vendor/tqdm/utils.py
@@ -365,7 +365,7 @@ def _screen_shape_linux(fp):  # pragma: no cover
             return cols, rows
         except:
             try:
-                return [int(os.environ[i]) - 1 for i in ("COLUMNS", "LINES")]
+                return [int(os.environ.get(i, 1)) - 1 for i in ("COLUMNS", "LINES")]
             except KeyError:
                 return None, None
 
--- a/conda/_vendor/tqdm/utils.py
+++ b/conda/_vendor/tqdm/utils.py
@@ -365,8 +365,8 @@ def _screen_shape_linux(fp):  # pragma: no cover
             return cols, rows
         except:
             try:
-                return [int(os.environ.get(i, 1)) - 1 for i in ("COLUMNS", "LINES")]
-            except KeyError:
+                return [int(os.environ[i]) - 1 for i in ("COLUMNS", "LINES")]
+            except KeyError, ValueError:
                 return None, None
 
 


conda/_vendor/tqdm/utils.py
```diff
@@ -365,7 +365,7 @@ def _screen_shape_linux(fp):  # pragma: no cover
             return cols, rows
         except:
             try:
-                return [int(os.environ[i]) - 1 for i in ("COLUMNS", "LINES")]
+                return [int(os.environ.get(i, 1)) - 1 for i in ("COLUMNS", "LINES")]
```
erastogi, line 368: If either of the environment variables are not set (return empty string), `int()` will fail.

dbast:
@anaconda-issue-bot check

conda/_vendor/tqdm/utils.py
chenghlee: This would still fail is `$COLUMNS` or `$LINES` were set to an empty string (e.g., `COLUMNS=`) or some other non-integer value. Perhaps a better approach would be to change the `except KeyError:` below to `except KeyError, ValueError`?

chenghlee (changes requested):

chenghlee:
Proposed PR to fix #10673

conda/_vendor/tqdm/utils.py
erastogi: Good point!
I will fix the PR.

chenghlee: Thanks!

chenghlee (approved):

conda/_vendor/tqdm/utils.py
```diff
@@ -368,6 +368,8 @@ def _screen_shape_linux(fp):  # pragma: no cover
                 return [int(os.environ[i]) - 1 for i in ("COLUMNS", "LINES")]
             except KeyError:
                 return None, None
+            except ValueError:
+                return [0, 0]
```
chenghlee, line 372: Is there a reason for returning `[0, 0]` rather than `(None, None)`? Looking at the [`tqdm.__init__(...)` docstring](https://github.com/conda/conda/blob/f47dc5e6f/conda/_vendor/tqdm/std.py#L858-L863) where `_screen_shape_wrapper(...)` gets called, this would (likely) suppress any progress meter, rather supporting dynamic resizing at a later time.

jezdez:
@chenghlee I think we should file a bug about this in the upstream project as well, or how have we so far handled bugfixes to vendored packages?

jezdez:
@erastogi Thank you for this PR, I've opened a similar PR in the upstream tqdm project and it was merged.

I'm hoping to see a release soon so we can pull in the change from there. I'm closing this PR in the mean time. There is still #10673 to track the progress of your original issue.

conda/_vendor/tqdm/utils.py
erastogi: In the earlier code before the refactoring, `1` was used as the default value. I tried to do the same here.

PR #10688, opened 2021-05-25
explicit install doesn't respect --download-only

jamesp:
Currently an "explicit" install doesn't respect the `--download-only` flag.  

This PR takes the `install.py` logic of exiting with code zero when `--download-only` is set: https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda/cli/install.py#L343-L345

and applies it in `explicit` after the fetch and extract action has occurred.


jamesp:
I've now signed the CLA, should be with you 👍

chenghlee:
@anaconda-issue-bot check

chenghlee (approved):

jezdez (approved):

PR #10689, opened 2021-05-25
Remove the double quotations in env path

Clark934:
An error will block the script if there are double quotations in env path. Just tried it in my windows 10 power shell. This can pass if I removed the double quotations.

┆Issue is synchronized with this [Jira Task](https://anaconda.atlassian.net/browse/CONDAORG2-471)


Clark934:
image
Tested it here.

jezdez:
@anaconda-issue-bot check

PR #10692, opened 2021-05-26
Don't load binstar_client until needed. Part 2.

jezdez:
This is to check if #10659 can be made using importlib.

### DIFF
--- a/conda_env/specs/binstar.py
+++ b/conda_env/specs/binstar.py
@@ -29,10 +29,11 @@ def import_binstar():
         from binstar_client import errors
         import binstar_client.utils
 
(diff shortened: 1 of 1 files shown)


conda_env/specs/binstar.py
```diff
@@ -112,7 +112,7 @@ def package(self):
         if self._package is None:
             try:
                 self._package = self.binstar.package(self.username, self.packagename)
-            except errors.NotFound:
+            except IndexError:
```
jezdez, line 115: `NotFound` is using IndexError as one parent class.

chenghlee (approved):

PR #10727, opened 2021-06-16
Fix /usr/bin/env shebang replacement with options

maresb:
When `/usr/bin/env` is called within a shebang, and options are provided to the executable, it is necessary to add the `-S` parameter. Otherwise the options are interpreted as part of the executable name.

### DIFF
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -179,8 +179,9 @@ def replace_long_shebang(mode, data):
                 # Unescape spaces.
                 executable_name = executable_name.replace(r"\ ", " ")
                 # Note that options contains a leading space.
-                assert (options == b'') or (options.startswith(b' '))
-                new_shebang = '#!/usr/bin/env -S "%s"%s' % (executable_name, options.decode('utf-8'))
+                options = options.decode('utf-8')
+                assert (options == '') or (options.startswith(' '))
+                new_shebang = '#!/usr/bin/env -S "%s"%s' % (executable_name, options)
                 data = data.replace(whole_shebang, new_shebang.encode('utf-8'))
 
     else:
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -174,7 +174,12 @@ def replace_long_shebang(mode, data):
             whole_shebang, executable, options = shebang_match.groups()
             if len(whole_shebang) > 127:
                 executable_name = executable.decode('utf-8').split('/')[-1]
-                assert '"' not in executable_name
+                # Ensure there's no quote or escaping weirdness.
+                assert ('"' not in executable_name) and (r'\\' not in executable_name)
+                # Unescape spaces.
+                executable_name = executable_name.replace(r"\ ", " ")
+                # Note that options contains a leading space.
+                assert (options == b'') or (options.startswith(b' '))
                 new_shebang = '#!/usr/bin/env -S "%s"%s' % (executable_name, options.decode('utf-8'))
                 data = data.replace(whole_shebang, new_shebang.encode('utf-8'))
 
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -174,7 +174,8 @@ def replace_long_shebang(mode, data):
             whole_shebang, executable, options = shebang_match.groups()
             if len(whole_shebang) > 127:
                 executable_name = executable.decode('utf-8').split('/')[-1]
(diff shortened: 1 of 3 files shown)


jezdez:
Can you elaborate on what the purpose of the change is? A way to reproduce this locally would be useful.

maresb:
@jezdez, thanks for asking! I have just realized that I seem to be trying to solve an unsolvable problem.

The unsolvable problem is how to handle arguments in a long shebang which is replaced with an `env` command by `conda.core.portability:replace_long_shebang`


As an example, let's suppose I build a conda package which includes a binary with:

```
#!/path/to/interpreter --some-arg
...
```

If this shebang line is too long after prefix substitution, it currently gets replaced by the following:
```
#!/usr/bin/env interpreter --some-arg
```

Unfortunately, [the behavior of this shebang is unspecified](https://stackoverflow.com/a/4304187). I was getting the error message

```
/usr/bin/env: ‘interpreter --some-arg’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```

This error arises because `env` receives the single argument `interpreter --some-arg` which is not an existing executable.

My patch here is an attempt to fix this with the solution suggested by the error message. Namely,
```
#!/usr/bin/env -S "interpreter" --some-arg
```

Here, the string `-S "interpreter" --some-arg` is passed as the single argument to `env`, but the `-S` flag indicates to `env` that this string should be parsed for spaces, and `env` correctly calls `interpreter` with the argument `--some-arg`.

What's really nasty here is that this `-S` parameter was apparently [introduced in coreutils 8.30](https://stackoverflow.com/a/52979955), released [in July 2018](https://github.com/coreutils/coreutils/releases). I only realized this just now. :cry:

In summary, the way `conda` currently adds arguments to any shortened shebang, it renders that shebang corrupt on any operating system [of type 3](https://stackoverflow.com/a/4304187).

My patch here fixes systems of type 3 with coreutils >=8.30, but would cause a regression on systems of type 2 with coreutils <8.30. I imagine that such a regression is unacceptable.

I'm now unfortunately at a loss for how to proceed. Did I articulate the issue clearly? If so, any suggestions?

jezdez:
Interesting find @maresb!

I would suggest to research if there are portable ways to figure out which coreutils are installed to do the rewriting conditionally?

OTOH there is a risk of breakage if users would upgrade coreutils.

Alternatively, we could write a conda specific "env" alternative that better handles the use case and decides during runtime which option to use?

maresb:
Hi @jezdez, thanks so much for thoroughly reading and understanding my explanation.

> Alternatively, we could write a conda specific "env" alternative that better handles the use case and decides during runtime which option to use?

This is a really nice idea, but I see a problem. A leading cause of having to rewrite the shebang with `/usr/bin/env` is when the prefix name is very long. If we rewrote `env`, then we would have to place the new `env` under the same prefix, and we haven't gained anything.

> I would suggest to research if there are portable ways to figure out which coreutils are installed...

We could probably parse this:

```
$ env --version
env (GNU coreutils) 8.30
Copyright (C) 2018 Free Software Foundation, Inc.
...
```

> OTOH there is a risk of breakage if users would upgrade coreutils.

Actually, things only get better with upgrading. The risk would be that the user downgrades coreutils.

> ...to do the rewriting conditionally?

I think there are four cases.

Recall that Type 2 means that the OS automatically processes whitespace as an argument delimiter in a shebang, and Type 3 means it does not. We should be able to distinguish between types 2 and 3 with a test.

The cases are

1. Type 2, old coreutils
    * Current behavior works, thanks to "Type 2"
    * This PR would break things due to old coreutils.
2. Type 2, new coreutils
    * Current behavior works, thanks to "Type 2"
    * This PR would also work.
3. Type 3, old coreutils
    * Current behavior is broken due to Type 3.
    * This PR would also be broken due to old coreutils.
4. Type 3, new coreutils
    * Current behavior is broken due to Type 3.
    * This PR would fix things.

So I think the correct behavior for each case would look something like

1. Use current behavior.
2. Use current behavior or use the solution of this PR.
3. Throw an error saying that env ≥ 8.30 is required for commands with arguments when the prefix name is too long.
4. Use the solution of this PR.

Unfortunately this issue has exceeded the time I have allocated for it, and I can't really pursue it any further. Shall I close this PR, open a corresponding issue, copy all of this correspondence into that new issue, and hope that someone else can fix it? (At least the issue would be documented.)

PR #10771, opened 2021-07-05
The RE used to split tokens is too narrow

asterbini:
I noticed that when using conda WITH tokens sometimes the token part is not recognized if it contains also the char '_' (underscore). 
This patch enlarge regular expression to split the token to all chars different from '/'

### DIFF
--- a/conda/common/url.py
+++ b/conda/common/url.py
@@ -268,7 +268,7 @@ def split_anaconda_token(url):
         >>> split_anaconda_token("https://10.2.3.4:8080/conda/t/tk-123-45")
         (u'https://10.2.3.4:8080/conda', u'tk-123-45')
     """
-    _token_match = re.search(r'/t/([a-zA-Z0-9-]*)', url)
+    _token_match = re.search(r'/t/([^/]*)', url)
     token = _token_match.groups()[0] if _token_match else None
     cleaned_url = url.replace('/t/' + token, '', 1) if token is not None else url
     return cleaned_url.rstrip('/'), token


jezdez:
@anaconda-issue-bot check

wagnerpeer:
I doubt that the regex `[^/]` is fully correct, as it would allow URLs/token to contain `?` and other special characters.
I haven't found a reference to the parts where token are generated, but I assume that token will always be alphanumeric with dividers of blocks like `-` and `_`. Hence, extending the regex to include `_` seems like the better option to me.

wagnerpeer (changes requested):
What about similar changes to [conda/conda/gateways/logging.py#L32](https://github.com/conda/conda/blob/248741a843e8ce9283fa94e6e4ec9c2fafeb76fd/conda/gateways/logging.py#L32)?

asterbini:
> What about similar changes to [conda/conda/gateways/logging.py#L32](https://github.com/conda/conda/blob/248741a843e8ce9283fa94e6e4ec9c2fafeb76fd/conda/gateways/logging.py#L32)?

Yes, it makes sense. I did not notice that RE, I just was fixing mamba/conda .

wolfv:
@wagnerpeer the regex you linked has teh same prolbme for the token matching though (`r'/t/[a-z0-9A-Z-]+/'  # token`) as in not matching a token with an underscore.

The problem here is that the first two letters of a token generated by anaconda.org are the first two letters of the username.

Apparently anaconda.org allows underscores in usernames.

So when your username is `x_blabla` your tokens will look like `x_-uuid4123-1231-1231-1321-aaaaaa` which breaks these regex's.

wagnerpeer:
> @wagnerpeer the regex you linked has teh same prolbme for the token matching though (`r'/t/[a-z0-9A-Z-]+/' # token`) as in not matching a token with an underscore.

Correct, that was the intention. Making aware of another location with the same issues.

> The problem here is that the first two letters of a token generated by anaconda.org are the first two letters of the username.
> 
> Apparently anaconda.org allows underscores in usernames.
> 
> So when your username is `x_blabla` your tokens will look like `x_-uuid4123-1231-1231-1321-aaaaaa` which breaks these regex's.

That explains the issue, but can you give a reference (documentation) on how a token is defined? Where did you find, that the first two characters are copied from the username?
If the statement holds true, then a solution could look like `r'/t/.{2}[a-z0-9A-Z-]+/' # token`?

asterbini:
> That explains the issue, but can you give a reference (documentation) on how a token is defined? Where did you find, that the first two characters are copied from the username?
> If the statement holds true, then a solution could look like `r'/t/.{2}[a-z0-9A-Z-]+/' # token`?

I just observed that my username on anaconda starts with 'a_' and the token start with the same two chars

PR #10795, opened 2021-07-16
Add creation of the conda-ci-linux-64 container

dbast:
The logic moved from https://github.com/ContinuumIO/docker-images

While the Dockerfile is untouched, the github action was updated to use
a build matrix, to publish to ghcr.io and to specify the Python version in
the image tag instead of the image name.

### DIFF
--- a/.github/workflows/conda_ci_container_cicd.yml
+++ b/.github/workflows/conda_ci_container_cicd.yml
@@ -1,0 +1,46 @@
+name: Build and Publish Conda CI Images
+on:
+  schedule:
+    - cron:  '0 0 * * 0'
+  pull_request:
+    paths:
+      - 'ci/Dockerfile'
+      - '.github/workflows/conda_ci_container_cicd.yml'
+  push:
+    paths:
+      - 'ci/Dockerfile'
+      - '.github/workflows/conda_ci_container_cicd.yml'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [3.7, 3.8, 3.9]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        if: github.ref == 'refs/heads/master'
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+          driver-opts: network=host
+
+      - name: build-push conda-ci-linux-64-python-${{ matrix.python }}
+        uses: docker/build-push-action@v2
+        with:
+          context: ./ci
+          builder: ${{ steps.buildx.outputs.name }}
+          file: ./ci/Dockerfile
+          build-args: PYTHON_VERSION=${{ matrix.python }}
+          tags: conda/conda-ci-linux-64:python${{ matrix.python }}-latest
+          push: ${{ github.ref == 'refs/heads/master' }}
+
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,0 +1,42 @@
+FROM debian:buster-slim
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+ENV PATH /opt/conda/bin:$PATH
+
+# hadolint ignore=DL3008
+RUN apt-get update --fix-missing && \
+    apt-get install -y --no-install-recommends wget build-essential bzip2 ca-certificates libglib2.0-0 libxext6 libsm6 libxrender1 git mercurial subversion sudo && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    useradd -m -s /bin/bash test_user && \
+    usermod -u 1001 test_user && \
+    groupmod -g 1001 test_user && \
+    echo "test_user ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh && \
+    /opt/conda/bin/conda clean -tipsy
+
+ARG PYTHON_VERSION=3.8
+
+# conda and test dependencies
+RUN /opt/conda/bin/conda install --update-all -y -c defaults \
+    conda conda-package-handling \
+    python=$PYTHON_VERSION pip pycosat requests ruamel_yaml cytoolz \
+    anaconda-client nbformat make \
+    pytest pytest-cov codecov radon pytest-timeout mock responses pexpect \
+    flake8 && \
+    /opt/conda/bin/conda clean --all --yes
+
+# conda-build and test dependencies
+RUN /opt/conda/bin/conda install --update-all -y -c defaults \
+        conda-build patch git \
+        perl pytest-xdist pytest-mock \
+        anaconda-client \
+        filelock jinja2 conda-verify pkginfo \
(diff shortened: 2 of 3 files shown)


dbast:
After merging another PR is required to update https://github.com/conda/conda/blob/master/.github/workflows/ci.yml to use the images from ghcr.io.

.github/workflows/conda_ci_container_cicd.yml
```diff
@@ -0,0 +1,46 @@
+name: Build and Publish Conda CI Images
+on:
+  schedule:
+    - cron:  '0 0 * * 0'
+  pull_request:
+    paths:
+      - 'ci/Dockerfile'
+      - '.github/workflows/conda_ci_container_cicd.yml'
```
jezdez, line 8: Could we use a shorter name for this file please, like `ci_image.yml`?

.github/workflows/conda_ci_container_cicd.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [3.7, 3.8, 3.9]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to DockerHub
```
jezdez, line 23:
```suggestion
      - name: Login to GitHub Container Registry
```

jezdez: I was wondering if this separate section is needed if the `docker/build-push-action` action also supports auth parameters out of the box? Or is this because of using buildx?

.github/workflows/conda_ci_container_cicd.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: build-push conda-ci-linux-64-python-${{ matrix.python }}
+        uses: docker/build-push-action@v2
+        with:
+          context: ./ci
+          builder: ${{ steps.buildx.outputs.name }}
+          file: ./ci/Dockerfile
+          build-args: PYTHON_VERSION=${{ matrix.python }}
+          tags: conda/conda-ci-linux-64:python${{ matrix.python }}-latest
+          push: ${{ github.ref == 'refs/heads/master' }}
```
jezdez, line 45: The login step above has the same condition but the buildx step doesn't. Is this intentional?

Could we move this condition to the whole job so we don't even build the image?

ci/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM debian:buster-slim
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+ENV PATH /opt/conda/bin:$PATH
+
+# hadolint ignore=DL3008
+RUN apt-get update --fix-missing && \
+    apt-get install -y --no-install-recommends wget build-essential bzip2 ca-certificates libglib2.0-0 libxext6 libsm6 libxrender1 git mercurial subversion sudo && \
```
jezdez, line 8: Those dependencies, are they all needed for the CI image? E.g. hg and svn seem to be pretty random. What's libxext6, libsm6 and libxrender1 for?

ci/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    useradd -m -s /bin/bash test_user && \
+    usermod -u 1001 test_user && \
+    groupmod -g 1001 test_user && \
+    echo "test_user ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh && \
+    /opt/conda/bin/conda clean -tipsy
```
jezdez, line 18: `-tipsy`? This must be an in-joke or something, but we should write it to be easier to understand.

For some reason I can't find a reference for the `-s` option in that clean subcommand call.

```suggestion
    /opt/conda/bin/conda clean --tarballs --index-cache --packages --yes
```
OR:
```suggestion
    /opt/conda/bin/conda clean --all
```

ci/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM debian:buster-slim
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+ENV PATH /opt/conda/bin:$PATH
+
+# hadolint ignore=DL3008
```
jezdez, line 6: What's that?

ci/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh && \
+    /opt/conda/bin/conda clean -tipsy
+
+ARG PYTHON_VERSION=3.8
+
+# conda and test dependencies
+RUN /opt/conda/bin/conda install --update-all -y -c defaults \
+    conda conda-package-handling \
```
jezdez, line 24: It would be great to sync those dependencies (also conda-build deps) with the `dev/tests-requirements.txt` for a better dev/ci parity. We should have the exact same environment in CI and development.

ci/Dockerfile
jezdez:
```suggestion
    apt-get install -y --no-install-recommends tini wget build-essential bzip2 ca-certificates libglib2.0-0 libxext6 libsm6 libxrender1 git mercurial subversion sudo && \
```

ci/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+        conda-build patch git \
+        perl pytest-xdist pytest-mock \
+        anaconda-client \
+        filelock jinja2 conda-verify pkginfo \
+        glob2 beautifulsoup4 chardet pycrypto \
+    && /opt/conda/bin/conda clean --all --yes
+
+USER test_user
+
+ENTRYPOINT [ "/bin/bash" ]
```
jezdez, line 42: This should probably wrap the container in a init process for local development and better process handling

```suggestion
ENTRYPOINT ["/usr/bin/tini", "--"]
CMD ["/bin/bash"]
```

jezdez (changes requested):
Thanks for opening this @dbast! I'm sorry that this is more than I thought, but I think it's worth reviewing this properly to not have to touch it again for a while :)

.github/workflows/conda_ci_container_cicd.yml
dbast: Login only works if credentials are available... they typically are restricted to master... but the workflow can still run on other branches without login/push to test building the images if there are updates.

ci/Dockerfile
dbast: A Dockerfile linker https://github.com/hadolint/hadolint

jezdez: Ah, neat, we should add that.

.github/workflows/conda_ci_container_cicd.yml
jezdez: Got it, thanks.

.github/workflows/conda_ci_container_cicd.yml
dbast: ah. that can be maybe simplified... also we don't need buildx... lets see

ci/Dockerfile
dbast: not sure if that was the original intention

ci/Dockerfile
dbast: the dependencies are replicated from the miniconda image ... I assume that some tests install conda packages that depend on those... the libs are the client libs regarding using X11 functionality used by conda packages.

ci/Dockerfile
jezdez: Yeah, I'm happy to do this in a separate PR TBH, this is an added benefit for the sake of developer experience that wasn't possible before.

ci/Dockerfile
dbast: added :)

jezdez (approved):

Issue #10815, opened 2021-07-28
Unable to install dependencies from requirements.txt as part of an environment.yml


drorata:
## Current Behavior

In my `environment.yml` I install some of the dependencies from a `requirements.txt` using `pip`. 
This part of the dependencies resolving fails with a `CondaEnvException` exception. See below more details.

### Steps to Reproduce

Consider the following `environment.yml`:

```yml
name: my_env
channels:
  - conda-forge
dependencies:
  - python=3.9.5
  - jupyter
  - nb_conda
  - pip
  - jupyter_contrib_nbextensions
  - pip
  - pip:
      - -r file:requirements.txt
```

and the `requirements.txt`:

```
pandas~=1.0
yfinance~=0.1
pandas_datareader~=0.8
pytest<7.0
loguru<0.6
beautifulsoup4~=4.9
python-dotenv<0.20
click<9.0
sendgrid<7.0
```

Running either `conda env create -f environment.yml` or `conda env update -f environment.yml` fails with the following error:

```
Installing pip dependencies: | Ran pip subprocess with arguments:
['/Users/username/anaconda3/envs/stocks-opportunities-monitor/bin/python', '-m', 'pip', 'install', '-U', '-r', '/Users/username/Dropbox/dev/stocks-opportunities-monitor/condaenv.812ekr2q.requirements.txt']
Pip subprocess output:

Pip subprocess error:
ERROR: Exception:
Traceback (most recent call last):
  File "/Users/username/anaconda3/envs/stocks-opportunities-monitor/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 173, in _main
    status = self.run(options, args)
  File "/Users/username/anaconda3/envs/stocks-opportunities-monitor/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 203, in wrapper
    return func(self, options, args)
  File "/Users/username/anaconda3/envs/stocks-opportunities-monitor/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 286, in run
    reqs = self.get_requirements(args, options, finder, session)
  File "/Users/username/anaconda3/envs/stocks-opportunities-monitor/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 384, in get_requirements
    for parsed_req in parse_requirements(
  File "/Users/username/anaconda3/envs/stocks-opportunities-monitor/lib/python3.9/site-packages/pip/_internal/req/req_file.py", line 135, in parse_requirements
[...]

PR #10820, opened 2021-07-30
Update Advanced Pip Example

mfansler:
## Overview
The Advanced Pip Example previously used an invalid `file` URI syntax. With Pip version 21.2.1, this syntax is no longer supported and results in `FileNotFoundError`. This PR updates the example to use a valid syntax.

## Related Issues
#10815 
https://github.com/pypa/pip/issues/10237
https://github.com/conda-forge/pip-feedstock/issues/76


jezdez (approved):
Thanks, this makes sense.

PR #10852, opened 2021-08-11
Refactor docs logic

dbast:
* Move to own workflow to trigger only on docs changes
* Remove git docs detect magic from ./ci/*_tests.sh that was used previously to avoid test runs on docs only changes.
* Move ci/build_docs.sh logic to docs/Makefile
* Use dedicated env for doc builds to avoid base env polution on local runs
* Use docs/Makefile targets during docs.yml flow run
* Convert Python version in .readthedocs.yml to string according to https://raw.githubusercontent.com/readthedocs/readthedocs.org/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.json

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,26 @@
+name: CI docs
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+     - 'docs/**'
+  pull_request:
+    paths:
+     - 'docs/**'
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: build the docs
+        run : |
+          source "${CONDA}/etc/profile.d/conda.sh"
+          make env-docs
+          conda activate conda-docs
+          cd docs
+          make html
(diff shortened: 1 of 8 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,27 @@
+name: CI tests
```
jezdez, line 1:
```suggestion
name: Docs
```

docs/Makefile
```diff
@@ -12,7 +12,11 @@ BUILDDIR      = _build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+environment:
+	conda create --name conda-docs --channel defaults python=3.8 --yes
+	conda run --name conda-docs pip install -r requirements.txt
```
jezdez, line 17:
```suggestion
	conda run --name conda-docs pip install -r docs/requirements.txt
```

dbast: hm. this is the Makefile in the docs folder ... prefixing with docs/ wont work here.

jezdez: Wait, this isn't right, I missed that this is in the docs directory. Could we instead use the parent directory Makefile?

dbast: we could also move the target to the Makefile in the repo root and call the target env-docs instead.

dbast: ok. wil do.

PR #10857, opened 2021-08-12
[WIP] Use pytest-test-groups for more parallelisation

dbast:


dbast:
wow. nice group run times ... the failures need more investigations ... do some tests depend on the test order, which they should not?

PR #10862, opened 2021-08-16
Cleanup Makefile

dbast:
* Always print to be executed commands for easier understanding while calling targets (removing the "@" prefix).
* Specify the shell to be bash with error handling so that errors in multi-line bash code with pipes are not masked.
* Setting all targets to be phony (producing no out file/folder with the same name for caching)
* Remove pytest args that are anyway not used anymore or set differently (the github workflows are not setting those and some are set via setup.cfg)
* Make clean-all more generic
* Don't define variables that only work inside test environments, as the Makefile is also used outside to e.g. create the docs or lint env. Fixes the message `/bin/sh: 0: Illegal option -` shown 3x on CI for each make call.


dbast:
failures should be un-related.

Issue #9862, opened 2020-04-20
unsatisfiable_hints affects whether an environment can be solved


rolweber:
Hello Team Anaconda,

I ran into a strange case where setting `unsatisfiable_hints` to `False` makes an installation step unsolvable, whereas it succeeds with the default of `True`. I narrowed it down to a combination of:
- the setting of `unsatisfiable_hints`
- a pin for the `conda` minor version
- the list of packages to be updated

The following Dockerfile reproduces the problem, with the current Miniconda installer 4.8.2.
See the last RUN statement for the problematic scenario.

```Dockerfile
FROM registry.access.redhat.com/ubi8/ubi-minimal

RUN /usr/bin/microdnf install \
    bzip2 \
    tar \
 && microdnf clean all \
 && rm -rf /var/cache/yum

ENV CONDA_HOME=/opt/conda \
    PATH=/opt/conda/bin:$PATH \
    CONDA=Miniconda3-py37_4.8.2-Linux-x86_64.sh \
    SHA256=957d2f0f0701c3d1335e3b39f235d197837ad69a944fa6f5d8ad2c686b69df3b

RUN time curl -sS --progress-bar --fail -o /tmp/$CONDA \
         https://repo.anaconda.com/miniconda/$CONDA \
 && echo >/tmp/conda.sha256sum.txt "$SHA256 */tmp/$CONDA" \
 && sha256sum -c /tmp/conda.sha256sum.txt \
 && /bin/bash /tmp/$CONDA -b -p "${CONDA_HOME}" \
 && rm -f /tmp/$CONDA


RUN conda config --system --set unsatisfiable_hints False \
 && echo >>"${CONDA_HOME}/conda-meta/pinned" "conda=4.8" \
 && conda install -n base conda=4.8.3 openssl=1.1.1f
```

Any one of these changes will let the installation succeed:
- keep `unsatisfiable_hints` at the default, or set it to `True`
- don't pin the `conda` minor version
- install only one of the packages
  - The problem remains if you replace `openssl=1.1.1f` with `certifi=2020.4.5.1`

This is not a blocker for me. Now that I've identified the circumstances, I can work around the problem.

rolweber:
I'm running into this problem again in a custom environment, without pinned packages. This is more serious for me.

With unsatisfiable_hints as True, an installation succeeds:
`conda install --dry-run pytest`
```txt
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

[...]

PR #10867, opened 2021-08-17
Fix `unsatisfiable_hints=false` bug (misreported conflicts)

jaimergp:
Comes from https://github.com/conda/conda/issues/9862

I don't know what's wrong but at least I found a way to reproduce in the unit tests. I'll edit as a find more details.

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -565,7 +565,7 @@ class Resolve(object):
 
         strict_channel_priority = context.channel_priority == ChannelPriority.STRICT
 
-        cache_key = strict_channel_priority, tuple(explicit_specs)
+        cache_key = strict_channel_priority, tuple(sorted(explicit_specs, key=str))
         if cache_key in self._reduced_index_cache:
             return self._reduced_index_cache[cache_key]
 
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -1260,8 +1260,7 @@ class Resolve(object):
         log.debug("Solve: Getting reduced index of compliant packages")
         len0 = len(specs)
 
-        reduced_index = self.get_reduced_index(
-            specs, exit_on_conflict=not context.unsatisfiable_hints)
+        reduced_index = self.get_reduced_index(specs)
         if not reduced_index:
             # something is intrinsically unsatisfiable - either not found or
             # not the right version
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -1260,8 +1260,8 @@ class Resolve(object):
         log.debug("Solve: Getting reduced index of compliant packages")
         len0 = len(specs)
 
-        reduced_index = self.get_reduced_index(specs,
-            exit_on_conflict=not context.unsatisfiable_hints)
+        reduced_index = self.get_reduced_index(
+            specs, exit_on_conflict=not context.unsatisfiable_hints)
         if not reduced_index:
             # something is intrinsically unsatisfiable - either not found or
             # not the right version
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -1260,7 +1260,8 @@ class Resolve(object):
         log.debug("Solve: Getting reduced index of compliant packages")
         len0 = len(specs)
 
-        reduced_index = self.get_reduced_index(specs)
+        reduced_index = self.get_reduced_index(specs,
+            exit_on_conflict=not context.unsatisfiable_hints)
         if not reduced_index:
             # something is intrinsically unsatisfiable - either not found or
             # not the right version
--- a/tests/test_resolve.py
+++ b/tests/test_resolve.py
@@ -13,7 +13,7 @@ import pytest
 
 from conda.base.context import context, conda_tests_ctxt_mgmt_def_pol
 from conda.common.compat import iteritems, itervalues
-from conda.common.io import env_var
+from conda.common.io import env_var, env_vars
 from conda.exceptions import UnsatisfiableError
 from conda.gateways.disk.read import read_python_record
 from conda.models.channel import Channel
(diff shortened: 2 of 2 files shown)


jaimergp:
Mmmm... my local tests say that it works _if_ you enable trace logging (`-vvv`) but it doesn't if you enable debug logging only (`-vv`). This is going to be interesting 🤣

jaimergp:
Update, it's not due to logging levels. The error is just not deterministic. Attaching a traced log of both a successful and failed attempts.

* [`unsat_log_trace_ok.txt`](https://github.com/conda/conda/files/7006695/unsat_log_trace_ok.txt)
* [`unsat_log_trace_failed.txt`](https://github.com/conda/conda/files/7006696/unsat_log_trace.txt)

jaimergp:
The diff of both logs reveal that the error is happening somewhere around here:

https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda/resolve.py#L1263-L1289

I guess `get_reduced_index` is not deterministic. Specifically this section:

https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda/resolve.py#L674-L686

Let's see if I can add more traces there!

jaimergp:
Yes, specifically, these lines:

```
 if pruned_to_zero and exit_on_conflict: 
     return {} 
```

implement a logic that is faulty. `pruned_to_zero` can end up storing perfectly valid package records. See the logs of this simple test. We are trying to install `a` and `b`, which both depend on `c`. No conflicts should be reported.

```python
from collections import OrderedDict
from conda.gateways.logging import initialize_logging, WARN, TRACE
initialize_logging(WARN)

from conda.models.records import PackageRecord
from conda.resolve import MatchSpec, Resolve, ResolvePackageNotFound

def simple_rec(name='a', version='1.0', depends=None, build='0',
               build_number=0, channel='channel-1'):
    if depends is None:
        depends = []
    return PackageRecord(**{
        'name': name,
        'version': version,
        'depends': depends,
        'build': build,
        'build_number': build_number,
        'channel': channel,
    })

index = (
    simple_rec(name='a', version='1.0', depends=['c >=1,<2']),
    simple_rec(name='b', version='1.0', depends=['c']),
    simple_rec(name='c', version='1.0'),
    simple_rec(name='c', version='2.0'),
)

r = Resolve(OrderedDict((prec, prec) for prec in index))
r.install(['a', 'b '])
```

Yet, as we process the package list, both `a` and `b` are labeled as zero-pruned, populating the `pruned_to_zero` set and hence enabling that early exit with an empty dictionary.

```
START (MatchSpec("b"), MatchSpec("a"))
-- ASSESSING b
---- REDUCED False
---- REDUCED False
---- PRUNED b
-- ASSESSING a
---- REDUCED True
---- REDUCED True
---- FILTERED a
-- ASSESSING a
---- REDUCED False
---- PRUNED a
START (MatchSpec("b"), MatchSpec("a"))
-- ASSESSING b
---- REDUCED False
---- REDUCED False
---- PRUNED b
-- ASSESSING a
---- REDUCED False
---- REDUCED False
---- PRUNED a
```

If I comment out that early exit (which is the same path followed when `unsatisfiable_hints` is `true`), the environment ends up being solved perfectly, as it should.

So I guess the issue comes from that pruning algorithm implemented in `filter_group`. `git blame` [shows](https://github.com/conda/conda/blame/33a142c16530fcdada6c377486f1c1a385738a96/conda/resolve.py#L674-L686) that a lot of people have touched this part of the code so no clue who to tag here...

Anyway, for now the fix is simple: do not early exit. However, somebody should look into that logic.

jaimergp:
Huh, this makes one unit test [fail](https://github.com/conda/conda/pull/10867/checks?check_run_id=3361310455#step:3:790).

How is that the expected behaviour? If the env is solvable, it shouldn't depend on the hints?

```python
def test_fast_error_on_unsat():
    installed = r.install(["zope.interface=4.1.1"])
    result = [rec.dist_str() for rec in installed]

    assert result == add_subdir_to_iter([
        'channel-1::nose-1.3.0-py33_0',
        'channel-1::openssl-1.0.1c-0',
        'channel-1::python-3.3.2-0',
        'channel-1::readline-6.2-0',
        'channel-1::sqlite-3.7.13-0',
        'channel-1::system-5.8-1',
        'channel-1::tk-8.5.13-0',
        'channel-1::zlib-1.2.7-0',
        "channel-1::zope.interface-4.1.1.1-py33_0",
    ])

    _installed = r.install(['python 2.7*'], installed=installed)
    result = [rec.dist_str() for rec in _installed]
    assert result == add_subdir_to_iter([
        'channel-1::nose-1.3.0-py27_0',
        'channel-1::openssl-1.0.1c-0',
        'channel-1::python-2.7.5-0',
        'channel-1::readline-6.2-0',
        'channel-1::sqlite-3.7.13-0',
        'channel-1::system-5.8-1',
        'channel-1::tk-8.5.13-0',
        'channel-1::zlib-1.2.7-0',
        'channel-1::zope.interface-4.0.5-py27_0',
    ])

    r._reduced_index_cache.clear()
    with env_var("CONDA_UNSATISFIABLE_HINTS", "False", stack_callback=conda_tests_ctxt_mgmt_def_pol):
        with pytest.raises(UnsatisfiableError):
            _installed = r.install(["python 2.7*"], installed=installed)
            # Test fails his here  
```

jaimergp:
Maybe my initial tests were wrong, so I am following the existing style now.

I am puzzled by the cache clearing stuff. If a `satisfiable_hints=true` attempt is done beforehand and we keep the cache (`get_index_r_1` are memoized), it works. If the cache is cleared, it doesn't, since it cannot exit through this early path here:

https://github.com/conda/conda/blob/88830971853188a812b049a6d97daad069bd97bd/conda/resolve.py#L570

This does not show up in a real CLI example. Both options work reliably after f5e1965 :

```
# Local machine
% cd ~/devel/conda
% docker run -it --rm -v ${PWD}:/conda ghcr.io/conda/conda:master-python3.9 bash
# now in Docker
% eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
% CONDA_UNSATISFIABLE_HINTS=true conda create -n hints-test --dry-run pytest python
% CONDA_UNSATISFIABLE_HINTS=false conda create -n hints-test --dry-run pytest python
```

If I revert the tuple-sorting change:

```
% CONDA_UNSATISFIABLE_HINTS=false conda create -n hints-test --dry-run pytest python
Solving environment: failed

UnsatisfiableError: 
Did not find conflicting dependencies. If you would like to know which
packages conflict ensure that you have enabled unsatisfiable hints.

conda config --set unsatisfiable_hints True
```

How can I reproduce this behaviour with unit tests? I've tried all the test indices (`get_index_r_1` et al) and none reproduce it :/

jaimergp:
I'll also note that I tried this patch on [this other PR](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/164) and it [didn't work](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=365149&view=logs&j=d1321064-f6c3-56b7-0172-6d994d01c836&t=39022612-2533-571e-7b41-94814e7d38bb&l=68) 😬

PR #10881, opened 2021-08-24
(WIP) Add libmamba solver backend

jaimergp:
> This is the main view of the libsolv proof of concept integration. Actual work should happen in smaller PRs targeting this feature branch.

## Status

* 2021.11.30, e90caa8 passes all tests in the conda suite, with the exception of known bugs and unsupported features. These have been annotated with `xfail` for reference.
* 2021.12.03, 1bb3ed9 re-enables unicode tests on Windows and makes sure mamba 0.19 works. We only depend on `libmambapy` now but we are still installing `mamba` fully to aid in development / debugging.
* For the rest of the month, we will clean up the conda classic logic, come up with a plugin architecture to install this separately and abstract away some common details across solvers.

__Note__

> History of descriptions in this PR is available in the "edited" dropdown.


jaimergp:
Ok, I have tried this locally and the following commands are technically functional:

```
CONDA_SOLVER_LOGIC=libsolv conda install anaconda --dry-run
CONDA_SOLVER_LOGIC=libsolv conda install remove pytest --dry-run
CONDA_SOLVER_LOGIC=libsolv conda update --all
```

The next step are, I guess:

* Review this implementation to see if it matches our expectations.
* Enable the integration tests for this logic and see what we find.
* Create smaller PRs to tackle whatever we find!

conda/core/solve.py
```diff
@@ -1029,6 +1039,250 @@ def _prepare(self, prepared_specs):
...
+            self._setup_state()
+            # 2. Create solver and needed flags, tasks and jobs
+            self._configure_solver()
+            # 3. Run the SAT solver
+            self._run_solver()
+            # 5. Export back to conda
+            return self._build_transaction()
+        finally:
+            # 6. Clean
+            self._state.clear()
```
FFY00, line 1083: Using `self._state` as an internal ephemeral argument is kind of anti-pattern. Wouldn't this be more idiomatic by using the state returned by `_setup_state` and passing it to the other functions?

conda/core/solve.py
```diff
@@ -1029,6 +1039,250 @@ def _prepare(self, prepared_specs):
...
+        super().__init__(prefix, channels, subdirs=subdirs, specs_to_add=specs_to_add,
+                         specs_to_remove=specs_to_remove, repodata_fn=repodata_fn,
+                         command=command)
+        self._state = {}
+
+    def solve_for_transaction(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL,
+                              ignore_pinned=NULL, force_remove=NULL, force_reinstall=NULL,
+                              should_retry_solve=False):
+        # Logic heavily based on Mamba's implementation:
+        # https://github.com/mamba-org/mamba/blob/master/mamba/mamba.py#L289 (solver parts)
```
jezdez, line 1068: Let's have a commit specific link here.

conda/core/solve.py
```diff
@@ -1029,6 +1039,250 @@ def _prepare(self, prepared_specs):
...
+        from mamba.utils import load_channels, get_installed_jsonfile, init_api_context
+        from mamba.mamba_api import Pool, Repo, PrefixData
+
+        init_api_context()
+
+        pool = Pool()
+
+        # TODO: Check if this update-related logic is needed here too
+        # Maybe conda already handles that beforehand
+        # https://github.com/mamba-org/mamba/blob/master/mamba/mamba.py#L426-L485
```
jezdez, line 1095: Same here

conda/core/solve.py
```diff
@@ -1029,6 +1039,250 @@ def _prepare(self, prepared_specs):
...
+            "installed_pkgs": installed_pkgs,
+        })
+
+        return self._state
+
+    def _configure_solver(self):
+        if self.specs_to_add or context.update_modifier == UpdateModifier.UPDATE_ALL:
+            return self._configure_solver_for_install()
+        elif self.specs_to_remove:
+            return self._configure_solver_for_remove()
```
FFY00, line 1125: Can we ever have both? The legacy solver seemed to be able to handle that from looking at the API, but I might be wrong.
Still, it would probably be a good idea to error out when we get the two actions, as we do not actually support them. Currently, if resolving for install, `self.specs_to_remove` will be silently ignored, which I imagine could lead to nightmare debugging :sweat_smile:

tests/core/test_solve.py
```diff
@@ -48,7 +48,7 @@ def get_solver(tmpdir, specs_to_add=(), specs_to_remove=(), prefix_records=(), h
     spec_map = {spec.name: spec for spec in history_specs}
     get_index_r_1(context.subdir)
     with patch.object(History, 'get_requested_specs_map', return_value=spec_map):
-        solver = Solver(tmpdir, (Channel('channel-1'),), (context.subdir,),
+        solver = _get_solver_logic()(tmpdir, (Channel('channel-1'),), (context.subdir,),
```
FFY00, line 51: Can we make this a fixture? It would also make it easier to parameterize the test.

FFY00 (commented):
Looks great! There's still testing to be done to uncover quirks this might not handle properly, but I am very happy with the progress.

I left some comments, let me know what you think. I can fix them for you if you want :blush:

conda/core/solve.py
jaimergp: I was mimicking their `SolverStateContainer` class behaviour (not a fan either), but without knowing which things we will have to move around. Now that we have a more or less clear idea of what things are being moved from A to B we can switch to explicit returns and signatures!

conda/core/solve.py
jaimergp: Not sure about this. Mamba does not allow both operations, I think. The API supports that because some solves might involve installing/updating stuff _and_ removing existing packages. However, the user will only be able to say "add this" or "remove this", but never "remove this and add this" at the same time. 

Anaconda should confirm anyway because I am not sure.

And yes, please check for that silent ignore, I forgot about it assuming this is coming straight from the CLI, but who knows how the Python API will be used.

tests/core/test_solve.py
jaimergp: Yes, agreed!

jaimergp:
Thanks for the comments @FFY00! Pretty good feedback, spot-on!

> I can fix them for you if you want

Please do :)

conda/core/solve.py
```diff
@@ -1029,6 +1039,251 @@ def _prepare(self, prepared_specs):
...
+        )
+        specs = [s.conda_build_form() for s in self.specs_to_remove]
+        solver.add_jobs(specs, api.SOLVER_ERASE | api.SOLVER_CLEANDEPS)
+
+        state["solver"] = solver
+        return solver
+
+    def _collect_specs_to_add(self, state):
+        """
+        Reimplement the logic found in super()._add_specs(), but simplified.
```
jezdez, line 1204: If this is a simplification, let's see if we can refactor _add_specs to simplify as well.

conda/core/solve.py
```diff
@@ -1029,6 +1039,251 @@ def _prepare(self, prepared_specs):
...
+        state = self._setup_state()
+        # 2. Create solver and needed flags, tasks and jobs
+        self._configure_solver(state)
+        # 3. Run the SAT solver
+        self._run_solver(state)
+        # 5. Export back to conda
+        return self._build_transaction(state)
+
+    def _setup_state(self):
+        from mamba.utils import load_channels, get_installed_jsonfile, init_api_context
```
jezdez, line 1081: We need to figure out a strategy for this, whether we can actually rely on runtime dependencies on Mamba since it could make maintenance of Conda much harder if API is used that may be less stable than Conda's API.

conda/core/solve.py
```diff
@@ -1029,6 +1039,251 @@ def _prepare(self, prepared_specs):
...
+                if not pkg.name.startswith("__"):  # skip virtual packages
+                    specs_to_add.append(MatchSpec(pkg.name))
+        # C. Implicitly requested via `install|update --update-deps`
+        elif context.update_modifier == UpdateModifier.UPDATE_DEPS:
+            final_specs = specs_to_add
+            for spec in specs_to_add:
+                prec = next(p for p in state["installed_pkgs"] if p.name == spec.name)
+                for dep in prec.depends:
+                    ms = MatchSpec(dep)
+                    if ms.name != "python":
```
jezdez, line 1220: That's a curious line, mind leaving a comment here?

conda/core/solve.py
```diff
@@ -1029,6 +1039,251 @@ def _prepare(self, prepared_specs):
...
+            specs_to_add = list(set(final_specs))
+        return [s.conda_build_form() for s in specs_to_add]
+
+    def _pin_python(self, state):
+        """
+        If Python is already present in the environment and the user didn't
+        include it in the requested packages, make sure it remains pinned to
+        the minor version.
+        """
+        installed_names = [rec.name for rec in state["installed_pkgs"]]
```
jezdez, line 1231: `state["installed_pkgs"]` is used a few times, we should make this a cached property or something.

conda/core/solve.py
```diff
@@ -1029,6 +1039,251 @@ def _prepare(self, prepared_specs):
...
+    def _build_transaction(self, state):
+        from mamba import mamba_api as api
+        from mamba.utils import to_txn
+
+        transaction = api.Transaction(
+            state["solver"],
+            api.MultiPackageCache(context.pkgs_dirs),
+            PackageCacheData.first_writable().pkgs_dir,
+        )
+        (names_to_add, names_to_remove), to_link, to_unlink = transaction.to_conda()
```
jezdez, line 1272: Mind elaborating what `to_conda` does? This seems like something we want to have implemented locally to be able to adopt to changes inside Conda easier.

jezdez (commented):
Just a few high-level comments

jaimergp:
Huh, all tests pass already? Really? o.o.

jezdez:
Uuuh, that's odd.. can we add a test to see if the `CONDA_SOLVER_LOGIC` env var works as expected?

jaimergp:
I have opened a new PR to enable integration tests and enumerate failures.

jaimergp:
Ah, I see why now. Most of the tests are testing `solve_for_final_state` which we haven't provided yet (most of the CLI targets `solve_for_transaction` instead). Since this method is _not_ in our subclass, it ends up calling the parent class. Will add "NotImplementedErrors" in those to mark the failures properly.

jaimergp:
There we go, some proper errors :) I'll enumerate in a separate PR.

jaimergp:
Ok, _now_ we are seeing implementation errors as expected!

jaimergp:
I cleaned the raw logs a bit to go through the errors and categorize them. Attached here for posterity.

- [mamba_errors_0.log](https://github.com/conda/conda/files/7090743/mamba_errors_0.log) (unit tests)
- [mamba_errors_1.log](https://github.com/conda/conda/files/7090745/mamba_errors_1.log) (integration tests group 1)
- [mamba_errors_2.log](https://github.com/conda/conda/files/7090746/mamba_errors_2.log) (integration tests group 2)
- [mamba_errors_3.log](https://github.com/conda/conda/files/7090757/mamba_errors_3.log) (integration tests group 3)

conda/core/solve.py
jaimergp: This is going to be tricky because it comes from the C++ API directly :/ I have simplified how we relied on `mamba.utils.to_txn` in #10893 and #10894.

conda/core/solve.py
jaimergp: Hm, `installed_pkgs` is only computed once while setting the state. After that it's only a dict lookup. What do you want to optimize?

jaimergp:
As of today, and considering the last commit in PR #10901, we have went from 63 failing integration tests to **47**.

jaimergp:
Description has been updated with latest failures (57 in total). I am excited because there are three of them that repeat across tests so hopefully that's a n-birds-one-stone situation.

jaimergp:
I've been looking into the update flag errors and maybe we need to reimplement some of the logic in https://github.com/conda/conda/blob/5eb0858fdfdcc16a53071513f681c4f420633b77/conda/core/solve.py#L860. Not sure how much of that is already on `libmamba`.

jaimergp:
This has been a nice week. As of Oct 1st, we are down to 33 failing tests on Linux. 10 days ago we were at 57!

jaimergp:
Checkpoint for the future. Our baseline is e90caa8.

jaimergp:
Rerunning to see if the rebuilt the images with mamba 0.19 also impacts Linux.

jaimergp:
Investigating errors in https://github.com/conda/conda/pull/11129

jaimergp:
Adding file logging in #11132

dpryan79:
For those of us watching from the peanut gallery could someone give a quick 1-line summary of how close this might be to coming out of draft status? I think I can speak for a lot of people when I say we would LOVE to see this functionality merged in! :)

jezdez:
> For those of us watching from the peanut gallery could someone give a quick 1-line summary of how close this might be to coming out of draft status? I think I can speak for a lot of people when I say we would LOVE to see this functionality merged in! :)

Thanks for the question, this is quite timely since we're currently in the process to prepare the release and started to track the preperation in the [new milestone for the libmamba solver](https://github.com/conda/conda/milestone/53). @jaimergp and me have discussed adding additional tickets soon to track non-technical tasks we need to get done to make it happen. I hope that helps!

jaimergp:
pre-commit.ci autofix

jaimergp:
Please do!

💐

jezdez:
Closing since this PR has served its purpose. Good bye, old friend. 😢

PR #10893, opened 2021-09-01
[Mamba POC] - Rework `to_txn` usage to fix virtual packages KeyErrors

jaimergp:
To fix the virtual package KeyErrors in the final transactions, we need to patch `mamba.utils.to_txn`. Instead, we are bringing that one over, since we will need to introduce more modifications in how that's being handled.

This should end up fixing the following issues listed in the main PR:

- [x] `KeyError: PackageRecord(_hash=-3340074908272569636, name='__glibc', ...)`, which affects the following tests:
  - [x] `IntegrationTests.test_clean_tarballs_and_packages`
  - [x] `IntegrationTests.test_noarch_python_package_with_entry_points`
  - [x] `IntegrationTests.test_noarch_python_package_without_entry_points`
  - [X] `IntegrationTests.test_remove_all` - fails but with different error `Failed: DID NOT RAISE <class 'conda.exceptions.PackagesNotFoundError'>`
  - [X] `IntegrationTests.test_remove_force_remove_flag` - fails but with different error `assert package_is_installed(prefix, 'python')`

If they are fixed, they won't appear in the failed report. If they do appear, it'll be for a different error, not `KeyError`.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1101,6 +1101,8 @@ class LibSolvSolver(Solver):
         # Maybe conda already handles that beforehand
         # https://github.com/mamba-org/mamba/blob/fe4ecc5061a49c5b400fa7e7390b679e983e8456/mamba/mamba.py#L426-L485
 
+        # See https://github.com/mamba-org/mamba/blob/89174c0dc06398c995895649f83862d9c8021acf/src/core/prefix_data.cpp#L13
+        # for the C++ implementation of PrefixData
         mamba_prefix_data = MambaPrefixData(self.prefix)
         mamba_prefix_data.load()
         installed_json_f, installed_pkgs = get_installed_jsonfile(self.prefix)
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -27,6 +27,7 @@ from ..common.compat import iteritems, itervalues, odict, text_type
 from ..common.constants import NULL
 from ..common.io import Spinner, dashlist, time_recorder
 from ..common.path import get_major_minor_version, paths_equal
+from ..common.url import split_anaconda_token, remove_auth
 from ..exceptions import PackagesNotFoundError, SpecsConfigurationConflictError, UnsatisfiableError
 from ..history import History
 from ..models.channel import Channel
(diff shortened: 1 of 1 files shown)


conda/core/solve.py
```diff
@@ -1100,8 +1101,8 @@ def _setup_state(self):
         # Maybe conda already handles that beforehand
         # https://github.com/mamba-org/mamba/blob/fe4ecc5061a49c5b400fa7e7390b679e983e8456/mamba/mamba.py#L426-L485
 
-        prefix_data = PrefixData(self.prefix)
-        prefix_data.load()
+        mamba_prefix_data = MambaPrefixData(self.prefix)
```
jezdez, line 1104: Could you please leave a comment here pointing to the C++ implementation of this class?

conda/core/solve.py
```diff
@@ -1272,27 +1273,83 @@ def _run_solver(self, state):
...
             PackageCacheData.first_writable().pkgs_dir,
         )
         (names_to_add, names_to_remove), to_link, to_unlink = transaction.to_conda()
         specs_to_add = [MatchSpec(m) for m in names_to_add]
         specs_to_remove = [MatchSpec(m) for m in names_to_remove]
 
-        return to_txn(
-            specs_to_add,
-            specs_to_remove,
+        to_link_records, to_unlink_records = [], []
```
jezdez, line 1291: Wondering if we want to keep the function structure for easier testing?

conda/core/solve.py
```diff
@@ -1272,27 +1273,83 @@ def _run_solver(self, state):
...
+            for i_rec in installed_pkgs:
+                # Do not try to unlink virtual pkgs
+                if not i_rec.is_unmanageable and i_rec.fn == pkg:
+                    final_precs.remove(i_rec)
+                    to_unlink_records.append(i_rec)
+                    break
+            else:
+                print("No package record found!")
+
+        for c, pkg, jsn_s in to_link:
```
jezdez, line 1315: What's `c` in this case?

jezdez (approved):
Just a few nits...

conda/core/solve.py
jaimergp: I want to think it's `c` for `channel` but I need to double check; this is coming straight from Mamba's code.

conda/core/solve.py
jaimergp: Done. Interestingly, it's not that complicated. It's essentially a JSON loader.

conda/core/solve.py
jaimergp: This is going away in #10894 so no need to work on it.

conda/core/solve.py
jaimergp: Double checked and fixed.

jaimergp:
@jezdez Comments have been addressed. Thanks!

jezdez (approved):

PR #10894, opened 2021-09-01
[Mamba POC] - Rework Solver API to rely on `Solver.solve_final_state`

jaimergp:
Built on top of #10893 

***

`Solver.solve_for_diff` and `Solver.solve_for_transaction` end up calling `Solver.solve_final_state`. The only requirement is to export some state back to the class. I don't like it, but it's how it works in the legacy `Solver`. We can discuss it.

This enables testing with `conda.core.test_solve` and the following issues noted in the main PR. Note we are not aiming at _fixing_ them, just uncovering the real error, since they were previously guarded by incomplete implementations.

`conda.core.test_solve` includes these:

- `tests/core/test_solve.py::test_only_deps_1`
    - Probably an API case we are not handling well (missing key in matchspec?)
- `tests/core/test_solve.py::test_force_remove_1`
    - Probably an API case we are not handling well (missing key in matchspec?)
- `tests/core/test_solve.py::test_prune_1`
    - Solution includes different package variants due to `features` weighting
- `tests/core/test_solve.py::test_no_deps_1`
    - Solution includes four packages that shouldn't be there.    
- `tests/core/test_solve.py::test_only_deps_2`
    - Solution includes a package that shouldn't be there.
- `tests/core/test_solve.py::test_update_all_1`
    - Subtle difference in the build string of a single package. Misconfigured solver?
- `tests/core/test_solve.py::test_broken_install`
    - Test relies on internal low-level object we do not use (Resolver). Annotate and skip.
- `tests/core/test_solve.py::test_conda_downgrade`
    - Some packages are not present in the proposed solution
- `tests/core/test_solve.py::test_auto_update_conda`
    - Auto updating conda ends up in downgraded version?
- `tests/core/test_solve.py::test_python2_update`
    - Unexpectedly unsolvable environment.
- `tests/core/test_solve.py::test_update_deps_1`
    - Different solutions found.
- `tests/core/test_solve.py::test_update_deps_2`
    - Different solutions found.
- `tests/core/test_solve.py::test_fast_update_with_update_modifier_not_set`
    - Different solutions found.
- `tests/core/test_solve.py::test_pinned_1`
    - Mishandled conflict.
- `tests/core/test_solve.py::test_force_reinstall_1`
    - Different solutions found.
- `tests/core/test_solve.py::test_channel_priority_churn_minimized`
    - Different solutions found.
- `tests/core/test_solve.py::test_priority_1`
    - Different solutions found.
- `tests/core/test_solve.py::test_features_solve_1`
    - Different solutions found.
- `tests/core/test_solve.py::test_freeze_deps_1`
    - Different solutions found.
- `tests/core/test_solve.py::test_current_repodata_fallback`
    - Missing package.
- `tests/core/test_solve.py::test_downgrade_python_prevented_with_sane_message`
    - Different solutions found.
- `tests/core/test_solve.py::test_indirect_dep_optimized_by_version_over_package_count`
    - Different solutions found.

Other errors:

- [x] `test_Solver_return_value_contract`
- [ ] `IntegrationTests.test_create_dry_run_json`, `IntegrationTests.test_create_dry_run_yaml` and others. I don't think we have implementes `conda create` yet (missing specs), but we will tackle this in a separate PR.


jaimergp:
Managed to export the ephemeral state of the in-memory channels to disk so Mamba can load them too. I tried hard not to break the existing code paths for Conda legacy, which involved readjusting the `mtime` attribute of the cache entry so they are more recent than the local dump.

Now both solver backends can run the tests. It seems that there are two different issues with `test_solve_1`. 
* The local dump exports `pip` as a dependency of Python, but it shouldn't. You can see this with Conda Legacy too if you disable the `mtime` readjustment in `helpers.py`
* Mamba also considers MKL as part of the solution, maybe something to do with the features?

I believe this has more to do with the local dump exporting mechanism than the actual Mamba solving behaviour. I'll keep looking into it.

jaimergp:
Seeing some side effects.

This passes:

```
CONDA_SOLVER_LOGIC=legacy pytest tests/core/test_solve.py --no-cov -vv -k test_aggressive_update_packages
```

Having `test_solve_1` run before `test_aggressive_update_packages` makes the latter fail:

```
CONDA_SOLVER_LOGIC=legacy pytest tests/core/test_solve.py --no-cov -vv -k "test_aggressive_update_packages or test_solve_1"
```

jaimergp:
Update: disabling memoization of the `get_index_r*` functions in `tests.helpers` makes everything pass again. I am assuming some tests have side effects that write to that cached object and it gets out of sync. The local copy _should not_ be queried from Conda legacy because the in-memory subdir-data copy should be returned right away from the cache.

jaimergp:
I think I found it:

* There's a fixture (with autouse=True) that will clear the cache of SubdirData automatically for every test
* That fixture calls `SubdirData.clear_cached_local_channel_data()`, which removes all the cache entries that start with `file://` ¬¬
* That's why my "expose state as a local file" trick only works if memoization is disabled: the cache entries for the local channels are never recreated! Unfair! This function is only used in unit tests anyway... so I guess we can customize it a bit :)

jaimergp:
Managed to pin down the issue with the added pip dependencies. Mamba respects this context option, but it needs to be present at Solver instantiation time again because the SubdirData cache is not used in Mamba (they have their own SubdirData C++ implementation).

jaimergp:
Ok, I'll leave it here for now.

First, I cut off some parts of the original drafted `solve_for_transaction` to just return the final state as `solve_final_state` expects. The diff and transaction objects are created by the parent class now with no problem. This allows tests to actually run as expected and uncover some errors.

The main issue now is that the synthetic data used for testing is not accessible by Mamba because state is stored differently, so we need to express that state in a way that Mamba understands. Namely:

*  The test machinery in `tests.core.test_solve` use fake channels populated with data from test files. Then, some helper functions modify that state with different specs to account for test cases and inject the modified subdir data in the class cache.
* The problem is that those in-memory changes never make it back to disk, so Mamba cannot access them. We fix this by exporting the fake `SubdirData` objects state to back to JSON and prefixing the channel name with the temporary location on disk that will host these files.  That way, Mamba can "download" the repodata.json.
* However, the local prefixes change the SubdirData cache behaviour and we need to patch that so Conda legacy gets the subdir data objects without disk access. Memoization and some automatic cache clearing gets in the way, so we need to repatch for every solver instance.

Without this, everything fails with "download errors". Now that that's fixed, we can actually see the actual solving errors with Mamba. Some categories:

- Exception messages are compared with string equality, which fails because Mamba uses different wordings. I added alternative messages in these cases.
- Different solver behaviour. Features / track_features is not fully implemented on Mamba, since this is technically deprecated. We tried to expose features as track_features in our exported disk state, which fixed some tests, but not all. We might never get this 100% right I guess.
- Some tests rely on `Solver.ssc` objects, which we are not using in our current implementation. These are skipped.


Hope it makes sense, exposing this one was tricky!

jaimergp:
This is ready for review @jezdez.

conda/core/solve.py
```diff
@@ -1268,30 +1260,65 @@ def _run_solver(self, state):
...
+            else:
+                key = split_anaconda_token(remove_auth(c))[0]
+            if key not in lookup_dict:
+                raise ValueError("missing key {} in channels: {}".format(key, lookup_dict))
+            sdir = lookup_dict[key]
+            rec = to_package_record_from_subjson(sdir, pkg, jsn_s)
+            final_precs.add(rec)
+            to_link_records.append(rec)
+
+        return IndexedSet(PrefixGraph(final_precs).graph)
```
jezdez, line 1321: Let's review this for performance regressions

tests/helpers.py
```diff
@@ -177,6 +181,74 @@ def add_feature_records_legacy(index):
         rec = make_feature_record(feature_name)
         index[rec] = rec
 
+
+def _export_subdir_data_to_repodata(subdir_data, index):
+    state = subdir_data._internal_state
```
jezdez, line 186: Leave a comment to explain that this is just transitionary

jezdez (approved):
LTGM with minor nits and comments

PR #10929, opened 2021-09-20
[Mamba POC] Enable windows tests

jaimergp:
Only to have an idea of how bad things are down there ;)


jezdez (approved):
LGTM!

jaimergp:
Ok, good news, Windows has some more failing tests than Linux, 72 vs 57. Not _that_ bad I hope!

PR #10935, opened 2021-09-23
Fetch and run plantuml via coursier

dbast:
Vendors coursier (40 KByte vs ~8 MByte plantuml.jar). Coursier bootstraps
itself, fetches and runs .jars from e.g. maven central including caching,
consistency checks and java dependecy resolution if required.

Coursier is from https://github.com/coursier/coursier/releases/download/v2.0.16/coursier


dbast:
Nice. Github docs workflow and RTD seem to be fine with it :)

jezdez (approved):
Nice, thank you!

PR #10936, opened 2021-09-23
[Mamba POC] Fix YAML parsing in `test_create_dry_run_yaml`

jaimergp:
Conda and Mamba have different standard outputs in CLI. This test hardcoded a line number to begin the YAML parsing; instead we find the first token without assuming any particular output before that line.


jezdez:
Would you mind fixing the linting issue?

jezdez (approved):
Other than that this looks good, thanks!

PR #10937, opened 2021-09-23
[Mamba POC] Handle UpdateModifier.UPDATE_DEPS correctly (all graph, not just leaves)

jaimergp:
This tries to fix `test_update_deps_flag_present`. Mamba code only handled the leave nodes I think. In this case we take conda's approach and do a pre-solve to obtain the full chain.

I've also reworked how the solver tasks are precomputed here. This might affect other tests too. Keep an eye on regressions.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1411,6 +1411,8 @@ class LibSolvSolver(Solver):
             # Remove pinned_specs and any python spec (due to major-minor pinning business rule).
             for spec in self._pinned_specs():
                 specs_map.pop(spec.name, None)
+            # TODO: This kind of version constrain patching is done several times in different
+            # parts of the code, so it might be a good candidate for a dedicate utility function
             if "python" in specs_map:
                 python = next((p for p in state["installed_pkgs"] if p.name == "python"), None)
                 if python is not None:
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -7,6 +7,7 @@ import copy
 from genericpath import exists
 from logging import DEBUG, getLogger
 from os.path import join
+from collections import defaultdict
 import sys
 from textwrap import dedent
 from itertools import chain
@@ -1095,9 +1096,11 @@ class LibSolvSolver(Solver):
         # 2. Create solver and needed flags, tasks and jobs
         self._configure_solver(
             state,
-            force_reinstall=kwargs["force_reinstall"],
-            deps_modifier=kwargs["deps_modifier"],
             update_modifier=kwargs["update_modifier"],
+            deps_modifier=kwargs["deps_modifier"],
+            ignore_pinned=kwargs["ignore_pinned"],
+            force_remove=kwargs["force_remove"],
+            force_reinstall=kwargs["force_reinstall"],
         )
         # 3. Run the SAT solver
         self._run_solver(state)
@@ -1108,10 +1111,10 @@ class LibSolvSolver(Solver):
             self,
             update_modifier=NULL,
             deps_modifier=NULL,
-            prune=NULL,
             ignore_pinned=NULL,
             force_remove=NULL,
             force_reinstall=NULL,
+            prune=NULL,
             should_retry_solve=False):
         """
         Context options can be overriden with the signature flags.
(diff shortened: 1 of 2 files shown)


conda/core/solve.py
```diff
@@ -1324,28 +1345,65 @@ def _configure_solver_for_remove(self, state):
...
+            force_remove=NULL,
+            force_reinstall=NULL):
         """
         Reimplement the logic found in super()._add_specs(), but simplified.
         """
+        from mamba import mamba_api as api
+        task_and_specs = defaultdict(set)
         # A. Explicitly requested packages via CLI (either update or install)
-        specs_to_add = list(self.specs_to_add)
+        task = (api.SOLVER_UPDATE if self._command == "update" else api.SOLVER_INSTALL)
```
jezdez, line 1361: No need for extra parentheses.
```suggestion
        task = api.SOLVER_UPDATE if self._command == "update" else api.SOLVER_INSTALL
```

conda/core/solve.py
```diff
@@ -1324,28 +1345,65 @@ def _configure_solver_for_remove(self, state):
...
+            specs_map = {name: MatchSpec(name) for name in update_names}
+
+            # Remove pinned_specs and any python spec (due to major-minor pinning business rule).
+            # Add in the original specs_to_add on top.
+            for spec in self._pinned_specs():
+                specs_map.pop(spec.name, None)
+            if "python" in specs_map:
+                python_rec = next((p for p in state["installed_pkgs"] if p.name == "python"), None)
+                if python_rec is not None:
+                    py_ver = ".".join(python_rec.version.split(".")[:2]) + ".*"
```
jezdez, line 1401: This seems a bit brittle, is there a chance that `python_rec.version` could be an empty string or nonsensible version?

conda/core/solve.py
```diff
@@ -1363,9 +1421,9 @@ def _pin_python(self, state, update_modifier=NULL):
             return f"python {major_minor_version}.*"
 
     def _pinned_specs(self):
-        pinned_specs = get_pinned_specs(context.target_prefix)
+        pinned_specs = get_pinned_specs(self.prefix)
```
jezdez, line 1424: Could you elaborate why this change?

conda/core/solve.py
```diff
@@ -1324,28 +1345,65 @@ def _configure_solver_for_remove(self, state):
...
+                deps_modifier=deps_modifier,
+                ignore_pinned=ignore_pinned,
+                force_remove=force_remove,
+                force_reinstall=force_reinstall)
+            graph = PrefixGraph(solved_pkgs)
+            update_names = set()
+            for spec in self.specs_to_add:
+                node = graph.get_node_by_name(spec.name)
+                update_names.update(ancest_rec.name for ancest_rec in graph.all_ancestors(node))
+            specs_map = {name: MatchSpec(name) for name in update_names}
```
jezdez, line 1392: Could we move this into a matching PrefixGraph method?

jezdez: Actually including the removal of pinned specs below?

jezdez (approved):
Just a few nits and questions :)

jaimergp:
Oops, sorry Jannis, I wasn't finished with this yet. I'll mark it as draft.

jezdez:
No problem ;)

jaimergp:
Phew, it shows that we are entering tricky territory now. Let's review this for now @jezdez

conda/core/solve.py
jaimergp: Some tests do pass the prefix explicitly to the class (which defaults to `context.target_prefix` if `prefix` is not passed explicitly).

conda/core/solve.py
```diff
@@ -1324,28 +1347,93 @@ def _configure_solver_for_remove(self, state):
...
+            for spec in specs_to_add_plus_deps:
+                node = graph.get_node_by_name(spec.name)
+                update_names.update(ancest_rec.name for ancest_rec in graph.all_ancestors(node))
+            specs_map = {name: MatchSpec(name) for name in update_names}
+
+            # Remove pinned_specs and any python spec (due to major-minor pinning business rule).
+            for spec in self._pinned_specs():
+                specs_map.pop(spec.name, None)
+            if "python" in specs_map:
+                python = next((p for p in state["installed_pkgs"] if p.name == "python"), None)
```
jezdez, line 1415: Let's add a TODO item here that we may want to abstract this into a utility API for other cases in which we want to restrict this spec map.

jezdez (approved):
Great, some nits.

PR #10944, opened 2021-09-28
[Mamba POC] Do not freeze explicitly (or implicitly) requested packages

jaimergp:
This fixes `test_install_freezes_env_by_default`. If a package is frozen _and_ requested through `specs_to_add` (or any of the implicitly added names via `--update-all`, `--update-deps`, etc), it creates a conflict in `libsolv` because it tries both to lock the currently installed and also explicitly install a new version. Thus, the solution is to not freeze those names mentioned in the job list.

Conda does something similar here:

https://github.com/conda/conda/blob/0b1312ce65bf0fbb8cbea3750f07c32a2492c57a/conda/core/solve.py#L613-L620

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1284,20 +1284,24 @@ class LibSolvSolver(Solver):
 
         # Configure jobs
 
-        # 1. Lock currently installed ones, if requested
-        if update_modifier is UpdateModifier.FREEZE_INSTALLED:
-            solver.add_jobs([p for p in prefix_data.package_records], api.SOLVER_LOCK)
-
-        # 2. Install/update user requested packages
+        # 1. Install/update user requested packages
         tasks = self._collect_specs_to_add(state,
                                            update_modifier=update_modifier,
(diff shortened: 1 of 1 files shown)


conda/core/solve.py
```diff
@@ -1284,21 +1284,25 @@ def _configure_solver_for_install(self,
...
-
-        # 2. Install/update user requested packages
+        # 1. Install/update user requested packages
         tasks = self._collect_specs_to_add(state,
                                            update_modifier=update_modifier,
                                            deps_modifier=deps_modifier,
                                            ignore_pinned=ignore_pinned,
                                            force_remove=force_remove,
                                            force_reinstall=force_reinstall)
+        explicitly_or_implicitly_requested_names = set()
```
jezdez, line 1294: Nice variable name ;) We could also just call it `requested_names` and leave a comment that they can be requested both ways. 😆

jezdez (approved):
LGTM with nits :)

PR #10945, opened 2021-09-28
[Mamba POC] Control verbosity on tests where stdout is parsed

jaimergp:
Finally an easy one. Libsolv will print logs to stdout if verbosity=3 is enabled, which pollutes the JSON output that needs to be parsed here. Note that the verbosity is set at first instantiation, so adjusting `CONDA_VERBOSITY` after won't have any effect. In other words, it needs to be done as early as possible.

This fixes `test_install_force_reinstall_flag`.


jezdez (approved):
Huzzah! 🥳

PR #10946, opened 2021-09-28
[Mamba POC] Allow unfrozen retries

jaimergp:
This fixes `test_noarch_python_package_reinstall_on_pyver_change` and opens the possibility of other attempts at doing increasingly relaxed solves (e.g. neutered specs, which we don't implement yet).

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1089,31 +1089,43 @@ class LibSolvSolver(Solver):
         if none_or_final_state is not None:
             return none_or_final_state
 
+        # By default, conda tries to freeze everything not in the specs
+        # to minimize the amount of packages needed; however sometimes
+        # that's too strict and we need to relax:
+
         # These tasks DO need a solver
         # 1. Populate repos with installed packages
         state = self._setup_state()
-        # 2. Create solver and needed flags, tasks and jobs
-        self._configure_solver(
-            state,
-            update_modifier=kwargs["update_modifier"],
-            deps_modifier=kwargs["deps_modifier"],
-            ignore_pinned=kwargs["ignore_pinned"],
-            force_remove=kwargs["force_remove"],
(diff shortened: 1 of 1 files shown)


jezdez (approved):

conda/core/solve.py
```diff
@@ -1089,31 +1089,43 @@ def solve_final_state(self, update_modifier=NULL, deps_modifier=NULL, prune=NULL
...
-        # 5. Refine solutions depending on the value of some modifier flags
-        return self._post_solve_tasks(
-            state,
-            update_modifier=kwargs["update_modifier"],
-            deps_modifier=kwargs["deps_modifier"],
-            ignore_pinned=kwargs["ignore_pinned"],
-            force_remove=kwargs["force_remove"],
-            force_reinstall=kwargs["force_reinstall"]
-        )
+        while True:
```
jezdez, line 1099: That while loop seems odd, could you elaborate why we need this here?

jezdez: If this is to catch the exceptions below and retry with the changed `update_modifier`, let's use a different type of flow control than a while loop like a separate method that implements the core logic of `solve_final_state`.

jaimergp: Yes, that's the intended use. Do you have any suggestions on how to implement the retry loop? Tagging @FFY00 in case they want to add their two cents too :)

jaimergp: I pushed a different approach, hopefully a bit more explicit. Let me know what you think! I am not a big fan of creating locally scoped functions but I'll let you decide on that.

jaimergp: Take into account that this will carry over to the changes in the `pin-history` branch family.

jaimergp:
Note that this PR alone fixes 7+ tests, so the retrying loop is kind of important :)

PR #10947, opened 2021-09-28
[Mamba POC] Pin historic entries

jaimergp:
I am not sure we should even be doing this, but `test_neutering_of_historic_specs` is failing and this fixes it. I need to assess whether:

a) This impacts other tests
b) This is actually done by conda at all. I think legacy uses a subtler approach to pinning historic entries (and not just _everything_ ever requested).

I also modified the test a bit to actually test whether `psutil` is kept at the requested version or not. Without this, the tests passes too, but `libsolv` is actually bringing psutil up to 5.8.0 instead of staying at 5.6.3.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -38,6 +38,7 @@ from ..models.match_spec import MatchSpec
 from ..models.prefix_graph import PrefixGraph
 from ..models.version import VersionOrder
 from ..resolve import Resolve
+import conda
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


jezdez:
@chenghlee I've added you to this since this is a bit outside my wheelhouse, I'd really would be interested if you know this area. Thank you!

jaimergp:
This fixes some tests but also breaks a previously passing `test_update_deps_1`. Gotta be careful with the pinning here!

PR #10948, opened 2021-09-28
Force remove miniconda in dev/start

kenodegard:
Some users (aka me) like to alias `rm` as `rm -i` so avoid that by including `-f`.


jezdez (approved):
That's a bit random TBH, but it doesn't hurt either.

PR #10949, opened 2021-09-28
Remove extra argv processing for Python2 win32 unicode

kenodegard:
This is a remnant of Python2 days and since conda is now (at least) Python3.6+ theres no need to keep the hack around.


jezdez:
Thanks but I have a branch already with this type of thing and I'd rather keep the removal in one big commit. There is *a lot* of this. I'll keep you posted when you can review it!

PR #10952, opened 2021-09-29
Allow to use add, append and prepend on mappable of sequences

janjagusch:
With this PR, I'm adding support for the `--add`, `--append` and `--prepend` operator for maps of sequences.

## Motivation

Parts of the config logic assume that the values of maps cannot be sequences, but that's not true. For example the values for the keys in `custom_multichannels` are sequences:

```
custom_multichannels:
  foo:
    - bar
```

This is something you currently cannot express through the CLI. Executing the following command

```
conda config --append custom_multichannels.foo "bar"
```

causes the following error

```
Key 'custom_multichannels.foo' is not a known sequence parameter.
```

With my changes this now becomes possible.

## Caveat

There is currently no way to distinguish between maps of scalars and maps of sequences. This change allows users to use `--add`, `--append` and `--prepend` with any map in the config, regardless of whether it takes a sequence as value. To avoid this, we would need a different way of storing the schema of nested configuration objects.

### DIFF
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -309,22 +309,26 @@ def execute_config(args, parser):
     # prepend, append, add
     for arg, prepend in zip((args.prepend, args.append), (True, False)):
         for key, item in arg:
+            key, subkey = key.split('.', 1) if '.' in key else (key, None)
             if key == 'channels' and key not in rc_config:
                 rc_config[key] = ['defaults']
-            if key not in sequence_parameters:
+            if key in sequence_parameters:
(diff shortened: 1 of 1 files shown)


janjagusch:
i have signed the pdf, could somebody review that please? 🙂

xhochy:
> on a more general note: i think it would make sense to add a [JSON schema](https://json-schema.org/) for the configuration file to improve the validation and make certain expression more flexible.

This would then probably go into https://github.com/conda/schemas

kenodegard:
@anaconda-issue-bot check

PR #10957, opened 2021-10-01
[Mamba POC] Do not drop python if it was requested along with --only-deps

jaimergp:
Built on top of https://github.com/conda/conda/pull/10946

***

Python is special-cased here because otherwise the directory tree for other packages is not available for the dependencies. This makes `test_create.py::test_create_only_deps_flag ` pass.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1338,7 +1338,7 @@ class LibSolvSolver(Solver):
             solver_options.append((api.SOLVER_FLAG_STRICT_REPO_PRIORITY, 1))
         solver = api.Solver(state["pool"], solver_options)
 
-        installed_names = [rec.name for rec in state["installed_pkgs"]]
+        installed_names = set(rec.name for rec in state["installed_pkgs"])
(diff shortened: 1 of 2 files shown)


conda/core/solve.py
```diff
@@ -1578,7 +1578,9 @@ def _post_solve_tasks(self,
                 if spec.name in original_prefix_map:
                     final_prefix_map[spec.name] = original_prefix_map[spec.name]
                 # Case B: it was never installed, make sure we don't add it.
-                else:
+                # Note that we special-case python, otherwise there's no path
+                # we can still packages to. This makes `test_create_only_deps_flag` pass.
```
jezdez, line 1582:
```suggestion
                # Python is special-cased here because otherwise the directory
                # tree for other packages is not available for the dependencies.
                # This makes test_create.py::test_create_only_deps_flag pass.
```

jezdez (approved):
Just a tiny update to the comment since your PR description was better.

PR #10958, opened 2021-10-01
[Mamba POC] Don't try removing packages that are not installed

jaimergp:
Built on top of https://github.com/conda/conda/pull/10957

***

This fixes `test_create.py::test_remove_all`, which tests https://github.com/conda/conda/issues/2154. This needs to be upstreamed in Mamba too.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1338,7 +1338,7 @@ class LibSolvSolver(Solver):
             solver_options.append((api.SOLVER_FLAG_STRICT_REPO_PRIORITY, 1))
         solver = api.Solver(state["pool"], solver_options)
 
-        installed_names = [rec.name for rec in state["installed_pkgs"]]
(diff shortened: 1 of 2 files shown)


conda/core/solve.py
```diff
@@ -1348,6 +1348,10 @@ def _configure_solver_for_remove(self, state):
             list(set(chain(self._history_specs(), aggresive_updates))),
             api.SOLVER_USERINSTALLED,
         )
+        not_installed = [s.name for s in self.specs_to_remove if s.name not in installed_names]
```
jezdez, line 1351: Let's leave a comment about the ticket here and use 
```suggestion
        not_installed = [spec.name for spec in self.specs_to_remove if spec.name not in installed_names]
```

jezdez (approved):
Woop woop!

PR #10959, opened 2021-10-01
[Mamba POC] Raise right exception type

jaimergp:
Built on top of https://github.com/conda/conda/pull/10958

***

Re-raise `RawStrUnsatisfiableError` as `PackagesNotFound` for this test to pass. I don't want to add this to the code itself because it's irrelevant to the user, and Mamba devs might end up working on a nicer way to do this.

See issue https://github.com/mamba-org/mamba/issues/1196 for more details.


jezdez (approved):
Good call, we can come back later again.

PR #10961, opened 2021-10-02
Update CONTRIBUTING.md

anuj-s1ngh:
added the link for conda-env issues.


kenodegard:
@anuj-s1ngh https://github.com/conda/conda-env is the old, _archived_ repo for the conda-env code, conda-env issues should be reported here to conda itself

PR #10964, opened 2021-10-05
Abstract solver running into separate method.

jezdez:
Hey Jaime, this is related to #10946 where I said that abstracting the retry/fallback system into different method would be better for flow control. What do you think?


jezdez:
Closing since this has become moot while we're refactoring this code more. /cc @jaimergp

PR #10983, opened 2021-10-11
[Mamba POC] Bring in more conda logic

jaimergp:
This PR implements a consolidated logic after studying the code in `Solver._collect_all_metadata()`, `Solver._add_specs()` and `Solver._post_sat_handling`. It's a reimplementation that doesn't use the SolverStateContainer features found in the `Resolve` instance they carry. More comments when needed, but it still needs to be documented thoroughly.

It's not perfect yet (conflict analysis is pending), but I am seeing more passing tests now.

This is an alternative implementation to the retry logic and history pinning found in #10946, #10947.

Let's see how many we can tackle now:

- [ ] `tests/test_create.py`
  - [x] `test_allow_softlinks`
  - [x] `test_clone_offline_multichannel_with_untracked`
  - [x] `test_conda_pip_interop_conda_editable_package` -- this passes locally! :/
  - [x] `test_conda_recovery_of_pip_inconsistent_env`
  - [x] `test_install_features`
  - [x] `test_install_force_reinstall_flag`
  - [x] `test_install_mkdir `
  - [x] `test_list_with_pip_wheel `
  - [x] `test_neutering_of_historic_specs`
  - [x] `test_noarch_python_package_reinstall_on_pyver_change`
  - [x] `test_noarch_python_package_without_entry_points`
  - [x] `test_offline_with_empty_index_cache`
  - [x] `test_package_optional_pinning`
  - [x] `test_packages_not_found `
  - [x] `test_pinned_override_with_explicit_spec`
  - [x] `test_remove_all `
  - [x] `test_remove_features `
  - [x] `test_update_with_pinned_packages`
  - [x] `test_create_env_default_packages`

- [ ] `tests/cli/test_cli_install.py`
  - [x] `test_find_conflicts_called_once`

- [x] `tests/conda_env/test_cli.py`
  - [x] `test_update `
  - [x] `test_update_env_json_output`

- [ ] `tests/core/test_solve.py`
  - [x] `test_pinned_1`
  - [x] `test_freeze_deps_1`
  - [x] `test_prune_1`
  - [x] `test_update_all_1`
  - [x] `test_conda_downgrade`
  - [x] `test_python2_update`
  - [ ] `test_fast_update_with_update_modifier_not_set`
  - [x] `test_force_reinstall_1`
  - [ ] `test_channel_priority_churn_minimized`
  - [x] `test_priority_1`
  - [x] `test_features_solve_1`
  - [x] `test_current_repodata_fallback`


jaimergp:
I did some research today to better understand the failures. Some notes to self for next week:

`tests/core/test_solve.py::test_fast_update_with_update_modifier_not_set`

* `python=2` env is forced to update with `python`, but `libedit` and `ncurses` do not get updated. No extra flags set.

`tests/core/test_solve.py::test_conda_downgrade`

* conda gets downgraded along with python (37 -> 36), but some packages that should not change did (but maybe they should?). Namely, `pysocks` and `urllib3`. Additionally, `openssl` and `xz` should have changed too, but didn't.

`tests/core/test_solve.py::test_channel_priority_churn_minimized`

* `itsdangerous-0.24-py37_1` should have changed to `itsdangerous-0.24-py_0`

`tests/test_create.py::IntegrationTests::test_conda_pip_interop_conda_editable_package`

For some reason this passes locally but not on CI. No clue why.

`tests/test_create.py::IntegrationTests::test_pinned_override_with_explicit_spec`

This tests checks whether a pinned python is overridden if an explicit spec is passed. I don't think we are implementing this correctly yet, so we can actually fix this :D 

Also I think the specs are passed in the wrong way here. The requested should be pinned, and the other way around:

```
conda.exceptions.SpecsConfigurationConflictError: Requested specs conflict with configured specs.
  requested specs: 
    - python=3.6.5
  pinned specs: 
    - defaults/linux-64::python==3.6.13=h12debd9_1
```

`tests/test_create.py::IntegrationTests::test_update_with_pinned_packages`

* One more case of pinned overrides. I think we need to force an update if a package is asked with no constrains. Mamba will declare no change as a good solution, while `conda` does change the env. We might need something similar to what we do with Python at the end of `compute_specs_map`.

jaimergp:
Fixed `tests/test_create.py::IntegrationTests::test_update_with_pinned_packages` but it's not pretty. This is fixed with the section 11 in `_compute_specs_map`. It implements a way to implicitly extend a bare spec (just the package name, as in `conda install numpy`) with the presently installed version in the environment (`conda install numpy!=x.y.z`), so it forces an update.

However this introduces failures in other tests that are testing for the "nothing to do here" situation, when the environment already satisfies the given tasks. I needed to special case pip and setuptools so they are implicitly restrained, to find a sweet spot there, but it doesn't feel right.

I don't understand why sometimes `conda update numpy` needs to regress something if the history is set up in just the right way, but other times is ok to leave things as is. Too many states interplaying.

The fix also required reporting all installed packages as conflicting (so they are unfrozen during state collection) as a last opportunity to fix conflicts, due to mamba not reporting all missing packages at once. We annotate the conflicts with a `.missing` attribute depending on the type of error message, because `.missing` packages shouldn't be restrained like above (e.g. asking for zmq > 2.2 won't work because there are no packages that satisfy that restrain alone). 

This is getting more and more complicated, and the history is always involved. A compelling reason to drop it in a future release, I believe.

jaimergp:
Down to 3.5 failing tests! `test_conda_pip_interop_conda_editable_package` passes locally :/

jaimergp:
All green in _both_ platforms, except four tiny issues on Windows:

Something about cleaning the index cache. They pass in isolation, but not with other tests in the same run. Fun stuff.

- [ ] `tests/test_create.py::IntegrationTests::test_clean_index_cache`
- [ ] `tests/test_create.py::IntegrationTests::test_use_index_cache`

Something about paths:

- [ ] `tests/test_create.py::IntegrationTests::test_install_tarball_from_local_channel`. This tries to create a Unix-like path for a Windows path and fails; e.g. `C:\Some\Dir` is turned into `/C:/Some/Dir`. Unclear if this is on us or mamba.
- [ ] `tests/test_create.py::IntegrationTests::test_legacy_repodata`. Fails with `RuntimeError: Multi-download failed.` and no more details. Probably path related as well. Unclear why it fails on Windows only, but this fake channel only has data for `noarch` so maybe Mamba is trying to download stuff from the win-64 directory and failing? Anyway, it says "legacy", we could skip :)

I will fix this first thing next week and merge for all green goodness.

jaimergp:
Also note that I [needed to use restricted unicode on Windows for now](https://github.com/conda/conda/pull/10983/files#diff-1e27c53a97d6c9242ea92544453665b553fe9fdf63f08717dc915f80e76fd60cR148). This is being fixed in libmamba as we speak and we will be able to undo that workaround in the future.

And I also took split `test_solvers.py` tests into legacy and libsolv for CI purposes. LibSolv is only tested in the `experimental-resolver.sh` script. That way the CI checks tell us immediately if something is failing on legacy or not.

jaimergp:
Looks like the index cache issue might be Windows leaving something open... Logs say:

```
DEBUG conda.gateways.disk.delete:unlink_or_rename_to_trash(136): renaming file path D:\Users\jrodriguez\Documents\conda\devenv\pkgs\cache\119df107.json to trash failed.  Output was: < empty >
WARNING conda.gateways.disk.delete:unlink_or_rename_to_trash(143): Could not remove or rename D:\Users\jrodriguez\Documents\conda\devenv\pkgs\cache\119df107.json.  Please remove this file manually (you may need to reboot to free file handles)
```

jaimergp:
Is this the first time everything's green? 🔥

jezdez (approved):
This is ready. Thank you @jaimergp!

PR #10992, opened 2021-10-14
Add Python 3.10 as a test target.

jezdez:
Fix #10991.

PR #10993, opened 2021-10-14
Build Docker image for Linux with Python 3.10.

jezdez:
Just the Docker image part of #10992 so we can use the image in the PR.


dbast:
That requires a conda release first... otherwise the conda in the current Dockerfile cannot create Python 3.10 environments.

jezdez:
Seems like building the 3.10 image worked now.

beeankha (approved):

kenodegard (approved):

PR #11009, opened 2021-10-29
Pre-commit, darker, blacken-docs, flake8, pylint, bandit

### LINKED ISSUE #10932: Add pre-commit to code base
### Checklist

### What is the idea?

We're testing code quality with flake8 right now, but don't have a good way to enforce code quality in local development environments. Instead of using a custom setup, let's use [pre-commit](https://pre-commit.com) for this and maybe even enable [pre-commit CI](https://pre-commit.ci) to get quicker results for PRs.

### Why is this needed

This is needed to streamline the development environment and enforce code quality checks for contributors before they are pushed to their forks.

### What should happen?

Add a pre-commit configuration and enable at least the flake8 hook.

### Additional Context

_No response_

kenodegard:
Adds necessary [pre-commit](https://pre-commit.com/) config file.

Do we want to update the `README`/`CONTRIBUTING` in this PR to reflect these changes?

After internal brainstorming and discussions we decided on using the following CQ tools:
- [darker](https://github.com/akaihola/darker) (we want the formatting to be gradually applied)
- [blacken-docs](https://github.com/asottile/blacken-docs)
- [flake8](https://github.com/PyCQA/flake8) (this is what we have been using)

We also include two other useful tools which are run but we currently do not monitor the results, these results can be reviewed by running pre-commit in verbose mode:
- [pylint](https://github.com/PyCQA/pylint)
- [bandit](https://github.com/PyCQA/bandit)

Resolves #10932
CC: @rchord @jezdez

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,9 @@
+[tool.black]
+line-length = 99
+target-version = ['py36', 'py37', 'py38']
+extend-exclude = '''
+^/(
+    conda/_vendor
+  | devenv
+)/
+'''
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,39 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+      # standard end of line/end of file cleanup
+      - id: mixed-line-ending
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      # ensure syntaxes are valid
+      - id: check-toml
+      - id: check-yaml
+      # catch git merge/rebase problems
+      - id: check-merge-conflict
+  - repo: https://github.com/akaihola/darker
+    rev: 1.3.2
+    hooks:
+      - id: darker
+        additional_dependencies: [black==21.9b0]
+        exclude: ^conda/_vendor/
+  - repo: https://github.com/asottile/blacken-docs
+    rev: v1.11.0
+    hooks:
+      - id: blacken-docs
+        additional_dependencies: [black==21.9b0]
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+  - repo: https://github.com/PyCQA/pylint
+    rev: v2.11.1
+    hooks:
+      - id: pylint
+        args: [--exit-zero]
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.0
+    hooks:
+      - id: bandit
+        args: [--exit-zero]
+        exclude: ^(conda/_vendor|tests)/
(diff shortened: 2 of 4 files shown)


dbast:
Nice! How about also changing the lint action and install `pre-commit` via `conda install conda-forge::pre-commit` and then running `pre-commit run -a` instead of flake8 at https://github.com/conda/conda/blob/master/.github/workflows/lint.yml#L31 ?

dbast:
How about also activating some of the hooks of the default pre-commit repo?, e.g.
```
- repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.0.1
   hooks:
   - id: mixed-line-ending
   - id: end-of-file-fixer
   - id: trailing-whitespace
```

If the pyproject.toml file only serves the purpose to provide parameters for black, while the underlying project is actually not pyproject.toml based (which is the case here), then the black parameters can also be set as pre-commit hook args e.g. 

```
-   repo: https://github.com/psf/black
    rev: 21.9b0
    hooks:
    -   id: black
        args: [-l, "99", --skip-magic-trailing-comma]
```

kenodegard:
@dbast I hadn't modified our CI yet since we've been discussing using pre-commit CI.

I will look into those default hooks, I hadn't included them since trailing whitespace and end of line is fixed by black/darker so I was avoiding redundancy.

I disagree with removing the config file as that will make the config inaccessible to users who may choose to use these tools directly. I have no problem using a different config file but it doesn't look like black/darker accept anything besides `pyproject.toml`.

pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.black]
+line-length = 99
+target-version = ['py36', 'py37', 'py38']
+extend-exclude = '''
+# A regex preceded with ^/ will apply only to files and directories
+# in the root of the project.
+/(_vendor|devenv|tests)/
```
jezdez, line 7: I think the tests should be included in the move to use black.

```suggestion
/(_vendor|devenv)/
```

jezdez (approved):
This is a good start!

pyproject.toml
kenodegard: Makes perfect sense, I don't remember why I included `tests` here...

kenodegard: Done

kenodegard:
@jezdez do we want to enable pre-commit CI after merging this in or do we want to do as @dbast suggested and do `pre-commit run -a` in the Lint GH action?

jezdez:
> @jezdez do we want to enable pre-commit CI after merging this in or do we want to do as @dbast suggested and do `pre-commit run -a` in the Lint GH action?

Let’s go with the service since we’re limited on number of runners already and we’ll soon add Python 3.10 which is added on top.

jezdez:
That said, documenting how to use pre-commit locally is worth the effort for contributors.

kenodegard:
Updated `CONTRIBUTING.md` with a static code analysis section and realized that I should add pre-commit to `tests/requirements.txt` that way it gets installed with `devenv`.

@jezdez I'll leave this for you to review one last time and merge as you see fit since you will also be able to enable pre-commit CI at the same time

jezdez (approved):
Love it!

PR #11012, opened 2021-11-01
GitHub Action to build/upload conda packages on CI tests success

chenghlee:
Build `conda` conda packages and upload them to the specified anaconda.org channel and label, which must be configured as per-fork repository secrets.  For this "official" conda/conda repository, I've already configured these to be the `conda-canary` channel and `dev` label.

The objective is to provide users access to development versions of `conda` in a way more similar to how they would upgrade released versions; i.e., via `conda update` and not via setting up a development environment.

[Example run of this GHA](https://github.com/chenghlee/conda/actions/runs/1409300954)


conda.recipe/run_test.sh
```diff
@@ -12,11 +14,11 @@ export TEST_PLATFORM=$(python -c "import sys; print('win' if sys.platform.starts
...
 conda info
-conda create -y -p ./built-conda-test-env python=3.9
+conda create -y -p ./built-conda-test-env python=3.${TEST_MINOR_VER}
 conda activate ./built-conda-test-env
 echo $CONDA_PREFIX
 [ "$CONDA_PREFIX" = "$PWD/built-conda-test-env" ] || exit 1
-[ $(python -c "import sys; print(sys.version_info[1])") = 9 ] || exit 1
-python -c '__requires__ = ["ruamel_yaml_conda >= 0.11.14"]; import pkg_resources' || exit 1
+[ $(python -c "import sys; print(sys.version_info[1])") = ${TEST_MINOR_VER} ] || exit 1
+#python -c '__requires__ = ["ruamel_yaml_conda >= 0.11.14"]; import pkg_resources' || exit 1
```
jezdez, line 22: Let's remove this line if it's not needed.

conda.recipe/run_test.sh
```diff
@@ -12,11 +14,11 @@ export TEST_PLATFORM=$(python -c "import sys; print('win' if sys.platform.starts
...
 conda deactivate
-pytest tests -m "not integration and not installed" -vv
+#pytest tests -m "not integration and not installed" -vv
```
jezdez, line 24: Same for this, no need to run the test suite as part of this process.

conda.recipe/meta.yaml
```diff
@@ -97,3 +94,4 @@ extra:
     - patricksnape
     - pelson
     - scopatz
+    - chenghlee
```
jezdez, line 97: Mind adding @kenodegard and me as well?

.github/workflows/deploy-dev-pkgs.yml
```diff
@@ -0,0 +1,82 @@
...
+      - master
+    types:
+      - completed
+
+jobs:
+  # Check for required configuration/secrets needed to build & upload.  Done
+  # this way because secrets cannot be used directly in GitHub Action
+  # conditionals; see https://github.com/actions/runner/issues/520 for details.
+  check-config:
+    runs-on: ubuntu-18.04
```
jezdez, line 17: Can we use ubuntu-20.04 here?

.github/workflows/deploy-dev-pkgs.yml
```diff
@@ -0,0 +1,82 @@
...
+  build-conda-pkgs:
+    needs:
+      - check-config
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      needs.check-config.outputs.okay == 'true'
+    strategy:
+      matrix:
+        include:
+          - runner: ubuntu-18.04
```
jezdez, line 46: Same, do we need to run 18.04 here or would 20.04 be also ok?

.github/workflows/deploy-dev-pkgs.yml
```diff
@@ -0,0 +1,82 @@
+name: Deploy "conda" conda packages
```
jezdez, line 1: Since the name is used in the actions UI could we just call this "Canary" and rename the yml file to canary.yml? Just a nit but "deploy" has a particular meaning on GitHub and we're not doing that here.

conda.recipe/build.sh
```diff
@@ -2,10 +2,10 @@
 
 echo $PKG_VERSION > conda/.version
 $PYTHON setup.py install --single-version-externally-managed --record record.txt
-if [[ ! $(uname) =~ M.* ]]; then
+if [[ ! $(uname) =~ MINGW* ]]; then
   rm -rf "$SP_DIR/conda/shell/*.exe"
 fi
 $PYTHON -m conda init --install
-if [[ ! $(uname) =~ M.* ]]; then
+if [[ $(uname) =~ MINGW* ]]; then
```
jezdez, line 9: Huh, that's been broken?

.github/workflows/deploy-dev-pkgs.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          set -eu
+          conda activate
+          conda update --yes conda
+          conda install --yes conda-build anaconda-client git
+          conda info
+          conda config --show-sources
+          conda list
+          conda build --croot=./pkgs conda.recipe
+          anaconda --token ${ANACONDA_ORG_TOKEN} upload \
```
kenodegard, line 78: Since this is a secrete can we just leave it as such and then remove the `env` defined above and move the above comment down here?
```suggestion
          anaconda --token ${{ secrets.ANACONDA_ORG_TOKEN }} upload \
```

conda.recipe/meta.yaml
```diff
@@ -69,6 +63,9 @@ test:
   imports:
     - conda
     - conda_env
+  files:
+    - test_conda.bat                # [win]
+    - test_conda.sh                 # [unix]
```
jezdez, line 68: Odd, does that mean the test files weren't part of the conda package?

jezdez (approved):
Nice, just minor comments, @chenghlee!

Related but not needed to solve in this PR: Do we know what the *other* recipe folder in the root of the repo is?

.github/workflows/deploy-dev-pkgs.yml
chenghlee: Good point; how about just "Build & upload"? ("Publish", it turns out, also has a specific meaning in GitHub.)

I purposely avoided the word "Canary" since forks of this repo won't have permissions to upload packages to that channel, but this GH Action should still be usable by forks (as long as they have their `ANACONDA_*` repository secrets properly configured).

conda.recipe/build.sh
chenghlee: The not (`!`) was definitely breaking things. The `M*` wasn't broken, but it was less than clear what OS/distribution was being checked for, so I decided to expand it.

.github/workflows/deploy-dev-pkgs.yml
chenghlee: Yup, can do. The recipe doesn't depend on anything from the underlying OS, so there's no reason not to.

.github/workflows/deploy-dev-pkgs.yml
chenghlee: Done; the GHA docs are a bit ambiguous about whether the `env` block is strictly needed to properly handle secrets, but logs from the test runs seem to confirm that it's not necessary; i.e., directly passing the token as suggested isn't causing it to show up in the logs.

conda.recipe/meta.yaml
chenghlee: These are meant to be smoke tests run after the `conda` conda package is generated; I'll go ahead and remove these references since since (1) the smoke tests are already done in `run_test.*`, and (2) the files are not actually present in the recipe directory (though I'm not sure why conda-build doesn't fail because of that).

conda.recipe/run_test.sh
chenghlee: Done

conda.recipe/run_test.sh
chenghlee: Done. Also removed similar chunks of dead/commented-out code in `run_test.bat`.

conda.recipe/meta.yaml
chenghlee: Done.

jezdez (approved):
LGTM!

PR #11019, opened 2021-11-04
Remove conda-docs workflow in favor of using Read The Docs PR integration.

jezdez:
Now that we've enabled the Read The Docs pull request integration, there is no need to waste actions runners on building the docs separately. The only downside is losing insight into the build log for community users but so far the build process on Read The Docs was stable enough to not make this a blocker.


kenodegard (approved):
Looks great! 🚀

Issue #10778, opened 2021-07-08
Replace custom vendor solution with vendoring package


jezdez:
The way we currently vendor packages is done in a way that makes it harder to automate security patches. We should start using the [vendoring](https://pypi.org/project/vendoring/) for this which was extracted from pip which also vendors a number of core dependencies.

The [README for pip's vendor directory](https://github.com/pypa/pip/blob/main/src/pip/_vendor/README.rst) has some good reasons why we should follow their example.

LtDan33:
If we are going to start using [vendoring](https://pypi.org/project/vendoring/) , wouldn't there be a concern over this? : `This tool has no stability promises -- it has only one intended user: pip. There may be unannounced changes to this codebase at any time, as long as the intended user (i.e. the pip project) is prepared for those changes.`

kenodegard:
@jezdez @LtDan33 hm i hadn't thought about the `auxlib` license ([ISC](https://en.wikipedia.org/wiki/ISC_license)), how do we want to handle that? I'm wondering if that will make it a little more convoluted to go the absorbing route?

PR #11025, opened 2021-11-05
Move/absorb auxlib

kenodegard:
In seeking to do #10778 we found that `auxlib` has deviated enough from the source that we are (currently) uninterested in trying to address this discrepancy.

As a short term solution we will move `auxlib` from `conda._vendor` to `conda` itself and work to eliminate dead code and otherwise refactor/absorb it into the `conda.utils` submodule.


jezdez (approved):
Thank you @kenodegard!

PR #11028, opened 2021-11-08
Update templates/workflows to use new label formats

kenodegard:
Since https://github.com/actions/stale/issues/551 has been resolved I went ahead and enabled `labels-to-add-when-unstale` and `labels-to-remove-when-unstale` as well.


jezdez (approved):
Thanks!

PR #11029, opened 2021-11-09
Automate moving to/from boards

kenodegard:
Implements workflow to auto add new issues to the triaging project board and once the backlog label is added to move the issue from triaging to the backlog project board.

I have successfully tested this with a personal dummy repo & projects.

In working on these automations I found that these GH Actions do not appear to work with beta projects and maintainers of said actions are hesitant to dig into it since beta projects promise builtin workflows/automation. So we will just need to keep any eye on the developments and at some point we will be able to switch from old projects to the beta projects.

### DIFF
--- a/.github/workflows/boards.yml
+++ b/.github/workflows/boards.yml
@@ -25,7 +25,7 @@ jobs:
           action: delete
           project: Triaging
           column: Ready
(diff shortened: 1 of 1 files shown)


jezdez (approved):
That reminds me, we should enable dependabot for GHA workflows.

PR #11035, opened 2021-11-10
Prepare for new menuinst; non-win platforms

jaimergp:
Part of the [CEP](https://github.com/conda/ceps/pull/8) efforts.

Right now this is using `menuinst` from the experimental builds available at [`jaimergp/label/menuinst_dev`](https://anaconda.org/jaimergp/menuinst/files), which are coming from this [other PR](https://github.com/conda/menuinst/pull/91) (`cep` branch).

Summary of changes:

- [X] Remove `on_win` checks for menuinst code paths.
- [X] Use the new `menuinst` API, falling back to `menuinst._legacy` when needed.
- [X] Enable `--shortcuts` and `--no-shortcuts` CLI flags in non-windows platforms.
- [X] Add `--shortcuts-only`†‡  flag to CLI and context, which adds better support for constructor's `menu_packages` option.
- [ ] Add `menuinst=2` specific tests in the integration suite

---

† This allows us to remove the `conda-standalone constructor --mk-menus/--rm-menus` code paths, simplifying the patches on the conda-standalone project. Note this can be further simplified by improving the `conda run` support on `conda-standalone`.

‡ This can be implemented with a dual `--shortcuts` too. If shortcuts is set to `True` (bare `--shortcuts`), same behaviour is kept. If the user sets it to a list of packages, then it behaves as `shortcuts_only` (e.g. `--shortcuts=package_a,package_b`). However, this is trickier to implement (and I am not sure it's even possible without big changes to `context` / `argparse`). Hence why I opted for `--shortcuts-only` above: it is one more flag, yes, but maintains better legacy behaviour and we can see precedents for this in the `--channel-priority` expansion. Note it can be used several times like `--shortcuts-only=python --shortcuts-only=numpy`.


jaimergp:
Looks like pydantic 1.9 is not available on Windows for py310, hence the failures on that platform.

jaimergp:
@kenodegard, I am seeing some environment stacking errors strikingly similar those described in https://github.com/conda/conda/pull/11257#issuecomment-1050531320 and related work (linux integration group 1). I don't get how or why changing menuinst stuff would affect that part of the code but what do I know... do you mind sharing any pointers you might have if you have time at some point? Thank you!

jaimergp:
Well turns out you should NOT install things after `conda init --dev` because if the dependencies pull in `conda` _again_ it might undo the initialization and you are back to store-bought conda, which might break tests unexpectedly. I am surprised it didn't break more stuff, but I guess we should add a test to check this condition specifically...

jaimergp:
Continues in https://github.com/conda/conda/pull/11882 (token needed for canary builds)

PR #11040, opened 2021-11-11
Enable one-click gitpod setup for Linux development (WIP)

jaimergp:
We can try with this link while this is being built:

[Open in Gitpod](https://gitpod.io/#https://github.com/conda/conda/tree/gitpod)

* [Github.dev](https://github.dev/conda/conda/tree/gitpod) (only VS Code, no container)
* Github Codespaces: Click on `< > Code` on the top right corner, next to the title, and choose Codespaces (2nd tab).

Btw this should work locally with VS Code's Remote containers too, I think!

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.194.3/containers/python-3
+{
+	"name": "Linux / Python 3.9",
+	"image": "ghcr.io/conda/conda-ci:master-linux-python3.9",
+	"containerEnv": {
+		"RUNNING_ON_DEVCONTAINER": "1"
+	},
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"python.pythonPath": "/opt/conda/bin/python",
+		"terminal.integrated.profiles.linux": {
+			"Conda Bash": {
+				"path": "/bin/bash",
+				"args": ["--rcfile", "/home/test_user/.bashrc"]
+			}
+		},
+		"terminal.integrated.defaultProfile.linux": "Conda Bash"
+	},
+
+	"extensions": [
+		"ms-python.python",
+		"ms-python.vscode-pylance",
+		"eamodio.gitlens"
+	],
+
+	"postAttachCommand": ["/bin/bash", "/workspaces/conda/dev/docker.bashrc"],
+
+	"remoteUser": "test_user"
+}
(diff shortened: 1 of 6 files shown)


jaimergp:
Before merge we need to change the Docker image referred in `/.gitpod.yml`.

This will run a VS Code instance in your browser, with all the initialization steps done for interactive debugging. I am sure we can think of more niceties, but this is a good start for a "one-click development setup". Try it!

[Open in Gitpod](https://gitpod.io/#https://github.com/conda/conda/tree/gitpod)

jezdez:
Yeah, it's a good idea to streamline the contribution process like this, but gitpod isn't free except for 50h/month which given the way this PR is written will likely be quickly surpassed by passing contributors.

I think we should look into using [GitHub Codespaces](https://github.com/features/codespaces) to see if we get the same there since we've started to recently pay for a team organization account and the [pricing scheme for codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-codespaces) isn't per-user but billed hourly (with the ability to set spending limits).

jaimergp:
I think those 50h/month are counted per individual user, not project. For first-time contributors it might be enough to get acquainted, and then switch to the script added in #11041.

jaimergp:
That said, I am happy to provide a Codespaces config if those are configurable like Gitpod.

dbast:
Another option that can be provided via URL in the README: Just adding "1s" to the repo URL also provides a crazy fast opening vs-like editor https://github1s.com/conda/conda  ... by adding a personal GH token it is possible to write to a forked repo and create a PR, where then pre-commit + general CI runs... thought I don't currently see a way here to run tests within the editor.

jezdez:
GitHub also officially also supports https://github.dev/conda/conda for a VS Code UI, just without container as in GitHub Codespaces. Start with by pressing ``.`` (dot) in any repo!

jezdez:
> I think those 50h/month are counted per individual user, not project. For first-time contributors it might be enough to get acquainted, and then switch to the script added in #11041.

Ah, that was NOT clear from the [pricing page](https://www.gitpod.io/pricing), even after reading the FAQ, this wasn't clear.

jezdez:
> That said, I am happy to provide a Codespaces config if those are configurable like Gitpod.

Yeah, here are the docs for it: https://docs.github.com/en/codespaces/customizing-your-codespace/configuring-codespaces-for-your-project

jaimergp:
Awesome, I'll add Codespaces here as well.

jaimergp:
Codespaces should be working too now! Check PR description.

jezdez (approved):
LGTM!

PR #11041, opened 2021-11-12
Add developer guide section to docs

jaimergp:
For now I am only including the root `CONTRIBUTING.md` (great resource already!), and nesting the architecture diagrams under that category.

This PR also adds a nice-to-have script to start the CI images in development mode. One-command setup to debug CI problems on Linux, yay!


.gitignore
```diff
@@ -36,3 +36,4 @@ conda.tmp/
 *.iml
 *.ipr
 *.iws
+dev/docker.bashrc
```
jezdez, line 39: What's the use case for this file?

jezdez (changes requested):
Love this idea, a developer guide! A note about the use of Docker though:

Docker ships with compose now by default and in my experience it provides a better user experience than the low-level `docker` CLI commands (e.g. easy container start and stop) that contributors often struggle with, when they don't have much container management experience.

Instead I'd suggest to simply replace the scripts in this PR with a docker-compose.yml file (or even separate ones that have the various Python versions pre-configured, or continue to use env vars). Compose also supports nested config files, which makes handling complex local setups easier.

All that said, I'm a bit worried that we're making developing on Linux even easier and continue to ignore Windows as the main platform of end users.

@dbast Going back to our previous discussions, how do you feel about trying to come up with Docker images that use the [Nano Server](https://hub.docker.com/_/microsoft-windows-nanoserver) base images that we could both use for CI as well as local development?

jaimergp:
> Instead I'd suggest to simply replace the scripts in this PR with a docker-compose.yml file

I like this. I started with the script to provide an alternative method to the already existing `dev/start` script. I'll look into this and see if I can make it work in the same way.

> how do you feel about trying to come up with Docker images that use the Nano Server base images that we could both use for CI as well as local development?

This is great! I didn't know about those images. Sadly I can't try any because the AWS Workspaces we have do not support nested virt :( Can't help much here without a Windows box, sadly.

jaimergp:
Hm, I've tried with this simple Docker compose config:

```
services:
  devel:
    image: ghcr.io/conda/conda-ci:master-linux-python${CONDA_DOCKER_PYTHON:-3.9}
    volumes:
      - .:/opt/conda-src
    command: ["bash"]
    stdin_open: true
    tty: true
```

We can start it with:

```
$> docker compose up -d
```

And then attach to it with

```
$> docker exec -it conda_devel_1 bash
```

However this has not initialized conda in dev mode, so they have to run this once in Bash:

```
$> eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
```

Three steps to get to the same state as the script is providing now. Is this really easier for a beginner? The conda initialization might get automated in the Docker image, that I get, but still requires two steps to get to bash with keywords and flags that do not really convey much (`up`, `exec`, `-it`). 

I'd rather have them run `dev/start_docker` and forget about all this complexity, but maybe I am not using Docker compose in the best way. Happy to hear tips and tricks!

jaimergp:
Hm, I can do `docker compose run devel` too, ok. That's fair.

Now I only need to add the start up script to the Docker image.

.gitignore
jaimergp: This file is now checked in the repo because we need it on other places too.

jaimergp:
Ok I withdraw my claims, I like this :) Docs updated with how to instructions.

jezdez:
Opened https://github.com/jaimergp/conda/pull/1 with a few updates to this! 🥳

jezdez:
@jaimergp I'm working on the conda 4.11 release in #11049 and wondered if you have time to get this in shape for the release? Is there anything I can do to help?

jaimergp:
@jezdez check now :)

jezdez (commented):
Problem with this approach is that GitHub filters out iframe tags: https://github.com/conda/conda/blob/636ef0c44b6c5fd65a8fbd2f3978107169c296b5/CONTRIBUTING.md#conda-contributor-license-agreement

jezdez (approved):

PR #11051, opened 2021-11-19
`SolverStateContainer` revamp

FFY00:


FFY00:
This provides a more iron out version of the tracked container and spec map we discussed. I decided to drop the dict inheritance and just make it a normal mutable mapping to reduce complexity. If you need a dictionary, you can do `dict(spec_map)`. I had to re-implement the mixin methods in the tracked map to make sure we are logging correctly there.

I think `TrackedMap` is finished implementation-wise, `SpecMap` is half-done, the base structure is done but we need to add the missing attributes, and finally, `SolverStateContainer` is mostly just a draft, I am not sure how the final structure should look like now that we added `SpecMap`, that should take form as I start looking into using it in the code and finishing `SpecMap`.

conda/core/solve.py
```diff
@@ -2153,6 +2155,155 @@ def working_state_reset(self):
...
 
 
+class TrackedMap(MutableMapping):
+    def __init__(self, name, data=None):
+        self._data = data
+        self._name = name
+        self._logger = logging.getLogger(self.__class__.__name__)
+
+    def _set(self, key, value, *, reason=None, _level=3):
+        msg = f'{self._name}[{key}] set to `{value}`'
```
jaimergp, line 2165: I'd like to have finer debugging here. Whether the key was new or not; if it wasn't, what the value was and the new one is.

conda/core/solve.py
```diff
@@ -2153,6 +2155,155 @@ def working_state_reset(self):
...
+        return self._data.popitem(key)
+
+    def clear(self):
+        self._logger.debug(f'{self._name} was cleared')
+        self._data.clear()
+
+    def update(self, data, **kwargs):
+        if data and kwargs:
+            raise ValueError('Can only update via a data object or keyword mapping')
+        self._logger.debug(f'{self._name}[{key}] was deleted', stacklevel=2)
```
jaimergp, line 2206: I think this is a copy-paste artifact right?

conda/core/solve.py
```diff
@@ -2153,6 +2155,155 @@ def working_state_reset(self):
...
+            for item in pool:
+                yield (pool, item)
+
+
+class SolverStateContainer:
+    def __init__(
+        self,
+        requested_specs_to_add: Mapping,
+        requested_specs_to_remove: Mapping,
+        prefix_data: Mapping,
```
jaimergp, line 2269: We only need to pass the `prefix` path here. Prefix data, history and pinnings can be calculated (one) from that.

conda/core/solve.py
```diff
@@ -2153,6 +2155,155 @@ def working_state_reset(self):
         self.final_environment_specs = None
 
 
+class TrackedMap(MutableMapping):
+    def __init__(self, name, data=None):
+        self._data = data
+        self._name = name
+        self._logger = logging.getLogger(self.__class__.__name__)
```
jezdez, line 2162: I think we want to use a logger name here that includes the full import path to be able to namespace it correctly.

conda/core/solve.py
```diff
@@ -2153,6 +2155,155 @@ def working_state_reset(self):
         self.final_environment_specs = None
 
 
+class TrackedMap(MutableMapping):
```
jezdez, line 2158: This should probably be a UserDict to reduce the implementation here a notch.

conda/core/solve.py
jezdez: Good idea, the less we have to pass around the easier it'll be to evolve the code in the future.

jezdez (changes requested):
Just a few minor nits on top of what @jaimergp said.

conda/core/solve.py
FFY00: That was the plan originally, but I decided against it to simplify the implementation. If we used `UserDict`, we would still have to reimplement everything to get the correct stack associated with the log, or we would have to have some magic that automatically calculates the correct stack level for the log call. To keep things easily maintainable I opted against the stack level calculation, I think reimplementing the methods is better as it is less complex and easier to read and understand the code. Since we would be reimplementing the methods, while promising a `dict` subclass, I thought it would be better to move to `MutableMapping` and limiting the API surface. There's no specific reason here to promise a `dict`.

Anyway, the `MutableMapping` implementation would also be pretty simple if we didn't have to override the mixin methods.

jaimergp:
Superseded by https://github.com/conda/conda/pull/11084

PR #11052, opened 2021-11-19
Use the appropriate token for GHA

kenodegard:
Previously the board automation wasn't working correctly so we needed to introduce a new token.


jezdez (approved):
🚀

PR #11068, opened 2021-11-29
Automatically handle milestoned/demilestoned/commented issues

kenodegard:
Adds the following automations:

- If an issue is milestoned:
  - the `backlog` label will be removed
  - the issue will be removed from the Backlog board
  - (fail-safe) the issue will be removed from the Triaging board
- If an issue is demilestoned:
  - the `backlog` label will be added
  - the issue will be added to the Backlog board in the `Do Next` column
- If an issue is commented by the author:
  - the `pending::feedback` label will be removed
  - the `pending::support` label is added

### DIFF
--- a/.github/workflows/boards.yml
+++ b/.github/workflows/boards.yml
@@ -1,13 +1,13 @@
 name: Automate Boards
 on:
   issues:
-    types: [opened, labeled]
+    types: [opened, labeled, milestoned, demilestoned]
 jobs:
-  # move to triaging board if new
+  # move to Triaging board if new
   triaging:
     runs-on: ubuntu-latest
     steps:
-      - uses: alex-page/github-project-automation-plus@v0.8.1
+      - uses: alex-page/github-project-automation-plus@v0
         if: github.event.action == 'opened' && !contains(github.event.issue.labels.*.name, 'backlog')
         with:
           action: update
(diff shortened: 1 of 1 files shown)


.github/workflows/boards.yml
```diff
@@ -1,35 +1,86 @@
...
+  issue_comment:
+    types: [created]
 jobs:
-  # move to triaging board if new
+  # move to Triaging board if new
   triaging:
     runs-on: ubuntu-latest
     steps:
-      - uses: alex-page/github-project-automation-plus@v0.8.1
+      - uses: alex-page/github-project-automation-plus@v0
```
jezdez, line 12: Let's... keep that pin and enable Dependabot or Renovate for GitHub actions, so we get action updates as PRs. E.g. https://docs.renovatebot.com/modules/manager/github-actions/

.github/workflows/boards.yml
```diff
@@ -1,35 +1,86 @@
...
+    steps:
+      - uses: dessant/label-actions@v2
+        if: github.event.action == 'milestoned'
+        with:
+          unlabel: backlog
+          github-token: ${{ secrets.PROJECT_TOKEN }}
+      - uses: alex-page/github-project-automation-plus@v0
+        with:
+          action: delete
+          project: Backlog
```
jezdez, line 61: Should we keep those names as env vars or something to reduce the risk of breaking when we update the workflows and forget a line?

jezdez (approved):
Just some nits..

.github/workflows/boards.yml
kenodegard: ah hadn't thought of that, agreed

rchord (approved):
LGTM. :shipit:

.github/workflows/boards.yml
rchord: Like the project names? I'd lean more towards verbosity in this case. The only benefit I see is just code reuse for it and then we'd obscure the name from the actual value being used when reviewing.

I won't die on this hill, but I'd vote to keep it here.

kenodegard: I started going down this path and I have 2 concerns:
1. the variable names end up being very descriptive of its value which I'd personally feel inclined to update if any of the names were modified in the future:
   ```yaml
   env:
     BOARD_TRIAGING: Triaging
     STATE_READY: Ready
     LABEL_BACKLOG: backlog
     LABEL_FEEDBACK: pending::feedback
   ```
2. the one potentially nice win would be to store the token as an envvar which eliminates all of those duplicate `with` statements but idk if GHA is smart enough to keep those obfuscated once they become envvars?
   ```yaml
   env:
     INPUT_REPO_TOKEN: ${{ secrets.PROJECT_TOKEN }}
     INPUT_GITHUB_TOKEN: ${{ secrets.PROJECT_TOKEN }}
   ```

PR #11071, opened 2021-11-30
[Mamba POC] Turn `conda.core.solve` module into a subpackage

jaimergp:
Doing several things here

* Opening a new namespace for the solvers, `conda.core.solve`. Former `conda.core.solve` _module_ is now `conda.core.solve.classic`. 
  * Some tests import stuff with a different origin from this module (e.g. `UpdateModifier` is in `conda.base.constants` but the tests import it from here directly... we might want to change this).
* Renaming "legacy" to "classic", following @rchord's advice.
* Renaming "libsolv" to "libmamba", as discussed with @wolfv.
* Checks for the value of `context.solver_logic` are done against the `SolverLogicChoice` enum instead of raw strings.

Other than that, it's just moving things around for tidiness. The code itself is the same and all tests should pass here too.


jaimergp:
We might want to rename the `solver::libsolv` tag too, for consistency (but the branches will remain prefixed with `poc-libsolv` for... historical reasons? :D)

jaimergp:
@jezdez @chenghlee -- flake8 complains about unused imports in `conda.core.solve.classic`. Is it ok to clean those up or do we need to keep them around? I'd rather clean but just double checking.

jezdez:
> 

Hmm, yeah, can you check if those imports are either used elsewhere in the codebase or conda-build? Otherwise I'd remove them.

jezdez:
> @jezdez @chenghlee -- flake8 complains about unused imports in `conda.core.solve.classic`. Is it ok to clean those up or do we need to keep them around? I'd rather clean but just double checking.

Hmm, yeah, can you check if those imports are either used elsewhere in the codebase or conda-build? Otherwise I'd remove them.

jezdez:
> We might want to rename the `solver::libsolv` tag too, for consistency (but the branches will remain prefixed with `poc-libsolv` for... historical reasons? :D)

I've renamed the tag, but you're right the branch is harder, we CAN rename it locally and push to GitHub but would "lose" the original PR tracking it. If you feel like we'd lose too much discussion, I'd say we should keep the name, but if most of the work happened in separate PRs I think we might as well change the name. GitHub does allow changing the base branch of PRs such as this PR, to a renamed feature branch, e.g. `libmamba-solver`.

jaimergp:
I don't mind too much about the existing branches. We can create the new ones with a `libmamba-` prefix and keep the old one as it is (because the discussion is indeed important). Artifacts of our own progress! :)

jaimergp:
One test broke on Windows, possibly due to `mamba 0.19` having been released. I am expecting this on Linux too, since it's about solver behaviour, but I'd rather investigate this in a separate PR following this one.

Everything else is passing and the imports have been cleaned up.

PR #11076, opened 2021-12-02
Libmamba 0.19 investigation

jaimergp:
The new version includes a patch for channel/subdir subpriorities that fixes some tests but makes another fail. It also includes Unicode fixes that might help with our windows issues. We'll take a look here.

Changes:

* Vendored `mamba.utils` into `conda.core.solve.libmamba_utils` as a temporary workaround to try different subpriorities.
* Escape local channels more robustly.

### DIFF
--- a/conda/common/url.py
+++ b/conda/common/url.py
@@ -383,9 +383,18 @@ def remove_auth(url):
 
 
 def escape_channel_url(channel):
+    if on_win and channel.startswith("file:"):
+        channel = channel.replace("\\", "/")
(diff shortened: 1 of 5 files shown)


jaimergp:
After rebuilding libcurl with unicode support on Windows this should work again. Rerunning to see if that's the case.

jaimergp:
Ok we are back on track :)

jaimergp:
@jezdez This is ready for a quick review!

jezdez (approved):

PR #11089, opened 2021-12-07
Automate sprint

kenodegard:
This adds automations to move issues from the backlog board to the sprint board. This does undo the milestone automation introduced in #11068 as I found that even if an issue has been earmarked for a specific milestone it doesn't mean anything with regards to where in the planning process the issue resides.

The sprint automation introduced here creates a circular effect between the backlog and sprint boards, e.g.:
- adding the backlog label (or removing the sprint label) will move it to the Backlog board (and add the backlog label)
- adding the sprint label (or removing the backlog label) will move it to the Sprint board (and add the sprint label)
- adding to the Backlog board (or removing from the Sprint board) will add the backlog label (and move it to the Backlog board)
- adding to the Sprint board (or removing from the Backlog board) will add the sprint label (and move it to the Sprint board)

### DIFF
--- a/.github/workflows/boards.yml
+++ b/.github/workflows/boards.yml
@@ -1,29 +1,18 @@
 name: Automate Boards
 on:
   issues:
-    types: [opened, labeled, milestoned, demilestoned]
+    types: [opened, labeled]
   issue_comment:
     types: [created]
+  project_card:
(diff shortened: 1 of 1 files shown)


kenodegard:
Superseded by #11145

PR #11093, opened 2021-12-09
Solve launching Conda on windows with a space in the username

Hepson:
Fixe #9757

Use the temp folder/file path considering spaces for windows users.

I didn't achieve to make it work reading the file content with the `for` loop. Instead of retrieving the last file line, we can do the opposite with the first one. In the end, it shouldn't matter because it's a file with only one line.

_NB: CLA form was submitted a few hours ago._


dbast:
@anaconda-issue-bot check

kenodegard:
@anaconda-issue-bot check

conda/shell/condabin/_conda_activate.bat
```diff
@@ -38,10 +38,10 @@
 )
 @ECHO Failed to create temp directory "%TMP%\conda-<RANDOM>\" & exit /b 1
 :tmp_file_created
-@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>%UNIQUE%
+@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>"%UNIQUE%"
 @IF %ErrorLevel% NEQ 0 @EXIT /B %ErrorLevel%
-@FOR /F %%i IN (%UNIQUE%) DO @SET _TEMP_SCRIPT_PATH=%%i
-@RMDIR /S /Q %UNIQUE_DIR%
+@SET /P _TEMP_SCRIPT_PATH=<"%UNIQUE%"
+@RMDIR /S /Q "%UNIQUE_DIR%"
```
jezdez, line 44: Hey @Hepson, could you elaborate what's going on here, why you changed the part where it iterates over the list of uniques?

tests/test_activate.py
```diff
@@ -2615,6 +2615,19 @@ def test_legacy_activate_deactivate_cmd_exe(self):
...
             assert conda_shlvl == '0', conda_shlvl
 
+    @pytest.mark.skipif(not which("cmd.exe"), reason="cmd.exe not installed")
+    def test_tmp_path_with_space_cmd_exe(self):
+        with InteractiveShell("cmd.exe") as shell:
+            tmp = shell.get_env_var("TMP")
+            tmp_with_space = join(tmp, "with space")
+            shell.sendline("SET TMP={}".format(tmp_with_space))
+
+            shell.sendline("activate {}".format(dev_arg))
```
jezdez, line 2625:
```suggestion
            shell.sendline(f"SET TMP={tmp_with_space}")

            shell.sendline(f"activate {dev_arg}")
```
f-strings are great!

conda/shell/condabin/_conda_activate.bat
kenodegard: @jezdez looks like `%UNIQUE%` is the first valid `!UNIQUE_DIR!\conda.tmp` path (see line 30-37 above) so no need to iterate over it?

kenodegard:
@anaconda-issue-bot check

conda/shell/condabin/_conda_activate.bat
Hepson: @jezdez  As per my understanding, we store the location of a unique temporary script in the "UNIQUE" file (_Aka. `@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>"%UNIQUE%"`_) that we try to retrieve.

The previous loop allows retrieving the last segment of this file content. However, the space seems to be a delimiter that will conclude in retrieving a partial path. Instead, I propose to replace it by reading the first line.

This solution will only work if we have effectively just one line, but I haven't the knowledge to assume it. I've forgotten the reason since. When I made these modifications, I was more confident that we get just one entry. I think that is a proper solution based on unit tests and functional tests.

conda/shell/condabin/_conda_activate.bat
```diff
@@ -29,7 +29,7 @@
 @REM This method will not work if %TMP% contains any spaces.
 @FOR /L %%I IN (1,1,100) DO @(
     SET UNIQUE_DIR=%TMP%\conda-!RANDOM!
```
jaimergp, line 31:
```suggestion
    SET "UNIQUE_DIR=%TMP%\conda-!RANDOM!"
```

Shouldn't this be quoted too? See https://stackoverflow.com/a/67542742/3407590 for quoted `SET` instructions.

conda/shell/condabin/_conda_activate.bat
```diff
@@ -29,7 +29,7 @@
...
-    MKDIR !UNIQUE_DIR! > NUL 2>&1
+    MKDIR "!UNIQUE_DIR!" > NUL 2>&1
     IF NOT ERRORLEVEL 1 (
         SET UNIQUE=!UNIQUE_DIR!\conda.tmp
```
jaimergp, line 34:
```suggestion
        SET "UNIQUE=!UNIQUE_DIR!\conda.tmp"
```

Quotes won't hurt here either, I guess.

conda/shell/condabin/_conda_activate.bat
jaimergp: As per my understanding, `conda shell.cmd.exe activate` only returns a single line with a temporary path to the activation script:

```
$ conda shell.cmd.exe activate
/var/folders/1b/z374sygs1s96t_xxxv0hr3lr0000gn/T/tmpot5p01v9.bat                   
```

So this seems to be safe.

---

Full explanation

1. User calls `conda activate "my env"`
2. This enters the `conda` shell command, defined in `conda/shell/condabin/conda.bat`. Since we asked for `activate`, it forwards to `conda/shell/condabin/_conda_activate.bat`, where this PR is adding changes.
3. `_conda_activate.bat` [will call](https://github.com/conda/conda/blob/92cfafcd65c338240d562d7a74cf7285f2eb5670/conda/shell/condabin/_conda_activate.bat#L41) `conda shell.cmd.exe <activate|deactivate>`, which takes us to `conda.cli.main`. There, `source`'d commands are [split](https://github.com/conda/conda/blob/92cfafcd65c338240d562d7a74cf7285f2eb5670/conda/cli/main.py#L125) from `install`-like commands. 
4. `main_sourced` will delegate to the relevant activator class. Since we have `cmd.exe`, then it calls [`conda.activate.CmdExeActivator.execute()`](https://github.com/conda/conda/blob/92cfafcd65c338240d562d7a74cf7285f2eb5670/conda/activate.py#L169).
5. `.execute()` delegates to the relevant subcommand (`activate` in our case), which delegates to `_finalize()` (a method that writes a given string to stdout OR a temporary file and returns its path).
6. This temporary path is printed to STDOUT.
7. We are back to `_conda_activate.bat`, which will now `@CALL` that temporary path.

As a result, we can see that `_finalize()` will only return the temporary path on Windows, so it's a single line, so this method is safe.

conda/shell/condabin/_conda_activate.bat
```diff
@@ -38,10 +38,10 @@
 )
 @ECHO Failed to create temp directory "%TMP%\conda-<RANDOM>\" & exit /b 1
 :tmp_file_created
-@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>%UNIQUE%
+@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>"%UNIQUE%"
 @IF %ErrorLevel% NEQ 0 @EXIT /B %ErrorLevel%
-@FOR /F %%i IN (%UNIQUE%) DO @SET _TEMP_SCRIPT_PATH=%%i
-@RMDIR /S /Q %UNIQUE_DIR%
+@SET /P _TEMP_SCRIPT_PATH=<"%UNIQUE%"
```
jaimergp, line 43: If I am understanding this correctly, `<"%A_PATH_VAR%` is the CMD syntax to get input from `stdin`, and since we use `SET /P`, it takes the first (and only) line?

Ref: https://stackoverflow.com/a/40163258/3407590

jaimergp:
I added some comments to clarify some doubts, but overall I think this looks like a potential solution to the spaces-in-TMP issue!

sysrage:
Hi all. I was just dealing with this bug and found what I believe is a "cleaner" fix. Some comments:
- Adding all the quotes suggested in the comments above actually breaks things. There's no need to do this, since most of the script is already using `%~s` for Windows shortnames. 
- The only place where spaces are still an issue is the output of `@"%CONDA_EXE%" %_CE_M% %_CE_CONDA% shell.cmd.exe %* 1>%UNIQUE%`. This command writes a path inside the unique temp file, which contains spaces.
- To read this file properly, we simply add `"tokens=*"` to the `FOR` loop and it reads the entire line as a single string.

```diff
< @FOR /F %%i IN (%UNIQUE%) DO @SET _TEMP_SCRIPT_PATH=%%i
---
> @FOR /F "tokens=*" %%i IN (%UNIQUE%) DO @SET _TEMP_SCRIPT_PATH=%%i
```

PR #11121, opened 2021-12-31
Plugin system implementation

FFY00:
This got a bit delayed with Christmas, and then I fell ill, but it is here at last.

The plugin system itself is fairly simple, it essentially consists of a few definitions in a new `conda.plugins` module.
Initially, I planned to only expose the plugin manager instance to the rest of the conda codebase via the context object, but that proved insufficient right on the first plugin implementation, as the argument parsing happens before the context object is created, and the context object creation is currently inherently dependent on the argument parsing. As such, the plugin manager is also exposed via a helper function in the context module. This might limit us in the future, but that could potentially be solved by refactoring the context object a bit.
Instead of exposing the plugin manager directly, I think we might benefit of having a wrapper around it that does all the data necessary gathering, but this is not clear to me at this point, hopefully it will become clearer as we implement more plugins and get an idea of their needs.

Overall, I think the code size is pretty small, which was my goal. It demonstrates how easily plugins can be implemented in the conda codebase.

Regarding the actual subcommand plugin implementation, I did explore improving things a little bit more in regards to the `argparse` API usage and how to plug this functionality into it, but ultimately settled on just setting `epilog` instead of overwriting `print_help`. Part of the reasoning was that jezdez expressed the wish to move away from `argparse` to a different module, like `click`, so spending more time trying to improve this design did not seem worth it. For the same reason, and backwards compatibility, I decided to give plugin users full control over the argument parsing, instead of locking them into a specific API, like integrating the hook directly into `argparse`

I still need to add the tests, but I think I am reasonably happy with the end result so far.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -38,11 +38,11 @@ from ..common.configuration import (Configuration, ConfigurationLoadError, MapPa
 from ..common._os.linux import linux_get_libc_version
 from ..common.path import expand, paths_equal
 from ..common.url import has_scheme, path_to_url, split_scheme_auth_token
-from ..common.decorators import env_override
 
 from .. import CONDA_SOURCE_ROOT
 
 from .. import plugins
+from ..plugins import archspec, cuda, linux, osx, windows  # noqa: F401
 
 try:
     os.getcwd()
@@ -136,6 +136,11 @@ def ssl_verify_validation(value):
 @functools.lru_cache(maxsize=None)  # XXX: Replace with functools.cache once we drop 3.8.
 def get_plugin_manager():
     pm = pluggy.PluginManager('conda')
+    pm.register(plugins.archspec)
+    pm.register(plugins.cuda)
+    pm.register(plugins.linux)
+    pm.register(plugins.osx)
+    pm.register(plugins.windows)
     pm.add_hookspecs(plugins)
     pm.load_setuptools_entrypoints('conda')
     return pm
@@ -907,12 +912,6 @@ class Context(Configuration):
         # DANGER: This is rather slow
         info = _get_cpu_info()
         return info['flags']
-
-    @memoizedproperty
-    @env_override('CONDA_OVERRIDE_CUDA', convert_empty_to_none=True)
-    def cuda_version(self):
-        from conda.common.cuda import cuda_detect
-        return cuda_detect()
 
     @property
     def category_map(self):
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -15,6 +15,8 @@ import struct
 from contextlib import contextmanager
 from datetime import datetime
 
+import pluggy
+
 from .constants import (APP_NAME, ChannelPriority, DEFAULTS_CHANNEL_NAME, REPODATA_FN,
                         DEFAULT_AGGRESSIVE_UPDATE_PACKAGES, DEFAULT_CHANNELS,
                         DEFAULT_CHANNEL_ALIAS, DEFAULT_CUSTOM_CHANNELS, DepsModifier,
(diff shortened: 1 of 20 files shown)


conda/plugins.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import Callable, List, NamedTuple, Optional, Tuple
+
+import pluggy
+
+
+_hookspec = pluggy.HookspecMarker('conda')
+hookimp = pluggy.HookimplMarker('conda')
+
+
+class CondaSubcommand(NamedTuple):
```
jezdez, line 11: I know this would be a big dependency, but should we use attrs for this? Maybe something that would be useful elsewhere in the code base?

conda/cli/conda_argparse.py
```diff
@@ -106,6 +110,27 @@ def __init__(self, *args, **kwargs):
...
+        pm = context.get_plugin_manager()
+        self._subcommands = sorted(itertools.chain(
+            *pm.hook.conda_cli_register_subcommands()
+        ), key=operator.attrgetter('name'))
+
+        # check conflicts
+        seen = []
+        for subcommand in self._subcommands:
+            if subcommand.name in seen:
+                raise PluginError(
```
jezdez, line 122: I'm not sure if this exception text is helpful to users as written, maybe add an explanation for where the subcommand came from? Generally we should be careful about low-level information in exception messages.

jezdez (commented):
Thanks @FFY00, this is a great start, thank you! I've left a few notes, but nothing blocks this from my perspective.

conda/plugins.py
jaimergp: I have been using `pydantic` on `menuinst`, if that helps. It's still not released or even merged, so we can change it too, but that's more of a CEP decision I am guessing? What for sure is not happening is us adding both `pydantic` and `attrs` to the dependency tree.

jezdez: Yeah, that's fair, both aren't great in terms of bus factor, so I'd say we should try to go with the one that has the most likely chance to be sustainable in the next years.

conda/base/context.py
```diff
@@ -125,6 +129,13 @@ def ssl_verify_validation(value):
     return True
 
 
+def get_plugin_manager():
```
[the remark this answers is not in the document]
FFY00: Hum, sure.

conda/plugins.py
FFY00: Does conda vendor its dependencies? I don't think so, right? Even if it does, bringing new dependencies is problematic as lots of downstreams are required to unvendor those dependencies to comply to guidelines when packaging Conda. I'd highly recommend avoiding external dependencies unless absolutely necessary. A namedtuple is perfectly suitable for this use-case.
IMO this snippet should not motivate bringing any dependency. If we do decide bringing to bring in `attrs` or something like that, we could update it to use it, though even in that case I'd be inclined to not do that because this is public API and that would lock us into the dependency even if we decide to move away from it because it is no longer the best option for the code that motivated bringing it in the first place.

This would, of course, be a different story if the code did call for a functionality provided by a package like `attrs`, but in this case, it does not, and IMO that would only make the code more complex anyway :sweat_smile:

FFY00:
I have been having some issues with the tests, it seems things get cached somewhere :confused:

FFY00:
Seems like the argparse instance is cached :facepalm:

FFY00:
The latest commit (416f3482561f8d39fa43b5b007b154f447b7cdee) implement virtual packages via the plugin system. Virtual packages defined by conda are implemented as core plugins, which exist under the `conda.plugins.*` namespace and are registered directly in `conda.base.context.get_plugin_manager`, instead of via entrypoints.

jezdez:
Closing, now that #11435 has been merged into the plugins feature branch.

PR #11135, opened 2022-01-11
Review builds on PR labelling

jezdez:
This is a slight extension of our existing canary build process to allow building review builds when labelling a pull request with the `build::review` label.

The idea here is to be able to try branches out easily in real Conda environments, e.g. the libmamba work.

TODOs:

* [x] Leave a comment when the build starts with a link to the workflow run
* [x] Remove label automatically once build is finished (or when workflow run is started)
* [x] Update comment with information how to try out the review build
* [x] Only allow building for conda maintainers
* [x] Move workflow into custom local action
* [x] Move this into conda/actions maybe?

Update: this depends on https://github.com/conda/actions/pull/10.


jezdez:
That seems to have worked, except that I didn't expect that labels need to be created upfront via anaconda.org, which made using the branch name as a anaconda.org label not possible.

So unless there is an option to proactively create the label via the anaconda.org API, I'd suggest to use a new anaconda.org label called `review` and just be okay that there may be overlap between pull requests.

@chenghlee any insights into labelling on anaconda.org?

jezdez:
The working workflow with the build is: https://github.com/conda/conda/runs/4782194924

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,91 @@
...
+          ANACONDA_ORG_TOKEN: ${{ secrets.ANACONDA_ORG_TOKEN }}
+        shell: bash -l {0}
+        run: |
+          test -n "${ANACONDA_ORG_CHANNEL}" || \
+              { echo "FATAL: ANACONDA_ORG_CHANNEL secret not configured" >&2; exit 1; }
+          test -n "${ANACONDA_ORG_LABEL}" || \
+              { echo "FATAL: ANACONDA_ORG_LABEL secret not configured" >&2; exit 1; }
+          test -n "${ANACONDA_ORG_TOKEN}" || \
+              { echo "FATAL: ANACONDA_ORG_TOKEN secret not configured" >&2; exit 1; }
+          echo "::set-output name=defined::true"
```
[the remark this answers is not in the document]
jezdez: Yep, agreed! I'll get on that

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,91 @@
...
+          ref: ${{ github.ref }}
+          clean: true
+          fetch-depth: 0
+      - uses: conda-incubator/setup-miniconda@v2
+      - name: Build & upload package
+        shell: bash -l {0}
+        # Note: In the `run` commands below, `ANACONDA_ORG_CHANNEL` and
+        # `ANACONDA_ORG_LABEL` don't really _need_ to be secrets (e.g., masked
+        # out in the logs), but I don't really know of another way of
+        # performing per-repository/fork configuration for GitHub Actions.
```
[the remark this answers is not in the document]
jezdez: I can only partially remember why that was, but IIRC @chenghlee wanted to make it reusable for forks?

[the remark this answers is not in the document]
kenodegard: Hm, can we change it to include the user login as part of the label if this workflow runs on a fork? That should standardize this and eliminate complexity?

jezdez:
The last run worked better: https://github.com/conda/conda/runs/4804504769?check_suite_focus=true

But I found that slashes in labels didn't really work for actually installing from Conda :-/

jezdez:
That seems to work:

```
❯ conda install -c conda-canary/label/conda-conda-pr-11135 conda
Solving environment: done

## Package Plan ##

  environment location: /Users/jezdez/.asdf/installs/python/miniconda3-latest

  added / updated specs:
    - conda


    certifi-2021.10.8          |   py39hecd8cb5_2         152 KB
    conda-4.11.0.22_g0804cf614 |           py39_0        16.9 MB  conda-canary/label/conda-conda-pr-11135
    openssl-1.1.1m             |       hca72f7f_0         2.2 MB
                                           Total:        19.3 MB

The following packages will be UPDATED:

  certifi                          2021.10.8-py39hecd8cb5_0 --> 2021.10.8-py39hecd8cb5_2
  conda              pkgs/main::conda-4.10.3-py39hecd8cb5_0 --> conda-canary/label/conda-conda-pr-11135::conda-4.11.0.22_g0804cf614-py39_0
  openssl                                 1.1.1l-h9ed2024_0 --> 1.1.1m-hca72f7f_0


```

.github/workflows/builds-review.yaml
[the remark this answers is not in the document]
jezdez: I think this is done?

jezdez:
@kenodegard does this make sense to you?

.github/actions/canary/actions.yml
```diff
@@ -0,0 +1,43 @@
+name: 'Canary releases'
```
kenodegard: Contender for `conda/actions`?

.github/workflows/builds-canary.yaml
```diff
@@ -0,0 +1,36 @@
...
+          - runner: windows-latest
+            subdir: win-64
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          clean: true
+          fetch-depth: 0
+      - uses: conda-incubator/setup-miniconda@v2
```
kenodegard, line 31: Should we include these in the composite action as well?

kenodegard (approved):
I am absolutely a fan of this change, very sleek 🚀

kenodegard:
@jezdez is `build::review` appropriately descriptive? `build::canary` would perhaps be clear? in which case we could see a future where we'd also have labels like `build::alpha`/`build::beta`/`build::candidate`

jezdez:
> @jezdez is `build::review` appropriately descriptive? `build::canary` would perhaps be clear? in which case we could see a future where we'd also have labels like `build::alpha`/`build::beta`/`build::candidate`

That's a good point, totally agreed.

The custom action seems to work, just like that: https://github.com/conda/conda/runs/4806814209?check_suite_focus=true

.github/actions/canary/actions.yml
jezdez: I think so? I'm not sure if it's useful for other projects generally, but then again maybe that's okay. Being able to install conda-build easily would be beneficial as well.

.github/workflows/builds-canary.yaml
jezdez: Ah, yeah, good call, this makes sense, since we're expecting a conda environment might as well as creating it there. 👍🏻

.github/actions/canary/actions.yml
kenodegard: This could be something that would be useful for `conda-build`? Or other plugins as that becomes more of a thing?

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,49 @@
...
+            subdir: osx-64
+          - runner: windows-latest
+            subdir: win-64
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          clean: true
+          fetch-depth: 0
```
kenodegard, line 28: Why 0? AFAIK this means that the entire repo history (for all branches) is cloned, is this necessary?

jezdez: I'm not sure, this comes from the previous canary action, the last commit should suffice, right?

kenodegard: ah yeah for both of these workflows I'm pretty sure the defaults for the `actions/checkout` action are all we need

jezdez:
@kenodegard Not sure where the Windows errors come from suddenly, tbh. But this depends on this PR now: https://github.com/conda/actions/pull/10

jezdez:
@kenodegard I think I'm done trying to get this working better, using a different github action proved to be a better way to implement this. Mind taking another look?

.github/actions/canary/action.yml
```diff
@@ -0,0 +1,113 @@
+name: 'Canary releases'
```
kenodegard, line 1: This can be removed now right?

.github/workflows/builds-review.yaml
kenodegard: @jezdez I still think we can remove the entire `with` clause from both of these workflows as the defaults should be sufficient

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+      # Clean checkout of specific git ref needed for package metadata version
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          clean: true
+          fetch-depth: 0
+
+      - name: Create and upload review build
+        uses: conda/actions/canary-release@canary-release
```
kenodegard, line 53: We need to remember to update this after merging https://github.com/conda/actions/pull/10

kenodegard:
@jezdez Hm weird that it only appended the status for macOS and Windows but not Linux?

jezdez:
@kenodegard 

>  I still think we can remove the entire with clause from both of these workflows as the defaults should be sufficient

That didn't work when I tried it, the specific call sets the env vars needed for building the version in https://github.com/conda/conda/blob/db4e1c0f2c1a8321a58679816a8ad6eb87993706/conda.recipe/meta.yaml#L3.

kenodegard:
@jezdez ooooh bc git describe needs the git history to work, gotcha, can we document this detail in the action `README.md` and in these workflows? Avoid anyone making a "smart" cleanup

jezdez:
> @jezdez Hm weird that it only appended the status for macOS and Windows but not Linux?

Totally, Linux was the first run, wonder if that was related?

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,62 @@
...
+on:
+  pull_request:
+    types:
+      - labeled
+
+jobs:
+  build:
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.label.name == 'build::review'
```
kenodegard, line 12: Before we forget!

```suggestion
      github.event.label.name == 'build::canary'
```

jezdez: Oh right, so I was wondering if we should define canary builds as those that are created when a branch is merged into the main branch, and "review build" if we want to review a pull request. That's why I thought we should use "build::review" for that workflow. Too convoluted?

kenodegard: Ah I see, fair enough for me, makes it easier to outline in our docs with clear terminology like this.

.github/workflows/builds-review.yaml
```diff
@@ -0,0 +1,71 @@
...
+            subdir: win-64
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - name: Check current actor
+        id: in_team
+        uses: conda/actions/user-in-team
+        with:
+          org: conda
+          team: conda-core
+          user: ${{ github.actor }}
```
jezdez, line 30: @kenodegard Does that look right?

kenodegard: Yup - just need to add the new release ref
```yml
uses: conda/actions/user-in-team@v22.1.2
```

jezdez: Super good, update comming..

kenodegard (approved):

Issue #10972, opened 2021-10-06
`conda run` fails at quoting command parameters


t-kalinowski:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help
- [X] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?

I attempt to run this command:
```bash
/home/tomasz/.local/share/r-miniconda/bin/conda run --name tf-2.5-cpu python -m pip install --upgrade 'tensorflow-cpu==2.5.*' tensorflow-hub scipy requests pyyaml 'Pillow<8.3' 'h5py' pandas
```
I get this output:
```
ERROR conda.cli.main_run:execute(33): Subprocess for 'conda run ['python', '-m', 'pip', 'install', '--upgrade', 'tensorflow-cpu==2.5.*', 'tensorflow-hub', 'scipy', 'requests', 'pyyaml', 'Pillow<8.3', 'h5py', 'pandas']' command failed.  (See above for error)
/home/tomasz/.local/share/r-miniconda/envs/tf-2.5-cpu/.tmpdwyu00x3: line 3: 8.3: No such file or directory

```
Adding `--no-capture-output` does not help.
The bare pip command that doesn't go through conda run succeeds:
```
/home/tomasz/.local/share/r-miniconda/envs/tf-2.5-cpu/bin/python -m pip install --upgrade 'tensorflow-cpu==2.5.*' tensorflow-hub scipy requests pyyaml 'Pillow<8.3' 'h5py' pandas
```

This is a very vanilla miniconda installation
### Conda info

```shell
active environment : None
       user config file : /home/tomasz/.condarc
 populated config files : 
          conda version : 4.10.3
    conda-build version : not installed
         python version : 3.9.5.final.0
       virtual packages : __cuda=11.2=0
                          __linux=5.11.0=0
                          __glibc=2.31=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /home/tomasz/.local/share/r-miniconda  (writable)
      conda av data dir : /home/tomasz/.local/share/r-miniconda/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
[...]

PR #11168, opened 2022-01-26
Rework `ensure_comspec_set` into `get_comspec`

kenodegard:
Better envvar handling and simplify usage.

Working towards #10972


jezdez (approved):
Nice refactor! This would be worth a news item I think?

PR #11172, opened 2022-01-28
Refactor `is_valid_prefix` into `validate_prefix`

kenodegard:
Another cleanup. This function is also unused in conda-build.

The big change with this function is adding the prefix as a passthrough which helps streamline usage.

Working towards #10972


jezdez (approved):
Thanks for adding the docstring parameters!

beeankha (approved):
🚢

PR #11179, opened 2022-01-31
Remove unnecessary `on_win` arg for `conda.utils.wrap_subprocess_call`

kenodegard:
At no point does `on_win` deviate from whats detected from the current system so its unnecessary to explicitly pass this in.

If anything the `if on_win` clauses in `conda.utils.wrap_subprocess_call` should simply be moved outside the function but that's out of scope for this change.

### DIFF
--- a/conda/gateways/subprocess.py
+++ b/conda/gateways/subprocess.py
@@ -16,8 +16,13 @@ from ..utils import wrap_subprocess_call
 from .logging import TRACE
 from .. import ACTIVE_SUBPROCESSES
 from ..auxlib.ish import dals
-from ..common.compat import (ensure_binary, string_types, encode_arguments,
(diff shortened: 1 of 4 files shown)


conda/gateways/subprocess.py
```diff
@@ -39,11 +44,20 @@ def _format_output(command_str, cwd, rc, stdout, stderr):
...
+        args,
+    )
+    process = Popen(
+        command_args,
+        cwd=cwd or prefix,
+        universal_newlines=False,
+        stdout=PIPE,
+        stderr=PIPE,
+        env=env,
+    )
```
beeankha, line 60: Yay for improved readability!

beeankha (approved):
👍

PR #11181, opened 2022-01-31
[libmamba] Refactor integration test helpers

jaimergp:
We'll need these too downstream, so let's refactor a bit. I wouldn't pay much attention to `conda/testing/integration.py` because that's essentially a copy-pasted chunk and it needs plenty of work. For now, just check that the new imports are ok and tests still pass.


jaimergp:
@jezdez this is ready now. The Windows errors are still the ones in `master`.

jezdez (approved):
LGTM!

PR #11182, opened 2022-01-31
Introduce `CONDA_SITE_PACKAGES`

kenodegard:
`dirname(CONDA_PACKAGE_ROOT)` is used a number of times in various corners of `conda`. Introducing a new global variable eliminating any repeat `dirname` calls.

Not set on the variable name or the necessity of this change so def up for a discussion.

### DIFF
--- a/conda/__init__.py
+++ b/conda/__init__.py
@@ -41,8 +41,13 @@ __url__ = "https://github.com/conda/conda"
 if os.getenv('CONDA_ROOT') is None:
     os.environ[str('CONDA_ROOT')] = sys.prefix
 
+#: The conda package directory.
 CONDA_PACKAGE_ROOT = abspath(dirname(__file__))
-CONDA_SITE_PACKAGES = dirname(CONDA_PACKAGE_ROOT)
+#: The path within which to find the conda package.
+#:
+#: If `conda` is statically installed this is the site-packages. If `conda` is an editable install
+#: or otherwise uninstalled this is the git repo.
+CONDA_SOURCE_ROOT = dirname(CONDA_PACKAGE_ROOT)
 
 def another_to_unicode(val):
     # ignore flake8 on this because it finds this as an error on py3 even though it is guarded
--- a/conda/__init__.py
+++ b/conda/__init__.py
@@ -42,6 +42,7 @@ if os.getenv('CONDA_ROOT') is None:
     os.environ[str('CONDA_ROOT')] = sys.prefix
 
 CONDA_PACKAGE_ROOT = abspath(dirname(__file__))
+CONDA_SITE_PACKAGES = dirname(CONDA_PACKAGE_ROOT)
 
 def another_to_unicode(val):
     # ignore flake8 on this because it finds this as an error on py3 even though it is guarded
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -15,7 +15,7 @@ from textwrap import dedent
 
 # Since we have to have configuration context here, anything imported by
 #   conda.base.context is fair game, but nothing more.
-from . import CONDA_PACKAGE_ROOT, CondaError
+from . import CONDA_PACKAGE_ROOT, CONDA_SITE_PACKAGES, CondaError
 from ._vendor.toolz import concatv, drop
 from .auxlib.compat import Utf8NamedTemporaryFile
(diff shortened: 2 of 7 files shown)


beeankha (approved):
This change makes sense to me; thanks also for cleaning up formatting/improving readability!

conda/activate.py
```diff
@@ -1048,26 +1059,28 @@ def __init__(self, arguments=None):
...
+                f"""
                 $Env:CONDA_EXE = "{context.conda_exe}"
                 $Env:_CE_M = ""
                 $Env:_CE_CONDA = ""
                 $Env:_CONDA_ROOT = "{context.conda_prefix}"
                 $Env:_CONDA_EXE = "{context.conda_exe}"
                 $CondaModuleArgs = @{{ChangePs1 = ${context.changeps1}}}
-                """.format(context=context))
+                """
+            ).strip
```
jezdez, line 1083:
```suggestion
            ).strip()
```

conda/base/context.py
```diff
@@ -613,12 +613,18 @@ def conda_exe_vars_dict(self):
...
-                                    os.pathsep, os.environ.get('PYTHONPATH', ''))),
-                                ('_CE_M', '-m'),
-                                ('_CE_CONDA', 'conda'),
-                                ('CONDA_PYTHON_EXE', sys.executable)])
+            return OrderedDict(
+                [
+                    ("CONDA_EXE", sys.executable),
+                    (
+                        "PYTHONPATH",
+                        os.pathsep.join((CONDA_SITE_PACKAGES, os.environ.get("PYTHONPATH", ""))),
```
jezdez, line 621: This is a bit weird since there is a very similar main API of `os.path.join(path, components, etc)` that has an uncanny similarity to this way of writing it. Could we use that instead or add a comment why we're not using it?

conda/__init__.py
```diff
@@ -42,6 +42,7 @@
     os.environ[str('CONDA_ROOT')] = sys.prefix
 
 CONDA_PACKAGE_ROOT = abspath(dirname(__file__))
+CONDA_SITE_PACKAGES = dirname(CONDA_PACKAGE_ROOT)
```
jezdez, line 45: "site-packages" has a significant meaning in Python packaging as the default place where packages are installed in, which makes me wonder if we should try to prevent using it in this case. Somewhere in this PR I saw the name "CONDA_SOURCE_ROOT" which may be better suited and prevents the overlap in meaning?

jezdez (approved):
Just a few nits and a question about naming the variable, LGTM otherwise!

conda/base/context.py
kenodegard: So this is using `os.pathsep` (e.g. `;` and `:`) not `os.sep` (e.g. '/' and '\') so I don't think there's any way to use `os.path.join` here?

kenodegard: But yes a comment highlighting this special case here would be good.

conda/__init__.py
kenodegard: Good call

conda/__init__.py
```diff
@@ -41,7 +41,13 @@
 if os.getenv('CONDA_ROOT') is None:
     os.environ[str('CONDA_ROOT')] = sys.prefix
 
+#: The conda package directory.
 CONDA_PACKAGE_ROOT = abspath(dirname(__file__))
+#: The path within which to find the conda package.
+#:
+#: If `conda` is statically installed this is the site-packages. If `conda` is an editable install
+#: or otherwise uninstalled this is the git repo.
+CONDA_SOURCE_ROOT = dirname(CONDA_PACKAGE_ROOT)
```
kenodegard, line 50: @jezdez @beeankha do either of you have experience defining docs for consts? Outside of defining them all at the top in the module docstring using the autodoc extension for sphinx seems to be the best option? Looks like urllib uses this.

beeankha: I don't have direct experience defining docs for consts, but after doing some poking around/research, it appears that the sphinx option is indeed best! 👍

jezdez: Yeah, autodoc is the way to go 👍🏻

conda/base/context.py
jezdez: Oh, d'oh, you're right, I misread the line! Thanks for the context :)

PR #11187, opened 2022-02-01
[libmamba] Error out if libmamba is chosen but unavailable

jaimergp:
Instead of falling back to the classic solver, explicitly error out.


jezdez (approved):
Sounds good to me!

PR #11189, opened 2022-02-01
Improve command line argument quoting

kenodegard:
Not in love with this implementation so alternative ideas would be very welcome.

We have already been relying on `subprocess.list2cmdline` to convert our list of str into a str on Windows. Unfortunately the implementation does not identify and handle what may look like redirects, file descriptors, or pipes. This resulted in:

<table>
<tr>
<td>End user input</td>
<td><code>conda run -n NAME pip install "numpy<1.22"</code></td>
</tr>
<tr>
<td>In <code>.tmp</code> script</td>
<td><code>pip install numpy<1.22</code></td>
</tr>
<tr>
<td>Expected</td>
<td><code>pip install "numpy<1.22"</code></td>
</tr>
</table>

So in addition to handling quotes and escape chars as `subprocess.list2cmd` does we also ensure we quote anything that looks like a redirect (`<` or `>`), a file descriptors (`&1` or `&2`), or a pipe (`|`).

This is the magical PR that solves #10972

### DIFF
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -263,9 +263,8 @@ def sys_prefix_unfollowed():
 def quote_for_shell(*arguments):
     """Properly quote arguments for command line passing.
 
-    For POSIX uses `shlex.join`, for Windows uses an implementation derived from `shlex.join` and
-    guided by
-    https://docs.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way#a-better-method-of-quoting.
+    For POSIX uses `shlex.join`, for Windows uses a custom implementation to properly escape
+    metacharacters.
 
     :param arguments: Arguments to quote.
     :type arguments: list of str
@@ -280,6 +279,9 @@ def quote_for_shell(*arguments):
 
 
 if on_win:
+    # https://ss64.com/nt/syntax-esc.html
+    # https://docs.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way
+
     # cmd metachars
     _CMD_ESCAPE_METACHARS = "()%!^<>&|"
     _CMD_DBLESCAPE_METACHARS = '"'
@@ -290,28 +292,21 @@ if on_win:
     _CMD_DBLESCAPE_RE = re.compile(fr"([{_CMD_DBLESCAPE_METACHARS}])")
 
     def _args_join(args):
-        """Return a shell-escaped string from a list of arguments.
+        """Return a shell-escaped string from *args*."""
 
-        Inspired by `shlex.join`.
-        """
-        return " ".join(map(_cmd_quote, args))
+        def quote(s):
+            # derived from shlex.quote
+            if not s:
+                return '""'
+            if not _CMD_UNSAFE_RE.search(s):
+                return s
+            # escape (^) metacharacters
+            s = _CMD_ESCAPE_RE.sub(r"^\1", s)
+            # doubly escape (\\^) metacharacters to avoid case of, e.g., quotes inside quoted arg
+            s = _CMD_DBLESCAPE_RE.sub(r"\\^\1", s)
+            return f'^"{s}^"'
 
-    def _cmd_quote(s):
-        """Return a shell-escaped version of the string.
-
-        Inspired by `shlex.quote`.
-        """
-        if not s:
-            return '""'
-        if not _CMD_UNSAFE_RE.search(s):
-            return s
-        # escape (^) metacharacters
-        s = _CMD_ESCAPE_RE.sub(r"^\1", s)
-        # doubly escape (\\^) metacharacters to avoid case of, e.g., quotes inside quoted arg
-        s = _CMD_DBLESCAPE_RE.sub(r"\\^\1", s)
-        return f'^"{s}^"'
-
-
+        return " ".join(quote(arg) for arg in args)
 else:
     try:
         from shlex import join as _args_join
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -282,14 +282,8 @@ if on_win:
     # https://ss64.com/nt/syntax-esc.html
     # https://docs.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way
 
-    # cmd metachars
-    _CMD_ESCAPE_METACHARS = "()%!^<>&|"
-    _CMD_DBLESCAPE_METACHARS = '"'
-
-    _CMD_UNSAFE_RE = re.compile(fr"[\s{_CMD_ESCAPE_METACHARS}{_CMD_DBLESCAPE_METACHARS}]")
-
-    _CMD_ESCAPE_RE = re.compile(fr"([{_CMD_ESCAPE_METACHARS}])")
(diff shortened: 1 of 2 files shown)


conda/utils.py
```diff
@@ -260,35 +260,71 @@ def sys_prefix_unfollowed():
...
+            or (">" in arg)
+            # file descriptor
+            or ("&" in arg)
+            # pipes
+            or ("|" in arg)
+            # spacing
+            or (" " in arg)
+            or ("\t" in arg)
+            or not arg
+        )
```
kenodegard, line 299: Previously `conda.utils.quote_for_shell` was quoting an argument that contained a newline. This is no longer the case. I haven't fully brainstormed my way through the pros and cons of this yet.

kenodegard: But I _think_ that needs to be added back...

kenodegard: And I added back quoting for newlines.

kenodegard:
`shlex.join` appears to accomplish the same thing except it uses single quotes instead of double quotes.

jezdez:
> [`shlex.join`](https://github.com/python/cpython/blob/e5e1441d41907f92cc3bb5de675a2c519068173d/Lib/shlex.py#L318-L334) appears to accomplish the same thing except it uses single quotes instead of double quotes:
> 
> ```python
> def join(split_command):
>     """Return a shell-escaped string from *split_command*."""
>     return ' '.join(quote(arg) for arg in split_command)
> 
> 
> _find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search
> 
> def quote(s):
>     """Return a shell-escaped version of the string *s*."""
>     if not s:
>         return "''"
>     if _find_unsafe(s) is None:
>         return s
> 
>     # use single quotes, and put single quotes into double quotes
>     # the string $'b is then quoted as '$'"'"'b'
>     return "'" + s.replace("'", "'\"'\"'") + "'"
> ```
> 
> Obviously the problem with using `shelx.join` is that it isn't Windows compliant.

Interesting, the comment in the Python docs is:

> Consider using functions that pass command arguments with lists such as subprocess.run() with shell=False.

Not sure if [subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run) is a help here though.

In the [shlex.split](https://docs.python.org/3/library/shlex.html#shlex.split) docs is an interesting note:

> This function operates in POSIX mode by default, but uses non-POSIX mode if the posix argument is false.

Used in Ansible for example: https://github.com/ansible/ansible/blob/41db6d8d35900d425df3228406db3fec61ab2269/lib/ansible/plugins/shell/powershell.py#L194-L195

jezdez:
Oh, and quoting in [Ansible for cmd](https://github.com/ansible/ansible/blob/41db6d8d35900d425df3228406db3fec61ab2269/lib/ansible/plugins/shell/cmd.py#L37-L57)?

kenodegard:
Yeah we can't use `subprocess.run` since we are generating a script to execute. `subprocess.run` on windows uses `list2cmdlist` and runs it as a script but on posix systems it uses the `posix_spawn` C function where it accepts the list of args.

Hm yeah the ansible command is _very_ similar to `shlex.join`. Perhaps we should start with a duplicate of `shlex.join` as well and then as more use cases are identified we can handle those?

jaimergp:
I think I have something [similar](https://github.com/conda/menuinst/pull/91/files#diff-58d58b9193027a652e314f90419a20fea7240dca32c86f065488901a8920c0dbR67-R115) on the new menuinst. It would be awesome to have this in a single place. I think that code also falls in the `<` trap.

jezdez:
here's the [Ansible version of the quote function](https://github.com/ansible/ansible/blob/41db6d8d35900d425df3228406db3fec61ab2269/lib/ansible/plugins/shell/cmd.py#L37-L57) for cmd.exe:

```python
# these are the metachars that have a special meaning in cmd that we want to escape when quoting
_find_unsafe = re.compile(r'[\s\(\)\%\!^\"\<\>\&\|]').search

def quote(s):
    # cmd does not support single quotes that the shlex_quote uses. We need to override the quoting behaviour to
    # better match cmd.exe.
    # https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

    # Return an empty argument
    if not s:
        return '""'

    if _find_unsafe(s) is None:
        return s

    # Escape the metachars as we are quoting the string to stop cmd from interpreting that metachar. For example
    # 'file &whoami.exe' would result in 'file $(whoami.exe)' instead of the literal string
    # https://stackoverflow.com/questions/3411771/multiple-character-replace-with-python
    for c in '^()%!"<>&|':  # '^' must be the first char that we scan and replace
        if c in s:
            # I can't find any docs that explicitly say this but to escape ", it needs to be prefixed with \^.
            s = s.replace(c, ("\\^" if c == '"' else "^") + c)

    return '^"' + s + '^"'
```

Here's the [shlex version of quote](https://github.com/python/cpython/blob/b5527688aae11d0b5af58176267a9943576e71e5/Lib/shlex.py#L323-L334):

```python
_find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search

def quote(s):
    """Return a shell-escaped version of the string *s*."""
    if not s:
        return "''"
    if _find_unsafe(s) is None:
        return s

    # use single quotes, and put single quotes into double quotes
    # the string $'b is then quoted as '$'"'"'b'
    return "'" + s.replace("'", "'\"'\"'") + "'"
```

jezdez:
> @jezdez Interesting how the Ansible implementation is using a character set but `shlex` uses a _negated_ character set, ultimately they are almost identical matches. The Ansible implementation is clearly cmd specific and it feels like overkill (especially compared with `subprocess.list2cmdline`)?

Yeah, let's just say that Ansible's implementation also can be more narrower focused on cmd.exe alone, while we're trying to cover more shells.

I agree with your thinking that if a) it's good enough for subprocess it's probably good enough for conda b) I have a hard top trying to cover all the edge cases without test cases for them. #10972 seems to have *some* but I feel like we'd need more to make an educated decision? Or is this distracting more to get a fix in?

> 
> > ```python
> > # cmd does not support single quotes that the shlex_quote uses. We need to override the quoting behaviour to
> > # better match cmd.exe.
> > # https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/
> > ```
> 
> This is very insightful, immediately rules out simply using `shlex` for both sh & cmd.

Yeah, indeed. So that leaves your list2cmdline implementation?

jezdez:
Another case in which we see an implementation in the wild that is the same as in subprocess: https://github.com/mahmoud/boltons/blob/395f690f4a24331c4554e2169ac18a15955a4eab/boltons/strutils.py#L769

jezdez (commented):
Python 3.7 doesn't have `shlex.join` yet.

jezdez:
@kenodegard Oh sorry, I missed that the devenv commits were in this branch as well and not only the quoting related ones when I rebased it with the rebase button 😬  Definitely ping me when I should take another look please, just wanted to check if the tests pass now that master works again. Sorry!

kenodegard:
No worries, I found the issue with the failing test so I should have this wrapped up later today

tests/test_utils.py
```diff
@@ -14,6 +14,9 @@
 import sys
 from conda.common.compat import on_win
 
+import pytest
+import subprocess
```
jezdez, line 18:
```suggestion
```

jezdez (approved):
Just a minor nit, but looks good otherwise, let's do it!

PR #11237, opened 2022-02-20
Fix conda environment variable echo, update example versions

manics:
The `echo $my_var` example was missing the `$`, and I added the Windows equivalent.
I also took the opportunity to update the JupyterLab version in the example environment command since 0.35 is very old!


kenodegard:
@anaconda-issue-bot check

PR #11243, opened 2022-02-21
Don't make devenv dir when running `dev/start --dry-run`

kenodegard:
When invoking `dev/start{,.bat}` as dry-run it shouldn't be making **any** system changes including any new directories.


jezdez (approved):
Good catch, you're right!

PR #11249, opened 2022-02-22
Define messages.yml for conda/infra#508 & conda/infra#509

kenodegard:
Defining `.github/messages.yml` as necessary for the changes proposed in conda/infra#508 & conda/infra#509.

@jezdez I haven't seen many users actually providing the information we are asking for in these blurbs (besides saying something to the effect of "this is still an issue"), perhaps we should trim the message down further? If we make the messages generic we could define them centrally in [conda/infra](https://github.com/conda/infra) instead of requiring `.github/messages.yml` to be defined for every repo.

### DIFF
--- a/.github/messages.yml
+++ b/.github/messages.yml
@@ -1,0 +1,65 @@
+stale-issue: |
+  Hi there, thank you for your contribution to Conda!
+
+  This issue has been **automatically marked as stale** because it has not had recent activity. It will be closed automatically if no further activity occurs.
+
+  If you would like this issue to remain open please:
+
+    1. Verify that you can still reproduce the issue in the latest version of Conda
+    2. Comment that the issue is still reproducible and include:
+      - What version of Conda you reproduced the issue on
+      - What OS and version you reproduced the issue on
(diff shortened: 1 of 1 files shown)


jezdez:
> Defining `.github/messages.yml` as necessary for the changes proposed in [conda/infra#508](https://github.com/conda/infra/pull/508) & [conda/infra#509](https://github.com/conda/infra/pull/509).
> 
> @jezdez I haven't seen many users actually providing the information we are asking for in these blurbs (besides saying something to the effect of "this is still an issue"), perhaps we should trim the message down further? If we make the messages generic we could define them centrally in [conda/infra](https://github.com/conda/infra) instead of requiring `.github/messages.yml` to be defined for every repo.

Yeah, that's fine by me, let's word-smith the message a bit, I want to make sure the tone is right. But in doubt I'd rather be a bit more verbose for the people that feel insulted by the stale/lock process. Being to brief has the downside of leaving too much room for speculation why we're doing this and asking detailed questions allows us to signal that we care about the issues and just need to "clean house" those that aren't useful anymore because of age etc.

PR #11255, opened 2022-02-23
Do not auto activate base before activating devenv

### LINKED ISSUE #11252: `devenv`: Install test dependency in base environment, not per-Python version environment
### Checklist

- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help
- [X] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?

While trying to write a test case for #11174 I noticed that the changes to the development setup done in #11233 contain a major regression I missed in the code review: the test dependencies [aren't installed in the base environment](https://github.com/conda/conda/pull/11233/files#diff-0da549fc3fb0b16c808b4c8172a4d14bf0ab79646b097db802395e88fe62fb1fR181) anymore but instead in a separate sub-environment (I think for ease of use).

I only now realize that this is counter-productive for dev/CI parity (since in CI we are also running the tests from the base environment) and prevents us from running tests that start at the lowest environment stack level (at least with the way the integration tests are written at the moment).

Please roll back the change to the dev environment again while we're in the run-up to the 4.12.0 release so we can consider alternatives again.

I think a way forward mid-term would be to simplify the new start script again to not create a separate environment that gets activated.

### Conda info

```shell
-
```

### Conda config

```shell
-
```

### Conda list

```shell
-
```

### Additional Context

_No response_

kenodegard:
Resolves #11252

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -132,15 +132,15 @@ def initialize_dev(shell, dev_env_prefix=None, conda_source_root=None):
     # > alias conda-dev='eval "$(python -m conda init --dev)"'
     # > eval "$(python -m conda init --dev)"
 
-    dev_env_prefix = expand(dev_env_prefix or sys.prefix)
+    prefix = expand(dev_env_prefix or sys.prefix)
     conda_source_root = expand(conda_source_root or os.getcwd())
 
-    python_exe, python_version, site_packages_dir = _get_python_info(dev_env_prefix)
+    python_exe, python_version, site_packages_dir = _get_python_info(prefix)
 
     if not isfile(join(conda_source_root, 'conda', '__main__.py')):
         raise CondaValueError("Directory is not a conda source root: %s" % conda_source_root)
 
-    plan = make_install_plan(dev_env_prefix)
+    plan = make_install_plan(prefix)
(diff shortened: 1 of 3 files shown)


conda/core/initialize.py
```diff
@@ -201,10 +201,9 @@ def initialize_dev(shell, dev_env_prefix=None, conda_source_root=None):
         sys_executable = abspath(sys.executable)
         if on_win:
             sys_executable = "$(cygpath '%s')" % sys_executable
-        builder += [
-            "eval \"$(\"%s\" -m conda \"shell.bash\" \"hook\")\"" % sys_executable,
-            "conda activate '%s'" % dev_env_prefix,
-        ]
+        builder += [f'eval "$("{sys_executable}" -m conda "shell.bash" "hook")"']
+        if context.auto_activate_base:
+            builder += [f"conda activate '{prefix}'"]
```
jezdez, line 206: This is smart, I don't think that should have been auto-activated before!

jezdez (approved):
Thanks for the quick dive-in Ken

PR #11193, opened 2022-02-02
Add support for libmamba integrations (clean up `poc-libsolv`)

jaimergp:
Supersedes #10881 

---

We will cleanup the long-lived `poc-libsolv` branch here to provide a clearer diff with no formatting:

Implemented changes (wip):

- [X] Add mamba dependencies to dev requirements (incl. Docker images) (dd847c2)
- [X] Add `solver_logic` Enum and `_get_solver_logic()` forwarder (09c10b9)
- [X] Tune `UnsatisfiableError` exceptions  (082ff64)
- [x] Rework logging to allow file logs with debug level (0808ce2, e3bab36)
- [x] Rework testing infrastructure to support downstream packages (includes fixes to original helpers for local channel serialization instead of in-memory cache override) (954e4a0, fa4d5b8, b55085b, 86dafe3, 0c72113, 0f9eb57, a255c1d)

Test changes:

- [x] change timestamp in test data so it's one SECOND older (not just one ms) (6a5f3aa)
- [x] Change expected logging levels (part of 0808ce2)

Other small fixes:
- [X] Fix URL encoding / decoding (otherwise mamba fails to load local channels) (e7a4d03)
- [X] add `fresh_context` contextmanager (needed in some downstream tests) (a025443)
- [X] resolve: move specs conversion before logging (the logger is enabled for debug for our file logs, so this subtle bug is uncovered) (0df41a4)
- [x] add context.override contextmanager (c8ffe7a)
- [x] add __str__ to auxlib.Null for cleaner depiction in logs (c0c2118)
- [x] default to REPODATA_FN is passed argument is falsey (1ad4cfb)
- [x] fix logging typo (unimportant) (29372e5)
- [X] temporary workaround for `.conda_build_form()` deficiencies described in #11200 (f26c643)

PR #11263, opened 2022-02-25
Update reload from import deprecated since py=3.4

erykoff:
The `from imp import reload` used in the vendored `five.py` has been deprecated since python 3.4.  This creates warnings that can be seen in third-party tests and the conda CI unit tests.  This PR fixes the import line.

### DIFF
--- a/conda/auxlib/_vendor/five.py
+++ b/conda/auxlib/_vendor/five.py
@@ -46,7 +46,7 @@ PY3 = sys.version_info[0] == 3
 try:
     reload = reload                         # noqa
 except NameError:                           # pragma: no cover
-    from imp import reload                  # noqa
+    from importlib import reload            # noqa
 
 try:
     from collections import UserList        # noqa


jezdez:
Thanks again for a cleanup PR! This is *really* appreciated.

That said, and I understand how frustrating that must be, the real fix is to either update the vendored auxlib library or preferable stop using it altogether (since its author is no longer a conda maintainer).

There is also the larger issue with the huge amount of Python 2 specific code that will need to be removed/cleaned up, too. Given the large amount code changes that are needed for adding the libmamba solver (which is is coming to an end at the moment, #10881, #11193 etc), I decided a few months ago to wait on large refactoring like that to reduce the risk to rolling out core changes like that.

So for now this PR can't be merged and also should be replaced with a more complete cleanup. I hope you understand :)

timj:
This warning is appearing every day and is pretty much the only warning coming out of our codebase (we use the conda API to obtain package information). I understand that the real fix is to remove all the python <3.3 compatibility code but is there a downside in accepting this one line change to `five.py`. I realize this is vendored code but it seems like it's vendored code that is not being developed anyhow.

You say this PR should be replaced with a more complete cleanup. If we were to embark on the bigger problem of stripping out all the compatibility code from `conda` how would we do that? I think you are implying that that would be blocked for now by that other work you are doing. Is there an issue that we can follow to learn when work on removing `five.py` and `six.py` (and all the other version < 3.3 code) can begin? If a third-party was doing the work would you like the refactor done as one if-check at a time, removing the use of one compatibility file at a time or as a complete PR with everything sorted out?

erykoff:
I've done a new PR removing `five.py` entirely here: https://github.com/conda/conda/pull/11267

Issue #11085, opened 2021-12-06
GitHub Actions not disabled for forks


mattkram:
### Checklist
* [x] I added a descriptive title
* [x] I searched for other issues and couldn't find a solution or duplication
* [x] I already searched in Google and didn't find any good information or help
* [x] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?
This is not necessarily a bug, but I'm filing an issue in case the `conda` team wants to take any action.

I forked `conda` in GitHub and began receiving emails for failed GitHub Actions:
![image](https://user-images.githubusercontent.com/11037737/144879254-4ae24929-6c74-453a-8988-985559182427.png)

It appears that GitHub actions are enabled by default when forking a repo. These run under the forking user's account, e.g.: https://github.com/mattkram/conda/actions/runs/1493517417

This can be updated in the settings of the forked repo, but I'm not sure if this can be done by default.

![image](https://user-images.githubusercontent.com/11037737/144881036-65bef6c9-c86d-4937-8906-772172a26ce4.png)

PR #11265, opened 2022-02-25
Only run GHAs on non-forks

kenodegard:
Resolves https://github.com/conda/conda/issues/11085

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,6 @@ concurrency:
 
 jobs:
   windows:
-    if: '!github.event.repository.fork'
     runs-on: windows-2019
     strategy:
       fail-fast: false
@@ -70,7 +69,6 @@ jobs:
           env_vars: OS,PYTHON
 
   linux:
-    if: '!github.event.repository.fork'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -21,6 +26,7 @@ concurrency:
 
 jobs:
   windows:
+    if: '!github.event.repository.fork'
```
jezdez, line 29: So I'm not sure if we should prevent the CI jobs to run in forks to be honest, they should not have any problem running there.

kenodegard: Hm, can we make tests like this dependent on an environment variable? I worry about forks running anything without the user being explicitly aware. We can do something like:
```suggestion
    if: >-
      !github.event.repository.fork
      || fromJSON(env.CONDA_GHA_ENABLED || "false")
```

jezdez: Yeah I understand where you come from, I think we should investigate what other projects that use GitHub Actions use are doing before committing to one solution. I'm also worried we'll forget to add the checks when we evolve the workflows in the future and then end up with partial workflows run in users' namespaces.

kenodegard: Fully agree on forgetting 😄, ok will look around at other projects.

kenodegard: I looked around and other projects don't seem to care about this at all for _any_ of their workflows 🤷🏻‍♂️

So to avoid overkill I'll only apply the fork check on workflows that deal with conda org specific things (e.g. project boards, etc)

jezdez (approved):
Thanks @kenodegard!

PR #11276, opened 2022-03-02
Libmamba issue form

LtDan33:
## Purpose
Have a place that people using the new experimental solver can go to specifically discuss issues and the conda dev team can collect the needed information to make it better

## Approach
Have an issue form that is very similar to the existing bug form. A label would also be auto-applied so that its easy to filter and find these issues.

## Other needed info
This will in time go away as we move this feature to the normal release and issues should occur less frequently.

### UPDATE 03/03: 
Added two labels which will make it easier to filter these issues as they come in. The `solver::libmamba` as well as a `tag::experimental`. Will be useful to track if we are having more issues than normal with experimental features.


jaimergp:
We probably need a field to upload / attach (but not paste) the log files too.

LtDan33:
There isn't a way to add a file upload to the form, but we can just request that they add it to a comment on the issue.

There also isn't a great way to see the preview, so I just added to my fork here : https://github.com/LtDan33/conda/issues/new?assignees=&labels=solver%3A%3Alibmamba%2Ctag%3A%3Aexperimental&template=libmamba-bug-report.yml

I also will create a PR to update the link that we show in the command line to point directly to this issue once merged.

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
```diff
@@ -0,0 +1,86 @@
...
+    id: context
+    attributes:
+      label: Additional Context
+      description: Add any additional context information or screenshots you think are useful.
+      
+  - type: markdown
+    attributes:
+      value: |
+       ## Important: 
+       Please attach the log files in a comment to this issue after you have used the `Submit new issue` button. The log files should be found on your computer at `<install path of anaconda>\envs\libmamba\pkgs\.logs\`. You will also see this exact path as it relates to your system on the command line after running anything with `--experimental-solver=libmamba`.
```
kenodegard, line 86: Looks like any `textarea` without the `render` attribute defined will accept attachments (see [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#:~:text=When%20this%20key%20is%20provided%2C%20the%20text%20area%20will%20not%20expand%20for%20file%20attachments%20or%20Markdown%20editing.)):

```suggestion      
  - type: textarea
    attributes:
      label: Log file
      description: |
        Please attach the associated log file (drag and drop the file into the textbox
        below). The log file can be found at `$CONDA_PREFIX\pkgs\.logs\`. The
        exact path is also printed to the command line when running commands with
        `--experimental-solver=libmamba`.
```

Assuming this all works as expected I'd also swap this "Log file" block with the "Additional Context" block.

LtDan33: Nice find, yeah that will be a cleaner way to do it.

I do think that the "CONDA_PREFIX" part is really confusing to an average user though, saying it in a more common language I think would be clearer.

kenodegard: Yeah I agree with `$CONDA_PREFIX` being non-obvious but I don't think there's a good way to broadly encapsulate any environment name/path otherwise?

LtDan33: Actually just adding a quick explanation of what it means will be good. I'll add that.

LtDan33: Could also have "the environment where libmamba is installed". It might not matter that much as the path would be shown. We can see if we need to reword if people are having issues.

jaimergp: It's actually "a `.logs` directory under your package cache directory", which can be `$CONDA_ROOT/pkgs`, `$CONDA_PREFIX/envs/<env>/pkgs` or `~/.conda/pkgs`, depending on the configuration and directory permissions.

kenodegard: Ah, I didn't know that, since there are 3 different places, asking the user to figure out where it may be located is probably too much.

I'd be in favor of dropping the `.logs` instructions and only specifying that the path is printed to the command line.

jaimergp: Yes it's a bit weird. It's the "first writable cache directory" (which might also change across runs!), but it uses the same logic as conda. I agree that we should just refer to the blurb posted at the beginning.

jezdez:
pre-commit.ci autofix

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
```diff
@@ -0,0 +1,84 @@
+name: Libmamba Bug Report (Experimental Feature)
+description: Create a report for the experimental solver
+labels: ['solver::libmamba', 'tag::experimental']
```
jezdez, line 3: We'll need to update the https://github.com/conda/conda/blob/master/.github/labels.yml file with the "tag::experimental" label.

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
```diff
@@ -0,0 +1,84 @@
+name: Libmamba Bug Report (Experimental Feature)
```
jezdez, line 1:
```suggestion
name: Libmamba Solver Feedback (Experimental Feature)
```
Let's make sure we're not using the wrong name for this, as libmamba isn't the one we're asking for feedback but the libmamba-based solver. I would also suggest to make this a bit different in framing it a bit more generically as "feedback" than a "bug report" as that reduces the barrier a bit, e.g. not every issue users may find with the solver may be able identifiable as a bug for them. For example we also want to get other feedback like user experience problems etc.

As such, could you please also rename the template file name to `libmamba-solver-feedback.yml`?

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
```diff
@@ -0,0 +1,84 @@
...
+          required: true
+        - label: I looked at the [README.md](https://github.com/conda-incubator/conda-libmamba-solver) and didn't see anything to help
+          required: true
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: And what should have happened instead? This really helps us triage quicker and greatly increases the chance that we can get around to solve your issue
+      placeholder: Tell us what you see!
+      value: "A wild Caterpie appeared!"
```
jezdez, line 40:
```suggestion
      value: "I tried to install package XYZ and these things happened instead:"
```
I'm not sure how to say this, but could we exchange this with something like this to reduce the risk of confusion?

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
```diff
@@ -0,0 +1,84 @@
...
+      render: shell
+
+  - type: textarea
+    attributes:
+      label: Log file
+      description: |
+        Please attach the associated log file (drag and drop the file into the textbox
+        below). The log file can be found at `$CONDA_PREFIX\pkgs\.logs\`. `$CONDA_PREFIX` is the path to the libmamba environment. The
+        exact path is also printed to the command line when running commands with
+        `--experimental-solver=libmamba`.
```
jezdez, line 84:
```suggestion
        the experimental libmamba solver enabled.
```
There are a few other way to enable the solver, so I think we can be a little less specific here.

jezdez (commented):
This is good, I just had a few word-smithing notes, nothing major.

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
kenodegard: I'm also not sure we need any placeholder text at all here? Just make sure it's a required field so they can't leave it empty?

jezdez: I'm... not sure

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
kenodegard: Do we really need a new tag for this? Are we trying to capture a difference between something being https://github.com/conda/conda/labels/solver%3A%3Alibmamba vs. https://github.com/conda/conda/labels/solver%3A%3Alibmamba + `[tag::experimental]`?

jezdez: You're right, it's fine to just use [solver::lib](https://github.com/conda/conda/labels/solver%3A%3Alibmamba).

LtDan33: We are going to have more experimental features and items in the future, if like to capture if they are generating more issues front the beginning rather than going back and retro actively applying

kenodegard: Discussed with @LtDan33 and will define `[tag::experimental]` so we can group this and future experimental features together.

.github/ISSUE_TEMPLATE/libmamba-bug-report.yml
kenodegard: Done

kenodegard (approved):

PR #11288, opened 2022-03-03
Remove old CLA code

kenodegard:
Since we now rely on [conda/clabot-config](https://github.com/conda/clabot-config) and [conda/conda-bots](https://github.com/conda/conda-bots) for our CLA checking we can remove the old code.


jezdez:
@kenodegard I've left a comment in an old, internal ticket about this to make sure we're okay deleting those files.

jezdez (approved):
We've discussed this internally and this is good to go.

PR #11310, opened 2022-03-10
Correction of RegisterPython description in Windows Installer arguements

### LINKED ISSUE #11266: Incorrect description of `RegisterPython` in Windows installer arguments
### What is the idea?

https://github.com/conda/conda/blob/master/docs/source/user-guide/install/windows.rst describes as:
```
/RegisterPython=[0|1]---Make this the system's default Python. 0 indicates JustMe, which is the default. 1 indicates AllUsers.
```
According to https://github.com/conda/constructor/blob/master/constructor/nsis/main.nsi.tmpl, `RegisterPython` controlls only whether registration into registry is processed or not.
It is controlled by `InstallationType` via `SetShellVarContext` whether `JustMe` or `AllUsers`.

### Why is this needed

This is an official document and similar descriptions have already been spread into other resources, not only in English.
It seems that investigating https://github.com/conda/constructor/blob/master/constructor/nsis/main.nsi.tmpl is the only way to understand actual behavior.

### What should happen?

It should be like:
```rst
* ``/RegisterPython=[0|1]``---Make this the system's default
  Python. 
  ``0`` indicates no registration, which is the default. ``1``
  indicates to be registered as ``/InstallationType`` specifies.
```
I'm not a native speaker of English, so please correct appropriately...

### Additional Context

This issue should be under `type::documentation` label.

ChrisPanopoulos:
@awwad I have revised the RegisterPython description that was inaccurate and raised as an issue. Please review my PR!

Fixes #11266

### DIFF
--- a/docs/source/user-guide/install/windows.rst
+++ b/docs/source/user-guide/install/windows.rst
@@ -44,8 +44,8 @@ argument. The following optional arguments are supported:
 * ``/AddToPath=[0|1]``---Default is ``0``
 * ``/RegisterPython=[0|1]``---Make this the system's default
   Python.
-  ``0`` indicates Python won't be registered as system's default. ``1``
-  indicates Python will be registered as system's default.
+  ``0`` indicates Python won't be registered as the system's default. ``1``
+  indicates Python will be registered as the system's default.
 * ``/S``---Install in silent mode.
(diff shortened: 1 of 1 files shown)


ChrisPanopoulos:
i just signed the contributors agreement. Can we proceed with the review?

PR #11349, opened 2022-03-21
Remove unused `auxlib` parts

### LINKED ISSUE #11338: Remove unused parts of `auxlib`
### What is the idea?

Remove unused submodules of `auxlib`.

### Why is this needed

Continuing work of absorbing/refactoring `auxlib`. (See https://github.com/conda/conda/pull/11025)

### What should happen?

Removal of unused submodules.

### Additional Context

_No response_

kenodegard:
Removing the following unused (or low usage) submodules of `conda.auxlib`:
- `configuration`: conda implements its own way of parsing and handling config files
- `crypt`: it's unused why keep it?
- `deprecation`: currently unused and probably better to use other tools like [`python-deprecated`](https://python-deprecated.readthedocs.io)
- `factory`: unused and documented as being incomplete, so why keep it?
- `path`: mostly unused, convenience functions not widely used

Resolves #11338


kenodegard:
@travishathaway This is what I'd compiled as safe to remove last week, I had not finished reviewing the remaining submodules of `auxlib`. Feel free to pick up where I left off or feel free to reassign this back to me if you don't think you'll have time to take a look.

travishathaway:
@kenodegard I checked out the rest of the sub-modules and it looks like they are all currently in use else in the repository. I don't believe anything else needs to be removed.

travishathaway (approved):
Looks good.

PR #11356, opened 2022-03-23
Remove expensive conda-build autouse fixture

kenodegard:
The packages built by this autouse fixture are only used in two places. Blindly building these packages for any test run is unnecessarily expensive (e.g. when trying to run individual tests locally).


jezdez (approved):
Well dang, that's a good idea!

PR #11360, opened 2022-03-24
Reorg + new tests for main_clean

kenodegard:
Depends on ~#11351~ & #11356

Discovered while writing tests for the untested functionality that:
- The `.trash` directory and `.trash` suffix no longer appear to be in use and we need to investigate removing it altogether
- Only the `--tempfiles` function requires a prefix but it doesn't check to ensure a prefix is passed but rather simply accepts any valid path, this is inconsistent and should be revisited

### DIFF
--- a/tests/cli/test_main_clean.py
+++ b/tests/cli/test_main_clean.py
@@ -7,9 +7,8 @@ from json import loads as json_loads
 from os import listdir
 from os.path import basename, isdir, join, exists
 from shutil import copy
-import pytest
 from conda.base.constants import CONDA_TEMP_EXTENSION
-from conda.core.subdir_data import create_cache_dir, SubdirData
+from conda.core.subdir_data import create_cache_dir
 from conda.testing.integration import make_temp_package_cache, run_command, Commands, make_temp_env
 
 
@@ -44,11 +43,6 @@ def _get_all(pkgs_dir):
 
 def _any_pkg(name, contents):
     return any(basename(c).startswith(f"{name}-") for c in contents)
-
-
-@pytest.fixture
-def clear_cache():
-    SubdirData._cache_.clear()
 
 
 # conda clean --force-pkgs-dirs
--- a/tests/cli/test_main_clean.py
+++ b/tests/cli/test_main_clean.py
@@ -36,8 +36,12 @@ def _get_all(pkgs_dir):
     return _get_pkgs(pkgs_dir), _get_tars(pkgs_dir), _get_index_cache()
 
 
-def _any_pkg(name, contents):
-    return any(basename(content).startswith(f"{name}-") for content in contents)
+def assert_any_pkg(name, contents):
+    assert any(basename(content).startswith(f"{name}-") for content in contents)
+
+
+def assert_not_pkg(name, contents):
+    assert not any(basename(content).startswith(f"{name}-") for content in contents)
 
 
 # conda clean --force-pkgs-dirs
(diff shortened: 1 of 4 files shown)


kenodegard:
@anaconda-issue-bot check

beeankha (approved):
👍

tests/cli/test_main_clean.py
```diff
@@ -0,0 +1,235 @@
...
+    return _get_pkgs(contents=contents), _get_tars(contents=contents), _get_index_cache()
+
+
+def _any_pkg(name, contents):
+    return any(basename(c).startswith(f"{name}-") for c in contents)
+
+
+@pytest.fixture
+def clear_cache():
+    SubdirData._cache_.clear()
```
jezdez, line 51: I wonder if we have a use for this elsewhere?

tests/cli/test_main_clean.py
```diff
@@ -0,0 +1,235 @@
...
+
+def _get_index_cache():
+    return glob(join(create_cache_dir(), "*.json"))
+
+
+def _get_tempfiles(pkgs_dir=None, *, contents=None):
+    return [
+        t
+        for t in contents or _get_contents(pkgs_dir)
+        if t.endswith((".trash", CONDA_TEMP_EXTENSION))
```
jezdez, line 36:
```suggestion
        content
        for content in contents or _get_contents(pkgs_dir)
        if content.endswith((".trash", CONDA_TEMP_EXTENSION))
```

tests/cli/test_main_clean.py
```diff
@@ -0,0 +1,235 @@
...
+def _get_contents(pkgs_dir):
+    return [join(pkgs_dir, d) for d in listdir(pkgs_dir)]
+
+
+def _get_pkgs(pkgs_dir=None, *, contents=None):
+    return [d for d in contents or _get_contents(pkgs_dir) if isdir(d)]
+
+
+def _get_tars(pkgs_dir=None, *, contents=None):
+    return [t for t in contents or _get_contents(pkgs_dir) if t.endswith((".tar.bz2", ".conda"))]
```
jezdez, line 25:
```suggestion
def _get_contents(pkgs_dir):
    return [join(pkgs_dir, name) for name in listdir(pkgs_dir)]


def _get_pkgs(pkgs_dir=None, *, contents=None):
    return [content for content in contents or _get_contents(pkgs_dir) if isdir(content)]


def _get_tars(pkgs_dir=None, *, contents=None):
    return [content for content in contents or _get_contents(pkgs_dir) if content.endswith((".tar.bz2", ".conda"))]
```

jezdez: I mean, this is just a minor nit, simplifies debugging maybe in the future.

tests/cli/test_main_clean.py
```diff
@@ -0,0 +1,235 @@
...
+    ]
+
+
+def _get_all(pkgs_dir):
+    contents = _get_contents(pkgs_dir)
+    return _get_pkgs(contents=contents), _get_tars(contents=contents), _get_index_cache()
+
+
+def _any_pkg(name, contents):
+    return any(basename(c).startswith(f"{name}-") for c in contents)
```
jezdez, line 46:
```suggestion
    return any(basename(content).startswith(f"{name}-") for content in contents)
```

tests/cli/test_main_clean.py
```diff
@@ -0,0 +1,235 @@
...
+        if t.endswith((".trash", CONDA_TEMP_EXTENSION))
+    ]
+
+
+def _get_all(pkgs_dir):
+    contents = _get_contents(pkgs_dir)
+    return _get_pkgs(contents=contents), _get_tars(contents=contents), _get_index_cache()
+
+
+def _any_pkg(name, contents):
```
jezdez, line 45: Since this function is used only for checking assertions I'd say it'd be fine to rename it to `assert_any_pkg` and do the assert check inside, would cleanup the tests below a bit more.

```
assert _any_pkg(pkg, _get_pkgs(pkgs_dir))
```
vs.

```
assert_any_pkg(pkg, pkgs_dir)
```

jezdez (commented):
Just minor variable nitpicking, and more oppurtunity to factor out `_any_pkg` function.

tests/cli/test_main_clean.py
kenodegard: Ah, this comes from a personal convention where if the list/dict comprehension fits on a single line I'll use one-char variables to keep it super short and sweet but I agree it produces inconsistencies so I'll make the change.

tests/cli/test_main_clean.py
kenodegard: No idea, do we want to move **all** fixtures into `conftest.py` so it'll be easier to discover?

tests/cli/test_main_clean.py
kenodegard: What do we do with the `assert not _any_pkg()` cases?

tests/cli/test_main_clean.py
kenodegard: Moved to `conftest.py`

tests/cli/test_main_clean.py
kenodegard: Done

tests/cli/test_main_clean.py
kenodegard: Done

tests/cli/test_main_clean.py
kenodegard: Done

tests/cli/test_main_clean.py
jezdez: ah, yeah, the pattern used in pytest heavy codebases is implementing `assert_not_any_pkg` or `assert_no_pkg`?

tests/cli/test_main_clean.py
jezdez: Thank you!

jezdez (approved):
Thank you for the quick updates!

Issue #11334, opened 2022-03-15
Replace vendored `urllib3` with builtin `urllib`


kenodegard:
We apparently only care about the URL parser from urllib3, which makes me wonder if there isn't a standalone parser that we can use instead (or if we can rely on an updated package in the defaults channel instead)?

_Originally posted by @jezdez in https://github.com/conda/conda/pull/11290#discussion_r826839406_

PR #11373, opened 2022-03-28
Issue 11334 replace vendored urllib3

travishathaway:
See https://github.com/conda/conda/issues/11334 for a full description of this issue.

The goal of this PR was removing our dependency on `urllib3` in our `_vendors` module. I was able to do this by instead relying on the `urllib.parse.urlparse`. On top of this, I copied over the `Url` object from `urllib3` and made some slight modifications to it to make updating the code easier.

The `conda.common.url` module augments the default parse behavior of the std lib `urllib.parse` module. We do this by defining our own `urlparse` method. Where this method differs from the std. lib. version is that it returns its own `Url` object instead of a `ParsedResult`. This `Url` object is much friendlier to those wishing to reconstitute a URL from various parts (like when stripping out username and password).

Acceptance Criteria:
- Passes all test
- New API has parity with previous


beeankha:
@anaconda-issue-bot check

kenodegard (commented):
Looks good just a few suggestions since `Url` is now our own implementation.

conda/common/url.py
```diff
@@ -125,11 +131,119 @@ def path_to_url(path):
...
+
+        return self.__class__(**new_kwargs)
+
+
+def parse_result_as_url_obj(parse_result: ParseResult) -> Url:
+    """
+    We use this to transform an urllib.parse.ParseResult object into a Url object
+    """
+    values = {fld: getattr(parse_result, fld, "") for fld in url_attrs}
+    return Url(**values)
```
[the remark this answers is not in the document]
travishathaway: I had actually considered this as well. I will go ahead and implement this and call it `from_parse_result`.

conda/common/url.py
```diff
@@ -125,11 +131,119 @@ def path_to_url(path):
...
+            return self.username
+
+    @property
+    def netloc(self):
+        if self.port:
+            return "%s:%d" % (self.hostname, self.port)
+        return self.hostname
+
+    def __str__(self):
+        scheme, path, params, query, fragment, username, password, hostname, port = self
```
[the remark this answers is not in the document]
travishathaway: Yeah, that's fair. I included it here because it was it's a property of the ParseResult object in the stdlib. Even there it's not used anymore.

conda/common/url.py
```diff
@@ -317,9 +431,14 @@ def split_scheme_auth_token(url):
...
-                        query=url_parts.query).url
-    return remainder_url, url_parts.scheme, url_parts.auth, token
+    remainder_url = Url(
+        hostname=url_parts.hostname,
+        port=url_parts.port,
+        path=url_parts.path,
+        query=url_parts.query,
+    )
+
+    return join(remainder_url.netloc, remainder_url.path), url_parts.scheme, url_parts.auth, token
```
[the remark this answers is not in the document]
travishathaway: Good catch! When I initially implemented this, I thought I was going to be able to use the `ParseResult` object (hence the `netloc` property access).

When I now look at it, I realize this `join` doesn't make sense as I can just use `str(remainder_url)` instead.

conda/common/url.py
```diff
@@ -125,11 +131,116 @@ def path_to_url(path):
...
+    "query",
+    "fragment",
+    "username",
+    "password",
+    "hostname",
+    "port",
+)
+
+
+class Url(namedtuple("Url", url_attrs)):
```
kenodegard, line 146: Hm since we're starting to include typing can we rework this to use `typing.NamedTuple`? See [this SO](https://stackoverflow.com/a/61844762) for why we need a private class for this to work, if we opt for this solution we should include that SO link as a reference. All references to `url_attrs` can use `Url._fields` instead.

```suggestion
class _Url(NamedTuple):
    scheme: str
    path: str
    query: str
    fragment: str
    username: str
    password: str
    hostname: str
    port: int

class Url(_Url):
```

conda/common/url.py
```diff
@@ -125,11 +131,116 @@ def path_to_url(path):
...
+        >>> u = Url(scheme='https', hostname='anaconda.com', path='/hello',
+        ...         username='user', password='pass')
+        >>> u1 = u.replace(username='', password='')
+        >>> str(u1)
+        'https://anaconda.com/hello'
+        """
+        new_kwargs = self.as_dict()
+        new_kwargs.update(kwargs)
+
+        return self.__class__(**new_kwargs)
```
kenodegard, line 228: Hm, actually both `as_dict` and `replace` are predefined as part of [`collections.namedtuple`/`typing.NamedTuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple). Should we just use those methods (`_asdict` and `_replace`) and/or alias them?

```suggestion
    replace = _Url._replace
```

conda/common/url.py
```diff
@@ -125,11 +131,116 @@ def path_to_url(path):
...
+        return self.hostname
+
+    def __str__(self):
+        scheme, path, query, fragment, username, password, hostname, port = self
+        url = ""
+
+        if scheme:
+            url += f"{scheme}://"
+        if password and username:
+            url += f"{username}:{password}" + "@"
```
kenodegard, line 197: Minor nit!

```suggestion
            url += f"{username}:{password}@"
```

kenodegard (approved):
I realized that `namedtuple` already provides `as_dict` & `replace` functionality so we should probably not reimplement those but other than that this looks good to me!

edit: also if you rebase to include #11378 the tests should pass again

conda/common/url.py
travishathaway: Hi @kenodegard , to be honest, I really wanted to use data classes for this, but figured because we still need to support Python 3.6 it wouldn't be a good idea.

Sounds like we will be dropping support for this soon though, right? Maybe it would be better for me to redo this as a data class instead? (NamedTuple is just a little awkward for this use case, IMHO)

kenodegard: True, personally I've just been suuuper hesitant to use dataclasses given [benchmarks](https://twitter.com/anthonypjshaw/status/1498912385268350976?s=20&t=9SMAs71S8zmBJL7J65Hspw) finding it to be significantly slower than normal classes.

Since `_asdict` & `_replace` are predefined in `namedtuple`/`NamedTuple` I'm leaning towards not converting this to a `dataclass`.

Yeah, don't worry about supporting 3.6, it'll be officially dropped in the next release.

kenodegard:
@travishathaway oh and let's also include the news about dropping `urllib3` for stdlib `urllib`.

jezdez (approved):
Just a minor nits! 💯

PR #11374, opened 2022-03-28
Check for conda before calling shell hook

wulmer:
We faced some issues with the PowerShell hook which gets installed via `conda init`. When a user uninstalls Anaconda, the hook isn't removed automatically which then leads to errors each time when the user opens a new PowerShell instance.

The patch in this PR simply adds a test for the conda executable 
**so that the shell hook is only called if conda exists**.


kenodegard (changes requested):
@wulmer thank you for this contribution!

Since this check is equally valid for all of the other shells would you be able to make the applicable changes for those shells as well?

wulmer:
> @wulmer thank you for this contribution!
> 
> Since this check is equally valid for all of the other shells would you be able to make the applicable changes for those shells as well?

Hi @kenodegard , sure, I'll try to fix this for other shells as well, however this might take some time...

conda/core/initialize.py
```diff
@@ -1504,7 +1509,9 @@ def _powershell_profile_content(conda_prefix):
     conda_powershell_module = dals("""
     #region conda initialize
     # !! Contents within this block are managed by 'conda init' !!
-    (& "{conda_exe}" "shell.powershell" "hook") | Out-String | Invoke-Expression
+    If (Test-Path "{conda_exe}") {{
+        (& "{conda_exe}" "shell.powershell" "hook") | Out-String | ?{{$_}} | Invoke-Expression
```
wulmer, line 1513: @kenodegard : we have had an issue today where `Test-Path "{conda_exe}"` still worked as `conda.exe` existed, but Python couldn't be executed due to some missing files/DLLs (it was a broken installation).

However this led to other scripts failing during PS1 initialization, as the line above then failed: conda was there, and could be called, but Python crashed → the whole profile script was treated as failed.

Should I add an extra layer of safety here and catch exceptions occurring during this conda hook call?

wulmer: At the moment, I am using this code:

```powershell
#region conda initialize
If (Test-Path "C:\Program Files\Anaconda3\Scripts\conda.exe") {
    try {
        (& "C:\Program Files\Anaconda3\Scripts\conda.exe" "shell.powershell" "hook") | Out-String | Invoke-Expression
    }
    catch {
        # some verbose warning message which tells the user
        # what to do, i.e. to reinstall Anaconda.
    }
}
#endregion
```

kenodegard (changes requested):
This looks good as is, it's just missing a news snippet.

Let's wait on adding any more checks to these shell interfaces. We're currently ramping up work on the plugin framework (https://github.com/conda/conda/issues/11820), and we expect all of these shell interfaces to be split into separate plugins in the near future.

kenodegard (approved):
@wulmer thanks for working on this!

Issue #11170, opened 2022-01-27
Python 3.11 support


kenodegard:
### Description

We need to investigate the changes necessary to support Python 3.11. Because this release cycle happens every year. It would also make sense for us to have a policy in place guiding the work that we need to regularly complete for this.

### Good to know

There is currently a rough draft of this document that could be used as a starting point for the work: 
https://github.com/conda-incubator/ceps/pull/24

```[tasklist]
- [ ] https://github.com/conda/conda/pull/12256
- [x] https://github.com/conda/conda/pull/12580
```

EwoutH:
Python 3.11 is now in beta, which means no new features will be added.

- [Python 3.11.0b1 is now available](https://blog.python.org/2022/05/python-3110b1-is-now-available.html)
- [What’s New In Python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html)

EwoutH:
Any update on this effort? Python 3.11 is now in the release candidate phase and will be released in two weeks. Many projects already support Python 3.11 and have wheels available, including NumPy, Pandas, SciPy, Matplotlib and Seaborn.

travishathaway:
@EwoutH, thanks for checking in. This issue is high on our priority list, and we hope to begin working on it in an upcoming sprint. I can not say for certain when this will be completed though.

EwoutH:
Thanks for getting back! Great to hear the team is working on this, keep us in the loop :)

johnnynunez:
Yeah! Python 3.11 support please. It should be the default.
Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite. See Faster CPython for details.
https://docs.python.org/3.11/whatsnew/3.11.html
https://www.phoronix.com/news/Python-3.11-Released

johnnynunez:
Benchmark https://www.phoronix.com/review/python-311-performance

EwoutH:
Hi @travishathaway! Do you have anything to share as of this moment?

EwoutH:
We're now over 4 months over rc1 and almost two months after stable release. Could we get an update on 3.11 support?

johnnynunez:
could you build all packages for python 3.11 in channel python 3.11? https://github.com/pytorch/vision/pull/7108
![image](https://user-images.githubusercontent.com/22727137/213490577-f7b56fb3-c9fd-4372-95d8-3bf49aaab892.png)

[...]

PR #11378, opened 2022-03-28
Add python constraint back for `--update-deps --only-deps`

kenodegard:
Due to recent updates to the available packages there seems to be a preference for Python 3.9 over Python 3.10 causing Python to be downgraded for the `--update-deps --only-deps` test:

```bash
# installs flask=2.0.1 & python=3.10.3
$ conda create -n deps flask=2.0.1

# is not supposed to update flask/python but it's downgrading python to 3.9.11
$ conda install -n deps flask python --update-deps --only-deps
```

This is yet another "short-term" fix for the larger issue of our testsuite depending on the everchanging defaults channel.

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -884,7 +884,7 @@ dependencies:
             result_before = subprocess_call_with_clean_env([python, "--version"])
             assert package_is_installed(prefix, "flask=2.0.1")
             assert package_is_installed(prefix, "jinja2=3.0.1")
-            run_command(Commands.INSTALL, prefix, "flask", "python", "--update-deps", "--only-deps")
+            run_command(Commands.INSTALL, prefix, "flask", "python=3.10", "--update-deps", "--only-deps")
             result_after = subprocess_call_with_clean_env([python, "--version"])
             assert result_before == result_after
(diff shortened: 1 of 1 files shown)


jezdez:
So IIRC we had trouble with this on the libmamba solver side (where we run these tests as well). @jaimergp do you think we can add this pin back again?

jaimergp:
I think this will break when Python 3.11 is out. The behavior being tested here is that the Python version shouldn't change across runs unless explicitly asked (e.g. `python=3.10`). The fix suggested here only patches this temporary hiccup because for whatever reason the solver is finding 3.9 more suitable now 🤷  This shouldn't be happening and in my opinion it's a bug, but who knows what the explanation is...

That said, we are [overriding](https://github.com/conda-incubator/conda-libmamba-solver/blob/f05f59cb138c3a99450f71a86fe0bd06a7024995/tests/test_modified_upstream.py#L105-L124) this test downstream. So it does not matter for us.

jezdez:
Thanks @jaimergp, yeah, we'll need to figure out a proper fix. #11170 tracks the Python 3.11 support.

jezdez (approved):

PR #11381, opened 2022-03-30
Remove duplicated instruction in manage-python.rst

simon9500:
While reading the conda docs I noticed one of the instructions was written twice.  It looks like it wasn't intended so I thought it was a good idea to correct it.  See attached image below:

image


kenodegard:
@anaconda-issue-bot check

PR #11386, opened 2022-03-31
Standardize tarball & pkg finding in `conda clean`

kenodegard:
This is in preparation for combining `rm_tarballs` and `rm_pkgs` since both are identically structured.

The primary changes here is to:
- move the `getsize` logic previously defined in `rm_tarballs` into `find_tarballs`
- combine the `pkgs_dirs` for-loop with the `pkg_sizes` for-loop in `find_pkgs`
- split out a `getsize` helper function shared between the two implementations

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -4,9 +4,9 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from logging import getLogger
-from os import listdir, lstat, walk
-from os.path import getsize, isdir, join
-from typing import Iterable, List
+from os import lstat, walk
+from os.path import isdir, join
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 import sys
 
 from ..base.constants import CONDA_PACKAGE_EXTENSIONS, CONDA_TEMP_EXTENSIONS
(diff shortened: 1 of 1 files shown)


conda/cli/main_clean.py
```diff
@@ -16,25 +17,57 @@
...
+                pass
+            else:
+                pkg_sizes.setdefault(pkgs_dir, {})[tar] = size
 
-def rm_tarballs(args, pkgs_dirs, total_size, verbose=True):
+    return {
+        "warnings": warnings,
+        "pkg_sizes": pkg_sizes,
+        "pkgs_dirs": {pkgs_dir: tuple(pkgs) for pkgs_dir, pkgs in pkg_sizes.items()},
+        "total_size": sum(sum(pkgs.values()) for pkgs in pkg_sizes.values()),
```
jezdez, line 66: I wonder if this would be good to have in an own private function like `_sum_sizes` since there is the same thing happening in `find_pkgs`?

conda/cli/main_clean.py
```diff
@@ -16,25 +17,57 @@
 _EXTS = (*CONDA_PACKAGE_EXTENSIONS, *(f"{e}.part" for e in CONDA_PACKAGE_EXTENSIONS))
 
 
-def find_tarballs():
+def _getsize(*parts: str, warnings: Optional[List[Tuple[str, Exception]]] = None) -> int:
```
jezdez, line 20:
```suggestion
def _get_size(*parts: str, warnings: Optional[List[Tuple[str, Exception]]] = None) -> int:
```
This is definitely more of a nit, feel free to ignore.

jezdez (commented):
This looks good, just a few questions around abstraction and naming.

jezdez (approved):

PR #11393, opened 2022-04-04
Bump vendored `cpuinfo` 4.0.0 → 8.0.0

### LINKED ISSUE #11332: Bump vendored `cpuinfo` version
Let's file a ticket to upgrade or upstream the changes here to cpuinfo.

_Originally posted by @jezdez in https://github.com/conda/conda/pull/11290#discussion_r826833053_

kenodegard:
### Description

Bump `py-cpuinfo` in order to eliminate the need for patching.

Resolves #11332 

### Checklist - did you ...

- [X] Add `news` entry to be included in the next release's `CHANGELOG`?
     

tools/vendoring/patches/cpuinfo.patch
```diff
@@ -1,44 +0,0 @@
...
-+	if re.match(r'^i\d86$|^x86$|^x86_32$|^i86pc$|^ia32$|^ia-32$|^bepc$', raw_arch_string):
- 		arch = 'X86_32'
- 		bits = 32
- 	elif re.match('^x64$|^x86_64$|^x86_64t$|^i686-64$|^amd64$|^ia64$|^ia-64$', raw_arch_string):
- 		arch = 'X86_64'
- 		bits = 64
- 	# ARM
-+	elif re.match('^arm64$|^arm64[a-z]$|^arm64-[a-z]$', raw_arch_string):
-+		arch = 'ARM_8'
-+		bits = 64
```
jezdez, line 19: I think this section is currently not covered by cpuinfo when looking at the regex patterns in https://github.com/workhorsy/py-cpuinfo/blob/220d5b05e26160474a384c8803e59488d4292228/cpuinfo/cpuinfo.py#L804-L806.

@chenghlee do you happen to know if we actually need `^arm64$|^arm64[a-z]$|^arm64-[a-z]$` or could we rely on `^armv8-a|aarch64|arm64$`?

jezdez (changes requested):
Let's make sure we're not accidentally breaking arm64 support with this.

https://github.com/conda/conda/commit/022d862dc7b86417433c829cfac722184a4755d9 was the change that added it (#10128).

tools/vendoring/patches/cpuinfo.patch
kenodegard: I don't think `arm64-?[a-z]` are a thing (looking at [this list from Wikipedia](https://en.wikipedia.org/wiki/Uname))?

kenodegard: The regex pattern used in our patch looks like a copy/paste of the `armv7`/`armv8` patterns.

jezdez: Yeah, @katietz confirmed your suspicion that this wasn't correct in the old patch and cpuinfo is up-to-date

jezdez (approved):

Issue #11358, opened 2022-03-23
Drop Python 3.6 Support


kenodegard:
### Summary

Python 3.6 reached EOL on 23 Dec 2021.

### Linked Issues & PRs
- https://github.com/conda/conda/pull/11364 
- Remove Python 3.6 references (e.g. `setup.py`)
- Include announcement in release notes

jezdez:
As a point of reference, Anaconda has stopped regularly building [Python 3.6 packages when Python 3.10 was released](https://anaconda.cloud/anaconda-repo-news). That's just another data point for this to be done rather sooner than later.

jaimergp:
conda-forge also dropped it recently.

PR #11398, opened 2022-04-05
Replace `conda.auxlib.collection.frozendict` with vendored `frozendict`

### LINKED ISSUE #11337: Replace `conda.auxlib.collection.frozendict` with vendored `frozendict`
### What is the idea?

We are using two implementations of `frozendict`. A slim version via `conda.auxlib.collection` and the complete one via the vendored `frozendict` package. We should identify why and whether there's a reason to continue doing so.

### Why is this needed

Cut down on "duplicate" code.

### What should happen?

Remove `conda.auxlib.collection.frozendict`.

### Additional Context

_No response_

kenodegard:
### Description

Removes custom `conda.auxlib.collection.frozendict` in favor of the already vendored `frozendict`.

Resolves #11337

### Checklist - did you ...

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -7,8 +7,8 @@ from collections import defaultdict, OrderedDict, deque
 import copy
 from logging import DEBUG, getLogger
 
-from .auxlib.collection import frozendict
 from .auxlib.decorators import memoize, memoizemethod
+from ._vendor.frozendict import frozendict
 from ._vendor.toolz import concat, groupby
 from ._vendor.tqdm import tqdm
 from .base.constants import ChannelPriority, MAX_CHANNEL_PRIORITY, SatSolverChoice
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -8,7 +8,7 @@ import copy
 from logging import DEBUG, getLogger
 
 from .auxlib.decorators import memoize, memoizemethod
-from ._vendor.frozendict import frozendict
+from ._vendor.frozendict import FrozenOrderedDict as frozendict
 from ._vendor.toolz import concat, groupby
 from ._vendor.tqdm import tqdm
 from .base.constants import ChannelPriority, MAX_CHANNEL_PRIORITY, SatSolverChoice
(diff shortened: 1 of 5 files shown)


beeankha (commented):
LGTM 👍

beeankha (approved):
Oops, selected "comment" vs "approve" earlier

conda/resolve.py
```diff
@@ -7,8 +7,8 @@
 import copy
 from logging import DEBUG, getLogger
 
-from .auxlib.collection import frozendict
 from .auxlib.decorators import memoize, memoizemethod
+from ._vendor.frozendict import frozendict
```
jezdez, line 11: If I understand this correctly, the vendored frozendict library doesn't use `collections.OrderedDict` as the one in auxlib but an `collections.abc.Mapping` as its base class, which I think (!) means that it's not going to be ordered by default anymore. This is probably going to have a negative impact on the `Resolve.__hash__` method since it won't changes the hash of the various values used with `frozendict`. Can we double-check if `collections.abc.Mapping` are ordered now on Python >= 3.7?

jezdez (commented):
This has a tricky ordering risk I think but in quick research I couldn't find out if `collections.abc.Mapping` is ordered like `dict` on Python 3.x or not.

conda/resolve.py
jezdez, line 11:
```suggestion
from ._vendor.frozendict import FrozenOrderedDict as frozendict
```

Maybe that would be enough to make this possible?

kenodegard: Actually, I don't think this makes a difference anymore given that we are [dropping Python 3.6 support](https://github.com/conda/conda/issues/11358). While `frozendict` inherits from `Mapping` it's using a private, internal `dict` instance:
https://github.com/conda/conda/blob/077616b2f3f19eae332613d304e68f4ca3caae76/conda/_vendor/frozendict/__init__.py#L18-L21

So both `frozendict` and `FrozenOrderedDict` are actually ordered implementations at this time.

PR #11415, opened 2022-04-14
extend user-agent if solver != default

jaimergp:
### Checklist - did you ...


conda/base/context.py
```diff
@@ -821,6 +821,8 @@ def user_agent(self):
         builder.append("%s/%s" % self.os_distribution_name_version)
         if self.libc_family_version[0]:
             builder.append("%s/%s" % self.libc_family_version)
+        if self.experimental_solver.value != "classic":
+            builder.append("solver/%s" % self.experimental_solver.value)
```
jezdez, line 825: Interesting that it's so easy! Do you think we can try adding the version of it as well?

jaimergp: Mmm, that requires running the `_get_solver_class()` function, importing the class, and expecting a version field somewhere. We haven't standardized this, and I was trying to keep this as agnostic as possible. If we want more details, we might want to rethink this; e.g. offering an API for downstream plugins to add stuff to the user agent string.

The alternative right now:

* Implement this from the `conda-libmamba-solver` side without such an API and revisit in the future.
* Quick API: add a `user_agent` property to the Solver class, and have the plugin define it?

There are lots of ways of fixing this but once we choose one we need to stick to it, I guess, so better think it through.

conda/base/context.py
```diff
@@ -822,7 +822,14 @@ def user_agent(self):
         if self.libc_family_version[0]:
             builder.append("%s/%s" % self.libc_family_version)
         if self.experimental_solver.value != "classic":
-            builder.append("solver/%s" % self.experimental_solver.value)
+            from ..core.solve import _get_solver_class
+
+            user_agent_str = "solver/%s" % self.experimental_solver.value
+            Solver = _get_solver_class()
```
jezdez, line 828: Thinking about this more, I think this defeats the purpose of the conda.base namespace, where no imports should happen other than conda.common, conda._vendor and conda.base (at least according to the README: https://github.com/conda/conda/tree/master/conda/base)

Can we do this appending in the place where the user-agent string is used instead?

conda/base/context.py
```diff
@@ -822,6 +822,15 @@ def user_agent(self):
...
         if self.libc_family_version[0]:
             builder.append("%s/%s" % self.libc_family_version)
+        if self.experimental_solver.value != "classic":
+            from ..core.solve import _get_solver_class
+
+            user_agent_str = "solver/%s" % self.experimental_solver.value
+            Solver = _get_solver_class()
+            if hasattr(Solver, "user_agent") and callable(Solver.user_agent):
+                # This has to be a static or class method
+                user_agent_str += f" {Solver.user_agent()}"
```
jezdez, line 832:
```suggestion
            try:
                # Solver.user_agent has to be a static or class method
                user_agent_str += f" {Solver.user_agent()}"
            except (AttributeError, TypeError):
                pass
```

jaimergp: The solver class might need to compute things on the fly (like the version of a library it depends on), hence the callable approach. I'll go EAFP for this!

jezdez (approved):
This just needs a news file TBH.

Issue #11294, opened 2022-03-04
Iterating over multiple repodatas


kenodegard:
https://github.com/conda-incubator/conda-libmamba-solver/pull/17#issuecomment-1059090555:
> > Oh, I see, we are iterating over the repodata_fns (e.g. ('current_repodata.json', 'repodata.json')) and creating a new solver for each one
> 
> We respect the outer loop but internally we are only using repodata.json, iirc. We shouldn't need to retry (if we did, it will fail again). This should noted as an issue so we can fix it!

Per @jaimergp's comment, we are unnecessarily iterating over several repodatas and invoking the solver multiple times.
https://github.com/conda/conda/blob/1905ece79f6d7d005911bf11a639503b1b2a215b/conda/cli/install.py#L238-L252

PR #11433, opened 2022-04-26
Define optional exception attribute to bypass the retry logic

jaimergp:
### Checklist - did you ...

This would close https://github.com/conda/conda/issues/11294.

*** 

I am not sure this is the best solution though. It's arguably the minimum change to make this work (we would only need to set that attribute in the exception raised by libmamba), but it also extends the role of exceptions in flow control, and without a well-defined API. 

Alternatives:
* Some default behaviour in `conda.cli.install` is only defined for `solver=classic` (but this makes everything messier and harder to reason about)
* The retry logic is passed to `conda.core.solve.Solver` directly (but then this would be inherited by other tools in the conda ecosystem like conda-build, conda-smithy, constructor...)
* Essentially, some other refactor that allows us to define CLI behaviours per solver type

Other alternatives:
* Make conda-libmamba-solver behave more like conda classic, using repodatas and retries. I don't like this because we would be doing unnecessary work imo (loading the index twice, etc).

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -285,6 +285,10 @@ def install(args, parser, command='install'):
                     raise PackagesNotFoundError(e._formatted_chains, channels_urls)
 
         except (UnsatisfiableError, SystemExit, SpecsConfigurationConflictError) as e:
+            if not getattr(e, "allow_retry", True):
+                # TODO: This is a temporary workaround to allow downstream libraries 
+                # to inject this attribute set to False and skip the retry logic
+                raise e
             # Quick solve with frozen env or trimmed repodata failed.  Try again without that.
             if not hasattr(args, 'update_modifier'):
                 if repodata_fn == repodata_fns[-1]:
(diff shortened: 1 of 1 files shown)


conda/cli/install.py
```diff
@@ -285,6 +285,10 @@ def install(args, parser, command='install'):
                     raise PackagesNotFoundError(e._formatted_chains, channels_urls)
 
         except (UnsatisfiableError, SystemExit, SpecsConfigurationConflictError) as e:
+            if not getattr(e, "allow_retry", True):
+                # TODO: This is a temporary workaround to allow downstream libraries 
+                # to inject this attribute set to False and skip the retry logic
```
jezdez, line 290: Can you elaborate here which error might be given that allow_retry and where it's implemented in conda-libmamba-solver?

jaimergp: Done!

jezdez:
@jaimergp This is good to go IMO.

jezdez (approved):

PR #11440, opened 2022-04-27
Include HTTP response's reason/message if present

### LINKED ISSUE #11438: Improve `CondaHTTPError` to include the response body
### Checklist

### What is the idea?

When HTTP errors occur `conda` intercepts a number of cases with a hardcoded error message. We'd like to augment/replace these hardcoded messages with messages provided in the response so we can provide more targeted information.

### Why is this needed

Provide more descriptive messaging for HTTP errors (e.g. 403 and 429) to explain why the user is experiencing those errors.

### What should happen?

The server-provided message should be passed through as-is to the user.

Here are some sample response bodies we'd want to provide:

#### 403
```json
{
  "reason": "NOT AUTHORIZED",
  "message": "Your IP address has been blocked due to..."
}
```

#### 429
```json
{
  "reason": "TOO MANY REQUESTS",
  "message": "Your IP address has been throttled due to..."
}
```

### Additional Context

_No response_

kenodegard:
### Description

Resolves #11438 

Attempts to extract the reason/message from the HTTP response body. If successful these are used instead of any hardcoded reason/message.

| Error | Old | New |
|---|---|---|
| 403 |  |  |
| 429 |  |  |

### Checklist - did you ...

### DIFF
--- a/news/11440-improve-http-error-message
+++ b/news/11440-improve-http-error-message
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Attempt parsing HTTP errors as a JSON and extract error details. If present, prefer these details instead of those hard-coded. (#11440)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/exceptions.py
```diff
@@ -531,30 +532,60 @@ class CondaHTTPError(CondaError):
...
+
+        # standardize arguments
+        url = maybe_unquote(url)
         status_code = status_code or '000'
         reason = reason or 'CONNECTION FAILED'
+        if isinstance(reason, str):
+            reason = reason.upper()
         elapsed_time = elapsed_time or '-'
+        if isinstance(elapsed_time, timedelta):
+            elapsed_time = text_type(elapsed_time).split(":", 1)[-1]
```
jezdez, line 553:
```suggestion
            elapsed_time = str(elapsed_time).split(":", 1)[-1]
```

conda/exceptions.py
```diff
@@ -444,34 +445,77 @@ def __init__(self, channel):
...
+                f"""
+                HTTP {status_code} {reason} for channel {channel_name} <{channel_url}>
+
+                """
+            )
+            # since message may include newlines don't include in f-string/dals above
+            + message
+        )
+
+        from .auxlib.logz import stringify
```
travishathaway, line 506: Would it be possible to move this to the top of the file? It's a real pet-peeve of mine to see these import statements here, especially when it's not necessary (e.g. because of circular reference issues)

conda/exceptions.py
```diff
@@ -444,34 +445,77 @@ def __init__(self, channel):
...
 
-        if channel.scheme == 'file':
-            message += dedent("""
-            As of conda 4.3, a valid channel must contain a `noarch/repodata.json` and
-            associated `noarch/repodata.json.bz2` file, even if `noarch/repodata.json` is
-            empty. Use `conda index %s`, or create `noarch/repodata.json`
-            and associated `noarch/repodata.json.bz2`.
-            """) % join_url(channel.location, channel.name)
+        response_details = (stringify(response, content_max_len=1024) or "") if response else ""
```
travishathaway, line 508: I was doing some tests of this locally myself, and discovered an interesting bug. With the `--json` flag turned on, I noticed that `response_details` was still always empty, even when there should have been something in there (e.g. 404 still has content to display).

What I discovered is that the `requests` library's `Response` objects actually evaluates as "falsy" when the status code is an error. Checkout the two examples below:

```python
import requests

resp = requests.get('https://api.github.com/users/travishathaway')

print(resp)   # <Response [200]>
print(bool(resp))   # True

resp = requests.get('https://api.github.com/users/asfaaasdfasdfasdf')

print(resp)   # <Response [404]>
print(bool(resp))  # False
```

Any who, this is all a really long-winded way of saying that you should use a strict `None` check in that if statement to avoid this situation 😅 

```python
response_details = (stringify(response, content_max_len=1024) or "") if response is not None else ""
```

cheers!

conda/exceptions.py
```diff
@@ -531,30 +575,60 @@ class CondaHTTPError(CondaError):
...
+                Elapsed: {elapsed_time}
+                {cf_ray}
+                """
+            )
+            # since message may include newlines don't include in f-string/dals above
+            + message
+        )
 
         from .auxlib.logz import stringify
         response_details = (stringify(response, content_max_len=1024) or '') if response else ''
```
travishathaway, line 620: We should also fix the bug I discovered above here and get that import statement out of here if possible.

travishathaway (changes requested):
Thanks for the quick turn around! I just had a couple suggestions based on a bug I found.

conda/exceptions.py
kenodegard: Absolutely, this was result of a copy-paste that I didn't rethink.

conda/exceptions.py
kenodegard: Good catch! However, since we store the parsed json now do we care for `response_details` anymore, can we just remove it?

travishathaway (approved):
This PR looks good

conda/exceptions.py
travishathaway: I would lean towards yes considering there's probably never been any useful information in there (HTTP errors would have always evaluated to false). But, I think it should stay put for now, considering removing it may have unintended consequences. It is a public piece of JSON, so someone's weird script may break if we take it out.

jezdez (approved):
Nice work!

conda/exceptions.py
kenodegard: I took this a step further and delegated the logic to `stringify`, if `stringify` gets a `None` it simply returns a `None` (this is effectively what it was doing before, `stringify` is only seeking to stringify special kinds of objects). So now we have:

```python
response_details = stringify(response, content_max_len=1024) or ""
```

PR #11455, opened 2022-05-04
Catch all exceptions in context.user_agent

jaimergp:
### Checklist - did you ...

Follow-up for #11415 - ignore `ImportError` too if the solver cannot be imported. Otherwise, `conda info` will fail (and it might be used for debugging purposes).

Actually, we might want to ignore _all_ exceptions in this block? 

@jezdez


jezdez:
@jaimergp Yeah, I know it's a pretty big anti-pattern, but we should swallow exceptions to make user experience not a problem. It's not great, but better than risking stuff.

conda/base/context.py
```diff
@@ -829,7 +829,7 @@ def user_agent(self):
             try:
                 # Solver.user_agent has to be a static or class method
                 user_agent_str += f" {_get_solver_class().user_agent()}"
-            except (AttributeError, TypeError):
+            except (AttributeError, TypeError, ImportError):
```
jezdez, line 832:
```suggestion
            except Exception:
```

jaimergp:
@jezdez - do we want to log the exception via `log.debug` or something? It could be painful to debug otherwise.

jezdez:
> @jezdez - do we want to log the exception via `log.debug` or something? It could be painful to debug otherwise.

Yeah, good idea.

beeankha:
@anaconda-issue-bot check

jezdez (approved):

PR #11481, opened 2022-05-17
CI: Add Python 3.11 jobs to matrix

EwoutH:
Adds Python 3.11 jobs to the CI build matrix. Python 3.11 is currently in beta and no new features will be added. By adding Python 3.11 to the CI at this point, we will be able to catch bugs and incompatibilities with Python 3.11 before its stable release.

Part of #11170.


jezdez:
To elaborate a bit why the CI image job is failing for 3.11: We're relying on the existence of the Python 3.11 package on the conda defaults channel, which means this is heavily relying on the existence of that work before we can continue with this PR.

/cc @chenghlee About 3.11 support.

EwoutH:
> We're relying on the existence of the Python 3.11 package on the conda defaults channel 

Thanks for getting back! Is there any place (i.e. issue) where this process can be tracked? (cc @chenghlee)

I also just rebased this PR

EwoutH:
No problem, thanks for the explanation and working on it!

Issue #3097, opened 2016-07-17
request: conda "rename" (or conda "move") for changing name of an environment


TomLav:
All is in the title :)

Today I wanted to rename my historic "dev" environment to "devel".

This did the trick:

`conda create --name devel --clone dev`
`conda remove --name dev --all`

This however seemed to (temporarily) duplicate disk usage. And maybe a specific command could do something smarter than "copy + remove", I do not know.

T.

eduartua:
Any advancement with this?

wandonye:
Is it ok to just change the name of the folder under env?

kaumaron:
I just tried your suggestion wandonye and it seems to be working so far.

H0R5E:
+1

shachar-i:
+1

ashander:
Yes this would be great. Not only does the work around posted above temporarily duplicate disk usage but it hits remote repos and uses network

dlime:
+1 :)

DheerajAgarwal:
+1

braham-snyder:
Though I too would like to see this feature implemented, "+1" posts send every subscriber a useless notification--I (and I suspect many other subscribers) would greatly appreciate it if you instead gave the OP a thumbs up (by clicking the smiley face in its top right corner).

(In case you didn't notice, GitHub's issue tracker can sort issues by the numbers of reactions they've garnered--ideally, IMO, we'd have explicit issue voting with votes displayed next to the number of comments as in https://github.com/isaacs/github/issues/9 or, e.g., GitLab, but that doesn't look like it will happen in the near future.)

ijstokes:
@kalefranz I'd like to suggest that this is a duplicate of #1845

sebma:
Hi,
I tried renaming my environment directory name in the filesystem from `test` to `test3` and inside `$HOME/.conda/environments.txt` file but it does not seem enough : 
```
$ source activate test2
$ ipython -V
-bash: /usr/local/miniconda3/envs/test3/bin/ipython: /usr/local/miniconda3/envs/test/bin/python: bad interpreter: No such file or directory
```

Any idea ?

sebma:
I found why : 

I seems the path to environment is included in the shebang of the scripts in the environment bin directory and also in other files.

Therefore unless script a recursive search'n'replace, rename the environment directory is a VERY BAD IDEA.

I won't do it next time, I promise :) !

sebma:
[...]

PR #11496, opened 2022-05-23
Add new `rename` command

travishathaway:
This PR is an implementation related to the following issue: https://github.com/conda/conda/issues/3097

## Details

This PR adds a new sub-command called `conda rename`. To get an idea of how it works, please see the following new man-page:

```
usage: conda rename [-h] [-n ENVIRONMENT | -p PATH] [--force] [-d] destination

Renames an existing environment

Options:

positional arguments:
  destination           New name for the conda environment

optional arguments:
  -h, --help            Show this help message and exit.
  --force               force creation of environment (removing a previously existing environment of the same name).
  -d, --dry-run         Only display what would have been done.

                        Name of environment.
  -p PATH, --prefix PATH
                        Full path to environment location (i.e. prefix).

examples:
    conda rename -n test123 test321
    conda rename --name test123 test321
    conda rename -p path/to/test123 test321
    conda rename --prefix path/to/test123 test321
```

## How to test

The new tests were written as integration tests in [tests/cli/test_main_rename.py](https://github.com/travishathaway/conda/blob/feature/conda-rename/tests/cli/test_main_rename.py).

So long as these tests pass, everything should be working. As you review this PR, please check there for any missing tests.

## Good to know

The changes to [tests/conda_env/test_cli.py](https://github.com/travishathaway/conda/blob/feature/conda-rename/tests/conda_env/test_cli.py) were made because we originally thought these were where the `rename` tests were going to live. After realizing this was not the case, rather than undoing our refactor efforts, we decided to leave the changes in place.

### DIFF
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -83,7 +83,8 @@ def is_active_prefix(prefix):
     Can be used a validation step to make sure operations are not being
     performed on the active prefix.
     """
-    return prefix == context.active_prefix or prefix == env_name(context.active_prefix)
+    active_prefix = context.active_prefix
+    return paths_equal(prefix, active_prefix) or prefix == env_name(context.active_prefix)
 
 
 def arg2spec(arg, json=False, update=False):
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -11,7 +11,7 @@ from warnings import warn
 
 from ..auxlib.ish import dals
 from ..base.constants import ROOT_ENV_NAME
-from ..base.context import context
+from ..base.context import context, env_name
 from ..common.constants import NULL
 from ..common.io import swallow_broken_pipe
 from ..common.path import paths_equal
@@ -76,6 +76,16 @@ def ensure_name_or_prefix(args, command):
         from ..exceptions import CondaValueError
         raise CondaValueError('either -n NAME or -p PREFIX option required,\n'
                               'try "conda %s -h" for more details' % command)
+
+def is_active_prefix(prefix):
+    """
+    Determines whether the args we pass in are pointing to the active prefix.
+    Can be used a validation step to make sure operations are not being
+    performed on the active prefix.
+    """
+    active_prefix_name = env_name(context.active_prefix)
+
+    return prefix == context.active_prefix or prefix == active_prefix_name
 
 
 def arg2spec(arg, json=False, update=False):
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from logging import getLogger
-from os.path import basename, dirname, isdir, isfile, join
+from os.path import basename, dirname, isdir, isfile, join, normcase
 import re
 import sys
 from warnings import warn
@@ -83,8 +83,11 @@ def is_active_prefix(prefix):
     Can be used a validation step to make sure operations are not being
     performed on the active prefix.
     """
-    active_prefix = context.active_prefix
-    return paths_equal(prefix, active_prefix) or prefix == env_name(context.active_prefix)
+    return (
+        paths_equal(prefix, context.active_prefix)
+        # normcasing our prefix check for Windows, for case insensitivity
+        or normcase(prefix) == normcase(env_name(context.active_prefix))
+    )
 
 
 def arg2spec(arg, json=False, update=False):
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -83,9 +83,7 @@ def is_active_prefix(prefix):
     Can be used a validation step to make sure operations are not being
     performed on the active prefix.
     """
-    active_prefix_name = env_name(context.active_prefix)
-
-    return prefix == context.active_prefix or prefix == active_prefix_name
+    return prefix == context.active_prefix or prefix == env_name(context.active_prefix)
 
 
 def arg2spec(arg, json=False, update=False):
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -54,14 +54,17 @@ def check_prefix(prefix, json=False):
                           "Spaces in paths can sometimes be problematic." % prefix)
 
 
-def clone(src_arg, dst_prefix, json=False, quiet=False, index_args=None):
+def clone(src_arg, dst_prefix, json=False, quiet=False, index_args=None, use_context=True):
     if os.sep in src_arg:
         src_prefix = abspath(src_arg)
         if not isdir(src_prefix):
             raise DirectoryNotFoundError(src_arg)
     else:
-        assert context._argparse_args.clone is not None
-        src_prefix = locate_prefix_by_name(context._argparse_args.clone)
+        if use_context:
+            assert context._argparse_args.clone is not None
+            src_prefix = locate_prefix_by_name(context._argparse_args.clone)
+        else:
+            src_prefix = locate_prefix_by_name(src_arg)
 
     if not json:
         print("Source:      %s" % src_prefix)
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -54,17 +54,13 @@ def check_prefix(prefix, json=False):
                           "Spaces in paths can sometimes be problematic." % prefix)
 
 
-def clone(src_arg, dst_prefix, json=False, quiet=False, index_args=None, use_context=True):
+def clone(src_arg, dst_prefix, json=False, quiet=False, index_args=None):
     if os.sep in src_arg:
         src_prefix = abspath(src_arg)
         if not isdir(src_prefix):
             raise DirectoryNotFoundError(src_arg)
     else:
-        if use_context:
-            assert context._argparse_args.clone is not None
-            src_prefix = locate_prefix_by_name(context._argparse_args.clone)
-        else:
-            src_prefix = locate_prefix_by_name(src_arg)
+        src_prefix = locate_prefix_by_name(src_arg)
 
     if not json:
         print("Source:      %s" % src_prefix)
--- a/conda/cli/main_rename.py
+++ b/conda/cli/main_rename.py
@@ -61,11 +61,11 @@ def safe_rm_destination(directory: str):
     tmpdir = tempfile.mkdtemp()
 
     try:
-        rename(directory, tmpdir)
+        rename(directory, tmpdir, force=True)
         yield
     except Exception as exc:
         # Error occurred, roll back change
-        rename(tmpdir, directory)
+        rename(tmpdir, directory, force=True)
         raise exc
 
 
--- a/conda/cli/main_rename.py
+++ b/conda/cli/main_rename.py
@@ -7,12 +7,11 @@ import os
 from typing import Callable
 
 from conda.base.context import context, locate_prefix_by_name, validate_prefix_name
+from conda.base.constants import DRY_RUN_PREFIX
 from conda.cli import common, install
 from conda.common.path import expand
 from conda.exceptions import CondaEnvException
 from conda.gateways.disk.delete import rm_rf
-
-DRY_RUN_PREFIX = "Dry run action:"
 
 
 def validate_src(args) -> str:
@@ -66,7 +65,7 @@ def execute(args, _):
         (
             install.clone,
             (src, dest),
-            {"quiet": context.quiet, "json": context.json, "use_context": False},
+            {"quiet": context.quiet, "json": context.json},
         )
     )
     actions.append((rm_rf, (src,), {}))
--- a/conda/cli/main_rename.py
+++ b/conda/cli/main_rename.py
@@ -3,15 +3,18 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
+import tempfile
+from contextlib import contextmanager
 from functools import partial
 import os
 
-from conda.base.context import context, locate_prefix_by_name, validate_prefix_name
-from conda.base.constants import DRY_RUN_PREFIX
-from conda.cli import common, install
-from conda.common.path import expand
-from conda.exceptions import CondaEnvException
-from conda.gateways.disk.delete import rm_rf
+from ..base.context import context, locate_prefix_by_name, validate_prefix_name
+from ..base.constants import DRY_RUN_PREFIX
+from ..cli import common, install
+from ..common.path import expand, paths_equal
+from ..exceptions import CondaEnvException
+from ..gateways.disk.delete import rm_rf
+from ..gateways.disk.update import rename
 
 
 def validate_src(args) -> str:
@@ -19,9 +22,7 @@ def validate_src(args) -> str:
     Validate that we are receiving at least one value for --name or --prefix
     and ensure that the "base" environment is not being renamed
     """
-    common.ensure_name_or_prefix(args, "env rename")
-
-    if context.target_prefix == context.root_prefix:
+    if paths_equal(context.target_prefix, context.root_prefix):
         raise CondaEnvException("The 'base' environment cannot be renamed")
 
     prefix = args.name if args.name else args.prefix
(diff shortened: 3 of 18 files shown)


kenodegard (changes requested):

docs/source/dev-guide/deep-dive-solvers.md
```diff
@@ -222,13 +222,12 @@ Explicit installs are taken care of by the [`explicit`][conda.misc:explicit] fun
...
-using the usual means (`cp`, `mv` or your favorite file manager): this will break your
-environment! Conda _environments_ are _not_ relocatable; some files might contain hardcoded
-paths to existing files in the original location, and those references will break with a
-rename. You can create a new environment anywhere you want, but once created they shall not
-change locations.
+`conda create` has a `--clone` flag that allows you to create a fully-working copy of an
+existing environment. This is needed because you cannot relocate an environment using `cp`,
+`mv`, or your favorite file manager without unintended consequences. Some files in a conda
+environment might contain hardcoded paths to existing files in the original location, and
+those references will break if `cp` or`mv` is utilized (conda environments _can_ be renamed
```
kathatherine, line 229:
```suggestion
those references will break if `cp` or `mv` is utilized (conda environments _can_ be renamed
```

beeankha: Good catch!!

news/11496-new-rename-command
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Implementation of a new `rename` subcommand
```
kenodegard, line 3:
```suggestion
* Implementation of a new `rename` subcommand. (#11496)
```

news/11496-new-rename-command
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* Corresponding documentation of `rename` subcommand
```
kenodegard, line 15:
```suggestion
* Corresponding documentation of `rename` subcommand. (#11496)
```

conda/cli/main_rename.py
```diff
@@ -0,0 +1,80 @@
...
+import os
+from typing import Callable
+
+from conda.base.context import context, locate_prefix_by_name, validate_prefix_name
+from conda.cli import common, install
+from conda.common.path import expand
+from conda.exceptions import CondaEnvException
+from conda.gateways.disk.delete import rm_rf
+
+DRY_RUN_PREFIX = "Dry run action:"
```
kenodegard, line 15: Since this is only used once let's bring it inline or make larger plans for it and pop it up into constants.

conda/cli/main_rename.py
```diff
@@ -0,0 +1,80 @@
...
+        actions.append((rm_rf, (dest,), {}))
+
+    actions.append(
+        (
+            install.clone,
+            (src, dest),
+            {"quiet": context.quiet, "json": context.json, "use_context": False},
+        )
+    )
+    actions.append((rm_rf, (src,), {}))
```
kenodegard, line 72: This is a personal styling preference that can be dismissed. But if we do go this route we need to include a `if not act_func` in the below `for`-loop.

```suggestion
    actions: list[tuple[Optional[Callable], Args, Kwargs]] = [
        (rm_rf, (dest,), {}) if args.force else (None, (), {}),
        (
            install.clone,
            (src, dest),
            {"quiet": context.quiet, "json": context.json, "use_context": False},
        ),
        (rm_rf, (src,), {}),
    ]

```

conda/cli/main_rename.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def execute(args, _):
+    """
+    Executes the command for renaming an existing environment
+    """
+    src = validate_src(args)
+    dest = validate_destination(args.destination, force=args.force)
+
+    actions: list[tuple[Callable, Args, Kwargs]] = []
```
kenodegard, line 60: What about using `functools.partial` instead of `tuple[Callable, Args, Kwargs]`?

conda/cli/main_rename.py
```diff
@@ -0,0 +1,80 @@
...
+            {"quiet": context.quiet, "json": context.json, "use_context": False},
+        )
+    )
+    actions.append((rm_rf, (src,), {}))
+
+    # We now either run collected actions or print dry run statement
+    for act_func, act_args, act_kwargs in actions:
+        if args.dry_run:
+            pos_args = ", ".join(act_args)
+            print(f"{DRY_RUN_PREFIX} {act_func.__name__} {pos_args}")
```
kenodegard, line 78: Is `act_kwargs` being intentionally ignored here?

conda/cli/install.py
```diff
@@ -54,14 +54,17 @@ def check_prefix(prefix, json=False):
...
         if not isdir(src_prefix):
             raise DirectoryNotFoundError(src_arg)
     else:
-        assert context._argparse_args.clone is not None
-        src_prefix = locate_prefix_by_name(context._argparse_args.clone)
+        if use_context:
+            assert context._argparse_args.clone is not None
+            src_prefix = locate_prefix_by_name(context._argparse_args.clone)
+        else:
+            src_prefix = locate_prefix_by_name(src_arg)
```
kenodegard, line 67:
```python
elif use_context:
    assert context._argparse_args.clone is not None
    src_prefix = locate_prefix_by_name(context._argparse_args.clone)
else:
    src_prefix = locate_prefix_by_name(src_arg)
```

kenodegard: Wait, isn't `context._argparse_args.clone == src_arg`?

The `clone` method is only called in one place previously:
https://github.com/conda/conda/blob/ae389d58d67dd8e29952e669c34f1495d034185a/conda/cli/install.py#L217

And `context._argparse_args` _is_ the parsed args as set here:
https://github.com/conda/conda/blob/ae389d58d67dd8e29952e669c34f1495d034185a/conda/cli/main.py#L78

So from what I'm reading we can simply flatten all of this and remove the `use_context` arg:
```
else:
    src_prefix = locate_prefix_by_name(src_arg)
```

kenodegard (changes requested):
Lots of good stuff here, a few nits about style/where things should be defined.

conda/cli/main_rename.py
travishathaway: @kenodegard, yes that is true. My reasoning is that it provides too much information to the user that really not that important for understanding the essence of what is going to performed.

conda/cli/install.py
travishathaway: Thanks for this. Yeah, I had made these changes before learning more about the context object itself. I will implement what you suggest.

conda/cli/main_rename.py
travishathaway: Good idea. I will move it constants.

conda/cli/main_rename.py
travishathaway: @kenodegard, having a hard time figure out why one would do that 🤔.  The type hint is trying to say that this will be a list of tuples which have a function, args and kwargs. I know that `functools.partial` has a similar function signature, but I'm not sure it's appropriate to use it in this way.

kathatherine (approved):
LGTM!

kenodegard (approved):
🚀

conda/cli/common.py
```diff
@@ -77,6 +77,14 @@ def ensure_name_or_prefix(args, command):
         raise CondaValueError('either -n NAME or -p PREFIX option required,\n'
                               'try "conda %s -h" for more details' % command)
 
+def is_active_prefix(prefix):
+    """
+    Determines whether the args we pass in are pointing to the active prefix.
+    Can be used a validation step to make sure operations are not being
+    performed on the active prefix.
+    """
+    return prefix == context.active_prefix or prefix == env_name(context.active_prefix)
```
jaimergp, line 86: Might be useful to apply `conda.common.path.paths_equal` here, to ensure cases like:

* Paths with relative markers (`..`) which would be equivalent
* Symlinked locations (some people do that... Y_Y)

conda/cli/main_rename.py
```diff
@@ -0,0 +1,66 @@
...
+from conda.exceptions import CondaEnvException
+from conda.gateways.disk.delete import rm_rf
+
+
+def validate_src(args) -> str:
+    """
+    Validate that the "base" environment is not being renamed and that the
+    currently active prefix isn't trying to be renamed.
+    """
+    if context.target_prefix == context.root_prefix:
```
jaimergp, line 22: Same here, use `paths_equal`.

conda/cli/main_rename.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Executes the command for renaming an existing environment
+    """
+    src = validate_src(args)
+    dest = validate_destination(args.destination, force=args.force)
+
+    actions: list[partial] = []
+
+    if args.force:
+        actions.append(partial(rm_rf, dest))
```
jaimergp, line 56: Should this be atomic? What if the rename ends up failing? The target environment gets deleted "for nothing". Maybe rename to something else first, and if everything succeeded, end up deleting for good. If something failed, undo the initial rename.

conda/cli/main_rename.py
```diff
@@ -0,0 +1,66 @@
...
+
+    actions.append(partial(install.clone, src, dest, quiet=context.quiet, json=context.json))
```
jaimergp, line 58: [`conda create --clone`](https://github.com/conda/conda/blob/310548bd9f9bb78204fe7d6c3d03334b54bd8083/conda/misc.py#L187) has a series of limitations that might prevent the environment from being renamed. It essentially relies on the package cache to reinstall the packages into a new location. But if the package cache is faulty, it might limit what `clone` is able to do.

There's this other tool, [`conda-prefix-replacement`](https://github.com/conda/conda-prefix-replacement) that might be better suited for this task? I know it's a big ask and I don't want to stall the review here, but I'd like to see if this could be considered, maybe as an optional flag?

kenodegard: @jaimergp AFAIK this has been a known and accepted downside of this approach. A broken/corrupt environment cannot be renamed. I don't think we are prepared ATM to revive `conda-prefix-replacement` to figure out if it does the correct thing. Perhaps we can view this PR as an initial POC and `conda-prefix-replacement` could be refactored into a plugin that replaces the `conda rename` command?

@travishathaway We should make sure these limitations are clearly outlined in the documentation.

conda/cli/main_rename.py
travishathaway: @jaimergp, great recommendation! I just implemented this in a recent commit. Let me know what you think.

conda/cli/main_rename.py
jaimergp: Fine with me!

conda/cli/main_rename.py
```diff
@@ -0,0 +1,94 @@
...
+            if args.dry_run:
+                print(f"{DRY_RUN_PREFIX} {func.func.__name__} {','.join(func.args)}")
+            else:
+                func()
+
+    if args.force:
+        with safe_rm_destination(destination):
+            clone_and_remove()
+    else:
+        clone_and_remove()
```
kenodegard, line 94: If the destination env already exists, specifying `--dry-run --force` will rename the destination to the tempdir.

conda/cli/main_rename.py
```diff
@@ -0,0 +1,94 @@
...
+    tmpdir = tempfile.mkdtemp()
+
+    try:
+        rename(directory, tmpdir)
+        yield
+    except Exception as exc:
+        # Error occurred, roll back change
+        rename(tmpdir, directory)
+        raise exc
+
```
kenodegard, line 68: Can we either make this private (`_safe_rm_destination`), inline it within `execute` below, or abstract it and define it in `conda.gateways.disk.update` (something like `renamecontext`).

I _think_ I prefer the `conda.gateways.disk.update.renamecontext` option since this is very generic functionality and is likely useful elsewhere.

conda/cli/main_rename.py
```diff
@@ -0,0 +1,94 @@
...
+def validate_src(args) -> str:
+    """
+    Validate that the "base" environment is not being renamed and that the
+    currently active prefix isn't trying to be renamed.
+    """
+    if context.target_prefix == context.root_prefix:
+        raise CondaEnvException("The 'base' environment cannot be renamed")
+
+    prefix = args.name if args.name else args.prefix
```
kenodegard, line 28: Since the other validate function explicitly passes in the arguments it uses can we do the same here to be consistent?

```suggestion
def validate_src(name: str | None, prefix: str | None) -> str:
    """
    Validate that the "base" environment is not being renamed and that the
    currently active prefix isn't trying to be renamed.
    """
    if context.target_prefix == context.root_prefix:
        raise CondaEnvException("The 'base' environment cannot be renamed")

    prefix = name or prefix
```

conda/cli/main_rename.py
```diff
@@ -0,0 +1,94 @@
...
+
+def validate_destination(dest: str, force: bool = False) -> str:
+    """Ensure that our destination does not exist"""
+    if os.sep in dest:
+        dest = expand(dest)
+    else:
+        dest = validate_prefix_name(dest, ctx=context, allow_base=False)
+
+    if not force and os.path.exists(dest):
+        raise CondaEnvException("Environment destination already exists. Override with --force.")
```
kenodegard, line 44:
```suggestion
        raise CondaEnvException(f"Environment destination ({dest}) already exists. Override with --force.")
```

kenodegard (changes requested):
The rename became atomic but running `--dry-run --force` is now unsafe.

conda/cli/main_rename.py
travishathaway: Yeah, that thought definitely crossed my mind too. I will refactor it so that it's available to other modules.

conda/cli/main_rename.py
travishathaway: Need a test for that! 😅

travishathaway: And thanks for point that out.

conda/cli/common.py
```diff
@@ -77,6 +77,15 @@ def ensure_name_or_prefix(args, command):
...
                               'try "conda %s -h" for more details' % command)
 
+def is_active_prefix(prefix):
+    """
+    Determines whether the args we pass in are pointing to the active prefix.
+    Can be used a validation step to make sure operations are not being
+    performed on the active prefix.
+    """
+    active_prefix = context.active_prefix
+    return paths_equal(prefix, active_prefix) or paths_equal(prefix, env_name(active_prefix))
```
beeankha, line 87: FYI, the reason I use the `active_prefix` variable here is because otherwise the linter repeatedly tries to re-format this bit of code to look like:
```
    return paths_equal(prefix, context.active_prefix) or paths_equal(
        prefix, env_name(context.active_prefix)
    )
```
...which I didn't think was super readable. This happened no matter how I reformatted the line; even something like the below didn't escape the linter's reformatting:
```
    return paths_equal(prefix, context.active_prefix) \
    or paths_equal(prefix, env_name(context.active_prefix))
```

kenodegard: Extra parentheses should do the trick:

```suggestion
    return (
        paths_equal(prefix, context.active_prefix)
        or paths_equal(prefix, env_name(context.active_prefix))
    )
```

beeankha: `darker` is still insisting on this formatting:
```
    return paths_equal(prefix, context.active_prefix) or paths_equal(
        prefix, env_name(context.active_prefix)
    )
```

Is there any way to bypass this in `darker` without changing default checks for everything else?

kenodegard: Ugh, black 🤦‍♂️

Adding comments like this will preserve the formatting, but the comments I came up with feel unnecessary

```python
    return (
        # check if the prefix is the active env path
        paths_equal(prefix, context.active_prefix)
        # check if the prefix is the active env name
        or prefix == env_name(context.active_prefix)
    )
```

Can also explicitly disable formatting, but personally, I don't think that's worth it, at this point, we might as well leave it with the `active_prefix` variable or just let black do its thing and leave it a little weird 🤷‍♂️

```python
    # fmt: off
    return (
        paths_equal(prefix, context.active_prefix)
        or prefix == env_name(context.active_prefix)
    )
    # fmt: on
```

Also, not sure we want to use `paths_equal` for the second condition?

kenodegard (approved):

wolfv:
Sorry for chiming in late -- wouldn't it be cleaner to have this as `conda env rename`? So that it's clear that an environment is renamed?

stardiviner:
I propose this suggestion too.

wolfv:
I’m all for deprecating conda env create, but other sub commands (such as conda env list or conda env delete) seem reasonable to me.

PR #11215, opened 2022-02-12
Allow `conda run` to work in read-only environments.

erykoff:
This fixes #11208 by using the system temp file directory instead of trying to create one in the conda environment when executing `conda run`.

Issue #11208, opened 2022-02-11
conda run crashes when targeted to an env on a read-only filesystem/path.


erykoff:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help
- [X] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?

If you try to do `conda run` into a read-only filesystem/env (as is sometimes used in shared environments) it crashes because it can't write a temporary file to the env location.

For example ...

This runs fine:
```
conda create -n testpy python=3.9
conda run -n testpy python
```
Then make it read-only:
```
chmod -w $CONDA_PREFIX/envs/testpy
conda run -n testpy python
```
And you get:
```
% conda run -n testpy python        

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/cli/conda_argparse.py", line 83, in do_call
        return getattr(module, func_name)(args, parser)
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/cli/main_run.py", line 25, in execute
        script_caller, command_args = wrap_subprocess_call(on_win, context.root_prefix, prefix,
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/utils.py", line 403, in wrap_subprocess_call
        with Utf8NamedTemporaryFile(mode='w', prefix=tmp_prefix, delete=False) as fh:
      File "/Users/erykoff/miniforge3/lib/python3.9/site-packages/conda/_vendor/auxlib/compat.py", line 81, in Utf8NamedTemporaryFile
        return NamedTemporaryFile(mode=mode, buffering=buffering, encoding=encoding,
      File "/Users/erykoff/miniforge3/lib/python3.9/tempfile.py", line 541, in NamedTemporaryFile
[...]

Issue #10690, opened 2021-05-26
Conda run fails with Permission denied if environment is read-only.


mforbes:
## Current Behavior

Trying to execute `conda run` attempts to write to the directory where the environment is install, which leads to a failur with a `Permission denied` error when the environment is in a read-only location (i.e. installed on a computing cluster, etc. where users do not have access.)

### Steps to Reproduce

```bash
$ conda create -yn tst python
$ conda env list
...
tst                      /data/apps/conda/envs/tst
...
$ find /data/apps/conda/envs/tst -exec chmod a-w {} \;
# Or more simply, just create the environment as a different user.
$ conda run -n tst python
```

<details open><summary>Full output</summary>

```bash
$ conda run -n tst python 

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/data/apps/conda/lib/python3.9/site-packages/conda/exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "/data/apps/conda/lib/python3.9/site-packages/conda/cli/main.py", line 84, in _main
        exit_code = do_call(args, p)
      File "/data/apps/conda/lib/python3.9/site-packages/conda/cli/conda_argparse.py", line 83, in do_call
        return getattr(module, func_name)(args, parser)
      File "/data/apps/conda/lib/python3.9/site-packages/conda/cli/main_run.py", line 25, in execute
        script_caller, command_args = wrap_subprocess_call(on_win, context.root_prefix, prefix,
      File "/data/apps/conda/lib/python3.9/site-packages/conda/utils.py", line 403, in wrap_subprocess_call
        with Utf8NamedTemporaryFile(mode='w', prefix=tmp_prefix, delete=False) as fh:
      File "/data/apps/conda/lib/python3.9/site-packages/conda/_vendor/auxlib/compat.py", line 81, in Utf8NamedTemporaryFile
        return NamedTemporaryFile(mode=mode, buffering=buffering, encoding=encoding,
      File "/data/apps/conda/lib/python3.9/tempfile.py", line 541, in NamedTemporaryFile
        (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
      File "/data/apps/conda/lib/python3.9/tempfile.py", line 251, in _mkstemp_inner
        fd = _os.open(file, flags, 0o600)
[...]

PR #11520, opened 2022-06-02
Control location of temporary directory via environmental variable

cswartzvi:
Currently, I work in an office that has some pretty strict security regulations, one of them is a group policy that forbids execution in the user and system temp folders - we are using Windows based systems. Because of this, `conda run` is blocked in both CMD and PowerShell, and `conda activate` is blocked in CMD. This is especially problematic for vscode python integration (which uses `conda run`) and interacting with environments that need to run batch files in ` $CONDA_PREFIX/etc/conda/activate.d` (because these will only run when activating the environment in CMD).

Following on from #11208 and the subsequent fix in #11215, I would like to introduce an option to toggle the location of the temp directory. This PR introduces an environmental variable `CONDA_USE_PREFIX_TEMP` (much like `CONDA_TEST_SAVE_TEMPS`) that, when defined, will instruct `auxlib.compat.Utf8NamedTemporaryFile` to create temporary files in `sys.prefix`. 

First time contributor here - I apologize if anything is out of order 😳.

### Checklist - did you ...

### DIFF
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -349,12 +349,9 @@ def wrap_subprocess_call(
         dev_mode,
         debug_wrapper_scripts,
         arguments,
-        use_system_tmp_path=False):
+):
     arguments = massage_arguments(arguments)
-    if not use_system_tmp_path:
-        tmp_prefix = abspath(join(prefix, '.tmp'))
-    else:
-        tmp_prefix = None
+    tmp_prefix = ".tmp"
     script_caller = None
     multiline = False
     if len(arguments) == 1 and '\n' in arguments[0]:
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -351,7 +351,7 @@ def wrap_subprocess_call(
         arguments,
 ):
     arguments = massage_arguments(arguments)
-    tmp_prefix = ".tmp"
+    tmp_prefix = ".conda_"
     script_caller = None
     multiline = False
     if len(arguments) == 1 and '\n' in arguments[0]:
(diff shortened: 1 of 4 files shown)


jezdez:
@conda-bot check

conda/utils.py
```diff
@@ -352,12 +352,9 @@ def wrap_subprocess_call(
...
         debug_wrapper_scripts,
         arguments,
-        use_system_tmp_path=False):
+):
     arguments = massage_arguments(arguments)
-    if not use_system_tmp_path:
-        tmp_prefix = abspath(join(prefix, '.tmp'))
-    else:
-        tmp_prefix = None
+    tmp_prefix = ".tmp"
```
cswartzvi, line 357: I think some explanation is required here. I came to the same conclusion as https://github.com/conda/conda/issues/10690#issuecomment-1031429467 and adjusted the `prefix` that is passed into `Utf8NamedTemporaryFile` so that it does not contain an absolute path - and therefore does not specify a location.

cswartzvi:
Drat. I know why those 'unit 1' tests are failing, I don't have write permissions for `sys.prefix` - that's what I get for not thinking beyond my system 🤦🏻‍♂️. Let me update the tests.

cswartzvi:
I believe that [d6f9f76](https://github.com/conda/conda/pull/11520/commits/d6f9f7656542238462737baf448542ab54b27870) should allow the 'unit 1' tests to pass on Linux.

cswartzvi:
This might be better as a full blown configuration option in `.condarc` as opposed to an environmental variable. Thoughts?

PR #11528, opened 2022-06-06
Add note about disabling adding to PATH during AllUsers installation (on Windows)

pseudoyim:
### Description

Add note to inform Windows users that they can no longer add to PATH when running in AllUsers mode. This note should help avoid more issues like this: https://github.com/ContinuumIO/anaconda-issues/issues/12956

The commit where the security fix was implemented: https://github.com/ContinuumIO/anaconda-distribution-installer/commit/301e84f84b63d654045d4d7871b726de39fc9bb5

### Checklist - did you ...


csoja (approved):
Thanks for adding this!

kathatherine (approved):
LGTM

Issue #10345, opened 2020-11-08
Return versions with `--from-history` flag while creating environment files


regaldude:
### Issue

If I run `conda env export --from-history`, I only get the packages that I have explicitly installed (as expected) but the output doesn't contain package versions. I would consider this bad practice since this can lead to reproducibility issues down the line.

### Suggestion

Command returns output with version numbers as the default behaviour. Define a `--no-vers` flag to let users explicitly opt out of this behaviour.

### Context

At work, most of my development happens on Windows while solutions are deployed on Linux. I had previously ran into an issue where the environment file created on Windows failed to create an environment on Linux. I eventually solved this using the `--from-history` flag. But now I have to manually copy & paste the version numbers in the name of reproducibility 😔

mfansler:
In general, I'd also like to see more control added to exporting. Other than the full export - which is completely appropriate for record keeping - the other two options are poor approximations for what users working cross-platform actually want. Namely, a minimal version of leaf node packages, plus an option to control versions.

Nevertheless, I still think the default behavior for the `--from-history` command makes sense as is, i.e., without versions. It captures exactly the explicit specifications the user has provided in the history. I think the argument could be made that if the user wanted a specific version, this should have been placed in the explicit specifications during installation. Additionally, one can adjust explicit specs with `conda install --update-specs`. However, I wouldn't be opposed to adding an opt-in option for `--from-history` to include versions.

@regaldude In the meantime, as an alternative, perhaps you might be interested in [the `conda-minify` package](https://github.com/jamespreed/conda-minify). It still seems like it doesn't hit exactly your middle ground of wanting only the user-requested packages as well as having versions, but it maybe it will fit your needs better than `--from-history`.

eiphy:
@regaldude I made a simple package to achieve this. See: https://github.com/eiphy/conda-export

regaldude:
Looks nice, @eiphy. Will take a look at this when I have some time.

maeghan-porter:
> Nevertheless, I still think the default behavior for the --from-history command makes sense as is, i.e., without versions. It captures exactly the explicit specifications the user has provided in the history. I think the argument could be made that if the user wanted a specific version, this should have been placed in the explicit specifications during installation.

I'm going to disagree on that one. Here is my personal example:
[...]

PR #11532, opened 2022-06-09
Fix #9628 and add more conda env export flexibility

### LINKED ISSUE #9628: include pip installs in `conda env export --from-history`
Hi - 

I like the idea of the --from-history tag for conda env export for cross-platform compatibility, and simpler looking .yml files. However, it does not export the pip installed packages, as it normally would without the --from-history tag. Since pip is often still used to install packages not available on conda channels, the utility of the --from-history tag is greatly minimized.

Wondering if it would be possible to track history of pip installs. If it is difficult to do this in the same way as for conda packages, i wonder if there could be an additional argument, e.g. '--include-pip-in-history', that would just append all the pip installed packages to the output of conda env export when --from-history is used.

Thanks a lot
Nick

jeskowagner:
### Description
This commit is a draft to address #9628. It needs further work to address tests and complete the checklist (see below).
In summary, it introduces more flexibility to `conda env export` in the form of two new flags, and adds new functionality to the existing `conda env export --from-history` in that this will now include pip dependencies.

The more detailed view is that three changes were made:

- Make `--from-history` respect `pip` dependencies. This needed some deeper changes and is the reason this commit is only a draft: it currently fails the test `test_from_history`, because that test does not fully emulate a "normal" conda install fully, leading
to no packages detected. I need help addressing that with a more targeted test.
- Introduce `--override-prefix` flag to `conda env export`, to allow for better portability to other systems
- Introduce `--ignore-pip` flag to `conda env export`, to allow control over pip dependencies if desired

### Checklist

- [ ] Add / update necessary tests? **No, PR fails tests, see Description**


jeskowagner:
I added an `xfail` to the unit test for `conda env export --from-history`. Reason is that now the logic for `--from-history` is different and not as easy to test using `patch`. With this addition, the PR could be merged as is or after adding a hint to this in the documentation.

jeskowagner:
@jezdez Squashed and rebased, 10fe774 should be ready for review.

Btw, I see that black accidentally reformatted some `'` to `"`. Is this a showstopper?

jeskowagner:
Sorry, had missed the pre-commit errors. Should be fixed now. Would appreciate a review @jezdez

dholth:
@jeskowagner I've tried to complete the merge in #12083, feel free to merge that into this PR if it LGTY

dholth:
By the way refomatting ' to " is a show_starter_ in my book

jeskowagner:
Thanks @dholth, I merged those changes and also merged in new changes from conda. Hopefully this time all CI passes nicely so we can get these features to users.

jeskowagner:
All pre-CI checks passed and no merges remain. Could someone please approve the CI workflow? @dholth, do you have authorisation for this?

jeskowagner:
Well spotted, thank you.

For my understanding, could you please clarify what the optimal expected behavior would be? So far, versions not being included in `conda env export --from-history` is inconsistent, in my opinion. Having build numbers is the default for `conda env export` (without the `--from-history` flag), and there is no documentation to suggest this should not be the case (see also #10345). That said, I appreciate that also in that issue there is differing opinions about it. I suppose we could add an opt-in flag to include version/build numbers in `--from-history` output, however, that makes documentation very confusing in my eyes (why would `conda env export` have a `--include-version` flag when that is the default, a user might ask).

AlbertDeFusco:
Yes, I see where things can be confusing. I'm taking the view that from history will record what was originally requested on the command line, at least by default. I can write up more with examples later

jeskowagner:
I see your point. If desired, we can change this quite easily by modifying
```python
if no_builds:
    dependencies = ["=".join((a.name, a.version)) for a in conda_precs]
else:
    dependencies = ['='.join((a.name, a.version, a.build)) for a in conda_precs]
```
to 

```python
if from_history:
    dependencies = [a.requested_spec for a in conda_precs]
elif no_builds:
    dependencies = ["=".join((a.name, a.version)) for a in conda_precs]
else:
    dependencies = ['='.join((a.name, a.version, a.build)) for a in conda_precs]
```

With that change, the environment you gave with the call `conda env export --from-history` gives
```yaml
name: export-test
channels:
  - conda-forge
dependencies:
  - pip
  - python=3.9
  - pip:
      - requests==2.27.0
prefix: <redacted>
```

I remain on the fence which approach is better, to be honest. Obviously a regression from the status quo is in favor of this approach, but then we can never make a fully reproducible environment if packages were under-specified. Unless, of course, one leaves out the `--from-history` flag. However, I tend to always include it, because the dependency grows very long very quickly, and often leads to solver problems.

AlbertDeFusco:
Yes, I agree that reproducibility is challenging and I'm in favor of providing more options to `conda env export` to cover more needs and `--history-include-version` seems fine to me.

These challenges have led me to develop [conda-project](https://github.com/conda-incubator/conda-project/blob/main/docs/source/user_guide.md) recently, which utilizes [conda-lock](https://github.com/conda-incubator/conda-lock). The concept is to combine the output of `conda env export --from-history` (using `requested_spec` variant), which I refer to as the _intentional_ spec with the _fully specified_ conda-lock file that has the added advantage of being cross-platform (the lockfile contains what is essentially `conda env export` for any or all platforms supported by Conda).

jeskowagner:
That's neat! So what should be the basic advice for conda users that have one or more of the problems listed in #11800? That is, do we want to move forward with `--history-include-version`, or is the current implementation fine? I am asking because I would like to avoid another round of changes, tests etc only to then hear that Anaconda does not approve of them. I think many people will benefit from the changes, yet it has been half a year since I opened the PR.

Thrillz:
> That's neat! So what should be the basic advice for conda users that have one or more of the problems listed in #11800? That is, do we want to move forward with `--history-include-version`, or is the current implementation fine? I am asking because I would like to avoid another round of changes, tests etc only to then hear that Anaconda does not approve of them. I think many people will benefit from the changes, yet it has been half a year since I opened the PR.

I believe `--history-include-version` is absolutely necessary if I want to share and be able to reproduce results on a different machine/environment. Please Include this implementation if possible @jeskowagner thanks 🙏

jeskowagner:
I see that @jezdez and @dholth have been assigned now. I would like to wait with changes to hear their view on preferred implementations, because ultimately @Anaconda will have to approve any changes.

jeskowagner:
@jezdez and @dholth, could I please get an update about your preferred implementation? I would like to move forward with this PR as it's a feature that I could use frequently. I can make the required changes but would like a statement about preferred implementation first. Thanks.

pgramme:
I don't really understand the need for `--history-include-version`.

To me the interest of `--from-history` is to report the _intent_ of the person who installed packages in the environment. So the current behavior (including version numbers only if provided in command-line) is great.

The motivation for proposing `--history-include-version` is to have reproducibility, but it is actually not sufficient. Indeed, running `conda create -n test pandas=1.3.0` now vs one year ago would produce two different environments, due to updated versions of the dependencies which have been published in the meanwhile and will be preferred now. Concrete example of difference: I now get intel-openmp=2023.1.0, which wasn't available last year.

The only way to have real reproducibility is to avoid using `--from-history` at all, IMHO.

But aside from this, I love the added functionality of reporting pip installs!

jeskowagner:
@pgramme Fair points. I think for many users `--from-history` is the starting point to derive a slim `yaml` that they can share with others. The issue with a "default" `conda list` or `conda env export` is that the produced files are essentially irreproducible, because they contain a lot of hardware-specific dependencies. 

As you say, `--from-history` brings the advantage of specifying the packages with the intent of the author.

`--history-include-version` similarly would allow people to share a minimal, but fully versioned `yaml` that others can use to get started with all the right versions. 

In an ideal world, the autor of the `yaml` has installed all the right versions by hand, but realistically this is almost never the case, and as a result, you end up with non-reproducible environments if just  using `--from-history`.

Hope that clarifies the intentions.

FSpanhel:
@jeskowagner I also hope that your PR soon gets approved!

In my team, we often use `conda env export --from-history` to create environment files which can be used to create environments on different platforms.  In our team, we have Windows, Mac and Linux users. Sure we could use docker etc. but the team mainly consists of data analysts whose focus should be analyzing data...

Besides, I am also giving lectures for data science. Here, I also encounter the problem that the students have different hardwares. I also used  `conda env export --from-history` here.

In both cases, it would be really benefical when the version numbers and pip packages could be listed!

jeskowagner:
[The Carpentries](https://carpentries-incubator.github.io/introduction-to-conda-for-data-scientists/04-sharing-environments/index.html) similarly recommend `conda env export --from-history` for device-independent environment definitions. When teaching this course I often get questions about how to include pip packages and fully versioned definitions. Would be nice to be able to have a better answer than "manually change the yaml after the export".

FSpanhel:
As mentioned before, I really think that this PR provides an important feature to conda.

It also seems that all concerns have been addressed?

So would it be possible that this PR is finally approved?

jeskowagner:
I agree that the proposed changes remain useful. However, note that I have not implemented the all changes because it is unclear what flags/changes would be deemed acceptable by @anaconda, see also [the above comment](https://github.com/conda/conda/pull/11532#issuecomment-1434491208). Likewise, I have stopped continously rebasing this feature while there is no timeline for it being reviewed. In all, I disagree that this feature proposal is stale, but warn that it's not ready to merge as is.

PR #11533, opened 2022-06-09
refactor SubdirData._process_raw_repodata_str

dholth:
It's odd that this doesn't accept a dict. conda-index uses SubdirData to generate `current_repodata.json`, and wastes a json encode / decode to send the data to SubdirData. Not tested.

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -489,7 +489,7 @@ class SubdirData(metaclass=SubdirDataType):
         self._names_index = _names_index = defaultdict(list)
         self._track_features_index = _track_features_index = defaultdict(list)
 
-        signatures = json_obj.get("signatures", {})
+        signatures = repodata.get("signatures", {})
 
         _internal_state = {
             'channel': self.channel,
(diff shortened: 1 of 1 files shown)


jezdez (approved):

dholth (commented):
reduce hungarian notation

conda/core/subdir_data.py
```diff
@@ -608,12 +610,12 @@ def read_mod_and_etag(path):
                 match_objects = take(3, re.finditer(REPODATA_HEADER_RE, m))
                 result = dict(map(ensure_unicode, mo.groups()) for mo in match_objects)
                 return result
-        except (BufferError, ValueError, OSError):  # pragma: no cover
+        except (BufferError, ValueError):  # pragma: no cover
```
dholth, line 613: This error was found by the type checker. OSError is handled in the next clause. IOError is a subclass of OSError, so it doesn't make sense to catch (IOError, OSError)

dholth:
https://github.com/conda/conda/blob/3273a2b1b3599ed2df2b606267dc8982e1a30f62/conda/core/subdir_data.py#L560 mutates the `repodata['packages']`. This isn't a problem for conda, but conda-index uses this method to make current_repodata and has to keep an unmodified (JSON serializable) repodata around.

conda/core/subdir_data.py
jezdez: Nice catch!

kenodegard (approved):
🚀

kenodegard:
@dholth hm this touches the same code block that is modified in #11545, conda-content-trust also wants the unmodified repodata so we rely on a deepcopy

dholth:
> @dholth hm this touches the same code block that is modified in #11545, conda-content-trust also wants the unmodified repodata so we rely on a deepcopy

No one's stopping it from continuing to use the old API...

jezdez (approved):

PR #11548, opened 2022-06-14
Manually updated doc conf.py copyright.

kathatherine:
### Checklist - did you ...

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -86,7 +86,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "conda"
-copyright = "2017, Anaconda, Inc"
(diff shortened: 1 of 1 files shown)


jezdez:
Updating the copyright line is an anti-pattern that we shouldn't follow as it indicates the first year of when the document was created (in this case when the documentation was first written and published) and denotes the start of the automatic copyright period (different in various jurisdictions), the Berne Convention applies.

A lot more details can be found here: https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code and of some examples from big corporations speaking publicly about this: https://opensource.google/documentation/reference/copyright

PR #11560, opened 2022-06-20
Update Dockerfile + ci-images flow to build multi arch images

dbast:
### Description

This is done to have linux/aarch64 conda-ci images, which perform better on Apple M1 machines compared to the linux/amd64 images.

The other images (linux/ppc64le, linux/s390x) we basically get for free and prove that the whole tooling is consistently available on all 4 linux platforms.

Once the multi-arch image is published it can be also used during the ci flow to e.g. run a short smoke test on all the non native platforms via emulation.

### Checklist - did you ...


Dockerfile
```diff
@@ -1,17 +1,28 @@
...
+    if [ "${UNAME_M}" = "x86_64" ]; then \
+        MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-${CONDA_VERSION}-Linux-x86_64.sh"; \
+    elif [ "${UNAME_M}" = "s390x" ]; then \
+        MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-${CONDA_VERSION}-Linux-s390x.sh"; \
+    elif [ "${UNAME_M}" = "aarch64" ]; then \
+        MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-${CONDA_VERSION}-Linux-aarch64.sh"; \
+    elif [ "${UNAME_M}" = "ppc64le" ]; then \
+        MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-${CONDA_VERSION}-Linux-ppc64le.sh"; \
+    fi && \
+    wget --quiet $MINICONDA_URL -O ~/miniconda.sh && \
```
jezdez, line 20: This inline shell script is worth abstracting into a script in `dev/linux/`, like `bootstrap.sh` or something. Maybe just pass `UNAME_M` and the `/opt/conda` as parameters?

jezdez (approved):
Neat!

Sidenote: I wish we'd have a Windows Docker setup using the servercore or nanocore image to do the same on Windows :-/

Dockerfile
dbast: Updated the Dockerfile to use built-in args instead of uname and added lots of  comments to make it more clear how the multi-platform stuff works. 

I personally prefer to have the platform/arch related if/else statement within the Dockerfile to have all the important multi-arch related information directly visible on one screen (still less than 100 LOC) without switching files.

dbast:
yay, all green :)

Had to remove pre-commit in tests/requirements.txt as not available for s390x... that can potentially be moved back after somebody converts the pre-commit-feedstock in conda-forge to be noarch (that should be possible with the right arguments). pre-commit anyway runs via the pre-commit-ci service with auto-commit and does not rely on the conda-ci containers.

@jezdez Regarding your windows container note: what would you expect from windows containers, just more speed? That should be also possible by 1. not installing miniconda (it is pre-installed on all public osx, linux, windows github action runners, [see](https://github.com/actions/virtual-environments/blob/main/images/win/scripts/Installers/Install-Miniconda.ps1)) and 2. using the [cache action](https://github.com/actions/cache) with a good cache key, e.g. `key: ${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('test\requirements') }}-${{ hashFiles('dev\windows\setup.bat') }}-${{ DAY_OF_WEEK }}`

Dockerfile
jezdez: > Updated the Dockerfile to use built-in args instead of uname and added lots of comments to make it more clear how the multi-platform stuff works.

This is good, serves the same purpose of making it configurable and not a big blob of script in Dockerfile.

> I personally prefer to have the platform/arch related if/else statement within the Dockerfile to have all the important multi-arch related information directly visible on one screen (still less than 100 LOC) without switching files.

Makes sense, thanks for the comments, this works nicely!

jezdez:
> yay, all green :)
> 
> Had to remove pre-commit in tests/requirements.txt as not available for s390x... that can potentially be moved back after somebody converts the pre-commit-feedstock in conda-forge to be noarch (that should be possible with the right arguments). pre-commit anyway runs via the pre-commit-ci service with auto-commit and does not rely on the conda-ci containers.

Fascinating, might make sense to open a ticket in the feedstock?

FWIW we're relying on pre-commit being installed in the development environment, so https://github.com/conda/conda/blob/775c20cb07753ea399acff13cee2e3529e08c765/docs/source/dev-guide/development-environment.md?plain=1#L110 needs to be updated with a note to install pre-commit until the conda-forge recipe is up-to-date.

> @jezdez Regarding your windows container note: what would you expect from windows containers, just more speed? That should be also possible by 1. not installing miniconda (it is pre-installed on all public osx, linux, windows github action runners, [see](https://github.com/actions/virtual-environments/blob/main/images/win/scripts/Installers/Install-Miniconda.ps1)) and 2. using the [cache action](https://github.com/actions/cache) with a good cache key, e.g. `key: ${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('test\requirements') }}-${{ hashFiles('dev\windows\setup.bat') }}-${{ DAY_OF_WEEK }}`

Yeah, I've been meaning to look into whether we can use the pre-installed miniconda for conda as well and just worry that we end up in a cyclic situation where the CI environment needs to do something different than what the old miniconda did.

Besides that speed component I'm wondering what we can do to further improve the parity between development environments and the CI environment, to reduce the barriers of entry for contributors. A readily available base image might be worth exploring since it'd get Windows devs set up faster. Just a theory though, not really in scope of this PR :)

dbast:
Readded pre-commit and removed s390x from the platform list... s390x can be added later when the pre-commit noarch situation is resolved. 

adding arm64/ppc64le already serves the purpose of a better Apple M1 experience and prove of consistency.

dbast:
all green again. ready for another review / merge. thanks

Issue #11367, opened 2022-03-25
CondaHTTPError: HTTP 000 CONNECTION FAILED


kenodegard:
### Summary

This is a catch-all error that is very unhelpful and very hard to diagnose (although there are a small handful of cases that typically causes it). Collecting reported issues to see if we can make a plan.

### Linked Issues & PRs

- https://github.com/conda/conda/issues/11272
- https://github.com/conda/conda/issues/11126
- https://github.com/conda/conda/issues/11082
- https://github.com/conda/conda/issues/11006
- https://github.com/conda/conda/issues/10711
- https://github.com/conda/conda/issues/10612
- https://github.com/conda/conda/issues/10311
- https://github.com/conda/conda/issues/10291
- https://github.com/conda/conda/issues/10288
- https://github.com/conda/conda/issues/10184
- https://github.com/conda/conda/issues/10170
- https://github.com/conda/conda/issues/10146
- https://github.com/conda/conda/issues/10111
- https://github.com/conda/conda/issues/9948
- https://github.com/conda/conda/issues/9746
- https://github.com/conda/conda/issues/9656
- https://github.com/conda/conda/issues/9555
- https://github.com/conda/conda/issues/9370
- https://github.com/conda/conda/issues/8945
- https://github.com/conda/conda/issues/8656
- https://github.com/conda/conda/issues/5220

jezdez:
Let's stop using `HTTP 000` errors as a way to tell users that something didn't work out during networking/IO.

biologyzhangbo:
I provid a solution to this issue, but it only works in mainland China.
First, delete the line containing "defaults" in the channels. Note that you should delete the entire line.
Next, replace all instances of "https" with "http" in the Tsinghua channel URLs.
Then you will find that your Conda can work normally.

PR #11564, opened 2022-06-21
Separate `SSLError` from `HTTPError`

kenodegard:
### Description

Previously `SSLErrors` were grouped with `HTTPErrors` and hence became more difficult to diagnose since we'd receive a non-obvious HTTP 000 error. This defines a new `CondaSSLErrors` and passes through the `SSLError` for further diagnosis.

See #11367 (not a complete solution as HTTP 000 is still possible).

### Checklist - did you ...


jezdez (changes requested):
This increases the number of our test runs drastically just for one test. Can you think of another way that would successfully monkeypatch this, so the tests are run in place and we don't have as much test matrix sprawl?

kenodegard:
@jezdez I agree, I dislike that I had to implement the tests this way, but I don't see a simple alternative. Trying to properly mock the import as part of the regular unit test suite is not straightforward. The problem is that `conda` gets imported as part of pytest startup (https://github.com/conda/conda/blob/6ed9db722e95234b618e5ec36825c3673ffbf302/tests/__init__.py) so even if the no SSL test is the very first test run, conda has already been imported making any mocking very convoluted.

tests/core/test_subdir_data.py
```diff
@@ -190,7 +193,20 @@ def test_fetch_repodata_remote_request_invalid_noarch(self):
...
+            fetch_repodata_remote_request(url, etag, mod_stamp)
+
+
+def test_no_ssl(mocker):
+    def CondaSession_get(*args, **kwargs):
+        raise SSLError("Got an SSL error")
+
+    mocker.patch.object(CondaSession, "get", CondaSession_get)
+
+    url = "https://www.fake.fake/fake/fake/noarch"
```
dholth, line 205: this url might be fake

jezdez (approved):

PR #11576, opened 2022-06-23
Fix GIT_DESCRIBE_REGEX to strip non-numeric characters at beginning

ryanvolz:
### Description

This fixes `GIT_DESCRIBE_TAG_PEP440` when the tag includes a prefix composed of lowercase letters, e.g. "v1.0.0" -> "1.0.0".

This regex is used to parse the output of `git describe` and turn it into a version, post commit number, and commit hash that can be formatted into a PEP440-compliant version number. The first part of this regex was presumably intended to match letters, '_', or '-' at the beginning of the string to strip them off exclusive of the version number. The pattern was incorrect, however, with an unescaped '-' being interpreted as a character range instead of the literal dash character. The result was that all lowercase letters except 'z' would not be stripped from the version.

The common case that this fixes is a git version tag that has a 'v' prefix, like "v1.0.0". Previously, this would result in an invalid version number of "v1.0.0". With this fix, the version number becomes "1.0.0".

The following test case shows the change:
```
import re
import conda.auxlib.packaging
print(conda.auxlib.packaging.GIT_DESCRIBE_REGEX.match('v1.0.0-52-gc50417f').groupdict())
```
produces `{'version': 'v1.0.0', 'post': '52', 'hash': 'c50417f'}` before the change and `{'version': '1.0.0', 'post': '52', 'hash': 'c50417f'}` after the change, correctly stripping the 'v' prefix.

### Checklist - did you ...

### DIFF
--- a/conda/auxlib/packaging.py
+++ b/conda/auxlib/packaging.py
@@ -82,7 +82,7 @@ import sys
 log = getLogger(__name__)
 
 Response = namedtuple('Response', ['stdout', 'stderr', 'rc'])
-GIT_DESCRIBE_REGEX = compile(r"(?:[_-a-zA-Z]*)"
+GIT_DESCRIBE_REGEX = compile(r"(?:[_\-a-zA-Z]*)"
                              r"(?P<version>[a-zA-Z0-9.]+)"
                              r"(?:-(?P<post>\d+)-g(?P<hash>[0-9a-f]{7,}))$")
 
--- a/news/11576-fix-get-describe-regex
+++ b/news/11576-fix-get-describe-regex
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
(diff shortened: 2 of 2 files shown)


ryanvolz:
I have recently e-signed the CLA, so checking an updating the bot should take care of that.

As far as I can tell, this doesn't have any tests or documentation currently, so I'm not sure what to do there.

travishathaway:
Hi @ryanvolz,

Thanks a lot for submitting this fix. If you'd like to create tests, I would create a new directory underneath `tests` called `auxlib` and create a new `test_packaging.py` file to hold the tests.

For the tests themselves, you will see that we have both older class-based `unittest` style tests and newer `pytest` style tests. We now prefer using `pytest` styled test. To get a good idea of what these look like, you can take a look at this recent pull request https://github.com/conda/conda/pull/11462 under the `tests` folder.

Once this pull request has some tests associated with, I will have a discussion with my fellow maintainers to see if this is a good candidate to be merged (seems pretty likely that it is though).

jezdez:
@conda-bot check

PR #11589, opened 2022-06-28
Add cytoolz as dependency

### LINKED ISSUE #11333: Replace `toolz` dependency with `cytoolz`
This makes me all kind of uncomfortable since cytoolz is a separate pacakge and us using it this way is totally backwards. We should rely on it instead directly or remove our footprint using it. Probably larger than this PR though, so worth opening a ticket to look into it.

_Originally posted by @jezdez in https://github.com/conda/conda/pull/11290#discussion_r826836808_

travishathaway:
### Description
This removes the `toolz` module from our vendored dependencies. Instead, we will rely explicitly on `cytoolz` from now on.

Resolves #11333 

### Checklist - did you ...


kenodegard (approved):

kenodegard (changes requested):
This warrants a news snippet.

kenodegard (approved):

kenodegard:
Just reviewed all of our other repos and found this impacts other repos. `conda._vendor.toolz` is used in both [`constructor`](https://github.com/conda/constructor/blob/7a6cf1d65035c41d5c8f4bb519a5eb8cff8ff9e4/constructor/conda_interface.py#L31-L33) and [`conda-libmamba-solver`](https://github.com/conda-incubator/conda-libmamba-solver/blob/640456f70731518f9d50e7c05b285b0a0e526d0c/conda_libmamba_solver/_libmamba.py#L18).

conda/activate.py
```diff
@@ -13,10 +13,11 @@
 import sys
 from textwrap import dedent
 
+from cytoolz.itertoolz import concatv, drop
```
jakirkham, line 16: There is a `tlz` namespace that tries `cytoolz` first and falls back to `toolz`. It also works regardless of which is installed. This is what we do in Dask ( https://github.com/dask/dask/pull/5949#issuecomment-595407707 )

```suggestion
from tlz.itertoolz import concatv, drop
```

kenodegard: hm interesting, and I see that `cytoolz` lists `toolz` as a dependency so it'll always be present

travishathaway: Thanks for the suggestion. I just updated all of our import statements to use `tlz` instead of `cytoolz` 👍

jakirkham: Thanks Travis! 🙏

kenodegard (approved):
🚀

kenodegard (approved):

PR #11597, opened 2022-06-30
Deprecate `conda.auxlib.decorators.memoize` in favor of `functools.lru_cache`

kenodegard:
### Description

Deprecating custom caching in favor of stdlib caching.

Notable observations:
1. `conda.auxlib.decorators.memoizedproperty` and `conda.auxlib.decorators.memoizemethod` implement a shared cache (shared within the object instance) which allows for the entire object's cache to be reset all at once (instead of individually per property/method as can be done with `lru_cache.cache_clear`). This makes it less straightforward to replace with `lru_cache` at this time.
2. `conda.exports.memoized` is yet another custom implementation of caching that is only still in use in `conda-build` and includes logic to skip caching when any of the arguments are unhashable. This will need to be addressed in `conda-build` separately.

### Checklist - did you ...

### DIFF
--- a/conda/auxlib/decorators.py
+++ b/conda/auxlib/decorators.py
@@ -50,9 +50,10 @@ def memoize(func):
     TypeError: Can't memoize a generator or non-hashable object!
     """
     warnings.warn(
-        "The `conda.auxlib.decorators.memoize` is pending deprecation and will be removed in a "
-        "future release. Please use `functools.lru_cache` instead.",
+        "The `conda.auxlib.decorators.memoize` decorator is pending deprecation and will be "
+        "removed in a future release. Please use `functools.lru_cache` instead.",
         PendingDeprecationWarning,
+        stacklevel=2,
     )
 
     func._result_cache = {}  # pylint: disable-msg=W0212
--- a/conda/auxlib/decorators.py
+++ b/conda/auxlib/decorators.py
@@ -4,6 +4,7 @@ try:
 except ImportError:
     from collections import Hashable
 from types import GeneratorType
+import warnings
 
 from .._vendor.six import wraps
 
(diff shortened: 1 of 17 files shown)


conda/exports.py
```diff
@@ -132,6 +135,12 @@ class memoized(object):  # pragma: no cover
     (not reevaluated).
     """
     def __init__(self, func):
+        warnings.warn(
+            "The `conda.exports.memoized` is pending deprecation and will be removed in a future "
+            "release. Please use `functools.lru_cache` instead.",
+            PendingDeprecationWarning,
```
jezdez, line 141: I'm wondering if it would be better to use a stacklevel of 2 here to make sure the deprecation warning shows up next to the code that uses the decorator than pointing to the implementation here.

```suggestion
            PendingDeprecationWarning, stacklevel=2
```

But OTOH I'm not sure how decorator initialization behaves in this case with the stack.

kenodegard: I don't think the `stacklevel` helps here. This is what I get without the `stacklevel` which correctly refers to `conda.exports` line 138:



And with `stacklevel` it refers to `stdin`:

jezdez: Ah, would you mind trying with a stacklevel of 1?

conda/auxlib/decorators.py
```diff
@@ -48,6 +49,12 @@ def memoize(func):
     Traceback (most recent call last):
     TypeError: Can't memoize a generator or non-hashable object!
     """
+    warnings.warn(
+        "The `conda.auxlib.decorators.memoize` is pending deprecation and will be removed in a "
+        "future release. Please use `functools.lru_cache` instead.",
```
jezdez, line 54:
```suggestion
        "The `conda.auxlib.decorators.memoize` decorator is pending deprecation and will be removed in a "
        "future release. Please use `functools.lru_cache` instead.",
```

jezdez (approved):

conda/exports.py
kenodegard: Wait, I think you're right and I'm wrong. We do want stacklevel 2 because we want the second case from above (not the first), we want the `<stdin>:2` file/line. We need users to see where _they_ have included `@memoized` not where `memoized` is defined in conda, right?

Issue #7617, opened 2018-08-03
conda config clear option


jakirkham:
Would be great to have an option for `conda config` to `clear` a list entry of all items. For example, `conda config --clear channels` would change `condarc` like so.

```yaml
...
channels: []
...
```

MesaJonathan:
I'd like to take a swing at this as my first open-source contribution

jakirkham:
Think you can just give it a try if you are interested. Thanks for looking into it Jonathan 😄

jakirkham:
(bump)

PR #11627, opened 2022-07-16
Add `conda config --clear` option

dharmaquark:
### Description

This PR introduces a new conda config clear option, https://github.com/conda/conda/issues/7617

#### Usage notes
This current approach will work for features that map to values that are lists, e.g.
`$ conda config --clear track_features` 

For keys that do not map to lists, running --clear will generate the following error:
```
$ conda config --clear quiet

CondaValueError: Key 'quiet' is not a known sequence parameter.
```

I'd be happy to add some tests for this, if this approach and scope looks correct (I'm not 100% sure if this will properly handle subkeys, or if it should? another open question is whether this should clear out keys that map to string values?)

### Checklist - did you ...


dharmaquark:
> We require contributors to sign our [Contributor License Agreement](https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement) and we don't have one on file for @dharmaquark.
> 
> In order for us to review and merge your code, please e-sign the [Contributor License Agreement PDF](https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement). We then need to manually verify your signature. We will ping the bot to refresh the PR status when we have confirmed your signature.

confirming, I have now signed the contributor license agreement

jezdez:
@conda-bot check

jezdez:
This needs news item before being merged.

conda/cli/main_config.py
```diff
@@ -365,6 +365,29 @@ def execute_config(args, parser):
                                 (item, key))
         rc_config[key] = [i for i in rc_config[key] if i != item]
 
+    # Clear
+    for key, in args.clear:
```
jaimergp, line 369:
```suggestion
    for key in args.clear:
```

I don't think we need that comma, do we?

conda/cli/main_config.py
```diff
@@ -365,6 +365,29 @@ def execute_config(args, parser):
...
+        key, subkey = key.split('.', 1) if '.' in key else (key, None)
```
jaimergp, line 370: Is this `key.subkey` notation supported in other commands too?

conda/cli/main_config.py
```diff
@@ -365,6 +365,29 @@ def execute_config(args, parser):
...
+        if key == 'channels' and key not in rc_config:
+            rc_config[key] = ['defaults']
```
jaimergp, line 372: This hardcoded value assumes the user started with `Miniconda`, which hardcodes `defaults`. However, other distributions like Miniforge or Mambaforge hardcode `conda-forge`. There might be other distros out there with different configurations too.

dharmaquark: ah I see- would it make sense to just remove this default logic, since this `clear` option would be emptying arrays versus returning to any default value?

conda/cli/main_config.py
dharmaquark: yup- this is supported for prepend, append, and add operations:
https://github.com/conda/conda/blob/448fc905536ffd5a39781aeb50076c7831e302dd/conda/cli/main_config.py#L311

does it make sense to also support for clear, for consistency? or would starting with only supporting primary keys cover most valid use cases of clearing out an config array?

conda/cli/main_config.py
dharmaquark: ah, without the `,` notation, this iterates over an array object, e.g. key == [keyvalue], the comma notation pulls the element out of the array. This approach mirrors the iteration logic that `remove key` uses:
https://github.com/conda/conda/blob/448fc905536ffd5a39781aeb50076c7831e302dd/conda/cli/main_config.py#L391

dharmaquark:
> This needs a news item before being merged.

should I add that as part of this PR changeset, or would that be something a conda admin would handle as part of release? Happy to help write that up if needed!

conda/cli/conda_argparse.py
```diff
@@ -531,6 +531,15 @@ def configure_parser_config(sub_parsers):
...
     )
+    config_modifiers.add_argument(
+        "--clear",
+        nargs=1,
+        action="append",
+        help="""Clear a list key. This clears out
+    all values for the key.""",
+        default=[],
+        metavar="KEY",
+    )
```
kenodegard, line 542: Let's tweak this slightly to avoid the [trailing comma that @jaimergp pointed out](https://github.com/conda/conda/pull/11627#discussion_r927447722).

```suggestion
    config_modifiers.add_argument(
        "--clear",
        action="append",
        help="Clear a list key. This clears out all values for the key.",
        metavar="KEY",
    )
```

kenodegard (changes requested):
@dharmaquark yes, please include the news snippet in this PR

conda/cli/main_config.py
kenodegard: Let's go ahead and drop this special conditional

jezdez:
@kenodegard Do you think we should move this to the next release milestone?

Issue #11517, opened 2022-06-02
Ensure conda uses inclusive language


travishathaway:
### Description

In order to create a welcoming environment for our community, we want to make sure this repository uses inclusive language and avoids problematic terms such as "master" for branch names or "whitelist"/"blacklist" for variable names.

### Linked Issues & PRs

- [x] https://github.com/conda/conda/pull/11570
- [x] Rename `master` branch to `main`
- [x] https://github.com/conda/conda/pull/11645
- [x] https://github.com/conda/conda/pull/11647
- [x] Rename any variables named `black*` to `deny*` (not present)
- [x] Fix ReadTheDocs config to make sure code is pulled from the correct branch (@jezdez)
- [x] Announcements need to be made in all public channels (conda slack team and conda-forge gitter) with instructions on how people can update their local development environments
    - @kenodegard: this may not be necessary since GitHub redirects/announces this for us, see https://github.com/github/renaming
    - [x] @jezdez done for the master/main change in conda/conda

travishathaway:
Here's a good how-to article describing how to rename the "master" branch to "main". It requires the appropriate rights in GitHub to complete though:

https://www.git-tower.com/learn/git/faq/git-rename-master-to-main/

travishathaway:
It's important to note that only org owners can make the necessary updates to this repository in order to complete this issue.

travishathaway:
It's important to note that only org owners can make the necessary updates to this repository in order to complete this issue.

travishathaway:
It's important to note that only org owners can make the necessary updates to this repository in order to complete this issue.

jezdez:
Hey @travishathaway, do you think we could make this ticket an epic and try to tackle each item listed above separately (since some are admin-style tasks and some are code changes needed)?

jezdez:
(delete duplicate comments as https://github.com/conda/conda/issues/11517#issuecomment-1153808221)

kenodegard:
Actually, the process of renaming master to main is even easier thanks to GitHub: https://github.com/github/renaming

It's just a matter of doing the rename in the GitHub web interface and the rest gets done for us.

PR #11645, opened 2022-07-21
Update gitpod/images to use main branch

kenodegard:
### Description

Missed a few references to the master branch. @jaimergp I'm no expert on GitPod so it would be great if you could take a look.

Ref: https://github.com/conda/conda/issues/11517

### Checklist - did you ...

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,5 +1,4 @@
-# TODO: Replace gitpod with master
-image: ghcr.io/conda/conda-ci:gitpod-linux-python3.9
+image: ghcr.io/conda/conda-ci:main-linux-python3.9
 tasks:
   - name: initialize
     init: |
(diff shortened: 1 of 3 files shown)


.gitpod.yml
```diff
@@ -19,7 +18,7 @@ vscode:
 
 github:
   prebuilds:
-    # enable for master branch
+    # enable for default branch
     master: true
```
jaimergp, line 22: Interestingly they still maintain `master` as the key here, but [the docs say](https://www.gitpod.io/docs/references/gitpod-yml#prebuildsmaster) it enables prebuilds for the "default" branch, so it looks like we are good?

.gitpod.yml
```diff
@@ -1,5 +1,4 @@
-# TODO: Replace gitpod with master
```
jaimergp, line 1: Good catch!

jaimergp (approved):

PR #11545, opened 2022-06-14
Change `conda-content-trust` from pre-solve to post-solve

kenodegard:
### Description

In an effort to speed up package signature verification we move the verification process to occur after solving. This way only the packages included in the transaction are validated (tens of packages, not thousands).

~A secondary goal was to move even more of the signature verification logic into `conda-content-trust` itself. This is in preparation for making this a conda plugin once the framework is available (#11435). With this change, conda knows/cares even less about what and how signature verification happens.~

Depends on #11646 
Resolves #11154
~See conda/conda-content-trust#29~

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/master/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #11646, opened 2022-07-21
Refactor content trust into submodule

kenodegard:
### Description

Splitting refactor from pre/post change in #11545 for easier review.

This is functionally the same as before.

### Checklist - did you ...


conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+                log.info(f"Loading root metadata from {path}.")
+                trusted = load_metadata_from_file(path)
+                break
+        else:
+            log.debug(
+                f"No root metadata in {context.av_data_dir}. " "Using built-in root metadata."
+            )
+
+        # Refresh trust root metadata
+        while True:
```
jezdez, line 104: Why does this need an endless loop?

conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # If key_mgr is unavailable from server, fall back to copy on disk
+        if not trusted and exists(path):
+            trusted = load_metadata_from_file(path)
+
+        return trusted
+
+    @staticmethod
+    def _fetch_channel_signing_data(signing_data_url, filename, etag=None, mod_stamp=None):
+        if not context.ssl_verify:
```
jezdez, line 170: This doesn't need to be a staticmethod and would be better suited to be a regular instance method to make use of standard OOP patterns.

conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+            warnings.simplefilter("ignore", InsecureRequestWarning)
+
+        session = CondaSession()
```
jezdez, line 173: Let's move this into a method, since this may become important in the future when we switch out the fetching backend.

conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # In certain cases (e.g., using `-c` access anaconda.org channels), the
+        # `CondaSession.get()` retry logic combined with the remote server's
+        # behavior can result in non-JSON content being returned.  Parse returned
+        # content here (rather than directly in the return statement) so callers of
+        # this function only have to worry about a ValueError being raised.
+        try:
+            str_data = json.loads(resp.content)
+        except json.decoder.JSONDecodeError as err:  # noqa
+            raise ValueError(f"Invalid JSON returned from {signing_data_url}/{filename}") from err
```
jezdez, line 213: Doesn't requests provide a JSON loading mechanism with advanced error handling?

conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+        # TODO: additional loading and error handling improvements?
+
+        return str_data
+
+    def __call__(self, info, fn, signatures):
+        if not self.enabled or fn not in signatures:
+            return
+
+        envelope = wrap_as_signable(info)
+        envelope["signatures"] = signatures[fn]
```
jezdez, line 224: Let's explain what we're doing here in a comment, please.

conda/content_trust/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+        try:
+            verify_delegation("pkg_mgr", envelope, self.key_mgr)
+        except SignatureError:
+            log.warn(f"invalid signature for {fn}")
+            status = "(WARNING: metadata signature verification failed)"
+        else:
+            status = "(INFO: package metadata is signed by Anaconda and trusted)"
```
jezdez, line 232: Do we know that to be true generally? I assumed we'd eventually want to delegate the signing to other parties like conda-forge, where this might be confusing users?

conda/content_trust/constants.py
```diff
@@ -0,0 +1,50 @@
...
+                "pubkeys": [
+                    "668a3217d72d4064edb16648435dc4a3e09a172ecee45dcab1464dcd2f402ec6",
+                    "508debb915ede0b16dc0cff63f250bde73c5923317b44719fcfc25cc95560c44",
+                    "6d4d5888398ad77465e9fd53996309187723e16509144aa6733015c960378e7a",
+                    "e0c88b4c0721bd451b7e720dfb0d0bb6b3853f0cbcf5570edd73367d0841be51",
+                ],
+                "threshold": 2,
+            },
+        },
+        "expiration": "2022-10-31T18:00:00Z",
```
jezdez, line 42: That's not far away, do we plan to update this in the September release?

conda/content_trust/__init__.py
kenodegard: Because we don't know how many signatures are in the chain of trust? We iteratively keep fetching a newer signature until there are no more to fetch. This could be switched to using a boolean conditional instead of `while True`/`break`.

conda/content_trust/__init__.py
kenodegard: You mean move the session definition itself? E.g.

```python
class _SignatureVerification:
    @property
    @lru_cache(maxsize=None)
    def session(self):
        return CondaSession()

    ...
```

conda/content_trust/__init__.py
kenodegard: I opted for a `staticmethod` to clearly indicate that this is truly a helper function and doesn't depend on the current instance of `self`.

What standard OOP patterns are you referring to?

conda/content_trust/__init__.py
kenodegard: You mean `resp.json()`? That's probably a good idea, can still catch the resulting `JSONDecodeError` and use a conda exception.

conda/content_trust/__init__.py
kenodegard: Already had this conversation with @awwad, we agreed this info needs to be extracted from `repodata.json` but that change is out of scope for this work.

conda/content_trust/__init__.py
jezdez: Yeah, let's do the boolean, I'm usually a little worried to see `while True` in parts of code that do network I/O.

conda/content_trust/__init__.py
jezdez: Standard OOP patterns like subclassing, for example. If we were to decide in the future to make fetching the signing data pluggable, which would require refactoring this part of the functionality, we'd have to change the signature of this method.

A good example that I think showcases this already is the use of the `CondaSession` in the method, which could easily be decomposed into an own method to separate data fetching from data processing and making this class a little more future-proof. Your intentions regarding emphasizing, that this is a helper method would still work (that's a good pattern), by using the leading underscore in the method name.

conda/content_trust/__init__.py
jezdez: Yep!

conda/content_trust/__init__.py
jezdez: Yeah, that!

conda/content_trust/__init__.py
jezdez: Understood, thank you!

conda/content_trust/__init__.py
kenodegard: You're right, I didn't realize the session idea was in this method

conda/content_trust/__init__.py
kenodegard: We discussed this offline and agree that ideally, the signatures would include an additional key, `has_next`, to indicate that there are more signatures to fetch in the chain of trust.

kenodegard:
@jezdez re. renaming the submodule, I think `verification` is equally confusing given our `conda-verify` package, any other ideas?

jezdez:
@kenodegard Hmm, that's tricky, maybe we shouldn't have an own submodule then and move the `_SignatureVerification` class as `SignatureVerification` to `conda/models/verification.py` and leave the constants in conda.base.constants?

conda/trust/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+    from conda_content_trust.signing import wrap_as_signable
+except ImportError:
+    # SignatureStatus.enabled handles this
+    pass
+
+
+log = getLogger(__name__)
+
+
+class _SignatureVerification:
```
jezdez, line 35: Since I usually worry about import time side-effects, could we move this into conda.trust.verification or something so when we add additional submodules to conda.trust we're not triggering the imports for conda-content-trust?

jezdez (dismissed):

jezdez (approved):

awwad (approved):

PR #11647, opened 2022-07-21
Rename `whitelist_channels` to `allowlist_channels`

kenodegard:
### Description

Part of inclusive language efforts. This is a low-impact change as no other conda org project refers to `whitelist_channels`.

Ref: https://github.com/conda/conda/issues/11517

> **Warning**
> pre-commit/darker/black will not be satisfied until this has been reviewed. Prior to merge we will apply the necessary changes.

### Checklist - did you ...

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -9,6 +9,7 @@ from itertools import chain
 from logging import getLogger
 import platform
 import sys
+import warnings
 
 from .package_cache_data import PackageCacheData
 from .prefix_data import PrefixData
@@ -26,6 +27,14 @@ from ..models.records import EMPTY_LINK, PackageCacheRecord, PackageRecord, Pref
 
 log = getLogger(__name__)
 
+
+def check_whitelist(channel_urls):
+    warnings.warn(
+        "`conda.core.index.check_whitelist` is pending deprecation and will be removed in a "
+        "future release. Please use `conda.core.index.check_allowlist` instead.",
+        PendingDeprecationWarning,
+    )
+    return check_whitelist(channel_urls)
 
 def check_allowlist(channel_urls):
     if context.allowlist_channels:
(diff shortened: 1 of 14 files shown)


conda/core/index.py
```diff
@@ -27,14 +27,14 @@
 log = getLogger(__name__)
 
 
-def check_whitelist(channel_urls):
-    if context.whitelist_channels:
-        whitelist_channel_urls = tuple(concat(
-            Channel(c).base_urls for c in context.whitelist_channels
+def check_allowlist(channel_urls):
```
jaimergp, line 30: I think mamba imports this function in their utils.py module:

https://github.com/mamba-org/mamba/blob/da44aadfaf88dfa3ce21656d671682626379cea6/mamba/mamba/utils.py#L71

We can add an alias for one release, with the corresponding deprecation warning?

kenodegard: Oh good point

kenodegard: Added

news/11647-allowlist_channels
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Rename `whitelist_channels` to `allowlist_channels`. (#11647)
```
beeankha, line 3: I might reword this to say that we're changing any instances of `whitelist` to `allowlist` and `blacklist` to `denylist`

beeankha (approved):
Left a non-blocking suggestion for the news file; otherwise LGTM!

news/11647-allowlist_channels
kenodegard: Hehe, I explicitly wrote this with the opposite mindset; outline the actual change, not the policy/objective that resulted in the change. I wrote it this way since the change only resulted in one meaningful variable change. 🤷🏻 

I have no strong opinion here so if others agree that this should be reworded I'll make the change.

jezdez: I think this is fine as written, it's not worth adding troll traps into the changelog. The issues and pull requests have more details than whatever we reasonably would have to write in the changelog to explain the policy change about naming things.

jezdez (approved):
🚀

Issue #11087, opened 2021-12-06
Document using `conda init` for TCL/LMOD


kenodegard:
Include documentation to outline setups with TCL/LMOD. We should also create destructor scripts for these kinds of use cases to help cleanup user environments when modules are unset.

See #11067

### TCL
```
set name    python
set version 3.9-anaconda-2021.11
set root    /usr/common/software/$name/$version

if { [ module-info shell ] == "bash" } {
    puts stdout "source $root/etc/profile.d/conda.sh ;"
}

if { [ module-info shell ] == "csh" } {
    puts stdout "source $root/etc/profile.d/conda.csh ;"
}
```

### LMOD
```
local root        = pathJoin("/global/common/software/nersc/pm-2021q4/sw", name, version)
execute{cmd="source '"..pathJoin(root, "/etc/profile.d/conda."..myShellType()).."'", modeA={"load"}}
```

jdblischak:
I confirmed that the above LMOD snippet is able to initialize conda when the module is loaded. However, it's unclear to me how to proceed to 1) auto-activate the base environment, and 2) remove conda from the PATH when the module is unloaded.

## Auto activate the base environment

When I use a personal installation of conda, I put the initialization code into my `.bashrc`, and it automatically activates the base environment when I login. From reading the code, it appears to me to be essentially running `etc/profile.d/conda.sh`, which is the same as the LMOD script is doing.

I confirmed that `auto_activate_base` is `True` for the conda installed as a module:

```
$ conda config --system --show auto_activate_base
auto_activate_base: True
```

## How to unload the module

After running `module unload`, the module is removed, but `condabin` is still in `PATH` and the conda env vars are still defined.

```sh
$ env | grep CONDA | cut -d'=' -f1
CONDA_SHLVL
CONDA_EXE
_CE_CONDA
CONDA_PYTHON_EXE
```

How can I de-initialize conda when the module is unloaded?

jdblischak:
I made some progress on unloading the module. I adapted the example from this [Lmod mailing list thread](https://sourceforge.net/p/lmod/mailman/message/36724692/) to unset the environment variables and remove `condabin/` from `PATH`:

```lua
cmd = "conda deactivate; " ..
      "unset ANACONDA3ROOT; unset PYTHONROOT; unset CONDA_EXE; " ..
      "unset CONDA_PYTHON_EXE; unset _CE_CONDA; unset _CE_M; " ..
[...]

PR #11649, opened 2022-07-22
Use `**kwargs` instead of `OrderedDict` for `_Activator.get_export_unset_vars`

kenodegard:
### Description

Since conda has [dropped Python 3.6 support](https://github.com/conda/conda/releases/tag/4.13.0), we can finally switch to using `**kwargs` instead of using `OrderedDict`.

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -76,21 +76,20 @@ class _Activator(object):
         unset_vars = []
         export_vars = {}
 
-        def split_export_unset(func=None, **kwargs):
-            for name, value in kwargs.items():
-                if value is None:
-                    unset_vars.append(name.upper())
-                elif func and value:
-                    export_vars[name.upper()] = func(value)
-                else:
-                    export_vars[name.upper()] = value
-
         # split provided environment variables into exports vs unsets
-        split_export_unset(**kwargs)
+        for name, value in kwargs.items():
+            if value is None:
+                unset_vars.append(name.upper())
+            else:
+                export_vars[name.upper()] = value
 
         if export_metavars:
             # split meta variables into exports vs unsets
-            split_export_unset(func=self.path_conversion, **context.conda_exe_vars_dict)
+            for name, value in context.conda_exe_vars_dict.items():
+                if value is None:
+                    unset_vars.append(name.upper())
+                else:
+                    export_vars[name.upper()] = self.path_conversion(value) if value else value
         else:
             # unset all meta variables
             unset_vars.extend(context.conda_exe_vars_dict)
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -65,8 +65,7 @@ class _Activator(object):
         self._raw_arguments = arguments
         self.environ = os.environ.copy()
 
-    # Once Python2 dies odargs can become kwargs again since dicts are ordered since 3.6.
-    def get_export_unset_vars(self, odargs):
+    def get_export_unset_vars(self, **kwargs):
         """
         :param kwargs: environment variables to export. The `conda_exe_vars` meta
                        variables are also exported by default. If you do not want
@@ -78,7 +77,6 @@ class _Activator(object):
         :return: A OrderedDict of env vars to export ordered the same way as kwargs.
                  And a list of env vars to unset.
         """
-        kwargs = odargs
         conda_exe_vars_None = False if ('conda_exe_vars' not in kwargs or
                                         kwargs['conda_exe_vars'] is not None) else True
         conda_exe_unset_vars = []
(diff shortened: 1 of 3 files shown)


conda/activate.py
```diff
@@ -65,8 +63,7 @@ def __init__(self, arguments=None):
         self._raw_arguments = arguments
         self.environ = os.environ.copy()
 
-    # Once Python2 dies odargs can become kwargs again since dicts are ordered since 3.6.
-    def get_export_unset_vars(self, odargs):
+    def get_export_unset_vars(self, *, conda_exe_vars=True, **kwargs):
```
travishathaway, line 66: Maybe it would make more sense to have `conda_exe_vars` be `export_conda_vars` instead? I say this because associating a boolean with a verb (or action in this case) is perhaps a little clearer. When I see a plural noun like `conda_exe_vars`, I instantly think, "oh this is going to be a sequence or something".

Just a suggestion, no need to implement

conda/activate.py
```diff
@@ -75,48 +72,45 @@ def get_export_unset_vars(self, odargs):
...
-            elif v is None:
-                unset_vars.append(k.upper())
-            else:
-                export_vars[k.upper()] = v
-        # Just to make writing tests a bit more ergonomic we add these to the end.
-        unset_vars += conda_exe_unset_vars
-        export_vars = OrderedDict(chain(export_vars.items(), conda_exe_vars_export.items()))
+        export_vars = {}
+
+        def split_export_unset(func=None, **kwargs):
```
travishathaway, line 81: Perhaps it would be better not to shadow the already existing `kwargs` variable. I usually go with `**kwargs_` (add a tail) or `**inner_kwargs` to avoid this.

conda/activate.py
```diff
@@ -75,48 +72,45 @@ def get_export_unset_vars(self, odargs):
...
+                    unset_vars.append(name.upper())
+                elif func and value:
+                    export_vars[name.upper()] = func(value)
+                else:
+                    export_vars[name.upper()] = value
+
+        # split provided environment variables into exports vs unsets
+        split_export_unset(**kwargs)
+
+        if conda_exe_vars is None:
```
travishathaway, line 93: I wonder if this is even necessary to have after having read this comment:

https://github.com/conda/conda/blob/6efd518f7b88d5682aa42b8b178d25a9e48bd1b2/conda/activate.py#L418-L425

It surely isn't being hit in our tests, so maybe if we do keep it, we could at least cover this line? Definitely a difficult decision to make when dealing with legacy code. I would err on the side of caution and keep it, but I would also add a new test case that at least reach this line, so that we know it works if used elsewhere.

travishathaway (approved):
I think this pull request is good as-is, but there are a couple things I commented on that I would like you to consider (namely, the line that isn't being covered by our tests at the moment). I'd also like to thank you for removing some very confusing import statements! Death to OrderDict! 🤣

conda/activate.py
kenodegard: Hm, I hadn't fully processed this comment earlier.

This is related to the de-initialization discussions in https://github.com/conda/conda/issues/11087.

I'm inclined to "fix" the deactivate command in the future to unset all conda envvars after deactivating the last env (the only thing to remain would be `condabin` on the `PATH` which we could argue is also unnecessary) so this feature may become useful again in the future. Let's leave it for now.

**But** since it's an unused feature at the moment I have no problems (1) making it a positional argument (to better distinguish it from the other keyword args) and (2) renaming it as suggested above.

conda/activate.py
kenodegard: Hm, personally I've never been a fan of using alternatives for `args` or `kwargs` unless I absolutely must (e.g., for decorators).

I'm also not a fan of the proposed implementation, what I really want to do was:

```python
def get_export_unset_vars(self, **kwargs):
    unset_vars = []
    export_vars = {}

    for name, value in {**kwargs, **context.conda_exe_vars_dict}.items():
        if value is None:
            unset_vars.append(name.upper())
        else:
            export_vars[name.upper()] = value
```

But things got weird with the `conda_exe_vars` conditional and the `path_conversion` only being applied to the `context.conda_exe_vars_dict` (and not to `kwargs`).

I'll simmer on this a little longer. 🤔

kenodegard: I decided to drop the helper function since it didn't help with the overall code readability

jezdez (approved):
Just needs a news item, even if it's an internal refactoring, people may be relying on internal state being stored as an OrderedDict somewhere out there.

Issue #7335, opened 2018-05-23
RFE: Support FIPS mode - do not use md5 hash


opoplawski:
**I'm submitting a...**
  - [ ] bug report
  - [X] feature request

md5 hashes are unsafe and are disabled in FIPS mode.

See also #5282 

### Steps to Reproduce
Boot with fips=1 
```
# /usr/bin/conda create -y -n scipy35-anaconda -c anaconda  python=3.5 astropy basemap h5py jupyter matplotlib mkl mysql-connector-python netcdf4 notebook numba pandas pip scikit-learn scipy seaborn sunpy sympy xarray nbbrowserpdf tensorflow                            
Solving environment: failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/usr/lib/python3.4/site-packages/conda/exceptions.py", line 819, in __call__
        return func(*args, **kwargs)
      File "/usr/lib/python3.4/site-packages/conda/cli/main.py", line 78, in _main
        exit_code = do_call(args, p)
      File "/usr/lib/python3.4/site-packages/conda/cli/conda_argparse.py", line 77, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/usr/lib/python3.4/site-packages/conda/cli/main_create.py", line 11, in execute
        install(args, parser, 'create')
      File "/usr/lib/python3.4/site-packages/conda/cli/install.py", line 235, in install
        force_reinstall=context.force,
      File "/usr/lib/python3.4/site-packages/conda/core/solve.py", line 505, in solve_for_transaction
        force_remove, force_reinstall)
      File "/usr/lib/python3.4/site-packages/conda/core/solve.py", line 438, in solve_for_diff
        final_precs = self.solve_final_state(deps_modifier, prune, ignore_pinned, force_remove)
      File "/usr/lib/python3.4/site-packages/conda/core/solve.py", line 179, in solve_final_state
        index, r = self._prepare(prepared_specs)
      File "/usr/lib/python3.4/site-packages/conda/core/solve.py", line 560, in _prepare
        self.subdirs, prepared_specs)
      File "/usr/lib/python3.4/site-packages/conda/core/index.py", line 174, in get_reduced_index
        subdir_datas = tuple(SubdirData(Channel(url)) for url in channel_urls)
      File "/usr/lib/python3.4/site-packages/conda/core/index.py", line 174, in <genexpr>
[...]

PR #11658, opened 2022-07-26
add usedforsecurity=False for hashlib.md5 for FIPS compliance

topherocity:
### Description

This pull request is a fix for a problem highlighted in #7335.

Conda fails to install packages on FIPS-enabled systems because use of the MD5 algorithm is disallowed for security purposes. Since conda is not using it for such a purpose, this pull request makes use of the `usedforsecurity=False` flag in `hashlib.md5` (available since Python 3.9 in CPython, and earlier on RHEL/CentOS systems).

### Checklist - did you ...

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -905,8 +905,11 @@ def cache_fn_url(url, repodata_fn=REPODATA_FN):
     #    are looking for the cache under keys without this.
     if repodata_fn != REPODATA_FN:
         url += repodata_fn
-    md5 = hashlib.md5(ensure_binary(url)).hexdigest()
-    return '%s.json' % (md5[:8],)
+    try:
+        md5 = hashlib.md5(ensure_binary(url))
+    except ValueError:
+        md5 = hashlib.md5(ensure_binary(url), usedforsecurity=False)
+    return '%s.json' % (md5.hexdigest()[:8],)
 
 
 def add_http_value_to_dict(resp, http_key, d, dict_key):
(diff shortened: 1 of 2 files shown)


topherocity:
CLA signed.

conda/core/subdir_data.py
```diff
@@ -905,8 +905,11 @@ def cache_fn_url(url, repodata_fn=REPODATA_FN):
     #    are looking for the cache under keys without this.
     if repodata_fn != REPODATA_FN:
         url += repodata_fn
-    md5 = hashlib.md5(ensure_binary(url)).hexdigest()
-    return '%s.json' % (md5[:8],)
+    try:
+        md5 = hashlib.md5(ensure_binary(url))
+    except ValueError:
+        md5 = hashlib.md5(ensure_binary(url), usedforsecurity=False)
```
chenghlee, line 911: Since we don't use MD5 for security purposes anyways, I think we should just set `usedforsecurity=False` in all cases and skip the `try`/`except` block. (The code for metadata signature verification and associated package checksumming use SHA-256.)

chenghlee (changes requested):

topherocity:
Added TODO as requested. 

Closes #7335.

chenghlee (approved):

jezdez:
@conda-bot check

jezdez (approved):

rkperkins:
The following was brought to my attention - “The solution does not cover every place that hashlib is used to create an algorithm, only the one that explicitly uses the md5 constructor and not `hashlib.new`. If you check the GitHub link, it shows two other locations that would need to be changed, assuming they are, in fact, not using it for security”.

PR #11659, opened 2022-07-26
Forward slashes in channel names breaks channel notice cache

### LINKED ISSUE #11657: Forward slashes in channel names breaks channel notice cache
### Checklist

### What happened?

When a channel name has forward slashes, this creates an illegal file name for our cache entries and causes errors like the following:

https://github.com/napari/packaging/runs/7519992448?check_suite_focus=true

We need to make sure that we are properly sanitizing channel names so that they contain no illegal characters. This needs to be true for "unix" like operating systems as well as Windows. This Stack Overflow post is a good starting point for listing these characters: https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

First reported by @jaimergp (thank you!)

travishathaway:
Fixes #11657.

With this PR, I decided to be a bit more defensive and not only strip forward slashes but other characters that are known to cause problems on Windows too (I used this Stack Overflow article as a reference: https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names)

This may or may not be a good idea, but is fairly trivial to change (just need to update the constant in `conda.base.constants`.

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -140,12 +140,6 @@ else:
         'powershell',
     )
 
-# These characters are not allowed in filenames for these types of operating systems
-ILLEGAL_FILENAME_CHARS_NIX = "/"
-
-ILLEGAL_FILENAME_CHARS_WIN = rf"\\|#*?><:\"{ILLEGAL_FILENAME_CHARS_NIX}"
-
-
 # Maximum priority, reserved for packages we really want to remove
 MAX_CHANNEL_PRIORITY = 10000
 
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -140,6 +140,7 @@ else:
         'powershell',
     )
 
+
 # Maximum priority, reserved for packages we really want to remove
 MAX_CHANNEL_PRIORITY = 10000
 
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -140,6 +140,11 @@ else:
         'powershell',
     )
 
+# These characters are not allowed in filenames for these types of operating systems
+ILLEGAL_FILENAME_CHARS_NIX = "/"
+
+ILLEGAL_FILENAME_CHARS_WIN = rf"\\|#*?><:\"{ILLEGAL_FILENAME_CHARS_NIX}"
+
 
 # Maximum priority, reserved for packages we really want to remove
 MAX_CHANNEL_PRIORITY = 10000
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -3,12 +3,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from datetime import datetime
+import hashlib
 from pathlib import Path
 from urllib import parse
 from typing import NamedTuple, Optional, Sequence
 
 from ..base.constants import NoticeLevel
-from ..utils import sanitize_filename
 
 
 class ChannelNotice(NamedTuple):
@@ -77,9 +77,16 @@ class ChannelNoticeResponse(NamedTuple):
 
     @classmethod
     def get_cache_key(cls, url: str, name: str, cache_dir: Path) -> Path:
-        """Returns the place where this channel response will be stored as cache"""
+        """
+        Returns the place where this channel response will be stored as cache
+
+        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
+        problematic characters). We don't care the that this is a weak hash.
+        """
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
-        cache_filename = sanitize_filename(f"{name}{path}.json")
+        bytes_filename = bytes(f"{name}{path}", "utf-8")
+        md5_hash = hashlib.md5(bytes_filename)
+        cache_filename = f"{md5_hash.hexdigest()}.json"
 
         return cache_dir.joinpath(cache_filename)
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -8,6 +8,7 @@ from urllib import parse
 from typing import NamedTuple, Optional, Sequence
 
 from ..base.constants import NoticeLevel
+from ..utils import sanitize_filename
 
 
 class ChannelNotice(NamedTuple):
@@ -79,5 +80,6 @@ class ChannelNoticeResponse(NamedTuple):
         """Returns the place where this channel response will be stored as cache"""
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+        cache_filename = sanitize_filename(f"{name}{path}")
+
         return cache_dir.joinpath(cache_filename)
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -5,7 +5,6 @@
 from datetime import datetime
 import hashlib
 from pathlib import Path
-from urllib import parse
 from typing import NamedTuple, Optional, Sequence
 
 from ..base.constants import NoticeLevel
@@ -76,15 +75,11 @@ class ChannelNoticeResponse(NamedTuple):
             return None
 
     @classmethod
-    def get_cache_key(cls, url: str, name: str, cache_dir: Path) -> Path:
+    def get_cache_key(cls, url: str, cache_dir: Path) -> Path:
         """
-        Returns the place where this channel response will be stored as cache
-
-        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
-        problematic characters). We don't care the that this is a weak hash.
+        Returns the place where this channel response will be stored as cache by hashing the url.
         """
-        url_obj = parse.urlparse(url)
-        bytes_filename = f"{name}{url_obj.path}".encode()
+        bytes_filename = url.encode()
         md5_hash = hashlib.sha256(bytes_filename)
         cache_filename = f"{md5_hash.hexdigest()}.json"
 
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -84,9 +84,8 @@ class ChannelNoticeResponse(NamedTuple):
         problematic characters). We don't care the that this is a weak hash.
         """
         url_obj = parse.urlparse(url)
-        path = url_obj.path.replace("/", "-")
-        bytes_filename = bytes(f"{name}{path}", "utf-8")
-        md5_hash = hashlib.md5(bytes_filename)
+        bytes_filename = f"{name}{url_obj.path}".encode()
+        md5_hash = hashlib.sha256(bytes_filename)
         cache_filename = f"{md5_hash.hexdigest()}.json"
 
         return cache_dir.joinpath(cache_filename)
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -80,6 +80,6 @@ class ChannelNoticeResponse(NamedTuple):
         """Returns the place where this channel response will be stored as cache"""
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
-        cache_filename = sanitize_filename(f"{name}{path}")
+        cache_filename = sanitize_filename(f"{name}{path}.json")
 
         return cache_dir.joinpath(cache_filename)
--- a/conda/notices/types.py
+++ b/conda/notices/types.py
@@ -80,7 +80,7 @@ class ChannelNoticeResponse(NamedTuple):
         Returns the place where this channel response will be stored as cache by hashing the url.
         """
         bytes_filename = url.encode()
-        md5_hash = hashlib.sha256(bytes_filename)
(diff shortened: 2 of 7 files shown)


conda/base/constants.py
```diff
@@ -140,6 +140,11 @@
         'powershell',
     )
 
+# These characters are not allowed in filenames for these types of operating systems
+ILLEGAL_FILENAME_CHARS_NIX = "/"
+
+ILLEGAL_FILENAME_CHARS_WIN = rf"\\|#*?><:\"{ILLEGAL_FILENAME_CHARS_NIX}"
```
kenodegard, line 146:
```suggestion
ILLEGAL_FILENAME_CHARS_WIN = rf'\\|#*?><:"{ILLEGAL_FILENAME_CHARS_NIX}'
```

conda/notices/types.py
```diff
@@ -79,5 +80,6 @@ def get_cache_key(cls, url: str, name: str, cache_dir: Path) -> Path:
         """Returns the place where this channel response will be stored as cache"""
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+        cache_filename = sanitize_filename(f"{name}{path}")
```
kenodegard, line 83: What happens if there are collisions? Two different channels with the same package but after sanitization they have the same `cache_filename`?

Is there value in using a hashing algorithm instead of sanitizing?

kenodegard (changes requested):

conda/notices/types.py
travishathaway: Yeah, I thought of this scenario as well. A hashing algorithm is good idea. This should actually be a pretty trivial change to make.

travishathaway: It would only make debugging more difficult later... but I don't think that's a deal breaker.

conda/utils.py
```diff
@@ -534,3 +535,13 @@ def safe_open(*args, **kwargs):
...
+
+
+def sanitize_filename(filename: str, replacement_char: str = "_") -> str:
+    """
+    Strips all illegal/problematic characters from a filename and replaces them
+    with `replacement_char` (defaults to "_"). We apply Windows rules
+    (see ``conda.base.constants.ILLEGAL_FILENAME_CHARS_WIN``) to err on the
+    safe side.
+    """
+    return re.sub(rf"[{ILLEGAL_FILENAME_CHARS_WIN}]", replacement_char, filename)
```
jezdez, line 547: (this is probably moot if we use a hash instead)

It's a good practice to compile regex patterns ahead of running this function so it only happens once per thread/process and not on each execution of `re.sub`.

Since the code defined in `conda/base/` is supposed to be very fast to import (according to https://github.com/conda/conda/blob/8c835dd3b54b15ed68ef1a58ba8862a0e1ddcbae/conda/base/README.md), where compiling a regex pattern may not fit the bill, I'd suggest to define this on top of the `conda/utils.py` file:

```python
ILLEGAL_FILENAME_CHARS_WIN_RE = re.compile(rf"[{ILLEGAL_FILENAME_CHARS_WIN}]")
```

And then use it below:

```suggestion
def sanitize_filename(filename: str, replacement_char: str = "_") -> str:
    """
    Strips all illegal/problematic characters from a filename and replaces them
    with `replacement_char` (defaults to "_"). We apply Windows rules
    (see ``conda.base.constants.ILLEGAL_FILENAME_CHARS_WIN``) to err on the
    safe side.
    """
    return ILLEGAL_FILENAME_CHARS_WIN_RE.sub(replacement_char, filename)
```

conda/notices/types.py
jezdez: This reminds me of the short discussion we had about using [requests-cache](https://github.com/conda/conda/pull/11462#discussion_r902484853) to reduce the amount of effort to maintain the notices cache system.

I'd suggest to create a SHA256 hash with the provided URL of the channel and not look back, it probably is enough®.

jezdez (changes requested):
Agree with Ken that this should be made simpler by using a hash per channel URL instead of adding complexity by introducing additional platform specific information.

conda/notices/types.py
travishathaway: I re-implemented this with an md5. This is not a problem right? We don't care about this stuff being visible.

travishathaway:
Great suggestion @kenodegard! This PR looks way better than the first version I came up with. I left the regression test more or less as-is though. Perhaps a little overkill, but it can't hurt.

conda/notices/types.py
travishathaway: ☝️ trivial change, so no worries if it's not okay. I wouldn't think that they would ever deprecate/remove this, right?

conda/notices/types.py
```diff
@@ -76,8 +77,16 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
...
+        Returns the place where this channel response will be stored as cache
+
+        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
+        problematic characters). We don't care the that this is a weak hash.
+        """
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+        bytes_filename = bytes(f"{name}{path}", "utf-8")
+        md5_hash = hashlib.md5(bytes_filename)
```
kenodegard, line 89: Unfortunately, we cannot do [binary f-strings](https://peps.python.org/pep-0498/#no-binary-f-strings) but we can use `str.encode` (where the default encoding is utf-8):

```suggestion
        md5_hash = hashlib.md5(f"{name}{path}".encode())
```

kenodegard (approved):
Yup, this definitely looks better!

conda/notices/types.py
jezdez: Is there a reason why this is using md5 and not sha256?

conda/notices/types.py
jezdez: Since we're using SHA256 for many other parts of conda, I'd suggest to just use that instead of MD5. It probably is computationally not more expensive for these short strings than MD5 and consistency works in our favor IMO.

conda/notices/types.py
```diff
@@ -76,8 +77,16 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
...
     def get_cache_key(cls, url: str, name: str, cache_dir: Path) -> Path:
-        """Returns the place where this channel response will be stored as cache"""
+        """
+        Returns the place where this channel response will be stored as cache
+
+        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
+        problematic characters). We don't care the that this is a weak hash.
+        """
         url_obj = parse.urlparse(url)
         path = url_obj.path.replace("/", "-")
```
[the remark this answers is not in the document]
jezdez: I don't think we need to do any of that, hashing the URL should suffice since that'll be stable enough to make the cache work.

conda/notices/types.py
travishathaway: No, reason. Just an arbitrary choice 😅 . Like I said, super trivial to change.

conda/notices/types.py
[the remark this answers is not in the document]
travishathaway: Good catch. Will fix.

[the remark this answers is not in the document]
kenodegard: @jezdez do you mean that you think we should skip the `parse.urlparse` step and just hash the plain URL?

conda/notices/types.py
```diff
@@ -76,8 +77,15 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
...
+        Returns the place where this channel response will be stored as cache
+
+        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
+        problematic characters). We don't care the that this is a weak hash.
+        """
         url_obj = parse.urlparse(url)
-        path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+        bytes_filename = f"{name}{url_obj.path}".encode()
+        md5_hash = hashlib.sha256(bytes_filename)
```
dholth, line 88: change the variable name to `hash`. it would also be reasonable to truncate the hash.

conda/notices/types.py
```diff
@@ -76,8 +77,15 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
...
+        """
+        Returns the place where this channel response will be stored as cache
+
+        We use a naive hashing algorithm to avoid issues with filenames (e.g. illegal or
+        problematic characters). We don't care the that this is a weak hash.
+        """
         url_obj = parse.urlparse(url)
-        path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+        bytes_filename = f"{name}{url_obj.path}".encode()
```
dholth, line 87: bold move calling `.encode()` with no arguments, usually this is spelled `.encode('utf-8')`

conda/notices/types.py
```diff
@@ -76,8 +77,15 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
 
     @classmethod
     def get_cache_key(cls, url: str, name: str, cache_dir: Path) -> Path:
-        """Returns the place where this channel response will be stored as cache"""
+        """
+        Returns the place where this channel response will be stored as cache
```
dholth, line 81: Not following the "Return the place" instead of "Returns the place" grammar?

conda/notices/types.py
kenodegard: Meh, I dislike reiterating the [default value](https://docs.python.org/3/library/stdtypes.html#str.encode) 🤷🏻‍♂️

conda/notices/types.py
kenodegard: won't truncating the hash lead to possible collisions?

conda/notices/types.py
dholth: Cool, as long as it's not `sys.getdefaultencoding()`, I see that the default is always "utf-8" at least since Python 3.5+

conda/notices/types.py
dholth: There's a nice approximation at https://en.wikipedia.org/wiki/Birthday_attack. 64 bits or 8 bytes (half the length of md5) gives about a one in a million chance of collision, if you've hashed 6.1 million items.

conda/notices/types.py
dholth: I'm curious about the channel name / path instead of the full URL. Conda usually expects channel names to be unique right?

conda/notices/types.py
dholth: I'm curious about the channel name / path instead of the full URL. Conda usually expects channel names to be unique right?

conda/notices/types.py
dholth: (A pep 257 style guide https://peps.python.org/pep-0257/#one-line-docstrings)

travishathaway: The sentence is written with as if it begins with "This function..." So, fully spelled out:

> This function returns the place where this channel response will be stored as cache

I've always been pretty liberal with my usage of sentence fragments in doc strings, but would be open to changing that habit if it makes it more clear.

conda/notices/types.py
travishathaway: Thanks @dholth, this function has been slowly mangled by layers of comments, and I forgot the that U in URL stands for "Unique" 🙈 . I'm going to update this to be more clear and just to hash the URL string instead. No channel name needed!

conda/notices/types.py
dholth: Not a big deal

conda/notices/types.py
```diff
@@ -75,9 +75,12 @@ def _parse_iso_timestamp(iso_timestamp: Optional[str]) -> Optional[datetime]:
...
-        url_obj = parse.urlparse(url)
-        path = url_obj.path.replace("/", "-")
-        cache_filename = f"{name}{path}"
+    def get_cache_key(cls, url: str, cache_dir: Path) -> Path:
+        """
+        Returns the place where this channel response will be stored as cache by hashing the url.
+        """
+        bytes_filename = url.encode()
+        md5_hash = hashlib.sha256(bytes_filename)
+        cache_filename = f"{md5_hash.hexdigest()}.json"
```
jezdez, line 84:
```suggestion
        sha256_hash = hashlib.sha256(bytes_filename)
        cache_filename = f"{sha256_hash.hexdigest()}.json"
```

jezdez (approved):
Nice work, thank you!

conda/notices/types.py
travishathaway: oops 🙃 . Good catch (would have been confusing otherwise)

PR #11626, opened 2022-07-15
Support CUDA detection on WSL

jaimergp:
### Description

Support CUDA detection on WSL. We are just adding a known location for `libcuda.so`, which corresponds to the WSL path (WSL is detected as Linux).

Closes https://github.com/conda/conda/issues/11615
Closes https://github.com/conda/conda/issues/11488

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/master/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #11667, opened 2022-07-28
Detect CUDA driver version in subprocess

XuehaiPan:
### Description

Followed by #11626, see https://github.com/conda/conda/pull/11626#issuecomment-1186553105.

The environment variable `CUDA_VISIBLE_DEVICES` can cause `libcuda.cuInit()` fails (returns non-zero). The CUDA driver detector detects the driver version in the current process. The user-defined environment variable will let it fail.

For example:

- An empty `CUDA_VISIBLE_DEVICES` will cause `libcuda.cuInit()` returns `CUDA_ERROR_NO_DEVICE` (100).

```console
$ CUDA_VISIBLE_DEVICES='' python3 -c 'from conda.plugins.virtual_packages import cuda; print(cuda.cuda_version())'
None
```

- An invalid `CUDA_VISIBLE_DEVICES` will cause `libcuda.cuInit()` returns `CUDA_ERROR_INVALID_DEVICE` (101).

```console
$ CUDA_VISIBLE_DEVICES='0,0' python3 -c 'from conda.plugins.virtual_packages import cuda; print(cuda.cuda_version())'
None
```

------

In this PR, we detect the CUDA driver version in a spawned subprocess, which unset the environment variable `CUDA_VISIBLE_DEVICES` before calling `libcuda.cuInit()`.

```console
$ CUDA_VISIBLE_DEVICES='' python3 -c 'from conda.plugins.virtual_packages import cuda; print(cuda.cuda_version())'
11.7

$ CUDA_VISIBLE_DEVICES='0,0' python3 -c 'from conda.plugins.virtual_packages import cuda; print(cuda.cuda_version())'
11.7
```

### Checklist - did you ...

### DIFF
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import ctypes
 import functools
+import multiprocessing as mp
+import os
 import platform
 from contextlib import suppress
 
@@ -23,55 +25,33 @@ def cuda_version():
 
     Returns: version string (e.g., '9.2') or None if CUDA is not found.
     """
-    # platform-specific libcuda location
-    system = platform.system()
-    if system == "Darwin":
-        lib_filenames = [
-            "libcuda.dylib",  # check library path first
-            "/usr/local/cuda/lib/libcuda.dylib",
-        ]
-    elif system == "Linux":
-        lib_filenames = [
-            "libcuda.so",  # check library path first
-            "/usr/lib64/nvidia/libcuda.so",  # RHEL/Centos/Fedora
-            "/usr/lib/x86_64-linux-gnu/libcuda.so",  # Ubuntu
-        ]
-    elif system == "Windows":
-        lib_filenames = ["nvcuda.dll"]
-    else:
-        return None  # CUDA not available for other operating systems
-
-    # open library
-    if system == "Windows":
-        dll = ctypes.windll
-    else:
-        dll = ctypes.cdll
-    for lib_filename in lib_filenames:
-        with suppress(Exception):
-            libcuda = dll.LoadLibrary(lib_filename)
-            break
-    else:
-        return None
-
-    # Get CUDA version
+    # Do not inherit file descriptors and handles from the parent process.
+    # The `fork` start method should be considered unsafe as it can lead to
+    # crashes of the subprocess. The `spawn` start method is preferred.
+    ctx = mp.get_context("spawn")
+    queue = ctx.SimpleQueue()
     try:
-        cuInit = libcuda.cuInit
-        flags = ctypes.c_uint(0)
-        ret = cuInit(flags)
-        if ret != 0:
-            return None
+        # Spawn a subprocess to detect the CUDA version
+        detector = ctx.Process(
+            target=_cuda_driver_version_detector_target,
+            args=(queue,),
+            name="CUDA driver version detector",
+            daemon=True,
+        )
+        detector.start()
+        detector.join(timeout=60.0)
+    finally:
+        # Always cleanup the subprocess
+        try:
+            detector.kill()  # requires Python 3.7+
+        except AttributeError:
+            pass
 
-        cuDriverGetVersion = libcuda.cuDriverGetVersion
-        version_int = ctypes.c_int(0)
-        ret = cuDriverGetVersion(ctypes.byref(version_int))
-        if ret != 0:
-            return None
-
-        # Convert version integer to version string
-        value = version_int.value
-        return "%d.%d" % (value // 1000, (value % 1000) // 10)
-    except Exception:
+    if queue.empty():
         return None
+
+    result = queue.get()
+    return result
 
 
 @functools.lru_cache(maxsize=None)
@@ -87,3 +67,81 @@ def conda_virtual_packages():
     cuda_version = cached_cuda_version()
     if cuda_version is not None:
         yield CondaVirtualPackage("cuda", cuda_version, None)
+
+
+def _cuda_driver_version_detector_target(queue):
+    """
+    Attempt to detect the version of CUDA present in the operating system in a
+    subprocess.
+
+    On Windows and Linux, the CUDA library is installed by the NVIDIA
+    driver package, and is typically found in the standard library path,
+    rather than with the CUDA SDK (which is optional for running CUDA apps).
+
+    On macOS, the CUDA library is only installed with the CUDA SDK, and
+    might not be in the library path.
+
+    Returns: version string (e.g., '9.2') or None if CUDA is not found.
+             The result is put in the queue rather than a return value.
+    """
+    # Platform-specific libcuda location
+    system = platform.system()
+    if system == "Darwin":
+        lib_filenames = [
+            "libcuda.dylib",  # check library path first
+            "/usr/local/cuda/lib/libcuda.dylib",
+        ]
+    elif system == "Linux":
+        lib_filenames = [
+            "libcuda.so",  # check library path first
+            "/usr/lib64/nvidia/libcuda.so",  # RHEL/Centos/Fedora
+            "/usr/lib/x86_64-linux-gnu/libcuda.so",  # Ubuntu
+            "/usr/lib/wsl/lib/libcuda.so",  # WSL
+        ]
+    elif system == "Windows":
+        lib_filenames = ["nvcuda.dll"]
+    else:
+        queue.put(None)  # CUDA not available for other operating systems
+        return
+
+    # Open library
+    if system == "Windows":
+        dll = ctypes.windll
+    else:
+        dll = ctypes.cdll
+    for lib_filename in lib_filenames:
+        with suppress(Exception):
+            libcuda = dll.LoadLibrary(lib_filename)
+            break
+    else:
+        queue.put(None)
+        return
+
+    # Empty `CUDA_VISIBLE_DEVICES` can cause `cuInit()` returns `CUDA_ERROR_NO_DEVICE`
+    # Invalid `CUDA_VISIBLE_DEVICES` can cause `cuInit()` returns `CUDA_ERROR_INVALID_DEVICE`
+    # Unset this environment variable to avoid these errors
+    os.environ.pop("CUDA_VISIBLE_DEVICES", None)
+
+    # Get CUDA version
+    try:
+        cuInit = libcuda.cuInit
+        flags = ctypes.c_uint(0)
+        ret = cuInit(flags)
+        if ret != 0:
+            queue.put(None)
+            return
+
+        cuDriverGetVersion = libcuda.cuDriverGetVersion
+        version_int = ctypes.c_int(0)
+        ret = cuDriverGetVersion(ctypes.byref(version_int))
+        if ret != 0:
+            queue.put(None)
+            return
+
+        # Convert version integer to version string
+        value = version_int.value
+        queue.put("%d.%d" % (value // 1000, (value % 1000) // 10))
+        return
+    except Exception:
+        queue.put(None)
+        return
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import ctypes
 import functools
+import itertools
 import multiprocessing as mp
 import os
 import platform
@@ -88,7 +89,9 @@ def _cuda_driver_version_detector_target(queue):
     system = platform.system()
     if system == "Darwin":
         lib_filenames = [
-            "libcuda.dylib",  # check library path first
+            "libcuda.1.dylib",  # check library path first
+            "libcuda.dylib",
+            "/usr/local/cuda/lib/libcuda.1.dylib",
             "/usr/local/cuda/lib/libcuda.dylib",
         ]
     elif system == "Linux":
@@ -98,8 +101,13 @@ def _cuda_driver_version_detector_target(queue):
             "/usr/lib/x86_64-linux-gnu/libcuda.so",  # Ubuntu
             "/usr/lib/wsl/lib/libcuda.so",  # WSL
         ]
+        # Also add libraries with version suffix `.1`
+        lib_filenames = list(
+            itertools.chain.from_iterable((f"{lib}.1", lib) for lib in lib_filenames)
+        )
     elif system == "Windows":
-        lib_filenames = ["nvcuda.dll"]
+        bits = platform.architecture()[0].replace("bit", "")  # e.g. "64" or "32"
+        lib_filenames = [f"nvcuda{bits}.dll", "nvcuda.dll"]
     else:
         queue.put(None)  # CUDA not available for other operating systems
         return
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -3,7 +3,7 @@
 import ctypes
 import functools
 import itertools
-import multiprocessing as mp
+import multiprocessing
 import os
 import platform
 from contextlib import suppress
@@ -29,11 +29,11 @@ def cuda_version():
     # Do not inherit file descriptors and handles from the parent process.
     # The `fork` start method should be considered unsafe as it can lead to
     # crashes of the subprocess. The `spawn` start method is preferred.
-    ctx = mp.get_context("spawn")
-    queue = ctx.SimpleQueue()
+    context = multiprocessing.get_context("spawn")
+    queue = context.SimpleQueue()
     try:
         # Spawn a subprocess to detect the CUDA version
-        detector = ctx.Process(
+        detector = context.Process(
             target=_cuda_driver_version_detector_target,
             args=(queue,),
             name="CUDA driver version detector",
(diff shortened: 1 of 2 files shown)


conda/plugins/virtual_packages/cuda.py
```diff
@@ -35,13 +91,15 @@ def cuda_version():
             "libcuda.so",  # check library path first
             "/usr/lib64/nvidia/libcuda.so",  # RHEL/Centos/Fedora
             "/usr/lib/x86_64-linux-gnu/libcuda.so",  # Ubuntu
+            "/usr/lib/wsl/lib/libcuda.so",  # WSL
```
XuehaiPan, line 94: The fix for WSL search path was added in #11626, but was accidentally removed in #11854. Readd it back here.

- #11626
- #11854

jezdez: OH wow, thanks for catching that.

jezdez:
@chenghlee @jakirkham Do you happen to know who we could ask to review this?

conda/plugins/virtual_packages/cuda.py
```diff
@@ -23,7 +25,63 @@ def cuda_version():
 
     Returns: version string (e.g., '9.2') or None if CUDA is not found.
     """
-    # platform-specific libcuda location
+    # Do not inherit file descriptors and handles from the parent process
+    # The `fork` start method should be considered unsafe as it can lead to
+    # crashes of the subprocess
+    ctx = mp.get_context("spawn")
+    queue = ctx.SimpleQueue()
```
jakirkham, line 32: Is there a way to do this without starting a subprocess? This feels kind of brittle

jakirkham:
Think @seibert wrote the original CUDA driver version detection logic.

@kkraus14 and @jaimergp may also have thoughts (Jaime in particular has gone down the CUDA Windows rabbit hole before).

Am happy to look as well. Though this may be something that occurs in the new year.

conda/plugins/virtual_packages/cuda.py
XuehaiPan: > Is there a way to do this without starting a subprocess?

@jakirkham Yes, another approach is to unset `"CUDA_VISIBLE_DEVICES"` in `os.environ` before calling `cuInit()` in the main process.

```python
# Empty `CUDA_VISIBLE_DEVICES` can cause `cuInit()` returns `CUDA_ERROR_NO_DEVICE`
# Invalid `CUDA_VISIBLE_DEVICES` can cause `cuInit()` returns `CUDA_ERROR_INVALID_DEVICE`
os.environ.pop("CUDA_VISIBLE_DEVICES", None)
```

The point here using subprocess is separate the `.so` handle (`fd`s) namespaces. Although it adds more complexity.

1. We never release the `.so` handle once we load `libcuda.so`. The `conda` program will always keep the NVIDIA device (`/dev/nvidia-uvm`) and kernel module (`nvidia_uvm`) in use while solving the environment. A benefit of using a subprocess here is we can release the `fd` as the subprocess ends. We use caches to memorize the result, so the subprocess will start only once.
2. Currently, `conda` is a standalone program, and it is not meant to be `import`ed. Unset and `CUDA_VISIBLE_DEVICES` and call `cuInit()` in the main process seems safe here. But if someone calls `cuInit()` with invalid `CUDA_VISIBLE_DEVICES` in advance. Then unsetting `CUDA_VISIBLE_DEVICES` and call `cuInit()` in conda internals will always fail. The `.so` handlers and `fd`s are messed up in the main process.

kkraus14:
Instead of using the CUDA driver API to accomplish this, we could use the nvml library which doesn't have the same restrictions with regards to `CUDA_VISIBLE_DEVICES`. 

Some example code:
```python
from ctypes import *

libnvml = CDLL("libnvidia-ml.so.1")
status = libnvml.nvmlInit_v2()  # Should check the result of status here to ensure it succeeded

cudaDriverVersion = c_int()
status = libnvml.nvmlSystemGetCudaDriverVersion_v2(byref(cudaDriverVersion))  # Should check the result of status here to ensure it succeeded

# cudaDriverVersion is now a ctypes c_int with the CUDA driver version
```

kkraus14:
`nvmlInit_v2` was added in the CUDA 5.5 release of NVML so this should be fine. We can move to use `nvmlInit` if we want to be even more conservative (which is silly IMO).

`nvmlSystemGetCudaDriverVersion_v2` was added in CUDA 10 release of NVML so maybe we should avoid using this. It looks like `nvmlSystemGetCudaDriverVersion` was added between CUDA 8 and CUDA 9 in driver 378 (https://github.com/NVIDIA/nvidia-settings/commit/c9e545365748181b3c685bf2a736ade37e3c997f) so that would be pretty good as well.

<hr>

A potentially bad idea: it looks like you can call `cuDriverGetVersion` without calling `cuInit` first and it still returns the correct version without error. This is definitely relying on implementation details of the CUDA driver and could change without notice in the future.

jezdez:
Hey all! Thank you @XuehaiPan for continuing to push for change.

I'm trying to understand the correctness of this patch given that we plan to tag [conda 23.1.0](https://github.com/conda/conda/milestone/58) January 16. Is this ready?

On a different notw, there is a larger discussion we could have: now that this virtual package is implemented as a plugin, we _could_ release this in a separate `conda-cuda-virtual-package` (or similar) package, maybe co-maintained with folks from Nvidia? @jakirkham @kkraus14 Any opinions on this?

XuehaiPan:
> Is this ready?

@jezdez I think it's ready, but it's depending on the opinions of the code reviewers.

------

Let me summarize the changes in this PR:

1. Re-add the `libcuda.so` search path for WSL support.
2. Put the whole CUDA detection function in a separate subprocess:

    1. **Main process**: Start a new subprocess with the `spawn` start method.
    1. **Subprocess**: Detect the CUDA version in the subprocess and put the result in a `SimpleQueue`.
        - New change: unset `CUDA_VISIBLE_DEVICES` before initializing the CUDA driver context.
    1. **Main process**: receive the result from the subprocess and return the result to the function caller.

**Rationale**

1. **Loading a `.so` file in the main process is not safe.** If the driver library is corrupted, then the `conda` program always fails (Segmentation Fault). Putting the CUDA detector in a daemon process, if it fails, the caller can return `None` rather than crash.
2. **Better process environment management.** The CUDA driver library requires valid `CUDA_VISIBLE_DEVICES`. Using subprocesses, we can separate the environment rather than mess them together.
3. **Early release of the CUDA library handle.** We never release the `.so` library handle once we load it. If we load the CUDA library in the main process, it will cause the NVIDIA device and kernel module in use during the whole `conda` program lifetime. With subprocess, we can release the handle as soon as we get the result and exit the subprocess.

conda/plugins/virtual_packages/cuda.py
```diff
@@ -35,13 +96,15 @@ def cuda_version():
             "libcuda.so",  # check library path first
```
kkraus14, line 96: I would also add `libcuda.so.1` here as that's what a typical `ldconfig` reports and is the `SONAME` in the actual `libcuda` binaries

kkraus14:
It feels really funky to unset the `CUDA_VISIBLE_DEVICES` environment variable in a spawned subprocess. Almost feels like this is conda overstepping its bounds.

We could use NVML and avoid having to mess with the environment variable and then decide if we want to use the subprocess to more nicely load and unload the library / isolate it from crashing the main conda process.

conda/plugins/virtual_packages/cuda.py
XuehaiPan: I add extra library search paths with `.1` suffix on Linux and macOS. On Windows, extra `nvcuda64.dll` is added.

XuehaiPan:
> Almost feels like this is conda overstepping its bounds.

@kkraus14 I agree. I can open another PR to refactor with the NVML library.

One concern is the version handling and backward compatibility guarantee of the NVML library functions. Different NVML libraries ships different symbols. Newer versioned functions may have extra arguments, and the new and the old ones sometimes are not inter-replaceable.

That may require extra maintenance work. For example, check whether the library is new enough to have newer versioned functions with `hasattr(lib, '<symbol_vN>')`, otherwise fallback to `_v(N-1)`. Then call the function with proper arguments (may be different between versions). Every time when a new NVIDIA driver is released, we will also need to update the version handling again.

```cpp
#define NVML_INIT_FLAG_NO_GPUS      1   //!< Don't fail nvmlInit() when no GPUs are found
#define NVML_INIT_FLAG_NO_ATTACH    2   //!< Don't attach GPUs

nvmlReturn_t DECLDIR nvmlInit(void);
nvmlReturn_t DECLDIR nvmlInit_v2(void);
nvmlReturn_t DECLDIR nvmlInit_v3(unsigned int flags);
nvmlReturn_t DECLDIR nvmlInitWithFlags(unsigned int flags);
nvmlReturn_t DECLDIR nvmlShutdown(void);

nvmlReturn_t DECLDIR nvmlSystemGetCudaDriverVersion(int *cudaDriverVersion);
nvmlReturn_t DECLDIR nvmlSystemGetCudaDriverVersion_v2(int *cudaDriverVersion);
```

<details>
<summary>nvml.h</summary>

```cpp
/***************************************************************************************************/
/** @defgroup nvmlInitializationAndCleanup Initialization and Cleanup
 * This chapter describes the methods that handle NVML initialization and cleanup.
 * It is the user's responsibility to call \ref nvmlInit_v2() before calling any other methods, and
 * nvmlShutdown() once NVML is no longer being used.
 *  @{
 */
/***************************************************************************************************/

#define NVML_INIT_FLAG_NO_GPUS      1   //!< Don't fail nvmlInit() when no GPUs are found
#define NVML_INIT_FLAG_NO_ATTACH    2   //!< Don't attach GPUs

/**
 * Initialize NVML, but don't initialize any GPUs yet.
 *
 * \note nvmlInit_v3 introduces a "flags" argument, that allows passing boolean values
 *       modifying the behaviour of nvmlInit().
 * \note In NVML 5.319 new nvmlInit_v2 has replaced nvmlInit"_v1" (default in NVML 4.304 and older) that
 *       did initialize all GPU devices in the system.
 *
 * This allows NVML to communicate with a GPU
 * when other GPUs in the system are unstable or in a bad state.  When using this API, GPUs are
 * discovered and initialized in nvmlDeviceGetHandleBy* functions instead.
 *
 * \note To contrast nvmlInit_v2 with nvmlInit"_v1", NVML 4.304 nvmlInit"_v1" will fail when any detected GPU is in
 *       a bad or unstable state.
 *
 * For all products.
 *
 * This method, should be called once before invoking any other methods in the library.
 * A reference count of the number of initializations is maintained.  Shutdown only occurs
 * when the reference count reaches zero.
 *
 * @return
 *         - \ref NVML_SUCCESS                   if NVML has been properly initialized
 *         - \ref NVML_ERROR_DRIVER_NOT_LOADED   if NVIDIA driver is not running
 *         - \ref NVML_ERROR_NO_PERMISSION       if NVML does not have permission to talk to the driver
 *         - \ref NVML_ERROR_UNKNOWN             on any unexpected error
 */
nvmlReturn_t DECLDIR nvmlInit_v2(void);

/**
 * nvmlInitWithFlags is a variant of nvmlInit(), that allows passing a set of boolean values
 *       modifying the behaviour of nvmlInit().
 *       Other than the "flags" parameter it is completely similar to \ref nvmlInit_v2.
 *
 * For all products.
 *
 * @param flags                                 behaviour modifier flags
 *
 * @return
 *         - \ref NVML_SUCCESS                   if NVML has been properly initialized
 *         - \ref NVML_ERROR_DRIVER_NOT_LOADED   if NVIDIA driver is not running
 *         - \ref NVML_ERROR_NO_PERMISSION       if NVML does not have permission to talk to the driver
 *         - \ref NVML_ERROR_UNKNOWN             on any unexpected error
 */
nvmlReturn_t DECLDIR nvmlInitWithFlags(unsigned int flags);

/**
 * Shut down NVML by releasing all GPU resources previously allocated with \ref nvmlInit_v2().
 *
 * For all products.
 *
 * This method should be called after NVML work is done, once for each call to \ref nvmlInit_v2()
 * A reference count of the number of initializations is maintained.  Shutdown only occurs
 * when the reference count reaches zero.  For backwards compatibility, no error is reported if
 * nvmlShutdown() is called more times than nvmlInit().
 *
 * @return
 *         - \ref NVML_SUCCESS                 if NVML has been properly shut down
 *         - \ref NVML_ERROR_UNINITIALIZED     if the library has not been successfully initialized
 *         - \ref NVML_ERROR_UNKNOWN           on any unexpected error
 */
nvmlReturn_t DECLDIR nvmlShutdown(void);

/** @} */

/**
 * Retrieves the version of the CUDA driver.
 *
 * For all products.
 *
 * The CUDA driver version returned will be retreived from the currently installed version of CUDA.
 * If the cuda library is not found, this function will return a known supported version number.
 *
 * @param cudaDriverVersion                    Reference in which to return the version identifier
 *
 * @return
 *         - \ref NVML_SUCCESS                 if \a cudaDriverVersion has been set
 *         - \ref NVML_ERROR_INVALID_ARGUMENT  if \a cudaDriverVersion is NULL
 */
nvmlReturn_t DECLDIR nvmlSystemGetCudaDriverVersion(int *cudaDriverVersion);

/**
 * Retrieves the version of the CUDA driver from the shared library.
 *
 * For all products.
 *
 * The returned CUDA driver version by calling cuDriverGetVersion()
 *
 * @param cudaDriverVersion                    Reference in which to return the version identifier
 *
 * @return
 *         - \ref NVML_SUCCESS                  if \a cudaDriverVersion has been set
 *         - \ref NVML_ERROR_INVALID_ARGUMENT   if \a cudaDriverVersion is NULL
 *         - \ref NVML_ERROR_LIBRARY_NOT_FOUND  if \a libcuda.so.1 or libcuda.dll is not found
 *         - \ref NVML_ERROR_FUNCTION_NOT_FOUND if \a cuDriverGetVersion() is not found in the shared library
 */
nvmlReturn_t DECLDIR nvmlSystemGetCudaDriverVersion_v2(int *cudaDriverVersion);

/**
 * Macros for converting the CUDA driver version number to Major and Minor version numbers.
 */
#define NVML_CUDA_DRIVER_VERSION_MAJOR(v) ((v)/1000)
#define NVML_CUDA_DRIVER_VERSION_MINOR(v) (((v)%1000)/10)
```

</details>

kkraus14:
Given that none of the older functions are deprecated or removed we could just use them and hold off on any kind of versioning until needed.

See my comment here: https://github.com/conda/conda/pull/11667#issuecomment-1364168424

Driver 378, which was when `nvmlSystemGetCudaDriverVersion` was included, released ~6 years ago so I think it's pretty reasonable that we don't need to be able to detect the CUDA version on drivers older than that. Besides, the oldest version of the CUDA toolkit packaged as a conda package is 8.0 and its only packaged for Windows on the `anaconda` channel. The oldest version packaged on conda-forge is 9.2.

kkraus14 (dismissed):
LGTM other than refactoring to use nvml instead of the driver library, but we can do that in a follow up

conda/plugins/virtual_packages/cuda.py
```diff
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import ctypes
 import functools
+import itertools
+import multiprocessing as mp
```
jezdez, line 6:
```suggestion
import multiprocessing
```

conda/plugins/virtual_packages/cuda.py
```diff
@@ -23,25 +26,90 @@ def cuda_version():
...
+            target=_cuda_driver_version_detector_target,
+            args=(queue,),
+            name="CUDA driver version detector",
+            daemon=True,
+        )
+        detector.start()
+        detector.join(timeout=60.0)
+    finally:
+        # Always cleanup the subprocess
+        detector.kill()  # requires Python 3.7+
```
jezdez, line 46:
```suggestion
    # Do not inherit file descriptors and handles from the parent process.
    # The `fork` start method should be considered unsafe as it can lead to
    # crashes of the subprocess. The `spawn` start method is preferred.
    context = multiprocessing.get_context("spawn")
    queue = context.SimpleQueue()
    try:
        # Spawn a subprocess to detect the CUDA version
        detector = context.Process(
            target=_cuda_driver_version_detector_target,
            args=(queue,),
            name="CUDA driver version detector",
            daemon=True,
        )
        detector.start()
        detector.join(timeout=60.0)
    finally:
        # Always cleanup the subprocess
        detector.kill()  # requires Python 3.7+
```

jezdez (commented):
Some minor code fixes.

jezdez (approved):

jezdez:
@XuehaiPan @kkraus14 Before I forget, do you feel like it's important to file a ticket for the refactoring to use the nvml library?

Alternatively, does anyone know how likely it would be for us to extract the cuda virtual package into a separate repo? Any takers from Nvidia?

jakirkham:
It's probably a good idea to have an issue

We could look into a plugin, but that won't be for a while

Highest priority for me is getting new packages together ( https://github.com/conda-forge/staged-recipes/issues/21382 ) and starting CUDA 12 builds

leofang:
NVML should be the way to go (e.g. see [here](https://github.com/NVIDIA/cuQuantum/blob/b7f847c16413e114ddc91d415f18d44d24dd3261/benchmarks/cuquantum_benchmarks/_utils.py#L145-L176)). Sorry to see/raise this late, but this feels a bit awkward to me 😅 Like Keith said, the needed NVML symbols have existed for [very long time](https://github.com/conda/conda/pull/11667#issuecomment-1364168424), it should be preferred for detecting the driver version without initializing the context.

PR #11676, opened 2022-08-01
Better support prefixes with spaces

jaimergp:
### Description

`conda` does not have bullet-proof support for spaces in its environment paths. As a matter of fact, environment names with spaces are forbidden:

```
$ conda create -n "name with spaces" python

CondaValueError: Invalid environment name: 'name with spaces'
  Characters not allowed: ('/', ' ', ':', '#')
```

However, little can conda do when the base prefix already has spaces; e.g. Miniconda was installed to a place like `/Users/Jaime RGP/anaconda`. This is prevented at the `constructor` level, which forbids the user from choosing such a location. This is unfortunate because in some cases the user might have no other choice (e.g. managed installs where the only writable location is the user directory, which is created with `Name Last name` syntax).

Additionally, a user can still end up in that situation with a command like:

```
$ conda create -p "/tmp/env with spaces" python
```

This is not prevented, but a warning is emitted:

```
WARNING: A space was detected in your requested environment path:
'/tmp/env with spaces'
Spaces in paths can sometimes be problematic. 
```

This PR tries to fix some common issues with spaces and Unix shebang strings:

#### Python entry points
- Python entry points were naively using the path to `python` as is in the shebang; since a space was included in the path, the kernel parses it wrong and fails with a "file not found error". 
- Why? Because `#!/prefix/with spaces/python` is understood as run `/prefix/with` using argument `spaces/python`. 
- The fix for this borrows inspiration from `pypa/distlib` using a cool shell/python hybrid string :) Comments in the code.

#### Replaced hardcoded prefixes in shebangs

- Prefix replacement didn't check for the target prefix and spaces. Even a correctly built package with no spaces might acquire spaces in this step if it gets installed to a location with spaces. 
- What we do here is to escape the spaces as `\ ` (this does nothing yet) and then check for that substring in the executable path. 
- In that case, we apply the `/usr/bin/env` trick that was already being used to fix "path too long" errors. Note that this trick has a drawback: the environment MUST be active in order to use the correct executable. An alternative solution would be ideal here, but we can't assume the script will be written in Python, so the distlib trick doesn't work.

#### Info messages

I also amended some warning/info strings:

- If the prefix uses spaces, provide more information about what could go wrong.
- Quote the name or path to the freshly created environment if it contains spaces in the stdout instructions.

#### More info

If this gets through, we can accept https://github.com/conda/constructor/pull/449

### Checklist - did you ...


jezdez:
pre-commit.ci autofix

jaimergp:
Adding here the solution for #11694 - note this will still fail, we need to patch the underlying Miniconda installation. Coming in the next commit to make sure the current one fails as expected.

jaimergp:
Ah wait, not needed. `conda init --dev` recreates the installed launchers from the bundled ones :D Tests should be passing then!

tests/test_create.py
```diff
@@ -337,6 +337,8 @@ def test_noarch_python_package_with_entry_points(self):
             if on_win:
                 exe_path += ".exe"
             assert isfile(exe_path)
+            output = check_output([exe_path, "--help"], text=True)
+            assert "Usage: flask" in output
```
jaimergp, line 341: This is the new test that uncovered the Unicode / entry point / launcher hell on Windows.

jezdez: Yikes, that's so simple!

conda/core/initialize.py
```diff
@@ -746,7 +747,10 @@ def make_entry_point(target_path, conda_prefix, module, func):
         # no shebang needed on windows
         new_ep_content = ""
     else:
-        new_ep_content = "#!%s\n" % join(conda_prefix, get_python_short_path())
+        from .portability import generate_shebang_for_entry_point
```
jezdez, line 750: Can we move this to the top of the file or does this trigger a circular import?

conda/core/package_cache_data.py
```diff
@@ -326,7 +326,7 @@ def _make_single_record(self, package_filename):
                 # JsonDecodeError if info/index.json is partially extracted or corrupted
                 #   python 2.7 raises ValueError instead of JsonDecodeError
                 #   ValueError("No JSON object could be decoded")
-                log.debug("unable to read %s\n  because",
+                log.debug("unable to read %s\n  because %r",
                           join(extracted_package_dir, 'info', 'index.json'), e)
```
jezdez, line 330: Huh!

conda/core/portability.py
```diff
@@ -172,12 +187,33 @@ def replace_long_shebang(mode, data):
...
+    # Explanation: these lines are both valid Python and shell :)
+    # 1. Python will read it as a triple-quoted multiline string; end of story
+    # 2. The shell will see:
+    #       * '' (empty string)
+    #       * 'exec' "path/with spaces/to/python" "this file" "arguments"
+    #       * ' ''' (quoted space followed by empty string)
+    if len(shebang) > 127 or " " in shebang:
+        shebang = "#!/bin/sh\n"
+        shebang += "'''exec' " + f'"{executable}" "$0" "$@"'
+        shebang += "\n' '''\n"
```
jezdez, line 217: Do we need to consider macOS' longer max shebang compatibility here as well? https://github.com/pypa/distlib/blob/91aa92e64/distlib/scripts.py#L145-L146

conda/gateways/disk/create.py
```diff
@@ -128,16 +128,15 @@ def create_python_entry_point(target_full_path, python_full_path, module, func):
         'import_name': import_name,
     }
     if python_full_path is not None:
-        shebang = u'#!%s\n' % python_full_path
+        from ...core.portability import generate_shebang_for_entry_point
```
jezdez, line 131: It would be wonderful to have this at the top of the file.

conda/gateways/disk/create.py
```diff
@@ -128,16 +128,15 @@ def create_python_entry_point(target_full_path, python_full_path, module, func):
...
-        from ...core.portability import replace_long_shebang  # TODO: must be in wrong spot
-        shebang = replace_long_shebang(FileMode.text, shebang)
+        shebang = generate_shebang_for_entry_point(python_full_path)
     else:
         shebang = None
 
     with codecs.open(target_full_path, mode='wb', encoding='utf-8') as fo:
         if shebang is not None:
-            fo.write(shebang.decode('utf-8'))
+            fo.write(shebang)
```
jezdez, line 139: Noice!

news/11676-prefix-spaces
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Update bundled Python launchers for Windows (`conda/shell/cli-*.exe`) to match the ones found in conda-build. (#11676)
+
+### Bug fixes
+
+* Better support for shebang instructions in prefixes with spaces. (#11676)
+* Fix `noarch` entry points in Unicode-containing prefixes on Windows. (#11676)
```
jezdez, line 8:
```suggestion
* Better support for shebang instructions in prefixes with spaces. (#11676)
* Fix `noarch` entry points in Unicode-containing prefixes on Windows. (#11694)
```

Epic!

tests/core/test_portability.py
```diff
@@ -79,3 +85,62 @@ def test_replace_long_shebang(self):
...
+        assert new_expected_data == new_data
+
+        # long shebang with escaped spaces in prefix
+        shebang = b"#!/" + b"she\\ bang/" *20 + b"python --and --flags -x"
+        assert len(shebang) > 127
+        data = b'\n'.join((shebang, content_line, content_line, content_line))
+        new_data = replace_long_shebang(FileMode.text, data)
+        new_shebang = b"#!/usr/bin/env python --and --flags -x"
+        new_expected_data = b'\n'.join((new_shebang, content_line, content_line, content_line))
+        assert new_expected_data == new_data
```
jezdez, line 116: Could we make these cases one test case each?

jezdez (dismissed):
Amazing sleuthing @jaimergp! Just some code nits, but nothing major, this is huge win!

conda/core/package_cache_data.py
jaimergp: Yes, I ran into this error while debugging. The missing placeholder causes the exception to be thrown for other reasons and it's veeery misleading.

conda/core/initialize.py
jaimergp: Let's see :)

conda/gateways/disk/create.py
jaimergp: I didn't put it there because `conda.core` is "off-limits" within `conda.gateways`, according to the README :(

jezdez: Le sigh.

FFY00:
FYI, here's an explanation for shell/python hybrid entrypoint.

https://github.com/pypa/installer/pull/4#issuecomment-623668717

conda/core/package_cache_data.py
kenodegard: @jaimergp see https://github.com/conda/conda/pull/11679, need to rebase

conda/core/portability.py
```diff
@@ -27,24 +27,28 @@
...
+
 
 class _PaddingError(Exception):
     pass
 
 
-def update_prefix(path, new_prefix, placeholder=PREFIX_PLACEHOLDER, mode=FileMode.text,
-                  subdir=context.subdir):
+def update_prefix(
+    path, new_prefix, placeholder=PREFIX_PLACEHOLDER, mode=FileMode.text, subdir=context.subdir
```
kenodegard, line 38:
```suggestion
    path,
    new_prefix,
    placeholder=PREFIX_PLACEHOLDER,
    mode=FileMode.text,
    subdir=context.subdir,
```

conda/core/portability.py
```diff
@@ -69,6 +73,19 @@ def _update_prefix(original_data):
...
 def replace_prefix(mode, data, placeholder, new_prefix):
     if mode == FileMode.text:
+        if not on_win:
+            # if new_prefix contains spaces, it might break the shebang!
+            # handle this by escaping the spaces early, which will trigger a
+            # /usr/bin/env replacement later on
+            newline_pos = data.find(b"\n")
+            if newline_pos > -1:
+                shebang_line, rest_of_data = data[:newline_pos], data[newline_pos:]
+                shebang_placeholder = b"#!" + placeholder.encode('utf-8')
```
kenodegard, line 83:
```suggestion
                shebang_placeholder = f"#!{placeholder}".encode('utf-8')
```

conda/core/portability.py
```diff
@@ -165,19 +182,40 @@ def replace_long_shebang(mode, data):
...
         shebang_match = re.match(SHEBANG_REGEX, data, re.MULTILINE)
         if shebang_match:
             whole_shebang, executable, options = shebang_match.groups()
-            if len(whole_shebang) > 127:
-                executable_name = executable.decode('utf-8').split('/')[-1]
-                new_shebang = '#!/usr/bin/env %s%s' % (executable_name, options.decode('utf-8'))
-                data = data.replace(whole_shebang, new_shebang.encode('utf-8'))
+            prefix, executable_name = executable.decode("utf-8").rsplit("/", 1)
+            if len(whole_shebang) > MAX_SHEBANG_LENGTH or "\\ " in prefix:
+                new_shebang = "#!/usr/bin/env %s%s" % (executable_name, options.decode("utf-8"))
```
kenodegard, line 194:
```suggestion
                new_shebang = "#!/usr/bin/env {executable_name}{options.decode('utf-8')}"
```

conda/core/portability.py
```diff
@@ -165,19 +182,40 @@ def replace_long_shebang(mode, data):
...
+    # Explanation: these lines are both valid Python and shell :)
+    # 1. Python will read it as a triple-quoted multiline string; end of story
+    # 2. The shell will see:
+    #       * '' (empty string)
+    #       * 'exec' "path/with spaces/to/python" "this file" "arguments"
+    #       * ' ''' (quoted space followed by empty string)
+    if len(shebang) > MAX_SHEBANG_LENGTH or " " in shebang:
+        shebang = "#!/bin/sh\n"
+        shebang += "'''exec' " + f'"{executable}" "$0" "$@"'
+        shebang += "\n' '''\n"
```
kenodegard, line 219: oof this was hard to read, but nice to see that this hack is still possible

```suggestion
        shebang = dals(
            f"""
            #!/bin/sh
            '''exec' "{executable}" "$0" "$@"
            ' '''
            """
        )
```

tests/core/test_path_actions.py
```diff
@@ -219,10 +219,19 @@ def test_CreatePythonEntryPointAction_noarch_python(self):
         with open(py_ep_axn.target_full_path) as fh:
             lines = fh.readlines()
             first_line = lines[0].strip()
+            second_line = lines[1].strip()
+            third_line = lines[2].strip()
             last_line = lines[-1].strip()
```
kenodegard, line 224:
```suggestion
            lines = fh.read()
            last_line = lines.splitlines()[-1].strip()
```

tests/core/test_path_actions.py
```diff
@@ -219,10 +219,19 @@ def test_CreatePythonEntryPointAction_noarch_python(self):
...
         if not on_win:
             python_full_path = join(self.prefix, get_python_short_path(target_python_version))
-            assert first_line == "#!%s" % python_full_path
+            if " " in self.prefix:
+                # spaces in prefix break shebang! we use this python/shell workaround
+                # also seen in virtualenv
+                assert first_line == "#!/bin/sh"
+                assert second_line == "'''exec' " + f'"{python_full_path}" "$0" "$@"'
+                assert third_line == "' '''"
```
kenodegard, line 232:
```suggestion
                assert lines.startswith(
                    dals(
                        f"""
                        #!/bin/sh
                        '''exec' "{python_full_path}" "$0" "$@"
                        ' '''
                        """
                    )
                )
```

tests/core/test_path_actions.py
```diff
@@ -219,10 +219,19 @@ def test_CreatePythonEntryPointAction_noarch_python(self):
...
+            else:
+                assert first_line == "#!%s" % python_full_path
```
kenodegard, line 234:
```suggestion
                assert lines.startswith(f"#!{python_full_path}\n")
```

tests/core/test_portability.py
```diff
@@ -4,78 +4,151 @@
...
+    """
+    In order to deal with spaces and shebangs, we first escape the spaces
+    in the shebang and then post-process it with the /usr/bin/env trick.
+
+    However, we must NOT escape other occurrences of the prefix in the file.
+    """
+    new_prefix = "/a/path/with/s p a c e s"
+    contents = f"""#!{PREFIX_PLACEHOLDER}/python
+data = "{PREFIX_PLACEHOLDER}"
+"""
```
kenodegard, line 143:
```suggestion
    contents = dals(
        f"""
        #!{PREFIX_PLACEHOLDER}/python
        data = "{PREFIX_PLACEHOLDER}"
        """
    )
```

kenodegard (changes requested):
Minor nits for easier reading

kenodegard (approved):

jaimergp:
I think this is ready to merge, @conda/conda-core!

jezdez (approved):

PR #11435, opened 2022-04-26
Add Plugin Mechanism

beeankha:
A `conda` plugin system was initially implemented by @FFY00 in https://github.com/conda/conda/pull/11121; this initial PR will remain open for reference until all of the work discussed in Issue https://github.com/conda/conda/issues/11112 is complete. 

* * * 

This particular PR's purpose is to split out the mechanism/infrastructure parts of that work so that we can merge the new plugins feature via manageable chunks that can be fully tested _before_ implementing specific plugins (e.g., virtual packages, authentication, shell integration, license acceptance, etc). 

We will also be adding documentation relevant to the work being implemented in this PR (a simple tutorial, an overview of _how_ the infrastructure portion of the plugins works, etc)

PR #11680, opened 2022-08-02
Experimental 🚧 👷 👷‍♀️: Initial plugins implementation

travishathaway:
This pull request was largely inspired by this pull request: https://github.com/conda/conda/pull/11435

What I am trying to do differently here is to give plugin authors a way to define a command line interface for their sub-commands. This is accomplished by telling `argparse` in `conda` not to parse arguments for sub-commands it knows to belong to plugins.

This is accomplished by overriding the `parse_args` method on our custom `ArgumentParser` class:

https://github.com/travishathaway/conda/blob/9277f1e439f890bbbf557499d4045f646bb76034/conda/cli/conda_argparse.py#L228-L236

If it's a known plugin, we simply do nothing. Under the hood, it simply checks the value of `sys.argv[1]` to see if that string matches a name in our registered plugins list.

Later in `do_call`:

https://github.com/travishathaway/conda/blob/9277f1e439f890bbbf557499d4045f646bb76034/conda/cli/conda_argparse.py#L107-L113

We perform a naive check (maybe there's better way to do this?) to see that the `args` object doesn't have a `func` attribute. If doesn't we proceed to find the sub-command and execute its action callable.

Because this just calls a function without having parsed any arguments thus far, this leaves us free to do whatever kind of argument parsing we want. Below is an example program using `click` as the argument parser:

```python
import click
import conda.plugins

__version__ = "0.1.0"

SUMMARY = "Example implementation of conda-doctor command"
SUBCOMMAND_NAME = "doctor"


@click.group()
def cli():
    pass


@cli.command(SUBCOMMAND_NAME)
@click.version_option(__version__, prog_name="conda-doctor")
@click.argument('param_one')
def command(param_one):
    print(param_one)


@conda.plugins.register
def conda_subcommands():
    yield conda.plugins.CondaSubcommand(
        name=SUBCOMMAND_NAME,
        summary=SUMMARY,
        action=cli
    )


if __name__ == '__main__':
    cli()
```

The `click` program needs to be setup as-if it was a program using sub-commands which is why we define a `click.group`. Otherwise, it works just like you would expect it.


conda/cli/conda_argparse.py
```diff
@@ -85,17 +88,52 @@ def generate_parser():
...
+            "This is a helpful error message for plugin authors! This happens when you do not "
+            "appropriately set the p.set_defaults(func=entry_point_function)"
+        )
+
+        sys.exit(1)
+
+    if isinstance(args.func, str):
+        relative_mod, func_name = args.func.rsplit(".", 1)
+        # func_name should always be 'execute'
+        from importlib import import_module
```
jezdez, line 122: This can move to the top of the module.

conda/cli/conda_argparse.py
```diff
@@ -85,17 +88,52 @@ def generate_parser():
...
+        return getattr(module, func_name)(args, parser)
+
+    elif callable(args.func):
+        args.func(args, parser)
+
+    else:
+        raise CondaError(
+            "There is something mis-configured with your subcommands. This could be due to an "
+            "incorrectly configured plugin. Please try uninstalling any problematic plugins "
+            "you may have to remove this error."
```
jezdez, line 135: Would it help to print the `args` here so users can identify what is not working?

conda/plugins/__init__.py
```diff
@@ -23,14 +23,11 @@ class CondaSubcommand(NamedTuple):
...
-    :param action: Callable that will be run when the subcommand is invoked.
+    :param add_argument_parser: Function that adds the parser to the main conda argument parser
     """
     name: str
     summary: str
-    action: Callable[
-        [List[str]],  # arguments
-        Optional[int],  # return code
-    ]
+    add_argument_parser: Callable
```
jezdez, line 30: FWIW @FFY00 and me considered this pattern in the past and erred on the side of NOT making the `argparse` API accessible to end users, to prevent wild growth since `argparse` is arguable not a great API and we wouldn't want to encourage following anti-patterns. Supporting click etc was specifically not a design goal.

I'd suggest to look into whether we can renaming the existing design a `BasicCondaSubcommand` and adding a `ArgparseCondaSubcommand` class to handle the `argparse` specific APIs.

jezdez (commented):
FWIW the idea for the initial plugin implementation was to stay close to the previous `conda-*` API and have a similar behavior and specifically not provide an `argparse` specific plugin hook.

That was a design decision to reduce the effort needed for existing packages extending conda with `conda-*` scripts to have a transition path and gain some experience before we support more advanced techniques like supporting click etc.

I've left a few comments and think this still has value FWIW, e.g. we could move this into an own `ArgparseCondaSubcommand` class that could be an “advanced” option for plugins with more complex needs. That said, generally speaking, I would caution us to leaning into `argparse` given its obvious API problems. I'd rather have us work towards supporting tools like [click](https://click.palletsprojects.com/en/8.1.x/) or [typer](https://typer.tiangolo.com/).

conda/cli/conda_argparse.py
travishathaway: Yeah, definitely would be nice to have some detailed debugging information here just like the error report we normally show.

Issue #11689, opened 2022-08-05
`ExceptionHandler.write_out` prints to the wrong stream


kenodegard:
### What happened?

Several of us have encountered this problem a few times when tinkering with conda's argparse. E.g. https://github.com/conda/conda/pull/11435#pullrequestreview-1059455307:

```bash
$ conda my-plugin   # subcommand just raises an exception in its body
[y/N]:
```

https://github.com/conda/conda/blob/8b846957c4338188706959756e7dad1619727c59/conda/exceptions.py#L1134-L1143

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : base
    active env location : /Users/kodegard/anaconda/devenv/Darwin/envs/devenv-3.8-c
            shell level : 1
       user config file : /Users/kodegard/.condarc
 populated config files : /Users/kodegard/.condarc
          conda version : 4.13.0.post66+402d4f4f0
    conda-build version : 3.21.9
         python version : 3.8.13.final.0
       virtual packages : __osx=10.16=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /Users/kodegard/anaconda/devenv/Darwin/envs/devenv-3.8-c  (writable)
      conda av data dir : /Users/kodegard/anaconda/devenv/Darwin/envs/devenv-3.8-c/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /Users/kodegard/anaconda/devenv/Darwin/envs/devenv-3.8-c/pkgs
                          /Users/kodegard/.conda/pkgs
       envs directories : /Users/kodegard/anaconda/devenv/Darwin/envs/devenv-3.8-c/envs
                          /Users/kodegard/.conda/envs
               platform : osx-64
             user-agent : conda/4.13.0.post66+402d4f4f0 requests/2.28.1 CPython/3.8.13 Darwin/21.5.0 OSX/10.16
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```

</details>

<details>
<summary><code>conda config</code></summary>

```shell
==> /Users/kodegard/.condarc <==
auto_update_conda: False
auto_activate_base: False
auto_stack: 10
changeps1: False
channels:
  - defaults
report_errors: False
```

</details>

<details>
[...]

PR #11691, opened 2022-08-05
Simplify `ExceptionHandler._calculate_ask_do_upload ` logic

kenodegard:
### Description

Minor refactor while working on https://github.com/conda/conda/issues/11689. Opted not to include a news snippet since these are private methods.

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1182,7 +1182,7 @@ class ExceptionHandler(object):
     def handle_unexpected_exception(self, exc_val, exc_tb):
         error_report = self.get_error_report(exc_val, exc_tb)
         self.print_unexpected_error_report(error_report)
-        self._do_upload(error_report)
+        self._upload(error_report)
         rc = getattr(exc_val, 'return_code', None)
         return rc if rc is not None else 1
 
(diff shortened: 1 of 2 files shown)


conda/exceptions.py
```diff
@@ -1299,7 +1291,8 @@ def print_expected_error_report(self, error_report):
             message_builder.append('')
             self.write_out('\n'.join(message_builder))
 
-    def _calculate_ask_do_upload(self):
+    def _do_upload(self, error_report) -> None:
+        """Determine whether or not to upload the error report."""
```
jezdez, line 1295: I get the reason for the refactor but don't think `_do_upload` is a good name for this function, since it doesn't really handle the "determine whether or not to upload the error report" use case. What about this?
```suggestion
    def _should_upload(self, error_report) -> None:
        """Determine whether or not to upload the error report."""
```

jezdez (approved):
Just a minor naming issue.

Not sure about the need to use underscore prefixes for all those methods either, is this just defensive programming in case someone monkeypatches things?

kenodegard:
Yeah, I'm trying to be more defensive in indicating what downstream users should and shouldn't be trying to reuse.

conda/exceptions.py
kenodegard: since it now encapsulates the entire upload process, can we just call it `_upload`?

jezdez: Yeah, that's fine, too.

jezdez (approved):

Issue #11698, opened 2022-08-08
Missing tlz import during build


kenodegard:
### What happened?

The canary builds are failing with a `tlz` import error. This is due to the build invoking `conda init --install`.

See https://github.com/conda/conda/runs/7731850991?check_suite_focus=true#step:3:761

### Tasks

- [x] https://github.com/conda/conda/pull/11699
- [x] https://github.com/conda/conda/pull/11700
- [x] https://github.com/conda-forge/conda-feedstock/pull/172
- [ ] https://github.com/AnacondaRecipes/conda-feedstock/pull/10
- [x] https://github.com/conda/conda-build/pull/4556
- [x] https://github.com/conda/conda/pull/11704
- [x] https://github.com/conda/conda/pull/11709

jakirkham:
Perhaps we should install both `toolz` and `cytoolz`. Thought this was [already the case in conda-forge]( https://github.com/conda-forge/cytoolz-feedstock/blob/412878b301496fb3ee81989fa8e303f2b19909ab/recipe/meta.yaml#L30 ), but maybe we are missing something

kenodegard:
[The canary builds are still failing](https://github.com/conda/conda/runs/7754554546?check_suite_focus=true#step:3:775)

After some digging it's all due to an innocent `path/to/python -m conda` call that should be `path/to/conda`, a short-term fix is proposed in https://github.com/conda/conda/pull/11709

dholth:
Everything we use from toolz

- [ ] from toolz import 
- [ ] merge, 
- [ ] take, 
- [ ] interleave, 
- [ ] excepts, 
- [ ] merge_with, 
- [ ] concatv, 
- [ ] concat, 
- [ ] accumulate, 
- [ ] groupby, 
- [ ] unique

PR #11709, opened 2022-08-09
Add `conda._vendor.toolz` fallback for `tlz`

kenodegard:
### Description

Yet another patch to fix the broken canary builds. Adding fallback imports to `conda._vendor.toolz` for all `tlz` imports.

This is necessary because of how `conda build` builds packages:
1. `conda build` generates the necessary build environment (which contains the new conda dependencies, e.g. `toolz`)
2. `conda build` generates a `build_env_setup.sh` script where at the very end, it runs something like:
    ```
    ...
    eval "$('path/to/python' -m conda shell.bash hook)"
    conda activate <ENV1>
    conda activate --stack <ENV2>
    ```
    this occurs in [conda_build.build._write_sh_activation_text](https://github.com/conda/conda-build/blob/11192e0054a3cf42a8d2ee976080a707b1b640e0/conda_build/build.py#L1824-L1827).
3. `conda build` changes directories to the working directory (in this case a shallow clone of our conda repo)
4. `conda build` runs the `build_env_setup.sh` from our working directory so when `path/to/python -m conda ...` is invoked the CWD (working directory) is added to Python's search path, and hence the conda source code is loaded even though all we are attempting to do is bootstrap the build environment.

The correct fix is to change the `path/to/python -m conda ...` command to `path/to/conda ...` but this is a chicken/egg problem, so we need to keep this PR's fix in place until conda-build is updated and released.

Xref: #11698 

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -11,7 +11,10 @@ import re
 import sys
 from textwrap import dedent
 
-from tlz.itertoolz import concatv, drop
+try:
+    from tlz.itertoolz import concatv, drop
+except ImportError:
+    from conda._vendor.toolz.itertoolz import concatv, drop
 
 # Since we have to have configuration context here, anything imported by
 #   conda.base.context is fair game, but nothing more.
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -17,7 +17,10 @@ from contextlib import contextmanager
 from datetime import datetime
 import warnings
 
-from tlz.itertoolz import concat, concatv, unique
+try:
+    from tlz.itertoolz import concat, concatv, unique
+except ImportError:
+    from conda._vendor.toolz.itertoolz import concat, concatv, unique
 
 from .constants import (
     APP_NAME,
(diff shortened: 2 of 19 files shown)


jezdez (approved):
This issue keeps on giving!

dholth:
Could we replace the `tlz` abbreviation with `toolz` everywhere?

jakirkham:
Am curious, why not just use `tlz` as-is?

dholth:
I would replace all the concat/concatv with the equivalent itertools calls (already a C module) and redirect all import fallbacks through an intermediate "the toolz we ulze" module...

kenodegard:
@dholth `tlz` will attempt to import from `cytoolz` first and falls back to `toolz`, yes we can replace `concat`/`concatv` with native options in Python (`itertools.chain.from_iterable` and arg unpacking), tracking that work in https://github.com/conda/conda/issues/11718

@jakirkham would love to but unfortunately, since conda/conda-build haven't been built/packaged with `cytoolz`/`toolz` as a dependency in the past, we're running into a chicken/egg problem when building/packaging this new conda (conda-build manages to invoke conda's source code via `python -m conda shell.bash hook` when creating the build environment)

jakirkham:
Sorry @kenodegard that was a question intended for @dholth

PR #11719, opened 2022-08-10
indirection to toolz through new conda.tlz module

dholth:
### Description

Redirect toolz through a module, with only the used toolz (found with regex + sets on conda/**/*.py), for a more pleasant single `try / except`. Provide a single place to redefine the most obvious toolz in terms of itertools, reducing our toolz dependency.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -11,7 +11,7 @@ import re
 import sys
 from textwrap import dedent
 
-from conda.tlz import concatv, drop
+from conda.tlz import concatv
 
 # Since we have to have configuration context here, anything imported by
 #   conda.base.context is fair game, but nothing more.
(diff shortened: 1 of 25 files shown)


jezdez (commented):
I'm not feeling strongly about this, since we hope to get rid of it soon, but I wouldn't mind. @kenodegard what's your thoughts on this?

kenodegard:
I am not a fan of adding a short-term submodule, especially since we are attempting to remove all of this in the first place. So I don't think we want this since the try-excepts added in https://github.com/conda/conda/pull/11709 are intended to be removed in a release or two. Also, `conda._vendor.toolz` effectively implements the indirection proposed here:

https://github.com/conda/conda/blob/8b846957c4338188706959756e7dad1619727c59/conda/_vendor/toolz/__init__.py#L8-L17

so if we'd want this, we should just revert all `tlz` imports back to `conda._vendor.toolz`.

dholth:
The other point would be to make it more difficult to unintentionally use more toolz. Did you say we already have a bug for removing toolz altogether

PR #11737, opened 2022-08-19
Reverse conditional in devenv startup script to work on ZSH

kenodegard:
### Description

Not sure why this was causing an issue on ZSH since it's supposed to support the same variable expansions as in bash, but alas, here we are.

### Checklist - did you ...


jezdez (approved):
Not sure where this is coming from, I have no reason to doubt this works :)

kenodegard:
I don't remember opening this PR and can seem to replicate the issue anymore, so I'm in favor of closing this

PR #11749, opened 2022-08-22
Modify `ArgumentParser` to print to `stderr` instead of `stdout`

### LINKED ISSUE #11748: `conda shell.posix activate/deactivate` should write out to `fd` 3, not `stdout`
### Description

~Since `conda activate` and `conda deactivate` are special commands that actually invoke hidden subcommands (i.e. `conda shell.posix activate|deactivate`) and evaluate what is printed to `stdout` we need a different way to communicate with the user for things like `conda activate --help`.~

After the initial PR #11749 we realized that we didn't like the consequences of moving everything over to `stderr`.

Instead, we reimagined the problem and realized it should instead be possible to use different file descriptors (`fd`) to pipe out the content that the `conda activate/deactivate` command is expected to evaluate.

```[tasklist]
- [x] https://github.com/conda/conda/pull/11862
- [ ] https://github.com/conda/conda/pull/11865
- [ ] https://github.com/conda/conda/pull/11812
- [ ] Change: Need to update `dev/**/*` scripts to use the new `--fd` and `--tmp` arguments when calling `conda init --dev`
- [ ] Change: Expose `--fd` and `--tmp` to `conda shell.* hook`
- [ ] Add: Implement `argparse` for `conda.cli.main.main_sourced`
```

kenodegard:
### Description

Working towards supporting `argparse` for `activate`/`deactivate` commands. Since those commands are based on evaluating the output printed to `stdout` we need to ensure instead that all `argparse` outputs are dumped to `stderr`.

Resolves #11748 

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -161,8 +161,15 @@ class ArgumentParser(ArgumentParserBase):
 
         super(ArgumentParser, self).error(message)
 
-    def print_help(self):
-        super(ArgumentParser, self).print_help()
+    def print_usage(self, file=None):
+        if file is None:
+            file = sys.stderr
+        super().print_usage(file)
+
+    def print_help(self, file=None):
+        if file is None:
+            file = sys.stderr
+        super().print_help(file)
 
         if sys.argv[1:] in ([], [''], ['help'], ['-h'], ['--help']):
             from .find_commands import find_commands
@@ -170,8 +177,8 @@ class ArgumentParser(ArgumentParserBase):
             if other_commands:
                 builder = ['']
                 builder.append("conda commands available from other packages:")
-                builder.extend('  %s' % cmd for cmd in sorted(other_commands))
-                print('\n'.join(builder))
+                builder.extend("  %s" % cmd for cmd in sorted(other_commands))
+                self._print_message("\n".join(builder), file)
 
 
 def _exec(executable_args, env_vars):


conda/cli/conda_argparse.py
```diff
@@ -161,17 +161,24 @@ def error(self, message):
...
         if sys.argv[1:] in ([], [''], ['help'], ['-h'], ['--help']):
             from .find_commands import find_commands
             other_commands = find_commands()
             if other_commands:
                 builder = ['']
                 builder.append("conda commands available from other packages:")
-                builder.extend('  %s' % cmd for cmd in sorted(other_commands))
-                print('\n'.join(builder))
+                builder.extend("  %s" % cmd for cmd in sorted(other_commands))
+                self._print_message("\n".join(builder), file)
```
dholth, line 181: It looks like these lines should be executed by the test, but pytest-cov doesn't find it because it's in a subprocess

travishathaway (changes requested):
@kenodegard, I have a couple concerns about switching help messages to `stderr`. It basically breaks a common pattern of being able to page through help pages with the following command:

```bash
conda --help | less
```

☝️ I use that a lot for many commands (not just `conda`). I imagine others may too.

Furthermore, I really don't think it is a good idea to begin with. Help messages are not error messages, and this could have other unintended consequences for our users.

PR #11754, opened 2022-08-23
Fix failing stacking tests

kenodegard:
### Description

Stacking tests randomly started to fail. See https://github.com/conda/conda/runs/7978894522?check_suite_focus=true#step:3:1723

After much debugging found that the test is still using the old (before dropping the prefix injection) shell interface scripts:
Screen Shot 2022-08-23 at 3 15 48 PM

This is a short-term solution as we continue working on redesigning the CI scripts (e.g. https://github.com/conda/conda-build/issues/4558)

### Checklist - did you ...

### DIFF
--- a/dev/linux/integration.sh
+++ b/dev/linux/integration.sh
@@ -5,8 +5,8 @@ set -o errtrace -o pipefail -o errexit
 TEST_SPLITS="${TEST_SPLITS:-1}"
 TEST_GROUP="${TEST_GROUP:-1}"
 
-eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
 sudo su root -c "/opt/conda/bin/conda install -yq conda-build"
+eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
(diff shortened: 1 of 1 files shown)


dev/linux/integration.sh
```diff
@@ -5,8 +5,8 @@ set -o errtrace -o pipefail -o errexit
 TEST_SPLITS="${TEST_SPLITS:-1}"
 TEST_GROUP="${TEST_GROUP:-1}"
 
-eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
 sudo su root -c "/opt/conda/bin/conda install -yq conda-build"
```
dholth, line 8: is this better than `sudo /opt/conda/bin/conda install -yq conda-build`

kenodegard: No idea, sought to keep the fix as small as possible as a larger overhaul is needed soon anyhow.

kenodegard: I honestly think most of these quirks can be removed without any ill effects

beeankha (approved):

PR #11755, opened 2022-08-23
specify --basetemp

dholth:
### Description

The tests warn against having this on a separate filesystem. Does this pass more quickly?


dholth:
The tests should actually run `conda info` to find the writeable directory, not `/opt/conda`, and check if it is linkable. But `/tmp/` fails that test too.

jezdez:
@dholth Not sure if this is still something you intend to work on, do you know what we'd need to do to have a good base temp directory?

dholth:
IIRC the error message may be wrong. When running under docker the source code is on a readonly shared filesystem and everything else isn't.

jezdez:
@dholth Let's see if mounting /tmp as a tmpfs helps?

jezdez:
Okay, let's close it out then. Thanks for the context Daniel!

PR #11766, opened 2022-08-27
remove toolz.drop()

dholth:
### Description

Replace toolz.drop() with stdlib equivalent.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -202,8 +202,7 @@ class _Activator(object):
         if arguments is None or len(arguments) < 1:
             raise_invalid_command_error()
 
-        command = arguments[0]
-        arguments = tuple(arguments[1:])
+        command, *arguments = arguments
         help_flags = ('-h', '--help', '/?')
         non_help_args = tuple(arg for arg in arguments if arg not in help_flags)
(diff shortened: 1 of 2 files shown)


conda/activate.py
```diff
@@ -203,7 +203,7 @@ def raise_invalid_command_error(actual_command=None):
             raise_invalid_command_error()
 
         command = arguments[0]
-        arguments = tuple(drop(1, arguments))
+        arguments = tuple(arguments[1:])
```
kenodegard, line 206:
```python
command, *arguments = arguments
```

kenodegard (changes requested):
Minor suggestion but otherwise good to go

conda/activate.py
dholth: The old code seemed really excited about keeping everything as a tuple()

kenodegard: yeah, I noticed, I guess leaving it as a tuple provides some level of guarantee 🤷🏼‍♂️

but in this case `arguments` is only a local var and isn't passed to another function, so I see very little potential for dangerous mutations

kenodegard (approved):

PR #11767, opened 2022-08-29
tests: remove dead code

dholth:
### Description

`def fixurl()` is a broken function that is not called. Appears to be left over from Python 2.


kenodegard (approved):

dholth:
Thanks @kenodegard

PR #11784, opened 2022-08-31
Add `conda.plugins.CondaSubcommand` rich comparisons

kenodegard:
### Description

Adding comparisons and hashing for easier sorting and contains/unique/conflict checking.

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -12,6 +12,7 @@ from argparse import (
     _CountAction,
     _HelpAction,
 )
+from functools import chain
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
@@ -121,21 +122,14 @@ class ArgumentParser(ArgumentParserBase):
             self.description += "\n\nOptions:\n"
 
         pm = context.get_plugin_manager()
-        self._subcommands = sorted(
-            (
-                subcommand
-                for subcommands in pm.hook.conda_subcommands()
-                for subcommand in subcommands
-            ),
-            key=lambda subcommand: subcommand.name,
-        )
+        self._subcommands = sorted(chain.from_iterable(pm.hook.conda_subcommands()))
 
         # Check for conflicts
         seen = set()
         conflicts = [
             subcommand
             for subcommand in self._subcommands
-            if subcommand.name in seen or seen.add(subcommand.name)
+            if subcommand in seen or seen.add(subcommand)
         ]
         if conflicts:
             raise PluginError(
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -12,7 +12,7 @@ from argparse import (
     _CountAction,
     _HelpAction,
 )
-from functools import chain
+from itertools import chain
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
--- a/conda/plugins/__init__.py
+++ b/conda/plugins/__init__.py
@@ -28,6 +28,24 @@ class CondaSubcommand(NamedTuple):
         Optional[int],  # return code
     ]
 
+    def __eq__(self, other):
+        return other.name == self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __lt__(self, other):
+        return other.name < self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __le__(self, other):
+        return other.name <= self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __gt__(self, other):
+        return other.name > self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __ge__(self, other):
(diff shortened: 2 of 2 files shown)


conda/cli/conda_argparse.py
```diff
@@ -12,6 +12,7 @@
     _CountAction,
     _HelpAction,
 )
+from functools import chain
```
pkmooreanaconda: I think chain is in ```itertools``` rather than ```functools```.

I'm not sure if this is something that has moved around between Python versions

https://docs.python.org/3/library/itertools.html

conda/cli/conda_argparse.py
```diff
@@ -121,21 +122,14 @@ def __init__(self, *args, **kwargs):
...
-        )
+        self._subcommands = sorted(chain.from_iterable(pm.hook.conda_subcommands()))
 
         # Check for conflicts
         seen = set()
         conflicts = [
             subcommand
             for subcommand in self._subcommands
-            if subcommand.name in seen or seen.add(subcommand.name)
+            if subcommand in seen or seen.add(subcommand)
```
pkmooreanaconda, line 132: I had to stare at this one for a bit until I understood what was going on with the short circuit eval.  I think the logic might be clearer (but perhaps uglier?) written out as a loop but this isn't the place to make that call

pkmooreanaconda (changes requested):
Looks good to go once the test failures are sorted.

conda/cli/conda_argparse.py
```diff
@@ -12,6 +12,7 @@
     _CountAction,
     _HelpAction,
 )
+from itertools import chain
```
pkmooreanaconda, line 15: 👍

pkmooreanaconda (approved):
Looks good pending Windows tests

conda/cli/conda_argparse.py
jezdez: I'm not sure if we need to store the whole subcommand hook instance in the set, what do we get from that?

conda/plugins/__init__.py
```diff
@@ -28,6 +28,24 @@ class CondaSubcommand(NamedTuple):
...
+        return other.name <= self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __gt__(self, other):
+        return other.name > self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __ge__(self, other):
+        return other.name >= self.name if isinstance(other, type(self)) else NotImplemented
+
+    def __hash__(self):
+        return hash(self.name)
```
jezdez, line 47: I wish we'd be able to use dataclasses or pydantic instead of overloading all of these :( @FFY00 Remind me, why did we go with NamedTuple again instead of data classes?

kenodegard: Per my understanding of comparison magic methods:
- `cmp`/`__cmp__` was dropped in Python 3,
- the implied shortcut of only defining `__eq__` and one of `__lt__`, `__le__`, `__gt__`, or `__ge__` from Python 2 requires using [`@functools.total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering) in Python 3, the docs state that using this decorator has a performance hit so I opted not to use it

conda/cli/conda_argparse.py
kenodegard: @pkmooreanaconda I agree this isn't the clearest trick in the book, I've struggled for years trying to make this trick more obvious.

I considered making a new helper function where it would be easier to explain what is happening, something like:

```python
def unique(*iterables):
    # unique list of items in order of occurrence
    ordered = []
    # for every item in our iterables, we check to see if it has already been seen,
    # if it has, its added to the list of duplicates, otherwise it's appended to ordered
    duplicates = [
        item
        for item in chain.from_iterable(iterables)
        # shortcircuit for whether to add to duplicates list or to ordered list
        if item in ordered or ordered.append(item)
    ]
    return ordered, duplicates
```

And then in `conda/cli/conda_argparse.py`:

```python
self._subcommands, conflicts = unique(*pm.hook.conda_subcommands())
```

@jezdez you're right we don't need the entire object, I'd originally thought to define `conda.plugins.CondaSubcommand.__str__` to avoid the extra step of extracting the name in the conflicts error

dholth: I would tend to want to just put this inline...

I see toolz unique() doesn't collect duplicates.

The old dict key : true trick could be used to do an ordered set, then iterate over the keys to return a list.

-1 on "if item in list"

dholth:
```
nonduplicates = {}
duplicates = [ item for item in chain.from_iterable(iterables) if nonduplicates.setdefault(item, True) ]
return list(nonduplicates.keys()), duplicates
```

kenodegard:
This is stale.

PR #11785, opened 2022-08-31
Bring `get_plugin_manager` inline to `conda.base.context`

kenodegard:
### Description

Since the `get_plugin_manager` isn't called directly, let's just fold the entire implementation into the `context` object.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -155,14 +155,6 @@ def ssl_verify_validation(value):
     return True
 
 
-@functools.lru_cache(maxsize=None)  # FUTURE: Python 3.9+, replace w/ functools.cache
-def get_plugin_manager():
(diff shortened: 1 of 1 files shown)


jezdez (commented):
I don't remember why we implemented this as a module level function, but this would simplify the implementation. Can you think about uses of the plugin manager outside of the context-aware scenarios?

kenodegard:
@jezdez no idea, perhaps @beeankha knows?

an alternative question, since we now have the `conda.plugins` submodule can we just move this there?

kenodegard:
@jezdez ah looks like it's origin is from the [getting started docs](https://pluggy.readthedocs.io/en/latest/#the-host):
Screen Shot 2022-09-12 at 4 56 22 PM

Issue #11600, opened 2022-07-01
Create a "fetch" plugin prototype


FFY00:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

There are several use cases that are desirable to be customizable that could be tackled by implementing a "repo interaction" plugin. Some examples are, authentication, replacing/modifying the transport layer, or even interfacing with custom repo implementations.

The idea behind this is that conda would provide a plugin that allows to completely replace the code where it interacts with the remote repository.

This has some challenges, as the current codebase is not in the best shape to implement this. And there are several parts of the code that interact with the repo and need hooking into.

Additionally, ideally, it would be good to also provide the original implementation as a plugin, so that people can re-use parts of it if they just need minor modifications.

FFY00:
Even though the plugin scope will be quite broad, we should gather all the requirements in this issue, so that we can make sure all of them are addressed.

cc @jezdez

jezdez:
Part of the use cases are covered in https://github.com/conda-incubator/ceps/pull/14

FFY00:
Okay, there are several code paths which this plugin can impact, depending on the design.

- `conda.connection.adapters`
	- Essentially exposes the `requests` package
- `conda.connection.session.CondaSession`
	- Abstracts `requests.Session`, which handles remote interaction (HTTP(S), FTP, etc.)
- `conda.subdir_data.fetch_repodata_remote_request`
	- Fetches the repo data from the remote repo
		- Receives a URL and the target repodata file and performs an HTTP request
- `conda.core.subdir_data.SubdirData._load`
	- Loads the
- `conda.trust.signature_verification._SignatureVerification._fetch_channel_signing_data`
	- Fetches signing data from the repo

Currently, the repo interaction is implemented by a couple functions which receive the URL and repodata name. They then use `CondaSession` to access the remote, and also implement cache. Changing this is tricky as the code is closely coupled. The channel loading by `SubdirData.query_all`, but I am not sure if there are hidden places that go around that code path.

After carefully reviewing our options, I think the best design for now is replacing the url that gets passed around with a callable that fetches the data for a given repodata name. However, this is not optimal, and in my opinion it'd be good to uncouple the repodata from these steps, passing just a callable that already has a repodata in mind, however that is not possible right now.

[...]

PR #11788, opened 2022-09-01
[WIP] Repo plugin prototype

FFY00:
Per https://github.com/conda/conda/issues/11600#issuecomment-1238462096

> The goal is to provide a design for the plugin data structures and extract the current implementation to what will become the default implementation plugin.

This PR just implements the data structures and prepares the codebase for the plugin system to be able to hook into it.

TODO:

- [X] Implement repodata fetching
- [ ] Implement package fetching

After this, we can open a proper PR against the plugins branch.

### DIFF
--- a/conda/core/repo.py
+++ b/conda/core/repo.py
@@ -1,0 +1,261 @@
+import abc
+import bz2
+import json
+import logging
+import os
+import warnings
+
+from typing import List, Dict, Optional, TypedDict
+
+from conda.auxlib.ish import dals
+from conda.auxlib.logz import stringify
+
+from conda.base.context import context
+from conda.base.constants import CONDA_HOMEPAGE_URL, REPODATA_FN
+from conda.common.compat import ensure_text_type
+from conda.common.url import join_url, maybe_unquote
+from conda.exceptions import (
+    CondaDependencyError,
+    CondaHTTPError,
+    CondaSSLError,
+    UnavailableInvalidChannel,
+    ProxyError,
+)
+from conda.gateways.connection import (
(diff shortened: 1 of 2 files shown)


beeankha:
@FFY00 could you please point this PR to the `feature/plugins` feature branch?

beeankha:
(connecting issue https://github.com/conda/conda/issues/11600)

FFY00:
I have purposely not based this on the plugins branch, as that's fast moving, and I didn't want to have to deal with keeping up with it for the prototype. I'll definitely point at the plugins branch when actually proposing an implementation. Does this sound reasonable to you? Or do you want me to change the approach? :sweat_smile:

beeankha:
> I have purposely not based this on the plugins branch, as that's fast moving, and I didn't want to have to deal with keeping up with it for the prototype. I'll definitely point at the plugins branch when actually proposing an implementation. Does this sound reasonable to you? Or do you want me to change the approach? 😅

One reason we might want to point to the feature branch sooner rather than later is so that we can see if everything integrates together well, but I understand not wanting to keep up with a feature branch while things are still in the early stages! 👍

FFY00:
The repodata implementation seems to be working properly now, we just need to port the tests and add some backport imports to get the tests to pass.

dholth:
I'm working on a less dramatic extension of this change at https://github.com/conda/conda/pull/12003 which introduces `gateways.repodata` and should pass tests.

I think it will be necessary to get this working as a not-plugin and then refactor it, again, into a plugin once we have several use cases working nicely.

Issue #5021, opened 2017-04-07
Exclude Pre-Releases


gabm:
I read about versioning quite in detail in [PEP 440](https://www.python.org/dev/peps/pep-0440) and conda's effort to include it #1601...

That seems to work very well, but I was wondering, whether one could exclude pre-releases from conda? 

If I have two packages: 
```
1.0.0
1.0.1dev1
```
... then clearly the second is newer than the first. So if I call `conda install mypkg` I'll get the second. 

Is there a way, so that conda ignores developement releases, if not specifically demanded?  [PEP 440 - Handling of Pre-Releases](https://www.python.org/dev/peps/pep-0440/#handling-of-pre-releases) says:

> Dependency resolution tools SHOULD also allow users to request the following alternative behaviours:
> - accepting pre-releases for all version specifiers
> - excluding pre-releases for all version specifiers (reporting an error or warning if a pre-release is already installed locally, or if a pre-release is the only way to satisfy a particular specifier)

Conda currently does the first option by default, can I achieve the second?

kalefranz:
Lots of discussion regarding versions in conda at https://github.com/conda/conda/issues/2071

teake:
I'm also interested in getting this to work. I couldn't find any discussion about this in the aforementioned issue. I've naively tried

```
$ conda install "mypackage!=*dev*"
```
but that gives an `InvalidVersionSpecError: Invalid version spec: =!=*dev` error (which might be a bug?). Other variants with the wildcard `*` also failed. Any help here is appreciated.

gabrielcnr:
I also need a way to constrain versions on my recipe that exclude any "dev" or "post release" versions.

analog-cbarber:
One hacky workaround for this problem is to make a copy of your last official release with the next version and a `.prev` tag (you don't necessarily have to build it again, you can copy the already built package, change its index.json entries appropriately and rename), so that anyone using `>=` will get that version instead of an earlier dev version.

For instance, if your last official release is `1.2.3` and your next dev version is going to be `1.2.4.dev0` then
you can make a copy of `1.2.3` as `1.2.4.prev`, so that `>=1.2.3` will result in you getting `1.2.4.prev`. If someone
wants a dev version they have to use one of

```
conda install 'mypackage==1.2.4.dev0'
conda install 'mypackage>=1.2.4.dev0,<1.2.4.prev
conda install 'mypackage>=1.2.4.dev0,!=1.2.4.prev'
```

[...]

PR #11797, opened 2022-09-05
PEP 440 prerelease handling

analog-cbarber:
### Description

Adds optional support for [recommended PEP 440 prerelease handling](https://peps.python.org/pep-0440/#handling-of-pre-releases).

This adds a new `prerelease_behavior` solver configuration option with the following options:

* `allow` -Treat prereleases the same as other versions. This is the default and is the same as the previous behavior.

* `exclude` - Exclude prerelease versions from consideration.

* `limit` - only include a prerelease version in a solution if it is already present in the environment index or it is the
  the only way to satisfy the match spec. This is the behavior recommended by PEP 440.

An additional `prerelease_allowed_packages` config option has also been added. This is a list of package
specifiers for packages whose versions should never be considered to be a prerelease. This is a way to get
deal with packages such as openssl that have regular versions that would otherwise look like a prerelease.
There is no corresponding command line option, because users can simply add a package spec to the command
line.

### Rationale

This addresses issue #5021.

This is an important feature for those who want to be able to publish pre-release versions of their packages without
the fear that users will pick them up prematurely. Because prereleases are not currently handled different from other
releases, package publishers must either be extremely careful to ensure compatibility/safety or must publish them on
an alternate channel. As a result, prereleases on the public conda channels are currently fairly uncommon. Even with
an alternate channel, there is no way to specify to prevent users from picking up any other prereleases on that channel even if they only needed one for a particular package.

With this change, when the behavior is set to `limit`, the treatment of prerelease packages will be the same as it is with pip
and package publishers will be able to safely publish prerelease versions without the risk that users will get those versions before they are ready.

### Implementation notes

* This adds a `is_prerelease` method to the Context class. This seems to mostly just have properties and descriptive methods, but I could not think of a better place to put this.

* The prerelease version test is simply a regular expression: `r'.*\.\d*(?!post\d*$)[a-zA-Z]+\d*$'`.
  I don't know if this covers all the cases that technically should be considered, but having looked at
  the contents of the public indexes and conda-forge, I think it should match all prerelease versions
  that I have seen in the field.

* There are unit tests for everything except for the code in the cli/install that avoids accepting a solution from
  current_repodata.json when limiting prereleases.

### Future work

Since it would be more in line with the PEP 440 recommendations and the current pip behavior, 
It would be reasonable to make `limit` the default behavior at some point in the future. If so, we
would probably want to emit a deprecation warning prior to that when using the default.

The `current_repodata.json` file currently just contains the most recent version of each package, but
if that version is a prerelease version that is used in a solution under the `limit` behavior, the code will
have to retry using the full repodata thus defeating the optimization. To avoid this the `current_repodata.json`
file should contain both the latest version of the package and the latest non-prerelease version of the package.
To take advantage of that, the `current_repodata.json` would also have to contain some sort of indicator that
it implements the new format.


analog-cbarber:
The test failure appears to be a flakey failure. Does not appear to have anything to do with these changes.

analog-cbarber:
I am sure you are all busy, but is this going to get reviewed eventually? Is there something else I need to do?

analog-cbarber:
More flakey failures in unrelated tests.

jezdez:
> I am sure you are all busy, but is this going to get reviewed eventually? Is there something else I need to do?

Hey @analog-cbarber, thank you for opening this PR, we definitely value your contribution and will get to reviewing this as soon as we can. PEP 440 is certainly something we need to take into account.

Since this changing conda solver behavior, I think it would help to hear what @chenghlee, @wuoulf and @jaimergp think about the proposal.

analog-cbarber:
Thanks! 

Note that the default behavior is the same as the current behavior, so nothing should change unless you specify different prerelease behavior.

jaimergp:
I need to take a deeper look, but from a quick glance I must say this is a really high quality PR! Code reads well, added convenience methods where necessary, touched up the docs and added tests. Thank you so much!

I'll have to use this locally as well to get a feeling on how the different options work and then provide a thorough review.

We'll also need to think of how this can be ported over to libmamba/libsolv.ç

Thanks again!

conda/resolve.py
```diff
@@ -977,6 +983,10 @@ def generate_version_metrics(self, C, specs, include0=False):
 
         sdict = {}  # Dict[package_name, PackageRecord]
 
+        prerelease_penalty = 0
+        if context.prerelease_behavior is not PrereleaseBehavior.ALLOW:
+            prerelease_penalty = 100
```
jaimergp, line 988: Is there any specific reason we use `100` as penalty?

conda/base/context.py
```diff
@@ -609,6 +620,20 @@ def aggressive_update_packages(self):
         from ..models.match_spec import MatchSpec
         return tuple(MatchSpec(s) for s in self._aggressive_update_packages)
 
+    @memoizedproperty
+    def prerelease_allowed_specs(self):
+        from ..models.match_spec import MatchSpec
+        return tuple(MatchSpec(s) for s in context.prerelease_allowed_packages)
```
jaimergp, line 626: We should probably de-duplicate items here by having the MatchSpec `merge`d. E.g. `MatchSpec.merge(*specs)`.

conda/base/context.py
```diff
@@ -609,6 +620,20 @@ def aggressive_update_packages(self):
...
+
+    def is_prerelease(self, rec):
+        """True if PackageRecord or MatchSpec specifies a prerelease
+
+        Will return false if package is a prerelease but matches a spec in
+        prerelease_allowed_packages.
+        """
+        return (rec.is_prerelease and
+                not any(spec.match(rec) for spec in self.prerelease_allowed_specs))
```
jaimergp, line 635: I am not sure I like this method here. I know you want to add a context-aware method but to me it feels foreign to have this kind of version-specific logic here.

Maybe we can remove it from here and leave the VersionOrder and MatchSpec implementations only, as methods (not properties) that take `prerelease_allowed_specs` as an optional kwarg, defaulting to None, which is replaced by the context option at call-time. Is that reasonable?

conda/base/context.py
```diff
@@ -1396,6 +1423,28 @@ def description_map(self):
...
+            prerelease_behavior=dals(
+                """
+                Specifies how to handle prerelease versions (e.g. 1.2.3.dev3) when installing
+                or updating. The options are 'allow' (the default), 'limit', and 'exclude'.
+
+                When 'allow' is selected, prerelease versions are treated the same as any
+                other version.
+
+                When 'limit' is selected, prerelease packages will only be used if already
+                present or is the only way to satisfy a dependency.
```
jaimergp, line 1435: I might have missed this, but where are we checking whether a given package is already using a pre-release version and thus pre-release updates should be allowed? This sounds like the trickiest mode to implement, so we need tests covering these cases specifically.

conda/cli/conda_argparse.py
```diff
@@ -1708,6 +1708,13 @@ def add_parser_solver_mode(p):
         default=NULL,
         help="Ignore pinned file.",
     )
+    solver_mode_options.add_argument(
+        "--prerelease-behavior",
```
jaimergp, line 1712: I think this can just be `--prerelease`. Behavior sounds redundant here, in the context of a choice flag. Let's hear from other reviewers though, I am not a native speaker :)

conda/cli/install.py
```diff
@@ -237,6 +238,7 @@ def install(args, parser, command='install'):
         UpdateModifier.UPDATE_SPECS)) and not newenv
 
     for repodata_fn in repodata_fns:
+        is_last_repodata = repodata_fn == repodata_fns[-1]
```
jaimergp, line 241: Awesome. This kind of change improves readability a lot!

conda/cli/install.py
```diff
@@ -264,13 +266,19 @@ def install(args, parser, command='install'):
                     force_reinstall=context.force_reinstall or context.force,
                     should_retry_solve=(_should_retry_unfrozen or repodata_fn != repodata_fns[-1]),
                 )
+
+                if (context.prerelease_behavior is PrereleaseBehavior.LIMIT and
+                        not is_last_repodata and
+                        unlink_link_transaction.adds_prereleases):
```
jaimergp, line 272: I know the rest of the codebase is written with a different style, but as per the new style guide, we are trying to become "darker" little by little :D As a result, can we write this like this?

```suggestion
                if (
                    context.prerelease_behavior is PrereleaseBehavior.LIMIT 
                    and not is_last_repodata
                    and unlink_link_transaction.adds_prereleases
                ):
```

conda/core/link.py
```diff
@@ -188,6 +188,15 @@ def __init__(self, *setups):
...
 
+    @property
+    def adds_prereleases(self) -> bool:
+        """True if a new/updated spec is a prerelease version"""
+        for stp in self.prefix_setups.values():
+            for prec in list(stp.link_precs) + list(stp.update_specs):
+                if context.is_prerelease(prec):
+                    return True
+        return False
+
```
jaimergp, line 199: Is this doing a per-package check or a per-transaction check? I understand that the PEP440 specifies that if a _package_ is already using a pre-release, then it should be able to update to a pre-release too. However, this doesn't imply that the whole environment should opt-in. Is that too strict of a reading? Happy to hear otherwise, but needs to be well documented.

conda/models/match_spec.py
```diff
@@ -222,6 +222,11 @@ def is_name_only_spec(self):
                 and 'name' in self._match_components
                 and self.name != '*')
 
+    @property
+    def is_prerelease(self) -> bool:
+        """True if spec contains an exact version spec for a prerelease version."""
+        return self.version is not None and self.version.is_prerelease
```
jaimergp, line 228: Note that version specs can have `*` globs, so I don't know how the pre-release regex string is handling those, or what the answer should be. I'd like to hear your thoughts on this... I guess it should depend on the context option chosen?

conda/models/records.py
```diff
@@ -262,6 +263,11 @@ class PackageRecord(DictSafeMixin, Entity):
         default=None, required=False, nullable=True, default_in_dump=False
     )
 
+    @property
+    def is_prerelease(self) -> bool:
+        """True if package has prerelase version ending with an alphabetic tag other than 'post'"""
+        return bool(version_prerelease_re.match(self.version))
+
```
jaimergp, line 270: I'd rather see this as a method that takes the optional kwarg mentioned earlier for the context configuration. E.g. `openssl <3` should never make this return True.

conda/models/version.py
```diff
@@ -502,6 +503,13 @@ def __init__(self, vspec):
         vspec_str, matcher, is_exact = self.get_matcher(vspec)
         super(VersionSpec, self).__init__(vspec_str, matcher, is_exact)
 
+    @property
+    def is_prerelease(self) -> bool:
+        """True if exact match and spec is for prerelease version ending with alphabetic
+        tag other than 'post'
+        """
+        return self.is_exact() and bool(version_prerelease_re.match(self.spec_str))
```
jaimergp, line 511: What happens with the `*` glob versions here?

Also, why should it _end_  with such a suffix only? I think a version like `1.2.3.dev1.post1` should still be considered a pre-release. Maybe that's a wrong way of using `post` but I guess we can argue it's _not_ a stable release. Is it a better test if it _contains_ a non-post alphanumeric tag?

tests/models/test_version.py
```diff
@@ -324,3 +324,34 @@ def test_pep_440_arbitrary_equality_operator(self):
         # We're going to leave the not implemented for now.
         with pytest.raises(InvalidVersionSpec):
             VersionSpec("===3.3.2")
+
+    def test_prerelease_re(self):
```
jaimergp, line 328: I'd like to see tests for `*` containing versions.

tests/models/test_match_spec.py
```diff
@@ -476,6 +476,22 @@ def test_license_match(self):
         assert MatchSpec("*[license='*gpl*']").match(record)
         assert MatchSpec("*[license='*v3+']").match(record)
 
+    def test_is_prerelease(self):
```
jaimergp, line 479: I'd like to see tests for `*` containing versions.

conda/base/context.py
```diff
@@ -355,6 +356,16 @@ class Context(Configuration):
...
+    prerelease_behavior = ParameterLoader(PrimitiveParameter(PrereleaseBehavior.ALLOW,
+                                                             element_type=PrereleaseBehavior))
+    prerelease_allowed_packages = ParameterLoader(
+        SequenceParameter(
+            PrimitiveParameter("", element_type=str),
+            default=("openssl <=2.99.99",),
+            string_delimiter='&',
+        )
+    )
+
```
jaimergp, line 368: I don't think we need to expose both the raw string _and_ a spec. This should be stored as `_prerelease_allowed_packages`, with a Parameter alias `prerelease_allowed_packages`. Then, the property below can just be `prerelease_allowed_packages`. You can see this pattern in other property-post-processed keys.

jaimergp (commented):
OK, here we go! I'll start by repeating what I said above. This is very high quality contribution! Thanks for taking the time to research and understand the issue thoroughly, and then posting a high quality solution! I've learnt a lot reviewing and checking the background context needed. Thanks again!

You'll see plenty of comments, but in general the main points are:
* The `context` object might not need the `is_prelease` method there, since a check for the PackageRecord, MatchSpec or VersionOrder objects might be enough.
* I am concerned about the ramifications of the `limit` mode, which is [the default mode suggested by PEP440](https://peps.python.org/pep-0440/#handling-of-pre-releases). If we implement this, we need to decide how we are going to interpret the PEP440 directives, specially when it comes to the "allow pre-releases for packages already using a pre-release" directive, if we understand it in a package-by-package basis. This of course would need specific tests.
* How we handle `*`-containing versions.

I am also curious about what happens to projects that are _only_ available as pre-releases (for now). The most popular example is `black`. I think `pip` handles this with an exception, but it still begs the question: should a prerelease-only project be always available regardless the `prerelease=xxxx` option chosen? PEP440 seems to be addressing this in the 'limit'  mode implementation but, again, we need tests for that too.

On top of this, I checked open issues in `pypa/pip` and other tools, to see what shortcomings the community might have found before us. Please take a look at the following references and let's see if we can establish whether the current status of the PR suffers from the same issue(s):

* Intersection with other flags and/or commands? https://github.com/pypa/pip/issues/8983
* This part of the PEP440 'limit' mode: https://github.com/pypa/pip/issues/9121
* This other issue about automated opt-in via the specifier. It sounds like pip enables RCs if the specifier mentions RCs, but it shoudl only mean so for the current stable branch, not every subsequent RC? https://github.com/pypa/pip/issues/7579

Last but not least, a ramification of this feature. Allowing prereleases has effects on two levels: creating the index and solving the environment. On the index creation side, it determines the availability of the package. The solving side means whether we pick it or not. In that vein, wonder if we should extend the availability checks to `conda search`. 

Apologies for the length of the review! This is an extremely useful and important feature, and I am very grateful we are progressing in the right direction. Thanks again for the nth time :)

conda/resolve.py
analog-cbarber: No. It just needs to be a large finite number to ensure that the prerelease solutions go last.

I think that 100 is actually probably not large enough. Probably should just make it 1000000 or something.

conda/base/context.py
analog-cbarber: Sure, why not. However, in practice I would expect this to only contain a list of simple package names without any versions.

conda/base/context.py
analog-cbarber: Yes, I was a bit reluctant to put this here, but I was trying to avoid code duplication. I will consider that.

conda/models/version.py
analog-cbarber: This is really only used on the MatchSpecs that are generated from actual package versions. It isn't used on the input match specs. Thus the `is_exact` test.

I wanted a relatively simple and easy to use implementation that covers existing usage. In theory, there could be such a package version, but I did not observe anything like that when I looked through the conda-forge repodata indexes.

We could try to support stuff like that, but it is probably a waste of time to devote energy and code complexity on theoretical cases. I think we can just document the behavior and expect people to version their packages accordingly.

conda/models/records.py
analog-cbarber: ok

conda/cli/conda_argparse.py
analog-cbarber: I thought it would be clearest for this to exactly match the config parameter, especially since there is another config parameter `prerelease_allowed_packages`. The current conda argparse configuration already allows abbreviations, so users will actually be able to write `--prerelease limit` (although I have not tried this myself).

In practice, I expect that this option will not be used that often. Most organizations will want to just tell everyone to configure the option in their global or project `.condarc`.

conda/core/link.py
analog-cbarber: This is doing a per-package check. I see nothing in the PEP 440 spec that suggests that an existing pre-release version should enable an update to a different pre-release:

> By default, dependency resolution tools SHOULD:
>
> * accept already installed pre-releases for all version specifiers
> * accept remotely available pre-releases for version specifiers where there is no final or post release that satisfies the version specifier
> * exclude all other pre-releases from consideration

So I think this implementation satisfies the spec. However, I did not look to see how pip implements this.

I think it would be reasonable to allow an installed pre-release to upgrade to a new pre-release, but only if there is no intervening regular release. For instance, if you already had `foo 1.2.3.dev2` installed, it would be ok to upgrade to `foo 1.2.3.dev4` as long as there is no `foo 1.2.4` that satisfies the original match spec. I am not sure how to implement that or whether it would be worth the trouble.

analog-cbarber:
Thanks for the review. I will update the code when I get the chance.

I think it would make sense to update `conda search` as well, but I think that would only mean excluding pre-releases when the behavior is set to exclude them.

analog-cbarber:
I should also say that I think that it will be hard to really test this without using it in real world situations for a while. Obviously, we should add any useful unit tests that make sense, but it will be hard to anticipate everything.

However, since this will not change the default behavior (at least for quite a while), it may be sufficient to just clearly advertise it as a new "experimental" feature and that slight adjustments could possibly be expected. Then those of us who intend to use the new limit mode can try it out in our own organizations to see if it meets everyone's expectations.

analog-cbarber:
I am interested in working on this some more, but will not have the bandwidth for at least several weeks.

PR #11803, opened 2022-09-07
Fixes an issue that we are having with carnary builds

travishathaway:
This pull request adds the same changes that were implemented in the official conda feedstock. See this issue for more information: https://github.com/conda/conda-build/issues/4572


kenodegard (changes requested):
I don't think the `cytoolz` change is necessary for this fix

jezdez (approved):

kenodegard (approved):

Issue #11608, opened 2022-07-08
Using asyncio/threads to download packages


shughes-uk:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other speed issues and didn't find a duplicate

### What happened?

This is a continuation of an email thread with @LtDan33 and @pzwang 

I wrote a hacky little experiment that pre-populates the `pkgs` cache using asyncio to download packages listed in a conda-lock lockfile. I found it to be much faster than the default conda implementation. I suspect it would be faster than using multiple threads too. 

Mostly looking to optimize start time for dask clusters, I can create the environment and wrap it up as a docker image, but it ends up being slower than doing this.

In benchmarking I found this to be as fast or faster than mamba's implementation. Definitely much faster when mamba falls back to what I think is conda's single threaded implementation!

My very hacky use of a processpool executor to farm out validation/extraction of the package tarballs probably needs some extra thought, it was quite fast though!

```python
import asyncio
import hashlib
from concurrent.futures import ProcessPoolExecutor
from functools import partial
from pathlib import Path
from typing import Dict
from urllib.parse import urlparse

import aiofiles
import aiohttp
import yaml
from conda_package_handling import api

cheap = yaml.load(Path("./conda-lock.yml").open(), Loader=yaml.SafeLoader)
root = Path("~/miniconda3/pkgs").expanduser()
root.mkdir(exist_ok=True)


def extract(sha256: str, file: str, target: str):
    hasher = hashlib.new("sha256")
    with open(file, "rb") as fh:
        for chunk in iter(partial(fh.read, 8192), b""):
            hasher.update(chunk)
    assert hasher.hexdigest() == sha256
    print("Extracting", file, target)
    api.extract(file, target)


async def main(packages: Dict):
    async with aiohttp.ClientSession() as session:
        tasks = []
        urls = []
        for package in packages:
            if package["platform"] == "linux-aarch64" and package["manager"] == "conda":
                url = package["url"]
                parsed = urlparse(url)
                pth = Path(parsed.path)
                pth.name
                tasks.append(
                    download_url(
                        sha256=package["hash"]["sha256"],
                        session=session,
                        fp=root / pth.name,
                        url=package["url"],
                    )
                )
                urls.append(url)
        await asyncio.gather(*tasks)
        async with aiofiles.open(root / Path("urls.txt").expanduser(), mode="w") as f:
            for url in urls:
                await f.write(url + "\n")
[...]

PR #11807, opened 2022-09-08
download and extract packages in a thread pool

dholth:
### Description

Conda download and extract is aggressively single-threaded. Make it parallel. For #11608

See also https://github.com/dholth/conda-benchmarks/blob/main/benchmarks/conda_install.py#L42

Requires `pip install git+https://github.com/conda-incubator/conda-package-streaming#egg=conda-package-streaming` (released version does not include extract API); older libarchive-based extractor appears to be not threadsafe? (it uses `os.chdir()` for example)

The tests are difficult, it is very easy to accidentally time "packages already cached" as conda likes to look in a number of cache directories.

### Checklist - did you ...

### DIFF
--- a/conda/core/package_cache_data.py
+++ b/conda/core/package_cache_data.py
@@ -13,6 +13,12 @@ from os.path import basename, dirname, getsize, join
 from sys import platform
 from tarfile import ReadError
 
+from concurrent.futures import as_completed
+from concurrent.futures.thread import ThreadPoolExecutor
+
+DOWNLOAD_THREADS = 10
+EXTRACT_THREADS = None  # ThreadPoolExecutor will set cpu count or max 32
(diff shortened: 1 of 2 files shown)


shughes-uk:
Having the number of download threads be either configurable or default to cpu count would be appealing. I work with machines that have 300gbit network access and would love to be able to utilize all of it!

dholth:
If you can get the asv benchmark to run, maybe you can try it with bigger
thread counts. However the # of connections can't (easily :)) be > the
number of packages!

dholth:
https://github.com/conda/conda/pull/11841 downloads and extracts in separate thread pools, and they are configurable with a `CONDA_FETCH_THREADS=` setting

PR #11812, opened 2022-09-08
Use fd 3 instead of hijacking stdout

### LINKED ISSUE #11748: `conda shell.posix activate/deactivate` should write out to `fd` 3, not `stdout`

kenodegard:
### Description

After starting with #11748 we found that we didn't like how we were hijacking stdout/stderr. So instead, let's just use file descriptor 3 that way we can still depend on stdout/stderr for their correct purposes.

ref: [A simple producer/consumer example](https://unix.stackexchange.com/a/537435)
ref: [A simple Python example](https://stackoverflow.com/a/61536168)
ref: [Is it safe to use fd 3](https://unix.stackexchange.com/a/320588)

This is still in POC territory, still missing the fix for:
- [ ] (t)csh
- [ ] cmd.exe
- [ ] PowerShell
- [ ] fish
- [ ] xonsh

Resolves #11748 
Depends on #11862 
Depends on #11865 

### Checklist - did you ...

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -108,7 +108,17 @@ def main_sourced(shell, *args, **kwargs):
         raise CondaError("%s is not a supported shell." % shell)
 
     activator = activator_cls(args)
-    print(activator.execute(), end="")
+    script = activator.execute()
+
+    try:
+        from os import fdopen
+
+        with fdopen(3, "w") as fh:
+            fh.write(script)
+    except OSError:
+        # the file descriptor 3 hasn't been opened so it cannot be written to, just write out
+        # to stdout instead as the user is likely running this for debugging purposes
+        print(script, end="")
     return 0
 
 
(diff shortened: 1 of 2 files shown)


kenodegard:
this was a valuable learning and research effort but is unlikely to proceed in the current form

PR #11835, opened 2022-09-12
add ThreadPoolExecutor to package_cache_data

dholth:
### Description

A closer solution for #11608

So far, a clumsy translation of a clumsy initial process. But the asv benchmark shows the right speedup.

### Checklist - did you ...


dholth:
It would make sense to essentially rename repodata threads to fetch_threads (or have it follow fetch_threads), since they are both about concurrent http transfers

PR #11836, opened 2022-09-12
Move `get_plugin_manager` to `conda.plugins`

kenodegard:
### Description

An alternative solution to #11785. Keeps all plugin/pluggy related content within the plugin submodule.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -5,14 +5,12 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from collections import OrderedDict
 
-import functools
 from errno import ENOENT
 from functools import lru_cache
 from logging import getLogger
 import os
 from os.path import abspath, basename, expanduser, isdir, isfile, join, split as path_split
 import platform
-import pluggy
 import sys
 import struct
 from contextlib import contextmanager
@@ -63,8 +61,6 @@ from ..common.url import has_scheme, path_to_url, split_scheme_auth_token
 from ..common.decorators import env_override
 
 from .. import CONDA_SOURCE_ROOT
-
-from .. import plugins
 
 try:
     os.getcwd()
@@ -153,14 +149,6 @@ def ssl_verify_validation(value):
                     "certificate bundle file, or a path to a directory containing "
                     "certificates of trusted CAs." % value)
     return True
-
-
-@functools.lru_cache(maxsize=None)  # FUTURE: Python 3.9+, replace w/ functools.cache
-def get_plugin_manager():
-    pm = pluggy.PluginManager('conda')
-    pm.add_hookspecs(plugins)
-    pm.load_setuptools_entrypoints('conda')
-    return pm
 
 
 class Context(Configuration):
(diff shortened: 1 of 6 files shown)


conda/base/context.py
```diff
@@ -398,14 +386,6 @@ def __init__(self, search_path=None, argparse_args=None):
         super(Context, self).__init__(search_path=search_path, app_name=APP_NAME,
                                       argparse_args=argparse_args)
 
-        # Add plugin support
-        self._plugin_manager = get_plugin_manager()
-
-    @property
-    @functools.lru_cache(maxsize=None)  # FUTURE: Python 3.9+, replace w/ functools.cache
-    def plugin_manager(self):
-        return self._plugin_manager
```
jezdez, line 407: The reason we considered putting it into the context object, is that it's at the core of the conda startup process and is appropriate for loading user-specific configuration like the plugins.

conda/plugins/__init__.py
```diff
@@ -5,34 +5,12 @@
...
-@_hookspec
-def conda_subcommands() -> Iterable[CondaSubcommand]:
-    """
-    Register external subcommands in conda.
-
-    :return: An iterable of subcommand entries.
-    """
+manager = pluggy.PluginManager("conda")
+manager.add_hookspecs(specs)
+manager.load_setuptools_entrypoints("conda")
```
jezdez, line 16: FWIW we wanted to make sure that no import-time side effect would trigger the plugin discovery, since it may have unintended downsides to how plugins are loaded. That's why this wasn't in the `conda.plugins` module itself but hidden in a function when the context is loaded during startup. So the way this is written is a risk to add complications when plugins do something weird in their entrypoints or similar.

conda/plugins/specs.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
jezdez, line 1: I like the idea of putting the specs and plugin objects into a separate file, I think this hadn't been done because it wasn't clear how we'll evolve the code. I wonder how quickly this would become unwieldy, the more plugin hooks we support, though. Maybe it would be best to separate them by use case, e.g. put the subcommand specific code into `conda.plugins.subcommands`? Will the use of the @_hookspec decorator enough to trigger the registration of the plugin hook spec then?

kenodegard: I only did this because `pluggy.PluginManager.add_hookspecs` expects a namespace, either a module or a class. AFAIK we need both the `@_hookspec` and the `pluggy.PluginManager.add_hookspecs` for the specs to be registered.

I wasn't super happy with using `conda.plugins.specs`, not sure about the directory structure, perhaps we should remove this specs submodule until we have more types of plugins and a better feel for how we want this laid out?

conda/base/context.py
kenodegard: True, but I'm starting to suspect that we are misusing the context object, isn't it an anti-pattern to use a large stateful singleton? Moreover, our context object containing singletons feels like a no-no.

Keeping all of this within `conda.plugins` should also help with discoverability and is a step towards reorganizing conda into "topics" instead of "levels".

Issue #6902, opened 2018-02-20
Override condarc path


pitrou:
**I'm submitting a...**
  - [ ] bug report
  - [x] feature request

## Current Behavior

When `conda` is run, it will unconditionally pick up the current user config. This is detrimental in some cases, for example when conda is run programmatically by some integration tests.

## Expected Behavior

One should be able either 1) either to override the conda config lookup path (for example using a `CONDARC=/path/to/my/condarc` environment variable) or 2) to disable config reading entirely (for example using a `CONDA_IGNORE_CONFIG=1` environment variable).

Note that option 1 can emulate option 2 if it is allowed to specify `CONDARC=/dev/null` or `CONDARC=/somenonexistentpath`.

goanpeca:
Should this also include a CLI argument?

pitrou:
For the use case I'm thinking of, an environment variable is enough.

kalefranz:
This issue juxtaposes sysadmins needing to lock down configuration via read-only config files (e.g. `/etc/conda/condarc`) against general user-friendliness and flexibility.  We should definitely strive for the second, but we also need to make sure we don't violate the security model here.

mbargull:
@pitrou, for context, `conda` uses a couple of different paths to load the configuration from:
https://github.com/conda/conda/blob/4.4.10/conda/base/constants.py#L25-L43
```python
SEARCH_PATH = (
    '/etc/conda/.condarc',
    '/etc/conda/condarc',
    '/etc/conda/condarc.d/',
    '/var/lib/conda/.condarc',
    '/var/lib/conda/condarc',
    '/var/lib/conda/condarc.d/',
    '$CONDA_ROOT/.condarc',
    '$CONDA_ROOT/condarc',
    '$CONDA_ROOT/condarc.d/',
    '~/.conda/.condarc',
    '~/.conda/condarc',
    '~/.conda/condarc.d/',
    '~/.condarc',
    '$CONDA_PREFIX/.condarc',
    '$CONDA_PREFIX/condarc',
    '$CONDA_PREFIX/condarc.d/',
    '$CONDARC',
)
```
I guess you are most interested in (only?) overriding the user-specific paths? I.e.,
```python
    '~/.conda/.condarc',
    '~/.conda/condarc',
    '~/.conda/condarc.d/',
    '~/.condarc',
```
Also note that there already is a `$CONDARC`, though this one is (like the other files) additive and doesn't disable the other configuration files.

pitrou:
[...]

PR #11838, opened 2022-09-13
Provide a way to override condarc search path: `CONDA_NO_CONFIG_SEARCH_PATH`

jaimergp:
### Description

Some uses of `conda` require working in absolute isolation from user configurations. One example is conda-standalone as used in `constructor`.

Some issues originating from this problem can be found here: https://github.com/conda/constructor/issues/568. 

The solution is simple: in the presence of a non-empty `NO_CONDARC` environment variable, the default search path is overridden and thus no configuration is loaded from disk, at all. 

### Checklist - did you ...

Closes https://github.com/conda/constructor/issues/542

Questions: 

* An alternative solution to this problem would be exposing `CONDA_SEARCH_PATH` as a configurable list. Maybe not linked directly in the context object (bootstrapping issue). Instead of using `NO_CONDARC`, the user could set `CONDA_SEARCH_PATH=""`, achieving the same effect. This would be more flexible but also confusing and unneeded?

### DIFF
--- a/tests/base/test_context.py
+++ b/tests/base/test_context.py
@@ -564,13 +564,13 @@ def test_condarc_search_path_override(tmpdir):
         reset_context()
         assert "uniquely_named_channel" in context.channels
 
-        # $CONDA_RC_SEARCH_PATH overrides the default search path, but
+        # $CONDA_CONFIG_SEARCH_PATH overrides the default search path,
         # but we can still include CONDARC in that path :)
-        with env_var("CONDA_RC_SEARCH_PATH", "$CONDARC"):
+        with env_var("CONDA_CONFIG_SEARCH_PATH", "$CONDARC"):
             reset_context()
             assert "uniquely_named_channel" in context.channels
 
         # If we set it to the empty string, then no file will be loaded
-        with env_var("CONDA_RC_SEARCH_PATH", ""):
+        with env_var("CONDA_CONFIG_SEARCH_PATH", ""):
             reset_context()
             assert "uniquely_named_channel" not in context.channels
--- a/tests/base/test_context.py
+++ b/tests/base/test_context.py
@@ -547,3 +547,24 @@ def test_validate_prefix_name(prefix, allow_base, mock_return_values, expected):
         else:
             actual = validate_prefix_name(prefix, ctx, allow_base=allow_base)
             assert actual == str(expected)
+
+
+def test_condarc_search_path_override(tmpdir):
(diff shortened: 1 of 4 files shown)


chenghlee:
Would setting this override the system-/install-level configuration files (i.e., those in `$CONDA_ROOT`)? I'm know sysadmins who use those configuration files to, e.g., constrain what channels users can access, and it's probably not a good idea to give users a mechanism to override that.

kenodegard:
How about `CONDA_CONFIG_SEARCH_PATH`? This would perhaps be more clear to those unfamiliar with the history of "rc"?

Something to consider, git offers a [`GIT_CONFIG_NOSYSTEM` boolean environment variable](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables). Not sure how much value there is in exposing the `SEARCH_PATH` versus a boolean.

tests/base/test_context.py
```diff
@@ -547,3 +547,30 @@ def test_validate_prefix_name(prefix, allow_base, mock_return_values, expected):
...
+                channels: ["uniquely_named_channel"]
+                """
+            )
+        )
+    # $CONDARC allows us to add one more file or directory to the search path
+    with env_var("CONDARC", custom_condarc):
+        reset_context()
+        assert "uniquely_named_channel" in context.channels
+
+        # $CONDA_RC_SEARCH_PATH overrides the default search path, but
```
pavelzw, line 567:
```suggestion
        # $CONDA_RC_SEARCH_PATH overrides the default search path,
```

jaimergp:
Hm, I keep thinking about this and overriding the default search path with full flexibility might incur into a support team asking "have you changed the default search path? what is the value?", which is something I don't think it's needed.

So... inspired by your suggestion, @kenodegard, the last approach is this one:
* Users can set `CONDA_NO_CONFIG_SEARCH_PATH` to change the default search path to just `$CONDARC`.
* If they _also_ set `CONDARC` to a file or directory, this will be loaded.
* With no `CONDARC` env var, no file is loaded.

This allows people to override the default condarc, if needed, with a custom one, but also prevents issues coming from misconfiguring `CONDA_SEARCH_PATH`.

Does this look like a compromise?

pavelzw:
Sounds good to me 👍🏻

jaimergp:
Turns out this has been attempted before. See prior discussion: https://github.com/conda/conda/issues/6902.

There were concerns about the "security model" for sysadmins.

This might live as a perma-patch on conda-standalone for constructor purposes.

PR #11841, opened 2022-09-13
Download packages in parallel

dholth:
### Description

Download packages in parallel.

Anticipates thread-safe conda extract()

For #11608.

- [x] Progress bars should not write on top of each other
- [x] Fetch threads defaults to `None = (# of cores or 32 max)`; should default to e.g. 10 (which is a default maximum connections per server in requests)

### Checklist - did you ...


dholth:
Fetch threads should probably be 10 and extract threads 4 (informed by asv benchmark)

dholth:
The relevant benchmarks is https://dholth.github.io/conda-benchmarks/#conda_install.TimeInstall.time_explicit_install?branch=parallel-download&branch=main&x-axis=threads

When the latency is 0, the main branch is about as fast as this branch. When the latency increases, threading helps tremendously because we can wait for the per-package delays in parallel.

(We can't [extract in threads yet,](https://dholth.github.io/conda-benchmarks/#extract.TimeExtract.time_extract?branch=main&x-axis=threads) since libarchive is not threadsafe.)

dholth:
https://asciinema.org/a/7hf1BkGd8ZdUsXW8qx85Wl4Zt

FFY00:
Hey, any ETA on merging this? This conflicts with #11600, and I think it's easier to just implement the plugin hook on top of this directly instead of resolving conflicts later.

jezdez (approved):
Just a few questions.. LGTM otherwise.

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+        if action:
+            action.reverse()
+
+
+def done_callback(f, actions, progress_bar: ProgressBar, exceptions: list, finish=False):
+    try:
+        f.result()
+    except Exception as e:
+        do_reverse(progress_bar, *reversed(actions))
+        exceptions.append(e)
```
jezdez, line 876:
```suggestion
def done_callback(future, actions, progress_bar: ProgressBar, exceptions: list, finish=False):
    try:
        future.result()
    except Exception as e:
        do_reverse(progress_bar, *reversed(actions))
        exceptions.append(e)
```

conda/core/package_cache_data.py
```diff
@@ -4,14 +4,19 @@
...
 from collections import defaultdict
+from concurrent.futures import as_completed
 from errno import EACCES, ENOENT, EPERM, EROFS
 from json import JSONDecodeError
 from logging import getLogger
 from os import scandir
 from os.path import basename, dirname, getsize, join
 from sys import platform
 from tarfile import ReadError
+from concurrent.futures.thread import ThreadPoolExecutor as Executor
```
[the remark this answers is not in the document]
dholth: We should add a get_executor / executor factory that chooses the right one (or a DummyExecutor when threads are disabled). However I need to double check that the DummyExecutor supports all the features I use.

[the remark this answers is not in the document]
dholth: Used original name.

conda/core/package_cache_data.py
```diff
@@ -644,83 +713,109 @@ def execute(self):
             # TODO: use logger
```
travishathaway, line 713: It would be really nice to have a doc string in this method explaining what is going on at a high level.

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
         return hash(self) == hash(other)
 
 
+# called from ProgressiveFetchExtract.execute()
+def do_cache_action(prec, cache_axn, progress_bar, download_total=1.0):
+    # dummy action to simplify code
```
travishathaway, line 829:
```suggestion
def do_cache_action(prec, cache_axn, progress_bar, download_total=1.0):
    """
    This is function gets called from `ProgressiveFetchExtract.execute`
    """
    # dummy action to simplify code
```

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+    if not cache_axn:
```
travishathaway, line 830: cache_action? Not sure what `axn` means. Let's remove that ambiguity.

conda/core/package_cache_data.py
travishathaway: Please feel free to add whatever else you think would be important as a doc string. Perhaps even mention that it is getting called from within a `ThreadPoolExecutor` (right?)

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+
+    else:
+        download_total = 0
+        progress_update_cache_axn = None
+
+    cache_axn.execute(progress_update_cache_axn)
+    return prec
+
+
+def do_extract_action(prec, extract_axn, progress_bar):
```
travishathaway, line 847: This function signature leaves me wondering what the arguments are. Is this a place where type hinting could help? Also a doc string here would be very nice 🙏

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+        return prec
+    extract_axn.verify()
+    # currently unable to do updates on extract;
+    # likely too fast to bother
+    extract_axn.execute(None)
+    progress_bar.update_to(1.0)
+    return prec
+
+
+def do_cleanup(progress_bar, *actions):
```
travishathaway, line 859: Looks like progress bar is an used argument. Could we remove it?

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+    for action in actions:
+        if action:
+            action.cleanup()
+
+
+def do_reverse(progress_bar, *actions):
```
travishathaway, line 865: `progress_bar` is also unused here.

conda/core/package_cache_data.py
```diff
@@ -729,10 +824,67 @@ def __eq__(self, other):
...
+    for action in actions:
+        if action:
+            action.reverse()
+
+
+def done_callback(future, actions, progress_bar: ProgressBar, exceptions: list, finish=False):
```
travishathaway, line 871:
```suggestion
def done_callback(future: Future, actions: tuple[Union[CacheUrlAction, ExtractPackageAction], ...], progress_bar: ProgressBar, exceptions: list[Exception], finish: bool = False):
```

Might as well go all the way if you're going to use type hints 😅 . If you decide to incorporate this change, be sure to add the following to the top of the file:

```python

from typing import Union
```

travishathaway: Also, this function would benefit from a doc string too 🙏

travishathaway (changes requested):
Looking good so far, I would just like to see some doc strings added and we could also improve the readability with some better usage of type hints. There were also a couple of functions with unused arguments that need to be updated.

conda/core/package_cache_data.py
travishathaway: Also, I that `cache_axn` variable should be spelled out `cache_action`. I know it's used else where in the code and you're just following convention, but it's a bad one that we should change 😄

conda/core/package_cache_data.py
travishathaway: Realizing now that it's used else where in the code, but let's not follow a bad convention!

tests/core/test_package_cache_data.py
```diff
@@ -7,86 +7,97 @@
...
     url="%s/%s/%s" % (CONDA_PKG_REPO, subdir, zlib_conda_fn),
 )
 
+
 def test_ProgressiveFetchExtract_prefers_conda_v2_format():
     # force this to False, because otherwise tests fail when run with old conda-build
-    with env_var('CONDA_USE_ONLY_TAR_BZ2', False, stack_callback=conda_tests_ctxt_mgmt_def_pol):
+    # zlib is available in local "linux-64" subdir
+    with env_vars(
+        {"CONDA_USE_ONLY_TAR_BZ2": False, "CONDA_SUBDIR": "linux-64"},
```
dholth, line 83: It looks like this repository has plenty of subdirs but not osx-arm64

travishathaway (approved):

imxtx:
One drawback: The downloading process is not shown which is very misleading. I thought Conda was stuck.

PR #11752, opened 2022-08-23
update dev/start for arm64

dholth:
### Description

Add architecture to download URL for development scrip to benefit arm64; fix #11751

PR #11851, opened 2022-09-16
Expose --mach and --arch to devenv script

kenodegard:
### Description

We added Mac ARM support to the devenv script (https://github.com/conda/conda/pull/11752) but occasionally still find it desirable to use the x86_64 installer.

### Checklist - did you ...

### DIFF
--- a/dev/start
+++ b/dev/start
@@ -12,6 +12,7 @@ cleanup() {
     unset _BASEEXE
     unset _CMD
     unset _DEFAULT_ARCH
+    unset _DEFAULT_DEVENV
     unset _DEFAULT_DRYRUN
     unset _DEFAULT_MACH
     unset _DEFAULT_PYTHON
--- a/dev/start
+++ b/dev/start
@@ -11,35 +11,44 @@ cleanup() {
     unset _AUTOBASE
     unset _BASEEXE
     unset _CMD
+    unset _DEFAULT_ARCH
+    unset _DEFAULT_DRYRUN
+    unset _DEFAULT_MACH
+    unset _DEFAULT_PYTHON
+    unset _DEFAULT_UPDATE
     unset _DEVENV
     unset _DRYRUN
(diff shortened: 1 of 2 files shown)


dev/start
```diff
@@ -67,19 +94,21 @@ while [ $# -gt 0 ]; do
...
         -h|--help)
             # since zsh is MacOS standard fallback to $0 if not in bash
             echo "Usage: source ${BASH_SOURCE:-$0} [options]"
             echo ""
             echo "Options:"
-            echo "  -p, --python  VERSION  Python version for the env to activate. (default: 3.8)"
-            echo "  -u, --update           Force update packages. (default: update every 24 hours)"
+            echo "  -p, --python  VERSION  Python version for the env to activate. (default: ${_DEFAULT_PYTHON})"
+            echo "      --mach    MACHINE  Python version for the env to activate. (default: ${_DEFAULT_MACH})"
+            echo "      --arch    ARCH     Python version for the env to activate. (default: ${_DEFAULT_ARCH})"
```
jezdez, line 107: Just out of curiosity, is there a reason not to allow `-m` and `-a` as well?

jezdez (approved):
Just a minor question about short option names, thank you!

dev/start
kenodegard: social engineering, I tend to only expose the short options for those that I want users to use frequently, if there's only a long option I interpret it as "think again" and "read the docs before proceeding"

jezdez: Hmm, I get that, it's a common pattern in end user CLIs, but given that `dev/start` is hardly a tool used by end users, I think it's fair to set them up for ease of use instead of adding a barrier. That said, that's just my thoughts, not really strongly opinionated about it.

beeankha (dismissed):
👍

beeankha (approved):
🚀

Issue #11847, opened 2022-09-15
`conda env update` fails when notices are enabled and the base environment's python is being updated


travishathaway:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

This was first reported in this [mamba](https://github.com/mamba-org/mamba) issue:

- https://github.com/mamba-org/mamba/issues/1925

The bug appears when updating `conda`'s (or `mamba`'s) root/base environment's Python version. We have been able to successfully recreate this in linux Docker containers.

### How to recreate

Using this Dockerfile (or whatever other Dockerfile that has the latest `conda` installed)

```Dockerfile
FROM --platform=linux/amd64 debian:stable-slim

RUN apt update -y && apt upgrade -y
RUN apt install curl -y
RUN cd /tmp/ && curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
RUN bash /tmp/Miniconda3-latest-Linux-x86_64.sh -p /opt/conda -b
RUN /opt/conda/bin/conda init --all
RUN /opt/conda/bin/conda install -y -c defaults python conda-build
```

Build and run this docker image...

Once on the command line:

```bash
$ python --version  # let's see which python version we have
Python 3.7.13
$ echo "name: base
channels:
  - defaults
dependencies:
  - python=3.10" > /tmp/environment.yml  # create an environment.yml file with a newer version of Python
$ conda env update -f /tmp/environment.yml
💥  # same error pops up
```

The following command will also trigger this error (do this instead of running `conda env update`)

```bash
$ conda install python=3.10
```

### How to fix

We should look at how `CondaSession` and the `requests` library handles loading TLS root certificates. There's a chance we may be able to load these ahead of time in memory to avoid these situations.

Experimentation for a better solution is encouraged as well.

Issue #11743, opened 2022-08-22
Conda install fails with: Invalid URL 'notices.json'


nejcskofic:
### What happened?

Running command `conda install --quiet "company.h5-base>=3.0,<4.0" -c bioconda -c conda-forge -c s3://company-conda-channels/dev -c defaults` fails in retrieving notices step:

<details>
<summary>Error Report</summary>

```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
Retrieving notices: ...working... failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/exceptions.py", line 1125, in __call__
        return func(*args, **kwargs)
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/cli/main.py", line 86, in main_subshell
        exit_code = do_call(args, p)
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 93, in do_call
        return getattr(module, func_name)(args, parser)
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/notices/core.py", line 75, in wrapper
        display_notices(
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/notices/core.py", line 39, in display_notices
        channel_notice_responses = http.get_notice_responses(channel_name_urls, silent=silent)
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/notices/http.py", line 36, in get_notice_responses
        return tuple(
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/notices/http.py", line 39, in <genexpr>
        (
      File "/opt/miniconda3/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
        yield fs.pop().result()
      File "/opt/miniconda3/lib/python3.8/concurrent/futures/_base.py", line 437, in result
        return self.__get_result()
      File "/opt/miniconda3/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
        raise self._exception
      File "/opt/miniconda3/lib/python3.8/concurrent/futures/thread.py", line 57, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/opt/miniconda3/lib/python3.8/site-packages/conda/notices/http.py", line 42, in <lambda>
[...]

PR #11852, opened 2022-09-16
Fix for conda update and install issue related to notices

travishathaway:
This pull request is an attempt to fix the bug reported in the following issue:

- https://github.com/conda/conda/issues/11847


travishathaway:
This pull request should also resolve the following issue:
- https://github.com/conda/conda/issues/11743

conda/notices/core.py
```diff
@@ -48,15 +48,29 @@ def display_notices(
...
     channel_notices = filter_notices(channel_notices, limit=limit, exclude=viewed_notices)
-    if len(channel_notices) == 0:
-        return
 
-    views.print_notices(channel_notices)
+    return ChannelNoticeSet(
+        channel_notices=channel_notices,
+        viewed_channel_notices=num_viewed_notices,
+        total_number_channel_notices=num_total_notices,
+    )
```
jezdez, line 56: Could we align the variable names here with the attribute names of ChannelNoticeSet?

conda/notices/types.py
```diff
@@ -24,6 +24,28 @@ class ChannelNotice(NamedTuple):
     interval: Optional[int]
 
 
+class ChannelNoticeSet(NamedTuple):
```
jezdez, line 27: The name is a little misleading since it's not a Python set after all, could we use this instead?

```suggestion
class ChannelNotices(NamedTuple):
```

conda/notices/core.py
```diff
@@ -66,19 +80,21 @@ def notices(func):
...
-
         if is_channel_notices_enabled(context):
-            display_notices(
+            channel_notice_set = retrieve_notices(
                 limit=context.number_channel_notices,
                 always_show_viewed=False,
                 silent=True,
             )
+            return_value = func(*args, **kwargs)
+            display_notices(channel_notice_set)
```
jezdez, line 92: Let's add a comment here to explain that we first retrieve the notices before calling the decorated function, since it may have side-effects of not doing that.

conda/__init__.py
```diff
@@ -148,3 +149,7 @@ def _default(self, obj):
 
 _default.default = JSONEncoder().default
 JSONEncoder.default = _default
+
+# Add s3:// support to urllib.parse
+urllib.parse.uses_netloc.append("s3")
+urllib.parse.uses_relative.append("s3")
```
jezdez, line 155: Can you elaborate why this is needed?

jezdez (approved):
Just minor naming suggestions and comment requests. Nice work!

kenodegard (changes requested):

conda/__init__.py
travishathaway: @jezdez, are you talk about in the code or here?

This is need so that we can use `urllib.parse.urljoin` with s3. I will leave that in the code as well.

conda/notices/types.py
travishathaway: My only concern is that the name is too similar to the above data type `ChannelNotice`.

In defining this, I was trying to mirror what Django does with its `QuerySet` object. For notices, this is the object getting returned from our `retrieve_notices` function, so it is in a sense a set of query results. How about, `ChannelNoticeResultSet` to be more clear and avoid confusion?

conda/__init__.py
dholth: I had to look this up as well by reading the Python source code. Changing global undocumented Python state is ugly.

travishathaway: Yeah, this definitely isn't ideal 😭 (I felt gross writing it). Maybe there's a better way to handle url joins for s3 urls? 🤔 I'll take a look at this....

travishathaway: @dholth, I've implemented an alternative approach. Please see the most recent commit.

conda/notices/core.py
```diff
@@ -35,69 +36,111 @@ def display_notices(
...
+
+    for channel in channels:
+        name = channel.name or channel.location
+
+        if type(channel) is Channel:
+            channel_name_and_urls.append((join_url(channel.base_url, NOTICES_FN), name))
+
+        elif type(channel) is MultiChannel:
+            for url in channel.base_urls:
+                channel_name_and_urls.append((join_url(url, NOTICES_FN), name))
```
jezdez, line 138: I thought `isinstance` checks are still sensible?

```suggestion
        if isinstance(channel, Channel):
            channel_name_and_urls.append((join_url(channel.base_url, NOTICES_FN), name))

        elif isinstance(channel, MultiChannel):
            for url in channel.base_urls:
                channel_name_and_urls.append((join_url(url, NOTICES_FN), name))
```

jezdez (approved):
Just a minor question about replying `type()` with `isinstance()`. Nicely done!

conda/notices/core.py
travishathaway: That's what I thought at first too, but with the changes you proposed, you will always go to the first block because `MultiChannel` is a sub-class of `Channel`. That's why I put the more strict "type" checking in there.

I think I have a better idea for how to do this though now that I'm seeing this code again:
```suggestion
        for url in channel.base_urls:
            channel_name_and_urls.append((join_url(url, NOTICES_FN), name))
```

Both classes define their own `base_urls` property that returns a tuple.

jezdez: Ah in that case you can also change the order of the if statements to first check for `MultiChannel` and then `Channel`

jezdez (approved):

conda/notices/core.py
```diff
@@ -35,69 +36,104 @@ def display_notices(
...
+        channel_notices=channel_notices,
+        viewed_channel_notices=viewed_channel_notices,
+        total_number_channel_notices=total_number_channel_notices,
+    )
+
+
+def display_notices(channel_notice_set: ChannelNoticeResultSet) -> None:
+    """
+    Prints the channel notices to std out
+    """
```
kenodegard, line 67: If we offer a default, we can avoid the change in `main_notices.py`:
```suggestion
def display_notices(channel_notice_set: ChannelNoticeResultSet | None = None) -> None:
    """
    Prints the channel notices to std out
    """
    if channel_notice_set is None:
        channel_notice_set = retrieve_notices()
```

conda/notices/core.py
```diff
@@ -35,69 +36,104 @@ def display_notices(
...
-        for channel in channels
-        for base_url in channel.base_urls
-    )
+    channel_name_and_urls = []
+
+    for channel in channels:
+        name = channel.name or channel.location
+
+        for url in channel.base_urls:
+            channel_name_and_urls.append((join_url(url, NOTICES_FN), name))
```
kenodegard, line 134: We can add the s3 protocol such that `urllib.parse.urljoin` will work:

```
urllib.parse.uses_relative.append("s3")
urllib.parse.uses_netloc.append("s3")
```

Alternatively, I'd suggest we inline the helper function:

```suggestion
            channel_name_and_urls.append((f"{url.rstrip('/')}/{NOTICES_FN}", name))
```

conda/notices/core.py
```diff
@@ -35,69 +36,104 @@ def display_notices(
...
 
 def get_channel_name_and_urls(
-    channels: [Sequence[Channel]],
-) -> Sequence[Tuple[ChannelUrl, ChannelName]]:
-    """Return a sequence of Channel URL and name"""
-
-    def ensure_endswith(value: str, ends: str) -> str:
-        return value if value.endswith(ends) else f"{value}{ends}"
+    channels: Sequence[Union[Channel, MultiChannel]],
+) -> Sequence[tuple[ChannelUrl, ChannelName]]:
```
kenodegard, line 116: Good practice for the function input types to be generic but for the output type to be explicit:

```suggestion
) -> list[tuple[ChannelUrl, ChannelName]]:
```

tests/cli/test_main_notices.py
```diff
@@ -214,3 +231,66 @@ def test_cache_names_appear_as_expected(
...
+
+    with mock.patch(
+        "conda.notices.fetch.get_notice_responses", wraps=fetch.get_notice_responses
+    ) as fetch_mock:
+        # Second run of install; notices should not be retrieved
+        run(
+            f"conda install -n {env_one} -c local --override-channels -y pre_link_messages_package"
+        )
+
+        assert len(fetch_mock.mock_calls) == 0
```
kenodegard, line 278:
```suggestion
        assert fetch_mock.assert_not_called()
```

tests/cli/test_main_notices.py
```diff
@@ -214,3 +231,66 @@ def test_cache_names_appear_as_expected(
...
+        # "wraps" probably makes this more nested than it normally would be, but the docs
+        # were not clear on that.
+        ((url, name), *_), *_ = args
+
+        assert url.startswith("file://")
+        assert name == "local"
+
+    with mock.patch(
+        "conda.notices.fetch.get_notice_responses", wraps=fetch.get_notice_responses
+    ) as fetch_mock:
```
kenodegard, line 272: We can combine this with the prior context and just reset the mock wrapper:

```suggestion
        fetch_mock.reset_mock()
```

kenodegard (changes requested):

conda/notices/core.py
travishathaway: I choose to get rid of the urllib.parse modification because it felt a little hacky. You don't think so?

kenodegard: Ah, I was trying to figure out where I'd seen this before, in that case, I'd just go with inlining this

travishathaway: Will do. It's not going to be as inline as you have it because I really don't like having two levels of quotes (one with `"` and the other with `'`).

conda/notices/core.py
travishathaway: I understand the desire for this change, but for me, this violates the, "separation of concerns" and "loose coupling" principles that I am trying to follow. A display function should not also perform a retrieval. It should merely define an interface for accepting a data structure that you would like displayed.

I hope you understand where I am coming from.

jezdez: +1 to @travishathaway I think that composition patterns like this are more effective than subtle defaults that make code harder to read.

jezdez (approved):

kenodegard (approved):

Issue #10131, opened 2020-08-05
Plugin mechanism for virtual packages


jakirkham:
Virtual packages have been popular since their introduction with `__cuda` ( https://github.com/conda/conda/pull/8267 ). Since then we have added `__glibc` ( https://github.com/conda/conda/pull/9358 ), `__osx` ( https://github.com/conda/conda/pull/9349 ), and are now looking at architecture specific ones ( https://github.com/conda/conda/pull/9930 ). As the use cases here expand, wonder if we should begin discussing a plugin architecture for the community to supply virtual packages and how these can be enabled and which ones we include by default. Thinking about this in the context of maintenance, extensibility, ease of use, etc. Thoughts? :slightly_smiling_face:

┆Issue is synchronized with this [Jira Task](https://anaconda.atlassian.net/browse/CONDAORG-144)

chenghlee:
I think it's a great idea. I can definitely see the utility, especially for community-supported packages that relies on some platform/host feature that isn't common enough to pull into the conda core code base (e.g., anything that relies on certain hardware or OS features being present).

jakirkham:
cc @kkraus14

jezdez:
I'm in full support for this feature request, and would like to suggest to use the plugin mechanism that is used by pytest, datasette and tox (and others): https://pluggy.readthedocs.io/

jakirkham:
To add something else here, it would be good to provide a way to override these virtual package plugins

jezdez:
@jakirkham I think that's the idea, yup!

hmaarrfk:
It would be good to have a package for the nvidia driver version. `__nvidia` ???

carterbox:
It would be good to have a package for __cudaarch ?? So binaries can be less fat or ship without PTX?

beeankha:
Due to the merging of the [virtual package plugin implementation feature](https://github.com/conda/conda/pull/11854), I'm going to go ahead and close this issue! If any specific virtual package plugins are requested to be implemented for the future, please feel free to create a new issue for it and add the https://github.com/conda/conda/labels/plugins label.

PR #11854, opened 2022-09-16
Implement Virtual Packages Plugins

### LINKED ISSUE #11811: Virtual Package Plugins
### Checklist

### What is the idea?

This work will be a continuation of the initial plugin mechanism implementation for `conda`.

### Why is this needed?

An implementation of virtual packages via the plugin system would be a good use case for the new plugin system. 

### What should happen?

[Virtual packages defined by `conda`](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html) can be implemented as core plugins under the `conda.plugins.*` namespace and registered directly in `conda.base.context.get_plugin_manager` instead of via entrypoints. 

### Additional Context

This work is a continuation of the initial plugin implementation [here](https://github.com/conda/conda/pull/11435) (connected to issue https://github.com/conda/conda/issues/11112).

beeankha:
This work will enable virtual packages plugins to be able to hook into `conda`. This is something with precedented demand:

- https://github.com/conda/conda/issues/10131

Fixes #11811


kenodegard (commented):
I'm confused, this implements the hook, but the hook isn't used anywhere? Further, what's the intention of the hook? Will simply installing the plugin set the virtual package as always "on" with a fixed version? E.g., if I'm on macOS I'll always have a corresponding OS package like `conda-plugin-osx-10.6` installed?

jezdez:
Yeah, this currently doesn't have the code to make use of the hook with the existing, included virtual packages, that were implemented in https://github.com/conda/conda/pull/11121/commits/e8f9a42e84c6e4a632ab02251d2300504f6a486a. @beeankha Could you look into porting this over to your code please?

beeankha:
Some of the CI test failures are due to this issue: https://github.com/conda/conda/issues/11905

conda/plugins/__init__.py
```diff
@@ -36,3 +36,23 @@ def conda_subcommands() -> Iterable[CondaSubcommand]:
...
+
+    :param name: Virtual package name (e.g., ``my_custom_os``).
+    :param version: Virtual package version (e.g., ``1.2.3``).
+    """
+    name: str
+    version: Optional[str]
+
+
+@_hookspec
+def conda_virtual_package_plugin() -> Iterable[CondaVirtualPackage]:
```
jezdez, line 53: For the subcommands we used `conda_subcommands` as the name, so we should stay consistent and not use the `_plugin` suffix but `conda_virtual_packages` (plural since it can `yield` multiple).

jezdez (dismissed):
Besides the naming issue, this LGTM!

conda/core/index.py
```diff
@@ -246,6 +193,22 @@ def get_archspec_name():
         return machine
 
 
+def _supplement_index_with_system(index):
+    registered_names = []
```
jezdez, line 197:
```suggestion
def _supplement_index_with_system(index):
    """
    Loads and populates virtual package records from conda plugins
    and adds them to the provided index, unless there is a naming
    conflict.
    """
    registered_names = []
```

jezdez (approved):

travishathaway (approved):
I have got a couple of things that I would like to see addressed that I believe should be pretty easy to do and leave the code in better shape. I don't want this to be a blocker though, so that's why I'm giving my approval. These recommendations are optional, and you can do what you want with them.

First, I think we should organize our `plugins` module so that it has room to grow. This can be accomplished by adding a sub-module `virtual_packages` to organize all the new files you added (e.g. `cuda.py`). After doing this, the directory structure would look like this:

```
./plugins
├── __init__.py
└── virtual_packages
    ├── __init__.py
    ├── archspec.py
    ├── cuda.py
    ├── linux.py
    ├── osx.py
    └── windows.py
```

Second, I think we should have a special `register_virtual_packages` function. Perhaps this could live in `conda/plugins/virtual_packages/__init__.py`? I'd like to see all the `pm.register` calls moved there. Here's what that would look like:

```python
def register_virtual_packages(pm: PluginManager) -> None:
   """
   Please add a nice doc string too explaining why this function exists!
   """
    pm.register(archspec)
    pm.register(cuda)
    pm.register(linux)
    pm.register(osx)
    pm.register(windows)
```

This would later be called and imported in the `get_plugin_manager` function.

Also, I think that you should mention something in the documentation about the following virtual package names being reserved:

- archspec
- cuda
- unix
- linux
- libc (more maybe?)
- win
- osx

If someone were to implement their own version of those, it would error out right? That should definitely be documented. It's okay not to do this now, but please make a ticket to do so in the future.

Also, the documentation tells me how to create a virtual package plugin, but I'm not necessary sold on why or when I might need to do this. I'm guessing we're still trying to figure this out ourselves, but let's make sure to come back to this piece of documentation in the future to present a more compelling use case.

That's it for now.

news/11854-virtual-packages-plugins
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a new plugin hook for virtual packages
```
kenodegard, line 3:
```suggestion
* Add a new plugin hook for virtual packages. (#11854)
```

conda/plugins/windows.py
```diff
@@ -0,0 +1,15 @@
...
+
+from .. import plugins
+
+
+@plugins.register
+def conda_virtual_packages():
+    if platform.system() != "Windows":
+        return
+
+    yield plugins.CondaVirtualPackage("win", None)
```
kenodegard, line 15: I understand this replicates current functionality, but why don't we include the version here?

kenodegard (changes requested):
A few minor edits, otherwise this looks great!

Would also like to see all of these virtual packages pushed down into a submodule as @travishathaway suggested.

conda/core/index.py
```diff
@@ -246,6 +193,27 @@ def get_archspec_name():
...
+                f"`{package.name}` key. Multiple conda plugins "
+                "are registering this virtual package via the "
+                "`conda_virtual_packages` hook, please make sure "
+                "you don't have any incompatible plugins installed."
+            )
+        registered_names.append(package.name)
+        rec = _make_virtual_package(f"__{package.name}", package.version)
+        index[rec] = rec
+
+
```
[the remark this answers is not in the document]
beeankha: Done!

dholth (dismissed):
Cool, it does use a simpler cache when maxsize is None https://github.com/python/cpython/blob/c11b667a1d8c13095b295d1a8c4ab6e4ccf0f895/Lib/functools.py#L549

beeankha:
pre-commit.ci autofix

jakirkham:
cc @kkraus14 @wolfv @CJ-Wright (including a few people who may have thoughts on this)

beeankha:
🚨 This PR is currently blocked by CI test failures (we are actively working on them!) 🚨

travishathaway:
Not sure where to put this as it relates to the `feature/plugins` branch, but I have encountered this repeatedly while working in this branch. Any who, here goes...

I would like to propose that we remove adding the `plugin_manager` to the context object:

https://github.com/beeankha/conda/blob/869d09564b0bbf7494c30772dfba88ee9c2364eb/conda/base/context.py#L414-L417

I think that the `get_plugin_manager` function will suffice and we are already seeing places in the code where we use either that function or the property on the context object:

- https://github.com/beeankha/conda/blob/869d09564b0bbf7494c30772dfba88ee9c2364eb/conda/cli/conda_argparse.py#L119  (uses `get_plugin_manager`)
- https://github.com/beeankha/conda/blob/869d09564b0bbf7494c30772dfba88ee9c2364eb/conda/core/index.py#L174 (uses the `context` object)

It's definitely small, but I think sticking to one method will help make our code easier to read.

beeankha:
@travishathaway re: your comment [here](https://github.com/conda/conda/pull/11854#issuecomment-1288781705)
> I would like to propose that we remove adding the plugin_manager to the context object

I think this is most likely fine to do, but let's tackle it in a separate PR since this one has already grown a lot in scope!

news/11854-virtual-packages-plugins
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a new plugin hook for virtual packages. (#11854)
```
beeankha, line 3:
```suggestion
* Add a new plugin hook for virtual packages and convert existing code for virtual packages (from index.py) into plugins. (#11854)
```

travishathaway (approved):

kenodegard (approved):
LGTM 🚀

PR #11857, opened 2022-09-19
Refactor subdir_data for incremental .jlap repodata

dholth:
### Description

Refactor subdir_data. It will need to provide the fetch code with a place to keep some state, and not worry about last-modified/etag in the same way in SubdirData.

How can we manage all the error conditions in jlapper? Should we incorporate a formal state machine? Can we detect a certain number of failures and fetch such channels without the new file?

Includes the parallel-downloads branch https://github.com/conda/conda/pull/11841

### Checklist - did you ...


dholth:
It appears `--experimental-solver libmamba` downloads repodata.json in its own way...

PR #11863, opened 2022-09-20
Fixes bug found during release of 22.9.0

travishathaway:
This is a proposed fix for an issue initially discovered while creating a new pull request for the [AnacondaRecipes/conda-feedstock](https://github.com/AnacondaRecipes/conda-feedstock) repository.

Pull request:

- https://github.com/AnacondaRecipes/conda-feedstock/pull/12

### Checklist - did you ...

### DIFF
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -28,7 +28,7 @@ SHEBANG_REGEX = (br'^(#!'  # pretty much the whole match string
                  br'(.*)'  # the rest of the line can contain option flags
                  br')$')  # end whole_shebang group
 
-MAX_SHEBANG_LENGTH = 512 if on_win else 127
+MAX_SHEBANG_LENGTH = 127
 
 
 class _PaddingError(Exception):
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -13,7 +13,7 @@ import sys
 from ..auxlib.ish import dals
 from ..base.constants import PREFIX_PLACEHOLDER
 from ..base.context import context
-from ..common.compat import on_win, on_linux
+from ..common.compat import on_win, on_linux, on_mac
 from ..exceptions import CondaIOError, BinaryPrefixReplacementError
 from ..gateways.disk.update import CancelOperation, update_file_in_place_as_binary
 from ..models.enums import FileMode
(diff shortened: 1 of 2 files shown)


conda/core/portability.py
```diff
@@ -28,7 +28,7 @@
                  br'(.*)'  # the rest of the line can contain option flags
                  br')$')  # end whole_shebang group
 
-MAX_SHEBANG_LENGTH = 127 if on_linux else 512  # Not used on Windows
+MAX_SHEBANG_LENGTH = 127 if (on_linux or on_mac) else 512  # Not used on Windows
```
kenodegard, line 31: Flip it?

```suggestion
MAX_SHEBANG_LENGTH = 512 if on_win else 127
```

travishathaway: Yeah, that's easy to parse! (for humans)

conda/core/portability.py
```diff
@@ -13,7 +13,7 @@
 from ..auxlib.ish import dals
 from ..base.constants import PREFIX_PLACEHOLDER
 from ..base.context import context
-from ..common.compat import on_win, on_linux
+from ..common.compat import on_win, on_linux, on_mac
```
kenodegard, line 16: Should be able to remove both of these now:

```suggestion
from ..common.compat import on_win
```

beeankha (dismissed):
👍

kenodegard (changes requested):
As a bugfix this needs a news snippet

beeankha (approved):

jezdez (approved):

conda/core/portability.py
```diff
@@ -28,7 +28,7 @@
                  br'(.*)'  # the rest of the line can contain option flags
                  br')$')  # end whole_shebang group
 
-MAX_SHEBANG_LENGTH = 127 if on_linux else 512  # Not used on Windows
+MAX_SHEBANG_LENGTH = 512 if on_win else 127
```
jaimergp, line 31: Windows has no concept of shebang, so if the intent here is to have 127 for both Linux and Mac, I'd say we keep it simple:

```suggestion
MAX_SHEBANG_LENGTH = 127
```

kenodegard: Ah, in which case I think we all misunderstood the purpose of this code then, this change shouldn't be merged.

cc @jezdez @travishathaway

travishathaway: @kenodegard, I think this still needs to be merged. My suggestion would be to go with what @jaimergp put forth. It fixes a bug that was found during a build test.

PR #11862, opened 2022-09-20
Flatten script building for `conda init --dev`

kenodegard:
### Description

Refactor/cleanup while working on #11812 

### Checklist - did you ...

- [x] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [x] ~Add / update necessary tests?~
- [x] ~Add / update outdated documentation?~

PR #11865, opened 2022-09-20
Allow `conda init --dev` to output to either fd or tmp

kenodegard:
### Description

Expose controlling where `conda init --dev` is output. There are three ways to output:
- stdout (default)
- a file descriptor (e.g., `--fd 3`), or
- a (temp) file (e.g., `--tmp $TMPDIR/conda-init-dev.$$.$RANDOM`)

This way, the shell interface can choose how to receive the initialization script (as opposed to being told how it must receive it).

Depends on #11862 

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -14,6 +14,7 @@ from argparse import (
 from logging import getLogger
 import os
 from os.path import abspath, expanduser, join
+from pathlib import Path
 from subprocess import Popen
 import sys
 from textwrap import dedent
@@ -698,7 +699,6 @@ def configure_parser_create(sub_parsers):
 def configure_parser_init(sub_parsers):
     help = "Initialize conda for shell interaction."
     descr = help
-
     epilog = dals(
         """
         Key parts of conda's functionality require that it interact directly with the shell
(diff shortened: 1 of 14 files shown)


jezdez (dismissed):
Thank you for that refactor!

kenodegard:
pre-commit.ci autofix

conda/cli/conda_argparse.py
```diff
@@ -684,9 +665,22 @@ def configure_parser_init(sub_parsers):
...
+        type=int,
+        help="When --dev is specified use --fd to pipe to a file descriptor besides stdout.",
+        default=None,
+    )
+    dev_piping_group.add_argument(
+        "--tmp",
+        type=Path,
+        help="When --dev is specified use --tmp to write to a temporary file instead of stdout.",
+        default=None,
+    )
```
kenodegard, line 683: Defining a mutex group where all of the args are `SUPPRESS`-ed breaks sphinx...

travishathaway (changes requested):
I noticed there's a lot of uncovered lines. Is there a way for to test this even if we have to use something like `subprocess.Popen`? I would feel a little better about this pull request if we could bump up the test coverage.

I'll be happy to be persuaded otherwise though if this represents a too big of a burden.

kenodegard:
this was a valuable learning and research effort but is unlikely to proceed in the current form

PR #11873, opened 2022-09-23
Investigating collapsing `conda_env` into `conda`

kenodegard:
### Description

This is the first attempt to collapse the two codes together.

This reimagines how we should organize `conda.cli` to avoid import penalties while avoiding everything `argparse` related being defined in `conda.cli.conda_argparse`. Every first-class subcommand is defined as a submodule within `conda.cli` and each submodule defines a `_parser` and `__main__` submodules. The bonus here is that we can import our parsers without the import penalty of importing `__main__` and we also expose a new way of running conda commands, e.g.:

```bash
python -m conda.cli.env --help
python -m conda.cli.env.config --help
python -m conda.cli.env.config vars --help
python -m conda.cli.env.config.vars.list --help
python -m conda.cli.env.config.vars.list --help
python -m conda.cli.env.config.vars.list --help
python -m conda.cli.env.create --help
python -m conda.cli.env.export --help
python -m conda.cli.env.list --help
python -m conda.cli.env.remove --help
python -m conda.cli.env.update --help
```

All other non-cli modules from `conda_env` are moved over into `conda.env`.

### Checklist - did you ...


kenodegard:
this was good research but we aren't ready for this work yet, closing since it's stale

PR #11877, opened 2022-09-23
Root user implies `conda init --system`

### LINKED ISSUE #9083: conda init should support modifying /etc/bashrc
If I run `conda init bash` as root, then it adds conda initialization to `/root/.bashrc`.

What I'd actually like to do, is update `/etc/bashrc` instead, so all users get that config (not just root). It would be nice if `conda init` had something like a `--global` option, that made it update `/etc/bashrc` instead of `/root/.bashrc`.

(While I could manually copy the section from `/root/.bashrc` to `/etc/bashrc`, I would like to avoid that if I can. This is being run from a Dockerfile which is installing conda into a container; the Dockerfile runs as root but then conda itself will be run from some other non-root user accounts in the docker container.)

(The same point applies to `/root/.zshrc` vs `/etc/zshrc`, `/root/.config/fish/config.fish` vs `/etc/fish/config.fish`, etc. I really only care about bash, but if you were to address this, you'd want to address it for as many of your supported shells as you feasibly could.)

(I realise some would argue that sharing a conda install between multiple users is maybe not a good idea. In actuality, the users don't add any environments at runtime, nor modify any environments, the container is just being used to run an application, and if we need to modify the environment, we actually do that by modifying the container build process, rather than doing it at runtime.)

shivance:
### Description

This PR is an attempt to fix the `conda init` ran as root, to initialize conda environment for all the users!

Resolves #9083 

### Checklist - did you ...

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -462,7 +462,9 @@ def make_initialize_plan(conda_prefix, shells, for_user, for_system, anaconda_pr
                 },
             })
 
-        if for_system:
+        from ..common._os import is_admin 
+
+        if for_system or is_admin():
             plan.append({
                 'function': init_sh_system.__name__,
                 'kwargs': {
@@ -470,7 +472,7 @@ def make_initialize_plan(conda_prefix, shells, for_user, for_system, anaconda_pr
                     'conda_prefix': conda_prefix,
                     'reverse': reverse,
                 },
-            })
+            })     
 
     if 'fish' in shells:
         if for_user:


kenodegard:
Instead of hardcoding this condition into the initialize function, I think that it makes more sense to define this as a conditional default for the `--system` option itself in `conda/cli/conda_argparse.py`:

https://github.com/conda/conda/blob/8730563d531e3d83bb14a23e97b0c2edcfb09589/conda/cli/conda_argparse.py#L718-L723

@jezdez what are your thoughts here?

kenodegard:
@conda-bot check

shivance:
> Instead of hardcoding this condition into the initialize function, I think that it makes more sense to define this as a conditional default for the `--system` option itself in `conda/cli/conda_argparse.py`:
> 
> https://github.com/conda/conda/blob/8730563d531e3d83bb14a23e97b0c2edcfb09589/conda/cli/conda_argparse.py#L718-L723
> 
> @jezdez what are your thoughts here?

@kenodegard As I'm new to conda, I'll try explaining what I understood the flow of program execution. 
When `conda init` is called.. `execute()` function from conda\cli\main_init.py executes, which in turn calls `initialize()` from conda\core\initialize.py with `for_system` passed through args , this argument tells what to do if the conda was initialized for the root !
These args can be controlled from conda\cli\argparse.py 

Earlier, I directly hard coded in initialize.py to check if root is being run throug `is_admin()` call,
now as per your suggestion, I'm supposed to add condition in the argparser file itself.

There are couple of things I'm cluless about
```
   1. How to define conditions through argparser?
   2. The condition remains same ? if system was the argument, then we initialize with path as '/etc/profile.d/conda.sh'  . 
  
```

https://github.com/shivance/conda/blob/ed6d03cb0b071faf7e4d6c490362f4fd837ec34d/conda/core/initialize.py#L437-L475

PR #11879, opened 2022-09-26
Remove toolz.accumulate

dholth:
### Description

Remove toolz.accumulate (only used once in project)

### Checklist - did you ...


conda/common/path.py
```diff
@@ -118,8 +113,15 @@ def explode_directories(child_directories, already_split=False):
...
     maybe_split = lambda x: x if already_split else x.split('/')
-    return set(concat(accumulate(join, maybe_split(directory))
-                      for directory in child_directories if directory))
+
+    exploded = set()
+    for directory in (directory for directory in child_directories if directory):
+        path = ""
+        for part in maybe_split(directory):
+            path = join(path, part)
+            exploded.add(path)
```
kenodegard, line 122: What about using `itertools.accumulate`?

jezdez: +1

dholth: Rewritten

jezdez (dismissed):
Huzzah!

jezdez (dismissed):

conda/common/path.py
```diff
@@ -85,14 +83,13 @@ def tokenized_startswith(test_iterable, startswith_iterable):
     return all(t == sw for t, sw in zip(test_iterable, startswith_iterable))
 
 
-def get_all_directories(files):
+def get_all_directories(files: Iterable[str]) -> list[tuple[str]]:
     return sorted(set(tuple(f.split('/')[:-1]) for f in files) - {()})
```
dholth, line 87: I'm confused about how this works with Windows

conda/common/path.py
```diff
@@ -113,12 +110,14 @@ def _process(x, y):
     return tuple('/'.join(leaf) for leaf in leaves)
 
 
-def explode_directories(child_directories, already_split=False):
+def explode_directories(child_directories: Iterable[tuple[str, ...]]):
```
dholth, line 113: simplify signature (require already_split) since it's only called one way

conda/common/path.py
kenodegard: is [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) called earlier?

conda/common/path.py
```diff
@@ -85,14 +83,13 @@ def tokenized_startswith(test_iterable, startswith_iterable):
...
-def get_all_directories(files):
+def get_all_directories(files: Iterable[str]) -> list[tuple[str]]:
     return sorted(set(tuple(f.split('/')[:-1]) for f in files) - {()})
 
 
-def get_leaf_directories(files):
-    # type: (List[str]) -> List[str]
-    # give this function a list of files, and it will hand back a list of leaf directories to
-    #   pass to os.makedirs()
+def get_leaf_directories(files: Iterable[str]) -> Sequence[str]:
```
kenodegard, line 90: Let's use os.PathLike when we are expecting paths:

```suggestion
def get_leaf_directories(files: Iterable[os.PathLike]) -> tuple[str, ...]:
```

conda/common/path.py
dholth: Could it be looking at tarfile paths

conda/common/path.py
kenodegard, line 113:
```suggestion
def explode_directories(child_directories: Iterable[tuple[str, ...]]) -> set[str]:
```

kenodegard: Unsure if we should deprecate this change:

```
def explode_directories(child_directories: Iterable[tuple[str, ...]], already_split=NULL):
    if already_split != NULL:
        warn(...)
```

kenodegard (changes requested):
Minor edits, otherwise looking good

conda/common/path.py
jezdez: That's a little weird, when following the code path it has a ton of indirection via `PathsData`, but I think no `os.path.normpath` use.

dholth: In `conda.models.records`
```
class PathsData(Entity):
    # from info/paths.json
    paths_version = IntegerField()
    paths = ListField(PathData)
```

dholth (commented):
[exception PendingDeprecationWarning

    Base class for warnings about features which are obsolete and expected to be deprecated in the future, but are not deprecated at the moment.

    This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is preferred for already active deprecations.](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning)
    
The `already_split` feature is gone. It would have to split the paths before sending them to `chain.from_iterable(...)` otherwise this function will split by characters.

dholth:
pre-commit.ci autofix

PR #11880, opened 2022-09-26
remove encoding pragma (not needed in Python 3)

dholth:
### Description

Configure pre-commit to remove encoding-pragma; utf-8 is default in Python 3

### Checklist - did you ...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,7 @@ repos:
       # catch git merge/rebase problems
       - id: check-merge-conflict
       - id: fix-encoding-pragma
+        args: [--remove]
         exclude: ^conda/auxlib/
   - repo: https://github.com/akaihola/darker
     rev: 1.5.1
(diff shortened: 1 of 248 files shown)


jezdez (approved):
I think this is a good idea, but I'd be interested what @conda-core thinks about it?

kenodegard:
Looks good to me, will need to run pre-commit CI before merging

dholth:
I was surprised to see the pragma magically appear on a new Python file, then I found the existing "add the pragma" setting.

jezdez:
pre-commit.ci autofix

jezdez (approved):

kenodegard:
We should remove the pre-commit pragma hook now

dbast:
maybe we could even activate the pyupgrade pre-commit hook (as in conda-build) and get rid of more, like the future imports etc, it also would have removed the encoding pragma.

jezdez:
> maybe we could even activate the pyupgrade pre-commit hook (as in conda-build) and get rid of more, like the future imports etc, it also would have removed the encoding pragma.

Yeah, most likely, would it only run pyupgrade on changed files in CI?

kenodegard:
https://github.com/conda/conda/pull/11903

PR #11889, opened 2022-09-27
Replace "experimental solver" with just "solver"

### LINKED ISSUE #11875: Drop `experimental_solver` context key and replace with a new one named `solver`, aliasing the former to the latter for a couple releases

jaimergp:
### Checklist - did you ...

Closes #11875


conda/base/context.py
```diff
@@ -338,10 +338,21 @@ class Context(Configuration):
...
+        aliases=('experimental_solver',),
     )
 
+    @property
+    def experimental_solver(self):
+        # TODO: Remove in a later release
+        warnings.warn(
+            "'context.experimental_solver' is deprecated and will be removed. "
+            "Please use 'context.solver' instead.",
+            DeprecationWarning
```
kenodegard, line 352:
```suggestion
            PendingDeprecationWarning
```

conda/cli/conda_argparse.py
```diff
@@ -245,6 +246,13 @@ def __call__(self, parser, namespace, values, option_string=None):
         items.extend(values or [self.const])
         setattr(namespace, self.dest, items)
 
+
+class DeprecatedAction(Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        warnings.warn(f"Option {self.option_strings} is deprecated!", DeprecationWarning)
+        super().__call__(parser, namespace, values, option_string)
```
kenodegard, line 253: I think we want a `PendingDeprecationAction` first.

kenodegard (commented):
The deprecations need to be marked as pending first per CEP 9.

conda/base/context.py
dholth: Python says https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning `This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is preferred for already active deprecations.`

jaimergp:
Thinking about the issue template, we might want to remove it altogether? But not in this PR, so it stays available til next release.

conda/base/constants.py
```diff
@@ -251,12 +251,16 @@ class SatSolverChoice(ValueEnum):
...
 
 
-class ExperimentalSolverChoice(ValueEnum):
+class SolverChoice(ValueEnum):
     CLASSIC = 'classic'
     LIBMAMBA = 'libmamba'
     LIBMAMBA_DRAFT = 'libmamba-draft'
 
 
+# TODO: Remove in a later release - compatibility alias
```
LtDan33, line 260: Do we want to call out the planned release we want to remove this in?

conda/cli/conda_argparse.py
```diff
@@ -1779,19 +1797,28 @@ def add_parser_prune(p):
...
+    )
+    group.add_argument(
         "--experimental-solver",
-        dest="experimental_solver",
-        choices=[v.value for v in ExperimentalSolverChoice],
-        help="EXPERIMENTAL. Choose which solver backend to use.",
+        action=PendingDeprecationAction,
+        dest="solver",
+        choices=[v.value for v in SolverChoice],
+        help="DEPRECATED. Please use '--solver' instead.",
```
LtDan33, line 1821: Even though we have the `deprecated` message, should we still have what this is intended to do? "Choose which solver backend to use."

conda/base/constants.py
jaimergp: I'd like to do that, but I think we need to come up with a mechanism / protocol to handle this across the codebase. Either a specific comment syntax or a time-controlled unit test that keeps track and notifies us?

conda/cli/conda_argparse.py
jaimergp: They show up together in the help message, so I don't know if repeating it again helps clarify or adds noise:

```
Solver Mode Modifiers:
  --strict-channel-priority
                        Packages in lower priority channels are not considered if a package with the same name appears in a higher priority channel.
  --no-channel-priority
                        Package version takes precedence over channel priority. Overrides the value given by `conda config --show channel_priority`.
  --no-deps             Do not install, update, remove, or change dependencies. This WILL lead to broken environments and inconsistent behavior. Use at your own risk.
  --only-deps           Only install dependencies.
  --no-pin              Ignore pinned file.
  --solver {classic,libmamba,libmamba-draft}
                        Choose which solver backend to use.
  --experimental-solver {classic,libmamba,libmamba-draft}
                        DEPRECATED. Please use '--solver' instead.
  --force-reinstall     Ensure that any user-requested package for the current operation is uninstalled and reinstalled, even if that package already exists in the environment.
  --freeze-installed, --no-update-deps
                        Do not update or change already-installed dependencies.
  --update-deps         Update dependencies that have available updates.
  -S, --satisfied-skip-solve
                        Exit early and do not run the solver if the requested specs are satisfied. Also skips aggressive updates as configured by the 'aggressive_update_packages' config setting. Use 'conda info --describe
                        aggressive_update_packages' to view your setting. --satisfied-skip-solve is similar to the default behavior of 'pip install'.
  --update-all, --all   Update all installed packages in the environment.
  --update-specs        Update based on provided specifications.
```

conda/base/constants.py
travishathaway: I think it's a good idea that we automate this (perhaps a custom PyLint parser?), but this work is going to be very involved and should be handled separately.

I will create a spike issue for implementing a deprecation mechanism that can be detected by our CI workflow.

travishathaway (approved):
This looks good. @jaimergp, let me know if you need any further reviews on this in case you make anymore updates. @jezdez is out this week 🌴 😎

conda/base/constants.py
```diff
@@ -251,12 +251,16 @@ class SatSolverChoice(ValueEnum):
...
+ExperimentalSolverChoice = SolverChoice
```
jaimergp, line 261: Do we need to mark this as PendingDeprecation too? How can we do it in a transparent way? Subclassing and overriding `__getattribute__` or similar?

conda_env/cli/main_remove.py
```diff
@@ -32,7 +32,7 @@ def configure_parser(sub_parsers):
     )
 
     add_parser_prefix(p)
-    add_parser_experimental_solver(p)
+    add_parser_solver(p)
```
jaimergp, line 35: Should this one also have an alias for now?

travishathaway: I don't think so. This has not been in conda for that long, so I really don't think there are other projects out there depending on it. Plus, this would be a weird thing for other people to depend on.

conda/base/constants.py
travishathaway: If this can be easily done, then let's do it.

PR #11913, opened 2022-09-30
substitute for toolz.groupby

dholth:
### Description

As part of the "don't depend on toolz" project, replace groupby with own implementation.

### Checklist - did you ...


dholth:
Unfortunately it's slower, whether this matters depends on how much we groupby.

```
from tlz import groupby as tlz_groupby
from conda.tlz import groupby_to_dict
import timeit
import random


dataset = [{"k": random.randint(0, 100), "v": i} for i in range(1000)]


assert tlz_groupby(lambda d: d['k'], dataset) == groupby_to_dict(lambda d: d['k'], dataset)

print("toolz version", timeit.timeit("tlz_groupby(lambda d: d['k'], dataset)", globals=globals(), number=100000))
print("our version", timeit.timeit("groupby_to_dict(lambda d: d['k'], dataset)", globals=globals(), number=100000))
```

kenodegard (commented):
Ah, I see how `itertools.groupby` differs significantly from that in `toolz`. Hm, not super excited about introducing `conda.tlz`, perhaps we can move this into `conda.utils` instead?

dholth:
@jezdez I renamed it and moved it into conda.common. I know the function does not return an iterator, but it accepts one.

dholth:
In conda-index, I replaced groupby with inline code and it was faster and simpler. For example conda uses it to split some things into a True and False set (a case that is probably simpler but not faster without groupby)

PR #11915, opened 2022-09-30
fixed errors when renaming without an active env

FelisNivalis:
`active_prefix` returns empty string instead of `None` when no active env

### Description

Got errors when trying to rename an environment with no active environments
```
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/exceptions.py", line 1129, in __call__
        return func(*args, **kwargs)
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/cli/main.py", line 86, in main_subshell
        exit_code = do_call(args, p)
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/cli/conda_argparse.py", line 93, in do_call
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/cli/main_rename.py", line 53, in execute
        source = validate_src(args.name, args.prefix)
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/cli/main_rename.py", line 28, in validate_src
        if common.is_active_prefix(prefix):
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/cli/common.py", line 87, in is_active_prefix
        paths_equal(prefix, context.active_prefix)
      File "/home/.../miniconda3/lib/python3.9/site-packages/conda/common/path.py", line 57, in paths_equal
        return abspath(path1) == abspath(path2)
      File "/home/.../miniconda3/lib/python3.9/posixpath.py", line 375, in abspath
        path = os.fspath(path)
    TypeError: expected str, bytes or os.PathLike object, not NoneType
```

So basically when no active env, `context.active_prefix` returned a `None`, which further caused `os.fspath` to raise an error.

I just fixed it in a simplest way so that `context.active_prefix` always returns a string, but if, for example, `None` means something special for `active_prefix`, we can try to figure out another way.

### Checklist - did you ...


jezdez:
@conda-bot check

news/11915-no-active-env-renaming-error
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixed errors when renaming without an active env
```
beeankha, line 7:
```suggestion
* Fixed errors when renaming without an active env (#11915)
```

beeankha:
Hi @FelisNivalis , thank you for your contribution! We really appreciate you taking the time to submit this suggested fix. 

The `active_prefix()` method in the `Context` class (which you made code changes for in this pull request) always expects the root prefix to have a value instead of `''` or `None`, so that's something we would not like to alter; additionally, the `Context` object affects many other parts of the codebase, so changing it should generally be avoided since there will definitely be some unforeseeable consequences.

Additionally, changing the `active_prefix()` method and attempting a `rename` while conda is deactivated means that there is no straightforward way to explicitly prevent conda users from inadvertently renaming their base environment. We need to avoid this scenario since renaming `base` irreparably breaks conda.

The more ideal path forward may be to alter the `is_active_prefix()` function in [common.py](https://github.com/conda/conda/blob/main/conda/cli/common.py) to detect whether or not a prefix is active and return `False` if there isn't one. This will ensure that the `TypeError` you were seeing previously will not surface during an attempted rename, and that an environment can get successfully renamed even when conda is not active.

I've pushed a new commit to this pull request; if you agree to this change, we'll go forward with this PR and keep your name attached to it as a contributor since you so kindly found this bug and submitted a fix. Please let us know your thoughts!

travishathaway (approved):

PR #11286, opened 2022-03-03
ci-skip.yml to circumvent skipped required checks

kenodegard:
When checks are marked as required but also skipped (e.g. when only docs are modified) we get into a catch 22 state. We address this issue by defining a dummy workflow to satisfy the checks.

See [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

Should resolve merging block for https://github.com/conda/conda/pull/11278

PR #10864, opened 2021-08-17
Test suite improvements for usability, correctness and performance

dbast:
NOTE: This was previous opened in #10854 but got closed accidentally.

This does a number of things:
- Use more runners for the integration tests by splitting the tests into groups
- Fix Codecov reporting on Windows
- Separates the documentation build into own workflow (only triggered upon changes to docs files)
- Combine various requirements files.
- Add pytest-rerunfailures to be able to rerun flaky tests.
- Refactor a few unittest style tests to be pytest tests to reduce flakiness.
- Move pytest config file and fixtures in better places.
- Fix temp folder issue on Windows.
- Cleanup Makefile.
- Lint on all platforms, in an own workflow file.

Fix #10846.

PR #11941, opened 2022-10-06
Fix CI workflows to always run

jezdez:
### Description

It seems as if the work in #10864 in combination with #11286 might have created a scenario in which tests were skipped for PR containing docs.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,6 +240,7 @@ jobs:
     steps:
       - name: Download test results
         uses: actions/download-artifact@v3
+
       - name: Upload combined test results
         # provides one downloadable archive of all .coverage/test-report.xml files
         # of all matrix runs for further analysis.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -252,7 +249,13 @@ jobs:
...
+
       - name: Test Summary
         uses: test-summary/action@v1
         with:
           paths: ./test-results-${{ github.sha }}-**/test-report*.xml
+
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
```
kenodegard, line 261: This will always be true since the `needs` condition will prevent the `analyze` job from running if anything failed.

jezdez: Even with the `if: always()` condition above?

jezdez: I've following the example on https://github.com/re-actors/alls-green#usage, e.g. https://github.com/pyca/cryptography/blob/main/.github/workflows/ci.yml#L518-L531

jezdez: https://github.com/orgs/community/discussions/26733#discussioncomment-3253155 has some background

kenodegard: oh I see, the analysis job is always run, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs

so we want the analysis job to run but "fail" if any of the dependencies also failed?

jezdez: Yeah, basically

kenodegard (approved):

PR #11960, opened 2022-10-13
Merge plugins feature branch into main.

jezdez:
### Description

This PR basically exists to:

a) easily update the feature branch by using GitHub's regular rebase/merge button
b) have a place where we can do a final review

### Checklist - did you ...


beeankha (approved):

travishathaway (approved):

kenodegard (approved):

jaimergp:
I think the conda recipe at `/recipe` needs to have `pluggy` as a dependency. Otherwise the PR preview builds fail :D [See run here](https://github.com/conda/conda/pull/12043#issuecomment-1298800258).

kenodegard:
@jaimergp pluggy is listed as a dependency, the build failures is related to how conda-build runs `python -m conda` (afaik your isolation PR is the fix we need), this is the same issue that we've encountered when we attempted to remove toolz from vendored

dholth:
Still it seems like there ought to be possible to install pluggy into whichever environment conda build uses to run "python -m conda", before conda-build runs

jaimergp:
Oh, I see @kenodegard, thanks for the clarification. Let's see if we can get https://github.com/conda/conda-build/pull/4604 in!

PR #11962, opened 2022-10-13
Check for actions updates every week.

jezdez:
### Description

I noticed that we have a number of actions using deprecated GitHub APIs and we should probably automatically get update PRs for these.

### Checklist - did you ...


dholth:
monthly would also be fine

dholth (approved):

PR #11963, opened 2022-10-13
Update a few actions to reduce warnings.

jezdez:
### Description

Some of the actions are outdated and need to be bumped to upgrade to Node 16. This is related to #11962 which would automate this.

### Checklist - did you ...


dholth (approved):
Great! Maybe the new versions will even be better.

PR #11971, opened 2022-10-13
Prototype for a conda session manager and context extension (do not merge!)

travishathaway:
This is a prototype for a conda session manager and how we would be able to extend this using our plugin system. Additionally, it also defines a plugin hook for extending the context object.

These implementations are intentionally incomplete as they are more geared towards starting a conversation about the best way to implement such things in the current conda code base.

There are extensive comments within the code itself that I highly encourage you to read, especially the comments surrounding the extending the `Context` object.


jezdez:
Just removed the emoji from the PR title to reduce visual clutter in actions overview. :)

Issue #11975, opened 2022-10-14
conda's old vendored six.py emits import warnings on python=3.10


erykoff:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

I set up a new env with the latest `conda=22.9.0` and `python=3.10`.  Then if I do:

```
$ python -Wall
>>> from conda.cli.python_api import Commands
```
It prints...
```
<frozen importlib._bootstrap>:914: ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()
<frozen importlib._bootstrap>:914: ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()
...
```
This is coming from the old vendored version of `six.py` which needs to be updated to 1.16.0 to properly support `python=3.10`.

### Conda Info

```shell
active environment : testconda
    active env location : /Users/erykoff/miniforge3/envs/testconda
            shell level : 1
       user config file : /Users/erykoff/.condarc
 populated config files : /Users/erykoff/miniforge3/.condarc
                          /Users/erykoff/.condarc
          conda version : 4.13.0
    conda-build version : not installed
         python version : 3.9.2.final.0
       virtual packages : __osx=12.6=0
                          __unix=0=0
                          __archspec=1=arm64
       base environment : /Users/erykoff/miniforge3  (writable)
      conda av data dir : /Users/erykoff/miniforge3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-arm64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /Users/erykoff/miniforge3/pkgs
                          /Users/erykoff/.conda/pkgs
       envs directories : /Users/erykoff/miniforge3/envs
                          /Users/erykoff/.conda/envs
               platform : osx-arm64
             user-agent : conda/4.13.0 requests/2.28.1 CPython/3.9.2 Darwin/21.6.0 OSX/12.6
                UID:GID : 501:20
             netrc file : None
           offline mode : False

```
```


### Conda Config

```shell
==> /Users/erykoff/miniforge3/.condarc <==
channels:
  - conda-forge

==> /Users/erykoff/.condarc <==
auto_activate_base: False
```
```


### Conda list

```shell
# packages in environment at /Users/erykoff/miniforge3/envs/testconda:
#
# Name                    Version                   Build  Channel
[...]

PR #11976, opened 2022-10-14
Update vendored six.py from 1.10.0 to 1.16.0.

erykoff:
This fixes importlib warnings on python=3.10.  See #11975.

### Checklist - did you ...

### DIFF
--- a/conda/_vendor/six.py
+++ b/conda/_vendor/six.py
@@ -1,6 +1,4 @@
-"""Utilities for writing code that runs on Python 2 and 3"""
-
-# Copyright (c) 2010-2015 Benjamin Peterson
+# Copyright (c) 2010-2020 Benjamin Peterson
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


dholth:
This is great!

It looks like we import six in three places.

conda/exports

conda/auxlib/compat

conda/auxlib/decorators

Could we delete six?

I'm unsure about the `necessary for conda-build` comment in exports.py

erykoff:
I was just saying on a slack thread with colleagues "I have not tried to figure out how to get rid of conda’s use of six.py."  It would certainly be preferable to updating six, but I worry that there are tendrils further in than I have time for at the moment.

erykoff:
Okay, I see this in `conda-build`: https://github.com/conda/conda-build/blob/main/conda_build/conda_interface.py#L71-L72 So removing `six.py` from `conda` definitely isn't possible until all uses of these exports are scrubbed from `conda-build` (at least!).  ☹️

erykoff:
Attempt to remove six.py on #11979.

erykoff:
I updated `vendor.txt`, but I'm hopeful that this PR is unnecessary if #11979 is accepted.

travishathaway:
@erykoff, I like the approach in https://github.com/conda/conda/pull/11979 much more and will help you review that and account for any dependency issues.

Issue #11918, opened 2022-10-03
Stacktrace Shown rather than standard error message when libmamba gets HTTP 404


remkade:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I looked at the [README.md](https://github.com/conda-incubator/conda-libmamba-solver) and didn't see anything to help

### What happened?

When you try to use a non-existent channel in conda with the default solver you get a helpful HTTP 404 error.

```
> conda create --dry-run -n intel-test -c main/nonexistent -c main tensorflow
Collecting package metadata (current_repodata.json): failed

UnavailableInvalidChannel: HTTP 404 NOT FOUND for channel intel/validation <https://conda.anaconda.org/intel/validation>

The channel is not accessible or is invalid.

You will need to adjust your conda configuration to proceed.
Use `conda config --show channels` to view your configuration's current state,
and use `conda config --show-sources` to view config file locations.
```

But when you use libmamba:

```
> conda create --dry-run -n intel-test -c main/nonexistent -c main --experimental-solver=libmamba tensorflow
Traceback (most recent call last):
  File "/Users/kleaders/anaconda3/lib/python3.8/site-packages/conda_libmamba_solver/solver.py", line 224, in solve_final_state
    index = LibMambaIndexHelper(
  File "/Users/kleaders/anaconda3/lib/python3.8/site-packages/conda_libmamba_solver/solver.py", line 87, in __init__
    self._index = load_channels(
  File "/Users/kleaders/anaconda3/lib/python3.8/site-packages/conda_libmamba_solver/mamba_utils.py", line 117, in load_channels
    index = get_index(
  File "/Users/kleaders/anaconda3/lib/python3.8/site-packages/conda_libmamba_solver/mamba_utils.py", line 98, in get_index
    is_downloaded = dlist.download(api.MAMBA_DOWNLOAD_FAILFAST)
RuntimeError: Multi-download failed.
```

### Conda info

```Shell
active environment : base
    active env location : /Users/kleaders/anaconda3
            shell level : 1
       user config file : /Users/kleaders/.condarc
 populated config files : /Users/kleaders/.condarc
          conda version : 22.9.0
    conda-build version : 3.21.9
         python version : 3.8.13.final.0
       virtual packages : __osx=10.16=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /Users/kleaders/anaconda3  (writable)
[...]

PR #11993, opened 2022-10-18
Add solver plugin hook

jaimergp:
### Description

Originally opened by @jezdez at https://github.com/jaimergp/conda/pull/2, but the target PR was merged before I saw it so... (re)-creating it here for better visibility

### Checklist - did you ...


conda/base/context.py
```diff
@@ -850,17 +837,17 @@ def user_agent(self):
...
             try:
                 # Solver.user_agent has to be a static or class method
                 user_agent_str += f" {_get_solver_class().user_agent()}"
             except Exception as exc:
                 log.debug(
                     "User agent could not be fetched from solver class '%s'.",
-                    self.solver.value,
+                    self.solver,
                     exc_info=exc
                 )
```
jaimergp, line 852: If we pursue [this suggestion](https://github.com/conda/conda/issues/11918#issuecomment-1265619129), the whole user agent logic can be removed (a solver should not query anything on its own) because we would be using conda fetchers anyway.

conda/plugins/solvers.py
```diff
@@ -0,0 +1,59 @@
...
+        ),
+        key=lambda solver: solver.name,
+    )
+    # Check for conflicts
+    seen = set()
+    conflicts = [
+        solver
+        for solver in solvers
+        if solver.name in seen or seen.add(solver.name)
+    ]
```
jaimergp, line 46: Is this checking for duplicate entries across plugins?

conda/plugins/specs.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@_hookspec
+def conda_solvers() -> Iterable[CondaSolver]:
+    """
+    Register external solvers in conda.
+
+    :return: An iterable of solvers entries.
+    """
+    ...
```
jaimergp, line 58: Why does this hook point need an ellipsis while the `conda_subcommand` does not?

tests/cli/test_main_clean.py
```diff
@@ -187,7 +187,7 @@ def test_clean_tempfiles(clear_cache):
 def test_clean_logfiles(clear_cache):
```
jaimergp, line 187: Do we need to remove this test or are we leaving it here so people can remove old logfiles from previous installations?

jezdez:
pre-commit.ci autofix

conda/plugins/types.py
```diff
@@ -24,11 +24,25 @@ class CondaSubcommand(NamedTuple):
...
     name: str
     version: str | None
+
+
+class CondaSolver(NamedTuple):
+    """
+    A conda solver.
+
+    :param name: Subcommand name (e.g., ``conda my-subcommand-name``).
+    :param backend: Callable that will instantiated as the solver backend.
```
travishathaway, line 42:
```suggestion
    :param backend: Callable that will be instantiated as the solver backend.
```

conda/plugins/types.py
```diff
@@ -24,11 +24,25 @@ class CondaSubcommand(NamedTuple):
...
+    """
+    name: str
+    backend: Callable[
+        [list[str]],
+        int | None,
```
travishathaway, line 47: Doesn't this need to be Sovler object. Looks a lot like the CondaSubCommand type 🤔

conda/plugins/types.py
```diff
@@ -24,11 +24,25 @@ class CondaSubcommand(NamedTuple):
...
 
     name: str
     version: str | None
+
+
+class CondaSolver(NamedTuple):
+    """
+    A conda solver.
+
+    :param name: Subcommand name (e.g., ``conda my-subcommand-name``).
```
travishathaway, line 41: Let's update this 😁 (looks like it was copied over from the CondaSubCommand type.

conda/plugins/types.py
jezdez: True, that's weird!

conda/plugins/manager.py
```diff
@@ -36,11 +38,45 @@ def load_plugins(self, *plugins) -> list[str]:
                 plugin_names.append(plugin_name)
         return plugin_names
 
+    def get_registered_plugins(self, plugin_name: str) -> list:
+        """
+        Return all registered plugins for the given name.
+        """
+        if not plugin_name:
```
travishathaway, line 45: If we do a check below to see if the hook attribute exists on the `self.hook` object, is this check really necessary? Even if the string value is empty it will still raise a nice exception and not blow up in a stack trace.

Furthermore, by being a positional argument, it's already assumed that we will pass in a value and because we use the `str` type hint, a string is strongly encouraged 😅 

I think we could ditch this check and make the code a little more legible. Definitely optional though.

travishathaway (commented):
I think this pull request is almost ready to go. Do we have any plans to address the missing code coverage, specifically around the error cases? How difficult will these be to test?

jaimergp:
Re-running failing job because of a shell timeout (current value is 12s, imo a bit low; #12043 raises it to 30s).

conda/base/context.py
jaimergp: I don't anticipate any challenging changes. It might be just a single line, plus testing.

jaimergp: That said, it might still be useful to keep user-agent info around to inform future choices like when/if to change the default solver choice. 

Alternatively, a user-agent plugin hook could be used? But perhaps this opens too many gates :D

conda/cli/conda_argparse.py
```diff
@@ -210,6 +181,8 @@ def error(self, message):
         super().error(message)
 
     def print_help(self):
+        # This code path is pending deprecation; it should be removed when subcommands
+        # via executables goes away
```
jaimergp, line 185: Should we add a warning then?

conda/core/index.py
```diff
@@ -174,7 +174,7 @@ def _supplement_index_with_system(index):
     """
     registered_names = []
     pm = get_plugin_manager()
-    for package in chain(*pm.hook.conda_virtual_packages()):
+    for package in pm.get_registered_plugins("virtual_packages"):
```
jaimergp, line 177: I know this is not part of  this PR but a thought: should all these "magic strings" be part of a predefined Enum?

conda/core/solve.py
```diff
@@ -23,59 +23,48 @@
...
+    for solver in solvers:
+        solvers_mapping[solver.name.lower()] = solver.backend
 
-    raise ValueError(
-        f"You have chosen a non-default solver backend ({key}) "
-        f"but it was not recognized. Choose one of "
-        f"{[v.value for v in SolverChoice]}"
-    )
+    if key in solvers_mapping:
+        return solvers_mapping[key]
```
jaimergp, line 61: What about:

```
try:
  return solvers_mapping[key]
except KeyError:
  raise ValueError(...)
```

Maybe `CondaValueError`?

conda/plugins/types.py
```diff
@@ -24,11 +24,23 @@ class CondaSubcommand(NamedTuple):
...
+class CondaSolver(NamedTuple):
+    """
+    A conda solver.
+
+    :param name: Solver name (e.g., ``custom-solver``).
+    :param backend: Callable that will be instantiated as the solver backend.
+    """
+
+    name: str
+    backend: Callable
```
jaimergp, line 46: Shoudn't this type be a class of `conda.core.solve.Solver`? Or a `Protocol` implementing the API? A `Callable` is too broad imo. If we want to leave as is, then a pointer about the expected API contract (`__init__` args, public methods and their return types) should be added to the docs.

jaimergp (commented):
This is looking nice!

I added some comments that do not really change the current work much. I would like to take a deeper look at the tests and how it feels to use this locally but I don't have time right now, so expect a 2nd round of comments tomorrow. Thanks!

conda/plugins/manager.py
jezdez: FWIW I added it to handle passing in an empty string, but that'll still be handled by the check for the hook below that. 👍

conda/cli/conda_argparse.py
jezdez: Nah, not sure why this is here tbh

conda/core/index.py
jezdez: Yeah, probably, the name is derived from the hookspec we specify, not sure how to best maintain it, or if this is needed for the first launch though

conda/core/solve.py
jezdez: Good idea!

conda/plugins/types.py
jezdez: Will using `conda.core.solve.Solver` here include subclasses of it or do we need to use `typing.Type[conda.core.solve.Solver]`?

tests/plugins/test_solvers.py
```diff
@@ -0,0 +1,114 @@
...
+from conda.core import solve
+from conda.exceptions import PluginError
+from conda import plugins
+
+
+log = logging.getLogger(__name__)
+
+
+class Solver(solve.Solver):
+    pass
```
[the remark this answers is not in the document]
jezdez: Just iterating over the test and left-over. Removing it..

conda/plugins/types.py
jezdez: I remember now why this was a problem, our use of the plugin manager in the various parts of conda triggered a cyclic import error when importing from the `conda.core.solve` here. We could inline these imports there and see if that helps.

jezdez: I've moved the loading of the plugin manager back into the context object (as it used to be) in https://github.com/conda/conda/pull/11993/commits/36430f1ceacc441357917ff7e75cfbdb28994b41, so that I don't have to inline it all over the place.

travishathaway (approved):
Don't see any obvious problems. Let me know if you change anything and need a second round of review.

conda/plugins/types.py
```diff
@@ -24,11 +25,23 @@ class CondaSubcommand(NamedTuple):
...
+class CondaSolver(NamedTuple):
+    """
+    A conda solver.
+
+    :param name: Solver name (e.g., ``custom-solver``).
+    :param backend: Callable that will be instantiated as the solver backend.
+    """
+
+    name: str
+    backend: Solver
```
travishathaway, line 47:
```suggestion
    backend: type[Solver
```

Discovered this today while looking at the code in my IDE. Otherwise, it will complain because it expected to receive an instantiated `Solver` object.

travishathaway (changes requested):

conda/plugins/types.py
jezdez: Yeah, I asked that in https://github.com/conda/conda/pull/11993#discussion_r1019000802

jezdez: @travishathaway This should have been this though? 
```suggestion
    backend: type[Solver]
```

travishathaway: Yes, sorry for the syntax error!

conda/base/context.py
jezdez: Yeah, a user-agent extension plugin hook has been discussed and we should make use of it going forward. But not in the coming release.

conda/base/context.py
```diff
@@ -414,6 +412,11 @@ def post_build_validation(self):
             errors.append(error)
         return errors
 
+    @property
+    def plugin_manager(self):
```
travishathaway, line 416:
```suggestion
    def plugin_manager(self):
    """
    This is the preferred way of accessing the ``PluginManager`` object for this application.
    This property appears here to avoid problems with cyclical imports elsewhere in the code.
    """
     
```

travishathaway (changes requested):
Couple of additions to the docs that I think will help us later (especially the comment on the plugin_manager property for the context object 😅).

Other than that, looks good.

travishathaway (approved):

jezdez (approved):

jaimergp (commented):
I played a bit with the plugin PRs (this one and the sibling in `conda-libmamba-solver`) and it works as expected.

I don't know how to feel about `conda.core.solve._get_solver_class()` still being the default way of instantiating the configured solver. It feels a bit weird to have this as a private function (downstream libraries will need to call this), which was okay during the "experimental" phase, but maybe not anymore. Not saying we should change it now, but at some point we might need to think a mechanism that feels better supported / guaranteed to work robustly.

I can't "approve" this because I opened it, but let this comment be equivalent! 

✅ 👍 👌 🆗  :D

Issue #11893, opened 2022-09-27
Spaces in prefix fix breaks passing the `$PS1` environment variable


kenodegard:
### What happened?

After releasing conda 22.9.0, users (predominantly Linux) identified an issue (https://github.com/conda/conda/issues/11885) where the prompt would no longer update correctly on posix. After some digging, the issue appears because of the spaces in prefix fix (#11676).

How to replicate:

1. Update conda:
    
    ```
    conda update conda
    ```

2. At this point, depending on your system, the issue may already be present. If not, you'll need to continue to mock the issue users are encountering.

3. Modify `$CONDA_PREFIX/bin/conda`'s shebang to use the long prefix/space in prefix trick:
 
    So from this:
    ```
    #!/Users/kodegard/.conda/4.12.0/3.9/bin/python
    # -*- coding: utf-8 -*-
    ...
    ```

    To this:
    ```
    #!/bin/sh
    '''exec' "/Users/kodegard/.conda/4.12.0/3.9/bin/python" "$0" "$@"
    ' '''
    # -*- coding: utf-8 -*-
    ...
    ```

4. Running `conda activate` will now break your prompt.

This has to do with how environment variables are passed to the sub-shell. Only exported environment variables are passed along and since `$PS1` is not an exported value it gets lost with the `exec` call (this is why the bash shell interface for activate/deactivate calls conda as `eval "$(PS1=${PS1:-} conda shell.posix activate ...)"` to correctly pass along the `$PS1` value). Now that our entry points are invoking another sub-shell that PS1 isn't being passed along to the exec's sub-shell, conda isn't able to identify the presence of a prompt.

A fix is needed as this issue likely propagates to all entry points & scripts that conda installs, i.e., I would be unable to pass environment variables to executables:

```
LIFE=42 $CONDA_PREFIX/bin/executable
```

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : None
            shell level : 0
       user config file : /Users/kodegard/.condarc
 populated config files : /Users/kodegard/.condarc
          conda version : 22.9.0
    conda-build version : 3.22.0
         python version : 3.9.12.final.0
       virtual packages : __osx=10.16=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /Users/kodegard/.conda/4.12.0/3.9  (writable)
      conda av data dir : /Users/kodegard/.conda/4.12.0/3.9/etc/conda
  conda av metadata url : None
[...]

PR #11970, opened 2022-10-13
Use `python -m conda` instead of `PREFIX/bin/conda` in shell functions

jaimergp:
### Description

Resolves #11893.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #11995, opened 2022-10-19
Use `python -m conda` instead of `PREFIX/bin/conda` in shell functions 

jaimergp:
### Description

Replaces #11970.

Resolves https://github.com/conda/conda/issues/11893.

### Checklist - did you ...


jaimergp:
This has uncovered a bug in `conda-build`. Go figure :D

conda/base/context.py
```diff
@@ -661,35 +659,44 @@ def signing_metadata_url_base(self):
...
+        bin_dir = "Scripts" if on_win else "bin"
+        exe = "conda.exe" if on_win else "conda"
+        env_vars = {
+            "CONDA_EXE": os.path.join(sys.prefix, bin_dir, exe),
+            # -I enables 'isolated' mode: no local or user paths are searched
+            # for modules (only in $PREFIX/lib/pythonX.Y/*); it also ignores
+            # all PYTHON* env vars
+            "_CE_I": "-I",
+            "_CE_M": "-m",
+            "_CE_CONDA": "conda",
```
jaimergp, line 685: I'd argue we don't rely on this environment variables anymore. They were needed before, when `__conda_exe()` and friends were defined like this:

```bash
$CONDA_EXE $_CE_M $_CE_CONDA $@
```

In normal circumstances, these variables were defined as:

* `CONDA_EXE`: path to `$BASE_PREFIX/bin/conda` (Unix) or `$BASE_PREFIX/Scripts/conda.exe` (Windows) -- in other words, the Python entry points.
* `_CE_M` and `_CE_CONDA`: empty string

In `dev` mode, they would change to
* `CONDA_EXE`: path to `python` executable in the base environment
* `_CE_M`: `-m`
* `_CE_CONDA`: `conda`

With this hack, we could do `conda.exe install ...` in normal mode and `python -m conda install ...`  in dev mode. However, since we are now ALWAYS using `python -m conda`, it makes no sense to provide this option. If anything, it can create issues because the user can accidentally delete or redefine `_CE_M` and `_CE_CONDA` and break the integration.

What I propose is to leave it like:

```
$CONDA_PYTHON_EXE $_CE_I -m conda $@
```

We still need to configure whether to use `-I` or not; we do use it for normal mode, but we omit it for dev mode so PYTHONPATH can be read. Right now, all the code in this PR assumes `_CE_M` and `_CE_CONDA` are always defined, but I'd say we just hardcode it because these do not need to be configurable.

conda/core/portability.py
```diff
@@ -203,26 +203,31 @@ def replace_long_shebang(mode, data):
...
-    #       * ' ''' (quoted space followed by empty string)
+    #    * '' (empty string)
+    #    * 'exec' "path/with spaces/to/python" "this file" "arguments"
+    #    * # ''' (inline comment with three quotes, ignored by shell)
+    if "/_h_env_placehold" in executable:
+        # This is being used during a conda-build process,
+        # which uses long prefixes on purpose. This will be replaced
+        # with the real environment prefix at install time. Do not
+        # do nothing for now.
+        return shebang
```
jaimergp, line 224: This should fix the packaging issues of newer condas, where we self-apply the workaround because the host prefix is too long.

conda/activate.py
```diff
@@ -16,7 +15,7 @@
 
 # Since we have to have configuration context here, anything imported by
 #   conda.base.context is fair game, but nothing more.
-from . import CONDA_PACKAGE_ROOT, CONDA_SOURCE_ROOT
+from . import CONDA_PACKAGE_ROOT
 from .auxlib.compat import Utf8NamedTemporaryFile
 from .base.constants import PREFIX_STATE_FILE, PACKAGE_ENV_VARS_DIR, CONDA_ENV_VARS_UNSET_VAR
 from .base.context import ROOT_ENV_NAME, context, locate_prefix_by_name
```
jaimergp, line 21: I wonder if we could move these utility functions to a different module instead of the heavy `context` object. This would make the shell integration faster to run, I believe.

Issue #11905, opened 2022-09-29
`conda_rename` Tests Fail in Feature Branches


beeankha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

All of the tests in `tests/cli/conda_main_rename.py` [fail on feature branches](https://github.com/conda/conda/actions/runs/3152698373/jobs/5128256911#step:3:536) due to the `notify_outdated_conda` warning (see below for an example):

```
___________ ERROR at setup of test_cannot_rename_active_env_by_name ____________
Traceback (most recent call last):
  File "/opt/conda-src/tests/cli/test_main_rename.py", line 33, in env_one
    run(f"conda create -n {TEST_ENV_NAME_1} -y")
  File "/opt/conda-src/conda/testing/helpers.py", line 148, in run_inprocess_conda_command
    with argv(encode_arguments(shlex_split_unicode(command))), captured(disallow_stderr) as c:
  File "/opt/conda/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/opt/conda-src/conda/testing/helpers.py", line 90, in captured
    raise Exception("Got stderr output: %s" % c.stderr)
Exception: Got stderr output: 

==> WARNING: A newer version of conda exists. <==
  current version: 4.14.0.post42+f55209477
  latest version: 22.9.0

Please update conda by running

    $ conda update -n base -c defaults conda
```

PR #11996, opened 2022-10-19
Fix error output bug in rename tests

beeankha:
Fixes https://github.com/conda/conda/issues/11905

All of the tests in `tests/cli/conda_main_rename.py` [fail on feature branches](https://github.com/conda/conda/actions/runs/3152698373/jobs/5128256911#step:3:536) due to the `notify_outdated_conda` warning.

### DIFF
--- a/tests/cli/test_main_rename.py
+++ b/tests/cli/test_main_rename.py
@@ -163,10 +163,12 @@ def test_rename_with_force(env_one, env_two):
     # Do a force rename
     run(f"conda rename -n {TEST_ENV_NAME_1} {TEST_ENV_NAME_2} --force", disallow_stderr=False)
 
+    (_, _, exit_code), _ = list_envs()
 
     assert locate_prefix_by_name(TEST_ENV_NAME_2)
     with pytest.raises(EnvironmentNameNotFound):
         locate_prefix_by_name(TEST_ENV_NAME_1)
+    assert exit_code is None
 
     # Clean up
     run(f"conda rename -n {TEST_ENV_NAME_2} {TEST_ENV_NAME_1}", disallow_stderr=False)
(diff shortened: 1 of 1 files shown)


tests/cli/test_main_rename.py
```diff
@@ -165,18 +161,15 @@ def test_rename_with_force(env_one, env_two):
...
-    run(f"conda rename -n {TEST_ENV_NAME_1} {TEST_ENV_NAME_2} --force")
+    run(f"conda rename -n {TEST_ENV_NAME_1} {TEST_ENV_NAME_2} --force", disallow_stderr=False)
 
-    (out, err, exit_code), data = list_envs()
-    result = data.get("envs", [])
 
     assert locate_prefix_by_name(TEST_ENV_NAME_2)
     with pytest.raises(EnvironmentNameNotFound):
         locate_prefix_by_name(TEST_ENV_NAME_1)
-    assert err == ""
```
sven6002, line 176: you could add the same assert error is `None`

tests/cli/test_main_rename.py
```diff
@@ -189,18 +182,15 @@ def test_rename_with_force_with_errors(env_one, env_two):
...
         assert error_message in err
-
-    (out, err, exit_code), data = list_envs()
-    result = data.get("envs", [])
+        assert exit_code == 1
 
     # Make sure both environments still exist
     assert locate_prefix_by_name(TEST_ENV_NAME_2)
     assert locate_prefix_by_name(TEST_ENV_NAME_1)
-    assert err == ""
```
sven6002, line 203: instead of removing still keep it to check for if there is an error in the environment. assert error is None

dholth: `def run_inprocess_conda_command(command, disallow_stderr: bool = True): ... return c.stdout, c.stderr, exit_code`

Since we are outputting the new message to stderr, `err` will contain that message instead of the empty string.

dholth (approved):
LGTM!

PR #12005, opened 2022-10-21
Change 'tqdm' from vendored dependency to real dependency for conda

ForgottenProgramme:
### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -43,6 +43,7 @@ requirements:
     - ruamel.yaml >=0.11.14,<0.17
     - setuptools >=31.0.1
     - toolz >=0.8.1
+    - tqdm >=4.64.1
   run_constrained:
     - conda-build >=3
     - conda-env >=2.6
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ install_requires = [
     "requests >=2.20.1",
     "ruamel.yaml >=0.11.14",
     "menuinst ; platform_system=='Windows'",
+    "tqdm >=4.64.1",
 ]
 
 def package_files(*root_directories):


dholth:
Looks good, except the lower version should be much less (possibly as low as tqdm >= 4). We only have 4.61.0 in defaults: https://metayaml-conda.fly.dev/metayaml/newest_by_platform?name=tqdm

dholth (changes requested):
I would change this to "tdmq >=4" since defaults only has `4.64.0` and there is no special reason to require the newest version. Here's how other things in defaults have declared a tqdm dependency: https://metayaml-conda.fly.dev/metayaml/reverse_dependencies?name=tqdm

kenodegard (changes requested):
Beware that we have had a lot of issues removing vendored code in favor of a dependency in the same release.

I suggest we only add tqdm as a dependency in the next release and mark the vendored code as pending deprecation per CEP8.

cc @jezdez

travishathaway (dismissed):

kenodegard (changes requested):
Needs a news snippet

dholth (approved):

beeankha (approved):

kenodegard (dismissed):

beeankha (approved):

jonashaag:
Could it be that this broke the recipe in the repo? We test Conda nightly with Mamba and a few days ago the build broke: https://github.com/mamba-org/mamba/actions/workflows/conda_nightly.yml Although I'm not quite sure why the build broke on this particular day, adding `tqdm` to the `host:` requirements seems to fix it.

jonashaag:
@jezdez @ForgottenProgramme

kenodegard:
@jonashaag I wonder if this is potentially related to the need to introduce the isolated mode in conda-build:
- https://github.com/conda/conda-build/pull/4604
- https://github.com/conda/conda-build/pull/4625

> If this is the case, does the Conda CI not test the recipe?

What do you mean?

Let's discuss further in a new issue for better visibility

PR #12014, opened 2022-10-25
Fix crashes in MatchSpec, VersionSpec parsing

pkmooreanaconda:
### Description
This PR attempts to address the set of crashes referenced in https://github.com/conda/conda/issues/11999.  These fixes are intended to be minimal in order to reduce the chances of unintended MatchSpec behavior changes.  The changes can be split into three commits:

1. Regex-In-MatchSpec related fixes:  These fixes wrap attempts to compile regular expressions built from user input in order to catch regular expressions and raise InvalidMatchSpec with an informative error message instead.

2. version.py fix: Fix a bug in handling the "local version separator (+)."  I've made a minimal correction to this code which makes versions like "+", "+a", "+1", and "+1.2" invalid.  These cases are invalid because it doesn't make sense to only specify a local version.  Cases like "1.2+", "1+", "a+" are already caught as invalid.

3. "==" fix: "==" was unintentionally treated differently than "<=" and ">=" due to some weird translation code ([link](https://github.com/conda/conda/blob/352bcb3f1667c4aea9d1b04cf02a0e1af468dbd3/conda/models/match_spec.py#L701)).  This meant that cases line "numpy<=" and "numpy>=" would raise an error while "numpy==" would crash.  I made a small fix to bypass the translation code in the "numpy==" case so invalid version "==" can reach the existing error handling code like its friends.  The way this is handled is very weird.  We allow the strings "<=", ">=" and (now) "==" to be considered versions for a while after parsing and catch them later on with an "invalid operator" error.  But that's a job for another time.

### Checklist


kenodegard:
@conda-bot check

pkmooreanaconda:
pre-commit.ci autofix

tests/models/test_version.py
```diff
@@ -301,6 +301,28 @@ def test_invalid_version_specs(self):
...
+            VersionSpec("+1.2+")
+        with pytest.raises(InvalidVersionSpec):
+            VersionSpec("++")
+        # Fuzzer-identified crasher:
+        with pytest.raises(InvalidVersionSpec):
+            VersionSpec("c +, 0/|0 *")
+
+        # Test for mishandling of '==' without a version number
+        with pytest.raises(InvalidVersionSpec):
+            VersionSpec("==")
```
chenghlee, line 320: Should we add this test case?
```
with pytest.raises(InvalidVersionSpec):
    VersionSpec("=")
```

If `==` is invalid, then `=` (single equals) should be as well.

chenghlee (changes requested):

tests/models/test_version.py
pkmooreanaconda: That case was "working" before so I didn't change it.  Current behavior is:
 ```conda search "numpy="``` is equivalent to ```conda search "numpy"```

It could be a breaking change for some people (but I agree with your assessment of how it SHOULD work).

kenodegard: Let's open a spike to investigate this further.

conda/models/match_spec.py
```diff
@@ -821,11 +834,16 @@ def __init__(self, value):
...
+        try:
+            if value.startswith('^') and value.endswith('$'):
+                self._re_match = re.compile(value).match
+            elif '*' in value:
+                value = re.escape(value).replace('\\*', r'.*')
+                self._re_match = re.compile(r'^(?:%s)$' % value).match
+        except re.error as e:
+            raise InvalidMatchSpec(value,
+                                   'Contains an invalid regular expression. "%s"' % e)
+            pass
```
kenodegard, line 846:
```suggestion
            raise InvalidMatchSpec(value, f"Contains an invalid regular expression. '{e}'")
```

conda/models/match_spec.py
```diff
@@ -936,13 +954,17 @@ class ChannelMatch(GlobStrMatch):
...
+            if isinstance(value, str):
+                if value.startswith('^') and value.endswith('$'):
+                    self._re_match = re.compile(value).match
+                elif '*' in value:
+                    self._re_match = re.compile(r'^(?:%s)$' % value.replace('*', r'.*')).match
+                else:
+                    value = Channel(value)
+        except re.error as e:
+            raise InvalidMatchSpec(value,
+                                   'Contains an invalid regular expression. "%s"' % e)
```
kenodegard, line 967:
```suggestion
            raise InvalidMatchSpec(value, f"Contains an invalid regular expression. '{e}'")
```

kenodegard (changes requested):
Minor nits, since we don't want to change behavior here let's move the `=` discussion into a new issue/spike.

tests/models/test_version.py
kenodegard: See https://github.com/conda/conda/issues/12046

chenghlee (approved):

PR #12019, opened 2022-10-26
tests: use readline instead of regex capture to avoid deadlock

dholth:
### Description

Test change. Does pexpect buffer size affect performance? (larger = slower)?

Appears to hang if the regex capture is larger than the buffer size, causing failures in osx-64 CI. readline() is better than increasing the buffer indefinitely.

Do we have to worry about multi-line env var's?

### Checklist - did you ...

### DIFF
--- a/dev/macos/setup.sh
+++ b/dev/macos/setup.sh
@@ -2,9 +2,12 @@
 set -ex
 
 # Download the Minio server, needed for S3 tests
-curl -LO https://dl.minio.io/server/minio/release/darwin-amd64/minio
+if [[ ! -f minio ]]
+then
+    curl -LO https://dl.minio.io/server/minio/release/darwin-amd64/minio
+fi
 chmod +x minio
-sudo mv minio /usr/local/bin/minio
+sudo cp minio /usr/local/bin/minio
 
 # restoring the default for changeps1 to have parity with dev
 conda config --set changeps1 true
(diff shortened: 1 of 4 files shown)


dev/macos/setup.sh
```diff
@@ -2,9 +2,12 @@
...
 
 # Download the Minio server, needed for S3 tests
-curl -LO https://dl.minio.io/server/minio/release/darwin-amd64/minio
+if [[ ! -f minio ]]
+then
+    curl -LO https://dl.minio.io/server/minio/release/darwin-amd64/minio
+fi
 chmod +x minio
-sudo mv minio /usr/local/bin/minio
+sudo cp minio /usr/local/bin/minio
```
dholth, line 10: keep around so we can cache it

tests/test_activate.py
```diff
@@ -108,6 +108,8 @@
 }
 '''
 
+HDF5_VERSION = "1.12.1"
```
dholth, line 111: available on osx-arm64 et al

tests/test_activate.py
```diff
@@ -1976,18 +1978,17 @@ def get_env_var(self, env_var, default=None):
             self.sendline("@echo %%%s%%" % env_var)
             self.expect("@echo %%%s%%\r\n([^\r]*)\r" % env_var)
             value = self.p.match.groups()[0]
-        elif self.shell_name == 'powershell':
+        elif self.shell_name in ('powershell', 'pwsh'):
```
dholth, line 1981: it is called pwsh on mac and IDK on windows, but, the last else: sortof-worked for powershell

dholth:
In summary, not rock solid, but notably better.

PR #12023, opened 2022-10-27
Convert `conda.core.initialize` unittest to pytest

kenodegard:
### Description

Flatten unittest class into pytest functions and refacted setup/teardown as fixture.

Also contains minimal reformatting of c-style string formatting.

Prework for adding new tests in #11865.

### Checklist - did you ...

### DIFF
--- a/tests/core/test_initialize.py
+++ b/tests/core/test_initialize.py
@@ -40,9 +40,6 @@ from conda.models.enums import LinkType
 from conda.testing.helpers import tempdir
 
 
-log = getLogger(__name__)
-
-
 @pytest.fixture
 def verbose():
     orig_verbosity = os.environ.get("CONDA_VERBOSITY")
--- a/tests/core/test_initialize.py
+++ b/tests/core/test_initialize.py
@@ -6,7 +6,6 @@ import ntpath
 import os
 from os.path import abspath, dirname, isfile, join, realpath
 import sys
-from unittest import TestCase
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/core/test_initialize.py
```diff
@@ -19,709 +17,767 @@
...
-    from unittest.mock import Mock, patch
-except ImportError:
-    from unittest.mock import Mock, patch
-
 
 log = getLogger(__name__)
 
 
-class InitializeTests(TestCase):
+@pytest.fixture
```
dholth, line 46: Try adding `fixture(scope="module")` to run this code once, instead of once per test.

dholth (approved):

tests/core/test_initialize.py
kenodegard: I wasn't sure if we needed the context to be reset once or per function, it looks like everything is good when I change it

dholth: I checked the docs, and setUp and tearDown are run after each method. Sorry for the inconvenience.

It does look like some of the test alter context, and would benefit from the `reset_context()` call.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

tests/core/test_initialize.py
```diff
@@ -19,964 +18,1065 @@
...
 from conda.models.enums import LinkType
 from conda.testing.helpers import tempdir
 
-try:
-    from unittest.mock import Mock, patch
-except ImportError:
-    from unittest.mock import Mock, patch
-
 
 log = getLogger(__name__)
```
sven6002, line 43: Could be removed..

sven6002 (approved):
Refactoring is a possible ticket for future.. but it not deterring the working of the test cases, in test_initialize.py

tests/core/test_initialize.py
```diff
@@ -1,14 +1,12 @@
...
-
-
 from distutils.sysconfig import get_python_lib
 from logging import getLogger
 import ntpath
 import os
 from os.path import abspath, dirname, isfile, join, realpath
 import sys
-from unittest import TestCase
+from unittest.mock import patch
```
dholth, line 9: there is a pytest fixture called `mocker`

dholth: It is the unittest.mock interface but it cleans up automatically. Example https://github.com/conda/conda-package-handling/blob/conda-package-handling-2.0/tests/test_cli.py#L30

dholth (approved):
LGTM

Issue #11825, opened 2022-09-09
Add extra configuration parameters to channels in conda configuration


travishathaway:
### What is the idea?

In order for to support authentication methods in a way that does not require us to either put a token or a username/password in the URL (e.g. anaconda.org token authentication method and basic HTTP authentication), we need a way to make the channel aware of which "fetch" backend to use. These backends will be implemented in the future by various fetch plugins.

How do we do that? We add a new configuration parameter called "channel_settings" that will hold additional setting for individual channels:

Here's what that looks like:

```yaml
channels:
  - defaults
  - https://package-corp.com
channel_settings:
  - channel: https://package-corp.com
    fetch: basic-auth
```

Using this method, we'll be able to switch out fetch backends on a per channel basis.

This method will allow us to attach arbitrary parameters to individual channels, so not only will the immediate use case for "fetch" be covered, but future use cases too.

### More background information

Relevant issues:

- #11600 
- #11696 

### Tasks

- [x] Update conda configuration so that it supports these extra configuration parameters for channels
- [x] Create a way for us to inspect this config. These extra parameters should show up when we run `conda config --show` and `conda config --get channels`
- [x] Add new tests to make sure that we can properly access these new parameters

wolfv:
Interesting idea. I think there are already some other configuration ways to override some channel-specific settings, it might be interesting to consolidate them (`custom_channel` and `custom_multichannel`).

Also, in the mamba project we implement the same "rc"-file and we read the conda one, so it might be cool to properly specify what options you'll add and what additional uses you envision in the future.

In the powerloader PR we added a `mirrors` option that looks like so:

```
mirrors:
  conda-forge:
     - https://mymirror
     - oci://somemirror
     - ...
```

However, that could also be a key to the entry in this dictionary here.

travishathaway:
Hi @wolfv,

I think that's a good idea. This could be the first step in simplifying conda's configuration.

As we move to adding more plugin hooks in conda, we want to keep these configurations fairly flexible because anticipating every future use case would be difficult. So, in that spirit it is kind of difficult to say exactly what will be there. Channel authentication is our immediate use case though.

We can say that a schema will be enforced, and the channel options will be simple key, value pairs at first (`str, str`). Would that suffice and allow you to predictably continue to parse these configuration files?

travishathaway:
The issue description has been updated and is worth reading if you have been following along 👍

Issue #7022, opened 2018-03-12
Certain custom_multichannels entries produce warnings


mbargull:
**I'm submitting a...**
  - [x] bug report
  - [ ] feature request

## Current Behavior

If there are channel entries in a `custom_multichannels` configuration that are suffixes of another channel, `conda` prints a warning:
```
WARNING: The remote server could not find the noarch directory for the
requested channel with url: https://conda.anaconda.org/bioconda/label/deprecated/conda.anaconda.org/bioconda
```
Triggered by the prefix check at https://github.com/conda/conda/blob/4.4.11/conda/models/channel.py#L413.
Despite that warning, it looks like the correct channel URLs are used since subsequent operations succeed.

### Steps to Reproduce

```
$ docker run --rm -it continuumio/miniconda3
root@f5d1379434b5:/# conda update -qy conda >/dev/null
root@f5d1379434b5:/# conda --version
conda 4.4.11
root@f5d1379434b5:/# cat > ~/.condarc <<'EOF'
> custom_multichannels:
>   bioconda-repro:
>     - bioconda
>     - conda-forge
>     - defaults
>     - bioconda/label/deprecated
>     - bioconda/label/broken
> EOF
root@f5d1379434b5:/# conda search --info bioconda-repro::julia=0.3.11
Loading channels: | 
WARNING: The remote server could not find the noarch directory for the
requested channel with url: https://conda.anaconda.org/bioconda/label/deprecated/conda.anaconda.org/bioconda

It is possible you have given conda an invalid channel. Please double-check
your conda configuration using `conda config --show`.

If the requested url is in fact a valid conda channel, please request that the
channel administrator create `noarch/repodata.json` and associated
`noarch/repodata.json.bz2` files, even if `noarch/repodata.json` is empty.
$ mkdir noarch
$ echo '{}' > noarch/repodata.json
$ bzip2 -k noarch/repodata.json
done
julia 0.3.11 0
--------------
file name   : julia-0.3.11-0.tar.bz2
name        : julia
version     : 0.3.11
build string: 0
build number: 0
size        : 28.6 MB
arch        : x86_64
constrains  : ()
platform    : Platform.linux
license     : MIT
subdir      : linux-64
url         : https://conda.anaconda.org/bioconda/label/deprecated/linux-64/julia-0.3.11-0.tar.bz2
md5         : b8afb4685ec0cd983f78f8dc66fe009f
dependencies: 
```

## Expected Behavior

[...]

PR #12033, opened 2022-10-29
Add ability to add arbitrary configuration parameters to channels

travishathaway:
Please see this issue for a full explanation of why this pull request exists:

- https://github.com/conda/conda/issues/11825

#### No full support for `conda config --set channels`

This change does not fully support setting channels with extra configuration parameters via the command line. I decided not to do this because it goes against the "MVP" spirit of this pull request. I will be making another issue for it though so that we can support this in the future. I'm not sure how useful this would be though.

#### Not loudly advertised in our documentation

I think it is best for this to go out more-or-less silently at first. But, as pointed out in this comment:

- https://github.com/conda/conda/issues/11825#issuecomment-1296181916

this pull request could allow us to consolidate existing channel related configuration parameters found elsewhere in our configuration (and hopefully make it easier to understand). I will be making a spike issue to explore doing this.

#### Risks involved

These changes are what I consider to be high-risk. We need to make sure that the changes I have introduced are backwards compatible and that they do not break when reading existing configuration files or environment variables. Please take extra care to make sure there are no edge cases I've forgotten while writing my tests.

Suggestions for additional tests that I can write to make these changes as robust as possible are greatly welcomed.


conda/common/configuration.py
```diff
@@ -1101,6 +1156,110 @@ def load(self, name, match):
...
             match.valueflags(self._element_type),
+            validation=self._validation,
+        )
+
+
+#: These are the allowable types for a UnionSequenceParameter
+UnionElementTypes = Union[MapParameter, SequenceParameter, PrimitiveParameter]
+
+
+class UnionSequenceParameter(Parameter):
```
kenodegard, line 1167: What about abstracting the existing `SequenceParameter`? One `element_type` is homogenous, multiple is heterogeneous.

travishathaway: This could definitely be done with what I have right now. I will investigate. It's kind of weird to have two classes which essentially handle the same thing.

travishathaway: @kenodegard, thanks for this suggestion. I have performed a refactor, and we now only have a single `SequenceParameter` class.

tests/common/test_configuration.py
```diff
@@ -578,11 +583,13 @@ def test_expanded_variables(self):
     def test_nested(self):
         config = SampleConfiguration()._set_raw_data(
             load_from_string_data('nestedFile1', 'nestedFile2'))
+        # breakpoint()
```
beeankha, line 586:
```suggestion
```

travishathaway: Thanks 😊

tests/common/test_configuration.py
```diff
@@ -600,3 +661,14 @@ def test_object(self):
...
+    def test_invalid_config_for_sequence(self):
+        config = SampleConfiguration()._set_raw_data(
+            load_from_string_data("invalid_config_object")
+        )
+        valid_types = "Sequence[PrimitiveParameter]"
+
+        with pytest.raises(InvalidElementTypeError) as exc_info:
+            assert config.env_var_list == ()
+
+        assert exc_info.value.valid_types == valid_types
```
beeankha, line 674: 🎉

beeankha (dismissed):

conda/common/configuration.py
```diff
@@ -1049,51 +1060,130 @@ def load(self, name, match):
...
+    More examples of what this looks like are available in ``tests/common/test_configuration.py``
     """
-    _type = tuple
 
-    def __init__(self, element_type, default=(), validation=None, string_delimiter=','):
+    def __init__(
+        self,
+        default: Sequence | None = None,
+        primitive_type: PrimitiveParameter = None,
+        map_type: SequenceElementTypes = None,
```
kenodegard, line 1081: I think we should make this even more abstract. This is currently too specific to what we are trying to accomplish with the `channels` context argument. Something like:

```
SequenceElementTypes = PrimitiveParameter | MapParameter | SequenceParameter

class SequenceParameter(Parameter):
    def __init__(
        self,
        element_type: SequenceElementTypes | list[SequenceElementTypes],
        default: Sequence | None = None,
        validation=None,
        string_delimiter: str = ",",
    ):
        ...
```

kenodegard (changes requested):

conda/common/configuration.py
travishathaway: Hi @kenodegard,

I spent a good amount of time deliberating your suggestion and going back over a lot of the code contained within `conda/common/configuration.py` and `conda/base/context.py`.

What I concluded is that although it's alluring to overload the `element_type` parameter, this parameter really only wants to point at a single type. Let me explain what I mean with an example...

Suppose we have the following configuration defined:

```python
from conda.common.configuration import (
    Configuration,
    SequenceParameter,
    PrimitiveParameter,
   MapParameter,
    ParameterLoader, 

class MyConfig(Configuration):
    complex_seq = ParameterLoader(
        SequenceParameter(element_type=(
            PrimitiveParameter(0, element_type=int),
            PrimitiveParameter("", element_type=str),
            MapParameter(PrimitiveParameter("", element_type=str))
        ))
    )
```

We've added in `PrimitiveParameter` twice, which may cause undefined behavior. When we run across a number, do we parse it as a string or an integer? Defining `element_type` as a sequence of `Parameter` types leaves us open to this type of ambiguity in our API.

But, what if we just want to enforce that users only pass in one of each type. In this case, we would have to implement some ugly validation logic within these objects to verify that. There must be a better way...

This is why I opted to create two new keyword arguments for this `Parameter` type. To avoid ambiguities like in the above example, we want to establish a one-to-one relation between what is found in the configuration and how it ultimately ends up getting parsed.

I would gladly add an additional `sequence_type` so that it supports all possible values, but I figured that would be a little overkill for this pull request. I hope that I have documented the class well enough that a would-be future developer wouldn't have a difficult time adding this behavior in the future if it is ever needed.

Let me know what you think and if I have adequately convinced you 😎 .  I'm always happy to discuss this code and the rationale behind the changes more.

kenodegard: I've always preferred the (sometimes extreme) abstractions for stuff like this since I've always benefitted from it eventually. 🤷🏼‍♂️

I would rely on the order of `element_types` to dictate the parsing precedence, so parsing `"0"` will always return a `0`. If the `PrimitiveParameter` were switched, we'd expect a `"0"` to always be `"0"`.

This is so close to implementing true homogeneous and heterogenous sequence types that not implementing it IMO results in a more complicated solution that will be harder to document.

But I don't care to die on this hill. I'll let others chime in and you can make a decision from there.

travishathaway: @kenodegard ,

I went back and made some revisions. I have more-or-less implemented the type of API you would like for `SequenceParameter` and now we can have a combination of primitive values, maps and sequences.

I opted to use a `element_types` parameter that is just a sequence rather than the sequence or single value function signature you had suggested with the `element_type` parameter.

It still works the same "under the hood", but I've made sure to make class attributes like `_map_type`, `_primitive_type` and `_sequence_type` protected so people know they are not supposed to use these directly.

Thanks for your help in discussing/evolving this pull request with me! Let me know if there's still anything else that needs to be resolved, or whether the newest batch of changes aligns with what you had imagined.

travishathaway: @kenodegard ,

You can disregard part of that previous comment. After seeing the code diff once more after my changes, I realized it would be better to have that `element_type` argument and give it the ability to accept either a single Parameter type or a sequence of them.

This should hopefully be everything you are looking for now.

Cheers ✌️

news/12033-adds-arbitrary-config-parameters-to-channels
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* In configuration files, channels can now be defined as a either a list of strings or a list of strings and mappings.
+  This allows us to attach arbitrary configuration parameters to individual channels.
```
beeankha, line 4:
```suggestion
  This allows us to attach arbitrary configuration parameters to individual channels. (#12033)
```

beeankha (approved):
Other than the need to add a PR (or issue) number to the news snippet, this LGTM!

conda/base/context.py
```diff
@@ -792,16 +837,37 @@ def channels(self):
...
+        channel_params = {}
+
+        for channel in self._channels:
+            if isinstance(channel, Mapping):
+                keys = channel.keys()
+                if len(keys) > 0:
+                    name, *_ = keys
+                    channel_params[name] = channel.get(name)
+            else:
+                channel_params[channel] = frozendict()
```
kenodegard, line 868:
```suggestion
        for channel in filter(None, self._channels):
            if isinstance(channel, Mapping):
                name = next(iter(channel))
                channel_params[name] = channel.get(name)
            else:
                channel_params[channel] = frozendict()
```

conda/testing/helpers.py
```diff
@@ -836,3 +839,20 @@ def convert_to_dist_str(solution):
...
+def temp_context(condarc: str) -> Generator[Context, None, None]:
+    """
+    Context manager that resets the context with the provide `condarc` string.
+    This string should be in the YAML format.
+    """
+    context_obj = reset_context(())
+    rd = dict(
+        testdata=YamlRawParameter.make_raw_parameters("testdata", yaml_round_trip_load(condarc))
+    )
+    context._set_raw_data(rd)
```
kenodegard, line 854:
```suggestion
    testdata = YamlRawParameter.make_raw_parameters("testdata", yaml_round_trip_load(condarc))
    context._set_raw_data({"testdata": testdata})
```

conda/common/configuration.py
```diff
@@ -1432,7 +1540,10 @@ def describe_parameter(self, parameter_name):
         if not isiterable(et):
             et = [et]
 
-        if isinstance(parameter._element_type, Parameter):
+        # If it doesn't have  a __name__ attribute it's an object not a class
+        has_name = all(hasattr(element, "__name__") for element in et)
```
kenodegard, line 1544: What about using `inspect.isclass` instead?

kenodegard (changes requested):
I like the switch back to using `element_type`, this makes it more backward- & forward-compatible. Thanks for taking the time! A few suggestions and some questions.

conda/base/context.py
travishathaway: Nice usage of `filter`, `next` and `iter` 👍 🔥

conda/common/configuration.py
```diff
@@ -1052,51 +1063,153 @@ def load(self, name, match):
...
+        return element_type
 
+    @staticmethod
+    def _pick_element_type(
+        element_type: type[SequenceElementTypes], elements: Sequence[SequenceElementTypes]
+    ) -> SequenceElementTypes | None:
+        """Iterates over ``elements`` and picks the first ``element_type`` encountered"""
+        for elm in elements:
+            if isinstance(elm, element_type):
+                return elm
```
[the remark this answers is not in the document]
travishathaway: I like the idea of moving this to `conda.common.iterators`.

This is right on the border of readability for a one liner in my opinion. I usually err on the side of providing a nice alias for these types of things to ease the cognitive load for future maintainers.

conda/common/configuration.py
travishathaway: I'll check that out. Thanks!

tests/common/test_configuration.py
```diff
@@ -32,8 +45,10 @@
           - bugs
           - daffy
           - tweety
-    """),
-    'file2': dals("""
+    """
```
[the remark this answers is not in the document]
travishathaway: This was most likely darker auto-formatting or complaining about things. I remember have to make a bunch of weird updates to this file because it wasn't passing darker checks.

I'll give it another shot though.

conda/common/configuration.py
```diff
@@ -1432,7 +1533,11 @@ def describe_parameter(self, parameter_name):
...
 
-        if isinstance(parameter._element_type, Parameter):
+        is_class = all(inspect.isclass(element) for element in et)
+
+        # if it's not a Parameter object, we make sure to also exclude all type variables
+        # (i.e. classes) from the next step
+        if isinstance(parameter._element_type, Parameter) or not is_class:
             element_types = tuple(
                 _et.__class__.__name__.lower().replace("parameter", "") for _et in et)
         else:
```
kenodegard, line 1543: Hm, after re-reading this, can we rewrite the whole thing? Isn't the goal to distinguish between classes vs. instances and do different things?

```python
element_types = tuple(
    # if a class is found, return its name as is
    element.__name__
    if inspect.isclass(element)
    # if an instance is found, return a simplified name
    else element.__class__.__name__.lower().replace("parameter", "")
    for element in et
)
```

conda/common/iterators.py
```diff
@@ -18,3 +21,24 @@ def groupby_to_dict(keyfunc, sequence):
     for key, group in itertools.groupby(sequence, keyfunc):
         result[key].extend(group)
     return dict(result)
+
+
+T = TypeVar("T")
+
+
+def next_instance(objects: Iterable[T], object_type: type[T]) -> T | None:
```
kenodegard, line 29: The typing is too strict, there is no requirement that the iterable elements must be an instance of `T`:

```suggestion
def next_instance(objects: Iterable[object], object_type: type[T] | tuple[type[T], ...]) -> T | None:
```

We could also reuse what [typeshed](https://github.com/python/typeshed/blob/7c15e6998f28272a5e308524dadfc5aa00aff08b/stdlib/builtins.pyi#L1327-L1332) does?

What do you think about replacing `object_type` with `classinfo` to mirror `isinstance`/`issubclass`?

travishathaway: Good catch. I think I will just re-write it as:

```python
from typing import Any

# snip, snip...

def next_instance(objects: Iterable[Any], object_type: type[T]) -> T | None:
```

That way it will be more generic.

I wouldn't want the second argument to accept multiple values though (i.e. `tuple[type[T], ...]`). The function is supposed to return the _next_ instance of `type`, which means just one single instance. Being able to define multiple types just doesn't make much sense to me. Would it just return the first type found in the tuple? If that's the case, why not just pass in a single type to begin with. Hope that makes sense.

kenodegard (approved):

conda/testing/helpers.py
```diff
@@ -836,3 +839,18 @@ def convert_to_dist_str(solution):
 @pytest.fixture()
 def solver_class():
     return context.plugin_manager.get_cached_solver_backend()
+
+
+@contextmanager
+def temp_context(condarc: str) -> Generator[Context, None, None]:
```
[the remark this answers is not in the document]
jezdez: Last thought, could this also be written as a pytest fixture?

jezdez (changes requested):
A few questions and naming feedback, but this looks almost ready to go! Nice work @travishathaway

conda/base/context.py
```diff
@@ -792,16 +837,33 @@ def channels(self):
...
+
+    @property
+    def channel_parameters(self) -> frozendict:
+        """
+        Return parameters associated with all channels if there are any.
+        It will be an empty frozendict if there are no parameters.
+        """
+        channel_params = {}
+
+        for channel in filter(None, self._channels):
```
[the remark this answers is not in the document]
travishathaway: I think this may have been a little too defensive coding on my part. Pretty sure this could be removed.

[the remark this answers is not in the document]
travishathaway: I mean the only thing I could think of is the following example:

```yaml
channels:
  - defaults
  - ""
```

But that is pretty silly and contrived 😂

conda/common/configuration.py
```diff
@@ -206,7 +210,7 @@ def __repr__(self):
         return str(vars(self))
 
     @abstractmethod
-    def value(self, parameter_obj):
+    def value(self, parameter_obj=None):
```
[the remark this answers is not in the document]
travishathaway: This was to keep my type checker in PyCharm from showing an error. I can switch it back to keep it consistent.

conda/common/configuration.py
```diff
@@ -1359,12 +1465,7 @@ def check_source(self, source):
 
             if match is not None:
                 loaded_parameter = parameter.type.load(key, match)
-                # untyped_value = loaded_parameter.value
-                # if untyped_value is None:
-                #     if isinstance(parameter, SequenceLoadedParameter):
-                #         untyped_value = ()
-                #     elif isinstance(parameter, MapLoadedParameter):
-                #         untyped_value = {}
+
```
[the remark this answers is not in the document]
travishathaway: My dad always told me: commenting code out!?!? Boy don't you know that's why God created revision control! 😂  (my dad never said that).

conda/common/iterators.py
```diff
@@ -18,3 +21,24 @@ def groupby_to_dict(keyfunc, sequence):
     for key, group in itertools.groupby(sequence, keyfunc):
         result[key].extend(group)
     return dict(result)
+
+
+T = TypeVar("T")
+
+
+def next_instance(objects: Iterable[Any], object_type: type[T]) -> T | None:
```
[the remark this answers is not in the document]
travishathaway: Thanks for asking.

In type hinting, the `T` would simply be a generic for matching against an instantiated object. What we want instead is a `type` generic (that's what `type[T]` is). A `type` is precisely what `isinstance` is expecting as a second argument. 

So, in plain English the function signature translates roughly to: `object_type` should be the `type` we want to look for and then we return that instantiated type or `None`. The type that you pass in has to match the type that your return.

The Generic magic happens because it applies to multiple types (e.g. `str` -> `str`, `list` -> `list).

Hope that helps!

conda/testing/helpers.py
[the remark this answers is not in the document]
travishathaway: @jezdez, do you know if it's possible to pass arguments to pytest fixtures? The reason it's a context manager is that I wanted the ability to add whatever YAML string I wanted.

conda/common/configuration.py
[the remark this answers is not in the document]
jezdez: 🤣

conda/common/configuration.py
```diff
@@ -1052,51 +1065,144 @@ def load(self, name, match):
...
+    More examples of what this looks like are available in ``tests/common/test_configuration.py``
     """
-    _type = tuple
 
-    def __init__(self, element_type, default=(), validation=None, string_delimiter=','):
+    def __init__(
+        self,
+        element_type: SequenceElementTypes | Sequence[SequenceElementTypes],
+        default: Sequence | None = None,
```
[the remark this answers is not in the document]
travishathaway: @jezdez, we do this in the `load` method below. The class that ultimately determines the value that, for example, the `Context` object gets is the `SequenceLoadedParameter` class. Fun stuff, ehh? 😅.

Down there, we have a `if value is None` check that guards against this (line 1187)

Just to be extra certain, I will change this to default to an empty tuple 👍.

conda/common/configuration.py
```diff
@@ -1052,51 +1065,144 @@ def load(self, name, match):
...
-    Parameter type for a Configuration class that holds a sequence (i.e. list) of Parameters.
+    The SequenceParameter class allows you to define a flexible sequence that
+    can accept multiple data types.
+
+    You can put up to three different values in ``element_types`` for either
+    PrimitiveParameter, MapParameter or SequenceParameter types. Combining different
+    types could, for example, allow you to parse YAML files with mixed data type
+    sequences (e.g. strings and mappings or strings, lists and mappings).
+
+    More examples of what this looks like are available in ``tests/common/test_configuration.py``
```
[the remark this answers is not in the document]
travishathaway: @travishathaway ❤️ s doc-strings 😁

conda/testing/helpers.py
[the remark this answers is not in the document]
jezdez: Yeah, this is possible by using [indirect parameterization](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization), which is useful, especially if you're testing the same thing over and over and want to parametrize the test function anyway:

```python

@pytest.fixture
def context_from_yaml(request):
    """
    Fixture that resets the context with the provided parameter.
    This string should be in the YAML format.
    """
    test_context = reset_context(())
    loaded_yaml = YamlRawParameter.make_raw_parameters(
         "testdata",
         yaml_round_trip_load(request.param),
    )
    context._set_raw_data({"testdata": loaded_yaml})

    yield test_context

    reset_context()


@pytest.mark.parametrize("context_from_yaml", ["<YAML CONTENT>"], indirect=True)
def test_custom_condarc(context_from_yaml):
    ...

```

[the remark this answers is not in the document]
jezdez: Also see the example for using indirect parameters only on a few of the parameters of a test function: https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

conda/base/context.py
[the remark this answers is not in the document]
jezdez: Yeah, I think that'd be a pretty odd edge case :)

conda/common/configuration.py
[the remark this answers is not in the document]
jezdez: I'm not sure I follow PyCharm's reason to require a default for this case TBH

conda/testing/helpers.py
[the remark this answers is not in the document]
travishathaway: Thanks for this trick!

conda/common/configuration.py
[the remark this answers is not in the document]
travishathaway: @jezdez, I actually went back and looked at this more today. It was more than just a type checker thing. It has to do with the way that `EnvRawParameter` is handled differently than everything else.

`EnvRawParameter`'s implementation of `value` is the only one which actually does anything with the `parameter_obj` value.  Presumably, it does this to handle the "splitting" of environment variables with a delimiter. Interestingly enough, passing in `None` here doesn't break any of that.

To be extra certain, I also created a new test with a custom delimiter, and still everything works. So, I'm not exactly certain what is going on. You can checkout this method for yourself here:

https://github.com/conda/conda/blob/c9aa8ff6e365692e55961fa7477f06c3f0fc5178/conda/common/configuration.py#L230

I have since rewritten the method, `get_element_type_from_value` to explicitly pass in `None` and removed the default values of `None` for the other `value` definitions.

travishathaway:
@jezdez, @kenodegard  and @beeankha this one is ready for another look over.

news/12033-adds-arbitrary-config-parameters-to-channels
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* In configuration files, channels can now be defined as a either a list of strings or a list of strings and mappings.
```
beeankha, line 3:
```suggestion
* In configuration files, channels can now be defined as either a list of strings or a list of strings and mappings.
```

Small nitpick here!

beeankha (dismissed):

tests/base/test_context.py
```diff
@@ -534,3 +535,78 @@ def test_validate_prefix_name(prefix, allow_base, mock_return_values, expected):
...
+
+
+BASIC_CONDARC = "channels: testing"
+
+
+@pytest.mark.parametrize("context_from_yaml", [BASIC_CONDARC], indirect=True)
+def test_condarc_file_with_single_channel(context_from_yaml):
+    """
+    Tests the case where "channels" is defined as a single string value
+    """
```
beeankha, line 608: Really great test docstrings here!! Thank you 👍

news/12033-adds-arbitrary-config-parameters-to-channels
travishathaway: Thanks!

beeankha (dismissed):

conda/common/iterators.py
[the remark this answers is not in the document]
mbargull: Please note that `type.__getitem__` is a [Python 3.9+ functionality](https://docs.python.org/3/library/stdtypes.html#generic-alias-type).
This should still not cause trouble at runtime for 3.7/3.8 due to the [postponed evaluation](https://peps.python.org/pep-0563/) brought in via `from __future__ import annotations`.

However, beware that using features/properties of later versions can give a false sense of confirmation from a type checker. Usually it's good to also run a type checker for the lowest supported version.
(^-- That is a very general note. In this certain case it's not a big deal, probably. But if/when we have type checking as part of the CI, it will likely complain.)

tests/base/test_context.py
```diff
@@ -534,3 +535,78 @@ def test_validate_prefix_name(prefix, allow_base, mock_return_values, expected):
...
+
+#: Invalid condarc contain bad values for the parameters, specifically "bad_param"
+INVALID_CONDARC_WITH_CHANNEL_PARAMETERS_1 = """
+channels:
+  - defaults
+  - conda-forge
+  - http://localhost:
+      bad_param:
+        - test_one  # <-- lists are not allowed here
+        - test_two
```
mbargull, line 558: A test with multiple keys would be good too, e.g.:
```yaml
  - http://localhost:
      authentication: test-auth
      test_param: test-value
    http://::1:
      authentication: test-auth
      test_param: test-value
```

mbargull:
> #### Risks involved
> 
> These changes are what I consider to be high-risk. We need to make sure that the changes I have introduced are backwards compatible and that they do not break when reading existing configuration files or environment variables. Please take extra care to make sure there are no edge cases I've forgotten while writing my tests.

I agree these changes are high-risk.
We also have to keep in mind that "reading existing configuration" with a new version is not the only direction it can go.
There is also the case of reading new configuration files with older versions and/or alternative implementations (e.g. -- cc @wolfv -- `micromamba`; plus maybe also even auxiliary tools that are able to read `.condarc` files).
For better or worse, `conda` has a fixed set of paths (`conda.base.constants.SEARCH_PATH`) it reads its configuration from. If any of those paths' config files (which might be non-writable by the current user) have new/incompatible entries then an older version/alternative implementation can be rendered non-functional by that single file. (These scenarios can be easily constructed, e.g.: A working group has a main Conda installation at `/var/lib/conda` that uses these new features and a member who gets back from extended leave (e.g., paternal leave, etc. pp.) wants to work with their slightly outdated Conda installation in their home directory. -- IDK how people/groups use Conda; but back when the work from https://www.anaconda.com/blog/conda-configuration-engine-power-users was done, it must've had some use cases/customers...).

Also, the error reporting on broken/incompatible config files is not great. E.g., with the proposed test case
```yaml
channels:
  - defaults
  - conda-forge
  - http://localhost:
      authentication: test-auth
      test_param: test-value
```
users of (then) outdated versions/alternative implementations currently get
<details><summary>

```sh
# conda --version
conda 22.11.1
# conda config --show
```
</summary>

```python
WARNING conda.exceptions:print_unexpected_error_report(1240): KeyError('pkgs_dirs')
  File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1118, in __call__
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
    exit_code = do_call(args, p)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 91, in do_call
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 33, in execute
    execute_config(args, parser)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 162, in execute_config
    d = {key: getattr(context, key) for key in paramater_names}
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 162, in <dictcomp>
    d = {key: getattr(context, key) for key in paramater_names}
  File "/opt/conda/lib/python3.10/site-packages/conda/base/context.py", line 803, in channels
    return tuple(IndexedSet((*local_add, *self._channels)))
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1236, in __get__
    expanded = merged.expand() if self._expandvars else merged
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 557, in expand
    new_value = type(self.value)(v.expand() for v in self.value)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 557, in <genexpr>
    new_value = type(self.value)(v.expand() for v in self.value)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 555, in expand
    new_value = type(self.value)((k, v.expand()) for k, v in self.value.items())
  File "/opt/conda/lib/python3.10/site-packages/conda/_vendor/frozendict/__init__.py", line 21, in __init__
    self._dict = self.dict_cls(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 555, in <genexpr>
    new_value = type(self.value)((k, v.expand()) for k, v in self.value.items())
AttributeError: 'YamlRawParameter' object has no attribute 'expand'

  File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1238, in print_unexpected_error_report
    message_builder.append(get_main_info_str(error_report['conda_info']))
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_info.py", line 236, in get_main_info_str
    info_dict['_' + key] = ('\n' + 26 * ' ').join(info_dict[key])
KeyError: 'pkgs_dirs'

      File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1118, in __call__
        return func(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
        exit_code = do_call(args, p)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 91, in do_call
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 33, in execute
        execute_config(args, parser)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 162, in execute_config
        d = {key: getattr(context, key) for key in paramater_names}
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 162, in <dictcomp>
        d = {key: getattr(context, key) for key in paramater_names}
      File "/opt/conda/lib/python3.10/site-packages/conda/base/context.py", line 803, in channels
        return tuple(IndexedSet((*local_add, *self._channels)))
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1236, in __get__
        expanded = merged.expand() if self._expandvars else merged
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 557, in expand
        new_value = type(self.value)(v.expand() for v in self.value)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 557, in <genexpr>
        new_value = type(self.value)(v.expand() for v in self.value)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 555, in expand
        new_value = type(self.value)((k, v.expand()) for k, v in self.value.items())
      File "/opt/conda/lib/python3.10/site-packages/conda/_vendor/frozendict/__init__.py", line 21, in __init__
        self._dict = self.dict_cls(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 555, in <genexpr>
        new_value = type(self.value)((k, v.expand()) for k, v in self.value.items())
    AttributeError: 'YamlRawParameter' object has no attribute 'expand'

`$ /opt/conda/bin/conda config --show`

  environment variables:
conda info could not be constructed.
KeyError('pkgs_dirs')

future releases of conda.
Would you like conda to send this report to the core maintainers? [y/N]: N

No report sent. To permanently opt-out, use

    $ conda config --set report_errors false

```
</details>
<details>
<summary>

```sh
# conda --version
conda 22.11.1
# conda config --validate
```
</summary>

```python
WARNING conda.exceptions:print_unexpected_error_report(1240): KeyError('pkgs_dirs')
  File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1118, in __call__
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
    exit_code = do_call(args, p)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 91, in do_call
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 33, in execute
    execute_config(args, parser)
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 220, in execute_config
    context.validate_all()
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1383, in validate_all
    validation_errors = list(chain.from_iterable(self.check_source(source)[1]
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1383, in <genexpr>
    validation_errors = list(chain.from_iterable(self.check_source(source)[1]
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1366, in check_source
    typed_value = loaded_parameter.typify(match.source)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 591, in typify
    return LoadedParameter._typify_data_structure(self.value, source, element_type)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 604, in _typify_data_structure
    return type(value)(v.typify(source) for v in value)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 604, in <genexpr>
    return type(value)(v.typify(source) for v in value)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 591, in typify
    return LoadedParameter._typify_data_structure(self.value, source, element_type)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 602, in _typify_data_structure
    return type(value)((k, v.typify(source)) for k, v in value.items())
  File "/opt/conda/lib/python3.10/site-packages/conda/_vendor/frozendict/__init__.py", line 21, in __init__
    self._dict = self.dict_cls(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 602, in <genexpr>
    return type(value)((k, v.typify(source)) for k, v in value.items())
AttributeError: 'YamlRawParameter' object has no attribute 'typify'

  File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1238, in print_unexpected_error_report
    message_builder.append(get_main_info_str(error_report['conda_info']))
  File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_info.py", line 236, in get_main_info_str
    info_dict['_' + key] = ('\n' + 26 * ' ').join(info_dict[key])
KeyError: 'pkgs_dirs'

      File "/opt/conda/lib/python3.10/site-packages/conda/exceptions.py", line 1118, in __call__
        return func(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
        exit_code = do_call(args, p)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 91, in do_call
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 33, in execute
        execute_config(args, parser)
      File "/opt/conda/lib/python3.10/site-packages/conda/cli/main_config.py", line 220, in execute_config
        context.validate_all()
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1383, in validate_all
        validation_errors = list(chain.from_iterable(self.check_source(source)[1]
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1383, in <genexpr>
        validation_errors = list(chain.from_iterable(self.check_source(source)[1]
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 1366, in check_source
        typed_value = loaded_parameter.typify(match.source)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 591, in typify
        return LoadedParameter._typify_data_structure(self.value, source, element_type)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 604, in _typify_data_structure
        return type(value)(v.typify(source) for v in value)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 604, in <genexpr>
        return type(value)(v.typify(source) for v in value)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 591, in typify
        return LoadedParameter._typify_data_structure(self.value, source, element_type)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 602, in _typify_data_structure
        return type(value)((k, v.typify(source)) for k, v in value.items())
      File "/opt/conda/lib/python3.10/site-packages/conda/_vendor/frozendict/__init__.py", line 21, in __init__
        self._dict = self.dict_cls(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/conda/common/configuration.py", line 602, in <genexpr>
        return type(value)((k, v.typify(source)) for k, v in value.items())
    AttributeError: 'YamlRawParameter' object has no attribute 'typify'

`$ /opt/conda/bin/conda config --validate`

  environment variables:
conda info could not be constructed.
KeyError('pkgs_dirs')

future releases of conda.
Would you like conda to send this report to the core maintainers? [y/N]: N

No report sent. To permanently opt-out, use

    $ conda config --set report_errors false

```
</details>
```sh
# micromamba --version
1.1.0
# micromamba config list
error    libmamba Bad conversion of configurable 'channels' from source '~/.condarc' : yaml-cpp: error at line 4, column 5: bad conversion
show_banner: false
```

----
All in all, I'm quite doubtful that changing the structure/type of the existing configuration key is worth it.
Adding a new alternative key would be less disruptive (since `conda` just ignores unknown keys).
(Plus the potential opportunity to improve/fix (e.g., see gh-7022) `custom_channel`/`custom_multichannel` by incorporation those as mentioned by @wolfv in https://github.com/conda/conda/issues/11825#issuecomment-1296181916 -- granted, that might need some more head scratching...)

mbargull:
> #### No full support for `conda config --set channels`
> 
> This change does not fully support setting channels with extra configuration parameters via the command line. I decided not to do this because it goes against the "MVP" spirit of this pull request. I will be making another issue for it though so that we can support this in the future. I'm not sure how useful this would be though.

That's quite reasonable. Handling nested entries via simple CLI parameters and/or custom environment variable parsing is no fun. If this is needed at some point, we'd be better off with some generic approach like passing on objects via JSON (somewhat like `conda-build --variants`). (Or JSON Patch if manipulation is required -- but manipulation is another little can of worms...)

jezdez:
Thank you for your feedback @mbargull, it comes at the right time. You're right that backward compatibility needs more attention for this change to land, e.g. reviewing how older versions of conda would react to a changed config format.

I would suggest as the next steps:

1. Weigh the pros and cons about adding this feature under a new name in detail, e.g. a `channel_map` config parameter.

  Should it be added as a new config-only option (no CLI integration, to work around the issues of updating complex config options via CLIs)?

  Would config parsers handle the new config option with reasonable error messages?

  How would be the order of importance look like when build the definitive list channels out of the `channels` and `channel_map` config options?

2. Are there other ways to provide additional options to channels that have a more lenient format, e.g. by leaning into URIs?

```
channels:
  - defaults
  - conda-forge
  - http://localhost?authentication=test-auth&test_param=test-value
  - http://some.hoster.tld/?authentication=test-auth&list_param=value_1&list_param=value_2
```

travishathaway:
Here's alternative implementation I came up with today:

- https://github.com/conda/conda/pull/12239

Still working on it, but I think it's basically good to go.

Issue #11718, opened 2022-08-10
Drop pre-Python 3.7 codes


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

Old Python code is less concise than newer code, let's get rid of unnecessary stuff.

### Linked Issues & PRs

- [ ] Replace `odict` with standard `dict` (since its now ordered by default)
- [ ] Replace `toolz.itertoolz.concatv` with arg unpacking
- [ ] Replace `toolz.itertoolz.concat` with `itertools.chain.from_iterable`
- [ ] Replace `toolz.itertoolz.groupby` with `itertools.groupby`

dholth:
We ran pyupgrade https://github.com/conda/conda/pull/11909

jakirkham:
Thanks Daniel! 🙏

Looks like some things (like replacing `odict` with `dict`) still remain

kenodegard:
The code base is now solidly Python 3.7+ with dropping all `toolz`/`collections.OrderedDict` usage in favor of using stdlib functionality.

dholth:
Amazing!

PR #12039, opened 2022-11-01
Remove `toolz.dicttoolz.merge(_with)` usage

kenodegard:
### Description

This initially started as an effort to remove usage of `toolz.dicttoolz.merge` and `toolz.dicttoolz.merge_with` but found that the variables could benefit from renaming for greater clarity.

Xref: https://github.com/conda/conda/issues/11718

### Checklist - did you ...


dholth:
This one might deserve a test comparing the new and old techniques

kenodegard:
@dholth I agree that additional tests are desirable, but I don't see an easy way (without an even more significant refactor) to add more meaningful tests outside of the ones that already exist in `tests/common/test_configuration.py`. These tests are not unit tests, but they indirectly test the `merge` methods (I caught several errors in my implementation with these tests). Do you think this is sufficient or should this work be postponed until additional refactoring & tests can be done?

beeankha (approved):
I vote for additional refactoring & tests being done in a separate PR if they are deemed necessary; in the meantime, this looks good to me! 👍

tests/common/test_iterators.py
```diff
@@ -0,0 +1,46 @@
...
+    old_merge = merge_with(sum, mappings)
+
+    # new style
+    grouped_map = {}
+    for mapping in mappings:
+        for key, value in mapping.items():
+            grouped_map.setdefault(key, []).append(value)
+    new_merge = {key: sum(values) for key, values in grouped_map.items()}
+
+    assert old_merge == new_merge
```
dholth, line 46: It would be neat to use the deprecated decorator or a timestamp check on this kind of test, which we should remove once the new code is proven.

kenodegard: Since we're on track to remove `tqdm` entirely this should be easy to find when we get around to it?

PR #12042, opened 2022-11-01
Use GNU tar to successfully restore Windows github actions cache

dholth:
### Description

Windows cache isn't restoring correctly. Does the other slash work better?


dholth:
It works. The secret sauce is to use the tar that comes with git. Doesn't care about forward versus backslash.

```
Cache restored successfully
Cache restored from key: Windows-conda-2-py3.10-ecd82714a54c2ae893966e370621d9b86e49165de195e3103066d41d6925998e
```

dholth:
Turns out the Windows problem is a link between directories, that is created after the cache step. We can cache against C:\Miniconda instead.

Issue #11885, opened 2022-09-27
conda 22.9.0 breaks bash command prompt


SimonPeatman:
### What happened?

I have upgraded from 4.14.0 to 22.9.0 and now my bash command prompt is broken.  I used to see:
```(conda-env) hostname:/path/to/working/dir$ ```

But now I just get:
```(conda-env) ```

In other words, rather than prepending the name of the conda environment in parentheses to the value of the PS1 environment variable, conda is overwriting PS1 altogether.  Doing ```conda deactivate``` leaves PS1 as an empty string.

### The Fix

Run the following and restart your shell for the changes to take effect:

```
conda init bash
```

See: https://github.com/conda/conda/issues/11885#issuecomment-1259519190

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : himawari
    active env location : /home/users/peatman/miniconda3/envs/himawari
            shell level : 2
       user config file : /home/users/peatman/.condarc
 populated config files : /home/users/peatman/.condarc
          conda version : 22.9.0
    conda-build version : not installed
         python version : 3.9.7.final.0
       virtual packages : __linux=3.10.0=0
                          __glibc=2.17=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /home/users/peatman/miniconda3  (writable)
      conda av data dir : /home/users/peatman/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/users/peatman/miniconda3/pkgs
                          /home/users/peatman/.conda/pkgs
       envs directories : /home/users/peatman/miniconda3/envs
                          /home/users/peatman/.conda/envs
               platform : linux-64
             user-agent : conda/22.9.0 requests/2.28.1 CPython/3.9.7 Linux/3.10.0-1160.66.1.el7.x86_64 centos/7.9.2009 glibc/2.17
                UID:GID : 7050034:26030
             netrc file : /home/users/peatman/.netrc
           offline mode : False
```

</details>
<details>
<summary><code>conda config</code></summary>

```shell
==> /home/users/peatman/.condarc <==
auto_activate_base: False
```

</details>
<details>
<summary><code>conda list</code></summary>

```shell
[...]

PR #12043, opened 2022-11-01
Make sure conda does not use polyglot shebangs

jaimergp:
### Description

Minimum work to fix https://github.com/conda/conda/issues/11885 and close https://github.com/conda/conda/issues/11893.

More robust solution long-term available at https://github.com/conda/conda/pull/11995.

Basically we prevent `conda` from using the polyglot shebang approach with two checks:

* If conda-build is being used and the executable is inside `$PREFIX`, we are most probably packaging `conda`. Do nothing and let prefix replacement handle that at install time (`replace_long_shebang()` right above our function)
* If somebody uses `conda init --install` outside conda-build (there's no reason to, but :shrug:), we ensure they end up with a `#!/usr/bin/env python` shebang by calling the function with a new flag.

### Checklist - did you ...

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -775,7 +775,7 @@ def make_entry_point(target_path, conda_prefix, module, func):
         new_ep_content = ""
     else:
         python_path = join(conda_prefix, get_python_short_path())
-        new_ep_content = generate_shebang_for_entry_point(python_path)
+        new_ep_content = generate_shebang_for_entry_point(python_path, with_usr_bin_env=True)
 
     conda_extra = dals("""
     # Before any more imports, leave cwd out of sys.path for internal 'conda shell.*' commands.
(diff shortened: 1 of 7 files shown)


jaimergp:
Ready for review but I want to check the canaries that come out of this PR once #12047 is addressed.

jezdez (approved):

beeankha (approved):

jezdez (approved):

jaimergp:
I asked about the canaries because I wanted to review how the entry points were being rendered 😬

kenodegard:
conda-build 3.23.0 is on defaults now, we'll get the canary builds on the main branch shortly

jaimergp:
Awesome, the test changes in the recipe should reveal the shebangs properly. So we just need to check the logs.

PR #12049, opened 2022-11-02
ensure test dependencies even if we are ahead of Docker image; quiet …

dholth:
### Description

Avoid progress bars in test logs while setting up environment; make sure test dependencies are installed even if Docker image is behind.

We haven't added tests dependencies in a while, but if we do, the current situation requires us to merge the tests dependencies to main beforehand. Instead, always try to install them – if they are already installed, this should be quick.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,13 +42,24 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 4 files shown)


dholth:
Included in other PR.

PR #12050, opened 2022-11-02
Defer list of PackageRecord objects

### LINKED ISSUE #8500: Using a fast JSON parser for repo data
This came up recently in our meeting at AnacondaCON. Writing it down so we can track it.

@mariusvniekerk, suggested we should look at using another faster JSON parser for reading the repo data. Here is [a blogpost]( https://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/ ) with some Python JSON parser benchmarks. All of these are in Conda-forge. It looks like we should choose between `ujson` and `python-rapidjson`. Considering loading is the priority for users loading, `ujson` looks better. Though `python-rapidjson` is a bit faster for writing.

dholth:
### Description

Conda spends time building PackageRecord for every package, even if it only needs a few of them. This is slower when loading a fresh repodata.json.

Lazily create them as needed. Fix #8500 ?

This can't use the older pickle data; I've commented that part out for benchmarking anyway.

Before this can be merged we would need to benchmark operations that touch many packages, to make sure we aren't much slower than when eagerly creating them.

PackageRecord uses an auxlib Entity class that is slow. Maybe it is validating all records as soon as they are created?

Searching for a single named package without lazy PackageRecord:

```
conda: 2187.81 ms
libmamba: 1701.37 ms
```

With lazy PackageRecord:

```
conda: 783.34 ms
libmamba: 1714.94 ms
```

It is possible that the conda+pickled version is faster, if you can re-use the pickle cache, but this test deliberately clears the cache each time.

### Checklist - did you ...


conda/core/subdir_data.py
```diff
@@ -322,7 +344,7 @@ def _pickle_me(self):
 
     def _read_local_repdata(self, etag, mod_stamp):
         # first try reading pickled data
-        _pickled_state = self._read_pickled(etag, mod_stamp)
+        _pickled_state = False  # self._read_pickled(etag, mod_stamp)
```
dholth, line 347: We can change `REPODATA_PICKLE_VERSION` in this module to fix the loads-wrong-pickle problem. However we don't like having to spend the time to save the pickle in the first place. Whether the pickle is worth it could depend on how often you run conda versus how often the repository gets updated.

dholth:
Need to do a few simple benchmarks of loading json, loading/saving pickle, and iterating over all lazy package records compared to the eager main branch.

dholth:
Here are some tests from https://github.com/dholth/conda-benchmarks/blob/main/benchmarks/subdir_data.py against main/osx-64 repodata.json

The pickle speed looks wrong; it should be faster, because it is disabled, on the lazy branch.

```
2):
[ 50.00%] ·· Benchmarking conda-py3.10-conda-package-handling2.0.0a4
[ 54.17%] ··· subdir_data.TimeSubdirData.time_load_json                  169±5ms
[ 58.33%] ··· ...data.TimeSubdirData.time_load_json_query_all            172±4ms
[ 62.50%] ··· ...data.TimeSubdirData.time_load_json_query_one            169±8ms
[ 66.67%] ··· subdir_data.TimeSubdirData.time_load_pickle            1.32±0.06ms
[ 70.83%] ··· subdir_data.TimeSubdirData.time_save_pickle            1.18±0.02ms
[ 75.00%] ··· subdir_data.TimeSubdirData.time_subdir_data               302±20μs
[ 75.00%] · For conda commit 3f8db559 <main> (round 2/2):
[ 75.00%] ·· Building for conda-py3.10-conda-package-handling2.0.0a4.
[ 75.00%] ·· Benchmarking conda-py3.10-conda-package-handling2.0.0a4
[ 79.17%] ··· subdir_data.TimeSubdirData.time_load_json                  937±7ms
[ 83.33%] ··· ...data.TimeSubdirData.time_load_json_query_all            899±6ms
[ 87.50%] ··· ...data.TimeSubdirData.time_load_json_query_one           894±10ms
[ 91.67%] ··· subdir_data.TimeSubdirData.time_load_pickle            1.28±0.02ms
[ 95.83%] ··· subdir_data.TimeSubdirData.time_save_pickle            1.37±0.02ms
[100.00%] ··· subdir_data.TimeSubdirData.time_subdir_data               276±20μs
       before           after         ratio
     [3f8db559]       [86153959]
     <main>           <lazy-instantiate-packagerecord>
-     1.37±0.02ms      1.18±0.02ms     0.86  subdir_data.TimeSubdirData.time_save_pickle
-         899±6ms          172±4ms     0.19  subdir_data.TimeSubdirData.time_load_json_query_all
-        894±10ms          169±8ms     0.19  subdir_data.TimeSubdirData.time_load_json_query_one
-         937±7ms          169±5ms     0.18  subdir_data.TimeSubdirData.time_load_json

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE INCREASED.
```

conda/core/subdir_data.py
dholth: Also, if we pickle the un-parsed repodata we haven't saved the user their repodata -> PackageRecord time on the next run.

If we wanted to get complex, we could do the deferred parsing during another download.

conda/core/subdir_data.py
```diff
@@ -98,6 +98,22 @@ def __call__(cls, channel, repodata_fn=REPODATA_FN):
...
+
+    def __getitem__(self, i):
+        if isinstance(i, slice):
+            return self.__class__(self.data[i])
+        else:
+            record = self.data[i]
+            if not isinstance(record, PackageRecord):
+                record = PackageRecord(**record)
+                self.data[i] = record
+            return record
```
dholth, line 114: conda-build wants to iterate over `_package_records`  so this list keeps things compatible, and could simplify the rest of this PR (but the iter_records() with looser guarantees is nice)

dholth:
I think we could stop doing pickle entirely with this PR.

It needs better benchmarks. If you iterate over all the PackageRecord often without downloading a new repodata.json then you might want the pickle. Otherwise, it is a waste of time. If you're using the new solver, then you don't need PackageRecord since all the action is inside libmamba-solver.

jezdez:
@dholth Could you rebase this please?

dholth:
@jezdez updated

conda/core/subdir_data.py
```diff
@@ -549,15 +571,15 @@ def _process_raw_repodata(self, repodata, state=None):
...
-                package_kwargs.update({"fn": fn, "url": join_url(channel_url, fn)})
-                package_record = PackageRecord(**package_kwargs)
-
-                _package_records.append(package_record)
-                _names_index[package_record.name].append(package_record)
-                for ftr_name in package_record.track_features:
-                    _track_features_index[ftr_name].append(package_record)
+                # lazy
+                # package_record = PackageRecord(**info)
+                info["fn"] = fn
```
dholth, line 576: should consider copying this

dholth:
Finally got a stable benchmark I'm happy with.

```
       before           after         ratio
     [c9aa8ff6]       [044cf5df]
     <main>           <lazy-instantiate-packagerecord>
-     1.39±0.04ms      1.23±0.04ms     0.89  subdir_data.TimeSubdirData.time_save_pickle
-      1.93±0.03s       1.17±0.01s     0.61  subdir_data.TimeSubdirData.time_load_json_query_all
-      1.85±0.04s          362±2ms     0.20  subdir_data.TimeSubdirData.time_load_json_query_one
-         923±8ms          173±3ms     0.19  subdir_data.TimeSubdirData.time_load_json
-      1.01±0.02s          187±4ms     0.18  subdir_data.TimeSubdirData.time_subdir_data

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE INCREASED.
```

Mostly as expected: query_all (instantiate all PackageRecord) is similar speed - but where did the other 1.93 - 1.17 seconds go? but query_one (instantiates all PackageRecord with a given name) is faster. On the other hand we want to spend time_subdir_data seconds plus network access, and hand the entire json to libmamba.

jezdez:
@dholth To your point above:

> Before this can be merged we would need to benchmark operations that touch many packages, to make sure we aren't much slower than when eagerly creating them.

What do we need to do to implement speed benchmarking for this? We have a dedicated [GHA workflow for "perfomance" testing](https://github.com/conda/conda-libmamba-solver/blob/main/.github/workflows/performance.yml) in CLS: which effectively just runs [a set of slow tests](https://github.com/conda/conda-libmamba-solver/blob/568fbd84ec0976b2492c51794867fa7441792390/tests/test_performance.py) we identified while working on the solver in a timebox. Could you reuse these here?

conda/core/subdir_data.py
jezdez: Hm, no idea if we have any constraints on copying, but if there were, that promise is very likely broken somewhere and then probably not worth keeping.

conda/core/subdir_data.py
jezdez: > Whether the pickle is worth it could depend on how often you run conda versus how often the repository gets updated.

My best guess is that conda commands are often run in short succession until an environment is set up correctly, and then not updated until a reason comes along.

> If we wanted to get complex, we could do the deferred parsing during another download.

If possible, we should simplify, unless it loses us correctness.

conda/core/subdir_data.py
```diff
@@ -80,6 +79,22 @@ def __call__(cls, channel, repodata_fn=REPODATA_FN):
         return subdir_data_instance
 
 
+class LazyPackageRecords(UserList):
```
jezdez, line 82: In the spirit of naming things is hard, let's focus on the positive and just name this:

```suggestion
class PackageRecordList(UserList):
```

conda/core/subdir_data.py
dholth: Will keep it as is. I'm more confident that we don't need to copy.

dholth:
@jezdez I think my comment about benchmarking is outdated. This benchmark https://github.com/conda/conda-benchmarks/blob/main/benchmarks/subdir_data.py is pretty good.

The one benchmark that is missing would be to compare "a bunch of operations that use the eager-packagerecords pickle" versus a bunch of operations that use the lazy one.

We could skip doing that benchmark. The new code is fast; it should be very rare to need to instantiate over all PackageRecords (the one deferred operation) many times while the cache is still valid; if you have to instantiate them all, it's not slower than the old code.

We don't need to continuously benchmark this. It would be nice but we just need to make sure the technique is worthwhile ahead of time.

dholth:
Giving it a try with unchanged pickle code (does create the pickle).

dholth:
It would be more interesting to test the write / don't write pickle on a slower disk (the developer laptop disk is very fast)

conda/core/subdir_data.py
```diff
@@ -52,7 +51,7 @@
 
 log = getLogger(__name__)
 
-REPODATA_PICKLE_VERSION = 29
+REPODATA_PICKLE_VERSION = 31
```
dholth, line 54: 30 is currently being used by repodata-jlap (and maybe we can stop pickling in that branch)

jezdez: I'd favor the removal of the pickle code if we can assume it won't have a negative side effect for performance and correctness

dholth: I don't have time to verify whether that is the case. Instead, I've simplified this PR.

dholth: https://github.com/conda/conda/issues/12277

jezdez: Good decision, thank you!

jezdez (approved):
🚀

jezdez:
@dholth Oh, could you add a news item?

jezdez (approved):

PR #12058, opened 2022-11-04
Move `conda.models.plugins` to `conda.plugins.types`

kenodegard:
### Description

POC of internal discussion.

Moving `conda.models.plugins` to `conda.plugins.types` for stricter domain organization and to avoid adding more confusion to `conda.models`.

My intention with the import rules of `conda/plugins/__init__.py` is to expose everything necessary to make a plugin. The plugin implementations (e.g., `conda.plugins.virtual_pacakges`) should never be sideloaded by importing `conda.plugins`.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/__init__.py
+++ b/conda/plugins/__init__.py
@@ -1,3 +1,4 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 from .hookspec import hookimpl  # noqa: F401
+from .types import CondaVirtualPackage, CondaSubcommand  # noqa: F401
--- a/conda/plugins/hookspec.py
+++ b/conda/plugins/hookspec.py
@@ -7,7 +7,7 @@ from collections.abc import Iterable
 
 import pluggy
 
-from ..models.plugins import CondaSubcommand, CondaVirtualPackage
+from .types import CondaSubcommand, CondaVirtualPackage
 
 spec_name = "conda"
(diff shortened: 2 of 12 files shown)


conda/plugins/__init__.py
```diff
@@ -1,3 +1,4 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 from .hookspec import hookimpl  # noqa: F401
+from .types import CondaVirtualPackage, CondaSubcommand  # noqa: F401
```
jezdez, line 4: FWIW this is the thing that makes me worried, that I mentioned in our conversations: `conda.plugins.types` will eventually have non-trivial functionality and import stuff from elsewhere in conda or even 3rd party library when we add new plugin hooks. Importing the types in the `conda.plugins` may then trigger import-time side effects that are not in line with the need to have a fast-loading, dependency-free plugin system.

@dholt You have a lot of experience with Python's import system, what's your take on this?

dholth: Checking conda's init https://github.com/conda/conda/blob/main/conda/__init__.py loaded as a side effect of importing conda dot anything...

I'm not sure how complex types will get but agree it seems like a good idea to avoid conda.models

jezdez: Thanks for the input, let's go ahead then.

jezdez (approved):

jezdez:
pre-commit.ci autofix

jezdez (approved):

PR #12061, opened 2022-11-04
Update Docstring in Hookspec Module

beeankha:
When `self` is declared in a plugin function decorated with `@plugins.hookimpl`, the following error is output:

```
    hookimpl definition: conda_subcommands(self)
    Argument(s) {'self'} are declared in the hookimpl but can not be found in the hookspec
```

Editing the docstrings in `hookspec` module in order to avoid a plugin developer from encountering this error.


dholth (approved):

jezdez:
I'm wondering if this was a weird rebase issue, since in fact the self parameters were missing in the `CondaSpecs` methods!

PR #12062, opened 2022-11-04
Prevent [name=...] from overwriting name specified outside of brackets

pkmooreanaconda:
### Description

Letting [name=...] overwrite the name specified outside of brackets in MatchSpecs allows for confusing cases where it seems one package will be installed but, in reality, another one is installed.

It's worth noting that this is a minor change to MatchSpec behavior.  We have no indication that this behavior is relied upon anywhere but its worth keeping in mind.  We should be confident that we aren't negatively impacting anyone before merging this PR.

### Checklist - did you ...

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -8,6 +8,7 @@ from functools import reduce
 from logging import getLogger
 from operator import attrgetter
 from os.path import basename
+import warnings
 import re
 
 try:
@@ -742,6 +743,17 @@ def _parse_spec_str(spec_str):
         components['build'] = build
 
     # anything in brackets will now strictly override key as set in other area of spec str
+    # EXCEPT FOR: name
+    # If we let name in brackets override a name outside of brackets it is possible to write
+    # MatchSpecs that appear to install one package but actually install a completely different one
+    # e.g. tensorflow[name=* version=* md5=<hash of pytorch package> ] will APPEAR to install
+    # tensorflow but actually install pytorch.
+    if "name" in components and "name" in brackets:
+        warnings.warn(
+            f"'name' specified both inside ({brackets['name']}) and outside ({components['name']})"
+            " of brackets. the value outside of brackets ({components['name']}) will be used."
+        )
+        del brackets["name"]
     components.update(brackets)
     components['_original_spec_str'] = original_spec_str
     _PARSE_CACHE[original_spec_str] = components
--- a/news/12062-fix-name-overwrite
+++ b/news/12062-fix-name-overwrite
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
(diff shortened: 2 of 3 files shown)


conda/models/match_spec.py
```diff
@@ -742,6 +743,17 @@ def _parse_spec_str(spec_str):
...
 
     # anything in brackets will now strictly override key as set in other area of spec str
+    # EXCEPT FOR: name
+    # If we let name in brackets override a name outside of brackets it is possible to write
+    # MatchSpecs that appear to install one package but actually install a completely different one
+    # e.g. tensorflow[name=* version=* md5=<hash of pytorch package> ] will APPEAR to install
+    # tensorflow but actually install pytorch.
+    if "name" in components and "name" in brackets:
+        warnings.warn(
+            f"'name' specified both inside ({brackets['name']}) and outside ({components['name']})"
```
dholth, line 753: specifies?

news/12062-fix-name-overwrite
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixed a confusing case where [name=...] could overwrite the name specified outside of brackets. (#12062)
```
dholth, line 7: How about something like "Prefer the outer name when a MatchSpec specifies a package's name twice `package-name[name=package-name]`"

Unfortunately we can write MatchSpec's without an outer package name?

tests/models/test_match_spec.py
```diff
@@ -655,6 +655,30 @@ def test_parse_spec_str_with_brackets(self):
             "build": "3",
         }
 
+        # Ensure 'name' within brackets can't override the name specified outside of brackets
+        assert _parse_spec_str(
+            "tensorflow[name=* version=* md5=253b922ecdb5a30884875948b8904983]"
```
dholth, line 660: this is a mistake, md5 should not be used for anything :)

news/12062-fix-name-overwrite
pkmooreanaconda: Yeah,  "*[name=...]" is an accepted MatchSpec.  Something of the flavor "*[sha256=...]" is the intended way to install the package with exactly that hash.

tests/models/test_match_spec.py
pkmooreanaconda: https://github.com/search?q=%22md5%22+filename%3Aenvironment.yml&type=Code

And yet it is out there!  I'm happy to add another test for sha256 or whatever the appropriate hash is

conda/models/match_spec.py
pkmooreanaconda: "MatchSpec specifies 'name' both inside (...) and outside (...) of brackets..." ?

dholth (commented):
news update

dholth (approved):

jezdez (approved):

PR #12067, opened 2022-11-07
use python 3.9 for test_conda_pip_interop_conda_editable_package

dholth:
### Description

Was seeing `ValueError: Unregistered SAT variable name: conda-forge/noarch::urllib3-1.26.11-pyhd8ed1ab_0` in this test. Update Python to make it easier to test on modern systems.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -1603,8 +1603,9 @@ dependencies:
 
 
     def test_conda_pip_interop_conda_editable_package(self):
-        with env_var('CONDA_RESTORE_FREE_CHANNEL', True, stack_callback=conda_tests_ctxt_mgmt_def_pol):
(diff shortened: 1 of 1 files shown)


jezdez (approved):

travishathaway:
@dholth, I took a look at this today and was not able to get this test running successfully myself.

I think we should go back to the drawing board with this particular test. There's a lot of stuff going on here I'm not happy about. One of them is the reliance on the `urllib3` library and the way we retrieve it from GitHub. I think it would be a better idea to simply rely on a test package that we create ourselves to test that pip interoperability continues to work in the way we expect it.

I would say that if this particular test is proving problematic, we could simply skip it for now and create an issue to come be later and refactor it.

dholth:
I was able to get the same test working in another branch. Not sure what makes it so fragile. We should try to decouple our tests from changing network repositories as much as possible.

Issue #11079, opened 2021-12-03
Fish shell 2.x support broken by removed evals


w-m:
### Checklist
* [x] I added a descriptive title
* [x] I searched for other issues and couldn't find a solution or duplication
* [x] I already searched in Google and didn't find any good information or help
* [x] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?
I'm running Ubuntu 18.04 LTS. Using fish shell, version 2.7.1. This is an outdated version of fish, but it's the one that ships with that Ubuntu version.

After updating conda, I got the following error when creating a new shell: 

```
~/anaconda3/etc/fish/conf.d/conda.fish (line 2): Variables may not be used as commands. In fish, please define a function or use 'eval $CONDA_EXE'.
        $CONDA_EXE
        ^
in function “conda”
        called on standard input
```

The environment is completely broken, it's not possible to use `conda` in this state. The cause seems to be #10696, which removed the evals. This breaks fish shell v2.7.1 support.

A fix for me was to update fish shell by installing release 3 through the ppa (https://launchpad.net/~fish-shell/+archive/ubuntu/release-3).

### Conda info
```shell
active environment : base
    active env location : /home/morgenstern/anaconda3
            shell level : 1
       user config file : /home/morgenstern/.condarc
 populated config files : /home/morgenstern/.condarc
          conda version : 4.11.0
    conda-build version : 3.21.4
         python version : 3.8.10.final.0
       virtual packages : __cuda=11.3=0
                          __linux=5.4.0=0
                          __glibc=2.27=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /home/morgenstern/anaconda3  (writable)
      conda av data dir : /home/morgenstern/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/morgenstern/anaconda3/pkgs
                          /home/morgenstern/.conda/pkgs
       envs directories : /home/morgenstern/anaconda3/envs
                          /home/morgenstern/.conda/envs
               platform : linux-64
[...]

PR #10696, opened 2021-05-27
Remove extra evals in fish shell

johnflavin:
Solve issues with shell interpretation and escaping characters in `fish` shell by removing superfluous `eval`s.

This fix is the same as discussed in #9145.

Fixes #9145
Fixes #7611
Fixes #10643

PR #12070, opened 2022-11-08
Fix fish activation scripts

jonashaag:
### Description

From https://github.com/mamba-org/mamba/pull/2101:

In fish `(subproc)` will replace newlines with spaces. Thus, `eval (subproc)` will evaluate the output of `subproc` with newlines replaced with spaces. We need `| source` to keep the newlines.

### Checklist - did you ...

### DIFF
--- a/conda/shell/etc/fish/conf.d/conda.fish
+++ b/conda/shell/etc/fish/conf.d/conda.fish
@@ -69,10 +69,10 @@ function conda --inherit-variable CONDA_EXE
         set -e argv[1]
         switch $cmd
             case activate deactivate
-                eval ($CONDA_EXE shell.fish $cmd $argv)
+                $CONDA_EXE shell.fish $cmd $argv | source || return $status
             case install update upgrade remove uninstall
-                $CONDA_EXE $cmd $argv
-                and eval ($CONDA_EXE shell.fish reactivate)
+                $CONDA_EXE $cmd $argv || return $status
+                $CONDA_EXE shell.fish reactivate | source || return $status
             case '*'
                 $CONDA_EXE $cmd $argv
         end


jezdez:
@jonashaag Thanks for the patch, really appreciate the upstreaming from mamba!

@johnflavin Since you last worked on this part of the fish support in #10696 and have responded in #11079 about whether we should continue to support fish < 3.x (we don't, even though #11079 hasn't been solved yet).

I wonder if you have any insights into whether this PR here would actually be preferable to using `eval`?

johnflavin:
> I wonder if you have any insights into whether this PR here would actually be preferable to using `eval`?

I am honored that I have become the local expert on `fish`, though I can't say I am the best person for the role. 😬 

From the fish [`eval` docs](https://fishshell.com/docs/current/cmds/eval.html):

> If the command does not need access to stdin, consider using [source](https://fishshell.com/docs/current/cmds/source.html#cmd-source) instead.

Makes me think that this change is a 👍🏼 .

johnflavin:
I took a look at https://github.com/mamba-org/mamba/pull/2101, and I would like this change from there to come here as well:
```diff
-  set -gx PATH $MAMBA_ROOT_PREFIX/condabin $PATH
+  fish_add_path --move $MAMBA_ROOT_PREFIX/condabin
```
Is it possible to get this `fish_add_path` change? (With the appropriate non-mamba environment variable, of course.)

jezdez:
@johnflavin Given that `fish_add_path` was added in fish 3.2.0, I'm a little concerned this'll be the same as with #11079. It was released in March 2021, so not that long ago and probably hasn't hit the various Linux distributions. Do you have details why this would be useful to use?

PR #12071, opened 2022-11-08
Create Python 3.11 conda-ci image

dbast:
### Checklist - did you ...

Context to the pycrypto replacement:
* https://github.com/conda-forge/pycrypto-feedstock/pull/27
* https://github.com/conda-forge/conda-smithy/pull/1128

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -46,7 +46,7 @@ jobs:
       packages: write
     strategy:
       matrix:
-        python: ['3.7', '3.8', '3.9', '3.10']
+        python: ['3.7', '3.8', '3.9', '3.10', '3.11']
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


dbast:
3 lines of change which bring a conda-ci python 3.11 image into existence :) Not sure though if the pycrypto to pycryptodome switch is fine.

dbast:
This is green and should not change the existing images... only adds a new one for Python 3.11 ... after a merge another PR could extend the matrix to actually use the new image.

cjmartian (approved):
LGTM

jezdez:
This isn't the right strategy since it would effectively create a pretty drastically different image just for 3.11 and by that add inconsistency in the test system, which is already pretty brittle.

Let's first open a ticket to track the work before we jump into partial solutions and discuss the pros and cons of switching to conda-forge for development.

dbast:
@jezdez While I understand your decision to plan this first before starting something, the inconsistencies you refer to are exactly the same inconsistencies users are experiencing every day by mixing a miniconda3 base installs with defaults and conda-forge packages ... so having one such frankenstein image for testing would just represent the reality ... and it is just an one-liner which produces that image, which can be easily disabled in case testing like this is too horrible.

Issue #11633, opened 2022-07-18
Collapse `conda_env` into `conda`


kenodegard:

jaimergp:
> conda env remove | 🤔 combine with conda remove?

For completeness, this issue seems relevant: https://github.com/conda/conda/issues/11092

dharhas:
> continue to make envs a first-class citizen of conda, as we've learned again and again, a large bulk of install/dependency issues arise from users installing everything into their base env, collapsing conda env commands into conda will help simplify the CLI and should help with educating users

So a bit confused about this logic. Not sure how removing the `env` namespace actually helps improve best practices.

To me the issue was different behavior between similar commands. i.e. `conda create` and `conda env create`

One advantage about the `env` namespace is that it makes it clear whether you are running related in a env or not. I always found the `-n`  flag problematic and opaque to new users.

To me what `conda env list` or `conda env remove` is a lot clearer than `conda info --envs` or `conda remove --name`

To be really heretical (or logical) an alternate proposal could be that `conda create` gets replaces by `conda env create` because then you have a nice dichotomy where plain `conda` commands apply to the current environment and ones in the `env` name space apply to a particular environment. The other option is to use `--env` instead of `--name` to refer to an environment.

Disclaimer, to some extent I'm bike shedding. I care more about having a single consistent way to do things that these particulars.

dharhas:
>  possibly even phasing out the base environments outright and having a strong core functionality that understands environments

I would strongly support ☝🏽

kenodegard:
Another example of `conda` vs `conda env` confusion: https://github.com/conda/conda/issues/11651

itcarroll:
A feature of `conda env create` that I would like to see brought into `conda create` during this "collapse": including the environment name in a file that is read by default.  With `conda env create` you do not have to specify `--name` on the command line as you do with `conda create`. That's really useful for CI workflows that clone a repo with an environment.yml.

If it worked out that "collapsing" the way `conda env create` handles the default YAML file affected other commands, I would not mind 😃 . Just `conda create && conda activate`, with no `--name` after cloning would be grand.

jaimergp:
One small comment for the table that @kenodegard created in the opening message:

`conda env export` is (almost) equivalent to `conda list --export`. This could be easily replicated there, specially when / if we have https://github.com/conda/conda/issues/14070.

jaimergp:
[...]

PR #12074, opened 2022-11-08
add environment variable expansion for repository url

### LINKED ISSUE #10018: Token substitution in YAML env file
I have the following env file:

```yaml
channels:
  - t/$ANACONDA_TOKEN/MYCHANNEL
  - conda-forge
  - anaconda

dependencies:
  - python ==3.7.6
  - myprivatepacakge
```

`ANACONDA_TOKEN` is correctly defined in my bash. But when I run the following, the token is not substituted:

```bash
conda create -y -n test_env
conda activate test_env
conda env update -f test_env.yml
```

I got the following error:

```bash
CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://conda.anaconda.org$anaconda_token/MYCHANNEL/linux-64/repodata.json>
Elapsed: -
```

Note that if I substitute `$ANACONDA_TOKEN` by the real token, it works well.

Is that a bug or just a not supported feature?

Related PR: https://github.com/conda/conda/pull/7826

sanzoghenzo:
### Description

closes #10018

Adds `os.path.expandvars` to expand the input url at the beginning of `split_anaconda_token`.

### Checklist - did you ...


sanzoghenzo:
I already signed the CLA, what should I do to fix the failing check?
Thanks

beeankha:
@conda-bot check

travishathaway:
Hi @sanzoghenzo,

Thanks for this pull request and sorry for the delayed response. At first glance this looks like a very reasonable change. I'm going to review this in depth next week.

jezdez:
@awwad @pkmooreanaconda Do you have any concerns around that?

awwad:
I think this is okay.  It doesn't seem like you can do much damage with a strange environment variable in this context.  It should be noted that `expandvars` will leave the variable name in if it is not defined.

awwad (approved):

jezdez (changes requested):
To stay consistent with other configuration parameters that are being expanded, this needs to be updated to happen at the configuration validation level. Doing this elsewhere on the connection level breaks the separation of concerns with regard to handling configuration values.

The code in question where this SHOULD work is https://github.com/conda/conda/blob/main/conda/base/context.py#L269-L273

news/10018-add-environment-variable-expansion-for-repository-url
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* add environment variable expansion for repository url (#10018)
```
jezdez, line 3: It might be worth elaborating in more detail how this looks like, as it IS a change of behavior and might be useful for end-users to read in the changelog about

pkmooreanaconda:
> To stay consistent with other configuration parameters that are being expanded, this needs to be updated to happen at the configuration validation level. Doing this elsewhere on the connection level breaks the separation of concerns with regard to handling configuration values.
> 
> The code in question where this SHOULD work is
> 
> https://github.com/conda/conda/blob/d63470c4148868c3643b66533dd038a5913ac989/conda/base/context.py#L275-L279

Some thoughts on where this behavior change should take place:

* If the purpose of this change is to allow insertion of a conda token (and **only** a conda token) from an environment variable, it may be better to specifically use the contents of CONDA_TOKEN and perform the replacement in a more limited fashion.  The current change effectively expands any environment variables throughout the url.

* If the purpose of this change is to allow expansion of any environment variable at any point in a url, it is probably better to do this expansion earlier in the parsing process (I believe this is what @jezdez's change implements) so the remainder of the parsing code has a chance to operate on a complete url.  As it stands, parsing functions that examine the url before it is run through split_anaconda_token() will receive a URL with variables while post-token-split functions will receive an expanded one.

sanzoghenzo:
Thanks for the reviews.
@pkmooreanaconda you're right that the needed variable expansion I had in mind is only the conda token,
but I envisioned a case in which a user can have multiple private repositories (for example an anaconda one and a self-hosted one using quetz), hence multiple conda tokens to substitute. With this in mind, limiting the substitution to CONDA_TOKEN doesn't fully solve the issue.

I will try to learn about the validation process @jezdez pointed to and get back with a more correct solution.

sanzoghenzo:
> To stay consistent with other configuration parameters that are being expanded, this needs to be updated to happen at the configuration validation level. Doing this elsewhere on the connection level breaks the separation of concerns with regard to handling configuration values.
> 
> The code in question where this SHOULD work is
> 
> https://github.com/conda/conda/blob/d63470c4148868c3643b66533dd038a5913ac989/conda/base/context.py#L275-L279

Hi @jezdez 
I tried to take a look at this, but it seems to me that the code you pointed to is only validating the arguments passed via CLI (and .condarc, if I understand correctly), but it doesn't validate the contents of the environment YAML file passed with the `--file` argument of `conda env create`, hence the issue #10018 .

In this use case, the channel is taken as is until `conda_env.installers.conda:_solve` calls `conda.models.channel:prioritize_channels`, this in turn creates the `Channel` objects using the `Channel.from_url` method, that parses the url and extracts the token.

I honestly cannot find a better place to do the environment variable expansion.

travishathaway:
Hi @jezdez and @sanzoghenzo,

I dug into this a little to figure out the best way forward. @jezdez, you were right in trying to nudge @sanzoghenzo in the direction of code that is responsible for parsing configuration, but unfortunately there exists multiple classes in conda for doing this. In conda, there is both `conda` and `conda_env` and both of these modules have their own classes and functions for parsing configuration, specifically YAML files. This is because `conda_env` used to be its own project and was brought in to the main `conda` code sometime ago.

@sanzoghenzo, I believe where you want to look instead is the following class:

https://github.com/conda/conda/blob/main/conda_env/env.py#L208

In order to add support for expanding environment variables, we could perhaps add a "protected" method on this class that would iterate through the channel names and then replace the token there. This way, the behavior can remain with this particular class and is coupled to that particular way of parsing the environment config file. Even better, we avoid having to modify anything under `conda.common.url`.

@jezdez and @sanzoghenzo, please let me know your thoughts.

sanzoghenzo:
> @sanzoghenzo, I believe where you want to look instead is the following class:
> 
> https://github.com/conda/conda/blob/main/conda_env/env.py#L208
> 
> In order to add support for expanding environment variables, we could perhaps add a "protected" method on this class that would iterate through the channel names and then replace the token there. This way, the behavior can remain with this particular class and is coupled to that particular way of parsing the environment config file. Even better, we avoid having to modify anything under `conda.common.url`.

Hi @travishathaway , thanks for taking the time to review this!

I think it's best to leave the `Environment` class as is, since it can be created in various ways based on the detected `Spec` class, and to avoid adding logic/coupling to what seems to be a data class.

I would instead implement it in the [`from_yaml` function](https://github.com/conda/conda/blob/fc3371d7fe6b50df9a260a8d50ab13dc1d12ff63/conda_env/env.py#L146), or better yet as a function called by it.

The code changed in my commits will be useless for the proposed solution, should I close this PR and start from scratch?

jezdez:
> I dug into this a little to figure out the best way forward. @jezdez, you were right in trying to nudge @sanzoghenzo in the direction of code that is responsible for parsing configuration, but unfortunately there exists multiple classes in conda for doing this.

@travishathaway The problem remains, there should be only one place where token substitution happens, close to the configuration system and it's not enough to do this in conda-env and conda separately.

We have an item in https://github.com/conda/conda/issues/11633 to "Combine conda env update with conda update and make the former an alias" which would be appropriate to try to do as prerequisite of #10018.

@sanzoghenzo You're right, I missed that this was related to the env subcommand as @travishathaway correctly pointed out. I would like to suggest for this to be a great cause to refactor `env update` to use the same code path to make sure string substitution happens only in one place.

sanzoghenzo:
> We have an item in #11633 to "Combine conda env update with conda update and make the former an alias" which would be appropriate to try to do as prerequisite of #10018.
> 
> @sanzoghenzo You're right, I missed that this was related to the env subcommand as @travishathaway correctly pointed out. I would like to suggest for this to be a great cause to refactor `env update` to use the same code path to make sure string substitution happens only in one place.

Sounds right!
Unfortunately I don't think I'll have the time to tackle this... While waiting for the big refactor, is it OK if I go ahead with the changes of the `conda_env` part?

travishathaway:
@jezdez, do you believe we could still merge this while waiting for the work to unify `conda_env` and `conda` build?

travishathaway (approved):

jezdez (approved):

Issue #9497, opened 2019-12-11
conda 4.7.12 ProxyError - proxy server settings that work with conda 4.3 fail with conda 4.7.12


johngergely:
## Current Behavior

I am using conda behind a firewall. The repositories have been whitelisted and proxy server access is authorized for my desktop. I have used this configuration successfully for a long time with conda 4.3. In helping a new colleague configure his Anaconda installation with conda v 4.7.12, I have come to find out that the same proxy server settings fail with the new version.

### Steps to Reproduce

using conda 4.7.12
```
(base) PS C:\> conda update conda --verbose
Collecting package metadata (current_repodata.json): ...working... Retrying (Ret
ry(total=2, connect=None, read=None, redirect=None, status=None)) after connecti
on broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunnel connection
failed: 403 Forbidden'))': /pkgs/r/noarch/current_repodata.json

Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) a
fter connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunne
l connection failed: 403 Forbidden'))': /pkgs/msys2/win-64/current_repodata.json


Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) a
fter connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunne
l connection failed: 403 Forbidden'))': /pkgs/main/noarch/current_repodata.json

Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) a
fter connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunne
l connection failed: 403 Forbidden'))': /pkgs/r/win-64/current_repodata.json

Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) a
fter connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunne
l connection failed: 403 Forbidden'))': /pkgs/main/win-64/current_repodata.json


Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) a
fter connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunne
l connection failed: 403 Forbidden'))': /pkgs/msys2/noarch/current_repodata.json

Traceback (most recent call last):
  File "C:\Users\gerg\Anaconda_test\lib\site-packages\urllib3\connectionpool
.py", line 594, in urlopen
[...]

PR #12078, opened 2022-11-09
Add troubleshooting for ProxyError in user guide

tykutar:
### Description

Please refer to this issue -> https://github.com/conda/conda/issues/9497

I was stumped for a really long time when I encountered the same ProxyError. It turned out the solution was extremely simple. I think this addition will be highly beneficial for similar users.

### Checklist - did you ...

### DIFF
--- a/docs/source/user-guide/configuration/use-condarc.rst
+++ b/docs/source/user-guide/configuration/use-condarc.rst
@@ -347,6 +347,9 @@ here overrides that default:
   proxy_servers:
       http: http://user:pass@corp.com:8080
       https: https://user:pass@corp.com:8080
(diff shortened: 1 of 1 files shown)


travishathaway:
Hi @tykutar,

We appreciate your contribution, but we do not feel like this is an appropriate place for this particular tip.

If you look further down in the documentation, it says:

```
Be careful not to use ``http`` when you mean https or
``https`` when you mean http.
```

This clearly contradicts what your contribution suggests. Mixing these two schemes is not something that we want to encourage.

I hope this makes sense. Please re-open this pull request if you have any questions.

PR #12079, opened 2022-11-09
Initial Apple M1 testing with cirrus-ci

dbast:
### Description

Running tests natively on Apple M1 (osx-arm64, Apple Silicon).

Cirrus CI seems to be currently the only Apple M1 CI provider, which is free for OpenSource Projects ~~and requires enablement via https://github.com/marketplace/cirrus-ci for the org/repo~~(got enabled).

This currently runs one single marked Test Class with a Python version matrix (3.8, 3.9, 3.10). This can be extended over time by marking more test to run via cirrus in case there is any interest to run a test natively on osx-arm64.

~~Some results can be seen in my fork that got enabled for cirrus-ci, see e.g. https://github.com/dbast/conda/pull/1/checks?check_run_id=9385205816~~ Can now be seen as part of the CI runs in this PR.

TODO:
* ~~Cirrus CI enablement for the org/repo~~ Done
* ~~fixing the commented out "only_if:" statement~~ Done
* ~~investigate current failures, by fixing/changing setup.sh/unit.sh/integration.sh/tests.~~ switched to marked tests only
* ~~Extending the test task to test a matrix of TEST_GROUP/PYTHON/UNIT/INTEGRATION (can als be done with later PRs)~~ Python matrix is there.

The elegance of cirrus is that it can be also tested locally with full isolation, by running (initial run is slow due to the VM download, see also https://github.com/cirruslabs/tart):
* brew install cirruslabs/cli/cirrus
* cirrus run 

conda-forge also discussed the usage of cirrus-ci in https://conda-forge.org/docs/orga/minutes/2022-09-21.html

### Checklist - did you ...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,21 @@
+macos_instance:
+  image: ghcr.io/cirruslabs/macos-ventura-base:latest
+
+env:
+  CIRRUS_CLONE_DEPTH: '0'
+
+task:
+  name: macos-m1
+  #only_if: $CIRRUS_BRANCH == 'main' || $CIRRUS_BRANCH == 'feature/.*' || $CIRRUS_BRANCH =~ 'pull/.*'
+  environment:
+    PYTHON: '3.10'
+    TEST_SPLITS: '3'
+    TEST_GROUP: '1'
+  test_script: |
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -O ~/miniconda.sh
+    bash ~/miniconda.sh -b -p $HOME/miniconda3
+    source $HOME/miniconda3/etc/profile.d/conda.sh
+    conda create --name conda-test-env "python=${PYTHON}" -y
+    conda activate conda-test-env
+    ./dev/macos/setup.sh
+    ./dev/macos/unit.sh
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,6 +12,11 @@ task:
     TEST_SPLITS: '3'
     TEST_GROUP: '1'
 
+  info_script: |
+    sysctl -n hw.ncpu
+    vm_stat
+    df -h
+
   miniconda_script: |
     wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -O ~/miniconda.sh
     bash ~/miniconda.sh -b -p $HOME/miniconda3
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -37,4 +37,4 @@ task:
   test_script: |
     source $HOME/miniconda3/etc/profile.d/conda.sh
     conda activate conda-test-env
-    pytest -m "cirrus" -v
+    python -m pytest -m "cirrus" -v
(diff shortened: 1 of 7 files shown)


jezdez:
@dbast Nice!! I've added the app from the marketplace to see how it works..

.cirrus.yml
```diff
@@ -0,0 +1,21 @@
...
+env:
+  CIRRUS_CLONE_DEPTH: '0'
+
+task:
+  name: macos-m1
+  #only_if: $CIRRUS_BRANCH == 'main' || $CIRRUS_BRANCH == 'feature/.*' || $CIRRUS_BRANCH =~ 'pull/.*'
+  environment:
+    PYTHON: '3.10'
+    TEST_SPLITS: '3'
+    TEST_GROUP: '1'
```
jezdez, line 13: Remember that this'll only run the first test group of three, so this would not test the whole test suite.

dbast: Yeah, goal is to first get one test group running by harmonising the scripts to run for osx-64 on the GitHub Runners and osx-arm64 on Cirrus CI. 

Once that works, extending to all test types / suits / groups is the next step.

dbast:
The test suite runs... but there are failures... any ideas where they come from?

jaimergp:
👀  This is awesome!

@dbast - did you have any issues with concurrency and parallel jobs? IIRC, Cirrus gives the project a single VM with no isolation among jobs... but maybe I misread?

dbast:
@jaimergp My understanding so far was that cirrus-ci uses their own developed [tart-vm](https://github.com/cirruslabs/tart) (Tart uses Apple's own Virtualization.Framework for near-native performance) to split physical M1 MACs with e.g. 12 cores into 4 VMs with each 3 cores.... but this somehow contradicts what they currently state at https://cirrus-ci.org/faq/#are-there-any-limits 

But I did not try parallel / concurrent jobs yet.... the tests by this PR afaik all run serially.

dbast:
This PR works... but 8 out of 280 tests of the selected test group fail. The intention of the PR is not to fix all those. Running all unit + integration test via cirrus would also take too long, I assume. Should we instead run only some specific tests via cirrus which are especially relevant to M1 Macs? @conda/conda-core What do you think?

dbast:
@jaimergp test run tells `sysctl -n hw.ncpu`  = 4. so they split physical machines into several VMs each 4 cores.

.cirrus.yml
dbast: went a different route by running only specifically marked tests via cirrus. If we find out over time that there are enough resources maybe entire test suites can be enabled.

dbast:
Current cirrus error
```ERROR :: Running tests from a non-base Python interpreter.  Tests requires installing menuinst and that causes stderr  output when activated.``` 

strange

dholth:
#12273

dbast:
pre-commit.ci autofix

mikemhenry:
This is awesome!

jezdez:
Closing in favor of #13105 since the [limits for cirrus' runner for public repositories](https://cirrus-ci.org/faq/#are-there-any-limits) is pretty limiting (500 mins on macOS per month max). We'll find a different way to support Apple Silicon, maybe via the trial of Cirrus paid CI runners as mentioned in #13105.

PR #12090, opened 2022-11-11
features/repodata jlap

### LINKED ISSUE #11640: Implement .jlap incremental repodata
### Checklist

### What is the idea?

Implement jlap repodata to reduce bandwidth fetching repodata.json

### What should happen?

The conda HTTP cache needs to be replaced. We want to have a per-user cache that better understands http caching rules, and transparently applies diffs when available. This cache may be compatible with mamba. The pickled solver data might remain per-environment.

```[tasklist]
- [ ] https://github.com/conda/conda/pull/12003
- [ ] https://github.com/conda-incubator/ceps/pull/20
- [ ] https://github.com/conda/conda/pull/11447
- [ ] https://github.com/conda/conda/pull/11914
- [ ] https://github.com/conda/conda/pull/12090
- [ ] write unit tests / code coverage tests for branch
- [ ] asv benchmarks?
- [ ] Write test asserting "304 not modified" adds 30 seconds to the local cache validity (avoids checking again for 30s) instead of checking every time after the cache expires once
- [ ] locking: never write cached repodata.json without rewriting cache headers at the same time, while a lock is held. Necessary for jlap only, but wouldn't hurt to have correct locking always.
- [ ] Audit for XXX TODO markers
- [ ] Trace through the code to double-chechk "equivalent when jlap is not enabled" behavior
```

dholth:
### Description

Incremental repodata(.jlap) support on top of the features/repodata-gateway refactor.

Install jsonpatch, then

`CONDA_DEBUG=1 conda search --experimental=jlap <package>`

or add `experimental=jlap` to `~/.condarc`

Resolves #11640

### Checklist - did you ...


conda/base/context.py
```diff
@@ -327,6 +327,7 @@ class Context(Configuration):
     number_channel_notices = ParameterLoader(PrimitiveParameter(5, element_type=int))
     _verbosity = ParameterLoader(
         PrimitiveParameter(0, element_type=int), aliases=('verbose', 'verbosity'))
+    experimental_jlap = ParameterLoader(PrimitiveParameter(False))
```
jaimergp, line 330: One Thing I Regret from our libmamba experiment is having used `experimental` in the flags and object names. It caused unnecessary disruption and tons of deprecation warnings when we removed it. We were given the following advice but had to ignore it because we just wanted to release it without further changes. That was a mistake.

The advice was to have a `--experimental` namespace, that takes features that should be turned on. Only a single flag, only a single container in the `context`. This is what `pip` does with `--experimental-feature`.

dholth: It looks like I can do an append_list argument... and collect a list / set of feature strings

dholth: Or possibly a custom Action() class.

jaimergp: For reference, `pip`'s implementation: https://github.com/pypa/pip/blob/9aa422da16e11b8e56d3597f34551f983ba9fbfd/src/pip/_internal/cli/cmdoptions.py#L976

dholth: Thanks! I did find the choices argument and made this change. https://github.com/conda/conda/pull/12090/commits/1c7be2c192db4ad879d16c8e2b46cd4b932dce24

It is smart to have pip's naming that doesn't need to change if it becomes non-experimental though.

conda/testing/integration.py
```diff
@@ -388,6 +388,8 @@ def make_temp_channel(packages):
             f.write(json.dumps(repodata, cls=EntityEncoder))
         with open(join(noarch_dir, "repodata.json"), "w") as f:
             f.write(json.dumps({}, cls=EntityEncoder))
+        with open(join(noarch_dir, "repodata.json.bz2"), "w") as f:
+            f.write("required")
```
dholth, line 392: not actually required; will remove

dev/windows/setup.bat
```diff
@@ -17,7 +17,8 @@ cd \conda_src || goto :error
 CALL \conda_bin\scripts\activate.bat || goto :error
 CALL conda create -n conda-test-env -y python=%PYTHON% pywin32 --file=tests\requirements.txt || goto :error
 CALL conda activate conda-test-env || goto :error
-CALL conda update openssl ca-certificates certifi || goto :error
+:: make sure all test requirements are installed
+CALL conda install --quiet -y -c defaults --file tests\requirements.txt || goto :error
```
dholth, line 21: Did this line run?

conda/core/subdir_data.py
```diff
@@ -67,14 +79,17 @@ def __call__(cls, channel, repodata_fn=REPODATA_FN):
...
             cache_entry = SubdirData._cache_[cache_key]
-            if cache_key[0].startswith("file://"):
-                file_path = url_to_path(channel.url() + "/" + repodata_fn)
-                if exists(file_path):
-                    if cache_entry._mtime > getmtime(file_path):
+            if cache_key[0] and cache_key[0].startswith("file://"):
+                channel_url = channel.url()
+                if channel_url:
+                    file_path = url_to_path(channel_url + "/" + repodata_fn)
+                    if exists(file_path) and cache_entry._mtime > getmtime(file_path):
```
dholth, line 86: Should be `>=` ?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+
+        class _lock_impl:
+            def __init__(self, fd):
+                self.fd = fd
+
+            def __enter__(self):
+                for attempt in range(LOCK_ATTEMPTS):
+                    try:
+                        # msvcrt locking does something similar
+                        fcntl.lockf(self.fd, fcntl.LOCK_EX | fcntl.LOCK_NB, 1, LOCK_BYTE)
```
kenodegard, line 522: Don't we need to return here if the locking is successful?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+            self.load(state_only=True)
+        except FileNotFoundError:
+            self.state.clear()
+        return self.state
+
+    def save(self, data: str):
+        """
+        Write data to <repodata>.json cache path, synchronize state.
+        """
+        temp_path = self.cache_dir / self.cache_dir.with_suffix(f".{os.urandom(4).hex()}.tmp")
```
kenodegard, line 638: This doesn't seem quite right, do you mean to use `self.name`?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+
+        try:
+            with temp_path.open("x") as temp:  # exclusive mode, error if exists
+                temp.write(data)
```
kenodegard, line 642: Can we use `write_text` here?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+        """
+        Rename path onto <repodata>.json path, synchronize state.
+
+        Relies on path's mtime not changing on move. `temp_path` should be
+        adjacent to `self.cache_path_json` to be on the same filesystem.
+        """
+        with self.cache_path_state.open("a+") as state_file, _lock(state_file):
+            # "a+" avoids trunctating file before we have the lock and creates
+            state_file.seek(0)
+            state_file.truncate()
```
kenodegard, line 662: Since how the file is opened is important should we perhaps be doing the `open` within the lock context? something like:

```suggestion
        with _lock(self.cache_path_state) as state_file:
```

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+            max_age = context.local_repodata_ttl
+        elif context.local_repodata_ttl == 1:
+            max_age = get_cache_control_max_age(self.state.cache_control)
+        else:
+            max_age = 0
+
+        max_age *= 10**9  # nanoseconds
+        now = time.time_ns()
+        refresh = self.state.get("refresh_ns", 0)
+        return (now - refresh) > max_age
```
kenodegard, line 702: Reuse `timeout` here?

```suggestion
    def stale(self):
        """
        Compare state refresh_ns against cache control header and
        context.local_repodata_ttl.
        """
        return self.timeout() < 0
```

kenodegard (commented):
Some initial comments

conda/deprecations.py
```diff
@@ -6,7 +6,17 @@
...
+    from packaging.version import parse, Version
+except ImportError:
+    warnings.warn("`packaging` not available; deprecation warnings will not display.")
+
+    class Version:
+        def __lt__(self, other):
+            return True
+
+    def parse(version):
+        return Version()
```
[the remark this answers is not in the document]
dholth: It could be duplicated in another PR. Tests were failing hard without it.

setup.cfg
```diff
@@ -67,3 +79,4 @@ exclude_lines =
     raise NotImplementedError
     if __name__ == .__main__.:
     def __repr__
+    \s*\.\.\.$
```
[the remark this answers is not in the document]
dholth: This is three dots `...` indicating an empty function, in an ABC or interface class.

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+    def _lock_impl(fd):  # type: ignore
+        tell = fd.tell()
+        fd.seek(LOCK_BYTE)
+        msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK, 1)  # type: ignore
+        try:
+            fd.seek(tell)
+            yield
+        finally:
+            fd.seek(LOCK_BYTE)
+            msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
```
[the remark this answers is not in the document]
dholth: We do need to unlock the file no matter what

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+
+            def __exit__(self, *exc):
+                fcntl.lockf(self.fd, fcntl.LOCK_UN, 1, LOCK_BYTE)
+
+
+def _lock(fd):
+    if "jlap" in context.experimental or "lock" in context.experimental:
+        # locking required for jlap
+        return _lock_impl(fd)
+    return _lock_noop(fd)
```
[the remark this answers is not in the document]
dholth: Not yet

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+LOCK_ATTEMPTS = 10
+LOCK_SLEEP = 1
+
+
+@contextmanager
+def _lock_noop(fd):
+    """
+    When locking is not available.
+    """
+    yield
```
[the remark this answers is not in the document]
dholth:
```
>>> import contextlib
>>> foo = contextlib.nullcontext()
>>> with foo('bar', 'baz'):
...     print('hello')
TypeError: 'nullcontext' object is not callable
```

conda/gateways/repodata/__init__.py
dholth: Yes. It must complete the other attempts really quickly if the lock is already acquired!

conda/gateways/repodata/__init__.py
dholth: The purpose of `"x"` is to error if the supposedly new file already exists, so we can't use `write_text` here. Probably being too cautious.

conda/gateways/repodata/__init__.py
dholth: Maybe? I feel like there may have been a reason not to.

conda/gateways/repodata/__init__.py
dholth: timeout loses precision when dividing by `/ 1e9` and is only for a certain debug/log message. Probably not a big deal but it might be better to move or remove the log message.

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +317,434 @@ def conda_http_errors(url, repodata_fn):
...
+
+        return False, datetime.datetime.now(tz=datetime.timezone.utc)
+
+    def set_has_format(self, format: str, value: bool):
+        key = f"has_{format}"
+        self[key] = {
+            "last_checked": datetime.datetime.now(tz=datetime.timezone.utc).isoformat()[
+                : -len("+00:00")
+            ]
+            + "Z",
```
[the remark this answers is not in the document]
dholth: What is this change for

conda/core/subdir_data.py
```diff
@@ -578,12 +617,18 @@ def _process_raw_repodata(self, repodata, state=None):
         return _internal_state
 
 
+@deprecated("23.1", "23.5", addendum="Cache headers are now stored in a separate file.")
```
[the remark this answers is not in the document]
dholth: The separate file was added in 23.1, and the cache headers from here will always be unavailable.

conda/core/subdir_data.py
```diff
@@ -66,14 +80,17 @@ def __call__(cls, channel, repodata_fn=REPODATA_FN):
...
+                if channel_url:
+                    file_path = url_to_path(channel_url + "/" + repodata_fn)
+                    if exists(file_path) and cache_entry._mtime > getmtime(file_path):
                         return cache_entry
             else:
                 return cache_entry
-        subdir_data_instance = super().__call__(channel, repodata_fn)
+        subdir_data_instance = super().__call__(
+            channel, repodata_fn, RepoInterface=get_repo_interface()
+        )
```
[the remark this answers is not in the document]
dholth: This is a pattern that makes it easy to pass a mock RepoInterface to a constructor for testing, although we don't have a test that does that. `get_repo_interface` chooses whether we are using jlap or classic.

conda/gateways/repodata/__init__.py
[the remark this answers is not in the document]
dholth: `datetime.now(tz=timezone.utc).isoformat().sub` `AttributeError: 'str' object has no attribute 'sub'` I wanted to restrict searching for "+00:00", to only the end of the string.

.github/workflows/tests.yml
```diff
@@ -259,6 +259,7 @@ jobs:
           -e TEST_GROUP
           ghcr.io/conda/conda-ci:main-linux-python${{ matrix.python-version }}${{ matrix.default-channel == 'conda-forge' && '-conda-forge' || '' }}
           bash -c "source /opt/conda/etc/profile.d/conda.sh; \
+                   sudo /opt/conda/bin/conda install --quiet -y -c defaults --file tests/requirements.txt; \
```
dholth, line 262: I'm surprised this hasn't been merged already but we have to make sure our new test dependencies (flask) are installed, and the cached docker image might not have those.

conda/gateways/repodata/__init__.py
dholth: It breaks out of the loop now.

conda/core/subdir_data.py
```diff
@@ -355,25 +374,38 @@ def _load(self):
...
-            touch(self.cache_path_json)
-            _internal_state = self._read_local_repodata(mod_etag_headers)
+            cache.refresh()
+            # touch(self.cache_path_json) # not anymore, or the .state.json is invalid
+            # self._save_state(mod_etag_headers)
+            _internal_state = self._read_local_repodata(cache.state)
             return _internal_state
         else:
-            if not isdir(dirname(self.cache_path_json)):
+            if not isdir(dirname(self.cache_path_json)):  # What happens if it is a directory?
```
dholth, line 383: could we just mkdir_p all the time

conda/core/subdir_data.py
```diff
@@ -355,25 +374,38 @@ def _load(self):
...
                 mkdir_p(dirname(self.cache_path_json))
             try:
-                cache_path_json = self.cache_path_json
-                with io_open(cache_path_json, "w") as fh:
-                    fh.write(raw_repodata_str or "{}")
-                # quick thing to check for 'json matches stat', or store, check a message digest:
-                mod_etag_headers["mtime"] = pathlib.Path(cache_path_json).stat().st_mtime
-                self._save_state(mod_etag_headers)
+                if raw_repodata_str is RepodataOnDisk:
+                    # this is handled very similar to a 304. Can the cases be merged?
```
dholth, line 387: We should eventually push more out to RepodataCache so that SubdirData only does package searching. Should be simpler and avoid some duplicate read/write/parse cycles

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +315,436 @@ def conda_http_errors(url, repodata_fn):
...
+        except (KeyError, ValueError, TypeError) as e:
+            log.warn("error parsing `has_` object from `<cache key>.state.json`", exc_info=e)
+            self.pop(key)
+
+        return False, datetime.datetime.now(tz=datetime.timezone.utc)
+
+    def set_has_format(self, format: str, value: bool):
+        key = f"has_{format}"
+        self[key] = {
+            "last_checked": datetime.datetime.now(tz=datetime.timezone.utc).isoformat()[
```
dholth, line 441: utcnow should not be used, this is the new way according to the pytz maintainer

conda/gateways/repodata/jlapper.py
```diff
@@ -0,0 +1,420 @@
...
+
+class Jlap304NotModified(Exception):
+    pass
+
+
+class JlapSkipZst(Exception):
+    pass
+
+
+def process_jlap_response(response: Response, pos=0, iv=b""):
```
dholth, line 62: mark

conda/gateways/repodata/__init__.py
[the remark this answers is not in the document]
kenodegard: this can be disregarded, I hadn't meant to include this idea

conda/core/subdir_data.py
```diff
@@ -618,26 +663,11 @@ def make_feature_record(feature_name):
     )
 
 
-def cache_fn_url(url, repodata_fn=REPODATA_FN):
```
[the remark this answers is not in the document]
kenodegard: Ah I missed that 👍🏻

conda/gateways/repodata/__init__.py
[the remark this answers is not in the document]
kenodegard: that's not how you use it, you'd end up with doing the following instead of defining `_lock_noop`:

```
def _lock(fd):
    if "jlap" in context.experimental or "lock" in context.experimental:
        # locking required for jlap
        return _lock_impl(fd)
    return contextlib.nullcontext()
```

conda/gateways/repodata/__init__.py
dholth: The difference between `r+` and `a+` that matters to us, is that `r+` errors if the file does not exist while `a+` will create the file, and seek to the end of the file. `w+` will truncate and seek to the beginning, which is not something we want to do before we have the lock.

tests/test_deprecations.py
```diff
@@ -1,10 +1,11 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+import importlib
 import sys
 
 import pytest
 
-from conda.deprecations import DeprecationHandler, DeprecatedError
+from conda.deprecations import DeprecatedError, DeprecationHandler
```
dholth, line 8: a sorted import is a happy import

conda/gateways/repodata/__init__.py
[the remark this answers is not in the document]
dholth: Since we use the same function for "locking wanted but not supported on platform" it is simpler to retain `_lock_noop`

conda/deprecations.py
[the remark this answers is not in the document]
dholth: moved to a separate refactors PR

setup.cfg
```diff
@@ -42,6 +42,18 @@ ignore = E126,E133,E226,E241,E242,E302,E704,E731,E722,W503,E402,W504,F821,E203
 exclude = .asv*,build/*,.tox/*,dev*/*,env/*,test_data/*,tests/*,ve/*,utils/*,*/_vendor/*,conda/compat.py,conda/common/compat.py,conda_env/compat.py
 
 
+[coverage:run]
+# store relative paths in coverage information
+relative_files = True
```
dholth, line 47: This is an attempt to run coverage in and out of docker, and combine the results. It doesn't work perfectly but it is better than nothing.

tests/test_resolve.py
```diff
@@ -18,10 +18,6 @@
 from conda.testing.helpers import TEST_DATA_DIR, add_subdir, add_subdir_to_iter, \
     get_index_r_1, get_index_r_4, raises
```
dholth, line 20: It turns out this test was duplicated in the pluggable solvers refactor, but this original was not removed. Another branch tests removing this file, which would get rid of a lot of seemingly spontaneous test failures.

tests/test_deprecations.py
```diff
@@ -152,3 +153,22 @@ def test_topic_remove(deprecated_v3):
     # alerting developer that a module needs to be removed
     with pytest.raises(DeprecatedError):
         deprecated_v3.topic("2.0", "3.0", topic="Some special topic")
+
+
+def test_packaging_unavailable(mocker):
```
dholth, line 158: Another branch skips this test.

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -0,0 +1,402 @@
...
+            for block in response.iter_content(chunk_size=1 << 14):
+                repodata.write(block)
+    if response.request:
+        log.info("Download %d bytes %r", length, response.request.headers)
+    return response  # can be 304 not modified
+
+
+def request_url_jlap_state(
+    url, state: RepodataState, get_place=get_place, full_download=False, *, session: Session
+):
```
kenodegard, line 247: This is a sufficiently complex function and needs to be split up into smaller units

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -0,0 +1,402 @@
...
+def withext(url, ext):
+    return re.sub(r"(\.\w+)$", ext, url)
+
+
+@contextmanager
+def timeme(message):
+    begin = time.monotonic()
+    yield
+    end = time.monotonic()
+    log.debug("%sTook %0.02fs", message, end - begin)
```
kenodegard, line 186: I'm wondering if this is better suited as a monkey patch of `logging.Logger` since this would be useful to have elsewhere in conda:

```
@contextmanager
def timeme(message, *args, **kwargs):
    begin = time.monotonic()
    yield
    end = time.monotonic()
    self.debug(message + " Took %0.02fs", *args, end - begin, **kwargs)

logging.Logger.timeme = timeme
del timeme
```

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -0,0 +1,402 @@
...
+    if response.status_code == 304:
+        raise Jlap304NotModified()
+
+    def lines() -> Iterator[bytes]:
+        yield from response.iter_lines(delimiter=b"\n")  # type: ignore
+
+    buffer = JLAP.from_lines(lines(), iv, pos)
+
+    # new iv == initial iv if nothing changed
+    pos, footer, _ = buffer[-2]
```
kenodegard, line 78: All of these buffer indices (`-1`, `-2`, and `-3`), are effectively magic numbers as it isn't immediately clear what these indices are referring to. Since the buffer object is an instance of a custom list implementation can we create properties instead that return this information in a way that is more descriptive?

conda/core/subdir_data.py
```diff
@@ -3,59 +3,76 @@
...
+    if "jlap" in context.experimental:
+        try:
+            from conda.gateways.repodata.jlap.interface import JlapRepoInterface
+
+            return JlapRepoInterface
+        except ImportError as e:  # pragma: no cover
+            warnings.warn(
+                "Could not load the configured jlap repo interface. "
+                f"Is the required jsonpatch package installed?  {e}"
+            )
```
kenodegard, line 71: I'm thinking we should mark `jsonpatch` as a dependency (in the recipe and in `requirements.txt`) instead of including this extra try-except.

conda/gateways/repodata/jlap/core.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    lines: iterator over input split by '\n', with '\n' removed.
+    pos: initial position
+    """
+    for line in lines:
+        yield pos, line
+        pos += len(line) + 1
+
+
+class JLAP(UserList):
```
kenodegard, line 39: Since we never access the internal `data` attribute here there's no benefit to using `collections.UserList` over `list[tuple[int, str, str]]`:

```suggestion
class JLAP(list[tuple[int, str, str]]):
```

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -0,0 +1,402 @@
...
+from conda.base.context import context
+from conda.gateways.connection import Response, Session
+from conda.gateways.repodata import RepodataState
+
+from .core import JLAP
+
+log = logging.getLogger(__name__)
+
+
+DIGEST_SIZE = 32  # 160 bits a minimum 'for security' length?
```
kenodegard, line 30: This is also defined in `conda.gateways.repodata.jlap.core`, would it help to move all of the jlap constants into a `conda.gateways.repodata.jlap.constants` submodule?

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -0,0 +1,402 @@
...
+LATEST = "latest"
+JLAP_UNAVAILABLE = "jlap_unavailable"
+ZSTD_UNAVAILABLE = "zstd_unavailable"
+
+
+def hash():
+    """
+    Ordinary hash.
+    """
+    return blake2b(digest_size=DIGEST_SIZE)
```
kenodegard, line 45: This is a constant. Why isn't it stored as one?

kenodegard (changes requested):
I think I fully understand what's happening here now, I like where this is going

conda/gateways/repodata/jlap/fetch.py
dholth: This creates a new hasher object each time.

conda/gateways/repodata/jlap/core.py
dholth: I did try extending list earlier but no that doesn't behave as desired. Iirc you wind up covering back to base list by mistake.

conda/gateways/repodata/jlap/fetch.py
dholth: I would like to reformulate it as a formal state machine.

conda/gateways/repodata/jlap/fetch.py
dholth: This code is absolutely not focused on being general code that is useful elsewhere in conda (apart from its intendend use as a repodata adapter). I think logging does have a built-in timestamp-between-log-messages features.

conda/core/subdir_data.py
dholth: Maybe. How experimental is the feature? jsonpatch is 14kb and its dependency jsonpointer is 11kb.

conda/gateways/repodata/jlap/fetch.py
dholth: This digest size is used for a different purpose (digest of the reconstituted repodata.json file) and the jlap `core` keyed hash digest size is used for the integrity of the individual lines in the jlap file. They don't necessarily have to be the same size, even though they are in this application.

conda/gateways/repodata/jlap/fetch.py
dholth: @kenodegard would you like to take on the refactor for this function, submitting it back as a PR on top of this one? It should be easier now that there is only one download-and-hash(is_zst=True/False) function.

conda/gateways/repodata/jlap/fetch.py
kenodegard: Maybe I'm missunderstanding but I see this:

```python
>>> from hashlib import blakeb2
>>> blake2b(digest_size=32).digest() == blake2b(digest_size=32).digest()
True
```

conda/gateways/repodata/jlap/fetch.py
kenodegard: Yes, logging includes timestamps, but it's not quite the same as timing something.

If we don't think this should be general code, then I don't think we should introduce it here in the first place either.

conda/gateways/repodata/jlap/fetch.py
kenodegard: If they're used for different things and mean different things they need to be called different things. Including a docstring would also help.

conda/gateways/repodata/jlap/fetch.py
dholth: something like this

```
b2 = blake2b(digest_size=32)
b2.update(b'some data')
b2.digest()

b2.update(b'some data')
b2.digest()
```

conda/gateways/repodata/jlap/fetch.py
dholth: On the one hand I see what you mean, on the other hand the context is that they are in different files. Can rename the constant.

conda/gateways/repodata/jlap/fetch.py
kenodegard: ah, of course

what do we gain by defining this `hash` helper function instead of inlining the `blake2d` initialization?

dholth: we never pass the wrong digest size by mistake, and could change the hash function to blake3 or whatever if it becomes availabel

conda/core/subdir_data.py
dholth: In progress over here https://github.com/conda/conda/pull/12461

conda/gateways/repodata/jlap/fetch.py
dholth: Like https://github.com/conda/conda/pull/12090/commits/4e70c5e4168342d8610be0dbd091005a25f91d16 ?

conda/core/subdir_data.py
dholth: I think I agree it makes sense to just depend on jsonpatch all the time. Added to https://github.com/conda/conda/pull/12461 which is intended to be the next step after this `features/repodata-jlap`

conda/gateways/repodata/jlap/fetch.py
dholth: It is standard Python that -1 is the last element of the list, -2 is the next to last and so on...?

jezdez (approved):
I accidentally reviewed a very similar branch in https://github.com/conda/conda/pull/12425 and many if not all suggestions had been transferred here. Since most of the code is behind a feature flag, I think we should merge and then iterate over it more (e.g. #12461)

conda/core/subdir_data.py
```diff
@@ -262,31 +292,27 @@ def _iter_records_by_name(self, name):
...
+        assert Path(state.cache_path_state) == Path(self.cache_path_state)
+        assert state.repodata_fn == self.repodata_fn
+        return state.save()
 
     def _load(self):
-        try:
-            mtime = getmtime(self.cache_path_json)
-        except OSError:
+        cache = self.repo_cache
+        cache.load_state()  # XXX should this succeed even if FileNotFound?
```
dholth, line 311: Pretty sure this does succeed on file not found, with RepoCache/RepodataState creating the state file if it is not available.

conda/core/subdir_data.py
dholth: Since the follow-on PR improves the RepodataOnDisk situation, we could have it return a str here always even when that sacrifices efficiency. Then this soon-to-be-eliminated version of the code would be a little simpler.

sven6002 (approved):
Looks good to me

PR #12099, opened 2022-11-14
Fix More Crashes in MatchSpec and Version Parsing

pkmooreanaconda:
### Description

Further fuzzing efforts have identified the following cases crashing cases.  More details once I chase them down.

"(-cuda*|*cnumpy==1.0uda|????*cu"
"numpy= " (the trailing space is important here)
"aults::n [channel=ana.conda,versiond=3]"
"a[=)(=b"

Found another while looking at this:
".*!=1.9,^)\$=0"

Interestingly, it only crashes when it is passed to VersionSpec via MatchSpec, not directly.

### Checklist

### DIFF
--- a/conda/common/url.py
+++ b/conda/common/url.py
@@ -403,7 +403,7 @@ def split_conda_url_easy_parts(known_subdirs, url):
     cleaned_url, token = split_anaconda_token(url)
     cleaned_url, platform = split_platform(known_subdirs, cleaned_url)
     _, ext = strip_pkg_extension(cleaned_url)
-    cleaned_url, package_filename = cleaned_url.rsplit('/', 1) if ext else (cleaned_url, None)
+    cleaned_url, package_filename = cleaned_url.rsplit('/', 1) if ext and '/' in cleaned_url else (cleaned_url, None)
 
     # TODO: split out namespace using regex
     url_parts = urlparse(cleaned_url)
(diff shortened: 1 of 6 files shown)


conda/common/url.py
```diff
@@ -403,7 +403,7 @@ def split_conda_url_easy_parts(known_subdirs, url):
     cleaned_url, token = split_anaconda_token(url)
     cleaned_url, platform = split_platform(known_subdirs, cleaned_url)
     _, ext = strip_pkg_extension(cleaned_url)
-    cleaned_url, package_filename = cleaned_url.rsplit('/', 1) if ext else (cleaned_url, None)
+    cleaned_url, package_filename = cleaned_url.rsplit('/', 1) if ext and '/' in cleaned_url else (cleaned_url, None)
```
pkmooreanaconda, line 406: I am not at all convinced that this is the correct way to solve this crash.  We are allowing some very messed up URLs and then they fail once the hit the HTTP(S) machinery further along in execution.  I played around with creating an "InvalidUrl" exception but ran into a wall getting it to import correctly.

Side note, I think this import issue is why we're raising ValueError instead of a more meaningful error in this file but that's an adventure for another day.

travishathaway (changes requested):
I'm hesitant to accept this pull request with out test coverage of the newly added exceptions. I tried messing around with `treeify` a little to see if I could trigger an exception, but wasn't able (otherwise, I would have committed the test case directly this pull request).

How difficult is it to trip these new exceptions?

jezdez (dismissed):

jezdez:
@travishathaway Do you feel your review concerns were taken care of?

travishathaway (approved):

Issue #11342, opened 2022-03-16
Move `conda` out of `base` env


kenodegard:
### Summary

There is a long history of issues where users install anything and everything into their `base` environment (previously called the `root` environment) which invariably at some point (or after running `conda update conda`) results in some `conda` dependency being upgraded (or downgraded) and `conda` itself breaking. Efforts have been made in the past to circumvent this issue by vendoring `conda`'s dependencies (see https://github.com/conda/conda/issues/2825) but to this day `conda` is still not fully siloed and independent.

The idea for splitting `conda` out of the `base`/`root` environment is an old one (predating 2016) and has been earmarked as a conda 5.0 feature for quite some time.

#### Proposal
1. Define a new "class" of environments: hidden environments
    a. Most likely would just be an environment prefixed with a `.`
    b. Modifying `conda info --envs` and `conda env list` to hide hidden envs by default
4. Allow `conda` to be run/shell functions initialized out of any environment
5. Define a new `conda` command/args for upgrading `conda`:
    ```bash
    conda update --conda

    # the above proposal is effectively the same as the following:
    conda update --prefix HIDDEN_CONDA_ENV conda

    # adding this new `--conda` argument allows us to make a
    # more clean interface for say installing plugins:
    conda install --conda conda-doctor

    # or even to make it convenient to migrate existing installs to the
    # new siloed env setup:
    conda create --conda
    ```
6. Modify installer to install with a base env and a hidden conda env

#### Related ideas
- https://github.com/conda/conda/issues/4929
- https://github.com/conda/conda/issues/4943

kenodegard:
micromamba has already moved in this direction.

> The lack of a "base" environment makes things less brittle vs. traditional conda/mamba."
> https://link.medium.com/dCB8wwNzQob

LtDan33:
This is a great CEP candidate. That would be the next step.

jaimergp:
I've started some experiments in this direction:

- Idea for installers: supply a `default` env (see https://github.com/conda-forge/miniforge/issues/636).
- Idea for `conda`: lock `base` by default ([see this logic](https://github.com/conda/conda/pull/14113/commits/fbfad1610c8bcc14b6f4c36dae3bde45a8f9c8b8)) or, even better, come up with a CEP for PEP-668 style locking (EXTERNALLY-MANAGED for conda).
[...]

Issue #3912, opened 2016-11-18
Private application environments


kalefranz:
In the future, we would like to offer a miniconda installer that installs conda into a *private env*, while leaving the *default user env* empty except for links of conda's executables into the default user env. It would be naive and a disservice to users to treat conda as a special case. 

Private named environments are generally useful for any "top-level application" that the prototypical user is likely to install only once on the system.  Most users, for example, wouldn't (intentionally) install multiple instances of Firefox on their systems.  Advantageously, an application may leverage explicitly pinned dependencies for its own operational safety without worrying about creating unsatisfiability in the user's broader default environment.  Examples could include Anaconda Navigator and Spyder.

The convention for a private application environment will be the application's name, prefixed and suffixed by an underscore. This distinguishes a private environment from the established use case of an environment name prefixed by underscore (prefixed only, not suffixed); the prefix-only convention allows conda to be installed in the environment without complaint.

An application may request that it be installed in a private named environment by indicating so in [non-index package metadata](https://github.com/conda/conda-build/pull/1539).  The request will be respected for conda create, install, update, and remove commands when

* the *user default env* is the active environment, and
* the user does not specify the `--name` or `--prefix` command-line flags.

Private environment preference will only be respected for packages explicitly requested by the user.  Packages installed as dependencies of a user-requested package will not be installed in private environments, but rather the environment into which the user-requested package is installed.

When a package specifies a preferred private environment, it need not be a private environment for itself.  It can instead be the private environment of another package.  For example, conda-build can indicate that its preferred private environment is the conda private environment.

When a private environment request is respected, entry points are created in the *user default environment*, linking executable-marked paths in the private prefix's `bin`/`Scripts` directory. The entry point for an `application` executable will be similar to 

```
#!/bin/bash
CONDA_ROOT="$(dirname $0).."
source "$CONDA_ROOT/bin/activate" "$CONDA_ROOT/envs/_myapp_"
exec "$CONDA_ROOT/envs/_myapp_/bin/myapp" "$@"
```

A package may limit the executables being linked by providing an explicit list of desired paths to symlink in the package's non-index metadata.

To track user-requested packages that have been installed in private environments, a file
`conda-meta/private_envs.json` is used.  It is present only in the *user default env* `conda-meta` directory.  This tracking is necessary to unlink packages (either during update or remove) that have been installed in private environments.

The issue https://github.com/conda/conda/issues/3074 discusses the more-limited conda-specific implementation.

mingwandroid:
[...]

PR #12101, opened 2022-11-14
Deprecate private conda env

kenodegard:
### Description

First introduced in conda 4.2 and updated in 4.3, conda introduced a private conda environment (`_conda` and `_conda_` respectively). This special env appears to have been a special local testing environment prior developers of conda used. This private env is undocumented, no longer in use, and adds an undesired complexity so we will deprecate it.

Xref: https://github.com/conda/conda-build/pull/4629

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -8,7 +8,7 @@ from functools import lru_cache
 from logging import getLogger
 from typing import Optional
 import os
-from os.path import abspath, basename, expanduser, isdir, isfile, join, split as path_split
+from os.path import abspath, expanduser, isdir, isfile, join, split as path_split
 import platform
 import sys
 import struct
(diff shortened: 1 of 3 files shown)


conda/base/context.py
```diff
@@ -475,7 +475,13 @@ def arch_name(self):
 
     @property
     def conda_private(self):
-        return conda_in_private_env()
+        warnings.warn(
+            "`conda.base.context.context.conda_private` is pending deprecation and will be "
+            "removed in a future release. It's meaningless and any special meaning it may have "
```
mbargull, line 480: Because of the wording here, I'm unsure if the history about this is known to everyone.

This was a feature planned for 4.4 which would've helped (and still would) to protect the Conda installation against unintentional breakage. The idea was to make the `base` (back then `root`) environment empty (or populated by the non-`conda` deps of the `anaconda` package) and have `conda` installed into `_conda_`, neatly tucked away from everything else.

This was conceptually/structurally a big change that touched a lot of different bits and pieces, i.e., it was not an easy feat but ultimately/unfortunately resource constraints hindered its completion.
Though by now, saying it doesn't have much meaning anymore is apt, I guess.

See https://github.com/conda/conda/issues/3912 for context.

mbargull: (And right after writing this I saw https://github.com/conda/conda/issues/11342 mentioned https://github.com/conda/conda/issues/3912 :smile: .)

PR #12105, opened 2022-11-15
Suggestions for the original pull request: 12003

travishathaway:
Here are my suggestions for improvement:

- I think we should divorce the SubdirData and CondaRepoInterface objects from each other. This coupling doesn't seem worth it. If you decide to keep this coupling, please make the property public instead of protected, so we know it's okay to use this property from outside the `SubdirData` object.
- In `conda/gateways/repodata/__init__.py` I chose to remove the `**kwargs` from the `__init__` method of the `CondaRepoInterface` object because they were not being used at.
- Added some more type hinting where necessary.

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -239,8 +239,8 @@ class SubdirData(metaclass=SubdirDataType):
             log.debug("Could not load state", exc_info=True)
             return {}
 
-    def _save_state(self, state: dict) -> None:
-        pathlib.Path(self.cache_path_state).write_text(json.dumps(state, indent=True))
+    def _save_state(self, state: dict):
+        return pathlib.Path(self.cache_path_state).write_text(json.dumps(state, indent=True))
 
     def _load(self):
         try:
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -25,7 +25,6 @@ from genericpath import getmtime, isfile
 from conda.common.iterators import groupby_to_dict as groupby
 from conda.gateways.repodata import (
     CondaRepoInterface,
-    RepoInterface,
     RepodataIsEmpty,
     Response304ContentUnchanged,
 )
@@ -50,7 +49,6 @@ from ..models.records import PackageRecord
 from ..trust.signature_verification import signature_verification
 
 log = getLogger(__name__)
-stderrlog = getLogger("conda.stderrlog")
 
 REPODATA_PICKLE_VERSION = 29
 MAX_REPODATA_VERSION = 1
@@ -169,16 +167,6 @@ class SubdirData(metaclass=SubdirDataType):
         self._loaded = False
         self._key_mgr = None
 
-    @property
-    def _repo(self) -> RepoInterface:
-        # XXX one for `current_repodata.json` and one for `repodata.json`
-        return CondaRepoInterface(
-            self.url_w_credentials,
-            self.repodata_fn,
-            cache_path_json=self.cache_path_json,
-            cache_path_state=self.cache_path_state,
-        )
-
     def reload(self):
         self._loaded = False
         self.load()
@@ -238,19 +226,21 @@ class SubdirData(metaclass=SubdirDataType):
             self.load()
         return iter(self._package_records)
 
-    # replaces old in-band mod_and_etag headers
-    def _load_state(self):
+    def _load_state(self) -> dict:
+        """
+        Replaces old in-band mod_and_etag headers
+        """
         try:
             state_path = pathlib.Path(self.cache_path_state)
             state = json.loads(state_path.read_text())
             log.debug("Load state from %s", state_path)
             return state
-        except:
+        except (json.JSONDecodeError, OSError):
             log.debug("Could not load state", exc_info=True)
             return {}
 
-    def _save_state(self, state):
-        return pathlib.Path(self.cache_path_state).write_text(json.dumps(state, indent=True))
+    def _save_state(self, state: dict) -> None:
+        pathlib.Path(self.cache_path_state).write_text(json.dumps(state, indent=True))
 
     def _load(self):
         try:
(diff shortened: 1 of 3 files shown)


conda/gateways/repodata/__init__.py
```diff
@@ -57,14 +55,21 @@ class Response304ContentUnchanged(Exception):
...
+    Provides an interface for retrieving repodata data from channels
+    """
 
-        self._log = logging.getLogger(__name__)
-        self._stderrlog = logging.getLogger("conda.stderrlog")
+    #: Channel URL
+    _url: str
 
-    def repodata(self, state: dict) -> str | None:
+    #: Filename of the repodata file; defaults to value of conda.base.constants.REPODATA_FN
```
dholth, line 65: does sphinx pick up the #: on the previous line; when I was documenting conda-package-handling I had trouble getting this to work https://github.com/conda/conda-package-handling/blob/main/src/conda_package_handling/api.py#L31

conda/core/subdir_data.py
```diff
@@ -238,18 +226,20 @@ def iter_records(self):
...
+        """
+        Replaces old in-band mod_and_etag headers
+        """
         try:
             state_path = pathlib.Path(self.cache_path_state)
             state = json.loads(state_path.read_text())
             log.debug("Load state from %s", state_path)
             return state
-        except:
+        except (json.JSONDecodeError, OSError):
```
dholth, line 238: nice

conda/gateways/repodata/__init__.py
```diff
@@ -57,14 +55,21 @@ class Response304ContentUnchanged(Exception):
 
 
 class CondaRepoInterface(RepoInterface):
-    def __init__(self, url: str, repodata_fn: str | None, **kwargs) -> None:
```
dholth, line 60: It takes **kwargs because we want to pass extra properties to the subclass in features/repodata-jlap

conda/gateways/repodata/__init__.py
```diff
@@ -84,21 +89,21 @@ def repodata(self, state: dict) -> str | None:
...
             return ensure_text_type(
-                bz2.decompress(resp_content) if filename.endswith(".bz2") else resp_content
+                bz2.decompress(resp_content) if _filename.endswith(".bz2") else resp_content
             ).strip()
 
-        json_str = maybe_decompress(filename, resp.content)
+        json_str = decompress_if_bzip2(filename, resp.content)
 
-        # We explictly no longer add these tags to the large `resp.content` json
+        # We no longer add these tags explicitly to the large `resp.content` json
```
dholth, line 106: we can remove the word explicitly

conda/gateways/repodata/__init__.py
```diff
@@ -84,21 +89,21 @@ def repodata(self, state: dict) -> str | None:
...
             resp.raise_for_status()
 
         if resp.status_code == 304:
             raise Response304ContentUnchanged()
 
-        def maybe_decompress(filename, resp_content):
+        def decompress_if_bzip2(_filename, resp_content):
             return ensure_text_type(
-                bz2.decompress(resp_content) if filename.endswith(".bz2") else resp_content
+                bz2.decompress(resp_content) if _filename.endswith(".bz2") else resp_content
```
dholth, line 101: should be fine without the added underscore? we'll remember to remove the python2/3 transition method ensure_text_type

conda/gateways/repodata/__init__.py
```diff
@@ -57,14 +55,21 @@ class Response304ContentUnchanged(Exception):
...
 
-    def repodata(self, state: dict) -> str | None:
+    #: Filename of the repodata file; defaults to value of conda.base.constants.REPODATA_FN
+    _repodata_fn: str | None
+
+    def __init__(self, url: str, repodata_fn: str | None) -> None:
+        self._url = url
+        self._repodata_fn = repodata_fn or REPODATA_FN
+
+    def fetch_repodata(self, state: dict) -> str | None:
```
dholth, line 72: We might fetch it or we might load it from cache.

conda/core/subdir_data.py
```diff
@@ -169,16 +167,6 @@ def __init__(self, channel, repodata_fn=REPODATA_FN):
...
 
-    @property
-    def _repo(self) -> RepoInterface:
-        # XXX one for `current_repodata.json` and one for `repodata.json`
-        return CondaRepoInterface(
-            self.url_w_credentials,
-            self.repodata_fn,
-            cache_path_json=self.cache_path_json,
-            cache_path_state=self.cache_path_state,
-        )
```
dholth, line 180: we did wind up using this twice, giving us a single place to update when we need to pass different arguments for different parameters to CondaRepoInterface as might be needed for jlap. Should we have a test for this method

conda/gateways/repodata/__init__.py
travishathaway: I've never actually tried run it through Sphinx 🤔. TBH, I've always gone with a doc string directly below the variable name, but recently I was advised to start using this format...

Regardless, this isn't actually documented anywhere in any style guide, so it doesn't matter. It would be very nice to have a documenting style guide for this project though.

conda/gateways/repodata/__init__.py
travishathaway: I did this because my IDE was complaining about this variable also being defined in the outer scope too. Not a deal breaker though.

conda/core/subdir_data.py
```diff
@@ -238,18 +226,20 @@ def iter_records(self):
             self.load()
```
jezdez, line 226:
```suggestion
            self.load(this)
```

conda/core/subdir_data.py
```diff
@@ -238,18 +226,20 @@ def iter_records(self):
...
         return iter(self._package_records)
 
-    # replaces old in-band mod_and_etag headers
-    def _load_state(self):
+    def _load_state(self) -> dict:
+        """
+        Replaces old in-band mod_and_etag headers
+        """
```
jezdez, line 232:
```suggestion
    def _load_state(self) -> dict:
        """
        Replaces this old in-band mod_and_etag headers
        """
```

conda/gateways/repodata/__init__.py
kenodegard: I know the `#:` syntax will work for global vars and class vars.

[Sphinx autodoc docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)

I don't think it's possible (or desirable) to document local variables so a plain `#` is best here.

dholth:
Thanks!

PR #12125, opened 2022-11-23
Implement `conda.deprecations.deprecated` decorator

kenodegard:
### Description

Added a new decorator to simplify the deprecation process by automating the transition from pending deprecation to deprecated and then raising `DeprecatedError` once the code is expected to be removed.

There are several parts of `conda.deprecations.deprecated`. It can be used as:

1. a decorator deprecating entire function, method, or class (e.g., https://github.com/conda/conda/pull/12125/files#diff-ade4c48ccdde7569779251423b096e113c2c7f4481f325967511e761447ff7e8R11-R12):

    ```python
    # some_module.py

    @deprecated("23.3.0", "23.9.0")
    def some_function(one, two, three):
        ...
    ```

    ```pycon
    >>> import some_module
    >>> some_module.some_function()
    <stdin>:1: PendingDeprecationWarning: some_module.some_function is pending deprecation and will be removed in 23.9.0.
    ```

2. a decorator deprecating keyword arguments (e.g., https://github.com/conda/conda/pull/12125/files#diff-cb683bb6edba28f0584eb5733b791fb794ca440c217992b51e3781a60ad592feR112-R113):

    ```python
    # some_module.py

    # deprecate argument
    @deprecated.argument("23.3.0", "23.9.0", "three")
    def some_function1(one, two):
        ...

    # rename argument
    @deprecated.argument("23.3.0", "23.9.0", "three", rename="third")
    def some_function2(one, two, third):
        ...
    ```

    ```pycon
    >>> import some_module
    >>> some_module.some_function1(1, 2, three=3)
    <stdin>:1: PendingDeprecationWarning: some_module.some_function1(three) is pending deprecation and will be removed in 23.9.0.

    >>> some_module.some_function2(1, 2, three=3)
    <stdin>:1: PendingDeprecationWarning: some_module.some_function2(three) is pending deprecation and will be removed in 23.9.0. Use 'third' instead.
    ```

3. a function deprecating modules (e.g., https://github.com/conda/conda/pull/12125/files#diff-d16b69da5d3b68d0b16f2a607e13462a782dd90dd2d1143ed42a71f53bfc0891R3):

    ```python
    # some_module.py

    deprecated.module("23.3.0", "23.9.0")
    ```

    ```pycon
    >>> import some_module
    <stdin>:1: PendingDeprecationWarning: some_module is pending deprecation and will be removed in 23.9.0.
    ```

4. a function deprecating constants/globals:

    ```python
    # some_module.py

    deprecated.constant("23.3.0", "23.9.0", "SOME_CONSTANT", 42)
    ```

    ```pycon
    >>> import some_module
    >>> some_module.SOME_CONSTANT
    <stdin>:1: PendingDeprecationWarning: some_module.SOME_CONSTANT is pending deprecation and will be removed in 23.9.0.
    ```

5. a function deprecating a topic (e.g., https://github.com/conda/conda/pull/12125/files#diff-c2c5d10a71b1188c28e3e8c3350d01635574004be719683cf7203045fed4ce3cR165-R169):

    ```python
    # some_module.py

    def logic(...):
        if ...:
            pass
        else:
            deprecated.topic("23.3.0", "23.9.0", topic="This feature/logic path/topic")
    ```

    ```pycon
    >>> import some_module
    >>> some_module.logic(...)
    some_module.py:7: PendingDeprecationWarning: This feature/logic path/topic is pending deprecation and will be removed in 23.9.0.
    ```

Xref: https://github.com/conda/conda/issues/12276

### Checklist - did you ...


dholth:
Will this work for our function that is only deprecated if you pass an old keyword argument

kenodegard:
> Will this work for our function that is only deprecated if you pass an old keyword argument

Added the ability to deprecate _keyword_ arguments via `@conda._deprecate.argument`.

conda/__init__.py
```diff
@@ -148,3 +148,85 @@ def _default(self, obj):
...
+        _current: str | None = None,
+    ):
+        """
+        Args:
+            deprecate_in: Version in which code will be marked as deprecated.
+            remove_in: Version in which code is expected to be removed.
+            addendum: Additional messaging. Useful to indicate what should be used instead.
+            _current: Testing variable, shouldn't be used in practice.
+        """
+        from packaging import version
```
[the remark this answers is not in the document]
kenodegard: 🤷🏻‍♂️ because conda's version should always be PEP440? But nothing wrong with using `conda.models.version.normalized_version` instead especially given the speed difference:

```
python -m timeit -s 'from conda.models.version import normalized_version as parse' -- 'parse("22.3.0+93_kep98nrt_py0")'
1000000 loops, best of 5: 293 nsec per loop

python -m timeit -s 'from packaging.version import parse' -- 'parse("22.3.0+93_kep98nrt_py0")'
1000000 loops, best of 5: 8.84 usec per loop
```

Unfortunately, importing from `conda.models.version` results in a circular import, so perhaps this isn't worthwhile.

kenodegard:
the RTD failure looks to be due to a shallow clone of this repo, so a valid `conda.__version__` cannot be found

dholth:
Does conda depend on `packaging`?

kenodegard:
@dholth ah our testing environment has `packaging` but not the packaged conda, we have `pkg_resources` from `setuptools` available though 🤔 

I'll look more into the option to just use conda's version parser

dholth:
I might do a split('.') and int(first two elements) instead of a general purpose one...

dholth (dismissed):
How is this used outside of the test_deprecated module?

kenodegard:
@dholth I intend to either update all current pending/deprecations with a commit here or in a new PR once we agree on whether we like this implementation

dholth:
@kenodegard I wasn't sure how this would look in actual use. In the tests, you have to specify the "current conda version" as well as the deprecated-in and disabled-in versions. It would be helpful to see one real deprecation outside of the tests.

conda/__init__.py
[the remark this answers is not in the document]
kenodegard: Ended up implementing a dumb parser: https://github.com/conda/conda/pull/12125/commits/61dc68a73efc3e138038a81858c83eb2c0d82086

travishathaway (approved):

beeankha (dismissed):
This is a really useful addition, thank you @kenodegard !

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+        if self._version < deprecate_version:
+            self._category = PendingDeprecationWarning
+            message = f"{{name}} is pending deprecation and will be removed in {remove_in}."
+        elif self._version < remove_version:
+            self._category = DeprecationWarning
+            message = f"{{name}} is deprecated and will be removed in {remove_in}."
+        else:
+            self._category = None
+            message = f"{{name}} was slated for removal in {remove_in}."
+        self._message = lambda name: f"{message}{addendum}".format(name=name)
```
jezdez, line 195: I'm a little worried by the way this is written, the mix of f-strings and old-style `str.format()`ing etc. Can we just use this?

```suggestion

        if self._version < deprecate_version:
            category = PendingDeprecationWarning
            warning = f"is pending deprecation and will be removed in {remove_in}."
        elif self._version < remove_version:
            category = DeprecationWarning
            warning = f"is deprecated and will be removed in {remove_in}."
        else:
            category = None
            warning = f"was slated for removal in {remove_in}."

        def message(name):
            parts = [name, warning]
            if addendum:
                parts.append(addendum)
            return " ".join(parts)

        self._category = category
        self._message = lambda name: message(name)

```

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+        """Deprecation decorator for keyword arguments."""
+        # provide a default addendum if renaming and no addendum is provided
+        if rename and not addendum:
+            addendum = f"Use '{rename}' instead."
+
+        self = cls(deprecate_in=deprecate_in, remove_in=remove_in, addendum=addendum, stack=stack)
+        self._argument = argument
+        self._rename = rename
+
+        return self
```
jezdez, line 247:
```suggestion
        deprecation = cls(deprecate_in=deprecate_in, remove_in=remove_in, addendum=addendum, stack=stack)
        deprecation._argument = argument
        deprecation._rename = rename

        return deprecation
```

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+    def __call__(self, func: Callable) -> Callable:
+        """Deprecation decorator for functions, methods, & classes."""
+        # detect function name
+        fullname = f"{func.__module__}.{func.__qualname__}"
+        if self._argument:
+            fullname = f"{func.__module__}.{func.__qualname__}({self._argument})"
+
+        # alert developer that it's time to remove something
+        if not self._category:
+            raise RuntimeError(self._message(fullname))
```
jezdez, line 208: It might be useful to configure the type of exception that is used as a fallback.

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+    ) -> None:
+        """Deprecation function for a topic."""
+        self = cls(deprecate_in=deprecate_in, remove_in=remove_in, addendum=addendum, stack=stack)
+
+        # alert developer that it's time to remove something
+        if not self._category:
+            raise RuntimeError(self._message(topic))
+
+        # alert user that it's time to remove something
+        warnings.warn(self._message(topic), self._category, stacklevel=2 + self._stack)
```
jezdez, line 334:
```suggestion
        deprecation = cls(deprecate_in=deprecate_in, remove_in=remove_in, addendum=addendum, stack=stack)

        # alert developer that it's time to remove something
        if not deprecation._category:
            raise RuntimeError(deprecation._message(topic))

        # alert user that it's time to remove something
        warnings.warn(deprecation._message(topic), deprecation._category, stacklevel=2 + deprecation._stack)
```

jezdez: It's a little weird to me to see `self` used in a class method

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+
+    @classmethod
+    def _factory(cls, version: str) -> _deprecated:
+        return type("_deprecated", (cls,), {"_version": cls._parse_version(version)})
```
jezdez, line 338: Assuming we agree to allow other developers of conda related projects to make use of this, I think we should use a class to clearly instantiate the `decorated` helper, e.g. in a `conda_acme/__init__.py`:

```python
from conda import deprecations

__version__ = "23.1.0"

# A deprecation helper using the current version
deprecated = deprecations.Handler(__version__)

```

where `Handler` is a regular class with a number of instance methods for each use case (argument, constant etc)

jezdez (changes requested):
The name and location of this handler rub me the wrong way, since we'd shoehorn this into the `conda` namespace, making it "private" using the leading underscore *and* overwrite the base class with a dynamically created subclass.

I think that's far too exciting for this functionality, it should not get in the way and not risk the runtime issues of the decorated code. It's as if we're telling people importing conda (e.g. 3rd party developers) that they shouldn't use it, but it's still there, just for the conda code itself and nothing else.

I would suggest doing the following:

- move this into a new module `conda/deprecations.py` (analog to `conda/exceptions.py`)
- remove the leading underscore to enable other conda libraries to depend on it to deprecate things
- rewrite it using `wrapt` since [it's the most correct solution](https://hynek.me/articles/decorators/) for writing decorators and we need to be correct (`functools.wraps` isn't enough)

We _could_ release the result as an actual package on PyPI (but then why not use https://github.com/briancurtin/deprecation?)

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+    _version: tuple[int, ...]
+    _category: Warning
+    _message: Callable[str, str]
+    _argument: str | None = None
+    _rename: str | None = None
+    _stack: int
+
+    @staticmethod
+    def _parse_version(version: str) -> tuple[int, ...]:
+        """A naive version parser to avoid circular imports. Do not use for other purposes."""
```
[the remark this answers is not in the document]
kenodegard: I'm all for using `packaging` instead but is this PR a good enough reason for adding another dependency?

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+
+        def try_int(string: str) -> int | str:
+            try:
+                return int(string)
+            except ValueError:
+                return string
```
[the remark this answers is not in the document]
kenodegard: I hate these relatively dumb helper functions (if only try-except could be natively written as a oneliner). I'm all for extracting it or rewriting it into something cleaner.

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+        # alert user that it's time to remove something
+        warnings.warn(self._message(topic), self._category, stacklevel=2 + self._stack)
+
+    @classmethod
+    def _factory(cls, version: str) -> _deprecated:
+        return type("_deprecated", (cls,), {"_version": cls._parse_version(version)})
+
+
+# initialize conda's deprecation decorator with the current version
+_deprecated = _deprecated._factory(__version__)
```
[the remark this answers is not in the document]
kenodegard: I thought so too and struggled with this for a while, if we switch to using `packaging` I can make this quirk magically disappear 🪄

conda/__init__.py
kenodegard: So a new `DeprecationError` derived from `RuntimeError`?

conda/__init__.py
kenodegard: gotcha, I like to do this as it makes it very clear (to me) that it's an instance of the `cls` as opposed to being some other object 🤷🏼‍♂️ but I can see how it may be perceived as confusing

conda/__init__.py
kenodegard: This was the intended use case I had envisioned for the `_factory` method, renaming it to `Handler` confuses me.

Is the intention to push `__init__` and `__call__` down a level? So instead of:

```
@deprecated(...)
def foo(...):
    ...
```

We do:

```
@deprecated.function(...)
def foo(...):
    ...
```

conda/__init__.py
kenodegard: This too was an area I wasn't too happy with, these are some good ideas, what about this:

```suggestion
        if self._version < deprecate_version:
            category = PendingDeprecationWarning
            warning = f"is pending deprecation and will be removed in {remove_in}."
        elif self._version < remove_version:
            category = DeprecationWarning
            warning = f"is deprecated and will be removed in {remove_in}."
        else:
            category = None
            warning = f"was slated for removal in {remove_in}."

        self._category = category
        self._message = lambda name: f"{name} {warning} {addendum}".strip()
```

conda/__init__.py
```diff
@@ -148,3 +139,209 @@ def _default(self, obj):
...
+    @staticmethod
+    def _get_module(stack: int) -> tuple[ModuleType, str]:
+        import inspect  # expensive
+
+        try:
+            frame = inspect.stack()[2 + stack]
+            module = inspect.getmodule(frame[0])
+            return (module, module.__name__)
+        except (IndexError, AttributeError):
+            raise RuntimeError("unable to determine the calling module") from None
```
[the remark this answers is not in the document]
kenodegard: I'm explicitly catching any issues with the user-provided `stack` value used in `inspect.stack()[2 + stack]` since there is no guarantee that the user provided a reasonable (or even positive) value (e.g., `inf` or `-inf`).

kenodegard:
> The name and location of this handler rub me the wrong way, since we'd shoehorn this into the conda namespace, making it "private" using the leading underscore and overwrite the base class with a dynamically created subclass.

Yup I was struggling with this. I originally considered implementing it in `conda.deprecated` (back when the implementation was super small and simple), I agree that the scope has outgrown its location (I'm unsure if I'm happy with how much this tries to do).

> I think that's far too exciting for this functionality, it should not get in the way and not risk getting runtime issues of the decorated code. It's as if we're telling people importing conda (e.g. 3rd party developers) that they shouldn't use it, but it's still there, just for the conda code itself and nothing else.

The intention was always for conda and friends to be able to rely on this.

> - move this into a new module `conda/deprecations.py` (analog to `conda/exceptions.py`)
remove the leading underscore to enable other conda libraries to depend on it to deprecate things

👍🏻 

> - rewrite it using `wrapt` since [it's the most correct solution](https://hynek.me/articles/decorators/) for writing decorators and we need to be correct (`functools.wraps` isn't enough)

Soooooo, yes, however, I'm not sure I want to introduce `wrapt` if this is our only purpose for it. These are functions we intend to deprecate anyhow, if users need to jump through more hoops to learn about a deprecated function 🤷🏼‍♂️. `functools.wrap` is good enough IMO for this relatively small impact helper function.

> - We could release the result as an actual package on PyPI (but then why not use https://github.com/briancurtin/deprecation?)

We could if we think it's a meaningful addition to the Python community. 🤷🏼‍♂️

I did question using an existing tool when I initially started on this, but we collectively agreed that this was a core functionality. The caveat at the time being we wanted a "tiny function" and this is no longer "tiny."

dholth:
@kenodegard can we use this in conda-build?

conda/__init__.py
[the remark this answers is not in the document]
jezdez: I think it is, we should start relying on standard Python packaging tooling for things that relate to conda's status as a tool written in Python.

conda/__init__.py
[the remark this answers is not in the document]
jezdez: Oh, good point!

conda/__init__.py
jezdez: Works for me!

conda/__init__.py
jezdez: Yeah, or even a parameter. But an own exception class would suffice as well.

conda/__init__.py
jezdez: Yeah, I understand, looking how this could be confusing during debugging etc mostly.

conda/__init__.py
jezdez: My concerns is that the class does too much for its own sake, wrapping various functionality into one class (a generic factory, a few specific factories per decorated type, a general helper to retain state, a decorator).  The "handler" name was a suggestion to encourage being clear about it being a complex class and decoupling some of its parts.

Other than that, it's also trying to be "private" (lots of underscores) where [we know](https://www.hyrumslaw.com/) and want to encourage that it'll be used long-term, so why even try to hide the implementation and not optimize for ease of understanding?

My guess is that you're trying to make this code as small as possible, but it's impacting the readability and complexity of the code without an upside.

jezdez:
@kenodegard I think we should try to implement this in a separate module at least and see if it feels like a useful addition once done.

I'm pretty sure the conda ecosystem will follow whatever deprecation policy and implementation conda uses (over time), so it's important we make it reusable and easy to maintain.

conda/__init__.py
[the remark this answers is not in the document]
kenodegard: this is no longer needed since we use `packaging`

conda/__init__.py
kenodegard: all good points, I def wanted to make this as small as possible while still covering our use cases (conflicting objectives...)

conda/deprecations.py
```diff
@@ -0,0 +1,266 @@
...
+            if name == constant:
+                warnings.warn(message, category, stacklevel=2 + stack)
+                return value
+
+            if super_getattr:
+                return super_getattr(name)
+
+            raise AttributeError(f"module '{fullname}' has no attribute '{name}'")
+
+        module.__getattr__ = __getattr__
```
jezdez, line 192: 🔥

news/12125-added-conda._deprecated
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Added new `conda._deprecated` decorator and `conda._deprecated.module` functions for easier/standardized deprecation. (#12125)
```
jezdez, line 3: Do we need to publish this as `conda._deprecated`? I assumed we'd announce the existence of `conda.deprecations` instead, so we don't have to worry what happens in `conda.__init__` long-term?
```suggestion
* Added a new `conda.deprecations` module for easier/standardized deprecation, including a `deprecated` decorator and functions to mark modules for deprecation.  (#12125)
```

jezdez (approved):
I think that looks awesome! Two small nits:

- Shouldn't we document the decorator somewhere for the conda community maintainers to find (unless you want to wait for it to settle)?
- the import from `conda.deprecations` in the various packages shouldn't be "private". I think it's not providing enough information to 3rd parties why it has a leading underline. And we know that 3rd parties still import "private" variables from allover the place, if they feel like it. `from .deprecations import deprecated` is fine.

news/12125-added-conda._deprecated
kenodegard: oops, this was just an oversight on my part

kenodegard:
@jezdez My reasoning for importing as `_deprecated` was only to indicate that _users_ shouldn't be using it and to keep the namespace "cleaner." Neither is a strong enough reason, so I'll remove the aliases.

Yes, we want to document this. I've added a followup task in https://github.com/conda/conda/issues/12276 to add documention.

jezdez (approved):

beeankha (approved):

Issue #12018, opened 2022-10-26
Add auth handler plugin hook


travishathaway:
### What's the idea?

In order to allow conda to handle more means of authentication, a new plugin hook will be created. Please see the following pull request for more detailed information:

- https://github.com/conda/conda/pull/12911

### More background information

- [Conda Fetch Plugin Case Study: HTTP Basic Authentication](https://hackmd.io/hEfgjgMsT-WSQYVRc_9N2A?view) HackMD Document
- https://github.com/conda/conda/issues/11696
- https://github.com/conda/conda/pull/11971

dholth:
We need a good test harness that spins up proxies with various authentication methods, and any other flavor of repository we can think of.

travishathaway:
> 

This is work that should be included in a separate issue.

dholth:
Also, it may want to accept a Channel object instead of a channel string.

dholth:
e.g.

1. Conda opens 5 parallel requests to protected-repo
2. They all fail, but they pause, blocking conda
3. The session manager opens a user's web browser exactly once to do an OIDC flow (log in)
4. They all try again with the new login cookie
5. Conda finishes the transfer
6. If they tried again (step 4) but the new login cookie did not work, do not repeat step 3.

PR #12139, opened 2022-12-01
Channel aware session manager.

travishathaway:
This pull request implements everything described in the following issue:

- https://github.com/conda/conda/issues/12018

Please see that issue for full background details.


conda/gateways/connection/session.py
```diff
@@ -42,6 +46,33 @@ def close(self):
         raise NotImplementedError()
 
 
+@lru_cache(None)
+def get_channel_name_from_url(url: str) -> str | None:
```
dholth, line 50: I notice this is different than https://github.com/conda/conda/blob/main/conda/models/channel.py#L34 . Am unfamiliar with MultiChannel.

travishathaway: Probably want to use this instead:

https://github.com/conda/conda/blob/bdacdbb6079cc877e52a322d511c666aa8e89276/conda/models/channel.py#L451

jezdez:
Closing in favor of #12212

PR #12142, opened 2022-12-02
Fixes slow running test in the `tests/cli/test_main_notices.py` file

### LINKED ISSUE #12092: Slow `test_notices_appear_once_when_running_decorated_commands` test
### What happened?

Seems like the [test_notices_appear_once_when_running_decorated_commands](https://github.com/conda/conda/blob/9d51b702b41f47b6b53686a0ba2d013218dd5c71/tests/cli/test_main_notices.py#L236-L278) test is slow, e.g. [in this run with ~217 seconds](https://github.com/conda/conda/actions/runs/3448472489/jobs/5755526559#step:7:514). Wondering if the mock isn't working and it's timing out or something.

/cc @travishathaway

travishathaway:
For this test, it's not actually necessary for us to even run the install command. We just want to check whether the notices are running as we expect them to.

So, instead of relying on the fixture that runs `conda-build` under the covers, we've removed it and instead just mock out the install function.

Fix #12092.

### DIFF
--- a/tests/cli/test_main_notices.py
+++ b/tests/cli/test_main_notices.py
@@ -244,39 +244,39 @@ def test_notices_appear_once_when_running_decorated_commands(tmpdir, env_one, no
     To ensure this test runs appropriately, we rely on using a pass-thru mock
     of the `conda.notices.fetch.get_notice_responses` function. If this function
     was called and called correctly we can assume everything is working well.
+
+    This test intentionally does not make any external network calls and never should.
     """
     offset_cache_file_mtime(NOTICES_DECORATOR_DISPLAY_INTERVAL + 100)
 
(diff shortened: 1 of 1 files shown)


jezdez (approved):
Love it, thank you for the quick fix!

jezdez:
Oh wait, seems like something is failing with `ValueError: not enough values to unpack (expected at least 1, got 0)`

jezdez (changes requested):
Seems like a test is failing now with a ValueError

jezdez:
@travishathaway I don't think we even need to test for the use of "local" name and the url in the failing test tbh

jezdez (approved):

PR #12148, opened 2022-12-03
Fix missing __unix virtual package in 22.11

mbargull:
### Description

While publishing the latest `conda` release for conda-forge we encountered solver issues when the CI used `conda=22.11.0`. Particularly, the issues were seen when building Python 3.11 variants.
During those builds solving for the test environment failed since `responses` is a test requirement in the `conda-feedstock`. `responses` requires `pysocks` has only been recently (before a Python 3.11 variant of it would have been built) converted to a `noarch: python` package that depends on `__unix`/`__win` in https://github.com/conda-forge/pysocks-feedstock/pull/20 .

Turns out, `conda=22.11.0` does not create virtual packages for `__unix` and `__win`.

ref: https://github.com/conda-forge/admin-requests/pull/523
### Checklist - did you ...

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -145,12 +145,12 @@ def _supplement_index_with_cache(index):
             index[pcrec] = pcrec
 
 
-def _make_virtual_package(name, version=None, build_string='0'):
+def _make_virtual_package(name, version=None, build_string=None):
     return PackageRecord(
             package_type=PackageType.VIRTUAL_SYSTEM,
             name=name,
             version=version or '0',
-            build_string=build_string,
+            build_string=build_string or '0',
             channel='@',
             subdir=context.subdir,
             md5="12345678901234567890123456789012",
@@ -184,9 +184,8 @@ def _supplement_index_with_system(index):
             )
         registered_names.append(package.name)
 
-        rec = _make_virtual_package(f"__{package.name}", package.version)
-
-        if package.version is not None:
+        if package.name is not None:
+            rec = _make_virtual_package(f"__{package.name}", package.version, package.build)
             index[rec] = rec
 
 
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -174,6 +174,8 @@ def _supplement_index_with_system(index):
     registered_names = []
     packages = context.plugin_manager.get_hook_results("virtual_packages")
     for package in packages:
+        if package.name is None:
+            continue
         if package.name in registered_names:
             raise PluginError(
                 "Conflicting virtual package entries found for the "
@@ -184,9 +186,8 @@ def _supplement_index_with_system(index):
             )
         registered_names.append(package.name)
 
-        if package.name is not None:
-            rec = _make_virtual_package(f"__{package.name}", package.version, package.build)
-            index[rec] = rec
+        rec = _make_virtual_package(f"__{package.name}", package.version, package.build)
+        index[rec] = rec
 
 
 def get_archspec_name():
--- a/conda/plugins/virtual_packages/archspec.py
+++ b/conda/plugins/virtual_packages/archspec.py
@@ -7,4 +7,6 @@ from .. import hookimpl, CondaVirtualPackage
 def conda_virtual_packages():
     from ...core.index import get_archspec_name
 
-    yield CondaVirtualPackage("archspec", get_archspec_name())
+    archspec_name = get_archspec_name()
+    if archspec_name is not None:
+        yield CondaVirtualPackage("archspec", "1", archspec_name)
--- a/conda/plugins/virtual_packages/archspec.py
+++ b/conda/plugins/virtual_packages/archspec.py
@@ -8,5 +8,5 @@ def conda_virtual_packages():
     from ...core.index import get_archspec_name
 
     archspec_name = get_archspec_name()
-    if archspec_name is not None:
+    if archspec_name:
         yield CondaVirtualPackage("archspec", "1", archspec_name)
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -84,4 +84,6 @@ def cached_cuda_version():
 
 @hookimpl
 def conda_virtual_packages():
-    yield CondaVirtualPackage("cuda", cached_cuda_version())
+    cuda_version = cached_cuda_version()
+    if cuda_version is not None:
+        yield CondaVirtualPackage("cuda", cuda_version, None)
--- a/news/12148-restore-default-virtual-packages
+++ b/news/12148-restore-default-virtual-packages
@@ -1,0 +1,21 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Restore default virtual package specs as in 22.9.0 (#12148)
+  - re-add `__unix`/`__win` packages
+  - restore `__archspec` version/build string composition
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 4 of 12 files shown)


kenodegard:
Ah, I see you just found the spot, why not just leave everything as is and just remove the `if package.version is not None:`? See https://github.com/conda/conda/commit/0e385a59916c178ded358b0df7373cea3ecf16d1

mbargull:
> Ah, I see you just found the spot, why not just leave everything as is and just remove the `if package.version is not None:`? See [0e385a5](https://github.com/conda/conda/commit/0e385a59916c178ded358b0df7373cea3ecf16d1)

Because I assumed this was added for a reason/design decision and I rather wanted to prepare this for a quick patch release if needed. Also, I changed this to `package.name is not None` since that check makes sense anyway.

Otherwise, the `archspec` virtual package was changed without any notice of it in the changelog, so this restores it to the previous behavior. Overall, this should now create the specs for all virtual packages as before. (Okay, at least I guess, because I didn't check if the version composition for glibc/linux/osx/cuda changed too.)

kenodegard:
Yeah you're right, it looks like the build_string of the virtual packages got lost in the refactor

conda/core/index.py
```diff
@@ -145,12 +145,12 @@ def _supplement_index_with_cache(index):
...
 
 
-def _make_virtual_package(name, version=None, build_string='0'):
+def _make_virtual_package(name, version=None, build_string=None):
     return PackageRecord(
             package_type=PackageType.VIRTUAL_SYSTEM,
             name=name,
             version=version or '0',
-            build_string=build_string,
+            build_string=(build_string if build_string is not None else '0'),
```
kenodegard, line 153:
```suggestion
            build_string=build_string or '0',
```

news/12148-restore-default-virtual-packages
```diff
@@ -0,0 +1,21 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Restore default virtual package specs as in 22.9.0
```
kenodegard, line 7:
```suggestion
* Restore default virtual package specs as in 22.9.0 (#12148)
```

kenodegard (changes requested):

conda/core/index.py
mbargull: Oh, I misremembered that these are allowed to be empty.
Looking at the code again, I think you are right, and they should always be non-empty.

mbargull:
Thanks for the review!
(I amended the previous commits since the changes were small and I'd rather not clutter the history.)

kenodegard (dismissed):

mbargull:
Sorry, added another small commit since I noticed it didn't make sense to add `None` to `registered_names` when no package is created for it anyway..

kenodegard (approved):
Good catch

beckermr:
Wow. I'm surprised the tests didn't catch this. Good work @mbargull!

jezdez:
What a surprising small unwanted change! Epic sleuthing @mbargull @kenodegard @beckermr!!

conda/core/index.py
```diff
@@ -145,12 +145,12 @@ def _supplement_index_with_cache(index):
...
 
 
-def _make_virtual_package(name, version=None, build_string='0'):
+def _make_virtual_package(name, version=None, build_string=None):
     return PackageRecord(
             package_type=PackageType.VIRTUAL_SYSTEM,
             name=name,
             version=version or '0',
-            build_string=build_string,
+            build_string=build_string or '0',
```
dholth, line 153: Is it important to have an empty build string

jezdez: FWIW this was the previous behavior: https://github.com/conda/conda/blob/22.9.0/conda/core/index.py#L152-L163

conda/core/index.py
```diff
@@ -174,6 +174,8 @@ def _supplement_index_with_system(index):
     registered_names = []
     packages = context.plugin_manager.get_hook_results("virtual_packages")
     for package in packages:
+        if package.name is None:
```
jezdez, line 177: This is silencing a buggy virtual package instance, that doesn't have a name, I'd rather not swallow that here. We could raise a `PluginError` instead of course, which might be useful to help user identify where the problem is?

conda/plugins/virtual_packages/archspec.py
```diff
@@ -7,4 +7,6 @@
 def conda_virtual_packages():
     from ...core.index import get_archspec_name
 
-    yield CondaVirtualPackage("archspec", get_archspec_name())
+    archspec_name = get_archspec_name()
+    if archspec_name is not None:
+        yield CondaVirtualPackage("archspec", "1", archspec_name)
```
jezdez, line 12: Ooooh, that's an interesting one, not sure how this was lost, compared to https://github.com/conda/conda/blob/22.9.0/conda/core/index.py#L217-L219

conda/plugins/virtual_packages/cuda.py
```diff
@@ -84,4 +84,6 @@ def cached_cuda_version():
 
 @hookimpl
 def conda_virtual_packages():
-    yield CondaVirtualPackage("cuda", cached_cuda_version())
+    cuda_version = cached_cuda_version()
+    if cuda_version is not None:
+        yield CondaVirtualPackage("cuda", cuda_version, None)
```
jezdez, line 89: It's surprising that the virtual plugin API here doesn't take `None` into account as a valid value, which would ignore it during solving.

jezdez (approved):
I have more questions now about how this was not covered by the test suite, but this is obviously more aligned in how it worked until 22.9.0.

conda/core/index.py
beckermr: yes let's not change stuff because it is not obvious that we need it. too much history in this code base.

mbargull: We'd probably break stuff with empty strings. There is code similar to this in `resolve` for example:
```py
prec = PackageRecord(name="a", version="1", build="", build_number=0)
spec = MatchSpec(f"{prec.name} {prec.version} {prec.build}")
# this results in:
str(spec) == "a==1"
spec.get("build") is None
# i.e., with that MatchSpec from string above, we loose the build string completely
```

conda/core/index.py
mbargull: Let's decide on how to handle such things when considering https://github.com/conda/conda/issues/12149 .

conda/plugins/virtual_packages/cuda.py
mbargull: I don't follow...?

conda/plugins/virtual_packages/archspec.py
mbargull, line 12:
```suggestion
    if archspec_name:
        yield CondaVirtualPackage("archspec", "1", archspec_name)
```

conda/core/index.py
jezdez: Sounds good.

conda/plugins/virtual_packages/cuda.py
jezdez: It's unfortunate that we can't use `None` indicating the unavailability of the virtual package and have to check for it before yielding the virtual package object. I know it's a backward compatibility issue, but still. Nothing to do there FWIW.

jezdez (approved):

conda/plugins/virtual_packages/cuda.py
mbargull: The API for plugins is still fresh, so I don't know if it's set in stone already to worry too much about backward compatibility. Then again, I also haven't thought about what the best way to communicate absence is; possibly a more structured way to communicate it could be need in future (but IDK).
But well, things to worry about for a release after `22.11.1` ;)

PR #12164, opened 2022-12-09
Replace tlz concat with itertools

dholth:
### Description

toolz concat forwards to itertools; we can forward ourselves instead.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -9,10 +9,7 @@ from os.path import isfile, join
 import sys
 from textwrap import wrap
 
-try:
-    from tlz.itertoolz import concat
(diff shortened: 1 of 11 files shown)


dholth:
Some usages not matching the literal `from tlz.itertoolz import concat` string remain.

kenodegard (changes requested):
Can we use `itertools.chain.from_iterable` instead of creating a new alias?

dholth:
> Can we use `itertools.chain.from_iterable` instead of creating a new alias?

I'm trying to make this PR easy, since we wound up deleting the same work in a previous unmerged PR. I can check if refactor -> inline is available...

PR #12174, opened 2022-12-12
put test skip runs in own concurrency group

dholth:
### Description

Don't cancel tests because a non-testing commit is added.

If this is in effect, will real tests (possibly in a previous commit in the same pull request) still gate the merge?

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests-skip.yml
+++ b/.github/workflows/tests-skip.yml
@@ -35,7 +35,7 @@ concurrency:
   # concurrency group while a previous build is running it will be canceled.
   # Repeated pushes to a PR will cancel all previous builds, while multiple
   # merges to main will not cancel.
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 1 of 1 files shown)


kenodegard:
ah!

> If this is in effect, will real tests (possibly in a previous commit in the same pull request) still gate the merge?

No

What I want the path filtering to distinguish between is if there are any code changes in this PR (not just in the last commit), this seems to be what we are missing.

kenodegard:
per the GH [docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#git-diff-comparisons):

> GitHub generates the list of changed files using two-dot diffs for pushes and three-dot diffs for pull requests:
> - **Pull requests:** Three-dot diffs are a comparison between the most recent version of the topic branch and the commit where the topic branch was last synced with the base branch.

So all commits pushed to a PR are supposed to be compared with the paths/paths-ignore lists to determine which workflow to run.

Based on what we've recently experienced, I don't think this is what is occurring.

jezdez:
Closing since #12180 made this moot.

PR #12179, opened 2022-12-14
different mechanism for testing get_env_var

dholth:
### Description

Get environment variables with one command instead of three. Might avoid a race condition?

### Checklist - did you ...


tests/test_activate.py
```diff
@@ -2123,16 +2123,19 @@ def get_env_var(self, env_var, default=None):
...
         else:
-            self.sendline('echo get_var_start')
-            self.sendline(self.print_env_var % env_var)
-            self.sendline('echo get_var_end')
-            self.expect('get_var_start\n')
+            # strip redundant 'echo ' (how varied is print_env_var for this shell?)
+            assert " " in (self.print_env_var % "")
+            print_env_var = (self.print_env_var % env_var).split(" ", 1)[-1]
+            # send in one command, to try to avoid race conditions
+            self.sendline(r"echo get_var_start\\n" + print_env_var + r"\\nget_var_end")
```
jezdez, line 2130: The prefix and suffix strings, do they need to be raw strings or would this also work?

```suggestion
            self.sendline("echo get_var_start\n" + print_env_var + "\nget_var_end")
```

jezdez (commented):
Just a quick question

tests/test_activate.py
dholth: we are definitely trying to pass escape sequences to bash, which will then interpret them.

Issue #11890, opened 2022-09-27
Docs needed on `channel::package` syntax available in environment.yaml file


kathatherine:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Per comments on #7202, the documentation is lacking on the `channel::package` syntax available in environment.yaml files, in order to specify the channel from which a package should be downloaded during environment creation. Docs on this should be added to https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually, and probably referenced elsewhere in the Managing environments page.

### Why is this needed?

So users are aware this feature exists and can more easily specify package locations during environment creation.

jezdez:
Closed in #12213.

skandermoalla:
I couldn't find anywhere whether the channel would also apply to the dependencies of the package prefixed with the channel, or if it only applies to the package itself (and its dependencies would be searched in the other channels first).
Anyone can help?
@jezdez

PR #12213, opened 2023-01-04
Document channel::package syntax for environment.yml files

ReveStobinson:
### Description

Adds documentation requested in #11890.

Documented `channel::package` syntax for dependency specification in a `note:` in the **Sharing an Environment** section of `docs/source/user-guide/tasks/manage-environments.rst`. Also added `news` file, but it may not be needed for a change of this magnitude.

### Checklist - did you ...

### DIFF
--- a/docs/source/user-guide/tasks/manage-environments.rst
+++ b/docs/source/user-guide/tasks/manage-environments.rst
@@ -850,6 +850,8 @@ EXAMPLE: A more complex environment file:
        - Flask-Testing
 
 .. note::
+   **Using wildcards**
+
    Note the use of the wildcard * when defining the patch version
    number. Defining the version number by fixing the major and minor
    version numbers while allowing the patch version number to vary
@@ -857,7 +859,8 @@ EXAMPLE: A more complex environment file:
    to get any bug fixes whilst still maintaining consistency of
    software environment.
 
-.. note::
+   **Specifying channels outside of "channels"**
+
    You may occasionally want to specify which channel conda will
    use to install a specific package. To accomplish this, use the
    `channel::package` syntax in `dependencies:`, as demonstrated
--- a/docs/source/user-guide/tasks/manage-environments.rst
+++ b/docs/source/user-guide/tasks/manage-environments.rst
@@ -842,7 +842,7 @@ EXAMPLE: A more complex environment file:
    dependencies:
      - python=3.9
      - bokeh=2.4.2
-     - numpy=1.21.*
+     - conda-forge::numpy=1.21.*
      - nodejs=16.13.*
      - flask
      - pip
@@ -856,6 +856,15 @@ EXAMPLE: A more complex environment file:
    allows us to use our environment file to update our environment
    to get any bug fixes whilst still maintaining consistency of
    software environment.
+
+.. note::
+   You may occasionally want to specify which channel conda will 
+   use to install a specific package. To accomplish this, use the
+   `channel::package` syntax in `dependencies:`, as demonstrated
+   above with `conda-forge::numpy` (version numbers optional). The
+   specified channel does not need to be present in the `channels:`
+   list, which is useful if you want some—but not *all*—packages
+   installed from a community channel such as `conda-forge`.
 
 You can exclude the default channels by adding ``nodefaults``
 to the channels list.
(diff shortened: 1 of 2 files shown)


beeankha:
@conda-bot check

beeankha (dismissed):

chenghlee (dismissed):

travishathaway:
Hi @ReveStobinson,

Thank for you very much for this contribution. This additional example helps reiterate the fact that this is possible to do within `environmnet.yml` config files.

I made some slight changes to what you had proposed (I hope this is okay! 😅). I couldn't help but notice that this "admonition" ("note" box) appears right next to another one.  To help reduce visual clutter and to make it easier for the reader, I have combined the two into a single box and also added paragraph headers for each. I think the end product is slightly better.

Please checkout the preview on this pull request to let us know what you think. If we don't hear back soon, we will go ahead and merge this pull request.

Cheers ✌️,

--
Travis from the conda team 🐍

travishathaway (approved):

jezdez (approved):

James-Leslie:
Hi there,
What if a package needs to use multiple channels?

For example, the `datasets` package from huggingface would usually be installed with (taken from bottom of [this page](https://huggingface.co/docs/datasets/installation)): 
```bash
conda install -c huggingface -c conda-forge datasets
```

How would I specify this in my `.yml` file?

```yml
name: myenv
channels:
  - conda-forge
  - defaults
dependencies:
  - numpy
  - pandas
  - huggingface::transformers
  - huggingface::datasets
  - pytorch::pytorch
```

ReveStobinson:
@James-Leslie **TL;DR,** your `.yml` file looks correct.
> Hi there, What if a package needs to use multiple channels?
> 
> For example, the `datasets` package from huggingface would usually be installed with (taken from bottom of [this page](https://huggingface.co/docs/datasets/installation)):
> 
> ```shell
> conda install -c huggingface -c conda-forge datasets
>```

A single package _on its own_ won't ever need multiple channels to install, but its dependencies might, which is what I understand is the case in the example you gave from the Huggingface docs. In this case that was fairly obvious, and your `environment.yml` was indeed correct.

I _could_ forsee an issue in a slightly different (seemingly unlikely?) case where you might not know which of the two channels your packages comes from. As I see it, this would mostly affect highly specific packages only available on smaller channels.

### Example

Consider a fictional package called `frog_stuff`, which is maintained by one University professor who never checks his email. His install documentation is one line:

> Use conda to install `frog_stuff` with the following command: `conda install -c bioconda -c paulscherrerinstitute frog_stuff`. 

You might not know which channel `frog_stuff` is actually hosted on, so you don't know which channel to put in `channel` and which to put before `::frog_stuff`.  If `frog_stuff is only on one of these two channels, one combination will invariably error out.

Typically, you'd be able to figure out fairly easily which channel holds the dependencies, and which holds the "base" package (as you did with `huggingface::datasets`). In a less obvious case like `frog_stuff`, the context clues aren't there, so your best bet is probably either searching both channels for the package in question, or simple trial and error.

James-Leslie:
Hi @ReveStobinson (great username - I always enjoy swapping letters around like that 😂) thanks very much for that helpful response.

Would you be able to help me in defining the `.yml` file for including the `pytorch-cuda=11.7` package?

It has similar conda installation instructions:
```bash
conda install pytorch-cuda=11.7 -c pytorch -c nvidia
```

My understanding is that the `pytorch-cuda` package comes from the `pytorch` channel, but its dependencies all come from the `nvidia` channel.

So if I were to specify `pytorch::pytorch-cuda`, how does conda know where to get the dependencies that come from the `nvidia` channel? Should I add `nvidia` to the list of channels?

ReveStobinson:
@James-Leslie 
> So if I were to specify `pytorch::pytorch-cuda`, how does conda know where to get the dependencies that come from the `nvidia` channel? Should I add `nvidia` to the list of channels?

**Yes!** Sorry for the late reply, but this is exactly the use case where I discovered this undocumented syntax that prompted this PR. 

`pytorch-cuda` comes from the `pytorch` channel, but it will also install a ton of libcuda and cudnn-type stuff for you, and some of those dependencies (or possibly their most up-to-date versions) are hosted on the `nvidia` channel. So, you would indeed just specify `- nvidia` in your `channels:` list at the top of your `.yml` file.

PR #12239, opened 2023-01-11
Adds a new config parameter: `channel_settings`

travishathaway:
This pull request was created in response to the discussion on a previous pull request:

- https://github.com/conda/conda/pull/12033

It is also related to this ticket:

closes: https://github.com/conda/conda/issues/11825

I concluded that changing `channels` they way I was proposing in the above pull request was a little too risky. Not only was there a significant change of a potentially brittle configuration handling module (`conda.core.configuration`), there was also the chance of breaking older versions of conda.

You will notice this pull request is considerably smaller as it introduces a separate `channel_settings` configuration parameter and does not involve changing anything in the configuration handling module. Older versions of conda will simply ignore this new parameter thus reducing the chances of affect older versions.


jezdez:
@travishathaway Just to make sure I understand correctly, you're looking to add this without also committing the use case, right?

tests/base/test_context.py
```diff
@@ -42,7 +42,8 @@
 class ContextCustomRcTests(TestCase):
 
     def setUp(self):
-        string = dals("""
+        string = dals(
+            """
```
kenodegard, line 46: Since this got formatted, can we indent the text block below to align with these open quotes?

news/12239-add-channel-settings-parameter
```diff
@@ -0,0 +1,3 @@
+### Enhancements
+
+* Adds a new "channel_settings" configuration parameter that will be used to override arbitrary settings on per channel basis
```
kenodegard, line 3:
```suggestion
* Adds a new "channel_settings" configuration parameter that will be used to override arbitrary settings on per channel basis. (#12239)
```

kenodegard (changes requested):
Minor corrections, otherwise looks good

tests/base/test_context.py
dholth: Can we remove dals and just use """ with correctly-formatted inner contents?

beeankha (approved):
Aside from the changes suggested by @kenodegard and others, this looks good to me! 👍

tests/base/test_context.py
travishathaway: @dholth, to be honest, I've always thought the `dals` function was a little silly, so I wouldn't mind doing it, but I believe it's better to just stick with existing conventions 🤷

kenodegard (dismissed):

jezdez (approved):
Thanks @travishathaway, when you get closer to using this for specific features (e.g. fetch plugin), please mention this here so we can keep track of it historically.

conda/base/context.py
```diff
@@ -963,6 +966,7 @@ def category_map(self):
             "Channel Configuration": (
                 "channels",
                 "channel_alias",
+                "channel_settings",
```
mbargull, line 969: Can we add this to `"Hidden and Undocumented"` for now (maybe with a `# Once this is used, move it to "Channel Configuration".` or similar comment)?

conda/base/context.py
```diff
@@ -1201,6 +1205,13 @@ def description_map(self):
                 The list of conda channels to include for relevant operations.
                 """
             ),
+            channel_settings=dals(
+                """
+                A list of mappings that allows overriding certain settings for a single channel.
+                Each list item should include at least the "channel" key and the setting you would
```
mbargull, line 1211: Since this PR can be seen as a first "reserve the config key" step, I'm okay if the following is done in a subsequent PR (but before `channel_settings` sees usage):
Some checks should be done while reading the configuration:
  - If `channel` key is not present, either
    1. error or
    2. warn and discard

    the entry (IDK/haven't thought about what's preferred here).
- In case of duplicate entries with the same `channel` key in the same file,
  1. error/warn, and
  2. if only warn, decide if first or last entry should be chosen.
- In case of duplicate entries with the same `channel` key in *different* files, decide if we want to
  1. override settings in higher priority config files, or
  2. merge the channel's setting keys.

  (the first option is easier to reason about).

mbargull (approved):
LGTM!
Left only minor suggestions that can mostly be done in a later PR.
(Sorry, didn't get to review this earlier.)

conda/base/context.py
travishathaway: Thanks for the feedback @mbargull.

I will open an issue with this information in it to make sure it is addressed before we begin using this feature.

beeankha (approved):

PR #12241, opened 2023-01-12
Repodata state format per draft CEP

dholth:
### Description

Standardized field names, plus mtime checks to ignore `X.state.json` if `X.json` was changed.

Although the CEP is not finished, we want to merge the mtime checks to make sure an outdated `state.json` is ignored. `*.state.json` stores fields previously in-lined into cached `repodata.json`, files named with hexadecimal `<prefix>.json` in places like `~/miniconda3/pkgs/cache/`

https://github.com/conda-incubator/ceps/pull/46

### Checklist - did you ...

### DIFF
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -13,22 +13,22 @@ from contextlib import closing
 from errno import EACCES, ENODEV, EPERM, EROFS
 from functools import partial
 from io import open as io_open
+from itertools import chain, islice
 from logging import getLogger
 from mmap import ACCESS_READ, mmap
 from os.path import dirname, exists, isdir, join, splitext
+from pathlib import Path
 from time import time
 
 from genericpath import getmtime, isfile
 
-from itertools import islice, chain
-
 from conda.common.iterators import groupby_to_dict as groupby
 from conda.gateways.repodata import (
-    CondaRepoInterface,
-    RepoInterface,
-    RepodataIsEmpty,
-    Response304ContentUnchanged,
     CacheJsonState,
+    CondaRepoInterface,
+    RepodataIsEmpty,
+    RepoInterface,
+    Response304ContentUnchanged,
 )
 
 from .. import CondaError
@@ -250,8 +250,8 @@ class SubdirData(metaclass=SubdirDataType):
         return CacheJsonState(self.cache_path_json, self.cache_path_state, self.repodata_fn).load()
 
     def _save_state(self, state: CacheJsonState):
-        assert state.cache_path_json == self.cache_path_json
-        assert state.cache_path_state == self.cache_path_state
+        assert Path(state.cache_path_json) == Path(self.cache_path_json)
+        assert Path(state.cache_path_state) == Path(self.cache_path_state)
         assert state.repodata_fn == self.repodata_fn
         return state.save()
 
@@ -642,7 +642,7 @@ def fetch_repodata_remote_request(url, etag, mod_stamp, repodata_fn=REPODATA_FN)
     subdir = SubdirData(Channel(url), repodata_fn=repodata_fn)
 
     try:
-        cache_state = CacheJsonState(None, None, repodata_fn)
+        cache_state = subdir._load_state()
         cache_state.etag = etag
         cache_state.mod = mod_stamp
         raw_repodata_str = subdir._repo.repodata(cache_state)
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -80,6 +80,58 @@ class SubdirDataType(type):
         return subdir_data_instance
 
 
+class CacheJsonState(dict):
+    """
+    Load/save `.state.json` that accompanies cached `repodata.json`
+    """
+
+    def __init__(self, cache_path_json, cache_path_state, repodata_fn):
+        self.cache_path_json = pathlib.Path(cache_path_json)
+        self.cache_path_state = pathlib.Path(cache_path_state)
+        self.repodata_fn = repodata_fn
+
+    def load(self) -> dict:
+        """
+        Cache headers and additional data needed to keep track of the cache are
+        stored separately, instead of the previous "added to repodata.json"
+        arrangement.
+        """
+        try:
+            state_path = self.cache_path_state
+            log.debug("Load %s cache from %s", self.repodata_fn, state_path)
+            with state_path.open("r") as s:
+                state = json.load(s)
+            # json and state files should match
+            json_stat = self.cache_path_json.stat()
+            if not (
+                state.get("mtime_ns") == json_stat.st_mtime_ns
+                and state.get("size") == json_stat.st_size
+            ):
+                # clear mod, etag, cache_control to encourage re-download
+                state.update({"etag": "", "mod": "", "cache_control": ""})
+            for alias in "_etag", "_mod", "_cache_control":
+                if alias[1:] in state:
+                    state[alias] = state.pop(alias[1:])
+            self.update(state)
+            return self
+        except (json.JSONDecodeError, OSError):
+            log.debug("Could not load state", exc_info=True)
+            self.clear()
+            return self
+
+    def save(self, state: dict):
+        """
+        Must be called after writing cache_path_json, as its mtime is included in .state.json
+        """
+        json_stat = self.cache_path_json.stat()
+        serialized = {}
+        serialized.update({"mtime_ns": json_stat.st_mtime_ns, "size": json_stat.st_size})
+        for alias in "_etag", "_mod", "_cache_control":
+            if alias in state:
+                serialized[alias[1:]] = state[alias]
+        return pathlib.Path(self.cache_path_state).write_text(json.dumps(serialized, indent=True))
+
+
 class SubdirData(metaclass=SubdirDataType):
     _cache_ = {}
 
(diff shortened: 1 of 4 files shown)


dholth:
@wolfv

dholth:
I realized there is no backwards compat reason to keep state as a dict. Probably will go back and pass it around as an object instead, will make aliasing etag / mtime / ... much easier.

dholth:
It looks like there are some bugs with the existing `.state.json`, so we'll need to merge an improved version of this PR before the release.

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,88 @@ def conda_http_errors(url, repodata_fn):
...
+        """
+        Cache headers and additional data needed to keep track of the cache are
+        stored separately, instead of the previous "added to repodata.json"
+        arrangement.
+        """
+        try:
+            state_path = self.cache_path_state
+            log.debug("Load %s cache from %s", self.repodata_fn, state_path)
+            with state_path.open("r") as s:
+                state = json.load(s)
```
kenodegard, line 315:
```suggestion
            log.debug("Load %s cache from %s", self.repodata_fn, self.cache_path_state)
            state = json.loads(self.cache_path_state.read_text())
```

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,88 @@ def conda_http_errors(url, repodata_fn):
             e.response,
             caused_by=e,
         )
+
+
+class CacheJsonState(dict):
```
kenodegard, line 295: This implementation defines both `etag/mod/cache_control` and private aliases `_etag/_mod/_cache_control`, however, the `_` prefixed aliases are unused. I think it's ok to remove the private aliases.

tests/core/test_subdir_data.py
```diff
@@ -307,12 +309,45 @@ def test_metadata_cache_clearing(platform=OVERRIDE_PLATFORM):
...
+    mod = "last modified time"
+
+    state = CacheJsonState(cache_json, cache_state, "repodata.json")
+    state["_mod"] = mod
+    state["_cache_control"] = "cache control"
+    state["_etag"] = "etag"
+    state.save()
+
+    on_disk_format = json.loads(cache_state.read_text())
+    print("disk format", on_disk_format)
```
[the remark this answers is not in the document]
dholth: it's in a test, you'll never see it until you debug

conda/core/subdir_data.py
```diff
@@ -622,7 +597,40 @@ def cache_fn_url(url, repodata_fn=REPODATA_FN):
     return f"{md5.hexdigest()[:8]}.json"
 
 
+def read_mod_and_etag(path):
```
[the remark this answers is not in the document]
dholth: This is for getting the in-band mod and etag, which will no longer be generated, so this function will no longer be used by conda.

conda/gateways/repodata/__init__.py
dholth: This is probably wrong, since I should be using the new class?

conda/gateways/repodata/__init__.py
dholth: I removed them just now.

dholth: This class is currently a mess.

conda/core/subdir_data.py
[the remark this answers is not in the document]
kenodegard: I understand that, my question is, why did this function get moved from its original location on line 566/560 before the `get_cache_control_max_age` function? When deprecating stuff, we should keep the edits to a minimum.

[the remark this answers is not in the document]
dholth: Only because I wanted to put it next to the other deprecated function.

conda/gateways/repodata/__init__.py
kenodegard: this is the new state class?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,97 @@ def conda_http_errors(url, repodata_fn):
...
+                state.update({"etag": "", "mod": "", "cache_control": "", "size": 0})
+            for field in (
+                "etag",
+                "mod",
+                "cache_control",
+                "size",
+            ):
+                alias = field
+                if alias in state:
+                    self[field] = state[alias]
```
kenodegard, line 331: We should define a private class level `_fields` variable (e.g., `_fields = ("etag", "mod", "cache_control", "size")`).

```suggestion
            self.update({field: state[field] for field in self._fields if field in state})
```

kenodegard, line 331: We should define a private class level `_fields` variable (e.g., `_fields = ("etag", "mod", "cache_control", "size")`).

```suggestion
            self.update({field: state[field] for field in self._fields if field in state})
```

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,96 @@ def conda_http_errors(url, repodata_fn):
...
+        serialized = {"mtime_ns": json_stat.st_mtime_ns, "size": json_stat.st_size}
+        for field in (
+            "etag",
+            "mod",
+            "cache_control",
+            "size",
+        ):
+            alias = field
+            if field in self:
+                serialized[alias] = self[field]
```
kenodegard, line 351:
```suggestion
        serialized.update({field: self[field] for field in self._fields if field in self})
```

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,96 @@ def conda_http_errors(url, repodata_fn):
...
+    @property
+    def mod(self) -> str:
+        """
+        Last-Modified header or ""
+        """
+        return self.get("mod", "")
+
+    @mod.setter
+    def mod(self, value):
+        self["mod"] = value or ""
```
kenodegard, line 363: Per the way we serialize in `save` shouldn't these properties clear empty values?

```suggestion
        if value:
            self["mod"] = value
        elif "mod" in self:
            del self["mod"]
```

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,96 @@ def conda_http_errors(url, repodata_fn):
...
+    def load(self):
+        """
+        Cache headers and additional data needed to keep track of the cache are
+        stored separately, instead of the previous "added to repodata.json"
+        arrangement.
+        """
+        try:
+            state_path = self.cache_path_state
+            log.debug("Load %s cache from %s", self.repodata_fn, state_path)
+            state = json.loads(state_path.read_text())
```
jezdez, line 314: Should we handle exceptions in case the cache file is corrupt?

conda/core/subdir_data.py
```diff
@@ -240,26 +241,19 @@ def iter_records(self):
             self.load()
         return iter(self._package_records)
 
-    def _load_state(self) -> dict:
+    def _load_state(self):
```
jezdez, line 244:
```suggestion
    def _load_state(self) -> CacheJsonState:
```
Does that help in any way?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,96 @@ def conda_http_errors(url, repodata_fn):
             e.response,
             caused_by=e,
         )
+
+
+class CacheJsonState(dict):
```
jezdez, line 295: I feel like subclassing `dict` and not implementing the parent `__init__` signature seems wrong, could we instead use `collections.UserDict` for this to retain the ability to access a regular dict?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +290,96 @@ def conda_http_errors(url, repodata_fn):
...
+            ):
+                alias = field
+                if alias in state:
+                    self[field] = state[alias]
+        except (json.JSONDecodeError, OSError):
+            log.debug("Could not load state", exc_info=True)
+            self.clear()
+        return self
+
+    def save(self):
```
jezdez, line 337:
```suggestion
    def save(self) -> pathlib.Path:
```

tests/core/test_subdir_data.py
```diff
@@ -307,12 +310,47 @@ def test_metadata_cache_clearing(platform=OVERRIDE_PLATFORM):
...
+    underscore-prefixed field names. Assert state is only loaded if it matches
+    cached json.
+    """
+    cache_json = tmp_path / "cached.json"
+    cache_state = tmp_path / "cached.state.json"
+
+    cache_json.write_text("{}")
+
+    CacheJsonState(cache_json, cache_state, "repodata.json").save()
+
```
jezdez, line 325: Could we check for the returned path here?

conda/gateways/repodata/__init__.py
jezdez:
```suggestion
class RepodataState(dict):
```
Assuming we use this in the future, it might make sense to forgo the "cache" name and just call it what it is, the representation of the repodata state?

jezdez (changes requested):
This is coming together, nice!

But! There is still an ongoing CEP discussion, so it's a little premature to merge this presently, since we're about to do the 23.1.0 release (today) and the code in question doesn't benefit from a lot of churn if we rush and have to fix this.

I'm assigning it to the 23.1.1 release milestone to indicate that it's okay to ship this way before the next regular release.

tests/core/test_subdir_data.py
dholth: `write_text` returns the number of bytes written, not a Path

conda/gateways/repodata/__init__.py
dholth: On L332

dholth:
Needs a test to check that state added to `_mod` etc. keys by RepoInterface implementations, lands in the correct place in the .state.json files.

dholth:
Folding into the jlap branch.

PR #12252, opened 2023-01-19
Implement custom `unique` function

kenodegard:
### Description

Inlined a simplified version of toolz's `unique` function.

Xref https://github.com/conda/conda/issues/11718

### Checklist - did you ...

### DIFF
--- a/conda/common/iterators.py
+++ b/conda/common/iterators.py
@@ -1,9 +1,9 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
-
 """
 Replacements for parts of the toolz library.
 """
+from __future__ import annotations
 
 import itertools
 import collections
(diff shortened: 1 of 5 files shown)


conda/common/iterators.py
```diff
@@ -18,3 +19,20 @@ def groupby_to_dict(keyfunc, sequence):
     for key, group in itertools.groupby(sequence, keyfunc):
         result[key].extend(group)
     return dict(result)
+
+
+T = TypeVar("T")
+
+
+def unique(sequence: Sequence[T]) -> Generator[T, None, None]:
```
dholth, line 27: could be a `Sequence[Any]` or "hashable" (or just Sequence?) since we don't need every element in the sequence to be the same type

kenodegard: oops, you're totally right

dholth: actually try 'Iterable'

jezdez (approved):

beeankha (approved):

conda/common/iterators.py
dholth: This is okay but `Sequence` is about being able to index by integers, iterable is about for ... in

PR #12253, opened 2023-01-19
Replace `toolz.itertoolz.interleave` with `zip`

kenodegard:
### Description

No need to depend on `toolz`'s `interleave` when `zip` does the job.

Xref https://github.com/conda/conda/issues/11718

### Checklist - did you ...


dholth:
It's hard to tell if this does the same thing as the original. Might be necessary to split it out into a function and write a test comparing the two? https://github.com/conda/conda/blob/main/conda/_vendor/toolz/itertoolz.py#L236 is straightforward until it gets to the `except StopIteration` bit.

beeankha (approved):

PR #12262, opened 2023-01-20
Add .jlap, .json.zst extensions to paths to make them downloadable

### LINKED ISSUE #12261: conda should allow more extensions in conda.common.path / downloads
### Checklist

### What happened?

Conda should accept additional extensions in paths.py

A part of conda's http stack tries to parse URLs as channels, for binstar/anaconda.org authentication. 

https://github.com/conda/conda/blob/main/conda/gateways/connection/session.py#L129

This eventually ends up in https://github.com/conda/conda/blob/main/conda/common/path.py#L341

Then conda can't download files not having one of those extensions.

As a fix, add more extensions to this file.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

dholth:
### Description

Due to the way conda's network stack works today, we can only download certain extensions, otherwise these are stripped or taken to be part of a channel name. We anticipate needing to download `.json.zst` and `.jlap` in the near future; add these. Is otherwise harmless (may prevent users from including these extensions in `/some/channelname.json.zst/osx-64/` for example).

Also, reformat a test with black and pymodernize.

Fix #12261 

### Checklist - did you ...

### DIFF
--- a/conda/common/path.py
+++ b/conda/common/path.py
@@ -31,6 +31,10 @@ PATH_MATCH_REGEX = (
     r"|//"              # windows UNC path
 )
 
+# any other extension will be mangled by CondaSession.get() as it tries to find
+# channel names from URLs, through strip_pkg_extension()
+KNOWN_EXTENSIONS = (".conda", ".tar.bz2", ".json", ".jlap", ".json.zst")
+
 
 def is_path(value):
     if '://' in value:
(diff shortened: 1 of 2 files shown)


jezdez (changes requested):
Please remove the test formatting, this shouldn't be committed with the change in question, which is reasonable

dholth:
Interesting, I did have to cherry pick and force push to get the deformatted file past pre-commit.

jezdez (approved):
Thanks @dholth!

PR #12265, opened 2023-01-21
remove useless output from conda.csh

simonLeary42:
I found that the CSH environment setup script prints a bunch of useless output every time I want to activate it.
```
usage: conda [-h] [-V] command ...

conda is a tool for managing and deploying applications, environments and packages.

Options:

positional arguments:
  command
    clean        Remove unused packages and caches.
    compare      Compare packages between conda environments.
    config       Modify configuration values in .condarc. This is modeled after the git config command. Writes to the user .condarc
                 file (/home/simonleary_umass_edu/.condarc) by default. Use the --show-sources flag to display all identified
                 configuration locations on your computer.
    create       Create a new conda environment from a list of specified packages.
    info         Display information about current conda install.
    init         Initialize conda for shell interaction.
    install      Installs a list of packages into a specified conda environment.
    list         List installed packages in a conda environment.
    package      Low-level conda package utility. (EXPERIMENTAL)
    remove       Remove a list of packages from a specified conda environment.
    rename       Renames an existing environment.
    run          Run an executable in a conda environment.
    search       Search for packages and display associated information.The input is a MatchSpec, a query language for conda packages.
                 See examples below.
    uninstall    Alias for conda remove.
    update       Updates conda packages to the latest compatible version.
    upgrade      Alias for conda update.
    notices      Retrieves latest channel notifications.

optional arguments:
  -h, --help     Show this help message and exit.
  -V, --version  Show the conda version number and exit.

conda commands available from other packages:
  content-trust
  env
  pack
```

I just pipe stdout to `/dev/null` and the output goes away. If I'm not mistaken, this `if` statement checks if the return code == 0, and piping stdout shouldn't change anything. Error messages shouldn't be hidden because stderr is unchanged.

### DIFF
--- a/conda/shell/etc/profile.d/conda.csh
+++ b/conda/shell/etc/profile.d/conda.csh
@@ -1,30 +1,11 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 
-if (! $?_CONDA_EXE) then
-  set _CONDA_EXE="${PWD}/conda/shell/bin/conda"
-else
-  if ("$_CONDA_EXE" == "") then
-      set _CONDA_EXE="${PWD}/conda/shell/bin/conda"
-  endif
-endif
+set correct_conda_alias="source ${_CONDA_ROOT}/etc/profile.d/conda.csh"
 
-if ("`alias conda`" == "") then
-    if ($?_CONDA_EXE) then
-        # _CONDA_PFX is named so as not to cause confusion with CONDA_PREFIX
-        # If nested backticks were possible we wouldn't use any variables here.
-        set _CONDA_PFX=`dirname "${_CONDA_EXE}"`
-        set _CONDA_PFX=`dirname "${_CONDA_PFX}"`
-        alias conda source "${_CONDA_PFX}/etc/profile.d/conda.csh"
-        # And for good measure, get rid of it afterwards.
-        unset _CONDA_PFX
-    else
-        alias conda source "${PWD}/conda/shell/etc/profile.d/conda.csh"
-    endif
+if (!("`alias conda`" == "$correct_conda_alias")) then
     setenv CONDA_SHLVL 0
-    if (! $?prompt) then
-        set prompt=""
-    endif
+    alias conda $correct_conda_alias
 else
     switch ( "${1}" )
         case "activate":


dholth (dismissed):

simonleary-umass-edu:
I was mistaken, it does not check if exit code == 0, it checks if the returned text is an empty string. If stdout goes nowhere then it will always be an empty string and the if statement will always be true. I think the problem is that it's executing the output of the alias command. In bash you could fix it using `[[ ... ]]` instead of `( ... )` but not in csh.

simonleary-umass-edu:
On my local machine, it's not executing the output of the alias command.
```sh
#!/bin/csh
if ("`alias conda`" == "") then
        echo "no alias"
endif
alias conda "echo hi"
if ("`alias conda`" == "") then
        echo "no alias"
else
        echo "yes alias"
endif
```

```
no alias
yes alias
```

simonleary-umass-edu:
Conda's `conda.csh` works normally with the `csh` shell but it prints all that output with the `tcsh` shell. But my little test script works normally with both `csh` and `tcsh`.

simonleary-umass-edu:
Conda's `conda.csh` works normally with `tcsh` on my machine but it prints all that output only on the remote system I'm working on.

simonleary-umass-edu:
so the thing that makes the output is this:
```
            $_CONDA_EXE $argv[1-]
```
When you source the script with no arguments and conda alias already defined, it just does `conda`. That is the bug.

simonleary-umass-edu:
Because the `conda` command goes back to this script, there's no way for the script to tell the difference between "I am initializing conda but I already have a conda alias" and "I just called the conda command with no arguments".

simonleary-umass-edu:
My original PR will break conda completely, I think. Because it will always just initialize and never actually run the conda command.

travishathaway:
Hi @simonleary-umass-edu,

I appreciate your effort on this! Thanks for the information on how to better recreate this. I'll see if there's anything I can come up with on my end.

We definitely don't want to break conda 😅 , so keep us updated if you come up with a better technique.

Also, please feel welcome to switch this pull request into "draft mode" if you are still figuring out the best solution. This helps us stay organized.

simonleary-umass-edu:
This could be fixed by separating the two halves of `conda.csh` into `conda-init.csh` and `conda.csh`. But that seems like a big change and might make people uncomfortable.

The workaround for my system (using Lmod environment modules) is to make sure that `conda` is unalias'ed before `conda.csh` is run.

simonleary-umass-edu:
I changed a lot of stuff in my last commit.

I found that on my system, CONDA_EXE, _CONDA_ROOT, _CONDA_EXE, CONDA_PYTHON_EXE were hardcoded in the conda.csh script. I assumed that the installer script added these lines. The precense of _CONDA_ROOT (which is two directories up from CONDA_EXE) removes the need to do any `dirname` stuff.

I also found that these lines:
```sh
    if (! $?prompt) then
        set prompt=""
    endif
```
Do nothing because they set a local variable right before the script ends.

The `conda.sh` script doesn't try and come up with a new `CONDA_EXE` -> `"${PWD}/conda/shell/etc/profile.d/conda.csh"`, so I don't see why `conda.csh` should behave differently. I also assume that `CONDA_EXE` will always be hardcoded at the top of the file as it was for me, meaning that $?_CONDA_EXE would always be true.

simonleary-umass-edu:
I also solved the problem of finding the difference between “initialization with conda aliased” and “i’m calling conda with no arguments”

PR #12268, opened 2023-01-23
Move coverage from setup.cfg to ease debugging

dholth:
### Description

I always run tests in the debugger, in vscode, but `--cov` prevents the debugger from working.

Instead, enable coverage in individual test runner scripts, for CI.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -218,7 +218,7 @@ jobs:
           -e TEST_GROUP
           ghcr.io/conda/conda-ci:main-linux-python${{ matrix.python-version }}
           bash -c "source /opt/conda/etc/profile.d/conda.sh; \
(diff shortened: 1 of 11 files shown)


jezdez:
We'll need a way to collect coverage data and report about it for people that aren't using VS Code, will this collect and print out coverage reports locally when following the contribution policy?

dholth:
We would have to ask people to say `--cov` or make sure it is in a test script that we also ask them to run.

jezdez (changes requested):
This would require updating to mention the need to specify --cov: https://github.com/conda/conda/blob/main/docs/source/dev-guide/development-environment.md

jezdez:
Oh BTW, I think it would be cool to learn how you're using the VS Code debugger BTW.

PR #12271, opened 2023-01-24
modify test to account for boltons in defaults

dholth:
### Description

A test requires boltons to be absent from defaults but it was added on 23 Jan 2023. Use `--override-channels` to look in the specified channel only, instead. Presumably boltons is in the private version of this test channel, but I'm not a member.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -1824,7 +1824,18 @@ dependencies:
             assert yml_obj['channels'] == [channel_url.replace('cqgccfm1mfma', '<TOKEN>'), 'defaults']
 
             with pytest.raises(PackagesNotFoundError):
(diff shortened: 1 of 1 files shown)


kenodegard (approved):
hm I don't like this test using the `data-portal` channel, but this fix is good enough

dholth:
> hm I don't like this test using the `data-portal` channel, but this fix is good enough

We should delete the test instead? Does the test even make sense?

kenodegard:
> > hm I don't like this test using the `data-portal` channel, but this fix is good enough
> 
> We should delete the test instead? Does the test even make sense?

The test makes sense and is reasonable to keep IMO

PR #12274, opened 2023-01-25
Improve repodata.json cache state implementation; fix some tests

dholth:
### Description

Test improvements, repodata.json.state improvements from features/repodata-jlap that are not directly related to jlap support. Should make conda better, and the other PR simpler.

### Checklist - did you ...


conda/core/subdir_data.py
```diff
@@ -415,7 +429,11 @@ def _pickle_valid_checks(self, pickled_state, mod, etag):
         yield "_pickle_version", pickled_state.get("_pickle_version"), REPODATA_PICKLE_VERSION
         yield "fn", pickled_state.get("fn"), self.repodata_fn
 
-    def _read_pickled(self, state):
+    def _read_pickled(self, state: RepodataState):
+        if not isinstance(state, RepodataState):
+            _state = RepodataState(self.cache_path_json, self.cache_path_state, self.repodata_fn)
```
dholth, line 434: Maybe add an easier RepodataState-from-dict-or-RepodataState constructor.

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +300,102 @@ def conda_http_errors(url, repodata_fn):
...
+        """
+        try:
+            state_path = self.cache_path_state
+            log.debug("Load %s cache from %s", self.repodata_fn, state_path)
+            state = json.loads(state_path.read_text())
+            # json and state files should match
+            json_stat = self.cache_path_json.stat()
+            if not (
+                state.get("mtime_ns") == json_stat.st_mtime_ns
+                and state.get("size") == json_stat.st_size
```
dholth, line 339: Are we testing the cache invalidation flow thoroughly enough, to verify that it does or does not use the cache depending, and that it doesn't think that a concurrent conda overwrote repodata.json while we are trying to overwrite repodata.state.json?

tests/core/test_subdir_data.py
```diff
@@ -276,7 +279,7 @@ def test_metadata_cache_works(platform=OVERRIDE_PLATFORM):
 
     with env_vars(
         {"CONDA_PLATFORM": platform}, stack_callback=conda_tests_ctxt_mgmt_def_pol
-    ), patch.object(CondaRepoInterface, "repodata", return_value={}) as fetcher:
+    ), patch.object(CondaRepoInterface, "repodata", return_value="{}") as fetcher:
```
dholth, line 282: just an incorrect return type

conda/core/subdir_data.py
```diff
@@ -339,17 +342,27 @@ def _load(self):
                 self.url_w_repodata_fn,
             )
             touch(self.cache_path_json)
+            # change mtime in state file, or cache will be thrown away next time:
+            self._save_state(mod_etag_headers)
             _internal_state = self._read_local_repodata(mod_etag_headers)
             return _internal_state
         else:
-            if not isdir(dirname(self.cache_path_json)):
+            if not isdir(dirname(self.cache_path_json)):  # What happens if it is a directory?
```
dholth, line 350: Oh, it is using isdir() instead of an exists() check.

dholth:
It is really hard to write repodata.json and an associated state + mtime correctly without locking. Could it work something like this:

- separate read-from/write to paths in case we are updating repodata.json (would be important for jlap feature)
- get mtime etc. from named temporary "write repodata here" file
- write .state.json with mtime, size etc. possibly also with a temporary file / move strategy.
- move "write repodata here" on top of permanent path

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +304,291 @@ def conda_http_errors(url, repodata_fn):
...
+        with temp_path.open("wx") as temp:  # exclusive mode, error if exists
+            temp.write(data)
+
+    def replace(self, temp_path: Path):
+        """
+        Rename path onto <repodata>.json path, synchronize state.
+
+        Relies on path's mtime not changing on move.
+        """
+        with self.cache_path_state.open("w+") as state_file, _lock(state_file):
```
dholth, line 555: does `w+` (open for writing and reading) truncate the file, before bothering to check if we can get the lock?

conda/gateways/repodata/__init__.py
```diff
@@ -289,3 +306,371 @@ def conda_http_errors(url, repodata_fn):
...
+                # clear mod, etag, cache_control to encourage re-download
+                state.update({"etag": "", "mod": "", "cache_control": "", "size": 0})
+            self.update(state)  # allow all fields
+        except (json.JSONDecodeError, OSError):
+            log.debug("Could not load state", exc_info=True)
+            self.clear()
+        return self
+
+    @deprecated("23.3", "23.9", addendum="use RepodataCache")
+    def save(self):
```
dholth, line 358: Added this in 23.1

dholth:
Closing as too complicated to maintain separate from the features/repodata-jlap branch (which includes all of these changes)

PR #12279, opened 2023-01-25
Use ruff

dholth:
### Description

Refs #12445 

ruff is a faster linter.

Before: `pre-commit run --all  58.99s user 4.11s system 126% cpu 49.833 total`

With ruff: `pre-commit run --all  9.92s user 2.04s system 456% cpu 2.622 total`

It is frustrating when a long CI run gets thrown out because of failing pre-commit checks. You can add pre-commit locally, so it runs before "git commit", but it breaks flow when that takes too long to run. Instead, use ruff, and local, checked commits are delightful again.

### Checklist - did you ...


jezdez:
I understand what you're trying to do, but we should for the sake of stability first check how ruff is maintained currently (a solo maintainer project), I'll reach to them.

.pre-commit-config.yaml
```diff
@@ -45,15 +45,11 @@ repos:
...
+    # Ruff version.
+    rev: 'v0.0.233'
     hooks:
-      - id: flake8
-  - repo: https://github.com/PyCQA/pylint
-    rev: v2.16.0b1
-    hooks:
-      - id: pylint
-        args: [--exit-zero]
+      - id: ruff
```
dholth, line 52: ruff can also do the bandit checks possibly with an added bandit dependency

dbast:
seems like they re-implement bandit lints and track that here https://github.com/charliermarsh/ruff/issues/1646

dholth:
> seems like they re-implement bandit lints and track that here [charliermarsh/ruff#1646](https://github.com/charliermarsh/ruff/issues/1646)

Great, then we would simply enable, probably the "S" prefix, in pyproject.toml

dbast:
@dholth Pretty much like pre-commit finishing in under 3 secs! that e.g. allows running it on any file change via vs code and thus incorporate the pre-commit feedback file coding instead of somewhen later.

There are quite some projects listed also doing ruff https://github.com/charliermarsh/ruff

ruff or not to ruff... the important thing is that pylint gets removed, as it is slow and with exit-zero not bringing any value.

The numbers without pylint, but still with flake8 (no ruff yet) are also pretty good already: `12.76s user 2.59s system 452% cpu 3.395 total`

kenodegard:
@dholth @dbast I agree with removing both `pylint` and `bandit`, we just never got around to do anything meaningful with them

I'm unsure about switching to `ruff`, it's simply too new and potentially volatile

dholth:
I don't see the risk in adding a new linter, to replace a broken linter. If it stops working, remove it again?

jezdez:
I'm willing to reconsider this again now that Charlie has created a [company to support ruff](https://astral.sh/blog/announcing-astral-the-company-behind-ruff) development

tests/test_exports.py
kenodegard, line 1: The tests defined here are pointless to have

kenodegard:
pre-commit.ci autofix

kenodegard (approved):

beeankha (approved):

PR #12284, opened 2023-01-27
list CURL_CA_BUNDLE as option too

bbodenmiller:
List `URL_CA_BUNDLE` as non-standard certificates option too


bbodenmiller:
Closing. `CURL_CA_BUNDLE` does not appear to work for some reason.

PR #12285, opened 2023-01-27
Better CLI parsing support for plugins

travishathaway:
This is a pull request to improve the experience of writing subcommand plugins for conda. The goal is to better support the usage of other libraries for parsing command line options and arguments such as `click`.

The most important change in this pull request is the change to [`conda.cli.conda_argparse.ArgumentParser`](https://github.com/travishathaway/conda/blob/932d8c1ec97b76b5da1743ec4e351c84b6a785e8/conda/cli/conda_argparse.py#L212-L225). Before, we relied on the argument parser to hit an error before we routed to the appropriate subcommand. With this approach, we override the `parse_args` method of `ArgumentParser` to instead intercept the subcommand and return a `Namespace` object we use later to invoke the actual subcommand (see: [conda/cli/conda_argparse.py#L86-L98](https://github.com/travishathaway/conda/blob/932d8c1ec97b76b5da1743ec4e351c84b6a785e8/conda/cli/conda_argparse.py#L86-L98)).

On top of refactoring the code to make the execute path a little more clear for plugin subcommand execution, this pull request also makes it easier for other libraries to integrate with conda, especially [click ](https://click.palletsprojects.com/).

Here is an example of what is now possible with this improvement: 

```python
import click

from conda.plugins import hookimpl, CondaSubcommand


# Defining this group here properly nests this as a conda subcommand in click
@click.group("conda")
def cli():
    pass


@cli.command()
def my_command():
    """
    Subcommand for invoking conda bot wars!!!
    """
    print("Test command")


@hookimpl
def conda_subcommands():
    yield CondaSubcommand(
        name="my_command",
        summary=my_command.__doc__,
        action=cli,
        no_sys_argv=True
    )
```


conda/cli/conda_argparse.py
```diff
@@ -199,6 +210,21 @@ def _check_value(self, action, value):
...
+    def parse_args(self, args=None, namespace=None):
+        """
+        We override this method to check if we are running from a known plugin subcommand.
+        If we are, we do not want to handle argument parsing as this is delegated to the plugin
+        subcommand. We instead return a ``Namespace`` object with ``plugin_subcommand`` defined,
+        which is a ``conda.plugins.CondaSubcommand`` object.
+        """
+        if len(sys.argv) > 1:
+            name = sys.argv[1]
+            subcommand = next((sub for sub in self._subcommands if sub.name == name), None)
```
beeankha, line 222: 🤩

tests/plugins/test_subcommands.py
```diff
@@ -22,6 +24,25 @@ def conda_subcommands(self):
         )
 
 
+class SubcommandPluginWithCustomParser:
+    """
+    This version of our test subcommand does not except any values for it's
```
beeankha, line 29:
```suggestion
    This version of our test subcommand does not except any values for its
```

beeankha (dismissed):

conda/plugins/types.py
```diff
@@ -13,14 +13,16 @@ class CondaSubcommand(NamedTuple):
...
     """
 
     name: str
     summary: str
     action: Callable[
-        [list[str]],  # arguments
+        ...,  # arguments
         int | None,  # return code
     ]
+    no_sys_argv: bool = False
```
jezdez, line 25: Double-negative which looks weird when writing it as `no_sys_argv=True`, should be:
```suggestion
    pass_sys_argv: bool = True
```

jezdez (changes requested):
I'm uncertain if I understand the implementation method of the PR, why does the `click` example you posted in the PR description reimplement the `conda` main command?

Do you think non-grouped click commands work with this? Do we need to pass to click (or other CLI libraries like typer) additional information what has alredy been parsed by the main argparse parser (e.g. `sys.argv[1:]`)?

From the click example, does this work?
```
import click

from conda.plugins import hookimpl, CondaSubcommand


@click.command()
@click.option('--count', default=1, help='Number of greetings.')
@click.option('--name', prompt='Your name',
              help='The person to greet.')
def hello(count, name):
    """Simple program that greets NAME for a total of COUNT times."""
    for x in range(count):
        click.echo(f"Hello {name}!")


@hookimpl
def conda_subcommands():
    yield CondaSubcommand(
        name="my_command",
        summary=hello.__doc__,
        action=hello,
        pass_sys_argv=False,
    )
```

conda/plugins/types.py
travishathaway: Thanks for catching this. Will change 👍

travishathaway:
Hi @jezdez,

Thanks for your review and getting me to think a little bit more about whether this is in fact actually needed.

To answer your question, no that won't work the way you have written it, but this will:

```python
@click.command()
@click.argument("name")
def say_hi(name):
    """
    Say hi
    """
    print(f"Hi, {name}")


def wrapper(args):
    say_hi(args=args, prog_name="conda say_hi")


@hookimpl
def conda_subcommands():
    yield CondaSubcommand(
        name="say_hi",
        summary=say_hi.__doc__,
        action=wrapper
    )
```

You'll notice that we had to add an extra `wrapper` function and then add some (not super well documented) arguments while calling the command function. This method has the benefit of not having to change the existing API at all with the downside of having to define an additional wrapper function.

I think this method is better, but I would still like to keep the work I have done to refactor the sub-command argument parsing (minus the addition to the `CondaSubcommand` class), so I will go back and change some things.

In the meantime, please let me know what you think of the above example.

travishathaway:
@beeankha  and @jezdez this is ready for a second review.

beeankha (dismissed):

news/12285-improve-plugin-subcommand-arg-parsing
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Refactor subcommand argument parsing to make it easier to understand (#12285)
```
jezdez, line 3: This should add a mention of the change in when plugin subcommands are being called, earlier in the stack.

conda/cli/conda_argparse.py
```diff
@@ -199,6 +205,21 @@ def _check_value(self, action, value):
...
+            if subcommand:
+                return Namespace(plugin_subcommand=subcommand)
```
jezdez, line 219: A little easier to understand:

```suggestion
            for subcommand in self._subcommands:
                if subcommand.name == name:
                    return Namespace(plugin_subcommand=subcommand)
```

Do we need to pass `args` and `namespace` here, too?

conda/cli/conda_argparse.py
```diff
@@ -84,6 +86,14 @@ def generate_parser():
 
 
 def do_call(args, parser):
+    """
+    Serves as the primary entry point for commands referred to in this file and for
+    all registered plugin subcommands.
+    """
+    # First, check if this is a plugin subcommand; if this attribute is present then it is
```
jezdez, line 93: To be clear, `args` here is a `Namespace` object?

jezdez (changes requested):
I'm a little worried that we don't add additional tests for this as it changes the stack order of running commands. When the `conda-*` legacy behavior is removed, the users of that API will need to be able to learn about the new order somehow.

Related to that, I think with this change, it'll be possible to override built-in commands via plugins, which isn't something we should accidentally add as it could enable people to override install etc.

Could we instead move the check for subcommands to after it looks for built-in subcommands?

conda/cli/conda_argparse.py
travishathaway: No we do not. The `parse_args` method only returns a single `Namespace` object.

conda/cli/conda_argparse.py
travishathaway: Yes.

travishathaway: Most places you see `args` in conda, it is indeed a `Namespace` object.

travishathaway:
@jezdez,

Thanks for the thorough review. I have already incorporated some of your suggestions.

I need to go back to the drawing board to make sure I have taken into account what you mentioned in your comment. This will take some time. I hope to get to this sometime next week.

conda/cli/conda_argparse.py
```diff
@@ -199,6 +226,45 @@ def _check_value(self, action, value):
         else:
             super()._check_value(action, value)
 
+    @staticmethod
+    def _get_attempted_override_error(name: str, summary: str):
+        return (
+            f"The plugin '{name}: {summary}' is trying to override "
+            f"the built command {name}, which is not allowed. Please uninstall this plugin "
```
travishathaway, line 233:
```suggestion
            f"the built-in command {name}, which is not allowed. Please uninstall this plugin "
```

travishathaway:
Hi @jezdez,

I just wanted to ping you again regarding this pull request. I have gone back and refactored the code a bit. The code now has a static list of "built-in" commands that cannot be overridden. If a plugin does attempt to override it, we simply show an error and skip the invocation of it.

I left the `env` commands out of this though because I think we'd like to eventually re-migrate that to be a plugin-in too? (not sure here).

I would be happy to hear your thoughts once more on this pull request.

Thanks!

conda/cli/conda_argparse.py
```diff
@@ -199,6 +226,45 @@ def _check_value(self, action, value):
...
+            name = sys.argv[1]
+            for subcommand in self._subcommands:
+                if subcommand.name == name:
+                    if name.lower() in BUILTIN_COMMANDS:
+                        error_msg = self._get_attempted_override_error(
+                            subcommand.name, subcommand.summary
+                        )
+                        log.error(error_msg)
+                    else:
+                        return Namespace(plugin_subcommand=subcommand)
```
kenodegard, line 252: I don't think there's any value in defining this as a helper method, lets's inline it below

tests/plugins/test_subcommands.py
```diff
@@ -46,10 +59,36 @@ def test_help(plugin, cli_main, capsys):
...
+    # setup
+    plugin = SubcommandPlugin(name=builtin_command)
+    expected_error = ArgumentParser._get_attempted_override_error(plugin.name, plugin.summary)
+    plugin_manager.register(plugin)
+
+    # run code under test
+    run(f"conda {builtin_command} --help")
+
+    # assertions; make sure we got the right error messages and didn't invoke the custom command
+    assert mock_log.error.mock_calls == [mock.call(expected_error)]
```
kenodegard, line 92: Can we use `pytest.raises` instead?

conda/cli/conda_argparse.py
```diff
@@ -84,6 +107,14 @@ def generate_parser():
 
 
 def do_call(args, parser):
+    """
+    Serves as the primary entry point for commands referred to in this file and for
+    all registered plugin subcommands.
+    """
+    # First, check if this is a plugin subcommand; if this attribute is present then it is
+    if hasattr(args, "plugin_subcommand") and isinstance(args.plugin_subcommand, CondaSubcommand):
```
kenodegard, line 115:
```suggestion
    if getattr(args, "plugin_subcommand", None):
```

kenodegard (changes requested):
Hm, I like this

tests/plugins/test_subcommands.py
```diff
@@ -1,23 +1,30 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+from unittest import mock
+
```
jezdez, line 4: No need to import mock here since `mocker.call` also works, https://pytest-mock.readthedocs.io/en/latest/usage.html

conda/cli/conda_argparse.py
```diff
@@ -199,6 +226,45 @@ def _check_value(self, action, value):
...
             super()._check_value(action, value)
 
+    @staticmethod
+    def _get_attempted_override_error(name: str, summary: str):
+        return (
+            f"The plugin '{name}: {summary}' is trying to override "
+            f"the built-in command {name}, which is not allowed. Please uninstall this plugin "
+            "to stop seeing this error message"
+        )
+
```
jezdez, line 236:
```suggestion
```
Don't think we need a static method here, inlining below is okay.

conda/cli/conda_argparse.py
```diff
@@ -199,6 +226,45 @@ def _check_value(self, action, value):
...
+        """
+        if len(sys.argv) > 1:
+            name = sys.argv[1]
+            for subcommand in self._subcommands:
+                if subcommand.name == name:
+                    if name.lower() in BUILTIN_COMMANDS:
+                        error_msg = self._get_attempted_override_error(
+                            subcommand.name, subcommand.summary
+                        )
+                        log.error(error_msg)
```
jezdez, line 250:
```suggestion
                        error_message = dals(f"""
                            The plugin '{subcommand.name}' ({subcommand.summary})
                            is trying to override the built-in command {name},
                            which is not allowed. Please uninstall the plugin
                            to stop seeing this error message.
                        """)
                        log.error(error_message)

```

tests/plugins/test_subcommands.py
```diff
@@ -32,12 +39,18 @@ def plugin(mocker, plugin_manager):
 
 
 def test_invoked(plugin, cli_main):
+    """
+    Ensure we are able to invoke our command after creating it
+    """
```
jezdez, line 44: Thank you for these docstring updates!! 🤩

tests/plugins/test_subcommands.py
jezdez: This should test the `log.error` call with [`caplog`](https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture)

jezdez (changes requested):
A few nits around how the error is tested (if it's done via the logging system, we can test for it). Once that's handled, I think it's good to go.

tests/plugins/test_subcommands.py
travishathaway: Neat! Thanks for the tip.

conda/cli/conda_argparse.py
travishathaway: In my opinion, the value comes from being able to more easily read the `parse_args` method. It also helps to decrease the overall code complexity of that method.

But, it seems like Jannis also thinks this is a good idea, so I will inline it. We can always refactor later 😉.

tests/plugins/test_subcommands.py
travishathaway: That wouldn't be possible because we do not raise an exception, right? 😅

travishathaway: We just want to make sure the call to `logger.error` was done in the way we expected it to be (i.e. printing a message to the user about the attempt to override a built-in sub-command).

conda/cli/conda_argparse.py
```diff
@@ -199,6 +225,34 @@ def _check_value(self, action, value):
...
+            name = sys.argv[1]
+            for subcommand in self._subcommands:
+                if subcommand.name == name:
+                    if name.lower() in BUILTIN_COMMANDS:
+                        error_msg = dals(
+                            f"The plugin '{subcommand.name}: {subcommand.summary}' is trying "
+                            f"to override the built-in command {name}, which is not allowed. "
+                            "Please uninstall this plugin to stop seeing this error message"
+                        )
+                        log.error(error_msg)
```
jezdez, line 247:
```suggestion
                        error_message = dals(
                            f"The plugin '{subcommand.name}: {subcommand.summary}' is trying "
                            f"to override the built-in command {name}, which is not allowed. "
                            "Please uninstall this plugin to stop seeing this error message"
                        )
                        log.error(error_message)
```

jezdez (approved):

tests/plugins/test_subcommands.py
kenodegard: oooh this is logger 🤦🏼‍♂️, can we use `caplog` instead?

https://docs.pytest.org/en/6.2.x/logging.html

travishathaway: Sounds good. I was not aware of this fixture.

travishathaway: @kenodegard , although this is a great idea, the implementation is proving to be exceedingly difficult. Therefore, I propose keeping the mocking behavior instead.

jezdez (approved):

conda/cli/conda_argparse.py
```diff
@@ -199,6 +225,34 @@ def _check_value(self, action, value):
...
+        """
+        We override this method to check if we are running from a known plugin subcommand.
+        If we are, we do not want to handle argument parsing as this is delegated to the plugin
+        subcommand. We instead return a ``Namespace`` object with ``plugin_subcommand`` defined,
+        which is a ``conda.plugins.CondaSubcommand`` object.
+        """
+        plugin_subcommand = None
+
+        if len(sys.argv) > 1:
+            name = sys.argv[1]
```
kenodegard, line 238: Both of the following are possible and succeed:

```
conda --debug info
conda info --debug
```

So we shouldn't be exclusively checking the second argument. Is this something we should address here or deal with later?

travishathaway: @kenodegard, thanks for bring this up. I think we don't have to worry about this for now.

Even if you tried to override the `info` command, it still wouldn't succeed in either case because we would either be checking the sub-command against "info" (which it would recognize as a built-in) or checking to see if `--debug` was a valid sub-command name.

Maybe you could elaborate a little more on exactly what you see as the risk here or how this would break things.

kenodegard: The `info` subcommand was just an example, any subcommand can be run with certain flags present to the left of the subcommand:

```
conda --debug create
conda --debug config
conda --debug install
conda --debug search
```

Even the old way of invoking plugins would work:

```
conda --debug build
```

But this implementation wouldn't work whenever conda-build is refactored to use the new plugin framework since we are only looking at the second argument.

travishathaway: Fair point. Because this currently doesn't break any of that behavior, I believe it is okay to do nothing for now. When we finally do decide to remove that old behavior of invoking sub-commands, we need to make sure that it accounts for this.

kenodegard (approved):

PR #12297, opened 2023-01-30
Tune OSX test splits

dholth:
### Description

Update: Initially this split the OSX integration tests into only 2 runs, but the "add conda-forge CI" merge conflicts with that. The PR retains use of test durations and the more accurate splitting algorithm.

According to https://jerry-git.github.io/pytest-split/#usage, you only need to store durations after significant changes to the test suite.

Since we spend 6-12 minutes in Setup Environment, two unit tests will save us that much runner time, might help with limits?

We will need to fix whatever bz2 cache problems conda had.

Added a `tests/combine...` script that could help us to periodically update our test durations, with a long comment explaining how to get the artifacts back down from github actions.

### Checklist - did you ...


.github/workflows/tests.yml
```diff
@@ -245,10 +247,19 @@ jobs:
...
         test-group: ['1', '2', '3']
+        # fewer test splits for quicker unit tests, keep 3 splits for integration
+        test-splits: [2, 3]
+        exclude:
+          - test-type: unit
+            test-group: '3'
+          - test-type: unit
+            test-splits: 3
+          - test-type: integration
+            test-splits: 2
```
jezdez, line 258: Oh to clarify, this basically binds the test type with the test split nummber, by excluding the pairs that we don't want, and then also corrects on the unit test type not needing to run a test group 3, right?

dev/macos/setup.sh
```diff
@@ -11,8 +11,9 @@ sudo cp minio /usr/local/bin/minio
...
-# make sure the caching works correctly
-conda config --set use_only_tar_bz2 true
+# .conda-only will become the norm. previously claimed "only tar.bz2" was
+# required for caching; does this still break caching?
+conda config --set use_only_tar_bz2 false
 # install all test requirements
-conda install --quiet --name conda-test-env --yes --file tests/requirements.txt
-conda update openssl ca-certificates certifi
+conda install --yes --name conda-test-env --file tests/requirements.txt
+conda update --yes openssl ca-certificates certifi
```
jezdez, line 19: No quiet anymore?

jezdez (commented):
Neat, let's try this!

dev/macos/setup.sh
dholth: Maybe wanted to see the logs

dholth:
Now we download the run artifacts, combine all timing and commit that.

jezdez (approved):

dholth:
I'm comparing an alternative PR https://github.com/conda/conda/actions/runs/4053810805/jobs/6974890705 using the "no durations available" strategy, to this PR https://github.com/conda/conda/actions/runs/4055329924

I'm not sure things are more even? We could try averaging test durations for OSX only (since it is the most troubled runner), for many runs, or by providing a per-platform set of durations chosen with `--durations-path`. There's also the lesat-duration splitting algorithm, which changes the test order to get a more even split. https://jerry-git.github.io/pytest-split/#splitting-algorithms

jezdez (approved):

dholth:
We'd better check the test order dependency (they are not reordered yet, just put into different groups) for that failing Windows shell test.

.test_durations
```diff
@@ -0,0 +1,1198 @@
+{
```
jaimergp, line 1: Are these files expected to be committed or should they be `gitignore`d?

tests/test_resolve.py
```diff
@@ -99,9 +100,12 @@ def test_iopro_mkl(self):
         ])
 
     def test_mkl(self):
-        a = r.install(['mkl 11*', MatchSpec(track_features='mkl')])
-        b = r.install(['mkl'])
-        assert a == b
+        SubdirData._cache_.clear()
+        Channel._cache_.clear()
```
jaimergp, line 104: Should this be part of a `setUp` method for all tests?

.test_durations
dholth: These do get committed, so that the tests have an estimate of how long the tests will take.

tests/test_resolve.py
dholth: This is still an experiment, and apparently, it does not help this test to pass.

conda/testing/helpers.py
```diff
@@ -305,7 +305,7 @@ def _patch_for_local_exports(name, subdir_data, channel, index):
     subdir_data._mtime = float("inf")
 
 
-@lru_cache(maxsize=None)
+# @lru_cache(maxsize=None)
```
dholth, line 308: We need a fresh `index_r_1` to fix a test run order dependency on test_mkl. This is only called a few times.

tests/test_misc.py
```diff
@@ -85,27 +85,30 @@ def test_explicit_missing_cache_entries(mocker):
...
-        AssertionError, match="Missing package cache records for: pkgs/linux-64::foo==1.0.0=py_0"
-    ):
-        explicit(
-            [
-                "http://test/pkgs/linux-64/foo-1.0.0-py_0.tar.bz2",  # does not exist
-                PackageCacheData.get_all_extracted_entries()[0].url,  # exists
-            ],
-            "",
-        )
+    with make_temp_env() as prefix:  # ensure writable env
```
dholth, line 88: Another order dependency. After enough tests have run, `context.target_prefix` has packages; before enough tests have run, the read-only /opt/conda in the tests container has no packages. ?? so how does it get packages later during the normal test runs, since it is read-only? (we do rm -rf the package cache at the beginning presumably to solve some other problem)

tests/test_resolve.py
dholth: Update: this test can pass

beeankha (approved):

PR #12302, opened 2023-01-31
cache minio for windows

dholth:
### Description

Should save time downloading minio.exe. Downloaded where the cache can find it, then copied to (I'm not sure where) CONDA_PREFIX is.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -96,6 +96,16 @@ jobs:
             ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-py${{ env.PYTHON }}-${{
             hashFiles('recipe/meta.yaml', 'dev/windows/setup.bat', 'tests/requirements.txt') }}
 
+      - name: Cache minio
+        uses: actions/cache@v3
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -96,6 +96,16 @@ jobs:
...
+      - name: Cache minio
+        uses: actions/cache@v3
+        env:
+          # Increase this value to reset cache
+          CACHE_NUMBER: 3
+        with:
+          path: minio.exe
+          key:
+            ${{ runner.os }}-minio-${{ env.CACHE_NUMBER }}
+
```
jezdez, line 108: I know the minio setup is in the `setup.bat`, could we extract the one line from there to have the setup and cache section live side-by-side here?

dholth: I suppose? I don't think about cache as necessarily being directly related/adjacent, it just tries to make files available, and at the end of the test run uploads those back into the cache.

jezdez (approved):

dholth:
Thanks for the merge!

PR #12309, opened 2023-02-01
remove check for conda-build that doesn't understand .conda

dholth:
### Description

conda-build 3.18.3 is from 2019. Remove a check, constrain the minimum conda-build version to be at least 3.18.3.

### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -50,7 +50,7 @@ requirements:
     - pluggy >=1.0.0
     - tqdm >=4
   run_constrained:
-    - conda-build >=3
+    - conda-build >=3.18.3
     - conda-env >=2.6
     - conda-content-trust >=0.1.1
     - cytoolz >=0.8.1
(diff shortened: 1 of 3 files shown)


jezdez:
I think we should create a ticket for this as it might be triggering unforeseen compat issues, e.g. teams using old conda-builds etc.

dholth:
> I think we should create a ticket for this as it might be triggering unforeseen compat issues, e.g. teams using old conda-builds etc.

#12318

jezdez (approved):

recipe/meta.yaml
```diff
@@ -50,7 +50,7 @@ requirements:
     - pluggy >=1.0.0
     - tqdm >=4
   run_constrained:
-    - conda-build >=3
+    - conda-build >=3.18.3
```
jezdez, line 53: @dholth Could you also send PRs for the feedstocks to make sure we're not missing this?

Issue #12310, opened 2023-02-01
channel notices doesn't handle i/o errors


beckermr:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

I was running conda, and on the system I am on, sometimes I cannot access my home dir. conda flaked out due to an io error.

### Conda Info

```shell
$ conda info

     active environment : base
    active env location : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3
            shell level : 1
       user config file : /u/ki/beckermr/.condarc
 populated config files : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3/.condarc
                          /u/ki/beckermr/.condarc
          conda version : 22.11.1
    conda-build version : not installed
         python version : 3.9.9.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.12=0
                          __linux=2.6.32=0
                          __unix=0=0
       base environment : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3  (writable)
      conda av data dir : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3/pkgs
                          /u/ki/beckermr/.conda/pkgs
       envs directories : /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3/envs
                          /u/ki/beckermr/.conda/envs
               platform : linux-64
             user-agent : conda/22.11.1 requests/2.28.1 CPython/3.9.9 Linux/2.6.32-754.33.1.el6.x86_64 rhel/6.10 glibc/2.12 solver/libmamba conda-libmamba-solver/22.8.1 libmambapy/1.1.0
                UID:GID : 10150:1092
             netrc file : /u/ki/beckermr/.netrc
           offline mode : False
```

### Conda Config

```shell
$ conda config --show-sources
==> /nfs/slac/des/fs1/g/sims/beckermr/mambaforge3/.condarc <==
channels:
[...]

PR #12312, opened 2023-02-01
Channel Notices doesn't handle i/o errors

travishathaway:
Fix for this issue:

- https://github.com/conda/conda/issues/12310

### DIFF
--- a/conda/notices/core.py
+++ b/conda/notices/core.py
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
-from functools import wraps
+import logging
 import time
+from functools import wraps
 from typing import Sequence
 
 from ..base.context import context, Context
@@ -18,6 +19,8 @@ from .types import ChannelNotice, ChannelNoticeResponse, ChannelNoticeResultSet
 # Used below in type hints
 ChannelName = str
 ChannelUrl = str
+
+logger = logging.getLogger(__name__)
 
 
 def retrieve_notices(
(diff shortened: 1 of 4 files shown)


conda/notices/core.py
```diff
@@ -44,7 +44,11 @@ def retrieve_notices(
 
     # We always want to modify the mtime attribute of the file if we are trying to retrieve notices
     # This is used later in "is_channel_notices_cache_expired"
-    cache_file.touch()
+    try:
```
beckermr, line 47: This catches the specific error I had, but I think there other i/o operations below that would also fail if the cache file cannot be changed/read/etc., right?

dholth: There's a get first writable function used in subdir_data that might be the thing.

travishathaway: @beckermr, good catch. I still want to figure out how to properly test this so I rest assured I know what I'm doing 😅. .

travishathaway:
@beckermr, did a more thorough job today. Would you mind looking it over once more? 😄

tests/cli/test_main_notices.py
```diff
@@ -296,3 +296,41 @@ def test_notices_work_with_s3_channel(notices_cache_dir, notices_mock_http_sessi
...
+    writable, I still want commands to run and not end up failing.
+    """
+    env_name = "testenv"
+    error_message = "Can't touch this"
+
+    with mock.patch("conda.notices.cache.open") as mock_open, mock.patch(
+        "conda.notices.core.logger.error"
+    ) as mock_logger:
+        mock_open.side_effect = [PermissionError(error_message)]
+        run(f"conda create -n {env_name} -y -c local --override-channels")
```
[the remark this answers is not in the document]
travishathaway: I've added the check. Successful runs of those functions return `None` though because it's not actually spawning a sub-process.

jezdez (approved):

PR #12335, opened 2023-02-08
Build Docker images with conda-forge base too

jaimergp:
### Description

Comes from https://github.com/conda/conda-libmamba-solver/pull/120.

This PR provides an additional Docker image for CI testing. It just provides one more option to debug installation issues. We are often mixing channels (technically unsupported) to provide extra dependencies not available on `defaults` (project not packaged, version out of date, etc).

This would help maintainers choose if they want to also test with a pure conda-forge setup, or maybe just for local debugging when trying to reproduce issues users with a pure conda-forge setup are facing.

### Checklist - did you ...


.github/workflows/images.yml
```diff
@@ -15,6 +15,8 @@ on:
   # NOTE: github.event context is push payload:
   # https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push
   push:
+    branches:
+      - main
```
jezdez, line 19: That would take away the ability to update the image in a branch, which we've used a few times, making it easier to handle feature branches without changing the main image.

.github/workflows/images.yml
```diff
@@ -76,14 +83,14 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           flavor: |
             latest=false
-            suffix=-linux-python${{ matrix.python }}
+            suffix=-linux-python${{ matrix.python }}${{ matrix.channel == 'conda-forge' && '-conda-forge' || '' }}
```
jezdez, line 86: I don't think we've agreed to maintain the main image name that way, I could just see this being 
```suggestion
            suffix=-linux-python${{ matrix.python }}-${{ matrix.channel }}
```

Dockerfile
```diff
@@ -19,22 +19,39 @@ FROM --platform=$TARGETPLATFORM debian:stable-slim AS buildbase
...
+            MINICONDA_URL="https://github.com/conda-forge/miniforge/releases/${CONDA_VERSION}/download/Miniforge3-Linux-ppc64le.sh"; \
+        else \
+            echo "Not supported source channel & target architecture: ${conda_docker_default_channel} & ${TARGETARCH}"; \
+            exit 1; \
+        fi; \
     else \
-        echo "Not supported target architecture: ${TARGETARCH}"; \
-        exit 1; \
+        echo "conda_docker_default_channel value ${conda_docker_default_channel} not supported"; \
+        exit 1; \
```
jezdez: @dbast probably has opinions on this, but could we try to move some of this into a script so it's not as painful to look at?

travishathaway:
Hi @jaimergp,

Thanks for the pull request. I would advise also adding a little text about this in our "development environment" docs:

- https://docs.conda.io/projects/conda/en/stable/dev-guide/development-environment.html

.github/workflows/images.yml
jezdez: Hmpf, yeah, true. Okay, fair enough.

dbast (approved):

jezdez (approved):

Issue #12063, opened 2022-11-04
`conda env list` returning "TypeError: expected str, bytes or os.PathLike object, not NoneType"


kwt5152:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

While trying to run `conda env list` as root I receive the following error message.

```
(base) root@host:~# conda env list

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/usr/local/miniconda3/lib/python3.9/site-packages/conda/exceptions.py", line 1129, in __call__
        return func(*args, **kwargs)
      File "/usr/local/miniconda3/lib/python3.9/site-packages/conda_env/cli/main.py", line 80, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/usr/local/miniconda3/lib/python3.9/site-packages/conda_env/cli/main_list.py", line 36, in execute
        info_dict = {'envs': list_all_known_prefixes()}
      File "/usr/local/miniconda3/lib/python3.9/site-packages/conda/core/envs_manager.py", line 84, in list_all_known_prefixes
        environments_txt_file = get_user_environments_txt_file(home_dir)
      File "/usr/local/miniconda3/lib/python3.9/site-packages/conda/core/envs_manager.py", line 24, in get_user_environments_txt_file
        return expand(join(userhome, '.conda', 'environments.txt'))
      File "/usr/local/miniconda3/lib/python3.9/posixpath.py", line 76, in join
        a = os.fspath(a)
    TypeError: expected str, bytes or os.PathLike object, not NoneType

`$ /usr/local/miniconda3/bin/conda-env list`

  environment variables:
                 CIO_TEST=<not set>
  CONDA_AUTO_UPDATE_CONDA=false
        CONDA_DEFAULT_ENV=base
                CONDA_EXE=/usr/local/miniconda3/bin/conda
             CONDA_PREFIX=/usr/local/miniconda3
    CONDA_PROMPT_MODIFIER=(base)
         CONDA_PYTHON_EXE=/usr/local/miniconda3/bin/python
               CONDA_ROOT=/usr/local/miniconda3
              CONDA_SHLVL=1
           CURL_CA_BUNDLE=<not set>
                     PATH=/usr/local/miniconda3/bin:/usr/local/miniconda3/condabin:/usr/local/sb
[...]

PR #12345, opened 2023-02-10
Bugfix for issue 12063

travishathaway:
This is a bug fix for the following issue:

- https://github.com/conda/conda/issues/12063

This basically just makes the handling of strings a little more robust by filtering out `None` values. There is also a test included to make sure this is covered. I use some mocking to recreate the error state.

### DIFF
--- a/tests/core/test_envs_manager.py
+++ b/tests/core/test_envs_manager.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 
-
+from argparse import Namespace
 from logging import getLogger
 import os
 from os.path import isdir, join, lexists
@@ -13,8 +13,9 @@ from unittest.mock import patch
 from conda.auxlib.collection import AttrDict
 from conda.base.constants import PREFIX_MAGIC_FILE
 from conda.base.context import context, reset_context, conda_tests_ctxt_mgmt_def_pol
+from conda.common.compat import on_win
 from conda.common.io import env_var
-from conda.common.path import paths_equal
+from conda.common.path import paths_equal, expand
 from conda.core.envs_manager import list_all_known_prefixes, register_env, \
     get_user_environments_txt_file, \
     unregister_env, _clean_environments_txt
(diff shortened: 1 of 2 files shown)


news/12345-bugfix-for-list-all-known-prefixes
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Makes the handling of strings in `list_all_known_prefixes` more robust (#12345)
```
dholth, line 7: Make not Makes, but "Filter out `None` path values from `pwd.getpwall()` on Unix systems, for users without home directories, when running as root. (#usually bug number not PR number)"

This "conda should do something for all users when run as admin" behavior is pretty weird and makes me nervous.

travishathaway: Yes, me too.

I was considering to remove it completely, but that would probably break a lot of stuff. The worse part is how it includes all the system users too 😭.

tests/core/test_envs_manager.py
```diff
@@ -112,3 +113,30 @@ def test_list_all_known_prefixes_with_permission_error(mock_clean_env, mock_get_
...
+    mock_is_admin, mock_getpwall, mock_clean_env, mock_context, tmp_path
+):
+    """
+    Regression test for a bug first indentified in this issue: https://github.com/conda/conda/issues/12063
+
+    Tests to make sure that `None` values are filtered out of the `search_dirs` variable in the
+    `list_all_known_prefixes` function.
+    """
+    mock_is_admin.return_value = True
+    mock_getpwall.return_value = [Namespace(pw_dir=expand("~")), Namespace(pw_dir=None)]
```
dholth, line 133: I would use something besides Namespace here, either an inline class or a dataclass or namedtuple, to avoid wondering if this had anything to do with argparse

news/12345-bugfix-for-list-all-known-prefixes
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Filter out None path values from pwd.getpwall() on Unix systems, for users without home directories, when running as root. (#12063)
```
dholth, line 7: These are probably system users, if you `cat /etc/passwd` there are a ton of users for different programs.

dholth (approved):
LGTM!

Issue #12324, opened 2023-02-06
In xonsh, typing `conda --help` causes error


battaglia01:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

There seems to be some weird parsing bug if conda is followed by anything starting with a - in xonsh. For instance, we get all of the following:

```
$ conda --help
usage: xonsh command
xonsh: error: the following arguments are required: command
```

```
$ conda --version
usage: xonsh command
xonsh: error: the following arguments are required: command
```

This even happens if we explicitly try subprocess mode, e.g. $(conda --help) gives the same thing.

Things like `conda init --help` do work. It's only if we do `conda --help` that we get this error.

There doesn't seem to be any traceback.

I've posted this also over at the xonsh repository:

https://github.com/xonsh/xonsh/issues/5038

They've suggested the problem is upstream at conda and related to something happening here:

https://github.com/conda/conda/blob/fdbedc49d4a61fb2121614de22795cfccc562b0b/conda/shell/conda.xsh#L83-L84

https://github.com/conda/conda/blob/fdbedc49d4a61fb2121614de22795cfccc562b0b/conda/shell/conda.xsh#L23-L34

### Conda Info

```shell
active environment : base
    active env location : /Users/mike/Library/miniconda3
            shell level : 6
       user config file : /Users/mike/.condarc
 populated config files :
          conda version : 23.1.0
    conda-build version : not installed
         python version : 3.9.12.final.0
       virtual packages : __archspec=1=arm64
                          __osx=12.5=0
                          __unix=0=0
       base environment : /Users/mike/Library/miniconda3  (writable)
      conda av data dir : /Users/mike/Library/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/osx-arm64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /Users/mike/Library/miniconda3/pkgs
                          /Users/mike/.conda/pkgs
       envs directories : /Users/mike/Library/miniconda3/envs
                          /Users/mike/.conda/envs
               platform : osx-arm64
             user-agent : conda/23.1.0 requests/2.28.1 CPython/3.9.12 Darwin/21.6.0 OSX/12.5
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

[...]

PR #12352, opened 2023-02-13
Fix `xonsh` Command Parser

beeankha:
This is a bug fix for the following issue:
- https://github.com/conda/conda/issues/12324

See the comment in a related bug ticket here for the source of this fix: https://github.com/xonsh/xonsh/issues/5038#issuecomment-1426635346

**Before the fix:**

Screen Shot 2023-02-13 at 3 50 01 PM

**After the fix (`--help` menu shows up properly!):**

Screen Shot 2023-02-13 at 3 49 31 PM

### Checklist - did you ...

- [ ] ~~Add / update outdated documentation?~~

### DIFF
--- a/conda/shell/conda.xsh
+++ b/conda/shell/conda.xsh
@@ -23,7 +23,7 @@ def Env():
 def _parse_args(args=None):
     from argparse import ArgumentParser
     p = ArgumentParser(add_help=False)
-    p.add_argument('command')
+    p.add_argument('command', nargs='?', default=None)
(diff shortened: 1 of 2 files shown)


travishathaway (commented):
Have you thought about how you would potentially test this? I'm guessing the reason this error existed in the first place was because we weren't properly testing this little corner of the code.

If it could be done without too much work (less than 2 hours), I would go for it. Otherwise, I would be happy to give my approval.

Issue #12332, opened 2023-02-07
Deprecate `use_only_tar_bz2` flag or warn when it is in use, since conda-forge only produces `.conda` now


dholth:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Now that `.conda` is widely, and sometimes exclusively used, the use-only-tar-bz2 flag has outlived its usefulness. 

e.g. https://github.com/scipy/scipy/pull/17921

Users who are using this flag will increasingly miss out on all new conda-forge packages, and it's likely that all conda publishers will switch to `.conda`-only in the future.

We could warn if the flag is in use, and especially add text to PackageNotFound's error message warning about the flag.

It might not be practical to search for the missing packages in the .conda section of repodata.json, since at that point in the code we may have thrown away that part of the data.

jezdez:
I think this can be closed via #12353.

jaimergp:
I think we need to add a similar warning to `conda search`. See conversation in https://matrix.to/#/!SOyumkgPRWoXfQYIFH:matrix.org/$168245707744TmQdB:gitter.im?via=matrix.org&via=gitter.im&via=staffchat.ethz.ch

PR #12353, opened 2023-02-14
add note to exception if 'use_only_tar_bz2' is enabled

jaimergp:
### Description

Comes from https://github.com/conda/conda/issues/12332, and https://github.com/conda-incubator/setup-miniconda/issues/267

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -618,7 +618,7 @@ class PackagesNotFoundError(CondaError):
             and use the search bar at the top of the page.
             """)
             from .base.context import context
-            
+
             if context.use_only_tar_bz2:
                 message += dals("""
                 Note: 'use_only_tar_bz2' is enabled. This might be omitting some
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -601,7 +601,7 @@ class PackagesNotFoundError(CondaError):
         format_list = lambda iterable: '  - ' + '\n  - '.join(str(x) for x in iterable)
 
         if channel_urls:
-            message = dals(f"""
+            message = dals("""
             The following packages are not available from current channels:
 
             %(packages_formatted)s
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -601,7 +601,7 @@ class PackagesNotFoundError(CondaError):
         format_list = lambda iterable: '  - ' + '\n  - '.join(str(x) for x in iterable)
 
         if channel_urls:
-            message = dals("""
+            message = dals(f"""
             The following packages are not available from current channels:
 
             %(packages_formatted)s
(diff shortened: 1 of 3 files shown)


conda/exceptions.py
```diff
@@ -617,6 +617,13 @@ def __init__(self, packages, channel_urls=()):
 
             and use the search bar at the top of the page.
             """)
+            from .base.context import context
+
```
dholth, line 621: Let's add a test to this, could go right inside 

```def test_use_only_tar_bz2(platform=OVERRIDE_PLATFORM):
    channel = Channel(join(CHANNEL_DIR, platform))
    SubdirData.clear_cached_local_channel_data()
    with env_var("CONDA_USE_ONLY_TAR_BZ2", True, stack_callback=conda_tests_ctxt_mgmt_def_pol):
        sd = SubdirData(channel)
        precs = tuple(sd.query("zlib"))
        assert precs[0].fn.endswith(".tar.bz2")
```

Only need to raise the exception, and check the message to get code coverage. No need to really "not find" a package.

jezdez: I think this hasn't reached the level of documentation, but it'd be worth it

conda/exceptions.py
```diff
@@ -617,6 +617,13 @@ def __init__(self, packages, channel_urls=()):
 
             and use the search bar at the top of the page.
             """)
+            from .base.context import context
```
jezdez, line 620: Do we need to import this inline? This is a little unfortunate.

jaimergp: See precedent in other exceptions. I am not sure why, maybe import times?

conda/exceptions.py
jaimergp: Added test in `test_exceptions.py`.

tests/test_exceptions.py
```diff
@@ -490,3 +492,17 @@ def test_BinaryPrefixReplacementError(self):
           original data Length: 1404
           new data length: 1104
         """).strip()
+
+    @patch("conda.exceptions.os.isatty", return_value=True)
+    def test_PackagesNotFoundError_use_only_tar_bz2(self, isatty_mock):
+        note = "use_only_tar_bz2"
+        for use_only_tar_bz2 in (True, False):
+            expected = ["", note][use_only_tar_bz2]
```
dholth, line 500: @jaimergp we could do it this way. We don't have the requirement to "not show the note" when the flag is not set, and this regex matcher is bad at negative testing.

jaimergp: As you wish, but I find this way of assigning booleans a bit unreadable. I'd rather see this in code I am maintaining:

```suggestion
            expected = "use_only_tar_bz2" if use_only_tar_bz2 else ""
```

The `raises` + `match` combo is killer though.

tests/test_exceptions.py
```diff
@@ -490,3 +492,17 @@ def test_BinaryPrefixReplacementError(self):
           original data Length: 1404
           new data length: 1104
         """).strip()
+
+    @patch("conda.exceptions.os.isatty", return_value=True)
```
dholth, line 496: this patch is probably useless now

jaimergp:
```suggestion
```

jezdez (approved):

dholth (dismissed):
muy bien

dholth:
Since we are using squash merge, and because I'm bad at rebase, I've been using merge-from-main and never rebase-and-force-push

dholth (approved):

jezdez (approved):

dholth (approved):

PR #12354, opened 2023-02-14
improve reliability of test_prefix_graph

dholth:
### Description

This test can fail because the channel decides to include its full URL (did we have another channel with almost the same name in some in-memory cache?)

Be more robust against that case.


sven6002 (approved):
looks good to me

Issue #12098, opened 2022-11-14
Optional dependencies in environment file


HealthyPear:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Define and install optional dependencies from an environment file (apologies if this is already possible, but I couldn't do it)

### Why is this needed?

It should resemble what is possible in setuptools

### What should happen?

A conda recipe defined by a file `environment.yaml` would look something like

```yaml
name : my_env
channels:
 - conda-forge
dependencies:
 - A
 - B
 - pip
 - pip:
    - G
optional_dependencies
    C:
      - C
    other:
      - D
      - E
      pip:
      - F
    seaborn:
      - seaborn
    all:
      - C
      - other
```

and then the installation command should accept a similar syntax to `pip`, namely

- `conda install -f environment.yaml` would install only (required) `dependencies`
- `conda install -f environment.yaml[C]` would install (required) `dependencies` + `C`
- `conda install -f environment.yaml[C, seaborn]` would install (required) `dependencies` + `C` + `seaborn`, but not `other`
- `conda install -f environment.yaml[all]` would install (required) `dependencies` + all optional depedencies

Not sure what would be the best syntax for the installation command, this is just an example.
Also, the `all` keyword might be also implicit.

giamic:
I agree with this feature request, it would be great if it was possible!

RogerAylagas:
This feature would be very useful! Any update on its progress?

travishathaway:
Hi @HealthyPear,

It sounds like you may have come up with a new feature that others may find useful! 😎 I personally think this could provide a nice way to extend the `environment.yml` file, but I can't say for sure when the primary development team could prioritize this.

Therefore, I would encourage you (or others) to fork the conda repository, come up with a working prototype, and to submit a pull request. We would be happy to help refine this and bring in other community opinions on the matter. 

Thanks for filing this feature request. If we decide to move on this ourselves, we will update you here.

jeskowagner:
I saw @astaric's work on this in #12357 and would love to see this feature merged. One use case is to specify whether CPU or GPU versions of packages should be installed (e.g. `conda install -f environment.yaml[GPU]`).

goyalyashpal:
This would be useful to me too, for specifying packages which is not a dependency in any way, but still stuff which maybe required in the environment.

Packages like: 
[...]

PR #12357, opened 2023-02-15
conda_env: add optional dependencies

astaric:
### Description

Prototype for optional dependencies in environment.yml file as requested in https://github.com/conda/conda/issues/12098.

Adds additional `optional_dependencies` section to environment.yml. Once can list several groups of dependencies there which do not get installed by default. Each group of dependencies can be selected with `--extras "group_name"` flag to `conda env create` / `conda env update` commands. Flag `--all-extras` enables installation of all optional dependencies.

This implementation uses additional command line flags for enabling the optional dependencies (flags have same names as corresponding flags in `poetry` package manager) instead of `-f environment.yml[group1,group2]` syntax described in issue . Optional dependencies have to be specified explicitly (listing all the packages), current implementation does not support referencing other optional groups.

### Checklist - did you ...


conda_env/env.py
```diff
@@ -177,11 +177,13 @@ def parse(self):
         if not self.raw:
             return
 
+        self.clear()
```
astaric, line 180: Before these two changes in parse, if pip "section" was specified multiple times in the environment file, all but the last occurrence were silently ignored. Since there were no warnings, I considered it a bug and changed the behaviour. This change now merges dependencies form all pip "sections" into the "pip" key in the dictionary.

If the old behaviour was intentional, I can revert this change, add a warning if config has multiple pip sections and move the merging logic to `add` or `enable_optional_dependencies` function.

beeankha:
@conda-bot check

XuehaiPan:
gentle bump for this

PR #12360, opened 2023-02-15
stub cache-info command

dholth:
### Description

This is intended to be used by Navigator, to decouple it from our cache filename and state format decisions.

With pluggy, this could also be defined outside of the conda source tree, and located through setuptools entry_points.txt . See https://docs.conda.io/projects/conda/en/latest/dev-guide/plugins/index.html#basic-hook-and-entry-point-examples for an example of an out-of-tree plugin.

(XXX update this branch so that even the in-tree plugin is loaded through entry points)

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -6,7 +6,7 @@ import functools
 
 import pluggy
 
-from . import solvers, virtual_packages
+from . import solvers, virtual_packages, subcommands
 from .hookspec import CondaSpecs, spec_name
 from ..auxlib.ish import dals
 from ..base.context import context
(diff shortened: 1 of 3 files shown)


jezdez:
I'm wondering if we could work on this as a general purpose cache subcommand to handle various operations on conda's cache?
- clearing cache (are there different levels of cache users might be required to cache independently?)
- warming the repodata cache (vs JLAP)
- print cache info
- ??

jezdez:
This might be useful to do as a new option for the conda info subcommand, e.g. `conda info --cache`. Closing in the meantime.

Issue #11563, opened 2022-06-21
Request to add Cisco Umbrella Root CA to signer certificate store


naatje80:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project (see the above section)

### What is the idea?

Our company recently enabled "SSL/TLS/HTTPS inspection" for the Cisco Umbrella. This is a feature of there "Intelligent Proxy" service, which is part of Cisco Umbrella. By enabling this, all HTTPS traffic is decrypted for inspection and forwarded to the receiver. Unfortunately, the HTTPS traffic is now signed by the Cisco Umbrella signer corticate, which is not part of the signer certificate database of the conda package manager. A temp. workaround is to disable ssl_verification, but I would like to know if it would be possible include the ROOT CA of Cisco Umbrella to the certificate database. I know that there are also solutions to create your own keystore and use this keystore for verification instead. However, this would require manual work to keep this keystore up-to-date.

### Why is this needed

Automatic verification of HTTPS traffic signed by Cisco Umbrella Intelligent Proxy service

### What should happen?

It would be nice if the Cisco Umbrally ROOT CA, which is available on the following page, https://www.cisco.com/security/pki/, could be added to the signer certificate keystore.

sandersn80:
Any thoughts regarding this topic?

sandersn80:
Please keep this topic open, because it has not been fixed. Could someone please comment on this topic?

dholth:
The best way to solve this would be to use the system's certificate store, which should have your corporate trust, which is something others in the Python http ecosystem have worked on.

dholth:
https://github.com/conda/conda/pull/12361 is a proof-of-concept that uses `truststore` to get certificates. Currently it borrows from pip's implementation, and pip must also be installed.

travishathaway:
Look into OpenSSLv3's ability to read from the Windows cert store.

dholth:
Should be fixed via #13075

Try setting `ssl_verify: truststore` on conda 23.9, available soon.

PR #12361, opened 2023-02-15
proof of concept truststore ssl context

dholth:
### Description

Help users with certain SSL configurations. Crib session from pip. Requires truststore, which appears to require Python 3.10+

Fix https://github.com/conda/conda/issues/11563 and similar SSL-intercept proxy issues

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -29,7 +29,10 @@ from ...common.url import (
 )
 from ...exceptions import ProxyError
 
-from pip._internal.network.session import HTTPAdapter
+try:
+    from pip._internal.network.session import HTTPAdapter
+except ImportError:
+    from . import HTTPAdapter
 
 log = getLogger(__name__)
 RETRIES = 3
@@ -48,6 +51,8 @@ CONDA_SESSION_SCHEMES = frozenset(
 
 def _ssl_context():
     try:
+        # must use this one, for the feature to work
+        from pip._internal.network.session import HTTPAdapter  # noqa
         import ssl
         import truststore
 
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -4,8 +4,16 @@
 from logging import getLogger
 from threading import local
 
-from . import (AuthBase, BaseAdapter, HTTPAdapter, Session, _basic_auth_str,
-               extract_cookies_to_jar, get_auth_from_url, get_netrc_auth, Retry)
+from . import (
+    AuthBase,
+    BaseAdapter,
+    Session,
+    _basic_auth_str,
+    extract_cookies_to_jar,
+    get_auth_from_url,
+    get_netrc_auth,
+    Retry,
+)
 from .adapters.ftp import FTPAdapter
 from .adapters.localfs import LocalFSAdapter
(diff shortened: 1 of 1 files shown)


conda/gateways/connection/session.py
```diff
@@ -4,38 +4,67 @@
...
+)
 
-class EnforceUnusedAdapter(BaseAdapter):
 
+def _ssl_context():
+    try:
+        import ssl
+        import truststore
+
+        return truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
```
dholth, line 54: Non-reformatting change 1: get ssl context from truststore

conda/gateways/connection/session.py
```diff
@@ -61,14 +90,20 @@ def __call__(cls):
...
         self.auth = CondaHttpAuth()  # TODO: should this just be for certain protocol adapters?
 
         self.proxies.update(context.proxy_servers)
 
+        ssl_context = _ssl_context()
+        self.ssl_context = ssl_context
+        if ssl_context:
+            log.info("Have truststore ssl_context")
+        else:
+            log.info("Lack truststore ssl_context")
```
dholth, line 105: non-reformatting change 2:  make ssl_context

conda/gateways/connection/session.py
```diff
@@ -78,19 +113,21 @@ def __init__(self):
...
-                          status_forcelist=[413, 429, 500, 503],
-                          raise_on_status=False)
-            http_adapter = HTTPAdapter(max_retries=retry)
+            retry = Retry(
+                total=context.remote_max_retries,
+                backoff_factor=context.remote_backoff_factor,
+                status_forcelist=[413, 429, 500, 503],
+                raise_on_status=False,
+            )
+            http_adapter = HTTPAdapter(max_retries=retry, ssl_context=self.ssl_context)
```
dholth, line 122: 3. pip's HTTPAdapter (import changed at top of file) uses a mixin class to add the ssl_context constructor argument, and pass it down the stack to where it needs to go. We should have our own implementation of "specify the ssl_context" instead of using pip's.

conda/gateways/connection/session.py
```diff
@@ -4,38 +4,72 @@
...
+from ...common.url import (
+    add_username_and_password,
+    get_proxy_username_and_pass,
+    split_anaconda_token,
+    urlparse,
+)
 from ...exceptions import ProxyError
 
+try:
+    from pip._internal.network.session import HTTPAdapter
```
jezdez, line 33: Import from pip like this will be a problem since pip's API isn't supported at all, that's why the module is named `_internal`. We can't rely on it in conda sadly, especially in a part that is security sensitive like this feature.

dholth: Conda does quite a few "try to import from requests / fall back to pip's vendored _internal version". However the main problem with this PR is that we should re-implement (or vendor?) the pip `_internal` classes used.

The classes we import from pip take care of adding a ssl context argument and propagating it all the way down to the lower-level urllib3 layer.

sethmlarson:
Since https://github.com/conda/conda/pull/13075 just got merged I believe this PR can be closed?

jezdez:
Yep!

PR #12378, opened 2023-02-21
Add info about `conda remove -n <env-name> --all`  in conda CLI help

### LINKED ISSUE #12358: Document the opposite of `conda create`
### Checklist

### What is the idea?

A new conda environment is created with `conda create`. I'm not sure what the correct method is to remove a conda environment.

### Why is this needed?

Some Internet searches show that the correct way to do it is `conda env remove -n [environment]`, however `conda --help` lists the `env` command as legacy. Furthermore, this command is listed as being "from other packages", which would indicate that it's not part of conda itself.

### What should happen?

There should be a command to remove an environment, or at least the current command should be documented and discoverable.

### Additional Context

There was discussion on this in #723 but this didn't seem to have a satisfying resolution.

ForgottenProgramme:
### Description

Resolves #12358 

`conda remove -n <env-name> --all` command can be used to delete environments. This is not mentioned in the conda CLI help (`conda --help`). However it is mentioned in `conda remove` CLI help (`conda remove --help`). 
Because the main `conda` CLI help only mentions that `remove` command is used to remove packages from an environment, it is unintuitive that it could also be used to remove the environment itself. 
Moreover the CLI help doesn't mention any other command to remove environments (eg. `conda env remove`). 

Whether `conda env remove` (and other `conda env` commands) should be listed on CLI help is a part of the [ongoing discussion about collapsing `conda env` into `conda`](https://github.com/conda/conda/issues/11633).
While that discussion continues in the background, I feel we should be able to make small reversible changes that improve user experience immediately. 

In this PR I added to the CLI help message that `remove` command will delete environments when used with the `--all` flag.

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -1058,7 +1058,10 @@ def configure_parser_package(sub_parsers):
 
 
 def configure_parser_remove(sub_parsers, aliases):
-    help_ = "Remove a list of packages from a specified conda environment."
+    help_ = (
+        "Remove a list of packages from a specified conda environment."
+        + " Use `--all` flag to remove all packages and the environment itself."
+    )
     descr = dals(
         f"""
         {help_}
@@ -1081,6 +1084,10 @@ def configure_parser_remove(sub_parsers, aliases):
         Remove a list of packages from an environemnt 'myenv'::
 
             conda remove -n myenv scipy curl wheel
+
+        Remove all packages from environment `myenv` and the environment itself::
+
+            conda remove -n myenv --all
 
         """
     )
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -1059,8 +1059,8 @@ def configure_parser_package(sub_parsers):
 
 def configure_parser_remove(sub_parsers, aliases):
     help_ = (
-        "Remove a list of packages from a specified conda environment."
-        + " Use `--all` flag to remove all packages and the environment itself."
+        "Remove a list of packages from a specified conda environment. "
+        "Use `--all` flag to remove all packages and the environment itself."
     )
     descr = dals(
         f"""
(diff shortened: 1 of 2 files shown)


travishathaway:
This pull request is looking good so far. I just have a couple suggestions that could make it even better.

1. Can we also update this line: https://github.com/conda/conda/blob/main/conda/cli/conda_argparse.py#L1061 to say that the command can be used for remove environments?
2. Can we add an example show how to remove an environment with this command? (https://github.com/conda/conda/blob/main/conda/cli/conda_argparse.py#L1073)

travishathaway (dismissed):

beeankha (approved):

conda/cli/conda_argparse.py
```diff
@@ -1058,7 +1058,10 @@ def configure_parser_package(sub_parsers):
 
 
 def configure_parser_remove(sub_parsers, aliases):
-    help_ = "Remove a list of packages from a specified conda environment."
+    help_ = (
+        "Remove a list of packages from a specified conda environment."
+        + " Use `--all` flag to remove all packages and the environment itself."
```
kenodegard, line 1063: [Adjacent strings will auto concatenate](https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation):

```suggestion
        "Remove a list of packages from a specified conda environment. "
        "Use `--all` flag to remove all packages and the environment itself."
```

news/12378-add-info-about-conda-remove-in-conda-CLI-help
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Added to `conda` CLI help that `conda remove -n <myenv> --all` can be used to delete environments
+#12378
```
kenodegard, line 4:
```suggestion
* Added to `conda` CLI help that `conda remove -n <myenv> --all` can be used to delete environments. (#12378)
```

kenodegard (changes requested):

kenodegard (approved):

PR #12411, opened 2023-02-28
Move durations into tests directory.

jezdez:
### Description

Follow up to #12297 

### Checklist - did you ...


.github/workflows/tests.yml
```diff
@@ -156,7 +156,7 @@ jobs:
           name: test-results-${{ github.sha }}-${{ runner.os }}-${{ matrix.default-channel }}-${{ matrix.python-version }}-${{ matrix.conda-subdir }}-${{ matrix.test-type }}-${{ matrix.test-group }}
           path: |
             .coverage
-            .test_durations_${OS}
+            tests\durations\${OS}.json
```
dholth, line 159: Do we have to use \ here or does GHA normalize path separators for this kind of thing

dholth (approved):

.github/workflows/tests.yml
jezdez: Not sure, it's probably safer to just use backslashes, despite the eye sore

dholth: Something about YAML complicated "when is \ an escape character" parsing

PR #12413, opened 2023-02-28
use 2 unit test splits for osx, to reduce concurrency

dholth:
### Description

Two unit-tests splits for OSX. Will save us ~6 minutes of setup time since we avoid running the third split. Since the integration tests are much slower, a bigger unit tests run won't hold anything up.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -283,35 +283,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - default-channel: 'conda-forge'
-            python-version: '3.9'
-            test-type: 'unit'
-            test-group: '1'
-          - default-channel: 'conda-forge'
-            python-version: '3.9'
-            test-type: 'unit'
-            test-group: '2'
-          - default-channel: 'conda-forge'
-            python-version: '3.9'
-            test-type: 'unit'
+        default-channel: ['defaults', 'conda-forge']
+        python-version: ['3.9']
+        test-type: ['unit', 'integration']
+        test-group: ['1', '2', '3']
+        # fewer test splits for quicker unit tests, keep 3 splits for integration
+        test-splits: [2, 3]
+        exclude:
+          - test-type: unit
             test-group: '3'
-          - default-channel: 'defaults'
-            python-version: '3.9'
-            test-type: 'integration'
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -283,35 +283,28 @@ jobs:
...
             test-group: '3'
+          - test-type: unit
+            test-splits: 3
+          - test-type: integration
+            test-splits: 2
+          # odd choice kept for parity with previous commit
+          - test-type: integration
+            default-channel: 'conda-forge'
+          - test-type: unit
+            default-channel: 'defaults'
```
kenodegard, line 303: Isn't this oddity actually a bug?

tests/test_resolve.py
```diff
@@ -127,7 +127,7 @@ def test_anaconda_nomkl(self):
 
 def test_generate_eq_1():
     # avoid cache from other tests which may have different result
-    r._reduced_index_cache = {}
+    _index, r = get_index_r_1()
```
kenodegard, line 130: Doesn't belong in this PR

dholth: If I leave this out of this PR, then the tests fail.

kenodegard: that's odd, shouldn't this be failing on `main` too then?

dholth: It is re-using a module-scope cached `r`, and something (no idea what) in another test sets it up for success or failure before we run here. So these new splits have improved the quality of our test suite.

The `least_duration` algorithm isn't as aggressive as randomizing test order, but it can split each file's tests into multiple runners compared to the naive algorithm which runs the 1st third of the tests in one batch, and the 2nd third in another batch and so on.

kenodegard: ah I see, carry on then

dholth: auto-merge enabled

.github/workflows/tests.yml
dholth: If we must change this to something besides "only run on defaults" or "randomly pick which channel to run against", let's wait until we have enough runners.

dholth:
I don't have anything else to add to this PR, ready to merge.

kenodegard (approved):

PR #12414, opened 2023-03-01
Mark `hashsum_file` for deprecation

kenodegard:
### Description

Found two implementations of file hash digests. The original implementation (that was added back for conda-build):

https://github.com/conda/conda/blob/584b1603923d495ce2d2e7cdcec05eb0d242bdd9/conda/utils.py#L226-L235

and the newer implementation:

https://github.com/conda/conda/blob/584b1603923d495ce2d2e7cdcec05eb0d242bdd9/conda/gateways/disk/read.py#L61-L69

Deprecating the `conda.utils.hashsum_file` implementation in favor of `conda.gateways.disk.read.compute_sum`.

### Checklist - did you ...


conda/gateways/disk/read.py
```diff
@@ -58,23 +61,30 @@ def yield_lines(path):
...
 
-def _digest_path(algo, path):
-    if not isfile(path):
+@deprecated("23.9", "24.3", addendum="Use `conda.gateways.disk.read.compute_sum` instead.")
+def _digest_path(algo: Literal["md5", "sha256"], path: str | os.PathLike) -> str:
+    return compute_sum(path, algo)
+
+
+# FUTURE: Python 3.11+, replace with hashlib.file_digest
+def compute_sum(path: str | os.PathLike, algo: Literal["md5", "sha256"]) -> str:
```
jezdez, line 70: I'm not sure if we need to check for the file existence here anymore, we could already mimic the `hashlib.filedigest` API here to make the deprecation easier long-term. IOW I'd rather have the core of this be a backport of `file_digest` that we remove at some point.

```suggestion
# FUTURE: Python 3.11+, replace with hashlib.file_digest
def file_digest(fileobj, digest: Literal["md5", "sha256"], /) -> str:
```

conda/utils.py
```diff
@@ -218,21 +220,15 @@ def human_bytes(n):
 urlpath = url_path = path_to_url
 
 
-def md5_file(path):  # pragma: no cover
+@deprecated("23.9", "24.3", addendum="Use `conda.gateways.disk.read.compute_md5sum` instead.")
```
jezdez, line 223: I don't think we should suggest using a md5 specific tool if `file_digest(open(path, "b"), "md5")` also works.

kenodegard: `compute_md5(path)` == `file_digest(open(path, "b"), "md5").hexdigest()`

dholth:
`conda_package_handling.utils` has a `checksums` and a `checksum` function I worked on, that uses a tuned `2**18` byte buffer size and can compute multiple checksums at once in threads.

conda/utils.py
kenodegard: just reviewed all instances of these helper functions and replacing everything with `file_digest().hexdigest()` adds quite a mouthful and decreases overall readability so I'm thinking we probably want to keep `compute_digest`, `compute_md5`, and `compute_sh256` helper functions

jezdez: What rubs me the wrong way is us not deprecating enough here, I wouldn't mind if we'd keep `compute_digest` to wrap the `open` and call `hexdigest()`, but not have `compute_md5` and `compute_sha256`, since it's literally just `compute_digest(path, "md5")` and `compute_digest(path, "sha256")`, which isn't decreasing readability.

jezdez:
@dholth Taking a look at https://github.com/conda/conda-package-handling/blob/323530b96a51b4ee52958372d3502f103cb52341/src/conda_package_handling/utils.py#L447-L481 that seems to be useful, not sure if this wouldn't be considered a new feature though, probably out of scope of this particular ticket?

conda/utils.py
kenodegard: Gotcha, I can get behind dropping the algorithm-specific implementations. 👍🏻

jezdez (approved):

PR #12424, opened 2023-03-01
miscellaneous refactors originating from features/repodata-jlap

dholth:
### Description

A few simple refactors.

### Checklist - did you ...


kenodegard (dismissed):
> should only ever be needed during unit tests

👍🏻

tests/test_resolve.py
```diff
@@ -55,6 +51,7 @@ def assert_have_mkl(self, precs, names):
     #     self.assertEqual(r.explicit(['pycosat 0.6.0 notarealbuildstring']), None)
 
     def test_empty(self):
+        _, r = get_index_r_1()
```
dholth, line 54: this only takes 0.14 seconds on my machine

.github/workflows/tests.yml
```diff
@@ -259,6 +259,7 @@ jobs:
           -e TEST_GROUP
           ghcr.io/conda/conda-ci:main-linux-python${{ matrix.python-version }}${{ matrix.default-channel == 'conda-forge' && '-conda-forge' || '' }}
           bash -c "source /opt/conda/etc/profile.d/conda.sh; \
+                   sudo /opt/conda/bin/conda install --quiet -y -c defaults --file tests/requirements.txt; \
                    pytest --cov=conda -k test_DepsModifier_contract"
```
jaimergp, line 263: This could use `unit.sh`, which could take some CLI args for pytest (default "-k "not integration").

tests/conftest.py
```diff
@@ -37,7 +37,7 @@ def pre_link_messages_package():
 def clear_cache():
     from conda.core.subdir_data import SubdirData
 
-    SubdirData._cache_.clear()
+    SubdirData.clear_cached_local_channel_data(exclude_file=False)
```
dholth, line 40: This is more OO but a really ugly method name.

jezdez:
> > I don't think we should remove test_resolve until we've removed the classic solver
> 
> The failing tests in test_resolve block all work on conda, but [`main`/tests/test_solvers.py](https://github.com/conda/conda/blob/main/tests/test_solvers.py?rgh-link-date=2023-03-02T15%3A36%3A52Z) tests the classic Solver class with code, in helpers, that was copied almost verbatim from test_resolve

Ah, thanks for the clarification, I was looking at https://github.com/conda/conda/blob/main/tests/plugins/test_solvers.py 🤦  Carry on!

dholth:
@jezdez OK to merge?

conda/deprecations.py
```diff
@@ -6,7 +6,33 @@
 from types import ModuleType
 import warnings
 
-from packaging.version import parse, Version
+try:
```
jezdez, line 9: I don't think we need this, it rubs me the wrong way that we implement a custom version parser just for `deprecations`.

The `packaging` dependency exists and if not, then we need to find the reason why it's not working and not paper over it.

jezdez (changes requested):

conda/deprecations.py
dholth: I do like this, I think the packaging import and full Version parsing is overkill for deprecations, and I'm concerned that we have too many dependencies. However I did remove it in the other PR, and if tests pass we can cherry pick that commit here.

tests/http_test_server.py
```diff
@@ -18,6 +18,10 @@ def run_test_server(directory: str) -> http.server.ThreadingHTTPServer:
     """
 
     class DualStackServer(http.server.ThreadingHTTPServer):
+        daemon_threads = False  # These are per-request threads
+        allow_reuse_address = True  # Good for tests
+        request_queue_size = 64  # Should be more than the number of test packages
```
dholth, line 23: I'm guessing request_queue_size < number of conda request threads is the real problem. It doesn't hurt to have a long queue.

allow_reuse_address is unchanged from the superclass default.

If daemon_threads keeps the test suite alive for noticeably longer (seems unlikely but who knows) we can take it out.

conda/deprecations.py
kenodegard: `packaging` is exclusively used here today but don't we expect to use it more as `distutils` is deprecated?

dholth: I've split it out into a different PR. Maybe we do, or maybe deprecation is a special case since it has to be accessed so early in conda's initialization cycle.

dholth:
I think the test server fix takes this one over the top (if the tests pass).

jezdez (approved):

PR #12428, opened 2023-03-02
Reformat news snippet filename and content

beeankha:
Updating a news snippet to reflect the PR number in the filename and the related issue number in the file content. 

We probably should standardize and document news snippet protocol in our contribution guide!


kenodegard:
Need to remove the old news snippet

kenodegard (approved):

PR #12431, opened 2023-03-02
implement minimal Version class for deprecations

dholth:
### Description

Dependencies? Can be a problem.

I was having a problem with missing `packaging` during some conda testing. Since deprecations needs a fast, lightweight, and featureless `Version` this implements that.

### Checklist - did you ...

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -2,13 +2,45 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
+import warnings
 from functools import wraps
 from types import ModuleType
-import warnings
-
-from packaging.version import parse, Version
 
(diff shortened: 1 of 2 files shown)


jezdez:
I don't think we should do this for a few reasons:

- this doesn't solve the problem you're stating, which is "a core dependency isn't available" but tries to work around it in conflict with prior design decisions
- should have been a ticket ~"packaging dependency missing during tests" first before you spend time on it
- it's adding a custom version parser without respecting the PEP standards we agreed to use more and adds technical debt
- makes wrong assumptions about what deprecations needs (stability and correctness) and about the performance impact

dholth:
`packaging` wants to be a core dependency but is only used here. This was broken out of another PR, done out of necessity due to broken tests.

Issue #12436, opened 2023-03-03
Drop Python 3.7 support


kenodegard:
### What is the idea?

Drop Python 3.7 support.

### Why is this needed?

Neither `pkgs/main` (a.k.a., `defaults`) nor `conda-forge` package for Python 3.7 anymore.

### What should happen?

Update dependencies and remove Python 3.7 from the testing matrix.

jezdez:
Closed by #12439.

PR #12439, opened 2023-03-03
Removing Python 3.7 from matrix.  Also, making exclude matrix on linu…

ryanskeith:
### Description

Addresses issue #12436 to drop Python 3.7 Support.

Also, syncs what is used in tests matrix in windows and linux section (which may have been on purpose).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -177,7 +177,7 @@ jobs:
           - default-channel: 'conda-forge'
             python-version: '3.8'
           - default-channel: 'conda-forge'
-            python-version: '3.10'
+            python-version: '3.9'
     env:
       OS: Linux
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -171,17 +169,15 @@ jobs:
...
         test-group: ['1', '2', '3']
         exclude:
           # exclude all but one Python version for conda-forge
-          - default-channel: 'conda-forge'
-            python-version: '3.7'
           - default-channel: 'conda-forge'
             python-version: '3.8'
           - default-channel: 'conda-forge'
-            python-version: '3.9'
+            python-version: '3.10'
```
kenodegard, line 180: I think this should stay 3.9:

```suggestion
            python-version: '3.9'
```

kenodegard (changes requested):

jezdez (approved):

kenodegard (approved):

jezdez:
Seems like #12446 might fix this.

jezdez (approved):

PR #12438, opened 2023-03-03
Disable user site-packages during de[activation]

jjhelmus:
### Description

Disable the user site-packages directory when retrieving the shell code for activation, deactivation and reactivation from CONDA_EXE. The output from these calls are evaluated by the shell and any extraneous output introduced by the user site-packages directory, such as by usercustomize.py or .pth files, will be evaluated.

An alternative would be to use Python in isolated mode (`-I`) to run the CONDA_EXE script. This would also ignore the PYTHON* environment variables and unsafe paths.

Note that this only adds this to Unix style shells.  Windows specific shells (cmd, powershell) may still access the user site-packages directory during activation.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [x] Add / update outdated documentation?

closes #12444

PR #12446, opened 2023-03-06
Init conda after installing requirements in tests

jjhelmus:
### Description

When setting up to run the unit tests initialize conda only after the requirements are installed since installing requirements may itself install conda undoing any initialization.

### Checklist - did you ...

### DIFF
--- a/dev/linux/unit.sh
+++ b/dev/linux/unit.sh
@@ -8,9 +8,9 @@ git config --global --add safe.directory /opt/conda-src
 TEST_SPLITS="${TEST_SPLITS:-1}"
 TEST_GROUP="${TEST_GROUP:-1}"
 
-eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
 # make sure all test requirements are installed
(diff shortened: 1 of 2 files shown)


jjhelmus:
The ordering of these operations is the cause of the failure in #12438.

jezdez (approved):

jezdez:
Nice find!

dholth:
Will this make Windows use the github actions cache of packages more effectively?

dev/linux/unit.sh
```diff
@@ -8,9 +8,9 @@ git config --global --add safe.directory /opt/conda-src
 TEST_SPLITS="${TEST_SPLITS:-1}"
 TEST_GROUP="${TEST_GROUP:-1}"
 
-eval "$(sudo /opt/conda/bin/python -m conda init --dev bash)"
 # make sure all test requirements are installed
 sudo /opt/conda/bin/conda install --quiet -y -c defaults --file tests/requirements.txt
```
jaimergp, line 12: This line shouldn't be adding `defaults`. I'll open an issue and a PR.

PR #12460, opened 2023-03-07
Handle ImportError gracefully while loading plugins.

### LINKED ISSUE #12455: Conda should not crash on plugin load "ImportError"
### Checklist

### What happened?

I have the `conda-libmamba-solver` plugin installed but something went wrong with my conda environment. Apparently libarchive 19 is installed instead of 13? Could this have something to do with conda no longer requiring libarchive by itself?

Since `conda` doesn't load in this state, the remedy is to go into the `site-packages` directory and manually delete the `conda_libmamb_solver*` directories associated with the plugin.

Instead, `conda` should warn when a plugin produces an `ImportError`, then proceed without the plugin.

```
% CONDA_DEBUG=1 conda search -c dholth conda
  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/conda/plugins/manager.py", line 61, in load_setuptools_entrypoints
    return super().load_setuptools_entrypoints(*args, **kwargs)
  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/pluggy/_manager.py", line 287, in load_setuptools_entrypoints
    plugin = ep.load()

...

  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/conda_libmamba_solver/__init__.py", line 7, in <module>
    from .solver import LibMambaSolver
  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/conda_libmamba_solver/solver.py", line 19, in <module>
    import libmambapy as api
  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/libmambapy/__init__.py", line 7, in <module>
    raise e
  File "/Users/dholth/miniconda3/lib/python3.10/site-packages/libmambapy/__init__.py", line 4, in <module>
    from libmambapy.bindings import *  # noqa: F401,F403
ImportError: dlopen(/Users/dholth/miniconda3/lib/python3.10/site-packages/libmambapy/bindings.cpython-310-darwin.so, 0x0002): Library not loaded: '@rpath/libarchive.13.dylib'
  Referenced from: '/Users/dholth/miniconda3/lib/libmamba.2.0.0.dylib'
  Reason: tried: '/Users/dholth/miniconda3/lib/libarchive.13.dylib' (no such file), '/Users/dholth/miniconda3/lib/python3.10/site-packages/libmambapy/../../../libarchive.13.dylib' (no such file), '/Users/dholth/miniconda3/lib/python3.10/site-packages/libmambapy/../../../libarchive.13.dylib' (no such file), '/Users/dholth/miniconda3/bin/../lib/libarchive.13.dylib' (no such file), '/Users/dholth/miniconda3/bin/../lib/libarchive.13.dylib' (no such file), '/usr/local/lib/libarchive.13.dylib' (no such file), '/usr/lib/libarchive.13.dylib' (no such file)
```

jezdez:
### Description

Resolves #12455.

### Checklist - did you ...


tests/plugins/test_manager.py
```diff
@@ -85,13 +85,24 @@ def test_load_plugins_error(plugin_manager, mocker):
...
+
+def test_load_setuptools_entrypoints_importerror(caplog, plugin_manager, mocker):
+    with mocker.patch(
+        "pluggy._manager.importlib_metadata.distributions",
+        side_effect=ImportError("load_setuptools_entrypoints ImportError"),
+    ):
+        caplog.set_level(logging.INFO, logger="conda.plugins.manager")
+        plugin_manager.load_setuptools_entrypoints("conda")
+        assert plugin_manager.get_plugins() == set()
+        assert "load_setuptools_entrypoints ImportError" in caplog.text
```
dholth, line 108: I bet you could add an `ImportError` in a different way. See https://github.com/conda/conda/blob/main/tests/gateways/test_repodata_gateway.py#L113 or https://github.com/conda/conda/blob/196148c25db8e5aa2a78124ce2232818d6316fc3/tests/test_deprecations.py#L170

dholth:
The other thing we might want to do is audit our plugins for too-eager imports. Ideally scanning the entry point would avoid importing most of the plugin implementation.

jezdez:
> The other thing we might want to do is audit our plugins for too-eager imports. Ideally scanning the entry point would avoid importing most of the plugin implementation.

I'm not following, scanning the entry point?

dholth:
When a plugin is imported, just to be added to the list of `conda x`, it should not do very much work. Including importing things.

A plugin author might want to arrange a plugin with separate folders myplugin/cli and myplugin/implementation so that "import myplugin.cli" doesn't import whatever dependencies "myplugin.implementation" uses to do its thing. Inside "myplugin.cli:do_the_plugin()", import "myplugin.implementation".

Then unrelated commands remain fast.

Firefox used to say "these extensions are slowing down your browser" and tell you how long each one took to load. Then you could decide whether to continue to use those extensions. It also has a safe mode.

jezdez:
I see, lazy loading plugins, but I don't see how this works with pluggy as it uses importlib as the central system to load entrypoints based plugins which loads plugins with the import system. I'm not sure how to achieve this given these constraints. I'll take a look at it tomorrow.

dholth:
It depends on the type of the plugin.

For conda_subcommands it is simple. Image some subcommand uses sqlalchemy.

A. someplugin.py
```python
import sqlalchemy
def custom_command():
    sqlalchemy.do_a_thing()

@plugins.hookimpl
def conda_subcommands(self):
    yield CondaSubcommand(
        name="custom",
        summary="test custom command",
        action=self.custom_command,
    )
```

B. anotherplugin.py
```python
def custom_command():
    import sqlalchemy
    sqlalchemy.do_a_thing()

@plugins.hookimpl
def conda_subcommands(self):
    yield CondaSubcommand(
        name="custom",
        summary="test custom command",
        action=self.custom_command,
    )
```

The second plugin is faster.

Authors would be more likely to do this accidentally, by having an `__init__.py` that imports a bunch of things sitting between your entry points hook and the `@plugins.hookimpl` call.

jezdez:
> It depends on the type of the plugin.
> 
> For conda_subcommands it is simple. Image some subcommand uses sqlalchemy.
> 
> A. someplugin.py
> 
> ```python
> import sqlalchemy
> def custom_command():
>     sqlalchemy.do_a_thing()
> 
> @plugins.hookimpl
> def conda_subcommands(self):
>     yield CondaSubcommand(
>         name="custom",
>         summary="test custom command",
>         action=self.custom_command,
>     )
> ```
> 
>  
> B. anotherplugin.py
> 
> ```python
> def custom_command():
>     import sqlalchemy
>     sqlalchemy.do_a_thing()
> 
> @plugins.hookimpl
> def conda_subcommands(self):
>     yield CondaSubcommand(
>         name="custom",
>         summary="test custom command",
>         action=self.custom_command,
>     )
> ```
> 
>  
> The second plugin is faster.
> 
> Authors would be more likely to do this accidentally, by having an `__init__.py` that imports a bunch of things sitting between your entry points hook and the `@plugins.hookimpl` call.

Yeah, I get that, import-time side effects have a downside. For subcommands we could allow to optionally pass a dotted import path as the action parameter and lazy-import it when we need it, to move the moment of loading to later in the stack.

conda/plugins/manager.py
```diff
@@ -59,10 +60,10 @@ def load_setuptools_entrypoints(self, *args, **kwargs) -> int:
         """
         try:
             return super().load_setuptools_entrypoints(*args, **kwargs)
+        except ImportError as err:
+            warnings.warn(f"WARNING: Error while loading conda plugin: {err}")
         except Exception as err:
-            raise PluginError(
-                f"Error while loading conda plugins from entrypoints: {err}"
-            )
+            raise PluginError(f"Error while loading conda plugin: {err}")
```
dholth, line 66: Does this halt conda?

dholth: Is the plugin hook name or dotted-name available in *args, **kwargs? Is it in the exception? Could we format that string into the error message?

dholth (approved):

conda/plugins/manager.py
jezdez: > Is the plugin hook name or dotted-name available in *args, **kwargs? Is it in the exception? Could we format that string into the error message?

No, the name of the entrypoint we're loading is just `conda` and the values passed is just that. We might be able to do more, but would implement more of the feature on our own than to rely on what is there in pluggy. I think it's a good idea since right now, it's hard for users to identify what's causing the loading error.

jezdez: > Does this halt conda?

Yeah, `PluginError` is a subclass of `CondaError` and will be gracefully handled like the other subclasses.

dholth: So if you have a plugin that raises another exception besides ImportError, you will still have to manually delete it, or use a different conda to update that environment before you can do anything with conda?

jezdez: Are you suggesting catching all exceptions when loading plugins from 3rd party packages and printing a warning instead?

dholth: Yes, I think that might be useful.

conda/plugins/manager.py
```diff
@@ -50,19 +54,38 @@ def load_plugins(self, *plugins) -> list[str]:
...
+        :return: The number of plugins loaded by this call.
         """
-        try:
-            return super().load_setuptools_entrypoints(*args, **kwargs)
-        except Exception as err:
-            raise PluginError(
-                f"Error while loading conda plugins from entrypoints: {err}"
-            )
+        count = 0
+        for dist in list(distributions()):
```
dholth, line 67: Is this the retconned "same thing setuptools did" equivalent function?

conda/plugins/manager.py
```diff
@@ -50,19 +54,38 @@ def load_plugins(self, *plugins) -> list[str]:
...
+                except Exception as err:
+                    # not using exc_info=True here since the CLI loggers are
+                    # set up after CLI initialization and argument parsing,
+                    # meaning that it comes too late to properly render
+                    # a traceback
+                    log.warning(f"Could not load conda plugin `{entry_point.name}`:\n\n{err}")
+                    continue
+                self.register(plugin, name=entry_point.name)
+                count += 1
+        return count
```
dholth, line 88: How do we get this line to run

news/12460-handle-plugins-importerror
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Handle Python import errors gracefully when loading
+  conda plugins from entrypoints. (#12460)
```
dholth, line 4: solid description

conda/plugins/manager.py
jezdez: Yep, importlib.metadata is the pkg_resources equivalent

conda/plugins/manager.py
jezdez: Not following, this just returns the count, there is no return statement in the previous code. This contains code from https://github.com/pytest-dev/pluggy/blob/c18d5da87f33bc5359c6eeccf4e85a194842d907/src/pluggy/_manager.py#L352-L377 BTW

dholth:
Did we fake a working plugin too as long as we're in here?

dholth (approved):
This should help.

jezdez:
> Did we fake a working plugin too as long as we're in here?

Nah, but it's a good idea to do that in a separate PR.

Issue #12445, opened 2023-03-06
Review pre-commit config


kenodegard:
### Summary

We've received complaints about the pre-commit process being slow. We should revisit all checks and review their necessity, value, and potential alternatives (if any).

```[tasklist]
### Checks to keep
- [ ] mixed-line-ending
- [ ] end-of-file-fixer
- [ ] trailing-whitespace
- [ ] check-toml
- [ ] check-yaml
- [ ] check-merge-conflict
- [ ] pyupgrade
- [ ] darker
- [ ] blacken-docs
- [ ] flake8 (https://github.com/charliermarsh/ruff)
- [ ] pylint (https://github.com/charliermarsh/ruff/issues/970)
- [ ] bandit
- [ ] insert-license
```

jezdez:
Here's the ruff roadmap: https://github.com/charliermarsh/ruff/issues/1992

dholth:
Before I was using pre-commit, I would often push a change, have all the tests pass, and have to re-run everything because some file had the wrong number of newlines at the end. This wastes a couple of hours. So pre-commit is helping me to enforce a style that no human reader, and no compiler would notice.

Then I set up pre-commit locally. I get the same error, but it takes less time to correct locally, and I'm more likely to be able to use the results of the slow tests to gate a merge. But it interrupts flow to commit, get the error, re-add the changed file, commit again, and push.

pre-commit is significantly slower than "git commit" by itself. With `ruff` that speed difference almost goes away. We might also be able to disable the slowest linter.

kenodegard:
Closing since we've replaced both flake8/black with ruff/ruff-format

PR #12466, opened 2023-03-08
Speedup pre-commit by factor 15

dbast:
### Description

This removes pylint and bandit from pre-commit, as both are slow and their feedback is anyway already ignored, due to running with exit-zero.

This locally reduces the pre-commit runtime from ~43sec to 2.9sec and thus makes it possible to run pre-commit in a loop during local development to constantly provide feedback / style the code.

This partially resolves https://github.com/conda/conda/issues/12445 (removing more hooks does not make anything significant faster.... the move to ruff for more speed can be done via a separate PR).

### Checklist - did you ...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,18 +49,6 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
-  - repo: https://github.com/PyCQA/pylint
-    rev: v2.16.4
-    hooks:
-      - id: pylint
-        args: [--exit-zero]
-  - repo: https://github.com/PyCQA/bandit
-    rev: 1.7.4
-    hooks:
-      - id: bandit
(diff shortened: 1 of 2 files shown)


dholth:
Thanks!

kenodegard (approved):
I assume these linters take so much longer to run because they're finding so many errors (that we then ignore). I agree with removing them from pre-commit, we can always add them back once we've gone through and resolved all of the errors raised by the respective tool.

dholth:
@kenodegard I think you will find unfortunately that the linters take the same amount of time, whether or not they find errors; but I haven't timed it. They are probably parsing the Python code (multiple times?) and passing it through dozens or hundreds of individual "look for a specific linter problem" functions. You can try it yourself with `pre-commit run --all` to take out the try-to-lint-only-changed-files variable.

PR #12474, opened 2023-03-10
Add improved repo/cache interface, polish jlap feature

dholth:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #12475, opened 2023-03-10
Add improved repo/cache interface, polish jlap feature (#12474)

dholth:
follow-up to features/repodata-jlap branch

### Checklist - did you ...


conda/gateways/repodata/__init__.py
```diff
@@ -663,6 +698,246 @@ def timeout(self):
...
+
+        :return: (repodata contents, state including cache headers)
+        """
+        parsed, state = self.whatever_subdir_data_used_to_do()
+        if isinstance(parsed, str):
+            return json.loads(parsed), state
+        else:
+            return parsed, state
+
+    def fetch_latest_str(self) -> tuple[str, RepodataState]:
```
dholth, line 743: It might be possible to get rid of the `_str` variant. SubdirData needs the json, and libmamba-solver only needs to be able to load it again from disk.

It is very awkward to change the functions below this, that have to download and return the index data, to either return str, parsed json or a Path.

tests/requirements.txt
jezdez, line 1: coverage>=7 and pytest-cov>=4 aren't available on defaults yet for the platforms we build multi-arch docker images yet, which was added in https://github.com/conda/conda/pull/11560

I've created package update requests with the Anaconda packaging team

jezdez (dismissed):
LGTM!

conda/gateways/repodata/__init__.py
```diff
@@ -660,6 +699,242 @@ def timeout(self):
...
+        Changes as we mutate self.repodata_fn.
+        """
+        return self.repo_interface_cls(
+            self.url_w_credentials,
+            self.repodata_fn,
+            cache_path_json=self.cache_path_json,
+            cache_path_state=self.cache_path_state,
+        )
+
+    def whatever_subdir_data_used_to_do(self) -> tuple[dict | str, RepodataState]:
```
dholth, line 803: @jezdez this function should probably be cleaned up

dholth:
@jezdez this branch renames the cache file from .state.json to .info.json, and it has some awkward method names.

dev/linux/setup.sh
```diff
@@ -4,7 +4,7 @@ set -o errtrace -o pipefail -o errexit
 
 apt-get update --fix-missing
 apt-get install -y --no-install-recommends \
-    tini wget build-essential bzip2 ca-certificates \
+    tini wget curl build-essential bzip2 ca-certificates \
```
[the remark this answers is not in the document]
dholth: https://github.com/conda/conda/actions/runs/4511518869/jobs/7943765244#step:7:1448

dholth:
I could use some advice cleaning up the `def whatever_subdir_data_used_to_do(self) -> tuple[dict | str, RepodataState]:` function, and possibly the new replacement for RepoInterface. With an eye towards making sure this really avoids duplicate work when passing the index data between the network stack and SubdirData.

tests/gateways/test_repodata_gateway.py
```diff
@@ -335,11 +351,82 @@ def test_cache_json(tmp_path: Path):
...
+    assert isinstance(state, RepodataState)
+
+
+def test_repodata_fetch_formats_2(
+    package_server: socket,
+    tmp_path: Path,
+):
+    """
+    Test that RepodataFetch can convert json to str, for coverage. Should we
+    have this feature?
```
dholth, line 415: I think we did remove this feature, this test is incomplete and should be removed.

conda/core/subdir_data.py
```diff
@@ -219,17 +198,26 @@ def __init__(
...
 
     @property
     def repo_cache(self) -> RepodataCache:
-        return RepodataCache(self.cache_path_base, self.repodata_fn)
+        return self.repo_fetch.repo_cache
+
+    @property
+    def repo_fetch(self) -> RepodataFetch:
+        """
+        Most modern "exactly what we need" interface.
```
travishathaway, line 213: I'm must admit, I am little confused by what this doc string is trying to communicate. Are you trying to say that this is the preferred way to instantiate a `RepodataFetch` object?

dholth: We started with `fetch_repodata_remote_request` (at the end of the file), then we moved through `_repo`, `repo_cache` and finally `repo_fetch`, all of which remain to support compatibility rules.

conda/core/subdir_data.py
```diff
@@ -460,37 +342,12 @@ def _read_local_repodata(self, state: RepodataState):
...
-            )
-            # taken care of by _process_raw_repodata():
-            assert self._internal_state is _internal_state
-            self._pickle_me()
-            return _internal_state
+        raw_repodata_str, state = self.repo_fetch.read_cache()
+        _internal_state = self._process_raw_repodata_str(raw_repodata_str, state)
+        # taken care of by _process_raw_repodata():
+        assert self._internal_state is _internal_state
+        self._pickle_me()
```
travishathaway, line 349: pickle_rick

conda/gateways/repodata/__init__.py
```diff
@@ -656,6 +710,251 @@ def timeout(self):
...
+
+    Provide a variety of formats since some ``RepoInterface`` have to
+    ``json.loads(...)`` anyway, and some clients don't need the Python data
+    structure at all.
+    """
+
+    cache_path_base: Path
+    channel: Channel
+    repodata_fn: str
+    url_w_subdir: str
```
travishathaway, line 726: Could we please replace `_w_` with `_with_`? I think that's what that means right?

At the very least, it would be nice to have that fully spelled out above with a comment.

conda/core/subdir_data.py
travishathaway: Okay, should probably mention that in the doc string 😁.

conda/gateways/repodata/__init__.py
```diff
@@ -656,6 +710,251 @@ def timeout(self):
...
+    url_w_credentials: str
+    repo_interface_cls: Any
```
travishathaway, line 728: You could use `type` here. That's ultimately what we will be passing in.

conda/core/subdir_data.py
```diff
@@ -460,37 +344,12 @@ def _read_local_repodata(self, state: RepodataState):
...
-        else:
-            _internal_state = self._process_raw_repodata_str(
-                raw_repodata_str, cache.state
-            )
-            # taken care of by _process_raw_repodata():
-            assert self._internal_state is _internal_state
-            self._pickle_me()
-            return _internal_state
+        raw_repodata_str, state = self.repo_fetch.read_cache()
+        _internal_state = self._process_raw_repodata_str(raw_repodata_str, state)
```
dholth, line 348: Not sure why the coverage doesn't reach here

conda/gateways/repodata/jlap/interface.py
```diff
@@ -34,45 +42,85 @@ def __init__(
...
+                    state[ETAG_KEY] = headers.get("etag")
+                    state[LAST_MODIFIED_KEY] = headers.get("last-modified")
+                    state[CACHE_CONTROL_KEY] = headers.get("cache-control")
 
-        # Indicate that subdir_data mustn't rewrite cache_path_json
-        raise RepodataOnDisk()
+                self._cache.state.update(state)
+
+            if temp_path.exists():
+                self._cache.replace(temp_path)
```
dholth, line 111: Need to find an elegant way to prevent the caller from also calling `cache.refresh()` writing the (small) state/info file again.

conda/gateways/repodata/__init__.py
```diff
@@ -656,6 +710,251 @@ def timeout(self):
...
+        """
+        return self.repo_interface_cls(
+            self.url_w_credentials,
+            repodata_fn=self.repodata_fn,
+            cache_path_json=self.cache_path_json,
+            cache_path_state=self.cache_path_state,
+            cache=self.repo_cache,
+        )
+
+    def fetch_latest(self) -> tuple[dict | str, RepodataState]:
```
dholth, line 809: This code is very similar to what was removed from `subdir_data.py`

conda/gateways/repodata/jlap/fetch.py
```diff
@@ -376,15 +403,21 @@ def request_url_jlap_state(
             )
 
             if apply:
-                with timeme("Load "), json_path.open() as repodata:
+                with timeme("Load "):
                     # we haven't loaded repodata yet; it could fail to parse, or
                     # have the wrong hash.
-                    repodata_json = json.load(repodata)  # check have_hash here
                     # if this fails, then we also need to fetch again from 0
+                    repodata_json = json.loads(cache.load())
```
dholth, line 410: This locks the `.info.json` file and loads both it (cache.state) and the cached index data while holding the lock.

dholth:
It's strange that `FAILED tests/cli/test_common.py::test_confirm_yn_no - Failed: DID NOT RAISE <class 'conda.exceptions.CondaSystemExit'>` should fail, and on OSX only.

The interface between fetchers and "write cache headers back to cache.state or passed-in state dict" should be refined.

conda/core/subdir_data.py
```diff
@@ -5,37 +5,33 @@
...
+from os.path import exists, join, splitext
 from pathlib import Path
 from time import time
 
 from genericpath import getmtime, isfile
 
-try:
-    from boltons.setutils import IndexedSet
-except ImportError:  # pragma: no cover
-    from .._vendor.boltons.setutils import IndexedSet
```
beeankha, line 25: We need to keep this import as-is, since `_vendor.boltons` has been deprecated (see [#12453](https://github.com/conda/conda/pull/12453/files#diff-ff8022ba5d08310acbbc6f95e1bc77a53f69e00ffe6daa51026af8880f89612dR24-R27))

conda/core/subdir_data.py
```diff
@@ -718,18 +582,19 @@ def make_feature_record(feature_name):
     )
 
 
+@deprecated(
+    "23.1",
+    "24.1",
```
beeankha, line 587:
```suggestion
    "23.9",
    "24.3",
```

beeankha: (per [CEP 9 deprecation policy](https://github.com/conda-incubator/ceps/blob/main/cep-9.md))

conda/gateways/repodata/__init__.py
```diff
@@ -430,7 +477,8 @@ def has_format(self, format: str) -> tuple[bool, datetime.datetime | None]:
             return (value, last_checked)
         except (KeyError, ValueError, TypeError) as e:
             log.warn(
-                "error parsing `has_` object from `<cache key>.state.json`", exc_info=e
+                "error parsing `has_` object from `<cache key>a separate file`",
+                exc_info=e,
```
beeankha, line 481: > `<cache key>a separate file`

Is this correct?

conda/gateways/repodata/__init__.py
```diff
@@ -471,15 +519,15 @@ def __setitem__(self, key: str, item: Any) -> None:
...
+            key = self._aliased[key]
         else:
             raise KeyError(key)
         return super().__getitem__(key)
 
 
 class RepodataCache:
     """
-    Handle caching for a single repodata.json + repodata.state.json
+    Handle caching for a single repodata.json + repodataa separate file
```
beeankha, line 530: > `repodataa separate file`

Similar issue as the above, I think there might have been a search + replace gone awry

conda/gateways/repodata/__init__.py
beeankha: This was defined previously so it looks like we should keep it as `url_w_subdir`

conda/gateways/repodata/__init__.py
```diff
@@ -709,6 +714,251 @@ def timeout(self):
...
+                if raw_repodata is RepodataOnDisk:
+                    # this is handled very similar to a 304. Can the cases be merged?
+                    # we may need to read_bytes() and compare a hash to the state, instead.
+                    # XXX use self._repo_cache.load() or replace after passing temp path to jlap
+                    raw_repodata = self.cache_path_json.read_text()
+                    cache.state["size"] = len(raw_repodata)  # type: ignore
+                    stat = self.cache_path_json.stat()
+                    mtime_ns = stat.st_mtime_ns
+                    cache.state["mtime_ns"] = mtime_ns  # type: ignore
+                    cache.refresh()
```
dholth, line 911: This cache.refresh() is redundant with this branch's version of jlap/fetch.py but is still needed with the "classic" fetcher. How should the fetcher communicate that the cache [metadata] is also up-to-date?

jezdez: Could you pass in which fetcher is currently in use, to decide whether to call `cache.refresh()` or not?

Either by requesting that information from conda's context, or by implementing a `ClassicRepodataCache` and a `JLAPRepodataCache` and the related `ClassicRepodataFetch` and `JLAPRepodataFetch` configured with them?

dholth: We do already check with hasattr; maybe the contract of the new method is that the cache does not have to be updated on the caller side.

jezdez: That would suffice to me

jaimergp:
I am looking at how constructor will integrate this new interface, now that `fetch_repodata_remote_request` will be deprecated.

Apparently this function returned `None` for non-existing `subdir`s. The new interface equivalent (`SubdirData(Channel("file:///my-channels/jaime/linux-aarch64")).repo_fetch.fetch_latest()[0]`) will return `'{}'`.

The `None` type was problematic in `constructor` because it expected a valid repodata.json schema later on, and a workaround was added in https://github.com/conda/constructor/pull/512. This makes me wonder whether we should return a [schema](https://github.com/conda/schemas/blob/main/repodata-1.schema.json) (ignore the signature stuff there) compliant repodata.json with the required keys set to their default values (empty dicts or lists most of the time).


In other words, assuming a valid channel (this means it provides at least `noarch`) is requested a non-existing subdir (e.g. exotic architecture), which value should be returned by `.fetch_latest()` and friends?

- `'null'` (so it's parsed as `None`)
- `'{}'` (so it's parsed as `dict()`)
- Minimally compliant repodata.json:
```python
{
    '_url': <requested url>,  # ??? I don't think this is compliant
    'info': {
        'subdir': <requested subdir>
    },
    'packages': {},
    'packages.conda': {},
    'removed': []
}
```

beeankha (approved):

dholth:
@jaimergp I have also considered whether we should return a complete repodata.json (with no packages) for empty results, but haven't implemented that. Other existing code does return an empty dict {} for offline mode.

jaimergp:
Got it, thanks for explaining the rationale. If there's a precedent for empty dicts, let's leave it like that. We can create an issue to investigate whether returning fully formed yet empty repodatas needs to be done generally.

dholth:
I'm dissatisfied with/not sure the save-headers for jlap requests is correct; do you have an opinion on whether this changes "classic" non-jlap requests?. Let's link to conda-libmamba-solver's usage another time to make sure it works correctly with the "new API requires the cache object" detail.

jaimergp:
I am checking at https://github.com/conda/conda-libmamba-solver/pull/210 but the testing infra is broken now (due to not using a SubdirData subclass anymore).

conda/gateways/repodata/jlap/interface.py
```diff
@@ -24,8 +30,10 @@ def __init__(
         self,
         url: str,
         repodata_fn: str | None,
+        *,
```
jezdez, line 33: ⭐

jezdez (approved):

conda/gateways/repodata/__init__.py
kenodegard: gotcha, so in this PR we will refrain from additional changes here 👍🏼

conda/gateways/repodata/__init__.py
```diff
@@ -709,6 +714,251 @@ def timeout(self):
...
+            ):
+                log.debug(
+                    "Using cached data for %s at %s forced. Returning empty repodata.",
+                    self.url_w_repodata_fn,
+                    self.cache_path_json,
+                )
+                return (
+                    {},
+                    cache.state,
+                )  # XXX basic properties like info, packages, packages.conda? instead of {}?
```
kenodegard, line 841:
```suggestion
                # TODO: basic properties like info, packages, packages.conda? instead of {}?
                return {}, cache.state
```

conda/gateways/repodata/__init__.py
```diff
@@ -709,6 +714,251 @@ def timeout(self):
...
+        else:
+            if context.use_index_cache:
+                log.debug(
+                    "Using cached repodata for %s at %s because use_cache=True",
+                    self.url_w_repodata_fn,
+                    self.cache_path_json,
+                )
+
+                _internal_state = self.read_cache()
+                return _internal_state
```
kenodegard, line 852:
```suggestion
                return self.read_cache()
```

conda/gateways/repodata/__init__.py
```diff
@@ -709,6 +714,251 @@ def timeout(self):
...
+        except Response304ContentUnchanged:
+            log.debug(
+                "304 NOT MODIFIED for '%s'. Updating mtime and loading from disk",
+                self.url_w_repodata_fn,
+            )
+            cache.refresh()
+            # touch(self.cache_path_json) # not anymore, or the a separate file is invalid
+            # self._save_state(mod_etag_headers)
+            _internal_state = self.read_cache()
+            return _internal_state
```
kenodegard, line 899:
```suggestion
            return self.read_cache()
```

kenodegard (commented):
minor comments, thanks for working to break this PR down into smaller pieces

dholth:
🪩

conda/testing/fixtures.py
```diff
@@ -69,3 +70,15 @@ def reset_conda_context():
     yield
 
     reset_context()
+
+
+@pytest.fixture()
+def temp_package_cache(tmp_path_factory):
```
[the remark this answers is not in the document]
dholth: No, but it is possible that it does something similar.

jaimergp:
Just saw there's a missing f-string `f` in this line:

https://github.com/conda/conda/blob/72977501b6c3bc6573ec0e9ffce630fed64229cd/conda/gateways/repodata/__init__.py#L516

PR #12478, opened 2023-03-13
Make `conda info --verbose` work

### LINKED ISSUE #12420: Make `conda info -v` work
### What is the idea?

There is no `conda info -v` command to display a detailed/verbose `conda info` report. It is `conda info --all` instead, which is a bit unintuitive. The most commonly used flag for displaying detailed versions of the regular command is `-v`. 

### Why is this needed?

conda CLI should support the standard CLI commands

### What should happen?

`conda info -v` should work. 
`conda info --all` should also continue to work, as users/developers might have gotten used to that command.

### Additional Context

See [comment](https://github.com/conda/conda/issues/12259#issuecomment-1430309637).

ForgottenProgramme:
Resolves #12420 

Turn default `-v` flag off, and aliase `--all` with `-v`/`--verbose` for `conda info`

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -373,6 +373,7 @@ def configure_parser_info(sub_parsers):
         "--all",
         "-v",
         "--verbose",
+        dest="verbosity",
         action="store_true",
         help="Show all information.",
     )
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -312,12 +312,12 @@ def execute(args, parser):
 
     options = 'envs', 'system'
 
-    if args.all or context.json:
+    if args.verbosity or context.json:
         for option in options:
             setattr(args, option, True)
     info_dict = get_info_dict(args.system)
 
-    if (args.all or all(not getattr(args, opt) for opt in options)) and not context.json:
+    if (args.verbosity or all(not getattr(args, opt) for opt in options)) and not context.json:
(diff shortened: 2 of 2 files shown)


conda/cli/conda_argparse.py
```diff
@@ -361,15 +361,18 @@ def configure_parser_info(sub_parsers):
...
         action='store_true',
         default=NULL,
         help=SUPPRESS,
     )
     p.add_argument(
-        '-a', "--all",
+        "-a",
+        "--all",
+        "-v",
+        "--verbose",
```
jezdez, line 375: The problem is that `verbosity` is also a config setting (.condarc etc) and I don't think this change will appropriately handle the case of loading it from there, as it's not using `dest="verbosity"`.

I would suggest to apply this:
```suggestion
        "-a",
        "--all",
        "-v",
        "--verbose",
        dest="verbosity",
```

and then update the CLI handler (`main_info.execute`) to use `args.verbosity` instead of `args.all`.

jezdez:
There is a test file for the info subcommand, and it might be useful to add one for the new "-a means -v" feature

kenodegard (commented):
I'm uncertain about this changes since its hijacking the `--verbose` flag thats already been standardized across the rest of conda.

What if instead we just deprecate `-a`/`--all` in favor of `-v`/`--verbose`?

jaimergp:
Another possible solution is to have `-v` _imply_ `-a`?

PR #12482, opened 2023-03-13
Change how `IndexedSet` is asserted

jaimergp:
### Description

Covering for a problem found in https://github.com/conda/conda-libmamba-solver/pull/151.

If we mix the vendored and not vendored `IndexedSet`, the `assert` instance test can fail, since it doesn't test the interface, just classes identities.

So... I changed how that interface is ensured: the object must support list indexing and set differences to be considered `IndexedSet`-like. Not perfect, but probably good enough.

### Checklist - did you ...


jezdez (approved):

dholth:
We merged some pre-boltons-refactor SubdirData into post-boltons-refactor conda; it would also be valid to expunge the vendored boltons use from conda.

PR #12483, opened 2023-03-14
Indicate current shell when misconfigured

johnyaku:
### Description

On HPC systems, the shell used by the job scheduler may not always be the same as that used in interactive sessions. On our HPC, for example, the job scheduler uses `csh` while interactive sessions start in `bash`. This confuses users who may have configured, say, `~/.bashrc` to initialise conda but may not be aware that the job scheduler uses a different shell by default. These errors can be difficult to debug.

This PR includes the path to the current shell (taken from the `$SHELL` environment variable) in the error message. Hopefully this will make it easier to spot the fact that the current shell is not the expected shell, and to adjust the job submission flags accordingly.

This PR uses `os.popen()`. Not sure if this is best practice, but the spirit of the PR is to make these errors easier to debug. 

### Checklist - did you ...

Don't think this change is big enough to warrant any of the above.


conda/exceptions.py
```diff
@@ -298,9 +298,14 @@ def __init__(self, command):
         init_loggers(context)
         if command in activate_commands:
             # TODO: Point users to a page at conda-docs, which explains this context in more detail
+            # Perhaps this link would be appropriate? https://docs.conda.io/projects/conda/en/latest/commands/init.html
+            # Remind users what shell they are currently using.
+            # Can be different in job schedulers versus interactive sessions, for example
+            shell = os.popen('echo $SHELL').read().rstrip()
```
travishathaway, line 304:
```suggestion
            shell = os.getenv("SHELL")
```

No need to open a sub process here. Was there a special reason you were doing that?

conda/exceptions.py
```diff
@@ -298,9 +298,14 @@ def __init__(self, command):
         init_loggers(context)
         if command in activate_commands:
             # TODO: Point users to a page at conda-docs, which explains this context in more detail
+            # Perhaps this link would be appropriate? https://docs.conda.io/projects/conda/en/latest/commands/init.html
+            # Remind users what shell they are currently using.
+            # Can be different in job schedulers versus interactive sessions, for example
```
travishathaway, line 303:
```suggestion
```

It's not necessary to include these extra pieces of documentation. I believe it's pretty straightforward what's going on here.

travishathaway (changes requested):
Hi @johnyaku,

Thank you for submitting this pull request. I have a couple recommendations for improvement that I would like you to check out.

On top of those, another thing you should address is how this will appear for users on Windows. The process of determining the active running shell is a bit different; therefore, I don't think it's important to include. But, I would rather not have these users see empty parenthesis (e.g. `()`).

travishathaway:
@johnyaku,

I have also updated this branch with the latest changes from main, so you will have to perform the following command to get everything working correctly locally for you:

```
git pull
```

kenodegard:
Unfortunately the `$SHELL` environment variable isn't a good indicator of what shell `conda` was invoked from (see https://stackoverflow.com/questions/55597797/detect-whether-current-shell-is-powershell-in-python/55598796#55598796)

Issue #11321, opened 2022-03-11
Remove deprecated code


kenodegard:
### Summary

A relatively low hanging fruit to remove clearly old code that has been marked deprecated.

### Linked Issues & PRs

- https://github.com/conda/conda/issues/11318
- https://github.com/conda/conda/issues/11319
- https://github.com/conda/conda/issues/11320

PR #12493, opened 2023-03-15
Define deprecation helper for `argparse.Action`

kenodegard:
### Description

Define `deprecations.deprecated.action` for easier argument deprecation. This replaces the prior `PendingDeprecationAction` and `DeprecationAction` implementations.

With this change we also mark `conda env create --force` as pending deprecation in favor of `conda env create --yes` which is the standard argument elsewhere in conda.

Xref #11321

### Checklist - did you ...


jezdez (approved):
Looks good in general, cleaner than before. Just FYI there is also a `--force` in the rename command.

conda/deprecations.py
```diff
@@ -120,6 +121,35 @@ def inner(*args, **kwargs):
...
+    def action(
+        self,
+        deprecate_in: str,
+        remove_in: str,
+        action: Action,
+        *,
+        addendum: str | None = None,
+        stack: int = 0,
+    ):
+        class DeprecationMixin:
```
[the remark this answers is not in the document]
kenodegard: I'm not too worried with a slowdown here since it's a deprecation in favor of something else.

### `main` Benchmarks
Screenshot 2023-03-21 at 10 08 14 AM

### `kenodegard:argparse-action-deprecation` Benchmarks
Screenshot 2023-03-21 at 10 08 28 AM

jezdez (approved):
Thanks for making the change @kenodegard, assuming a rebase this could be merged.

conda/deprecations.py
```diff
@@ -124,6 +125,50 @@ def inner(*args, **kwargs):
...
+                        if inner_self.option_strings
+                        # if not a flag/switch, use the destination itself
+                        else f"`{inner_self.dest}`"
+                    ),
+                    addendum=addendum,
+                )
+
+                # alert developer that it's time to remove something
+                if not category:
+                    raise DeprecatedError(message)
```
travishathaway, line 157: I'm a little curious why this isn't being covered by tests. Isn't the point of this function to eventually trigger such a state? Seems like it would be  pretty import to make sure this line is covered.

docs/source/dev-guide/deprecations.md
travishathaway, line 1: Thanks for writing this documentation. It seems like this example may be a good use case to include in the `tests/test_deprecations.py` file too, no? Perhaps this could be the test that helps us get to 100% coverage for this pull request.

travishathaway (changes requested):
Looks good so far, but let's try to get testing coverage on the missing line. It would also be good to accommodate this new use case in the `tests/test_deprecations.py` file.

jezdez (dismissed):

conda/deprecations.py
kenodegard: You're right, added applicable tests

docs/source/dev-guide/deprecations.md
kenodegard: You're right, added applicable tests and spruced up the old tests with docstrings

jezdez (dismissed):

docs/source/dev-guide/deprecations.md
```diff
@@ -107,6 +107,35 @@ def bar(enabled=True):
 <stdin>:1: PendingDeprecationWarning: foo.bar(is_true) is pending deprecation and will be removed in 24.3. Use 'enabled' instead.
 ```
 
+## `argparse.Action`
+
+Occasionally there is a need to deprecate CLI arguments. For this we provide a helper function to monkeypatch any `argparse.Action`:
```
travishathaway, line 112:
```suggestion
Occasionally, there is a need to deprecate CLI arguments. For this, we provide a helper function to monkeypatch any `argparse.Action`:
```

travishathaway (approved):

Issue #9943, opened 2020-05-16
conda create --clone not copying pinned file


itsallonetome:
## Current Behavior

$ `conda create --name fastai --clone pytorch`
creates a duplicate environment but omits the `conda-meta/pinned` file so an update installs "wrong" versions.

### Steps to Reproduce

$ `conda update --all` on an environment with pinned packages
$ `conda create --name badun --clone goodenv` to create a clone
$ `conda activate badun`
$ `conda update --all` updates the packages which should be pinned.

## Expected Behavior

The pinned packages in the previous environment are pinned in the cloned environment.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : pytorch
    active env location : /home/conda/miniconda3/envs/pytorch
            shell level : 3
       user config file : /home/conda/.condarc
 populated config files : /home/conda/.condarc
                          /home/conda/miniconda3/envs/pytorch/.condarc
          conda version : 4.8.3
    conda-build version : not installed
         python version : 3.7.7.final.0
       virtual packages : __glibc=2.17
       base environment : /home/conda/miniconda3  (writable)
           channel URLs : https://conda.anaconda.org/pytorch/linux-64
                          https://conda.anaconda.org/pytorch/noarch
                          https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/conda/miniconda3/pkgs
                          /home/conda/.conda/pkgs
       envs directories : /home/conda/miniconda3/envs
                          /home/conda/.conda/envs
               platform : linux-64
             user-agent : conda/4.8.3 requests/2.23.0 CPython/3.7.7 Linux/3.10.0-1127.el7.x86_64 rhel/7.6 glibc/2.17
                UID:GID : 1002:1002
             netrc file : None
           offline mode : False

```
</p></details>

<details open><summary><code>`conda config --show-sources`</code></summary><p>

```
==> /home/conda/.condarc <==
channels:
  - defaults
  - conda-forge

==> /home/conda/miniconda3/envs/pytorch/.condarc <==
disallowed_packages:
  - cudnn
  - cudatoolkit
channels:
  - pytorch
  - conda-forge
  - defaults

```
[...]

Issue #1769, opened 2015-11-02
cloning leaves out the pinning


michaelaye:
I just cloned an numpy 1.9.\* pinned environment like so:

``` bash
$ conda create -n py35clone --clone py35
Fetching package metadata: ....
src_prefix: '/usr/local/python3/miniconda/envs/py35'
dst_prefix: '/usr/local/python3/miniconda/envs/py35clone'
Packages: 41
Files: 3994
Linking packages ...
[      COMPLETE      ]|#########################################################################| 100%
#
# To activate this environment, use:
# $ source activate py35clone
#
# To deactivate this environment, use:
# $ source deactivate
#
```

but the pinned file was not copied over. System: OSX 10.10.5

megies:
Can confirm, just wanted to report the same.

megies:
Any comments by the devs/maintainers? Is this considered a feature? For me, pinning is essential to usage of anaconda envs and thus should be a part of making an exact clone (especially since #1967 has kept breaking my envs frequently for months with numpy updates not making sure that all packages compiled against numpy were still compatible).

michaelaye:
I would say that cloning currently is not reliable, at least I cannot rely on it so I prefer to do things by hand for now. Even using a package list file, I don't get consistent envs, see #2315 and #2316

PR #12499, opened 2023-03-18
Include pinned file with clone

scdub:
### Description

Issues addressed:

 - https://github.com/conda/conda/issues/1769
 - https://github.com/conda/conda/issues/9943

For long term reproducible environments, the `pinned` file is an important way of ensuring environment consistency independently of direct package dependencies. This PR is cautious by adding this as an optional `--pinned` flag to the clone operation, but arguably this should be on by default as the documentation mentions creating "an exact copy of an environment". This flag could be removed and the behavior enabled by default in a reworking of this PR if that is preferable.

### Checklist - did you ...

I can update the documentation to reflect the appropriate changes as needed, but figured it would be better to review the code changes first.

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -716,6 +716,11 @@ def configure_parser_create(sub_parsers):
         dest="dev",
         default=NULL,
     )
+    p.add_argument(
+        "--pinned",
+        action="store_true",
+        help="Copy pinned file when cloning the environment with `--clone`.",
+    )
     p.set_defaults(func='.main_create.execute')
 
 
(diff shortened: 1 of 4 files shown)


kenodegard (changes requested):
@scdub this is a great find!

To get this going again are you able to rebase this? I'm also thinking we should change this to be enabled by default with the option to disable.

If not let me know and I'll get this going for you!

tullis:
Hi @kenodegard - We at Wikimedia would really value this feature as part of conda clone and I think your suggestion of making it copy the pinned file by default is also great.

You mentioned that you might take it forward if you didn't hear back from @scdub - is this still a possibility? Many thanks.

PR #12503, opened 2023-03-20
add emscripten-32 platform support

wolfv:
### Description

We need support for the `emscripten-32` platform in [`emscripten-forge`](https://github.com/emscripten-forge/recipes)

### Checklist - did you ...


wolfv:
I don't know why adding a new platform would automatically mean that it is supported (by e.g. Anaconda). In my mind this is not necessarily strongly connected (if Anaconda starts serving emscripten-32, then it would be "supported" IMO - or do you support e.g. `linux-32`?). If you want I can add a comment in the code. 

Alternatively, if we had hooks to configure additional platforms somehow, we could try to make that work as well.

There are no concrete plans for emscripten-forge except for we're continuing to build it out and covering more and more packages. It's working pretty great so far.

The long term goal is still to move this to conda-forge, but for that we're waiting for a better compatible FORTRAN compiler (likely `LFortran`).

jezdez (changes requested):
Thinking about this a little more:

- let's add a news item to the PR, clarifying that the platform is being added to support the emscripten-forge project
- add a comment where to find more information about the platform (or lack thereof)

wolfv:
We had some discussions about this at @prefix-dev and maybe 

- `emscripten-wasm32`
- `wasi-wasm32`

would be better suited because then we know that the "arch" is wasm32 and the "OS" is either WASI or Emscripten.

PR #12504, opened 2023-03-20
Make split line f-string

dholth:
### Description

Line split to satisfy linter missed f-prefix

### Checklist - did you ...


jezdez (approved):

dholth:
Would it have been better to mark the line with an "ignore line length" comment, instead of splitting the line in this way, avoiding this bug?

PR #12518, opened 2023-03-22
Improve error message about -p/-n flags

### LINKED ISSUE #9762: Clarifying -n vs -p "error" messages
New(ish) users don't always use the right flags. `-n` instead of `-p` is a pretty common failure mode. A recent interaction with a novice-ish conda user produced the following:

> The error message is perplexing given that the supplied path is a legal Windows path. (I also tried enclosing the path in double quotes, to no avail).
and the terminal output:

```
(base) C:\>conda create --clone base -n C:\ProgramData\Anaconda3\envs\py37
 
CondaValueError: Invalid environment name: 'C:\ProgramData\Anaconda3\envs\py37'
 Characters not allowed: ('/', ' ', ':', '#')
```

Suggested resolution would be to check if the argument passed to `-n` is a full path and, if so, suggest to the user that they may be meaning to use `-p` instead

ForgottenProgramme:
Resolves #9762 

Improves error message to clarify that the user should use the `-p` flag instead of the `-n` flag when passing an environment path.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1776,7 +1776,7 @@ def validate_prefix_name(prefix_name: str, ctx: Context, allow_base=True) -> str
                 f"""
                 Invalid environment name: {prefix_name!r}
                 Characters not allowed: {PREFIX_NAME_DISALLOWED_CHARS}
-                If you are specifying a path to an environment, the `-p` 
+                If you are specifying a path to an environment, the `-p`
                 flag should be used instead.
                 """
             )


kathatherine (dismissed):

kathatherine (commented):
The error works for me overall. For a newbie, would they understand the term "passing"? Is that something we want to cater to by changing it to "using"? Just a suggestion.

conda/base/context.py
```diff
@@ -1776,6 +1776,7 @@ def validate_prefix_name(prefix_name: str, ctx: Context, allow_base=True) -> str
                 f"""
                 Invalid environment name: {prefix_name!r}
                 Characters not allowed: {PREFIX_NAME_DISALLOWED_CHARS}
+                Note that with an environment path, the `-p` flag should be used instead.
```
travishathaway, line 1779:
```suggestion
                If you are specifying a path to an environment, the `-p` flag should be used instead.
```

I think technically they are not referred to as "environment paths" (officially we call them "environment prefixes"). I thought this suggestion makes it a little clearer.

conda/base/context.py
```diff
@@ -1776,6 +1776,7 @@ def validate_prefix_name(prefix_name: str, ctx: Context, allow_base=True) -> str
                 f"""
                 Invalid environment name: {prefix_name!r}
                 Characters not allowed: {PREFIX_NAME_DISALLOWED_CHARS}
+                If you are specifying a path to an environment, the `-p` flag should be used instead.
```
travishathaway, line 1779:
```suggestion
                If you are specifying a path to an environment, the `-p` 
                flag should be used instead.
```

travishathaway (approved):

conda/base/context.py
```diff
@@ -1776,6 +1776,8 @@ def validate_prefix_name(prefix_name: str, ctx: Context, allow_base=True) -> str
                 f"""
                 Invalid environment name: {prefix_name!r}
                 Characters not allowed: {PREFIX_NAME_DISALLOWED_CHARS}
+                If you are specifying a path to an environment, the `-p` 
+                flag should be used instead.
```
travishathaway, line 1780:
```suggestion
                If you are specifying a path to an environment, the `-p`
                flag should be used instead.
```

travishathaway (approved):

jezdez (approved):

beeankha (approved):

PR #12523, opened 2023-03-22
Test that cache revalidate updates ttl

dholth:
### Description

This test is meant to check a case that I was concerned about getting wrong.

1. Download repodata
2. Cache expires after 30 seconds
3. Will the cache "stay expired", checking "304 Not Modified" every time after the original expiration? Or will it add another 30 seconds and correctly check no more frequently than once every 30 seconds?

### Checklist - did you ...

### DIFF
--- a/tests/gateways/test_jlap.py
+++ b/tests/gateways/test_jlap.py
@@ -5,6 +5,7 @@ Test that SubdirData is able to use (or skip) incremental jlap downloads.
 """
 import datetime
 import json
+import time
 from pathlib import Path
 from socket import socket
 from unittest.mock import Mock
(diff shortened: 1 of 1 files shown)


tests/gateways/test_jlap.py
```diff
@@ -28,6 +29,10 @@
 from conda.gateways.repodata.jlap import core, fetch, interface
 from conda.models.channel import Channel
 
+import logging
+
+log = logging.getLogger(__name__)
+
```
jezdez, line 35:
```suggestion
```

jezdez (dismissed):

jezdez (approved):

Issue #12535, opened 2023-03-27
conda clean: error in main_clean.py: UnboundLocalError: local variable 'stat' referenced before assignment


bernt-matthias:
### What happened?

xref https://github.com/conda/conda/issues/12187 .. not sure why this has been closed

running `conda clean -p -i -y` raises the following error:

```
Traceback (most recent call last):
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/exceptions.py", line 1118, in __call__
    return func(*args, **kwargs)
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
    exit_code = do_call(args, p)
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 91, in do_call
    return getattr(module, func_name)(args, parser)
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main_clean.py", line 298, in execute
    json_result = _execute(args, parser)
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main_clean.py", line 282, in _execute
    json_result["packages"] = pkgs = find_pkgs()
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main_clean.py", line 98, in find_pkgs
    size = sum(
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main_clean.py", line 99, in <genexpr>
    _get_size(root, file, warnings=warnings)
  File "/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/lib/python3.10/site-packages/conda/cli/main_clean.py", line 28, in _get_size
    if stat.st_nlink > 1:
UnboundLocalError: local variable 'stat' referenced before assignment
```

<details>
<summary><code>ERROR REPORT</code></summary>

```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
[...]

PR #12536, opened 2023-03-27
Ignore paths that cannot be stat-ed

kenodegard:
### Description

`conda clean` produced an `UnboundLocalError` if any errors occurred while stat-ing paths to remove. This change captures this error for display but otherwise ignores the issue and lets the user resolve the issue on their own.

Resolves https://github.com/conda/conda/issues/12535

### Checklist - did you ...


conda/cli/main_clean.py
```diff
@@ -128,8 +130,8 @@ def rm_pkgs(
     from ..utils import human_bytes
 
     if verbose and warnings:
-        for fn, exception in warnings:
-            print(exception)
+        for path, exception in warnings:
+            print(path, exception)
```
dholth, line 134: Call this function with verbose and warnings in the tests

conda/cli/main_clean.py
```diff
@@ -1,41 +1,43 @@
...
-from typing import Any, Dict, Iterable, List, Tuple
+from typing import Any, Iterable
 import sys
 
 from ..base.constants import CONDA_PACKAGE_EXTENSIONS, CONDA_TEMP_EXTENSIONS, CONDA_LOGS_DIR
 from ..base.context import context
 
 log = getLogger(__name__)
-_EXTS = (*CONDA_PACKAGE_EXTENSIONS, *(f"{e}.part" for e in CONDA_PACKAGE_EXTENSIONS))
+_EXTENSIONS = (*CONDA_PACKAGE_EXTENSIONS, *(f"{ext}.part" for ext in CONDA_PACKAGE_EXTENSIONS))
```
dholth, line 15: Do we use `.part` while downloading packages? It might be a good idea but I don't recall the code that does so.

kenodegard: not sure

conda/cli/main_clean.py
```diff
@@ -132,8 +137,8 @@ def rm_pkgs(
     from .common import confirm_yn
 
     if verbose and warnings:
-        for fn, exception in warnings:
-            print(exception)
+        for warning in warnings:
+            print(warning)
```
travishathaway, line 141: Is there relatively straightforward way to add coverage to these lines?

jezdez (approved):

Issue #12549, opened 2023-03-29
Optimize imports in hot paths and commands


jaimergp:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other speed issues and didn't find a duplicate

### What happened?

While researching https://github.com/conda-forge/conda-feedstock/pull/204#issuecomment-1488257409, I found out a surprising amount of imports in hot code paths.

I am profiling with this approach:

* `PYTHONPROFILEIMPORTTIME=1 conda [COMMAND] 2> conda_[COMMAND].import.profile`
* `tuna conda_[COMMAND].import.txt` ([tuna](https://github.com/nschloe/tuna) project)

---

`conda activate` should be fast and shouldn't need any exotic deps; it's just generating shell code. We can check for that with `python -Im conda shell.bash activate`.

[conda_activate.import.txt](https://github.com/conda/conda/files/11099249/conda_activate.import.txt)

![image](https://user-images.githubusercontent.com/2559438/228498199-01376e81-01c9-48e2-be58-40afd3bdf4c6.png)

That `distutils.spawn ` import call ends up in this tree of imports

![image](https://user-images.githubusercontent.com/2559438/228498365-494aa0fd-5ed3-4191-8ebf-74848198ddc6.png)

Mostly because of https://github.com/conda/conda/blob/88a5a04ff105a8af1379e3280e6a286521d9be46/conda/common/path.py#L18, which can be replaced with `shutil.which` if anything, and of course be deferred til really needed in that module. Where is it needed? Just a wrapper function... named [`which`](https://github.com/conda/conda/blob/88a5a04ff105a8af1379e3280e6a286521d9be46/conda/common/path.py#L333) 🤦  So the whole thing can be avoided and eventually removed after deprecating it.

---

`conda info` shouldn't take too long either.

[conda_info.import.txt](https://github.com/conda/conda/files/11099227/conda_info.import.txt)

![image](https://user-images.githubusercontent.com/2559438/228499560-e32ed24a-a7d3-47cd-b99a-be4ccc03e63c.png)

Again we see the `conda.common.path` -> `distutils` mess mentioned above, but there's also a new one: `conda.core.solve` importing `conda.core.index` which goes through a spiral of imports that account for 25% of the execution time (and we only need this import to retrieve the user agent).

### Debug

```shell
<Empty>
```

### Conda info

```shell
active environment : None
       user config file : /home/test_user/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 23.1.0
[...]

PR #12550, opened 2023-03-29
Optimize imports for `conda activate` calls

jaimergp:
### Description

Part of https://github.com/conda/conda/issues/12549.

### Checklist - did you ...


jaimergp:
Timings for `conda activate`. Note I edited `__init__.py` so it doesn't use `__version__.py` ([explanation here](https://github.com/conda/conda/issues/12549#issuecomment-1488363135)), but those changes are not in this PR (unrelated work)

# Tuna plots

```bash
$ git checkout <ref>
$ PYTHONPROFILEIMPORTTIME=1 python -m conda shell.bash activate 2> <ref>.txt
$ tuna <ref>.txt
```

Starting point (`main`):

image

[main.log.txt](https://github.com/conda/conda/files/11100406/main.log.txt)


Last commit so far (`dd27949`):

image

[dd27949.log.txt](https://github.com/conda/conda/files/11100407/dd27949.log.txt)


---

# Benchmarks

## `time` * 5

```
$ time python -m conda shell.bash activate
```

### ref = main
Result | 1   | 2   | 3   | 4   | 5   |
------ | :-: | :-: | :-: | :-: | :-: |
real   | `0m0.571s` | `0m0.310s` |  `0m0.292s` | `0m0.360s` | `0m0.310s` |
user   | `0m0.281s` | `0m0.188s` |  `0m0.182s` | `0m0.202s` | `0m0.178s` |
sys    | `0m0.097s` | `0m0.042s` |  `0m0.035s` | `0m0.056s` | `0m0.036s` |

### ref = dd27949

Result | 1   | 2   | 3   | 4   | 5   |
------ | :-: | :-: | :-: | :-: | :-: |
real   |` 0m0.323s` | `0m0.201s` | `0m0.211s` | `0m0.176s` | `0m0.176s` 
user   |` 0m0.131s` | `0m0.087s` | `0m0.103s` | `0m0.084s` | `0m0.084s` 
sys    |` 0m0.054s` | `0m0.030s` | `0m0.033s` | `0m0.028s` | `0m0.028s` 

## `timeit`

```
$ python -m timeit 'from subprocess import run; import sys; run([sys.executable, "-mconda", "shell.bash", "activate"])'
```

### ref = main

```
1 loop, best of 5: 251 msec per loop
```

### ref = dd27949

```
2 loops, best of 5: 148 msec per loop
```

conda/exception_handler.py
```diff
@@ -0,0 +1,337 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+import sys
+from functools import lru_cache, partial
+from logging import getLogger
+import os
```
jezdez, line 6:
```suggestion
import os
import sys
from functools import lru_cache, partial
from logging import getLogger
```

conda/common/path.py
```diff
@@ -330,7 +331,10 @@ def _translation(found_path):  # NOQA
 
 
 def which(executable):
-    return find_executable(executable)
+    """Backwards-compatibility wrapper. Use `shutil.which` directly if possible."""
+    from shutil import which
+
+    return which(executable)
```
jezdez, line 337: This should be marked for deprecation if there is a stdlib alternative

conda/deprecations.py
```diff
@@ -8,7 +8,7 @@
 
 from packaging.version import parse, Version
 
-from .__version__ import __version__
+from . import __version__
```
jezdez, line 11: IIRC `conda/__init__.py` imports `deprecations`, wouldn't that trigger a cyclic import?

conda/common/path.py
jaimergp: Yes, but we need to update all `which` uses in the project first. I didn't do it yet so this PR doesn't grow too much. My concern is that if we add a DeprecationWarning and some config tells Python to print these (e.g. env var) this might pollute the output in some commands. Although it should go to stderr so maybe not that big of a deal.

dholth: I agree that it's useful to redirect calls through an alias to be able to change the way conda works with one edit, instead of many; an alias has no maintenance cost; and there are other more profitable deprecations such as any code that is right now in the way of producing a user-visible improvement to the product.

jezdez: @dholth This makes it sound like you're not in favor of our deprecation policy, which I beg to differ. Deprecation is needed (and it's totally fine to do it when the actual uses of which are changed) since many 3rd party libraries are importing random stuff from conda. The deprecations are the way to tell them about it.

conda/deprecations.py
jezdez: @kenodegard Can you take a look at this proposal please?

conda/activate.py
```diff
@@ -764,7 +764,7 @@ def native_path_to_unix(paths):  # pragma: unix no cover
     # cygpath.  When this happens, we have two different virtual POSIX machines, rooted at
     # different points in the Windows filesystem.  We do our path conversions with one and
     # expect the results to work with the other.  It does not.
-    from .common.path import which
+    from shutil import which
```
jezdez, line 767: Thinking about this a little more, we need to look into the differences between `distutils.spawn.find_exectuable` (which is what `conda.common.path.which` uses) and `shutil.which`:

https://github.com/python/cpython/blob/c1e71ce56fdb3eab62ad3190d09130f800e54610/Lib/shutil.py#L1452-L1530

vs

https://github.com/python/cpython/blob/d58ff6a616276c205e480225f588140f480271b7/Lib/distutils/spawn.py#L95-L129

In particular, `shutil.which` will:

- prepend `PATH` on Windows with the CWD
- will not only look for exe files but also the ones set in the `PATHEXT` env var, or when not set the list of `.COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC`
- runs access checks (if it exists and is executable) on the found path, which may have performance impacts on network drives

I just want to make sure we test these on Windows before committing to this change

jaimergp: The Windows suite passes but we can run more things locally if needed. `find_executable` [also checks for existence](https://github.com/python/cpython/blob/d58ff6a616276c205e480225f588140f480271b7/Lib/distutils/spawn.py#L126), which is hopefully as slow on those cases as existence+access.

We _could_ add `os.curdir` to the end of PATH on purpose on Windows but I feel that would be more surprising than leaving it as is. I suspect there were strong reasons to add the default behaviour in `shutil.which`.

What kind of tests did you have in mind?

conda/common/path.py
```diff
@@ -359,7 +362,10 @@ def _translation(found_path):  # NOQA
...
```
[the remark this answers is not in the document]
jezdez: This won't suffice, since the prepending happens after it checks for `path`: https://github.com/python/cpython/blob/c1e71ce56fdb3eab62ad3190d09130f800e54610/Lib/shutil.py#L1494-L1500

`shutil.which` does this to mimic the `which` command line tool I think, which is probably a reasonable assumption, but it does mean a slight change in behavior.

It would be beneficial to find out where `which` is used and in what context to gauge to potential blast radius for this change in behavior. Based on that we could derive the appropriate way to inform the users (deprecation, breaking change etc). This problem might be more academic than real.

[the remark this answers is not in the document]
jaimergp: I see, it does check whether `curdir` is there so we can add it to the end of the list, but again, I concur with you that we might be overthinking this. If anything, I would have found `find_executable`'s behaviour buggy because `curdir` is searched for on Windows almost always.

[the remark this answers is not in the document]
jaimergp: And we can always copy/vendor `find_executable` directly. [It's a small, self-contained function](https://github.com/pypa/setuptools/blob/fca6d742d4a05b7ce5be1328d9eb517ca96cce7e/setuptools/_distutils/spawn.py#L75).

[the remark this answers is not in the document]
jezdez: Thinking about this more, I think we should be fine.

conda/activate.py
jezdez: Let's keep it the way you suggested, since I think `shutil` is a more complete approach than `distutils`' version, and it's also maintained.

jezdez:
@jaimergp I think this is good to go

jaimergp:
Are we still in time for inclusion in `23.5`?

conda/deprecations.py
kenodegard: even if this change works, why would we want it? it feels like its the opposite of what this PR is attempting to do

isn't it safer to import stuff from wherever it is defined instead of from other places where it was imported?

jezdez (approved):

beeankha (approved):

jaimergp:
🚀🚀🚀

PR #12554, opened 2023-03-30
Update Python auto formatting tooling

### LINKED ISSUE #12522: Reformat with black
### What is the idea?

We [announced that we would reformat conda projects with black.](https://conda.discourse.group/t/intent-to-format-all-of-conda-conda-build-with-black/182).

### Why is this needed?

It will be easier to reformat everything all the time, instead of having to avoid reformatting unchanged code.

### What should happen?

```[tasklist]
- [ ] format code with black https://github.com/conda/conda/pull/12554
- [ ] update pre-commit to stop using darker https://github.com/conda/conda/pull/12554
- [ ] update pre-commit to use black https://github.com/conda/conda/pull/12554
- [ ] exclude formatted code commit from git history/blame https://github.com/conda/conda/pull/12554
```

kenodegard:
### Description

Adds:
- isort
- black

Removes:
- darker

This will be merged as is (no squash merge).

Resolves #12522

### Checklist - did you ...


.pre-commit-config.yaml
```diff
@@ -29,30 +41,46 @@ repos:
...
+        files: \.py$
+        args: [--license-filepath, .github/disclaimer.txt, --no-extra-eol]
+        exclude: ^conda/auxlib/
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.3.1
     hooks:
+      # upgrade standard Python codes
       - id: pyupgrade
-        args: ["--py37-plus"]
+        args: [--py38-plus]
```
dholth, line 64: Harsh to switch back to an unquoted "reader has to remember yaml's string rules" string?

pyproject.toml
```diff
@@ -1,13 +1,16 @@
 [tool.black]
 line-length = 99
```
dholth, line 2: 99 columns of code on the wall. Would have liked black's shorter default but that would be an even bigger change.

.pre-commit-config.yaml
kenodegard: oh, my intention is to enable an auto formatter for YAMLs in a separate PR, it does this for us

pyproject.toml
kenodegard: I'm ok with removing this, just say the word

dholth: comparing...

jezdez: I wouldn't mind the shorter version, too

kenodegard: I consider this decided then 😄

dholth (dismissed):

travishathaway (approved):

PR #12582, opened 2023-04-10
Add `conda rename` support to `conda.testing.integration.run_command`

kenodegard:
### Description

Updated `conda.testing.integration.run_command` to support either `--prefix` or `--name` and to support `conda rename`. This allows us to refactor test_main_rename to replace the subprocess based `conda.testing.helpers.run_inprocess_conda_command` invocations.

Derived from https://github.com/conda/conda/pull/12493

### Checklist - did you ...


tests/cli/test_main_rename.py
```diff
@@ -4,285 +4,227 @@
...
-    """Used to get an environment created using -p flag"""
-    # Setup
-    tmpdir = tempfile.mkdtemp()
-    run(f"conda create -p {tmpdir} -y", disallow_stderr=False)
+def prefix_rename() -> Generator[Path, None, None]:
+    """A teardown fixture, removes the renamed environment if present."""
+    with tempfile.TemporaryDirectory() as root:
+        prefix = Path(root, uuid.uuid4().hex)
 
-    yield tmpdir
```
dholth, line 60: This is a little confusing; exiting the TemporaryDirectory() context manager didn't delete the entire temporary directory before? Does the remove command give us (removing the environment from a list of env's textfile?) or faster deletes compared  to standard temp cleanup?

kenodegard: good catch, we dont need the `run_command(Commands.REMOVE, ...` below since we now use the `tempfile.TemporaryDirectory` context

travishathaway (changes requested):
I would rather not have this merged. Let's talk about this some more in a synchronous manner 😄 📞 .

tests/cli/test_main_rename.py
```diff
@@ -4,285 +4,223 @@
...
 
 
-def list_envs():
-    out, err, exit_code = run(ENV_LIST_COMMAND)
-    data = json.loads(out)
+@pytest.fixture
+def prefix_rename() -> Generator[Path, None, None]:
+    """A teardown fixture, removes the renamed environment if present."""
+    with tempfile.TemporaryDirectory() as root:
+        prefix = Path(root, uuid.uuid4().hex)
```
jezdez, line 57: This should use the [tmp_path fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) instead of a context manager inside a generator fixture

```suggestion
@pytest.fixture
def prefix_rename(tmp_path: Path) -> Generator[Path, None, None]:
    """A teardown fixture, removes the renamed environment if present."""
    prefix = tmp_path / uuid.uuid4().hex
```

jezdez (changes requested):
It might be useful to provide an extra fixture for an "uuid4 named temp directory" and this needs to stop using `tempfile` directly since `pytest` is better at handling temporary files than us. E.g. `tmp_path` allows to be configured with the `--basetemp` command line function, making it easier to for example use fast in-memory temp directories.

tests/cli/test_main_rename.py
dholth: Do we need a hex-named subdirectory of tmp_path or is tmp_path good enough by itsefl?

tests/cli/test_main_rename.py
```diff
@@ -4,283 +4,223 @@
...
+    """A setup fixture, creates an empty testing environment and removes it at the end if present."""
+    with tempfile.TemporaryDirectory() as root:
+        prefix = Path(root, uuid.uuid4().hex)
+        run_command(Commands.CREATE, str(prefix))
 
-def test_rename_by_path_success(env_one):
-    with tempfile.TemporaryDirectory() as temp_dir:
-        new_name = str(pathlib.Path(temp_dir).joinpath("new-env"))
-        run(f"conda rename -n {TEST_ENV_NAME_1} {new_name}", disallow_stderr=False)
+        yield prefix
```
beeankha, line 79: @jezdez should this `yield` statement also be indented by one? (as suggested [here](https://github.com/conda/conda/pull/12582/files#r1163841617) and [here](https://github.com/conda/conda/pull/12582/files#r1163844302))

PR #12592, opened 2023-04-13
Introducing `conda_cli`, `tmp_env`, and `path_factory` fixtures

kenodegard:
### Description

We have two different run subcommand implementations that effectively do the same thing. Each adds unnecessary complexities that we are better off avoiding all together.

```python
from conda.testing.integration import run_command, Commands
from pathlib import Path

def test_info_envs():
    out, err, code = run_commands(Commands.INFO, None, "--envs")

def test_create(tmp_path: Path):
    out, err, code = run_commands(Commands.CREATE, tmp_path, "--yes")
```

vs.

```python
from conda.testing.helpers import run_inprocess_conda_command
from pathlib import Path

def test_info_envs():
    out, err, code = run_inprocess_conda_command("conda info --envs")

def test_create(tmp_path: Path):
    out, err, code = run_inprocess_conda_command(f"conda create --prefix {tmp_path}" --yes")
```

Here we propose rewriting these into a pytest fixture:

```python
from conda.testing.integration import CondaCLIFixture
from pathlib import Path

def test_info_envs(conda_cli: CondaCLIFixture):
    out, err, code = conda_cli("info", "--envs")

def test_create(conda_cli: CondaCLIFixture, tmp_path: Path):
    out, err, code = conda_cli("create", "--prefix", tmp_path, "--yes")
```

In addition to the new `conda_cli` fixture we also implemented:
- `tmp_env` fixture to replace `conda.testing.integration.make_temp_env`
- `path_factory` fixture which does `tmp_path / uuid.uuid4().hex` for us

### Checklist - did you ...


conda/cli/main_remove.py
```diff
@@ -55,12 +54,11 @@ def execute(args, parser):
...
         if prefix == context.root_prefix:
             raise CondaEnvironmentError(
-                "cannot remove root environment,\n"
-                "       add -n NAME or -p PREFIX option"
+                "cannot remove root environment, add -n NAME or -p PREFIX option"
             )
         if not isfile(join(prefix, "conda-meta", "history")):
             raise DirectoryNotACondaEnvironmentError(prefix)
-        print("\nRemove all packages in environment %s:\n" % prefix, file=sys.stderr)
+        print(f"\nRemove all packages in environment {prefix}:\n")
```
dholth, line 61: changes stderr to stdout

conda/testing/__init__.py
```diff
@@ -156,3 +165,94 @@ def conda_check_versions_aligned():
...
+@dataclass
+class CondaCLIFixture:
+    capsys: CaptureFixture
+
+    def __call__(self, *argv: str, no_capture: bool = False) -> tuple[str, str, int]:
+        """Mimic what is done in `conda.cli.main.main`"""
+        # drop the conda argument if provided
+        if argv[0] == "conda":
+            argv = argv[1:]
+
```
dholth, line 179: needs test

tests/test_create.py
```diff
@@ -1361,7 +1361,7 @@ def test_conda_run_1(self):
             use_restricted_unicode=False, name=str(uuid4())[:7]
         ) as prefix:
             output, error, rc = run_command(Commands.RUN, prefix, "echo", "hello")
-            assert output == "hello" + os.linesep
+            assert output == f"hello{os.linesep}\n"
```
dholth, line 1364: If os.linesep is `\n` are we looking for `\n\n` now

conda/cli/main_info.py
```diff
@@ -345,9 +345,7 @@ def execute(args, parser):
     if (
         args.all or all(not getattr(args, opt) for opt in options)
     ) and not context.json:
-        stdout_logger = getLogger("conda.stdoutlog")
-        stdout_logger.info(get_main_info_str(info_dict))
-        stdout_logger.info("\n")
+        print(get_main_info_str(info_dict) + "\n")
```
dholth, line 348: Interesting way to send messages presumably to stdout with stdout_logger instead of print. I wonder if that flexibility was ever used.

conda/cli/main_remove.py
kenodegard: correct, IMO the text printed does not belong in stderr:

```bash
$ conda create -n scratch python --yes
...
$ conda remove -n scratch --all --yes > /dev/null

Remove all packages in environment /Users/kodegard/.conda/arm64/23.1.0/3.10/envs/scratch:

```

```bash
$ conda create -n scratch python --yes
...
$ conda remove -n scratch --all 2> /dev/null

## Package Plan ##

  environment location: /Users/kodegard/.conda/arm64/23.1.0/3.10/envs/scratch


The following packages will be REMOVED:

  bzip2-1.0.8-h620ffc9_4
  ca-certificates-2023.01.10-hca03da5_0
  certifi-2022.12.7-py311hca03da5_0
  libffi-3.4.2-hca03da5_6
  ncurses-6.4-h313beb8_0
  openssl-1.1.1t-h1a28f6b_0
  pip-23.0.1-py311hca03da5_0
  python-3.11.2-hc0d8a6c_0
  readline-8.2-h1a28f6b_0
  setuptools-65.6.3-py311hca03da5_0
  sqlite-3.41.2-h80987f9_0
  tk-8.6.12-hb8d0fd4_0
  tzdata-2023c-h04d1e81_0
  wheel-0.38.4-py311hca03da5_0
  xz-5.2.10-h80987f9_1
  zlib-1.2.13-h5a0b063_0


Preparing transaction: done
Verifying transaction: done
Executing transaction: done
```

tests/test_create.py
kenodegard: not entirely sure why this change was necessary but the change was test driven:
- [Windows failure](https://github.com/conda/conda/actions/runs/4696058510/jobs/8325797700#step:9:498)
- [Linux failure](https://github.com/conda/conda/actions/runs/4696058510/jobs/8325800428#step:3:276)

sven6002: Question: is the bash shell introducing the `\n`

tests/cli/test_subcommands.py
```diff
@@ -0,0 +1,230 @@
...
+
+def test_install(conda_cli: CondaCLIFixture, tmp_env: TmpEnvFixture):
+    with tmp_env() as prefix:
+        out, err, code = conda_cli(
+            "install",
+            *("--prefix", prefix),
+            "ca-certificates",
+            "--yes",
+        )
+        assert out
```
sven6002, line 80: `assert out is not None` should be the tested

conda/testing/__init__.py
```diff
@@ -156,3 +165,94 @@ def conda_check_versions_aligned():
         )
         with open(version_file, "w") as fh:
             fh.write(version_from_git)
+
+
+@dataclass
```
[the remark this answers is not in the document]
kenodegard: Originally I had:

```
class CondaCLIFixture:
    def __init__(self, capsys: CaptureFixture):
        self.capsys = capsys
```

Which is what a dataclass does 🤷🏼‍♂️

tests/conftest.py
```diff
@@ -5,6 +5,8 @@
 
 import pytest
 
+from conda.testing import conda_cli, path_factory, tmp_env
+
```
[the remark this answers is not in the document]
kenodegard: isort + black handles this formatting, if we remove it, it'll be re-added

tests/cli/test_subcommands.py
kenodegard: `conda_cli` returns `tuple[str, str, int]`, `out` will always be `not None` so that would be a redundant check

what I'm asserting here is whether anything was printed to `stdout` so an empty string is considered a failure

conda/testing/__init__.py
kenodegard: resolved by removing check, no value in being extra cautious here

beeankha (dismissed):

conda/cli/main_remove.py
jezdez: yeah that seemed to be a bug

conda/cli/main_info.py
jezdez: I'm curious about this change, the `conda.stdoutlog` logger behaves the same as `print()`?

jezdez (approved):
Some questions, but generally this is a big step forward

travishathaway (dismissed):
@kenodegard,

Looking good so far. Thanks for the very clear examples on the description of the pull request. That made this a lot easier to review.

Before we merge this, could we please update this article in the documentation?

https://docs.conda.io/projects/conda/en/stable/dev-guide/writing-tests/integration-tests.html

Shouldn't be too difficult.

conda/cli/main_info.py
kenodegard: It's effectively the same for the CLI.

I had to change it because the logging was resulting in the output getting collected differently so the new `conda_cli` didn't work correctly for `run` commands.

Per my understanding, an effort was previously undertaken to replace all print (to both stdout and stderr) with this kind of logging logic. My best guess is this was to potentially allow third-party tools (like navigator) to better redirect/interface with the text streams. This effort was never completed and the logging system has become over complicated as a result. (I'll ping the navigator team to check that this won't cause issues)

jezdez (approved):

beeankha (approved):

PR #12603, opened 2023-04-18
Simplify recipe and add missing tests

kenodegard:
### Description

Simplified recipe script to avoid Windows complexities. Also added additional package tests to improve packaging confidence.

> **Warning**
> For now, since both `run_test.*` and `test/commands` are defined only the former are run. This is a known issue and a fix is in the works, see https://github.com/conda/conda-build/pull/4429.

### Checklist - did you ...


sven6002 (approved):

recipe/run_test.bat
```diff
@@ -1,26 +1,27 @@
...
-FOR /F "delims=" %%i IN ('python -c "import sys; print(sys.version_info[1])"') DO set "ENV_PYTHON_MINOR_VERSION=%%i"
-IF NOT "%ENV_PYTHON_MINOR_VERSION%" == "%TEST_MINOR_VER%" EXIT /B 1
-@REM CALL conda deactivate
-rd /s /q "%TEMP%\built-conda-test-env"
+
+:: display conda details
+CALL conda info --all
+
+:: create, activate, and deactivate a conda environment
+CALL conda create --yes --prefix ".\built-conda-test-env" "m2-patch"
```
jezdez, line 17: Can this test environment live in a tmp folder instead so we don't fill up the local dev directory?

jezdez (approved):
LGTM with comments

recipe/run_test.bat
kenodegard: AFAIK these tests are run within a temp directory so this relative path is relative to that testing location and will be cleaned up after testing

jezdez: Ah, gotcha, nice!

Issue #12606, opened 2023-04-20
`conda clean` cannot delete directory because `rsync` error handling code is broken


JubilantJerry:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Overview
`conda clean --all` results in `WARNING: cannot remove, file permissions: [directory]` for several directories in the package cache.
All the inner contents of the directories are deleted. But the directories themselves are not.
A `.empty` file gets created in whatever directory my shell is currently on while running `conda clean --all`, and stays even after the end of the `conda clean --all` command.

### About the directories
The directories are all writeable. I can successfully delete the directories by directly running `rm -r [directory]` in the shell.
The directory's owner is not the same as the user running the command. But the group permissions are such that the directory can still be deleted.
The below shows one of the directories that `conda clean --all` could not delete:
```
(pytorch) jerry:~/Documents$ ls -l /srv/mambaforge/pkgs/ | grep pyzmq-24.0.1-py37h0c0c2a8_0
drwxrwsr--+  2 henry conda-users   4096  4月 20 17:38 pyzmq-24.0.1-py37h0c0c2a8_0
(pytorch) jerry:~/Documents$ groups
jerry sudo conda-users
```

### Root cause
I've looked into the code and the culprit is that the deletion is handled using `rsync`, which attempts to change the permissions, group, ownership, and timestamps of the directory because it is called with the `-a` flag.

This alone is only part of the problem. If `rsync` fails `conda` intends to fall back to using `shutil.rmtree`, which works fine in my case.
But the debug logging line itself raises a `NameError`, because it references a non-existent variable. So the error handling code that uses `shutil.rmtree` is unreachable.

The buggy line of code is: https://github.com/conda/conda/blob/main/conda/gateways/disk/delete.py#L128 :
```
except CalledProcessError:
    log.debug(
        f"removing dir contents the fast way failed.  Output was: {out}"  # 'out' is not defined because the exception prevents the variable from being set
    )
```

### Proposed solution
Fixing the logging line to not reference the variable will suffice, because the `shutil.rmtree` will work.
Alternatively, `rsync` should be run with flags `-rlD` instead of `-a`, to produce the same behavior as `shutil.rmtree`.
I've prepared a pull request with both changes made.

### Conda Info

```shell
active environment : pytorch
    active env location : /home/jerry/mambaforge/envs/pytorch
            shell level : 6
       user config file : /home/jerry/.condarc
[...]

PR #12607, opened 2023-04-20
Fix rsync deletion and its error handling path

JubilantJerry:
See issue #12606

### Description

Changed `rsync` call so that it doesn't attempt to synchronize permissions, owner, or timestamps against the `.empty` directory.
In other words, `rsync` is called with `-rlD` instead of `-a`.
Also fixed broken error handling path, which originally made error handling unreachable because of undefined variable issue (`NameError`)

### Checklist - did you ...

The change is very minor.


travishathaway:
@JubilantJerry,

It looks like the "black" auto code formatting failed for this pull request. Did you make the changes via GitHub? It's better to do this from a cloned version of the repository on your local computer so that the pre-commit hooks can run and make the desired changes.

JubilantJerry:
I've indeed used GitHub to do the change last time. I've figured out how to set up the precommit hook and have recreated the commit after cloning, the "black" issue should be resolved now.

travishathaway:
Hi @JubilantJerry,

This pull request sparked quite an interesting discussion amongst the conda maintainers. We are currently debating the validity of actually using `rsync` to remove directories containing many files.

We think now may be a good time to remove `rsync` altogether (this would hopefully also resolve the bug you initially reported) and instead just use something like `shutil.rmtree`.

Because `rsync` may have provided performance benefits, we want to first make sure that switching to `shutil.remove` won't cause any significant slow-downs. You're more than welcome to help out with this if you are interested. You can also update this pull request to use `shutil.tree` instead and remove the usage of `rsync`.

Otherwise, we will have to conduct this research first ourselves and will not likely be merging this pull request any time soon.

Please keep an eye on this pull request for future updates.

PR #12509, opened 2023-03-21
Convert `setup.py` to `pyproject.toml` using hatch build system

kenodegard:
### Description

This converts the `setup.py` to `pyproject.toml` using hatchling and hatch-vcs. Since we no longer rely on our custom version injection (i.e., `conda/.version`) we are also able to mark `conda.auxlib.packaging` for deprecation.

Resolves #12508 
Resolves #12525

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #12627, opened 2023-04-29
Ensure all paths in `conda.sh` are prefix-replaceable on Windows

### LINKED ISSUE #12626: `conda.sh` has incorrect paths on Windows
### What happened?

Previously, our recipe included a `sed` statement:

https://github.com/conda/conda/blob/ca036ed747d955d23fd0abd83295d96ddf74cc40/recipe/build.sh#L8-L10

This was removed for a simpler recipe implementation in #12603 (which in turn was prompted by issues with packaging conda for Python 3.11, see #12256).

Removing `sed` resulted in the `conda.sh` breaking on Windows + MSYS2. The issue is evident in the first few lines of `conda.sh`:

#### Before Recipe Change (`defaults::conda=23.3.1`)

```bash
export CONDA_EXE='C:/Users/kodegard/.conda/miniconda\Scripts\conda.exe'
export _CE_M=''
export _CE_CONDA=''
export CONDA_PYTHON_EXE='C:/Users/kodegard/.conda/miniconda\python.exe'
```

#### After Recipe Change (`conda-canary/label/dev::conda=23.3.1+23`)

```bash
export CONDA_EXE='/d/a/conda/conda/pkgs/conda_1682642894222/_h_env/Scripts/conda.exe'
export _CE_M=''
export _CE_CONDA=''
export CONDA_PYTHON_EXE='C:/Users/kodegard/.conda/miniconda\python.exe'
```

kenodegard:
### Description

Fix converting Windows path to unix path for cygwin and msys2.

The proposal is to have the PosixActivator conform to what the other activators already do. Within the shell scripts loaded at startup (e.g., the profile) we will rely on `cygpath` (which at that point should be the most correct `cygpath`) to convert our paths. When running `conda activate` we will still perform the path conversion in Python as was done previously.

Resolves #12626

### Checklist - did you ...


conda/activate.py
```diff
@@ -17,8 +17,9 @@
     isdir,
     join,
 )
+from pathlib import Path
 from textwrap import dedent
-from typing import Iterable
+from typing import Callable, Iterable
```
travishathaway, line 22:
```suggestion
from collections.abc import Callable, Iterable
```

news/12627-fix-msys2-conda.sh-paths
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
```
travishathaway, line 3: I would also argue that this pull request performs a significant refactor of the way that `Activator` classes are defined in `conda/activate.py`. I would mention this as an enhancement.

dbast:
PR #12509 actually lead to this issue on current main. So either this PR is merged or #12509 reverted as part of next release.

conda.sh is used in "git bash" (part of official windows git installer) and in many cases on GitHub actions windows runners.

beeankha:
pre-commit.ci autofix

dbast:
The title of this PR is a bit misleading... its not about using cygpath but more like **Ensure all paths in conda.sh are prefix replaceable on windows**.

news/12627-fix-msys2-conda.sh-paths
```diff
@@ -0,0 +1,19 @@
...
+
+### Bug fixes
+
+* Stop pre-converting paths to unix style on Windows in `conda.sh`. Instead rely on `cygpath` at runtime. (#12627)
```
dbast, line 7:
```suggestion
* Stop pre-converting paths to unix style on Windows in `conda.sh`, so that they are prefix replaceable upon installation, which got broken by #12509. It also relies on `cygpath` at runtime, which all msys2/cygwin bash versions on windows should have available. (#12627)
```

dbast (approved):

beeankha (approved):

conda/activate.py
```diff
@@ -34,7 +35,7 @@
 from .common.path import paths_equal
 
 
-class _Activator:
+class _Activator(metaclass=abc.ABCMeta):
```
jezdez, line 38: Woah!

conda/activate.py
```diff
@@ -814,102 +811,83 @@ def ensure_fs_path_encoding(value):
...
     if paths is None:
         return None
-    from subprocess import PIPE, CalledProcessError, Popen
 
-    from conda.auxlib.compat import shlex_split_unicode
+    if not on_win:
+        return path_identity(paths)
+
+    # on windows, uses cygpath to convert windows native paths to posix paths
+    from subprocess import run
```
jezdez, line 824: Why is this import here and not at the top of the file?

conda/activate.py
```diff
@@ -814,102 +811,83 @@ def ensure_fs_path_encoding(value):
...
-    return final[0] if single_path else tuple(final)
-
-
-def path_identity(paths):
+def path_identity(paths: str | Iterable[str] | None) -> str | tuple[str] | None:
+    if paths is None:
+        return None
+
     if isinstance(paths, str):
         return os.path.normpath(paths)
```
jezdez, line 856:
```suggestion
    if paths is None:
        return None
    elif isinstance(paths, str):
        return os.path.normpath(paths)
```

jezdez: You might combine this with the line below and an else clause, see next suggestion

conda/activate.py
```diff
@@ -814,102 +811,83 @@ def ensure_fs_path_encoding(value):
...
-    elif paths is None:
-        return None
-    else:
-        return tuple(os.path.normpath(_) for _ in paths)
+    return tuple(os.path.normpath(path) for path in paths)
```
jezdez, line 857:
```suggestion
    else:
        return tuple(os.path.normpath(path) for path in paths)
```

conda/activate.py
```diff
@@ -814,102 +811,83 @@ def ensure_fs_path_encoding(value):
...
+    if paths is None:
+        return None
 
-        self.unset_var_tmpl = "unset %s"
-        self.export_var_tmpl = "export %s='%s'"
-        self.set_var_tmpl = "%s='%s'"
-        self.run_script_tmpl = '. "%s"'
+    if isinstance(paths, str):
+        return paths.replace("\\", "/")
+    return tuple([path.replace("\\", "/") for path in paths])
```
jezdez, line 868:
```suggestion
    elif isinstance(paths, str):
        return paths.replace("\\", "/")
    else:
        return tuple([path.replace("\\", "/") for path in paths])
```

conda/activate.py
```diff
@@ -929,46 +907,40 @@ def _update_prompt(self, set_vars, conda_prompt_modifier):
...
 
-    def _hook_preamble(self):
-        result = ""
+    def _hook_preamble(self) -> str:
+        result = []
         for key, value in context.conda_exe_vars_dict.items():
             if value is None:
                 # Using `unset_var_tmpl` would cause issues for people running
                 # with shell flag -u set (error on unset).
-                # result += join(self.unset_var_tmpl % key) + '\n'
```
jezdez, line 938: That line is dropped, any idea why?

conda/activate.py
```diff
@@ -1276,10 +1221,10 @@ def _build_activator_cls(shell):
...
     shell_etc = shell.split("+")
     activator, formatters = shell_etc[0], shell_etc[1:]
-    bases = [activator_map[activator]]
 
+    bases = [activator_map[activator]]
     for f in formatters:
         bases.append(formatter_map[f])
 
-    cls = type("Activator", tuple(bases), {})
+    cls = type("Activator", tuple(reversed(bases)), {})
```
jezdez, line 1229: Any reason why the base classes are reversed here?

jezdez (commented):
Just some minor questions and nits

conda/activate.py
beeankha: I think it's to visually separate where the Windows path-related work is being done, though I'm not 100% sure if that's the reason. If you think it's best to have the import statements at the top of the file, I can move them.

conda/activate.py
dbast: What do you mean? That line is a comment line, the next line is converted to append and the `\n` handling moved to the end of the function, which is more generic.

conda/activate.py
jezdez: Gotcha, thanks for elaborating, writing this down for our future selves suffices :)

conda/activate.py
jezdez: Generally speaking, I find it a bit of a code smell when imports are allover the place in a module. The only reason to do this is to make sure the module is imported faster. But `subprocess` is probably a module that is loaded during the startup of conda somewhere as well, making this moot.

@jaimergp Have you had any concerns around this working on the very related #12550?

jezdez: For this this isn't a blocker for merging this..

jezdez (approved):

beeankha (approved):

Issue #12669, opened 2023-05-15
parametrize cache keys, extension to more easily track repodata state/info draft CEP


dholth:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Replace strings with constants to make it easier to track https://github.com/conda-incubator/ceps/pull/48

PR #12673, opened 2023-05-16
name hash algorithm in keys to better track draft CEP

dholth:
### Description

At first I thought it would be more conservative to keep the hash key names the same, but the other PR addressing #12669 changed the whole filename. Change those key names to represent the hash algorithm used. [The draft CEP is here as a PR against the conda/ceps repository.](https://github.com/conda-incubator/ceps/pull/48)

The nominal hash and actual hash are used to track jlap changes. The nominal hash comes from the upstream file, hashed as formatted on the server. We look for those hashes in the `.jlap` changelog. The actual hash is our on-disk hash since we do not format the saved json exactly the same as the server; instead, we format it in a more compact way.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/jlap/fetch.py
+++ b/conda/gateways/repodata/jlap/fetch.py
@@ -31,8 +31,8 @@ DIGEST_SIZE = 32  # 256 bits
 
 JLAP_KEY = "jlap"
 HEADERS = "headers"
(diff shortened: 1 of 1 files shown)


kenodegard (approved):

dholth:
Thanks.

jezdez (approved):

PR #12683, opened 2023-05-17
fix: use requests.exceptions.JSONDecodeError for ensuring compatibility with different json implementations used by requests

johanneskoester:
Otherwise, conda misses JSONDecodeErrors generated by simplejson if that is in the same env and thereby used by the requests package for decoding the json response. In turn, this can lead to aborting source download after trying the first given source in a meta.yaml instead of trying all of them.
Also see requests [JSONDecodeError implementation](https://github.com/psf/requests/blob/302225334678490ec66b3614a9dddb8a02c5f4fe/requests/exceptions.py#L31), which either inherits from simplejson's JSONDecodeError or from the Python std libs json.JSONDecodeError.

### Checklist - did you ...


jezdez (approved):

dholth (approved):

pyproject.toml
```diff
@@ -56,6 +56,7 @@ source = "vcs"
 
 [tool.hatch.version.raw-options]
 local_scheme = "dirty-tag"
+version_scheme = "calver-by-date"
```
kenodegard, line 59: Whoops, this was accidentally added back (removed in https://github.com/conda/conda/pull/12708)

PR #12691, opened 2023-05-18
Add missing `f` for fstring and update imports

beeankha:
Resolves [this comment](https://github.com/conda/conda/pull/12475#issuecomment-1552767334).

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -19,14 +19,14 @@ from os.path import dirname
 from pathlib import Path
 from typing import Any
 
-from conda import CondaError
-from conda.auxlib.logz import stringify
-from conda.base.constants import CONDA_HOMEPAGE_URL, REPODATA_FN
-from conda.base.context import context
-from conda.common.url import join_url, maybe_unquote
-from conda.core.package_cache_data import PackageCacheData
-from conda.deprecations import deprecated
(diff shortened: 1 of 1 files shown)


sven6002 (dismissed):
Looks good to me..

conda/gateways/repodata/__init__.py
```diff
@@ -513,7 +512,7 @@ def __setitem__(self, key: str, item: Any) -> None:
         if key in self._aliased:
             key = key[1:]  # strip underscore
         if key in self._strings and not isinstance(item, str):
-            warnings.warn('Replaced non-str RepodataState[{key}] with ""')
+            warnings.warn(f'Replaced non-str RepodataState[{key}] with ""')
```
jaimergp, line 515: Two things:

- Should we also log the `item` that got replaced?
- Should this be `logging.warning` instead?

beeankha: I consulted with @sven6002 and @kenodegard and we think it makes more sense for now to only use the `warnings` module exclusively for deprecations and `logging` everywhere else. The code has been updated accordingly!

jaimergp:
> Change the way we warn

Sounds like the catchy phrase at the end of a song 😂

jaimergp (approved):

jezdez (approved):

PR #12696, opened 2023-05-19
Don't set `$CONDA_PREFIX` in `context.__init__`

kenodegard:
### Description

Previously when `conda.base.context` is initialized it sets `$CONDA_PREFIX` based on whether `--name` or `--prefix` is present.

Under normal usage this works fine, but for testing purposes and for resetting the context purpose this causes issues (see [failure](https://github.com/conda/conda/actions/runs/5019301959/jobs/8999646050?pr=12693#step:3:529) occurring in recent test refactor https://github.com/conda/conda/pull/12693).

Previously the workflow was as follows:

1. Initialize `conda.base.context` at startup/import (this occurs before CLI argument parsing so neither `--name` nor `--prefix` are "defined" yet irrespective of what the user provided on the CLI)
    - since `--name` or `--prefix` aren't "defined" yet, `$CONDA_PREFIX` is left as is
2. Reinitialize `conda.base.context` with the parsed arguments (e.g., `conda create --prefix PREFIX`)
    - when the user invokes a conda subcommand that requires `--name` or `--prefix` (e.g., `create`, `install`, etc.), `--name` or `--prefix` become "defined" and hence `$CONDA_PREFIX` is set
3. Attempt to reset the context but this time without a subcommand that requires `--name` or `--prefix` (e.g. `conda clean`)
    - `$CONDA_PREFIX` remains unchanged even though we've done a reset

This fixes the issue by no longer setting `$CONDA_PREFIX` at all. Instead we use a custom implementation of `os.path.expandvars` (inspired by `string.Template`) that accepts both the path to expand and a mapping of variables to use during expansion:

```python
from conda.common.configuration import custom_expandvars

custom_expandvars("path/$VAR/somewhere", VAR="to")  # path/to/somewhere
custom_expandvars("path/${VAR}/somewhere", VAR="to")  # path/to/somewhere
custom_expandvars("path/%VAR%/somewhere", VAR="to")  # path/to/somewhere
```

Xref https://github.com/conda/conda/issues/12704

### Checklist - did you ...


kenodegard:
pre-commit.ci autofix

conda/common/configuration.py
```diff
@@ -1347,32 +1316,86 @@ def __init__(cls, name, bases, attr):
...
-    def _set_search_path(self, search_path):
-        self._search_path = IndexedSet(search_path)
-        self._set_raw_data(load_file_configs(search_path))
+    @staticmethod
+    def _expand_search_path(
+        search_path: Iterable[Path | Template | str],
+        **kwargs,
+    ) -> Iterable[Path]:
+        for search in search_path:
+            # use string.Template instead of os.path.expand so additional variables can be passed
```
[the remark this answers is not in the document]
jezdez: There is an issue with using [`string.Template`](https://docs.python.org/3/library/string.html#template-strings) here instead since `os.path.expandvars` [has this in addition](https://docs.python.org/3/library/os.path.html#os.path.expandvars):

> On Windows, `%name%` expansions are supported in addition to `$name` and `${name}`.

In other words, we'll have to somehow retrofit this to `string.Template` if we want to switch to it.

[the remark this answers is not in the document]
jezdez: Maybe this would suffice?

```python
class WindowsTemplate(string.Template):
    delimiter = "%"
```

From the docs:

> Advanced usage: you can derive subclasses of [Template](https://docs.python.org/3/library/string.html#string.Template) to customize the placeholder syntax, delimiter character, or the entire regular expression used to parse template strings. To do this, you can override these class attributes:
> 
> delimiter – This is the literal string describing a placeholder introducing delimiter. The default value is $. Note that this should not be a regular expression, as the implementation will call [re.escape()](https://docs.python.org/3/library/re.html#re.escape) on this string as needed. Note further that you cannot change the delimiter after class creation (i.e. a different delimiter must be set in the subclass’s class namespace).

news/12696-do-not-set-context-CONDA_PREFIX
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Do not mock `$CONDA_PREFIX` when `--name` or `--prefix` is provided. (#12696)
```
travishathaway, line 3: I would probably mention something about the introduction of `Template` to parse environment variables. Could be helpful with tracking down bugs with the release notes in the future.

conda/common/configuration.py
```diff
@@ -1382,27 +1405,30 @@ def _set_argparse_args(self, argparse_args):
...
+        if source in self.raw_data:
+            del self.raw_data[source]
+
+        self.raw_data[source] = ArgParseRawParameter.make_raw_parameters(argparse_args)
+
         self._reset_cache()
         return self
 
-    def _set_raw_data(self, raw_data):
+    def _set_raw_data(self, raw_data: Mapping[Hashable, dict]):
```
dholth, line 1431: `Hashable` not too specific :)

conda/testing/helpers.py
```diff
@@ -93,6 +93,11 @@ def capture_json_with_argv(
         raise
 
 
+@deprecated(
+    "24.3",
+    "24.9",
+    addendum="Use `mocker.patch('conda.base.context.Context.active_prefix')` instead.",
+)
```
dholth, line 100: Heavy handed to reach for mocker directly. Could this function be rewritten to call mocker?

tests/base/test_context.py
```diff
@@ -357,6 +357,8 @@ def test_target_prefix(self):
                 reset_context((), argparse_args=AttrDict(name="blarg", func="create"))
                 assert context.target_prefix == join(envs_dirs[0], "blarg")
 
+        reset_context()
```
dholth, line 360: Is there a helper or fixture that automatically resets context?

tests/cli/test_config.py
```diff
@@ -87,9 +88,12 @@ def test_invalid_yaml():
     )
     try:
         with make_temp_condarc(condarc) as rc:
-            run_command(Commands.CONFIG, "--file", rc, "--add", "channels", "test")
+            try:
+                run_command(Commands.CONFIG, "--file", rc, "--add", "channels", "test")
+            except ScannerError as err:
```
dholth, line 93: _scanners_

conda/common/configuration.py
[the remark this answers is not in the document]
kenodegard: Unfortunately `string.Template` doesn't support `%...%` without extra work. Further more `string.Template` treated escaped delimiters differently than `os.path.expandvars` does:

```
# on all OSes
assert string.Template("$$").safe_substitute() == "$"

# on Unix
assert os.path.expandvars("$$") == "$$"

# on Windows
assert os.path.expandvars("$$") == "$"
```

Because of this I ended up implementing a new function that properly handles `$VAR`, `${VAR}`, and `%VAR%` and leaves all other patterns as-is.

conda/testing/helpers.py
kenodegard: yes, just didn't want to add even more new stuff into this PR

tests/base/test_context.py
kenodegard: this reset may actually no longer be necessary, let me try to remove it

kenodegard: yup, it's no longer needed given the other changes done in `context.__init__`

travishathaway (approved):

jezdez (approved):

PR #12707, opened 2023-05-23
Add support for sha256 field in matchspec

### LINKED ISSUE #12654: Allow sha256 to be included in match spec
### Checklist

### What is the idea?

Currently a package match specification can include the md5 of the package to fully specify the package, for example `python[md5=bd5ef8b4531bcd58ac2c00393a7d8c3f]`.  It would be nice if the sha256 hash could be supplied instead.

### Why is this needed?

This would be beneficial as md5 is no longer considered a safe hash function and replacing it with another hash is often recommended. 

### What should happen?

_No response_

### Additional Context

_No response_

jaimergp:
### Description

Closes #12654.

This was simpler than expected!

### Checklist - did you ...

### DIFF
--- a/docs/source/user-guide/concepts/pkg-search.rst
+++ b/docs/source/user-guide/concepts/pkg-search.rst
@@ -40,7 +40,7 @@ example image. The search below will return the same list of packages as the sta
 
    $ conda search "numpy[channel=conda-forge, subdir=linux-64, version=1.17.*, build=py38*]"
 
-At the time of writing, this notation supports the following key-value pairs:
+This notation supports the following key-value pairs:
 
 .. program-output:: cd .. && python -c 'import os; import sys; src_dir = here = os.path.abspath(os.path.dirname("../setup.py")); sys.path.insert(0, src_dir); import conda.models.match_spec as M; print("-", "\n- ".join([f"{field:20} # validated via {M._implementors.get(field, M.ExactStrMatch).__name__}" for field in sorted(M.MatchSpec.FIELD_NAMES)]))'
    :shell:
--- a/docs/source/user-guide/concepts/pkg-search.rst
+++ b/docs/source/user-guide/concepts/pkg-search.rst
@@ -40,6 +40,11 @@ example image. The search below will return the same list of packages as the sta
 
    $ conda search "numpy[channel=conda-forge, subdir=linux-64, version=1.17.*, build=py38*]"
 
+At the time of writing, this notation supports the following key-value pairs:
+
+.. program-output:: cd .. && python -c 'import os; import sys; src_dir = here = os.path.abspath(os.path.dirname("../setup.py")); sys.path.insert(0, src_dir); import conda.models.match_spec as M; print("-", "\n- ".join([f"{field:20} # validated via {M._implementors.get(field, M.ExactStrMatch).__name__}" for field in sorted(M.MatchSpec.FIELD_NAMES)]))'
+   :shell:
+
 Key-value pair notation can be used at the same time as standard notation.
 
 .. code-block:: none
--- a/docs/source/user-guide/concepts/pkg-search.rst
+++ b/docs/source/user-guide/concepts/pkg-search.rst
@@ -42,7 +42,7 @@ example image. The search below will return the same list of packages as the sta
 
 This notation supports the following key-value pairs:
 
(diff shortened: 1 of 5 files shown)


beeankha (dismissed):

jaimergp:
Added a snippet to [autodocument](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/pkg-search.html) the accepted fields (and their validators). Works locally and mimics what we do for the Configuration page.

This is how it looks:

image


(`sphinx.programoutput` assumes you want a code block, no way around that :/)

docs/source/user-guide/concepts/pkg-search.rst
```diff
@@ -40,6 +40,11 @@ example image. The search below will return the same list of packages as the sta
 
    $ conda search "numpy[channel=conda-forge, subdir=linux-64, version=1.17.*, build=py38*]"
 
+This notation supports the following key-value pairs:
+
+.. program-output:: cd .. && python -c 'import os; import sys; src_dir = here = os.path.abspath(os.path.dirname("../setup.py")); sys.path.insert(0, src_dir); import conda.models.match_spec as M; print("-", "\n- ".join([f"{field:20} # validated via {M._implementors.get(field, M.ExactStrMatch).__name__}" for field in sorted(M.MatchSpec.FIELD_NAMES)]))'
```
jezdez, line 45: Okay, `setup.py` doesn't exist anymore, should we use `pyproject.toml` to locate the conda folder instead? Why is this needed in the first place?

tests/models/test_match_spec.py
```diff
@@ -1216,32 +1216,37 @@ def test_build_number_merge(self):
...
+            specs = (
+                MatchSpec(f"python[{hash_type}=FFBADD11]"),
+                MatchSpec("python=1.2.3"),
+                MatchSpec(f"python[{hash_type}=ffbadd11]"),
+            )
+            with pytest.raises(ValueError):
+                MatchSpec.merge(specs)
+
+    def test_hash_merge_wo_name(self):
+        for hash_type in ("md5", "sha256"):
```
jezdez, line 1239: This would be an ideal case for a parameterized pytest test, but may be out of scope.

OTOH I think we could port all tests in this module to pytest? /cc @kenodegard WDYT?

jaimergp: Yea I tried that first but methods are not parameterizeable directly :/

docs/source/user-guide/concepts/pkg-search.rst
jaimergp: I just copied things from the configuration.rst page. I don't even think `setup.py` is being used at all there; maybe it was just a documentation thing (so we know it's the folder containing setup.py?). `src_dir = os.path.abspath("..")` should be enough.

And I guess the whole thing is to have development `conda` in PYTHONPATH if it's not installed, but maybe this is not needed anymore in the current workflow?

jezdez: I think it's not needed anymore since we've modernized the documentation building last year, but I could be mistaken. Does it still work if you remove it?

tests/models/test_match_spec.py
jezdez: If you feel adventurous I think it should be easy to convert to function-based tests.

jezdez (commented):
LGTM, with comments

tests/models/test_match_spec.py
jaimergp: Let's do that in a follow-up PR to keep this changeset cleaner. I can see how other implementations might point to this one for reference.

docs/source/user-guide/concepts/pkg-search.rst
jaimergp: It works as long as `conda` is installed in the docs environment, in editable mode. Otherwise you get this error, silently:

image

jaimergp: I guess ReadTheDocs has been configured to install the project in the build env, so I'll remove it.

jezdez (approved):

jezdez:
The error is a legitimate problem in how the signature verification tests are executed if conda-content-trust is broken: https://github.com/conda/conda/issues/12755

docs/source/user-guide/concepts/pkg-search.rst
jezdez: FWIW there was a regression about this after all: #12768

Issue #12593, opened 2023-04-14
Add a "pre-command" hook to conda


travishathaway:
### What is the idea?

To allow plugin authors to add extra functionality to conda, we would like to add new plugin hook that gets invoked before calling a conda sub-command (e.g. create, install, etc.).

### Why is this needed?

This is needed to give plugin authors the flexibility they need when adding new features to conda. A good example would be interrupting program to start-up to ask for and store credentials for a login protected channel.

### What should happen?

- [x] Do some preliminary research on other examples like this that use pluggy (regarding the hook call order)
- [ ] A new `pre_command` plugin hook should be added to conda
- [ ] Tests for this new hook will be written and execute a working example of it
- [ ] Documentation will be updated to describe this new plugin hook

### Additional Context

More information about the technical specifications can be found in this CEP proposal:

- https://github.com/conda-incubator/ceps/pull/45

travishathaway:
One question that came up today while working on the `pre_command` plugin hook was:

- What happens if the underlying pre-command action function raises an error?
  - [ ] Log it and continue on with normal program flow
  - [ ] Treat it like any other error in conda and route it through our exception handler?

I'm leaning towards the first. In my opinion, the authors should be doing their due diligence to catch and handle errors and if one happens to slip through the cracks, it shouldn't negatively affect the rest of the command.

If anyone has any other ideas of how to handle this, please let me know.

jezdez:
> One question that came up today while working on the `pre_command` plugin hook was:
> 
> * What happens if the underlying pre-command action function raises an uncaught exception?
>   
>   * [ ]  Log it and continue on with normal program flow
>   * [ ]  Treat it like any other error in conda and route it through our exception handler?
> 
> I'm leaning towards the first. In my opinion, the authors should be doing their due diligence to catch and handle errors and if one happens to slip through the cracks, it shouldn't negatively affect the rest of the command.
> 
> If anyone has any other ideas of how to handle this, please let me know.

We can’t assume plugins to always work but also can’t deduce if the successful run of a plugin might be required for a user or not. In other words we need to fail loudly and not ignore a bad run and just move on (by default).

The user has made a choice about installing and configuring the pre_command plugin, so they’ll need to be able to debug it as well. We could add an option to ignore the result of a plug-in optionally, but that’s maybe over engineering.

Issue #12138, opened 2022-12-01
Lock Environments


JucaVaz:
### What is the idea?

What if there was a way to "lock" specific conda environments and don't let any packages be installed/updated? For example, lock the base environment, so in case I forgot to activate my conda environment it would avoid me from polluting the base environment with packages unless I disable that specific feature. Throwing at me some kind of error? Looks like would be a nice feature to have, instead of doing `conda install --revision` every time I mess up a package installation.

travishathaway:
Hi @JucaVaz,

Thanks for this feature request. This is a very sensible idea for a new ability in conda, but is unlikely to actually be added to the core of conda. Fortunately, this is a great for a plugin. I want to keep this open to highlight a potential use case for a future plugin.

As it stands, we currently do not have the appropriate plugin hook to make this work, but in a future release, it is likely that we will add a plugin hook that allow plugin authors to run code before a command runs. After adding this hook, writing a plugin for this type of locking behavior would actually be relatively simple.

I am going to add our plugin label to this issue so that we can keep track of it.

Thank you very much for this idea.

--
conda team ✌️ 🐍

bernt-matthias:
Has there been any progress on this? I just recently tried to (re)create all my environments in a parallel way and failed horribly (with lots of broken packages). 

Also we are using conda environment more and more in multi-user setups .. where it would be nice if they would be able to use a shared package cache.

bernt-matthias:
Maybe a duplicate of https://github.com/conda/conda/issues/6742

travishathaway:
Hi @bernt-matthias,

So, this feature request doesn't mean locking like in the context of parallel computing, rather it means locking an environment so that you can no longer make changes to it.

This is essentially how it would work:

```
conda envlock -n base
Successfully locked based
```

and then, if I tried to install something to base:

```
conda install -n base python=3.11
Cannot install to environment "base" because it is currently locked
```

I hope that makes a little more sense now 😅.

bernt-matthias:
Thanks for the explanation @travishathaway . This sounds super useful as well :). Actually for my environments I always `chmod -R ugo-w ENVDIR` to have a similar behavior (wo a pretty error message).

But you are right, I was looking for the wrong thing in this issue.

travishathaway:
@JucaVaz,

I'm happy to report that I have released a plugin that essentially does what you want. Please check out the README on this repository to learn more: https://github.com/travishathaway/conda-guard

To install, run this command:

```
[...]

Issue #12149, opened 2022-12-03
Plugin code executed in core modules should have extra error handling


mbargull:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

With the new plugin system being used in `conda.core.index`, we could run into problematic scenarios in case of buggy plugins. In the future, if some package which registers itself as a plugin is installed, its code can bring `conda install`/`conda remove` etc. down. Meaning if a plugin (or a combination of plugins) cause some errors, they effectively can't be uninstalled with the same `conda` installation:

Two scenarios:

1. If some arbitrary error is thrown in `core.index._supplement_index_with_system` via `context.plugin_manager.get_hook_results("virtual_packages")`, it is currently not caught early and thus breaks `core.index.get_reduced_index`.
2. If two identically named virtual packages are created by the same or multiple plugins, this currently throws an uncaught `PluginError` which again breaks `core.index.get_reduced_index`.

Since `core.index.get_reduced_index` is a core functionality used from `core.solve`, a breakage there makes any package modifications impossible, esp. including removal of the hypothetical error-producing plugin package.

### Conda Info

```shell
Source as of conda=22.11.0
```

### Tasks

- [x] #12460
- [x] #12748
- [x] #12660

travishathaway:
This issue came up in a separate pull request:

- https://github.com/conda/conda/pull/12712#discussion_r1206369287

I did little work exploring how we may be able to disable plugins and came across these helpful methods on the PluginManager object:

- https://pluggy.readthedocs.io/en/stable/#blocking

@jezdez is suggested that we simply go with a `--no-plugins` flag for all commands. Using this would allow users to regain operation of a broken install command. I endorse this approach 😎.

jezdez:
Closing since this was done.

PR #12712, opened 2023-05-25
Pre command plugin hook

travishathaway:
Fixes: #12593 

This pull request adds a new `pre_command` plugin hook to conda. This will give plugin authors the added flexibility of being able to "interrupt" normal command flow by execute arbitrary code before a command is run.

To illustrate how all of this could be used, I created the following project:

- https://github.com/travishathaway/conda-envlock

This project was inspired by the following feature request:

- https://github.com/conda/conda/issues/12138

### Checklist


travishathaway:
This pull request also fixes a small bug I found in the documentation. The code samples for the various plugin hooks currently don't show correctly.

Compare this:
- https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/solvers.html

with this:
- https://continuumio-conda-docs--12712.com.readthedocs.build/projects/conda/en/12712/dev-guide/plugins/solvers.html

to see what I'm talking about.

conda/plugins/manager.py
```diff
@@ -159,6 +161,24 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
 
         return backend
 
+    def run_pre_command_hooks(self, command: str, args: Namespace) -> None:
+        """
+        Runs ``CondaPreCommand.action`` functions registered by the ``conda_pre_commands`` hook.
+
+        This method takes special care to handle any exceptions raised during the running of these
+        hooks so that it doesn't interfere with the normal operation of the conda command.
```
jezdez, line 169: Still true?

conda/plugins/manager.py
```diff
@@ -159,6 +161,24 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
...
+
+        :param command: name of the command that is currently being invoked
+        :param args: arguments of the running command
+        """
+        # Load pre-commands available to run
+        pre_command_hooks = self.get_hook_results("pre_commands")
+
+        for pre_command in pre_command_hooks:
+            pre_command = cast(CondaPreCommand, pre_command)
```
jezdez, line 178: Do we need this here?

docs/source/dev-guide/plugins/index.rst
```diff
@@ -113,6 +113,7 @@ API reference
    solvers
    subcommands
    virtual_packages
+   pre_command
```
jezdez, line 116: Let's rename that module to plural

conda/plugins/hookspec.py
```diff
@@ -96,3 +98,30 @@ def conda_virtual_packages():
...
+        """
+        Register pre-commands functions in conda.
+
+        **Example:**
+
+        .. code-block:: python
+
+           from conda import plugins
+
+           PLUGIN_NAME = "custom_plugin"
```
jezdez, line 113:
```suggestion
```

This is otherwise confusing users between "plugin" and individual hook implementations, one plugin is a collection of hook implementations.

conda/plugins/hookspec.py
```diff
@@ -96,3 +98,30 @@ def conda_virtual_packages():
...
+
+
+           def custom_plugin_pre_commands_action(command_name, args):
```
jezdez, line 116:
```suggestion
           def example_pre_command(command, args):
```

conda/plugins/hookspec.py
```diff
@@ -96,3 +98,30 @@ def conda_virtual_packages():
...
+               print("pre-command action")
+
+
+           @plugins.hookimpl
+           def conda_pre_commands():
+               yield CondaPreRun(
+                   name=f"{PLUGIN_NAME}_pre_commands",
+                   action=custom_plugin_pre_commands_action,
+                   run_for={"install", "create"},
+               )
```
jezdez, line 126:
```suggestion
               yield CondaPreRun(
                   name="example-pre-command",
                   action=example_pre_command,
                   run_for={"install", "create"},
               )
```

conda/cli/conda_argparse.py
```diff
@@ -123,6 +124,9 @@ def do_call(args, parser):
 
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
+    command_name = relative_mod.replace(".main_", "")
+
+    context.plugin_manager.run_pre_command_hooks(command_name, args)
```
jezdez, line 129:
```suggestion
    command = relative_mod.replace(".main_", "")

    context.plugin_manager.run_pre_command_hooks(command, args)
```

conda/plugins/manager.py
```diff
@@ -159,6 +161,24 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
 
         return backend
 
+    def run_pre_command_hooks(self, command: str, args: Namespace) -> None:
```
jezdez, line 164: I think this function would be better suited for the subject code where commands are called, than the plugin manager since the plugin manager exists to query plugin metadata, not handle actions of hook implementations. Or maybe you could split it between a function for getting a list of `pre_commands` and a function to call them in the CLI code?

docs/source/dev-guide/plugins/pre_command.rst
jezdez, line 1: Let's rename to pre_commands.rst

docs/source/dev-guide/plugins/pre_command.rst
```diff
@@ -0,0 +1,16 @@
+========================
+Pre-command
+========================
```
jezdez, line 3:
```suggestion
============
Pre-commands
============
```

docs/source/dev-guide/plugins/pre_command.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Conda command start-up can be extended with the ``conda_pre_commands`` plugin hook.
```
jezdez, line 5:
```suggestion
Conda commands can be extended with the ``conda_pre_commands`` plugin hook.
```

news/12712-pre-command-plugin-hook
```diff
@@ -0,0 +1,7 @@
+### Enhancements
+
+* Adds new "pre_command" hook which allows plugins to interrupt the normal start-up process of specific commands
+
+### Docs
+
+* Documentation for how to use the new "pre_command" plugin hook
```
jezdez, line 7:
```suggestion
* Documentation for how to use the new "pre_commands" plugin hook
```

tests/plugins/test_pre_command.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from unittest import mock
+
```
jezdez, line 4:
```suggestion
```

tests/plugins/test_pre_command.py
```diff
@@ -0,0 +1,70 @@
...
+import pytest
+
+from conda import plugins
+from conda.plugins.types import CondaPreCommand
+
+PLUGIN_NAME = "custom_pre_command"
```
jezdez, line 10:
```suggestion
```

conda/plugins/manager.py
travishathaway: Good catch! Not true anymore.

dholth: OTOH it is a little tricky if you've installed something that breaks the installer, preventing the uninstall of that thing. Do we have a "no plugins" safe mode flag?

conda/plugins/manager.py
jezdez: Main reason: separation of concerns between plugin code and CLI code for example.

conda/plugins/manager.py
travishathaway: This was what I originally had in that method:

```python
        # Load pre-commands available to run
        pre_command_hooks = self.get_hook_results("pre_commands")

        for pre_command in pre_command_hooks:
            if command in pre_command.run_for:
                try:
                    pre_command.action(args)
                except Exception as exc:
                    error, *_ = traceback.format_exception_only(type(exc), exc)
                    error = error.strip()

                    log.error(
                        f'Pre-command action for the plugin "{pre_command.name}" failed with: '
                        f"{error}"
                    )
```

It would basically just catch everything and then log it. Should we consider re-introducing this. @dholth's concern about not being able to uninstall a faulty plugin does concern me 😅.

Another way around this would be to introduce a new setting called called `disabled_plugins`. You would only need to de-activate faulty plugins. We could also perhaps automatically disable a plugin if it were to raise an uncaught exception somewhere. Somewhere in the error message, we could tell the user what happened.

Definitely a can of worms but worth discussing.

travishathaway: @jezdez and @dholth,

I was able to update this pull request today with the addition of a `disabled_plugins` setting. Please let me know what you think and whether this could be a solution to the problem of "misbehaving plugins".

conda/cli/main.py
```diff
@@ -63,15 +63,17 @@ def generate_parser(*args, **kwargs):
 def main_subshell(*args, post_parse_hook=None, **kwargs):
     """Entrypoint for the "subshell" invocation of CLI interface. E.g. `conda create`."""
     # defer import here so it doesn't hit the 'conda shell.*' subcommands paths
+    from ..base.context import context
     from .conda_argparse import generate_parser
 
-    args = args or ["--help"]
+    # This needs to be called to give the argument parser access to context variables
+    context.__init__()
```
travishathaway, line 70: This is weird 😭. Any suggestions of how to change it? I don't think the performance penalty is that big, so it might not be that bad to leave in.

conda/plugins/manager.py
jezdez: As I mentioned in person, I don't think we should catch all errors of plugins and ignore them and move on. The functioning of a plugin may be required for the user (imagine a plugin setting auth credentials or something), so it's important to raise any issue with similar to other errors that conda shows to users. Plugins are part of the conda runtime environment after all and need to be treated that way as well.

Let's add a global `--no-plugins` CLI options as the escape hatch Daniel alluded to. And separate that into an own PR related to https://github.com/conda/conda/issues/12149 to not make review harder here.

I'm a little concerned that `disabled_plugins`, is going to be a painful user interface for end users, since there is no presedence so far to listing plugin names (will it be plugin names, or dotted paths to implementation hooks?). Basically, it's too low-level for regular users IMO.

conda/cli/main.py
jezdez: Please don't do that, that's the fundamental problem of the CLI and context system, we can't change the stack order of the CLI/config system, for risk of triggering side-effects that are related to config loading.

Let's skip the `disabled_plugins` config option in favor of CLI options `--no-plugins` and maybe `--disable-plugin=my-example-plugin --disable-plugin=conda-libmamba-solver`

conda/plugins/manager.py
travishathaway: I summed up our discussion on https://github.com/conda/conda/issues/12149 . So, let's move that conversation over there for now. I am considering this thread resolved.

conda/plugins/hookspec.py
```diff
@@ -96,3 +98,28 @@ def conda_virtual_packages():
...
+           from conda import plugins
+
+
+           def example_pre_command(command, args):
+               print("pre-command action")
+
+
+           @plugins.hookimpl
+           def conda_pre_commands():
+               yield CondaPreRun(
```
jezdez, line 120:
```suggestion
               yield CondaPreCommand(
```

conda/plugins/manager.py
```diff
@@ -159,6 +161,24 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
 
         return backend
 
+    def get_pre_command_hook_actions(
```
jezdez, line 164: Technically a generator shouldn't use “get” but this:
```suggestion
    def yield_pre_command_hook_actions(
```

tests/plugins/test_manager.py
```diff
@@ -10,6 +10,7 @@
 from conda.core import solve
 from conda.exceptions import PluginError
 from conda.plugins import virtual_packages
+from conda.plugins.manager import CondaPluginManager
```
jezdez, line 13:
```suggestion
```

docs/source/dev-guide/plugins/index.rst
```diff
@@ -113,6 +113,7 @@ API reference
    solvers
    subcommands
    virtual_packages
+   pre_commands
```
jezdez, line 116:
```suggestion
   pre_commands
   solvers
   subcommands
   virtual_packages
```

conda/cli/conda_argparse.py
```diff
@@ -123,6 +129,12 @@ def do_call(args, parser):
 
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
+    command = relative_mod.replace(".main_", "")
+
+    # Run the pre_command actions
+    actions = context.plugin_manager.get_pre_command_hook_actions(command, args)
+    tuple(action() for action in actions)
+
     return getattr(module, func_name)(args, parser)
```
jezdez, line 139:
```suggestion
    def _run_pre_commands(self, command, args) -> None:
         for action in context.plugin_manager.get_pre_command_hook_actions(command, args):
             action(command, args)
```

conda/cli/conda_argparse.py
```diff
@@ -115,6 +115,12 @@ def do_call(args, parser):
     """
     # First, check if this is a plugin subcommand; if this attribute is present then it is
     if getattr(args, "plugin_subcommand", None):
+        # Run the pre_command actions
+        actions = context.plugin_manager.get_pre_command_hook_actions(
+            args.plugin_subcommand.name, args
+        )
+        tuple(action() for action in actions)
```
jezdez, line 122:
```suggestion
        # Run the pre_command actions
        _run_pre_commands(args.plugin_subcommand.name, args)
```

conda/cli/conda_argparse.py
```diff
@@ -123,6 +129,12 @@ def do_call(args, parser):
 
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
+    command = relative_mod.replace(".main_", "")
+
+    # Run the pre_command actions
+    actions = context.plugin_manager.get_pre_command_hook_actions(command, args)
+    tuple(action() for action in actions)
```
jezdez, line 136:
```suggestion
	_run_pre_commands(args.plugin_subcommand.name, args)
```

conda_env/cli/main.py
```diff
@@ -45,6 +45,13 @@ def do_call(args, parser):
     # func_name should always be 'execute'
     from importlib import import_module
 
+    # Run the pre_command actions
+    command = relative_mod.replace(".main_", "")
+    actions = context.plugin_manager.get_pre_command_hook_actions(
+        f"env_{command}", args
+    )
+    tuple(action() for action in actions)
```
jezdez, line 53:
```suggestion
    _run_pre_commands(f"env_{command}", args)
```
But this also needs an import above for `conda.cli.conda_argparse._run_pre_commands`

conda/cli/conda_argparse.py
jezdez: Important bit is that you hadn't passed command name and args, and there is a bit of repetition going on that we can move into `_run_pre_commands` to be able to have just one code path for running the pre commands here.

conda/plugins/manager.py
```diff
@@ -159,6 +161,24 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
...
+
+        :param command: name of the command that is currently being invoked
+        :param args: arguments of the running command
+        """
+        # Load pre-commands available to run
+        pre_command_hooks = self.get_hook_results("pre_commands")
+
+        for pre_command in pre_command_hooks:
+            if command in pre_command.run_for:
+                yield functools.partial(pre_command.action, command, args)
```
jezdez, line 180: Does the partial give us anything here compared to just passing command and args? Partials are annoying to debug since they have the parameters baked into them, while passing the variables doesn't take up more memory and enables to show up as local variables during debugging.

jezdez (commented):
Some more design comments, I think this is getting SUPER close!

jezdez (approved):

conda/plugins/manager.py
```diff
@@ -169,9 +184,7 @@ def get_plugin_manager() -> CondaPluginManager:
     plugin_manager = CondaPluginManager()
     plugin_manager.add_hookspecs(CondaSpecs)
     plugin_manager.load_plugins(
-        solvers,
-        *virtual_packages.plugins,
-        *subcommands.plugins,
+        solvers, *virtual_packages.plugins, *subcommands.plugins
```
kenodegard, line 187:
```suggestion
        solvers,
        *virtual_packages.plugins,
        *subcommands.plugins,
```

news/12712-pre-command-plugin-hook
```diff
@@ -0,0 +1,7 @@
+### Enhancements
+
+* Adds a new "pre_commands" hook which allows plugins to run code before specific conda commands.
```
kenodegard, line 3:
```suggestion
* Add a new "pre_commands" hook allowing plugins to run code before a `conda` subcommand. (#12712)
```

news/12712-pre-command-plugin-hook
```diff
@@ -0,0 +1,7 @@
...
+
+### Docs
+
+* Documentation for how to use the new "pre_commands" plugin hook
```
kenodegard, line 7:
```suggestion
* Add documentation for how to use the new "pre_commands" plugin hook. (#12712)
```

conda/cli/conda_argparse.py
```diff
@@ -123,9 +124,21 @@ def do_call(args, parser):
...
     return getattr(module, func_name)(args, parser)
 
 
+def _run_pre_command_hooks(command: str, args) -> None:
+    """
+    Helper function used to gather applicable pre_command hook functions
+    and then run them.
+    """
+    actions = context.plugin_manager.yield_pre_command_hook_actions(command)
+    tuple(action(command, args) for action in actions)
```
kenodegard, line 139: I prefer expanded for-loops instead of list comprehensions if we arent returning, storing, or otherwise using the results, I find it to be less error prone (e.g., dropping the `tuple` here would make this a generator that never executes)

```suggestion
    for action in actions:
        action(command, args)
```

kenodegard (commented):
elegant!

conda/cli/conda_argparse.py
jezdez: Gave that feedback in my previous review as well 🤣

travishathaway: What can I say. I'm a sucker for one liners 😏... Will change it 👌 .

conda/plugins/manager.py
travishathaway: @kenodegard, I prefer this but `black` does not 😭. Let's just keep what the code formatter wants.

jezdez (approved):

conda/plugins/manager.py
kenodegard: its all about the magic trailing comma: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

Issue #12621, opened 2023-04-27
SPIKE: Research for potential shell plugin hook


kalawac:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Determine the proposed architecture method for plugins that handle, at minimum, conda's `activate`, `deactivate` and `reactivate` processes.

### Why is this needed?

As per #12451:

> Shell interfaces are difficult to support as they differ widely, and the development team has limited experience with many of the exotic shells supported to offer meaningful support.
> Let's rely on the plugin infrastructure to allow plugins that extend the supported shell interfaces. This means that most of the shells supported out of the box today will be moved into standalone plugins, and we will request community support for future development efforts.

An example that ports the desired functionality into a plugin is [here](https://github.com/kalawac/simple-bash-plugin). The current process flow for the activate process can be seen [here](https://github.com/kalawac/conda/tree/activate-flow/conda/shell#readme). 

Currently, there are three core decision points that will require exploration and, ultimately, consensus:
- What are the architectural criteria that plugins should conform to? For example:
    - Should the plugin have a shell script bundled with it?
    - Should conda or the plugin modify the user's shell profile to allow for automatic evaluation?
    - Is it possible and/or desirable to have a plugin that avoids using a shell script at all?
    - Do we need a new plugin hook? If yes, what should that plugin hook do?
- What are the minimum criteria for a plugin's UX?
- How should the plugin be packaged? What is the migration path?
- In simplifying the code, should we keep `hook` and `command` functionality bundled in with `activate`, `deactivate`, and `reactivate`?

### What should happen?

- [x] Research approaches on plugin architecture, in alignment with the questions and decision points outlined above
- [x] Create document to track arguments for and against each approach
- [x] Create prototypes to better understand UX considerations for each approach

### Additional Context

I’m currently considering the following approaches:

- Inject code logic into the user’s shell profile that will allow the statements printed by the python subcommand to be evaluated (as `conda init` does now)
    - Would likely require a new plugin hook implementation that would actually yield the relevant shell logic and make the changes
- Use modules and functions available in more recent versions of Python (e.g., argparse, [os.execve](https://docs.python.org/3/library/os.html#os.execve)) to simplify the code logic and evaluate the shell commands without needing to exit Python (that is, eliminate the need for separate shell scripts) and implement the plugins as a pure Python subcommand
- Continue to use shell scripts: each new plugin would require the subcommand and the related shell script.
[...]

PR #12721, opened 2023-05-26
Shell Plugin Prototypes

kalawac:
### Description
Prototype plugins for spike issue #12621. The prototypes demonstrate three proposed architectural methods for plugins that handle, at minimum, conda's `activate`, `deactivate` and `reactivate` processes. See also the related gist with [discussion points and architectural considerations for each prototype scenario](https://gist.github.com/kalawac/8a05141237a59f0f4e9830096704eb4f).

### Checklist - did you ...

- ~~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- ~~Add / update outdated documentation?~~


conda/plugins/shells/posix_activate.py
jezdez, line 1: Could you move these into subdirectories please, like `conda/plugins/shells/posix/activate.py` etc?

kalawac: OK, will do later today. These are all POSIX prototypes for now.

conda/plugins/shells/posix/posix_activate.py
```diff
@@ -0,0 +1,224 @@
...
+        help="Reactivate the current conda environment, updating environment variables",
+    )
+    reactivate.add_argument(
+        "--dev", action="store_true", default=False, help=argparse.SUPPRESS
+    )
+
+    try:
+        args = parser.parse_args(argv)
+    except BaseException:
+        # avoid evaluation of help strings
```
kenodegard, line 172: we want to keep exceptions targeted so we don't end up catching too much:

```suggestion
    except SystemExit:
        # SystemExit: help blurb was printed, intercepting SystemExit(0) to avoid
        # evaluation of help strings by shell interface
```

conda/plugins/shells/posix/posix_activate.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+
+
+@hookimpl
+def conda_shell_plugins():
+    yield CondaShellPlugins(
+        name="posix_plugin_current_logic",
+        summary="Plugin for POSIX shells: handles conda activate, deactivate, and reactivate",
+        activator=PosixPluginActivator,
+    )
```
kenodegard, line 224: these plugins for the `conda_shell_plugins` hook aren't used anywhere (yet)

conda/plugins/shells/posix/posix_activate.py
```diff
@@ -0,0 +1,224 @@
...
+
+            if args.stack is None:
+                self.stack = context.auto_stack and context.shlvl <= context.auto_stack
+            else:
+                self.stack = args.stack
+
+        return
+
+
+def get_parsed_args(argv: list[str]) -> argparse.Namespace:
```
kenodegard, line 100: Since these are identical for all 3 prototypes can we move this into a new `common.py` submodule? will help make this easier to review

conda/plugins/shells/posix/posix_os_exec_shell.py
```diff
@@ -0,0 +1,277 @@
...
+    )
+    activate.add_argument(
+        "--dev", action="store_true", default=False, help=argparse.SUPPRESS
+    )
+
+    commands.add_parser("deactivate", help="deactivate the current environment")
+    commands.add_parser(
+        "reactivate",
+        help="reactivate the current environment, updating environment variables",
+    )
```
kenodegard, line 156: these are duplicates with what is defined below and can be removed:

```suggestion
```

conda/plugins/shells/posix/posix_activate.py
kalawac: Yes, the gist indicates that the hook is just a placeholder for the moment. Eventually, there'll be some logic specific to the shell plugins there. I thought it was important to indicate that there probably would be a hook of some kind.

conda/plugins/shells/posix/posix_activate.py
kalawac: `get_parsed_args` actually has a small difference across the three (the version in `posix_activate.py` sets `self.command` while the other two don't. I can look for other logic that's common across all three and extract it, though.

conda/plugins/shells/posix/posix_os_exec_shell.py
kalawac: This happened during the rebase process; I missed it during the manual merge. Now fixed.

PR #12725, opened 2023-05-26
Deprecate `conda.plugins.subcommands.doctor.cli.get_prefix`

kenodegard:
### Description

`conda.plugins.subcommands.doctor.cli.get_prefix` duplicates existing functionality that is solved by doing one of the following:

```python
from conda.base.context import context, determine_target_prefix
args = <argparse parsing>

# option 1
prefix = determine_target_prefix(context, args)

# option 2
prefix = context.target_prefix
```

### Checklist - did you ...

### DIFF
--- a/conda/plugins/subcommands/doctor/cli.py
+++ b/conda/plugins/subcommands/doctor/cli.py
@@ -4,10 +4,9 @@ from __future__ import annotations
 
 import argparse
 
-from ....base.context import context, locate_prefix_by_name
-from ....cli.common import validate_prefix
+from ....base.context import context
 from ....cli.conda_argparse import add_parser_prefix
-from ....exceptions import CondaEnvException
+from ....deprecations import deprecated
 from ... import CondaSubcommand, hookimpl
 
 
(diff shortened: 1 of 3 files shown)


conda/plugins/subcommands/doctor/cli.py
```diff
@@ -28,34 +27,21 @@ def get_parsed_args(argv: list[str]) -> argparse.Namespace:
...
-        return validate_prefix(args.prefix)
-
-    if context.active_prefix:
-        return context.active_prefix
-
-    raise CondaEnvException(
-        "No environment specified. Activate an environment or specify the "
-        "environment via `--name` or `--prefix`."
-    )
+    context.__init__(argparse_args=args)
```
travishathaway, line 34: I think we can leave this out. It should be the caller's responsibility to properly initialize context.

Not that it really matters though. Hopefully no one else uses this function 🙄.

travishathaway (changes requested):
Just a minor request. Otherwise, it looks good.

conda/plugins/subcommands/doctor/cli.py
kenodegard: I added this only to ensure that this backport results in the same value as before (whether or not the caller initialized the context)

but yeah this function has existed for only a short moment so theres low risk in it being used elsewhere

travishathaway: Okay, let's just keep it as-is.

travishathaway (approved):

PR #12738, opened 2023-06-01
small linting fixes

kalawac:
### Description
Fixes for the following concerns:
- The type hint for `_Activate.path_conversion` was generating an issue: `Expected parameter type list or ...`
- Lines in the class doc string were one space off of the appropriate spacing.
- Updated hanging indent level in class doc string to be in line with [PEP 8](https://peps.python.org/pep-0008/).

### Checklist - did you ...

- ~~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- ~~Add / update necessary tests?~~
- ~~Add / update outdated documentation?~~


conda/activate.py
```diff
@@ -81,11 +83,11 @@ def get_export_unset_vars(self, export_metavars=True, **kwargs):
...
         :param export_metavars: whether to export `conda_exe_vars` meta variables.
         :param kwargs: environment variables to export.
-                       .. if you pass and set any other variable to None, then it
-                       emits it to the dict with a value of None.
+                        .. if you pass and set any other variable to None, then it
+                        emits it to the dict with a value of None.
 
         :return: A dict of env vars to export ordered the same way as kwargs.
-                 And a list of env vars to unset.
+                And a list of env vars to unset.
```
kenodegard, line 90: Why are we making these changes?

kalawac: This is what I was referencing in the second bullet point. The indentation was one space off.

kenodegard: lets actually follow the standard hanging indent policy instead (see https://stackoverflow.com/a/50455731/5239932)

kalawac: The issue was because the spacing should be in multiples of four. I took care not to change the indent level. There were three spaces instead of four in two cases and five spaces instead of four in one case.

kenodegard: gotcha, lets abide by standard hanging indent policy instead, so only indent the hanging line once:

```
        :param kwargs: environment variables to export.
            .. if you pass and set any other variable to None, then it
            emits it to the dict with a value of None.
        :return: A dict of env vars to export ordered the same way as kwargs.
            And a list of env vars to unset.
```

kalawac: OK. I'll update the pull request and the code to reflect that I'm also making that fix.

kenodegard (approved):

beeankha (approved):

PR #12742, opened 2023-06-02
add environment variable expansion for environment variables

### LINKED ISSUE #12645: Variable expansion when setting environment variables from env.yaml
### Checklist

### What is the idea?

at present writing:

```
variables:
  PARAMS: ${CONDA_PREFIX}/share/params
```

yeilds

```
$ echo $PARAMS
${CONDA_PREFIX}/share/params
```

it would be great if it could yeild something like:

```
/home/ciaran/mambaforge/envs/env/share/params
```

### Why is this needed?

This would allow the user to set environment variables that point to files in the conda environment, without requiring the CONDA_PREFIX to be hardcoded in. 

This would allow the environment to be shareable, whilst still allowing the environment variable to be set (my CONDA_PREFIX is /home/ciaran/mambaforge, someone elses CONDA_PREFIX could be /home/OTHERNAME/mambaforge, and so my environment variables do not travel well)

### What should happen?

the user should be able to rely on variable expansion within the variables section of the yaml - i.e. writing ${CONDA_PREFIX}/share/params leads to the variable being set in conda with ${CONDA_PREFIX} replaced by their CONDA_PREFIX

### Additional Context

_No response_

rafaelmartins:
### Description

This closes #12645.

### Checklist - did you ...

### DIFF
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -164,7 +164,7 @@ def from_yaml(yamlstr, **kwargs):
         for key, value in kwargs.items():
             data[key] = value
     _expand_channels(data)
-    _expand_envvars(data)
+    _expand_variables(data)
     return Environment(**data)
 
 
@@ -175,7 +175,7 @@ def _expand_channels(data):
     ]
 
 
-def _expand_envvars(data):
+def _expand_variables(data):
     """Expands enviroment variables for the variables found in the yaml data"""
     variables = {}
     for key, value in data.get("variables", {}).items():
--- a/conda_env/env.py
+++ b/conda_env/env.py
@@ -164,6 +164,7 @@ def from_yaml(yamlstr, **kwargs):
         for key, value in kwargs.items():
             data[key] = value
     _expand_channels(data)
+    _expand_envvars(data)
     return Environment(**data)
 
 
@@ -172,6 +173,14 @@ def _expand_channels(data):
     data["channels"] = [
         os.path.expandvars(channel) for channel in data.get("channels", [])
     ]
+
+
+def _expand_envvars(data):
+    """Expands enviroment variables for the variables found in the yaml data"""
+    variables = {}
+    for key, value in data.get("variables", {}).items():
+        variables[key] = os.path.expandvars(value)
+    data["variables"] = variables
 
 
 def from_file(filename):
--- a/news/12645-add-environment-variable-expansion-for-variables
+++ b/news/12645-add-environment-variable-expansion-for-variables
@@ -1,0 +1,19 @@
+### Enhancements
+
+* allow the use of environment variables for variables defined in environment.yaml  (#12645)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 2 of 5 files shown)


rafaelmartins:
> We require contributors to sign our [Contributor License Agreement](https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement) and we don't have one on file for @rafaelmartins.
> 
> In order for us to review and merge your code, please e-sign the [Contributor License Agreement PDF](https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement). We then need to manually verify your signature, merge the PR ([conda/infrastructure#773](https://github.com/conda/infrastructure/pull/773)), and ping the bot to refresh the PR.

signed

conda_env/env.py
```diff
@@ -164,6 +164,7 @@ def from_yaml(yamlstr, **kwargs):
         for key, value in kwargs.items():
             data[key] = value
     _expand_channels(data)
+    _expand_envvars(data)
```
jezdez, line 167:
```suggestion
    _expand_variables(data)
```

conda_env/env.py
```diff
@@ -174,6 +175,14 @@ def _expand_channels(data):
     ]
 
 
+def _expand_envvars(data):
```
jezdez, line 178:
```suggestion
def _expand_variables(data):
```

news/12645-add-environment-variable-expansion-for-variables
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* allow the use of environment variables for variables defined in environment.yaml  (#12645)
```
jezdez, line 3:
```suggestion
* Allow the use of environment variables for `variables` defined in environment.yml  (#12645)
```

jezdez (commented):
As mentioned in https://github.com/conda/conda/pull/12074 I don't think we should pile on `env update` like this, this is broadening the scope of the config parsing differences between different parts of conda, which is opposite to what we should be doing.

IOW, this really needs a refactoring.

https://github.com/conda/conda/pull/12074#issuecomment-1370642952 is an important insight:
> [..] but unfortunately there exists multiple classes in conda for doing this. In conda, there is both conda and conda_env and both of these modules have their own classes and functions for parsing configuration, specifically YAML files. This is because conda_env used to be its own project and was brought in to the main conda code sometime ago.

https://github.com/conda/conda/issues/11633 remains true and I feel we're working backwards here basically.

rafaelmartins:
> As mentioned in #12074 I don't think we should pile on `env update` like this, this is broadening the scope of the config parsing differences between different parts of conda, which is opposite to what we should be doing.
> 
> IOW, this really needs a refactoring.
> 
> [#12074 (comment)](https://github.com/conda/conda/pull/12074#issuecomment-1370642952) is an important insight:
> 
> > [..] but unfortunately there exists multiple classes in conda for doing this. In conda, there is both conda and conda_env and both of these modules have their own classes and functions for parsing configuration, specifically YAML files. This is because conda_env used to be its own project and was brought in to the main conda code sometime ago.
> 
> #11633 remains true and I feel we're working backwards here basically.

do you have a suggestion of proper design for this?

travishathaway:
@rafaelmartins,

The proper design for this would essentially see us adding conda env abilities (e.g. being about to read directly from an environment file using `conda update`). Because of that, we might deprecate the usage of `conda env` commands in the future (this still isn't a sure thing though).

Therefore, we are currently hesitant about adding new features/abilities to anything under `conda env`. I hope you understand, and I apologize for not communicating that more effectively earlier.

PR #12757, opened 2023-06-06
Add "detect altered packages" health check

ForgottenProgramme:
### Description

This PR adds a new health check "detect altered packages" to `conda doctor`. 
Altered packages are defined as any packages in a conda environment that are no longer in the state as they were when first installed/built. `conda doctor` detects such packages/files by looking for changes in sha256 checksums.

### Checklist - did you ...

**Note for reviewers**
Kindly do not commit directly to this PR. :)


conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +33,36 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+        required_data = data["paths_data"]["paths"]
+
+        for path in required_data:
+            _path = path.get("_path")
+            old_sha256 = path.get("sha256_in_prefix")
+            file_location = f"{prefix}/{_path}"
+            if os.path.isfile(file_location):
+                new_sha256 = generate_sha256_checksum(file_location)
+            else:
+                continue
```
travishathaway, line 58:
```suggestion
            if not os.path.isfile(file_location):
                continue
            new_sha256 = generate_sha256_checksum(file_location)
```

It's usually better to invert these statements to avoid an extra `else` clause if possible.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +33,36 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
 
+def find_altered_packages(prefix: str | Path) -> list[str]:
+    """Finds altered packages"""
+    altered_packages = {}
+
+    def generate_sha256_checksum(filepath) -> str:
+        with open(filepath, "rb") as f:
+            bytes = f.read()
+            hash = hashlib.sha256(bytes).hexdigest()
+            return hash
```
travishathaway, line 44: I think it would be a good idea to still wrap this in a try/except block and handle the potential `OSError` group of errors. Earlier, you ran into a `FileNotFoundError` but there are more you could run into. For example, you could also run into a `PermissionsError`.

To provide a better message to our users and not just blow up in a huge stack trace, you could catch this error and then re-raise it as a `CondaError` (or one of its sub-classes). To see what I'm talking about, checkout my `conda-envlock` project:

https://github.com/travishathaway/conda-envlock/blob/main/envlock/main.py#L85-L90

In conda-envlock, I actually subclass `CondaError` error to create my own `CondaEnvLockError`. You could consider doing this for conda-doctor by making a `CondaDoctorError` error. It would look something like this:

```python
from conda.exceptions import CondaError

class CondaDoctorError(CondaError):
    """
    Error raised when we attempt to perform health checks for a conda environments
    """
```

I hope this makes sense.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,43 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+        try:
+            with open(filepath, "rb") as f:
+                bytes = f.read()
+                hash = hashlib.sha256(bytes).hexdigest()
+                return hash
+        except OSError as err:
+            raise CondaError(
+                f"Could not generate checksum for file {filepath}"
+                + f" because of the following error: {err}."
+            )
```
jezdez, line 52: Please use [`conda.gateways.disk.read.compute_sum`](https://github.com/conda/conda/blob/ce0740ae171b4a2755b8c4f924f830e7cdefedcf/conda/gateways/disk/read.py#L74-L84) instead of implementing an own function for this. It doesn't do the error handling as you're doing here, but you can do that where you call `generate_sha256_checksum`. You'll have to catch `OSError` as well, the raised `PathNotFoundError` is a subclass of `OSError`.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,43 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+
+    prefix = Path(prefix)
+    for file in (prefix / "conda-meta").glob("*.json"):
+        data = json.loads(file.read_text())
```
jezdez, line 56: Let's add error handling for which this loading might fail for some reason.

jezdez (commented):
Just a few nits

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,44 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+    altered_packages = {}
+
+    prefix = Path(prefix)
+    for file in (prefix / "conda-meta").glob("*.json"):
+        try:
+            data = json.loads(file.read_text())
+        except Exception as e:
+            raise CondaError(
+                f"Could not load the json file {file}"
+                + f" because of the following error: {e}."
```
kenodegard, line 49: Adjoining strings are auto concatenated and are technically faster/cheaper (they're concatenated during compile time, not during runtime). It's also safer to leave spaces at the end of strings instead of the front (similar to how we leave newlines at the end of the string).

```suggestion
                f"Could not load the json file {file} "
                f"because of the following error: {e}."
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,44 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+            )
+        required_data = data["paths_data"]["paths"]
+
+        for path in required_data:
+            _path = path.get("_path")
+            if _path is None:
+                continue
+            old_sha256 = path.get("sha256_in_prefix")
+            file_location = f"{prefix}/{_path}"
+            if not os.path.isfile(file_location):
```
kenodegard, line 59:
```suggestion
            file_location = prefix / _path
            if not file_location.is_file():
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,44 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+                continue
+
+            try:
+                new_sha256 = compute_sum(file_location, "sha256")
+            except OSError as err:
+                raise CondaError(
+                    f"Could not generate checksum for file {file_location}"
+                    + f" because of the following error: {err}."
```
kenodegard, line 67:
```suggestion
                    f"Could not generate checksum for file {file_location} "
                    f"because of the following error: {err}."
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +35,44 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+                )
+
+            if old_sha256 is not None and old_sha256 != new_sha256:
```
kenodegard, line 70: move the `old_sha256 is None` check up before we compute `new_sha256`, no need to compute the sha if we don't have something to compare it with:

```python
_path = path.get("_path")
old_sha256 = path.get("sha256_in_prefix")
if _path is None or old_sha256 is None:
    continue
```

kenodegard (changes requested):

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: thank you for explaining this, @kenodegard !

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: Thank you for explaining this, @kenodegard !

kenodegard (changes requested):
We are already starting to see duplicated patterns between the missing files and altered packages checks, can we abstract out the common stuff? Something like:

```python
def find_missing_files(prefix: Path, package: dict) -> tuple[str, str, str]:
    for file in package.get("files", []):
        if not (prefix / file).exists():
            yield "missing", package["fn"], file

def find_altered_packages(prefix: Path, package: dict) -> tuple[str, str, str]:
    for path in package.get("paths_data", {}).get("paths", []):
        _path = path.get("_path")
        old_sha256 = path.get("sha256_in_prefix")
        if _path is None or old_sha256 is None:
            continue

        fullpath = prefix / _path
        if not fullpath.is_file():
            # this **should** have been identified in find_missing_files
            # doesn't hurt to re-report it, deduplicate later
            yield "missing", package["fn"], _path
            continue

        try:
            new_sha256 = compute_sum(fullpath, "sha256")
        except OSError as err:
            yield "error", package["fn"], err
            continue

        if old_sha256 != new_sha256:
            yield "altered", package["fn"], _path

def display_health_checks(prefix: str, verbose: bool) -> None:
    ...

    checks = {}
    for file in (prefix / "conda-meta").glob("*.json"):
        try:
            package = json.loads(file.read_text())
        except Exception as err:
            checks.setdefault("errors", {}).setdefault(file.stem, set()).append(err)
            continue
        
        try:
            for category, fn, path in chain(
                find_missing_files(prefix, package),
                find_altered_packages(prefix, package),
            ):
                checks.setdefault(category, {}).setdefault(fn, set()).add(path)
        except Exception as err:
            checks.setdefault("errors", {}).setdefault(package["fn"], set()).append(err)

    # print out checks["errors"]
    # print out checks["missing"]
    # print out checks["altered"]
```

ForgottenProgramme:
@kenodegard 

```
if not fullpath.is_file():
            # this **should** have been identified in find_missing_files
            # doesn't hurt to re-report it, deduplicate later
            yield "missing", package["fn"], _path
            continue
```
> this **should** have been identified in find_missing_files

It is; The `find_missing_files()` function identifies the missing files but it does not fix their "missing-ness". So we do need to run the `fullpath.is_file()` check in the `find_altered_files()` function again. 
Alternative would be `find_altered_files()` function skips checking for the missing files reported by `find_missing_files()` function.

ForgottenProgramme:
@kenodegard, thank you for your feedback.

We plan to convert `conda doctor` into a plugin hook as the next step. And that would definitely require a significant refactor of the code anyway. 
I think I would like to keep the code as is for now, and I will for sure take into account your suggestions about using `yield` statements etc when I work on coverting `conda doctor` into a plugin hook and refactor the code in the process.

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -67,6 +111,11 @@ def test_display_health_checks(env_ok: tuple[Path, str, str, str], verbose: bool
         {"CONDA_PREFIX": prefix},
         stack_callback=conda_tests_ctxt_mgmt_def_pol,
     ):
-        display_health_checks(prefix, verbose=verbose)
-        (prefix / bin_doctor).unlink()
-        display_health_checks(prefix, verbose=verbose)
+        display_health_checks(
```
travishathaway, line 114: Can you please add some assertions to this test using the `capsys` fixture available from pytest? (https://docs.pytest.org/en/6.2.x/capture.html#accessing-captured-output-from-a-test-function)

It's nice that we're calling this functions, but we also want to make sure that the output looks like what we expect it to.

travishathaway (dismissed):
It's looking good, but I really want to see some assertions on the last test. Please check out the link I provided to the `capsys` testing fixture from pytest.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,6 +37,43 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+            try:
+                new_sha256 = compute_sum(file_location, "sha256")
+            except OSError as err:
+                raise CondaError(
+                    f"Could not generate checksum for file {file_location} "
+                    f"because of the following error: {err}."
+                )
+
+            if old_sha256 != new_sha256:
+                altered_packages.setdefault(file.stem, []).append(_path)
```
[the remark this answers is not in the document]
ForgottenProgramme: Because we want to print the name of the package in the output, not the json file name.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -30,19 +37,77 @@ def find_packages_with_missing_files(prefix: str | Path) -> dict[str, list[str]]
...
+def find_altered_packages(prefix: str | Path) -> dict[str, list[str]]:
+    """Finds altered packages"""
+    altered_packages = {}
+
+    prefix = Path(prefix)
+    for file in (prefix / "conda-meta").glob("*.json"):
+        try:
+            metadata = json.loads(file.read_text())
+        except Exception:
+            logger.error("Could not load the json file {file}")
```
jezdez, line 49:
```suggestion
        except Exception as exc:
            logger.error("Could not load the json file {file} because of the following errpr: {exc}")
```

jezdez (dismissed):
Just a minor note about elaborating on the error when the json file can't be loaded, otherwise this LGTM now!

kenodegard (commented):
Can we add tests for the missing coverage?

We also need a news snippet for this.

jezdez (approved):

ForgottenProgramme:
@jezdez, thanks for noticing all those unused imports! :D

news/12757-add-detect-altered-packages-health-check
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a new health check to conda doctor that detects altered packages in an environment by comparing expected and actual sha256 checksums.
```
kenodegard, line 3:
```suggestion
* Add a new health check to `conda doctor` detecting altered packages in an environment by comparing expected and computed sha256 checksums. (#12757)
```

kenodegard (dismissed):

kenodegard (approved):

beeankha (approved):

PR #12228, opened 2023-01-10
Generic Plugin Hook Prototype

beeankha:
This is a prototype related to the Generic Plugin Hook CEP proposal: 
- https://github.com/conda-incubator/ceps/pull/45

🚧 🏗👷This is currently a work in progress and is not yet ready for review or merge! 👷 🏗🚧

PR #12758, opened 2023-06-06
Add a "Post Command" Plugin Hook

beeankha:
This pull request adds a new `post_command` plugin hook to conda. This will give plugin authors the added flexibility of being able to execute arbitrary code _after_ a command is run.

This work is based on the initial (now closed) [Generic Plugin Hook prototype PR](https://github.com/conda/conda/pull/12228), and a follow-up to the recently merged [Pre Command Plugin Hook PR](https://github.com/conda/conda/pull/12712).

Please read the [related CEP](https://github.com/conda-incubator/ceps/pull/45/files?short_path=d36507c#diff-d36507c8dee02cb85d2f722e68079d1d1dc2b108fc39f4726843db28fe82eaf0) for more information!


conda/plugins/hookspec.py
```diff
@@ -123,3 +129,28 @@ def conda_pre_commands():
...
+
+        **Example:**
+
+        .. code-block:: python
+
+           from conda import plugins
+
+
+           def example_post_command(command, args):
+               print("post-command action")
```
jezdez, line 146:
```suggestion
           def example_post_command(command, args, result):
               print("post-command action")
```

jezdez: Here's what I meant in our meeting on Monday, I know this is mostly copy-pasting the pre_command code, but thought I catch this early. The post command should also be given the result (error code? subcommand return value?) of the command that was run.

conda/cli/conda_argparse.py
```diff
@@ -115,8 +116,11 @@ def do_call(args, parser):
     """
     # First, check if this is a plugin subcommand; if this attribute is present then it is
     if getattr(args, "plugin_subcommand", None):
-        _run_pre_command_hooks(args.plugin_subcommand.name, args)
-        return args.plugin_subcommand.action(sys.argv[2:])
+        _run_command_hooks("pre", args.plugin_subcommand.name, sys.argv[2:])
+        result = args.plugin_subcommand.action(sys.argv[2:])
+        _run_command_hooks("post", args.plugin_subcommand.name, sys.argv[2:], result)
```
jezdez, line 121: I wonder if we should do some error handling here, following #12149 this seems prudent?
```suggestion
        try:
            _run_command_hooks("post", args.plugin_subcommand.name, sys.argv[2:], result)
        except Exception as exc:
            do_something_with_it(exc)
        finally:
            return result
            
```

conda/cli/conda_argparse.py
```diff
@@ -125,20 +129,26 @@ def do_call(args, parser):
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
     command = relative_mod.replace(".main_", "")
-    _run_pre_command_hooks(command, args)
+    _run_command_hooks("pre", command, args)
+    result = getattr(module, func_name)(args, parser)
+    _run_command_hooks("post", command, args, result)
 
-    return getattr(module, func_name)(args, parser)
+    return result
```
jezdez, line 136: same as above?

conda/cli/conda_argparse.py
```diff
@@ -125,20 +129,26 @@ def do_call(args, parser):
...
 
 
-def _run_pre_command_hooks(command: str, args) -> None:
+def _run_command_hooks(hook_type: CommandHookTypes, *args) -> None:
```
jezdez, line 139: The command is always needed to be passed (we don't have a "catch-all pre/post command hook"), let's just keep the argument, so we don't have to unpack the args later.

```suggestion
def _run_command_hooks(hook_type: CommandHookTypes, command: str, *args, **kwargs) -> None:
```

conda/cli/conda_argparse.py
```diff
@@ -125,20 +129,26 @@ def do_call(args, parser):
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
     command = relative_mod.replace(".main_", "")
-    _run_pre_command_hooks(command, args)
+    _run_command_hooks("pre", command, args)
+    result = getattr(module, func_name)(args, parser)
+    _run_command_hooks("post", command, args, result)
```
jezdez, line 134: To reduce confusion with `*args`/`*kwargs`, what if we'd rename `args` to `arguments`?

conda/cli/conda_argparse.py
```diff
@@ -125,20 +129,26 @@ def do_call(args, parser):
...
-    Helper function used to gather applicable pre_command hook functions
+    Helper function used to gather applicable "pre" or "post" command hook functions
     and then run them.
+
+    The values in *args are passed directly through to the "pre" or "post" command
+    hook function.
     """
-    actions = context.plugin_manager.yield_pre_command_hook_actions(command)
+    command, *_ = args
+    actions = context.plugin_manager.yield_command_hook_actions(hook_type, command)
```
jezdez, line 148:
```suggestion
    actions = context.plugin_manager.yield_command_hook_actions(hook_type, command)
```

conda/cli/conda_argparse.py
```diff
@@ -125,20 +129,26 @@ def do_call(args, parser):
...
 
     for action in actions:
-        action(command, args)
+        action(*args)
```
jezdez, line 151:
```suggestion
        action(*args, **kwargs)
```

conda/plugins/hookspec.py
```diff
@@ -123,3 +129,28 @@ def conda_pre_commands():
...
+
+        **Example:**
+
+        .. code-block:: python
+
+           from conda import plugins
+
+
+           def example_post_command(command, args, result):
+               print("post-command action")
```
jezdez, line 146:
```suggestion
           def example_post_command(command, arguments, result):
               print("post-command action")
```

conda/plugins/manager.py
```diff
@@ -160,19 +160,19 @@ def get_solver_backend(self, name: str = None) -> type[Solver]:
 
         return backend
 
-    def yield_pre_command_hook_actions(self, command: str) -> Iterable[Callable]:
+    def yield_command_hook_actions(self, hook_type: CommandHookTypes, command):
```
jezdez, line 163:
```suggestion
    def yield_command_hook_actions(self, hook_type: CommandHookTypes, command: str):
```

conda/plugins/types.py
```diff
@@ -62,3 +65,17 @@ class CondaPreCommand(NamedTuple):
...
+    Allows a plugin hook to execute after an invoked conda command is run.
+
+    :param name: Post-command name (e.g., ``custom_plugin_post_commands``).
+    :param action: Callable which contains the code to be run.
+    :param run_for: Represents the command(s) this will be run on (e.g. install or create).
+    """
+
+    name: str
+    action: Callable
+    run_for: set[str]
```
jezdez, line 81: Do you think it would be useful to add a feature to optionally receive the result of the command?

jezdez (commented):
Just some naming/conceptual questions

conda/cli/conda_argparse.py
travishathaway: Is this because whatever happened in the command is assumed to have succeeded at this point? (otherwise an exception would have been thrown), and therefore, we don't want to let an uncaught error blow up in a big stack trace?

Something similar came up here: https://github.com/conda/conda/pull/12712#discussion_r1205679348

I'd be more than happy to implement what I originally had planned for `pre_command` for `post_command`.

Ultimately, the plugin should be held responsible for catching and handling all exceptions, but if one falls through the cracks, we could tell users about it and how they might fix the situation (e.g. remove the plugin or file a bug issue with the plugin author).

conda/cli/conda_argparse.py
travishathaway: Sounds good.

conda/cli/conda_argparse.py
travishathaway: I'll change the function signature (`args` -> `arguments`) and add a type annotation to make it super clear.

conda/plugins/types.py
travishathaway: What would the result be? I think most commands either return `None` or throw an exception if there was an error. In the latter case, the `post_command` would never be invoked. Should that behavior be different? (call the `post_command` regardless of whether the command succeeded or not?

travishathaway:
> Just a few tiny nits, think this can go in, given that we're tracking the docs in #12793

I'll address these nits in that pull request.

jezdez:
> > Just a few tiny nits, think this can go in, given that we're tracking the docs in #12793
> 
> I'll address these nits in that pull request.

Yeah, the `sys.argv` related comment wasn't documentation related FWIW

PR #12765, opened 2023-06-06
Introduce `durations.yml` workflow

kenodegard:
### Description

Introduce a new `durations.yml` workflow to run on a weekly schedule updating the duration files based on recent runs. This will help to keep the durations up to date as we continue to evolve the test suite.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/durations.yml
+++ b/.github/workflows/durations.yml
@@ -1,0 +1,53 @@
+name: Update Durations
+
+on:
+  # Every Sunday at 00:00 UTC
+  # https://crontab.guru/#0_0_*_*_0
+  schedule:
+    - cron: '0 0 * * 0'
+
+  workflow_dispatch:
+
+jobs:
+  update-durations:
+    runs-on: ubuntu-latest
+    steps:
+      - name: download recent artifacts
+        run: >
+          gh run list
+            --branch main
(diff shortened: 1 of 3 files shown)


.github/workflows/durations.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: python tests/durations/combine.py ${{ runner.temp }}/artifacts/
+
+      - name: create updated durations PR
+        permissions:
+          contents: write
+          pull-requests: write
+        uses: peter-evans/create-pull-request@v5
+        with:
+          push-to-fork: conda-bot/conda
+          token: ${{ secretes.DURATIONS_TOKEN }}
```
kenodegard, line 45: Need to create this secret before this workflow will work

jezdez (dismissed):

.github/workflows/durations.yml
jezdez: Do you need help with this?

kenodegard: I'm actually thinking to update this to not use a fork and to use elevated permissions instead of defining a new token

dholth:
Makes sense to me

jezdez (approved):

PR #12781, opened 2023-06-08
Add CITATION.cff

schackartk:
### Description

This pull request adds a CITATION.cff file to the root of the repository and would resolve #12779. As discussed in that issue, it provides several ways to help users cite conda. Hopefully this will help give conda it's due credit when it is used!

### Considerations

What information should be included in the citation file is up to the maintainers. I have done my best to fill out the information, but it can be difficult to know what to include (hence the need for such a file!)

A few things that I would suggest for further consideration:

- Who to list as authors? I see that conda does a good job of keeping updated lists. If it is possible to automate the task, it may be useful to add a full author list to the file and keep it up to date. Many projects use a broader approach like "The conda development team" or "Contributors to conda" to simplify this.
- What to use as the date? This is not required. It can be useful, but requires keeping it up to date.
- Should the version number be included in the file? Again, this is useful, but required more upkeep if it is not automated.

### Checklist - did you ...

- [X] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [X] Add / update outdated documentation?

*Note*, I added it to the "Enhancements" section of the news template. Should it be that section, "Documentation", or "Other" instead?

PR #12782, opened 2023-06-08
Move `durations` into `tools` to avoid globbing collision in `changes` job

kenodegard:
### Description

https://github.com/dorny/paths-filter uses globs (using [picomatch](https://github.com/micromatch/picomatch)) to detect file changes. While negated statements are permitted they aren't joined with earlier globs, e.g. if we have the following list of filters:

```
- 'tests/**'
- '!tests/durations/**'
```

this will end up matching everything, all the time since paths-filter is looking if any of the patterns matches a given change (see [this recent non-code change triggering the full test suite](https://github.com/conda/conda/pull/12781) and [this update durations PR](https://github.com/conda/conda/pull/12783)). So instead of fighting with globs lets just move the durations script and jsons out of `tests` and into `tools` instead and avoid the issue entirely.

Xref https://github.com/conda/conda/pull/12765
Xref https://github.com/conda/conda/pull/12778
Xref https://github.com/conda/conda/pull/12780

### Checklist - did you ...


jezdez:
Can we move this into `dev/` instead, not a big fan of generic folder names such as `tools`.

kenodegard:
hehe I actually want to get rid of `dev` and would rather we use `tools`

what do you mean by `tests/durations`? thats where they live currently and this causes the globbing issue

jezdez:
> hehe I actually want to get rid of `dev` and would rather we use `tools`

ah, cool, didn't know that, carry on, let's use `tools`

> what do you mean by `tests/durations`? thats where they live currently and this causes the globbing issue

oh right, that was just a brain fart

jezdez (approved):

Issue #12760, opened 2023-06-06
Add pre-solve and post-solve plugin hooks


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Similar to the pre-/post-command hooks (see https://github.com/conda/conda/issues/12595), the pre-/post-solve hooks will occur directly before and after the sovler is invoked.

### Why is this needed?

This will provide a clean interface for functionality like conda-content-trust or even refactoring features like `pinned_packages`.

### What should happen?

Introduce new pre-/post-command hooks.

travishathaway:
@kenodegard I think it would be really nice to have a specification document for this feature like we did for pre/post commands: https://github.com/conda-incubator/ceps/pull/45

kenodegard:
Both pre- and post-solve hooks have been added in #13053

jezdez:
w00t!

PR #12788, opened 2023-06-10
Implement `repodata_patches` plugin hook

kenodegard:
### Description

After tinkering with the pre-solve (#12760) idea I realized that what I actually had in mind was a repodata patching hook.

The idea here is to allow for local manipulation of the `repodata.json` after fetching/caching but before usage.

POC plugin: https://github.com/kenodegard/conda-repodata

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -756,9 +756,13 @@ class RepodataFetch:
         """
         parsed, state = self.fetch_latest()
         if isinstance(parsed, str):
-            return json.loads(parsed), state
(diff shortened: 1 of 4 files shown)


jezdez:
@kenodegard @jaimergp This bubbled thanks to stalebot, do we still need this? I wonder if this would be helpful for the wheel efforts?

kenodegard:
I'm not sure how useful this would be for the wheels effort, I still view this as a debugging tool and a discovery path for standardization of repodata hotfixing/patching

Issue #12769, opened 2023-06-07
When the user's existing shell environment variable is equal to the value of the conda environment's variable on conda activation, then that environment variable become unset upon conda deactivation


jst1001:
### What happened?

If the user's existing shell environment variable is equal to the value of the conda environment's variable on conda activation, then that environment variable is unset upon conda deactivation. Please see the below example: MYVAR ends up being blank after conda deactivation, when it should be equal to 2.

```
(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$ cat .conda/envs/env7/conda-meta/state
{"env_vars": {"MYVAR": "2"}}
(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$ export MYVAR=2
(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$ echo $MYVAR # <---- At this point, MYVAR is the same as what the conda environment will want it to be
2
(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$ conda activate env7
Activating env7
(env7)jthompson@cluster:~$ echo $MYVAR 
2
(env7)jthompson@cluster:~$ conda deactivate
Deactivating env7
(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$ echo $MYVAR 

(NO CONDA ENVIRONMENT SET!)jthompson@cluster:~$
```

I believe that there is a bug inside the conda activate python script: `conda/conda/activate.py`. This happens because conda isn't creating a `__CONDA_SHLVL_` variable when the environment variable would be equal before and after conda activation. This bug seems to have been introduced in conda version 23.1.0 (2023-01-17) from a single line change in commit ID: `b038a8e092ca62b20286a9830f263fc007d2c09e` . Performing a git blame on `conda/conda/activate.py` from that commit shows:

`b038a8e092 (Dario Curreri       2022-12-07 16:10:25 +0100  342)         clobber_vars = set(filter(lambda var: env_vars[var] != os.environ[var], clobber_vars))`

The commit message states: "Avoid warning when environment variables have the same value of OS ones". It is true that this line avoids the warning message, but it does more than that. It also prevents the conda backup variable, in my example `__CONDA_SHLVL_0_MYVAR`, from being created. This produces unwanted results of the environment variable being unset when later deactivating from the environment.

[...]

PR #12789, opened 2023-06-11
Fix equal env variable deactivation #12769

dariocurr:
### Description

Resolves https://github.com/conda/conda/issues/12769

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -370,19 +370,17 @@ class _Activator(metaclass=abc.ABCMeta):
 
         # get clobbered environment variables
         clobber_vars = set(env_vars.keys()).intersection(os.environ.keys())
-        overwritten_clobber_vars = dict()
-        for clobber_var in clobber_vars:
-            os_value = os.environ[clobber_var]
-            clobber_value = env_vars[clobber_var]
-            if clobber_value != os_value:
-                overwritten_clobber_vars[clobber_var] = (os_value, clobber_value)
+        overwritten_clobber_vars = [
+            clobber_var
+            for clobber_var in clobber_vars
+            if os.environ[clobber_var] != env_vars[clobber_var]
+        ]
         if overwritten_clobber_vars:
             print(
                 "WARNING: overwriting environment variables set in the machine",
                 file=sys.stderr,
             )
-            for key, values in overwritten_clobber_vars.items():
-                print(f"`{key}`: from `{values[0]}` to `{values[1]}`", file=sys.stderr)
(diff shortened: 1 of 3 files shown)


conda/activate.py
mitchrosen, line 1: Rather than change the user experience, the simplest change that fixes the bug described in [conda#12769](https://github.com/conda/conda/issues/12769) would be to remove original lines 373-375.  Your new lines 373-379 and then finally 384-385 are big changes from what the user expects now, could be a huge list if there are many env_vars and some of these vars could have very long definitions which will all be spit out at every "conda activate."  

Who is demanding this information on a "conda activate"?

mitchrosen:
Hi @dariocurr.  Thanks for taking on this change to address issue [conda#12769](https://github.com/conda/conda/issues/12769).  It makes sense for you to take this step as we did identify your check-in as being the root cause of the bug.

Full disclosure here, I and my colleague, the author of [conda#12769](https://github.com/conda/conda/issues/12769), are complete newbies to Conda's code-base and should not be relied upon for sage advice.  But, I would say it's a very big deal to change the user-facing experience following "conda activate."  I suggest it is very important to seek out a long and deep conversation among very seasoned contributors before making fundamental changes to the user interface.  Even if it turns out to be a good idea, I wonder if this type of change should be in a debug-level update.  Besides making changes that will break user expectations and potentially break people's scripts, your version will print out variable settings on every conda activate that may very well be long and involved and could be unwelcome to the average user.

dariocurr:
Hi @mitchrosen. I see your points.
I think the last commit is the least intrusive solution.
It fixes the bug in #12769 and keeps the previous behavior of not printing variables if they have the same value.

news/12769-fix-bug-deactivation
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+*   <news item>
+
+### Bug fixes
+
+*   Fix bug: When the user's existing shell environment variable is equal to the value of the conda environment's variable on conda activation, then that environment variable become unset upon conda deactivation (#12769)
```
kenodegard, line 7:
```suggestion
*   Fix erroneous `conda deactivate` behavior of unsetting preexisting environment variables that are identical to those set during `conda activate`. (#12769)
```

kenodegard (approved):
looks good!

thanks @jst1001 for reporting, @dariocurr for coding, and @mitchrosen for reviewing!

dariocurr:
My pleasure!

Issue #12793, opened 2023-06-13
Revise index page for plugins in documentation


travishathaway:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Our first pass at creating a guide for plugins was good, but I think it would be better to use the [index page](https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/index.html) as a place to hold a quick start guide. This quick start guide should follow what Flask does: https://flask.palletsprojects.com/en/2.3.x/quickstart/. Right at the beginning we see a full working example that is then explained step by step.

### Why is this needed?

This should hopefully make it easier for users to understand how to write a plugin.

### What should happen?

- [ ] Show a full working example using more than one plugin (e.g. `pre_command` and `subcommand` hooks)
- [ ] Use some important conda data structures (e.g. `conda.base.context.context`) to show a very basic usage of the conda API
- [ ] Talk about some other useful parts of the conda API (e.g. `CondaSession`, `Channel`, etc.)
- [ ] Remove discussion of pluggy and make that more of a footnote.

### Additional Context

The general idea here is giving plugin authors enough of a push in the right direction to begin implementing real-world applications using modules in conda.

travishathaway:
Be sure to address these nits:

https://github.com/conda/conda/pull/12758#pullrequestreview-1476983226

PR #12802, opened 2023-06-15
Revising the index page of the plugins section in our documentation

travishathaway:
Fixes: #12793

### Description

Please see the issue above for a full description.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/__init__.py
+++ b/conda/plugins/__init__.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """
 In this module, you will find everything relevant to conda's plugin system.
-It contains all of the code that plugin authors will use to write plugins
+It contains all of the code that plugin authors will use to write plugins,
 as well as conda's internal implementations of plugins.
 
 **Modules relevant for plugin authors**
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -50,7 +50,7 @@ class CondaPluginManager(pluggy.PluginManager):
         """
         Load the provided list of plugins and fail gracefully on error.
         The provided list of plugins can either be classes or modules with
-        :attr:`~conda.plugins.hook_impl`.
+        :attr:`~conda.plugins.hookimpl`.
         """
         plugin_names = []
         for plugin in plugins:
--- a/docs/source/dev-guide/plugins/index.rst
+++ b/docs/source/dev-guide/plugins/index.rst
@@ -87,7 +87,7 @@ More examples
 
 To see more examples of conda plugins, please visit the following resources:
 
-- `conda-plugins-template`_: this is a repository with full examples that could be used a starting point for your plugin
+- `conda-plugins-template`_: This is a repository with full examples that could be used a starting point for your plugin
 
 Using other plugin hooks
 ------------------------
@@ -127,7 +127,7 @@ A note on licensing
 
 For more information on which license to use for your custom plugin, please reference
 the `"Choose an Open Source License"`_ site. If you need help figuring out exactly
-which one to use, we advise to seeking a qualified legal professional.
+which one to use, we advise communicating with a qualified legal professional.
 
 
 .. _Pluggy: https://pluggy.readthedocs.io/en/stable/


conda/cli/conda_argparse.py
```diff
@@ -117,9 +117,10 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
     """
     # First, check if this is a plugin subcommand; if this attribute is present then it is
     if getattr(arguments, "plugin_subcommand", None):
-        _run_command_hooks("pre", arguments.plugin_subcommand.name, sys.argv[2:])
+        plugin_arguments = sys.argv[2:]
+        _run_command_hooks("pre", arguments.plugin_subcommand.name, plugin_arguments)
         result = arguments.plugin_subcommand.action(sys.argv[2:])
-        _run_command_hooks("post", arguments.plugin_subcommand.name, sys.argv[2:])
+        _run_command_hooks("post", arguments.plugin_subcommand.name, plugin_arguments)
```
kenodegard, line 123: this means that the pre-command can modify the argument list and post-command will get the modified list, however the plugin itself will still receive the unmodified args

I can see value in allowing a pre-command plugin to modify the arguments before passing it to the subcommand but that is not what is currently happening and if this is a desired feature we should document it as such

news/12802-revise-plugins-index-page
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* Revising the plugins index page to make it easier to understand how to create a conda plugin.
```
kenodegard, line 15:
```suggestion
* Revise the plugins index page to make it easier to understand how to create a conda plugin. (#12802)
```

kenodegard (changes requested):

conda/cli/conda_argparse.py
jezdez: I don't think the pre command hook or the subcommand should be able to modify the arguments, we should handle `sys.argv` once, before all that and make it immutable and then pass them through the command flow. If we ever want to enable `sys.argv` modification it should become an own feature

conda/cli/conda_argparse.py
```diff
@@ -117,9 +117,10 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
     """
     # First, check if this is a plugin subcommand; if this attribute is present then it is
     if getattr(arguments, "plugin_subcommand", None):
-        _run_command_hooks("pre", arguments.plugin_subcommand.name, sys.argv[2:])
+        plugin_arguments = sys.argv[2:]
```
jezdez, line 120: This makes the arguments immutable, which I think is what we need to have consistency between the various steps of the command flow
```suggestion
        plugin_arguments = tuple(sys.argv[2:])
```

jezdez (changes requested):

conda/cli/conda_argparse.py
travishathaway: Perhaps it's better to make a copy and then pass it in as a tuple?

conda/cli/conda_argparse.py
travishathaway: Below, for normal conda commands, we pass in an `argparse.Namespace` object. These are mutable. I will also update the code to pass in just a tuple there as well.

Plus, I'm now realizing how inconsistent that was too 🙈 (i.e. for built-in commands you get one thing, for plugin commands something totally different).

Thanks for catching this @kenodegard and @jezdez.

conda/cli/conda_argparse.py
```diff
@@ -130,9 +132,9 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
     module = import_module(relative_mod, __name__.rsplit(".", 1)[0])
 
     command = relative_mod.replace(".main_", "")
-    _run_command_hooks("pre", command, arguments)
+    _run_command_hooks("pre", command, plugin_arguments)
```
jezdez, line 135:
```suggestion
    _run_command_hooks("pre", command, arguments)
```

conda/cli/conda_argparse.py
```diff
@@ -130,9 +132,9 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
...
     result = getattr(module, func_name)(arguments, parser)
-    _run_command_hooks("post", command, arguments)
+    _run_command_hooks("post", command, plugin_arguments)
```
jezdez, line 137:
```suggestion
    _run_command_hooks("post", command, arguments)
```

jezdez (changes requested):
This needs a number of changes for consistency:

- fully populate the plugin subcommand namespace
- pass the namespace to the pre- and post hooks (they are immutable)
- continue to pass the `sys.argv` to plugin subcommands to enable advanced plugin implementations (e.g. via click)
- continue to pass the parsed namespace to internal non-plugin subcommands

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +52,77 @@ a package out of the custom plugin hook shown above:
...
+Our plugins system is built on top of the great `Pluggy`_ library. For more information about how it
+works, we suggest heading over to their `documentation`_.
+
+
+A note on licensing
+-------------------
+
+When licensing plugins, we recommend using licenses such as BSD-3_, MIT_, and
+`Apache License 2.0`_. Some ``import`` statements may possibly require the GPLv3_
+license, which ensures that the software being licensed is open source.
```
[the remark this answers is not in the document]
travishathaway: 😬. Sounds like a good idea.

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +52,77 @@ a package out of the custom plugin hook shown above:
...
 
    from setuptools import setup
 
    setup(
-       name="my-conda-plugin",
+       name="example",
        install_requires="conda",
-       entry_points={"conda": ["my-conda-plugin = my_plugin"]},
+       entry_points={"conda": ["example = example"]},
        py_modules=["my_plugin"],
```
travishathaway, line 73:
```suggestion
       py_modules=["example_plugin"],
```

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +52,77 @@ a package out of the custom plugin hook shown above:
...
+Using other plugin hooks
+------------------------
 
-   Conda plugin hook implementation marker.
+For examples of how to use other plugin hooks, please read their respective documentation pages:
 
 .. toctree::
    :maxdepth: 1
 
-   manager
```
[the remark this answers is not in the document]
travishathaway: I removed it because it's not a plugin hook and isn't important for plugin authors to be aware of. The plugin manager should only ever be used inside of conda.

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +52,77 @@ a package out of the custom plugin hook shown above:
...
+-------------
 
+To see more examples of conda plugins, please visit the following resources:
 
-API reference
--------------
+- `conda-plugins-template`_: this is a repository with full examples that could be used a starting point for your plugin
 
-.. autoattribute:: conda.plugins::hookimpl
-   :no-value:
```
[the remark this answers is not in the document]
travishathaway: Yeah, that's a far ask.

Otherwise, it just looked really weird having this there:

Bildschirmfoto 2023-06-20 um 12 13 44

I'm not sure that would really make sense for a new reader.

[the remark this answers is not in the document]
jezdez: Sure, let's fix the docstring of the hookimpl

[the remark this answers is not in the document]
travishathaway: Even if the doc string was better. I still do not think it would be a good fit for this particular page. The quick start already shows how to use it. I think this would be a better fit for a dedicated "API" page where users could go for a further detailed reference.

travishathaway:
I removed all code changes from this pull request so we can focus on the documentation instead.

conda/plugins/__init__.py
```diff
@@ -8,3 +32,6 @@
     CondaSubcommand,
     CondaVirtualPackage,
 )
+
+# Used for sphinx-autoapi
+__all__ = ()
```
jezdez, line 37: Remember that this will ALSO prevent imports such as `from conda.plugins import *` from working

travishathaway: @jezdez, I removed these for now.

docs/source/dev-guide/plugins/index.rst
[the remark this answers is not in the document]
travishathaway: We now have the full Plugin API linked to at the bottom and this pull request has made some updates to this documentation too.

beeankha (changes requested):

beeankha:
A few more things I found below, accompanied with screenshots:

* * * 

In `conda.plugins.manager.py`, there's a docstring that appears to get cut off:
Screen Shot 2023-06-22 at 4 58 50 PM

* * *

In the same `manager.py` file, there's a docstring that needs linebreaks (this might not be an easy fix and is not necessarily a blocker):
Screen Shot 2023-06-22 at 5 00 28 PM

* * * 

Another thing in that `manager.py` file is an instance of `log` by itself in a table with no accompanying text:
Screen Shot 2023-06-22 at 4 59 18 PM

docs/source/dev-guide/plugins/index.rst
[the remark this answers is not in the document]
kenodegard: so `manager.rst` should be deleted then?

also the `plugins-api` link added to the bottom doesn't appear anywhere in this page

docs/source/dev-guide/plugins/index.rst
```diff
@@ -1,52 +1,48 @@
...
+
+   def command(arguments: list[str]):
+       print("Conda subcommand!")
 
 
    @conda.plugins.hookimpl
    def conda_subcommands():
-       ...
+       yield plugins.CondaSubcommand(
+           name="example", action=command, summary="Example of a conda subcommand"
```
kenodegard, line 27: the trialing comma trick!

```suggestion
           name="example",
           action=command,
           summary="Example of a conda subcommand",
```

conda/plugins/hookspec.py
```diff
@@ -15,9 +21,15 @@
     CondaVirtualPackage,
 )
 
+#: Name used for organizing conda hook specifications
 spec_name = "conda"
```
kenodegard, line 25:
```suggestion
spec_name = "conda"
"""Name used for organizing conda hook specifications"""
```

unless there was some reason for leaving this hidden from sphinx-autoapi

docs/source/dev-guide/plugins/index.rst
```diff
@@ -125,3 +136,5 @@ API reference
 .. _`Apache License 2.0`: https://www.apache.org/licenses/LICENSE-2.0
 .. _GPLv3: https://www.gnu.org/licenses/gpl-3.0.en.html
 .. _`"Choose an Open Source License"`: https://choosealicense.com/
+.. _`conda-plugins-template`: https://github.com/conda/conda-plugin-template/tree/main/tutorials
```
kenodegard, line 139:
```suggestion
.. _`conda-plugins-template`: https://github.com/conda/conda-plugin-template
```

kenodegard:
Re. the `log` constant with no description, this will show up throughout a significant portion of the code so we probably want to do a refactor and make all of those private, alternatively we can configure sphinx-autoapi to explicitly ignore any `log` variables (https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#event-autoapi-skip-member)

dholth:
Very simple. :) I would dramatically simplify the language to "disable plugins". The "only builtin commands" qualifier makes no sense. I guess it means "doesn't affect conda commands that are started in a separate process (legacy) or themselves provided by (doesn't start with conda.?) plugins.

docs/source/dev-guide/plugins/index.rst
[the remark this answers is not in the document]
beeankha: I removed that link at the bottom of the page (there's already a `:doc:`-style link in the API section)

jezdez (dismissed):

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +54,82 @@ a package out of the custom plugin hook shown above:
...
+       name="conda-example-plugin",
        install_requires="conda",
-       entry_points={"conda": ["my-conda-plugin = my_plugin"]},
-       py_modules=["my_plugin"],
+       entry_points={"conda": ["conda-example-plugin = example_plugin"]},
+       py_modules=["example_plugin"],
    )
 
+In both examples shown above, we define an entry point for conda. It's important to make sure
+that the entry point is for "conda" and that it points to the correct module in your plugin package.
```
kathatherine, line 79: Would we want to put "conda" in backticks here?

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +54,82 @@ a package out of the custom plugin hook shown above:
...
-use case. Be sure to credit the original author of the plugin, and keep in mind that
-licenses can be altered depending on the situation.
+More examples
+-------------
 
-For more information on which license to use for your custom plugin, please reference
-the `"Choose an Open Source License"`_ site.
+To see more examples of conda plugins, please visit the following resources:
 
+- `conda-plugins-template`_: this is a repository with full examples that could be used a starting point for your plugin
```
kathatherine, line 90:
```suggestion
- `conda-plugins-template`_: This is a repository with full examples that could be used a starting point for your plugin
```

docs/source/dev-guide/plugins/index.rst
```diff
@@ -56,67 +54,82 @@ a package out of the custom plugin hook shown above:
...
+For even more detailed information about our plugin system, please the see the
+:doc:`Plugin API </dev-guide/api/conda/plugins/index>` section.
+
+
+A note on licensing
+-------------------
+
+For more information on which license to use for your custom plugin, please reference
+the `"Choose an Open Source License"`_ site. If you need help figuring out exactly
+which one to use, we advise to seeking a qualified legal professional.
```
kathatherine, line 130:
```suggestion
which one to use, we advise communicating with a qualified legal professional.
```
Could also use "seeking out", but wasn't sure how colloquial that phrase was.

kathatherine (changes requested):
Couple things

docs/source/dev-guide/plugins/index.rst
beeankha: Since this section is talking about entry points and we're referring to conda as the specific example, I think the quote marks are appropriate.

conda/plugins/manager.py
```diff
@@ -44,7 +50,7 @@ def __init__(self, project_name: str | None = None, *args, **kwargs) -> None:
     def load_plugins(self, *plugins) -> list[str]:
         """
         Load the provided list of plugins and fail gracefully on error.
-        The provided list plugins can either be classes or modules with
+        The provided list of plugins can either be classes or modules with
         :attr:`~conda.plugins.hook_impl`.
```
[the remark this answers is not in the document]
beeankha: Good catch!!

kathatherine (approved):
👍 Thanks!

docs/source/dev-guide/plugins/index.rst
kathatherine: Sorry, just saw this. I meant doing `"conda"`, to be clear. But no worries either way.

Issue #12740, opened 2023-06-01
Available subcommands appear on help pages they probably shouldn't be


travishathaway:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When you run `conda create --help` it will show the normal help output, but it also shows the following at the end:

```
conda commands available from other packages:
 doctor - A subcommand that displays environment health report
```

I think the desired behavior should be to only show this when running `conda --help`.

### Conda Info

```shell
n/a
```

### Conda Config

```shell
n/a
```

### Conda list

```shell
n/a
```

travishathaway:
I recently discovered that this bug also causes the examples to disappear completely 😱. Run `conda env create --help` to see this.

PR #12814, opened 2023-06-20
Revise subcommand plugin hook API

jezdez:
### Description

This is a follow-up on #12802 (which will be reduced in scope on documentation only in the next few day).

We need to revise the subcommand plugin hook to have more consistent handling of how arguments are parsed and passed to subcommands provided by plugins, versus those that are built-in and are configured in the traditional way. Effectively, this also enables plugin subcommands more complex integrations into the `argparse`-based main `conda` command.

There are also some fixes for inconsistent help texts, rendering issues etc.

Fixes: https://github.com/conda/conda/issues/12740

### Checklist - did you ...


kenodegard (commented):
I don't think this warrants an entirely new plugin hook/type. IMO it would be cleaner and easier to document if the `CondaSubcommand` were to offer an optional `configure_argparse` function instead.

conda/cli/conda_argparse.py
```diff
@@ -361,6 +368,27 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        # plugin subcommands later
+        parser.add_argument(
+            "plugin_args",
+            nargs=argparse.REMAINDER,
+            help=argparse.SUPPRESS,  # to hide it from the help output
+        )
+        try:
+            plugin_subcommand.configure_parser(parser)
+        except NotImplementedError:
+            pass
```
kenodegard, line 389: I think this collection of remaining arguments should only occur if the plugin doesn't define `configure_parser`:

```suggestion
        try:
            plugin_subcommand.configure_parser(parser)
        except NotImplementedError:
            # we store all other arguments here, so we can pass them to the
            # plugin subcommands later
            parser.add_argument(
                "plugin_args",
                nargs=argparse.REMAINDER,
                help=argparse.SUPPRESS,  # to hide it from the help output
            )
```

kenodegard: I also think we need to disable `argparse`'s default `--help` argument when we enable this `plugin_args` argument since we'd want the plugin to intercept the `--help` argument and handle it as they see fit

jezdez: I've switched the argument parser to handle option conflicts more gracefully, which should suffice to make this possible.

conda/plugins/types.py
```diff
@@ -34,6 +38,9 @@ class CondaSubcommand(NamedTuple):
         [list[str]],  # arguments
         int | None,  # return code
     ]
```
kenodegard, line 40: Will it be better to use overloading here?

```python
@dataclass
class CondaSubcommand:
    name: str
    summary: str
    action: Callable[[Namespace | tuple[str]], int | None]
    configure_parser: Callable[[ArgumentParser], None] | None = field(default=None)

    @overload
    def __init__(self, name: str, summary: str, action: Callable[[Namespace], int | None], configure_parser: Callable[[ArgumentParser], None]) -> None:
        ...

    @overload
    def __init__(self, name: str, summary: str, action: Callable[[tuple[str]], int | None], configure_parser: None = None) -> None:
        ...

```

jezdez: Huh, where is the `overload` decorator coming from?

kenodegard: `typing.overload`

conda/plugins/types.py
```diff
@@ -16,7 +18,8 @@
 """The two different types of `conda_*_commands` hooks that are available"""
 
 
-class CondaSubcommand(NamedTuple):
+@dataclass
+class CondaSubcommand:
```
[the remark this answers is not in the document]
kenodegard: Yeah, `NamedTuple` made sense for the initial plugins, these more advanced types are better described with dataclasses (AFAIK the overloading example above is not possible with `NamedTuple`)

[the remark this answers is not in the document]
jezdez: Yep, makes sense to me

kenodegard:
Thanks to @kalawac's work on adding tests to the `conda shells` we found that the current subcommand plugins implementation that uses `sys.argv[2:]` breaks `conda_cli` fixture testing since the passed in arguments are ignored in favor of whatever the `sys.argv` value (which will be whatever we define to run pytest.

Because of that, this refactor is even more pressing IMO.

kenodegard:
ooo just stumbled over this: https://pluggy.readthedocs.io/en/stable/#opt-in-arguments

pluggy supports opt-in arguments, if i understand correctly this means downstream plugins can selectively choose which arguments to accept (like pytest fixtures)!

conda/cli/conda_argparse.py
```diff
@@ -97,42 +100,48 @@ def generate_parser():
     configure_parser_info(sub_parsers)
     configure_parser_init(sub_parsers)
     configure_parser_install(sub_parsers)
+    configure_parser_notices(sub_parsers)
     configure_parser_list(sub_parsers)
```
kenodegard, line 104: I assume the intent here was for it to be alphabetical:

```suggestion
    configure_parser_list(sub_parsers)
    configure_parser_notices(sub_parsers)
```

jezdez: 🤦

conda/cli/conda_argparse.py
```diff
@@ -104,44 +107,51 @@ def generate_parser():
...
+    if plugin_subcommand:
+        # pass on the rest of the plugin specific args or fall back to
+        # the whole discovered arguments
+        try:
+            plugin_args = tuple(args.plugin_args)
+        except AttributeError:
+            plugin_args = args
+        context.plugin_manager.apply_pre_commands(plugin_subcommand.name)
+        result = plugin_subcommand.action(plugin_args)
+        context.plugin_manager.apply_post_commands(plugin_subcommand.name)
```
kenodegard, line 144:
```suggestion
        context.plugin_manager.invoke_pre_commands(plugin_subcommand.name)
        result = plugin_subcommand.action(plugin_args)
        context.plugin_manager.invoke_post_commands(plugin_subcommand.name)
```

jezdez: Okay, I'm backing away from the computer now...

conda/plugins/types.py
kenodegard: this is good enough: https://github.com/conda/conda/pull/12814/commits/cb3efa5fe1551382d1616ce2afa55aaca70ec43b

conda/plugins/types.py
```diff
@@ -22,15 +29,19 @@ class CondaSubcommand(NamedTuple):
 
     :param name: Subcommand name (e.g., ``conda my-subcommand-name``).
     :param summary: Subcommand summary, will be shown in ``conda --help``.
+    :param setup: Callable that will be run when the subcommand parser is set up.
     :param action: Callable that will be run when the subcommand is invoked.
```
kenodegard, line 33:
```suggestion
    :param action: Callable that will be run when the subcommand is invoked.
    :param configure_parser: Callable that will be run when the subcommand parser is initialized.
```

beeankha (approved):

kenodegard (approved):

PR #12816, opened 2023-06-20
Delete `test_import.py`

kenodegard:
### Description

This test did not catch import issues previously (https://github.com/conda/conda/pull/12545) so we are likely better off relying on coverage reports.

Furthermore the original purpose for these tests is no longer valid (Python 3 compatibility).

### Checklist - did you ...


jezdez (approved):
🔥

beeankha (approved):

Issue #12748, opened 2023-06-05
Add a `--no-plugins` option to conda for core commands


travishathaway:
### What's the idea?

When a plugin triggers an error, we need a way to temporarily disable all plugins so that it can be uninstalled if necessary.

### What should be done?

- [ ] Add a new option called `--no-plugins` that will be used to temporarily disable plugins (this should apply to all "[core commands](https://github.com/conda/conda/tree/main/conda/cli)").
- [ ] We should update the the error report page to mention this option.
- [ ] Write tests to ensure this behavior works as expected.

### Additional context

This is probably how we will be able to disable plugins:

- https://pluggy.readthedocs.io/en/stable/#blocking

jezdez:
(didn't know @travishathaway had already asked @dholth about working on this)

PR #12824, opened 2023-06-22
Create a `--no-plugins` option to disable external conda plugins

travishathaway:
### Description

Fixes: #12748 

Please see the above issue for more details.

### Usage

When you want to disable an external plugin from running during built-in conda commands you can use the new option like so:

```
conda --no-plugins install python
```

This option has to be provided before the sub-command.

### Checklist - did you ...


travishathaway:
This is currently being applied for `conda env` commands. Is that okay? I think it's okay because `conda env` commands are necessary to remove a problematic plugin. The goal of this option is to allow users to repair their conda base environment by giving them the ability to remove a problematic plugin.

conda/plugins/manager.py
```diff
@@ -175,6 +175,14 @@ def yield_command_hook_actions(self, hook_type: CommandHookTypes, command: str):
             if command in command_hook.run_for:
                 yield command_hook.action
 
+    def disable_external_plugins(self) -> None:
+        """
+        Disables all currently registered plugins except built-in conda plugins
+        """
+        for name, _ in self.list_name_plugin():
+            if not name.startswith("conda.plugins"):
```
jezdez, line 183:
```suggestion
            if not name.startswith("conda."):
```
Let's not presume that all plugins will forever live in the `conda.plugins` package.

travishathaway:
The end of the unexpected error report now looks like this:

```

consider using the --no-plugins option to turn off external plugins.

Example: conda --no-plugins install <package> 
```

Maybe that's a little too much 🤷‍♂️. Not married to keeping that in there. I just thought it would be a good place to put that type of information.

conda/cli/conda_argparse.py
```diff
@@ -85,6 +85,12 @@ def generate_parser():
         action="store_true",
         help=SUPPRESS,
     )
+    p.add_argument(
+        "-P",
```
jezdez, line 89: This might be confusing with the lower-case `-p` for `--path`?
```suggestion
```

conda/exception_handler.py
```diff
@@ -174,6 +174,15 @@ def print_unexpected_error_report(self, error_report):
                 "An unexpected error has occurred. Conda has prepared the above report."
             )
             message_builder.append("")
+            message_builder.append(
+                "If you suspect this error is being caused by a malfunctioning plugin,"
+            )
+            message_builder.append(
+                "consider using the --no-plugins option to turn off external plugins."
```
jezdez, line 181: I wonder if users understand what "external" means, can we somehow use "built-in"?

```suggestion
                "consider using the --no-plugins option to only enable built-in plugins."
```

jezdez (commented):
I'm a little uncertain if a binary toggle is enough, if we only disable non-built-ins, what if the problem is in a built-in plugin?

conda/cli/conda_argparse.py
travishathaway: Fair. Do you have better suggestion? 😁 We could also just not even include an alias for it.

travishathaway: I was thinking about not even including it originally.

jezdez: Yeah agreed, let's just keep `--no-plugins`, that's descriptive and explicit.

Should we ALSO add a config option BTW?

jezdez (dismissed):
This is a good thing, let's merge and maybe refine it later by adding a dedicated config option as well? I think that might be useful for CI scripts for those that truly don't want to have plugins enabled, period.

kenodegard (changes requested):
I’m thinking we should make this a `context` value instead, so instead of:

`conda --no-plugins`

which as you noted wont work for `conda env`, we do:

`CONDA_NO_PLUGINS=true conda`

this would also allow users to “permanently” disable plugins via `.condarc`

we can also configure it similarly to how `context.channels` works where the CLI argument overrides the `.condarc` value

conda/base/constants.py
```diff
@@ -319,3 +319,5 @@ class NoticeLevel(ValueEnum):
 
 # Not all python namespace packages are registered on PyPI. If a package
 # contains files in site-packages, it probably belongs in the python namespace.
+
+NO_PLUGINS = False
```
jezdez, line 323: Let's add a comment here what this is

conda/base/context.py
```diff
@@ -446,6 +444,18 @@ def experimental_solver(self):
...
 
+    ####################################################
+    #               Plugin Configuration               #
+    ####################################################
+
+    _no_plugins = ParameterLoader(
+        PrimitiveParameter(NO_PLUGINS, element_type=(bool, NoneType)),
+        aliases=(
+            "no_plugins",
+            "no_plugin",
```
jezdez, line 455: I don't think we need `no_plugin` here

conda/base/context.py
```diff
@@ -487,6 +499,24 @@ def plugin_manager(self):
...
+            self._argparse_args
+            and "no_plugins" in vars(self._argparse_args)
+            and self._argparse_args.no_plugins is not None
+        ):
+            return True
+        elif os.getenv("NO_PLUGINS") is not None:
+            return True
+        else:
+            return False
+
```
jezdez, line 519:
```suggestion
        return (
            self._argparse_args
            and "no_plugins" in self._argparse_args
            and self._argparse_args["no_plugins"] is not None
        ) or os.getenv("NO_PLUGINS") is not None

```

conda/base/context.py
```diff
@@ -1499,6 +1530,11 @@ def description_map(self):
             #     environments and inconsistent behavior. Use at your own risk.
             #     """
             # ),
+            no_plugins=dals(
+                """
+                Disable or enable all currently-registered plugins except built-in conda plugins.
```
jezdez, line 1535:
```suggestion
                Disable all currently-registered plugins, except built-in conda plugins.
```

conda/cli/conda_argparse.py
```diff
@@ -115,7 +120,9 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
     Serves as the primary entry point for commands referred to in this file and for
     all registered plugin subcommands.
     """
-    # First, check if this is a plugin subcommand; if this attribute is present then it is
+    if hasattr(arguments, "no_plugins") and arguments.no_plugins or context._no_plugins:
```
jezdez, line 123: Shouldn't this just refer to `context.no_plugins`? I think it would be good to implement this logic in one place.

conda/cli/conda_argparse.py
```diff
@@ -85,6 +85,11 @@ def generate_parser():
         action="store_true",
         help=SUPPRESS,
     )
+    p.add_argument(
+        "--no-plugins",
+        action="store_true",
+        help="Disable usage of all plugins during built-in conda commands (e.g. 'install')",
```
jezdez, line 91:
```suggestion
        help="Disable all plugins that are not built into conda.",
```

conda/exception_handler.py
```diff
@@ -169,11 +169,21 @@ def print_unexpected_error_report(self, error_report):
...
+                "",
+                "If you suspect this error is being caused by a malfunctioning plugin,",
+                "consider using the --no-plugins option to turn off external plugins.",
+                "",
+                "Example: conda --no-plugins install <package>",
+                "",
+                "Alternatively, you can set the 'no_plugins' option to run the command",
+                "without external (i.e., non-conda) plugins enabled. You can do this",
+                "by running:",
+                "    $ conda config --set no_plugins true",
```
jezdez, line 184:
```suggestion
                "If you suspect this error is being caused by a malfunctioning plugin,",
                "consider using the --no-plugins option to turn off plugins.",
                "",
                "Example: conda --no-plugins install <package>",
                "",
                "Alternatively, you can set the 'no_plugins' option to run the command",
                "without plugins enabled. You can do this by running:",
                "    $ conda config --set no_plugins true",
```

jezdez (changes requested):
Some tweaks

conda/base/context.py
beeankha: I was thinking of the most common and harmless typo instance to cover here, but I will go ahead and remove it! 👍

conda/exception_handler.py
```diff
@@ -169,11 +169,20 @@ def print_unexpected_error_report(self, error_report):
...
+                "If you suspect this error is being caused by a malfunctioning plugin,",
+                "consider using the --no-plugins option to turn off plugins.",
+                "",
+                "Example: conda --no-plugins install <package>",
+                "",
+                "Alternatively, you can set the 'no_plugins' option to run the command",
+                "without plugins enabled. You can do this by running:",
+                "    $ conda config --set no_plugins true",
+                "",
             )
```
jezdez, line 185: `message_builder` is a Python list and needs to be extended:

```suggestion
            message_builder.extend([
                "",
                "An unexpected error has occurred. Conda has prepared the above report."
                "",
                "If you suspect this error is being caused by a malfunctioning plugin,",
                "consider using the --no-plugins option to turn off plugins.",
                "",
                "Example: conda --no-plugins install <package>",
                "",
                "Alternatively, you can set the 'no_plugins' option to run the command",
                "without plugins enabled. You can do this by running:",
                "    $ conda config --set no_plugins true",
                "",
            ])
```

jezdez (changes requested):

beeankha:
pre-commit.ci autofix

conda/base/context.py
ForgottenProgramme: > I was thinking of the most common and harmless typo instance to cover here
I think that's a good idea.

conda/base/context.py
```diff
@@ -487,6 +496,19 @@ def plugin_manager(self):
...
+        This property is used to determine if plugins should be loaded or not.  It is set by
+        the ``--no-plugins`` command line flag.  It can also be set by the ``NO_PLUGINS``
+        environment variable.  If neither of those are set, then the default value is used.
+        """
+        return (
+            self._argparse_args
+            and "no_plugins" in self._argparse_args
+            and self._argparse_args["no_plugins"] is not None
+        ) or os.getenv("NO_PLUGINS") is not None
+
```
kenodegard, line 511: We don't need to define a new `NO_PLUGINS` envvar, `CONDA_NO_PLUGINS` is a freebee we get by adding `_no_plugins = ParameterLoader(...)` to the context.

conda/base/context.py
```diff
@@ -487,6 +496,19 @@ def plugin_manager(self):
...
+        """
+        This property is used to determine if plugins should be loaded or not.  It is set by
+        the ``--no-plugins`` command line flag.  It can also be set by the ``NO_PLUGINS``
+        environment variable.  If neither of those are set, then the default value is used.
+        """
+        return (
+            self._argparse_args
+            and "no_plugins" in self._argparse_args
+            and self._argparse_args["no_plugins"] is not None
+        ) or os.getenv("NO_PLUGINS") is not None
```
kenodegard, line 510: This is also missing the fallback to `self._no_plugins` if `self._argparse_args` doesn't define `--no-plugins`.

conda/base/context.py
jezdez: Oh that's interesting!

conda/cli/conda_argparse.py
```diff
@@ -115,7 +120,9 @@ def do_call(arguments: argparse.Namespace, parser: ArgumentParser):
     Serves as the primary entry point for commands referred to in this file and for
     all registered plugin subcommands.
     """
-    # First, check if this is a plugin subcommand; if this attribute is present then it is
+    if context._no_plugins:
```
kenodegard, line 123: Here we want to reference the special property we created:

```suggestion
    if context.no_plugins:
```

conda/base/context.py
jezdez:
```suggestion
        return (
            self._argparse_args
            and "no_plugins" in self._argparse_args
            and self._argparse_args["no_plugins"] is not None
        ) or self._no_plugins
```

Would that suffice?

conda/plugins/manager.py
```diff
@@ -181,6 +181,14 @@ def yield_command_hook_actions(self, hook_type: CommandHookTypes, command: str):
             if command in command_hook.run_for:
                 yield command_hook.action
 
+    def disable_external_plugins(self) -> None:
+        """
+        Disables all currently registered plugins except built-in conda plugins
+        """
+        for name, _ in self.list_name_plugin():
+            if not name.startswith("conda."):
+                self.set_blocked(name)
```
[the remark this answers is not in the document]
jezdez: You're right, the name is created by https://github.com/pytest-dev/pluggy/blob/1158d387b7622a9b4d66bf8bcdc1c38effe9f497/src/pluggy/_manager.py#L255-L264 only during registration if it's not provided, e.g. when we load internal plugin, then the name will be the dotted Python path to the module of the plugin.

But for plugins provided by 3rd party packages via entrypoints, it goes through the [`CondaPluginManager.load_entrypoints`](https://github.com/conda/conda/blob/main/conda/plugins/manager.py#L67-L99) which will use the name of the entrypoint, e.g. here's the output on my machine:

```
>>> pprint.pprint(context.plugin_manager.list_name_plugin())
[('conda.plugins.solvers',
  <module 'conda.plugins.solvers' from '/path/to/conda/conda/plugins/solvers.py'>),
 ('conda.plugins.virtual_packages.archspec',
  <module 'conda.plugins.virtual_packages.archspec' from '/path/to/conda/conda/plugins/virtual_packages/archspec.py'>),
 ('conda.plugins.virtual_packages.cuda',
  <module 'conda.plugins.virtual_packages.cuda' from '/path/to/conda/conda/plugins/virtual_packages/cuda.py'>),
 ('conda.plugins.virtual_packages.freebsd',
  <module 'conda.plugins.virtual_packages.freebsd' from '/path/to/conda/conda/plugins/virtual_packages/freebsd.py'>),
 ('conda.plugins.virtual_packages.linux',
  <module 'conda.plugins.virtual_packages.linux' from '/path/to/conda/conda/plugins/virtual_packages/linux.py'>),
 ('conda.plugins.virtual_packages.osx',
  <module 'conda.plugins.virtual_packages.osx' from '/path/to/conda/conda/plugins/virtual_packages/osx.py'>),
 ('conda.plugins.virtual_packages.windows',
  <module 'conda.plugins.virtual_packages.windows' from '/path/to/conda/conda/plugins/virtual_packages/windows.py'>),
 ('conda.plugins.subcommands.doctor.cli',
  <module 'conda.plugins.subcommands.doctor.cli' from '/path/to/conda/conda/plugins/subcommands/doctor/cli.py'>),
 ('conda-libmamba-solver',
  <module 'conda_libmamba_solver.plugin' from '/path/to/miniconda/lib/python3.10/site-packages/conda_libmamba_solver/plugin.py'>)]
```

See the `conda-libmamba-solver` at the end? That's the name of the entrypoint provided by conda plugin authors.

I would suggest adding a condition to `load_entrypoints` that would check if the entrypoint name is not starting with `conda.` and continue otherwise.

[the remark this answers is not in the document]
jezdez: To clarify, basically we should be looking into making sure that when we load plugins, from 3rd party apps, that we can successfully disable them, independent of what the authors have provided as name.

There are a few ways to do that, one of which could be that we mark all built-in plugins with a name prefix such as `conda:` or something, but verify that the loaded plugin module is actually inside the `conda` package namespace as well (during disabling?). Then, when we want to disable any 3rd party plugins, we can simply disable any that don't have the `conda:` prefix in the name.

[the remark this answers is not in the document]
kenodegard: My first thought was that since `self.list_name_plugin()` tells us the `(name: str, plugin: module)`, we can just inspect the `plugin` itself and ensure it's loaded from the `conda` module. But now I'm wondering if this too could be spoofed via a [packaging namespace](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/)?

tests/plugins/test_manager.py
```diff
@@ -94,3 +98,32 @@ def test_load_entrypoints_importerror(plugin_manager, mocker, monkeypatch):
...
+    with the --no-plugins flag
+    """
+    assert not context.no_plugins
+    out, err, code = conda_cli("--no-plugins", "info")
+    assert pre_command_plugin.pre_command_action.mock_calls == []
+    assert err == ""
+
+
+def test_disable_plugins_with_env_var(
+    conda_cli, plugin_manager, mocker, monkeypatch, pre_command_plugin
```
kenodegard, line 117:
```suggestion
    conda_cli, monkeypatch, pre_command_plugin
```

kenodegard (changes requested):

kenodegard:
pre-commit.ci autofix

conda/plugins/manager.py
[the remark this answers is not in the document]
jezdez: I've tweaked the `disable_external_plugins` method a little to account for module and class based plugins and fallback to automatically disable the plugin if it can't be identified.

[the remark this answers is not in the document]
jezdez: In [`6707f1e` (#12824)](https://github.com/conda/conda/pull/12824/commits/6707f1e3509691e0fb86bf127a120c9337a78431) I mean

conda/plugins/manager.py
```diff
@@ -181,6 +182,25 @@ def yield_command_hook_actions(self, hook_type: CommandHookTypes, command: str):
...
+                # the plugin is actually a module, which should have a __name__
+                name: str | None = getattr(plugin, "__name__", None)
+            else:
+                # disable all plugins that can't be easily identified
+                # via their __module__ name (e.g. a class)
+                name: str | None = getattr(plugin, "__module__", None)
+            if (
+                name is None or
+                not name.startswith("conda.plugins")
+            ):
```
beeankha, line 201:
```suggestion
            if name is None or not name.startswith("conda.plugins"):
```

conda/plugins/manager.py
```diff
@@ -181,6 +182,22 @@ def yield_command_hook_actions(self, hook_type: CommandHookTypes, command: str):
...
+        """
+        for registered_name, plugin in self.list_name_plugin():
+            # retrieves the name of the plugin module
+            if isinstance(plugin, types.ModuleType):
+                # the plugin is actually a module, which should have a __name__
+                name: str | None = getattr(plugin, "__name__", None)
+            else:
+                # disable all plugins that can't be easily identified
+                # via their __module__ name (e.g. a class)
+                name: str | None = getattr(plugin, "__module__", None)
```
kenodegard, line 197: `module.__name__` is _very_ easily spoofed (can just say `__name__ = "foo"` in module) so `module.__spec__.name` is safer

and `object.__module__` is just a copy of `module.__name__` so it suffers from the same spoofing problem so I'm thinking we should consider doing:

```suggestion
            module = inspect.getmodule(plugin)
            if name is None:
                # probably not good to raise an exception here,
                # perhaps better to just block any plugin where
                # we cannot detect a safe name? or perhaps this
                # is the only case where we fallback to the
                # spoofable name?
                raise Exception(...)
            name = module.__spec__.name
```

beeankha:
pre-commit.ci autofix

conda/plugins/manager.py
kenodegard: resolved by https://github.com/conda/conda/pull/12869 and https://github.com/conda/conda/pull/12824/commits/a0a188ef5386c98b3334af6d5a4e40f7b246e2ff

conda/base/context.py
```diff
@@ -444,6 +444,14 @@ def experimental_solver(self):
         aliases=("conda-build", "conda_build"),
     )
 
+    ####################################################
+    #               Plugin Configuration               #
+    ####################################################
+
+    no_plugins = ParameterLoader(
+        PrimitiveParameter(NO_PLUGINS, element_type=(bool, NoneType)),
+    )
```
kenodegard, line 453: Lets simplify this to only allow True/False:

```suggestion
    no_plugins = ParameterLoader(PrimitiveParameter(NO_PLUGINS))
```

jezdez (approved):

jezdez (approved):

beeankha (approved):

kenodegard (approved):

PR #12856, opened 2023-07-08
Misleading Proxy Server Configuration in Documentation

padeoe:
### Description

I noticed a potential issue in the Conda documentation regarding the proxy server configuration.

Current example:
```
proxy_servers:
    http: http://user:pass@corp.com:8080
    https: https://user:pass@corp.com:8080
```
However, the 'https' key should also be specified using the HTTP protocol ('http://'), not HTTPS:

```
proxy_servers:
    http: http://user:pass@corp.com:8080
    https: http://user:pass@corp.com:8080
```

The `http` and `https` keys indicate the type of traffic to route, not the protocol of the proxy server itself. As written, this could mislead users into thinking that the proxy server should use the HTTPS protocol for `https` traffic.

I suggest updating the documentation to prevent confusion.

Thank you for considering this issue.

### Checklist - did you ...


travishathaway:
@conda-bot check

docs/source/user-guide/configuration/use-condarc.rst
```diff
@@ -346,7 +346,7 @@ here overrides that default:
 
   proxy_servers:
       http: http://user:pass@corp.com:8080
-      https: https://user:pass@corp.com:8080
+      https: http://user:pass@corp.com:8080
```
travishathaway, line 349:
````suggestion
      https: https://user:pass@corp.com:8080
      
```{admonition} Mixing HTTPS and HTTP
For the ``https`` key, it is okay to use a server which only accepts the HTTP protocol (e.g. ``http://user:pass@corp.com:8080``
```
````
@padeoe,

My only concern with your suggested improvement is that it may also confuse readers who would expect the `https` key to point at an HTTPS server.

I have another suggestion though (please see above). We could add an admonition directly below the example the informs users about being able to use either HTTP or HTTPS for the `https` key.

Let me know what you think.

Comment (padeoe) [reply docs/source/user-guide/configuration/use-condarc.rst:349]: Thank you for your prompt response.

I understand your concern about potentially confusing readers. However, I'd like to clarify a key point: the protocol of the proxy server (HTTP or HTTPS) is typically fixed and not interchangeable based on the type of traffic (http or https) it's routing.

For instance, if a user has an HTTP proxy server, both the http and https keys should point to an http:// URL. Similarly, if a user has an HTTPS proxy server, both keys should point to an https:// URL. It's not a mix-and-match scenario.

The current documentation might lead users to believe that they should use an https:// URL for the https key even if their proxy server only supports HTTP. This could result in misconfigurations.

To make this clearer, perhaps we could rephrase the admonition:
```{admonition} Note on Proxy Server Protocol
The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
```
Feel free to adjust the wording as you see fit!

padeoe: @travishathaway

travishathaway: @padeoe,

I think the admonition you propose makes sense. I will get the opinion of some colleagues on this too.

travishathaway, line 349:
````suggestion
      https: http://user:pass@corp.com:8080
      
```{admonition} Mixing HTTPS and HTTP
The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
```
````

**Comment by travishathaway** [comment]
@padeoe,

Just FYI, I committed you suggestion to the branch. I will get the opinion of some others now and will see if this is something we would like to merge to our documentation.

File: docs/source/user-guide/configuration/use-condarc.rst [hunk 2]
```diff
@@ -346,7 +346,11 @@ here overrides that default:
...
+      
+```{admonition} Mixing HTTPS and HTTP
+The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
+```
```
travishathaway, line 353:
````suggestion

```{admonition} Mixing HTTPS and HTTP
The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
```
````

File: docs/source/user-guide/configuration/use-condarc.rst [hunk 3]
```diff
@@ -346,7 +346,11 @@ here overrides that default:
 
   proxy_servers:
       http: http://user:pass@corp.com:8080
-      https: https://user:pass@corp.com:8080
+      https: http://user:pass@corp.com:8080
+
+```{admonition} Mixing HTTPS and HTTP
+The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
```
jezdez, line 352:
```suggestion
The protocol in the URL (either ``http://`` or ``https://``) should match the actual protocol of your proxy server. The keys ``http`` and ``https`` in the above example merely indicate the type of traffic to route, not the protocol of the proxy server itself. Ensure that both keys use the correct protocol based on your proxy server's configuration.
```

jezdez (approved):
This config option is sadly easy to get wrong, maybe it would be good to provide some real world examples or a tutorial?

kathatherine (approved):
Hi, @padeoe, thank you so much for your patience on this. Your change was a bit lost in the shuffle, but we have rediscovered it! There were some pretty big changes to the conda docs organization since your addition, so I moved your changes to the correct file they should be in. Thank you for your contribution!

kenodegard (approved):

PR #12869, opened 2023-07-11
Modify plugin loading to only use canonical names

kenodegard:
### Description

This changes how entry point plugins are registered. Instead of using the entry point name as the plugin name we use the canonical name (this is creates more consistency with how first-party plugins are loaded).

This is inspired by the spoofing concerns raised while implementing `--no-plugins` (https://github.com/conda/conda/pull/12824).

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -141,7 +141,7 @@ class CondaPluginManager(pluggy.PluginManager):
         specname = f"{self.project_name}_{name}"  # e.g. conda_solvers
         hook = getattr(self.hook, specname, None)
         if hook is None:
-            raise PluginError(f"Could not load `{specname}` plugins.")
+            raise PluginError(f"Could not find requested `{name}` plugins")
 
         plugins = sorted(
             (item for items in hook() for item in items),
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import functools
 import logging
 from importlib.metadata import distributions
+from inspect import getmodule, isclass
 
 import pluggy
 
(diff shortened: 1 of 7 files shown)


jezdez (changes requested):
We should retrofit `get_canonical_name` to also check if the registered plugin is a class (as in my suggested PR #https://github.com/conda/conda/pull/12814), instead of a module, since the default of `__name__` is pretty non-descriptive and may increase the risk of overlaps.

conda/plugins/manager.py
```diff
@@ -46,23 +47,44 @@ def __init__(self, project_name: str | None = None, *args, **kwargs) -> None:
             self.get_solver_backend
         )
 
-    def load_plugins(self, *plugins) -> list[str]:
+    def get_canonical_name(self, plugin: object) -> str:
+        # detect the fully qualified module name
+        prefix = "<unknown_module>"
+        if module := getmodule(plugin):
+            prefix = module.__spec__.name if module.__spec__ else module.__name__
```
kenodegard, line 54: Since we've taken all of this work to avoid spoofing abilities I'm also inclined to consider any module without the `__spec__` attribute to be malformed so perhaps this is better:

```suggestion
        if (module := getmodule(plugin)) and module.__spec__:
            prefix = module.__spec__.name
```

tests/plugins/test_manager.py
```diff
@@ -31,18 +33,18 @@ def conda_solvers(self):
 
 def test_load_no_plugins(plugin_manager):
     plugin_names = plugin_manager.load_plugins()
-    assert plugin_names == []
+    assert not plugin_names
```
travishathaway, line 36: I think this assertion statement would be a little clearer if we changed it to:

```suggestion
    assert plugin_names == 0
```

It would also be more consistent with the assertions in the tests below.

jezdez: Agreed

jezdez (approved):

tests/plugins/data/test-plugin/test_plugin-1.0.dist-info/entry_points.txt
```diff
@@ -1,2 +1,4 @@
-[conda]
-conda-test-plugin=test_plugin.plugin
+[test_plugin]
```
kenodegard, line 1: I believe this change fixes the noisy logging that would bleed over into other tests:

```
WARNING conda.plugins.manager:load_entrypoints(86): Could not load conda plugin `conda-test-plugin`:

No module named 'package_that_does_not_exist'
```

kenodegard: cc @kalawac

tests/plugins/test_manager.py
```diff
@@ -86,11 +86,11 @@ def test_load_plugins_error(plugin_manager):
     plugin_manager.load_plugins(VerboseSolverPlugin)
     # then try again to trigger a PluginError via the `ValueError` that
     # pluggy.PluginManager.register throws on duplicate plugins
-    with pytest.raises(PluginError) as exc:
+    with pytest.raises(
+        PluginError, match="Error while loading first-party conda plugin"
+    ):
```
jezdez, line 91: 😍

travishathaway (approved):

PR #12887, opened 2023-07-14
Build ePub, PDF and HTML zip formats as well.

### LINKED ISSUE #12886: Provide downloadable epub and PDFs versions of the docs
### Checklist

### What is the idea?

It would be good to re-enable the EPUBs and PDFs for the conda docs.

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

jezdez:
Fix #12886

### Checklist - did you ...

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -23,7 +23,7 @@ python:
   install:
     - requirements: docs/requirements.txt
 
-# Build PDF & ePub
+# Build PDF, ePub and zipped HTML
 formats:
   - epub
   - pdf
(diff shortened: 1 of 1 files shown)


kathatherine (approved):
This looks good to me and according to the RTD docs, all that needs to happen. So simple! Is this something we would want to do for all the versions that are missing the downloads? I guess we'd have to rebuild them all once. idk if that's something we want to undertake. Going forward, this'll be good, though.

jezdez:
I would stick to the release branches for those, e.g. 23.5.x, since we don't enable all tags, there are like 150 of those and I don't think it adds anything of value to end users.

Issue #11921, opened 2022-10-03
Implement user authentication for trust metadata


AMWatkins314:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Conda does not support authentication for the signature verification files `1.root.json` or `key_mgr.json` which are pointed to by the URL `signing_metadata_url_base` (as defined in condarc)
See: 
- https://github.com/conda/conda/blob/main/conda/trust/signature_verification.py#L109-L114
`# TODO: support fetching root data with credentials`

- https://github.com/conda/conda/blob/main/conda/trust/signature_verification.py#L144-L148

- https://github.com/conda/conda/blob/main/conda/trust/signature_verification.py#L187-L201
  ```
  # TODO: Figure how to handle authn for obtaining trust metadata,
  # independently of the authn used to access package repositories.
  ```

### Why is this needed?

The condarc file allows users to specify where the metadata files live via `signing_metadata_url_base`.
The `signing_metadata_url_base` can point to a location that requires user authentication.
An example of this is a repository configured to require user authentication, like Nexus or Artifactory.

### What should happen?

The user configures`signing_metadata_url_base` to point to a URL requiring user authentication. The user authentication can either be specified in a netrc file or as basic auth embedded in the URL (https://user:pwd@fqdn). When the user runs `conda install` or `conda create` the user authentication processes, the user is granted access, and packages that would be installed show the correct package signature verification information.

awwad:
There are two reasons I know of to work on this:
- This is currently affecting users of conda / repository infrastructure who require that all contact with the server include user authentication.
- Consolidation of http request code / file download code in conda seems like a worthwhile goal.  (There are probably other issues with having duplicated `requests` calls.)

jezdez:
This is certainly something achievable given conda's network I/O gateway system and our currently on-going work to make this pluggable (e.g. to be extended with additional authentication schemes). I would consider the fetching of content trust keys to be a specific application of the problem here.

jezdez:
@AMWatkins314 Apologies for the long pause in communication, we've implemented the fix in #12889 but sadly missed the merge window for the July release of conda. The next _planned_ release is in September.

jezdez:
@AMWatkins314 FYI this'll be fixed in the September release cycle in ~1 month.

PR #12889, opened 2023-07-16
Directly suppress use of binstar token when fetching trust metadata

chenghlee:
### Description

Drop `lambda` weirdness and simply (but temporarily) disable use of the conda token mechanism when fetching trust metadata.  One useful side effect of doing this is that it's now possible to use standard HTTP "Basic" auth mechanisms and `$HOME/.netrc` to protect access to such trust metadata, which is a requirement in certain enterprise repository/mirroring use cases.

This should resolve conda/conda#11921.

(Note, however, that this PR only supports HTTP "Basic" auth via the `signing_metadata_url_base` URL or the user's netrc file. It is *not* a generic solution for authentication, which will have to wait on the networking plug-in work to be completed.)

### Checklist - did you ...


conda/trust/signature_verification.py
```diff
@@ -193,23 +193,27 @@ def _fetch_channel_signing_data(
...
+            # Seems reasonable, since we (probably) don't want the headaches of
+            # dealing with protected, per-channel trust metadata.
             #
-            # TODO: Figure how to handle authn for obtaining trust metadata,
-            # independently of the authn used to access package repositories.
+            # Note: Setting `auth=None` here does allow trust metadata to be
+            # protected using standard HTTP basic auth mechanisms, with the
+            # login information being provided in the user's netrc file.
+            saved_token_setting = context.add_anaconda_token
+            context.add_anaconda_token = False
```
jezdez, line 205: I worry a little about thread-safety here, since context is just a module attribute and behind `self.session` is the LRU cached instance of the `CondaSession` class, which behind the scenes runs the requests/urllib3 machinery, but is using a thread local to store session instances per thread.

E.g. in case an exception occurs during `self.session.get`, it won't reset `context.add_anaconda_token` again, which may or may not have issues downline when it arrives at the exception handler. Would a 

```python
saved_token_setting = context.add_anaconda_token
context.add_anaconda_token = False
try:
    ...
finally:
    context.add_anaconda_token = saved_token_setting
```

chenghlee: Ooohhh...good catch. Modified the PR to add the suggested `finally` block (and to remove the basically pointless `except`).

jezdez (approved):

jezdez (approved):

PR #12911, opened 2023-07-21
Auth handler plugin hook

travishathaway:
### Description

Fixes: #12018

This pull request represents a working prototype that is ready for feedback.

### Example use case

To see how this new plugin hook could be used, please check out the plugin I'm currently authoring called, "conda-auth" over here -> https://github.com/travishathaway/conda-auth. That plugin makes use of the [keyring](https://github.com/jaraco/keyring) package to securely store credentials.

In order to make use of this plugin, you need to first configure it on a per-channel basis. This is what that looks like in the `.condarc` file for a channel configured to use HTTP basic auth:

```yaml
channels:
  - http://localhost:8042
channel_settings:
  - channel: http://localhost:8042
    auth: conda-basic-auth
    username: user
```

After that's setup up you can run an `install` or `create` command. The first time it's run, the program start-up with be interrupted by the `conda_pre_commands` hook to ask for a password. Subsequent runs will not ask for a password because it has been stored in the user's keyring via the [keyring](https://github.com/jaraco/keyring) package.

### Checklist - did you ...


tests/gateways/test_connection.py
```diff
@@ -108,3 +116,51 @@ def test_s3_server(minio_s3_server):
...
+    condarc_data = {"channels": channels}
+    yml = YAML()
+    condarc_stream = StringIO()
+    yml.dump(condarc_data, stream=condarc_stream)
+    condarc_stream.seek(0)
+
+    with temp_context(condarc_stream.read()):
+        channel_name = get_channel_name_from_url(url)
+
+        assert expected == channel_name
```
kenodegard, line 166: all of these tests can be achieved using the `CONDA_CHANNELS` environment variable so I recommend not introducing `temp_context`:

```suggestion
    monkeypatch.setenv("CONDA_CHANNELS", ",".join(channels))
    reset_context()
    assert context.channels == channels

    assert expected == get_channel_name_from_url(url)
```

travishathaway: Thanks. I will check out your suggestion.

jezdez:
I'm concerned about the design proposed here, as it adds another level of abstraction on top of `requests`, but doesn't align with the other levels or even recognize that they exist:

- conda.gateways.adapters.* -- custom `requests` adapters to handle the network I/O

  I think it'd be great to have a new plugin hook dedicated to the gateway adapters, e.g., so we could easily separate the maintenance of the FTP and S3 adapters into separate packages, e.g., `conda-ftp-adapter` and `conda-s3-adapter`.

  Each might also provide additional auth classes, of course, so `conda-ftp` and `conda-s3` might actually be more fitting.

- `conda.gateways.connection.session.CondaHttpAuth` -- a custom `requests` auth class that actually implements anaconda.org specific authentication

  This is a tricky one, as it does two things in one, but requests' auth system allows multiple authentication backends (via the `auth` attribute) to be configured per session object.

  Could we, for ease of migration, split these up into two classes (`URLTokenAuth`/"binstar" and `URLBasicAuth`/password in URL) and set them as default, but allow them to be overridden by classes provided via a plugin hook?

  This might be useful if we created a separate `conda-auth` package that comes with a `auth` subcommand (think: `conda auth login https://anaconda.org/my-channel` etc) and the various authentication backends.

  The upside: separation of maintenance and being able to implement different authentication mechanisms in one place.

  To slightly tweak your example above:

  ```
  channels:
  - http://localhost:8042/
  channel_settings:
  - channel: http://localhost:8042/
    auth:
      - url-token-auth
      - keyring-basic-auth
  ```

  `url-token-auth` and `keyring-basic-auth` would be the names of the `requests` auth backends provided via the plugin system.

jezdez (commented):
@travishathaway I've sent a whole other PR with changes to your fork (https://github.com/travishathaway/conda/pull/1) that I think are needed to tweak the naming and functionality of this. But this is clearly "just before the finish line" type feedback!

jezdez:
pre-commit.ci autofix

jezdez (dismissed):
This is good to go in my opinion, assuming we do the usual round of doc updates, blog post, etc. I also take it that we'll have to tweak the CEP?

conda/gateways/connection/session.py
```diff
@@ -60,6 +64,48 @@ def close(self):
...
+
+    # If for whatever reason a channel name can't be determined, (should be unlikely)
+    # we just return the default session object.
+    if channel_name is None:
+        return CondaSession()
+
+    channel_settings = {}
+    for settings in context.channel_settings:
+        if settings.get("channel") == channel_name:
+            channel_settings = settings
```
dholth, line 93: tests?

travishathaway: Nice catch. I just added another test that should bump the coverage of this function to 💯.

conda/plugins/manager.py
```diff
@@ -202,6 +203,16 @@ def get_solver_backend(self, name: str | None = None) -> type[Solver]:
 
         return backend
 
+    def get_auth_handler(self, name: str) -> type[AuthBase] | None:
+        """
+        Get the auth handler with the given name or None
+        """
+        auth_handlers = self.get_hook_results("auth_handlers")
+        matches = tuple(item for item in auth_handlers if item.name == name.strip())
```
kenodegard, line 211: For both the solver and subcommands we only compare `str.lower()` names, do we want to do the same here?

travishathaway: Yeah, that makes sense.

conda/gateways/connection/session.py
```diff
@@ -60,6 +64,61 @@ def close(self):
...
 
 
+def get_channel_name_from_url(url: str) -> str | None:
+    """
+    Given a URL, determine the channel it belongs to and return its name.
+    """
+    for channel in get_channel_objs(context):
+        for base_url in channel.base_urls:
+            if url.lower().startswith(base_url):
+                return channel.canonical_name
```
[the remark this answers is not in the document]
travishathaway: hmm, I would have thought that those would appear in channels too.. 😦

If that's not the case, then I'm not sure how we would support that...

beeankha (approved):

conda/gateways/connection/session.py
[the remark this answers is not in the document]
travishathaway: @kenodegard,

I thought about this some more today, and I don't believe this is a show stopper. I did some digging around and this is what I found...

When you pass something to `MatchSpec` that looks like this: `http://localhost/channel::package`, the URL is stripped away from the package name, which indicates to me that it is not meant to be used with URLs.

I am pretty sure this is only supposed to be used with channel names, and follow this approach I was able to get everything working. Below is how I set up my `.condarc` file by making use of the `channel_alias` setting:

```yaml
channels:
  - channel

channel_alias: http://localhost:8041

channel_settings:
  - channel: channel
    auth: conda-auth-oauth2
    login_url: http://localhost:8041/login
```

When you run the following command, everything works as expected (i.e. the correct auth handler is used):

```
conda create -n tester channel::travis_test
```

I actually think that this use case closely mirrors what actual users with private channels will be doing.

Thanks for bringing this up though. This was an important use case to think through. Please let me know if I have missed anything else important to consider.

conda/plugins/types.py
```diff
@@ -102,3 +104,30 @@ class CondaPostCommand(NamedTuple):
...
+    """
+    Base class that we require all plugin implementations to use to be compatible.
+
+    Authentication is tightly coupled with individual channel; therefore, an additional
+    ``channel_name`` property must be set on the ``requests.auth.AuthBase`` class.
+    """
+
+    def __init__(self, channel_name: str):
+        self.channel_name = channel_name
+        super().__init__()
```
jezdez, line 119: Will this make it harder to re-use existing requests auth handler classes? I would suggest to provide a mixin class for those cases.

```suggestion
class ChannelAuthMixin:
    """
    Class mixin to make all plugin implementations that use an existing
    (e.g. 3rd party) requests authentication handler compatible.
    
    Please use the concrete :class:`conda.plugins.types.ChannelAuthBase`
    in case you're creating an own implementation.
    """

    def __init__(self, channel_name: str):
        self.channel_name = channel_name
        super().__init__()


class ChannelAuthBase(ChannelAuthMixin, AuthBase):
    """
    Base class that we require all plugin implementations to use to be compatible.

    Authentication is tightly coupled with individual channel; therefore, an additional
    ``channel_name`` property must be set on the ``requests.auth.AuthBase`` class.
    """
```

conda/gateways/connection/session.py
```diff
@@ -107,6 +107,18 @@ def session_manager(url: str):
...
+def get_session_storage_key(auth) -> str:
+    """
+    Function that determines which storage key to use for our CondaSession object caching
+    """
+    if auth is None:
+        return "default"
+
+    auth_type = type(auth)
+
+    return f"{auth_type.__module__}.{auth_type.__qualname__}::{auth.channel_name}"
```
jezdez, line 119: Remember that in requests lore, auth can also be a tuple, this might break that contract.

conda/plugins/types.py
travishathaway: I thought about following the mixin pattern here first, but then realized we would have two classes both wanting to define their own `__init__` method. That won't work, right? 😅

conda/gateways/connection/session.py
jezdez: Yep, that should do the trick!

conda/plugins/types.py
jezdez: Yeah, indeed, the method resolution order matters, so using `super()` is needed and following the right order. But since this is an API that would be used not rarely and the authors will likely be advanced enough IMO, I think it's not a problem to expect that.

conda/gateways/connection/session.py
```diff
@@ -70,21 +132,30 @@ def __new__(mcs, name, bases, dct):
...
+            session = cls._thread_local.sessions[storage_key] = super().__call__(
+                **kwargs
+            )
+
+        return session
 
 
 class CondaSession(Session, metaclass=CondaSessionType):
-    def __init__(self):
+    def __init__(self, auth: AuthBase | None = None):
```
beeankha, line 152: Since we add a check in `get_session_storage_key()` to see if `auth` is a tuple, should we update the type hinting here to include that possibility?
```suggestion
    def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
```

conda/gateways/connection/session.py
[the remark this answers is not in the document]
kenodegard: `https://conda.anaconda.org/conda-forge/noarch::nose2` is a valid package spec:

`conda create -n scratch https://conda.anaconda.org/conda-forge/noarch::nose2`

But even if thats found to be irrelevant I still think doing `conda.models.channel.Channel.from_url(url).canonical_name` will be more robust instead of using this custom `get_channel_name_from_url(url)` implementation

jezdez:
pre-commit.ci autofix

jezdez:
pre-commit.ci autofix

kenodegard (approved):

jezdez (approved):

beeankha (approved):

PR #12914, opened 2023-07-21
lazy-load boto3 to save ~400ms startup time

dholth:
### Description

Address #12549 (partial).

On my machine using the scalene profiler:

```
. dev/start.sh -p 3.10
conda install scalene
echo "import conda.__main__" > con.py
python3 -m scalene con.py
```

`import boto3` takes about 350-450ms if boto3 is installed.

Instead, only import when s3 channels are actually used. If the user happens to have boto3 installed at all in their base environment, but doesn't use s3 channels, this should speed things up.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/adapters/s3.py
+++ b/conda/gateways/connection/adapters/s3.py
@@ -6,17 +6,26 @@ from logging import LoggerAdapter, getLogger
 from tempfile import SpooledTemporaryFile
 
 have_boto3 = have_boto = False
-try:
-    import boto3
 
-    have_boto3 = True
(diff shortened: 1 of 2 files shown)


kenodegard (commented):
Never been a fan of the imports with `have_boto` booleans. Can we move the import statements into the `S3Adapter.send` method and get rid of the `have_boto` booleans?

kenodegard (approved):

jezdez:
Nicely done, I've wanted to 🔥 that code for a LOONG time.

PR #12919, opened 2023-07-24
replace inspect.stack() with sys._getframe() to save ~170ms -> ~90ms

dholth:
### Description

We only use a single frame, so use `sys._getframe` instead of `inspect.stack()`.

### Checklist - did you ...

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -3,6 +3,7 @@
 """Tools to aid in deprecating code."""
 from __future__ import annotations
 
+import sys
 import warnings
 from argparse import Action
 from functools import wraps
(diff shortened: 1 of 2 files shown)


kenodegard:
[Per the docs](https://docs.python.org/3/library/sys.html):

> **CPython implementation detail:** This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.

Does this mean we should include a fallback to `inspect.stack`?

dholth:
No, it means we should gracefully not show deprecation messages if `sys._getframe` is not available. Not likely to find a Python that has `inspect.stack()` but not `sys._getframe`

jezdez (approved):
Looking at the [GitHub search for `sys._getframe`](https://github.com/search?q=sys._getframe+language%3APython&type=code&l=Python), it seems as if quite a few rely on it, I think this should be fine®

jezdez (approved):

Issue #5253, opened 2017-05-08
conda vs conda env has too much confusion


jgmorrison:
Many issues have been brought up by customers using different methods we have put forward. This leads to confusion on when to use conda or conda env. 

Here are some examples of duplicate actions:

list available environments:

    conda info -e
    conda env list

The following has had failures with "conda env" and has been suggested to use the "--all" flag with conda instead, https://github.com/conda/conda/issues/3215. Two ways are confusing especially if one doesn't always work.

    conda env remove -n env_name
    conda remove -n env_name —-all

The docs suggest the following  to build identical environments, https://conda.io/docs/using/envs.html#build-identical-conda-environments

    conda list -n test —explicit > testexp.txt
    conda create -n test —file testexp.txt

But suggest sharing environments with the following, https://conda.io/docs/using/envs.html#share-an-environment

    conda env export -n env_name > environment.yml
    conda env create —file environment.yml

JeremyEnns:
If we are no longer developing conda env should we deprecate it and consolidate our docs around the most up-to-date way?

ijstokes:
@JeremyEnns who says that we're deprecating `conda env`? @kalefranz is that true? I use it all the time.  I think many people do.

@jgmorrison I don't disagree that it is confusing.  It was my wish that the thing that became `anaconda-project` would provide a single-system tiered-syntax that would do away with the confusion from `conda env` environments (`conda env export`) and `conda` environments (`conda list -e`), giving us just one command and one syntax.  Instead we've ended up with 3...

kalefranz:
> who says that we're deprecating `conda env`?

Not me.

However, it hasn't gotten the attention from me that the core conda code has.  At some point this will change.  I have every intention to actually integrate them more tightly, especially now that they're in the same repo.

All of the current ways of doing things (e.g. `conda info -e` and `conda env list`) will stay functional for a long time.  We will give ample deprecation warnings before we shut anything down, and those warnings will also teach the *right* way to do things, whatever that ends up being.

I agree that confusion between conda and conda-env is still an area I'd classify as a mess that needs cleaned up.

Kharms:
Which of those two is recommended as of now?

jkleckner:
I have had conda environments get quite tangled and unable to recreate them from a spec-file.  Is there a recommended flow to avoid this?

soapy1:
I hope this documentation will make this more clear https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html

[...]

PR #12952, opened 2023-08-01
Create env from fsspec uri

AlbertDeFusco:
### Description

This allows an [FSSpec](https://filesystem-spec.readthedocs.io/en/latest/index.html) URI to be provided as a `remote_definition` to `conda env create`.

This allows reading data using any of the [built-in protocols](https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations) or from a growing list of [3rd party packages](https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations) that can be installed.

If the remote definition is not of the form `protocol://path` it will attempt to find it using the BinstarSpec.

Here are some examples using built-in drivers:

```text
conda env create https://github.com/AlbertDeFusco/windturbines/raw/main/environment.yml
conda env create github://albertdefusco:windturbines@main/environment.yml
```

### Checklist - did you ...


conda_env/specs/__init__.py
```diff
@@ -13,6 +13,7 @@
 from conda.gateways.connection.session import CONDA_SESSION_SCHEMES
 
 from .binstar import BinstarSpec
```
dholth, line 15: BinstarSpec!

jezdez:
Hey @AlbertDeFusco, thanks for the PR! Could you please open a feature request issue, so we can discuss if this is something we want to add?

Regarding your proposal, on first glance: Both `fsspec` and `aiohttp` are non-trivial dependencies and given #11633 (and #5253 really), I don't think this should be a feature added to conda as proposed. Basically we're trying to merge conda-env into conda, not add to it.

Here's my proposal:

- Both the existing (e.g. `binstar`) and the proposed `fsspec` environment specs could be served by a new plugin hook (e.g. named `env_specs`) via #13015
- conda could then only ship with the requirements.txt and environment.yaml specs as the "default" specs
- binstar could be provided by a `anaconda-cloud` plugin (or similar)
- fsspec could be provided by a `conda-fsspec` plugin, maintained in the conda community organization
- this would leave the door open to all kinds of remote service env specs

AlbertDeFusco:
Perfect. I'm happy to migrate this to an extension.

Here's the issue: https://github.com/conda/conda/issues/13017

conda_env/specs/remote.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Remote spec using fsspec"""
+from functools import cached_property
+
+import fsspec
+from fsspec import available_protocols
+from fsspec.core import split_protocol
```
jjhelmus, line 8: This will require `fsspec` to be a required dependency of conda.  
If this is not desired the import of `RemoteSpec` could be delayed until needed in the __init__.py file so that the dependency is optional and only required if a remote spec is needed.

AlbertDeFusco: I'm cool with that if it might get this going faster while the `conda env` effort continues

jezdez:
We've refactored the environment handling a lot and I think this needs another approach as a result.

Issue #12977, opened 2023-08-08
Toggling logger verbosity causes issues for test suite


kenodegard:
### What happened?

Ran into into some funky verbosity/logging issues because of our mixed usage of what `--verbose` means (both `conda clean` and `conda doctor` expose this). Historically `--verbose` was exclusively used to indicate increasing the logging verbosity but recently we’ve also used it to print out more detailed info (compare for example, `conda info --all` versus `conda info --verbose`).

To replicate this today on main run the following, note the test order is important:

```
pytest \
    tests/plugins/subcommands/doctor/test_cli.py::test_conda_doctor_happy_path_verbose \
    tests/cli/test_main_run.py::test_run_returns_nonzero_errorlevel
```

The issue is caused by `--verbose` being treated as an alias for `context.verbosity` which sets the logging verbosity in `conda.cli.main::main_subshell` (see `init_loggers`) and toggling the verbosity/logging in our test suite results in stream closed errors as the old handlers are replaced with new ones (I think this happens [here](https://github.com/conda/conda/blob/c691f729a942ab0a3303424a31d0e1e5b07143d9/conda/common/io.py#L320-L341)):

```
Traceback (most recent call last):
  File "/Users/kodegard/.conda/devenv/Darwin/arm64/envs/devenv-3.10-c/lib/python3.10/logging/__init__.py", line 1103, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
```

I believe there are two possible fixes:
1. don’t replace logging handlers, rather mutate them in place (not 100% sure if this is safe to do)
2. don’t use `--verbose` to print more info, introduce a new argument like `--detailed`

### Prior Discussions

- https://github.com/conda/conda/issues/12420

ForgottenProgramme:
>don’t use --verbose to print more info, introduce a new argument like --detailed

I like this idea.

PR #12980, opened 2023-08-08
Introducing `--details` for `info`, `clean`, and `doctor`

kenodegard:
### Description

Since `--verbose` is an alias that correlates to `context.verbosity` which is used to set the logging verbosity we realized we shouldn't be using this argument to indicate "more info". We've previously found that the prior "more info" argument (`--all`) is ill suited as it cannot be used universally (e.g. `conda clean --all` means something completely different). So instead let's deprecate these older inconsistent usages in favor of `--details`:

| Old | New |
|---|---|
| `conda info [-a\|--all]` | `conda info [-D\|--details]` |
| `conda clean [-v\|--verbose]` | `conda clean [-D\|--details]` |
| `conda doctor [-v\|--verbose]` | `conda doctor [-D\|--details]` |

Resolves https://github.com/conda/conda/issues/12977
Alternative to #12985

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -13,6 +13,7 @@ from argparse import (
     _CountAction,
     _HelpAction,
     _StoreAction,
+    _StoreTrueAction,
 )
 from importlib import import_module
 from logging import getLogger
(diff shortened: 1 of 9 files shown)


kenodegard:
Closing in favor of #12985

PR #12985, opened 2023-08-09
Inject new detailed verbosity level

### LINKED ISSUE #12977: Toggling logger verbosity causes issues for test suite

kenodegard:
### Description

Previously verbosity exclusively controlled the logging levels. This has caused issues as we've wanted to offer detailed outputs without modifying the logging levels (e.g. `conda clean --verbose`, `conda info --all`, and `conda doctor --verbose`). Each of these implementations breaks something:

- `conda clean` has a internal differentiation between verbose and verbosity (huh?!?!)
- `conda info` has long used `--all` which is counter to UX standards
  - First introduced in conda 1.4 in https://github.com/conda/conda/commit/f696c8e4cf559f59458dd9a9df29f3bb435b53eb
- `conda doctor` treats detailed output and the first verbosity level for logging as the same thing

Resolves #12977
Resolves https://github.com/conda/conda/issues/12420
Alternative to #12980

### Checklist - did you ...


conda/cli/conda_argparse.py
```diff
@@ -2076,3 +2058,22 @@ def add_parser_default_packages(p):
...
+            "Can be used multiple times. Once for detailed output, twice for INFO logging, "
+            "thrice for DEBUG logging, four times for TRACE logging."
+        ),
+        dest="verbosity",
+        default=NULL,
+    )
+    parser.add_argument(
+        "--debug",
+        action="store_true",
+        help=SUPPRESS,
```
kenodegard, line 2078:
```suggestion
        help=SUPPRESS,
        default=NULL,
```

conda/gateways/logging.py
```diff
@@ -21,8 +21,13 @@
 from ..common.io import _FORMATTER, attach_stderr_handler
 
 log = getLogger(__name__)
-TRACE = 5  # TRACE LOG LEVEL
-VERBOSITY_LEVELS = (WARN, INFO, DEBUG, TRACE)
+VERBOSITY_LEVELS = {
+    0: WARN,  # standard output
+    1: WARN,  # -v, detailed output
```
kenodegard, line 26:
```suggestion
    0: 100,  # standard output
    1: 100,  # -v, detailed output
```

conda/base/context.py
```diff
@@ -955,7 +955,12 @@ def binstar_upload(self):
 
     @property
     def verbosity(self):
-        return 2 if self.debug else self._verbosity
+        #                   0 = logging.WARN, standard output
+        #           -v    = 1 = logging.WARN, detailed output
+        #           -vv   = 2 = logging.INFO
+        # --debug = -vvv  = 3 = logging.DEBUG
+        #           -vvvv = 4 = logging.TRACE
+        return 3 if self.debug else self._verbosity
```
jezdez, line 963: This doesn't actually conform with the logging module, `logging.WARN` was long renamed to `logging.WARNING` and the levels are two digit integers.

```suggestion
        #                    0 = logging.WARNING, standard output
        #           -v    = 10 = logging.WARNING, detailed output
        #           -vv   = 20 = logging.INFO
        # --debug = -vvv  = 30 = logging.DEBUG
        #           -vvvv = 40 = logging.TRACE
        return logging.DEBUG if self.debug else self._verbosity
```

BTW `logging.TRACE` is only added in `conda/gateways/logging.py`, so technically isn't `4` or `40` but `5`!

conda/cli/main.py
```diff
@@ -10,14 +10,14 @@ def init_loggers(context=None):
...
     if context and context.json:
         # Silence logging info to avoid interfering with JSON output
         for logger in ("conda.stdout.verbose", "conda.stdoutlog", "conda.stderrlog"):
             getLogger(logger).setLevel(CRITICAL + 1)
 
-    if context:
-        if context.verbosity:
-            set_verbosity(context.verbosity)
+    # set verbosity level, if none set default to no verbosity
+    set_verbosity(context and context.verbosity or 0)
```
jezdez, line 20:
```suggestion
    set_verbosity(context and context.verbosity or NOTSET)
```

But it needs an additional import of `NOTSET` above.

conda/cli/main_search.py
```diff
@@ -29,7 +29,7 @@ def execute(args, parser):
     if args.envs:
         with Spinner(
             "Searching environments for %s" % spec,
-            not context.verbosity and not context.quiet,
+            context.verbosity < 2 and not context.quiet,
```
jezdez, line 32: The `2` should be a constant to easier understand this, e.g.:

```suggestion
            context.verbosity < logging.INFO and not context.quiet,
```

conda/gateways/logging.py
```diff
@@ -21,8 +21,13 @@
...
+    2: INFO,  # -vv, info logging
+    3: DEBUG,  # -vvv, debug logging
+    4: (TRACE := 5),  # -vvvv, trace logging
+}
```
jezdez, line 30: I think this translation between `logging` and our CLI/logging setup will invite incompatibility when integrating plugins into conda. For better or worse, we need to use the logging levels the logging library provides, as it's the lowest common denominator when writing stdlib based packages.

In logging lore, `0` means `NOTSET`, meaning that everything should be logged. I think that's close to what we mean as “standard output”.

jezdez (changes requested):
This needs some consolidation between logging and our CLI levels, since it's essentially a matter of following best practices that the stdlib has set to make it easy for 3rd party packages/plugins to integrate well.

- we need to refer to all levels with their `logging` constants
- we can use `logging.NOTSET` as meaning the "standard output"
- our "TRACE" level should be 5 (too low)

conda/base/context.py
jezdez: A `--trace` CLI options (even if hidden), might be useful as well.

kenodegard: The verbosity values are simply a count of the number of `-v`s present (see our custom `conda.cli.conda_argparse.NullCountAction`). So my comments here are better written as:

```suggestion
        #                   0 → logging.WARN, standard output
        #           -v    → 1 → logging.WARN, detailed output
        #           -vv   → 2 → logging.INFO
        # --debug = -vvv  → 3 → logging.DEBUG
        #           -vvvv → 4 → logging.TRACE
        return 3 if self.debug else self._verbosity
```

What if we leave `context.verbosity` as is (make it easy to say `if context.verbose`) and then introduce a new `context.log_level` that maps the verbosity to the logging level?

jezdez: > What if we leave context.verbosity as is (make it easy to say if context.verbose) and then introduce a new context.log_level that maps the verbosity to the logging level?

I like that!

conda/gateways/logging.py
kenodegard: this mapping is marked for deprecation, see `context.log_level` instead

conda/cli/main.py
kenodegard: this oddity has been elimited by deprecating passing `context` into `init_loggers`

conda/cli/main_search.py
kenodegard: we introduce helpful aliases instead, the aliases produce much more readable code:

```
context.verbose and not context.quiet
```

jezdez (approved):
Wonderful, I think this is a great improvement!

jezdez (approved):

PR #12994, opened 2023-08-10
Handle jlap response without request e.g. for file:///

### LINKED ISSUE #12963: jlap: error with file:/// repositories
### Checklist

### What happened?

`--experimental=jlap` errors when pointed to a `file:///` repository containing `.jlap`:

For example,

`.condarc`:
```
channels:
   - file:///Users/dholth/prog/conda/repo
experimental: [jlap]
```

In the `repo` folder, download `repodata.json`, `repodata.json.zst`, `repodata.jlap` into noarch/ and your arch (e.g. osx-arm64/) folders. A copy from defaults is fine for testing.

Now try something like a `conda search` a few times until the second or third time where it would normally consult `.jlap` (which it only does after fetching `repodata.json`).

An error appears since the "pretend requests can talk to file" system doesn't understand headers.

```
      File ".../conda/gateways/repodata/jlap/fetch.py", line 131, in request_jlap
        log.debug("request headers: %s", pprint.pformat(response.request.headers))
    AttributeError: 'NoneType' object has no attribute 'headers'
```

### Conda Info

Version 23.7.2

dholth:
### Description

Fix #12963 

Test needs to be updated to point to the same place the test http server is getting its files; do we need to test with and without .jlap on the filesystem; do we work correctly when there is a range request or do we detect that those are unsupported; or should we simply not do the jlap bits when we are not on http or s3.

### Checklist - did you ...


tests/gateways/test_jlap.py
```diff
@@ -278,6 +320,8 @@ def test_jlap_sought(
     package_repository_base: Path,
 ):
     """Test that we try to fetch the .jlap file."""
+    (package_repository_base / "osx-64" / "repodata.jlap").unlink(missing_ok=True)
```
jezdez, line 323: Is this just a precaution? Or does this point to a larger issue of test cleanup failing elsewhere?

jezdez (approved):

tests/gateways/test_jlap.py
dholth: The test web server thread points to a single directory. We could use other strategies like copying those files into a per-test directory or creating a per-test scoped fixture to cleanup.

PR #13000, opened 2023-08-17
Add healthcheck for checking whether the environment is listed on environments.txt

ForgottenProgramme:
As an exploratory step towards converting `conda doctor` in to a plugin hook, this PR adds a new healthcheck which has a different structure from the previous two healthchecks (missing files, altered files). 
It's different because this health check doesn't require looping over the files in the `conda-meta` dir. Also it returns a `bool` versus a `dict`. 

Such an exploration of the possible types of healthchecks will allow us to decide whether the `conda doctor` plugin hook should require a standardized pattern and return type for the plugins being registered with it or not. 

### Checklist - did you ...


jezdez:
13000!! 🎊

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -20,6 +20,18 @@ def display_report_heading(prefix: str) -> None:
     print(f"Environment Health Report for: {Path(prefix)}\n")
 
 
+def check_envs_txt_file(prefix: str | Path) -> bool:
+    """Checks whether the environment is listed in the environments.txt file"""
+    home = Path.home()
+    envs_txt_file = home / ".conda" / "environments.txt"
```
kenodegard, line 26: Use `conda.core.envs_manager.get_user_environments_txt_file()` instead.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -107,3 +119,9 @@ def display_health_checks(prefix: str, verbose: bool = False) -> None:
                 print(f"{package_name}: {len(altered_files)}")
     else:
         print(f"{OK_MARK} There are no packages with altered files.\n")
+
+    print("3. Environment listed in environments.txt file:\n")
+    if check_envs_txt_file(prefix):
+        print("Yes\n")
+    else:
+        print("No\n")
```
kenodegard, line 127: Let's keep this as concise as possible:

```suggestion
    present = OK_MARK if check_envs_txt_file(prefix) else EX_MARK
    print("3. Environment listed in environments.txt file: {present}\n")
```

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -83,6 +85,22 @@ def env_altered_files(env_ok: tuple[Path, str, str, str]) -> tuple[Path, str, st
...
+    """Test that runs for the case when the env is listed on the environments.txt file"""
+    prefix, _, _, _ = env_ok
+    tmp_envs_txt_file = tmp_path / "envs.txt"
+    with open(tmp_envs_txt_file, "w") as f:
+        f.write(f"{prefix}")
+
+    mocker.patch(
+        "conda.core.envs_manager.get_user_environments_txt_file",
+        return_value=tmp_envs_txt_file,
+    )
```
kenodegard, line 100: This actually patches the "wrong" function.

When we import a function as done above:

```python
from conda.core.envs_manager import get_user_environments_txt_file
```

We end up making a unique _pointer_ to that specific instance of the `get_user_environments_txt_file` function.

This means that when we mock the actual `get_user_environments_txt_file` function only the function defined at `conda.core.envs_manager` is mocked and not the one at `conda.plugins.subcommands.doctor.health_checks`.

So there are two solutions to this, (1) mock the `get_user_environments_txt_file` in `conda.plugins.subcommands.doctor.health_checks`:

```suggestion
    mocker.patch(
        "conda.plugins.subcommands.doctor.health_checks.get_user_environments_txt_file",
        return_value=tmp_envs_txt_file,
    )
```

Or (2) change the import in `conda.plugins.subcommands.doctor.health_checks` to be:

```python
from conda.core import envs_manager
...
envs_txt_file = envs_manager.get_user_environments_txt_file()
```

kenodegard (changes requested):
This mocking exercise exposed a design issue with how in `conda/plugins/subcommands/__init__.py` we write:

```python
from .doctor import cli as doctor
```

This results in the submodules within the `conda.plugins.subcommands.doctor` module becoming inaccessible, try running the following and see how even though the import works the code path doesn't:

```python
>>> import conda.plugins.subcommands.doctor.health_checks
>>> conda.plugins.subcommands.doctor.health_checks
  File "<string>", line 2, in <module>
AttributeError: module 'conda.plugins.subcommands.doctor.cli' has no attribute 'health_checks'
```

kenodegard (changes requested):

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -7,19 +7,37 @@
...
+def check_envs_txt_file(prefix: str | Path) -> bool:
+    """Checks whether the environment is listed in the environments.txt file"""
+    try:
+        envs_txt_file = Path(get_user_environments_txt_file())
+        content = envs_txt_file.read_text()
+        if str(prefix) in content:
+            return True
+        else:
+            return False
+    except Exception as err:
```
kenodegard, line 34: targeted exceptions are better since they wont accidentally catch or hide an unexpected error:

```suggestion
    except (IsADirectoryError, FileNotFoundError, PermissionError) as err:
        # IsADirectoryError: path isn't a file
        # FileNotFoundError: path doesn't exist
        # PermissionError: path cannot be read
```

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -83,6 +85,38 @@ def env_altered_files(env_ok: tuple[Path, str, str, str]) -> tuple[Path, str, st
...
 
 
+def test_listed_on_envs_txt_file(
+    tmp_path: Path, mocker: MockerFixture, env_ok: tuple[Path, str, str, str]
+):
+    """Test that runs for the case when the env is listed on the environments.txt file"""
+    prefix, _, _, _ = env_ok
+    tmp_envs_txt_file = tmp_path / "envs.txt"
+    with open(tmp_envs_txt_file, "w") as f:
+        f.write(f"{prefix}")
```
kenodegard, line 95: `pathlib.Path`'s `write_text` would be good to use here

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -83,6 +85,38 @@ def env_altered_files(env_ok: tuple[Path, str, str, str]) -> tuple[Path, str, st
...
+
+
+def test_not_listed_on_envs_txt_file(
+    tmp_path: Path, mocker: MockerFixture, env_ok: tuple[Path, str, str, str]
+):
+    """Test that runs for the case when the env is not listed on the environments.txt file"""
+    prefix, _, _, _ = env_ok
+    tmp_envs_txt_file = tmp_path / "envs.txt"
+    with open(tmp_envs_txt_file, "w") as f:
+        f.write("Not environment name")
```
kenodegard, line 111: ditto

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -83,6 +85,38 @@ def env_altered_files(env_ok: tuple[Path, str, str, str]) -> tuple[Path, str, st
...
+    prefix, _, _, _ = env_ok
+    tmp_envs_txt_file = tmp_path / "envs.txt"
+    with open(tmp_envs_txt_file, "w") as f:
+        f.write(f"{prefix}")
+
+    mocker.patch(
+        "conda.plugins.subcommands.doctor.health_checks.get_user_environments_txt_file",
+        return_value=tmp_envs_txt_file,
+    )
+    assert check_envs_txt_file(prefix) == True
```
kenodegard, line 101:
```suggestion
    assert check_envs_txt_file(prefix)
```

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -83,6 +85,38 @@ def env_altered_files(env_ok: tuple[Path, str, str, str]) -> tuple[Path, str, st
...
+    prefix, _, _, _ = env_ok
+    tmp_envs_txt_file = tmp_path / "envs.txt"
+    with open(tmp_envs_txt_file, "w") as f:
+        f.write("Not environment name")
+
+    mocker.patch(
+        "conda.plugins.subcommands.doctor.health_checks.get_user_environments_txt_file",
+        return_value=tmp_envs_txt_file,
+    )
+    assert check_envs_txt_file(prefix) == False
```
kenodegard, line 117:
```suggestion
    assert not check_envs_txt_file(prefix)
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -7,19 +7,37 @@
...
     """Displays our report heading."""
     print(f"Environment Health Report for: {Path(prefix)}\n")
 
 
+def check_envs_txt_file(prefix: str | Path) -> bool:
+    """Checks whether the environment is listed in the environments.txt file"""
+    try:
+        envs_txt_file = Path(get_user_environments_txt_file())
+        content = envs_txt_file.read_text()
+        if str(prefix) in content:
```
kenodegard, line 30: what would happen here if `/path/to/dir/nested/env` is in `environment.txt` but we are checking if `/path/to/dir` is a registered environment?

kenodegard (changes requested):

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -7,19 +7,40 @@
...
+    try:
+        envs_txt_file = Path(get_user_environments_txt_file())
+        with envs_txt_file.open() as f:
+            lines = f.readlines()
+            for line in lines:
+                line = line.strip()
+                if Path(prefix).samefile(line):
+                    return True
+            return False
+
```
kenodegard, line 36: a few suggestions here:
1. only generate the prefix `Path` object once (instead of once for every entry in `environments.txt`)
2. move commands that will not throw an exception outside of the try-except (keeps the body of the try-except as concise as possible)
3. remove unnecessary helper variables

```suggestion
    prefix = Path(prefix)
    envs_txt_file = Path(get_user_environments_txt_file())

    try:
        with envs_txt_file.open() as fh:
            for line in fh.readlines():
                if prefix.samefile(line.strip()):
                    return True
            return False
```

kenodegard (dismissed):

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: Got it!
Thanks for explaining, @kenodegard !

beeankha (approved):

jezdez (approved):

PR #13014, opened 2023-08-22
remove cgi import

### LINKED ISSUE #13013: Remove cgi module from   conda/conda/gateways/connection/adapters/ftp.py
### Checklist

### What happened?

`conda/conda/gateways/connection/adapters/ftp.py:19: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13`

Use the Python 3.13+ equivalent.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

dholth:
### Description

Fix #13013

I can't imagine that conda ever needed to STOR files to ftp? It certainly doesn't show up in the tests.

What's the best way to deprecate this?

We could import cgi in the naughty functions, deprecate those and avoid the annoying import-time cgi warning.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/adapters/ftp.py
+++ b/conda/gateways/connection/adapters/ftp.py
@@ -17,10 +17,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 import ftplib
+import os
 from base64 import b64decode
-from email.message import EmailMessage
 from io import BytesIO, StringIO
 from logging import getLogger
+
+from conda.deprecations import deprecated
 
 from ....common.url import urlparse
 from ....exceptions import AuthenticationError
@@ -55,6 +57,7 @@ class FTPAdapter(BaseAdapter):
         self.func_table = {
             "LIST": self.list,
             "RETR": self.retr,
+            "STOR": self.stor,
             "NLST": self.nlst,
             "GET": self.retr,
         }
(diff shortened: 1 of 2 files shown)


dholth:
This removes the "you imported cgi!" warning from the `conda-content-trust` test run.

conda/gateways/connection/adapters/ftp.py
```diff
@@ -266,10 +268,13 @@ def build_response(request, data, code, encoding):
...
 
+@deprecated("24.3", "24.9")
 def parse_multipart_files(request):
     """Given a prepared request, return a file-like object containing the
     original data. This is pretty hacky.
     """
+    import cgi
+
     # Start by grabbing the pdict.
     _, pdict = cgi.parse_header(request.headers["Content-Type"])
```
jezdez, line 279: Let's not deprecate part of the ftp adapters that way (especially since we'll have a way to remove it outright when #13054 merges) and instead switch to the email standard library as described in https://peps.python.org/pep-0594/#cgi

For the header parsing:

```suggestion
from email.message import Message

def parse_separated_header(content_type: str):
    # Adapted from https://peps.python.org/pep-0594/#cgi
    message = Message()
    message['Content-Type'] = content_type
    return dict(message.get_params())


def parse_multipart_files(request):
    """Given a prepared request, return a file-like object containing the
    original data. This is pretty hacky.
    """
    # Start by grabbing the pdict.
    pdict = parse_separated_header(request.headers["Content-Type"])
```

The multipart parsing is a little harder I think, I haven't looked into it though

jezdez (changes requested):

dholth:
I originally reimplemented the functions with email. See earlier commits in
this pr.

Then realized conda has never and will never need to upload files using ftp
and arrived at the simpler and not needed debugging method of moving the
import into the function. This fixes the main problem of seeing CGI
depreciation messages during unrelated test runs.

jezdez:
Ah I didn't see that, interesting, I think my question is if there are some kind of Anaconda internal uses of this code given that it exists, my spidey senses are tingling that this is too specific code to be there for no reason. Did you by any chance do some background check when it was added etc?

dholth:
@jezdez https://github.com/conda/conda/commit/4370e42d32#diff-83b351e0e586cfa26949fd97d2edcd43edbdb0023222acf58ad56cf560870011R3 or the comment at the top of the current file explains it. This code was written for [`requests-ftp`](https://pypi.org/project/requests-ftp/#description). It happened to have extra features that conda doesn't need.

jezdez (approved):

PR #13024, opened 2023-08-23
Add new `tiny_package` instead of `curl` for activation stacking tests

kenodegard:
### Description

`curl`'s `openssl` dependency was recently updated to include activation scripts which broke our tests.

To resolve this we stop depending on `curl` and instead use a dummy package with a tiny executable instead.

### Checklist - did you ...


JeanChristopheMorinPerso:
@kenodegard More accurately, `openssl` recently added activation scripts on Windows, not curl. See https://github.com/AnacondaRecipes/openssl-feedstock/pull/21.

kenodegard:
oops, thanks for the correction

jezdez:
Fixed via #13025.

PR #13031, opened 2023-08-25
Fix pins in `test_reorder_channel_priority`

jaimergp:
### Description

Comes from https://github.com/conda/conda/commit/14d611af65cfa17fd9a4945abe3296d382eaa48e#diff-68cc63c694632f5b92eb4a9942f1cad884cb3b4e75a645ba71d49c249f3859e7R19

This test fails when run with conda-libmamba-solver. I honestly do not understand why it works with conda classic. A pin should constrain what's installable, it doesn't act as a "lock this package to whatever we have installed", afaik. If that's the case, the [documentation](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#preventing-packages-from-updating-pinning) is wrong, but there are mentions in the [changelog](https://github.com/conda/conda/blob/95e90f63516fa95ac013c2ea6919beafb8779067/CHANGELOG.md?plain=1#L2625) where it's implied otherwise.

### Checklist - did you ...

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1393,9 +1393,10 @@ def get_pinned_specs(prefix):
         match_spec = MatchSpec(spec, optional=True)
         pins.append(match_spec)
         if match_spec.name and not any(
-            v for k, v in match_spec._match_components.items()
+            v
+            for k, v in match_spec._match_components.items()
             if k not in ("name", "target", "optional")
-        ): # spec is a name-only spec
+        ):  # spec is a name-only spec
             if not prefix_data:
                 prefix_data = PrefixData(prefix)
             msg = (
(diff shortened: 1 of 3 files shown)


jezdez:
@jaimergp Did we get more eyes on people with historic knowledge about the classic solver?

jaimergp:
> @jaimergp Did we get more eyes on people with historic knowledge about the classic solver?

@jezdez - Planning to bring this up in the next conda community call. That said, this (standard or not) expected behaviour should be now implementable via https://github.com/conda/conda-libmamba-solver/pull/270, so this is not a priority I think.

jaimergp:
Fixed in https://github.com/conda/conda-libmamba-solver/pull/270 via patched test, but for other reasons. The pin works as in the original test now, so no need for this PR.

Issue #12993, opened 2023-08-10
Conda hangs indefinitely Collecting package metadata


manueldeljesus:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Any conda install or update command hangs indefinitely at _Collecting package metadata_. I would expect the command to show the list of packages to update or to install.

### Conda Info

```shell
active environment : base
    active env location : /home/manuel/miniconda3
            shell level : 1
       user config file : /home/manuel/.condarc
 populated config files : /home/manuel/.condarc
          conda version : 23.7.2
    conda-build version : not installed
         python version : 3.9.16.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.36=0
                          __linux=6.4.8=0
                          __unix=0=0
       base environment : /home/manuel/miniconda3  (writable)
      conda av data dir : /home/manuel/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/manuel/miniconda3/pkgs
                          /home/manuel/.conda/pkgs
       envs directories : /home/manuel/miniconda3/envs
                          /home/manuel/.conda/envs
               platform : linux-64
             user-agent : conda/23.7.2 requests/2.31.0 CPython/3.9.16 Linux/6.4.8-100.fc37.x86_64 fedora/37 glibc/2.36
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> /home/manuel/.condarc <==
channel_priority: strict
channels:
  - conda-forge
  - defaults

==> cmd_line <==
debug: False
json: False
```

### Conda list

```shell
# packages in environment at /home/manuel/miniconda3:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
boltons                   23.0.0             pyhd8ed1ab_0    conda-forge
brotli-python             1.0.9            py39h5a03fae_9    conda-forge
bzip2                     1.0.8                h7f98852_4    conda-forge
[...]

PR #13050, opened 2023-08-29
configuring retry behavior to ignore "retry-after" headers

travishathaway:
### Description

Fixes: #12993 

This pull request configures conda to ignore the "Retry-After" header sometimes returned by requests returning a HTTP 429 error code (rate limited). Sometimes these values can be set to very large numbers (>30 seconds) and will cause conda to hang for what feels like an indefinite period.

We already provide other configuration options to tweak the behavior of retries (e.g. `remote_max_retries` and `remote_backoff_factor`), so I think it is okay for us to disregard this header.

I would be happy to discuss this change here though if anyone disagrees.

### Checklist - did you ...


jezdez:
Have we confirmed with Anaconda IT that the header isn't legitimately being set this high via a server side backoff counter?

travishathaway:
@jezdez, I checked on this, and it appears that this backoff counter is being set to 24 hours, which is very high for any client program.

Beyond what Anaconda specifically does, I think it makes sense for us to disregard this considering improperly configured servers can effectively break conda and cause it to hang indefinitely with no meaningful feedback to our users.

Ideally, it would be nice if urllib3 exposed a `retry_after_header_max` setting we could set to conditionally ignore values which would a bad user experience (i.e. > 30 seconds).

This could also be solved by configuring the server correctly. But, I believe Including this in conda would ensure that this behavior wouldn't flair up again if a server were to be configured incorrectly in the future.

jezdez:
> @jezdez, I checked on this, and it appears that this backoff counter is being set to 24 hours, which is very high for any client program.
> 
> Beyond what Anaconda specifically does, I think it makes sense for us to disregard this considering improperly configured servers can effectively break conda and cause it to hang indefinitely with no meaningful feedback to our users.
> 
> Ideally, it would be nice if urllib3 exposed a `retry_after_header_max` setting we could set to conditionally ignore values which would a bad user experience (i.e. > 30 seconds).
> 
> This could also be solved by configuring the server correctly. But, I believe Including this in conda would ensure that this behavior wouldn't flair up again if a server were to be configured incorrectly in the future.

This makes sense to me, thanks for elaborating!

jezdez (approved):
If we can, let's try testing this somehow, we might be able to find inspiration in the requests tests. Other than that, this is good

travishathaway:
I will attempt to come up with a testing scenario for this.

news/13050-ignore-retry-after-header
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Configure conda to ignore "Retry-After" header to avoid the scenarios when this value is very large and causes conda to hang indefinitely.
```
beeankha, line 7:
```suggestion
* Configure conda to ignore "Retry-After" header to avoid the scenarios when this value is very large and causes conda to hang indefinitely. (#13050)
```

beeankha (approved):
Just one small change request for news file

beeankha (approved):

travishathaway:
I thought about the testing scenario some more today and have decided against writing a test. The reason being is that we would essentially be testing to see if urllib3 works as expected. This is testing that should be instead handled by urllib3.

Plus, I think it would be pretty complicated to set up without much benefit.

Issue #11154, opened 2022-01-18
Significant performance impact when enabling `--enable-signature-verification` with conda token


CheeseMochi:
### What happened?
Enabling signature verification via conda/conda-token/conda-content-trust introduces a noticeable increase in the time it takes to perform a conda install. 

To test this out, I timed conda installs of a few packages (scikit-learn, jupyter, and a few others) on clean environment with and without "--enable-signature-verification" and saw real times ~2x and user times >2x with the feature enabled. Note: I performed this with packages that have already been downloaded to remove the network as a factor.

I expect some additional overhead for signature verification, but not quite the 2x observed.

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : tokentest
    active env location : /opt/anaconda3/envs/tokentest
            shell level : 2
       user config file : /Users/nunya/.condarc
 populated config files : /Users/nunya/.condarc
                          /opt/anaconda3/envs/tokentest/.condarc
          conda version : 4.11.0
    conda-build version : 3.21.7
         python version : 3.9.7.final.0
       virtual packages : __osx=10.16=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /opt/anaconda3  (writable)
      conda av data dir : /opt/anaconda3/etc/conda
  conda av metadata url : https://repo.anaconda.cloud/repo
           channel URLs : https://repo.anaconda.cloud/repo/main/osx-64
                          https://repo.anaconda.cloud/repo/main/noarch
                          https://repo.anaconda.cloud/repo/r/osx-64
                          https://repo.anaconda.cloud/repo/r/noarch
                          https://repo.anaconda.cloud/repo/msys2/osx-64
                          https://repo.anaconda.cloud/repo/msys2/noarch
          package cache : /opt/anaconda3/pkgs
                          /Users/nunya/.conda/pkgs
       envs directories : /opt/anaconda3/envs
                          /Users/nunya/.conda/envs
               platform : osx-64
             user-agent : conda/4.11.0 requests/2.26.0 CPython/3.9.7 Darwin/21.1.0 OSX/10.16
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```

</details>

<details>
<summary><code>conda config</code></summary>

```shell
==> /Users/nunya/.condarc <==
default_python: None
extra_safety_checks: False
add_anaconda_token: True
channels:
  - defaults
default_channels:
  - https://repo.anaconda.cloud/repo/main
[...]

PR #13053, opened 2023-09-01
Implement post-solve hook for signature-verification

kenodegard:
### Description

This implements a post-solve hook (proposed in https://github.com/conda/conda/issues/12760) for use by the signature-verification process.

This builds on the work done in https://github.com/conda/conda/pull/13016 but for clarity this is currently targeting the `main` branch so we can see the full diff of changes.

I have tested this with:
- classic solver
- libmamba solver
- channels with signatures (e.g. https://repo.anaconda.cloud/repo)
- channels without signatures (e.g. conda-forge)

Resolves https://github.com/conda/conda/issues/11154

### Checklist - did you ...


conda/core/solve.py
```diff
@@ -135,29 +138,34 @@ def solve_for_transaction(
...
+            force_remove,
+            force_reinstall,
+            should_retry_solve,
+        )
+        # TODO: Only explicitly requested remove and update specs are being included in
+        #   History right now. Do we need to include other categories from the solve?
+
+        # plugins run any post-solve processes here before performing the transaction
+        for post_solve in context.plugin_manager.get_hook_results("post_solves"):
+            post_solve.action(self._repodata_fn, unlink_precs, link_precs)
```
jezdez, line 156: Let's move this into the plugin manager class as `invoke_post_solves` so we can reuse it elsewhere if we need.

```suggestion
        # plugins run any post-solve processes here before performing the transaction
        context.plugin_manager.invoke_post_solves(self._repodata_fn, unlink_precs, link_precs)
```

Untested method:

```python
    def invoke_post_solves(
            self,
            repodata_fn: str,
            unlink_precs: tuple[tuple[PackageRecord, ...]],
            link_precs: tuple[tuple[PackageRecord, ...]],
        ) -> None:
        """
        Invokes ``CondaPostSolve.action`` functions registered with ``conda_post_solves``.

        :param repodata_fn: name of the command that is currently being invoked
        :param unlink_precs: packages records to remove from an environment
        :param link_precs: packages records to add to an environment
        """
        for hook in self.get_hook_results("post_solves"):
            hook.action(repodata_fn, unlink_precs, link_precs)

```

conda/core/solve.py
```diff
@@ -1394,8 +1402,11 @@ def get_pinned_specs(prefix):
 
 
 def diff_for_unlink_link_precs(
-    prefix, final_precs, specs_to_add=(), force_reinstall=NULL
-):
+    prefix,
+    final_precs,
+    specs_to_add=(),
+    force_reinstall=NULL,
+) -> tuple[tuple[PackageRecord, ...], tuple[PackageRecord, ...]]:
```
jezdez, line 1409: Can we define this type at the top of the file somehow as `DiffResult` or something?

conda/models/records.py
```diff
@@ -418,6 +423,11 @@ def record_id(self):
             self.channel.name, self.subdir, self.name, self.version, self.build
         )
 
+    @property
+    @lru_cache(maxsize=None)
+    def metadata(self) -> Metadata:
+        return Metadata()
```
jezdez, line 429: We shouldn't wrap methods with `lru_cache` since the garbage collector can't clean it up (https://rednafi.com/python/lru_cache_on_methods/) or am I missing a reason here for this somehow?

```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.metadata = lru_cache(maxsize=None)(self._metadata)

    def _metadata(self) -> Metadata:
        return Metadata()
```

jezdez: Or would this suffice?

```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.metadata = Metadata()
```

conda/plugins/hookspec.py
```diff
@@ -208,3 +209,31 @@ def conda_auth_handlers():
                     auth_handler=EnvironmentHeaderAuth,
                 )
         """
+
+    @_hookspec
+    def conda_post_solves(self) -> Iterable[CondaPostSolve]:
+        """
+        Register post-solve functions in conda.
```
jezdez, line 216:
```suggestion
        Register post-solve functions in conda that are used in the
        general solver API, after the solver has provided the package
        records to add or remove from the conda environment.
```

conda/plugins/hookspec.py
```diff
@@ -208,3 +209,31 @@ def conda_auth_handlers():
...
+
+           from conda import plugins
+
+
+           def example_post_solve(
+               repodata_fn: str,
+               unlink_precs: tuple[PackageRecord, ...],
+               link_precs: tuple[PackageRecord, ...],
+           ):
+               print("post-solve action")
```
jezdez, line 230: Hmm, let's add a realistic example here, this is going to show up in the docs after all. Maybe printing the to-be-added and to-be-removed packages?

conda/plugins/post_solves.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Register the signature verification plugin."""
```
jezdez, line 3:
```suggestion
"""Register the built-in post_solves hook implementations."""
```

conda/plugins/post_solves.py
jezdez, line 1: It might be beneficial to follow the same pattern as the other built-ins and move this into a `conda/plugins/post_solves/trust.py` module if we want to expand on it, adding a `conda/plugins/post_solves/__init__.py` that has a `plugins` iterable.

news/11545-move-content-trust-into-submodule
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Rework signature verification from pre-solve to post-solve process. (#11545)
```
jezdez, line 3: This is worth mentioning since it's not just an implementation detail.

```suggestion
* Rework signature verification from pre-solve to post-solve process. (#11545)

* Add a new `post-solves` plugin hook. (#11545)
```

jezdez (changes requested):
I think this looks pretty good and needs the usual docs updates to render it in the developer section. I also think we have a few missing areas with test coverage around this that we should figure out how to test.

Generally speaking, my spidey-senses are tingling that it might not take much to also add a `pre_solves` hook at the same time, even if we don't have a use case for a built-in right away.

conda/plugins/post_solves.py
```diff
@@ -0,0 +1,14 @@
...
+from . import CondaPostSolve, hookimpl
+
+
+@hookimpl
+def conda_post_solves():
```
dholth, line 8: How many solves?

conda/trust/signature_verification.py
```diff
@@ -227,23 +232,58 @@ def _fetch_channel_signing_data(
                 f"Invalid JSON returned from {signing_data_url}/{filename}"
             )
 
-    def __call__(self, info, fn, signatures):
-        if not self.enabled or fn not in signatures:
+    def verify(self, repodata_fn: str, record: PackageRecord):
+        repodata, _ = SubdirData(
+            record.channel,
+            repodata_fn=repodata_fn,
+        ).repo_fetch.fetch_latest_parsed()
```
dholth, line 239: I added this API. Is this the correct way to hold it, or should conda interact with the higher-level SubdirData? Can we add an "always use the cache/don't download newer repodata than was sent to the solver" argument?

conda/trust/signature_verification.py
```diff
@@ -227,23 +232,58 @@ def _fetch_channel_signing_data(
...
+        # short-circuit if no signature is defined for this package
+        if record.fn not in signatures:
+            record.metadata.add(f"(no signatures found for {repodata.fn})")
             return
+        signature = signatures[record.fn]
+
+        # extract metadata to be verified
+        if record.fn.endswith(CONDA_PACKAGE_EXTENSION_V1):
+            info = repodata["packages"][record.fn]
+        elif record.fn.endswith(CONDA_PACKAGE_EXTENSION_V2):
```
dholth, line 258: This system of endswith/repodata[key] is unfortunate.

news/11545-move-content-trust-into-submodule
dholth: Verify signatures on to-be-installed packages instead of on all packages.

conda/plugins/hookspec.py
dholth: in conda that are called after the solver has [solved]

conda/models/records.py
```diff
@@ -244,6 +244,11 @@ class PathsData(Entity):
     paths = ListField(PathData)
 
 
+class Metadata(set):
+    def __str__(self) -> str:
+        return " ".join(self)
```
dholth, line 249: I never metadata I didn't like

Is this better than set's builtin __str__?

conda/plugins/post_solves.py
jezdez: What do you mean?

news/11545-move-content-trust-into-submodule
jezdez: Updated my suggestion

conda/plugins/post_solves.py
dholth: I would name the function `conda_post_solve` without the plural.

kenodegard: I agree it doesn't sound right however we are returning an iterable of plugins and all of the other hooks we have are also plural

conda/trust/signature_verification.py
kenodegard: agreed, open to suggestions

conda/plugins/post_solves.py
jezdez: Yeah, it's a pattern in pluggy to use plural when the hookimpl is able to return iterables of plugins

conda/plugins/hookspec.py
kenodegard: see https://github.com/conda/conda/pull/13053/commits/85462f9e2b67b8028cc08c3955857736c33fae93

conda/core/solve.py
kenodegard: `DiffResult` seemed to have limited usability so I opted for `tuple[PackageRecordSequence, PackageRecordSequence]` instead, see https://github.com/conda/conda/pull/13053/commits/85462f9e2b67b8028cc08c3955857736c33fae93

conda/core/solve.py
kenodegard: see https://github.com/conda/conda/pull/13053/commits/3c0f8b7a3de60bacf88580e25aa0da6e078f8834

conda/plugins/post_solves.py
kenodegard: see https://github.com/conda/conda/pull/13053/commits/74e02799af2a7808ea370ef479632ee28d35ee45

conda/trust/signature_verification.py
kenodegard: This was inspired by what conda-libmamba-solver does [here](https://github.com/conda/conda-libmamba-solver/blob/1443a464480dff908e96f3153d2521a429929637/conda_libmamba_solver/index.py#L232)

happy to create a new helper method in `SubdirData` for this

cc @jaimergp

conda/models/records.py
kenodegard: see https://github.com/conda/conda/pull/13053/commits/ce45724620ab7f09605e909e271b8e2ec1294242

conda/models/records.py
kenodegard: see https://github.com/conda/conda/pull/13053/commits/ce45724620ab7f09605e909e271b8e2ec1294242

dholth:
I think it is likely that a future equivalent of conda-content-trust would need to

- sign [a secure digest of] repodata.json as a whole, establishing trust for the set of packages and their metadata
- sign individual packages by fetching additional signature files and performing an online verification

In that system we would not have to pay attention to preserving literal repodata.json fragments but we might need to figure out how to do a decent amount of additional network access.

kenodegard:
pre-commit.ci autofix

conda/plugins/post_solves/signature_verification.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@hookimpl
+def conda_post_solves():
+    from ...trust.signature_verification import signature_verification
+
+    yield CondaPostSolve(
+        name="signature-verification",
+        action=signature_verification,
+    )
```
jezdez, line 14: So, maybe crazy idea, should this move into conda-content-trust? (including the trust module)

jezdez (commented):
I'm wondering if we should split the PR into just the solve plugin hooks and move the majority of the content trust code into cct?

conda/plugins/post_solves/signature_verification.py
jezdez: We've discussed this, and want to merge this first, then refactor.

jezdez (approved):

jezdez:
Ken and I discussed moving it into CCT but want to first merge it and then iterate.

PR #13062, opened 2023-09-05
Catch `PermissionErrors` raised by `conda.cli.find_commands.find_commands`

### LINKED ISSUE #13058: conda fails if PATH contains unreadable directories
### What happened?

The most recent conda patch added a feature where conda will manually scan all of the directories in PATH variable for files beginning with `conda-`. This was added in https://github.com/conda/conda/pull/13006. This method does not account for the possibility that there may be unreadable or non-existent directories in the PATH like shells do. When any conda process is run with this condition, it will exit with 

```[Errno 13] Permission denied: '/path/to/unreadable/directory/in/PATH'```

or similar.

This module should either handle such errors gracefully by skipping that directory or checking whether the directory exists and is readable before calling `os.listdir()`. Another tool could be used that collects all executables in the system PATH, but despite the commit title, the current version of this module doesn't appear to check if the file is executable, so perhaps that is not desired.

This occurs with in all cases with 23.7.3 on Linux systems with a PATH variable containing an unreadable directory.

### PR DESCRIPTION
**Pull request by kenodegard**
### Description

While we check whether a path is a directory we don't also check whether the user has read access resulting in `PermissionError` if any paths are restricted.

Resolves #13058

### Checklist - did you ...

### DIFF
--- a/tests/cli/test_find_commands.py
+++ b/tests/cli/test_find_commands.py
@@ -1,0 +1,71 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+import os
+from pathlib import Path
+
+import pytest
+from pytest import MonkeyPatch
+
+from conda.cli.find_commands import find_commands, find_executable
+from conda.common.compat import on_win
+
+
+@pytest.fixture
+def faux_path(tmp_path: Path, monkeypatch: MonkeyPatch) -> Path:
+    if not on_win:
+        # make a read-only location, none of these should show up in the tests
+        permission = tmp_path / "permission"
+        permission.mkdir(mode=0o333, exist_ok=True)
+        (permission / "conda-permission").touch()
+        (permission / "conda-permission.bat").touch()
+        (permission / "conda-permission.exe").touch()
+        monkeypatch.setenv("PATH", str(permission), prepend=os.pathsep)
+
+    # bad executables
+    bad = tmp_path / "bad"
+    bad.mkdir(exist_ok=True)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/cli/test_find_commands.py [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+
+import pytest
+from pytest import MonkeyPatch
+
+from conda.cli.find_commands import find_commands, find_executable
+from conda.common.compat import on_win
+
+
+@pytest.fixture
+def faux_path(tmp_path: Path, monkeypatch: MonkeyPatch) -> Path:
```
jezdez, line 14: I see what you did there! :D

jezdez (dismissed):

tests/gateways/test_connection.py
```diff
@@ -221,9 +221,15 @@ def test_get_session_with_channel_settings_no_handler(mocker):
...
-    mock_context.channel_settings = ({"channel": "defaults", "auth": "dummy_two"},)
+    mock = mocker.patch(
+        "conda.plugins.manager.CondaPluginManager.get_auth_handler",
+        return_value=None,
+    )
+    mocker.patch(
+        "conda.base.context.Context.channel_settings",
+        new_callable=mocker.PropertyMock,
+        return_value=({"channel": "defaults", "auth": "dummy_two"},),
+    )
```
kenodegard, line 232: With the addition of new tests above the testing order was shuffled and exposed incorrect context mocking in this unrelated test, see https://github.com/conda/conda/actions/runs/6079217031/job/16498615642#step:8:909

kenodegard:
superseded by #13086

PR #13075, opened 2023-09-08
Use OS Truststore in corporate HTTPS inspection proxies

### LINKED ISSUE #13097: Support operating system root certificate stores for TLS connections
### Checklist

### What is the idea?

This would enable us to work around issues that show up often in corporate environments, e.g. #11563.

### Why is this needed?

Increasing the likelihood that connections to repos are secure and pass through corporate proxies, for example.

### What should happen?

We should adopt the [truststore](https://truststore.readthedocs.io/) package as a core dependency.

### Additional Context

_No response_

zeehio:
### Description

Closes #13097.

On (typically corporate) environments, we sometimes find that all traffic is routed through a MITM HTTPS server, that decrypts, inspects and re-encrypts all https traffic. This MITM inspector generates custom SSL certificates, and the self-signed Certificate Authority that signs those custom certificates is trusted in all company laptops. Typically it is installed in the OS trust store.

On such environments, conda can't validate properly the certificates, because it uses `requests`. The [instructions](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#ssl-verification-ssl-verify) make it very easy to disable SSL validation, which is in general considered a bad practice. It is mentioned that a custom certificate file can be provided as well, but that's harder to configure, especially if the user does not know how to get that certificate in the first place, or in what format should it be.

Issue #11563 is one example of this problem.

A simpler approach is to use the trust store. This approach has been submitted at:
- https://github.com/conda/conda/pull/12361

However that PR has been lingering for months with comments that I believe haven't been addressed.

This PR is an alternative implementation, inspired on that PR (thanks @dholth!), that:

- Does not use any internal pip function
- Hides all changes behind an environment variable `CONDA_USE_TRUSTSTORE=True`.

The implementation is rather simple, and it worked on my corporate environment. It is hidden behind an environment variable `CONDA_USE_TRUSTSTORE=True` so it can be released and get wider testing without risks. I guess it could be exposed as an option if necessary, but since I lack the context of the conda project, I would appreciate some pointers on what would be the best option name `--truststore` and where it should be implemented.

### Checklist - did you ...

I would appreciate your feedback before doing further work such as a news entry and documentation changes.

Thanks


dholth:
This is so easy!

One way to package this would be as a https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/pre_commands.html conda pre-command plugin. Instead of defining any real pre-command actions, such a plugin could depend on conda >= 23.7, truststore, and python >= 3.10, be distributed as a tiny conda package, and unconditionally add truststore's monkeypatch when conda searches for its plugins.

zeehio:
> This is so easy!

Thanks!

> 
> One way to package this would be as a https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/pre_commands.html conda pre-command plugin. Instead of defining any real pre-command actions, such a plugin could depend on conda >= 23.7, truststore, and python >= 3.10, be distributed as a tiny conda package, and unconditionally add truststore's monkeypatch when conda searches for its plugins.

Let's say we call that plugin something like `conda-plugin-truststore`.

I believe we would have a bootstrapping issue. A user will get errors downloading `conda-plugin-truststore` because the user needs `conda-plugin-truststore` to download it. I'd like to avoid asking anyone to run any command with `ssl_verify: false` or messing with certificates. A manual download from the anaconda website would require downloading the `truststore` package and the plugin, so it is not very convenient in my opinion...

What if instead of an environment variable we accepted [`ssl_verify: truststore`](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#ssl-verification-ssl-verify), then we changed this PR to honor that setting, and we make it so the anaconda installer (and miniconda?) ships `truststore`? Do you think that would be feasible? I guess growing the installers is not something we are looking forward to do, but I do not see any other simple way of avoiding the bootstrapping problem we would have with a plugin here.

zeehio:
Out of the options described above, I believe bundling `truststore` as a dependency (like pip is considering to do) would be the best option to avoid the bootstrapping issue I described above.

Since I had a bit of time and wanted to attempt to bring something mergeable to the table, I attempted the change, including some tests and documentation.

Some additional points:

- The `truststore` package is developed by sethmlarson, the maintainer of urllib3 and requests packages, (someone `conda` already trusts).
- The `truststore` package is small (~20kB).
- While `truststore` only works on python 3.10 and above, it is tested for compatible python syntax with python 3.7 and above. This ensures the package is installable and has a valid syntax on older versions. If used on python < 3.10, it raises an `ImportError` with the message "truststore requires python 3.10 or above", so users easily know that it is not an option for them.

jezdez:
Thanks for the due diligence @zeehio, this is great to see, here are some thoughts to make sure we have a realistic way forward:

- Vendoring the `truststore` library in conda is not the right direction for conda, e.g. pip follows a different strategy since it has a far less used bootstrapping story (`get-pip.py` is not as often used since `ensurepip` is included in CPython).

  The conda ecosystem, on the other hand, has `miniconda`/`miniforge` and, of course, the Anaconda Distribution installer(s) which contains additional dependencies not vendored.

  In the past year or so, **we've also slowly reduced the vendored libraries in conda** to decrease the maintenance surface.

- The way forward here is to make `truststore` a dependency of conda, so it can go through the usual package update process as other critical packages, such as Seth's urllib3. That means **we'll need to package it on defaults and conda-forge** to set the stage. 

  I'll talk to Seth about this first to make sure this is something he's okay with, given the potential added maintenance burden. `truststore` is currently Python >=3.10 only, which is for the conda ecosystem that is supports Python 3.8 and 3.9 fully still (and will likely for a while).

- Users like you in regulated environments will be able to use future installers, that include `truststore` as part of the default dependency set.

- As to how to enable the `truststore` monkeypatch, from my experience **this should happen as early as possible in the startup process** of the conda CLI, so a plugin hook implementation might actually be a little bit too late.

jezdez:
Oh, I forgot, could you please open a new issue describing the feature request?

This PR doesn't currently have an issue attached, and #11563 is too specific. With a dedicated issue, it's much less likely to get lost during planning and allows the conda maintainers to coordinate the work to package `truststore` etc.

Thank you again!

zeehio:
> Thanks for the due diligence @zeehio, this is great to see, here are some thoughts to make sure we have a realistic way forward:
> 
>     * Vendoring the `truststore` library in conda is not the right direction for conda, e.g. pip follows a different strategy since it has a far less used bootstrapping story (`get-pip.py` is not as often used since `ensurepip` is included in CPython).
>       The conda ecosystem, on the other hand, has `miniconda`/`miniforge` and, of course, the Anaconda Distribution installer(s) which contains additional dependencies not vendored.
>       In the past year or so, **we've also slowly reduced the vendored libraries in conda** to decrease the maintenance surface.

Sorry, I didn't use the right verb. I meant including `truststore` in the anaconda and miniconda installers. This pull request already makes `truststore` as a dependency.

> 
>     * The way forward here is to make `truststore` a dependency of conda, so it can go through the usual package update process as other critical packages, such as Seth's urllib3. That means **we'll need to package it on defaults and conda-forge** to set the stage.

It's already available on conda-forge 

https://anaconda.org/conda-forge/truststore

>       I'll talk to Seth about this first to make sure this is something he's okay with, given the potential added maintenance burden. `truststore` is currently Python >=3.10 only, while the conda ecosystem supports Python 3.8 and 3.9 still (and will likely for a while).
> 

Installing `truststore` on python>=3.7 is supported. Enabling truststore works on python 3.10 and up only. I don't expect it to work on earlier versions.

My suggestion with respect to this is to decide what happens when a user attempts to use `ssl_verify: truststore` on python 3.8 and 3.9:

- The user gets an error mentioning that `truststore` is not available and conda stops.
- The user gets a warning, and conda assumes the default certificate store (from requests)

I guess the second option would be preferable. What do you think?

>     * Users like you in regulated environments will be able to use future installers, that include `truststore` as part of the default dependency set.
> 

Great!

>     * As to how to enable the `truststore` monkeypatch, from my experience **this should happen as early as possible in the startup process** of the conda CLI, so a plugin hook implementation might actually be a little bit too late.

It should work with the current approach committed here then

dholth:
https://truststore.readthedocs.io/en/latest/#user-guide

zeehio:
> https://truststore.readthedocs.io/en/latest/#user-guide

Yes, it is installable but it raises an import error.

The impact of this is that the setting `ssl_verify: truststore` under python <3.10 can't be supported and must be interpreted as `ssl_verify: true` or something like that, maybe with a warning.

I plan to change the code to raise such warning, if that's ok. I agree it is not ideal, but I guess we are blocked on  earlier python versions due to `ssl` package or underlying OpenSSL limitations.

So on "conda 24" (or whatever) `ssl_verify: truststore` works if the environment uses python 3.10, otherwise it behaves as `ssl_verify: true`.

I don't think we can go beyond this unfortunately

dholth:
FYI pip doesn't use the monkeypatch; instead, it adjust requests' session object to be able to pass the truststore ssl context all the way down to urllib3. Both approaches have merit.

sethmlarson:
Happy to see this PR, will be following along and let me know if there's anything that needs doing from Truststore's side.

One downside of the monkeypatching method is it's not possible without testing to know whether all connections are /actually/ using Truststore's SSLContext implementation because import order/changes and the test suite would pass whether Truststore is in use or not (you'd only find out once users start reporting issues).

It also has the potential to "leak" into other places where Truststore's SSLContext might not be wanted, but I suspect that that case isn't an issue for Conda.

zeehio:
I see good reasons for switching from monkey patching to just passing the ssl context through the HTTPAdapter, so I am implementing the changes.

I'll finish the changes later today/tomorrow, since right now I am struggling to install the pre-commits locally: I am behind an https certificate inspector, and the pre-commit tool gives me https validation errors.

jezdez:
I concur about the monkeypatch, conda is embedded in some project in which accidentally leaking the SSL context via the monkeypatch wouldn't be great.

@sethmlarson Thanks for your kind message, I'm stoked to see you support it, which might solve a real problem for many packaging users throughout the communities. Thank you for working on it!

jezdez:
@zeehio Speaking about pre-commit, you may simply comment here `pre-commit.ci autofix` and it'll autofix the issue (most of them) via the CI here and commit the changes to the branch, no need to run it locally.

jezdez:
pre-commit.ci autofix

conda/gateways/connection/adapters/http.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2008-2023 The pip developers
+# SPDX-License-Identifier: MIT
+#
+"""Defines HTTP transport adapter for CondaSession (requests.Session).
+
+Taken from pip (https://github.com/pypa/pip/blob/8c24fd2a80bad21aa29aec02fb48bd89a1e8f5e1/src/pip/_internal/network/session.py#L254
```
dholth, line 8: I think we will need to include https://github.com/pypa/pip/blob/8c24fd2a80bad21aa29aec02fb48bd89a1e8f5e1/LICENSE.txt

tests/test_create.py
```diff
@@ -1258,7 +1258,7 @@ def test_conda_config_validate(clear_package_cache: None):
             assert len(exc.value.errors) == 2
             str_exc_value = str(exc.value)
             assert (
-                "must be a boolean, a path to a certificate bundle file, or a path to a directory containing certificates of trusted CAs"
+                "must be a boolean, a path to a certificate bundle file, a path to a directory containing certificates of trusted CAs or 'truststore' to use the OS certificate store."
```
dholth, line 1261:
```suggestion
                "must be a boolean, a path to a certificate bundle file, a path to a directory containing certificates of trusted CAs, or 'truststore' to use the OS certificate store."
```

conda/gateways/connection/session.py
```diff
@@ -156,6 +157,26 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
...
+                import ssl
+
+                import truststore
+
+                ssl_context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+            except ImportError:
+                warnings.warn(
+                    dals(
+                        """
+                        The `ssl_verify: truststore` setting is only supported on python-3.10 or above.
```
dholth, line 172:
```suggestion
                        The `ssl_verify: truststore` setting is only supported on python >=3.10.
```
A conda matchspec?

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
   "setuptools >=60.0.0",
   "toolz >=0.8.1",
   "tqdm >=4",
+  "truststore >=0.8.0",
```
dholth, line 38:
```suggestion
  "truststore >=0.8.0; python_version>='3.10'",
```

jezdez:
I've requested building the `truststore` package internally at Anaconda to be shipped on Anaconda's main channel (referring to the conda-forge feedstock).

jezdez:
I've opened #13097 to track this.

zeehio:
I can squash all commits if you prefer (or maybe they can be squashed at merge time automatically, I don't mind).

I think I took care of all comments/suggestions on this PR. If there is anything left feel free to ping me.

jezdez:
Thanks @zeehio, we squash commits on merge automatically and don't allow merge commits by default.

conda/gateways/connection/adapters/http.py
zeehio: I copied the whole license file in the header

conda/gateways/connection/__init__.py
```diff
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 try:
     from requests import ConnectionError, HTTPError, Session
-    from requests.adapters import BaseAdapter, HTTPAdapter
+    from requests.adapters import DEFAULT_POOLBLOCK, BaseAdapter
+    from requests.adapters import HTTPAdapter as BaseHTTPAdapter
```
jezdez, line 6: This change of import will likely break some 3rd party conda packages, let's keep using the name `HTTPAdapter` here and then import in the new `adapters/http.py` using a new local name.

```suggestion
    from requests.adapters import DEFAULT_POOLBLOCK, BaseAdapter, HTTPAdapter
```

conda/gateways/connection/__init__.py
```diff
@@ -17,7 +18,13 @@
...
 except ImportError:  # pragma: no cover
     from pip._vendor.requests import ConnectionError, HTTPError, Session  # noqa: F401
-    from pip._vendor.requests.adapters import BaseAdapter, HTTPAdapter  # noqa: F401
+    from pip._vendor.requests.adapters import (  # noqa: F401
+        DEFAULT_POOLBLOCK,
+        BaseAdapter,
+    )
+    from pip._vendor.requests.adapters import (  # noqa: F401
+        HTTPAdapter as BaseHTTPAdapter,
+    )
```
jezdez, line 27:
```suggestion
    from pip._vendor.requests.adapters import (  # noqa: F401
        DEFAULT_POOLBLOCK,
        BaseAdapter,
        HTTPAdapter,
    )
```

conda/gateways/connection/adapters/http.py
```diff
@@ -0,0 +1,80 @@
...
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+"""
+from typing import TYPE_CHECKING, Any, Optional
+
+from .. import DEFAULT_POOLBLOCK, BaseHTTPAdapter
```
jezdez, line 37:
```suggestion
from .. import DEFAULT_POOLBLOCK, HTTPAdapter as BaseHTTPAdapter
```

conda/gateways/connection/session.py
```diff
@@ -156,6 +157,26 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
...
+                import ssl
+
+                import truststore
+
+                ssl_context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+            except ImportError:
+                warnings.warn(
+                    dals(
+                        """
+                        The `ssl_verify: truststore` setting is only supported on python >=3.10.
```
jezdez, line 172:
```suggestion
                        The `ssl_verify: truststore` setting is only supported on Python 3.10 or later.
```

conda/gateways/connection/session.py
```diff
@@ -156,6 +157,26 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
...
+                        Falling back to `ssl_verify: true`
```
jezdez, line 173: This is misleading, it doesn't fall back to `ssl_verify = True` but to the default of the parent class' `verify` default value.

conda/gateways/connection/session.py
```diff
@@ -182,8 +203,6 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
 
         self.headers["User-Agent"] = context.user_agent
 
-        self.verify = context.ssl_verify
-
```
jezdez, line 186: Something is wrong here with the `verify` boolean, we stop setting it here but above it's not being set if `truststore` is configured but doesn't work for some reason.

I think the fallback should not be the default of the session parent class but bubbling up as an error to the user since it may lead to surprising results.

E.g. if a user has `truststore` configured explicitly, they should be told the reason in case it doesn't work, since the non-OS certs may not work for their use case after all.

conda/gateways/connection/session.py
```diff
@@ -156,6 +157,26 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
...
 
+        ssl_context = None
+        if context.ssl_verify == "truststore":
+            try:
+                import ssl
+
+                import truststore
+
+                ssl_context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+            except ImportError:
```
jezdez, line 168: Are there other exceptions that may be thrown here other than `ImportError`?

docs/source/user-guide/troubleshooting.rst
```diff
@@ -298,7 +312,13 @@ Make sure your conda is up-to-date: ``conda --version``
...
 If not, run: ``conda update conda``
 
-Temporarily set your ``ssl_verify`` variable to false, upgrade the requests package, and then
+Try using the operating system certificate store. Set you ``ssl_verify`` variable to ``truststore``
+using the following command::
+
+    conda config --set ssl_verify truststore
+
+If using the OS certificate store does not solve your issue, temporarily set your ``ssl_verify`` variable
+to false, upgrade the requests package, and then
```
jezdez, line 321:
```suggestion
If using the operating system certificate store does not solve your issue, temporarily
set your ``ssl_verify`` variable to false, upgrade the requests package, and then
```

news/13075-ssl-verify-truststore
```diff
@@ -0,0 +1,3 @@
+### Enhancements
+
+* `ssl_verify: truststore` lets conda use the operating system certificate store  (requires python >= 3.10) (#13075)
```
jezdez, line 3:
```suggestion
* Add support for `truststore` to the `ssl_verify` config option, enabling conda to use the operating system certificate store (requires Python 3.10 or later). (#13075)
```

conda/base/context.py
```diff
@@ -156,11 +156,12 @@ def default_python_validation(value):
...
     if isinstance(value, str):
-        if not isfile(value) and not isdir(value):
+        if not value == "truststore" and not isfile(value) and not isdir(value):
             return (
                 "ssl_verify value '%s' must be a boolean, a path to a "
-                "certificate bundle file, or a path to a directory containing "
-                "certificates of trusted CAs." % value
+                "certificate bundle file, a path to a directory containing "
+                "certificates of trusted CAs, or 'truststore' to use the OS certificate store."
+                % value
```
jezdez, line 164: Better to spell out what we mean with "OS".

```suggestion
                "certificate bundle file, a path to a directory containing "
                "certificates of trusted CAs, or 'truststore' to use the "
                "operating system certificate store." % value
```

conda/base/context.py
```diff
@@ -1701,8 +1702,8 @@ def description_map(self):
                 browser. By default, SSL verification is enabled, and conda operations will
                 fail if a required url's certificate cannot be verified. Setting ssl_verify to
                 False disables certification verification. The value for ssl_verify can also
-                be (1) a path to a CA bundle file, or (2) a path to a directory containing
-                certificates of trusted CA.
+                be (1) a path to a CA bundle file, (2) a path to a directory containing
+                certificates of trusted CA, or (3) 'truststore' to use the OS certificate store.
```
jezdez, line 1706:
```suggestion
                be (1) a path to a CA bundle file, (2) a path to a directory containing
                certificates of trusted CA, or (3) 'truststore' to use the
                operating system certificate store.
```

conda/gateways/connection/session.py
```diff
@@ -156,6 +157,26 @@ def __init__(self, auth: AuthBase | tuple[str, str] | None = None):
...
+                warnings.warn(
```
jezdez, line 169: See below for my remark about the need to bubble up the error here, if users specifically configure to use `truststore`, we need to raise a `CondaError` or something instead of assuming ssl_verify=True (e.g. they may prefer to configure a custom cert bundle instead).

jezdez (changes requested):
This is excellent, almost done, just a few more tweaks around error messaging and changing the default fallback if something fails while loading `truststore`. We need to bubble up the import error instead of falling back to ssl_verify=True.

docs/source/user-guide/troubleshooting.rst
dholth: This suggestion is odd without a minimum supported version of the requests package?

zeehio: I agree with you it is odd. However it appeared like that on the troubleshooting guide, so I did not want to change it without reason. If I had to guess the reason behind this "update requests" is that the upgrade of the requests package may pull  an upgraded certificate bundle, that may address the connection issue.

conda/gateways/connection/session.py
zeehio: None that I can think of

conda/gateways/connection/session.py
zeehio: I changed it so it now raises a CondaError with a message stating that truststore is only available on python 3.10 or above.

I also made sure `self.verify = True` when `truststore` to ensure requests validates the certificate.

zeehio:
pre-commit.ci autofix

conda/gateways/connection/session.py
zeehio: I raise a `CondaError` now

zeehio:
Thanks, I believe I have addressed the suggestions.

The test failure seems to be related to a random connection issue that can be fixed by rerunning the job.

conda/gateways/connection/session.py
zeehio: To be more explicit about this:

- truststore defines SSLContext here https://github.com/sethmlarson/truststore/blob/459c75006a11265d2dbbfaed5e389cd61d7c93c5/src/truststore/_api.py#L57
- it eventually calls the SSLContext from the `ssl` package defined here https://github.com/python/cpython/blob/17a335dd0291d09e1510157a4ebe02932ec632dd/Lib/ssl.py#L484C3-L484C3

I don't see any specific documented exception or raised exception there. I could add a "catch all" exception here and reraise as a CondaError but I believe that if there is any issue I haven't seen it might be better to let it raise than to just add code out of fear of missing something.

Feel free to check yourself if you have further concerns about this, or tell me what are those concerns and I'll do my best to address them 😃 

(And thanks for the review!)

dholth:
We have some code in test_jlap to create a ssl package server; is it possible to add that certificate to truststore's set to test without external servers?

sethmlarson:
> We have some code in test_jlap to create a ssl package server; is it possible to add that certificate to truststore's set to test without external servers?

@dholth Truststore does this sort of testing [using mkcert](https://github.com/sethmlarson/truststore/blob/main/tests/conftest.py).

dholth:
> > We have some code in test_jlap to create a ssl package server; is it possible to add that certificate to truststore's set to test without external servers?
> 
> @dholth Truststore does this sort of testing [using mkcert](https://github.com/sethmlarson/truststore/blob/main/tests/conftest.py).

Or that could be considered testing truststore and not testing whether conda is using truststore.

jezdez (dismissed):

jezdez:
The process to build the `truststore` package for the Anaconda's defaults channel is ongoing, and I hope this to be done in time for the feature freeze for conda 23.9.0 this week. Let's not merge this PR until then.

jezdez:
https://github.com/AnacondaRecipes/truststore-feedstock/pull/2 specifically

jezdez:
This has been released on the main channel: https://anaconda.org/main/truststore/files

jezdez (approved):

PR #13054, opened 2023-09-01
Add transport adapters plugin hook

jezdez:
### Description

Fixes #13176.

This is a continuation of the work done for the auth-handlers plugin hook in #12911.

It basically allows expanding the session backend system, to replace or add the requests transport adapters used and opens the door to moving the S3 and the FTP adapters into their own packages, or add new ones like OCI.

It might also enable us to additionally support other http libraries that have better support for HTTP/2 etc.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13084, opened 2023-09-11
Use `conda-forge::minio-server`

kenodegard:
### Description

Instead of implementing a manual download/caching for MinIO we can turn it into a conda package.

### Checklist - did you ...


kenodegard:
I don't have anything against eliminating MinIO entirely assuming we still get the the same level of coverage

jezdez:
> s3 is http, and conda isn't doing anything s3-specific. Our existing http(s) package server fixture introduced in the last 6 months serves the necessary package files without requiring a large binary or an upload step.

I'm not following, we obviously do a bunch of S3 related things in the [S3 adapter](https://github.com/conda/conda/blob/4a76fea527a729421d895acfca1abfa508b50683/conda/gateways/connection/adapters/s3.py) that are very S3 specific. The e2e test is supposed to make sure we're doing the right thing and not only monkeypatching boto3's session. That would only prove that we know how to use boto3 and not that conda works with a S3-compatible server API.

dholth:
I'm not seeing that much going on in the very short s3.py. We configure boto3 to work with a non-AWS server exactly as we would for minio. These tests do not monkey-patch boto3 except to pass it a different configuration [similar to the conda-package-streaming tests](https://github.com/conda/conda-package-streaming/blob/main/tests/test_s3.py#L12-L19). The [bucket name becomes the first part of the URL](https://github.com/conda/conda/blob/4a76fea527a729421d895acfca1abfa508b50683/conda/gateways/connection/adapters/s3.py#L75), later the s3 adapter copies boto3 response headers.

jaimergp:
The signature monkeypatching can be undone now, btw. See https://github.com/boto/boto3/issues/1375#issuecomment-1699234689.

jezdez:
Let's just go with this approach and then maybe strip out the S3 adapter out of Conda long-term when the #13054 has been merged.

jezdez (approved):

PR #13086, opened 2023-09-11
23.7.x find commands permission error

kenodegard:
### Description

While we check whether a path is a directory we don't also check whether the user has read access resulting in PermissionError if any paths are restricted.

Resolves https://github.com/conda/conda/issues/13058

Target `23.7.x` instead of `main`, supersedes #13062

### Checklist - did you ...

### DIFF
--- a/conda/cli/find_commands.py
+++ b/conda/cli/find_commands.py
@@ -6,7 +6,7 @@ import re
 import sys
 import sysconfig
 from functools import lru_cache
-from os.path import basename, expanduser, isdir, isfile, join
+from os.path import basename, expanduser, isfile, join
 
 from ..common.compat import on_win
 
@@ -67,14 +67,13 @@ def find_commands(include_others=True):
 
     res = set()
     for dir_path in dir_paths:
-        if not isdir(dir_path):
-            continue
         try:
             for entry in os.scandir(dir_path):
                 m = pat.match(entry.name)
                 if m and entry.is_file():
                     res.add(m.group(1))
-        except PermissionError:
+        except (FileNotFoundError, PermissionError):
+            # FileNotFoundError: directory doesn't exist
             # PermissionError: user doesn't have read access
             continue
     return tuple(sorted(res))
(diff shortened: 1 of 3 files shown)


conda/cli/find_commands.py
```diff
@@ -61,16 +61,20 @@ def find_commands(include_others=True):
...
 
     if on_win:
-        pat = re.compile(r"conda-([\w\-]+)\.(exe|bat)$")
+        pat = re.compile(r"conda-([\w\-]+)(\.(exe|bat))?$")
     else:
         pat = re.compile(r"conda-([\w\-]+)$")
 
     res = set()
     for dir_path in dir_paths:
         if not isdir(dir_path):
```
dholth, line 70: We could even move the try/except up here and catch not a directory error + permission error (or OSError?)

dholth (dismissed):

conda/cli/find_commands.py
kenodegard: see https://github.com/conda/conda/pull/13086/commits/59adb5c6901a722f03667a2e4353e285378b2a71

beeankha (approved):

conda/cli/find_commands.py
```diff
@@ -67,14 +67,13 @@ def find_commands(include_others=True):
...
     for dir_path in dir_paths:
-        if not isdir(dir_path):
-            continue
         try:
             for entry in os.scandir(dir_path):
                 m = pat.match(entry.name)
                 if m and entry.is_file():
                     res.add(m.group(1))
-        except PermissionError:
+        except (FileNotFoundError, PermissionError):
```
dholth, line 75: A call to `os.scandir('conda.py')` produces `NotADirectoryError` which is a subclass of `OsError`.

kenodegard: I just tested this locally:

```
$ python -c "import os; os.scandir('yellow')"
  File "<string>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'yellow'
```

dholth: Add an existing file to PATH.

kenodegard: ugh

PR #13101, opened 2023-09-14
catch JSONDecodeError when loading repodata cache metadata

### LINKED ISSUE #13056: Treat JSONDecodeError on repodata.info.json as a warning, equivalent to a missing repodata.info.json
### Checklist

### What happened?

Parallel downloads can corrupt the repodata metadata. Since we will soon lock by default, this will probably be less of an issue. If `<repodata cache key>.info.json` can't be decoded, show a warning, then behave as if `*.info.json` was missing, download fresh repodata.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

dholth:
### Description

Fix #13056 

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -580,7 +580,7 @@ class RepodataCache:
         try:
             self.load(state_only=True)
         except (FileNotFoundError, json.JSONDecodeError) as e:
-            if isinstance(e,  json.JSONDecodeError):
+            if isinstance(e, json.JSONDecodeError):
                 warnings.warn(f"{e.__class__.__name__} loading {self.cache_path_state}")
             self.state.clear()
         return self.state
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -579,7 +579,9 @@ class RepodataCache:
         """
         try:
             self.load(state_only=True)
-        except FileNotFoundError:  # or JSONDecodeError?
+        except (FileNotFoundError, json.JSONDecodeError) as e:
+            if isinstance(e,  json.JSONDecodeError):
+                warnings.warn(f"{e.__class__.__name__} loading {self.cache_path_state}")
             self.state.clear()
         return self.state
 
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -609,7 +609,7 @@ class RepodataCache:
         adjacent to `self.cache_path_json` to be on the same filesystem.
         """
         with self.cache_path_state.open("a+") as state_file, lock(state_file):
-            # "a+" avoids trunctating file before we have the lock and creates
+            # "a+" creates the file if necessary, does not trunctate file.
             state_file.seek(0)
             state_file.truncate()
             stat = temp_path.stat()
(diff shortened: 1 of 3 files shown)


conda/gateways/repodata/__init__.py
```diff
@@ -579,7 +579,9 @@ def load_state(self):
         """
         try:
             self.load(state_only=True)
```
dholth, line 581: I notice we also load the state in the `load()` method. It happens that we always load the state in this method first, to avoid parsing `repodata.json` twice when it is outdated. Conceivably we would need the same checks in load() however.

conda/gateways/repodata/__init__.py
```diff
@@ -579,7 +579,9 @@ def load_state(self):
...
-        except FileNotFoundError:  # or JSONDecodeError?
+        except (FileNotFoundError, json.JSONDecodeError) as e:
+            if isinstance(e, json.JSONDecodeError):
+                warnings.warn(f"{e.__class__.__name__} loading {self.cache_path_state}")
```
jezdez, line 584: Is a warning the right thing here? Should we use a debug logger?

```suggestion
        except (FileNotFoundError, json.JSONDecodeError) as exc:
            if isinstance(exc, json.JSONDecodeError):
                warnings.warn(f"{exc.__class__.__name__} loading {self.cache_path_state}")
```

tests/gateways/test_jlap.py
```diff
@@ -302,6 +302,55 @@ def test_repodata_state(
...
+
+        # Corrupt the cache state. Double json could happen when (unadvisably)
+        # running conda in parallel, before we added locks-by-default.
+        sd.cache_path_state.write_text(sd.cache_path_state.read_text() * 2)
+
+        # now try to re-download
+        SubdirData.clear_cached_local_channel_data(exclude_file=False)
+        sd2 = SubdirData(channel=test_channel)
+
+        # warnings.warn(f"{e.__class__.__name__} loading {self.cache_path_state}")
```
jezdez, line 349:
```suggestion
        # warnings.warn(f"{exc.__class__.__name__} loading {self.cache_path_state}")
```

jezdez (approved):
Minor nits only

beeankha (approved):

conda/gateways/repodata/__init__.py
jezdez: > `e` is the idiomatic variable name for an exception.

Agree to disagree, descriptive variable names are also idiomatic, make the code more accessible for beginners.

> A logger could be appropriate.

Should the logger be a separate PR or should there be a follow-up ticket?

dholth: The two main choices for this symbol that is usually defined on one line and only used on the following line appear to be `e` and a random 2-5 character string starting with the letter `e`. In conda 136 are `e` out of 199 total `except .* as`.

https://github.com/search?type=code&q=%2Fexcept+.*+as+e%3A%2F+language%3APython

https://github.com/search?q=org%3Aconda+%2Fexcept+.*+as+%5Cw%2B%3A%2F&type=code

jezdez: Thanks for the additional data!

PR #13108, opened 2023-09-15
prefer log.warning() in conda.gateways.repodata

dholth:
### Description

Follow-up on #13056. Is it better to use log.warning() over warnings.warn()?

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -581,7 +581,7 @@ class RepodataCache:
             self.load(state_only=True)
         except (FileNotFoundError, json.JSONDecodeError) as e:
             if isinstance(e, json.JSONDecodeError):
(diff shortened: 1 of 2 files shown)


beeankha (approved):

conda/gateways/repodata/__init__.py
```diff
@@ -581,7 +581,7 @@ def load_state(self):
             self.load(state_only=True)
         except (FileNotFoundError, json.JSONDecodeError) as e:
             if isinstance(e, json.JSONDecodeError):
-                warnings.warn(f"{e.__class__.__name__} loading {self.cache_path_state}")
+                log.warning(f"{e.__class__.__name__} loading {self.cache_path_state}")
```
jezdez, line 584: The main difference is that `warnings.warn` will only show up once for the same parameter, while `log.warning` will emit to the configured log output every occurrence. More infos: https://stackoverflow.com/a/37979724

PR #13111, opened 2023-09-15
reindex test repository, fixing checksums

dholth:
### Description

Some of the test suite repository checksums were incorrect, matching the .tar.bz2 versions of those packages. This makes it impossible to install those `.conda`-format packages in tests. Fixed by running conda-index and preserving the packages that are only in index.json but not on disk used for solver tests.

### DIFF
--- a/tests/data/conda_format_repo/linux-64/current_repodata.json
+++ b/tests/data/conda_format_repo/linux-64/current_repodata.json
@@ -2,21 +2,7 @@
   "info": {
     "subdir": "linux-64"
   },
-  "packages": {
-    "libgcc-ng-8.2.0-hdf63c60_1.tar.bz2": {
-      "build": "hdf63c60_1",
-      "build_number": 1,
-      "depends": [],
-      "license": "GPL",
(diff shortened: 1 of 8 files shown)


jezdez (approved):

kenodegard:
Does this fix mean we can remove this: https://github.com/conda/conda/blob/9a23bc4d6e19b2a913a10e8e382dfb3b97152730/tests/gateways/test_connection.py#L108-L109

dholth:
I do want to remove USE_ONLY_TAR_BZ2. After fixing this index, it was producing another error about not being able to seek that I couldn't reproduce locally - an error that I've seen before for some test packages, generated with buggy test code, that have the same file twice in a `.tar`.

We should make sure conda-package-streaming is able to deal with all test packages. If that works I'm afraid we'll have to hunt for any remote test packages used in the tests. Or, conda's package cache is re-used during tests and getting corrupted by some other test - which could be duplicated by choosing the same test order, by invoking the failing 1/3d of the tests with the same options used in CI.

PR #13128, opened 2023-09-19
Add response string to repodata JSONDecodeError

### LINKED ISSUE #13110: Friendly JSONDecodeError alternative when repodata.json can't be parsed?
### Checklist

### What is the idea?

Users can see a `JSONDecodeError` with basically no information, that goes something like this, when we try to parse the direct http body of a repodata request.

```
fetched https://the-repository/platform/repodata.json
JSONDecodeError <expected json on byte 0>
```

It might be because you got HTML "please log into the cafe wifi" instead of JSON.

Instead, we could print the HTTP response headers "content-type was ..." and we could print the first ~80 characters of the response string.

'Could not decode repodata.json. Got "<html><head><title>I am a Proxy</ti"..., Content-Type: html" instead.' 

### Why is this needed?

Better support users behind a proxy.

Unfortunately JSON is not necessarily decoded until after we discard HTTP headers. IIRC JLAP stores a selection of headers in the `<repodata>.info.json` i.e. the cache object's state property which could be retrieved to fill out the error message.

We should be able to test this with the same techniques used in `test_jlap` or `test_repodata_gateway / test_subdir_data`, by setting up a local package server and putting e.g. a html file named 'repodata.json' in one of our test folders.

dholth:
### Description

Fix #13110

This should make it much easier to figure out what is going on when repodata.json is not parseable.

Do we need to sanitize the response so it can't print HTML or escape characters?

It would be better if we converted this to a CondaError instead of surfacing the Python exception. We could raise a subclass of JSONDecodeError with the data and convert to CondaError at a much higher level?

### Checklist - did you ...


travishathaway:
@dholth,

It think it would be a better idea to address these issues just slightly further up the call stack.

What if instead of returning a JSON string that has be parsed later, [fetch_latest](https://github.com/conda/conda/blob/32647334148ab4d2109656954d602648eb7be95b/conda/gateways/repodata/__init__.py#L761) always returned a parsed value and we forced it to be dealt with there.

We could also change [read_cache](https://github.com/conda/conda/blob/32647334148ab4d2109656954d602648eb7be95b/conda/gateways/repodata/__init__.py#L882) to work the same way too. That way we could also separate the error messages so that it's easier for us differentiate invalid JSON coming from a network request vs. coming from bad cache values on disk.

Also, I would expect the error message to look similar to what we have in `read_cache` right now for problems related to faulty copies of cached data.

In the future, it might also be nice to figure out how to implement some retry logic if a faulty cached version is found (i.e. just remove it and grab another fresh copy from the server).

conda/gateways/repodata/__init__.py
```diff
@@ -715,7 +718,14 @@ def fetch_latest_parsed(self) -> tuple[dict, RepodataState]:
...
         if isinstance(parsed, str):
-            return json.loads(parsed), state
+            try:
+                return json.loads(parsed), state
+            except json.JSONDecodeError as e:
+                e.args = (
+                    f'{e.args[0]}; got "{parsed[:ERROR_SNIPPET_LENGTH]}"',
+                    *e.args[1:],
+                )
+                raise
```
dholth, line 728: Is there a way for me to get the URL here...

jezdez (commented):
Could we create a higher level tool that wraps `json.loads` or something that would enable us to handle this holistically through out the codebase?

dholth:
I'm not sure about a global wrapper for json.loads(). Could be tricky to decide when to keep the input string or figure out where to attach additional useful information such as the URL that produced the unparseable input.

A good followup would be to write a test that adds a bad channel (html-formatted repodata.json) and check that conda still loads any other configured channels.

news/11804-annotate-jsondecodeerror
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Show first few characters of undecodeable response if repodata.json raises
+  JSONDecodeError (#11804)
```
beeankha, line 4:
```suggestion
* Show first few characters of undecodeable response if `repodata.json` raises
  `JSONDecodeError`. (#11804)
```

kenodegard:
pre-commit.ci autofix

kenodegard (approved):

PR #13151, opened 2023-09-26
Log missing or non-string RepodataState items at debug level instead …

dholth:
…of warning

### Description

This can be triggered by the server not returning an Etag header or by another *.info.json-writer (does mamba use the new extension?) for example. Log at debug level.


jezdez (approved):

beeankha (approved):

conda/gateways/repodata/__init__.py
```diff
@@ -468,7 +468,7 @@ def __setitem__(self, key: str, item: Any) -> None:
         if key in self._aliased:
             key = key[1:]  # strip underscore
         if key in self._strings and not isinstance(item, str):
-            log.warn(f'Replaced non-str RepodataState[{key}] with ""')
+            log.debug('Replaced non-str RepodataState[%s] with ""', key)
```
dholth, line 471: This form doesn't perform the substitution unless debug logs are enabled. Not exactly a hot path but why not?

PR #13157, opened 2023-09-27
Only continue tests.yml workflow if not canceled

kenodegard:
### Description

Sometimes we need to cancel workflows and we expect them to quickly finish so we can restart them. Currently this is not possible as several jobs will `always()` run even when prior jobs have been canceled.

Xref https://github.com/conda/conda-build/pull/5018

### Checklist - did you ...


jezdez (approved):
OMG this was so annoying!

PR #13159, opened 2023-09-28
Fix CMD.exe activation test

kenodegard:
### Description

Looks like git moved on the GitHub runner (?) (or perhaps it changed on conda-forge?) so the CMD.exe test started to fail. Fixed the issue by being more careful with how we "reset" the `PATH` for the CMD.exe test.

Merging this to the 23.9.x release branch since we need this fix in order to merge #13155 (we will also need this fix for any 23.9.x patches and the upcoming 23.10.x special release).

### Checklist - did you ...

### DIFF
--- a/tests/test_activate.py
+++ b/tests/test_activate.py
@@ -2455,7 +2455,7 @@ class InteractiveShell:
                 self.expect_exact(value)
                 self.expect(".*\n")
             else:
-                self.expect(f"{value}\n")
+                self.expect(f"{value}\r?\n")
         except:
             print(f"{self.p.before=}", file=sys.stderr)
             print(f"{self.p.after=}", file=sys.stderr)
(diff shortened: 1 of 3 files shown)


tests/test_activate.py
```diff
@@ -2438,34 +2438,37 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
         # value is actually a regex
         self.sendline(self.print_env_var % env_var)
         try:
             if use_exact:
-                self.p.expect_exact(value)
+                self.expect_exact(value)
                 self.expect(".*\n")
             else:
-                self.expect("%s\n" % value)
+                self.expect(f"{value}\r?\n")
```
kenodegard, line 2458: Auto-handle `\r\n` and `\n`.

tests/test_activate.py
```diff
@@ -2940,112 +2942,127 @@ def test_powershell_PATH_management(shell_wrapper_integration: tuple[str, str, s
...
 @pytest.mark.skipif(not which("cmd.exe"), reason="cmd.exe not installed")
 @pytest.mark.integration
-def test_cmd_exe_basic_integration(shell_wrapper_integration: tuple[str, str, str]):
+def test_cmd_exe_basic_integration(
+    shell_wrapper_integration: tuple[str, str, str],
+    monkeypatch: MonkeyPatch,
+):
     prefix, charizard, _ = shell_wrapper_integration
+    conda_bat = Path(CONDA_PACKAGE_ROOT, "shell", "condabin", "conda.bat")
+    monkeypatch.setenv("PATH", STRIPPED_PATH)
```
kenodegard, line 2986: `monkeypatch.setenv` instead of `env_vars`

dholth: Don't we use conda env_vars because it resets the context? Will that be a problem compared to monkeypatch?

kenodegard: I've been trying to move away from `env_var(s)?` in general

In this case resetting the context has no impact since the context doesn't do anything with `PATH`

jezdez (approved):

beeankha (approved):

Issue #12418, opened 2023-03-01
Calling `conda` command on `cmd` exits without return code in CI if not called via `call`


jaimergp:
### What happened?

I observed this in https://github.com/conda/conda-libmamba-solver/pull/133. I was debugging the `conda` installation via `conda info -a` and then doing some more patching, but those operations afterwards were never executed (including the `if errorlevel 1 exit 1` guard!).

Turns out that running `conda` commands on CMD returns immediately (at least on our CI). This might be a `EXIT` command somewhere, without [the `/b` flag](https://ss64.com/nt/exit.html)? It needs to be guarded with a `CALL`.

This returns early (`OK` is never printed!):

```yaml
      - shell: cmd
        run: |
          conda info -a
          if errorlevel 1 exit 1
          echo "OK"
```

This proceeds correctly (`OK` is printed):

```yaml
      - shell: cmd
        run: |
          CALL conda info -a
          if errorlevel 1 exit 1
          echo "OK"
```

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : base
    active env location : C:\conda_bin\envs\conda-test-env
            shell level : 1
       user config file : C:\Users\runneradmin\.condarc
 populated config files : 
          conda version : 23.1.0.post50+81beee710
    conda-build version : 3.23.3
         python version : 3.10.9.final.0
       virtual packages : __archspec=1=x86_64
                          __win=0=0
       base environment : C:\conda_bin\envs\conda-test-env  (writable)
      conda av data dir : C:\conda_bin\envs\conda-test-env\etc\conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\conda_bin\envs\conda-test-env\pkgs
                          C:\Users\runneradmin\.conda\pkgs
                          C:\Users\runneradmin\AppData\Local\conda\conda\pkgs
       envs directories : C:\conda_bin\envs\conda-test-env\envs
                          C:\Users\runneradmin\.conda\envs
                          C:\Users\runneradmin\AppData\Local\conda\conda\envs
               platform : win-64
[...]

PR #13162, opened 2023-09-29
Remove `dev/linux`, `dev/macos`, `dev/windows` in favor of streamlined GH workflow

kenodegard:
### Description

Changes include:
- Bring `dev/linux` inline in `tests.yml`
- Remove linux images
- Use standard QEMU images instead of custom images
- Bring `dev/macos` inline in `tests.yml`
- Bring `dev/windows` inline in `tests.yml`
- ~Enable alpha sorting on `requirements.txt` files~ (cherry picked into https://github.com/conda/conda/pull/13166) 

Updates to `tests.yml`:
- Test matrix runs 3 integration tests and 2 unit tests per OS/Python/channel combo
- ~Adds Allure reports~ (no longer desired)

Two bash activation tests on Windows proved difficult to resolve neatly within this already large PR. Opted to mark them as skip and to tackle them separately.

Xref https://github.com/conda/conda-build/pull/5019
Resolves https://github.com/conda/conda/issues/12418

### Checklist - did you ...


jezdez:
@kenodegard FWIW conda-libmamba-solver is using the [images to run conda tests](https://github.com/conda/conda-libmamba-solver/blob/17c8338029b1c951da2c34d4a8e4dfc5be7e69b3/.github/workflows/tests.yml#L65) and it's super critical to not do a pull rug just before we do the October release.

There is probably a great opportunity to make it easier to maintain the cls tests as well, just be aware that it does take additional effort to stabilize such a change, please include that in your risk assessment. I'd love to get you and Jaime coordinate on this, to decide in which order and timeframe this change may be applied in cls as well.

jezdez:
@kenodegard Now that we're depending on conda-libmamba-solver, I'm wondering if it's time to reinvestigate this, since we may want to reduce the duplication of test runs between the conda repo and the conda-libmamba-solver repo. @jaimergp any suggestion related to this?

.github/workflows/tests.yml
```diff
@@ -64,85 +62,102 @@ jobs:
...
+          - test-type: unit
+            test-group: 3
     env:
-      OS: Windows
-      PYTHON: ${{ matrix.python-version }}
-      TEST_SPLITS: 3
-      TEST_GROUP: ${{ matrix.test-group }}
+      ALLURE_DIR: ${{ github.workspace }}\allure-results
+      PYTEST_MARKER: ${{ matrix.test-type == 'unit' && 'not integration' || 'integration' }}
+      PYTEST_NUMPROCESSES: '0'
```
jezdez, line 82: Does this mean "automatic"?

.github/workflows/tests.yml
```diff
@@ -64,85 +62,102 @@ jobs:
...
+            default-channel: conda-forge
+          - test-type: unit
+            test-group: 3
     env:
-      OS: Windows
-      PYTHON: ${{ matrix.python-version }}
-      TEST_SPLITS: 3
-      TEST_GROUP: ${{ matrix.test-group }}
+      ALLURE_DIR: ${{ github.workspace }}\allure-results
+      PYTEST_MARKER: ${{ matrix.test-type == 'unit' && 'not integration' || 'integration' }}
```
jezdez, line 81: Oooh, that's smart! Much nicer :)

.github/workflows/tests.yml
```diff
@@ -64,85 +62,102 @@ jobs:
...
+          key: cache-${{ env.HASH }}
 
-      - name: Python ${{ matrix.python-version }} on ${{ matrix.default-channel }}, ${{ matrix.test-type }} tests, group ${{ matrix.test-group }}
-        shell: cmd
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          condarc-file: .github\condarc-${{ matrix.default-channel }}
+          python-version: ${{ matrix.python-version }}
+          run-post: false  # skip post cleanup
```
jezdez, line 106: This'll run the tests in a non-base environment, are we sure that this won't be a problem with tests that assume to be running in base?

.github/workflows/tests.yml
```diff
@@ -214,43 +278,45 @@ jobs:
...
+          && conda create
+            --name test
+            --quiet --yes
+            --file tests/requirements.txt
+            --file tests/requirements-ci.txt
+            python=${{ matrix.python-version }}
+          && conda activate test
+          && conda info --verbose
+          && conda list --show-channel-urls
+          && pytest tests/test_api.py::test_DepsModifier_contract"
```
jezdez, line 319: We could combine this with the benchmark tests instead of this arbitrary old test?

jezdez (commented):
This is an amazing attempt to clean up some of this mess, but there are open questions:

- Without the Docker image, we should remove the gitpod config and docs
- What's the expected developer experience for dev/CI parity? Can we use https://github.com/nektos/act to run the same tests locally (at least on the native platforms)?
- How can conda-libmamba-solver run the conda tests with this new regime without access to pre-built Docker images?

conda/core/package_cache_data.py
```diff
@@ -71,10 +72,10 @@
 class PackageCacheType(type):
     """This metaclass does basic caching of PackageCache instance objects."""
 
-    def __call__(cls, pkgs_dir):
+    def __call__(cls, pkgs_dir: str | os.PathLike | Path):
         if isinstance(pkgs_dir, PackageCacheData):
             return pkgs_dir
-        elif pkgs_dir in PackageCacheData._cache_:
+        elif (pkgs_dir := str(pkgs_dir)) in PackageCacheData._cache_:
```
jaimergp, line 78: I'm still a bit green with `:=` does this assignment propagate to the block under `else` or is it local to this block only?

tests/cli/test_main_clean.py
```diff
@@ -69,119 +68,124 @@ def has_pkg(name: str, contents: Iterable[str | Path]) -> bool:
...
-        pkgs_dir = Path(pkgs_dir)
-        assert pkgs_dir.is_dir()
+    # pkgs_dir is a directory
+    assert tmp_pkgs_dir.is_dir()
 
-        with tmp_env(pkg):
-            stdout, _, _ = conda_cli("clean", "--force-pkgs-dirs", "--yes", "--json")
-            json.loads(stdout)  # assert valid json
+    with tmp_env(pkg):
+        stdout, _, _ = conda_cli("clean", "--force-pkgs-dirs", "--yes", "--json")
```
jaimergp, line 82: can we assert this here just in case? It was the original issue:

```
assert context.pkgs_dirs == (tmp_pkgs_dir,)
```

conda/core/package_cache_data.py
kenodegard: yes, it propagates

it's a normal local variable, so any subsequent code can access it

the walrus operator did not change how scopes work

tests/cli/test_main_clean.py
kenodegard: this assertion is in the new `tmp_pkgs_dir` fixture

jaimergp: It's only part of it. It tests that it's the first one, but it doesn't assert that it's the only one:

```
assert context.pkgs_dirs[0] == pkgs_dir_str
```

In `test_main_clean` we need `context.pkgs_dirs` to ignore user config. Otherwise prepopulated caches might affect the outcome of the test (e.g. the user cache at ~/.conda/pkgs already has `zlib` (now `small-package`?) due to other tests or CI configs, so conda won't download it, so the new tmp pkgs_dir is not populated).

kenodegard: opted to mock `context.pkgs_dirs` instead of monkeypatching to resolve this issue, see https://github.com/conda/conda/pull/13511

.github/workflows/tests.yml
kenodegard: looks like we're good, we seem to have naturally drifted away from this expectation

.github/workflows/tests.yml
kenodegard: the docs only says this is the default value if `--pdb` is specified

looking at pytest-xdist's source it looks like a zero value means it runs inline to the spawning process, with `-n 1` it will spawn a child `popen` process to run the tests

.github/workflows/tests.yml
```diff
@@ -64,85 +62,87 @@ jobs:
...
+          --file tests\requirements-${{ runner.os }}.txt
+          --file tests\requirements-ci.txt
+          --file tests\requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
+
+      - name: Conda Info
+        run: conda info --verbose
+
+      - name: Conda List
+        run: conda list --show-channel-urls
```
jaimergp, line 121: I'd also add a `conda config --show-sources`.

jaimergp:
> What's the expected developer experience for dev/CI parity? Can we use https://github.com/nektos/act to run the same tests locally (at least on the native platforms)?

I'm also interested about the DX story for interactive debugging locally.

> How can conda-libmamba-solver run the conda tests with this new regime without access to pre-built Docker images?

I guess our only chance is to copy things and adjust the steps, and then have an upstream checker to manually port changes in the CI files ([something similar to this kind of system](https://github.com/napari/packaging/issues/105) we have in napari/packaging). Reusable workflows are tricky because we'd need to add extra steps and modify some commands and it's more trouble than it's worth for just one project, IMO.

.github/workflows/tests.yml
```diff
@@ -38,10 +35,11 @@ jobs:
...
       code: ${{ steps.filter.outputs.code }}
     steps:
-      - uses: actions/checkout@v3
-        # dorny/paths-filter needs git clone for push events
+      - uses: actions/checkout@v4
+        # dorny/paths-filter needs git clone for non-PR events
         # https://github.com/dorny/paths-filter#supported-workflows
         if: github.event_name != 'pull_request'
+
       - uses: dorny/paths-filter@v2.11.1
```
jaimergp, line 43: The `filters` argument here needs to be updated to drop the `dev/` files.

.github/workflows/tests.yml
jaimergp, line 1: Would you consider adding a single, yet parametrized when appropriate, workflow job for all platforms? I see a lot of duplication, but maybe the parametrization is more trouble than it's worth.

kenodegard: I was thinking to try that in a subsequent PR

.github/workflows/tests.yml
kenodegard: https://github.com/conda/conda/pull/13162/commits/80e38e8a3ce880ffcb57ebfb9a6723a51db5752c

.github/workflows/tests.yml
kenodegard: https://github.com/conda/conda/pull/13162/commits/e7fc559182b3c8dbd9271a664639e362f7c06ac3

.github/workflows/tests.yml
kenodegard: we could but why? combining them would results in benchmarks being run on all three archs (`linux-64`, `linux-arm64`, and `linux-ppc64le`)

kenodegard:
> > What's the expected developer experience for dev/CI parity? Can we use [nektos/act](https://github.com/nektos/act) to run the same tests locally (at least on the native platforms)?
> 
> I'm also interested about the DX story for interactive debugging locally.

I don't see any reason to change the DX, if anything this change to CI actually aligns it more closely with the local DX workflow.

> > How can conda-libmamba-solver run the conda tests with this new regime without access to pre-built Docker images?
> 
> I guess our only chance is to copy things and adjust the steps, and then have an upstream checker to manually port changes in the CI files ([something similar to this kind of system](https://github.com/napari/packaging/issues/105) we have in napari/packaging). Reusable workflows are tricky because we'd need to add extra steps and modify some commands and it's more trouble than it's worth for just one project, IMO.

Yeah I expect CLS to either completely diverge from conda's CI workflow and build their own images or adopt these proposed changes.

kenodegard:
Hm the `windows, conda-forge, integration` tests are take significantly longer to run than `windows, defaults, integration` (we're looking at 40-45min versus 3+hrs)

jaimergp:
> Hm the `windows, conda-forge, integration` tests are take significantly longer to run than `windows, defaults, integration` (we're looking at 40-45min versus 3+hrs)

Maybe a missing `--override-channels` in some tests. Some assume defaults as default and end up mixing channels which makes classic cry.

jaimergp:
I'd expect failures. osx-arm64 channels are less populated and I'm sure we'll find old tests that assumed availability of some version in osx-64. That is to say that it won't be a single commit change (changing the runner label) and maybe it's better to isolate that change in a separate PR.

kenodegard:
pytest-codspeed appears to mess with how `stdout`/`stderr` are handled and so `capsys` is unable to properly intercept the streams which is why our benchmark tests are failing, see https://github.com/CodSpeedHQ/pytest-codspeed/issues/24

kenodegard:
We're waiting on https://github.com/CodSpeedHQ/pytest-codspeed/pull/25 so until then we are using my pytest-codspeed fork

tests/__init__.py
```diff
@@ -6,10 +6,6 @@
 
 initialize_logging()
 
-from conda.testing import (
-    conda_ensure_sys_python_is_base_env_python,
-    conda_move_to_front_of_PATH,
-)
+from conda.testing import conda_move_to_front_of_PATH
 
-conda_ensure_sys_python_is_base_env_python()
```
jezdez, line 14: 🔥 🤗 🔥

jezdez (dismissed):

beeankha (approved):

tests/test_create.py
```diff
@@ -1405,7 +1405,7 @@ def test_shortcut_absent_does_not_barf_on_uninstall(
     request.addfinalizer(lambda: rmtree(shortcut_file.parent, ignore_errors=True))
 
     # including --no-shortcuts should not get shortcuts installed
-    with tmp_env("console_shortcut", "--no-shortcuts", prefix=prefix):
+    with tmp_env("main::console_shortcut", "--no-shortcuts", prefix=prefix):
```
jaimergp, line 1408: This might add channels to the solver when  `conda-forge` is the default channel. Can we instead use a conda-forge shortcut? e.g. https://anaconda.org/conda-forge/miniforge_console_shortcut

.github/workflows/tests.yml
```diff
@@ -151,58 +152,99 @@ jobs:
...
 
-      - name: Upload test results
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          conda-solver: libmamba
+          condarc-file: .github/condarc-${{ matrix.default-channel }}
+          python-version: ${{ matrix.python-version }}
+          run-post: false  # skip post cleanup
```
jaimergp, line 205:
```suggestion
        with:
          condarc-file: .github/condarc-${{ matrix.default-channel }}
          run-post: false  # skip post cleanup
```

These options are not used in the other CI entries. Also, since we are running `conda install` next, we can save one solve and add `python=${{ matrix.python-version }}` instead there.

.github/workflows/tests.yml
```diff
@@ -151,58 +152,99 @@ jobs:
...
+
+      - name: Conda Install
+        run: conda install
+          --yes
+          --file tests/requirements.txt
+          --file tests/requirements-${{ runner.os }}.txt
+          --file tests/requirements-ci.txt
+          --file tests/requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
```
jaimergp, line 214:
```suggestion
          ${{ env.REQUIREMENTS_TRUSTSTORE }}
          python=${{ matrix.python-version }}
```

Save one solve

.github/workflows/tests.yml
```diff
@@ -151,58 +152,99 @@ jobs:
...
+
+      - name: Conda Info
+        run: conda info --verbose
```
jaimergp, line 217: Don't we want `conda info --all`?

.github/workflows/tests.yml
```diff
@@ -211,29 +253,65 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: ~/conda_pkgs_dir
+          key: cache-${{ env.HASH }}
+
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          condarc-file: .github/condarc-defaults
+          python-version: ${{ matrix.python-version }}
```
jaimergp, line 286:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -211,29 +253,65 @@ jobs:
...
+          run-post: false  # skip post cleanup
+
+      - name: Conda Install
+        run: conda install
+          --yes
+          --file tests/requirements.txt
+          --file tests/requirements-${{ runner.os }}.txt
+          --file tests/requirements-ci.txt
+          --file tests/requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
```
jaimergp, line 296:
```suggestion
          ${{ env.REQUIREMENTS_TRUSTSTORE }}
          python=${{ matrix.python-version }}
```

.github/workflows/tests.yml
```diff
@@ -211,29 +253,65 @@ jobs:
...
+
+      - name: Pip Install
```
jaimergp, line 298:
```suggestion
      - name: Install CodSpeed
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
+          ". /opt/conda/etc/profile.d/conda.sh
+          && conda create
+            --name test
+            --yes
+            --file tests/requirements.txt
+            --file tests/requirements-${{ runner.os }}.txt
+            --file tests/requirements-ci.txt
+            python=${{ matrix.python-version }}
+          && conda activate test
+          && conda info --verbose
```
jaimergp, line 365:
```suggestion
          && conda info --all
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
-          # when locally dowloading and comparing results of different workflow runs.
-          name: test-results-${{ github.sha }}-linux-${{ matrix.platform }}-qemu-${{ matrix.default-channel }}-${{ matrix.python-version }}
-          path: |
-            .coverage
-            test-report.xml
-          retention-days: 1
+          ${{ matrix.image }}
+          bash -lc
+          ". /opt/conda/etc/profile.d/conda.sh
+          && conda create
```
jaimergp, line 357:
```suggestion
          && set -x
          && conda create
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
+          miniconda-version: latest  # auto-install conda if missing
+          architecture: ${{ runner.arch }}
+
+      - name: Conda Install
+        run: conda install
+          --yes
+          --file tests/requirements.txt
+          --file tests/requirements-ci.txt
+          --file tests/requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
```
jaimergp, line 440:
```suggestion
          ${{ env.REQUIREMENTS_TRUSTSTORE }}
          python=${{ matrix.python-version }}
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
-          ./dev/macos/setup.sh
-
-      - name: Python ${{ matrix.python-version }} on ${{ matrix.default-channel }}, ${{ matrix.test-type }} tests, group ${{ matrix.test-group }}
-        shell: bash -l {0}
-        run: |
-          ./dev/macos/${{ matrix.test-type }}.sh
-
-      - uses: codecov/codecov-action@v4
+          run-post: false  # skip post cleanup
+          miniconda-version: latest  # auto-install conda if missing
```
jaimergp, line 431: This will always install Miniconda, won't it? What about:

```suggestion
          miniconda-version: ${{ matrix.arch == 'osx-arm64' && 'latest' || null }}  # Miniconda not preinstalled in arm64 runners
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
+      - name: Conda Install
+        run: conda install
+          --yes
+          --file tests/requirements.txt
+          --file tests/requirements-ci.txt
+          --file tests/requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
+
+      - name: Conda Info
+        run: conda info --verbose
```
jaimergp, line 443:
```suggestion
        run: conda info --all
```

jaimergp:
It does look solver related:

```
4083.60s call     tests/cli/test_env.py::test_update[classic]
2469.53s call     tests/cli/test_env.py::test_update_env_only_pip_json_output[classic]
2288.18s call     tests/cli/test_env.py::test_create_dry_run_yaml[classic]
1761.93s call     tests/cli/test_env.py::test_create_valid_env_json_output[classic]
1760.75s call     tests/cli/test_env.py::test_remove_dry_run[classic]
1755.67s call     tests/cli/test_env.py::test_set_unset_env_vars[classic]
424.98s call     tests/test_create.py::test_shortcut_absent_when_condarc_set[classic]
306.58s call     tests/cli/test_cli_install.py::test_find_conflicts_called_once[classic]
279.13s call     tests/cli/test_main_remove.py::test_remove_globbed_package_names[classic]
261.81s call     tests/test_create.py::test_dont_remove_conda_2[classic]
120.10s call     tests/test_create.py::test_install_tarball_from_file_based_channel[classic]
```

.github/workflows/tests.yml
```diff
@@ -393,40 +482,38 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - name: Download test results
-        uses: actions/download-artifact@v4
+      - name: Download Artifacts
+        uses: actions/download-artifact@v3
```
jaimergp, line 486:
```suggestion
        uses: actions/download-artifact@v4
```

To fix warnings:

> aggregate
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/download-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

.github/workflows/tests.yml
```diff
@@ -393,40 +482,38 @@ jobs:
...
-    name: Analyze results
     needs: [windows, linux, linux-qemu, linux-benchmarks, macos, aggregate]
     if: '!cancelled()'
 
     runs-on: ubuntu-latest
     steps:
-      - name: Decide whether the needed jobs succeeded or failed
-        uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe
+      - name: Determine Success
+        uses: re-actors/alls-green@v1.2.2
```
jaimergp, line 509: Dependabot usually recommends pinning the hash to avoid tag replacement attacks. Dependabot can recognize the hashes as version approach and update them as necessary too, so I'd recommend switching back to hashes in 3rd party (i.e. no `actions/*`) actions.

.github/workflows/tests.yml
```diff
@@ -393,40 +482,38 @@ jobs:
...
-        # provides one downloadable archive of all .coverage/test-report.xml files
-        # of all matrix runs for further analysis.
+      - name: Upload Combined Test Results
+        # provides one downloadable archive of all matrix run test results for further analysis
         uses: actions/upload-artifact@v4
         with:
           name: test-results-${{ github.sha }}-all
-          path: test-results-${{ github.sha }}-*
-          retention-days: 90  # default: 90
+          path: test-results-*
```
jaimergp, line 493: There's a warning now in the Summary:

> [aggregate](https://github.com/conda/conda/actions/runs/8058911732/job/22096776879#step:3:10)
No files were found with the provided path: test-results-*. No artifacts will be uploaded.

.github/workflows/images.yml
jaimergp, line 1: I think this should be announced in the `news/`. And I'd also check if the docs require updates.

jaimergp: Also, we should decide if this means we will stop supporting DevContainer and Gitpod, which are not removed in this PR. Maybe the can be reconfigured to use the regular Miniconda/Miniforge images, but right now they point at the "latest" CI images.

jaimergp: And speaking of `news/`, maybe we should also mention that this PR fixes https://github.com/conda/conda/issues/12418 as the description states.

.github/workflows/tests.yml
kenodegard: From my experience I haven't seen dependabot updating hashes when a new version was available, perhaps theres some config option we need to enable?

tests/test_create.py
kenodegard: See https://github.com/conda/conda/pull/13162/commits/d6748bb3920533320ef2a276f5c4c06f7eb79a86

.github/workflows/tests.yml
```diff
@@ -53,96 +53,97 @@ jobs:
...
-          call .\dev\windows\${{ matrix.test-type }}.bat
+          path: ~/conda_pkgs_dir
+          key: cache-${{ env.HASH }}
 
-      - uses: codecov/codecov-action@v4
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          condarc-file: .github\condarc-${{ matrix.default-channel }}
+          python-version: ${{ matrix.python-version }}
```
kenodegard, line 103:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -53,96 +53,97 @@ jobs:
...
+          run-post: false  # skip post cleanup
+
+      - name: Conda Install
+        run: conda install
+          --yes
+          --file tests\requirements.txt
+          --file tests\requirements-${{ runner.os }}.txt
+          --file tests\requirements-ci.txt
+          --file tests\requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
```
kenodegard, line 113:
```suggestion
          ${{ env.REQUIREMENTS_TRUSTSTORE }}
          python=${{ matrix.python-version }}
```

.github/workflows/tests.yml
```diff
@@ -247,141 +325,152 @@ jobs:
...
-        if: matrix.default-channel == 'defaults'
-        env:
-          CONDA_SOLVER: classic
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
         with:
-          activate-environment: conda-test-env
-          miniconda-version: "latest"
+          condarc-file: .github/condarc-${{ matrix.default-channel }}
           python-version: ${{ matrix.python-version }}
```
kenodegard, line 429:
```suggestion
```

pyproject.toml
```diff
@@ -120,7 +120,7 @@ target-version = "py38"
 # see also https://docs.astral.sh/ruff/rules/
 select = ["E", "W", "F", "I", "D1", "UP", "ISC", "TCH", "T10", "FA"]
 # E402 module level import not at top of file
-# E501 line too long
+# E501 line too long (82 > 79 characters)
```
kenodegard, line 123:
```suggestion
# E501 line too long
```

.github/workflows/tests.yml
kenodegard: and we don't want to do `bash -lcx ...`?

kenodegard:
> Something about the benchmark tests looks like mixed channels as well, running way too long.

https://docs.codspeed.io/faq#why-are-my-benchmarks-taking-more-time-when-run-with-the-codspeed-github-action

.github/workflows/tests.yml
```diff
@@ -247,141 +327,159 @@ jobs:
...
-          - test-type: unit
-            test-splits: 3
-          - test-type: integration
-            test-splits: 2
-          # odd choice kept for parity with previous commit
-          - test-type: integration
-            default-channel: 'conda-forge'
-          - test-type: unit
-            default-channel: 'defaults'
+            test-group: 3
```
dholth, line 411: Is this line continuing to do fewer unit test groups on OSX to preserve limited resources?

.github/workflows/tests.yml
kenodegard: I'm not seeing this in the latest run, can you confirm?

.github/workflows/tests.yml
kenodegard: yup, the macOS matrix only runs 5 jobs:

- osx-64 + Python 3.8 + defaults + unit 1
- osx-64 + Python 3.8 + defaults + unit 2
- osx-arm64 + Python 3.12 + conda-forge + integration 1
- osx-arm64 + Python 3.12 + conda-forge + integration 2
- osx-arm64 + Python 3.12 + conda-forge + integration 3

.github/workflows/images.yml
kenodegard: See https://github.com/conda/conda/pull/13162/commits/e6fd209145979c755b2a761068fa91b3684151ef

jezdez (approved):
A nice to have would be to move the codspeed action inline while we're waiting for the PR to merge upstream, or add it to conda/actions. I don't mind either way, and we can also do it as a post-merge step

beeankha (approved):

dbast (approved):

.github/workflows/tests.yml
```diff
@@ -247,141 +327,160 @@ jobs:
...
-        shell: bash -l {0}
-        run: |
-          ./dev/macos/${{ matrix.test-type }}.sh
-
-      - uses: codecov/codecov-action@v4
+          condarc-file: .github/condarc-${{ matrix.default-channel }}
+          run-post: false  # skip post cleanup
+          # conda not preinstalled in arm64 runners
+          miniconda-version: ${{ (matrix.default-channel == 'defaults' && matrix.arch == 'osx-arm64') && 'latest' || null }}
+          miniforge-version: ${{ (matrix.default-channel == 'conda-forge' && matrix.arch == 'osx-arm64') && 'latest' || null }}
```
jaimergp, line 440: This construct might be useful in the other OSs as well, to get a fresh conda-forge-provided install with Miniforge, which might help with the long integration runtimes on Linux & Windows items that test against conda-forge.

.github/workflows/tests.yml
jaimergp: Confirmed!

.github/workflows/tests.yml
jaimergp: See this one PR https://github.com/conda-forge/docker-images/pull/256. [Config here](https://github.com/conda-forge/docker-images/blob/main/.github/dependabot.yml).

kenodegard: ah good good, I've opened https://github.com/conda/conda/issues/13657 to address this

PR #13171, opened 2023-10-03
Remove imports from pip/requests's vendored packages

kenodegard:
### Description

Stop relying on vendored packages provided by `pip` and `requests` and instead depend on the dependency itself.

Removes the following fallback imports:
- `pip._vendor.requests`
- `requests.packages.chardet` (added `charset-normalizer` as dependency)
- `pip._vendor.requests.packages.chardet` (added `charset-normalizer` as dependency)

> **Note**
> Opted to depend on `charset-normalizer` instead of `chardet` since the former is also the preferred library used by `requests`.

### Checklist - did you ...


beeankha (dismissed):

conda/cli/main_info.py
```diff
@@ -118,30 +118,14 @@ def print_package_info(packages):
...
+    except ImportError as err:
+        # ImportError: conda-build is not installed
+        log.debug("Unable to import conda-build: %s", err)
         conda_build_version = "not installed"
-    except Exception as e:  # pragma: no cover
-        conda_build_version = "Error %s" % e
-    else:  # pragma: no cover
-        conda_build_version = conda_build.__version__
+    except Exception as err:
+        log.error("Error to import conda-build: %s", err)
```
kenodegard, line 127:
```suggestion
        log.error("Error importing conda-build: %s", err)
```

conda/base/context.py
```diff
@@ -1043,13 +1043,15 @@ def _override(self, key, value):
     @memoizedproperty
     def requests_version(self):
         try:
```
kenodegard, line 1045:
```suggestion
        # used in User-Agent as "requests/<version>"
        # if unable to detect a version we expect "requests/unknown"
        try:
```

beeankha (dismissed):

recipe/meta.yaml
```diff
@@ -35,6 +35,7 @@ requirements:
     - tqdm >=4
   run:
     - boltons >=23.0.0
+    - chardet
```
kenodegard, line 38:
```suggestion
    - charset-normalizer
```

beeankha (dismissed):

jezdez (dismissed):
charset-normalizer is a pretty old version on defaults: https://anaconda.org/main/charset-normalizer

But a recent one on conda-forge: https://anaconda.org/conda-forge/charset-normalizer

Might be useful to review https://github.com/Ousret/charset_normalizer/blob/master/CHANGELOG.md if there are any features or fixes needed.

Otherwise LGTM

conda/cli/main_info.py
```diff
@@ -118,30 +118,14 @@ def print_package_info(packages):
...
+    except ImportError as err:
+        # ImportError: conda-build is not installed
+        log.debug("Unable to import conda-build: %s", err)
         conda_build_version = "not installed"
-    except Exception as e:  # pragma: no cover
-        conda_build_version = "Error %s" % e
-    else:  # pragma: no cover
-        conda_build_version = conda_build.__version__
+    except Exception as err:
+        log.error("Error importing conda-build: %s", err)
```
dholth, line 127: Could we remove the special ImportError handling and keep except Exception

kenodegard: of course we can, I'm indifferent either way

kenodegard: actually they indicate different kinds of information that gets displayed in `conda info` so I think its best to leave this

dholth: I mean could they be combined to make code coverage testing easier? I see that printing the exception is one way it shows the difference between import and other errors.

beeankha (approved):

Issue #13206, opened 2023-10-16
Error when using custom platform/subdir with an extra dash (e.g. `linux-64-2`)


calebzulawski:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

It errors at this line:
https://github.com/conda/conda/blob/fa57da0d2105b091d5ddffe50e379c8581b4d475/conda/core/index.py#L185

```
Traceback (most recent call last):
  File "conda/exceptions.py", line 1118, in __call__
  File "conda/cli/main.py", line 69, in main_subshell
  File "conda/cli/conda_argparse.py", line 91, in do_call
  File "conda/notices/core.py", line 109, in wrapper
  File "conda/cli/main_create.py", line 41, in execute
  File "conda/cli/install.py", line 264, in install
  File "conda/core/solve.py", line 136, in solve_for_transaction
  File "conda/core/solve.py", line 179, in solve_for_diff
  File "conda/core/solve.py", line 284, in solve_final_state
  File "conda/common/io.py", line 84, in decorated
  File "conda/core/solve.py", line 423, in _collect_all_metadata
  File "conda/core/index.py", line 175, in _supplement_index_with_system
  File "conda/plugins/manager.py", line 77, in get_hook_results
  File "conda/plugins/manager.py", line 78, in <genexpr>
  File "conda/plugins/virtual_packages/archspec.py", line 10, in conda_virtual_packages
  File "conda/core/index.py", line 196, in get_archspec_name
ValueError: too many values to unpack (expected 2)
```

### Additional Context

I'm using `conda-lock` installed via `mamba`, which is why I'm missing some of the conda context.  `conda-lock` is installing conda somewhere, I'm not sure the version.

My `.condarc` has
```
platform: linux-64-cxx03
```

travishathaway:
@calebzulawski,

Please see the discussion in the pull request I opened. There is still an ongoing discussion around the best possible solution.

PR #13207, opened 2023-10-17
Allow platform names with more than one dash

travishathaway:
Fixes: #13206

### Description

This improves the logic around separating a platform name into its distinct parts.

### Checklist - did you ...

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -182,7 +182,7 @@ def _supplement_index_with_system(index):
 def get_archspec_name():
     from conda.base.context import _arch_names, _platform_map, non_x86_machines
 
-    target_plat, target_arch = context.subdir.split("-")
+    target_plat, target_arch = context.subdir.split("-", maxsplit=1)
     # This has to reverse what Context.subdir is doing
     if target_arch in non_x86_machines:
         machine = target_arch
--- a/news/13207-platform-name-bugfix
+++ b/news/13207-platform-name-bugfix
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Improves the logic around separating a platform name into its distinct parts (#13207)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/core/index.py
```diff
@@ -182,7 +182,7 @@ def _supplement_index_with_system(index):
 def get_archspec_name():
     from conda.base.context import _arch_names, _platform_map, non_x86_machines
 
-    target_plat, target_arch = context.subdir.split("-")
+    target_plat, target_arch = context.subdir.split("-", maxsplit=1)
```
jezdez, line 185: With a platform of "linux-64" this correctly parses the arch into `64` and in line 192 returns "x86_64".

With this change applied, it may fix the `ValueError` reported in the ticket, but with the custom platform of `linux-64-cxx03` won't be returning `x86_64` anymore but `None`. @chenghlee do you have a suggestion how to handle custom platforms like this?

chenghlee: > do you have a suggestion how to handle custom platforms like this?

I would recommend custom platforms replace `-`s with `_`s in the [OS] platform and [hardware] architecture strings. Adding `maxsplit=1` smells like the wrong thing to do, because:

1. It implicitly asserts that _only_ the architecture component can have a `-`, which [I suspect] will inevitably lead to "but why can't my OS/platform have a `-` instead/as well?!" somewhere down the road.
2. "`subdir` contains one and only one `-`" has been around so long that it's effectively an undocumented but fundamental assumption throughout the conda ecosystem. So while adding `maxsplit` would solve  the user's _immediate_ problem, allowing `-` in either the platform or architecture subdir component will likely break something else in the conda ecosystem.

TL;DR: Supporting `-` in either the platform or architecture component of `subdir` is an ecosystem-wide, CEP-level initiative, and _not_ something we should add in a single line of code.

wolfv: I agree with Cheng here. We definitely also encode this assumption in mamba & rattler. Using another character would be better IMO.

calebzulawski: Since I'm not totally sure what detecting the platform is actually used for, is the suggestion to use `linux-64_cxx03` instead?  Wouldn't that still fail to detect as `x86_64`?  I'm also curious, what's the harm of failing to detect the platform correctly?

PR #13218, opened 2023-10-19
remove ruamel_yaml import/except fallback

dholth:
### Description

We removed `ruamel_yaml` a year ago.

The `conda install required-dependency` errors are odd - they might be a little hard to follow through on...

### Checklist - did you ...

### DIFF
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -54,14 +54,9 @@ try:
     from ruamel.yaml.reader import ReaderError
     from ruamel.yaml.scanner import ScannerError
 except ImportError:  # pragma: no cover
-    try:
-        from ruamel_yaml.comments import CommentedMap, CommentedSeq
-        from ruamel_yaml.reader import ReaderError
-        from ruamel_yaml.scanner import ScannerError
-    except ImportError:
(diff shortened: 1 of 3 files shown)


conda/common/configuration.py
```diff
@@ -54,14 +54,9 @@
...
-    try:
-        from ruamel_yaml.comments import CommentedMap, CommentedSeq
-        from ruamel_yaml.reader import ReaderError
-        from ruamel_yaml.scanner import ScannerError
-    except ImportError:
-        raise ImportError(
-            "No yaml library available. To proceed, conda install ruamel.yaml"
-        )
+    raise ImportError(
+        "No yaml library available. To proceed, conda install ruamel.yaml"
```
dholth, line 58: Should this message go away or be replaced with "reinstall conda with a different environment's conda"?

jezdez: Good question!

jezdez:
@dholth I don't think we need this fallback at all anymore since we're depending on `ruamel.yaml` now and it should just be installed upon upgrade, right?

jezdez (approved):

PR #13231, opened 2023-10-23
Resume downloads / refactor download() function

### LINKED ISSUE #8694: resume for single package download

### LINKED ISSUE #12752: Implement resume partial/incomplete package downloads
### Checklist

### What is the idea?

This feature request is readdressing past requests for providing a resume feature in conda:
https://github.com/conda/conda/issues/7
https://github.com/conda/conda/issues/800

As AI models and their supporting tools are now in the 1GB+ size, for those of us in countries where electricity and internet are consistently interrupted or slow, it's proving almost impossible to complete large updates. For example, pytorch-2.0.1-py3.10_cuda11.8_cudnn8_0.tar.bz2 has never completed in the last 24 hours and each restart starts from the beginning. Even for people in locations with fast internet, building in redundancy and resume logic will help reduce network congestion.

Possible implementation strategies may include:

- allow option for alternate download tool (wget, ...)
- check for partial downloads and complete those before starting a fresh download
- allow check/option for removing orphaned downloads

I am happy to be a tester for any effort applied to this feature request.

### Why is this needed?

As AI models and their supporting tools are now in the 1GB+ size, for those of us in countries where electricity and internet are consistently interrupted or slow, it's proving almost impossible to complete large updates.

### What should happen?

**Current:** 
User: conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 pytorch-cuda=11.7 -c pytorch -c nvidia

System: hangs or aborts about 2-3 hours

User: conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 pytorch-cuda=11.7 -c pytorch -c nvidia

System: Starts downloading all packages from the beginning!

**Ideal:**
User: conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 pytorch-cuda=11.7 -c pytorch -c nvidia

System: hangs or aborts about 2-3 hours

User: conda install pytorch==2.0.0 torchvision==0.15.0 torchaudio==2.0.0 pytorch-cuda=11.7 -c pytorch -c nvidia

System: Continues downloading all packages from where it was interrupted

### Additional Context

_No response_

dholth:
### Description

Fix #12752 

Begin by refactoring `download()` function, moving error handling outside of main function.

I've made this so that partial downloads only happen when size, hash is given which I think is done for packages. "Only resume partial download if etag matches" would be a different classic move.

I tested this by turning off wifi during a download. CTRL-C handling during parallel downloads is poor and does not stop most downloads. #13234 

The progress bars didn't initially make it to 100%, they did not take the already-downloaded content into account, should be working with latest revision.

The hashes didn't match when I'd typed in the 206 response code incorrectly, this is fixed and tested now.

I check the hash of `.partial` before putting it on its destination, but delete `.partial` for 4xx HTTP errors that are unlikely to succeed on retry. Unlike the current system that might leave files with the wrong hash but correct filename.

Does the current code check the hash of already-downloaded packages before unpacking?

### Checklist - did you ...


conda/gateways/connection/download.py
```diff
@@ -209,6 +118,126 @@ def download(
...
+        checksum_builder = NullChecksumBuilder()
+
+    size_builder = 0
+
+    with open(target_full_path, "wb") as fh:
+        streamed_bytes = 0
+        for chunk in resp.iter_content(chunk_size=CHUNK_SIZE):
+            # chunk could be the decompressed form of the real data
+            # but we want the exact number of bytes read till now
+            streamed_bytes = resp.raw.tell()
```
dholth, line 183: weird flex if this function is used for downloading packages, do we use the same function for repodata.json that might be `Content-Encoding: gzip`?

conda/gateways/connection/download.py
```diff
@@ -35,118 +36,26 @@
...
-                raise CondaError(
-                    message,
-                    url=url,
-                    target_path=target_full_path,
-                    content_length=content_length,
-                    downloaded_bytes=streamed_bytes,
-                )
+        yield
 
-        except OSError as e:
```
dholth, line 122: PR moves this catch up a level into the parent try/except block and fixes hardcoded 104

conda/gateways/connection/download.py
```diff
@@ -209,6 +118,126 @@ def download(
...
+            # but we want the exact number of bytes read till now
+            streamed_bytes = resp.raw.tell()
+            try:
+                fh.write(chunk)
+            except OSError as e:
+                message = "Failed to write to %(target_path)s\n  errno: %(errno)d"
+                # TODO: make this CondaIOError
+                raise CondaError(message, target_path=target_full_path, errno=e.errno)
+
+            checksum_builder.update(chunk)
```
dholth, line 191: updated version should checksum-on-file at the end since we can't resume a checksum

beeankha (approved):
I've manually tested this out and compared against current behavior on `main`, confirming that resuming downloads seem to be happening as intended with these code changes 👍

beeankha (approved):

PR #13237, opened 2023-10-25
Remove double-colon Examples::, replacing with Example: in CLI help

scruel:
### Description

OMIT

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -25,7 +25,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     description = summary
     epilog = dals(
         """
-        Examples::
+        Examples:
 
             conda clean --tarballs
         """
--- a/conda/cli/main_notices.py
+++ b/conda/cli/main_notices.py
@@ -24,7 +24,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     )
     epilog = dals(
         """
-        Examples::
+        Examples:
 
         conda notices
 
--- a/conda/cli/main_rename.py
+++ b/conda/cli/main_rename.py
@@ -31,7 +31,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     )
     epilog = dals(
         """
-        Examples::
+        Examples:
 
             conda rename -n test123 test321
 


beeankha:
@conda-bot check

dholth:
Was this done to format `Examples::` in https://docs.conda.io/projects/conda/en/stable/commands/clean.html#Output,%20Prompt,%20and%20Flow%20Control%20Options

beeankha:
> Was the double-colon :: included so that Examples:: would have a block formatted below itself in https://docs.conda.io/projects/conda/en/stable/commands/clean.html#Output,%20Prompt,%20and%20Flow%20Control%20Options

@dholth I think you're right, the formatting is now different:

_Before_
Screenshot 2023-10-25 at 1 01 46 PM

_After_
Screenshot 2023-10-25 at 1 01 29 PM

scruel:
Oh, interesting, after checking the other files, now I know what the double-colon is for. Its not quite clear for me to know the meaning of it.

scruel:
BTW, can we use indent rather than double-colon here, for making some section as a code block?
By doing this, we can respect the original MarkDown syntax, and make those in-code strings cleaner.

kathatherine:
I don't believe that would work. The double colon is required, as far as I'm aware, for the code formatting to exist when the next paragraph is indented: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

Definitely would be happy to be proved wrong, though. It's unfortunate that the double colon makes its way into the CLI help. @beeankha, did your local build without the double colons have any indents for the commands that are supposed to have code formatting, but don't?

beeankha:
@kathatherine the indents still show up if they're manually put into the file (see first two screenshots below); however, if indents are NOT put in manually in the docstring, the `::` seems to have no effect! (see the last two screenshots + scroll to the bottom of [this page](https://docs.conda.io/projects/conda/en/latest/commands/notices.html) for an example)
* * *

**Manually-placed indent:**

_[code]_
Screenshot 2023-10-25 at 1 48 49 PM

_[docs page]_
Screenshot 2023-10-25 at 1 01 29 PM

* * * 

**Non-effective double colon:**

_[code]_
Screenshot 2023-10-25 at 1 49 32 PM
_[docs page]_
Screenshot 2023-10-25 at 1 49 21 PM

kathatherine:
Yes, both the indent and double colon seem to be required, which is unfortunate for the CLI help. The only other way I know of is adding a `..code-block:: <LINTER-VALUE>` directive, but then that would probably show up in the CLI, which would be way more confusing than a double colon.

scruel:
> Yes, both the indent and double colon seem to be required, which is unfortunate for the CLI help. The only other way I know of is adding a `.. code-block:: <LINTER-VALUE>` directive, but then that would probably show up in the CLI, which would be way more confusing than a double colon.

How about using extra indent (8) to mark lines as code block, and reduce the extra indent (8 to 4) in the CLI?

kathatherine:
@scruel I don't know if that would work, but maybe we're not understanding each other. You're welcome to push more changes and we can test it out to see if we can make things better.

scruel:
@beeankha For your "Non-effective double colon:" example, I do think the code indent is wrong, otherwise why `conda notices` section is not be included in a code block? IMO some contributions might also confused about this syntax and they did some simple copy/paste workflows to create this command :)

beeankha:
> For your "Non-effective double colon:" example, I do think the code indent is wrong, otherwise why conda notices section is not be included in a code block?

@scruel if you're agreeing that an indentation is necessary there, then I think adding it via this PR would be great; you're correct that it's not showing up on the [currently live docs page](https://docs.conda.io/projects/conda/en/23.9.x/commands/notices.html). I think we can also remove `conda notices` as an example and just have `conda notices -c defaults` showing.

jezdez:
This seems to go in circles, do we actually have an issue here? https://github.com/conda/conda/pull/13257 seems to just replace "Examples" with "Example"

jezdez:
For the record, double-colons are part of the reStructureText markup language used by Python doc strings and Sphinx and should _not_ be removed: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

jezdez:
Closing as wontfix since the double colons are needed to render the code examples correctly in the docs.

Issue #12713, opened 2023-05-25
`conda init` cannot get elevated due to a `conda.gateways.subprocess` bug


jaimergp:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

We have seen this in several cases:

- https://github.com/conda/conda/issues/11530
- https://github.com/conda/conda/issues/12698
- https://github.com/conda/conda/issues/11885#issuecomment-1308346028

Today I stumbled upon this bug myself and found out why. If the user is installing to a location that is not user writable, `conda init` might ask for `sudo` access. Then it hits the `conda.gateways.subprocess` bug, where `stdin` has been passed as `str` from `conda.core.initialize` but it expected `bytes` instead.

According to the [`subprocess.Popen.communicate()` docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate):

> Interact with process: Send data to stdin. Read data from stdout and stderr, until end-of-file is reached. Wait for process to terminate and set the [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode) attribute. The optional input argument should be data to be sent to the child process, or None, if no data should be sent to the child. **If streams were opened in text mode, input must be a string. Otherwise, it must be bytes.**

(emphasis mine)

I think `conda.gateways.subprocess` needs to do some checks to ensure type consistency across streams.

### Conda Info

```shell
active environment : None
            shell level : 0
       user config file : /home/test_user/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 23.5.1.dev5
    conda-build version : 3.24.0
         python version : 3.10.10.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.31=0
                          __linux=5.10.25=0
                          __unix=0=0
       base environment : /opt/conda  (read only)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /opt/conda/pkgs
                          /home/test_user/.conda/pkgs
       envs directories : /home/test_user/.conda/envs
                          /opt/conda/envs
               platform : linux-64
             user-agent : conda/23.5.1.dev5 requests/2.28.2 CPython/3.10.10 Linux/5.10.25-linuxkit debian/11 glibc/2.31
                UID:GID : 1001:1001
             netrc file : None
[...]

Issue #12698, opened 2023-05-21
Anaconda3 Command Line Install for MacOS 13.3.1: TypeError Bug 


Rahul-Rut:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

OS: MacOS 13.3.1
Anaconda3-2023.03-1-MacOSX-arm64
I had previously installed Anaconda, but had to uninstall it for various reasons, and now when I tried reinstall-ing it using the pkg file, I kept running into an error and failed to install it. Upon trying the Command Line Install, I ran into the following Error Report:

`>>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/exceptions.py", line 1132, in __call__
        return func(*args, **kwargs)
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/cli/main.py", line 69, in main_subshell
        exit_code = do_call(args, p)
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/cli/conda_argparse.py", line 122, in do_call
        return getattr(module, func_name)(args, parser)
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/cli/main_init.py", line 33, in execute
        return initialize(context.conda_prefix, selected_shells, for_user, args.system,
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/core/initialize.py", line 113, in initialize
        run_plan_elevated(plan2)
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/core/initialize.py", line 709, in run_plan_elevated
        result = subprocess_call(
      File "/Users/USERNAME/anaconda3/lib/python3.10/site-packages/conda/gateways/subprocess.py", line 98, in subprocess_call
        stdout, stderr = process.communicate(input=stdin)
      File "/Users/USERNAME/anaconda3/lib/python3.10/subprocess.py", line 1154, in communicate
        stdout, stderr = self._communicate(input, endtime, timeout)
      File "/Users/USERNAME/anaconda3/lib/python3.10/subprocess.py", line 1985, in _communicate
        input_view = memoryview(self._input)
    TypeError: memoryview: a bytes-like object is required, not 'str'

### Conda Info

```shell
active environment : None
[...]

PR #13240, opened 2023-10-26
gateways.subprocess: fix subprocess_call when stdin is bytes

### LINKED ISSUE #11530: Anaconda failed to install on Mac OS Monterey Version 12.1
### Checklist

- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help
- [X] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?

When I try to install Anaconda Graphical Installer AND Command Line Installer, I get an error and it fails to install.

The GUI Installer gives a generic error message simply stating that Anaconda failed to install.

Instead with the command line error, I get the following python traceback after saying "yes" to "Do you wish the installer to initialize Anaconda3 by running conda init":

```

      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/exceptions.py", line 1114, in __call__
        return func(*args, **kwargs)
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/cli/main.py", line 86, in main_subshell
        exit_code = do_call(args, p)
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/cli/conda_argparse.py", line 90, in do_call
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/cli/main_init.py", line 50, in execute
        return initialize(context.conda_prefix, selected_shells, for_user, args.system,
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/core/initialize.py", line 118, in initialize
        run_plan_elevated(plan2)
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/core/initialize.py", line 687, in run_plan_elevated
        result = subprocess_call(
      File "/Users/gguarro/anaconda3/lib/python3.9/site-packages/conda/gateways/subprocess.py", line 100, in subprocess_call
        stdout, stderr = process.communicate(input=stdin)
      File "/Users/gguarro/anaconda3/lib/python3.9/subprocess.py", line 1134, in communicate
        stdout, stderr = self._communicate(input, endtime, timeout)
      File "/Users/gguarro/anaconda3/lib/python3.9/subprocess.py", line 1959, in _communicate
        input_view = memoryview(self._input)
    TypeError: memoryview: a bytes-like object is required, not 'str'
```

I also get this output:

`$ /Users/gguarro/anaconda3/bin/conda init zsh`

```
  environment variables:
                 CIO_TEST=<not set>
               CONDA_ROOT=/Users/gguarro/anaconda3
           CURL_CA_BUNDLE=<not set>
                 INFOPATH=/opt/homebrew/share/info:
                  MANPATH=/opt/homebrew/share/man::
                     PATH=/Library/Frameworks/Python.framework/Versions/3.10/bin:/opt/homebrew/b
                          in:/opt/homebrew/sbin:/Library/Frameworks/Python.framework/Versions/3.
                          7/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
       REQUESTS_CA_BUNDLE=<not set>
            SSL_CERT_FILE=<not set>

     active environment : None
       user config file : /Users/gguarro/.condarc
 populated config files : 
          conda version : 4.13.0
    conda-build version : 3.21.9
         python version : 3.9.12.final.0
       virtual packages : __osx=12.1=0
                          __unix=0=0
                          __archspec=1=arm64
       base environment : /Users/gguarro/anaconda3  (writable)
      conda av data dir : /Users/gguarro/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                          https://repo.anaconda.com/pkgs/r/osx-arm64
          package cache : /Users/gguarro/anaconda3/pkgs
                          /Users/gguarro/.conda/pkgs
       envs directories : /Users/gguarro/anaconda3/envs
                          /Users/gguarro/.conda/envs
               platform : osx-arm64
             user-agent : conda/4.13.0 requests/2.27.1 CPython/3.9.12 Darwin/21.2.0 OSX/12.1
                UID:GID : 503:20
             netrc file : None
           offline mode : False


future releases of conda.
Would you like conda to send this report to the core maintainers?
```

### Conda info

```shell
Anaconda failed to install so conda command is not found.
```

### Conda config

```shell
Conda command is not found.
```

### Conda list

```shell
Conda command is not found.
```

### Additional Context

_No response_

FFY00:
Fixes #11530, #12698, and #12713


FFY00:
I opted to open the process in text mode so that we don't have to decode the streams afterwards, instead of simply encoding the `stdin` argument to a string, but that'd be a valid fix too.

FFY00:
Running the test here to see there's actually anything that calls `subprocess_call` with `stdin` being bytes.

dholth:
Makes sense to me.

Should we switch to [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) as recommended by the Python docs or are we using something that is Popen-only?


```
New in version 3.6: Added encoding and errors parameters.

New in version 3.7: Added the text parameter as an alias for universal_newlines.
```

We have to go to https://docs.python.org/3/library/io.html#io.TextIOWrapper to get a proper explanation for `errors` parameter of Popen or run(), which is 

> errors is an optional string that specifies how encoding and decoding errors are to be handled. Pass 'strict' to raise a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) exception if there is an encoding error (the default of None has the same effect), or pass 'ignore' to ignore errors. (Note that ignoring encoding errors can lead to data loss.) 'replace' causes a replacement marker (such as '?') to be inserted where there is malformed data. 'backslashreplace' causes malformed data to be replaced by a backslashed escape sequence. When writing, 'xmlcharrefreplace' (replace with the appropriate XML character reference) or 'namereplace' (replace with \N{...} escape sequences) can be used. Any other error handling name that has been registered with [codecs.register_error()](https://docs.python.org/3/library/codecs.html#codecs.register_error) is also valid.

conda/gateways/subprocess.py
```diff
@@ -101,15 +101,12 @@ def subprocess_call(
         stdout=pipe,
         stderr=pipe,
         env=env,
+        text=True,  # open streams in text mode so that we don't have to decode
     )
     ACTIVE_SUBPROCESSES.add(process)
 
     # decode output, if not PIPE, stdout/stderr will be None
     stdout, stderr = process.communicate(input=stdin)
```
jaimergp, line 109: What if `stdin` is `bytes`? Do we need to decode that one?

conda/gateways/subprocess.py
```diff
@@ -101,15 +101,12 @@ def subprocess_call(
         stdout=pipe,
         stderr=pipe,
         env=env,
+        text=True,  # open streams in text mode so that we don't have to decode
```
jaimergp, line 104:
```suggestion
        text=True,  # open streams in text mode so that we don't have to decode
        errors="replace",
```

travishathaway:
xref: https://github.com/conda/conda/issues/13423

dholth (dismissed):

jezdez (changes requested):
This needs news and test.

conda/gateways/subprocess.py
dholth: Comparing to the above similar `any_subprocess` function that retains the hasattr(stdout, "decode")... code, should it match this updated version?

Can we use Python 3's builtin subprocess module "preferred API" instead?

kenodegard:
@jezdez test added in https://github.com/conda/conda/pull/13240/commits/0f499aeb18a196f1c59aaeb20f170aed42f7fae6 and news in https://github.com/conda/conda/pull/13240/commits/576fa690eb835905f5ce4e888ea79ac441172fc9 / https://github.com/conda/conda/pull/13240/commits/c194691da9028d3f91d280416e340a15de82d035

jezdez (approved):

conda/gateways/subprocess.py
kenodegard: @dholth probably

PR #13173, opened 2023-10-04
Inline imports in `conda.cli.main_*` and break up `conda.cli.conda_argparse`

kenodegard:
### Description

Resolves #13172

Changes:
- Lazy imports for all relative imports in `conda.cli.main_*`
- Move all argparse configuration functions from `conda.cli.conda_argparse` to the respective `conda.cli.main_*`
- Move custom `argparse.Actions` into `conda.cli.actions` (e.g., `NullCountAction`)
- Move all helper argparse functions into `conda.cli.helpers` (e.g., `add_parser_prefix`)

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #13242, opened 2023-10-26
Automatically configure subcommands via `list_submodules`.

scruel:
### Description
Automatically search subcommands by calling `list_submodules` util function, so that we can execute their `configure_parser` function in `conda.cli.conda_argparse.generate_parser` easily, see https://github.com/conda/conda/pull/13242/commits/425dfbe76ea27ee9c8cefc5e67f1dd8ea4347a5b.

===
IMO with `list_submodules` function, we can make this repo to be easier to add new commands. In the end, for creating a new command, the only thing we need to do will be created a new file that named as `main_<xxx>.py` (based on `TEMPLATE.md`) in `command` folder, rather than like now, we need to copy/paste those command names and paths everywhere in the source code.
This PR is mainly for demonstration, it is only meant to improve the maintainability and should not change or break any existing behavior.
Before continuing to modify the remain code, I would like to hear some feedback. :)

### Checklist - did you ...


travishathaway:
Hi @scruel,

Thank for you submitting this pull request and sparking a discussion. I do admire how this pull request helps conda adhere better to DRY (don't repeat yourself) principles, but I don't believe it's what this project needs in order to be easier to maintain. I give a couple reasons why below.

### We don't add commands that often

If we were dealing with a system that had to be more dynamic and adding new commands was a regular occurrence, this would be more important. But for conda, this is not the case, and we don't add new commands enough to warrant something like this.

### For new commands, we now prefer the "subcommand" plugin hook

When we do want to add new subcommands, we now prefer adding them as separate plugins. We have a dedicated plugin hook called "subcommands" that specifically allows us to do this. Using this plugin hook allows us to load new subcommands in a more dynamic way. More information on our plugin system can be found here: https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/index.html

An example of a new subcommand being offered via plugins is [conda-auth](https://github.com/conda-incubator/conda-auth)

I'll leave this pull request open so that others can chime in on the discussion.

scruel:
I am implementing my own CLI interface, and I realize that conda repo has related stuffs that I can learn from. So I started reading the related code, and during that time, I found that I might be able to improve some code in this repo.
The original impose of creating this PR is that I noticed you are currently refracting this part of the code, for example https://github.com/conda/conda/pull/13173, and this PR can make code cleaner in this opportunity for revision.

> We don't add commands that often

Sure, conda is currently a stable project, so this change seems not necessary now, but who knows, and what we do here can be a "once for all" thing to make the future better. :)

> For new commands, we now prefer the "subcommand" plugin hook

It's a cool design, but it's obvious to me that it's mainly for adding outside subcommands, rather than inside subcommands.
IMO users won't consider installing additional libraies to enable a subcommand if not necessary, and if a subcommand will be frequently used in the future, most of the user will require conda to have it built in.

@travishathaway, thank you for leaving it open, let's wait and see what else we can do.

travishathaway:
@scruel,

The subcommand plugin hooks can also be used for internal commands. For example, we have recently add a new `conda doctor` command and have chosen to go this route:

- https://github.com/conda/conda/tree/main/conda/plugins/subcommands

In the future, a more meaningful refactor would actually be to move all commands to this directory, so that they are all registered as "internal" plugins. You can see another example of this by looking at the "quetz" project (https://github.com/mamba-org/quetz)

The refactoring that we are currently doing here https://github.com/conda/conda/pull/13173 is relate to our work here:

- https://github.com/conda/conda/issues/11633

Although it's not on the roadmap yet, after completing that epic, we could very well move to refactor all conda subcommands as internally registered subcommand plugins.

We really appreciate your interest in this! Keep the questions coming if you have them 😄.

scruel:
Closed due to loss of interest.

PR #13244, opened 2023-10-26
Convert canary builds to be noarch

dbast:
There is no technical reason why a package build on macos-{64, arm64} or linux-{64, ppc64le, s390x} should be any different. So building two noarch variants of conda, one for unix and one for windows. The run constrain ensures that only the right variant can be installed.

While its is the goal of prod/defaults to build recipes for all platform / python combination to e.g. reveal if dependencies are missing in some of the combinations, the the goal of conda canary releases may differ in this regards to make canary builds as easy testable as possible on all platforms / python combinations.

Inspired by https://github.com/conda-forge/constructor-feedstock/blob/main/recipe/meta.yaml

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -411,8 +411,6 @@ jobs:
         include:
           - runner: ubuntu-latest
             subdir: linux-64
-          - runner: macos-latest
-            subdir: osx-64
           - runner: windows-latest
             subdir: win-64
     runs-on: ${{ matrix.runner }}
--- a/news/13244-canary-noarch
+++ b/news/13244-canary-noarch
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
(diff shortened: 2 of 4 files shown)


dholth (dismissed):
Makes sense to me.

dbast:
rebased

travishathaway:
@dbast,

Thanks for this pull request. I'd be happy to test this out. Will this get automatically added to canary builds? I thought there was a label for this but apparently that's not the case.

dbast:
@travishathaway Looked at the flows: so the label trigger based canary build flow failed due to no access to the canary upload token ... thats expected as workflows triggered from branches from a repo fork don't (and afaik never will) get repo secrets access ... so I have to:
* close this PR
* push the noarch branch as internal repo branch 
* create a new PR
* that new PR can be labeled after an initial review and will then produce a canary from the PR
* the resulting package can be tested to further review the PR :)

dbast:
@travishathaway here we go https://github.com/conda/conda/pull/13254

PR #13245, opened 2023-10-26
Issue 13243 Fix proposal for paths with up to 99 spaces

omdaniel:
### Description

This is a proposed fix to #13243 where Get-Conda Environment in Conda.psm1 ignores any part of path after the first space is encountered.

The approach is not fully robust in that paths with greater than 99 spaces would still have issues but for an extra character you could have 999 space paths! I've thought of some other methods but the ergonomics on this change involve minimal alternations to the current original implementation 

### Checklist


dbast:
@conda-bot check

PR #13246, opened 2023-10-26
check cancellation in download threads' progress callback

### LINKED ISSUE #13234: Conda does not stop parallel downloads on CTRL-C
### Checklist

### What happened?

On CTRL-C, conda does not stop parallel downloads. Instead, the progress bars continue to display.

Is the CTRL-C signal reaching a random download thread (or randomly, the manager thread)?

When CTRL-C is pressed, conda should immediately halt even if it is in the middle of a download.

This may be caused by libmamba-solver

conda 23.9

dholth:
### Description

Fix #13234

See also https://github.com/conda/conda-libmamba-solver/issues/337 which must be applied for CTRL-C handling to work at all, when libmamba is in use.

Python doesn't allow us to send exceptions to or kill individual futures. Instead, we check a flag as a side effect of progress bar handling. Since the progress bar is updated frequently this is fast.

Filter `CancelledError` to prevent shutdown messages that look like

```
CancelledError()
CancelledError()
CancelledError()
CancelledError()
CondaError: Cancelled
CondaError: Cancelled
CondaError: Cancelled
CondaError: Cancelled
```

(one for each pending or in-progress download) when it is cancelled.

"wifi turned off" does not instantly cancel downloads but does work. It will retry a couple of connections and then exit. Mainly unrelated to this PR.

### Checklist - did you ...


conda/core/package_cache_data.py
```diff
@@ -816,40 +824,48 @@ def execute(self):
...
-                extract_future.add_done_callback(
-                    partial(
-                        done_callback,
-                        actions=(cache_action, extract_action),
-                        exceptions=exceptions,
-                        progress_bar=progress_bars[prec_or_spec],
-                        finish=True,
-                    )
-                )
+                    for completed_future in as_completed(futures, timeout=1):
```
dholth, line 829: According to one stackoverflow post, at least subprocess waits work better when there is a timeout. Perhaps this is also true for as_completed?

news/context-fetch-threads
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Allow `CONDA_FETCH_THREADS`/`fetch_threads`, which sets parallel package
+  downloads, to be overridden from its default value of `5`.
```
beeankha, line 8: Looks like this news file was included by accident, from https://github.com/conda/conda/pull/13263

news/13234-cancel-parallel-downloads
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Improve CTRL-C (cancellation) handling of parallel dowload threads (#13234)
```
beeankha, line 7:
```suggestion
* Improve CTRL-C (cancellation) handling of parallel download threads. (#13234)
```

news/context-fetch-threads
dholth: I merged that PR into this one on purpose.

beeankha (approved):
I tested out this PR locally and everything seems to work with ctrl+c interruptions during conda env creation and also when installing packages into environments; when I did the same things in `main`, the creation of envs/installing of packages wouldn't immediately stop with a ctrl+c, and also sometimes I would get a `CorruptedEnvironmentError` that I couldn't seem to fix (i.e. this PR seems to solve those issues!)

Before this gets merged, one minor change I'd like to request is for the two news files to be combined into one.

beeankha (approved):

PR #13254, opened 2023-10-27
Convert unix canary builds to be noarch

dbast:
There is no technical reason why a package build on macos-{64, arm64} or linux-{64, ppc64le, s390x} should be any different. So building two noarch variants of conda, one for unix and one for windows. The run constrain ensures that only the right variant can be installed.

While its is the goal of prod/defaults to build recipes for all platform / python combination to e.g. reveal if dependencies are missing in some of the combinations, the the goal of conda canary releases may differ in this regards to make canary builds as easy testable as possible on all platforms / python combinations.

Inspired by https://github.com/conda-forge/constructor-feedstock/blob/main/recipe/meta.yaml

This is a followup PR via internal branch (for canary builds via added label) to https://github.com/conda/conda/pull/13244

### Checklist - did you ...


jaimergp (dismissed):

dbast:
this produces some interesting error: installing via `conda install -c conda-canary/label/conda-conda-pr-13254 conda` on osx-arm64 fails with `RemoveError: '__unix' is a dependency of conda and cannot be removed from`... what can that be?

recipe/meta.yaml
```diff
@@ -34,6 +36,7 @@ requirements:
     - ruamel.yaml >=0.11.14,<0.18
     - tqdm >=4
   run:
+    - __unix                       # [unix]
```
travishathaway, line 39: Why was this added as a run dependency? Seems a little weird to me.

jaimergp: Since it will land on `noarch`, we need to exclude it from the index in Windows. This requirement makes the solver ignore it because `__unix` is not present on Windows.

See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-packages-with-os-specific-dependencies

dholth: Remind me why the same noarch build can't also be used on Windows?

dbast: there are windows only run dependencies... so we can do a windows and unix noarch build... the windows noarch build failed during testing with conda.exe not found error... so this PR currently produces a unix noarch build and regular windows builds...

jaimergp: We can have a Windows-noarch buils and save the per-Python builds.

kenodegard:
> this produces some interesting error: installing via `conda install -c conda-canary/label/conda-conda-pr-13254 conda` on osx-arm64 fails with `RemoveError: '__unix' is a dependency of conda and cannot be removed from`... what can that be?

@dbast I get the same

jaimergp:
It seems to come from this function:

https://github.com/conda/conda/blob/3e5989de84908a25358ab8e566e9ba00a02e8629/conda/core/link.py#L689-L690

However, `__unix` should never be in the unlink list. The solver transaction should have trimmed it out. Looks like a bug. Does it happen with `libmamba` too?

dbast:
Yes, that happens here with libmamba.

recipe/meta.yaml
jezdez: @dbast What do you think about @jaimergp's suggestion of a Windows specific noarch build to cover for the Windows runtime dependencies?

kenodegard: This needs to be refactored now that we've added universal menuinst support (which should remove the need for a special windows package).

PR #13263, opened 2023-10-30
Fix CONDA_FETCH_THREADS parameter

dholth:
### Description

Would always be the default of 5 even if `CONDA_DEFAULT_THREADS` was set. IIRC I was also unable to set `fetch_threads` in `~/.condarc`?

What is the correct precedence? I think when default threads is overridden, then all other thread counts must be set or else they equal default threads. Otherwise the individual tunables should win out.

We could try to preserve a distinction between 0 (use DummyExecutor, do not use threading at all), 1 (use a single real Thread), and None (pick a default based on testing or core count/system resources). The current code doesn't seem well suited for that.

### Checklist - did you ...

### DIFF
--- a/news/context-fetch-threads
+++ b/news/context-fetch-threads
@@ -1,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Allow `CONDA_FETCH_THREADS`/`fetch_threads`, which sets parallel package
+  downloads, to be overridden from its default value of `5`.
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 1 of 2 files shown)


news/context-fetch-threads
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Allow `CONDA_FETCH_THREADS`/`fetch_threads`, which sets parallel package
+  downloads, to be overridden from its default value of `5`.
```
beeankha, line 8:
```suggestion
  downloads, to be overridden from its default value of `5`. (#13263)
```

beeankha (approved):

PR #13266, opened 2023-10-30
Implement `__conda` virtual package

### LINKED ISSUE #13308: Add `__conda` virtual package functionality
### Checklist

### What is the idea?

Ahead of the work to make conda's virtual packages pluggable in #11811, @isuruf opened https://github.com/conda/conda/pull/11764 to add a `__conda` virtual packages, so that packages can rely on a specific version of conda as part of their dependencies. As he described in the PR:

> This is useful for when limiting a package to a conda version. (For eg: we wanted to limit python 3.10 installation to conda <4.13 because of noarch: python issues there, but couldn't and had to use an ugly hack to allow python 3.10 to work on conda 4.13)

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

kenodegard:
### Description

Add the `__conda` virtual package using the virtual package plugin hook.

Also expand the virtual package test suite for better coverage.

Supersedes #11764 

Resolves #13308

### Checklist - did you ...

### DIFF
--- a/tests/plugins/test_virtual_packages.py
+++ b/tests/plugins/test_virtual_packages.py
@@ -116,34 +116,42 @@ def get_virtual_precs() -> Iterable[PackageRecord]:
     "subdir,expected",
     [
         # see conda.base.constants.KNOWN_SUBDIRS
-        ("emscripten-wasm32", []),
-        ("freebsd-64", ["__archspec", "__unix"]),
-        ("linux-32", ["__archspec", "__linux", "__unix"]),
-        ("linux-64", ["__archspec", "__linux", "__unix"]),
-        ("linux-aarch64", ["__archspec", "__linux", "__unix"]),
-        ("linux-armv6l", ["__archspec", "__linux", "__unix"]),
-        ("linux-armv7l", ["__archspec", "__linux", "__unix"]),
-        ("linux-ppc64", ["__archspec", "__linux", "__unix"]),
-        ("linux-ppc64le", ["__archspec", "__linux", "__unix"]),
-        ("linux-riscv64", ["__archspec", "__linux", "__unix"]),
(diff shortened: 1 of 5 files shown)


tests/plugins/test_virtual_packages.py
```diff
@@ -92,71 +98,113 @@ def test_cuda_override_none(clear_cuda_version):
...
-                context.subdirs,
-                (),
-                context.repodata_fns[0],
-            )
-            virtual = [p for p in packages if p.channel.name == "@"]
-            osx_exported = any("osx" in p.name for p in virtual)
-            assert osx_exported == bool(version)
+    monkeypatch.setenv("CONDA_SUBDIR", "osx-64")
+    monkeypatch.setenv("CONDA_OVERRIDE_OSX", version or "")
+    reset_context()
```
jaimergp, line 200: Won't this leave the `context` object in a modified state after the test is done? Or is there a fixture reverting that silently?

kenodegard: technically yes it will remain in a modified state, in most cases this wont be an issue (`conda_cli` for example ends up effectively doing a `reset_context`)

jezdez (approved):

tests/plugins/test_virtual_packages.py
jaimergp: I'm just concerned this ends up being a corner case where we leak state into other tests that are not doing any monkey patching or context resetting, and given our "split-by-duration" logic, test ordering is shuffled all the time. So we might even run things side by side with other module. It's just safer to restore the context once done.

beeankha (approved):

isuruf (approved):

PR #12996, opened 2023-08-11
Lock repodata cache by default

dholth:
### Description

TODO figure out what to do with `--experimental=lock`.  The intention would be to keep all `--experimental=` options forever to not break configs as unknown ones create errors in the argument parser, but it might not be necessary to retain them in the command line help.

Fix #12920 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13273, opened 2023-11-01
enable repodata.jlap/repodata.json.zst by default

### LINKED ISSUE #13120: Enable .jlap by default
### Checklist

### What is the idea?

[`.jlap` provides bandwidth savings for repodata updates.](https://conda.discourse.group/t/how-we-reduced-condas-index-fetch-bandwidth-by-99/257/4) by transferring differences between `repodata.json` instead of transferring the entire file each time. Can we enable `.jlap` by default? When the remote server does not provide `.jlap`, this fact is cached for a week before it checks again.

Conda's `.jlap` code also enables `repodata.json.zst` which is not available in the older non-jlap code path. Servers such as `anaconda.org` may have increased or decreased CPU usage compared to the current `repodata.json`' `Content-Encoding: gzip`. (If `gzip` happens on the reverse proxy, it's offloaded from the application server; if `gzip` happens in the application server, `zst` should be a faster alternative)

### Why is this needed?

Repeated conda index updates will see dramatically reduced bandwith requirements.

### What should happen?

Right now jlap is enabled by placing `experimental: [jlap]` in condarc or `--experimental=jlap` on the command line.

Instead, `--jlap / --no-jlap` could be passed.

The `experimental` version of option should continue to work and should never produce an error just by being there.

### Additional Context

We will have to improve tests against non-https channels.

### LINKED ISSUE #13256: Enable repodata.json.zst by default
### Checklist

### What is the idea?

repodata.json.zst is faster to download than repodata.json and will always be a win over the latter. Depending on your network versus disk bandwidth it may be faster than jlap, and it is usable on the first request e.g. in a CI system when repodata.json is not yet cached. Enable repodata.json.zst by default.

We could use the jlap network code, that already knows how to fetch from alternate URL's depending on whether `.zst` and `jlap` is available or not, and add a separate "zst but not jlap" flag.

dholth:
### Description

Enable jlap by default. Modeled on #12996

Fix #13120 
Fix #13256 

We would want to update https://conda.discourse.group/t/how-we-reduced-condas-index-fetch-bandwidth-by-99/257 with the new enable/disable instructions.

### Checklist - did you ...

### DIFF
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -210,12 +210,10 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
             "conda config --describe repodata_fns."
         ),
     )
+    # Land additional experimental options here as needed; don't fail if old
+    # ones passed.
     channel_customization_options.add_argument(
-        "--experimental",
-        action="append",
-        choices=["jlap", "lock"],
-        help="jlap: Download incremental package index data from repodata.jlap; implies 'lock'. "
-        "lock: use locking when reading, updating index (repodata.json) cache. Now enabled.",
+        "--experimental", action="append", choices=["jlap", "lock"], help=SUPPRESS
     )
     channel_customization_options.add_argument(
         "--no-lock",
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -220,6 +220,11 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
         action="store_true",
         help="Disable locking when reading, updating index (repodata.json) cache. ",
     )
+    channel_customization_options.add_argument(
+        "--no-repodata-zst",
+        action="store_true",
+        help="Do not check for repodata.json.zst.",
+    )
     return channel_customization_options
 
 
--- a/tests/gateways/test_jlap.py
+++ b/tests/gateways/test_jlap.py
@@ -16,7 +16,7 @@ import requests
 import zstandard
 
 import conda.gateways.repodata
-from conda.base.context import conda_tests_ctxt_mgmt_def_pol, reset_context
+from conda.base.context import conda_tests_ctxt_mgmt_def_pol, context, reset_context
 from conda.common.io import env_vars
 from conda.core.subdir_data import SubdirData
 from conda.exceptions import CondaHTTPError, CondaSSLError
@@ -359,6 +359,7 @@ def test_jlap_flag(use_jlap):
         {"CONDA_NO_JLAP": not use_jlap},
         stack_callback=conda_tests_ctxt_mgmt_def_pol,
     ):
+        expected = use_jlap
         assert context.no_jlap is not use_jlap
 
         # now using a subclass of JlapRepoInterface for "check zstd but not jlap"
--- a/tests/gateways/test_jlap.py
+++ b/tests/gateways/test_jlap.py
@@ -352,9 +352,9 @@ def test_repodata_info_jsondecodeerror(
         assert any(record[0].startswith("JSONDecodeError") for record in records)
 
 
-@pytest.mark.parametrize("use_jlap", ["jlap", "jlapopotamus", "jlap,another", ""])
+@pytest.mark.parametrize("use_jlap", [True, False])
 def test_jlap_flag(use_jlap):
-    """Test that CONDA_EXPERIMENTAL is a comma-delimited list."""
+    """Test that CONDA_NO_JLAP controls the repo interface class."""
     with env_vars(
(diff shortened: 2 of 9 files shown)


tests/gateways/test_jlap.py
```diff
@@ -367,11 +367,12 @@ def warning(*args, **kwargs):
 def test_jlap_flag(use_jlap):
```
kenodegard, line 367: Looks like we should change the parametrization such that `use_jlap` is a boolean

tests/gateways/test_jlap.py
```diff
@@ -367,11 +367,12 @@ def warning(*args, **kwargs):
...
     with env_vars(
-        {"CONDA_EXPERIMENTAL": use_jlap},
+        {"CONDA_NO_JLAP": not use_jlap},
         stack_callback=conda_tests_ctxt_mgmt_def_pol,
     ):
-        expected = "jlap" in use_jlap.split(",")
-        assert ("jlap" in context.experimental) is expected
+        # expected = "jlap" in use_jlap.split(",")
+        # assert ("jlap" in context.experimental) is expected
+        expected = use_jlap
```
kenodegard, line 375:
```suggestion
        assert context.no_jlap is not use_jlap
```

tests/gateways/test_jlap.py
```diff
@@ -367,11 +367,12 @@ def warning(*args, **kwargs):
...
 
         expected_cls = interface.JlapRepoInterface if expected else CondaRepoInterface
```
kenodegard, line 377:
```suggestion
        expected_cls = interface.JlapRepoInterface if use_jlap else CondaRepoInterface
```

kenodegard (changes requested):

travishathaway (changes requested):
@dholth,

I'm still a little worried about the concerns raised in this comment:

https://github.com/conda-incubator/ceps/pull/20#issuecomment-1773250540

Before merging this, I would like to see some profiling that compares before and after statistics using these two new features. I do not want to make this the default without having some concrete measurements to justify it.

dholth:
We can save warm-cache users 2 seconds and cold-cache users 4 seconds fetching `conda-forge/linux-64/repodata.json` with this change as-is. It will cause [anaconda.org](http://anaconda.org/) to serve more dynamically generated repodata.json.zst but less gzip.
We could choose to remove the alternate zst and jlap formats if there was a dramatic problem. The conda client would notice those were missing and only check again every 7 days.
The “fast” implementation requires patched libmamba, conda-libmamba-solver etc. and won't be available for 23.11. It addresses the complaint that higher-bandwidth users spend more time and RAM parsing and re-serializing a complete repodata.json compared to unpacking a fresh repodata.json.zst. But without this change conda only downloads `repodata.json; Content-Encoding: gzip`. That `gzip` version is about 2 or 4 seconds slower for conda-forge/linux-64. The Python conda also does not attempt to save RAM with partial repodata.json parsing.
I’d like to see us ship the existing feature today. It will make conda faster for everyone. Then it can become magically twice faster in the next release with an updated libmamba etc., so that the bandwidth and time spent to update even conda-forge/linux-64/repodata.json melts away.

beeankha:
@dholth can this PR be closed since https://github.com/conda/conda/pull/13452 got merged?

dholth:
This one also enables `.jlap` by default, but the recently merged work enables `.zst` only. I think we ought to copy the flag implementation from here to be able to disable `.zst` in case the remote server is broken.

jezdez (commented):
Thinking about these disabling options a little more, I wonder if we're missing out an edge case, where a user wants to disable it via the config system, but enable JLAP for a specific command line call. Do we support this use case?

Do you think it would be hard to reverse `no_jlap` into a `jlap` option (which defaults to `True`) and make sure the `--no-<name>` form also works as expected?

conda/cli/helpers.py
```diff
@@ -210,12 +210,10 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
...
     )
+    # Land additional experimental options here as needed; don't fail if old
+    # ones passed.
     channel_customization_options.add_argument(
-        "--experimental",
-        action="append",
-        choices=["jlap", "lock"],
-        help="jlap: Download incremental package index data from repodata.jlap; implies 'lock'. "
-        "lock: use locking when reading, updating index (repodata.json) cache. Now enabled.",
+        "--experimental", action="append", choices=["jlap", "lock"], help=SUPPRESS
```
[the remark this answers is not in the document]
dholth: It would be good to always have a --no-option / --option "click"-style.

I added this experimental section so that I could easily add new flags. I wanted to leave them there forever so that conda releases didn't break anyone's config.

Issue #13271, opened 2023-11-01
conda rename using prefix deleted environment after cloning and silently failed(?)


pieris98:
### Reproduction steps

1. ran `conda env create -f ./environment.yml`  (file is from [here](https://github.com/nmwsharp/diffusion-net/blob/master/environment.yml)) this failed with output:

    ```
    CondaValueError: prefix already exists: /root/miniconda3
	```

1. I tried to create a conda environment with prefix:  `conda env create --prefix /root/miniconda3/envs -f ./environment.yml ` (I'm not sure if this was a correct command but it created an env with all packages so I assumed I made a minor mistake)

    Relevant output:
    ```
    done
    To activate this environment, use
         $ conda activate /root/miniconda3/envs
     To deactivate an active environment, use
         $ conda deactivate
    ```

2. I tried to rename with a dryrun first using `conda rename -p /root/miniconda3/envs -d diffnet`

    Output:
    ```
    Dry run action: clone /root/miniconda3/envs,/root/miniconda3/envs/diffnet
    Dry run action: rm_rf /root/miniconda3/envs
    ```

3. Actually ran the rename command without dryrun: `conda rename -p /root/miniconda3/envs  diffnet`

    Output:
    ```
    Source:      /root/miniconda3/envs
    Destination: /root/miniconda3/envs/diffnet
    Packages: 18
	Files: 55977
	
	Downloading and Extracting Packages
	Downloading and Extracting Packages
	Preparing transaction: done
	Verifying transaction: done
	Executing transaction: done
	```

#### Result:

`/root/miniconda3/envs` _directory_ deleted! and all environments within it deleted! Running `conda env list` outputs:

```
conda environments:

base                  *  /root/miniconda3
```

I believe this is a conda bug therefore I filed it to the team. Please let me know if you could reproduce the issue as well.

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : base                                                                                                                                                        
    active env location : /root/miniconda3                                                                                                                                            
            shell level : 1                                                                                                                                                           
       user config file : /root/.condarc                                                                                                                                              
 populated config files :                                                                                                                                                             
          conda version : 23.5.2                                                                                                                                                      
    conda-build version : not installed                                                                                                                                               
         python version : 3.11.4.final.0
       virtual packages : __archspec=1=x86_64
                          __cuda=12.2=0
                          __glibc=2.35=0
                          __linux=5.15.0=0
                          __unix=0=0
[...]

PR #13280, opened 2023-11-03
Fix prefix `rename` issue

beeankha:
🚧 _This PR is still a work in progress and not currently ready for review_ 🚧 

* * *

Resolves https://github.com/conda/conda/issues/13271

* * * 

With this change, anyone who creates an env with an `/envs` at the end like so:

```
$ conda env create --prefix /root/miniconda3/envs
```

... will still be able to do so, but will _not_ be able to rename this environment. 

If someone creates an environment named `envs` and thus has a prefix path with an additional `/envs` at the end (e.g., `/root/miniconda3/envs/envs`), they will still be able to rename that environment since it's not the directory that contains other environments.

* * *

While investigating this issue I found that if a rename was attempted on a non-existing environment (e.g., `/root/miniconda3/envs/env` (without the `s` at the end) vs `/root/miniconda3/envs/envs`),  then this traceback would show up:

```

      File "/GitHub/conda/conda/exception_handler.py", line 17, in __call__
        return func(*args, **kwargs)
      File "/GitHub/conda/conda/cli/main.py", line 83, in main_subshell
        exit_code = do_call(args, parser)
      File "/GitHub/conda/conda/cli/conda_argparse.py", line 194, in do_call
        result = getattr(module, func_name)(args, parser)
      File "/GitHub/conda/conda/cli/main_rename.py", line 128, in execute
        source = validate_src()
      File "/GitHub/conda/conda/deprecations.py", line 125, in inner
        return func(*args, **kwargs)
      File "/GitHub/conda/conda/deprecations.py", line 125, in inner
        return func(*args, **kwargs)
      File "/GitHub/conda/conda/cli/main_rename.py", line 89, in validate_src
        if Path(context.target_prefix).samefile(context.root_prefix):
      File "/GitHub/conda/devenv/Darwin/arm64/envs/devenv-3.10-c/lib/python3.10/pathlib.py", line 1006, in samefile
        st = self.stat()
      File "/GitHub/conda/devenv/Darwin/arm64/envs/devenv-3.10-c/lib/python3.10/pathlib.py", line 1097, in stat
        return self._accessor.stat(self, follow_symlinks=follow_symlinks)
    FileNotFoundError: [Errno 2] No such file or directory: '/GitHub/conda/devenv/Darwin/arm64/envs/devenv-3.10-c/envs/envs'

{output shortened}

consider using the --no-plugins option to turn off plugins.

Example: conda --no-plugins install <package>

Alternatively, you can set the CONDA_NO_PLUGINS environment variable on
the command line to run the command without plugins enabled.

Example: CONDA_NO_PLUGINS=true conda install <package>

future releases of conda.
Would you like conda to send this report to the core maintainers? [y/N]:
```

After this code change we would simply see:
```
CondaEnvException: The environment you are trying to rename does not exist
```


travishathaway (changes requested):
Looks good so far, but it appears you have forgotten to add content to the news file 🙃.

conda/cli/main_rename.py
```diff
@@ -79,17 +79,24 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
 @deprecated.argument("24.3", "24.9", "prefix")
 def validate_src() -> str:
     """
-    Validate that we are receiving at least one value for --name or --prefix
-    and ensure that the "base" environment is not being renamed
+    Validate that we are receiving at least one legitimate value for --name or
```
dholth, line 82:
```suggestion
    Validate that we are receiving at least one valid value for --name or
```

beeankha: I was struggling with that wording, thank you for this suggestion!

ForgottenProgramme (dismissed):

conda/cli/main_rename.py
```diff
@@ -79,17 +79,24 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
     """
     from ..base.context import context
     from ..exceptions import CondaEnvException
 
-    if Path(context.target_prefix).samefile(context.root_prefix):
-        raise CondaEnvException("The 'base' environment cannot be renamed")
-
-    if Path(context.target_prefix).samefile(context.active_prefix):
-        raise CondaEnvException("Cannot rename the active environment")
+    if Path(context.target_prefix).exists():
```
kenodegard, line 88: instead of re-creating the Path object several times set it ahead of time:

```suggestion
    prefix = Path(context.target_prefix)
    if prefix.exists():
```

kenodegard (changes requested):
I wonder if a better fix (or at least in addition to) is to disallow making `$CONDA_ROOT/envs` an environment in the first place. This would be done in `conda.cli.install.check_prefix`.

travishathaway (approved):

conda/cli/install.py
```diff
@@ -56,6 +56,10 @@ def check_prefix(prefix, json=False):
         raise CondaValueError(
             f"Cannot create a conda environment with '{os.pathsep}' in the prefix. Aborting."
         )
+    if prefix.endswith("/envs"):
```
kenodegard, line 59: I'm hesitant of making the condition this broad, we don't care if `/some/other/path/envs` is used or even `conda create --name envs`, we only care if `$CONDA_PREFIX/envs` is used.

I believe a better option is to check if `prefix.parent` is a conda env and if so to disallow certain protected directory names (`bin`, `etc`, `conda-meta`, `envs`, etc.).

We also should account for a symlink pointing to `$CONDA_PREFIX/envs`.

kenodegard: cc @chenghlee

conda/cli/install.py
```diff
@@ -56,6 +56,37 @@ def check_prefix(prefix, json=False):
...
+        "/sbin",
+        "/share",
+        "/shell",
+        "/ssl",
+    ]
+    for dir in common_conda_dirs:
+        if dir in prefix:
+            raise CondaValueError(
+                f"Cannot create a prefix containing '{dir}'. Aborting."
+            )
```
kenodegard, line 88: I suggest we create either a helper function or a new `context` property that returns a tuple of "protected" directories (a superset of the `context.envs_dirs` value). That way we can do a simpler check:

```python
if prefix in context.protected_dirs:
    raise CondaValueError
```

Also where does this list come from?

beeankha:
Closing in favor of https://github.com/conda/conda/pull/13387

PR #13291, opened 2023-11-07
Increase logger level for CLI tests.

jezdez:
### Description

This should supersede #13284 as less drastic and focusing on the subcommand tests alone.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13286, opened 2023-11-06
Merge `23.10.x` into `main`.

jezdez:
Xref https://github.com/conda/conda/issues/13156

> **Warning**
> DO NOT SQUASH MERGE!

PR #13284, opened 2023-11-06
tests: Filter urllib3 retry warning log messages

mbargull:
### Description

Some tests like those in tests.cli.test_subcommands assert that there is no output on stderr.
If urllib3 retries a connection, it logs a warning message which lets those tests fail.

refs:
    - https://github.com/conda/conda-libmamba-solver/pull/358#issuecomment-1793574369

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13295, opened 2023-11-07
Sidestep urllib retries issue

kenodegard:
### Checklist - did you ...


travishathaway (commented):
Why not just delete this line?

jezdez:
@kenodegard I think we've had a few more of just this one test fail in the past weeks, that's why I thought a retry error could be ignored. Do you feel strongly by selectively fixing this?

kenodegard:
@travishathaway this is IMO a "short term" fix, the retries shouldn't be happening in the first place so I left it commented so future us can "fix" it (perhaps as part of the collapsing `conda_env` into `conda` effort)

@jezdez I worry that the urllib retries logging message is a red herring, if it were truly just a flaky test I would expect it to be much more erratic and to cause failures in other tests (e.g. `tests/cli/test_subcommands.py::test_create`). It's now failed 5 consecutive times for #13286 (see https://github.com/conda/conda/actions/runs/6784928710/job/18461168607?pr=13286) not to mention the failures originally observed in https://github.com/conda/conda-libmamba-solver/pull/358#issuecomment-1793574369.

mbargull:
> @jezdez I worry that the urllib retries logging message is a red herring, if it were truly just a flaky test I would expect it to be much more erratic and to cause failures in other tests (e.g. `tests/cli/test_subcommands.py::test_create`). It's now failed 5 consecutive times for #13286 (see https://github.com/conda/conda/actions/runs/6784928710/job/18461168607?pr=13286) not to mention the failures originally observed in [conda/conda-libmamba-solver#358 (comment)](https://github.com/conda/conda-libmamba-solver/pull/358#issuecomment-1793574369).

Nope, not a red herring.
Before I opened https://github.com/conda/conda/pull/13284 I did a quick test with
```sh
while true ; do curl -L https://repo.anaconda.com/pkgs/main/osx-arm64/repodata.json >/dev/null ; done
```
and it usually gives responses right away, but also regularly with a 5s or a bit over 10s delay.
The current default timeout is at `9.15`s.
Also the logs clearly indicate that it is only the retry log warnings that are on `stderr`.

mbargull (changes requested):
Let's go with https://github.com/conda/conda/pull/13291 .
Yes, the `assert not err` is too coarse. Completely ignoring it might be too, though.
If it's just the connection retry logs that affected this by now, we might just leave it as is for now.

kenodegard:
closing in favor of #13291

Issue #13297, opened 2023-11-07
Resolve system-level .condarc source differences in documentation, conda config initialization, and search path


ghost:
### What is the idea?

(I'm writing this for Windows.  There may be similar issues on other platforms but I have not tried.)

1) The documentation on [this page](https://conda.io/projects/conda/en/latest/user-guide/configuration/admin-multi-user-install.html) and [this page](https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html) and similar pages should be updated to no longer suggest placing .condarc files in the directory where Anaconda was installed, or the "root directory".  Instead, it should suggest placing them in a separate folder "C:/ProgramData/conda/".

2) When `conda config --system --set KEY VALUE` is run for the first time, the created file should be at C:/ProgramData/conda/.condarc instead of C:/ProgramData/Anaconda3/.condarc or C:/ProgramData/miniconda3/.condarc or whatever the install path was set to.

### Why is this needed?

The install paths for an "All Users" installation of Anaconda and Miniconda are not on the [search path](https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#searching-for-condarc).  Therefore, if a user of a multi-user installation accidentally or intentionally installs conda in one of their user environments, the system .condarc file in the install directory is ignored by conda because it is not on the search path.  This is because sys.prefix resolves to the environment prefix where conda is installed by the user.

This allows users to ignore system configuration defaults, as well as override #!final configuration parameters.  It is also confusing for administrators trying to set up multi-user installations.

### What should happen?

`(base) C:\Users\username>conda config --show-sources`
`==> C:\ProgramData\conda\.condarc <==`

### Additional Context

To reproduce:

- Place a .condarc file in the base environment directory (where Anaconda or Miniconda was installed for "All Users")
- Run `conda create -n new_env conda`
- Activate the new environment: `conda activate new_env`
- Check where configs are found: `conda config --show-sources`
- Observe that the .condarc file in the install directory is not listed.  Any restrictions on channels, packages, etc no longer apply to this environment.

kenodegard:
@shaese-auw you're right, this is equally applicable to unix systems as well:

https://github.com/conda/conda/blob/e69b2353c2f14fbfc9fd3aa448ebc991b28ca136/conda/base/constants.py#L28-L42

fPkX6F1nGTX:
[...]

PR #13301, opened 2023-11-08
System condarc location update (#13297)

samhaese:
### Description

Addresses issue #13297:
- References to "root directory" are now replaced with "system conda directory"
- An explanation of system conda directory is included in the multi-user administration page
- ~~The default system .condarc location is no longer the base environment~~

### Checklist - did you ...


samhaese:
pre-commit.ci autofix

travishathaway:
@conda-bot check

ForgottenProgramme (changes requested):
Hey, @samhaese!
Thank you for submitting this PR. :)
Could you please rename your news file to: `13297-system-condarc-location`  i.e. remove the `.txt` extention ?

mbargull (changes requested):
This would a big breaking change.
`conda config --system ...` always meant the installation's own configuration file.
This change would break a couple of setups, e.g.,
- multiple Conda installations on the same machine with different configurations,
- "packing up" pre-configured installations (e.g., restoring installations in CI caches).

While I agree that `--system` is an unfortunate naming (personally, I'd prefer something like `--base` for the current one and have `--system` like you suggested), it is not a change that can be done with a considerable deprecation phase, if at all.

kenodegard:
@mbargull agreed, thoughts on adding a new argument instead? `--global` or `--universal`?

samhaese:
I removed the change to the default system condarc path.  Instead, I added new language to the docs to clarify that there is a difference between a system-level and installation-level configuration, and that administrators of a multi-user installation should use system-level if they want to enforce #!final.

samhaese:
pre-commit.ci autofix

PR #13318, opened 2023-11-10
Add option to skip flexible search

### LINKED ISSUE #13315: Add option to skip flexible search in `conda search`
### Checklist

### What is the idea?

Provide an option to `conda search` to prevent the flexible search from being executed. For example, a flag like `--exact-name` to skip the code:

https://github.com/conda/conda/blob/25e4b1f77faa10c8efd44508de8a5bb011ff3520/conda/cli/main_search.py#L226-L233

### Why is this needed?

There is currently no way to do *only* an exact search with the existing code. If a user only wants exact search and the package is missing, they unnecessarily incur a second flexible search. Further, if the second search is successful, the exit code comes out as 0 - only if both searches are unsuccessful is an exit code 1 given.

For example, I was checking the status of a package `conda-forge::r-nlmixr2` and expecting to get an exit code 1 when it was not found. However, instead the search changed to `*r-nlmixr2*` which enable other packages to be found, resulting in an exit code 0. One thus needs to parse the output or work with JSON to determine if the exact search failed.

### What should happen?

If a package is not available with the exact name provided, then no flexible search should be performed and an exit code of 1 should be returned.

### Additional Context

_No response_

mfansler:
### Description

Adding option `--no-flexible-search` to `conda search` CLI allow users to skip fallback to flexible search. Default behavior is left intact.

Closes #13315 

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_search.py
+++ b/conda/cli/main_search.py
@@ -90,12 +90,16 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         default=NULL,
     )
     p.add_argument(
+        "--no-flexible-search",
+        action="store_true",
+        help="Do not perform flexible search if initial search fails.",
+    )
+    p.add_argument(
         "match_spec",
         default="*",
         nargs="?",
         help=SUPPRESS,
     )
-
     p.add_argument(
         "--canonical",
         action="store_true",
@@ -223,7 +227,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             SubdirData.query_all(spec, channel_urls, subdirs),
             key=lambda rec: (rec.name, VersionOrder(rec.version), rec.build),
         )
-    if not matches and spec.get_exact_value("name"):
+    if not matches and not args.no_flexible_search and spec.get_exact_value("name"):
         flex_spec = MatchSpec(spec, name="*%s*" % spec.name)
         if not context.json:
             print(f"No match found for: {spec}. Search: {flex_spec}")
--- a/news/13315-skip-flexible-search-option
+++ b/news/13315-skip-flexible-search-option
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Add `--no-flexible-search` option to skip flexible search. (#13315)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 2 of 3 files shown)


dholth:
Should we start doing [click](https://pypi.org/project/click/) style --flexible-search/--no-flexible-search (both options) for this kind of option

mfansler:
> Should we start doing [click](https://pypi.org/project/click/) style --flexible-search/--no-flexible-search (both options) for this kind of option

I think that might lead to confusion. Were I to see just a `--flexible-search` option, I'd expect `conda search --flexible-search name` would query `*name*` from the beginning, rather than as a fallback. Perhaps I should rename it to `--skip-flexible-search` to make it more semantically transparent that flexible is a default?

Honestly, I'd prefer this flexible fallback be entirely removed - one can always add a wildcard for flexibility. However, people asked for it (#7336 #7717), so I'm trying to be conservative.

travishathaway:
@mfansler,

I think this is a reasonable addition. I do agree with you that the default behavior of `conda search` is not preferable. I have been considering a complete re-write of `conda search` and have it just be a plugin (maybe call it `conda find` 🤷)

The point being is that we should avoid any bigger refactors of this command because rewriting it from the ground up would be the best thing IMO. We could then properly deprecate `conda search` and make the new search command available in the same way that `conda-libmamba-solver` is currently available (as a plugin, but bundled with conda installers).

Once those tests are passing, I'd be happy to give my approval.

dholth:
This looks like an interesting alternative, I don't see where it is exposed though. https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/repoquery.py is interesting.

jezdez:
@dholth There is a `repoquery` subcommand shipped with conda-libmamba-solver.

FWIW I don't think we should reinvent the `search` command, given that `repoquery` exists, unless of course the old search command supports things that it doesn't, which would then mean we could upstream some of that into mamba.

mfansler:
I elected to switch option name to `--skip-flexible-search`. Hopefully I've gotten the test correct now. Thanks everyone for the feedback! ❤️

dholth:
How do you feel about the word "exact"

mfansler:
> How do you feel about the word "exact"

Ambivalent.

mfansler:
> How do you feel about the word "exact"

No strong preference. You thinking `--exact-only`?

mfansler:
Anything else needed here? I believe this is ready for review.

travishathaway (approved):

beeankha (approved):

PR #13334, opened 2023-11-15
Use correct path for package cache on macOS.

jezdez:
### Description

I think this might reduce macOS runtime a little.

### Checklist - did you ...

- ~[ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -263,7 +263,7 @@ jobs:
       fail-fast: false
       matrix:
         default-channel: ['defaults', 'conda-forge']
-        python-version: ['3.9']
+        python-version: ['3.11']
         test-type: ['unit', 'integration']
         test-group: ['1', '2', '3']
         # fewer test splits for quicker unit tests, keep 3 splits for integration
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -294,9 +294,9 @@ jobs:
         uses: actions/cache@v3
         env:
           # Increase this value to reset cache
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -294,9 +294,9 @@ jobs:
         uses: actions/cache@v3
         env:
           # Increase this value to reset cache
-          CACHE_NUMBER: 4
+          CACHE_NUMBER: 5
         with:
-          path: ~/conda_pkgs_dir
+          path: ~/.miniconda/pkgs
```
jezdez, line 299: But maybe I'm wrong: https://github.com/conda-incubator/setup-miniconda/blob/99a08d57b7564700c6a69d09e1ff94f7e6482e7d/src/outputs.ts#L35

dholth: I get a ~/.conda/envs etc. for some commands, it is what conda does when it can't write the base installation (like how we set it up in Docker)

kenodegard: https://github.com/conda-incubator/setup-miniconda#caching-packages

dholth: Yikes. I guess it's possible to track down whether the github action sets the conda cache path with an environment variable, or to look up conda's exact cache path rules.

jezdez: Yeah, it was a trial to see if the cache was a problem for the macOS runners, the link I posted above sets the path it seems.

PR #13338, opened 2023-11-16
Ensure that the parent file of environments.txt exists on environment registration

marcoesters:
### Description

The `RegisterEnvironmentLocationAction` does not have a `verify()` step in its execution. Because of this the `.conda` directory for the `environments.txt` file is not created in some circumstances and the environment cannot be registered.

Add the `verify()` step to the execution.

### Checklist - did you ...

### Notes

* There is currently no test for this action and the path to the `.conda` directory is more or less hard-coded. So, I'm not sure how to test this in a safe way without affecting user systems.
* An alternative implementation would be to create the `.conda` directory in the `register_env` function. We could consider implementing both.


travishathaway:
Hi @marcoesters,

I went digging around the code today for everything that happens as part of the [PrefixActionGroup](https://github.com/conda/conda/blob/4.11.0/conda/core/link.py#L123), `RegisterEnvironmentLocationAction` being one of those.

What I discovered is that `RegisterEnvironmentLocationAction` does in fact have a `verify` method that is invoked before it's `execute` method. At a high level, this is happening here:

https://github.com/conda/conda/blob/3c0af6175bc930ae14c972f9d44426ed7dab7554/conda/core/link.py#L341-L353

Line 343 above will go through all the registered actions and call their `verify` methods, which `RegisterEnvironmentLocationAction` defines here:

https://github.com/conda/conda/blob/3c0af6175bc930ae14c972f9d44426ed7dab7554/conda/core/path_actions.py#L1049-L1059

That method will attempt to create a folder if it doesn't already exist.

All of this is to say that I do not believe adding this extra call to `.verify` actually does anything. This would explain why that statement is being missed by our code coverage checker. By the time it reaches calling that method, the `verify` method has already been called and `_verified` has been set to true. Furthermore, because we already call  `verify` at a higher level, this would introduce redundant, unnecessary calls to this method.

If you need some more background information about how these actions work, I found this document very helpful:

- https://docs.conda.io/projects/conda/en/latest/dev-guide/deep-dives/install.html#action-groups-and-actions-in-detail

Please let me know what you think. I'll be happy to discuss this further.

marcoesters:
@travishathaway thank you for your feedback! The warning that I'm seeing is coming from inside the `register_env` function. These lines in particular: https://github.com/conda/conda/blob/main/conda/core/envs_manager.py#L56-L58

I think a better way would be to add a call to `makedirs` to create the directory if it doesn't exist. I updated the code to reflect that. I think that's generally a good idea to ensure that directory exists before writing to that file.

If this doesn't solve the issue, the problem may indeed be in `constructor` or `menuinst`

travishathaway (dismissed):

conda/core/envs_manager.py
```diff
@@ -46,6 +46,16 @@ def register_env(location):
         # Nothing to do. Location is already recorded in a known environments.txt file.
         return
 
+    try:
+        os.makedirs(os.path.dirname(user_environments_txt_file), exist_ok=True)
```
jezdez, line 50:
```suggestion
    user_environments_txt_directory = os.path.dirname(user_environments_txt_file)
    try:
        os.makedirs(user_environments_txt_directory, exist_ok=True)
```

conda/core/envs_manager.py
```diff
@@ -46,6 +46,16 @@ def register_env(location):
...
+    except OSError as exc:
+        log.warn(
+            "Unable to register environment. "
+            "Could not create `.conda` directory in home directory. "
```
jezdez, line 54:
```suggestion
            f"Could not create {user_environments_txt_directory}. "
```

news/13338-add-verify-to-env-location-action
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Add verify step to RegisterEnvironmentLocationAction to create a .conda directory for the environments.txt file
```
jezdez, line 7:
```suggestion
* Create the `~/.conda` directory before trying to write to the `environments.txt` file. (#13338)
```

jezdez (changes requested):
Looking good, but needs to be more specific for users

jezdez (approved):

PR #13356, opened 2023-11-21
Enable sphinx nitpicky mode + turn warnings into errors to find broken refs

dbast:
`fail_on_warning: true` is equivalent to `-W --keep-going`.

### Checklist - did you ...


travishathaway:
@dbast,

I would most, if not all, of these warnings are coming from `autoapi`. I wonder if there is a way to allow warnings from this extension and disallow them from others.

travishathaway:
Nevermind,

It looks like this can be configured.

travishathaway:
Yes, I looked at those. These are going to be pretty tricky ignore. We'd most likely have to come up with some regex to do that for us.

It might be better to just build the theme, start some background Python HTTP server and then manually check the redirects? 🤷

dbast:
yeah. closing the PR... was interesting to see all the warnings... such checks have to be in place when a project starts, as later nobody wants to fix that many findings ... :/

PR #13357, opened 2023-11-22
Use `callback=reset_context` in `conda.plan`

jaimergp:
### Description

- Fixes https://github.com/conda/conda-libmamba-solver/issues/393
- Fixes https://github.com/conda/conda-libmamba-solver/issues/386

Thanks to all the research done by @mbargull:

We need to use `reset_context` instead of `reset_context_default` to prevent he latter from clearing configurations coming from `.condarc` files (see https://github.com/conda/conda-libmamba-solver/issues/393 for details). We also need `callback` instead of `stack_callback`, because:

> 1. conda.base.context.ContextStack is broken
>     (pop/replace don't call self.apply).
> 2. conda.plan is the only consumer of ContextStack
>     => we know nothing else adds to the stack => no need for it.
> 3. We know always use default search_path => no need for distinctions.

### Checklist - did you ...

### DIFF
--- a/conda/plan.py
+++ b/conda/plan.py
@@ -20,7 +20,7 @@ except ImportError:  # pragma: no cover
     from ._vendor.boltons.setutils import IndexedSet
 
 from .base.constants import DEFAULTS_CHANNEL_NAME, UNKNOWN_CHANNEL
-from .base.context import context, stack_context_default
+from .base.context import context, stack_context
 from .common.io import dashlist, env_vars, time_recorder
 from .common.iterators import groupby_to_dict as groupby
 from .core.index import LAST_CHANNEL_URLS, _supplement_index_with_prefix
@@ -493,7 +493,7 @@ def install_actions(
             "CONDA_ALLOW_NON_CHANNEL_URLS": "true",
             "CONDA_SOLVER_IGNORE_TIMESTAMPS": "false",
         },
-        stack_callback=stack_context_default,
+        stack_callback=stack_context,
     ):
         from os.path import basename
 
--- a/conda/plan.py
+++ b/conda/plan.py
@@ -20,7 +20,7 @@ except ImportError:  # pragma: no cover
     from ._vendor.boltons.setutils import IndexedSet
 
 from .base.constants import DEFAULTS_CHANNEL_NAME, UNKNOWN_CHANNEL
-from .base.context import context, stack_context
+from .base.context import context, reset_context
 from .common.io import dashlist, env_vars, time_recorder
 from .common.iterators import groupby_to_dict as groupby
 from .core.index import LAST_CHANNEL_URLS, _supplement_index_with_prefix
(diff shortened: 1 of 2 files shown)


mbargull:
Another thing I noticed was that `SubdirData._cache_` does not use `add_pip_as_python_dependency` as a key.
We should add it here https://github.com/conda/conda/blob/23.10.0/conda/core/subdir_data.py#L77 .
Plus, the `SubdirData.__init__` should then store the value of `context.add_pip_as_python_dependency` as an instance member (since loading of the actual data is not done during `__init__` immediately, i.e., the value from `context.add_pip_as_python_dependency` could change between calls to `__init__` and the value's actual usage in `_process_raw_repodata` afterwards.)

jaimergp:
Nothing seems to be breaking. Errors are due to https://github.com/conda/conda/issues/13360.

mbargull:
(I had this commit https://github.com/conda/conda/commit/2b308b112010070b4c9f2a02c2843cc9aebcdd86 in here too
>  Fix ContextStack.pop/replace reset last used args
> 
> N.B: After the recent change to conda.plan.install_actions to not use
> stack_context_default anymore, conda.base.ContextStack and related
> functions are only used for test code.
> As such, that code may be marked as deprecated and moved to the tests
> entirely.

to call `ContextStack`'s `apply` in `pop`/`replace`.
That broke a lot of tests immediately (probably esp. due to changing `replace`).
So, let's not change that code for now :D.
But the nota bene still applies: We may want to mark it as deprecated and move it to the tests.)

mbargull:
Tests fail because they are broken.
They load the index with `add_pip_as_python_dependency=False` through `conda.testing.helpers` but don't set `context.add_pip_as_python_dependency` accordingly.
Plus there's a comment at https://github.com/conda/conda/blob/23.10.0/conda/testing/helpers.py#L600
>         # We need CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY=false here again (it's also in
>         # get_index_r_*) to cover solver logics that need to load from disk instead of
>         # hitting the SubdirData cache

that kind of foreshadows more issues when fixing the tests...

mbargull:
Okay, the `SubdirData._cache_` `add_pip`-awareness needs more investigation.
I'm splitting this off into a separate PR so we can at least go ahead with the `conda.plan.install_actions` change which on its own should unblock (at least many, if not all) `conda-libmamba-solver`+`conda-build` issues.

mbargull:
> I'm adding this PR to the next milestone since it is effectively a blocker for `conda-libmamba-solver` (and thus `conda>=23.10` with its new solver default) adoption in common `conda-build` uses (e.g., for conda-forge).

And a note for other reviewers: The changes here are pretty much confined to `conda-build` since `conda.plan.install_actions` is not used within `conda` itself. Meaning, this is high impact for `conda-build` but low impact for `conda`.

jaimergp:
I think it's time to mark `tests/core/test_solve.py::test_conda_downgrade[libmamba]` as `xfail`able (strict=False). See https://github.com/conda/conda/pull/13367.

kenodegard (approved):

Issue #13292, opened 2023-11-07
Modernize `tests/test_create.py`


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

`tests/test_create.py` contain 102 tests. A lot of them contain outdated testing functionality and needs updating. Several prior attempts have been made to update this file but each effort failed as theres simply too many tests to do all at once. Let's instead take this piecemeal and do one at a time.

### Linked Issues & PRs

- [ ] `test_install_python_and_search`
- [ ] `test_run_preserves_arguments`
- [ ] `test_create_install_update_remove_smoketest`
- [ ] `test_install_broken_post_install_keeps_existing_folders`
- [ ] `test_safety_checks`
- [ ] `test_json_create_install_update_remove`
- [ ] `test_not_writable_env_raises_EnvironmentNotWritableError`
- [ ] `test_conda_update_package_not_installed`
- [ ] `test_noarch_python_package_with_entry_points`
- [ ] `test_noarch_python_package_without_entry_points`
- [ ] `test_noarch_python_package_reinstall_on_pyver_change`
- [ ] `test_noarch_generic_package`
- [ ] `test_override_channels`
- [ ] `test_create_empty_env`
- [ ] `test_strict_channel_priority`
- [ ] `test_strict_resolve_get_reduced_index`
- [ ] `test_list_with_pip_no_binary`
- [ ] `test_list_with_pip_wheel`
- [ ] `test_compare_success`
- [ ] `test_compare_fail`
- [ ] `test_install_tarball_from_local_channel`
- [ ] `test_tarball_install`
- [ ] `test_tarball_install_and_bad_metadata`
- [ ] `test_update_with_pinned_packages`
- [ ] `test_pinned_override_with_explicit_spec`
- [ ] `test_remove_all`
- [ ] `test_allow_softlinks`
- [ ] `test_remove_features`
- [ ] `test_dash_c_usage_replacing_python`
- [ ] `test_install_prune_flag`
- [ ] `test_remove_force_remove_flag`
- [ ] `test_install_force_reinstall_flag`
- [ ] `test_create_no_deps_flag`
- [ ] `test_create_only_deps_flag`
- [ ] `test_install_update_deps_flag`
- [ ] `test_install_only_deps_flag`
- [ ] `test_install_update_deps_only_deps_flags`
- [ ] `test_install_features`
- [ ] `test_clone_offline_simple`
[...]

PR #13410, opened 2023-12-08
Refactoring tests for `test_create.py`

travishathaway:
### Description

xref: https://github.com/conda/conda/issues/13292

I decided to just delete the test, "test_noarch_python_package_reinstall_on_pyver_change" because it does not make sense to have around any more. It is probably very seldom that users are actually switching major Python versions and this test is very slow to run.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -418,8 +418,13 @@ def test_not_writable_env_raises_EnvironmentNotWritableError(
     with tmp_env() as prefix:
         make_read_only(prefix / PREFIX_MAGIC_FILE)
 
-        with pytest.raises(CondaMultiError) as exc:
-            conda_cli("install", f"--prefix={prefix}", "ca-certificates", "--yes")
+        _, _, exc = conda_cli(
+            "install",
+            f"--prefix={prefix}",
+            "ca-certificates",
+            "--yes",
+            raises=CondaMultiError,
+        )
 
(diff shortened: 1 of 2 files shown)


kenodegard (changes requested):
`tmp_env` returns a `pathlib.Path` `prefix` so we should update all of the `os.path` stuff to use `pathlib.Path` logic instead, e.g.:

`isfile(join(prefix, py_file))` → `(prefix / py_file).is_file()`

tests/test_create.py
```diff
@@ -406,92 +407,129 @@ def assert_json_parsable(content):
...
 
+        with pytest.raises(CondaMultiError) as exc:
+            conda_cli("install", f"--prefix={prefix}", "ca-certificates", "--yes")
 
-def test_conda_update_package_not_installed():
-    with make_temp_env() as prefix:
-        with pytest.raises(PackageNotInstalledError):
-            run_command(Commands.UPDATE, prefix, "sqlite", "openssl")
+        assert len(exc.value.errors) == 1
+        assert type(exc.value.errors[0]) is EnvironmentNotWritableError
```
kenodegard, line 425:
```suggestion
        assert isinstance(exc.value.errors[0], EnvironmentNotWritableError)
```

kenodegard (approved):

PR #13419, opened 2023-12-11
add --keep-env flag 

### LINKED ISSUE #12485: ENH(remove): Add option to keep the (empty) environment despite removal of all packages
### What is the idea?

Conda environments are more than a list of packages. Metadata like channel priority, subdir etc. are also a crucial part.

However, the CLI seems to assume that empty environments serve no purpose - this is implicit in the behaviour of `conda remove`: when you remove all packages, it deletes the environment: `conda remove --all -n ENV_NAME` not just removes all packages, it _also_ removes the environment and its metadata.

Also, when you remove all packages without explicitly passing `--all` the environment itself _with_ the metadata is removed, despite this not necessarily being what the user wants. This has been discussed before in https://github.com/conda/conda/issues/10467

I would like to suggest a CLI option for `conda remove` that tells `conda remove` _not_ to delete the environment even if all packages are uninstalled.

This is consistent with environment creation: `conda create` can create _empty_ environments, so the existence of empty environments _is_ valid. Hence there's no reason why one shouldn't go back to empty environments from a full environment with remove.

### What should happen?

Ideally, the behaviour of `conda remove` would be changed to not delete the environment if all packages are gone. But since this is quite a deep change, it may be easier to add a new option that changes that default behaviour: `conda remove --all --keep-env -n ENV_NAME` could be how one tells remove to not remove the environment.

This should also work for `conda remove --keep-env scipy` in cases where remove would remove all packages without the shortcut `--all` being passed.

### Additional Context

See this StackOverflow question for further context: https://stackoverflow.com/questions/75731916/how-to-remove-all-packages-from-a-conda-environment-but-not-keep-environment-i

### Checklist

ForgottenProgramme:
### Description

Resolves #12485 

Adding a new `--keep-env` flag to be used with `conda remove --all`. It allows users to remove all the packages in the environment while retaining the environment itself.

- [x] Add a `--keep-env` flag to retain the environment when using `conda remove --all`
- [x] Create a new module `test_main_remove.py`. 
- [x] Move `test_remove_all` from `test/test_create.py` to this new module.
- [x] Use `tmp_env()` instead of `make_temp_env()` and `conda_cli()` instead of `run_command()`.
- [x] delete the original test from `tests/test_create.py`
- [x] Write test for `keep-env` flag.

### Checklist - did you ...


tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+)
+from conda.gateways.disk.delete import path_is_clean
+from conda.testing import CondaCLIFixture, TmpEnvFixture
+from conda.testing.integration import (
+    PYTHON_BINARY,
+    package_is_installed,
+)
+
+
+def test_remove_all(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
```
kenodegard, line 21: This is moved from `tests/test_create.py`, remember to delete the original

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+    with tmp_env("python") as prefix:
+        assert exists(join(prefix, PYTHON_BINARY))
+        assert package_is_installed(prefix, "python")
+
+        # regression test for #2154
+        with pytest.raises(PackagesNotFoundError) as exc:
+            conda_cli("remove", f"--prefix={prefix}", "python", "foo", "numpy")
```
kenodegard, line 28: `conda_cli` does not magically insert `--yes` like the older `run_command` implementation did so we need to be explicit

```suggestion
            conda_cli("remove", f"--prefix={prefix}", "python", "foo", "numpy", "--yes")
```

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+        exception_string = repr(exc.value)
+        assert "PackagesNotFoundError" in exception_string
+        assert "- numpy" in exception_string
+        assert "- foo" in exception_string
+
+        conda_cli("remove", f"--prefix={prefix}", "--all")
```
kenodegard, line 34:
```suggestion
        conda_cli("remove", f"--prefix={prefix}", "--all", "--yes")
```

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+        assert path_is_clean(prefix)
+
+
+def test_remove_all_keep_env(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
+    with tmp_env("python") as prefix:
+        assert exists(join(prefix, PYTHON_BINARY))
+        assert package_is_installed(prefix, "python")
+
+        conda_cli("remove", f"--prefix={prefix}", "--all", "--keep-env")
```
kenodegard, line 43:
```suggestion
        conda_cli("remove", f"--prefix={prefix}", "--all", "--keep-env", "--yes")
```

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from os.path import (
+    exists,
+    join,
+)
+
```
kenodegard, line 7:
```suggestion
```

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+import pytest
+
+from conda.exceptions import (
+    PackagesNotFoundError,
+)
```
kenodegard, line 12:
```suggestion
from conda.exceptions import PackagesNotFoundError
```

tests/cli/test_main_remove.py
```diff
@@ -0,0 +1,44 @@
...
+from conda.gateways.disk.delete import path_is_clean
+from conda.testing import CondaCLIFixture, TmpEnvFixture
+from conda.testing.integration import (
+    PYTHON_BINARY,
+    package_is_installed,
+)
```
kenodegard, line 18:
```suggestion
from conda.testing.integration import PYTHON_BINARY, package_is_installed
```

kenodegard (changes requested):
`tmp_env` returns a `pathlib.Path` object so we should use it as such

travishathaway:
@ForgottenProgramme,

I think it would be really nice to add an example to the help page:

https://github.com/conda/conda/blob/349ca85ce327ed76cfa514498f36d050c946b1ec/conda/cli/main_remove.py#L43-L58

jezdez (approved):

travishathaway (approved):

beeankha (approved):

kenodegard (approved):

PR #13449, opened 2023-12-19
Fix leaky tests

kenodegard:
### Description

`tests/test_create.py::test_legacy_repodata` recently started to fail on `main`, not sure whether due to a test reordering or whether some other change occurred to defaults. Either way the test uncovered leaky tests (`tests/test_create.py:: test_create_env_different_platform_*`). Here we refactor those tests to use a new local package.

We also propose a monkeypatch for the `monkeypatch` fixture where we check whether the monkeypatch set/unset any `CONDA_` prefixed environment variables and if so we undo those changes and call `reset_context`. This should offer a more robust context resetting across all of our tests.

The initial changes resulted in additional leaky behavior being uncovered resulting in necessary updates to `tests/test_exceptions.py::test_print_unexpected_error_message_upload_*`.

And apparently `tests/test_misc.py::test_explicit_missing_cache_entries` wasn't as robust as we previously thought, refactored to use once of our local packages instead of `heapdict`.

### Checklist - did you ...


tests/conftest.py
```diff
@@ -22,6 +24,23 @@
...
+    yield monkeypatch
+
+    # reset context if any CONDA_ variables were set/unset
+    if any(
+        obj is os.environ and name.startswith("CONDA_")
+        for obj, name, _ in monkeypatch._setitem
+    ):
+        log.error("Resetting context")
+        monkeypatch.undo()
+        reset_context()
```
kenodegard, line 42: @jaimergp @travishathaway thoughts on this? this should allow us to stop using the `env_var`/`env_vars` contexts with the callbacks

jaimergp: How stable is `_setitem`? I'm a bit concerned it might break silently. Also, why `log.error` and not just `log.info`? I would also log the detected `CONDA_*` vars just to have some extra info.

tests/conftest.py
```diff
@@ -22,6 +24,23 @@
     "conda.testing.fixtures",
 )
 
+log = getLogger(__name__)
```
jaimergp, line 27: You might want to use a different `__name__`, like `conda.tests` so it uses the same config as everything else.

tests/test-recipes/arch_package/meta.yaml
```diff
@@ -0,0 +1,9 @@
+package:
+  name: arch-package
+  version: 1.0
+
+build:
+  number: 0
+
+about:
+  summary: A small architecture specific pacakge
```
jaimergp, line 9:
```suggestion
  summary: A small architecture specific package
```

tests/test_exceptions.py
jaimergp, line 1: These changes are unrelated to the PR title, right? I would rename the PR for clarity. I guess this is a byproduct of adding the monkeypatch context awareness.

tests/test_create.py
```diff
@@ -2979,38 +2980,48 @@ def test_multiline_run_command():
...
+            raises=DryRunExit,
+        )
+        result = json.loads(stdout)
+        assert result["success"]
+        assert any(
+            pkg["name"] == "arch-package" and pkg["platform"] == platform
+            for pkg in result["actions"]["LINK"]
+        )
+        assert not stderr
+        assert code  # DryRunExit
```
jaimergp, line 3024: I thought DryRunExit had exit code 0, since it inherits from `CondaExitZero`:

https://github.com/conda/conda/blob/4064f528bca7281115cbf07eea82527a14d4c821/conda/__init__.py#L123-L124

tests/conftest.py
```diff
@@ -22,6 +24,23 @@
...
+
+
+@pytest.fixture(name="monkeypatch")
```
jaimergp, line 30: Are we simply shadowing the original `monkeypatch` with this? Can we add this to an importable location in `conda.testing`? This will be useful in other projects like `conda-libmamba-solver`.

jaimergp: Actually we should move all these fixtures to somewhere in `conda.testing`. I had to copy a few over to conda-libmamba-solver to make the new test style work.

tests/test_exceptions.py
kenodegard: combining the two tests caused a reshuffle so these tests started to fail on macOS: https://github.com/conda/conda/actions/runs/7267643889/job/19801909296?pr=13449

tests/test_create.py
kenodegard: we changed `conda_cli` to return the pytest `ExceptionInfo` object: https://github.com/conda/conda/blob/4064f528bca7281115cbf07eea82527a14d4c821/conda/testing/__init__.py#L193-L195

tests/conftest.py
kenodegard: Done, see https://github.com/conda/conda/pull/13449/commits/e00239cc839d810a0b1ecfabea62743d82092224

tests/conftest.py
kenodegard: We got this for free by moving the fixture to `conda.testing`, see https://github.com/conda/conda/pull/13449/commits/e00239cc839d810a0b1ecfabea62743d82092224

tests/conftest.py
kenodegard: > How stable is `_setitem`?

It's an original feature of the `MonkeyPatch` implementation (https://github.com/pytest-dev/pytest/commit/c17a09adafeb8bc36b70f182c904ce7bcfa0e97e), so almost 15 years stable

> why `log.error` and not just `log.info`

this snapshot is from before I updated it to a `log.debug`, see https://github.com/conda/conda/pull/13449/commits/a96508a12bee9576d86bb1875b2277e73b3eb53d

> also log the detected `CONDA_*` vars just to have some extra info

good idea

tests/test_create.py
kenodegard: reworked it to:

```python
assert isinstance(excinfo.value, DryRunExit)
```

tests/test_exceptions.py
kenodegard: updated title and description

beeankha (approved):

conda/testing/__init__.py
```diff
@@ -240,9 +243,6 @@ def conda_cli(capsys: CaptureFixture) -> CondaCLIFixture:
     """Fixture returning CondaCLIFixture instance."""
     yield CondaCLIFixture(capsys)
 
-    # restore to prior state
-    reset_context()
-
```
kenodegard, line 245: removed since it didn't actually reset anything, see new `context_aware_monkeypatch` instead

conda/testing/__init__.py
```diff
@@ -295,7 +295,6 @@ def __call__(
         """
         prefix = Path(prefix or self.path_factory())
 
-        reset_context([prefix / "condarc"])
```
kenodegard, line 298: removed since this is already included as part of invoking the subcommands (see `context.__init__` calls with the parsed arguments)

tests/conftest.py
```diff
@@ -35,6 +35,18 @@ def test_recipes_channel(monkeypatch: MonkeyPatch) -> Path:
...
+    local = Path(__file__).parent / "data" / "legacy_repodata"
+    monkeypatch.setenv("CONDA_BLD_PATH", str(local))
+    monkeypatch.setenv("CONDA_USE_LOCAL", "true")
+    reset_context()
+    assert local.samefile(context.bld_path)
+    assert context.use_local
+
+    return local
+
+
```
kenodegard, line 49: No longer necessary

```suggestion
```

tests/conftest.py
```diff
@@ -17,6 +16,7 @@
 
 pytest_plugins = (
     # Add testing fixtures and internal pytest plugins here
+    "conda.testing",
```
jezdez, line 19: Much better!

tests/test_create.py
```diff
@@ -2979,38 +2980,56 @@ def test_multiline_run_command():
...
+            == platform
+        )
 
+        stdout, stderr, excinfo = conda_cli(
+            "install",
+            f"--prefix={prefix}",
+            "arch-package",
+            "--dry-run",
+            "--json",
+            raises=DryRunExit,
```
jezdez, line 3020: Noice!

jezdez (approved):

PR #13452, opened 2023-12-20
Enable repodata.json.zst by default

### LINKED ISSUE #13256: Enable repodata.json.zst by default

dholth:
### Description

Since `repodata.json.zst` is a performance slam dunk, enable it by default.

Uses the jlap networking code to handle checking for the two possible formats `.json.zst` or `.json`.

The jlap networking code should be read through and cleaned up.

Fix #13256 

### Checklist - did you ...


conda/gateways/repodata/__init__.py
```diff
@@ -106,6 +106,13 @@ def get_repo_interface() -> type[RepoInterface]:
                 f"Is the required jsonpatch package installed?  {e}"
             )
 
+    try:
+        from .jlap.interface import ZstdRepoInterface
+
+        return ZstdRepoInterface
+    except ImportError:  # pragma: no cover
```
dholth, line 113: This fallback (on the jlap interface) was originally in case jsonpatch was not installed. Now this function avoids a circular import. `zstandard` and `jsonpatch` are regular conda dependencies today.

conda/gateways/repodata/jlap/interface.py
```diff
@@ -121,3 +121,28 @@ def repodata_parsed(self, state: dict | RepodataState) -> dict | None:
...
+        return super().should_check_format(format)
+
+
+class ZstdRepoInterface(JlapRepoInterface):
+    """
+    Support repodata.json.zst (if available) without checking .jlap
+    """
+
+    def _repodata_state_copy(self, state: dict | RepodataState):
+        return RepodataStateSkipFormat(dict=state, skip_formats=["jlap"])
```
jezdez, line 148: I'm not super happy that we're implementing the zst repodata API via the JLAP API, even though they are not that related. But I don't see a clear way to implement it independently, either. Do you?

dholth: That would include a significant refactor of [repodata/jlap/fetch](https://github.com/conda/conda/blob/aca84aa1328ca36eab1ab7f6d3956bf0a3d4b7c0/conda/gateways/repodata/jlap/fetch.py#L277-L281) which contains the logic to look for alternate formats and remember whether or not they are available.

jezdez: Gotcha, an opportunity for another time

jezdez (approved):

conda/gateways/repodata/jlap/interface.py
dholth: We also have an opportunity to add a flag or configuration to be able to disable `repodata.json.zst` checks (add both formats to the new skip-formats class), this could be written without knowledge of the networking details.

PR #13460, opened 2023-12-22
Enable performance tracking via continuous benchmarking.

jezdez:
### Description

This sets up codespeed following their docs: https://docs.codspeed.io/ci/github-actions

I've selected a number of basic tests to run for now, we can tweak that list in the future.

### Checklist - did you ...


.github/workflows/tests.yml
```diff
@@ -204,6 +204,37 @@ jobs:
...
+  linux-benchmarks:
+    # only run test suite if there are code changes
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.11', '3.12']
```
jezdez, line 217: Not sure if it makes sense to benchmark different versions of Python here, and if it should be 3.8 and 3.12 TBH. Any thoughts on this?

travishathaway: I would just define a single Python version for now. I don't think it will bring much to run benchmarks for two different versions because we're mostly focused on a relative measurements between conda versions or commits and not between Python versions (although that may be interesting to know).

If we do add different versions, it makes sense to add the latest and oldest supported versions (3.8 and 3.12).

jezdez: Let's make it simpler and just use one version

travishathaway (approved):

tests/cli/test_subcommands.py
```diff
@@ -210,11 +214,11 @@ def test_env_remove(conda_cli: CondaCLIFixture, tmp_env: TmpEnvFixture):
         assert not code
 
 
+@pytest.mark.benchmark
```
jaimergp, line 217: In the future I'd rather see a `tests/benchmarks/` directory with benchmark-focused tests so we don't mix IO too much, since codspeed since to be focusing on CPU bound tasks.

jaimergp (approved):

dev/linux/benchmarks.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -o errtrace -o pipefail -o errexit
+
+### Prevent git safety errors when mounting directories ###
+git config --global --add safe.directory /opt/conda-src
+
+# make sure all test requirements are installed
+sudo /opt/conda/bin/conda install --quiet -y --file tests/requirements.txt "conda-forge::menuinst>=2" --repodata-fn=repodata.json
```
kenodegard, line 9:
```suggestion
sudo /opt/conda/bin/conda install --quiet -y --file tests/requirements.txt --repodata-fn=repodata.json
```

kenodegard (changes requested):

tests/cli/test_subcommands.py
jezdez: Yeah, that makes sense, I even started with it at first and then noticed that I was duplicating a lot of efforts. I started from the tests in conda-benchmarks, but thought to first get some experience with codspeed.

kenodegard (approved):

PR #13465, opened 2023-12-27
Type hints for conda.cli.main_search

samhaese:
### Description

This is a small PR that adds a few docstrings and type hints to `conda.cli.main_search`.  This is part of #13437 

I didn't add any docstring to configure_parser since its use is consistent across all of the main_* cli modules.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #13477, opened 2024-01-04
Fix #13476

### LINKED ISSUE #13476: "Port could not be cast to integer value" when environment variable is missing
### Checklist

### What happened?

I'm getting `"Port could not be cast to integer value"` when I have a URL that contains an environment variable as the token that is not set on the system.

This happens when you use Conda from PyCharm with their Conda integration.

### Conda Info

_No response_

### Conda Config

```shell
channels:
  - https://example.com:1234/t/${MISSING}/get
```

### Conda list

_No response_

### Additional Context

_No response_

jonashaag:
### Description

Resolves #13476

### Checklist - did you ...

### DIFF
--- a/conda/common/url.py
+++ b/conda/common/url.py
@@ -353,7 +353,7 @@ def split_anaconda_token(url):
         >>> split_anaconda_token("https://10.2.3.4:8080/conda/t/tk-123-45")
         (u'https://10.2.3.4:8080/conda', u'tk-123-45')
     """
-    _token_match = re.search(r"/t/([a-zA-Z0-9-]*)", url)
+    _token_match = re.search(r"/t/([a-zA-Z0-9$%{}_-]+)", url)
     token = _token_match.groups()[0] if _token_match else None
     cleaned_url = url.replace("/t/" + token, "", 1) if token is not None else url
     return cleaned_url.rstrip("/"), token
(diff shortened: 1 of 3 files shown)


travishathaway:
Hi @jonashaag,

Thanks for submitting this pull request. I think this is a reasonable change, but I also wonder the context where this being used that's causing a problem. Usually, if you go to the `conda.base.context.context` object and access the `channels` property, these environment variables are replaced so this should never be a problem.

I would be interested to know exactly how to recreate this error in PyCharm.

jonashaag:
PyCharm calls `conda info --envs --json` which crashes with the example configuration

travishathaway:
Thanks for the clarification.

I would be happy to approve these changes once a news file has been added (see https://github.com/conda/conda/pull/13465/files#diff-f0269936f11d1c774728706e82ef9dc19c87966185edd9c021c470354fb90366 for an example).

I'm also going to have one more maintainer look this over.

jezdez (changes requested):
This is a duplicate of #10771, or at least an instance of the same limited token matching. The PR solves this problem by simply matching for any characters, which I think we weren't super comfortable without an improved test coverage.

Could you add a test case for that PR as well and then expand on it?

jonashaag:
What specific example would you like to have tested?

Issue #13437, opened 2023-12-17
Add type hints and/or reST docstrings


samhaese:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

A lot of the conda modules have no type hinting, incomplete type hinting, no docstrings, or docstrings with varying formats (reST, Google).

How open are the maintainers to PR's that add type hints and/or reST-style doc strings?  I'm using `conda.deprecations` as a reference for style.

travishathaway:
Hi @samhaese,

Thanks for creating this. We would be open to any pull request adding more type hinting or consolidating our documentation style. `conda.deprecations` is a good reference to use.

My advice to you if you want to submit such pull requests is to keep them narrow in scope (e.g. focus on one module at a time or limit lines to changed to around 1,000). This makes the pull requests smaller and easier for us to review. We have also begun to render our API documentation here recently:

- https://docs.conda.io/projects/conda/en/stable/dev-guide/api/conda/index.html

Please make sure that it renders correctly there when making any changes.

Let us know if you need any help on this. We are happy to assist.

travishathaway:
One more thing:

It would be nice if you could make a check list of the things that you think should be updated on this issue. That way, others could potentially pick up some of this work. We could also add our own suggestions to this list too.

samhaese:
The task list and new layout of the issue has been added to help identify where people can easily contribute.  I'll do another pass soon and mark modules completed that don't actually need any additional type hints or doc strings.

Maybe this is a "good first issue"?

jezdez:
FWIW, since this came up in #13445, we should make use of sphinx-autoapi's ability to use typehints [automatically to render return types](https://sphinx-autoapi.readthedocs.io/en/latest/how_to.html#how-to-include-type-annotations-as-types-in-rendered-docstrings) and not duplicate the information in the source code.

samhaese:
@Nathann03 I've been keeping each pull request to one module at a time so far.  As you look through the code, find a module you think you can tackle and submit a PR linking to this issue and I'll add it to the list at the top.

jaimergp:
Mentioning these tips here in case someone wants to identify areas of work for this issue in the future. We are slowly adding types here and there!

With `mypy`, you can get a report of "imprecision" using this command:

```
mypy --txt-report report.txt -p conda 
```

With `pyright`, you can leverage this score annotator:

```
pyright --verifytypes conda --ignoreexternal
```

[...]

PR #13480, opened 2024-01-07
Added type hints to conda.base.constants.py

samhaese:
### Description

This small PR that only modifies `conda.base.constants` is part of the effort to add type hints and doc strings across the conda package (#13437).

Notes: 
 * I understand the `base` modules need to be highly performant to I hope these modifications haven't affected that at all.
 * So far, no type hints were added to any Enum subclass.

### Checklist - did you ...


alippai:
I'm not a conda contributor, but I'm curious what's this good for? The function / class typings I understand, but for constants you are changing Literal values to more generic ones. Is this desired? Does mypy, pyling, pyright or other tools encourage the explicit constant/literal annotation?

travishathaway:
@samhaese,

You need to add this line before the other Python imports:

```
```

This should fix the failing tests.

travishathaway:
Hi @alippai,

Thanks for your input. Maybe what we want in this case is to use `Final` instead? (https://docs.python.org/3/library/typing.html?highlight=literal#typing.Final). Basically we want the readers of the code and the type checks to know that most of these values should not be changed.

alippai:
That sounds promising, adding simply the implicit `Final` will make it a `Final[Literal[myvalue]]` in mypy. I hope other tools have the same special handling (instead of `Final[Any]`.

jaimergp:
@conda/conda-maintainers, ready for review.

dholth (dismissed):
Thanks!

kenodegard (approved):

PR #13483, opened 2024-01-08
ensure `PackageRecord.timestamp` is dumped in milliseconds

jaimergp:
### Description

This is needed for https://github.com/conda/conda-libmamba-solver/pull/414, which identifies a bug in how we export synthetic repodata to disk. `PackageRecord.timestamp` returns a `float` in seconds, but the JSON repodata should have an `int` in milliseconds. `TimestampField` has the necessary logic but it's not active by default; we are simply flipping that `in_dump` value so `PackageRecord.dump()` includes it.

### Checklist - did you ...

### DIFF
--- a/conda/testing/helpers.py
+++ b/conda/testing/helpers.py
@@ -220,12 +220,14 @@ def _export_subdir_data_to_repodata(subdir_data: SubdirData):
     packages = {}
     packages_conda = {}
     for pkg in subdir_data.iter_records():
+        if pkg.timestamp:
+            # ensure timestamp is dumped as int in milliseconds
+            # (pkg.timestamp is a kept as a float in seconds)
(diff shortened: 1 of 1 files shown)


conda/testing/helpers.py
```diff
@@ -220,12 +220,14 @@ def _export_subdir_data_to_repodata(subdir_data: SubdirData):
     packages = {}
     packages_conda = {}
     for pkg in subdir_data.iter_records():
+        if pkg.timestamp:
+            # ensure timestamp is dumped as int in milliseconds
+            # (pkg.timestamp is a kept as a float in seconds)
```
dholth, line 225: I bet we use time.time() to get the float timestamp in conda-build as well? Assuming it would be hard to switch to ints everywhere?

dholth (approved):
Thanks

PR #13490, opened 2024-01-10
Fix `context` bug with `CONDA_REPODATA_FNS` and improve some tests in `test_solve`

jaimergp:
### Description

Some tests in `test_solve` were using the cached solver backend, so the solver parametrization was not working properly.

Additionally, we explicitly set `CONDA_REPODATA_FNS` so libmamba can recognize it was asked for (and not injected in `conda.cli.install`). 

While doing this I discovered that the current `conda` will crash if `CONDA_REPODATA_FNS` is set to something _not_ including `repodata.json` because a list is expected but `context` returns a tuple.

### Checklist - did you ...

### DIFF
--- a/tests/core/test_solve.py
+++ b/tests/core/test_solve.py
@@ -3170,10 +3170,10 @@ def test_freeze_deps_1(tmpdir):
 
 def test_current_repodata_usage(tmpdir):
     # force this to False, because otherwise tests fail when run with old conda-build
-    with env_var(
-        "CONDA_USE_ONLY_TAR_BZ2", False, stack_callback=conda_tests_ctxt_mgmt_def_pol
-    ):
-        solver = context.plugin_manager.get_cached_solver_backend()(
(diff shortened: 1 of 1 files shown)


jaimergp:
```
4084.33s call     tests/test_create.py::test_install_features[classic]
```

o.o WHAT

jezdez:
@jaimergp Any idea where the 3.12 test failures comes from?

jaimergp:
I'm looking into it. I can reproduce locally and only with 3.12. Very intriguing. Also the test is _slow_. 4000s in CI, 1000s locally o.o

jaimergp:
I think it's an interaction between `main` and `conda-forge` that makes the classic solver faint :P These integration tests shouldn't assume compatibility with the pre-configured channels when issuing a `--channel=xxx` argument, so I added an `--override-channels` flag and now it passes (and super fast!).

jezdez (approved):

PR #13498, opened 2024-01-12
Create local `track_features`/`features` packages for testing

kenodegard:
### Description

Relying on the `main` or `conda-forge` channels to accurately offer valid `track_features`/`features` packages no longer makes sense since at least `features` is considered deprecated and discouraged.

So instead we create local recipes that we can test against.

### Checklist - did you ...


kenodegard:
@jaimergp any idea why the new `test_remove_features_downgrade` test doesn't work correctly? I don't understand why that one would fail when `test_remove_features_upgrade` succeeds

tests/test-recipes/track_feature_v1/meta.yaml
```diff
@@ -0,0 +1,8 @@
+package:
+  name: track_feature
+  version: 1.0
```
jaimergp, line 3:
```suggestion
  version: "1.0"
```

In case you see something weird in the versioning.

tests/test-recipes/track_feature_v2/meta.yaml
```diff
@@ -0,0 +1,8 @@
+package:
+  name: track_feature
+  version: 2.0
```
jaimergp, line 3:
```suggestion
  version: "2.0"
```
Same.

tests/test-recipes/feature_v1/meta.yaml
```diff
@@ -0,0 +1,12 @@
+package:
+  name: feature
+  version: 1.0
```
jaimergp, line 3:
```suggestion
  version: "1.0"
```

tests/test-recipes/feature_v2/meta.yaml
```diff
@@ -0,0 +1,12 @@
+package:
+  name: feature
+  version: 2.0
```
jaimergp, line 3:
```suggestion
  version: "2.0"
```

conda/core/solve.py
```diff
@@ -625,7 +626,13 @@ def _remove_specs(self, ssc):
...
                 if rec_has_a_feature and rec.name in ssc.specs_from_history_map:
                     spec = ssc.specs_map.get(rec.name, MatchSpec(rec.name))
-                    spec._match_components.pop("features", None)
+                    spec._match_components = frozendict(
+                        {
+                            key: value
+                            for key, value in spec._match_components.items()
+                            if key != "features"
+                        }
+                    )
```
jaimergp, line 635: Why do we need a frozen dict here?

tests/test_create.py
```diff
@@ -1127,38 +1127,6 @@ def test_allow_softlinks(
         assert (prefix / "fonts" / "Inconsolata-Bold.ttf").is_symlink()
 
 
-@pytest.mark.skipif(on_win, reason="nomkl not present on windows")
-def test_remove_features(
```
jaimergp, line 1131: This is now in `test_features` under a new name, right?

tests/test_create.py
```diff
@@ -1351,42 +1319,6 @@ def test_install_update_deps_only_deps_flags(
         assert package_is_installed(prefix, "dependency=2.0")
 
 
-@pytest.mark.skipif(on_win, reason="nomkl not present on windows")
-def test_install_features(
```
jaimergp, line 1355: I assume here too? New name under `test_features`?

tests/test_features.py
```diff
@@ -0,0 +1,88 @@
...
+    # install an older track_feature
+    with tmp_env("track_feature=1.0", "feature") as prefix:
+        assert package_is_installed(prefix, "track_feature=1.0")
+        assert package_is_installed(prefix, "feature=1.0")
+
+        # remove the track_feature1, expecting to see track_feature2 be installed instead
+        conda_cli(
+            "remove", f"--prefix={prefix}", "--features", "track_feature1", "--yes"
+        )
+        assert package_is_installed(prefix, "track_feature=2.0")
```
jaimergp, line 68: Aren't you running into `PrefixData` cache issues? Or do we have an autouse fixture to clear it after `conda_cli` uses?

jaimergp (approved):
Just a few questions, but overall LGTM!

conda/core/solve.py
kenodegard: `spec._match_components` is already a `frozendict`, so the `pop` failed (the new tests exposed this)

tests/test_create.py
kenodegard: Correct, moved it and expanded on it

tests/test_create.py
kenodegard: Correct, moved it and expanded on it

jaimergp:
> @jaimergp any idea why the new `test_remove_features_downgrade` test doesn't work correctly? I don't understand why that one would fail when `test_remove_features_upgrade` succeeds

I took a look locally with:

```
CONDA_VERBOSITY=4 pytest -v tests/test_features.py::test_remove_features_downgrade --capture=tee-sys 2>&1 | less -R
```

Essentially, the `remove --features` command solution is a no-op:

```
DEBUG conda.core.solve:solve_final_state(427): solved prefix /tmp/pytest-of-test_u
ser/pytest-48/test_remove_features_downgrade0/6b920d6709fe478bb1182154003a93b9
  solved_linked_dists:
    local/noarch::feature-2.0-track_feature2_0
    local/noarch::track_feature-2.0-0

INFO conda.core.link:__init__(210): initializing UnlinkLinkTransaction with
  target_prefix: /tmp/pytest-of-test_user/pytest-48/test_remove_features_downgrade
0/6b920d6709fe478bb1182154003a93b9
  unlink_precs:
    
  link_precs:
```

i.e. The solution we get is the same as the starting point. This triggers a false positive in 

https://github.com/conda/conda/blob/c9ceeaea107fdf94c76e86b22dd5799f640998e3/conda/cli/install.py#L441-L445

The exception should have been "Couldn't remove". The track-features package is there but the solver must be buggy and doesn't remove it.

tests/test_features.py
```diff
@@ -0,0 +1,88 @@
...
+):
+    # install a newer track_feature
+    with tmp_env("track_feature=2.0", "feature") as prefix:
+        assert package_is_installed(prefix, "track_feature=2.0")
+        assert package_is_installed(prefix, "feature=2.0")
+
+        # remove the track_feature2, expecting to see track_feature1 be installed instead
+        conda_cli(
+            "remove", f"--prefix={prefix}", "--features", "track_feature2", "--yes"
+        )
```
jaimergp, line 86: I think this is broken behavior. The solver refuses to downgrade.

The equivalent `install` command does work though:

```
conda_cli("install", f"--prefix={prefix}", "track_feature=1.0", "--yes")
```

So maybe we just need to make a note about the broken behavior and move on. It's not like the users are left without options. Hopefully not too many depend on this flag. Which begs the question: should we deprecate `--features` as a flag? It's only available on `conda remove` 🤷

mbargull: > @mbargull do you have any insights here?

Unfortunately not.
I've mostly worked with/around the `features` stuff in the way it's nowadays still used, i.e., with `track_features` used just to "weigh down" versions/builds.
I also know how `features` are (or at least were around 2019 +/- a year or so) handled in `conda.resolve` -- but I didn't much looked into how they were handled in explicit specs in `conda.core.solve` etc. ...

> I'd be happy to mark the failing test as xfail

If you wanted to check if that test case was handled as you expected in older versions, versions `4.3.34`, `4.5.12`, `4.7.12` would probably candidates to test.
In case it isn't already marked as deprecated, I'd certainly suggest to mark the `--features` CLI arg as pending removal.

kenodegard: tested with `conda 4.5.12` and got the same `PackagesNotFoundError` error so perhaps this was never possible

I will mark this as xfail

jezdez (approved):

PR #13504, opened 2024-01-16
add --no-repodata-zst flag

### LINKED ISSUE #13491: Add flag or environment variable to disable checking repodata.json.zst
### Checklist

### What is the idea?

Add a flag or environment variable or config option to disable looking for `repodata.json.zst` in case the remote server returns bad data for `repodata.json.zst` (instead of e.g. a nice 404). Normally `repodata.json.zst` is used if it is available, but if it is not available that is also remembered and not checked again for 7 days.

dholth:
### Description

Fix #13491 

If remote returns valid data, or a normal 404 then `repdodata.json.zst` will provide a nice speedup or a weekly re-check and fallback to `repodata.json`. When a server returns unparseable `repodata.json.zst` or a bad error code, this flag enables the older network code and does not look for `repodata.json.zst` at all.

### Checklist - did you ...

### DIFF
--- a/news/13491-no-repodata-zst-flag
+++ b/news/13491-no-repodata-zst-flag
@@ -1,0 +1,20 @@
+### Enhancements
+
+* Add `--no-repodata-zst` flag to skip `repodata.json.zst` check. Useful if
+  remote returns unparseable `repodata.json.zst` instead of correct data or 404.
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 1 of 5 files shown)


news/13491-no-repodata-zst-flag
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Add `--no-repodata-zst` flag to skip `repodata.json.zst` check. Useful if
+  remote returns unparseable `repodata.json.zst` instead of correct data or 404.
```
beeankha, line 4:
```suggestion
  remote returns unparseable `repodata.json.zst` instead of correct data or 404. (#13504)
```

jezdez (changes requested):
It's worth providing a test that double checks the zst compressed repodata isn't fetched.

dholth:
Yes I will review the coverage tests locally; the new test will be similar to other tests that check the fetcher class when the flag is set.

jezdez (approved):

beeankha (approved):

PR #13510, opened 2024-01-18
Prevent accidental `context` reinitialization after using `make_temp_package_cache`

jaimergp:
### Description

This fixes a tricky issue seen in https://github.com/conda/conda-libmamba-solver/pull/423#issuecomment-1898149457 and then https://github.com/conda/conda-libmamba-solver/pull/426.

It all starts with these errors:

```
FAILED tests/cli/test_main_clean.py::test_clean_tarballs - AssertionError: assert False
 +  where False = has_pkg('zlib', [])
 +    where [] = _get_tars('/var/folders/mm/pltwc2yj1jx192t6dzy9zsrr0000gn/T/1dd3 dcafbe/pkgs')
FAILED tests/cli/test_main_clean.py::test_clean_all[True] - AssertionError: assert False
 +  where False = has_pkg('zlib', [PosixPath('/var/folders/mm/pltwc2yj1jx192t6dzy9zsrr0000gn/T/9650 cbfdae/pkgs/cache')])
FAILED tests/cli/test_main_clean.py::test_clean_all_mock_lstat[True] - AssertionError: assert False
 +  where False = has_pkg('zlib', [PosixPath('/var/folders/mm/pltwc2yj1jx192t6dzy9zsrr0000gn/T/beba ecfdab/pkgs/cache')])
```

In other words, the tests expect to find a certain tarball in those paths. These have been set up as a temporary package cache location. It should be the _only_ path in `context.pkgs_dirs`, `make_temp_package_cache` had ensured that the context object was reset with an empty `search_path=()` to ignore the user configuration. In fact, there's an assertion to make sure that's the case:

https://github.com/conda/conda/blob/f825c702c4a1f915cbe5aaef627ed349c4205826/conda/testing/integration.py#L344

However, when `tmp_env` and `conda_cli` are used right after, they end up calling `conda.cli.main.main_subshell`, which mercilessly resets the context with the default search path. Since some OSs in our CI uses setup-miniconda, which sets up `pkgs_dirs` to `~/conda_pkgs_dirs` in `~/.condarc`, this ends up leaking back into the tests.

In some cases, `ProgressiveFetchExtract` ends up choosing `~/conda_pkgs_dirs` instead of the temporary location, and the test assertion fails.

My proposed fix is to expose `search_path` privately in `conda.cli.main.main_subshell` so then it can be invoked properly from `conda_cli` and friends. However, this is a bit too verbose in the tests. There might be a better way?

### Checklist - did you ...

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -44,7 +44,7 @@ def generate_parser(*args, **kwargs):
     return generate_parser(*args, **kwargs)
 
 
-def main_subshell(*args, post_parse_hook=None, **kwargs):
+def main_subshell(*args, post_parse_hook=None, _context_search_path=None, **kwargs):
     """Entrypoint for the "subshell" invocation of CLI interface. E.g. `conda create`."""
     # defer import here so it doesn't hit the 'conda shell.*' subcommands paths
     from ..base.context import context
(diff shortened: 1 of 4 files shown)


kenodegard:
I alternatively propose: https://github.com/conda/conda/pull/13162/commits/da8a885fe8dd3f5fa790058a7b17688e79a6cddd

kenodegard:
Closing in favor of #13511

PR #13518, opened 2024-01-23
Move / update more `test_create.py` tests

beeankha:
### Description

More work related to https://github.com/conda/conda/issues/13292. The following tests will be updated:

- `test_packages_not_found`
- `test_conda_pip_interop_dependency_satisfied_by_pip`
- `test_install_freezes_env_by_default`
- `test_create_from_extracted`
- `test_install_mkdir`


tests/test_install.py
```diff
@@ -207,3 +213,150 @@ def test_read_no_link(tmpdir):
...
+            "--yes",
+        )
+
+        stdout, _, _ = conda_cli("list", f"--prefix={prefix}", "--json")
+        pkgs_after_install = json_loads(stdout)
+
+        for pkg in pkgs:
+            for pkg_after in pkgs_after_install:
+                if pkg["name"] == pkg_after["name"]:
+                    assert pkg["version"] == pkg_after["version"]
```
kenodegard, line 246:
```suggestion
        PrefixData._cache_.clear()
        prefix_data = PrefixData(prefix)
        for pkg in pkgs:
            assert prefix_data.get(pkg["name"]).version == pkg["version"]
```

tests/test_install.py
```diff
@@ -207,3 +213,150 @@ def test_read_no_link(tmpdir):
...
+            assert isdir(dir)
+            assert exists(file)
+            with pytest.raises(DirectoryNotACondaEnvironmentError):
+                stdout, _, _ = conda_cli(
+                    "install", f"--prefix={dir}", "python", "--mkdir", "--yes"
+                )
+                assert (
+                    "The target directory exists, but it is not a conda environment."
+                    in stdout
+                )
```
kenodegard, line 264: Ditto

```suggestion
            with pytest.raises(
                DirectoryNotACondaEnvironmentError,
                match="The target directory exists, but it is not a conda environment.",
            ):
                conda_cli("install", f"--prefix={dir}", "python", "--mkdir", "--yes")
```

tests/test_install.py
```diff
@@ -207,3 +213,150 @@ def test_read_no_link(tmpdir):
...
+                "python-dateutil",
+                "python",
+                "--mkdir",
+                "--yes",
+            )
+            assert package_is_installed(dir, "python")
+            assert package_is_installed(dir, "python-dateutil")
+
+    finally:
+        rm_rf(prefix, clean_empty_parents=True)
```
kenodegard, line 286: a try-finally with `rm_rf(prefix)` is no longer required since we use `tmp_env` which uses our `path_factory` fixture which in turn uses the `tmp_path` fixture which automatically cleans up the temporary directory upon test completion

kenodegard (changes requested):

kenodegard (approved):

Issue #13552, opened 2024-01-31
Allow plugins to define arbitrary configuration parameters


travishathaway:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

We want conda plugins to be able to define their own configuration parameters.

### Why is this needed?

Certain plugins may need to collect configuration that is specific to that particular plugin. We want to allow this and not have to hard code these configuration parameters in conda itself.

### What should happen?

A while back, I worked on an implementation that involved modifying the `conda.base.context.Context` class at runtime:

- https://github.com/conda/conda/pull/11971/files#diff-423030a00e56775c2e10d719b1bd322be14c99c150210954d344938a98c0a83bR54

I wrote some suggestions in that comment that should be followed:

- Make sure that existing configuration parameters cannot be overridden
- Ideally, users will just need to define and return the name of the configuration parameter and the `ParameterLoader` class with a parameter definition inside

PR #13554, opened 2024-02-01
Add ability for plugins to define settings

travishathaway:
### Description

Fixes: #13552

### Example

Here is how this plugin is used:

#### Plugin code

Use the `conda_settings` plugin hook to return a `CondaSetting` object containing a `name`, `loader` and `description`. Plugin authors are able define new settings exactly as they are currently defined for the `Context` object. See that class definition for more examples to see what all is possible when defining new settings.

```python
from conda.common.configuration import ParameterLoader, PrimitiveParameter
from conda.plugins import hookimpl, CondaSetting

@hookimpl
def conda_settings():
    name = "param_one"
    loader = ParameterLoader(
        PrimitiveParameter("", element_type=str)
    )
    description = "param one description"

    yield CondaSetting(
        name=name,
        loader=loader,
        description=description
    )
```

#### Accessing the settings via `context`

New plugin settings are available under `context.plugins`.

```python

context.plugins.param_one
```

#### Using the setting (end user experience)

Either define in a config file:

```yaml
plugins:
  param_one: some_value
```

Or via environment variables:

```bash
export CONDA_PLUGINS_PARAM_ONE='some_value'
```

### Checklist - did you ...


conda/cli/main.py
```diff
@@ -63,6 +65,12 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
         "verbosity": pre_args.verbosity,
     }
 
+    try:
+        # load extra parameters defined by plugins
+        context.plugin_manager.load_configuration_parameters()
+    except PluginError as exc:
+        logging.warning(exc.message)
```
travishathaway, line 72: Not really sure what to do here about the logging call. Below is a call to `init_loggers` but this method really needs to be called as soon as possible in this function.

Ideally, I would move the `init_loggers` call to the top and then create a `logger` object inside this function to properly log. Maybe that's not necessary though 🤷. I would appreciate some feedback here on this.

conda/plugins/types.py
```diff
@@ -184,3 +185,22 @@ class CondaPostSolve:
...
     action: Callable[[str, tuple[PackageRecord, ...], tuple[PackageRecord, ...]], None]
+
+
+@dataclass
+class CondaConfigurationParameter:
+    """
+    Return type to use when defining a conda configuration parameter plugin hook.
+
+    For details on how this is used, see
+    :meth:`~conda.plugins.hookspec.CondaSpecs.conda_post_solves`.
```
jaimergp, line 196: I think this reference points to a different thing.

tests/plugins/test_configurtaion_parameters.py
jaimergp, line 1: The file name should be `test_configuration_parameters.py`

jaimergp:
Am I understanding correctly that plugin authors would be able to _inject_ toplevel attributes to the `context`? Or will they get their own namespace? I'd advocate for the latter, so folks access their settings via: `context.plugins.conda_libmamba_solver.xxxx` instead of `context.xxxx`. I'm adding `.plugins` to the path so there's no risk of someone claiming a namespace of a top-level attribute name we might need in the future.

conda/plugins/types.py
travishathaway: ahh, left over from copy and paste. Thanks for catching it.

travishathaway:
@jaimergp,

That makes sense and is something I had considered doing. We could actually have a cleaner implementation that way to as we would avoid having to modify the `Context` class with `setattr`. I'll mess around with this today and see what I come up with.

conda/cli/main.py
jezdez: Plugins implementing this functionality will require similar access to a fully initialized logging system, so we don't accidentally allow them to register the default output. This is a sign of a tightly coupled system sadly, so the solution might be to make context initialization lazier and move to a later stage. I'm afraid this effort may require some refactoring in the config system essentially.

conda/plugins/hookspec.py
```diff
@@ -301,3 +301,35 @@ def conda_post_solves():
...
+           from conda import plugins
+           from conda.common.configuration import (
+               ParameterLoader,
+               PrimitiveParameter,
+               SequenceParameter,
+           )
+
+
+           @plugins.hookimpl
+           def conda_post_solves():
```
jezdez, line 325:
```suggestion
           def conda_configuration_parameters():
```

conda/plugins/hookspec.py
```diff
@@ -301,3 +301,35 @@ def conda_post_solves():
                    action=example_post_solve,
                )
         """
+
+    @_hookspec
+    def conda_configuration_parameters(self) -> Iterable[CondaPostSolve]:
```
jezdez, line 306: Also needs an import

```suggestion
    def conda_configuration_parameters(self) -> Iterable[CondaConfigurationParameter]:
```

conda/plugins/manager.py
```diff
@@ -361,6 +378,33 @@ def invoke_post_solves(
...
 
+    def load_configuration_parameters(self) -> None:
+        """
+        Iterates through all registered configuration parameters and adds them to the context class.
+
+        :raises PluginError: Happens when plugins attempt to override existing parameters
+
+        TODO:
+            - Need to add configuration parameters to ``Context.category_map`` and
+              ``Context.description_map``
```
jezdez, line 389: Could these because dynamic properties?

conda/plugins/manager.py
```diff
@@ -361,6 +378,33 @@ def invoke_post_solves(
...
+        """
+        existing_parameters = context.list_parameters()
+        error_params = []
+
+        for name, loader in self.get_configuration_parameters().items():
+            if name in existing_parameters:
+                error_params.append(name)
+            name = loader._set_name(name)
+            setattr(Context, name, loader)
+            setattr(Context, "parameter_names", Context.parameter_names + (name,))
```
jezdez, line 399: This is a bit of code smell that we need to overcome with a context API, it's not great we have to monkeypatch `Context` like this.

conda/plugins/hookspec.py
```diff
@@ -301,3 +301,35 @@ def conda_post_solves():
...
+           )
+
+
+           @plugins.hookimpl
+           def conda_post_solves():
+               name = "example_option"
+               description = "This is an example option"
+               loader = ParameterLoader(
+                   PrimitiveParameter(name, element_type=str), aliases=(name,)
+               )
```
jezdez, line 330: I think it might be useful to construct the loader instance later in the process and have a `loader` (a callable) _and_ a `parameter` argument (a `Parameter` instance), where the loader argument defaults to `ParameterLoader`.

jezdez (commented):
Good direction, even though "conda configurations parameter" is a bit of a mouthful, I don't have a good different suggestion. Maybe "config parameter"?

conda/plugins/manager.py
jezdez: E.g. in the Context class (not sure if this exists already)
```python
def add_parameter(self, name, loader):
    setattr(self, name, loader)
    self.parameter_names += (name,)
```

But maybe there is more to do in the way parameter names are handled, that would not require setting instance (or class) attributes like this

conda/plugins/manager.py
travishathaway: I think they already are by using the `@property` decorator.

We just need to implement a proper setters for these properties, so we can update them.

conda/plugins/manager.py
travishathaway: I agree that there needs to be an API on Context for this.

I think this might have to be a `@classmethod` though. I remember it not working when I tried to set it on the instance itself.

jaimergp:
> Another random thought: maybe it makes sense to just automatically give plugins their own namespace?

I'm not sure how I feel about that one. I think it makes sense, but what if we have a package publishing several plugins? Is it a namespace per plugin or per package? For example, I could have a `conda explain` subcommand that runs the solver and explains things, but the same package is also providing a `post-solve` plugin to inject that behaviour in `conda install`. Two plugin namespaces? I think it makes sense to have it per package, but I don't think we are tracking provenance right now, so maybe that info is not available?

travishathaway:
@jaimergp,

I think you highlighted that this is potentially a can of worms, and I am also not excited about the technical implementation. Like I said, it's just going to complicate the code even further.

Let's just isolate plugin settings to the `plugins` namespace and let plugins fight over spots there. We may wish to encourage naming practices such as `<your-plugin>_<your-setting-name>`.

jaimergp:
Maybe we do want to encourage "at least one namespace per package", instead of plain values in the top-level. The plugin API for the config object would only allow to return a dictionary, not scalars. That way authors are forced to use it.

And another question: what's the story for environment variables overrides? `CONDA_PLUGINS_<NAMESPACE>__<SETTING>`?

travishathaway:
@jaimergp,

The environment variables work like you would expect:

`CONDA_PLUGINS_<SETTING_NAME>`

See this test for an example:

https://github.com/travishathaway/conda/blob/694906f2e3354a22ed1e54e094c149a33ff65602/tests/plugins/test_settings.py#L111-L127

If we want to do what you proposed regarding the mapping, I think the best way to do that would be to have the plugin authors themselves define a namespace and then nest the settings underneath that.

Here's what it might look like:

```python

@hookimpl
def conda_settings():
    return CondaSetting(
        namespace="conda_plugin_one",
        name="option_one",
        loader=loader,
        description="option one"
    )
```

This would then be accessible on the context object like this:

```python
>>> context.plugins.conda_plugin_one.option_one
"some value"
```

Getting this to work appropriately is going to be a little tricky because of the way that the configuration code works. We basically have to dynamically generate new class objects and then attach these new settings all at runtime. This is definitely doable but comes with an added complexity cost that I would personally prefer avoiding.

If we had a configuration system that handled dynamically defining recursive structures like this, it wouldn't be so bad 😅. I'm open to suggestions though if I have some how gotten this wrong and there is an easier way to do it.

conda/plugins/hookspec.py
travishathaway: I do agree that having to provide the `ParameterLoader` object feels a little boiler-platy, but I'm not sure it's bad enough to warrant wrapping it ourselves and adding another layer of abstraction. I'm guessing that most people who end up using this will be looking at the `Context` class to see examples. It would be nice for them to be able to more-or-less copy and paste stuff from there and have it work.

If we wrapped it in another layer of abstraction, it also might be harder for authors to fully understand what's going on.

tests/plugins/test_settings.py
```diff
@@ -0,0 +1,127 @@
+# Copyright (C) 2012 Anaconda, Inc
```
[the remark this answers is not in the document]
kenodegard: Copyrights are auto inserted by pre-commit

conda/plugins/manager.py
```diff
@@ -19,6 +19,7 @@
 
 from ..auxlib.ish import dals
 from ..base.context import context
+from ..common.configuration import PluginConfig
```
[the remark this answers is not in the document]
kenodegard: We fully switched to relative imports back in November https://github.com/conda/conda/pull/13281

It's just a style preference

conda/base/context.py
```diff
@@ -489,6 +496,11 @@ def __init__(self, search_path=None, argparse_args=None, **kwargs):
         self._set_env_vars(APP_NAME)
         self._set_argparse_args(argparse_args)
 
+        # set plugin configuration parameters
+        if initialize_plugin_config:
+            self.plugin_manager.load_settings()
+            self.plugins = PluginConfig(context.raw_data)
+
```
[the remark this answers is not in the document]
travishathaway: I wouldn't be opposed to this. I had to do this because I was running into lots of circular import errors.

We could even do something like the following to lazy load the `plugins` property itself:

```python
def __init__(self, search_path=None, argparse_args=None, plugin_config=None):
    # snip, snip...

    # This is just here in case we ever want to override what `plugins` is on object instantiation
    if plugin_config is not None:
        self._plugins = plugin_config


@property
def plugins():
    if not self._plugins:
        self.plugins = PluginConfig(context.raw_data)
    return self._plugins
```

conda/common/configuration.py
```diff
@@ -1609,3 +1609,69 @@ def typify_parameter(self, parameter_name, value, source):
...
+    to :class:`PluginConfig`.
+    """
+    new_data = {}
+
+    for source, config in data.items():
+        if plugin_data := config.get("plugins"):
+            for param_name, raw_param in plugin_data.value(None).items():
+                new_data[source] = {param_name: raw_param}
+
+        if source == "envvars":
```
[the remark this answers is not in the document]
travishathaway: I don't think so because there are no error conditions in this function. Here, we simply walk through the dictionary that's provided to try and pluck out what we want for the return dictionary. If don't find anything, we simply return an empty dictionary.

conda/common/configuration.py
```diff
@@ -1609,3 +1609,69 @@ def typify_parameter(self, parameter_name, value, source):
...
         raise NotImplementedError()
+
+
+def get_plugin_config_data(
+    data: dict[Path, dict[str, RawParameter]],
+) -> dict[Path, dict[str, RawParameter]]:
+    """
+    This is used to move everything under the key "plugins" from the provided dictionary
+    to the top level of the returned dictionary. The returned dictionary is then passed
+    to :class:`PluginConfig`.
```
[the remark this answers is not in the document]
travishathaway: That's thankfully already handled for us via auto-api:

https://continuumio-conda-docs--13554.com.readthedocs.build/projects/conda/en/13554/dev-guide/api/conda/common/configuration/index.html#conda.common.configuration.get_plugin_config_data

We could add comments for the parameters themselves, but I think that would be a little overkill for this function. I also feel the type hint is detailed enough.

conda/plugins/types.py
```diff
@@ -187,3 +188,22 @@ class CondaPostSolve:
...
+
+    :param name: name of the setting (e.g., ``config_param``)
+    :param description: description of the setting that should be targeted
+                        towards users of the plugin
+    :param loader: ParameterLoader class containing the setting definition
+    """
+
+    name: str
+    description: str
+    loader: ParameterLoader
```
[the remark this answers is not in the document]
travishathaway: We've discussed this before and deemed that unnecessary. This project has a solid deprecation policy, so we would rely on that for retiring properties.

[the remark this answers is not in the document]
travishathaway: See https://github.com/conda-incubator/ceps/blob/main/cep-9.md for more information.

tests/plugins/test_settings.py
```diff
@@ -0,0 +1,127 @@
...
+    PrimitiveParameter,
+    YamlRawParameter,
+    yaml_round_trip_load,
+)
+from conda.plugins.manager import CondaPluginManager
+
+log = logging.getLogger(__name__)
+
+#: Name for a string type parameter
+STRING_PARAMETER_NAME = "string_parameter"
```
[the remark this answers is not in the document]
travishathaway: That's fair, but I personally think that's a little too much boilerplate code for tests. On public API, yes, that totally makes sense.

tests/plugins/test_settings.py
```diff
@@ -0,0 +1,127 @@
...
+                "testdata", yaml_round_trip_load(CONDARC_TEST_ONE)
+            )
+        }
+    )
+
+    context.plugins = PluginConfig(context.raw_data)
+    return setting_plugin_manager
+
+
+def test_get_settings(setting_plugin_manager):
```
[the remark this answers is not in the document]
travishathaway: My goal with these tests is to reach the minimum amount of coverage. We already test the configuration more extensively in testing modules like:

- https://github.com/conda/conda/blob/main/tests/base/test_context.py

so I think it is safe to keep the testing coverage lighter here and only focus on testing what has been added rather than various permutations of what a settings file could be.

conda/plugins/manager.py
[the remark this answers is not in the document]
schuylermartin45: I've optimized C/C++ headers to get build times down before but have never heard of optimizing Python imports. That's kind of terrifying.

conda/common/configuration.py
```diff
@@ -1609,3 +1609,69 @@ def typify_parameter(self, parameter_name, value, source):
...
+    is called in :meth:`conda.plugins.manager.CondaPluginManager.load_settings`.
+    We do this because ``CondaPluginManager`` has access to all registered plugin hooks.
+    """
+
+    def __init__(self, data):
+        self._cache_ = {}
+        self.raw_data = get_plugin_config_data(data)
+
+    @classmethod
+    def add_setting(cls, name: str, loader: ParameterLoader) -> None:
```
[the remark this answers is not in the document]
travishathaway: I don't think that will be possible because we need the `PluginManager` to be able to modify the settings for plugins. This method provides the interface for that and making it private would block that.

I also think that exposing these methods is not ideal. One potential way around this is to dynamically generate the type via a function that the `PluginManager` could use. Here's what that might look like:

```python

class PluginConfigBase(metaclass=ConfigurationType):
    """This is included to be able to define the metaclass"""
    pass

def create_plugin_config_type(plugin_settings: list[CondaSetting]) -> type:
    def __init__(self, data):
        self._cache_ = {}
        self.raw_data = get_plugin_config_data(data)

    plugin_config_type = type("PluginConfig", (PluginConfigBase), {
        "__init__": __init__,
    })

    for plugin_setting in plugin_settings:
        setattr(plugin_config_type, plugin.name, plugin.loader)

    return plugin_config_type
```

Using this method means no more special `add_settings` or `remove_all_settings` methods because the `PluginConfig` type will now be seen as "immutable" If any changes need to be made, we simply generate a new type and throw away the old one. It does look a little weird though because it involves dynamic type generation.

I think I like this method more though. Would be interested to hear your thoughts.

conda/common/configuration.py
```diff
@@ -1609,3 +1609,54 @@ def typify_parameter(self, parameter_name, value, source):
...
+            "__init__": __init__,
+        },
+    )
+
+    for name, loader in plugin_settings.items():
+        plugin_config_type.parameter_names += (name,)
+        name = loader._set_name(name)
+        setattr(plugin_config_type, name, loader)
+
+    return plugin_config_type
```
jezdez, line 1662: Please no meta programming like this, it's going to make things much harder later, it's one of those cases where being clever isn't worth it

conda/base/context.py
```diff
@@ -1154,6 +1162,17 @@ def cpu_flags(self):
...
+    def plugins(self):
+        """Holds plugin related configuration settings"""
+        if self._plugins is None:
+            plugin_config_type = create_plugin_config_type(
+                self.plugin_manager.get_settings()
+            )
+            self._plugins = plugin_config_type(self.raw_data)
+
+        return self._plugins
+
```
jezdez, line 1175: Dynamically creating types are a red flag in the year 2024, let's not do that. If you want to lazily load the plugins, just use `functools.cached_property`.

```suggestion
	from functools import cached_property

    @cached_property
    def plugins(self):
        """Holds plugin related configuration settings"""
        PluginType = create_plugin_config_type(self.plugin_manager.get_settings())
        return PluginType(self.raw_data)
```

conda/plugins/hookspec.py
jezdez: > It would be nice for them to be able to more-or-less copy and paste stuff from there and have it work.
I don't think they'll copy-paste from conda and want an easier API, will likely not want to override the used loader at all.

E.g. this would be easier:

```
       @plugins.hookimpl
       def conda_settings():
           name = "example_option"
           description = "This is an example option"
           yield plugins.CondaSetting(
               name=name,
               description=description,
               parameter=PrimitiveParameter(name, element_type=str), aliases=(name,)
           )
```

news/13554-add-plugin-hook-for-configuration
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Adds a new plugin hook giving plugin authors the ability to define new configuration parameters (#13554)
```
jezdez, line 3:
```suggestion
* Adds a new plugin hook giving plugin authors the ability to define new conda settings. (#13554)
```

conda/common/configuration.py
```diff
@@ -1609,3 +1609,54 @@ def typify_parameter(self, parameter_name, value, source):
...
+    """
+    This is used to move everything under the key "plugins" from the provided dictionary
+    to the top level of the returned dictionary. The returned dictionary is then passed
+    to :class:`PluginConfig`.
+    """
+    new_data = {}
+
+    for source, config in data.items():
+        if plugin_data := config.get("plugins"):
+            for param_name, raw_param in plugin_data.value(None).items():
```
jezdez, line 1626: What does `plugin_data.value(None)` do? Any way to have a nicer API that explains what that does?

jezdez (changes requested):
I like the "settings" name better for this plugin hook, since "configuration" refers to the machinery that handles settings, nice change!

I believe that we shouldn't dynamically create plugin classes though, it indicates that the abstraction isn't following composition but just expands on the antipatterns of the configuration system. So short of requiring a refactor of the config system (e.g. move to attrs or pydantic), let's find another way.

conda/common/configuration.py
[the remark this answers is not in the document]
jezdez: Just adding that I don't agree that dynamically creating classes is superior to the class methods, just that the class methods should not be bunched up with the configuration system like that. Just move them into a separate in-between tool class like `PluginConfiguration` with `add` and `remove` steps, or something, that allows us glue the config and plugin systems together.

conda/common/configuration.py
travishathaway: I looked at this closer. This assumes that we always get a dictionary back from the `.value()`, but that assumption is naive.

We need to add an extra check to make sure that the value we get back from that call some sort of mapping. If it isn't, we should skip it.

travishathaway:
@jezdez, @schuylermartin45 and @kenodegard this is ready for another round of review.

conda/base/context.py
```diff
@@ -1858,6 +1873,10 @@ def description_map(self):
 
 def reset_context(search_path=SEARCH_PATH, argparse_args=None):
     global context
+
+    # reset plugin config params
+    remove_all_plugin_settings()
+
     context.__init__(search_path, argparse_args)
     context.__dict__.pop("_Context__conda_build", None)
     from ..models.channel import Channel
```
schuylermartin45, line 1882: I know this isn't part of this PR, but I see the circular dependency comment below and I'm wondering if we should look into improving that issue. At the very least, the comment about importing to avoid a circular dependency is a few lines after the import is done, and that seems a little confusing/easy to miss.

conda/base/context.py
```diff
@@ -478,7 +480,12 @@ class Context(Configuration):
 
     no_plugins = ParameterLoader(PrimitiveParameter(NO_PLUGINS))
 
-    def __init__(self, search_path=None, argparse_args=None, **kwargs):
+    def __init__(
+        self,
+        search_path=None,
+        argparse_args=None,
+        **kwargs,
+    ):
```
kenodegard, line 488: Since theres no long a change here we should avoid reformatting

```suggestion
    def __init__(self, search_path=None, argparse_args=None, **kwargs):
```

kenodegard, line 488: Since there's no longer a change here we should avoid reformatting

```suggestion
    def __init__(self, search_path=None, argparse_args=None, **kwargs):
```

conda/base/context.py
```diff
@@ -2109,6 +2128,7 @@ def _first_writable_envs_dir():
 
 try:
     context = Context((), None)
+
```
kenodegard, line 2131:
```suggestion
```

conda/common/configuration.py
```diff
@@ -1609,3 +1609,76 @@ def typify_parameter(self, parameter_name, value, source):
...
+    to the top level of the returned dictionary. The returned dictionary is then passed
+    to :class:`PluginConfig`.
+    """
+    new_data = defaultdict(dict)
+
+    for source, config in data.items():
+        if plugin_data := config.get("plugins"):
+            plugin_data_value = plugin_data.value(None)
+
+            if not isinstance(plugin_data_value, frozendict):
```
kenodegard, line 1628: Would it be safer to check for `dict`?

conda/common/configuration.py
```diff
@@ -1609,3 +1609,76 @@ def typify_parameter(self, parameter_name, value, source):
...
+                continue
+
+            for param_name, raw_param in plugin_data_value.items():
+                new_data[source][param_name] = raw_param
+
+        if source == EnvRawParameter.source:
```
kenodegard, line 1634: We would never expect both `config.get("plugins")` and `source == EnvRawParameter.source` to be true for the same source (right?), so we should combine these for clarity:

```
        if plugin_data := config.get("plugins"):
            ...
        elif source == EnvRawParameter.source:
            ...
```

conda/plugins/types.py
```diff
@@ -187,3 +188,23 @@ class CondaPostSolve:
...
+    """
+    Return type to use when defining a conda setting plugin hook.
+
+    For details on how this is used, see
+    :meth:`~conda.plugins.hookspec.CondaSpecs.conda_settings`.
+
+    :param name: name of the setting (e.g., ``config_param``)
+    :param description: description of the setting that should be targeted
+                        towards users of the plugin
+    :param loader: ParameterLoader class containing the setting definition
```
jezdez, line 204:
```suggestion
    :param parameter: Parameter instance containing the setting definition
    :param aliases: alternative names of the setting
```

jezdez (approved):
Minor issues around tests like found by @schuylermartin45 and docs, but generally really looking good

tests/common/test_configuration.py
```diff
@@ -841,3 +842,78 @@ def test_expand_search_path(tmp_path):
...
+    with condarc.open("w") as fp:
+        fp.write(
+            dals(
+                """
+            plugins:
+              option_one: value_one
+              option_two: value_two
+        """
+            )
+        )
```
kenodegard, line 862:
```suggestion
    condarc.write_text(
        dals(
            """
            plugins:
              option_one: value_one
              option_two: value_two
            """
        )
    )
```

tests/plugins/test_settings.py
```diff
@@ -0,0 +1,135 @@
...
+    it is read correctly.
+    """
+    monkeypatch.setenv(
+        f"CONDA_PLUGINS_{STRING_PARAMETER_NAME.upper()}", STRING_PARAMETER_ENV_VAR_VALUE
+    )
+    reset_context()
+
+    try:
+        del context.plugins
+    except AttributeError:
```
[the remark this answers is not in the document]
kenodegard: Actually `del context.plugins` will always raise an `AttributeError` and never do anything since `plugins` is a read-only `property`

kenodegard (changes requested):

tests/plugins/test_settings.py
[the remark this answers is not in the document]
travishathaway: @schuylermartin45 @kenodegard,

"plugins" is a cached property and the according the Python documentation, deleting it is the way to clear the cache:

> The cached value can be cleared by deleting the attribute. This allows the cached_property method to run again.

(source: https://docs.python.org/3/library/functools.html#functools.cached_property)

That is wrapped in try/except block because sometimes this cache has not been set yet (i.e. the property still hasn't been accessed). Albeit, it looks a little ugly in the tests, but that was the best solution I could find to ensure the tests pass every time.

conda/base/context.py
travishathaway: @schuylermartin45,

Figuring out how to rework the organization of the code in conda is a right of passage for most new developers on the project. I encourage you to check out my [Honda project](https://github.com/travishathaway/honda) for my own attempt at doing so.

Please use your fresh eyes and energy to help move the project forward!

conda/base/context.py
travishathaway: Agreed. Surprise the code formatter didn't catch this.

conda/common/configuration.py
travishathaway: We actually want `collections.abc.Mapping` instead. Checking for `dict` here will fail.

tests/plugins/test_settings.py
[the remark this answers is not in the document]
schuylermartin45: From the same docs:

> The cached_property does not prevent a possible race condition in multi-threaded usage. The getter function could run more than once on the same instance, with the latest run setting the cached value. If the cached property is idempotent or otherwise not harmful to run more than once on an instance, this is fine. If synchronization is needed, implement the necessary locking inside the decorated getter function or around the cached property access.

So there's a race condition. Ideally our tests should be deterministic. I think we have a couple of options:
1) Make the cache thread safe with a synchronization mechanism
2) Use `@property` (I'm guessing that's a non-starter and there's a reason you're using the cached version)
3) Mock the cache. I'm only okay with this if we expect that the only way this property is being used in a multi-threaded way is during the unit testing. Otherwise the race condition will crop up under normal use.

conda/base/context.py
schuylermartin45: Lol good name on that project. Like I said, at the very least, I think the comment should be bumped up to be above the import so the next time someone goes _WTF_ it'll be a lot easier to find :)

tests/plugins/test_settings.py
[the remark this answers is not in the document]
schuylermartin45: Similar to 3 I suppose you could also guarantee initialization order of the cache with a fixture.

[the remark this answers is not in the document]
kenodegard: oh wow I did not know that `functools.cached_property` had implemented yet another way to clear the cache, I was expecting the same `cache_clear()` that `functools.lru_cache` has

I'd be in favor of moving the try-except into a fixture, I also think we should be clearing the `context.plugins` value for both setup and teardown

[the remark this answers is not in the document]
travishathaway: @kenodegard, put this in a fixture with some more documentation.

@schuylermartin45, as we discussed, we are not worried at all about this running in a multi-threaded environment 😱. The `context` object is not thread safe!

schuylermartin45 (approved):
Approving as my comments have been addressed. Thanks!

conda/common/configuration.py
kenodegard, line 1: I think all of the new code added here to `conda.common.configuration` actually belongs in `conda.base.context`.

I view configuration as being the abstract implementation and the plugins being a concrete implementation, the `PluginConfig` docstrings alludes to this as well:

> ```
> The object created by this class should only be accessed via
> :class:`conda.base.context.Context.plugins`.
> ```

kenodegard (changes requested):

conda/common/configuration.py
travishathaway: Makes sense to me. I will move it over.

travishathaway:
@kenodegard, this is ready for another review

jezdez (approved):

kenodegard (approved):

jezdez:
Huzzah!

PR #13556, opened 2024-02-01
Fix conda pre, post and health check plugin examples

jjhelmus:
The examples yield from plugins.Conda<plugin> rather than using Conda<plugin> directly.  

### Description

Fix the conda pre, post and health check plugin example so that they correctly use the classes from the `plugins` module.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release 

### DIFF
--- a/conda/plugins/hookspec.py
+++ b/conda/plugins/hookspec.py
@@ -143,7 +143,7 @@ class CondaSpecs:
 
            @plugins.hookimpl
            def conda_pre_commands():
-               yield CondaPreCommand(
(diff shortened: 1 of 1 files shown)


jjhelmus:
Given this change is minimal, does it need a news entry?

kenodegard (approved):
@jjhelmus good catch! No need for a news entry

jezdez (approved):

PR #13557, opened 2024-02-02
Deprecate conda.models.dist and conda.plan

mbargull:
### Checklist - did you ...

Following the removal of `conda.models.dist` and `conda.plan`'s usage in `conda-build` via https://github.com/conda/conda-build/pull/5074 , this moves the few functions of these modules that are used within `conda` itself to other modules and marks both modules as pending for deprecation.


conda/core/link.py
```diff
@@ -1614,3 +1622,61 @@ def messages(prefix):
                 return m
     finally:
         rm_rf(path)
+
+
+def _get_best_prec_match(precs):
```
mbargull, line 1627: We may want to change the name to not carry the underscore prefix since it's used in an other module (`conda.misc` also imported `.plan._get_best_prec_match` before).

conda/models/dist.py
```diff
@@ -17,12 +17,19 @@
...
+from ..history import dist_str_to_quad, strip_extension  # noqa: F401
 from .channel import Channel
 from .package_info import PackageInfo
 from .records import PackageRecord
 
 log = getLogger(__name__)
 
+deprecated.module(
+    "24.3",
+    "24.9",
```
mbargull, line 29: This is the earliest I'd be comfortable with.
`Dist` is used elsewhere outside of https://github.com/conda .
Unless it induces noticeable maintenance burden, we can also postpone the actual removal to early 2025 or so.
Marking this for `24.9` just gives us the possibility remove earlier.

conda/models/dist.py
```diff
@@ -17,12 +17,19 @@
...
+    addendum="Use `conda.models.records.PackageRecord` instead.",
```
mbargull, line 30: `PackageRecord` of course does not cover every functionality of this module one to one.
Feel free to add additional deprecation notes to symbols of this module as you see fit!

conda/plan.py
```diff
@@ -23,17 +23,13 @@
...
-from .history import History
+from .core.index import LAST_CHANNEL_URLS
+from .core.link import _get_best_prec_match, revert_actions  # noqa: F401
+from .deprecations import deprecated
 from .instructions import FETCH, LINK, SYMLINK_CONDA, UNLINK
 from .models.channel import Channel, prioritize_channels
 from .models.dist import Dist
 from .models.enums import LinkType
-from .models.match_spec import ChannelMatch
-from .models.prefix_graph import PrefixGraph
```
mbargull, line 36: Should we add `deprecated.constant(...)`s for all re-exported symbols this removes?

PR #13559, opened 2024-02-02
Catch ZstdError in jlap/repodata.json.zst handling

### LINKED ISSUE #13558: Fallback to repodata.json if repodata.json.zst raises a zstandard exception
### Checklist

### What happened?

conda should be able to fall back to repodata.json if repodata.json.zst is not zst, not only if it 404's.

### Conda Info

24.1.0

dholth:
### Description

Fix #13558

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/jlap/fetch.py
+++ b/conda/gateways/repodata/jlap/fetch.py
@@ -18,6 +18,8 @@ import jsonpatch
 import zstandard
 from requests import HTTPError
 
+from conda.common.url import mask_anaconda_token
+
 from ....base.context import context
 from ...connection import Response, Session
 from .. import (
@@ -117,7 +119,7 @@ def request_jlap(
     if etag and not ignore_etag:
         headers["if-none-match"] = etag
 
-    log.debug("%s %s", url, headers)
+    log.debug("%s %s", mask_anaconda_token(url), headers)
 
     assert session is not None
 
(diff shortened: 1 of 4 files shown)


conda/gateways/repodata/jlap/fetch.py
```diff
@@ -304,7 +304,10 @@ def request_url_jlap_state(
                     )
                 else:
                     raise JlapSkipZst()
-            except (JlapSkipZst, HTTPError) as e:
+            except (JlapSkipZst, HTTPError, zstandard.ZstdError) as e:
+                if isinstance(e, zstandard.ZstdError):
+                    # will this include the token
```
jezdez, line 309: I think that's reasonable, so maybe sanitize it? Or leave out the URL?

jezdez (approved):
Not sure if we actually need the debugging statement, but I concur it's good to catch zstandard decompression exceptions

dholth:
In general this entire code path could be more clear about whether the user's web server is configured correctly, and which variant they wind up getting. If we don't say something then the user would not know that they have a "404 page that returns a 200 code" for `repodata.json.zst`

jezdez:
Then lets make this and the other possible cases result in more specific (and actionable) logging statement? Info level?

dholth:
Improved log message with sanitized URL, and a bonus sanitized URL in a similar log message.

conda/gateways/repodata/jlap/fetch.py
dholth: Improved.

tests/core/test_solve.py
```diff
@@ -1895,7 +1895,7 @@ def test_python2_update(tmpdir):
         )
         assert convert_to_dist_str(final_state_1) == order1
 
-    specs_to_add = (MatchSpec("python=3"),)
+    specs_to_add = (MatchSpec("python=3.7"),)
```
dholth, line 1898: This part started failing for no runners seemingly on its own.

dholth: https://github.com/conda/conda-libmamba-solver/issues/317

tests/core/test_solve.py
```diff
@@ -1949,6 +1949,11 @@ def test_python2_update(tmpdir):
                     "channel-4::pycosat-0.6.3-py37h14c3975_0",
                 )
             )
+
+            pprint("Important parts")
+            pprint(set(important_parts))
+            pprint("Full solution")
+            pprint(set(full_solution))
```
jezdez, line 1956: Do we need those?

Issue #11850, opened 2022-09-16
conda rename: `unexpected error has occurred - TypeError: expected str, bytes or os.PathLike object, not NoneType`


thesofakillers:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

I am trying to use the new `conda rename` functionality to rename an environment `foo` to `bar`. I run:

`conda deactivate`

and then

`conda rename -n foo -d bar`

I get the following error:

<details>

```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<
 
    Traceback (most recent call last):
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/exceptions.py", line 1125, in __call__
        return func(*args, **kwargs)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/cli/main.py", line 86, in main_subshell
        exit_code = do_call(args, p)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 93, in do_call
        return getattr(module, func_name)(args, parser)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/cli/main_rename.py", line 53, in execute
        source = validate_src(args.name, args.prefix)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/cli/main_rename.py", line 28, in validate_src
        if common.is_active_prefix(prefix):
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/cli/common.py", line 87, in is_active_prefix
        paths_equal(prefix, context.active_prefix)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/site-packages/conda/common/path.py", line 53, in paths_equal
        return abspath(path1) == abspath(path2)
      File "/Users/thesofakillers/miniconda3/lib/python3.8/posixpath.py", line 374, in abspath
        path = os.fspath(path)
    TypeError: expected str, bytes or os.PathLike object, not NoneType
 
`$ /Users/thesofakillers/miniconda3/bin/conda rename -p /Users/thesofakillers/miniconda3/envs/xlaicl claficle`
 
  environment variables:
                 CIO_TEST=<not set>
                CONDA_EXE=/Users/thesofakillers/miniconda3/bin/conda
[...]

PR #13565, opened 2024-02-05
Enable `conda rename` to not throw a `TypeError` when conda is not activated

beeankha:
### Description

Fixes https://github.com/conda/conda/issues/11850


dholth (approved):
Thanks

Issue #13567, opened 2024-02-05
`conda activate` performance regressions


mbargull:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The time needed for `conda activate` has gone up in the recent versions again.
We made some performance improvements for `conda>=4.6` in gh-7122 and `conda>=23.5` in gh-12550, but have quickly regressed on this in recent versions.

The following gives an overview of `conda shell.posix activate` run times and
- uses `python=3.8` so we can compare over a wide range of `conda` versions,
- uses `setuptools=58` for `conda<=23.5` to factor out the impact of `setuptools>=60`'s `distutils` injection,
- uses `ruamel.yaml=0.16` since later versions got slower to import (tracking this upstream in https://sourceforge.net/p/ruamel-yaml/tickets/504/ ; already figured out the culprits/needed changes; will work with upstream to address this),
- has `ruamel_yaml` forcefully removed for the older `conda` versions to ensure the specified `ruamel.yaml` is used,

and was run on Linux `6.6.10`, glibc `2.17`, packages from `conda-forge` (apart from `defaults::conda=24.1.0`), with environments in memory on a `tmpfs`, and scaling governor/energy performance preference set to `performance`/`performance` on a 6 year old laptop with an `i7-8550U`.
(Running the same with `powersave`/`powersave` CPU settings causes times to rise upward of 200ms for the latest versions, i.e., very noticible delays.)

| time     | `conda`   | `ruamel.yaml` | `setuptools` | `python` | notes |
|----------|-----------|---------------|--------------|----------|-------|
| `0.128s` | `4.7.12`  | `0.16.12`     | `68.2.2`     | `3.8.18` | slow due to `setuptools>=60`'s `distutils` (fixed in gh-12550)
| `0.065s` | `4.7.12`  | `0.16.12`     | `58.5.3`     | `3.8.18` | no `python=3.8`-compatible builds for `<=4.6` available, so this is base
| `0.065s` | `4.8.5`   | `0.16.12`     | `58.5.3`     | `3.8.18` |
| `0.065s` | `4.9.2`   | `0.16.12`     | `58.5.3`     | `3.8.18` |
| `0.065s` | `4.10.3`  | `0.16.12`     | `58.5.3`     | `3.8.18` |
| `0.065s` | `4.11.0`  | `0.16.12`     | `58.5.3`     | `3.8.18` |
[...]

Issue #13541, opened 2024-01-30
Debug code guarded with `log.isEnabledFor(logging.DEBUG)` is always run


mbargull:
### What happened?

In `conda>=4.12` there is a seemingly innocuous change in the logging code:
- https://github.com/conda/conda/pull/11193/commits/487a02bbcb131e48f5ca280bfe39a431604b6d17#diff-f5c41f7e31bcc7aced31c4cad11ada6abc5478d93492e4b77715528d4642666eL331-L340
- https://github.com/conda/conda/pull/11193/commits/487a02bbcb131e48f5ca280bfe39a431604b6d17#diff-5c8832d586a9cd64e3162e388dfd6569fd77a0fe2031e04f2e287d4c088bb675R212

which changed the logger's level to `NOTSET` which in turn renders `log.isEnabledFor(logging.DEBUG)` as always true.

Having those `isEnabledFor`-guarded code always run increases the resource usage tremendously.
Local tests with the `conda-forge` channel, `libmamba` solver and a cleaned index showed 5-6 times peak memory usage (we're talking a couple of GiB here...) and runtime nearly doubled.
Interestingly, with the `classic` solver we seem to take slightly different code paths with much less impact.
The peak memory is happening
- in https://github.com/conda/conda/blob/23.11.0/conda/auxlib/logz.py#L165
- called from https://github.com/conda/conda/blob/23.11.0/conda/gateways/repodata/__init__.py#L152
- originating from https://github.com/conda/conda-libmamba-solver/blob/23.12.0/conda_libmamba_solver/index.py#L239

### Additional Context

Full reproducer showing the reduced resource usage when the logger level is not set to `NOTSET`:
<details><summary>output</summary>

```sh
 podman run --rm -it quay.io/condaforge/linux-anvil-cos7-x86_64 sh -ileuc "$( cat << 'end-of-script'                                                                                                                                                                             
conda create --quiet --yes --prefix=./conda-23.11 conda=23.11 python=3.11 patch time mamba
conda deactivate && unset CONDA_SHLVL && . ./conda-23.11/etc/profile.d/conda.sh && conda activate
test() {
  case ${1} in (clean) "${2}" clean --quiet --yes --index ;; esac
  printf '%80s\n' | tr ' ' '='
  printf 'TEST: %s\n' "${2} (${1})${3+ ${3}}"
  command time \
    --format='time: %es | peak memory: %MKB' \
    "${2}" create \
      --solver="${3-libmamba}" \
[...]

PR #13568, opened 2024-02-06
Optimize import time for `conda activate`

mbargull:
### Description

Avoid some imports to speed up `conda activate`.

Fixes gh-13567.

| `python`<br />`=3.8.18`,<br />`ruamel.yaml`<br />`=0.16.12` | `python`<br />`=3.11.7`,<br />`ruamel.yaml`<br />`=0.16.12` | `python`<br />`=3.8.18`,<br />`ruamel.yaml` <br /> `=0.18.5` | `python`<br />`=3.11.7`,<br />`ruamel.yaml`<br />`=0.18.5` | `python`<br />`=3.8.18`,<br />`ruamel.yaml` <br /> `=0.18.6` | `python`<br />`=3.11.7`,<br />`ruamel.yaml`<br />`=0.18.6` | changes |
|-------|-------|-------|-------|-------|-------|-------|
| 0.089 | 0.093 | 0.093 | 0.098 | 0.090 | 0.95 | base commit 59af01b from `main` |
| 0.084 | 0.088 | 0.089 | 0.093 | | | `Avoid costly .gateways.logging import for activate` |
| 0.086 | 0.091 | 0.091 | 0.096 | | | `Avoid costly argparse import for activate` |
| 0.086 | 0.090 | 0.090 | 0.095 | | | `Use packaging.version in .deprecations only if needed` |
| 0.080 | 0.081 | 0.086 | 0.093 | | | `Avoid costly inspect.getmodule in .deprecations` |
| 0.086 | 0.090 | 0.090 | 0.095 | | | `Defer platformdirs import for conda activate` |
| 0.068 | 0.071 | 0.074 | 0.081 | 0.069 | 0.072 | everything above combined |
| 0.065 | 0.068 | | | 0.065 | 0.68 | everything above plus `Defer tempfile import for auxlib.compat` |

(Edit: Added `ruamel.yaml=0.18.6` timings after improvements re. https://sourceforge.net/p/ruamel-yaml/tickets/504/ .)

### Checklist - did you ...

- ~~[ ] Add / update necessary tests?~~

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -25,8 +25,8 @@ class DeprecatedError(RuntimeError):
 # inspired by deprecation (https://deprecation.readthedocs.io/en/latest/) and
 # CPython's warnings._deprecated
 class DeprecationHandler:
-    _version_str: str
-    _version_tuple: tuple[int, ...]
+    _version_str: str | None
+    _version_tuple: tuple[int, ...] | None
     _version_object: Version | None
 
     def __init__(self, version: Version | str):
@@ -40,7 +40,10 @@ class DeprecationHandler:
         if version is not None:
             self._version_str = str(version)
             if not isinstance(version, str):
-                self._version_object = version
+                from packaging.version import Version
+
+                if isinstance(version, Version):
+                    self._version_object = version
             # Try to parse the version string as a simple tuple[int, ...] to
             # avoid packaging.version import and costlier version comparisons.
             self._version_tuple = self._get_version_tuple(self._version_str)
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -25,68 +25,52 @@ class DeprecatedError(RuntimeError):
 # inspired by deprecation (https://deprecation.readthedocs.io/en/latest/) and
 # CPython's warnings._deprecated
 class DeprecationHandler:
-    _version_str: str | None
+    _version: str | None
     _version_tuple: tuple[int, ...] | None
     _version_object: Version | None
 
-    def __init__(self, version: Version | str):
+    def __init__(self, version: str):
         """Factory to create a deprecation handle for the specified version.
 
         :param version: The version to compare against when checking deprecation statuses.
         """
-        self._version_str = None
+        self._version = version
+        # Try to parse the version string as a simple tuple[int, ...] to avoid
+        # packaging.version import and costlier version comparisons.
+        self._version_tuple = self._get_version_tuple(version)
         self._version_object = None
-        self._version_tuple = None
-        if version is not None:
-            self._version_str = str(version)
-            if not isinstance(version, str):
-                from packaging.version import Version
 
-                if isinstance(version, Version):
-                    self._version_object = version
-            # Try to parse the version string as a simple tuple[int, ...] to
-            # avoid packaging.version import and costlier version comparisons.
-            self._version_tuple = self._get_version_tuple(self._version_str)
+    @staticmethod
+    def _get_version_tuple(version: str) -> tuple[int, ...] | None:
+        """Return version as non-empty tuple of ints if possible, else None.
 
-    @property
-    def _version(self) -> Version:
-        """Populate and return self._version_object with parsed version string."""
-        # Lazily import this to reduce import time for conda activate.
+        :param version: Version string to parse.
+        """
+        try:
+            return tuple(int(part) for part in version.strip().split(".")) or None
+        except (AttributeError, ValueError):
+            return None
+
+    def _version_less_than(self, version: str) -> bool:
+        """Test whether own version is less than the given version.
+
+        :param version: Version string to compare against.
+        """
+        if self._version_tuple:
+            if version_tuple := self._get_version_tuple(version):
+                return self._version_tuple < version_tuple
+
+        # If self._version or version could not be represented by a simple
+        # tuple[int, ...], do a more elaborate version parsing and comparison.
+        # Avoid this import otherwise to reduce import time for conda activate.
         from packaging.version import parse
 
         if self._version_object is None:
             try:
-                self._version_object = parse(self._version_str)
+                self._version_object = parse(self._version)
             except TypeError:
                 self._version_object = parse("0.0.0.dev0+placeholder")
-        return self._version_object
-
-    @staticmethod
-    def _get_version_tuple(version_str: str) -> tuple[int, ...]:
-        """Parse version as a tuple of ints if possible, else return empty tuple.
-
-        :param version_str: Version string to parse.
-        """
-        try:
-            return tuple(int(part) for part in version_str.strip().split("."))
-        except (AttributeError, ValueError):
-            return tuple()
-
-    def _version_less_than(self, version_str: str) -> bool:
-        """Test whether own version is less than the given version.
-
-        :param version_str: Version string to compare against.
-        """
-        if self._version_tuple:
-            if version_tuple := self._get_version_tuple(version_str):
-                return self._version_tuple < version_tuple
-
-        # If self._version_str or version_str could not be represented by a
-        # simple tuple[int, ...], do a more elaborate version parse and compare.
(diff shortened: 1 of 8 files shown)


conda/deprecations.py
```diff
@@ -285,6 +335,20 @@ def _get_module(self, stack: int) -> tuple[ModuleType, str]:
 
         try:
             frame = sys._getframe(2 + stack)
+            # Shortcut finding the module by manually inspecting loaded modules.
+            if hasattr(frame, "f_code") and hasattr(frame.f_code, "co_filename"):
```
kenodegard, line 339: duck typing instead?

and looking at this again it appears that 338-354 should actually be defined outside this try-except:

```python
try:
    frame = sys._getframe(2 + stack)
except IndexError:
    # IndexError: 2 + stack is out of range
    raise DeprecatedError("unable to determine the calling module")

try:
    filename = frame.f_code.co_filename
except AttributeError:
    # AttributeError: frame.f_code.co_filename is undefined
    pass
else:
    for module in sys.modules.values():
        ...

import inspect
...
```

conda/deprecations.py
```diff
@@ -285,6 +335,20 @@ def _get_module(self, stack: int) -> tuple[ModuleType, str]:
```
kenodegard, line 335: FYI the original `import inspect` is still on the line above

mbargull: Argh, forgot to add that to the commit!

conda/deprecations.py
mbargull: > duck typing instead?

I don't mind either way; added your suggestion!

> and looking at this again it appears that 338-354 should actually be defined outside this try-except:

Ah, I didn't even look at the exception type. Makes sense -- added!

conda/deprecations.py
```diff
@@ -22,17 +25,64 @@ class DeprecatedError(RuntimeError):
...
+
+        if self._version_object is None:
+            try:
+                self._version_object = parse(self._version_str)
+            except TypeError:
+                self._version_object = parse("0.0.0.dev0+placeholder")
+        return self._version_object
+
+    @staticmethod
+    def _get_version_tuple(version_str: str) -> tuple[int]:
```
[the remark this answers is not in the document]
mbargull: I always forget the ellipsis :facepalm: -- thanks, added!

mbargull:
Thanks for the early review, Ken!
I'll do some timing on this and possibly look for more pieces (but I think upstream work in `ruamel.yaml` would cover most remaining stuff already).

kenodegard:
@mbargull the benchmarking tests (via CodSpeed) don't run correctly ATM but can we include a benchmark test (mark it with `@pytest.mark.benchmark`) so we can track this better once we get CodSpeed working?

mbargull:
> @mbargull the benchmarking tests (via CodSpeed) don't run correctly ATM but can we include a benchmark test (mark it with `@pytest.mark.benchmark`) so we can track this better once we get CodSpeed working?

As for really benchmarking, I don't know, actually.
These activation/import times are very sensitive, so you'd need to be careful of the setup and how you'd compare results.
You need a reasonably quiet system (e.g., not too much CPU, disk activity in the background) and definitely should run timings for the merge base in the same job to get valid target data points to compare against.
The Python interpreter gives us a rather small budget to work with and is at the same time very sensitive to background work (there are just many CPU cycles and disk accesses involved in these startup/import phases...).
I have no idea how CodSpeed works/operates and as such can't really tell how things would work out.

We could instead rather have some proxies we could use to assess this, though, by not looking at the timings themselves but rather look for "known bad/slow imports".
E.g., parse the output of `PYTHONPROFILEIMPORTTIME=1 conda shell.${SHELL_NAME} activate 2>&1 > /dev/null` and ensure that `inspect` etc. is not imported.
(This would be platform and shell specific, e.g., `tempfile` would probably be imported for `shell.cmd.exe`.)

I wanted to open an issue for us to track addition of basic performance regression tests anyway, so we can avoid regressing on this, but also the massive memory+cpu one from gh-13541 .

dholth:
`len(sys.modules)` would be a pretty good proxy.

conda/deprecations.py
```diff
@@ -22,17 +25,65 @@ class DeprecatedError(RuntimeError):
 # inspired by deprecation (https://deprecation.readthedocs.io/en/latest/) and
 # CPython's warnings._deprecated
 class DeprecationHandler:
-    _version: Version
+    _version_str: str
+    _version_tuple: tuple[int, ...]
```
kenodegard, line 29:
```suggestion
    _version_str: str | None
    _version_tuple: tuple[int, ...] | None
```

conda/deprecations.py
```diff
@@ -22,17 +25,65 @@ class DeprecatedError(RuntimeError):
...
         """Factory to create a deprecation handle for the specified version.
 
         :param version: The version to compare against when checking deprecation statuses.
         """
+        self._version_str = None
+        self._version_object = None
+        self._version_tuple = None
+        if version is not None:
+            self._version_str = str(version)
+            if not isinstance(version, str):
```
kenodegard, line 42: Technically anything could be passed in, so just checking for `str` is incomplete?

```suggestion
            if isinstance(version, Version):
```

kenodegard (approved):
Two minor suggestions

conda/deprecations.py
mbargull:
```suggestion
            if not isinstance(version, str):
                from packaging.version import Version
                
                if isinstance(version, Version):
```
otherwise we'd need to import `packaging.version` unconditionally again.

kenodegard: actually why don't we just change the version input to only expect `str`? we initialize the `DeprecationHandler` as:

```python
deprecated = DeprecationHandler(__version__)
```

and I wouldn't expect anyone to pass in a `Version` object (even in our tests we only test with passing in `None` or `str`):

```python
deprecated = DeprecationHandler(Version(...))
```

mbargull: Yes, I'd be in favor of that.
I just kept the `Version` input because it was there and I didn't care to look if/how things are actually used :D.

mbargull: Done and simplified the code accordingly; looks clearer with much less boilerplate.

kenodegard (dismissed):

kenodegard (approved):

jezdez:
For the record this change should never have been merged since it implements a custom version parsing scheme, which is know to be a source of hard-debug issues in packaging ecosystems.

mbargull:
> For the record this change should never have been merged since it implements a custom version parsing scheme, which is known to be a source of hard-debug issues in packaging ecosystems.

Explain? In case you are referring to `.deprecations._get_version_tuple`/`.deprecations._version_less_than`:
These are entirely internal functions and `.deprecations.deprecated` *should* only be used with `conda`'s own version as input.
And even then: The parsing therein is only used when both versions of the comparison can be converted via `tuple(int(part) for part in version.strip().split("."))` without raising any errors -- if there are any errors, it gracefully falls back to `packaging.version.parse`.

To summarize: It is not meant for general usage but even then gracefully falls back to established version parsing code.
Hence, I do not share your concerns at all.

(Rather than reverting this back to always using `packaging.version`, I'd even suggest changing the interface such that it doesn't take a string that has to be parsed expensively, if you'd really want to amend this.)

jezdez:
> To summarize: It is not meant for general usage but even then gracefully falls back to established version parsing code. Hence, I do not share your concerns at all.

I think you've misunderstood the goal of the deprecations framework, it's supposed to be used by other conda related libraries, in support of plugin based conda ecosystem, and not just internally.

mbargull:
> I think you've misunderstood the goal of the deprecations framework, it's supposed to be used by other conda related libraries, in support of plugin based conda ecosystem, and not just internally.

Ah, okay, I was not aware of that.
If I had been, I'd certainly been more stringent about this, i.e., included the changes from gh-13681 from the get go.

PR #13577, opened 2024-02-08
Move `conda env export` to `conda export`

beeankha:
### Description

Related epic:
- https://github.com/conda/conda/issues/11633

### Checklist - did you ...


conda/env/env.py
```diff
@@ -63,77 +63,6 @@ def validate_keys(data, kwargs):
...
-                channels.insert(0, canonical_name)
-    return Environment(
-        name=name,
-        dependencies=dependencies,
-        channels=channels,
-        prefix=prefix,
-        variables=variables,
-    )
-
-
```
beeankha, line 136: Oops; will copy back `from_environment` to the correct location

conda/cli/main_env_export.py
kenodegard, line 1: I agree this should be an alias for `conda export` but I don't think we should mark `conda env export` for deprecation (yet)

conda/cli/main_export.py
```diff
@@ -0,0 +1,131 @@
...
+    add_parser_json(p)
+
+    p.add_argument(
+        "--from-history",
+        default=False,
+        action="store_true",
+        required=False,
+        help="Build environment spec from explicit specs in history",
+    )
+    p.set_defaults(func="conda.cli.main_env_export.execute")
```
kenodegard, line 91:
```suggestion
    p.set_defaults(func="conda.cli.main_export.execute")
```

news/13577-conda-export-command
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Introducing the `conda export` command to replace `conda env export`. (#13577)
```
kenodegard, line 3:
```suggestion
* Move `conda env export` to `conda export` and alias the old command to the new command. (#13577)
```

tests/cli/test_main_export.py
kenodegard, line 1: is this a duplicate of the `conda env export` tests? can we combine them and parameterize the tests instead?

kenodegard (changes requested):

tests/cli/test_main_export.py
beeankha: The tests in `tests/cli/test_main_export.py` are not duplicates of the `conda env export`-related tests in `tests/cli/test_env.py`; eventually once `conda env export` is deprecated we can move (or possibly delete) some of those related `test_env.py` tests into `test_main_export.py`.

beeankha: After thinking about it a bit more, I think it would be best to move the `conda env export`-related tests from `tests/cli/test_env.py` into `test_main_export.py`; at the same time we could move some of the functions used in multiple tests (e.g. [`env_is_created`](https://github.com/conda/conda/blob/main/tests/cli/test_env.py#L476-L491)) into a test fixtures file so that they can be used in multiple places. In order to keep things as simple as possible, I'll do this work in a separate PR.

conda/cli/main_env.py
```diff
@@ -43,7 +43,7 @@ def configure_parser(sub_parsers: _SubParsersAction | None, **kwargs) -> Argumen
     )
     main_env_config.configure_parser(env_parsers)
     main_env_create.configure_parser(env_parsers)
-    main_env_export.configure_parser(env_parsers)
+    main_export.configure_parser(env_parsers)
```
kenodegard, line 46: Ahh! Yes! This is exactly what we want!

conda/cli/main_export.py
```diff
@@ -114,6 +116,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     if args.file is None:
         stdout_json(env.to_dict()) if args.json else print(env.to_yaml(), end="")
     else:
+        filename = args.file
+        # check for the proper file extension; otherwise when the export file is used later,
+        # the user will get a file parsing error
+        if not filename.endswith((".yml", ".yaml")):
```
kenodegard, line 122: Can we use one of the yaml constants we already have defined? E.g. https://github.com/conda/conda/blob/6488061bdb0c46251bddc245d673b119e8c12e16/conda/common/configuration.py#L1310

beeankha: I wasn't aware of this constant, thanks for pointing it out!

travishathaway (approved):

conda/cli/main_export.py
kenodegard, line 1: Unfortunately this rename also needs a deprecation

kenodegard (changes requested):

kenodegard (changes requested):

kenodegard (approved):

PR #13598, opened 2024-02-15
Provide more informative `MatchSpec` errors

### LINKED ISSUE #11203: Output corresponding package name when `InvalidVersionSpec` is raised
### Checklist

### What is the idea?

When `conda env export` fails with an `InvalidVersionSpec` exception, it would be nice if the error message included the name of the package whose metadata contains the invalid character.

### Why is this needed

Including dependencies, I have hundreds of packages in my environment. It's incredibly difficult to track down which package causes an InvalidVersionSpec exception. If I had this information, I could try using a different version of the package, raise an issue with the maintainers, manually change the metadata on my local machine, or get rid of the package altogether. Without this information, I'm flying blind.

### What should happen?

When `conda env export` fails with an `InvalidVersionSpec` exception, the error message would be something like
`Invalid version <something>: invalid character(s). Dependency of <insert package here>`

### Additional Context

Here is the traceback of an error I got after calling `conda env export -v --no-builds -n base`:
```
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 38, in __call__
    return cls._cache_[arg]
KeyError: '>=4.2.2<6.0.0'

  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 38, in __call__
    return cls._cache_[arg]
KeyError: '4.2.2<6.0.0'

  File "/opt/conda/lib/python3.8/site-packages/conda/exceptions.py", line 1080, in __call__
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/conda_env/cli/main.py", line 80, in do_call
  File "/opt/conda/lib/python3.8/site-packages/conda_env/cli/main_export.py", line 109, in execute
    env = from_environment(name, prefix, no_builds=args.no_builds,
  File "/opt/conda/lib/python3.8/site-packages/conda_env/env.py", line 109, in from_environment
    precs = tuple(PrefixGraph(pd.iter_records()).graph)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/prefix_graph.py", line 39, in __init__
    parent_match_specs = tuple(MatchSpec(d) for d in node.depends)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/prefix_graph.py", line 39, in <genexpr>
    parent_match_specs = tuple(MatchSpec(d) for d in node.depends)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/match_spec.py", line 55, in __call__
    return super(MatchSpecType, cls).__call__(**parsed)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/match_spec.py", line 177, in __init__
    self._match_components = self._build_components(**kwargs)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/match_spec.py", line 403, in _build_components
    return frozendict(_make_component(key, value) for key, value in iteritems(kwargs))
  File "/opt/conda/lib/python3.8/site-packages/conda/models/match_spec.py", line 403, in <genexpr>
    return frozendict(_make_component(key, value) for key, value in iteritems(kwargs))
  File "/opt/conda/lib/python3.8/site-packages/conda/models/match_spec.py", line 417, in _make_component
    matcher = _implementors[field_name](value)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 40, in __call__
    val = cls._cache_[arg] = super(SingleStrArgCachingType, cls).__call__(arg)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 497, in __init__
    vspec_str, matcher, is_exact = self.get_matcher(vspec)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 546, in get_matcher
    self.matcher_vo = VersionOrder(vo_str)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 40, in __call__
    val = cls._cache_[arg] = super(SingleStrArgCachingType, cls).__call__(arg)
  File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 172, in __init__
    raise InvalidVersionSpec(vstr, "invalid character(s)")
conda.exceptions.InvalidVersionSpec: Invalid version '4.2.2<6.0.0': invalid character(s)
```

I was eventually able to resolve the issue by adding the `--from-history` flag, but I don't think that's an ideal solution.

There have been other issues raised here concerning `InvalidVersionSpec`, such as #9624, but the solutions given there vary wildly and only work for specific packages. I think adding this feature would make the issue much more solvable for a developer who encounters it. 

From looking at the traceback and the codebase, I think a good solution would be to put a try-except [here](https://github.com/conda/conda/blob/master/conda/models/prefix_graph.py#L39), and print the name of the node along with the error message.

jaimergp:
### Description

Catch `InvalidSpec` during `MatchSpec` initialization to report which spec was responsible of the exception.

Closes #11203 

### Checklist - did you ...


jaimergp:
Failing as expected:

```
FAILED tests/models/test_match_spec.py::test_invalid_version_reports_spec[pkg 4.2.2<6.0.0] - conda.exceptions.InvalidVersionSpec: Invalid version '4.2.2<6.0.0': invalid character(s)
FAILED tests/models/test_match_spec.py::test_invalid_version_reports_spec[pkg>=0.10.0,<1.0.0<py312|>=0.13.0,<1.0.0>=py312] - conda.exceptions.InvalidVersionSpec: Invalid version '1.0.0<py312': invalid character(s)
```

Adding fix now.

conda/models/match_spec.py
```diff
@@ -33,6 +33,17 @@
...
 class MatchSpecType(type):
     def __call__(cls, spec_arg=None, **kwargs):
+        try:
+            return cls._call(spec_arg, **kwargs)
+        except InvalidSpec as e:
+            msg = ""
+            if spec_arg:
+                msg += f"{spec_arg}"
+            if kwargs:
+                msg += " " + ", ".join(f"{k}={v}" for k, v in kwargs.items())
```
kenodegard, line 43: Isn't this the same info as is added to the f-string below on line 77?

jaimergp: It is, and kept like that for backwards compatibility, but this also adds that error message to any other exceptions raised during `MatchSpec` initialization (like the version problem reported in the issue).

kenodegard: Ah I see

kenodegard (dismissed):

conda/models/match_spec.py
```diff
@@ -33,6 +33,17 @@
...
+            raise InvalidMatchSpec(msg, details=e) from e
+
+    def _call(cls, spec_arg=None, **kwargs):
```
jezdez, line 46: This looks off due to the name of the argument, it takes a class object and is called from the __call__ method, I think this should be a classmethod and not an instance method to not be handed down to MatchSpec instances?

jaimergp: I'm borrowing the signature from `__call__` above. I'm not sure what gets passed to the method here, since we are in the Metaclass land now. Should `__call__` also be a classmethod by that definition?

jezdez: Reading up on this a little more, you're right, the __call__ method of type subclasses get passed the class and not instances and it's a misleading case of this meta-programming pattern that the _call method looks like taking an instance. This is bound to confuse people (as it did me), so I would suggest to simply not call a separate method with a confusing name `_call` and just inline the try/except block. That should suffice to handle this and not hurt us long-term.

jaimergp: Done!

jezdez (approved):

PR #13609, opened 2024-02-20
Enable fast coverage tracking

dbast:
See also:

* https://github.com/nedbat/coveragepy/releases/tag/7.4.0
* https://docs.python.org/3/whatsnew/3.12.html#pep-669-low-impact-monitoring-for-cpython
* https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring.html

Trying the option as inspired by @jaimergp to see
* what is the potential speed gain from it
* what breaks
* if that correctly falls back to the old method in case of Python <3.12.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+env:
+  COVERAGE_CORE: sysmon
+
 jobs:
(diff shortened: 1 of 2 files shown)


dbast:
yay... defaults still has `coverage-7.2.2`, but >=7.4 is required :(

dbast:
learnings:
* `COVERAGE_CORE: sysmon` can only be set with Python >=3.12, otherwise coverage errors instead of falling back to the old method
* can not see any speed improvement by comparing the different runs as part of the commits of this PR

thus closing.

PR #13628, opened 2024-02-27
Fix auxlib.logz.stringify excess memory allocation

mbargull:
### Description

This addresses the excess resource usage observed in gh-13541.

Essentially, the whole `repodata.json` is parsed via `requests.Response.json` and then dumped into a formatted string -- with both parts allocating multiple GB in RAM for large channels like `conda-forge`.
All this happens in a context only for (intended) debug logging output which is meant to constrain the log message length (and in turn memory usage); the proposed changes just apply the intended size constraints.

(It only handles the severe run time impact but doesn't fix the underlying cause for gh-13541.)

### Checklist - did you ...

### DIFF
--- a/conda/auxlib/logz.py
+++ b/conda/auxlib/logz.py
@@ -1,5 +1,5 @@
 from itertools import islice
-from json import JSONEncoder, dumps
+from json import JSONEncoder, dumps, loads
 from logging import getLogger, INFO, Formatter, StreamHandler, DEBUG
 from sys import stderr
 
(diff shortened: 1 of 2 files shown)


dholth:
This is concerning. It would be better to never instantiate `response.text` for example the jlap / repodata.json.zst code tries to stream the response directly to disk. Thanks for working on this!

mbargull:
> It would be better to never instantiate `response.text` for example the jlap / repodata.json.zst code tries to stream the response directly to disk.

Yes, ideally we'd always stream this.
As is, the [code calling `stringify` also uses `requests.Response.text`](https://github.com/conda/conda/blob/24.1.2/conda/gateways/repodata/__init__.py#L159-L169), so this should not pose much impact.
I.e., there are multiple places for us to replace whenever we get to using streaming for all `repodata.json` handling.

jezdez:
This reminds me that we need to find a way to track memory usage in the performance tracking somehow.

jezdez (approved):

kenodegard (approved):

PR #13648, opened 2024-03-04
Update dev container

zklaus:
### Checklist - did you ...


.devcontainer/devcontainer.json
kenodegard, line 1: Let's switch the indentations to spaces instead of tabs.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM mcr.microsoft.com/devcontainers/miniconda:0-3
+
+# Copy environment.yml (if found) to a temp location so we update the environment. Also
+# copy "noop.txt" so the COPY instruction does not fail if no environment.yml exists.
+# COPY environment.yml* .devcontainer/noop.txt /tmp/conda-tmp/
+# RUN if [ -f "/tmp/conda-tmp/environment.yml" ]; then umask 0002 && /opt/conda/bin/conda env update -n base -f /tmp/conda-tmp/environment.yml; fi \
+#     && rm -rf /tmp/conda-tmp
+
+COPY tests/requirements.txt /tmp/conda-tmp/
```
kenodegard, line 9: If the intention is for this container to be used for testing (e.g., to run pytest) we will need to also copy and install from `tests/requirements-Linux.txt` and `tests/requirements-ci.txt`

see `dev/start`:

https://github.com/conda/conda/blob/0f8fb581cdfb1716b22596dace01747bc978eeb1/dev/start#L241-L253

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM mcr.microsoft.com/devcontainers/miniconda:0-3
```
jezdez, line 1:
```suggestion
FROM continuumio/miniconda3:latest
```

.devcontainer/requirements-dev.txt
```diff
@@ -0,0 +1,2 @@
+esbonio
```
jezdez, line 1: Is this something we need? It adds a whole number of dependencies into the developer toolchain that we'd have to support going forward.

zklaus: This is the language server for restructured text and as such helpful for writing docs, though I have to admit that it is giving me a bit of trouble. If we want to move mainly to MyST for docs, maybe this is less crucial.

I am not sure I share your concern for future support. This is, after all, only a development environment, so imho it is easier switch things out. In this case, for example, if we think rst is still relevant for now, but we move to MyST later on, we can just drop this.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN umask 0002 \
+    && /opt/conda/bin/conda install \
+    -n base --yes --quiet --override-channels \
+    --channel=defaults \
+    --file=/tmp/conda-tmp/requirements.txt \
+    --file=/tmp/conda-tmp/requirements-ci.txt \
+    --file=/tmp/conda-tmp/requirements-Linux.txt \
+    --file=/tmp/conda-tmp/requirements-dev.txt \
+    && rm -rf /tmp/conda-tmp
```
jaimergp, line 17: I think we can move this logic to `postCreateCommand` and avoid the COPY above.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+
+    // Features to add to the dev container. More info: https://containers.dev/features.
+    // "features": {},
+
+    // Use 'forwardPorts' to make a list of ports inside the container available locally.
+    // "forwardPorts": [],
+
+    // Use 'postCreateCommand' to run commands after the container is created.
+    "postCreateCommand": "/opt/conda/bin/python -m conda init --dev bash",
```
jaimergp, line 17:
```suggestion

    // Features to add to the dev container. More info: https://containers.dev/features.
    // "features": {},

    // Use 'forwardPorts' to make a list of ports inside the container available locally.
    // "forwardPorts": [],

    // Use 'postCreateCommand' to run commands after the container is created.
    "postStartCommand": "/opt/conda/bin/python -m conda init --dev bash",
```

jaimergp, line 17:
```suggestion
    "postStartCommand": "/opt/conda/bin/python -m conda init --dev bash",
```

jaimergp:
I've been experimenting with this over at https://github.com/conda/conda-libmamba-solver/pull/451 and works suuuper nicely :)

.devcontainer/Dockerfile
zklaus: Well, it works, but it leaves you with a container that is run as root, whereas the MSFT version is run as the user `vscode`---arguably a better approximation of what a user might find. It achieves that not by directly deriving from our image, but by copying only the `/opt/conda` folder from `continuumio/miniconda3:latest` into the new image, which is based on `mcr.microsoft.com/devcontainers/base:1-bullseye`.

.devcontainer/Dockerfile
zklaus: We can. The upside is the avoided copy. The downside is potentially worse caching. If we keep that in the Dockerfile, rebuilding the devcontainer should reuse the installed dependencies and only require updating the entrypoint script, right?

zklaus:
I had a chat with @jaimergp and subsequently simplified this PR a bit. There are many things imaginable with devcontainers like improved rst support and installation with an unprivileged user account instead of root in the container, but this represents a simple, functional starting point that can be improved in the future with specific development goals.

.github/dependabot.yml
kenodegard, line 1: What will this update? I don't see any versions defined in the devcontainer

kenodegard: are the extensions defined in `devcontainer.json` supposed to be versioned?

zklaus: Good catch. This is for updating the devcontainer `features`. However, since we are now deriving from a "naked" docker container which doesn't have any features and we are not adding any features here either, this is superfluous.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,34 @@
...
+                "ms-toolsai.jupyter",
+                "lextudio.restructuredtext-pack",
+                "lextudio.restructuredtext@prerelease",
+                "swyddfa.esbonio@prerelease",
+                "ms-azuretools.vscode-docker"
+            ]
+        }
+    }
+
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
```
jaimergp, line 32: We are already connecting as root anyway, right?

zklaus: Yes. This only becomes relevant when the default user is a different one, for example because the devcontainer is based on another container that uses a different user or when making use of the [common-utils feature](https://github.com/devcontainers/features/tree/main/src/common-utils).

jezdez (dismissed):

jezdez:
That reminds me, we should update the dev environment docs to highlight this change: https://github.com/conda/conda/blob/a2e5e6601441c0f70bdc75d013cbd728af1dea2d/docs/source/dev-guide/development-environment.md

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,34 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/miniconda
+{
+    "name": "Miniconda (Python 3)",
```
jaimergp, line 4:
```suggestion
    "name": "Miniconda (default-channel=defaults)",
```

.devcontainer/devcontainer.json
jaimergp: In that case I'd change the comment, because right now that "instead" is a bit misleading.
```suggestion
    // Adjust to connect as non-root instead. More info: https://aka.ms/dev-containers-non-root.
```

jezdez (approved):

PR #12874, opened 2023-07-12
Add (back) `cygpath` fallback

kenodegard:
### Description

The `cygpath` fallback logic was removed in https://github.com/conda/conda/pull/12627 because the code stated:

https://github.com/conda/conda/blob/84f0d2b56c2e81b62841f8926d2ee319182fdb67/conda/activate.py#L850-L851

So we incorrectly assumed that `conda` would always have `cygpath` available on Windows (this is in fact not the case, see https://github.com/conda/conda/issues/12873).

We add back the fallback logic until a better solution (if there is one) is decided upon.

Resolves #12873

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #13649, opened 2024-03-04
support multiple MSYS2 environments

ifitchet:
#### Rationale

MSYS2 supports multiple [environments](https://www.msys2.org/docs/environments/) including several native compilation environments as well as the POSIX-oriented MSYS environment and this PR aims provide a cleaner mechanism to support them.

The native compilation environments differ in regards to which compiler suite is used (GCC or LLVM) and which C/C++ runtimes are used.

##### History

conda will only support a single native compilation environment plus the MSYS environment.  One of the MSYS2 `mingw32` or `mingw64` native compilation environments is reworked into `{prefix}/Library/mingw-w64` and the MSYS environment is placed in `{prefix}/Library` becoming, in effect, `{prefix}/Library/usr/*`.

(The `mingw-w64` name is the base of the MSYS2 [native compilation environment recipes](https://github.com/msys2/MINGW-packages), eg. `mingw-w64-zlib`.)

The following (pertinent) directories are added to the user's PATH in this order:

1. `{prefix}/Library/mingw-w64/bin` - the MSYS2 native compilation environment binaries
1. `{prefix}/Library/usr/bin` - the MSYS2 POSIX environment binaries
1. `{prefix}/Library/bin` - the conda package binary install location

**NOTE**: The final item, `{prefix}/Library/bin`, is problematic.  Internally, MSYS2 has nothing in `/bin` and mounts `/usr/bin` onto `/bin` (presumably to enable the usage of `#! /bin/bash` etc.) but `/usr/bin` is `{prefix}/Library/usr/bin` and `/bin` is `{prefix}/Library/bin`.  That is, MSYS2 utilities may have the conda package binary install location obscured with an overlying mount of `{prefix}/Library/usr/bin`.

#### Proposal

The proposal is for conda to support any of the 64-bit MSYS2 native compilation environments and always include the existing `mingw-w64` system for backwards compatibility.  No change will occur to the MSYS environment.

The new environments will use the same distinguishing directory as the MSYS2 distribution, so the UCRT64 environment will use a `{prefix}/Library/ucrt64` directory, CLANG64 the `{prefix}/Library/clang64` directory etc..

There is an immediate benefit here as the "repack" of the MSYS2 tarball into a conda archive becomes a far simpler `mv`.

The mechanism will be for a **dynamic** lookup of the native compilation environment distinguishing directories plus the current `mingw-w64` directory.

##### Order and PATH Insertion

1. there is an order to the dynamic lookup of the distinguishing directories: [`ucrt64`, `clang64`, `mingw64`, `clangarm64`]
   
   Each `{dir}` in the above list is looked for as `{prefix}/Library/{dir}`.

   1. the first detected directory is the **only** one put on the user's PATH as `{prefix}/Library/{dir}/bin`
   
      Extra matching directories will elicit a warning but will **not** be put on the user's PATH.
   
   1. if no distinguishing directory is found then nothing will be put on the user's PATH
   
1. Then `{prefix}/mingw-w64/bin`, `{prefix}/usr/bin` and `{prefix}/bin` will be put on the user's PATH as usual
   
#### Existing conda Behaviour

If nothing else is done the existing conda behaviour will largely continue unchanged: `{prefix}/Library/mingw-w64/bin` will be put on the user's PATH.

#### Packages Bundling MSYS2

The `main/git` package (and others?) is a repack of a [Git For Windows](https://github.com/git-for-windows/git) release which bundles an entire MSYS2 tree internally.  This currently includes the underlying MSYS2 `mingw64` hierarchy.

Currently, conda does not add the `mingw64/bin` directory to the user's PATH.  (If you run the supplied `Git Bash` Start Menu item then `/mingw64/bin` is added to that shell's PATH -- `Git Bash` appears to be a re-spin of the MSYS2 `MSYS2 MINGW64` Start Menu item.)

The conda packaging puts this tree in `{prefix}/Library` meaning that `{prefix}/Library/mingw64` may be picked up as a distinguishing directory and `{prefix}/Library/mingw64/bin` will be put on the user's PATH:

* if no prior distinguishing directories exist then `{prefix}/Library/mingw64/bin` will be put on PATH ahead of the (potentially non-existent) `{prefix}/Library/mingw-w64/bin`
  
  Here the user will have the git-related executables appear on their PATH which they never had had previously.
  
  **NOTE** If the `{prefix}/Library/mingw-w64` existed and had similarly named executables in it then they would not be visible any more.  See the section on Anaconda below.

* if a prior distinguishing directory does exist, say, `clang64`, then `{prefix}/Library/clang64` will be used instead of `{prefix}/Library/mingw64`
  
  Here the user would not see the git-related executables just as previously.

(The Git For Windows people [don't seem inclined to switch to UCRT64](https://github.com/git-for-windows/git/issues/3167#issuecomment-1240475437) at the moment.)

#### Anaconda Distribution

Anaconda distribution does come bundled with a `{prefix}/Library/mingw-w64` directory which contains `jq.exe` and `libwinpthread-1.dll` (the `m2w64-jq` MSYS2 package and dependency).  If someone installs `main/git` (or a similar package with a bundled MSYS2) then those executables may be obscured.

#### Package mutexes

Packages in the native compilation environments **should** associate themselves with a discriminating mutex to minimise conflicts.  We would propose `msys2-mingw-w64-mutex` with a build string of `ucrt64`, `clang64` etc..

The rationale for the package mutex is that whilst the code changes to `activate.py` will only put one native compilation environment on the PATH it is better all round if you can be prevented from installing potentially confusing libraries in the same conda environment in the first place.

There is a question here about backporting mutexes for, say, the existing `m2w64-*` packages which is raised but left unanswered.  A mutex build string of `m2w64` will be defined for `m2w64-*` packages.

Similarly, a mutex for the MSYS (POSIX) environment, `msys2-msys-mutex` has been defined with a build string of `msys2`.  A variant with a mutex build string of `m2` will be defined for `m2-*` packages.

#### Dynamic Lookups vs CLI flags, Environment Variables

It was not the intention to have dynamic lookups but it turns out to be necessary as, for example, tests in multi-output recipes are unable to be directed as to which native compilation environment should be used.

`conda activate {env}` could use a CLI flag or an environment variable reasonably easily although there is the risk that the user's choice could conflict with the conda environment's actual state.

#### Unix to Native Prefixes

In practice, `conda activate|deactivate` supply the code with a Windows native pathname, eg. `C:\path\to\prefix` however during a build, the supplied `{prefix}` is a Unix-like pathname, eg. `/c/path/to/prefix`.  The code catches both MSYS2-style Unix pathnames, `/c/...`, and Cygwin-style Unix pathnames, `/cygdrive/c/...`.

Ostensibly this shouldn't matter except that the Python `os.isdir()` function only works for Windows native pathnames.

Consequently the previously existing `native_path_to_unix()` has been mirrored into a `unix_path_to_native()` function.

### Checklist - did you ...


ifitchet:
Should I renumber the news article to reflect the PR#?

beeankha:
@conda-bot check

beeankha:
> Should I renumber the news article to reflect the PR#?

Yes, that's fine; alternatively, the number in the name of the news file can be the number associated with the related GitHub issue.

news/13999-msys2-environments
```diff
@@ -0,0 +1,17 @@
...
+  Windows native compilers and libraries.  `mingw-w64` will continue
+  to be added to the user's `PATH` as a fallback location.
+
+  This allows for multiple MSYS2 environments to be accommodated: for
+  example either the GCC suite using either UCRTBASE or MSVCRT or the
+  LLVM suite using UCRTBASE.
+
+  New MSYS2 packages can now use the upstream installation prefixes,
+  `{prefix}/ucrt64`, `{prefix}/clang64`, etc. and will be found and
+  used in preference automatically.
```
beeankha, line 16:
```suggestion
  used in preference automatically. (#13649)
```

In addition to the suggested change above, I was wondering if there is any way that this news file entry could get reduced to one or two sentences with no line/paragraph breaks? Please see items in the [release changelog notes](https://github.com/conda/conda/releases) for reference/examples; these news files get aggregated into the changelog during each release cycle so we want to be as concise as possible while retaining a lot of detail, which can certainly be challenging sometimes! 😆

beeankha:
pre-commit.ci autofix

tests/test_activate.py
```diff
@@ -3099,3 +3099,108 @@ def test_activate_and_deactivate_for_uninitialized_env(conda_cli):
...
+        mkdir_p(join(td, "conda-meta"))
+        activate_d_dir = mkdir_p(join(td, "etc", "conda", "activate.d"))
+        activate_d_1 = join(activate_d_dir, "see-me.sh")
+        activate_d_2 = join(activate_d_dir, "dont-see-me.bat")
+        touch(join(activate_d_1))
+        touch(join(activate_d_2))
+
+        activate_env_vars = join(td, PREFIX_STATE_FILE)
+        with open(activate_env_vars, "w") as f:
+            f.write(ENV_VARS_FILE)
```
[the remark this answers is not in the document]
beeankha: Where is `ENV_VARS_FILE` defined?

tests/test_activate.py
```diff
@@ -3099,3 +3099,110 @@ def test_activate_and_deactivate_for_uninitialized_env(conda_cli):
...
+
+        write_pkg_env_vars(td)
```
beeankha, line 3124: Should this be [`write_pkgs(td)`](https://github.com/conda/conda/blob/main/tests/test_activate.py#L186) or something else?

tests/test_activate.py
```diff
@@ -3099,3 +3099,110 @@ def test_activate_and_deactivate_for_uninitialized_env(conda_cli):
...
+                exp_dir = join(library_dir, dir, "bin")
+                assert ce_act.path_conversion(exp_dir) in paths
+
+        if no_paths:
+            for dir in no_paths:
+                exp_dir = join(library_dir, dir, "bin")
+                assert ce_act.path_conversion(exp_dir) not in paths
+
+        ps_act = PowerShellActivator()
+        builder = ps_act.build_activate(td)
```
beeankha, line 3149: This instance of the `builder` variable doesn't seem to be utilized, should this line be changed like what's shown below? 
```suggestion
        ps_act.build_activate(td)
```

conda/activate.py
```diff
@@ -894,6 +924,68 @@ def _translation(match):
         return tuple(unix_path.split(":"))
 
 
+def unix_path_to_native(
```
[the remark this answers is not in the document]
ifitchet: There might well be but I was looking to create a reflection of the `native_path_to_unix()` function immediately above to perform the reverse function.
`native_path_to_unix()` is used several times in `activate.py` but it would appear that no-one previously needed the reverse.

FWIW, in both cases, I believe the functionality is, in practice, implemented by calling the Cygwin/MSYS2 command `cygpath`, with `-u`/`-w` flags as necessary, to massage the names appropriately leaving the "do it in Python" parts of the functions unused/untested.

conda/activate.py
```diff
@@ -894,6 +924,68 @@ def _translation(match):
...
+    elif not on_win:
+        return path_identity(paths)
+
+    # short-circuit if we don't get any paths
+    paths = paths if isinstance(paths, str) else tuple(paths)
+    if not paths:
+        return "." if isinstance(paths, str) else ()
+
+    # on windows, uses cygpath to convert posix paths to windows native paths
+    from shutil import which
```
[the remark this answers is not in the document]
ifitchet: As noted above, this is a reflection of the existing `native_path_to_unix()` function.

When making one's first forays into someone else's legacy code I always abide by the maxim: "don't try to fix their tech debt."

conda/activate.py
```diff
@@ -894,6 +924,68 @@ def _translation(match):
...
+                found_path.group(1)
+                .replace("/", "\\")
+                .replace("", ":")
+                .replace("/", "//")
+            )
+            return "\\" + found.rstrip("\\")
+
+        win_path = (
+            re.sub(
+                r"([a-zA-Z]:[\/\\\\]+(?:[^:*?\"<>|;]+[\/\\\\]*)*)",
```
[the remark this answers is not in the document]
ifitchet: See above, not on this pass.

tests/test_activate.py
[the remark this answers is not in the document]
ifitchet: I was standing on the shoulders of giants... before they moved.

`ENV_VARS_FILE` was last seen in `test_activate.py` in 7679aed181 before that code was rewritten in 298ca49ff0 (12 days ago).
I guess I'll need to revisit.

As for mocking it out @schuylermartin45 I was following in the style of existing tests which created physical environments.  One of the key issues with these MSYS2 tests is that we want to create physical directories that the code can _stat(2)_ to determine which MSYS2 environment is present.

tests/test_activate.py
ifitchet: See the previous comment about the code changing "a bit" in the last few weeks.  I'll need to revisit.

tests/test_activate.py
ifitchet: To be fair to myself, I just cut'n'paste this code from elsewhere at the time.
It looks like Ken has been changing things around a bit recently.
I'll revisit.

ifitchet: `builder` wasn't used in the previous variant (I cut'n'paste) but it is now.  I've reworked this code to follow in the same style as the new code.

tests/test_activate.py
```diff
@@ -3099,3 +3099,138 @@ def test_activate_and_deactivate_for_uninitialized_env(conda_cli):
...
+
+        if exp_paths:
+            for dir in exp_paths:
+                exp_dir = join(library_dir, dir, "bin")
+                assert ce_act.path_conversion(exp_dir) in paths
+
+        if no_paths:
+            for dir in no_paths:
+                exp_dir = join(library_dir, dir, "bin")
+                assert ce_act.path_conversion(exp_dir) not in paths
```
kenodegard, line 3158: @ifitchet is this the only part of the test you actually care about? everything above this appears to just be initialization?

ifitchet: @kenodegard The tests want to:

1. create a physical environment because
2. they will create any (optional) `create_dirs` before
3. activating the environment (and the whole point is that the environment has different PATHs because of (2))
4. check for the things we (now) expect `exp_paths`
5. check for the absence of things we shouldn't have `no_paths`

Other than that, the rest of it is cut'n'paste but I don't know how much of that is _required_ to bootstrap a physical environment in the test code.  I'm all for cutting it back if we can.

conda/activate.py
```diff
@@ -618,6 +618,42 @@ def _get_starting_path_list(self):
...
+                    f"WARNING: {prefix}: {variants} MSYS2 envs exist: please check your dependencies",
+                    file=sys.stderr,
+                )
+                print(
+                    f"WARNING: conda list -n {self._default_env(prefix)}",
+                    file=sys.stderr,
+                )
+
+            if msystem:
+                yield self.sep.join((prefix, "Library", msystem, "bin"))
```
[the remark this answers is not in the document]
kenodegard: see https://github.com/conda/conda/pull/13649/commits/32cb5e54fb6a6992b35527f5058f5f3321a3b1a0

conda/activate.py
```diff
@@ -894,6 +938,76 @@ def _translation(match):
...
+            )
+            return "\\" + found.rstrip("\\")
+
+        win_path = (
+            re.sub(
+                r"([a-zA-Z]:[\/\\]+(?:[^:*?\"<>|;]+[\/\\\\]*)*)", _translation, joined
+            )
+            .replace(":/", ";/")
+            .rstrip(";")
+        )
```
kenodegard, line 998: This fallback logic is incorrect, as written it is converting from Windows to Unix, not Unix to Windows

ifitchet: Uh, three things:

1. it should be the reflection of the same parts of `native_path_to_unix()` immediately above
2. I reworked it without thinking hard so you could well be right
3. I don't think any of it is called in practice as `cygpath` is used

Do we have any tests for "no cygpath on PATH"?

kenodegard: I have previously gone down the whole path of surely `cygpath` is always available and that was a completely wrong assumption, see https://github.com/conda/conda/pull/12874

ifitchet: Oops and Hmm.
But we don't run any tests that handle the missing cygwin case (well, not from what I could see, otherwise I would have tested the code ;) )

In the meanwhile, yes, the regex on line 994 looks suspiciously similar to its counterpart in the other function.

ifitchet: I'm wondering if we should pull both of these fallback clauses out into standalone functions which we can, at least, test fairly easily even if they're only called in relatively rare circumstances.

conda/activate.py
```diff
@@ -618,6 +618,38 @@ def _get_starting_path_list(self):
...
+                # cygwin /cygdrive/c/
+                if re.match("^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*", prefix):
+                    path = unix_path_to_native(path)
+
+                if isdir(path):
+                    variants.append(variant)
+
+            if len(variants) > 1:
+                print(
+                    f"WARNING: {prefix}: {variants} MSYS2 envs exist: please check your dependencies",
```
skupr-anaconda, line 642: @ifitchet Would it be better for end-users to see an example of wrong dependencies in the environment? When I saw this warning message I couldn't get what it means

conda/activate.py
kenodegard: I've already updated the tests in this PR to test both the `cygpath` and fallback codes but we def can split the fallback out if that will make things easier

conda/activate.py
ifitchet: @skupr-anaconda probably.

However, I don't know how to extract such information.  What I do have is the list of possible variants in my hands so at least I can give _some_ clues as to why they won't necessarily see all the packages they have installed on their PATH.

If, though, that is confusing, then it is less useful.

conda/activate.py
ifitchet: Hmm, some thinking out loud:

- `/x/here/there` is MSYS2-speak for `X:\here\there`
- FYI `/cygdrive/x/here/there` is Cygwin-speak for `X:\here\there`
- `/here/there` is otherwise relative to the MSYS2 root: `{prefix}\Library\here\there`

Luckily, the calling code does have `{prefix}` to hand so at least it can be passed to `unix_path_to_native()`

ifitchet: Hmm, I hadn't noticed (because I had no need to) that the MSYS2 and Cygwin `cygpath` programs are not mutually aware and will treat the other's form as a regular root filesystem path, ie. prepend Windows path to Unix root:

MSYS2:
```
(devenv-3.10-c) C:\Users\IanFitchet\src\ifitchet\conda>cygpath --path -w /x/here/there
X:\here\there

(devenv-3.10-c) C:\Users\IanFitchet\src\ifitchet\conda>cygpath --path -w /cygdrive/x/here/there
C:\Users\IanFitchet\src\ifitchet\conda\devenv\Windows\envs\devenv-3.10-c\Library\cygdrive\x\here\there

```

Cygwin:
```
dev-admin@EC2AMAZ-RNGBGN7 ~
$ cygpath --path -w /cygdrive/x/here/there
X:\here\there

dev-admin@EC2AMAZ-RNGBGN7 ~
$ cygpath --path -w /x/here/there
C:\cygwin64\x\here\there

```

The upshot of this is that, if we are going to test `cygpath` and the fallback code in `unix_path_to_native()` in the same function then are either testing against an MSYS2 `cygpath` or a Cygwin `cygpath` but we can't test against both (unless you can give `pytest` multiple possible answers).

We _could_ separate out the `cygpath` parts in one test and limit it to MSYS2 (as that's what the CI installs?) and have the fallback code tested for either MSYS2 or Cygwin variants in another test.

Or, as I'm going to do, comment out the Cygwin test variants for now and leave them as food for thought for a later visit.

tests/test_activate.py
```diff
@@ -1105,40 +1107,186 @@ def make_dot_d_files(prefix, extension):
...
+#
+# The only tests where *_prefix are used are root filesystem ones.
+windows_prefix = context.target_prefix
+if windows_prefix.startswith("/"):
+    # toggle the prefixes
+    unix_prefix = windows_prefix
+    # yes, this is inception
+    windows_prefix = unix_path_to_native(unix_prefix, unix_prefix)
+else:
+    unix_prefix = native_path_to_unix(windows_prefix)
```
ifitchet, line 1176: @kenodegard I suspect this block (and the commentary) should be in a `if on_win:` clause as much because the tests are in a `skipif (not on_win, ...)` decorator as anything else.

conda/activate.py
```diff
@@ -618,6 +618,38 @@ def _get_starting_path_list(self):
...
+            # mingw-w64 is a legacy variant used by m2w64-* packages
+            #
+            # We could include clang32 and mingw32 variants
+            variants = []
+            for variant in ["ucrt64", "clang64", "mingw64", "clangarm64"]:
+                path = self.sep.join((prefix, "Library", variant))
+
+                # MSYS2 /c/
+                # cygwin /cygdrive/c/
+                if re.match("^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*", prefix):
```
kenodegard, line 634: will it only ever be a windows drive path? or could we also say:

```suggestion
                if "/" in prefix:
```

conda/activate.py
```diff
@@ -860,25 +892,37 @@ def native_path_to_unix(
...
         unix_path = (
-            re.sub(r"([a-zA-Z]:[\/\\]+(?:[^:*?\"<>|;]+[\/\\]*)*)", _translation, joined)
+            re.sub(
+                r"""
+                (?P<drive>[a-z]:)?
+                (?P<path>[\/\\]+(?:[^:*?\"<>|;]+[\/\\]*)*)
+                """,
+                _translation,
+                joined,
+                flags=re.VERBOSE | re.IGNORECASE,
```
jezdez, line 924: Oh, huh, `re.IGNORECASE`, I thought Cygwin is case-sensitive, could this be an incompatibility perhaps?

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
...
+        return path_identity(paths)
+
+    # short-circuit if we don't get any paths
+    paths = paths if isinstance(paths, str) else tuple(paths)
+    if not paths:
+        return "." if isinstance(paths, str) else ()
+
+    # on windows, uses cygpath to convert posix paths to windows native paths
+    from shutil import which
+    from subprocess import run
```
jezdez, line 956: Let's move these to the top of the file, they'll be imported in `sys.modules` already.

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
...
+        # if present, use cygpath to convert paths since its more reliable
+        win_path = run(
+            [cygpath, "--windows", "--path", joined],
+            text=True,
+            capture_output=True,
+            check=True,
+        ).stdout.strip()
+    except FileNotFoundError:
+        log.warning(
+            "cygpath not available, falling back to less reliable path conversion"
```
jezdez, line 978:
```suggestion
            "cygpath is not available, falling back to manual path conversion"
```

tests/test_activate.py
```diff
@@ -1105,40 +1107,292 @@ def make_dot_d_files(prefix, extension):
     touch(join(prefix, "etc", "conda", "deactivate.d", f"deactivate1{extension}"))
 
 
+@pytest.mark.skipif(
+    not on_win,
```
marcoesters, line 1111: Shouldn't this be run outside of Windows then to make sure that this is the case?

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
...
+                return elem
+
+            # continue performing substitutions until a match is found
+            elem, subs = RE_DRIVE.subn(_translation_drive, elem)
+            if not subs:
+                elem, subs = RE_MOUNT.subn(_translation_mount, elem)
+            if not subs:
+                elem, subs = RE_ROOT.subn(_translation_root, elem)
+
+            return re.sub(r"/+", r"\\", elem)
```
jezdez, line 1057: It'd be good to not define those functions dynamically on every run of `unix_path_to_native` and let the patterns compile on module level to reuse the values on consecutive runs of `unix_path_to_native`.

You could also simply turn `unix_path_to_native` into a `UnixToNativeConverter` class with the translation functions as instance methods, if you'd like to encapsulate the functionality.

conda/activate.py
```diff
@@ -860,25 +892,37 @@ def native_path_to_unix(
...
+                r"""
+                (?P<drive>[a-z]:)?
+                (?P<path>[\/\\]+(?:[^:*?\"<>|;]+[\/\\]*)*)
+                """,
+                _translation,
+                joined,
+                flags=re.VERBOSE | re.IGNORECASE,
+            )
             .replace(";", ":")
             .rstrip(";")
```
jezdez, line 927:
```suggestion
            .replace(ntpath.pathsep, posixpath.pathsep)
            .rstrip(ntpath.pathsep)
```

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
         return tuple(unix_path.split(":"))
```
jezdez, line 938:
```suggestion
        return tuple(unix_path.split(posixpath.pathsep))
```

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
...
+
+    # It is very easy to end up with a bash in one place and a cygpath in another due to e.g.
+    # using upstream MSYS2 bash, but with a conda env that does not have bash but does have
+    # cygpath.  When this happens, we have two different virtual POSIX machines, rooted at
+    # different points in the Windows filesystem.  We do our path conversions with one and
+    # expect the results to work with the other.  It does not.
+
+    bash = which("bash")
+    cygpath = str(Path(bash).parent / "cygpath") if bash else "cygpath"
+    joined = paths if isinstance(paths, str) else ":".join(paths)
```
jezdez, line 966:
```suggestion
    joined = paths if isinstance(paths, str) else posixpath.pathsep.join(paths)
```

conda/activate.py
```diff
@@ -894,6 +938,140 @@ def _translation(match):
...
+                elem, subs = RE_MOUNT.subn(_translation_mount, elem)
+            if not subs:
+                elem, subs = RE_ROOT.subn(_translation_root, elem)
+
+            return re.sub(r"/+", r"\\", elem)
+
+        # The conda prefix can be in a drive letter form
+        prefix = translate(prefix)
+
+        win_path = ";".join(map(translate, joined.split(":")))
```
jezdez, line 1062:
```suggestion
        win_path = ntpath.pathsep.join(map(translate, joined.split(posixpath.pathsep)))
```

jezdez (commented):
Minor nits about nesting in conversion function and suggestions, nothing major though. Nice to see so many tests! 

I do wonder if a) there is prior art elsewhere we could verify this with b) there is a best practice from MS we could get inspired from, I wonder if they've had to account for Cygwin when they rolled out WSL.

tests/test_activate.py
```diff
@@ -3099,3 +3353,87 @@ def test_activate_and_deactivate_for_uninitialized_env(conda_cli):
...
+    activator_cls: type[_Activator],
+) -> None:
+    with tmp_env() as prefix:
+        # create MSYS2 directories
+        (library := prefix / "Library").mkdir()
+        for path in create:
+            (library / path / "bin").mkdir(parents=True)
+
+        activator = activator_cls()
+        paths = activator._replace_prefix_in_path(str(prefix), str(prefix))
```
marcoesters, line 3431: Is this a roundabout way to get the directories in `PATH` because there is no getter function like that in the activator?

conda/activate.py
marcoesters: The skeleton of the two functions is essentially the same. The biggest differences I see are:

* `cygpath` uses `--windows` instead of `--unix`
* The regexes are different
* The pathsep is different.

The functions could be consolidated with a `target` input that determines what the target nomenclature ought to be. That may be out of scope for this PR though.

conda/activate.py
kenodegard: The `re.IGNORECASE` only impacts the drive letter matching here, updated to use `[A-Za-z]` instead.

conda/activate.py
kenodegard: Done, see https://github.com/conda/conda/pull/13649/commits/4d98451d83de3f5f3ba1447dfcaa4a0d4630e6d4

tests/test_activate.py
kenodegard: I agree this isn't great but fixing it is out of scope for this PR. Alternatively, we'd do something like:

```suggestion
        paths = (
            *activator._get_path_dirs(prefix),
            *activator.path_conversion(activator._get_starting_path_list())
        )
```

So IMO one "private" method is better than two of them.

tests/test_activate.py
kenodegard: On Windows the `native_path_to_unix` and `unix_path_to_native` functions will short-circuit to the `path_identity` function instead which won't error, they will simply return gibberish for the input values.

conda/activate.py
```diff
@@ -618,6 +618,38 @@ def _get_starting_path_list(self):
     def _get_path_dirs(self, prefix):
         if on_win:  # pragma: unix no cover
             yield prefix.rstrip("\\")
+
+            # We need to stat(2) for possible environments because
+            # tests can't be told where to look!
+            #
+            # mingw-w64 is a legacy variant used by m2w64-* packages
+            #
+            # We could include clang32 and mingw32 variants
```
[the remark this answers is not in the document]
kenodegard: @ifitchet?

jezdez (approved):

conda/activate.py
ifitchet: @kenodegard to me that seems risky.  Can I create an expecting-to-be-Windows conda environment with `/` in the name?  Probably: `md here/there\conda-meta` (or whatever the minimum conda env is).
That said, anyone who mixes `/` and `\` in the same tree _and_ is using MSYS2 is probably in for a world of pain for other reasons.
I might think that `prefix.startswith("/")` would be OK but can you create an environment that _doesn't_ start at the filesystem root?  Can I `conda activate here\there` and what will the value of `prefix` be, will it always be resolved to the filesystem root?

conda/activate.py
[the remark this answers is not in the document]
ifitchet: Erm, good question.
I limited it to 64-bit variants as "we" don't built 32-bit variants of any packages any more.
The choice of "we" and 64-bit only is a little parochial and may not cover the wider conda community's needs.

The cost will be two extra (filename munge + _stat(2)_) for everyone on Windows each time this function is called (`conda activate|deactivate|build`).

conda/activate.py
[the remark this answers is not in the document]
ifitchet: @jezdez We don't need the match, we're just looking to identify MSYS2 or Cygwin pathnames hence the grouping for alternates.
I don't know enough the the detail of Python's `re` to know if `(?:...)` is a saving.

ifitchet:
@jezdez 

> I wonder if they've had to account for Cygwin when they rolled out WSL

Not really, they have another spin on `cygpath`:

```
ifitchet@CPC-ifitc-2HB9V:~$ wslpath
wslpath: Invalid argument
Usage:
    -a    force result to absolute path format
    -u    translate from a Windows path to a WSL path (default)
    -w    translate from a WSL path to a Windows path
    -m    translate from a WSL path to a Windows path, with '/' instead of '\'

EX: wslpath 'c:\users'
```

and, FWIW, they use `/mnt` for Windows drives (and other things):

```
ifitchet@CPC-ifitc-2HB9V:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
...
C:\             128G  115G   13G  90% /mnt/c
...
```

(on an Ubuntu/WSL2 instance at any rate)

marcoesters (approved):

PR #13651, opened 2024-03-05
Add `--json` to conda notices

ForgottenProgramme:
### Description

Add `--json` flag to `conda notices`.
This will flag will lead to all notices being printed in json format.

### Checklist - did you ...


conda/notices/fetch.py
```diff
@@ -38,7 +39,7 @@ def get_notice_responses(
     """
     executor = ThreadPoolExecutor(max_workers=max_workers)
 
-    with Spinner("Retrieving notices", enabled=not silent):
+    with Spinner("Retrieving notices", enabled=not silent, json=context.json):
```
ForgottenProgramme, line 42: To ensure the "retrieving notices" prefix and the progress bar doesn't show up with `--json`.

conda/notices/views.py
```diff
@@ -15,17 +17,35 @@ def print_notices(channel_notices: Sequence[ChannelNotice]):
...
+                    "channel_name": channel_notice.channel_name,
+                    "message": channel_notice.message,
+                    "level": channel_notice.level.name,
+                    "created_at": str(channel_notice.created_at),
+                    "expired_at": str(channel_notice.expired_at),
+                    "interval": channel_notice.interval,
+                }
+                for channel_notice in channel_notices
+            ],
+            indent=2,
```
ForgottenProgramme, line 34: `indent=2`
So that the output is more readable

conda/notices/views.py
```diff
@@ -15,17 +17,34 @@ def print_notices(channel_notices: Sequence[ChannelNotice]):
...
+            [
+                {
+                    "id": channel_notice.id,
+                    "channel_name": channel_notice.channel_name,
+                    "message": channel_notice.message,
+                    "level": channel_notice.level.name.lower(),
+                    "created_at": channel_notice.created_at.isoformat(),
+                    "expired_at": channel_notice.expired_at.isoformat(),
+                    "interval": channel_notice.interval,
+                }
```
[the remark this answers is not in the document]
ForgottenProgramme: I added it. 
Once reviewed and approved, I will delete the commented code.

beeankha (dismissed):

conda/notices/views.py
```diff
@@ -15,17 +17,27 @@ def print_notices(channel_notices: Sequence[ChannelNotice]):
     """
     current_channel = None
 
-    for channel_notice in channel_notices:
-        if current_channel != channel_notice.channel_name:
+    if context.json:
+        json_output = json.dumps(
+            [
+                ChannelNotice.notice_to_json_notice(channel_notice)
```
jezdez, line 23: This isn't a correct use of the instance method you defined above, you're passing in an instance, while all that would be needed is calling it on the instance you already have here:
```suggestion
                channel_notice.to_dict()
```

conda/notices/types.py
```diff
@@ -28,6 +28,17 @@ class ChannelNotice(NamedTuple):
     expired_at: datetime | None
     interval: int | None
 
+    def notice_to_json_notice(self):
```
jezdez, line 31: A standard name for such a method is `to_dict` or `for_json`. We have many classes that implement `to_dict`, so I would suggest to use that.
```suggestion
    def to_dict(self):
```

tests/cli/test_main_notices.py
```diff
@@ -71,6 +73,44 @@ def test_main_notices(
...
+        return_value=(channel_str,),
+    )
+    args, parser = conda_notices_args_n_parser
+    messages = ("Test One", "Test Two")
+    messages_json = get_test_notices(messages)
+    json_list = messages_json.get("notices")
+
+    for item in json_list:
+        item.update({"channel_name": channel_str})
+        item.update({"interval": "null"})
```
jezdez, line 98:
```suggestion
        item.update({"channel_name": channel_str, "interval": "null"})
```

jezdez (changes requested):

tests/cli/test_main_notices.py
```diff
@@ -71,6 +73,43 @@ def test_main_notices(
...
+        new_callable=mocker.PropertyMock,
+        return_value=(channel_str,),
+    )
+    args, parser = conda_notices_args_n_parser
+    messages = ("Test One", "Test Two")
+    messages_json = get_test_notices(messages)
+    json_list = messages_json.get("notices")
+
+    for item in json_list:
+        item.update({"channel_name": channel_str, "interval": "null"})
```
kenodegard, line 97: Would it be better to add this to `get_test_notices`? cc @travishathaway

ForgottenProgramme: I think Travis didnt want to add this to the `get_test_notices` func itself because then we might need to change the other funcs where `get_test_notices` is being called.

kenodegard: I agree that `channel_name` isn't part of the API endpoint but isn't `interval` included in `notices.json`? https://github.com/conda/conda/blob/078e7ee79381060217e1ec7f9b0e9cf80ecc8f3f/conda/notices/types.py#L56-L69

travishathaway: @kenodegard,

No, it's not: https://github.com/conda-incubator/ceps/blob/main/cep-6.md#what-file-format-will-this-message-have-and-what-will-it-contain

I think I was originally planning to have it there and must have forgotten to remove it from conda 🙃.

travishathaway: I could add removing (deprecating) it as an item for this issue:

- https://github.com/conda/conda/issues/13652

It would make sense to have that change there as we may wish to refactor some of that code anyways.

kenodegard (approved):

jezdez (approved):

beeankha (approved):

PR #13663, opened 2024-03-06
Upload a release archive during release.

### LINKED ISSUE #13399: Publish stable artifacts with checksums during release
### Checklist

### What is the idea?

Currently GH provides autogenerated archives with releases. However these have the downside of having unstable checksums. As a result, when they are used as a source in package builds (like in conda-forge), they may pass at one point in time and later fail due to checksum mismatches

Given this, wonder if we can consider a different option where an artifact is generated and uploaded as part of the release process with a checksum. That way consumers of these artifacts will know the artifacts are static and have a checksum they can count on to verify those artifacts

### Why is this needed?

Would improve downstream packaging experience by providing better reliability

### What should happen?

Am a little unsure what the current release process looks like. So what should be done will depend a bit on how that release process is run

Do see that we have [a Rever file]( https://github.com/conda/conda/blob/08a19a3d84c024b8a8c3fdc0e5246cff3175d0c5/rever.xsh#L1 ). If that is what we are using, we could specify [`$GHRELEASE_ASSETS` (like in `conda-smithy`)]( https://github.com/conda-forge/conda-smithy/blob/7a6269113a525205dbef32140280f14d74479c8d/rever.xsh#L19 )

If we are not using Rever, maybe we could use [a GH Actions step to upload artifacts]( https://github.com/actions/upload-artifact )

There might be other reasonable choices depending on what fits best in our release process

### Additional Context

Recently ran into this when releasing 23.11.0 ( https://github.com/conda-forge/conda-feedstock/pull/228#issuecomment-1839685975 ). Though this is not the first time we have seen this issue with GH autogenerated artifacts

jezdez:
### Description

Closes #13399.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -1,0 +1,52 @@
+name: Upload release
+
+on:
+  # https://docs.github.com/en/webhooks/webhook-events-and-payloads#release
+  release:
+    types: [published]
+
+concurrency:
+  # Concurrency group that uses the workflow name and PR number if available
+  # or commit SHA as a fallback. If a new build is triggered under that
+  # concurrency group while a previous build is running it will be canceled.
+  # Repeated pushes to a PR will cancel all previous builds, while multiple
+  # merges to main will not cancel.
+  group: ${{ github.workflow }}-${{ github.ref_name || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  # create compressed tarballs and upload them to the published release
+  upload:
+    if: >-
+      !cancelled()
+      && !github.event.repository.fork
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        # test lower version (w/ defaults) and upper version (w/ defaults and conda-forge)
+        file-format: ['gz', 'bz2']
+    env:
+      FILE_NAME: "${{ github.event.repository.name }}-${{ github.ref_name }}.tar.${{ matrix.file-format }}"
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v4
+
+      - name: Gzip Source
+        if: matrix.file-format == 'gz'
+        run: tar -cfzv ${{ env.FILE_NAME }} ${{ github.workspace }}
+
+      - name: Bzip2 Source
+        if: matrix.file-format == 'bz2'
+        run: tar -cfjv ${{ env.FILE_NAME }} ${{ github.workspace }}
+
+      - name: Run sha256sum
+        run: sha256sum ${{ env.FILE_NAME }} > "${{ env.FILE_NAME }}.sha256sum"
+
+      - name: Upload Release
+        run: gh release
+          upload
+          --clobber "${{ github.ref_name }}"
+          --repo "${{ github.repository }}"
+          "${{ env.FILE_NAME }}"
+          "${{ env.FILE_NAME }}.sha256sum"
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -27,39 +27,22 @@ jobs:
     strategy:
       fail-fast: true
     env:
-      REPO_NAME: "${{ github.event.repository.name }}"
+      ARCHIVE_NAME: "${{ github.event.repository.name }}-${{ github.ref_name }}"
     steps:
       - name: Checkout Source
         uses: actions/checkout@v4
 
       - name: Archive Source
         run: |
-          mkdir -p "$RUNNER_TEMP/release"
+          mkdir -p release
+          git archive --prefix="${{ env.ARCHIVE_NAME }}/" --output "release/${{ env.ARCHIVE_NAME }}.tar.gz" HEAD
+          sha256sum "release/${{ env.ARCHIVE_NAME }}.tar.gz" | awk '{print $1}' > "release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum"
 
-          echo ::group::Creating gzip archive
-          export ARCHIVE_NAME="$RUNNER_TEMP/release/$REPO_NAME-$GITHUB_REF_NAME.tar"
-          tar \
-          --dereference --hard-dereference \
-          --directory "$GITHUB_WORKSPACE" \
-          -cvzf "$ARCHIVE_NAME.gz" \
-          .
-          sha256sum "$ARCHIVE_NAME.gz" | awk '{print $1}' > "$ARCHIVE_NAME.gz.sha256sum"
-          echo ::endgroup::
-
-          echo ::group::Creating bz2 archive
-          tar \
-          --dereference --hard-dereference \
-          --directory "$GITHUB_WORKSPACE" \
-          -cvjf "$ARCHIVE_NAME.bz2" \
-          .
-          sha256sum "$ARCHIVE_NAME.bz2" | awk '{print $1}' > "$ARCHIVE_NAME.bz2.sha256sum"
-          echo ::endgroup::
-
-      - name: Upload Archives
+      - name: Upload Archive
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           gh release upload \
-          --clobber "$GITHUB_REF_NAME" \
-          --repo "$GITHUB_REPOSITORY" \
-          $RUNNER_TEMP/release/*
+          --clobber "${{ github.ref_name }}" \
+          --repo "${{ github.repository }}" \
+          release/*
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -21,7 +21,7 @@ jobs:
   # create source archive and upload it to the published release
   # URL to the archive: https://github.com/conda/<repo>/releases/download/<tag>/<repo>-<tag>.tar.gz
   upload:
-    if: !github.event.repository.fork
+    if: '!github.event.repository.fork'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: true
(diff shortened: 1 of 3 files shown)


.github/workflows/upload.yml
```diff
@@ -0,0 +1,52 @@
...
+  # create compressed tarballs and upload them to the published release
+  upload:
+    if: >-
+      !cancelled()
+      && !github.event.repository.fork
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        # test lower version (w/ defaults) and upper version (w/ defaults and conda-forge)
```
dbast, line 27:
```suggestion
```

.github/workflows/upload.yml
```diff
@@ -0,0 +1,52 @@
...
+        file-format: ['gz', 'bz2']
```
dbast, line 28: are both formats required?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,52 @@
...
+    env:
+      FILE_NAME: "${{ github.event.repository.name }}-${{ github.ref_name }}.tar.${{ matrix.file-format }}"
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v4
+
+      - name: Gzip Source
+        if: matrix.file-format == 'gz'
+        run: tar -cfzv ${{ env.FILE_NAME }} ${{ github.workspace }}
```
dbast, line 37: this has the risk to compress things you don't want, this should be better (this can also run for both archive types instead of a matrix):
```suggestion
        run: git archive HEAD --output ${{ env.FILE_NAME }}
```

.github/workflows/upload.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Bzip2 Source
+        if: matrix.file-format == 'bz2'
+        run: tar -cfjv ${{ env.FILE_NAME }} ${{ github.workspace }}
```
dbast, line 41: see above

.github/workflows/upload.yml
jezdez: Oh you were faster with the review than expected, I've played with it a little more, I'm not sure what you may not want in the tarballs for the release artifacts?

.github/workflows/upload.yml
jezdez: I just settled on tgz only.

.github/workflows/upload.yml
dbast: reasons for not using tar:
* tar is not reproducible (depends on order in filesystem, locale, timezone, timestamps)
* if the workspace is dirty (e.g. when testing the command locally) then that dirty stuff gets also compressed (would require running `git clean -fdx && git reset --hard` before)... not an issue directly after checkout in that workflow

git archive was made just for the reason to create reproducible fs/os independent archives only compressing what exactly belongs to a commit.

jezdez: Yeah, I knew about this, thanks for reminding me again

dbast (dismissed):

jezdez:
@kenodegard I could also copy this over to conda/infra so we can sync it with conda and conda-build centrally. Any concerns with that? Should we first try it out here?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,48 @@
...
+
+permissions:
+  contents: write
+
+jobs:
+  # create compressed tarballs and upload them to the published release
+  upload:
+    if: >-
+      !cancelled()
+      && !github.event.repository.fork
```
kenodegard, line 25:
```suggestion
    if: "!github.event.repository.fork"
```

.github/workflows/upload.yml
```diff
@@ -0,0 +1,48 @@
...
+          sha256sum "release/${{ env.ARCHIVE_NAME }}.tar.gz" | awk '{print $1}' > "release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum"
+
+      - name: Upload Archive
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release upload \
+          --clobber "${{ github.ref_name }}" \
+          --repo "${{ github.repository }}" \
+          release/*
```
kenodegard, line 48: We can use yaml string concatenation instead:

```suggestion
        run: gh release upload
          --clobber "${{ github.ref_name }}"
          --repo "${{ github.repository }}"
          release/*
```

.github/workflows/upload.yml
```diff
@@ -0,0 +1,48 @@
...
+      ARCHIVE_NAME: "${{ github.event.repository.name }}-${{ github.ref_name }}"
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v4
+
+      - name: Archive Source
+        run: |
+          mkdir -p release
+          git archive --prefix="${{ env.ARCHIVE_NAME }}/" --output "release/${{ env.ARCHIVE_NAME }}.tar.gz" HEAD
+          sha256sum "release/${{ env.ARCHIVE_NAME }}.tar.gz" | awk '{print $1}' > "release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum"
```
kenodegard, line 39: Can we split this up into separate steps, I found it to be much cleaner to do so in the recent CI refactor

```suggestion
      - name: Create Release Directory
        run: mkdir -p release

      - name: Archive Source
        run: git archive
          --prefix="${{ env.ARCHIVE_NAME }}/"
          --output="release/${{ env.ARCHIVE_NAME }}.tar.gz"
          HEAD

      - name: Compute Checksum
        run: sha256sum "release/${{ env.ARCHIVE_NAME }}.tar.gz"
          | awk '{print $1}'
          > "release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum"
```

kenodegard (commented):
Let's try it here first and then move it over?

kenodegard:
Sandbox repo showing this working: https://github.com/conda-sandbox/upload-archive/releases/tag/2.0

Are these the tarball names we want? We'd have to change Anaconda's and conda-forge's recipes to use this new URL instead? Is it possible to override the "Source code (tar.gz)" instead?

Screenshot 2024-03-07 at 11 28 48

jezdez:
> Sandbox repo showing this working: [conda-sandbox/upload-archive@`2.0` (release)](https://github.com/conda-sandbox/upload-archive/releases/tag/2.0)
> 
> Are these the tarball names we want? We'd have to change Anaconda's and conda-forge's recipes to use this new URL instead?

IIRC we used to have them as artifacts before, I can look for the change back in the day tomorrow. Yes, we would update the URLs in the recipes when we do the release in the next few days.

> Is it possible to override the "Source code (tar.gz)" instead?

Not to my knowledge, since that's the automatic git tag based download. We're just trying to remove the stability problem with github tag URLs.

> 
> ```
> https://github.com/conda-sandbox/upload-archive/releases/download/2.0/upload-archive-2.0.tar.gz
> https://github.com/conda-sandbox/upload-archive/archive/refs/tags/2.0.tar.gz
> ```
> 
> Screenshot 2024-03-07 at 11 28 48

kenodegard:
Didn't find any extra assets for any `conda` releases.

On `conda-build` the last extra asset was added in https://github.com/conda/conda-build/releases/tag/3.21.8. This changed when we dropped `gh_release` from `rever.xsh` in https://github.com/conda/conda-build/pull/4483/files#diff-ac0e1c1da42964b557dbc025c5e788bae7ab7cac5d6454d7fc90c0686d6e48d6L4-L6.

Looking at this old URL I think the url proposed here is perfectly reasonable:

```
https://github.com/conda/conda-build/releases/download/3.21.8/conda-build-3.21.8.tar.gz
```

jezdez:
> Didn't find any extra assets for any `conda` releases.
> 
> On `conda-build` the last extra asset was added in [conda/conda-build@`3.21.8` (release)](https://github.com/conda/conda-build/releases/tag/3.21.8). This changed when we dropped `gh_release` from `rever.xsh` in [conda/conda-build#4483 (files)](https://github.com/conda/conda-build/pull/4483/files#diff-ac0e1c1da42964b557dbc025c5e788bae7ab7cac5d6454d7fc90c0686d6e48d6L4-L6).
> 
> Looking at this old URL I think the url proposed here is perfectly reasonable:
> 
> ```
> https://github.com/conda/conda-build/releases/download/3.21.8/conda-build-3.21.8.tar.gz
> ```

Ah, right, I misremembered, thanks for finding that.

.github/workflows/upload.yml
```diff
@@ -0,0 +1,54 @@
...
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  # create source archive and upload it to the published release
+  # URL to the archive: https://github.com/conda/<repo>/releases/download/<tag>/<repo>-<tag>.tar.gz
+  upload:
+    if: !github.event.repository.fork
```
kenodegard, line 24: we need this to be quoted since it starts with a `!`: https://results.pre-commit.ci/run/github/6235174/1709882115.GVlSQPmNRfmcGfUd4aFBWQ

jezdez: Oh woops, my IDE yelled at me telling me I'm wrong, will undo

jakirkham:
Thanks all! 🙏

Looking forward to taking this for a spin on the next release 😀

mbargull:
Thanks very much for putting this together!
Shall we copy this over to `conda-build` in a separate PR or do we have some general recurring `conda`<-->`conda-build` update process?

kenodegard:
@mbargull we were going to test run this for the March release and assuming all goes well we will move this over to https://github.com/conda/actions and/or https://github.com/conda/infrastructure and sync it back out here and to conda-build

mbargull:
Sounds good, thank you!

PR #13673, opened 2024-03-08
Ensure all supported shells are tested

kenodegard:
### Description

The activation tests are configured to be skipped if the shell in question is not present so we have little oversight on which shells are actually tested by our CI.

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,7 +136,7 @@ flake8-type-checking = {exempt-modules = [], strict = true}
 
 [tool.pytest.ini_options]
 minversion = 3.0
-testpaths = ["tests"]
+testpaths = ["tests/activate"]
 norecursedirs = [
   "tests/data/*",
   "tests/integration/*",
--- a/tests/activate/test_ash.py
+++ b/tests/activate/test_ash.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from shutil import which
+
+import pytest
+
+from conda.common.compat import on_mac, on_win
+
+pytestmark = [
+    pytest.mark.skipif(on_mac, reason="ash is not available on macOS"),
+    pytest.mark.skipif(on_win, reason="ash is not available on Windows"),
+]
+
+
+def test_ash_available():
+    assert which("ash")
--- a/tests/activate/test_bash.py
+++ b/tests/activate/test_bash.py
@@ -1,0 +1,7 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from shutil import which
+
+
+def test_bash_available():
+    assert which("bash")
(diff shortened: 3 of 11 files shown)


tests/shell/__init__.py
```diff
@@ -85,54 +120,37 @@ class InteractiveShellType(type):
...
-            # .. and this one from the `macOS` terminal:
-            # pexpect.exceptions.EOF: End Of File (EOF).
-            # 'args': ('-x',),
-            "init_command": (
-                f'set _CONDA_EXE="{CONDA_PACKAGE_ROOT}/shell/bin/conda"; '
-                f"source {CONDA_PACKAGE_ROOT}/shell/etc/profile.d/conda.csh;"
-            ),
+            # "args": ("-v", "-x"),  # for debugging
+            # unset conda alias before calling conda shell hook
+            "init_command": f'unalias conda;\neval "`{EXE_UNIX} -m conda shell.csh hook {DEV_ARG}`"',
```
kenodegard, line 125: We are forced to do this since the `conda.csh` script actually serves two different purposes. The first time it's called (i.e., `alias conda` is undefined) it defines the `conda` alias. Future calls are treated as actual `conda` commands.

I'm thinking we should solve this by splitting the `conda.csh` script into two scripts, a setup script that defines the alias and a `conda` "function" script.

kenodegard:
ugh the Windows failures on bash are an oooooold issue, we've been ignoring the failures ever since https://github.com/conda/conda/pull/13162

tests/shell/test_posix.py
```diff
@@ -192,33 +182,38 @@ def test_basic_integration(
...
-        sh.sendline(f"conda {install} --blah")
-        sh.assert_env_var("?", "2", use_exact=True)
-        sh.sendline("conda list --blah")
-        sh.assert_env_var("?", "2", use_exact=True)
+        sh.sendline(f'conda {install} --blah || echo "exitcode=$?"')
+        sh.expect_exact("error: unrecognized arguments: --blah")
+        sh.expect_exact("exitcode=2")
+        sh.sendline('conda list --blah || echo "exitcode=$?"')
+        sh.expect_exact("error: unrecognized arguments: --blah")
+        sh.expect_exact("exitcode=2")
```
kenodegard, line 206: Hm so something with `bash` running in the GH runner on Windows causes `echo $?` to not work as a standalone call after running `conda install --blah`, no idea why but doing this instead works...

tests/shell/test_posix.py
```diff
@@ -192,33 +182,38 @@ def test_basic_integration(
         )
         sh.expect(r"Executing transaction: ...working... done.*\n")
         sh.assert_env_var("?", "0", use_exact=True)
+        sh.sendline('cat /proc/$$/status | grep "^Pid:"')
```
kenodegard, line 185:
```suggestion
```

tests/shell/test_posix.py
kenodegard: I would assume GitHub has a post-processor instead of trying to override `echo` `printf` etc.

kenodegard: weirdly an earlier `echo $?` in the test works so 🤷🏼‍♂️

PR #13681, opened 2024-03-11
Use stricter version parsing in `.deprecations`

mbargull:
### Description

Use stricter version parsing in `.deprecations` to ensure `DeprecationHandler._get_version_tuple` only ever parses a subset of what is valid for `packaging.version.parse`.

### Checklist - did you ...

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -22,19 +22,17 @@ class DeprecatedError(RuntimeError):
     pass
 
 
-# inspired by deprecation (https://deprecation.readthedocs.io/en/latest/) and
-# CPython's warnings._deprecated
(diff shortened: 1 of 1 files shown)


jezdez:
This is the wrong way to go about resolving my concerns with the custom version parser that was added in #13568, it just makes it more special. Thanks for trying to accommodate my concerns though!

mbargull:
FWIW, https://github.com/conda/conda/pull/13681/commits/a8adddb09e4b8eb99c0112365b3c0578dcba31da would be how I think we should go about this since it's more logically structured.
(I didn't want to introduce new types etc. in gh-13568 which is why this was added directly to `DeprecationHandler` then.)

PR #13684, opened 2024-03-12
Remove unused setuptools remainings

dbast:
conda once was once packaged by setuptools + wheel extension.

#12509 did the switch to hatch, which has its own wheel builder: https://github.com/pypa/hatch/blob/master/backend/src/hatchling/builders/wheel.py#L160

Thus the wheel package (containing the setuptools wheel extension) and the MANIFEST.in file (describes what setuptools should include) can be removed. The hatch file selection is not based on the MANIFEST.in, see https://hatch.pypa.io/1.9/plugins/builder/wheel/#default-file-selection

### Checklist - did you ...

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,0 @@
-include CHANGELOG.md
-include LICENSE.txt
-include MANIFEST.in
-include README.md
-include setup.cfg
-include setup.py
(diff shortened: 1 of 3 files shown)


dholth (approved):
Thanks for getting rid of MANIFEST.in :)

Issue #12126, opened 2022-11-23
Make conda pip installable via pypi


costrouc:
### What is the idea?

Users should be able to `pip install conda` . Looks like[ in the past](https://pypi.org/project/conda/) this worked and was abandoned around `4.3.16 `

### Why is this needed?

First I wanted to outline some of the reasons that conda should be pypi instalable:
 - other package managers e.g. npm are pip installable
 - conda already supports installing conda within it wouldn't be much different than pip installing conda into an environment
 - conda is also a library and often times I need access to conda.core.*
 - it feels wrong in my setup.py etc. to list all dependencies except conda
 - installation instructions for conda vary by OS/distribution. Why not just use `python -m ensure-pip; pip install conda` this certainly isn't saying that the other install instructions aren't important. setup.py can do anything it likes so we could

Some use cases:
  - using conda as a library (not as a command for interacting with conda internals e.g. package cache) without requiring installing conda via miniconda etc. I've often seen myself needing access to `conda.core.*`
 
My personal primary motivation is the new build system for python https://hatch.pypa.io/latest/. It has great support for plugins into the build, run, environments, publish. All of this is controlled via the `pyproject.toml`. This is an opportunity for conda to share a common file `pyproject.toml`  without conda having to make any changes. 

There is already a plugin https://pypi.org/project/hatch-conda/ which allows using conda environments within the pyproject.toml. But there is one major drawback to this package... it assumes that conda is already installed. It would be nice if conda could be boostrapped via pip so that conda can be effortlessly used in pyproject.toml files.

### What should happen?

The user should be able to 
```
pip install conda
```

jezdez:
Thank you for the issue, repeating what we shortly discussed on the conda Slack:

> I don’t have the full history of why it’s like that, but it’s come up a few time since I’ve worked on conda. I think it’s fundamentally a question of how bootstrapping works. `ensurepip` and `ensureconda` imply that there is already a trustworthy Python installation available on the system to run them. I think there was a design decision in the past that this would not suffice to create a "viable" environment for conda to run in.
>
> In other words, the question isn’t particularly about whether conda can be installed with pip, but whether there is the ability and interest in maintaining conda when it runs in environments that haven’t been bootstrapped with conda itself (think package verification, ABI compatibility etc)

That said, I'm interested in exploring this as an alternative way to integrate the conda stack with the PyPA stack, as I totally hear you on the reasons why this would be useful.

[...]

PR #13690, opened 2024-03-13
Add sdist to release uploads

dbast:
#13663 added the workflow to upload a reproducible release tarball to the release artifacts to not rely on the non-stable Github mechanisms that provide those.

While archiving the entire repo via `git archive` is nice, the real Pythonic artifact is a sdist (Source Distribution) file as also packages upload to pypi.

This PR adds the sdist creation to the upload flow and also changes the sha256sum calculation to produce a combined SHA256SUMS file, as the sdist file name is not something we influence and can rely on in the flow, but is given by hatch and the configuration in the pyproject.toml. So a tagged releases produces a conda-$tag.tar.gz (size 550 KB, name produced by hatch) and conda-$tag_full.tar.gz (size 21 MB, name defined in the flow) file and a combined SHA256SUMS file (huge size difference!). The sdist produced by hatch is also reproducible (according to the hatch documentation).

The sdist file conda-$tag.tar.gz should be complete by concept and should be enough as source file for the feedstocks to build conda and it is the first step, if we ever upload conda to pypi (after all the dependencies are available there). 

(The build command can also easily produce a wheel additionally by adding the `--wheel` arg: `python -m build --sdist --wheel --outdir release`)

### Checklist - did you ...

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Archive Source
         run: git archive
           --prefix="${{ env.ARCHIVE_NAME }}/"
-          --output="release/${{ env.ARCHIVE_NAME }}_full.tar.gz"
+          --output="release/${{ env.ARCHIVE_NAME }}_archive.tar.gz"
           HEAD
 
       - name: Setup Python
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -37,13 +37,24 @@ jobs:
       - name: Archive Source
         run: git archive
           --prefix="${{ env.ARCHIVE_NAME }}/"
-          --output="release/${{ env.ARCHIVE_NAME }}.tar.gz"
+          --output="release/${{ env.ARCHIVE_NAME }}_full.tar.gz"
           HEAD
 
-      - name: Compute Checksum
-        run: sha256sum "release/${{ env.ARCHIVE_NAME }}.tar.gz"
-          | awk '{print $1}'
-          > "release/${{ env.ARCHIVE_NAME }}.tar.gz.sha256sum"
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install build
+        run: pip install build
+
+      - name: Build Source Distribution
+        run: python -m build --sdist --outdir release
+
+      - name: Compute Checksums
+        run: |
+          cd release
+          sha256sum * > SHA256SUMS
 
       - name: Upload Archive
         env:


.github/workflows/upload.yml
```diff
@@ -37,13 +37,24 @@ jobs:
...
+      - name: Install build
+        run: pip install build
+
+      - name: Build Source Distribution
+        run: python -m build --sdist --outdir release
+
+      - name: Compute Checksums
+        run: |
+          cd release
+          sha256sum * > SHA256SUMS
```
jezdez, line 57: Is SHA256SUMS a common name for that?

.github/workflows/upload.yml
```diff
@@ -37,13 +37,24 @@ jobs:
       - name: Archive Source
         run: git archive
           --prefix="${{ env.ARCHIVE_NAME }}/"
-          --output="release/${{ env.ARCHIVE_NAME }}.tar.gz"
+          --output="release/${{ env.ARCHIVE_NAME }}_full.tar.gz"
```
jezdez, line 40:
```suggestion
          --output="release/${{ env.ARCHIVE_NAME }}_archive.tar.gz"
```

jezdez (commented):
Given that we haven't solved https://github.com/conda/conda/issues/12126 I'm inclined to not do this, it could give users the idea that installing conda via an sdist and the Python Packaging stack is supported. It's not supported at the moment.

.github/workflows/upload.yml
dbast: Didn't find a standard yet, but it is e.g. used by Debian since ages, see e.g. https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/SHA256SUMS

dbast:
> Given that we haven't solved #12126 I'm inclined to not do this, it could give users the idea that installing conda via an sdist and the Python Packaging stack is supported. It's not supported at the moment.

Well, I see it more like a consistency check (not a statement that conda is installed via sdist):
* `python -m build --sdist .` runs hatch to create the sdist tarball 
* conda-build runs `{{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv && {{ PYTHON }} -m conda init --install` and thus also runs hatch to do the pip install
* this is a consistency check as hatch will only install things when called via pip as it puts into the sdist via build call... the intermediate sdist allows to inspect what hatch considers as relevant to be installed / packed and see if that is correct... (had some old CI pipelines years ago, where we first created the sdist, then installed from it to ensure the sdist is complete)

dbast:
There are two other arguments that would speak for the sdist as distribution archive:
* the sdist has the correct hatch derived version already burned-in, thus we don't rely on a correct https://github.com/conda/conda/blob/main/.git_archival.txt file + artifact name.
* providing only the sdist to downstream feedstocks is a kind of governance pattern as it enforces consistency, by preventing that arbitrary files straight from the repo can be copied into the conda package, as everything has to then be from the sdist only.

.github/workflows/upload.yml
jezdez: Ah, that's where I saw it, thanks for the pointer. Yeah, it's as good as any, and some precedence is useful.

zklaus:
Just stumbled upon this PR in a discussion with Tania. The context is that `conda` is a dependency for `conda-store`, but commented out in the `pyproject.toml` because it's not on Pypi. My personal opinion from past experience in the climate science world is that Python distribution packages should be published to Pypi even when they are not pip installable, in that case usually only as sdists, of course.

Among other things that enables a meaningful `pip check` for downstream packages and allows them to declare complete dependencies in their respective `pyproject.toml` files.

jezdez:
> Just stumbled upon this PR in a discussion with Tania. The context is that `conda` is a dependency for `conda-store`, but commented out in the `pyproject.toml` because it's not on Pypi. My personal opinion from past experience in the climate science world is that Python distribution packages should be published to Pypi even when they are not pip installable, in that case usually only as sdists, of course.

While I appreciate your experience with Python packaging in your field of work, I'm afraid you're wrong about conda being a Python packaging tool. It's a common misunderstanding and a serious barrier of entry for Python users to adopt conda, sadly, with considerable impact on adoption. Imagine all the "regular Python users" that would benefit from conda!

Sadly pip will create wheel files out of sdists automatically and install them and there is no way to mark a package as uploaded to PyPI but unsupported to my knowledge. From my experience, once it's uploaded to PyPI, the floodgates will open and users will start to install it from there.

In other words, publishing sdists on PyPI is not appropriate before we've solved #12126 and made sure everything works as expected (especially handling the shell integration and the dependencies). This should not be done accidentally or just because it makes a release process easier.

> Among other things that enables a meaningful `pip check` for downstream packages and allows them to declare complete dependencies in their respective `pyproject.toml` files.

You're assuming that conda is currently a fully Python packaging compatible project, but we aren't there yet (maybe [PEP 725](https://peps.python.org/pep-0725/) will help). pip check will completely ignore the current way of specifying conda dependencies (packages on Anaconda's main channel or conda-forge).

zklaus:
I understand there are many factors to consider and these decisions should not be made lightly. To clarify, I did not mean to imply that conda is "only" a Python packaging tool, merely that anything that installs a `.dist-info` directory is a [Python distribution package](https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/) and that it is useful to have the metadata and dependency relations available from a single source like PyPI.

I have helped with the packaging of many non-Python packages for conda-forge, sometimes completely independent from Python (such as subversion), but mostly with a Python binding somewhere along the chain. In those cases `pip check` will of course not guarantee that the binary dependencies are in order, but we get the Python side covered as well as possible.

Having said all that, I hear you: Solve #12126 and everybody is happy :smiley:

PR #13716, opened 2024-03-22
Use `FutureWarning` for CLI deprecations

### LINKED ISSUE #13704: `--force` no longer appears to be supported
### Checklist

### What happened?

I have several automated workflows that use `--force` when creating an environment. I use this flag to bypass all the `[Y/N]` prompts when running from a user-less environment.

This started happening after the latest release in my GitHub workflows.

Excerpt:
```
usage: conda [-h] [-v] [--no-plugins] [-V] COMMAND ...
conda: error: unrecognized arguments: --force
```

Here are some examples of failed runs with the latest version of `conda`:
- https://github.com/anaconda/conda-recipe-manager/actions/runs/8169639733/job/22797339349?pr=11
- https://github.com/anaconda/conda-recipe-manager/actions/runs/8169639733/job/22797338604?pr=11
- https://github.com/anaconda/conda-recipe-manager/actions/runs/8169639733/job/22797339000?pr=11

### Conda Info

```shell
Will get later, running from GitHub Workflow container.
```

### Conda Config

```shell
Will get later, running from GitHub Workflow container.
```

### Conda list

```shell
Will get later, running from GitHub Workflow container.
```

### Additional Context

_No response_

kenodegard:
### Description

We found that using `DeprecationWarning` for CLI deprecations does not offer the correct behavior (and modifying the default behavior is not ideal as it messes with downstream expectations).

As such we propose to raise `FutureWarning` instead for CLI deprecations. (thanks @mbargull for suggesting this)

With the proposed changed here `FutureWarning` is raised instead of `DeprecationWarning` only for `argparse.Action` deprecations (see `conda.deprecations.deprecated.action`). Otherwise the behavior has remained the same.

This PR also happens to complete the type annotations for `conda.deprecations` (and enables strict static type checking).

Resolves #13704 

### Checklist - did you ...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,17 @@ repos:
         args: [--fix]
       # compatible replacement for black
       - id: ruff-format
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.9.0
+    hooks:
+      - id: mypy
+        language_version: "3.12"
+        args: [--follow-imports=silent, --strict]
+        additional_dependencies: [packaging >=23.0]
+        files: |
+          (?x)^(
+            conda/deprecations.py
+          )$
   - repo: meta
     # see https://pre-commit.com/#meta-hooks
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,17 +68,6 @@ repos:
         args: [--fix]
       # compatible replacement for black
       - id: ruff-format
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.9.0
-    hooks:
-      - id: mypy
-        language_version: "3.12"
-        args: [--follow-imports=silent, --strict]
-        additional_dependencies: [packaging >=23.0]
-        files: |
-          (?x)^(
-            conda/deprecations.py
-          )$
   - repo: meta
     # see https://pre-commit.com/#meta-hooks
     hooks:
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -6,15 +6,26 @@ from __future__ import annotations
 
 import sys
 import warnings
+from argparse import Action
 from functools import wraps
 from types import ModuleType
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from argparse import Action
-    from typing import Any, Callable
+    from argparse import ArgumentParser, Namespace
+    from typing import Any, Callable, ParamSpec, Self, TypeVar
 
     from packaging.version import Version
+
+    T = TypeVar("T")
+    P = ParamSpec("P")
+
+    ActionType = TypeVar("ActionType", bound=type[Action])
+    PendingType = TypeVar("PendingType", bound=type[Warning])
+    DeprecatedType = TypeVar("DeprecatedType", bound=type[Warning])
+
+    UserDeprecationType = type[PendingDeprecationWarning] | type[FutureWarning]
+    DevDeprecationType = type[PendingDeprecationWarning] | type[DeprecationWarning]
 
 from . import __version__
 
@@ -30,7 +41,7 @@ class DeprecationHandler:
     _version_tuple: tuple[int, ...] | None
     _version_object: Version | None
 
-    def __init__(self, version: str):
+    def __init__(self: Self, version: str) -> None:
         """Factory to create a deprecation handle for the specified version.
 
         :param version: The version to compare against when checking deprecation statuses.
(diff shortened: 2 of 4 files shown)


jezdez:
Good idea!

schuylermartin45 (approved):
I appreciate the extra effort to parameterize these tests! Looks good otherwise.

beeankha (dismissed):

.pre-commit-config.yaml
```diff
@@ -68,6 +68,17 @@ repos:
...
+    rev: v1.9.0
+    hooks:
+      - id: mypy
+        language_version: "3.12"
+        args: [--follow-imports=silent, --strict]
+        additional_dependencies: [packaging >=23.0]
+        files: |
+          (?x)^(
+            conda/deprecations.py
+          )$
```
jezdez, line 81: Hmm, I think that should be discussed in the team first before we're adding it, not sure if everyone is familiar running mypy and how to resolve issues, also kind of not related to the PR

conda/deprecations.py
```diff
@@ -350,22 +374,76 @@ def _get_module(self, stack: int) -> tuple[ModuleType, str]:
...
+        :param addendum: Additional messaging. Useful to indicate what to do instead.
+        """
+        return self._generate_message(
+            deprecate_in,
+            remove_in,
+            prefix,
+            addendum,
+            PendingDeprecationWarning=PendingDeprecationWarning,
+            DeprecationWarning=DeprecationWarning,
+        )
```
jezdez, line 423: These two seem a little repetitive, is this needed because of the different return type?

pyproject.toml
```diff
@@ -177,6 +177,10 @@ pythonpath = [
   "tests/plugins/data/test-plugin",
 ]
 
+[tool.mypy]
+mypy_path = "conda"
+modules = ["conda"]
```
jezdez, line 182: Do we need to also do this for `conda_env`?

tests/test_deprecations.py
```diff
@@ -1,268 +1,200 @@
...
+)
+
+parametrize_user = pytest.mark.parametrize(
+    "deprecated,warning,message",
+    [PENDING, FUTURE, REMOVE],
+)
+parametrize_dev = pytest.mark.parametrize(
+    "deprecated,warning,message",
+    [PENDING, DEPRECATED, REMOVE],
+)
```
jezdez, line 51: Neat

jezdez (commented):
I think turning on mypy by default should be a separate conversation from this change to the deprecations system

conda/deprecations.py
kenodegard: This is repetitive because in `_get_user_messages` we use `FutureWarning` instead of `DeprecationWarning`

.pre-commit-config.yaml
kenodegard: the idea was that once a file was fully type annotated we wouldn't allow ourselves to regress

but yes I agree, its unrelated enough and should be extracted and put it into a new PR

schuylermartin45: I agree with Ken that we should slowly enforce `mypy` over time.

Fair warning, I have had some inconsistencies with dependencies when running it through `pre-commit` over having `pre-commit` execute a `make` script that runs with the environment dependencies, but that's a story for another time.

kenodegard:
@jezdez I see what you're doing, I originally did what you're suggesting but thought that having `_get_user_message` versus `_generate_developer_message` would be more clear and would make it easier to add new deprecation patterns in the future

let me experiment some more with this

jezdez:
> @jezdez I see what you're doing, I originally did what you're suggesting but thought that having `_get_user_message` versus `_generate_developer_message` would be more clear and would make it easier to add new deprecation patterns in the future
> 
> let me experiment some more with this

Ah, got it, I'm not sure if we couldn't refactor it later when/if we add another audience, for now, it seems worthwhile to try `FutureWarning` first

kenodegard:
@jezdez ah I just remembered this evolved this way since `mypy` doesn't understand setting a default value for a generic type, e.g., the following fails:

```python
DeprecationType = TypeVar("DeprecationType", bound=type[Warning])

def _generate_message(
    self: Self,
    deprecate_in: str,
    remove_in: str,
    prefix: str,
    addendum: str | None,
    *,
    deprecation_type: DeprecationType = DeprecationWarning,  # <-- error
) -> tuple[type[PendingDeprecationWarning] | DeprecationType | None, str]:
```

anyhow, went ahead and removed `_generate_user_message` and `_generate_developer_message` in favor of simpler typing

pyproject.toml
kenodegard: `mypy` from this PR

beeankha (approved):

Issue #13685, opened 2024-03-12
Authenticode Sign cli-32.exe and cli-64.exe prebuilt binaries


Callek:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Windows users when they `conda install` a package with entrypoints end up with an executable that is not codesigned. We should provide a codesigned copy of our stubs to make the executables produced as part of that process also signed.

### Why is this needed?

Users may be on a more locked down system requiring executable files, like this, be signed.  Having codesigned copies of these stubs helps unlock that potential for our users.

### What should happen?

The conda release process, and possibly CI, ensures these binaries are signed such that they may be used by windows users for launching python entrypoints on windows packages.

### Additional Context

Anaconda Inc, is willing to sign the existing binaries ahead of a more inclusive solution designed in the open for the Conda Community.

This would apply both to `conda`'s https://github.com/conda/conda/tree/main/conda/shell and `conda-build`'s [https://github.com/conda/conda-build/blob/main/conda_build/cli-{32,64}.exe](https://github.com/conda/conda-build/blob/main/conda_build/cli-%7B32,64%7D.exe) as sourced at https://github.com/conda/conda-build/tree/main/conda_build/launcher_sources

jaimergp:
There are some launchers in conda/menuinst too which might need the same treatment.

Callek:
> There are some launchers in conda/menuinst too which might need the same treatment.

Those are not pre-built though correct?  And they can be signed as part of the build process itself?

jaimergp:
Yes, they are, check menuinst/data.

Callek:
> Yes, they are, check menuinst/data.

Ahhh - https://github.com/conda/menuinst/tree/main/menuinst/data - those are osx-only binaries (from what I can tell) and the recipe for menuinst seems to be compiling the windows ones. (Authenticode signing is windows only)

References:
* https://github.com/conda/menuinst/blob/db528854422d765d9423eb7c705e06b0c7ade905/setup.py#L11
* https://github.com/conda-forge/menuinst-feedstock/blob/100bc037d6759541cb8e55194037ae1239f0c90b/recipe/meta.yaml#L35
* https://github.com/AnacondaRecipes/menuinst-feedstock/blob/692c881bd53914f1283b68b310ff62cf17b6d814/recipe/meta.yaml#L36

jezdez:
Resolved by #13721.

jezdez:
[...]

PR #13721, opened 2024-03-22
Provide codesigned stub exe's

Callek:
### Description

- This provides codesigned exe's for conda's stubs
- Relates to #13685
- Signatures provided on behalf of Anaconda

### Checklist - did you ...


tests/test_codesigned.py
```diff
@@ -0,0 +1,98 @@
...
+    return True if signtool_unsupported_because() else False
+
+
+@pytest.mark.skipif(signtool_unsupported(), reason=signtool_unsupported_because())
+@pytest.mark.parametrize("stub_file_name", ["cli-32.exe", "cli-64.exe"])
+def test_stub_exe_signatures(stub_file_name: str) -> None:
+    """Verify that signtool verifies the signature of the stub exes"""
+    logger = logging.getLogger(__name__)
+    stub_file = STUB_FOLDER / stub_file_name
+    logger.info(f"Checking for signature in {stub_file}")
```
travishathaway, line 91: Is there a reason for the logger call here? Isn't it already clear from the output of pytest which value this variable has?

tests/test_codesigned.py
```diff
@@ -0,0 +1,98 @@
...
+    signtool_exe = find_signtool()
+    try:
+        # raises CalledProcessError if this fails
+        check_call([signtool_exe, "verify", "/pa", "/v", stub_file])
+    except CalledProcessError as e:
+        logger.error(f"Command failed with exit code {e.returncode}")
+        raise
```
travishathaway, line 98:
```suggestion
    completed_process = subprocess.run([signtool_exe, "verify", "/pa", "/v", stub_file])
    assert completed_process.return_code == 0
```

I think this would be a little more concise. We're also making use of assertions here rather than just raising an error. This should be a little easier to debug in pytest results in the future in case this test begins failing.

tests/test_codesigned.py
```diff
@@ -0,0 +1,98 @@
...
+from conda.common.compat import on_win
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+REPO_ROOT = (Path(HERE) / "..").resolve().absolute()
+STUB_FOLDER = REPO_ROOT / "conda" / "shell"
+
+
+@lru_cache(maxsize=None)
+def find_signtool() -> str | None:
+    """Tries to find signtool
```
travishathaway, line 23: Is there any documentation to this signtool program you could link to here? I'm thinking this might be nice for future developers who need to debug this when it starts failing.

news/13721-sign-stubs
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* (windows) the stub-exe's that are used for python entry_point's in packages are now codesigned. (Thanks to Anaconda, Inc)
```
beeankha, line 3:
```suggestion
* For Windows users, the stub-exe's that are used for Python entrypoints in packages are now codesigned. (#13721)
```

tests/test_codesigned.py
```diff
@@ -0,0 +1,95 @@
...
+                # Use most recent signtool version
+                if not dirname.endswith(".0"):
+                    continue  # next dirname
+                if int(dirname.replace(".", "")) < max_version:
+                    continue  # next dirname
+
+                maybe_signtool_path = os.path.join(path, dirname, "x64", "signtool.exe")
+                if os.path.exists(maybe_signtool_path):
+                    signtool_path = maybe_signtool_path
+    return signtool_path
```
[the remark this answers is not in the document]
Callek: I took some slight inspiration from https://github.com/Azure/azure-code-signing-action.

`which` isn't enough because on these images we do not have this in the PATH by default, due to the multiple SDKs that github installs.  We could in theory use the `path` parameter but the version of windows SDK's does vary from system to system and I wanted to prefer what is already set in the path and not override it.

The version info comparison is explicitly so we always take the most recent among whatever sdk is installed

I've also seen evidence, via googling that `"C:\\Program Files\\Windows Kits\\10\\bin"` and `"C:\\Windows\\System32"` may contain signtool in some circulstances, so I wrote this additionally so that it is more likely to succeed on local developer systems.

tests/test_codesigned.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def signtool_unsupported() -> bool:
+    return bool(signtool_unsupported_because())
+
+
+@pytest.mark.skipif(signtool_unsupported(), reason=signtool_unsupported_because())
+@pytest.mark.parametrize("stub_file_name", ["cli-32.exe", "cli-64.exe"])
+def test_stub_exe_signatures(stub_file_name: str) -> None:
+    """Verify that signtool verifies the signature of the stub exes"""
```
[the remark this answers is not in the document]
Callek: This is to verify that the pre-built stubs are signed with a valid (chains to windows root) signature.  This wouldn't be the case if someone went to refresh the exe's and didn't sign them.

The test is so that we don't get surprised down the line by someone updating the binaries and not having had them signed prior to the PR.

conda/shell/cli-32.exe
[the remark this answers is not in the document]
Callek: From my prior research these were created from https://github.com/conda/conda-build/tree/main/conda_build/launcher_sources - For my use-case here I did not rebuild what we've already been shipping for ages and merely took the existing binaries and signed them.

We likely will want to rebuild these in the future, and have stricter provenance guarantees around those binaries -- not to mention the upstream changes those sources have incurred since the last binary updates https://github.com/python/cpython/compare/main/PC/launcher.c -- and additionally since python3.7 there is now a `launcher2` in upstream sources: https://github.com/python/cpython/blob/main/PC/launcher2.c

jezdez (dismissed):

jezdez:
Retargeted at the release branch since we plan to do a 24.3.1 release with these changes.

jezdez (approved):

jezdez:
@Callek Hmm, the rebasing didn't properly reduce the changes to what you had planned, could you rebase on top of 24.3.x please?

Callek:
> @Callek Hmm, the rebasing didn't properly reduce the changes to what you had planned, could you rebase on top of 24.3.x please?

Sure, I do that within the hour

Callek:
> > @Callek Hmm, the rebasing didn't properly reduce the changes to what you had planned, could you rebase on top of 24.3.x please?
> 
> Sure, I'll do that within the hour

...and Done

chenghlee (approved):

jezdez:
The performance regressions are odd, @Callek any idea what that could be?

Callek:
> The performance regressions are odd, @Callek any idea what that could be?

Really odd if this affected performance...  it didn't error on my initial pr. Maybe it's comparing against the wrong base?

Callek:
> > The performance regressions are odd, @Callek any idea what that could be?
> 
> Really odd if this affected performance... it didn't error on my initial pr. Maybe it's comparing against the wrong base?

EDIT: I didn't think it had a performance issue in initial PR...   But now that I'm looking it may have

Taking one of the simpler tests `test_env_update` it seems this creates a new (temporary) environment...

https://github.com/Callek/conda/blob/0434968525bb2c956d7e9f6d5dcacf1af2008099/tests/cli/test_subcommands.py#L239-L240

This would be using the new binaries for stubs...  Secondly the CodSpeed is showing a diff in call stacks between the base and head commits, which is puzzling a bit:

image
image


In All the regressed tests it seems to be adding a new `PyObject_Vectorcall` invocation...

@jezdez how should I proceed?

Callek:
Sorry for additional noise, but looks like we're only running benchmarks on py3.12 on *linux* so I'd have ZERO idea why this could be impactful here... and even posit that this change is mere noise-in-the-machine

tests/test_codesigned.py
```diff
@@ -0,0 +1,95 @@
...
+    except CalledProcessError as exc:
+        reason = f"signtool: something went wrong while running 'signtool verify /?', output:\n{exc.output}\n"
+    return reason
+
+
+def signtool_unsupported() -> bool:
+    return bool(signtool_unsupported_because())
+
+
+@pytest.mark.skipif(signtool_unsupported(), reason=signtool_unsupported_because())
```
jezdez, line 88: Let's see if inlining this skipping into the test may prevent the tests to degrade performance wise

jezdez:
As proven by the test PR https://github.com/conda/conda/pull/13787 the performance degradation seems to be something entirely different, or just a branch targeting issue in Codspeed. Let's merge this as-is.

PR #13724, opened 2024-03-23
Add search algorithm for environments.txt file (remove hardcoded path)

jrom99:
### Description

As seem on https://github.com/conda/conda/issues/13453, the `environments.txt` file is a configuration file that is currently hardcoded to `~/.conda/environments.txt` and is one (among many) files responsible for the creation of `~/.conda`. This commit tries to reduce the dependency on such directory.

The current behavior of `get_user_environments_txt_file` is ambiguous in that it may return a file to write (default behavior) or multiple files to read (during the execution of `list_all_known_prefixes`). This commit adds a search algorithm for it so that it is not written to `~/.conda` by default (but it's still able to find it there), and adds a docstring specifying the "file to write" behavior.

This commit expands `list_all_known_prefixes` search algorithm to include all config paths specified in <https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#searching-for-condarc>.

One should be aware that another user on the same computer may choose to set their `$XDG_CONFIG_HOME` to a value that is different from `$HOME/.config`. Since this was not dealt with before (and I'm not sure what should be the expected behavior when one lists environments with admin privileges), I choose not to implement this.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes? No.
     
- [x] Add / update necessary tests? No.
- [x] Add / update outdated documentation? No.

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -164,6 +164,9 @@ CONDA_LOGS_DIR = ".logs"
 UNKNOWN_CHANNEL = "<unknown>"
 REPODATA_FN = "repodata.json"
 
+#: Name of file where environment paths are listed
+ENVIRONMENTS_FN = "environments.txt"
+
 #: Default name of the notices file on the server we look for
 NOTICES_FN = "notices.json"
 
--- a/conda/core/envs_manager.py
+++ b/conda/core/envs_manager.py
@@ -7,6 +7,9 @@ from errno import EACCES, ENOENT, EROFS
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, normpath
 
+from platformdirs import user_config_dir
+
+from ..base.constants import APP_NAME, ENVIRONMENTS_FN
 from ..base.context import context
 from ..common._os import is_admin
 from ..common.compat import ensure_text_type, on_win, open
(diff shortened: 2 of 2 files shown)


bigodel:
adding a +1 to this feature

jayqi:
I'm also in favor of conda supporting XDG and annoyed that I can't get rid of the `~/.conda` directory because of `~/.conda/environments.txt`. 

However, I don't think `XDG_CONFIG_HOME` and `user_config_dir` is appropriate for `environments.txt`. This is a data file and not a config file. People often sync config files to other machines, but it doesn't make sense to sync a list of the environments on one machine to another. Instead, this should use `XDG_DATA_HOME` and `platformdirs.user_data_dir`.

jrom99:
> I'm also in favor of conda supporting XDG and annoyed that I can't get rid of the `~/.conda` directory because of `~/.conda/environments.txt`. 
> 
> However, I don't think `XDG_CONFIG_HOME` and `user_config_dir` is appropriate for `environments.txt`. This is a data file and not a config file. People often sync config files to other machines, but it doesn't make sense to sync a list of the environments on one machine to another. Instead, this should use `XDG_DATA_HOME`, `~/.local/share`, and `platformdirs.user_data_dir`. 

Oh, that is true! I'll update the code to the correct position. I based this change on the fact that environments.txt is located next to condarc, which is a config file.

I think there are also other XDG_DATA dirs available to search, so I'll try to link the correct documentation.

jrom99:
Sorry for the long wait, the other path I mentioned is `site_data_dir`, but since it is not searched for writing **or** for reading, I don't think we should worry about it.

The problem of a user selecting a different XDG_DATA_DIR than default still applies though, and I'd need a maintainer to know what should be the expected behavior here, since implementing a way to discover each user's XDG_DATA_DIR would be really prone to bugs.

jrom99:
Still waiting for a reviewer.

PR #13732, opened 2024-03-25
Stop using `log.trace(...)`

jaimergp:
### Description

This is a long-standing issue that prevents `conda` from being used as a library. We can't test `SubdirData` in isolation unless we first call `conda.gateways.logging.initialize_logging` which, among other things, patches `logger.Logger` to have a new method `.trace()` which essentially does `Logger.log(5, ...)`.

I recommend we deprecate it and stop using it. This is the first step to allow easier testability / library usage in other projects.

### Checklist - did you ...

### DIFF
--- a/conda/common/_logic.py
+++ b/conda/common/_logic.py
@@ -5,6 +5,8 @@ from array import array
 from itertools import combinations
 from logging import DEBUG, getLogger
 
+from conda.gateways.logging import trace
+
 log = getLogger(__name__)
 
 
@@ -588,7 +590,7 @@ class Clauses:
         nterms = len(coeffs)
         if nterms and coeffs[-1] > hi:
             nprune = sum(c > hi for c in coeffs)
-            log.trace("Eliminating %d/%d terms for bound violation" % (nprune, nterms))
+            trace(log, "Eliminating %d/%d terms for bound violation" % (nprune, nterms))
             nterms -= nprune
         else:
             nprune = 0
@@ -686,10 +688,10 @@ class Clauses:
         try0 = 0
         for peak in (True, False) if maxval > 1 else (False,):
             if peak:
-                log.trace("Beginning peak minimization")
+                trace(log, "Beginning peak minimization")
                 objval = peak_val
             else:
-                log.trace("Beginning sum minimization")
+                trace(log, "Beginning sum minimization")
                 objval = sum_val
 
             objective_dict = {a: c for c, a in zip(coeffs, lits)}
@@ -706,7 +708,7 @@ class Clauses:
             if trymax and not peak:
                 try0 = hi - 1
 
-            log.trace("Initial range (%d,%d)" % (lo, hi))
+            trace(log, "Initial range (%d,%d)" % (lo, hi))
             while True:
                 if try0 is None:
                     mid = (lo + hi) // 2
@@ -722,14 +724,14 @@ class Clauses:
                     self.Require(self.LinearBound, lits, coeffs, lo, mid, False)
 
                 if log.isEnabledFor(DEBUG):
-                    log.trace(
+                    trace(log, 
                         "Bisection attempt: (%d,%d), (%d+%d) clauses"
                         % (lo, mid, nz, self.get_clause_count() - nz)
                     )
                 newsol = self.sat()
                 if newsol is None:
                     lo = mid + 1
-                    log.trace("Bisection failure, new range=(%d,%d)" % (lo, hi))
+                    trace(log, "Bisection failure, new range=(%d,%d)" % (lo, hi))
                     if lo > hi:
                         # FIXME: This is not supposed to happen!
                         # TODO: Investigate and fix the cause.
@@ -742,7 +744,7 @@ class Clauses:
                     bestsol = newsol
                     bestval = objval(newsol, objective_dict)
                     hi = bestval
-                    log.trace("Bisection success, new range=(%d,%d)" % (lo, hi))
+                    trace(log, "Bisection success, new range=(%d,%d)" % (lo, hi))
                     if done:
                         break
                 self.m = m_orig
--- a/conda/common/_logic.py
+++ b/conda/common/_logic.py
@@ -590,7 +590,9 @@ class Clauses:
         nterms = len(coeffs)
         if nterms and coeffs[-1] > hi:
             nprune = sum(c > hi for c in coeffs)
-            log.log(TRACE, "Eliminating %d/%d terms for bound violation" % (nprune, nterms))
+            log.log(
+                TRACE, "Eliminating %d/%d terms for bound violation" % (nprune, nterms)
+            )
             nterms -= nprune
         else:
             nprune = 0
--- a/conda/common/_logic.py
+++ b/conda/common/_logic.py
@@ -724,9 +724,10 @@ class Clauses:
                     self.Require(self.LinearBound, lits, coeffs, lo, mid, False)
 
                 if log.isEnabledFor(DEBUG):
-                    trace(log, 
+                    trace(
+                        log,
                         "Bisection attempt: (%d,%d), (%d+%d) clauses"
-                        % (lo, mid, nz, self.get_clause_count() - nz)
+                        % (lo, mid, nz, self.get_clause_count() - nz),
                     )
                 newsol = self.sat()
                 if newsol is None:
--- a/conda/common/_logic.py
+++ b/conda/common/_logic.py
@@ -5,7 +5,7 @@ from array import array
 from itertools import combinations
 from logging import DEBUG, getLogger
 
-from conda.gateways.logging import TRACE
+from .constants import TRACE
 
 log = getLogger(__name__)
 
--- a/conda/common/_logic.py
+++ b/conda/common/_logic.py
@@ -729,7 +729,10 @@ class Clauses:
                     log.log(
                         TRACE,
                         "Bisection attempt: (%d,%d), (%d+%d) clauses",
-                        lo, mid, nz, self.get_clause_count() - nz,
+                        lo,
+                        mid,
+                        nz,
+                        self.get_clause_count() - nz,
                     )
                 newsol = self.sat()
                 if newsol is None:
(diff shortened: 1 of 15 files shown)


jaimergp:
There are several questions to resolve while doing this:

- Deprecate `log.trace` or just remove it? I really hope no one is using it.
- Really consider if we actually need a `logLevel = 5` or if those ~70 `Logger.trace()` calls can be promoted to simple `log.debug()` ones.
- If we do want that, implement them as `Logger.log(5, ...)` instead of `trace()` and only check for `isEnabledFor` when necessary.
- Add a custom linter to warn against usage of `log.trace()` in the future, in case we have some developer inertia.

cc @zklaus who's looking into this.

mbargull:
@jaimergp, @zklaus, thanks for looking into these things!

> - Deprecate `log.trace` or just remove it? I really hope no one is using it.

We should mark this for deprecation (and I'm very much +1 for removing it after that!).

> - If we do want that, implement them as `Logger.log(5, ...)` instead of `trace()` and only check for `isEnabledFor` when necessary.

Yes, replacing `logger.trace(...)` with `logger.log(TRACE, ...)` is what we should do.
`log(TRACE, ...)` is not too inconvenient and patching `logging.Logger` is not warranted for a negligible convenience thing.
As for `isEnabledFor`, IIRC, `logging.Logger.log` does it anyway (please confirm, though!), so no need to worry about that.

> - Alternatively, if we want custom logging, do so from the get-go and stop monkey-patching the root Logger class. Just create a subclass and instantiate the loggers from the helper module (e.g. `from logging import getLogger` would become `from conda.gateways.logging import getLogger`, assuming no circular imports).
> - (Optional) Add a custom linter to warn against usage of `log.trace()` in the future, in case we have some developer inertia, or `from logging import getLogger` in case we go full-custom.

If we go with a normal deprecation cycle and change internal uses to `.log(TRACE, ...)`, we should already be set, IMO:
- Using `log(<level>, ...)` is convenient enough and already supported by the stdlib, i.e., no customization needed.
- A `@conda.deprecation.deprecated(...)` decorator will provide usage warning.

----
^-- I intentionally left out comments about the log level itself and how/if we want to use in the future ;).
IMO, a discussion about that should be held separately to not cause unnecessary delay for a deprecation cycle for `trace()`.

jaimergp:
> As for `isEnabledFor`, IIRC, `logging.Logger.log` does it anyway (please confirm, though!), so no need to worry about that.

[Correct](https://github.com/python/cpython/blob/54514467c26c657eb1a6f8885786297979689d2d/Lib/logging/__init__.py#L1608-L1609)! This is great news! I can just add `.log(TRACE,...)` instead of the awkward `trace(log,...)`.

jezdez:
@kenodegard You looked at the logging code last, could you please take a look at the proposal here?

@jaimergp I'm inclined to use a subclassed logger instead of dropping trace (given the obvious usefulness, right?) to get rid of the code smell during initialization, even if it may introduce new bootstrap problems. It could be a prep step to start using more advanced logging tech like [structlog](https://www.structlog.org/) later though. Or we make that jump right away? Consider me a supporter of a more radical approach to solving this problem.

I'm thinking `logging.getLogger(__name__)(conda.common.constants.TRACE, ...)` isn't a great API and will confuse the casual contributors, the logging library hasn't gotten less awkward sadly over the years.

jaimergp:
> I'm inclined to use a subclassed logger instead of dropping trace 

I looked into using [`logging.setLoggerClass`](https://docs.python.org/3/library/logging.html#logging.setLoggerClass), but I saw two impediments:

* We will still need to call that function somewhere (probably in `conda.__init__`) to initialize things.
* It will have side effects to _all_ projects using `logging` once `conda` has been imported. Not that they are big given that we are "just" adding one method.

That's the cost we need to pay for the convenience.

> given the obvious usefulness, right?

Up to personal taste, I guess. While I don't find it super useful, [I searched on Github for similar uses](https://github.com/search?type=code&q=%2F%28log%7Clogger%7Clogging%29.trace%5C%28%2F+language%3APython&l=Python) and it seems to be somewhat popular.

* [`dvc`](https://github.com/iterative/dvc/blob/dd2b515c97a6757c6c3522329d6259e0e86592df/dvc/log.py) uses `trace` as an alias for `debug`
* [`kivy`](https://github.com/kivy/kivy/blob/5d17b27dbf245283617edc549a195ef816d82065/kivy/logger.py) has somewhat complex machinery but they do GUI development so maybe different needs
* [`faceswap`](https://github.com/deepfakes/faceswap/blob/64a7b5812e861de39746cc72fcdad6d0c3473748/lib/logger.py) does the subclass method. Note their initialization in [`__init__.py`](https://github.com/deepfakes/faceswap/blob/64a7b5812e861de39746cc72fcdad6d0c3473748/lib/__init__.py).

There are probably more examples. My original question was whether we actually need a trace level or not. I think we do (unfortunately, imo) for historical reasons and the fact that there are four verbosity levels supported in the CLI:

* `--quiet`: Warnings and errors. Basic reports.
* No flags: Same, plus extended progress reports.
* `-v`: Same plus Info.
* `-vv`: Same plus Debug.
* `-vvv`: Same plus Trace.

---

I think one way forward would be:

* We stop import `logging` directly.
* Instead we require `from conda.gateways import logging`.
* `conda.gateways.logging` has `getLogger` too, but it always return our subclass.
* A custom linter warns about the direct `logging` import and suggests using `conda.gateways.logging` instead.

If we must absolutely import from `logging` directly:

* We need to use `setLoggerClass`. This will affect all loggers and libraries. We shouldn't do this as a library. As an application it's ok, though. That separation is not yet available, sadly.
* We need to initialize things as soon as possible (in `conda.__init__`).
* This might have an added import cost for things like `conda activate`.
* To reduce the import cost, we would have to move `conda.gateways.logging` to `conda.logging` preferrably. 


I know I wrote a lot so apologies in advance. My preferred choice would the `from conda.gateways import logging` drop-in replacement + a little linter for future contributions.

dholth:
+1 on removing log.trace. It is completely useless.

conda/common/_logic.py
```diff
@@ -742,7 +747,7 @@ def sum_val(sol, objective_dict):
                     bestsol = newsol
                     bestval = objval(newsol, objective_dict)
                     hi = bestval
-                    log.trace("Bisection success, new range=(%d,%d)" % (lo, hi))
+                    log.log(TRACE, "Bisection success, new range=(%d,%d)" % (lo, hi))
```
dholth, line 750: Should be , lo, hi (defer string formatting when trace is disabled)

jaimergp:
```suggestion
                    log.log(TRACE, "Bisection success, new range=(%d,%d)", lo, hi)
```
Like this?

jaimergp:
FWIW, I also prefer `log.log(TRACE, ...)` if we agree that `log.trace(...)` is not worth the trouble.

conda/common/_logic.py
```diff
@@ -588,7 +590,9 @@ def LinearBound(self, lits, coeffs, lo, hi, preprocess, polarity):
         nterms = len(coeffs)
         if nterms and coeffs[-1] > hi:
             nprune = sum(c > hi for c in coeffs)
-            log.trace("Eliminating %d/%d terms for bound violation" % (nprune, nterms))
+            log.log(
+                TRACE, "Eliminating %d/%d terms for bound violation" % (nprune, nterms)
```
jaimergp, line 594:
```suggestion
                TRACE, "Eliminating %d/%d terms for bound violation", nprune, nterms
```

conda/common/_logic.py
```diff
@@ -706,7 +710,7 @@ def sum_val(sol, objective_dict):
             if trymax and not peak:
                 try0 = hi - 1
 
-            log.trace("Initial range (%d,%d)" % (lo, hi))
+            log.log(TRACE, "Initial range (%d,%d)" % (lo, hi))
```
jaimergp, line 713:
```suggestion
            log.log(TRACE, "Initial range (%d,%d)", lo, hi)
```

conda/common/_logic.py
```diff
@@ -722,14 +726,15 @@ def sum_val(sol, objective_dict):
                     self.Require(self.LinearBound, lits, coeffs, lo, mid, False)
 
                 if log.isEnabledFor(DEBUG):
-                    log.trace(
+                    log.log(
+                        TRACE,
                         "Bisection attempt: (%d,%d), (%d+%d) clauses"
-                        % (lo, mid, nz, self.get_clause_count() - nz)
+                        % (lo, mid, nz, self.get_clause_count() - nz),
```
jaimergp, line 732:
```suggestion
                        lo, mid, nz, self.get_clause_count() - nz,
```

conda/common/_logic.py
```diff
@@ -722,14 +726,15 @@ def sum_val(sol, objective_dict):
                     self.Require(self.LinearBound, lits, coeffs, lo, mid, False)
 
                 if log.isEnabledFor(DEBUG):
-                    log.trace(
+                    log.log(
+                        TRACE,
                         "Bisection attempt: (%d,%d), (%d+%d) clauses"
```
jaimergp, line 731:
```suggestion
                        "Bisection attempt: (%d,%d), (%d+%d) clauses",
```

conda/common/_logic.py
```diff
@@ -722,14 +726,15 @@ def sum_val(sol, objective_dict):
...
-                        % (lo, mid, nz, self.get_clause_count() - nz)
+                        % (lo, mid, nz, self.get_clause_count() - nz),
                     )
                 newsol = self.sat()
                 if newsol is None:
                     lo = mid + 1
-                    log.trace("Bisection failure, new range=(%d,%d)" % (lo, hi))
+                    log.log(TRACE, "Bisection failure, new range=(%d,%d)" % (lo, hi))
```
jaimergp, line 737:
```suggestion
                    log.log(TRACE, "Bisection failure, new range=(%d,%d)", lo, hi)
```

jaimergp:
pre-commit.ci autofix

travishathaway (approved):

jezdez (approved):

Issue #13709, opened 2024-03-19
Add `reporters` as a new configuration option


travishathaway:
### What's the idea?

Add a new configuration option, `reporters`, to conda.

### Why is this needed?

We want to be able to support multiple backends for writing to various output streams (e.g. files and stdout) and adding this new configuration option will make this process easier. Our primary use case for this is to add additional backends such as [rich](https://rich.readthedocs.io) in the future.

### What should be done?
- [ ] Add new `reporters` configuration option. This will be a list of mappings (see examples below). This will only be configurable via configuration files.
- [ ] Currently available options for verbosity (e.g. `--verbose`)  and output type (e.g. `--json`) will implicitly set values within `reporters` to make the CLI interface more user friendly (see examples below).
- [ ] ~Make sure this is documented~ _This is not being documented right now because setting this option has no effect, and we do not want to confuse users with such options at this point._

#### Examples

This is the default setting. It prints the regular command output to standard out:

```yaml
reporters:
  - backend: stdlib
    output: stdout
```

Setting JSON as output for standard out:

```yaml
reporters:
  - backend: json
    output: stdout
```

Writing JSON to a file and printing the normal output to stdout:

```yaml
reporters:
  - backend: json
    output: file
    filename: file.json
  - backend: stdlib
    output: stdout
```

jaimergp:
This assumes users can only have one output set a time. That's an ok assumption, but I want to discuss whether simultaneous outputs would make sense (e.g. regular output to stdout, json to file). In that case, should this allow for a more complex type like a list of dicts too? That can be done in further PRs, but again, just making sure we are not making an assumption implicitly.

jezdez:
> This assumes users can only have one output set a time. That's an ok assumption, but I want to discuss whether simultaneous outputs would make sense (e.g. regular output to stdout, json to file). In that case, should this allow for a more complex type like a list of dicts too? That can be done in further PRs, but again, just making sure we are not making an assumption implicitly.

I concur, I wouldn't mind if we'd make --output basically a list and design the interface so that it can behave as a stream of outputs, with optional filters

chenghlee:
I like the idea, but I'm not sure that calling this option simply `output` is the best choice, especially as an "top-level" key-value pair in `condarc`. IMO, just seeing  `output: json` or `output: ["json", "text"]` doesn't make it at least somewhat obvious _which_ output is being referenced.

Also, the `--output` argument already exists for `conda-build`:
```
[...]

Issue #13707, opened 2024-03-19
Refactor conda to allow for configurable display backends


travishathaway:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

Conda currently has the `json` option which will trigger it to display only JSON to standard out. Otherwise it uses mostly print statements and loggers to print messages to the user in the normal flow of running commands. For this epic, we want to enable the addition of new display managers such as rich by refactoring conda.

This refactor will proceed in various steps and the issues linked here reflect those steps. By the end of the refactor we will be able to set a new configuration option, `output`, that will trigger the usage of a particular display backend. The current `json` option will become an alias for `output=json`.

### Individual components in conda needing a refactor

These are the individual components within conda itself that will need to be refactored to allow for new layouts and displays via the `output` configuration variable:

- [ ] Refactor error handler to allow for configurable `ouput` backend
- [ ] Refactor "Spinner" and "ProgressBar" to allow for configurable `output` backend
- [ ] Refactor confirmation to allow for a configurable `output` backend
- [ ] Refactor "update", "package plan" and "environment activation" messages to allow for configurable `output` backend
- [ ] Refactor "verification" step to allow for configurable `output` backend

These components are primarily used within `create`, `update` and `install`. As we go through these, we may want to combine multiple tasks into a single pull request. But, this separation should be a good starting point that will keep pull request size manageable and easier to review.

### Commands in conda needing to be refactored

If the above refactor has been performed correctly, then the refactor for many of the commands listed below should require none or a very small amount of work. The commands requiring the most work here will be ones such as `config` or `info` which rely heavily on print statements contained within their own modules.

This list serves as both a check list to make sure everything has been covered and as a starting point for issues for commands requiring a more in depth refactor effort.

- [ ] Refactor `create`, `update` and `install` commands to allow for a configurable `output` backend
- [ ] #13692
- [ ] Refactor `config` command to allow for a configurable `output` backend
- [ ] Refactor `compare` command to allow for a configurable `output` backend
- [ ] Refactor `list` command to allow for a configurable `output` backend
- [ ] Refactor `notices` command to allow for a configurable `output` backend
- [ ] Refactor `remove` command to allow for a configurable `output` backend
- [ ] Refactor `search` command to allow for a configurable `output` backend

[...]

PR #13736, opened 2024-03-26
Add `reporters` setting to conda

travishathaway:
### Description

Fixes: #13709

Adds a new `reporters` setting to conda. This is currently not used and is undocumented. It will be used more as we progress with the following epic:

- https://github.com/conda/conda/issues/13707

### Checklist - did you ...


conda/base/context.py
```diff
@@ -1165,6 +1171,48 @@ def cpu_flags(self):
         info = _get_cpu_info()
         return info["flags"]
 
+    @memoizedproperty
+    @unique_sequence_map(
+        unique_key="backend",
+        property_name="_reporters",
+        allowed_keys={"json", "stdlib"},
+    )
+    def reporters(self) -> Sequence[Mapping[str, str]]:
```
schuylermartin45, line 1180: I'm curious why we are using the generic types here. Wouldn't it be more explicit/clear for our users if we specified a specific type like a `list[dict[str, str]]`?

conda/base/context.py
```diff
@@ -1165,6 +1171,48 @@ def cpu_flags(self):
...
+        if not self._reporters:
+            return (
+                {
+                    "backend": "stdlib",
+                    "output": "stdout",
+                    "verbosity": context.verbosity,
+                },
+            )
+
+        return self._reporters
```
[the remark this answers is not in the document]
kenodegard: Logging and warning occurs in the `unique_sequence_map`. See lines 1633-1638 in `conda.common.configuration` below: https://github.com/conda/conda/pull/13736/files#diff-e3e902a85b6d7a36bb1244ff9d9c7540a946c57fa967e76f9e176130e6c38b97R1633-R1638

conda/base/context.py
kenodegard: Agreed, a good rule of thumb is to use abstract inputs and concrete outputs

travishathaway: Will make it more specific. I think I will leaving mapping there though because it can theoretically return either a `tuple[dict[str, str]]` or `tuple[frozendict[str,str]]`. Sticking with the generic `Mapping` feels like a more elegant way of defining it.

conda/base/context.py
```diff
@@ -1165,6 +1171,48 @@ def cpu_flags(self):
...
+
+        if context.quiet:
+            return (
+                {
+                    "backend": "stdlib",
+                    "output": "stdout",
+                    "verbosity": self.verbosity,
+                    "quiet": True,
+                },
+            )
```
kenodegard, line 1202: I'm not convinced this is the correct thing to do for `context.quiet`. I'm wondering if quiet should be inserted into whatever the reporters value.

It's also harder to reason about this since the reporters aren't in use yet so maybe just leave a comment and revisit once the next PR is opened?

kenodegard (changes requested):
Let's add some more tests so we can reach 100% code coverage on the new `unique_sequence_map` function

conda/base/context.py
travishathaway: I think your correct on that. I will just insert `context.quiet` as a value to this dictionary.

travishathaway:
@kenodegard and @jezdez,

I have just finished another round of refactoring and test additions. This is ready for review again.

conda/base/context.py
travishathaway: I have vastly simplified this method in a recent commit. Please check it out and let me know if there are still any concerns.

conda/common/configuration.py
```diff
@@ -1600,3 +1601,56 @@ def typify_parameter(self, parameter_name, value, source):
 
     def get_descriptions(self):
         raise NotImplementedError()
+
+
+def unique_sequence_map(
+    *, unique_key: str, property_name: str, allowed_keys: set | None = None
```
kenodegard, line 1607: `property_name` is unused

```suggestion
    *, unique_key: str, allowed_keys: set | None = None
```

conda/common/configuration.py
```diff
@@ -1600,3 +1601,56 @@ def typify_parameter(self, parameter_name, value, source):
...
+    should be regarded as unique. This decorator will handle removing duplicates and
+    merging to a single sequence.
+    """
+
+    def inner_wrap(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            sequence_map = func(self, *args, **kwargs)
+            known_keys = set()
+            new_sequence_mapping = []
```
kenodegard, line 1621: We should use a `dict` instead with the `unique_key` as they key:

```suggestion
            new_sequence_mapping = {}
```

conda/common/configuration.py
```diff
@@ -1600,3 +1601,56 @@ def typify_parameter(self, parameter_name, value, source):
...
+    Used to validate properties on :class:`Configuration` subclasses defined as a
+    ``SequenceParameter(MapParameter())`` where the map contains a single key that
+    should be regarded as unique. This decorator will handle removing duplicates and
+    merging to a single sequence.
+    """
+
+    def inner_wrap(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            sequence_map = func(self, *args, **kwargs)
```
kenodegard, line 1619:
```suggestion
        def wrapper(*args, **kwargs):
            sequence_map = func(*args, **kwargs)
```

conda/common/configuration.py
kenodegard, line 1607: I don't think I agree with the `allowed_keys` as implemented here. This means our supported reporters are hardcoded in conda, instead I envision reporter plugins. Something more akin to what we do for the solver.

kenodegard (changes requested):

conda/common/configuration.py
travishathaway: @kenodegard,

I don't understand why this should be dictionary instead of a list. Could you please explain why?

conda/common/configuration.py
travishathaway: @kenodegard,

One way to solve this is by defining this function argument as a callback function which will allow it to be lazy-loaded at runtime.

Does this sound like a good idea to you?

travishathaway: But, I will also state here that I find it very unlikely that "reporters" will be turned into a plugin in the future, so I believe this could be over optimization for a problem that is unlikely to ever arise.

conda/common/configuration.py
kenodegard: The `known_keys` and `new_sequence_mapping` are tightly coupled with a one-to-one relationship:

```
        def wrapper(*args, **kwargs):
            new_sequence_mapping = {}

            for mapping in func(*args, **kwargs):
                unique_key_value = mapping.get(unique_key)

                if unique_key_value is None:
                    ...

                if unique_key_value in known_keys:
                    ...

                if allowed_keys is not None and unique_key_value not in allowed_keys:
                    ...

                new_sequence_mapping[unique_key_value] = mapping

            return tuple(new_sequence_mapping.values())
```

conda/common/configuration.py
kenodegard: Why can't these be plugin hooks from the get-go? Plugins aren't just for third-party use to add new functionality but also as a way for us to modularize and organize stuff within conda.

There are absolutely cases where additional reporters could be defined, e.g., conda-tui could define a new type of reporter that uses textual instead of rich or Navigator could desire a custom reporter to interface with conda instead of relying on the builtin json reporter.

I think about our plugin initiatives through the lens of "if we already had this plugin hook what could we have already accomplished sooner/faster?" And in this case, if we had a reporter hook previously defined, a third-party plugin could easily have introduced rich support years ago with minimal effort.

travishathaway: I hear those concerns, but I believe that discussion is better to be had in a separate issue. I would really like to keep this implementation as simple as possible for now and not have to worry about the complexities surrounding setting up another plugin hook for this.

In the future, this should be pretty easy to retro-fit to support that use case. Like I said, we could simply change this decorator function to optionally accept a callback function to determine all possible values for `allowed_keys`. That will allow us to dynamically determine the value of that argument at runtime and therefore take into account all the possible values registered by plugin hooks.

So, in order to keep this pull request going forward, I kindly ask that you open a separate issue and attach it to the ongoing epic for this work:

- https://github.com/conda/conda/issues/13707

I would be more than happy to have a conversation about plugin hooks there.

kenodegard (dismissed):

news/13736-add-reporters-setting
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Adds a new `reporters` setting for configure output
```
beeankha, line 3:
```suggestion
* Adds a new `reporters` setting for configure output. (#13736)
```

beeankha (dismissed):
Requested a change for the news file but otherwise LGTM!

beeankha (dismissed):

conda/base/context.py
```diff
@@ -1168,6 +1174,25 @@ def cpu_flags(self):
...
+        """
+        if not self._reporters:
+            return (
+                {
+                    "backend": "json" if self.json else "stdlib",
+                    "output": "stdout",
+                    "verbosity": self.verbosity,
+                    "quiet": self.quiet,
+                },
+            )
```
jezdez, line 1192: I assumed `json` would become an own reporter and not tied to stdout? E.g. `--json` is a shortcut for `--reporter=json`?

Also, why do we need the `reporters` setting? Wouldn't that list of reporters come from plugins that provide a reporter implementation?

Conda would ship with two out of the box:

```suggestion
        if not self._reporters:
            return (
                {
                    "backend": "plaintext",
                    "output": "stdout",
                    "verbosity": self.verbosity,
                    "quiet": self.quiet,
                },
                {
                    "backend": "json",
                    "output": "stdout",
                    "verbosity": self.verbosity,
                    "quiet": self.quiet,
                 },
             )
```

travishathaway: @jezdez,

The reporters listed there are the currently active ones, not the available ones.

The `reporters` setting is meant to be a convenience for other parts of the code and to make it easier to link it with what has been defined in the configuration file. 

It doesn't do anything right now, but as we progress with the refactor we will be relying on `context.reporters` as the source of truth for all the output rendering that needs to occur for the running process, which could be multiple things depending on what has been configured (e.g. we could choose both `output = file` and `output = stdout` to write to both stdout and a file at the same time).

Please let me know if this makes sense and whether you would like to discuss it more.

jezdez: > Please let me know if this makes sense and whether you would like to discuss it more

Ah, gotcha, I misunderstood the purpose, thanks for elaborating!

conda/base/context.py
```diff
@@ -1168,6 +1174,25 @@ def cpu_flags(self):
...
+    @unique_sequence_map(unique_key="backend")
+    def reporters(self) -> tuple[Mapping[str, str]]:
+        """
+        Determine the value of reporters based on other settings and the ``self._reporters``
+        value itself.
+        """
+        if not self._reporters:
+            return (
+                {
+                    "backend": "json" if self.json else "stdlib",
```
jezdez, line 1187:
```suggestion
                    "backend": "json" if self.json else "console",
```

tests/base/test_context.py
```diff
@@ -70,6 +71,11 @@
       rsync: 'false'
     aggressive_update_packages: []
     channel_priority: false
+    reporters:
+      - backend: json
+        output: test.json
+      - backend: stdlib
```
jezdez, line 77:
```suggestion
      - backend: console
```

tests/base/test_context.py
```diff
@@ -759,3 +765,68 @@ def value(self, _):
...
     assert plugin_config_data == {}
+
+
+def test_reporters_from_config_file(testdata):
+    """
+    Ensure that the ``reporters`` property returns the correct values
+    """
+    assert context.reporters == (
+        {"backend": "json", "output": "test.json"},
+        {"backend": "stdlib", "output": "stdout"},
```
jezdez, line 776:
```suggestion
        {"backend": "console", "output": "stdout"},
```

tests/base/test_context.py
```diff
@@ -759,3 +765,68 @@ def value(self, _):
...
+    """
+    Ensure that the ``reporters`` property returns the correct values when ``context.quiet``
+    is true.
+    """
+    args = SimpleNamespace(quiet=True)
+    reset_context((), args)
+
+    assert context.reporters == (
+        {
+            "backend": "stdlib",
```
jezdez, line 810:
```suggestion
            "backend": "console",
```

jezdez (approved):
Just concerned about the "stdlib" name, that has a particular meaning (e.g. the built-in set of libraries shipped with CPython), which isn't something that conda users may know, and it might be confusing.

travishathaway:
@jezdez and @beeankha,

Can I please get your approval once more. If this passes checks, feel free to automatically merge it upon completion. Otherwise, I will take care of it.

jezdez (approved):

jezdez:
@travishathaway @ForgottenProgramme I thought we wanted roll this back before 24.5.0?

travishathaway:
@jezdez,

Unfortunately, it looks like it slipped in the `24.5.x` release 🙈. I apologize for not staying on top of this.

PR #13735, opened 2024-03-26
Improve logging

zklaus:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13750, opened 2024-04-01
save certificate for use in tests

dholth:
### Description

An attempt to speed up the tests by not creating a new certificate all the time.

To address codspeed regressions in #13735 

### Checklist - did you ...


kenodegard:
you can also update the test to isolate the code to benchmark versus the setup/teardown:

```python
@pytest.mark.benchmark
def test(benchmark):
    # setup

    @benchmark
    def _():
        # code to benchmark

    # teardown
```

dholth:
Sounds good, looking at those functions that happened to be tagged as `benchmark`, a lot of what it's doing is test setup.
It would make more sense to benchmark single functions like `fetch.download_and_hash` or `repo.repodata()`. Will also have to consider whether those would mess with the "count function calls with a mocker" part of the test.

tests/adhoc_cert.py
jezdez, line 1: Couldn't we just depend on Werkzeug in the tests? Aren't we already using flask in the tests?

news/13750-test-certificate
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Use static ssl certificate in test suite to save time (#13750)
```
jezdez, line 19:
```suggestion
* Use static ssl certificate in test suite to save time. (#13750)
```

Issue #13671, opened 2024-03-08
Improve support for `accept-ranges: none` combined with `.partial` resume downloads suppport


dholth:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Conda 23.11 added [`.partial` package downloads.](https://github.com/conda/conda/pull/13231) which makes a range request to resume an interrupted package download.

When the web server returns `accept-ranges: none` (200 instead of 206 partial response), then conda should truncate the `.partial` file to 0 and download it again from the beginning.

Add unit test for this feature.

dholth:
#13653

dholth:
We have a flask and also a http.server based test server in the suite, and the simpler built-in Python one probably doesn't support accept-ranges. It wouldn't send accept-ranges: none but that probably doesn't matter for this test; conda would need to look for 206 partial content or 200 OK

PR #13785, opened 2024-04-09
Adding test for "Accept-Ranges: none" scenario with partial downloads

travishathaway:
### Description

Fixes: #13671 

### Checklist - did you ...

### DIFF
--- a/tests/fixtures_jlap.py
+++ b/tests/fixtures_jlap.py
@@ -53,6 +53,17 @@ def download_file(subdir, name):
     return flask.send_from_directory(Path(base, subdir), name)
 
 
+@app.route("/none-accept-ranges/")
+def none_accept_ranges():
+    """
+    Returns an empty request with "Accept-Ranges" set to "none"
+    """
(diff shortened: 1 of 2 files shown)


tests/fixtures_jlap.py
```diff
@@ -53,6 +53,17 @@ def download_file(subdir, name):
...
 
+@app.route("/none-accept-ranges/")
+def none_accept_ranges():
+    """
+    Returns an empty request with "Accept-Ranges" set to "none"
+    """
+    response = flask.Response("test content test content test content")
+    response.headers["Accept-Ranges"] = "none"
+
+    return response
```
dholth, line 64: This may work. We will have to check that flask ignores the range request headers, and returns all the bytes even if we ask for only some of the bytes.
The test suite also contains a `http_test_server` fixture based on Python's `simplehttpserver`. It wouldn't return `accept-ranges: none`, which would be OK since we ignore that response header, but it may ignore the range request headers.

dholth (approved):
Thanks!

PR #13818, opened 2024-04-19
Add reporter handler, output handler and reporter manager class structure to conda

### LINKED ISSUE #13814: Create the class structure used for reporter handlers and output handlers
### What's the idea?

In order to support a way for us to grow and customize the way output is rendered via the the `reporters` setting, we will need to create a class structure that can subsequently be turned into plugin hooks.

We have decided to create to different types of classes for this:

- `ReporterHandler` - this will hold many methods that correspond to widgets of output (e.g. `detail_view`, `table_view`). These classes define how it will look.
- `OutputHandler` - this will hold a single `render` method and will determine how the output is rendered. For example, this could be to stdout out but also to a file or even a network stream.

### What should be done?

- [x] Figure out which module to place this code in
- [x] Write tests for the code and ensure it has 100% coverage

### Additional context

More information about this class design can be found in the following document:

- https://hackmd.io/Ii7CzdaMTxGZRWVKR2iu9A

ForgottenProgramme:
### Description

This PR adds the `ReporterHandler` and `OutputHandler` abstract classes and the `ReporterManager` class to conda. 
This structure will be used to update the output of various conda commands to use the new reporter setting. 

resolves #13814. 

### Checklist - did you ...

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -694,7 +694,7 @@ class time_recorder(ContextDecorator):  # pragma: no cover
 
 
 Reporter = Dict[str, Union[bool, int, str]]
-DetailRecord = dict[str, Union[str, int, bool]]
+DetailRecord = Dict[str, Union[str, int, bool]]
 
 
 class ReporterHandlerBase(ABC):
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -23,7 +23,7 @@ from logging import CRITICAL, NOTSET, WARN, Formatter, StreamHandler, getLogger
 from os.path import dirname, isdir, isfile, join
 from threading import Event, Lock, RLock, Thread
 from time import sleep, time
-from typing import Dict, Union
+from typing import TYPE_CHECKING
 
 from ..auxlib.decorators import memoizemethod
 from ..auxlib.logz import NullHandler
@@ -693,8 +693,9 @@ class time_recorder(ContextDecorator):  # pragma: no cover
             os.makedirs(dirname(self.record_file))
 
 
-Reporter = Dict[str, Union[bool, int, str]]
-DetailRecord = Dict[str, Union[str, int, bool]]
+if TYPE_CHECKING:
+    Reporter = dict[str, bool | int | str]
+    DetailRecord = dict[str, str | int | bool]
 
 
 class ReporterHandlerBase(ABC):
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -2,11 +2,14 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Common I/O utilities."""
 
+from __future__ import annotations
+
 import json
 import logging
 import os
 import signal
 import sys
+from abc import ABC, abstractmethod
 from collections import defaultdict
 from concurrent.futures import Executor, Future, ThreadPoolExecutor, _base, as_completed
 from concurrent.futures.thread import _WorkItem
@@ -20,6 +23,7 @@ from logging import CRITICAL, NOTSET, WARN, Formatter, StreamHandler, getLogger
 from os.path import dirname, isdir, isfile, join
 from threading import Event, Lock, RLock, Thread
 from time import sleep, time
+from typing import Union
 
 from ..auxlib.decorators import memoizemethod
 from ..auxlib.logz import NullHandler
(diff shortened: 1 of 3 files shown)


conda/common/io.py
```diff
@@ -689,6 +693,114 @@ def _ensure_dir(self):
...
+DetailRecord = Dict[str, Union[str, int, bool]]
+
+
+class ReporterHandlerBase(ABC):
+    """
+    Base class for all reporter handlers
+    """
+
+    @abstractmethod
+    def detail_view(self, data: DetailRecord, **kwargs) -> str: ...
```
travishathaway, line 706: Can we add comments for these two methods? (`detail_view` and `string_view`) It would be good to know why they exists and how they are expected to be implemented. When authors are creating their own sub-classes based on this class they will be looking to these comments for guidance.

conda/common/io.py
```diff
@@ -689,6 +693,114 @@ def _ensure_dir(self):
...
+
+    @abstractmethod
+    def string_view(self, data: str, **kwargs) -> str: ...
+
+
+class ConsoleHandler(ReporterHandlerBase):
```
travishathaway, line 712: A comment here would be good to explain how this particular reporter handler behaves.

conda/common/io.py
```diff
@@ -689,6 +693,114 @@ def _ensure_dir(self):
...
+            row_header = header.ljust(longest_header, " ")
+            table_str += f"{row_header} : {value}\n"
+
+        return table_str
+
+    def string_view(self, data: str, **kwargs) -> str:
+        return data
+
+
+class JSONHandler(ReporterHandlerBase):
```
travishathaway, line 727: It would be nice to have a comment here as well to briefly describe how this behaves.

tests/common/test_io.py
```diff
@@ -80,3 +92,64 @@ def test_attach_stderr_handler():
     assert c.stdout == ""
     assert "test message" in c.stderr
     assert debug_message in c.stderr
+
+
+def test_console_handler():
```
travishathaway, line 97: Please add a comment to explain why this test exists. Please do this for the other tests too.

travishathaway (changes requested):
Looking good. I would just like to see some more comments in the code.

conda/common/io.py
```diff
@@ -689,6 +693,130 @@ def _ensure_dir(self):
...
+    Reporter handler for standard output (classic output).
+    """
+
+    def detail_view(self, data: DetailRecord, **kwargs) -> str:
+        table_str = ""
+        longest_header = len(sorted(data.keys(), key=len).pop())
+
+        for header, value in data.items():
+            row_header = header.ljust(longest_header, " ")
+            table_str += f"{row_header} : {value}\n"
```
kenodegard, line 731:
```suggestion
            table_str += f"{row_header:<{longest_header}} : {value}\n"
```

kenodegard (changes requested):
Let's add some more tests and get the code coverage for this new code closer to 100%.

conda/common/io.py
```diff
@@ -689,6 +693,130 @@ def _ensure_dir(self):
             os.makedirs(dirname(self.record_file))
 
 
+Reporter = Dict[str, Union[bool, int, str]]
+DetailRecord = Dict[str, Union[str, int, bool]]
```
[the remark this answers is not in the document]
kenodegard: I see you tried to change this and got test failures so you reverted the change.

A few things to remember with regards to type annotations:

- When we say `from __future__ import annotations` this changes how type annotations are _stored_. Without this import, type annotations will always be evaluated at runtime irrelevant of whether we care to perform any type checking operations. This makes it difficult to use modern type annotations when we're still supporting older versions of Python (as you discovered). When we include `from __future__ import annotations` the type annotations are stored internally as a _raw string_ and hence has no impact on our runtime code (even when running older versions of Python).

- Unfortunately, only the type annotation syntax itself is impacted by the future import. E.g.:

    ```python
    variable: int

    def function() -> str:
        ...
    ```

    So your type aliases (`Reporter` and `DetailRecord`), which are plain Python assignments, are _not_ stored as raw strings even though we include `from __future__ import annotations`.

- To resolve this issue we want to ensure that our type aliases are only evaluated when we are type checking. This is best done by putting our type annotations into a type checking block:

    ```python

    if TYPE_CHECKING:
        Reporter = dict[str, bool | int | str]
        DetailRecord = dict[str, str | int | bool]
    ```

    See other examples already in conda: https://github.com/conda/conda/blob/be5adf514550260952f20a807748f7f3ae297f10/conda/deprecations.py#L12-L23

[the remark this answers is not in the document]
travishathaway: Thanks for figure this out @kenodegard. I was wondering why that wasn't working as expected.

conda/common/io.py
kenodegard: If the suggestion above looks magical take a look at https://docs.python.org/3/library/string.html#format-specification-mini-language for more details on the format string mini-language:

Screenshot 2024-04-24 at 17 52 27

kenodegard (changes requested):

conda/common/io.py
```diff
@@ -689,6 +693,131 @@ def _ensure_dir(self):
...
+        self._reporters = reporters
+        self._reporter_handlers = {
+            "console": ConsoleHandler(),
+            "json": JSONHandler(),
+        }
+        self._output_handlers = {
+            "stdout": StdoutHandler(),
+        }
+
+    def render(self, name, data, **kwargs) -> None:
```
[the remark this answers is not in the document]
ForgottenProgramme: Hey, Ken!
I now understand your suggested changes better. Thank you. 

>I expect render to know that a str is a string and how to print it (using string_view). It should also know that a dict is a mapping and how to print it (using detail_view).

The render method is intended to point to a type of "template". The methods that we’ve defined on the ReporterHandler classes are these templates. For now, we only have "string_view" and "detail_view" which depend on the data type they receive. 
But not all of these "templates"/"views" will necessarily depend on the data type passed in. Therefore we do want to link which "view" method gets called to the data type.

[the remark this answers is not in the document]
kenodegard: > But not all of these "templates"/"views" will necessarily depend on the data type passed in.

If that's the case, a compelling example is necessary to sway my opinion. The implementation and the arguments for why we want it this way look and sound like we are over engineering. It's ok for us to over engineer but we need to have documented examples for why we are doing this.

In any case, I expect the vast majority of renders to be based on the data type. So the format/template name should at least be an optional key and when not provided it is determined from the data type.

PR #13842, opened 2024-04-24
Implement transaction action plugins

### LINKED ISSUE #13795: Plugin system: transaction action plugins
### Checklist

### What is the idea?

Add a plugin hook to allow developing new transaction actions.

### Why is this needed?

The transaction system is the natural fit for many post-solve operations that require access to the recently solved environment. I'm seeing this need in `conda-pip` and `conda-checkpoints`, for example.

The transaction API is also a natural fit for rollbacks and other operations that affect the disk (like placing files in strategic paths with `EXTERNALLY-MANAGED` or menuinst shortcuts).

### What should happen?

A new `transaction` plugin hook could be added to classes that inherit the [`_Action`](https://github.com/conda/conda/blob/d7167adc5ac2f96ea1f56fd193610556a77c5d78/conda/core/path_actions.py#L84) API.

### Additional Context

This is similar to [post-solves](https://github.com/conda/conda/pull/13053) in a way, but not quite. Post-solves run _before_ the transaction is created, don't have rollbacks and are more designed for additional checks, not operations in disk.

jaimergp:
### Description

Closes #13795 

### Checklist - did you ...

### DIFF
--- a/conda/plugins/hookspec.py
+++ b/conda/plugins/hookspec.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
         CondaSetting,
         CondaSolver,
         CondaSubcommand,
+        CondaTransactionAction,
         CondaVirtualPackage,
     )
 
@@ -331,4 +332,49 @@ class CondaSpecs:
                    parameter=PrimitiveParameter("default_value", element_type=str),
                    aliases=("example_option_alias",),
                )
+        """
+
+    @_hookspec
+    def conda_transaction_action(self) -> Iterable[CondaTransactionAction]:
+        """
+        Register new transaction action
+
+        The example below defines a simple action that places a marker file in prefix.
+
+        **Example:**
+
+        .. code-block:: python
+
+        from conda import plugins
+        from conda.core.path_actions import CreateInPrefixPathAction
+        from conda.gateways.disk.update import touch
+        from conda.gateways.disk.delete import rm_rf
+
+        @plugins.hookimpl
+        def conda_transaction_action():
+            yield plugins.CondaTransactionAction(
+                name="marker_file_action",
+                action=MarkerFileAction,
+            )
+
+        class MarkerFileAction(CreateInPrefixPathAction):
+            @classmethod
+            def create_actions(
+                cls, transaction_context, package_info, target_prefix, requested_link_type
+            ):
+                return (cls(transaction_context, package_info, target_prefix),)
+
+            def __init__(self, transaction_context, package_info, target_prefix):
+                super().__init__(
+                    transaction_context, package_info, None, None, target_prefix, "my_marker"
+                )
+                self._file_created = False
+
+            def execute(self):
+                log.log(TRACE, "touching marker file %s", self.target_full_path)
+                self._file_created = touch(self.target_full_path)
+
+            def reverse(self):
+                if self._file_created:
+                    rm_rf(self.target_full_path)
         """
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from typing import Callable
 
     from ..common.configuration import Parameter
+    from ..core.path_actions import _Action
     from ..core.solve import Solver
     from ..models.match_spec import MatchSpec
     from ..models.records import PackageRecord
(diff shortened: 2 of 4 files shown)


jaimergp:
Currently deciding where to add hooks. I started by adding a single generic `CondaTransactionAction`, but that API assumes very little about _when_ it will be called and with which parameters.

The meat happens in `UnlinkLinkTransaction`. Roughly we have something like:

* Actions called per package record
* Actions called per path individually (each file and directory)
* Actions called for several paths in bulk
* Actions called per environment

These are duplicated for unlink & link. So we have something like eight possible hooks:

* Unlink package record hook
* Unlink individual path hook
* Unlink multiple paths hook
* Unlink environment hook
* Link package record hook
* Link individual path hook
* Link multiple paths hook
* Link environment hook

The API is split between Actions (unaware of the transaction order) and ActionGroups (which are directly handled by the transaction).

I'll need to prototype a bit more to see how this looks like.

conda/plugins/hookspec.py
```diff
@@ -332,3 +333,48 @@ def conda_settings():
                    aliases=("example_option_alias",),
                )
         """
+
+    @_hookspec
+    def conda_transaction_action(self) -> Iterable[CondaTransactionAction]:
```
jezdez, line 338: Per pluggy parlance hooks should be plural since they are implemented not as singular instances but always as a list, or in our case generators as we're using `yield`

conda/plugins/types.py
```diff
@@ -210,3 +211,19 @@ class CondaSetting:
     description: str
     parameter: Parameter
     aliases: tuple[str, ...] = tuple()
+
+
+@dataclass
+class CondaTransactionAction:
```
jezdez, line 217: action action 💥

jaimergp:
Ay further comments here @jezdez (or anyone from @conda/conda-core)? If not I think I'll proceed with the solution described at https://github.com/conda/conda/pull/13842#issuecomment-2074827357.

jezdez:
> Ay further comments here @jezdez (or anyone from @conda/conda-core)? If not I think I'll proceed with the solution described at [#13842 (comment)](https://github.com/conda/conda/pull/13842#issuecomment-2074827357).

> Still eight hooks, which don't guarantee execution order, and would force plugins to inspect the UnlinkLinkTransaction instance to find the objects they need, which means more potential for API breakage.

That's going to set in stone the un-API that `UnlinkLinkTransaction` is, which isn't really about just unlinking IIUC. `UnlinkLinkTransaction` is totally named wrong, it's in fact just _the_ Transaction, capital T.

The tricky thing is that we're trying extend different API levels, that are stacked and have different constraints, some related to the transaction order, some not. That's going to make things a lot harder for plugin authors to understand how to build reasonable plugins without breaking the API.

We don't neccearily need a pre and post step for every action (3. below), e.g. calling a hook plugin for every file that is being handled sounds risky in terms of performance for example. While calling a plugin with the list of files that the action is about to be taken, that might not be as problematic.

There are three level all weirdly intertwined:

1. We derive the first level from the values the transation is handling on a high level:

- CondaTransaction(Pre|Post)Unlink (gets list of package records)
- CondaTransaction(Pre|Post)Link (gets list of package records)
- CondaTransaction(Pre|Post)Removal (gets list of matchspec)
- CondaTransaction(Pre|Post)Update (gets list of matchspec)

Those are just for data preperation, I wonder if those could also be done with pre and post solver plugins?

> The other alternative is to forget about the `Action` API and simply add a series of pre/post hooks in each part of the `LinkUnlinkTransaction` API:
> 
> * `pre_prepare`, `post_prepare`
> * `pre_verify`, `post_verify`
> * `pre_execute`, `post_execute`
> * `pre_reverse`, `post_reverse`
> 
> Still eight hooks, which don't guarantee execution order, and would force plugins to inspect the `UnlinkLinkTransaction` instance to find the objects they need, which means more potential for API breakage.

2. That's the second level, which will all get a transaction instance passed (previously called "LinkUnlinkTransaction".

3. And then in addition we port the `(Path)Action` type to being a plugin based interface, we already have specific plugins built-in (that can't be overwritten): "unlink" , "remove_menus", "unregister", "link", "entry_point", "compile", "make_menus", "record", "register", each could live in a separate module (`conda.transaction.actions.*`), implement seperate phases of the transaction and plugin authors could add aditional actions.

jaimergp:
Thanks for the detailed reply @jezdez! I need to think a little bit more about it, but for now:

> Those are just for data preperation, I wonder if those could also be done with pre and post solver plugins?

The limitation I see there is that the post-solve hook is run _before_ the IO work has started. For example, directories might not be in place yet, nor the artifacts have been downloaded, etc. For example, if I want to write some files, I could anticipate _where_ to write it (but we don't have any info about the target prefix the solver was passed!). However, if for some reason the transaction fails (bad download, disk full, etc), there won't be any rollbacks for the post-solve actions. This hook is for me more useful to prevent the transaction from happening, before IO work begins.


So in a nutshell, what do you think it's the sweet spot here? I think we can forget about your level (1), focus on number (2), and then at some point tackle (3) in a refactor.

jezdez:
Thinking about your question more, I think I mostly see this as a pipeline problem with an increasing level of data available.

Level 1 is about generic transaction action events
Level 2 is about specific transaction executions
Level 3 is about specific transaction phases.

It seems most useful to work on level 2 first (with a cleaned up transaction instance API for easy inspection in plugin hook implementations), then level 1, then level 3.

jaimergp:
Closed by https://github.com/conda/conda/pull/14754

PR #13211, opened 2023-10-18
Convert `path` to `pathlib.Path` before adding to `PrefixData` cache

kenodegard:
### Description

Weird things occur when the cache keys are a blend of `str` and `pathlib.Path`. Standardize them such that they're always `pathlib.Path`.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

Issue #12421, opened 2023-03-01
Mark `conda/plan.py` for Deprecation (and do related refactoring work as necessary)


beeankha:
### What is the idea?

There is [a `TODO` comment](https://github.com/conda/conda/blob/main/conda/plan.py#L42) marking the `conda/plan.py` module for deprecation that got added in the pull request titled ["separate conda package metadata under new root-level key in repodata (#8639)"](https://github.com/conda/conda/pull/8639).

Deprecating/removing that module (and [related test file](https://github.com/conda/conda/blob/main/tests/test_plan.py)) entirely is not a straightforward task, as there are references to the contents of `conda/plan.py` in files such as [`conda/misc.py`](https://github.com/conda/conda/blob/main/conda/misc.py) and [`conda/exports.py`](https://github.com/conda/conda/blob/main/conda/exports.py). This deprecation will require:

```[tasklist]
- [x] Investigations into which files contain references to any functions/methods in [`conda/plan.py`](https://github.com/conda/conda/blob/main/conda/plan.py)
- [x] Figuring out what sort of refactoring efforts are required as a result of those references
```

jezdez:
@beeankha Is there anything we need to do about this ticket?

beeankha:
@jezdez nothing in particular is blocking this work afaik, we just need to get started on it; if it's more complicated than anticipated we can move it to the 24.7.x milestone. I'll do at least the preliminary research (task 1) this week!

beeankha:
Discoveries for Step 1 of this task:

* **[`print_dists()`](https://github.com/conda/conda/blob/main/conda/plan.py#L45):** 
  * only used within `conda/plan.py`, in the `display_actions()` function

* **[`display_actions()`](https://github.com/conda/conda/blob/main/conda/plan.py#L56):**
  * used in `conda/exports.py` in [`display_actions()`](https://github.com/conda/conda/blob/main/conda/exports.py#L171) (it's [imported as `_display_actions` with a leading underscore added](https://github.com/conda/conda/blob/main/conda/exports.py#L67)
[...]

PR #13869, opened 2024-04-26
Stop importing from `conda/exports.py`

beeankha:
While investigating a fix for https://github.com/conda/conda/pull/13860, an instance of [`reset_context()` was found in `conda/exports.py`](https://github.com/conda/conda/blob/main/conda/exports.py#L82); in order to avoid resetting the context unintentionally, as well as to avoid utilizing the `conda/exports.py` module (which is meant to be used by conda-build vs conda), this PR attempts to remove/edit/update any import statement in the conda codebase that is importing from `conda/exports.py`. 

Certain cases were left alone, such as the imports from `exports.py` in `tests/test_create.py` in [`test_list_with_pip_no_binary`](https://github.com/conda/conda/pull/13211/files#diff-1e27c53a97d6c9242ea92544453665b553fe9fdf63f08717dc915f80e76fd60cR596) and [`test_rm_rf`](https://github.com/conda/conda/pull/13211/files#diff-1e27c53a97d6c9242ea92544453665b553fe9fdf63f08717dc915f80e76fd60cR651), since those are specifically related to regression testing.

Also related to the following deprecation issue:
- https://github.com/conda/conda/issues/12421


schuylermartin45 (commented):
Looks reasonable to me, but I am not familiar enough with the implications of this change.

jezdez:
@beeankha The description of the PR doesn't match the content of the PR, since it changes only things in `conda.plan`, but not `conda.exports`. Could we update that somehow?

beeankha:
> @beeankha The description of the PR doesn't match the content of the PR, since it changes only things in conda.plan, but not conda.exports. Could we update that somehow?

I agree that the wording is confusing; I meant we want to remove any import statements that are importing anything from the `conda/exports.py` module. I updated the title of this PR, hopefully that helps a bit. I'll see if I can edit the description to be a bit more clear.

jezdez (approved):

kenodegard (dismissed):

kenodegard (approved):

PR #13880, opened 2024-05-01
Add `conda.core.index.Index` as a faster drop-in replacement of the realized dictionary index.

zklaus:
### Description

This was triggered by https://github.com/conda/conda-build/issues/5154.
The goal is to introduce an Index API as a clean interface for the index, so as to allow well-defined interactions and avoid direct manipulation of Solver._index.
Currently, the index is handled as a simple dict, which is manipulated in various places to update with changed state.
This is particularly relevant in conda-build, where the index needs to be updated after the creation of new artifacts.
Introducing a standardized API makes this reloading more explicit.
The class `Index` also follows the spirit of `PackageRecordList` to avoid instantiating `PackageRecord`s where not necessary, yielding both memory and time savings.

Additionally, this introduces memray into the testing facilities. With this PR, it becomes possible to use a `pytest.mark.memray` marker, which will make the test run in the `linux-memray` job (in addition to any other runs that may occur) under the observation of the pytest-memray plugin, which allows us to track memory usage.
Currently, this only leads to a little summary blob in the test logs; a future PR may expand upon this with persistent tracking of memory requirements similar to the benchmarking we do with codspeed.

### Checklist - did you ...


dholth:
How does this compare to https://github.com/conda/conda/blob/be5da40dc6f18de7d0690afa1fefff13955ef4ab/conda/core/subdir_data.py#L92

zklaus:
@dholth, I added a bit of explanation to the PR description. Basically, it is inspired by `PackageRecordList` and tries to bring the lazy `PackageRecord` instantiation from there (and the related `SubdirData` and friends) to the index.

Tracking memory usage and time with `pytest-memray`, the current dict approach uses 50s and
```
Allocation results for tests/core/test_index.py::test_get_index_lazy at the high watermark

         📦 Total memory allocated: 1.8GiB
         📏 Total allocations: 104
         📊 Histogram of allocation sizes: |█▇▂▇▁|
         🥇 Biggest allocating functions:
                - raw_decode:/opt/conda/lib/python3.12/json/decoder.py:353 -> 1.1GiB
                - __hash__:/workspaces/conda-workspace/conda/conda/models/records.py:300 -> 182.5MiB
                - __set__:/workspaces/conda-workspace/conda/conda/auxlib/entity.py:441 -> 124.0MiB
                - join:/workspaces/conda-workspace/conda/conda/common/url.py:315 -> 83.0MiB
                - _pkey:/workspaces/conda-workspace/conda/conda/models/records.py:293 -> 43.1MiB
```
and the new lazy approach 10s and
```
Allocation results for tests/core/test_index.py::test_get_index_lazy at the high watermark

         📦 Total memory allocated: 1.2GiB
         📏 Total allocations: 91
         📊 Histogram of allocation sizes: |▁█▂▁ |
         🥇 Biggest allocating functions:
                - raw_decode:/opt/conda/lib/python3.12/json/decoder.py:353 -> 822.1MiB
                - decode:<frozen codecs>:322 -> 243.6MiB
                - raw_decode:/opt/conda/lib/python3.12/json/decoder.py:353 -> 146.5MiB
                - join:/workspaces/conda-workspace/conda/conda/common/url.py:315 -> 5.0MiB
                - _process_raw_repodata:/workspaces/conda-workspace/conda/conda/core/subdir_data.py:489 -> 2.0MiB
```

i.e. a reduction by 40s and 600MB, though a bunch of integration tests are currently failing.

dholth:
That's phenomenal.

.github/workflows/tests.yml
```diff
@@ -393,6 +393,70 @@ jobs:
...
+        run: python -m conda info --verbose
+
+      - name: Conda Config
+        run: conda config --show-sources
+
+      - name: Conda List
+        run: conda list --show-channel-urls
+
+      - name: Run memray
+        run: $CONDA/envs/test/bin/pytest -k memray --memray
```
jaimergp, line 458:
```suggestion
        run: python -m pytest -k memray --memray
```

Does this not work?

zklaus: Yes, I think it does. It's just that I copied it from the benchmark one. However, in that case I think it's because of the additional codspeed runner. Good catch.

zklaus:
pre-commit.ci autofix

zklaus:
pre-commit.ci autofix

jezdez (changes requested):
This looks very promising, but IMO could go a little further in subsuming the supplement functions into `Index` instance methods (not static nor class methods).

The `get_reduced_index` could also use the `Index` internally instead of creating an ad-hoc dict instead.

tests/core/test_index.py
```diff
@@ -164,3 +166,82 @@ def test_basic_get_reduced_index():
...
+        "win-64": dict(
+            channel="conda-forge",
+            name="vim",
+            version="9.1.0356",
+            build="py312h275cf98_0",
+            build_number=0,
+        ),
+    }
+    subdir = PLATFORMS[(platform.system(), platform.machine())]
+    index = get_index(channel_urls=["conda-forge"], platform=subdir)
```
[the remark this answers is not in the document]
zklaus: This is to test the `prepend` feature, analogous to the other `test_get_index_xxx_platform` tests.

conda/core/index.py
```diff
@@ -148,7 +267,13 @@ def _supplement_index_with_prefix(index: dict[Any, Any], prefix: str) -> None:
     """
     # supplement index with information from prefix/conda-meta
     assert prefix
-    for prefix_record in PrefixData(prefix).iter_records():
+    if isinstance(index, Index):
+        return
+    if isinstance(prefix, PrefixData):
```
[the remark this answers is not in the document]
zklaus: Functionally, it doesn't matter, of course, due to the return in the first if branch. Semantically, the first "if" deals with the index argument, the second if-else with the prefix argument, so on balance, I'd prefer to keep it as is. Does that work for you?

zklaus:
Thanks for the quick review, @jezdez!

> This looks very promising, but IMO could go a little further in subsuming the supplement functions into `Index` instance methods (not static nor class methods).

Dang, I already had the deprecations in place :smile: 

I removed them again because this is designed now as a drop-in replacement and it can be difficult to know what other code might rely on functions like these. For example, we have https://github.com/conda/conda-libmamba-solver/blob/1ac0aee96c4618c29ae2d2491fdc2ccd2b3305d4/conda_libmamba_solver/state.py#L186.

But if you are happy to remove them, then so am I.

We could even go further. Iterating an index is always expensive, particularly for large channels, so why not deprecate that? With the `Index` class in place we could to that by deprecating `.__iter__`, `.keys`, `.values`, `.items`.

> 
> The `get_reduced_index` could also use the `Index` internally instead of creating an ad-hoc dict instead.
Sure. Let me add that.

jezdez:
> Thanks for the quick review, @jezdez!
> 
> > This looks very promising, but IMO could go a little further in subsuming the supplement functions into `Index` instance methods (not static nor class methods).
> 
> Dang, I already had the deprecations in place 😄
> 
> I removed them again because this is designed now as a drop-in replacement and it can be difficult to know what other code might rely on functions like these. For example, we have https://github.com/conda/conda-libmamba-solver/blob/1ac0aee96c4618c29ae2d2491fdc2ccd2b3305d4/conda_libmamba_solver/state.py#L186.

To be clear, we should still provide stub functions that are properly deprecated per CEP-9 that call `Index` behind the scenes. But for CLS specifically we maintain the API, so this is relatively painless by wrapping it into a try/except or similar techniques of inspecting conda's version and using a different import, or wrapping function. There are a few ways to do this.

> But if you are happy to remove them, then so am I.
> 
> We could even go further. Iterating an index is always expensive, particularly for large channels, so why not deprecate that? With the `Index` class in place we could to that by deprecating `.__iter__`, `.keys`, `.values`, `.items`.

Migrating from dict with these methods to UserDict implies the methods remain, so not sure if deprecating the iterator API makes sense

> > The `get_reduced_index` could also use the `Index` internally instead of creating an ad-hoc dict instead.
> > Sure. Let me add that.

news/13880-lazy-index
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Add `conda.core.index.Index` as drop-in replacement of realized dictionary index. (#13880)
```
dholth, line 3:
```suggestion
* Add `conda.core.index.Index` as a faster drop-in replacement of realized dictionary index. (#13880)
```

zklaus:
pre-commit.ci autofix

conda/core/index.py
```diff
@@ -55,31 +55,46 @@ def __init__(
...
         self.track_features = context.track_features
+        self.add_system = add_system
+        self.system_packages = {
+            (
+                rec := _make_virtual_package(
+                    f"__{package.name}", package.version, package.build
+                )
+            ): rec
+            for package in context.plugin_manager.get_virtual_packages()
+        }
```
jaimergp, line 97: I think this could be a property or a method. Doing this in initialization regardless the value of `add_system` might be expensive (e.g. the `__cuda` package has a non negligible overhead).

zklaus:
To give a bit of an update: It took me some time last week to make this work with `get_reduced_index`, which uncovered some bugs and rough edges. Now it works, but as you can see from the codspeed comment above this results in some unacceptable slow down.
Figuring out why that is, it ultimately comes down to the combination of the solver doing a bunch of iterating over the packages, for example [`Resolve.__init__`](https://github.com/conda/conda/blob/49f2614c5d0e9092ba199ea7adb9a0bc94aedb80/conda/resolve.py#L120) has no less than 5 full iterations over the index, and the size of the lazy index. Since for iterating the lazy `Index` behaves exactly as the current index with a realized dictionary, this is of course much slower with the full index compared to the reduced index. 
I spent some time to try and remove the iterating from the solver, but ultimately gave up on that because it is too much, out-of-scope for this PR, and possibly to invasive a change.
Let's try to add a reduction to the `Index` class instead.

conda/core/index.py
```diff
@@ -48,10 +49,360 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        add_system=False,
+    ) -> None:
+        if use_local:
+            channels = ["local"] + list(channels)
+        if prepend:
+            channels += context.channels
+        self._channels = channels
+        if subdirs:
+            if platform:
+                log.warn("subdirs is %s, ignoring platform %s", subdirs, platform)
```
jezdez, line 73:
```suggestion
                log.warning("subdirs is %s, ignoring platform %s", subdirs, platform)
```

zklaus:
Note (for discussion?): This PR does remove the implicit initialization of the logging from `get_index`. If you relied on that, you should probably call `initialize_logging` explicitly.

jaimergp:
> Note (for discussion?): This PR does remove the implicit initialization of the logging from `get_index`. If you relied on that, you should probably call `initialize_logging` explicitly.

Good catch! Let's put this in the `news/` file so people see it in the changelog.

zklaus:
pre-commit.ci autofix

zklaus:
pre-commit.ci autofix

zklaus:
pre-commit.ci autofix

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+
+    There are four types of sources for package information, namely
+
+    Channels
+        represent packages available from standard sources identified with a url, mostly online,
+        but can also be on a local filesystem using the `file://` scheme.
+        Programatically, channels are represented by :class:`conda.models.Channel`, their data
+        is fetched using :class:`conda.index.subdir_data.SubdirData`.
+
+        For more information see :ref:`condepts-channels`.
```
jaimergp, line 72:
```suggestion
        For more information see :ref:`concepts-channels`.
```

I think?

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
 
 
+class Index(UserDict):
+    """The ``Index`` provides information about available packages from all relevant sources.
+
+    There are four types of sources for package information, namely
+
+    Channels
+        represent packages available from standard sources identified with a url, mostly online,
+        but can also be on a local filesystem using the `file://` scheme.
```
jaimergp, line 68:
```suggestion
        but can also be on a local filesystem using the ``file://`` scheme.
```

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        use_local=False,
+        use_cache=None,
+        prefix=None,
+        repodata_fn=context.repodata_fns[-1],
+        add_system=False,
+    ) -> None:
+        if use_local:
+            channels = ["local"] + list(channels)
+        if prepend:
+            channels += context.channels
```
jaimergp, line 105: Is this prepending or appending? 🤔

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            urls = Channel(channel).urls(True, subdirs)
+            check_allowlist(urls)
+            expanded_channels = [Channel(url) for url in urls]
+            self.channels[channel] = [
+                SubdirData(expanded_channel, repodata_fn=repodata_fn)
+                for expanded_channel in expanded_channels
+            ]
+            self.expanded_channels.extend(expanded_channels)
+        LAST_CHANNEL_URLS.clear()
+        LAST_CHANNEL_URLS.extend(self.expanded_channels)
```
jaimergp, line 126: Can you add a note about this global state bit and why is it needed? We should aim to remove it at some point.

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            self.prefix_path = prefix.prefix_path
+        else:
+            self.prefix_path = prefix
+        self._prefix_data = None
+        self.use_cache = True if use_cache is None and context.offline else use_cache
+        self.track_features = context.track_features
+        self.add_system = add_system
+
+    def _make_virtual_package(
+        self, name: str, version: str | None = None, build_string: str | None = None
```
jaimergp, line 139: Public and static?

```suggestion
    @staticmethod
    def make_virtual_package(
       name: str, version: str | None = None, build_string: str | None = None
```

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            }
+        return self._features
+
+    @property
+    def prefix_data(self):
+        if self._prefix_data is None and self.prefix_path:
+            self._prefix_data = PrefixData(self.prefix_path)
+        return self._prefix_data
+
+    def reload(self, prefix=False):
```
jaimergp, line 193: What is `prefix` here for?

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        if prefix:
+            if self.prefix_data:
+                self.prefix_data.reload()
+
+    def __repr__(self):
+        channels = ", ".join(self.channels.keys())
+        return f"Index(channels=[{channels}])"
```
jaimergp, line 200:
```suggestion
        return f"<Index(channels=[{channels}])>"
```

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+    def data(self):
+        try:
+            return self._data
+        except AttributeError:
+            self._realize()
+            return self._data
+
+    @data.setter
+    def data(self, value):
+        self._data = value
```
jaimergp, line 225: At what point do we need to override the data via the API? Not sure if this is a reasonable use case.

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+
+    def _supplement_index_dict_with_prefix(self, index_dict):
```
jaimergp, line 227: Do we need this as an instance method?

```suggestion
    @staticmethod
    def supplement_index_dict_with_prefix(index_dict):
```

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+                # If the package is not in the repodata, use the local data.
+                # If the channel is known but the package is not in the index, it
+                # is because 1) the channel is unavailable offline, or 2) it no
+                # longer contains this package. Either way, we should prefer any
+                # other version of the package to this one. On the other hand, if
+                # it is in a channel we don't know about, assign it a value just
+                # above the priority of all known channels.
+                index_dict[prefix_record] = prefix_record
+
+    def _supplement_index_dict_with_cache(self, index_dict: dict[Any, Any]) -> None:
```
jaimergp, line 264: Same here.

conda/core/index.py
```diff
@@ -51,7 +58,457 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            prefix,
+            repodata_fn,
+            add_system,
+        )
+        self.specs = specs
+        self._derive_reduced_index()
+
+    def __repr__(self):
+        channels = ", ".join(self.channels.keys())
+        return f"ReducedIndex(spec={self.specs}, channels=[{channels}])"
```
jaimergp, line 415:
```suggestion
        return f"<ReducedIndex(spec={self.specs}, channels=[{channels}])>"
```

news/13880-lazy-index
```diff
@@ -0,0 +1,21 @@
+### Enhancements
+
+* Add `conda.core.index.Index` as a faster drop-in replacement of realized dictionary index. (#13880)
+  Note: This removes the implicit initialization of the logging from `get_index`. If you relied on that, you should probably call `conda.gateways.logging.initialize_logging` explicitly.
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
```
jaimergp, line 10: I think we have deprecated more things in this PR. Like `get_reduced_index`. Double check.

conda/core/index.py
jezdez: Please no more static methods, they are a code smell of a weak architecture, especially in "core" code. We can easily add a separate utility function if needed. Can you elaborate why this would be needed to be a public, static method?

conda/core/index.py
jezdez: As as above, let's use the concrete class name here.

conda/core/index.py
jaimergp: This was initially a private function in the top module that was moved here as part of the review, but it requires no state of the class to work. IOW, I would need to instantiate the class just to use the method and discard it. But anyway, Klaus and I talked about it and then convened that the best place for this is to have alternative constructors in `PackageRecord` (there's also a `make_feature_record` function somewhere).

zklaus: Jupp. I made the change in https://github.com/conda/conda/pull/13880/commits/191528e799bcfb86360c760702d77fde776a4040 and it has the nice side-effect of addressing two `TODO`s.

jezdez: Nice, I like that change much better, thanks for talking about it!

jezdez (approved):

conda/core/index.py
jezdez: Oh, good point that's.. not ideal variable naming, the code currently appends.

zklaus: So this had me confused as well. A little explanation: Currently, `get_index` produces a simple dictionary, which it does by going through the list of channels and adding *or overwriting* packages. That means that channels later in the list effectively have higher priority for equal packages. In the new code this is done by traversing the list of channels in `__getitem__` in reversed order (l.247).

Effectively this code is inlining `calculate_channel_urls` and `all_channel_urls`.

conda/core/index.py
zklaus: I am fairly certain that I was forced to add this for a `UserDict` method. `UserDict.copy`?

conda/models/records.py
```diff
@@ -244,38 +244,105 @@ class PathsData(Entity):
...
+    """:class:`conda.models.channel.Channel`: The channel where the package can be found."""
     subdir = SubdirField()
+    """str: The subdir, i.e. ``noarch`` or a platform (``linux-64`` or similar).
+
+    Part of the :attr:`_pkey`.
+    """
     fn = FilenameField(aliases=("filename",))
+    """str: The filename of the package.
+
+    Only part of the :attr:`_pkey` if :ref:`separate_format_cache <auto-config-reference>` is true (default: false).
```
jaimergp, line 298:
```suggestion
    Only part of the :attr:`_pkey` if :ref:`separate_format_cache <auto-config-reference>` is ``true`` (default: ``false``).
```

conda/core/index.py
```diff
@@ -51,7 +58,456 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        self,
+        channels=(),
+        prepend=True,
+        platform=None,
+        subdirs=None,
+        use_local=False,
+        use_cache=None,
+        prefix=None,
+        repodata_fn=context.repodata_fns[-1],
+        add_system=False,
```
jaimergp, line 100: Can we add type hints in these new API / methods?

conda/core/index.py
```diff
@@ -51,7 +58,456 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        specs,
+        channels=(),
+        prepend=True,
+        platform=None,
+        subdirs=None,
+        use_local=False,
+        use_cache=None,
+        prefix=None,
+        repodata_fn=context.repodata_fns[-1],
+        add_system=False,
```
jaimergp, line 396: Same here with the type hints.

conda/core/subdir_data.py
```diff
@@ -553,6 +554,11 @@ def _get_base_url(self, repodata: dict, with_credentials: bool = True) -> str:
         return self.url_w_subdir
 
 
+@deprecated(
+    "24.9",
+    "25.3",
+    addendum="Use `conda.core.models.records.PackageRecord.make_feature_record` instead.",
```
jaimergp, line 560: I am wondering whether we should really provide this new API at all given how "in-practice-deprecated" features have been for a while.

conda/models/records.py
```diff
@@ -244,38 +244,105 @@ class PathsData(Entity):
...
         default=None, required=False, nullable=True, default_in_dump=False
     )
+    """str: The download url of the package."""
     sha256 = StringField(
         default=None, required=False, nullable=True, default_in_dump=False
     )
+    """str: The sha256 checksum of the package."""
 
     @property
     def schannel(self):
```
jaimergp, line 321: I know this is not your fault, but what in the world is that `s` for 😂

conda/models/records.py
```diff
@@ -411,6 +478,46 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.metadata = set()
 
+    @classmethod
+    def make_feature_record(cls, feature_name):
```
jaimergp, line 482: Maybe we shouldn't really add this (as per the other comment above). Thoughts @jezdez?

jaimergp: Nevermind, classic solver still uses it. We need to keep it.

conda/core/subdir_data.py
zklaus: We (i.e. Jaime) noticed that the classic solver still needs it, so for now, it stays.

.github/workflows/tests.yml
```diff
@@ -393,6 +393,70 @@ jobs:
...
+    defaults:
+      run:
+        # https://github.com/conda-incubator/setup-miniconda#use-a-default-shell
+        shell: bash -el {0}  # bash exit immediately on error + login shell
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.12']
+    env:
+      REQUIREMENTS_TRUSTSTORE: ${{ contains('3.10|3.11|3.12', matrix.python-version) && '--file tests/requirements-truststore.txt' || '' }}
```
jezdez, line 412:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -393,6 +393,70 @@ jobs:
...
+
+      - name: Conda Install
+        run: >
+          conda install
+          --yes
+          --file tests/requirements.txt
+          --file tests/requirements-${{ runner.os }}.txt
+          --file tests/requirements-ci.txt
+          --file tests/requirements-s3.txt
+          ${{ env.REQUIREMENTS_TRUSTSTORE }}
```
jezdez, line 443:
```suggestion
          --file tests/requirements-truststore.txt
```

conda/core/subdir_data.py
jezdez, line 560:
```suggestion
    addendum="Use `conda.core.models.records.PackageRecord.feature` instead.",
```

conda/models/records.py
jezdez, line 482:
```suggestion
    def feature(cls, feature_name) -> PackageRecord:
```

conda/resolve.py
```diff
@@ -178,19 +178,21 @@ def __hash__(self):
...
         if filter is None:
             filter = {}
         else:
             filter.clear()
 
         filter.update(
-            {make_feature_record(fstr): False for fstr in self.trackers.keys()}
+            {
+                PackageRecord.make_feature_record(fstr): False
+                for fstr in self.trackers.keys()
```
jezdez, line 189:
```suggestion
                PackageRecord.feature(name): False
                for name in self.trackers.keys()
```

conda/resolve.py
```diff
@@ -178,19 +178,21 @@ def __hash__(self):
...
+            }
         )
         if features:
-            filter.update({make_feature_record(fstr): True for fstr in features})
+            filter.update(
+                {PackageRecord.make_feature_record(fstr): True for fstr in features}
```
jezdez, line 194:
```suggestion
                {PackageRecord.feature(name): True for name in features}
```

conda/resolve.py
```diff
@@ -810,7 +809,8 @@ def filter_group(_specs):
 
         # Determine all valid packages in the dependency graph
         reduced_index2 = {
-            prec: prec for prec in (make_feature_record(fstr) for fstr in features)
+            prec: prec
+            for prec in (PackageRecord.make_feature_record(fstr) for fstr in features)
```
jezdez, line 813:
```suggestion
            for prec in (PackageRecord.feature(name) for name in features)
```

conda/testing/helpers.py
```diff
@@ -147,7 +147,7 @@ def add_feature_records_legacy(index):
             all_features.update(rec.track_features)
 
     for feature_name in all_features:
-        rec = make_feature_record(feature_name)
+        rec = PackageRecord.make_feature_record(feature_name)
```
jezdez, line 150:
```suggestion
        rec = PackageRecord.feature(feature_name)
```

news/13880-lazy-index
```diff
@@ -0,0 +1,28 @@
...
+
+* Mark `conda.core.index.get_index` as pending deprecation. Use `conda.core.index.Index` instead. (#13880)
+* Mark `conda.core.index.get_reduced_index` as pending deprecation. Use `conda.core.index.ReducedIndex` instead. (#13880)
+* Mark `conda.core.index.fetch_index` as pending deprecation. Use `conda.core.index.Index` instead. (#13880)
+* Mark `conda.core.index._supplement_index_with_prefix` as pending deprecation. Use `conda.core.index.Index.reload` instead. (#13880)
+* Mark `conda.core.index._supplement_index_with_cache` as pending deprecation. Use `conda.core.index.Index.reload` instead. (#13880)
+* Mark `conda.core.index._supplement_index_with_features` as pending deprecation. Use `conda.core.index.Index.reload` instead. (#13880)
+* Mark `conda.core.index._supplement_index_with_system` as pending deprecation. Use `conda.core.index.Index.reload` instead. (#13880)
+* Mark `conda.core.index._make_virtual_package` as pending deprecation. Use `conda.models.records.PackageRecord.make_virtual_package` instead. (#13880)
+* Mark `conda.core.subdir_data.make_feature_record` as pending deprecation. Use `conda.models.records.PackageRecord.make_feature_record` instead. (#13880)
```
jezdez, line 20:
```suggestion
* Mark `conda.core.subdir_data.make_feature_record` as pending deprecation. Use `conda.models.records.PackageRecord.feature` instead. (#13880)
```

conda/core/subdir_data.py
```diff
@@ -553,6 +554,11 @@ def _get_base_url(self, repodata: dict, with_credentials: bool = True) -> str:
...
+)
 def make_feature_record(feature_name):
     # necessary for the SAT solver to do the right thing with features
     pkg_name = f"{feature_name}@"
```
jezdez, line 564: Let's just return `PackageRecord.feature(feature_name)` here to not duplicate the implementation and only leave a stub, basically.

conda/core/index.py
```diff
@@ -51,7 +58,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        Individual packages from the package are usually represented by :class:`conda.models.records.PackageCacheRecord`.
+
+    Virtual Packages
+        represent properties of the system, not actual conda packages in the normal sense. These are,
+        for example, system packages that inform the solver about the operating system in use, or
+        track features that can be used to steer package priority.
+
+        Individual virtual packages are represented by special :class:`conda.models.records.PackageRecord`,
+        see :meth:`conda.models.records.PackageRecord.make_virtual_package` and
+        :meth:`conda.models.records.PackageRecord.make_feature_record`.
```
jezdez, line 97:
```suggestion
        see :meth:`conda.models.records.PackageRecord.virtual_package` and
        :meth:`conda.models.records.PackageRecord.feature`.
```

conda/core/index.py
```diff
@@ -51,7 +58,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            if has_data:
+                self._supplement_index_dict_with_prefix()
+        if cache:
+            self._cache_entries = PackageCacheData.get_all_extracted_entries()
+            if has_data:
+                self._supplement_index_dict_with_cache()
+        if features:
+            self._features = {
+                (rec := PackageRecord.make_feature_record(feature)): rec
+                for feature in context.track_features
```
jezdez, line 252:
```suggestion
                (rec := PackageRecord.feature(track_feature)): rec
                for track_feature in context.track_features
```

conda/core/index.py
```diff
@@ -51,7 +58,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            }
+            if has_data:
+                self._data.update(self.features)
+        if system:
+            self._system_packages = {
+                (
+                    rec := PackageRecord.make_virtual_package(
+                        f"__{package.name}", package.version, package.build
+                    )
```
jezdez, line 261:
```suggestion
                    rec := PackageRecord.virtual_package(
                        f"__{package.name}", package.version, package.build
                    )
```

conda/core/index.py
```diff
@@ -51,7 +58,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        except AttributeError:
+            pass
+        try:
+            return self.system_packages[key]
+        except KeyError:
+            pass
+        if self.track_features and key.name.endswith("@"):
+            for feature in self.track_features:
+                if feature == key.name[:-1]:
+                    return PackageRecord.make_feature_record(feature)
```
jezdez, line 451:
```suggestion
            for track_feature in self.track_features:
                if track_feature == key.name[:-1]:
                    return PackageRecord.feature(track_feature)
```

conda/core/index.py
```diff
@@ -51,7 +58,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            self._supplement_index_dict_with_prefix()
+
+        # add feature records for the solver
+        known_features = set()
+        for rec in self._data.values():
+            known_features.update((*rec.track_features, *rec.features))
+        known_features.update(context.track_features)
+        for ftr_str in known_features:
+            rec = PackageRecord.make_feature_record(ftr_str)
+            self._data[rec] = rec
```
jezdez, line 613:
```suggestion
        for known_feature in known_features:
            rec = PackageRecord.feature(known_feature)
            self._data[rec] = rec
```

conda/core/index.py
```diff
@@ -231,10 +815,11 @@ def _supplement_index_with_features(
     :param features: A list of feature names to add to the index.
     """
     for feature in chain(context.track_features, features):
-        rec = make_feature_record(feature)
+        rec = PackageRecord.make_feature_record(feature)
```
jezdez, line 818:
```suggestion
        rec = PackageRecord.feature(feature)
```

conda/core/index.py
```diff
@@ -243,8 +828,12 @@ def _supplement_index_with_system(index: dict[PackageRecord, PackageRecord]) ->
 
     :param index: The package index to supplement.
     """
+    if isinstance(index, Index):
+        return
     for package in context.plugin_manager.get_virtual_packages():
-        rec = _make_virtual_package(f"__{package.name}", package.version, package.build)
+        rec = PackageRecord.make_virtual_package(
+            f"__{package.name}", package.version, package.build
+        )
```
jezdez, line 836:
```suggestion
        rec = PackageRecord.virtual_package(
            f"__{package.name}", package.version, package.build
        )
```

conda/core/index.py
```diff
@@ -197,6 +771,11 @@ def _supplement_index_with_cache(index: dict[Any, Any]) -> None:
             index[pcrec] = pcrec
 
 
+@deprecated(
+    "24.9",
+    "25.3",
+    addendum="Use `conda.core.models.records.PackageRecord.make_virtual_package` instead.",
```
jezdez, line 777:
```suggestion
    addendum="Use `conda.core.models.records.PackageRecord.virtual_package` instead.",
```

news/13880-lazy-index
```diff
@@ -0,0 +1,28 @@
+### Enhancements
+
+* Add `conda.core.index.Index` as a faster drop-in replacement of realized dictionary index. (#13880)
+  Note: This removes the implicit initialization of the logging from `get_index`. If you relied on that, you should probably call `conda.gateways.logging.initialize_logging` explicitly.
```
jezdez, line 4:
```suggestion
* Add `conda.core.index.Index` as a faster drop-in replacement of the realized dictionary index. (#13880)

  Note: This removes the implicit initialization of the logging from `get_index`. If you relied on that, you should probably call `conda.gateways.logging.initialize_logging` explicitly.
```

jezdez (changes requested):
Final pass from me, I promise, this looks very good, my feedback is mostly about API naming

conda/core/subdir_data.py
zklaus: Good catch! Done.

zklaus:
pre-commit.ci autofix

jezdez (dismissed):

travishathaway (dismissed):
First off, thanks for taking the time to create a new API for us to use to manipulate the index. But, unfortunately, I cannot approve this pull request as it currently stands because the test coverage, especially on the `Index` class is lacking. For example, the `features`, `__copy__` and `_supplement_index_dict_with_cache` appear to not be tested at all.

I could be swayed from this opinion, but I will need sufficient reasons why this is not necessary to include at this time. Ideally, with all new code, we should be trying as hard as possible to achieve 100% test coverage.

zklaus:
Thanks for the feedback, @travishathaway! Looking into adding tests, I noticed that in https://github.com/conda/conda/pull/6924, `conda/models/index_record.py` was renamed to `conda/models/records.py`, but so far the corresponding test module retains the name of `tests/models/test_index_record.py`. Should we rename the test file to make it easier to find the right tests?

jezdez:
> First off, thanks for taking the time to create a new API for us to use to manipulate the index. But, unfortunately, I cannot approve this pull request as it currently stands because the test coverage, especially on the `Index` class is lacking. For example, the `features`, `__copy__` and `_supplement_index_dict_with_cache` appear to not be tested at all.

To be clear, see the comment from [CodSpeed about the speed improvements](https://github.com/conda/conda/pull/13880#issuecomment-2088392932), the reason to centralize index access is primarily to improve the speed of a pretty central part of conda.

Test coverage shouldn't reduce, ideally, but this isn't just a random new API.

travishathaway:
> Thanks for the feedback, @travishathaway! Looking into adding tests, I noticed that in #6924, `conda/models/index_record.py` was renamed to `conda/models/records.py`, but so far the corresponding test module retains the name of `tests/models/test_index_record.py`. Should we rename the test file to make it easier to find the right tests?

Yes, I think that's a great idea.

conda/core/index.py
```diff
@@ -51,7 +58,566 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        return f"<ReducedIndex(spec={self.specs}, channels=[{channels}])>"
+
+    def _derive_reduced_index(self) -> None:
+        records = IndexedSet()
+        collected_names = set()
+        collected_track_features = set()
+        pending_names = set()
+        pending_track_features = set()
+
+        def push_spec(spec: MatchSpec) -> None:
```
kenodegard, line 533: `push_spec` is called several times below, often within a loop, thoughts on updating it to support `*specs`?

```python
def push_specs(*specs: MatchSpec | str) -> None:
    for spec in map(MatchSpec, specs):
        ...
```

that way instead of writing:

```python
push_spec(MatchSpec(record.name))
for _spec in combined_depends:
    push_spec(_spec)
if record.track_features:
    for ftr_name in record.track_features:
        push_spec(MatchSpec(track_features=ftr_name))
```

we can write:

```python
push_specs(record.name, *combined_deps, *record.track_features)
```

looks like `push_record` could benefit from similar changes

tests/core/test_index.py
```diff
@@ -164,3 +257,182 @@ def test_basic_get_reduced_index():
...
+    main_prec = PackageRecord(**DEFAULTS_SAMPLE_PACKAGES[subdir])
+    conda_forge_prec = PackageRecord(**CONDAFORGE_SAMPLE_PACKAGES[subdir])
+
+    assert main_prec == index[main_prec]
+    assert conda_forge_prec == index[conda_forge_prec]
+
+
+class TestIndex:
```
kenodegard, line 274: why are we organizing these tests this way? if this organization style is necessary a separate test file would be more appropriate?

conda/models/records.py
```diff
@@ -244,38 +244,105 @@ class PathsData(Entity):
...
+    and reference of :class:`PackageRecord` s in mappings, _different_ objects appear identical.
+    The fields taken into account are marked in the following list of attributes.
+    The subclasses do not add further attributes to the :attr:`_pkey`.
+    """
+
     name = StringField()
+    """str: The name of the package.
+
+    Part of the :attr:`_pkey`.
+    """
```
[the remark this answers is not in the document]
zklaus: Hm. I tried that now, but it seems to make the documentation disappear, iow, autoapi doesn't parse it correctly?
See [here](https://continuumio-conda-docs--13880.com.readthedocs.build/projects/conda/en/13880/dev-guide/api/conda/models/records/index.html#conda.models.records.PackageRecord.name).

How should we handle this?

conda/core/index.py
```diff
@@ -139,16 +696,33 @@ def dist_str_in_index(index: dict[Any, Any], dist_str: str) -> bool:
...
-    assert prefix
-    for prefix_record in PrefixData(prefix).iter_records():
+    if isinstance(prefix, PrefixData):
+        prefix_data = prefix
+        prefix_path = prefix.prefix_path
+    else:
+        prefix_path = Path(prefix)
+        prefix_data = PrefixData(prefix)
+    if isinstance(index, Index):
+        if not prefix_path.samefile(Path(index.prefix_path)):
```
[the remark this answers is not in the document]
zklaus: Yes. Do you consider that a bad thing? I had changed the code at some point, but ultimately decided that that's not a bad message for a non-existent prefix directory.

conda/core/index.py
```diff
@@ -51,7 +58,566 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+                self._data.update(self.features)
+        if system:
+            self._system_packages = {
+                (
+                    rec := PackageRecord.virtual_package(
+                        f"__{package.name}", package.version, package.build
+                    )
+                ): rec
+                for package in context.plugin_manager.get_virtual_packages()
+            }
```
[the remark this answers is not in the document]
zklaus: I'm open to this, but who else is using `get_virtual_packages`? Feels like at least a deprecation cycle may be needed, no?

conda/core/index.py
```diff
@@ -51,7 +58,566 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+            if has_data:
+                self._supplement_index_dict_with_prefix()
+        if cache:
+            self._cache_entries = PackageCacheData.get_all_extracted_entries()
+            if has_data:
+                self._supplement_index_dict_with_cache()
+        if features:
+            self._features = {
+                (rec := PackageRecord.feature(track_feature)): rec
+                for track_feature in context.track_features
```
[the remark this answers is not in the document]
zklaus: Thanks for finding this smell. I opted for removing `self.track_features` instead. This was in place originally to allow for adding custom tracking features just to the index, as supported by the current `_supplement_index_with_features` method. However, this was not used anywhere in the code base, Jaime informed me that track features are seen as old fashioned and together we decided not to support the addition of custom tracking features for indexes. Hence, I think it's best to just remove it.

tests/core/test_index.py
zklaus: I discussed this in a call with Travis. We quickly agreed that it would be nice to group the tests for the index class together in some form. Having it in a class certainly was convenient for me to run those tests as a group. It also shows up nicely grouped in the log and allows for the fixtures here to be scoped more narrowly.

You are of course right that that could be achieved by putting them in a separate file, but that would break the source file---test file correspondence.

I don't feel overly strong about this.

Are there any downsides to using a class here?

conda/models/records.py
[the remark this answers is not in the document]
zklaus: Btw, this problem also exists for the examples you linked, see https://docs.conda.io/projects/conda/en/stable/dev-guide/api/conda/index.html#conda.CONDA_PACKAGE_ROOT.

conda/core/index.py
zklaus: Nice and done, thanks!

conda/core/index.py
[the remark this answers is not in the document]
jaimergp: I'd suggest adding a new method for the plugin manager: `.get_virtual_package_records()`, or a keyword argument for `get_virtual_packages(as_records=True)`. And then deprecate accordingly if needed.

conda/core/index.py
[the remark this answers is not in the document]
kenodegard: We've been bit by this a few times already so just being cautious, if the intention is to bubble up the `FileNotFoundError` it is probably better to explicitly catch it here and raise a more informative error

conda/models/records.py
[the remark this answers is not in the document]
kenodegard: grrr that means something in the sphinx setup broke, perhaps when we did the big docs refactor?

cc @travishathaway

conda/core/index.py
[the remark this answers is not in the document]
kenodegard: I like @jaimergp's first suggestion

tests/core/test_index.py
kenodegard: no downsides really, I've just been focused on squashing all unittest classes into pytest functions for so long that this just feels off

if we feel the need to organize the tests into a grouping like this does it suggest the source code should be organized differently too?

kenodegard: what if we move the new `Index` and `ReducedIndex` classes into a new `conda.models.index`?

conda/models/records.py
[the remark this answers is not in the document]
travishathaway: I vaguely remember having a hard time using this particular syntax for adding module level variable comments. I will look into it.

conda/core/index.py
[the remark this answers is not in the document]
zklaus: Fair enough. I'll change it to only check for equality.

conda/core/index.py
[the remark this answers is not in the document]
zklaus: Done

conda/core/index.py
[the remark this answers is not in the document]
zklaus: Done.

tests/core/test_index.py
zklaus: If you don't feel very very strongly about it, I'd rather leave it as is and get this in. I think the test structure now reflects the code structure well and makes it easy to run tests together that belong together.

I get the feeling about the old unittest classes and the associated trauma :smile:. Travis also brought that up. But on balance I think it's a valid use of ordinary classes in the pytest framework.

conda/models/records.py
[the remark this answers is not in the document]
travishathaway: @zklaus,

Sounds like a good idea.

tests/core/test_index.py
jezdez: I concur with Ken, let's do this correctly and move it into an own module, it's a relatively simple refactoring which might be harder to do later on.

kenodegard: As I stated above:

> if we feel the need to organize the tests into a grouping like this does it suggest the source code should be organized differently too?

that's why I'm suggesting that the `Index` and `ReducedIndex` classes are more appropriately defined in `conda.models.index`

I have no problem with using classes like this in the test suite, I see it rather as a suggestion that something else is incorrectly organized

zklaus: I am happy to move `Index` and `ReducedIndex` into a new `conda.models.index` module.

However, from #14248, it seems a bit unclear whether that is actually the right thing to do.
In `conda.models.__init__`, it is stated that
> Models are data transfer objects or "light-weight" domain objects with no appreciable logic
other than their own validation. Models are used to pass data between layers of the stack. In
many ways they are similar to ORM objects.  Unlike ORM objects, they are NOT themselves allowed
to load data from a remote resource.  Thought of another way, they cannot import from
``conda.gateways``, but rather ``conda.gateways`` imports from ``conda.models`` as appropriate
to create model objects from remote resources.
>
> Conda modules importable from ``conda.models`` are
>
> - ``conda._vendor``
> - ``conda.common``
> - ``conda.models``

On the one hand, I can see and argument for the index being mostly an aggregate data transfer object, collecting information from the four different sources of package metadata.

On the other hand, they do load data from remote sources via `SubdirData` and they need access to a number of `conda.core` modules, namely `package_cache_data`, `prefix_data`, and `subdir_data`, as well as the `Context` from `conda.base.context` and exceptions from `conda.exceptions`.

These are not allowed in `conda.models` according to the module docstring.

Should we go ahead and move them anyway? How do we address those discrepancies?

kenodegard: We decided to not move `Index`/`ReducedIndex` over to `conda.models` at this time and will rather investigate further in #14248

conda/core/index.py
```diff
@@ -51,7 +59,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+        if self.prefix_data:
+            self._supplement_index_dict_with_prefix()
+        if self.use_cache:
+            self._supplement_index_dict_with_cache()
+        self._data.update(self.features)
+        if self.use_system:
+            self._data.update(self.system_packages)
+
+    def _retrieve_from_channels(self, key: PackageRecord) -> PackageRecord | None:
+        for subdir_datas in reversed(self.channels.values()):
```
kenodegard, line 378: Doesn't this incorrectly prioritize packages? If the user defines:

```yaml
channels:
  - conda-forge
  - default
use_local: True
```

wont `default` take precedence over `local`?

conda/core/index.py
```diff
@@ -51,7 +59,565 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+          prefix: associate prefix with this index and add its packages.
+          repodata_fn: filename of the repodata, default taken from config, almost always "repodata.json".
+          use_system: if ``True``, add system packages, that is virtual packages defined by plugins, usually used
+            to make intrinsic information about the system, such as cpu architecture or operating system, available
+            to the solver.
+        """
+        if use_local:
+            channels = ["local"] + list(channels)
+        if prepend:
+            channels += context.channels
```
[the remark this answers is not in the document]
zklaus: I think you are right. I have reworked this part using `IndexedSet` in the spirit of `calculate_channel_urls` and `all_channel_urls`. Let me know what you think.

conda/core/index.py
zklaus: I have only translated the old logic into the new framework, that is, I have looked at [`fetch_index`](https://github.com/conda/conda/blob/aee55cc8ca6b588e3ffbf11ce1df103b6c4c145b/conda/core/index.py#L104) and concluded that the for loop goes through this same list, adding records one channel at a time. That means that later entries in the channel list will overwrite earlier ones if they collide.

Now going doing this lazily, returning the matching entry that would be returned in the old approach means to go through the list backwards and find the first one that way.

Does that make sense?

kenodegard: I see what you mean. I still don't follow how this translates to conda correctly prioritizing channels but since this replicates the old behavior I'm happy to proceed here.

kenodegard (approved):

conda/plugins/manager.py
```diff
@@ -341,9 +342,20 @@ def get_subcommands(self) -> dict[str, CondaSubcommand]:
...
+        "25.3",
+        addendum="Use `conda.plugins.manager.get_virtual_package_records` instead.",
+    )
     def get_virtual_packages(self) -> tuple[CondaVirtualPackage, ...]:
         return tuple(self.get_hook_results("virtual_packages"))
 
+    def get_virtual_package_records(self) -> tuple[PackageRecord, ...]:
+        return tuple(
+            cvp.to_virtual_package()
+            for cvp in self.get_hook_results("virtual_packages")
```
jezdez, line 356:
```suggestion
            hook.to_virtual_package()
            for hook in self.get_hook_results("virtual_packages")
```

kenodegard (approved):

dholth (approved):
Nice.
A few code coverage warnings look like they would be easy to cover.

PR #12771, opened 2023-06-08
Treat deprecation warnings as errors

kenodegard:
### Description

We should treat deprecation warnings as errors for greater awareness and visibility.

Resolves #12372

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #13886, opened 2024-05-02
Refactor `main_config` into a modular API

kenodegard:
### Description

`main_config` previously defined a god-function (and still does). This made it difficult to programmatically interact with the `conda config` functionality.

Here we extract individual components from the god-function (e.g., `get_key`, `set_key`, `remove_item`, `remove_key`, `read_rc`, `write_rc`, `set_keys`). These can then be easily called from other parts of conda as well as from the god-function.

Cherry picked from https://github.com/conda/conda/pull/12771

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -11,6 +11,7 @@ conda.cli.main_remove for the entry points into this module.
 import os
 from logging import getLogger
 from os.path import abspath, basename, exists, isdir, isfile, join
+from pathlib import Path
 
 from .. import CondaError
 from ..auxlib.ish import dals
@@ -47,7 +48,7 @@ from ..models.match_spec import MatchSpec
 from ..plan import revert_actions
 from . import common
 from .common import check_non_admin
-from .python_api import Commands, run_command
+from .main_config import set_keys
 
 log = getLogger(__name__)
 stderrlog = getLogger("conda.stderr")
@@ -482,13 +483,9 @@ def handle_txn(unlink_link_transaction, prefix, args, newenv, remove_op=False):
     if newenv:
         touch_nonadmin(prefix)
         if context.subdir != context._native_subdir():
-            run_command(
-                Commands.CONFIG,
-                "--file",
-                os.path.join(prefix, ".condarc"),
-                "--set",
-                "subdir",
-                context.subdir,
+            set_keys(
+                ("subdir", context.subdir),
+                path=Path(prefix, ".condarc"),
             )
         print_activate(args.name or prefix)
 
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -353,9 +353,13 @@ def print_config_item(key, value):
                 stdout_write(" ".join(("--add", key, repr(item))))
 
 
-def get_key(
-    key: str, config: dict, *, json: dict[str, Any] = {}, warnings: list[str] = []
-):
+def _get_key(
+    key: str,
+    config: dict,
+    *,
+    json: dict[str, Any] = {},
+    warnings: list[str] = [],
+) -> None:
     from ..base.context import context
 
     key_parts = key.split(".")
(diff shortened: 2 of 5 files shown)


beeankha (approved):

conda/cli/install.py
```diff
@@ -482,13 +483,9 @@ def handle_txn(unlink_link_transaction, prefix, args, newenv, remove_op=False):
...
-            run_command(
-                Commands.CONFIG,
-                "--file",
-                os.path.join(prefix, ".condarc"),
-                "--set",
-                "subdir",
-                context.subdir,
+            set_keys(
+                ("subdir", context.subdir),
+                path=Path(prefix, ".condarc"),
```
kenodegard, line 488: This is the desired change that prompted this work

jaimergp:
I only took a quick look now. Will get deeper later. But, for now, a quick question: are we expecting to use these functions in layers other than the application? Thinking of whether they should be defined in the library layer instead of in the CLI module.

kenodegard:
Yes, in the future this should be moved out of `conda.cli` 

I didn't spend any more time thinking about this since this was originally just a concise change in https://github.com/conda/conda/pull/12771

conda/cli/main_config.py
```diff
@@ -346,9 +353,121 @@ def print_config_item(key, value):
...
+) -> None:
+    from ..base.context import context
+
+    key_parts = key.split(".")
+
+    if key_parts[0] not in context.list_parameters():
+        if context.json:
+            warnings.append(f"Unknown key: {key_parts[0]!r}")
+        else:
+            print(f"Unknown key: {key_parts[0]!r}", file=sys.stderr)
```
travishathaway, line 371: I'm curious what the reason was for switching from `stderr_write` here?

conda/cli/main_config.py
```diff
@@ -346,9 +353,121 @@ def print_config_item(key, value):
                 stdout_write(" ".join(("--add", key, repr(item))))
 
 
-def execute_config(args, parser):
+def _get_key(
```
travishathaway, line 356: I think `_get_key` is a little confusing for a function name. Usually `get*` functions/methods return something. This just prints some output.

Would `_show_key` be a better name for this?

travishathaway: Another thing that might this function easier to reason about: it's really just trying to figure out the correct value to use for `sub_config`. Couldn't we just return this as a value?

I just find this function confusing because the ultimate outcome is either mutating the  `json` dictionary or printing something. It would be nice to rip that last little piece of logic out and further restrict what this function is responsible for.

travishathaway (approved):
I think the pull request is good to go. The comments I left were more just food for thought but not required to implement.

Thanks for taking a stab at refactoring a part of conda that has been neglected for a while 😄.

conda/cli/main_config.py
kenodegard: > Usually `get*` functions/methods return something.

No disagreement there, the helper functions were named based on what CLI arguments trigged the code path/where they are called within `execute`. I do not necessarily agree with their implementations.

The goal of this PR was only to remove usage of deprecated functionality (https://github.com/conda/conda/pull/13886#discussion_r1589739364). Additional refactoring for this is desired, necessary, and expected after the 24.5.x release.

conda/cli/main_config.py
kenodegard: `stderr_write` is `getLogger("conda.stderr").info`

As mentioned in past logging discussions it is not clear when and why we should be using the logging infrastructure to print to the terminal instead of a `print` statement. This is all something I anticipate the output backend + logging refactors will resolve for us. Until that time I'm continuing to remove these in favor of simple `print` statements whenever possible.

conda/cli/main_config.py
```diff
@@ -346,9 +353,121 @@ def print_config_item(key, value):
...
+    key: str,
+    config: dict,
+    *,
+    json: dict[str, Any] = {},
+    warnings: list[str] = [],
+) -> None:
+    from ..base.context import context
```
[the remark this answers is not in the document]
kenodegard: I expect future refactors of the config logic to evolve into a more complete config manager. Until then the goal was to leave this functional and as minimal of a change possible.

jezdez (approved):

Issue #13908, opened 2024-05-09
Add updated cheatsheet PDF to cheatsheet page


kathatherine:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Until the cheatsheet automation is complete, we can add the manually-created cheatsheet PDF to the cheatsheet page as an available download. https://docs.conda.io/projects/conda/en/latest/user-guide/cheatsheet.html

PR #13909, opened 2024-05-10
Added cheatsheet PDF download to cheatsheet page

kathatherine:
### Description

Closes https://github.com/conda/conda/issues/13908.

Added cheatsheet PDF download to the new cheatsheet text page. I didn't add an extra folder or symlink back in just yet. I'm not sure if we are going to want to have a new PDF for each version or not.

I also added back in the tips from the PDF, since it'll be more directly comparable. These may or may not be helpful, but can probably be fit into the auto PDF once we figure all of that out.

### Checklist - did you ...


docs/source/user-guide/cheatsheet.rst
```diff
@@ -2,11 +2,17 @@
 Cheatsheet
 ==========
 
-Refer to the conda cheatsheet for basic commands for creating and managing environments, installing packages, and importing and exporting environments.
+The conda cheatsheet contains the most important information about using conda, such as basic commands for creating and managing environments, installing packages, and importing and exporting environments.
+
+See the :download:`conda cheatsheet <cheatsheets/conda-cheatsheet.pdf>` PDF (3 MB) for a downloadable, single-page version.
```
jezdez, line 7:
```suggestion
See the :download:`conda cheatsheet <conda-cheatsheet.pdf>` PDF (3 MB) for a downloadable, single-page version.
```

docs/source/user-guide/cheatsheets/conda-cheatsheet.pdf
jezdez, line 1: There is also a symlink under [`docs/source/user-guide/conda-cheatsheet.pdf`](https://github.com/kathatherine/conda/blob/13908-add-updated-cheatsheet-pdf-to-cheatsheet-page/docs/source/user-guide/conda-cheatsheet.pdf) (note the lack of `cheatsheets`) that points to `docs/source/user-guide/cheatsheets/conda-4.14.pdf` at the moment.

I suggest to rename this file you're adding to `docs/source/user-guide/cheatsheets/conda-24.4.0.pdf` and updating the symlink to point to that. That way we let the old URLs working

kathatherine: Good point, Jannis. I'll do that.

docs/source/user-guide/conda-cheatsheet.pdf
```diff
@@ -1 +1 @@
-cheatsheets/conda-4.14.pdf
\ No newline at end of file
+cheatsheets/conda-cheatsheet-24.4.0.pdf
```
jezdez, line 1:
```suggestion
cheatsheets/conda-24.4.0.pdf
```

docs/source/user-guide/cheatsheets/conda-cheatsheet-24.4.0.pdf
jezdez, line 1: Sorry for being a stickler, could we name the file `conda-24.4.0.pdf` like the other files in the folder?

jezdez (approved):
Naming aside, this looks great! Thank you, @kathatherine!

beeankha (approved):
Other than Jannis's comment [here](https://github.com/conda/conda/pull/13909#discussion_r1597250464) (which I agree with), this looks good to go!

docs/source/user-guide/cheatsheets/conda-cheatsheet-24.4.0.pdf
kathatherine: Not a problem! I missed that the first time because I wasn't reading carefully enough. I appreciate you being a stickler.

kathatherine:
Also, one more thing, @jezdez, I deleted the old cheatsheet PDFs in the cheatsheet PR before this one. Your comment above makes me think we actually want to keep those. I can add them back in, if so.

beeankha:
> Also, one more thing, @jezdez, I deleted the old cheatsheet PDFs in the cheatsheet PR before this one. Your comment above makes me think we actually want to keep those. I can add them back in, if so.

I know you were asking @jezdez but I want to add my opinion that I don't think that any of the old cheat sheets should be on that page, since a lot of them have outdated info and folks can go to an older version of the docs in order to access the previous cheat sheets (but I defer to what @jezdez thinks is best!)

jezdez:
> Also, one more thing, @jezdez, I deleted the old cheatsheet PDFs in the cheatsheet PR before this one. Your comment above makes me think we actually want to keep those. I can add them back in, if so.

Oh I missed that. My worry is that someone was linking to the old files and it's now 404'ing. Maybe we could resolve that with a redirect instead?

jezdez:
> > Also, one more thing, @jezdez, I deleted the old cheatsheet PDFs in the cheatsheet PR before this one. Your comment above makes me think we actually want to keep those. I can add them back in, if so.
> 
> I know you were asking @jezdez but I want to add my opinion that I don't think that any of the old cheat sheets should be on that page, since a lot of them have outdated info and folks can go to an older version of the docs in order to access the previous cheat sheets (but I defer to what @jezdez thinks is best!).
> 
> PS - I might be misunderstanding; if you're talking about reverting the deletion of an old cheat sheet from an older version of the docs, then yes please add it back!

Gotcha, I missed the deletion, my only worry is breaking links to the old PDFs via the stable or latest docs builds that would end being a 404 without resolution.

kathatherine:
Gotcha. Apologies for the misunderstanding! I downloaded them from stable and added them back in.

jezdez (approved):

PR #13928, opened 2024-05-21
Implement conda env installers plugin hook

### LINKED ISSUE #16: Implement conda env installers plugin hook
conda env installer hooks are required bring `conda env create` and `conda create/install/update` closer to parity. 

This POC PR has a lot of thoughts and ideas in it https://github.com/conda/conda/pull/13928. However, the change set is very large and there are some details to iron out before conda can use the installer plugins.

This branch: https://github.com/conda/conda/compare/main...soapy1:installer-hooks?expand=1
just introduces the hooks.

This change proposes that an installer must support 2 operations:
* install -> this takes the list of packages to install (as specified by the user) and the prefix to install them into. It will install the (solved) environment packages into the prefix
* dry_run -> this takes the list of packages to install (as specified by the user) and the prefix to install into. It will return the list of packages to install, but not install them.

For example, to define a new installer plugin:
```
import sys
from subprocess import run
from conda import plugins
python = sys.executable

def install(prefix, specs, args):
    run([python, "-m", "pip", "install", "--prefix", prefix, *specs])

def dry_run(prefix, specs, args):
    run([python, "-m", "pip", "install", "--prefix", prefix, *specs, "--dry-run"])

@plugins.hookimpl
def conda_installers():
    yield plugins.CondaInstaller(
        name="pip",
        types=["pip"],
        install=install,
        dry_run=dry_run,
    )
```

Here's the current ongoing PR: https://github.com/conda/conda/pull/14794

### LINKED ISSUE #13833: `conda_env` installers / providers plugin hooks
### Checklist

### What is the idea?

```yaml
name: bla
channels:
  - conda-forge
dependencies:
  - python
  - pip
  - pip:
    - requests
```

### Why is this needed?

### What should happen?

### Additional Context

jaimergp:
### Description

Part of https://github.com/conda/conda/issues/13833

### Checklist - did you ...


conda/cli/main_env_create.py
```diff
@@ -170,19 +171,19 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-                except InvalidInstaller:
+                except InvalidInstaller as exc:
                     raise CondaError(
                         dals(
                             f"""
                             Unable to install package for {installer_type}.
 
                             Please double check and ensure your dependencies file has
                             the correct spelling. You might also try installing the
-                            conda-env-{installer_type} package to see if provides
```
jaimergp, line 181: Did you know you were supposed to install `conda-env-node` or something like that? o.o

jezdez: Err, TIL!

jezdez: I guess this _was_ planned in the past, but never went anywhere? Seeing https://pypi.org/project/conda-env/ it feels like something went wrong in the past.

conda/env/installers/base.py
```diff
@@ -2,18 +2,20 @@
...
-        Gets the installer for the given environment.
+    Gets the installer for the given environment.
 
     Raises: InvalidInstaller if unable to load installer
     """
-    try:
-        return importlib.import_module(f"conda.env.installers.{name}")
-    except ImportError:
-        raise InvalidInstaller(name)
+    from conda.base.context import context
```
jezdez, line 19: Do we need to inline this?

conda/env/installers/pip.py
```diff
@@ -68,6 +68,10 @@ def _pip_install_via_requirements(prefix, specs, args, *_, **kwargs):
     return get_pip_installed_packages(stdout)
 
 
+def dry_run(*args, **kwargs):
+    return None
```
jezdez, line 72: Huh?

conda/plugins/types.py
```diff
@@ -16,13 +16,35 @@
...
+        def __call__(
+            self,
+            prefix: str,
+            specs: Iterable[str],
+            args: Namespace,
+            env: Environment,
+            *_,
+            **kwargs,
+        ) -> Iterable[str]: ...
+
```
jezdez, line 37:
```suggestion


```

conda/plugins/types.py
```diff
@@ -16,13 +16,35 @@
...
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace
-    from typing import Callable
+    from typing import Callable, Iterable, Protocol
 
     from ..common.configuration import Parameter
     from ..core.solve import Solver
+    from ..env.env import Environment
     from ..models.match_spec import MatchSpec
     from ..models.records import PackageRecord
```
jezdez, line 26:
```suggestion


```

conda/plugins/types.py
```diff
@@ -16,13 +16,35 @@
...
 
+    class CondaEnvInstallerInstallProtocol(Protocol):
```
jezdez, line 27:
```suggestion
    class EnvInstallerInstallProtocol(Protocol):
```

conda/plugins/types.py
```diff
@@ -16,13 +16,35 @@
...
+            self,
+            prefix: str,
+            specs: Iterable[str],
+            args: Namespace,
+            env: Environment,
+            *_,
+            **kwargs,
+        ) -> Iterable[str]: ...
+
+    class CondaEnvInstallerDryRunProtocol(Protocol):
```
jezdez, line 38:
```suggestion
    class EnvInstallerDryRunProtocol(Protocol):
```

conda/plugins/types.py
```diff
@@ -210,3 +232,23 @@ class CondaSetting:
...
+    :param name: name of the installer (e.g., ``pip``)
+    :param types: accepted sections from the environment.yml (e.g. conda, pip).
+    :param install: Callable which contains the code to be run for the install operation.
+    :param dry_run: Callable which contains the code to be run for the dry-run installs.
+    """
+
+    name: str
+    types: Iterable[str]
+    install: CondaEnvInstallerInstallProtocol
+    dry_run: CondaEnvInstallerDryRunProtocol
```
jezdez, line 254:
```suggestion
    install: EnvInstallerInstallProtocol
    dry_run: EnvInstallerDryRunProtocol
```

jezdez (commented):
Looking good in general, but needs docs, news and tests

conda/cli/main_env_create.py
dholth: In the future

conda-env-conda

conda/env/installers/base.py
dholth: When it says "for the given environment" does it mean "for a particular directory in ~/miniconda3/envs/"?

Could it say "ecosystem" or "package type"?

conda/plugins/types.py
dholth: Do we use the word Protocol instead of Abstract now

jaimergp:
@soapy1 is kindly taking over this one 🚀

conda/base/context.py
```diff
@@ -522,6 +523,13 @@ def __init__(self, search_path=None, argparse_args=None, **kwargs):
         self._set_env_vars(APP_NAME)
         self._set_argparse_args(argparse_args)
 
+    def add_environment_file_config_source(self, file, data):
```
soapy1, line 526: In addition to configuration coming from argparse args, environment vars and `.condarc` config files, `environment.yaml` files provide some amount of configuration. For now, `environment.yaml` is only able to provide channel configuration. But in the future, I expect that users might be able to specify more config for their environment. 
This aims to provide a way to inject configuration from the environment file into the running context.

conda/env/env.py
```diff
@@ -264,6 +264,10 @@ def save(self):
         with open(self.filename, "wb") as fp:
             self.to_yaml(stream=fp)
 
+    def get_configuration_settings(self):
```
soapy1, line 267: This represents all the configuration parameters that a user is able to set in an `environment.yaml`

conda/env/installers/base.py
```diff
@@ -1,19 +1,14 @@
...
-import importlib
+Dynamic installer loading.
+"""
 
-from ...exceptions import InvalidInstaller
+from conda.deprecations import deprecated
+from conda.installers.base import get_installer  # noqa
+from conda.exceptions import InvalidInstaller  # noqa
 
+deprecated.module("24.9", "25.3", addendum="Use `conda.installers.base` instead.")
```
soapy1, line 12: I think it makes more sense for installer plugins to not be in the `conda.env` module since they don't particularly have aaything to do with conda environments. For example, you can imagine extending the available list of installers to include uv. So, I think it makes sense to deprecate this module and pull it under `conda.installers`.

conda/installers/conda.py
```diff
@@ -0,0 +1,57 @@
...
+
+def dry_run(specs: Iterable[str], context: Context, *args, **kwargs) -> Iterable[str]:
+    """Do a dry run of the environment solve"""
+    solver = _solve(tempfile.mkdtemp(), specs, context)
+    pkgs = solver.solve_final_state()
+    return [str(p) for p in pkgs]
+
+
+def install(prefix: str, specs: Iterable[str], context: Context, *args, **kwargs) -> Iterable[str]:
+    """Install packages into an environment"""
```
soapy1, line 41: This still needs work in order to support all the install options required for `conda.cli.install.install`

conda/installers/pip.py
```diff
@@ -0,0 +1,93 @@
...
+            else:
+                log.warning(
+                    f"CONDA_TEST_SAVE_TEMPS :: retaining pip requirements.txt {requirements.name}"
+                )
+    return get_pip_installed_packages(stdout)
+
+
+def _get_pip_workdir(context: Context) -> str:
+    pip_workdir = None
+    if hasattr(context._argparse_args, "file"):
```
soapy1, line 66: I don't like this relying on `conda._argparse_args`. It would be better if an installer was completely decoupled from arguments coming in from the command line. 

Unfortunately, in order to support current patterns for installing with pip, conda needs to have a notion of where the environment file is. Still thinking about how to make that information available to the pip installer.

conda/plugins/types.py
```diff
@@ -21,13 +21,35 @@
...
-    from typing import Any, Callable
+    from typing import Any, Callable, Iterable, Protocol
 
+    from ..base.context import Context
     from ..common.configuration import Parameter
     from ..core.solve import Solver
     from ..models.match_spec import MatchSpec
 
 
+    class EnvInstallerInstallProtocol(Protocol):
```
soapy1, line 32: This is quite a bit different from the v1 of this PR. I think the install plugins should be decoupled from argparse args and any notion of environments. 

Instead an installer should deal in 
* what packages need to be installed
* where they need to be installed
* settings/context for solving/installing

And should be returning the list of packages that it has installed.

jezdez (commented):
A few design comments and general discussion about the way we've established the plugins system in the past few years. As a general code feedback, passing around context objects is sadly not something we'll realistically get to without serious widespread refactoring, so relying on it as a global is the better strategy.

conda/cli/main_env_create.py
```diff
@@ -125,6 +125,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         directory=os.getcwd(),
     )
     env = spec.environment
+    context.add_environment_file_config_source(args.file, env.get_configuration_settings())
```
jaimergp, line 128: I like this solution to the configuration overrides! However we need to ensure it lands at the right time with CLI and env var sources. For example, in `conda env create -f env.yml -n mouse`, the env name is `mouse` not whatever `env.yml` carries. So CLI takes precedence. Same with `conda env create -f env.yml -n mouse -c defaults`; CLI takes precedence. Is this precedence order respected right now?

conda/cli/main_env_create.py
```diff
@@ -155,43 +156,47 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         pkg_specs = env.dependencies.get(installer_type, [])
         pkg_specs.extend(args_packages)
 
-        solved_env = installer.dry_run(pkg_specs, args, env)
-        if args.json:
-            print(json.dumps(solved_env.to_dict(), indent=2))
-        else:
-            print(solved_env.to_yaml(), end="")
+        solved_packages = installer.dry_run(pkg_specs, context)
```
jaimergp, line 159: I'm not sure about this method name. It's more like `solve_records` (inspired by `Solver.solve_for_final_state()`), so we are adding a layer in the middle just when we deal some files, but in `install.py` we call the solver directly 🤔 There's something about that inconsistency that I don't like but I don't have a solution right now.

And yea, as Jannis said, we use the global `context` or pass the necessary bits via parameters.

conda/cli/main_env_update.py
```diff
@@ -143,22 +144,20 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+                    Unable to install package for {installer_type}.
 
-                    Please double check and ensure you dependencies file has
-                    the correct spelling.  You might also try installing the
-                    conda-env-{0} package to see if provides the required
-                    installer.
+                    Please double check and ensure your dependencies file has
+                    the correct spelling. You might also try installing the
+                    corresponding conda plugin for this type (e.g. conda-pypi
+                    for 'pip').
```
jaimergp, line 152: I'm not sure we should hardcode particular plugins in the error message. The plugin nuance is good to have, but I wouldn't mention conda-pypi. If anything, we could link to the [conda/plugins](https://github.com/conda-incubator/plugins) repo (which we haven't really used too much), but that can be done separately.

conda/env/installers/base.py
jaimergp: Even if we were to move things from `conda.env` elsewhere, I wouldn't do it in this PR. It increases the diff and hinder readability, so let's revert for now. 🙏

conda/installers/conda.py
jaimergp, line 1: I like this idea and the simple API. It would be awesome to apply this to `conda.cli.install` and see if it needs some more details. For example, I don't think we are passing all the CLI flags like `--no-deps` or `--update-deps`. These don't matter for `conda env create` but do for `conda install` and friends.

conda/installers/conda.py
```diff
@@ -0,0 +1,57 @@
...
+
+    channels = IndexedSet(Channel(url) for url in _channel_priority_map)
+    subdirs = IndexedSet(basename(url) for url in _channel_priority_map)
+
+    solver_backend = context.plugin_manager.get_cached_solver_backend()
+    solver = solver_backend(prefix, channels, subdirs, specs_to_add=specs)
+    return solver
+
+
+def dry_run(specs: Iterable[str], context: Context, *args, **kwargs) -> Iterable[str]:
```
jaimergp, line 33: I'd call this `solve` not `dry_run`. It can be used as a dry-run only functionality, but that's an application layer detail, not concerned with the library.

conda/installers/conda.py
```diff
@@ -0,0 +1,57 @@
...
+    """Do a dry run of the environment solve"""
+    solver = _solve(tempfile.mkdtemp(), specs, context)
```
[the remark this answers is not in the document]
jaimergp: That's the target prefix. It should be `context.target_prefix` if not provided. I think `conda create --dry-run` handles the tempdir part already, so it would be good to unify it.

conda/installers/conda.py
```diff
@@ -0,0 +1,57 @@
...
+from boltons.setutils import IndexedSet
+
+from ..base.constants import UpdateModifier
+from ..base.context import Context
+from ..common.constants import NULL
+from ..exceptions import UnsatisfiableError
+from ..models.channel import Channel, prioritize_channels
+
+
+def _solve(prefix, specs, context: Context):
```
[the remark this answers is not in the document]
soapy1: This is already existing code that shows up in this diff because of moving the `conda.env.installer` module to `conda.installer`. I think there is definitely room for improvement here though. Will take a peek at what is happening in other parts of the code for ideas.
 
ref: https://github.com/conda/conda/blob/main/conda/env/installers/conda.py#L18

conda/installers/conda.py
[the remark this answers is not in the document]
soapy1: Same as above https://github.com/conda/conda/pull/13928#discussion_r2006266304 

> I think conda create --dry-run handles the tempdir part already, so it would be good to unify it.

Yep! Lets tackle that in a follow up pr that gets these installer plugins working for `conda create`?

conda/installers/conda.py
soapy1: Ya, it would be nice if everything that is using installers is able to use the plugin. So, imho it's worth it to spend some extra time to make sure this API will allow that to be a possibility. 

Very good point about scope creep. I'm going to spin out another branch to test out getting `conda install` to also use the installer plugins. We can turn that into a PR when we are more ready. Let's keep the scope of this PR to introducing the api for installer plugins.

conda/installers/conda.py
soapy1: Hmm, this implements the [conda env plugin `dry_run` interface ](https://github.com/conda/conda/blob/ee1e1c223d7ad80eab516dc0b5651ada37a8fe36/conda/plugins/types.py#L180). 

Are you proposing to rename `dry_run` to `solve` for all installers?

conda/installers/pip.py
```diff
@@ -0,0 +1,93 @@
...
+        requirements.close()
+        # pip command line...
+        # see https://pip.pypa.io/en/stable/cli/pip/#exists-action-option
+        pip_cmd = ["install", "-U", "-r", requirements.name, "--exists-action=b"]
+        if dry_run:
+            pip_cmd += "--dry_run"
+        stdout, stderr = pip_subprocess(pip_cmd, prefix, cwd=pip_workdir)
+    finally:
+        # Win/Appveyor does not like it if we use context manager + delete=True.
+        # So we delete the temporary file in a finally block.
```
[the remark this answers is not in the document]
soapy1: I'm not sure. But this part is same as above https://github.com/conda/conda/pull/13928#discussion_r2006266304

I'd love to not dig into this for the purpose of this PR. But can do if you think that's worthwhile.

conda/installers/conda.py
soapy1: For reference her is another branch that gets the `conda create` command to use this installer plugin https://github.com/conda/conda/compare/main...soapy1:conda:env-installers-hook-conda-create?expand=1. There is still lots of work to get it working all the way. But it does work for simple use cases (like the smoke tests).

conda/plugins/types.py
soapy1: > smells like yet another API layer we're creating inside conda to manage conda

yes, that was what I was going for. Is there a particular reason why this is not a good direction for this type of plugin?

> Just note that plugin implementations are supposed to provide just enough functionality to do what's needed, stay lightweight, and implement a conservative level of provided conda APIs

yep, I agree. That's why I propose installer's not dealing in things like the environment model. Instead they should deal in lists of packages to install and paths to the install location. Like IMO an installer should not need to know that an environment is a concept that exists, and much less an `environment.yaml`. That way these plugins can be reused in the `conda create` commands and we can deduplicate some of this code. 

Is there a different interface for an installer plugin that would better represent what an installer should be receiving as inputs/outputs?

conda/base/context.py
soapy1: Good point, can do!

conda/common/configuration.py
```diff
@@ -408,6 +408,37 @@ def make_raw_parameters_from_file(cls, filepath):
             return cls.make_raw_parameters(filepath, yaml_obj) or EMPTY_MAP
 
 
+class EnvironmentSpecificationRawParameter(RawParameter):
+    """A raw parameter that can be injected"""
```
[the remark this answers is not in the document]
soapy1: can do!

conda/installers/pip.py
```diff
@@ -0,0 +1,93 @@
...
+            try:
+                pip_workdir = op.dirname(op.abspath(file))
+                if not os.access(pip_workdir, os.W_OK):
+                    pip_workdir = None
+            except AttributeError:
+                pip_workdir = None
+    return pip_workdir
+
+
+def dry_run(specs: Iterable[str], context: Context, *args, **kwargs) -> Iterable[str]:
```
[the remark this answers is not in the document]
soapy1: can do!

conda/cli/main_env_create.py
soapy1: linking to this other thread about this approach to adding config to the context https://github.com/conda/conda/pull/13928#discussion_r2005017049 

I haven't tried to see if precedence is preserved, but definitely will need to iron that out!

conda/plugins/types.py
soapy1: @jezdez do you have opinions or thoughts around the form of this interface? Given the size of this PR, I'd love to split this part that defines the installer hooks into a separate PR.

conda/cli/main_env_create.py
```diff
@@ -155,43 +156,47 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        solved_env = Environment(
+            name=env.name, dependencies=solved_packages, channels=context.channels
+        )
```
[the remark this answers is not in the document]
soapy1: @jezdez just following up to see if this seems like an ok change to you?

jaimergp:
Continues at https://github.com/conda/conda/pull/13928

PR #13930, opened 2024-05-21
Draft conda env spec hooks

### LINKED ISSUE #18: Environment specifier hooks
Implement plugin hooks for the different formats accepted by `conda env create` and friends.

# PRs

- [X] ~demo pr: https://github.com/conda/conda/pull/13930~
- [X] https://github.com/conda/conda/pull/14710
- [X] https://github.com/conda/conda/pull/14814
- [X] https://github.com/conda/conda/pull/14823
- [X] https://github.com/conda/conda/pull/14900
- [x] https://github.com/conda/conda/pull/14914
- [x] https://github.com/conda/conda/pull/14916
- [x] https://github.com/conda/conda/pull/14918
- [x] https://github.com/conda/conda/pull/14877
- [x] https://github.com/conda/conda/pull/14925

### LINKED ISSUE #13833: `conda_env` installers / providers plugin hooks

jaimergp:
### Description

Part of #13833. The tricky part here is determining how to find a good API that allows to easily map a _handler_ to a _resource_. 

`conda env` currently has two types of resource:

- A file-like resource, as passed with `-f, --file`. This is either a local file on disk or a remote location under HTTP(s), FTP or S3 protocols. This is the most common way to use `conda env create` in the community, I'd say. This category defines two types of specs, which are found by the resource _extension_:
  - YAML files: your beloved environment.yml
  - TXT files: the plain text requirements.txt (same as `conda create --file`?) 
- A remote definition of an environment. This argument doesn't require any flags. This category only has one spec so far:  
  - Binstar environments (as provided by Anaconda.org). I don't think these are used too often in the community but maybe there's a huge user somewhere in the corporate world, who knows.

> Note that a file-like resource can be remote / downloadable, so the "remote definition" is a bit of a misnomer in a way.

Looking at it from the handler perspective, we have three classes that implement an implicit API:

```python
class BaseEnvSpec:
    msg: str | None = None

    def can_handle(self) -> bool:
        # this is an instancemethod, not a classmethod!
        # so it requires initialization, which means different __init__ signatures...
        # can we change that? I think we should :D
        raise NotImplementedError

    @property
    def environment(self) -> Environment:
        raise NotImplementedError
```

The key being that [`Environment`](https://github.com/conda/conda/blob/49f2614c5d0e9092ba199ea7adb9a0bc94aedb80/conda/env/env.py#L210) object. So as long as this API is fulfilled, then _any_ format could be implemented.

> Looking at https://github.com/conda/conda/pull/12952, I think that should be a file-like resource, not a "remote definition" 🤷

I think these two categories of resource only add friction but it might be too late to disambiguate. Maybe we could deprecate `-f, --file` and do:

```
$ conda env create # implicitly looks for environment.yml in cwd
$ conda env create environment.yml
$ conda env create https://github.com/owner/repo/raw/environment.yml
$ conda env create requirements.txt
$ conda env create owner/name  # only check in binstar if this path doesn't exist locally, as a backwards compatibility layer
$ conda env create binstar://owner/name  # force anaconda.org via fsspec?
```

However that might be a separate issue not tackled here. The point now is how to design the plugin API so we can arbitrarily match a resource to its handler. The [current logic](https://github.com/conda/conda/pull/13930/files#diff-7d861030d2e81b5082cd0b27c9af9b3b18ff05031d0e20184da093fb4a710f8eR408) is:

1. Check for handlers that handle `remote_definition` with explicit protocols (like the fsspec PR)
2. Special case the protocol-less binstar handler
3. At this point, no remote definitions were passed, so we check for file-like resources. We check its extension and its existence / remote protocol availability (it will be downloaded later). 
 - I feel this extension-only check is insufficient and we should maybe have an additional URI syntax with keywords like `?environment_type=yml` for URIs that are not path addressed but maybe content addressed (e.g. with a hash).

So to support that logic, the EnvSpec API can have a `extensions` and `protocols` matcher. That's ok I think.

Then there's some UX weirdness with the `TXT` handler and the lack of names in that environment format. I feel this should be streamlined or deprecated since `conda create` already handles that.

TLDR:

- I don't like the need for the `--file` flag just to give the Binstar handler a better UX.
- The Binstar handler would become its own `conda-env-binstar` plugin (or whatever name).
- I'd like to deprecate `-f, --file` and give the Binstar handler its own protocol so we can use the positional argument as the only way to specify an environment definition, remote or not. 
- This way we can get rid of the file-like / remote definition difference weirdness in the APIs. If we agree we can get there, then I'd design the APIs with that in mind.
- The file-like match-by-extension approach might be insufficient for some plugins. Thinking of a future where might have `environment.v2.yml` and how to handle that.

### Checklist - did you ...


conda/env/specs/binstar.py
```diff
@@ -20,7 +21,7 @@
 ENVIRONMENT_TYPE = "env"
 
 
-class BinstarSpec:
+class BinstarSpec(BaseEnvSpec):
```
jezdez, line 24: I wish we'd just yeet this into an own package

jaimergp:
xref https://conda.discourse.group/t/notice-upcoming-sunset-of-projects-notebooks-and-environments-features-on-anaconda-org/672

> Projects, Notebooks, and Environments features on [Anaconda.org](https://anaconda.org/) will be removed on August 20th, 2024.

This means that we should probably deprecate the `binstar` spec provider and the (protocol-less) positional CLI.

jaimergp:
Continues at https://github.com/conda/conda/pull/14710

jaimergp:
Closed by https://github.com/conda/conda/pull/14710

PR #13936, opened 2024-05-24
`conda list --explicit` should not include auth details by default

jaimergp:
### Description

I noticed that some of my `conda-meta/*.json` files have auth details in them. `conda list --explicit` will happily include the auth details in the URL by default IN PLAIN TEXT. For a file that is meant to be shared, I think that default option is too big of a security risk so in this PR:

- I changed the default behaviour so it does NOT print auth details.
- I add an `--auth` flag to `conda list` to opt-in including auth details (in case this is used internally with additional security measurements in place).

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_list.py
+++ b/conda/cli/main_list.py
@@ -117,9 +117,9 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     )
     p.add_argument(
         "--auth",
-        action="store_true",
-        default=False,
-        dest="auth",
+        action="store_false",
+        default=True,
+        dest="remove_auth",
         help="In explicit mode, leave authentication details in package URLs. "
         "They are removed by default otherwise.",
     )
@@ -244,10 +244,10 @@ def print_packages(
     return exitcode
 
 
-def print_explicit(prefix, add_md5=False, with_auth=False):
+def print_explicit(prefix, add_md5=False, remove_auth=True):
     from ..base.constants import UNKNOWN_CHANNEL
     from ..base.context import context
-    from ..common.url import remove_auth, split_anaconda_token
+    from ..common import url as common_url
     from ..core.prefix_data import PrefixData
 
     if not isdir(prefix):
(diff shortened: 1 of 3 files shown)


conda/cli/main_list.py
```diff
@@ -236,9 +244,10 @@ def print_packages(
     return exitcode
 
 
-def print_explicit(prefix, add_md5=False):
+def print_explicit(prefix, add_md5=False, with_auth=False):
```
jezdez, line 247: I like the imperative name better for this new argument, since it makes clear that the function _does_ process the URL and doesn't add authentication as implied with "with"

```suggestion
def print_explicit(prefix, add_md5=False, remove_auth=True):
```

conda/cli/main_list.py
```diff
@@ -236,9 +244,10 @@ def print_packages(
...
     from ..base.constants import UNKNOWN_CHANNEL
     from ..base.context import context
+    from ..common.url import remove_auth, split_anaconda_token
```
jezdez, line 250:
```suggestion
    from ..common import url as common_url
```

conda/cli/main_list.py
```diff
@@ -252,6 +261,8 @@ def print_explicit(prefix, add_md5=False):
         if not url or url.startswith(UNKNOWN_CHANNEL):
             print("# no URL for: {}".format(prefix_record["fn"]))
             continue
+        if not with_auth:
+            url = remove_auth(split_anaconda_token(url)[0])
```
jezdez, line 265:
```suggestion
        if remove_auth:
            url = common_url.remove_auth(common_url.split_anaconda_token(url)[0])
```

conda/cli/main_list.py
```diff
@@ -286,7 +297,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         return 0
 
     if args.explicit:
-        print_explicit(prefix, args.md5)
+        print_explicit(prefix, args.md5, args.auth)
```
jezdez, line 300:
```suggestion
        print_explicit(prefix, args.md5, args.remove_auth)
```

conda/cli/main_list.py
```diff
@@ -115,6 +115,14 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         dest="pip",
         help="Do not include pip-only installed packages.",
     )
+    p.add_argument(
+        "--auth",
+        action="store_true",
+        default=False,
+        dest="auth",
```
jezdez, line 122: I think that's the correct incantation for the flip of the argument logic suggested below:

```suggestion
        action="store_false",
        default=True,
        dest="remove_auth",
```

jezdez (commented):
The new argument as proposed is misleading as this remove things from the URL, not adds auth as the "with_auth" name implies.

jaimergp:
> The new argument as proposed is misleading as this remove things from the URL, not adds auth as the "with_auth" name implies.

Excellent point. Suggestions applied and passing!

travishathaway (approved):

jezdez (approved):

PR #13937, opened 2024-05-24
Do not include auth details in `conda-meta/*.json` files by default

jaimergp:
### Description

Related to https://github.com/conda/conda/pull/13936.

### Checklist - did you ...


conda/core/prefix_data.py
```diff
@@ -112,7 +113,7 @@ def _get_json_fn(self, prefix_record):
             )
         return fn + ".json"
 
-    def insert(self, prefix_record):
+    def insert(self, prefix_record, dump_auth=False):
```
travishathaway, line 116:
```suggestion
    def insert(self, prefix_record, include_tokens=False):
```

I think this is a little more descriptive.

conda/core/prefix_data.py
```diff
@@ -132,8 +133,14 @@ def insert(self, prefix_record):
                 context,
             )
             rm_rf(prefix_record_json_path)
-
-        write_as_json_to_file(prefix_record_json_path, prefix_record)
+        if dump_auth:
```
travishathaway, line 136:
```suggestion
        if include_tokens:
```

travishathaway (commented):
Left a suggestion for a slightly better variable name.

I was curious though. Are we okay with making this default behavior? Is this the type of thing that gets to bypass the deprecation notices?

jaimergp:
There was no option before and I would say this is a security oversight. I'm not aware of any conversations discussing this in the past, but I didn't run a thorough investigation. 

The risk here is that the anaconda-client tokens, which should / could be protected by the OS permission system under `.continuum/anaconda-client`, are being exported silently to a location that might not benefitting from the same security treatment. I think that constitutes a bug, not a "behavior".

conda/core/prefix_data.py
jezdez: To stay consistent with the argument added in (my review of) #13936:

```suggestion
    def insert(self, prefix_record, remove_auth=True):
```

conda/core/prefix_data.py
```diff
@@ -132,8 +133,14 @@ def insert(self, prefix_record):
...
+            write_as_json_to_file(prefix_record_json_path, prefix_record)
+        else:
+            dumped_prefix_record = prefix_record.dump()
+            dumped_prefix_record["url"] = remove_auth(
+                mask_anaconda_token(prefix_record.url)
+            )
+            write_as_json_to_file(prefix_record_json_path, dumped_prefix_record)
```
jezdez, line 143:
```suggestion
        if remove_auth:
            prefix_record_json = prefix_record.dump()
            prefix_record_json["url"] = remove_auth(
                mask_anaconda_token(prefix_record.url)
            )
        else:
            prefix_record_json = prefix_record
        write_as_json_to_file(prefix_record_json_path, prefix_record_json)
```

jezdez (dismissed):
Approved with comments

jaimergp:
pre-commit.ci autofix

jaimergp:
> Approved with comments

Comments applied, requesting reapproval :D

travishathaway (approved):
I checked with the Anaconda Navigator team on this, and they do not foresee any difficulties with such a change. Therefore, I'm happy to approve it now 😄 🚀

PR #13904, opened 2024-05-08
Add a `keep_case` flag which ensures that env variables are not uppercased

ForgottenProgramme:
### Description
resolves #13713 

At present `conda` converts all environment variables into uppercase. 
This PR adds a `keep_case` context variable, which allows users to store their environment variables as is.

When the `--keep-case` flag is used, environment variables are saved in "preserved case". When the flag is not used variables are saved in upper case (the original conda behavior). 

**Tasks**

- [x] Add a new `keep_case` variable to context
- [x] store environment variables in preserved case when `keep_case` is `True`
- [x] write tests
- [x] write more tests
- [x] Parametrize the tests!

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Issue #13713, opened 2024-03-20
lowercase environment variables not showing


krishnaispresent:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When setting environment variables in his conda environment, by using lowercase letters for the variable key, it appears conda sets the key to uppercase when it sets the variable:

(base) krishna@Krishnas-MBP ~ % conda env config vars set my_test_var="hello" -n envvartest
(base) krishna@Krishnas-MBP ~ % conda activate envvartest
(envvartest) krishna@Krishnas-MBP ~ % echo $my_test_var

(envvartest) krishna@Krishnas-MBP ~ % echo $MY_TEST_VAR
hello
(envvartest) krishna@Krishnas-MBP ~ % conda env config vars list
my_test_var = hello
Snippet from activate.py:
        # split provided environment variables into exports vs unsets
        for name, value in kwargs.items():
            if value is None:
                unset_vars.append(name.upper())
            else:
                export_vars[name.upper()] = value

### Conda Info

```shell
(base) krishna@Krishnas-MBP Desktop % conda info


     active environment : base
    active env location : /Users/krishna/anaconda3
            shell level : 1
       user config file : /Users/krishna/.condarc
 populated config files : /Users/krishna/.condarc
          conda version : 24.1.2
    conda-build version : 3.27.0
         python version : 3.11.5.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=24.1.2=0
                          __osx=14.2=0
                          __unix=0=0
       base environment : /Users/krishna/anaconda3  (writable)
      conda av data dir : /Users/krishna/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.cloud/repo/main/osx-arm64
                          https://repo.anaconda.cloud/repo/main/noarch
                          https://repo.anaconda.cloud/repo/r/osx-arm64
                          https://repo.anaconda.cloud/repo/r/noarch
                          https://repo.anaconda.cloud/repo/msys2/osx-arm64
                          https://repo.anaconda.cloud/repo/msys2/noarch
          package cache : /Users/krishna/anaconda3/pkgs
                          /Users/krishna/.conda/pkgs
       envs directories : /Users/krishna/anaconda3/envs
                          /Users/krishna/.conda/envs
               platform : osx-arm64
[...]

PR #13938, opened 2024-05-24
Save preserved case as well as upper cased env vars

ForgottenProgramme:
### Description
resolves # https://github.com/conda/conda/issues/13713

This PR explores the second possible solution to the problem of `conda` uppercasing users' environment variable names. 

(check https://github.com/conda/conda/pull/13904 to see the first possible solution: introducing a `--keep-case` flag). 

The changes introduced in this PR cause `conda` to save two new environment variables: one in its original preserved case (case as defined by the user) and one in the upper case. 

The idea is to save both preserved case and upper case variables for now, and give a deprecation warning about the upper casing of the env var names. And eventually, deprecate the upper casing behavior. Therefore after this deprecation process, only the case defined by the user will be used for variable names. 

To Do:

- [ ] add deprecation warning

### Checklist - did you ...


jezdez:
Closed in favor of #13943.

PR #13941, opened 2024-05-24
Do not require name or prefix if we are running `create` with dry-run

jaimergp:
### Description

Folks often rely on `create --dry-run` to run solver diagnostics. For this operation, prefixes or env names are not needed, yet they are still required in the CLI. It's common to see invocations such as:

```
$ conda create -n unused --dry-run ...
```

Or the shorthand (with `x` being the env name):

```
$ conda create -dnx ...
```

However it should be perfectly fine to run:

```
$ conda create --dry-run ...
```

This is what we see in this PR.

### Checklist - did you ...


conda/cli/main_create.py
```diff
@@ -100,14 +98,25 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
-    from ..exceptions import CondaValueError
+    from ..exceptions import ArgumentError, CondaValueError
     from ..gateways.disk.delete import rm_rf
     from ..gateways.disk.test import is_conda_environment
     from .common import confirm_yn
     from .install import install
 
+    if not args.name and not args.prefix:
+        if context.dry_run:
+            context._argparse_args.prefix = os.path.join(mktemp(), "dry-run-create-env")
```
travishathaway, line 114:
```suggestion
            args.prefix = os.path.join(mktemp(), "dry-run-create-env")
            context.__init__(argparse_args=args)
```

I think this is slightly better. No need to access a protected property when we can simply re-initialize the object. Performance penalty should be negligible.

Also, Could we make  `"dry-run-create-env"` a constant and define it in `conda.base.constants`. Not necessary, but would help keep the code a little more organized and provide a central place to look up this value in case it is ever needed.

tests/test_create.py
```diff
@@ -2589,3 +2589,15 @@ def test_repodata_v2_base_url(
         platform,
     )
     assert package_is_installed(prefix, "ca-certificates")
+
```
travishathaway, line 2592: Could we please add a new test case to test `conda create python` (i.e. no `--name` or `--prefix` provided), so we can cover the new `raise ArgumentError` exception you added in `main_create.py`?

travishathaway (changes requested):
Just a couple requests for changes. Please let me know if seem unreasonable. Will be happy to review once more.

conda/cli/main_create.py
jaimergp: Are we sure we are not overriding some stuff from `os.environ` with that call? 🤔 I always get nervous when I reinitialize the context outside of the tests 😬 

+1 to the 2nd suggestion.

tests/test_create.py
jaimergp: Done :)

conda/cli/main_create.py
jaimergp: Applied your suggestions. The new tests seem to be passing locally.

travishathaway: No reason to worry about overriding stuff from `os.environ`. I double checked the code. We just need to make sure we initialize it like we do in `conda.cli.main.main_subshell` by pass the argparse args:

- https://github.com/conda/conda/blob/a261ade3d0a276a214bff314e54c9a9dbb8c282e/conda/cli/main.py#L72

travishathaway (approved):
🚀

travishathaway:
Looking good. Thanks for being receptive to my suggestions 😄.

beeankha (approved):
Thanks for doing this work, @jaimergp !

jezdez (approved):

Issue #13396, opened 2023-12-04
Epic: `rename`/prefix-related fixes


beeankha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

This epic groups together a few issues related to placing guardrails against dangerous prefix paths as well as refactoring the `main_rename.py` module. All of these efforts are a result of research that was done for fixing the following bug:
* https://github.com/conda/conda/issues/13271

### Linked Issues & PRs

- [x] #13271 (
  - [ ] #13280)
  - [x] https://github.com/conda/conda/pull/13955
- [x] #13386
  - [x] #13387
- [x] #13395 
- [x] #11850

Issue #13386, opened 2023-12-01
Prevent protected directory paths from being able to be overriden as a conda environment prefix without a warning


beeankha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

While investigating a solution to [conda rename using prefix deleted environment after cloning and silently failed(?) #13271](https://github.com/conda/conda/issues/13271), it was discovered that there's a way to override protected directories (e.g., `bin`, `etc`, `conda-meta`, `envs`, etc.) by setting the protected subdirectory's path as the desired prefix. We need to set guardrails in place in order to prevent this from happening, since overriding those subdirectories can break a user's conda installation and/or unintentionally delete all of their environments.

PR #13955, opened 2024-05-31
Prevent protected directory paths from being able to be overridden as a conda environment prefix

beeankha:
### Description

This PR groups together a few fixes related to placing guardrails against "dangerous" prefix paths as well as refactoring the `main_rename.py` module in order to centralize some prefix checks in the `conda/cli/install.py` module (so that those checks can also be used by the `conda create` and `conda env create` commands). 

Additionally, the [`--force` flag in the `rename` command](https://github.com/conda/conda/blob/main/conda/cli/main_rename.py#L62-L68) is marked as deprecated per the work in this PR, in order to be consistent with modules like [`conda env create`](https://github.com/conda/conda/pull/13168/files#diff-3c57253ee42c86d103638d85e853415ae28aad66bb1e602a044916a18741bf4eR90-R100).

Related issues/epic:
- https://github.com/conda/conda/issues/13396
- https://github.com/conda/conda/issues/13271
- https://github.com/conda/conda/issues/13386

Previous attempt at this fix, for reference: 
- https://github.com/conda/conda/pull/13280

### Checklist - did you ...


kenodegard (changes requested):

conda/cli/install.py
```diff
@@ -64,7 +66,59 @@
...
+            if dir in os.listdir(parent_dir):
+                error = textwrap.dedent(
+                    f"""
+                    The specified prefix '{prefix}'
+                    contains a protected directory: {dir}. Cannot create an environment at this location.
+                    """
+                )
+
+    if error:
+        raise CondaEnvException(error, json)
```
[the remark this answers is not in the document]
beeankha: I've simplified this check to look for the `history` file within the `conda-meta` directory. Doing local testing with different scenarios (including naming an environment `conda-meta`, which is totally allowed 😆) shows that `conda rename` and `conda env create` both are still unable to accidentally recreate a scenario like what was posted in #13271.

tests/cli/test_main_install.py
```diff
@@ -0,0 +1,142 @@
+# Copyright (C) 2012 Anaconda, Inc
```
beeankha, line 1: These particular tests used to be in `tests/test_install.py` but since they test the `conda install` command I separated them into their own file in the `tests/cli` directory with the other CLI command-specific tests.

tests/env/test_create.py
```diff
@@ -286,3 +286,17 @@ def test_create_env_json(
 
     for string in stdout and stdout.split("\0") or ():
         json.loads(string)
+
+
+def test_protected_dirs_exist(conda_cli: CondaCLIFixture, tmp_env: TmpEnvFixture):
```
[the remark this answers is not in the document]
beeankha: Since the "`check_protected_dirs()`" guardrail now only checks for the existence of `/conda-meta/history` vs a list of potentially protected directories (which I simplified since it [might raise false alarms](https://github.com/conda/conda/pull/13955#discussion_r1628983021)), I just added another test in `test_main_rename.py` that tries to rename a protected directory instead of parametrizing in this file.

conda/cli/main_create.py
```diff
@@ -129,14 +129,17 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-            f"WARNING: A conda environment already exists at '{context.target_prefix}'\n"
-            "Remove existing environment",
+            f"WARNING: A conda environment already exists at '{context.target_prefix}'\n\n"
+            "Remove existing environment?\nThis will remove ALL directories contained within "
+            "this specified prefix directory, including any other conda environments.\n\n",
             default="no",
             dry_run=False,
         )
-        log.info("Removing existing environment %s", context.target_prefix)
+        log.info(f"Removing existing environment {context.target_prefix}")
```
kenodegard, line 138: We prefer to leave logging as `%` style formatting to avoid unnecessary templating when the log level results in the message not being printed:

```suggestion
        log.info("Removing existing environment %s", context.target_prefix)
```

kenodegard (changes requested):

conda/cli/install.py
```diff
@@ -64,7 +67,57 @@
...
+        if os.path.isdir(conda_meta_dir):
+            history_file = os.path.join(conda_meta_dir, "history")
+            if os.path.isfile(history_file):
+                raise CondaEnvException(
+                    f"The specified prefix '{prefix}' "
+                    "contains a protected directory, 'conda-meta', and a 'history' file within it. "
+                    "Creating an environment in this location will potentially risk the deletion "
+                    "of other conda environments. Aborting.",
+                    json,
+                )
```
jezdez, line 96: Should we use pathlib for this as well?

```suggestion
def check_protected_dirs(prefix: str | Path, json: bool = False) -> None:
    """Ensure that the new prefix does not contain protected directories."""
    parent_dir = Path(prefix).parent.resolve()

    if parent_dir.exists():
        conda_meta_dir = parent_dir / "conda-meta"
        if conda_meta_dir.is_dir():
            history_file = conda_meta_dir / "history"
            if history_file.is_file():
                raise CondaEnvException(
                    f"The specified prefix '{prefix}' "
                    "contains a protected directory, 'conda-meta', and a 'history' file within it. "
                    "Creating an environment in this location will potentially risk the deletion "
                    "of other conda environments. Aborting.",
                    json,
                )
```

conda/cli/main_rename.py
```diff
@@ -83,27 +88,34 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
 
 
+@deprecated(
+    "24.9",
+    "25.3",
+    addendum="Use `conda.cli.install.validate_new_prefix` instead.",
+)
 def validate_destination(dest: str, force: bool = False) -> str:
     """Ensure that our destination does not exist"""
+    import os
```
jezdez, line 117: Let's move this import out of the function

tests/cli/test_main_install.py
```diff
@@ -0,0 +1,142 @@
...
+            assert prefix_data.get(pkg["name"]).version == pkg["version"]
+
+
+def test_install_mkdir(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
+    with tmp_env() as prefix:
+        file = prefix / "tempfile.txt"
+        file.write_text("test")
+        dir = prefix / "conda-meta"
+        assert isdir(dir)
+        assert exists(file)
```
jezdez, line 52:
```suggestion
        assert dir.is_dir()
        assert file.exists()
```

jezdez (commented):
Just a few pathlib related comments

beeankha:
pre-commit.ci autofix

jezdez (dismissed):

conda/cli/install.py
```diff
@@ -64,7 +67,57 @@
 stderrlog = getLogger("conda.stderr")
 
 
-def check_prefix(prefix, json=False):
+def validate_prefix_exists(prefix: str | Path) -> Path:
```
jaimergp, line 70: Should we always promote to `Path`? Looks like a side effect that might be surprising to those passing a string.

conda/cli/install.py
```diff
@@ -64,7 +67,57 @@
...
+                    json,
+                )
+
+
+def validate_new_prefix(dest: str, force: bool = False) -> str:
+    """Ensure that the new prefix does not exist."""
+    from ..base.context import context, validate_prefix_name
+    from ..common.path import expand
+
+    if os.sep in dest:
```
jaimergp, line 104: What if someone passes a path like `some/thing` in Windows? `os.sep` is `\` there.

beeankha: To address this concern, I initially added a check in the `validate_destination()` function in `main_rename.py` to ensure that `/` was not allowed in Windows (and alternatively that `\` was not allowed on UNIX systems). However, this appeared to be redundant since the [`PREFIX_NAME_DISALLOWED_CHARS` constant](https://github.com/conda/conda/blob/main/conda/base/constants.py#L181-L186) (and the [recently-added `PREFIX_NAME_DISALLOWED_CHARS_WIN`](https://github.com/conda/conda/pull/13975/files#diff-3fe51c3b31f7c2db9a7e6b7616764e3f7777393adf967debc093939a6eae1d87R188)) already checks for specific characters like this. 

I [added `\` as a disallowed character in `constants.py`](https://github.com/conda/conda/pull/13955/files#diff-3fe51c3b31f7c2db9a7e6b7616764e3f7777393adf967debc093939a6eae1d87R186) and also added [a couple of tests in `test_main_rename.py`](https://github.com/conda/conda/pull/13955/files#diff-8d2bdaf3c1457e2fd059bd64d06d67132b20f427802fced0162693334dd50a38R307-R334) to ensure that the `PREFIX_NAME_DISALLOWED_CHARS` and `PREFIX_NAME_DISALLOWED_CHARS_WIN` constants are indeed filtering out those characters. Please let me know what you think of these changes!

jaimergp: Wait, does this mean that if I pass a path like `C:\Something\With^%disallowed?characters` directly, the path won't be validated, only expanded, because of the presence of `\`? I don't understand why it needs to be exclusive :/

Additionally, I don't think it'll cause problems, but _technically_  `\` is an allowed character in Unix paths, right?

beeankha: > _technically_ `\ `is an allowed character in Unix paths, right?

Yes, but allowing it here will potentially break cross-platform compatibility for things like `environment.yml` so I think it's best to leave it in the "disallowed characters" list.

One thing I'm wondering is if we should maybe combine `PREFIX_NAME_DISALLOWED_CHARS` and `PREFIX_NAME_DISALLOWED_CHARS_WIN` and default to disallowing more characters (so default to `PREFIX_NAME_DISALLOWED_CHARS_WIN`) in order to ensure cross-platform compatibility?

jaimergp: If the two lists are converging into the same thing, yes. You make a good point about cross-platform compat. Maybe that deserves its own PR for traceability/discussion in case it breaks something.

My only concern is that this change ends up preventing the activation and manipulation of existing environments because of validation happening for anything passed via `-n` (`conda install -n 'this used \\ to work' ...`).

kenodegard (changes requested):

beeankha:
pre-commit.ci autofix

conda/cli/install.py
```diff
@@ -64,7 +67,51 @@
...
+    prefix = Path(prefix)
+    if not prefix.exists():
+        raise CondaEnvException("The environment you have specified does not exist.")
+
+
+def check_protected_dirs(prefix: str | Path, json: bool = False) -> None:
+    """Ensure that the new prefix does not contain protected directories."""
+    history_file = Path(prefix).parent / "conda-meta" / "history"
+
+    if history_file.is_file():
```
kenodegard, line 83: Hm the recent convos in https://github.com/conda/conda/pull/14112 reminds me that we probably want to use `conda.gateways.disk.tests.is_conda_environment` instead. Also begs the question whether these helpers should also be defined in `conda.gateways.disk.test`.

conda/cli/install.py
kenodegard: This function was previously defined in `conda.cli.main_rename`. Since this is simply moving the function and not changing it I'd be fine with leaving this untouched and revisit it at a later time.

conda/cli/main_rename.py
```diff
@@ -57,47 +62,68 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+    else:
+        check_protected_dirs(prefix)
 
-    return context.target_prefix
+    return str(prefix)
 
 
+@deprecated(
+    "24.9",
+    "25.3",
```
kenodegard, line 124: Need to bump this until next year:

```suggestion
    "25.3",
    "25.9",
```

kenodegard (changes requested):

kenodegard (approved):

jezdez (approved):

PR #13963, opened 2024-06-07
Replace logger.warn w/ logger.warning

zklaus:
### Description

This replaces calls to `logger.warn` with `logger.warning`. The former has been deprecated since Python 3.3 and generates `DeprecationWarning`s that can make output particularly in debugging and testing more noisy than necessary.

Sibling to conda/conda-build#5364.

### Checklist - did you ...

As this is a purely internal change, I don't thing any of the following items apply.
However, it does remove the hack in gateways.logging to suppress the deprecation warnings.
Does that need special consideration?

### Other repos
@jezdez noted that the same issue is present in other repos in the ecosystem.
They may be addressed in a similar manner and in detail are the following:

- [x] conda/conda-index (conda/conda-index#173)
- [x] conda/conda-libmamba-solver (conda/conda-libmamba-solver#476)
- [x] conda/conda-lock (conda/conda-lock#646)
- [x] conda/conda-verify (conda/conda-verify#115)
- [x] conda/menuinst (conda/menuinst#220)

### DIFF
--- a/conda/gateways/logging.py
+++ b/conda/gateways/logging.py
@@ -253,6 +253,3 @@ def trace(self, message, *args, **kwargs):
 
 logging.addLevelName(TRACE, "TRACE")
 logging.Logger.trace = trace  # type: ignore[attr-defined]
-
-# suppress DeprecationWarning for warn method
-logging.Logger.warn = logging.Logger.warning  # type: ignore[method-assign]
--- a/conda/core/envs_manager.py
+++ b/conda/core/envs_manager.py
@@ -51,7 +51,7 @@ def register_env(location):
     try:
         os.makedirs(user_environments_txt_directory, exist_ok=True)
     except OSError as exc:
-        log.warn(
+        log.warning(
             "Unable to register environment. "
             f"Could not create {user_environments_txt_directory}. "
             f"Reason: {exc}"
@@ -64,7 +64,7 @@ def register_env(location):
             fh.write("\n")
     except OSError as e:
         if e.errno in (EACCES, EROFS, ENOENT):
-            log.warn(
+            log.warning(
                 "Unable to register environment. Path not writable or missing.\n"
                 "  environment location: %s\n"
                 "  registry file: %s",
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -1587,7 +1587,7 @@ def run_script(
                     )
                 raise LinkError(message)
             else:
-                log.warn(
+                log.warning(
                     "%s script failed for package %s\n"
                     "consider notifying the package maintainer",
                     action,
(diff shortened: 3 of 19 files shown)


conda/gateways/logging.py
```diff
@@ -253,6 +253,3 @@ def trace(self, message, *args, **kwargs):
 
 logging.addLevelName(TRACE, "TRACE")
 logging.Logger.trace = trace  # type: ignore[attr-defined]
-
-# suppress DeprecationWarning for warn method
-logging.Logger.warn = logging.Logger.warning  # type: ignore[method-assign]
```
jezdez, line 258: This _may_ result in additional deprecation warnings, if any of the conda-related packages use the old name, as suggested in https://github.com/search?q=org:conda+lang:python+.warn(&type=code

I would keep it, but deprecate it with our `deprecated.topic` tool. @kenodegard do you think that's appropriate?

zklaus: You are right that this might lead to more deprecation warnings. A refined search (excluding `warnings.warn`, which remains correct, and the conda and conda-build repos, which are dealt with in this PR and #5364 and #5365) [shows matches in 26 files](https://github.com/search?q=org%3Aconda+lang%3Apython+%22.warn%28%22+NOT+%22warnings.warn%28%22+NOT+repo%3Aconda%2Fconda+NOT+repo%3Aconda%2Fconda-build&type=code).

Perhaps we should address those repos as well?

jezdez: I think so, and please remember that no all dependencies are always installing the newest version of something, e.g. by monkeypatching something like that, the conda maintainers have made it really hard to undo it again, unless the minimal conda versions in the other packages are bumped

kenodegard: not sure how much added value there is in us also marking this as deprecated when it’s already marked as deprecated in stdlib

plus marking this as deprecated with our tooling will be unnecessarily complex (our tooling assumes we are deprecating stuff within our own modules, not deprecating monkeypatches from other modules)

+1 on doing a best effort to fine any `logging.warn` in conda and conda-incubator projects and updating those

zklaus: Those in `conda` are fixed (see linked PRs in description), surprisingly, there doesn't seem to be anything in `conda-incubator`, see https://github.com/search?q=org%3Aconda-incubator+lang%3Apython+%22.warn%28%22&type=code.

kenodegard (dismissed):

beeankha (approved):

conda/gateways/logging.py
jezdez: Thanks @kenodegard!

jezdez:
@zklaus Could you please add a news file?

PR #13969, opened 2024-06-11
Checksum comparisons should be case insensitive - use bytes directly

jaimergp:
### Description

`rattler` returns `sha256` with uppercase alpha characters. Python's `hashlib.hexdigest()` returns it with lowercase. Either works and both are ok since SHA256 should be case insensitive. We just need to make our comparisons insensitive too.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/download.py
+++ b/conda/gateways/connection/download.py
@@ -180,7 +180,7 @@ def download_partial_file(
             checksum = sha256 if sha256 else md5
             try:
                 checksum_bytes = bytes.fromhex(checksum)
-            except ValueError as exc:
+            except (ValueError, TypeError) as exc:
                 raise CondaValueError(exc) from exc
             hasher = hashlib.new(checksum_type)
             target.seek(0)
--- a/conda/gateways/connection/download.py
+++ b/conda/gateways/connection/download.py
@@ -176,7 +176,7 @@ def download_partial_file(
         target.seek(0)
         if md5 or sha256:
             checksum_type = "sha256" if sha256 else "md5"
-            checksum = sha256 if sha256 else md5
+            checksum = (sha256 if sha256 else md5).lower()
             hasher = hashlib.new(checksum_type)
             target.seek(0)
             while read := target.read(CHUNK_SIZE):
--- a/conda/gateways/connection/download.py
+++ b/conda/gateways/connection/download.py
@@ -24,6 +24,7 @@ from ...exceptions import (
     CondaDependencyError,
     CondaHTTPError,
     CondaSSLError,
+    CondaValueError,
     ProxyError,
     maybe_raise,
 )
@@ -176,15 +177,18 @@ def download_partial_file(
         target.seek(0)
         if md5 or sha256:
             checksum_type = "sha256" if sha256 else "md5"
-            checksum = (sha256 if sha256 else md5).lower()
+            checksum = sha256 if sha256 else md5
+            try:
+                checksum_bytes = bytes.fromhex(checksum)
+            except ValueError as exc:
+                raise CondaValueError(exc) from exc
             hasher = hashlib.new(checksum_type)
             target.seek(0)
             while read := target.read(CHUNK_SIZE):
                 hasher.update(read)
 
-            actual_checksum = hasher.hexdigest()
-
-            if actual_checksum != checksum:
+            if hasher.digest() != checksum_bytes:
+                actual_checksum = hasher.hexdigest()
                 log.debug(
                     "%s mismatch for download: %s (%s != %s)",
                     checksum_type,
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -2,9 +2,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import hashlib
 import os
+from contextlib import nullcontext
 from os.path import exists, isfile
 from pathlib import Path
 from tempfile import mktemp
+from typing import Any, Callable
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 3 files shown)


jezdez (approved):

jezdez:
Not sure if there is a test for this, but it might be useful to add one

dholth:
This is a change in the file format.

jaimergp:
> This is a change in the file format.

Can you elaborate? Which file?

jaimergp:
> Not sure if there is a test for this, but it might be useful to add one

Added one :)

dholth:
OK, maybe these uppercased checksums are only coming from a rattler API.

Should the news mention that the comparison is case-insensitive in (the name of the function)

conda/gateways/connection/download.py
```diff
@@ -176,7 +176,7 @@ def check(target):
         target.seek(0)
         if md5 or sha256:
             checksum_type = "sha256" if sha256 else "md5"
-            checksum = sha256 if sha256 else md5
+            checksum = (sha256 if sha256 else md5).lower()
```
dholth, line 179: A different way: `bytes.fromhex(checksum) == hasher.digest()`
An even more different way: `int(hashlib.sha256().hexdigest(),16)`

dholth: https://github.com/dholth/conda/commit/54e01e54290b349ad569816971af10173f130c57

jezdez: Note, that if we're creating a bytes object with the `fromhex` classmethod, we'll need to account for input validation (e.g. `bytes.fromhex("foobar")` will raise a `ValueError`) while the current string based comparison is more forgiving

dholth: Observe that in the sharded repodata CEP, checksums are transmitted as bytes and are not converted to hex.

wolfv: We can change this in rattler if it's faster that way :D Maybe we should actually type the checksum as a bytes object that is returned to Python anyways.

wolfv: here is the PR: https://github.com/mamba-org/rattler/pull/752

wolfv: That change is merged @jaimergp - not sure when we'll do the next release though as a few people are on vacation.

jaimergp: Thanks @wolfv! And no worries, no rush :)

jaimergp: @dholth @jezdez - added checks in bytes now. It also tests for bad strings.

jezdez (dismissed):

jaimergp:
@conda/conda-core - this is ready to merge (the previously submitted approval was dismissed by the last commit).

tests/test_fetch.py
```diff
@@ -347,3 +348,33 @@ def __init__(self, status_code):
...
+                size=size,
+                sha256=123456 if sha256_type == "bad-type" else "not-an-hex-string",
+            )
+    else:
+        download(
+            url,
+            output_path,
+            size=size,
+            sha256=sha256.upper() if sha256_type == "upper" else sha256,
+        )
```
kenodegard, line 380: I prefer using a `contextlib.nullcontext` instead of repeating the download logic:

```
@pytest.mark.parametrize(
    "raises,get_sha256",
    [
        pytest.param(False, lambda x: x, id="original"),
        pytest.param(False, str.upper, id="upper"),
        pytest.param(True, lambda x: "not-an-hex-string", id="gibberish"),
        pytest.param(True, lambda x: 123456, id="bad-type"),
    ],
)
def test_checksum_checks_bytes(
    tmp_path: Path,
    package_repository_base,
    package_server,
    raises: bool,
    get_sha256: Callable[[str], Any],
):
    host, port = package_server.getsockname()
    base = f"http://{host}:{port}/test"
    package_name = "zlib-1.2.11-h7b6447c_3.conda"
    url = f"{base}/linux-64/{package_name}"
    package_path = package_repository_base / "linux-64" / package_name
    sha256 = checksum(package_path, algorithm="sha256")
    size = package_path.stat().st_size
    output_path = tmp_path / package_name

    with pytest.raises(CondaValueError) if raises nullcontext():
        download(url, output_path, size=size, sha256=get_sha256(sha256))
```

conda/gateways/connection/download.py
```diff
@@ -177,14 +178,17 @@ def check(target):
         if md5 or sha256:
             checksum_type = "sha256" if sha256 else "md5"
             checksum = sha256 if sha256 else md5
+            try:
+                checksum_bytes = bytes.fromhex(checksum)
+            except (ValueError, TypeError) as exc:
+                raise CondaValueError(exc) from exc
```
kenodegard, line 184: Isn't the `from exc` already the default behavior?

kenodegard (approved):

conda/gateways/connection/download.py
jaimergp: What do you mean by default behavior here? The output is different in this test:

```
>>> try:
...   raise ValueError("First exception")
... except ValueError as exc:
...   raise TypeError(exc) from exc
... 
  File "<stdin>", line 2, in <module>
ValueError: First exception

The above exception was the direct cause of the following exception:

  File "<stdin>", line 4, in <module>
TypeError: First exception
>>> try:
...   raise ValueError("First exception")
... except ValueError as exc:
...   raise TypeError(exc)
... 
  File "<stdin>", line 2, in <module>
ValueError: First exception

  File "<stdin>", line 4, in <module>
TypeError: First exception
```

tests/test_fetch.py
jaimergp: Oh, nice.  Thanks! Committed the result.

conda/gateways/connection/download.py
jaimergp: Yea, I wouldn't force this on every contributor. I do it out of habit when I'm explicitly turning one exception into another so there's a clear intent for that papertrail. But the difference might be too small to make it, e.g., an imposed code style rule for the project. There's a nice discussion about this in [this SO answer](https://stackoverflow.com/questions/24752395/python-raise-from-usage) and in the [originating PEP](https://legacy.python.org/dev/peps/pep-3134/).

kenodegard (approved):

Issue #11092, opened 2021-12-08
`conda env remove` does not run unlink transactions (pre_unlink, remove_menus, etc)


jaimergp:
### Checklist
* [x] I added a descriptive title
* [x] I searched for other issues and couldn't find a solution or duplication
* [x] I already searched in Google and didn't find any good information or help
* [x] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?
One way to delete environments is through `conda env remove -n name`. 

My expectation is that this will uninstall all packages in the environment and then remove the directory. 

However, it doesn't do so unless a [list of packages is passed](https://github.com/jaimergp/conda/blob/5b847665f02a0495d2eda50ec74be86745df063d/conda/cli/main_remove.py#L62) (?), possibly to optimize some IO; it goes straight to a `rm -rf prefix` call, bypassing ALL unlink actions. This is because `package_names` is (perhaps deliberately) [left out of the args list](https://github.com/jaimergp/conda/blob/5b847665f02a0495d2eda50ec74be86745df063d/conda_env/cli/main_remove.py#L44-L47) in the call from `conda env remove` to `conda remove`.

The easy fix would be to add `package_names=None` to the passed args, but this might have been an intentional decision. I am not aware if this is documented, though.

This results in some artifacts left in the system, like the shortcuts created by menuinst, which should be removed by the RemoveMenuAction objects created during the unlink transaction.

### Conda info
<details>
<summary><code>conda info</code></summary>

```shell
conda info

     active environment : None
       user config file : /Users/jrodriguez-guerra/.condarc
 populated config files : /Users/jrodriguez-guerra/.condarc
          conda version : 4.10.3.post86+5b847665f
    conda-build version : not installed
         python version : 3.9.7.final.0
       virtual packages : __osx=11.6.1=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /Users/jrodriguez-guerra/.local/conda/envs/pydantic  (writable)
      conda av data dir : /Users/jrodriguez-guerra/.local/conda/envs/pydantic/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /Users/jrodriguez-guerra/.local/conda/envs/pydantic/pkgs
                          /Users/jrodriguez-guerra/.conda/pkgs
[...]

PR #13977, opened 2024-06-12
Alias `conda env remove` command to `conda remove --all`

beeankha:
This PR targets Step 2 (removing duplicate subcommands) of the ["refactoring and deprecating `conda env`-related commands and modules"](https://github.com/conda/conda/issues/11633) epic. Specifically, this PR aliases the `conda env remove` command to `conda remove --all`; this means that even though the execution of the command is mostly reliant on `main_remove.py`, anyone who uses/prefers the older `conda  env remove` command will still be able to get the results that they expect.

Depends-on:
- https://github.com/conda/conda/pull/14173


kenodegard:
What do we need to do in order to make `conda remove` and `conda env remove` look and act the same? I.e. same arguments and same default behavior?

jezdez:
> What do we need to do in order to make `conda remove` and `conda env remove` look and act the same? I.e. same arguments and same default behavior?

- review the current options
- compare the implementations and find differences
- refactor argument handling code if needed

beeankha:
> What do we need to do in order to make conda remove and conda env remove look and act the same? I.e. same arguments and same default behavior?

I might be misunderstanding (so please correct me if I'm wrong!), but do we actually _want_ to make `conda env remove` do the same things as/have the extra features that `conda remove` has? `conda env remove` is established as only being able to remove specified environments and all of the packages within it, while `conda remove` has the same functionality and more. Thus, anyone who has been using `conda env remove` will only expect the single use case, and we're aiming to functionally alias/deprecate that command with this PR; I feel like if we make `conda env remove` able to do the same things as `conda remove`, that would _expand_ the functionality of a `conda env` command which (from my understanding) doesn't align with what we want to do with the "`conda env` collapse" efforts.

jaimergp:
`conda remove --all` is preferred also because of this issue: https://github.com/conda/conda/issues/11092

conda/cli/main_env_remove.py
```diff
@@ -52,34 +54,28 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+    p.set_defaults(
+        func="conda.cli.main_remove.execute",
+        all=True,
+        channel=None,
+        features=None,
+        override_channels=None,
+        use_local=None,
+        use_cache=None,
+        offline=None,
+        package_names="",
```
beeankha, line 66: Before this `package_names=""` addition, the output for `conda env remove [environment]` was as follows:

```
Remove all packages in environment /conda/.../envs/remove_me:

Everything found within the environment /conda/.../envs/remove_me), including any conda environment configurations and any non-conda files, will be deleted. Do you wish to continue?
 (y/[n])? 
 ```
 
 After the addition, the command is much more like `conda remove [environment] --all`:
 
 ```
 Remove all packages in environment /conda/.../envs/remove_me:


## Package Plan ##

  environment location: 
/conda/.../envs/remove_me


The following packages will be REMOVED:

  bzip2-1.0.8-h80987f9_6
  ca-certificates-2024.7.2-hca03da5_0
  expat-2.6.2-h313beb8_0
  libcxx-14.0.6-h848a8c0_0
  libffi-3.4.4-hca03da5_1
  ncurses-6.4-h313beb8_0
  openssl-3.0.14-h80987f9_0
  pip-24.2-py312hca03da5_0
  python-3.12.4-h99e199e_1
  readline-8.2-h1a28f6b_0
  setuptools-72.1.0-py312hca03da5_0
  sqlite-3.45.3-h80987f9_0
  tk-8.6.14-h6ba3021_0
  tzdata-2024a-h04d1e81_0
  wheel-0.43.0-py312hca03da5_0
  xz-5.4.6-h80987f9_1
  zlib-1.2.13-h18a0788_1


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Everything found within the environment (/conda/.../envs/remove_me), including any conda environment configurations and any non-conda files, will be deleted. Do you wish to continue?
 (y/[n])?
 ```

**Review by travishathaway — DISMISSED** [review]
🚀

**Comment by travishathaway** [comment]
I don't think this makes that big of difference, but the default option that is chosen when removing an environment differs. For the old `conda env remove` it defaults to `no` but for `conda remove --all` it defaults to `yes`.

Again, not that big of a concern, but it's worth pointing out.

File: conda/cli/main_env_remove.py [hunk 2]
```diff
@@ -11,6 +11,8 @@
     _SubParsersAction,
 )
 
+from conda.deprecations import deprecated
```
jezdez, line 14:
```suggestion
from ..deprecations import deprecated
```

conda/cli/main_env_remove.py
```diff
@@ -52,34 +54,28 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+        package_names=[],
+        force=True,
+        pinned=None,
+        keep_env=False,
+    )
 
     return p
 
 
+@deprecated("25.3", "99.9", addendum="Use `conda.cli.main_remove.execute` instead.")
```
jezdez, line 75: I'm not sure if that is a helpful deprecation notice, since it'll be triggered when a user is calling `conda env remove`. Using the function may be misleading basically, how about this:
 ```suggestion
@deprecated("25.3", "99.9", addendum="Use `conda remove --all ...` instead.")
```

Comment (kenodegard) [reply conda/cli/main_env_remove.py:75]: This is not necessary. Users no longer invoke this `execute` function since the `func` default (see above) is now `conda.cli.main_remove.execute` instead.

File: tests/cli/test_env.py [hunk 4]
```diff
@@ -375,12 +375,11 @@ def test_update_env_no_action_json_output(
...
 
 @pytest.mark.integration
-def test_remove_dry_run(env1: str, conda_cli: CondaCLIFixture):
+def test_remove_dry_run(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
     # Test for GH-10231
-    create_env(ENVIRONMENT_CA_CERTIFICATES)
-    conda_cli("env", "create")
-    conda_cli("env", "remove", f"--name={env1}", "--dry-run")
-    assert env_is_created(env1)
+    with tmp_env() as prefix:
```
kenodegard, line 380: This is no longer the same test since `ca-certificates` is not installed, was this an intentional change?

conda/cli/main_env_remove.py
jezdez: Oh excellent, I missed that!

jezdez (dismissed):

conda/cli/main_env_remove.py
kenodegard, line 75: Oh we want an actual removal version here though since this execute function no longer serves a purpose.

```suggestion
@deprecated("25.3", "25.9", addendum="Use `conda.cli.main_remove.execute` instead.")
```

kenodegard (changes requested):

kenodegard (approved):

Issue #13610, opened 2024-02-20
Conda cannot be activated on Windows if AppLocker is used


pnsaevik:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The problem is described in the issue https://github.com/ContinuumIO/anaconda-issues/issues/9450

In short, when AppLocker is used on a Windows system (which is a common company-wide policy), conda is blocked.

The user gets the following unhelpful response: "This program is blocked by group policy. For more information, contact your system administrator."

The problem is that conda activation creates a temporary script with a random name, puts it in the 'Temp' folder, and tries to execute it. This looks too much like malware to AppLocker, so it gets blocked.

Something similar happens if one tries to activate conda through PowerShell.

I don't know what would be the best solution to this problem. A workaround is outlined in https://github.com/ContinuumIO/anaconda-issues/issues/9450.

### Additional Context

![image](https://github.com/conda/conda/assets/12728107/f79f3edf-a654-4275-a360-2fe30b93bb71)

travishathaway:
Hi @pnsaevik,

Thanks for filing this detailed issue, and we also really appreciate the proposed work around that you have shared with others. It appears that this problem is widespread enough to warrant potentially updating conda itself to use a solution like yours.

At this point, I think it would make sense to work together to draft a pull request in order to further discuss your solution and see how we could possibly integrate this into conda. Would you be able/willing to create this pull request yourself? Is there anything we could help with?

I'm going to share this issue with other maintainers so they can chime in on the discussion too.

skupr-anaconda:
Could you please clarify if it will behave in the same way as before if you run the `Anaconda Prompt` as **Administrator**?
![Screenshot 2024-02-21 at 10 39 37](https://github.com/conda/conda/assets/79282962/81967def-2571-46cb-abfb-ab0c505d7ddf)

pnsaevik:
Hi @travishathaway, I'm quite busy the coming month, but after that I might be able to contribute with some code/pull request. But there are other challenges as well. My proposed workaround is hackish and might be patched by Windows updates. While local variants of the workaround are possible, a standardized, open-source implementation might be vulnerable to exploits. I fear that a "proper" solution would require a more substantial rewrite of the activation system, at the risk of breaking established workflows for other users. If people in the development team are familiar with such issues, we should discuss alternative solutions before we start writing code.

pnsaevik:
[...]

PR #13979, opened 2024-06-13
Remove intermediate temporary file for CMD activation

kenodegard:
### Description

Previously `conda activate` in Windows CMD would first create a temporary file at `%TMP%\conda-%RANDOM%\conda.tmp`, it would then call `conda shell.cmd.exe activate` which in turn would write the activation code to another temporary file (with a randomized name), this randomized temporary file name is then printed to stdout and stored in the first temporary file.

This PR removes the second temporary file and instead redirects the activation code to the first temporary file.

Xref #13610

### Checklist - did you ...


kenodegard:
Superseded by https://github.com/conda/conda/pull/14607

Issue #12853, opened 2023-07-06
`conda doctor` should suggest how to fix the diagnosed problems 


ForgottenProgramme:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

conda doctor should make suggestions on how the user might be able to fix the diagnosed problems. 
For example in case of reported missing files, `conda doctor` might suggest uninstalling and then reinstalling the corrupt packages. 
In some cases it might not be straightforward to suggest a solution to fix the failing health check. In such a case `conda doctor` might explain "why" the health check might be failing. 

### Why is this needed?

It aligns with the goal of helping conda users help themselves.

### What should happen?

`conda doctor` also makes suggestions on how to fix the diagnosed problem.

PR #14037, opened 2024-07-15
Add "requests_ca_bundle" health check to `conda doctor`

### LINKED ISSUE #12905: Add health check for environment variables like `REQUESTS_CA_BUNDLE`
### Checklist

### What is the idea?

The env variable REQUESTS_CA_BUNDLE can break conda/anaconda-client if it is pointing to a file that doesn't exist.  Since anaconda-client is integrated/used by conda this can cause problems even if conda has `ssl_verify: false` set.  

It would be nice to catch and report this error so a user knows what to do to fix it -- the windows command to run and the *nix command to run to fix this error.

### Why is this needed?

This is a rare but very subtle problem that most users wouldn't notice.  It's an environment variable that can break anaconda-client and therefore break conda when running normal commands.  But since the problem is twice removed from conda, it is hard to diagnose (ie: it is an env variable used by anaconda-client that breaks conda commands). 

### What should happen?

`conda doctor` could do any of the following:

1. report that this environment variable is set and affects operations of conda (???) and anaconda-client
2. If it is pointing to a file that doesn't exist or has bad file permissions, report that it is a bad setting and how to fix it.

### Additional Context

_No response_

ForgottenProgramme:
### Description

resolves #12905 

Add a health check to verify the env var REQUESTS_CA_BUNDLE. 

- [x] check that it points to an existent path
- [x] if yes, verify the ca bundle at this path
- [x] Use sessions 

### Checklist - did you ...


tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -245,6 +247,20 @@ def test_not_env_txt_check_action(
     assert X_MARK in captured.out
 
 
+def test_requests_ca_bundle_check_action(
+    env_ok: tuple[Path, str, str, str, str], capsys
+):
+    prefix, _, _, _, _ = env_ok
+    os.environ["REQUESTS_CA_BUNDLE"] = "non/existent/path"
```
kenodegard, line 254: Let's use `monkeypatch` here instead in order to properly cleanup after the test completes

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -163,8 +163,19 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
     print(f"Environment listed in environments.txt file: {present}\n")
 
 
+def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
+    import os
```
kenodegard, line 167: stdlib imports should be left at the top of the file

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -163,8 +163,19 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
...
+
+    if os.environ.get("REQUESTS_CA_BUNDLE"):
+        if not Path(os.environ.get("REQUESTS_CA_BUNDLE")).exists():
```
kenodegard, line 170:
```suggestion
    if (path := os.getenv("REQUESTS_CA_BUNDLE")) and not Path(path).exists():
```

kenodegard (changes requested):

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: @kenodegard , then we don't need the following two lines, right?
https://github.com/conda/conda/blob/main/conda/plugins/subcommands/doctor/health_checks.py#L19-L20

travishathaway (changes requested):
I really liked @bkreider's idea about suggesting how to fix this if problems are found. Could we implement that here? It would also be nice to use the `conda.common.compat.os_win` variable to conditionally show either Windows or *nix help.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -163,8 +160,16 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
     print(f"Environment listed in environments.txt file: {present}\n")
 
 
+def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
+    if (path := os.getenv("REQUESTS_CA_BUNDLE")) and not Path(path).exists():
```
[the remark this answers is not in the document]
ForgottenProgramme: Yeah. I was in two minds about this when I wrote the `REQUESTS_CA_BUNDLE` check. 
Displaying a "health check passes ✅ " message is consistent with how we display outputs for other health checks.  
At the same time I thought "REQUESTS_CA_BUNDLE check passes" would confuse users who are not advanced and don't know what `REQUESTS_CA_BUNDLE` is. 🤔 

BUT 
Now that I'm thinking about it again: maintaining consistency is probably a better idea. 👍🏼 
I would change the "check passes" message to "env var `REQUESTS_CA_BUNDLE` points to a valid file. ✅ " so that it's more descriptive for non-advanced users.

ForgottenProgramme:
@travishathaway, I agree. There is an [open issue](https://github.com/conda/conda/issues/12853) for making `conda doctor` suggest how to fix diagnosed problems. I am thinking of implementing this in the coming days.
However, I will keep the scope of this PR limited to only adding this new health check.

travishathaway:
> @travishathaway, I agree. There is an [open issue](https://github.com/conda/conda/issues/12853) for making `conda doctor` suggest how to fix diagnosed problems. I am thinking of implementing this in the coming days. However, I will keep the scope of this PR limited to only adding this new health check.

Thanks for linking to that issue. If there's already a separate issue for that, let's address the suggestions there. I'm okay with leaving that out for this pull request.

beeankha (approved):

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -163,8 +160,20 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
...
```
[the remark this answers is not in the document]
ForgottenProgramme: I agree with you. 
Do you think we should go one step further and check if the file is corrupt or not?
Or another option could be to add a note saying that the file exists but `conda doctor` doesn't check for the validity of the certificate bundle or whether it's a corrupt file.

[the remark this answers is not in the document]
ForgottenProgramme: I updated the PR. Now I am first checking if the env var is set or not, if yes, then I check if the path exists or not, if yes, I do a `requests.get()` to see if the ca_bundle is valid or not.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -163,8 +162,29 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
...
+def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
+    if not os.getenv("REQUESTS_CA_BUNDLE"):
+        return
+    elif not Path(os.getenv("REQUESTS_CA_BUNDLE")).exists():
+        print(
+            f"{X_MARK} Env var `REQUESTS_CA_BUNDLE` is pointing to a non existent file.\n"
+        )
+    else:
+        try:
+            response = requests.get("https://example.com")
```
travishathaway, line 174: There are two things I'd like to mention for this line.

In conda, we do not use `requests` directly. Instead, we use the `get_session` function which returns a `CondaSession` object. We should also do this here so the doctor check is as similar to regular conda HTTP requests as possible.

For an example, of how this is used please see:

- https://github.com/conda/conda/blob/6148dda2527f610c658dab8e6172b07b6c64004a/conda/notices/fetch.py#L64

The second thing I'd like to mention is that we should probably use a different URL. I would suggest either `https://conda.org` or `https://anaconda.org`.

@kenodegard, do you have any thoughts on the test URL?

travishathaway (changes requested):
I think cod coverage is a little off because when I run the tests locally, I see that these lines are not covered:

Bildschirmfoto 2024-08-05 um 16 07 00

I would really like to see 100% coverage on all the new lines written. I would highly suggest utilizing mocks to accomplish this. Using mocks, we could reduce the need for external network requests in the tests too. This should help the test run faster.

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: Thanks. I will look into the `CondaSession` suggestion. 
So I had the same question. I wanted to use conda.org too. But I learned that https://example.com is a valid website and it is, IIUC, guaranteed to always be available.

kenodegard: Both conda.org and anaconda.org will give false negatives for those with restrictive intranets. What if we just use one of the urls the user is expecting to work? e.g.:

```python
conda.models.channel.Channel(conda.base.context.context.channels[0]).urls()[0]
```

kenodegard: could also iterate over all of those URLs and report whether or not they work? would be another nice health check for those trying to debug proxy/network issues etc.

travishathaway: @kenodegard,

That was my first thought. I vote for that approach.

ForgottenProgramme:
pre-commit.ci autofix

ForgottenProgramme:
On further discussions it was realized that using `context.channel_alias.urls()[0]` is more appropriate as using `context.channel_alias.urls()[0]`  will also account for the cases where the user's machine is connected to an intranet (as  `context.channel_alias.urls()[0]` will in that case point to an internal url).

ForgottenProgramme:
About "example.com": 
It is an IANA reserved domain. More information here: https://www.iana.org/help/example-domains

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -5,22 +5,25 @@
...
-    import os
-
 logger = getLogger(__name__)
 
 OK_MARK = "✅"
 X_MARK = "❌"
+CA_BUNDLE_TEST_URL = context.channel_alias.urls()[
+    0
+]  # using one of the channel aliases url ensures that the health check also runs on machines that are on an intranet connection. As in such a case the `context.channel_alias.urls()[0]` might be pointing to an internal url.
+# CA_BUNDLE_TEST_URL = "https://example.com"  # IANA reserved domain (more info: https://www.iana.org/help/example-domains)
```
jezdez, line 26: Accessing `context.channel_alias` outside the function may trigger an import-time side effect, which might be a problem since plugins are loaded pretty early in the startup process of conda. Let's move that into the check function below (I'll add a suggestion)

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -5,22 +5,25 @@
 from __future__ import annotations
 
 import json
+import os
 from logging import getLogger
 from pathlib import Path
-from typing import TYPE_CHECKING
+
+import requests
```
jezdez, line 12: Please use the conda session for this (as it pre-configures authentication, TLS certs and proxy handling):

(please move this to the correct import line below)

```suggestion
from ...gateways.connection.session import get_session
```

See my suggestion below how to get the session object (which has the same interface as if you'd use the "requests" module you imported here first.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -150,8 +153,29 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
...
+        )
+    else:
+        try:
+            response = requests.get(CA_BUNDLE_TEST_URL)
+            if response:
+                print(f"{OK_MARK} `REQUESTS_CA_BUNDLE` was verified.\n")
+        except OSError as e:
+            print(
+                f"{X_MARK} The following error occured while verifying `REQUESTS_CA_BUNDLE`: {e}\n"
+            )
```
jezdez, line 171: Here I suggest:

- inlining the call to `context.channel_alias`
- using `requests.exceptions.RequestException` for errors that may be thrown when fetching the file (please add that import at the top of the file)
- using the conda session object as suggested above
- storing the requests ca bundle in a variable for reuse
- making use of `response.raise_for_status()` since just checking for `if response` isn't enough, it basically will always be true, but it may be an HTTP 404 response or another HTTP error that we want to bubble up to the user (see [requests docs about it](https://requests.readthedocs.io/en/latest/user/quickstart/#response-status-codes)) to prevent false positives

```suggestion
	# using one of the channel aliases url ensures that the health check also
	# runs on machines that are on an intranet connection. As in such a case
	# the `context.channel_alias.urls()[0]` might be pointing to an internal url.
	ca_bundle_test_url = context.channel_alias.urls()[0]

	requests_ca_bundle = os.getenv("REQUESTS_CA_BUNDLE")
    if not requests_ca_bundle:
        return
    elif not Path(requests_ca_bundle).exists():
        print(
            f"{X_MARK} Env var `REQUESTS_CA_BUNDLE` is pointing to a non existent file.\n"
        )
    else:
        session = get_session(ca_bundle_test_url)
        try:
            response = session.get(ca_bundle_test_url)
            response.raise_for_status()
            print(f"{OK_MARK} `REQUESTS_CA_BUNDLE` was verified.\n")
        except (OSError, RequestException) as e:
            print(
                f"{X_MARK} The following error occured while verifying `REQUESTS_CA_BUNDLE`: {e}\n"
            )
```

jezdez (changes requested):

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
...
+    response.status_code = 200
+    mocker.patch(
+        "conda.gateways.connection.session.CondaSession.get", return_value=response
+    )
+    requests_ca_bundle_check(prefix, verbose=True)
+    captured = capsys.readouterr()
+    assert f"{OK_MARK} `REQUESTS_CA_BUNDLE` was verified.\n" in captured.out
+
+
+def test_requests_ca_bundle_check_action_fail_1(
```
kenodegard, line 269:
```suggestion
def test_requests_ca_bundle_check_non_existent_path(
```

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
...
 
 
+def test_requests_ca_bundle_check_action_passes(
+    env_ok: tuple[Path, str, str, str, str],
+    capsys,
+    monkeypatch,
+    tmp_path: Path,
+    mocker: MockerFixture,
+):
+    from requests import Response
```
kenodegard, line 254: Move this to the imports at the top of this file.

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
...
+
+    prefix, _, _, _, _ = env_ok
+    monkeypatch.setenv("REQUESTS_CA_BUNDLE", str(tmp_path))
+    reset_context()
```
kenodegard, line 258: `reset_context` is not necessary for this test

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
...
+    captured = capsys.readouterr()
+    assert f"{OK_MARK} `REQUESTS_CA_BUNDLE` was verified.\n" in captured.out
+
+
+def test_requests_ca_bundle_check_action_fail_1(
+    env_ok: tuple[Path, str, str, str, str], capsys, monkeypatch
+):
+    prefix, _, _, _, _ = env_ok
+    monkeypatch.setenv("REQUESTS_CA_BUNDLE", "non/existent/path")
+    reset_context()
```
kenodegard, line 274: `reset_context` is not necessary for this test

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
...
+        in captured.out
+    )
+
+
+def test_requests_ca_bundle_check_action_fail_2(
+    env_ok: tuple[Path, str, str, str, str], capsys, monkeypatch, tmp_path: Path
+):
+    prefix, _, _, _, _ = env_ok
+    monkeypatch.setenv("REQUESTS_CA_BUNDLE", str(tmp_path))
+    reset_context()
```
kenodegard, line 288: `reset_context` is not necessary for this test

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -242,6 +244,56 @@ def test_not_env_txt_check_action(
     assert X_MARK in captured.out
 
 
+def test_requests_ca_bundle_check_action_passes(
+    env_ok: tuple[Path, str, str, str, str],
+    capsys,
+    monkeypatch,
```
kenodegard, line 250: Let's use typing for all of these:

```suggestion
    capsys: CaptureFixture,
    monkeypatch: MonkeyPatch,
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -150,8 +151,38 @@ def env_txt_check(prefix: str, verbose: bool) -> None:
         print(f"{X_MARK} The environment is not listed in the environments.txt file.\n")
 
 
+def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
+    ca_bundle_test_url = context.channel_alias.url()[
+        0
+    ]  # using one of the channel aliases url ensures that the health
+    # check also runs on machines that are on an intranet connection.
+    # As in such a case the `context.channel_alias.urls()[0]` might
+    # be pointing to an internal url.
```
kenodegard, line 160:
```suggestion
    # Use a channel aliases url since users may be on an intranet and
    # have customized their conda setup to point to an internal mirror.
    ca_bundle_test_url = context.channel_alias.url()[0]
```

kenodegard (changes requested):

tests/plugins/subcommands/doctor/test_health_checks.py
ForgottenProgramme: why? 🤔

jezdez (approved):
Awesome!

travishathaway (approved):

kenodegard (approved):

Issue #8149, opened 2019-01-27
RemoveError: 'requests' is a dependency of conda and cannot be removed from conda's operating environment


majeedus:
## Current Behavior

Having trouble installing a third party library called bottleneck and I have not seen this error before using Windows 10 with Anaconda installed:

conda install bottleneck

### Environment
Operating System: (Windows 10 Pro)
Python Version: 3.6.5

https://pastebin.com/Qz5tp09J

### Steps to Reproduce

```
(base) C:\WINDOWS\system32>conda install bottleneck
Collecting package metadata: done
Solving environment: done

## Package Plan ##

  environment location: C:\Users\XYZ\Anaconda3

  added / updated specs:
    - bottleneck


The following NEW packages will be INSTALLED:

  bottleneck         pkgs/main/win-64::bottleneck-1.2.1-py36h452e1ab_1
  certifi            pkgs/main/win-64::certifi-2018.11.29-py36_0
  numpy-base         pkgs/main/win-64::numpy-base-1.14.3-py36h555522e_1
  pip                pkgs/main/win-64::pip-18.1-py36_0
  setuptools         pkgs/main/win-64::setuptools-40.6.3-py36_0
  zstd               pkgs/main/win-64::zstd-1.3.7-h508b16e_0

The following packages will be UPDATED:

  ca-certificates                              2018.03.07-0 --> 2018.12.5-0
  libarchive                               3.3.2-h798a506_7 --> 3.3.3-h0643e63_5
  libpng                                  1.6.34-h79bbb47_0 --> 1.6.36-h2a8f88b_0
  openssl                                 1.0.2o-h8ea7d77_0 --> 1.1.1a-he774522_0
  python                                   3.6.5-h0c2934d_0 --> 3.6.8-h9f7ef89_0
  qt                                   5.9.5-vc14he4a7d60_0 --> 5.9.7-vc14h73c81de_0
  sqlite                                  3.23.1-h35aae40_0 --> 3.26.0-he774522_0
  vc                                          14-h0510ff6_3 --> 14.1-h0510ff6_4
  vs2015_runtime                               14.0.25123-3 --> 14.15.26706-h3a45250_0
  wheel                                       0.31.1-py36_0 --> 0.32.3-py36_0

The following packages will be DOWNGRADED:

  libssh2                                  1.8.0-hd619d38_4 --> 1.8.0-h7a1dbc1_4


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: failed

[...]

PR #14051, opened 2024-07-18
Skip spurious RemoveError when conda dependency is provided by pip

### LINKED ISSUE #14050: If a conda dependency is provided by pip, it throws RemoveError on unrelated package installs
### Checklist

### What happened?

I used pip to upgrade pluggy to version 1.5.0 in the base environment. Now when I want to install an unrelated package into base `conda install -n base rapidfuzz` I get `RemoveError: 'pluggy' is a dependency of conda and cannot be removed from
conda's operating environment.`

This is caused by https://github.com/conda/conda/blob/6140dc238a6c639432d92aaea407e3a16cc9e729/conda/core/link.py#L764-L773 which errors because pluggy is also not in `pkg_names_already_linked` thanks to pip.

`conda list` shows
```
pluggy                    1.5.0                    pypi_0    pypi
```

### Conda Info

```shell
% conda info

     active environment : base
    active env location : /Users/dholth/miniconda3
            shell level : 1
       user config file : /Users/dholth/.condarc
 populated config files : /Users/dholth/.condarc
          conda version : 24.5.0
    conda-build version : 24.5.1
         python version : 3.10.13.final.0
                 solver : libmamba (default)
                          __conda=24.5.0=0
                          __osx=14.5=0
                          __unix=0=0
       base environment : /Users/dholth/miniconda3  (writable)
      conda av data dir : /Users/dholth/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                          https://repo.anaconda.com/pkgs/r/osx-arm64
          package cache : /Users/dholth/miniconda3/pkgs
                          /Users/dholth/.conda/pkgs
       envs directories : /Users/dholth/miniconda3/envs
                          /Users/dholth/.conda/envs
               platform : osx-arm64
             user-agent : conda/24.5.0 requests/2.31.0 CPython/3.10.13 Darwin/23.5.0 OSX/14.5 solver/libmamba conda-libmamba-solver/23.11.0 libmambapy/1.5.3 aau/0.4.3 c/3Pd2XMG8Ewd_NqK47OhyjA s/tX-ytR0Amf6BlP8fXYaHwA e/aHwMQRuXzL6N3zBvhz2heg
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```
```


### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by dholth**
### Description

Fix #14050.

When pip upgrades a conda dependency (e.g. `pluggy`), the package disappears from `conda-meta` but remains installed. Conda's `RemoveError` check then blocks *any* unrelated install/remove operation in the base environment because it sees the dependency as missing.

The root cause is in `UnlinkLinkTransaction._verify`: if a conda dependency isn't in `pkg_names_already_lnkd` *and* isn't being linked in the current transaction, it raises `RemoveError` — even though the transaction has nothing to do with that dependency.

This change removes the `pkg_names_already_lnkd` check so `RemoveError` only fires when a conda dependency is actually being *unlinked* by the current transaction, which is the case we actually want to guard against.

A test (`test_dont_remove_conda_3`) reproduces the scenario using `conda-pypi` to convert and install conda into a test environment, then deleting a dependency's `conda-meta` entry to simulate the pip-upgrade situation.

### Checklist - did you ...

### CODE REVIEW

**Review by jezdez — COMMENTED** [review]
This looks like an innocuous change, but I worry about changing stuff like this in the transaction system without a test TBH. @jaimergp do you have more experience in this area by any chance?

**Comment by jaimergp** [comment]
Yea, I'm not sure about this. The fix shouldn't be to ignore that check, but to include the non-conda dependency in the transaction somehow. That said, this change might be innocuous and could be ok, but I'd like a test too.

**Comment by dholth** [comment]
I wasn't able to trigger the error except when I was actually trying to uninstall a conda dependency. I think it may have been overly pessimistic when it was written.

**Comment by dholth** [comment]
Not stale

**Comment by dholth** [comment]
I was able to write a test that [recreates the error condition without the fix](https://github.com/conda/conda/pull/14330/changes#diff-1e27c53a97d6c9242ea92544453665b553fe9fdf63f08717dc915f80e76fd60c)  and passes with the fix. It's involved but it takes advantage of conda-pypi to quickly package up conda and install it into an environment, instead of testing either an editable install or an older version of conda from a channel. I wasn't able to determine why an editable install of conda bypasses the check. Note that the check only applies in the base environment.
We've received a recent report of this issue popping up so I've revived the PR.

File: tests/test_create.py [hunk 1]
```diff
@@ -2354,6 +2355,143 @@ def test_dont_remove_conda_2(
...
 
+def test_dont_remove_conda_3(
+    conda_cli: CondaCLIFixture, tmp_env: TmpEnvFixture, monkeypatch: MonkeyPatch
+):
+    """
+    If conda thinks its core dependency is uninstalled (happens when pip
+    upgrades a dependency) it could produce spurious RemoveError, blocking
+    further use of conda.
+    """
+    with tmp_env("conda", "conda-pypi") as prefix:
```
jezdez, line 2369: Should we pin the version of conda-pypi to a specific one here?

tests/test_create.py
```diff
@@ -2354,6 +2355,143 @@ def test_dont_remove_conda_2(
...
+            meta_path.unlink()
+
+        # reload() to update cached results
+        PrefixData(str(prefix)).reload()
+        print(
+            f"{conda_dependency} installed? {package_is_installed(prefix, conda_dependency)}"
+        )
+
+        # we will likely add this dependency; what's a better small package with
+        # no dependencies?
```
jezdez: We could use any of the local test deps?

tests/test_create.py
dholth: If it broke we'd be more interested in fixing the latest conda-pypi.

dholth: It might be worth skipping this test after this PR has been merged. It's elaborate and unlikely to be reintroduced.

tests/test_create.py
jezdez: done in [`9adce68` (this PR)](https://github.com/conda/conda/pull/14051/commits/9adce687631672dd29e47c5f0df959c6fb38fe5c)

jezdez (approved):

conda/core/link.py
```diff
@@ -829,8 +827,7 @@ def _verify_transaction_level(prefix_setups):
             for conda_dependency in conda_linked_depends:
                 dep_name = MatchSpec(conda_dependency).name
                 if dep_name not in pkg_names_being_lnkd and (
-                    dep_name not in pkg_names_already_lnkd
```
dholth, line 832: It seems like this part anticipates an installation where `_verify_transaction_level()` would be called during an intermediate phase or I suspect they were being too cautious.

dholth:
Fix #8149

PR #14057, opened 2024-07-21
feat: `.toml` support for creating env

### LINKED ISSUE #10633: Support creating environment from pyproject.toml

harnvo:
### Description
Fixes #10633

This is another PR to support `pyproject.toml` 

The code was largely taken from #12666.

My modification includes:
- the environment name would be project's name by default if not given one
- Anyone using this feature must explicitly tell conda to not use any conda channels in `pyproject.toml`:

```
...
[tool.conda]
channel=[]
```

This is to alleviate concerns from https://github.com/conda/conda/pull/12666#issuecomment-1628391662.

- Uses `pyproject.toml` from real-world projects as test cases.

### TODO list

- [ ] Utlize `requires-python` to build environment

I would like to include this feature. However, as specified by [python's documentation](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-requires-python), it's grammar is not compatible with conda. 

Note that in practice most project uses either `>=x.x`, `<=x.x` or `==x.x`, so having support for this feat would not be problematic for most projects. So far I have never seen **any** python projects using grammars not handleable by conda.

- [ ] default to create environment from `pyproject.toml` if conda could not find `environment.yml`
> Not sure if I should really do this. This is likely to break some downstream projects since this is a change of conda's behavior.

- [ ] Update Documents
> Would update the Document once the feature is freezed.

- [ ] Integration Test
> There are some test cases not suitable for unit-test.

- [ ] support installing optional environment via syntactic sugar
> I am think about something like `[test]pyproject.toml`, but I am not sure if it is reasonable to introduce another hack just to enable this feature.

### Checklist - did you ...


harnvo:
@conda-bot check

jaimergp:
I feel this type of support would belong better in a Python-specific plugin like the one we are implementing at [`conda-pypi`](https://github.com/conda-incubator/conda-pypi). We will also need some more plugin hooks in conda and conda env to allow additional file formats.

harnvo:
> I feel this type of support would belong better in a Python-specific plugin like the one we are implementing at [`conda-pypi`](https://github.com/conda-incubator/conda-pypi). 

Not sure if I have understood everything correctly. "Having support for `pyproject.toml`" means we could configure & manage our project with `conda` + `pyproject.toml`. This is something cannot be done soloy with better support for `PyPI`.

IMO, having another `FileSpec` is mandatory for this feature, and it is just the matter of how. Maybe you meant a plugin that makes PEP 508 syntax compatible with conda?

> We will also need some more plugin hooks in conda and conda env to allow additional file formats.

I would agree having more plugin hooks is a less hacky way for implementing this feature. However, I believe the biggest problem lies in the fact that `conda` assumes one `file` contains exactly one environment, which is no longer true for `pyproject.toml`.

travishathaway:
@conda-bot check

jaimergp:
These days you can install the `conda-pypi` alpha and do `conda pip install .` on your PyPI project.

One key aspect to remember is that `conda` is a language-agnostic package manager. Python has been special cased in the past but in the future we'd rather have language-specific plugins dealing with the oddities of each ecosystem. We are slowly adding more and more plugin hooks to make this possible.

I don't have a clear idea of what you are trying to achieve when you say "Add support for pyproject.toml". I can see several possible strategies:

- Install a project from their git repository, installing Python and pip with conda, and the rest from PyPI
- Create an environment using _both_ pyproject.toml and environment.yml?
- Create an environment using only pyproject.toml, which would have a new `tool.conda` section with the conda deps?

I'm not trying to be dismissive, but I am more interested in seeing this implemented at the right level, with the right scope, and maybe as a plugin (which could be eventually be included by default as a conda dependency; being a plugin doesn't have to mean it's not part of the "core" experience). I appreciate the prototypes because the inform the conversation, I think we would benefit from discussing the scope back in the originating issue.

matterhorn103:
Aside from the other concerns raised, I feel obliged to point out that the consensus on the python.org forums e.g. in [this thread](https://discuss.python.org/t/projects-that-arent-meant-to-generate-a-wheel-and-pyproject-toml/29684) or [this one](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233) is that the `project.dependencies` table in `pyproject.toml` should *not* be used for any purposes other than for building a package for distribution (via PyPI).

I.e. PyPI packages must specify their own dependencies there, as should things being *built*, but it *should not be used to provide arbitrary lists of dependencies* required to be installed in an environment to run or work on some project.

[PEP-735](https://peps.python.org/pep-0735/) is being worked on to provide a standardized way for precisely this, and once it's passed, maybe Conda could support that in some form.

But while I don't know enough about packaging for Conda distribution to know how Conda *packages* specify their own dependencies (perhaps it's also in `environment.yml`), I *do* know that `environment.yml` is often used for projects that are not being either installed or packaged for distribution – the exact use-case which is being explicitly discouraged from utilizing `project.dependencies`.

So for now I think it is a really bad idea to try and make Conda support dependency specification within `pyproject.toml` *unless* the supported place to specify the dependencies is contained entirely within a `[tools.conda]` table.

Hence [my suggestion](https://github.com/conda/conda/issues/10633#issuecomment-2222670268) in the issue you linked as to how `[tools.conda.environment]` could be parsed as if it was an `environment.yml` file.

dholth:
I disagree with the forums; why not allow such a list to be directly installed? Deliberately unhelpful.

We could have a general facility for installing a list of requirements given an arbitrary key within thing-that-can-be-parsed-as-a-dict (toml, json, yaml). "install <file.toml> --from-key tool.conda.requirements or --from-key forbidden-by-forums.requirements"

matterhorn103:
Yeah I know, it does seem a bit pointlessly stubborn. But those discussions are settled it seems, and surely it would also be pretty deliberately unhelpful to move now, while a standardization effort is ongoing, to actively support the one way in which such lists of dependencies definitely *won't* be included in a `pyproject.toml`. Doesn't hurt to wait for the outcome.

harnvo:
> We could have a general facility for installing a list of requirements given an arbitrary key within thing-that-can-be-parsed-as-a-dict (toml, json, yaml). "install <file.toml> --from-key tool.conda.requirements or --from-key forbidden-by-forums.requirements"

I agree with this general idea. However, this does not solve his concern IMO. 

`project.dependencies` are used by setuptools. Allowing conda to create environment from `project.dependencies` without channel restriction could make conda-managed projects uninstallable via setuptools.

Issue #12462, opened 2023-03-08
Support for `pyproject.toml`


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

`pyproject.toml` and build systems are all the rage in the Python world today. conda needs to figure out how it fits into this world.

### Linked Issues & PRs

- [ ] https://github.com/conda/conda/issues/10633
- [ ] https://github.com/conda/conda/issues/12126
- [ ] https://github.com/conda/conda-build/issues/4774
- [ ] https://github.com/conda/conda-build/issues/4002
- [ ] https://github.com/conda/conda-build/issues/3919
- [ ] https://github.com/conda/conda-build/issues/3563

wvreeven:
Is there any update on this? I see that is has been removed from the planning. Does this mean that this will not be implemented? Any information would be very welcome.

kenodegard:
@wvreeven no progress on this yet

The move from the Planning board to Epics was just us reorganizing tickets (and GitHub being surprisingly noisy about it), no deeper meaning.

xuanswe:
Hi, do you have an idea about the time when this feature could be done?

shifenhutu:
i use https://github.com/astral-sh/uv now, it's better

PR #14067, opened 2024-07-23
Add spec and tests to enable reading environment table from pyproject.toml

### LINKED ISSUE #10633: Support creating environment from pyproject.toml

matterhorn103:
### Description

Added a spec to demonstrate how conda could enable the creation of a conda environment from a `pyproject.toml` file, if the user desires. This has been quite a popular request, see #12462 

Closes #10633 

The approach is that I put forward in #10633:

- The contents of an ordinary conda `environment.yml` can alternatively be included in a `pyproject.toml` file under a `[tool.conda.environment]` table.
- The support is read-only (a conda environment cannot be exported to a pyproject.toml)
- The table has to be formatted with TOML syntax, naturally, so it must be translated from YAML syntax, but this is trivial
- The syntax for requirement strings is identical to a normal conda environment file
- The parsed table is passed directly to the normal YAML parser, so anything that works in a normal `environment.yml` works here too

Unlike previous approaches (#12666 and #14057), no premature judgement or decision is made on how to process dependencies specified in other tables of the TOML, as the officially correct(/at all acceptable) place to store arbitrary lists of dependencies for an environment in a `pyproject.toml` is [yet to be finalized](https://peps.python.org/pep-0735/), and as I understand it there are remaining technical challenges when it comes to passing PEP-440 (PyPI-style) dependency specifications. Such functionality could be added at a later date, when an appropriate approach crystallizes. For now, this solution provides a stop-gap.

As such, this would not allow users to use a single dependency list for both PyPI tools and conda tools, but it would allow consolidation into a single file that can support two sets of users.

### Checklist - did you ...

### DIFF
--- a/conda/env/specs/__init__.py
+++ b/conda/env/specs/__init__.py
@@ -13,6 +13,7 @@ from ...exceptions import (
 from ...gateways.connection.session import CONDA_SESSION_SCHEMES
 from .binstar import BinstarSpec
 from .requirements import RequirementsSpec
+from .pyproject import PyProjectSpec
 from .yaml_file import YamlFileSpec
 
 FileSpecTypes = Union[Type[YamlFileSpec], Type[RequirementsSpec]]
@@ -25,7 +26,9 @@ def get_spec_class_from_file(filename: str) -> FileSpecTypes:
     :raises EnvironmentFileExtensionNotValid | EnvironmentFileNotFound:
     """
     # Check extensions
-    all_valid_exts = YamlFileSpec.extensions.union(RequirementsSpec.extensions)
+    all_valid_exts = set().union(
+        YamlFileSpec.extensions, RequirementsSpec.extensions, PyProjectSpec.extensions
+    )
     _, ext = os.path.splitext(filename)
 
     # First check if file exists and test the known valid extension for specs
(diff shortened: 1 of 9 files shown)


matterhorn103:
Could anyone point me in the direction of a guide on how to build conda itself?

matterhorn103:
Would it be possible for someone to merge https://github.com/conda/infrastructure/pull/968?

travishathaway:
@conda-bot check

matterhorn103:
pre-commit.ci autofix

matterhorn103:
@conda-bot check

travishathaway:
Hi @matterhorn103,

Thanks a lot for your work on this issue. I wanted to give you some feedback on this pull request and this issue in general.

First, the pull request itself looks reasonable, but without being able to support all versions of Python that our users have installed (< 3.11), this would provide and an inconsistent experience, and in my opinion make this not eligible to be merged. We are also very conservative about adding new dependencies, so that route may be difficult too.

Second, the maintainers are currently moving to push PyPI specific behavior outside the main conda project (see: https://github.com/conda-incubator/conda-pypi). Merging a pull request like this would run contrary to these efforts.

In order to have this feature be handled by external plugin, we need to add the ability to add new environment specifications as a plugin hook. If you would like to help with this, I would suggest engaging with the maintainers on https://conda.zulipchat.com/ (I'm there too :smile:).

The cool thing is that once we have this plugin hook setup, you should be able to copy and paste a lot of the work you did here to make it work.

PR #14065, opened 2024-07-23
Revert #13975

jaimergp:
### Description

Reverts #13975 after concerns raised in https://github.com/conda/conda/issues/12558#issuecomment-2245517955.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Issue #12558, opened 2023-03-31
Windows: conda cannot activate environments with `!` in the name


marcoesters:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

On Windows, [constructor](https://github.com/conda/constructor) disallows installing `conda` into paths that contain the special characters `^%!=,`: https://github.com/conda/constructor/blob/main/constructor/nsis/main.nsi.tmpl#L854

So, I got curious and tried to create conda environment with these:

```
conda create -n foo^%,!=
```

It created the environment `foo,!=`

```
# conda environments:                                                                                                                              
#   

base                  *  C:\miniconda3                                                                                                             
foo,!=                   C:\miniconda3\envs\foo,!= 
```

I verified that the directory exists. When I tried to activate this environment using `conda activate foo,!=`, I get the following error:

```                                                                                                                                                                                                     
EnvironmentNameNotFound: Could not find conda environment: foo,=                                                                                   
You can list all discoverable environments with `conda info --envs`.    
```

It seems like `!` is a problematic character, so `conda` should maybe not allow environments with that character on Windows. Maybe the list of forbidden characters should be synchronized with `constructor`, too?

### Conda Info

```shell
active environment : base                                                                                                                     
    active env location : C:\miniconda3                                                                                                            
            shell level : 1                                                                                                                        
       user config file : C:\Users\dev-admin\.condarc                                                                                              
 populated config files : C:\miniconda3\.condarc                                                                                                   
                          C:\Users\dev-admin\.condarc                                                                                              
          conda version : 23.3.1                                                                                                                   
    conda-build version : 3.24.0                                                                                                                   
         python version : 3.9.16.final.0                                                                                                           
       virtual packages : __archspec=1=x86_64                                                                                                      
                          __win=0=0                                                                                                                
       base environment : C:\miniconda3  (writable)                                                                                                
      conda av data dir : C:\miniconda3\etc\conda                                                                                                  
  conda av metadata url : None                                                                                                                     
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64                                                                               
                          https://repo.anaconda.com/pkgs/main/noarch                                                                               
                          https://repo.anaconda.com/pkgs/r/win-64                                                                                  
                          https://repo.anaconda.com/pkgs/r/noarch                                                                                  
                          https://repo.anaconda.com/pkgs/msys2/win-64                                                                              
[...]

PR #14073, opened 2024-07-24
[DRAFT] Warn for problematic chars in env name on Windows

mbargull:
### Description

This is a suggestion to handle @jezdez's comment from https://github.com/conda/conda/issues/12558#issuecomment-2245951439 :
> I’d suggest to work on an improved patch that only presents users with a note that future versions of conda will change env name handling.

This is only composed quickly and is untested, carries no news file, etc.
Since I'm not able to work on this further, please do feel free to amend or incorporate into gh-14065 as you see fit.

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -185,11 +185,7 @@ PREFIX_NAME_DISALLOWED_CHARS = {
     "#",
 }
 # Includes disallowed characters from constructor
-PREFIX_NAME_DISALLOWED_CHARS_WIN = {
-    "/",
-    " ",
-    ":",
-    "#",
+PREFIX_NAME_DISCOURAGED_CHARS_WIN = {
     "^",
     "%",
     "!",
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -13,6 +13,7 @@ import os
 import platform
 import struct
 import sys
+import warnings
 from collections import defaultdict
 from contextlib import contextmanager
 from errno import ENOENT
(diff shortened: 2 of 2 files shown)


conda/base/constants.py
```diff
@@ -185,11 +185,7 @@
     "#",
 }
 # Includes disallowed characters from constructor
-PREFIX_NAME_DISALLOWED_CHARS_WIN = {
-    "/",
-    " ",
-    ":",
-    "#",
+PREFIX_NAME_DISCOURAGED_CHARS_WIN = {
```
mbargull, line 188: It could even make sense to not add this symbol here or at least prefix it and underscore since it its TBD how this will be handled.

conda/base/context.py
```diff
@@ -2068,25 +2069,30 @@ def validate_prefix_name(prefix_name: str, ctx: Context, allow_base=True) -> str
...
+            disallowed=PREFIX_NAME_DISALLOWED_CHARS,
+        )
+    if on_win and PREFIX_NAME_DISCOURAGED_CHARS_WIN.intersection(prefix_name):
+        warnings.warn(
+            dals(
+                f"""
+                Environment name {prefix_name!r}
+                contains characters from {PREFIX_NAME_DISCOURAGED_CHARS_WIN}
+                which are known to cause issues.
+                Future conda versions may disallow them in environment names.
```
mbargull, line 2092: This is intentionally worded vaguely because the whole process on figuring out which/how many of those characters should be discouraged and when to implement it is still TBD.

PR #14093, opened 2024-07-30
Deprecate `tests.env.utils.make_temp_envs_dir`

kenodegard:
### Description

Implement a new fixture, `tmp_envs_dir`, mirroring the `tmp_pkgs_dir` fixture to replace `make_temp_envs_dir` which required additional manual setup.

### Checklist - did you ...

### DIFF
--- a/tests/env/test_create.py
+++ b/tests/env/test_create.py
@@ -15,7 +15,6 @@ from conda.testing import CondaCLIFixture, PathFactoryFixture
 from conda.testing.integration import package_is_installed
 
 from . import support_file
-from .utils import make_temp_envs_dir
 
 
 def get_env_vars(prefix):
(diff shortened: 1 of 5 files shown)


tests/env/test_create.py
```diff
@@ -27,229 +26,218 @@ def get_env_vars(prefix):
...
+def test_create_update(
+    conda_cli: CondaCLIFixture,
+    monkeypatch: MonkeyPatch,
+    tmp_envs_dir: Path,
+):
+    env_name = uuid4().hex[:8]
+    prefix = tmp_envs_dir / env_name
 
+    conda_cli(
+        *("env", "create"),
```
travishathaway, line 38: I understand this syntax was here before, but what's the point of doing this? Isn't it basically the same as doing:

```python
conda_cli(
    "env", "create",
    "--name", env_name,
    "--file", support_file("example/environment_pinned.yml")
)
```

travishathaway (approved):
Looks good. I had one question that was wasn't related to the PR though.

tests/env/test_create.py
kenodegard: it was done this way since ruff/black would reformat it:

```python
conda_cli(
    "env",
    "create",
    "--name",
    env_name,
    "--file",
    support_file("example/environment_pinned.yml"),
)
```

which is harder to read since it is unclear that the arguments belong to the flag

this style was adopted and used before I realized that `argparse` supports `=`, so the preferred style is now:

```python
conda_cli(
    "env",
    "create",
    f"--name={env_name}",
    f"--file={support_file('example/environment_pinned.yml')}",
)
```

but I deemed doing that here was out of scope

Issue #14095, opened 2024-07-31
Replace `env_var`/`env_vars` with `monkeypatch.setenv`


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

The `env_var` and `env_vars` with-statement contexts are better written using `monkeypatch.setenv` (less code nesting and overall easier to read tests).

We cannot outright deprecate either `env_var` or `env_vars` since they're used outside of the test suite but removing them from the test suite will give us a better idea of the next steps we can take.

### Tasks
- [X] #14096
- [X] #14097 
- [X] #14098
- [X] https://github.com/conda/conda/pull/14105
- [X] https://github.com/conda/conda/pull/14106
- [X] https://github.com/conda/conda/pull/14165
- [X] https://github.com/conda/conda/pull/14166

jaimergp:
I think we only have one instance left, @agriyakhetarpal! Awesome work:

https://github.com/conda/conda/blob/a67cb54a24fa2665a12b205243136252508a79c9/conda/testing/fixtures.py#L130-L139

After that, shall we deprecate `env_var` and `env_vars`, @kenodegard?

agriyakhetarpal:
Actually, besides the fixture you suggested, there are a few more left:

https://github.com/conda/conda/blob/a67cb54a24fa2665a12b205243136252508a79c9/tests/core/test_initialize.py#L631-L638

https://github.com/conda/conda/blob/a67cb54a24fa2665a12b205243136252508a79c9/tests/core/test_initialize.py#L1052-L1062

https://github.com/conda/conda/blob/a67cb54a24fa2665a12b205243136252508a79c9/conda/testing/helpers.py#L293-L305

I'll pick them all up!

agriyakhetarpal:
I'll note that we have a few internal uses of the `env_vars` contexts, but we've dropped all uses in the test suite. We should now be in a position to start the deprecation process for it (for v25.11.0 or later).

PR #14105, opened 2024-08-01
`test_connection.py`: Replace `env_var` with `monkeypatch.setenv`

kenodegard:
### Description

Replace `env_var` usage in `tests/gateways/test_connection.py` with `monkeypatch.setenv`.

Xref #14095

### Checklist - did you ...


ForgottenProgramme (approved):
Why I approved this PR:
You wanted to use `monkeypatch` in `inner_s3_test` , and since it is being called inside `test_s3_server_with_mock` and `test_s3_server`, you updated their signatures also to contain `monkeypatch`. 
You are now using `monkeypatch.setenv` instead of `with env_vars():`. And `monkeypatch.setattr` instead of `patch.object`.
You wrote assert statements after calling `monkeypatch.setenv` to confirm that those env vars were set correctly.

PR #14113, opened 2024-08-01
Add support for YML files in `conda create --file` (WIP)

jaimergp:
### Description

Looking into parts of https://github.com/conda/conda/issues/11633 (`conda env create` -> `conda create`).

This is the minimum effort required to have YML support in `conda create`. It has a few problems UX-wise, though:

- There's no yes/no prompt
- The reported information is different (terser)
- We can't get the --name from the environment.yml file
- It doesn't support the pseudo-plugin system

This is just here to see what breaks and inform of the challenges in `conda.cli.install` to make a case for a bigger refactor. 

<details>

<summary>What I would like to have instead</summary>

I envision a new `Environment` class that CLIs need to fulfill with details such as packages, channels or solver settings, and then this class delegates to the adequate install backends (solver, explicit...). This starts to build up a homogeneized Environment schema that could look like a potential `conda.toml` or `environment.yml` v2. That's a big change so I'll draft a roadmap to get there if interested. My idea would be to have a single file that represents the input state of the environment, and then operating on the environment would mean editing that file (on disk or virtually), and then applying the changes to disk. Again, borrowing a few concepts from the Pixi model.

</details>

Edit: I went ahead and implemented parts of the `details` dropdown. See this comment: https://github.com/conda/conda/pull/14113#issuecomment-2299327969

### Checklist - did you ...


jaimergp:
> I expanded a little bit on that `details` tag above.

---

# conda environment modification UX

There are four main commands:

- `conda create`: Creates a new environment (identified by name or full path). Accepts a list of MatchSpec strings and/or a list of TXT files containing MatchSpecs. Paths and URLs are also accepted.
- `conda install`: Same as create, but it expects an _existing_ environment.
- `conda remove`: Removes package(s) from the environment, and their dependents. Can also remove the full environment if `--all` is passed.
- `conda update`: Installs the latest version of a package already installed, as long as it's solvable.

All of these can be tackled as the same type of input if expressed as a single input file. Let's assume that file is `environment.yml` for practical purposes.

- `conda create -n env -c conda-forge python=3.10`:

```yaml
name: env
channels:
- conda-forge
dependencies:
- python=3.10
```

- `conda install -n env numpy` just means adding `numpy` to the implicit file:

```yaml
name: env
channels:
- conda-forge
dependencies:
- python=3.10
- numpy=*  #
```

- `conda update -n python` means unpinning `python`, solving, and then repinning to the resolved version. Note this command only accepts package _names_. 

```yaml
name: env
channels:
- conda-forge
dependencies:
- python=*  # would become python=3.12 
- numpy=*
```

- `conda remove numpy` is as obvious as it sounds; remove its entry and resolve:

```yaml
name: env
channels:
- conda-forge
dependencies:
- python=*  # would become python=3.12
```

However `conda remove` can also remove dependencies not _explicitly_  in the input file, but maybe transitory dependencies that are part of the one of the variants. For example, `mkl` BLAS for `numpy`. We can `conda remove mkl` and we would obtain `openblas` instead, while still keeping numpy. We would need an extra input field to _prevent_ a package from being installed, similar to `run_constrained` in `meta.yaml`. This can operate as a series of pins too!

```yaml
name: env
channels:
- conda-forge
dependencies:
- python=*  # would become python=3.12
constraints:
- mkl <0 # removes or prevents installation of mkl
```

## `--file` flags

The `--file` flag (which can be specified multiple times) accepts TXT files. These TXT files have newline-separated MatchSpecs or URLs. `#`-leading lines are ignored. If `@EXPLICIT` is in the file, it's considered an explicit install with no solver invocation: the URLs are just fetched and linked in order of appearance.

Since they only specify specs, they can be concatenated easily. They are meant to complement the CLI, not replace it!

`environment.yml` files are _not_ supported by this option (yet). This format has more details whose concatenation is not as obvious: what to do with `name` or `channels`? Who wins? First? Last? These would need to be solved before added to the new input file format. As a result we would start only accepting one of these at a time.

## Solving-affecting flags

There are also a number of solver flags that can affect the result of the solve:

- Postprocess dependencies of passed specs _after_ the solve. These can only be implemented under the "everything is a file" model by post-processing the hypothetical lockfile that gets created. Note this leaves the environment in an inconsistent state.
    - `--no-deps`: only the packages corresponding to the spec are installed.
    - `--only-deps`: everything _but_ the passed packages is installed.

- Adjust how dependencies are dealt with. These flags are useful to solve complications of dealing with the "installed packages should not change unless necessary" behaviour that `conda` prefers. In my opinion, these could be implemented as a series of `constraints` for `freeze-installed`, or as a series of `spec=*` in `dependencies`:
    - `--specs-satisfied-skip-solve`: if the installed stuff satisfies the constraints, do not update even if there are newer versions compatible.
    - `--freeze-installed`: constrain everything else while solving the new specs.
    - `--update-deps`: force the update of the dependencies of the specs we passed (consider it a partial update-all).
    - `--update-specs`: this is the default behaviour when not freezing.
    - `--update-all`: this is in principle the default behaviour of a fresh input file. So it's a matter of resolving it again with the new repodata.

There are also some extra CLI flags that concern how the channels are fetched and can have an effect on the solution:

- `--repodata-fn`: which repodata file(s) are fetched from the remote channels(s). This _can_ affect the solution.
- `--no/strict-channel-priority`: how to deal with several channels at once.
- from condarc, `only_tar_bz2`. This could be deprecated.

And of course, the `--solver` flag.

## All other flags

Everything else in the CLI should be considered a runtime option that does not affect the solution of the environment, and hence could be just kept around when necessary. For example, `--copy` can be used without issues in the CLI and its presence won't affect which packages are installed.

## The proposed schema

This is the proposed schema for a more explicit input file that can potentially replace the state stored in `conda-meta/history`, `conda-meta/state` and `conda-meta/pinned`.

```yaml
name: str
description: str
last_modified: datetime
channels: list of str  # these should be ideally URLs for fully resolved channels
channel_options:
  repodata_fn: list of str
  # maybe authentication stuff
platforms: list of str
solver_options: dict
  solver: str
  channel_priority: flexible or strict
  use_only_tar_bz2: bool
  aggressive_update_packages: list of str
dependencies: list of str or dict of (str, list(str))
constraints: list of str  # conda-meta/pinned
variables: dict of (str, str)  # conda-meta/state
```

travishathaway:
@jaimergp,

I'm curious why you see the need to eventually support multiple `--file` options. Wouldn't this just needlessly overcomplicate our implementation and the CLI interface? Under what circumstances would this be useful?

Another thing that we could do to make the CLI even more simplified would be to either accept a `--file` option or list of MatchSpecs on the command line. This would also help further simplify our implementation.

jezdez:
To support @travishathaway, I've mentioned this to @jaimergp in person already, I also think multiple `--file` invocations with `.yml` file are a nice-to-have since it would imply resolving a merging strategy first. We can take a look at how micromamba handles it of course, but that's not just a refactor then anymore, but a larger feature addition. Could we achieve the end goals of this ticket basically in multiple step, to reduce the code churn? Deprecation of existing flags and code paths need to be accounted for as well.

jaimergp:
Hello @conda/conda-core! This PR is still in draft but we have reached a milestone here. `conda create --file` passes the `conda env create -f` tests :) 

Let me recap what I've done here:

- Created a `conda.cli.install2` module that reimplements parts of `conda.cli.install`.
- Added a new `Environment` class that is able to accumulate much of the input data necessary to operate on an environment. It has a weird scope overlap with `PrefixData` but it's more CLI-ish, if that makes sense. Still unsure if this would be an implementation detail or a first-class API citizen (maybe encapsulating the logic of a potential environment.yml v2 file format).
  - The `merge` classmethod allows us to combine `Environment` objects, regardless the source (CLI, txt, yml...). This is so we can support multiple TXT files, but it also handles multiple YMLs if you want. The CLI data is massaged into an `Environment` object too, so this makes it super easy to combine all the possible sources. 
  - Fun fact: This is mostly so we can deal with the annoying feature of `environment.yml` being able to provide an env name / path, and then have it overridden in the CLI (sometimes). I don't like this but we need it for a smooth transition I guess.
- I've also split the main `install` function in smaller ones, and rewrote some of them so we can get `Transaction` objects out of them. This allows us to have the same UX across input sources. 
  - Note that `conda env create` never asked for confirmation or reported the summary of the transaction. Same with explicit files. So, technically, the correct translation for `conda env create -f some.yml` is `conda create --file some.yml --yes`.

There's still some more work to do (like a smarter file format detector that doesn't have to deal with the now deprecated Anaconda.org environments), but you get the idea.

jaimergp:
pre-commit.ci autofix

kcpevey:
Excited to see this! I need to specify my optional test/docs dependencies in a separate file from my core dependencies. It would be great to be able to use the yml format instead of the txt format for specifying multiple files 💜

jezdez:
@jaimergp @soapy1 Where are we with this, is this PR still needed?

jaimergp:
Nah this is still open mostly for historic reasons, but I think @soapy1's work supersedes this first exploration. Let's close so it's clear this won't get anywhere. It served its purpose 💐

PR #14117, opened 2024-08-02
Adapt for logging lock internal changes in Python 3.13

musicinmybrain:
### Description

In Fedora Rawhide (the development version), the system Python is now 3.13.0rc1. We are therefore testing conda (and everything else) on Python 3.13 regardless of whether “official” support is ready or not.

This PR fixes errors like:

```
+ conda info
  […]
  File "/builddir/build/BUILD/conda-24.7.1-build/BUILDROOT/usr/lib/python3.13/site-packages/conda/common/io.py", line 272, in _logger_lock
    logging._acquireLock()
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'logging' has no attribute '_acquireLock'
```

Currently in Python 3.13, the `logging` functions `_prepareFork` and `_afterFork` correspond closely to the old `_acquireLock` and `_releaseLock`, which have been removed.

As long as conda is already using non-public APIs, it’s easiest to just call *those* on Python 3.13. That approach could perhaps break in the future if these functions start doing something other than acquire/release a module lock, but copying their contents could *also* break if the internal locking semantics change in a future Python, so :shrug:.

(See also https://github.com/celery/billiard/pull/404.)

### Checklist - did you ...

- [X] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes? **I didn’t consider this “significant” enough for an entry, but let me know if I need to add one.**
     
- [x] Add / update necessary tests? **Covered by existing tests, when executed on Python 3.13**
- [x] Add / update outdated documentation? **Nothing to do.**

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -269,11 +269,16 @@ def argv(args_list):
 
 @contextmanager
 def _logger_lock():
-    logging._acquireLock()
+    try:
+        # Python 3.13+
+        acquire, release = logging._prepareFork, logging._afterFork
+    except AttributeError:
+        acquire, release = logging._acquireLock, logging._releaseLock
+    acquire()
     try:
         yield
     finally:
-        logging._releaseLock()
+        release()
 
 
 @contextmanager


musicinmybrain:
CLA signed…

travishathaway:
@conda-bot check

jezdez:
We don't currently run the tests on Python 3.13, so it's a little hard to see if this has a negative impact on runtime behavior etc

travishathaway:
@musicinmybrain,

Can you please add a news file for this pull request. You just need to put a bullet point under "Other". We recently had to add a change for xonsh regarding a deprecated import path, so that would be a good example to look at:

- https://github.com/conda/conda/blob/main/news/14047-update-xonsh-support

Please use the pull request number in the title and at the end of the line of the bullet point.

musicinmybrain:
Thanks for the follow-up. I added a news item.

SheMelody:
Could we hurry up? This is preventing me from using conda in my environment.

bdice:
Hi, what is the next step for this fix? I just ran into this while attempting to help conda-forge with the Python 3.13 migration for the `conda` package. Could this be reviewed/merged?

xref: https://github.com/conda-forge/conda-feedstock/issues/246#issuecomment-2629559879

jakirkham (approved):
Thanks Ben! 🙏

FWIW LGTM

In terms of Python 3.13 CI, filed a separate issue to track adding that: https://github.com/conda/conda/issues/14554

PR #14131, opened 2024-08-05
Refactor `classic` solver into its own subpackage

jaimergp:
### Description

Consider this a preview of what a `conda-classic-solver` package would look like. I still need to add the deprecation forwarders in the original places, but I'll get to do that once we establish wether we want to keep the solver here, or in its own subpackage that it's built separately.


jezdez (commented):
I agree we should move this into a `conda-pycosat-solver` package

jaimergp:
Superseded by https://github.com/conda/conda/pull/14170

PR #14150, opened 2024-08-13
Reporter backends refactor

travishathaway:
### Description

This refactor was prompted by the problems and solutions I initially posted in this document:

- https://hackmd.io/-zqp1n2HQX2QIQUy-MEnng

### Checklist - did you ...


conda/reporters.py
```diff
@@ -6,27 +6,62 @@
...
+    of output.
+    """
+    if context.json:
+        backend = context.reporter_backends.json.get(
+            "backend", DEFAULT_JSON_REPORTER_BACKEND
+        )
+    else:
+        backend = context.reporter_backends.console.get(
+            "backend", DEFAULT_CONSOLE_REPORTER_BACKEND
+        )
```
jezdez, line 37: To be clear, `console` will remain the fallback/default reporter backend, basically?

conda/base/constants.py
```diff
@@ -266,6 +266,12 @@ class SatSolverChoice(ValueEnum):
 DEFAULT_SOLVER = "libmamba"
 CLASSIC_SOLVER = "classic"
 
+#: The name of the default json reporter backend
+DEFAULT_JSON_REPORTER_BACKEND = "json"
+
+#: The name of the default console reporter backend
+DEFAULT_CONSOLE_REPORTER_BACKEND = "classic"
```
jezdez, line 273: I don't think "classic" fits here as a name since that would imply it would not be offered anymore like the classic solver, which is a concrete class etc, while the default console rendering is just built-in.

How about this?

```suggestion
DEFAULT_CONSOLE_REPORTER_BACKEND = "default"
```

jezdez (changes requested):
Ultimately this is a step back, but I understand your concerns about a meta configuration system as described in your hackmd document but I think the proposed configuration with mappings of mappings is unprecendented in condarc.

1. Since we already have a settings plugin hook, we don't have to nest the config option you propose, so instead of 

```yaml
reporter_backends:
  console:
    backend: rich
```

```yml
reporter_backends:
  console: rich
```

should suffice.

2. In fact, the "reporter backends" nomenclature seems a little odd now that there aren't any outputs anymore, it feels a little heavy-handed. How about just "outputs" now? I think that fits well with CLI jargon.

```yml
outputs:
  console: rich
```

3. If the `rich` output plugin requires additional settings, it could use the [settings plugin hook](https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/settings.html) to define additional config options, so users can override them:

```yml
outputs:
  console: rich
plugins:
  rich:
    info: dim cyan
    warning: magenta
    danger: bold red
```


So I'd suggest to:

- rename "reporter backends" to just "outputs"
- rely on the settings hook for custom config options for output plugins
- not look back

kenodegard:
If we simplify it as @jezdez suggests above we reduce this to a 1-1 mapping from settings to any given reporter backend and prevents the users from defining more than one backend "flavor" at a time (e.g., the user cannot define a b/w rich output and a colored rich output). Is this desired?

If I understand correctly, going the above route would also mean the `reporter_backends`/`outputs` mapping is simply an alias mapping which seems unnecessary. Wouldn't it just be easier to just refer to it as the `rich` reporter instead of the `console` reporter?

travishathaway:
@jezdez,

I like what you have proposed as it helps simplify the configuration even further. Maybe this could be even further simplified by using the setting `console` instead though:

```
console: rich
plugins:
  rich:
    info: blue
    warning: cyan
```

If we do this, we enable easily switching the reporter backend via a command line option (e.g. `--console=rich`).

Funny enough, this resembles my initial proposal for this feature 🙃...

@kenodegard, I think that proposal addresses what you were talking about, right?

travishathaway:
I went ahead and added the `--console` option to conda too. This will allow us to specify which console backend we want to use at the CLI, which could be useful. For example:

```bash
conda info --console=rich
```

conda/reporters.py
```diff
@@ -6,27 +6,60 @@
...
+            DEFAULT_CONSOLE_REPORTER_BACKEND
+        )
 
-        if reporter is None or output is None:
-            continue
+        if reporter is None:
+            raise CondaError(
+                "There are no available reporter backends to render output. This conda installation"
+                " is most likely corrupt and requires a re-installation."
+            )
```
kenodegard, line 52: What about moving both of these checks into `get_reporter_backend` instead and simplify this into:

```python
reporter = context.plugin_manager.get_reporter_backend(backend, default=DEFAULT_CONSOLE_REPORTER_BACKEND)
```

conda/reporters.py
```diff
@@ -6,27 +6,60 @@
...
-            if render_func is None:
-                raise AttributeError(f"'{style}' is not a valid reporter backend style")
-        else:
-            render_func = getattr(renderer, "render")
+    if style is not None:
+        render_func = getattr(renderer, style, None)
+        if render_func is None:
+            raise AttributeError(f"'{style}' is not a valid reporter backend style")
+    else:
+        render_func = getattr(renderer, "render")
```
kenodegard, line 61: It would seem that we intend to call this `render` in place of most `print` statements in the code, correct? If that's the case, both fetching the `reporter` backend and detecting the `render_func` (with/without the style) should be performed in a cached function call.

conda/base/context.py
```diff
@@ -420,7 +415,10 @@ class Context(Configuration):
     dry_run = ParameterLoader(PrimitiveParameter(False))
     error_upload_url = ParameterLoader(PrimitiveParameter(ERROR_UPLOAD_URL))
     force = ParameterLoader(PrimitiveParameter(False))
-    json = ParameterLoader(PrimitiveParameter(False))
+    json = ParameterLoader(PrimitiveParameter(False, element_type=(bool, str)))
```
kenodegard, line 418: I don't think this is necessary and probably overcomplicates json output. The user should instead just say `--console=alt-json`.

travishathaway: I like this. Turning the the CLI option of `--json` to either be a boolean on/off or a option that accepts strings is going to be a pain! We would have to implement something custom ourselves to handle it.

conda/reporters.py
travishathaway: Good idea. I'm going to move the retrieval of the render function into a separate cached function.

conda/env/specs/binstar.py
```diff
@@ -17,9 +18,12 @@
 
     from ...env.env import Environment
 
-ENVIRONMENT_TYPE = "env"
+deprecated.module("24.7", "25.9")
 
+deprecated.constant("24.7", "25.9", "ENVIRONMENT_TYPE", "env")
```
jezdez, line 23: I'm not sure I follow why this change shows up here, did [`483f36a` (#14150)](https://github.com/conda/conda/pull/14150/commits/483f36a1609fa91dff920478d11e5884b0c09a96) leak into this PR?

conda/plugins/manager.py
```diff
@@ -366,18 +360,33 @@ def get_virtual_packages(self) -> tuple[CondaVirtualPackage, ...]:
     def get_reporter_backends(self) -> tuple[CondaReporterBackend, ...]:
         return tuple(self.get_hook_results("reporter_backends"))
 
-    def get_reporter_backend(self, name: str) -> CondaReporterBackend | None:
-        for reporter_backend in self.get_reporter_backends():
+    def get_reporter_backend(self, name: str) -> CondaReporterBackend:
```
jezdez, line 363: Let's improve this with a more strict fallback handling, similar to the solver loading behavior:

```python
    def get_reporter_backend(self, name: str) -> CondaReporterBackend:
        """
        Attempts to find a reporter backend while providing a fallback option if it is
        not found.

        This method must return a valid ``CondaReporterBackend`` object.
        """
        reporter_backends = {
            reporter_backend.name: reporter_backend
            for reporter_backend in self.get_reporter_backends()
        }
        reporter_backend = reporter_backends.get(name, None)
        if reporter_backend is None:
            log.warning(
                f'Unable to find reporter backend: "{name}"; '
                f'falling back to using "{DEFAULT_CONSOLE_REPORTER_BACKEND}"'
            )
            return reporter_backends.get(DEFAULT_CONSOLE_REPORTER_BACKEND)
        else:
            return reporter_backend

```

conda/plugins/manager.py
```diff
@@ -366,18 +360,33 @@ def get_virtual_packages(self) -> tuple[CondaVirtualPackage, ...]:
...
-            if reporter_output.name == name:
-                return reporter_output
+        for reporter_backend in reporter_backend_objs:
+            if reporter_backend.name == DEFAULT_CONSOLE_REPORTER_BACKEND:
+                return reporter_backend
+
+        raise CondaError(
+            "There are no available reporter backends to render output. This conda installation"
+            " is most likely corrupt and requires a re-installation."
+        )
```
jezdez, line 389: As mentioned above, that's not a valid outcome of this method and doesn't provide meaningful help to users. Conda essentially _has_ to ship a default console reporter backend that works, similar to the solver.

conda/plugins/reporter_backends/console.py
```diff
@@ -62,7 +62,7 @@ def conda_reporter_backends():
     Reporter backend for console
     """
     yield CondaReporterBackend(
-        name="console",
+        name=DEFAULT_CONSOLE_REPORTER_BACKEND,
```
jezdez, line 65: Let's use this for the strict definition of the default console reporter backed via a prioritized call order, similar how we use the `tryfirst` parameter of pluggy to prevent overriding the classic solver in https://github.com/conda/conda/blob/156a69dee64ce94f39c77271d3e558457e04fb43/conda/plugins/solvers.py#L9

That will assure hat the plugin loading method above will be able to always have a default console reporter backend to load.

```python
@hookimpl(tryfirst=True)  # make sure the default console reporter backend can't be overridden
def conda_reporter_backends():
    """
    Reporter backend for console
    """
    yield CondaReporterBackend(
        name=DEFAULT_CONSOLE_REPORTER_BACKEND,
        description="Default implementation for console reporting in conda",
        renderer=ConsoleReporterRenderer,
    )
```

conda/plugins/reporter_backends/json.py
```diff
@@ -43,7 +44,7 @@ def conda_reporter_backends():
     that can be passed to reporter outputs.
     """
     yield CondaReporterBackend(
-        name="json",
+        name=DEFAULT_JSON_REPORTER_BACKEND,
```
jezdez, line 47: Same here as with the default console reporter backend:

```python
@hookimpl(tryfirst=True)  # make sure the default json reporter backend can't be overridden
def conda_reporter_backends():
    """
    Reporter backend for JSON

    This is the default reporter backend that returns objects as JSON strings.
    """
    yield CondaReporterBackend(
        name=DEFAULT_JSON_REPORTER_BACKEND,
        description="Default implementation for JSON reporting in conda",
        renderer=JSONReporterRenderer,
    )
```

conda/reporters.py
```diff
@@ -6,27 +6,56 @@
...
+    from typing import Callable
 
-def render(data, style: str | None = None, **kwargs) -> None:
-    for settings in context.reporters:
-        reporter = context.plugin_manager.get_reporter_backend(settings.get("backend"))
-        output = context.plugin_manager.get_reporter_output(settings.get("output"))
+logger = logging.getLogger(__name__)
+
+
+@lru_cache
```
jezdez, line 25: Since the default size of the lru_cache may be larger depending on the objects handled by the function

```suggestion
@lru_cache(maxsize=None)
```

dev/start
```diff
@@ -199,7 +199,7 @@ if ! [ -f "${_DEVENV}/conda-meta/history" ]; then
     else
         [ -f "${_INSTALLER}/miniconda.exe" ] || _CMD='powershell.exe -Command "Invoke-WebRequest -Uri 'https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe' -OutFile '${_INSTALLER}/miniconda.exe' | Out-Null"'
     fi
-    if [ ${_CMD:+0} = 0 ]; then
+    if ! [ ${_CMD:-0} = 0 ]; then
```
jezdez, line 202: Another commit that should be in this PR?

tests/plugins/test_health_checks.py
```diff
@@ -25,6 +26,7 @@ def plugin_manager_with_doctor_command(plugin_manager):
     Registers the `conda doctor` subcommand
     """
     plugin_manager.load_plugins(doctor)
+    plugin_manager.load_plugins(*reporter_backends_plugins)
```
jezdez, line 29: That seems not ideal to handle for every test module, let's add a fixture to install must-have plugins in the test suite?

conda/env/specs/binstar.py
travishathaway: Yeah, that was an error with the pull request. Thanks for catching.

jezdez (approved):

PR #4447, opened 2017-01-27
subprocess_call function

kalefranz:
This PR is based on advice from http://stackoverflow.com/questions/13593223/making-sure-a-python-script-with-subprocesses-dies-on-sigint

addresses #4443

PR #14157, opened 2024-08-20
Implement `conda.common.path._cygpath` with roundtrip testing

kenodegard:
### Description

After working on both #14029 and #14146 it became clear that the Windows to Unix and Unix to Windows fallback conversions (for when `cygpath` is not available) were incomplete and did not offer proper roundtrip conversions.

We add a more complete test to cover proper roundtrip conversions (see `tests/common/test_path.py`) but also improve overall coverage. As part of this move we migrated the following and provide appropriate deprecations:
- `conda.activate._Cygpath` → `conda.common.path._cygpath`
- `conda.activate.native_path_to_unix` → `conda.common.path.win_path_to_unix`
- `conda.activate.unix_path_to_native` → `conda.common.path.unix_path_to_win`

Closes #14146
Closes #14029 
Xref #14068
Xref #13649

### Checklist - did you ...


kenodegard:
@ifitchet since you spearheaded the initial `_Cygpath` implementation in #13649, I'd greatly appreciate your input!

tests/common/test_path.py
```diff
@@ -193,3 +210,193 @@ def test_path_identity(tmp_path: Path) -> None:
...
+def test_path_conversion_falsy(
+    paths: str | Iterable[str] | None, expected: str | Iterable[str] | None
+) -> None:
+    assert win_path_to_unix(paths) == expected
+    assert unix_path_to_win(paths) == expected
+
+
+@pytest.mark.parametrize(
+    # NOTE: we automatically test for paths with redundant slashes
+    # NOTE: for windows paths we offer two roots; LIB (the resolved path) and ALT (the same path but with indirection)
```
ifitchet, line 234: Can you expand on what you mean by `{LIB}` and `{ALT}` (maybe example values)?

tests/common/test_path.py
```diff
@@ -193,3 +210,193 @@ def test_path_identity(tmp_path: Path) -> None:
...
+        pytest.param("//", "\\\\", True, id="UNC"),
+        pytest.param("//mount", "\\\\mount", True, id="UNC"),
+        pytest.param("//mount/", "\\\\mount\\", True, id="UNC"),
+        pytest.param("//mount//", "\\\\mount\\", None, id="UNC"),
+        pytest.param(None, "\\\\mount\\", "//mount/", id="UNC"),
+        pytest.param("//mount/CaSe", "\\\\mount\\CaSe", True, id="UNC"),
+        pytest.param("//mount//CaSe", "\\\\mount\\CaSe", None, id="UNC"),
+        pytest.param(None, "\\\\mount\\CaSe", "//mount/CaSe", id="UNC"),
+        # drive (1 leading slash + 1 letter)
+        # /c & /C doesn't roundtrip
```
ifitchet, line 272:
```suggestion
        # /c & /C doesn't roundtrip because the normal form is /c/ -- see below
```

tests/common/test_path.py
```diff
@@ -193,3 +210,193 @@ def test_path_identity(tmp_path: Path) -> None:
...
+        pytest.param("/c", "C:\\", None, id="drive"),
+        pytest.param("/C", "C:\\", None, id="drive"),
+        # c: & C: doesn't roundtrip
```
ifitchet, line 275:
```suggestion
        # c: & C: doesn't roundtrip because the normal form is c:\ (or c:\\ in a string)
```

tests/common/test_path.py
```diff
@@ -193,3 +210,193 @@ def test_path_identity(tmp_path: Path) -> None:
...
+        pytest.param(None, "c:\\drive", "/c/drive", id="drive"),
+        pytest.param(None, "C:\\drive", "/c/drive", id="drive"),
+        pytest.param(None, "c:\\drive", "/c/drive", id="drive"),
+        pytest.param("/c/drive/CaSe", "C:\\drive\\CaSe", True, id="drive"),
+        pytest.param(None, "c:\\drive\\CaSe", "/c/drive/CaSe", id="drive"),
+        # relative path
+        pytest.param("relative", "relative", True, id="relative"),
+        pytest.param("relative/", "relative\\", True, id="relative"),
+        pytest.param("relative/CaSe", "relative\\CaSe", True, id="relative"),
+        # odd cases
```
ifitchet, line 294: Hardly odd, a whole category of other problems!  😆 

```suggestion
        # PATH forms
```

tests/common/test_path.py
```diff
@@ -193,3 +210,193 @@ def test_path_identity(tmp_path: Path) -> None:
...
+        pytest.param("path:", "path;.", None, id="colon"),
+        pytest.param(None, "path;.", "path:.", id="colon"),
+        pytest.param(None, "path;", "path", id="colon"),
```
ifitchet, line 297: I would add some variants where the "blank" is at the start, `":path"` etc.

tests/common/test_path.py
kenodegard: Is this clear?

```suggestion
    # NOTE: for windows paths we offer two roots for testing:
    #         - LIB (resolved path), e.g., Z:\\fake\\root
    #         - ALT (indirect path), e.g., Z:\\fake\\non_root\\.\\..\\root
```

ifitchet: `fake` had me wondering for a moment, maybe `somewhere`?

But yes, `ALT` is a yet-to-be-resolved pathname then should equate to `LIB`, got it!

kenodegard: Oops! I also forgot to include the `Library` dir in the examples!

> `fake` had me wondering for a moment, maybe `somewhere`?

totally fair, is it still clear if I simply just use `root` and `non_root` like so:

```suggestion
    # NOTE: for windows paths we offer two roots for testing:
    #         - LIB (resolved path), e.g., Z:\\root\\Library
    #         - ALT (unresolved path), e.g., Z:\\non_root\\.\\..\\root\\Library
```

and "unresolved" is probably clearer than calling it "indirect", I'll make similar updates elsewhere

tests/common/test_path.py
kenodegard: good idea, added in https://github.com/conda/conda/pull/14157/commits/f93c1babef0fa5ad83be4c4c23b7b6c4a5d445ca

tests/common/test_path.py
ifitchet: At the risk of confusing everyone... 😄 MSYS2 (technically, the MSYS environment) computes that the Unix root `/` is two directories up from wherever `msys-2.0.dll` is installed.  If you install MSYS2 in some directory, ie. you'll end up with `Z:\somewhere\usr\lib\msys-2.0.dll` then the Unix root, `/`, equates to `Z:\somewhere`.

_We_ install MSYS2 packages in `Z:\somewhere\envs\env\Library`, ie. we'll have an `Z:\somewhere\envs\env\Library\usr\lib\msys-2.0.dll` which means the Unix `/` is `Z:\somewhere\envs\env\Library`.

So the `Library` bit is an artifact of where we choose to install packages and not a function of MSYS2 in any way.

So I would either stick to `Z:\root` (which I thought was fine but maybe I'm attuned to MSYS2 now!) or, if you want to use `Library` in there to match where we install packages then _don't_ use `root` as, for your examples, the Unix root filesystem, `/`, is `Z:\root\Library` and the `root` bit is now confusing!  😕  It isn't the "root" of anything (Unixy), just the parent directory of `Library`.

kenodegard: Ah I see what you mean 🤔

kenodegard: And the Cygwin root is defined in the same way?

ifitchet: Yes, exactly the same.  For want of a better description, MSYS2 is a (very) sawn-off Cygwin.  Plus some patches/convenience changes.

Some things _look_ quite different, `/cygdrive/z/foo` != `/z/foo`, but that's just a small patch (to `msys-2.0.dll` 😉 ) to change the prefix used for Windows drives.

kenodegard: I've attempted to untangle the difference between the prefix and root in https://github.com/conda/conda/pull/14157/commits/95c019ddf652f7b6261e4dc2b69bc8338fee9073

conda/common/path/_cygpath.py
```diff
@@ -0,0 +1,348 @@
...
+    # normalize path to remove duplicate slashes, .., and .
+    prefix = ntpath.normpath(prefix)
+
+    # root is defined as:
+    #   {root}\usr\lib\msys-2.0.dll
+    # the conda community chose to define that as:
+    #   %CONDA_PREFIX%\Library\usr\lib\msys-2.0.dll
+    # so we need to match the root path as:
+    #   {prefix}\Library
+    # ref: https://github.com/conda/conda/pull/14157#discussion_r1725384636
```
ifitchet, line 96: Can I persuade you to change the commentary?

```suggestion
    # MSYS2's root filesystem, /, is defined relative to this DLL:
    #   {root}\usr\lib\msys-2.0.dll
    # the conda community has chosen to install that DLL as:
    #   %CONDA_PREFIX%\Library\usr\lib\msys-2.0.dll
    # so we can infer the root path is:
    #   {prefix}\Library
    # ref: https://github.com/conda/conda/pull/14157#discussion_r1725384636
```

tests/common/test_path.py
ifitchet: Heh, that commit ended up being a lot of changes! 😄 

I've added a suggestion to change the commentary of that block, every little helps for whomever comes after!

conda/common/path/__init__.py
```diff
@@ -406,12 +367,136 @@ def is_package_file(path):
...
+def win_path_to_unix(
+    paths: PathType | PathsType | None,
+    prefix: PathType | None = None,
+    *,
+    cygdrive: bool = False,
+) -> str | tuple[str, ...] | None:
+    """Convert Windows paths to Unix paths.
+
+    .. note::
+        Can be expected to produce unexpected results when run on Unix.
```
travishathaway, line 474:
```suggestion
        Produces unexpected results when run on Unix.
```

conda/common/path/__init__.py
```diff
@@ -406,12 +367,136 @@ def is_package_file(path):
...
+def unix_path_to_win(
+    paths: PathType | PathsType | None,
+    prefix: PathType | None = None,
+    *,
+    cygdrive: bool = False,
+) -> str | tuple[str, ...] | None:
+    """Convert Unix paths to Windows paths.
+
+    .. note::
+        Can be expected to produce unexpected results when run on Unix.
```
travishathaway, line 494:
```suggestion
        Produces unexpected results when run on Unix.
```

conda/common/path/__init__.py
```diff
@@ -406,12 +367,136 @@ def is_package_file(path):
...
+        cygpath_arg = "--windows"
+        cygpath_fallback = posix_to_nt
+        to_pathsep = ntpath.pathsep
+        to_sep = ntpath.sep
+
+    # It is very easy to end up with a bash in one place and a cygpath in another due to e.g.
+    # using upstream MSYS2 bash, but with a conda env that does not have bash but does have
+    # cygpath.  When this happens, we have two different virtual POSIX machines, rooted at
+    # different points in the Windows filesystem.  We do our path conversions with one and
+    # expect the results to work with the other.  It does not.
```
travishathaway, line 416: I am not sure why you put this comment here. What is it trying to accomplish?

conda/common/path/__init__.py
```diff
@@ -406,12 +367,136 @@ def is_package_file(path):
...
+    cygpath = str(Path(bash).parent / "cygpath") if bash else "cygpath"
+    joined = (
+        str(paths)
+        if isinstance(paths, (str, os.PathLike))
+        else from_pathsep.join(map(str, paths))
+    )
+
+    try:
+        # if present, use cygpath to convert paths since its more reliable
+        converted = subprocess.run(
```
travishathaway, line 429: Is there a reason we aren't using  `conda.gateways.subprocess.subprocess_call` here?

conda/common/path/__init__.py
```diff
@@ -406,12 +367,136 @@ def is_package_file(path):
...
+        converted = cygpath_fallback(joined, prefix, cygdrive)
+    except subprocess.CalledProcessError as err:
+        log.error(
+            "Unexpected cygpath error\n  %s: %s\n  stdout: %s\n  stderr: %s",
+            err.__class__.__name__,
+            err,
+            err.stdout.strip(),
+            err.stderr.strip(),
+        )
+        raise
```
travishathaway, line 449: I'm curious whether we can use the `cygpath_fallback` function in this and the other exception case.

Are we absolutely sure we've reached some sort of unrecoverable error state that warrants halting the execution of the entire program? Could we also raise a subclass of CondaError with an actionable error message so that the user isn't presented with a traceback?

conda/common/path/__init__.py
kenodegard: Counter argument, why should we use `conda.gateways.subprocess.subprocess_call` and not move to deprecate it? It is not at all clear that it does anything better or safer. It appears to have initially been implemented to do a better job handling signals, see https://github.com/conda/conda/pull/4447 and https://stackoverflow.com/questions/13593223/making-sure-a-python-script-with-subprocesses-dies-on-sigint, at this point it smells like old Python 2 code.

conda/common/path/__init__.py
kenodegard: no reason beyond it "wasn't done previously", added in https://github.com/conda/conda/pull/14157/commits/5272cc9693db65b7fcc90557e1ddb6e9f28e259c

travishathaway: Understood. The way the pull request is presented makes it look like you wrote that code, so it's hard to determine what was already there and what you have recently added.

I don't have a problem keeping it the way it was.

conda/common/path/__init__.py
travishathaway: Fair point!

Maybe we should at least update the comment for `conda.gateways.subprocess.subprocess_call` then:

> This utility function should be preferred for all conda subprocessing.

This can be saved for another pull request though.

conda/common/path/__init__.py
jezdez, line 1: That's a lot of code to put into a `__init__.py`, could we separate the code paths a little more and build an API in the module init?

kenodegard: I took a stab at splitting it up into smaller submodules and ended up with `conda.common.path.directories`, `conda.common.path.python`, and `conda.common.path.windows`.

The functions left in `__init__.py` are pretty small. Several of the functions defined here have very limited usage so I'd hope we could find ways to deprecate them entirely at some point.

jezdez (approved):

PR #14170, opened 2024-08-23
Refactor `classic` solver out

jaimergp:
### Checklist - did you ...

Supersedes https://github.com/conda/conda/pull/14167


jaimergp:
Needs https://github.com/conda/conda/pull/14475 before cleanup.

tests/test_create.py
```diff
@@ -2457,7 +2457,7 @@ def test_conda_downgrade(
         PrefixData._cache_.clear()
         subprocess_call(
             [
-                python_exe,
+                sys.executable,
```
jaimergp, line 2460: @kenodegard, can you confirm this fix is correct? Judging by the comment above, I think so? This makes the test pass.

.github/workflows/tests.yml
jaimergp, line 1: We'll need to publish [`conda-classic-solver`](https://github.com/conda/conda-classic-solver) in both `defaults` and `conda-forge` before we can undo this.

PR #14179, opened 2024-08-27
Handle 'null' for noarch values

sumanth-manchala:
### Description

We recently had an artifactory issue where it was deserialising empty noarch on meta.yaml for some of our internally built packages to `noarch: null` on repodata.json and this broke the entire conda workflows that are using our internal channels with the below error.

While we are getting artifactory to fix the repodata, I figured that NoarchType enum handle null values too like in this PR.

If the earlier usage is intentional and a hard check not to handle null values on NoarchType, feel free to close this PR.

```
  File "/<>/conda/conda/models/enums.py", line 164, in coerce
    val = NoarchType.generic if boolify(val) else None
  File "/<>/conda/conda/auxlib/type_coercion.py", line 170, in boolify
    raise TypeCoercionError(value, "The value %r cannot be boolified." % value)
conda.auxlib.type_coercion.TypeCoercionError: The value 'null' cannot be boolified.

  File "/<>/conda/conda/models/enums.py", line 166, in coerce
    raise CondaUpgradeError(
conda.exceptions.CondaUpgradeError: The noarch type for this package is set to 'null'.
The current version of conda is too old to install this package.
Please update conda.
```

### Checklist - did you ...

### DIFF
--- a/conda/models/enums.py
+++ b/conda/models/enums.py
@@ -161,7 +161,7 @@ class NoarchType(Enum):
                 val = NoarchType.generic
             else:
                 try:
-                    val = NoarchType.generic if boolify(val) else None
+                    val = NoarchType.generic if boolify(val, nullable=True) else None
                 except TypeCoercionError:
                     raise CondaUpgradeError(
                         dals(


travishathaway:
@conda-bot check

jezdez:
not stale

wolfv:
From what I've heard, this is fixed with newer Artifactory versions.

sumanth-manchala:
@wolfv , yes this is resolved with the newer version of artifactory but we have a significant numbers of builds with malformed repodata records which we need to keep. 

As per our internal teams we can only delete entries from repodata. Hence we are running a custom miniforge env with this patch which requires maintaince/coordination between multipe teams for an upgrade.

Surprisingly, we only hit the above exception only during conda-build and these malformed records are handled well with libmamba/pixi.

Let me know if this patch go in on conda so that I can add an unit test for missing coverage. else please feel free to close this PR. Thank you for the followup!

jezdez:
pre-commit.ci autofix

ryanskeith (approved):
This looks sound to me.

jezdez:
Thank you @sumanth-manchala!

PR #14196, opened 2024-08-29
 Add `denylist_channels` config option.

### LINKED ISSUE #14176: Global channel denylist
### Checklist

### What is the idea?

Add the possibility to force-disable/denylist some channels

### Why is this needed?

We do not want to use the defaults channel but it randomly gets pulled in by some projects, environment.yaml's, etc.
We cannot watch every user or piece of software to not use the defaults channel.

### What should happen?

Users would see a warning that some channel (e.g. defaults) has been blacklisted and will not be used for dependency resolution.

### Additional Context

Anaconda now sues organizations for using their channels:
https://www.reuters.com/legal/litigation/intel-sued-copyright-infringement-over-ai-software-2024-08-09/

jezdez:
### Description

This adds a `denylist_channels` config option, which complements the already existing `allowlist_channels` config option.

Fix #14176.

### Checklist - did you ...

### DIFF
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -52,7 +52,7 @@ def check_allowlist(channel_urls: list[str]) -> None:
     for channel_url in channel_urls:
         channel = Channel(channel_url)
         for channel_base_url in channel.base_urls:
-            if denylist_channel_urls and channel_base_url in denylist_channel_urls:
+            if channel_base_url in denylist_channel_urls:
                 raise ChannelDenied(channel)
             if (
                 context.allowlist_channels
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -49,16 +49,17 @@ def check_allowlist(channel_urls: list[str]) -> None:
             for denylist_channel in context.denylist_channels
         )
     )
-    for channel_url in channel_urls:
-        channel = Channel(channel_url)
-        for channel_base_url in channel.base_urls:
-            if channel_base_url in denylist_channel_urls:
-                raise ChannelDenied(channel)
-            if (
-                context.allowlist_channels
-                and channel_base_url not in allowlist_channel_urls
-            ):
-                raise ChannelNotAllowed(channel)
+    if allowlist_channel_urls or denylist_channel_urls:
+        for channel_url in channel_urls:
+            channel = Channel(channel_url)
+            for channel_base_url in channel.base_urls:
+                if channel_base_url in denylist_channel_urls:
+                    raise ChannelDenied(channel)
+                if (
+                    allowlist_channel_urls
+                    and channel_base_url not in allowlist_channel_urls
+                ):
+                    raise ChannelNotAllowed(channel)
 
 
 LAST_CHANNEL_URLS = []
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -52,7 +52,7 @@ def check_allowlist(channel_urls: list[str]) -> None:
             ):
                 raise ChannelNotAllowed(Channel(channel_url))
 
-    elif context.denylist_channels:
+    if context.denylist_channels:
         denylist_channel_urls = tuple(
             chain.from_iterable(
                 Channel(denylist_channel).base_urls
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -54,7 +54,10 @@ def check_allowlist(channel_urls: list[str]) -> None:
         for channel_base_url in channel.base_urls:
             if denylist_channel_urls and channel_base_url in denylist_channel_urls:
                 raise ChannelDenied(channel)
-            if context.allowlist_channels and channel_base_url not in allowlist_channel_urls:
+            if (
+                context.allowlist_channels
+                and channel_base_url not in allowlist_channel_urls
+            ):
                 raise ChannelNotAllowed(channel)
 
 
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -37,35 +37,25 @@ def check_allowlist(channel_urls: list[str]) -> None:
     :raises ChannelNotAllowed: If any URL is not in the allowlist.
     :raises ChannelDenied: If any URL is in the denylist.
     """
-    if context.allowlist_channels:
-        allowlist_channel_urls = tuple(
-            chain.from_iterable(
-                Channel(allowlist_channel).base_urls
-                for allowlist_channel in context.allowlist_channels
-            )
+    allowlist_channel_urls = tuple(
+        chain.from_iterable(
+            Channel(allowlist_channel).base_urls
+            for allowlist_channel in context.allowlist_channels
         )
-        for channel_url in channel_urls:
-            channel_base_urls = Channel(channel_url).base_urls
-            if not all(
-                channel_base_url in allowlist_channel_urls
-                for channel_base_url in channel_base_urls
-            ):
-                raise ChannelNotAllowed(Channel(channel_url))
-
-    if context.denylist_channels:
-        denylist_channel_urls = tuple(
-            chain.from_iterable(
-                Channel(denylist_channel).base_urls
-                for denylist_channel in context.denylist_channels
-            )
+    )
+    denylist_channel_urls = tuple(
+        chain.from_iterable(
+            Channel(denylist_channel).base_urls
+            for denylist_channel in context.denylist_channels
         )
-        for channel_url in channel_urls:
-            channel_base_urls = Channel(channel_url).base_urls
-            if any(
-                channel_base_url in denylist_channel_urls
-                for channel_base_url in channel_base_urls
-            ):
-                raise ChannelDenied(Channel(channel_url))
+    )
+    for channel_url in channel_urls:
+        channel = Channel(channel_url)
+        for channel_base_url in channel.base_urls:
+            if denylist_channel_urls and channel_base_url in denylist_channel_urls:
+                raise ChannelDenied(channel)
+            if context.allowlist_channels and channel_base_url not in allowlist_channel_urls:
+                raise ChannelNotAllowed(channel)
 
 
 LAST_CHANNEL_URLS = []
--- a/conda/core/index.py
+++ b/conda/core/index.py
@@ -12,7 +12,7 @@ from boltons.setutils import IndexedSet
 
 from ..base.context import context
 from ..common.io import ThreadLimitedThreadPoolExecutor, time_recorder
-from ..exceptions import ChannelNotAllowed, InvalidSpec
+from ..exceptions import ChannelDenied, ChannelNotAllowed, InvalidSpec
 from ..gateways.logging import initialize_logging
 from ..models.channel import Channel, all_channel_urls
 from ..models.enums import PackageType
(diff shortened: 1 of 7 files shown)


conda/core/index.py
```diff
@@ -35,17 +35,37 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+                for denylist_channel in context.denylist_channels
+            )
+        )
+        for channel_url in channel_urls:
+            channel_base_urls = Channel(channel_url).base_urls
+            if any(
+                channel_base_url in denylist_channel_urls
+                for channel_base_url in channel_base_urls
+            ):
+                raise ChannelDenied(Channel(channel_url))
```
jezdez, line 68: I opted to not refactor this further since it really is just two different cases after all.

conda/core/index.py
```diff
@@ -35,17 +35,37 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
-                raise ChannelNotAllowed(Channel(url))
+        for channel_url in channel_urls:
+            channel_base_urls = Channel(channel_url).base_urls
+            if not all(
+                channel_base_url in allowlist_channel_urls
+                for channel_base_url in channel_base_urls
+            ):
+                raise ChannelNotAllowed(Channel(channel_url))
+
+    elif context.denylist_channels:
```
jaimergp, line 55: If they are mutually exclusive, I think we should error out early so `denylist` is not ignored, and this should be documented in their `context` docstrings.

jaimergp:
Given the timing and need for this feature, I think we should also add some docs to at least one of these pages:

- https://docs.conda.io/projects/conda/en/stable/user-guide/tasks/manage-channels.html
- https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/index.html

jaimergp:
Food for thought, `context.channels` is not the only way we feed channel configurations into different parts of the stack:

In the general solver logic, it's possible to pass `channels` directly via the signature, overriding `context.channels`:

https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/solve.py#L93-L94

In conda-libmamba-solver, installed packages can push their channels outside the configuration:

https://github.com/conda/conda-libmamba-solver/blob/ceae580d6e6b51573cb88cc5411c323ff33084c3/conda_libmamba_solver/solver.py#L191-L197

In SubdirData, we should error out in the passed channel is not in allowed or is denied:

https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/subdir_data.py#L175

Downloads via ProgressFetchExtract should check for URLs in this block. This is important for @EXPLICIT installs:

https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/package_cache_data.py#L745-L763


So all in all, I think we are ok without covering the solver cases, but we MUST cover SubdirData and ProgressiveFetchExtract because those are the blocks that might hit denied channels one way or another (repodata and artifact download, respectively).

conda/core/index.py
jezdez: Nah, this is an oversight only, they should not be mutually exclusive

jezdez:
> Food for thought, `context.channels` is not the only way we feed channel configurations into different parts of the stack:
> 
> In the general solver logic, it's possible to pass `channels` directly via the signature, overriding `context.channels`:
> 
> https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/solve.py#L93-L94
> 
> In conda-libmamba-solver, installed packages can push their channels outside the configuration:
> 
> [conda/conda-libmamba-solver@`ceae580`/conda_libmamba_solver/solver.py#L191-L197](https://github.com/conda/conda-libmamba-solver/blob/ceae580d6e6b51573cb88cc5411c323ff33084c3/conda_libmamba_solver/solver.py#L191-L197)

I'd suggest fixing this in conda-libmamba-solver, basically call the 

> In SubdirData, we should error out in the passed channel is not in allowed or is denied:
> 
> https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/subdir_data.py#L175

Doesn't the call to `check_allowlist` in `SubdirData.query_all` suffice? Could you provide a code path (since it's a whole mess and I don't have time to dig further) where the MUST becomes clear?

> Downloads via ProgressFetchExtract should check for URLs in this block. This is important for @explicit installs:
> 
> https://github.com/conda/conda/blob/ef93541d0132837ede502a2b62e88d9169cdc872/conda/core/package_cache_data.py#L745-L763

I don't follow, how is `ProgressFetchExtract` related to this? This looks far too deep in the call stack to verify a channel list.

> So all in all, I think we are ok without covering the solver cases, but we MUST cover SubdirData and ProgressiveFetchExtract because those are the blocks that might hit denied channels one way or another (repodata and artifact download, respectively).

A MUST is pretty stark given that it seems to me that those are additional cases that can be improved. We're about to release the September release and this has to be in there.

conda/core/index.py
jezdez: Nevermind, had to refactor

docs/source/user-guide/configuration/settings.rst
```diff
@@ -97,6 +97,63 @@ handler called, "test-auth-handler" registered via the aforementioned plugin hoo
...
+    - conda-forge
+
+An example for the ``denylist_channels`` configuration is (which would deny the conda-forge channel):
+
+.. code-block:: yaml
+
+  denylist_channels:
+    - conda-forge
+
+An example for both the ``allowlist_channels`` and ``denylist_channels`` configuration is (which would explicitely allow the defaults channel but deny the conda-forge channel):
```
jezdez, line 130:
```suggestion
An example for both the ``allowlist_channels`` and ``denylist_channels`` configuration is
(which would explicitely allow the defaults channel but deny the conda-forge channel):
```

jaimergp:
> I'd suggest fixing this in conda-libmamba-solver, basically call the `check_allowlist` function.
> 
> >


The only `check_allowlist()` call in the whole conda codebase is in `SubdirData.query_all()` and `conda.core.index.get_index()` 😱 

https://github.com/conda/conda/blob/2fbf0cc62e8539f97591933082f8fa7b0c0f8ac1/conda/core/subdir_data.py#L129

I would have expected this to be more in the CLI layer than the library layer (maybe in `conda.cli.install`). I'll add it to conda-libmamba-solver anyway.

> Doesn't the call to `check_allowlist` in `SubdirData.query_all` suffice? Could you provide a code path (since it's a whole mess and I don't have time to dig further) where the MUST becomes clear?

That's only `query_all`, which is a classmethod. There's no check in `__init__` so all other methods (like only downloading the repodata.json via the `load` / `repo_fetch` interfaces) won't trigger the warnings.

> I don't follow, how is ProgressFetchExtract related to this? This looks far too deep in the call stack to verify a channel list

`@EXPLICIT` lockfiles won't download the repodata (they are "channel-less" in a way), they will make `conda` download the artifacts directly from the channel. If those artifacts are hosted in `pkgs/main` that's also a potential violation of the TOS.

conda/core/index.py
```diff
@@ -35,17 +35,30 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
         )
-        for url in channel_urls:
-            these_urls = Channel(url).base_urls
-            if not all(this_url in allowlist_channel_urls for this_url in these_urls):
-                raise ChannelNotAllowed(Channel(url))
+    )
+    for channel_url in channel_urls:
+        channel = Channel(channel_url)
+        for channel_base_url in channel.base_urls:
+            if denylist_channel_urls and channel_base_url in denylist_channel_urls:
```
jaimergp, line 55:
```suggestion
            if channel_base_url in denylist_channel_urls:
```

Do we need the first part of the check? I think that's only necessary for the allowlist.

docs/source/user-guide/configuration/settings.rst
```diff
@@ -97,6 +97,65 @@ handler called, "test-auth-handler" registered via the aforementioned plugin hoo
    schema must match exactly to the channel URL, so a pattern like ``*`` is not valid.
 
 
+``allowlist_channels`` and ``denylist_channels``: Allow or deny specific channels
+--------------------------------------------------------------------------------
```
jaimergp, line 101:
```suggestion
---------------------------------------------------------------------------------
```

docs/source/user-guide/configuration/settings.rst
```diff
@@ -97,6 +97,65 @@ handler called, "test-auth-handler" registered via the aforementioned plugin hoo
...
+
+With ``allowlist_channels`` and ``denylist_channels``, you can allow or deny specific channels
+from being used in conda operations. This is useful for restricting the channels that conda
+can access, especially in enterprise or multi-user environments.
+
+The denylist takes precedence over the allowlist. If a channel is in both lists, it is denied.
+
+**Examples:**
+
+An example which allows the ``defaults`` and ``conda-forge`` channels with the ``alllowlist_channels``
```
jaimergp, line 116:
```suggestion
An example which allows the ``defaults`` and ``conda-forge`` channels with the ``allowlist_channels``
```

docs/source/user-guide/configuration/settings.rst
```diff
@@ -97,6 +97,65 @@ handler called, "test-auth-handler" registered via the aforementioned plugin hoo
...
+    - conda-forge
+
+An example which denies the ``conda-forge`` channel with the ``denylist_channels`` setting is:
+
+.. code-block:: yaml
+
+  denylist_channels:
+    - conda-forge
+
+An example which explicitely allows the ``defaults`` channel but denies the ``conda-forge`` channel
```
jaimergp, line 132:
```suggestion
An example which explicitly allows the ``defaults`` channel but denies the ``conda-forge`` channel
```

docs/source/user-guide/configuration/settings.rst
```diff
@@ -97,6 +97,65 @@ handler called, "test-auth-handler" registered via the aforementioned plugin hoo
...
+    - defaults
+  denylist_channels:
+    - https://conda.anaconda.org/conda-forge/linux-64
+
+An example that denies using the ``defaults`` channel:
+
+.. code-block:: yaml
+
+  denylist_channels:
+    - defaults
```
jaimergp, line 157: Since `defaults` can be configured via `default_channels`, would it make more sense to have the `repo.anaconda.com` URLs listed here?

jezdez: Okay, let me actually mention that `default_channels` part and have both, that'll be important to clarify

jezdez: be9872aea27a0af2a2da43c69777db1796b30a6b

conda/core/index.py
```diff
@@ -35,17 +35,30 @@ def check_allowlist(channel_urls: list[str]) -> None:
...
+    for channel_url in channel_urls:
+        channel = Channel(channel_url)
+        for channel_base_url in channel.base_urls:
+            if channel_base_url in denylist_channel_urls:
+                raise ChannelDenied(channel)
+            if (
+                context.allowlist_channels
+                and channel_base_url not in allowlist_channel_urls
+            ):
+                raise ChannelNotAllowed(channel)
```
kenodegard, line 61: two edits:
 1. only iterate over channel_urls if either allowlist or denylist are defined
 2. use `allowlist_channel_urls` instead of `context.allowlist_channels`

```suggestion
    if allowlist_channel_urls or denylist_channel_urls:
        for channel_url in channel_urls:
            channel = Channel(channel_url)
            for channel_base_url in channel.base_urls:
                if channel_base_url in denylist_channel_urls:
                    raise ChannelDenied(channel)
                if (
                    allowlist_channel_urls
                    and channel_base_url not in allowlist_channel_urls
                ):
                    raise ChannelNotAllowed(channel)
```

kenodegard (changes requested):

Issue #14174, opened 2024-08-23
Raise error if attempting to remove a non-conda env path


kenodegard:
> shouldn't attempting to remove a non-existent env fail? currently it silently succeeds which is confusing:
> <img width="805" alt="Screenshot 2024-08-23 at 15 34 57" src="https://github.com/user-attachments/assets/03d5d697-4e16-4272-850d-8ec72f8b2c1c">

_Originally posted by @kenodegard in https://github.com/conda/conda/pull/14173#discussion_r1729489413_

beeankha:
The plot thickens... [this issue from 2016](https://github.com/conda/conda/issues/3489) was filed specifically to _remove_ the error output when attempting to remove a non-existent env. There is even a [regression test](https://github.com/conda/conda/blob/main/tests/test_create.py#L2306-L2310) for it. Adding the "pending discussion" label to this issue.

PR #14199, opened 2024-08-29
Raise error when attempting to remove non-existent environment

beeankha:
### Description

To resolve https://github.com/conda/conda/issues/14174

### Checklist - did you ...


conda/cli/main_remove.py
```diff
@@ -165,23 +167,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-        # not to exist afterwards, would you not? If not (fine, I can see the argument
-        # about deleting people's work in envs being a very bad thing indeed), but if
-        # being careful is the goal it would still be nice if after `conda remove --all`
-        # to be able to do `conda create` on the same environment name.
-        #
-        # try:
-        #     rm_rf(prefix, clean_empty_parents=True)
-        # except:
-        #     log.warning("Failed rm_rf() of partially existent env {}".format(prefix))
+    if args.all and not is_conda_environment(prefix):
```
kenodegard, line 173: Don't we _always_ want to check if the prefix is valid? Shouldn't it still fail here if I say `conda remove -n missing zlib`?

tests/cli/test_env.py
```diff
@@ -159,25 +137,24 @@ def test_deprecated_binstar(conda_cli: CondaCLIFixture):
...
-    create_env(ENVIRONMENT_CA_CERTIFICATES)
-    conda_cli("env", "create")
-    assert env_is_created(env1)
-
-    stdout, _, _ = conda_cli("info", "--json")
-    parsed = json.loads(stdout)
-    assert [env for env in parsed["envs"] if env.endswith(env1)]
+    with tmp_env() as prefix:
+        create_env(ENVIRONMENT_CA_CERTIFICATES)
+        assert is_conda_environment(prefix)
```
kenodegard, line 147: This isn't testing what it's supposed to test. Breaking it down:
1. `with tmp_env() as prefix`: creates an empty environment
2. `create_env(ENVIRONMENT_CA_CERTIFICATES)`: creates an `environment.yml` file
3. `assert is_conda_environment(prefix)`: ensures the env from 1. exists

The env file from 2. is never used!

We're gonna want something like this instead:

```suggestion
    create_env(ENVIRONMENT_CA_CERTIFICATES)
    prefix = path_factory()
    conda_cli(
        "env",
        "create",
        f"--prefix={prefix}",
        # "--file=environment.yml",  # the implicit default
    )
    assert is_conda_environment(prefix)
    assert package_is_installed(prefix, "ca-certificates")
```

kenodegard (changes requested):
One concern I have with the tests is how most are writing a `environment.yml` file into my CWD via the `create_env` helper function. Let's use something like `monkeypatch.chdir(tmp_path)` to avoid side effects and keep our CWD clean.

jezdez (approved):

kenodegard (approved):

PR #14235, opened 2024-09-10
Remove no-op `context.use_only_tar_bz2` logic

kenodegard:
### Description

The following code block will always evaluate to `use_only_tar_bz2 = False` so we can remove it (plus none of our subcommands supports providing a `--use-only-tar-bz2` flag anyhow):

https://github.com/conda/conda/blob/8ba894077c7ddfbcc4e258dc4cefcd1fae5a460d/conda/base/context.py#L972-L975

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -969,18 +969,10 @@ class Context(Configuration):
         #    is only called when use_only_tar_bz2 is first called.
         import conda_package_handling.api
 
-        use_only_tar_bz2 = False
-        if self._use_only_tar_bz2 is None:
-            if self._argparse_args and "use_only_tar_bz2" in self._argparse_args:
-                use_only_tar_bz2 &= self._argparse_args["use_only_tar_bz2"]
         return (
-            (
-                hasattr(conda_package_handling.api, "libarchive_enabled")
-                and not conda_package_handling.api.libarchive_enabled
-            )
-            or self._use_only_tar_bz2
-            or use_only_tar_bz2
-        )
+            hasattr(conda_package_handling.api, "libarchive_enabled")
+            and not conda_package_handling.api.libarchive_enabled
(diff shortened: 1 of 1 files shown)


conda/base/context.py
```diff
@@ -969,18 +969,10 @@ def use_only_tar_bz2(self):
...
         return (
-            (
-                hasattr(conda_package_handling.api, "libarchive_enabled")
-                and not conda_package_handling.api.libarchive_enabled
-            )
-            or self._use_only_tar_bz2
-            or use_only_tar_bz2
-        )
+            hasattr(conda_package_handling.api, "libarchive_enabled")
+            and not conda_package_handling.api.libarchive_enabled
```
kenodegard, line 974:
```suggestion
            not getattr(conda_package_handling.api, "libarchive_enabled", False)
```

kenodegard: @dholth thoughts on simplifying it this way? I've also checked and `conda_package_handling.api.libarchive_enabled` has been available since `CPH 1.4.0` (https://github.com/conda/conda-package-handling/commit/ab7bf9cc5f9e62a2829db81fd37820fb4c5c4677) so since we depend on `CPH>=2.2.0` the constant should always be defined

dholth: getattr is better but `not conda_package_handling.api.libarchive_enabled` is best.

conda/base/context.py
```diff
@@ -969,18 +969,9 @@ def use_only_tar_bz2(self):
         #    is only called when use_only_tar_bz2 is first called.
```
dholth, line 969:
```suggestion
# Import inside function to avoid circular import
```
Or is it really a PATH issue

dholth (approved):

PR #14236, opened 2024-09-11
Update AUTHORS.md

scw:
My name ended up getting duplicated in this list, I believe because I had at least one commit under a separate account, `scdub`, but was later able to recover the credentials for my primary account, `scw`. Reduce the list down to the one me, presuming there isn't another conda committing copy of myself lurking.

### Description

Minor change, so didn't add any other components alongside it.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -375,7 +375,6 @@ Authors are sorted alphabetically.
 * Shannon Green
 * Shaun Thompson
 * Shaun Walbridge
-* Shaun Walbridge
 * Sidhant Bansal
 * Simon Li
 * Simon McLaren


jezdez (approved):

kenodegard:
Modifying the `AUTHORS.md` file isn't enough to correct this since that file is auto generated by `rever` during every release, we would also need to modify `.authors.yml`:

https://github.com/conda/conda/blob/aa0fb6f3ae669a5ade575d340555aa6a9f71ef5e/.authors.yml#L147-L156

While `rever` supports aliases and alternative names it does not offer a means of combining multiple GitHub IDs into one entry. This doesn't really matter though since the GitHub ID is only used when generating the next release's changelog entry, so we could do something like this:

```yaml
- name: Shaun Walbridge
  email: shaun.walbridge@gmail.com
  alternate_emails:
    - 46331011+scdub@users.noreply.github.com
  github: scw  # alternate: scdub
```

PR #14239, opened 2024-09-12
Change how arguments are passed in in `Invoke-Conda` function

seanmamasde:
### Description

Due to breaking change in core .NET libraries, where passing `string.Empty` to `Environment.SetEnvironmentVariable` no longer deletes the environment variable, see [dotnet breaking chnage](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/9.0/empty-env-variable) and [this GitHub issue](https://github.com/PowerShell/PowerShell/issues/24268)

This solution should mitigate this issue PowerShell version above 7.5.0-preview.4. (This is just a workaround, further fixing how empty env vars are handled might be needed.)

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #14240, opened 2024-09-12
Change how arguments are passed in in `Invoke-Conda` function

seanmamasde:
### Description

Due to breaking change in core .NET libraries, where passing `string.Empty` to `Environment.SetEnvironmentVariable` no longer deletes the environment variable, see [dotnet breaking chnage](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/9.0/empty-env-variable) and [this GitHub issue](https://github.com/PowerShell/PowerShell/issues/24268)

This solution should mitigate this issue PowerShell version above 7.5.0-preview.4. (This is just a workaround, further fixing how empty env vars are handled might be needed.)

supersedes #14239

### Checklist - did you ...

### DIFF
--- a/conda/shell/condabin/Conda.psm1
+++ b/conda/shell/condabin/Conda.psm1
@@ -125,6 +125,11 @@ function Exit-CondaEnvironment {
         conda install toolz
 #>
 function Invoke-Conda() {
+    # To mitigate breaking change in .NET 9
+    # see github issue: https://github.com/PowerShell/PowerShell/issues/24268
+    if ($Env:_CE_M -eq "") { $Env:_CE_M = $null }
+    if ($Env:_CE_CONDA -eq "") { $Env:_CE_CONDA = $null }
+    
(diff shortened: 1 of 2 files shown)


kenodegard:
thanks @seanmamasde for helping to debug this

FYI in the future there's no need to close and open a new PR to modify a PR, you just need to push the additional changes to the original branch

since the activation/deactivate logic may unset other environment variables I think we want to correct this a little deeper in the code, can you take a look at my alternative implementation in https://github.com/conda/conda/pull/14246 and let me know what you think?

seanmamasde:
yeah you're right. I think #14246 is how it should be done instead. I was using PowerShell 7.5.0-preview 4 with miniforge3 24.7.1.0 when this happened, since it's not merge yet. Thanks for the insight on the PR thing, I didn't know that is possible. I am going to close this PR in favor of the one you mentioned.

Issue #14217, opened 2024-09-06
Channel handling improvements


jezdez:
### Checklist

- [X] I added a descriptive title
- [X] I searched open issues and couldn't find a duplicate

### What?

Following feedback from end users around concerns related to channels with terms of services and other legal restrictions, we can improve conde to handle channel management a little better by default. The goal of this initiative is to work on a consistent resolution to user experience issues as well as overcoming some tech debt in conda's code base.

### Why?

Various stakeholders in the conda ecosystem have reported concerns about the pre-configuration of conda to favor channels from Anaconda Inc. While distributions like [miniforge](https://github.com/conda-forge/miniforge) come pre-configured for conda-forge, conda itself hard-codes Anaconda Inc channels within its code base, due to the way conda was incubated historically.

Conda has been a full Open-Source project with a permissive license and governed by [a steering council per its governance policy](https://github.com/conda/governance) for a while now. As such, this pre-configuration has detrimental effects on the trust and reputation of the conda OSS project among conda users that aren't Anaconda Inc customers.

### User impact

Less confusing way of handling channels, no accidental use of an unwanted channel, conda is more agnostic to how channels are run etc.

### Goals

- user focus, reduce legal risks for end users
- overcome historic decisions on how conda is configured, pay down tech-debt

### Tasks

- [ ] https://github.com/conda/conda/issues/12356
- [ ] https://github.com/conda/conda/issues/14178
- [ ] https://github.com/conda/conda/issues/14176

### This epic is blocked by:

- [ ] <dependency epic or issue>

### This epic blocks:

- [ ] <dependency epic or issue>

jezdez:
@jaimergp I've started this epic, in case you see other tickets, feel free to mention them here

SylvainCorlay:
As the conda project transitions towards a more open governance, it makes perfect sense for it to be developed with **vendor agnosticism** in mind, and I salute @jezdez for opening this issue. As pointed out in this issue, the fact that Anaconda channels are hardcoded as the default channels in the codebase is one instance in which the project is currently not vendor agnostic, although it could easily be overriden (like seen in miniforge).

Additionally, it would be beneficial to update the project's **readme** and **documentation** to:
 - clearly state that the miniconda and anaconda installers advertised in the documentation may incur costs. (I think it is perfectly fine to advertise these installers, and to mention that the associated costs come with e.g. commercial support or other benefits).
[...]

PR #14269, opened 2024-09-24
Mark `restore_free_channel` for deprecation.

### LINKED ISSUE #14231: Deprecate the `restore_free_channel` configuration option.

jezdez:
### Description

Fix #14231.
Refs #14217.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -869,6 +869,15 @@ class Context(Configuration):
             default_channels = list(self._default_channels)
 
         if self.restore_free_channel:
+            deprecated.topic(
+                "24.9",
+                "25.3",
+                topic="Adding the 'free' channel as it existed prior to conda 4.7.",
+                addendum="Add 'https://repo.anaconda.com/pkgs/free' to your channels "
+                "list manually after 'defaults', if you still need content from "
+                "the 'free' channel.",
+                deprecation_type=FutureWarning,
+            )
             default_channels.insert(1, "https://repo.anaconda.com/pkgs/free")
 
         reserved_multichannel_urls = {
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -873,9 +873,9 @@ class Context(Configuration):
                 "24.9",
                 "25.3",
                 topic="Adding the 'free' channel as it existed prior to conda 4.7.",
-                addendum="Add 'https://repo.anaconda.com/pkgs/free' to your channels "
-                "list manually after 'defaults', if you still need content from "
-                "the 'free' channel.",
+                addendum="See "
+                "https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/free-channel.html "
+                "for more details.",
                 deprecation_type=FutureWarning,
             )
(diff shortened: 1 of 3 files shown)


conda/base/context.py
```diff
@@ -869,6 +869,15 @@ def custom_multichannels(self):
...
         if self.restore_free_channel:
+            deprecated.topic(
+                "24.9",
+                "25.3",
+                topic="Adding the 'free' channel as it existed prior to conda 4.7.",
+                addendum="Add 'https://repo.anaconda.com/pkgs/free' to your channels "
+                "list manually after 'defaults', if you still need content from "
+                "the 'free' channel.",
+                deprecation_type=FutureWarning,
+            )
```
beeankha, line 880: Along with this deprecation marker, should we also add some `TODO` comments reminding us to remove (in conda in version 25.3) other mentions of the `restore_free_channel` in this file, e.g. https://github.com/conda/conda/blob/main/conda/base/context.py#L402, https://github.com/conda/conda/blob/main/conda/base/context.py#L1204 and https://github.com/conda/conda/blob/main/conda/base/context.py#L1737-L1742?

conda/base/context.py
```diff
@@ -869,6 +869,15 @@ def custom_multichannels(self):
             default_channels = list(self._default_channels)
 
         if self.restore_free_channel:
+            deprecated.topic(
+                "24.9",
+                "25.3",
+                topic="Adding the 'free' channel as it existed prior to conda 4.7.",
+                addendum="Add 'https://repo.anaconda.com/pkgs/free' to your channels "
+                "list manually after 'defaults', if you still need content from "
+                "the 'free' channel.",
```
jaimergp, line 878: Not sure if this is the way 🤔 Back then strict channel priority was not so popular so free was source for some other versions... This would work with flexible priority, but for strict priority maybe the recommendation should be to add free in next to the other repo.anaconda.com channels under the `default_channels` multichannel.

conda/base/context.py
jezdez: I am not convinced that's needed since we'll do a pass over the feature ahead of 25.3. I also find the TODO comments suspect since they are super hand wavy.

conda/base/context.py
jezdez: Do you have a suggested alternative?

jezdez: How about this?

```suggestion
                addendum="Add `https://repo.anaconda.com/pkgs/free` next to the "
                "other repo.anaconda.com hosted channel in the `default_channels` "
                "configuration option, if you still need content from the `free` "
                "channel.",
```

kenodegard: I'd in favor of adding this info to the free channel docs page and referencing that:

```suggestion
                addendum="See "
                "https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/free-channel.html "
                "for more details.",
```

jezdez: I've added bf903e9c5d290a145f93497b784c9216438d459b to handle this, does that suffice?

kenodegard (approved):

PR #13940, opened 2024-05-24
Print transaction summary in explicit installs

jaimergp:
### Description

When install via `conda create --file explicit.txt`, with `explicit.txt` having `@EXPLICIT` in it, you don't get much info in the terminal:

```
$ conda create -nx --file explicit.txt 
Downloading and Extracting Packages:


Downloading and Extracting Packages:

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
```

With this one-line PR, we print the usual summary we see in all commands:

```
$ conda create -nx --file explicit.txt 


Downloading and Extracting Packages:


## Package Plan ##

  environment location: /opt/conda/envs/x

  added / updated specs:
    - conda-forge/linux-aarch64::docutils==0.20.1=py312h8025657_3
    - conda-forge/linux-aarch64::expat==2.6.2=h2f0025b_0
    - conda-forge/linux-aarch64::fmt==10.2.1=h2a328a1_0
    - conda-forge/linux-aarch64::gettext==0.21.1=ha18d298_0
    - conda-forge/linux-aarch64::icu==73.2=h787c7f5_0
    - conda-forge/linux-aarch64::libarchive==3.7.2=hd2f85e0_1
    - conda-forge/linux-aarch64::libexpat==2.6.2=h2f0025b_0
    - conda-forge/linux-aarch64::libgcc-ng==13.2.0=hf8544c7_5
    - conda-forge/linux-aarch64::libgomp==13.2.0=hf8544c7_5
    - conda-forge/linux-aarch64::libiconv==1.17=h31becfc_2
    - conda-forge/linux-aarch64::libmamba==1.5.7=hea3be6c_0
    - conda-forge/linux-aarch64::libmambapy==1.5.7=py312h1e39527_0
    - conda-forge/linux-aarch64::libnsl==2.0.1=h31becfc_0
    - conda-forge/linux-aarch64::libsqlite==3.45.2=h194ca79_0
    - conda-forge/linux-aarch64::libstdcxx-ng==13.2.0=h9a76618_5
    - conda-forge/linux-aarch64::libuuid==2.38.1=hb4cce97_0
    - conda-forge/linux-aarch64::libxcrypt==4.4.36=h31becfc_1
    - conda-forge/linux-aarch64::libxml2==2.12.6=h3091e33_0
    - conda-forge/linux-aarch64::libzlib==1.2.13=h31becfc_5
    - conda-forge/linux-aarch64::minio-server==2024.03.10.02.53.48=hcefe29a_0
[...]

PR #14271, opened 2024-09-24
Remove CreateNonAdminAction

marcoesters:
### Description

The .nonadmin file is created by constructor-based installers to indicate that the installed application does not require admin privileges. This is used by the Windows uninstaller for constructor-based installers (e.g., Miniconda, Miniforge) and menuinst.

When removing a package via `conda remove`, the .nonadmin file is removed. This will require users to have admin rights for subsequent calls of menuinst or when running the uninstaller. If a user does not have those privileges, those calls will fail.

The culprit is the [`CreateNonadminAction`](https://github.com/conda/conda/blob/main/conda/core/path_actions.py#L596-L619) which is added to the linking transaction: https://github.com/conda/conda/blob/main/conda/core/link.py#L1144-L1151

This adds the .nonadmin file to the paths list for each package and is removed during `conda remove`. Since it is an environment-level file, however, the .nonadmin file should not be touched by the installation transactions. https://github.com/conda/conda/pull/13940 has recently added all necessary `touch_nonadmin` commands to environment creations, so this action is obsolete.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2666,3 +2666,20 @@ def test_create_dry_run_without_prefix(
 
 def test_create_without_prefix_raises_argument_error(conda_cli: CondaCLIFixture):
     conda_cli("create", "--json", "ca-certificates", raises=ArgumentError)
+
+
+def test_nonadmin_file_untouched(
+    conda_cli: CondaCLIFixture,
+    tmp_env: TmpEnvFixture,
+):
+    channel = Path(__file__).parent / "test-recipes" / "noarch"
+    with tmp_env() as prefix:
+        nonadmin_file = prefix / ".nonadmin"
+        nonadmin_file.touch()
+        assert nonadmin_file.is_file()
+        conda_cli(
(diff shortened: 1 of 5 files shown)


jezdez (approved):

tests/test_create.py
```diff
@@ -2663,3 +2663,20 @@ def test_create_dry_run_without_prefix(
 
 def test_create_without_prefix_raises_argument_error(conda_cli: CondaCLIFixture):
     conda_cli("create", "--json", "ca-certificates", raises=ArgumentError)
+
+
+def test_nonadmin_file_untouched(
```
travishathaway, line 2668: Should this be a windows only test?

marcoesters: The MacOS and Linux installers create the `.nonadmin` file, too, even though it's not propagated into other environments. This is why I am testing this on all platforms.

However, if you think that this test is redundant for non-Windows platforms, I can restrict it to Windows.

travishathaway (approved):

PR #14282, opened 2024-09-30
Add protected directories guardrail to `conda create` command

beeankha:
### Description

A "protected directories" guardrail for the `conda create` command was moved out of `conda/cli/install.py` and into `conda/cli/main_rename.py` while doing work for issue https://github.com/conda/conda/issues/13386; this PR adds the guardrail back in to `conda/cli/main_create.py`.

### Checklist - did you ...


tests/cli/test_main_install.py
```diff
@@ -62,9 +62,9 @@ def test_install_mkdir(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
...
 
-        conda_cli("create", f"--prefix={dir}", "--yes")
-        conda_cli("install", f"--prefix={dir}", "python", "--mkdir", "--yes")
-        assert package_is_installed(dir, "python")
+        conda_cli("create", f"--prefix={prefix}", "--yes")
+        conda_cli("install", f"--prefix={prefix}", "python", "--mkdir", "--yes")
+        assert package_is_installed(prefix, "python")
 
         rm_rf(prefix, clean_empty_parents=True)
         assert path_is_clean(dir)
```
beeankha, line 70: I'm not sure if this `test_install_mkdir` is necessary/actually testing what it's supposed to. The regression test for #4849 seems to require specific versions to be listed in order to ensure that the regression is being tested, but this would mean that intermittent updating of this test would be required over time as some specific versions of packages may no longer be available from certain channels, etc.

kenodegard: This test is supposed to be for the `--mkdir` argument which has been slated for deprecation and will be removed in 25.3.

Looking at the original implementation I agree that this no longer is testing for the regression:

https://github.com/conda/conda/commit/1985160dac27c4c05d079015e65a8b63141b87aa#diff-1e27c53a97d6c9242ea92544453665b553fe9fdf63f08717dc915f80e76fd60cR928-R937

I'd opt to remove the regression comment and leave this test as is since it'll be removed next year anyhow.

tests/cli/test_main_install.py
```diff
@@ -62,14 +62,13 @@ def test_install_mkdir(tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture):
         ):
             conda_cli("install", f"--prefix={dir}", "python", "--mkdir", "--yes")
 
-        conda_cli("create", f"--prefix={dir}", "--yes")
-        conda_cli("install", f"--prefix={dir}", "python", "--mkdir", "--yes")
-        assert package_is_installed(dir, "python")
+        conda_cli("create", f"--prefix={prefix}", "--yes")
+        conda_cli("install", f"--prefix={prefix}", "python", "--mkdir", "--yes")
+        assert package_is_installed(prefix, "python")
```
beeankha, line 67: This part of the test has been updated to no longer point to `prefix / "conda-meta"`.

PR #14341, opened 2024-10-23
Make `conda doctor` accept `$REQUESTS_CA_BUNDLE` that point to a valid file

muffato:
### Description

Hello,

On my system, `REQUESTS_CA_BUNDLE` is a local path. It makes `conda doctor` complain:

```console
$ echo $REQUESTS_CA_BUNDLE 
/etc/ssl/certs/ca-certificates.crt
$ conda doctor
(...)
❌ The following error occured while verifying `REQUESTS_CA_BUNDLE`: Invalid URL 'h': No scheme supplied. Perhaps you meant https://h?
```

It looks like #14037 made `conda doctor` expects either an invalid path, or a URL. In this PR, I make it allow valid paths

### Checklist - did you ...

### DIFF
--- a/conda/plugins/subcommands/doctor/health_checks.py
+++ b/conda/plugins/subcommands/doctor/health_checks.py
@@ -6,13 +6,13 @@ from __future__ import annotations
 
 import json
 import os
-import re
 from logging import getLogger
 from pathlib import Path
 
 from requests.exceptions import RequestException
 
 from ....base.context import context
+from ....common.url import is_url
 from ....core.envs_manager import get_user_environments_txt_file
 from ....exceptions import CondaError
 from ....gateways.connection.session import get_session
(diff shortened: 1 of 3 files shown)


conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -159,7 +160,9 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
     requests_ca_bundle = os.getenv("REQUESTS_CA_BUNDLE")
     if not requests_ca_bundle:
         return
-    elif not Path(requests_ca_bundle).exists():
+    elif Path(requests_ca_bundle).exists():
+        print(f"{OK_MARK} `REQUESTS_CA_BUNDLE` was verified.\n")
+    elif re.match("^https?://", requests_ca_bundle) is None:
```
kenodegard, line 165: This results in the `$REQUESTS_CA_BUNDLE` not being validated (see the `else` clause). We need something like the following instead:

```suggestion
    elif not (
        Path(requests_ca_bundle).exists()
        or re.match("^https?://", requests_ca_bundle)
    ):
```

We also have a `conda.common.url.is_url` convenience function that can be used:

```suggestion
    elif not (
        Path(requests_ca_bundle).exists()
        or is_url(requests_ca_bundle)
    ):
```

kenodegard (changes requested):
@muffato thanks for taking a look at this!

conda/plugins/subcommands/doctor/health_checks.py
muffato: Thanks for showing me `is_url`. I've added it in the latest commit. I've also changed the order of the `if` to hopefully make things more straightforward:

1. Is it a URL ? Yes → defer to `requests`
2. Otherwise, does the file exist ? Yes → we're good
3. Otherwise (i.e. the file doesn't exist); show an error

muffato:
Closing this pull-request as I had misinterpreted the problem. Will create a new one in a few minutes

Issue #14355, opened 2024-10-30
env create gives FutureWarning: `remote_definition` is deprecated


raybellwaves:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I ran `conda env create -f environment-dev.yml -n myenv` and got

```
/opt/conda/lib/python3.12/argparse.py:2006: FutureWarning: `remote_definition` is deprecated and will be removed in 25.9. Use `conda env create --file=URL` instead.
  action(self, namespace, argument_values, option_string)
```

It also seems this message has popped up in the docs https://docs.conda.io/projects/conda/en/latest/commands/env/create.html under positional argumation

### Conda Info

```
(base) root@850a544f31cb:/# conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
                          /root/.condarc
          conda version : 24.9.0
    conda-build version : 24.5.1
         python version : 3.12.7.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=x86_64
                          __conda=24.9.0=0
                          __glibc=2.31=0
                          __linux=6.6.51=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : HIDDEN
                          https://conda.anaconda.org/nodefaults/linux-64
                          https://conda.anaconda.org/nodefaults/noarch
                          https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/24.9.0 requests/2.32.3 CPython/3.12.7 Linux/6.6.51-0-virt ubuntu/20.04.6 glibc/2.31 solver/libmamba conda-libmamba-solver/24.9.0 libmambapy/1.5.9
                UID:GID : 0:0
             netrc file : /root/.netrc
           offline mode : False
```
### Conda Config
```
channels:
  - conda-forge

==> /root/.condarc <==
channels:
  - https://HIDDEN.jfrog.io/artifactory/api/conda/conda/dtn
  - https://HIDDEN.jfrog.io/artifactory/api/conda/conda
  - nodefaults
show_channel_urls: True

[...]

PR #14359, opened 2024-10-31
Update `deprecated.action()` to account for positional arguments with no value specified

beeankha:
Resolves https://github.com/conda/conda/issues/14355

Every positional argument in the conda CLI is either required or defaults to `None` or `[]`, so [checking to see if `values` exists](https://github.com/conda/conda/pull/14359/files#diff-55373dab266c863a6f3a85e6ed1700a8e532ca7c027f6650093e18fd0cb66c4dR229) avoids the activation of an unnecessary deprecation warning.

### DIFF
--- a/tests/test_deprecations.py
+++ b/tests/test_deprecations.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import sys
-from argparse import ArgumentParser, _StoreAction
+from argparse import ArgumentParser, _StoreAction, _StoreTrueAction
 from contextlib import nullcontext
 from typing import TYPE_CHECKING
 
@@ -139,11 +139,15 @@ def test_action(
         parser = ArgumentParser()
         parser.add_argument(
             "--foo",
+            action=deprecated.action("2.0", "3.0", _StoreTrueAction),
+        )
+        parser.add_argument(
+            "bar",
             action=deprecated.action("2.0", "3.0", _StoreAction),
         )
 
         with pytest.warns(warning, match=message):
-            parser.parse_args(["--foo", "some_value"])
+            parser.parse_args(["--foo", "bar"])
 
 
 @parametrize_dev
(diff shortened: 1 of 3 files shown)


jezdez (approved):

tests/test_deprecations.py
```diff
@@ -139,11 +139,11 @@ def test_action(
         parser = ArgumentParser()
         parser.add_argument(
             "--foo",
-            action=deprecated.action("2.0", "3.0", _StoreTrueAction),
+            action=deprecated.action("2.0", "3.0", _StoreAction),
         )
```
kenodegard, line 143: I think we should explicitly test this as a positional and leave the `--foo` as it was:

```suggestion
        )
        parser.add_arguemnt(
            "bar",
            action=deprecated.action("2.0", "3.0", _StoreAction),
        )
```

tests/test_deprecations.py
```diff
@@ -141,9 +141,13 @@ def test_action(
...
             action=deprecated.action("2.0", "3.0", _StoreTrueAction),
         )
+        parser.add_argument(
+            "bar",
+            action=deprecated.action("2.0", "3.0", _StoreAction),
+        )
 
         with pytest.warns(warning, match=message):
-            parser.parse_args(["--foo"])
+            parser.parse_args(["--foo", "bar"])
```
kenodegard, line 150: This will only test the `--foo` warning so we need to add a second case:

```suggestion
            parser.parse_args(["--foo"])

        with pytest.warns(warning, match=message):
            parser.parse_args(["bar"])
```

beeankha: The `--foo` flag passed without a value won't trigger a deprecation warning since the warning only comes up for marked flags that contain a value, thus the `"some_value"` addition to that line.

kenodegard (approved):

PR #14370, opened 2024-11-05
pre-commit.ci fixes

beeankha:
Resolving pre-commit.ci failures that came up after resolving some merge conflicts.

### DIFF
--- a/conda/reporters.py
+++ b/conda/reporters.py
@@ -8,11 +8,11 @@ from __future__ import annotations
 
 import logging
 import sys
-from functools import lru_cache
+from functools import cache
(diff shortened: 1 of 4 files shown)


conda/reporters.py
```diff
@@ -25,7 +25,7 @@
 logger = logging.getLogger(__name__)
 
 
-@lru_cache(maxsize=None)
+@cache
```
beeankha, line 28: This returns the same as `@lru_cache(maxsize=None)` ([reference](https://docs.python.org/3/library/functools.html#functools.cache))

ForgottenProgramme (approved):

PR #14382, opened 2024-11-12
Make request headers dynamic

### LINKED ISSUE #14381: Adjust `conda_request_headers` plugin hook to be more dynamic
### Checklist

### What is the idea?

Make `conda_request_headers` more dynamic where we can filter headers based on the full URL and not just the host/netloc/domain.

### Why is this needed?

The `conda_request_headers` plugin hook was implemented at the session level, thus the headers themselves are cached with the `CondaSession`. This means the headers are only fine tuned to the URL's host/netloc/domain and doesn't offer additional tuning (e.g., fetching `repodata.json` versus a package URL).

### What should happen?

Refactor/move the `conda_request_headers` plugin hook closer to the request itself instead of attaching it to the session.

### Additional Context

Xref https://github.com/conda/conda/pull/14325

kenodegard:
### Description

Closes #14381 

### Checklist - did you ...

### DIFF
--- a/conda/plugins/hookspec.py
+++ b/conda/plugins/hookspec.py
@@ -17,7 +17,6 @@ import pluggy
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
-    from ..common.url import Url
     from .types import (
         CondaAuthHandler,
         CondaHealthCheck,
@@ -396,14 +395,12 @@ class CondaSpecs:
         """
 
     @_hookspec
-    def conda_request_headers(
-        self, method: str, url: Url
-    ) -> Iterable[CondaRequestHeader]:
+    def conda_session_headers(self, host: str) -> Iterable[CondaRequestHeader]:
         """
         Register new HTTP request headers
 
         The example below defines how to add HTTP headers for all requests
-        with the hostname of ``example.com``
+        with the hostname of ``example.com``.
 
         **Example:**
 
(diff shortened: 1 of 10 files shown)


conda/plugins/hookspec.py
```diff
@@ -410,11 +413,14 @@ def conda_request_headers(self) -> Iterable[[CondaRequestHeader]]:
...
            @plugins.hookimpl
-           def conda_request_headers():
-               yield plugins.CondaRequestHeader(
-                   name="Example-Header",
-                   description="This is an example HTTP header",
-                   value="example",
-                   hosts={"example.com", "sub.example.com"},
-               )
+           def conda_request_headers(method: str, url: conda.common.url.Url):
+               if url.scheme in ("https", "http") and url_parts.netloc in {
```
jezdez, line 417:
```suggestion
               if url.scheme in ("https", "http") and url.netloc in {
```

conda/plugins/hookspec.py
```diff
@@ -410,11 +413,14 @@ def conda_request_headers(self) -> Iterable[[CondaRequestHeader]]:
...
-               )
+           def conda_request_headers(method: str, url: conda.common.url.Url):
+               if url.scheme in ("https", "http") and url.netloc in {
+                   "example.com",
+                   "sub.example.com",
+               }:
+                   yield plugins.CondaRequestHeader(
+                       name="Example-Header",
+                       value="example",
+                   )
```
jezdez, line 424: I think this conditional yielding is something new in the plugin API that would be good to explore first a little, as a required first-level API, at least. To me, this smells like the type of API that might backfire since it would wholesale allow plugins to add headers without defining for what endpoints.

I thought we'd define `scheme`, `netloc` etc as attributes of the conda request header class, to be validated/checked against later in the process if the request location would match that. We could alternatively use `conda.common.url.Url` as the domain object and use a dunder comparison to decide if the session request location equals the request header URL. But, seeing how the `Url.__str__` is just using string interpolation for recreating the URL, makes me worried TBH, not even using `unparse` :-/

kenodegard: > I think this conditional yielding is something new in the plugin API that would be good to explore first a little, as a required first-level API, at least.

FYI these hook args are opt-in only (https://pluggy.readthedocs.io/en/stable/#opt-in-arguments), dunno if that helps us much here but something interesting to keep in mind.

travishathaway (dismissed):
I had a chance to test this out with a plugin I'm currently developing and am feeling pretty good about these changes. I would be happy to accept it as-is.

@jezdez, I do understand your concern about the new usage of arguments for these hooks, but I do believe it makes more efficient use of the pluggy API. Otherwise, we would have to define some sort of callable to attach to the `CondaRequestHeader` object and that wouldn't be as elegant and require more boilerplate code.

beeankha (dismissed):

jezdez (approved):

conda/plugins/hookspec.py
```diff
@@ -395,26 +395,58 @@ def conda_reporter_backends():
...
-                   hosts={"example.com", "sub.example.com"},
-               )
+           def conda_request_headers(host: str):
+               if host in HOSTS:
+                   yield plugins.CondaRequestHeader(
+                       name="Example-Header",
+                       value="example",
+                   )
+        """
+        yield from ()
```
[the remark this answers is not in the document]
kenodegard: this is to satisfy `mypy`, AFAIK the contents of this `hookspec` is irrelevant to `pluggy`

jezdez (dismissed):

jezdez (approved):

Issue #13015, opened 2023-08-22
Combine `conda env create` with `conda create` and make the former an alias


jezdez:
We desire consistent behavior between `conda create` and `conda env create`.

Some of the inconsistencies between the two commands:

| feature | `conda create` | `conda env create` |
|---|---|---|
| list dependencies on CLI | ✅ | ❌ |
| `requirements.txt` | ✅ | ✅ |
| `environment.yaml` | ❌ | ✅ |
| remote `environment.yaml` | ❌ | ✅ |
| specify additional channels (`--channel=CHANNEL`) | ✅ | ❌ |
| override existing channels (`--override-channels`) | ✅ | ❌ |

beeankha:
Connecting https://github.com/conda/conda/pull/14113 here

jaimergp:
xref https://github.com/conda/conda/pull/14446

danyeaw:
Not stale.

janxkoci:
I need to create an environment using a yml file, but conda does not let me - is there any workaround? Version 25.11.0, installed last week.

PS: sorry for the multiple messages, GitHub moving to Azure has some glitches, I had to use phone because GitHub didn't have free server to log me in on a laptop - ridiculous!

PPS: I see errors as in https://github.com/conda/conda/issues/13676. The workaround is to use mamba.

GF-Huang:
Issue still 2026.

janxkoci:
Yes, the functionality is currently still broken. However, my understanding is that only YML support is broken, while other [specification formats](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#sharing-an-environment), like txt or json, may work.

Or as I said above - you can use mamba with YML files.

PR #14446, opened 2024-12-16
Deprecate `-f` as alias of `--force` in favor of `--file`

jaimergp:
### Description

In the future, we will fold `conda env create` under `conda create`. `conda env create` uses `-f` for file paths, but `conda create` only had `--file` because `-f` was being used for the suppressed / deprecated `--force` flag. If it's true that the `env` folding is not ready yet, we can already start the flag migration process here to have it ready when time comes.

Part of https://github.com/conda/conda/issues/13015

### Checklist - did you ...

### DIFF
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -70,6 +70,7 @@ def add_parser_create_install_update(p, prefix_required=False):
     # Add the file kwarg. We don't use {action="store", nargs='*'} as we don't
     # want to gobble up all arguments after --file.
     p.add_argument(
+        # "-f",  # FUTURE: 25.7: Enable this after deprecating alias in --force
         "--file",
         default=[],
         action="append",
@@ -449,7 +450,7 @@ def add_parser_package_install_options(p: ArgumentParser) -> _ArgumentGroup:
         "Package Linking and Install-time Options"
     )
     package_install_options.add_argument(
-        "-f",
+        "-f",  # FUTURE: 25.7: Remove here and add as alias to --file
         "--force",
         action="store_true",
         default=NULL,
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -11,6 +11,7 @@ conda.cli.main_remove for the entry points into this module.
 from __future__ import annotations
 
 import os
+import sys
 from logging import getLogger
 from os.path import abspath, basename, exists, isdir, isfile, join
 from pathlib import Path
@@ -31,6 +32,7 @@ from ..core.index import (
 from ..core.link import PrefixSetup, UnlinkLinkTransaction
 from ..core.prefix_data import PrefixData
 from ..core.solve import diff_for_unlink_link_precs
+from ..deprecations import deprecated
 from ..exceptions import (
     CondaEnvException,
     CondaExitZero,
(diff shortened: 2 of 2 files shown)


conda/cli/helpers.py
```diff
@@ -449,7 +450,7 @@ def add_parser_package_install_options(p: ArgumentParser) -> _ArgumentGroup:
         "Package Linking and Install-time Options"
     )
     package_install_options.add_argument(
```
kenodegard, line 452: The deprecation you want is `25.9` and `26.3` (I just opened a PR to add the deprecation formula to the [docs](https://github.com/conda/conda/pull/14447)).

We also have a `deprecated.action` helper:

```suggestion
    package_install_options.add_argument(
        "-f",
        dest="force",
        action=deprecated.action(
            "25.9",
            "26.3",
            argparse._StoreTrueAction,
            addendum="Use `--force` instead.",
        ),
        default=False,
        help=SUPPRESS,
    )
    package_install_options.add_argument(
```

jaimergp: 15 months for this change seems too long in this specific case. This has been ˝functionally deprecated" (suppressed) for a long time now, so I thought we could skip the PendingDeprecation stage and kick the six-month period with 25.1.

Also this message is not accurate because `--force` does not have any effect anymore; users need to resort to `--force-reinstall` or `--yes`, depending on the intended semantics.

kenodegard: 1. it sounds like we should be deprecating `--force` not just trying to deprecate `-f`
2. if we feel that strongly about deprecating this we can go with 25.3 and 25.9?

jaimergp: About (1), I think we are doing two things: deprecating _some_  uses of `--force` (with install and update, but not with remove) and repurposing `-f`.

About (2), I still think 25.1 and 25.7 is generous enough, but I'll defer that to the broader team.

jezdez: I don't think we can soften the deprecation policy, since then, we start debating every deprecation, 2026 sounds far, but it's not really, it's just 12.5 months away

jaimergp: I would agree if `conda install -f numpy` had a different outcome than `conda install numpy` but these days it doesn't. Folks using `-f` would get a warning about the upcoming meaning change for six months, and I feel that's sufficient notice. 

I just want to avoid the weird time window where users can do `mamba create -f env.yml` but have to do `conda create --file env.yml` only to end up learning that six months later they'll be able to do `conda create -f env.yml` too.

conda/cli/helpers.py
```diff
@@ -444,12 +446,24 @@ def add_parser_networking(p: ArgumentParser) -> _ArgumentGroup:
...
         "-f",
+        dest="force",
+        action=deprecated.action(
+            "25.9",
+            "26.3",
+            _StoreTrueAction,
+            addendum="Use `--force` instead.",
+        ),
+        default=False,
```
kenodegard, line 463:
```suggestion
        default=NULL,
```

conda/cli/install.py
```diff
@@ -189,6 +191,14 @@ def install(args, parser, command="install"):
...
     if context.force_32bit and prefix == context.root_prefix:
         raise CondaValueError("cannot use CONDA_FORCE_32BIT=1 in base env")
+    if getattr(args, "force", None) and "-f" in sys.argv:
+        deprecated.topic(
+            "25.9.0",
+            "26.3.0",
+            topic="Deprecated -f flag will be repurposed as a --file alias.",
+            addendum="The -f alias for --force has not been used for long. In the future, "
+            "it will be repurposed as an alias for --file.",
+        )
```
kenodegard, line 201: isn't this already handled by the action deprecation?

jaimergp: True. Can be removed.

kenodegard: Finding some issue with the deprecation messages when testing locally, still debugging

kenodegard:
```suggestion
```

kenodegard:
pre-commit.ci autofix

Issue #14448, opened 2024-12-17
`CONDA_SUBDIR` causes wrong virtual package versions to be reported


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

```
$ docker run -it continuumio/miniconda3:24.9.2-0
# CONDA_SUBDIR=osx-64 conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 24.9.2
    conda-build version : not installed
         python version : 3.12.7.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=x86_64
                          __conda=24.9.2=0
                          __osx=12.7=0   # <<<< THIS IS WRONG
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : osx-64
             user-agent : conda/24.9.2 requests/2.32.3 CPython/3.12.7 Linux/5.15.49-linuxkit-pr debian/12.7 glibc/2.36 solver/libmamba conda-libmamba-solver/24.9.0 libmambapy/1.5.8 aau/0.4.4 c/. s/. e/.
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

This reports `__osx=12.7` but that's wrong, because there's no way we can know. It should be `0`. The 12.7 comes from `platform.version()`, which reports the underlying Debian version.

### Conda Info

```shell
active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 24.9.2
    conda-build version : not installed
         python version : 3.12.7.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=24.9.2=0
                          __glibc=2.36=0
                          __linux=5.15.49=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
[...]

Issue #13832, opened 2024-04-22
__osx virtual package version depends on what SDK CPython in the base environment was built with


jjhelmus:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

On the macOS of x86-64 the version of the `__osx` virtual package depend on what SDK CPython in the base environment was built with. Specifically it can either be >=10.16,<11.0 if CPython was built with macOS SDK <11 or >=11.0 if built with SDK >=11. This is a result of the 10.x compatibility rules that comes into play when reporting the OS version starting with Big Sur. For example see [Big Sur is both 10.16 and 11.0 – it’s official](https://eclecticlight.co/2020/07/21/big-sur-is-both-10-16-and-11-0-its-official/). 

This difference in versions for the `__osx` virtual packages make it difficult for package builders to determine an appropriate constraint to add when they need to describe the version of macOS their package is compatible with. This also introduced a hurdle for end users who may be blocked from installing packages that use have a dependency on `__osx>=11.0`.

As an example of this inconsistency:

``` 
❯ sw_vers
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224

❯ conda create --name test_py39_defaults --platform osx-64 python=3.9 conda --yes
...
❯ ./mc/envs/test_py39_defaults/condabin/conda info | grep __osx
                          __osx=10.16=0
❯ conda create --name test_py39_cf --platform osx-64 python=3.9 conda --override-channels --channel conda-forge --yes
❯ ./mc/envs/test_py39_cf/condabin/conda info | grep __osx
                          __osx=14.4.1=0
```

One potential solution for this issue would be to translate the versions >=10.16,<11.0 into it's >=11.0 equivalent. This could be tricky going forward as future version are not known AFAIK.

### Conda Info

```shell
❯ conda info

     active environment : base
    active env location : /Users/jhelmus/mc
            shell level : 1
       user config file : /Users/jhelmus/.condarc
 populated config files : /Users/jhelmus/.condarc
          conda version : 24.3.0
    conda-build version : 24.3.0
         python version : 3.10.13.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=24.3.0=0
                          __osx=14.4.1=0
                          __unix=0=0
       base environment : /Users/jhelmus/mc  (writable)
      conda av data dir : /Users/jhelmus/mc/etc/conda
  conda av metadata url : None
[...]

Issue #13178, opened 2023-10-05
`user-agent` on osx-64 shows incorrect MacOS version


pseudoyim:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

**Observed behavior:** When running `conda info` on an Intel x86_64 Mac (`osx-64`), the `user-agent` value for the OS version appears to be incorrect, consistently showing version number `10.16` - even though the machine clearly has a newer version of MacOS installed (the correct MacOS version can be seen when running the `sw_vers` command in a Mac terminal).

**Expected behavior:**  The OS version displayed on the `user-agent` line should match the `ProductVersion` value returned when running `sw_vers` in a Mac terminal.

### Conda Info

```shell
(see screenshots under "Additional Context")
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

The screenshot below shows commands run on an Intel x86_64 Macbook Pro, running MacOS 13.4.1
The user has installed [Anaconda3-2023.07-2-MacOSX-x86_64.pkg](https://repo.anaconda.com/archive/Anaconda3-2023.07-2-MacOSX-x86_64.pkg)

Observe:
- `user-agent` line from `conda info` displays `OSX/10.16` (incorrect)
- The `sw_vers` command displays `ProductVersion: 13.4.1` (correct)

<img width="1324" alt="Screen Shot 2023-09-20 at 1 28 58 PM" src="https://github.com/conda/conda/assets/9688260/401cdae5-6b21-40dc-b1a3-cbc5cfbdc826">

<br>
===================================================

While investigating the cause of this discrepancy, I found:
- [This StackOverflow post](https://stackoverflow.com/questions/65290242/pythons-platform-mac-ver-reports-incorrect-macos-version) was very helpful.
- [This](https://github.com/conda/conda/blob/e9870ccaaa1ee1c57ff38b075f308e6c051c72a7/conda/base/context.py#L1060-L1093) is how `conda` derives the MacOS version for the `user-agent` line. It uses Python's `platform.mac_ver()` function, and the CPython folks seem to be aware of how the `10.16` value can be returned on machines running MacOS >=11 (see comment on test [here](https://github.com/python/cpython/blob/5b8f0246834f211db0ea83b89277489abc2521ed/Lib/test/test_platform.py#L348-L349).
[...]

PR #14449, opened 2024-12-17
Virtual packages fixes for `__archspec`, `__osx` and `__linux`

jaimergp:
### Description

- Closes https://github.com/conda/conda/issues/14448
- Closes https://github.com/conda/conda/issues/13832

### Checklist - did you ...

### DIFF
--- a/conda/plugins/virtual_packages/linux.py
+++ b/conda/plugins/virtual_packages/linux.py
@@ -24,12 +24,13 @@ def conda_virtual_packages():
     # numeric component; note that this breaks version ordering for
     # development (`-rcN`) kernels, but that can be a TODO for later.
     dist_version = os.environ.get("CONDA_OVERRIDE_LINUX")
-    if not dist_version:
+    if dist_version is None:
         dist_name, dist_version = context.platform_system_release
         if dist_name != "Linux":
-            dist_version = ""
-    m = re.match(r"\d+\.\d+(\.\d+)?(\.\d+)?", dist_version)
-    yield CondaVirtualPackage("linux", m.group() if m else "0", None)
+            dist_version = "0"
+    if dist_version:
+        m = re.match(r"\d+\.\d+(\.\d+)?(\.\d+)?", dist_version)
+        yield CondaVirtualPackage("linux", m.group() if m else "0", None)
 
     libc_family, libc_version = linux_get_libc_version()
     if not (libc_family and libc_version):
--- a/conda/plugins/virtual_packages/linux.py
+++ b/conda/plugins/virtual_packages/linux.py
@@ -28,9 +28,8 @@ def conda_virtual_packages():
         dist_name, dist_version = context.platform_system_release
         if dist_name != "Linux":
             dist_version = "0"
-    if dist_version:
-        m = re.match(r"\d+\.\d+(\.\d+)?(\.\d+)?", dist_version)
-        yield CondaVirtualPackage("linux", m.group() if m else "0", None)
+    m = re.match(r"\d+\.\d+(\.\d+)?(\.\d+)?", dist_version)
+    yield CondaVirtualPackage("linux", m.group() if m else "0", None)
 
     libc_family, libc_version = linux_get_libc_version()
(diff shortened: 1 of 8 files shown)


conda/plugins/virtual_packages/linux.py
```diff
@@ -23,8 +23,11 @@ def conda_virtual_packages():
...
     # development (`-rcN`) kernels, but that can be a TODO for later.
-    _, dist_version = context.platform_system_release
-    dist_version = os.environ.get("CONDA_OVERRIDE_LINUX", dist_version)
+    dist_version = os.environ.get("CONDA_OVERRIDE_LINUX")
+    if dist_version is None:
+        dist_name, dist_version = context.platform_system_release
+        if dist_name != "Linux":
+            dist_version = "0"
     m = re.match(r"\d+\.\d+(\.\d+)?(\.\d+)?", dist_version)
     yield CondaVirtualPackage("linux", m.group() if m else "0", None)
```
jaimergp, line 32:
```suggestion
    if dist_version:
        yield CondaVirtualPackage("linux", m.group() if m else "0", None)
```

This would align behavior with the __osx and __glibc behavior, but there's a test protecting the original implementation. Why? Should we deprecate this and align behaviors?

travishathaway: I tried searching through GitHub to find examples of people doing this in the wild but was unsuccessful. That doesn't mean it doesn't happen, but I also think it gives us a valid reason for deprecating current behavior.

I personally cannot understand why if you were going to override the Linux version, you would leave it as `""` if you want it to appear as `0`. In this case, you should just set it to `0`.

travishathaway:
@jaimergp,

Looks like we just need to add `__archspec` as an expected value to the `test_subdir_override` test, right? From the changes you made, it looks like we want to this value to be appearing most of the time now.

jaimergp:
> From the changes you made, it looks like we want to have this value to be appearing most of the time now.

I had a little bug. `CONDA_OVERRIDE_ARCHSPEC=""` should hide `__archspec` as per current behaviour. This might not be the best idea, but for now I'm not changing that behavior. We'll have the [CEP](https://github.com/conda/ceps/pull/103) for that :)

conda/plugins/virtual_packages/linux.py
jaimergp: Yea, same 🤷 Maybe they didn't want to allow `__linux` hides. Or maybe it was an oversight 🤷 A conversation for the [CEP](https://github.com/conda/ceps/pull/103) :D

travishathaway (approved):

conda/plugins/virtual_packages/linux.py
jezdez: This smells like a truthiness regression when we ported the virtual package to the plugin interface: https://github.com/conda/conda/pull/12293 fixed something like that for `CONDA_OVERRIDE_GLIBC`

jezdez: I can't find a test that tested CONDA_OVERRIDE_LINUX before the plugin API port, have you found one in the [old code](https://github.com/conda/conda/pull/11854/files#diff-a2a8adb5655f8100ed16e96feed682b9d2982d2010503cf4976b48847a608930L175-L203)?

jaimergp: They were always creating `__linux` as long as the target subdir was `linux-*`. There was no override with `""`. The [original tests](https://github.com/beeankha/conda/blob/2651937da04b16199d2068999bfbde63d2d17fce/tests/core/test_index.py#L66-L85) only covered truthy overrides.

kenodegard (commented):
Can we consistently use either `os.getenv` xor `os.environ.get`.

And can we keep the if-clauses in the same order (e.g., do the `is None` check first, etc).

jaimergp:
@kenodegard, done. Also refactored a bit to include the fix https://github.com/conda/conda/issues/13178 for free (the user agent won't misreport either).

PR #14468, opened 2025-01-03
Fix recursion error in `conda.sh`'s `__conda_reactivate`

jaimergp:
### Checklist - did you ...

### DIFF
--- a/conda/shell/etc/profile.d/conda.sh
+++ b/conda/shell/etc/profile.d/conda.sh
@@ -29,7 +29,7 @@ __conda_activate() {
 
 __conda_reactivate() {
     # FUTURE: conda 25.9, remove this function
-    echo `__conda_reactivate` is deprecated and will be removed in 25.9. Use `__conda_activate reactivate` instead. 1>&2
+    echo "'__conda_reactivate' is deprecated and will be removed in 25.9. Use '__conda_activate reactivate' instead." 1>&2
     __conda_activate reactivate
 }
 


conda/shell/etc/profile.d/conda.sh
```diff
@@ -29,7 +29,7 @@ __conda_activate() {
 
 __conda_reactivate() {
     # FUTURE: conda 25.9, remove this function
-    echo `__conda_reactivate` is deprecated and will be removed in 25.9. Use `__conda_activate reactivate` instead. 1>&2
+    echo "'__conda_reactivate' is deprecated and will be removed in 25.9. Use '__conda_activate reactivate' instead." 1>&2
```
jjerphan, line 32: Thank you, Jaime. :pray:

I have one UX question based on https://github.com/conda-forge/miniforge/issues/700#issuecomment-2568526971.

If users of conda or mamba have their setup uses `__conda_reactivate`, will they know:
 - 1) why this warning is show
 - 2) where the warning is issued from
 - 3) how to resolve the problem (here replace instance of `__conda_reactivate` with `__conda_activate reactivate)

?

jaimergp: AFAICS this is just a standard error message, no fancy deprecation framework in use, etc. So if that information needs to be included, we need to add it manually. Something like this?

```suggestion
    echo "Warning(conda.sh): '__conda_reactivate' is deprecated and will be removed in 25.9. '__conda_activate reactivate' is the recommended replacement. Consider updating conda and/or mamba to make this warning disappear." 1>&2
```

I don't think we have a way of detecting how the shell function was called without introducing a bigger surface for potential problems.

jjerphan: LGTM!

jjerphan (approved):
Thank you!

hmaarrfk:
to be honest, who is this warning for? end users are going to bear the brunt of living with this. I would be more in favor of removing this entirely.

hmaarrfk:
We have already dealt with the downstream issues in mamba, and we now have mamba 1.5.12 that doesn't use this anymore.

hmaarrfk:
Trying to expedite this since it is really affecting (me!) and other conda-forge users https://github.com/conda-forge/conda-feedstock/pull/251

jaimergp:
> We have already dealt with the downstream issues in mamba, and we now have mamba 1.5.12 that doesn't use this anymore.

Were repodata patches submitted too to set the upper bounds as needed?

hmaarrfk:
Anyway,  this works, i suggest we take the path of least resistance.

Today mamba 1.5.12 + conda 2.11.NEW would not have warnings for end users.

which is what I am trying to ensure.

hmaarrfk:
> Were repodata patches submitted too to set the upper bounds as needed?

Even with the new structure of repo-data patching, it is really mentally exhausting for me to write them. I won't have time to write them.

jaimergp:
> Even with the new structure of repo-data patching, it is really mentally exhausting for me to write them. I won't have time to write them.

No probs, just wanted to check whether they were needed. Added https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/934

jezdez (approved):

jezdez:
@hmaarrfk I think we intend to remove this in the January release, do you have suggestions how to do this better?

hmaarrfk:
Please consider that a "late january release" is really late for what seems like a "lockup".

I already patched conda-forge's implementation since it was hitting many many users.

---
How to do it better

Do not let any of this delay this release.

I'm not sure if there is the concept of a "stack level" but I find that warnings in python that don't have stacklevel's set correctly are effectively useless. 

So in my mind, a user that sees this will be like "what I am supposed to do this information".

What we want them to do is to raise their voice with whoever is calling conda's commands for them.
In this case it is mamba. 

(My thesis was that mamba 1.5.12 already addressed the core of the problem, so it isn't so critical that this warning exists if it is to communicate with them)

PR #14475, opened 2025-01-07
CI: Set PYTHONPATH=CONDA_SOURCE_ROOT

jaimergp:
### Description

Observed in https://github.com/conda/conda/pull/14170. 

`tests/test_shell.py::test_stacking` and `tests/test_install.py::test_windows_entry_point` rely on calls to `conda` subprocesses executed via the entry point (e.g. `$PREFIX/bin/conda`). As per the Python docs, this way of running a script only adds the parent directory of the script to `sys.path`, _instead_ of the working directory (which is added when we call `python` prompt, `python -c "some code"` or `python -m module`).

The different `sys.path` without the working directory misses our cloned `conda` copy, finding the "production" copy (e.g. latest release as installed in the test environmetn). We are not overwriting that one with `pip install` in CI for reasons related to shell initialization, IIRC.

Anyway, the fix is to set PYTHONPATH in CI too, as we were doing in the dev scripts for local development. I also added a fallback to the tests themselves in case they are executed individually, plus a chunky comment.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -141,7 +141,7 @@ jobs:
         # Windows is sensitive to long paths, using `--basetemp=${{ runner.temp }} to
         # keep the test directories shorter
         env:
-          PYTHONPATH: "."
+          PYTHONPATH: .
         run: >
           pytest
           --cov=conda
@@ -247,7 +247,7 @@ jobs:
 
       - name: Run Tests
         env:
-          PYTHONPATH: "."
+          PYTHONPATH: .
         run: >
           pytest
           --cov=conda
@@ -338,7 +338,7 @@ jobs:
       - name: Run Benchmarks
         uses: CodSpeedHQ/action@513a19673a831f139e8717bf45ead67e47f00044 # v3.2.0
         env:
-          PYTHONPATH: "."
+          PYTHONPATH: .
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: $CONDA/envs/test/bin/pytest --codspeed
@@ -467,7 +467,7 @@ jobs:
 
       - name: Run memray
         env:
-          PYTHONPATH: "."
+          PYTHONPATH: .
         run: python -m pytest -k memray --memray
 
   # macos test suite
@@ -558,7 +558,7 @@ jobs:
 
       - name: Run Tests
         env:
-          PYTHONPATH: "."
+          PYTHONPATH: .
         run: >
           pytest
           --cov=conda
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,6 +140,8 @@ jobs:
       - name: Run Tests
         # Windows is sensitive to long paths, using `--basetemp=${{ runner.temp }} to
         # keep the test directories shorter
+        env:
+          PYTHONPATH: "."
         run: >
           pytest
           --cov=conda
@@ -244,6 +246,8 @@ jobs:
         run: conda list --show-channel-urls
 
       - name: Run Tests
+        env:
+          PYTHONPATH: "."
         run: >
           pytest
           --cov=conda
@@ -333,6 +337,8 @@ jobs:
 
       - name: Run Benchmarks
         uses: CodSpeedHQ/action@513a19673a831f139e8717bf45ead67e47f00044 # v3.2.0
+        env:
+          PYTHONPATH: "."
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: $CONDA/envs/test/bin/pytest --codspeed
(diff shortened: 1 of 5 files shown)


jaimergp:
cc @kenodegard

jezdez:
FWIW https://github.com/pytest-dev/pytest-env exists

jaimergp:
> FWIW [pytest-dev/pytest-env](https://github.com/pytest-dev/pytest-env?rgh-link-date=2025-01-07T13%3A44%3A17Z) exists

We need to set it to the project directory, not sure how `pytest-env` interprets `"."` 🤔

tests/test_install.py
```diff
@@ -126,6 +128,21 @@ def test_windows_entry_point():
...
+        # have not been overwritten with `pip install -e . --no-deps`. This doesn't happen
+        # in other tests because they run with the equivalent of `python -m conda`. However,
+        # this family of tests relies on conda (shell function) which calls conda (Python entry
+        # point). When a script is called this way, it bypasses the automatic "working directory
+        # is first on sys.path" behavior you find in `python -m` style calls. See
+        # https://docs.python.org/3/library/sys_path_init.html for details.
+        env = os.environ.copy()
+        env["PYTHONPATH"] = CONDA_SOURCE_ROOT
+    else:
+        env = os.environ
```
kenodegard, line 145: isn't this a use case for `monkeypatch.setenv`?

.github/workflows/tests.yml
```diff
@@ -140,6 +140,8 @@ jobs:
       - name: Run Tests
         # Windows is sensitive to long paths, using `--basetemp=${{ runner.temp }} to
         # keep the test directories shorter
+        env:
+          PYTHONPATH: .
         run: >
           pytest
```
kenodegard, line 146: AFAIK this is the equivalent to `python -m pytest` (see https://docs.pytest.org/en/stable/how-to/usage.html#calling-pytest-through-python-m-pytest), can we use that instead? I think `python -m` has greater precedence, will be less magical to new devs, and less likely to be reverted in the future (thinking about future Ken trying to cleanup and simplify stuff)?

```suggestion
        run: >
          python -m pytest
```

jaimergp: It won't have an effect for the subprocess that run `conda create` because those end up doing something akin to `python path/to/conda-entry-point` which doesn't inherit the parent `sys.path` nor adds `.` itself. So I think we do need `PYTHONPATH` :(

tests/test_install.py
jaimergp: We need it in the subprocess only if it's not set. I think the subprocess will inherit the parent env in Unix, but not in Windows. At least I remember a conversation with @marcoesters in constructor or menuinst where he was chasing a strange bug that ended up being due to `env` not being passed as expected.

.github/workflows/tests.yml
kenodegard: in that case this sounds more like something we can solve with an autouse fixture?

tests/test_install.py
kenodegard: hm can we track that down? per Python docs there is no difference between Unix and Windows re. env inheritance in subprocesses?

inline with the [autouse fixture suggestion](https://github.com/conda/conda/pull/14475#discussion_r1907591471) I'm thinking we only need to do the following:

```python
@pytest.fixture(scope="session", autouse=True)
def PYTHONPATH():
    if "PYTHONPATH" in os.environ:
        return

    with pytest.MonkeyPatch.context() as monkeypatch:
        monkeypatch.setenv("PYTHONPATH", CONDA_SOURCE_ROOT)
        yield
```

marcoesters: @jaimergp, I remember this vaguely. From what I can track down, the subprocess module handles copying explicitly: https://github.com/python/cpython/blob/3.12/Lib/subprocess.py#L1757-L1758

For Windows, it defers to `_winapi` where I can only find an explicit copy for non-`None` values: https://github.com/python/cpython/blob/3.12/Modules/_winapi.c#L1272-L1277

Otherwise, `wenvironment` is `NULL`. However, I do not know what is done under the hood when the process opens. I did notice a difference when setting `env` to `None` and explicitly copying the entire environment. 

However, I don't think I used `monkeypatch` back then. This is the PR in question: https://github.com/conda/constructor/pull/771

I'd say if monkeypatch works, use monkeypatch.

jaimergp: That autouse fixture is neat. Let me give it a try.

.github/workflows/tests.yml
jaimergp: Done!

kenodegard (approved):

Issue #14479, opened 2025-01-09
`reset_context` does not reinitialize plugin settings


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

As we have seen with this issue:

- https://github.com/anaconda/conda-anaconda-telemetry/issues/87

The `reset_context` function is used to reset the context. We currently remove the plugin settings initialized at runtime, but we fail to reinitialize them. This causes in conda-build, but only when it is being run for the first time.

### Conda Info

```shell
n/a
```

### Conda Config

```shell
n/a
```

### Conda list

```shell
n/a
```

PR #14480, opened 2025-01-09
Reinitializes plugin settings in `reset_context`

travishathaway:
### Description

Fixes: #14479 

### Checklist - did you ...

  - This was not tested because it is extremely difficult to test. During the development of this feature, I manually QA tested this according to the bug initially discovered here: https://github.com/anaconda/conda-anaconda-telemetry/issues/87 

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -15,7 +15,7 @@ import struct
 import sys
 from collections import defaultdict
 from collections.abc import Mapping
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from errno import ENOENT
 from functools import cache, cached_property
 from itertools import chain
@@ -1923,7 +1923,8 @@ def reset_context(search_path=SEARCH_PATH, argparse_args=None):
     from ..reporters import _get_render_func
 
     # reload plugin config params
-    context.plugin_manager.load_settings()
+    with suppress(AttributeError):
+        del context.plugins
 
     _get_render_func.cache_clear()
 
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1908,7 +1908,7 @@ class Context(Configuration):
 def reset_context(search_path=SEARCH_PATH, argparse_args=None):
     global context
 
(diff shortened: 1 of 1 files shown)


conda/base/context.py
```diff
@@ -1922,6 +1922,9 @@ def reset_context(search_path=SEARCH_PATH, argparse_args=None):
     # clear function cache
     from ..reporters import _get_render_func
 
+    # reload plugin config params
+    context.plugin_manager.load_settings()
+
```
jezdez, line 1927: This moves the plugin initialization much earlier in the process, since `reset_context` is called in `conda_build/__init__.py`, have you verified with manual QA that this doesn't trip up other plugin loading?

travishathaway: This doesn't cause any problems for me during manual QA testing. This is valid concern though.

I looked at the conda-build testing configuration. Is there a reason we don't use the canary build of conda over there? Seems like this is the type of issue that would be caught early there if we had that configured.

kenodegard: > I looked at the conda-build testing configuration. Is there a reason we don't use the canary build of conda over there? Seems like this is the type of issue that would be caught early there if we had that configured.

We do use canary conda builds for testing, e.g.:
- [linux (3.12, canary, serial)](https://github.com/conda/conda-build/actions/runs/12696494127/job/35390660203)
- [windows (3.12, canary, serial)](https://github.com/conda/conda-build/actions/runs/12696494127/job/35390656381)
- [macos (3.12, canary, serial)](https://github.com/conda/conda-build/actions/runs/12696494127/job/35390656990)

kenodegard: would it be safer to clear the cached plugin settings and allow them to be lazily reloaded?

```python
# clear cached plugin settings if set
with contextlib.suppress(AttributeError):
    del context.plugins
python

kenodegard (approved):

jezdez (approved):

PR #14504, opened 2025-01-20
Make use of GitHub's new Linux on ARM64 runner.

### LINKED ISSUE #11127: Failure to create (and run) post-link.sh on macos
### Checklist
* [x] I added a descriptive title
* [x] I searched for other issues and couldn't find a solution or duplication
* [x] I already searched in Google and didn't find any good information or help
* [x] I looked at the [docs](https://docs.conda.io/projects/conda/en/latest/user-guide/index.html) and didn't see anything to help

### What happened?
Test `test_post_link_run_in_env` will fail on OSX only because it cannot find the  `post-link.sh` script in `_conda_test_env_activated_when_post_link_executed`. Interestingly, Linux passes just fine. I checked and the contents of both linux and osx tarballs are the same. None actually have the `post-link.sh` script in the tree, but it's listed in the recipe. At some point `conda install` must be creating it on the spot, but it's unclear how. Technically, `conda-build` [is generating these](https://github.com/conda/conda-build/blob/0a6efdc/conda_build/build.py#L128) to begin with. 

So it could be:

* Packages are not built following the specs (but why does Linux work in that case?)
* The _package extraction_ logic where the file must be getting created (how?!) is different in Linux than OSX. Maybe not explicitly, but a case-sensitivity issue?

### Conda info
```shell
See CI in https://github.com/conda/conda/pull/10875
```

### Conda config
```shell
See CI in https://github.com/conda/conda/pull/10875
```

### Conda list
```shell
See CI in https://github.com/conda/conda/pull/10875
```

### Additional Context
_No response_

jezdez:
### Description

Resolves #11127

### Checklist - did you ...


dbast:
@jezdez Is there an added benefit to run __all__ linux tests both on linux64 and linux-aarch64? The test failures I looked at seem to be bugs in the tests (and not in conda), where e.g. test packages are missing etc. Something we should fix.

I am normally all for running as many tests, doing as much CI with extended matrix etc as possible, was there ever a bug that only existed on linux-X but not on linux-Y and vice versa? Won't this just burn lots of additional Action minutes of the GH org? Maybe it would make sense to run the different test-groups each on a different linux platform, but not on both (and a PR label enables running all on all)?

.github/workflows/tests.yml
```diff
@@ -270,6 +270,109 @@ jobs:
             test-report.xml
           retention-days: 1  # temporary, combined in aggregate below
 
+  # linux-arm64 test suite
+  linux-arm64:
```
kenodegard, line 274: is this a copy past from the `linux` job with minimal adjustments? if so I think we should just amend the `linux` job and add `ubuntu-24.04-arm` to the test matrix?

kenodegard: It can be: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability

Screenshot 2025-01-31 at 15 12 35

tests/test_create.py
```diff
@@ -2511,12 +2511,10 @@ def test_run_script_called(tmp_env: TmpEnvFixture):
...
-@pytest.mark.xfail(on_mac, reason="known broken; see #11127")
-def test_post_link_run_in_env(tmp_env: TmpEnvFixture):
-    test_pkg = "_conda_test_env_activated_when_post_link_executed"
-    # a non-unicode name must be provided here as activate.d scripts
-    # are not executed on windows, see https://github.com/conda/conda/issues/8241
-    with tmp_env(test_pkg, "--channel=conda-test") as prefix:
+@pytest.mark.integration
+def test_post_link_run_in_env(test_recipes_channel: Path, tmp_env: TmpEnvFixture):
+    test_pkg = "post_link_run_in_env_package"
+    with tmp_env(test_pkg) as prefix:
```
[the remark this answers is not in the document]
kenodegard: nice!

kenodegard (approved):

PR #14510, opened 2025-01-22
Add support for plugin settings to `conda config`

### LINKED ISSUE #13661: Add better integration of settings plugin hook into `conda config`
### Checklist

### What is the idea?

The newly created settings plugin hook allows plugin authors to create their own settings which will then appear alongside other conda settings on the context object under "plugins" (e.g. `conda.plugins.conda_plugin_setting`).

On the initial implementation, integration into `conda config` was overlooked. See:

- https://github.com/conda/conda/pull/13554

for more information.

### Why is this needed?

In order to provide a better user experience, we want it to be possible to set and display settings via the `conda config` command.

#### Scenario one: read configuration

If I have the following defined as settings:

```yaml
plugins:
  conda_plugin_setting: "some value"
```

I want to be able to run:

```
conda config --show
```

and this value in the output

#### Scenario two: write a setting

If a plugin has defined the setting `conda_plugin_setting` as a string parameter, I should be able to run the following command to set it:

```
conda config --set plugins_conda_plugin_setting "some value"
```

If a plugin has defined the setting `conda_plugin_setting` as a list of strings, I should be able to run the following command to set it:

```
conda config --set plugins_conda_plugin_setting "item_one,item_two,item_three"
```

### Additional Context

_No response_

travishathaway:
### Description

Fixes: #13661 

In addition to resolve the issue above, this pull request also performs a little refactoring of the current plugin settings code. These were necessary refactors to make the code work better with the `conda config` command and the settings system more generally.

### Checklist - did you ...

  - [x] We could add a special page in the documentation for "Plugin settings", but I don't believe it's necessary right now.
  - [x] Follow up issue created: https://github.com/conda/conda/issues/14511

### Note for reviewers

The `conda.cli.main_config` is definitely in need of a refactor. I tried my best to "spruce" things up a little while I was there, but I intentionally went the easier route while adding the plugin functionality to it because I would like to just ship something that works.


travishathaway:
Getting very close with this one. I think I only need to implement the correct behavior for `conda config --show` and  `conda config --show-sources`.

conda/base/context.py
```diff
@@ -2202,77 +2201,43 @@ def _first_writable_envs_dir():
     raise NoWritableEnvsDirError(context.envs_dirs)
 
 
+@deprecated(
+    "25.3",
+    "25.9",
+    addendum="Use `conda.base.context.plugins.raw_data` instead.",
```
kenodegard, line 2207:
```suggestion
    addendum="Use `conda.base.context.context.plugins.raw_data` instead.",
```

conda/plugins/config.py
```diff
@@ -0,0 +1,111 @@
...
+
+    When this class is updated via the :func:`add_plugin_setting` function it adds new setting
+    properties which can be accessed later via the context object.
+
+    We currently call that function in
+    :meth:`conda.plugins.manager.CondaPluginManager.load_settings`.
+    because ``CondaPluginManager`` has access to all registered plugin settings via the settings
+    plugin hook.
+    """
+
```
kenodegard, line 44: should we be initializing `parameter_names` here?

```suggestion

parameter_names: tuple[str, ...] = ()

```

conda/plugins/config.py
```diff
@@ -0,0 +1,111 @@
...
+    @classmethod
+    def add_plugin_setting(
+        cls, name: str, parameter: Parameter, aliases: tuple[str, ...] = ()
+    ):
+        """
+        Adds a setting to the :class:`PluginConfig` class
+        """
+        cls.parameter_names = cls.parameter_names + (name,)
```
kenodegard, line 52:
```suggestion
        cls.parameter_names = (*cls.parameter_names, name)
```

conda/base/context.py
```diff
@@ -2202,77 +2201,43 @@ def _first_writable_envs_dir():
     raise NoWritableEnvsDirError(context.envs_dirs)
 
 
+@deprecated(
+    "25.3",
+    "25.9",
```
kenodegard, line 2206: Don't want to skip the pending deprecation window:

```suggestion
    "25.9",
    "26.3",
```

conda/base/context.py
```diff
@@ -2202,77 +2201,43 @@ def _first_writable_envs_dir():
...
-    plugin hook.
-    """
-
-    def __init__(self, data):
-        self._cache_ = {}
-        self.raw_data = get_plugin_config_data(data)
 
+@deprecated(
+    "25.3",
+    "25.9",
```
kenodegard, line 2219:
```suggestion
    "25.9",
    "26.3",
```

conda/base/context.py
```diff
@@ -2202,77 +2201,43 @@ def _first_writable_envs_dir():
...
-    PluginConfig.parameter_names = PluginConfig.parameter_names + (name,)
-    loader = ParameterLoader(parameter, aliases=aliases)
-    name = loader._set_name(name)
-    setattr(PluginConfig, name, loader)
+    return PluginConfig.add_plugin_setting(name, parameter, aliases)
 
 
+@deprecated(
+    "25.3",
+    "25.9",
```
kenodegard, line 2234:
```suggestion
    "25.9",
    "26.3",
```

kenodegard (changes requested):

jezdez (dismissed):

conda/cli/main_config.py
```diff
@@ -276,12 +281,14 @@ def parameter_description_builder(name):
 
     if details["parameter_type"] == "primitive":
         builder.append(
-            "{} ({})".format(name, ", ".join(sorted({et for et in element_types})))
+            "{} ({})".format(
+                f"{name_prefix}{name}", ", ".join(sorted({et for et in element_types}))
+            )
```
kenodegard, line 286:
```suggestion
            f"{name_prefix}{name} ({', '.join(sorted(set(element_types)))})"
```

conda/cli/main_config.py
```diff
@@ -276,12 +281,14 @@ def parameter_description_builder(name):
...
         )
     else:
         builder.append(
             "{} ({}: {})".format(
-                name,
+                f"{name_prefix}{name}",
                 details["parameter_type"],
                 ", ".join(sorted({et for et in element_types})),
```
kenodegard, line 293:
```suggestion
                ", ".join(sorted(set(element_types))),
```

conda/cli/main_config.py
```diff
@@ -355,22 +375,45 @@ def print_config_item(key, value):
...
+    """
+    Logic to determine if the key is a valid setting.
+    """
+    if context is None:
+        from ..base.context import context
+
+    first, *rest = key.split(".")
+    if (
+        first == "plugins"
+        and len(rest) > 0
```
kenodegard, line 388:
```suggestion
        and rest
```

conda/cli/main_config.py
```diff
@@ -390,21 +433,31 @@ def _get_key(
...
+        return
 
-        raise CondaKeyError(key, "unknown parameter")
+    first, *rest = key.split(".")
+
+    if first == "plugins":
+        base_context = context.plugins
+        base_config = config.setdefault("plugins", {})
+        parameter_name = rest[0]
+        rest = rest[1:]
```
kenodegard, line 445:
```suggestion
        parameter_name, *rest = rest
```

conda/cli/main_config.py
```diff
@@ -390,21 +433,31 @@ def _get_key(
...
+    else:
+        base_context = context
+        base_config = config
+        parameter_name = first
+
+    parameter_type = base_context.describe_parameter(parameter_name)["parameter_type"]
+
+    if parameter_type == "primitive" and len(rest) == 0:
```
kenodegard, line 453:
```suggestion
    if parameter_type == "primitive" and not rest:
```

conda/cli/main_config.py
```diff
@@ -414,29 +467,43 @@ def _set_key(key: str, item: Any, config: dict) -> None:
...
         # KeyError: key_parts[0] is an unknown parameter
         from ..exceptions import CondaKeyError
 
         raise CondaKeyError(key, "unknown parameter")
 
-    if parameter_type == "sequence" and len(key_parts) == 1:
-        (key,) = key_parts
-        if key not in config:
-            if key != "channels":
+    if parameter_type == "sequence" and len(rest) == 0:
```
kenodegard, line 492:
```suggestion
    if parameter_type == "sequence" and not rest:
```

conda/cli/main_config.py
```diff
@@ -563,20 +630,46 @@ def execute_config(args, parser):
...
+        d["plugins"] = {}
+
+        # sort to make sure "plugins" appears in the right spot
+        d = {key: value for key, value in sorted(d.items())}
+
+        for key in provided_plugin_parameters:
+            value = getattr(context.plugins, key)
+            if isinstance(value, Mapping):
+                d["plugins"][key] = dict(value)
+            elif isinstance(value, tuple) and len(value) == 0:
```
kenodegard, line 668:
```suggestion
            elif isinstance(value, tuple) and not value:
```

conda/cli/main_config.py
```diff
@@ -610,24 +703,52 @@ def execute_config(args, parser):
...
+
+            if not_params or not_plugin_params:
                 from ..common.io import dashlist
                 from ..exceptions import ArgumentError
 
+                error_params = not_params | not_plugin_params
                 raise ArgumentError(
-                    f"Invalid configuration parameters: {dashlist(not_params)}"
+                    f"Invalid configuration parameters: {dashlist(error_params)}"
                 )
```
kenodegard, line 731: Isn't this a near duplicate of the above? Can we use a helper function instead?

conda/cli/main_config.py
```diff
@@ -610,24 +703,52 @@ def execute_config(args, parser):
...
+                        separators=(",", ": "),
+                        cls=EntityEncoder,
+                    )
+                )
+                stdout_write(
+                    json.dumps(
+                        [
+                            context.plugins.describe_parameter(name)
+                            for name in provided_plugin_parameters
+                        ],
```
kenodegard, line 751: I think this is the same thing?

```suggestion
                        [
                            *(
                                context.describe_parameter(name)
                                for name in provided_parameters
                            ),
                            *(
                                context.plugins.describe_parameter(name)
                                for name in provided_plugin_parameters
                            ),
                        ],
```

conda/cli/main_config.py
```diff
@@ -563,20 +630,46 @@ def execute_config(args, parser):
...
+            not_plugin_params = set(provided_plugin_parameters) - set(all_plugin_names)
+
+            if not_params or not_plugin_params:
                 from ..common.io import dashlist
                 from ..exceptions import ArgumentError
 
+                error_params = not_params | not_plugin_params
                 raise ArgumentError(
-                    f"Invalid configuration parameters: {dashlist(not_params)}"
+                    f"Invalid configuration parameters: {dashlist(error_params)}"
```
kenodegard, line 652: Isn't this list of `error_params` potentially confusing? Since we removed the `plugins.` prefix above wont it have non-plugin params and plugin params all jumbled together?

kenodegard (changes requested):
By trying to keep the refactor to a minimum I think we have inadvertently added new errors.

As an example, the following script (with a mocked context object) should pass but currently fails:

```python
from argparse import Namespace
from conda.cli.main_config import _key_exists

class Context(Namespace):
    def list_parameters(self):
        return self.parameter_names

    @property
    def parameter_names(self):
        return tuple(self.__dict__)

context = Context(
    foo="bar",
    plugins=Context(baz="qux"),
    json=False,
)

assert _key_exists("foo", [], context) # passes
assert _key_exists("plugins.baz", [], context) # passes

assert not _key_exists("missing", [], context) # passes
assert not _key_exists("plugins.missing", [], context) # fails
```

conda/cli/main_config.py
travishathaway: I don't agree with this suggestion because I believe that `len(rest) > 0` is easier to read than just evaluating its truthiness. When you use `len` you immediately know you're dealing with some sort of iterable data type, which makes the code easier to reason about.

I'm pretty sure there's no performance gain either and if there is one its pretty minimal.

travishathaway:
@kenodegard,

I appreciate the review, and you have importantly pointed to some gaps in coverage we have (more on that later). But, while the review is appreciated, I do have some problems with your example above. Please allow me to explain...

While I do concede that the above does fail, in your example, the `_key_exists` function is not being used in the correct context. On the `Context` class, parameters are special types of objects (`ParameterLoader`). Because of this, `plugins` will never be returned by the `list_parameters` method and when that's true, the `_key_exists` function works as expected.

A more realistic example for how the `Configuration` classes are being in conda would be the following:

```python
import os
from argparse import Namespace
from conda.cli.main_config import _key_exists
from conda.common.configuration import (
    Configuration, ParameterLoader, PrimitiveParameter
)
from conda.plugins.config import PluginConfig

class Plugins(PluginConfig):
    baz = ParameterLoader(PrimitiveParameter("default"))

class Context(Configuration):
    foo = ParameterLoader(PrimitiveParameter(True))
    json = ParameterLoader(PrimitiveParameter(True))

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self._set_env_vars("TESTER")

    @property
    def plugins(self):
        return Plugins(self.raw_data)

class Args(Namespace):
    foo: bool
    json: bool

# Plugin configuration values can only come from environment variables
# or config files
os.environ["TESTER_PLUGINS_BAZ"] = "something"

args = Args(foo=False, json=False)
context = Context(argparse_args=args)

assert context.plugins.baz == "something"
assert not context.foo
assert not context.json

assert _key_exists("foo", [], context) # passes
assert _key_exists("plugins.baz", [], context) # passes

assert not _key_exists("missing", [], context) # passes
assert not _key_exists("plugins.missing", [], context) # passes
```

After writing the above example, I realized that `_key_exists` is not being directly tested in our test suite. Thanks for helping point this out. To address this, I added the following test to our test suite. See the commit below for details:

- https://github.com/conda/conda/pull/14510/commits/df07755e21b43fab0196085c72fe6a2a6b55de60

I hope this can convince you that this pull request is ready to be merged. If you have any more concerns, I'm happy to hear them.

conda/cli/main_config.py
```diff
@@ -355,22 +375,45 @@ def print_config_item(key, value):
...
+
+    first, *rest = key.split(".")
+    if (
+        first == "plugins"
+        and len(rest) > 0
+        and rest[0] in context.plugins.parameter_names
+    ):
+        return True
+
+    if first not in context.list_parameters():
```
kenodegard, line 393: Why do we use `Configuration.parameter_names` for plugins but `Configuration.list_parameters()` for non-plugins?

travishathaway: We could switch it to use `list_parameters` instead. I'll do it.

travishathaway: This must have been an oversight because `list_parameters` was being used everywhere else in that module.

travishathaway: Address this in my latest commit.

kenodegard: Just a thought, does this mean that parameters cannot be looked up by their aliases?

tests/plugins/test_settings.py
```diff
@@ -138,3 +176,277 @@ def test_load_plugin_config_with_env_var(
...
+        "value_one",
+    )
+
+    if isinstance(expected_output, Exception):
+        with pytest.raises(ArgumentError, match=expected_output.message):
+            out, err, _ = conda_cli("config", "--show", f"plugins.{parameter_name}")
+    else:
+        out, err, _ = conda_cli("config", "--show", f"plugins.{parameter_name}")
+        assert not err
+        assert out == expected_output
```
kenodegard, line 431: There are existing patterns in our test suite of using `contextlib.nullcontext` for this kind of testing:

```suggestion
    with pytest.raises(ArgumentError, match=expected_output.message) if isinstance(expected_output, Exception) else nullcontext():
        out, err, _ = conda_cli("config", "--show", f"plugins.{parameter_name}")
        assert not err
        assert out == expected_output
```

tests/plugins/test_settings.py
```diff
@@ -138,3 +176,277 @@ def test_load_plugin_config_with_env_var(
...
+@pytest.mark.parametrize(
+    "parameter_name, expected_output",
+    [
+        (STRING_PARAMETER_NAME, "plugins:\n  string_parameter: value_one\n"),
+        (SEQ_PARAMETER_NAME, "plugins:\n  seq_parameter: []\n"),
+        (MAP_PARAMETER_NAME, "plugins:\n  map_parameter: {}\n"),
+        (
+            "non_existent_parameter",
+            ArgumentError(
+                "Invalid configuration parameters: \n  - non_existent_parameter"
```
kenodegard, line 398:
```suggestion
                "Invalid configuration parameters: \n  - plugins.non_existent_parameter"
```

tests/cli/test_config.py
```diff
@@ -464,8 +464,8 @@ def test_set_unconfigured_key(conda_cli: CondaCLIFixture):
 def test_set_invalid_key(conda_cli: CondaCLIFixture):
     key, to_val = "invalid_key", "a_bogus_value"
     with make_temp_condarc(CONDARC_BASE) as rc:
-        with pytest.raises(CondaKeyError, match=r"'invalid_key': unknown parameter"):
-            conda_cli("config", "--file", rc, "--set", key, to_val)
+        out, err, _ = conda_cli("config", "--file", rc, "--set", key, to_val)
+        assert err == f"Unknown key: '{key}'\n"
```
kenodegard, line 468: So an exception is no longer raised, does this mean the exit code is no longer non-zero?

tests/cli/test_main_config.py
```diff
@@ -281,3 +285,71 @@ def test_config_env_does_not_exist(
         conda_cli(
             "config", "--get", "channels", "--prefix", "ireallydontexist", "--json"
         )
+
+
+@pytest.mark.parametrize("is_json", [True, False], ids=["json_true", "json_false"])
```
kenodegard, line 290: I wouldn't include ids here. My rule is to only include ids if it is hard to figure out which test case is failing.

But if we are first using ids I think we should use `pytest.param` to keep the mapping to a minimum:

```suggestion
@pytest.mark.parametrize(
    "is_json",
    [
        pytest.param(True, id="is_json=True"),
        pytest.param(False, id="is_json=False"),
    ]
)
```

tests/plugins/test_settings.py
```diff
@@ -138,3 +176,277 @@ def test_load_plugin_config_with_env_var(
...
+    Ensure that string parameter types work correctly as a plugin setting
+    """
+    condarc = tmp_path / "condarc"
+    out, err, _ = conda_cli(
+        "config",
+        "--file",
+        condarc,
+        "--set",
+        f"plugins.{STRING_PARAMETER_NAME}",
+        "env_var_value",
```
kenodegard, line 192: Personally I like to do something like this to help keep the arguments tightly associated:

```suggestion
        f"--file={condarc}",
        *("--set", f"plugins.{STRING_PARAMETER_NAME}", "env_var_value"),
```

kenodegard (commented):
Some more questions, small test correction, and some optional nits

conda/cli/main_config.py
travishathaway: Will add a `_validate_provided_parameters` function.

conda/cli/main_config.py
travishathaway: Thanks for pointing this out. This was actually a latent bug which our test coverage wasn't addressing. I have changed it and updated the test coverage to make sure the JSON output is correctly output when using plugin parameters.

conda/cli/main_config.py
travishathaway: Yeah, that's true, but you currently cannot do this with normal parameters either, so I'm not worried about it. This is something that should be addressed with a separate pull request.

tests/cli/test_main_config.py
travishathaway: I'll just get rid of the IDs. It was an auto suggestion by co-pilot that I thought was worth keeping.

tests/plugins/test_settings.py
travishathaway: I see where you're coming from, but I would rather not do this.

tests/cli/test_config.py
travishathaway: Yes, I changed it to no longer raise and exception and return error code, and I will change it back to the way it was.

The reason I changed it originally is that our handling of error codes in `conda config` is inconsistent. For example, `conda config --get does_not_exist` does not return a non-zero error code when it probably should.

conda/cli/main_config.py
```diff
@@ -25,6 +25,8 @@
     from argparse import ArgumentParser, Namespace, _SubParsersAction
     from typing import Any
 
+    from conda.base.context import Context
```
kenodegard, line 28:
```suggestion
    from ..base.context import Context
```

kenodegard (dismissed):

jezdez (dismissed):

jezdez (approved):

PR #14558, opened 2025-02-04
Remove `git` & `bash` from CI dependencies

kenodegard:
### Description

We may not need `git` as an explicit CI dependency anymore. Per comments and `git blame`, `git` was necessary for `auxlib` to work. The relevant parts that depended on `git` appear to have been removed so we're gonna try to remove `git`.

### Checklist - did you ...


kenodegard:
hm the `cygpath` code is sensitive to removing `bash`/`cygpath` which are indirect dependencies of `git`

PR #14606, opened 2025-02-19
Ignore flaky stderr checks

kenodegard:
### Description

Recent testing has shown the `tests/test_create.py::test_json_create_install_update_remove` test to be very flaky on Linux ARM and results in the test failing something like 90% of the time. The failure occurs when under network connection retries which are printed to stderr. So the proposed change is to stop testing for `assert not stderr`.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -358,41 +358,38 @@ def test_safety_checks_disabled(
 
 
 def test_json_create_install_update_remove(
-    path_factory: PathFactoryFixture,
+    tmp_path: Path,
     conda_cli: CondaCLIFixture,
-    capsys: CaptureFixture,
 ):
     # regression test for #5384
 
-    def assert_json_parsable(content):
-        string = None
-        try:
-            for string in content and content.split("\0") or ():
+    def is_json_parsable(content: str) -> bool:
+        for string in content and content.split("\0") or ():
+            try:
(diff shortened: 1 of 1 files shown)


tests/test_create.py
```diff
@@ -358,135 +358,125 @@ def test_safety_checks_disabled(
...
     conda_cli: CondaCLIFixture,
-    capsys: CaptureFixture,
 ):
     # regression test for #5384
 
-    def assert_json_parsable(content):
-        string = None
-        try:
-            for string in content and content.split("\0") or ():
+    def is_json_parsable(content: str) -> bool:
```
ForgottenProgramme, line 366: Can we simply rename this test like that? 🤔 
Do we not need to deprecate it applying the deprecation policy?

ForgottenProgramme (dismissed):
The function is now changed to return `False` or `True` based on whether an exception occurs or not. 
So `assert is_json_parsable(stdout)` evaluates to False in case of an exception and we don't need to use `assert not stderr`.

tests/test_create.py
ForgottenProgramme: I think you mean it's a local func and inaccessible outside the test! :D 
Makes sense! 👍🏼

ForgottenProgramme (approved):

PR #14610, opened 2025-02-19
Remove `defaults` channel injection

kenodegard:
### Description

Removing `defaults` channel injection. Raise exception if no channels are defined.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -194,24 +194,6 @@ def ssl_verify_validation(value):
     return True
 
 
-def _warn_defaults_deprecation():
-    deprecated.topic(
-        "24.9",
-        "25.3",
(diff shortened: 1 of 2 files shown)


kenodegard:
Outstanding question: should we allow the channels list to be empty? or do we raise an exception?

jaimergp:
> should we allow the channels list to be empty? or do we raise an exception?

micromamba raises an exception in this case. I think that's fine. In practice, Miniconda/Miniforge will have configured the channels via `.condarc` so this should not impact most users.

OTOH, `conda` knows how to use the package cache as a "channel" when called in offline mode. So mayyybe in those cases it should allow an empty channel list (only if offline).

kenodegard:
Stale, closing

PR #14635, opened 2025-02-28
Stop relying on the name mangled double underscore attribute for dist string comparison.

jezdez:
### Description

This might help with the transient `test_solve_1` errors, maybe? 

### Checklist - did you ...

### DIFF
--- a/conda/testing/helpers.py
+++ b/conda/testing/helpers.py
@@ -732,7 +732,6 @@ def convert_to_dist_str(solution):
     return tuple(prec.dist_str(canonical_name=False) for prec in solution)
 
 
-
 @pytest.fixture()
 def solver_class():
     return context.plugin_manager.get_solver_backend()


conda/testing/helpers.py
```diff
@@ -729,15 +729,19 @@ def get_solver_cuda(
...
-        prec.channel._Channel__canonical_name = canonical_name
-    return tuple(dist_str)
+    return tuple(dist_str(prec) for prec in solution)
+
+
+def dist_str(prec):
+    # This is needed to remove the local path prefix in the
+    # dist_str() calls, otherwise we cannot compare them
+    return "{}{}::{}-{}-{}".format(
+        prec.channel.name,
```
jaimergp, line 739:
```suggestion
        prec.channel.canonical_name,
```

I think we don't want to sidestep all [this logic here](https://github.com/conda/conda/blob/2fc3bc82d53dae164905b7042e0dc8d81346a07e/conda/models/channel.py#L214-L253).

kenodegard: as I read it the `prec.channel.canonical_name` was not used, it was overridden with `prec.channel.name` and restored after calling `prec.dist_str()`

jaimergp: The consequence of this change is that using `name` on `repo.anaconda.com/pkgs/main` will render as `pkgs/main/<subdir>::...` instead of `defaults`. If that's intended, then ok.

kenodegard: With the following test script:

```python
from conda.models.records import PackageRecord
from conda.models.channel import Channel
from conda.testing.helpers import convert_to_dist_str

for channel in map(Channel, ("defaults", "https://repo.anaconda.com/pkgs/main")):
    print(f"{channel.name=}")
    print(f"{channel.canonical_name=}")

    prec = PackageRecord(
        channel=channel,
        subdir=context.subdir,
        name="conda",
        version="25.1.1",
        build="py39hca03da5_0",
        build_number=0,
    )

    try:
        print(f"{convert_to_dist_str([prec])=}")
    except AttributeError as err:
        print(f"{err=}")
    print()
```

I get the following on `main`:

```
channel.name='defaults'
channel.canonical_name='defaults'
err=AttributeError("'MultiChannel' object has no attribute '_Channel__canonical_name'")

channel.name='pkgs/main'
channel.canonical_name='defaults'
convert_to_dist_str([prec])=('pkgs/main/osx-arm64::conda-25.1.1-py39hca03da5_0',)
```

Versus this PR:

```
channel.name='defaults'
channel.canonical_name='defaults'
convert_to_dist_str([prec])=('defaults/osx-arm64::conda-25.1.1-py39hca03da5_0',)

channel.name='pkgs/main'
channel.canonical_name='defaults'
convert_to_dist_str([prec])=('pkgs/main/osx-arm64::conda-25.1.1-py39hca03da5_0',)
```

AFAIK this is a safe change to make and also explains why `PackageRecord.dist_str` was not used since it uses `canonical_name`. I think I'd be in favor of modifying `PackageRecord.dist_str` to remove redundancy:

```
def dist_str(self, canonical_name: bool = True) -> str:
    return "{}{}::{}-{}-{}".format(
        self.channel.canonical_name if canonical_name else self.channel.name,
        ("/" + self.subdir) if self.subdir else "",
        self.name,
        self.version,
        self.build,
    )
```

kenodegard (approved):

Issue #12137, opened 2022-12-01
Add `--name/-n` option to `conda config`


maximlt:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

`conda config` can be used to configure an **activated** environment with `conda config --env ...`. I believe this is a cool feature of conda. However, as I said the environment needs to be activated which poses some problem, I think, when conda is controlled programmatically through its command-line interface as you need to fiddle with the shell (I'd be happy to be proven wrong on this point!).

Adding `--name/-n` to `conda config` means that it would no longer be required to activate the environment to configure it. It would also make conda's CLI more consistent, as `--name/-n` is an option of `conda create/install/update/remove(and more?)`.

### Why is this needed?

The HoloViz group has a developer tool called `pyctdev` that is meant to be a unified interface for managing a Python project (e.g. `doit test_unit`, `doit build_package`, `doit build_docs`, ...). The user can either choose conda or pip as their dependency manager, i.e. `pyctdev` knows how to parse and convert the dependencies listed in a setup.py file so that they can be installed with conda. We're trying to see [how we can improve this tool](https://github.com/pyviz-dev/pyctdev/issues/103) and being able to configure an environment without activating it would be a cool feature.

### Additional Context

I tried to look for related issues but looking for `conda config` just returns hundred of issues, sorry if this is a duplicate!

travishathaway:
Hi @maximlt,

This is a very reasonable pull request that actually shouldn't be too hard to implement. To this, we really just need to figure out which configuration file we need to edit. That's exactly what this code here does:

https://github.com/conda/conda/blob/main/conda/cli/main_config.py#L223-L233

The good thing is that we are already have the functions to determine if a particular environment exists.

On top of of `-n/--name`, I would also recommend that we add `-p/--prefix` to stay consistent with the rest of the commands.

travishathaway:
As a temporary work around, you could implement something like this:

```bash
# Do this from the base environment
export MY_ENV_NAME='my-cool-env'
conda config --append channels cool-new-channel --file "$CONDA_PREFIX/envs/$MY_ENV_NAME/.condarc"
```

This will effectively do the same thing granted you are run this from a base environment.

bleakprestiger:
Can I Take This Issue ?

maximlt:
@bleakprestiger sure!

harivamsi9:
is anyone working on this issue rn?
if not I would like to contribute

[...]

PR #14648, opened 2025-03-05
Allow user to specify `--name` or `--prefix` as part of `conda config` command

soapy1:
### Description

Fixes https://github.com/conda/conda/issues/12137

This PR introduces two new flag options to the `conda config` command:
* `--name` allows users to specify the name of an environment to apply the config to
* `--prefix` allows the user to specify the prefix to apply the config to

For example, if a user wants to add a new channel to their environment named `myenv` they may do:
```
$ conda config --append channels my-new-channel -n myenv
```

This will update the `.condarc` file in `<path to conda deir>/envs/myenv/.condarc`.

For invalid environment names or prefixes provided, an error will be returned to the user:

```
$ conda config --get channels --prefix /i/dont/exist

EnvironmentLocationNotFound: Not a conda environment: /i/dont/exist
```

### Checklist - did you ...


conda/cli/main_config.py
```diff
@@ -120,6 +120,21 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         ),
     )
     location.add_argument("--file", action="store", help="Write to the given file.")
+    location.add_argument(
```
soapy1, line 123: Calling out that I added these here instead of using the [`add_parser_prefix` helper](https://github.com/conda/conda/blob/main/conda/cli/helpers.py#L128). Not sure if this is the right way to go here.

The motivation was to group these flags with the other `Config File Location Selection` flags. So, with this approach when you do a `conda config --help` you see

```
Config File Location Selection:
  Without one of these flags, the user config file at '/home/sophia/.condarc' is used.

  --system              Write to the system .condarc file
                        at '/home/sophia/projects/conda/de
                        venv/Linux/x86_64/envs/devenv-
                        3.10-c/.condarc'.
  --env                 Write to the active conda
                        environment .condarc file (/home/s
                        ophia/projects/conda/devenv/Linux/
                        x86_64/envs/devenv-3.10-c). If no
                        environment is active, write to
                        the user config file
                        (/home/sophia/.condarc).
  --file FILE           Write to the given file.
                        Write to the .condarc file for the
                        specified environment name.s
  -p PATH, --prefix PATH
                        Write to the .condarc file for the
                        specified environment prefix.
```

Opposed to using the `add_parser_prefix` helper which would produce
```
Config File Location Selection:
  Without one of these flags, the user config file at '/home/sophia/.condarc' is used.

  --system              Write to the system .condarc file
                        at '/home/sophia/projects/conda/de
                        venv/Linux/x86_64/envs/devenv-
                        3.10-c/.condarc'.
  --env                 Write to the active conda
                        environment .condarc file (/home/s
                        ophia/projects/conda/devenv/Linux/
                        x86_64/envs/devenv-3.10-c). If no
                        environment is active, write to
                        the user config file
                        (/home/sophia/.condarc).
  --file FILE           Write to the given file.

                        Write to the .condarc file for the
                        specified environment name.s
  -p PATH, --prefix PATH
                        Write to the .condarc file for the
                        specified environment prefix.
```

conda/cli/main_config.py
```diff
@@ -676,6 +693,11 @@ def execute_config(args, parser):
             rc_path = user_rc_path
     elif args.file:
         rc_path = args.file
+    elif args.prefix or args.name:
+        if is_conda_environment(context.target_prefix):
```
soapy1, line 697: Not sure if this is too strict? Should a user be able to specify a `prefix` or `environment` that does not exist?

conda/cli/main_config.py
```diff
@@ -120,6 +120,21 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         ),
     )
     location.add_argument("--file", action="store", help="Write to the given file.")
+    location.add_argument(
+       "-n",
+        "--name",
+        action="store",
+        help="Write to the .condarc file for the specified environment name.s",
```
dholth, line 127:
```suggestion
        help="Write to the .condarc file for the specified environment name.",
```

conda/cli/main_config.py
```diff
@@ -676,6 +693,11 @@ def execute_config(args, parser):
             rc_path = user_rc_path
     elif args.file:
         rc_path = args.file
+    elif args.prefix or args.name:
+        if is_conda_environment(context.target_prefix):
+            rc_path = join(context.target_prefix, ".condarc")
+        else:
+            raise EnvironmentLocationNotFound(context.target_prefix)
```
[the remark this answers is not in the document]
dholth: thanks for testing this case, it looks like codecov doesn't mark it as covered for unknown reasons.

conda/cli/main_config.py
kenodegard: I think it is important to reuse `add_parser_prefix` for consistency. I also think the mutex offered by adding this directly here is important (we loose this by using `add_parser_prefix`).

So I think we should do a small edit to `add_parser_prefix` and support passing in either a `ArgumentParser` or a `_MutuallyExclusiveGroup`.

soapy1: Hmmm, ya the mutex bit is important.

>So I think we should do a small edit to add_parser_prefix and support passing in either a ArgumentParser or a _MutuallyExclusiveGroup

Landed on extracting a `add_parser_prefix_to_group` function. The existing `add_parser_prefix` uses this new function to add the `--name` and `--prefix` arguments so it's functionality doesn't change in any way. And then this also allows other mutually exclusive argument groups to pull in these options.

kenodegard:
pre-commit.ci autofix

news/14648-conda-config-name-and-prefix-option
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Allow user to specify --name or --prefix as part of conda config command. (#14648)
```
jaimergp, line 7:
```suggestion
* Allow user to specify --name or --prefix as part of conda config command. (#12137 via #14648)
```

kenodegard (approved):

PR #14670, opened 2025-03-13
Refactor `repodata_fn` iteration in `conda.cli.install` and smaller cleanups

soapy1:
### Description
The goal of this PR is to refactor `conda.cli.install` by:
* de-duplicating validation
* moving validation of command line argument combinations to the cli command layer
* moving appropriate validations to the cli command layer
* reduce the responsibilities of the `conda.cli.install.install` function

This refactor will make it easier to add new behaviours to the different installation workflows. In particular, taking steps to have yaml files supported by the `conda create --file` command (ref: https://github.com/conda/conda/pull/14113)

### Checklist - did you ...


conda/cli/common.py
```diff
@@ -263,3 +267,51 @@ def validate_prefix(prefix):
         raise EnvironmentLocationNotFound(prefix)
 
     return prefix
+
+
+def validate_subdir_config():
+    """Validates that the configured subdir is ok. A subdir that is different from
+    the native system is only allowed if it comes from the global configration, or
```
jaimergp, line 274:
```suggestion
    the native system is only allowed if it comes from the global configuration, or
```

conda/cli/common.py
```diff
@@ -263,3 +267,51 @@ def validate_prefix(prefix):
...
+        # its subdir to a newer env.
+        context_sources = context.collect_all()
+        if context_sources.get("cmd_line", {}).get("subdir") == context.subdir:
+            pass  # this is ok
+        elif context_sources.get("envvars", {}).get("subdir") == context.subdir:
+            pass  # this is ok too
+        # config does not come from envvars or cmd_line, it must be a file
+        # that's ok as long as it's a base env or a global file
+        elif not paths_equal(context.active_prefix, context.root_prefix):
+            # this is only ok as long as it's base environment
```
jaimergp, line 293:
```suggestion
            # this is only ok as long as it's NOT base environment
```

If I'm reading correctly?

conda/cli/install.py
```diff
@@ -172,120 +172,144 @@ def get_revision(arg, json=False):
...
-                "Creating new environment for a non-native platform %s",
-                context.subdir,
-            )
+        # new environments should not have a prefix that exists
+        pass
     elif isdir(prefix):
-        delete_trash(prefix)
-        if not isfile(join(prefix, "conda-meta", "history")):
+        # if the prefix exists (and this is not a new environment)
+        # the conda-meta/history file must also exists - if it does not
```
jaimergp, line 195:
```suggestion
        # the conda-meta/history file must also exist - if it does not
```

conda/cli/install.py
```diff
@@ -172,120 +172,144 @@ def get_revision(arg, json=False):
...
+        # then this is not a valid conda environment
+        if isfile(join(prefix, "conda-meta", "history")):
```
jaimergp, line 197: https://github.com/conda/conda/pull/14668, aren't you checking if this file is writable too? Could do both here.

conda/cli/install.py
```diff
@@ -172,120 +172,144 @@ def get_revision(arg, json=False):
...
                 if not path_is_clean(prefix):
                     raise DirectoryNotACondaEnvironmentError(prefix)
-        else:
-            # fall-through expected under normal operation
-            pass
     else:
-        raise EnvironmentLocationNotFound(prefix)
+        # if this is not a new env and the prefix does not exist, then
+        # there is no existing environment - this is an error
+        EnvironmentLocationNotFound(prefix)
```
jaimergp, line 209:
```suggestion
        raise EnvironmentLocationNotFound(prefix)
```

conda/cli/install.py
```diff
@@ -172,120 +172,144 @@ def get_revision(arg, json=False):
...
-        # Override defaults if they are specified at the command line
-        names = [MatchSpec(pkg).name for pkg in args_packages]
-        for default_package in context.create_default_packages:
-            if MatchSpec(default_package).name not in names:
-                args_packages.append(default_package)
 
-    context_channels = context.channels
-    index_args = {
+def get_index_args(args) -> dict[str, any]:
+    """Retruns a dict of args required for fetching an index
```
jaimergp, line 213:
```suggestion
    """Returns a dict of args required for fetching an index
```

conda/cli/install.py
```diff
@@ -479,6 +452,72 @@ def install(args, parser, command="install"):
     handle_txn(unlink_link_transaction, prefix, args, newenv)
 
 
+def install_revision(args, parser):
```
jaimergp, line 455: Since both `install` and `install_revision` end up with `handle_txn()`, would it make sense to architect this refactor in a way that `install*` functions return the transaction, and this is handled in a layer above?

conda/cli/install.py
soapy1: Yep! I think #14668 is scoped much more tightly and can be reviewed/merged before this. Will opt to pull those changes in when they get merged :fireworks:

conda/cli/install.py
soapy1: There is also `install_clone` to handle. This currently does not return a transaction, instead it executes the cloning action and prints the activation message. I think this can be refactored to be more in line with `install` and `install_revision`. 

However, I think it makes sense to me for the cli layer to deal with:
* validating that combinations of command line arguments are valid
* doing preliminary checks + validations to ensure that the whatever actions can be successful
* doing any post installation steps/validations/messaging

Then the install layer to deal with pulling together the appropriate plan for installation (given the command line arguments) and executing it.

How does that sound to you?

conda/cli/install.py
```diff
@@ -297,53 +360,20 @@ def install(args, parser, command="install"):
                     " packages \nconda create --help for details"
                 )
         if "@EXPLICIT" in specs:
+            # short circuit to installing explicit if explicit specs are provided
```
soapy1, line 363: This part of the code is kind of weird. If there is a mix of `--file`s that specify regular packages and explicit package then this will produce an error. For example, consider the following two environment files:

env1.txt
```
@EXPLICIT
https://repo.anaconda.com/pkgs/main/linux-64/_libgcc_mutex-0.1-main.conda
https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda
```

env2.txt
```
numpy
```

Running a `conda create` against these files individually works. Providing both files as arguments produces an error:
```
$ conda create -n testenv-mixed --file env.txt --file env2.txt 

ParseError: Could not parse explicit URL: file:///home/sophia/projects/conda/numpy
```

Is this the intended behaviour here?

Would it be better to more clearly raise an error that specifying explicit files/packages are not compatible with specifying packages? That would include any combination of:
* explicit specs (eg. package files or `--file` arguments that have `@EXPLICIT` in them
* packages specified as part of `conda create` or as an entry in a `--file` argument

soapy1:
pre-commit.ci autofix

conda/cli/install.py
soapy1: Linking https://github.com/conda/ceps/pull/79 which describes the difference between these 2 types of input files

jaimergp: I wouldn't allow mixing `@EXPLICIT` (or, in the future, any other lockfile) with regular txt files. They have different purposes.

Basically, I wouldn't want `conda` to even touch the solver when all it needs to do is download, extract and link a series of packages. In that regard, I'd separate input fils in two categories: solverless (lockfiles like `@EXPLICIT` and `conda-lock.yml`) and solvable (environment.yml, requirements.txt).

conda/cli/install.py
jaimergp: Yes, in general I'd like to have a streamlined way of running all necessary pre-checks, then creating a transaction object, then reporting and post-checks.

Creating the transaction object here is the key step, because not every installation process results in one right now. Clones, explicits, revisions, and regular `conda create|install|update|remove` operations should all result in a transaction. Once we have a transaction, then reporting can be unified.

jaimergp: I think that's what I mean by transaction is what you meant by "installation plan", so we are in agreement.

soapy1:
Looks like the failing tests are flaky ones.

Since this is a refactor PR I didn't add too many more tests. But very open to being more thorough about that if it seems like a good idea.

jaimergp:
Yea unfortunately libmamba exhibits some non-deterministic behavior on our CI tests that we haven't been able to identify or fix. A rerun usually fixes them though.

conda/cli/install.py
soapy1: Cool, this is also what I was thinking. I think this should not be addressed in this PR since it's a change in behaviour for conda. I'll open up an issue to raise an error if the user is mixing types of `file` arguments.

conda/cli/install.py
soapy1: Hmmm, ya I think we have a similar vision. I think this PR doesn't get all the way there. There is still more work to be done on aspects like collecting and parsing input files etc. I would advocate for punting on solving that fully for the purpose of this PR, so as not to explode the scope too much (I'd say the scope of this PR is to create a set of reusable validations, clarify the difference between different installation workflows and reduce the responsibility of `conda.cli.install.install`). But certainly address it in follow up work.

conda/cli/install.py
```diff
@@ -361,76 +399,65 @@ def install(args, parser, command="install"):
...
+
+    @retry_package_not_found(
+        repodata_fns=repodata_fns,
+        index_args=index_args,
+        retry_extra_errors=(
+            UnsatisfiableError,
+            SpecsConfigurationConflictError,
+            SystemExit,
+        ),
+    )
```
jaimergp, line 417: Having a decorated inline function is a bit strange. Would it be better to have `retry_package_not_found` as a context manager instead?

soapy1: ya, true! Giving a context manager a go

conda/cli/install.py
```diff
@@ -172,119 +171,222 @@ def get_revision(arg, json=False):
         raise CondaValueError(f"expected revision number, not: '{arg}'", json)
 
 
-def install(args, parser, command="install"):
-    """Logic for `conda install`, `conda update`, and `conda create`."""
+class TryRepodata:
```
dholth, line 174: Neat

soapy1:
pre-commit.ci autofix

conda/cli/install.py
soapy1: @jaimergp I've updated this part of the code to use a context manager approach. Does this look better to you? Will add some extra unit tests to ensure that its expressing the right behaviour if this is the direction we want to go.

soapy1:
pre-commit.ci autofix

soapy1:
Looks like these errors are not related to the changes from this PR

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+        # Other solvers might implement their own internal retry logic without
+        # depending --freeze-install implicitly like conda classic does. Example
+        # retry loop in conda-libmamba-solver:
+        # https://github.com/conda-incubator/conda-libmamba-solver/blob/da5b1ba/conda_libmamba_solver/solver.py#L254-L299
+        # If we end up raising UnsatisfiableError, we annotate it with `allow_retry`
+        # so we don't have go through all the repodatas and freeze-installed logic
+        # unnecessarily (see https://github.com/conda/conda/issues/11294). see also:
+        # https://github.com/conda-incubator/conda-libmamba-solver/blob/7c698209/conda_libmamba_solver/solver.py#L617
+        if (
+            (exc_type in self.allowed_errors)
```
jaimergp, line 206: This should be an `isinstance()` check I think?

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+            and (self.repodata != self.last_repodata)
+            and getattr(exc_value, "allow_retry", True)
+        ):
+            return True
+        elif exc_type == ResolvePackageNotFound:
```
jaimergp, line 211: Same here, use `isinstance()`?

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+
+class Repodatas:
+    def __init__(self, repodata_fns, index_args, allows_errors=()):
+        self.repodata_fns = repodata_fns
+        self.index_args = index_args
+        self.success = False
+        self.allowed_errors = (
+            ResolvePackageNotFound,
+            PackagesNotFoundError,
+        ) + allows_errors
```
jaimergp, line 233:
```suggestion
        self.allowed_errors = (
            ResolvePackageNotFound,
            PackagesNotFoundError,
            *allows_errors,
        )
```

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+    """
+    prefix_data = PrefixData(prefix)
+    for spec in specs:
+        spec = MatchSpec(spec)
+        if not spec.is_name_only_spec:
+            raise CondaError(
+                f"Invalid spec for 'conda update': {spec}\nUse 'conda install' instead."
+            )
+        if not prefix_data.get(spec.name, None):
+            raise PackageNotInstalledError(prefix, spec.name)
```
jaimergp, line 321: Same here.

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+
+
+def install(args, parser, command="install"):
+    """Logic for `conda install`, `conda update`, and `conda create`."""
+    prefix = context.target_prefix
+    newenv = bool(command == "create")
+    isupdate = bool(command == "update")
+    isinstall = bool(command == "install")
+    isremove = bool(command == "remove")
```
jaimergp, line 330: The `bool()` calls here are redundant 🤔

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+
+    # common validations for all types of installs
+    validate_install_command(prefix=prefix, newenv=newenv)
+
+    # this is sort of a hack.  current_repodata.json may not have any .tar.bz2 files,
+    #    because it deduplicates records that exist as both formats.  Forcing this to
+    #    repodata.json ensures that .tar.bz2 files are available
+    if context.use_only_tar_bz2:
+        args.repodata_fns = ("repodata.json",)
```
jaimergp, line 339: We should only override if the user didn't provide their own. e.g. it would be surprising if we do this when they provided `--repodata-fn=my-custom-repodata.json`.

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
+
+    # ensure trash is cleared from existing prefix
+    if isdir(prefix):
+        delete_trash(prefix)
```
jaimergp, line 343: Should we swallow exceptions here if there's an error while cleaning up? Maybe it shouldn't be fatal.

conda/cli/install.py
```diff
@@ -361,123 +408,122 @@ def install(args, parser, command="install"):
...
                     update_modifier=update_modifier,
                     force_reinstall=context.force_reinstall or context.force,
                     should_retry_solve=(
-                        _should_retry_unfrozen or repodata_fn != repodata_fns[-1]
+                        _should_retry_unfrozen or repodata != repodata_fns[-1]
                     ),
                 )
-            # we only need one of these to work.  If we haven't raised an exception,
-            #   we're good.
-            break
```
jaimergp, line 410: We are not breaking early if successful anymore?

conda/cli/main_install.py
```diff
@@ -137,4 +139,18 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             file=sys.stderr,
         )
 
-    return install(args, parser, "install")
+    # Ensure provided combination of command line argments are valid
```
jaimergp, line 142:
```suggestion
    # Ensure provided combination of command line arguments are valid
```

conda/cli/install.py
```diff
@@ -172,119 +170,201 @@ def get_revision(arg, json=False):
...
-            )
+        # new environments should not have a prefix that exists
+        pass
     elif isdir(prefix):
-        delete_trash(prefix)
-        if not isfile(join(prefix, "conda-meta", "history")):
+        # if the prefix exists (and this is not a new environment)
+        # the conda-meta/history file must also exist - if it does not
+        # then this is not a valid conda environment
+        if isfile(join(prefix, "conda-meta", "history")):
```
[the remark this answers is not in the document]
jaimergp: Also `conda-meta/history` is a constant: `PREFIX_MAGIC_FILE`.

jaimergp (changes requested):
I dropped a few comments as I read the changes. We are on the right track but two main things came to mind.

The first one: There are many prefix validations in different layers. 

If we are refactoring this part, it should be simpler. I know it was way more involved before, but right now I can see:
  - `conda.cli.main_create:execute`:  `check_prefix()`, `check_protected_dirs()`, several directory existence checks.
  - `conda.cli.main_install:execute`:  `validate_prefix()`
  - `conda.cli.main_update:execute`:  no `validate_prefix()`
  - `conda.cli.install:validate_install_command`: More directory checks, `conda-meta/history` existence, and whether is writable via `validate_prefix_is_writable`.
  - `conda.cli.main_remove` is not even in this PR, but there are a few prefix checks there as well. Same with the `conda.cli.main_env_*` family.
  
Seeing all these `check_prefix`, `validate_prefix`, `validate_prefix_is_writable`, `check_protected_dirs` utility functions, I wonder if we would be better off with a class that encompasses all these checks for a target directory. Many of these functions are overlapping and we could make more out of these system calls. Something like a `TargetPrefixValidator`.

---

The second one: the repodata loop refactor. It seems to me that is unrelated to the main goal of this PR, and could be better off on its own PR. If we do keep it here, we have to make sure that we are indeed breaking the repodata / unfreeze loop when successful. I think I saw a `break` clause being removed and I'm uncertain why.

soapy1:
Hmmm, I think it will be better to split this PR out into a few smaller, more focused changes. Giving that a go

conda/cli/install.py
soapy1: I think this would be a pretty big change in behaviour and not appropriate to be in a refactor PR. For reference, compare this to the main branch https://github.com/conda/conda/blob/main/conda/cli/install.py#L182

Can you create a seperate issue (or is this a bug?) to address this.

conda/cli/install.py
soapy1: We are. This part of the control flow is handled in the iterator/context manager bit. See https://github.com/conda/conda/pull/14670/files#diff-30852b2709103c10b41b1639d623f9d7eb8588f44f24ccf8cb35e8b4171866b0R238 and https://github.com/conda/conda/pull/14670/files#diff-30852b2709103c10b41b1639d623f9d7eb8588f44f24ccf8cb35e8b4171866b0R187

conda/cli/install.py
jaimergp: I'll raise an issue and assign you, thanks.

jaimergp: https://github.com/conda/conda/issues/14745 - can't assign you until you comment there, though.

jaimergp:
Opened https://github.com/conda/conda/issues/14747 to track the first bit in https://github.com/conda/conda/pull/14670#pullrequestreview-2742443805.

jaimergp:
This PR has been split in several ones down the line, so it's currently holding only the following changes:

- The `repodata_fn` iteration now uses a context manager driven loop.
- Some post-create tasks have moved to `conda.cli.main_create` (touch nonadmin file, print activate info).
- Some well defined checks have been refactored into its own little function for better readability.

Note that this PR interfaces a lot with https://github.com/conda/conda/pull/14750. I think we should merge this first, and then the other one.

jaimergp (approved):

PR #2537, opened 2016-05-23
conda 4.2 configuration engine

kalefranz:
Still TODO:

- [ ] Much more robust validation error messaging.
- [ ] Messaging on what parameters are coming from where.
- [ ] Further integrate `context` into cli execute logic (get rid of the need for using `args` so much).
- [ ] Get test coverage back up.
- [ ] `!top` and `!bottom` for lists
- [ ] Make environment variables splittable into lists (with a separator).
- [ ] Consider `!final`, `!enforce`, `!override` instead of `!important`.
- [ ] Put binstar token handling back in.
- [ ] Better docstring documentation of fresh code.

PR #14678, opened 2025-03-17
Define context.default_env

jaimergp:
### Description

Instead of relying on `${CONDA_DEFAULT_ENV}` directly, use it via the `context` system.

Make sure this is used in `conda activate` too, instead of the hardcoded `base` instance there.

This would allow installers to ship a non-base default environment that can be auto activated on shell initialization, preventing users from unintentionally installing things in `base`.

Comes from https://github.com/conda/conda/issues/11342#issuecomment-2474969605.

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -304,7 +304,7 @@ class _Activator(metaclass=abc.ABCMeta):
                     + str(remainder_args)
                     + "\n"
                 )
-            self.env_name_or_prefix = remainder_args and remainder_args[0] or "base"
+            self.env_name_or_prefix = remainder_args and remainder_args[0] or context.default_prefix
 
         else:
             if remainder_args:
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -304,7 +304,9 @@ class _Activator(metaclass=abc.ABCMeta):
                     + str(remainder_args)
                     + "\n"
                 )
-            self.env_name_or_prefix = remainder_args and remainder_args[0] or context.default_prefix
+            self.env_name_or_prefix = (
+                remainder_args and remainder_args[0] or context.default_prefix
+            )
 
         else:
             if remainder_args:
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1532,7 +1532,7 @@ class Context(Configuration):
                 The name of the environment to be used when no name or prefix is provided,
                 and no other environment is active. Defaults to '{ROOT_ENV_NAME}'.
                 """
-            )
+            ),
             default_threads=dals(
                 """
                 Threads to use by default for parallel operations.  Default is None,
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -221,11 +221,14 @@ class Context(Configuration):
     auto_update_conda = ParameterLoader(
         PrimitiveParameter(True), aliases=("self_update",)
     )
-    auto_activate_base = ParameterLoader(PrimitiveParameter(True))
+    auto_activate_base = ParameterLoader(
+        PrimitiveParameter(True), aliases=("auto_activate_default_env",)
+    )
     auto_stack = ParameterLoader(PrimitiveParameter(0))
     notify_outdated_conda = ParameterLoader(PrimitiveParameter(True))
     clobber = ParameterLoader(PrimitiveParameter(False))
     changeps1 = ParameterLoader(PrimitiveParameter(True))
+    default_env = ParameterLoader(PrimitiveParameter(ROOT_ENV_NAME, element_type=str))
     env_prompt = ParameterLoader(PrimitiveParameter("({default_env}) "))
     create_default_packages = ParameterLoader(
         SequenceParameter(PrimitiveParameter("", element_type=str))
@@ -747,7 +750,7 @@ class Context(Configuration):
     def default_prefix(self):
         if self.active_prefix:
             return self.active_prefix
-        _default_env = os.getenv("CONDA_DEFAULT_ENV")
+        _default_env = self.default_env
         if _default_env in (None, ROOT_ENV_NAME, "root"):
             return self.root_prefix
         elif os.sep in _default_env:
@@ -1269,6 +1272,7 @@ class Context(Configuration):
                 "auto_activate_base",
                 "auto_stack",
                 "changeps1",
+                "default_env",
                 "env_prompt",
                 "json",
                 "console",
@@ -1391,7 +1395,7 @@ class Context(Configuration):
             ),
             auto_activate_base=dals(
                 """
-                Automatically activate the base environment during shell initialization.
+                Automatically activate the default (base) environment during shell initialization.
                 """
             ),
             auto_update_conda=dals(
@@ -1523,6 +1527,12 @@ class Context(Configuration):
             #     the version used by conda itself.
             #     """
             # ),
+            default_env=dals(
+                f"""
+                The name of the environment to be used when no name or prefix is provided,
+                and no other environment is active. Defaults to '{ROOT_ENV_NAME}'.
+                """
+            )
             default_threads=dals(
                 """
                 Threads to use by default for parallel operations.  Default is None,
--- a/news/14666-default-env
+++ b/news/14666-default-env
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Add `context.default_env` option, previously only available via `CONDA_DEFAULT_ENV` environment variable. (#14666)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


jaimergp:
We'll also need to override this hardcoded `base` reference in the Docker images: https://github.com/anaconda/docker-images/blob/85cd4168dbbeada3713135044d6d03744cb99235/miniconda3/debian/Dockerfile#L60

jaimergp:
I'm having trouble with `CONDA_DEFAULT_ENV` being set by `conda activate`, which gets in the way of the context env var sourcing.

This env var was added to the context in https://github.com/conda/conda/pull/2537) with no further context. But before that it was already used by the activation engine for some reason; the details are lost in https://github.com/conda/conda/pull/5044.

My understanding is that `CONDA_DEFAULT_ENV` is used as a fallback for `context.default_prefix`. This is different from `.active_prefix` and `.target_prefix`. Let's see:

- `context.target_prefix`: this is the clearer one. The environment we will operate on.
- `context.active_prefix`: whatever `$CONDA_PREFIX` says it is. Set by activation engine.
- `context.default_prefix`: If `active_prefix` is set, that one. If not, `$CONDA_DEFAULT_ENV` dictates the returned value. If absent, it is `base`.

What I don't get is why we need `CONDA_DEFAULT_ENV` if `CONDA_PREFIX` is already there. It seems like `CONDA_DEFAULT_ENV` stores the name of the environment in the prompt, and that it needs to be kept around for stacked activation? 🤔

jaimergp:
The other theory is that is used to simply export the name of the currently active environment so other tools can check that. Whether that was intentional or not, [it's being used as such pretty popularly](https://github.com/search?q=CONDA_DEFAULT_ENV&ref=opensearch&type=code) so we won't be able to use `default_env` as the context variable for sure.

Back to the drawing board.

jjhelmus:
I like this approach. Should the hook parameter be renamed as well to `auto_active` from `auto_activate_base`. If the old name is needed for compatibility is it reasonable to add `auto_activate = auto_activate_base` to the hook method and use the new name?

conda/activate.py
```diff
@@ -194,12 +194,10 @@ def hook(self, auto_activate_base: bool | None = None) -> str:
             builder.append(preamble)
         if self.hook_source_path:
             builder.append(self.hook_source_path.read_text())
-        if (
-            auto_activate_base is None
-            and context.auto_activate_base
-            or auto_activate_base
-        ):
-            builder.append("conda activate base\n")
+        if auto_activate_base is None and context.auto_activate or auto_activate_base:
```
jjhelmus, line 197:
```suggestion
        if auto_activate is None and context.auto_activate or auto_activate:
```
And either rename the method parameter or create a new variable in the block with the preferred name

jaimergp:
Correct! The only problem with that approach is that a bare `conda activate` still activates `base` instead of the configured default activation environment. Only a minor nit though.

conda/base/context.py
```diff
@@ -1510,6 +1529,13 @@ def description_map(self):
                 a list of channel names and/or channel urls.
                 """
             ),
+            default_activation_env=dals(
```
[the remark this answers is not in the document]
jaimergp: We cannot use `default` env because `CONDA_DEFAULT_ENV` is a reserved env var used during activation, holding the name of the activated environment. I tried and failed terribly (see first round of commits).

It's mostly about which environment gets activated by default by either a bare `conda activate` or via `context.auto_activate = True`. Not sure what other "default envs" one can think of, because for example with `conda install` you need to either have an active environment or you pass `-n/-p`.

conda/activate.py
```diff
@@ -189,17 +189,19 @@ def reactivate(self):
         )
 
     def hook(self, auto_activate_base: bool | None = None) -> str:
+        # NOTE: auto_activate_base was renamed to auto_activate in 25.5.
+        # We didn't rename it in the signature for backwards compatibility.
+        # They are otherwise equivalent.
+        auto_activate = auto_activate_base
```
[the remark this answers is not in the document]
jaimergp: Oh, `deprecated.argument()` handles most of the complexity here. Easier than expected. Yea let's do that.

conda/shell/condabin/conda_auto_activate.bat
jaimergp, line 1: Please double check these changeset here 😬 Everytime I need to use `FOR` in Batch for `$(...)` equivalents my heart hurts.

conda/activate.py
jaimergp: Ok went for it and deprecated the old one.

conda/shell/condabin/conda_auto_activate.bat
```diff
@@ -3,6 +3,9 @@
 :: Helper script to conda_hook.bat.  This is a separate script so that the Anaconda Prompt
 :: can enable this option while only conda_hook.bat can be used in AutoRun.
 
-@FOR /F "delims=" %%i IN ('@CALL "%CONDA_EXE%" config --show auto_activate_base') DO @SET "__conda_auto_activate_base=%%i"
-@IF NOT "x%__conda_auto_activate_base:True=%"=="x%__conda_auto_activate_base%" @CALL "%CONDA_BAT%" activate base
-@SET __conda_auto_activate_base=
+@FOR /F "delims=" %%i IN ('@CALL "%CONDA_EXE%" config --show auto_activate') DO @SET "__conda_auto_activate=%%i"
+@FOR /F "delims=" %%i IN ('@CALL "%CONDA_EXE%" config --show default_activation_env') DO @SET "__conda_auto_activate_name=%%i"
+@IF NOT "x%__conda_auto_activate:True=%"=="x%__conda_auto_activate%" @CALL "%CONDA_BAT%" activate "%__conda_auto_activate_name:default_activation_env: =%
```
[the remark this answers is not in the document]
jaimergp: Correct, thanks for pointing it out!

conda/shell/condabin/conda_auto_activate.bat
```diff
@@ -3,6 +3,9 @@
...
+)
```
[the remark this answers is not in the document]
jaimergp: Oops, leftover. Removing.

conda/activate.py
```diff
@@ -188,18 +188,17 @@ def reactivate(self):
             self._yield_commands(self.build_reactivate()), self.tempfile_extension
         )
 
-    def hook(self, auto_activate_base: bool | None = None) -> str:
+    @deprecated.argument("25.5", "26.1", "auto_activate_base", rename="auto_activate")
```
kenodegard, line 191:
```suggestion
    @deprecated.argument("25.9", "26.3", "auto_activate_base", rename="auto_activate")
```

conda/base/context.py
```diff
@@ -1187,6 +1190,15 @@ def console(self) -> str:
             return DEFAULT_JSON_REPORTER_BACKEND
         return self._console
 
+    @property
+    @deprecated(
+        "25.5",
+        "26.1",
```
kenodegard, line 1196:
```suggestion
        "25.9",
        "26.3",
```

news/14666-default-env
```diff
@@ -0,0 +1,20 @@
...
+* Add `context.default_activation_env` setting, which allows users to customize which environment should get auto activated on initialization (if `context.auto_activate` is enabled), as well as `conda activate` calls without arguments. Defaults to `base`. (#14666)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Mark `context.auto_activate_base` as pending deprecation in favor of `context.auto_activate`.
+  It will be permanently removed in 26.1. (#14666)
```
kenodegard, line 12: We typically haven't included the actual removal version in the changelog, is this something we should start including?

```suggestion
* Mark `conda.base.context.context.auto_activate_base` as pending deprecation. Use `conda.base.context.context.auto_activate` instead.  (#14666)
```

conda/activate.py
```diff
@@ -188,18 +188,17 @@ def reactivate(self):
...
             builder.append(self.hook_source_path.read_text())
-        if (
-            auto_activate_base is None
-            and context.auto_activate_base
-            or auto_activate_base
-        ):
-            builder.append("conda activate base\n")
+        if auto_activate is None and context.auto_activate or auto_activate:
+            builder.append(
+                f"conda activate '{context.default_activation_env or ROOT_ENV_NAME}'\n"
```
kenodegard, line 200: Since we do this in several places are we not better off doing this in a property in the `context` object?

```python
class Context(...):
     _default_activation_env = ParameterLoader(PrimitiveParameter(ROOT_ENV_NAME), aliases=("default_activation_env",)

     @property
```

news/14666-default-env
jaimergp: I think it's useful to put in the notes, yes.

jaimergp:
```suggestion
* Mark `conda.base.context.context.auto_activate_base` as pending deprecation. Use `conda.base.context.context.auto_activate` instead. It will be permanently removed in 26.3. (#14666)
```

conda/activate.py
jaimergp: Oh, nice, yes.

PR #14703, opened 2025-03-25
Add `conda init --condabin` to add `condabin/` to `PATH`

### LINKED ISSUE #20: Add `conda init --condabin`
https://github.com/conda/conda/pull/14703

jaimergp:
### Description

Part of https://github.com/conda/constructor/issues/960 and https://github.com/conda/constructor/pull/965.

### Checklist - did you ...


jaimergp:
@marcoesters I'd also appreciate your feedback here, since it will be mostly used in `constructor` and `conda-standalone`.

marcoesters:
> @conda/conda-maintainers, this is ready for review. @marcoesters I'd also appreciate your feedback here, since it will be mostly used in `constructor` and `conda-standalone`.

I think from a `constructor` and `conda-standalone` perspective, the changes look mostly fine. The only thing that needs to change on the `conda-standalone` is how the "init reverse" plan is handled: https://github.com/conda/conda-standalone/blob/12e05855d2135fa9e82510a01c4ef2a2a3d93a65/src/entry_point.py#L278

It currently has `make_initialize_plan` hard-coded. The only difference between `make_initialize_plan` and `make_condabin_plan` in their inputs is that the former contains `anaconda_prompt` as input. Since the prompt is nowadays handled via `menuinst`, it that one should probably be deprecated though.

To make `conda-standalone` more robust, I suggest creating a top-level `make_plan`-like function that takes `ContentType` as input. In `conda-standalone`, `ContentType` can then be imported and iterated over (maybe make this an Enum?).

conda/core/initialize.py
```diff
@@ -1201,7 +1429,31 @@ def _config_fish_content(conda_prefix):
     return conda_initialize_content
 
 
-def init_fish_user(target_path, conda_prefix, reverse):
+def _config_fish_content_to_add_condabin_to_path(conda_prefix):
+    condabin_dir = join(conda_prefix, "condabin")
+    if on_win:
+        from ..activate import native_path_to_unix
+
+        condabin_dir = native_path_to_unix(condabin_dir)
```
kenodegard, line 1437: This has been deprecated:

```suggestion
        from ..common.path import win_path_to_unix

        condabin_dir = win_path_to_unix(condabin_dir)
```

kenodegard (dismissed):
I have concerns about adding anything new to `conda.core.initialize`. What would it take for us to go the plugin hook route instead?

jaimergp:
> I have concerns about adding anything new to conda.core.initialize. What would it take for us to go the plugin hook route instead?

Can you clarify? Would you like to see a new subcommand, or adding plugin hooks to `init`?

I added it here because I'd like to reuse the revert framework and it makes sense to me to keep all initialization code in the same spot. Ideally we would end up deprecating the whole `conda init` shell function business and just leave the `PATH` stuff, but that's for the future.

The code paths here are sufficiently uncoupled to not add much risk to the PR, imo.

news/14703-condabin
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Provide `conda init --condabin`, which only adds the `condabin/` directory to the PATH environment variable instead of installing a shell function. (#14703)
```
jaimergp, line 3:
```suggestion
* Provide `conda init --condabin`, which only adds the `$CONDA_PREFIX/condabin` directory to the PATH environment variable instead of installing a shell function. (#14703)
```

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -429,9 +429,13 @@ behavior uniform across operating systems. Conda 4.4 allowed
 support so that conda works faster and less disruptively on
 a wide variety of shells (bash, zsh, csh, fish, xonsh, and more).
 Now these shells can use the ``conda activate`` command.
-Removing the need to modify PATH makes conda less disruptive to
-other software on your system. For more information, read the
-output from ``conda init --help``.
+
+Alternatively, ``conda init --condabin`` will not install a shell
+function in your profile. Instead, it will only add the ``condabin/``
```
jaimergp, line 434:
```suggestion
function in your profile. Instead, it will only add the ``$CONDA_PREFIX/condabin/``
```

conda/core/initialize.py
```diff
@@ -1679,6 +2038,56 @@ def init_cmd_exe_registry(target_path, conda_prefix, reverse=False):
...
+            for part, normalized_part in zip(path_parts, normalized_path_parts)
+            if normalized_part != normalized_condabin_path
+        ]
+        new_value = ";".join(new_path_parts)
+    else:
+        # Add condabin_path if not already present (typically at the beginning)
+        if normalized_condabin_path not in normalized_path_parts:
+            # Prepend condabin_path
+            path_parts.insert(0, condabin_path)
+        new_value = ";".join(path_parts)
```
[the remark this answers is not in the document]
jaimergp: That code is mostly for `bin`, `Scripts` and `Library\bin`, so I'm not sure how many similarities we can bring together. It takes into account some edge cases that do not apply here too (especially in the "replace" bits). And it's multi-platform, while this is Windows only.

I don't know, I feel that this code is clear and self-contained as is, while adding a dependency on some other code paths might introduce breakages in the future as that code is touched up and maintained.

tests/core/test_initialize.py
```diff
@@ -1059,6 +1065,161 @@ def _read_windows_registry_mock(target_path, value=None):
     assert c.stdout.strip().splitlines()[-1][1:] == expected
 
 
+@pytest.mark.skipif(not on_win, reason="windows-only test")
```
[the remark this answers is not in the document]
jaimergp: Thanks for the feedback. They all had a similar structure, so opted for a huuge parameterized list case. It's fast though.

jaimergp:
> The only thing that needs to change on the conda-standalone is how the "init reverse" plan is handled: [conda/conda-standalone@12e0585/src/entry_point.py#L278](https://github.com/conda/conda-standalone/blob/12e05855d2135fa9e82510a01c4ef2a2a3d93a65/src/entry_point.py#L278)

I think for now we can add the extra logic to handle `--condabin` in conda-standalone to see how it looks like.

> Since the prompt is nowadays handled via menuinst, it that one should probably be deprecated though.

Yep, agreed, I don't think we are using this anywhere. Should be handled in a separate PR though.

> To make conda-standalone more robust, I suggest creating a top-level make_plan-like function that takes ContentType as input. In conda-standalone, ContentType can then be imported and iterated over (maybe make this an Enum?).

There's also a `make_install_plan` function though, so I'm not sure how to create a high-level API for all three. Can we handle that in a later refactor? This PR is pretty unwieldy already.

conda/core/initialize.py
```diff
@@ -1312,7 +1570,28 @@ def _config_xonsh_content(conda_prefix):
     return conda_initialize_content
 
 
-def init_xonsh_user(target_path, conda_prefix, reverse):
+def _config_xonsh_content_to_add_condabin_to_path(conda_prefix):
+    condabin_dir = join(conda_prefix, "condabin")
+    if on_win:
+        from ..activate import native_path_to_unix
+
+        condabin_dir = native_path_to_unix(condabin_dir)
```
kenodegard, line 1578:
```suggestion
        from ..common.path import win_path_to_unix

        condabin_dir = win_path_to_unix(
```

conda/core/initialize.py
```diff
@@ -1439,7 +1727,50 @@ def _bashrc_content(conda_prefix, shell):
     return conda_initialize_content
 
 
-def init_sh_user(target_path, conda_prefix, shell, reverse=False):
+def _bashrc_content_to_add_condabin_to_path(conda_prefix, shell):
+    condabin_path = join(conda_prefix, "condabin")
+    if on_win:
+        from ..activate import native_path_to_unix
+
+        condabin_path = native_path_to_unix(condabin_path)
```
kenodegard, line 1735:
```suggestion
        from ..common.path import win_path_to_unix

        condabin_dir = win_path_to_unix(
```

conda/cli/main_init.py
```diff
@@ -36,6 +36,11 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         They're therefore implemented in ways specific to each shell. Each shell must be configured
         to make use of them.
 
+        The --condabin option adds the "$CONDA_PREFIX/condabin" directory to the PATH environment variable.
+        This directory only contains the conda executable and will not pollute your PATH so much.
+        On most shells, a small snippet is added to the shell profile. For CMD, the PATH environment
+        variable is modified directly in the registry.
```
travishathaway, line 42:
```suggestion
        The --condabin option adds the "$CONDA_PREFIX/condabin" directory to the PATH environment variable.
        This directory only contains the conda executable and does not contain other executables from other
        programs or libraries installed in the base conda environment. On most shells, a small snippet is added to the
        shell profile. For CMD, the PATH environment variable is modified directly in the registry.
```

@jaimergp, I was looking at this again with fresh eyes and found the phrase, "not pollute your PATH so much" to be a little ambiguous. I tried to add a succinct explanation of what you meant. Please include if you think it's an improvement.

travishathaway (approved):

Issue #13344, opened 2023-11-17
JSONDecodeError after updating conda


ShadeAlsha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

After successfully updating my Conda version using the command, Conda stopped functioning and is now displaying the following error.

Conda update command: 
`conda install -n base conda=23.7.3 conda-build=3.26.1`

After the update, I am getting this error:
`Traceback (most recent call last):
  File "/anaconda/lib/python3.8/site-packages/conda/exception_handler.py", line 17, in __call__
    return func(*args, **kwargs)
  File "/anaconda/lib/python3.8/site-packages/conda/cli/main.py", line 54, in main_subshell
    parser = generate_parser(add_help=True)
  File "/anaconda/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 115, in generate_parser
    configure_parser_create(sub_parsers)
  File "/anaconda/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 767, in configure_parser_create
    add_parser_solver(solver_mode_options)
  File "/anaconda/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 1975, in add_parser_solver
    solver.name for solver in context.plugin_manager.get_hook_results("solvers")
  File "/anaconda/lib/python3.8/site-packages/conda/base/context.py", line 492, in plugin_manager
    from ..plugins.manager import get_plugin_manager
  File "/anaconda/lib/python3.8/site-packages/conda/plugins/__init__.py", line 28, in <module>
    from .hookspec import hookimpl  # noqa: F401
  File "/anaconda/lib/python3.8/site-packages/conda/plugins/hookspec.py", line 16, in <module>
    from .types import (
  File "/anaconda/lib/python3.8/site-packages/conda/plugins/types.py", line 15, in <module>
    from ..core.solve import Solver
  File "/anaconda/lib/python3.8/site-packages/conda/core/solve.py", line 30, in <module>
    from ..exceptions import (
  File "/anaconda/lib/python3.8/site-packages/conda/exceptions.py", line 16, in <module>
    from requests.exceptions import JSONDecodeError
ImportError: cannot import name 'JSONDecodeError' from 'requests.exceptions' (/anaconda/lib/python3.8/site-packages/requests/exceptions.py)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/anaconda/bin/conda", line 13, in <module>
[...]

PR #14709, opened 2025-03-25
Remove `JSONEncoder` monkeypatching

kenodegard:
### Description

While investigating the recent build issues I noticed that we are importing and doing too much in `conda/__init__.py`, especially with monkey patching the Python stdlib (something we should be moving away from whenever possible).

So here we consolidate the json encoding logic into one place (`conda.common.serialize.json`). We can mirror this move for the yaml logic in a subsequent PR.

### Checklist - did you ...


conda/auxlib/entity.py
```diff
@@ -972,19 +973,4 @@ def update(self, E=None, **F):
...
-        elif hasattr(obj, 'to_json'):
-            return obj.to_json()
-        elif hasattr(obj, 'as_json'):
-            return obj.as_json()
-        elif isinstance(obj, Enum):
-            return obj.value
-        elif isinstance(obj, Path):
-            return str(obj)
-        return JSONEncoder.default(self, obj)
+deprecated.constant("25.9", "26.3", "EntityEncoder", CondaJSONEncoder, addendum="Use `conda.common.serialize.json.CondaJSONEncoder` instead.",)
```
dholth, line 976: The title of this PR was misleading. This is not monkeypatching since we are not setting properties on the global json module. It is subclassing and using a [standard json module feature.](https://docs.python.org/3/library/json.html#json.dump)

This does seem like a slow function. Would a faster version keep a mapping of `type(obj) -> (needs custom handling)` to avoid doing hasattr all the time.

Or we continue to move away from auxlib and don't do any of this.

conda/cli/main_config.py
```diff
@@ -627,12 +615,9 @@ def execute_config(args, parser):
                 )
             if context.json:
                 stdout_write(
-                    json.dumps(
+                    dumps(
                         [context.describe_parameter(name) for name in paramater_names],
```
dholth, line 619: parameter or parameter

conda/common/serialize/__init__.py
```diff
@@ -60,11 +60,30 @@ def yaml_safe_dump(object, stream=None):
...
+
+
+@deprecated(
+    "25.9",
+    "26.3",
+    addendum="Use `conda.common.serialize.json.dumps(sort_keys=True)` instead.",
+)
 def json_dump(object):
-    return json.dumps(
-        object, indent=2, sort_keys=True, separators=(",", ": "), cls=EntityEncoder
```
dholth, line 69: A funny choice from past conda developers, to make only one of them include a space... for me, the most compact version is best.

conda/__init__.py
```diff
@@ -158,9 +156,25 @@ def conda_signal_handler(signum, frame):
...
-    if hasattr(obj, "to_json"):
+    elif hasattr(obj, "to_json"):
+        deprecated.topic(
+            "25.9",
+            "26.3",
+            "Monkey-patching `json.JSONEncoder` to support `obj.to_json()`",
+            addendum="Use `conda.common.serialize.json.CondaJSONEncoder` instead.",
+        )
         return obj.to_json()
     return _default.default(obj)
```
dholth, line 180: Re prior comments, I see that the monkeypatch is at the unchanged bottom of this file and so doesn't show up in the diff.

conda/cli/main_config.py
kenodegard: already fixed in https://github.com/conda/conda/pull/14510

conda/common/serialize/__init__.py
kenodegard: 🤷🏼‍♂️

jezdez:
pre-commit.ci autofix

conda/common/serialize/json.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # auxlib entity types
+        if hasattr(obj, "dump"):
+            return obj.dump()
+        elif hasattr(obj, "__json__"):
+            return obj.__json__()
+        elif hasattr(obj, "to_json"):
+            return obj.to_json()
+        elif hasattr(obj, "as_json"):
+            return obj.as_json()
```
jaimergp: This would benefit from a `for` loop I think?

```suggestion
        # auxlib entity types
        for attr in ("dump", "__json__", "to_json", "as_json"):
            if method := getattr(obj, attr, None):
                return method() 

```

conda/common/serialize/json.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # default
+        return super().default(obj)
+
+
+def dump(*args, **kwargs):
+    if kwargs.get("cls") is None:
+        kwargs["cls"] = CondaJSONEncoder
+    if kwargs.get("indent") is None:
+        kwargs["indent"] = 2
```
jaimergp:
```suggestion
    kwargs.setdefault("cls", CondaJSONEncoder)
    kwargs.setdefault("indent", 2)
```

conda/common/serialize/json.py
```diff
@@ -0,0 +1,58 @@
...
+    return json.dump(*args, **kwargs)
+
+
+def dumps(*args, **kwargs):
+    if kwargs.get("cls") is None:
+        kwargs["cls"] = CondaJSONEncoder
+    if kwargs.get("indent") is None:
+        kwargs["indent"] = 2
```
jaimergp:
```suggestion
    kwargs.setdefault("cls", CondaJSONEncoder)
    kwargs.setdefault("indent", 2)
```

conda/core/prefix_data.py
```diff
@@ -487,10 +483,10 @@ def _load_single_record(self, prefix_record_json_path: PathType) -> None:
         log.debug("loading prefix record %s", prefix_record_json_path)
         with open(prefix_record_json_path) as fh:
             try:
-                json_data = json_load(fh.read())
+                json_data = json.loads(fh.read())
```
jaimergp: Why not pass the descriptor instead?

```suggestion
                json_data = json.load(fh)
```

jezdez:
pre-commit.ci autofix

conda/core/prefix_data.py
dholth: loads is faster than load(fh) iirc significantly

conda/auxlib/entity.py
```diff
@@ -239,7 +239,6 @@
 from datetime import datetime
 from enum import Enum
 from functools import reduce
-from json import JSONEncoder, dumps as json_dumps, loads as json_loads
```
dholth, line 242: Is this a Python optimization to try to avoid repeated "json.loads" lookups?

conda/exceptions.py
```diff
@@ -15,16 +14,14 @@
 from traceback import format_exception, format_exception_only
 from typing import TYPE_CHECKING
 
-from requests.exceptions import JSONDecodeError
```
dholth, line 18: https://github.com/conda/conda/issues/13344

https://github.com/psf/requests/issues/5794

and others...

Maybe the issue of requiring requests' JSONDecodeError is no longer a problem? In case requests is using simplejson apparently.

conda/auxlib/entity.py
jezdez: I'm not sure, that seems unreasonable, the [`json module`](https://github.com/python/cpython/blob/main/Lib/json/__init__.py) seems pretty straightforward about access to the various module level functions like `loads`

conda/exceptions.py
jezdez: Oh, that's a good point, at least the compat code still exists in requests, https://github.com/psf/requests/blob/91a3eabd3dcc4d7f36dd8249e4777a90ef9b4305/src/requests/exceptions.py#L31-L52

jezdez: Oh, I see our new json module here imports the correct compat exception class though: https://github.com/conda/conda/pull/14709/files#diff-84797ed824a58ab550574945be3861f96016e6b0fd2daa654b73925f398b77e2R11-R49

jezdez: Done something about it in [`932b044` (#14709)](https://github.com/conda/conda/pull/14709/commits/932b044e65d622d643e1de32f8fd418ca7630778)

jezdez:
pre-commit.ci autofix

jezdez (approved):

PR #14710, opened 2025-03-25
Add env specs hooks - part 1

### LINKED ISSUE #18: Environment specifier hooks

soapy1:
### Description

This PR follows https://github.com/conda/conda/pull/13930 in adding hooks for env specs plugins. These plugins will allow for new ways of defining conda environments. Currently, `conda env create` supports `environment.yaml` and `requirements.txt` (binstar environments are deprecated) files for defining a conda environment.

In addition to defining the plugin hooks, this PR refactors
* `conda.env` module so that the existing env spec classes implement the plugin interface 
* updates `conda env create` to use the plugins to read the environment file

For example, to define a new env spec plugin:

```python
import json
import random
from pathlib import Path
from subprocess import run
from conda import plugins
from .conda.plugins.types import EnvironmentSpecifierBase
from conda.env.env import Environment

packages = ["python", "numpy", "scipy", "matplotlib", "pandas", "scikit-learn"]
class RandomSpec(EnvironmentSpecifierBase):
    extensions = {".random"}

    def __init__(self, filename: str):
        self.filename = filename

    def can_handle(self):
        return random.random() < 0.5

    def environment(self):
        return Environment(
            name="".join(random.choice("0123456789abcdef") for i in range(6)),
            dependencies=[random.choice(packages) for i in range(6)],
        )

@plugins.hookimpl
    yield plugins.CondaEnvSpec(
        name="jpg",
        handler_class=RandomSpec,
    )
```

In part 2, will do some refactoring around `conda.env.env` and the yaml file spec handler. Todo in part 2:
- [ ] stop using `env.specs.__init__::detect` (it is deprecated in this pr)
- [ ] move all the code pieces for reading yaml environment specs into the `YamlFileSpec` plugin
- [x] improve docs - https://github.com/conda/conda/pull/14814

### Checklist - did you ...

### DIFF
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -361,7 +361,7 @@ def print_activate(env_name_or_prefix):  # pragma: no cover
         print(message)  # TODO: use logger
 
 
-def validate_env_file_exists(filename: str):
+def validate_file_exists(filename: str):
     """
     Validate the existence of an environment file.
 
@@ -378,10 +378,12 @@ def validate_env_file_exists(filename: str):
         EnvironmentFileNotFound: If the file does not exist and is not a valid URL.
     """
     url_scheme = filename.split("://", 1)[0]
-    if url_scheme in CONDA_SESSION_SCHEMES:
-        return
-    else:
+    if url_scheme == "file":
+        filename = expand(filename.split("://", 1)[-1])
+    elif url_scheme not in CONDA_SESSION_SCHEMES:
         filename = expand(filename)
+    else:
+        return
 
     if not exists(filename):
         raise EnvironmentFileNotFound(filename=filename)
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -5,7 +5,6 @@
 import re
 import sys
 from logging import getLogger
-from os.path import dirname, isdir, isfile, join, normcase
 from os.path import (
     dirname,
     exists,
@@ -20,7 +19,7 @@ from ..base.constants import PREFIX_MAGIC_FILE
 from ..base.context import context, env_name
 from ..common.constants import NULL
 from ..common.io import swallow_broken_pipe
-from ..common.path import paths_equal, expand
+from ..common.path import expand, paths_equal
 from ..deprecations import deprecated
 from ..exceptions import (
     CondaError,
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -7,11 +7,8 @@ import sys
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, normcase
 from os.path import (
-    abspath,
     dirname,
     exists,
-    expanduser,
-    expandvars,
     isdir,
     isfile,
     join,
@@ -23,7 +20,7 @@ from ..base.constants import PREFIX_MAGIC_FILE
 from ..base.context import context, env_name
 from ..common.constants import NULL
 from ..common.io import swallow_broken_pipe
-from ..common.path import paths_equal
+from ..common.path import paths_equal, expand
 from ..deprecations import deprecated
 from ..exceptions import (
     CondaError,
@@ -370,7 +367,7 @@ def validate_env_file_exists(filename: str):
     if url_scheme in CONDA_SESSION_SCHEMES:
         return
     else:
-        filename = abspath(expanduser(expandvars(filename)))
+        filename = expand(filename)
 
     if not exists(filename):
         raise EnvironmentFileNotFound(filename=filename)
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -362,6 +362,21 @@ def print_activate(env_name_or_prefix):  # pragma: no cover
 
 
 def validate_env_file_exists(filename: str):
+    """
+    Validate the existence of an environment file.
+
+    This function checks if the given ``filename`` exists as an environment file.
+    If the `filename` has a URL scheme supported by ``CONDA_SESSION_SCHEMES``,
+    it assumes the file is accessible and returns without further validation.
+    Otherwise, it expands the given path and verifies its existence. If the file
+    does not exist, an ``EnvironmentFileNotFound`` exception is raised.
+
+    Parameters:
+        filename (str): The path or URL of the environment file to validate.
+
+    Raises:
+        EnvironmentFileNotFound: If the file does not exist and is not a valid URL.
+    """
     url_scheme = filename.split("://", 1)[0]
     if url_scheme in CONDA_SESSION_SCHEMES:
         return
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -7,7 +7,15 @@ import sys
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, normcase
 from os.path import (
-    abspath, dirname, exists, expanduser, expandvars, isdir, isfile, join, normcase
+    abspath,
+    dirname,
+    exists,
+    expanduser,
+    expandvars,
+    isdir,
+    isfile,
+    join,
+    normcase,
 )
 
 from ..auxlib.ish import dals
@@ -25,8 +33,8 @@ from ..exceptions import (
     EnvironmentNotWritableError,
     OperationNotAllowed,
 )
-from ..gateways.disk.test import file_path_is_writable
 from ..gateways.connection.session import CONDA_SESSION_SCHEMES
+from ..gateways.disk.test import file_path_is_writable
 from ..models.match_spec import MatchSpec
 from ..reporters import render
 
@@ -363,6 +371,6 @@ def validate_env_file_exists(filename: str):
         return
     else:
         filename = abspath(expanduser(expandvars(filename)))
-    
+
     if not exists(filename):
         raise EnvironmentFileNotFound(filename=filename)
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -6,6 +6,9 @@ import re
 import sys
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, normcase
+from os.path import (
+    abspath, dirname, exists, expanduser, expandvars, isdir, isfile, join, normcase
+)
 
 from ..auxlib.ish import dals
 from ..base.constants import PREFIX_MAGIC_FILE
@@ -17,11 +20,13 @@ from ..deprecations import deprecated
 from ..exceptions import (
     CondaError,
     DirectoryNotACondaEnvironmentError,
+    EnvironmentFileNotFound,
     EnvironmentLocationNotFound,
     EnvironmentNotWritableError,
     OperationNotAllowed,
 )
 from ..gateways.disk.test import file_path_is_writable
+from ..gateways.connection.session import CONDA_SESSION_SCHEMES
 from ..models.match_spec import MatchSpec
 from ..reporters import render
 
@@ -350,3 +355,14 @@ def print_activate(env_name_or_prefix):  # pragma: no cover
             """
         )
         print(message)  # TODO: use logger
+
+
+def validate_env_file_exists(filename: str):
+    url_scheme = filename.split("://", 1)[0]
+    if url_scheme in CONDA_SESSION_SCHEMES:
+        return
+    else:
+        filename = abspath(expanduser(expandvars(filename)))
+    
+    if not exists(filename):
+        raise EnvironmentFileNotFound(filename=filename)
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -88,14 +88,17 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..auxlib.ish import dals
     from ..base.context import context, determine_target_prefix
     from ..core.prefix_data import PrefixData
+    from .common import validate_env_file_exists
     from ..env import specs as install_specs
     from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..exceptions import CondaEnvException, InvalidInstaller
 
-    spec = install_specs.detect(
-        filename=args.file,
-    )
+    # validate incoming arguments
+    validate_env_file_exists(args.file)
+    
+    # detect the file format and get the env representation
+    spec = install_specs.detect(filename=args.file)
     env = spec.environment
 
     if not (args.name or args.prefix):
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -88,15 +88,15 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..auxlib.ish import dals
     from ..base.context import context, determine_target_prefix
     from ..core.prefix_data import PrefixData
-    from .common import validate_env_file_exists
     from ..env import specs as install_specs
     from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..exceptions import CondaEnvException, InvalidInstaller
+    from .common import validate_env_file_exists
 
     # validate incoming arguments
     validate_env_file_exists(args.file)
-    
+
     # detect the file format and get the env representation
     spec = install_specs.detect(filename=args.file)
     env = spec.environment
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -92,10 +92,10 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..exceptions import CondaEnvException, InvalidInstaller
-    from .common import validate_env_file_exists
+    from .common import validate_file_exists
 
     # validate incoming arguments
-    validate_env_file_exists(args.file)
+    validate_file_exists(args.file)
 
     # detect the file format and get the env representation
     spec = install_specs.detect(filename=args.file)
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -107,7 +107,6 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             # TODO Refactor common.get_prefix
             name = os.environ.get("CONDA_DEFAULT_ENV", False)
             if not name:
-                msg = "Unable to determine environment\n\n"
                 msg = dals(
                     """
                     Unable to determine environment
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -89,14 +89,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import context, determine_target_prefix
     from ..core.prefix_data import PrefixData
     from ..env import specs as install_specs
-    from ..env.env import get_filename, print_result
+    from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..exceptions import CondaEnvException, InvalidInstaller
 
     spec = install_specs.detect(
-        name=args.name,
-        filename=get_filename(args.file),
-        directory=os.getcwd(),
+        filename=args.file,
     )
     env = spec.environment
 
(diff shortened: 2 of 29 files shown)


conda/env/env.py
```diff
@@ -265,15 +264,6 @@ def save(self):
             self.to_yaml(stream=fp)
 
 
-def get_filename(filename):
```
soapy1, line 268: not sure if this needs to be deprecated instead of removed like this.

conda/cli/common.py
```diff
@@ -281,3 +285,14 @@ def validate_prefix_is_writable(prefix: str) -> str:
     if isdir(dirname(test_path)) and file_path_is_writable(test_path):
         return prefix
     raise EnvironmentNotWritableError(prefix)
+
+
+def validate_env_file_exists(filename: str):
+    url_scheme = filename.split("://", 1)[0]
+    if url_scheme in CONDA_SESSION_SCHEMES:
```
soapy1, line 292: Should conda try to see if the remote file exists? It wasn't doing that previously. Maybe out of scope for this PR.

soapy1:
pre-commit.ci autofix

soapy1:
@conda/conda-maintainers this is ready for review, would love some feedback when you have time to take a look :heart_eyes:

conda/env/env.py
jaimergp: We need to deprecate it.

peytondmurray (commented):
Two thoughts about `EnvironmentSpecifierBase`:

1. The current implementation allows the user to create environment specifier classes for files that can't be parsed by the specifier. I guess that's okay because everywhere in the code that environment specifiers are used, they immediately check `spec.can_handle()`, but IMO it's a bit awkward to have this two-step initialization process:

```python
    spec = spec_hook.handler_class(filename)
    if spec.can_handle():
        # proceed with handling the environment
```

Can the `can_handle()` call just be moved to inside `__init__()`? That way we avoid the need to remember to call it everywhere, and we won't be able to instantiate `EnvironmentSpecifier` classes for files they can't handle.

2. All the `EnvironmentSpecifierBase` subclasses do is instantiate `Environment` objects. Is this interface needed? If I was writing this from scratch I think I'd just make a classmethod on the Environment like this:

```python
class Environment:
    ...
    @classmethod
    def from_spec(cls, spec) -> Environment:
        # Get all the pluggy hooks here, and return the
        # environment from whichever one successfully parses it.
        return env
```

In other words, `CondaEnvironmentSpecifier` could just take a name of the extension that you want to parse, and a function that parses to an environment. Thoughts?

conda/env/specs/requirements.py
```diff
@@ -16,27 +18,51 @@ class RequirementsSpec:
...
+
+    @deprecated("25.9", "26.3", addendum="This method is not used anymore.")
     def _valid_file(self):
         if os.path.exists(self.filename):
             return True
         else:
             self.msg = "There is no requirements.txt"
             return False
 
+    @deprecated("25.9", "26.3", addendum="This method is not used anymore.")
```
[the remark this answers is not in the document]
jaimergp: Looking at the [changelog](https://github.com/conda/conda/blob/main/CHANGELOG.md) and searching for `._`, reveals plenty of deprecations of private names, so I don't think we have a choice here.

conda/cli/main_env_update.py
```diff
@@ -87,34 +87,37 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
     env = spec.environment
 
     if not (args.name or args.prefix):
         if not env.name:
-            # Note, this is a hack fofr get_prefix that assumes argparse results
+            # Note, this is a hack for get_prefix that assumes argparse results
             # TODO Refactor common.get_prefix
             name = os.environ.get("CONDA_DEFAULT_ENV", False)
             if not name:
                 msg = "Unable to determine environment\n\n"
```
kenodegard, line 109:
```suggestion
```

conda/env/specs/__init__.py
```diff
@@ -43,6 +57,10 @@ def get_spec_class_from_file(filename: str) -> FileSpecTypes:
     raise EnvironmentFileNotFound(filename=filename)
 
 
+@deprecated.argument("25.9", "26.3", "name")
+@deprecated.argument(
+    "25.9", "26.3", "directory", addendum="Specify the full path in filename"
+)
```
kenodegard, line 63: Also need to remove the deprecated arguments from the signature

conda/env/specs/requirements.py
```diff
@@ -16,27 +18,51 @@ class RequirementsSpec:
     msg = None
     extensions = {".txt"}
 
+    @deprecated.argument("24.7", "26.3", "name")
     def __init__(self, filename=None, name=None, **kwargs):
```
kenodegard, line 22:
```suggestion
    def __init__(self, filename=None, **kwargs):
```

conda/plugins/types.py
```diff
@@ -347,3 +348,45 @@ class CondaRequestHeader:
 
     name: str
     value: str
+
+
+class EnvironmentSpecifierBase(ABC):
```
kenodegard, line 353: I think we should use `EnvironmentSpec` since it's already called `RequirementsSpec` and `YamlFileSpec`.

Sure we can rename all of these but I'm pretty sure we are not about to rename `MatchSpec` so I think we should be consistent and stick to the `*Spec` convention.

soapy1:
@peytondmurray thanks for taking a look at this PR! 

### Q 1. part 1

> but IMO it's a bit awkward to have this two-step initialization process 

ya, good catch! I think the plugin manager layer should be responsible for calling the `can_handle` function. So, a conda maintainer should only ever get the env spec by calling the plugin manager, something like 
```
    spec_hook = context.plugin_manager.get_environment_specifier_handler(
        filename=filename,
    )
    spec = spec_hook.handler_class(filename)
```

and the `get_environment_specifier_handler` is looping thru all the provided plugins to return the first positive result for a plugin that `can_handle` (https://github.com/conda/conda/blob/346ce8b599afa152e4cf4903ef1767f09ab7c30e/conda/plugins/manager.py#L477). 

### Q 1. part 2
>  Can the can_handle() call just be moved to inside `__init__()`? 

I think this is a perfectly good pattern to use. I think the plugin maintainer can make this decision.

### Q 2. part 1
> All the EnvironmentSpecifierBase subclasses do is instantiate Environment objects. Is this interface needed? If I was writing this from scratch I think I'd just make a classmethod on the Environment 

hmmm, so I interpret the root of this issue being the awkwardness in this part of the code https://github.com/conda/conda/blob/346ce8b599afa152e4cf4903ef1767f09ab7c30e/conda/env/specs/__init__.py#L64. I think the suggested code block
```
class Environment:
    ...
    @classmethod
    def from_spec(cls, spec) -> Environment:
        spec_hook = context.plugin_manager.get_environment_specifier_handler(
          filename=filename,
        )
        spec = spec_hook.handler_class(filename)
        return spec.environment
```
 is a nice way to make parsing the environment work more smoothly. I think there is a little more refactoring that needs to be done to make this work. I would love to follow up with this as one of the items to address in part 2

### Q 2. part 2
> In other words, CondaEnvironmentSpecifier could just take a name of the extension that you want to parse, and a function that parses to an environment. Thoughts?

I think just using an extension is not enough information to determine if a particular EnvSpec plugin can handle a particular file. For example, you can imagine multiple EnvSpec plugins that use `.yaml` files but support different schemas.

conda/cli/main_env_update.py
```diff
@@ -87,34 +87,37 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-    spec = install_specs.detect(
-        name=args.name,
-        filename=get_filename(args.file),
-        directory=os.getcwd(),
-    )
+    # validate incoming arguments
+    validate_env_file_exists(args.file)
+
+    # detect the file format and get the env representation
+    spec = install_specs.detect(filename=args.file)
```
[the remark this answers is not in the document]
soapy1: I think this is a good suggestion but I think I won't add it to this PR:
(1) `install_specs.detect` is getting deprecated in this PR so I feel a little sketchy changing it's return type
(2) I'll address the whole awkwardness around this part of the code in the follow up PR (https://github.com/conda/conda/pull/14710#issuecomment-2842812261)

conda/env/specs/__init__.py
```diff
@@ -52,12 +70,11 @@ def detect(
...
-        spec = spec_class(name=name, filename=filename, directory=directory)
-        if spec.can_handle():
-            return spec
-
+    spec_hook = context.plugin_manager.get_environment_specifier_handler(
+        filename=filename,
+    )
+    spec = spec_hook.handler_class(filename)
+    if spec.can_handle():
+        return spec
```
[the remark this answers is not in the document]
soapy1: I think you are right! This is marked for deprecation now so I don't want to change up the interface for this. But will address the issue of how to detect specs more elegantly in a follow up PR.

jezdez (commented):
I like where this is going, I have only naming nitpicking, and one big question, assuming we do plugin hooks for environment installers (`env_installers` plugin hook) and environment lockers (`env_lockers`) (https://github.com/conda/conda-environments-project-mgmt/issues/36#issuecomment-2842984851), I think it might be more consistent to name the plugin hook implemented here `env_specs` and use "spec" as the main nomenclature. Thoughts?

soapy1:
pre-commit.ci autofix

soapy1:
@travishathaway thanks for taking a look at this PR! You're absolutely right, writing good docs is an important step to adding a feature :book: I took a stab at writing some better docs for this. 

Unfortunately, it looks like there is too much histor in this branch so read the docs is not able to check out the lastest commit. This seems like probably a bug related to https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/vcs_support/backends/git.py#L28.

soapy1:
Hmmm, it looks like maybe some of these failing windows tests are flaky? It looks there is something weird going on when executing pip. Some of the errors look like:

> ValueError: Unable to find resource t64.exe in package pip._vendor.distlib

jaimergp:
It might be one of the new setuptools releases, or the new pip release. Also see https://github.com/pypa/pip/issues/13368. It looks like `defaults` is missing some the `distlib` executables: https://conda-metadata-app.streamlit.app/?q=pkgs%2Fmain%2Fnoarch%2Fpip-25.1-pyhc872135_0.conda

jaimergp:
I opened https://github.com/AnacondaRecipes/pip-feedstock/issues/36

docs/source/dev-guide/plugins/environment_specifiers.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Using the Plugin
+----------------
+Once this plugin is registered, users will be able to create environments from the
+types of files specified by the plugin. For example to create a `random` environment
+using the plugin defined above:
+
+.. code-block:: bash
+
+   conda env create --file /doesnt/matter/any/way.random
```
travishathaway, line 69: Could we use something a little more realistic like `.xml` or `.json`? The `random` example feels a bit too abstract, and I think it would help future readers to use something that might actually be implemented instead.

docs/source/dev-guide/plugins/environment_specifiers.rst
```diff
@@ -0,0 +1,71 @@
+======================
+Environment Specifiers
+======================
+
+Conda can create environments from several file formats. Currently, conda natively
+supports creating environments from:
+
+* a yaml `environment.yaml`
+* an "explicit" text file
```
travishathaway, line 9:
```suggestion
* A :ref:`YAML environment.yaml file <create-env-file-manually>`
* An :ref:`"explicit" text file <identical-conda-envs>`
```

We could make this slightly better with direct links to other parts of the documentation 😄.

travishathaway: I also tested this myself, and verified that's the correct format too.

jezdez (commented):
I'm sorry for flip-flopping on this, I realized the weirdness about the naming of `environment_spec`, `env_spec`, `Environment Specifieres` and the existing `<NAME>Spec` classes made it more appealing to use `env_spec` as a shorthand, but I think I actually prefer your spelling of the API of `environment_spec` better.

Otherwise this seems really really good, I'll do some manual testing today for further review (writing a plugin etc).

conda/env/specs/binstar.py
```diff
@@ -60,7 +60,7 @@ def valid_name(self) -> bool:
         Validates name
         :return: True or False
         """
-        if re.match("^(.+)/(.+)$", str(self.name)) is not None:
+        if re.match(r"^[a-z-]+\/[a-z-\/]*$", str(self.name)) is not None:
```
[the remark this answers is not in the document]
soapy1: Ah yes, thanks for calling this out! When I did some manual testing against names of the form `<namespace>/<name>` this works. If there is another regex that is used on the server side, it would be good to add here. I obviously don't have visibility into that.

But calling out that this regex needs to be updated because the current regex `"^(.+)/(.+)$"` will match file paths. So like, `can_handle` might return true for like a `environment.yaml`.

docs/source/dev-guide/plugins/environment_specifiers.rst
soapy1: hmmm, I think the not standard file extension is on purpose here. It is a succinct way to denote that some file should be handled by this `RandomSpec` class.

For example, the implementation:

```
class RandomSpec(EnvironmentSpecBase):
                extensions = {".random"}

                def __init__(self, filename: str):
                    self.filename = filename

                def can_handle(self):
                    for ext in RandomSpec.extensions:
                        if self.filename.endswith(ext):
                            return True
```

Each `can_handle` function must only return `True` for environment files that it can read, and must not interfere with other environment spec plugins. So, a plugin that `can_handle` any file with the `.json` extension file would be really bad, because then no other plugin would be able use that extension (without some kind of undefined behaviour). Instead, that type of plugin should be doing something like defining a schema for the environment file and trying to parse the file into that scheme. Then it can return a True/False response based on if that parsing was successful or not.

IMO the point of this example is to outline the bare minimum pieces of code that a plugin author must write in order to have a functioning plugin. Adding complexity, like providing a fully useful plugin with a schema for an environment, etc. does not really serve this purpose. But, having a more complete example would be nice. I thik the `YamlFileSpec` plulgin is a good example. So I'll link to this in the docs.

kenodegard (commented):
I like this change

I want to work to deprecate `conda.env` which means moving these specs elsewhere but I'm ok with waiting until after the binstar deprecation in 25.9

conda/cli/common.py
```diff
@@ -350,3 +359,29 @@ def print_activate(env_name_or_prefix):  # pragma: no cover
             """
         )
         print(message)  # TODO: use logger
+
+
+def validate_env_file_exists(filename: str):
```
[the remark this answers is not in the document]
soapy1: >Should we check for existence if it is a file:// filename?

It does! It checks all the `CONDA_SESSION_SCHEMES`. That includes
```
CONDA_SESSION_SCHEMES = frozenset(
    (
        "http",
        "https",
        "ftp",
        "s3",
        "file",
    )
)
```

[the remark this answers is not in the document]
kenodegard: Right but we aren't then also checking if it exists on disk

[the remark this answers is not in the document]
soapy1: > This is also a pretty general purpose function, are we better off renaming to `validate_exists`

I think `validate_exists` is too general of a name, especially to be in the `cli.common` module. Like, validate that what exists?

This function validates that a file exists on disk, or that it is a remote file that conda knows how to read. So could do something like `validate_file_exists`? Is that an improvement?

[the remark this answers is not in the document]
soapy1: > Right but we aren't then also checking if it exists on disk

oh! I see what you mean. Ya, this could be better. Taking a look :eye:

soapy1:
pre-commit.ci autofix

docs/source/dev-guide/plugins/environment_specifiers.rst
travishathaway: > So, a plugin that can_handle any file with the .json extension file would be really bad, because then no other plugin would be able use that extension (without some kind of undefined behaviour).

This is a bit concerning because I can easily see a situation where multiple plugins would want to define and env specification that both use the `.yaml` file format or even `.txt`. Do we have any way to warn users when they have installed multiple plugins that both handle `.yaml` or `.txt` files?

To add a warning like this, we should simple see if the number `can_handle` matches is greater than one. We could then simply output a warning message saying there were multiple matches and say which env spec handler was chosen. Otherwise, this could lead to an extremely difficult to debug situation.

On top of this, we should explicitly warn of this exact situation in the developer documentation. It looks like you tried to describe this situation in the hint above, but I think it's worth going into more detail about how this can arise and why it's a problem.

I still think it's better to use a real file extension in the example, but I won't block on this.

travishathaway:
I just wanted to reiterate something I said in another comment above...

Currently when two plugins define overly permissive `can_handle` methods (e.g. both want to handle `.yml` files), this can cause indeterminate behavior with no warnings to the user. Before this pull request is merged, I would like to see a warning added to the `get_environment_specifier_handler` when more than one handler is matched. In this case, it's at least clear which handler was chosen and we can send a warning message to the user about possible ways to resolve it (e.g. remove the plugin causing the conflict).

docs/source/dev-guide/plugins/environment_specifiers.rst
soapy1: > Do we have any way to warn users when they have installed multiple plugins that both handle .yaml or .txt files?

I want to call out here that this PR suggests moving away from relying on file extensions as the way that conda chooses which environment_spec plugin to use. The interface suggested in this PR:
* _pushes the responsibility of determining if an environment_spec plugin is valid for a particular environment definition to the plugin author_. The plugin author has the ability to do whatever actions they need to determine if they can read the provided input. Practically, this should be something like, the plugin author has a model of their environment (this is an implementation detail of the plugin) and the `can_handle` function determines if the provided input is compatible with this model.  In this way, users may have multiple plugins that can read files of the same extension. 
* _pushes the reponsibility of determining which environment_spec plugin to use to the conda plugin manager_. The plugin manager must rely on each plugin's implementation of `can_handle` to determine if the provided environment definition has a way of being converted into an environment object that conda can do things with.

>To add a warning like this, we should simple see if the number can_handle matches is greater than one. We could then simply output a warning message saying there were multiple matches and say which env spec handler was chosen. Otherwise, this could lead to an extremely difficult to debug situation.

Good call out! I think this is the responsibility of the plugin manager. Will update this

> On top of this, we should explicitly warn of this exact situation in the developer documentation. It looks like you tried to describe this situation in the hint above, but I think it's worth going into more detail about how this can arise and why it's a problem.

True! I think maybe there are a few ways to improve this documentation, as you point out. Lets do that in a follow up PR so that feedback for the docs and implementation don't get mixed up/lost.

conda/plugins/manager.py
```diff
@@ -476,6 +487,33 @@ def load_settings(self) -> None:
...
+            if hook.environment_spec(filename).can_handle():
+                found.append(hook)
+
+        if len(found) == 1:
+            return found[0]
+        elif len(found) > 0:
+            # raise an error if there is more than one plugin found
+            names = ", ".join([hook.name for hook in found])
+            raise PluginError(
+                f"Multiple environment_spec plugins registered for the environment specified in '{filename}'. Please review plugins: {names}"
```
soapy1, line 509: took a stab at writing a helpful message here. I feel like there are better words to use, but I can't find them. Feel free to push directly to my branch or suggest a better message!

jezdez (dismissed):

soapy1:
pre-commit.ci autofix

jezdez (commented):
Last tweak to the `can_handle` code, then I think this is ready to go in.

conda/env/specs/binstar.py
```diff
@@ -60,7 +61,7 @@ def valid_name(self) -> bool:
...
```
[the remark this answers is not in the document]
soapy1: Awesome, thanks for following up on this!

soapy1:
pre-commit.ci autofix

jezdez (approved):
Yay, thank you @soapy1! 🌟

PR #14726, opened 2025-04-02
add "environment consistency" health check to `conda doctor`

### LINKED ISSUE #14715: Add 'check if environment is consistent with solver' health check to `conda doctor`
### Checklist

### What is the idea?

Per discussion with @jaimergp, the solution to [this](https://github.com/orgs/conda/projects/17/views/6?pane=issue&itemId=102041911) issue from the project board can start with a `conda doctor` health check.

### What should the health check be?
`conda doctor` will get all the entries in `PrefixData.iter_records()`  and submit them to the solver, and check if the solver can find a solution that is exactly what PrefixData says. 
If yes, we say that the environment is consistent else not. 

Would be good if `conda doctor` could also print which packages are causing the inconsistency. 

### Why is this needed?

We want to make `conda doctor` provide even more information to users about their conda environment health. 
So that`conda doctor` is a tool to diagnose inconsistent environments.

### What should happen?

A "environment consistency against the solver" health check is added to `conda doctor`

### Additional Context

_No response_

ForgottenProgramme:
### Description

This PR adds a new health check to `conda doctor` called "environment consistency check". 

resolves https://github.com/conda/conda/issues/14715

#### Todo

- [x] output why the environment is inconsistent
- [x] Write tests
- [x] remove the following text from the outputed health report 
```
Channels:
 - file:///var/folders/tg/gstz8jc94rb75hnd7pfsxjvr0000gp/T/tmpg2n97khs
Platform: osx-arm64
Solving environment: done
``` 

### Checklist - did you ...


conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +180,57 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+        if record.fn.endswith(".conda"):
+            repodata["packages.conda"][record.fn] = record_data
+        else:
+            repodata["packages"][record.fn] = record_data
+
+    # create fake channel with package records from pd
+
+    with TemporaryDirectory() as tmp_dir:
+        noarch_path = Path(tmp_dir) / "noarch"
+        non_noarch_path = Path(tmp_dir) / "subdir"
```
jaimergp, line 207:
```suggestion
        non_noarch_path = Path(tmp_dir) / context.subdir
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +180,57 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+
+        noarch_path.mkdir(parents=True, exist_ok=True)
+        non_noarch_path.mkdir(parents=True, exist_ok=True)
+
+        noarch_repodata = noarch_path / "repodata.json"
+        # non_noarch_repodata = non_noarch_path / "repodata.json"
+
+        noarch_repodata.write_text(json.dumps(repodata))
+
+    fake_channel = Channel(str(noarch_repodata.resolve()))
```
jaimergp, line 217: Channel can the parent of `noarch/` and infer the subdirs, which in this case is, conveniently, `tmp_dir`.

```suggestion
    fake_channel = Channel(tmp_dir)
```

I don't remember if the Solver requires a subdir-aware Channel though, so in that case it would have to be this:

```python
channels = [Channel(str(non_noarch_path)), Channel(str(noarch_path))]
```

But in no case it should be the direct `repodata.json` path.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +180,57 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+
+    specs = History(prefix).get_requested_specs_map()  # get specs from the history file
```
jaimergp, line 219: I'm not sure we want to do this check in particular. This is too lenient (which is ok), but we will also need the stricter version:

```python
specs = [MatchSpec(name=record.name, version=record.version, build=record.build_string) for record in prefix_data.iter_records()]
```

This way we can detect if the user patched the environment with a direct archive install (e.g. `conda install path/to/my/custom/numpy-2-0.conda` and things like that).

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +180,54 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+        if record.fn.endswith(".conda"):
+            pkg_type = "packages.conda"
+        else:
+            pkg_type = "packages"
+
+        if record_data["subdir"] not in repodatas:
+            repodatas[record_data["subdir"]] = {}
+        if pkg_type not in repodatas[record_data["subdir"]]:
+            repodatas[record_data["subdir"]][pkg_type] = {}
+        repodatas[record_data["subdir"]][pkg_type][record.fn] = record_data
```
kenodegard, line 203: An alternative is to use [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault):

```suggestion
        repodata = repodatas.setdefault(record_data["subdir"], {})
        repodata.setdefault(pkg_type, {})[record.fn] = record_data
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +180,54 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+    # create fake channel with package records from pd
+    with TemporaryDirectory() as tmp_dir:
+        for subdir in repodatas:
+            path = Path(tmp_dir) / subdir
+            path.mkdir(parents=True, exist_ok=True)
+            repodata = path / "repodata.json"
+            if repodatas[subdir]["packages.conda"]:
+                repodata.write_text(json.dumps(repodatas[subdir]["packages.conda"]))
+            if repodatas[subdir]["packages"]:
+                repodata.write_text(json.dumps(repodatas[subdir]["packages"]))
```
kenodegard, line 214: Currently you are only writing `packages.conda` or `packages` to the `repodata.json`, we want both:

```suggestion
        for subdir, repodata in repodatas.items():
            path = Path(tmp_dir, subdir)
            path.mkdir(parents=True, exist_ok=True)
            (path / "repodata.json").write_text(json.dumps(repodata))
```

kenodegard (commented):
Need to add tests (both consistent and inconsistent environments) to better understand if this check is doing what we expect it to be doing.

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: @kenodegard, I am using two `if` statements and not `if else`.

jaimergp: `write_text` clears the contents first, doesn't it?

conda/plugins/manager.py
```diff
@@ -423,6 +423,7 @@ def invoke_health_checks(self, prefix: str, verbose: bool) -> None:
                 hook.action(prefix, verbose)
             except Exception as err:
                 log.warning(f"Error running health check: {hook.name} ({err})")
+                raise
                 continue
```
jaimergp, line 427: This is superfluous if you are raising the exception. However, this will prevent the other checks from happening. You might want to accumulate `err` in a list, and _then_ raise with `CondaMultiError`.

ForgottenProgramme: oh this was only for testing purposes. I forgot to remove it, sorry!

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -206,30 +210,30 @@ def consistent_env_check(prefix: str, verbose: bool) -> None:
...
-        # a channel must always contain a "noarch/repodata.json" even if it's empty
-        # create a noarch directory and a populate it with an empty repodata.json in case it doesn't exist.
-        noarch_path = Path(tmp_dir) / "noarch" / "repodata.json"
-        noarch_path.parent.mkdir(parents=True, exist_ok=True)
-        if not noarch_path.exists():
-            noarch_path.write_text(json.dumps({}))
-
+        new_env_path = (
+            Path(tmp_dir) / "environment"
+        )  # a new environment to run the solver in
```
jaimergp, line 215: In general, put the comments _before_ the line. That will also have the side effect of not needing the weird parentheses for formatting.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -206,30 +210,30 @@ def consistent_env_check(prefix: str, verbose: bool) -> None:
...
 
-        # a channel must always contain a "noarch/repodata.json" even if it's empty
-        # create a noarch directory and a populate it with an empty repodata.json in case it doesn't exist.
-        noarch_path = Path(tmp_dir) / "noarch" / "repodata.json"
-        noarch_path.parent.mkdir(parents=True, exist_ok=True)
-        if not noarch_path.exists():
-            noarch_path.write_text(json.dumps({}))
-
+        new_env_path = (
+            Path(tmp_dir) / "environment"
```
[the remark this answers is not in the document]
ForgottenProgramme: okay

conda/plugins/subcommands/doctor/health_checks.py
ForgottenProgramme: done

jjhelmus:
Is a solve really necessary here. Iterating over the PrefixRecords and testing if their dependencies are met by packages in the prefix is sufficient to validate an environment. This process can be done using MatchSpecs and not the solver.

tests/plugins/subcommands/doctor/test_health_checks.py
```diff
@@ -326,3 +394,29 @@ def test_json_cannot_be_loaded(env_ok: tuple[Path, str, str, str, str]):
...
+
+def test_env_consistency_check_pass(
+    env_consistent: tuple[Path, str, str, str, str],
+    capsys: CaptureFixture,
+):
+    """Test that runs for the case when the environment is consistent"""
+    prefix, _, _, _, package = env_consistent
+    consistent_env_check(prefix=prefix, verbose=True)
+    captured = capsys.readouterr()
+    assert f"{OK_MARK} The environment is consistent.\n" in captured.out
```
kenodegard, line 407: Let's refactor this to use `tmp_env` instead, like we do in `test_env_consistency_check_fails`.

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -175,6 +181,64 @@ def requests_ca_bundle_check(prefix: str, verbose: bool) -> None:
...
+
+    # create a repodatas dict with noarch subdir already created as it is pre-requisite for creating a channel
+    repodatas = {"noarch": {}}
+
+    # TODO create a list using iter_records and check its length, if 0 exit
+
+    # segregate the package records based on subdir/architecture and package type
+    for record in pd.iter_records():
+        record_data = dict(record.dump())
+        if record.fn.endswith(".conda"):
```
kenodegard, line 202: let's use `conda.base.constants.CONDA_PACKAGE_EXTENSION_V2`:

```suggestion
        if record.fn.endswith(CONDA_PACKAGE_EXTENSION_V2):
```

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -6,13 +6,19 @@
...
 import os
+from contextlib import nullcontext, redirect_stdout
 from logging import getLogger
 from pathlib import Path
+from tempfile import TemporaryDirectory
 
 from requests.exceptions import RequestException
 
+from conda.models.channel import Channel
+from conda.models.match_spec import MatchSpec
```
kenodegard, line 17:
```suggestion
from ....models.channel import Channel
from ....models.match_spec import MatchSpec
```

kenodegard (changes requested):

jjhelmus:
Can a similar check be done by validating that all dependencies for each packages installed in the environment in question are met? This seems favorable given it can be accomplished without engaging the solver or creating mock channels.

The following snippet performs this check:
``` Python
def consistent_env_check(prefix: str, verbose: bool) -> None:

    # get the plugin manager from context
    pm = context.plugin_manager

    # find virtual packages
    virtual_packages = {str(p.name): p for p in  pm.get_virtual_package_records()}

    # get prefix data
    prefix_data = PrefixData(prefix)

    issues = []
    # look at each package in the environment for conflicts
    prefix_record: PrefixRecord
    for prefix_record in prefix_data.iter_records():
        for depend in prefix_record.depends:
            matchspec = MatchSpec(depend)
            package_record_with_name = prefix_data.get(
                matchspec.name,
                default=virtual_packages.get(matchspec.name, None),
            )
            if package_record_with_name is None:
                issues.append(f"No package {matchspec.name} required by {prefix_record} ({matchspec})")
                continue
            if not matchspec.match(package_record_with_name):
                issues.append(f"{package_record_with_name} does not match {prefix_record} ({matchspec})")
                continue
    if issues:
        print(
            f"{X_MARK} The environment is not consistent due to the following issues:\n",
            ''.join(issues),
        )
    else:
        print(f"{OK_MARK} The environment is consistent.\n")
```

PR #14733, opened 2025-04-04
V2 Environment spec

peytondmurray:
### Description

This PR adds a class that can parse a V2 environment spec of the form https://gist.github.com/jaimergp/4209c4c90d51b1bb07fe7293095f7c70.

This PR is intended as an initial implementation, and as a playground for testing this spec. Please note the **spec implemented here is not intended to be final**.

### Notes

- I added an `EnvironmentBase` class which defines a few common methods that I think it would be nice for all subclasses to implement, because they're used elsewhere. But I think it would also be nice to use this as an opportunity to clean up the API a little bit. For example, there are a number of functions in `conda/env/env.py` which IMO should be classmethods, including `from_environment`, `from_yaml`, and `from_file`. These generate `EnvironmentV1` instances, but it would be nice to have the `EnvironmentBase` be an ABC that requires subclasses to implement these as well.
- I implemented the spec as it is in https://gist.github.com/jaimergp/4209c4c90d51b1bb07fe7293095f7c70, but I think there are some changes that would clean up the implementation that would be good to consider - see below.

### Checklist - did you ...

### Some thoughts about the spec itself

- I'd like to advocate for _not_ using yaml. As widespread as it is, I've worked on a number of projects which use yaml configuration files where users (including me) waste time getting whitespace just right so that it parses, fight with implicit strings, and workaround all the other yaml pitfalls that we encounter on the regular. JSON doesn't have comments, so I'd vote not to use it. TOML isn't perfect especially for complex configuration files, but for environment.yml files it could work, especially since tomllib has been part of the standard library since Python 3.11.
- Currently, commonly used settings like `channels: ...` live at the top level of the project. _Some_ of these override user configuration that lives at `.condarc`, but not all of them (e.g. `name` and `description`). I think it makes sense to nest these into a `config: ...` section, keeping the configuration variables that override `.condarc` settings together. In this PR, these live in the `EnvironmentConfig` object, while options like `description` etc live in `EnvironmentV2.options`.
- There's currently a proposal for conditional requirements that look like (in dict form)

```python
{
  "if": "__win",
  "then": "flask",
}
```

Is there a chance that it would make sense to make the `then` a list instead? For example:

```python
{
  "if": "__win",
  "then": [
    "flask",
    "matplotlib",
  ],
}
```

### DIFF
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -92,7 +92,7 @@ def from_environment(
     if from_history:
         history = History(prefix).get_requested_specs_map()
         deps = [str(package) for package in history.values()]
-        return Environment(
+        return EnvironmentV1(
             name=name,
             dependencies=deps,
             channels=list(context.channels),
@@ -133,7 +133,7 @@ def from_environment(
             canonical_name = prec.channel.canonical_name
             if canonical_name not in channels:
                 channels.insert(0, canonical_name)
-    return Environment(
+    return EnvironmentV1(
         name=name,
         dependencies=dependencies,
         channels=channels,
@@ -154,7 +154,7 @@ def from_yaml(yamlstr, **kwargs):
         for key, value in kwargs.items():
             data[key] = value
     _expand_channels(data)
-    return Environment(**data)
+    return EnvironmentV1(**data)
 
 
 def _expand_channels(data):
@@ -278,7 +278,11 @@ class EnvironmentConfig:
         return "\n".join(lines)
 
 
-class EnvironmentSpecV2:
+class EnvironmentBase:
+    """A class representing an ``environment.yaml`` file"""
+
+
+class EnvironmentV2(EnvironmentBase):
     """A class representing a V2 environment.yaml."""
 
     def __init__(
@@ -296,7 +300,7 @@ class EnvironmentSpecV2:
         self.config = config if config else EnvironmentConfig()
 
     @classmethod
-    def from_file(cls, filename: os.PathLike[str]) -> EnvironmentSpecV2:
+    def from_file(cls, filename: os.PathLike[str]) -> EnvironmentV2:
         with open(filename) as f:
             data = yaml_safe_load(f.read())
 
@@ -332,6 +336,11 @@ class EnvironmentSpecV2:
         result.update(self.options)
         return result
 
+    def to_yaml(self, stream=None) -> Any | None:
+        out = yaml_safe_dump(self.to_dict, stream)
+        if stream is None:
+            return out
+
     def to_file(self, filename: os.PathLike):
         with open(filename, "w") as f:
             yaml_safe_dump(self.to_dict(), stream=f)
@@ -362,7 +371,7 @@ class EnvironmentSpecV2:
         return "\n".join(lines)
 
 
-class Environment:
+class EnvironmentV1(EnvironmentBase):
     """A class representing an ``environment.yaml`` file"""
 
     def __init__(
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -2,11 +2,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Environment object describing the conda environment.yaml file."""
 
+from __future__ import annotations
+
 import json
 import os
 import re
+from dataclasses import dataclass, field
+from functools import total_ordering
 from itertools import chain
 from os.path import abspath, expanduser, expandvars
+from typing import Any
 
 from ..base.context import context
 from ..cli import common, install
@@ -205,6 +210,116 @@ class Dependencies(dict):
         """Add a package to the ``Environment``"""
         self.raw.append(package_name)
         self.parse()
+
+
+@total_ordering
+class Requirement:
+    def __init__(self, spec: str | dict[str, str]):
+        if isinstance(spec, str):
+            self.spec = spec
+            self.condition = None
+        elif isinstance(spec, dict):
+            for key in ["if", "then"]:
+                if key not in spec:
+                    raise ValueError(f"Conditional requirement missing an `{key}` key.")
+
+            self.spec = spec["then"]
+            self.condition = spec["if"]
+        else:
+            raise ValueError(f"Invalid requirement: {spec}")
+
+    def __repr__(self) -> str:
+        result = self.spec
+        if self.condition:
+            result = f"{result} (if {self.condition})"
+        return result
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, Requirement):
+            return self.spec == other.spec
+
+        raise NotImplementedError
+
+    def __lt__(self, other: Requirement) -> bool:
+        return self.spec < other.spec
+
+
+class Requirements:
+    def __init__(
+        self,
+        raw_requirements: list[str | dict[str, str]],
+        raw_pypi_requirements: list[str],
+    ):
+        self.requirements: list[Requirement] = []
+        self.pypi_requirements: list[Requirement] = []
+
+        for raw_req in raw_requirements:
+            self.requirements.append(Requirement(raw_req))
+
+    def __repr__(self) -> str:
+        lines = []
+        for req in sorted(self.requirements):
+            lines.append(repr(req))
+
+        return "\n".join(lines)
+
+
+@dataclass
+class EnvironmentConfig:
+    channels: list[str] | None = field(default_factory=list)
+    channel_priority: str | None = None
+    repodata_fn: str | None = None
+    variables: dict[str, str] = field(default_factory=dict)
+
+
+class EnvironmentSpecV2:
+    """A class representing a V2 environment.yaml."""
+
+    def __init__(
+        self,
+        requirements: Requirements,
+        name: str | None = None,
+        **options,
+    ):
+        self.name = name
+        self.requirements = requirements
+        self.options = options
+
+    @classmethod
+    def from_file(cls, filename: os.PathLike[str]) -> EnvironmentSpecV2:
+        with open(filename) as f:
+            data = yaml_safe_load(f.read())
+
+        breakpoint()
+        requirements = Requirements(
+            raw_requirements=data.get("requirements"),
+            raw_pypi_requirements=data.get("pypi-requirements"),
+        )
+
+        groups = data.get("groups")
+
+        return cls(
+            name=data.get("name"),
+            requirements=requirements,
+            version=data.get("version", 2),
+            description=data.get("description"),
+            config=EnvironmentConfig(
+                channels=data.get("channels"),
+                channel_priority=data.get("channel-priority"),
+                repodata_fn=data.get("repodata-fn"),
+                variables=data.get("variables", {}),
+            ),
+        )
+
+    def to_dict(self) -> dict:
+        result = {}
+        result["name"] = self.name
+        result.update(self.options)
+        return result
+
+    def to_file(self, filename: os.PathLike):
+        with open(filename, "w") as f:
+            yaml_safe_dump(self.to_dict(), stream=f)
 
 
 class Environment:
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -217,6 +217,18 @@ class Dependencies(dict):
 
 @total_ordering
 class Requirement:
+    """A single requirement for a package.
+
+    Can either be a single spec string (e.g. "flask"), or a dictionary containing
+    a conditionally present package, and a spec to include if that condition is present.
+    For example:
+
+    {
+        "if": "__linux",    # <-- if __linux is present, also include flask as a requirement
+        "then": "flask",
+    }
+    """
+
     def __init__(self, spec: str | dict[str, str]):
         if isinstance(spec, str):
             self.spec = spec
@@ -246,8 +258,15 @@ class Requirement:
     def __lt__(self, other: Requirement) -> bool:
         return self.spec < other.spec
 
+    def serialize(self) -> str | dict[str, str]:
+        if self.condition:
+            return {"if": self.condition, "then": self.spec}
+        return self.spec
+
 
 class Requirements:
+    """A set of requirements for a package."""
+
     def __init__(
         self,
         raw_requirements: list[str | dict[str, str]],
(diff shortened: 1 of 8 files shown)


conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+        groups: dict[str, Requirements] | None = None,
+        name: str | None = None,
+        config: EnvironmentConfig | None = None,
+        **options,
+    ):
+        self.name = name
+        self.requirements = requirements if requirements else Requirements()
+        self.groups = groups if groups else {}
+        self.options = options
+        self.config = config if config else EnvironmentConfig()
```
soapy1, line 415: How does the config for an environment interact with a users `.condarc`?

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+
+        requirements = ["=".join((a.name, a.version)) for a in conda_precs]
+        pypi_requirements = [f"{a.name}=={a.version}" for a in pip_precs]
+
+        channels = [prec.channel.canonical_name for prec in conda_precs]
+        name = get_env_name(prefix)
+
+        return cls(
+            name=name if name else None,
+            requirements=Requirements.from_strs(requirements, pypi_requirements),
```
soapy1, line 607: What does `requirements` mean in this case? It seems like in other `from_*` methods the requirements are the user requested specs. For example in `from_history` they come from `History(prefix).get_requested_specs_map()`. But in this case, it looks like the requirements are all the conda and pip packages installed in the prefix. Does the environment object need to have a notion between the difference between these 2 things?

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+
+
+class EnvironmentConfig:
+    """A class that contains configuration variables for an environment.
+
+    These variables override user configuration, e.g. settings in .condarc.
+    """
+
+    def __init__(self, **options):
+        self.options = options
```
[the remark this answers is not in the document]
peytondmurray: Any environment metadata that _isn't_ a configuration variable that overrides settings in `.condarc`. For example, I think `description` and arguably `name` should be here, whereas `channels` should be in the `EnvironmentConfig`.

conda/env/env.py
peytondmurray: I think it should override `.condarc`. For example when users specify `channels` in an environment.yml, they expect it to override the default channels and whatever they've specified in `.condarc`.

conda/env/env.py
peytondmurray: Yeah, this is the difficulty/distinction between loading from history vs a prefix. If you're just trying to create an `EnvironmentV2` from a prefix, to my knowledge there's no way to distinguish the specs the user requested vs the specs that were installed (except by looking at the `history`, but there's a separate function for that).

There's added difficulty in that the `history` file doesn't record PyPI packages. If you have ideas about how these two functions could be improved, I'd be happy to implement it.

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+            name=name if name else None,
+            requirements=Requirements.from_strs(requirements, pypi_requirements),
+            config=EnvironmentConfig(
+                channels=channels,
+                variables=variables,
+            ),
+        )
+
+
+class EnvironmentV1(EnvironmentBase):
```
jaimergp, line 615: This has to be named `Environment` for backwards compatibility. You can simply add an alias if you want, assuming this new class is API compatible with the old `Environment` (e.g. didn't change signatures in existing methods, or removed methods, attributes, etc).

conda/env/env.py
```diff
@@ -291,3 +696,22 @@ def print_result(args, prefix, result):
             common.stdout_json_success(prefix=prefix, actions=actions)
     else:
         install.print_activate(args.name or prefix)
+
+
+def get_env_name(prefix: os.PathLike) -> str:
```
jaimergp, line 701: I'm sure there's functionality for this somewhere else in the code. Maybe not as a function, but I've seen this logic elsewhere. I think I'll address this as part of https://github.com/conda/conda/pull/14750, though, so let's leave this thread open.

conda/env/env.py
jaimergp: Using `Environment` would also reduce the diff of this PR a bit :)

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+        if self.condition:
+            return {"if": self.condition, "then": self.spec}
+        return self.spec
+
+    def __str__(self) -> str:
+        return repr(self)
+
+
+class Requirements:
+    """A set of conda and PyPI requirements for a package."""
```
jaimergp, line 286: I don't like having these two aspects combined in a single class 🤔 They should be separate imo. Any particular reason why you merged them here?

conda/env/env.py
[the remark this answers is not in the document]
jaimergp: I don't think these fit here. Shouldn't name and description be top level attributes in the EnvironmentV2 class?

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+    """A class representing a V2 environment.yaml."""
+
+    def __init__(
+        self,
+        requirements: Requirements | None = None,
+        groups: dict[str, Requirements] | None = None,
+        name: str | None = None,
+        config: EnvironmentConfig | None = None,
+        **options,
+    ):
```
jaimergp, line 410: Where's the description?

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+        pip_precs = sorted(
+            (
+                *grouped_precs.get(PackageType.VIRTUAL_PYTHON_WHEEL, ()),
+                *grouped_precs.get(PackageType.VIRTUAL_PYTHON_EGG_MANAGEABLE, ()),
+                *grouped_precs.get(PackageType.VIRTUAL_PYTHON_EGG_UNMANAGEABLE, ()),
+            ),
+            key=lambda x: x.name,
+        )
+
+        requirements = ["=".join((a.name, a.version)) for a in conda_precs]
```
jaimergp, line 599: I think `PackageRecord` has a `to_match_spec()` method or similar. And also, why are we not adding the build string?

conda/env/env.py
```diff
@@ -207,7 +219,400 @@ def add(self, package_name):
...
+        """Create an Environment from the prefix of an environment on disk.
+
+        :param prefix: Prefix of the target conda environment
+        :return: EnvironmentV2 object containing all the conda packages in the prefix
+        """
+        pfd = PrefixData(prefix, pip_interop_enabled=True)
+        variables = pfd.get_environment_env_vars()
+
+        precs = tuple(PrefixGraph(pfd.iter_records()).graph)
+        grouped_precs = groupby(lambda x: x.package_type, precs)
```
jaimergp, line 580: Why grouped? The toposort above is enough imo.

jaimergp:
> TOML isn't perfect especially for complex configuration files, but for environment.yml files it could work, especially since tomllib has been part of the standard library since Python 3.11.

I'm fine with that!

> For example, there are a number of functions in conda/env/env.py which IMO should be classmethods, including from_environment, from_yaml, and from_file. These generate EnvironmentV1 instances, but it would be nice to have the EnvironmentBase be an ABC that requires subclasses to implement these as well.

Agreed here, but let's focus on v2 first.

> There's currently a proposal for conditional requirements that look like (in dict form) [...] Is there a chance that it would make sense to make the then a list instead?

Yes, that can be done. We just need to merge those in the parent list.

> Currently, commonly used settings like channels: ... live at the top level of the project. Some of these override user configuration that lives at .condarc, but not all of them (e.g. name and description). I think it makes sense to nest these into a config: ... section, keeping the configuration variables that override .condarc settings together. In this PR, these live in the EnvironmentConfig object, while options like description etc live in EnvironmentV2.options

I'm neutral to group condarc-derived options under `config` or similar, but I _do_ want `name` and `description` to be top-level both in the spec and in the class signature and attributes. I don't really want to do `env.options.name` instead of `env.name`.

jaimergp:
Also, I might have missed this, but how does this interplay with `conda env create`? Can I just use it now, or is the plugin bit not implemented here? That would be understandable given how https://github.com/conda/conda/pull/14710 is still in progress, but wanted to make sure that was not an oversight on my end.

conda/env/env.py
peytondmurray: Because it makes sense conceptually - a package has a set of requirements. Those requirements can be either conda or pypi requirements or both. Maybe this shouldn't be a separate class, and instead each `Requirement` can be either a `PyPIRequirement` or a `CondaRequirement`? I'll try that approach now.

conda/env/env.py
peytondmurray: Ack, I can add a description now - usually I don't for `__init__` because it's always the same: "Initialize an instance of <whatever class>". But here I think having annotations for e.g. `options` would be helpful.

conda/env/env.py
peytondmurray: This was actually ripped from `conda.env.nev.from_environment`, so that's why it may look familiar. I'll add in the build string now.

conda/env/env.py
peytondmurray: No idea, this is just lifted from `from_environment`. I'll simplify this now.

conda/env/env.py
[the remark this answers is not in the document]
peytondmurray: Okay, sorry but I was looking at the wrong line number previously. Let me revise my answer from before:

- The `EnvironmentConfig` contains anything that needs to override `.condarc`, such as `channels`
- The `EnvironmentV2` also contains an `options` attribute which is used for other environment-specific metadata.

---

Normally I'd agree with you about `name` being a top level attribute of the `EnvironmentV2`, but not every environment has a name (which I actually think is not a good thing). The description is currently optional as well. You can also imagine other metadata that you'd like to attach to an environment (for example, tags), and we don't yet have a place for that.

What I'm really saying is `name` and `description` get first-class treatment insofar as they have special dedicated attributes on the environment, but then they aren't actually used everywhere. Furthermore, the current system doesn't really work for any additional metadata one could want to attach to an environment.

If I can propose a path forward:

1. Every `EnvironmentV2` has a `name` attribute which is not optional
2. Any other metadata can go in an `options` or `metadata` attribute which is a potentially empty dict of other python primitives. Users can store additional info like environment tags or permissions information here.

Thoughts?

peytondmurray:
> I'm neutral to group condarc-derived options under config or similar, but I do want name and description to be top-level both in the spec and in the class signature and attributes. I don't really want to do env.options.name instead of env.name.

I'm fine if people want it to be this way. My main issue with this is that specfile as it stands doesn't separate settings that you override in .condarc with environment names or descriptions or anything else. Conceptually they are doing very different things for them all to be top level.

```yaml
name: test-env
requirements:
  - <whatever>
channels:           # <-- IMO this should be totally separate, nested under a `config` key or something
  - conda-forge
```

conda/env/env.py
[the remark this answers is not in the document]
soapy1: For (1) I think sometimes when a `name` is not associated with an environment the `prefix` is used instead. There are a bunch of places where `env_name_or_prefix` is used. Is this ok to be the top level parameter instead of forcing the `name` to be required?

For (2) I  think the conda config stuff is pretty complicated. It might be nice to seperate this information from other environment metadata type information.

conda/env/env.py
```diff
@@ -207,7 +222,430 @@ def add(self, package_name):
...
+        raise NotImplementedError
+
+    def to_yaml(self) -> str | None:
+        raise NotImplementedError
+
+    def save(self):
+        raise NotImplementedError
+
+
+class EnvironmentV2(EnvironmentBase):
```
soapy1, line 340: Should this be implemented as a EnvSpec eg. in `conda.env.specs` instead of as a new type of environment?

peytondmurray: That would be fine by me!

conda/env/env.py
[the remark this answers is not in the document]
peytondmurray: > For (1) I think sometimes when a name is not associated with an environment the prefix is used instead. There are a bunch of places where env_name_or_prefix is used. Is this ok to be the top level parameter instead of forcing the name to be required?

Is the fact that environment names are not required a good thing? Maybe this isn't the place to make radical changes, but IMO the current design is confusing, and I guess I see this PR as a chance to improve things. For example, if I have an unnamed environment, when I do `conda env list` I get a blank space where the environment name should be:

```bash
$ conda env list

# conda environments:
#
base                 * /home/pdmurray/Desktop/workspace/conda/devenv/envs/devenv-3.10-c
test_env               /home/pdmurray/Desktop/workspace/conda/devenv/envs/devenv-3.10-c/envs/test_env
                       /opt/miniforge
```

But internally `conda` refers to this environment by its prefix. However there's no way to search for it by name, and functionally speaking it's very annoying to activate or modify this prefix since I have to specify the path. If I could have conda behave any way I wanted, I'd make it so that environments _must_ have unique names. Is there an actual use case for having a multiple unnamed environments living in various prefixes?

> For (2) I think the conda config stuff is pretty complicated. It might be nice to seperate this information from other environment metadata type information.

Yep, that's what I'm proposing here - stuff that overrides conda configuration (env variables, channels, channel priority, etc) goes into `EnvironmentV2.config`. All environment metadata (e.g. tags, permissions, whatever) goes in `EnvironmentV2.metadata`.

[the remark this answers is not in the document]
jjhelmus: >  Is there an actual use case for having multiple unnamed environments living in various prefixes?

Absolutely. I use local environments created using `-p,--prefix` to develop projects where the environment is local to the source code or project working directory.  The workflow is:

```
git clone <repo>
cd <project>
conda create --prefix ./env <environment-specs>
conda activate ./env
```

I tend to rarely used named environment outside of my `base` environment and a working environment that is activated as part of shell startup.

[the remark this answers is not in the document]
jjhelmus: That said `prefix` rather than name being used in an input file such as `environment.yaml` is not common and I would likely override this anywhere unless it was part of the setup on a container or as part of a deployment strategy.

conda/env/env.py
soapy1: Just want to clarify my question above. I'm getting a bit confused between what ought to be a user facing environement spec and an internal representation of an environment. Currently, the `conda.env.env.Environment` class is pulling double duty and representing both of these ideas.

Thinking about this more, I think it's a good idea to seperate these two ideas. Since the environment spec plugin stuff has been merged, I think the user facing (environment file definition) part of this can be implemented as a plugin.

PR #14750, opened 2025-04-09
Teach `PrefixData` how to perform prefix checks and validations

### LINKED ISSUE #14747: Consolidate prefix validation logic
### Checklist

### What is the idea?

`conda`'s CLI runs a few checks to make sure that the target prefix of an operation exists, is in good shape, etc. There are a few utility functions spread around the code base, with overlapping checks, scopes and responsibilities. This is not only inefficient (e.g. repeated filesystem calls to check whether `prefix` exists), but also difficult to maintain and improve. @soapy1 has been digging this up in a couple PRs (e.g. https://github.com/conda/conda/pull/14670, which brought up [this review](https://github.com/conda/conda/pull/14670#pullrequestreview-2742443805)).

Utility functions that check prefix paths in some way:

- [`check_prefix()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L96)
- [`check_protected_dirs()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_rename.py#L72)
- [`validate_prefix()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/common.py#L254)
- [`validate_prefix_is_writable()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/common.py#L271)
- [`validate_prefix_exists()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L68)
- [`validate_new_prefix()`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L77)

And there are a few more checks without a dedicated function in the CLI layer:

- In [`conda list`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_list.py#L230-L233) and [`--explicit`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_list.py#L263-L266)
- [`conda compare`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_compare.py#L67) (which already uses `PrefixData`!)
- `conda [env]create|install|update|remove` have too many too list. [Example (1)](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_create.py#L131), [(2)](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L106), [(3)](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L127), [(4)](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/install.py#L243)
- [`conda activate`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/activate.py#L344)
- The [`envs_manager` helper](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/core/envs_manager.py#L108)
- Strange [`which_prefix`](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/cli/main_package.py#L293) helper in `conda package`
- [Linking logic](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/core/link.py#L976)
- [`.nonadmin` marker logic](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/misc.py#L231) for Windows

... and that's just checking for `os.path.isdir()`.

### Why is this needed?

This will also be a problem with the implementation of frozen environments (CEP 22) for `base` env protection. @travishathaway also noticed the "entanglement" in his `conda-protect` plugin (see [comment](https://github.com/conda-incubator/conda-protect/blob/9ca71859290c89cb7cb17abf51479de30fd20fcb/conda_protect/main.py#L259-L260)).

### What should happen?

I'd like to propose a refactor where we consolidate all these utility functions in a class, where we can control the common logic better.

It should have these properties:

- Common logic for path manipulation and checks.
- Awareness of stale caches.
- Methods for fine-grained checks.
- Single point of access to all prefix checks. Just finding all the utility functions (listed above) was a bit of a pain.

I think the best way to place this in is `PrefixData`, where we already have an [`is_writable` property](https://github.com/conda/conda/blob/3259a9c914446a44b12e7c4fa191972892b8e394/conda/core/prefix_data.py#L259).

### Additional Context

_No response_

jaimergp:
### Description

Closes #14747 

### Checklist - did you ...


jaimergp:
I'm finding some inconsistencies in the current design, where we are coupling existing and non-existing environments in the same class. Thinking about it, I think it'd be better to decouple these two aspects in two separate classes:

- The existing `PrefixData`, which would only concern _existing_ environments and their records.
- A new `PrefixState`, that would take care of everything else, like variables, configuration, nonadmin and frozen markers, etc. Maybe it's also a good place for name and path validators, but I need to think about that. If not, maybe these validators should be free functions.

To be continued.

jaimergp:
I took another look and the previously existing methods in `PrefixData` do not necessarily assumed existence of the environment. Splitting the new methods to a new class didn't clean it up too much either, and created more questions to the developer, like "is this method in State or in Data?", so I think I'll leave it as is for now.

@conda/conda-maintainers, I think this is ready for review.

jaimergp:
pre-commit.ci autofix

conda/core/prefix_data.py
```diff
@@ -84,6 +100,35 @@ def __init__(
             else context.pip_interop_enabled
         )
 
+    @classmethod
+    def from_name(cls, name: str, **kwargs):
```
peytondmurray, line 104:
```suggestion
    def from_name(cls, name: str, **kwargs) -> PrefixData:
```

conda/core/prefix_data.py
```diff
@@ -84,6 +100,35 @@ def __init__(
...
+        if "/" in name or "\\" in name:
+            raise CondaValueError("Environment names cannot contain path separators")
+        try:
+            return cls(locate_prefix_by_name(name))
+        except EnvironmentNameNotFound:
+            cls(name).validate_name()
+            return cls(Path(_first_writable_envs_dir(), name), **kwargs)
+
+    @classmethod
+    def from_context(cls, validate: bool = False):
```
peytondmurray, line 114:
```suggestion
    def from_context(cls, validate: bool = False) -> PrefixData:
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
             # neither prefix exists, raw comparison
             return self.prefix_path.resolve() == other.prefix_path.resolve()
 
+    def exists(self):
```
peytondmurray, line 145:
```suggestion
    def exists(self) -> bool:
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+        try:
+            return self.prefix_path.is_dir()
+        except OSError:
+            return False
+
+    def is_environment(self):
```
peytondmurray, line 151:
```suggestion
    def is_environment(self) -> bool:
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+        try:
+            return self._magic_file.is_file()
+        except OSError:
+            return False
+
+    def is_base(self):
```
peytondmurray, line 157:
```suggestion
    def is_base(self) -> bool:
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+        return paths_equal(str(self.prefix_path), context.root_prefix)
+
+    @property
+    def is_writable(self):
```
peytondmurray, line 161:
```suggestion
    def is_writable(self) -> bool:
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+        self.assert_exists()
+        if not self.is_environment():
+            raise DirectoryNotACondaEnvironmentError(self.prefix_path)
+
+    def assert_writable(self):
+        self.assert_environment()
+        if not file_path_is_writable(self._magic_file):
+            raise EnvironmentNotWritableError(self.prefix_path)
+
+    def validate_path(self, expand_path: bool = False) -> Path:
```
peytondmurray, line 184:
```suggestion
    def validate_path(self, expand_path: bool = False):
```

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+            )
+
+        if " " in prefix_str:
+            log.warning(
+                "Environment paths should not contain spaces. Prefix: '%s'",
+                prefix_str,
+            )
+        parent = self.__class__(self.prefix_path.parent)
+        if parent.is_environment():
+            raise CondaValueError(
```
peytondmurray, line 202: Should environments ever be able to be nested? I can't tell if there's a legitimate use case for an environment to contain another environment at any level at all.

conda/core/prefix_data.py
```diff
@@ -380,6 +491,9 @@ def _load_site_packages(self):
 
         return new_packages
 
+    # endregion
```
peytondmurray, line 494: Leftover from debugging?

tests/cli/test_main_rename.py
```diff
@@ -75,8 +77,12 @@ def test_rename_by_name_success(
     env_one: str,
     env_rename: str,
 ):
-    conda_cli("rename", "--name", env_one, env_rename)
+    import sys
 
+    out, err, rc = conda_cli("rename", "--name", env_one, env_rename)
+    assert not rc
+    print(out)
+    print(err, file=sys.stderr)
```
peytondmurray, line 85:
```suggestion
```

peytondmurray (changes requested):

peytondmurray:
Thanks for this - mostly minor changes here, otherwise looks good! :rocket:

conda/core/prefix_data.py
```diff
@@ -97,6 +142,80 @@ def __eq__(self, other: Any) -> bool:
...
+    def is_writable(self):
+        if self.__is_writable == NULL:
+            if not self.is_environment():
+                is_writable = None
+            else:
+                is_writable = file_path_is_writable(self._magic_file)
+            self.__is_writable = is_writable
+        return self.__is_writable
+
+    def assert_exists(self):
```
travishathaway, line 170: I'm not super excited about the names for these three methods. Asserts have a pretty specific meaning in Python, and I wouldn't want to confuse the caller by having them think that they are getting some sort of sub-class of an assertion error here or that the method itself is running a series of `assert` statements.

How about we take a lead from the beloved `requests` API, specifically the `raise_for_status` method on the response? We could call these methods:

- `exists_or_raise`
- `is_environment_or_raise`
- `is_writable_or_raise`

That way it's pretty clear what's happening within the method itself and the caller can guess that they will have to handle some sort of exception.

conda/core/prefix_data.py
travishathaway, line 104: If we expect these to be used from outside of this project (which they should!), we need to make sure to add a nice doc string to the method clearly stating why this method should be used and what it returns.

conda/core/prefix_data.py
travishathaway, line 114: This should also get a nice doc string. When I first saw this method in the code itself, I was left wondering, "okay, but which context are you talking about?" Let's make sure the doc string clarifies that so developers can easily find it.

conda/core/prefix_data.py
```diff
@@ -84,6 +100,35 @@ def __init__(
...
+    @property
+    def name(self) -> str:
+        if self == PrefixData(context.root_prefix):
+            return ROOT_ENV_NAME
+        for envs_dir in context.envs_dirs:
+            if paths_equal(envs_dir, self.prefix_path.parent):
+                return self.prefix_path.name
+        return ""
+
+    # region Checks
```
travishathaway, line 130:
```suggestion
```
🧹 ?

conda/core/prefix_data.py
```diff
@@ -84,6 +100,35 @@ def __init__(
...
+    @classmethod
+    def from_context(cls, validate: bool = False):
+        inst = cls(context.target_prefix)
+        if validate:
+            inst.validate_path()
+            inst.validate_name()
+        return inst
+
+    @property
+    def name(self) -> str:
```
travishathaway, line 122: It would be good to document this too if we expect others to be using it.

Another thing that would be nice would be explaining what an empty string means. I'm pretty sure it means it's not a named environment, but totally not clear by reading the code itself.

travishathaway (changes requested):
Looking good so far, and I haven't reviewed all of it yet, but I wanted to cut my review short so you could focus on the changes I requested so far. Please ping me to discuss or when you are ready for me to take another look!

conda/base/context.py
```diff
@@ -2196,10 +2199,12 @@ def determine_target_prefix(ctx, args=None):
...
         return expand(prefix_path)
     else:
-        return validate_prefix_name(prefix_name, ctx=ctx)
+        from ..core.prefix_data import PrefixData
+
+        return str(PrefixData.from_name(prefix_name).prefix_path)
 
 
-def _first_writable_envs_dir():
+def _first_writable_envs_dir(create=True):
```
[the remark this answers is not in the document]
jaimergp: I was experimenting with using this function without side effects in `PrefixData.from_name()`. I may use it there with `create=False` but I haven't decided yet.

conda/core/prefix_data.py
```diff
@@ -57,11 +70,13 @@ def __call__(
     ):
         if isinstance(prefix_path, PrefixData):
             return prefix_path
-        elif (prefix_path := Path(prefix_path)) in PrefixData._cache_:
-            return PrefixData._cache_[prefix_path]
+        prefix_path = Path(prefix_path)
+        cache_key = prefix_path, pip_interop_enabled
+        if cache_key in PrefixData._cache_:
```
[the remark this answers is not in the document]
jaimergp: Good catch, thanks.

conda/core/prefix_data.py
jaimergp: No, this is a standard IDE feature to navigate the minimap and fold related code blocks. See https://www.codu.co/articles/using-regions-in-vs-code-to-organize-your-code-womhkang

conda/core/prefix_data.py
jaimergp: I do like `assert_*`. That very specific meaning is exactly what I mean: raise if the following clause is not true. The `*_or_raise` form sounds a bit too Rust-y for my taste.

My intent here was to have booleans returned by the `is_*` family, and exceptions raised by the `assert_*` family. I also thought of `check_*` for the latter, but I felt that `assert` was clearer.

I'll let other folks chime in here to resolve the "tie", but I do like `assert_*` a lot.

conda/core/prefix_data.py
jaimergp: Yep, didn't put the time on these yet because I didn't know how many of these new API endpoints would survive a review :P Looks like we are keeping them so I'll write them down.

conda/core/prefix_data.py
jaimergp: Environments can be nested in this way:

```
$CONDA_ROOT
$CONDA_ROOT/envs/new-conda-env
$CONDA_ROOT/envs/new-conda-env/envs/another-conda-subenv
```

Not that we _should_ do that, but it's perfectly legal. We use this in the local dev setup, not sure if intentionally, but it does happen there.

What we are avoiding here is having this situation:

```
$CONDA_ROOT
$CONDA_ROOT/my-new-env
$CONDA_ROOT/envs/new-conda-env/another-conda-subenv
```

jaimergp: This type of nesting, btw, is because `envs_dirs` includes the `base`-like (it has `conda` in it) environment path as one of the possible locations, as you have seen in your PRs. This is one of the reasons I want to remove that coupled behaviour.

conda/core/prefix_data.py
travishathaway: I'll leave this thread open in case anyone else wants to chime in on it.

jaimergp: FWIW, a couple Python function definition searches that are relevant to the discussion:

- [`assert_*`](https://github.com/search?q=%2Fdef+assert_%5CS%2B%2F+language%3APython+NOT+is%3Afork&ref=opensearch&type=code): 134k files
- [`*_or_raise`](https://github.com/search?q=%2Fdef+%5CS%2B_or_raise%2F+language%3APython+NOT+is%3Afork&ref=opensearch&type=code): 5.3k files

conda/base/context.py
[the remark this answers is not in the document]
jaimergp: Its only user in `context.py` has been deprecated, so we might just as well move it somewhere else where it makes more sense. I'll take a look.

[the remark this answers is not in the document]
jaimergp: Done.

jaimergp:
pre-commit.ci autofix

conda/cli/install.py
```diff
@@ -337,7 +333,7 @@ def install_clone(args, parser):
 
 
 def install(args, parser, command="install"):
-    """Logic for `conda install`, `conda update`, and `conda create`."""
+    """Logic for `conda install`, `conda update`, `conda remove`, and `conda create`."""
```
jaimergp, line 336:
```suggestion
    """Logic for `conda install`, `conda update`, and `conda create`."""
```

travishathaway (approved):

PR #14660, opened 2025-03-11
Silence the error when unset variable in fish

zeyugao:
### Description

Fix https://github.com/conda/conda/issues/14659, avoid failure when unset a variable in fish

### Checklist - did you ...

~- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~
~- [ ] Add / update necessary tests?~
~- [ ] Add / update outdated documentation?~

PR #14752, opened 2025-04-11
fix: preserve conda exit status in fish wrapper function

holgerdell:
Fixes an issue where the conda fish wrapper function was incorrectly returning
exit status 4 after successful conda operations. This occurred because the
`eval ($CONDA_EXE shell.fish reactivate)` command fails with status 4 when
attempting to unset non-existent variables with `set -e`.

The fix captures the original conda command's exit status before reactivation
and explicitly returns it, ensuring the function returns the status of the
actual conda operation rather than the status of the environment reactivation.

This maintains the intended behavior where reactivation only occurs for
successful commands while properly preserving error codes for error handling.

### Description

I noticed that the command `conda update --all` returns an incorrect exit code of 4 after using `conda activate` in the `fish` shell. The reason is that the fish function `conda` returns the exit code of `eval ($CONDA_EXE shell.fish reactivate)`, which in turn is 4 if the variable undefined and then unset by `set -e _CE_M` in `$CONDA_EXE shell.fish reactivate`. This pull request fixes this issue.

### Checklist - did you ...


dbast:
@conda-bot check

conda/shell/etc/fish/conf.d/conda.fish
```diff
@@ -72,7 +72,11 @@ function conda --inherit-variable CONDA_EXE
                 eval ($CONDA_EXE shell.fish $cmd $argv)
             case install update upgrade remove uninstall
                 $CONDA_EXE $cmd $argv
-                and eval ($CONDA_EXE shell.fish reactivate)
+                set -l cmd_status $status
+                if test $cmd_status -eq 0
+                    eval ($CONDA_EXE shell.fish reactivate)
+                end
+                return $cmd_status
```
kenodegard, line 79: I think we still want the return code of reactivate if its run, so either of the following:

```suggestion
                set -l cmd_status $status
                if test $cmd_status -eq 0
                    eval ($CONDA_EXE shell.fish reactivate)
                    return $status
                end
                return $cmd_status
```

```suggestion
                set -l cmd_status $status
                if test $cmd_status -ne 0
                    return $cmd_status
                end
                eval ($CONDA_EXE shell.fish reactivate)
```

kenodegard (changes requested):

conda/shell/etc/fish/conf.d/conda.fish
holgerdell: Your suggested changes are functionally equivalent to the one-line version before my suggested commit. So are you saying it is intended behavior that `conda update` fails with an exit code of 4 in the fish shell when e.g. the environment variable _CE_M is not set? I don’t think so. A correct, real alternative is if instead the fish code returned by `reactivate` is changed so that this situation does not cause an error. Would this be preferred?

holgerdell:
@kenodegard I changed the pull request: Instead of editing `conda.fish`, I am proposing an edit to the `set -e` command that causes the exit status of 4. I believe this issue only occurs on recent versions of `fish`, I tested it on version `4.0.2`.

holgerdell:
This is a duplicate of #14660.

scott-5:
The problem is solved and the pr can be closed. @holgerdell

PR #14754, opened 2025-04-12
Add pre/post transaction plugin hooks

### LINKED ISSUE #13795: Plugin system: transaction action plugins

peytondmurray:
### Description

This PR adds a pre/post-transaction hooks plugin system. Pre- (post-) transaction hooks accept a subclass of `Action`, and are guaranteed to run before (after) every other action is run. Plugin authors must define the following methods:

1. ``verify``
2. ``execute``
3. ``reverse`` (only if ``execute`` raises an exception)
4. ``cleanup``

and have access to the following attributes:

- `transaction_context`
- `target_prefix`
- `unlink_precs`
- `link_precs`
- `remove_specs`
- `update_specs`
- `neutered_specs`

In this way, you can define plugins that have access to e.g. what links are being generated/removed and where. For tests, I wrote a few that followed pretty closely the ones I found in that same plugin test directory.

Closes #13795.

Related: https://github.com/conda/conda/pull/13842

### Checklist - did you ...


jaimergp:
Clever! This is cleaner and might actually work. Pinging @jezdez too.

My two cents:
- Instead of adding a new verb, `run()`, with similar semantics, why not rename the existing `execute()` as `def _execute()`, and then turn `execute(): _execute(); post_transaction_plugins()`?
- Is it time to make `_Action()` public?
- This might incur in a LOT of extra calls for IO-heavy transactions. I think some of those `_Action()` subclasses are created per file copy even. I worry about the potential overheads here. I'd like to see numbers / tests where something big is installed and we add a random 1-100ms sleep in the plugin.

conda/plugins/types.py
```diff
@@ -347,3 +348,20 @@ class CondaRequestHeader:
...
+    For details on how this is used, see
+    :meth:`~conda.plugins.hookspec.CondaSpecs.conda_post_transactions`.
+
+    :param name: Post transaction name (this is just a label).
+    :param run: Function to run after each transaction; accepts a single parameter which
+        is the action that has just been executed
+    """
+
+    name: str
+    run: Callable[[_Action], None]
```
jaimergp, line 367: Should the plugin declare on which `_Action` instances it should run? This would allow us to skip a lot of calls. Otherwise, the plugin itself needs to check the `_Action` type every time, possibly just to ignore it. If it should run for everything, then it could just `_Action`. In other words:

```suggestion
    run: Callable[[_Action], None]
    action_type: _Action = _Action
```

conda/plugins/manager.py
```diff
@@ -366,6 +373,14 @@ def get_subcommands(self) -> dict[str, CondaSubcommand]:
             for subcommand in self.get_hook_results("subcommands")
         }
 
+    def run_post_transaction_hooks(self, action: _Action) -> None:
+        """Run the post transaction hooks for the given action.
+
+        :param action: Action for which post transaction hooks are to be run.
+        """
+        for hook in self.get_hook_results("post_transactions"):
+            hook.run(action)
```
jaimergp, line 382:
```suggestion
            if isinstance(action, hook.action_type):
                hook.run(action)
```

I don't think this would help much, though... We are just skipping one call.

jezdez (commented):
Agree with Jaime, let’s not rename the execute functions like this and let’s publish `Action` for sure. A few more comments in-line.

tests/plugins/test_post_transactions.py
```diff
@@ -0,0 +1,72 @@
...
+
+from typing import TYPE_CHECKING
+from unittest import mock
+
+import pytest
+
+from conda import plugins
+from conda.core.path_actions import _Action
+
+# Attempt to load libmamba solver plugin since it's much faster;
```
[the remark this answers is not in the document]
peytondmurray: Understandable; it's really that the classic solver is quite slow and thus these tests are quite slow. Maybe that's an indication that the solve should be mocked out?

conda/plugins/manager.py
peytondmurray: It'll help if the hook is configured to be run only on very specific actions, at least :shrug:

peytondmurray:
> This might incur in a LOT of extra calls for IO-heavy transactions. I think some of those _Action() subclasses are created per file copy even. I worry about the potential overheads here. I'd like to see numbers / tests where something big is installed and we add a random 1-100ms sleep in the plugin.

Okay, some benchmarks on my local machine, using the tests in `tests/plugins/test_post_transactions.py`. These two tests are very similar, but `test_post_transaction_invoked` uses a dummy plugin that calls a function that does nothing, while `test_post_transaction_overhead` uses a plugin that calls a function which adds a random overhead between 1-100ms. The test time is utterly dominated by the length of time it takes to solve the environment:

```
74.41s call     tests/plugins/test_post_transactions.py::test_post_transaction_invoked
37.40s call     tests/plugins/test_post_transactions.py::test_post_transaction_overhead
31.04s call     tests/plugins/test_post_transactions.py::test_post_transaction_raises_exception
```

Let me see if there's a straightforward way to mock the solve, but keep the `UnlinkLinkTransaction` where these hooks are being called from.

conda/core/path_actions.py
```diff
@@ -82,20 +82,29 @@
...
+        # at the end of a verification run, all errors will be raised as a CondaMultiError
         # after successful verification, the verify method should set self._verified = True
         raise NotImplementedError()
 
     @abstractmethod
     def execute(self):
+        """Execute the action."""
         raise NotImplementedError()
 
+    def _execute(self):
```
kenodegard, line 102: What about making this a `__call__` instead?

peytondmurray:
Okay, measuring execution times it looks like if the hook is called on every action, i.e. `action_type=Action`:

[no overhead] elapsed time: 3.9611058235168457
[random overhead] elapsed time: 402.5158312320709

But if for example the hooks use `action_type=CreatePrefixRecordAction`, the dummy plugin is only ~1s faster than the random overhead plugin on my local machine:

[no overhead] elapsed time: 3.776529550552368
[random overhead] elapsed time: 4.754676103591919

So it really depends on what action type is being targeted.

conda/core/path_actions.py
peytondmurray: My only objection here is that the diff would get expanded. Really I'd defer to the community on this one - if there's consensus here I'd be happy to change it.

conda/core/path_actions.py
```diff
@@ -82,20 +82,29 @@
 )
 
 
-class _Action(metaclass=ABCMeta):
+class Action(metaclass=ABCMeta):
```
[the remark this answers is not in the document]
peytondmurray: Hmm, `@deprecated` doesn't seem to play nice with class definitions, so I'm instead decorating the `__init__` method:

```python
class _Action(Action):
    @deprecated("25.9", "26.3", addendum="Use `Action` instead.")
```

As it stands if you decorate the class itself you get an unhelpful error if you try to inherit from it. Here's an example which is already in `main`:

```python
In [1]: from conda.common.io import Spinner

In [2]: class Bar(Spinner):
   ...:     pass
   ...: 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 class Bar(Spinner):
      2     pass

TypeError: function() argument 'code' must be code, not str
```

and I can't get the deprecation message to appear when using the Spinner anyway, so I think for now just deprecating `_Action.__init__` will be sufficient.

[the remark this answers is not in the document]
kenodegard: For renames I've found the `deprecated.constant` to be sufficient, something like:

```python
deprecated.constant(
    "25.9",
    "26.3",
    "_Action",
    Action,
    addendum="Use `conda.core.path_actions.Action` instead.",
)
```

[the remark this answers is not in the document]
kenodegard: And interesting find w.r.t. the `Spinner` deprecation, will need to look into that

jaimergp:
> Let me see if there's a straightforward way to mock the solve

You can use an `@EXPLICIT` lockfile to do that via `conda create --file ...`. There are some at https://github.com/conda/conda-libmamba-solver/tree/main/tests/data, but you can essentially place a few URLs in there and that's all.

The 400s one for `action_type=Action` is a bit concerning, I wonder if we should add different types of hooks:

- post_transaction_hook: would be limited to the `Action` subclasses that only run a few times. This would be recommended.
- post_transaction_slow_hook: would allow all the others.

Do you think that's a good idea? (Names can be different) Alternatively, really loud warnings in the documentation would help. Do we actually have use cases for the `Action` subclasses that are called VERY often?

peytondmurray:
Alternatively, we could _prevent_ users from passing `Action` as a trigger... they'd then be forced to pass only the subclasses they want instead?

jaimergp:
I like that! I think we can easily achieve that with `issubclass(X, Action) and X != Action`?

conda/core/path_actions.py
jaimergp: I don't think we want to expose a private method via a public `instance()` interface. If anything, the public `execute()` should be the one getting `__call__`, but I like how the current code is.

peytondmurray:
Yep, thinking the same thing :handshake:

conda/core/path_actions.py
kenodegard: > My only objection here is that the diff would get expanded.

> I don't think we want to expose a private method via a public `instance()` interface

I'm confused, I'm probably misunderstanding something here.

As I understand this we need subclasses to implement their individual `execute` functions but we implemented this new `_execute` wrapper to properly call the post transaction hook. AFAIK the `execute` is currently invoked in only one place, in `conda.core.link.UnlinkLinkTransaction`.

What I'm suggesting is that we should not (1) be calling a private method outside of the class itself and (2) have nearly identical methods where it is not clear which should be used in what scenario (should `Action.execute` ever be called? or should it only ever be `Action._execute`?).

By switching to `__call__` we resolve the first concern and hide the second concern:

```python
# conda.core.path_actions
class Action:
    def __call__(self):
        result = self.execute()
        context.plugin_manager.run_post_transaction_hooks(self)
        return result

# conda.core.link
class UnlinkLinkTransaction:
    def _execute_actions(axngroup):
        ...
        for action in axngroup.actions:
            action()
        ...
```

jaimergp: So what you are proposing is, given the current state of the PR:

- `execute` becomes `__call__`
- `_execute` becomes `execute`

That could work and I don't mind either. I'll let Peyton choose!

peytondmurray:
Cool, so the way to do this is to use `__post_init__` on the `CondaPostTransaction` class. That way, we only need to check once when the plugin is instantiated :+1:.

jjhelmus:
To reduce the overhead what if the `execute` method (or a new `post_execute` method) of each `Action` subclass was patched beforehand to call all applicable plugins? This would require a one time setup costs which evaluated each possible subclass.

Another idea would be to use a dispatch table that when called initially for each subclass determines the necessary plugin routines to run. Subsequent calls would run the cached list of routines for the given class.

conda/core/path_actions.py
peytondmurray: I'm fine with the proposed changes, but wanted to comment:

> What I'm suggesting is that we should not (1) be calling a private method outside of the class itself

To be fair there are not clear lines between what I'd traditionally say are private and public APIs here. For example - `_Action` I would argue is not (or rather _should not_) have been part of the public API, and yet we are going through a deprecation cycle for this :shrug:

peytondmurray:
> To reduce the overhead what if the execute method (or a new post_execute method) of each Action subclass was patched beforehand to call all applicable plugins? This would require a one time setup costs which evaluated each possible subclass.

Ahh, so instead of having the Action call the plugin manager, which loops over all the hooks and passes the Action instance to each hook which matches, the Action instead asks the plugin manager for the list of hooks and iterates over the matching ones itself?

conda/plugins/types.py
```diff
@@ -347,3 +349,31 @@ class CondaRequestHeader:
...
+        be an Action subclass, or a tuple of Action subclasses
+    """
+
+    name: str
+    run: Callable[[Action], None]
+    action_type: type[Action] | tuple[type[Action]]
+
+    def __post_init__(self):
+        """Validate that the action_type is an Action subclass."""
+        if not is_only_subclass(self.action_type, Action):
```
travishathaway, line 375: I don't think it makes sense to be doing this type of post-initialization type enforcement for our type classes. If we began doing it, it should be taken care of in a separate pull request and applied to all applicable type classes.

Let's just keep this pull request as simple as possible.

travishathaway:
@peytondmurray and @jaimergp,

I thought the original spirit of #13795 was to create a plugin hook that would allow you to add new types of actions and not simply something that would be executed after each action run. I'm not sure this pull request addresses that original wish.

I can't help but think what was originally described in #13795 would be a better fit. If we do this, the plugin system will be easier to understand from the perspective of developers outside of conda because they wouldn't have to learn about each action defined and run.

kenodegard:
I think we're missing real world examples of this plugin hook in use. Can we create a PR in `conda-pypi` and/or `conda-checkpoints` using this new hook to help confirm that it does indeed solve our use case?

jaimergp:
`conda-checkpoints` is probably the easiest one.

peytondmurray:
> I think this can be achieved with the code, a new action would be basically a new Action subclass that the plugin package provides and that the hook implementation wires up with the run callback. It does mean that Action has to become a "public" API though, you're right. Does that concern you?

A little more context to add to this: this PR follows the discussion from https://github.com/conda/conda/pull/13842, which originally targeted https://github.com/conda/conda/issues/13795 but shifted to a discussion about post-transaction hooks rather than focusing on injecting new transactions.

travishathaway:
> I think this can be achieved with the code, a new action would be basically a new `Action` subclass that the plugin package provides and that the hook implementation wires up with the `run` callback. It does mean that `Action` has to become a "public" API though, you're right. Does that concern you?

@jezdez, yes, this is the main crux of what concerns me the most about this plugin hook. This part of the code, while better structured and easier to reason about than others, is still poorly documented. If we want to expose these action classes, we need to make sure they are documented well and provide examples that approximate real world use cases.

See my current work on Reporter Backends for an example of what it could look like:

- https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/reporter_backends.html

In that plugin, I actually provide links to two real implementations plugin authors can use as a reference. Ideally, we could do the same thing here even if we need to link to an external plugin.

conda/auxlib/type_coercion.py
```diff
@@ -263,3 +265,23 @@ def typify_data_structure(value, type_hint=None):
...
+        base_class itself). If ``obj_class`` is a tuple, True is returned only if all elements
+        of the tuple are base classes; False otherwise
+    """
+
+    if isinstance(obj_class, tuple):
+        classes = obj_class
+    else:
+        classes = (obj_class,)
+
+    return all(
```
jaimergp, line 285: Isn't this supposed to be `any`?

jaimergp:
> I thought the original spirit of https://github.com/conda/conda/issues/13795 was to create a plugin hook that would allow you to add new types of actions and not simply something that would be executed after each action run. I'm not sure this pull request addresses that original wish.

I'm thinking about this comment and actually there's a good point. If a user doesn't really pick the right `Action` to hook onto, then the same plugin might run several times.

I think a good compromise between the original intent (where we were trying to add new Action classes) and this smarter approach is (where we hook on to existing Actions), is to do the following:

- We add a few blank Actions at the right levels of abstraction. We can call them `[AbstractionLevel]PluginHookAction`. I think we need to add as many as mentioned in this comment: https://github.com/conda/conda/pull/13842#issuecomment-2074805623
- We add four plugin hooks like the one proposed in this PR, with the ability to hook on to a given `*PluginHookAction` (and only those!).
  - `pre_execute`
  - `post_execute` (this is the one we have in this PR)
  - `pre_reverse`
  - `post_reverse` 
- We instrument `UnlinkLinkTransaction` so the relevant `*PluginHookAction` instances are added to the transaction, depending on whether there are active plugin hooks there or not.

Does that sound like too much? It's a compromise between the initially proposed strategy, which involved a LOT of refactoring, and the feedback received for this one.

jaimergp:
For reference, this is the class three in `_Action` (generated from this [SO answer](https://stackoverflow.com/a/54614686)):

```
_Action
└---PathAction
    └---PrefixPathAction
        └---CreateInPrefixPathAction
            └---LinkPathAction
                └---PrefixReplaceLinkAction
            └---MakeMenuAction
            └---CreatePythonEntryPointAction
            └---CreatePrefixRecordAction
            └---UpdateHistoryAction
        └---RemoveFromPrefixPathAction
            └---UnlinkPathAction
                └---RemoveLinkedPackageRecordAction
            └---RemoveMenuAction
    └---RegisterEnvironmentLocationAction
    └---UnregisterEnvironmentLocationAction
    └---CacheUrlAction
    └---ExtractPackageAction
└---MultiPathAction
    └---CompileMultiPycAction
        └---AggregateCompileMultiPycAction
```

travishathaway:
@jaimergp,

I like the approach of adding placeholders during parts of the transaction. But, wouldn't it be better for the plugin hooks to correspond more closely to the abstract methods defined on this class?

- https://github.com/conda/conda/blob/5607154ddd418bcb39cd8ea76aaa44e46d798b95/conda/core/path_actions.py#L88-L109

That would be:
- verify
- execute
- reverse
- cleanup
- verified

If these new "plugin actions" are being defined separately, I'm not sure there's a need to use the idea of "pre" and "post".

Another thing I'm considering is the order in which the registered plugin actions are run. Or we going to have to implement some sort of weird "z-index" (e.g. from CSS element ordering) integer field here to establish which plugins should be run before others?

jaimergp:
> But, wouldn't it be better for the plugin hooks to correspond more closely to the abstract methods defined on this class?

Yea, actually makes sense. If there's nothing to wrap, we don't have to have pre- and post-. I'll update my comment.

> Another thing I'm considering is the order in which the registered plugin actions are run.

Yep, that's tricky. For now I think we should just say "order of execution cannot be guaranteed, assume arbitrary and architect your plugin accordingly" in the docs.

peytondmurray:
Moving back into draft pending completion. I'm still unsure about the proposed implementation - let me restate it here since it's not immediately apparent to me based on the conversation above, and so that I can make sure we're all on the same page:

1. Implement a `PluginHookAction` which unconditionally gets added to the `ActionGroup` instances that are created by the `UnlinkLinkTransaction`.
2. The `PluginHookAction` doesn't do anything except call plugin hooks on the following functions:

- verify
- execute
- reverse
- cleanup
- verified

---

Do people want separate classes for the different stages at which these plugin hooks get executed? I.e. should we have `CondaPostVerifyTransaction`, `CondaPostExecuteTransaction`, ...? Alternatively they could all be one plugin hook class, where the user specifies at what stage they want the plugin run:

```python
CondaPostTransaction(
    name="my_transaction",
    run=my_hook_function,
    stage="verify",
)
```

or similar.

---

Another implication of this implementation strategy is that the plugin hook functions loose any information about what the `UnlinkLinkTransaction` is doing. With the current implementation where hooks are run on each action, the hook function gets passed the `Action` instance itself. Plugin authors can examine the attributes of the action to figure out what it is doing e.g. linking to a specific directory, removing links, etc. The proposed method loses that information, because we are just adding a new dummy action to the `UnlinkLinkTransaction` action groups; they don't get passed any information about their sibling actions.

conda/auxlib/type_coercion.py
```diff
@@ -263,3 +265,23 @@ def typify_data_structure(value, type_hint=None):
 
 def maybecall(value):
     return value() if callable(value) else value
+
+
+def is_only_subclass(obj_class: Union[Any, tuple[Any]], base_class: type) -> bool:
```
kenodegard, line 270: Let's add `from __future__ import annotations` at the top:

```suggestion
def is_only_subclass(obj_class: type | tuple[type, ...], base_class: type) -> bool:
```

kenodegard, line 270: I think "strict" is a better term here.

```suggestion
def is_strict_subclass(obj_class: Union[Any, tuple[Any]], base_class: type) -> bool:
```

conda/auxlib/type_coercion.py
kenodegard: No. The check is asserting that all of the classes are strict subclasses.

conda/auxlib/type_coercion.py
```diff
@@ -263,3 +265,23 @@ def typify_data_structure(value, type_hint=None):
...
+    """
+
+    if isinstance(obj_class, tuple):
+        classes = obj_class
+    else:
+        classes = (obj_class,)
+
+    return all(
+        issubclass(item, base_class) and item != base_class for item in classes
+    )
```
kenodegard, line 287: `issubclass` is fragile to objects being passed in, let's catch those errors, e.g.:

```pycon
>>> issubclass(12, int)
TypeError: issubclass() arg 1 must be a class
```

```suggestion
    try:
        return all(
            issubclass(item, base_class) and item != base_class for item in classes
        )
    except TypeError:
        # TypeError: item is not a class
        return False
```

tests/plugins/test_post_transactions.py
kenodegard, line 1: We prefer to use the `mocker` fixture provided by `pytest_mock` instead of `unittest.mock`

tests/plugins/test_post_transactions.py
```diff
@@ -0,0 +1,175 @@
...
+
+@pytest.fixture()
+def post_transaction_plugin_overhead(plugin_manager_with_reporter_backends):
+    # Explicitly load the solver, since this is a dummy plugin manager and not the default
+    plugin_manager_with_reporter_backends.load_plugins(plugins.solvers)
+
+    plugin = OverheadActionPlugin()
+    with mock.patch.object(
+        plugin, "overhead_hook", wraps=plugin.overhead_hook
+    ) as mock_hook:
```
kenodegard, line 86: I believe this is the same as using `mocker.spy`, see https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

kenodegard:
Oops, I see now y'all moved this back to draft, feel free to ignore my comments

conda/core/link.py
```diff
@@ -926,6 +947,15 @@ def _execute(self, all_action_groups):
                         #   call something that isn't there anymore
                         for axngroup in make_menu_actions:
                             UnlinkLinkTransaction._execute_actions(axngroup)
+
+                # Execute any user-defined post-transaction actions
+                for exc in self.execute_executor.map(
```
peytondmurray, line 952: Probably not necessary to pass to the executor, since there will typically be very few FinalTransactionActions, but if many are implemented this may be useful to have here.

peytondmurray:
Since plugin developers will subclass `FinalTransactionAction`, there's no longer a need to deprecate `_Action`. Thoughts?

peytondmurray:
Failures look like a couple of flaky requests.

conda/core/path_actions.py
```diff
@@ -82,20 +83,27 @@
 )
 
 
-class _Action(metaclass=ABCMeta):
+class Action:
```
travishathaway, line 86: Wouldn't it be easier to make this the base Abstract class? If we did that, we wouldn't have to duplicate the abstract method definitions in `FinalTransactionAction`.

conda/plugins/hookspec.py
```diff
@@ -256,6 +257,45 @@ def conda_health_checks():
...
+                def execute(self):
+                    print(
+                        self.transaction_context,
+                        self.target_prefix,
+                        self.unlink_precs,
+                        self.link_precs,
+                        self.remove_specs,
+                        self.update_specs,
+                        self.neutered_specs,
+                    )
```
travishathaway, line 286:
```suggestion
                    )
                def verify(self):
                    print("Performing verify actions...")
                    self._verified = True

                def reverse(self):
                    print("Reversing...")

                def cleanup(self):
                    print("Cleaning up...")
```

It would be nice for the example to define all abstract methods.

docs/source/dev-guide/plugins/post_transactions.rst
```diff
@@ -0,0 +1,15 @@
+=================
+Post-transactions
+=================
+
+:ref:`Solver transactions <solver_api_transactions>` can be extended with the
+``conda_post_transactions`` plugin hook. This plugin hook accepts a function which will be called
+after any action is executed, with the action object as the only argument. Arbitrary code can be
+executed via this mechanism, but remember that even simple solves can generate thousands of actions
+- so running computationally expensive code via this plugin mechanism can be slow.
```
travishathaway, line 9:
```suggestion
:ref:`Solver transactions <solver_api_transactions>` can be extended with the
``conda_post_transactions`` plugin hook. This plugin hook accepts a subclass of ``FinalTransactionAction``
which will be instantiated and appended to the end of conda's transaction action list. When defining
the class you must define the following methods:

- ``execute``: this is the primary place to put code you wish to run during the transaction.
- ``verify``: this is run before the transaction is executed. This is a good place to check for
  conditions that would cause the transaction to fail.
- ``cleanup``: this is run after the transaction is executed. This is a good place to clean up any
  resources that were created during the transaction.
- ``reverse``: in the case of a failure, this allows you to define any reversal procedures.

```

jaimergp:
We're getting closer! I have some comments / asks:

- Can you update the top-level description of the PR? We've gone through several iterations and designs, and I don't think the current description reflects the latest changes.
- Should we also add pre-transaction hooks using the same design we've agreed on here? As a potential use case: review packages about to be installed or uninstalled (e.g. an institution would forbid a certain license from being used, and this can be implemented via a plugin). Not sure if they are super useful because they are very similar to `post_solve`, _but_ lockfiles processing won't hit the solver, so I think we do need a pre-transaction hook for those cases.

jaimergp:
> Since plugin developers will subclass `FinalTransactionAction`, there's no longer a need to deprecate `_Action`. Thoughts?

Yea maybe we can revert this because it's no longer needed as a public interface.

peytondmurray:
> Yea maybe we can revert this because it's no longer needed as a public interface.

Actually, in response to a comment above, we are not defining a `FinalTransactionAction` and instead just using the `Action` base class.

conda/core/path_actions.py
```diff
@@ -82,27 +85,84 @@
...
 
+    def __init__(
+        self,
+        transaction_context,
+        target_prefix,
+        unlink_precs,
+        link_precs,
+        remove_specs,
+        update_specs,
+        neutered_specs,
```
travishathaway, line 121:
```suggestion
        transaction_context: dict[str, str] | None = None,
        target_prefix: str | None = None,
        unlink_precs: Iterable[PackageRecord] | None = None,
        link_precs: Iterable[PackageRecord] | None = None,
        remove_specs: Iterable[MatchSpec] | None = None,
        update_specs: Iterable[MatchSpec] | None = None,
        neutered_specs: Iterable[MatchSpec] | None = None,
```

Let's add some nice type annotations because this will be a public class. I also think that each argument should be optional in case we want to change what we pass these objects in the future.

peytondmurray: Thanks for this - IIRC the calling functions don't have type annotations for these, but it's nice to have the insider knowledge for this :)

conda/core/link.py
```diff
@@ -193,6 +194,7 @@ class PrefixActionGroup(NamedTuple):
     make_menu_action_groups: Iterable[ActionGroup]
     entry_point_action_groups: Iterable[ActionGroup]
     prefix_record_groups: Iterable[ActionGroup]
+    final_transaction_groups: Iterable[ActionGroup]
```
jaimergp, line 197: Unfortunately, namedtuple args are positional, so this is an API change that needs to go through the deprecation cycle.

What if we create a new `PrefixActionGroup` namedtuple with the new arguments, and deprecate the old one? e.g. `PrefixActionGroupV2` or similar? Even better if the new object is a dataclass so it doesn't suffer from this problem in the future?

news/13795-transaction-hooks
```diff
@@ -0,0 +1,19 @@
...
+
+* Add the ability to define pre- and post-transaction hooks. (#14754)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
```
jaimergp, line 11: Needs entries for the deprecations. See example syntax at https://github.com/conda/conda/blob/1ec58ac92abf05907786c329c65989d79ced6c2e/news/14750-prefixdata#L12-L22

conda/plugins/hookspec.py
```diff
@@ -256,6 +258,104 @@ def conda_health_checks():
         """
         yield from ()
 
+    @_hookspec
+    def conda_pre_transactions(self) -> Iterable[CondaPreTransaction]:
```
jaimergp, line 262: Risking being pedantic here, but these are not "transactions", but "actions". I know it's more verbose, but the technically correct term would be `conda_pre_transaction_actions`. Similarly for the other objects and the `post-` variants.

jaimergp (dismissed):
Looking pretty good!

We do need a couple changes:

- Some deprecations (see above)
- Rename `(pre|post)_transactions` to `(pre|post)_transaction_actions`.

conda/core/link.py
peytondmurray: Ack you're right! Wow, the `PrefixActionGroup` feels like such an implementation detail I'm surprised it's not private. Deprecation, and the dataclass was nearly a drop-in replacement :rocket:

peytondmurray:
My one hesitation here is about having `PrefixActionGroupV2` be public. Really it seems like an implementation detail, and we're already going through one deprecation cycle for its predecessor. Is there actually a reason to make this public?

If not, is there any way I can make it private? Normally I'd prefix the name of it with an underscore, but I'm also having to deprecate `_Action` here as well...

travishathaway:
> My one hesitation here is about having `PrefixActionGroupV2` be public. Really it seems like an implementation detail, and we're already going through one deprecation cycle for its predecessor. Is there actually a reason to make this public?
> 
> If not, is there any way I can make it private? Normally I'd prefix the name of it with an underscore, but we're also deprecating `_Action` here too...

If it makes you feel any better, we also deprecate things with underscores in front of them:

- https://github.com/conda/conda/blob/a6288c22108503574a5ea8108d232a76eb9e8d39/conda/core/index.py#L684-L735

docs/source/dev-guide/deep-dives/install.md
```diff
@@ -559,6 +559,8 @@ Their order is roughly how they happen in practice:
 7. `make_menu_action_groups`, composed of `MakeMenuAction` actions.
 9. `prefix_record_groups`, records installed packages in the environment via
    `CreatePrefixRecordAction` actions.
+10. `initial_action_groups`, includes any user-defined pre-transaction actions.
+11. `final_action_groups`, includes any user-defined post-transaction actions.
```
travishathaway, line 563:
```suggestion
10. `initial_action_groups`, includes any plugin-defined pre-transaction actions.
11. `final_action_groups`, includes any plugin-defined post-transaction actions.
```

Right? "user-defined" is technically true, but I think "plugin-defined" is a little more specific.

travishathaway (dismissed):
I think this is looking good! Great job on even taking the time to update the "deep dives" section.

conda/core/link.py
```diff
@@ -179,10 +181,45 @@ class PrefixSetup(NamedTuple):
...
+    make_menu_action_groups: Iterable[ActionGroup]
+    entry_point_action_groups: Iterable[ActionGroup]
+    prefix_record_groups: Iterable[ActionGroup]
+    initial_action_groups: Iterable[ActionGroup] = ()
+    final_action_groups: Iterable[ActionGroup] = ()
+
+    def __iter__(self) -> Generator[Iterable[ActionGroup], None, None]:
+        for field in fields(self):
+            yield getattr(self, field.name)
+
```
kenodegard, line 220: Can we instead just make the new `dataclass` quack like a `NamedTuple` and just deprecate the tuple parts? Something like:

```
from dataclasses import dataclass, fields, astuple, field
from typing import Iterable, Generator, Any, Iterator, Tuple

def deprecate_PrefixActionGroup_tuple() -> None:
    deprecated.topic(
        "25.9",
        "26.3",
        topic="`conda.core.link.PrefixActionGroup` as a NamedTuple",
        addendum="Utilize `conda.core.link.PrefixActionGroup` as a dataclass instead.",
    )

@dataclass
class PrefixActionGroup:
    remove_menu_action_groups: Iterable[ActionGroup]
    unlink_action_groups: Iterable[ActionGroup]
    unregister_action_groups: Iterable[ActionGroup]
    link_action_groups: Iterable[ActionGroup]
    register_action_groups: Iterable[ActionGroup]
    compile_action_groups: Iterable[ActionGroup]
    make_menu_action_groups: Iterable[ActionGroup]
    entry_point_action_groups: Iterable[ActionGroup]
    prefix_record_groups: Iterable[ActionGroup]
    initial_action_groups: Iterable[ActionGroup] = ()
    final_action_groups: Iterable[ActionGroup] = ()
    
    # quack like a tuple
    def __getitem__(self, index):
        deprecate_PrefixActionGroup_tuple()
        return astuple(self)[index]
    
    def __iter__(self) -> Iterator[Iterable[ActionGroup]]:
        deprecate_PrefixActionGroup_tuple()
        return iter(astuple(self))
    
    def __len__(self) -> int:
        deprecate_PrefixActionGroup_tuple()
        return len(fields(self))
    
    def __iter_fields__(self) -> Generator[Iterable[ActionGroup], None, None]:
        yield from astuple(self)
    
    def __eq__(self, other):
        if isinstance(other, tuple):
            deprecate_PrefixActionGroup_tuple()
            return astuple(self) == other
        return super().__eq__(other)
```

peytondmurray: We definitely can, but is this actually substantially better than what we have? Yes, it does avoid us having a `PrefixActionGroupV2`, but making it quack like a `NamedTuple` doesn't actually make it a `NamedTuple`. So if the primary concern here is about backwards compatibility, `issubclass(PrefixActionGroup, tuple) == False` whereas `issubclass(PrefixActionGroupV2, tuple) == True`. And to my knowledge we are not using `__len__`, `__iter_fields__`, `__eq__`, or `__getitem__`, so I just don't see the need here.

We could probably cook up some multiple inheritance scheme to get this to actually behave like a namedtuple. If other folks are really interested in having this or the suggestion above be the path forward I'm happy to do it, but again for what is effectively an internal only implementation detail I just don't see the point.

kenodegard: Can define `__instancecheck__` and `__subclasscheck__` magic methods to address  `isinstance` and `issubclass`

🤷🏼‍♂️ just trying to think of alternatives to needing to introduce a V2 suffix

conda/core/link.py
```diff
@@ -551,7 +587,35 @@ def _prepare(
...
+                hook.action(
+                    transaction_context,
+                    target_prefix,
+                    unlink_precs,
+                    link_precs,
+                    remove_specs,
+                    update_specs,
+                    neutered_specs,
+                )
+            )
```
kenodegard, line 616: The pattern established by other plugin hooks is to define a helper method in the plugin manager to do this so we don't need to have to call `get_hook_restults` with the magic string outside of `conda.plugins.manager`:

```suggestion
        pre_transaction_actions = context.plugin_manager.get_pre_transaction_actions(
            transaction_context,
            target_prefix,
            unlink_precs,
            link_precs,
            remove_specs,
            update_specs,
            neutered_specs,
        )
        post_transaction_actions = context.plugin_manager.get_post_transaction_actions(
            transaction_context,
            target_prefix,
            unlink_precs,
            link_precs,
            remove_specs,
            update_specs,
            neutered_specs,
        )
```

conda/core/path_actions.py
```diff
@@ -82,27 +89,84 @@
...
+    def execute(self) -> None:
+        """Execute the action.
+
+        Called after ``self.verify()``. If this function raises an exception,
+        ``self.reverse()`` will be called.
+        """
+
+    def __call__(self) -> None:
+        """Execute the action."""
+        return self.execute()
```
kenodegard, line 156: I don't think this serves a purpose anymore, correct? If so we should revert this and just call `action.execute` as before.

kenodegard (dismissed):

conda/core/link.py
jaimergp: Can we instead do something like this?

- Introduce PrefixActionGroupDataclass, already deprecated and scheduled to be renamed as simply PrefixActionGroup
- Deprecate the old PrefixActionGroup, which will be replaced by PrefixActionGroupDataclass when time comes
- We start using PrefixActionGroupDataclass internally

jaimergp: Although this is a bit like mutually canceling deprecations; downstream users won't have the chance to change their code in time 😬 I'd rather stick to well-known APIs instead of trying to each objects how to be have like others. Even if we make it quack like a tuple, the amount and order of elements is different, so it's still an API breakage.

I say we keep the solution as is. If we don't like the V2 suffix, I think `Dataclass` is descriptive enough. Other alternatives:

- `PrefixActions`
- `GroupedPrefixActions`
- `PrefixActionGrouper`

jezdez (changes requested):
Let's fix those names and then merge this thing. This is awesome work @peytondmurray!

peytondmurray:
Okay, here's where we stand:

- `PrefixActionGroup` is deprecated, with suggestion to use `PrefixActionGroupDataclass` instead
- `PrefixActionGroupV2` renamed to `PrefixActionGroupDataclass`
- `PrefixActionGroupDataclass` deprecated, to be renamed in 26.3 to `PrefixActionGroup`, following @jaimergp's suggestion

Thanks folks for the support!

peytondmurray:
100% agreed :no_good: :coffee: :no_good:

jezdez (approved):

peytondmurray:
> @peytondmurray We need to pay much closer attention to detail here. The PrefixAction class is marked as deprecated and points to PrefixActionGroup, but below the PrefixActionGroup is marked as deprecated with a message that says Use PrefixActions instead.

I wouldn't call this an inattention to detail - you'll notice that the original implementation here included a `PrefixActionGroupV2` which didn't have any deprecation. After the discussion with 4 code owners above, group consensus was to move forward with the cyclic deprecation, and rename this object now. I understand your objection: this results in a deprecation that in some sense users are unable to solve, but in practice these are `PendingDeprecationWarning` which many users will never see, as the default warning filter ignores it. If we don't want to be doing this I'm happy to revert that change, and we should escalate warnings into errors across the test suite (as in https://github.com/conda/conda/pull/12771).

> I think the intent is to rename one to the other. However, this doesn't really work since the deprecation warnings are circular and so downstream users cannot switch from one class to the other to remove the warnings.
>
> If the two offer identical functionality and are drop-in replacements, a workable approach is to instead simply do the replacement now and leave PrefixActions as a pointer to PrefixActionGroup with a deprecation warning on PrefixActions only.
>
> If that won't work, then you'll need to go through two deprecation cycles or pick a new unique name or something.

As discussed above `PrefixActions` is not a drop in replacement for `PrefixActionGroup`, so we'll either need to live with a new name for this or go through two deprecation cycles. Looks like this is already being handled in https://github.com/conda/conda/pull/12771?

PR #14755, opened 2025-04-12
Fix PowerShell hook error on startup with restricted execution policy

badihian:
### Description

Fixes a bug in the PowerShell activation logic where Conda initialization fails if the user's execution policy is set to `Restricted`. This patch modifies `_powershell_profile_content` to:

- Check if the execution policy is not `Restricted`
- Wrap the hook invocation in a `try/catch` block
- Fallback silently with a `Write-Verbose` message if policy prevents execution

This improves compatibility with default-restricted Windows environments and prevents errors on shell startup.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes? → *Added below*
- [x] Add / update necessary tests? → *Patch is runtime-guarded and non-breaking, doesn't affect core logic*
- [x] Add / update outdated documentation? → *N/A — internal shell hook*

---

#### News entry suggestion (add to `news/` directory):

**fixes**

Fix PowerShell hook error on startup when execution policy is restricted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,8 @@ rever/
 
 # setuptools-scm
 conda/_version.py
+
+# Development/testing files
+hook_test.ipynb
+test_profile.ps1
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -45,8 +45,3 @@ rever/
 
 # setuptools-scm
 conda/_version.py
-
-# Development/testing files
-hook_test.ipynb
-test_profile.ps1
-.vscode/
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1711,8 +1711,12 @@ def _powershell_profile_content(conda_prefix):
         f"""
     #region conda initialize
     # !! Contents within this block are managed by 'conda init' !!
-    If (Test-Path "{conda_exe}") {{
-        (& "{conda_exe}" "shell.powershell" "hook") | Out-String | ?{{$_}} | Invoke-Expression
+    try {{
+        if ((Get-ExecutionPolicy -Scope CurrentUser) -ne 'Restricted') {{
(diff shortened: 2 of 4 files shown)


badihian:
@conda-bot I have signed the Contributor License Agreement

dbast:
@conda-bot check

.gitignore
```diff
@@ -45,3 +45,8 @@ rever/
 
 # setuptools-scm
 conda/_version.py
+
+# Development/testing files
+hook_test.ipynb
+test_profile.ps1
+.vscode/
```
travishathaway, line 52:
```suggestion
```

Please remove all changes to `.gitignore`.

badihian: I removed the changes, let me know if that looks good now.

badihian:
Hi @travishathaway,

Thank you for your guidance. I have pushed a testing file. Hopefully that works, but if you would rather I remove the file so that you might simply recreate it yourself on your local machine, that would be understandable. Let me know how you would like to proceed.

travishathaway:
Hi @badihian,

We appreciate the help. I would like to ask that you remove the testing file because we need to run a full integration test to see what's happening for Power Shell users.

We'll be able to do that by simply updating our Power Shell profiles by running `conda init` within a development environment.

To test locally, we'll need to do the following:

1. Have "restricted" mode set and attempt to run conda on Power Shell start-up (we'll want to observe the original error, then apply your suggested changes to ensure it works as expected)
2. Have "restricted" mode disabled and attempt to run conda Power Shell start-up (we'll want to make sure that the new code doesn't have any unintended consequences)

If you have any ideas how this could be automated in GHA Windows runners, let us know! I'm not sure how this "restricted" mode could be dynamically enabled and disabled there though, and automatically testing things like shell start-up still remains quite difficult for us to do.

Issue #14759, opened 2025-04-15
Errors seen in GHA test runners on main branch


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Every time a test run fails on "main" a new issue is created (see: https://github.com/conda/conda/issues?q=is%3Aissue%20state%3Aopen%20Tests%20failed%20%20author%3Aconda-bot)

This issue is an attempt to aggregate what's been failing so we can at least begin recognizing some patterns (e.g. which tests have been failing the most).

As we add the test failures to this issue, we will close the ones automatically opened by conda-bot.

### Conda Info

```shell
n/a
```

### Conda Config

```shell
n/a
```

### Conda list

```shell

```

kenodegard:
The `tests/shell/test_csh.py::test_basic_integration[tcsh]` failure was identified and resolved as part of the `25.3.x` releases, see https://github.com/conda/conda/issues/14694 & https://github.com/conda/conda/pull/14707

jaimergp:
See https://github.com/conda/conda/pull/14764 for the red ones.

kenodegard:
Closing in favor of the new slow test epic: https://github.com/conda/conda/issues/15637

PR #14764, opened 2025-04-17
Run known flaky tests in subprocess

jaimergp:
### Description

A potential solution for 🔴 tests in https://github.com/conda/conda/issues/14759

Ideally we would use `pytest-rerunfailures` + `pytest-xdist` (see ["Recover from hard crashes"](https://github.com/pytest-dev/pytest-rerunfailures#recover-from-hard-crashes)), but I didn't find a clean way to do that for only a few tests (we'd need to run _everything_ under xdist). Also [we'd miss the stdout/stderr output](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers). Instead we just need to check for a sentinel env var.

### Checklist - did you ...

### DIFF
--- a/conda/testing/helpers.py
+++ b/conda/testing/helpers.py
@@ -742,7 +742,7 @@ def solver_class():
 
 
 def forward_to_subprocess(
-    request_node, *cli_args, **subprocess_kwargs
+    request_node, reruns: int = 0, *cli_args, **subprocess_kwargs
 ) -> subprocess.CompletedProcess | None:
     if not os.getenv("RERUN_IN_SUBPROCESS"):
         args = cli_args or (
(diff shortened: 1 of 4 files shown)


jaimergp:
[In this run](https://github.com/conda/conda/actions/runs/14512855891/job/40715311627?pr=14764), you can see how the subprocess shielded from the segfault. Now we can add subprocess-based re-runs to workaround the flakiness.

jaimergp:
Reruns in action:

- [`test_solve_1`](https://github.com/conda/conda/actions/runs/14517653798/job/40730544747?pr=14764#step:12:382)

PR #14766, opened 2025-04-19
Implement `conda-meta/frozen` protections (CEP 22)

### LINKED ISSUE #14746: Implement CEP 22 support for frozen environments
### Checklist

### What is the idea?

Teach `conda` how to interpret `conda-meta/frozen` markers.

### Why is this needed?

Community standard that would allow installers to protect `base`, among others.

### What should happen?

Discussed in [CEP 22](https://github.com/conda/ceps/blob/main/cep-0022.md). 

### Additional Context

_No response_

jaimergp:
### Description

Closes #14746 

### Checklist - did you ...


jezdez (dismissed):
This is really nicely scoped, let's do this for the May release?

conda/cli/helpers.py
```diff
@@ -205,6 +205,15 @@ def add_output_and_prompt_options(p: ArgumentParser) -> _ArgumentGroup:
     return output_and_prompt_options
 
 
+def add_parser_frozen_env(p: ArgumentParser):
+    p.add_argument(
+        "--ignore-frozen-env",
```
kenodegard, line 210: This sounds like we are ignoring the env itself, not the protection:

```suggestion
        "--ignore-frozen-env-protection",
```

```suggestion
        "--override-frozen-env",
```

```suggestion
        "--override-frozen",
```

conda/core/prefix_data.py
```diff
@@ -259,6 +273,22 @@ def assert_writable(self) -> None:
         if not file_path_is_writable(self._magic_file):
             raise EnvironmentNotWritableError(self.prefix_path)
 
+    def assert_not_frozen(self) -> None:
```
kenodegard, line 276: Since all of the calls are:

```python
if context.protect_frozen_envs:
    prefix_data.assert_not_frozen()
```

Can we combine it:

```suggestion
    def assert_not_frozen(self, override: bool = False) -> None:
        if override:
            return
```

conda/cli/helpers.py
kenodegard: After reading through everything I think I prefer `--override-frozen`

conda/core/prefix_data.py
jaimergp: It feels weird to have a dedicated boolean to turn this function into a no-op. The override flag is an application layer detail so I'd rather have those uncoupled and I don't think it's that bad to have a couple of `if`s in the CLI modules.

conda/cli/helpers.py
jaimergp: I was going at it as in `--ignore-env-is-frozen`. Is that better?

We also have `--freeze-installed-packages`, so that's why I feel it's better to mention `env` explicitly in the flag name, for clarity.

jaimergp: > We could use --break-base-environment as an alternative name

We can freeze any environment, not just base, so maybe we can use `--break-environment`?

kenodegard: `--override-frozen`/`--override-frozen-env` makes more sense to me and wont require reading the docs to understand what is happening.

I prefer we stick to existing conda lingo. I consider `--override-channels` to have a very clear meaning in `conda` (or do I use that more than the average user?) so using the same terminology here makes it easier to understand. 🤷🏼‍♂️

Whatever is decided I think the argument needs to explicitly reference "frozen", forcing the user to have to learn two terms ("frozen env" and "break env") for a pretty simple feature is overkill.

jezdez (approved):
Let's go with one, if needed we can tweak later with aliases.

PR #14775, opened 2025-04-22
Add `os::cygwin` & `os::msys2` labels

travishathaway:
### Description

I've been noticing more and more Git BASH issues, and I realized that we don't have a way to easily categorize them.

In the past, we used a combination of `cli::posix` and `os::windows`, but this isn't super clear:

https://github.com/conda/conda/issues?q=is%3Aissue%20state%3Aopen%20label%3Acli%3A%3Aposix%20label%3Aos%3A%3Awindows

It would be better to just have a specific label for this particular shell. Below is an example of the type of issue that would make a good candidate for it:

- https://github.com/conda/conda/issues/14758

### Checklist - did you ...


ForgottenProgramme (dismissed):

kenodegard:
I think we want `cli::git-bash`

`cli::windows-bash` has unclear meaning and could refer to msys2/cygwin bash, git-bash, or WSL bash

ifitchet:
Not wanting to muddy the waters but...
_MSYS2_, the project, has an `MSYS` POSIX-aware environment which contains `bash`.  The `MSYS` environment is derived from the POSIX-on-Windows project, Cygwin.  That is available in the `msys2` channel as `m2-bash` (quite old) and on `main` as `msys2-bash` (relatively new).
`git`, on `main`, is a repack of Git for Windows which uses both the `MSYS` and `MINGW64` _MSYS2_ environments and, therefore, its `bash` is largely the same as the above, albeit at at different release.
WSL comes in two forms, WSL1 is the paravirtualised original Linux (Ubuntu until recently) variant and WSL2 is the Hyper-V fully virtualised variant.  I can't see that `bash` would be different between the two WSL variants (modulo speed of accessing files, IIRC) as it's just `bash` in a POSIX environment.
The chances are that these issues are being flagged as a "`bash` on Windows" issue (for which read "Unix shell on WIndows", cf. `csh`, `dash`, `fish`, ...) which will likely be a `/` <-> `\`, drive letter or similar "Windows isn't Unix" issue.  Which specific `bash`/shell can be flagged up on a per-issue basis.  No particular need for a label-per-variant-of-`bash`-as-above.

travishathaway:
@ifitchet,

Thanks for chiming in. So, what exactly are you recommending here? 😅 Should we add this label as I have proposed it?

travishathaway:
@kenodegard,

If msys2/cygwin bash and git-bash are essentially the same thing, then I think this label would make a good umbrella for both of those.

ifitchet:
@travishathaway I'm not recommending anything, just pointing out that there are more variants which fall into several implementation groups. 😄 
Personally, I would lean towards your `cli::windows-bash` as a "generic" placeholder (in the sense of "we know what it means" rather than the unwieldy `cli::POSIX-shell-on-Windows` and make it clear in the description which particular shell it was found on and/or affects.
My supposition would be that most issues will affect all POSIX-y shells on Windows and only a few will be specific to `fish`, `bash`, ...

kenodegard:
hm we already have https://github.com/conda/conda/labels/os%3A%3Awsl so issues with bash on WSL/WSL2 would be tagged with https://github.com/conda/conda/labels/os%3A%3Awsl and https://github.com/conda/conda/labels/cli%3A%3Aposix

maybe we should treat this as a different OS? `os::cygwin-mingw` (named after the "firsts")

so issues with git-bash would be `os::cygwin-mingw` and https://github.com/conda/conda/labels/cli%3A%3Aposix

never say never, a user could always come along with https://github.com/conda/conda/labels/cli%3A%3Afish on MSYS2 (https://packages.msys2.org/packages/fish)...

I'm also ok with keeping the proposed `cli::windows-bash` 🤷🏼‍♂️

ifitchet:
_Oof!_ Let's not get started on confusing AF as _mingw_, now https://www.mingw-w64.org/, is, IIUC, the Linux cross-compiler for generating native Windows binaries.
We're talking about Windows implementations of Unix toolchains to generate native Windows binaries[1].

I would also avoid combining Cygwin and MSYS2 in any name.  Cygwin is an all singing all dancing POSIX implementation (X11 and Emacs!) whereas MSYS2's (second attempt at a _M_ inimal _SYS_ tem to generate native Windows binaries -> _MSYS2_) relationship to it is a patched core POSIX library, `cygwin1.dll` -> `msys-2.0.dll`, with some subtle behavioural changes (including mounting `/usr/bin` over `/bin` which is where conda installs things and is partly why MSYS2 should only be used during a build and not, for conda, at runtime -- albeit the new `msys2-*` variants avoid this) and little else.

If `os::cygwin` is (potentially) a thing then `os::msys2` should be fine: both imply Windows but they clearly identify which "toolkit" is in use and the corresponding quirks.

[1] Native Windows binaries suggesting that they only need the stock Windows C/C++ runtimes (`msvcrt`/`ucrt`) although once GCC/Clang have been used you'll get references to their support libraries which should come out in the dependency tree wash.

travishathaway:
> I think my brain just broke, but I concur, adding TWO labels might be sufficient as well. So new plan:
> 
>     * `os::cygwin`
> 
>     * `os::msys2`
> 
> 
> ..to be combined with `cli::posix`, which would fit into the fact that users can combine for example `os::linux` and `cli::powershell` (because MS is now a major contributor to Linux) etc. #lifechoices
> 
> Do you agree @travishathaway?

Sounds good! I'll update the pull request.

travishathaway:
@ForgottenProgramme and @jezdez, ready for another round of review.

.github/labels.yml
```diff
@@ -66,6 +66,14 @@
     color: '#91e4e5'
     aliases: [type-shell-xonsh]
 
+# OS
+  - name: os::cygwin
+    description: pertains to Cygwin, commonly run on Windows
+    color: '#d0e0e3'
+  - name: os::msys2
+    description: pertains to MSYS2, commonly run on Windows
+    color: '#d0e0e3'
```
ForgottenProgramme, line 75: Are we making them the same color on purpose?

travishathaway: No, it's because that's the same color that's used for the other "OS" labels in the conda/infrastructure repository:

https://github.com/conda/infrastructure/blob/9bec3e981c5a4f41a37ee667084fd005e6d3ab62/.github/global.yml#L223-L248

.github/labels.yml
```diff
@@ -66,6 +66,14 @@
     color: '#91e4e5'
     aliases: [type-shell-xonsh]
 
+# OS
+  - name: os::cygwin
+    description: pertains to Cygwin, commonly run on Windows
+    color: '#d0e0e3'
+  - name: os::msys2
+    description: pertains to MSYS2, commonly run on Windows
```
kenodegard, line 74: Should we explicitly specify call out git-bash to help remember?

```suggestion
    description: pertains to MSYS2 and Git for Windows
```

travishathaway: Yeah, that's a good idea. Thanks!

.github/labels.yml
ForgottenProgramme: okay, got it

ForgottenProgramme (approved):

PR #14776, opened 2025-04-22
Add type hints to the context module

jaimergp:
### Description

A little bit of DX self-care is nice. xref https://github.com/conda/conda/issues/13437

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #14777, opened 2025-04-22
Do not use IndexedSet for deduplication

jaimergp:
### Description

Noticed this unnecessary `IndexedSet` incursions when all we need is `dict.fromkeys` for deduplication. I initially saw this in https://github.com/conda/conda/pull/14776 and thought it would be more widespread, so that's why I started in this branch, but to my surprise that's everything.

We do have a lot of (indiscriminate?) `IndexedSet` usage but I won't touch the apparently legitimate cases here. Just the `tuple(IndexedSet(...))` bits.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -23,7 +23,6 @@ from os.path import abspath, exists, expanduser, isdir, isfile, join
 from os.path import split as path_split
 from typing import TYPE_CHECKING
 
-from boltons.setutils import IndexedSet
 from frozendict import frozendict
 
 from .. import CONDA_SOURCE_ROOT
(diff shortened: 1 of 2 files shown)


jjhelmus:
Answering my own question, `set` does not guarantee that the order of elements is preserved, `dict.fromkeys` does.

jjhelmus:
[`unique_iter`](https://boltons.readthedocs.io/en/latest/iterutils.html#boltons.iterutils.unique_iter) from `boltons.iterutils` is another option as a replacement. Depends on if the goal is clarity or removing the use of boltons.

jjhelmus (approved):

jaimergp:
> Depends on if the goal is clarity or removing the use of boltons.

I'm slowly trying to get rid of `boltons` when there are (now) Python idioms that maybe didn't exist back when the code was written. e.g. `dict` didn't have guaranteed insertion order til recently.

kenodegard:
I'm 100% on board with this, just FYI we also have `conda.common.iterators.unique`

PR #14780, opened 2025-04-22
Pluginize `PrefixData` interoperability

### LINKED ISSUE #14070: `conda list` plugin hooks
### Checklist

### What is the idea?

`conda list` is a relatively simple command that prints the information returned by the `PrefixData` object, in different ways. I'd like to provide plugin hooks so we can:

- Extend what information is shown per conda package (e.g. license)
- Add new output formats (e.g. a conda-lock file)
- Extend which packages are listed, even if they are not conda-packaged.

### Why is this needed?

I see several benefits:

- We can refactor some code that doesn't really belong in the 'conda' core experience (see below)
- We can flexibly implement features like https://github.com/conda/conda/issues/11094
- We can add support for several output formats easily

### What should happen?

- Add `PrefixData` plugin hooks to collect non-conda packages.
  - `PrefixData` has some "extensions" to load packages that are _not_ installed by `conda`.
  - This is mostly what `pip_interop_enabled` does. It feels like bolted on and not really part of the design; more like a workaround to avoid pitfalls when users overwrite conda packages with `pip`-installed ones. 
  - In my opinion, this Python-specific code should be moved to `conda-pypi`, where a `PrefixData` plugin would be in charge of providing the required information.
  - Note: This is currently a biiit linked to the solver because `conda` tries to be smart about it, so this needs to be handled carefully. Definitely feasible though, not a blocker, but a word of caution :)

- Pluginize `conda list` formats. Right now they are hardcoded in the CLI layer, with some `if`s in the for loop that iterates the `PrefixData` object, or in a separate function altogether.

- Add a new format plugin that allows folks to add a template string to print their required information. Fields available are those in a `PrefixRecord` object.

### Additional Context

Without this, our `@EXPLICIT` integrations in `conda-pypi` are a bit [intricate](https://github.com/conda-incubator/conda-pypi/blob/5f985413b6b69d5c1ddf89840b95318a300b9634/conda_pypi/cli/list.py#L8-L22) (we can only print lines after `conda` is done...).

jaimergp:
### Description

Part of https://github.com/conda/conda/issues/14070

Adds plugin hooks for `PrefixData` loaders and refactors the current PyPI records loading logic as a plugin. The same can be used to list e.g. R packages, Node, and others.

### Checklist - did you ...


conda/core/prefix_data.py
```diff
@@ -320,8 +316,9 @@ def load(self):
             )
             for meta_file in conda_meta_json_paths:
                 self._load_single_record(meta_file)
-        if self._pip_interop_enabled:
-            self._load_site_packages()
+        if self._pip_interop_enabled:  # TODO: Rename as general interop
```
kenodegard, line 319: Alternatively (or in addition) the interoperability could be toggled per plugin?

conda/core/prefix_data.py
```diff
@@ -470,6 +467,7 @@ def _load_single_record(self, prefix_record_json_path):
     # region Python records
 
     @property
+    @deprecated("25.9", "26.3", addendum="Use PrefixData.get('python').")
```
kenodegard, line 470: any downsides to changing the body here to do that?

```python
@property
@deprecated("25.9", "26.3", addendum="Use PrefixData.get('python').")
def _python_pkg_record(self):
    """Return the prefix record for the package python."""
    return self.get("python")
```

news/14780-prefixdata-hooks
```diff
@@ -0,0 +1,22 @@
...
+
+* Add a plugin hook for `PrefixData` loaders and use it to wrap the existing PyPI interoperability features. These are now present in `conda.plugins.prefix_data_loaders.pypi`. (#14780)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Mark `conda.core.prefix_data.get_python_version_for_prefix()` as pending deprecation. Use `conda.core.prefix_data.PrefixData(prefix).get("python").version` instead. It will be permanently removed in 26.3. (#14750)
```
kenodegard, line 11: Thoughts on moving all of the deprecation info to the front and leave only the fix at the end:

```suggestion
* Mark `conda.core.prefix_data.get_python_version_for_prefix()` as pending deprecation, to be removed in 26.3. Use `conda.core.prefix_data.PrefixData(prefix).get("python").version` instead. (#14750)
```

conda/core/prefix_data.py
jaimergp: Yes but not at this API level, I think. Interoperability can be used in two very different ways:

- To report extra stuff in `conda list`
- To enable the solver to "see" more packages as "installed" and consider them as part of the solve process

For (1) we can have as much interoperability as we want because it's harmless, but for (2) it can cause a lot of havoc. So I was more leaning towards having a global bool switch `with_loader_plugins`. Then each plugin can be enabled / disabled via config I guess? It would also bring some trouble with the metaclass cache stuff...

But for now, we only need this part. I can rename it now, though.

conda/core/prefix_data.py
jaimergp: Since it's marked for deprecation, I'd rather not touch it.

news/14780-prefixdata-hooks
jaimergp: I like it! Makes more sense.

jezdez (approved):

conda/plugins/prefix_data_loaders/pypi.py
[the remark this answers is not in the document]
jaimergp: Yep, eventually I want to do that, but I don't think `conda-pypi` is ready for prime time yet. For now it will have to live here and then we can move it. The split is done, next step is just a copy-paste + deprecate exercise :D

jjhelmus:
These plugins will need the ability to register additional package types if they are to be effective. The `pypi` plugin introduced here only works because conda already has [package types for common Python distributions](https://github.com/conda/conda/blob/25.3.1/conda/models/enums.py#L119-L122) which [`conda.gateway.disk.read::read_python_record`](https://github.com/conda/conda/blob/25.3.1/conda/gateways/disk/read.py#L252-L321) uses. A future R or node plugin would not work unless the detected packages were of a type already supported by conda.

Additionally, are these plugins allowed to serialize information about the packages they detect? If so can they do this into `conda-meta`? Once done can conda manager (update/remove) these packages? What is the expected behavior if a plugin serializes this data and is then uninstalled?

jaimergp:
> These plugins will need the ability to register additional package types if they are to be effective.

Good catch. I wonder though if we should abandon this idea super-specific `PackageType`s and just have all the `conda` types, and then non-conda (with perhaps a distinction between manageable and unmanageable, but that's a solver detail that is not really used these days). 

> Additionally, are these plugins allowed to serialize information about the packages they detect? If so can they do this into conda-meta? Once done can conda manager (update/remove) these packages? What is the expected behavior if a plugin serializes this data and is then uninstalled?

I don't think that's needed or expected. `conda-meta/*.json` should only be for conda packages or, if we stretch it, packages that `conda` knows how to manage. Managing non-conda packages seems to be out of the scope for this plugin hook, where we only offer a "read-only" view of the prefix state. The idea seems to be that these `PackageRecords` only exist in memory so they are exposed in `conda list`.

jaimergp:
Yes, I think the most we can do for now is to expose non-conda packages to `conda list` so it can, well, list them in the output. Maybe to the solver as well so they show up as "installed" and they are not overwritten by a conda package, but that code path is not really finalized yet.

jjhelmus:
> Good catch. I wonder though if we should abandon this idea super-specific PackageTypes and just have all the conda types, and then non-conda (with perhaps a distinction between manageable and unmanageable, but that's a solver detail that is not really used these days).

I like this idea. Perhaps adding two new types, `OtherManageable` and `OtherNonManageable` would allow the flexibility for plugins to use.

conda/plugins/prefix_data_loaders/pypi.py
```diff
@@ -0,0 +1,161 @@
...
+from typing import TYPE_CHECKING
+
+from ...auxlib.exceptions import ValidationError
+from ...common.path import (
+    get_python_site_packages_short_path,
+    win_path_ok,
+)
+from ...common.pkg_formats.python import get_site_packages_anchor_files
+from ...gateways.disk.delete import rm_rf
+from ...gateways.disk.read import read_python_record
```
jjhelmus, line 22: Same as above, inline `read_python_record` and deprecate?

jjhelmus (approved):

conda/plugins/prefix_data_loaders/pypi.py
```diff
@@ -0,0 +1,161 @@
...
+from os.path import basename
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from ...auxlib.exceptions import ValidationError
+from ...common.path import (
+    get_python_site_packages_short_path,
+    win_path_ok,
+)
+from ...common.pkg_formats.python import get_site_packages_anchor_files
```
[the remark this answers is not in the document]
jaimergp: 100%. The whole `pkg_formats` subpackage is ripe to remove from there. It's oddly specific and it should definitely move. I'll do it here now that there have been a few reviews so the larger diff won't be that scary.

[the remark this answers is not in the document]
jaimergp: Actually I'll merge this as is and proceed with the refactors it in another PR, it's getting a bit unwieldy.

PR #14784, opened 2025-04-25
Run Pyright as part of CI

jezdez:
### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -329,7 +329,7 @@ jobs:
           fetch-depth: 0
 
       - name: Hash + Timestamp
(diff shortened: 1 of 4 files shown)


kenodegard:
I don't think we should merge this in, from experience we are unlikely to actually go look at it

But we can leave this open and use it as a metric as we add more typing in other PRs

Having this test here to occasionally run (similar to how every few months I've been rebasing the [elevate deprecations to errors PR](https://github.com/conda/conda/pull/12771) to detect unhandled deprecations) will help us gauge where to add typing first and how much more we have to do before we can enable this in pre-commit

jaimergp:
I've found this [post](https://discuss.python.org/t/is-there-any-tool-that-can-report-type-coverage-of-a-project/34962/4) where they added mypy CI + coverage.io support, so maybe that's a bit more useful. Join me at https://github.com/conda/conda/pull/14818!

PR #14794, opened 2025-05-01
Add installer plugin hooks

### LINKED ISSUE #16: Implement conda env installers plugin hook

### LINKED ISSUE #13833: `conda_env` installers / providers plugin hooks

soapy1:
### Description

This PR aims to break up the work started in https://github.com/conda/conda/pull/13928.

Addresses https://github.com/conda/conda-environments-project-mgmt/issues/16

It adds hooks that define the interface for installer plugins. 

For example, to create a pip installer plugin:

```
import sys
from subprocess import run
from conda import plugins
from conda.plugins.types import CondaInstaller, InstallerBase

python = sys.executable

class MyPipInstaller(InstallerBase):
    def __init__(self, **kwargs):
        pass

    def install(self, prefix, specs, *args, **kwargs):
        return "installing {specs} into {prefix}"

    def dry_run(self, prefix, specs, *args, **kwargs):
        return "DRYRUN: installing {specs} into {prefix}"


@plugins.hookimpl
def conda_installers():
    yield CondaInstaller(
        name="pip",
        types=["pip"],
        installer=MyPipInstaller,
    )
```

This PR *does not* add any installer plugins or use the installer plugin systems. For samples on how this can be achieved see the following branches:
* https://github.com/conda/conda/pull/13928/files#diff-b088be378fc4d3c8240b5a24fd9d51409bb64226eb874be18a10eafadec79f78
* https://github.com/conda/conda/compare/main...soapy1:conda:write-env-file-config-to-env-condarc?expand=1

### Checklist - did you ...


soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

jezdez (commented):
All of this seems reasonable, but of course it doesn't implement the use yet, and I'd love for us to port the use of conda.env.installers.base.get_installer to use this API as well. Would you be willing to work on that once we've merged this?

soapy1:
> and I'd love for us to port the use of conda.env.installers.base.get_installer to use this API as well. Would you be willing to work on that once we've merged this?

yes, there is still more work to do here!

jjhelmus:
Should we differentiate in the descriptions of this plugin hook between packages that are installed, visible and manageable by conda itself (e.g. .tar.bz2 and .conda artifacts) and those installed by plugins which are not visible or manageable by conda?  This is semantic but to users seeing "conda also natively supports installing packages with pip" might give the impression that pip installed packages have the same support as conda package.

soapy1:
Updated the docs!

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

jezdez:
@soapy1 @jaimergp do we intend to resurrect this again?

soapy1:
@jezdez yep, I'm planning on picking this up again this week. Will also incorporate some of the stuff we've learned from doing the environment spec and export plugin hooks :pray: 

Do you think there is any overlap here with the conda-pip/conda-pupa work? Or, are there any new requirements/constraints of the installer hooks?

soapy1:
pre-commit.ci autofix

dholth:
I did think it might be useful to say "I want these to be installed with pupa". The difference from pip would mean "install a pypi package but maybe get its dependences from conda packages" or "install a local source tree in a way that is more fully understood by conda"

jaimergp:
I wonder if the section names should be ecosystems instead of explicit tools; e.g. `pypi` instead of `pip`. Then users can configure which tool they want for each ecosystem. But that might be a departure of the initial tool-centric design.

dholth:
For pupa, the tool works differently than pip in ways the user might care about (do I want my dependencies of pypi packages from pypi or from conda-if-possible?) but accesses the same ecosystem.

jezdez:
Given the slippery slope of reinventing pip, we should be clear that we intend to adopt the patterns in pupa for conda-pypi that allow us to have a consistent way to install from PyPI and PyPI-compatible indexes. That's what the name in the ecosystem denotes. Let's focus on users not having to know what pupa is, since all they care is getting their packages from PyPI-like indexes installed without conda breaking their conda environmment or any other packages already installed there (e.g. from a conda package).

dholth:
@jezdez there is also "the easiest way to convert Python projects to .conda packages to share in the conda-package-sharing-ecosystem" feature, and the "modern editable installs" feature which have nothing to do with installer hooks but something to do with pupa

jezdez:
> @jezdez there is also "the easiest way to convert Python projects to .conda packages to share in the conda-package-sharing-ecosystem" feature, and the "modern editable installs" feature which have nothing to do with installer hooks but something to do with pupa

That is true, good reminder

docs/source/dev-guide/plugins/index.rst
```diff
@@ -101,6 +101,8 @@ For examples of how to use other plugin hooks, please read their respective docu
    environment_exporters
    environment_specifiers
    health_checks
+   installers
+   request_headers
```
jezdez, line 105:
```suggestion
```

jezdez:
I think we can close this given how we're progressed with conda-pypi?

PR #14796, opened 2025-05-02
Experiment -  using installer plugins

soapy1:
### Description

This PR is an experiment to try to remove refrences to cmd line argument in the install layer of conda. Eventually, it would be nice to get installers as plugins. And I think it would be good to be able to seperate notions of command line arguments from installing.

This experiment depends on the context object being able to:
1. collect configuration from command line arguments and
2. give them the correct precedence or merge them appropriately with other levels of config

### Checklist - did you ...


conda/cli/install.py
```diff
@@ -419,8 +418,6 @@ def install(args, parser, command="install"):
     if (isinstall or isremove) and args.update_modifier == NULL:
         update_modifier = UpdateModifier.FREEZE_INSTALLED
     deps_modifier = context.deps_modifier
-    if isupdate:
-        deps_modifier = context.deps_modifier or DepsModifier.UPDATE_SPECS
```
soapy1, line 423: `DepsModifier` does not have an attribute called `UPDATE_SPECS` https://github.com/conda/conda/blob/main/conda/base/constants.py#L201

PR #14807, opened 2025-05-06
Cleanup: `DepsModifier` does not have an attribute `UPDATE_SPECS`

soapy1:
### Description

This PR cleans up a small line in `cli/install.py`. The `DepsModifier` class does not have an attribute `UPTATE_SPECS`, it is defined as:

```
class DepsModifier(Enum):
    """Flags to enable alternate handling of dependencies."""

    NOT_SET = "not_set"  # default
    NO_DEPS = "no_deps"
    ONLY_DEPS = "only_deps"

    def __str__(self):
        return self.value
```
ref: https://github.com/conda/conda/blob/main/conda/base/constants.py#L201

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -21,7 +21,6 @@ from .. import CondaError
 from ..base.constants import (
     REPODATA_FN,
     ROOT_ENV_NAME,
-    DepsModifier,
     UpdateModifier,
 )
 from ..base.context import context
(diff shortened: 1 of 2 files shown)


dholth (approved):

soapy1:
> Is this just a case that is rarely or never triggered? It looks like this reference has been there for a while?


I think the `DepsModifier.UPDATE_SPECS` is never hit because `context.deps_modifier` is never `None`.  So like, if I update the diff to be 

```
if isupdate:
        deps_modifier = DepsModifier.UPDATE_SPECS
```

and then run a `conda update` you get the error:
```
      File "/home/sophia/projects/conda/conda/cli/install.py", line 425, in install
        deps_modifier = DepsModifier.UPDATE_SPECS
      File "/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/lib/python3.10/enum.py", line 437, in __getattr__
        raise AttributeError(name) from None
    AttributeError: UPDATE_SPECS
```

kenodegard (approved):

PR #14811, opened 2025-05-08
Make `PrefixData` lazy

jaimergp:
### Description

Hopefully a small improvement for very large environments with a lot of installed packages.

### Checklist - did you ...

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import json
 import os
 import re
+from collections import UserDict
 from logging import getLogger
 from os.path import basename, lexists
 from pathlib import Path
@@ -90,6 +91,16 @@ class PrefixDataType(type):
             prefix_data_instance = super().__call__(prefix_path, interoperability)
             PrefixData._cache_[cache_key] = prefix_data_instance
             return prefix_data_instance
+
+
+class PrefixRecordDict(UserDict):
+    """Lazily convert dict entries to PrefixRecord."""
+
+    def __getitem__(self, package_name: str) -> PrefixRecord:
+        record = self.data[package_name]
+        if not isinstance(record, PrefixRecord):
+            self.data[package_name] = record = PrefixRecord(**record)
+        return record
 
 
 class PrefixData(metaclass=PrefixDataType):
@@ -353,7 +364,7 @@ class PrefixData(metaclass=PrefixDataType):
 
     @time_recorder(module_name=__name__)
     def load(self) -> None:
-        self.__prefix_records = {}
+        self.__prefix_records = PrefixRecordDict()
         _conda_meta_dir = self.prefix_path / "conda-meta"
         if lexists(_conda_meta_dir):
             conda_meta_json_paths = (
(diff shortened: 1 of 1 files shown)


conda/core/prefix_data.py
```diff
@@ -486,36 +505,26 @@ def _load_single_record(self, prefix_record_json_path: PathType) -> None:
         log.debug("loading prefix record %s", prefix_record_json_path)
         with open(prefix_record_json_path) as fh:
             try:
-                json_data = json_load(fh.read())
+                data = json_load(fh.read())
```
dholth, line 508: https://github.com/conda/conda/blob/main/conda/common/serialize.py#L63-L64

jaimergp:
@dholth, given your experience with the lazy `SubdirData` refactor, do you think this is an ok adaptation for `PrefixData`? Let me know!

jaimergp:
If you are thinking of `conda list` users, not much of a change because in there we need to load everything.

But for the `PrefixData` consumers (which includes some bits of the solver plugins), `.query` and `.get` are now more efficient since we don't have to read and parse everything before getting the one result. See this snippet in `conda-libmamba-solver` for an example of the things we need to do to workaround the non-laziness:

https://github.com/conda/conda-libmamba-solver/blob/899c0a5ddd39fc461658aa0f376e9ed7127cb895/conda_libmamba_solver/solver.py#L1025-L1039

This is expected to be more noticeable in large environments (100+ packages), which is the case of Anaconda Distribution's `base`, where the `auto_update_conda` checks eat a couple seconds in the background, often when the solver has already finished finding a solution.

conda/core/prefix_data.py
```diff
@@ -93,6 +94,16 @@ def __call__(
             return prefix_data_instance
 
 
+class PrefixRecordDict(UserDict):
+    """Lazily convert dict entries to PrefixRecord."""
+
+    def __getitem__(self, package_name: str) -> PrefixRecord:
+        record = self.data[package_name]
+        if not isinstance(record, PrefixRecord):
+            self.data[package_name] = record = PrefixRecord(**record)
```
ForgottenProgramme, line 109: so this is the same as:
```
record=PrefixRecord(**record)
self.data[package_name]=record
```

jaimergp: Yep, see "Assign the same value to multiple variables" in https://note.nkmk.me/en/python-multi-variables-values/, for example.

jezdez:
I think we can safely move that to the next milestone.

jezdez:
@ryanskeith Since it's in the 25.11.x milestone, I just fix the merge conflicts, not sure if we're more ready than in the past

conda/core/prefix_data.py
```diff
@@ -543,6 +554,14 @@ def query(
         if isinstance(param, str):
             param = MatchSpec(param)
         if isinstance(param, MatchSpec):
+            if (
+                param.name
+                and param.name != "*"
+                and (record := self.get(param.name, None))
+            ):
+                if param.match(record):
+                    return (r for r in (record,))
```
jezdez, line 565:
```suggestion
                    # Yes, we do want to return the same
                    # type for all queries
                    return (r for r in (record,))
```

PR #14816, opened 2025-05-08
Strict environment yaml validation

soapy1:
### Description

Currently, the `YamlFileSpec` plugin, which provides functionality for reading `environment.yaml` files is very permissive in the files that is can handle. It validates that the provided file has a given set of keys, but if there is extra data, it will strip it out and continue to work perfectly ok. This flexibility is really great for forwards/backwards compatibility. However, environment_spec plugins must not be so permissive in handling environment files, so as not to interfer with other plugins. 

For example, given the current `YamlFileSpec.cam_handle` implementation, the `YamlFileSpec` plugin will be selected for all the following files even if it shouldn't:

valid environment.yaml
```
name: nlp
dependencies:
  - nltk
```

not a valid environment.yaml - actually an empty json files
```
{ }
```

not a valid environment.yaml - actually a json file
```
{
   "name": "mysimpletest",
   "conda_deps": ["numpy", "pandas"]
}
```

not a valid environment.yaml - a made up yaml format
```
env_version: 2.0
name: nlp
deps:
  - nltk
```

This PR suggests resolving this issue by introducing a `strict` mode to the environment.yaml key validation step. When strict mode is enabled, reading a file will return an error if an invalid key is encountered.

#### Warning

This approach will be a breaking change for users that have `environment.yaml` files with keys outside of ("name", "dependencies", "prefix", "channels", "variables"). `environment.yaml` files that use additional keys will no longer be detected as valid and conda will return a `EnvironmentSpecPluginNotDetected` error.

### Checklist - did you ...

### DIFF
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -13,7 +13,11 @@ from ..common.path import expand
 from ..common.serialize import json, yaml_safe_dump, yaml_safe_load
 from ..core.prefix_data import PrefixData
 from ..deprecations import deprecated
-from ..exceptions import EnvironmentFileEmpty, EnvironmentFileNotFound
+from ..exceptions import (
+    EnvironmentFileEmpty,
+    EnvironmentFileNotFound,
+    EnvironmentSectionNotValid,
+)
 from ..gateways.connection.download import download_text
 from ..gateways.connection.session import CONDA_SESSION_SCHEMES
 from ..history import History
@@ -24,8 +28,10 @@ from ..models.match_spec import MatchSpec
 VALID_KEYS = ("name", "dependencies", "prefix", "channels", "variables")
 
 
-def validate_keys(data, kwargs):
-    """Check for unknown keys, remove them and print a warning"""
+def validate_keys(data, kwargs, strict=False):
+    """Check for unknown keys. In strict mode, will raise an error.
+    Otherwise remove the unknown keys and print a warning.
+    """
     invalid_keys = []
     new_data = data.copy() if data else {}
     for key in data.keys():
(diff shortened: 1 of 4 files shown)


conda/env/env.py
```diff
@@ -26,8 +30,10 @@
 VALID_KEYS = ("name", "dependencies", "prefix", "channels", "variables")
 
 
-def validate_keys(data, kwargs):
-    """Check for unknown keys, remove them and print a warning"""
+def validate_keys(data, kwargs, strict=False):
```
jaimergp, line 31: Let's add type hints here.

conda/exceptions.py
```diff
@@ -1295,6 +1295,12 @@ class PluginError(CondaError):
     pass
 
 
+class EnvironmentSectionNotValid(CondaError):
+    def __init__(self, filename: str, sections: list[str],plural: str, verb: str, *args, **kwargs):
+        msg = f"The following section{plural} on '{filename}' {verb} invalid and will be ignored: {dashlist(sections)}"
```
jaimergp, line 1373: If this is an error, why are "ignoring" sections?


```suggestion
        msg = f"The following section{plural} on '{filename}' {verb} invalid: {dashlist(sections)}"
```

conda/env/env.py
```diff
@@ -39,13 +45,16 @@ def validate_keys(data, kwargs):
...
+        if strict:
+            raise EnvironmentSectionNotValid(filename, invalid_keys, plural, verb)
+        else:
+            print(
+                f"\nEnvironmentSectionNotValid: The following section{plural} on "
+                f"'{filename}' {verb} invalid and will be ignored:"
+            )
+            for key in invalid_keys:
+                print(f" - {key}")
+            print()
```
jaimergp, line 55: Let's use `log.warning` for this, in a single call.

jaimergp (commented):
Looks good!

I'll add a reference to [CEP 24](https://github.com/conda/ceps/blob/main/cep-0024.md), where this file was standardized. The CEP doesn't go into detail about erroring, but it does recommend to warn. As long as the `strict` boolean is only used internally for plugin detection, we are fine, but in my opinion we should not raise errors in the CLI if the environment file contains extra keys, just warnings.

Also, let's make sure the added signatures have type hints.

PR #14725, opened 2025-04-01
Ensure conda does not allow for mixing input file types

soapy1:
### Description

Refs #14684

Depends on resolving https://github.com/conda/conda-environments-project-mgmt/issues/15 and https://github.com/conda/conda-environments-project-mgmt/issues/18

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #14820, opened 2025-05-09
Implement explicit environment spec file support based on `CEP 23`.

### LINKED ISSUE #47: Add environment spec plugin for `explicit` input file types
### What is the idea?

Conda uses [`explicit` spec files to build identical environments on the same platforms ](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments). 

Currently, explicit files are handled in the `cli/install.py` module. It would be nice if the parsing of these files were handled in a plugin.

### Why is this needed?

This will help drive towards `conda` and `conda env` commands becoming more consistent with eachother.

### What should happen?

conda should have a built in plugin that handles explicit spec files

### LINKED ISSUE #14819: Add environment specifier plugin support for explicit input files (CEP23)
### Checklist

### What is the idea?

`conda create --file explicit.txt` supports installing from an explicit specs file, but this format is handled directly by the CLI and not through the environment spec plugin system introduced for extensibility.

### Why is this needed?

This brings feature parity between `conda env` and `conda create/install/update` by allowing all environment specification formats—including explicit files—to be handled uniformly via plugins.

### What should happen?

Allow explicit spec files (e.g. `--file explicit.txt`) to be handled by the environment spec plugin interface. This would enable support for explicit files to be provided as a plugin, similar to other environment spec formats.

### Additional Context

Refs https://github.com/conda/conda/issues/14684 and https://github.com/conda/conda-environments-project-mgmt/issues/15

jezdez:
### Description

- Fix #14819.
- Fix https://github.com/conda/conda-environments-project-mgmt/issues/47

### Checklist - did you ...


jjhelmus:
Creating an environment using this plugin requires a solve. A solve is not required when using `conda create --file explicit.txt`. Should this be documented or mentioned to users?

jjhelmus:
Testing this out it looks like all explicit files are detected as being handled by the `requirements.txt` plugin as well.

```
❯ cat explicit_single.txt 
@EXPLICIT
https://conda.anaconda.org/conda-forge/osx-64/libcurl-7.69.1-hc0b9707_0.tar.bz2

❯ conda env create --prefix blah --file explicit_single.txt

PluginError: Too many plugins found that can handle the environment file 'explicit_single.txt':

explicit, requirements.txt

Please make sure that you don't have any overlapping plugins installed.
```

conda/env/installers/conda.py
```diff
@@ -43,8 +59,35 @@ def dry_run(specs, args, env, *_, **kwargs):
...
+    :return: Installation result information
+    :rtype: dict
+
+    .. note::
+        This implementation follows CEP-23, which states: "When an explicit input file is
+        processed, the conda client SHOULD NOT invoke a solver."
+    """
+    # Handle explicit environments separately per CEP-23 requirements
+    if isinstance(env, ExplicitEnvironment):
+        return _install_explicit_environment(prefix, specs, env)
```
jjhelmus, line 88: This is a special case but one that will come up for any plugins that reads a definition for a locked environment (including pixi.lock or conda-lock.yml files).  In these cases a solve is unnecessary, should not be performed and in some cases (e.g. packages outside of a channel are referenced) may not be possible. I think providing a path for plugins to express this information would be useful but likely requires an expansion of the `Environment` class.

For pixi.lock and conda-lock.yml, parsing can return a set of `PackageRecord` instances for the conda requirements, where as requirement and environment.yaml files would return a set of `MatchSpec` instances.

conda/env/installers/conda.py
```diff
@@ -53,16 +96,79 @@ def install(prefix, specs, args, env, *_, **kwargs):
             update_modifier=UpdateModifier.FREEZE_INSTALLED,
         )
     except (UnsatisfiableError, SystemExit) as exc:
-        # See this comment for 'allow_retry' details
-        # https://github.com/conda/conda/blob/b4592e9eb0/conda/cli/install.py#L417-L429
-        if not getattr(exc, "allow_retry", True):
```
soapy1, line 58: what is the motivation for removing this retry bit?

soapy1 (commented):
Neat! This works when I (pull in the changes from #14725 and) try to reproduce the issue described in https://github.com/conda/conda/pull/14725#issuecomment-2867623756.

soapy1:
> I wonder if we should merge those PRs together?

For me, its ok if they are seperate since they implementation doesn't overlap and they have 2 distinct goals. But, if you think it would be helpful to merge them, that sounds fine as well.

soapy1 (approved):

conda/env/installers/conda.py
jjhelmus: Another option would be to attaching this metadata to the `Environment` class itself which is returned from `EnvironmentSpecBase.environment` method. 

Either option is reasonable and will allow for plugins for various lock formats.

conda/env/installers/conda.py
```diff
@@ -2,21 +2,30 @@
...
 
 def _solve(prefix, specs, args, env, *_, **kwargs):
-    """Solve the environment"""
+    """Solve the environment.
+
+    :param prefix: Installation target directory
+    :param specs: Package specifications to install
+    :param args: Command-line arguments
+    :param env: Environment object
+    :return: Solver object
```
jjhelmus, line 27: Is it worthing adding or convert these to type annotation?

conda/env/installers/conda.py
```diff
@@ -53,16 +96,78 @@ def install(prefix, specs, args, env, *_, **kwargs):
...
+        try:
+            with open(filename) as f:
+                explicit_specs = [
+                    line.strip()
+                    for line in f
+                    if line.strip() and not line.strip().startswith("#")
+                ]
+            log.debug(f"Using package specs from explicit file: {filename}")
+        except OSError as e:
+            log.warning(f"Could not read explicit file {filename}: {e}")
```
jjhelmus, line 144: This logic exists in two (or even three see conda.cli.spec_from_line) locations. Can this be de-duped or collected into a utility function. 
Fine leaving this as a future refactor.

conda/env/installers/conda.py
```diff
@@ -53,16 +96,78 @@ def install(prefix, specs, args, env, *_, **kwargs):
...
-            prune=getattr(args, "prune", False), update_modifier=NULL
+        # Handle solver errors (original code continues here)
+        raise exc
+
+    return _execute_transaction(unlink_link_transaction)
+
+
+def _install_explicit_environment(
+    prefix: str, specs: list, env: ExplicitEnvironment
+) -> dict:
```
jjhelmus, line 107:
```suggestion
) -> dict[str, Any]:
```

soapy1:
In general, I think this looks good to me and seems to work in all the ways I expect it to. I want to call out 2 other things happening in parallel here:

1) once the environment model stuff get's merged we'll want to change the environment spec plugins to be using that environment opposed to the `conda.env.env.Environment`. I think this change is fully compatible with what is happening in this PR. But merging this first will add to the work required to move over to that desired final state of all plugins using the new environment model.

2) @jjhelmus's [pr for implementing installing from a lockfile](https://github.com/conda/conda/pull/14868/files#diff-770280f144ccd2f54a49acb17077dd9109c45b8cd1fd8f26c594e728dbfad6c5) proposes another way of installing explicit packages. Is this the direction we want to go in for installing packages directly? If so, is this PR compatible with being able to move over to this new method?

conda/env/installers/conda.py
jjhelmus: Is there a need to get this feature ASAP? If not I'd prefer to wait until the Environment class is merged. Once that work is done this can be re-worked to use that class.

soapy1:
> I'm not sure I follow, are you asking if we can first merge this environment model and then this feature branch?

Sorry, that was a lot of words on my part. I think this is good to merge as is.

conda/env/installers/conda.py
```diff
@@ -33,18 +55,53 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
-    solved_env = Environment(
+    return Environment(
         name=env.name, dependencies=[str(p) for p in pkgs], channels=env.channels
     )
-    return solved_env
 
 
-def install(prefix, specs, args, env, *_, **kwargs):
-    """Install packages into an environment"""
+def install(prefix: str, specs: list, args, env, *_, **kwargs) -> dict:
```
ForgottenProgramme:
```suggestion
def install(prefix: str, specs: list, args: Namespace, env: Environment, *_, **kwargs) -> dict:
```

jezdez:
pre-commit.ci autofix

conda/gateways/disk/read.py
```diff
@@ -32,13 +32,36 @@
...
+    :param filename: Path to the file
+    :return: List of non-empty, non-comment lines
+    :raises OSError: If the file cannot be read
+    :raises FileNotFoundError: If the file does not exist
+    """
+    # Convert to Path object for consistent handling
+    path = Path(filename)
+    if not path.exists():
+        raise FileNotFoundError(f"File {path} does not exist")
+    # Use Path.read_text() to read the file content
```
jaimergp: This can be race-y, `try/except` better?

conda/gateways/disk/read.py
```diff
@@ -32,13 +32,36 @@
...
+    content = path.read_text()
+    return [
+        line.strip()
+        for line in content.splitlines()
+        if line.strip() and not line.strip().startswith("#")
+    ]
```
jaimergp: We won't probably ingest huge files but I think it'd be better to use the classic `for line in f` idioms, and maybe use a generator

```suggestion
    with path.open() as f:
        for line in f:
            line = line.strip()
            if line and not line.startswith("#"):
                yield line
```

jaimergp: This is actually what the previous function was doing so I'm not sure (yet) what this refactor is for (I guess I'll find out later).

conda/env/explicit.py
```diff
@@ -0,0 +1,51 @@
...
+        name : str, optional
+            The name of the environment
+        dependencies : List[str], optional
+            The list of package specifications (URLs in this case)
+        channels : List[str], optional
+            The list of channels
+        prefix : str, optional
+            The installation prefix
+        filename : str, optional
+            The path to the explicit file this environment was created from
```
jaimergp: Are we included types in the docstrings now? I think we are deferring those to the type hints. Also I see that we are contributing different types of docstring syntax (rst and numpy, mostly) in different PRs it would be nice to finally standardize on this.

conda/env/specs/requirements.py
```diff
@@ -65,19 +92,173 @@ def can_handle(self) -> bool:
...
+                # or just: channel::package==version=build
+                line = line.split("::", 1)[1]
+
+                # This ensures the line is in the format expected by explicit()
+                # which requires direct URLs to packages, not channel-prefixed specifications
+            # Handle file paths by expanding variables and converting to URLs
+            if not line.startswith(("http://", "https://", "file://")):
+                line = os.path.expanduser(os.path.expandvars(line))
+                if os.path.isabs(line) and not line.startswith("file://"):
+                    line = path_to_url(line)
```
jaimergp: The previous implementation in `conda create` uses a [regex](https://github.com/conda/conda/blob/e341a1f8df6da0b599e71db09566ffc5e535550a/conda/misc.py#L52-L59) to do all this validation. Can we just use that here?

If we won't, are we fine with subtly different implementations in `conda create` and `conda env create`?

conda/env/specs/requirements.py
```diff
@@ -65,19 +92,173 @@ def can_handle(self) -> bool:
...
-    def environment(self):
-        dependencies = []
-        with open(self.filename) as reqfile:
-            for line in reqfile:
-                line = line.strip()
-                if not line or line.startswith("#"):
-                    continue
-                dependencies.append(line)
-        return Environment(dependencies=dependencies)
+    def environment(self) -> Environment:
```
jaimergp:
```suggestion
    def environment(self) -> ExplicitEnvironment:
```

conda/env/specs/requirements.py
```diff
@@ -65,19 +92,173 @@ def can_handle(self) -> bool:
...
+        """
+        try:
+            packages = self._parse_explicit_file()
+            if packages is None:
+                raise ValueError(f"Unable to handle file {self.filename}: {self.msg}")
+
+            # Create an explicit environment with the packages
+            # Using the typed ExplicitEnvironment class signals that this is from an explicit file
+            # and should bypass the solver according to CEP-23
+            return ExplicitEnvironment(dependencies=packages, filename=self.filename)
```
jaimergp: Not sure I'm a fan of adding this class when we are about to receive a generalized API for this in https://github.com/conda/conda/pull/14870

conda/env/explicit.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Explicit environment implementation for conda."""
+
+from __future__ import annotations
+
+from .env import Environment
+
+
+class ExplicitEnvironment(Environment):
```
jaimergp: Isn't the purpose of this whole class to flip a boolean in `Environment` informing that the specs here contained are fully resolved MatchSpecs and that they should be taken verbatim into a transaction bypassing the solver?

In other words, can we add that boolean to `Environment` and call it a day?

jaimergp (changes requested):
I have several concerns with this PR:

- I'm not sure I would consider `@EXPLICIT` a different format than `requirements.txt`, so I'm not sure I would have a different backend for this instead of teaching the existing `requirements` parser to handle explicit bits. After the only difference is what type of spec they ship (query vs resolved).
- Also, we are reimplementing behaviour existing in `conda create --file`, but with slightly different behavior (conversion of filepaths to URLs?); I don't recall the code doing that in `conda create`, but maybe I missed something. I'd strongly prefer if both commands end up using the very same logic.
- I'm not sure we want to introduce yet another `Environment` subclass, when we are talking about deprecating it in favor of https://github.com/conda/conda/pull/14870. That class will be able to handle both cases (explicit and not explicit) gracefully, so I don't know why we can't wait til that class ships before accepting this.


With all that in mind, , I thought we had a general goal of merging `conda env create` with `conda create` and end up deprecating `conda env create`. So I think it's more valuable to spend time teaching `conda create` how to do what `conda env` does, instead of the other way around. This is what @soapy1's refactors are directed to. Are we going to allow users to run `conda env create -f explicit.txt` in addition to `conda create --file explicit.txt` only to remove it later? I don't think it makes much sense to spend resources in this area, to be honest. We are adding more code to do the same thing in a different area of the CLI, adding tests and allocating engineering resources to something that will be eventually phased out.

If this was an exercise to try out the new plugin hooks, then cool with me, but I fail to see the added value if we accept that `conda env` will be eventually removed. 

There's a scenario where the conda env plugins end up backing up the logic behind `conda create`. In that case, this code might be useful in the future when we get there, but it feels like we should focus first on pluginizing `conda create` input file support after having gone through the necessary refactors.

conda/env/explicit.py
jezdez: Sure, that was mostly to use subclassing and type checks in the code, I can update that

conda/env/explicit.py
jezdez: I can change the style, we're using different ones for a while now

conda/env/installers/conda.py
jezdez: @jaimergp that's the similar conversation you had above, I'll change to the boolean in the environment class

conda/env/specs/requirements.py
jezdez: Sure

conda/gateways/disk/read.py
jezdez: Good idea

conda/env/specs/requirements.py
jezdez: Agree, when this PR was opened, we didn't have the class, so there are some leftovers from that like that, the good news with #14870 merge-ready, I can fix this

conda/gateways/disk/read.py
jezdez: Sure, I like generators, easy to switch

jezdez:
> I have several concerns with this PR:
> 
>     * I'm not sure I would consider `@EXPLICIT` a different format than `requirements.txt`, so I'm not sure I would have a different backend for this instead of teaching the existing `requirements` parser to handle explicit bits. After the only difference is what type of spec they ship (query vs resolved).

CEP 23 makes a differentiation, which is why I implemented it that way, but honestly I'm not hung-up about it.

> 
>     * Also, we are reimplementing behaviour existing in `conda create --file`, but with slightly different behavior (conversion of filepaths to URLs?); I don't recall the code doing that in `conda create`, but maybe I missed something. I'd strongly prefer if both commands end up using the very same logic.

You're assuming that this solves every step of the way, this is a snapshot while we didn't have a number of other pieces in place, including the environment class and the install refactor. In other words, I agree that they should end up using the same code paths.

>     * I'm not sure we want to introduce yet another `Environment` subclass, when we are talking about deprecating it in favor of [Add a data model for an environment #14870](https://github.com/conda/conda/pull/14870). That class will be able to handle both cases (explicit and not explicit) gracefully, so I don't know why we can't wait til that class ships before accepting this.

Yep, when this PR was started we only barely started the discussion that may end in something like #14870, I've asked for review for a bit now.

> With all that in mind, , I thought we had a general goal of merging `conda env create` with `conda create` and end up deprecating `conda env create`. So I think it's more valuable to spend time teaching `conda create` how to do what `conda env` does, instead of the other way around. This is what @soapy1's refactors are directed to. Are we going to allow users to run `conda env create -f explicit.txt` in addition to `conda create --file explicit.txt` only to remove it later? I don't think it makes much sense to spend resources in this area, to be honest. We are adding more code to do the same thing in a different area of the CLI, adding tests and allocating engineering resources to something that will be eventually phased out.

I'm sorry, but I need to clearly call out the unreasonable response to this, this PR doesn't stop unifying `conda env create` and `conda create`. I've discussed with Sophia that those refactors are worth having, but they haven't been merged yet.

> If this was an exercise to try out the new plugin hooks, then cool with me, but I fail to see the added value if we accept that `conda env` will be eventually removed.
> 
> There's a scenario where the conda env plugins end up backing up the logic behind `conda create`. In that case, this code might be useful in the future when we get there, but it feels like we should focus first on pluginizing `conda create` input file support after having gone through the necessary refactors.

Okay, we can do that, we already have the environment_specifier plugin hook, and have already proposal to refactor the installer backend to elevate the way conda env does things.

jaimergp:
> Yep, when this PR was started we only barely started the discussion that may end in something like https://github.com/conda/conda/pull/14870, I've asked for review for a bit now.

Ok, I missed that temporal context, thanks for the reminder. That PR will get merged anytime now, so we can use it the moment it's in main.

> I'm sorry, but I need to clearly call out the unreasonable response to this, this PR doesn't stop unifying conda env create and conda create. I've discussed with Sophia that those refactors are worth having, but they haven't been merged yet.

Glad to hear that. For a moment I thought we were abandoning that adventure but good to see it's still on.

---

So to recap, I'll be happy to approve this once it uses Sophia's `Environment` data model. My other concerns about `conda create` vs `conda env create` are now gone with the replies above, since we will keep working towards that step by step. Thanks for baring with me!

jezdez:
I want to do a light refactor as suggested by you and Jonathan, and base it on top of #14870, I think that'll be a good step in between, I'll do another call for review when it happens. If other PRs like the install refactor land first (even older in #14794), we can reconsider if this is still giving us forward momentum.

conda/env/env.py
```diff
@@ -180,6 +180,7 @@ class Dependencies(dict):
     def __init__(self, raw, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.raw = raw
+        self.explicit = bool(raw and raw[0] == "@EXPLICIT")
```
jaimergp: AFAIK `@EXPLICIT` doesn't have to be the first line in the lockfile. So maybe this should be 
```suggestion
        self.explicit = bool(raw and "@EXPLICIT" in raw)
```

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
     subdirs = IndexedSet(basename(url) for url in _channel_priority_map)
 
     solver_backend = context.plugin_manager.get_cached_solver_backend()
+    assert solver_backend is not None  # get_cached_solver_backend never returns None
```
jaimergp: Should we raise a `CondaError` subclass here?

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
+    """Install packages into a conda environment.
+
+    This function handles two main paths:
+    1. For environments with explicit_specs (from @EXPLICIT files): Bypasses the solver
+       and directly installs packages using conda.misc.explicit() as required by CEP-23.
+    2. For regular Environment instances: Uses the solver to determine the optimal
+       package set before installation.
+
+    :param prefix: The target installation path for the environment
+    :type prefix: str
```
jaimergp: I don't think we need to double annotate the types here if we have type hints already.

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
+def install(
+    prefix: str, specs: list, args: Namespace, env: Environment, *_, **kwargs
+) -> dict | None:
+    """Install packages into a conda environment.
+
+    This function handles two main paths:
+    1. For environments with explicit_specs (from @EXPLICIT files): Bypasses the solver
+       and directly installs packages using conda.misc.explicit() as required by CEP-23.
+    2. For regular Environment instances: Uses the solver to determine the optimal
+       package set before installation.
```
jaimergp, line 84: Love this bit! Super clear :)

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
+
+        # Determine which package specs to use (priority order):
+        explicit_specs = None
+        filename = env.filename
+
+        # 1. Try to read from original file if available (most reliable source)
+        if filename:
+            explicit_specs = list(yield_lines(filename))
+            if explicit_specs:
+                log.debug(f"Using package specs from explicit file: {filename}")
```
jaimergp:
```suggestion
                log.debug("Using package specs from explicit file: %s", filename})
```

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
+            else:
+                log.warning(f"Could not read explicit file {filename} or file is empty")
```
jaimergp:
```suggestion
                log.warning("Could not read explicit file %s or file is empty", filename)
```

conda/gateways/disk/read.py
```diff
@@ -50,12 +50,17 @@ def yield_lines(path):
 
     """
     try:
-        with open_utf8(path) as fh:
-            for line in fh:
+        # Convert to Path object for consistent handling
```
jaimergp: I think this comment is a bit superfluous and we don't need it.

tests/env/test_explicit_environment.py
```diff
@@ -0,0 +1,89 @@
...
+    env = Environment(dependencies=["@EXPLICIT", "test"])
+    assert env.dependencies.explicit
+
+    regular_env = Environment()
+    assert not regular_env.dependencies.explicit
+
+
+def test_explicit_environment_initialization(explicit_urls):
+    """Test Environment initialization with explicit parameters."""
+    explicit_deps = ["@EXPLICIT"] + explicit_urls
```
jaimergp, line 38: Can we add a test where `@EXPLICIT` is not the first line, just to make sure it works elsewhere in the file?

conda/env/installers/conda.py
```diff
@@ -29,22 +48,90 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
...
+
+        # 2-3. Fall back to provided specs or dependencies from environment
+        if not explicit_specs:
+            explicit_specs = specs if specs else env.dependencies.raw
+            log_msg = (
+                "Using specs provided to installer"
+                if specs
+                else "Using dependencies from environment"
+            )
+            log.debug(log_msg)
```
jaimergp: I don't fully understand the purpose of this fallback. If the explicit_dependencies are empty, there's nothing to install.

What we are missing here is to use the non-explicit specs as the `history` contents, but I think that's missing in the current `explicit()` function API anyway.

jaimergp (commented):
I have some questions about the precedence of the `explicit()` code paths in the absence of `explicit_specs`. Other than that, the rest looks solid. I'll take a look locally tomorrow and extend my review with maybe some more comments. In the meantime we can address the non controversial feedback. Thanks!

conda/misc.py
```diff
@@ -142,12 +155,16 @@ def explicit(specs, prefix, verbose=False, force_extract=True, index=None):
             else:
                 precs_to_remove.append(prec)
 
+    # Only record user-requested specs in history, not all packages
+    # This prevents dependencies from polluting the user's request history
+    update_specs_for_history = tuple(requested_specs) if requested_specs else ()
```
jezdez: Can we do that?

conda/env/installers/conda.py
jezdez: You're right! I've refactored this in [`8be6a51` (#14820)](https://github.com/conda/conda/pull/14820/commits/8be6a5159b0b6e0dbc3162d0a5dc97e72bb754ae) and essentially track what the requested specs were, and also only update the history with those

tests/env/test_explicit_environment.py
jezdez: Added in [`c462de1` (#14820)](https://github.com/conda/conda/pull/14820/commits/c462de12a78774a4be24325523bc894f800e02c7)

conda/gateways/disk/read.py
jezdez: [`72082aa` (#14820)](https://github.com/conda/conda/pull/14820/commits/72082aab1570db2a07a5d1e862c89e4dab9b0101)

conda/env/installers/conda.py
jezdez: [`72082aa` (#14820)](https://github.com/conda/conda/pull/14820/commits/72082aab1570db2a07a5d1e862c89e4dab9b0101)

conda/env/installers/conda.py
jezdez: [`72082aa` (#14820)](https://github.com/conda/conda/pull/14820/commits/72082aab1570db2a07a5d1e862c89e4dab9b0101)

conda/env/installers/conda.py
jezdez: [`72082aa` (#14820)](https://github.com/conda/conda/pull/14820/commits/72082aab1570db2a07a5d1e862c89e4dab9b0101)

conda/env/installers/conda.py
jezdez: [`11d2563` (#14820)](https://github.com/conda/conda/pull/14820/commits/11d256314e183879434ff6705fc37701c847b3fe)

conda/env/installers/conda.py
```diff
@@ -29,22 +48,101 @@ def _solve(prefix, specs, args, env, *_, **kwargs):
     subdirs = IndexedSet(basename(url) for url in _channel_priority_map)
 
     solver_backend = context.plugin_manager.get_cached_solver_backend()
+    if solver_backend is None:
```
soapy1, line 51: In what situation is the `solver_backend` None?

soapy1 (approved):

conda/misc.py
jaimergp: To maintain backwards compatibility I'd fallback to `tuple(sp[0] for sp in specs_pcrecs if sp[0]),` instead of `()`. You could argue it's wrong, but I don't think we are better off with an empty history instead.

conda/env/installers/conda.py
jezdez: The `get_cached_solver_backend` returns the `backend` attribute of the loaded solver plugin, but that doesn't enforce that `backend` is of `type[Solver]`, it's just a hint and we don't do runtime validation. So this is just a little bit of defensive programming to prevent malicious or misconfigured solvers to breaking the core installer functionality and let it fail with a graceful `CondaError` instead of a `TypeError: 'NoneType' object is not callable`.

conda/misc.py
jezdez: Done in [`c8ca580` (#14820)](https://github.com/conda/conda/pull/14820/commits/c8ca5806f5b721b74242255bd6e4f819fc8e494f)

jaimergp (approved):

PR #14821, opened 2025-05-09
Add `SolvedRecord`

jaimergp:
### Description

This sits between `PackageRecord` and `PrefixRecord`, simply adding `requested_spec` so it can be used in lockfiles without requiring the artifact on disk.

xref https://github.com/conda-incubator/conda-lockfiles/issues/8

### Checklist - did you ...


jaimergp:
pre-commit.ci autofix

dholth (approved):

conda/models/records.py
```diff
@@ -636,9 +648,6 @@ class PrefixRecord(PackageRecord):
 
     # app = ComposableField(App, required=False)
 
-    #: str: The :class:`MatchSpec` that the user requested or ``None`` if dependency it was installed as a dependency.
-    requested_spec = StringField(required=False)
```
jjhelmus, line 640: Does this need to be deprecated from `PrefixRecord` before removal?

jjhelmus:
I like the idea of adding this new class but since it is not used would it make more sense to wait on this change until the business logic requires it?

conda/models/records.py
jaimergp: `PrefixRecord` inherits from `SolvedRecord`, so we didn't really remove it.

jaimergp:
Given how we were planning to use this in plugins, I thought it would be nice to have this simple change ready in the May release, instead of July, but no strong feelings.

travishathaway (approved):

PR #14870, opened 2025-05-21
Add a data model for an environment

soapy1:
### Description

This PR adds a data model for a conda Environment. In follow up PRs this data model will be used to:
* refactor the `cli.install` module
* update the environment_spec interface to use this Environment model (opposed to the environment.yaml specific model)
* save declarative environment state

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -76,8 +76,7 @@ CONDA_HOMEPAGE_URL: Final = "https://conda.io"
 ERROR_UPLOAD_URL: Final = "https://conda.io/conda-post/unexpected-error"
 DEFAULTS_CHANNEL_NAME: Final = "defaults"
 
-KNOWN_SUBDIRS: Final = (
-    "noarch",
+PLATFORMS: Final = (
     "emscripten-wasm32",
     "wasi-wasm32",
     "freebsd-64",
@@ -97,6 +96,10 @@ KNOWN_SUBDIRS: Final = (
     "win-arm64",
     "zos-z",
 )
+KNOWN_SUBDIRS: Final = (
+    "noarch",
+) 
+KNOWN_SUBDIRS += PLATFORMS
 PLATFORM_DIRECTORIES = KNOWN_SUBDIRS
 
 RECOGNIZED_URL_SCHEMES: Final = ("http", "https", "ftp", "s3", "file")
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -96,9 +96,7 @@ PLATFORMS: Final = (
     "win-arm64",
     "zos-z",
 )
-KNOWN_SUBDIRS: Final = (
-    "noarch",
-) 
+KNOWN_SUBDIRS: Final = ("noarch",)
 KNOWN_SUBDIRS += PLATFORMS
 PLATFORM_DIRECTORIES = KNOWN_SUBDIRS
 
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -96,8 +96,7 @@ PLATFORMS: Final = (
     "win-arm64",
     "zos-z",
 )
-KNOWN_SUBDIRS: Final = ("noarch",)
-KNOWN_SUBDIRS += PLATFORMS
+KNOWN_SUBDIRS: Final = ("noarch", *PLATFORMS)
 PLATFORM_DIRECTORIES = KNOWN_SUBDIRS
 
 RECOGNIZED_URL_SCHEMES: Final = ("http", "https", "ftp", "s3", "file")
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -1,0 +1,136 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Conda environment data model"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from logging import getLogger
+from typing import TYPE_CHECKING
+
+from ..exceptions import CondaValueError
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from .match_spec import MatchSpec
+    from .records import PackageRecord
+
+
+log = getLogger(__name__)
+
+
+@dataclass
+class Environment:
+    # Prefix the environment is installed into (required).
+    prefix: str
+
+    # The platform this environment may be installed on (required)
+    platform: str
+
+    # Environment level configuration, eg. channels, solver options, etc.
+    config: dict[str, Any] = field(default_factory=dict)
+
+    # Map of other package types that conda can install. For example pypi packages.
+    external_packages: dict[str, list]  = field(default_factory=dict)
+
+    # The complete list of specs for the environment.
+    # eg. after a solve, or from an explicit environemnt spec
+    explicit_specs: list[PackageRecord] = field(default_factory=list)
+
+    # Environment name
+    name: str |  None = None
+
+    # User requested specs for this environment.
+    requested_specs: list[MatchSpec] = field(default_factory=list)
+
+    # Environment variables to be applied to the environment.
+    variables: dict[str, str] = field(default_factory=dict)
+
+    def __post_init__(self):
+        # an environment must have a name of prefix
+        if not self.prefix:
+            raise CondaValueError("'Environment' needs a 'prefix'.")
+
+    @classmethod
+    def merge(cls, *environments):
+        """
+        Keeps first name and/or prefix.
+        Concatenates and deduplicates requirements.
+        Reduces configuration and variables (last key wins).
+        """
+        name = None
+        prefix = None
+        platform = None
+        names = [env.name for env in environments if env.name]
+        prefixes = [env.prefix for env in environments if env.prefix]
+        
+        if names:
+            name = names[0]
+            if len(names) > 1:
+                log.debug("Several names passed %s. Picking first one %s", names, name)
+        
+        if prefixes:
+            prefix = prefixes[0]
+            if len(prefixes) > 1:
+                log.debug(
+                    "Several prefixes passed %s. Picking first one %s", prefixes, prefix
+                )
+        
+        platforms = [env.platform for env in environments if env.platform]
+        # Ensure that all environments have the same platform
+        if len(set(platforms)) == 1:
+            platform = platforms[0]
+        else:
+            raise CondaValueError(f"Conda can not merge environments of different platforms. Recieved environments with plafroms {platforms}")
+
+        requested_specs = list(
+            dict.fromkeys(
+                requirement for env in environments for requirement in env.requested_specs
+            )
+        )
+
+        explicit_specs = list(
+            dict.fromkeys(
+                requirement for env in environments for requirement in env.explicit_specs
+            )
+        )
+
+        variables = {k: v for env in environments for (k, v) in env.variables.items()}
+
+        config = {}
+        external_packages = {}
+        for env in environments:
+            # Config items can be any type, merge them so that lists get 
+            # concatenated, dicts get merged, and primitive types get clobbered.
+            for k, v in env.config.items():
+                if k not in config:
+                    config[k] = v
+                elif isinstance(config[k], list) and isinstance(v, list):
+                    config[k].extend(v)
+                elif isinstance(config[k], dict) and isinstance(v, dict):
+                    config[k].update(v)
+                else:
+                    log.debug("merging configs, clobbering value %s with value %s")
+                    config[k] = v
+
+            # External packages map values are always lists of strings. So,
+            # we'll want to concatentate each list.
+            for k, v in env.external_packages.items():
+                if k in external_packages:
+                    for val in v:
+                        if val not in external_packages[k]:
+                            external_packages[k].append(val)
+                elif isinstance(v, list):
+                    external_packages[k] = v
+
+        return cls(
+            config=config,
+            external_packages= external_packages,
+            explicit_specs=explicit_specs,
+            name=name,
+            platform=platform,
+            prefix=prefix,
+            requested_specs=requested_specs,
+            variables=variables,
+        )
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -1,6 +1,6 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
-"""Conda environment data model"""
+"""EXPERIMENTAL Conda environment data model"""
 
 from __future__ import annotations
 
@@ -22,6 +22,12 @@ log = getLogger(__name__)
 
 @dataclass
 class Environment:
+    """
+    **Experimental** While experimental, expect both major and minor changes across minor releases.
+
+    Data model for a conda environment.
+    """
+
     #: Prefix the environment is installed into (required).
     prefix: str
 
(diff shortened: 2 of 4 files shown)


soapy1:
pre-commit.ci autofix

conda/models/environment.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Environment level configuration, eg. channels, solver options, etc.
+    config: dict[str, Any] = field(default_factory=dict)
+
+    # Map of other package types that conda can install. For example pypi packages.
+    external_packages: dict[str, list] = field(default_factory=dict)
+
+    # The complete list of specs for the environment.
+    # eg. after a solve, or from an explicit environemnt spec
+    explicit_specs: list[MatchSpec] = field(default_factory=list)
```
jjhelmus, line 38: I don't think explicit specs can always be modeled completely using `MatchSpec`. A different, new?, type might be needed here.

conda/models/environment.py
```diff
@@ -0,0 +1,141 @@
...
+    prefix: str
+
+    # The platform this environment may be installed on (required)
+    platform: str
+
+    # Environment level configuration, eg. channels, solver options, etc.
+    config: dict[str, Any] = field(default_factory=dict)
+
+    # Map of other package types that conda can install. For example pypi packages.
+    external_packages: dict[str, list] = field(default_factory=dict)
```
jjhelmus, line 34:
```suggestion
    external_packages: dict[str, list[Any] = field(default_factory=dict)
```

soapy1: hmmmm, intially I was imaging this as a `list[str]`. But, I can see where we may want to have a more specific datatype. Taking this approach will have some implications on how environments get merged. Thanks for raising this. Will look into it!

conda/models/environment.py
soapy1: I'm definitely not super set to using MatchSpecs here. What kind of information is missing from a MatchSpec? What kind of information would you want to capture in a new explicit type?

jaimergp: Aren't they (proto-)`PackageRecord`? As in, they are guaranteed to match a single record.

soapy1: Ya, I think a `PackageRecord` is a good option. From the [docstring](https://github.com/conda/conda/blob/main/conda/models/records.py#L247):

> Representation of a concrete package archive (tarball or .conda file).

    It captures all the relevant information about a given package archive, including its source,
    in the following attributes.

jjhelmus: If `PackageRecord` is used here it would be good to be explicit if they expected to be sorted topologically. 

If topoligical order is required then the creator of the Environment is responsible for ensuring that dependencies are listed before dependents. 

If any order is allowed then ordering will need to be done when converting an Environment into a concrete environment on disk.

soapy1: Ok, these are some good insights! Given this information, I think `PackageRecord` is the right thing to use here. `PackageCacheRecord` is for packages that are already downloaded. I think that kind of information should not be in the Environment object. 

> If PackageRecord is used here it would be good to be explicit if they expected to be sorted topologically.

I think the Environment object's only responsibility is to hold all the information about the environment. And then some operation is done against the Environment object to get the transactions that need to be done to install. The sorting should be done as part of this step, I think. This poc follows that process https://github.com/conda/conda/pull/14113/files#diff-7f08889b327d1140a6232cc0429f997e3dfd19e74814252312d581d2e5911ec0R28

soapy1:
pre-commit.ci autofix

conda/models/environment.py
jaimergp: We also have `SolvedRecord` now, which is `PackageRecord` + a `requested_spec` field, in case that's a better fit.

soapy1: The sources of information that can populate an environment object are:
* the cli
* environment specs
* lockfiles

I don't think any of these have a notion of a package associated with a requested spec, so I don't think it's a good fit.

soapy1: Ya, I'd say an environment should just have `explicit_specs` or `requested_specs`. Currently in conda, if you [mix these types of specs you will get an error](https://docs.google.com/document/d/1Pr0R01he8T3Bg0-YTlnBR9xlwjvG_lOW_UBuSdE60qU/edit?userstoinvite=jhelmus%40anaconda.com&sharingaction=manageaccess&role=writer&tab=t.0). Though, the environment model itself probably doesn't need to make that assumption?  

In any case, given both a set of requested specs and all the package records for an environment, it should be pretty straightforward to build the `SolvedRecords` for the environment.

jaimergp: I think there's a use case for having both `explicit_specs` and `requested_specs` present, but in that case we'd treat `requested_specs` differently. Let's recap:

- `requested_specs` only: fed to solver to obtain a list of PackageRecords that is then fed to the Transaction object.
- `explicit_specs` only: we build the Transaction object directly, no solver. 
- `explicit_specs` + `requested_specs`: we build the Transaction object directly, no solver, but use `requested_specs` to populate the `conda-meta/history` file in the new environment. The names in `requested_specs` need to be a subset of `explicit_specs` and match their corresponding records. Otherwise it's an input error.

conda/models/environment.py
```diff
@@ -0,0 +1,142 @@
...
+    from .match_spec import MatchSpec
+    from .records import PackageRecord
+
+
+log = getLogger(__name__)
+
+
+@dataclass
+class Environment:
+    # Prefix the environment is installed into (required).
```
jaimergp, line 25: Small thing but I think this comment syntax is parsed magically by Sphinx in some way:

```suggestion
    #: Prefix the environment is installed into (required).
```

soapy1:
pre-commit.ci autofix

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+    from .records import PackageRecord
+
+
+log = getLogger(__name__)
+
+
+@dataclass
+class Environment:
+    """
+    **Beta** While in beta, expect both major and minor changes across minor releases.
```
soapy1, line 26: Marking as in beta, following the pattern defined in the [api module](https://github.com/conda/conda/blob/main/conda/api.py).

soapy1:
pre-commit.ci autofix

conda/models/environment.py
jaimergp: xref https://github.com/conda/conda/issues/14895, we'll probably standardize those api module ideas via that issue.

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+                    "Several prefixes passed %s. Picking first one %s", prefixes, prefix
+                )
+
+        platforms = [env.platform for env in environments if env.platform]
+        # Ensure that all environments have the same platform
+        if len(set(platforms)) == 1:
+            platform = platforms[0]
+        else:
+            raise CondaValueError(
+                f"Conda can not merge environments of different platforms. Recieved environments with plafroms {platforms}"
```
jaimergp:
```suggestion
                "Conda can not merge environments of different platforms. "
                f"Received environments with platforms: {platforms}"
```

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        )
+
+        variables = {k: v for env in environments for (k, v) in env.variables.items()}
+
+        config = {}
+        external_packages = {}
+        for env in environments:
+            # Config items can be any type, merge them so that lists get
+            # concatenated, dicts get merged, and primitive types get clobbered.
```
jaimergp, line 149: Just highlighting that the `.merge()` logic in [`conda.common.configuration`](https://github.com/conda/conda/blob/main/conda/common/configuration.py) seems to be a bit more involved than this. Maybe this simplification is ok for the involved keys, but in that case we should make sure we are erroring out on unexpected keys.

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+                elif isinstance(config[k], list) and isinstance(v, list):
+                    config[k].extend(v)
+                elif isinstance(config[k], dict) and isinstance(v, dict):
+                    config[k].update(v)
+                else:
+                    log.debug("merging configs, clobbering value %s with value %s")
+                    config[k] = v
+
+            # External packages map values are always lists of strings. So,
+            # we'll want to concatentate each list.
```
jaimergp:
```suggestion
            # we'll want to concatenate each list.
```

news/14870-add-environment-data-model
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a data model to represent an environment. (#14870)
```
jaimergp:
```suggestion
* Add an experimental data model to represent an environment. (#14870)
```

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+    prefix: str
+
+    #: The platform this environment may be installed on (required)
+    platform: str
+
+    #: Environment level configuration, eg. channels, solver options, etc.
+    config: dict[str, Any] = field(default_factory=dict)
+
+    #: Map of other package types that conda can install. For example pypi packages.
+    external_packages: dict[str, list[Any]] = field(default_factory=dict)
```
jaimergp:
```suggestion
    external_packages: dict[str, list[str]] = field(default_factory=dict)
```

Are we sure this is Any?

conda/models/environment.py
```diff
@@ -0,0 +1,151 @@
...
+    #: Environment name
+    name: str | None = None
+
+    #: User requested specs for this environment.
+    requested_packages: list[MatchSpec] = field(default_factory=list)
+
+    #: Environment variables to be applied to the environment.
+    variables: dict[str, str] = field(default_factory=dict)
+
+    def __post_init__(self):
```
jaimergp, line 59: I think we need some extra validation here:

- `platform` must be one of the non-noarch [KNOWN_SUBDIRS](https://github.com/conda/conda/blob/eb3b48191ae0135dd16c9a5d7cb40131d312dc77/conda/base/constants.py#L79)
- if both `requested_packages` and `explicit_packages` are set, each `requested_packages` entry should match one (and only one) record in `explicit_packages`.
- `explicit_packages` cannot contain records for more than one package name. IOW, `len([record.name for record in explicit_packages]) == len(explicit_packages)`.

jaimergp:
I think this is in good shape to go ahead with approvals and merge given that:

- Typos are fixed
- Validations are extended a bit
- The merge code paths are checked briefly for compatibility
- We mark this as experimental as discussed in https://github.com/conda/conda/issues/14895

Good job with the discovery and design! 👏

conda/models/environment.py
soapy1: I think the validations for the environment model should only fail for environment configurations that are outright not possible.

> platform must be one of the non-noarch [KNOWN_SUBDIRS](https://github.com/conda/conda/blob/eb3b48191ae0135dd16c9a5d7cb40131d312dc77/conda/base/constants.py#L79)

I think this is a good one. You can not have an environment for a platform that conda does not support.

> if both requested_packages and explicit_packages are set, each requested_packages entry should match one (and only one) record in explicit_packages.

I think this is too strict. Currently, users are not able to install a mix of explicit packages and requested specs. I think this validation should be done in not this layer.

> explicit_packages cannot contain records for more than one package name. IOW, len([record.name for record in explicit_packages]) == len(explicit_packages).

I think this is also a good one. You can not have multiple of same package installed in an environment.

conda/models/environment.py
soapy1: The involved keys will be:
```
"aggressive_update_packages",
"channel_priority",
"channels" (comprised of: "custom_channels", "custom_multichannels")
"default_channels",
"channel_settings",
"deps_modifier",
"disallowed_packages",
“default_packages”
"pinned_packages",
"repodata_fns",
"solver"
"update_modifier",
"use_only_tar_bz2",
```

soapy1:
pre-commit.ci autofix

conda/models/environment.py
soapy1: The configuration class adds additional complexity in merging by adding logic for things like `final` markers in `.condarc` configs.  This kind of information will be lost is the context is serialized to a simple dict.

conda/models/environment.py
jezdez: I don't think the **Beta** is a reliable standard to implement, I'd rather keep this more vague ("beta" has a very distinct meaning in the software development lifecycle theory, it's in the middle of "alpha", "release candidates" and "final"). I think we were pretty lax with this related to the solver API (read: the API module is a lie, _all_ of conda is effectively the API).
```suggestion
    Experimental, see https://github.com/conda/conda/issues/14895 for more information.
```

conda/models/environment.py
jezdez: Wondering, will the markers have already been evaluated by the time we reach this point? Those markers _are_ heavily used and cannot be ignored.

jezdez (commented):
Nicely done! I think this looks excellent already, my only concerns:
- is the config merge behavior acceptable to be dropped or am I misunderstanding the code?
- CodeCov says there are a bunch of branches that aren't tested yet, could you add cases for the ValueError please?

conda/models/environment.py
jaimergp: I assume we are taking these from the initialized `context` at this point, so my answer would be yes.

conda/models/environment.py
jaimergp: > Currently, users are not able to install a mix of explicit packages and requested specs. I think this validation should be done in not this layer.

My suggestion here comes from the notion that in this mode `requested_specs` is just metadata for `explicit_specs`. i.e. the list of specs that will end up in the `history` file, and the values that will populate the `requested_spec` field in the corresponding `conda-meta/*.json` file for that record.

conda/models/environment.py
soapy1: I agree with Jaime, these values should be pulled from the initialized contexts, so markers are taken into account. This only applies to generating an environemnt from a cli command. I think it's still an open question for how users should be able to define config in their environemnt specs and how conda should merge this into the environemtn object. 

I have another POC here https://github.com/conda/conda/compare/main...soapy1:conda:env-model-with-config that tries to make this config field be a new `EnvironmentConfig` type.

conda/models/environment.py
soapy1: Added

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

conda/models/environment.py
jezdez: Okay, that answers it for me, as long as we're essentially not sidestepping the centralized config loading, I'm fine with this as-is.

conda/models/environment.py
```diff
@@ -0,0 +1,180 @@
...
+
+    #: Prefix the environment is installed into (required).
+    prefix: str
+
+    #: The platform this environment may be installed on (required)
+    platform: str
+
+    #: Environment level configuration, eg. channels, solver options, etc.
+    #: TODO: may need to think more about the type of this field and how
+    #:       conda should be merging configs between environments
```
jezdez, line 40: 👍🏻

jezdez (dismissed):

conda/base/constants.py
```diff
@@ -97,6 +96,8 @@
     "win-arm64",
     "zos-z",
 )
+KNOWN_SUBDIRS: Final = ("noarch",)
+KNOWN_SUBDIRS += PLATFORMS
```
jaimergp:
```suggestion
KNOWN_SUBDIRS: Final = ("noarch", *PLATFORMS)
```

The type checkers will complain otherwise

jaimergp (approved):

PR #14877, opened 2025-05-23
Explicitly selected env spec plugin

soapy1:
### Description

This PR:
* adds a cli arg `--env-spec`/`--environment-specifier` to `conda env` commands to users can explicitly select the environment spec plugin they want to use when creating/updating an environment using an environment file
* removes references to the function `env.specs.detect`
* does not substitue improving the environment spec plugin detection process

This PR suggests that environment spec plugin selection works as follows:
* If no plugin is selected, conda should try to detect what environment spec plugin should be used against the file
* If the user requests a specific plugin with the `--env-spec` cli arg or `env_spec` config parameter, conda will try to use the plugin with that name
  * if a environment spec plugin with that name is not registered, return an error
  * if the requested environment spec plugin can not handle the input file, return an error

Here are some examples of this change:

The environment spec plugin is automatically detected if no argument is provided
```
$ conda env create --file tests/env/support/simple.yml
Channels:
 - conda-forge
Platform: linux-64
Solving environment: done
#
#
#     $ conda activate nlp
#
#
#     $ conda deactivate
```

Return an error to the user if an unregistered plugin is requested
```
$ conda env create --file tests/env/support/simple.yml --env-spec idontextis

                        [--env-spec-plugin {binstar,environment.yml,requirements.txt}] [--no-default-packages] [--json]
                        [--console CONSOLE] [-v] [-q] [-d] [-y] [--solver {classic,libmamba}] [--subdir SUBDIR]
                        [remote_definition]
conda env create: error: argument --env-spec-plugin: invalid choice: 'idontextis' (choose from 'binstar', 'environment.yml', 'requirements.txt')
```

Return an error to the user if a plugin that won't work is requested
```
$ conda env create --file tests/env/support/simple.yml --env-spec binstar

PluginError: Requested plugin 'binstar' is unable to handle environment spec 'tests/env/support/simple.yml'
```

Set the environment spec using env vars
```
$ CONDA_ENV_SPEC=idontexist conda env create --file tests/env/support/simple.yml

CondaValueError: You have chosen an unrecognized environment specifier type (idontexist). Choose one of: binstar, environment.yml, requirements.txt
```

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1389,7 +1389,7 @@ class Context(Configuration):
                 # prevent modifications to envs marked with conda-meta/frozen
             ),
             "Plugin Configuration": ("no_plugins",),
-            "Environment Managment Configuration": ("environment_specifier",),
+            "Experimental": ("environment_specifier",),
         }
 
     def get_descriptions(self) -> dict[str, str]:
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -245,6 +245,12 @@ class Context(Configuration):
     clobber = ParameterLoader(PrimitiveParameter(False))
     changeps1 = ParameterLoader(PrimitiveParameter(True))
     env_prompt = ParameterLoader(PrimitiveParameter("({default_env}) "))
+
+    # environment_specifier is an EXPERIMENTAL config parameter
+    environment_specifier = ParameterLoader(
+        PrimitiveParameter(None, element_type=(str, NoneType)), aliases=("env_spec",)
+    )
+
     create_default_packages = ParameterLoader(
         SequenceParameter(PrimitiveParameter("", element_type=str))
     )
@@ -1383,6 +1389,7 @@ class Context(Configuration):
                 # prevent modifications to envs marked with conda-meta/frozen
             ),
             "Plugin Configuration": ("no_plugins",),
+            "Environment Managment Configuration": ("environment_specifier",),
         }
 
     def get_descriptions(self) -> dict[str, str]:
@@ -1634,6 +1641,14 @@ class Context(Configuration):
                 of '{name}' if the active environment is a conda named environment ('-n'
                 flag), or otherwise holds the value of '{prefix}'. Templating uses python's
                 str.format() method.
+                """
+            ),
+            environment_specifier=dals(
+                """
+                **EXPERIMENTAL** While experimental, expect both major and minor changes across minor releases.
+
+                The name of the environment specifier plugin that should be used for this context.
+                If not specified, the plugin manager will try to detect the plugin to use.
                 """
             ),
             execute_threads=dals(
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -376,7 +376,7 @@ def validate_environment_files_consistency(files: list[str]) -> None:
 
     # Get types for all files using the plugin manager
     file_types = {
-        file: context.plugin_manager.get_environment_specifier_name(file)
+        file: context.plugin_manager.get_environment_specifier(file).name
         for file in files
     }
     # If there's more than one unique type, raise an error
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -576,6 +576,19 @@ def add_parser_verbose(parser: ArgumentParser | _ArgumentGroup) -> None:
     )
 
 
+def add_parser_environment_specifier(p: ArgumentParser) -> None:
+    from ..base.context import context
+    from ..common.constants import NULL
+
+    p.add_argument(
+        "--environment-specifier",
+        "--env-spec",  # for brevity
+        choices=context.plugin_manager.get_environment_specifiers(),
+        default=NULL,
+        help="(EXPERIMENTAL) Specify the environment specifier plugin to use.",
+    )
+
+
 def comma_separated_stripped(value: str) -> list[str]:
     """
     Custom type for argparse to handle comma-separated strings with stripping
--- a/conda/cli/main_compare.py
+++ b/conda/cli/main_compare.py
@@ -102,7 +102,6 @@ def compare_packages(active_pkgs, specification_pkgs) -> tuple[int, list[str]]:
 def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import context
     from ..core.prefix_data import PrefixData
-    from ..env import specs
     from ..exceptions import SpecNotFound
     from ..gateways.connection.session import CONDA_SESSION_SCHEMES
     from .common import stdout_json
@@ -118,7 +117,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         else:
             filename = abspath(expanduser(expandvars(args.file)))
 
-        spec = specs.detect(name=args.name, filename=filename, directory=os.getcwd())
+        spec_hook = context.plugin_manager.get_environment_specifier(
+            source=filename,
+            name=context.environment_specifier,
+        )
+        spec = spec_hook.environment_spec(filename)
         env = spec.environment
 
         if args.prefix is None and args.name is None:
--- a/conda/env/specs/__init__.py
+++ b/conda/env/specs/__init__.py
@@ -26,7 +26,7 @@ if TYPE_CHECKING:
 @deprecated(
     "25.9",
     "26.3",
-    addendum="Use conda.base.context.plugin_manager.get_environment_specifiers.",
+    addendum="Use conda.base.context.plugin_manager.detect_environment_specifier.",
 )
 def get_spec_class_from_file(filename: str) -> FileSpecTypes:
     """
@@ -71,8 +71,8 @@ def detect(
     :raises SpecNotFound: Raised if no suitable spec class could be found given the input
     """
     try:
-        spec_hook = context.plugin_manager.get_environment_specifiers(
-            filename=filename,
+        spec_hook = context.plugin_manager.detect_environment_specifier(
+            source=filename,
         )
     except EnvironmentSpecPluginNotDetected as e:
         raise SpecNotFound(e.message)
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1297,7 +1297,7 @@ class SpecNotFound(CondaError):
 
 
 class EnvironmentSpecPluginNotDetected(SpecNotFound):
-    def __init__(self, name, plugin_names, *args, **kwargs):
+    def __init__(self, name: str, plugin_names: Iterable[str], *args, **kwargs):
         self.name = name
         msg = dals(
             f"""
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1287,7 +1287,16 @@ class EnvironmentFileNotDownloaded(CondaError):
         super().__init__(msg, *args, **kwargs)
 
 
-class EnvironmentSpecPluginNotDetected(CondaError):
+class PluginError(CondaError):
+    pass
+
+
+class SpecNotFound(CondaError):
+    def __init__(self, msg: str, *args, **kwargs):
+        super().__init__(msg, *args, **kwargs)
+
+
+class EnvironmentSpecPluginNotDetected(SpecNotFound):
     def __init__(self, name, plugin_names, *args, **kwargs):
         self.name = name
         msg = dals(
@@ -1299,15 +1308,6 @@ class EnvironmentSpecPluginNotDetected(CondaError):
             """
         )
         super().__init__(msg, *args, **kwargs)
-
-
-class SpecNotFound(CondaError):
-    def __init__(self, msg: str, *args, **kwargs):
-        super().__init__(msg, *args, **kwargs)
-
-
-class PluginError(CondaError):
-    pass
 
 
 def maybe_raise(error: BaseException, context: Context):
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -508,39 +508,92 @@ class CondaPluginManager(pluggy.PluginManager):
         """
         return PluginConfig(data)
 
-    def get_environment_specifiers(self, filename: str) -> CondaEnvironmentSpecifier:
+    def get_environment_specifiers(self) -> dict[str, CondaEnvironmentSpecifier]:
         """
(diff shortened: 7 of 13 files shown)


conda/base/context.py
```diff
@@ -245,6 +245,7 @@ class Context(Configuration):
     clobber = ParameterLoader(PrimitiveParameter(False))
     changeps1 = ParameterLoader(PrimitiveParameter(True))
     env_prompt = ParameterLoader(PrimitiveParameter("({default_env}) "))
+    env_spec_plugin = ParameterLoader(PrimitiveParameter(""))
```
soapy1, line 248: should users also be able to set this in a `condarc`?

conda/plugins/manager.py
```diff
@@ -554,6 +595,22 @@ def get_environment_specifiers(self, filename: str) -> CondaEnvironmentSpecifier
             name=filename, plugin_names=[hook.name for hook in hooks]
         )
 
+    def get_environment_specifier(self, filename: str, plugin_name=None) -> CondaEnvironmentSpecifier:
```
soapy1, line 598: Is it worth is to have this function? Can callers of the plugin manager just call `detect_environment_spec_plugin_from_file` or `get_explicit_environment_specifier` directly?

jezdez (changes requested):
This needs some regiggering of words I think, we should absolutely not call out the existence of "plugins" in the cli or config as that is not consistent how we've mapped plugin provided functionality to the user interface.

conda/base/context.py
```diff
@@ -1372,7 +1373,10 @@ def category_map(self) -> dict[str, tuple[str, ...]]:
                 "protect_frozen_envs",
                 # prevent modifications to envs marked with conda-meta/frozen
             ),
-            "Plugin Configuration": ("no_plugins",),
+            "Plugin Configuration": (
+                "env_spec_plugin",
```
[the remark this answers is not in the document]
soapy1: True! I don't think this configuration really matches any of the other sections. Could do "Basic Conda Configuration" or "Output, Prompt, and Flow Control Configuration"? What do you think?

conda/cli/main_compare.py
```diff
@@ -118,7 +117,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         else:
             filename = abspath(expanduser(expandvars(args.file)))
 
-        spec = specs.detect(name=args.name, filename=filename, directory=os.getcwd())
+        spec_hook = context.plugin_manager.get_environment_specifier(
+            filename=filename,
+            plugin_name=context.env_spec_plugin,
+        )
+        spec = spec_hook.environment_spec(filename)
```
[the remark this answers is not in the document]
soapy1: Can do if you prefer this. IMO there are so many other types of specs, that it's helpful to be redundant in this case.

conda/env/specs/binstar.py
```diff
@@ -36,7 +36,7 @@ class BinstarSpec(EnvironmentSpecBase):
 
     msg = None
 
-    def __init__(self, name=None):
+    def __init__(self, name=None, **kwargs):
```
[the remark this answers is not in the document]
soapy1: mmmm, maybe I'm getting a bit ahead of myself here. This really doesn't solve any problems and isn't in scope for this pr. Thanks for keeping me honest!

conda/plugins/manager.py
```diff
@@ -508,12 +509,59 @@ def get_config(self, data) -> PluginConfig:
...
+    def get_environment_specifier_names(self) -> dict[str, CondaEnvironmentSpecifier]:
+        """
+        Returns a list of all the environment specifier plugin names
+        """
+        return {
+            hook.name.lower(): hook
+            for hook in self.get_hook_results("environment_specifiers")
+        }
+
+    def get_explicit_environment_specifier(
```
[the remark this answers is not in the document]
soapy1: >I'm not sure if I understand this method name, what is explicit about it?

This method is meant to allow callers to explicitly provide the name of the plugin they want to load (instead of relying on detecting the plugin). Merging this into `get_environment_specifier`.

> And why does the plugin implementation leak into the plugin manager?

I don't really see where this implementation leak is happening? Are you referring to initializing the plugin and calling `can_handle()`? Those are things defined in the plugin interface, so should be safe to use in this context.

conda/plugins/manager.py
```diff
@@ -508,12 +509,59 @@ def get_config(self, data) -> PluginConfig:
...
+
+        if found[0].environment_spec(filename).can_handle():
+            return found[0]
+        else:
+            raise PluginError(
+                f"Requested plugin '{plugin_name}' is unable to handle environment spec '{filename}'"
+            )
+
+    def detect_environment_spec_plugin_from_file(
+        self, filename: str
```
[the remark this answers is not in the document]
soapy1: hmmm, this is a good question. Maybe this is named a little bit strict. Like does a spec need to come from a file on disk, or is it ok to come from a remote file, or not file (like the binstar spec). I would be into renaming this from `filename` -> `source` or something.

conda/env/specs/__init__.py
```diff
@@ -26,7 +26,7 @@
 @deprecated(
     "25.9",
     "26.3",
-    addendum="Use conda.base.context.plugin_manager.get_environment_specifiers.",
+    addendum="Use conda.base.context.plugin_manager.detect_environment_specifier.",
```
soapy1, line 29: It's maybe spicy to be changing this deprecation warning here? Not sure how conda's process for this is.

conda/base/context.py
```diff
@@ -245,6 +245,9 @@ class Context(Configuration):
     clobber = ParameterLoader(PrimitiveParameter(False))
     changeps1 = ParameterLoader(PrimitiveParameter(True))
     env_prompt = ParameterLoader(PrimitiveParameter("({default_env}) "))
+    environment_specifier = ParameterLoader(
```
soapy1, line 250: should this get marked as experimental?

jezdez (dismissed):

conda/base/context.py
jezdez: yes

conda/env/specs/__init__.py
jezdez: Oh, hmm, I'm not sure. @kenodegard did we have a case like this before? Could we keep a get_environment_specifiers stub around as an alias until this deprecation cycle is done?

soapy1:
pre-commit.ci autofix

soapy1:
hmmm, looks like the docs failure is unrelated to this PR. Error has to do with access to anaconda.org?
```

CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/r/noarch/repodata.json>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.

If your current network has https://repo.anaconda.com blocked, please file
a support request with your network engineering team.

'https//repo.anaconda.com/pkgs/r/noarch'
```

conda/cli/common.py
```diff
@@ -376,7 +376,7 @@ def validate_environment_files_consistency(files: list[str]) -> None:
 
     # Get types for all files using the plugin manager
     file_types = {
-        file: context.plugin_manager.get_environment_specifier_name(file)
+        file: context.plugin_manager.get_environment_specifier(file).name
```
soapy1, line 379: https://github.com/conda/conda/pull/14725/ also introduced some changes to the plugin manager that are at odds with the changes made to the plugin manager in this PR. After rebasing, I've straighened this out to be consistent with what we chatted about here https://github.com/conda/conda/pull/14877/files#r2109000518. This change is a side effect of this.

soapy1:
pre-commit.ci autofix

conda/base/context.py
```diff
@@ -1383,6 +1389,7 @@ def category_map(self) -> dict[str, tuple[str, ...]]:
                 # prevent modifications to envs marked with conda-meta/frozen
             ),
             "Plugin Configuration": ("no_plugins",),
+            "Environment Managment Configuration": ("environment_specifier",),
```
jaimergp:
```suggestion
            "Environment Management Configuration": ("environment_specifier",),
```

Hm, do we need a new section for this? If it's experimental, we should maybe put it under "undocumented" for now?

conda/cli/helpers.py
```diff
@@ -576,6 +576,19 @@ def add_parser_verbose(parser: ArgumentParser | _ArgumentGroup) -> None:
...
 
 
+def add_parser_environment_specifier(p: ArgumentParser) -> None:
+    from ..base.context import context
+    from ..common.constants import NULL
+
+    p.add_argument(
+        "--environment-specifier",
+        "--env-spec",  # for brevity
+        choices=context.plugin_manager.get_environment_specifiers(),
```
jaimergp, line 586: Can we depend on `context` bis at CLI initialization? My understanding is that the context needs the parsed CLI args to finish initialization, so I'm not sure we have discovered anything about plugins yet here 🤔 Would we need to validate later?

conda/exceptions.py
```diff
@@ -1287,7 +1287,16 @@ def __init__(self, username: str, packagename: str, *args, **kwargs):
...
+    pass
+
+
+class SpecNotFound(CondaError):
+    def __init__(self, msg: str, *args, **kwargs):
+        super().__init__(msg, *args, **kwargs)
+
+
+class EnvironmentSpecPluginNotDetected(SpecNotFound):
     def __init__(self, name, plugin_names, *args, **kwargs):
```
jaimergp:
```suggestion
    def __init__(self, name: str, plugin_names: Iterable[str], *args, **kwargs):
```

conda/plugins/manager.py
```diff
@@ -508,39 +508,92 @@ def get_config(self, data) -> PluginConfig:
...
+        if len(found) == 1:
+            if found[0].environment_spec(source).can_handle():
+                return found[0]
+            else:
+                raise PluginError(
+                    f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
+                )
+        elif len(found) > 1:
+            raise PluginError(
+                f"More than one environment_spec plugin named {name} found"
```
jaimergp: Should we report which?
```suggestion
                f"More than one environment_spec plugin named {name} found: {found}"
```

news/14877-option-to-choose-environment-spec-plugin
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add config option and cli argument to allow users to select which environment specifier plugin to use. (#14877)
```
jaimergp:
```suggestion
* Add config option `environment_specifier` and CLI argument (`--environment-specifier, --env-spec`) to allow users to select which environment specifier plugin to use. (#14877)
```

conda/base/context.py
soapy1: I don't think this config field really fits in any of the other categories. Open to suggestions.

I don't think this should go in the `undocumented` section. But maybe could add an "experimental" section?

conda/cli/helpers.py
soapy1: Erm, ya this is not great, I'd say. This happens other places in the code, for example with the `--solver` cli argument https://github.com/conda/conda/blob/main/conda/cli/helpers.py#L423.

>  My understanding is that the context needs the parsed CLI args to finish initialization,

yes, that'll be for slurping up all the config from all the different places it comes from. I don't think this is changing anything for plugin loading though. Like, all the plugins get discovered when the [plugin manager gets initialized with `plugin.manager::get_plugin_manager`](https://github.com/conda/conda/blob/main/conda/plugins/manager.py#L651). 

So, I think setting `choices` is an ok thing to do here. It is weird that configuration and other things that are expected to be globally accessible (like the plugin manager) and shoved into the context object.

conda/base/context.py
jezdez: Oh, I'm in favor of an experimental section!

conda/cli/common.py
jezdez: Thank you!

conda/cli/helpers.py
jezdez: That's a fair point, I've used something like this before, should maybe used for the solver option as well

```python
import argparse

class LazyChoicesAction(argparse.Action):
    def __init__(self, option_strings, dest, choices_func, **kwargs):
        self.choices_func = choices_func
        super().__init__(option_strings, dest, **kwargs)
    
    def __call__(self, parser, namespace, values, option_string=None):
        valid_choices = self.choices_func()
        if values not in valid_choices:
            raise argparse.ArgumentTypeError(
                f"invalid choice: {values!r} (choose from {valid_choices})"
            )
        setattr(namespace, self.dest, values)
```

```suggestion
        action=LazyChoicesAction,
        choices_func=context.plugin_manager.get_environment_specifiers,
```

conda/plugins/manager.py
```diff
@@ -554,6 +603,24 @@ def get_environment_specifiers(self, filename: str) -> CondaEnvironmentSpecifier
             name=filename, plugin_names=[hook.name for hook in hooks]
         )
 
+    def get_environment_specifier(
```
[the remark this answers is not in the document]
jezdez: I like this

jezdez (commented):
The lazy population of the choices is good to fix before merge, other than that this looks great

conda/cli/helpers.py
jaimergp: Oh, I had missed we do that with the solver already. I think we can apply Jannis' suggestion and fix the solver bits in a separate PR to aid with traceability.

conda/base/context.py
jaimergp: Me too, let's move it to experimental.

jaimergp (dismissed):
I'm fine with leaving `choices=` as is for now provided we check the `Lazy*` approach @jezdez suggested in a new PR. Let's move the documentation bits to a new Experimental section and merge here!

soapy1:
Pushed up the category map change. Follow up PR with the `Lazy` approch coming soon

jaimergp (approved):

jezdez (approved):

PR #14888, opened 2025-05-29
Environment spec plugin detection improvements

soapy1:
### Description

This PR suggests some improvements to conda's environment spec detection system. conda expands the environment spec plugin interface to have a more opinionated set of validations for determining an environment spec plugin. So, instead of `can_handle`, a plugin must implement `validate_filename`, `validate_schema`. 

ref: https://docs.google.com/document/d/1LtB2QKuF72FsnuBzWYsw4wOL2w_boyvYys5taiOE2zE/edit?tab=t.0

### Checklist - did you ...

### DIFF
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -419,12 +419,24 @@ class EnvironmentSpecBase(ABC):
     """
 
     @abstractmethod
-    def can_handle(self) -> bool:
+    def validate_source_name(self) -> bool:
         """
-        Determines if the EnvSpec plugin can read and operate on the
-        environment described by the `filename`.
+        Determine if the EnvironmentSpec plugin can parse the target
+        source name. If the source spec is a filename, this check
+        may be used to ensure the file has the right extension and format.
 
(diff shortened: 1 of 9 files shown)


conda/plugins/types.py
```diff
@@ -419,12 +419,24 @@ class EnvironmentSpecBase(ABC):
     """
 
     @abstractmethod
-    def can_handle(self) -> bool:
+    def validate_source_name(self) -> bool:
```
soapy1, line 422: I'm not really a fan of this approach in general. It has all the downsides of the current implementation but no real benefit. I think the argument can be made to say that splitting the types of validations may make it more likely for a plugin author to want to follow a best practice. But it would be better to build tools for plugin authors that force them into making good decisions.

jezdez: I agree with that sentiment

soapy1: cool, in this case I'll close this. It was nice to try it on and see if it fits :jeans:

PR #14898, opened 2025-06-03
Allow aliases to be used in `conda config --set` and others

jaimergp:
### Description

Fixes regression introduced in https://github.com/conda/conda/pull/14678, and reported in https://github.com/conda-incubator/setup-miniconda/issues/401. `conda config --set` does not allow aliases to be used, but IMO it should.

Resolves https://github.com/conda/conda/issues/14899

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -652,7 +652,7 @@ def execute_config(args, parser):
     if args.file and (args.show is not None or args.describe is not None):
         from ..base.context import Context
 
-        context = Context(search_path=(args.file,))
+        context = Context(search_path=(args.file,), argparse_args=args)
 
     stdout_write = getLogger("conda.stdout").info
     stderr_write = getLogger("conda.stderr").info
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -394,6 +394,8 @@ def _key_exists(key: str, warnings: list[str], context=None) -> bool:
         return True
 
     if first not in context.list_parameters():
+        if context._name_for_alias(first):
+            return True
         if context.json:
             warnings.append(f"Unknown key: {key!r}")
         else:
@@ -419,6 +421,10 @@ def _get_key(
     if not _key_exists(key, warnings, context):
         return
 
+    if alias := context._name_for_alias(key):
+        key = alias
+        key_parts = alias.split(".")
+
     sub_config = config
     try:
         for part in key_parts:
@@ -440,6 +446,9 @@ def _set_key(key: str, item: Any, config: dict) -> None:
         from ..exceptions import CondaKeyError
 
         raise CondaKeyError(key, "unknown parameter")
+
+    if alias := context._name_for_alias(key):
+        key = alias
 
     first, *rest = key.split(".")
 
@@ -595,7 +604,7 @@ def _validate_provided_parameters(
     from ..common.io import dashlist
     from ..exceptions import ArgumentError
 
-    all_names = context.list_parameters()
+    all_names = context.list_parameters(aliases=True)
     all_plugin_names = context.plugins.list_parameters()
 
     not_params = set(parameters) - set(all_names)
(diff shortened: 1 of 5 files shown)


conda/cli/main_config.py
```diff
@@ -523,9 +532,15 @@ def _remove_key(key: str, config: dict) -> None:
...
-        # KeyError: part not found, nothing to remove
+        from ..base.context import context
         from ..exceptions import CondaKeyError
 
+        if alias := context._name_for_alias(key):
+            try:
+                return _remove_key(alias, config)
+            except CondaKeyError:
+                pass  # raise with original name
+        # KeyError: part not found, nothing to remove
```
kenodegard: the pattern used elsewhere in the code is to leave these exception annotations at the top of the block

kenodegard:
I think we should include a warning when the key is an alias

More interesting question is if we should use the actual key if an alias is provided (I think most users would want this, some users who use multiple different conda versions would not want this)

jaimergp:
> More interesting question is if we should use the actual key if an alias is provided

I'm forwarding the alias to the canonical key in all cases. I'll warn on `--set` because the effects won't be seen immediately on output (for `--get`, `--show`, `--describe` users will see that a different key name was returned).

---

What I don't fully get is how `conda config --add channels ...` worked before, given that that field is also an alias 🤔 Is there some special logic for lists and dicts?

jezdez:
> > More interesting question is if we should use the actual key if an alias is provided
> 
> I'm forwarding the alias to the canonical key in all cases. I'll warn on `--set` because the effects won't be seen immediately on output (for `--get`, `--show`, `--describe` users will see that a different key name was returned).
> 
> What I don't fully get is how `conda config --add channels ...` worked before, given that that field is also an alias 🤔 Is there some special logic for lists and dicts?

Is `--add` maybe working slightly differently compared to `--set`?

conda/common/configuration.py
```diff
@@ -1479,6 +1485,18 @@ def _set_raw_data(self, raw_data: Mapping[Hashable, dict]):
         self._reset_cache()
         return self
 
+    def _name_for_alias(self, alias: str, ignore_private: bool = True) -> str | None:
```
jezdez: I think that's a cool method, can we make it not private?

conda/common/configuration.py
```diff
@@ -1479,6 +1485,18 @@ def _set_raw_data(self, raw_data: Mapping[Hashable, dict]):
...
+        return next(
+            (
+                p._name
+                for p in self.__class__.__dict__.values()
+                if isinstance(p, ParameterLoader)
+                and alias in p.aliases
+                and (ignore_private and not p._name.startswith("_"))
```
kenodegard: I think we should do a slight refactor in the `ConfigurationType` metaclass and convert the `parameter_names` into a mapping (names to parameters) that we can use here instead:

```suggestion
                for name, parameter in self.parameter_names.items()
                if alias in parameter.aliases
                and (ignore_private and not name.startswith("_"))
```

Also the parameters starting with `_` aren't "private", they simply have a more intricate getter so I'm not sure we need to include the `ignore_private` logic

conda/common/configuration.py
```diff
@@ -1314,6 +1314,21 @@ def __init__(cls, name, bases, attr):
             for name, p in cls.__dict__.items()
             if isinstance(p, ParameterLoader)
         )
+        cls.parameter_names_and_aliases = tuple(
+            name for p in cls.parameter_loaders.values() for name in p._names
+        )
+
+    @property
+    def parameter_loaders(cls):
```
kenodegard: Move this to the `Configuration` class itself so we don't have to do `self.__class__.parameter_loaders`?

jezdez: Ok

jezdez: we need to turn it into a classmethod though then

kenodegard: hm we can leave it as a metaclass property but alias it in the instance object?

```python
class ConfigurationType(type):
    @property
    def _parameter_loaders(cls) -> dict[str, ParameterLoader]:
        return {
            name: param
            for name, param in cls.__dict__.items()
            if isinstance(param, ParameterLoader)
        }

    def __call__(cls, *args, **kwargs):
        self = super().__call__(*args, **kwargs)
        self._parameter_loaders = cls._parameter_loaders
        return self
```

on reflection I think this is getting too far into the weeds and isn't helping with correcting the actual parameter aliasing issue, probably best to revert it to what you originally had?

jezdez (approved):

PR #14901, opened 2025-06-03
Replace `context.conda_exe` with `context.conda_exe_vars_dict`

kenodegard:
### Description

Inspired by work done in https://github.com/conda/conda/pull/12771

Replace all usage of `context.conda_exe` with `context.conda_exe_vars_dict`.

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -96,10 +96,12 @@ class _Activator(metaclass=abc.ABCMeta):
 
     unset_var_tmpl: str
     export_var_tmpl: str
+    path_var_tmpl: str
     set_var_tmpl: str
     run_script_tmpl: str
 
     hook_source_path: Path | None
+    inline_hook_source: bool
 
     def __init__(self, arguments=None):
         self._raw_arguments = arguments
@@ -193,7 +195,10 @@ class _Activator(metaclass=abc.ABCMeta):
         if preamble := self._hook_preamble():
             builder.append(preamble)
         if self.hook_source_path:
-            builder.append(self.hook_source_path.read_text())
+            if self.inline_hook_source:
+                builder.append(self.hook_source_path.read_text())
+            else:
+                builder.append(self.run_script_tmpl % self.hook_source_path)
         if auto_activate is None and context.auto_activate or auto_activate:
             builder.append(f"conda activate '{context.default_activation_env}'\n")
         postamble = self._hook_postamble()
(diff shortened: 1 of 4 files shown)


conda/activate.py
```diff
@@ -890,8 +911,11 @@ class PosixActivator(_Activator):
     tempfile_extension = None  # output to stdout
     command_join = "\n"
 
-    unset_var_tmpl = "unset %s"
+    # Using `unset %s` would cause issues for people running
+    # with shell flag -u set (error on unset).
```
kenodegard, line 915: I don't think we should worry about users running POSIX with `-u` set. Engineering `conda` to work under such a strict mode feels hacky, and our support for it is already inconsistent since we've previously used both `export %s=''` and `unset %s` in POSIX environment activation, making our `-u` compatibility haphazard at best.

beckermr:
Thanks for the PR @kenodegard! LGTM from my point of view.

jezdez (dismissed):
Thanks, that's a nice to review PR!

jezdez:
The failure in `test_csh_basic` seems legit tho

conda/activate.py
```diff
@@ -890,10 +911,13 @@ class PosixActivator(_Activator):
...
+    unset_var_tmpl = "export %s=''"  # unset %s
     export_var_tmpl = "export %s='%s'"
+    path_var_tmpl = "export %s=\"$(cygpath '%s')\"" if on_win else export_var_tmpl
     set_var_tmpl = "%s='%s'"
-    run_script_tmpl = '. "%s"'
+    run_script_tmpl = ". \"`cygpath '%s'`\"" if on_win else '. "%s"'
```
travishathaway, line 920: Pretty sure this line is causing the test errors. Why was`cygpath` added here?

conda/activate.py
kenodegard: @jezdez want me to switch back to `unset` here or do that in a followup PR?

conda/activate.py
kenodegard: Not this line, but the one further down in the `CshActivator` on line 962

You're right the code didn't previously do this. This was added to mirror what was already being done for `export`.

jezdez (approved):

conda/activate.py
kenodegard: https://github.com/conda/conda/issues/14904 for followup

PR #14902, opened 2025-06-03
Replace hard-coded `C:\ProgramData` for `.condarc` search paths

### LINKED ISSUE #14912: Search paths for system-wide `.condarc` files are hard-coded on Windows
### Checklist

### What happened?

The search paths for `.condarc` files on Windows are [hard-coded](https://github.com/conda/conda/blob/0e4ba4e50dc419558a572e52e3409fe6302d477a/conda/base/constants.py#L39-L43) on Windows to be in `C:\ProgramData`.

While it is common for Windows to be installed in the `C:` drive and it is common for `ProgramData` to be on the same drive as the Windows installation, there is no guarantee for this to be true.

It is best practice to use drive-independent variables such as the `%ProgramData%` environment variable.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

marcoesters:
### Description

The search paths for `.condarc` files contain hard-coded paths for `ProgramData`. While that path is correct for most systems, it is possible to redirect that directory to a different drive, e.g., because disk space in the system drive is full or dual boots. A more robust approach is using known folder IDs as implemented by `platformdirs`.

Closes #14912 

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -37,7 +37,7 @@ APP_NAME: Final = "conda"
 
 SEARCH_PATH: tuple[str, ...] = ()
 
-if on_win and (PROGRAMDATA := environ.get("PROGRAMDATA")):  # pragma: no cover
+if on_win and (PROGRAMDATA := environ.get("ProgramData")):  # pragma: no cover
     PROGRAMDATA = PureWindowsPath(PROGRAMDATA).as_posix()
     SEARCH_PATH += (
         f"{PROGRAMDATA}/conda/.condarc",
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -12,7 +12,9 @@ from __future__ import annotations
 
 import struct
 from enum import Enum, EnumMeta
+from os import environ
 from os.path import join
+from pathlib import PureWindowsPath
 from typing import TYPE_CHECKING
 
 from ..common.compat import on_win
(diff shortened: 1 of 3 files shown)


jezdez:
Please file a ticket for this, or find one that fits this. We need to keep track of Windows changes for posterity :)

marcoesters:
> Please file a ticket for this, or find one that fits this. We need to keep track of Windows changes for posterity :)

Happy to do so, but I want to point out that this is currently not listed as a requirement in the contributor guidelines.

jezdez:
> > Please file a ticket for this, or find one that fits this. We need to keep track of Windows changes for posterity :)
> 
> Happy to do so, but I want to point out that this is currently not listed as a requirement in the contributor guidelines.

Thanks for the note, but I didn't say anything about a requirement, you must have missed the part about "bug reports are always welcome" and I wanted to encourage you to take another step to create or find a bug report that matches this proposed change.

As context, I responded like this to highlight that in the case of Windows topics, I know it's important to keep track of it due to the importance to the users (~50% of users are on Windows) and there are our knowledge gaps for Windows topics. E.g. if there is a chance we can close an existing ticket with your code contribution, we should absolutely make sure to do that.

jezdez:
Could we use `platformdirs` for this, since it has an "official" way to get those paths? In the case of `ProgramData` it's `platformdirs.site_data_dir()`.

Check out the implementation for the fallback-based system using `CSIDL_COMMON_APPDATA` as the basis: https://github.com/tox-dev/platformdirs/blob/main/src/platformdirs/windows.py

More info: https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid

marcoesters:
> Thanks for the note, but I didn't say anything about a requirement, you must have missed the part about "bug reports are always welcome" and I wanted to encourage you to take another step to create or find a bug report that matches this proposed change.

I did not find an existing issue, so I created one #14912. I did not miss the "bug reports are always welcome", but did not interpret this as "we prefer you file an issue before creating a PR". I suggest the guidelines be updated if that's what is desired.

> Could we use `platformdirs` for this, since it has an "official" way to get those paths? In the case of `ProgramData` it's `platformdirs.site_data_dir()`.
> 
> Check out the implementation for the fallback-based system using `CSIDL_COMMON_APPDATA` as the basis: https://github.com/tox-dev/platformdirs/blob/main/src/platformdirs/windows.py
> 
> More info: https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid

I have been looking into it. `platformdirs` uses the `CSIDL` system, which Microsoft discourages: https://learn.microsoft.com/en-us/windows/win32/shell/csidl

I will do a deeper dive and see what the implications are of using one over the other. I think if it's all the same, `platformdirs` is the better choice. I will update the description once I've come to a full conclusion before I remove the PR from the draft stage.

marcoesters:
I implemented the `platformdirs` variant. In a virtual machine, I moved `ProgramData` into the `D:` drive and the `.condarc` inside `D:\ProgramData\conda` was picked up by `conda info`. It also has the advantage that I don't have to worry about case-sensitive systems like Cygwin or git bash.

I'm still looking into how to test this - as far as I can see, there are no tests for reading those global files yet.

jezdez:
not stale?

PR #14905, opened 2025-06-04
Replace `get_index` and `get_reduced_index` usage

kenodegard:
### Description

Followup to https://github.com/conda/conda/pull/13880

Stop using deprecated `conda.core.index.get_index` and `conda.core.index.get_reduced_index`.

Extracted from https://github.com/conda/conda/pull/12771

### Checklist - did you ...

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -37,7 +37,7 @@ from ..models.prefix_graph import PrefixGraph
 from ..models.version import VersionOrder
 from ..reporters import get_spinner
 from ..resolve import Resolve
-from .index import _supplement_index_with_system, get_reduced_index
+from .index import Index, ReducedIndex
 from .link import PrefixSetup, UnlinkLinkTransaction
 from .prefix_data import PrefixData
 from .subdir_data import SubdirData
@@ -64,6 +64,9 @@ class Solver:
       * :meth:`solve_for_diff`
       * :meth:`solve_for_transaction`
     """
+
+    _index: ReducedIndex | None
+    _r: Resolve | None
 
     def __init__(
         self,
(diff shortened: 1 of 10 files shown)


kenodegard:
Blocked by #14924

conda/core/solve.py
```diff
@@ -1261,7 +1263,7 @@ def _notify_conda_outdated(self, link_precs):
                     file=sys.stderr,
                 )
 
-    def _prepare(self, prepared_specs):
+    def _prepare(self, prepared_specs) -> tuple[ReducedIndex, Resolve]:
```
jaimergp, line 1266:
```suggestion
    def _prepare(self, prepared_specs: Iterable[MatchSpec]) -> tuple[ReducedIndex, Resolve]:
```

I think? Double check though.

tests/core/test_index.py
```diff
@@ -147,9 +145,15 @@ def patch_pkg_cache(mocker, pkg_cache_entries):
     mocker.patch("conda.base.context.context.track_features", ("test_feature",))
 
 
-def test_supplement_index_with_system():
-    index = {}
-    _supplement_index_with_system(index)
+def test_deprecated_supplement_index_with_system() -> None:
```
jaimergp, line 148: Why are we adding return annotations to tests? 🤔 They never return anything, do they?

kenodegard: At one point `mypy` wouldn't by default type check functions unless either the args or return was annotated, so I added these `None` returns to enable the type checking in my IDE for functions without any args, I've just gotten in the habit of consistently adding them 🤷🏼‍♂️

travishathaway (approved):
I think this is ready to merge. Let's wait for the other assigned reviewers to chime in too.

jaimergp (commented):
Leaving aside I just realized that some tests are annotated with `-> None` return types, and that that's not within the scope of this PR (we should address that conversation separately), the changes proposed here look ok to me.

jaimergp (approved):
Leaving aside I just realized that some tests are annotated with `-> None` return types, and that that's not within the scope of this PR (we should address that conversation separately), the changes proposed here look ok to me.

PR #14911, opened 2025-06-05
Refactor cli layer: use the plugin system to read input files

soapy1:
### Description

This PR refactors the CLI layer to:
* merge all environment specification coming from cli args and environemt specs into a common `model.environment.Environment` object
* creates a transaction object from the final `Environment`

This refactor is a big change to how conda's cli layer. In order to ensure that we are not changing functionality in unexpected ways. [I've outlined all the ways conda package installing related commands work, including noting some bugs](https://docs.google.com/document/d/1Pr0R01he8T3Bg0-YTlnBR9xlwjvG_lOW_UBuSdE60qU/edit?tab=t.0#heading=h.fp7hlxnc6giv) (this includes `conda install`, `conda env create`, `conda remove`, etc.). 

Some important changes in this PR:
*  **`conda install/update` commands can now read environment.yaml files**: Previsouly, only `conda env` commands could read environment.yaml files. This PR introduces using the plugin system to detect the type of the provided `--file` source. So, any plugin installed can be used.

Things not included in this PR (can be included in follow up PRs):
* being able to select which environment spec plugin to use by a cli flag
* updating all install paths to use `handle_txn` to execute the installing transaction

### Checklist - did you ...

- addresses:
  -  https://github.com/conda/conda-environments-project-mgmt/issues/37
  - https://github.com/conda/conda-environments-project-mgmt/issues/15 and https://github.com/conda/conda-environments-project-mgmt/issues/64
- fixes https://github.com/conda/conda-environments-project-mgmt/issues/63


conda/cli/install.py
```diff
@@ -315,6 +318,61 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    files: list[str] = (),
+    inject_default_packages: bool = True,  
+) -> Environment:
+    # First, let's create an 'Environment' for the information exposed in the CLI (no files)
+    if inject_default_packages:
+        names = {MatchSpec(pkg).name for pkg in specs}
+        for pkg in context.create_default_packages:
+            pkg_name = MatchSpec(pkg).name
+            if pkg_name not in names:
+                specs.append(pkg_name)
```
jaimergp:
```suggestion
                specs.append(pkg)
```

In case the user configured version constraints

conda/cli/install.py
```diff
@@ -315,6 +318,61 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    )
+
+    # Now let's process potential files passed via --file
+    file_envs = []
+    if files:
+        for path in files:
+            # parse the file
+            spec_hook = context.plugin_manager.get_environment_specifiers(path)
+            file_env = spec_hook.environment_spec(path).environment
+            file_envs.append(file_env)
```
jaimergp:
```suggestion
    for path in files:
        # parse the file
        spec_hook = context.plugin_manager.get_environment_specifiers(path)
        file_env = spec_hook.environment_spec(path).environment
        file_envs.append(file_env)
```

conda/cli/install.py
```diff
@@ -334,71 +392,49 @@ def install_clone(args, parser):
...
-            return
-    specs.extend(common.specs_from_args(args_packages, json=context.json))
+    if context.use_only_tar_bz2:
+        args.repodata_fns = ("repodata.json",)
+
+    env = _assemble_environment(
+        name=args.name,
+        prefix=context.target_prefix,
+        specs=[s.strip("\"'") for s in args.packages],
+        files=args.file,
```
jaimergp, line 418:
```suggestion
        files=(args.file,),
```

conda/cli/install.py
```diff
@@ -315,6 +318,61 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+            # parse the file
+            spec_hook = context.plugin_manager.get_environment_specifiers(path)
+            file_env = spec_hook.environment_spec(path).environment
+            file_envs.append(file_env)
+
+    if context.dry_run:
+        cli_env.prefix = os.path.join(mktemp(), UNUSED_ENV_NAME)
+        return Environment.merge(cli_env, *file_envs)
+    else:
+        return Environment.merge(cli_env, *file_envs)
```
jaimergp: If any of the files specifies an environment name or prefix, will be overridden, or does the first one win?

conda/cli/install.py
```diff
@@ -315,6 +318,61 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
             raise PackageNotInstalledError(prefix, spec.name)
 
 
+def _assemble_environment(
+    name: str | None = None,
+    prefix: str | None = None,
+    specs: list[str] = (),
+    files: list[str] = (),
```
jaimergp:
```suggestion
    specs: Iterable[str] = (),
    files: Iterable[str] = (),
```

conda/cli/install.py
```diff
@@ -24,6 +24,7 @@
     UpdateModifier,
 )
 from ..base.context import context
+from ..base.constants import  UNUSED_ENV_NAME
```
jaimergp:
```suggestion
from ..base.constants import UNUSED_ENV_NAME
```

jaimergp (commented):
Loving how this is starting to look!

conda/cli/install.py
soapy1: currently, the implementation in the Environment model is so that the first one wins. And, it will log if there are several prefixes specified. 
https://github.com/conda/conda/blob/main/conda/models/environment.py#L240

conda/cli/install.py
```diff
@@ -315,82 +327,64 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    files: Iterable[str] = (),
+    inject_default_packages: bool = True,
+) -> Environment:
+    # First, let's create an 'Environment' for the information exposed in the CLI (no files)
+    if inject_default_packages:
+        names = {MatchSpec(pkg).name for pkg in specs}
+        for pkg in context.create_default_packages:
+            pkg_name = MatchSpec(pkg).name
+            if pkg_name not in names:
+                specs.append(pkg)
```
kenodegard, line 343: Per the type annotation `specs` may not necessarily be a list

So what if we do this instead (since the ultimate goal is to add missing default packages):


```suggestion
        # include default packages but override with specs of the same name
        specs = {
            MatchSpec(pkg).name: pkg
            for pkg in (*context.create_default_packages, *specs)
        }.values()
```

soapy1:
pre-commit.ci autofix

conda/cli/common.py
```diff
@@ -134,11 +141,25 @@ def arg2spec(arg, json=False, update=False):
...
+    :param json: unused
+    :param update: whether the command being run is an update command
+    :returns: str
+    """
+    spec = arg2matchspec(arg, update=update)
     return str(spec)
 
 
-def specs_from_args(args, json=False):
+def specs_from_args(args: list[str], json: bool = False):
```
jaimergp:
```suggestion
def specs_from_args(args: Iterable[str], json: bool = False) -> list[str]:
```

conda/cli/install.py
```diff
@@ -315,82 +327,64 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    for spec in specs:
+        if is_package_file(spec):
+            fetch_explicit_packages.append(spec)
+        else:
+            requested_packages.append(common.arg2matchspec(spec))
+
+    # transform explicit packages into package records
+    explicit_packages = []
+    if fetch_explicit_packages:
+        explicit_packages = get_package_records_from_explicit(fetch_explicit_packages)
```
jaimergp: This function has a potential side-effect of pre-downloading artifacts, only to build a transaction that is then presented to the user to be accepted or rejected with a prompt.

Is there a chance we can create `PackageRecord` shims with just the info provided in the URL (should be enough: channel, subdir, package name, version, build) so we can assemble a good enough Transaction to ask for confirmation? Then, once the user has chosen "yes" in the prompt, we do perform the downloads and create the actual transaction object from the Package Cache info.

conda/cli/install.py
```diff
@@ -315,82 +327,64 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
 
-    # common validations for all types of installs
-    validate_install_command(prefix=prefix, command=command)
+    return Environment.merge(cli_env, *file_envs)
 
-    if context.use_only_tar_bz2:
-        args.repodata_fns = ("repodata.json",)
 
+def install(args, parser, command="install"):
+    """Logic for `conda install`, `conda update`, `conda remove`, and `conda create`."""
```
jaimergp: I'm not sure we are using this for `conda remove` after all. Can you double check and, in that case, update the docstring?

```suggestion
    """Logic for `conda install`, `conda update`, and `conda create`."""
```

soapy1: Yep, `conda remove` does it's own thing

conda/cli/install.py
soapy1: I like this idea, will give it a try!

soapy1: Hmmm, ok so not sure if this is going to work out. In particular, it's not always possible to get the channel or subdir from a explicit package url. For example, if it is a path to a local file, or a package sitting on a fileserver. So, the transaction object that is created without this information is not quite right. 

This works out on the main branch of conda, since there is no confirmation when installing an explicit package. If the user uses the `--dry-run` flag to install an explicit package, basically nothing happens.

A dry run
```
 $ conda install ~/miniconda3/pkgs/flask-2.2.5-pyhd8ed1ab_0.conda  --dry-run


DryRunExit: Dry run. Exiting.
```

Installing an explicit package:
```
$ conda install ~/miniconda3/pkgs/flask-2.2.5-pyhd8ed1ab_0.conda                       

## Package Plan ##

  environment location: /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/ndf

  added / updated specs:
    - <unknown>/noarch::flask==2.2.5=pyhd8ed1ab_0


  flask              <unknown>/noarch::flask-2.2.5-pyhd8ed1ab_0 


Preparing transaction: done
Verifying transaction: done
Executing transaction: done

```

I think this is all a bit awkward, but I don't think this PR is changing this behaviour in particular.

jaimergp: > For example, if it is a path to a local file, or a package sitting on a fileserver.

In those cases you can extract directly, I think? I mean, if we don't have a URL because it's a relative path, then we have access to the tarball. And if we have a URL, we have a good-enough channel. The full information will be fetched later anyway.

If you prefer not to tackle this now, that's ok, but let's create an issue and leave a note in the code with a comment.

In terms of the awkward UI, maybe we can do something like:

```python
with spinner("Pre-fetching records"...), redirect_stdout(os.devnull):
   # fetch records here
```

This way we don't this odd double "Downloading and Extracting packages" bit.

soapy1: Created an issue for this https://github.com/conda/conda-environments-project-mgmt/issues/76

soapy1:
pre-commit.ci autofix

conda/cli/install.py
```diff
@@ -315,82 +327,71 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
-    validate_install_command(prefix=prefix, command=command)
+    return Environment.merge(cli_env, *file_envs)
 
-    if context.use_only_tar_bz2:
-        args.repodata_fns = ("repodata.json",)
 
+def install(args, parser, command="install"):
+    """Logic for `conda install`, `conda update`, and `conda create`."""
     newenv = command == "create"
     isupdate = command == "update"
```
soapy1, line 392: Updating is a little bit different from creating an environment or installing packages into an environment. For example, it doesn't really make sense to be using environment spec plugins to be updating packages. Might need to rework this some.

soapy1: demo for an alternative https://github.com/conda/conda/compare/main...soapy1:conda:refactor-cli-layer-split-out-update?expand=1

conda/cli/install.py
```diff
@@ -315,82 +327,71 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
 
-def install(args, parser, command="install"):
-    """Logic for `conda install`, `conda update`, `conda remove`, and `conda create`."""
-    prefix = context.target_prefix
+    if context.dry_run:
+        cli_env.prefix = os.path.join(mktemp(), UNUSED_ENV_NAME)
 
-    # common validations for all types of installs
-    validate_install_command(prefix=prefix, command=command)
+    return Environment.merge(cli_env, *file_envs)
```
soapy1, line 386: currently, this environment does not actually represent the final state of the environment to install. It is made up of information coming from the cli and from any specified environment spec files. The specs included in this environment model are just the specs to be added and do not include the specs that are already installed in the environment.

soapy1:
pre-commit.ci autofix

conda/cli/install.py
soapy1: a kinda related ux problem is what the meaning of `conda install --file <environemnt_spec>` is. Like, should this command be removing packages that are not included in the environment spec? `conda env update` has a `--prune` flag that allows for this. Should `conda install` also support this?

jaimergp: > currently, this environment does not actually represent the final state of the environment to install. It is made up of information coming from the cli and from any specified environment spec files. The specs included in this environment model are just the specs to be added and do not include the specs that are already installed in the environment.

Correct, the CLI doesn't know (and never knew) too much about the target environment. That's handled by the solver (via `history`, `pinned` and so on). We can consider this `Environment` class a `EnvironmentRequest` so to say. Adding a comment reporting this nuance would be useful, I think!

> conda env update has a --prune flag that allows for this. Should conda install also support this?

The `solver` layer supports `prune` for all operations; the CLI flag was only exposed to `conda env` IIRC. I think it would be sensible to allow `--prune` in `conda install` too.

conda/cli/install.py
```diff
@@ -379,18 +464,57 @@ def install(args, parser, command="install"):
...
+        retry_unfrozen=should_retry_unfrozen,
+        cmd="update",
+    )
+    handle_txn(unlink_link_transaction, prefix, args, newenv=False)
+
+
+def install(args, parser, command="install"):
+    """Logic for `conda install`, and `conda create`."""
+    newenv = command == "create"
+    isupdate = command == "update"
```
jaimergp, line 392: Do we still need `isupdate` if `install` only handles `create` and `install`?

conda/cli/common.py
```diff
@@ -134,11 +142,25 @@ def arg2spec(arg, json=False, update=False):
...
 
+
+def arg2spec(arg: str, json: bool = False, update: bool = False) -> str:
+    """
+    Ensure that an input package argument can be converted into
+    a MatchSpec. Will raise an error if the package can not be converted
+    or is malformed.
+
+    :param arg: the package provided as an argument on the cli
+    :param json: unused
```
jaimergp: Shall we mark as deprecated?

conda/cli/common.py
```diff
@@ -134,11 +142,25 @@ def arg2spec(arg, json=False, update=False):
...
+    :param update: whether the command being run is an update command
+    :returns: str
+    """
+    spec = arg2matchspec(arg, update=update)
     return str(spec)
 
 
-def specs_from_args(args, json=False):
+def specs_from_args(args: Iterable[str], json: bool = False) -> list[str]:
```
jaimergp: Deprecate `json`?

jaimergp:
@soapy1, I was playing with this PR locally and I think we need some more tests because I quickly found one crash (easy fix, see https://github.com/conda/conda/pull/15026). My suggestion is to bring over the parametrization fixtures introduced in https://github.com/conda/conda/pull/14113/files#diff-b4cf2ce772c22eededba8ee6d0c5c34d165f58bd22cfb143ba5233ccb5b5ec70R35-R44.

conda/cli/install.py
```diff
@@ -379,18 +464,57 @@ def install(args, parser, command="install"):
...
                 )
         if "@EXPLICIT" in specs:
             # short circuit to installing explicit if explicit specs are provided
-            explicit(specs, prefix, verbose=not context.quiet)
-            return
+            raise CondaError(
+                "Cannot update packages from an explicit file. Please try"
+                "\n`conda install` instead"
+            )
     specs.extend(common.specs_from_args(args_packages, json=context.json))
```
jaimergp: Looks like `conda update` does not support the plugin system. Is this intentional? I tried passing some environment.yml files and it crashes with a not-so-informative message:

```
$ conda update --file tests/env/support/with-pip.yml 

CondaValueError: could not parse 'name: pip' in: tests/env/support/with-pip.yml

$ conda update --file tests/env/support/empty_env.yml 

CondaValueError: could not parse '- conda-forge' in: tests/env/support/empty_env.yml
```

I think we should still support the plugin system in `conda update` but only allow name-only specs. That also excludes the `pip` section, I guess.

jaimergp:
`ensure_update_specs_exist()` should accumulate all invalid specs and report them together all at once. The message should also say _why_ they are not valid (`conda update` only accepts name-only specs that are installed).

jaimergp:
Also, there's a surprising amount of [results](https://github.com/search?q=%22conda+update+--file%22+NOT+is%3Afork&ref=opensearch&type=code) for `conda update --file` 🤔 I don't think many of those are working commands, though; we didn't take YML files as input there. Maybe they all meant `conda env update`?

I was searching for that because I was questioning the usefulness of having `conda update --file` when `conda update` is mostly concerned with CLI specs (same as `conda remove`, which does _not_ accept a `--file` flag). But I guess that ship has sailed so we have to keep supporting `--file` for a while. If anything we can start questioning whether to deprecate `conda update --file` because we could argue that most of those uses are tackled by `conda install` anyway.

conda/cli/install.py
soapy1: Ya, I was testing out what the cli might feel like if the `update` command was not using the the plugin system to address some of the questions brought up in [the cli notes doc](https://docs.google.com/document/d/1Pr0R01he8T3Bg0-YTlnBR9xlwjvG_lOW_UBuSdE60qU/edit?tab=t.0#heading=h.mfa42jk1ewst). You're right, this is still quite awkward. Will give it another crack with the plugin system

conda/cli/install.py
soapy1: yes! but, recombining these so, will keep this around.

conda/cli/common.py
soapy1: let's do this in another small PR so this one stays focused on getting the environment object in right.

soapy1: https://github.com/conda/conda/pull/15028

conda/cli/common.py
kenodegard, line 1: revert this file?

conda/cli/install.py
```diff
@@ -309,88 +321,77 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
         spec = MatchSpec(spec)
         if not spec.is_name_only_spec:
             raise CondaError(
-                f"Invalid spec for 'conda update': {spec}\nUse 'conda install' instead."
+                f"'conda update' only supports name-only spec, but received: {spec}\nUse 'conda install' to install a specific version of a package."
```
kenodegard, line 324:
```suggestion
                f"'conda update' only supports name-only spec, but received: {spec}\n"
                f"Use 'conda install' to install a specific version of a package."
```

tests/models/test_environment.py
```diff
@@ -238,6 +238,13 @@ def test_merge_configs_primitive_none_values_order():
     result = EnvironmentConfig.merge(config1, config2)
     assert result.use_only_tar_bz2 is True
 
+
+def test_merge_configs_primitive_none_values_order_two():
+    config1 = EnvironmentConfig(
+        use_only_tar_bz2=True,
+    )
```
kenodegard, line 246:
```suggestion
    config1 = EnvironmentConfig(use_only_tar_bz2=True)
```

tests/test_create.py
```diff
@@ -520,7 +520,12 @@ def test_conda_update_package_not_installed(
             raises=PackageNotInstalledError,
         )
 
-        with pytest.raises(CondaError, match="Invalid spec for 'conda update'"):
+
+def test_conda_update_package_is_not_name_only_spec(
+    tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture
```
kenodegard, line 525:
```suggestion
    tmp_env: TmpEnvFixture,
    conda_cli: CondaCLIFixture,
```

soapy1:
pre-commit.ci autofix

soapy1:
This PR is getting a little bit confusing to manage with the refactor to use the environment model + the reconciling of the `--file` backends. Going to break this up some. 
* refactr to use the environemnt model: https://github.com/conda/conda/pull/15039

soapy1:
This is superseded by https://github.com/conda/conda/pull/15547

PR #14913, opened 2025-06-05
Introduce dataclass to represent an environment config

soapy1:
### Description

This PR suggests adding some rigidity to the Environment's model notion of config. Previously, Environment config was just stored as a plain dict. This is ok, but leaves some abiguity around what type of data ought to present in the config.

Providing this dataclass will give plugin authors a more concrete understanding of:
* the limits of what configuration they can control, and
* what information they _should_ be passing into conda

Further, providing this dataclass will give downstream consumers of the Environment model some assurances about what environment configuratio information should be available to them. For example, consider how the Environment model will be used to generate a set of transactions for conda to execute for package installation/update/removal (https://github.com/conda/conda/pull/14911).

This PR provides the ability to merge multiple environment configs together. It is important to note that this simple approach to merging config is not a substitute for the merging of config sources that happens in the context object. 

### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -36,39 +36,62 @@ class EnvironmentConfig:
     Data model for a conda environment config.
     """
 
-    aggressive_update_packages: bool
+    aggressive_update_packages: bool | None = None
 
-    channel_priority:  ChannelPriority
+    channel_priority:  ChannelPriority | None = None
 
-    channels: list[str]
+    channels: list[str]  = field(default_factory=list)
 
-    channel_settings: dict[str, str]
+    channel_settings: dict[str, str] = field(default_factory=dict)
 
-    deps_modifier: DepsModifier
+    deps_modifier: DepsModifier | None = None
 
-    disallowed_packages: list[str]
+    disallowed_packages: list[str] = field(default_factory=list)
 
-    pinned_packages: list[str]
+    pinned_packages: list[str] = field(default_factory=list)
 
-    repodata_fns: list[str]
+    repodata_fns: list[str] = field(default_factory=list)
 
     sat_solver: SatSolverChoice | None = None
 
     solver: str | None = None
 
-    track_features: list[str]
+    track_features: list[str] = field(default_factory=list)
 
-    update_modifier: UpdateModifier
+    update_modifier: UpdateModifier | None = None
 
-    use_only_tar_bz2: bool
+    use_only_tar_bz2: bool | None = None
 
-    @classmethod
-    def merge(cls, *config):
+    def merge(self, other):
         """
         **Experimental** While experimental, expect both major and minor changes across minor releases.
 
-        Merges multiple EnvironemntConfigs
+        Merges an EnvironmentConfig into this one. Merging rules are:
+        * Primitive types get clobbered
+        * Lists get appended to
+        * Dicts get updated
         """
+        # Return early if there is nothing to merge
+        if other is None:
+            return self
+
+        # Ensure that we are merging another EnvironmentConfig
+        if not isinstance(other, self.__class__):
+            raise CondaValueError("Cannot merge EnvironmentConfig with non-EnvironmentConfig")
+
+        self.aggressive_update_packages = other.aggressive_update_packages
+        self.channel_priority = other.channel_priority
+        self.channels.extend(other.channels)
+        self.channel_settings.update(other.channel_settings)
+        self.deps_modifier = other.deps_modifier
+        self.disallowed_packages.extend(other.disallowed_packages)
+        self.pinned_packages.extend(other.pinned_packages)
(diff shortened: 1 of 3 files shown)


soapy1:
pre-commit.ci autofix

conda/models/environment.py
```diff
@@ -8,19 +8,91 @@
 from logging import getLogger
 from typing import TYPE_CHECKING
 
-from ..base.constants import PLATFORMS
+from ..base.constants import (
+    PLATFORMS,
+)
```
jaimergp: I think this can be reverted.

conda/models/environment.py
```diff
@@ -8,19 +8,91 @@
...
+
+    solver: SatSolverChoice | None = None
+
+    track_features: list[str] = field(default_factory=list)
+
+    update_modifier: UpdateModifier | None = None
+
+    use_only_tar_bz2: bool | None = None
+
+    def merge(self, other):
```
jaimergp:
```suggestion
    @classmethod 
    def merge(cls, *other: EnvironmentConfig) -> EnvironmentConfig:
```

Don't we usually set these as `classmethod`s? At least in other APIs like `MatchSpec`, there's a precedent.

If you prefer a regular method, should we accept several `other`?

conda/models/environment.py
```diff
@@ -8,19 +8,91 @@
...
+
+        # Ensure that we are merging another EnvironmentConfig
+        if not isinstance(other, self.__class__):
+            raise CondaValueError(
+                "Cannot merge EnvironmentConfig with non-EnvironmentConfig"
+            )
+
+        self.aggressive_update_packages = other.aggressive_update_packages
+        self.channel_priority = other.channel_priority
+        self.channels.extend(other.channels)
```
jaimergp: Should we deduplicate here?

conda/models/environment.py
```diff
@@ -8,19 +8,91 @@
...
+        self.channel_settings.update(other.channel_settings)
+        self.deps_modifier = other.deps_modifier
+        self.disallowed_packages.extend(other.disallowed_packages)
+        self.pinned_packages.extend(other.pinned_packages)
+        self.repodata_fns.extend(other.repodata_fns)
+        self.solver = other.solver
+        self.track_features.extend(other.track_features)
```
jaimergp: Should we deduplicate these lists?

news/14913-environment-model-config
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Introduce dataclass to represent an environment config. (#14913)
```
jaimergp:
```suggestion
* Introduce new dataclass `conda.models.environment.EnvironmentConfig` to represent `Environment` settings, a subset of `context` settings that impact environment creation and management. (#14913)
```

conda/models/environment.py
```diff
@@ -8,19 +8,91 @@
...
+
+    deps_modifier: DepsModifier | None = None
+
+    disallowed_packages: list[str] = field(default_factory=list)
+
+    pinned_packages: list[str] = field(default_factory=list)
+
+    repodata_fns: list[str] = field(default_factory=list)
+
+    solver: SatSolverChoice | None = None
```
jaimergp: Is this `SatSolverChoice`? I think this is incorrect; this is supposed to be the solver plugin, which cannot be known statically, so you'll need to set this to `str`.
```suggestion
    solver: str | None = None
```

jaimergp (commented):
Looking good! I have some comments about the `.merge()` method and list deduplication, nothing major. Let's address those.

conda/models/environment.py
soapy1: ah ya `sat_solver` and `solver` are 2 different settings and both should be represented, like 
```
sat_solver: SatSolverChoice | None = None
solver: str | None = None
```

soapy1:
pre-commit.ci autofix

conda/models/environment.py
```diff
@@ -12,15 +12,144 @@
...
+        return self
+
+    @classmethod
+    def merge(cls, *configs: EnvironmentConfig) -> EnvironmentConfig:
+        """
+        **Experimental** While experimental, expect both major and minor changes across minor releases.
+
+        Merges a list of EnvironmentConfigs into a single one. Merging rules are:
+        * Primitive types get clobbered if subsequent configs have a value, otherwise keep the last set value
+        * Lists get appended to
```
jaimergp:
```suggestion
        * Lists get appended to and deduplicated
```

jaimergp (dismissed):

jaimergp (approved):

conda/models/environment.py
soapy1: good catch, added this and a few other docstring corrections

jezdez (approved):

PR #14914, opened 2025-06-05
Allow env spec plugins to enable/disable autodetection

soapy1:
### Description

This PR introduces a way for environment spec plugins to declare that they should not be automatically detected. So, these plugins would only be accessible to users if they explicitly request the plugin by cli arg or configuration.

For example, consider the [environment.yml spec plugin](https://github.com/conda/conda/blob/main/conda/env/specs/yaml_file.py). One could implement a new plugin based on it, with autodetection for their new plugin disabled:
```
class YamlNoAutoDetect(YamlFileSpec):
   detection_supported = False

@hookimpl
        name="secret-environment.yml",
        environment_spec=YamlNoAutoDetect,
    )
```

This plugin will never be selected if not explicitly requested. In order to use the `YamlNoAutoDetect` plugin, users must run a command like:

```
$ conda env create --file special.yml --env-spec secret-environment.yml
```

When no plugin is detected, conda will produce an error calling out plugins with autodetection disabled:

```
$ conda env create --file tests/env/support/simple.yml             

EnvironmentSpecPluginNotDetected: Environment at tests/env/support/simple.yml is not readable by any installed environment specifier plugins.

Available plugins: 
    - binstar
    - requirements.txt

Found compatible plugins but they must be explicitly selected.
Request conda to use these plugins by providing
the cli argument `--environment-spec PLUGIN_NAME`:

    - secret-environment.yml
 ```

depends on the changes to the plugin manager introduced in https://github.com/conda/conda/pull/14877

### Checklist - did you ...

### DIFF
--- a/conda/env/specs/binstar.py
+++ b/conda/env/specs/binstar.py
@@ -24,7 +24,6 @@ deprecated.module("24.7", "25.9")
 deprecated.constant("24.7", "25.9", "ENVIRONMENT_TYPE", "env")
 
 
-@deprecated("24.7", "25.9")
 class BinstarSpec(EnvironmentSpecBase):
     """
     spec = BinstarSpec('darth/deathstar')
@@ -36,6 +35,7 @@ class BinstarSpec(EnvironmentSpecBase):
 
     msg = None
 
+    @deprecated("24.7", "25.9")
     def __init__(self, name=None):
         self.name = name
 
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1301,7 +1301,7 @@ class EnvironmentSpecPluginNotDetected(SpecNotFound):
         self,
         name: str,
         plugin_names: Iterable[str],
-        autodetect_disabled_plugins: Iterable[str] = [],
+        autodetect_disabled_plugins: Iterable[str] = (),
         *args,
         **kwargs,
     ):
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1310,11 +1310,18 @@ class EnvironmentSpecPluginNotDetected(SpecNotFound):
             f"""
             Environment at {name} is not readable by any installed environment specifier plugins.
 
-Available plugins: {dashlist(plugin_names, 4)}
+            Available plugins: {dashlist(plugin_names, 16)}
+
             """
         )
         if len(autodetect_disabled_plugins) > 0:
-            msg += f"\nAvailable plugins with autodetection disabled. Request conda to use these plugins by providing the cli argument `--environment-spec PLUGIN_NAME`: {dashlist(autodetect_disabled_plugins, 4)}"
+            msg += dals(
+                f"""
+                Found compatible plugins but they must be explicitly selected.
+                Request conda to use these plugins by providing
+                the cli argument `--environment-spec PLUGIN_NAME`:
+                """
+            ) + dashlist(autodetect_disabled_plugins, 4)
         super().__init__(msg, *args, **kwargs)
 
 
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1297,16 +1297,18 @@ class SpecNotFound(CondaError):
 
 
 class EnvironmentSpecPluginNotDetected(SpecNotFound):
-    def __init__(self, name: str, plugin_names: Iterable[str], *args, **kwargs):
+    def __init__(
+            self, name: str, plugin_names: Iterable[str], autodetect_disabled_plugins: Iterable[str] = [], *args, **kwargs):
         self.name = name
         msg = dals(
             f"""
-            Environment at {name} is not readable by any installed
-            environment specifier plugins.
+            Environment at {name} is not readable by any installed environment specifier plugins.
 
-            Available plugins: {dashlist(plugin_names, 4)}
+Available plugins: {dashlist(plugin_names, 4)}
             """
         )
+        if len(autodetect_disabled_plugins) > 0:
+            msg += f"\nAvailable plugins with autodetection disabled. Request conda to use these plugins by providing the cli argument `--environment-spec PLUGIN_NAME`: {dashlist(autodetect_disabled_plugins, 4)}"
         super().__init__(msg, *args, **kwargs)
 
 
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1298,7 +1298,13 @@ class SpecNotFound(CondaError):
 
 class EnvironmentSpecPluginNotDetected(SpecNotFound):
     def __init__(
-            self, name: str, plugin_names: Iterable[str], autodetect_disabled_plugins: Iterable[str] = [], *args, **kwargs):
+        self,
+        name: str,
+        plugin_names: Iterable[str],
+        autodetect_disabled_plugins: Iterable[str] = [],
+        *args,
+        **kwargs,
+    ):
         self.name = name
         msg = dals(
             f"""
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1316,7 +1316,7 @@ class EnvironmentSpecPluginNotDetected(SpecNotFound):
         )
         if len(autodetect_disabled_plugins) > 0:
             msg += dals(
-                f"""
+                """
                 Found compatible plugins but they must be explicitly selected.
                 Request conda to use these plugins by providing
                 the cli argument `--environment-spec PLUGIN_NAME`:
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -565,24 +565,23 @@ class CondaPluginManager(pluggy.PluginManager):
         hooks = self.get_hook_results("environment_specifiers")
         found = []
         for hook in hooks:
-            if hook.enable_autodetection:
-                log.debug("EnvironmentSpec hook: checking %s", hook.name)
-                if hook.environment_spec(source).can_handle():
-                    log.debug(
-                        "EnvironmentSpec hook: %s can be %s",
-                        source,
-                        hook.name,
-                    )
-                    found.append(hook)
-                else:
-                    log.debug(
-                        "EnvironmentSpec hook: %s can NOT be handled by %s",
-                        source,
-                        hook.name,
-                    )
-            else:
+            log.debug("EnvironmentSpec hook: checking %s", hook.name)
+            if not hook.environment_spec.detection_supported():
                 log.debug(
                     "EnvironmentSpec hook '%s' does not support autodetection, skipping",
+                    hook.name,
+                )
+            elif hook.environment_spec(source).can_handle():
+                log.debug(
+                    "EnvironmentSpec hook: %s can be %s",
+                    source,
+                    hook.name,
+                )
+                found.append(hook)
+            else:
+                log.debug(
+                    "EnvironmentSpec hook: %s can NOT be handled by %s",
+                    source,
                     hook.name,
                 )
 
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by jezdez** [comment]
pre-commit.ci autofix

File: conda/plugins/manager.py [hunk 1]
```diff
@@ -508,39 +508,98 @@ def get_config(self, data) -> PluginConfig:
...
+                        source,
+                        hook.name,
+                    )
+                    found.append(hook)
+                else:
+                    log.debug(
+                        "EnvironmentSpec hook: %s can NOT be handled by %s",
+                        source,
+                        hook.name,
+                    )
```
jezdez: @jaimergp @soapy1 I wonder if this `enable_autodetection` option is yet another predicate about whether a plugin "can handle", since the same information could simply be implemented on the env spec level instead of the pluggy interface level. I'd prefer the env spec level to all the env spec business logic covered there

soapy1: I would say that autodetection and the ability for a plugin to parse an environment spec are seperate ideas. Both of these ideas need to come together eventually for conda to decide if an environment spec plugin is going to be used. However, `can_handle` is about deteriming if a plugin is able to read/parse a given environment spec. And `enable_autodetection` is about a plugin author being able to opt out of the plugin autodetection process.

imo the whole auto detection process is the responsibility of the plugin manager. So, I think the most appropriate place for a plugin author to be able to set this information is in the bit where the plugin get's registered. 

I'm definitely open to suggestions here. But, I would strongly recommend that the mechanism for opting in-to or out-of autodetection is very explicit. Like, it would be confusing if opting out of autodetection was done by having `can_handle` return a `None` value opposed to a `True`/`False` value.

conda/plugins/manager.py
```diff
@@ -565,18 +565,24 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
...
             else:
                 log.debug(
-                    "EnvironmentSpec hook: %s can NOT be handled by %s",
-                    source,
+                    "EnvironmentSpec hook '%s' does not support autodetection, skipping",
                     hook.name,
                 )
```
kenodegard: I'd do the following instead:

```python
for hook in hooks:
    log.debug(...checking...)
    if not hook.enable_autodetection:
        log.debug(...skipping...)
    elif hook.environment_spec(source).can_handle():
        log.debug(...maybe...)
        found.append(hook)
    else:
        log.debug(...no...)
```

conda/plugins/manager.py
jezdez: I don't agree with your hypothesis that plugin authors should be allowed to create plugins that have two states, auto-detectable or not, I'd say those are fundamentally different behaviors, and it's visible by the error message below, which doesn't give users any recourse to resolve the problem.

In other words, I don't agree that this boolean should be stored in the plugin type, since it _is_ business logic if a plugin should be used to automatically detect the environment spec from a provided file. The deciding code would use the same functionality as we'd implement in the EnvironmentSpecBase, rather than what is in the plugin manager, which doesn't care about the implementation details of environment specs at all.

I'd suggest we either move this into the main `EnvironmentSpecBase` and do something like a `detection_supported()`, an instance method that can reuse information from the instance:

```suggestion
            env_spec = hook.environment_spec(source)
            if env_spec.detection_supported():
                log.debug("EnvironmentSpec hook: checking %s", hook.name)
                if env_spec.can_handle():
                    log.debug(
                        "EnvironmentSpec hook: %s can be %s",
                        source,
                        hook.name,
                    )
                    found.append(hook)
                else:
                    log.debug(
                        "EnvironmentSpec hook: %s can NOT be handled by %s",
                        source,
                        hook.name,
                    )
```

or we add another plugin type like `class OptionalEnvironmentSpec(EnvironmentSpecBase): ...`, just for the purpose of providing an optional environment spec, plugin authors would be able to able to switch the behavior by using this different type.

soapy1: Lets go for adding a `detection_supported` function opposed to a new type. Pushed up the change.

soapy1:
pre-commit.ci autofix

conda/exceptions.py
```diff
@@ -1297,16 +1297,24 @@ def __init__(self, msg: str, *args, **kwargs):
...
-            Environment at {name} is not readable by any installed
-            environment specifier plugins.
+            Environment at {name} is not readable by any installed environment specifier plugins.
 
-            Available plugins: {dashlist(plugin_names, 4)}
+Available plugins: {dashlist(plugin_names, 4)}
             """
         )
+        if len(autodetect_disabled_plugins) > 0:
+            msg += f"\nAvailable plugins with autodetection disabled. Request conda to use these plugins by providing the cli argument `--environment-spec PLUGIN_NAME`: {dashlist(autodetect_disabled_plugins, 4)}"
```
soapy1: This can probably be worded better

conda/exceptions.py
```diff
@@ -1297,16 +1297,24 @@ def __init__(self, msg: str, *args, **kwargs):
...
+    ):
         self.name = name
         msg = dals(
             f"""
-            Environment at {name} is not readable by any installed
-            environment specifier plugins.
+            Environment at {name} is not readable by any installed environment specifier plugins.
 
-            Available plugins: {dashlist(plugin_names, 4)}
+Available plugins: {dashlist(plugin_names, 4)}
```
jaimergp, line 1313:
```suggestion
            Available plugins: {dashlist(plugin_names, 4)}
```

conda/exceptions.py
jaimergp:
```suggestion
            msg += dals(
                f"""
                Found compatible plugins but they must be explicitly selected.
                Request conda to use these plugins by providing
                the cli argument `--environment-spec PLUGIN_NAME`:
                """
            ) + dashlist(autodetect_disabled_plugins, 4)}
```

conda/exceptions.py
jaimergp: What's with the indentation here?

conda/plugins/types.py
```diff
@@ -420,6 +420,21 @@ class EnvironmentSpecBase(ABC):
     Base class for all env specs.
     """
 
+    @classmethod
```
jaimergp: Should this be a class method or a regular instance method? As a class method, it doesn't have much instance context to access so might as well make it a class attribute.

conda/exceptions.py
soapy1: Ya, this is weird. Without this change the formatting is not nice. Before this change the output is  :

```
EnvironmentSpecPluginNotDetected: Environment at tests/env/support/simple.yml is not readable by any installed environment specifier plugins.

    Available plugins: 
- binstar
- requirements.txt
```

soapy1: oh, I see actually this can be fixed a better way by setting the right indentation on `dashlist`

soapy1:
pre-commit.ci autofix

conda/plugins/types.py
soapy1: Ya, a class attribute sounds good to me. I think the determining if a plugin has autodetect enabled should be done without having to instantiate the object. This is just a setting that a plugin author can set as they see fit. It should not have any extra logic associated with it.

conda/env/specs/binstar.py
```diff
@@ -36,6 +35,7 @@ class BinstarSpec(EnvironmentSpecBase):
 
     msg = None
 
+    @deprecated("24.7", "25.9")
```
jaimergp, line 38: Why this change? 🤔

conda/exceptions.py
```diff
@@ -1297,16 +1297,31 @@ def __init__(self, msg: str, *args, **kwargs):
 
 
 class EnvironmentSpecPluginNotDetected(SpecNotFound):
-    def __init__(self, name: str, plugin_names: Iterable[str], *args, **kwargs):
+    def __init__(
+        self,
+        name: str,
+        plugin_names: Iterable[str],
+        autodetect_disabled_plugins: Iterable[str] = [],
```
jaimergp:
```suggestion
        autodetect_disabled_plugins: Iterable[str] = (),
```

Immutable is better.

conda/plugins/types.py
```diff
@@ -420,6 +420,13 @@ class EnvironmentSpecBase(ABC):
     Base class for all env specs.
     """
 
+    # Determines if the EnvSpec plugin should be included in the set
+    # of available plugins checked during environment_spec plugin detection.
+    # If set to False, the only way to use the plugin will be through explicitly
+    # requesting it as a cli argument or setting in .condarc. By default,
+    # autodetection is enabled.
+    detection_supported: bool = True
```
jaimergp: Probably a good idea to mark it as a [ClassVar](https://docs.python.org/3/library/typing.html#typing.ClassVar)

```suggestion
    detection_supported: ClassVar[bool] = True
```

news/14914-env-spec-plugins-can-opt-out-of-autodetection
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Allow env spec plugins to enable/disable auto detection. (#14914)
```
jaimergp:
```suggestion
* Allow env spec plugins to opt-out of auto detection. (#14914)
```

conda/exceptions.py
```diff
@@ -1297,16 +1297,31 @@ def __init__(self, msg: str, *args, **kwargs):
...
 
-            Available plugins: {dashlist(plugin_names, 4)}
             """
         )
+        if len(autodetect_disabled_plugins) > 0:
+            msg += dals(
+                """
+                Found compatible plugins but they must be explicitly selected.
+                Request conda to use these plugins by providing
+                the cli argument `--environment-spec PLUGIN_NAME`:
```
jaimergp, line 1322: Ami I missing something or is this argument not implemented yet?

jaimergp (commented):
Looking good, but I think we are missing `--environment-spec` somewhere? Or is this supposed to come in a follow up PR?

conda/exceptions.py
soapy1: This was implemented in another PR. If you pull this one down and try it out it should all work as described

conda/env/specs/binstar.py
soapy1: good to call this one out! I left a note in the commit message:

> Annotating the whole class makes passing refrences to the
class awkward, since it causes 'BinstarSpec' to be a function
opposed to a class. This causes things like CondaEnvironmentSpecifier
to have the wrong type for it's environment_spec field.

So, before this change something like:
```
for hook in hooks:
        if not hook.environment_spec.detection_supported:
            . . .
```

would error on the binstar plugin because the `environment_spec` type is a function instead of a class because it has been decorated.

Moving the deprecation decorator to the `__init__` method still produces the result of logging the deprecation whenever the binstar class is instantiated, but avoids the above mentioned awkwardness.

conda/exceptions.py
jaimergp: Duh, true! xref here for visibility: https://github.com/conda/conda/pull/14877

conda/env/specs/binstar.py
jaimergp: Hm, I'm not sure that's the intended design. cc @kenodegard. We might need an issue to track this if it's not there yet.

jaimergp (approved):

jaimergp:
@jezdez, given your previous comments in this PR, can you take another look?

jezdez (approved):

conda/env/specs/binstar.py
jezdez: Yeah, that's an odd one, I wouldn't block on it TBH, given that the deprecation is ending soon

jezdez: @kenodegard Do you have a suggestion how to handle this?

PR #14916, opened 2025-06-06
Handle exceptions coming from environment spec plugins

soapy1:
### Description

This PR introduces some more error handling into the environemnt_spec plugin detection process. Now,
if a plugin has an unhandled exception in their `can_handle` function:
* detecting plugins (`detect_environment_specifier`) will capture + log errors coming from plugins, instead of erroring out
* explicitly requesting a plugin (`get_environment_specifier_by_name`) with unhandled exceptions will wrap the error up in a conda exception and provide information about the error to the user

See the attached tests for examples of how this works.

This PR is just handling exceptions during the plugin detection phase, so that would be in the plugin implementation of the `can_handle` function. This PR does not introduce any error handling for retrieving an environment from the plugin.

depends on the changes to the plugin manager introduced in https://github.com/conda/conda/pull/14877 and https://github.com/conda/conda/pull/14914

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -542,13 +542,10 @@ class CondaPluginManager(pluggy.PluginManager):
                 f"{', '.join([hook.name for hook in hooks])}"
             )
         if len(found) == 1:
+            # Try to load the plugin and check if it can handle the environment spec
             try:
                 if found[0].environment_spec(source).can_handle():
                     return found[0]
-                else:
-                    raise PluginError(
-                        f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
-                    )
             except Exception as e:
                 raise PluginError(
                     dals(
@@ -558,6 +555,11 @@ class CondaPluginManager(pluggy.PluginManager):
                         {type(e).__name__}: {e}
                         """
                     )
+                )
+            else:
+                # If the plugin was not able to handle the environment spec, raise an error
+                raise PluginError(
+                    f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
                 )
         elif len(found) > 1:
             raise PluginError(
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -542,11 +542,22 @@ class CondaPluginManager(pluggy.PluginManager):
                 f"{', '.join([hook.name for hook in hooks])}"
             )
         if len(found) == 1:
-            if found[0].environment_spec(source).can_handle():
-                return found[0]
-            else:
+            try:
+                if found[0].environment_spec(source).can_handle():
+                    return found[0]
+                else:
+                    raise PluginError(
+                        f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
+                    )
+            except Exception as e:
                 raise PluginError(
-                    f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
+                    dals(
+                        f"""
+                        An error occured when handling '{source}' with plugin '{name}'.
+
+                        {type(e).__name__}: {e}
+                        """
+                    )
                 )
         elif len(found) > 1:
             raise PluginError(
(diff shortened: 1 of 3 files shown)


soapy1:
pre-commit.ci autofix

conda/plugins/manager.py
```diff
@@ -567,21 +578,29 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
...
+                            source,
+                            hook.name,
+                        )
+                except Exception as e:
+                    log.error(
+                        "EnvironmentSpec hook: an error occurred when handling '%s' with plugin '%s'. %s",
+                        source,
+                        hook.name,
+                        e,
+                    )
```
jezdez, line 605: Any reasons not to make this a `PluginError`?

soapy1: This is part of the detection process where conda loops thru all the installed plugins to see if anything can handle the provided environment spec, so we don't want to throw an error. For example, if the user installs a broken plugin that causes an exception in it's `can_handle`  function we don't want conda to completely break. Instead, conda should take that as a singal that the plugin is not able to handle the plugin, log that an error was thrown, and continue with the detection process.

soapy1: That's in contrast to `get_environment_specifier_by_name`, where the caller is specifically requesting a plugin. If that plugin throws an error in a place where it's not suppose to, conda should catch that error and bundle it up as a `PluginError`. You can see that above on line 549

conda/plugins/manager.py
```diff
@@ -542,11 +542,22 @@ def get_environment_specifier_by_name(
...
+            except Exception as e:
                 raise PluginError(
-                    f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
+                    dals(
+                        f"""
+                        An error occured when handling '{source}' with plugin '{name}'.
+
+                        {type(e).__name__}: {e}
+                        """
+                    )
```
kenodegard: Need to rework this otherwise the first `PluginError` will be be caught and wrapped in the second `PluginError`

kenodegard (changes requested):

conda/plugins/manager.py
soapy1: good catch! I also updated the tests to reflect the change required here. Thanks

conda/plugins/manager.py
```diff
@@ -542,12 +542,25 @@ def get_environment_specifier_by_name(
...
                 )
+
+            # If the plugin was not able to handle the environment spec, raise an error
+            raise PluginError(
+                f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
+            )
```
kenodegard: would this be clearer if we did a try-except-else instead?

tests/plugins/test_env_specs.py
```diff
@@ -129,10 +157,14 @@ def test_raises_an_error_if_named_plugin_can_not_be_handled(
     """
     Ensures that an error is raised if the user requests a plugin exists, but can't be handled
     """
-    with pytest.raises(PluginError):
+    with pytest.raises(PluginError) as e:
```
kenodegard: Personally prefer using the `match` keyword instead:

```suggestion
    with pytest.raises(
        PluginError,
        match=r"Requested plugin 'rand-spec' is unable to handle environment spec",
    ):
```

tests/plugins/test_env_specs.py
```diff
@@ -167,3 +199,36 @@ def test_explicitly_select_a_non_autodetect_plugin(
...
+    """
+    Ensures that explicitly selecting a plugin that has errors is handled appropriately
+    """
+    filename = "test.random"
+    with pytest.raises(PluginError) as e:
+        plugin_manager.get_environment_specifier_by_name(filename, "naughty")
+        assert (
+            "An error occured when handling '{filename}' with plugin 'naughty'."
+            in str(e)
+        )
```
kenodegard:
```suggestion
    with pytest.raises(
        PluginError,
        match=rf"An error occured when handling '{filename}' with plugin 'naughty'.",
    ):
        plugin_manager.get_environment_specifier_by_name(filename, "naughty")
```

kenodegard (dismissed):
Looks good, minor suggestions

conda/plugins/manager.py
soapy1: ya, that sounds good to me

kenodegard (dismissed):

conda/plugins/manager.py
jaimergp: Can we add the `exc_info=e` info in a debug statement to facilitate debugging? We are using this pattern in places like https://github.com/conda/conda/blob/eacc4c4d7dd3b88524e1ca8eec9d5c2146de167a/conda/core/envs_manager.py#L226-L227

```suggestion
                except Exception as e:
                    log.error(
                        "EnvironmentSpec hook: an error occurred when handling '%s' with plugin '%s'. %s",
                        source,
                        hook.name,
                        e,
                    )
                    log.debug("%r", e, exc_info=e)
```

news/14916-handle-exceptions-from-env-spec-plugins
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Handle exceptions coming from environment spec plugins can_handle functions. (#14916)
```
jaimergp:
```suggestion
* Handle exceptions coming from environment spec plugins `can_handle` functions. (#14916)
```

kenodegard (approved):

soapy1:
Looks like this test failure is due to a network error. Probably just needs a rerun

jezdez (approved):

PR #14925, opened 2025-06-10
Lazily evaluate cli options originating from the plugin manager

soapy1:
### Description
Follow up to https://github.com/conda/conda/pull/14877

This PR lazily evaluates `choices` for cli arguments that require information from the plugin manager. This will ensure that the plugin manager and context are fully initialized before calling on them.

The message returned to the user when selecting an invalid plugin is unchanged:

```

$ conda env create --env-spec idontexist         [9:59:00]
                        [--environment-specifier ENVIRONMENT_SPECIFIER] [--no-default-packages]
                        [--json] [--console CONSOLE] [-v] [-q] [-d] [-y] [--solver SOLVER]
                        [--subdir SUBDIR]
                        [remote_definition]
conda env create: error: argument '--env-spec': invalid choice: 'idontexist' (choose from 'binstar', 'environment.yml', 'requirements.txt')
```

```
$ conda install --solver  idontexist            [10:00:20]
usage: conda install [-h] [--revision REVISION] [--override-frozen] [-n ENVIRONMENT | -p PATH]
                     [-c CHANNEL] [--use-local] [--override-channels] [--repodata-fn REPODATA_FNS]
                     [--experimental {jlap,lock}] [--no-lock]
                     [--repodata-use-zst | --no-repodata-use-zst] [--strict-channel-priority]
                     [--no-channel-priority] [--no-deps | --only-deps] [--no-pin] [--copy]
                     [--no-shortcuts] [--shortcuts-only SHORTCUTS_ONLY] [-C] [-k] [--offline]
                     [--json] [--console CONSOLE] [-v] [-q] [-d] [-y] [--download-only]
                     [--show-channel-urls] [--file FILE] [--solver SOLVER] [--force-reinstall]
                     [--freeze-installed | --update-deps | -S | --update-all | --update-specs]
                     [--clobber] [--dev]
                     [package_spec ...]
conda install: error: argument '--solver': invalid choice: 'idontexist' (choose from 'classic', 'libmamba')
```
### Checklist - did you ...

### DIFF
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 from argparse import (
     SUPPRESS,
+    Action,
     BooleanOptionalAction,
     _HelpAction,
     _StoreAction,
@@ -17,6 +18,21 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, _ArgumentGroup, _MutuallyExclusiveGroup
+
+
+class LazyChoicesAction(Action):
+    def __init__(self, option_strings, dest, choices_func, **kwargs):
+        self.choices_func = choices_func
+        super().__init__(option_strings, dest, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        valid_choices = self.choices_func()
+        if values not in valid_choices:
+            choices_string = ", ".join(f"'{val}'" for val in valid_choices.keys())
+            parser.error(
+                f"argument '{option_string}': invalid choice: {values!r} (choose from {choices_string})"
+            )
+        setattr(namespace, self.dest, values)
 
 
 class _ValidatePackages(_StoreAction):
@@ -420,7 +436,8 @@ def add_parser_solver(p: ArgumentParser) -> None:
     group.add_argument(
         "--solver",
         dest="solver",
-        choices=context.plugin_manager.get_solvers(),
+        action=LazyChoicesAction,
+        choices_func=context.plugin_manager.get_solvers,
         help="Choose which solver backend to use.",
         default=NULL,
     )
@@ -583,7 +600,8 @@ def add_parser_environment_specifier(p: ArgumentParser) -> None:
     p.add_argument(
         "--environment-specifier",
         "--env-spec",  # for brevity
-        choices=context.plugin_manager.get_environment_specifiers(),
+        action=LazyChoicesAction,
+        choices_func=context.plugin_manager.get_environment_specifiers,
         default=NULL,
         help="(EXPERIMENTAL) Specify the environment specifier plugin to use.",
     )
--- a/news/14925-lazily-load-cli-choices
+++ b/news/14925-lazily-load-cli-choices
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Lazily evaluate cli options originating from the plugin manager. (#14925)
+
+### Bug fixes
+
+* <news item>
+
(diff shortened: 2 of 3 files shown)


conda/cli/helpers.py
```diff
@@ -19,6 +20,21 @@
...
+        super().__init__(option_strings, dest, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        valid_choices = self.choices_func()
+        if values not in valid_choices:
+            choices_string = ", ".join(f"'{val}'" for val in valid_choices.keys())
+            parser.error(
+                f"argument '{option_string}': invalid choice: {values!r} (choose from {choices_string})"
+            )
+        setattr(namespace, self.dest, values)
```
kenodegard, line 35:
```suggestion
from ..auxlib.type_coercion import maybecall

class LazyChoicesAction(Action):
    @property
    def choices(self):
        return maybecall(self._choices)
    
    @choices.setter
    def choices(self, value):
        self._choices = value
```

conda/cli/helpers.py
```diff
@@ -420,7 +436,8 @@ def add_parser_solver(p: ArgumentParser) -> None:
     group.add_argument(
         "--solver",
         dest="solver",
-        choices=context.plugin_manager.get_solvers(),
+        action=LazyChoicesAction,
+        choices_func=context.plugin_manager.get_solvers,
```
kenodegard, line 440:
```suggestion
        choices=context.plugin_manager.get_solvers,
```

conda/cli/helpers.py
```diff
@@ -583,7 +600,8 @@ def add_parser_environment_specifier(p: ArgumentParser) -> None:
     p.add_argument(
         "--environment-specifier",
         "--env-spec",  # for brevity
-        choices=context.plugin_manager.get_environment_specifiers(),
+        action=LazyChoicesAction,
+        choices_func=context.plugin_manager.get_environment_specifiers,
```
kenodegard, line 604:
```suggestion
        choices=context.plugin_manager.get_environment_specifiers,
```

kenodegard (changes requested):
Small suggestion to avoid implementing `Action.__call__`

conda/cli/helpers.py
soapy1: I think it's better to override the `__call__` function. From the [argparse docs](https://docs.python.org/3/library/argparse.html):

> The recommended way to create a custom action is to extend [Action](https://docs.python.org/3/library/argparse.html#argparse.Action), overriding the __call__() method and optionally the __init__() and format_usage() methods. You can also register custom actions using the [register()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.register) method and reference them by their registered name.

It's also mentioned in the [Action class part of the docs](https://docs.python.org/3/library/argparse.html#action-classes)

> Action instances should be callable, so subclasses must override the __call__() method, which should accept four parameters:

When I apply this diff to my branch I get an error when I use valid options, for example:

```
$ conda install flask --solver classic
. . .
File "/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/lib/python3.10/argparse.py", line 868, in __call__
        raise NotImplementedError(_('.__call__() not defined'))
    NotImplementedError: .__call__() not defined
. . .
```

kenodegard: Ah we'd need to subclass `_StoreAction` instead.

kenodegard (approved):

jaimergp (approved):

PR #14928, opened 2025-06-11
Implement experimental feature handling and update CLI for experiments

### LINKED ISSUE #14895: Policy for experimental APIs
### Checklist

### What is the idea?

Introduce a policy where we mark code as not bound to CEP 9 deprecation cycles.

I suggest we adopt a convention where non-stable API names are prefixed with `experimental`:

```python
EXPERIMENTAL_CONSTANT = ...
class ExperimentalClass:
    ...
def experimental_function(arg, experimental_arg = None):
    ...

experimental_name = ...
```

### Why is this needed?

We are bound to CEP 9 when it comes to deprecation cycles. This CEP is supposed to alleviate concerns about code breakage when we iterate on features, remove dead code, etc. However, it also ties us to API contracts for new code that maybe needs to change often during a few release cycles while we iron out all the necessary details to make it stable.

### What should happen?

Experimental-labeled code is not subject to deprecation cycles, but should not remain marked as experimental forever (looking at `conda.api`). This should also be reflected in their docstring prominently, as well as in the release notes.

After a sensible time, the experimental code should either be marked as stable by renaming the appropriate prefix, or deprecated and removed if the experiment turned out not be fruitful.

I don't think we need a decorator in `conda.deprecations`. Naming, docstrings and release notes are sufficient.

### Additional Context

This would help PRs that introduce APIs like https://github.com/conda/conda/pull/14870 be merged without concerns about "being ready for production" within a single release cycle.

jezdez:
### Description

Fix #14895

- Introduces `@experimental` decorator for managing experimental features, including registration and validation.
- Added `--experiments` argument to the CLI to list experimental features and their expiration versions.
- Updated `InfoRenderer` to render experimental features in the output.
- Enhanced tests to cover the new experimental functionality and ensure proper registration and validation of experimental features.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -81,6 +81,11 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         help="List environment variables.",
     )
     p.add_argument(
+        "--experiments",
+        action="store_true",
+        help="List experimental features and their expiration versions.",
+    )
+    p.add_argument(
         "--root",
         action=deprecated.action(
             "25.9",
@@ -409,7 +414,9 @@ def get_main_info_str(info_dict: dict[str, Any]) -> str:
 
 
 #: Possible components for the info command to render
-InfoComponents = Literal["base", "channels", "envs", "system", "detail", "json_all"]
+InfoComponents = Literal[
+    "base", "channels", "envs", "system", "detail", "json_all", "experiments"
+]
 
 
 class InfoRenderer:
@@ -430,6 +437,7 @@ class InfoRenderer:
             "envs": "envs_list",
             "system": None,
             "json_all": None,
+            "experiments": None,
         }
 
     def render(self, components: Iterable[InfoComponents]):
@@ -502,6 +510,33 @@ class InfoRenderer:
 
         return "\n".join(output)
 
+    def _experiments_component(self) -> str | dict[str, list[dict[str, str]]]:
+        """Render experimental features component."""
+        from conda.deprecations import experimental
+
+        if self._context.json:
+            rv = {"experiments": []}
+            for experiment in experimental.scan():
+                rv["experiments"].append(
+                    {
+                        "name": experiment["prefix"],
+                        "until": experiment["until"],
+                        "addendum": experiment["addendum"] or "",
+                    }
+                )
+            return rv
+        else:
+            experiments = experimental.scan()
+            if not experiments:
+                return "No experimental features currently active."
+
+            lines = ["Experimental Features:"]
+            for experiment in experiments:
+                lines.append(f"  • {experiment['prefix']} (until {experiment['until']})")
+                if experiment["addendum"]:
+                    lines.append(f"    {experiment['addendum']}")
+            return "\n".join(lines)
+
     def _json_all_component(self) -> dict[str, Any]:
         return self._info_dict
 
@@ -529,8 +564,11 @@ def iter_info_components(args: Namespace, context: Context) -> Iterable[InfoComp
     if args.unsafe_channels:
         yield "channels"
 
+    if args.experiments:
+        yield "experiments"
+
     if (
-        (args.all or (not args.envs and not args.system))
+        (args.all or (not args.envs and not args.system and not args.experiments))
         and not context.json
         and not args.base
         and not args.unsafe_channels
@@ -543,7 +581,12 @@ def iter_info_components(args: Namespace, context: Context) -> Iterable[InfoComp
     if (args.system or args.all) and not context.json:
         yield "system"
 
-    if context.json and not args.base and not args.unsafe_channels:
+    if (
+        context.json
+        and not args.base
+        and not args.unsafe_channels
+        and not args.experiments
+    ):
         yield "json_all"
 
 
@@ -557,6 +600,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
      * ``conda info --unsafe-channels``
      * ``conda info --envs``
      * ``conda info --system``
+     * ``conda info --experiments``
     """
 
     from ..base.context import context
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -513,30 +513,37 @@ class InfoRenderer:
     def _experiments_component(self) -> str | dict[str, list[dict[str, str]]]:
         """Render experimental features component."""
         from conda.deprecations import experimental
+
         experiments = experimental.scan(check=True)
 
         if self._context.json:
-            rv = {"experiments": []}
-            for experiment in experiments:
-                rv["experiments"].append(
+            return {
+                "experiments": [
                     {
                         "name": experiment["prefix"],
                         "until": experiment["until"],
                         "addendum": experiment["addendum"] or "",
                     }
-                )
-            return rv
+                    for experiment in experiments
+                ]
+            }
         else:
             if not experiments:
                 return "No experimental features currently active."
 
-            lines = ["Experimental Features:"]
-            for experiment in experiments:
-                lines.append(
-                    f"  • {experiment['prefix']} (until {experiment['until']})"
-                )
+            def format_experiment(experiment):
+                yield f"  • {experiment['prefix']} (until {experiment['until']})"
                 if experiment["addendum"]:
-                    lines.append(f"    {experiment['addendum']}")
+                    yield f"    {experiment['addendum']}"
+
+            lines = [
+                "Experimental Features:",
+                *[
+                    line
+                    for experiment in experiments
+                    for line in format_experiment(experiment)
+                ],
+            ]
             return "\n".join(lines)
 
     def _json_all_component(self) -> dict[str, Any]:
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -513,10 +513,11 @@ class InfoRenderer:
     def _experiments_component(self) -> str | dict[str, list[dict[str, str]]]:
         """Render experimental features component."""
         from conda.deprecations import experimental
+        experiments = experimental.scan(check=True)
 
         if self._context.json:
             rv = {"experiments": []}
-            for experiment in experimental.scan():
+            for experiment in experiments:
                 rv["experiments"].append(
                     {
                         "name": experiment["prefix"],
@@ -526,13 +527,14 @@ class InfoRenderer:
                 )
             return rv
         else:
-            experiments = experimental.scan()
             if not experiments:
                 return "No experimental features currently active."
 
             lines = ["Experimental Features:"]
             for experiment in experiments:
-                lines.append(f"  • {experiment['prefix']} (until {experiment['until']})")
+                lines.append(
+                    f"  • {experiment['prefix']} (until {experiment['until']})"
+                )
                 if experiment["addendum"]:
                     lines.append(f"    {experiment['addendum']}")
             return "\n".join(lines)
@@ -564,7 +566,7 @@ def iter_info_components(args: Namespace, context: Context) -> Iterable[InfoComp
     if args.unsafe_channels:
         yield "channels"
 
-    if args.experiments:
+    if getattr(args, "experiments", False):
         yield "experiments"
 
     if (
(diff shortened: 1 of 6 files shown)


tests/test_deprecations.py
```diff
@@ -54,7 +55,7 @@
 @parametrize_dev
 def test_function(
     deprecated: DeprecationHandler,
-    warning: DevDeprecationType | None,
+    warning: type[Warning] | None,
```
jezdez, line 58: Those were leftovers from another change

tests/test_experiments.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+def test_scan_check_real_features_not_expired() -> None:
+    """Test that real experimental features in the codebase haven't expired beyond grace period."""
+    # Use actual conda version to check real experimental features
+    experimental = ExperimentHandler(__version__)
+
+    # This will raise ExperimentConcluded if any real experiments have expired
+    # beyond their grace period, fulfilling requirement 6
+    experimental.scan(check=True)
```
jezdez, line 273: This is the load-bearing test that raises an exception once experiments have surpassed their expiration date

conda/deprecations.py
```diff
@@ -422,4 +603,319 @@ def _generate_message(
         )
 
 
+class ExperimentalFeatureVisitor(ast.NodeVisitor):
+    """Simple AST visitor to find experimental decorators."""
```
dholth: "simple"

conda/cli/main_info.py
```diff
@@ -502,6 +510,42 @@ def _system_component(self) -> str:
...
+                    }
+                    for experiment in experiments
+                ]
+            }
+        else:
+            if not experiments:
+                return "No experimental features currently active."
+
+            def format_experiment(experiment):
+                yield f"  • {experiment['prefix']} (until {experiment['until']})"
```
jaimergp: We are not using non-ASCII symbols in other parts of the CLI, not sure if we want to respect that for now and just do a simple dash:

```suggestion
                yield f"  - {experiment['prefix']} (until {experiment['until']})"
```

conda/cli/main_package.py
```diff
@@ -18,7 +18,7 @@
 def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser:
     from .helpers import add_parser_prefix
 
-    summary = "Create low-level conda packages. (EXPERIMENTAL)"
+    summary = "Create low-level conda packages."
```
jaimergp, line 21: Is this related to this PR?

conda/models/environment.py
```diff
@@ -87,10 +92,12 @@ def __post_init__(self):
                     )
 
     @classmethod
+    @experimental(
```
jaimergp, line 95: Is this necessary if the whole module is experimental?

conda/models/environment.py
```diff
@@ -173,8 +180,8 @@ def merge(cls, *environments):
             external_packages=external_packages,
             explicit_packages=explicit_packages,
             name=name,
-            platform=platform,
-            prefix=prefix,
+            platform=platform or "",
+            prefix=prefix or "",
```
jaimergp, line 184: These are required, aren't they?

conda/deprecations.py
jezdez: Heh, no lies detected. The alternatives were... not simple.

conda/cli/main_package.py
jezdez: When I was searching the code base, this popped up, and given that the package subcommand isn't an actively run experiment, I thought to correct it, but I can separate it in a different PR for better tracking

conda/models/environment.py
jezdez: For this case, it is probably fine to just mark the module to be an experiment. Context is that I wanted to call out that the class and classmethod were separately marked as experimental, with different reasons.

conda/models/environment.py
jezdez: Yeah, in the `merge` classmethod above, `platform` is set to `None`, but the type is required to be `str` in the dataclass, same for the `prefix`.

conda/deprecations.py
jezdez:
```suggestion
    """An AST visitor to find experimental decorators."""
```

conda/cli/main_package.py
jaimergp: As long as it's clearly spelt out in the news item, it's ok to leave here imo.

conda/deprecations.py
jaimergp, line 1: I was expecting something way simpler, where we don't even AST-traverse anything, more like a decorator that given a set of runtime conditions annotates the inner function in a global list or something. In pseudocode:

```python
_experimental_names = set()
def experimental(name: Any):
  if os.environ.get("PYTEST_CURRENT_TEST") or ("info", "--experimental") in cli_args:
    global _experimental_names
    _experimental_names.add(name)
  return func
```

Not every elegant or as flexible as the proposed handler here, but way less code to understand, test and maintain. Any strong reasons for the more complex approach or could we compromise a bit in the honor of simplicity?

dholth:
At first glance I found it very strange to expose the `--experiments` CLI since the ordinary conda user should not have to care. I also worry about having to think about whether code is experimental or not, instead of e.g. the pip "this entire section of the codebase is not API" approach.

jezdez:
@dholth What interface to bubble up experiments would you not find strange? Regarding the pip strategy, that has famously led to a whole ecosystem being frustrated about the authors, but was a compromise due to the lack of resources to maintain a plugin API.

conda/deprecations.py
jezdez: Just the usual, parsing CLI arguments is funny business and doing it live like this has led to odd edge cases, I think code parsing is relatively safe since it's a Python built-in and most of the code is to provide a familiar API coming from the deprecations' framework that we're already maintaining (hence the many shared code paths). I'm not suggesting your idea isn't also valid, but that we'd eventually want more (as we did with the deprecations) and then start building the tower of complexity back up. I basically jumped a few steps forward in that discussion.

jezdez: I should add this is a suggested implementation, to facilitate the conversation in #14895 ;)

conda/cli/main_package.py
jezdez: Good point, will do.

conda/cli/main_info.py
```diff
@@ -502,6 +510,42 @@ def _system_component(self) -> str:
 
         return "\n".join(output)
 
+    def _experiments_component(self) -> str | dict[str, list[dict[str, str]]]:
+        """Render experimental features component."""
+        from conda.deprecations import experimental
```
jaimergp:
```suggestion
        from ..deprecations import experimental
```

conda/deprecations.py
```diff
@@ -132,36 +334,57 @@ def argument(
...
+                prefix=prefix,
+                addendum=final_addendum,
                 deprecation_type=deprecation_type,
             )
 
-            # alert developer that it's time to remove something
-            if not category:
-                raise DeprecatedError(message)
+            # Get the function signature to determine parameter positions
+            import inspect
```
jaimergp, line 365: Why is this import all the way down here?

conda/deprecations.py
```diff
@@ -422,4 +603,319 @@ def _generate_message(
...
+        for keyword in call.keywords:
+            if keyword.arg == arg_name and isinstance(keyword.value, ast.Constant):
+                return keyword.value.value
+        return None
+
+
+class ExperimentHandler(BaseHandler):
+    """Handler for experimental features that tracks usage without warning users."""
+
+    def __init__(self: Self, version: str) -> None:
```
jaimergp:
```suggestion
    def __init__(self, version: str):
```

I thought we had established that `self` must not be typed, and `__init__` does not need a return type either.

conda/deprecations.py
```diff
@@ -422,4 +603,319 @@ def _generate_message(
...
+
+        # Use visitor
+        visitor = ExperimentalFeatureVisitor(module_name, self)
+        visitor.visit(tree)
+        return visitor.features
+
+    # Consolidated decorator method
+    def __call__(
+        self: Self,
+        until: str,
```
jaimergp, line 838: There's no `since` argument? The way I saw this API was like we would indicate the experiment start time, and then establish a maximum duration we would tolerate. This way we have an idea of how long the experiment has been running for when the expiry error shows up in pytest (and not be so tempted to just extend the experiment lifetime).

conda/models/environment.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
-"""EXPERIMENTAL Conda environment data model"""
+"""Conda environment data model"""
```
jaimergp, line 3: Personally I'd like to retain the docstring marks, and would even advocate for a test that ensures that all deprecated objects have `*EXPERIMENTAL*` in their docstring so folks checking the code can really see it (not sure if they would see the `experimental` mark amid all the top-level import/logging noise).

conda/deprecations.py
jaimergp: I see your points. I'm just unsure if we would be ok with maintaining this a year from now when we realize that we had not been listing a type of node because of a bug or something and now we have to re-read and re-understand the new classes.

But I won't block this on that issue. The code is written and it works, so at this point is easier to just go ahead.

jaimergp:
I'd like @kenodegard to take a look at this because of the refactors in the `deprecations` module.

jaimergp:
> At first glance I found it very strange to expose the --experiments CLI since the ordinary conda user should not have to care.

I kind of agree with @dholth here. Maybe we can just make `python -m conda.deprecations` executable?

conda/deprecations.py
```diff
@@ -422,4 +603,319 @@ def _generate_message(
...
+        if not isinstance(decorator, ast.Call):
+            return None
+
+        # Check if it's @experimental(...) or @experimental.method(...)
+        until = None
+        addendum = None
+
+        if isinstance(decorator.func, ast.Name) and decorator.func.id == "experimental":
+            # @experimental(until="...")
+            until = self._get_call_arg(decorator, "until") or (
```
soapy1, line 661: I think the idea of having something be marked experimental until a certain date or release is pretty intense. Like, what if the experimental feature is not ready to be not experimental by this time? Or, what if the experiment has run it's course and needs to be removed?

soapy1:
It looks like conda already has some notion of experimental features. There is this context field [`experimental`](https://github.com/conda/conda/blob/main/conda/base/context.py#L500). Looks like it's currently only used for repodata `jlap` support: https://github.com/conda/conda/blob/main/conda/gateways/repodata/__init__.py#L106. Does the notion of marking experiments introduced in this PR work with this existing notion of experiments? Or, does it need to?

jezdez:
> It looks like conda already has some notion of experimental features. There is this context field [`experimental`](https://github.com/conda/conda/blob/main/conda/base/context.py?rgh-link-date=2025-06-14T00%3A26%3A14Z#L500). Looks like it's currently only used for repodata `jlap` support: [`main`/conda/gateways/repodata/__init__.py#L106](https://github.com/conda/conda/blob/main/conda/gateways/repodata/__init__.py?rgh-link-date=2025-06-14T00%3A26%3A14Z#L106). Does the notion of marking experiments introduced in this PR work with this existing notion of experiments? Or, does it need to?

It's overlapping in meaning, the experimental field was added as a simplistic feature flag for the libmamba solver, then later used for the JLAP efforts. The latter will likely be removed soon by @dholth as part of the sharded repodata feature work, so I wouldn't think the context item is something to worry about too much. A more distributed feature flagging system (using subject names in the context) is probably more useful anyway.

conda/models/environment.py
jezdez: Okay

conda/deprecations.py
jezdez: Adding the since argument now

conda/deprecations.py
jezdez: Just a leftover from experimenting moving up

conda/deprecations.py
jezdez: Ok, I'll have an answer for this in a bit

jezdez:
@conda/conda-maintainers Please provide feedback if you're interesting in moving forward with this. The current status quo of adding "experimental" to docstrings are IMO not an alternative.

jezdez:
Closing since it doesn't seem to spark joy.

PR #14886, opened 2025-05-29
Implement environment export plugin hook

jezdez:
### Description

More on this in https://github.com/conda/conda-environments-project-mgmt/issues/52 (tickets in conda repo to follow)

Relies on https://github.com/conda/conda/pull/14820
Resolves https://github.com/conda/conda/issues/15048

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #14937, opened 2025-06-13
Environment Spec plugins should return an Environment model

### LINKED ISSUE #62: Make environment specifier hook API return the new `Environment` dataclass object

soapy1:
### Description

This PR updates the `EnvironmentSpecBase` interface to return a `model.Environment` (environment model)  instead of an `env.env.Environment` (environment.yml model). The `env.env.Environment` does not fully describe a conda environment. On the other hand, `model.Environment` can fully describe an environment. So, this type makes more sense to be the return type for env spec plugins. For more information on the Environment model see https://docs.google.com/document/d/1Q_M66kFGCYLuaqAqez_jlsqxnQEQh5tSUBtvEz3ixdA/edit?tab=t.gmwr9pinfepw.

This PR also updates `conda env` commands and env installers to use the new Environment model. 

This is a refactor PR, there should be no changes in conda's behaviour.

fixes https://github.com/conda/conda-environments-project-mgmt/issues/62

depends on https://github.com/conda/conda/pull/14820

### Checklist - did you ...


jezdez:
Could we start this up again, now that #14820 is merged? #14886 could really need a way to convert the legacy environment classes to the new models, see [_create_environment_from_prefix](https://github.com/conda/conda/pull/14886/files#diff-45818606a34d3f3f3b62eacc4c4a7733a8c4ff3a3670e43a5bdda7e26d6f420aR18) here

soapy1:
Absolutely! Will whip it up in just a bit

conda/env/installers/conda.py
```diff
@@ -56,18 +57,18 @@ def _solve(
 
 def dry_run(
     specs: list[str], args: Namespace, env: Environment, *_, **kwargs
-) -> Environment:
+) -> EnvironmentYaml:
```
soapy1, line 58: We want to continue to return the `conda.env.env.Environment` here so that we aren't changing the installer api. Eventually this should move over to use the new Environment model. But I think it's better to chunk that in another piece of work to keep this change focused on the env spec api.

tests/env/test_explicit_environment.py
```diff
@@ -1,108 +0,0 @@
-# Copyright (C) 2012 Anaconda, Inc
```
soapy1, line 1: We don't want to support an explicit environment.yaml anymore. The `conda.env.env.Environment` class should just represent the conda environment.yaml format.

tests/plugins/test_explicit_spec.py
```diff
@@ -1,102 +0,0 @@
-# Copyright (C) 2012 Anaconda, Inc
```
soapy1, line 1: These tests are duplicates of the ones in `tests/env/specs/test_explicit.py`

tests/env/installers/test_conda_installer_explicit.py
```diff
@@ -4,9 +4,12 @@
 
 import pytest
 
-from conda.env.env import Environment
```
soapy1, line 7: Maybe this whole file can be deleted?

soapy1:
pre-commit.ci autofix

tests/env/test_explicit_environment.py
jezdez: Wait, what's "an explicit environment.yaml"? `@EXPLICIT` input files are defined in [CEP 23](https://conda.org/learn/ceps/cep-0023), we can't not support them

conda/env/installers/conda.py
jezdez: I can't parse this statement

conda/cli/main_env_update.py
```diff
@@ -154,7 +154,10 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     # e.g. due to conda_env being upgraded or Python version switched.
     installers = {}
 
-    for installer_type in env.dependencies:
+    if len(env.requested_packages) > 0:
+        installers["conda"] = get_installer("conda")
+
+    for installer_type in env.external_packages:
```
jezdez: Why do we only get and set the conda installer if there are requested packages? This makes me wonder why we have a separation between requested and external packages

jaimergp: `requested` packages are only for `conda` packages. `external` are everything else (PyPI and whatever ecosystem we end up adding).

conda/cli/main_env_create.py
```diff
@@ -185,13 +185,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             installer = get_installer(installer_type)
             result[installer_type] = installer.install(prefix, args_packages, args, env)
 
-        if len(env.dependencies.items()) == 0:
+        if len(env.external_packages.items()) == 0 and len(env.requested_packages) == 0:
```
jaimergp:
```suggestion
        if not env.external_packages and not env.requested_packages:
```

conda/cli/main_env_update.py
```diff
@@ -154,7 +154,10 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     # e.g. due to conda_env being upgraded or Python version switched.
     installers = {}
 
-    for installer_type in env.dependencies:
+    if len(env.requested_packages) > 0:
```
jaimergp:
```suggestion
    if env.requested_packages:
```

conda/misc.py
```diff
@@ -177,6 +158,50 @@ def explicit(
...
+        [spec, next(PackageCacheData.query_all(spec), None)] for spec in specs
+    )
+
+    # Assert that every spec has a PackageCacheRecord
+    specs_with_missing_pcrecs = [
+        str(spec) for spec, pcrec in specs_pcrecs if pcrec is None
+    ]
+    if specs_with_missing_pcrecs:
+        if len(specs_with_missing_pcrecs) == len(specs_pcrecs):
+            raise AssertionError("No package cache records found")
```
jaimergp: That's not the right type of exception to raise here, maybe `SpecNotFound`?

conda/misc.py
```diff
@@ -177,6 +158,50 @@ def explicit(
...
+        else:
+            missing_precs_list = ", ".join(specs_with_missing_pcrecs)
+            raise AssertionError(
```
jaimergp: Same here.

conda/misc.py
```diff
@@ -177,6 +158,50 @@ def explicit(
...
+    # Extract the list of specs
+    fetch_specs = list(_match_specs_from_explicit(lines))
+
+    if context.dry_run:
+        raise DryRunExit()
+
+    # Try to get the specs from cache first
+    try:
+        return _get_package_record_from_specs(fetch_specs)
+    except AssertionError:
```
jaimergp: Update after changing the type of exception above.

conda/misc.py
```diff
@@ -177,6 +158,50 @@ def explicit(
...
+        # If not found in cache, fetch them from the network and try again
+        pfe = ProgressiveFetchExtract(fetch_specs)
+        pfe.execute()
```
jaimergp: This is a no-op if the packages are already cached, so maybe we don't need the `try / except`? It's true that it prints some text to stdout even if no actions are taken, but that can be fixed in a separate PR.

conda/cli/main_compare.py
```diff
@@ -130,11 +130,10 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
     active_pkgs = {pkg.name: pkg for pkg in get_packages(prefix)}
-    specification_pkgs = []
-    if "conda" in env.dependencies:
-        specification_pkgs = specification_pkgs + env.dependencies["conda"]
-    if "pip" in env.dependencies:
-        specification_pkgs = specification_pkgs + env.dependencies["pip"]
+    specification_pkgs = env.requested_packages
+    if len(env.external_packages) > 0:
+        for package_type, packages in env.external_packages.items():
+            specification_pkgs = specification_pkgs + packages
```
jaimergp: If `env.external_packages` is guaranteed to be a dict (even if empty), then we can just do:

```suggestion
    for package_type, packages in env.external_packages.items():
        specification_pkgs = specification_pkgs + packages
```

conda/env/env.py
```diff
@@ -266,6 +267,31 @@ def save(self):
         with open(self.filename, "wb") as fp:
             self.to_yaml(stream=fp)
 
+    def to_environment_model(self):
```
jaimergp:
```suggestion
    def to_environment_model(self) -> EnvironmentModel:
```

conda/env/installers/conda.py
jaimergp: I think it would be ok to make these changes here to make sure we have covered all the API changes and nothing lingers.

tests/env/test_explicit_environment.py
soapy1: TL;DR 
* To clarify, this PR is still supporting explicit input files. 
* `conda.env.env.Environment` should only be representing `environment.yaml` files.

Previously, this was implemented by making `conda.env.env.Environment` support reading these files (that was in https://github.com/conda/conda/pull/14820). That was an ok approach at that time since env spec plugins were meant to be returning `conda.env.env.Environment`.

However, now we want env spec plugins to return `conda.models.environment.Environment`. So, I think we should back out the changes to  `conda.env.env.Environment` that allow it to represent an explicit environment and restore it to it's original purpose which is to represent an `environment.yaml` file. And `ExplicitSpecPlugin` should create a `conda.models.environment.Environment`.

conda/misc.py
soapy1: It looks like this error doesn't really have to do with finding specs in the package cache. I don't want to co-opt it here. Will create a new error class for this.

soapy1:
pre-commit.ci autofix

conda/cli/main_compare.py
```diff
@@ -130,11 +130,9 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
 
     active_pkgs = {pkg.name: pkg for pkg in get_packages(prefix)}
-    specification_pkgs = []
-    if "conda" in env.dependencies:
-        specification_pkgs = specification_pkgs + env.dependencies["conda"]
-    if "pip" in env.dependencies:
-        specification_pkgs = specification_pkgs + env.dependencies["pip"]
+    specification_pkgs = env.requested_packages
+    for package_type, packages in env.external_packages.items():
+        specification_pkgs = specification_pkgs + packages
```
kenodegard:
```suggestion
    specification_pkgs = env.requested_packages
    for packages in env.external_packages.values():
        specification_pkgs = specification_pkgs + packages
```

conda/cli/main_env_create.py
```diff
@@ -170,7 +170,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         installer_type = "conda"
         installer = get_installer(installer_type)
 
-        pkg_specs = env.dependencies.get(installer_type, [])
+        pkg_specs = env.requested_packages
         pkg_specs.extend(args_packages)
```
kenodegard: Will the `list.extend` here extend the `Environment` model as well?

```suggestion
        pkg_specs = [*env.requested_packages, *args_packages]
```

conda/cli/main_env_create.py
```diff
@@ -185,13 +185,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+            installer = get_installer(installer_type)
+            result[installer_type] = installer.install(
+                prefix, env.requested_packages, args, env
+            )
+
+            # install all other external packages
+            for installer_type, pkg_specs in env.external_packages.items():
                 try:
                     installer = get_installer(installer_type)
                     result[installer_type] = installer.install(
```
kenodegard: We can flatten this, right? Don't need to include the `if`-clause since both `env.requested_packages` and `env.external_packages` will be lists?

conda/env/env.py
```diff
@@ -266,6 +267,31 @@ def save(self):
...
+    def to_environment_model(self) -> EnvironmentModel:
+        """Convert the ``Environment`` into a ``model.Environment`` object"""
+        config = EnvironmentConfig(channels=self.channels)
+
+        external_packages = {}
+        if "pip" in self.dependencies.keys() and len(self.dependencies.get("pip")) > 0:
+            external_packages["pip"] = self.dependencies.get("pip")
+
+        if self.prefix is None:
+            self.prefix = context.target_prefix
```
kenodegard: Do we really want to modify `self` when exporting?

conda/env/env.py
```diff
@@ -266,6 +267,31 @@ def save(self):
         with open(self.filename, "wb") as fp:
             self.to_yaml(stream=fp)
 
+    def to_environment_model(self) -> EnvironmentModel:
+        """Convert the ``Environment`` into a ``model.Environment`` object"""
+        config = EnvironmentConfig(channels=self.channels)
+
+        external_packages = {}
+        if "pip" in self.dependencies.keys() and len(self.dependencies.get("pip")) > 0:
+            external_packages["pip"] = self.dependencies.get("pip")
```
kenodegard: Since we don't check if `self.dependencies["pip"]` is an iterable we might as well just keep it simple by only check for falsy:

```suggestion
        external_packages = {}
        if (pip_dependencies := self.dependencies.get("pip")):
            external_packages["pip"] = pip_dependencies
```

conda/env/specs/binstar.py
```diff
@@ -109,7 +109,7 @@ def environment(self) -> Environment:
         )
         if req is None:
             raise EnvironmentFileNotDownloaded(self.username, self.packagename)
-        return from_yaml(req.text)
+        return from_yaml(req.text).to_environment_model()
```
kenodegard, line 132: This return type doesn't match the type annotation

conda/env/specs/requirements.py
```diff
@@ -115,7 +115,10 @@ def environment(self) -> Environment:
...
         dependencies_list = list(yield_lines(self.filename))
+        requested_packages = [MatchSpec(dep) for dep in dependencies_list]
+
         return Environment(
-            dependencies=dependencies_list,
-            filename=self.filename,
+            prefix=context.target_prefix,
+            platform=context.subdir,
+            requested_packages=requested_packages,
         )
```
kenodegard: We're swapping out the `conda.env.env.Environment` return type with `conda.models.environment.Environment`, should we be announcing the change via deprecation?

kenodegard: Should we update `conda.models.environment.Environment` to instead be a drop in replacement for `conda.env.env.Environment` so we can offer a cleaner deprecation path?

conda/misc.py
```diff
@@ -177,6 +159,48 @@ def explicit(
...
+    ]
+    if specs_with_missing_pcrecs:
+        if len(specs_with_missing_pcrecs) == len(specs_pcrecs):
+            raise SpecNotFoundInPackageCache("No package cache records found")
+        else:
+            missing_precs_list = ", ".join(specs_with_missing_pcrecs)
+            raise SpecNotFoundInPackageCache(
+                f"Missing package cache records for: {missing_precs_list}"
+            )
+    return [rec[1] for rec in specs_pcrecs]
```
kenodegard, line 182: Would `groupby_to_dict` help here?

```suggestion
    grouped_specs = groupby_to_dict(specs, lambda spec: next(PackageCacheData.query_all(spec), None) is not None)

    pcrecs_exists = grouped_specs.get(True) or []
    pcrecs_missing = grouped_specs.get(False) or []

    # Assert that every spec has a PackageCacheRecord
    if pcrecs_missing and not pcrecs_exists:
        raise SpecNotFoundInPackageCache("No package cache records found")
    elif specs_missing:
        missing_precs_list = ", ".join(pcrecs_missing)
        raise SpecNotFoundInPackageCache(
            f"Missing package cache records for: {missing_precs_list}"
        )
    return pcrecs_exists
```

conda/cli/main_env_create.py
soapy1: I don't hink we can flatten this. I think these 2 pieces of code have different intentions.

 In the first part of the `if` (`if not env.external_packages and not env.requested_packages`) represents the situation where the user is trying to install an empty environment. Conda will try to install an empty set of packages into the environment. 

That is:
```
pkg_specs = []
installer = get_installer(installer_type)
result[installer_type] = installer.install(prefix, pkg_specs, args, env)
```

the `else` part represents the situation where the user is trying to install some packages. So conda will install the set of packages defined in the file
```
result[installer_type] = installer.install(
    prefix, env.requested_packages, args, env
)
```

conda/env/env.py
soapy1: Good point, lets not do that

conda/env/specs/requirements.py
soapy1: I would strongly advocate for not updating `conda.models.environment.Environment` to be a drop in replacement for `conda.env.env.Environment`. In particular `conda.env.env.Environment` is meant to represent an `environment.yaml` file so has some features that we wouldn't want to carry over and support in `conda.models.environment.Environment`.

soapy1:
pre-commit.ci autofix

conda/env/specs/requirements.py
kenodegard: In that case I think we need to rename `conda.env.env.Environment` to `conda.env.env.EnvironmentYaml` or similar (I'd do that in a separate PR)

conda/misc.py
soapy1: This doesn't quite work as written because conda needs to be returning the package cache records instead of the matchspecs. Made some slight modifications and it works. But might be worth a second look.

soapy1:
pre-commit.ci autofix

conda/env/specs/requirements.py
jezdez: I like that

jezdez (dismissed):

conda/cli/main_env_create.py
kenodegard: right, but if there are no packages to install then `env.requested_packages` will be an empty list so it is the same thing? and if `env.external_packages` is empty the `for`-loop will be a noop?

soapy1: ah ya, you're right! will pull that change in

conda/misc.py
```diff
@@ -177,6 +160,54 @@ def explicit(
...
+
+    # Assert that every spec has a PackageCacheRecord
+    if missing and not existing:
+        raise SpecNotFoundInPackageCache("No package cache records found")
+    elif missing:
+        missing_precs_list = ", ".join([str(tup[0]) for tup in missing])
+        raise SpecNotFoundInPackageCache(
+            f"Missing package cache records for: {missing_precs_list}"
+        )
+    return [tup[1] for tup in existing]
```
kenodegard: Ah, I glossed over the 1 index, in that case I don't see any reason to use `groupby_to_dict`:

```suggestion
    existing = []
    missing = []
    for spec in specs:
        pcrec = next(PackageCacheData.query_all(spec), None)
        if pcrec is None:
            missing.append(spec)
        else:
            existing.append(pcrec)

    # Assert that every spec has a PackageCacheRecord
    if missing and not existing:
        raise SpecNotFoundInPackageCache("No package cache records found")
    elif missing:
        raise SpecNotFoundInPackageCache(
            f"Missing package cache records for: {', '.join(missing)}"
        )
    return existing
```

conda/misc.py
kenodegard: See https://github.com/conda/conda/pull/14937#discussion_r2178605100 for followup

conda/misc.py
soapy1: cool, I'll revert the commit that pulled in this change

soapy1:
pre-commit.ci autofix

jezdez (dismissed):

conda/env/specs/binstar.py
```diff
@@ -28,7 +29,7 @@ class BinstarSpec(EnvironmentSpecBase):
     """
     spec = BinstarSpec('darth/deathstar')
     spec.can_handle() # => True / False
-    spec.environment # => YAML string
+    spec.env # => YAML string
```
kenodegard:
```suggestion
    spec.environment # => YAML string
    spec.env # => conda.models.environment.Environment model
```

kenodegard (dismissed):

conda/env/specs/binstar.py
```diff
@@ -95,7 +97,8 @@ def file_data(self) -> list[dict[str, str]]:
         return [data for data in self.package["files"] if data["type"] == "env"]
 
     @cached_property
-    def environment(self) -> Environment:
+    @deprecated("25.9", "26.3", addendum="This method is not used anymore, use 'env'")
```
kenodegard: CEP9 calls for all deprecations to be pending for 2+ releases, since we're only one release away from the deprecation release in September we have to bump this to next year:

```suggestion
    @deprecated("26.3", "26.9", addendum="This method is not used anymore, use 'env'")
```

conda/env/specs/requirements.py
```diff
@@ -101,11 +104,12 @@ def can_handle(self) -> bool:
         return True
 
     @property
-    def environment(self) -> Environment:
+    @deprecated("25.9", "26.3", addendum="This method is not used anymore, use 'env'")
```
kenodegard:
```suggestion
    @deprecated("26.3", "26.9", addendum="This method is not used anymore, use 'env'")
```

conda/env/specs/yaml_file.py
```diff
@@ -47,7 +50,14 @@ def can_handle(self):
             return False
 
     @property
-    def environment(self):
+    @deprecated("25.9", "26.3", addendum="This method is not used anymore, use 'env'")
```
kenodegard:
```suggestion
    @deprecated("26.3", "26.9", addendum="This method is not used anymore, use 'env'")
```

news/14937-return-environment-model
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
```
kenodegard:
```suggestion
* Mark `conda.env.specs.binstar.BinstarSpec.environment` as pending deprecation in 26.9. Use `conda.env.specs.binstar.BinstarSpec.env` instead. (#14397)
* Mark `conda.env.specs.requirements.RequirementsSpec.environment` as pending deprecation in 26.9. Use `conda.env.specs.requirements.RequirementsSpec.env` instead. (#14397)
* Mark `conda.env.specs.yaml_file.YamlFileSpec.environment` as pending deprecation in 26.9. Use `conda.env.specs.yaml_file.YamlFileSpec.env` instead. (#14397)
```

kenodegard (changes requested):

conda/env/specs/requirements.py
soapy1: I've opened up another pr for renaming  `conda.env.env.Environment` to `conda.env.env.EnvironmentYaml` https://github.com/conda/conda/pull/14981

jezdez (approved):

kenodegard (approved):

PR #14942, opened 2025-06-16
Fixes bug related to `envvars_force_uppercase` setting

### LINKED ISSUE #14934: Setting `envvars_force_uppercase` to `false` lowercases other environment variables
### Checklist

### What happened?

When setting the `envvars_force_uppercase` to `false`:

```
conda config --set envvars_force_uppercase false
```

This changes to the activation code to output the following (`conda shell.posix activate`):

```
unset _CE_M
unset _CE_CONDA
PS1='(base) '
export path='/Users/username/miniconda3/bin:/Users/username/miniconda3/condabin'
export conda_prefix='/Users/username/miniconda3'
export conda_shlvl='1'
export conda_default_env='base'
export conda_prompt_modifier='(base) '
export CONDA_EXE='/Users/username/miniconda3/bin/conda'
export CONDA_PYTHON_EXE='/Users/username/miniconda3/bin/python'
```

You'll notice that environment variables like `PATH` and `CONDA_PREFIX` are lowercase which is not what we want. This causes all kinds of weird side-effects further on in the session. For example, look at my changing command prompt:

```
(test) root@7a7752c8176e:/# conda deactivate
(test) root@7a7752c8176e:/# conda activate base
(base) (test) root@7a7752c8176e:/# conda activate test
(test) (test) root@7a7752c8176e:/# conda activate base
(base) (test) (test) root@7a7752c8176e:/# conda activate test
(test) (test) (test) root@7a7752c8176e:/# 😬 
```

Furthermore, it also seems like this can break activation altogether when a variable has been set beforehand for the environment via:

```
conda env config vars set foo=bar --name test
```

This was the error I received on `osx-arm64` with `zsh` after trying to activate this environment:

```
(eval):export:4: path: inconsistent type for assignment
```

### Conda Info

```shell
active environment : base
            shell level : 1
 populated config files : /opt/conda/.condarc
                          /root/.condarc
          conda version : 25.3.1
         python version : 3.13.2.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=x86_64
                          __conda=25.3.1=0
                          __glibc=2.36=0
                          __linux=5.15.49=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/25.3.1 requests/2.32.3 CPython/3.13.2 Linux/5.15.49-linuxkit debian/12.10 glibc/2.36 solver/libmamba conda-libmamba-solver/25.4.0 libmambapy/2.0.5 aau/0.7.0 c/. s/. e/.
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> /opt/conda/.condarc <==
channels:
  - defaults

==> /root/.condarc <==
channels:
  - defaults
envvars_force_uppercase: False
```

### Conda list

```shell
n/a
```

### Additional Context

I tested against `25.5.x` and `25.3.x` and saw similar behavior.

travishathaway:
### Description

Closes: #14934 

### Checklist - did you ...


news/14934-bugfix-for-envvars-force-uppercase
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixes a bug discovered when setting `envvars_force_uppercase = False`
```
jaimergp:
```suggestion
* Fixes a bug discovered when setting `envvars_force_uppercase = False`. (#14934 via #14942)
```

ForgottenProgramme (dismissed):
Before the `envvars_force_uppercase` flag was introduced, `conda` was by default uppercasing all the env vars. Now that the uppercasing can be switched off, the env vars created and set in activate.py are no longer uppercased automatically (when this flag is set to `False`) and that causes the reported bug. 

I think this  PR adds a suitable fix for this bug. 👍🏼 

You also parametrized all the existing tests to ensure they pass in both cases: 
1. `envvars_force_uppercase` is set to `True`.
2. `envvars_force_uppercase` is set to `False`

jezdez:
@travishathaway Like we spoke earlier, here's a proposed integration test: https://github.com/travishathaway/conda/pull/4

tests/shell/__init__.py
```diff
@@ -288,3 +288,14 @@ def clear(self) -> None:
 
     def path_conversion(self, *args, **kwargs):
         return self.activator.path_conversion(*args, **kwargs)
+
+
+def get_env_var_safe(sh: InteractiveShell, varname: str) -> str:
+    """Safely get environment variable, handling different shell behaviors."""
+    try:
+        value = sh.get_env_var(varname, "")
+        value = None if "Undefined variable" in value else value
```
travishathaway: Special exception for `csh` and `tcsh`.

travishathaway:
@jezdez,

Thanks for those suggestions! I didn't realize we had such nice classes setup already for shell integration tests 🫶.

I basically plucked the changes you made and committed them directly to this branch. I had to make a couple updates because of some failing tests, so it would be worth it for your take a another look.

travishathaway:
@jezdez,

I saw some timeout issues in the test runners 😦. I bumped the timeout up to 60 seconds. If this doesn't work, we might have to rethink our approach here or further debug why the tests are taking so long.

travishathaway:
Timeout issues are continuing to happen even after bumping it up to 60 seconds 😭. I'm not sure if bumping this up to a very high number is practical given it would probably significantly increase the time our tests need to complete.

travishathaway:
Looks like it's only happening for `xonsh`, `fish` and `powershell`.

jezdez:
Let's remove those tests then

travishathaway:
@jezdez,

Reset the pull request to the way it originally was but kept your additional change regarding unsetting environment variables.

jezdez (approved):

conda/activate.py
```diff
@@ -533,11 +533,17 @@ def build_deactivate(self):
...
 
+        # Handle environment variables that need to be unset during deactivation
         for env_var in old_conda_environment_env_vars.keys():
             if save_value := os.getenv(f"__CONDA_SHLVL_{new_conda_shlvl}_{env_var}"):
                 export_vars[env_var] = save_value
             else:
-                unset_vars.append(env_var)
+                # Apply case conversion for environment variables that need to be unset
+                if context.envvars_force_uppercase:
+                    unset_vars.append(env_var.upper())
```
ForgottenProgramme, line 543: if `envvars_force_uppercase` is set to true,  why do we need to `.upper()`?

conda/activate.py
```diff
@@ -533,11 +533,17 @@ def build_deactivate(self):
...
+                else:
+                    unset_vars.append(env_var)
```
ForgottenProgramme, line 545:
```suggestion
                if context.envvars_force_uppercase:
                    unset_vars.append(env_var)
                else:
                    unset_vars.append(env_var.upper())
```

conda/activate.py
travishathaway: We do that here to ensure it's uppercase. When the `envvars_force_uppercase` is `True` everything has to be uppercase. When it's `False`, we allow the variable's case to be whatever the user has designated.

Issue #12013, opened 2022-10-25
Issues with Proxy 407 handling in 22.9.0


byjrack:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

We use authenticated proxies for egress to public Internet.  The Basic auth does work user:pass@proxy:port, but all sorts of bad on storing secrets in plaintext on the machine and they come up in DLP scans.

We support ntlm/negotiate which is the normal SSO model, but I don't see the hooks in conda for requests to support that type of response.

oob
```
Collecting package metadata (current_repodata.json): failed

ProxyError: Conda cannot proceed due to an error in your proxy configuration.
Check for typos and other configuration errors in any '.netrc' file in your home directory,
any environment variables ending in '_PROXY', and any other system-wide proxy
configuration settings.
```

with verbosity=3
```
DEBUG urllib3.util.retry:increment(594): Incremented Retry for (url='/pkgs/r/osx-64/current_repodata.json'): Retry(total=2, connect=None, read=None, redirect=None, status=None)
WARNING urllib3.connectionpool:urlopen(812): Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 407 Proxy Authentication Required'))': /pkgs/r/osx-64/current_repodata.json
```
[The session code](
https://github.com/conda/conda/blob/main/conda/gateways/connection/session.py#L141) seems to have a hook on the request to catch the 407 response and handle an interactive input for Basic, but it's not being invoked.  Based on the trace it's not even getting out of the core lib.

I am going to try and do some breakpoint testing and trace the flows, but is this on the radar?

 

### Conda Info

```shell
DEBUG conda.gateways.logging:set_verbosity(236): verbosity set to 3

     active environment : None
       user config file : /Users/user/.condarc
 populated config files : /Users/user/.condarc
          conda version : 22.9.0
    conda-build version : not installed
         python version : 3.9.12.final.0
       virtual packages : __osx=10.16=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /usr/local/Caskroom/miniconda/base  (writable)
      conda av data dir : /usr/local/Caskroom/miniconda/base/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/main/noarch
[...]

PR #14945, opened 2025-06-17
Enhance `ProxyError` to accept custom messages and add tests for default and custom error messages

jezdez:
### Description

While looking at some proxy related issue reports from users, I noticed that one case of `ProxyError` was being instantiated with an argument, that it doesn't actual support. Here's the call in `session.py`: https://github.com/conda/conda/blob/fbeec0b3ebc756e1cfc1339bfa0c2343079fac15/conda/gateways/connection/session.py#L324-L334

Refs #12013?

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -474,15 +474,16 @@ class CondaOSError(CondaError, OSError):
 
 
 class ProxyError(CondaError):
-    def __init__(self):
-        message = dals(
+    def __init__(self, message: str | None = None):
+        if message is None:
+            message = dals(
+                """
+            Conda cannot proceed due to an error in your proxy configuration.
+            Check for typos and other configuration errors in any '.netrc' file in your home directory,
+            any environment variables ending in '_PROXY', and any other system-wide proxy
+            configuration settings.
             """
-        Conda cannot proceed due to an error in your proxy configuration.
-        Check for typos and other configuration errors in any '.netrc' file in your home directory,
-        any environment variables ending in '_PROXY', and any other system-wide proxy
-        configuration settings.
-        """
-        )
+            )
         super().__init__(message)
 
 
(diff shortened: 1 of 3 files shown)


conda/exceptions.py
```diff
@@ -474,15 +474,16 @@ def __init__(self, message: str, **kwargs):
...
-        message = dals(
+    def __init__(self, message: str | None = None):
+        if message is None:
+            message = dals(
+                """
+            Conda cannot proceed due to an error in your proxy configuration.
+            Check for typos and other configuration errors in any '.netrc' file in your home directory,
+            any environment variables ending in '_PROXY', and any other system-wide proxy
+            configuration settings.
             """
```
kenodegard:
```suggestion
                Conda cannot proceed due to an error in your proxy configuration.
                Check for typos and other configuration errors in any '.netrc' file in your home directory,
                any environment variables ending in '_PROXY', and any other system-wide proxy
                configuration settings.
                """
```

tests/test_exceptions.py
```diff
@@ -803,3 +804,22 @@ def test_PackagesNotFoundError_use_only_tar_bz2(
...
             channel_urls=["https://repo.anaconda.org/pkgs/main"],
         )
+
+
+def test_proxy_error_default_message() -> None:
+    """Test ProxyError with default and custom messages."""
+    # Test default message
+    exc_default = ProxyError()
+    default_message = str(exc_default)
+    assert "proxy configuration" in default_message.lower()
```
ForgottenProgramme: Why do we need the `.lower()` in this assert?

jezdez:
```suggestion
    assert "proxy configuration" in default_message
```

jezdez: No good reason!

ForgottenProgramme (approved):

PR #11794, opened 2022-09-02
Provide `--platform` / `--subdir` flag to create (and remember) an environment for a platform different than native

jaimergp:
### Description

With the release of M1 processors in Apple products, a common question is "how do I create an Intel environment in my M1 laptop?". This is technically possible today but involves two (obscure) steps:

1. Set `CONDA_SUBDIR` to trick `conda` into thinking we are in a different platform.
2. Make `conda` remember that choice _after_ environment creation and activation via `conda config --env --set subdir <platform>`.

This PR simplifies that setup into a single step: `conda create` and `conda env create` will now accept a `--platform` CLI flag that automates the process.

This follows these precedents:

* `conda search` supports a `--platform` / `--subdir` flag to find packages in other platforms
* `micromamba create` [supports](https://github.com/mamba-org/mamba/blob/0cc361562a3d0c712d042f1c10c790eb8133a366/libmamba/src/api/create.cpp#L79-L90) a `--platform` flag, the value of which will get saved to `$PREFIX/.mambarc`.

This is how the `conda create` CLI help would look like:

```
$ conda create --help
usage: conda create [-h] [--clone ENV] (-n ENVIRONMENT | -p PATH) [-c CHANNEL] [--use-local]
                    [--override-channels] [--repodata-fn REPODATA_FNS] [--strict-channel-priority]
                    [--no-channel-priority] [--no-deps | --only-deps] [--no-pin] [--copy] [-C] [-k]
                    [--offline] [-d] [--json] [-q] [-v] [-y] [--download-only]
                    [--show-channel-urls] [--file FILE] [--no-default-packages]
                    [--experimental-solver {classic,libmamba,libmamba-draft}] [--subdir SUBDIR]
                    [--dev]
                    [package_spec [package_spec ...]]

Create a new conda environment from a list of specified packages. To use the created environment, use 'conda activate envname' to look in that directory first.  This command requires either the -n NAME or -p PREFIX option.

Options:

positional arguments:
  package_spec          Packages to install or update in the conda environment.

optional arguments:
  -h, --help            Show this help message and exit.
  --clone ENV           Path to (or name of) existing local environment.
  --file FILE           Read package versions from the given file. Repeated file specifications can
                        be passed (e.g. --file=file1 --file=file2).
  --dev                 Use `sys.executable -m conda` in wrapper scripts instead of CONDA_EXE. This
                        is mainly for use during tests where we test new conda source against old
                        Python versions.

Target Environment Specification:
[...]

PR #14956, opened 2025-06-23
Fixing problem with `--platform` option for `conda env create`

### LINKED ISSUE #14949: Support target platform specification (e.g., --platform/--subdir) in conda env update and conda update commands
### Checklist

### What is the idea?

Support user specification of a target platform when trying to update packages in an environment.

### Why is this needed?

Since at least version [24.3.0](https://github.com/conda/conda/releases/tag/24.3.0), users have been able to create environments with packages compiled for different platforms (architectures). Specifically, `conda env create` and `conda create` commands have supported `--platform/--subdir` options. This is particularly useful for Apple Silicon (ARM-based) Mac users (`osx-arm64` platform) trying to install packages only available for the `osx-64` platform, which can still be run on ARM-based Macs through Rosetta.

However, no similar target platform options currently exist for updating packages, whether via `conda env update` or `conda update`.

### What should happen?

`conda env update` and `conda update` should support `--platform` and/or `--subdir` options.

### Additional Context

There are currently at least 2 workarounds for updating packages targeting a different platform. Here, I assume that the environment.yaml file has been updated since the environment was initially created.

1. Remove and re-create the environment.

   ```bash
   conda env remove -y -n myenv

   # example targeting the osx-64 platform
   conda env create --platform osx-64 -f environment.yaml
   ```

2. Set the `subdir` conda config variable.

   - This was the main way to *create* environments targeting different platforms in older versions of conda, which was superseded by the `--platform` or `--subdir` options to `conda env create` and `conda create`. Unfortunately, this remains necessary for updating packages (if you don't want to remove and re-create the environment).

   ```bash
   # example targeting the osx-64 platform
   config --env --set subdir osx-64

   conda env update -f environment.yaml
   ```

travishathaway:
### Description

Closes: #14949 

Currently, when users run `conda env create --platform=<platform> ...` this will initially use the correct platform because `context.subdir` is being set correctly, but subsequent calls to `conda env update` or `conda install` will use erroneously use the system's platform.

This happens because we are not creating the `.condarc` file in the environment itself which has the `subdir` set correctly in it. To fix this, I simply copied over the code from `conda/cli/install.py` where this was being done correctly to `conda/cli/main_env_create.py`.

### Checklist - did you ...


news/14956-bugfix-for-env-create-platform
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixes issue where `--platform` option was not being respected on subsequent environment operations after initial `conda env create` call (#14949 via #14956)
```
jezdez:
```suggestion
* Respect `--platform` option on subsequent environment operations after initial `conda env create` call. (#14949 via #14956)
```

tests/env/test_create.py
```diff
@@ -36,6 +36,30 @@ def get_env_vars(prefix):
...
+    env_file.write_text(
+        f"""
+        name: test-env
+        channels:
+          - {test_recipes_channel}
+        dependencies:
+          - dependency
+        """
+    )
+    return env_file
```
jezdez: This seems like a lot of effort for one test case, unless you plan to use it elsewhere, inlining in the test is worth it.

conda/cli/main_env_create.py
```diff
@@ -209,6 +211,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
                         )
                     )
 
+        if context.subdir != context._native_subdir():
+            set_keys(
+                ("subdir", context.subdir),
+                path=Path(prefix, ".condarc"),
+            )
```
jezdez, line 218: Hmm, I'm not sure if that is what should be done, this writes to the environment and doesn't inform users about this.

jezdez (commented):
I don't think writing implicitly to the config file is okay when running `conda env create` with ``--platform``, at least the users should be informed about this change of behavior.

conda/cli/main_env_create.py
jaimergp: This is how we did it in https://github.com/conda/conda/pull/11794

jezdez: Ah, true, thanks for the pointer!

jezdez (approved):

PR #14951, opened 2025-06-19
Add file locking health check

ForgottenProgramme:
### Description

This PR adds a new health check to `conda doctor` called the "File locking health check". 
This checks whether file locking is supported on a system or not. 
The check is based on the logic in the `gateways/disk/lock.py` file:

- it checks for existence of `msvcrt` package on `windows` systems
- it checks for `fcntl` in UNIX like systems

Based on the presence or absence of these packages, it concludes whether file locking is supported or not. 

Alternative to #14957
Closes #14957

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #14957, opened 2025-06-23
Add `file locking` health check (use lock.py file logic)

ForgottenProgramme:
### Description

Second PR for "adding file locking health check to conda doctor"

Alternative to #14951
Closes #14951

### Checklist - did you ...

### DIFF
--- a/conda/gateways/disk/lock.py
+++ b/conda/gateways/disk/lock.py
@@ -23,8 +23,6 @@ def _lock_noop(fd):
     yield
 
 
-_lock_noop.locking = False
-
 try:  # pragma: no cover
     import msvcrt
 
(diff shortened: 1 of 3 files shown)


kenodegard (changes requested):
I prefer this implementation over https://github.com/conda/conda/pull/14951 since it includes checks for `context.no_lock`.

Since file locking behavior varies across filesystems, we cannot rely solely on importing `fcntl`/`msvcrt` or testing lock acquisition with a temporary file. A successful lock on a temporary file doesn't guarantee that we can acquire locks in other locations like the package cache, as the temporary file might reside on a different mount or drive. Instead, we need to test lock acquisition capability in each specific location where conda will attempt to create locks.

We also need to add a timeout if the lock cannot be immediately acquired so the health check wont hang.

ForgottenProgramme:
Closing this in favour of https://github.com/conda/conda/pull/15114

PR #14963, opened 2025-06-25
Separate requirements.txt and explicit env specs

### LINKED ISSUE #14684: conda should not allow mixing different types of input `file`s when creating an environment
### Checklist

### What happened?

conda `create` and `install` commands support a `--file` argument which is a list of files to install. Files can either be [explicit files or regular text files](https://github.com/conda/ceps/pull/79). If there is a mix of `--file`s that specify regular packages and explicit package then this will produce a `ParseError`. For example, consider the following two environment files:

env1.txt (explicit file)
```
@EXPLICIT
https://repo.anaconda.com/pkgs/main/linux-64/_libgcc_mutex-0.1-main.conda
https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda
```

env2.txt (regular text file)
```
numpy
```

Running a `conda create` against these files individually works. Providing both files as arguments produces an error:
```
$ conda create -n testenv-mixed --file env1.txt --file env2.txt 

ParseError: Could not parse explicit URL: file:///home/sophia/projects/conda/numpy
```

It is correct to raise an error for this situation, but it would be helpful if the error more accurately described what is going wrong and how to fix it. For example, it would be better for the result to be something like

```
$ conda create -n testenv-mixed --file env1.txt --file env2.txt 

CondaError: can not mix `file` types

env1.txt is found to be an explicit type file
env2.txt is found to be requirements.txt type file
```

---

### Checklist

- [x] https://github.com/conda/conda/pull/14725
- [x] https://github.com/conda/conda-environments-project-mgmt/issues/47

soapy1:
### Description

This PR introduces an env spec to represent an explicit environment. An explicit environment is one that lists out the full set of dependencies for an environment as a set of records. It is unrelated to a requirements.txt environment spec. However, both these specs might share a `.txt` file extension.

A user can export an explicit environment by running `conda list --explicit`. This will produce a file like:

```
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: linux-64
# created-by: conda 25.5.2.dev36+g5c3f32e7f
@EXPLICIT
https://repo.anaconda.com/pkgs/main/linux-64/_libgcc_mutex-0.1-main.conda
https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda
https://repo.anaconda.com/pkgs/main/linux-64/ld_impl_linux-64-2.40-h12ee557_0.conda
https://repo.anaconda.com/pkgs/main/linux-64/libstdcxx-ng-11.2.0-h1234567_1.conda
https://repo.anaconda.com/pkgs/main/linux-64/nlohmann_json-3.11.2-h6a678d5_0.conda
. . .
```

Fixes https://github.com/conda/conda/issues/14684#issuecomment-3004772951

### Checklist - did you ...

### DIFF
--- a/conda/env/specs/requirements.py
+++ b/conda/env/specs/requirements.py
@@ -20,11 +20,6 @@ class RequirementsSpec(EnvironmentSpecBase):
     """
     Reads dependencies from requirements files (including explicit files)
     and returns an Environment object from it.
-
-    This unified spec handles both regular requirements.txt files and explicit
-    environment files marked with @EXPLICIT. The Environment class automatically
-    detects whether dependencies are explicit based on the presence of the
-    @EXPLICIT marker.
     """
 
     msg: str | None = None
@@ -75,7 +70,7 @@ class RequirementsSpec(EnvironmentSpecBase):
             return True
         self.msg = "The environment does not have a name"
         return False
-
+    
     def can_handle(self) -> bool:
         """
         Validates that this spec can process the environment definition.
@@ -97,6 +92,12 @@ class RequirementsSpec(EnvironmentSpecBase):
             self.msg = f"File {self.filename} does not have a supported extension: {', '.join(self.extensions)}"
             return False
 
+        # Ensure this is not an explicit file. Requirements.txt and explicit files
+        # may sometimes share file extension.
+        dependencies_list = list(yield_lines(self.filename))
+        if "@EXPLICIT" in dependencies_list:
+            return False
+
         return True
 
     @property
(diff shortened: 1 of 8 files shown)


soapy1:
pre-commit.ci autofix

conda/env/specs/requirements.py
```diff
@@ -97,15 +92,20 @@ def can_handle(self) -> bool:
             self.msg = f"File {self.filename} does not have a supported extension: {', '.join(self.extensions)}"
             return False
 
+        # Ensure this is not an explicit file. Requirements.txt and explicit files
```
soapy1, line 95: I would say in general, the approach of matching just the file extension to check if an env spec plugin can parse an environment spec is not sufficient. I would call this a "greedy plugin". Punting on fixing this issue in this PR since other built in plugins also have this issue. We'll want to tackle that in a more thoughtful way soon.

jezdez (approved):

PR #14968, opened 2025-06-26
Stop using `native_path_to_unix`

kenodegard:
### Description

Replace deprecated `conda.activate.native_path_to_unix` with `conda.common.path.win_path_to_unix`.

Update tests to mirror codebase.

Extracted from https://github.com/conda/conda/pull/12771

### Checklist - did you ...


tests/common/_os/test_windows.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from conda.common._os.windows import is_admin_on_windows
+from conda.common.compat import on_win
+
+
+def test_is_admin_on_windows():
+    result = is_admin_on_windows()
+    if on_win:
+        assert result is False or result is True
```
jezdez: Huh, both can be true?

kenodegard: 🤔 not sure why, I just moved this test from a different location

ifitchet: Schrödinger never really gained traction in the testing world...

conda/core/initialize.py
```diff
@@ -1403,9 +1403,9 @@ def install_conda_csh(target_path, conda_prefix):
 def _config_fish_content(conda_prefix):
     conda_exe = join(conda_prefix, BIN_DIRECTORY, "conda.exe" if on_win else "conda")
     if on_win:
-        from ..activate import native_path_to_unix
```
dholth, line 1406: Should "if on_win" go inside the path_to_unix function?

kenodegard: Good question but I think that's out of scope here

The `*_path_to_*` functions includes fallback logic when `cygpath` isn't available so these functions actually work fine on non-Windows too (do we need them to? probably not)

jezdez (approved):

jezdez:
Let's open new tickets for the suggested changes

PR #14971, opened 2025-06-26
[FIX] Add unknown Arch when `platform.machine() == 'unknown'`

### LINKED ISSUE #14969: [BUG] `conda.models.enums.Arch.from_sys` is broken when `platform.machine() == "AMD64"`
### Checklist

### What happened?

Calling `conda.models.enums.Arch.from_sys()` on Windows fails. `from_sys()` calls `platform.machine()`, which returns `"AMD64"`; this key is then used to index the `Arch` class; since `AMD64` isn't a valid member, a `KeyError` is thrown: https://github.com/conda/conda/blob/6b926e7d08efcb1dc9d7d78b64f67a2eeeda420c/conda/models/enums.py#L35.

From the docs for `platform.machine()`

> Returns the machine type, e.g. 'AMD64'. An empty string is returned if the value cannot be determined.

So I think we should also add a way to handle unknown platforms as well.

### Conda Info

```shell
active environment : None
       user config file : /home/pdmurray/.condarc
 populated config files : /opt/miniforge/.condarc
                          /home/pdmurray/.condarc
          conda version : 25.3.2.dev94
         python version : 3.12.10.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=zen2
                          __conda=25.3.2.dev94=0
                          __glibc=2.41=0
                          __linux=6.12.34=0
                          __unix=0=0
       base environment : /opt/miniforge  (read only)
      conda av data dir : /opt/miniforge/etc/conda
  conda av metadata url : None
          package cache : /opt/miniforge/pkgs
                          /home/pdmurray/.conda/pkgs
       envs directories : /home/pdmurray/.conda/envs
                          /opt/miniforge/envs
               platform : linux-64
             user-agent : conda/25.3.2.dev94 requests/2.32.3 CPython/3.12.10 Linux/6.12.34-1-lts arch/rolling glibc/2.41 solver/libmamba conda-libmamba-solver/25.3.0 libmambapy/2.1.1
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
This is configuration-independent.
```

### Conda list

```shell

```

### Additional Context

_No response_

peytondmurray:
### Description

This PR modifies `conda.models.enums.Arch` to add a few new members, and so that `Arch.from_sys()` no longer raises an unhandled `KeyError` when `platform.machine()` isn't an `Arch` member name. Closes #14969.

In detail:

- Added an `Arch.unknown` member as a catch-all for unknown machine types. From the [python docs](https://docs.python.org/3/library/platform.html#platform.machine), `platform.machine()` can be 'unknown'.
- Added a `Arch.AMD64` member, since the python docs also say that's a possibility, and because on Windows github runners that's the return value of `platform.machine()`.

A few questions for maintainers:
- Is this considered a "significant" change? I'm happy to add a news item if so.
- I'm also happy to add tests. I've left them out for now because we're not currently testing anything else in `conda.models.enums.Arch`, and this change is trivial, but I can easily add some if that would be good.

### Checklist - did you ...


conda/models/enums.py
```diff
@@ -15,6 +15,7 @@
 class Arch(Enum):
     x86 = "x86"
     x86_64 = "x86_64"
+    AMD64 = "AMD64"
```
jaimergp: I think this should be x86_64 too, they are equivalent.

peytondmurray:
I don't _think_ the failure is related - will try updating the branch now to retrigger tests.

conda/models/enums.py
```diff
@@ -25,14 +26,18 @@ class Arch(Enum):
...
 
     @classmethod
     def from_sys(cls):
         if sys.platform == "zos":
             return cls["z"]
-        return cls[machine()]
+        try:
+            return cls(machine())
+        except ValueError:
+            return Arch.unknown
```
kenodegard, line 40: I'm not sure we want to blindly cast all unknown machine types to `Arch.unknown`, per docs:

> An empty string is returned if the value cannot be determined.

What if we only cast empty strings to `Arch.unknown`? If `platform.machine()` returns something but we haven't added it here yet I'm thinking raising an error would be more informative instead of treating it as unknown? Or maybe we should treat it as something else, e.g., `Arch.unhandled` `Arch.unmapped` `Arch.unsupported`?

peytondmurray: I sifted through the repo looking for use cases that could inform the decision to go this route, and I realized this is actually not being used anywhere. I'm fine making this fail with a helpful error message, it may encourage future users to add any platforms that we've forgotten here. I'll wait for some others to weigh in on this before making the change, but I like the idea :+1:

peytondmurray:
After a discussion with @soapy1, I'm going to close this one out. Originally I intended to use `conda.models.enums.Arch` and `conda.models.enums.Platform` to get system information for use in `conda-declarative`, but it seems like

1. `Arch` isn't used anywhere and should probably be removed anyway. It has somewhat overlapping functionality with `conda.base.constants.PLATFORMS`.
2. The preferred way of getting the current platform is to use `context.subdir` (???), which in some cases is a valid `conda.base.constants.PLATFORMS`. I'll leave sorting out the `subdir`/`PLATFORMS` relationship for another issue.

PR #14978, opened 2025-07-01
Minor environment specifier fixes

kenodegard:
### Description

Type checker was complaining about the return types so I've adjusted the if-clauses accordingly.

Switched to using `self.get_environment_specifiers()` instead of `self.get_hook_results("environment_specifiers")` since the former does plugin name normalization (lowercase).

Correct how we collect autodetect disabled plugins for error reporting.

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1314,7 +1314,7 @@ class EnvironmentSpecPluginNotDetected(SpecNotFound):
 
             """
         )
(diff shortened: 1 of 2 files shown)


soapy1 (approved):
Looks great!

ForgottenProgramme (approved):

jezdez (approved):

PR #14980, opened 2025-07-02
`conda.gateways.disk.lock` raises `LockError` instead of `OSError`

### LINKED ISSUE #15077: Add tests for `gateways/disk/lock.py` module
### Checklist

### What happened?

`gateways/disk/lock.py` is not tested

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

ForgottenProgramme:
### Description

resolves #15077 

This PR:
1. Adds a tests file for `gateways/disk/lock.py`.
2. Uses `LockError` in the `lock.py` module.
3. Modifies the `lock()` function/class to accept "number of lock attempts" variable, default is set to 10 for backwards compatibility. 

Tests added:
1. That a LockError is displayed when the locking functions in either `msvcrt` or `fcntl` generate an OSError. 
4. Happy path case (locking works)
5. When a process tries to lock an already locked file (LockError generated, operation not allowed). 

### Checklist - did you ...

### DIFF
--- a/conda/gateways/disk/lock.py
+++ b/conda/gateways/disk/lock.py
@@ -11,6 +11,7 @@ import warnings
 from contextlib import contextmanager
 
 from ...base.context import context
+from ...exceptions import LockError
 
 LOCK_BYTE = 21  # mamba interop
 LOCK_ATTEMPTS = 10
@@ -30,13 +31,16 @@ try:  # pragma: no cover
     def _lock_impl(fd):  # type: ignore
         tell = fd.tell()
         fd.seek(LOCK_BYTE)
-        msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK, 1)  # type: ignore
         try:
-            fd.seek(tell)
-            yield
-        finally:
-            fd.seek(LOCK_BYTE)
-            msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+            msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK, 1)  # type: ignore
+            try:
+                fd.seek(tell)
+                yield
+            finally:
+                fd.seek(LOCK_BYTE)
+                msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+        except OSError as e:
+            raise LockError("Failed to acquire lock.") from e
 
 except ImportError:
     try:
@@ -61,9 +65,9 @@ except ImportError:
                             self.fd, fcntl.LOCK_EX | fcntl.LOCK_NB, 1, LOCK_BYTE
                         )
                         break
-                    except OSError:
+                    except OSError as e:
                         if attempt > LOCK_ATTEMPTS - 2:
-                            raise
+                            raise LockError("Failed to acquire lock.") from e
                         time.sleep(LOCK_SLEEP)
 
             def __exit__(self, *exc):
--- a/conda/gateways/disk/lock.py
+++ b/conda/gateways/disk/lock.py
@@ -43,6 +43,9 @@ try:  # pragma: no cover
         except OSError as e:
             if e.errno in (errno.EACCES, errno.EAGAIN):
                 raise LockError("Failed to acquire lock.") from e
+            else:
+                raise
+
 
 except ImportError:
     try:
@@ -71,6 +74,8 @@ except ImportError:
                         if e.errno in (errno.EACCES, errno.EAGAIN):
                             if attempt > LOCK_ATTEMPTS - 2:
                                 raise LockError("Failed to acquire lock.") from e
+                        else:
+                            raise
                         time.sleep(LOCK_SLEEP)
 
             def __exit__(self, *exc):
--- a/conda/gateways/disk/lock.py
+++ b/conda/gateways/disk/lock.py
@@ -14,12 +14,11 @@ from ...base.context import context
 from ...exceptions import LockError
 
 LOCK_BYTE = 21  # mamba interop
-LOCK_ATTEMPTS = 10
 LOCK_SLEEP = 1
 
 
 @contextmanager
-def _lock_noop(fd):
+def _lock_noop(fd, lock_attempts):
     """When locking is not available."""
     yield
 
@@ -28,7 +27,7 @@ try:  # pragma: no cover
     import msvcrt
 
     @contextmanager
-    def _lock_impl(fd):  # type: ignore
+    def _lock_impl(fd, lock_attempts):  # type: ignore
         tell = fd.tell()
         fd.seek(LOCK_BYTE)
         try:
@@ -59,11 +58,12 @@ except ImportError:
     else:
 
         class _lock_impl:
-            def __init__(self, fd):
+            def __init__(self, fd, lock_attempts):
                 self.fd = fd
+                self.lock_attempts = lock_attempts
 
             def __enter__(self):
-                for attempt in range(LOCK_ATTEMPTS):
+                for attempt in range(self.lock_attempts):
                     try:
                         # msvcrt locking does something similar
                         fcntl.lockf(
@@ -71,7 +71,7 @@ except ImportError:
                         )
                         break
                     except OSError:
-                        if attempt > LOCK_ATTEMPTS - 2:
+                        if attempt > self.lock_attempts - 2:
                             raise LockError("Failed to acquire lock.")
                         time.sleep(LOCK_SLEEP)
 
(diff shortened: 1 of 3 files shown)


conda/gateways/disk/lock.py
```diff
@@ -30,13 +31,16 @@ def _lock_noop(fd):
...
-            msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+            msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK, 1)  # type: ignore
+            try:
+                fd.seek(tell)
+                yield
+            finally:
+                fd.seek(LOCK_BYTE)
+                msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+        except OSError as e:
+            raise LockError("Failed to acquire lock.") from e
```
kenodegard: The reason the test is failing is because we are catching ALL `OSErrors` that occur, both those raised during locking but also those raised within the context.

We need the following to raise an `OSError` but currently this will raise a `LockError` even if the locking was successful:
```python
with lock(...):
    raise OSError
```

kenodegard (changes requested):

conda/gateways/disk/lock.py
```diff
@@ -30,13 +31,16 @@ def _lock_noop(fd):
...
-            msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+            msvcrt.locking(fd.fileno(), msvcrt.LK_LOCK, 1)  # type: ignore
+            try:
+                fd.seek(tell)
+                yield
+            finally:
+                fd.seek(LOCK_BYTE)
+                msvcrt.locking(fd.fileno(), msvcrt.LK_UNLCK, 1)  # type: ignore
+        except LockError as e:
+            print("Failed to acquire lock.", e)
```
kenodegard: This exception will never happen

`msvcrt.locking` does not raise a `LockError`, it raises an `OSError`

conda/gateways/disk/lock.py
```diff
@@ -61,13 +70,16 @@ def __enter__(self):
                             self.fd, fcntl.LOCK_EX | fcntl.LOCK_NB, 1, LOCK_BYTE
                         )
                         break
-                    except OSError:
+                    except OSError as e:
                         if attempt > LOCK_ATTEMPTS - 2:
-                            raise
+                            raise LockError("Failed to acquire lock.") from e
```
kenodegard: `as e`/`from e` is the default stack trace linking so we don't need to do that here

conda/gateways/disk/lock.py
```diff
@@ -61,13 +70,16 @@ def __enter__(self):
...
                         time.sleep(LOCK_SLEEP)
 
             def __exit__(self, *exc):
-                fcntl.lockf(self.fd, fcntl.LOCK_UN, 1, LOCK_BYTE)
+                try:
+                    fcntl.lockf(self.fd, fcntl.LOCK_UN, 1, LOCK_BYTE)
+                except OSError as e:
+                    raise LockError("Failed to release lock.") from e
```
kenodegard: ditto

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,110 @@
...
+
+def test_locking_not_supported(monkeypatch: MonkeyPatch, tmp_path):
+    tmp_file = tmp_path / "testfile"
+    tmp_file.write_bytes(b"test")
+    monkeypatch.delitem(sys.modules, "fcntl", raising=False)
+    monkeypatch.delitem(sys.modules, "msvcrt", raising=False)
+    monkeypatch.setattr(builtins, "__import__", monkey_import_ImportError)
+
+    with pytest.raises(ImportError):
+        lock(tmp_file)
```
kenodegard: This test doesn't verify what we intended.

When we import from `conda.gateways.disk.lock`, both `fcntl` and `msvcrt` are already imported. Any import errors would have been raised and handled at that point. The current test is only checking if we can mock an import error when accessing these modules.

Additionally, the `conda.gateways.disk.lock` module is designed to output a warning that locking is not supported rather than raising an `ImportError`. We should refactor this test to verify that the appropriate warning is issued instead.

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,110 @@
...
+):
+    tmp_file = tmp_path / "testfile"
+    tmp_file.write_bytes(b"test")
+
+    mocker.patch("fcntl.lockf", side_effect=OSError)
+    monkeypatch.setattr("conda.gateways.disk.lock.LOCK_ATTEMPTS", 1)
+    with pytest.raises(LockError):
+        with tmp_file.open("r+b") as f:
+            with lock(f):
+                pass
```
kenodegard: These can be combined:

```suggestion
def test_LockError_raised(
    mocker: MockerFixture,
    monkeypatch: MonkeyPatch,
    tmp_path: Path,
):
    (tmp_file := tmp_path / "testfile").touch()

    mocker.patch("msvcrt.locking" if on_win else "fcntl.lockf", side_effect=OSError)
    monkeypatch.setattr("conda.gateways.disk.lock.LOCK_ATTEMPTS", 1)

    with pytest.raises(LockError):
        with tmp_file.open("r+b") as fh, lock(fh):
            pass
```

tests/gateways/disk/test_lock.py
ForgottenProgramme: Maybe we can continue asserting that an ImportError is raised and then also assert that the appropriate warning is raised?

ForgottenProgramme:
woohoo! 🎉 
The tests are finally passing. @kenodegard !

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,65 @@
...
+from conda.exceptions import LockError
+from conda.gateways.disk.lock import (
+    _lock_impl,
+    lock,
+)
+
+
+def test_LockError_raised(
+    mocker: MockerFixture, monkeypatch: MonkeyPatch, tmp_path: Path
+):
```
kenodegard:
```suggestion
def test_LockError_raised(mocker: MockerFixture, tmp_path: Path):
```

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+from pytest import MonkeyPatch
+from pytest_mock import MockerFixture
+
+from conda.common.compat import on_win
+from conda.exceptions import LockError
+from conda.gateways.disk.lock import (
+    _lock_impl,
+    lock,
+)
```
kenodegard:
```suggestion
from conda.gateways.disk.lock import _lock_impl, lock
```

kenodegard (changes requested):

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_LockError_raised(mocker: MockerFixture, tmp_path: Path):
+    tmp_file = tmp_path / "testfile"
+    tmp_file.touch()
+
+    mocker.patch("msvcrt.locking" if on_win else "fcntl.lockf", side_effect=OSError)
+    with pytest.raises(LockError):
+        with tmp_file.open("r+b") as f:
+            with lock(f, lock_attempts=1):
```
kenodegard:
```suggestion
            with _lock_impl(f, lock_attempts=1):
```

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_lock_acquired_success(mocker: MockerFixture, tmp_path: Path):
+    tmp_file = tmp_path / "testfile"
+    tmp_file.touch()
+
+    mocker.patch("conda.gateways.disk.lock.lock", return_value=_lock_impl)
+
+    with tmp_file.open("r+b") as f:
+        with lock(f):
```
kenodegard:
```suggestion
def test_lock_acquired_success(tmp_path: Path):
    tmp_file = tmp_path / "testfile"
    tmp_file.touch()

    with tmp_file.open("r+b") as f:
        with _lock_impl(f, lock_attempts=1):
```

tests/gateways/disk/test_lock.py
```diff
@@ -0,0 +1,60 @@
...
+            # Because we are able to use lock(), that means lock acquisition succeeded.
+            pass
+
+
+def lock_wrapper(path):
+    import time
+
+    try:
+        with path.open("r+b") as fd:
+            with lock(fd, lock_attempts=1):
```
kenodegard:
```suggestion
            with _lock_impl(fd, lock_attempts=1):
```

kenodegard (dismissed):

kenodegard:
pre-commit.ci autofix

kenodegard (approved):

PR #14985, opened 2025-07-03
When deprecating a class wrap `__init__`

kenodegard:
### Description

We were previously treating all class deprecations the same as a function/method. This worked great until we tried to access class attributes which wasn't possible when the deprecation wrapper is a function.

Instead we now detect if a deprecation is for a class in which case we attempt to wrap the `__init__` constructor instead.

### Checklist - did you ...

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -96,12 +96,12 @@ class DeprecationHandler:
         :param stack: Optional stacklevel increment.
         """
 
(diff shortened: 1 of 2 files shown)


jezdez (commented):
Ah, that's good, I think I stumbled over this in #14895

jezdez (approved):

PR #14986, opened 2025-07-03
Create `EnvironmentConfig` from context

soapy1:
### Description

This PR adds the ability to create an `EnvironmentConfig` object from the current running conda context. For example developers requiring the `EnvironmnentConfig` for the current context can use:

```
config = EnvironmentConfig.from_context()
```

This can be used for collecting information about the current command running environemtn, exporting the current environment, etc.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1277,10 +1277,7 @@ class Context(Configuration):
     @property
     def environment_settings(self) -> dict[str, Any]:
         """Returns a dict of environment related settings"""
-        return {
-            key: getattr(self, key)
-            for key in self.environment_context_keys
-        }
+        return {key: getattr(self, key) for key in self.environment_context_keys}
 
     @property
     def category_map(self) -> dict[str, tuple[str, ...]]:
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1255,6 +1255,34 @@ class Context(Configuration):
         return self._default_activation_env or ROOT_ENV_NAME
 
     @property
+    def environment_context_keys(self) -> list[str]:
+        return [
+            "aggressive_update_packages",
+            "channel_priority",
+            "channels",
+            "channel_settings",
+            "custom_channels",
+            "custom_multichannels",
+            "deps_modifier",
+            "disallowed_packages",
+            "pinned_packages",
+            "repodata_fns",
+            "sat_solver",
+            "solver",
+            "track_features",
+            "update_modifier",
+            "use_only_tar_bz2",
+        ]
+
+    @property
+    def environment_settings(self) -> dict[str, Any]:
+        """Returns a dict of environment related settings"""
+        return {
+            key: getattr(self, key)
+            for key in self.environment_context_keys
+        }
+
+    @property
     def category_map(self) -> dict[str, tuple[str, ...]]:
         return {
             "Channel Configuration": (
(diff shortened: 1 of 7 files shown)


conda/base/context.py
```diff
@@ -1254,6 +1254,34 @@ def auto_activate_base(self) -> bool:
     def default_activation_env(self) -> str:
         return self._default_activation_env or ROOT_ENV_NAME
 
+    @property
+    def environment_context_keys(self) -> list[str]:
```
soapy1, line 1258: I don't think this needs to be it's own function here per se. But, trying to follow the pattern used other places in this module, for example `category_map`.

conda/models/environment.py
```diff
@@ -126,6 +127,17 @@ def _merge(self, other: EnvironmentConfig) -> EnvironmentConfig:
...
 
+    @classmethod
+    def from_context(cls) -> EnvironmentConfig:
+        """
+        **Experimental** While experimental, expect both major and minor changes across minor releases.
+
+        Create an EnvironmentConfig from the current context
+        """
+        field_names = [f.name for f in fields(cls)]
+        environment_settings = {key: value for key, value in context.environment_settings.items() if key in field_names}
```
soapy1: This ensures that we are only passing in the data that the `EnvironmentConfig` class needs. We might want to do some additional processing here in the future?

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

conda/base/context.py
jezdez: I like this

tests/conftest.py
```diff
@@ -39,6 +42,56 @@
...
+@pytest.fixture
+def context_testdata() -> None:
+    reset_context()
+    context._set_raw_data(
+        {
+            "testdata": YamlRawParameter.make_raw_parameters(
+                "testdata", yaml_round_trip_load(TEST_CONDARC)
+            )
+        }
+    )
```
jaimergp: This is better defined in one of the modules under `conda.testing`.

jaimergp (dismissed):

conda/base/context.py
jaimergp: Hm, I kind of agree with you, but the alternative (place it in `Environment`) is not better either. I'll defer to others for a final say, but this is not a blocker imo.

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

jaimergp:
Thanks! Also needs a `news/` item.

soapy1:
pre-commit.ci autofix

jaimergp (dismissed):

conda/models/environment.py
```diff
@@ -126,6 +127,21 @@ def _merge(self, other: EnvironmentConfig) -> EnvironmentConfig:
...
         return self
 
+    @classmethod
+    def from_context(cls) -> EnvironmentConfig:
+        """
+        **Experimental** While experimental, expect both major and minor changes across minor releases.
+
+        Create an EnvironmentConfig from the current context
+        """
+        field_names = [f.name for f in fields(cls)]
```
jezdez:
```suggestion
        field_names = {field.name for field in fields(cls)}

```
Minor optimization

jezdez (approved):

PR #14999, opened 2025-07-09
Ensure file types can not be mixed in the conda install command

soapy1:
### Description

https://github.com/conda/conda/issues/14684 added a check to ensure that `conda create/update` commands do not allow users to mix file types provided with the `--file` command.
 
This makes `conda install` consistent with the create and update commands.

For example:
```
$ conda install --file tests/env/support/explicit.txt --file tests/env/support/add-pip.yml                      

EnvironmentFileTypeMismatchError: Cannot mix environment file formats.

'tests/env/support/explicit.txt' is a explicit format file
'tests/env/support/add-pip.yml' is a environment.yml format file
```

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_install.py
+++ b/conda/cli/main_install.py
@@ -128,6 +128,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
 def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import context
     from ..exceptions import CondaValueError
+    from .common import validate_environment_files_consistency
     from .install import get_revision, install, install_revision
 
     if context.force:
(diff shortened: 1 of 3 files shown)


conda/cli/main_install.py
```diff
@@ -140,6 +141,9 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             file=sys.stderr,
         )
 
+    # Validate that input files are of the same format type
+    validate_environment_files_consistency(args.file)
```
soapy1, line 145: could also do a small refactor to move this common validation into `conda.cli.install.install`.

tests/test_create.py
```diff
@@ -2764,7 +2764,7 @@ def test_python_site_packages_path(
         assert (prefix / sp_dir / "sample.py").is_file()
 
 
-def test_dont_allow_mixed_file_arguments(
+def test_dont_allow_mixed_file_arguments_create(
```
ForgottenProgramme: Can we parameterize this test for `create` and `install` ?
Then we won't need two tests that have mostly the same code. That would also make it simple to extend this test for other commands in the future.

soapy1: good idea! pushed up this change.

soapy1:
pre-commit.ci autofix

kenodegard (approved):

PR #15000, opened 2025-07-09
EnvironmentConfig lists of values should be `tuples`

soapy1:
### Description

Iterable values coming from the context objects are usually returned as `tuples`. This PR changes the `EnvironmentConfig` types to more closely match the context.

### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -64,7 +64,7 @@ class EnvironmentConfig:
     use_only_tar_bz2: bool | None = None
 
     def _append_without_duplicates(self, first: tuple, second: tuple) -> tuple:
-        return tuple(dict.fromkeys(item for item in first+second))
+        return tuple(dict.fromkeys(item for item in first + second))
 
     def _merge_channel_settings(
         self, first: tuple[dict[str, str]], second: tuple[dict[str, str]]
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -37,39 +37,38 @@ class EnvironmentConfig:
     Data model for a conda environment config.
     """
 
-    aggressive_update_packages: list[str] = field(default_factory=list)
+    aggressive_update_packages: tuple[str] = field(default_factory=tuple)
 
     channel_priority: ChannelPriority | None = None
 
-    channels: list[str] = field(default_factory=list)
+    channels: tuple[str] = field(default_factory=tuple)
 
-    channel_settings: list[dict[str, str]] = field(default_factory=list)
+    channel_settings: tuple[dict[str, str]] = field(default_factory=tuple)
 
     deps_modifier: DepsModifier | None = None
 
-    disallowed_packages: list[str] = field(default_factory=list)
+    disallowed_packages: tuple[str] = field(default_factory=tuple)
 
-    pinned_packages: list[str] = field(default_factory=list)
+    pinned_packages: tuple[str] = field(default_factory=tuple)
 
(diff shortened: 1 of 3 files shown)


conda/models/environment.py
```diff
@@ -162,7 +162,7 @@ def from_context(cls) -> EnvironmentConfig:
         field_names = {field.name for field in fields(cls)}
 
         environment_settings = {
-            key: value
+            key: list(value) if isinstance(value, tuple) else value
```
soapy1: hmmm, I'm not completely sure that this is a good idea. Is there a case when we want an `EnvironmentConfig` field to be a `tuple`? Probably?

jaimergp: I'd expect `EnvironmentConfig` items to be tuples rather than lists because I consider them "static" rather than something I'll modify. Merging envs is a side detail that shouldn't be more important than the "data vessel" role of the dataclass, so if anything the complexity of merging tuples should be done in the merge methods.

IOW, I don't think we need to force a `list` here, but manage the tuple merge correctly.

soapy1: that approach sounds good to me!

soapy1:
pre-commit.ci autofix

soapy1:
pre-commit.ci autofix

peytondmurray:
This is a nice improvement, thanks! I double checked the field types of the `EnvironmentConfig` and compared them to corresponding attributes of the `context` object in my local dev environment, and they (nearly) line up at this point:

```
                    field_name              field_type        context_field_type
0   aggressive_update_packages              tuple[str]           <class 'tuple'>
1             channel_priority  ChannelPriority | None  <enum 'ChannelPriority'>
2                     channels              tuple[str]           <class 'tuple'>
3             channel_settings   tuple[dict[str, str]]           <class 'tuple'>
4                deps_modifier     DepsModifier | None     <enum 'DepsModifier'>
5          disallowed_packages              tuple[str]           <class 'tuple'>
6              pinned_packages              tuple[str]           <class 'tuple'>
7                 repodata_fns              tuple[str]           <class 'tuple'>
8                   sat_solver  SatSolverChoice | None  <enum 'SatSolverChoice'>
9                       solver              str | None             <class 'str'>
10              track_features              tuple[str]           <class 'tuple'>
11             update_modifier   UpdateModifier | None   <enum 'UpdateModifier'>
12            use_only_tar_bz2             bool | None        <class 'NoneType'>
```

There are a few remaining incompatibilities, for example the enum types in the context do not support `None` as a valid setting, but maybe that's best handled in a separate issue, or maybe not at all? I guess it depends on what purpose the `EnvironmentConfig` is really supposed to serve, and what role if any the `context` has in relation to it.

soapy1:
Nice @peytondmurray, thanks this is really helpful! 

> for example the enum types in the context do not support None as a valid setting, 

I think in this case it's ok for `EnvironmentConfig` to be different from the context. For example, if these enum values can not be `None` in the `EnvironmentConfig` then they need to always be populated with something valid. This can either be (1) the default value for the config, or (2) a null value. 

The `EnvironmentConfig` object is meant to be created by environment spec plugins. It's valid for an environment spec to not support all the possible environment config fields. In this case, I think these plugin authors shouldn't have the responsibility of choosing good defaults for these pieces of config that they don't want to control. It's maybe best for default values to always be coming from the same source. In this case, the context object is a good source of truth for good default values. 

Definitely open to looking at this another way!

kenodegard (approved):

jezdez (approved):

PR #15002, opened 2025-07-09
Add plugin type base class for name normalization

kenodegard:
### Description

While we initially didn't expect every conda plugin to have a name we have clearly fallen into this pattern so we might as well standardize it.

And while at it might as well include post-init cleanup to lowercase and strip the name.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -42,7 +42,7 @@ from .subcommands.doctor import health_checks
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
-    from typing import Callable, Literal
+    from typing import Literal
 
     from requests.auth import AuthBase
 
@@ -80,17 +80,9 @@ class CondaPluginManager(pluggy.PluginManager):
 
     #: Cached version of the :meth:`~conda.plugins.manager.CondaPluginManager.get_solver_backend`
     #: method.
-    get_cached_solver_backend: Callable[[str | None], type[Solver]]
+    get_cached_solver_backend = None
 
-    #: Cached version of the :meth:`~conda.plugins.manager.CondaPluginManager.get_session_headers`
-    #: method.
-    get_cached_session_headers: Callable[[str], dict[str, str]]
-
-    #: Cached version of the :meth:`~conda.plugins.manager.CondaPluginManager.get_request_headers`
-    #: method.
-    get_cached_request_headers: Callable[[str, str], dict[str, str]]
-
-    def __init__(self, project_name: str | None = None, *args, **kwargs):
+    def __init__(self, project_name: str | None = None, *args, **kwargs) -> None:
         # Setting the default project name to the spec name for ease of use
         if project_name is None:
             project_name = spec_name
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -284,7 +284,7 @@ class CondaPluginManager(pluggy.PluginManager):
         if invalid:
             raise PluginError(
                 f"Invalid plugin names found for `{name}`:\n"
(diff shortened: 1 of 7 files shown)


conda/plugins/manager.py
```diff
@@ -80,13 +80,24 @@ class CondaPluginManager(pluggy.PluginManager):
...
+    get_cached_request_headers: Callable[[str, str], dict[str, str]]
+
+    def __init__(self, project_name: str | None = None, *args, **kwargs):
         # Setting the default project name to the spec name for ease of use
         if project_name is None:
             project_name = spec_name
         super().__init__(project_name, *args, **kwargs)
+        self.cache_init()
+
+    def cache_init(self) -> None:
```
kenodegard: `cache_init` or `init_cache`?

soapy1 (dismissed):

conda/plugins/manager.py
kenodegard: This change is unrelated and not necessary for this PR, removing:

```suggestion
```

conda/plugins/types.py
```diff
@@ -38,7 +39,25 @@
 
 
 @dataclass
-class CondaSubcommand:
+class CondaBasePlugin:
```
jezdez:
```suggestion
class CondaPlugin:
```

I'm not sure if "base" adds anything basically; if we're introducing a `dataclass` based base class, then for all intents and purposes that class becomes the representation of a conda plugin. We didn't do that in the past since we weren't sure yet how this would evolve, but I think we can just call it `CondaPlugin` bacause of that.

conda/plugins/types.py
```diff
@@ -38,7 +39,25 @@
...
+    """
+    Base class for all conda plugins.
+    """
+
+    name: str
```
jezdez, line 48: Should we add a code comment here that explains that this is a user-facing name? We'll print it in error messages, etc.

conda/plugins/manager.py
```diff
@@ -268,23 +268,28 @@ def get_hook_results(self, name, **kwargs):
...
+            or plugin.name != plugin.name.lower().strip()
+        ]
         if invalid:
             raise PluginError(
                 dals(
                     f"""
                     Invalid plugin names found:
 
-                    {", ".join([str(plugin) for plugin in invalid])}
+                    {", ".join([f"{plugin}:{plugin.name}" for plugin in invalid])}
```
jezdez: Huh, what will `plugin` be printed as here?

jezdez (commented):
Some clarifying questions

conda/plugins/manager.py
kenodegard: I thought we could provide more info like which module the plugin came from but this didn't do that so I've reverted those changes

conda/plugins/manager.py
```diff
@@ -266,48 +267,48 @@ def get_hook_results(self, name, **kwargs):
...
-
-        # Check for invalid names
-        invalid = [plugin for plugin in plugins if not isinstance(plugin.name, str)]
+        # hook() returns a generator of all plugins for a given specname,
+        # unfortunately this generator does not offer any information about which
+        # package/module the plugin is defined in, this makes reporting errors in
+        # a meaningful way to users difficult
+        plugins = [plugin for plugins in hook(**kwargs) for plugin in plugins]
+
+        # Validate plugin names since plugins may not properly inherit from CondaPluginBase
```
kenodegard:
```suggestion
        # Validate plugin names since plugins may not properly inherit from CondaPlugin
```

jezdez (approved):

soapy1 (approved):

PR #15013, opened 2025-07-15
Avoid modifying tests/env/support/ during test run

### LINKED ISSUE #15011: Test suite modifies tests/env/support
### Checklist

### What happened?

`test_create_advanced_pip` and possibly others modify the source checkout, confusing pytest discovery and the debugger. At least one of these is because pip-called-by-conda uses paths relative to a .yml input.

Instead, copy the .yml to a temporary folder before running conda.

### Conda Info

```shell
HEAD as of 2025-07-15
```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

dholth:
### Description

A different strategy might be to copy this entire folder to a temporary path as a session scoped fixture.

Fix #15011 

### Checklist - did you ...


kenodegard:
What if we refactor `support_file` into fixtures instead that handles the copying for us?

```python
SUPPORT_DIR = Path(__file__).parent / "env" / "support"


@pytest.fixture
def support_file_url(support_file_server_port: int) -> Callable[[str], str]:
    def support_file_url(path: str) -> str:
        return f"http://127.0.0.1:{support_file_server_port}/{filename}"

    return support_file_url


@pytest.fixture
def support_file(path_factory: PathFactoryFixture) -> Callable[[str], Path]:
    def support_file(path: str) -> Path:
        src = SUPPORT_DIR / path
        dst = path_factory()
        shutil.copytree(src, dst)
        return dst

    return support_file
```

dholth:
> What if we refactor `support_file` into fixtures instead that handles the copying for us?
> 
> ```python
> SUPPORT_DIR = Path(__file__).parent / "env" / "support"
> 
> 
> @pytest.fixture
> def support_file_url(support_file_server_port: int) -> Callable[[str], str]:
>     def support_file_url(path: str) -> str:
>         return f"http://127.0.0.1:{support_file_server_port}/{filename}"
> 
>     return support_file_url
> 
> 
> @pytest.fixture
> def support_file(path_factory: PathFactoryFixture) -> Callable[[str], Path]:
>     def support_file(path: str) -> Path:
>         src = SUPPORT_DIR / path
>         dst = path_factory()
>         shutil.copytree(src, dst)
>         return dst
> 
>     return support_file
> ```

Added

tests/conftest.py
```diff
@@ -86,6 +87,22 @@ def support_file_server_port(
...
 
 
+@pytest.fixture
+def support_file_isolated(tmp_path):
+    """
+    Copy support files to temporary path, avoid polluting source checkout.
+    """
+    source = Path(__file__).parents[0] / "env" / "support"
+    base = tmp_path / "support"
+    if not base.exists():  # tmp_path is session scoped
```
kenodegard, line 97: `tmp_path` is function scoped, so this will be copied every time this fixture is used

dholth: It crashes without the "if not base.exists()" check because the destination file is already there.

kenodegard: Since we don't specify the scope it defaults to the function scope. Also, if we set the scope to session:

```python
@pytest.fixture(scope="session")
def support_file_isolated(tmp_path):
    ...
```

`tmp_path` will error:

Screenshot 2025-07-16 at 14 42 04

dholth: Works fine for me, with no explicit scope. Fixture is used only twice in methods that were both named "test_create_advanced_pip" before this PR. It does seem like the same fixture receives the same tmp_path both times it's instantiated for the tests, which is acceptable here.

```
% pytest -k advanced_pip
====================================================================== test session starts ======================================================================
platform darwin -- Python 3.12.2, pytest-8.3.4, pluggy-1.5.0 -- /Users/dholth/prog/conda/devenv/Darwin/arm64/envs/devenv-3.12-c/bin/python
cachedir: .pytest_cache
conda.__file__: /Users/dholth/prog/conda/conda/__init__.py
rootdir: /Users/dholth/prog/conda
configfile: pyproject.toml
testpaths: tests
plugins: cov-6.0.0, timeout-2.2.0, rerunfailures-12.0, xdoctest-1.2.0, xprocess-1.0.2, mock-3.14.0, split-0.10.0
collected 2395 items / 2393 deselected / 2 selected                                                                                                             

tests/env/test_create.py::test_create_advanced_pip PASSED                                                                                                 [ 50%]
tests/env/test_env.py::test_env_advanced_pip PASSED          
```

dholth: @kenodegard if you need a specific change to merge this PR, please make it directly in the branch.

jezdez (approved):

PR #15022, opened 2025-07-18
Add `conda doctor` check for if the pinned file is ill formatted

### LINKED ISSUE #7978: Raise error if `conda-meta`'s `pinned` file is ill-formed
Currently if someone adds content to `conda-meta`'s `pinned` file that is ill-formed, it proceeds without raising or in the best case scenario may raise a tangential error message downstream. However this can really start to cause problems. For example, if one accidentally adds quotes around content in `pinned` that results in this content being blissfully ignored by Conda, causing surprising downstream consequences. Would be nice in these and other cases to validate `pinned` parses correctly and raise with an informative error if it doesn't. Thus blocking unintentional environment changes (which `pinned` is intended to avoid) and helping the user to diagnose what went wrong in `pinned`.

soapy1:
### Description

This PR proposes adding a check to `conda doctor` to inspect for malformed lines in the `conda-meta/pinned` file in an env. This solution does not check for malformed pinned file on every run since it's hard to determine if a file is actually malformed or not.

In this check, we'll inspect each pinned package. If the package does not exist in the prefix, that might indicate that the package is not formed correctly. Note, that a package in a pinned file but not in the prefix __might not__ be malformed, even though it might be reported as such.

For example, this may catch typos or extra punctuation added to package names. 

For example, 

for no pinned file, or an empty file
```
$ conda doctor
. . .
✅ No pinned specs found in /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned.
```

for a well formatted file
```
$ cat /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned
numpy == 1.11

$ conda doctor
. . .
✅ The pinned file in /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned is seems well formatted.
```

for a badly formatted file, eg extra quotes
```
$ cat /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned
"numpy" 1.11

$ conda doctor
. . . 
❌ The following specs in /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned are maybe malformed:

    - "numpy"
 ```

for a bady formatted file, eg. multiple packages specified on a line
```
$ cat /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned
numpy == 1.11, scipy == 0.14*

$ conda doctor
❌ Unable to read pinned file at /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned:
        Invalid spec 'numpy == 1.11, scipy == 0.14* optional=True': Invalid version 'scipy==0.14': invalid character(s)
```

for a well formatted file, but numpy is not installed in the environment (false positive) 
```
$ cat /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned
numpy == 1.11

$ conda doctor
. . .
❌ The following specs in /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/pinned-ex/conda-meta/pinned are maybe malformed:

    - numpy
```

fixes: https://github.com/conda/conda/issues/7978

Other changes in this PR:
* introduces a new constant `PREFIX_PINNED_FILE` to describe the path to the pinned file

### Checklist - did you ...

### CODE REVIEW

**Comment by soapy1** [comment]
pre-commit.ci autofix

File: tests/plugins/subcommands/doctor/test_health_checks.py [hunk 1]
```diff
@@ -400,3 +400,26 @@ def test_env_consistency_constrains_not_met(
...
+    ],
+)
+def test_pinned_will_formatted_check(
+    tmp_env: TmpEnvFixture,
+    conda_cli: CondaCLIFixture,
+    pinned_file: str,
+    expected_output: str,
+):
+    with tmp_env() as prefix:
+        (prefix / "conda-meta" / "pinned").write_text(pinned_file)
```
soapy1: It's maybe better to not have these tests reading/writing to the filesystem so much. Can refactor this some to not depend on the filesystem if it seems like a good idea.

conda/core/prefix_data.py
```diff
@@ -594,6 +594,21 @@ def set_nonadmin(self) -> None:
             self.prefix_path.mkdir(parents=True, exist_ok=True)
             (self.prefix_path / ".nonadmin").touch()
 
+    def get_pinned_specs(self) -> tuple[MatchSpec]:
```
kenodegard, line 629:
```suggestion
    def get_pinned_specs(self) -> tuple[MatchSpec, ...]:
```

conda/core/prefix_data.py
```diff
@@ -594,6 +594,21 @@ def set_nonadmin(self) -> None:
...
+        """Find pinned specs from file and return a tuple of MatchSpec."""
+        pin_file = self.prefix_path / "conda-meta" / "pinned"
+        if pin_file.exists():
+            with pin_file.open() as f:
+                from_file = (
+                    i
+                    for i in f.read().strip().splitlines()
+                    if i and not i.strip().startswith("#")
+                )
```
kenodegard, line 638:
```suggestion
            from_file = (
                spec
                for spec in pin_file.read_text().strip().splitlines()
                if spec and not spec.strip().startswith("#")
            )
```

conda/core/solve.py
```diff
@@ -1387,23 +1386,13 @@ def working_state_reset(self):
         self.final_environment_specs = None
 
 
-def get_pinned_specs(prefix):
+def get_pinned_specs(prefix: str) -> tuple[MatchSpec]:
```
kenodegard, line 1389:
```suggestion
def get_pinned_specs(prefix: str) -> tuple[MatchSpec, ...]:
```

conda/core/solve.py
```diff
@@ -1387,23 +1386,13 @@ def working_state_reset(self):
...
-        from_file = ()
-
-    return tuple(
-        MatchSpec(spec, optional=True)
-        for spec in (*context.pinned_packages, *from_file)
+    context_pinned_packages = tuple(
+        MatchSpec(spec, optional=True) for spec in context.pinned_packages
     )
+    prefix_data = PrefixData(prefix_path=prefix)
+    return context_pinned_packages + prefix_data.get_pinned_specs()
```
kenodegard, line 1395:
```suggestion
    return (
        *(MatchSpec(spec, optional=True) for spec in context.pinned_packages),
        *PrefixData(prefix_path=prefix).get_pinned_specs(),
    )
```

tests/plugins/subcommands/doctor/test_health_checks.py
kenodegard: I'm not too worried about this

kenodegard (dismissed):
Nice!

tests/plugins/subcommands/doctor/test_health_checks.py
jezdez: Same

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -238,6 +239,42 @@ def consistent_env_check(prefix: str, verbose: bool) -> None:
...
 
+def pinned_well_formatted_check(prefix: str, verbose: bool) -> None:
+    prefix_data = PrefixData(prefix_path=prefix)
+
+    try:
+        pinned_specs = prefix_data.get_pinned_specs()
+    except Exception as err:
+        print(
+            f"{X_MARK} Unable to read pinned file at {prefix}/conda-meta/pinned:\n\t{err}"
+        )
```
jezdez: Wondering if we should catch `OSError` here specifically and then have another case that deals with unknown errors, e.g.:
```suggestion
    except OSError as err:
    	# Create path here to account for cross-platform paths
        pinned_path = prefix / "conda-meta" / "pinned"
        print(
            f"{X_MARK} Unable to open pinned file at {pinned_path}:\n\t{err}"
        )
    except Exception as err:
        print(
            f"{X_MARK} An error occured trying to read pinned file at {prefix}/conda-meta/pinned:\n\t{err}"
        )
```

That said, maybe it would be wise to create this in `conda/base/constants.py`:

```
PREFIX_PINNED_FILE: Final[PathType] = join("conda-meta", "pinned")
```

There are quite a few instances of manual path creation through out the code base.

```suggestion
    except OSError as err:
        # Create path here to account for cross-platform paths
        print(
            f"{X_MARK} Unable to open pinned file at {prefix / PREFIX_PINNED_FILE}:\n\t{err}"
        )
    except Exception as err:
        print(
            f"{X_MARK} An error occurred trying to read pinned file at {prefix / PREFIX_PINNED_FILE}:\n\t{err}"
        )

File: conda/plugins/subcommands/doctor/health_checks.py [hunk 7]
```diff
@@ -238,6 +239,42 @@ def consistent_env_check(prefix: str, verbose: bool) -> None:
...
+        print(
+            f"{X_MARK} The following specs in {prefix}/conda-meta/pinned are maybe malformed:"
+        )
+        print(dashlist((spec.name for spec in maybe_malformed), indent=4))
+        return
+
+    # If there are no malformed packages, the pinned file is well formatted
+    print(
+        f"{OK_MARK} The pinned file in {prefix}/conda-meta/pinned is seems well formatted."
+    )
```
jezdez: Same `{prefix / PREFIX_PINNED_FILE}` for those other cases, using forward slashes will not be nice to Windows users

conda/plugins/subcommands/doctor/health_checks.py
```diff
@@ -249,3 +286,7 @@ def conda_health_checks():
     yield CondaHealthCheck(
         name="Consistent Environment Check", action=consistent_env_check
     )
+    yield CondaHealthCheck(
+        name="conda-meta/pinned Well Formatted Check",
```
jezdez:
```suggestion
        name=f"{PREFIX_PINNED_FILE} Well Formatted Check",
```

jezdez (changes requested):
This looks good in general, I think we need to use the opportunity to fix the path creation to `conda-metadata/pinned` in the code base though, like we do with other prefix files. Also, let's take into account that on Windows paths are constructed differently, we can make use of pathlib for those cases.

conda/plugins/subcommands/doctor/health_checks.py
soapy1: Good idea!

soapy1:
pre-commit.ci autofix

jezdez (approved):

PR #15026, opened 2025-07-21
fix `TypeError: can only concatenate tuple (not "list") to tuple`

jaimergp:
### Description

Small fix for some environment files in https://github.com/conda/conda/pull/14911. Reproduces with:

```
conda install --file tests/env/support/empty_env.yml
```

### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -64,7 +64,7 @@ class EnvironmentConfig:
     use_only_tar_bz2: bool | None = None
 
     def _append_without_duplicates(self, first: tuple, second: tuple) -> tuple:
-        return tuple(dict.fromkeys(item for item in first + second))
+        return tuple(dict.fromkeys(item for item in chain(first, second)))
 
     def _merge_channel_settings(
         self, first: tuple[dict[str, str]], second: tuple[dict[str, str]]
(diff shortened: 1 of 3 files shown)


kenodegard (dismissed):

conda/models/environment.py
```diff
@@ -64,7 +64,7 @@ class EnvironmentConfig:
     use_only_tar_bz2: bool | None = None
 
     def _append_without_duplicates(self, first: tuple, second: tuple) -> tuple:
```
kenodegard: Should we update the types?


```suggestion
    def _append_without_duplicates(self, first: Iterable[T], second: Iterable[T]) -> tuple[T, ...]:
```

soapy1 (dismissed):
It looks like the root of this issue is the `EnvironmentYaml` env spec plugin is not respecting the type annotations of the `EnvironmentConfig` class. I think this change is ok to make the model more resilient, but I think it would also be helpful to update the `EnvironmentYaml` plugin to return the correct type.

Can we do that in this PR as well? https://github.com/conda/conda/blob/main/conda/env/env.py#L271 needs to be updated to be  `config = EnvironmentConfig(channels=tuple(self.channels))`.

jaimergp:
Thanks @soapy1 @kenodegard, I applied your suggestions!

soapy1 (approved):

kenodegard (approved):

PR #15030, opened 2025-07-21
Simplify `get_environment_specifier_by_name` since names are unique

kenodegard:
### Description

Since `get_environment_specifiers` already returns a dictionary mapping plugin names to environment specifier plugins and the `get_hook_results` method already checks for duplicate plugin names we can simplify how we do name matching in `get_environment_specifiers_by_name`.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -267,48 +267,48 @@ class CondaPluginManager(pluggy.PluginManager):
         if hook is None:
             raise PluginError(f"Could not find requested `{name}` plugins")
 
-        plugins = [item for items in hook(**kwargs) for item in items]
+        # hook() returns a generator of all plugins for a given specname,
+        # unfortunately this generator does not offer any information about which
+        # package/module the plugin is defined in, this makes reporting errors in
+        # a meaningful way to users difficult
+        plugins = [plugin for plugins in hook(**kwargs) for plugin in plugins]
 
-        # Check for invalid names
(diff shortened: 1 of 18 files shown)


conda/plugins/manager.py
```diff
@@ -524,28 +525,27 @@ def get_environment_specifier_by_name(
...
         :param source: full path to the environment spec file/source
         :param name: name of the environment plugin to load
         :returns: an environment specifier plugin that matches the provided plugin name, or can handle the provided file
         """
         name = name.lower()
-        hooks = self.get_environment_specifiers()
-        found = [hook for hook_name, hook in hooks.items() if hook_name == name]
-
-        if not found:
-            raise CondaValueError(
```
kenodegard, line 536: Why was this previously a `CondaValueError`?

soapy1: The idea was that if no plugin is found then there is an error related to the invalid value that was passed into the function. Could also throw a `PluginError` here to be more consistent with the rest of this module.

jezdez: PluginError should be raised if there is an issue with the plugin, it's not a catchall for all plugin related errors

jezdez (changes requested):
I don't think we should accidentally change the exception for this one without careful review given how important the plugin system has become

soapy1 (approved):

jezdez (approved):

PR #15036, opened 2025-07-22
`CondaPluginManager` should raise `PluginError`s

kenodegard:
### Description

Shouldn't `CondaPluginManager` be raising a `PluginError` instead of a `CondaValueError`?

### Checklist - did you ...


soapy1 (dismissed):

soapy1 (commented):
Looks like the tests are just failing on updating the tests for this change.

jezdez:
Just restarted the failing tests, to see if they were transient

jezdez:
As background on why we used a `ValueError` exception for some of the errors so far:

We raise a `ValueError` when the user-provided value isn't correct and `PluginError` when something went wrong in the interaction with the plugin. I don't think those are the same error states, and we're probably a little too fuzzy so far. We could review the plugin manager code again, after it has grown so much over the years.

jezdez:
pre-commit.ci autofix

ForgottenProgramme (approved):

jezdez (changes requested):
Since this has gotten an approval and I didn't spell out my concerns clearly enough last time, I don't agree with this change since the specific error is a user-created error, and so a "value error," and not something that the plugin itself has caused, which is what `PluginError` is about.

PR #15039, opened 2025-07-24
Refactor cli layer: step 0 - Use new `Environment` dataclass to refactor install sources

soapy1:
### Description
This PR updates the `conda.cli.install` module to use the `Environment` dataclass to assemble an environment that can be solved and then installed. This refactor will support the work for reconciling the `conda install/create/update` and `conda env` backends described in https://github.com/conda/conda-environments-project-mgmt/issues/15.

There are a few other small changes that this PR suggests:
* making the `@EXPLICIT` marker a constant
* deprecating the `json` argument in the `conda.cli.common.specs_from_url`  function
* simplifies the should retry unfrozen logic

fixes:
* https://github.com/conda/conda-environments-project-mgmt/issues/63
* https://github.com/conda/conda-environments-project-mgmt/issues/69

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -313,8 +313,8 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
         spec = MatchSpec(spec)
         if not spec.is_name_only_spec:
             raise CondaError(
-                f"'conda update' only supports name-only spec, but received: {spec}\n"
-                f"Use 'conda install' to install a specific version of a package."
+                f"`conda update` only supports name-only spec, but received: {spec}\n"
+                f"Use `conda install` to install a specific version of a package."
             )
         if not prefix_data.get(spec.name, None):
             raise PackageNotInstalledError(prefix, spec.name)
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -491,13 +491,12 @@ class Environment:
         :param args: argparse Namespace containing command-line arguments
         :return: An Environment object representing the cli
         """
-        specs = [s.strip("\"'") for s in args.packages]
+        specs = [package.strip("\"'") for package in args.packages]
         if add_default_packages:
-            names = {MatchSpec(pkg).name for pkg in specs}
-            for pkg in context.create_default_packages:
-                pkg_name = MatchSpec(pkg).name
-                if pkg_name not in names:
-                    specs.append(pkg)
+            names = {MatchSpec(spec).name for spec in specs}
+            for default_package in context.create_default_packages:
+                if MatchSpec(default_package).name not in names:
+                    specs.append(default_package)
 
         requested_packages = []
         fetch_explicit_packages = []
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -22,7 +22,7 @@ from ..base.constants import (
     CMD_LINE_SOURCE,
     CONFIGURATION_SOURCES,
     ENV_VARS_SOURCE,
-    EXPLICIT_MARKER
+    EXPLICIT_MARKER,
     PREFIX_MAGIC_FILE,
 )
 from ..base.context import context, env_name
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -530,7 +530,7 @@ def test_conda_update_package_is_not_name_only_spec(
     """
     with tmp_env() as prefix:
         with pytest.raises(
-            CondaError, match="'conda update' only supports name-only spec"
+            CondaError, match="`conda update` only supports name-only spec"
         ):
             conda_cli("update", f"--prefix={prefix}", "conda-forge::*")
 


soapy1:
pre-commit.ci autofix

conda/base/constants.py
```diff
@@ -382,3 +382,7 @@ class NoticeLevel(ValueEnum):
 # Indicates whether or not external plugins (i.e., plugins that aren't shipped
 # with conda) are enabled
 NO_PLUGINS: Final = False
+
+# When this sting is present in an environment file, it indicates that the file
```
kenodegard:
```suggestion
# When this string is present in an environment file, it indicates that the file
```

tests/cli/test_cli_install.py
kenodegard, line 1: revert?

soapy1: yes!

conda/base/constants.py
soapy1: nice catch! just a small change so I'm going to shove this into the commit that originally added it

conda/cli/install.py
```diff
@@ -309,12 +316,75 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+                    if spec != EXPLICIT_MARKER
+                ]
+            )
+        except UnicodeError:
+            raise CondaError(
+                "Error reading file, file should be a text file containing"
+                " packages \nconda create --help for details"
+            )
+
+    for spec in specs:
```
kenodegard: If we filter out default packages based on `specs` shouldn't we do the same for specs defined in `files`?

soapy1: Good catch! This is consistent with the behaviour on the latest release of conda  (and older versions, I tested as old as 24.7.0). tbh I'm not sure if this is a bug or not, but I would certainly call it unexpected? I opened up a bug report in https://github.com/conda/conda-environments-project-mgmt/issues/86. Would you call this a bug?

soapy1:
pre-commit.ci autofix

conda/cli/install.py
kenodegard: as spelled out in the issue you opened I agree it is a bug, will continue conversation there

soapy1: Cool! Since this is a bug that already exists in the currently released conda, I would suggest fixing it in a follow up PR. That'll keep this PR lean and focused and (hopefully) easier to review.

soapy1:
pre-commit.ci autofix

conda/cli/install.py
soapy1: Follow up PR is queued up https://github.com/conda/conda/pull/15041

conda/base/constants.py
```diff
@@ -382,3 +382,7 @@ class NoticeLevel(ValueEnum):
 # Indicates whether or not external plugins (i.e., plugins that aren't shipped
 # with conda) are enabled
 NO_PLUGINS: Final = False
+
+# When this string is present in an environment file, it indicates that the file
+# describes an explicit environment spec.
+EXPLICIT_MARKER: Final = "@EXPLICIT"
```
jezdez, line 389: Finally!

conda/cli/install.py
```diff
@@ -309,88 +316,81 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    specs: list[str] = [],
+    files: list[str] = [],
+    inject_default_packages: bool = True,
+) -> Environment:
+    if inject_default_packages:
+        names = {MatchSpec(pkg).name for pkg in specs}
+        for pkg in context.create_default_packages:
+            pkg_name = MatchSpec(pkg).name
+            if pkg_name not in names:
+                specs.append(pkg)
```
jezdez: To clarify is appending the right move here, or do we need to insert it in front?

conda/cli/install.py
```diff
@@ -309,88 +316,81 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
-    """Executes an install of a new conda environment by cloning."""
-    prefix = context.target_prefix
-    index_args = get_index_args(args)
+def assemble_environment(
+    name: str | None = None,
+    prefix: str | None = None,
+    specs: list[str] = [],
+    files: list[str] = [],
+    inject_default_packages: bool = True,
+) -> Environment:
```
jezdez: Let's add a docstring

conda/cli/install.py
```diff
@@ -309,88 +316,81 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    if inject_default_packages:
```
jezdez: The `create_default_packages` parameter has this TODO:
```
# TODO: This is a bad parameter name. Consider an alternate.
```

But I'm not sure if `injecting` is better, since dependency injection has a particular meaning. Could we use `preinstall` or `auto-install`? Any other ideas?

conda/cli/install.py
```diff
@@ -309,88 +316,81 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+                [
+                    spec
+                    for spec in common.specs_from_url(fpath)
+                    if spec != EXPLICIT_MARKER
+                ]
+            )
+        except UnicodeError:
+            raise CondaError(
+                "Error reading file, file should be a text file containing"
+                " packages \nconda create --help for details"
```
jezdez:
```suggestion
                "Error reading file, file should be a text file containing packages\n"
                "See `conda create --help` for details."
```

conda/cli/install.py
```diff
@@ -309,88 +316,81 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
...
+    # transform explicit packages into package records
+    explicit_packages = []
+    if fetch_explicit_packages:
+        if len(fetch_explicit_packages) == len(specs):
+            explicit_packages = get_package_records_from_explicit(
+                fetch_explicit_packages
+            )
+        else:
+            raise CondaValueError(
+                "cannot mix specifications with conda package filenames"
```
jezdez:
```suggestion
                "Cannot mix specifications with conda package filenames"
```

conda/cli/install.py
jezdez: Let's move this to a `Environment.from_cli` class method

conda/cli/install.py
```diff
@@ -400,25 +400,49 @@ def install(args, parser, command="install"):
...
+        name=args.name,
+        prefix=context.target_prefix,
+        specs=[s.strip("\"'") for s in args.packages],
+        files=args.file,
+        inject_default_packages=command == "create" and not args.no_default_packages,
+    )
+
+    # install explicit specs
+    if len(env.explicit_packages) > 0 and len(env.requested_packages) == 0:
+        return install_explicit_packages(env.explicit_packages, env.prefix)
```
jezdez, line 355: This function should also not live in `conda.misc`, maybe a topic for another refactor later

conda/cli/install.py
```diff
@@ -400,25 +400,49 @@ def install(args, parser, command="install"):
...
-        or args.update_modifier
+        getattr(args, "update_modifier", NULL)
         not in (UpdateModifier.FREEZE_INSTALLED, UpdateModifier.UPDATE_SPECS)
     ) and not newenv
 
-    update_modifier = context.update_modifier
-    if (isinstall or isremove) and args.update_modifier == NULL:
+    update_modifier = env.config.update_modifier
+    if isinstall and args.update_modifier == NULL:
         update_modifier = UpdateModifier.FREEZE_INSTALLED
```
jezdez:
```suggestion
    if isinstall and args.update_modifier == NULL:
        update_modifier = UpdateModifier.FREEZE_INSTALLED
    else:
        update_modifier = env.config.update_modifier
```

conda/cli/install.py
```diff
@@ -400,25 +400,49 @@ def install(args, parser, command="install"):
...
     if REPODATA_FN not in repodata_fns:
         repodata_fns.append(REPODATA_FN)
 
-    args_set_update_modifier = getattr(args, "update_modifier", NULL) != NULL
     # This helps us differentiate between an update, the --freeze-installed option, and the retry
     # behavior in our initial fast frozen solve
     _should_retry_unfrozen = (
-        not args_set_update_modifier
-        or args.update_modifier
+        getattr(args, "update_modifier", NULL)
```
jezdez, line 371: `args.update_modifier == NULL` is checked a little below again, should we store this in a var?

conda/cli/install.py
soapy1: This is consistent with the current behaviour https://github.com/conda/conda/blob/main/conda/cli/install.py#L354-L357. In both this change and the main branch the `create_default_packages` end up listed after the packages from the cli. So, I think this is the right approach.

conda/cli/install.py
soapy1: hmm, ya the `create_default_package` is a bit of a confusing name. I think what it means to say is "default package that should be installed only when using the `create` command". How about `add_default_packages`, `append_default_packages`, `include_default_packages`?

conda/cli/install.py
soapy1: https://github.com/conda/conda/issues/15076

conda/cli/install.py
soapy1: This is not checking if `args.update_modifier == NULL`, it is checking for if `args.update_modifier` is not in `UpdateModifier.FREEZE_INSTALLED`, `UpdateModifier.UPDATE_SPECS`.

This is a little weird to read though, will simplify it further. The default value for `args.update_modifier` is `NULL`. So this can be rewritten
```
 _should_retry_unfrozen = (
        args.update_modifier not in (UpdateModifier.FREEZE_INSTALLED, UpdateModifier.UPDATE_SPECS)
    ) and not newenv
```

conda/cli/install.py
soapy1: I like this idea! I've pushed up this change. 

IMO at this stage, it's a little bit odd. Like, the part of code that reads the specs from the file requires some functions that exist in the `conda.cli` module. It's just kinda weird to be importing stuff from the cli layer into the `conda.models.Environment` module. Once this is piece of functionality can be replaced with reading the files using the environment spec plugins, it should look a lot better. I've left some TODOs noting this. 

So, I would (softly) advocate for keeping `conda.cli.install.assemble_environment` for now, and marking it as experimental. Then, when we get to the point where the `Environment.from_cli` implementation can really look good, we can do the small refactor. This approach tries to avoid leaving pieces of work incomplete in case we are not able to get to them in a timely manner.

conda/cli/install.py
soapy1: Ops, I was wrong. This is not an ok simplification. The `update_modifier` arg is only available for install commands. So this simplification can not be made.

soapy1:
pre-commit.ci autofix

conda/cli/install.py
jezdez: Cool, thanks!

conda/cli/install.py
jezdez: I like `add_default_packages`, since that fits the "appending" nature of the code and isn't awkward

conda/cli/install.py
jezdez: Sure, want to file a ticket?

jezdez: Oh, you already did it in [`8518bce` (#15039)](https://github.com/conda/conda/pull/15039/commits/8518bce56f4c2dc49fcfd095e0ed22a588e17d79), cool

conda/models/environment.py
```diff
@@ -467,3 +474,73 @@ def from_prefix(
...
+        Create an Environment model from command-line arguments.
+
+        This method will parse command-line arguments and create an
+        Environment object. This includes: reading files provided as
+        cli arguments, and pulling EnvironmentConfig from the context.
+
+        :param args: argparse Namespace containing command-line arguments
+        :return: An Environment object representing the cli
+        """
+        specs = [s.strip("\"'") for s in args.packages]
```
jezdez:
```suggestion
        specs = [package.strip("\"'") for package in args.packages]
```

conda/models/environment.py
```diff
@@ -467,3 +474,73 @@ def from_prefix(
...
+        if add_default_packages:
+            names = {MatchSpec(pkg).name for pkg in specs}
```
jezdez:
```suggestion
            names = {MatchSpec(spec).name for spec in specs}
```

conda/models/environment.py
```diff
@@ -467,3 +474,73 @@ def from_prefix(
...
+            for pkg in context.create_default_packages:
+                pkg_name = MatchSpec(pkg).name
+                if pkg_name not in names:
+                    specs.append(pkg)
```
jezdez:
```suggestion
            for default_package in context.create_default_packages:
                if MatchSpec(default_package).name not in names:
                    specs.append(default_package)
```

conda/cli/install.py
```diff
@@ -309,116 +313,70 @@ def ensure_update_specs_exist(prefix: str, specs: list[str]):
         spec = MatchSpec(spec)
         if not spec.is_name_only_spec:
             raise CondaError(
-                f"Invalid spec for 'conda update': {spec}\nUse 'conda install' instead."
+                f"'conda update' only supports name-only spec, but received: {spec}\n"
+                f"Use 'conda install' to install a specific version of a package."
```
jezdez:
```suggestion
                f"`conda update` only supports name-only spec, but received: {spec}\n"
                f"Use `conda install` to install a specific version of a package."
```

jezdez:
pre-commit.ci autofix

PR #15040, opened 2025-07-25
Limit anaconda_upload config value to bool

timhoffm:
Docs at https://docs.conda.io/projects/conda/en/stable/configuration.html say

```
# # anaconda_upload (NoneType, bool)
```

I was wondering what the meaning of the default value None means. When inspecting the code, it turns out that conda-build only ever uses the value in a boolean context, i.e. None is equivalent to False.

https://github.com/search?q=repo%3Aconda%2Fconda-build%20anaconda_upload&type=code
https://github.com/search?q=repo%3Aconda%2Fconda-build+binstar_upload&type=code (binstar_upload is an alias)

By removing support for None, code and semantics become clearer.

### Checklist - did you ...

  Is this worth a news entry? If so, which category should this be in?
- [ ] <s>Add / update necessary tests?</s>
- [ ] <s>Add / update outdated documentation?</s> - AFAICS the relevant config documentation is autogenerated.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -535,7 +535,7 @@ class Context(Configuration):
     # conda_build
     bld_path = ParameterLoader(PrimitiveParameter(""))
     anaconda_upload = ParameterLoader(
-        PrimitiveParameter(None, element_type=(bool, NoneType)),
+        PrimitiveParameter(False, element_type=bool),
         aliases=("binstar_upload",),
     )
     _croot = ParameterLoader(PrimitiveParameter(""), aliases=("croot",))
@@ -1068,7 +1068,7 @@ class Context(Configuration):
         ) or self._use_only_tar_bz2
 
     @property
-    def binstar_upload(self) -> bool | None:
+    def binstar_upload(self) -> bool:
         # backward compatibility for conda-build
         return self.anaconda_upload
 


travishathaway:
Hi @timhoffm,

Thanks for this contribution. I'm pretty sure this a safe change to make, but we can't ignore the fact that it still carries some risks. For example, other programs which use conda as a dependency may have come to depend on the value of `None` for `anaconda_upload`.

I'm not sure the risk-to-reward ratio is sufficient enough to warrant this change, so that's why I personally cannot approve and merge it. I do understand and appreciate your desire to make a cleaner API though!

If anyone else feels differently, I would be more than happy to hear their opinions here.

timhoffm:
Thanks for the reply.  If we don’t know who else is using this and you want to guarantee 
100% backward compatibility, we can at least:
- Document that None is equivalent to False (and only exists for backward compatibility).
- Change the default to False - this is not a functional change but clearer.


Side-note: There are also some more config parameters documented as bool or None, for which I suspect they have the same issue, but I haven’t investigated their code usage yet. - I would do that if something comes out of this PR.

kenodegard:
I'm pretty sure this was intended to be a tertiary choice just like `report_errors`:
- `None` → prompt user whether to upload
- `False` → no automatic uploads
- `True` → automatic uploads

timhoffm:
Indeed, digging into the history None meant "ask" at some point in 2013: https://github.com/conda/conda/commit/f9451226

However that code was deleted in 2014 https://github.com/conda/conda/commit/48aa78f8, claiming to be unused. - Without further context (there's no PR for that commit) I assume this is an oversight, and anaconda_upload was implicitly false since then...

Anyway, I'd like to have consistent and meaningful documentation. Stating anaconda_upload is None or bool without explaining what None does is deeply confusing. Please advice/decide how to resolve this.

travishathaway:
@timhoffm,

I think we can just better document it by adding in what @kenodegard mentioned. Maybe even noting that `None` is only there for legacy support.

timhoffm:
Not sure I understand what you mean. @kenodegard noted the *intended* mode of operation. But as written, conda does respect `None` → "prompt user whether to upload" since 2014. IMHO documenting that isn't helpful. Options are:

- either document `None` → "prompt user whether to upload", but then also implement that
- or just document "`None` is equivalent to `False` and exists for backward compatibility".

PR #15059, opened 2025-07-30
Rename requirements_txt export format to specs

kenodegard:
### Description

Looking to avoid conflating pip's `requirements.txt` with conda so let's rename this exporter format a more generic "specs" format since it is composed of `MatchSpec`s.

Depends on #15058 

### Checklist - did you ...

### DIFF
--- a/conda/plugins/environment_exporters/__init__.py
+++ b/conda/plugins/environment_exporters/__init__.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Built-in conda environment exporter plugins."""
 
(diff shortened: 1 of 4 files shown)


kenodegard:
After discussing with others, we've decided to maintain the current "requirements" terminology rather than switching to "specs" for the export format. Some key considerations:

- The `requirements.txt` format is already established in the community and has existing nomenclature in the codebase
- It aligns with historical usage patterns, even if somewhat overloaded with the Python packaging ecosystem
- Introducing a new term like "specs" could add confusion rather than clarity
- Our focus should be on backward compatibility while we work toward a declarative file format

PR #15061, opened 2025-07-31
Raise exceptions when renaming or removing the `default_activation_env`

### LINKED ISSUE #87: `default_activation_env`: effects on `conda rename` and `conda remove`
Users should be able to rename the `default_activation_env` environment to something they like better, but if they do so now (they have to deactivate it first), the setting in condarc might still point to the previous name, so the shell hook will fail on subsequent runs. `conda rename` should either prevent this (refuse to rename an env configured as `default_activation_env`) or handle it gracefully (update the value of `default_activation_env`).

With `conda remove --all` (which deletes the environment), something similar happens. We should probably refuse to remove an environment configured as default (as we do with `base`). Users need to pick another default env first.

peytondmurray:
### Description

This PR raises an exception if the user tries to remove or rename the `default_activation_env`.

Closes https://github.com/conda/conda-environments-project-mgmt/issues/87.

### Checklist - did you ...


conda/cli/main_remove.py
```diff
@@ -171,6 +171,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    # Get the PrefixData for the default_activation_env; this can either be
+    # a prefix or an env name, so we try both here.
+    default_activation_env_pfd = PrefixData(context.default_activation_env)
+    if not (
+        default_activation_env_pfd.exists()
+        and default_activation_env_pfd.is_environment()
+    ):
+        default_activation_env_pfd = PrefixData.from_name(
+            context.default_activation_env
+        )
```
kenodegard: let's add a new context property (e.g., `context.default_activation_prefix`) that does the work of testing for an absolute path versus calling `locate_prefix_by_name`

kenodegard (changes requested):

conda/base/context.py
```diff
@@ -1254,6 +1254,22 @@ def auto_activate_base(self) -> bool:
...
+        prefix; otherwise it is already a prefix, so just return it.
+
+        :return: Prefix of the default_activation_env
+        """
+        from ..exceptions import EnvironmentNameNotFound
+
+        try:
+            return locate_prefix_by_name(self.default_activation_env)
+        except EnvironmentNameNotFound:
+            return self.default_activation_env
```
kenodegard: Checking if the current value is already an absolute path is a cheaper operation than checking all of the `envs_dirs` for a possible env so we should reorder this.

conda/cli/main_remove.py
```diff
@@ -171,6 +172,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         msg = "Cannot remove current environment. Deactivate and run conda remove again"
         raise CondaEnvironmentError(msg)
 
+    if args.all and Path(context.target_prefix) == Path(
+        context.default_activation_prefix
+    ):
```
kenodegard: Let's reuse the `prefix_data` that was already created above on line 164:

```suggestion
    if args.all and prefix_data == PrefixData(
        context.default_activation_prefix
    ):
```

peytondmurray: Sure, happy to make the change! Question though, why this preference over comparing the paths?

kenodegard: PrefixData's equality is more complete 🤷🏼‍♂️ but also just to mirror the validation pattern from the prior if-clause

conda/cli/main_remove.py
```diff
@@ -171,6 +171,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         msg = "Cannot remove current environment. Deactivate and run conda remove again"
         raise CondaEnvironmentError(msg)
 
+    if args.all and prefix_data == PrefixData(context.default_activation_prefix):
+        raise CondaEnvironmentError(
```
jaimergp: I don't think we want to use `CondaEnvironmentError` here (check the class composition, it inherits from stdlib's `EnvironmentError`).

conda/cli/main_remove.py
```diff
@@ -171,6 +171,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+            "Cannot remove an environment if it is the default_activation_env."
```
jaimergp:
```suggestion
            "Cannot remove an environment if it is configured as `default_activation_env`."
```

conda/cli/main_rename.py
```diff
@@ -152,6 +152,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         context.active_prefix
     ):
         raise CondaEnvException("Cannot rename the active environment")
+
+    if source_prefix_data == PrefixData(context.default_activation_prefix):
+        raise CondaEnvException(
+            "Cannot rename an environment if it is the default_activation_env."
```
jaimergp:
```suggestion
            "Cannot rename an environment if it is configured as `default_activation_env`."
```

conda/cli/main_remove.py
peytondmurray: Hmm, that's what we're using just above for when the user tries to remove the default environment. What would you prefer? `CondaEnvException`?

kenodegard: Ah, yes please deprecate `CondaEnvironmentError`

jaimergp:
Test assertions need to be updated, I think @peytondmurray

peytondmurray:
Failure appears unrelated - py39 on Windows:

```
FAILED tests/cli/test_main_notices.py::test_notices_shown_after_previous_command_error - AssertionError: assert 'Test One' in 'Channels:
 - file:///D:/a/conda/conda/tests/data/test-recipes
Platform: win-64
Solving environment: done

## Package Plan ##

  environment location: C:\Miniconda\envs\test\envs\notices-test




Downloading and Extracting Packages: ...working... done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
#
#     $ conda activate notices-test
#
#
#     $ conda deactivate

'
```

tests/base/test_context.py
```diff
@@ -756,3 +758,33 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
...
+    context.default_activation_env can either be a name or a prefix path, so we check
+    that in either case, the context.default_activation_prefix is a path.
+    """
+    name = "test_env"
+
+    # Set up the test environment
+    conda_cli("config", "--set", "default_activation_env", context.root_prefix)
+    try:
+        locate_prefix_by_name(name)
+        conda_cli("remove", "--all", "--name", name, "--yes")
```
ForgottenProgramme: Couldn't this lead to any environment named "test_env" in the user's system being deleted?

jaimergp: Yes, that's a possibility. I think we could simply change this to the usual `tmp_env` fixture. The test also changes the global config, so maybe this should be marked so it's only run on CI.

peytondmurray: > Couldn't this lead to any environment named "test_env" in the user's system being deleted?

Ahh, good point about this affecting the user's environment. The reason I didn't use `tmp_env` here is because `tmp_env` creates environments with `--prefix`. I'll rewrite this to use `tmp_env` and extract the name out of the prefix with `context.env_name` instead.

I'll also reset the `default_activation_prefix` at the end.

tests/base/test_context.py
```diff
@@ -756,3 +758,34 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
...
+        conda_cli("config", "--set", "default_activation_env", context.root_prefix)
+        assert Path(prefix) != Path(context.default_activation_prefix)
+        conda_cli("config", "--set", "default_activation_env", prefix)
+        assert Path(prefix) == Path(context.default_activation_prefix)
+        conda_cli("config", "--set", "default_activation_env", name)
+        assert Path(prefix) == Path(context.default_activation_prefix)
+
+    # Set the default_activation_env back to the original setting. If there was no
+    # output at all, it means the key was never set; remove it entirely
+    if original_setting_output == "":
```
ForgottenProgramme: @peytondmurray, you can use the [`context_aware_monkeypatch`](https://github.com/conda/conda/blob/dfe2268a5f513113473d1b2cc700bae0bca308c3/conda/testing/fixtures.py#L561) for this. 
See [example](https://github.com/conda/conda/blob/dfe2268a5f513113473d1b2cc700bae0bca308c3/conda/testing/solver_helpers.py#L980) of its usage. 
This eliminates the need to unset the environment variables manually.

ForgottenProgramme (changes requested):

tests/base/test_context.py
```diff
@@ -756,3 +758,29 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
...
+        prefix_path = Path(prefix)
+
+        monkeypatch.setenv(key, context.root_prefix)
+        assert prefix_path != Path(context.default_activation_prefix)
+
+        monkeypatch.setenv(key, prefix)
+        assert prefix_path == Path(context.default_activation_prefix)
+
+        monkeypatch.setenv(key, name)
+        assert prefix_path == Path(context.default_activation_prefix)
```
ForgottenProgramme:
```suggestion
        name = env_name(prefix)
        prefix_path = Path(prefix)

        monkeypatch.setenv(key, context.root_prefix)
        reset_context()
        assert prefix_path != Path(context.default_activation_prefix)

        monkeypatch.setenv(key, prefix)
        reset_context()
        assert prefix_path == Path(context.default_activation_prefix)

        monkeypatch.setenv(key, name)
        assert prefix_path == Path(context.default_activation_prefix)
```

ForgottenProgramme: I think the `context` object is reading the variables only once after the first import. Any changes made to variable values after the first import are not being passed to the context. 
This is probably because the context variable values are cached. 
Since we are setting the `default_activation_prefix` to different values within the same test, we need to reset_context() after each such setting. We don't need to do it after the last `monkeypatch` because `context_aware_monkeypatch` resets the context at the end of the test.

peytondmurray: Yep, this explains the errors. Naively I thought that `default_activation_prefix`, being a property of the context, would be recomputed on each access. Reading the configuration documentation it looks like there is caching in the `ParameterLoader` layer, and there are explicit instructions to call `reset_context()` after changing any environment variables (although that has its own problems...). :+1:

tests/base/test_context.py
```diff
@@ -777,10 +777,13 @@ def test_default_activation_prefix(
...
         monkeypatch.setenv(key, context.root_prefix)
+        reset_context()
         assert prefix_path != Path(context.default_activation_prefix)
 
         monkeypatch.setenv(key, prefix)
+        reset_context()
         assert prefix_path == Path(context.default_activation_prefix)
 
         monkeypatch.setenv(key, name)
+        reset_context()
```
ForgottenProgramme, line 786: We might not need this last one since the fixture runs `reset_context()` at the end anyway.

ForgottenProgramme:
One last small comment above, other than that looks good to me! 👍🏼

tests/base/test_context.py
```diff
@@ -756,3 +758,32 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
...
+    """Test that the default_activation_prefix returns the correct prefix.
+
+    context.default_activation_env can either be a name or a prefix path, so we check
+    that in either case, the context.default_activation_prefix is a path.
+    """
+    key = "CONDA_DEFAULT_ACTIVATION_ENV"
+
+    with tmp_env() as prefix:
+        name = env_name(prefix)
+        prefix_path = Path(prefix)
```
kenodegard: `prefix` is already a `pathlib.Path`

conda/base/context.py
```diff
@@ -1256,6 +1256,22 @@ def auto_activate_base(self) -> bool:
     def default_activation_env(self) -> str:
         return self._default_activation_env or ROOT_ENV_NAME
 
+    @property
+    def default_activation_prefix(self) -> PathType:
```
kenodegard: Let's always return a `pathlib.Path` here:

```suggestion
    def default_activation_prefix(self) -> Path:
```

tests/cli/test_main_remove.py
```diff
@@ -106,3 +108,31 @@ def test_remove_globbed_package_names(
 def test_remove_nonexistent_env(conda_cli: CondaCLIFixture):
     with pytest.raises(EnvironmentLocationNotFound):
         conda_cli("remove", "--prefix=/non/existent/path", "--all")
+
+
+def test_remove_all_default_activation_env(
+    conda_cli: CondaCLIFixture,
+    tmp_env: TmpEnvFixture,
+    tmpdir: str,
```
kenodegard:
```suggestion
```

tests/cli/test_main_rename.py
```diff
@@ -328,3 +328,28 @@ def test_separator_chars_on_win(conda_cli: CondaCLIFixture, env_one: str):
             f"--name={bad_env_name}",
             env_two,
         )
+
+
+def test_rename_default_activation_env(
+    conda_cli: CondaCLIFixture,
+    tmp_env: TmpEnvFixture,
+    tmpdir: str,
```
kenodegard: We prefer using `tmp_path` over `tmpdir`

```suggestion
    tmp_path: Path,
```

tests/cli/test_main_remove.py
```diff
@@ -106,3 +108,31 @@ def test_remove_globbed_package_names(
...
+):
+    """Check that removing the default_activation_env raises an exception."""
+    with tmp_env() as env:
```
kenodegard: The convention we've followed in the codebase is to call this variable `prefix`:

```suggestion
    with tmp_env() as prefix:
```

kenodegard (changes requested):

tests/base/test_context.py
kenodegard: this reset is necessary for the last assertion to work

peytondmurray: Yes, it's needed for the context to pick up the env variable.

tests/cli/test_main_rename.py
peytondmurray: Both tests that were doing this have been updated.

conda/base/context.py
peytondmurray: Okay, I've casted the return value of `context.default_activation_prefix` to `Path` as well.

tests/cli/test_main_remove.py
peytondmurray: Okay, both tests that were doing this have been renamed to use `prefix`.

ForgottenProgramme (dismissed):

tests/base/test_context.py
```diff
@@ -756,3 +758,32 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
...
+        reset_context()
+        assert prefix_path != Path(context.default_activation_prefix)
+
+        monkeypatch.setenv(key, prefix)
+        reset_context()
+        assert prefix_path == Path(context.default_activation_prefix)
+
+        monkeypatch.setenv(key, name)
+        reset_context()
+        assert prefix_path == Path(context.default_activation_prefix)
```
kenodegard:
```suggestion
        monkeypatch.setenv(key, context.root_prefix)
        reset_context()
        assert prefix != context.default_activation_prefix

        monkeypatch.setenv(key, prefix)
        reset_context()
        assert prefix == context.default_activation_prefix

        monkeypatch.setenv(key, name)
        reset_context()
        assert prefix == context.default_activation_prefix
```

kenodegard (approved):

ForgottenProgramme (approved):

jezdez (approved):

PR #15075, opened 2025-08-05
conda list: Exit with return code 1 if the queried packages can't be found

### LINKED ISSUE #15074: `conda list <package>` should exit with return code `1`  if no matches are found
### Checklist

### What is the idea?

`conda list` should return 1 if the query cannot be found in the given environment. Right now it exits with 0 regardless the result.

### Why is this needed?

To be able to use `conda list` programmatically in scripts without resorting to piping into other commands. Right now users would need to resort to `jq` or grep for similar results:

```bash
conda list numpy | jq 'length > 0'
```

### What should happen?

Given an environment with a single package, `ca-certificates`, I'd expect the following outcomes:

Success: 

```bash
> conda list
# packages in environment at ~/.local/conda:
#
# Name                     Version          Build               Channel
ca-certificates            2025.6.15        hbd8a1cb_0          conda-forge
> echo $?
0
> conda list ca-certificates
# packages in environment at ~/.local/conda:
#
# Name                     Version          Build               Channel
ca-certificates            2025.6.15        hbd8a1cb_0          conda-forge
> echo $?
0
```

Error:

```bash
> conda list numpy
# packages in environment at ~/.local/conda:
#
# Name                     Version          Build               Channel
> echo $?
1
```

### Additional Context

This might be break existing scripts, so we probably need to introduce a new flag `--check` or similar.

jaimergp:
### Description

- Closes https://github.com/conda/conda/issues/15074

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_list.py
+++ b/conda/cli/main_list.py
@@ -146,6 +146,12 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         "They are removed by default otherwise.",
     )
     p.add_argument(
+        "--check",
+        action="store_true",
+        default=False,
+        help="Exit with code 1 if the queried package(s) cannot be found in the environment.",
+    )
+    p.add_argument(
         "regex",
         action="store",
         nargs="?",
@@ -244,6 +250,9 @@ def list_packages(
                 widths[idx] = value_length
 
         packages.append(row)
+
+    if regex and not packages:
+        exitcode = 1
 
     if reverse:
         packages = reversed(packages)
@@ -379,7 +388,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     if context.json:
         format = "canonical"
 
-    return print_packages(
+    exitcode = print_packages(
         prefix,
         regex,
         format,
@@ -388,3 +397,6 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         json=context.json,
         show_channel_urls=context.show_channel_urls,
     )
+    if args.check:
+        return exitcode
+    return 0
--- a/conda/cli/main_list.py
+++ b/conda/cli/main_list.py
@@ -246,7 +246,7 @@ def list_packages(
         packages.append(row)
 
     if regex and not packages:
-        raise CondaValueError(f"No packages match the '{regex}' query.")
+        raise CondaValueError(f"No packages match '{regex}'.")
 
     if reverse:
         packages = reversed(packages)
--- a/conda/cli/main_list.py
+++ b/conda/cli/main_list.py
@@ -146,12 +146,6 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         "They are removed by default otherwise.",
     )
     p.add_argument(
-        "--check",
-        action="store_true",
-        default=False,
-        help="Exit with code 1 if the queried package(s) cannot be found in the environment.",
-    )
-    p.add_argument(
         "regex",
         action="store",
         nargs="?",
@@ -252,7 +246,7 @@ def list_packages(
         packages.append(row)
 
     if regex and not packages:
-        exitcode = 1
+        raise CondaValueError(f"No packages match the '{regex}' query.")
 
     if reverse:
         packages = reversed(packages)
@@ -388,7 +382,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     if context.json:
         format = "canonical"
 
-    exitcode = print_packages(
+    return print_packages(
         prefix,
         regex,
         format,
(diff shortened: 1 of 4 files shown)


tests/cli/test_main_list.py
```diff
@@ -249,3 +249,26 @@ def test_fields_invalid(conda_cli):
...
+    out, err, rc = conda_cli("list", f"--prefix={sys.prefix}", "does-not-exist")
+    assert rc == 0
+    out, err, rc = conda_cli(
+        "list", f"--prefix={sys.prefix}", "--check", "does-not-exist"
+    )
+    assert rc == 1
+
+    # Make sure it works with --json too
+    out, err, rc = conda_cli(
+        "list", f"--prefix={sys.prefix}", "--check", "--json", "does-not-exist"
```
ForgottenProgramme: For completeness, do we also want to add a test case with `--json` but without `--check`?

ForgottenProgramme (commented):
For completeness, do we also want to add a test case with `--json` but without `--check`?

tests/cli/test_main_list.py
jaimergp: I don't think it's necessary, imo.

ForgottenProgramme (dismissed):

conda/cli/main_list.py
```diff
@@ -245,6 +251,9 @@ def list_packages(
 
         packages.append(row)
 
+    if regex and not packages:
+        exitcode = 1
+
```
jezdez: I wonder if a better way would be to handle this like in other commands that respond to user input. They raise exceptions, which automatically set the exit code to non-zero.

```suggestion
    if regex and not packages:
        raise PackagesNotFoundError(f"No packages match the '{regex}' regular expression.")

```

jezdez (changes requested):
The current behavior is not worth retaining, since running `conda list someregexthat` doesn't even give feedback to users what is happening.

conda/cli/main_list.py
```diff
@@ -145,6 +145,12 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         help="In explicit mode, leave authentication details in package URLs. "
         "They are removed by default otherwise.",
     )
+    p.add_argument(
+        "--check",
+        action="store_true",
+        default=False,
+        help="Exit with code 1 if the queried package(s) cannot be found in the environment.",
+    )
```
kenodegard: I understand this is to avoid script breakage but I really dislike this flag and would rather we just set the exitcode correctly for `conda list regex`

conda/cli/main_list.py
kenodegard: I like this, would eliminate the need for any of the other changes

jaimergp: I like it too!

jaimergp: Had to use `CondaValueError`, though. The `PackagesNotFoundError` messaging was not adequate for this context.

news/15075-list-exit-code
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* `conda list`  will raise an exception with exit code 1 if the query did not match any packages in the target environment. (#15074 via #15075)
```
kenodegard:
```suggestion
* `conda list pattern` will raise an exception with exit code 1 if the query did not match any packages in the target environment. (#15074 via #15075)
```

tests/cli/test_main_list.py
```diff
@@ -249,3 +249,13 @@ def test_fields_invalid(conda_cli):
...
     assert "invalid-field" in str(exc)
+
+
+def test_exit_codes(conda_cli):
+    # If the package is installed, with or without --check, the exit code must be 0
+    out, err, rc = conda_cli("list", f"--prefix={sys.prefix}", "conda")
+    assert rc == 0
+
+    conda_cli(
+        "list", f"--prefix={sys.prefix}", "does-not-exist", raises=CondaValueError
```
kenodegard:
```suggestion
        "list",
        f"--prefix={sys.prefix}",
        "does-not-exist",
        raises=CondaValueError,
```

kenodegard (dismissed):

jezdez (dismissed):

jezdez:
~Seems like the tests have a false positive?~ This seems to have been a bug when not providing a regex, but also using `--full-name`.

conda/cli/main_list.py
```diff
@@ -349,7 +352,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         )
 
     regex = args.regex
-    if args.full_name:
+    if regex and args.full_name:
```
jezdez, line 355: I think this was a bug, and created `^None$` as regex, which doesn't make sense.

jezdez (approved):

PR #15081, opened 2025-08-06
Get config from environment spec files

### LINKED ISSUE #88: Determine how config from environment spec files get applied
https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html describes how config precedence works in conda for config files, cli and environment variables. However, it does not describe how environment spec config gets applied.

Conda env commands have been able to [merge config coming from files into the running context](https://github.com/conda/conda/blob/main/conda/env/installers/conda.py#L27-L53). This did not present an issue issue since environment.yaml files only are able to update the channels setting in an environment. 

With the introduction of the `EnvironmentConfig` model, new environment specs can modify much more than just the `channels` setting. So, conda must have a mechanism for applying this config in the right places.

**Team**
Driver: @soapy1 
Navigator: @ryanskeith

soapy1:
### Description

This PR suggests a way of pulling in config from an environment spec file into the context object. Adding the environment config file as a new source of context allows the `Context` class to continue to have the reponsibility of collecting all sources of config and setting correct precedence. This also ensures that all pieces of conda will have access to the same config.

In this approach, the environment file is pulled in during context initialization when the cli is started. Conda will read the input cli arg `--file` for any install/create/update commands. Conda will track both the contents of the config and the full path to the source the config originated from.

Fixes https://github.com/conda/conda-environments-project-mgmt/issues/88
Notes and other demos: https://github.com/conda/conda-environments-project-mgmt/issues/88#issuecomment-3152191454

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -85,6 +85,7 @@ if TYPE_CHECKING:
     from ..common.configuration import Parameter, RawParameter
     from ..common.path import PathsType, PathType
     from ..models.channel import Channel
+    from ..models.environment import EnvironmentConfig
     from ..models.match_spec import MatchSpec
     from ..plugins.config import PluginConfig
     from ..plugins.manager import CondaPluginManager
@@ -556,6 +557,7 @@ class Context(Configuration):
         self,
         search_path: PathsType | None = None,
         argparse_args: Namespace | None = None,
+        env_spec_config: dict[Path, EnvironmentConfig] | None = None,
         **kwargs,
     ):
         super().__init__(argparse_args=argparse_args)
@@ -565,6 +567,8 @@ class Context(Configuration):
             # for proper search_path templating when --name/--prefix is used
             CONDA_PREFIX=determine_target_prefix(self, argparse_args),
         )
+        if env_spec_config:
+            self._set_environment_spec_data(env_spec_config)
         self._set_env_vars(APP_NAME)
         self._set_argparse_args(argparse_args)
 
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -557,7 +557,7 @@ class Context(Configuration):
         self,
         search_path: PathsType | None = None,
         argparse_args: Namespace | None = None,
-        env_spec_config: dict[Path, EnvironmentConfig] | None = None,
+        env_spec_paths: list[PathsType] | None = None,
         **kwargs,
     ):
         super().__init__(argparse_args=argparse_args)
@@ -573,10 +573,35 @@ class Context(Configuration):
             # for proper search_path templating when --name/--prefix is used
             CONDA_PREFIX=determine_target_prefix(self, argparse_args),
         )
-        if env_spec_config:
-            self._set_environment_spec_data(env_spec_config)
+
+        env_spec_config = self._get_env_spec_config_source(env_spec_paths)
+        self._set_environment_spec_data(env_spec_config)
+
         self._set_env_vars(APP_NAME)
         self._set_argparse_args(argparse_args)
+
+    def _get_env_spec_config_source(self, paths: list[PathsType] | None) -> dict[PathsType, EnvironmentConfig]:
+        """Get the config from every environment spec file provided. This function
+        will return a dict of full Paths to the environment spec file to the
+        config that originated from that file.
+
+        :param paths: list of full paths to environment spec files
+        :return: dict[Path, EnvironmentConfig] of all the configs for the provided paths
+        """
+        if paths is None:
+            return {}
+
+        # Load plugin manager
+        self.plugin_manager
+
+        # Load config from environment specs
+        env_spec_config = {}
+        for path in paths:
+            str_path = str(path)
+            spec_hook = self.plugin_manager.get_environment_specifier(str_path)
+            file_env = spec_hook.environment_spec(str_path).env
+            env_spec_config[path] = file_env.config
+        return env_spec_config
 
     def post_build_validation(self) -> list[ValidationError]:
         errors = []
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -562,6 +562,12 @@ class Context(Configuration):
     ):
         super().__init__(argparse_args=argparse_args)
 
+        # The order of these calls set the config precedence. From
+        # lowest to highest priority:
+        # - condarc files
+        # - config from environment spec files (if available)
+        # - environment variables
+        # - argparse arguments
         self._set_search_path(
             SEARCH_PATH if search_path is None else search_path,
             # for proper search_path templating when --name/--prefix is used
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -28,6 +28,7 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     # defer import here so it doesn't hit the 'conda shell.*' subcommands paths
     from ..base.context import context
     from ..common.path import expand
+    from ..exceptions import EnvironmentSpecPluginNotDetected
     from .conda_argparse import do_call, generate_parser, generate_pre_parser
 
     args = args or ["--help"]
@@ -55,13 +56,24 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
 
     # if we have a file argument, then we need to read it and pass its contents to the context
     env_spec_config = {}
-    if hasattr(args, "file"):
+    # A --file might be:
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    # 3) some other kind of file
+    if getattr(args, "file", None):
+        # Ensure that all the appropriate files will get processed
         paths = [args.file] if isinstance(args.file, str) else args.file
         for path in paths:
             full_path = expand(path)
-            spec_hook = context.plugin_manager.get_environment_specifier(full_path)
-            file_env = spec_hook.environment_spec(full_path).env
-            env_spec_config[Path(full_path)] = file_env.config
+            try:
+                spec_hook = context.plugin_manager.get_environment_specifier(full_path)
+                file_env = spec_hook.environment_spec(full_path).env
+                env_spec_config[Path(full_path)] = file_env.config
+            except EnvironmentSpecPluginNotDetected:
+                # The plugin manager will throw a EnvironmentSpecPluginNotDetected error
+                # if the provided --file is not an environment spec. This can be valid,
+                # so swallow the error and continue.
+                pass
 
     context.__init__(argparse_args=args, env_spec_config=env_spec_config)
     init_loggers()
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -27,8 +27,8 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     """Entrypoint for the "subshell" invocation of CLI interface. E.g. `conda create`."""
     # defer import here so it doesn't hit the 'conda shell.*' subcommands paths
     from ..base.context import context
+    from .common import validate_file_exists
     from .conda_argparse import do_call, generate_parser, generate_pre_parser
-    from ..common.path import expand
 
     args = args or ["--help"]
 
@@ -68,9 +68,10 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     env_spec_paths = []
     if read_input_file and getattr(args, "file", None):
         if isinstance(args.file, str):
-            env_spec_paths = [Path(expand(args.file))]
+            validate_file_exists(args.file)
+            env_spec_paths = [Path(args.file)]
         else:
-            env_spec_paths = [Path(expand(file)) for file in args.file]
+            env_spec_paths = [Path(file) for file in args.file if validate_file_exists(file)]
 
     context.__init__(argparse_args=args, env_spec_paths=env_spec_paths)
     init_loggers()
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -52,21 +52,20 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     context.__init__(argparse_args=pre_args)
 
     parser = generate_parser(add_help=True)
-    
-    # HACK: this piece of code determines if conda should be trying to read an input --file
-    # arg as part of setting up the context. Conda should just do this for the following 
-    # commands.Other commands may also support a --file argument, but these files are not 
-    # environment spec files.
-    read_input_file = True if args[0] in ("env", "install", "create", "update") else False
-    
     args = parser.parse_args(args, override_args=override_args, namespace=pre_args)
 
-    # if we have a file argument, then we need to read it and pass its contents to the context
+    # conda must determine if should be trying to read an input --file arg as part of
+    # setting up the context. To make this determination, check the conda sub command 
+    # (`cmd`) that is getting executed. Conda should just do this for the following sub
+    # commands. Other commands may also support a --file argument, but these files are not 
+    # environment spec files.
+    read_input_file = True if vars(args).get("cmd", "") in ("install", "create", "update") else False
+
     env_spec_config = {}
-    # A --file might be:
+    # if we have a valid file argument, then we need to read it and pass its contents 
+    # to the context. A --file might be:
     # 1) a list of environments if coming from conda install/create/update
     # 2) a single environment spec if coming from conda env
-    # 3) some other kind of file
     if read_input_file and getattr(args, "file", None):
         # Ensure that all the appropriate files will get processed
         paths = [args.file] if isinstance(args.file, str) else args.file
@@ -78,8 +77,8 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
                 env_spec_config[Path(full_path)] = file_env.config
             except EnvironmentSpecPluginNotDetected:
                 # The plugin manager will throw a EnvironmentSpecPluginNotDetected error
-                # if the provided --file is not an environment spec. This can be valid,
-                # so swallow the error and continue.
+                # if the provided --file is not an environment spec. This can be valid
+                # for the context of loading config, so swallow the error and continue.
                 pass
 
     context.__init__(argparse_args=args, env_spec_config=env_spec_config)
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -3,8 +3,8 @@
 """Entry point for all conda subcommands."""
 
 import sys
-
 from pathlib import Path
+
 
 def init_loggers():
     import logging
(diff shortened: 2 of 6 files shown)


conda/cli/main.py
```diff
@@ -51,7 +54,28 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
     args = parser.parse_args(args, override_args=override_args, namespace=pre_args)
 
-    context.__init__(argparse_args=args)
+    # if we have a file argument, then we need to read it and pass its contents to the context
+    env_spec_config = {}
+    # A --file might be:
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    # 3) some other kind of file
+    if hasattr(args, "file"):
```
jaimergp:
```suggestion
    if getattr(args, "file", None):
```

Might save that `or []` below.

soapy1: good suggestion!

conda/base/context.py
```diff
@@ -563,6 +565,8 @@ def __init__(
             # for proper search_path templating when --name/--prefix is used
             CONDA_PREFIX=determine_target_prefix(self, argparse_args),
         )
+        if env_spec_config:
+            self._set_environment_spec_data(env_spec_config)
```
soapy1: This makes the environment spec config have higher priority than config files, but lower priority than env vars and cli args.

conda/cli/main.py
```diff
@@ -49,9 +52,37 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    # 3) some other kind of file
+    if read_input_file and getattr(args, "file", None):
+        # Ensure that all the appropriate files will get processed
+        paths = [args.file] if isinstance(args.file, str) else args.file
+        for path in paths:
+            full_path = expand(path)
+            try:
+                spec_hook = context.plugin_manager.get_environment_specifier(full_path)
```
soapy1: This part.... I think is safe. The context object gets initialized a bit earlier on with the `pre_args`. So the `plugin_manager` should be fully loaded.

conda/cli/main.py
```diff
@@ -49,9 +52,37 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     context.__init__(argparse_args=pre_args)
 
     parser = generate_parser(add_help=True)
+    
+    # HACK: this piece of code determines if conda should be trying to read an input --file
+    # arg as part of setting up the context. Conda should just do this for the following 
+    # commands.Other commands may also support a --file argument, but these files are not 
+    # environment spec files.
+    read_input_file = True if args[0] in ("env", "install", "create", "update") else False
```
soapy1: Currently the environment.yaml plugin will read any file that ends in `.yaml`/`.yml`. condarc config files are ones that have the name `condarc`, or have the file extension `.yml`/`.yaml`. This causes some awkward behaviour where conda will try to use the environment.yaml plugin to read a condarc file. This obviously does not work.

There are 2 options to resolve this:
* restrict the set of commands where conda tries to load a --file arg into the context (this solution); or
* make the [environment.yaml plugin more conservative](https://github.com/conda/conda-environments-project-mgmt/issues/68)

jaimergp: For `env`, we only care about `env create` and `env update`, right? we might want to check for those only. Also, no `remove`?

jaimergp: Another question, what's `args` if we invoke `conda` like `conda --verbose create -d python`? This might not be robust enough. Do we have access to the subcommand invoked instead of relying on its position?

conda/cli/main.py
```diff
@@ -49,9 +52,37 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
-    context.__init__(argparse_args=args)
+    # if we have a file argument, then we need to read it and pass its contents to the context
+    env_spec_config = {}
+    # A --file might be:
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    # 3) some other kind of file
+    if read_input_file and getattr(args, "file", None):
+        # Ensure that all the appropriate files will get processed
+        paths = [args.file] if isinstance(args.file, str) else args.file
```
jaimergp: Hm, I thought we had an `ensure_list` utility function but that might be conda-build only.

conda/cli/main.py
jaimergp: What happens if `conda` is invoked with `--no-plugins`? 🤔

conda/common/configuration.py
```diff
@@ -410,6 +412,44 @@ def make_raw_parameters_from_file(cls, filepath):
...
+class EnvironmentSpecificationRawParameter(RawParameter):
+    """This class represents a raw parameter originating from an environment specification file"""
+
+    def value(self, parameter_obj):
+        if isiterable(self._raw_value):
+            children_values = []
+            for i in range(len(self._raw_value)):
+                children_values.append(
+                    EnvironmentSpecificationRawParameter(self.source, self.key, self._raw_value[i])
+                )
```
jaimergp: Why not iterate directly? 
```suggestion
            for value in self._raw_value:
                children_values.append(
                    EnvironmentSpecificationRawParameter(self.source, self.key, value)
                )
```

conda/common/configuration.py
```diff
@@ -410,6 +412,44 @@ def make_raw_parameters_from_file(cls, filepath):
             return cls.make_raw_parameters(filepath, yaml_obj) or EMPTY_MAP
 
 
+class EnvironmentSpecificationRawParameter(RawParameter):
```
jaimergp, line 416: We can start adding type hints here.

conda/common/configuration.py
```diff
@@ -1499,6 +1548,18 @@ def _set_argparse_args(self, argparse_args):
         self._reset_cache()
         return self
 
+    def _set_environment_spec_data(self, env_spec_config: dict[Path, EnvironmentConfig]):
```
jaimergp:
```suggestion
    def _set_environment_spec_data(self, env_spec_config: dict[Path, EnvironmentConfig]) -> None:
```

jaimergp:
I think this solution is the most robust of all the proposed ones (strategy 2 in https://github.com/conda/conda-environments-project-mgmt/issues/88#issuecomment-3152191454). If there's consensus, I recommend choosing this one and adding tests to make sure it works as intended.

conda/cli/main.py
soapy1: `--no-plugins` will disable all external plugins. So, the built in  plugins will still be accessible. Currently, that includes environment.yaml plugin.

conda/cli/main.py
soapy1: Yep, this solution is absolutly not robust enough to handle any of the pre argument that conda has. So, specifying stuff like `conda --no-plugins env create --file env.yml` will not work as expected. I think the option of making the environment.yaml plugin more conservative is the best way to go. Will continue to explore other options too.

soapy1:
pre-commit.ci autofix

conda/cli/main.py
```diff
@@ -51,7 +54,36 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
+    # to the context. A --file might be:
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    if read_input_file and getattr(args, "file", None):
+        # Ensure that all the appropriate files will get processed
+        paths = [args.file] if isinstance(args.file, str) else args.file
+        for path in paths:
+            full_path = expand(path)
+            try:
+                spec_hook = context.plugin_manager.get_environment_specifier(full_path)
```
soapy1: At some point conda will need to do this getting of the env spec plugin again. I think there is room for this to be more efficient. Maybe can be tackled as a seperate issue.

conda/cli/main.py
```diff
@@ -51,7 +54,36 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
+                file_env = spec_hook.environment_spec(full_path).env
+                env_spec_config[Path(full_path)] = file_env.config
+            except EnvironmentSpecPluginNotDetected:
+                # The plugin manager will throw a EnvironmentSpecPluginNotDetected error
+                # if the provided --file is not an environment spec. This can be valid
+                # for the context of loading config, so swallow the error and continue.
+                pass
+
+    context.__init__(argparse_args=args, env_spec_config=env_spec_config)
```
travishathaway: Could we refactor this into a separate "private" helper function in this file? It would also be better add most of the clarifying documentation into a doc string in stead of inline comments.

We could also have a conditional before calling `context.__init__` for the third time. For example:

```python
env_spec_config = _get_env_spec_config(args)  # <-- function that holds everything

if env_spec_config:
    context.__init__(argparse_args=args, env_spec_config=env_spec_config)
```

travishathaway (commented):
Overall, I agree with this approach. Let's make sure to update the documentation on config though, so that this change is reflected:

- https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/use-condarc.html#precedence

Maybe use ChatGPT to quickly regenerate that image?

conda/cli/main.py
soapy1: It looks like you can get the command out of the initall argparse args. I think that is a good way to go.

conda/base/context.py
jezdez: Let's add a comment explaining the order of the methods called here matters

conda/cli/main.py
```diff
@@ -3,6 +3,55 @@
...
+    This will only process `--file`'s for install, create, and update commands (including
+    conda env commands).
+
+    This function will return a dict of full Paths to the environment spec file to the
+    config that originated from that file.
+
+    :param args: argparse Namespace, parsed args
+    :return: dict[Path, EnvironmentConfig] of all the configs for the provided `--file` cli args
+    """
+    from ..base.context import context
```
jezdez: This is a possible conflict of order since it imports the context instance here, but it's called before `context.__init__` when being called in `main_subshell`. At that time the context is not fully initialized. Even if you're using it to get access to the plugin manager only, it would be better to move the functionality of `_get_env_spec_config_source` into `context.__init__` method.

conda/cli/main.py
```diff
@@ -3,6 +3,55 @@
...
+    from ..exceptions import EnvironmentSpecPluginNotDetected
+
+    # conda must determine if should be trying to read an input --file arg as part of
+    # setting up the context. To make this determination, check the conda sub command
+    # (`cmd`) that is getting executed. Conda should just do this for the following sub
+    # commands. Other commands may also support a --file argument, but these files are not
+    # environment spec files.
+    read_input_file = (
+        True if vars(args).get("cmd", "") in ("install", "create", "update") else False
+    )
```
jezdez: I don't think this is the right place to do this; it seems more like an argparse level thing to do, via a validator or similar. Whether an `--file` argument exists, should be able to be checked via the argparse namespace. Doing custom subcommand parsing here again will make this less realiable in the future.

conda/cli/main.py
```diff
@@ -3,6 +3,55 @@
...
+            full_path = expand(path)
+            try:
+                spec_hook = context.plugin_manager.get_environment_specifier(full_path)
+                file_env = spec_hook.environment_spec(full_path).env
+                env_spec_config[Path(full_path)] = file_env.config
+            except EnvironmentSpecPluginNotDetected:
+                # The plugin manager will throw a EnvironmentSpecPluginNotDetected error
+                # if the provided --file is not an environment spec. This can be valid
+                # for the context of loading config, so swallow the error and continue.
+                pass
```
jezdez: That's a type of functionality to live in the plugin manager, "give me a list of env_spec_configs for the given path and keep track of where it was loaded from".

conda/cli/main.py
```diff
@@ -51,7 +100,8 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
     parser = generate_parser(add_help=True)
     args = parser.parse_args(args, override_args=override_args, namespace=pre_args)
 
-    context.__init__(argparse_args=args)
+    env_spec_config = _get_env_spec_config_source(args)
+    context.__init__(argparse_args=args, env_spec_config=env_spec_config)
```
jezdez: As I said, let's move those into `context.__init__` (or a method in the `Context` class).

conda/common/configuration.py
```diff
@@ -411,6 +413,39 @@ def make_raw_parameters_from_file(cls, filepath):
             return cls.make_raw_parameters(filepath, yaml_obj) or EMPTY_MAP
 
 
+class EnvironmentSpecificationRawParameter(RawParameter):
```
jezdez, line 416: Let's refactor this, since it's almost identical to `ArgParseRawParameter(RawParameter)` and only differs in small parts.

conda/common/configuration.py
```diff
@@ -1397,6 +1439,8 @@ def __init__(self, search_path=(), app_name=None, argparse_args=None, **kwargs):
         self._validation_errors = defaultdict(list)
 
         self._set_search_path(search_path, **kwargs)
+        if env_spec_config:
+            self._set_environment_spec_data(env_spec_config)
```
jezdez, line 1443: Could we move this into the `_set_environment_spec_data` method so that all app logic to this new configuration source is located in one method? If `_get_env_spec_config_source` moves into the `Configuration` class itself, that might be a little more streamlined.

jezdez (changes requested):

conda/cli/main.py
soapy1: hmmm, this part is a little bit tricky. There are several commands that use the `--file` argument but it doesn't always mean the same thing. For example, there is also `conda config --file`. So, conda can't just look at if a `--file` argument exists. 

The `vars(args).get("cmd", "")` is pretty extra. Conda can just check `args.cmd`. Making this:
```
 read_input_file = (
        True if args.cmd in ("install", "create", "update") else False
    )
```

This `arg.cmd` comes from the [conda argparse module](https://github.com/conda/conda/blob/main/conda/cli/conda_argparse.py#L141).

jezdez: What is tricky about it? I'm aware of `args.cmd` being the subcommand selected by argparse, combined with my other suggestions in my review to refactor it to live in the Context class instead of separately, this can simply go away.

The use of the `_get_env_spec_config_source` function in `main_subshell` is the outlier, compared to the other things happening in it. It is supposed to only deal with CLI instantiation, not app logic such as loading the env spec. I think this `read_input_file` line essentially tries to implement a new subparser just for this specific use case.

conda/cli/main.py
soapy1: To be clear, here you are saying to just move the `_get_env_spec_config_source` into `context.__init__`. Then, `context.__init__` can take a list of paths to environment specs, like `context.__init__(argparse_args=args, env_spec_paths=env_spec_paths)`?

Or, are you suggesting that `context.__init__` extracts the environment spec file information directly from the `argparse_args` argument? I would advocate for not doing that. In this case `argparse_args` represents a source of configuration. By adding some logic to the context initializer to read values of the `argparse_args` then it gains more responsibility. This will likely make refactoring and extending this functionality more difficult in the future. Because then the argparse config sources and environment spec config sources are tightly coupled.

conda/cli/main.py
soapy1: Hmmm, so the crux of this issue is that `conda config` and `conda create/install/update` and `conda env create/update` all support the `--file` command, but they sometimes mean different things. So, if conda always tries to read the provided `--file` argument with the environment spec plugins, it will throw an error, because there are not environment spec plugins that know how to read the `--file` provided to the command `conda config`. There are 3 ways I can think of to avoid this issue:

1. only try to read the `--file` argument for commands that use the `--file`  to refer to environment specs (this implementation)
2. add an alias for  the `--file` argument in conda install/create/update commands and use that to determine if conda should read the input environment specs. For example in this branch https://github.com/conda/conda/compare/main...soapy1:conda:get-config-from-env-spec-files-add-env-file-alias?expand=1
3. catch and swallow the error from environment spec detection

Even with the refactor for this to live in the context class, the overloaded meaning of `--file` presents the same issue. I agree option (1) is not great. Option (3) is not very inspiring either. Conda will eventually produce an error if it can't read the environemnt spec file. How do you feel about option (2).

conda/common/configuration.py
soapy1: Is this suggestion to introduce a relationship between `ArgParseRawParameter` and `EnvironmentSpecificationRawParameter`? Like make this `class EnvironmentSpecificationRawParameter(ArgParseRawParameter)` and then override the relevant functions?

I would advocate to not do this to avoid coupling these two ideas. While the code is very similar and it will save a few lines of code to de-duplicate this, the idea of a parameter originating from arg parse and a parameter originating from an environment spec are seperate.

conda/cli/main.py
```diff
@@ -51,7 +53,27 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
+        True if args.cmd in ("install", "create", "update") else False
+    )
+    # if we have a valid file argument, then we need to read it and pass its contents
+    # to the context. A --file might be:
+    # 1) a list of environments if coming from conda install/create/update
+    # 2) a single environment spec if coming from conda env
+    env_spec_paths = []
+    if read_input_file and getattr(args, "file", None):
+        if isinstance(args.file, str):
+            validate_file_exists(args.file)
```
soapy1, line 71: I don't think I like this. 

If the file does not exist, the plugin manager will return an error saying that no plugin was found. This causes an error for the test `tests/cli/test_env.py::test_conda_env_create_no_file ` since it expects a `EnvironmentFileNotFound` error instead. Maybe it's better for the test to be updated?

tests/env/support/with-pip.yml
```diff
@@ -1,4 +1,6 @@
 name: pip
+channels:
+  - bioconda
```
soapy1, line 3: this is just for testing, will be removed

conda/common/configuration.py
soapy1: I like that idea, can do!

conda/common/configuration.py
```diff
@@ -411,6 +413,39 @@ def make_raw_parameters_from_file(cls, filepath):
...
 
+class EnvironmentSpecificationRawParameter(RawParameter):
+    """This class represents a raw parameter originating from an environment specification file"""
+
+    def value(self, parameter_obj):
+        if isiterable(self._raw_value):
+            children_values = []
+            for value in self._raw_value:
+                children_values.append(
+                    EnvironmentSpecificationRawParameter(self.source, self.key, value)
```
soapy1, line 424: It's important to note that the `source` is a file. Previously, the only other type of config that had a file type source was coming from condarc files. There might be assumptions in the code about if a config source is a file type then it is a condarc file. That assumption would be wrong with this change. So we'll need to dig into that.

conda/base/context.py
```diff
@@ -556,18 +557,52 @@ def __init__(
...
+            return {}
+
+        # Load plugin manager
+        self.plugin_manager
+
+        # Load config from environment specs
+        env_spec_config = {}
+        for path in paths:
+            str_path = str(path)
+            spec_hook = self.plugin_manager.get_environment_specifier(str_path)
```
soapy1, line 601: Using the plugin manager here is pretty weird. When this code is called, the context is not yet fully initialized. This means that env spec plugins must not use the context object. However, they currently do.

jezdez: Yeah, I think that's an issue

conda/common/configuration.py
jezdez: fair

PR #15085, opened 2025-08-06
Fix dev/start script to handle Anaconda Terms of Service acceptance

### LINKED ISSUE #15084: dev/start script fails due to Anaconda Terms of Service acceptance requirement
## Problem

The `dev/start` script fails when setting up the conda development environment due to Anaconda's Terms of Service (TOS) acceptance requirement. The error occurs because:

1. The script downloads and installs a fresh miniconda installation from `https://repo.anaconda.com/miniconda/`
2. This fresh installation comes with default Anaconda channels (`https://repo.anaconda.com/pkgs/main` and `https://repo.anaconda.com/pkgs/r`) that require TOS acceptance
3. The script attempts to create environments and install packages without first accepting the TOS
4. This results in a `CondaToSNonInteractiveError` with the message:

```
CondaToSNonInteractiveError: Terms of Service have not been accepted for the following channels. Please accept or remove them before proceeding:
    • https://repo.anaconda.com/pkgs/main
    • https://repo.anaconda.com/pkgs/r
```

## Root Cause

The issue stems from Anaconda's recent requirement for explicit TOS acceptance for their channels. The `dev/start` script was not updated to handle this requirement, causing it to fail for new developers setting up the conda development environment.

## Solution

Based on the [Anaconda TOS plugin documentation](https://www.anaconda.com/docs/getting-started/tos-plugin), we should modify the `dev/start` script to accept the TOS.

## Impact

This fix ensures that new developers can successfully set up the conda development environment without manual intervention, improving the developer experience and reducing setup friction.

## Related Documentation

- [Anaconda TOS Plugin Documentation](https://www.anaconda.com/docs/getting-started/tos-plugin)
- [GitHub Workflows](https://github.com/conda/conda/blob/main/.github/workflows/tests.yml) (which handle TOS differently in CI/CD)

ryanskeith:
Fixes #15084

This PR adds automatic TOS acceptance to the dev/start script to resolve the CondaToSNonInteractiveError that was preventing new developers from setting up the conda development environment.

The script now calls `conda tos accept` at key points:
- After miniconda installation
- Before environment creation
- Before updates and package installation

This ensures the development environment can be set up without manual TOS intervention.

### DIFF
--- a/dev/start
+++ b/dev/start
@@ -218,11 +218,25 @@ if ! [ -f "${_DEVENV}/conda-meta/history" ]; then
         echo "Error: failed to install development environment" 1>&2
         return 1
     fi
+
+    # Accept TOS for all channels
+    echo "Accepting Terms of Service for Anaconda channels..."
+    if ! PYTHONPATH="" "${_BASEEXE}" tos accept > /dev/null; then
+        echo "Warning: failed to accept Terms of Service automatically" 1>&2
+        echo "You may need to manually accept the Terms of Service when prompted" 1>&2
+    fi
 fi
 
 # create empty env if it doesn't exist
 if ! [ -d "${_ENV}" ]; then
     echo "Creating ${_NAME}..."
+
+    # Try to accept TOS before creating environment
+    echo "Ensuring Terms of Service are accepted..."
+    if ! PYTHONPATH="" "${_BASEEXE}" tos accept > /dev/null 2>&1; then
+        echo "Warning: Could not automatically accept TOS. You may be prompted to accept Terms of Service." 1>&2
+    fi
+
     if ! PYTHONPATH="" "${_BASEEXE}" create --yes --quiet "--prefix=${_ENV}" > /dev/null; then
         echo "Error: failed to create ${_NAME}" 1>&2
         return 1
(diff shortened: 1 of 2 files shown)


travishathaway:
Thanks for trying to find a solution to this!

jezdez:
A better way would be to set the env var `CONDA_PLUGINS_AUTO_ACCEPT_TOS=yes` because running `conda tos accept ..` will actually write to disk that you're consenting, and that shouldn't be something that is done as part of a dev script as it removes the ability to decide this individually.

dev/start
```diff
@@ -233,17 +247,29 @@ fi
...
+        echo "Warning: Could not automatically accept TOS for package installation. You may be prompted to accept Terms of Service." 1>&2
+    fi
+
     if ! PYTHONPATH="" "${_BASEEXE}" install \
         --yes \
         --quiet \
         "--prefix=${_ENV}" \
         --override-channels \
-        --channel=defaults \
+        --channel=conda-forge \
```
jezdez, line 272: That's not a change we should be doing

jezdez (changes requested):
Please switch to an env var approach

kenodegard:
> A better way would be to set the env var `CONDA_PLUGINS_AUTO_ACCEPT_TOS=yes` because running `conda tos accept ..` will actually write to disk that you're consenting, and that shouldn't be something that is done as part of a dev script as it removes the ability to decide this individually.

A file is still written to disk.

ryanskeith:
@travishathaway @jezdez @kenodegard , I put this in draft mode because I wanted to poll if we wanted to make the script interactive to prompt the dev user to accept or reject.  If accepted, we would use main if rejected, it would use conda-forge.  Thoughts on this enhancement?

travishathaway:
@ryanskeith,

I like the idea of being given a choice interactively for this.

jezdez:
The acceptance of the Terms of Service is a legal act that needs to consciously be done by a person and not automated wholesale. We've added `CONDA_PLUGINS_AUTO_ACCEPT_TOS` to the plugin to make it easier for users that use conda in CI to not be prompted (which usually doesn't work well in CI) and instead indicate their consent without interactively calling a command.

An additional prompt, that will even switch to conda-forge if denied automatically, will not improve the developer experience, since conda-forge's packages don't work identically to Anaconda's. We test both in CI because of this, since conda has to work for both distributions. Choosing to develop conda OSS with the conda-forge distribution should be a specific decision during the creation of the developer environment.

Additionally, the location where the plugin stores that consent is configurable, which means you can use `--site` for example, to store your consent globally on your computer.

```
Local Metadata Storage Location:
  --site                System-wide storage location.
  --system              Conda installation storage location.
  --user                User storage location. *(default)
  --env                 Conda environment storage location.
  --tos-root TOS_ROOT   Custom storage location.
```

What I'd be interested in is an option and the appropriate documentation for the `dev/start` and `dev\start.bat` scripts to set up the developer environment with conda-forge from the start; right now it's hardcoded `defaults`, which means Anaconda's repo.

jezdez:
I've opened #15163 as an alternative to this

jezdez:
Closing in favor of #15163.

PR #15100, opened 2025-08-12
Catch LockError Exception during partial download

### LINKED ISSUE #13921: Add an error message if file locking is not supported
### Checklist

### What is the idea?

Add an error message if file locking is not supported.

### Why is this needed?

Conda may be running a few hours without indications of a deadlock because file locking is not supported.

### What should happen?

If file locking is not supported and required by conda, it should fail early on instead of running for a few hours before failing.

### Additional Context

I was unable to install new packages on my conda environment and though it may be broken. After some trial and error, I decided to install move to a fresh anaconda installation, and create a new environment but conda failed after a few hours, supposedly because file locking was not available on our system.
If conda requires file locking, I would like this feature to be Implemented.

Here is the error I got:
```bash
...
      File ".../anaconda3/lib/python3.11/site-packages/conda/gateways/disk/lock.py", line 59, in __enter__
        fcntl.lockf(
    OSError: [Errno 37] No locks available

`$ .../anaconda3/bin/conda create -n fseft python=3.11 -y`

  environment variables:
                 CIO_TEST=<not set>
               CONDA_ROOT=.../anaconda3
           CURL_CA_BUNDLE=<not set>
               LD_PRELOAD=<not set>
                     PATH=.../anaconda3/bin:/usr/local/sbin:/usr/l
                          ocal/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/sn
                          ap/bin
       REQUESTS_CA_BUNDLE=<not set>
            SSL_CERT_FILE=<not set>

     active environment : None
       user config file : .../.condarc
 populated config files :
          conda version : 24.1.2
    conda-build version : 24.1.2
         python version : 3.11.7.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=cascadelake
                          __conda=24.1.2=0
                          __cuda=11.4=0
                          __glibc=2.27=0
                          __linux=4.15.0=0
                          __unix=0=0
       base environment : .../anaconda3  (writable)
      conda av data dir : .../anaconda3/etc/conda
  conda av metadata url : None
          package cache : .../anaconda3/pkgs
                          .../.conda/pkgs
       envs directories : .../anaconda3/envs
                          .../.conda/envs
               platform : linux-64
             user-agent : conda/24.1.2 requests/2.31.0 CPython/3.11.7 Linux/4.15.0-197-generic ubuntu/18.04.6 glibc/2.27 solver/libmamba conda-libmamba-solver/24.1.0 libmambapy/1.5.6 aau/0.4.3 c/LOjItEifpMzBAGMKrbUbwg s/OFnsg3lTuePlGdjAKdYoxg e/Wb8hgRYPNk76MKN_MSNJ9w
                UID:GID : 8575:8212
             netrc file : None
           offline mode : False
```

ForgottenProgramme:
### Description
resolves https://github.com/conda/conda/issues/13921
Make sure LockErrors are also caught in the partial download logic

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/download.py
+++ b/conda/gateways/connection/download.py
@@ -25,6 +25,7 @@ from ...exceptions import (
     CondaHTTPError,
     CondaSSLError,
     CondaValueError,
+    LockError,
     ProxyError,
     maybe_raise,
 )
@@ -252,6 +253,9 @@ def download_partial_file(
             partial_path.unlink()
         raise
     except ChecksumMismatchError:
+        partial_path.unlink()
+        raise
+    except LockError:
         partial_path.unlink()
         raise
 


conda/gateways/connection/download.py
```diff
@@ -254,6 +255,9 @@ def check(target):
     except ChecksumMismatchError:
         partial_path.unlink()
         raise
+    except LockError:
+        partial_path.unlink()
+        raise
```
jezdez, line 260:
```suggestion
    except (ChecksumMismatchError, LockError):
        partial_path.unlink()
        raise
```

since both error states do the same error handling, we can combine them into one clause

jezdez (changes requested):

jezdez:
@ForgottenProgramme actually, from the original ticket you can see the exception raised is a `OSError` exception, I think `LockError` may actually not be used (or at least I can't find where)!

kenodegard:
> @ForgottenProgramme actually, from the original ticket you can see the exception raised is a `OSError` exception, I think `LockError` may actually not be used (or at least I can't find where)!

@jezdez this was just changed in https://github.com/conda/conda/pull/14980

conda/gateways/connection/download.py
kenodegard: If we fail to lock on a partial path that means some other process has the lock so we should **not** unlink the partial path

jezdez:
> > @ForgottenProgramme actually, from the original ticket you can see the exception raised is a `OSError` exception, I think `LockError` may actually not be used (or at least I can't find where)!
> 
> 
> 
> @jezdez this was just changed in https://github.com/conda/conda/pull/14980

Ah excellent, I hadn't seen this yet, thanks for the pointer!

conda/gateways/connection/download.py
jezdez: Ok, ignore my suggestion then, it needs a different one

conda/gateways/connection/download.py
```diff
@@ -254,6 +255,9 @@ def check(target):
     except ChecksumMismatchError:
         partial_path.unlink()
         raise
+    except LockError:
+        partial_path.unlink()
```
jezdez, line 259:
```suggestion
```
@kenodegard Is this what you mean?

kenodegard: yes, I actually don't think we need this PR at all? lock failures now raise a `LockError` so we already get a pretty printed error instead of an error report

ForgottenProgramme (commented):
Should we close this then?
@kenodegard , @jezdez

ForgottenProgramme:
@kenodegard  and I discussed this offline and concluded that this change is actually not needed. LockErrors are now already being handled in the `lock.py` module.

PR #15110, opened 2025-08-18
Add support for `PackageRecord.requested_specs`

jaimergp:
### Description

Add this new field, discovered in https://github.com/conda/rattler/issues/1595#issuecomment-3176675370.

- Closes https://github.com/conda/conda/issues/15111

### Checklist - did you ...

### DIFF
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -934,7 +934,7 @@ class CreatePrefixRecordAction(CreateInPrefixPathAction):
         package_info,
         target_prefix,
         requested_link_type,
-        requested_spec: list[MatchSpec],
+        requested_spec: MatchSpec | list[MatchSpec],
         all_link_path_actions,
     ):
         extracted_package_dir = package_info.extracted_package_dir
@@ -958,7 +958,7 @@ class CreatePrefixRecordAction(CreateInPrefixPathAction):
         target_prefix,
         target_short_path,
         requested_link_type,
-        requested_spec: list[MatchSpec],
+        requested_spec: MatchSpec | list[MatchSpec],
         all_link_path_actions,
     ):
         super().__init__(


conda/core/path_actions.py
```diff
@@ -996,6 +996,7 @@ def paths_from_action(link_path_action):
             # self.package_info.index_json_record,
             self.package_info.package_metadata,
             requested_spec=str(self.requested_spec),
+            requested_specs=(str(self.requested_spec),),
```
jezdez: Should this only turned into a `str` once per run?

tests/test_create.py
```diff
@@ -1855,6 +1855,7 @@ def test_conda_pip_interop_pip_clobbers_conda(
                 ],
                 "paths_version": 1,
             },
+            "requested_specs": [],
```
jezdez: Why doesn't the singular field show up here?

tests/models/test_records.py
```diff
@@ -241,3 +244,15 @@ def test_record_spec_strings_inheritance(record_class, extra_kwargs):
...
+@pytest.mark.integration
+def test_requested_spec(tmp_env):
+    specs = ("ca-certificates", "ca-certificates>=2025")
+    with tmp_env(*specs) as prefix:
+        prefix_data = PrefixData(prefix)
+        record = prefix_data.get("ca-certificates")
+        # conda classic and conda-libmamba-solver merge the incoming match-specs
+        merged_spec = str(MatchSpec.merge(specs)[0])
+        assert record.requested_spec == merged_spec
+        assert record.requested_specs == (merged_spec,)
```
jezdez: I don't follow, wouldn't we want to have the unmerged specs here?

jaimergp: Yep, but this behaviour is not enabled yet in the other solvers because they always merged. We have to first make this field available, then we can start populating it correctly.

tests/test_create.py
jaimergp: Because it's null-y value is None and those are omitted by the dumper

jaimergp: Turns out one can set `nullable=True` in the field and then these go away. Should be simpler now.

tests/models/test_records.py
jaimergp: I also introduced a couple of attributes to help downstream solvers use the unmerged specs in their inputs.

jaimergp: Actually turns out we can make this happen in conda core already, no need for downstream solvers to do anything special.

jaimergp:
@jezdez, PTAL when you have time, thanks!

PR #15115, opened 2025-08-20
Conda should provide a helpful error message for invalid `create_default_packages` setting

soapy1:
### Description

Conda is unable to install a mixture of explicit packages (eg. `https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda`) and package specs (eg. `python > 3.10`). If a user tries to do this, they get an error message `CondaValueError: cannot mix specifications with conda package filenames`.  

A user may also set a `create_default_packages` setting. This will inject the specified default packages into an environment created with the `conda create` and `conda env create` commands. If this setting contains an explicit package, the user will see the same error message, `CondaValueError: cannot mix specifications with conda package filenames`.  However, it would be better if this error message was more helpful. 

This PR adds more context to the error message presented to the user when the `create_default_packages` setting contains a reference to an explicit package. 

Consider the condarc:
```
create_default_packages:
  - https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda
``` 

When trying to create an environment. 

**Previously**, this would produce an error when creating an environment with a package specified:
```
$ conda create -n test-default python                               

CondaValueError: cannot mix specifications with conda package filenames
```

And, it would create an environment with the explicit package if no package is specified:
```
$ conda create -n test-default      

## Package Plan ##

  environment location: /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/test-default

  added / updated specs:
    - pkgs/main/linux-64::ca-certificates==2025.2.25=h06a4308_0


. . . 
```

**Now**,  conda will produce an error:
```
$ conda create -n test-default

CondaValueError: Conda setting `create_default_packages` must not include references to explicit packages. Found explicit package reference 'https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda'
```

ref: https://github.com/conda/conda-environments-project-mgmt/issues/71

### Checklist - did you ...


conda/models/environment.py
```diff
@@ -517,6 +517,11 @@ def from_cli(
         if add_default_packages:
             names = {MatchSpec(spec).name for spec in specs}
             for default_package in context.create_default_packages:
+                if is_package_file(default_package):
+                    raise CondaValueError(
```
soapy1: Is this too strict? Should conda allow explicit packages in the `create_default_packages` setting?
* In every case that a user creates an environment and wants to specify packages, it will cause an error. For example `conda create -n myenv python`.
* It's maybe valid if the case where the user is not specifying packages to go in the created environment. For example `conda create -n myenv`
  * But, installing an explicit package means that none of the dependencies get installed. So if the user specifies an explicit reference to, say a flask package, the environment will be in a weird state.

soapy1: Another option is to have conda doctor check for if this setting is valid. Would that be a better alternative solution?

jezdez: I'd not raise the error immediately here, but first iterate over all of them to check if they are a package file, so that the `CondaValueError` can show a list if needed, preventing the users from having to run conda multiple times.

jezdez: Otherwise I think this is a fine exception. Maybe check if there is a more specific exception class in our small exception class collection?

conda/models/environment.py
```diff
@@ -517,6 +517,11 @@ def from_cli(
...
+                        "Conda setting `create_default_packages` must not include references to explicit packages. "
```
ryanskeith: TIL: this is a setting.  I am not sure how long this one has been around but I had to look it up which took me a bit of time.  Could we make this more descriptive perhaps?  Maybe `Conda config setting` or Conda Setting ... in .condarc?

ryanskeith (approved):
This looks good to me. I humbly suggest we could add more information to the message to help users understand what to do or where to go next.

conda/models/environment.py
danyeaw: Hi @soapy1,

What if we did a warn and continue approach to educate them but not block them? This way they would get a warning that their config is bad, but it would ignore the explicit packages, try to install other non-explicit packages in `create_default_packages` and keep going.

The warning could be something like `f"Skipping invalid explicit package in create_default_packages: '{default_package}'. Use package names like 'numpy' or specs like 'numpy>=1.20' instead."`

conda/models/environment.py
soapy1: Ya, this is true. How about adding some directions on how to fix the situation? Could add a note like: `Try using the command `conda config --show-sources` to verify your conda configuration.`

> I had to look it up

The conda config command has the ability to share a description for conda settings.  Though, looking it up certainly provides more information.

```
$ conda config --describe create_default_packages
# # create_default_packages (sequence: primitive)
# #   env var string delimiter: ','
# #   Packages that are by default added to a newly created environments.
# # 
# create_default_packages: []

```

conda/models/environment.py
kenodegard: I agree with collecting all of the failures into one message instead of failing on the first invalid package

I also think we should move this logic over into the context object and make `context.create_default_packages` a property that performs this validation

soapy1: Thanks for the feedback all!

> but first iterate over all of them to check if they are a package file

I like this idea! Added :fireworks: 

> What if we did a warn and continue approach to educate them but not block them?

I like this idea too. I think in general ignoring user config, even if invalid, can result in pretty hairy situations. For this case, it is always invalid for conda to be mixing explicit package references and specs. Having an explicit spec as a value in `create_default_packages` is invalid, and the solution is to always remove it. So, I think in this case removing the explicit package and just warning instead of erroring out is ok.

soapy1: > I also think we should move this logic over into the context object and make context.create_default_packages a property that performs this validation

Oh! I was also thinking about how the validation can be done in the context object. Just couldn't figure it out. Making it a property makes sense. I like that approach, will give it a go.

conda/base/context.py
```diff
@@ -1256,6 +1258,29 @@ def auto_activate_base(self) -> bool:
...
+            if is_package_file(package):
+                explicit_packages.append(package)
+            else:
+                default_packages.append(package)
+
+        if explicit_packages:
+            warnings.warn(
+                f"Ignoring invalid packages in `create_default_packages`: {dashlist(explicit_packages)}\n"
+                "\nExplicit package are not allowed, use package names like 'numpy' or specs like 'numpy>=1.20' instead.\n"
+                "Try using the command `conda config --show-sources` to verify your conda configuration.\n",
```
kenodegard: F-string all of them so they're identically indented and move all newlines to the end for consistency:

```suggestion
                f"Ignoring invalid packages in `create_default_packages`: {dashlist(explicit_packages)}\n"
                f"\n"
                f"Explicit package are not allowed, use package names like 'numpy' or specs like 'numpy>=1.20' instead.\n"
                f"Try using the command `conda config --show-sources` to verify your conda configuration.\n",
```

conda/base/context.py
```diff
@@ -1256,6 +1258,29 @@ def auto_activate_base(self) -> bool:
...
+        from ..common.io import dashlist
+        from ..common.path import is_package_file
+
+        explicit_packages = []
+        default_packages = []
+        for package in self._create_default_packages:
+            if is_package_file(package):
+                explicit_packages.append(package)
+            else:
+                default_packages.append(package)
```
kenodegard: FYI we have `conda.common.iterators.groupby_to_dict` that does the same thing

tests/cli/test_main_create.py
```diff
@@ -0,0 +1,35 @@
...
+        "https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda",
+    )
+    reset_context()
+    with pytest.warns(UserWarning) as warning:
+        conda_cli("create", "--prefix", tmp_path, "--yes")
+
+        assert (
+            "Ignoring invalid packages in `create_default_packages`: \n  - https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda\n\nExplicit package are not allowed, use package names like 'numpy' or specs like 'numpy>=1.20' instead.\nTry using the command `conda config --show-sources` to verify your conda configuration.\n"
+            in [str(warn.message) for warn in warning]
+        )
```
kenodegard: I'm thinking this should be more targeted and just test getting `context.create_default_packages`?

tests/cli/test_main_create.py
```diff
@@ -0,0 +1,35 @@
...
+    monkeypatch: MonkeyPatch,
+    conda_cli: CondaCLIFixture,
+    tmp_path: Path,
+):
+    monkeypatch.setenv(
+        "CONDA_CREATE_DEFAULT_PACKAGES",
+        "https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda",
+    )
+    reset_context()
+    with pytest.warns(UserWarning) as warning:
```
kenodegard: Use `match` key instead of needing `assert`:

```suggestion
        (url := "https://repo.anaconda.com/pkgs/main/linux-64/ca-certificates-2025.2.25-h06a4308_0.conda"),
    )
    reset_context()
    with pytest.warns(UserWarning, match=rf"Ignoring invalid packages.+{url}"):
```

kenodegard (changes requested):

soapy1:
pre-commit.ci autofix

kenodegard (approved):

jezdez (approved):

PR #15118, opened 2025-08-21
Create constant for all reserved base environment names

soapy1:
### Description

This PR adds a new constant `RESERVED_ENV_NAMES` that is a list of all the environment names that are reserved by conda. This includes "base" and "root". Conda code can refer to this new constant when doing environment name validations.

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -120,6 +120,10 @@ DEFAULT_CUSTOM_CHANNELS: Final = {
 DEFAULT_CHANNELS: Final = DEFAULT_CHANNELS_WIN if on_win else DEFAULT_CHANNELS_UNIX
 
 ROOT_ENV_NAME: Final = "base"
+RESERVED_ENV_NAMES: Final = (
+    ROOT_ENV_NAME,
+    "root",
+)
 UNUSED_ENV_NAME: Final = "unused-env-name"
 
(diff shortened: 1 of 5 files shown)


soapy1:
This is just a small refactor, does it need a news entry?

conda/base/constants.py
```diff
@@ -120,6 +120,10 @@
 DEFAULT_CHANNELS: Final = DEFAULT_CHANNELS_WIN if on_win else DEFAULT_CHANNELS_UNIX
 
 ROOT_ENV_NAME: Final = "base"
+RESERVED_ENV_NAMES: Final = (
```
ForgottenProgramme, line 123: Will we need to update this list following the `conda self` work? Since "default" will be the environment name that conda will create upon migration of the base env. And we probably would want to make "default" also a reserved name? 🤔

danyeaw: The new proposal is to call this new environment `base-copy`. We found during testing that users were very confused calling an environment default since channels are already called that.

kenodegard:
> This is just a small refactor, does it need a news entry?

Yeah, probably don't need a news file

danyeaw (approved):
Thanks @soapy1 - nice refactoring to make use of constants.

jezdez (approved):

PR #15120, opened 2025-08-22
Add environment.yaml schema validations as per cep-0024

soapy1:
### Checklist - did you ...


soapy1:
pre-commit.ci autofix

tests/env/support/empty_env.yml
```diff
@@ -1,3 +1,4 @@
 channels:
   - conda-forge
   - defaults
+dependencies:
```
soapy1: This is a change in behaviour. The environment.yaml cep says that `dependencies` is a required field. However this has never been enforced in the past.

> dependencies
Required, list[str | dict[str, Any]].

ref: https://github.com/conda/ceps/blob/main/cep-0024.md#dependencies

tests/env/support/invalid_keys.yml
```diff
@@ -1,6 +1,6 @@
 names: nlp
 chanels:
   - anaconda
-dependecies:
+dependencies:
```
soapy1: Same here, this is a change in behaviour caused by enforcing `dependencies` being a required field.

soapy1:
pre-commit.ci autofix

tests/env/support/invalid_keys.yml
jezdez: Hmm, this test fixture file exists to test invalid keys via the `from_file` loader, why do you fix it?

tests/env/support/empty_env.yml
jezdez: I think this is a bug in CEP 24

jezdez:
To be clear, rolling out this change will break a lot of people's user experience. I don't know why the CEP 24 process didn't catch this changing behavior.

conda/env/env.py
```diff
@@ -2,26 +2,171 @@
...
+
+def name_validation(name: str | None):
+    """Validate the contents of the name field.
+
+    Will ensure:
+      * name is a string
+      * is not the name of a reserved environment
+    """
+    field_type_validation("name", name, str)
+    if name in RESERVED_ENV_NAMES:
```
soapy1: This validation is maybe not appropriate here. This should not verify that the contents are valid, just that the form of the input file is correct?

soapy1:
> To be clear, rolling out this change will break a lot of people's user experience.

Yes, this introduces some breaking changes. Sorry, I should have outlined all of them as part of the description. Here is a breakdown of the change in behaviour.


This change will introduce breaking changes for the following senarios:
* environment.yaml file missing (or typo)  `dependencies` field.
  * _current behaviour_: conda will ignore this and create an empty environment
  * _proposed behaviour_: conda will not detect the file as a valid environment.yaml and will probably do nothing
* environment.yaml with invalid types for the `name`, `prefix`, `channels`, `varaibles` fields (if present).
  * _current behaviour_: conda will ignore all ill formatted fields, probably creating a different environment from the one the user intended to make
  * _proposed behaviour_: conda will not detect the file as a valid environment.yaml and will probably do nothing
* environment.yaml with reserved env name as the value for the `name` key
  * _current behaviour_: conda will produce an error about the name being invalid
  * _proposed behaviour_: conda will not detect the file as a valid environment.yaml and will probably do nothing

This change will not introduce breaking changes for:
* environment.yaml with extra keys. Currently, users can specify whatever extra keys they want in their environment.yaml. conda will ignore these and continue processing the request. Thie behaviour will stay the same.


I think it's important to note that more strictly enforcing validations on environment.yaml to any extent may cause breaking changes. But, without enforicing stricter validations, then the environment.yaml plugin will always report being able to read `.yaml`/`.yml` environment specs. This breaks environment spec plugin detection.

jezdez:
The proposed behaviors cannot go in as planned since they are all backward incompatible. The user experience is changing by enforcing the well-intentioned CEP.

> proposed behaviour: conda will not detect the file as a valid environment.yaml and will probably do nothing

What that means is the conda will not work as expected and create a lot of friction for those users that have environment files that are not compliant. Conda is already seen as user hostile, so we can't just change things like this. In fact the deprecation policy requires us to providing an off-ramp for deprecating behavior like this.

I think what we're needing here is:

- implementing this in an opt-in env spec that is more CEP compliant that users can switch to (think: `conda config --set env_spec cep-24` or similar)
- inform the user that their `environment.yml` is not compliant if we can detect it. Use the deprecations framework and [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning)
- the current env spec for `environment.yml` needs to know what compliant means basically
- roll this out over a number of releases per [CEP 9](https://conda.org/learn/ceps/cep-0009/)
- assuming we go further into implementing a `conda migrate` command via conda-self, we could provide a `conda migrate cep24 --input=my-old-environment.yml --output=environment.yml` functionality

soapy1:
@jezdez  I think you outline a good plan! Couple of clarifying questions:

1. Is this approach of making the `environment.yaml` plugin more conservative the solution we want to go with? For reference that'll be in line with the [responsible authors idea outlined in the design doc](https://docs.google.com/document/d/1LtB2QKuF72FsnuBzWYsw4wOL2w_boyvYys5taiOE2zE/edit?tab=t.0#heading=h.70acvvf5bevy).
2. Should we be updating the spec and this PR so that the `dependencies` key is not a required field for an `environment.yaml` file.

If the answer to (1) is yes, lets update this PR to start implementing the plan you've outlined.

soapy1:
Ok so, I've pushed up a change that makes conda warn that the input file is invalid instead of throwing an error. After implementing this and trying it out I think just warning on schema errors is kind of a clunky solution. Following the deprecation cycle, this means that env spec detection won't be working properly until the deprecation cycle for this is complete. In the mean time, any user trying to use a `.yaml` file for their env spec will need to specify a preferred env spec plugin (which is also the current behaviour). That is, in order for environment spec plugin detection to work, the environment.yaml plugin needs to report that it can not read a given file.

While this approach is more in line with the deprecation policy in cep-9, I think this behaviour is not super user friendly to folk who want to be adopting new features (like conda-lockfiles, which are currently using .yaml files). 

There are a few other options that I can think of to resolve this:
* Make the environment.yaml plugin a special case that runs only if no other plugin is able to read the input file. We'll still want to do the deprecation cycle as suggested in @jezdez's [comment].(https://github.com/conda/conda/pull/15120#issuecomment-3227175472).
* Add a "strict" cep-24 plugin that is enabled by default and disable the existing environment.yaml plugin by default. We'll need to build some new types of config to allow the user to enable/disable specific plugins to make this work.

soapy1:
pre-commit.ci autofix

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+    """
+    field_type_validation("prefix", prefix, str)
+
+
+def name_validation(name: str | None):
+    """Validate the contents of the name field.
+
+    Will ensure:
+      * name is a string
+      * is not the name of a reserved environment
```
kenodegard: it does?

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+
+    :param dict data: The contents of the environment.yaml
+    :returns errors: A list of EnvironmentFileInvalid exceptions that occurred during validation
+    """
+    errors = []
+    data_keys = data.keys()
+    # Ensure all required keys are present
+    for field in REQUIRED_KEYS:
+        if field not in data_keys:
+            errors.append(EnvironmentFileInvalid(f"Missing required field '{field}'"))
```
kenodegard:
```suggestion
    # Ensure all required keys are present
    for field in set(REQUIRED_KEYS).difference(data):
        errors.append(EnvironmentFileInvalid(f"Missing required field '{field}'"))
```

Or if we can make `REQUIRED_KEYS` a `frozenset` that would be even better IMO

conda/env/env.py
```diff
@@ -45,7 +191,7 @@ def validate_keys(data, kwargs):
             print(f" - {key}")
         print()
 
-    deps = data.get("dependencies", [])
+    deps = data.get("dependencies", []) or []
```
kenodegard:
```suggestion
    deps = data.get("dependencies") or []
```

conda/env/env.py
```diff
@@ -124,12 +270,35 @@ def from_environment(
...
         raise EnvironmentFileEmpty(filename)
+
+    # Perform schema validation. This will output a warning for any invalid schema.
+    errors = get_schema_errors(data)
+    if errors:
+        # Warn for all the schema errors in the environment
+        deprecated.topic(
+            "26.3",
+            "26.9",
+            topic="This environment file was found to not be compliant with cep-0024.",
```
kenodegard: `topic` needs to be a sentence fragment since we append "is pending deprecation and will be removed in..."

```suggestion
            topic="Using a non-compliant CEP-0024 environment file",
```

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+    for field in REQUIRED_KEYS:
+        if field not in data_keys:
+            errors.append(EnvironmentFileInvalid(f"Missing required field '{field}'"))
+
+    # Run validations on all the relevant fields
+    for key, validator in SCHEMA_VALIDATORS.items():
+        try:
+            validator(data.get(key, None))
+        except EnvironmentFileInvalid as err:
+            errors.append(err)
```
kenodegard, line 167: Since we've already checked for required keys shouldn't we just skip any validator checks if the key is missing?

```suggestion
    for key, validator in SCHEMA_VALIDATORS.items():
        try:
            validator(data[key])
        except KeyError:
            pass
        except EnvironmentFileInvalid as err:
            errors.append(err)
```

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+    "channels",
+    "variables",
+)
+VALID_KEYS = (*REQUIRED_KEYS, *OPTIONAL_KEYS)
+
+
+def field_type_validation(field_name: str, value: Any | None, value_type: Any) -> None:
+    """Validates the type of a value"""
+    if not value:
+        return
```
kenodegard: I don't think we want this base case, wont it cause unexpected behavior if the environment file uses falsy values?

conda/env/env.py
soapy1: good catch! Previously, this was checking the contents of the name. But I don't think that is the responsibility of the code that is validating the schema, so I backed that out. Will fix this comment as well!

soapy1:
Following up on the discussion around how to move towards a more strict environment.yml implementation without breaking existing malformed environments. I've pushed a demo up to the branch https://github.com/conda/conda/compare/main...soapy1:conda:cep-0024-env-spec-plugin?expand=1. It mostly implements the suggestions in https://github.com/conda/conda/pull/15120#issuecomment-3227175472. Things to note in this demo:
* there is a new environment spec called cep-0024 that implements a strict environment.yml. This plugin is enabled by default
* the environment.yml plugin is disabled by default. Users can explicitly use it by setting the `--environment-spec` flag.
* as part of the plugin detection process, there is a hack that tries to run the environment.yml plugin detection if no other plugin was detected. This way, non-cep-24 compliant environment.yaml files will still work as they have in the past. conda will output deprecation warnings with all the validation errors from the non-compliant environment.yaml.

Then once the deprecation cycle is over, we'll need to remove:
* the old environment.yml plugin (maybe rename the cep-24 plugin to environment.yml?)
* remove the environment.yml catch all/plugin detection hack

I think I'm not a huge fan of this approach. @jezdez @kenodegard what do you think?

conda/env/env.py
```diff
@@ -2,26 +2,166 @@
...
+                    raise EnvironmentFileInvalid(str(err))
+            elif isinstance(dep, dict):
+                # dict types are also allowed. There are no requirements
+                # for the form of this entry
+                pass
+            else:
+                # All other types are invalid
+                raise EnvironmentFileInvalid(
+                    "'{dep}' is an invalid type for a 'dependency'"
+                )
```
kenodegard: at this point we already know for certain that dependencies is a list so we can skip this extra nested level:

```suggestion
    for dep in dependencies:
        if isinstance(dep, str):
            # If the dependency is a string type, it must be
            # MatchSpec compatible.
            try:
                MatchSpec(dep)
            except InvalidMatchSpec as err:
                raise EnvironmentFileInvalid(str(err))
        elif isinstance(dep, dict):
            # dict types are also allowed. There are no requirements
            # for the form of this entry
            pass
        else:
            # All other types are invalid
            raise EnvironmentFileInvalid(
                "'{dep}' is an invalid type for a 'dependency'"
            )
```

conda/env/env.py
```diff
@@ -2,26 +2,166 @@
...
+      * all entries are strings
+    """
+    field_type_validation("channels", channels, list)
+
+    if channels:
+        for channel in channels:
+            if not isinstance(channel, str):
+                raise EnvironmentFileInvalid(
+                    "`channels` key must only contain strings. Found '{channel}'"
+                )
```
kenodegard:
```suggestion
    for channel in channels:
        if not isinstance(channel, str):
            raise EnvironmentFileInvalid(
                "`channels` key must only contain strings. Found '{channel}'"
            )
```

conda/env/env.py
```diff
@@ -2,26 +2,166 @@
...
+
+    Will ensure:
+      * variables is a dict
+      * all entries are strings
+    """
+    field_type_validation("variables", variables, dict)
+    if variables:
+        for value in variables.values():
+            if not isinstance(value, str):
+                raise EnvironmentFileInvalid("`variables` values must be strings")
```
kenodegard:
```suggestion
    for value in variables.values():
        if not isinstance(value, str):
            raise EnvironmentFileInvalid("`variables` values must be strings")
```

conda/env/env.py
```diff
@@ -2,26 +2,166 @@
...
+    Will produce errors to ensure:
+      * all required fields are present
+      * all fields contain valid data
+
+    :param dict data: The contents of the environment.yaml
+    :returns errors: A list of EnvironmentFileInvalid exceptions that occurred during validation
+    """
+    errors = []
+    # Ensure all required keys are present
+    for field in set(REQUIRED_KEYS).difference(data):
```
kenodegard:
```suggestion
    for field in REQUIRED_KEYS.difference(data):
```

conda/env/env.py
```diff
@@ -2,26 +2,166 @@
...
+        errors.append(EnvironmentFileInvalid(f"Missing required field '{field}'"))
+
+    # Run validations on all the relevant fields
```
kenodegard:
```suggestion
    # Run validations on all the relevant fields, extra keys are ignored
```

kenodegard:
> Following up on the discussion around how to move towards a more strict environment.yml implementation without breaking existing malformed environments. I've pushed a demo up to the branch [`main...soapy1:conda:cep-0024-env-spec-plugin`?expand=1 (compare)](https://github.com/conda/conda/compare/main...soapy1:conda:cep-0024-env-spec-plugin?expand=1). It mostly implements the suggestions in [#15120 (comment)](https://github.com/conda/conda/pull/15120#issuecomment-3227175472). Things to note in this demo:
> 
> * there is a new environment spec called cep-0024 that implements a strict environment.yml. This plugin is enabled by default
> * the environment.yml plugin is disabled by default. Users can explicitly use it by setting the `--environment-spec` flag.
> * as part of the plugin detection process, there is a hack that tries to run the environment.yml plugin detection if no other plugin was detected. This way, non-cep-24 compliant environment.yaml files will still work as they have in the past. conda will output deprecation warnings with all the validation errors from the non-compliant environment.yaml.
> 
> Then once the deprecation cycle is over, we'll need to remove:
> 
> * the old environment.yml plugin (maybe rename the cep-24 plugin to environment.yml?)
> * remove the environment.yml catch all/plugin detection hack
> 
> I think I'm not a huge fan of this approach. @jezdez @kenodegard what do you think?

I actually like the proposed deprecation process, sure there are moving pieces but when wont there be 🤷🏼‍♂️. Will we add those changes to this PR or will that be done separately?

soapy1:
> I actually like the proposed deprecation process, sure there are moving pieces but when wont there be 🤷🏼‍♂️. Will we add those changes to this PR or will that be done separately?

I think if y'all are in favour of this proposal I can merge the demo changes into here and it'll all be in one PR. If the review gets too difficult I can break it up.

soapy1:
pre-commit.ci autofix

conda/plugins/manager.py
```diff
@@ -621,12 +621,18 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
...
         if not found:
-            # raise error if no plugins found that can read the environment file
-            raise EnvironmentSpecPluginNotDetected(
-                name=source,
-                plugin_names=hooks,
-                autodetect_disabled_plugins=autodetect_disabled_plugins,
-            )
+            # HACK: if there was no plugin found, try to catch all `environment.yml` plugin
+            try:
+                return self.get_environment_specifier_by_name(
```
soapy1, line 620: This retry at the end with the `environment.yml` plugin is kinda a hack. But, it's what allows conda to remain backwards compatible while still supporting new environment spec use cases (like conda-lock.yml files). 

This needs to stay in until the deprecation cycle is complete. And it must be removed once the non-strict environment.yml is phased out. Since the deprecation cycle is pretty long, is there a better way to indicate this to the future maintainers that will need to clean up this functionality?

conda/env/specs/cep_24.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        if not self.filename:
+            return False
+
+        # Extract the file extension (e.g., '.txt' or '' if no extension)
+        _, file_ext = os.path.splitext(self.filename)
+
+        # Check if the file has a supported extension and exists
+        if not any(spec_ext == file_ext for spec_ext in YamlFileSpec.extensions):
+            return False
```
jaimergp:
```suggestion
        if file_ext.lower() not in self.extensions:
            return False
```

Case insensitive check in this class's `extensions` rather than the parent.

conda/plugins/manager.py
jaimergp: We usually add a `# FUTURE:` comment and mention the release number somewhere.

conda/plugins/manager.py
```diff
@@ -621,12 +621,18 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
...
+                    source=source, name="environment.yml"
+                )
+            except (PluginError, CondaValueError):
+                # raise error if no plugins found that can read the environment file
+                raise EnvironmentSpecPluginNotDetected(
+                    name=source,
+                    plugin_names=hooks,
+                    autodetect_disabled_plugins=autodetect_disabled_plugins,
+                )
```
jaimergp:
```suggestion
            except (PluginError, CondaValueError) as exc:
                # raise error if no plugins found that can read the environment file
                raise EnvironmentSpecPluginNotDetected(
                    name=source,
                    plugin_names=hooks,
                    autodetect_disabled_plugins=autodetect_disabled_plugins,
                ) from exc
```

conda/env/env.py
```diff
@@ -2,26 +2,167 @@
...
+            except InvalidMatchSpec as err:
+                raise EnvironmentFileInvalid(str(err))
+        elif isinstance(dep, dict):
+            # dict types are also allowed. There are no requirements
+            # for the form of this entry
+            pass
+        else:
+            # All other types are invalid
+            raise EnvironmentFileInvalid(
+                "'{dep}' is an invalid type for a 'dependency'"
```
jaimergp:
```suggestion
                f"'{dep}' is an invalid type for a 'dependency'"
```

conda/env/env.py
```diff
@@ -2,26 +2,167 @@
...
+    """Validate the contents of the dependencies field.
+
+    Will ensure:
+      * dependencies are a list
+      * all string dependencies are MatchSpec compatible
+      * the only other type allowed is a dict
+    """
+    field_type_validation("dependencies", dependencies, list)
+
+    for dep in dependencies:
```
jaimergp: Should we accumulate all the exceptions found in the loop and raise an ExceptionGroup at the end if needed?

conda/env/specs/cep_24.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+log = getLogger(__name__)
+
+
+class Cep24YamlFileSpec(EnvironmentSpecBase):
+    _environment = None
+    extensions = {".yaml", ".yml"}
+
+    def __init__(self, filename=None, **kwargs):
```
jaimergp:
```suggestion
    def __init__(self, filename: str | None = None, **kwargs):
```

tests/env/support/empty_env.yml
soapy1: Following up here. Do we want to update the cep so that this is not a required field?

soapy1:
pre-commit.ci autofix

tests/env/test_create.py
```diff
@@ -160,12 +160,16 @@ def test_create_empty_env(
     env_name = uuid4().hex[:8]
     prefix = tmp_envs_dir / env_name
 
-    conda_cli(
-        *("env", "create"),
-        *("--name", env_name),
-        *("--file", support_file("empty_env.yml")),
-    )
-    assert prefix.exists()
+    with pytest.warns(
```
soapy1: use `pytest.deprecated_call` instead

soapy1:
pre-commit.ci autofix

conda/env/env.py
```diff
@@ -2,26 +2,169 @@
...
+                errors.append(EnvironmentFileInvalid(str(err)))
+        elif isinstance(dep, dict):
+            # dict types are also allowed. There are no requirements
+            # for the form of this entry
+            pass
+        else:
+            # All other types are invalid
+            errors.append(
+                EnvironmentFileInvalid(f"'{dep}' is an invalid type for a 'dependency'")
+            )
```
jezdez:
```suggestion
    for dependency in dependencies:
        if isinstance(dependency, str):
            # If the dependency is a string type, it must be
            # MatchSpec compatible.
            try:
                MatchSpec(dependency)
            except InvalidMatchSpec as err:
                errors.append(EnvironmentFileInvalid(str(err)))
        elif isinstance(dependency, dict):
            # dict types are also allowed. There are no requirements
            # for the form of this entry
            pass
        else:
            # All other types are invalid
            errors.append(
                EnvironmentFileInvalid(
                    f"'{dependency}' is an invalid type for a 'dependency'"
                )
            )
```

PR #15138, opened 2025-08-27
Get config from env spec files

### LINKED ISSUE #88: Determine how config from environment spec files get applied

soapy1:
### Description

This PR suggests a way of pulling in config from an environment spec file into the context object. Adding the environment config file as a new source of context allows the `Context` class to continue to have the reponsibility of collecting all sources of config and setting correct precedence. This also ensures that all pieces of conda will have access to the same config.

In this approach, the context is re-initialized after the environment spec file is parsed. The context will accept one (merged) `EnvironmentConfig` model as a config source. It's precedence is above condarc config, and under cli args/env vars.

Fixes https://github.com/conda/conda-environments-project-mgmt/issues/88
Notes and other demos: 
* https://github.com/conda/conda-environments-project-mgmt/issues/88#issuecomment-3152191454
* https://github.com/conda/conda/pull/15081

### Checklist - did you ...


conda/cli/main_env_create.py
```diff
@@ -131,6 +131,8 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     )
     spec = spec_hook.environment_spec(args.file)
     env = spec.env
+    # reset the context to pick up config from the environment spec
+    context.__init__(argparse_args=args, env_spec_config=env.config)
```
soapy1, line 135: In order to pick up config from the environment spec the context object needs to be re-initialized with the environment config. I'm not a huge fan of calling `context.__init__` here I think. Alternatives for this re-initialization include:
* in the plugin manager as part of getting the env spec plugin - this is also not a good approach. The plugin manager shouldn't have the responsibility of re-initializing the context
* somewhere earlier in the application/cli startup process - it would be better to not read the environment spec multiple times. Though, this is the kind of thing that can be cached by the plugin manager probably. So maybe this is not much of a concern.

conda/base/context.py
```diff
@@ -556,15 +557,24 @@ def __init__(
         self,
         search_path: PathsType | None = None,
         argparse_args: Namespace | None = None,
+        env_spec_config: EnvironmentConfig | None = None,
```
soapy1, line 563: I think I like this approach of accepting a single `EnvironmentConfig` here more than accepting a list of env spec files as in [pr#15081](https://github.com/conda/conda/pull/15081/files#diff-b088be378fc4d3c8240b5a24fd9d51409bb64226eb874be18a10eafadec79f78R560). 

Couple of thoughts about this:
* I don't think the context object should have any buisness with calling the plugin manager
* It would be better to not need to do the environment file reading thing multiple times. Having the interface for initializing the context object be a file forces this to happen at least twice - once when the context is initialized with the env spec information. And another time when the `Environment` model is getting created. 
* The `EnvironmentConfig` object has a handy [`merge` function](https://github.com/conda/conda/blob/main/conda/models/environment.py#L186). So with this approach, the caller is responsible for passing the final environment config to the context object. This moves the responsibility of collecting multiple environment spec files to the caller.
* Since this `EnvironmentConfig` is maybe a merge of multiple `EnvironmentConfig`s, it's not appropriate to store the path to the env spec file in the `context.raw_data` as in #15081

conda/cli/main_env_create.py
soapy1: Also considered using `reset_context` here instead. I don't think that's what we want here. `reset_context` tries to clear some plugin config.

soapy1:
pre-commit.ci autofix

conda/cli/main_env_update.py
```diff
@@ -109,6 +109,8 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     )
     spec = spec_hook.environment_spec(args.file)
     env = spec.env
+    # reset the context to pick up config from the environment spec
+    context.__init__(argparse_args=args, env_spec_config=env.config)
```
soapy1, line 113: I think since the context needs to be re-initialized after the env spec pulgins have run, the env spec plugins should not be allowed

soapy1:
@jezdez when you have the chance, can you take a look at this as an alternative to https://github.com/conda/conda/pull/15081. I've tried to add some notes about the decisions made here. I'm interested to hear your thoughts as well.

conda/cli/main_env_create.py
soapy1: Having the context be re-initialized after the env spec plugins have run raises the question, should env spec plugins be allowed to access the context object? :grimacing:  

On one hand, you could make the argument that the context object is not fully initialized when the env spec plugins run. Becuase, they are meant to produce an output that get's added to the context.

On the other hand, the config coming from env spec plugin is limited to the [`environment_context_keys`](https://github.com/conda/conda/blob/main/conda/base/context.py#L1300) that are stored in an EnvironmentConfig. So, this configuration is only changing environment specific configuration and does not effect stuf like how the cli is run. This is a little bit loosey-goosey. This can more fully resolved by seperating out application level/run time/environment level config from the context. That is way out of scope, but maybe there is a middle ground?

So, I would say, it's relatively safe for env spec plugins to be using values from the context object. But, there are certainly improvements to be made that would inspire more confidence.

conda/common/configuration.py
```diff
@@ -1500,6 +1545,17 @@ def _set_argparse_args(self, argparse_args):
...
         return self
 
+    def _set_environment_spec_data(self, config: EnvironmentConfig) -> None:
+        source = EnvironmentSpecificationRawParameter.source
+        if source in self.raw_data:
+            del self.raw_data[source]
+
+        self.raw_data[source] = (
+            EnvironmentSpecificationRawParameter.make_raw_parameters(
+                {k: v for k, v in vars(config).items() if v is not None}
```
jaimergp, line 1555: Do we need to run some checks here to make sure the config keys are valid context keys? Or is that logic above the `Configuration` object? I'm concerned about little oversights like keys being renamed or introduced with a typo and all of a sudden we have a context object with something like both `solver_options` and `solver_settings`.

conda/base/context.py
```diff
@@ -559,15 +560,24 @@ def __init__(
...
```
jaimergp, line 563: Instead of passing `EnvironmentConfig` directly, which we are only using directly as part of the `vars()` call, what if we just allow an arbitrary dictionary here, `pre_cli_config` or `config_file_overrides` or something like that? I don't think we _need_ the `EnvironmentConfig` object per se. If it's about the key validation, I think that should be handled by the `Configuration` base class already, isn't it? Like, if I pass `CONDA_ENV_DOES_NOT_EXIST=1` in my env vars, it just ignores it.

tests/base/test_context.py
```diff
@@ -838,3 +840,58 @@ def test_create_default_packages_will_warn_for_explicit_packages(
     ):
         # Ensure only valid packages are returned
         assert context.create_default_packages == (PYTHON_SPEC,)
+
```
jaimergp, line 843: Can you add a test where the env spec config has an invalid key? What's the behaviour there? Silent omission or error?

conda/common/configuration.py
soapy1: Part of the motivation for making the input type here be an `EnvironmentConfig` is to enforce some control/guardrails so that something like a typo will not result in this type of `solver_options`/`solver_settings` issue. Like, at least the input is guarenteed to be aligned with some common configration idea. 

But, I'll be the first to admit that this is still not strict enough to have total confidence that this piece of code will remain bug free. The `EnvironmentConfiguration` class is [related to the `Context` object pretty loosely](https://github.com/conda/conda/blob/main/conda/models/environment.py#L170). So, it would be very easy to introduce a bug by mistakenly renaming  [any of the `context.environment_context_keys`](https://github.com/conda/conda/blob/main/conda/base/context.py#L1300) and not making the equivalent change in the `EnvironmentConfig` class. The solution I would like to see here is something like, the context class gets broken up so that different objects represent all the different layers of config that exists. One of those layers of config should be environment specific. That should replace the `EnvironmentConfig` object, or be tightly coupled in some way. That's way out of scope for this work. I think this is an ok (but not perfect) solution to push things forward and leave space for a better solution in the future.

conda/base/context.py
soapy1: I think it is possible to make this more generic. A couple of reasons why I decided to make this an `EnvironmentConfig` instead:
* this is meant to represent a new config source. I don't think it would be good to have this mechanism be abused by other parts of the code that might want to inject config into the context.
* the configuration object has a notion of a "config source" for each type of config it supports. For example the [`ENV_VARS_SOURCE` ](https://github.com/conda/conda/blob/main/conda/common/configuration.py#L221). Making this an `EnvironmentConfig` makes the config source clear.
* introducing a specifially scoped new config source makes it clear what level of precedence the config coming from that souce will take.

I think adopting a new config source should be specifially not generic or extendable (like, adding a new config source is ok, but it should not piggy back on existing config sources). This is to ensure clarity regarding the config source and how the config gets applied.

But if these are not convincing arguments, I'm happy to change.

conda/common/configuration.py
jaimergp: Do we have a test where we ensure that the `EnvironmentConfig` keys are present in `context`? It's only one part of getting things right, but it's something.

tests/base/test_context.py
soapy1: I think this is not really a possible situation. Since the env spec config is (loosely) coupled with the fields of the context object. If the `EnvironmentConfig` had an invalid key it would be a programmer error. I added a test for this ([as suggested](https://github.com/conda/conda/pull/15138/files#r2318214149)) in tests/models/test_environment.py. Unless I am misunderstanding?

conda/common/configuration.py
soapy1: good idea! This will at least protect against programmer error if changes are made to the `EnvironmentConfig` or the `Context` field names.

soapy1:
The codspeed issues raised here should be resolved by https://github.com/conda/conda/pull/15149 and https://github.com/conda/conda/pull/15150

conda/common/configuration.py
```diff
@@ -411,6 +413,40 @@ def make_raw_parameters_from_file(cls, filepath):
...
+            return deepfreeze(self._raw_value)
+
+    def keyflag(self):
+        return None
+
+    def valueflags(self, parameter_obj):
+        return None if isinstance(parameter_obj, PrimitiveLoadedParameter) else ()
+
+    @classmethod
+    def make_raw_parameters(cls, from_map: dict[str, Any]):
```
soapy1, line 439: should this also have a `make_raw_parameters_from_environment_config(cls, config: EnvironmentConfig`?

tests/base/test_context.py
soapy1: I've also added a similar test to what you suggested to the configuration tests.

soapy1:
pre-commit.ci autofix

soapy1:
It would be nice to update the [config-precedence graphic](https://github.com/conda/conda/blob/main/docs/source/img/config-precedence.png). I wouldn't know how to edit this image to add a new config source. I could recreate it slightly differently? Does the anaconda docs team want to weight in on that? cc/ @kathatherine

kathatherine:
> It would be nice to update the [config-precedence graphic](https://github.com/conda/conda/blob/main/docs/source/img/config-precedence.png). I wouldn't know how to edit this image to add a new config source. I could recreate it slightly differently? Does the anaconda docs team want to weight in on that? cc/ @kathatherine

I have no idea where the source file is for that. You are welcome to recreate it slightly differently. It would be better to have the source file controlled by the conda org as well, to make editing it easier.

soapy1:
@kathatherine, I opted to remove the image and replace it with a list https://github.com/conda/conda/pull/15138/commits/c6b4d7583e9d2a1cc9576c2ce02a76f11d3e668e. Does that work?

kathatherine:
> @kathatherine, I opted to remove the image and replace it with a list [c6b4d75](https://github.com/conda/conda/commit/c6b4d7583e9d2a1cc9576c2ce02a76f11d3e668e). Does that work?

Oh, 100%. Probably a much better solution. 👍

soapy1:
Closing this out in favour of an alternative approach. Demo: https://github.com/conda/conda/compare/main...soapy1:conda:get-config-from-env-spec-files-env-level-condarc?expand=1

jaimergp:
@soapy1 can you open a PR for that branch? I like its simplicity but I have some comments. Thanks!

soapy1:
@jaimergp  I opened up a draft here if you want to drop some comments https://github.com/conda/conda/pull/15199. But @ryanskeith is going to pick up this work. I will also drop some comments in!

PR #12910, opened 2023-07-20
Implement 2-phase parser with greedy subparsers

kenodegard:
### Description

Splits CLI parser into a 2-phase parser. Phase 1 only parses `conda`'s options (e.g., `conda --json`, `conda --no-plugins`, etc.) allowing us to disable plugins before generating the complete parser for phase 2. A benefit of the 2-phase parser is any option parsed in phase 1 can appear anywhere on the CLI, so the following are identical:

```
conda --json doctor
conda doctor --json
```

Note that `conda doctor` has not explicitly defined `--json` as one of its arguments:

https://github.com/conda/conda/blob/b89a1470aed2383047cef05ff1e738bf0337eeec/conda/plugins/subcommands/doctor/cli.py#L22-L30

In addition to the 2-phase parser we also implement a custom `argparse._SubParserAction` that will greedily consume all unmatched arguments, this is the change that resolves the `argparse.REMAINDER` issue.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15149, opened 2025-09-02
Only init the context with pre args once

soapy1:
### Description

Initializing the context object is costly. This issue surfaced in the work getting config from environment specs. See https://github.com/conda/conda/pull/15138 and https://github.com/conda/conda/pull/15138. In these PR's codspeed indicates that adding an extra context initialization call can add [up to 2 seconds to running a conda command](https://github.com/conda/conda/pull/15081#issuecomment-3161168592).

This PR aims to reduce the number of times context (re) initialization occurs during the application startup process.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -41,16 +41,18 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
         "verbosity": pre_args.verbosity,
     }
 
-    context.__init__(argparse_args=pre_args)
-    if context.no_plugins:
+    # disable plugins if requested
+    if pre_args.no_plugins:
         context.plugin_manager.disable_external_plugins()
 
-    # reinitialize in case any of the entrypoints modified the context
+    # initialize the context with the pre_args
     context.__init__(argparse_args=pre_args)
 
+    # generate the full parser
     parser = generate_parser(add_help=True)
     args = parser.parse_args(args, override_args=override_args, namespace=pre_args)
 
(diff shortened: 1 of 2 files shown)


conda/cli/main.py
```diff
@@ -41,16 +41,18 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
         "verbosity": pre_args.verbosity,
     }
 
-    context.__init__(argparse_args=pre_args)
```
soapy1, line 44: The `no_plugins`  argument is added in the [pre_parsers](https://github.com/conda/conda/blob/main/conda/cli/conda_argparse.py#L116), so I believe it is safe to use `pre_args.no_plugins` here to determine if the plugin manager should disable external plugins.

The context object is [initalized as part of it's setup process](https://github.com/conda/conda/blob/main/conda/base/context.py#L2424). So there is no need to initialize the context in order to use the plugin manager.

So, it should be ok to remove this first context initialization with pre_args. We'll still initialize the context with pre_args after disabling external plugins. And then initialize the context with the full arg parse arguments after generating the parser. This [should still respoect the 2-phase parser ](https://github.com/conda/conda/pull/12910) setup.

kenodegard: changing this means `no_plugins: true` in condarc has no effect

soapy1: ah, you're right! This is not an ok solution. Closing this pr. Thanks :100:

PR #15150, opened 2025-09-02
Cache reading condarc files

soapy1:
### Description

Initializing the context object is costly. This issue surfaced in the work getting config from environment specs. See https://github.com/conda/conda/pull/15138 and https://github.com/conda/conda/pull/15138. In these PR's codspeed indicates that adding an extra context initialization call can add https://github.com/conda/conda/pull/15081#issuecomment-3161168592.

Looking at the codspeed flamegraph, it looks like this is due to overhead with reading files. Every time a context object is re-initialized it must read all the config files. Since config files are not meant to change while conda is running a command, we can speed this up by caching the results of reading the config file.

### Checklist - did you ...

### DIFF
--- a/conda/common/configuration.py
+++ b/conda/common/configuration.py
@@ -398,7 +398,7 @@ class YamlRawParameter(RawParameter):
 
         This function will cache the result for each filepath. In order to re-read the same
         file path with updated content, be sure to clear this cache.
-        
+
         For example::
 
             YamlRawParameter.cache_clear()


conda/common/configuration.py
```diff
@@ -390,6 +390,7 @@ def make_raw_parameters(cls, source, from_map):
         return EMPTY_MAP
 
     @classmethod
+    @cache
```
soapy1, line 394: This assumes that the contents of config files do not change during the execution of one conda command. I think this is a safe assumption to make? Though I can see that being wrong. For example, is conda meant to be able to run some actions after a `conda config --file` command updates a given command file?

Using this approach, if conda needs to re-initialize the context due to a config file being updated, then the calling code must clear this cache. For example `YamlRawParameter.make_raw_parameters_from_file.cache_clear()`.

tests/cli/test_main_remove.py
```diff
@@ -114,16 +114,13 @@ def test_remove_nonexistent_env(conda_cli: CondaCLIFixture):
...
 ):
     """Check that removing the default_activation_env raises an exception."""
     with tmp_env() as prefix:
-        conda_cli(
-            "config",
-            "--set",
-            "default_activation_env",
-            prefix,
-        )
+        monkeypatch.setenv("CONDA_DEFAULT_ACTIVATION_ENV", prefix)
```
soapy1, line 121: I think this approach of setting config for tests is more consistent with rest of the test suite

tests/cli/test_config.py
```diff
@@ -48,6 +52,7 @@ def make_temp_condarc(value=None):
         if value:
             with open(temp_path, "w") as f:
                 f.write(value)
+        YamlRawParameter.make_raw_parameters_from_file.cache_clear()
```
soapy1: This is required since the test updates a config file and resets the context. In order to pick up the new changes to the file we must clear the cache.

jezdez:
Woah, nice find, up to 20% faster in the performance tests above!

conda/common/configuration.py
jezdez: I think this is probably really tricky to implement, e.g. post command plugin hooks are also possible to hook into the config subcommand. Is there a code path for file config file modification we can use to invalidate the cache?

jezdez: Also, is it safe to use `functools.cache` with class methods?

tests/cli/test_config.py
jezdez: This is a bit of a mouthful. Let's add a cache invalidation function to `conda.common.configuration.YamlRawParameter` as an interface like:

```python
@classmethod
def cache_clear(cls) -> None:
    cls.make_raw_parameters_from_file.cache_clear()
```

```suggestion
        YamlRawParameter.cache_clear()
```

conda/common/configuration.py
jezdez: Let's call the cache invalidation in `reset_context`, so we can skip the manual cache invalidation

soapy1: > Is there a code path for file config file modification we can use to invalidate the cache?

Lets try the invalidation in the `reset_context` approach. I think this added functionality in `reset_context` is still in line with the purpose of that function.

> Also, is it safe to use functools.cache with class methods?

I believe it is in this case. No state is meant to be changing inside this function and there is a 1-to-1 relationship between input and ouptut. From a memory perspective, `YamlRawParameter` is not really instantiated as a class as it's primary function. Instead, it's primary function is to use this classmethod to create a dict representation of a given config file. So, there shouldn't be an issue with instances of `YamlRawParmeter` not getting cleaned up and exploding the memory. [(Link to docs)](https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls)

tests/cli/test_config.py
soapy1: Good idea!

soapy1:
pre-commit.ci autofix

kenodegard:
I think we should solve this differently at a lower level

Instead of caching these fancy `YamlRawParameter` objects let's cache the I/O + parsing to yaml

We also recently cleaned up the json parsers:

https://github.com/conda/conda/blob/8e9eb19cb0bf2acb551b4ebbb091affff12af362/conda/common/serialize/json.py

I had every intention of refactoring our yaml functions to mirror that so I think we should do that first and then implement caching for both json and yaml

kenodegard:
marking as a draft while we work on the yaml refactor

soapy1:
Re-opening this PR since the yaml refactor has been completed. 

I tried out setting up caching at the file reading level. But, after trying it out and running into some issues with the jlap tests, and codspeed tests indicating a decrease in performance, I think this PR suggests a better solution for caching the contents of condarc files. More thoughts on this in [that pr](https://github.com/conda/conda/pull/15593#issuecomment-3752514841).

@kenodegard did you have a different idea for a yaml/json parsing caching solution?

soapy1:
> The only concern is potential uses of this functionality that assumes non-cached behavior; maybe it could be called out to work around it in the docs or the docstring?

I added a note in the `make_raw_parameters_from_file` docstring :rocket:

conda/common/configuration.py
```diff
@@ -391,7 +391,18 @@ def make_raw_parameters(cls, source, from_map):
...
     def make_raw_parameters_from_file(cls, filepath):
+        """
+        Read the provided file path and convert the contents into configuration parameters.
+
+        This function will cache the result for each filepath. In order to re-read the same
+        file path with updated content, be sure to clear this cache. For example
+
+        ```
+        YamlRawParameter.cache_clear()
+        ```
```
jezdez:
```suggestion
        file path with updated content, be sure to clear this cache.
        
        For example::

            YamlRawParameter.cache_clear()
```

soapy1:
pre-commit.ci autofix

PR #15162, opened 2025-09-04
Use `default=NULL` for `--override-frozen`

jaimergp:
### Description

@marcoesters reported this while debugging the new protected base installers. Turns out the `CONDA_PROTECT_FROZEN_ENVS` env var had no effect on the config because `--override-frozen` had `action="store_false"`, whose default value is `True`. Since the CLI layer has precedence over the env vars to populate the context, this implicit default was overriding the env var config, _even_ if there was no explicit CLI flag passed!

The fix, as seen in other bits of the CLI config, is to use an explicit `default=NULL`, which is special cased in `configuration.py`. This might be hitting other booleans `store_{true,false}` in the CLI config. Took a while to understand what was going on 😬 

### Checklist - did you ...

### DIFF
--- a/tests/cli/test_cli_install.py
+++ b/tests/cli/test_cli_install.py
@@ -130,7 +130,7 @@ def test_emscripten_forge(
         assert package_is_installed(prefix, "pyjs")
 
 
-def test_frozen_env_cep22(tmp_env, conda_cli):
+def test_frozen_env_cep22(tmp_env, conda_cli, monkeypatch):
     with tmp_env("ca-certificates") as prefix:
         prefix_data = PrefixData(prefix)
         prefix_data._frozen_file.touch()
@@ -145,7 +145,7 @@ def test_frozen_env_cep22(tmp_env, conda_cli):
             "update", "-p", prefix, "ca-certificates", raises=EnvironmentIsFrozenError
         )
 
-        # Bypass protection
+        # Bypass protection with CLI flag
         conda_cli(
             "install",
             "-p",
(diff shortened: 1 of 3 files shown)


tests/cli/test_cli_install.py
```diff
@@ -130,7 +130,7 @@ def test_emscripten_forge(
         assert package_is_installed(prefix, "pyjs")
 
 
-def test_frozen_env_cep22(tmp_env, conda_cli):
+def test_frozen_env_cep22(tmp_env, conda_cli, monkeypatch):
```
ForgottenProgramme, line 133: why do you have a CEP number in the test name?

jaimergp: That's the CEP that introduced the `conda-meta/frozen` file.

tests/cli/test_cli_install.py
```diff
@@ -174,6 +174,34 @@ def test_frozen_env_cep22(tmp_env, conda_cli):
         )
         assert rc == 0
 
+        # Bypass protection with env var
+        with monkeypatch.context() as monkeyctx:
+            monkeyctx.setenv("CONDA_PROTECT_FROZEN_ENVS", "0")
```
ForgottenProgramme, line 179: Should we also test for `CONDA_PROTECT_FROZEN_ENVS = 1`?

tests/cli/test_cli_install.py
ForgottenProgramme: yes, but do we need to have that here in the test name? 🤷🏼‍♀️

tests/cli/test_cli_install.py
jaimergp: If this case works, the other one should work too if we trust the configuration system. The regression here was the the env var was being overridden with the CLI silently, which this test covers. You can reproduce the original issue if you remove the `default=NULL` line from the diff.

tests/cli/test_cli_install.py
jaimergp: When I introduced this test (not in this PR), I thought "frozen env" might not be informative enough.

ForgottenProgramme: I think now there is enough knowledge/awareness about the idea of "frozen envs" and we probably don't need the `cep22` suffix.

jezdez (approved):

PR #15165, opened 2025-09-04
Add configuration support for multiple platforms

### LINKED ISSUE #15158: New command line configuration option in condarc to specify which platforms lock files should be generated
### What is the idea?
Implement a mechanism to specify target platforms when generating lock files, allowing users to create platform-specific lock files beyond their current platform.

### Why is this needed?
Currently, users can only generate lock files for their current platform. This creates challenges for:

- Cross-platform development teams needing consistent environments
- CI/CD pipelines running on different platforms
- Projects that need to ensure compatibility across multiple operating systems
- Organizations deploying applications in heterogeneous computing environments

### What should happen?

1. Add a new `--platform` option to the `conda export` command that accepts one or more platform identifiers
1. Implement a configuration option in condarc (either `extra_platforms` or `export_platforms`)
3. When specified, conda should generate lock files for all requested platforms, not just the current one

### Additional Context
The configuration option (either `extra_platforms` or `export_platforms`) would allow users to set default export platforms in their condarc file, which would then be used whenever they run conda export without explicitly specifying platforms. This provides a convenient way to standardize lock file generation across a team or project.

### Unanswered Questions

How should we handle formats that only support a single platform when multiple platforms are specified?

- **Option A:** Generate multiple output files with platform-specific naming (e.g., environment-linux-64.yml, environment-osx-64.yml)
- **Option B:** Raise an exception and require users to specify a format that supports multiple platforms
- **Option C:** Generate only for the current platform and warn about the others

Do we also need to support an `--override-platforms` argument mirroring `--channel`/`--override-channels`?

### Proposed Implementation

1. Extend the conda plugin system to support platform-specific lock file generation
2. Modify the `conda export` command to accept a `--platform` flag, e.g., `--platform linux-64 --platform osx-64 --platform win-64`
3. Update the condarc schema to include either an `extra_platforms` or `export_platforms` option that accepts a list of platforms
4. When generating lock files, conda should use `conda.base.context.context.extra_platforms`/`conda.base.context.context.export_platforms` to extrapolate an environment per platform (see https://github.com/conda/conda/issues/15157)

ryanskeith:
### Description

This PR implements the configuration option and command line flag for the conda export command for multiplatform export outputs.

Closes #15158

Changes made:
- Added export_platforms configuration parameter to conda/base/context.py with alias extra_platforms
- Added --platform command line argument to conda export command (supports multiple platforms)
- Updated documentation with platform-specific export examples and condarc configuration
- Added comprehensive tests for both single and multiple platform scenarios

Key features:
- Users can specify target platforms via --platform linux-64 --platform osx-64
- Configuration can be set in condarc: export_platforms: [linux-64, osx-64, win-64]
- Falls back to current platform if no platforms specified
- This provides the foundation for future multi-platform export functionality while maintaining backward compatibility.
  - NOTE: Right now, the code only accepts the first platform for exporting
  
Checklist - did you ...
- [ ] Add a file to the news directory 1 for the next release's release notes?

### DIFF
--- a/conda/cli/main_export.py
+++ b/conda/cli/main_export.py
@@ -68,8 +68,13 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         "--platform",
         "--subdir",
         action="append",
-        dest="platforms",
+        dest="export_platforms",
         help="Target platform(s)/subdir(s) for export (e.g., linux-64, osx-64, win-64)",
+    )
+    p.add_argument(
+        "--override-platforms",
+        action="store_true",
+        help="Override the platforms specified in the condarc",
     )
     add_parser_prefix(p)
 
@@ -134,17 +139,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from .common import stdout_json
 
     # Determine target platforms for export
-    target_platforms = []
-    if args.platforms:
-        # Use platforms specified on command line
-        target_platforms = args.platforms
-    elif context.export_platforms:
-        # Use platforms from condarc configuration
-        target_platforms = list(context.export_platforms)
-
-    # If no platforms specified, use current platform only
-    if not target_platforms:
-        target_platforms = [context.subdir]
+    target_platforms = context.export_platforms
 
     # TODO: Check if platform targets are valid
 
@@ -190,7 +185,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     env = Environment.from_prefix(
         prefix=prefix,
         name=env_name(prefix),
-        platform=context.subdir,
+        platform=target_platforms[0],
         from_history=args.from_history,
         no_builds=args.no_builds,
         ignore_channels=args.ignore_channels,
--- a/conda/cli/main_export.py
+++ b/conda/cli/main_export.py
@@ -66,8 +66,9 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     # for the current environment.  We may want to change the helper.
     p.add_argument(
         "--platform",
+        "--subdir",
         action="append",
-        help="Target platform(s) for export (e.g., linux-64, osx-64, win-64)",
+        help="Target platform(s)/subdir(s) for export (e.g., linux-64, osx-64, win-64)",
     )
     add_parser_prefix(p)
 
--- a/conda/cli/main_export.py
+++ b/conda/cli/main_export.py
@@ -57,6 +57,18 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         action="store_true",
         help="Do not include .condarc channels",
     )
+    # NOTE: This is a different platform option from the one in helpers.py
+    # This is because we want to:
+    # - Allow users to specify multiple platforms for export
+    # - Change the help so that it's clearer that this is for export
+    #
+    #  The add_parser_platform in helpers.py is used to specify a single platform/subdir
+    # for the current environment.  We may want to change the helper.
+    p.add_argument(
+        "--platform",
+        action="append",
+        help="Target platform(s) for export (e.g., linux-64, osx-64, win-64)",
+    )
     add_parser_prefix(p)
 
     p.add_argument(
@@ -119,6 +131,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import env_name
     from .common import stdout_json
 
+    # Determine target platforms for export
+    target_platforms = []
+    if args.platform:
+        # Use platforms specified on command line
+        target_platforms = args.platform
+    elif context.export_platforms:
+        # Use platforms from condarc configuration
+        target_platforms = list(context.export_platforms)
+
+    # If no platforms specified, use current platform only
+    if not target_platforms:
+        target_platforms = [context.subdir]
+
+    # TODO: Check if platform targets are valid
+
     # Early format validation - fail fast if format is unsupported
     target_format = args.format
     environment_exporter = None
@@ -152,10 +179,14 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     prefix = context.target_prefix
 
     # Create models.Environment directly
+    # TODO: In the future, this will support generating exports for multiple platforms
+    # For now, we use the first target platform
+    target_platform = target_platforms[0] if target_platforms else context.subdir
+
     env = Environment.from_prefix(
         prefix=prefix,
         name=env_name(prefix),
-        platform=context.subdir,
+        platform=target_platform,
         from_history=args.from_history,
         no_builds=args.no_builds,
         ignore_channels=args.ignore_channels,
--- a/conda/cli/main_export.py
+++ b/conda/cli/main_export.py
@@ -138,9 +138,6 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import env_name
     from .common import stdout_json
 
-    # Determine target platforms for export
-    target_platforms = context.export_platforms
-
     # TODO: Check if platform targets are valid
 
     # Early format validation - fail fast if format is unsupported
@@ -180,12 +177,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     #       we need to specify the source platform and then export to
     #       the target platform?  If so, is this done in the
     #       environment_exporter?
-    # target_platform = target_platforms[0] if target_platforms else context.subdir
+    # target_platforms = context.export_platforms
 
     env = Environment.from_prefix(
         prefix=prefix,
         name=env_name(prefix),
-        platform=target_platforms[0],
+        platform=context.subdir,
         from_history=args.from_history,
         no_builds=args.no_builds,
         ignore_channels=args.ignore_channels,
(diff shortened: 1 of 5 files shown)


conda/cli/main_export.py
```diff
@@ -57,6 +57,18 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
     )
+    # NOTE: This is a different platform option from the one in helpers.py
+    # This is because we want to:
+    # - Allow users to specify multiple platforms for export
+    # - Change the help so that it's clearer that this is for export
+    #
+    #  The add_parser_platform in helpers.py is used to specify a single platform/subdir
+    # for the current environment.  We may want to change the helper.
+    p.add_argument(
+        "--platform",
```
kenodegard, line 68: since we also use the "subdir" terminology in conda should we support both flags?

conda/cli/main_export.py
```diff
@@ -57,6 +57,18 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+        action="append",
+        help="Target platform(s) for export (e.g., linux-64, osx-64, win-64)",
```
kenodegard: Should we make this a `choices` based on `context.known_subdirs`?

conda/cli/main_export.py
```diff
@@ -119,6 +131,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
     from .common import stdout_json
 
+    # Determine target platforms for export
+    target_platforms = []
+    if args.platform:
+        # Use platforms specified on command line
+        target_platforms = args.platform
+    elif context.export_platforms:
+        # Use platforms from condarc configuration
+        target_platforms = list(context.export_platforms)
```
kenodegard: Should we follow this pattern (CLI overrides the context), or do we follow the pattern established by `--channel` (CLI extends the context, unless `--override-channels` is provided)?

Or do we introduce a more explicit syntax similar to `--ignore`/`--extend-ignore` in flake8? `--extend-platform` vs `--platform`?

cc @jezdez @jaimergp

tests/cli/test_main_export.py
```diff
@@ -811,3 +811,43 @@ def test_export_explicit_format_validation_errors(
...
+
+
+def test_export_platform_argument(conda_cli: CondaCLIFixture) -> None:
+    """Test export with platform argument."""
+    name = uuid.uuid4().hex
+    stdout, stderr, code = conda_cli(
+        "export",
+        f"--name={name}",
+        "--platform",
+        "linux-64",
```
kenodegard:
```suggestion
        "--platform=linux-64",
```

tests/cli/test_main_export.py
```diff
@@ -811,3 +811,43 @@ def test_export_explicit_format_validation_errors(
...
+def test_export_multiple_platforms(conda_cli: CondaCLIFixture) -> None:
+    """Test export with multiple platform arguments."""
+    name = uuid.uuid4().hex
+    stdout, stderr, code = conda_cli(
+        "export",
+        f"--name={name}",
+        "--platform",
+        "linux-64",
+        "--platform",
+        "osx-64",
```
kenodegard:
```suggestion
        "--platform=linux-64",
        "--platform=osx-64",
```

conda/cli/main_export.py
ryanskeith: That's a good idea.  Do you foresee any trouble with supporting all known choices?  Perhaps we might want to down select if there are things we think will break.

conda/cli/main_export.py
ryanskeith: Sure

conda/cli/main_export.py
ryanskeith: To me, this feels most natural.  Admittedly, I don't use export a lot.

conda/cli/main_export.py
kenodegard: The potentially odd one would be `noarch`

conda/cli/main_export.py
soapy1: Is this meant to provide the ability to override the target platform with cli flag? 

If there is not specific overriding functionality, then conda should probably just let the [config precedence logic](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#precedence) take take of determining the right value for this setting. On startup, conda will slurp up all the config from every source and the context object will ensure that the precedence described in the docs is preserved.

jezdez: @kenodegard I'm in favor of following established conda config patterns since constructing lists via CLI options isn't a great user experience (append versus insert etc).

conda/cli/main_export.py
```diff
@@ -152,6 +180,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     prefix = context.target_prefix
 
     # Create models.Environment directly
+    # TODO: Figure out how to handle source and target platforms.  Do we
+    #       we need to specify the source platform and then export to
+    #       the target platform?  If so, is this done in the
+    #       environment_exporter?
+    # target_platform = target_platforms[0] if target_platforms else context.subdir
```
jezdez: What do you plan to do with this TODO?

conda/cli/main_export.py
```diff
@@ -119,6 +132,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        target_platforms = args.platform
+    elif context.export_platforms:
+        # Use platforms from condarc configuration
+        target_platforms = list(context.export_platforms)
+
+    # If no platforms specified, use current platform only
+    if not target_platforms:
+        target_platforms = [context.subdir]
+
+    # TODO: Check if platform targets are valid
```
jezdez, line 141: Can you elaborate what you mean with this?

conda/cli/main_export.py
```diff
@@ -57,6 +57,19 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+    # - Allow users to specify multiple platforms for export
+    # - Change the help so that it's clearer that this is for export
+    #
+    #  The add_parser_platform in helpers.py is used to specify a single platform/subdir
+    # for the current environment.  We may want to change the helper.
+    p.add_argument(
+        "--platform",
+        "--subdir",
+        action="append",
+        help="Target platform(s)/subdir(s) for export (e.g., linux-64, osx-64, win-64)",
```
jezdez:
```suggestion
        action="append",
        dest="platforms",
        help="Target platform(s)/subdir(s) for export (e.g., linux-64, osx-64, win-64)",
```

That'll make it easier to understand when we refer to the parsed values being a list, rather than one item, see below

conda/cli/main_export.py
```diff
@@ -119,6 +132,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..base.context import env_name
     from .common import stdout_json
 
+    # Determine target platforms for export
+    target_platforms = []
+    if args.platform:
+        # Use platforms specified on command line
+        target_platforms = args.platform
```
jezdez:
```suggestion
    if args.platforms:
        # Use platforms specified on command line
        target_platforms = args.platforms
```

conda/cli/main_export.py
jezdez: @ryanskeith To clarify, we should follow Ken's proposal here that the CLI doesn't override the configured platforms unless an override flag is provided. My assumption is that most users have a platform they want to export for (their currently running platform, that they are on), and if they would like to provide additional ad-hoc platforms via the CLI, they'll need to specify if that platform shouldn't be included automatically.

conda/cli/main_export.py
ryanskeith: This is setting the default for the next actions.  If no CLI or config options have been set, set the target_platforms to the users platform.  I think this is the most widely used case.

conda/cli/main_export.py
ryanskeith: This is not the job if this PR.  This is a note to the implementer about what concerns may need to be addressed.

conda/cli/main_export.py
```diff
@@ -152,10 +176,16 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    #       we need to specify the source platform and then export to
+    #       the target platform?  If so, is this done in the
+    #       environment_exporter?
+    # target_platform = target_platforms[0] if target_platforms else context.subdir
+
     env = Environment.from_prefix(
         prefix=prefix,
         name=env_name(prefix),
-        platform=context.subdir,
+        platform=target_platforms[0],
```
ForgottenProgramme: Hey, @ryanskeith ! 
Why are we doing `target_platforms[0]` here?

kenodegard: This PR only adds the `--platform` CLI argument and `export_platforms` context variable, the export command doesn't support multiple platforms yet

ForgottenProgramme (changes requested):
@ryanskeith, I think the tests are failing because the description map has not been updated to include `export_platforms`. 

See https://github.com/conda/conda/blob/main/conda/base/context.py#L1438

conda/cli/main_export.py
jaimergp: You can use `conda.base.constants.PLATFORMS` for that, which excludes `noarch`.

ForgottenProgramme (approved):
LGTM!

travishathaway (approved):
LGTM 🚀 🦖

conda/cli/main_export.py
jezdez: Oh, I meant, elaborate in the comment.

PR #15177, opened 2025-09-09
Unlink target file, not just "cache_path_json" if exists

dholth:
### Description

Bugfix for https://github.com/conda/conda/blob/main/news/15083-bytes-cache-for-shards

### Checklist - did you ...

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -630,7 +630,7 @@ class RepodataCache:
             try:
                 temp_path.rename(target)
             except FileExistsError:  # Windows
-                self.cache_path_json.unlink()
+                target.unlink()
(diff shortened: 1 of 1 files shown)


conda/gateways/repodata/__init__.py
```diff
@@ -630,7 +630,7 @@ def replace(self, temp_path: Path, target=None):
             try:
                 temp_path.rename(target)
             except FileExistsError:  # Windows
-                self.cache_path_json.unlink()
+                target.unlink()
```
dholth, line 633: In L631 we try to move `temp_path` onto `target`, not `cache_path_json`; so we must unlink `target` if that fails.
An older version of this function only used `cache_path_json` and didn't have `target`.

travishathaway (approved):

PR #15183, opened 2025-09-09
Extrapolate environment for a different platform

### LINKED ISSUE #15157: Add Environment Extrapolation Function for Cross-Platform Compatibility
### Checklist

### What is the idea?

Create a new function that can take an existing `conda.models.environment.Environment` instance and extrapolate/transform it for a different target platform while preserving compatible specifications.

### Why is this needed?

This functionality is necessary for supporting multiplatform lockfiles in conda environments.

### What should happen?

1. Given a `conda.models.environments.Environment` object, extrapolate a set of packages to solve for on a different platform. Derive packages using a combination of techniques:
    - topographical sorting of packages
    - from history file
2. We need to limit the total number of solves, ideally only one per platform

### Additional Context

_No response_

kenodegard:
### Description

In order to export a multiplatform file format (e.g., lockfiles) we need to support extrapolating an environment for different platforms. We add an extrapolation function that uses the conda history + PyPI package detection to solve for a new environment.

Closes https://github.com/conda/conda/issues/15157

### Checklist - did you ...


conda/models/environment.py
```diff
@@ -403,10 +403,7 @@ def from_prefix(
 
         # Handle --from-history case
         if from_history:
-            history = History(prefix)
-            spec_map = history.get_requested_specs_map()
-            # Get MatchSpec objects from history; they'll be serialized to bracket format later
-            requested_packages = list(spec_map.values())
+            requested_packages = cls.from_history(prefix)
```
jezdez, line 407: This seems like it should be an instance method?

conda/models/environment.py
```diff
@@ -546,3 +543,40 @@ def from_cli(
             explicit_packages=explicit_packages,
             config=EnvironmentConfig.from_context(),
         )
+
+    def from_history(self: Self | str) -> list[MatchSpec]:  # type: ignore[misc]
+        prefix = self.prefix if isinstance(self, Environment) else str(self)
```
jezdez: 👀

kenodegard: yeah while this works, it is...surprising...so probably best to make this a staticmethod

jezdez (dismissed):

jezdez (approved):

PR #15197, opened 2025-09-11
Create a parent class of `Environment` representing an environment not on disk

kenodegard:
### Description

While working on https://github.com/conda/conda/pull/15183 we found that not every environment needs to define a prefix.

A solution proposed here is to create a parent class of Environment to represent a non-existent environment.

Possible name choices:
- [ ] `VirtualEnvironment`
- [ ] `HypotheticalEnvironment`
- [ ] others?
- [x] `AbstractEnvironment`

### Checklist - did you ...


jezdez:
"Virtual environment" is a very loaded term in Python, let's not use that, please?

How about `Environment`/`PrefixEnvironment` instead?

conda/models/environment.py
```diff
@@ -208,16 +209,13 @@ def merge(cls, *configs: EnvironmentConfig) -> EnvironmentConfig:
 
 
 @dataclass
-class Environment:
+class VirtualEnvironment:
```
soapy1: What type of things are meant to be using this `VirtualEnvironment`? And which things should be opting for the `Environment`?

Like, when you get the (1) output from an environment spec plugin, or (2) create an environment from a prefix, or (3) create an environment from cli args?

Why not make the `prefix` optional in the existing `Environment` model, instead of adding this new class?

kenodegard:
> "Virtual environment" is a very loaded term in Python, let's not use that, please?
> 
> How about `Environment`/`PrefixEnvironment` instead?

I agree. I was just trying to avoid renaming the current `Environment` object where `prefix` is already required.

Is there some other name we can use that means the environment doesn't exist yet? `BlueprintEnvironment`? `AbstractEnvironment`?

conda/models/environment.py
kenodegard: When describing an environment on disk we'd use `Environment`.

When talking about an environment that doesn't exist yet it would be this no-prefix environment. We could even create a "write" method in this no-prefix environment that writes the env to disk and returns a new `Environment`.

> (1) output from an environment spec plugin

yes, plugins would return the no-prefix environment, conda would then write the env to disk and produce an `Environment` with the prefix set

> (2) create an environment from a prefix

since the model describes something on disk it would be the `Environment` model

> (3) create an environment from cli args

depends on whether its a create or install/update command

e.g.: the new extrapolate method in https://github.com/conda/conda/pull/15183 would use the no-prefix environment model

soapy1: I think this is a neat idea. It would be good if you could merge a bunch of `Virtual/Abstract/Proto-Environments` together and be able to generate an `Environment`. And then you can use that single environment to do a solve against and install whatever needs to be installed.

One thing that is not completely clear to me with this approach is how the `prefix` specified in an environment spec will get included in the configuration that conda will use.

jezdez:
> Is there some other name we can use that means the environment doesn't exist yet? BlueprintEnvironment? AbstractEnvironment?

Abstract is good IMO

jezdez:
@kenodegard What's your take on #15281 compared to this?

kenodegard:
Closing in favor of the simpler #15281

PR #15199, opened 2025-09-11
Use fresh_context and EnvironmentConfig to create new environment

soapy1:
### Description

      
Fixes https://github.com/conda/conda-planning/issues/88

### Checklist - did you ...


conda/cli/main_env_create.py
```diff
@@ -109,6 +111,16 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     return p
 
 
+def write_env_condarc(prefix: str, config: EnvironmentConfig):
+    from .main_config import _write_rc
+    path = Path(prefix, ".condarc")
```
soapy1: is there a `.condarc` constant somewhere that can be used?

conda/cli/main_env_create.py
```diff
@@ -109,6 +111,16 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
     return p
 
 
+def write_env_condarc(prefix: str, config: EnvironmentConfig):
+    from .main_config import _write_rc
```
soapy1: would be nice to not import this `_` function. Maybe it would be good to extract a shared utility or something?

conda/cli/main_env_create.py
```diff
@@ -156,6 +168,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     prefix_data.validate_path()
     prefix_data.validate_name()
 
+    # HACK: write condarc to prefix
+    mkdir_p(prefix)
```
soapy1: I'm not sure if it's ok to be making this prefix like this. I think usually there is some prefix path action that deals with this. So, I'm not sure if this is ok?

conda/cli/main_env_create.py
ryanskeith: There is but it would look kinda bad as it stores both `.condarc` and `condarc`.

See: https://github.com/conda/conda/blob/80fa8d32dc706f6ed5900e68104674862030b21c/conda/common/configuration.py#L1336

conda/cli/main_env_create.py
ryanskeith: I looked around and I think I am ok with this for now.  Nothing does exactly what we want.

conda/cli/main_env_create.py
```diff
@@ -141,70 +154,97 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    # Convert all environment configuration to environment variables for fresh_context
+    # NOTE: Only channels and environment variables are currently supported in
+    # environment.yaml files.
+    env_vars = {}
+    # Add environment variables from YAML (convert all values to strings)
+    if env.variables:
+        env_vars.update({k: str(v) for k, v in env.variables.items()})
+
+    if env.config.channels:
+        env_vars["CONDA_CHANNELS"] = ",".join(env.config.channels)
```
soapy1: This needs to work for all the configuration options in the `EnvironmentConfig` object. And should be extensible so that developers can add more configuration options in the future.

For example, the environment spec api supports specs that can set any of the configuration  defined in `EnvironmentConfig`. If a spec decides it needs that config to exist, then conda should respect that.

conda/cli/main_env_create.py
```diff
@@ -141,70 +154,97 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-    # common.ensure_override_channels_requires_channel(args)
-    # channel_urls = args.channel or ()
-
-    result = {"conda": None, "pip": None}
+    # Convert all environment configuration to environment variables for fresh_context
+    # NOTE: Only channels and environment variables are currently supported in
+    # environment.yaml files.
+    env_vars = {}
+    # Add environment variables from YAML (convert all values to strings)
+    if env.variables:
```
soapy1: These `env.variables` are not the same as env vars for the purpose of configuration. They are just environment variables that the user wants setup in their environment.

conda/cli/main_env_create.py
```diff
@@ -141,70 +154,97 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        env_vars["CONDA_CHANNELS"] = ",".join(env.config.channels)
+
+    # TODO: Add other config options when they're supported in environment.yaml
+    # if env.config.channel_priority:
+    #     env_vars['CONDA_CHANNEL_PRIORITY'] = str(env.config.channel_priority)
+    # if env.config.solver:
+    #     env_vars['CONDA_SOLVER'] = env.config.solver
+
+    # Use fresh_context to load all environment configuration without writing to disk
+    with fresh_context(argparse_args=args, env=env_vars):
```
soapy1, line 190: Thats a neat idea! Taking this approach, we don't need to change conda's behaviour when it comes to writing an environment level condarc when an environment is created. @jezdez @kenodegard what do you think?

tests/cli/test_main_config.py
```diff
@@ -37,6 +37,106 @@
     )
 
 
+@pytest.mark.integration
+def test_env_create_with_channels_from_yaml(
```
soapy1: I think these tests would fit better in `tests/cli/test_env.py`

soapy1 (commented):
Interesting approach. I like the idea of using env vars instead of writing files to get information into the context. Just a few questions.

conda/cli/main_env_create.py
soapy1: I think this can be written up like https://github.com/conda/conda/pull/15232

conda/cli/main_env_create.py
ryanskeith: Right now the environment.yaml only supports channels and environment variables.  So only those can be implemented at this point in time.  As we change what is allowed in an environment.yaml, we can add them on.

conda/cli/main_env_create.py
ryanskeith: So, we don't want to use them as we are setting up the environment?

tests/cli/test_main_config.py
ryanskeith: I struggled with where to put them.  I can move them there.

conda/cli/main_env_create.py
soapy1: ok so, after chatting with @ryanskeith, I think we do want to include these here. For example, its pretty reasonable for someone to have an environment.yml

```
name: myenv
channels:
  - conda-forge
dependencies:
  - python
variables:
  VARIABLE: value
  CONDA_SOLVER: libmamba
```

On the main branch of conda this environment.yml environment configuration variables (`CONDA_SOLVER`) will not get applied on initial environment creation. But, this configuration will get persisted to the created environment.

So, this line has a bit of change in behaviour. Now, the environment variables defined here will get included into the context on environment creation.

ryanskeith:
It was decided to leave out the .condarc writing and use of environment variables for now.

ryanskeith (dismissed):
Looks good to me.

conda/cli/main_env_create.py
dholth: [I see the definition.](https://github.com/conda/conda/blob/b014c30eac546799ab751d316a063ef3f22dc5cd/conda/base/context.py#L2057) It creates a context with a new environment, and then creates another context with the old environment variables.
Is that different than using the context stack, preserving the old environment, and popping the new one when we're done here? Do the stacked contexts combine?

conda/cli/main_env_create.py
```diff
@@ -142,70 +143,91 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-    # common.ensure_override_channels_requires_channel(args)
-    # channel_urls = args.channel or ()
+    # Convert all environment configuration to environment variables for fresh_context
+    # NOTE: Only channels and environment variables are currently supported in
+    # environment.yaml files.
+    env_vars = {}
 
-    result = {"conda": None, "pip": None}
+    if env.config.channels:
+        env_vars["CONDA_CHANNELS"] = ",".join(env.config.channels)
```
soapy1: I was doing a bit more QA and found that channels added using env vars get overriden. I think this needs to be updated to respect existing environment variables. 

For example, consider the environment.yml
```
channels:
  - bioconda
dependencies:
  - python
```

running the command `$ CONDA_CHANNELS=abc conda env create --file tests/env/support/simple.yml`, I would expect the channels conda uses to be `abc`, `bioconda`, and my configured channels  `defaults` and `conda-forge`. Insead the result I get is:

```
$ CONDA_CHANNELS=abc conda env create --file tests/env/support/simple.yml

Channels:
 - bioconda
 - defaults
 - conda-forge
Platform: linux-64
. . .
```

It's missing the `abc` channel sepecified as an env var. I _think_ it's ok to address this by appending to the existing list from the `CONDA_CHANNELS` env var. Though maybe that would be re-implementing the merging functionality that the `Configuration` class is meant to provide?

conda/cli/main_env_create.py
ryanskeith: @dholth, if there is better way to stack, I would use this as well.  What I was building off of was renit of the context with:
```
# reload the context
 ```
 This is after writing a temp .condarc file.  I found this fresh_context which seemed to be a more elegant approach where I could naturally add both arguments and items from the environment.yaml.

File: conda/cli/main_env_create.py (hunk 8 above)
Comment (ryanskeith) [reply conda/cli/main_env_create.py]: I can append for now.  I think I will add this case to the discussion of what is the priority order, overrides, etc when the environment.yaml defines.

**Review by ryanskeith — APPROVED** [review]
LGTM

**Comment by soapy1** [comment]
I think this needs one more reviewer before it can be merged.  I'm going to unresolve some comments for visibility.

File: conda/cli/main_env_create.py [hunk 9]
```diff
@@ -142,70 +170,93 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    # if env.config.channel_priority:
+    #     env_vars['CONDA_CHANNEL_PRIORITY'] = str(env.config.channel_priority)
+    # if env.config.solver:
+    #     env_vars['CONDA_SOLVER'] = env.config.solver
 
-    if args.dry_run:
-        installer_type = "conda"
-        installer = get_installer(installer_type)
+    # Use fresh_context to load all environment configuration without writing to disk
+    with fresh_context(argparse_args=args, env=env_vars):
```
kenodegard, line 190: This is not identical to an environment condarc since environment variables have higher precedence over the `$CONDARC` file.

conda/cli/main_env_create.py
kenodegard: Since both `conda-lock.yml` and `pixi.lock` also only support defining `channels` I'm ok with only supporting `channels` for the moment

conda/cli/main_config.py
```diff
@@ -848,13 +848,9 @@ def execute_config(args, parser):
...
     if os.path.exists(rc_path):
-        with open(rc_path) as fh:
-            # round trip load required because... we need to round trip
-            rc_config = yaml_round_trip_load(fh) or {}
+        rc_config = _read_rc(rc_path)
     elif os.path.exists(sys_rc_path):
-        # In case the considered rc file doesn't exist, fall back to the system rc
-        with open(sys_rc_path) as fh:
-            rc_config = yaml_round_trip_load(fh) or {}
+        rc_config = _read_rc(sys_rc_path)
```
jezdez, line 853: There _are_ differences between `_read_rc` and this implementation, namely that it uses pathlib's `read_text`, which will call [`io.text_encoding()`](https://docs.python.org/3/library/io.html#io.text_encoding): https://github.com/python/cpython/blob/49e83e31bd45e513a3caaa39b5789e95caa78ac1/Lib/pathlib/__init__.py#L774-L782. The use of `io.text_encoding()` might lead to a raised `EncodingWarning` in case users have [`PYTHONWARNDEFAULTENCODING`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNDEFAULTENCODING) set.

Another concern is that the old implementation passed in the file handler instead of a string.

ryanskeith: So we have two different parts of the code trying to accomplish a condarc read.  If there is a reason to worry about the Encoding warning, should we change the code in `_read_rc` to do the same thing here?

Looking at the ruamel-yaml docs, it can use a fh, string, or Path object (anything with read method).  Internally, that specific function variable is called [string](https://github.com/conda/conda/blob/6a1104eb00262251756d89b58b8f608037661ccc/conda/common/serialize/__init__.py#L35-L36).  In other places in the code, we seem to use both passing a file handle and strings.

conda/cli/main_env_create.py
ryanskeith: 1. Would this be a better approach than to you? The other approach was [closed out](https://github.com/conda/conda/pull/15138) in favor of this one.  Instead of writing to disk for the new .condarc, I moved to just creating the `fresh_context`.

2. Oh good thought. This is being done during Environment creation.  So the question is does the Environment RC file have a higher precedence/priority than the user or system RC file.  If it does, this seems like an ok thing to do.

soapy1: Hmmm, so I think the right precedence is (highest first):

1. Command line parameter
2. Environment variables
3. Config from environment file
4. Config files specified by ``$CONDARC`` env var
5. Config files (by parse order)

So I think setting the environment level condarc is not sufficient to get the right behaviour. That condarc config will be overwritten if a user specified something like:
```
CONDARC=/paht/to/my/other/condarc conda env create ...
```

I think the environment variable option is also not such a great option because this means re-implementing config merging logic.

conda/cli/main_config.py
jezdez: > So we have two different parts of the code trying to accomplish a condarc read. If there is a reason to worry about the Encoding warning, should we change the code in _read_rc to do the same thing here?

To be clear, the reason why I think we need to worry about that is that if a code change may lead to it being raised, it means conda will stop working as it is implicitly handling the the encoding of the file per platform. Since I don't think we require a specific encoding for condarc, I wonder if this would basically introduce a case in which user would have a UX regression.

> Looking at the ruamel-yaml docs, it can use a fh, string, or Path object (anything with read method). Internally, that specific function variable is called string. In other places in the code, we seem to use both passing a file handle and strings.

That's probably fine? I wish we'd use one way, so we can handle errors defensively.

soapy1:
I'm going to go ahead and close this as we've decided to reconsider how to approach this problem

PR #15206, opened 2025-09-12
Remove uses of `env_var` and `env_vars` contexts in `tests/core/test_subdir_data.py`

agriyakhetarpal:
### Description

Contributes towards #14095; this PR replaces these contexts in `tests/core/test_subdir_data.py` with `monkeypatch.setenv(<...>)`.

### Checklist - did you ...

### DIFF
--- a/tests/core/test_subdir_data.py
+++ b/tests/core/test_subdir_data.py
@@ -187,7 +187,7 @@ def test_use_only_tar_bz2(monkeypatch: MonkeyPatch, platform=OVERRIDE_PLATFORM):
 
 def test_subdir_data_coverage(monkeypatch: MonkeyPatch, platform=OVERRIDE_PLATFORM):
     # disable SSL_VERIFY to cover 'turn off warnings' line
-    monkeypatch.setattr("conda.models.channel.Channel._cache_", {}, raising=False)
+    monkeypatch.setattr("conda.models.channel.Channel._cache_", {})
     monkeypatch.setenv("CONDA_PLATFORM", platform)
     monkeypatch.setenv("CONDA_SSL_VERIFY", "false")
     reset_context()
--- a/tests/core/test_subdir_data.py
+++ b/tests/core/test_subdir_data.py
@@ -6,12 +6,12 @@ from logging import getLogger
 from os.path import join
 from pathlib import Path
 from time import sleep
+from typing import TYPE_CHECKING
 
 import pytest
 
 from conda import CondaError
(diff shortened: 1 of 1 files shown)


tests/core/test_subdir_data.py
```diff
@@ -138,62 +151,57 @@ def test_cache_fn_url_repo_anaconda_com():
...
     # disable SSL_VERIFY to cover 'turn off warnings' line
-    with (
-        ChannelCacheClear(),
-        env_vars(
-            {"CONDA_PLATFORM": platform, "CONDA_SSL_VERIFY": "false"},
-            stack_callback=conda_tests_ctxt_mgmt_def_pol,
-        ),
-    ):
-        channel = Channel(url_path(join(CHANNEL_DIR_V1, platform)))
+    monkeypatch.setattr("conda.models.channel.Channel._cache_", {}, raising=False)
```
agriyakhetarpal:
```suggestion
    monkeypatch.setattr("conda.models.channel.Channel._cache_", {})
```

tests/core/test_subdir_data.py
```diff
@@ -138,62 +151,57 @@ def test_cache_fn_url_repo_anaconda_com():
...
     # disable SSL_VERIFY to cover 'turn off warnings' line
-    with (
-        ChannelCacheClear(),
-        env_vars(
-            {"CONDA_PLATFORM": platform, "CONDA_SSL_VERIFY": "false"},
-            stack_callback=conda_tests_ctxt_mgmt_def_pol,
-        ),
-    ):
-        channel = Channel(url_path(join(CHANNEL_DIR_V1, platform)))
+    monkeypatch.setattr("conda.models.channel.Channel._cache_", {})
```
agriyakhetarpal, line 190: I also attempted to remove the `ChannelCacheClear()` context manager, but I think I might need to rework this. Would it make sense to have a fixture like this:

```python
@pytest.fixture
def clear_channel_cache():
    Channel._cache_.clear()
    yield
    Channel._cache_.clear()
```

or could I just call `Channel._cache_.clear()` at the start and the end of this test?

jaimergp: This seems to be enough. To be honest all these metaclass-driven caches should be cleared per test to avoid side effects.

Strictly speaking, the fixture would only need the second `clear()` after `yield`, if you want to go that way, but I prefer clearing it explicitly here without the magic of a fixture. It seems that in this case the point of this contextmanager was to guarantee it was cleared once the block exits, even if case of failures.

jezdez (approved):

PR #15213, opened 2025-09-14
Report frozen environments with `+`

jaimergp:
### Description

Closes https://github.com/conda-incubator/conda-self/issues/50

### Checklist - did you ...

### DIFF
--- a/conda/plugins/reporter_backends/console.py
+++ b/conda/plugins/reporter_backends/console.py
@@ -203,11 +203,11 @@ class ConsoleReporterRenderer(ReporterRendererBase):
         if not output:
             return ""
 
-        output = ["", "# conda environments:", "#", "# *  -> active", "# 🔒 -> frozen"]
+        output = ["", "# conda environments:", "#", "# *  -> active", "# + -> frozen"]
 
         def disp_env(prefix):
             active = "*" if prefix == context.active_prefix else " "
-            protected = "🔒" if Path(prefix, PREFIX_FROZEN_FILE).is_file() else "  "
+            protected = "+" if Path(prefix, PREFIX_FROZEN_FILE).is_file() else " "
             if prefix == context.root_prefix:
                 name = ROOT_ENV_NAME
             elif any(
--- a/conda/plugins/reporter_backends/console.py
+++ b/conda/plugins/reporter_backends/console.py
@@ -12,11 +12,16 @@ import sys
 from errno import EPIPE, ESHUTDOWN
 from itertools import cycle
 from os.path import basename, dirname
+from pathlib import Path
 from threading import Event, Thread
 from time import sleep
 from typing import TYPE_CHECKING
 
-from ...base.constants import DEFAULT_CONSOLE_REPORTER_BACKEND, ROOT_ENV_NAME
+from ...base.constants import (
+    DEFAULT_CONSOLE_REPORTER_BACKEND,
+    PREFIX_FROZEN_FILE,
+    ROOT_ENV_NAME,
+)
 from ...base.context import context
 from ...common.io import swallow_broken_pipe
 from ...common.path import paths_equal
@@ -198,10 +203,11 @@ class ConsoleReporterRenderer(ReporterRendererBase):
         if not output:
             return ""
 
-        output = ["", "# conda environments:", "#"]
+        output = ["", "# conda environments:", "#", "# *  -> active", "# 🔒 -> frozen"]
 
         def disp_env(prefix):
             active = "*" if prefix == context.active_prefix else " "
+            protected = "🔒" if Path(prefix, PREFIX_FROZEN_FILE).is_file() else "  "
             if prefix == context.root_prefix:
                 name = ROOT_ENV_NAME
             elif any(
@@ -210,7 +216,7 @@ class ConsoleReporterRenderer(ReporterRendererBase):
                 name = basename(prefix)
             else:
                 name = ""
-            return f"{name:20} {active} {prefix}"
+            return f"{name:20} {active} {protected} {prefix}"
 
         for env_prefix in prefixes:
             output.append(disp_env(env_prefix))
--- a/tests/cli/test_main_info.py
+++ b/tests/cli/test_main_info.py
@@ -112,7 +112,7 @@ def test_info_envs_frozen(conda_cli: CondaCLIFixture, tmp_env):
 
         stdout, stderr, err = conda_cli("info", "--envs")
         assert stdout == ConsoleReporterRenderer.envs_list(prefixes)
-        assert "🔒" in stdout
+        assert " + " in stdout
         assert not stderr
         assert not err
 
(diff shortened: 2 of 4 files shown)


jaimergp:
Example output:

```
conda env list

# conda environments:
#
# *  -> active
# 🔒 -> frozen
base                            /Users/jrodriguez/.local/conda
cf-scripts                      /Users/jrodriguez/.local/conda/envs/cf-scripts
conda-24.5                      /Users/jrodriguez/.local/conda/envs/conda-24.5
conda-build-dev                 /Users/jrodriguez/.local/conda/envs/conda-build-dev
conda-dev                  *    /Users/jrodriguez/.local/conda/envs/conda-dev
conda-docs                      /Users/jrodriguez/.local/conda/envs/conda-docs
frozen                       🔒 /Users/jrodriguez/.local/conda/envs/frozen
libmambapy211                   /Users/jrodriguez/.local/conda/envs/libmambapy211
```

tests/plugins/reporter_backends/test_console.py
```diff
@@ -51,7 +51,7 @@ def test_console_reporter_renderer_envs_list(mocker):
 
     output = console_reporter_renderer.envs_list(["/tmp/envs"])
 
-    assert "envs                   /tmp/envs" in output
```
ForgottenProgramme, line 54: what are we checking here? 🤔

ForgottenProgramme:
@jaimergp ,
I checked the output locally, it looks good. 👍🏼  
Could you answer my question above? After that I am ready to give this a green tick mark.

conda/plugins/reporter_backends/console.py
```diff
@@ -198,10 +203,11 @@ def envs_list(prefixes: Iterable[PathType], output=True, **kwargs) -> str:
         if not output:
             return ""
 
-        output = ["", "# conda environments:", "#"]
+        output = ["", "# conda environments:", "#", "# *  -> active", "# 🔒 -> frozen"]
 
         def disp_env(prefix):
             active = "*" if prefix == context.active_prefix else " "
+            protected = "🔒" if Path(prefix, PREFIX_FROZEN_FILE).is_file() else "  "
```
jezdez: Do we _need_ to use a Unicode lock here? I think we should not mix it with plaintext markers we already have here.

```suggestion
        output = ["", "# conda environments:", "#", "# *  -> active", "# ! -> frozen"]

        def disp_env(prefix):
            active = "*" if prefix == context.active_prefix else " "
            protected = "!" if Path(prefix, PREFIX_FROZEN_FILE).is_file() else "  "
```

tests/cli/test_main_info.py
```diff
@@ -103,6 +105,18 @@ def test_info_envs(conda_cli: CondaCLIFixture):
...
 
 
+def test_info_envs_frozen(conda_cli: CondaCLIFixture, tmp_env):
+    with tmp_env("ca-certificates") as prefix:
+        Path(prefix, PREFIX_FROZEN_FILE).touch()
+        prefixes = list_all_known_prefixes()
+
+        stdout, stderr, err = conda_cli("info", "--envs")
+        assert stdout == ConsoleReporterRenderer.envs_list(prefixes)
+        assert "🔒" in stdout
```
jezdez:
```suggestion
        assert "!" in stdout
```

conda/plugins/reporter_backends/console.py
jaimergp: We don't have to, but we can, and Yane has been advocating for more emoji use in the CLI in every UX review we've had. 🤷

tests/plugins/reporter_backends/test_console.py
jaimergp: Adding the protected marker adds a couple of extra spaces that needed to be adjusted in the existing tests. Arguably we could rewrite this to use a regex though, but 🤷

conda/plugins/reporter_backends/console.py
ForgottenProgramme: We do want to make `conda` a bit cooler, right? 🤷🏼 
I think the lock emoji looks good! 🔒  
That said I know that there were some issues in displaying checkmark emojis (for `conda doctor`) on windows. 🤷🏼‍♀️

jezdez: To be clear, thinking about this more, I'm concerned about two things:
- inconsistencies with the rest of the command output (we should use the reporter backends to switch to a wholly "rich" output and not port it piece by piece, to not break scripts and integrations)
- I don't think we've done enough review and prep in the various platform/shell combinations to know for sure that this won't break things. The `conda doctor` has the advantage of being a niche subcommand compared to `conda info`, and because it's relatively new, it has a lower risk of impacting existing scripts and build systems that call conda

tests/cli/test_main_info.py
jaimergp: Went with `+` because we might want to use `!` for broken environments in the future.

conda/plugins/reporter_backends/console.py
jaimergp: If you have seen emoji issues on Windows, then I'll revert.

That said, I don't know why a lot of widely used CLI tools get to use emoji in their regular output (e.g. `black`) but we are still stuck in ASCII world.

tests/cli/test_main_info.py
jezdez: Fair!

conda/plugins/reporter_backends/console.py
jezdez: Same, I'm not against emojis, creating a more up-to-date output is why we started to implement the reporter backend plugin hook, it was deprioritized, as you know, we do have https://github.com/conda-incubator/conda-rich that implements part of this, but it hasn't replaced the plain text output IIRC. @travishathaway may know more about this

travishathaway: I would personally prefer us to keep things more basic and not use emojis. I'm not sure if we'll ever pick up the reporter backends initiative again though.

jezdez: Pinging @dashagurova and @yane-anaconda about this, I would recommend to pick the reporters back up and launch in the next quarters, making a richer interface a possibility

jezdez (approved):

ForgottenProgramme (approved):

tests/cli/test_main_info.py
ForgottenProgramme: Good idea.

conda/plugins/reporter_backends/console.py
travishathaway: #MakeCondaBeatiful 💋 🌈

yane-anaconda:
Sorry seeing this late! But I like the lock. It's more clear that I can't do anything with that env. The "!" is usually used if there in an error/warning they should attend to and the visual language we are trying to use across anaconda products. But we can always get some extra opinions in the future. There is always the possible route of just putting the word "Locked" or "Protected". It should probably match the command terminology.

PR #15217, opened 2025-09-15
Drop `env_vars` usage in helpers and fixtures

agriyakhetarpal:
### Description

This PR removes these contexts from the `temp_package_cache()` fixture and some of the helpers in `conda/testing/helpers.py`, and it is among the last tasks to handle on #14095.

### Checklist - did you ...

### DIFF
--- a/tests/base/test_context.py
+++ b/tests/base/test_context.py
@@ -268,6 +268,7 @@ def test_context_parameters_have_descriptions(context_testdata: None):
 
 def test_context_override_with_reset(monkeypatch):
     from conda.base.context import context, reset_context
+
     original = context.add_pip_as_python_dependency
     with context._override("add_pip_as_python_dependency", False):
         assert context.add_pip_as_python_dependency is False


agriyakhetarpal:
Argh, my bad, I forgot to make some changes as I was only testing a subset of the test suite (`pytest -k test_solver`, not `pytest -k solve`). I've marked this as a draft again.

conda/testing/helpers.py
```diff
@@ -294,12 +298,11 @@ def _get_index_r_base(
...
-        with env_var(
-            "CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY",
-            str(add_pip).lower(),
-            stack_callback=conda_tests_ctxt_mgmt_def_pol,
-        ):
-            sd._process_raw_repodata_str(json.dumps(repodata))
+
+        monkeypatch.setenv("CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY", str(add_pip).lower())
+        reset_context()
+        sd._process_raw_repodata_str(json.dumps(repodata))
```
jaimergp: This introduces a large number of diff lines in the PR, and maybe we can get by without using `monkeypatch` here so we don't have to pass it everywhere.

```suggestion
        with context._override("add_pip_as_python_dependency", add_pip):
            sd._process_raw_repodata_str(json.dumps(repodata))
```

Let's see if this works?

agriyakhetarpal:
I agree, the diff is quite large as is. Let's try this!

conda/testing/helpers.py
```diff
@@ -299,9 +299,8 @@ def _get_index_r_base(
         sd = SubdirData(channel)
         subdir_datas.append(sd)
 
-        monkeypatch.setenv("CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY", str(add_pip).lower())
-        reset_context()
-        sd._process_raw_repodata_str(json.dumps(repodata))
+        with context._override("add_pip_as_python_dependency", add_pip):
+	        sd._process_raw_repodata_str(json.dumps(repodata))
```
agriyakhetarpal:
```suggestion
        with context._override("add_pip_as_python_dependency", add_pip):
            sd._process_raw_repodata_str(json.dumps(repodata))
```

jaimergp:
Looks like that didn't work 🤔 Pretty annoying that we have to send the fixture around... I guess `reset_context` usage in tests _resets_ the manual patch in `_override()`. We might need something smarter there.

Can you add a unit test for `context._override()` that tests it against `reset_context()`? Something like:

```python
def test_context_override(monkeypatch):
  with context._override("add_pip_as_python_dependency", False):
    assert context.add_pip_as_python_dependency is False
    monkeypatch.setenv("CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY", "True")
    reset_context()
    # We assume it doesn't change buuut maybe it does?
    assert context.add_pip_as_python_dependency is False
```

I suspect we might need to change `_override()` to something like this:

```python
    def _override(self, key: str, value: Any) -> Iterator[None]:
        envkey = f"CONDA_{key.upper()}"
        oldenv = os.environ.pop(envkey, None)
        os.environ[envkey] = str(value)
        old = getattr(self, key)
        setattr(self, key, value)
        try:
            yield
        finally:
            setattr(self, key, old)
            if oldenv:
                os.environ[envkey] = oldenv
            else:
                del os.environ[envkey]
```

jaimergp:
Gentle ping @agriyakhetarpal

dholth:
What if we had a new version of the env_vars function that accepted a monkeypatch parameter, so that the code looked mostly the same

agriyakhetarpal:
Hmm, I had actually tried adding the test case previously and noticed it passing locally. Let me fix the merge conflicts that have occurred in the meantime.

Using it in the actual code still causes test failures, as `_override()` doesn't correctly handle environment variables that `reset_context()` is reading from. I think we have the following options:
- modify `_override()` to also set the environment variable
- Keep using `monkeypatch`, but perhaps restructure to minimise where it's passed (not sure how or how much)
- something else?

> What if we had a new version of the env_vars function that accepted a monkeypatch parameter, so that the code looked mostly the same

Thanks, @dholth. However, I think we're eventually trying to deprecate this function; I'm not sure if that would work? (unless I am misunderstanding you, sorry!).

agriyakhetarpal:
@jaimergp – I removed `monkeypatch.setenv()` from `_get_solver_base()` too, not just `_get_index_r_base` (mixing up `context._override()` and `monkeypatch` was likely the problem). That should work! Let me see if that replicates on CI as well, or if my local test suite is playing with me.

conda/testing/helpers.py
```diff
@@ -254,6 +257,7 @@ def _patch_for_local_exports(name, subdir_data):
 
 
 def _get_index_r_base(
+    monkeypatch: MonkeyPatch,
```
jaimergp: You don't need this anymore, do you? If so, we can remove and clear the diff quite a bit.

agriyakhetarpal: Yup, working through removing them locally :)

agriyakhetarpal:
Hmm, I don't think `context._override()` is doing enough – the `monkeypatch.setenv()` approach was forcing the solver via an environment variable, but this only overrides it for the scope of the context manager

conda/testing/fixtures.py
```diff
@@ -169,38 +166,37 @@ def test_foo():
...
     request: FixtureRequest,
-    monkeypatch: MonkeyPatch,
     solver: Solver,
 ) -> Iterable[Solver]:
     # clear cached solver backends before & after each test
     context.plugin_manager.get_cached_solver_backend.cache_clear()
     request.addfinalizer(context.plugin_manager.get_cached_solver_backend.cache_clear)
 
-    monkeypatch.setenv("CONDA_SOLVER", solver)
+    mp = request.getfixturevalue("monkeypatch")
```
jaimergp, line 197: Nice.

agriyakhetarpal:
Hi @jaimergp, this should be ready to go as the tests have been passing. I've merged the `main` branch just now, and I had previously added a test as per your suggestion. Thank you!

tests/base/test_context.py
```diff
@@ -266,6 +266,16 @@ def test_context_parameters_have_descriptions(context_testdata: None):
...
 
 
+def test_context_override_with_reset(monkeypatch):
+    from conda.base.context import context, reset_context
+
+    with context._override("add_pip_as_python_dependency", False):
+        assert context.add_pip_as_python_dependency is False
+        monkeypatch.setenv("CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY", "True")
+        reset_context()
+        assert context.add_pip_as_python_dependency is False
```
jaimergp:
```suggestion
    original = context.add_pip_as_python_dependency
    with context._override("add_pip_as_python_dependency", False):
        assert context.add_pip_as_python_dependency is False
        monkeypatch.setenv("CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY", "True")
        reset_context()
        assert context.add_pip_as_python_dependency is False
    assert context.add_pip_as_python_dependency is original
```

jaimergp:
pre-commit.ci autofix

agriyakhetarpal:
Thanks for the adjustment!

PR #15232, opened 2025-09-16
Make constant for '.condarc'

soapy1:
### Description

This PR adds a constant for `.condarc` and `condarc` file names. It also replaces instances of the magic string ".condarc" with the constant.

### Checklist - did you ...


soapy1:
pre-commit.ci autofix

travishathaway:
Thanks for doing this! Good idea!

conda/common/configuration.py
```diff
@@ -1333,7 +1333,8 @@ def __call__(cls, *args, **kwargs):
         return self
 
 
-CONDARC_FILENAMES = (".condarc", "condarc")
+DEFAULT_CONDARC_FILENAME: Final = ".condarc"
+CONDARC_FILENAMES = (DEFAULT_CONDARC_FILENAME, "condarc")
```
kenodegard: Since the non-`.` is also used (e.g., `sys_rc_path`) should we include that as well?

```suggestion
DEFAULT_CONDARC_FILENAME: Final = ".condarc"
ALTERNATIVE_CONDARC_FILENAME: Final = "condarc"
CONDARC_FILENAMES = (DEFAULT_CONDARC_FILENAME, ALTERNATIVE_CONDARC_FILENAME)
```

ryanskeith (commented):
I like all these changes that are being made.

Do we also want to make a handy function that returns the .condarc path for the dev?  Though it may not be used enough to warrant a function.

soapy1:
>Do we also want to make a handy function that returns the .condarc path for the dev?

I'm not sure what you mean by this? Which condarc path?

ryanskeith:
@soapy1, I was thinking something like:
```
def get_conda_rc_path(env: bool)
```
It seems it may be a common pattern that we are doing.   Though it may be so little that it is not worth it.

soapy1:
hmmm, there are lots of places where conda looks for condarc files: https://github.com/conda/conda/blob/main/conda/base/constants.py#L36-L72. I would say, lets not write that helper function until we actually need it. Or at least it's out of scope for this PR.

ryanskeith (approved):
These look great to me.

PR #15236, opened 2025-09-17
Dont warn invalid environment.yml keys multiple times

soapy1:
### Description

Previously, the environment.yml plugin would produce duplicate errors for invalid keys. For example, consider  the environment.yml:

```
channels:
  - anaconda
dependencies:
  - nltk
variables:
  VARIABLE: value
wrong: surely
```

would produce:
```
$ conda env create --file tests/env/support/test.yml --env-spec environment.yml

EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong


EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong

. . .
```

This change makes the message only appear once:
```
$ conda env create --file tests/env/support/test.yml --env-spec environment.yml

EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong
``` 
### Checklist - did you ...

### DIFF
--- a/conda/env/specs/yaml_file.py
+++ b/conda/env/specs/yaml_file.py
@@ -66,5 +66,5 @@ class YamlFileSpec(EnvironmentSpecBase):
     @property
     def env(self) -> Environment:
         if not self._environment:
-            self.can_handle()
+            self._environment = env.from_file(self.filename, suppress_invalid_key_warning=True)
         return self._environment.to_environment_model()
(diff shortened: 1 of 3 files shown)


conda/env/specs/yaml_file.py
```diff
@@ -66,5 +66,5 @@ def environment(self) -> EnvironmentYaml:
     @property
     def env(self) -> Environment:
         if not self._environment:
-            self.can_handle()
+            self._environment = env.from_file(self.filename, suppress_invalid_key_warning=True)
```
soapy1, line 69: We don't really need to be calling the `can_handle` function here. The `env` function is not responsible for validation, it just needs to load the file.

Issue #15139, opened 2025-08-27
Investigate improvements to environment spec plugin api


soapy1:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Currently, the environment spec plugin api requires plugin authors to accept a source/file path as part of initialization. For example:
```
def __init__(self, source_path: str):
    # code to read the content
```

This means that every plugin needs to open up the file and read the same content. As the number of plugins grow, this can become a performance bottle neck. It would reduce the amount of io operations conda had to perform if plugins accepted the contents of the environment spec instead. Then conda (core) could do things like cache the file contents and provide a standardized way to read from different file sources. 

### What should happen?

* Investigate if this approach provides an improvement in performance
* If so, let's update the api + all existing plugins that depend on these plugin hooks

soapy1:
Some other improvements include:
- [ ] better error messages when a plugin's `can_handle` function fails
- [ ] when a user selects an env spec plugin by name, conda should share any errors from the plugin with the user

PR #15237, opened 2025-09-17
Don't load environment.yml as part of its can_handle

soapy1:
### Description

Previously, the environment.yml plugin would produce duplicate errors for invalid keys. For example, consider  the environment.yml:

```
channels:
  - anaconda
dependencies:
  - nltk
variables:
  VARIABLE: value
wrong: surely
```

would produce:
```
$ conda env create --file tests/env/support/test.yml --env-spec environment.yml

EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong


EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong

. . .
```

This change makes the message only appear once:
```
$ conda env create --file tests/env/support/test.yml --env-spec environment.yml

EnvironmentSectionNotValid: The following section on 'tests/env/support/test.yml' is invalid and will be ignored:
 - wrong
``` 
### Checklist - did you ...

### DIFF
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -289,7 +289,7 @@ def from_yaml(yamlstr: str, **kwargs) -> EnvironmentYaml:
         deprecated.topic(
             "26.3",
             "26.9",
-            topic="Using a non-compliant CEP-0024 environment file",
+            topic="The environment file is not fully CEP 24 compliant",
             addendum=(
                 "In the future, this configuration will be rejected. Please fix the following "
                 "errors in order to make the configuration valid: "
--- a/conda/env/specs/yaml_file.py
+++ b/conda/env/specs/yaml_file.py
@@ -8,6 +8,7 @@ import os
 from logging import getLogger
 from typing import TYPE_CHECKING
 
+from ...common.serialize import yaml_safe_load
 from ...deprecations import deprecated
 from ...exceptions import CondaValueError
 from ...plugins.types import EnvironmentSpecBase
@@ -41,8 +42,7 @@ class YamlFileSpec(EnvironmentSpecBase):
         This can handle if:
             * the provided file exists
             * the provided file ends in the supported file extensions (.yaml or .yml)
-            * the env file can be interpreted and transformed into
-              a `conda.env.env.Environment`
+            * the yaml file can be loaded and is not empty
 
         :return: True or False
         """
@@ -57,11 +57,15 @@ class YamlFileSpec(EnvironmentSpecBase):
             return False
 
         try:
-            self._environment = env.from_file(self.filename)
-            return True
+            yamlstr = env.load_file(self.filename)
+            data = yaml_safe_load(yamlstr)
+            if data is None:
+                return False
         except Exception:
             log.debug("Failed to load %s as a YAML.", self.filename, exc_info=True)
             return False
+
+        return True
 
     @property
     @deprecated("26.3", "26.9", addendum="This method is not used anymore, use 'env'")
@@ -69,7 +73,7 @@ class YamlFileSpec(EnvironmentSpecBase):
         if not self._environment:
             if not self.can_handle():
                 raise CondaValueError(f"Cannot handle environment file: {self.msg}")
-            # can_handle() succeeded and set self._environment, so it should not be None
+            self._environment = env.from_file(self.filename)
 
         if self._environment is None:
             raise CondaValueError("Environment could not be loaded")
@@ -78,5 +82,5 @@ class YamlFileSpec(EnvironmentSpecBase):
     @property
     def env(self) -> Environment:
         if not self._environment:
-            self.can_handle()
+            self._environment = env.from_file(self.filename)
         return self._environment.to_environment_model()
--- a/news/15237-dont-parse-whole-environment-as-part-of-can_handle
+++ b/news/15237-dont-parse-whole-environment-as-part-of-can_handle
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Don't load the environment.yml environment as part of it's `can_handle` function. This function should just verify if the plugin will be able to interpret the provided file. For the environment.yml plugin, it can read any `.yml`/`.yaml` file. (#15237)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/env/test_create.py
+++ b/tests/env/test_create.py
@@ -400,3 +400,26 @@ def test_create_env_custom_platform(
 
         assert config.is_file()
         assert f"subdir: {platform}" in config.read_text()
+
+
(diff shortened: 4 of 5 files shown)


conda/env/specs/yaml_file.py
```diff
@@ -44,20 +42,15 @@ def can_handle(self):
         if not any(spec_ext == file_ext for spec_ext in YamlFileSpec.extensions):
             return False
 
-        try:
-            self._environment = env.from_file(self.filename)
```
soapy1, line 60: The `environment.yml` plugin is able to handle any yaml file thrown at it. As long as the file can be read and parsed to yaml, it's good to go. The work in https://github.com/conda/conda/issues/15139 will update the environment spec api to read and cache the file contents.

conda/env/specs/yaml_file.py
```diff
@@ -66,5 +59,5 @@ def environment(self) -> EnvironmentYaml:
     @property
     def env(self) -> Environment:
         if not self._environment:
-            self.can_handle()
+            self._environment = env.from_file(self.filename)
```
soapy1, line 85: now we can load the environment when the user asks for the environment to be loaded.

conda/env/env.py
```diff
@@ -124,12 +270,35 @@ def from_environment(
...
         raise EnvironmentFileEmpty(filename)
+
+    # Perform schema validation. This will output a warning for any invalid schema.
+    errors = get_schema_errors(data)
+    if errors:
+        # Warn for all the schema errors in the environment
+        deprecated.topic(
+            "26.3",
+            "26.9",
+            topic="Using a non-compliant CEP-0024 environment file",
```
jezdez:
```suggestion
            topic="The environment file is not fully CEP 24 compliant",
```

dholth:
What is its [can_handle](https://www.printables.com/model/203804-can-handle)?

kenodegard:
is this intended to be a followup to https://github.com/conda/conda/pull/15120?

soapy1:
>is this intended to be a followup to https://github.com/conda/conda/pull/15120?

yep. It uses some of the changes in #15120

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Environment object describing the conda environment.yaml file."""
 
+from __future__ import annotations
+
 import os
 import re
 from itertools import chain
+from typing import TYPE_CHECKING
```
ryanskeith, line 10:
```suggestion
from typing import Any
```
As we aren't doing any imports and using `from __future__ import annotations`, I don't think we need the TYPE_CHECKING block at this time.

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
 from ..gateways.connection.download import download_text
 from ..gateways.connection.session import CONDA_SESSION_SCHEMES
 from ..history import History
 from ..models.environment import Environment as EnvironmentModel
 from ..models.environment import EnvironmentConfig
 from ..models.match_spec import MatchSpec
 
-VALID_KEYS = ("name", "dependencies", "prefix", "channels", "variables")
+if TYPE_CHECKING:
+    from typing import Any
```
ryanskeith, line 35:
```suggestion
```
See above comment.

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+def channels_validation(channels: list):
+    """Validate the contents of the channels field.
+
+    Will ensure:
+      * channels is a list
+      * all entries are strings
+    """
+    field_type_validation("channels", channels, list)
+
+    for channel in channels:
```
ryanskeith, line 119: This loop is interesting to me as it will impose YAML typing rules onto channels names.  So I couldn't for install name a channel 1234, true, or null.

This may be ok as the above are horrible names but it is worth keeping in mind that name rules are being imposed here.

conda/env/env.py
```diff
@@ -2,26 +2,172 @@
...
+def variables_validation(variables: dict[str, str]):
+    """Validate the contents of the variables field.
+
+    Will ensure:
+      * variables is a dict
+      * all entries are strings
+    """
+    field_type_validation("variables", variables, dict)
+    for value in variables.values():
+        if not isinstance(value, str):
```
ryanskeith: This may cause problems as environment variable values are regularly numbers and booleans.

conda/env/specs/cep_24.py
```diff
@@ -0,0 +1,64 @@
...
+            return False
+
+        try:
+            yamlstr = env.load_file(self.filename)
+            data = yaml_safe_load(yamlstr)
+            errors = env.get_schema_errors(data)
+            if errors:
+                return False
+            return True
+        except Exception:
```
ryanskeith: This is a big Exception.  It may be nice to have at least one split to differentiate between file and system errors and other errors to narrow down what the problem is for the user.  Help them determine if it is the file or the contents of the file.

Right now we are ok as I think the biggest one, "if the file exists", happens further up in the cli chain but an API, it might be nice to check as well.

soapy1:
@ryanskeith it looks like all this review has more to do with https://github.com/conda/conda/pull/15120. I'll apply the changes there. Lemme filp this one into draft mode so it's more clear that this is a follow up PR.

conda/env/env.py
soapy1: I tested this out and found some interesting results. Currently (on this branch as well as the main branch) if you have an environment.yml like

```
name: 1234
dependencies:
  - python
```

conda will crash out when it tries to determin the target prefix

```
. . .
      File "/home/sophia/projects/conda/conda/cli/main_env_create.py", line 137, in execute
        prefix = determine_target_prefix(context, args)
      File "/home/sophia/projects/conda/conda/base/context.py", line 2304, in determine_target_prefix
        if prefix_name is not None and not prefix_name.strip():  # pragma: no cover
    AttributeError: 'int' object has no attribute 'strip'
```
I'll create an issue for this bug.

Users can name thier environments 1234, true or null, they just need to be strings. I think this is the right behaviour.  So, something like:

```
name: "1234"
dependencies:
  - python
 ```
works ok.

soapy1: https://github.com/conda/conda/issues/15250

soapy1: oh, I missread this as "name" instead of channel. My bad. I don't see how one could have a channel named as not a string type. This refers to a url or path, so it should never be a bool or int.

conda/env/env.py
soapy1: You're right! This validation is too strict, backing it out.

conda/env/env.py
soapy1: This produces an error https://results.pre-commit.ci/run/github/6235174/1758557431.7KHA_cK-RMuk-35ai4IdHg

kenodegard (approved):

PR #15142, opened 2025-08-28
Add `CondaMultiPlatformEnvironmentExporter`

kenodegard:
### Description

Add subclass of `CondaEnvironmentExporter` to support multiplatform exports.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15240, opened 2025-09-18
Alternative `CondaMultiPlatformEnvironmentExporter`

kenodegard:
…nvironmentExporter

### Description

Alternative to https://github.com/conda/conda/pull/15142

The code is identical with the exception of abstracting the environment extrapolation logic away from `conda.cli.main_export`. Instead each CondaEnvironmentExporter implements a `__call__` that standardizes erroring (if multiple platforms are requested for a single platform exporter) or extrapolating (for each platform requested).

### Checklist - did you ...


conda/plugins/types.py
```diff
@@ -522,3 +524,32 @@ def __post_init__(self):
...
+
+    :param name: name of the exporter (e.g., ``environment-yaml-multi``)
+    :param aliases: user-friendly format aliases (e.g., ("yaml",))
+    :param default_filenames: default filenames this exporter handles (e.g., ("environment.yml", "environment.yaml"))
+    :param export: callable that exports an Environment collection to string format
+    """
+
+    export: Callable[[Iterable[Environment]], str]
+
+    def __call__(self, env: Environment) -> str:
```
soapy1, line 551: Hmmm, I think this is a neat idea. I like that:
* the plugin does not need to do any extra work to extrapolate the environment per platform
* the plugin interface is simple - it just takes a list of environments

But, I think this pattern for using `__call__` is pretty different from the way the rest of the plugins work, and introduces a new pattern for interacting with plugins. imo, it would be better to follow existing patterns.

Issue #15239, opened 2025-09-18
Add `sharded` to `experimental` features


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

We'll be soon testing sharded repodata currently being developed in the conda-libmamba-solver project. We would like include this as an experimental feature we can reference in conda's configuration system.

kenodegard:
Since `sharded` is implemented in a separate plugin it seems inappropriate to hardcode a constant in `conda` to enable it, should we make this list pluggable instead?

kenodegard:
Also, once working on this particular piece of code we should move the `--experimental` out of the `add_parser_channels` helper and into a new `add_parser_experimental`.

travishathaway:
> Since `sharded` is implemented in a separate plugin it seems inappropriate to hardcode a constant in `conda` to enable it, should we make this list pluggable instead?

@kenodegard, I think ideally this would be a nice thing to do, but I currently see the benefit of doing so as extremely marginal. Let's just go the easy route this time and add a new string as a possible choice for that option.

Also, could you please add your requested refactor under the list of, "What should happen?" 🙏

danyeaw:
Is this still needed now that we have the environmental variable feature flag?

travishathaway:
No, will close.

PR #15244, opened 2025-09-19
[POC] Add `CondaExperimentalFeature` to expand `--experimental` choices

kenodegard:
### Description

Inspired by https://github.com/conda/conda/issues/15239

- Adds a new `LazyMixin` to replace the `LazyChoicesAction`.
- Updates `conda.cli.conda_argparse.ArgumentParser` to check `choices` for us (instead of leaving it to the action).
- Updates the `HelperFormatter` to respect newlines in descriptions.

### Checklist - did you ...

### DIFF
--- a/news/15239-add-experimental-features
+++ b/news/15239-add-experimental-features
@@ -1,0 +1,20 @@
+### Enhancements
+
+* Add `conda.cli.actions.LazyMixin` to support lazy evaluated fields for ArgumentParser arguments. (#15244)
+* Add `conda.plugins.types.CondaExperiementalFeature` plugin type allowing downstream packages to extend the experimental feature selection that can be toggled via `--experimental`. (#15239 via #15244)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Mark `conda.cli.helpers.LazyChoicesAction` as pending deprecation, to remove in 26.9. Use `conda.cli.actions.LazyStoreAction` instead. (#15244)
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/cli/helpers.py
```diff
@@ -10,16 +10,48 @@
...
+        return maybecall(self._help)
+
+    @help.setter  # type: ignore[no-redef]
+    def help(self, value):
+        """Store value as is, if it is a callable it will be evaluated on first access."""
+        self._help = value
+
+
+LazyAppendAction = type("LazyAppend", (LazyMixin, _AppendAction), {})
+LazyStoreAction = type("LazyStore", (LazyMixin, _StoreAction), {})
```
kenodegard:
```suggestion
LazyAppendAction = type("LazyAppendAction", (LazyMixin, _AppendAction), {})
LazyStoreAction = type("LazyStoreAction", (LazyMixin, _StoreAction), {})
```

conda/cli/actions.py
```diff
@@ -56,3 +63,30 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+            cache = super().__getattribute__("_cache_")
+        except AttributeError:
+            cache = self._cache_ = {}
+
+        # populate cache if value doesn't exist
+        if name not in cache:
+            cache[name] = maybecall(super().__getattribute__(name))
+
+        # return cached value
+        return cache[name]
```
jezdez, line 88: Huh, that seems interesting, did you create that yourself? I'm not super stoked about us extending argparse in such a way since it diverts from the core behavior a lot and laziness implementations often lead to unexpected side-effects in code that expects concrete implementations.

conda/cli/helpers.py
```diff
@@ -456,8 +465,8 @@ def add_parser_solver(p: ArgumentParser) -> None:
     group.add_argument(
         "--solver",
         dest="solver",
-        action=LazyChoicesAction,
-        choices_func=context.plugin_manager.get_solvers,
+        action=LazyStoreAction,
+        choices=lambda: sorted(context.plugin_manager.get_solvers()),
```
jezdez, line 469: Alright, that's quite the change in behavior, have we tested this manually appropriately? I'm not sure if the test suite covers this well enough.

conda/cli/actions.py
kenodegard: The original implementation was:

```python
class LazyMixin:
    @memoizedproperty
    def choices(self):
        return maybecall(self._choices)

    @choices.setter  # type: ignore[no-redef]
    def choices(self, value):
        self._choices = value

    @memoizedproperty
    def help(self):
        return maybecall(self._help)

    @help.setter  # type: ignore[no-redef]
    def help(self, value):
        self._help = value
```

Happy to return to this.

The current implementation is just borrowing stuff from `conda.auxlib.decorators.memoizedproperty` and abstracting away from the explicit `choices`/`help` so any value can be lazy, which is most likely overkill.

conda/cli/conda_argparse.py
```diff
@@ -228,12 +242,33 @@ def _check_value(self, action, value):
...
+
         # extend to properly handle when we accept multiple choices and the default is a list
-        if action.choices is not None and isiterable(value):
+        if isiterable(value):
             for element in value:
-                super()._check_value(action, element)
+                check_value(action, element)
         else:
-            super()._check_value(action, value)
+            check_value(action, value)
```
kenodegard, line 271: If we're ok with how Python changed from quoting the choices to unquoting the choices (see the test suite below) in stdlib's `ArgumentParser.check_value` we can simplify this:

```python
if action.choices is None:
    # no choices specified, nothing to check
    return
elif isiterable(value):
    for element in value:
        super().check_value(action, element)
else:
    super().check_value(action, value)
```

conda/cli/helpers.py
kenodegard: Testing can definitely be improved, was looking for initial approval before dedicating additional time to this

conda/cli/actions.py
dholth: I like the idea of not putting `@memoizedproperty` everywhere since how many times are we parsing the arguments.
We might call this "deferred" instead of "lazy".

conda/cli/conda_argparse.py
dholth: Very OK with letting Python choose the quoting

Issue #15159, opened 2025-09-03
Update export plugin hook to support multiplatform formats


kenodegard:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

The export plugin needs to support receiving multiple `conda.models.environment.Environment` instances (e.g., one per platform) allowing the plugins to implement exporting multiple platforms to a single export format.

### Why is this needed?

This is necessary to support multiplatform lockfiles.

### What should happen?

Add a new subclass of `conda.plugins.types.CondaEnvironmentExporter` (e.g., `CondaMultiPlatformEnvironmentExporter`) to support export formats that can represent multiple platforms.

PR #15253, opened 2025-09-23
Add `CondaEnvironmentExporter.multiplatform_export`

kenodegard:
### Description

Alternative to #15142 & #15240

Returns to the idea that `conda.plugins.types` are only supposed to be containers moving stuff from the plugin into `conda` and not implementors of any logic (which plugins could modify).

Closes https://github.com/conda/conda/issues/15159 (the issue)
Closes #15142 (orig PR)
Closes #15240 (alt PR)

### Checklist - did you ...


conda/cli/main_export.py
```diff
@@ -189,7 +186,20 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
-    exported_content = environment_exporter.export(env)
+    signature = inspect.signature(environment_exporter.export).parameters
+    if "extra_platforms" in signature:
```
soapy1: Does this mean that plugins can either implement an export function with the function signature

```
def multi_platform_export(
        self, env: Environment, extra_envs: Iterable[Environment]
    ) -> str:
```

or 

```
def multi_platform_export(
        self, env: Environment, extra_platforms: Iterable[str]
    ) -> str:
```

What is the goal of providing both types of interfaces? I think it would be better to commit to supporting just one of these patterns.

conda/plugins/types.py
```diff
@@ -37,6 +38,12 @@
         dict[str, PrefixRecord],
     ]
 
+    SinglePlatformEnvironmentExport = Callable[[Environment], str]
+    MultiPlatformEnvironmentExport = (
+        Callable[[Environment, Iterable[Environment]], str]
```
soapy1: Is the first `Environment` meant to be treated any differently from any of the other environments in the second argument `Iterable[Environment]`? Or, what is the motivation for seperating these opposed to something like ` Callable[[Iterable[Environment]], str]`?

conda/cli/main_export.py
kenodegard: it occurred to me while implementing https://github.com/conda-incubator/conda-lockfiles/pull/55 that a plugin may have a better idea for how to extrapolate an environment and wants to derive the other envs on its own 🤷🏼‍♂️

but on reflection it sounds like that would be better off as a different plugin hook, exporters should only care about exporting envs to a format, not doing any extrapolation

I'll remove `extra_platforms`

conda/plugins/types.py
kenodegard: Well the idea being that there has to be at least one env (the current platform) so we skip needing to do any length checks

I also found myself doing `env, *envs = envs` in every example implementation

news/15159-add-CondaMultiPlatformEnvironmentExporter
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add `CondaMultiPlatformEnvironmentExporter`, a subclass of `CondaEnvironmentExporter`, supporting export formats with multiple platforms. (#15159 via #15240)
```
kenodegard:
```suggestion
* Add `CondaEnvironmentExporter.multiplatform_export` to support export formats with multiple platforms. (#15159 via #15253)
```

soapy1 (approved):

conda/cli/main_export.py
```diff
@@ -189,7 +184,21 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
-    exported_content = environment_exporter.export(env)
+    if environment_exporter.multiplatform_export:
```
ryanskeith, line 198: Could we comment this if else statement?  It seems confusing as far as logic goes.  The first clause is doing a multiplatform export but the second says that Multiple platforms are not support.  I know there is a nuanced difference between the two operations (if I understand the modalities correctly).

kenodegard: Is https://github.com/conda/conda/pull/15253/commits/4d20751404f355564953241064f6ac0a10e792d4 sufficient?

ryanskeith: The code that runs is clear to me. So I don't think lines 191, 195 and 202 are necessary.  The logic is what is confusing me.

The problem is I am reading it this way:
```
if environment_export is multiplatform:
   export multiple platforms
elif context says to export multiple platforms:
   This is an error. I can't export multiple platforms
```

kenodegard: reworked it in https://github.com/conda/conda/pull/15253/commits/c0c6d1af5643e203c4fc7b13ef80e0e877fb0463, does it read better now?

ryanskeith: Yes!

conda/cli/main_export.py
```diff
@@ -189,7 +193,16 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
-    exported_content = environment_exporter.export(env)
+    # Export using the appropriate method
+    envs = [env.extrapolate(platform) for platform in context.export_platforms]
+    if environment_exporter.multiplatform_export:
+        exported_content = environment_exporter.multiplatform_export(envs)
+    elif environment_exporter.export:
+        exported_content = environment_exporter.export(envs[0])
```
soapy1, line 201: should this be the current platform env opposed to the first env in the extrapolated list of export platforms?

```suggestion
        exported_content = environment_exporter.export(env)
```

soapy1 (approved):

conda/cli/main_export.py
kenodegard: by using the first env from the extrapolated list means we support single platform exports for a platform different from the current platform, e.g., say we're on `osx-arm64` by using `envs[0]` we can do:

```
conda export --override-platforms --platform win-64 --format explicit
```

PR #15254, opened 2025-09-23
Use `EnvironmentConfig.channels` when extrapolating envs

kenodegard:
### Description

When extrapolating the environments for different platforms don't rely on `context` being set instead pass in the values directly to the solver.

### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -566,7 +566,7 @@ class Environment:
             with repodata_manager as repodata_fn:
                 solver = solver_backend(
                     prefix=prefix,
-                    channels=context.channels,
+                    channels=self.config.channels,
                     subdirs=(platform, "noarch"),
                     specs_to_add=requested_packages,
                     repodata_fn=repodata_fn,


conda/models/environment.py
```diff
@@ -566,7 +566,7 @@ def extrapolate(self, platform: str) -> Environment:
             with repodata_manager as repodata_fn:
                 solver = solver_backend(
                     prefix=prefix,
-                    channels=context.channels,
+                    channels=self.config.channels,
```
soapy1, line 569: Hmmm, I think this means that the channels for the extrapolated envs might be different from the current platform env. For example, consider the user `.condarc`

```
channels:
  - defaults
  - conda-forge
  - my-custom-channel
```

And an environment file with the channel configuration
```
channels:
  - conda-forge
```

Then, installing the environment locally will pull from the channels `conda-forge`, `defaults`, and `my-custom-channel`. But extrapolating will produce an environment with the channels `conda-forge`. So, this solve will be substantially different from the one that installs the environment onto the current system.

Also, channels is an optional field in the environment.yml ([cep-24](https://github.com/conda/ceps/blob/main/cep-0024.md#channels)) and is not required by the `EnvironmentConfig` model. So, there is a good chance of `self.config.channels` being empty. I assume this would produce an error (since the defauls channel will no longer be injected anywhere)?

So, I think this needs to use `context.channels`.

PR #15255, opened 2025-09-23
Deprecate `conda.plugins.hookspec.spec_name`

kenodegard:
### Description

Is would seem that the `conda.plugins.hookspec.spec_name` constant is identical to `conda.base.constants.APP_NAME` so we should deprecate one in favor of the other.

### Checklist - did you ...


dholth:
Since this is specifically for third party extensions I would keep the alias forever

PR #15265, opened 2025-09-26
Fix `conda.exports.__getattr__`

kenodegard:
### Description

Fix issue where `conda.exports.ANYTHING` would always return `None`.

Caught while testing `conda 25.9.x`.

### Checklist - did you ...

### DIFF
--- a/conda/exports.py
+++ b/conda/exports.py
@@ -104,12 +104,15 @@ string_types = str
 text_type = str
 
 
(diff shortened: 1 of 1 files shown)


ryanskeith (approved):
Looks great.  `__getattr__` 's job is either to return the attribute name (which it is doing) OR raise the `AttributeError` which it is now doing.

PR #15270, opened 2025-09-27
Feat/conda witness

kriscoleman:
### Checklist - did you ...

### DIFF
--- a/.github/workflows/conda-witness-integration-test.yml
+++ b/.github/workflows/conda-witness-integration-test.yml
@@ -6,32 +6,33 @@ on:
     branches:
       - feat/conda-witness
     paths:
-      - '.github/workflows/conda-witness-integration-test.yml'
+      - ".github/workflows/conda-witness-integration-test.yml"
+      - "Makefile"
 
 permissions:
   contents: read
-  id-token: write  # For Sigstore signing
+  id-token: write # For Sigstore signing
 
 jobs:
   integration-test:
     name: Test Conda + Witness Integration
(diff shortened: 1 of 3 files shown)


jezdez:
Hi @kriscoleman, before you continue, could you please open an issue with a feature request and explain a little what witness.dev is any why this is useful for conda? Thank you!

jezdez:
@kriscoleman Thanks again for opening this pull request. Since I haven't heard from you after my comment above, and you pushed a few commits since then, I just wanted to respond quickly that while [`witness`](https://github.com/in-toto/witness) seems to be an interesting CNCF/in-toto project, there is no need to add it to the conda CLI code directly.

As such, I'm going to close this pull request as "not planned" but would strongly encourage you to implement it as a separate [conda plugin](https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/index.html) (see many examples on https://github.com/conda-incubator/conda-plugins) instead. The subcommand plugin hook can be used to extend conda with a `witness` related subcommand (looking at the code, there is no change in conda's core needed).

Also, if you're interested, you could ask to develop the project (e.g. a `conda-witness` package?) via the [conda incubator](https://github.com/conda/governance?tab=readme-ov-file#incubation).

PS: In case you didn't know, there also is a legacy [`conda verify`](https://github.com/conda/conda-verify) project which I wouldn't recommend overlapping like you're proposing here. Maybe use a differently named subcommand to not create an unexpected user experience?

Thanks again!

PR #15272, opened 2025-09-29
Update test under settings plugin hook

travishathaway:
### Description

xref: https://github.com/conda/conda-libmamba-solver/issues/715

In order for this test to be more robust, we need to add a mock to it to make sure it won't break when other plugins are installed like `conda-libmamba-solver`.

### Checklist - did you ...


dholth (approved):
LGTM

jezdez:
I'd suggest keeping the test, but removing the libmamba solver from the plugins in the tests instead. I don't think we should imply plugins to be part of conda's default test suite.

travishathaway:
> I'd suggest keeping the test, but removing the libmamba solver from the plugins in the tests instead. I don't think we should imply plugins to be part of conda's default test suite.

I'll take a stab at doing that. Probably just involves slapping a mock on the `PluginManager` class... 🤔

dholth:
> I'd suggest keeping the test, but removing the libmamba solver from the plugins in the tests instead. I don't think we should imply plugins to be part of conda's default test suite.

Some plugins like the conda-libmamba-solver plugin will be part of conda's default test suite...

travishathaway:
@jezdez,

Yep, that was super easy to add. Ready for review again.

PR #15273, opened 2025-09-29
Add fish shell dev environment script

danyeaw:
### Description

I was pairing with @ForgottenProgramme today and discussing dev environment setup. Using bash or zsh are great choices, but I would like to be able to set up an environment using fish as well. In the future it would also be nice to have a PowerShell version as well.

Tradeoffs of this is it is one more thing for us to maintain, however, the main differences to the bash is some syntax changes - so shouldn't be a lot of ongoing maintenance.

### Checklist - did you ...

### DIFF
--- a/dev/start.fish
+++ b/dev/start.fish
@@ -1,0 +1,385 @@
+#!/usr/bin/env fish
+# NOTE: This script should be sourced! The shebang is only here to help syntax highlighters.
+
+function cleanup
+    set -e _ARCH
+    set -e _BASEEXE
+    set -e _CHANNEL_NAME
+    set -e _CHOICE
+    set -e _CMD
+    set -e _DEFAULT_ARCH
+    set -e _DEFAULT_DEVENV
+    set -e _DEFAULT_DRYRUN
+    set -e _DEFAULT_INSTALLER
+    set -e _DEFAULT_MACH
+    set -e _DEFAULT_PYTHON
+    set -e _DEFAULT_UPDATE
+    set -e _DEVENV
+    set -e _DOWNLOAD_URL
+    set -e _DRYRUN
+    set -e _ENV
+    set -e _ENVEXE
+    set -e _INSTALLER
+    set -e _INSTALLER_DISPLAY
+    set -e _INSTALLER_FILE
+    set -e _INSTALLER_INITIAL
+    set -e _INSTALLER_TYPE
+    set -e _MACH
+    set -e _NAME
+    set -e _PYTHON
+    set -e _PYTHONEXE
+    set -e _SCRIPT_PATH
+    set -e _SHELL_TYPE
+    set -e _SRC
+    set -e _UPDATE
+    set -e _UPDATED
+end
+
+function updating
+    # check if explicitly updating or if 24 hrs since last update
+    test "$_UPDATE" = "true" && return 0
+    test -f "$_UPDATED" || return 0
+    set current_time (date +%s)
+    set file_time (date -r "$_UPDATED" +%s)
+    set diff_time (math $current_time - $file_time)
+    test $diff_time -lt 86400
+end
+
+cleanup
+
+# get source path
+set _SCRIPT_PATH (status filename)
+set _SRC (dirname (dirname $_SCRIPT_PATH))
+set _SRC (cd "$_SRC" 2>&1 > /dev/null; pwd -P)
+
+# define default values
+set _DEFAULT_PYTHON "3.10"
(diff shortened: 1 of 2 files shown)


dev/start.fish
```diff
@@ -0,0 +1,385 @@
...
+            echo "  -p, --python    VERSION    Python version for the env to activate. (default: $_DEFAULT_PYTHON)"
+            echo "  -i, --installer INSTALLER  Installer to use: miniconda or miniforge, can also be defined in ~/.condarc. (default: $_DEFAULT_INSTALLER)"
+            echo "  -m, --mach      MACHINE    Machine type for the env to activate. (default: $_DEFAULT_MACH)"
+            echo "  -a, --arch      ARCH       Architecture for the env to activate. (default: $_DEFAULT_ARCH)"
+            echo "  -u, --update               Force update packages. (default: $_DEFAULT_UPDATE, update every 24 hours)"
+            echo "  -d, --devenv    PATH       Path to base env install, can also be defined in ~/.condarc."
+            echo "                             Path is appended with machine hardware name, see --mach. (default: $_DEFAULT_DEVENV)"
+            echo "  -n, --dry-run              Display env to activate. (default: $_DEFAULT_DRYRUN)"
+            echo "  -h, --help                 Display this."
+            return 0
```
soapy1: I get an error when I source this script in a fish shell `3.x`. It looks like maybe this should be [`exit 0`](https://fishshell.com/docs/current/cmds/exit.html) instead. This is also compatible with fish 4.x. I think this change needs to happen for every `return` statement.

```suggestion
            exit 0
```

dev/start.fish
```diff
@@ -0,0 +1,385 @@
...
+echo "Initializing shell integration..."
+eval (env CONDA_AUTO_ACTIVATE=0 "$_ENVEXE" shell.fish hook) > /dev/null
+if not test $status -eq 0
+    echo "Error: failed to initialize shell integration" 1>&2
+    return 1
+end
+
+# activate env
+echo "Activating $_NAME..."
+if not conda activate "$_ENV" > /dev/null
```
soapy1: Hmmm, I get an error here about conda not being initialized properly
```
> source ./dev/start.fish
Choose conda installer:
  1) miniconda (default - Anaconda defaults channel)
  2) miniforge (conda-forge channel)

Enter choice [1]: 2
Deactivating 1 environment(s)...
Warning: failed to deactivate environment, continuing...
Updating devenv-3.10-miniforge...

Update shell scripts...
Initializing shell integration...
Activating devenv-3.10-miniforge...

CondaError: Run 'conda init' before 'conda activate'

Error: failed to activate devenv-3.10-miniforge
```

This is happens for fish 4.x and 3.x, without an existing `devenv/`

jezdez:
We're already struggling to maintain the POSIX shell script, and the Batchfile for Windows, I'm not in favor of adding another custom shell script like this but instead invest into getting away from the need for those scripts.

dev/start.fish
danyeaw: Thanks @soapy1, I fixed these return conditions!

danyeaw:
Hi @jezdez, ya, I can understand that - although I'm going to be sad if I have to swap shells to contribute to conda. I can commit to keeping it updated to match the bash/zsh script. What would be a step towards us getting rid of this, putting effort in to conda-spawn?

dev/start.fish
danyeaw: Thanks, this should be fixed hopefully now!

kenodegard:
Instead of adding more start scripts we should update the contribution docs explaining how to setup a dev environment without using the start scripts, it boils down to:

1. install miniconda/miniforge
2. create `devenv` environment: `conda create -n devenv --file tests/requirements.txt ...`
3. activate `devenv` environment as `base` (this step will differ for each shell)
4. install source as pip-editable: `pip --no-deps -e .`

dholth:
> Instead of adding more start scripts we should update the contribution docs explaining how to setup a dev environment without using the start scripts, it boils down to:
> 
>     1. install miniconda/miniforge
> 
>     2. create `devenv` environment: `conda create -n devenv --file tests/requirements.txt ...`
> 
>     3. activate `devenv` environment as `base` (this step will differ for each shell)
> 
>     4. install source as pip-editable: `pip --no-deps -e .`

Hopefully soon, 4. will be "conda pypi -e ."

danyeaw:
I like that idea to document the manual steps, and then continue optimize them. I'll try to take that on.

Issue #15205, opened 2025-09-12
`environment.yml` does not respect the `prefix` field


soapy1:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Consider an environment.yml file
```
# env.yml
prefix: /tmp/myenv
dependencies:
  - python
```

**Expected behaviour**
A user would expect running `conda env create --file env.yml` to:
* create a new environment at `/tmp/myenv`
* install python into that prefix

**Actual behaviour**
Conda does not respect `prefix` field:

```
(conda:base) $ conda env create --file env.yml

CondaEnvException: Unable to create environment
Please re-run this command with one of the following options:
* Provide an environment name via --name or -n
* Provide a path on disk via --prefix or -p
```

or, specifying a prefix in the command line
```
$ conda env create --file tests/env/support/simple.yml -p /tmp/other    [8:28:06]
Channels:
 - defaults
 - conda-forge
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: done

Downloading and Extracting Packages:

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate /tmp/other
#
# To deactivate an active environment, use
#
#     $ conda deactivate
```

### Conda Info
* Running on the main branch of conda
* Similar results for conda going back to `24.3.x`

PR #15276, opened 2025-09-29
Ensure the environment spec prefix field is respsected

soapy1:
### Description

fixes: https://github.com/conda/conda/issues/15205

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_env_create.py
+++ b/conda/cli/main_env_create.py
@@ -121,7 +121,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     # FIXME conda code currently requires args to have a name or prefix
     # don't overwrite name if it's given. gh-254
     if args.prefix is None and args.name is None:
-        if env.name is None:  # requirements.txt won't populate Environment.name
+        if env.name is None and env.prefix is None:  # requirements.txt won't populate Environment.name
             msg = dals(
                 """
                 Unable to create environment
@@ -132,6 +132,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             )
             raise CondaEnvException(msg)
         args.name = env.name
+        args.prefix = env.prefix
 
     prefix = determine_target_prefix(context, args)
     prefix_data = PrefixData(prefix)
--- a/tests/env/test_create.py
+++ b/tests/env/test_create.py
@@ -400,3 +400,50 @@ def test_create_env_custom_platform(
 
         assert config.is_file()
         assert f"subdir: {platform}" in config.read_text()
+
+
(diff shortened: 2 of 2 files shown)


soapy1:
pre-commit.ci autofix

ryanskeith (commented):
The new catch is great.  I wonder though if we need to check if the name and prefix are set correctly from the file.

soapy1:
> I  wonder though if we need to check if the name and prefix are set correctly from the file.

Like as part of the tests?

ryanskeith:
> > I  wonder though if we need to check if the name and prefix are set correctly from the file.
> 
> Like as part of the tests?

I was thinking more as part of the initialization.  I think conda uses the folder name as the environment name.  So if prefix is set to something like `~/miniconda/envs/bob` then the name of the environment is `bob`.  If a user then creates a environment.yaml to doesn't have the name and the path ending the same, the user would be confused when they try to activate the environment with the name.  They may have set the name as `bob` but if the prefix is `~/miniconda/envs/barb`.  The environment name is `barb`.

soapy1:
@ryanskeith good point, that is a confusing situation.

currently, conda will produce an error if a prefix and name are specified.
```
$ conda env create --file env.yml --name bo
b --prefix /tmp/bobby
                        [--environment-specifier {cep-24,environment.yml,explicit,requirements.txt}]
                        [--no-default-packages] [--json] [--console CONSOLE] [-v] [-q] [-d] [-y]
                        [--solver {classic,libmamba}] [--subdir SUBDIR]
conda env create: error: argument -p/--prefix: not allowed with argument -n/--name
```

This is true for the `conda create` command as well. How do you feel about raising a similar error if both the name and prefix is specified by the environment?

ryanskeith:
@soapy1, yes, I think something similar would be in order with the caveat that if they are in agreement it shouldn't raise an error.

soapy1:
pre-commit.ci autofix

conda/cli/main_env_create.py
```diff
@@ -131,7 +133,20 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+                f"environment spec '{args.file}' specifies both environment "
+                "'name' and 'prefix'. Please specify only one or the other."
+            )
+            raise CondaEnvException(msg)
+        # Set the args.name to the env name so that it taken into account
+        # when determining the target prefix
+        args.name = env.name or args.name
+        # If the prefix from the environment is the same as the one from the 
+        # context then we don't need to update the args.prefix
+        if env.prefix != context.target_prefix:
```
soapy1, line 148: I think this makes sense. But would appreciate it if someone would double check this.

PR #15242, opened 2025-09-18
`dev/start` script followup to #15163

kenodegard:
### Description

A few followup adjustments to the start scripts:

1. removes `_CHOICE` and instead reuses `_INSTALLER_TYPE` with "normalization"
2. reorder setting defaults, reading from condarc, and validating to streamline process
   - combines `_INSTALLER_TYPE` prompt error handling with `_INSTALLER_TYPE` validation
4. add note about how to silence the prompt by setting `installer_type` in `~/.condarc`
5. add exit in batch script if prompt is terminated (e.g., CTRL-C)

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #15279, opened 2025-09-30
Improve devenv start script in Windows

jezdez:
### Description

Keep `dev/start.bat` working on Windows by fixing the installer prompt. Defaulting to miniconda now uses valid `IF NOT ... == ...` checks, and we reset/error-check the prompt so a plain `Enter` works in PowerShell without exiting.

Refs #15242

### Checklist - did you ...


dev/start.bat
```diff
@@ -72,13 +72,21 @@
...
 @IF "%_INSTALLER_TYPE%"=="" @SET "_INSTALLER_TYPE=miniconda"
 @IF "%_INSTALLER_TYPE%"=="2" @SET "_INSTALLER_TYPE=miniforge"
-@IF "%_INSTALLER_TYPE%"!="miniconda" @IF "%_INSTALLER_TYPE%"!="miniforge" (
+@IF NOT "%_INSTALLER_TYPE_PROMPT%"=="0" (
+    @IF NOT "%_INSTALLER_TYPE_RAW%"=="" (
+        @ECHO Error: failed to read installer selection 1>&2
+        @EXIT /B 1
+    )
+)
+@IF /I NOT "%_INSTALLER_TYPE%"=="miniconda" @IF /I NOT "%_INSTALLER_TYPE%"=="miniforge" (
```
kenodegard: we dont do any case-insensitive matching elsewhere so no need to allow it here:

```suggestion
@IF NOT "%_INSTALLER_TYPE%"=="miniconda" @IF NOT "%_INSTALLER_TYPE%"=="miniforge" (
```

dev/start.bat
```diff
@@ -72,13 +72,21 @@
...
 @ECHO Note: This choice can be overridden by setting the 'installer_type' key in ~\.condarc.
 @ECHO.
+@CALL :RESET_ERRORLEVEL
 @SET /P "_INSTALLER_TYPE=Enter choice [1]: "
-@IF NOT %ErrorLevel%==0 @EXIT /B 1
+@SET "_INSTALLER_TYPE_PROMPT=%ErrorLevel%"
+@SET "_INSTALLER_TYPE_RAW=%_INSTALLER_TYPE%"
 :: normalize user input
 @IF "%_INSTALLER_TYPE%"=="1" @SET "_INSTALLER_TYPE=miniconda"
 @IF "%_INSTALLER_TYPE%"=="" @SET "_INSTALLER_TYPE=miniconda"
```
kenodegard: Since there is no way to distinguish between an `Enter` and `Ctrl-C` (both result in `%_INSTALLER_TYPE%==""` **and** `%ErrorLevel%==1`) we need to just pick one behavior:

```suggestion
:: terminated or empty prompt returns errorlevel 1, set default
@IF %ErrorLevel%==1 @SET "_INSTALLER_TYPE=miniconda"
@CALL :RESET_ERRORLEVEL
:: normalize user input
@IF "%_INSTALLER_TYPE%"=="1" @SET "_INSTALLER_TYPE=miniconda"
```

dev/start.bat
```diff
@@ -72,13 +72,21 @@
...
 @IF "%_INSTALLER_TYPE%"=="2" @SET "_INSTALLER_TYPE=miniforge"
-@IF "%_INSTALLER_TYPE%"!="miniconda" @IF "%_INSTALLER_TYPE%"!="miniforge" (
+@IF NOT "%_INSTALLER_TYPE_PROMPT%"=="0" (
+    @IF NOT "%_INSTALLER_TYPE_RAW%"=="" (
+        @ECHO Error: failed to read installer selection 1>&2
+        @EXIT /B 1
+    )
+)
```
kenodegard: Since there is no way to distinguish between an `Enter` and `Ctrl-C` (both result in `%_INSTALLER_TYPE%==""` and `%ErrorLevel%==1`) this extra check doesn't help us:

```suggestion
```

dev/start.bat
```diff
@@ -297,6 +307,9 @@
 @SET _UPDATED=
 @GOTO :EOF
```
kenodegard: Let's make the cleanup include the reset:

```suggestion
@GOTO :RESET_ERRORLEVEL
```

kenodegard (changes requested):
We should also update both `DEVENV_CONDARC` and `INSTALLER_TYPE_CONDARC` routines to always exit with 0 that way they don't leak non-zero `%ErrorLevel%`

jezdez:
@kenodegard this works as expected

dev/start.bat
```diff
@@ -287,6 +289,8 @@
 @SET _INSTALLER_FILE=
 @SET _INSTALLER_INITIAL=
 @SET _INSTALLER_TYPE=
+@SET _INSTALLER_TYPE_PROMPT=
+@SET _INSTALLER_TYPE_RAW=
```
kenodegard: Don't need these anymore:

```suggestion
```

PR #15281, opened 2025-09-30
Don't require a prefix in the environment model

soapy1:
### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -216,9 +216,6 @@ class Environment:
     Data model for a conda environment.
     """
 
-    #: Prefix the environment is installed into (required).
-    prefix: str
-
     #: The platform this environment may be installed on (required)
     platform: str
 
@@ -237,6 +234,9 @@ class Environment:
     #: Environment name
     name: str | None = None
 
+    #: Prefix the environment is installed into.
+    prefix: str | None = None
+
     #: User requested specs for this environment.
     requested_packages: list[MatchSpec] = field(default_factory=list)
 
@@ -248,10 +248,6 @@ class Environment:
     virtual_packages: list[PackageRecord] = field(default_factory=list)
 
     def __post_init__(self):
-        # an environment must have a name of prefix
-        if not self.prefix:
-            raise CondaValueError("'Environment' needs a 'prefix'.")
-
         # an environment must have a platform
         if not self.platform:
             raise CondaValueError("'Environment' needs a 'platform'.")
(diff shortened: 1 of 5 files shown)


conda/models/environment.py
```diff
@@ -248,10 +248,6 @@ class Environment:
     virtual_packages: list[PackageRecord] = field(default_factory=list)
 
     def __post_init__(self):
-        # an environment must have a name of prefix
-        if not self.prefix:
```
ryanskeith, line 251: Perhaps, here we could make a check if prefix is supplied and name is not, the name could be auto-populated?

ryanskeith (dismissed):
These look great.  I am not seeing any detriments to the other parts of the code for this change.

conda/models/environment.py
soapy1: I don't think that is always possible to do. For example if a `prefix` is `/tmp/thing`, then there is no `name` that makes sense to use. 

We could autopopulate the name if the prefix is in the enrionments path - then it will have a valid `name`. But, I think it would be better if the environment model was just a simple container for environment information. And the calling conda is responsible for this type of validation.

conda/models/environment.py
```diff
@@ -216,9 +216,6 @@ class Environment:
     Data model for a conda environment.
     """
 
-    #: Prefix the environment is installed into (required).
-    prefix: str
-
     #: The platform this environment may be installed on (required)
     platform: str
```
kenodegard, line 219: thoughts on making this model `@dataclass(kw_only=True)`?

the order of these args no longer make sense to allow them to be positional IMO

soapy1: Ya, that makes sense to me! These are ordered alphebetically.

Pushed up that change.

soapy1: oh, looks like `kw_only=True` is only supported in python 3.10+. So this causes a bunch of tests to fail (https://github.com/conda/conda/actions/runs/18881919438/job/53887106569?pr=15281). Since 3.9 is EOL, lets let conda drop 3.9 support before making this change to the dataclass.

kenodegard: just created https://github.com/conda/conda/pull/15363

soapy1: Awesome, pulled those changes in :+1:

jezdez (approved):

jezdez:
@kenodegard since you worked on the other PR that this might be an alternative, I'll wait for your review

kenodegard (approved):

PR #15318, opened 2025-10-13
add `override_virtual_package` to context

### LINKED ISSUE #15156: Implement condarc-based method for overriding virtual packages
### Checklist

### What is the idea?

Virtual packages are either implemented for all platforms or on a per-platform basis (by checking `conda.base.context.context.subdir`). Subsequently, almost all virtual packages offer an override environment key. They follow a standard naming convention, e.g., `CONDA_OVERRIDE_CUDA`, but this is up to each plugin to follow (perhaps an opportunity to standardize?).

After checking for the override key, the actual virtual package detection logic is run.

What is missing is a standardized way to override these virtual packages through the existing context object, particularly by leveraging the condarc YAML configuration.

### Why is this needed?

This is necessary for exporting cross-platform lockfiles. When generating lockfiles for environments that need to work across different platforms, we currently don't have a clean way to specify virtual package requirements consistently. Environment variables work for manual operations but are insufficient for automated workflows and cross-platform compatibility.

### What should happen?

- [x] 1. Extend the conda context object to support virtual package overrides via condarc
- [x] 2. Create a unified interface that works consistently across all virtual packages
- [x] 3. Standardize the naming convention for all virtual package overrides (?)
- [ ] 5. Document the approach thoroughly, including examples for common use cases

### Additional Context

- Virtual packages like `__cuda`, `__archspec`, and others are critical for platform-specific dependencies
- Current workarounds involve setting environment variables before running conda operations, which is error-prone in automated environments
- This issue affects CI/CD pipelines, container builds, and cross-platform development workflows
- The context object in conda already aggregates configuration from condarc files, environment variables, and CLI arguments, so we can leverage this existing mechanism

### Proposed Implementation
Extend the condarc YAML format to support virtual package configuration:

```yaml
# In .condarc
virtual_packages:
  __cuda: "11.8"
  __archspec: "x86_64"
  __cudnn: "8.1"
  __mkl: "2023.1.0"
  # Other virtual packages as needed
```

The conda context object would:

1. Read these settings from condarc files
2. Apply them during package resolution and environment creation
3. Respect precedence rules (CLI args > environment variables > condarc settings)

This approach offers several advantages:

- Uses the existing context object in conda
- Configuration can be version-controlled
- Easy to share across team members
- Can be included in project-specific .condarc files
- Works seamlessly with existing conda workflows
- Provides a consistent configuration method across different environments

ForgottenProgramme:
### Description
Closes #15156 

- [x] Add `override_virtual_package` to context. 
- [x] Check for this context variable in `to_virtual_package()`.
- [x] Add tests

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -1091,7 +1091,7 @@ class Context(Configuration):
             return logging.WARNING  # 30
 
     @property
-    def override_virtual_packages(self) -> dict:
+    def override_virtual_packages(self) -> dict[str, str | None]:
         """Remove any dunders in the virtual_package name keys"""
         return {
             name.removeprefix("__"): value


soapy1:
After reading the description of the intended changes in https://github.com/conda/conda/issues/15156, I expect the following to behaviour: 

If I set the following condarc
```
virtual_packages:
  __cuda: "11.8"
```

Then conda should override the detected version of cuda to the set `11.8`. Note, my machine does not have cuda installed, so normally when I run a `conda info`, it reports `__cuda=0=0`. 

However, when I run a `conda info` with this condarc config, I see that my cuda version is not changed.

```
$ conda info | grep __cuda                                            
                          __cuda=0=0
```

Note, that if I update my condarc to 
```
_override_virtual_packages:
  __cuda: "11.8"
```

Then I get what I expect to be the right result.
```
$ conda info | grep __cuda                                            
                          __cuda=11.8=0
```

So, I think there is an error in how the context alias is setup? Could we also add some integration type tests for this?

soapy1 (dismissed):
Looks great! 

I did some QA on this as well and can confirm that the virtual package override is working and respecting the right levels of config.

Normally when I run a conda info I get the result
```
$ conda info | grep cuda
                          __cuda=0=0
```

If I update my condarc to
```
virtual_packages:
  __cuda: "11.8"
```
I get
```
$ conda info | grep cuda                
                          __cuda=11.8=0
```

With the same condarc, I can override the cuda version
```
$ CONDA_OVERRIDE_CUDA=12.0 conda info  | grep cuda                 
                          __cuda=12.0=0
```

I think it would still be good to get one more reviewer who is more familar with this workstream to review this.

conda/base/context.py
```diff
@@ -1085,6 +1090,14 @@ def log_level(self) -> int:
         else:
             return logging.WARNING  # 30
 
+    @property
+    def override_virtual_packages(self) -> dict:
+        """Remove any dunders in the virtual_package name keys"""
+        return {
+            k[2:] if k.startswith("__") else k: v
+            for k, v in self._override_virtual_packages.items()
```
kenodegard: We want to avoid single-char variables and we can use https://docs.python.org/3/library/stdtypes.html#str.removeprefix:

```suggestion
            name.removeprefix("__"): value
            for name, value in self._override_virtual_packages.items()
```

conda/base/context.py
```diff
@@ -2020,6 +2034,11 @@ def description_map(self) -> dict[str, str]:
                 Defaults to "{DEFAULT_CONSOLE_REPORTER_BACKEND}".
                 """
             ),
+            override_virtual_packages=dals(
+                """
+                Set virtual_packages to override values.
```
kenodegard:
```suggestion
                Set override values for virtual packages.
```

conda/plugins/types.py
```diff
@@ -124,14 +124,21 @@ class CondaVirtualPackage(CondaPlugin):
...
         # Overrides always yield a concrete value (string, NULL, or None),
         # so after this step, version/build will no longer be callables if they were overridden.
         if self.override_entity:
-            override_value = os.getenv(f"{APP_NAME}_OVERRIDE_{self.name}".upper())
+            override_value = os.getenv(
+                f"{APP_NAME}_OVERRIDE_{self.name}".upper()
+            )  # higher precedence
+            # check context only if `CONDA_OVERRIDE_<VIRTUAL_PACKAGE>` is not set.
+            if override_value is None and context.override_virtual_packages:
+                override_value = context.override_virtual_packages.get(f"{self.name}")
```
kenodegard:
```suggestion
            # environment variable has highest precedence
            override_value = os.getenv(f"{APP_NAME}_OVERRIDE_{self.name}".upper())
            # fallback to context
            if override_value is None:
                override_value = context.override_virtual_packages.get(self.name)
```

news/15318-add-virtual-packages-to-condarc
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a new `override_virtual_packages` (with alias `virtual_packages`) key to `condarc`.
```
kenodegard:
```suggestion
* Add a new `override_virtual_packages` (with alias `virtual_packages`) key to `condarc`. (#15156 via #15318)
```

kenodegard (changes requested):

conda/base/context.py
```diff
@@ -1085,6 +1090,14 @@ def log_level(self) -> int:
         else:
             return logging.WARNING  # 30
 
+    @property
+    def override_virtual_packages(self) -> dict:
```
kenodegard:
```suggestion
    def override_virtual_packages(self) -> dict[str, str | None]:
```

conda/plugins/types.py
kenodegard: the change here is reworking the trailing comment for compactness and greater readability

we should move trailing comments before the line it applies to if it causes the code to be reformatted across multiple lines

IOW we want code to look like:

```
some code here  # comment
```

or

```
# comment
some code here
```

not

```
(
    some code here
)  # comment
```

PR #15309, opened 2025-10-09
Fix `conda.models.environment.Environment` grouping of specs

kenodegard:
### Description

`Environment.from_cli` incorrectly categorizes an explicit URL as a `MatchSpec`.

Closes #15306 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #15320, opened 2025-10-13
perf: conda env list

hoxbro:
### Description

Noticed that `conda env list` took more time than I would expect it to. This PR waits for filling out `_info_dict` until it is needed. 

image

Will let CI run to see if this breaks anything. 

### Checklist - did you ...

### DIFF
--- a/news/15320-perf-conda-env-list
+++ b/news/15320-perf-conda-env-list
@@ -1,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Performance improvements for the `conda env list` command, which now defers
+  retrieving environment information until necessary (#15320)


travishathaway:
@conda-bot check

soapy1 (dismissed):
Neat improvement!

Playing around with this some, it looks like there is a significant speed improvement for just changing the `_info_dict` and `_info_dict["envs"]` into `@properties` and reorganizing the `InfoRenderer` initialization. I think it's safe to make these `@cached_properties` 👍 

This approach also provides some improvements for other `conda info` options. For example `conda info --base`

```
(pr/15320) $ hyperfine 'python -m conda info --base' --warmup 10                                 
Benchmark 1: python -m conda info --base
  Time (mean ± σ):     485.8 ms ±  86.0 ms    [User: 436.7 ms, System: 52.6 ms]
  Range (min … max):   388.1 ms … 665.1 ms    10 runs


 (main) $ hyperfine 'python -m conda info --base' --warmup 10                           
Benchmark 1: python -m conda info --base
  Time (mean ± σ):     685.2 ms ±  14.6 ms    [User: 599.7 ms, System: 93.4 ms]
  Range (min … max):   660.7 ms … 702.8 ms    10 runs
```

It would be good to get one more review from @conda/conda-maintainers before merging.

soapy1:
@hoxbro could you also add an a `news` entry for this change? For example, like the one included in this change https://github.com/conda/conda/pull/15309/files#diff-f57f719ee79204d66dbb0de020b028f0617971778c85476d2f43398a959c26d4. Thanks!

Issue #15268, opened 2025-09-26
Given conda a null value to channels causes Anaconda default channels to be used.


ryanskeith:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Having a .condarc file looking like this:

```
channels:
  -
```

or this:

```
channels:
  - null
```
will cause the following output on conda 25.9.0

```
conda install black --dry-run
Channels:
 - <unknown>
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /Users/rkeith/miniconda3/envs/c25.9

  added / updated specs:
    - black


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    black-25.1.0               |  py313hca03da5_0         533 KB
    click-8.2.1                |  py313hca03da5_0         337 KB
    mypy_extensions-1.0.0      |  py313hca03da5_0          13 KB
    pathspec-0.12.1            |  py313hca03da5_0          62 KB
    ------------------------------------------------------------
                                           Total:         945 KB

The following NEW packages will be INSTALLED:

  black              pkgs/main/osx-arm64::black-25.1.0-py313hca03da5_0
  click              pkgs/main/osx-arm64::click-8.2.1-py313hca03da5_0
  mypy_extensions    pkgs/main/osx-arm64::mypy_extensions-1.0.0-py313hca03da5_0
  pathspec           pkgs/main/osx-arm64::pathspec-0.12.1-py313hca03da5_0



DryRunExit: Dry run. Exiting.
```

It seems that if there is no validation for None values.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

This was found accidentally while testing the 25.9.x release of conda.  Not a show stopping bug but something that should be addressed.

PR #15323, opened 2025-10-13
Filter out 'None' channels

soapy1:
### Description

Fixes https://github.com/conda/conda/issues/15268

This PR will filter out `None` channels during channel validations. 

For example, for condarc's that look like
```
channels:
  -
```
or this:
```
channels:
  - null
```

**previously** with this configuration conda would install packages from the `main` channel. 

**with this change** conda will filter out `None` channels. So, a conda with this configuration will not have any valid channels and produce a `conda create --prefix /tmp/test black  --dry-run `. For example:

```

$ conda create --prefix /tmp/test black  --dry-run

NoChannelsConfiguredError: No channels are configured. Conda requires at least one channel to search for packages.
. . .

```

Note, conda still respects channel config sourced from environment variables and the cli. 

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -2262,7 +2262,11 @@ def validate_channels(channels: Iterator[str]) -> tuple[str, ...]:
                 if allowlist and url not in allowlist:
                     raise ChannelNotAllowed(channel)
 
-    return tuple(channel for channel in dict.fromkeys(channels) if Channel(channel).canonical_name != UNKNOWN_CHANNEL)
+    return tuple(
+        channel
+        for channel in dict.fromkeys(channels)
+        if Channel(channel).canonical_name != UNKNOWN_CHANNEL
+    )
 
 
 @deprecated(
--- a/tests/base/test_context.py
+++ b/tests/base/test_context.py
@@ -47,8 +47,9 @@ from conda.models.match_spec import MatchSpec
 from conda.utils import on_win
 
 if TYPE_CHECKING:
-    from pytest import MonkeyPatch
     from collections.abc import Iterator
+
+    from pytest import MonkeyPatch
 
     from conda.testing import PathFactoryFixture
     from conda.testing.fixtures import CondaCLIFixture, TmpEnvFixture
@@ -760,6 +761,7 @@ def test_check_allowlist_and_denylist(monkeypatch: MonkeyPatch):
     validate_channels(("defaults",))
     validate_channels((DEFAULT_CHANNELS[0], DEFAULT_CHANNELS[1]))
 
+
 @pytest.mark.parametrize(
     "channels,expected_channels",
     [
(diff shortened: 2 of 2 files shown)


conda/base/context.py
```diff
@@ -2241,7 +2242,7 @@ def validate_channels(channels: Iterator[str]) -> tuple[str, ...]:
                 if allowlist and url not in allowlist:
                     raise ChannelNotAllowed(channel)
 
-    return tuple(dict.fromkeys(channels))
+    return tuple(channel for channel in dict.fromkeys(channels) if Channel(channel).canonical_name != UNKNOWN_CHANNEL)
```
soapy1: Should conda produce an error here about not having any configured channels if there are no channels?

tests/test_create.py
```diff
@@ -632,6 +632,23 @@ def test_noarch_generic_package(test_recipes_channel: Path, tmp_env: TmpEnvFixtu
...
+    path_factory: PathFactoryFixture,
+) -> None:
+    conda_cli(
+        "create",
+        f"--prefix={path_factory()}",
+        "zlib",
+        "--yes",
+        "--channel",
+        None,
+        "--override-channels"
```
soapy1: Using `--override-channels` and specifying `--channels None` is another way to force conda to have a `None` channel config. I don't think there is another way to force this case in tests since there is a global `.condarc`

conda/base/context.py
jezdez: Yeah, that's what this method is well suited for, I'd double-check if that's the only place where channels are loaded/validated again, to make sure we're not raising an error, if channels could also be configured somehow else. But I think the context should take all of that already into account, right?

jezdez: So maybe a new `ChannelNotProvided` exception class?

jezdez (changes requested):

conda/base/context.py
soapy1: Looks like this `validate_channels` is also used in this cli helper module https://github.com/conda/conda/blob/main/conda/cli/helpers.py#L71.  In that context, this function is used to ensure that there are no requested packages are coming from channels on the deny list. In this context, it doesn't make sense to throw an error when there are no channels provided. 

Further, this doc string specifically mentions that this validates that all channels based on the allow/deny list. So, I think it makes sense to move this exception raising to the `channels` property. Defintely open to other suggestions too!

soapy1: This PR https://github.com/conda/conda/pull/15345 introduces raising an error about no channels being configured in the solver layer. Now that it has been merged, lets not check for the same error twice. The responsibliity of this function should be to just strip out any input that is not a valid channel. The `test_no_channels` test below demonstrates that a helpful error message is created when no valid channels are provided by the user.

soapy1:
pre-commit.ci autofix

jezdez (approved):

Issue #14308, opened 2024-10-11
Provide information about time of environment creation, last update and last activation


aovasylenko:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Extend `conda env list` with information about time of environment creation, last update and last activation

### Why is this needed?

As a user, i want to understand better, if there are environments in my system, which are not used any more. 
`conda env list` provides list of environments, known to conda, but it doesn't provide any hints, if environment is used. 
Providing information about creation time, last update and last activation could be really helpful at this point.

### What should happen?

`conda list env` and `conda list env --json` should include information about creation, update and last activation
(Feel free to adjust according to other constraints)

dashagurova:
I feel like this request is thematically related to [conda/conda#11480](https://github.com/conda/conda/issues/11480): making it easier for users to track and manage multiple environments. I support both ideas, as I often forget the purpose of environments I created just a week ago...

travishathaway:
I believe this type of feature would be better implemented as a plugin.

jaimergp:
This would need two sides:

- how to store standardized metadata about it; possibly part of https://github.com/conda/ceps/issues/67. Last modification of environment could come from `stat` of the `conda-meta/history`; the rest requires new metadata fields.
- how to export this information in the UI; the best way would be to open up a plugin hook for `conda info --envs`.
 
A workaround for now is to implement a new subcommand that implements the prototype and show a nice demo.

travishathaway:
> any pros/cons for plugin?

Pros:
- Full flexibility to implement it in any way you want (e.g. experiment with other CLI formatting libraries like rich)
- Does not add complexity to the current code base
- Conda env list may need some refactor before being able to reasonably accommodate such a feature

Cons:
- Users are required to install an additional plugin to enhance the ability of conda
- This may be considered part of the core functionality and we may not want to offload this to a plugin

With that being said, I think this could initially be developed as a plugin and then perhaps be considered as a replacement for `conda env list`. I say this because I believe a re-write of `conda env list` and by extension `conda list` might be a good idea and a plugin as a prototype of this would be a great starting point for this future work.

I hope that makes sense.

jaimergp:
Opened https://github.com/conda/conda/pull/15329 to add the necessary API. Unfortunately, the `info --envs` logic is very simple (only a list of prefixes), which is then enriched at print time. 

[...]

PR #15329, opened 2025-10-16
Add `.created` and `.last_modified` properties to `PrefixData`

jaimergp:
### Description

Part of https://github.com/conda/conda/issues/14308

### Checklist - did you ...

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -358,9 +358,14 @@ class PrefixData(metaclass=PrefixDataType):
     def created(self) -> datetime | None:
         """
         Returns the time when the environment was created, as evidenced by the `conda-meta`
-        directory creation time (if available). Falls back to conda-meta/.creation-timestamp.
-        If the environment does not exist or the creation time cannot be identified, returns
-        None.
+        directory creation time (if available). Falls back to `conda-meta/created_at`.
+
+        This may return None in the following cases:
+
+        - If the environment doesn't exist.
+        - If the creation time cannot be determined. This can happen for existing environments
+          created with conda versions before 25.11 in Linux systems, where the 'birthtime' metadata
+          is only available in kernels 4.11+, and not implemented in Python as of Python 3.14.
         """
         try:
             stat = (self.prefix_path / "conda-meta").stat()


conda/core/prefix_data.py
```diff
@@ -348,6 +350,43 @@ def validate_name(self, allow_base: bool = False) -> None:
...
+        Returns the time when the environment was last modified, as evidenced by the
+        `conda-meta/history` file modification time. If the environment does not exist, returns
+        None.
+        """
+        try:
+            stat = self._magic_file.stat()
+        except FileNotFoundError:
+            return None
+        else:
+            return datetime.fromtimestamp(stat.st_mtime, tz=timezone.utc)
```
dholth, line 412: There's a nanosecond one now also https://docs.python.org/3/library/os.html#os.stat_result.st_mtime_ns

jaimergp: Might not be present on some platforms and I don't think we need that type of accuracy here, so I decided not to use it.

conda/core/prefix_data.py
```diff
@@ -348,6 +351,59 @@ def validate_name(self, allow_base: bool = False) -> None:
...
+        except FileNotFoundError:
+            return None
+        else:
+            # On Windows, ctime represents creation time. On other platforms, ctime gives
+            # last metadata change; creation time comes from birthtime.
+            # Birthtime was introduced for Windows in Py312, hence the fallback below.
+            # On Linux, birthtime is only available in kernel 4.11+ via statx and won't be
+            # available in Python until 3.15. The fallback to ctime would be wrong
+            # because in this platform ctime is last metadata change (not creation time!).
+            if on_mac or on_win:
```
soapy1, line 386: So, on linux the creation time for any environments created before this change as added will be `None`. Is there a good place to document that caveat of how this works?

conda/core/prefix_data.py
```diff
@@ -108,6 +110,7 @@ class PrefixData(metaclass=PrefixDataType):
     """
 
     _cache_: dict[tuple[Path, bool | None], PrefixData] = {}
+    CREATION_TIMESTAMP_FILE = "conda-meta/.creation-timestamp"
```
jezdez: We have a bunch of other files in conda-meta that are not prefixed with a dot, let's just use something like

```suggestion
    CREATION_TIMESTAMP_FILE = "conda-meta/created"
```

or 

```suggestion
    CREATION_TIMESTAMP_FILE = "conda-meta/created_at"
```

Oh, and let's move the path into the `constants` module?

conda/core/prefix_data.py
```diff
@@ -348,6 +351,59 @@ def validate_name(self, allow_base: bool = False) -> None:
...
+                return None
+        except FileNotFoundError:
+            return None
+        else:
+            # On Windows, ctime represents creation time. On other platforms, ctime gives
+            # last metadata change; creation time comes from birthtime.
+            # Birthtime was introduced for Windows in Py312, hence the fallback below.
+            # On Linux, birthtime is only available in kernel 4.11+ via statx and won't be
+            # available in Python until 3.15. The fallback to ctime would be wrong
+            # because in this platform ctime is last metadata change (not creation time!).
```
jezdez, line 384: TIL, can we link to a source of this information for later reference?

conda/core/prefix_data.py
```diff
@@ -348,6 +351,59 @@ def validate_name(self, allow_base: bool = False) -> None:
...
+            if on_mac or on_win:
+                creation_time = getattr(stat, "st_birthtime", stat.st_ctime)
+            else:
+                try:
+                    creation_time = stat.st_birthtime
+                except AttributeError:
+                    # Fallback to magic file conda-meta/.creation-timestamp
```
jezdez, line 392: Needs filename change as mentioned above

conda/core/prefix_data.py
```diff
@@ -621,6 +677,18 @@ def unset_environment_env_vars(
...
+        """
+        Writes a .creation-time file in conda-meta with the current timestamp, meant
+        to be used by .created property as a fallback.
+        """
+        ts = (
+            self.created or self.last_modified or datetime.now(timezone.utc).timestamp()
+        )
+        tsfile = self.prefix_path / self.CREATION_TIMESTAMP_FILE
+        tsfile.parent.mkdir(parents=True, exist_ok=True)
+        tsfile.write_text(str(ts))
```
jezdez:
```suggestion
        timestamp = (
            self.created or self.last_modified or datetime.now(timezone.utc).timestamp()
        )
        timestamp_file = self.prefix_path / self.CREATION_TIMESTAMP_FILE
        timestamp_file.parent.mkdir(parents=True, exist_ok=True)
        timestamp_file.write_text(str(timestamp))
```

conda/core/prefix_data.py
jaimergp: Done!

jaimergp:
Let's merge https://github.com/conda/conda/pull/15330 first so we can add these new fields to the new `envs_details` dict

jaimergp:
Hm, actually let's do it the other way around. Merge here and then add them in https://github.com/conda/conda/pull/15330

Issue #15079, opened 2025-08-06
tests fail using Python 3.14 with `TypeError: _WorkItem.__init__() takes 3 positional arguments but 5 were given`


opoplawski:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I'm working on updating the conda package in Fedora to 25.7.0 and getting test failures like this:
```
___________________ TestLibMambaSolver.test_nonexistent_deps ___________________
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/_pytest/runner.py", line 341, in from_call
    result: TResult | None = func()
                             ~~~~^^
  File "/usr/lib/python3.14/site-packages/_pytest/runner.py", line 242, in <lambda>
    lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
            ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_hooks.py", line 512, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 167, in _multicall
    raise exception
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 139, in _multicall
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/_pytest/threadexception.py", line 92, in pytest_runtest_call
    yield from thread_exception_runtest_hook()
  File "/usr/lib/python3.14/site-packages/_pytest/threadexception.py", line 68, in thread_exception_runtest_hook
    yield
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 139, in _multicall
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
[...]

Issue #15088, opened 2025-08-07
Test failure with Python 3.14 - `AttributeError: 'UrlTest' object has no attribute '__annotations__'. Did you mean: '__annotate_func__'?`


opoplawski:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Building conda 25.7.0 with Python 3.14, test fail with:
```
________ test_urlparse[192.168.1.1:8080/path/to/resource-exp_url_obj0] _________
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/_pytest/runner.py", line 341, in from_call
    result: TResult | None = func()
                             ~~~~^^
  File "/usr/lib/python3.14/site-packages/_pytest/runner.py", line 242, in <lambda>
    lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
            ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_hooks.py", line 512, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 167, in _multicall
    raise exception
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 139, in _multicall
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/_pytest/threadexception.py", line 92, in pytest_runtest_call
    yield from thread_exception_runtest_hook()
  File "/usr/lib/python3.14/site-packages/_pytest/threadexception.py", line 68, in thread_exception_runtest_hook
    yield
  File "/usr/lib/python3.14/site-packages/pluggy/_callers.py", line 139, in _multicall
    teardown.throw(exception)
[...]

PR #15340, opened 2025-10-20
Fix Python 3.14 compatibility bugs in `ThreadPoolExecutor` implementation

travishathaway:
### Description

fixes: #15079, #15088

I didn't add any test coverage because we are already testing our custom implementation of `ThreadPoolExecutor` (`ThreadLimitedThreadPoolExecutor`) quite a bit during our existing integration tests.

### Checklist - did you ...


tests/common/test_url.py
```diff
@@ -116,7 +116,7 @@ def test_urlparse(test_url_str, exp_url_obj):
     """Tests a variety of different use cases for `conda.common.url.urlparse`."""
     answer = urlparse(test_url_str)
 
-    for attr in exp_url_obj.__annotations__.keys():
+    for attr in exp_url_obj._asdict().keys():
```
jaimergp: TIL that in `namedtuples`, the "leading underscore means private" convention [doesn't apply](https://docs.python.org/3/library/collections.html#collections.namedtuple):

> In addition to the methods inherited from tuples, named tuples support three additional methods and two attributes. To prevent conflicts with field names, the method and attribute names start with an underscore.

travishathaway: I learned that last week too! 😄

news/15340-python-314-fixes
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Several bug fixes related to Python 3.14 compatability (#15079 and #15088 via #15340)
```
agriyakhetarpal:
```suggestion
* Several bug fixes related to Python 3.14 compatibility (#15079 and #15088 via #15340)
```

conda/common/io.py
```diff
@@ -428,40 +427,18 @@ class ThreadLimitedThreadPoolExecutor(ThreadPoolExecutor):
...
+            return super()._adjust_thread_count()
+        except RuntimeError:
+            # RuntimeError: can't start new thread
+            # See https://github.com/conda/conda/issues/6624
+            if len(self._threads) > 0:
+                # It's ok to not be able to start new threads if we already have at least
+                # one thread alive.
+                pass
+            else:
+                raise
```
agriyakhetarpal, line 441: We could consider adding a test that verifies that when the thread limit is reached, we degrade gracefully – because `_adjust_thread_count` is still private API, even though it's better than `_WorkItem`. Perhaps something like this (please feel free to expand it):

```python
def test_thread_limited_executor_handles_thread_limit():
    executor = ThreadLimitedThreadPoolExecutor(max_workers=2)
    futures = [executor.submit(time.sleep, 0.1) for _ in range(10)]

    # should run without incurring a RuntimeError
    results = [f.result() for f in futures]
    assert len(results) == 10
```

or if you are confident with this (as you mentioned in the description, that is), please feel free to ignore this suggestion. :D

agriyakhetarpal (approved):
Hi @travishathaway, this looks great. Thank you!

conda/common/io.py
travishathaway: That's a good suggestion. I will add this test just so it's clear to future maintainers why this class exists in the first place.

dholth: Use monkeypatch/mock to have https://github.com/python/cpython/blob/v3.14.0/Lib/concurrent/futures/thread.py#L219 raise the tested exception on the second or third call, then run `pytest -k <named test> --cov` to verify that the error handling code runs.

news/15340-python-314-fixes
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Several bug fixes related to Python 3.14 compatibility (#15079 and #15088 via #15340)
```
jezdez:
```suggestion
* Fix several bug related to Python 3.14 compatibility. (#15079, #15088 via #15340)
```

news/15340-python-314-fixes
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix several bug related to Python 3.14 compatibility. (#15079, #15088 via #15340)
```
jezdez:
```suggestion
* Fix several bugs related to Python 3.14 compatibility. (#15079, #15088 via #15340)
```

tests/common/test_io.py
```diff
@@ -87,3 +89,17 @@ def test_attach_stderr_handler():
...
     assert debug_message in c.stderr
+
+
+def test_thread_limited_executor_handles_thread_limit():
+    """
+    Ensure our custom ThreadLimitedThreadPoolExecutor class does what it is
+    intended to do, which is not raise a RuntimeError if the max workers limit
+    is reached.
+    """
+    executor = ThreadLimitedThreadPoolExecutor(max_workers=2)
```
kenodegard: I replaced `ThreadLimitedThreadPoolExecutor` with `ThreadPoolExecutor` and this test still passes

kenodegard: can we parameterize this test so we can assert that the test will fail for `concurrent.futures.ThreadPoolExecutor` but pass for `conda.common.io.ThreadLimitedThreadPoolExecutor`?

tests/common/test_url.py
kenodegard: theres an attribute for that:

```suggestion
    for attr in exp_url_obj._fields:
```

travishathaway: The previous code was subconsciously affecting me 😂

tests/common/test_io.py
```diff
@@ -87,3 +92,40 @@ def test_attach_stderr_handler():
...
+            "concurrent.futures.ThreadPoolExecutor._adjust_thread_count",
+            side_effect=RuntimeError,
+        )
+
+    with thread_class(max_workers=2) as executor:
+        if should_fail:
+            with pytest.raises(RuntimeError):
+                _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
+        else:
+            _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
```
kenodegard: We use `contextlib.nullcontext` to avoid duplicate code:

```suggestion
        with pytest.raises(RuntimeError) if should_fail else nullcontext():
            assert [executor.submit(time.sleep, 0.1) for _ in range(10)]
```

conda/common/io.py
```diff
@@ -428,40 +427,18 @@ class ThreadLimitedThreadPoolExecutor(ThreadPoolExecutor):
...
+    def _adjust_thread_count(self):
+        try:
+            return super()._adjust_thread_count()
+        except RuntimeError:
+            # RuntimeError: can't start new thread
+            # See https://github.com/conda/conda/issues/6624
+            if len(self._threads) > 0:
+                # It's ok to not be able to start new threads if we already have at least
+                # one thread alive.
+                pass
```
dholth, line 439: Is this line of code executed by the tests

tests/common/test_io.py
```diff
@@ -87,3 +92,40 @@ def test_attach_stderr_handler():
...
+    Some historical background can be found here:
+        - https://github.com/conda/conda/pull/6653
+        - https://github.com/conda/conda/issues/7040
+    """
+    jobs = 10
+
+    if should_fail:
+        mocker.patch(
+            "concurrent.futures.ThreadPoolExecutor._adjust_thread_count",
+            side_effect=RuntimeError,
```
dholth: Do we need an _adjust_thread_count that raises RuntimeError if the requested number is > 2?

conda/common/io.py
travishathaway: My coverage report is showing that this is being covered although the if statement above is only partially covered.

Regardless, this is a step in the right direction considering the coverage of our previous version of this class:

- https://app.codecov.io/gh/conda/conda/blob/main/conda%2Fcommon%2Fio.py#L458

tests/common/test_io.py
travishathaway: I've removed setting `max_workers` so now this will be set to the number of processors available on the machine per the Python documentation.

Issue #15292, opened 2025-10-03
TypeError: conda.activate._Activator.get_export_unset_vars() got multiple values for keyword argument 'PATH'


marcmalade:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Hello, 
i've installed several virtual envs (comfyui, forgeui) with miniconda. 
In the past i had facefusion too, without any problem, runnning on its own virtual env.

I've reset my PC with windows 11 cuda 12.9, re-installed comfyui and forgeui without any problem but Facefusion isn't recognizing the virtual env i just created.
I've installed everything (torch==2.8.0+cu129, cudnn9.10.0, tensorrt==10.12.0.6 ) and the requirements.txt

When in try to activate facefusion virtual env it fails.
I know the problem comes from :    

 TypeError: conda.activate._Activator.get_export_unset_vars() got multiple values for keyword argument 'PATH'
`$ C:\ProgramData\miniconda3\Scripts\conda-script.py shell.cmd.exe activate facefusion`

But don't know which variable entry should i delete, or how to clean the facefusion env install

i 've tried from scratch several times, always the same problem

thank you for your help and knowledge ! 

### Conda Info

```shell
C:\Windows\System32>conda info

     active environment : None
            shell level : 0
       user config file : C:\Users\ARTHUR\.condarc
 populated config files : C:\ProgramData\miniconda3\.condarc
          conda version : 25.7.0
    conda-build version : not installed
         python version : 3.13.5.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=skylake
                          __conda=25.7.0=0
                          __cuda=13.0=0
                          __win=10.0.26100=0
       base environment : C:\ProgramData\miniconda3  (writable)
      conda av data dir : C:\ProgramData\miniconda3\etc\conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\ProgramData\miniconda3\pkgs
                          C:\Users\ARTHUR\.conda\pkgs
                          C:\Users\ARTHUR\AppData\Local\conda\conda\pkgs
[...]

PR #15342, opened 2025-10-20
Don't modify reserved environment variables when (de)/activating

soapy1:
### Description

Pt. 1 to address  https://github.com/conda/conda/issues/15292

Users should not be able to change their `PATH` environment variable through their environment config.

This change will ignore these reserved variables (currently just `PATH`). And, will print a warning to the user if a reserved environment variable is detected.

For example, to test this out:
```
// create a new env
$ conda create -n myenv

// set the `PATH` variable for the env
$ conda env config vars set PATH="notsogood" -n myenv

// activate the environment - notice 
// 1. no error  - on the main branch of conda this will produce an error 
// 2. warning message
 $ conda activate myenv                               
WARNING: the configured environment variable(s) for prefix '/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniforge/envs/myenv' are reserved and will be ignored: PATH.

Remove the invalid configuration with `conda env config vars unset -p /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniforge/envs/myenv PATH`.


// deactivate the environment - notice
// 1. no error  - on the main branch of conda this will produce an error 
// 2. warning message
$ conda deactivate                             
WARNING: the configured environment variable(s) for prefix '/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniforge/envs/myenv' are reserved and will be ignored: PATH.

Remove the invalid configuration with `conda env config vars unset -p /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniforge/envs/myenv PATH`.


```

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -820,10 +820,10 @@ class _Activator(metaclass=abc.ABCMeta):
         if collect_reserved_vars:
             print_reserved_vars = ", ".join(collect_reserved_vars)
             print(
-                f"WARNING: the configured environment variable(s) are reserved and "
-                f"will be ignored: {print_reserved_vars}.\n\n"
+                f"WARNING: the configured environment variable(s) for prefix '{prefix}' "
+                f"are reserved and will be ignored: {print_reserved_vars}.\n\n"
                 f"Remove the invalid configuration with `conda env config vars unset "
-                f"{' '.join(collect_reserved_vars)}`.\n",
+                f"-p {prefix} {' '.join(collect_reserved_vars)}`.\n",
                 file=sys.stderr,
             )
         return env_vars


soapy1:
pre-commit.ci autofix

news/15342-ignore-PATH-env-var
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Ignore and warn user if the `PATH` environment variable is set in the environment config.  Conda should not enable modifying the `PATH` to any arbitrary value.
```
jezdez:
```suggestion
* Ignore and warn user if the `PATH` environment variable is set in the environment config. Conda should not enable modifying the `PATH` to any arbitrary value. (#15342)
```

conda/activate.py
```diff
@@ -798,7 +799,17 @@ def _get_environment_env_vars(self, prefix):
...
+            if reserved in env_vars.keys():
+                print(
+                    f"WARNING: environment variable '{reserved}' is configured to be set "
+                    f"to '{env_vars.get(reserved)}'. However, this is a reserved environment "
+                    "variable. This configuration will be ignored.\n"
+                    f"You may remove this invalid configuration by removing the `{reserved}` "
+                    f"key from the file `{env_vars_file}`",
+                    file=sys.stderr,
+                )
+                env_vars.pop(reserved)
```
jezdez: Can we do this in a way that would not print this warning for every reserved env var but essentially warn once and list the reserved env vars that are found?

soapy1:
pre-commit.ci autofix

conda/activate.py
```diff
@@ -798,7 +799,21 @@ def _get_environment_env_vars(self, prefix):
                     )
                     print(f"variable {dup} duplicated", file=sys.stderr)
                 env_vars.update(prefix_state_env_vars)
+        for reserved in RESERVED_ENV_VARS:
+            collect_reserved_vars = []
+            if reserved in env_vars.keys():
+                env_vars.pop(reserved)
+                collect_reserved_vars.append(reserved)
 
+            if collect_reserved_vars:
```
ryanskeith: I think this should be dedented so it isn't in the for loop.

soapy1: that's an unfortunate typo. Good catch!

conda/activate.py
```diff
@@ -805,7 +799,26 @@ def _get_environment_env_vars(self, prefix):
...
+                # Remove from env_vars regardless
+                env_vars.pop(reserved)
+
+        if collect_reserved_vars:
+            print_reserved_vars = ', '.join(collect_reserved_vars)
+            print(
+                f"WARNING: the configured environment variable(s) are reserved and "
+                f"will be ignored: {print_reserved_vars}.\n\n"
+                f"Remove the invalid configuration with `conda env config vars unset "
+                f"{' '.join(collect_reserved_vars)}`.\n",
```
jezdez: @soapy1 I've changed the call to action to edit the state file directly with a call to `conda env config vars unset ..` since the state file shouldn't be modified by users directly

conda/activate.py
```diff
@@ -805,7 +799,26 @@ def _get_environment_env_vars(self, prefix):
...
+        
+        # Remove reserved environment variables and warn if they're being set
+        collect_reserved_vars = []
+        for reserved in RESERVED_ENV_VARS:
+            if reserved in env_vars:
+                # Only warn if the variable is actually being set (not unset)
+                if env_vars[reserved] != CONDA_ENV_VARS_UNSET_VAR:
+                    collect_reserved_vars.append(reserved)
+                # Remove from env_vars regardless
+                env_vars.pop(reserved)
```
jezdez, line 818: When I tested the reset logic, I found that vars that are unset are getting the value of `CONDA_ENV_VARS_UNSET_VAR` which means `***unset***` so we'll need to filter that out here, so even if users have unset the value, it'll be filtered out where this method is called, but we don't want to warn them here already, and only remove them from the vars. We can't change the return value of the `_get_environment_env_vars` method since the full list is needed for the deactivation.

jezdez:
pre-commit.ci autofix

conda/activate.py
soapy1: Cool, I think this is a good change! Thanks for pushing that up. 

I also have this [follow up PR](https://github.com/conda/conda/pull/15354) that warns the user if they are setting/unsetting a reserved environment variable. In that PR, conda will fully remove a reserved env var from the state file if it is unset. Opposed to giving that variable the `***unset***` value.

I think these 2 changes still work well together. So, @jezdez I'm happy to approve/merge if you are :100:

soapy1:
One more change to add here. When I was testing this out and flipping beteween environments with valid and invalid configured env vars, it was confusing to know which environment the warning is coming from. Adding the prefix name to the warning to clarify this.

soapy1:
> One more change to add here. When I was testing this out and flipping beteween environments with valid and invalid configured env vars, it was confusing to know which environment the warning is coming from. Adding the prefix name to the warning to clarify this.

pushed this change. Now when a user activates/deactivates an environment with reserved environment variables configured, they get a warning like:
```
WARNING: the configured environment variable(s) for prefix '/tmp/vars' are reserved and will be ignored: PATH.

Remove the invalid configuration with `conda env config vars unset -p /tmp/vars PATH`.
```

soapy1:
> If you're okay with my changes, I'd be happy to approve it, but we can work on it more if you'd like.

@jezdez I'm good with these changes. Do you want to give it a once thru before merging?

PR #15343, opened 2025-10-20
Don't allow users to configure reserved env vars in their environment

soapy1:
### Description

Pt. 2 to address https://github.com/conda/conda/issues/15292

This PR raises an error when a user attempts to add a reserved env var to an environment config.

For example, to test this:
```
$ conda env config vars set PATH=somevalue

OperationNotAllowed: Environment variable 'PATH' is reserved. It may not be modified as part of an environment configuration.
``` 

or, use `conda env` with the env.yml
```
name: test
channels:
  - conda-forge
dependencies:
  - zlib
variables:
  VARIABLE: value
  PATH: uhho
```

```
$ conda env create --file env.yml
Channels:
 - conda-forge
 - defaults
Platform: linux-64
Solving environment: done

Preparing transaction: done
Verifying transaction: done
Executing transaction: done

OperationNotAllowed: Environment variable 'PATH' is reserved. It may not be modified as part of an environment configuration.

$ conda env config vars list -n test
// no env vars set 
```

### Checklist - did you ...

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -596,12 +596,15 @@ class PrefixData(metaclass=PrefixDataType):
         Ensure that the set of env_var_names does not contain any reserved env vars.
         Will raise an OperationNotAllowed if a reserved env var is found.
         """
+        invalid_vars = []
         for var in RESERVED_ENV_VARS:
             if var in env_vars_names:
-                raise OperationNotAllowed(
-                    f"Environment variable '{var}' is reserved. It may not be "
-                    "modified as part of an environment configuration."
-                )
+                invalid_vars.append(var)
+        if invalid_vars:
+            raise OperationNotAllowed(
+                f"Environment variable(s) '{' '.join(invalid_vars)}' is reserved. It can not be "
+                "modified as part of an environment configuration."
+            )
 
     def get_environment_env_vars(self) -> dict[str, str] | dict[bytes, bytes]:
         prefix_state = self._get_environment_state_file()


conda/core/prefix_data.py
```diff
@@ -589,6 +591,18 @@ def _write_environment_state_file(self, state: dict[str, dict[str, str]]) -> Non
...
         env_vars_file.write_text(json.dumps(state, ensure_ascii=False))
 
+    def _ensure_no_reserved_env_vars(self, env_vars_names: Iterable[str]) -> None:
+        """
+        Ensure that the set of env_var_names does not contain any reserved env vars.
+        Will raise an OperationNotAllowed if a reserved env var is found.
+        """
+        for var in RESERVED_ENV_VARS:
+            if var in env_vars_names:
+                 raise OperationNotAllowed(
```
soapy1: This is pretty strict since if there are valid env vars to export in the set of `env_vars_names` they won't get exported. For example, if you have multiple variables definied in an `environment.yml` file.
 
Another approach is to:
* remove the reserved env var from the set of vars that are getting added to the config
* produce a log message/warning that the bad config is getting ignored.

soapy1:
pre-commit.ci autofix

conda/core/prefix_data.py
```diff
@@ -589,6 +591,18 @@ def _write_environment_state_file(self, state: dict[str, dict[str, str]]) -> Non
...
 
+    def _ensure_no_reserved_env_vars(self, env_vars_names: Iterable[str]) -> None:
+        """
+        Ensure that the set of env_var_names does not contain any reserved env vars.
+        Will raise an OperationNotAllowed if a reserved env var is found.
+        """
+        for var in RESERVED_ENV_VARS:
+            if var in env_vars_names:
+                raise OperationNotAllowed(
+                    f"Environment variable '{var}' is reserved. It may not be "
```
ForgottenProgramme:
```suggestion
                    f"Environment variable '{var}' is reserved. It can not be "
```

Lets use "can" instead of "may".

conda/core/prefix_data.py
```diff
@@ -589,6 +591,18 @@ def _write_environment_state_file(self, state: dict[str, dict[str, str]]) -> Non
         env_vars_file = self.prefix_path / PREFIX_STATE_FILE
         env_vars_file.write_text(json.dumps(state, ensure_ascii=False))
 
+    def _ensure_no_reserved_env_vars(self, env_vars_names: Iterable[str]) -> None:
+        """
+        Ensure that the set of env_var_names does not contain any reserved env vars.
+        Will raise an OperationNotAllowed if a reserved env var is found.
+        """
+        for var in RESERVED_ENV_VARS:
```
ryanskeith, line 600: A small enhance here could be to collect vars that are in Reserved ENV Vars so all the vars are declared at once.

conda/core/prefix_data.py
```diff
@@ -609,9 +624,8 @@ def set_environment_env_vars(
         self._write_environment_state_file(env_state_file)
         return env_state_file.get("env_vars")
 
-    def unset_environment_env_vars(
-        self, env_vars: dict[str, str]
-    ) -> dict[str, str] | None:
+    def unset_environment_env_vars(self, env_vars: list[str]) -> dict[str, str] | None:
+        self._ensure_no_reserved_env_vars(env_vars)
```
ryanskeith, line 631: Huh, I never thought about checking on the unset.  It should be rare but this will help with user education.

soapy1:
> I am torn if this should be a breaking error or a warning type of error with no action

Ya, I think you are right here. Probably better to warn the user. Will make that change.

soapy1:
Here is an alternative solution that warns the user instead of producing an error https://github.com/conda/conda/pull/15354

PR #15347, opened 2025-10-21
Remove extra indentation in _config_xonsh_content(...)

dholth:
### Description

In conda-libmamba-solver's run of the conda unit tests, [setup-miniconda adds a lower part of this file](https://github.com/conda-incubator/setup-miniconda/blob/f77e237980e9445f93e2033f0c4a695a697a967a/src/conda.ts#L387-L397) whose indentation does not match the upper part.

This causes conda's xonsh tests to fail, blocking unrelated conda-libmamba-solver work.

This PR doesn't fix that, but does use f-strings and removes dals().

### Checklist - did you ...


conda/core/initialize.py
```diff
@@ -1559,23 +1559,21 @@ def _config_xonsh_content(conda_prefix):
...
+    from types import ModuleType as _ModuleType
+    _mod = _ModuleType("xontrib.conda",
+                    "Autogenerated from $({conda_exe} shell.xonsh hook)")
+    __xonsh__.execer.exec($("{conda_exe}" "shell.xonsh" "hook"),
+                        glbs=_mod.__dict__,
+                        filename="$({conda_exe} shell.xonsh hook)")
+    _sys.modules["xontrib.conda"] = _mod
+    del _sys, _mod, _ModuleType
+# <<< conda initialize <<<
+"""
```
kenodegard, line 1576: Let's instead rely on string concatenation:

```suggestion
    conda_initialize_content = (
        f"# >>> conda initialize >>>\n"
        f"# !! Contents within this block are managed by 'conda init' !!\n"
        f'if !(test -f "{conda_exe}"):\n'
        f"    import sys as _sys\n"
        f"    from types import ModuleType as _ModuleType\n"
        f'    _mod = _ModuleType("xontrib.conda",\n'
        f'                    "Autogenerated from $({conda_exe} shell.xonsh hook)")\n'
        f'    __xonsh__.execer.exec($("{conda_exe}" "shell.xonsh" "hook"),\n'
        f"                        glbs=_mod.__dict__,\n"
        f'                        filename="$({conda_exe} shell.xonsh hook)")\n'
        f'    _sys.modules["xontrib.conda"] = _mod\n'
        f"    del _sys, _mod, _ModuleType\n"
        f"# <<< conda initialize <<<\n"
    )
```

dholth: Concatenated f-strings are much harder to edit but preserve indentation on the left margin. The linter will complain for every line that has an f-prefix but does not contain substitutions. To me this version is worse than the triple-quoted left-aligned string.

kenodegard: > The linter will complain for every line that has an f-prefix but does not contain substitutions.

Nope, the linter doesn't care about this: https://github.com/astral-sh/ruff/pull/10886#issuecomment-2056940825

This is how we've addressed string indentation elsewhere in conda. Of course most other long strings are continuous text for printing or errors, not code. 🤷🏼‍♂️

If we dislike concatenating then we should make this string a global so we can preserve the Python code indentations.

jaimergp:
What's the issue with `dals()`? Import time?

dholth:
I dislike `dals()` mainly because it has a small import-time overhead but also because it's a conda-specific function. The Python left margin is less important than any factor addressed by strategies that break the left margin. The subjective aesthetics of the code do not affect the end user in any way.

Issue #14274, opened 2024-09-26
Make strict channel priority the new default


jaimergp:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Migrate the factory settings so channel priority uses strict mode by default.

### Why is this needed?

Flexible priority imposes a large performance penalty on mixed channels, usually for little gain. Most of the time, the higher priority channels are expected to shadow the package names in the other channels anyway.

### What should happen?

1. Warn users who are implicitly using flexible channel priority that the default will change in a future version. Warning can be omitted by opting in early into strict (recommended) or explicitly using flexible priority in condarc.
2. After the warning period is over, flip the switch to strict priority by default.

### Additional Context

Comes from https://github.com/conda-forge/conda-forge.github.io/pull/2314#discussion_r1774070760

jaimergp:
Shall we try again?

PR #15359, opened 2025-10-27
Change default channel priority to strict

soapy1:
### Description

ref https://github.com/conda/conda/issues/14274

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -388,7 +388,7 @@ class Context(Configuration):
         aliases=("channel_alias",),
         expandvars=True,
     )
-    channel_priority = ParameterLoader(PrimitiveParameter(ChannelPriority.FLEXIBLE))
+    channel_priority = ParameterLoader(PrimitiveParameter(ChannelPriority.STRICT))
     _channels = ParameterLoader(
         SequenceParameter(PrimitiveParameter("", element_type=str), default=()),
         aliases=(


jaimergp:
Sorry, pressed the wrong button in the GH app but there's no button to mark as draft again 🤷

jezdez:
@jaimergp Not seeing this? 
CleanShot 2025-10-28 at 08 56 34@2x

soapy1:
Taking a look at these test failures, I think that is strict channel priority might be inconvenient for users that:
* have old packages in their environments
* have old/stale/not-updated channels in their configuration

Some notes on what tests are failing:
ref: https://docs.conda.io/projects/conda/en/stable/user-guide/tasks/manage-channels.html#strict-channel-priority

## tests/core/test_solve.py::test_update_deps_2
This test does require flexible channel priority.

## tests/test_create.py::test_conda_pip_interop_pip_clobbers_conda
This is a test that uses pretty old packages, six=1.9, pip=9.0, python=3.5.
It also uses the free channel to install packages. I think the failure
of this test more indicates that using the free channel with strict channel
priority enabled is broken. 
Might be a more effective test if the packages and channels are updated?

## tests/test_create.py::test_conda_pip_interop_compatible_release_operator
Similar to tests/test_create.py::test_conda_pip_interop_pip_clobbers_conda.

## tests/test_create.py::test_neutering_of_historic_specs
Looks like this failure has to do with the test forcing the channel. If 
you apply the diff

```
diff --git a/tests/test_create.py b/tests/test_create.py
index 5a81758d6..2d9818d25 100644
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2583,7 +2583,7 @@ def test_cross_channel_incompatibility(conda_cli: CondaCLIFixture, tmp_path: Pat
 def test_neutering_of_historic_specs(
     tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture
 ):
-    with tmp_env("main::psutil=5.6.3=py37h7b6447c_0") as prefix:
+    with tmp_env("psutil=5.6.3=py37h7b6447c_0") as prefix:
         conda_cli("install", f"--prefix={prefix}", "python=3.6", "--yes")
         d = (prefix / "conda-meta" / "history").read_text()
         assert re.search(r"neutered specs:.*'psutil==5.6.3'\]", d)
```

The test passes. I think this can still be an effective test if either:
* the test was updated to not force the channel of psutil; or
* the test was updated to enable flexible channel priority.

## tests/core/test_solve.py::test_conda_downgrade
This test setups up several different channels with conda dependencies. So,
I think it assumes flexible channel priority. It would be good to explicitly
add that configuration in this test. And maybe add a test that downgrades conda
that assumes strict channel priority. 

## tests/core/test_solve.py::test_solve_2
This test looks like it assumes flexible channel priority.

## tests/core/test_solve.py::test_channel_priority_churn_minimized
Fails with the classic solver, passes with the libmamba solver.
But this assumes flexible channel priority.

## tests/test_priority.py::test_reorder_channel_priority
This test only fails for the classic solver, with pinned_package. In particular,
this fails on a pinned package switching channels when all packages are updated
in the env. I don't think this should happen with strict channel priority enabled,
**this is maybe a bug**?

## tests/test_create.py::test_channel_usage_replacing_python
Just the classic solver fails, libmamba passes. Might be a bug.

## tests/core/test_solve.py::test_conda_downgrade
This one assumes flexible channel priority.


## tests/core/test_solve.py::test_cuda_fail_1
This one is an error about the output of the error string. When channel priority is set to strict, there is some extra info in the error.

## tests/core/test_solve.py::test_cuda_fail_2
same as tests/core/test_solve.py::test_cuda_fail_1.

jaimergp:
> @jaimergp Not seeing this?

On the website yes, but there's no button in the iOS app :/ I wrote that afk.

PR #15375, opened 2025-11-03
Switch to standard BSD 3-Clause

danyeaw:
### Description

Updates our current BSD 3-Clause license to be more standard:

1. Remove the all rights reserved
2. Number the 3 clauses

This was discussed on [Zulip](https://conda.zulipchat.com/#narrow/channel/457607-general/topic/All.20rights.20reserved.20in.20conda.20LICENSE.20files/with/547922679). Once we update this one, I can update the others in the org.

Source of the license formatting is the GitHub default BSD 3-Clause available if you use the new project wizard. It matches the OSI approved license.

### Checklist - did you ...

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,30 +1,31 @@
 BSD 3-Clause License
 
 Copyright (c) 2012, Anaconda, Inc.
-All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
 
-    * Redistributions of source code must retain the above copyright
-      notice, this list of conditions and the following disclaimer.
-    * Redistributions in binary form must reproduce the above copyright
-      notice, this list of conditions and the following disclaimer in the
-      documentation and/or other materials provided with the distribution.
-    * Neither the name of the copyright holder nor the names of its
-      contributors may be used to endorse or promote products
-      derived from this software without specific prior written permission.
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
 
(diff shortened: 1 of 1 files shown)


ForgottenProgramme (commented):
Hey, @danyeaw !
What difference does it make if we remove the "All rights reserved." statement?

LICENSE
```diff
@@ -1,30 +1,31 @@
 BSD 3-Clause License
 
 Copyright (c) 2012, Anaconda, Inc.
-All rights reserved.
```
jezdez, line 4: Let's leave a review comment here for future reference why this is being removed.

jezdez (changes requested):
Could you please add a source of the variant used here? I take it comes from https://opensource.org/license/bsd-3-clause?

LICENSE
danyeaw: All rights reserved is being removed because:

- Putting that statement on copyright notices hasn't been needed since 1989 when the U.S. joined the Berne Convention
- Anaconda is not reserving all rights, we are giving up some rights as outlined by the BSD 3-Clause license, so this is potentially confusing to community members working on the project
- I confirmed with the Anaconda legal team that this isn't helping anything

danyeaw:
Source of the license formatting is the GitHub default BSD 3-Clause available if you use the new project wizard. It matches the OSI license you linked to.

danyeaw:
> What difference does it make if we remove the "All rights reserved." statement?

Hi @scw - added a comment for the reasons to the changes requested, hopefully that clarifies the reason.

PR #15396, opened 2025-11-09
Update __init__.py

Rich-AM:
Added code to detect (via the chardet module) file type encoding, then pass the detected encoding to the appropriate '.read_text' function.  Also added 'log.info' statements to state which files are being read along with their detected file character encoding

### Description

Reason for the change.  

If a conda channel's `repodata.json` file contains any non-ASCII (e.g. UTF-8) characters, conda defaults to the local workstation regional file encoding.  For Linux and Mac, this is typically UTF-8, for Windows, it is the system regional default (cp-1252 Western Latin for most US based Windows systems).  If a single conda channel contains a UTF-8 encoded `repodata.json` file, it can cause a Windows installed version of conda to fail due to an 'invalid character' being detected.  Example: If the unicode character U+201D (the right double quote character) is in the `repodata.json` file for a channel added to the Windows machine conda environment, conda will fail even a simple command such as `conda search pandas`

The changes proposed in this `__init__.py` file merely uses the `chardet` module to detect the file's character encoding, then passes the detected encoding to the appropriate `.read_text` function.  Additionally `log.info` statements have been added to provide users with additional insight into which files are being read (cached or downloaded) as well as their detected file character encoding.

Lastly, all changes were bracketed by a string of 20 '#' with specific comments added.

Thank you for time and consideration of this request

### Checklist - did you ...

I did not do any of the following checklist items as I did not think this change request was significant enough

- ~~[ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- ~~[ ] Add / update necessary tests?~~

### DIFF
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -14,6 +14,11 @@ import pathlib
 import re
 import time
 import warnings
+
+####################
+import chardet                          # Added this import
+####################
+
 from collections import UserDict
 from contextlib import contextmanager
 from os.path import dirname
@@ -99,6 +104,14 @@ class RepoInterface(abc.ABC):
 class Response304ContentUnchanged(Exception):
     pass
 
+####################
+# Added the following lines to detect file encoding
+def detect_encoding(file_path):
+    with open(file_path, 'rb') as f:
+        raw_data = f.read()
+    result = chardet.detect(raw_data)
+    return result['encoding']
+####################
 
 def get_repo_interface() -> type[RepoInterface]:
     if "jlap" in context.experimental:
@@ -551,7 +564,13 @@ class RepodataCache:
                 if binary:
                     json_data = cache_path.read_bytes()
                 else:
-                    json_data = cache_path.read_text()
+                    ####################
+                    # Added the following lines to determine json file encoding (ascii, utf-8, etc)
+                    file_encoding = str(detect_encoding(str(cache_path)))
+                    log.info("Reading file: %s   \t Detected file encoding: %s", str(cache_path), file_encoding)
+                    # Added the encoding parameter to the following line
+                    json_data = cache_path.read_text(encoding=file_encoding)
+                    ####################
 
             json_stat = cache_path.stat()
             if not (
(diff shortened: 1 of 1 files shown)


dholth:
json must be utf-8 only.

It might be worth checking that conda works correctly under less-common default encodings.

https://peps.python.org/pep-0686/ makes utf-8 default in Python 3.15

travishathaway:
@conda-bot check

travishathaway:
Some additional context on this from ChatGPT:

On most Windows systems, the **default character encoding used by Python** is **`cp1252`** (also known as *Windows-1252*), which is a common Western European encoding.

However, this can vary depending on the system locale settings.
Here’s how it typically works:

* **`sys.getdefaultencoding()`** → Always returns `'utf-8'` in modern Python (since Python 3).
* **`locale.getpreferredencoding()`** → Returns the system’s default text encoding, which on Windows is often `'cp1252'` (but could be `'cp932'`, `'cp1251'`, etc., depending on the locale).
* **When opening files without specifying `encoding`** → Python uses the value from `locale.getpreferredencoding(False)` as the default encoding.

So, in short:

> ✅ On most English-language Windows installations, the default encoding is **`cp1252`**, though you should always explicitly specify `encoding="utf-8"` in file operations to ensure portability.

@Rich-AM,

With the above in mind and what you have described, I think you bring up some valid points here, but I don't think we'll accept the solution as you proposed because it's not as efficient as it could be (you load the entire file into memory and these files are normally many megabytes if not hundreds), and these errors aren't very widespread because most `repodata.json` from the most popular channels (e.g. `conda-forge` and Anaconda's `main`) don't contain non-ascii characters.

If you would like to see better character encoding in conda, I suggest opening a separate issue as a "feature" request to clearly definitely the problem and how we can solve it.

Also, for future pull requests to this repository, it's very important to have tests that validate the solution you have proposed. If you are ever unsure of how to write these tests or where to place them please reach out to us via these pull requests or our chat/message board: https://conda.zulipchat.com

Rich-AM:
@travishathaway
I understand the reasoning for your concerns for regarding memory, etc.

In reference to the comments from @dholth earlier regarding PEP-0686 (https://peps.python.org/pep-0686/), would you be willing to skip the function call to detect the file encoding in lieu of passing a deliberate parameter `encoding='utf-8'`? My reasoning is:
   1- Consistency. It will force all systems to use utf-8 as the default encoding  
   1- Backwards compatible to ascii,
   2- Inline with PEP-0686 intent to make Python's default file encoding utf-8
   3- Solve the, admitted uncommon, issues with most non-standard `repodata.json` files

If you would be willing to consider that option, I could resubmit that change as a new pull request

travishathaway:
@Rich-AM,

Thanks for reaching out again. I had the chance to talk this over with a couple other maintainers, and we came to the conclusion that adding these types of checks simply isn't a priority right now, and we don't want the added complexity this would introduce.

Like I said before, I'm not sure we've ever seen anyone else report this issue as being problem (please let me know if I'm wrong here). I am sorry to disappoint you like this, but I would suggest moving on for now.

If you're still interesting in contributing to conda, please check out some more of our open issues or hop on the Zulip chat I mentioned earlier. We'd be happy to get you started on a piece work that is more likely to get merged.

dholth:
It is a good idea to pass an explicit encoding to all text file access in
conda, but I'm not sure how large a change that would be. Especially all
JSON files which must be uft8.

Are you personally experiencing this encoding bug?

Issue #14762, opened 2025-04-16
Activate command exports PATH value that bypasses changes from deactivate scripts


padraic-shafer:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

The order of commands processed in `_Activator._yield_commands()` introduces "sticky" PATH entries for conda environments that use [de]activation scripts to modify the PATH variable. The issue becomes noticeable when more than one of these environments is added to the activation stack.

---
Example 1: Activate and Deactivate work for a SINGLE environment
```
$ conda activate env-A  # modifies the PATH variable

# Path now contains a custom path + the usual conda bin path
(env-A) $ echo $PATH
<path-to-env-A>/my/custom/path/A:<path-to-env-A>/bin:<ORIGINAL_PATH>

# Deactivation exports the PATH
# *before* running the environment's deactivation scripts
#
(env-A) $ conda shell.bash deactivate
export PATH='<path-to-env-A>/my/custom/path/A:$CONDA_PREFIX/bin:<ORIGINAL_PATH>'
. 'path-to-env-A>/etc/conda/deactivate.d/my_deactivate.sh.
unset CONDA_PREFIX
unset CONDA_DEFAULT_ENV
unset CONDA_PROMPT_MODIFIER
PS1=''
export CONDA_SHLVL='0'
export CONDA_EXE='/opt/conda/bin/conda'
export _CE_M=''
export _CE_CONDA=''
export CONDA_PYTHON_EXE='/opt/conda/bin/python'

# Deactivation correctly removes both path entries
(env-A) $ conda deactivate
$ echo $PATH
<ORIGINAL_PATH>
```

---
Example 2: Stacking 2 "activate"s botches the PATH
```
$ conda activate env-A  # modifies the PATH variable

# Path now contains a custom path + the usual conda bin path
(env-A) $ echo $PATH
<path-to-env-A>/my/custom/path/A:<path-to-env-A>/bin:<ORIGINAL_PATH>

# Activating a second environment runs the deactivation scripts
# from the old environment, but then exports the *partially modified*
# PATH that is based on the old environment (prior to deactivation scripts)
#
(env-A) $ conda shell.bash activate env-B
. '<path-to-env-A>/etc/conda/deactivate.d/my_deactivate.sh'
PS1='(env-B) '
export PATH='<path-to-env-A>/my/custom/path/A:<path-to-env-B>/bin:<ORIGINAL_PATH>'
[...]

PR #15398, opened 2025-11-10
Fix PATH handling in conda/conda#14762

cuiluyi:
This pull request resolves [[conda/conda#14762](https://github.com/conda/conda/issues/14762#event-17355311647)](https://github.com/conda/conda/issues/14762#event-17355311647), where the execution order during `conda activate [env]` did not consistently follow the intended sequence of exporting `PATH` before running `deactivate/*.sh` scripts.

### Changes

- Updated the return values of `build_deactivate()` and related functions to include a new field:

    ```python
    "export_path": export_path
    ```

    instead of placing `export_path` inside the `"export vars"` variable.

- This ensures that the `PATH` is exported first, maintaining the correct activation/deactivation behavior.

### Impact

With this change:

- `conda activate [env]` consistently follows the proper `export PATH` → `deactivate/*.sh` order.
- Improves environment activation reliability, especially in complex shell workflows.

### Testing

- All applicable tests have been updated and are passing.
- Verified activation/deactivation order in bash and zsh shells.

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -381,7 +381,7 @@ class _Activator(metaclass=abc.ABCMeta):
                 CONDA_PROMPT_MODIFIER=conda_prompt_modifier,
                 **env_vars,
             )
-            PATH=self.pathsep_join(self._add_prefix_to_path(prefix))
+            PATH = self.pathsep_join(self._add_prefix_to_path(prefix))
             deactivate_scripts = ()
         elif stack:
             export_vars, unset_vars = self.get_export_unset_vars(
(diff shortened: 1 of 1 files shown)


travishathaway:
@cuiluyi,

Can you please resubmit the CLA with the GitHub User IDs field set to `cuiluyi`? This needs to be your exact username.

Thanks.

cuiluyi:
Hi @travishathaway, I have re-submitted the CLA, and the GitHub User IDs field is now set to cuiluyi.
Please let me know if there’s anything else needed.

travishathaway:
@conda-bot check

travishathaway:
pre-commit.ci autofix

travishathaway:
@cuiluyi,

We appreciate this pull request, but we cannot accept this contribution because it presents too big of a change that could affect users in other ways. We have to respect the way conda currently functions not break our user's assumptions of how the software works.

Any change like this would ideally be first turned on and off by a new configuration variable, allowing users to opt-in to the new behavior. Additionally, there are no tests associated with this pull request, and this is problematic because of how complex this issue is. We would need any fixes for this problem to be properly and extensively tested.

Thanks for understanding and let us know if you have further questions.

PR #15407, opened 2025-11-11
Remove unnecessary try-except

kenodegard:
### Description

Removes:
1. unnecessary try-except-raise from conda.cli.main_compare:
    https://github.com/conda/conda/blob/0049f3a2eae767641bf8be5340f6e40572797197/conda/cli/main_compare.py#L132-L133
2. unnecessary assignment back to argparse namespace

### Checklist - did you ...

### DIFF
--- a/tests/cli/test_main_search.py
+++ b/tests/cli/test_main_search.py
@@ -2,16 +2,15 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
-import json
 import re
 from typing import TYPE_CHECKING
 
 import pytest
 import requests
 
-import conda.common.serialize.json
 from conda.base.context import context, reset_context
 from conda.cli.main_search import _pretty_record_format, pretty_record
+from conda.common.serialize import json
 from conda.exceptions import PackagesNotFoundError
 from conda.gateways.anaconda_client import read_binstar_tokens
 from conda.models.records import PackageRecord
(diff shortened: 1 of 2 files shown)


ForgottenProgramme (approved):

conda/cli/main_compare.py
```diff
@@ -106,31 +106,24 @@ def compare_packages(active_pkgs, specification_pkgs) -> tuple[int, list[str]]:
...
-
-        spec_hook = context.plugin_manager.get_environment_specifier(
-            source=filename,
-            name=context.environment_specifier,
-        )
-        spec = spec_hook.environment_spec(filename)
-        env = spec.env
-
-        if args.prefix is None and args.name is None:
-            args.name = env.name
```
jezdez, line 131: I don’t understand why this namespace attribute was set this way, is this needed somewhere else that we don’t see in this part of the code?

kenodegard: This assignment happens within the `execute` function which is only used for the `conda compare` command.

I see this assignment was added with the original `conda compare` command in https://github.com/conda/conda/commit/a35bb09fdf55e9a92d88e5a5dcf26a04a0258930

Since `args` is unused after this point in the code (the `context` object was reset with `args` _before_ `execute` is run) this assignment serves no purpose.

jezdez: Thank you!

PR #15420, opened 2025-11-18
Fix PackageRecord.dist_str() representation in `conda list`

jaimergp:
### Description

This is a callable, not a static field, so it needs to be called. Also, the help message used the repr of a list, which was not pretty or copy-pastable. Join it with commas instead.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_list.py
+++ b/conda/cli/main_list.py
@@ -83,7 +83,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         type=comma_separated_stripped,
         dest="list_fields",
         help="Comma-separated list of fields to print. "
-        f"Valid values: {', '.join(sorted(CONDA_LIST_FIELDS))}.",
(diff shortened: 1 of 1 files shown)


conda/cli/main_list.py
```diff
@@ -79,7 +83,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         type=comma_separated_stripped,
         dest="list_fields",
         help="Comma-separated list of fields to print. "
-        f"Valid values: {sorted(CONDA_LIST_FIELDS)}.",
+        f"Valid values: {', '.join(sorted(CONDA_LIST_FIELDS))}.",
```
dholth: space or no space

jaimergp: Both work as input (but the space requires quoting, of course). Probably easier without.
```suggestion
        f"Valid values: {','.join(sorted(CONDA_LIST_FIELDS))}.",
```

PR #15436, opened 2025-11-24
Validate HTTP headers from session plugins

### LINKED ISSUE #15405: `session|request_headers` plugins should check which headers are being passed
### Checklist

### What happened?

Looks like the logic for custom HTTP headers is not checking which headers are passed from plugins (only making sure that the headers from `requests` are [added last](https://github.com/conda/conda/blob/6832c1853c481c37f07e28775632787364b607d9/conda/gateways/connection/session.py#L234-L246) to avoid overwrites), which might lead to unexpected network errors if someone is not careful.

I wonder if we should prevent some known headers from being set from plugins; e.g. https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header. 

### Conda Info

```shell
N/A
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

_No response_

agriyakhetarpal:
### Description

I have added a list of forbidden HTTP request headers and HTTP method overrides to validate session plugins, such that plugins can't inject arbitrary headers that browsers/HTTP clients shouldn't allow user code to set.

Closes #15405

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -12,7 +12,6 @@ from typing import TYPE_CHECKING
 
 from ... import CondaError
 from ...auxlib.ish import dals
-from ...exceptions import PluginError
 from ...base.constants import CONDA_HOMEPAGE_URL
 from ...base.context import context
 from ...common.url import (
@@ -21,7 +20,7 @@ from ...common.url import (
     split_anaconda_token,
     urlparse,
 )
-from ...exceptions import OfflineError, ProxyError
+from ...exceptions import OfflineError, PluginError, ProxyError
 from ...models.channel import Channel
 from ..anaconda_client import read_binstar_tokens
 from . import (


conda/gateways/connection/session.py
```diff
@@ -232,15 +311,20 @@ def prepare_request(self, request: Request) -> PreparedRequest:
...
         url = urlparse(request.url)
         if url.scheme in ("https", "http"):
+            session_headers = _filter_forbidden_headers(
+                context.plugin_manager.get_cached_session_headers(host=url.netloc)
+            )
+            request_headers = _filter_forbidden_headers(
+                context.plugin_manager.get_cached_request_headers(
+                    host=url.netloc, path=url.path
+                )
+            )
```
jezdez: That's going to be a LOT of additional warnings printed, can you verify they are deduplicated?

jezdez (changes requested):
If a plugin sets any of those headers, I think we should consider it a bug in the plugin and should fail hard, not just warn users about it. This is the type of API that we shouldn't accidentally allow, as it would compromise conda's core network stack.

conda/gateways/connection/session.py
dholth: How many plugins can affect headers? get_cached_session/request_headers(), get_auth_handler(), telemetry?

dholth:
If I was an intentionally malicious plugin, wouldn't I just monkeypatch out the validation function?

jaimergp:
> If I was an intentionally malicious plugin, wouldn't I just monkeypatch out the validation function?

I was mostly concerned about preventing accidental breakages.

conda/gateways/connection/session.py
jezdez: The telemetry plugin is using the header plugin hooks FWIW

jezdez:
Hey @agriyakhetarpal, thanks for working on this!

I've addressed my review feedback in a separate PR against your branch: https://github.com/agriyakhetarpal/conda/pull/4

The main change is to raise a `PluginError` instead of warning and filtering when forbidden headers are detected. This ensures plugin bugs are immediately visible rather than silently failing.

Feel free to merge it into your branch, or let me know if you have any questions!

agriyakhetarpal:
Hi @jezdez, thanks for your PR! I had kind of left this PR alone for a while, as I hadn't taken it up in any sprints, so I am happy we are proceeding with it and I can track it a bit more actively. I've merged your PR here; let's see how it goes :) I would be happy to test this PR downstream in a few repos, since @soapy1 was suggesting that most of the code surrounding this might be untested.

agriyakhetarpal:
To prevent accidental footguns, I've started checking `conda-anaconda-telemetry` and `conda-anaconda-tos`, as those were the ones I could find dealing with these sorts of hooks, but TBH any enterprise plugins are "unknown unknowns"...

jezdez:
Neither of those are setting those headers FWIW

agriyakhetarpal:
Yup, that is what I'm seeing through a few GitHub code searches. We should be good!

Issue #166, opened 2013-06-25
symlink to conda in every environment?


ilanschnell:
Having a symlink to conda in every environment would allow to only have that environment's bin directory added to PATH, and would avoid potential confusion.

PR #15443, opened 2025-11-25
`MatchSpec`: Conditional dependencies, extras and flags

### LINKED ISSUE #15749: Support extras as part of MatchSpec
### Checklist

### What is the idea?

Add an extras key to MatchSpec's bracket syntax, enabling `conda install litellm[extras=proxy]`.

### Why is this needed?

This idea came about when talking to @AlbertDeFusco. Now that Rattler supports extras, it would be nice if when the solver is set to Rattler, that we can use extras syntax to declare a wanted package like `conda install litellm[extras=proxy]`.

### What should happen?

- extras is recognized as a valid bracket key, for example `litellm[extras=proxy,grpc]`)
- It is stored on MatchSpec alongside optional and target (not a record filter)
- The value is forwarded to the solver for optional dependency resolution

### Additional Context

This work is upstream-dependent on CEP PR [#111](https://github.com/conda/ceps/pull/111), which defines the repodata v3 format for extras.

@AlbertDeFusco created a branch to parse extras at one point here: https://github.com/AlbertDeFusco/conda/tree/feat/matchspec-extras

jaimergp:
### Description

Syntax described in https://github.com/conda/ceps/pull/164, https://github.com/conda/ceps/pull/165 and https://github.com/conda/ceps/pull/166, respectively.

### Checklist - did you ...

Pending changes:

- [X] verify the matches cover the full bracket content or raise. See d32019bab98822f16a21605e8a3a9383b518e4df
- [X] verify consecutive bracket pairs are excluded. See d32019bab98822f16a21605e8a3a9383b518e4df
- [ ] Extend coverage:
    - [x] `hash(MatchSpec("pkg[extras=[a]]"))`. See 8e3e7cdc51fe83b80b99576ebc4faa46423b8f68.
    - [x] `MatchSpec("pkg[extras=[a]]").match(some_record)`. Deferred, but added xfailing tests at 019db1aae73a28ca30ee452a87f1457fb51c7ff7.
    - [x] Roundtrip: `MatchSpec(str(MatchSpec(input))) == MatchSpec(input)` for all new fields. See e6dc53a2bbef88e0fec303e169b02626f7a199d2.
    - [X] Combining all three: `pkg[when=__win,extras=[a,b],flags=[cpu]]`. See a34ac0e842eb8f6efce86ffc70d06a0f0e63eaa0.
    - [X] Invalid YAML values: `extras=[True], extras=[1], extras=[null]`. Addressed by [4b063ca](https://github.com/conda/conda/pull/15443/commits/4b063cae84815e4af9f1be3df18e23b5501eab3e).
    - [x] `MatchSpec.merge` with extras/flags (exercises hashing and union paths). Added xfailing tests in 27d35b829893de6eb73ffaba0ab9fb260c7c54b9. Deferred to 2nd PR.
    - [ ] Broader regex hang tests beyond the single URL case
       - JRG: Can't tell how to find more cases.


conda/models/match_spec.py
```diff
@@ -710,11 +725,13 @@ def _parse_spec_str(spec_str):
...
         spec_str.strip()
 
-    # Step 1.b strip ' if ' anticipating future compatibility issues
-    spec_split = spec_str.split(" if ", 1)
-    if len(spec_split) > 1:
-        log.debug("Ignoring conditional in spec %s", spec_str)
-    spec_str = spec_split[0]
+    # Step 1.b save '; if ...' conditions
+    # NOTE: Conditional specs are experimental and subject to changes in syntax
+    if condition := re.match(r"(.*?)\s*;\s*if\s+(.*)\s*", spec_str):
```
jezdez: Should this be a compiled regex?

conda/models/match_spec.py
```diff
@@ -35,6 +36,12 @@
 except ImportError:
     from ..auxlib.collection import frozendict
 
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from conda.models.records import PackageRecord
```
jezdez:
```suggestion
    from .records import PackageRecord
```

conda/models/match_spec.py
```diff
@@ -383,6 +395,9 @@ def __str__(self):
         if brackets:
             builder.append("[{}]".format(",".join(brackets)))
 
+        if condition := self._match_components.get("condition"):
```
jezdez: :walrus:

conda/models/match_spec.py
jaimergp: Yep, although I don't know where I read that regex compilation didn't save that much time since some old version because the interpreter ends up caching them? Let me dig a source and if I can't find it I'll change it.

jaimergp: Hm, looks like yes, it's [cached](https://docs.python.org/3/library/re.html#re.compile), but only a limited amount of them. It might still make a good case for readability so I'll change it.

jaimergp: Added in [88d685d](https://github.com/conda/conda/pull/15443/commits/88d685dec29978cea3dae0bafb587896f693ab4a)

jaimergp:
Still a bit rough but let's get those comments in, @conda/conda-maintainers!

conda/models/match_spec.py
```diff
@@ -1234,4 +1313,6 @@ def match(self, other):
     "features": FeatureMatch,
     "license": CaseInsensitiveStrMatch,
     "license_family": CaseInsensitiveStrMatch,
+    "flags": ListOfStrMatch,  # FIXME: Must accept globs too
+    "extras": ListOfStrMatch,
```
jaimergp: This is temporary while we discuss the parsing changes in the blocks above. Ignore for now.

danyeaw:
I manually tested this PR and https://github.com/conda-incubator/conda-rattler-solver/pull/28 against a wheels channel. They are successfully resolving the `InvalidMatchspec` errors I was getting.

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
-    (["\']?)                 # optional opening quote
-    ([^\'"]*?)               # value
-    (\2)                     # matching closing quote
-    (?:[,\ ]|$)              # delimiter or end
+    (?P<key>[a-zA-Z0-9_-]+?)        # key
+    \s*=\s*                         # separator, with or without surrounding spaces
+    (?:                             # the value can be a:
+                                    # 1. maybe-quoted string
+        (?P<quote_s>["']?)            # optional opening quote
+        (?P<value>[^\'"\[\]]*?)       # value
```
kenodegard: We don't need to escape `'`? Also reordering to match prior line:

```suggestion
        (?P<value>[^"'\[\]]*?)        # value
```

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
+        (?P=quote_s)                  # matching closing quote
+        |                           # 2. a list of maybe-quoted strings
+        \[\s*                         # opening square bracket, with maybe spaces
+        (?P<value_list>
+            (                           # the value is represented by
+                (?P<quote_b>["']?)        # optional opening quote
+                (?P<value_b>[^\'"]*?)     # value
```
kenodegard:
```suggestion
                (?P<value_b>[^"']*?)      # value
```

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
+                (?P=quote_b)              # matching closing quote
+                (?:[,\ ]?)                # delimiter
```
kenodegard: Delimiter must be either a comma (`,`) or an actual space (` `)?

Can it be a comma followed by any number of whitespaces?

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
+    \s*=\s*                         # separator, with or without surrounding spaces
+    (?:                             # the value can be a:
+                                    # 1. maybe-quoted string
+        (?P<quote_s>["']?)            # optional opening quote
+        (?P<value>[^\'"\[\]]*?)       # value
+        (?P=quote_s)                  # matching closing quote
+        |                           # 2. a list of maybe-quoted strings
+        \[\s*                         # opening square bracket, with maybe spaces
+        (?P<value_list>
+            (                           # the value is represented by
```
kenodegard: this is an un-labeled group, do we need it for anything? if not should we make it a non-capture group?


```suggestion
            (?:                         # the value is represented by
```

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
+if TYPE_CHECKING:
+    from typing import Any
+
+    from .records import PackageRecord
+
 log = getLogger(__name__)
 
-_BRACKETS_RE: re.Pattern[str] = re.compile(r".*(?:(\[.*\]))")
+# Matches the whole square brackets section
+_BRACKETS_RE: re.Pattern[str] = re.compile(r".*?(?:(\[.*\]))")
```
kenodegard: A non-capture group wraps a capture group? Can we just drop the non-capture group then?


```suggestion
_BRACKETS_RE: re.Pattern[str] = re.compile(r".*?(\[.*\])")
```

kenodegard: This is a greedy match and will match any brackets within the outer brackets, is that desired?

conda/models/match_spec.py
```diff
@@ -35,20 +37,43 @@
...
+# Matches all the key-value pairs within the square brackets section
 _BRACKETS_KV_RE: re.Pattern[str] = re.compile(
```
kenodegard, line 54: The old pattern was very explicit that the only whitespace allowed was a normal space, now we mix `\s` and `\ `, is this intended?

conda/models/match_spec.py
jaimergp: I guess so? How would we write that, like this?
```suggestion
                (?:,|\ *)?                # delimiter
```

travishathaway:
@jaimergp,

One big problem that I discovered with this pull request is that you don't update the doc string and doc tests for the `MatchSpec` class. That class provides crucial documentation for our MatchSpec implementation so I would say it should be a hard requirement that it be updated too in this pull request.

conda/models/match_spec.py
kenodegard: I'm thinking this?

```suggestion
                (?:[,\ ]\ *)?             # delimiter
```

Or if we're also allowing leading spaces:

```suggestion
                (?:\ *[,\ ]\ *)?          # delimiter
```

But do we want extra spacing at all?

The current pattern matches:

- `value1 value2`
- `value1,value2`

Versus:

- `value1 ,value2`
- `value1, value2`
- `value1 , value2`

travishathaway:
@jaimergp,

Would you be open to adding more tests for `MatchSpec`. I was working with an LLM today and discovered some test cases that currently aren't being covered. I didn't find any bugs with your implementation though! (not sure if that's good or bad...)

I could add these tests directly to your branch if you'd like.

conda/models/match_spec.py
jaimergp: Yes, because we want the inner brackets in fields like `extras=[group1, group2]`.

conda/models/match_spec.py
jaimergp: Good catch, let me revert.

conda/models/match_spec.py
jaimergp: Yea I guess we need extra spacing.

jaimergp:
@kenodegard I applied your changes but then I had to pull some of them back because we have issues with square brackets in `when` specs. I added a few comments about it, I'm not explaining here too much so you can take a look with fresh eyes. Let me know if there are questions!

@travishathaway I applied some of your suggestions. Adding some extra tests revealed some problems. If you want to add some more test cases, go for it (feel free to push here directly), but please only tests for now.

---

I also discovered some silent failures with brackets. `numpy[version<2]` was silently dropping `[version<2]` so the test would always match any numpy version 😂 I added a warning and a deprecated topic to prevent this from being valid in the future.

jezdez (commented):
Went through this pretty carefully with a couple of extra sets of eyes. The parsing work is solid and the tests are good, but there are a few things that will crash at runtime once specs with extras/flags hit the solver. Details inline below, plus some broader items here.

Bracket parsing has a couple of rough edges that aren't covered by inline suggestions:

- `finditer` on the bracket content isn't anchored, so valid leading KV pairs followed by junk (like `version='>=1',garbage`) parse fine and silently drop the rest. For something as foundational as MatchSpec, it would be better to verify the matches cover the full bracket content or raise.

- `_BRACKETS_RE` with greedy inner (`\[.*\]`) captures `[a][b]` as one bracket section, so `name[a][b]` yields `a][b` inside the brackets and the error message suggests `extras=[a][b]` which isn't valid.

- The two FIXMEs (`quick copy of FeatureMatch`, `Must accept globs too`) should probably be tracked as issues before this lands so they don't rot.

Missing test coverage that would catch the blockers and some of the edge cases:

- `hash(MatchSpec("pkg[extras=[a]]"))` -- crashes right now
- `MatchSpec("pkg[extras=[a]]").match(some_record)` -- also crashes
- Roundtrip: `MatchSpec(str(MatchSpec(input))) == MatchSpec(input)` for all new fields
- Combining all three: `pkg[when=__win,extras=[a,b],flags=[cpu]]`
- Invalid YAML values: `extras=[True]`, `extras=[1]`, `extras=[null]`
- `MatchSpec.merge` with extras/flags (exercises hashing and union paths)
- Broader regex hang tests beyond the single URL case

A few design questions:

The old code silently stripped ` if ` conditions. Now they get folded into the version/build string instead of erroring. Should this get a deprecation warning for a cycle, or are we confident nothing in the wild uses it? A quick grep across conda-build and recipes would settle it.

`__repr__` sorts (set-like), `match()` does ordered list equality (sequence-like). For extras and flags the natural model is probably subset ("all requested are present"), not exact ordered match. Worth deciding now before the storage format is baked in. Any preference from the CEP discussion?

Parsing and storage for `when` look good, `match()` correctly skips it. But nothing evaluates it against virtual packages at solve time yet. If that's in [conda-rattler-solver#28](https://github.com/conda-incubator/conda-rattler-solver/pull/28), linking it from the description would help. Otherwise the news entry should note "parsing only, no behavioral effect yet".

The FIXME says "must accept globs too" and the test uses `blas:*`, but the matcher does exact list equality. Is glob support intended for this PR or a follow-up? If follow-up, worth filing an issue.

`26.5` to `26.11` is six months for syntax that's been silently accepted for years. Is that enough given how widely MatchSpec is used downstream? A year might be safer.

jezdez:
Sorry for the noise, that review didn't land some of the suggestions at the right lines. Fixing..

jezdez (commented):
Inline suggestions for the issues mentioned in the review above.

conda/models/match_spec.py
```diff
@@ -1166,6 +1235,46 @@ def exact_value(self):
...
+        if not value:
+            return []
+        elif isinstance(value, str):
+            return [
+                f
+                for f in (ff.strip() for ff in value.replace(" ", ",").split(","))
+                if f
+            ]
+        else:
+            return [f for f in (ff.strip() for ff in value) if f]
```
jezdez: `_convert` returns a `list`, which makes `_raw_value` unhashable, but `__hash__` on line 1270 calls `hash(self._raw_value)`. This will blow up the first time a `MatchSpec` with `extras` or `flags` ends up in a `frozenset` or dict key (which the solver does). Returning `tuple` everywhere should fix it:

```suggestion
    def _convert(self, value):
        if not value:
            return ()
        elif isinstance(value, str):
            return tuple(
                f
                for f in (ff.strip() for ff in value.replace(" ", ",").split(","))
                if f
            )
        else:
            return tuple(f for f in (ff.strip() for ff in value) if f)
```

conda/models/match_spec.py
```diff
@@ -299,6 +338,10 @@ def match(self, rec):
 
             rec = PackageRecord.from_objects(rec)
         for field_name, v in self._match_components.items():
+            if field_name == "when":
+                # Conditions do not apply to check whether a record
+                # matches a given match spec.
+                continue
```
jezdez, line 362: Skipping `when` here makes sense, but `PackageRecord` has no `extras` or `flags` attributes, and the `_match_individual` fallback calls `rec.__getattr__(field_name)` which returns `None`. `ListOfStrMatch.match(None)` then feeds `None` into `_convert` which returns `[]`/`()`, so the match silently fails for any spec that carries `extras` or `flags`. Either skip these fields too, or add the attributes to `PackageRecord` so match actually works.

conda/models/match_spec.py
```diff
@@ -786,16 +825,46 @@ def _parse_spec_str(spec_str):
...
+                    raise InvalidSpec(
+                        "'when' specs cannot include inner 'when' fields",
+                    )
+                # Q: Canonicalize or not?
+                value = str(when_spec)
             brackets[key] = value
+        if not brackets:
+            # No key-value pairs found but there was a outer square brackets match?
+            # That's invalid syntax (e.g. accidental `package[extra]`)
+            if "version" in brackets_str:
```
jezdez: This is a substring check, so `package[subversion<2]` or `package[my_version_thing]` would also match. Consider:

```suggestion
            if brackets_str.startswith("version"):
```

conda/models/match_spec.py
```diff
@@ -786,16 +825,46 @@ def _parse_spec_str(spec_str):
...
+            if not (key and value):
+                raise InvalidSpec(
+                    "key-value mismatch in brackets; a key or a value is missing",
                 )
-            if key == "version" and value:
-                value = _sanitize_version_str(value, match.groupdict().get("build"))
+            if key == "version":
+                value = _sanitize_version_str(value, groups.get("build"))
+            elif key in ("flags", "extras"):
+                value = tuple(yaml.loads(f"[{value.strip('[]')}]"))
```
jezdez: `yaml.loads` coerces unquoted values: `True` stays `True` (bool), `1` becomes `1` (int), `null` becomes `None`. So `flags=[True, 1, null]` would silently produce `(True, 1, None)` instead of strings. Worth either validating that every element is `str` after the parse, or documenting that non-string values are intentionally supported.

conda/models/match_spec.py
```diff
@@ -786,16 +825,46 @@ def _parse_spec_str(spec_str):
...
+            key = groups["key"]
+            value = groups["value"] or groups["value_list"]
+            if not (key and value):
+                raise InvalidSpec(
+                    "key-value mismatch in brackets; a key or a value is missing",
                 )
-            if key == "version" and value:
-                value = _sanitize_version_str(value, match.groupdict().get("build"))
+            if key == "version":
+                value = _sanitize_version_str(value, groups.get("build"))
```
jezdez: The new `_BRACKETS_KV_RE` regex has groups `key`, `value`, `value_list`, `quote_s`, but no `build` group. So `groups.get("build")` always returns `None` here. In the old code the single-regex approach captured `build` in the same match. Now that KV parsing is a second regex pass, the build string needs to come from a separate `brackets.get("build")` lookup after the loop, or version+build need to be handled together in a second pass.

conda/models/match_spec.py
```diff
@@ -806,8 +875,8 @@ def _parse_spec_str(spec_str):
         parens_str = parens_str[1:-1]
         m4b = _BRACKETS_KV_RE.finditer(parens_str)
         for match in m4b:
-            key, _, value, _ = match.groups()
-            parens[key] = value
+            groups = match.groupdict()
+            parens[groups["key"]] = groups["value"]
```
jezdez: This only reads `groups["value"]`, which is `None` when the input uses the list syntax (`value_list` group). The bracket-parsing loop on line 832 already does `groups["value"] or groups["value_list"]`, but this parens path does not.

```suggestion
            parens[groups["key"]] = groups["value"] or groups["value_list"]
```

conda/models/match_spec.py
```diff
@@ -786,16 +825,46 @@ def _parse_spec_str(spec_str):
...
+            elif key in ("flags", "extras"):
+                value = tuple(yaml.loads(f"[{value.strip('[]')}]"))
+            elif key == "when":
+                when_spec = MatchSpec(value)
+                if when_spec.get("when"):
+                    raise InvalidSpec(
+                        "'when' specs cannot include inner 'when' fields",
+                    )
+                # Q: Canonicalize or not?
+                value = str(when_spec)
```
jezdez: The inline question is a real design decision. Canonicalizing means `MatchSpec('foo[when=python >=3]')` round-trips as `foo[when='python >=3']` (normalized), which is good for dedup and solver caching. But it also means the user's original string is lost, which matters if the canonical form diverges from what they typed. I'd lean toward canonicalizing, since `MatchSpec.__str__` already normalizes everything else, and not canonicalizing means two specs that are semantically identical could compare differently.

conda/models/match_spec.py
```diff
@@ -186,6 +218,10 @@ class MatchSpec(metaclass=MatchSpecType):
         "conda-forge/linux-64::foo[version='>=1.0']"
         >>> str(MatchSpec('*/linux-64::foo>=1.0'))
         "foo[subdir=linux-64,version='>=1.0']"
+        >>> str(MatchSpec('package[extras=[a,b]]'))
+        "package[extras=['a', 'b']]"
+        >>> str(MatchSpec('package[when=__unix]'))
+        "package[when='__unix']"
```
jezdez: The doctest on line 223 shows `str(MatchSpec('package[when=__unix]'))` producing `"package[when='__unix']"` (quoted). But `__str__` on line 452-458 only quotes when the value contains `, `, `=`, or starts with `[`. The string `__unix` contains none of those, so it would actually emit `package[when=__unix]` (unquoted). Either the doctest is wrong, or `__str__` needs to always quote the `when` value.

news/15443-conditionals
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add support for conditional dependencies, optional dependency groups, and variant flags in `MatchSpec` expressions (introduced in [CEP PR#111](https://github.com/conda/ceps/pull/111)). (#15443)
```
jezdez: The PR description references three CEP drafts (conda/ceps PRs #164, #165, #166), but the news entry links to `ceps/pull/111`. The CEPs are still unnumbered drafts (`CEP XXXX`), so linking to the PRs directly is the right call, just with the correct PR numbers. The conda-rattler-solver companion PR ([conda-rattler-solver#203](https://github.com/conda/conda-rattler-solver/pull/203)) might also be worth mentioning here or in the PR description.

```suggestion
* Add support for conditional dependencies, optional dependency groups, and variant flags in `MatchSpec` expressions (introduced in [conda/ceps#164](https://github.com/conda/ceps/pull/164), [conda/ceps#165](https://github.com/conda/ceps/pull/165), and [conda/ceps#166](https://github.com/conda/ceps/pull/166)). (#15443)
```

jaimergp:
Solid review, thanks Jannis. All excellent points. I'll cover them in small commits so you can verify easily, but probably not today :D

danyeaw:
Hey @jaimergp, this could also be done in a follow-up PR. I was testing this branch against a conda-pypi test channel. Currently, this `when` value parser only handles a single MatchSpec, but conda-pypi already generates boolean and/or conditions from wheel environment markers:

`requests; sys_platform == "linux" or sys_platform == "darwin" → requests[when="(__linux or __osx)"]`


`requests; python_version not in "3.0, 3.1" → requests[when="(python!=3.0 and python!=3.1)"]`


These fail to parse:

```
>>> from conda.models.match_spec import MatchSpec
... 
... MatchSpec('requests[when="(__linux or __osx)"]')
  File "/Users/dyeaw/Projects/conda/conda/models/match_spec.py", line 114, in __call__
    parsed = _parse_spec_str(spec_arg)
  File "/Users/dyeaw/Projects/conda/conda/models/match_spec.py", line 842, in _parse_spec_str
    when_spec = MatchSpec(value)
  File "/Users/dyeaw/Projects/conda/conda/models/match_spec.py", line 143, in __call__
    raise InvalidMatchSpec(msg, details=e) from e
conda.exceptions.InvalidMatchSpec: Invalid spec '(__linux or __osx)': Invalid spec '(__linux or __osx)': no package name found in 'None'
```

Likely cause: `_PARENS_RE` uses `.*` before (, so any expression starting with ( has its content stripped, leaving nothing for the name parser.

jaimergp:
@danyeaw, [3564d03](https://github.com/conda/conda/pull/15443/commits/3564d03b2fea0e417b6392b2d0b2d088fbaa20c1) adds the necessary changes to the parser so it considers expressions. It's a crude attempt that can be refined later.

conda/models/match_spec.py
jaimergp: We should have a linter rule for these

conda/models/match_spec.py
jaimergp: Not relevant anymore, since `when` can also include expressions and canonicalizing those may involve too many changes. I lean towards leaving things as they are for simplicity. I'm only checking they are valid so far.

conda/models/match_spec.py
jaimergp: I think they are all supposed to be strings. I added a post-processing step and a test.

conda/models/match_spec.py
jaimergp: Added to records. Should probably refine that in a later PR though.

conda/models/match_spec.py
jaimergp: See 375e9b2415f065f0c98f0facc42c14520efc6722

jaimergp:
@jezdez, I covered your inline comments for now. Also found some more issues while looking at those.

jaimergp:
> finditer on the bracket content isn't anchored, so valid leading KV pairs followed by junk (like version='>=1',garbage) parse fine and silently drop the rest. For something as foundational as MatchSpec, it would be better to verify the matches cover the full bracket content or raise.

Added to PR description as checkbox "verify the matches cover the full bracket content or raise"

> _BRACKETS_RE with greedy inner (\[.*\]) captures [a][b] as one bracket section, so name[a][b] yields a][b inside the brackets and the error message suggests extras=[a][b] which isn't valid.

Added to PR description as checkbox "verify consecutive bracket pairs are excluded"

> The two FIXMEs (quick copy of FeatureMatch, Must accept globs too) should probably be tracked as issues before this lands so they don't rot.

I created an epic to track all the work involved in this effort: https://github.com/conda/conda/issues/16073

> Missing test coverage that would catch the blockers and some of the edge cases:
> 
> hash(MatchSpec("pkg[extras=[a]]")) -- crashes right now
> MatchSpec("pkg[extras=[a]]").match(some_record) -- also crashes
> Roundtrip: MatchSpec(str(MatchSpec(input))) == MatchSpec(input) for all new fields
> Combining all three: pkg[when=__win,extras=[a,b],flags=[cpu]]
> Invalid YAML values: extras=[True], extras=[1], extras=[null]
> MatchSpec.merge with extras/flags (exercises hashing and union paths)
> Broader regex hang tests beyond the single URL case

Added all these to the PR description too.

> A few design questions:
> 
> The old code silently stripped if conditions. Now they get folded into the version/build string instead of erroring. Should this get a deprecation warning for a cycle, or are we confident nothing in the wild uses it? A quick grep across conda-build and recipes would settle it.

I don't get it. If we silently stripped things, how is this "instead of erroring". We never supported, they just added support in the parser in case it ever got implemented in the solver, they wouldn't need to wait for ` if ` to be parsable. This expression is tricky to grep so I can't tell if it was used, but I'm not too concerned if there ancient specs using it, it's a mistake. It was also bad: it would split on build strings (or versions) that happened to be `if`.

> __repr__ sorts (set-like), match() does ordered list equality (sequence-like). For extras and flags the natural model is probably subset ("all requested are present"), not exact ordered match. Worth deciding now before the storage format is baked in. Any preference from the CEP discussion?

We should probably check what `rattler` decided here, but that part would be implemented in the PR for "matching fields", excluded here. I'm inclined to do subset checks, if you want my opinion now. At least for `flags`. For `extras`, it's weird; we don't really query on `depends` either, and in a way `extras` is "solver-only" metadata. What would be the question? "Match packages that exhibit these extras names?" Even if the names imply completely different packages in each case? (Assuming we only want to match on the keys, obviously; let's not consider matching the map).

> Parsing and storage for when look good, match() correctly skips it. But nothing evaluates it against virtual packages at solve time yet. If that's in conda-rattler-solver#28, linking it from the description would help. Otherwise the news entry should note "parsing only, no behavioral effect yet".

We are not equipped here to perform `when` evaluations (unless we want to add it to `classic`, to which I will loudly beg not to). I included this in https://github.com/conda/conda/issues/16073.

> The FIXME says "must accept globs too" and the test uses blas:*, but the matcher does exact list equality. Is glob support intended for this PR or a follow-up? If follow-up, worth filing an issue.

See https://github.com/conda/conda/issues/16073

> 26.5 to 26.11 is six months for syntax that's been silently accepted for years. Is that enough given how widely MatchSpec is used downstream? A year might be safer.

I only found three instances: https://github.com/Sydney-Informatics-Hub/AgReFed-Workshop/blob/d5ce77665cd4e4f8bc366230a2640a96cae1c29b/_import/environment.yml#L6, https://github.com/UBC-MDS/boardgame_rating_predictor/blob/8a79361132b6368c4a458cd9c91b5f2dd1949ce8/envboard.yaml#L29, https://github.com/UBC-MDS/dropout-predictions/blob/8ff9cffb823e8350a2cecc9300f41d2366428dd8/env/dropout_pred_env.yml#L28. I think 6 months is enough, in principle. But I'm fine with a longer cycle, or even permanent support. It's a ticking bomb that should be fixed, imo, though. [^1]

[^1]: We were silently dropping these, so if the user didn't notice the typo, it's because the solver was finding the expected solution anyway (dropping the version meant only the name was used, so it would always match any package). For "greater than", the solver defaults to latest version, so the resulting name-only spec should work in most cases; if not, then the user would investigate and find about the bug. For "smaller than", I guess they got lucky and no newer versions were published or another package in the environment was constraining it too. Regardless, if they are observed in the wild it's because they _happened_ to work.

tests/models/test_match_spec.py
```diff
@@ -1534,3 +1880,221 @@ def test_parse_spec_str_with_precompiled_regex(
...
+@pytest.mark.xfail(
+    reason=(
+        "_validate_when_spec splits naively on ' and '/'  or ', "
+        "corrupting quoted bracket values that contain those words"
+    )
+)
+def test_when_inner_bracket_with_quoted_boolean_word():
+    # A when value whose inner bracket spec contains a quoted string with "and"
+    # e.g. python[build="fast and slow"] — the split on ' and ' corrupts this
+    MatchSpec(r'pkg[when="python[build=\"fast and slow\"]"]')
```
jaimergp: @danyeaw, build strings cannot contain spaces 🤔

tests/models/test_match_spec.py
```diff
@@ -1534,3 +1880,221 @@ def test_parse_spec_str_with_precompiled_regex(
...
+def test_extras_flags_empty_value_raises(spec):
+    with pytest.raises(InvalidMatchSpec):
+        MatchSpec(spec)
+
+
+@pytest.mark.parametrize(
+    "spec",
+    [
+        pytest.param(
+            "pkg[extras=[a,]]",
```
jaimergp: I'm not sure this one should be bad syntax

tests/models/test_match_spec.py
```diff
@@ -1534,3 +1880,221 @@ def test_parse_spec_str_with_precompiled_regex(
...
+            marks=pytest.mark.xfail(
+                reason="Trailing comma in YAML flow sequence behavior not guaranteed to raise"
+            ),
+        ),
+        pytest.param(
+            "pkg[extras=[a,,b]]",
```
jaimergp: I agree this one should not be valid

jaimergp: Fixing this in YAML will require some customization of the parser, for now I'd say we'll need to leave with it.

tests/models/test_match_spec.py
danyeaw: Hey @jaimergp, ya, I agree - probably not a good test, we can remove it.

tests/models/test_match_spec.py
danyeaw: Hey @jaimergp, I pushed a change which makes this valid syntax now.

tests/models/test_match_spec.py
danyeaw: Since this is a small use case of have `,,` or `, ,`, I added a small regex which allows us to at least catch this error and give it to the user rather than throwing a YAML exception.

danyeaw (approved):

conda/models/match_spec.py
```diff
@@ -49,6 +62,38 @@
...
+    r"""
+    (?P<key>[a-zA-Z0-9_-]+?)        # key
+    \ *=\ *                         # separator, with or without surrounding spaces
+    (?:                             # the value can be a:
+                                      # 1. a list of maybe-quoted strings (key={extras,flags})
+                                      #    this has to be matched first so we can capture
+                                      #    comma-separated lists fully (e.g. flags=[a, b])
+                                      #    without being caught by 'when' specs like
+                                      #    `python[version='>=3,<4']`
+        \[ *                            # opening square bracket, with maybe spaces
```
kenodegard:
```suggestion
        \[\ *                           # opening square bracket, with maybe spaces
```

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+            "name": "*",
+            "track_features": (feature_name,),
+        }
+
+    # Step 1. strip '#' comment
+    if "#" in spec_str:
+        ndx = spec_str.index("#")
+        spec_str, _ = spec_str[:ndx], spec_str[ndx:]
+        spec_str.strip()
+
```
kenodegard, line 997: Note to self: dropped `if` stripping

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+                raise InvalidSpec(
+                    "No key-value pairs found in square brackets, "
+                    f"did you mean `extras=[{brackets_str}]`?",
+                )
+
+    # Re-process version field, whose syntax depends on the presence of a build field
+    if version_value := brackets.get("version"):
+        brackets["version"] = _sanitize_version_str(
+            version_value, brackets.get("build")
+        )
```
kenodegard, line 1095: Note to self: step 3 is the change

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+        escaped = found.replace('"', '\\"')
+        spec = spec.replace(found, f'"{escaped}"')
+    if len(specs) == 1:
+        # no boolean logic found, just check MatchSpec is parsable
+        return
+    try:
+        ast.parse(spec)
+    except (ValueError, SyntaxError) as exc:
+        raise InvalidSpec("'when' expression cannot be parsed") from exc
+    return
```
kenodegard, line 1232: \*\*eye twitch**

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+                # Backwards compatibility for a bit; supporting this was an accident!
+                brackets_str_version = brackets_str[len("version") :]
+                brackets["version"] = _sanitize_version_str(brackets_str_version, None)
+                deprecated.topic(
+                    "26.5",
+                    "26.11",
+                    topic="MatchSpec version expression without = separator",
+                    addendum=f"This MatchSpec syntax `{original_spec_str}` was accidentally "
+                    f"allowed in the past. Please use `version='{brackets_str_version}'`.",
+                )
```
kenodegard, line 1084: 26.11 is not a deprecation release:


```suggestion
                deprecated.topic(
                    "26.9",
                    "27.3",
                    topic="MatchSpec version expression without = separator",
                    addendum=f"This MatchSpec syntax `{original_spec_str}` was accidentally "
                    f"allowed in the past. Please use `version='{brackets_str_version}'`.",
                )
```

news/15443-conditionals
```diff
@@ -0,0 +1,19 @@
...
+
+* Add parsing and storage support for conditional dependencies, optional dependency groups, and variant flags in `MatchSpec` expressions (introduced in [conda/ceps#164](https://github.com/conda/ceps/pull/164), [conda/ceps#165](https://github.com/conda/ceps/pull/165), and [conda/ceps#166](https://github.com/conda/ceps/pull/166)). (#15443)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Accidentally allowed and silently ignored `package[version<1]`-like `MatchSpecs` will be removed in `26.11`. The square bracket syntax must always include `key=value` pairs. (#15443)
```
kenodegard:
```suggestion
* Accidentally allowed and silently ignored `package[version<1]`-like `MatchSpecs` will be removed in `27.3`. The square bracket syntax must always include `key=value` pairs. (#15443)
```

tests/models/test_match_spec.py
```diff
@@ -581,7 +586,14 @@ def test_bracket_matches():
     }
 
     assert MatchSpec("numpy<2").match(record)
-    assert MatchSpec("numpy[version<2]").match(record)
+    with pytest.warns(
+        (DeprecationWarning, PendingDeprecationWarning),
+        match="Please use `version='<2'`",
+    ):
```
kenodegard:
```suggestion
    with pytest.deprecated_call(match="Please use `version='<2'`"):
```

tests/models/test_match_spec.py
```diff
@@ -1534,3 +1883,198 @@ def test_parse_spec_str_with_precompiled_regex(
...
+    ],
+)
+def test_extras_invalid_list_syntax_raises(spec):
+    with pytest.raises(InvalidMatchSpec):
+        MatchSpec(spec)
+
+
+def test_bracket_trailing_junk_raises():
+    with pytest.raises(InvalidMatchSpec):
+        MatchSpec("pkg[build=0,junk]")
```
kenodegard: combine?


```suggestion
@pytest.mark.parametrize(
    "spec",
    [
        "pkg[extras=]",  # empty value — caught before YAML is called
        "pkg[flags=]",  # same for flags
        "pkg[extras=[a,,b]]",  # invalid list
        "pkg[build=0,junk]",  # trailing junk
    ],
)
def test_extras_flags_empty_value_raises(spec):
    with pytest.raises(InvalidMatchSpec):
        MatchSpec(spec)
```

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+            if not (key and value):
+                raise InvalidSpec(
+                    "key-value mismatch in brackets; a key or a value is missing",
+                )
+            if "][" in value:
+                raise InvalidSpec("Multiple bracket sections are not allowed")
+            if key in ("flags", "extras"):
+                if (value[0] == "[" and value[-1] != "]") or (
+                    value[-1] == "]" and value[0] != "["
+                ):
```
kenodegard:
```suggestion
                if (value[0] == "[") ^ (value[-1] == "]"):
```

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+        for b in a.split(" or "):
+            specs.append(b.strip(" ()'\""))
+    for found in specs:
+        parsed = MatchSpec(found)
+        if parsed.get("when"):
+            raise InvalidSpec(
+                f"'when' specs cannot include inner 'when' fields: {found} ",
+            )
+        escaped = found.replace('"', '\\"')
+        spec = spec.replace(found, f'"{escaped}"')
```
kenodegard, line 1224: I think we strip too much (see `b.strip(" ()'\"")`) for the escape/replace (see `spec.replace(found, f'"{escaped}"')`) to work correctly:

```python
from conda.models.match_spec import _validate_when_spec
>>> _validate_when_spec('("python" and "more")')
  File "/Users/kodegard/dev/core/conda/conda/models/match_spec.py", line 1020, in _validate_when_spec
    ast.parse(spec)
  File "/Users/kodegard/.conda/devenv/Darwin/arm64/envs/devenv-3.10-miniconda/lib/python3.10/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    (""python"" and ""more"")
     ^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?

The above exception was the direct cause of the following exception:

  File "/Users/kodegard/dev/core/conda/conda/models/match_spec.py", line 1022, in _validate_when_spec
    raise InvalidSpec("'when' expression cannot be parsed") from exc
conda.exceptions.InvalidSpec: 'when' expression cannot be parsed
```

maybe the initial strip should only do `b.strip(" ()")`? then do a second strip to remove quotes before parsing with MatchSpec?

conda/models/match_spec.py
```diff
@@ -898,6 +971,269 @@ def _parse_spec_str(spec_str):
...
+    >>> _validate_when_spec("(python and __unix)")
+    >>> _validate_when_spec("(python and __unix>=3)")
+    >>> _validate_when_spec("(python and __unix>=3 or __win)")
+    """
+    if not spec or not spec.strip():
+        raise InvalidSpec("'when' value must not be empty")
+    specs = []
+    for a in spec.split(" and "):
+        for b in a.split(" or "):
+            specs.append(b.strip(" ()'\""))
```
jaimergp:
```suggestion
            specs.append(b.strip(" ()").strip("'\""))
```

conda/models/match_spec.py
jaimergp: Added your suggestion in https://github.com/conda/conda/pull/15443/files#r3232895401

tests/models/test_match_spec.py
jaimergp: Lovely, thanks.

conda/models/match_spec.py
kenodegard: It needs to be split, the first strip is to remove the `()` (and any extra padding), the second strip is to remove quotes before passing to `MatchSpec`:

```suggestion
    for a in spec.split(" and "):
        for b in a.split(" or "):
            specs.append(b.strip(" ()"))
    for found in specs:
        parsed = MatchSpec(found.strip("'\""))
        if parsed.get("when"):
            raise InvalidSpec(
                f"'when' specs cannot include inner 'when' fields: {found} ",
            )
        escaped = found.replace('"', '\\"')
        spec = spec.replace(found, f'"{escaped}"')
```

Also isn't the AST parsing only to ensure the logical expressions is valid? IOW we could use a dummy value for replacements? I don't expect us to do this just confirming my understanding.

```python
    for a in spec.split(" and "):
        for b in a.split(" or "):
            specs.append(b.strip(" ()"))
    for found in specs:
        parsed = MatchSpec(found.strip("'\"")
        if parsed.get("when"):
            raise InvalidSpec(
                f"'when' specs cannot include inner 'when' fields: {found} ",
            )
        spec = spec.replace(found, f'"{uuid4().hex}"')  # use a placeholder
```

jaimergp: Yes, it's only for parsing now. We could replace it with `None` for all I care.

kenodegard: Followup, why are we allowing the individual specs within the `when` key to be quoted at all?

`"python==3"` is an invalid match spec:

```pycon
>>> from conda.models.match_spec import MatchSpec
>>> MatchSpec("'python==3'")
...
conda.exceptions.InvalidMatchSpec: Invalid spec ''python==3'': Invalid version '3'': invalid character(s)
```

jaimergp: CEP 29 [says](https://github.com/conda/ceps/blob/main/cep-0029.md#syntax):

> The value MUST be quoted with single ' or double " quotes if it contains spaces, commas, equal signs, or square brackets. Quoting rules follow [Python's string literals](https://docs.python.org/3/reference/lexical_analysis.html#strings).

And MatchSpecs MAY contain those, so we quote always just in case.

conda/models/match_spec.py
```diff
@@ -898,6 +971,267 @@ def _parse_spec_str(spec_str):
...
+    >>> _validate_when_spec("(python and __unix)")
+    >>> _validate_when_spec("(python and __unix>=3)")
+    >>> _validate_when_spec("(python and __unix>=3 or __win)")
+    """
+    if not spec or not spec.strip():
+        raise InvalidSpec("'when' value must not be empty")
+    specs = []
+    for a in spec.split(" and "):
+        for b in a.split(" or "):
+            specs.append(b.strip(" ()").strip("'\""))
```
kenodegard: As discussed offline we're going to drop quote stripping from within the `when` clause


```suggestion
            specs.append(b.strip(" ()"))
```

kenodegard: See https://github.com/conda/conda/pull/15443/commits/92016cb6089e01753e27c1436d2038b7d03a16ee

conda/models/match_spec.py
kenodegard: Resolved via https://github.com/conda/conda/pull/15443#discussion_r3235603498

kenodegard (approved):

PR #15450, opened 2025-11-27
Add tests for `conda.cli.common.print_activate`

agriyakhetarpal:
### Description

~I was reading through this file recently and noticed this unresolved TODO comment that had been there for quite some while now. This PR replaces the bare `print()` statement with a logger to print the message displayed when an environment is created.~ The `print_activate` function was untested, so I've added a couple of simple tests in this PR.

### Checklist - did you ...


agriyakhetarpal:
Thanks for the review and the extra context around logging, @jezdez. I hadn't seen that page in the documentation before. I've added a news entry. I wouldn't really count this as an enhancement or a bug fix, as there is no user-facing change, so I used the "Other" heading. Could you confirm if 45f7e498203e2df70cd249adfd415c720ab37298 is okay?

kenodegard:
I don't think we should be using the logger as the `TODO` says for two reasons:

1. the `conda.stdout` logger isn't compatible with the reporter backend and
2. the `conda.stdout` logger is inconsistently used in conda (used in a few core locations but most printout use `print`).

agriyakhetarpal:
Thanks for the context, @kenodegard – I wasn't aware of this. Should we close this PR, then?

jezdez:
We could repurpose this PR to add the tests?

PR #15453, opened 2025-11-28
Speed up our Windows CI by setting `TEMP` to the `D:` drive

agriyakhetarpal:
### Description

The fact that I/O operations on Windows CI are slow is a well-known issue with GitHub Actions runners (see https://github.com/actions/runner-images/issues/8755, which is marked as resolved but actually isn't). The idea with this PR is to move the temporary directory to the D: drive on Windows runners, since the D: drive is faster for reads/writes than the C: drive, where projects are checked out.

Several people have tested this, including `pip`; see https://ichard26.github.io/blog/2025/03/faster-pip-ci-on-windows-d-drive/ and https://github.com/pypa/pip/pull/13129, and have had convincing results. There are more details in the enclosed links.

We don't have a large enough sample size right now to see how much benefit we will get – the CI is noisy. However, we can get a coarse idea in a few days after merging this, when we can run the numbers with more data points. I have tried to get a sense of our Windows CI times over the past three months here: https://github.com/agriyakhetarpal/conda-ci-scripts (see the `notebooks/` and `data/` folders for more) with some dead-simple calculations. I attempted to test this on my fork here: https://github.com/agriyakhetarpal/conda/actions/runs/19772882689?pr=2.

Another approach is to create [virtual ReFS drives](https://learn.microsoft.com/en-us/windows-server/storage/refs/refs-overview) ([VHDs](https://learn.microsoft.com/en-us/windows-server/storage/disk-management/manage-virtual-hard-disks)), which I have not explored yet – it may or may not work for us. Only trying it out can reveal something conclusive. However, I would like to do so in a subsequent PR; in the meantime, this PR should provide us with some respite against our increasing CI time. For more details, see this PR in Astral's uv: https://github.com/astral-sh/uv/pull/3522 and the corresponding issue: https://github.com/astral-sh/uv/issues/3508

cc: @jaimergp, as this PR might be interesting for you :D

### Checklist - did you ...

- [x] Add / update necessary tests? ⚠️

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,7 @@ repos:
       # auto format Python codes within docstrings
       - id: blacken-docs
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.13
+    rev: v0.14.14
     hooks:
       # lint & attempt to correct failures (e.g. pyupgrade)
       - id: ruff-check
@@ -100,7 +100,7 @@ repos:
             tests/
           )
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.36.0
+    rev: 0.36.1
     hooks:
       # verify github syntaxes
       - id: check-github-workflows


agriyakhetarpal:
I have to note that the test case `test_trash_outside_prefix`, as defined below:

https://github.com/conda/conda/blob/1ed92caa83e6d845000facead21437e8755f8648/tests/test_install.py#L208-L216

is going to fail in CI here because it checks if `rm_rf()` can delete directories outside the conda prefix. On Windows, when `TEMP` is on D: and conda is on C:, `relpath()` can't compute a relative path between different drives, as they're inherently outside each other and will raise a `ValueError`.

I've gone ahead with monkeypatching `TEMP` for the scope of this test back to C: again, but if you have a better solution, please let me know!

jaimergp:
This is a good idea! I thought we had moved some stuff to D: already. The key point is to have _everything_ in the same drive, so conda can hard link files instead of copying things over. Root installation, pkgs cache, new environments, etc. That should give us a boost!

agriyakhetarpal:
> The key point is to have _everything_ in the same drive, so conda can hard link files instead of copying things over. Root installation, pkgs cache, new environments, etc. That should give us a boost!

Yes! This is similar to (maybe the same as?) the ReFS idea I've mentioned, for which we can use this action: https://github.com/marketplace/actions/setup-dev-drive. However, I'll prefer we do that in a subsequent PR, because:

1. I'd like to test how this performs over a week or so to see the actual results.
2. It actually caused a slight penalty for the pip test suite, and they eventually decided it wasn't worth the effort at the time: https://github.com/pypa/pip/pull/13123. So, I'd like to commit to _kaizen_ and make gradual, continuous improvements.

danyeaw:
It will be interesting to see some data that this indeed makes things go faster. I tried caching to the D:\ with setup-miniconda a few months ago, and although runner-to-runner variation is a thing, I found it to actually be slower. See https://github.com/conda-incubator/setup-miniconda/issues/406.

agriyakhetarpal:
Thanks for the extra context, @danyeaw. I wasn't aware that this had been tried before. In that case, I think one thing that I can do is to run the numbers on my conda fork by re-running the workflows a few times (5-7 successful runs should be fine, IMO), and that way we can have the data we need to draw a comparison without needing to exercise it in conda's CI.

jezdez:
## Windows CI Performance Analysis

I compared the PR's test run against a main branch run from the same time period to assess the impact of moving `TEMP` to the D: drive.

### Methodology
- **PR run**: 20228572542 (Dec 15, 2025) - WITH D: drive
- **Baseline**: 20347634605 (Dec 18, 2025) - main branch WITHOUT D: drive

### Results

#### Integration Tests: Significant Improvements ✅

| Job | D: drive | C: drive | Change |
|-----|----------|----------|--------|
| 3.10 defaults integration 1 | 21 min | 34 min | **-38%** |
| 3.10 defaults integration 2 | 23 min | 43 min | **-47%** |
| 3.10 defaults integration 3 | 21 min | 45 min | **-53%** |
| 3.13 defaults integration 1 | 18 min | 33 min | **-45%** |
| 3.13 defaults integration 2 | 21 min | 42 min | **-50%** |
| 3.13 defaults integration 3 | 20 min | 47 min | **-57%** |
| 3.13 conda-forge integration 1 | 34 min | 83 min | **-59%** |
| 3.13 conda-forge integration 2 | 31 min | 82 min | **-62%** |
| 3.13 conda-forge integration 3 | 34 min | 83 min | **-59%** |

**Average improvement for integration tests: ~50%**

#### Unit Tests: Mixed Results ⚠️

| Job | D: drive | C: drive | Change |
|-----|----------|----------|--------|
| 3.10 defaults unit 1 | 18 min | 21 min | -14% ✅ |
| 3.10 defaults unit 2 | 19 min | 17 min | +12% ❌ |
| 3.13 defaults unit 1 | 16 min | 19 min | -16% ✅ |
| 3.13 defaults unit 2 | 17 min | 17 min | 0% |
| 3.13 conda-forge unit 1 | 58 min | 51 min | **+14%** ❌ |
| 3.13 conda-forge unit 2 | 64 min | 57 min | **+12%** ❌ |

### Analysis

**The Good:**
- Integration tests show substantial speedups (40-60% faster), which aligns with the hypothesis that I/O-heavy operations benefit from the D: drive.

**The Concerning:**
- conda-forge unit tests are ~12-14% *slower* with D: drive. This may be related to @jaimergp's point about hard-linking - when conda is installed on C: but TEMP is on D:, cross-drive operations can't use hard links and must copy files instead.

**Additional Observation:**
- More recent runs on main (Jan 2026) show times similar to the D: drive PR, suggesting either GitHub runner improvements or significant run-to-run variance. More data points would help establish a clearer baseline.

### Recommendation

The PR shows clear benefits for integration tests. However:

1. **Investigate the conda-forge unit test regression** - worth understanding why these specific tests are slower
2. **Consider @jaimergp's suggestion** - moving the entire conda installation to D: (not just TEMP) could provide additional benefits by enabling hard-linking throughout the workflow
3. **Gather more data** - as @danyeaw noted, runner-to-runner variation is significant; merging this to collect more data points over time seems reasonable given the integration test improvements

Overall, the PR has merit and the integration test improvements are substantial enough to justify further exploration.

jezdez:
Let's see how this works with the current CI setup.

jezdez:
## Updated Windows CI Performance Analysis (After Merging Main)

Compared the latest run (Jan 26, 2026) after merging main with the December 15, 2025 run.

### Results

| Job | Dec 15 | Jan 26 | Change |
|-----|--------|--------|--------|
| 3.10 defaults unit 1 | 18 min | 17 min | -6% ✅ |
| 3.10 defaults unit 2 | 20 min | 14 min | **-31%** ✅ |
| 3.10 defaults integration 1 | 21 min | 21 min | ~ |
| 3.10 defaults integration 2 | 24 min | 25 min | +6% |
| 3.10 defaults integration 3 | 22 min | 24 min | +11% |
| 3.13 defaults unit 1 | 16 min | 14 min | -10% ✅ |
| 3.13 defaults unit 2 | 17 min | 14 min | **-21%** ✅ |
| 3.13 defaults integration 1 | 19 min | 20 min | +8% |
| 3.13 defaults integration 2 | 22 min | 23 min | +6% |
| 3.13 defaults integration 3 | 21 min | 22 min | +7% |
| 3.13 conda-forge unit 1 | 59 min | 26 min | **-56%** ✅ |
| 3.13 conda-forge unit 2 | 65 min | 24 min | **-63%** ✅ |
| 3.13 conda-forge integration 1 | 35 min | 31 min | -10% ✅ |
| 3.13 conda-forge integration 2 | 32 min | 32 min | ~ |
| 3.13 conda-forge integration 3 | 35 min | 34 min | -3% |

**Total Windows CI time: 424 min → 342 min (-19%)**

### Key Observations

**The big win**: The conda-forge unit tests that were previously *slower* with D: drive (compared to a different main baseline in my earlier analysis) are now **dramatically faster** - 56-63% improvement! This suggests:
1. The changes merged from main may have affected test behavior in a beneficial way
2. The combination of D: drive + current main state is working well together

**Mixed results for defaults integration tests**: The 3.13 defaults integration tests are now slightly slower (6-11%), while the December run showed them roughly on par. This could be normal run-to-run variance.

**Unit tests consistently faster**: All unit test jobs show improvements, ranging from 6% to 63%.

### Summary

After merging main, the overall picture is positive: **~19% reduction in total Windows CI time**. The dramatic improvement in conda-forge unit tests (from ~60min to ~25min each) is the biggest contributor. Worth continuing to monitor to see if these gains hold across multiple runs.

jezdez:
I would suggest marking this ready for review and then merging it into the main once the January release branch has been split off.

agriyakhetarpal:
Thanks @jezdez! I actually kind of left this PR for a while, and I think I ran it a few times on my fork in the first week of January, but later dropped the 🏀 on this. The Windows runners are failing to start today, tough luck. We should have enough data points after one month of merging this, IMO.

jezdez:
Agreed, I'm ready to merge this as is and readjust if needed.

tests/test_install.py
```diff
@@ -205,7 +206,17 @@ def test_binary(path_factory: PathFactoryFixture, subdir: str):
     assert tmp.read_bytes() == b"\x7fELF.../usr/local/lib/libfoo.so\0\0\0\0\0\0\0\0"
 
 
-def test_trash_outside_prefix():
+def test_trash_outside_prefix(monkeypatch: MonkeyPatch):
```
dholth, line 209: This test fails in the conda-libmamba-solver tests, which run conda's tests in a different environment. Does it pass on a local Windows machine?

PR #15447, opened 2025-11-26
Allow `conda activate` to spawn new shells if conda is not initialized

jaimergp:
### Description

- Closes https://github.com/conda-incubator/conda-spawn/issues/20
- Closes https://github.com/conda/conda/pull/15447

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15457, opened 2025-12-01
Keep parent shell environment

link89:
### Description

When starting a subshell from an activated conda environment, `CONDA_PREFIX` is lost because the init script reinitializes conda without preserving the parent's state.

This causes a problem: whenever I open a new shell session from an already-activated environment (e.g., py312)—such as launching screen, tmux, or opening VS Code from the terminal—the environment is reset back to base. As a result, I must manually reactivate the environment in every new shell.

I believe the conda initialization script should detect whether an environment is already active and, if so, automatically activate the same environment in the new shell.

**Solution**: Save `CONDA_PREFIX` before conda initialization, then reactivate after setup completes.

**Shells modified**:
- Bash/Zsh/Posix (Linux/macOS/Windows MSYS)
- Csh/Tcsh
- Fish (uses `set -g` for correct scoping)
- PowerShell
- Xonsh

Example init block (bash):
```bash
# >>> conda initialize >>>
__conda_env_keep="$CONDA_PREFIX"
__conda_setup="$('/path/to/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__conda_setup"
else
    # fallback logic...
fi
unset __conda_setup
[ -n "$__conda_env_keep" ] && conda activate "$__conda_env_keep"
unset __conda_env_keep
# <<< conda initialize <<<
```

### Checklist - did you ...

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1422,6 +1422,7 @@ def _config_fish_content(conda_prefix):
         """
         # >>> conda initialize >>>
         # !! Contents within this block are managed by 'conda init' !!
+        set -g __conda_env_keep "$CONDA_PREFIX"
         if test -f %(conda_exe)s
             eval %(conda_exe)s "shell.fish" "hook" $argv | source
         else
@@ -1431,6 +1432,10 @@ def _config_fish_content(conda_prefix):
                 set -x PATH "%(conda_prefix)s/bin" $PATH
             end
         end
+        if test -n "$__conda_env_keep"
+            conda activate "$__conda_env_keep"
+        end
+        set -e __conda_env_keep
         # <<< conda initialize <<<
         """
     ) % {
@@ -1566,6 +1571,7 @@ def _config_xonsh_content(conda_prefix):
         """
     # >>> conda initialize >>>
     # !! Contents within this block are managed by 'conda init' !!
+    __conda_env_keep = $CONDA_PREFIX if 'CONDA_PREFIX' in ${{...}} else None
     if !(test -f "{conda_exe}"):
         import sys as _sys
         from types import ModuleType as _ModuleType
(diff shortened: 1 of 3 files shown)


jaimergp:
Hm, I don't observe this in my local installation:

```bash
bash-3.2$ eval "$(conda shell.bash hook)"
(base) bash-3.2$ echo $CONDA_PREFIX
/Users/jrodriguez/.local/conda
(base) bash-3.2$ conda activate heavy
(heavy) bash-3.2$ echo $CONDA_PREFIX
/Users/jrodriguez/.local/conda/envs/heavy
# enter new shell
(heavy) bash-3.2$ bash
bash-3.2$ echo $CONDA_PREFIX
/Users/jrodriguez/.local/conda/envs/heavy
```

The PS1 var is lost but that's known (PS1 is not an exported var ever for... reasons).

link89:
@jaimergp Here is how it happens on WSL:

image


It is not just lost PS1, but the python executable is set back to base
And in .bashrc it is:

```bash
# >>> conda initialize >>>
__conda_setup="$('/home/henry/anaconda3/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__conda_setup"
else
    if [ -f "/home/henry/anaconda3/etc/profile.d/conda.sh" ]; then
        . "/home/henry/anaconda3/etc/profile.d/conda.sh"
    else
        export PATH="/home/henry/anaconda3/bin:$PATH"
    fi
fi
unset __conda_setup
# <<< conda initialize <<<
```

jaimergp:
Can you try setting `auto_activate` to false?

```
conda config --set auto_activate false
```

travishathaway:
@conda-bot check

link89:
@jaimergp I show both without and with `auto_activate false` in the screenshot, it do solve the problem, but the ps1 just lost which makes it looks confusing.
image

jaimergp:
PS1 being dropped is kind of expected because this var is not exported. We are looking into other ways of providing activation that would be more fitting to your use case:

- [`conda spawn`](https://github.com/conda-incubator/conda-spawn) creates new shell processes by default, and doesn't require initialization. This is a plugin that you need to install separately.
- I'm trying to bring this over to `conda activate` in #15447 

The problem with the logic proposed here is that it would conflict with other established use cases, where new shells are expected to start afresh. So, given that there are solutions and workarounds for your current situation, I'm inclined to reject this PR.

Let me know if the solutions I proposed are sufficient and, if so, we can add it to our documentation so other users know what to do in they find themselves in the same situation.

Thanks for understanding!

PR #15470, opened 2025-12-04
use explicit utf-8 encoding to read, write cached repodata and state

### LINKED ISSUE #15461: Call .read_text(), .write_text(), open() with encoding='utf-8' instead of system default encoding for repodata
### Checklist

### What happened?

In https://github.com/conda/conda/pull/15396 @Rich-AM noticed that `repodata.json` could be written, read in the system default encoding instead of utf-8 and that this could fail.

Instead, these, and probably all json in conda, should be read and written in utf-8. It's not necessary to detect and use any other encoding for these files.

Find all text-mode access to repodata.json and <cache key>.info.json, and pass an explicit utf-8 encoding.

### Conda Info

```shell
25.11
```

dholth:
### Description

Fix #15461 

May need a test; on the other hand, this is what we /implicitly/ do on most platforms.

### Checklist - did you ...


kenodegard:
Does this mean we should enable [PLW1514](https://docs.astral.sh/ruff/rules/unspecified-encoding/#unspecified-encoding-plw1514) ruff linting rule to ensure all reads set the encoding?

dholth:
This PR means to limit itself to cached .json. Those files certainly should be in utf-8 only; I did not audit the code for any files that may have to be read/written in other encodings.

conda/gateways/repodata/__init__.py
```diff
@@ -593,12 +593,14 @@ def save(self, data: str | bytes):
...
             target = self.cache_path_shards
+            encoding = {}  # none for binary mode
         else:
             mode = "x"
             target = self.cache_path_json
+            encoding = {"encoding": "utf-8"}  # explicit utf-8 for text mode
 
         try:
-            with temp_path.open(mode) as temp:  # exclusive mode, error if exists
+            with temp_path.open(mode, **encoding) as temp:  # type: ignore ; exclusive mode, error if exists
```
kenodegard, line 603: It'll be cleaner to pass `encoding=None` for binary mode (which means we can remove `# type: ignore`):

```suggestion
            encoding = None  # none for binary mode
        else:
            mode = "x"
            target = self.cache_path_json
            encoding = "utf-8"  # explicit utf-8 for text mode

        try:
            with temp_path.open(mode, encoding=encoding) as temp:  # exclusive mode, error if exists
```

dholth: Good catch

jaimergp: Was this applied or not? 🤔

dholth: It was applied.

jezdez:
Two thoughts:

- For existing cached repodata and state that may have been saved with a different encoding, what will be the migration path?
- Hardcoding the encoding this way in a few places eventually will mean some evolution of this code is going to forget the encoding parameter. Could we refactor this to have open/read/write available from a central object, so there is only one place where we need to define the encoding?

dholth:
It may already ignore and rewrite the metadata if there is a problem, which is what we would want for both.

jaimergp:
> Hardcoding the encoding this way in a few places eventually will mean some evolution of this code is going to forget the encoding parameter.

Agreed. At the very least this could be a global constant. Maybe also aliased as a class attribute in case there's a need for customization in subclasses.

dholth:
At least it will be the default in Python 3.15

jezdez:
@dholth I've opened https://github.com/conda/conda/pull/15600 as a proposal for my comments above, what do you think?

PR #15376, opened 2025-11-03
Expose configuration management interface

vshevchenko-anaconda:
### Description

Refactors conda config to make core logic available to other Python interfaces by extracting it into reusable classes. These classes can now be imported and used programmatically, not just through the CLI, enabling other parts of conda or third-party tools to manipulate configuration files without shelling out to conda config. The original implementation logic was preserved with minimal modifications to avoid behavioral changes.

Fix #15238.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #15473, opened 2025-12-05
Add 'channel_settings' to the MockContext

soapy1:
### Description

anaconda auth introduced using the `channel_settings` config from the context in https://github.com/anaconda/anaconda-auth/pull/175.

So, when running tests, the mock context must be able to provide this configuration. Otherwise, it will fail with the error

```
AttributeError: 'MockContext' object has no attribute 'channel_settings'
```

ref test failures in https://github.com/conda/conda/actions/runs/19971010416/job/57281147878?pr=15439

### Checklist - did you ...

~- [ ] Add / update necessary tests?~
~- [ ] Add / update outdated documentation?~


dholth:
I was just wondering how to control anaconda auth plugin from tests, this is great.

dholth (approved):

mcg1969 (commented):
Thank you for this. We added the ability for anaconda-auth to configure the "default" channel settings when installed.

tests/conftest.py
```diff
@@ -293,6 +293,7 @@ def __init__(self, *args, **kwargs):
             self.plugin_manager = mocker.MagicMock()
             self.repodata_fns = ["repodata.json", "current_repodata.json"]
             self.subdir = mocker.MagicMock()
+            self.channel_settings = []
```
soapy1, line 296: It would maybe be better if the `MockContext` more closely resembled the context object. That way changes in downstream dependencies would not cause these types of errors. Right now, this bug is causing test failures on the main branch, so I think this quick fix is appropriate. If there is interest, I can spin out an issue to dig into a better solution.

mattkram:
Thanks for this. FYI the auto-configuration behavior can be disabled via config or environment variable, which might help: https://github.com/anaconda/anaconda-auth/pull/181.

Out of curiosity, where is the config that's putting `anaconda-auth` into the conda CI? I'm glad for the feedback loop but just want to be aware of interdependencies here too :).

soapy1:
> Thanks for this. FYI the auto-configuration behavior can be disabled via config or environment variable, which might help: [anaconda/anaconda-auth#181](https://github.com/anaconda/anaconda-auth/pull/181).
> 
> Out of curiosity, where is the config that's putting `anaconda-auth` into the conda CI? I'm glad for the feedback loop but just want to be aware of interdependencies here too :).


I found that the anaconda-auth plugin gets installed by inspecting the [github actions logs (conda install section)](https://github.com/conda/conda/actions/runs/19974273591/job/57286440641?pr=15473)

You can see:
```

    anaconda-auth-0.12.0       |  py313h06a4308_0         103 KB
```

It looks like the `anaconda-client` package on the defaluts channel (though not the package that is on conda-forge) includes it as a dependency. For example:
```
$ conda create -n testenvsdf anaconda-client -c defaults --override-
channels
. . .

. . .
    anaconda-auth-0.12.0       |  py313h06a4308_0         103 KB
. . .
```

jezdez (changes requested):
Wait a second, I have major concerns with the way https://github.com/anaconda/anaconda-auth/pull/175 overwrites conda's context attributes, `context.channel_settings`, post-processing conda context attributes after startup basically, even though the user expects the config to be loaded in an idempotent way. The change in anaconda-auth essentially overtakes _all_ channel settings, which is not appropriate for the separation of concerns.

An appropriate way could be either:

- Introducing a new plugin hook that allows us to manage context settings safely via plugins (e.g. for self-declaration of channel settings, or maybe a `config_defaults` plugin hook that requires confirmation upon installation)
- updating a `condarc` file via the newly introduced config file management API in #15376 as contributed from another Anaconda team (since config files will be validated during startup of conda and prevent misconfiguration), e.g. via pre_command hooks, or post install scripts

While we're discussing how anaconda-auth will progress forward, this PR is on hold.

mcg1969:
I think we can navigate our disagreement about the anaconda-auth implemention in good faith, but I would say that blocking the continued development of conda because of the `anaconda-auth` issue is not a wise way forward. So while this particular PR might not be the best resolution, some mechanism that routes around the current build of `anaconda-auth` is likely wise.

jezdez:
@mcg1969 you're misunderstanding my reasons: this was a response to seeing a pull request like this being proposed on a Friday evening without prior conversations, since we _relaxed_ the review policy recently to only require one reviewer.

On review I found the change in anaconda-auth that triggered it particularly problematic since it was overwriting conda internals that are outside the scope of the plugin API, as noted in https://github.com/anaconda/anaconda-auth/pull/175/files#r2594060997.

As a senior maintainer of the conda project, I have that right, and no Anaconda-specific need overrules the policies the @conda/conda-maintainers team has set for itself.

tests/conftest.py
jezdez: Yeah, let's spin this out in a new ticket since you're right that downstream changes to the context should not break our conda CI.

That said, this was a deliberate decision to overwrite conda internals in https://github.com/anaconda/anaconda-auth/pull/175 due to a lack of alternatives that were readily available. We see now that it takes a lot more effort per https://github.com/anaconda/anaconda-auth/pull/192 and https://github.com/anaconda/anaconda-auth/pull/193.

Personally, I would even call into question the ability to modify context values after instantiation outside the `plugins` namespaces and would prefer if we considered a plugin hook (as we discussed a few times in the past) that would allow plugins to provide a default set of config values that would also include a user confirmation step. Think printing something like `Note: The 'anaconda-auth' conda plugin preconfigures your conda distribution for the following conda settings: channels, channel_settings, ssl_verify, auto_update_conda` (or similar). Basically a plugin hook that would allow plugins to mass-set config values via the config API.

soapy1:
superseded by https://github.com/conda/conda/pull/15476

PR #15476, opened 2025-12-05
Remove anaconda-client from test dependencies

soapy1:
### Description
anaconda-client is not required for running the test suite. Removing this dependency will also stop the anaconda-auth plugin from being pulled into test runs.

alternative to https://github.com/conda/conda/pull/15473

### Checklist - did you ...

### DIFF
--- a/tests/core/test_index.py
+++ b/tests/core/test_index.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import copy
-import platform
 from contextlib import nullcontext
 from logging import getLogger
 from typing import TYPE_CHECKING


kenodegard (approved):
Ah, a likely leftover from when the upload to anaconda.org was explicitly implemented here in this repo (instead of using the canary GH action).

tests/core/test_index.py
```diff
@@ -56,34 +56,40 @@
         "version": "2.3.9",
         "build": "py35_0",
         "build_number": 0,
+        "subdir": "linux-64",
```
soapy1: package records are compared based on the `_pkey` that includes the channel name, subdir, name, version, build number, build, and sometimes fn of the package. My intuition is that filling in the rest of these pkey values would make the tests pass, but the linux-64 tests still seem to fail. I have also confirmed that this package exists in the index (when testing locally). So, I'm not sure why this test is failing.

jezdez:
Tried a different approach with #15497 to get the `test_get_index_lazy` to work, @jaimergp will merge this into this PR

PR #15482, opened 2025-12-06
Add instructions for manual dev environment setup

danyeaw:
### Description

In #15273, we discussed creating manual dev environment instructions instead of adding automated scripts for other shells. This PR adds these instructions. Some decisions I took which are open to change to simplify the instructions:

- Only use Miniforge and conda-forge
- Simplify the dev env structure so it only installs in ./devenv instead of a nested structure based on OS, arch, and Python version. I think most of the time we just need one environment for local testing
- Use PowerShell for Windows, since I don't think someone would choose to manually do development on cmd.exe
- Make conda-pypi optional - doing an editable install is a pretty nice way to go, but it can be optional until it graduates to a conda project

### Checklist - did you ...

### DIFF
--- a/news/15482-manual-dev-env-instructions
+++ b/news/15482-manual-dev-env-instructions
@@ -12,7 +12,7 @@
 
 ### Docs
 
-* Add manual development environment setup instructions
+* Add manual development environment setup instructions. (#15482)
 
 ### Other
 


travishathaway:
I’ll be happy to test this later on my Windows workstation to make sure everything works. I was having an extreme amount of difficulty getting the dev scripts working so this addition seems like a great idea.

travishathaway:
One random thought I had as I was going through this is how annoying it is to have to manually copy and paste the commands. Maybe it's time to add something like this?

- https://sphinx-copybutton.readthedocs.io/en/latest/

docs/source/dev-guide/development-environment.md
```diff
@@ -161,6 +161,112 @@ $ source ./dev/start -i miniforge
...
+   $ mkdir -p ./devenv
+   ```
+
+2. Download and install miniforge:
+
+   ````{tab-set}
+
+   ```{tab-item} macOS
+   ```bash
+   $ curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-$(uname -m).sh -o ./devenv/miniforge.sh
```
travishathaway:
```suggestion
   $ curl -L "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-$(uname -m).sh" -o ./devenv/miniforge.sh
```

docs/source/dev-guide/development-environment.md
```diff
@@ -161,6 +161,112 @@ $ source ./dev/start -i miniforge
...
+   $ bash ./devenv/miniforge.sh -bfp ./devenv
+   ```
+
+   ```{tab-item} Linux
+   ```bash
+   $ curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-$(uname -m).sh -o ./devenv/miniforge.sh
```
travishathaway:
```suggestion
   $ curl -L "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-$(uname -m).sh" -o ./devenv/miniforge.sh
```

docs/source/dev-guide/development-environment.md
travishathaway: It's safer to put double quotes around the URL, especially when were invoking a sub-shell call with `$(...)`

travishathaway: This was the error I got with zsh after copying and pasting:

```
▶ curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-$\(uname -m).sh -o ./devenvtest/miniforge.sh
zsh: parse error near `)'
```

travishathaway:
I tested this on my mac and everything works ✅  (I used option A thought and not `conda pypi`)

danyeaw:
I think the Windows instructions need a few tweaks, let me update those before we merge it.

kenodegard:
> Simplify the dev env structure so it only installs in ./devenv instead of a nested structure based on OS, arch, and Python version. I think most of the time we just need one environment for local testing

True, especially if this is done manually. The nested structure was added to support macOS x86 vs ARM devenvs and for those using VMs with shared filesystems between the host and VM. Personally I've removed the need for much of this nested structure with my devenv living in my home directory and not within the repo itself. 🤷🏼‍♂️

jezdez:
> > Simplify the dev env structure so it only installs in ./devenv instead of a nested structure based on OS, arch, and Python version. I think most of the time we just need one environment for local testing
> 
> True, especially if this is done manually. The nested structure was added to support macOS x86 vs ARM devenvs and for those using VMs with shared filesystems between the host and VM. Personally I've removed the need for much of this nested structure with my devenv living in my home directory and not within the repo itself. 🤷🏼‍♂️

Should we default to that pattern?

news/15482-manual-dev-env-instructions
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* Add manual development environment setup instructions
```
jezdez:
```suggestion
* Add manual development environment setup instructions. (#15482)
```

jezdez (approved):

PR #15490, opened 2025-12-08
Cache reading environment spec file contents in core conda (opposed to in plugin implementation)

soapy1:
### Description

This change updates the environment spec api so that both the filename and data contained in the file are made available to the plugin implementation through the `self.filename` and `self.data` variables. It also clarifies these variables as part of the api for initializing the plugin.

This will allow for core conda to deal with reading the environment spec file. So core conda (and not plugin authors) is reponsible for:
* downloading the file if it is required
* caching reading the contents of the file to reduce io

The plugin author is reponsible for:
* converting the raw data into whatever format is relevant to the plugin (eg. yaml, json, dict, etc.)

#### Other changes

* Updating built in environment spec plugins to use the `data` field provided by the api
* Depcreation of `conda.env.env.load_file` - users should use `conda.common.io.load_file` instead

#### Performance improvement

* TODO: update these after resolving all code review

This increases the speed for creating/updating environments with the `conda env` command. For example:

On this branch
```
(env-spec-reduce-io) $ hyperfine 'conda env create -p /tmp/testenv --file tests/env/support/simple.yml' -r 10 -w 1 -c 'rm -rf /tmp/testenv'                                                       
Benchmark 1: conda env create -p /tmp/testenv --file tests/env/support/simple.yml
  Time (mean ± σ):      1.875 s ±  0.035 s    [User: 1.422 s, System: 0.317 s]
  Range (min … max):    1.841 s …  1.932 s    10 runs
```

On main
```
(main) $ hyperfine 'conda env create -p /tmp/testenv --file tests/env/support/simple.yml' -r 10 -w 1 -c 'rm -rf /tmp/testenv'                                                                     
Benchmark 1: conda env create -p /tmp/testenv --file tests/env/support/simple.yml
  Time (mean ± σ):      3.212 s ±  0.575 s    [User: 2.666 s, System: 0.499 s]
  Range (min … max):    2.500 s …  4.548 s    10 runs
```

ref: https://github.com/conda/conda/issues/15139

### Checklist - did you ...

### DIFF
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -572,15 +572,16 @@ class EnvironmentSpecBase(ABC):
     # autodetection is enabled.
     detection_supported: ClassVar[bool] = True
 
-    def __init__(self, filename: str, data: str | None = None):
+    def __init__(self, filename: str | None, data: str | None = None):
         """
         :param data: the contents of the environment spec file
         :param filename: the full path to the environment spec file
         """
         self.filename = filename
-        if data is None:
-            data = load_file(self.filename)
-        self.data = data
+        if data is None and self.filename:
+            self.data = load_file(self.filename)
+        else:
+            self.data = None
 
     @abstractmethod
     def can_handle(self) -> bool:
--- a/tests/env/specs/test_explicit.py
+++ b/tests/env/specs/test_explicit.py
@@ -7,6 +7,11 @@ from conda.models.environment import Environment
 from .. import support_file
 
 
+def test_no_environment_file():
+    spec = ExplicitSpec(filename=None)
+    assert not spec.can_handle()
+
+
 def test_can_handle_explicit():
     spec = ExplicitSpec(filename=support_file("explicit.txt"))
     assert spec.can_handle()
--- a/tests/env/specs/test_requirements.py
+++ b/tests/env/specs/test_requirements.py
@@ -8,6 +8,11 @@ from conda.models.environment import Environment
 from .. import support_file
 
 
+def test_no_environment_file():
+    spec = RequirementsSpec(filename="not-a-file")
+    assert not spec.can_handle()
+
+
 def test_no_name():
     spec = RequirementsSpec(filename=support_file("requirements.txt"))
     assert spec.can_handle()
--- a/tests/plugins/test_env_specs.py
+++ b/tests/plugins/test_env_specs.py
@@ -8,7 +8,6 @@ import pytest
 
 from conda import plugins
 from conda.exceptions import (
-    CondaFileIOError,
     CondaValueError,
     EnvironmentSpecPluginNotDetected,
     PluginError,
@@ -17,7 +16,7 @@ from conda.models.environment import Environment
 from conda.plugins.types import CondaEnvironmentSpecifier, EnvironmentSpecBase
 
 if TYPE_CHECKING:
-    from pytest_mock import MockerFixture
+    from pytest import MonkeyPatch
 
 
 class EmptySpec(EnvironmentSpecBase):
@@ -132,21 +131,13 @@ def naughty_spec_plugin(plugin_manager):
 
 
 @pytest.fixture()
-def mock_empty_load_file(mocker: MockerFixture):
+def mock_empty_load_file(monkeypatch: MonkeyPatch):
     """
     Fixture to mock out loading an environment spec file in the plugin manager.
     This will return an empty string regardless of if the file passed into
     `conda.plugins.manager.load_file` exists or not. Helpful for testing.
     """
-    mocker.patch(
-        "conda.plugins.manager.load_file",
-        return_value="",
-    )
-
-
-def test_no_environment_file():
-    with pytest.raises(CondaFileIOError):
-        EmptySpec(filename="")
+    monkeypatch.setattr("conda.plugins.manager.load_file", lambda x: "")
 
 
 def test_dummy_random_spec_is_registered(


conda/common/io.py
```diff
@@ -31,6 +31,28 @@
 IS_INTERACTIVE = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
 
+@cache
```
soapy1: I'm not entirely sure if this is ok.

soapy1:
pre-commit.ci autofix

conda/common/io.py
soapy1: This assumes that the contents of a file will not change during a single conda run. I think this is a safe assumption to make in the context of running conda as a cli. 

Howerver, some tests will update environment files and re-run conda multiple times. In these cases, the tests will need to invalidate this cache so changes to the files get picked up.

I think this is ok. Certainly open to other perspectives on this!

jezdez:
Hey! Had a chance to look through this - the idea of centralizing file reading in core conda makes a lot of sense, nice approach.

A few things caught my eye while reading through:

I noticed there might be a type mismatch in the new `load_file` in `common/io.py` - when reading from disk it returns bytes (`f.read()`) but `download_text()` probably returns a string? The docstring says "string" so just wanted to flag that. The original `env.load_file` does the UTF-8/UTF-16 decoding dance which might be needed here too.

Also - and I might be missing something here - but it looks like there are now two `load_file` functions? The new cached one in `common/io.py` and the original in `env/env.py`. The `can_handle()` methods in cep_24.py and yaml_file.py still seem to call `env.load_file()` which isn't cached, so they'd still hit the disk separately from when the base class loads `self.data`. Same deal with explicit.py and requirements.py using `yield_lines(self.filename)` directly instead of `self.data`.

One other thing - the `@cache` decorator is unbounded so memory could grow over time. Maybe `@lru_cache(maxsize=128)` or something? The test helper calls `cache_clear()` which is good but production code might want some invalidation strategy too eventually.

Small thing: the docstring in the new `__init__` says `:param file_path:` but the actual param is `filename`.

Oh and the news file still has some `<news item>` placeholders in it.

For plugin authors - this does change the base class signature (filename is now required, and it auto-loads the file). Might be worth noting somewhere for anyone with custom env spec plugins.

Anyway, looks like good progress on #15139!

conda/common/io.py
ryanskeith: In the regular case I think this is true as well.

conda/env/specs/yaml_file.py
ryanskeith, line 1: Same here as for the YamlFileSpec, the can_handle function rereads the file and should probably use the self.data field.

conda/plugins/manager.py
```diff
@@ -582,7 +582,7 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
             if hook.environment_spec.detection_supported:
                 log.debug("EnvironmentSpec hook: checking %s", hook_name)
                 try:
-                    if hook.environment_spec(source).can_handle():
+                    if hook.environment_spec(filename=source).can_handle():
```
ryanskeith: This is where the can_handle rereads hurt the most.  This is also where the code gains the benefit of the caching.  Maybe it could be more elegant to read the data file data once and feed it to the spec?  I am on the fence about this one, tbh.

conda/env/specs/cep_24.py
ryanskeith, line 1: Because the init of the child class was moved to the parent, the class is now reading the file and storing the contents in self.data.  This means in can_handle, we are re-reading the file again.  This is every time in manager.py see [here](https://github.com/conda/conda/blob/2eae53a802e7c520598fbdf32875211ddf994f35/conda/plugins/manager.py#L551).  The can_handle function need to adopt the same mechanism of reading from the data like the change in the env/environment property.

conda/env/specs/explicit.py
ryanskeith, line 1: This ExplicitSpec class is re-reading the file in the can_handle function via `yield_lines`.  We should pivot to using self.data as its source.

conda/env/specs/requirements.py
ryanskeith, line 1: This RequirementSpec class is re-reading the file in the can_handle function via yield_lines. We should pivot to using self.data as its source.

conda/plugins/manager.py
soapy1: hmmm, I think you are right. By passing the data in directly we can also skip on doing a lookup in the cache. Good suggestion, thanks!

conda/common/io.py
```diff
@@ -31,6 +32,26 @@
...
 
 
+@lru_cache(maxsize=128)
+def load_file(filename: str) -> str:
+    """Load and return a string from a given file"""
+    from ..exceptions import CondaFileIOError
+    from ..gateways.connection.download import download_text
+    from ..gateways.connection.session import CONDA_SESSION_SCHEMES
+
+    url_scheme = filename.split("://", 1)[0]
```
jezdez: I tend to want to use urlparse for this, to reduce manual URL parsing, I know we're using the same pattern elsewhere

conda/common/io.py
```diff
@@ -31,6 +32,26 @@
 IS_INTERACTIVE = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
 
+@lru_cache(maxsize=128)
+def load_file(filename: str) -> str:
```
jezdez, line 37: This looks like `conda.env.env.load_file`, but the behavior isn't similar since it raises a different exception, can we document that?

conda/common/io.py
```diff
@@ -31,6 +32,26 @@
...
+    url_scheme = filename.split("://", 1)[0]
+    if url_scheme in CONDA_SESSION_SCHEMES:
+        file_contents = download_text(filename)
+    elif not os.path.exists(filename):
+        raise CondaFileIOError(
+            filepath=filename,
+            message="File does not exist on disk",
+        )
+    else:
+        file_contents = Path(filename).read_text()
```
jezdez: Does this do the same as the old `load_file` to handle Unicode decode errors?

conda/env/specs/requirements.py
```diff
@@ -31,16 +30,16 @@ class RequirementsSpec(EnvironmentSpecBase):
 
     @deprecated.argument("24.7", "26.3", "name")
     def __init__(
-        self, filename: str | None = None, name: str | None = None, **kwargs
+        self, filename: str, data: str | None = None, name: str | None = None, **kwargs
```
jezdez: Is this backwards compatible?

conda/env/specs/requirements.py
```diff
@@ -101,7 +100,9 @@ def can_handle(self) -> bool:
 
         # Ensure this is not an explicit file. Requirements.txt and explicit files
         # may sometimes share file extension.
-        dependencies_list = list(yield_lines(self.filename))
+        dependencies_list = [
+            dep for dep in self.data.split("\n") if (dep and not dep.startswith("#"))
+        ]
```
jezdez, line 106: Why does this drop using yield_lines?

conda/env/specs/requirements.py
```diff
@@ -121,7 +122,9 @@ def environment(self) -> EnvironmentYaml:
             raise CondaValueError("No filename provided")
 
         # Convert generator to list since Dependencies needs to access it multiple times
-        dependencies_list = list(yield_lines(self.filename))
+        dependencies_list = [
+            dep for dep in self.data.split("\n") if (dep and not dep.startswith("#"))
+        ]
         return EnvironmentYaml(
```
jezdez, line 129: same

conda/env/specs/requirements.py
```diff
@@ -138,8 +141,11 @@ def env(self) -> Environment:
         if not self.filename:
             raise ValueError("No filename provided")
 
-        # Convert generator to list since Dependencies needs to access it multiple times
-        dependencies_list = list(yield_lines(self.filename))
+        # Get all the dependencies from the file, excluding lines that have
+        # been excluded by being prefixed with "#"
+        dependencies_list = [
+            dep for dep in self.data.split("\n") if not dep.startswith("#")
+        ]
```
jezdez, line 149: same

conda/env/env.py
```diff
@@ -333,7 +335,7 @@ def load_file(filename):
 
 def from_file(filename):
     """Load and return an ``EnvironmentYaml`` from a given file"""
-    yamlstr = load_file(filename)
+    yamlstr = io.load_file(filename)
     return from_yaml(yamlstr, filename=filename)
```
jezdez, line 349: Should we also deprecate this?

conda/plugins/hookspec.py
jezdez, line 690: This is an API change that _might_ create issues for implementators (I have no way to know if there are any)

conda/plugins/hookspec.py
```diff
@@ -681,9 +687,6 @@ def conda_environment_specifiers(self) -> Iterable[CondaEnvironmentSpecifier]:
             class RandomSpec(EnvironmentSpecBase):
                 extensions = {".random"}
 
-                def __init__(self, filename: str):
-                    self.filename = filename
-
                 def can_handle(self):
```
jezdez, line 690: This API change _might_ be a problem for implementers; I have no way to know how many there are

conda/plugins/types.py
```diff
@@ -571,6 +572,16 @@ class EnvironmentSpecBase(ABC):
...
+    def __init__(self, filename: str, data: str | None = None):
+        """
+        :param data: the contents of the environment spec file
+        :param filename: the full path to the environment spec file
+        """
+        self.filename = filename
+        if data is None:
+            data = load_file(self.filename)
+        self.data = data
+
```
jezdez: We actually thought to keep the plugin types lean and implement application logic elsewhere

tests/env/specs/test_explicit.py
```diff
@@ -7,11 +7,6 @@
 from .. import support_file
 
 
-def test_no_environment_file():
-    spec = ExplicitSpec()
-    assert not spec.can_handle()
-
-
```
jezdez, line 14: Hm, why do we remove this?

tests/env/specs/test_requirements.py
```diff
@@ -8,11 +8,6 @@
 from .. import support_file
 
 
-def test_no_environment_file():
-    spec = RequirementsSpec(filename="not-a-file")
-    assert not spec.can_handle()
-
-
 def test_no_name():
```
jezdez, line 16: same

tests/env/test_env.py
```diff
@@ -524,3 +526,13 @@ def test_prefix_validation(prefix, error_message):
...
+
+
+@pytest.mark.parametrize(
+    "function,raises",
+    [("load_file", TypeError)],
+)
+def test_env_deprecations(function: str, raises: type[Exception] | None) -> None:
+    raises_context = pytest.raises(raises) if raises else nullcontext()
+    with pytest.deprecated_call(), raises_context:
+        getattr(env_module, function)()
```
jezdez, line 545: If there are no more, we can keep this unparameterized.

tests/plugins/test_env_specs.py
```diff
@@ -111,7 +131,27 @@ def naughty_spec_plugin(plugin_manager):
...
+def mock_empty_load_file(mocker: MockerFixture):
+    """
+    Fixture to mock out loading an environment spec file in the plugin manager.
+    This will return an empty string regardless of if the file passed into
+    `conda.plugins.manager.load_file` exists or not. Helpful for testing.
+    """
+    mocker.patch(
+        "conda.plugins.manager.load_file",
+        return_value="",
+    )
```
jezdez: For mocks that don't require the magic of mock, I prefer to just use the monkeypatch fixture.

```suggestion
def mock_empty_load_file(monkeypatch):
    """
    Fixture to monkeypatch loading an environment spec file in the plugin manager.
    This will return an empty string regardless of if the file passed into
    `conda.plugins.manager.load_file` exists or not. Helpful for testing.
    """
    monkeypatch.setattr("conda.plugins.manager.load_file", lambda: "")
```

tests/plugins/test_env_specs.py
```diff
@@ -1,20 +1,38 @@
...
+    CondaFileIOError,
     CondaValueError,
     EnvironmentSpecPluginNotDetected,
     PluginError,
 )
 from conda.models.environment import Environment
 from conda.plugins.types import CondaEnvironmentSpecifier, EnvironmentSpecBase
 
+if TYPE_CHECKING:
+    from pytest_mock import MockerFixture
```
jezdez: Probably not needed if we switch to monkeypatch?

conda/common/io.py
soapy1: ah good idea. Looks like `Path.read_text` doesn't deal with unicode errors. Can add!

conda/common/io.py
soapy1: added to the deprecation addendum

conda/env/specs/requirements.py
soapy1: This suggests not using `yield_lines` because that function takes a file path and opens the file in order to produce lines. 

The goal of this PR is to reduce the io overhead of running these plugins by reading the data from the files once and then passing around that data to each plugin.

The most io intensive part is plugin detection. Previosly, each plugin's `can_handle` implementation would re-read the provided environment file. So, as more plugins are added we might expect a decrease in performance. 

Since this part of the code is not run during plugin detection, it could be ok to re-introduce using `yield_lines` here. Another option might be to introduce a new piece of funcitonality that has the same behaviour as `yield_lines` but takes in a string representing some file's contents and yields them. Maybe like a `yield_lines_from_data` (?), or modifying `yield_lines` to be `yield_lines(path: None | str = None, data: None | str = None)`.  

I think `yield_lines_from_data` is a bit of a mouthful, but might be a nice solution here. What do you think?

conda/env/specs/requirements.py
soapy1: good point, switched the arguments around so that `name` is still the 2nd argument.

conda/env/env.py
soapy1: I think it's good to deprecate this. It's only used in tests now :+1:

conda/plugins/hookspec.py
soapy1: I think this is a valid concern. Also, this concern also outlines the need for resolving https://github.com/conda/conda/issues/14895, to provide a way for having experimental public facing api's that don't need to go thru the deprecation policy. 

Note, this api is marked as being [`**EXPERIMENTAL**`](https://github.com/conda/conda/pull/15490/changes#diff-6556fdfa621b4832769318604b446edab8943f0f09bb1e34e4b1a71f8b939d48L658) including in the [docs](https://docs.conda.io/projects/conda/en/stable/dev-guide/plugins/environment_specifiers.html#conda.plugins.hookspec.CondaSpecs.conda_environment_specifiers).

## Thoughts about this

I think this api change is required and plugin authors may be required to make small changes in order to conform to this more explicit api.

### TL;DR

This change:
* formalizes the implicit `__init__` argument that represents the file name of the environment spec file
* will break plugins that have chosen different names for `filename`/`data`
* breaks [conda-lockfiles](https://github.com/conda-incubator/conda-lockfiles)

###  `filename` is the name of the file provided

Currently, it is implied that implementing plugins should have an `__init__` function that accepts the filename of the environment file. For example, whenever an environment spec is instantiated the filename for the environment file is passed in like `plugin.environment_spec(source)`. You can see this in a few places, for example
* https://github.com/conda/conda/blob/main/conda/plugins/manager.py#L551
* https://github.com/conda/conda/blob/main/conda/plugins/manager.py#L585

This means that working plugins are already accepting this argument (though we can have no knowledge of what they are doing with it).  I think it is good to document this more clearly by declaring that it is part of the api.

### `data` is the string representation of the file contents. If not provided, the `data` variable will automatically be populated from the contents of the file specified by `filename`

`data` in not a required argument. So, plugins that are currently not accepting this as an argument do not need to. However, given the plugin manager change to pass in both `filename` and `data` when instantiating a plugin ([`hook.environment_spec(filename=source, data=data).`](https://github.com/conda/conda/pull/15490/changes#diff-7d861030d2e81b5082cd0b27c9af9b3b18ff05031d0e20184da093fb4a710f8eR586) we might end up in a situation where the plugin produces an error about this argument not being defined if a plugin's `__init__` function  looks like:
*  `def __init__(self, source)`: This will produce an error about `data` not being definied when the line `hook.environment_spec(filename=source, data=data)` runs.
*  `def __init__(self, source, something: None | Any = None)`: Might produce an error because `something` will be populated with unexpected information. Though, it's important to note that `something` is not populated by conda in the latest released version.
* `def __init__(self, source, **kwargs)`: This should work ok, maybe 

### Passing named argumented eg. `hook.environment_spec(filename=source, data=data)`

Moving from positional arguments `hook.environment_spec(source)`  to named arguments `hook.environment_spec(filename=source, data=data)` will cause an issue with plugins that have implemented their own `__init__` functions with variables that represent the same thing as `filename`/`data` but are named differently. For example, [conda-lockfiles accepts a `path`](https://github.com/conda-incubator/conda-lockfiles/blob/main/conda_lockfiles/conda_lock/v1.py#L246).

conda/plugins/types.py
soapy1: Thinking about this concern + comparing this plugin to `CondaEnvironmentExporter` plugin, is it worth it to consider moving away from this `EnvironmentSpecBase` class and instead restructure the `CondaEnvironmentSpecifier` to look more like:

```
@dataclass
class CondaEnvironmentSpecifier(CondaPlugin):
    name: str
    can_handle: Callable[[PathType, str], bool]
    env: Callable[[PathType, str], Environment]
    detection_supported: bool
```

tests/env/specs/test_explicit.py
soapy1: letftover from refactor. Pulled back in!

tests/env/specs/test_requirements.py
soapy1: letftover from refactor. Pulled back in!

conda/plugins/types.py
soapy1: If we take this approach we might avoid any backwards compatiiblity concerns https://github.com/conda/conda/pull/15490#discussion_r2767935587 - like we can remove/deprecate the old implementation

soapy1: I tried this out https://github.com/conda/conda/compare/main...soapy1:conda:env-spec-reduce-io-2?expand=1

I think it's nice that this aligns more with how other plugins are structured. And with this approach lots of the stuff in `conda.env` can get deprecated maybe. But it is a bit of a departure from the original design, and maybe some bits give too much responsibility to the plugin manager?

soapy1: I spun this idea out into a demo with some notes https://github.com/conda/conda/pull/15731

soapy1:
closing in favour of https://github.com/conda/conda/pull/15731

PR #15497, opened 2025-12-10
Fix `test_get_index_lazy` by adding explicit subdir to `PackageRecord`

jezdez:
### Description

The test was failing because `PackageRecord.subdir` defaults to the current platform when not explicitly set. This caused a mismatch when looking up packages for a different platform (e.g., running on `osx-arm64` but testing `linux-aarch64` packages).

Changes:
- Add explicit `subdir` field to sample package data
- Parameterize `test_get_index_lazy` to test both defaults and conda-forge channels separately, with sample data inline in the decorator
- Add `prepend=False` to Index constructor to only query the specified channel

### Checklist - did you ...

### DIFF
--- a/tests/core/test_index.py
+++ b/tests/core/test_index.py
@@ -49,6 +49,7 @@ PLATFORMS = {
     ("Windows", "AMD64"): "win-64",
 }
 
+
 @pytest.fixture
 def pkg_cache_entries(mocker):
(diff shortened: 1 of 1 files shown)


jezdez:
pre-commit.ci autofix

jaimergp:
Absorbed in https://github.com/conda/conda/pull/15476

PR #15499, opened 2025-12-10
Add -O as an alias for --override-channels

jaimergp:
### Description

`--override-channels` is such a mouthful. @dholth suggested using `-O` (capital letter o), so here we are.

### Checklist - did you ...

### DIFF
--- a/news/15499-override-channels-alias
+++ b/news/15499-override-channels-alias
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Add `-O` as a flag alias for `--override-channels`. (#15499)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/cli/helpers.py
```diff
@@ -278,6 +278,7 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
         help="Use locally built packages. Identical to '-c local'.",
     )
     channel_customization_options.add_argument(
+        "-O",
         "--override-channels",
```
jezdez, line 282: `--override` could also be used, if that's easier to spell

kenodegard: I disagree, we have several override flags now (`--override-channels`, `--override-platforms`, `--override-frozen`) so `--override` is unclear

kenodegard:
Do we use uppercase aliases anywhere currently? If not should we standardize on capital means override?

`--override-channels` → `-C`
`--override-frozen` → `-F`
`--override-platforms` → `-P`

**(edit)** we have:

`--version` → `-V`
`--satisfied-skip-solve` → `-S`
`--use-index-cache` → `-C`

conda/cli/helpers.py
jaimergp: We have more overrides, sadly. I could compromise with `--oc`, à la `--sw, --stepwise` in pytest.

jaimergp:
Seems to be pretty arbitrary, first come, first serve basis? In that sense, `-O` could work (if it's not too confusing with `-o` or `-0`).

conda/cli/helpers.py
dholth: Surely override-channels is the most important; override-frozen says "please don't use me" and which "conda help" shows an "--override-platforms" tag. Agree that plain "--override" doesn't help with the too much to type problem or the disambiguate problem.

jezdez: Oh, good points, I digress.

jezdez:
> Do we use uppercase aliases anywhere currently? If not should we standardize on capital means override?
> 
> `--override-channels` → `-C` `--override-frozen` → `-F` `--override-platforms` → `-P`
> 
> **(edit)** we have:
> 
> `--version` → `-V` `--satisfied-skip-solve` → `-S` `--use-index-cache` → `-C`

I like this proposal, given that the `override` prefix exists in a number of options. @jaimergp how do you feel about it?

jaimergp:
We can't use `-C` because it means `--use-index-cache`. I guess we'll have to go with `--oc`?

dholth:
YAGNI?

jaimergp:
Can we find consensus? Vote with reactions like this:

- 👍 `--oc, --override-channels`
- 🚀 `-O, --override-channels`
- ❤️ `-X, --override-channels`
- 🎉 Something else (add a comment)

dholth:
My interpretation is that `-O` for override channels is fine, but the discussion got sidetracked with suggestions for overriding everything else. -O, -Ø, -Ö, -Ő, -𝓞

jezdez:
Given the consensus in the comments, I think this is ready to merge.

PR #15500, opened 2025-12-10
Do not create `__cuda` package if no version detected

jaimergp:
### Description

- Closes https://github.com/conda/conda/issues/15501
- Closes https://github.com/conda/conda-libmamba-solver/issues/809

### Checklist - did you ...

### DIFF
--- a/news/15500-no-cuda
+++ b/news/15500-no-cuda
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Do not expose `__cuda` virtual package when no CUDA version was detected. (#15500)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


beckermr:
@jaimergp do we need to fix up any of the other virtual packages? I honestly cannot tell from the code+comments how any of this works anymore.

jaimergp:
I checked and everything else seems in place. It was just that CUDA was returning `None`, which coerced to `0`. It should have been `NULL` instead, which does prevent the virtual package from being exposed.

The other implementations (micromamba, pixi) worked because they were not exposing `__cuda=0` themselves. `conda create --solver=classic` worked because... no clue, I guess different internal solver logic.

jaimergp:
I added this gotcha to https://github.com/conda/ceps/pull/103 so it's clear that the `__cuda` presence is meaningful and MUST NOT be exposed if no CUDA is detected.

jezdez:
Let's open a ticket in this repo since this warrants being very thorough on the documentation of the change.

jaimergp:
> Let's open a ticket in this repo since this warrants being very thorough on the documentation of the change.

Done: https://github.com/conda/conda/issues/15501

conda/plugins/virtual_packages/cuda.py
```diff
@@ -58,7 +58,7 @@ def cuda_version():
         return NULL
 
     result = queue.get()
-    return result
+    return result if result is not None else NULL
```
dholth, line 61: when None isn't NULL enough

Issue #15188, opened 2025-09-10
Refactor `conda.common.serialize.yaml_*`


kenodegard:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

We previously refactored and centralized the json parsing logic into `conda.common.serialize.json`: https://github.com/conda/conda/pull/14709

We should do the same for our yaml parsers.

### Why is this needed?

Provide a consistent DX for json and yaml.

### What should happen?

1. Inline `ruamel.yaml` imports
2. Mirror the `conda.common.serialize.json` functions in `conda.common.serialize.yaml`
3. Deprecate conda.common.serialize.yaml_*`

PR #15503, opened 2025-12-11
Refactor yaml parsers into `conda.common.serialize.yaml` (2)

soapy1:
### Description

Picks the work started in https://github.com/conda/conda/pull/15176 back up.

Fixes https://github.com/conda/conda/issues/15188.
Closes #15176

Deprecates (in `conda/common/serializers`):
* `_yaml_round_trip`
* `_yaml_safe`
* `yaml_round_trip_load`
* `yaml_safe_load`
* `yaml_round_trip_dump`
* `yaml_safe_dump`

Makes the yaml and json serializers have mirrored `load`/`loads` and `dump`/`dumps` functions.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2437,10 +2437,7 @@ def test_create_env_different_platform(
     with tmp_env(*args, shallow=False) as prefix:
         # check that the subdir is defined in environment's condarc
         # which is generated during the `conda create` command (via tmp_env)
-        assert (
-            yaml.read(path=prefix / DEFAULT_CONDARC_FILENAME)["subdir"]
-            == platform
-        )
+        assert yaml.read(path=prefix / DEFAULT_CONDARC_FILENAME)["subdir"] == platform
 
         stdout, stderr, excinfo = conda_cli(
             "install",


soapy1:
pre-commit.ci autofix

conda/common/serialize/yaml.py
```diff
@@ -0,0 +1,127 @@
...
+            f"Object of type {data.__class__.__name__} is not YAML serializable"
+        )
+
+
+CondaYAMLRepresenter.add_representer(None, CondaYAMLRepresenter.default)
+
+
+@cache
+def _yaml() -> ruamel.yaml.YAML:
+    parser = ruamel.yaml.YAML(typ="rt")
```
kenodegard, line 46: Since the roundtrip parser is essentially the safe parser with comment preservation, it's safe to use everywhere. However, it's more complex and potentially slower, so we should verify there's no significant performance penalty before replacing all usage of the safe parser.

soapy1: I wrote a one off benchmark test to see if there is much of a difference between the old `yaml_safe_load` and new `yaml.load`.

I added this test to `tests/common/test_serialize.py` (I don't think it's worth it to keep this test as part of the test suite so I didn't push it up to this branch).
```
@pytest.mark.benchmark
@pytest.mark.parametrize(
    "fn,yaml_str",
    [
        pytest.param(
            yaml_safe_load,
            YAML_COMMENT,
            id="yaml_safe_load/yaml_comment",
        ),
        pytest.param(
          yaml.loads,
          YAML_COMMENT,
          id="yaml.loads/yaml_comment",
        ),
        pytest.param(
            yaml_safe_load,
            "\n".join(f"{i}_field: value" for i in range(250)),
            id="yaml_safe_load/250 line doc",
        ),
        pytest.param(
          yaml.loads,
          "\n".join(f"{i}_field: value" for i in range(250)),
          id="yaml.loads/250 line doc",
        ),
    ]
)
def test_old_yaml_safe_load(fn, yaml_str):
    fn(yaml_str)
```


The benchmark results are as follows:
```
 $ python -m pytest tests/common/test_serialize.py --codspeed

 Benchmark Results                                            
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃                                            Benchmark ┃  Time (best) ┃ Rel. StdDev ┃ Run time ┃ Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│ test_old_yaml_safe_load[yaml_safe_load/yaml_comment] │    449,844ns │       15.7% │    2.93s │ 3,002 │
│     test_old_yaml_safe_load[yaml.loads/yaml_comment] │  1,243,285ns │        6.0% │    2.96s │ 2,240 │
│ test_old_yaml_safe_load[yaml_safe_load/250 line doc] │ 18,170,020ns │       19.9% │    3.01s │   162 │
│     test_old_yaml_safe_load[yaml.loads/250 line doc] │ 25,167,016ns │        3.4% │    3.00s │   116 │
└──────────────────────────────────────────────────────┴──────────────┴─────────────┴──────────┴───────┘
```

For moderate sized yaml files `yaml_safe_loads` is <0.1 s faster, and for  larger sized yaml files `yaml_safe_loads` is <0.05s faster. There is some pretty big stddev in the `yaml_safe_load` measurements. Not sure what that is from. But given these numbers, I'm happy to proceed with the round trip loader. And, we should be able to see some more performance gains once more caching is implemented.

What do you think?

soapy1: Running the benchmarks again I get more reasonable stddev for `yaml_safe_load` measurements

```
 Benchmark Results                                            
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃                                            Benchmark ┃  Time (best) ┃ Rel. StdDev ┃ Run time ┃ Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│ test_old_yaml_safe_load[yaml_safe_load/yaml_comment] │    881,100ns │        8.1% │    2.79s │ 2,958 │
│     test_old_yaml_safe_load[yaml.loads/yaml_comment] │  1,244,990ns │       20.2% │    3.00s │ 2,285 │
│ test_old_yaml_safe_load[yaml_safe_load/250 line doc] │ 18,078,793ns │        1.9% │    3.01s │   163 │
│     test_old_yaml_safe_load[yaml.loads/250 line doc] │ 24,882,968ns │        1.0% │    3.01s │   119 │
└──────────────────────────────────────────────────────┴──────────────┴─────────────┴──────────┴───────┘
```

kenodegard: Nice! I'm good with simply adopting the roundtrip parser then since it simplifies the code.

conda/common/serialize/__init__.py
```diff
@@ -12,30 +12,50 @@
 log = getLogger(__name__)
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

conda/common/serialize/__init__.py
```diff
@@ -12,30 +12,50 @@
...
 
-    parser = YAML(typ="rt")
+    parser = yaml.YAML(typ="rt")
     parser.indent(mapping=2, offset=2, sequence=4)
     return parser
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

conda/common/serialize/__init__.py
```diff
@@ -12,30 +12,50 @@
...
+    parser = yaml.YAML(typ="safe", pure=True)
     parser.indent(mapping=2, offset=2, sequence=4)
     parser.default_flow_style = False
     parser.sort_base_mapping_type_on_output = False
     return parser
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

conda/common/serialize/__init__.py
```diff
@@ -12,30 +12,50 @@
...
+    addendum="Use `conda.common.serialize.yaml.load()` instead.",
+)
 def yaml_round_trip_load(string):
     return _yaml_round_trip().load(string)
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

conda/common/serialize/__init__.py
```diff
@@ -46,6 +66,11 @@ def yaml_safe_load(string):
     return _yaml_safe().load(string)
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

conda/common/serialize/__init__.py
```diff
@@ -54,6 +79,11 @@ def yaml_round_trip_dump(object, stream=None):
         return ostream.getvalue()
 
 
+@deprecated(
+    "26.3",
+    "26.9",
```
kenodegard:
```suggestion
    "26.9",
    "27.3",
```

kenodegard (changes requested):
We should add explicit tests for the three different types of read/writes offered in both json and yaml to ensure they work as expected.

conda/common/serialize/json.py
```diff
@@ -32,18 +35,83 @@ def default(self, obj: Any) -> Any:
...
+
+@overload
+def write(obj: Any, **kwargs) -> str: ...
 
 
-def dumps(*args, **kwargs):
+def write(
+    obj: Any, *, fp: IO[str] | None = None, path: PathType | None = None, **kwargs
+) -> None | str:
+    if sum(value is not None for value in (fp, path)) > 1:
```
kenodegard: Would this be clearer if we flip it?


```suggestion
    if (fp, path).count(None) < 1:
```

conda/common/serialize/json.py
```diff
@@ -32,18 +35,83 @@ def default(self, obj: Any) -> Any:
...
+
+
+def read(
+    *,
+    text: str | None = None,
+    fp: IO[str] | None = None,
+    path: PathType | None = None,
+    **kwargs,
+) -> Any:
+    if sum(value is not None for value in (text, fp, path)) != 1:
```
kenodegard:
```suggestion
    if (text, fp, path).count(None) != 2:
```

conda/common/serialize/yaml.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@overload
+def write(obj: Any, *, path: PathType) -> None: ...
+
+
+def write(
+    obj: Any, *, fp: IO[str] | None = None, path: PathType | None = None
+) -> None | str:
+    if sum(value is not None for value in (fp, path)) > 1:
```
kenodegard:
```suggestion
    if (fp, path).count(None) < 1:
```

conda/common/serialize/yaml.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@overload
+def read(*, path: PathType) -> Any: ...
+
+
+def read(
+    *, text: str | None = None, fp: IO[str] | None = None, path: PathType | None = None
+) -> Any:
+    if sum(value is not None for value in (text, fp, path)) != 1:
```
kenodegard:
```suggestion
    if (text, fp, path).count(None) != 2:
```

conda/common/serialize/json.py
soapy1: hmmm, I think the most clear way to express this is to be more explicit. Like:
```
if fp and path:
  raise ValueError("At most one of fp or path must be provided")
```

If part of the motivation for this logic is to be able to easily extend this error checking to support more output types, then I think ` sum(value is not None for value in (fp, path)) > 1` is more correct.

For example consider the 3 values `a="a"`, `b="b"`, `c=None`. For providing all these values, we would want the error case to be triggered. Since we only want at most one of these values to be provided.
* ` sum(value is not None for value in (a,b,c)) >1` -> `True` error case is triggered
* `(a,b,c).count(None) < 1` -> `False` error is not triggered

So, comparing these solutions I think:
1. `if sum(value is not None for value in (fp, path)) ` is good, though maybe too future looking
2. `if fp and path` if the simplist and more explicit implementation
3. `if (fp, path).count(None) < 1` is not more clear than (1) imo, I think it's my least preferred option.

Would you be satisfied sticking with (1)? Open to other options too!

conda/common/serialize/json.py
soapy1: I think this suggestion makes things more clear, thanks!

conda/common/serialize/json.py
kenodegard: No, this wasn't meant to be extensible, was just grappling with whether we could get away with `fp and path` or if we needed to do `fp is not None and path is not None`

After more thought I dont think `fp: IO[str]` could ever be falsy? In which case let's go with the the most explicit option `if fp and path`

soapy1:
pre-commit.ci autofix

tests/common/test_serialize.py
```diff
@@ -0,0 +1,217 @@
...
+def test_read_write(
+    obj: Any,
+    text: str,
+    read: Callable,
+    write: Callable,
+    legacy_dump: Callable | None,
+    legacy_load: Callable | None,
+    tmp_path: Path,
+):
+    # Test read/write text
```
soapy1, line 143: It's maybe worth it to split up these different types of read/writes into different tests?

soapy1:
pre-commit.ci autofix

conda/common/serialize/__init__.py
```diff
@@ -46,6 +66,11 @@ def yaml_safe_load(string):
     return _yaml_safe().load(string)
 
 
+@deprecated(
+    "26.9",
+    "27.3",
+    addendum="Use `conda.common.serialize.yaml.dump()` instead.",
```
soapy1, line 72: Hmmm, calling out that `conda.common.serialize.yaml.dump()` is slightly different from this implementation. 
* In  `conda.common.serialize.yaml.dump()` , the `fp` arg (or in this function `stream`) is required
* In `yaml_round_trip_dump` if no stream is provided the function will retrun a string of the dumped object.  `conda.common.serialize.yaml.dump()` will always return `None`. 

For users that are looking to use the return of `yaml_round_trip_dump`, they should use `conda.common.serialize.yaml.dumps()`. 
For users that need to specify the output stream, they should use `conda.common.serialize.yaml.dump()`.

kenodegard: could update the deprecation message to say:

```suggestion
    addendum="Use `conda.common.serialize.yaml.dump()` or `conda.common.serialize.yaml.dumps()` instead.",
```

ryanskeith (dismissed):
Great add.

conda/env/env.py
```diff
@@ -419,9 +419,9 @@ def to_dict(self, stream=None):
     def to_yaml(self, stream=None):
         """Convert information related to the ``EnvironmentYaml`` into a ``yaml`` string"""
         d = self.to_dict()
-        out = yaml_safe_dump(d, stream)
         if stream is None:
-            return out
+            return yaml.dumps(d)
+        yaml.dump(d, stream)
```
kenodegard:
```suggestion
        return yaml.write(d, fp=stream)
```

conda/cli/condarc.py
```diff
@@ -337,14 +337,12 @@ def read(self, path: str | os.PathLike[str] | Path | None = None) -> dict[str, A
...
-        from ..common.serialize import yaml_round_trip_load
+        from ..common.serialize import yaml
 
         path = Path(path or self._path)
 
         try:
-            self._content = (
-                yaml_round_trip_load(Path(path or self._path).read_text()) or {}
-            )
+            self._content = yaml.loads(Path(path or self._path).read_text()) or {}
```
kenodegard:
```suggestion
            self._content = yaml.read(path=path) or {}
```

tests/test_create.py
```diff
@@ -2438,9 +2438,7 @@ def test_create_env_different_platform(
         # check that the subdir is defined in environment's condarc
         # which is generated during the `conda create` command (via tmp_env)
         assert (
-            yaml_round_trip_load((prefix / DEFAULT_CONDARC_FILENAME).read_text())[
-                "subdir"
-            ]
+            yaml.loads((prefix / DEFAULT_CONDARC_FILENAME).read_text())["subdir"]
```
kenodegard:
```suggestion
            yaml.read(path=prefix / DEFAULT_CONDARC_FILENAME)["subdir"]
```

conda/cli/condarc.py
```diff
@@ -358,11 +356,11 @@ def write(self, path: str | os.PathLike[str] | Path | None = None) -> None:
         :raises CondaError: If the file cannot be written.
         """
         from .. import CondaError
-        from ..common.serialize import yaml_round_trip_dump
+        from ..common.serialize import yaml
 
         path: Path = Path(path or self._path)
         try:
-            path.write_text(yaml_round_trip_dump(self.content))
+            path.write_text(yaml.dumps(self.content))
```
kenodegard:
```suggestion
            yaml.write(self.content, path=path)
```

conda/cli/condarc.py
soapy1: I think we still need the `path or self.path` bit here.

soapy1:
pre-commit.ci autofix

conda/cli/condarc.py
kenodegard: line 342 above does:

`path = Path(path or self._path)`

soapy1:
>Unanswered question is whether we should be using the read/write functions instead and only leave load(s)/dump(s) as backports (or perhaps we shouldn't have load(s)/dump(s) at all?).

@kenodegard imo, the load(s)/dump(s) functions provide a pleasant interface for reading + writing yaml/json. I'm a soft +1 for continuing to use these function names.

jezdez:
> > Unanswered question is whether we should be using the read/write functions instead and only leave load(s)/dump(s) as backports (or perhaps we shouldn't have load(s)/dump(s) at all?).
> 
> @kenodegard imo, the load(s)/dump(s) functions provide a pleasant interface for reading + writing yaml/json. I'm a soft +1 for continuing to use these function names.

I agree with that, load/dump/loads/dumps are classic APIs and they don't hurt to have around.

conda/cli/condarc.py
soapy1: true!

kenodegard (approved):

PR #15525, opened 2025-12-16
Elevate deprecation warnings to errors for the conda test suite

agriyakhetarpal:
### Description

This PR converts the deprecation warnings raised by the conda source code and other plugins/dependencies into hard errors at test suite execution time, and builds on @kenodegard's, @travishathaway's, and @beckermr's long-standing efforts in #12771. The benefit is that we can recognise the deprecations we introduce early on, rather than calling deprecated code in our test suite (unless doing so is warranted for other reasons).

cc: @jaimergp, who suggested that I reopen this as my own PR and show it some love to move it closer to the finish line :D

Most of the diff from the original PR is here, including my own changes on top; I've updated the error message match strings for some of them, and others were removed after merging the latest changes from `main` because the deprecation cycle was over and the source code was removed. I've been able to remove most of the old deprecation warning filters, as they are no longer applicable, and I also reduced the number of `pytest.deprecated_call()` calls to a minimum to ensure the test suite is green.

The commit history has not been compressed; there are simply way too many changes to keep track of (😄) – please let me know if you'd like me to squash the commits in any meaningful manner.

One question I still have after reading through the history of #12771 is https://github.com/conda/conda/pull/14754#issuecomment-2905580045, as it appears that we're still ignoring this circular import/deprecation warning problem and it hasn't been resolved yet. I haven't thoroughly followed any internal discussions on this work, so please let me know if there has been a resolution that I may have missed.

### Needs

- https://github.com/conda/conda/pull/15531

### Checklist - did you ...

### DIFF
--- a/conda/common/serialize/json.py
+++ b/conda/common/serialize/json.py
@@ -8,12 +8,7 @@ from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-# We must import frozendict before importing json to enable its monkeypatch over
-# json.JSONEncoder. This makes sure that our CondaJSONEncoder inherits from
-# frozendict's patched JSONEncoder, which natively supports serialising frozendicts.
-# See https://github.com/Marco-Sulla/python-frozendict/blob/a9744d61f42e86fc10c9c3668425abc4d485a9ec/src/frozendict/monkeypatch.py
-# and https://github.com/Marco-Sulla/python-frozendict/releases/tag/v2.3.6 for more.
-from frozendict import frozendict  # noqa: F401
+from frozendict import frozendict
 
 # detect the best json library to use
 from requests.compat import json
@@ -24,6 +19,11 @@ if TYPE_CHECKING:
 
 class CondaJSONEncoder(json.JSONEncoder):
     def default(self, obj: Any) -> Any:
+        # immutable types
+
+        if isinstance(obj, frozendict):
+            return dict(obj)
+
         # Python types
         if isinstance(obj, Enum):
             return obj.value
--- a/tests/common/test_json.py
+++ b/tests/common/test_json.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Tests for conda.common.serialize.json module."""
+
+from __future__ import annotations
+
+import json
+
+from frozendict import frozendict
+
+from conda.common.serialize.json import CondaJSONEncoder, dumps
+
+
+def test_condajsonencoder_serialises_frozendicts():
+    frozen = frozendict({"name": "test-package", "version": "1.0.0"})
+    result = json.dumps(frozen, cls=CondaJSONEncoder)
+
+    assert json.loads(result) == {"name": "test-package", "version": "1.0.0"}
+
+    data = {
+        "actions": {
+            "UNLINK": [
+                frozendict({"name": "package1"}),
+                frozendict({"name": "package2"}),
+            ]
+        }
+    }
+    result = json.dumps(data, cls=CondaJSONEncoder)
+    parsed = json.loads(result)
+
+    assert parsed["actions"]["UNLINK"][0]["name"] == "package1"
+    assert parsed["actions"]["UNLINK"][1]["name"] == "package2"
+
+
+def test_condajsonencoder_with_dumps():
+    frozen = frozendict({"key": "value"})
+    result = dumps(frozen)
+    assert json.loads(result) == {"key": "value"}


conda/common/serialize/json.py
```diff
@@ -17,6 +17,12 @@
 
 class CondaJSONEncoder(json.JSONEncoder):
     def default(self, obj: Any) -> Any:
+        # immutable types
+        from frozendict import frozendict
+
+        if isinstance(obj, frozendict):
+            return dict(obj)
+
```
agriyakhetarpal: Support for `frozendicts` for `CondaJSONEncoder` was introduced in b255997ca90bc2a500e310a43fdd5dc705a5f8f7 and was required for f73917f91113a15d96c5f549d48339dfc75c997f

pyproject.toml
```diff
@@ -163,6 +163,18 @@ doctest_optionflags = [
...
   "ELLIPSIS",
 ]
+filterwarnings = [
+  # elevate all deprecation warnings to errors
+  "error::PendingDeprecationWarning",
+  "error::DeprecationWarning",
+  "error::FutureWarning",
+  # ##### external library deprecations that we cannot fix in the conda source code
+  # - conda_content_trust.plugin uses old CondaSubcommand import path; remove when conda-content-trust plugin is updated
+  "ignore:conda.plugins.CondaSubcommand:PendingDeprecationWarning:conda_content_trust.plugin",
```
agriyakhetarpal, line 173: Please see https://github.com/conda/conda-content-trust/pull/238 for this

beckermr:
The issue with the original PR is that it touched too much of the code base to be properly reviewed. Can we break this one up into even smaller pieces so it can be merged bit by bit?

agriyakhetarpal:
Yes, sure – that works! I'll keep updating the PR description with PRs that spin out of this work.

conda/common/serialize/json.py
agriyakhetarpal: I have moved this to #15532; please review that first

agriyakhetarpal:
Hi @beckermr, this PR should now be in a much better state for review compared to the previous iteration from yesterday. I've also submitted two additional PRs (#15531 and #15532), which I've mentioned in the description. These should help further reduce the diff, and I'd recommend reviewing them before this PR. I can rebase here once they are merged.

The gist of this PR, after my recent cleanup commits, is that it does not contain any expired deprecations as of v25.11.0 (there is one for `conda.cli.main_compare.get_packages`, but that's not going away until v26.9) and includes only relevant code changes. Please let me know your thoughts! If you think it would still be better to separate any changes into their own constituent PRs, I can do that as well. Thanks! :)

beckermr:
bump @travishathaway and @kenodegard who decided it was too big to review originally - what do you all think? Can we go ahead or do we need even smaller chunks?

conda/common/serialize/json.py
travishathaway: @agriyakhetarpal,

Let's remove this addition per our conversation over here: https://github.com/conda/conda/pull/15532

tests/env/test_create.py
```diff
@@ -158,16 +158,26 @@ def test_create_empty_env(
     conda_cli: CondaCLIFixture,
     tmp_envs_dir: Path,
 ):
+    import warnings
```
travishathaway: Does it make sense to move this import to the top of the file? If not, we can leave it where it is.

travishathaway:
@beckermr,

I'm fine with the size of the pull request as it is now. I made a couple requests for changes. Once those are addressed, I'll be ready to give a ✅ . I'd also like @kenodegard to chime in to if he has a chance.

tests/env/test_create.py
kenodegard: I agree but don't think it should block merging, this will be removed again in the future 🤷🏼‍♂️

kenodegard (commented):
Wow!! Thank you thank you thank you!! This is great!

This is much smaller than this effort was at one point so I'm happy to merge this as is (once we remove the `frozendict` part).

No need to mess with the git history, it'll all get squashed on merge anyhow.

tests/env/test_create.py
agriyakhetarpal: I proceeded to move it to the top of the file, the linter will catch it when we remove it anyway :D

conda/common/serialize/json.py
agriyakhetarpal: Sorry, slipped my mind! Done, thanks!

pyproject.toml
agriyakhetarpal: We will need a new release of conda-content-trust to remove it – it doesn't need to block this PR, so I'll resolve this comment for now.

agriyakhetarpal:
The JSON encoder monkeypatching issue is tripping up the tests again. I investigated this a bit, and while @travishathaway's suggestion is correct that we don't need to merge #15532 – there is some nuance on how `frozendict` works with custom encoders. Leaving a comment here that I'll put in a fix for this in a moment, and explain what it is.

agriyakhetarpal:
The automatic JSON support in `frozendict` v2.3.6 and later only works for encoders defined _after_ `frozendict` is imported. Since `CondaJSONEncoder` is defined in an early-imported module, it can't benefit from `frozendict`'s monkeypatch. 

Before the `frozendict` import, we get a `<class 'json.encoder.JSONEncoder'>` class, and after doing so, we get a patched `<class 'frozendict._getFrozendictJsonEncoder.<locals>.FrozendictJsonEncoderInternal'>` one. Via a02010a865e2f69dce0797016f1d95f585749d47, `CondaJSONEncoder` will derive from this class instead.

I've left a brief inline comment in the code for those who may stumble across this at a later point in time.

conda/common/serialize/json.py
```diff
@@ -8,6 +8,13 @@
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+# We must import frozendict before importing json to enable its monkeypatch over
+# json.JSONEncoder. This makes sure that our CondaJSONEncoder inherits from
+# frozendict's patched JSONEncoder, which natively supports serialising frozendicts.
+# See https://github.com/Marco-Sulla/python-frozendict/blob/a9744d61f42e86fc10c9c3668425abc4d485a9ec/src/frozendict/monkeypatch.py
+# and https://github.com/Marco-Sulla/python-frozendict/releases/tag/v2.3.6 for more.
+from frozendict import frozendict  # noqa: F401
```
beckermr: My import-fu is weak. 

Will this always work? What happens if someone starts a new interpreter and then does

```python
import json
from json import JSONEncoder

from conda.common.serialize import *
```

?

Does it all still work?

TBH relying on a third-party package to **monkey patch the Python stdlib** is not a preferred solution. Monkey patching anything is usually a bad sign. We should not be relying on monkey patching and it costs us almost nothing to leave the frozendict if statement in the custom CondaJSONEncoder class.

beckermr: The fact that the tests are tripped up due to the monkey patching is the kind of fragility we don't need. A little bit of extra explicit code instead of an opaque global patching operation that depends on import order is obviously better IMHO.

agriyakhetarpal: > Does it all still work?

> The fact that the tests are tripped up due to the monkey patching is the kind of fragility we don't need. A little bit of extra explicit code instead of an opaque global patching operation that depends on import order is obviously better IMHO.

Hmm, yeah, it won't work with your code snippet :/ We either need an import right in `conda/__init__.py` (which you advise against), or we should reopen #15532?

agriyakhetarpal: or do you think we should go ahead with dropping `frozendict` and writing our own? https://github.com/conda/conda/pull/15532#issuecomment-3670397868

beckermr: We should reopen https://github.com/conda/conda/pull/15532 and merge that. @jezdez pointed out in previous comments that monkey patching is bad. I 100% agree with that.

travishathaway: @beckermr,

Please see my comment:

- https://github.com/conda/conda/pull/15532#issuecomment-3680704229

PR has been merged ✅.

PR #15527, opened 2025-12-16
Replace `#:` comments with trailing docstrings

### LINKED ISSUE #14356: Docstrings in the `#:` prefix style are not rendered in the API docs
### Checklist

### What happened?

In a recent discussion at https://github.com/conda/conda/pull/13880#discussion_r1755199016, we discovered that many docstrings for constants that follow the prepended `#:` are not rendered in the API docs.

It would be good to investigate whether that needs a change to the docstrings (it was suggested that an additional new line before the docstring is needed), or a change in the autoapi or sphinx configuration, and finally to make sure that all the docs are rendered.

### Additional Context

_No response_

kenodegard:
### Description

We previously used `#:` to document constants. This worked when we were using Sphinx's autodocs however we have since adopted sphinx-autoapi which does not support `#:`. Furthermore IDEs do not understand `#:` so everything points to us switching to the trailing docstring pattern when documenting constants.

To ensure we don't go down this path again in the future we also add a pre-commit hook to prevent adding `#:` docs.

Resolves #14356

### Checklist - did you ...

### DIFF
--- a/tests/plugins/test_settings.py
+++ b/tests/plugins/test_settings.py
@@ -22,27 +22,27 @@ from conda.plugins.manager import CondaPluginManager
 
 log = logging.getLogger(__name__)
 
-#: Name for a string type parameter
 STRING_PARAMETER_NAME = "string_parameter"
+"""Name for a string type parameter."""
 STRING_PARAMETER_ALIAS = "string_parameter_alias"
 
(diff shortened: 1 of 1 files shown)


travishathaway:
Thanks for taking care of this @kenodegard 👏

PR #15532, opened 2025-12-17
`CondaJSONEncoder`: support serialising `frozendicts`

agriyakhetarpal:
### Description

The `test_remove_all_json` test, as follows:

https://github.com/conda/conda/blob/053ab56aee99b9a5d76779a53b138b1fc1298cc1/tests/cli/test_subcommands.py#L160-L171

runs `conda remove --all --json`. This command calls `conda.common.serialize.json.dumps()`. However, since there is no native support for `frozendict`, the system falls back to the deprecated monkeypatched JSON encoder rather than `CondaJSONEncoder`. This is why running this test as part of the suite generates the following warning:

```python
PendingDeprecationWarning: Monkey-patching json.JSONEncoder to support frozendict is pending deprecation and will be removed in version 26.9
```

This PR adds support for serialising immutable dictionaries in `CondaJSONEncoder`. I have separated this change from #15525 (originally #12771) to reduce its diff.

### Checklist - did you ...

- [ ] **[⚠️ Do we need one?]** Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     

travishathaway:
@agriyakhetarpal,

I looked around on this and learned that it's not actually necessary to add this our `CondaJSONEncoder` because this is already being accounted for in `frozendict` itself.

Please see the following links for more information:
- https://github.com/Marco-Sulla/python-frozendict/releases/tag/v2.3.6
- https://github.com/Marco-Sulla/python-frozendict/issues/47

Conda currently has a requirement of `frozendict>=2.4.2`, so this shouldn't be an issue. When we finally remove the deprecated code paths, everything should JustWork™️ ✨ .

jezdez (approved):

travishathaway:
@jezdez,

Do you have any feedback on my comment? Am I correct in thinking that this pull request isn't necessary?

jezdez:
Oh huh, that comment didn't load in the tab I had this issue open before posting my comment. This changes things!

Looking at the change upstream, this monkeypatch seems like a pretty bad idea, [Marco-Sulla/python-frozendict@`v2.3.5...v2.3.6`#diff-f9ec75e660](https://github.com/Marco-Sulla/python-frozendict/compare/v2.3.5...v2.3.6#diff-f9ec75e660e1fd4560e72d07d2fda88aae4a9ce0c87e8e2777a56cc630c979da) and we added the pending deprecation note in https://github.com/conda/conda/pull/14709 specifically to stop monkeypatching locally as well. I would say this change here is technically correct, but you're right that the monkeypatching in https://github.com/Marco-Sulla/python-frozendict/blob/master/src/frozendict/monkeypatch.py seems to cover most cases. I wonder if there is any risk involved with the change in https://github.com/conda/conda/pull/14709 that we hadn't thought about. This is why monkeypatching is never a good API strategy!

jezdez:
This makes me want to stop relying on frozendict btw.

jezdez:
I wonder if https://peps.python.org/pep-0814/ goes anywhere and if there are backports? What about https://adamj.eu/tech/2022/01/05/how-to-make-immutable-dict-in-python/ should we simply roll it ourselves?

travishathaway:
> This makes me want to stop relying on frozendict btw.

My thoughts exactly! 😬

travishathaway:
> I wonder if https://peps.python.org/pep-0814/ goes anywhere and if there are backports? What about https://adamj.eu/tech/2022/01/05/how-to-make-immutable-dict-in-python/ should we simply roll it ourselves?

Maybe we should just create a separate issue for this work and not merge this pull request?

travishathaway:
@kenodegard, maybe you have some thoughts on this. Couldn't help but notice your name as I was skimming through the release notes in `frozendict`.

agriyakhetarpal:
Thanks for the conversation, y'all! For now, I'll remove this change from #15525 and replace it with a `deprecated_call` there, and I'll mark this pull request as a draft. I don't yet see a backport for PEP 814 work, so maybe having our own `frozendict` in `auxlib` is the best idea.

agriyakhetarpal:
This PR has been superseded by the change I described in https://github.com/conda/conda/pull/15525#issuecomment-3675801238. Thanks, everyone!

agriyakhetarpal:
I'm reopening this PR as per @beckermr's suggestion in https://github.com/conda/conda/pull/15525#discussion_r2636113770. The support that `frozendict` provides for custom encoders like ours is a bit finicky – mainly because it requires us to mess with the import order and ensure that `frozendict` is imported before `json`.

travishathaway:
Hi @beckermr,

Thanks for chiming in here and looking into this too. I read your feedback over [here](https://github.com/conda/conda/pull/15525) and think it's reasonable to include this pull request. Inspecting the type here doesn't have a huge cost and we add additional tests to ensure this expect behavior works.

Because of those reasons, I'll go ahead merge this pull request. But, with that being said, I think it's worth it to research what we could do instead of relying on a dependency that monkey patches the standard library. I think we can all agree that that's not a good idea, so I will create an issue specifically for that work.

Issue #14993, opened 2025-07-07
`conda run` could be faster


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched for other speed issues and didn't find a duplicate

### What happened?

`conda run` feels slow to run. A simple `echo hello`:

```
$ time conda run echo hello
hello

conda run echo hello  0.77s user 0.22s system 79% cpu 1.244 total
```

Meanwhile, `micromamba`:

```
$ time micromamba run echo hello
hello
micromamba run echo hello  0.02s user 0.01s system 45% cpu 0.079 total
```

A lot of that is Python vs C++, but looking at the code of `conda run`, I see one obvious opportunity for optimization, and then maybe a controversial one:

* Save one `conda` subprocess:
  * We initialize `conda` in the wrapper script by sourcing [`conda.sh`](https://github.com/conda/conda/blob/main/conda/shell/etc/profile.d/conda.sh) (or equivalent).
  * Then, we use this to call `conda activate`, which starts a new Python process to generate and eval shell code. 
  * However, we know the arguments already, so we could directly call `conda.activate.Activator(["shell.bash", "activate", target_prefix])` when we are generating the wrapper script and paste it there, saving one unnecessary `conda` process startup.
* Spicy 🌶️. I don't think `conda run` needs the plugin system, so it could be special cased along with `activate` and all the `shell.*` subcommands

### Debug

```shell
NA
```

### Conda info

```shell
active environment : None
       user config file : /Users/jrodriguez/.condarc
 populated config files : /Users/jrodriguez/.local/conda/.condarc
                          /Users/jrodriguez/.condarc
          conda version : 25.5.1
    conda-build version : not installed
         python version : 3.12.8.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=25.5.1=0
                          __osx=15.5=0
                          __unix=0=0
       base environment : /Users/jrodriguez/.local/conda  (writable)
      conda av data dir : /Users/jrodriguez/.local/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-arm64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /Users/jrodriguez/.local/conda/pkgs
                          /Users/jrodriguez/.conda/pkgs
       envs directories : /Users/jrodriguez/.local/conda/envs
                          /Users/jrodriguez/.conda/envs
               platform : osx-arm64
[...]

Issue #14871, opened 2025-05-22
Fails to activate environment on Windows, temporary env file is locked


adament:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

We recently upgraded to conda 25.3.1 and have started experiencing the following issue occasionally when activating environments:
```
The process cannot access the file because it is being used by another process.
The system cannot find the file C:\Users\BAPINV01\AppData\Local\Temp\__conda_tmp_24137.txt.
Failed to run 'conda activate olympos-env'.
```
I think is because the file is locked by antivirus software after the activation script has finished writing the file. We experience this issue in other cases. However because the activation script tries to read the file right after it has been written it is locked and we get a permission denied error.

### Conda Info

```shell
>conda info

     active environment : None
       user config file : C:\Users\admt1u_aen\.condarc
 populated config files : D:\tools\Miniconda3\.condarc
          conda version : 25.3.1
    conda-build version : 25.4.2
         python version : 3.13.3.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=broadwell
                          __conda=25.3.1=0
                          __win=10.0.14393=0
       base environment : D:\tools\Miniconda3  (writable)
      conda av data dir : D:\tools\Miniconda3\etc\conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/win-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : D:\tools\Miniconda3\pkgs
                          C:\Users\admt1u_aen\.conda\pkgs
                          C:\Users\admt1u_aen\AppData\Local\conda\conda\pkgs
       envs directories : D:\tools\Miniconda3\envs
                          C:\Users\admt1u_aen\.conda\envs
                          C:\Users\admt1u_aen\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/25.3.1 requests/2.32.3 CPython/3.13.3 Windows/2016Server Windows/10.0.14393 solver/libmamba conda-libmamba-solver/25.4.0 libmambapy/2.1.1
          administrator : True
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> D:\tools\Miniconda3\.condarc <==
local_repodata_ttl: 0
channel_priority: flexible
channels:
  - conda-forge
show_channel_urls: True
always_copy: True

==> envvars <==
allow_softlinks: False
```

### Conda list

```shell
[...]

PR #15534, opened 2025-12-19
perf: towards a faster `conda run` via inline environment activation

agriyakhetarpal:
### Description

This PR is a part of #14993, and is an effort to make `conda run` faster. Our approach here is to avoid calling `conda activate` in the wrapper script and instead use platform-specific activators. On POSIX platforms, we build the `PosixActivator` and have it write the activation actions to the wrapper script. For Windows, we disable file streaming for `CmdExeActivator` (save I/O overhead) and have it print the activation output to memory, whose lines we then iterate over and `CALL`/`SET` as needed. On Unix, it's even possible for us to run the commands as `sh -c "<...>"` instead of using a temporary file (which should be a little faster), but that's a platform-specific optimisation, and I've not proceeded with it for now.

I was optimistic with my expectations for results here, and they are pretty promising, too. I've used [`hyperfine`](https://github.com/sharkdp/hyperfine) for benchmarking, as follows:

<details>
<summary><h3>Benchmarking</h3></summary>

##### Before

```bash
hyperfine --warmup 10 --runs 1000 "conda run -- python --version"
Benchmark 1: conda run -- python --version
  Time (mean ± σ):      1.262 s ±  0.179 s    [User: 0.919 s, System: 0.303 s]
  Range (min … max):    1.118 s …  3.736 s    1000 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

##### After

```bash
hyperfine --warmup 10 --runs 1000 "conda run -- python --version" 
Benchmark 1: conda run -- python --version
  Time (mean ± σ):      1.022 s ±  0.334 s    [User: 0.716 s, System: 0.245 s]
  Range (min … max):    0.880 s …  7.534 s    1000 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 ```

which is a 19.02% speedup! 🔥

#### Test suite

Even running parts of the test suite shows nice results. Here is the output:

##### Before

```bash
hyperfine "pytest -k test_main_run.py" --warmup 5 --runs 100
Benchmark 1: pytest -k test_main_run.py
  Time (mean ± σ):     25.888 s ±  2.872 s    [User: 17.240 s, System: 5.871 s]
  Range (min … max):   22.594 s … 38.291 s    100 runs
 ```

##### After

```bash
hyperfine "pytest -k test_main_run.py" --warmup 5 --runs 100 
Benchmark 1: pytest -k test_main_run.py
  Time (mean ± σ):     19.559 s ±  3.858 s    [User: 12.411 s, System: 4.326 s]
  Range (min … max):   16.203 s … 36.211 s    100 runs
```

Let's hope that CodSpeed also provides us with similar numbers 😉

</details>

### Checklist - did you ...

- [ ] ~Add / update necessary tests?~ Not needed; there is no functional/behaviour change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,7 @@ repos:
       # auto format Python codes within docstrings
       - id: blacken-docs
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.13
+    rev: v0.14.14
     hooks:
       # lint & attempt to correct failures (e.g. pyupgrade)
       - id: ruff-check
@@ -100,7 +100,7 @@ repos:
             tests/
           )
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.36.0
+    rev: 0.36.1
     hooks:
       # verify github syntaxes
       - id: check-github-workflows
--- a/conda/gateways/connection/download.py
+++ b/conda/gateways/connection/download.py
@@ -16,8 +16,10 @@ from pathlib import Path
 from ... import CondaError
 from ...auxlib.ish import dals
 from ...auxlib.logz import stringify
+from ...base.constants import CONDA_HOMEPAGE_URL
 from ...base.context import context
 from ...common.io import time_recorder
+from ...common.url import join_url
 from ...exceptions import (
     BasicClobberError,
     ChecksumMismatchError,
@@ -28,6 +30,7 @@ from ...exceptions import (
     ProxyError,
     maybe_raise,
 )
+from ...models.channel import Channel
 from ..disk.delete import rm_rf
 from ..disk.lock import lock
 from . import (
@@ -320,16 +323,99 @@ def download_http_errors(url: str):
             )
 
     except (ConnectionError, HTTPError) as e:
-        help_message = dals(
-            """
-        An HTTP error occurred when trying to retrieve this URL.
-        HTTP errors are often intermittent, and a simple retry will get you on your way.
-        """
-        )
+        status_code = getattr(e.response, "status_code", None)
+
+        if status_code == 403:
+            channel = Channel(url)
+            if channel.token:
+                help_message = dals(
+                    f"""
+                    The token given for the URL has insufficient permissions to access this resource.
+
+                    You may not have the required permissions to access this channel or package.
+                    Consider requesting access from the channel owner.
+
+                    Use `conda config --show` to view your configuration's current state.
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+            elif context.channel_alias.location in url:
+                help_message = dals(
+                    f"""
+                    The remote server has indicated you do not have permission to access this resource.
+
+                    This may mean:
+                      (a) You are not authenticated. Check if authentication is required for this channel
+                          and verify your credentials are correctly configured.
+                      (b) You do not have access to this private channel or package. Contact the
+                          channel owner to request access.
+
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+            else:
+                help_message = dals(
+                    f"""
+                    You do not have permission to access this resource.
+
+                    This may indicate:
+                      - The channel requires authentication. Check your credentials.
+                      - You do not have access to this private channel or package.
+
+                    You will need to modify your conda configuration to proceed.
+                    Use `conda config --show` to view your configuration's current state.
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+
+        elif status_code == 401:
+            channel = Channel(url)
+            if channel.token:
+                help_message = dals(
+                    f"""
+                    The token given for the URL is invalid.
+
+                    You will need to adjust your conda configuration to proceed.
+
+                    Use `conda config --show` to view your configuration's current state.
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+            elif context.channel_alias.location in url:
+                help_message = dals(
+                    f"""
+                    The remote server has indicated you are using invalid credentials for this channel.
+
+                    You may need to:
+                      (a) Remove or update the invalid token from your configuration, or
+                      (b) Provide conda with a valid token directly.
+
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+            else:
+                help_message = dals(
+                    f"""
+                    The credentials you have provided for this URL are invalid.
+
+                    You will need to modify your conda configuration to proceed.
+                    Use `conda config --show` to view your configuration's current state.
+                    Further configuration help can be found at <{join_url(CONDA_HOMEPAGE_URL, "docs/config.html")}>.
+                    """
+                )
+
+        else:
+            help_message = dals(
+                """
+                An HTTP error occurred when trying to retrieve this URL.
+                HTTP errors are often intermittent, and a simple retry will get you on your way.
+                """
+            )
+
         raise CondaHTTPError(
             help_message,
             url,
-            getattr(e.response, "status_code", None),
+            status_code,
             getattr(e.response, "reason", None),
             getattr(e.response, "elapsed", None),
             e.response,
--- a/conda/gateways/repodata/__init__.py
+++ b/conda/gateways/repodata/__init__.py
@@ -238,7 +238,7 @@ Exception: {e}
 
     except (ConnectionError, HTTPError, ChunkedEncodingError) as e:
         status_code = getattr(e.response, "status_code", None)
-        if status_code in (403, 404):
+        if status_code == 404:
             if not url.endswith("/noarch"):
                 log.info(
                     "Unable to retrieve repodata (response: %d) for %s",
@@ -269,36 +269,64 @@ Exception: {e}
                         response=e.response,
                     )
 
+        elif status_code == 403:
+            channel = Channel(url)
+            if channel.token:
+                help_message = """\
+The token given for the URL has insufficient permissions to access this resource.
+
+You may not have the required permissions to access this channel or package.
+Consider requesting access from the channel owner.
+
+Use `conda config --show` to view your configuration's current state.
+Further configuration help can be found at <{}>.
+""".format(join_url(CONDA_HOMEPAGE_URL, "docs/config.html"))
+
+            elif context.channel_alias.location in url:
+                help_message = """\
+The remote server has indicated you do not have permission to access this resource.
+
+This may mean:
+  (a) You are not authenticated. Check if authentication is required for this channel
+      and verify your credentials are correctly configured.
+  (b) You do not have access to this private channel or package. Contact the
+      channel owner to request access.
+
+Further configuration help can be found at <{}>.
+""".format(join_url(CONDA_HOMEPAGE_URL, "docs/config.html"))
+
+            else:
+                help_message = """\
+You do not have permission to access this resource.
+
+This may indicate:
+  - The channel requires authentication. Check your credentials.
+  - You do not have access to this private channel or package.
+
+You will need to modify your conda configuration to proceed.
+Use `conda config --show` to view your configuration's current state.
+Further configuration help can be found at <{}>.
+""".format(join_url(CONDA_HOMEPAGE_URL, "docs/config.html"))
+
         elif status_code == 401:
             channel = Channel(url)
             if channel.token:
                 help_message = """\
-The token '{}' given for the URL is invalid.
+The token given for the URL is invalid.
 
-If this token was pulled from anaconda-client, you will need to use
-anaconda-client to reauthenticate.
-
-If you supplied this token to conda directly, you will need to adjust your
-conda configuration to proceed.
+You will need to adjust your conda configuration to proceed.
 
 Use `conda config --show` to view your configuration's current state.
 Further configuration help can be found at <{}>.
-""".format(
-                    channel.token,
-                    join_url(CONDA_HOMEPAGE_URL, "docs/config.html"),
-                )
+""".format(join_url(CONDA_HOMEPAGE_URL, "docs/config.html"))
 
             elif context.channel_alias.location in url:
-                # Note, this will not trigger if the binstar configured url does
-                # not match the conda configured one.
                 help_message = """\
 The remote server has indicated you are using invalid credentials for this channel.
 
-If the remote site is anaconda.org or follows the Anaconda Server API, you
-will need to
-    (a) remove the invalid token from your system with `anaconda logout`, optionally
-        followed by collecting a new token with `anaconda login`, or
-    (b) provide conda with a valid token directly.
+You may need to:
+  (a) Remove or update the invalid token from your configuration, or
+  (b) Provide conda with a valid token directly.
 
 Further configuration help can be found at <{}>.
 """.format(join_url(CONDA_HOMEPAGE_URL, "docs/config.html"))
--- a/news/15594-improve-403-error-messages
+++ b/news/15594-improve-403-error-messages
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Improve error messages for HTTP 403 (Forbidden) responses with context-aware guidance on authentication and permissions. (#15594)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/gateways/test_repodata_gateway.py
+++ b/tests/gateways/test_repodata_gateway.py
@@ -246,6 +246,36 @@ def test_coverage_conda_http_errors(monkeypatch: MonkeyPatch):
     ):
         raise HTTPError(response=Response(404))
 
+    # 403 should have context-aware error messages (unlike 404)
+    # Test 403 on anaconda.org channel
+    with (
+        pytest.raises(CondaHTTPError, match="do not have permission"),
+        conda_http_errors("https://conda.anaconda.org/noarch", "repodata.json"),
+    ):
+        raise HTTPError(response=Response(403))
+
+    # Test 403 with a token - should trigger a different message
+    with (
+        pytest.raises(CondaHTTPError, match="insufficient permissions"),
+        conda_http_errors(
+            "/t/dh-73683400-b3ee-4f87-ade8-37de6d395bdb/conda-forge/noarch",
+            "repodata.json",
+        ),
+    ):
+        raise HTTPError(response=Response(403))
+
+    # Test 403 on non-anaconda URL
+    monkeypatch.setenv("CONDA_ALLOW_NON_CHANNEL_URLS", "1")
+    reset_context()
+    with (
+        pytest.raises(CondaHTTPError, match="do not have permission"),
+        conda_http_errors("https://example.org/main/linux-64", "repodata.json"),
+    ):
+        context.channel_alias.location = "xyzzy"
+        raise HTTPError(response=Response(403))
+
+    reset_context()
+
     # A variety of helpful error messages should follow
     with (
         pytest.raises(CondaHTTPError, match="invalid credentials"),
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -336,11 +336,45 @@ def test_download_http_errors():
         raise SSLError()
 
     # A variety of helpful error messages should follow
+
+    # 401 on generic URL
     with (
-        pytest.raises(CondaHTTPError, match=str(401)),
+        pytest.raises(CondaHTTPError, match="credentials you have provided"),
         download_http_errors("https://example.org/file"),
     ):
         raise HTTPError(response=Response(401))
+
+    # 401 with token should trigger token-specific message
+    with (
+        pytest.raises(CondaHTTPError, match="token.*is invalid"),
+        download_http_errors(
+            "/t/dh-73683400-b3ee-4f87-ade8-37de6d395bdb/conda-forge/file"
+        ),
+    ):
+        raise HTTPError(response=Response(401))
+
+    # 403 on generic URL
+    with (
+        pytest.raises(CondaHTTPError, match="do not have permission"),
+        download_http_errors("https://example.org/file"),
+    ):
+        raise HTTPError(response=Response(403))
+
+    # 403 with token should trigger token-specific message
+    with (
+        pytest.raises(CondaHTTPError, match="insufficient permissions"),
+        download_http_errors(
+            "/t/dh-73683400-b3ee-4f87-ade8-37de6d395bdb/conda-forge/file"
+        ),
+    ):
+        raise HTTPError(response=Response(403))
+
+    # Other HTTP errors should use generic message
+    with (
+        pytest.raises(CondaHTTPError, match="An HTTP error"),
+        download_http_errors("https://example.org/file"),
+    ):
+        raise HTTPError(response=Response(500))
 
 
 @pytest.mark.parametrize(


agriyakhetarpal:
Argh; I should have run the full test suite locally instead of a few files – looks like there is some work ahead for me. I'll mark this as a draft for a bit until I fix the failures.

agriyakhetarpal:
close/reopen to rerun CI

conda/utils.py
```diff
@@ -245,7 +240,54 @@ def wrap_subprocess_call(
...
             # fh.write("@FOR /F \"tokens=100\" %%F IN ('chcp') DO @SET CONDA_OLD_CHCP=%%F\n")
             # fh.write('@chcp 65001>NUL\n')
-            fh.write(f'{silencer}CALL "{conda_bat}" activate "{prefix}"\n')
+
+            # We pursue activation inline here, which allows us to avoid
+            # spawning a `conda activate` process at wrapper runtime. We
+            # ensure that a sentinel conda-meta directory exists to pass
+            # activation validation.
+            conda_meta_dir = Path(prefix) / "conda-meta"
+            conda_meta_dir.mkdir(parents=True, exist_ok=True)
```
soapy1: TL;DR: I think this function should not be creating an environment here. Instead, it should fail if the user requests a `prefix` that is not a valid environment. 

hmmm, this means that a side effect of using this function with a `prefix` that does not already exist is, that it creates an empty environment at that `prefix`. I think we should probably avoid that. 

### behaviour on the main branch

On the main branch, this function will happily create a script with an invalid environment to activate. For example:
```
>>> from conda.utils import wrap_subprocess_call
>>> from conda.base.context import context
>>> wrap_subprocess_call(context.root_prefix, "/tmp/prefixdoesnotexist", False, None, ["python", "--version"])
('/tmp/tmp4cx0mu1i', ['/usr/bin/bash', '/tmp/tmp4cx0mu1i'])
>>> with open("/tmp/tmp4cx0mu1i") as f:
...     print(f.read())
... 
eval "$(/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/conda shell.posix hook)"
conda activate  /tmp/prefixdoesnotexist
python --version
_CONDA_EXE_RC=$?
exit $_CONDA_EXE_RC
```

When I run this file I get the result:
```
$ bash /tmp/tmp4cx0mu1i                         

EnvironmentLocationNotFound: Not a conda environment: /tmp/prefixdoesnotexist

Python 3.10.19
```

So, the `conda activate` part fails with the error `EnvironmentNameNotFound`, but the script does not fully exit. Instead it finishes, running the `python --version` command in the context of the base environment (given from the first line `eval "$(/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/conda shell.posix hook)"`). That is pretty misleading, not sure if it's  intentional, but the command is getting run in the base environment.

### behaviour on this branch

This branch is pretty similar, in that it will happily create a script for an environment that does not exist:

```
>>> from conda.utils import wrap_subprocess_call
>>> from conda.base.context import context
>>> wrap_subprocess_call(context.root_prefix, "/tmp/prefixdoesnotexistpr", False, None, ["python", "--version"])
('/tmp/tmp_nnrkvgu', ['/usr/bin/bash', '/tmp/tmp_nnrkvgu'])
>>> with open("/tmp/tmp_nnrkvgu") as f:
...     print(f.read())
... 
eval "$(/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/conda shell.posix hook)"
. "/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/etc/conda/deactivate.d/deactivate.sh"
export _CE_M=''
export _CE_CONDA=''
export PATH='/tmp/prefixdoesnotexistpr/bin:/usr/local/bin/google-cloud-sdk/bin:/home/sophia/.pixi/bin:/home/sophia/.config/Code/User/globalStorage/github.copilot-chat/debugCommand:/home/sophia/.config/Code/User/globalStorage/github.copilot-chat/copilotCli:/home/sophia/miniconda3/condabin:/usr/local/bin/google-cloud-sdk/bin:/home/sophia/.pixi/bin:/home/sophia/.cargo/bin:/home/sophia/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/sophia/.pulumi/bin:/home/sophia/.cargo/bin:/home/sophia/.vscode/extensions/ms-python.debugpy-2025.18.0-linux-x64/bundled/scripts/noConfigScripts:/home/sophia/.pulumi/bin:/home/sophia/.cargo/bin'
export CONDA_PREFIX='/tmp/prefixdoesnotexistpr'
export CONDA_SHLVL='2'
export CONDA_DEFAULT_ENV='/tmp/prefixdoesnotexistpr'
export CONDA_PROMPT_MODIFIER=''
export CONDA_PREFIX_1='/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda'
export CONDA_EXE='/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/conda'
export _CONDA_EXE='/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/conda'
export CONDA_PYTHON_EXE='/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/bin/python'
export _CONDA_ROOT='/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda'
python --version
_CONDA_EXE_RC=$?
exit $_CONDA_EXE_RC
``` 

But, when I run this file I get the result:
```
$ bash /tmp/tmp_nnrkvgu                               
/tmp/tmp_nnrkvgu: line 15: python: command not found
```

So this script successfully uses the right environment (presumably), but there is nothing installed so we get the result `command not found`.

And I can observe that an empty environment has been created
```
$ ls /tmp/prefixdoesnotexistpr                                  
conda-meta
```

### suggestion?
Maybe instead of creating an empty environment, it would be ok to produce an error if the provided `prefix` is invalid.
For example, for my diff:
```
diff --git a/conda/utils.py b/conda/utils.py
index d56252a3a..878eb21ee 100644
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -361,8 +361,8 @@ def wrap_subprocess_call(
             # spawning a `conda activate` process at wrapper runtime. We
             # ensure that a sentinel conda-meta directory exists to pass
             # activation validation.
-            conda_meta_dir = Path(prefix) / "conda-meta"
-            conda_meta_dir.mkdir(parents=True, exist_ok=True)
+            # conda_meta_dir = Path(prefix) / "conda-meta"
+            # conda_meta_dir.mkdir(parents=True, exist_ok=True)
 
             try:
                 _context.changeps1 = False
```

I have the result:
```
>>> from conda.utils import wrap_subprocess_call
>>> from conda.base.context import context
>>> wrap_subprocess_call(context.root_prefix, "/tmp/prefixdoesnotexistsuggestion", False, None, ["python", "--vers
ion"])
  File "/home/sophia/projects/conda/conda/deprecations.py", line 193, in inner
    return func(*args, **kwargs)
  File "/home/sophia/projects/conda/conda/utils.py", line 378, in wrap_subprocess_call
    activate_code = activator.activate()
  File "/home/sophia/projects/conda/conda/activate.py", line 175, in activate
    builder_result = self.build_activate(self.env_name_or_prefix)
  File "/home/sophia/projects/conda/conda/activate.py", line 324, in build_activate
    return self._build_activate_stack(env_name_or_prefix, False)
  File "/home/sophia/projects/conda/conda/activate.py", line 336, in _build_activate_stack
    raise EnvironmentLocationNotFound(prefix)
conda.exceptions.EnvironmentLocationNotFound: Not a conda environment: /tmp/prefixdoesnotexistsuggestion
```

conda/utils.py
```diff
@@ -245,7 +240,54 @@ def wrap_subprocess_call(
...
+
+            try:
+                _context.changeps1 = False
```
soapy1: What is the motivation for setting this `changeps1` configuration for this section of code? 
I don't think it's really good to be modifying the context like this. Another option might be to use this [fresh_context](https://github.com/conda/conda/blob/main/conda/base/context.py#L2077).

conda/utils.py
jaimergp: I agree with Sophia here. The behavior in `main` is clearly a bug.

conda/activate.py
```diff
@@ -335,8 +334,6 @@ def _build_activate_stack(self, env_name_or_prefix, stack):
                 from .exceptions import EnvironmentLocationNotFound
 
                 raise EnvironmentLocationNotFound(prefix)
-        elif env_name_or_prefix in (RESERVED_ENV_NAMES):
-            prefix = context.root_prefix
```
jaimergp, line 339: Nice opt. This is being done in `locate_prefix_by_name` already 🚀

conda/utils.py
jaimergp: I think it may be an attempt to save some code that won't have any impact. I would be surprised if this results in a measurable improvement, though.

conda/utils.py
```diff
@@ -245,7 +240,54 @@ def wrap_subprocess_call(
...
+
+                activator = activator_cls(activator_args)
+                activator._parse_and_set_args()
+                # CmdExeActivator writes an .env file by default; let's force in-memory output here
+                # so that we can embed the activation output directly into our wrapper script.
+                activator.tempfile_extension = None
+                activate_env = activator.activate()
+            finally:
+                _context.__dict__.pop("changeps1", None)
+                _context.__dict__.pop("dev", None)
```
jaimergp: Hm, why this? This is a destructive action for the global context.

conda/utils.py
```diff
@@ -245,7 +240,54 @@ def wrap_subprocess_call(
...
+                    fh.write(f"{silencer}IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%\n")
+                    continue
+
+                if "=" not in line:
+                    continue
+
+                key, value = line.split("=", 1)
+                fh.write(f'{silencer}SET "{key}="\n') if value == "" else fh.write(
+                    f'{silencer}SET "{key}={value}"\n'
+                )
```
jaimergp: Isn't this the same? If `value == ""`, then the `SET` line is still `"{key}="`.

```suggestion
                fh.write(f'{silencer}SET "{key}={value}"\n')
```

conda/utils.py
```diff
@@ -314,7 +356,32 @@ def wrap_subprocess_call(
...
+                activator_args.append(prefix)
+
+                activator = activator_cls(activator_args)
+                activator._parse_and_set_args()
+                activate_code = activator.activate()
+            finally:
+                _context.__dict__.pop("changeps1", None)
+                _context.__dict__.pop("dev", None)
+
+            fh.write(activate_code)
```
jaimergp, line 353: Does the activate code include `-e`? To make sure the behavior is consistent with Windows' early exit.

conda/utils.py
```diff
@@ -245,7 +240,54 @@ def wrap_subprocess_call(
...
+
+                activator_cls = _build_activator_cls("cmd.exe")
+                activator_args = ["activate"]
+                if dev_mode:
+                    activator_args.append("--dev")
+                activator_args.append(prefix)
+
+                activator = activator_cls(activator_args)
+                activator._parse_and_set_args()
+                # CmdExeActivator writes an .env file by default; let's force in-memory output here
```
jaimergp: I think there were security reasons the .env file was needed. Otherwise the script got blocked by some AV software or something like that. @kenodegard may be able to confirm.

conda/utils.py
jaimergp: cc @agriyakhetarpal: I share the same concerns as Sophia. Why is this needed at all?

jezdez:
This definitely needs a news file (all changes require one, otherwise users won't know what changed). Especially if it fixes bugs, or, like in this case, speeds up things.

conda/utils.py
agriyakhetarpal: This was required because I was trying to fix a failing test. I did that prior to the new year, so let me look up what the failure was...

agriyakhetarpal: Looks like https://github.com/conda/conda/actions/runs/20360253420/job/58504106160?pr=15534#step:11:3117 was the failure. I can take a fresh attempt for fixing this test, as I don't recall why it was failing.

conda/utils.py
agriyakhetarpal: Same; this was required to fix a Xonsh shell test for some reason, looking it up

conda/utils.py
agriyakhetarpal: Thanks for investigating this and finding this bug; I pushed 8cd432928af24b75463228cbae5969d781c1a2c3 and a5d6055b53efe74ba1a94a5cdd27856f3b0a5adf to resolve this!

conda/utils.py
agriyakhetarpal: Good point – it did not; but now does via e48d591ee6ffa370fcc3aab67d5d06d835c7f038. Thanks!

conda/utils.py
agriyakhetarpal: (I marked my above two comments as off-topic; I'm conflating issues.)

I recall that I did this to perform a micro-optimisation indeed: we don't need to bother with shell prompt modifications, as activation happens in a non-interactive subprocess script and changing the prompt has no visible effect. It does not result in a measurable improvement. I'll drop this.

Side note: I poked around with `fresh_context` a bit after this comment, and it turns out it's quite heavyweight because it reinitialises the entire context.

conda/utils.py
agriyakhetarpal: Do you mean #14871? This is what I could find from a quick search, and it's been open for a while now. But it doesn't look related to me.

tests/core/test_initialize.py
```diff
@@ -482,6 +482,7 @@ def test_install_conda_sh(verbose):
 def test_install_conda_fish(verbose):
     with tempdir() as conda_temp_prefix:
         target_path = join(conda_temp_prefix, "etc", "fish", "conf.d", "conda.fish")
+        context.dev = False
```
soapy1, line 485: What is the motivation for setting `context.dev` here?

news/15534-inline-env-activation-conda-run
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* `conda run` activates environments inline instead of using a subprocess, thus speeding up the runtime (#15534)
+
+### Bug fixes
+
+* <news item>
```
soapy1: Probably also worth it to include a note about the bug fix here too.

```suggestion
* Ensure an appropriate error code is returned when `conda run` is invoked with an invalid environment name or prefix. (#15534)
```

soapy1 (approved):
Looks great to me! Nice speed up 🤸

tests/core/test_initialize.py
agriyakhetarpal: The `test_install_conda_sh` test does it, but the corresponding tests for other shells were not doing so – this also fixes the tests that failed here: https://github.com/conda/conda/actions/runs/20972959121/job/60282210013?pr=15534

soapy1: hmmm, it's not good to be setting bits of the context object like this in normal conda code. It might be ok in test? Maybe someone can chime in on that.

Another way to wirte this might be to do
```
monkeypatch.setenv("CONDA_DEV", "false")
reset_context()
```

agriyakhetarpal: Yes, indeed, considering that this is test code that has been there for long, I think it should be fine... I can switch to using the monkeypatch + reset_context if that helps, sure! I'll also see if there are other places we can clean this up. :)

Edit: resolved with b4167b93f81e13ca6253f3ef0c04ef75048c5118. I found two more files where this was present, and a regex search with `^\s*context\.\w+\s*=\s*.+` I did not find any other cases where this kind of change would be applicable.

agriyakhetarpal:
I think the problem with these tests is that the parameter values are in `_cache_`. When `reset_context()` is called, the cache might not be properly cleared. I spent some time on this today, but I'm still trying to wrap my head around it... If I don't find it, I'd suggest we go back to setting `context.dev` because that's how we've had it in those tests for a long time...

conda/utils.py
```diff
@@ -254,7 +248,41 @@ def wrap_subprocess_call(
...
+
+            activator = activator_cls(activator_args)
+            activator._parse_and_set_args()
+            # CmdExeActivator writes an .env file by default; let's force in-memory output here
+            # so that we can embed the activation output directly into our wrapper script.
+            activator.tempfile_extension = None
+            activate_env = activator.activate()
+
+            for line in activate_env.splitlines():
+                line = line.rstrip("\r")
```
kenodegard: `str.splitlines` is supposed to already cover this:

https://docs.python.org/3/library/stdtypes.html#str.splitlines

conda/utils.py
```diff
@@ -254,7 +248,41 @@ def wrap_subprocess_call(
...
+                if not line:
+                    continue
+
+                if line.startswith("_CONDA_SCRIPT="):
+                    script = line.split("=", 1)[1]
+                    fh.write(f'{silencer}CALL "{script}"\n')
+                    # If any of these calls to the activation hook scripts fail, we want
+                    # to exit the wrapper immediately and abort `conda run` right away.
+                    fh.write(f"{silencer}IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%\n")
```
kenodegard: the choice to immediately exit here makes sense, should we do the same in `_conda_activate.bat`?

news/15534-inline-env-activation-conda-run
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* `conda run` activates environments inline instead of using a subprocess, thus speeding up the runtime (#15534)
```
kenodegard:
```suggestion
* Speedup `conda run` by inlining environment activation instead of using a subprocess. (#15534)
```

conda/utils.py
```diff
@@ -254,7 +248,41 @@ def wrap_subprocess_call(
...
+                    continue
+
+                if "=" not in line:
+                    continue
+
+                key, value = line.split("=", 1)
+                fh.write(f'{silencer}SET "{key}={value}"\n')
```
kenodegard: Can we write this to mirror what is done in `_conda_activate.bat`?


```suggestion
                key, value = line.split("=", 1)
                if key == "_CONDA_SCRIPT":
                    fh.write(f'{silencer}CALL "{value}"\n')
                    # If any of these calls to the activation hook scripts fail, we want
                    # to exit the wrapper immediately and abort `conda run` right away.
                    fh.write(f"{silencer}IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%\n")
                elif not value:
                    fh.write(f'{silencer}SET "{key}="\n')
                else:
                    fh.write(f'{silencer}SET "{key}={value}"\n')
```

conda/utils.py
agriyakhetarpal: Nice, I didn't know! Thanks!

Edit: resolved in 6c97cc849026fcff29215612a6c90afa84108da5

conda/utils.py
agriyakhetarpal: Yes, I think we should, though it would be a breaking change – even if it was always broken in that sense. I can change it.

agriyakhetarpal:
> An alternative approach would be to register a new activation class (e.g., `cmd.script`) that would return the activation formatted correctly and avoid reparsing:
> 
> ```python
> class CmdExeRunActivator(CmdExeActivator):
>     tempfile_extension = None
>     unset_var_tmpl = "SET %s="
>     export_var_tmpl = "SET %s=%s"
>     path_var_tmpl = export_var_tmpl
>     set_var_tmpl = export_var_tmpl
>     run_script_tmpl = "CALL %s\nIF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%"
> 
> def wrap_subprocess_call(...):
>     if on_win:
>         with Utf8NamedTemporaryFile(...) as fh:
>             ...
>             fh.write(f"{activator.activate()}\n")
>             ...
> ```
> 

Thanks for the review, @kenodegard! I really like this suggestion; I think it would make sense for us to have cleaner code.

> I also want us to get rid of the `--dev` option, is that something we can stop supporting for `conda run`?

Yes, I think so, as I've had a few run-ins with it since it has broken some tests over the course of this PR. Is it okay to leave that for a subsequent PR?

conda/utils.py
agriyakhetarpal: Added in 9323177666e6088e3a6ac41045f063fdf3b67bdc

agriyakhetarpal:
> > I also want us to get rid of the `--dev` option, is that something we can stop supporting for `conda run`?
> 
> Yes, I think so, as I've had a few run-ins with it since it has broken some tests over the course of this PR. Is it okay to leave that for a subsequent PR?

Looks like I found a TODO for this 😃: https://github.com/conda/conda/blob/97b8bfdf4caa3c75a934cdcbaa39200e7c21d8c0/conda/testing/fixtures.py#L273-L274

conda/utils.py
kenodegard: I went ahead and pulled this into a new PR #15616 since it was related to the recent #15602 and is somewhat separate from this PR

jezdez:
Thanks for all the work on this PR, @agriyakhetarpal! The ~10-19% performance improvement is a solid win.

I've reviewed the current state and agree with the approach of deferring the `--dev` removal to a follow-up PR. Let's land this performance improvement first and tackle that separately to avoid introducing additional risk.

@jaimergp @soapy1 @kenodegard — Could you give this a final look so we can get it merged? All CI checks are passing and the review feedback has been addressed.

tests/core/test_initialize.py
jaimergp, line 485: Shouldn't these be `monkeypatch.setenv("CONDA_DEV", "0"); reset_context()`?

jaimergp: (I do see that other tests are also doing this for `.dev`, even in user-facing code, so don't hold the PR just because of this).

jezdez: Changed in https://github.com/conda/conda/pull/15534/changes/f53f99764f3b3c87c99cdb1126f40f12f1958b22

agriyakhetarpal: Ideally, yes, they should be – @soapy1 raised it here: https://github.com/conda/conda/pull/15534#discussion_r2688511896 but ultimately I had to revert because https://github.com/conda/conda/pull/15534#issuecomment-3752393498

cc: @jezdez since you pushed f53f99764f3b3c87c99cdb1126f40f12f1958b22 just now

agriyakhetarpal: I would suggest reverting it, and I can take a look at it with fresh eyes in a new PR :)

jezdez: Reverted.

jezdez:
@jaimergp @agriyakhetarpal What's the status of #14993 once this PR is merged? The PR description says "it's part of" but I don't see another PR attached to the issue. Can we _close_ the issue once this is merged?

jaimergp:
#14993 proposed two things:

- what this PR has done
- special-casing `conda run` so the plugin system is not used and save a few more ms. In a way, `conda run` is much more like `conda (de)activate`, so things like virtual packages, solvers, etc, do not apply and do not need to be loaded at all.

jezdez:
> #14993 proposed two things:
> 
>     * what this PR has done
> 
>     * special-casing `conda run` so the plugin system is not used and save a few more ms. In a way, `conda run` is much more like `conda (de)activate`, so things like virtual packages, solvers, etc, do not apply and do not need to be loaded at all.

The special-casing seems like a bigger topic to consider (remember the runx convo?), let's open a new ticket with that?

agriyakhetarpal:
Most of what remains in #14993 (besides avoiding the plugin system, which I agree with discussing elsewhere in a new issue) is testing improvements to import time by restructuring module imports. Unfortunately, most of the experiments I performed at the time with my machine yielded very unreliable results, for a host of reasons. I can try them out in CI/CodSpeed to get some reliable numbers.

tests/core/test_path_actions.py
```diff
@@ -189,6 +189,10 @@ def test_CompileMultiPycAction_noarch_python(prefix: Path):
     mkdir_p(dirname(python_full_path))
     create_link(sys.executable, python_full_path, LinkType.softlink)
 
+    # create conda-meta directory like _execute_actions does in real installation
+    conda_meta_dir = join(prefix, "conda-meta")
+    mkdir_p(conda_meta_dir)
+
```
beckermr, line 195: This addition to the test suite is suspicious as well. One thing I found when deugging locally is this error in the messages:

```
DirectoryNotACondaEnvironmentError: The target directory exists, but it is not a conda environment.
Use 'conda create' to convert the directory to a conda environment.
  target directory: /home/runner/miniconda3/envs/test
```

That happens when the conda-meta dir is missing. From the change here, it appears that with the changfes in this PR suddenly this had to be done by hand? I wonder if we coded around a bug here instead of fixing it.

cc @jaimergp @agriyakhetarpal

beckermr:
OK @jaimergp @agriyakhetarpal Per my tests here, this PR is breaking linking+activation in a way that is not captured by the test suite. My guess is that the test suite envs are too simple. You can see my test of this here https://github.com/LSSTDESC/descwl-shear-sims/pull/238 where I patched this PR out live in the CI on a failing case and now it works.

beckermr:
@conda/conda-maintainers I strongly suggest we revert this PR and push a bugfix release ASAP. The bug here is a fundamental one in post-install linking and environment activation. 

These are core functions that have now been broken for ~3 months in a case that is common enough it was found twice independently. Further people are now patching their CI around avoiding conda versions 26.* which is the kind of thing folks used to do many years ago when conda was less stable.

agriyakhetarpal:
Hi @beckermr, thanks for raising this. I'll get on my computer and look at figuring out what the problem is on an immediate basis.

agriyakhetarpal:
From my preliminary investigations, it looks like this is related to the `set -e` changes in the activation hook calls, i.e., the "fail fast if activation fails" logic. I am moving away from my keyboard, but I will be back in a few hours to implement a fix and see if it fully resolves the linked issues. Edit: I think I am making progress and on the right path.

Also, I'll add a test that creates a conda environment with a package that has an `activate.d` script containing a benignly-failing command, plus a package with a post-link script.

jaimergp:
A hypothesis worth considering is that having added `set -e` is revealing an error in activation that had been failing silently for a bit, so the actual root cause may be somewhere else. Food for thought.

beckermr:
@jaimergp I don't think there has been a fundamental error, but I could be wrong of course.

Instead, what I think is happening is that bash/sh doesn't have good support for `try...except` idioms. So when the fail fast was added via `set -e`, things are no longer being cleaned up properly. This would explain the `set -e` might be at issue and the fact that in my failing test case, I saw an activation script being run with the wrong prefix.

beckermr:
@jaimergp @agriyakhetarpal one thing to remember is that this utility function is used in multiple different contexts in the code base, not just for conda run.

beckermr:
I think my hunch is looking close to correct.

Calling conda activate has a line (https://github.com/conda/conda/blob/main/conda/shell/etc/profile.d/conda.sh#L30) that does

```
ask_conda="$(PS1="${PS1:-}" __conda_exe shell.posix "$@")" || \return
```

~and this happens in a subshell so the calling process is not effected.~

~This is effectively a `try...except` idiom in bash, and that idiom was removed when the activate functionality was inlined.~

Edit: I don't think this is quite right actually, but something related it happening.

agriyakhetarpal:
I am trying to verify whether my hypotheses are correct with `gdk-pixbuf`, but I am seeing increased retries and occasional failures when accessing https://conda.anaconda.org/conda-forge/osx-arm64/repodata.json, so I can't test reliably. (I don't think it has to do with my internet connection).

The reproducer `conda create -n myenv -c hep-forge -c conda-forge root root-plus root-guard` as mentioned in https://github.com/conda-forge/gdk-pixbuf-feedstock/issues/61 seems to be `linux-64` specific and doesn't seem to work on `osx-arm64`. Would `conda create -n test-gdk-pixbuf -c conda-forge gdk-pixbuf librsvg -y` be an okay reproducer for `osx-arm64`, or should I be attempting a heavier solve?

beckermr:
I have not been able to reproduce anywhere but a very heavy CI run here: https://github.com/LSSTDESC/descwl-shear-sims/pull/238.

So I am just using that to test things by patching conda directly. :/

beckermr:
@jaimergp @agriyakhetarpal Removing the set commands fixes the issue. 

So I think the fix is in three parts. 

1. Back out this PR to fix the issue. A 20% gain in performance is not worth the bug. 
   Further, while removing the set commands fixes things, we
    - do not understand well enough what the expected behavior was originally to make changes here in the first place
    - do not have a good test case for this bug
    - do not fully understand the bug since it seems to also require big environments (and thus possibly involve a race condition?)

   Thus I do not think a further PR on top of this one is prudent.
2. Recognize that this utility function is used across the code base for different operations. In particular, it is used for both running linking scripts and for `conda run`. These operations may need different error handling semantics.
3. Once we understand 2, put in a proper optimization.

beckermr:
So again @conda/conda-maintainers let me very strongly state that we should revert this PR and push a bugfix release ASAP. The bug here is a fundamental one in post-install linking and environment activation, and further our collective understanding of the code base is what caused this bug to pass code review and be shipped in the first place. We should recognize that and be much more conservative in putting in changes into these operations.

agriyakhetarpal:
Thanks, @beckermr. I am still facing installation errors, so perhaps it's something to do with the CDN and my region, India. I've written up a patch file below with something. Could you please apply it and try it out there?

```patch
From 0ffbcd230c9c73ea482fac3cee499699e6f28824 Mon Sep 17 00:00:00 2001
From: Agriya Khetarpal <74401230+agriyakhetarpal@users.noreply.github.com>
Date: Fri, 10 Apr 2026 16:46:29 +0530
Subject: [PATCH] Testing something

---
 conda/activate.py |  4 +---
 conda/utils.py    | 29 ++++++++++++++++++++++-------
 2 files changed, 23 insertions(+), 10 deletions(-)

diff --git a/conda/activate.py b/conda/activate.py
index 8556c5e9a..675703018 100644
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -1035,9 +1035,7 @@ class CmdExeRunActivator(CmdExeActivator):
     export_var_tmpl = 'SET "%s=%s"'
     path_var_tmpl = export_var_tmpl
     set_var_tmpl = export_var_tmpl
-    # If any of these calls to the activation hook scripts fail, we want
-    # to exit the wrapper immediately and abort `conda run` right away.
-    run_script_tmpl = 'CALL "%s"\nIF %%ERRORLEVEL%% NEQ 0 EXIT /b %%ERRORLEVEL%%'
+    run_script_tmpl = 'CALL "%s"'
 
 
 class FishActivator(_Activator):
diff --git a/conda/utils.py b/conda/utils.py
index 39f6a26c1..bbf98a45b 100644
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -238,17 +238,26 @@ def wrap_subprocess_call(
             old_path = environ.get("PATH", "")
             if condabin_dir not in old_path.split(os.pathsep):
                 environ["PATH"] = f"{condabin_dir}{os.pathsep}{old_path}"
+
+            # Clear activation state so the activator sees a clean
+            # environment (CONDA_SHLVL=0). Otherwise it inherits the
+            # conda process's own activation state and (my hunch: may) compute
+            # wrong PATH entries or emit spurious deactivate scripts.
+            saved_env = {}
+            for key in ("CONDA_SHLVL", "CONDA_PREFIX", "CONDA_DEFAULT_ENV"):
+                if key in environ:
+                    saved_env[key] = environ.pop(key)
+
             try:
                 activator = activator_cls(activator_args)
                 activator._parse_and_set_args()
                 activate_script = activator.activate()
             finally:
                 environ["PATH"] = old_path
+                environ.update(saved_env)
 
             for line in activate_script.splitlines():
                 fh.write(f"{silencer}{line}\n")
-
-            fh.write(f"{silencer}IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%\n")
             if debug_wrapper_scripts:
                 fh.write("echo *** environment after *** 1>&2\n")
                 fh.write("SET 1>&2\n")
@@ -294,9 +303,6 @@ def wrap_subprocess_call(
             raise Exception("No compatible shell found!")
 
         with Utf8NamedTemporaryFile(mode="w", delete=False) as fh:
-            # If any of these calls to the activation hook scripts fail, we want
-            # to exit the wrapper immediately and abort `conda run` right away.
-            fh.write("set -e\n")
             if dev_mode:
                 from . import CONDA_SOURCE_ROOT
 
@@ -323,19 +329,28 @@ def wrap_subprocess_call(
             old_path = environ.get("PATH", "")
             if condabin_dir not in old_path.split(os.pathsep):
                 environ["PATH"] = f"{condabin_dir}{os.pathsep}{old_path}"
+
+            # Clear activation state so the activator sees a clean
+            # environment (CONDA_SHLVL=0). Otherwise it inherits the
+            # conda process's own activation state and (my hunch: may) compute
+            # wrong PATH entries or emit spurious deactivate scripts.
+            saved_env = {}
+            for key in ("CONDA_SHLVL", "CONDA_PREFIX", "CONDA_DEFAULT_ENV"):
+                if key in environ:
+                    saved_env[key] = environ.pop(key)
+
             try:
                 activator = activator_cls(activator_args)
                 activator._parse_and_set_args()
                 activate_code = activator.activate()
             finally:
                 environ["PATH"] = old_path
+                environ.update(saved_env)
 
             fh.write(activate_code)
 
             if debug_wrapper_scripts:
                 fh.write(">&2 echo '*** environment after ***'\n>&2 env\n")
-            # Disable exit-on-error for the user's command so we can capture its exit code.
-            fh.write("set +e\n")
             if multiline:
                 # The ' '.join() is pointless since mutliline is only True when there's 1 arg
                 # still, if that were to change this would prevent breakage.
-- 
2.53.0

```

agriyakhetarpal:
Ah, looks like you posted some further comments, and GitHub's UI didn't refresh. Please allow me some time to read through them and respond. Thanks!

beckermr:
I am happy to test the patch in the mean time while you all digest my comments. :) 

I need to run for an hour or so, but will reply here with the results of my test. I need to ensure I have conda 26.3.1 in my base env in the test as well for this patch above. Right now my patch is removing the set commands from conda 26.1.* and so did not test changes to your subsequent PR.

agriyakhetarpal:
Thank you for testing it!

I read your comments, and I agree that it is important to unblock your and @meiyasan's work. I'll create a PR to the 26.3.x patch line that reverts both [#15534](https://github.com/conda/conda/pull/15534) (this PR) and #15672 (my follow-up to this PR).

Next, I'll put together another PR that combines those two PRs, and applies the fixes from today, provided that they end up working correctly for you, based on your test. That PR can go through stricter peer review, until we are comfortable with merging it. I'd request you to look at that PR as well, at your convenience!

beckermr:
@agriyakhetarpal One thing to note is that when `conda shell.posix` is called, a new python process is created with `context.__init__()` called again. I suspect this bit is related to the subsequent fixes for global state changes that were put in once the hook function was removed. 

See the code here: 

https://github.com/conda/conda/blob/7dde4c75e13b57efcc369055ba9be7771bc53a5f/conda/cli/main.py#L77

This points also relates to the encapsulation issues I mentioned in that other PR. To keep things inlined and the same, you have to put in all of the changes in the current python process to make it look like it is a new process with `context.__init__()` called. Doing that in a way that is not fragile may require some level of work in the `context` object to allow it to deinit or reinit. Someone will need to go look.

beckermr:
> Thanks, @beckermr. I am still facing installation errors, so perhaps it's something to do with the CDN and my region, India. I've written up a patch file below with something. Could you please apply it and try it out there?
> 
> ```diff
> From 0ffbcd230c9c73ea482fac3cee499699e6f28824 Mon Sep 17 00:00:00 2001
> From: Agriya Khetarpal <74401230+agriyakhetarpal@users.noreply.github.com>
> Date: Fri, 10 Apr 2026 16:46:29 +0530
> Subject: [PATCH] Testing something
> 
> ---
>  conda/activate.py |  4 +---
>  conda/utils.py    | 29 ++++++++++++++++++++++-------
>  2 files changed, 23 insertions(+), 10 deletions(-)
> 
> diff --git a/conda/activate.py b/conda/activate.py
> index 8556c5e9a..675703018 100644
> --- a/conda/activate.py
> +++ b/conda/activate.py
> @@ -1035,9 +1035,7 @@ class CmdExeRunActivator(CmdExeActivator):
>      export_var_tmpl = 'SET "%s=%s"'
>      path_var_tmpl = export_var_tmpl
>      set_var_tmpl = export_var_tmpl
> -    # If any of these calls to the activation hook scripts fail, we want
> -    # to exit the wrapper immediately and abort `conda run` right away.
> -    run_script_tmpl = 'CALL "%s"\nIF %%ERRORLEVEL%% NEQ 0 EXIT /b %%ERRORLEVEL%%'
> +    run_script_tmpl = 'CALL "%s"'
>  
>  
>  class FishActivator(_Activator):
> diff --git a/conda/utils.py b/conda/utils.py
> index 39f6a26c1..bbf98a45b 100644
> --- a/conda/utils.py
> +++ b/conda/utils.py
> @@ -238,17 +238,26 @@ def wrap_subprocess_call(
>              old_path = environ.get("PATH", "")
>              if condabin_dir not in old_path.split(os.pathsep):
>                  environ["PATH"] = f"{condabin_dir}{os.pathsep}{old_path}"
> +
> +            # Clear activation state so the activator sees a clean
> +            # environment (CONDA_SHLVL=0). Otherwise it inherits the
> +            # conda process's own activation state and (my hunch: may) compute
> +            # wrong PATH entries or emit spurious deactivate scripts.
> +            saved_env = {}
> +            for key in ("CONDA_SHLVL", "CONDA_PREFIX", "CONDA_DEFAULT_ENV"):
> +                if key in environ:
> +                    saved_env[key] = environ.pop(key)
> +
>              try:
>                  activator = activator_cls(activator_args)
>                  activator._parse_and_set_args()
>                  activate_script = activator.activate()
>              finally:
>                  environ["PATH"] = old_path
> +                environ.update(saved_env)
>  
>              for line in activate_script.splitlines():
>                  fh.write(f"{silencer}{line}\n")
> -
> -            fh.write(f"{silencer}IF %ERRORLEVEL% NEQ 0 EXIT /b %ERRORLEVEL%\n")
>              if debug_wrapper_scripts:
>                  fh.write("echo *** environment after *** 1>&2\n")
>                  fh.write("SET 1>&2\n")
> @@ -294,9 +303,6 @@ def wrap_subprocess_call(
>              raise Exception("No compatible shell found!")
>  
>          with Utf8NamedTemporaryFile(mode="w", delete=False) as fh:
> -            # If any of these calls to the activation hook scripts fail, we want
> -            # to exit the wrapper immediately and abort `conda run` right away.
> -            fh.write("set -e\n")
>              if dev_mode:
>                  from . import CONDA_SOURCE_ROOT
>  
> @@ -323,19 +329,28 @@ def wrap_subprocess_call(
>              old_path = environ.get("PATH", "")
>              if condabin_dir not in old_path.split(os.pathsep):
>                  environ["PATH"] = f"{condabin_dir}{os.pathsep}{old_path}"
> +
> +            # Clear activation state so the activator sees a clean
> +            # environment (CONDA_SHLVL=0). Otherwise it inherits the
> +            # conda process's own activation state and (my hunch: may) compute
> +            # wrong PATH entries or emit spurious deactivate scripts.
> +            saved_env = {}
> +            for key in ("CONDA_SHLVL", "CONDA_PREFIX", "CONDA_DEFAULT_ENV"):
> +                if key in environ:
> +                    saved_env[key] = environ.pop(key)
> +
>              try:
>                  activator = activator_cls(activator_args)
>                  activator._parse_and_set_args()
>                  activate_code = activator.activate()
>              finally:
>                  environ["PATH"] = old_path
> +                environ.update(saved_env)
>  
>              fh.write(activate_code)
>  
>              if debug_wrapper_scripts:
>                  fh.write(">&2 echo '*** environment after ***'\n>&2 env\n")
> -            # Disable exit-on-error for the user's command so we can capture its exit code.
> -            fh.write("set +e\n")
>              if multiline:
>                  # The ' '.join() is pointless since mutliline is only True when there's 1 arg
>                  # still, if that were to change this would prevent breakage.
> -- 
> 2.53.0
> ```

This patch works per CI run: https://github.com/LSSTDESC/descwl-shear-sims/actions/runs/24245679484/job/70791233919?pr=238

So we are getting closer. However, see my recent comments on encapsulation of the various global/activation state changes. We need to do this in a more structured way.

PR #15546, opened 2025-12-24
Fix circular deprecation warning

travishathaway:
### Description

This fixes what I believe to be a circular deprecation error. We are deprecating `PrefixActions` and tell users to instead use `PrefixActionGroup`, but the `PrefixActionGroup` has a deprecation warning telling users to use `PrefixActions` 🤯.

### Checklist - did you ...


conda/core/link.py
```diff
@@ -222,7 +222,6 @@ def __iter__(self) -> Generator[Iterable[ActionGroup], None, None]:
             yield getattr(self, field.name)
 
 
-@deprecated("25.9", "26.3", addendum="Use PrefixActions instead.")
```
soapy1, line 225: Looking back [on the PR that made this change](https://github.com/conda/conda/pull/14754), I think the intention here was to deprecate this `PrefixActionGroup` and replace it with `PrefixActions`. Where the main differences between these two are:
* `PrefixActions` is not a `NamedTuple`, and
* `PrefixActions` has the two additoinal fields `initial_action_groups` and `final_action_groups`

Additionally, it looks like we want to keep the name `PrefixActionGroup`, so that's the motivation for the deprecation on this `PrefixActions`. So like, at the same time as the `PrefixActionGroup` (NamedTuple) gets removed, the `PrefixAction` dataclass can be renamed to `PrefixActionGroup`

ref: https://github.com/conda/conda/pull/14754#discussion_r2091784211

It's pretty confusing to have this type of circular deprecation warning, so it would be good to try to avoid this in the future. But, I think this change will break the intended plan of action for how to phase out the `PrefixActionGroup` named tuple and replace it with a dataclass. Since niether of these classes are drop in replacements for the other, I don't think this proposed change is a safe change to make.

travishathaway: Thanks for the feedback. I'll close this pull request then.

PR #15547, opened 2025-12-26
Demo: read all input environment spec files with an environment spec plugin

soapy1:
### Description

This demo does the following:

1.  Updates the environment model `from_cli` function to use the environment spec plugins. This means that the `conda create/update/install` commands will be able to read all available environment specs. For example: 
```
$ conda  create --file tests/env/support/simple.yml --dry-run
Channels:
 - defaults
Platform: linux-64
Solving environment: done

## Package Plan ##

  environment location: /tmp/tmprj2a4vp8/unused-env-name

  added / updated specs:
    - nltk


 . . .

DryRunExit: Dry run. Exiting.
```

2. Update the `conda env create` and `conda env update` command to use the same install code path as `conda create/install/update` commands

3. Update the `conda.cli.install.install` function to support the full definition of an environment. That includes, support installing external packages (ref: https://github.com/conda/conda/pull/14794), and support exporting environment variables.

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -342,7 +342,7 @@ def install(args, parser, command="install"):
     isinstall = command == "install"
 
     # fail early if using a deprecated option
-    if newenv and args.clone:
+    if newenv and getattr(args, "clone", False):
         deprecated.topic(
             "25.9",
             "26.3",
@@ -362,7 +362,7 @@ def install(args, parser, command="install"):
 
     env = Environment.from_cli(
         args=args,
-        add_default_packages=command == "create" and not args.no_default_packages,
+        add_default_packages=command == "create" and not getattr(args, "no_default_packages", False),
     )
 
     # for 'conda update' make sure:
--- a/conda/cli/main_env_update.py
+++ b/conda/cli/main_env_update.py
@@ -143,13 +143,14 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
 
     prefix = determine_target_prefix(context, args)
     prefix_data = PrefixData(prefix)
+    install_type = "install"
     if prefix_data.is_environment():
         prefix_data.assert_writable()
         if context.protect_frozen_envs:
             prefix_data.assert_not_frozen()
-
-    # CAN'T Check with this function since it assumes we will create prefix.
-    # cli_install.check_prefix(prefix, json=args.json)
-
-    install(args, parser, "install")
+    else:
+        install_type = "create"
+    
+    install(args, parser, install_type)
+        
     return 0


conda/cli/install.py
```diff
@@ -521,40 +527,60 @@ def revert_actions(prefix, revision=-1, index: Index | None = None):
     return UnlinkLinkTransaction(setup)
 
 
-def handle_txn(unlink_link_transaction, prefix, args, newenv, remove_op=False):
+def handle_txn(unlink_link_transaction, prefix, args, newenv, remove_op=False, env=None):
```
soapy1, line 555: I think `handle_txn` has too many responsibilities. It must decide if it should execute transactions and output some results (but not others) to the user. As a result, in order to get this demo working, this function has lots of hacks.
It would be better to seperate the concerns of executing the transactions and user output. This will take some refactoring here and in all of the callers of this function.

conda/cli/main_env_create.py
```diff
@@ -68,144 +71,71 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
     add_parser_platform(p)
 
-    p.set_defaults(func="conda.cli.main_env_create.execute")
+    # HACK: recreate the default environment file is `environment.yml` behaviour
+    env_yml_default_file = Path(expand("environment.yml"))
+    if env_yml_default_file.is_file():
+        p.set_defaults(file=["environment.yml"])
+   
+    p.set_defaults(
+        func="conda.cli.main_env_create.execute",
```
soapy1, line 87: There are some extra pre-processing of commands and validations that needs to be done in order to reconcile differences between the `conda create` and `conda env create` apis.

conda/cli/install.py
```diff
@@ -521,40 +527,60 @@ def revert_actions(prefix, revision=-1, index: Index | None = None):
...
+        except SystemExit as e:
+            raise CondaSystemExit("Exiting", e)
+
+    if env is not None:
+        # Is it ok to set environment variables and install external packages
+        # after handle_txn? handle_txn includes outputing a success message
+        # if the --json flag is provided.
+
+        # HACK: install pip packages
+        if env.external_packages.get("pip"):
```
soapy1, line 597: Will definitely need a better solution for installing `external_packages`. This `external_packages` idea comes from when we were introducing the environment model. We have maybe moved on from that idea, and it would be good to consider what this means as we continue to work on the `conda-pypi` project.

There is some prior work on making an installer plugin hook https://github.com/conda/conda/pull/14794. This is different from a plugin hook allowing conda to install different package types.

soapy1:
This feature is implemented!

PR #15552, opened 2025-12-29
Removing use of deprecated `codec.open`

travishathaway:
### Description

This is part of Python 3.14 compatibility. We recently decided to trigger errors with our test when use of deprecated code has been detected. This pull request removes the deprecated `codec.open` function in favor of `open` as suggested by the official documentation (https://docs.python.org/3/library/codecs.html#codecs.open).

### Checklist - did you ...

### DIFF
--- a/news/15552-remove-codec-open-use
+++ b/news/15552-remove-codec-open-use
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Remove use of deprecated `codecs.open` (deprecated since Python 3.14)
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/core/package_cache_data.py
```diff
@@ -554,7 +553,7 @@ def __iter__(self):
         return iter(self._urls_data)
 
     def add_url(self, url):
-        with codecs.open(self.urls_txt_path, mode="ab", encoding="utf-8") as fh:
+        with open(self.urls_txt_path, mode="a", encoding="utf-8") as fh:
```
soapy1, line 556: Looking at the documentation, the[ built in `open` function has a default value of `None` for the option `errors`](https://docs.python.org/3/library/functions.html#open) argument: On the other hand, `codecs.open` has a default value of `strict`. It looks like the built in `open` supports the same type of `strict` error handling.

To keep the same type of behaviour, should this also be updated to set `errors="strict"`?

```suggestion
        with open(self.urls_txt_path, mode="a", encoding="utf-8", errors="strict") as fh:
```

travishathaway: I looked at the Python documentation a little more and found this:

> 'strict' to raise a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) exception if there is an encoding error. The default value of None has the same effect.

So, I think `None` works too. Thanks for helping me look into this though. The documentation is very light on resources regarding migrating from `codecs.open` to `open`, so any extra scrutiny here is more than welcome.

PR #15591, opened 2026-01-13
Decode base64-encoded SHA256 in Python metadata

jaimergp:
### Description

The Python specification [says this field is base64 encoded](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file) but we were not decoding it. Do it so we have consistency in the SHA256 fields and `rattler` doesn't complain when parsing these records.

`=`-padding is needed to avoid this :

```
binascii.Error: Invalid base64-encoded string: number of data characters (1) cannot be 1 more than a multiple of 4
```

### Checklist - did you ...


dholth:
Thrilling.

PR #14869, opened 2025-05-21
(DNM, WIP) Adjust test suite for rattler solver

jaimergp:
### Description

WIP.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15592, opened 2026-01-13
Adjust test suite results for `conda-rattler-solver`

jaimergp:
### Description

Supersedes https://github.com/conda/conda/pull/14869. This only includes the testing changes.

### Checklist - did you ...

### DIFF
--- a/conda/testing/fixtures.py
+++ b/conda/testing/fixtures.py
@@ -590,7 +590,7 @@ class TmpChannelFixture:
         noarch.mkdir(parents=True)
 
         noarch_packages: dict[str, dict[str, object]] = {
-            "info": {"subdir": noarch},
+            "info": {"subdir": "noarch"},
             "packages": {},
             "packages.conda": {},
         }


jezdez:
I think expanding the hardcoding of the skip/xfail statements is a little sad, these were workarounds at the time, could we turn them into decorators or pytest markers instead?

jaimergp:
> I think expanding the hardcoding of the skip/xfail statements is a little sad, these were workarounds at the time, could we turn them into decorators or pytest markers instead?

We need them inside the test functions because other fixtures may change the value of `context.solver`, so not sure we can apply that suggestion for many of the changes here.

The bolder move would be to remove the tests that are only for classic, and dump them in `conda-classic-solver` 😎

jezdez:
> > I think expanding the hardcoding of the skip/xfail statements is a little sad, these were workarounds at the time, could we turn them into decorators or pytest markers instead?
> 
> We need them inside the test functions because other fixtures may change the value of `context.solver`, so not sure we can apply that suggestion for many of the changes here.
> 
> The bolder move would be to remove the tests that are only for classic, and dump them in `conda-classic-solver` 😎

Well.. do we have the time to finish classic solver?

jezdez:
I'm wondering if we should finish https://github.com/conda/conda/pull/14170

jaimergp:
> I'm wondering if we should finish https://github.com/conda/conda/pull/14170

I'm more than willing to work on it, but:

- We need to allocate resources in the sprints
- Note that this will mean one more solver in a different repo, which will have to copy over the conda test suite and workflows to run its tests. Not sure what the best direction is here, besides resuming the work that Klaus started with the solver test suite.

jaimergp:
I think the plan forward here is to keep working on [`pytest-conda-solvers`](https://github.com/conda-incubator/pytest-conda-solvers). When that's ready, refactoring classic out would make more sense, and we will be able to also clean up a lot of the solver-only tests here.

In the meantime, I think we are stuck with this approach and I'd rather merge this so `conda-rattler-solver` can depend on `main` directly instead of me coming back to this PR every now and then to sync the branch...

What do you think @jezdez?

conda/testing/solver_helpers.py
```diff
@@ -602,7 +604,7 @@ def test_nonexistent(self, env):
         with pytest.raises((ResolvePackageNotFound, PackagesNotFoundError)):
             env.install("numpy 1.5")
 
-    def test_timestamps_and_deps(self, env):
+    def test_timestamps_and_deps(self, env: SimpleEnvironment):
```
dholth, line 607: How simple is it

PR #15593, opened 2026-01-14
Cache reading yaml and json files

soapy1:
### Description

This PR demos caching all yaml/json file loading.

### Checklist - did you ...

### DIFF
--- a/tests/gateways/test_jlap.py
+++ b/tests/gateways/test_jlap.py
@@ -462,6 +462,9 @@ def test_jlap_sought(
     state["refresh_ns"] = state["refresh_ns"] - int(1e9 * 60)
     cache.cache_path_state.write_text(json.dumps(dict(state)))
 
+    from conda.common.serialize import json as common_json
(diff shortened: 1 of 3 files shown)


tests/gateways/test_jlap.py
```diff
@@ -462,6 +462,9 @@ def test_jlap_sought(
     state["refresh_ns"] = state["refresh_ns"] - int(1e9 * 60)
     cache.cache_path_state.write_text(json.dumps(dict(state)))
 
+    from conda.common.serialize import json as common_json
```
soapy1, line 465: I'm not sure why, but this is required to make the test assertion on line 506 work.

soapy1:
In general, I'm not a fan of this global caching approach. I think different pieces of code have different requirements around managing the lifespan of the cache. For example, condarc files don't change during an invocation of a conda command. But, other files, like bits of repodata _might_ change. I think it would be better to let each piece of code that deals with loading yaml/json decide on what level of caching is appropriate.

PR #15600, opened 2026-01-15
Improve consistency when handling cache files, require UTF-8

jezdez:
### Description

Proposal for #15470.

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -184,10 +184,6 @@ CONDA_LOGS_DIR: Final = ".logs"
 UNKNOWN_CHANNEL: Final = "<unknown>"
 REPODATA_FN: Final = "repodata.json"
 
-# Explicit encoding for cached repodata JSON and state files.
-# Use this constant to ensure consistent UTF-8 encoding across all cache file operations.
-CACHE_ENCODING: Final = "utf-8"
-
 NOTICES_FN: Final = "notices.json"
 """Default name of the notices file on the server we look for."""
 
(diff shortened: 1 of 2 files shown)


dholth:
Thanks. This is OK, but it's not hugely satisfying to have the constant "utf-8". Instead, someday we can require Python 3.15 and be utf-8 all the time :)

For the `*.info.json`, which this doesn't touch AFAICT, `read_text` and `write_text` can't be used because they would drop a lock. Instead, read(), write() must be used on the already- opened file.

I think the issue is rare enough that there should be no fallback to secondary encodings. Imagine a Python that saves to a non-ASCII-compatible encoding by default; would it work with the rest of conda? Instead, fetch again. Could the issue be caused by a different error writing `repodata.json` not related to the default encoding?

Yet another solution if the conda API doesn't require us to produce a str, is to `json.loads(bytes)` which does utf-8 decode inside the json module.

On a normally configured system `repodata.json` is probably saved with [ensure_ascii=True (the default)](https://docs.python.org/3/library/json.html#json.dump), compatible with utf-8 as required in the JSON RFC https://datatracker.ietf.org/doc/html/rfc8259

jezdez:
@dholth How about this, that's a little more defensive in terms of handling corrupt cache files?

dholth:
https://github.com/conda/conda/pull/15471 touches some of the same code.

dholth:
What if instead of "text or bytes" the distinction was "json or bytes"; the utf-8 constant was called `UTF_8="utf-8"`, and a bad text encoding returned an empty json object `{}` instead of making the upstream code.
I wonder how we would handle a valid json that is just a solitary string or number.

PR #15602, opened 2026-01-15
Fix conda activate failing when %TEMP% not writable

### LINKED ISSUE #15601: conda activate fails on Windows when %TEMP% is not writable
### Checklist

### What happened?

On Windows, `conda activate` fails with a cryptic error when the user doesn't have write access to the `%TEMP%` directory. This commonly happens on corporate-managed machines where:

1. User-level `%TEMP%` / `%TMP%` environment variables are not set
2. Only system-level `%TEMP%` exists, but users can't write to it

The error message is:
```
Failed to run 'conda activate ....'.
```

### Root Cause

The issue is in `conda/shell/condabin/_conda_activate.bat`. The script unconditionally uses `%TEMP%`:

```batch
SET "__conda_tmp=%TEMP%\__conda_tmp_%RANDOM%.txt"
```

If the user can't write to `%TEMP%`, the script fails when trying to create this temporary file.

### Suggested Fix

Add a fallback mechanism to use a writable directory. For example:

1. Try `%TEMP%` first
2. Fall back to `%LOCALAPPDATA%\Temp` if `%TEMP%` isn't writable
3. Or use the conda prefix directory as a last resort

### Upstream Report

This was originally reported as a miniforge issue: https://github.com/conda-forge/miniforge/issues/834

### Environment

- OS: Windows (corporate-managed machines)
- conda version: 25.9.1+

pirzada-ahmadfaraz:
## Summary

Fix `conda activate` failing on Windows when users don't have write access to `%TEMP%`.

This commonly happens on corporate-managed machines where:
- User-level `%TEMP%` / `%TMP%` environment variables are not set
- Only system-level `%TEMP%` exists, but users can't write to it

The fix adds a fallback chain in `_conda_activate.bat`:
1. Try `%TEMP%` first (default behavior, most common)
2. Fall back to `%LOCALAPPDATA%\Temp`
3. Fall back to `%USERPROFILE%\AppData\Local\Temp`
4. Fail with clear error message if none are writable

Also uses atomic directory creation (`MKDIR`) with retry loop to handle parallel activation scenarios safely.

## Related Issues

- Fixes #15601
- Upstream from: https://github.com/conda-forge/miniforge/issues/834

## Test Plan

- [ ] Test normal activation with writable `%TEMP%`
- [ ] Test with non-writable `%TEMP%` (should fall back to `%LOCALAPPDATA%\Temp`)
- [ ] Test parallel activation (multiple `conda activate` in parallel)
- [ ] Test with no writable temp (should fail with clear error message)

### DIFF
--- a/tests/shell/test_cmd_exe.py
+++ b/tests/shell/test_cmd_exe.py
@@ -138,6 +138,20 @@ def test_cmd_exe_activate_error(shell: Shell) -> None:
 
 
 @PARAMETRIZE_CMD_EXE
+def test_cmd_exe_activate_invalid_temp(shell: Shell) -> None:
+    """Test that activation fails gracefully with an invalid TEMP directory."""
+    with shell.interactive() as sh:
+        # Set TEMP to invalid path
+        sh.sendline("set TEMP=C:\\path\\that\\does\\not\\exist")
+
+        # Try to activate - should fail with TEMP error
+        sh.sendline(f"conda {activate} base")
+        sh.expect("ERROR: Failed to create temp file")
+        sh.expect("TEMP directory issue")
+        sh.assert_env_var("errorlevel", "1")
+
+
+@PARAMETRIZE_CMD_EXE
 def test_legacy_activate_deactivate_cmd_exe(
     shell_wrapper_integration: tuple[str, str, str],
     shell: Shell,


pirzada-ahmadfaraz:
@conda-bot

PR #15619, opened 2026-01-23
Conda should only include pip packages in `Environment.external_packages["pip"]`

soapy1:
### Description

Currently, when conda builds an `Environment` model from a given prefix using the `conda.models.environment.Environment.from_prefix` function, it will sort packages into 3 categories:
* external_packages
* requested_packages
* explicit_packages

All conda packages installed should be categorized as `explicit_packages`. And all packages installed by other means should be included in `external_packages`.

Currently, conda will correctly categorize packages installed with `pip` into the `external_packages`. However, it also adds those packages to the `explicit_packages` group. This is incorrect, since `explicit_packages` should only represent conda packages.

This PR fixes that by not inserting pip managed packages into the `explicit_packages` group.

part 1 to address: https://github.com/conda-incubator/conda-lockfiles/issues/77

### Checklist - did you ...

### DIFF
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -42,7 +42,7 @@ if TYPE_CHECKING:
 log = getLogger(__name__)
 
 
-EXTERNAL_PACKAGES_PYPI_KEY: Final = "pypi"
+EXTERNAL_PACKAGES_PYPI_KEY: Final = "pip"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


soapy1:
pre-commit.ci autofix

conda/models/environment.py
```diff
@@ -42,6 +42,9 @@
 log = getLogger(__name__)
 
 
+EXTERNAL_PACKAGES_PYPI_KEY: Final = "pip"
```
soapy1, line 45: `"pip"` appears as a magic string a lot of places in conda. Including this constant here to start moving away from that. We'll start just in this module and related tests first.

jezdez (approved):

PR #14607, opened 2025-02-19
Convert dynamic CMD scripts into static script

kenodegard:
### Description

Tinkering...details to come...

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #13975, opened 2024-06-12
Disallow additional characters in prefix names for Windows

jaimergp:
### Description

Closes https://github.com/conda/conda/issues/12558

These characters are problematic in Windows activation.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15628, opened 2026-01-26
Improve Windows cmd.exe special-character env activation

jezdez:
## Summary

This PR expands Windows CMD.EXE special-character coverage and aligns behavior with what is actually reliable in `cmd.exe` activation.

### Test Findings (CMD.EXE)

| Character | Status | Notes |
|-----------|--------|-------|
| `!` (exclamation) | **WORKS** | Fixed by PR #14607's batch script rewrite (removed `EnableDelayedExpansion`) |
| `^` (caret) | **UNSUPPORTED IN CMD ACTIVATE** | Activation now fails fast with a clear error if the target env path contains `^` |
| `=` (equals) | **WORKS** | Activation and prompt display work correctly |
| `%` (percent) | **WORKS** | Activation and prompt display work correctly |
| `(` (open paren) | **WORKS** | Activation and prompt display work correctly |
| `)` (close paren) | **WORKS** | Activation and prompt display work correctly |

### Conclusion

`^` remains fundamentally unsafe in `cmd.exe` activation paths due to parser semantics.  
Instead of brittle escape gymnastics, this PR makes CMD behavior explicit and predictable:

- **CMD.EXE**: reject activation when prefix path contains `^`, with actionable error text.
- **PowerShell / other shells**: unaffected.

This avoids breaking non-CMD workflows while preventing confusing runtime failures in CMD.

The original issue (#12558) reported that `!` couldn't activate environments - this was fixed by PR #14607.

### Background

PR #13975 attempted to fix this by disallowing problematic characters, but was reverted in #14065 because:
- It broke existing environments (e.g., `python=3.12` became unserviceable)
- No deprecation cycle was provided
- It blocked more characters than just the broken ones

### Related

- Issue: #12558
- Reverted PR: #13975
- Revert PR: #14065
- Batch script rewrite: #14607
- Comment with proposed plan: #12558 (comment)

## Test Plan

- [x] Run tests on Windows to characterize actual behavior
- [x] Confirm `!` now works (PR #14607 fixed it)
- [x] Identify '^' as broken in cmd.exe activation and make behavior explicit (clear error)
- [x] Review test results for remaining characters (`=`, `%`, `(`, `)`)
    - **All work!**
- [x] Use findings to determine next steps
    - **Treat `^` as explicit cmd.exe activation error (clear message)**
    - **Preserve non-cmd shell behavior (no global path/name ban)**

### DIFF
--- a/tests/shell/test_cmd_exe.py
+++ b/tests/shell/test_cmd_exe.py
@@ -242,7 +242,7 @@ def test_cmd_exe_special_char_env_activate_by_path(
         sh.sendline(f'conda {activate} "{prefix}"')
         if char == "^":
             sh.expect(
-                r"Cannot activate environments with '^' in their path from cmd.exe"
+                r"Cannot activate environments with '\^' in their path from cmd\.exe\."
             )
             sh.assert_env_var("errorlevel", "2")
             return
@@ -294,7 +294,7 @@ def test_cmd_exe_special_char_prompt_display(
         sh.sendline(f'conda {activate} "{prefix}"')
         if char == "^":
             sh.expect(
-                r"Cannot activate environments with '^' in their path from cmd.exe"
+                r"Cannot activate environments with '\^' in their path from cmd\.exe\."
             )
             sh.assert_env_var("errorlevel", "2")
             return


tests/base/test_context.py
```diff
@@ -648,6 +648,161 @@ def test_channel_alias_validation(value, expected):
...
+#
+# Background:
+# - PR #13975 tried to disallow these characters but was reverted in #14065
+# - The revert happened because it broke existing environments
+# - These tests help plan the proper fix for #12558
+# =============================================================================
+
+
+# Characters that are currently disallowed on all platforms
+ALWAYS_DISALLOWED_CHARS = {"/", " ", ":", "#"}
```
jaimergp: This is the same as `PREFIX_NAME_DISALLOWED_CHARS` in `conda.base.constants`. We should import that instead.

tests/base/test_context.py
```diff
@@ -648,6 +648,161 @@ def test_channel_alias_validation(value, expected):
...
+# These are candidates for future restriction on Windows
+WINDOWS_PROBLEMATIC_CHARS = {
+    "!",  # EnableDelayedExpansion consumes this
+    "^",  # CMD escape character
+    "%",  # Environment variable expansion
+    "=",  # Causes prompt corruption
+    "(",  # Special meaning in CMD
+    ")",  # Special meaning in CMD
+}
+
```
jaimergp: Do you think this is better defined in `conda.base.constants` too?

tests/base/test_context.py
```diff
@@ -648,6 +648,161 @@ def test_channel_alias_validation(value, expected):
...
+
+@pytest.mark.parametrize(
+    "char",
+    list(ALWAYS_DISALLOWED_CHARS),
```
jaimergp: Sort for deterministic parametrization:

```suggestion
    sorted(ALWAYS_DISALLOWED_CHARS),
```

tests/base/test_context.py
```diff
@@ -648,6 +648,161 @@ def test_channel_alias_validation(value, expected):
...
+@pytest.mark.parametrize(
+    "char",
+    [
+        pytest.param("!", id="exclamation"),
+        pytest.param("^", id="caret"),
+        pytest.param("%", id="percent"),
+        pytest.param("=", id="equals"),
+        pytest.param("(", id="open_paren"),
+        pytest.param(")", id="close_paren"),
+    ],
```
jaimergp: This should use the constant defined above. I don't think we need a display name for these (they are ASCII)

tests/base/test_context.py
```diff
@@ -648,6 +648,161 @@ def test_channel_alias_validation(value, expected):
...
+        ("myenv(test)", True),  # Parentheses are problematic
+        ("test!env", True),  # Exclamation is broken on Windows
+        # Invalid names (always disallowed)
+        ("my env", False),  # Space
+        ("my/env", False),  # Forward slash
+        ("my:env", False),  # Colon
+        ("my#env", False),  # Hash
+    ],
+)
+def test_env_name_validation_current_behavior(env_name, should_be_valid):
```
jaimergp: Isn't this test the same as the one in `test_prefix_data`? I don't think we need this, especially given the deprecated function.

tests/core/test_prefix_data.py
```diff
@@ -948,3 +948,163 @@ def test_timestamps(
...
+
+
+@pytest.mark.parametrize(
+    "env_name",
+    [
+        pytest.param("my env", id="space"),
+        pytest.param("my#env", id="hash"),
+        # Note: "/" and ":" are invalid filesystem characters and can't be
+        # tested with real directories. They are tested in test_context.py
+        # using mocks instead.
```
jaimergp: Duplicate comment?

tests/core/test_prefix_data.py
```diff
@@ -948,3 +948,163 @@ def test_timestamps(
...
+    """
+    Test that disallowed characters in env names raise CondaValueError.
+
+    Note: Some characters (like / and :) are also invalid in filesystem paths,
+    so they cannot be tested here with real directories. Those are tested
+    in tests/base/test_context.py using mocks.
+    """
+    env_path = tmp_path / "envs" / env_name
+    env_path.mkdir(parents=True, exist_ok=True)
+    (env_path / "conda-meta").mkdir(exist_ok=True)
```
jaimergp: Redundant. One `mkdir` with parents should be enough.

```suggestion
    (env_path / "conda-meta").mkdir(parents=True, exist_ok=True)
```

tests/core/test_prefix_data.py
```diff
@@ -948,3 +948,163 @@ def test_timestamps(
...
+
+def test_prefix_data_validate_name_base_not_allowed(
+    tmp_path: Path, mocker: MockerFixture
+):
+    """Test that 'base' is rejected when allow_base=False."""
+    # Create a directory named 'base'
+    envs_dir = tmp_path / "envs"
+    base_path = envs_dir / "base"
+    base_path.mkdir(parents=True, exist_ok=True)
+    (base_path / "conda-meta").mkdir(exist_ok=True)
```
jaimergp: Same here.

tests/core/test_prefix_data.py
```diff
@@ -948,3 +948,163 @@ def test_timestamps(
...
+        pd.validate_name(allow_base=False)
+
+
+def test_prefix_data_validate_name_base_allowed(tmp_path: Path, mocker: MockerFixture):
+    """Test that 'base' is accepted when allow_base=True."""
+    # Create a directory named 'base'
+    envs_dir = tmp_path / "envs"
+    base_path = envs_dir / "base"
+    base_path.mkdir(parents=True, exist_ok=True)
+    (base_path / "conda-meta").mkdir(exist_ok=True)
```
jaimergp: And here.

tests/core/test_prefix_data.py
```diff
@@ -948,3 +948,163 @@ def test_timestamps(
...
+    These characters cause issues on Windows CMD.EXE but are not blocked.
+    This test documents current behavior for issue #12558 planning.
+
+    When we implement the fix for #12558, these tests should be updated to
+    expect warnings or errors for these characters on Windows.
+    """
+    env_name = f"test{char}env"
+    env_path = tmp_path / "envs" / env_name
+    env_path.mkdir(parents=True, exist_ok=True)
+    (env_path / "conda-meta").mkdir(exist_ok=True)
```
jaimergp: And here.

tests/shell/test_cmd_exe.py
```diff
@@ -215,3 +215,247 @@ def test_legacy_activate_deactivate_cmd_exe(
...
+
+        # Get the prompt - it should contain the env name in parentheses
+        prompt = sh.get_env_var("PROMPT", "")
+
+        # Log for debugging
+        log.info(f"Prompt with '{special_char}': {prompt!r}")
+
+        # The prompt should start with the env name in parentheses
+        # and should NOT have the corruption pattern
+        expected_prefix = f"({env_name})"
```
jaimergp: Note this can be customized in `condarc`. Ensure `context` is patched to have this syntax.

tests/shell/test_cmd_exe.py
```diff
@@ -215,3 +215,247 @@ def test_legacy_activate_deactivate_cmd_exe(
...
+        sh.sendline(f'conda {activate} "{env_path}"')
+        sh.sendline("echo ACTIVATION_DONE")
+        sh.expect("ACTIVATION_DONE")
+
+        # Verify we're in the environment
+        shlvl = sh.get_env_var("CONDA_SHLVL", "0")
+
+        # Should be at shell level 1 if activation succeeded
+        assert shlvl == "1", (
+            f"Failed to activate existing env with '=' in name. "
```
jaimergp:
```suggestion
            "Failed to activate existing env with '=' in name. "
```

jezdez:
I've posted about the status of this PR in https://github.com/conda/conda/issues/12558#issuecomment-3816793425. I think this is ready for a round of reviews.

conda/activate.py
```diff
@@ -301,24 +301,30 @@ def _parse_and_set_args(self) -> None:
 
         self.command = command
 
+    def template_set_var(self, key: str, value: str) -> str:
+        return self.set_var_tmpl % (key, value)
+
+    def template_run_script(self, script: str) -> str:
+        return self.run_script_tmpl % script
```
jaimergp:
```suggestion
    def render_set_var(self, key: str, value: str) -> str:
        return self.set_var_tmpl % (key, value)

    def render_run_script(self, script: str) -> str:
        return self.run_script_tmpl % script
```

To avoid confusing the other template (`*_tmpl`). That said, I'm not sure these two methods increase readability that much. Personal preference, though, leave them if you want.

conda/activate.py
```diff
@@ -984,6 +990,31 @@ class CmdExeActivator(_Activator):
...
+        return value_str.replace("^", "^^")
+
+    def template_export_var(self, key: str, value: str) -> str:
+        return self.export_var_tmpl % (key, self._escape_cmd_special_chars(value))
+
+    def template_path_var(self, key: str, value: str) -> str:
+        return self.path_var_tmpl % (key, self._escape_cmd_special_chars(value))
+
+    def template_set_var(self, key: str, value: str) -> str:
+        return self.set_var_tmpl % (key, self._escape_cmd_special_chars(value))
```
jaimergp:
```suggestion
    def render_export_var(self, key: str, value: str) -> str:
        return self.export_var_tmpl % (key, self._escape_cmd_special_chars(value))

    def render_path_var(self, key: str, value: str) -> str:
        return self.path_var_tmpl % (key, self._escape_cmd_special_chars(value))

    def render_set_var(self, key: str, value: str) -> str:
        return self.set_var_tmpl % (key, self._escape_cmd_special_chars(value))
```

Ah, I see these now, which probably prompted the existence of the other two. Ok to keep, but let's group them in the same block so they show up together.

conda/activate.py
```diff
@@ -984,6 +990,31 @@ class CmdExeActivator(_Activator):
...
+        When the batch script reads the INI file with FOR /F and uses SET,
+        the ^ character is interpreted as an escape character. We need to
+        double it (^^) to preserve literal ^ characters in paths.
+
+        See: https://github.com/conda/conda/issues/12558
+        """
+        # Convert to string first (e.g., CONDA_SHLVL is an int)
+        value_str = str(value)
+        # ^ must be doubled to be preserved through CMD.EXE SET processing
+        return value_str.replace("^", "^^")
```
jaimergp: This general purpose function may belong to one of the `compat` modules, maybe? But if you want to keep shell-specific, ok to keep here.

jaimergp:
Some more comments; also https://github.com/conda/conda/pull/15628#discussion_r2733306954 and https://github.com/conda/conda/pull/15628#discussion_r2733304049 were marked as resolved without a reply or being addressed 🤔

jezdez:
> Some more comments; also [#15628 (comment)](https://github.com/conda/conda/pull/15628#discussion_r2733306954) and [#15628 (comment)](https://github.com/conda/conda/pull/15628#discussion_r2733304049) were marked as resolved without a reply or being addressed 🤔

I thought https://github.com/conda/conda/pull/15628/changes/c9fca0f734199b218eccd0f39ca87d4022b1d517 was enough 🤷🏻

conda/activate.py
jezdez: I think this is fine..

conda/activate.py
jezdez: [`ecea00e` (this PR)](https://github.com/conda/conda/pull/15628/commits/ecea00e367ecbdaf2fd68e4522882711fcbe1fcf)

conda/activate.py
jezdez: [`8fe052b` (this PR)](https://github.com/conda/conda/pull/15628/commits/8fe052bd14dc07c525aa1a95e9b130a037b3b0f2)

conda/activate.py
```diff
@@ -214,24 +214,30 @@ def execute(self):
         context.plugin_manager.invoke_post_commands(self.command)
         return response
 
-    def template_unset_var(self, key: str) -> str:
+    def render_unset_var(self, key: str) -> str:
```
kenodegard: Can we not rename these? Otherwise we'd need to deprecate

jezdez:
Addressed Ken's feedback in 16a2da8b8 - reverted the method renames back to the original `template_*` naming convention to avoid needing deprecation. The two new methods are now `template_set_var` and `template_run_script` to match the existing convention.

tests/core/test_prefix_data.py
```diff
@@ -956,3 +962,149 @@ def test_timestamps(
...
+    (env_path / "conda-meta").mkdir(parents=True, exist_ok=True)
+
+    pd = PrefixData(env_path)
+
+    if should_raise:
+        with pytest.raises(CondaValueError):
+            pd.validate_name()
+    else:
+        # Should not raise
+        pd.validate_name()
```
kenodegard: `contextlib.nullcontext` is good for this:

```suggestion
    with pytest.raises(CondaValueError) if should_raise else nullcontext():
        pd.validate_name()
```

conda/activate.py
```diff
@@ -984,6 +990,31 @@ class CmdExeActivator(_Activator):
...
+
+    def template_export_var(self, key: str, value: str) -> str:
+        return self.export_var_tmpl % (key, self._escape_cmd_special_chars(value))
+
+    def template_path_var(self, key: str, value: str) -> str:
+        return self.path_var_tmpl % (key, self._escape_cmd_special_chars(value))
+
+    def template_set_var(self, key: str, value: str) -> str:
+        return self.set_var_tmpl % (key, self._escape_cmd_special_chars(value))
+
```
ryanskeith: It may be fairly safe to leave this out. If I understand correctly, I think we may need to add this for this case:  `test^env` as env name? **shudder**

```suggestion

  def template_run_script(self, script: str) -> str:
      return self.run_script_tmpl % self._escape_cmd_special_chars(script)
 
```

ryanskeith (commented):
I am not sure if it really will affect things but we may consider applying the special character fix to template_run_script in the CmdExeActivator.

conda/activate.py
jezdez: Good catch, you're right. The script paths from activate.d/deactivate.d would also contain the caret if the prefix does. Added in 373e92004.

news/12558-special-char-tests
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix Windows CMD.EXE activation failing for environment paths containing `^` (caret) character. The caret was being consumed as an escape character during batch script processing. (#12558)
```
kenodegard:
```suggestion
* Fix Windows CMD.EXE activation failing for environment paths containing `^` (caret) character. The caret was previously consumed as an escape character during batch script processing. (#12558 via #15628)
```

news/12558-special-char-tests
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Add comprehensive test cases to characterize environment name validation behavior for Windows special characters (`!`, `=`, `^`, `%`, `(`, `)`). Tests confirm that `!` now works (fixed by PR #14607's batch script rewrite) and identify `^` as broken. (#12558)
```
kenodegard:
```suggestion
* Add comprehensive test cases to characterize environment name validation behavior for Windows special characters (`!`, `=`, `^`, `%`, `(`, `)`). Tests confirm that `!` now works (previously fixed in #14607) and identify `^` as broken. (#12558 via #15628)
```

tests/shell/test_cmd_exe.py
```diff
@@ -64,7 +69,7 @@ def test_cmd_exe_basic_integration(
 
         sh.sendline("chcp")
         sh.clear()
-        sh.assert_env_var("PROMPT", "(charizard).*")
```
kenodegard, line 67: This is a great example of how some of our assertion have been accidentally testing for the wrong thing.

The `assert_env_var` has two modes, a regex mode (the default) and an explicit mode (when `use_explicit=True` is passed).

Previously this assertion was checking to see if the `PROMPT` _contained_ the prefix basename, **not** that the prompt matched the `(...)` format.

kenodegard:
These latest failures makes me suspect we don't actually need to escape `^`

cc @ryanskeith @jezdez

kenodegard:
https://ss64.com/nt/syntax-esc.html

kenodegard:
Based on my testing we cannot support caret characters in the path while using quotes, no number of caret escapes + quotes will work:

```
❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.

❯ CALL "C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^^^^^^^^^that\etc\conda\activate.d\openssl_activate.bat"
The system cannot find the path specified.
```

only when unquoted and escaped does it work:

```
❯ CALL C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^that\etc\conda\activate.d\openssl_activate.bat
The system cannot find the path specified.

❯ CALL C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\this^^that\etc\conda\activate.d\openssl_activate.bat
# passes
```

So if we insist on supporting caret in CMD.exe we have to escape spaces and all other special chars. This is a brittle choice so I think the better choice is to raise an exception when trying to activate a prefix with a ^ in CMD.exe, if its necessary the user can switch to Powershell.

kenodegard:
It is now still possible to create an env with caret:

```
❯ conda create -n env^^^^with^^^^caret
Channels:
 - defaults
Platform: win-64
Solving environment: done

## Package Plan ##

  environment location: C:\Users\kodegard\.conda\devenv\Windows\envs\devenv-3.10-miniconda\envs\env^with^caret

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
#
#     $ conda activate env^with^caret
#
#
#     $ conda deactivate
```

But attempting to activate an env with caret from CMD.exe will now error:

```
❯ conda activate env^^^^^^^^with^^^^^^^^caret

CondaValueError: Cannot activate environments with '^' in their path from cmd.exe.
Use PowerShell or a caret-free environment path.

ERROR: 'conda activate env^with^caret' exited with code 1.
```

news/12558-special-char-tests
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix Windows CMD.EXE activation failing for environment paths containing `^` (caret) character. The caret was previously consumed as an escape character during batch script processing. (#12558 via #15628)
```
kenodegard:
```suggestion
* Improve Windows CMD.EXE activation failure for environment paths containing `^` (caret) character. (#12558 via #15628)
```

news/12558-special-char-tests
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Improve Windows CMD.EXE activation failure for environment paths containing `^` (caret) character. (#12558 via #15628)
```
kenodegard:
```suggestion
* Improve Windows CMD.EXE activation failure for environment paths containing `^` (caret) character. (#12558 via #15628)

### Bug fixes

* <news item>
```

kenodegard:
Failures are duplicate of https://github.com/conda/conda/issues/15755

tests/shell/test_cmd_exe.py
```diff
@@ -213,5 +215,140 @@ def test_legacy_activate_deactivate_cmd_exe(
...
+            sh.expect(
+                r"Cannot activate environments with '^' in their path from cmd.exe"
+            )
+            sh.assert_env_var("errorlevel", "2")
+            return
+
+        sh.clear()
+
+        # Get the prompt - it should contain the env name in parentheses
+        expected = f"{modifier}{prompt}".format(default_env=prefix)
```
ryanskeith, line 305: This feels weird to do both f-string and the .format at the same time.  Not necessary but perhaps we just need to do one format thing.
```suggestion
        expected = modifier.format(default_env=prefix) + prompt
```

tests/core/test_prefix_data.py
```diff
@@ -990,3 +991,77 @@ def test_conda_package_recognized_windows(empty_env, change_case):
...
+        ("(", False),
+        (")", False),
+        # Invalid chars, see PREFIX_NAME_DISALLOWED_CHARS
+        # ("/", True),
+        (" ", True),
+        # (":", True),
+        ("#", True),
+    ],
+)
+def test_prefix_data_validate_name(
```
ryanskeith, line 1020: It threw me for a second that the ^ wasn't throwing an error I see this is testing the general case and the cmd test is in another location.

tests/shell/test_cmd_exe.py
kenodegard: I find both styles equally unsatisfying so I'll just leave it as is

PR #15636, opened 2026-01-27
Add S3 download optimization to avoid extra file copy

crowecawcaw:
### Description
The S3 adapter currently downloads packages to a temporary file before streaming/copying them to their final location. It's a clever way of fitting S3 support into the request streaming adapter interface, but it requires packages to be copied. For large packages, the copy time can be significant.

This change introduces a new method to download a package from S3 directly to its final location to avoid the intermediate copy.

I ran a quick benchmark to compare download performance before and after this change. I downloaded 1GB and 10GB packages on a c6a.8xlarge EC2 instance from an S3 bucket in the same region. The optimization reduced download time from 6.0s to 4.2s (1.44x speedup) for a 1GB package and from 78s to 35s (2.27x speedup) for a 10GB package.

### Checklist - did you ...

- [n/a] Add / update outdated documentation?


jezdez:
@conda-bot check

conda/gateways/connection/adapters/s3.py
```diff
@@ -49,6 +52,46 @@ def send(
...
+
+        boto_callback = None
+        if progress_callback and size:
+            downloaded = [0]
+
+            def boto_callback(bytes_transferred):
+                downloaded[0] += bytes_transferred
+                progress_callback(downloaded[0] / size)
+
+        key.download_fileobj(fileobj, Callback=boto_callback)
```
jezdez: Consider adding exception handling to match the behavior of `_send_boto3()`. Currently, `BotoCoreError` and `ClientError` will propagate unhandled, whereas the existing code path catches them and returns a 404 response with error details.

```suggestion
        try:
            key.download_fileobj(fileobj, Callback=boto_callback)
        except (BotoCoreError, ClientError) as e:
            # Re-raise with conda-specific error for consistent error handling
            raise CondaError(
                "Error downloading file from S3: %(url)s\n%(exception)s",
                url=url,
                exception=repr(e),
            ) from e
```

jezdez:
Hi @crowecawcaw! Thanks for this contribution - the S3 download optimization is a great improvement! 🎉

I did a review and found it easier to submit the suggested changes as a PR against your branch rather than inline comments. Please take a look at https://github.com/crowecawcaw/conda/pull/3

The main changes are:
- Refactored the adapter detection to use a `DirectDownloadAdapter` Protocol instead of `hasattr()` for cleaner abstraction
- Added proper `CondaError` wrapping for boto3 exceptions
- Added test skips for libmamba solver due to a URL construction bug we discovered (see https://github.com/conda/conda-libmamba-solver/pull/866 and the fix in https://github.com/conda/conda-libmamba-solver/pull/867)

Feel free to merge that PR into your branch, or cherry-pick the changes you like. Let me know if you have any questions!

jezdez:
pre-commit.ci autofix

conda/gateways/connection/adapters/s3.py
```diff
@@ -49,6 +52,67 @@ def send(
...
+        """
+        from ....exceptions import CondaError
+
+        try:
+            from boto3.session import Session
+        except ImportError:
+            raise CondaError(
+                "boto3 is required for S3 channels. "
+                "Please install with `conda install boto3`\n"
+                "Make sure to run `conda deactivate` if you are in a conda environment."
```
jaimergp: We can also simplify this to
```suggestion
                "Please install with `conda install -n base boto3`"
```

conda/gateways/connection/__init__.py
```diff
@@ -16,3 +20,32 @@
...
+        size: int | None = None,
+    ) -> None:
+        """Download directly to a file object.
+
+        :param url: URL to download from
+        :param fileobj: File object to write to (binary write mode)
+        :param progress_callback: Optional callback(fraction) where fraction is 0.0-1.0
+        :param size: Content length (required for progress reporting)
+        """
+        ...
```
jaimergp:
```suggestion
```

Not needed, the docstring is itself a valid block.

PR #15641, opened 2026-01-27
Deprecate `conda run --dev` option

agriyakhetarpal:
### Description

I have marked the `--dev` option for the `conda run` CLI as pending deprecation ~and added a deprecation warning for `wrap_subprocess_call` for the `dev_mode` positional argument via `conda.deprecations.deprecated.topic`~. This simplifies `CondaCLIFixture._cast_args()` to just cast arguments to strings, and removes the automatic injection of `--dev`, as I believe that is already handled by the `PYTHONPATH` test fixture.

The corresponding `dev_mode` argument is used in `conda/core/link.py` right now for pre/post link scripts, so this deprecation is more about the user-facing `--dev` CLI option and not the internal mechanism, but please let me know if you would like me to change this.

This was brought over from the discussion in #15534.

cc: @kenodegard

### Checklist - did you ...


conda/cli/main_run.py
```diff
@@ -46,7 +47,12 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
 
     p.add_argument(
         "--dev",
-        action=NullCountAction,
+        action=deprecated.action(
+            "26.1",
+            "27.3",
```
agriyakhetarpal, line 52: Is it okay to tighten this deadline, perhaps, since IMO this is not something we've been intentionally exposing to the user?

conda/utils.py
```diff
@@ -211,6 +211,14 @@ def wrap_subprocess_call(
     debug_wrapper_scripts,
     arguments: Sequence[str],
 ):
+    if dev_mode:
+        deprecated.topic(
+            "26.1",
+            "27.3",
```
agriyakhetarpal: I have the same question here

conda/cli/main_run.py
jezdez: There is no wiggle room in the current CEP process; we've published this as a user-facing feature years ago, before the deprecation policy even existed. So we need to make sure it's not dropped out of thin air.

agriyakhetarpal: Got it, I'll switch!

jaimergp:
Thanks! Side note, there's a lot of logic in the activation scripts about "dev" mode that I'm not sure why we want to maintain. It's mostly for debugging, but I don't think anyone has used it in a long time, and we can always patch things locally to investigate whatever issue. It creates extra complexity for little gain, so if someone wants to give it a try, I think it's good for removal. Let's open an issue to discuss just in case.

PR #15642, opened 2026-01-27
Make TEMP paths adjustments on Windows more robust

agriyakhetarpal:
### Description

This PR is a safer version of the change in #15453 and now triggers when the current temp drive differs from the prefix drive such that downstream projects like the conda-libmamba-solver don't need to change to the D: drive for their testing just yet. I also added `TMP` just so that there are no surprises.

This change makes the test pass, as can be seen here: https://github.com/conda/conda-libmamba-solver/actions/runs/21403900422/job/61622352051?pr=860#step:15:1336

cc: @dholth

### Checklist - did you ...

### DIFF
--- a/conda/base/constants.py
+++ b/conda/base/constants.py
@@ -179,10 +179,6 @@ MAX_CHANNEL_PRIORITY: Final = 10000
 
 CONDA_PACKAGE_EXTENSION_V1: Final = ".tar.bz2"
 CONDA_PACKAGE_EXTENSION_V2: Final = ".conda"
-CONDA_PACKAGE_EXTENSIONS: Final = (
-    CONDA_PACKAGE_EXTENSION_V2,
-    CONDA_PACKAGE_EXTENSION_V1,
-)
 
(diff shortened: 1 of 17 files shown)


.github/workflows/tests.yml
```diff
@@ -103,6 +103,7 @@ jobs:
         run: |
           mkdir "D:\\Temp"
           echo "TEMP=D:\\Temp" >> $env:GITHUB_ENV
+          echo "TMP=D:\Temp"  >> $env:GITHUB_ENV
```
dholth: no double \\ needed here?

agriyakhetarpal: Typo, sorry!

agriyakhetarpal:
```suggestion
          echo "TMP=D:\\Temp"  >> $env:GITHUB_ENV
```

agriyakhetarpal:
I'm running the checks again because I'm not sure what caused the degraded performance

jezdez (approved):

kenodegard (approved):

Issue #15504, opened 2025-12-11
Proposed extension to `channel_settings` and `ChannelAuthBase` protocol


mattkram:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

## Summary

The `channel_settings` configuration items have two primary fields, `auth` and `channel`, which are currently used during the lookup process to find the appropriate auth handler plugin for a given request (see [here](https://github.com/conda/conda/blob/2af8e0f7255e1d06ea0bfcb6076c7427d101feee/conda/gateways/connection/session.py#L91-L112)). The schema of `channel_settings` is a [sequence of raw mappings](https://github.com/conda/conda/blob/2af8e0f7255e1d06ea0bfcb6076c7427d101feee/conda/base/context.py#L401).

In some cases, it would be useful for the user to be able to provide additional metadata to the plugin to aid in credential lookup or otherwise modifying the plugin's behavior. For example, this pattern is used in `conda-auth` to load `username` for the basic auth handler implementation (see [here](https://github.com/conda-incubator/conda-auth/blob/main/conda_auth/handlers/basic_auth.py#L67)).

However, it is not straightforward to access the `channel_settings` associated with a specific request from within a specific plugin. This has led to at least two implementations of similar behavior:

* In [`conda-auth`](https://github.com/conda-incubator/conda-auth/blob/main/conda_auth/handlers/base.py#L32-L40)
* Our [current proposed implementation](https://github.com/anaconda/anaconda-auth/pull/193/files#diff-2cc953cde15bd28e7570fd1cb39026eaa808838f4d367157dca889b685c14ab1R45-R86) in `anaconda-auth`, which just copies in conda's lookup logic, motivating this feature request.

I would like to propose either of the following approaches, which would help to allow plugin authors to more easily access the specific `channel_settings` that were used to select the specific plugin:

### Alternative 1

The current `ChannelAuthBase` protocol only uses one argument, the `channel_name` (via the [`ChannelNameMixin`](https://github.com/conda/conda/blob/2af8e0f7255e1d06ea0bfcb6076c7427d101feee/conda/plugins/types.py#L214-L234)). This could be extended to pass the `channel_settings` variable as well, which is already readily available [here](https://github.com/conda/conda/blob/2af8e0f7255e1d06ea0bfcb6076c7427d101feee/conda/gateways/connection/session.py#L125).

It could then be made available as `self.channel_settings` or other attribute the plugin class can use.

### Alternative 2

[...]

PR #15646, opened 2026-01-28
Extract helper function for easier retrieval of channel settings from context

mattkram:
### Description

Here, we propose a refactoring to extract a new helper function called `get_channel_settings(channel_name: str, url: str) -> dict[str, Any]` from within the `get_session` function. This will allow plugins that implement the `ChannelAuthBase` interface to be able to more easily access the `channel_settings` entry that was used by conda when the plugin was first selected.

More details for the motivation can be found in #15504.

I have not (yet) added specific unit tests, as this PR is intended to be a pure refactor and is covered by other tests inside `test_connection.py`, however I am happy to do so if it is requested.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/connection/session.py
+++ b/conda/gateways/connection/session.py
@@ -119,6 +119,26 @@ def get_channel_name_from_url(url: str) -> str | None:
 def get_channel_settings(channel_name: str, url: str) -> dict[str, Any]:
     """
     Return the specific item from the `channel_settings` list in the context.
+
+    When iterating through the list of potential items, we first look for an
+    exact match for the channel name, e.g.:
+
+    channel_settings:
+      - name: my-channel
+        auth: some-auth-handler
+
+    If the name does not match exactly, the value will be compared against the
+    request URL, with support for trailing wildcard. The scheme must be identical
+    (i.e. https:// or http://) between the specified pattern and the request URL,
+    to prevent downgrade attacks.
+
+    channel_settings:
+      - name: https://some-domain.com/some-prefix/*
+        auth: some-auth-handler
+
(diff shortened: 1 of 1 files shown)


jezdez:
Hey Matt! Thanks for working on this - making channel settings accessible to plugins is definitely useful.

One thought on the location: since `channel_settings` lives on the `context` object, would it make sense to add this as a method on the `Context` class in `conda/base/context.py` instead? Something like `context.get_channel_settings(channel_name, url)`.

That way:
- The lookup logic lives next to the data it operates on
- Plugin authors get a cleaner import path (they're likely already importing context)
- It keeps session.py focused on session handling

What do you think?

mattkram:
> What do you think?

I think that makes a ton of sense. From an external (plugin author) perspective, that would be a very convenient API, and also I think would be more discoverable than e.g. explicitly importing a helper function. A method seems to be a better approach. 

My initial implementation here was intentionally minimal, as I am less familiar with the overall architecture of conda's internals and didn't want to impose my own design too much without discussion. But I will refactor this a bit to make a method.

The other design aspect that I don't love is the passing of both `channel_name` and `url` arguments, but that was also done for a pure refactor. Perhaps there's some way we could just pass the channel name, since that is what is available inside the auth handler class as an attribute?

The alternative approach I mentioned in the issue would have the selected channel settings dictionary stored as an attribute on the ChannelBase instance. I would suggest this is an even more convenient and backwards-compatible extension to the class, with the caveat that we would need to be careful and likely have exception handling if that is set in the `__init__` method. I'm not sure whether that would add any overhead or be too complex, but since the session is cached it might be worth trying it?

PR #15648, opened 2026-01-29
Fix flaky `test_token_not_present_in_conda_create` test

jezdez:
## Summary

- Fix flaky `test_token_not_present_in_conda_create` integration test that was intermittently failing
- The test previously asserted `/t/<TOKEN>/` would appear in verbose output, but this was unreliable because channel URLs are parsed early and tokens are extracted/stored separately
- Add new tests to properly verify TokenURLFilter integration

## Changes

- Remove the flaky assertion for masked token in output (kept the critical security assertion that raw tokens must not leak)
- Add `test_token_url_filter_attached_to_loggers` to verify the filter is properly attached to conda loggers
- Add parametrized `test_token_url_filter_formats` for comprehensive token URL format coverage

## Root Cause

When a channel URL with a token (e.g., `https://host/t/token123/channel`) is set via `CONDA_CHANNELS`:
1. The token is immediately extracted during channel parsing (`split_scheme_auth_token`)
2. The clean URL (without `/t/token/`) is used for display and most logging
3. Token URLs only appear in HTTP request logging, which may not trigger with cached repodata or `--dry-run` mode

This made the assertion `assert "/t/<TOKEN>/" in all_output` flaky - it depended on specific code paths being hit during the test.

## Test Plan

- [x] Existing unit tests pass
- [x] New parametrized tests cover various token URL formats
- [x] Integration test still verifies the security guarantee (raw tokens don't leak)


soapy1 (approved):
Thanks for including a nice summary of the change! This looks great. Since this just a change to the unit tests, it shouldn't be effecting the codespeed tests. I think the regression reported from codspeed isn't quite right.

PR #15653, opened 2026-01-30
Increase granularity of exceptions raised in situations where packages are not found

### LINKED ISSUE #12197: Consolidate exceptions when a package is not found in the target environment
### Checklist

### What happened?

(Discovered in https://github.com/conda/conda-libmamba-solver/pull/101).

`conda` has several exceptions for "when a package is not found":

* `PackagesNotFoundError`, which can be used for two cases:
  - The package is not found in the target environment (e.g. with `conda update` or `conda remove`). Only takes a list of MatchSpecs or strings as `packages`.
  - If the package is not present in the configured channels (e.g. with `conda install` or `conda create`). This needs the `channels` kwarg.
* `PackageNotInstalledError`, only used for a _single_ package not being present in the target environment. Takes two arguments: prefix a single package name.

The problematic overlap is the "missing from target environment" situation. Both exceptions are used, but it depends on which command:

With `conda update`, conda early exits (as it should) because there's no package to update:

```
$ conda update this-does-not-exist

PackageNotInstalledError: Package is not installed in prefix.
  prefix: /opt/conda
  package name: this-does-not-exist
```

With `conda remove`, it could do the same (there's nothing to uninstall! we can early exit before trying to solve, right?), but it doesn't:

```
$conda remove this-does-not-exist
Solving environment: failed

PackagesNotFoundError: The following packages are missing from the target environment:
  - this-does-not-exist
```

It seems like these are intertwined and would need to be disambiguated. I'd argue for two `PackagesNotFound*` classes:

- `PackagesNotFoundInChannelsError(packages, channels)`: inherit from existing `PackagesNotFoundError` to maintain back-compat in `except` clauses.
- `PackagesNotFoundInPrefixError(packages, prefix)`: inherit from both `PackagesNotFoundError` and `PackageNotInstallerError` to maintain back-compat in `except` clauses.

Additionally, I'd promote the `conda remove` error condition to an early exit check, like it's done in `conda-libmamba-solver` already.

### Conda Info

```shell
active environment : base
            shell level : 1
       user config file : /home/test_user/.condarc
 populated config files : 
          conda version : 22.9.0.post118+ce4e810c9
    conda-build version : 3.23.1
         python version : 3.9.15.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.31=0
                          __linux=5.10.25=0
                          __unix=0=0
       base environment : /opt/conda  (read only)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
                          /home/test_user/.conda/pkgs
       envs directories : /home/test_user/.conda/envs
                          /opt/conda/envs
               platform : linux-64
             user-agent : conda/22.9.0.post118+ce4e810c9 requests/2.28.1 CPython/3.9.15 Linux/5.10.25-linuxkit debian/11 glibc/2.31
                UID:GID : 1001:1001
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
(empty)
```

### Conda list

```shell
# packages in environment at /opt/conda:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main    defaults
_openmp_mutex             5.1                       1_gnu    defaults
anaconda-client           1.11.0           py39h06a4308_0    defaults
attrs                     22.1.0           py39h06a4308_0    defaults
backports                 1.1                pyhd3eb1b0_0    defaults
backports.functools_lru_cache 1.6.4              pyhd3eb1b0_0    defaults
backports.tempfile        1.0                pyhd3eb1b0_1    defaults
backports.weakref         1.0.post1                  py_1    defaults
beautifulsoup4            4.11.1           py39h06a4308_0    defaults
boto3                     1.24.28          py39h06a4308_0    defaults
botocore                  1.27.59          py39h06a4308_0    defaults
brotlipy                  0.7.0           py39h27cfd23_1003    defaults
bzip2                     1.0.8                h7b6447c_0    defaults
c-ares                    1.18.1               h7f8727e_0    defaults
ca-certificates           2022.10.11           h06a4308_0    defaults
certifi                   2022.12.7        py39h06a4308_0    defaults
cffi                      1.15.1           py39h74dc2b5_0    defaults
cfgv                      3.3.1            py39h06a4308_0    defaults
chardet                   4.0.0           py39h06a4308_1003    defaults
charset-normalizer        2.0.4              pyhd3eb1b0_0    defaults
click                     8.0.4            py39h06a4308_0    defaults
clyent                    1.2.2            py39h06a4308_1    defaults
conda                     22.9.0.post118+ce4e810c9           dev_0    <develop>
conda-build               3.23.1           py39h06a4308_0    defaults
conda-content-trust       0.1.3            py39h06a4308_0    defaults
conda-libmamba-solver     22.8.1                   pypi_0    pypi
conda-package-handling    1.9.0            py39h5eee18b_1    defaults
conda-verify              3.4.2                      py_1    defaults
coverage                  6.3.2            py39h7f8727e_0    defaults
cryptography              38.0.1           py39h9ce1e76_0    defaults
curl                      7.85.0               h5eee18b_0    defaults
distlib                   0.3.2              pyhd3eb1b0_0    defaults
docutils                  0.18.1           py39h06a4308_3    defaults
execnet                   1.9.0              pyhd3eb1b0_0    defaults
expat                     2.4.9                h6a678d5_0    defaults
filelock                  3.6.0              pyhd3eb1b0_0    defaults
flake8                    4.0.1              pyhd3eb1b0_1    defaults
flit                      3.6.0              pyhd3eb1b0_0    defaults
flit-core                 3.6.0              pyhd3eb1b0_0    defaults
fmt                       8.1.1                hd09550d_1    defaults
future                    0.18.2           py39h06a4308_1    defaults
gdbm                      1.18                 hd4cb3f1_4    defaults
gettext                   0.21.0               hf68c758_0    defaults
git                       2.34.1          pl5262hc120c5b_0    defaults
glob2                     0.7                pyhd3eb1b0_0    defaults
icu                       58.2                 he6710b0_3    defaults
identify                  2.5.5            py39h06a4308_0    defaults
idna                      3.4              py39h06a4308_0    defaults
importlib-metadata        4.11.3           py39h06a4308_0    defaults
iniconfig                 1.1.1              pyhd3eb1b0_0    defaults
jinja2                    2.11.3             pyhd8ed1ab_2    conda-forge
jmespath                  0.10.0             pyhd3eb1b0_0    defaults
jsonschema                4.16.0           py39h06a4308_0    defaults
jupyter_core              4.11.2           py39h06a4308_0    defaults
krb5                      1.19.2               hac12032_0    defaults
ld_impl_linux-64          2.38                 h1181459_1    defaults
libarchive                3.6.1                hab531cd_0    defaults
libcurl                   7.85.0               h91b91d3_0    defaults
libedit                   3.1.20210910         h7f8727e_0    defaults
libev                     4.33                 h7f8727e_1    defaults
libffi                    3.3                  he6710b0_2    defaults
libgcc-ng                 11.2.0               h1234567_1    defaults
libgomp                   11.2.0               h1234567_1    defaults
liblief                   0.11.5               h295c915_1    defaults
libmamba                  1.0.0                ha06983f_0    defaults
libmambapy                1.0.0            py39ha06983f_0    defaults
libnghttp2                1.46.0               hce63b2e_0    defaults
libsolv                   0.7.22               he621ea3_0    defaults
libssh2                   1.10.0               h8f2d780_0    defaults
libstdcxx-ng              11.2.0               h1234567_1    defaults
libxml2                   2.9.14               h74e7548_0    defaults
lz4-c                     1.9.3                h295c915_1    defaults
markupsafe                2.0.1            py39h27cfd23_0    defaults
mccabe                    0.7.0              pyhd3eb1b0_0    defaults
nbformat                  5.5.0            py39h06a4308_0    defaults
ncurses                   6.3                  h5eee18b_3    defaults
nodeenv                   1.7.0            py39h06a4308_0    defaults
openssl                   1.1.1s               h7f8727e_0    defaults
packaging                 21.3               pyhd3eb1b0_0    defaults
patch                     2.7.6             h7b6447c_1001    defaults
patchelf                  0.13                 h295c915_0    defaults
pcre2                     10.37                he7ceb23_1    defaults
perl                      5.34.0               h5eee18b_2    defaults
pexpect                   4.8.0              pyhd3eb1b0_3    defaults
pip                       22.2.2           py39h06a4308_0    defaults
pkginfo                   1.8.3            py39h06a4308_0    defaults
platformdirs              2.5.2            py39h06a4308_0    defaults
pluggy                    1.0.0            py39h06a4308_1    defaults
pre-commit                2.20.0           py39hf3d152e_1    conda-forge
psutil                    5.9.0            py39h5eee18b_0    defaults
ptyprocess                0.7.0              pyhd3eb1b0_2    defaults
py                        1.11.0             pyhd3eb1b0_0    defaults
py-lief                   0.11.5           py39h295c915_1    defaults
pybind11-abi              4                    hd3eb1b0_0    defaults
pycodestyle               2.8.0              pyhd3eb1b0_0    defaults
pycosat                   0.6.4            py39h5eee18b_0    defaults
pycparser                 2.21               pyhd3eb1b0_0    defaults
pycrypto                  2.6.1           py39h27cfd23_10    defaults
pyflakes                  2.4.0              pyhd3eb1b0_0    defaults
pyopenssl                 22.0.0             pyhd3eb1b0_0    defaults
pyparsing                 3.0.9            py39h06a4308_0    defaults
pyrsistent                0.18.0           py39heee7806_0    defaults
pysocks                   1.7.1            py39h06a4308_0    defaults
pytest                    7.1.2            py39h06a4308_0    defaults
pytest-cov                3.0.0              pyhd3eb1b0_0    defaults
pytest-forked             1.3.0              pyhd3eb1b0_0    defaults
pytest-mock               3.7.0            py39h06a4308_0    defaults
pytest-rerunfailures      10.1               pyhd3eb1b0_0    defaults
pytest-split              0.8.0              pyhd8ed1ab_0    conda-forge
pytest-timeout            1.4.2                      py_0    defaults
pytest-xdist              2.5.0              pyhd3eb1b0_0    defaults
pytest-xprocess           0.21.0             pyhd8ed1ab_0    conda-forge
python                    3.9.15               haa1d7c7_0    defaults
python-dateutil           2.8.2              pyhd3eb1b0_0    defaults
python-fastjsonschema     2.16.2           py39h06a4308_0    defaults
python-libarchive-c       2.9                pyhd3eb1b0_1    defaults
python_abi                3.9                      2_cp39    conda-forge
pytz                      2022.1           py39h06a4308_0    defaults
pyyaml                    6.0              py39h7f8727e_1    defaults
readline                  8.2                  h5eee18b_0    defaults
reproc                    14.2.4               h295c915_1    defaults
reproc-cpp                14.2.4               h295c915_1    defaults
requests                  2.28.1           py39h06a4308_0    defaults
responses                 0.13.3             pyhd3eb1b0_0    defaults
ripgrep                   13.0.0               hbdeaff8_0    defaults
ruamel.yaml               0.17.21          py39h5eee18b_0    defaults
ruamel.yaml.clib          0.2.6            py39h5eee18b_1    defaults
ruamel_yaml               0.15.100         py39h27cfd23_0    defaults
s3transfer                0.6.0            py39h06a4308_0    defaults
setuptools                65.5.0           py39h06a4308_0    defaults
six                       1.16.0             pyhd3eb1b0_1    defaults
soupsieve                 2.3.2.post1      py39h06a4308_0    defaults
sqlite                    3.39.3               h5082296_0    defaults
tk                        8.6.12               h1ccaba5_0    defaults
toml                      0.10.2             pyhd3eb1b0_0    defaults
tomli                     2.0.1            py39h06a4308_0    defaults
tomli-w                   1.0.0              pyhd3eb1b0_0    defaults
toolz                     0.12.0           py39h06a4308_0    defaults
tqdm                      4.64.1           py39h06a4308_0    defaults
traitlets                 5.1.1              pyhd3eb1b0_0    defaults
tzdata                    2022f                h04d1e81_0    defaults
ukkonen                   1.0.1            py39hdb19cb5_0    defaults
urllib3                   1.26.12          py39h06a4308_0    defaults
virtualenv                20.16.2          py39h06a4308_0    defaults
wheel                     0.37.1             pyhd3eb1b0_0    defaults
xdoctest                  1.1.0              pyhd8ed1ab_0    conda-forge
xz                        5.2.6                h5eee18b_0    defaults
yaml                      0.2.5                h7b6447c_0    defaults
yaml-cpp                  0.7.0                h295c915_1    defaults
zipp                      3.8.0            py39h06a4308_0    defaults
zlib                      1.2.13               h5eee18b_0    defaults
zstd                      1.5.2                ha4553b6_0    defaults
```

### Additional Context

Studied in Docker development environment.

agriyakhetarpal:
### Description

This PR closes #12197; I have subclassed `PackagesNotFoundError` to add `PackagesNotFoundInChannelsError` and `PackagesNotFoundInPrefixError` to distinguish two situations: when a spec is not found within configured channels and when it is not found within an environment prefix on one's machine. This way, `PackageNotInstalledError` now (more accurately) derives from `PackagesNotFoundInPrefixError` instead of a more rudimentary `CondaError`.

<details><summary>Details of previous changes that were later split off into their own PRs</summary>
<p>

~As for other changes in this PR (please let me know if you would like me to split these into their own PRs before/after this PR):~
- ~We now exit early for the `conda remove` command when using the classic solver as well, which we previously only did for the CLS.~ (This has been reverted and removed from this PR, and split off into #15697 as suggested by @jaimergp in https://github.com/conda/conda/pull/15653#discussion_r2782856151)
- ~`conda search --envs <package>` now raises a `PackagesNotFoundInPrefixError` when the package is not found in any environment. It was previously just printing the `# Name            Version     Build  Channel        Location ` header with no entries, which wasn't very helpful, and I think having an error is better.~ (This has been reverted and removed from this PR, and split off into #15698 as suggested by @jaimergp in https://github.com/conda/conda/pull/15653#discussion_r2782859339).

</p>
</details> 

### Comparison of outputs

Here is a comparison of the outputs from both `conda remove` and `conda install`.

<details><summary><h4><code>conda remove</code></h4></summary>
<p>

`conda remove` now also outputs the prefix path of the active environment when a package to be removed is reportedly missing.

##### Before

```
PackagesNotFoundError: The following packages are missing from the target environment:

  - this-does-not-exist
```

##### After

```
PackagesNotFoundInPrefixError: The following packages are missing from the target environment:
  prefix: /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge

  - this-does-not-exist
```

</p>
</details>

<details><summary><h4><code>conda install</code></h4></summary>
<p>

When I run `conda install this-does-not-exist`, this is what I see:

##### Before

```
Channels:
 - conda-forge
Platform: osx-arm64
Solving environment: failed

  - this-does-not-exist

Current channels:

  - https://conda.anaconda.org/conda-forge

looking for, navigate to

    https://anaconda.org

```

##### After

```
Channels:
 - conda-forge
Platform: osx-arm64
Solving environment: failed

PackagesNotFoundInChannelsError: The following packages are not available from current channels:

  - this-does-not-exist

Current channels:

  - https://conda.anaconda.org/conda-forge

looking for, navigate to

    https://anaconda.org

```

</p>
</details> 

cc: @jaimergp

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -440,6 +440,7 @@ def install(args, parser, command="install"):
                 # no extra processing for subclasses of PackagesNotFoundError
                 raise
             except PackagesNotFoundError as e:
+                # If a backend raises the generic error, convert it to a more specific one
                 if e.channel_urls:
                     raise PackagesNotFoundInChannelsError(
                         e.packages, e.channel_urls
--- a/conda/cli/main_remove.py
+++ b/conda/cli/main_remove.py
@@ -246,14 +246,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
 
         try:
             txn = solver.solve_for_transaction()
-        # If a backend raises the generic error, we convert it to the more specific one
-        # here. TODO: is this the best place to do this? Should backends raise the specific
-        # errors, say the CLS should? I have thoughts either way but gotta read more about
-        # this. FIXME: handle before merge
         except (PackagesNotFoundInPrefixError, PackagesNotFoundInChannelsError):
             # no extra processing for subclasses of PackagesNotFoundError
             raise
         except PackagesNotFoundError as e:
+            # If a backend raises the generic error, convert it to a more specific one
             packages = getattr(e, "packages", None)
             if not packages:
                 packages = tuple(sorted(str(s) for s in specs))


tests/cli/test_main_search.py
```diff
@@ -177,6 +177,25 @@ def test_search_envs_json(conda_cli: CondaCLIFixture):
...
+def test_search_envs_not_found_json(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundInPrefixError) as excinfo:
+        conda_cli("search", "--envs", "--json", "package-that-does-not-exist-anywhere")
+    assert "package-that-does-not-exist-anywhere" in str(excinfo.value)
+
+
+def test_search_envs_not_found_info(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundInPrefixError) as excinfo:
+        conda_cli("search", "--envs", "--info", "package-that-does-not-exist-anywhere")
+    assert "package-that-does-not-exist-anywhere" in str(excinfo.value)
```
agriyakhetarpal: `--info` doesn't really do anything different right now for packages that are not found, but we can still test it anyway? Please let me know. It's probably good for code coverage, not test coverage, i.e., I am aware that I am artificially inflating the coverage.

agriyakhetarpal:
@travishathaway, to @jaimergp's comment in the issue description in #12197, which I write below:

> Additionally, I'd promote the conda remove error condition to an early exit check, like it's done in conda-libmamba-solver already.

you had responded with the following comment in https://github.com/conda/conda/issues/12197#issuecomment-3184251467:

> it looks like we already exit early when the `conda remove` is called and the package does not exist in the environment. So, I think the only thing left to be done here is simply refactor the exception classes.

I investigated this statement as part of my work on this issue and found that this is true for the libmamba solver, but not for the classic solver. `conda remove <spec> --solver classic` for a non-existent spec still attempted to query the repodata and solve the environment before raising an error. I have changed that behaviour in 6f1911e44a2dda88cca4f5af53f2287c21876a01, and it now exits early. PTAL when you have a chance!

conda/cli/install.py
```diff
@@ -379,12 +381,23 @@ def install(args, parser, command="install"):
...
+    # See https://github.com/conda/conda-libmamba-solver/blob/34aec802f890a3ece9c0f2f630de2ef136003624/conda_libmamba_solver/solver.py#L954-L967
+    _seen = set()
+    repodata_fns = [fn for fn in repodata_fns if not (fn in _seen or _seen.add(fn))]
+    if (
+        context.solver == "libmamba"
+        and not getattr(args, "repodata_fns", None)  # user did not explicitly set them
+        and len(repodata_fns) > 1
+    ):
+        repodata_fns = [REPODATA_FN]
+
```
agriyakhetarpal: I haven't worked through the murky details about this yet, but: some part of this should be in conda-libmamba-solver. We should add a `supports_repodata_fallback` method there (which would just return `False`) and then we should add a `_select_repodata_fn` method or similar to support handling both single and multiple repodata files and selecting the appropriate one. I'll prepare a PR for this, but since that won't get here until we do a release, this should help prevent running the same repodata attempt twice and failing twice before finally surfacing the error.

conda/cli/main_search.py
```diff
@@ -197,6 +197,16 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
             )
+
+        if not ordered_result:
+            from ..exceptions import PackagesNotFoundInPrefixError
+
+            # TODO: improve the prefix indication message? Currently just says "any of the searched environments"
+            # which is not very user friendly and not very informative of it.
+            raise PackagesNotFoundInPrefixError(
+                [spec], prefix="any of the searched environments"
+            )
```
agriyakhetarpal: One idea I have here is to call `list_all_known_prefixes()` and include the number of prefixes plus their names, but it doesn't feel as nice. Would it make sense to say something on the lines of, "use \`conda env list\` to see environments that were searched", or similar? Open to suggestions.

agriyakhetarpal: Or, we could customise this and call `conda env list` after this to report the environments... either way, I am mostly suggesting a quality-of-life improvement.

news/15653-granular-package-not-found-exceptions
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* Add `PackagesNotFoundInChannelsError` and `PackagesNotFoundInPrefixError` as specialised cases of `PackagesNotFoundError`, to distinguish between packages missing from configured channels versus packages missing from a specific environment prefix. (#12197 via #15653)
+* `conda remove <spec>` with the classic solver now performs an early existence check for the spec to be removed. (#15653)
```
jaimergp: This part would need to be in a separate PR. It doesn't just affect the classic solver, but also all invocations of `conda remove` (agnostic to the solver underneath).

news/15653-granular-package-not-found-exceptions
```diff
@@ -0,0 +1,20 @@
...
+
+### Bug fixes
+
+* `conda search --envs <package>` now raises an error when the package is not found in any environment. (#15653)
```
jaimergp: Also inclined to do this separately since it's, technically, a change in behavior we may need to revert or introduce with warnings.

conda/exceptions.py
```diff
@@ -783,13 +774,71 @@ def __init__(
...
+        # specifically "missing from prefix" and not a more general "not found".
+        CondaError.__init__(
+            self,
+            message,
+            packages=packages,
+            packages_formatted=packages_formatted,
+            prefix=prefix,
+            channel_urls=[],
+            channels_formatted="",
+        )
```
jaimergp, line 844: Ugh, this is such a tell-tale of the underlying base class mess.

conda/exceptions.py
```diff
@@ -783,13 +774,71 @@ def __init__(
...
+
+
+class PackageNotInstalledError(PackagesNotFoundInPrefixError):
```
jaimergp: Hm, this is very similar, semantically, to `PackagesNotFoundInPrefixError`. Should we keep both or decide for one of them?

news/15653-granular-package-not-found-exceptions
agriyakhetarpal: Thanks! Happy to move this to a separate PR (I've been wanting to do so too).

news/15653-granular-package-not-found-exceptions
agriyakhetarpal: Yes indeed. This is not related to the original issue; just something I discovered when working on this.

conda/exceptions.py
agriyakhetarpal: Both of them are the same, at least based on where `PackageNotInstalledError` is currently raised in the codebase.

My only reason to keep both has been backwards compatibility, which you raised in the original issue thread – so I'd be inclined to drop `PackagesNotFoundInPrefixError` and keep `PackageNotInstalledError`, while improving its message. Does that work for you?

news/15653-granular-package-not-found-exceptions
agriyakhetarpal: Reverted via 622ee913eb9e1801a854dbe81b49115d7be0b3cd and moved to #15697.

news/15653-granular-package-not-found-exceptions
agriyakhetarpal: Removed via c9198f2340480827e43c8fe233eb0fb5ec1eadc1 and put into #15698

conda/exceptions.py
jaimergp, line 728: Might be worth adding a docstring here suggesting the use of the more specific subclasses.

conda/exceptions.py
```diff
@@ -783,13 +774,71 @@ def __init__(
...
+        # the behaviour based on channel_urls, but is a case of the package(s) being
+        # specifically "missing from prefix" and not a more general "not found".
+        CondaError.__init__(
+            self,
+            message,
+            packages=packages,
+            packages_formatted=packages_formatted,
+            prefix=prefix,
+            channel_urls=[],
+            channels_formatted="",
```
jaimergp: Do we need these two `channel*` kwargs? They don't show up in the `message` template, so not sure why we need to pass them 🤔 Can you elaborate? Same about `packages` (maybe just for exception attribute inspection?).

conda/exceptions.py
jaimergp: I'd be fine with that, yes.

jaimergp: Ah wait, the only difference is whether a single package or more are passed, right? This one only takes one. Maybe the best approach, backwards-compatibility-wise, is to keep the same message for the case of a single package, but use the message from the new subclass for the case when there's more than one case. Maybe you get that for free with `dashlist()` anyway.

agriyakhetarpal: That works!

jaimergp: It'll be a bit weird to have `package_name: str | Iterable[str]`, but it could be worse. We can't rename without breaking API in case someone is using the kwarg name to instantiate this exception.

conda/misc.py
```diff
@@ -318,7 +318,10 @@ def clone_env(prefix1, prefix2, verbose=True, quiet=False, index_args=None):
                 drecs.remove(prec)
                 drecs.add(precs[0])
     if notfound:
-        raise PackagesNotFoundError(notfound)
+        # these specs were not found in the Index (i.e., not available from channels).
+        # index_args["channels"] was set from index_args.pop("channel_urls") above.
+        channel_urls = index_args.get("channels", ()) if index_args else ()
```
jaimergp:
```suggestion
        channel_urls = (index_args.get("channels") or ()) if index_args else ()
```

In case `channels` is defined but set to a scalar falsy, like `None` or `""`.

tests/test_create.py
```diff
@@ -2567,7 +2567,7 @@ def test_remove_spellcheck(
 
     with pytest.raises(
         PackagesNotFoundError,
-        match=r"The following packages are missing from the target environment:\s+- dependint",
+        match=r"The following packages are missing from the target environment:[\s\S]*- dependint",
```
jaimergp: Isn't `[\s\S]` essentially `.`? What's the new error message that requires this change in the regex? 🤔

conda/cli/main_remove.py
jaimergp, line 1: I'd say we should not try to be too prescriptive here and just let the solvers raise what they need to raise. I'd rather submit PRs to raise the more specific exceptions than keep this specialization code around forever.

conda/cli/install.py
```diff
@@ -434,6 +447,20 @@ def install(args, parser, command="install"):
...
+                    raise
+
+                if e.channel_urls:
+                    raise PackagesNotFoundInChannelsError(
+                        e.packages, e.channel_urls
+                    ) from e
+                else:
+                    raise PackagesNotFoundInPrefixError(
+                        e.packages, prefix=prefix
+                    ) from e
```
jaimergp, line 451: Same here as in `main_remove`, let's fix this at the source, we don't have that many solvers anyway.

agriyakhetarpal: Right, though we'd need to merge/release this first and then use it in the CLS code, which is why this exists here

tests/test_create.py
agriyakhetarpal: The error message spans multiple lines, so I loosened this. I'll see if I can simplify this. (not too proud of my regex fu).

conda/cli/install.py
jaimergp: We don't try twice. We have a hack later in the module. Check the `TryRepodata` class for `allow_retry`. I agree it's not the cleanest solution, but it's already there. Also this wouldn't be part of the scope of this PR anyway, so for now let's revert.

conda/cli/main_remove.py
jaimergp: And if we did want to do this, it would go in the `TryRepodata` context manager class.

conda/exceptions.py
agriyakhetarpal: I was wanting to use them later in the multiple-package `str` case, but otherwise they were carried over from the old class. We can drop them now.

conda/cli/install.py
agriyakhetarpal: Thanks! I reverted this for now in 2949837dc29f4fa50b59690d188fed3ae3642f8a, but would like to attempt addressing this again when once this PR merged.

conda/exceptions.py
agriyakhetarpal: I've changed this now. Eventually, I had to make this `package_name: str | Iterable[MatchSpec | PackageRecord | str]`, though.

conda/cli/main_remove.py
jezdez: @jaimergp Are you suggesting for this to _not_ use these new exception classes?

conda/cli/install.py
```diff
@@ -434,6 +436,20 @@ def install(args, parser, command="install"):
                     )
                 else:
                     raise e
+            except PackagesNotFoundError as e:
+                if isinstance(
+                    e, (PackagesNotFoundInChannelsError, PackagesNotFoundInPrefixError)
+                ):
+                    raise
+
```
kenodegard:
```suggestion
            except (PackagesNotFoundInChannelsError, PackagesNotFoundInPrefixError):
                # no extra processing for subclasses of PackagesNotFoundError
                raise
            except PackagesNotFoundError as e:
```

conda/cli/install.py
kenodegard: Could also make `PackagesNotFoundError` into a factory?

conda/cli/main_remove.py
```diff
@@ -228,14 +228,39 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        # If a backend raises the generic error, we convert it to the more specific one
+        # here. TODO: is this the best place to do this? Should backends raise the specific
+        # errors, say the CLS should? I have thoughts either way but gotta read more about
+        # this. FIXME: handle before merge
+        except PackagesNotFoundError as e:
+            if isinstance(
+                e, (PackagesNotFoundInPrefixError, PackagesNotFoundInChannelsError)
+            ):
+                raise
+
```
kenodegard:
```suggestion
        except (PackagesNotFoundInPrefixError, PackagesNotFoundInChannelsError):
            # no extra processing for subclasses of PackagesNotFoundError
            raise
        except PackagesNotFoundError as e:
```

conda/cli/main_remove.py
jaimergp: I'm saying that the solvers themselves should be the ones raising the appropriate exception and we should not need to a lot of conversion. classic is on the way out and we can fix this at the root of libmamba and rattler, so it feels like caring too much about backwards compatibility.

conda/cli/main_remove.py
```diff
@@ -228,14 +228,37 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        try:
+            txn = solver.solve_for_transaction()
+        # If a backend raises the generic error, we convert it to the more specific one
+        # here. TODO: is this the best place to do this? Should backends raise the specific
+        # errors, say the CLS should? I have thoughts either way but gotta read more about
+        # this. FIXME: handle before merge
+        except (PackagesNotFoundInPrefixError, PackagesNotFoundInChannelsError):
+            # no extra processing for subclasses of PackagesNotFoundError
+            raise
+        except PackagesNotFoundError as e:
```
kenodegard:
```suggestion
        except (PackagesNotFoundInPrefixError, PackagesNotFoundInChannelsError):
            # no extra processing for subclasses of PackagesNotFoundError
            raise
        except PackagesNotFoundError as e:
            # If a backend raises the generic error, convert it to a more specific one
```

conda/cli/install.py
```diff
@@ -434,6 +436,18 @@ def install(args, parser, command="install"):
                     )
                 else:
                     raise e
+            except (PackagesNotFoundInChannelsError, PackagesNotFoundInPrefixError):
+                # no extra processing for subclasses of PackagesNotFoundError
+                raise
+            except PackagesNotFoundError as e:
```
kenodegard, line 442:
```suggestion
            except PackagesNotFoundError as e:
                # If a backend raises the generic error, convert it to a more specific one
```

Issue #15657, opened 2026-01-30
conda fails in a strange way if channel name starts with target platform/subdir name


chenghlee:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

conda's internals do some sort of transformation that can cause unexpected failures if the target platform/subdir name (set either automatically as `context.subdir` or explicitly supplied via the `--platform` argument) happens to also be the starting substring of a path component in the channel URL. E.g.,

```
$ conda search -vv --override-channels -c chenghlee/label/linux-riscv64-testing --platform=linux-riscv64
Loading channels: | INFO conda.gateways.repodata.jlap.fetch:download_and_hash(265): Download 116 bytes {'User-Agent': 'conda/25.11.1 requests/2.32.5 CPython/3.13.11 Darwin/24.6.0 OSX/15.7.2 solver/libmamba conda-libmamba-solver/25.11.0 libmambapy/2.3.2 aau/0.7.5 c/QjCGMVBY9O9HMniuNOHNWA s/jVwdZtPmlHGuDoqWN3gnmI e/23bxfdvMcMCTUJDmyyJ3oi a/LznWnNdPTZufYZB5AGl7lw o/kA1VeQZ5QJq0EPqi6AwV8A', 'Accept-Encoding': 'gzip, deflate, br', 'Accept': '*/*', 'Connection': 'keep-alive'}
INFO conda.gateways.repodata.jlap.fetch:request_url_jlap_state(432): Apply 0 patches 45e0e7bc7fed24fc… → 45e0e7bc7fed24fc…
/ INFO conda.gateways.repodata:conda_http_errors(243): Unable to retrieve repodata (response: 404) for https://conda.anaconda.org/chenghlee/label-testing/linux-riscv64/linux-riscv64/repodata.json
done
Traceback (most recent call last):
  File "/Users/clee/Applications/miniconda3/lib/python3.13/site-packages/conda/exception_handler.py", line 29, in __call__
    return func(*args, **kwargs)
  File "/Users/clee/Applications/miniconda3/lib/python3.13/site-packages/conda/cli/main.py", line 61, in main_subshell
    exit_code = do_call(args, parser)
  File "/Users/clee/Applications/miniconda3/lib/python3.13/site-packages/conda/cli/conda_argparse.py", line 205, in do_call
    result = getattr(module, func_name)(args, parser)
[...]

PR #15658, opened 2026-01-30
Make `split_platform(...)` more robust

chenghlee:
### Description

Use the span matched by `_split_platform_re(...)` to remove the platform/subdir path component from `url`, rather than `str.replace(...)`. This prevents bugs if `"/" + platform` happens to be the start of some path element in the channel URL the precedes the actual platform subdirectory name; e.g., if your channel name contains a path element named `${platform}-testing`.

Fixes conda/conda#15657

### Checklist - did you ...


chenghlee:
Note that these changes do not affect conda's current behavior if a known platform's name happens to also be a complete subdirectory's name in the channel's URL; i.e., something like this still happens with this PR:
```
>>> from conda.models.channel import Channel
>>> Channel("a-channel/label/win-arm64/testing/something/new")
Channel("a-channel/label/testing/something/new/win-arm64")
>>> Channel("https://example.com/pkgs/win-64/testing/something/new")
Channel("pkgs/testing/something/new/win-64")
```

conda/common/url.py
```diff
@@ -376,8 +376,14 @@ def split_platform(known_subdirs: Iterable[str], url: str) -> tuple[str, str]:
...
-    platform = _platform_match.groups()[0] if _platform_match else None
-    cleaned_url = url.replace("/" + platform, "", 1) if platform is not None else url
+    if _platform_match:
+        start, stop = _platform_match.span(1)
+        platform = url[start:stop]
+        # "start-1" picks up preceding "/" that's not part of the capturing group
+        cleaned_url = url[: (start - 1)] + url[stop:]
+    else:
+        platform = None
+        cleaned_url = url
```
jezdez, line 386: Do we need to apply the same to this? https://github.com/conda/conda/blob/3e61fd9dd688e5fe67e6156857ea0dc95036f756/conda/common/url.py#L347-L366

chenghlee: Probably, but I suggest we do that as a separate issue + PR. Mostly because I suspect that's going to be a bigger refactor given how tokens need to be handled elsewhere in the code; e.g., it's not super clear to me why that regex is not defined as a constant and allows for the token to be an empty string and/or possibly conflated with a channel name.

chenghlee:
> @chenghlee, I think this is ready to merge. I agree that the scope of the changes need to be kept limited.
> 
> Please let me know if this okay to merge in its current state.

I'm okay with merging in its current state.

conda/common/url.py
```diff
@@ -376,8 +376,14 @@ def split_platform(known_subdirs: Iterable[str], url: str) -> tuple[str, str]:
 
     """
     _platform_match = _split_platform_re(known_subdirs).search(url)
-    platform = _platform_match.groups()[0] if _platform_match else None
-    cleaned_url = url.replace("/" + platform, "", 1) if platform is not None else url
+    if _platform_match:
+        start, stop = _platform_match.span(1)
+        platform = url[start:stop]
+        # "start-1" picks up preceding "/" that's not part of the capturing group
+        cleaned_url = url[: (start - 1)] + url[stop:]
```
kenodegard, line 383: We can avoid the messy start/stop index counting by updating the regex to use named capture groups:

```suggestion
        platform = match.group("platform")
        cleaned_url = match.group("channel") + match.group("path")
```

conda/common/url.py
kenodegard, line 395: Verbose regex with named capture groups:

```suggestion
    _platform_match_regex = rf"""
        (?P<channel>.+)
        /
        (?P<platform>{"|".join(map(str, known_subdirs))})
        (?P<path>/.*|$)
        """
    return re.compile(_platform_match_regex, flags=re.IGNORECASE 
```

tests/models/test_channel.py
```diff
@@ -1438,3 +1438,20 @@ def test_prioritize_channels():
         for weight, (url, channel) in enumerate(urls)
         for subdir in context.subdirs
     }
+
+
+def test_channel_url_has_platform_as_substr():
```
kenodegard, line 1443: This indirectly tests `split_platform` and only half of this tests `split_platform` (the non-example.com case) as the example.com case is parsed as a URL and is a different code path.

So we should add a unit test for `split_platform`.

kenodegard:
Closing, pending conversation re. https://github.com/conda/conda/issues/15657#issuecomment-3946339962

Issue #15652, opened 2026-01-30
CorruptedEnvironmentError on macOS when conda-meta contains ._* (AppleDouble) files


shenhaofang:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

When creating an env with conda create --prefix .conda python=3.12 on macOS (project on a non–HFS+ volume), conda finishes the transaction then raises CorruptedEnvironmentError because of conda-meta/._*.json files. Those are macOS AppleDouble files, not real conda metadata; conda is misclassifying them as corruption.

in my case, the output is:

> CorruptedEnvironmentError: The target environment has been corrupted. Corrupted environments most commonly
> occur when the conda process is force-terminated while in an unlink-link
> transaction.
>   environment location: /Volumes/Elements/work/.conda
>   corrupted file: /Volumes/Elements/work/.conda/conda-meta/._bzip2-1.0.8-h80987f9_6.json

### Conda Info

```shell
active environment : /Volumes/Elements/work/article_docx/.conda
    active env location : /Volumes/Elements/work/article_docx/.conda
            shell level : 2
       user config file : /Users/shenhaofang/.condarc
 populated config files : /Volumes/Elements/program/anaconda3/.condarc
          conda version : 25.1.1
    conda-build version : 24.9.0
         python version : 3.12.2.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=25.1.1=0
                          __osx=26.2=0
                          __unix=0=0
       base environment : /Volumes/Elements/program/anaconda3  (writable)
      conda av data dir : /Volumes/Elements/program/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/osx-arm64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /Volumes/Elements/program/anaconda3/pkgs
                          /Users/shenhaofang/.conda/pkgs
       envs directories : /Volumes/Elements/program/anaconda3/envs
                          /Users/shenhaofang/.conda/envs
               platform : osx-arm64
[...]

PR #15664, opened 2026-02-02
feat: filter out JSON file paths that begin with "._" to avoid errors…

shenhaofang:
… caused by files automatically generated by external hard drives on the macOS system affecting metadata loading.

resolve the issue https://github.com/conda/conda/issues/15652


conda/core/prefix_data.py
```diff
@@ -466,7 +466,7 @@ def load(self) -> None:
             conda_meta_json_paths = (
                 p
                 for p in (entry.path for entry in os.scandir(_conda_meta_dir))
-                if p[-5:] == ".json"
+                if p[-5:] == ".json" and p[:2] != "._"
```
soapy1: This seems like a reasonable approach to me. I had to do a bit of reading up on these types of files. But it looks like they provide extra metadata for macOS systems and are safe to be ignored.

Also to note that an alternative implementation is to check for the existence of a magic number in the header of the file. This will provide more certainty that the file is in fact an apple double file.

> This field, modeled after the UNIX magic number feature, specifies the file’s
format. Apple has defined the magic number for the AppleSingle format as
$00051600 or 0x00051600.

Something like:

```python
. . .
   with open(filepath, 'rb') as f:
        # read the first 4 bytes
        magic_bytes = f.read(4)

        # return early if the header has the apple double magic number
        if magic_number == 0x00051607:
            return 
. . .
```

This approach will double the amount of  times files get opened in the worse case. So, in this context I think it's appropriate to just check the filename prefix.

ref: 
https://www.loc.gov/preservation/digital/formats/fdd/fdd000625.shtml
https://nulib.com/library/AppleSingle_AppleDouble.pdf

soapy1 (changes requested):
This looks like a reasonable change to me. Thanks for taking the time to submit this bug fix!
It would be nice to have a test or so to go along with this change so that we don't get a regression in the future. Would you be open to adding a test for this?

jezdez:
@conda-bot check

shenhaofang:
> This looks like a reasonable change to me. Thanks for taking the time to submit this bug fix! It would be nice to have a test or so to go along with this change so that we don't get a regression in the future. Would you be open to adding a test for this?

Thank you for reviewing my PR. I have added tests for it.

shenhaofang:
@conda-bot check

jezdez:
@shenhaofang the check for @conda-bot is just there for the CLA check (I know, it's a little weirdly named). Reviews are done through @conda/conda-maintainers

jezdez (approved):

PR #15672, opened 2026-02-03
Drop redundant shell hook subprocess for `run` subcommand

### LINKED ISSUE #14993: `conda run` could be faster

agriyakhetarpal:
### Description

With this PR, the wrapper script no longer computes or evaluates the shell hook subprocess's output and is deemed redundant. Instead, we rely on the inline activators to set up activation environment variables directly, which are a superset of those the shell hook sets (i.e., the `CONDA_EXE`, `_CONDA_EXE`, `CONDA_PYTHON_EXE`, `_CONDA_ROOT`, `_CE_M`, `_CE_CONDA` variables plus `PATH`, `CONDA_SHLVL`, and `CONDA_PROMPT_MODIFIER`).

This grants us a ~2.89x speedup, i.e., a ~65.39% reduction in mean execution time. I tested with `hyperfine --warmup 10 --runs 100 "conda run -- python --version"`:

#### Before

```
Time (mean ± σ):      1.399 s ±  0.296 s    [User: 0.892 s, System: 0.364 s]
Range (min … max):    1.118 s …  2.621 s    100 runs
```

#### After

```
Time (mean ± σ):     484.1 ms ± 146.2 ms    [User: 308.8 ms, System: 125.6 ms]
Range (min … max):   350.6 ms … 1007.5 ms    100 runs
```

Closes #14993. We discussed the prospect of `conda run` not requiring the plugin system and agreed to converse about it in a new, separate issue: https://github.com/conda/conda/pull/15534#issuecomment-3800305523.

cc: @jaimergp

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_remove.py
+++ b/conda/cli/main_remove.py
@@ -227,6 +227,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             specs = specs_from_args(args.package_names)
         channel_urls = ()
         subdirs = ()
+
+        prefix_data = PrefixData(prefix)
+        if unmatched_specs := [
+            str(spec) for spec in specs if not tuple(prefix_data.query(spec))
+        ]:
+            raise PackagesNotFoundError(tuple(sorted(unmatched_specs)))
+
         solver_backend = context.plugin_manager.get_cached_solver_backend()
         solver = solver_backend(prefix, channel_urls, subdirs, specs_to_remove=specs)
         txn = solver.solve_for_transaction()
--- a/news/15697-faster-conda-remove-for-classic-solver
+++ b/news/15697-faster-conda-remove-for-classic-solver
@@ -1,0 +1,19 @@
+### Enhancements
+
+* `conda remove <spec>` with the classic solver now exits early if the spec to be removed does not exist in the environment. (#15697)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/cli/test_main_remove.py
+++ b/tests/cli/test_main_remove.py
@@ -25,6 +25,8 @@ from conda.testing.integration import (
 )
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from pytest import MonkeyPatch
 
     from conda.testing.fixtures import CondaCLIFixture, TmpEnvFixture
@@ -134,3 +136,20 @@ def test_remove_all_default_activation_env(
                 "--all",
                 f"--prefix={prefix}",
             )
+
+
+def test_remove_early_existence_check(
+    empty_env: Path, conda_cli: CondaCLIFixture, mocker
+):
+    """Verify that the (classic) solver is not invoked when packages don't exist in prefix."""
+    mock_solver_backend = mocker.patch(
+        "conda.base.context.context.plugin_manager.get_cached_solver_backend"
+    )
+
+    with pytest.raises(
+        PackagesNotFoundError,
+        match=r"(?s)missing from the target environment:.+nonexistent-package",
+    ):
+        conda_cli("remove", f"--prefix={empty_env}", "nonexistent-package", "--yes")
+
+    mock_solver_backend.assert_not_called()


agriyakhetarpal:
Re-running CI to get a better estimate for CodSpeed, since this PR is meant to be a performance improvement, not a regression. This PR remains ready for review, otherwise.

jezdez:
Hold on this changes fundamentally how the plugins are loaded this isn't just a bug fix.

agriyakhetarpal:
Yeah, it's more of a performance improvement rather than a bug fix

agriyakhetarpal:
I experimented with another step towards lazily generating parsers in this commit: https://github.com/conda/conda/pull/15672/changes/b02e916ee50e13d3646b8f906f2617c9339c4e4a, but reverted it in a later one since it breaks a few tests. Nevertheless, I do think I should clarify here that the majority of the speedup achieved in this PR has come from dropping the shell hook subprocess, not by wrangling the imports, which have a much lesser effect. 

Happy to hear any concerns, of course. Thanks!

agriyakhetarpal:
> I should clarify here that the majority of the speedup achieved in this PR has come from dropping the shell hook subprocess, not by wrangling the imports, which have a much lesser effect.

To elaborate: if we have the current change (i.e., both changing the import order + removing the shell hook subprocess):

```
Time (mean ± σ):     502.7 ms ± 141.8 ms    [User: 314.7 ms, System: 130.9 ms]
Range (min … max):   354.5 ms … 916.9 ms    100 runs
```

If I drop aee39a9a0eb0c23de0e43a7c84df797045ed43cd locally, i.e., if we keep just the shell hook subprocess removal change in this PR, I get this:

```
Time (mean ± σ):     512.5 ms ± 171.9 ms    [User: 321.3 ms, System: 132.8 ms]
Range (min … max):   360.3 ms … 1297.3 ms    100 runs
```

I tested with `hyperfine --warmup 10 --runs 100 "conda run -- python --version"`. Since there is not much difference in the results from changing the import order beyond purported theoretical improvements, I will revert the change for now. I agree that, since built-in plugins are core functionality as pointed out to me internally, this change is best discussed elsewhere.

conda/utils.py
```diff
@@ -311,35 +311,21 @@ def wrap_subprocess_call(
...
         with Utf8NamedTemporaryFile(mode="w", delete=False) as fh:
             # If any of these calls to the activation hook scripts fail, we want
             # to exit the wrapper immediately and abort `conda run` right away.
             fh.write("set -e\n")
             if dev_mode:
                 from . import CONDA_SOURCE_ROOT
 
-                fh.write(">&2 export PYTHONPATH=" + CONDA_SOURCE_ROOT + "\n")
-            hook_quoted = quote_for_shell(*conda_exe, "shell.posix", "hook", *dev_args)
+                fh.write(f"export PYTHONPATH={CONDA_SOURCE_ROOT}\n")
```
jaimergp:
```suggestion
                fh.write(f">&2 export PYTHONPATH={CONDA_SOURCE_ROOT}\n")
```

Accidental or intentional removal?

jaimergp:
> With this PR, the wrapper script no longer computes or evaluates the shell hook subprocess's output and is deemed redundant. 

Huh, so basically we missed this bit in the previous optimization work, right? Not sure why Codspeed is not picking up the improvements, though 🤔

agriyakhetarpal:
> > With this PR, the wrapper script no longer computes or evaluates the shell hook subprocess's output and is deemed redundant. 
> 
> Huh, so basically we missed this bit in the previous optimization work, right? Not sure why Codspeed is not picking up the improvements, though 🤔 

Yes, exactly!

I have no clue about CodSpeed, either, but it previously shower regressions, so I can investigate briefly if you'd suggest me to.

jaimergp:
FWIW, my benchmarks confirm the results. Pretty impressive!

```
$ hyperfine --warmup 2 -L branch main,conda-run-misc-speedups --prepare 'git checkout {branch} &> /dev/null' --command-name '{branch}' --export-markdown /var/folders/1b/z374sygs1s96t_xxxv0hr3lr0000gn/T/tmp.hvWVJ09RVM 'python -m conda run -- python -V'

Benchmark 1: main (branch = main)
  Time (mean ± σ):     908.6 ms ± 131.6 ms    [User: 719.3 ms, System: 142.6 ms]
  Range (min … max):   824.6 ms … 1257.3 ms    10 runs
 
Benchmark 2: conda-run-misc-speedups (branch = conda-run-misc-speedups)
  Time (mean ± σ):     327.7 ms ±  59.8 ms    [User: 235.9 ms, System: 53.2 ms]
  Range (min … max):   264.7 ms … 437.9 ms    10 runs
 
Summary
  conda-run-misc-speedups (branch = conda-run-misc-speedups) ran
    2.77 ± 0.65 times faster than main (branch = main)
```

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `main (branch = main)` | 908.6 ± 131.6 | 824.6 | 1257.3 | 2.77 ± 0.65 |
| `conda-run-misc-speedups (branch = conda-run-misc-speedups)` | 327.7 ± 59.8 | 264.7 | 437.9 | 1.00 |

conda/utils.py
```diff
@@ -311,35 +311,21 @@ def wrap_subprocess_call(
...
             if debug_wrapper_scripts:
                 fh.write(">&2 echo '*** environment before ***'\n>&2 env\n")
-                fh.write(f'>&2 echo "$({hook_quoted})"\n')
-            fh.write(f'eval "$({hook_quoted})"\n')
```
jezdez, line 339: Huh, we're completely removing the running of `shell.posix`? I don't understand how that is possible.

jezdez:
Please make sure to wait for Ken and me to finish our reviews

conda/utils.py
agriyakhetarpal: Yes! The inline activator sets the variables it was setting, plus a few extra ones that I listed in the description. 

Edit: oh, I assume you're asking why it was removed, and it's because we don't need any interactive features for `conda run`, so shell functions such as `__conda_activate()` or `__conda_hashr()` wouldn't be useful. We call the activator class to generate the activation commands and run the user command, and it saves one subprocess.

jaimergp: This can be considered a follow up to https://github.com/conda/conda/pull/15534, where the activation logic was inlined in the `conda run` script.

jaimergp: In 25.11.0, `conda run python -V` would generate this script:

```
eval "$(/Users/jrodriguez/.local/conda/bin/conda shell.posix hook)"
conda activate  /Users/jrodriguez/.local/conda/envs/conda-dev
python -V
```

With https://github.com/conda/conda/pull/15534 (current `main`), the `conda activate` call was inlined to avoid a subprocess call:

```bash
set -e
eval "$(/Users/jrodriguez/.local/conda/bin/conda shell.posix hook)"
export _CE_M=''
export _CE_CONDA=''
PS1='(conda-dev) '
export PATH='/Users/jrodriguez/.local/conda/envs/conda-dev/bin:/Users/jrodriguez/.local/conda/condabin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/opt/X11/bin:/Users/jrodriguez/noconda/bin:/Users/jrodriguez/.pixi/bin:/Users/jrodriguez/.local/conda/condabin:/Users/jrodriguez/.local/bin:/Users/jrodriguez/.lmstudio/bin:/Users/jrodriguez/.vscode/extensions/ms-python.debugpy-2025.18.0-darwin-arm64/bundled/scripts/noConfigScripts:/Users/jrodriguez/.pixi/bin:/Users/jrodriguez/.local/conda/condabin:/Users/jrodriguez/.local/bin'
export CONDA_SHLVL='1'
export CONDA_PROMPT_MODIFIER='(conda-dev) '
export CONDA_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/conda'
export _CONDA_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/conda'
export CONDA_PYTHON_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/python'
export _CONDA_ROOT='/Users/jrodriguez/.local/conda/envs/conda-dev'
. "/Users/jrodriguez/.local/conda/envs/conda-dev/etc/conda/activate.d/libxml2-split_activate.sh"
set +e
python -V
_CONDA_EXE_RC=$?
exit $_CONDA_EXE_RC
```

This PR removes the leftover `shell.posix hook` call, which was only needed to enable the original `conda activate` call, thus saving one more process:

```bash
set -e
export _CE_M=''
export _CE_CONDA=''
PS1='(conda-dev) '
export PATH='/Users/jrodriguez/.local/conda/envs/conda-dev/bin:/Users/jrodriguez/.local/conda/condabin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/opt/X11/bin:/Users/jrodriguez/noconda/bin:/Users/jrodriguez/.pixi/bin:/Users/jrodriguez/.local/conda/condabin:/Users/jrodriguez/.local/bin:/Users/jrodriguez/.lmstudio/bin:/Users/jrodriguez/.vscode/extensions/ms-python.debugpy-2025.18.0-darwin-arm64/bundled/scripts/noConfigScripts:/Users/jrodriguez/.pixi/bin:/Users/jrodriguez/.local/conda/condabin:/Users/jrodriguez/.local/bin'
export CONDA_SHLVL='1'
export CONDA_PROMPT_MODIFIER='(conda-dev) '
export CONDA_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/conda'
export _CONDA_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/conda'
export CONDA_PYTHON_EXE='/Users/jrodriguez/.local/conda/envs/conda-dev/bin/python'
export _CONDA_ROOT='/Users/jrodriguez/.local/conda/envs/conda-dev'
. "/Users/jrodriguez/.local/conda/envs/conda-dev/etc/conda/activate.d/libxml2-split_activate.sh"
set +e
python -V
_CONDA_EXE_RC=$?
exit $_CONDA_EXE_RC
```

agriyakhetarpal: Yes, correct ✅ In #15534, I was going even further by getting the inline activators to avoid changing the shell prompt (we don't need to bother with it for a non-interactive shell) and save some extra time, but that experiment didn't work out. I haven't made that change here to keep this PR narrow in scope.

jaimergp: You could argue that by removing `shell.posix hook` call, the `conda` shell function is no longer defined in the `conda run` script, so if the user depended on that function being present (e.g. manually calling `conda run -- conda activate my-env`), then that'll fail.

I doubt that was a widespread use case, but I did check and somehow it's a [pattern in Dockerfiles](https://github.com/search?q=%2Fconda+run.*%5Cs%2Bconda%5Cs%2B%2F++NOT+is%3Aarchived+NOT+is%3Afork&ref=opensearch&type=code). For example:

```Dockerfile
RUN conda run -n NeuralPLexer conda env update --file NeuralPLexer.yml && conda run -n NeuralPLexer conda clean -afy
``` 

So I think the main point we need to enable this optimization fully is to make sure that the base `conda` is in PATH. In other words, ensure that `$ROOT_PREFIX/condabin` is in PATH, @agriyakhetarpal.

---

Thanks for opening this question @jezdez!

agriyakhetarpal:
Thanks for your investigation, @jaimergp 🚀  I had no clue about these (uncommon, and a bit peculiar, if I can say so) user patterns. I'll mark this as a draft again until I figure out a way to implement https://github.com/conda/conda/pull/15672#discussion_r2788523040

agriyakhetarpal:
(gotta figure out a test for this)

conda/utils.py
jaimergp: Reminder here @agriyakhetarpal

conda/utils.py
```diff
@@ -352,6 +343,11 @@ def wrap_subprocess_call(
 
             fh.write(activate_code)
 
+            # Ensure that nested `conda` commands will work
+            # without the shell hook function.
+            condabin_dir = join(root_prefix, "condabin")
+            fh.write(f'export PATH="{condabin_dir}:$PATH"\n')
```
jaimergp: Not sure if we should add `condabin` first or last 🤔 Ideally we add it _before_ activating the target environment, so the target environment's `conda` takes precedence (I'm trying to follow the principle of least surprise here, not sure if you agree).

agriyakhetarpal: Yes, good point, we don't want the base conda to shadow a conda in the target environment.

agriyakhetarpal: See 97629cf6968061ff376827a06146b88a7717b080

jaimergp: Thanks, that's reasonable. I would still write it to the script _before_ we inline the activation code, so it behaves even more like the shell.hook logic. i.e., PATH should have:

- PATH entries added by activation
- condabin entry
- whatever was on PATH before that

The script should then do:

```bash
# preamble
set PATH="<condabin>:$PATH"
# inlined activation code
# user command
```

agriyakhetarpal: Ah, sorry, looks like I misunderstood you in that case. 😅 Let me have a look again!

agriyakhetarpal: The activator writes an absolute PATH string. So we'll need to somehow inject `condabin` into the process's PATH before the activator runs its `_get_starting_path_list()`. This sounds a bit messy.

agriyakhetarpal: Could you check 8c357c09b3a6a974c17059e3db7e5539adb63465, @jaimergp? I think it should achieve what we desire, but if you have a better way, LMK!

jaimergp: Yep, that works, thanks!

conda/utils.py
agriyakhetarpal: Sorry, I missed this comment. This is an intentional removal since it's a no-op. `export` is silent anyway, and there is no benefit in redirecting it to stderr. I assume that this came from someone's old debugging habit? 😄

kenodegard: looks like a consistency choice 🤷🏼‍♂️

tests/cli/test_main_run.py
```diff
@@ -504,6 +504,36 @@ def test_run_preserves_exit_code_despite_deactivation_failure(
         assert deactivation_marker_file.exists()
 
 
+def test_run_wrapper_has_condabin_on_path(
+    request: pytest.FixtureRequest,
+    tmp_env: TmpEnvFixture,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """Regression test to ensure condabin is on PATH in the wrapper script.
+    See https://github.com/conda/conda/pull/15672 for context.
```
kenodegard: This isn't a regression?

kenodegard:
```suggestion
    """Test to ensure condabin is on PATH in the wrapper script.
    See https://github.com/conda/conda/pull/15672 for context.
```

tests/cli/test_main_run.py
```diff
@@ -504,6 +504,36 @@ def test_run_preserves_exit_code_despite_deactivation_failure(
...
+
+    Without the shell hook, the conda shell functions are not defined,
+    so nested conda commands rely on the conda executable in ``condabin``
+    being discoverable via PATH.
```
kenodegard, line 517: Very small corner case but this also means `conda [de]activate` are no longer possible within a `conda run` command.

Currently the following passes (not that we can do much with it):

```
$ conda run -- conda activate scratch
$ echo $?
0
```

With this PR:

```
$ conda run -- conda activate scratch

CondaError: Run 'conda init' before 'conda activate'

ERROR conda.cli.main_run:execute(148): `conda run conda activate scratch` failed. (See above for error)
$ echo $?
1
```

tests/cli/test_main_run.py
agriyakhetarpal: Thanks, I'll take it. I meant "regression" in that it's a test for the future, though!

tests/cli/test_main_run.py
agriyakhetarpal: Yes, correct. I hope no one is `conda run` and `conda [de]activate` like this, though. `conda run -- conda activate something` “working” previously would be most likely accidental and not especially meaningful in any robust workflow, as activation would only affect the wrapper’s shell process, not the user’s parent shell, i.e., the user be essentially using an interactive shell mechanism inside a one-shot non-interactive one.

jaimergp: There are workarounds too; they can always eval the hooks directly if they reaaaally need this. I hope no one does. I agree on the accidental component of this working previously.

ryanskeith: Is there a use case for activation like this?  Should we find a way to disable this?

kenodegard: there isn't a good use case, I was just thinking about what this change technically breaks, my only qualm is the error isn't super clear but that's always been an issue with this particular error message

agriyakhetarpal:
Thanks for the approval, @kenodegard! The GitHub project automation moved it to "Approved" as a result – could you please move it back to "Blocked", since we do need to wait for @jezdez's review? Just to make sure this doesn't inadvertently get merged.

tests/cli/test_main_run.py
jezdez: Good riddance, I say

beckermr:
@jaimergp @agriyakhetarpal This PR probably needs to be backed out as well. We should roll back to this function as it was in conda 25.11.x and do this all at once with better testing.

conda/utils.py
beckermr, line 1: @agriyakhetarpal @jaimergp @jezdez This PR combined with the previous one has structural issues IMHO. Before both of those PRs, the various env vars in the hooks that were set, plus the activate process itself was encapsulated in a single spot. This was for good reason since anytime the activation process is changed, it should be changed everywhere. Now with this PR, details specific to activation appear in this utility function. That means if we changed the activation process, we now would need coordinated changes across the code base (i.e., we've lost encapsulation). We need to figure out how to inline things without loosing the encapsulation that was present before.

agriyakhetarpal: Thanks, @beckermr. I'll sit through this a bit more carefully after the 26.3.2 release proceeds. The faulty code is still in `main` (it hasn't been reverted there as noted in #15896). I'll wait until you have results from testing my proposed fix in #15534, and I'll start refactoring the activation code soon after. Thank you for being involved in this manner, and I hope it's okay if I ping you on further work in this direction going forward!

beckermr: Yep. I posted on the other PR the results of the test and it worked. However, the structural issues remain. Do ping me the resulting PR reintroducing these changes and I am more than happy to help review and think through them! 

I really appreciate all of your hard work on conda and your responsiveness on these critical bugs! ❤️

PR #15675, opened 2026-02-03
Build Python 3.14 canaries & add to test matrix

kenodegard:
### Description

With conda 26.1.0 released to main and conda-forge with Python 3.14 supported we can add 3.14 back to the canary builds and to the test matrix.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #14401, opened 2024-11-18
Add MSVC-built `win-arm64` launcher

jaimergp:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Taken directly from [`win-arm64/conda-launchers-24.7.1-h2722cbf_1.tar.bz2`](https://anaconda.org/conda-canary/conda-launchers/24.7.1/download/win-arm64/conda-launchers-24.7.1-h2722cbf_1.tar.bz2) at [`conda-canary/conda-launchers`](https://anaconda.org/conda-canary/conda-launchers/files?version=24.7.1), which were built in [this CI run](https://github.com/conda/conda-launchers/actions/runs/11898535180/job/33155248271) for [this commit](https://github.com/conda/conda-launchers/commit/d23e5578fb780bff2255c9180a7e5346b200a6d6).

Needs to be signed with Anaconda certificates.

PR #15678, opened 2026-02-04
Use bundled stub executables for Windows entry points instead of `Scripts/conda.exe`

kenodegard:
### Description

While adding Python 3.14 support to the canary builds (https://github.com/conda/conda/pull/15675), we found that Windows tests were failing with:

```
The path 'Scripts/conda.exe' specified in the package manifest cannot be found.
```

Highly misleading failures:
- https://github.com/conda/conda/actions/runs/21650181334/job/62437486991?pr=15675
    Screenshot 2026-02-04 at 08 57 15
- https://github.com/conda/conda/actions/runs/21676076697/job/62497374852?pr=15678
    Screenshot 2026-02-04 at 09 38 41

Caused by conda-forge's `conda-libmamba-solver` removing its explicit `conda` dependency to resolve circular build issues (see https://github.com/conda-forge/conda-libmamba-solver-feedstock/pull/41). This meant `conda` was no longer implicitly installed when setting up the test environment, so `Scripts/conda.exe` didn't exist.

The fix proposed here is to use the bundled `cli-32.exe`/`cli-64.exe` stub executables from `conda/shell/` to create Windows entry points, instead of relying on `$CONDA_PREFIX/Scripts/conda.exe` which is a `conda init` artifact.

When installing noarch packages with entry points on Windows, conda copies `Scripts/conda.exe` from `context.conda_prefix` to create executables like `pip.exe`, `wheel.exe`, etc. This required either:
1. conda to be installed as a conda package (which installs `Scripts/conda.exe`), or
2. running `conda init` (which creates `Scripts/conda.exe`)

This caused issues when running tests via `python -m pytest` without installing conda as a package, since `Scripts/conda.exe` wouldn't necessarily exist.

### Checklist - did you ...

### DIFF
--- a/news/15678-windows-entry-point-bundled-stubs
+++ b/news/15678-windows-entry-point-bundled-stubs
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Use bundled Windows stub executables (`conda/shell/cli-32.exe`, `conda/shell/cli-64.exe`) for creating entry points instead of relying on `Scripts/conda.exe` so that entry point creation works when conda is not installed as a conda package (e.g. tests via `python -m pytest`). Select the correct exe for the target env via a platform-to-stub mapping; for win-arm64 use `cli-64.exe` (relying on Windows x64-on-ARM64 emulation). Use target platform for `conda init` as well, fixing cross-architecture installs (e.g. win-32 host creating a win-64 env). (#15678)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>


conda/core/path_actions.py
```diff
@@ -370,8 +371,12 @@ def create_python_entry_point_windows_exe_action(
         requested_link_type,
         entry_point_def,
     ):
-        source_directory = context.conda_prefix
-        source_short_path = "Scripts/conda.exe"
+        # Use the bundled stub executables from conda/shell/ instead of
+        # relying on Scripts/conda.exe existing in context.conda_prefix.
+        # This allows conda to work when installed via pip or during testing.
```
jezdez: Well, that's a nice surprise, BRB preparing the PyPI release.. oh wait

kenodegard:
so the activation script tests still need `Scripts\conda.exe`

so probably best to add back the conda dependency

jaimergp:
Just mentioning that the current behavior on `main` is weird in its own way:

- `Scripts/conda.exe` actually comes from conda-build's copy, because we run `conda init` as part of the packaging steps, IIRC. This is the one that is used for all `entry_points` in a recipe for non-`noarch` packages.
- The bundled copy is obviously `conda`'s, and used for `noarch: python` packages at install time.

Right now these two copies should be the same because we sync'ed them in https://github.com/conda/conda/pull/11676, but just a word of caution in case they get out of sync again in the future.

conda/core/path_actions.py
```diff
@@ -370,8 +371,12 @@ def create_python_entry_point_windows_exe_action(
...
+        bits = 8 * struct.calcsize("P")
```
jaimergp: I'd swear I've seen this somewhere else in the code base. Also this will break for `win-arm64`.

tests/requirements-ci.txt
```diff
@@ -1,3 +1,4 @@
+conda  # for Windows shell ctivation tests
```
jaimergp:
```suggestion
conda  # for Windows shell activation tests
```

conda/core/path_actions.py
```diff
@@ -370,8 +371,12 @@ def create_python_entry_point_windows_exe_action(
         requested_link_type,
         entry_point_def,
     ):
-        source_directory = context.conda_prefix
-        source_short_path = "Scripts/conda.exe"
```
jaimergp, line 374: Interesting how nobody complained about this because we were always using the same platform, but this would be broken for cross-architecture environments (e.g. creating a `win-64` env on a `win-32` installation). In practice, not a problem, but it could have lead to wonky issues on `win-arm64` platforms where emulation were somehow not available.

conda/core/path_actions.py
jaimergp: I guess we should map the target platform back to the filename:

* `win-32` -> `cli-32.exe`
* `win-64` -> `cli-64.exe`
* `win-arm64` -> `cli-arm64.exe`

So, in other words, `target_platform.split("-")[-1]`

kenodegard: I suppose this means we should add cli-arm64.exe to conda too

kenodegard: We just need to establish (and document) the source of truth for these

jaimergp:
Replying to [this comment](https://github.com/conda/conda/pull/15678#discussion_r2770311521):

> I suppose this means we should add cli-arm64.exe to conda too

@kenodegard, what if we copy `cli-64.exe` as `cli-arm64.exe` here to let the emulation take over while we figure out the [native launcher](https://github.com/conda/conda/pull/14401) provenance? A symlink might work too as long as we resolve it, but I guess it's ok to have a full copy.

We will probably need something similar in `conda-build` too.

jaimergp:
@baszalmstra, emulation works out of the box in win-arm64, right? Do you think the strategy above makes sense?

Edit: Bas confirmed offline that it does work like that, so yep, we can copy.

conda/base/constants.py
```diff
@@ -107,6 +107,14 @@
 KNOWN_SUBDIRS: Final = ("noarch", *PLATFORMS)
 PLATFORM_DIRECTORIES = KNOWN_SUBDIRS
 
+# Windows subdir -> path under CONDA_PACKAGE_ROOT to the entry point stub exe.
+# Future source of truth: https://github.com/conda/conda-launchers
+WINDOWS_LAUNCHER_STUB_PATH: Final = {
+    "win-32": "shell/cli-32.exe",
+    "win-64": "shell/cli-64.exe",
+    "win-arm64": "shell/cli-arm64.exe",
```
kenodegard: Since we use an explicit mapping we could skip copying cli-64.exe to cli-arm64.exe?

jaimergp: Oh true, for now we can do that.

isuruf:
@kenodegard I independently did some work on this at https://github.com/conda/conda/pull/15690

conda/base/constants.py
kenodegard: See https://github.com/conda/conda/pull/15678/commits/0433acee20a87ddec828c7f703018d77fd428543

isuruf (approved):

jaimergp (approved):
Let's add a news/ file and merge. Thanks!

Issue #15233, opened 2025-08-11
Remove JLAP Support


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Clean up the codebase by removing existing JLAP support and refactoring the current network caching code that has become messy over time.

### Why is this needed?

The introduction of shard-based functionality makes the existing JLAP support redundant.

### What should happen?

Remove all JLAP-related code and refactor the network caching systems to be cleaner and more maintainable. This should simplify the codebase and make room for the new shard-based caching mechanisms.

### Additional Context

Ensure that removing JLAP support doesn't break any existing functionality that users depend on. The refactoring should improve code clarity and maintainability while preserving performance characteristics.

dholth:
We removed the "and refactor" from this issue to simplify. The refactor task should become a different issue.

It's simple to remove support for JLAP. It's skipped by default but the "alternate repodata formats" support also lets us find `repodata.json.zst`. This all lives inside the repodata gateway, and refactoring the whole repodata gateway to a certain standard is a larger task.

The repodata gateway is convoluted because it supports the old "fetch text" API and newer APIs that need to manipulate the cache at a lower level.

A specific tricky bit is to deal with locking the cache's .info.json whenever cache data is read or written so that this operation is atomic.

danyeaw:
#15687 opened. @dholth, please let me know if that's what you had in mind.

PR #15685, opened 2026-02-04
Removal of JLAP support

ryanskeith:
### Description

Addresses #15233, JLAP support.  A good experiment but now sharding has been adopted as the way to go.

- Removal of JLAP support 1st pass

### Checklist - did you ...


conda/cli/helpers.py
```diff
@@ -301,9 +301,8 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
     channel_customization_options.add_argument(
         "--experimental",
         action="append",
-        choices=["jlap", "lock"],
-        help="jlap: Download incremental package index data from repodata.jlap; implies 'lock'. "
-        "lock: use locking when reading, updating index (repodata.json) cache. Now enabled.",
+        choices=["lock"],
```
dholth, line 304: I designed this intending that old experiments should not break your scripts. Since lock is now the default and jlap is disabled, is there a way to hide the help text or not fail validation if you pass --experimental=jlap,lock

conda/gateways/repodata/__init__.py
```diff
@@ -58,7 +58,7 @@
 stderrlog = logging.getLogger("conda.stderrlog")
 
 
-# if repodata_shards.msgpack.zst, repodata.json.zst or repodata.jlap were unavailable, check again later.
+# if repodata_shards.msgpack.zst or repodata.json.zst were unavailable, check again later.
```
dholth, line 61:
```suggestion
# if an alternate repodata format was unavailable, check again later.
```

conda/gateways/repodata/__init__.py
```diff
@@ -101,25 +101,11 @@ class Response304ContentUnchanged(Exception):
...
-            warnings.warn(
-                "Could not load the configured jlap repo interface. "
-                f"Is the required jsonpatch package installed?  {e}"
-            )
-
-    if context.repodata_use_zst:
-        try:
-            from .jlap.interface import ZstdRepoInterface
-
-            return ZstdRepoInterface
```
dholth, line 119: `ZstdRepoInterface` is necessary. It is the only way to fetch `repodata.json.zst` which is radically faster than `repodata.json` because we don't have to wait for streaming `Content-Encoding: gzip` from the server.

Is it possible to remove `CondaRepoInterface`?

Do we have a test that asserts that `repodata.json.zst` is discovered and used?

conda/cli/helpers.py
ryanskeith: Good thought, I can look into that.

conda/gateways/repodata/__init__.py
ryanskeith: This was me being overzelous and removing the entire folder without reading deeply its two function nature.  Great question on the repodata.json.zst.  I will check.

ryanskeith: I am going to rework how I did this for better github tracking.

PR #15549, opened 2025-12-27
Option to disable changing fish_prompt function in conda init #15545

lsoksane:
### Description

Implements [issue #15545](https://github.com/conda/conda/issues/15545), that is, `conda shell.fish hook` does not change `fish_prompt` function if environment variable `CONDA_DISABLE_FISH_PROMPT` is set. Otherwise the behaviour is unchanged.

Only one if statement in the fish code is added, together with the corresponding end statement, and the indentation of the lines in between has been changed.  

The new  option should still be documented. 

Attached [recording.txt](https://github.com/user-attachments/files/24356612/recording.txt) shows a shell session demonstrating the new feature. 

Closes #15545

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #15686, opened 2026-02-04
Restore fish shell tests

danyeaw:
### Description

I noticed when reviewing #15549, that the fish shell tests were marked as `xfail` 👀.

I updated the test `InteractiveShell` context manager so that it uses a PTY interactive shell for testing it instead of pipes for fish. The tests were failing before because fish detects that it is running in non-tty and then disables the prompt rendering. Using `pexpect.spawn` ensures that the shell thinks it is in a real terminal.

Since launching an interactive shell is slow, I also tried to minimize the integration test to check minimal functionality.

### Checklist - did you ...

### DIFF
--- a/tests/shell/__init__.py
+++ b/tests/shell/__init__.py
@@ -191,27 +191,6 @@ class InteractiveShell(metaclass=InteractiveShellType):
         self.env = env or {}
 
     def __enter__(self):
-        env = {
-            **os.environ,
-            "CONDA_AUTO_ACTIVATE": "false",
-            "CONDA_AUTO_STACK": "0",
-            "CONDA_CHANGEPS1": "true",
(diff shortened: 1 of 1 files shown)


tests/shell/__init__.py
```diff
@@ -190,33 +191,40 @@ def __init__(
         self.env = env or {}
 
     def __enter__(self):
-        self.p = PopenSpawn(
+        env = {
```
kenodegard: I think this is simply moving the env definition, not changing any of the values, correct?

danyeaw: Hi Ken, that's correct and good point. I don't think that change to move the `env = {}` definition is needed, so I reverted it.

jezdez (approved):

PR #15693, opened 2026-02-06
Remove experimental JLAP support

### LINKED ISSUE #15233: Remove JLAP Support

ryanskeith:
### Description

Fixes #15233 for removing jlap support, ZstdRepoInterface which was a child class is still preserved and used.

**Removed:**
- JLAP experimental flag from CLI (`--experimental jlap`)
- `jlap` module directory and all JLAP-specific code
- `RepodataStateSkipFormat` class (no longer needed)
- `jsonpatch` dependency
-  Hashing code that was no longer required

**Preserved/Refactored:**
- `ZstdRepoInterface` - moved from `jlap/interface.py` to `zstd.py`
- Zstandard-compressed repodata (`.json.zst`) download functionality
- Fallback to `.json` when `.json.zst` unavailable
- `download_and_hash` changed to `download_repodata`
- All existing zstd repodata caching and state management

**Key Files:**
- `conda/gateways/repodata/zstd.py` - New home for `ZstdRepoInterface` and helpers
- `conda/gateways/repodata/__init__.py` - Updated `get_repo_interface()` to use `ZstdRepoInterface` directly
- `tests/gateways/test_zstd.py` - Updated tests (renamed from `test_jlap.py`)
- `tests/fixtures_package_server.py` - Renamed from `fixtures_jlap.py` for clarity

### Testing
- All 9 tests in `test_zstd.py` passing
- Tests verify zstd download, fallback to json, caching, and error handling
- Test fixtures renamed to reflect generic package server purpose

### Checklist - did you ...

### DIFF
--- a/tests/gateways/test_repodata_gateway.py
+++ b/tests/gateways/test_repodata_gateway.py
@@ -359,8 +359,6 @@ def test_repodata_fetch_formats(
     base = f"http://{host}:{port}/test"
     channel_url = f"{base}/osx-64"
 
-    repo_cls = CondaRepoInterface
-
     # we always check for *and create* a writable cache dir before fetch
     cache_path_base = tmp_path / "fetch_formats" / "xyzzy"
     cache_path_base.parent.mkdir(exist_ok=True)
@@ -368,7 +366,7 @@ def test_repodata_fetch_formats(
     channel = Channel(channel_url)
 
     fetch = RepodataFetch(
-        cache_path_base, channel, REPODATA_FN, repo_interface_cls=repo_cls
+        cache_path_base, channel, REPODATA_FN, repo_interface_cls=CondaRepoInterface
     )
 
     assert isinstance(fetch.cache_path_state, Path)  # coverage


conda/cli/helpers.py
```diff
@@ -301,9 +301,8 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
     channel_customization_options.add_argument(
         "--experimental",
         action="append",
-        choices=["jlap", "lock"],
-        help="jlap: Download incremental package index data from repodata.jlap; implies 'lock'. "
-        "lock: use locking when reading, updating index (repodata.json) cache. Now enabled.",
+        choices=["lock"],
```
ryanskeith: We have a choice here.  We could leave in jlap and now it's ignored or we need to adjust how the channel_customization handles items that are no longer listed.  I took it out for the time being for completeness but it does feel lacking.

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+    return re.sub(r"(\.\w+)$", ext, url)
+
+
+@contextmanager
+def timeme(message):
+    """Context manager for timing operations."""
+    begin = time.monotonic()
+    yield
+    end = time.monotonic()
+    log.debug("%sTook %0.02fs", message, end - begin)
```
jaimergp, line 52: I'd swear we have some timing utilities already somewhere else.

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def write(self, b: bytes):
+        self.hasher.update(b)
+        return self.backing.write(b)
+
+    def close(self):
+        self.backing.close()
+
+
+def download_and_hash(
```
jaimergp: This also looks like an utility that could be useful elsewhere (e.g. `.gateways`?)

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+
+def _is_http_error_most_400_codes(e: HTTPError) -> bool:
+    """Determine if HTTP error is a 4xx error (except 416)."""
+    if e.response is None:
+        return False
+    status_code = e.response.status_code
+    return 400 <= status_code < 500 and status_code != 416
+
+
+class ZstSkip(Exception):
```
jaimergp, line 116: We have usually defined all exceptions in `conda.exceptions`.

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+    """Exception to skip zst format check."""
+
+    pass
```
jaimergp: No need if there's a docstring already.


```suggestion
```

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+    from . import RepodataCache
+
+log = logging.getLogger(__name__)
+
+DIGEST_SIZE = 32
+NOMINAL_HASH = "blake2_256_nominal"
+ON_DISK_HASH = "blake2_256"
+
+
+def hash():
```
jaimergp: `hash()` is a [builtin](https://docs.python.org/3/library/functions.html#hash). We should pick a different name.

tests/gateways/test_repodata_gateway.py
```diff
@@ -358,19 +355,11 @@ def test_repodata_fetch_formats(
...
-
     host, port = package_server.getsockname()
     base = f"http://{host}:{port}/test"
     channel_url = f"{base}/osx-64"
 
-    if use_jlap:
-        repo_cls = JlapRepoInterface
-    else:
-        repo_cls = CondaRepoInterface
+    repo_cls = CondaRepoInterface
```
jaimergp: May as well just replace `repo_cls` with `CondaRepoInterface` in the instances below.

conda/cli/helpers.py
jaimergp: If you want to be smart, `choices` could be set to a magic list-like container that only lists the actually accepted values in the help message, but has a hidden list that also silently accepts (and ignores) removed values, maybe via subclassing `__contains__` or similar?

jaimergp: e.g.:

```python
from typing import Any, Iterable
from collections import UserList

class DeprecatedChoices(UserList):
    def __init__(self, valid: Iterable[Any], deprecated: Iterable[Any]):
        self.data = valid
        self.deprecated = deprecated
    
    def __contains__(self, value):
        return value in self.data or value in self.deprecated


choices = DeprecatedChoices(["valid", "also-valid"], ["deprecated"])
print(choices)
print("Is 'valid' allowed?", "valid" in choices)
print("Is 'deprecated' allowed?", "deprecated" in choices)
print("Is 'nope' allowed?", "nope" in choices)
```

which returns

```bash
$ python ./deprecated_list.py
['valid', 'also-valid']
Is 'valid' allowed? True
Is 'deprecated' allowed? True
Is 'nope' allowed? False
```

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,337 @@
...
+    json_path = cache.cache_path_json
+
+    hasher = hash()
+    is_fallback = False
+    with timeme(f"Download complete {url} "):
+        # Just try downloading .json.zst
+        try:
+            response = download_and_hash(
+                hasher,
+                withext(url, ".json.zst"),
```
soapy1, line 147: There is maybe an opportunity here to refactor all the instances of the string `.json.zst` into calling a constant like `REPODATA_JSON_ZST_EXTENSION`.

tests/gateways/test_zstd.py
```diff
@@ -0,0 +1,292 @@
...
+from conda.gateways.repodata.zstd import (
+    HashWriter,
+    ZstdRepoInterface,
+    download_and_hash,
+    hash,
+)
+from conda.models.channel import Channel
+
+
+def test_server_available(package_server: socket):
```
soapy1, line 39: Looks like this one is just testing the fixture? Can probably be removed since it's exercised in the tests already?

tests/gateways/test_zstd.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def test_repodata_state(
+    package_server: socket,
+    monkeypatch: MonkeyPatch,
+):
+    """Test that cache metadata file works correctly.  This test is valid for both .zst and .json repodata formats."""
+    host, port = package_server.getsockname()
+    base = f"http://{host}:{port}/test"
+    channel_url = f"{base}/osx-64"
```
soapy1, line 111: Does this need to use the osx-64 channel in particular, or could it use the current platform?

```suggestion
    channel_url = f"{base}/{context.subdir}"
```

soapy1 (commented):
🙌

conda/gateways/repodata/zstd.py
dholth: No longer needed (jlap only)

conda/gateways/repodata/zstd.py
dholth: Probably also no longer needed; jlap only. Although the standard "download and check hash" feature could use an update.

conda/gateways/repodata/zstd.py
dholth: Maybe but this is only used for flow control here.

tests/gateways/test_repodata_gateway.py
dholth: CondaRepoInterface (the old one) can't do compressed repodata. It should not be used at all.

tests/gateways/test_zstd.py
dholth: You may find any failures are easier to understand here than in the middle of the repodata code.

tests/gateways/test_zstd.py
dholth: There are local test files for only certain platforms.

dholth (commented):
Thank you for doing this. My concern is that the policy requires all removed symbols to be retained for a deprecation cycle until we update conda policy, so deleting modules is not allowed under that policy.

conda/gateways/repodata/zstd.py
ryanskeith: In the download_and_hash function, the zst path did seem to use the hashing function and that is why it was left in. See [here](https://github.com/conda/conda/blob/c814617cdb44cb8fec5d323c504ad7421227c548/conda/gateways/repodata/jlap/fetch.py#L249-L254). 

I will look into removing it completely if it is no longer required.

conda/gateways/repodata/zstd.py
ryanskeith: I will look.

conda/gateways/repodata/zstd.py
ryanskeith: I see the note below about download_and_hash and will look at the standard "download and check hash" function.

tests/gateways/test_repodata_gateway.py
ryanskeith: @dholth is there any value in testing `CondaRepoInterface` at all?

conda/gateways/repodata/zstd.py
dholth: That's probably because we need the hash when we later patch the repodata.json.zst with jlap. I wouldn't use the standard function here at all either since this code is performance critical.

conda/gateways/repodata/zstd.py
ryanskeith: So I did find a few timing functions in io.py.  One for log formatting, one for timeouts and lastly, `time_recorder` that does capture timing but it requires `CONDA_INSTRUMENTATION_ENABLED` to be an environment variable and it writes the result to a file.

This could be moved to conda.common.io in the future.

conda/cli/helpers.py
ryanskeith: That is interesting way to think of things.  I would vote to add this in on the next experimental feature we choose to put in.  I am liking the simplicity of just taking it out.  We might break a few scripts but I think they will be very few.

dholth: I am able to set in condarc `experimental: ["jlap", "eggs"]` and see both values in `conda.base.context.reset_context(); conda.base.context.context.experimental`. `conda install --dry-run --experimental=eggs conda` raises validation errors.
Breaking no scripts is better.
Since both experimental are deprecated, we could hide the help and not worry about list tricks.

ryanskeith:
@jaimergp responding to your "Let me know if you'd rather deal with that separately, though."  I would like to do it separately.  I think there are some general moves of functionality (like the `timeme` function) but also some work to do on streamlining some of the functions as well (like the download code as @dholth mentioned).

I think removing is a good solid foundation and then cleaning up the code a good separate step.

conda/cli/helpers.py
ryanskeith: @dholth that's a good idea as well.  Any objects @jaimergp ?

conda/cli/helpers.py
```diff
@@ -298,12 +298,12 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
             "conda config --describe repodata_fns."
         ),
     )
+    # jlap and lock experimental features are deprecated but are left in so older scripts will still work.
     channel_customization_options.add_argument(
         "--experimental",
         action="append",
```
soapy1: These can probably be deprecated if they are no longer used.

```suggestion
        action=deprecated.action(
            "26.3",
            "26.3",
            _AppendAction,
            addendum="Use `--force` instead.",
        ),
```

conda/cli/helpers.py
```diff
@@ -298,12 +298,12 @@ def add_parser_channels(p: ArgumentParser) -> _ArgumentGroup:
...
         choices=["jlap", "lock"],
-        help="jlap: Download incremental package index data from repodata.jlap; implies 'lock'. "
-        "lock: use locking when reading, updating index (repodata.json) cache. Now enabled.",
+        help="jlap and lock no longer supported.  These options will have no effect.",
```
soapy1:
```suggestion
        help="Deprecated: jlap and lock no longer supported.  These options will have no effect.",
```

conda/gateways/repodata/__init__.py
```diff
@@ -101,24 +101,11 @@ class Response304ContentUnchanged(Exception):
...
-        except ImportError as e:  # pragma: no cover
-            warnings.warn(
-                "Could not load the configured jlap repo interface. "
-                f"Is the required jsonpatch package installed?  {e}"
-            )
-
     if context.repodata_use_zst:
-        try:
-            from .jlap.interface import ZstdRepoInterface
+        # We need lazy import to avoid circular imports.
```
dholth, line 105: Some of my zstd code [has tested functionality](https://github.com/conda/conda-package-handling/blob/main/tests/test_degraded.py) to work in a degraded mode even when `zstandard` is not installed. May have been doing the same with try/except here.

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,303 @@
...
+    yield
+    end = time.monotonic()
+    log.debug("%sTook %0.02fs", message, end - begin)
+
+
+def build_headers(json_path: pathlib.Path, state: RepodataState):
+    """Build caching headers for a path and state."""
+    headers = {}
+    if json_path.exists():
+        etag = state.get("_etag")
```
dholth, line 59: Some code exists for backwards compatibility with `RepodataState | dict` including old property names. It might be possible to use constants for the header names and forget the old dict key names.

conda/gateways/repodata/zstd.py
```diff
@@ -0,0 +1,303 @@
...
+                json_path,
+                session=session,
+                state=state,
+                dest_path=temp_path,
+            )
+
+    # Update state with response headers (common for both zstd and fallback)
+    state["_mod"] = response.headers.get("last-modified")
+    state["_etag"] = response.headers.get("etag")
+    state["_cache_control"] = response.headers.get("cache-control")
```
dholth, line 178: Updated constants https://github.com/conda/conda/blob/main/conda/gateways/repodata/__init__.py#L64-L68

news/15233-remove-jlap-support
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Remove JLAP (JSON Line-oriented Patch) experimental support. Zstd-compressed repodata (`.json.zst`) functionality is preserved and continues to work as before. (#15233)
```
dholth, line 19:
```suggestion
* Remove support for JLAP incremental repodata updates. Zstd-compressed repodata (`.json.zst`) functionality is preserved and continues to work as before. (#15233)
```
JLAP stood for "json lines append penultimate" due to adding data from the next-to-last line in the file.

tests/gateways/test_zstd.py
```diff
@@ -0,0 +1,265 @@
+# Copyright (C) 2012 Anaconda, Inc
```
dholth, line 1: test_repodata_zstd or a name more attached to "repodata"

tests/gateways/test_zstd.py
```diff
@@ -0,0 +1,265 @@
...
+    expected = CondaRepoInterface if not repodata_use_zst else ZstdRepoInterface
+    monkeypatch.setenv("CONDA_REPODATA_USE_ZST", str(repodata_use_zst))
+    reset_context()
+    assert get_repo_interface() is expected
+
+
+def test_zstd_not_404(mocker, package_server, tmp_path):
+    """
+    Test that exception is raised if `repodata.json.zst` produces something
+    other than a 404. For code coverage.
```
dholth, line 209: We should run all of the gateways tests by themselves and observe coverage within the repodata gateway module.

jezdez:
@dholth I don't fully understand your last comments, are they supposed to be conditions for your approval or are they directions what to change before merge? After merge?

danyeaw:
We discussed this on Zulip at https://conda.zulipchat.com/#narrow/channel/457620-ceps/topic/Does.20CEP-9.20.28Deprecation.29.20apply.20to.20Experimental.20Features.3F/with/576277296. There wasn't concerns about us removing JLAP. We also discussed that we need to work on better capabilities to do feature flagging as we test and roll out new features.

Issue #13896, opened 2024-05-06
Create deb and rpm packages with each conda release


marcoesters:
### Checklist

- [X] I added a descriptive title
- [X] I searched open issues and couldn't find a duplicate

### What?

Anaconda produces deb and rpm packages with each conda release. They are currently available in the following repositories:

* deb: https://repo.anaconda.com/pkgs/misc/debrepo/
* rpm: https://repo.anaconda.com/pkgs/misc/rpmrepo/

This is currently accomplished by installing `conda` into a new environment and packaging the directory into a deb and an rpm package. These packages are then indexed and uploaded to the repositories mentioned above.

These repositories are maintained and controlled by Anaconda. `conda` is, however, an open-source project and should thus be the entity creating these packages.

### Why?

The current deb and rpm releases for conda are built in a private repository by Anaconda. `conda`, however, is an open source product and should thus be the entity producing these artifacts and managing the repositories.

Integrating this process into the release process ensures that these package formats are available with each `conda` release. This can potentially be expanded to other package formats.

### User impact

`conda` gains control over all conda-related artifacts. This consolidates the codebase and facilitates bug reports. It also provides the opportunity to build other package formats with a new repository.

### Goals

* Create a repository for deb and rpm packages owned and controlled by `conda`.
* Incorporate the generation of these packages into the release process.

### Tasks

- [ ] Create repositories on a conda-owned domain for deb and rpm packages
- [ ] Generate a GPG key to sign the packages
- [ ] Create conda and upload conda packages with each release
- [ ] Create workflow to build deb and rpm packages on release and upload to repository

maurerle:
Hi, is there anything one can do to get the ball rolling with this?

danyeaw:
Hi @maurerle - this isn't one of the priorities that the core team is working on at the moment. However, if you would like to work on creating a GitHub Actions workflow that builds deb and rpms, that would get us closer to the overall solution.

janxkoci:
As far as I know, deb & rpm packages require sudo/root permissions to install. This defeats the whole (or the main) point of conda for me.

Also, the current installation script runs everywhere - what's the benefit of distro-specific packages? All I see is extra maintenance burden..

maurerle:
@janxkoci on a multi-user HPC system, it would be good if there is a multi user installation which updates automatically through unattended-updates.
This is not possible if every user downloads its own copy to wherever they want.
Keeping track of installed programs and their location is also much easier with a package manager installation.

[...]

Issue #14826, opened 2025-05-12
conda-deb: usage of outdated SHA1 signature in debian package signing


maurerle:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

1. Follow information of Debian install: https://docs.conda.io/projects/conda/en/latest/user-guide/install/rpm-debian.html
2. run sudo apt update --audit
3. get the following:
```
Warning: https://repo.anaconda.com/pkgs/misc/debrepo/conda/dists/stable/InRelease: Policy will reject signature within a year, see --audit for details
Audit: https://repo.anaconda.com/pkgs/misc/debrepo/conda/dists/stable/InRelease: Sub-process /usr/bin/sqv returned an error code (1), error message is:
   Signing key on 34161F5BF5EB1D4BFBBB8F0A8AEB4F8B29D82806 is not bound:
              No binding signature at time 2024-05-15T22:40:11Z
     because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance
     because: SHA1 is not considered secure since 2026-02-01T00:00:00Z
```

There is plenty of time until 2026 but its best to have this is mind sooner than to be surprised by it :)

zukunftsfirmaDE:
I get this message on any apt update (Debian 13/forky) run. Since we are only roughly one month from this becoming an error, I would like to ask the developers to prioritize this and create a new non SHA1 signature, and ship it with an update in January.

Some people update only once per week, or only bi-monthly, so I would suggest you tackle this now.

Appreciate your work!

GerMalaz:
It became an error:
Sub-process /usr/bin/sqv returned an error code (1), error message is: Signing key on 34161F5BF5EB1D4BFBBB8F0A8AEB4F8B29D82806 is not bound:            No binding signature at time 2024-05-15T22:40:11Z   because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance   because: SHA1 is not considered secure since 2026-02-01T00:00:00Z

zukunftsfirmaDE:
How can we better support the people writing this code? 

I would like to offer a bounty of 25 EUR for someone to fix this today

bentalexanderhaase:
> How can we better support the people writing this code?
> 
> I would like to offer a bounty of 25 EUR for someone to fix this today

A simple (but janky) workaround is to force APT to use `gpgv` instead of `sqv` until Anaconda fixes this upstream:

```bash
[...]

PR #15694, opened 2026-02-06
Docs: remove RPM and Debian installation page

danyeaw:
### Description

Anaconda built rpm and deb installers in the past, but they stopped in 2025. This PR removes the documentation about it since it is out of date and this is no longer a supported installation format. Some content on the page about multi-user installations was worth hanging on to, so I moved it to the existing admin docs.

One thing I'm not sure about is we could also keep the page and present a warning that it is no longer supported and keep the instructions for older installations. However, I decided not to because I didn't like having it in the Table of Contents. We could also move it to some type of historical / archive section for reference.

Closes: https://github.com/conda/conda/issues/13896 and https://github.com/conda/conda/issues/14826

### Checklist - did you ...

### DIFF
--- a/news/15694-remove-rpm-deb-docs
+++ b/news/15694-remove-rpm-deb-docs
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* Remove docs about RPM and DEB packages since no longer supported.
+
+### Other
+
+* <news item>


travishathaway:
@danyeaw,

We don't need to worry about archiving anything because that's already done for us by Read the Docs.

Out of scope for this pull request, but something else to think about is this page:

- https://www.anaconda.com/blog/rpm-and-debian-repositories-for-miniconda

I would say we should either remove that or put a warning up about it not being supported anymore. @kathatherine, maybe you could help us with that?

travishathaway (approved):
Looks good. I'm curious whether the instructions we give for shared conda installations actually reflects what sys-admins are actually doing. Regardless, I think it provides a good start point for people to customize to their personal needs later.

kathatherine:
> Out of scope for this pull request, but something else to think about is this page:
> 
> * https://www.anaconda.com/blog/rpm-and-debian-repositories-for-miniconda
> 
> I would say we should either remove that or put a warning up about it not being supported anymore. @kathatherine, maybe you could help us with that?

Sure! I'll make a ticket with Anaconda's Marketing department to get that figured out.

kathatherine:
Popping back in here to say for posterity that the blog post is going to be archived by the end of the month, depending on ticket prioritization.

PR #15697, opened 2026-02-09
Make the classic solver exit early with `conda remove <spec>` if the spec does not exist in the environment

agriyakhetarpal:
### Description

Split off from #15653 (see https://github.com/conda/conda/pull/15653#discussion_r2782856151). The `conda remove <spec>` command exits early for non-existent specs in the environment when the libmamba solver is configured (running this command does not call the libmamba solver), but this does not currently happen with the classic solver. This PR changes that behaviour, such that we call `PackagesNotFoundError` early on to prevent the classic solver from being invoked for non-existent specs.

cc: @jaimergp

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_remove.py
+++ b/conda/cli/main_remove.py
@@ -232,9 +232,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         if unmatched_specs := [
             str(spec) for spec in specs if not tuple(prefix_data.query(spec))
         ]:
-            raise PackagesNotFoundError(
-                tuple(sorted(unmatched_specs))
-            )
+            raise PackagesNotFoundError(tuple(sorted(unmatched_specs)))
 
         solver_backend = context.plugin_manager.get_cached_solver_backend()
         solver = solver_backend(prefix, channel_urls, subdirs, specs_to_remove=specs)


jaimergp:
Thanks, Agriya. This is a good change, and it makes sense, but also begs the question: are we going to put any effort in the classic solver if we are thinking of deprecating it? Pinging @conda/conda-maintainers for their opinion.

Mine:

- This change is probably ok and not controversial, so it wouldn't hurt to add.
- But we really need to draw the line and say we have feature-frozen the classic solver as it undergoes the split and deprecation.

news/15697-faster-conda-remove-for-classic-solver
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* `conda remove <spec>` with the classic solver now exits early if the spec to be removed does not exist in the environment. (#15697)
```
soapy1, line 3:
```suggestion
* `conda remove <spec>` now exits early if the spec to be removed does not exist in the environment. (#15697)
```

This change is solver agnostic. Testing this out this the classic, libmamba and rattler solver I get the same result.

tests/cli/test_main_remove.py
```diff
@@ -134,3 +134,22 @@ def test_remove_all_default_activation_env(
                 "--all",
                 f"--prefix={prefix}",
             )
+
+
+def test_remove_early_existence_check(
+    tmp_env: TmpEnvFixture, conda_cli: CondaCLIFixture, mocker
+):
+    """Verify that the (classic) solver is not invoked when packages don't exist in prefix."""
+    with tmp_env("python") as prefix:
```
soapy1:
```suggestion
def test_remove_early_existence_check(
    empty_env: TmpEnvFixture, conda_cli: CondaCLIFixture, mocker
):
    """Verify that the (classic) solver is not invoked when packages don't exist in prefix."""
    with empty_env as prefix:
```

This is maybe an opportunity to use @kenodegard's [empty environment fixture](https://github.com/conda/conda/blob/main/conda/testing/fixtures.py#L544). Since this test should still be valid for an empty conda environment. Looks like it speeds these tests up quite a bit:

With `empty_env`
```
 $ hyperfine -r 10 'python -m pytest tests/cli/test_main_remove.py::test_remove_early_existence_check'
Benchmark 1: python -m pytest tests/cli/test_main_remove.py::test_remove_early_existence_check
  Time (mean ± σ):      1.076 s ±  0.064 s    [User: 1.000 s, System: 0.090 s]
  Range (min … max):    0.999 s …  1.187 s    10 runs
```

With `tmp_env`
```
$ hyperfine -r 10 'python -m pytest tests/cli/test_main_remove.py::test_remove_early_existence_check'
Benchmark 1: python -m pytest tests/cli/test_main_remove.py::test_remove_early_existence_check
  Time (mean ± σ):      5.599 s ±  0.251 s    [User: 4.794 s, System: 0.839 s]
  Range (min … max):    5.268 s …  5.981 s    10 runs
```

soapy1 (commented):
Looks good to me if other folks are also happy with this API change 👍

news/15697-faster-conda-remove-for-classic-solver
agriyakhetarpal: Oh, actually I added this because the libmamba solver already exits early, even without this change, even though the change is applicable to all solvers. This change is essentially a no-op for the libmamba solver.

tests/cli/test_main_remove.py
agriyakhetarpal: Oh, very nice! Thank you @soapy1, I missed that we had such a fixture :D Accepting and resolving.

tests/cli/test_main_remove.py
```diff
@@ -134,3 +134,22 @@ def test_remove_all_default_activation_env(
...
+        )
+
+        with pytest.raises(PackagesNotFoundError) as exc:
+            conda_cli("remove", f"--prefix={prefix}", "nonexistent-package", "--yes")
+
+        mock_solver_backend.assert_not_called()
+
+        exception_string = str(exc.value)
+        assert "nonexistent-package" in exception_string
+        assert "missing from the target environment" in exception_string
```
soapy1: Oh my, sorry to have stern you wrong here. This doesn't work with python >= 3.11 (I was originally testing this on python 3.10 :sweat: ). You can see this deprecation warning running it on python 3.11:
```
DeprecationWarning: pathlib.Path.__enter__() is deprecated and scheduled for removal in Python 3.13; Path objects as a context manager is a no-op
```

`empty_env` returns a Path so you don't need the `with empty_env....`. You can just use the path that it passes back as the prefix.

```suggestion
    empty_env: Path, conda_cli: CondaCLIFixture, mocker
):
    """Verify that the (classic) solver is not invoked when packages don't exist in prefix."""
    mock_solver_backend = mocker.patch(
        "conda.base.context.context.plugin_manager.get_cached_solver_backend"
    )

    with pytest.raises(PackagesNotFoundError) as exc:
        conda_cli("remove", f"--prefix={empty_env}", "nonexistent-package", "--yes")

    mock_solver_backend.assert_not_called()

    exception_string = str(exc.value)
    assert "nonexistent-package" in exception_string
    assert "missing from the target environment" in exception_string
```

tested on python 3.13 and 3.10 🙏

agriyakhetarpal: Ah, I'm sorry I missed the failed tests notification too, otherwise, I would have fixed it! Thank you!

~(However, it looks like the tests are failing again, so I'll take another look at this).~ Edit: the failures are unrelated, so all good.

conda/cli/main_remove.py
```diff
@@ -227,6 +227,15 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             specs = specs_from_args(args.package_names)
         channel_urls = ()
         subdirs = ()
+
+        prefix_data = PrefixData(prefix)
+        if unmatched_specs := [
+            spec for spec in specs if not tuple(prefix_data.query(spec))
+        ]:
+            raise PackagesNotFoundError(
+                tuple(sorted(str(spec) for spec in unmatched_specs))
```
kenodegard:
```suggestion
            str(spec) for spec in specs if not tuple(prefix_data.query(spec))
        ]:
            raise PackagesNotFoundError(
                tuple(sorted(unmatched_specs))
```

tests/cli/test_main_remove.py
```diff
@@ -134,3 +136,21 @@ def test_remove_all_default_activation_env(
...
+    )
+
+    with pytest.raises(PackagesNotFoundError) as exc:
+        conda_cli("remove", f"--prefix={empty_env}", "nonexistent-package", "--yes")
+
+    mock_solver_backend.assert_not_called()
+
+    exception_string = str(exc.value)
+    assert "nonexistent-package" in exception_string
+    assert "missing from the target environment" in exception_string
```
kenodegard: Lets use `pytest.raises(match)` instead:

```suggestion
    with pytest.raises(
        PackagesNotFoundError,
        match=r"(?s)missing from the target environment:.+nonexistent-package",
    ):
        conda_cli("remove", f"--prefix={empty_env}", "nonexistent-package", "--yes")

    mock_solver_backend.assert_not_called()
```

kenodegard:
pre-commit.ci autofix

PR #15698, opened 2026-02-09
Raise a `PackagesNotFoundError` for `conda search --envs <package>` if it does not exist

agriyakhetarpal:
### Description

Split off from #15653 (see https://github.com/conda/conda/pull/15653#discussion_r2782859339). Previously, `conda search --envs this-does-not-exist` simply printed out a header line as follows:

```
# Name            Version     Build  Channel        Location
```

with nothing below it. This PR changes behaviour, such that it now raises a `PackagesNotFoundError` for the non-existent package:

```
Searching environments for this-does-not-exist: failed

PackagesNotFoundError: The following packages are missing from the target environment:

  - this-does-not-exist
```

which, in my opinion, is clearer since it gives an indication to the user that the package they searched for isn't present, at least in the places where they're looking.

cc: @jaimergp 

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_remove.py
+++ b/conda/cli/main_remove.py
@@ -227,6 +227,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             specs = specs_from_args(args.package_names)
         channel_urls = ()
         subdirs = ()
+
+        prefix_data = PrefixData(prefix)
+        if unmatched_specs := [
+            str(spec) for spec in specs if not tuple(prefix_data.query(spec))
+        ]:
+            raise PackagesNotFoundError(tuple(sorted(unmatched_specs)))
+
         solver_backend = context.plugin_manager.get_cached_solver_backend()
         solver = solver_backend(prefix, channel_urls, subdirs, specs_to_remove=specs)
         txn = solver.solve_for_transaction()
--- a/news/15697-faster-conda-remove-for-classic-solver
+++ b/news/15697-faster-conda-remove-for-classic-solver
@@ -1,0 +1,19 @@
+### Enhancements
+
+* `conda remove <spec>` with the classic solver now exits early if the spec to be removed does not exist in the environment. (#15697)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/cli/test_main_remove.py
+++ b/tests/cli/test_main_remove.py
@@ -25,6 +25,8 @@ from conda.testing.integration import (
 )
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from pytest import MonkeyPatch
 
     from conda.testing.fixtures import CondaCLIFixture, TmpEnvFixture
@@ -134,3 +136,20 @@ def test_remove_all_default_activation_env(
                 "--all",
                 f"--prefix={prefix}",
             )
+
+
+def test_remove_early_existence_check(
+    empty_env: Path, conda_cli: CondaCLIFixture, mocker
+):
+    """Verify that the (classic) solver is not invoked when packages don't exist in prefix."""
+    mock_solver_backend = mocker.patch(
+        "conda.base.context.context.plugin_manager.get_cached_solver_backend"
+    )
+
+    with pytest.raises(
+        PackagesNotFoundError,
+        match=r"(?s)missing from the target environment:.+nonexistent-package",
+    ):
+        conda_cli("remove", f"--prefix={empty_env}", "nonexistent-package", "--yes")
+
+    mock_solver_backend.assert_not_called()


agriyakhetarpal:
Thanks for the review, @soapy1! Yes, I'd say the behaviour change is intentional, i.e., I believe it is better that we raise an error for a missing package as part of the search query.

Semantically speaking, end-user applications and websites with search functionality display some sort of indication if a search query doesn't raise any results, so if it is the case that we don't want to lean on raising an error here and still exit with a 0 return code, an alternative I'd suggest is that we should at the very least blurt out a helpful message that says "<package> was not found in the environment(s)" and list out the names and locations of the environments searched, or something similar in that regard. I think we agree that the header line right now isn't very helpful :D

jaimergp:
This in similar in scope to https://github.com/conda/conda/pull/15075, so even if we are breaking backwards compatibility with the exit code, we are actually fixing a bug, I'd say.

tests/cli/test_main_search.py
```diff
@@ -177,6 +177,24 @@ def test_search_envs_json(conda_cli: CondaCLIFixture):
     assert all(entry["package_records"][0]["name"] == search_for for entry in parsed)
 
 
+def test_search_envs_nonexistent(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundError) as excinfo:
+        conda_cli("search", "--envs", "does-not-exist")
```
ForgottenProgramme, line 182: Maybe it's worth adding an assertion for the exit code to make it clear that we actually expect a exit code 1 and that it's intentional. 
```suggestion
        stdout, stderr, code = conda_cli("search", "--envs", "does-not-exist")
        assert code == 1
```

agriyakhetarpal: I agree! Thanks for the suggestion, @ForgottenProgramme!

news/15698-search-envs-raises-error
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* `conda search --envs <package>` now raises an error when the package is not found in any environment. (#15698)
```
kenodegard: This isnt a bug:

```suggestion
* `conda search --envs <package>` now raises an error when the package is not found in any environment. (#15698)

### Bug fixes

* <news item>
```

tests/cli/test_main_search.py
```diff
@@ -177,6 +177,25 @@ def test_search_envs_json(conda_cli: CondaCLIFixture):
     assert all(entry["package_records"][0]["name"] == search_for for entry in parsed)
 
 
+def test_search_envs_nonexistent(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundError) as excinfo:
+        stdout, stderr, code = conda_cli("search", "--envs", "does-not-exist")
+        assert code == 1
+    assert "PackagesNotFoundError" in repr(excinfo.value)
```
kenodegard: 1. `assert code == 1` is never run since the exception will be raised and exit the with-clause
2. asserting the `repr(excinfo.value)` is a redundant check given we already require `PackagesNotFoundError` to be raised

```suggestion
    with pytest.raises(PackagesNotFoundError):
        conda_cli("search", "--envs", "does-not-exist")
```

tests/cli/test_main_search.py
```diff
@@ -177,6 +177,25 @@ def test_search_envs_json(conda_cli: CondaCLIFixture):
...
+
+
+def test_search_envs_nonexistent_json(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundError) as excinfo:
+        conda_cli("search", "--envs", "--json", "does-not-exist")
+    assert "PackagesNotFoundError" in repr(excinfo.value)
```
kenodegard:
```suggestion
def test_search_envs_nonexistent_json(conda_cli: CondaCLIFixture):
    with pytest.raises(PackagesNotFoundError):
        conda_cli("search", "--envs", "--json", "does-not-exist")
```

tests/cli/test_main_search.py
```diff
@@ -177,6 +177,25 @@ def test_search_envs_json(conda_cli: CondaCLIFixture):
...
+
+
+def test_search_envs_nonexistent_info(conda_cli: CondaCLIFixture):
+    with pytest.raises(PackagesNotFoundError) as excinfo:
+        conda_cli("search", "--envs", "--info", "does-not-exist")
+    assert "PackagesNotFoundError" in repr(excinfo.value)
```
kenodegard:
```suggestion
def test_search_envs_nonexistent_info(conda_cli: CondaCLIFixture):
    with pytest.raises(PackagesNotFoundError):
        conda_cli("search", "--envs", "--info", "does-not-exist")
```

dholth:
We should avoid showing the exception name to the user.

conda/cli/main_search.py
```diff
@@ -197,6 +197,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
                 }
                 for prefix, prefix_recs in prefix_matches
             )
+
+            if not ordered_result:
+                from ..exceptions import PackagesNotFoundError
+
+                raise PackagesNotFoundError([spec])
```
kenodegard, line 204: `conda list PATTERN` raises a `CondaValueError`, is there value in aligning with that/this?

dholth: Check with the nearest product manager; they would say "no implementation details". This is hard to do in Python of course.

jaimergp: PATTERN is not necessarily a "package spec" per se, it can be complicated regex.

news/15698-search-envs-raises-error
agriyakhetarpal: Maybe... I'd regard it as a bug since it showcases previously unintended behaviour. I'll choose to [disagree and commit](https://en.wikipedia.org/wiki/Disagree_and_commit), though. Thanks! :D

conda/cli/main_search.py
agriyakhetarpal: I would not lean towards a `CondaValueError`. When this PR is merged, #15653 will change this error to a more specific one, i.e., `PackagesNotFoundInPrefixError`, which should be clearer to the user.

jaimergp:
> We should avoid showing the exception name to the user.

That happens in _every_ error message we throw.

```
$ conda create -d does-not-exist
Channels:
 - conda-forge
Platform: osx-arm64
Solving environment: failed

  - does-not-exist

Current channels:

  - https://conda.anaconda.org/conda-forge

looking for, navigate to

    https://anaconda.org

```

PR #15701, opened 2026-02-12
Add aliases for environment spec plugins

soapy1:
### Description

part 1 to resolve https://github.com/conda-incubator/conda-lockfiles/issues/79

Key changes:
* support `aliases` for environment spec plugins
* align the aliases for export/env spec plugins 
* *does not* add the `cep-24` alias to the environment.yaml export plugin

This PR suggests adding `aliases` to environment spec plugins. This more closely aligns these plugins with environment export plugins. This way plugin authors may define the same set of aliases for their environment export and environment spec plugin implementations. Note, this does require plugin authors to be mindful about having consistent plugin names and aliases.

This would help cases where users expect to use the same format name for exporting and recreating an environment. For example:
```shell
# export environment. This is currently a completely valid and working command
$ conda export --format=environment-yaml --file=env.yml

# create a new environment from the export. On the latest released 
# conda this will fail because there is no `yml` plugin.
$ conda env create -n test --file=env.yml --env-spec=environment-yaml
```

Additionally, this PR updates the bulitin environment spec plugins to have similar aliases as their export counterparts. You can test this out using the above example.

### Checklist - did you ...

### DIFF
--- a/tests/env/test_create.py
+++ b/tests/env/test_create.py
@@ -441,11 +441,22 @@ def test_create_env_from_environment_yml_does_not_output_duplicate_warning(
     assert len(stdout.split("EnvironmentSectionNotValid")) == 2
 
 
+@pytest.mark.parametrize(
+    "target_format,file_name",
+    [
+        ("environment-yaml", "env.yaml"),
+        ("env.yml", "env.yaml"),
+        ("requirements", "env.txt"),
+        ("reqs", "env.txt"),
+    ],
+)
 @pytest.mark.integration
 def test_export_and_recreate_environment(
     conda_cli: CondaCLIFixture,
     tmp_env: TmpEnvFixture,
     path_factory: PathFactoryFixture,
+    target_format,
+    file_name,
 ):
     """
     Test that a user can recreate an environment with the same
@@ -454,36 +465,23 @@ def test_export_and_recreate_environment(
     """
     # Setup a simple environment
     with tmp_env("ca-certificates") as prefix:
-        # tuple of target format name and the file name to use
-        target_formats = [
-            ("environment-yaml", "env.yaml"),
-            ("env.yml", "env.yaml"),
-            ("requirements", "env.txt"),
-            ("reqs", "env.txt"),
-        ]
+        env_file_path = path_factory(file_name)
+        stdout, stderr, rc = conda_cli(
+            "export",
+            f"--prefix={prefix}",
+            f"--format={target_format}",
+            f"--file={env_file_path}",
+        )
+        assert rc == 0, "Unable to export env to format {target_format}"
 
-        for formats in target_formats:
-            target_format = formats[0]
-            file_name = formats[1]
-
-            # export the environment to yml format
-            env_file_path = path_factory(file_name)
-            stdout, stderr, rc = conda_cli(
-                "export",
-                f"--prefix={prefix}",
-                f"--format={target_format}",
-                f"--file={env_file_path}",
-            )
-            assert rc == 0, "Unable to export env to format {target_format}"
-
-            # recreate the environment
-            recreate_prefix = path_factory()
-            stdout, stderr, rc = conda_cli(
-                "env",
-                "create",
-                f"--prefix={recreate_prefix}",
-                f"--env-spec={target_format}",
-                f"--file={env_file_path}",
-                "--dry-run",
-            )
-            assert rc == 0, "Unable to recreate env from format {target_format}"
+        # recreate the environment
+        recreate_prefix = path_factory()
+        stdout, stderr, rc = conda_cli(
+            "env",
+            "create",
+            f"--prefix={recreate_prefix}",
+            f"--env-spec={target_format}",
+            f"--file={env_file_path}",
+            "--dry-run",
+        )
+        assert rc == 0, "Unable to recreate env from format {target_format}"


conda/plugins/environment_specifiers/cep_24.py
```diff
@@ -15,5 +15,6 @@ def conda_environment_specifiers():
 
     yield CondaEnvironmentSpecifier(
         name="cep-24",
+        aliases=("yaml", "yml", "environment-yaml"),
```
soapy1: hmmm, I'm not very sure if this is a good approach for solving this problem.

ryanskeith: I agree.  This is tricky. We have several formats that use the same file format.

soapy1:
> but if there were an end-to-end test for this new behavior, that would really help us evaluate whether we actually want this to be integrated into conda.

Good idea, added!

tests/env/test_create.py
```diff
@@ -442,3 +442,51 @@ def test_create_env_from_environment_yml_does_not_output_duplicate_warning(
...
+):
+    """
+    Test that a user can recreate an environment with the same
+    plugin name they used to export the environment.
+    Ref: https://github.com/conda-incubator/conda-lockfiles/issues/79
+    """
+    # Setup a simple environment
+    with tmp_env("python") as prefix:
+        # tuple of target format name and the file name to use
+        target_formats = [
```
soapy1: This is maybe more cleanly done by using pytest parameters. I opted not to in this case to avoid creating too many conda environments. But, maybe this sacrifices the clarity of the tests, happy to switch to a more common pattern if that feels better.

conda/plugins/environment_specifiers/cep_24.py
soapy1: To clarify, these aliases are names that users can use as arguments to the `--env-spec` flag. So these are all equivalent:
```
$ conda env create --file env.yml --env-spec yaml

$ conda env create --file env.yml --env-spec yml

$ conda env create --file env.yml --env-spec environment-yaml

$ conda env create --file env.yml --env-spec cep-24
```

jezdez:
```suggestion
        aliases=("environment.yml", "environment-yaml"),
```

jezdez (changes requested):
[CEP 24](https://conda.org/learn/ceps/cep-0024) speaks of the "environment.yml format" and I believe the "yml" format alias of the export command was purely for backward compatibility reasons left.

I would suggest we deprecate the underspecified "yml" format alias (maybe even hide it from the help text?) and then keep the env-spec and format plugins specific to environment.yml

conda/plugins/environment_specifiers/cep_24.py
soapy1: This suggestion is going to collide with the [`environment.yaml` plugin ](https://github.com/conda/conda/blob/main/conda/plugins/environment_specifiers/environment_yml.py#L17). We can add this alias once we finish up the deprecation cycle started in https://github.com/conda/conda/pull/15120

For example, this this change I get the error:
```
$ conda env create -p /tmp/somethg --file tests/env/support/simple.yml 

PluginError: Plugin name conflicts detected in environment specifiers.
Multiple plugins cannot use the same name or alias:
  - 'environment.yml' name or alias used by plugins: cep-24, environment.yml
 ```

File: conda/plugins/environment_specifiers/requirements_txt.py [hunk 3]
```diff
@@ -15,5 +15,6 @@ def conda_environment_specifiers():
 
     yield CondaEnvironmentSpecifier(
         name="requirements.txt",
+        aliases=("requirements", "reqs", "txt"),
```
jezdez:
```suggestion
        aliases=("requirements", "reqs"),
```

Is `txt` similar to `yml` in that we shouldn't allow in the env spec?

conda/plugins/environment_specifiers/cep_24.py
```diff
@@ -15,5 +15,6 @@ def conda_environment_specifiers():
 
     yield CondaEnvironmentSpecifier(
         name="cep-24",
+        aliases=("environment-yaml",),
```
ryanskeith: We could think about "env.yml" as well if desired to save the user typing.  Not a deal breaker for me.

tests/env/test_create.py
```diff
@@ -439,3 +439,50 @@ def test_create_env_from_environment_yml_does_not_output_duplicate_warning(
...
+    tmp_env: TmpEnvFixture,
+    path_factory: PathFactoryFixture,
+):
+    """
+    Test that a user can recreate an environment with the same
+    plugin name they used to export the environment.
+    Ref: https://github.com/conda-incubator/conda-lockfiles/issues/79
+    """
+    # Setup a simple environment
+    with tmp_env("python") as prefix:
```
jezdez: Do we need to install Python here?

conda/plugins/environment_specifiers/requirements_txt.py
ryanskeith: txt is a basic filename extension but I think we are ok as all the cool kids are going with structured file types.

tests/env/test_create.py
jezdez: I think it might be possible to do light envs using a smaller package

tests/env/test_create.py
soapy1: Yes, the requirements format needs to have packages to export. For example:

```
$ conda create -p /tmp/nothing   
Channels:
 - defaults
Platform: linux-64
Solving environment: done

## Package Plan ##

  environment location: /tmp/nothing



Proceed ([y]/n)? 


Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
#
#     $ conda activate /tmp/nothing
#
#
#     $ conda deactivate

$ conda export -p /tmp/nothing --format=requirements

CondaValueError: Cannot export requirements format: no requested packages found. Use 'explicit' format for environments with installed packages, or ensure the environment has package specifications.

```

We might want to consider this a bug? What do you think?

conda/plugins/environment_specifiers/requirements_txt.py
soapy1: we should remove it :fried_egg:

conda/plugins/environment_specifiers/cep_24.py
soapy1: If we add it here we should also add it to the export plugin

tests/env/test_create.py
soapy1: let's go for ca-certificates!

soapy1: Comparing this approach vs. parameterized with hyperfine I get the results

**one reused env**
```
$  hyperfine --warmup 2 --runs 10 'python -m pytest tests/env/test_create.py::test_export_and_recreate_environment'
Benchmark 1: python -m pytest tests/env/test_create.py::test_export_and_recreate_environment
  Time (mean ± σ):      5.833 s ±  0.098 s    [User: 4.234 s, System: 1.017 s]
  Range (min … max):    5.724 s …  6.049 s    10 runs
```

**pytest parameterized**
```
$  hyperfine --warmup 2 --runs 10 'python -m pytest tests/env/test_create.py::test_export_and_recreate_environment'
Benchmark 1: python -m pytest tests/env/test_create.py::test_export_and_recreate_environment
  Time (mean ± σ):      7.752 s ±  0.116 s    [User: 5.828 s, System: 1.444 s]
  Range (min … max):    7.607 s …  7.968 s    10 runs
```

About a 2 second different. I would say worth it for readability!

Issue #15717, opened 2026-02-19
Add pattern matching support to `default_filenames`


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Add `default_filenames` field to `CondaEnvironmentSpecifier` and implement two phase matching support using fnmatch patterns (like `*.conda-lock.yml`).

### Why is this needed?

We need the ability to pass in filenames to the `CondaEnvironmentSpecifier` plugin for running `conda env create --file <filename>` in order for plugins like conda-lockfiles to be able directly validate a format based on the filename instead of checking the format against all registered formats.

`CondaEnvironmentSpecifier` lacks `default_filenames` field (only exporter has it), and detection only supports exact matches. This prevents using patterns like `*.conda-lock.yml` that they must be supported as part of the conda-lock spec.

### What should happen?

Add the `default_filenames` field and add two phase matching:

1. Filename based filtering
2. Fallback autodetection (try all plugins)

Add field:
```python
@dataclass
class CondaEnvironmentSpecifier(CondaPlugin):
    name: str
    environment_spec: type[EnvironmentSpecBase]
    default_filenames: tuple[str, ...] = ()
```

Two-phase detection:
```python
# Phase 1: Filename-based filtering
candidates = [p for p in plugins 
              if any(fnmatch.fnmatch(basename, pat) for pat in p.default_filenames)]
if candidates:
    for plugin in candidates:
        if plugin.environment_spec(source).can_handle():
            return plugin

# Phase 2: Fallback autodetection (try all plugins)
for plugin in plugins:
    if plugin.environment_spec(source).can_handle():
        return plugin
```

**Acceptance Criteria:**
- [ ] Add `default_filenames` to `CondaEnvironmentSpecifier`
- [ ] Implement two-phase detection (filename filter → fallback)
- [ ] Update exporter detection to use fnmatch for consistency
- [ ] Backward compatible with exact filenames
- [ ] Add tests for pattern matching

### Additional Context

**Files:** `conda/plugins/types.py`, `conda/plugins/manager.py`

Issue #15718, opened 2026-02-19
Improve error messages for CondaEnvironmentSpecifier plugins


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

For plugins like conda-lockfiles, when no environment specifier plugins can handle a file I want conda to provide a clear error with guidance including the list of available formats with their filename patterns and commands to run to fix the error.

### Why is this needed?

Current errors don't explain filename pattern requirements or suggest using `--env-spec`. This was originally reported in https://github.com/conda-incubator/conda-lockfiles/issues/78 that the error just said `PluginError: Requested plugin 'conda-lock-v1' is unable to handle environment spec 'conda-lock'`.

### What should happen?

```bash
EnvironmentSpecError: Unable to detect format for 'conda-lock'

Available formats and filename patterns:
  - conda-lock-v1: *.conda-lock.yml (e.g., dev.conda-lock.yml)
  - pixi-lock-v6, rattler-lock-v6: pixi.lock

Suggestions:
  1. Rename to match a pattern (e.g., conda-lock.yml)
  2. Use: --env-spec conda-lock-v1
```

**Acceptance Criteria:**
- [ ] Query plugins for `default_filenames` patterns
- [ ] List available formats when no plugin can handle file
- [ ] Suggest using `--env-spec`
- [ ] Format parse errors clearly

### Additional Context

**Files:** `conda/plugins/manager.py`, `conda/exceptions.py`

danyeaw:
Hi @yane-anaconda, could I please get your input on error message wording? Thanks!

yane-anaconda:
@danyeaw I had a [couple options here](https://anaconda.atlassian.net/browse/URSA-244?atlOrigin=eyJpIjoiZDZiYTViZDUwNDgwNDEwZmE3NWY5Nzk2Y2ZlZjFjZjkiLCJwIjoiaiJ9)! What do you think of them?

Sorry for the late reply! Was on vacay all last week 🌴

dashagurova:
@kenodegard @jezdez 
Do we need to encode the format version directly in the option value? Like `conda-lock-v1` and `rattler-lock-v6`
Instead, could we simplify this to: 
- pixi-lock → maps to Rattler lock v6
- conda-lock →  maps to conda lock v1 (or later version when it comes around)
When a new upstream version is supported, we update our implementation, but the CLI surface stays stable and requires fewer choices/decisions from users?

[...]

PR #15719, opened 2026-02-20
Adds new `default_filenames` to `CondaEnvironmentSpecifier`

travishathaway:
### Description

fixes: #15717

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -63,6 +63,7 @@ if TYPE_CHECKING:
         CondaEnvironmentSpecifier,
         CondaHealthCheck,
         CondaPackageExtractor,
+        CondaPlugin,
         CondaPostCommand,
         CondaPostSolve,
         CondaPostTransactionAction,
@@ -516,22 +517,72 @@ class CondaPluginManager(pluggy.PluginManager):
         return PluginConfig(data)
 
     def get_environment_specifiers(
-        self, supports_detection: bool | None = None
+        self, *, supports_detection: bool | None = None, with_aliases: bool = True
     ) -> dict[str, CondaEnvironmentSpecifier]:
         """
-        Returns a mapping from environment specifier name to environment specifier.
+         Returns a mapping from environment specifier name to environment specifier.
+
+         :param supports_detection: ternary value that returns either everything, only supporting
+                                    detection or not supporting detection.
+        :param with_aliases: whether to include aliased values of environment specifiers.
         """
         if supports_detection is None:
-            return {
-                hook.name: hook
-                for hook in self.get_hook_results("environment_specifiers")
-            }
+            env_spec_hooks = [
+                h for h in self.get_hook_results("environment_specifiers")
+            ]
         else:
-            return {
-                hook.name: hook
-                for hook in self.get_hook_results("environment_specifiers")
-                if hook.environment_spec.detection_supported == supports_detection
-            }
+            env_spec_hooks = [
+                h
+                for h in self.get_hook_results("environment_specifiers")
+                if h.environment_spec.detection_supported == supports_detection
+            ]
+
+        if not with_aliases:
+            return {hook.name: hook for hook in env_spec_hooks}
+
+        try:
+            return self._get_name_and_alias_mapping(env_spec_hooks)
+        except PluginError as err:
+            raise PluginError(
+                f"Plugin name conflicts detected in environment specifiers.\n{err}"
+            )
+
+    def _get_name_and_alias_mapping(
+        self, plugins: Iterable[CondaPlugin]
+    ) -> dict[str, CondaEnvironmentSpecifier]:
+        """
+        Get a mapping from plugin names (including aliases) to plugin.
+
+        :param plugins: List of plugins that have a name and aliases attribute.
+        :return: Dict mapping format name to CondaEnvironmentExporter
+        :raises PluginError: If multiple exporters use the same format name or alias
+        """
+        mapping = {}
+        conflicts = {}  # format_name -> set of plugin names
+
+        for plugin in plugins:
+            for format_name in (plugin.name, *plugin.aliases):
+                if format_name in mapping:
+                    if format_name not in conflicts:
+                        conflicts[format_name] = {mapping[format_name].name}
+                    conflicts[format_name].add(plugin.name)
+                else:
+                    mapping[format_name] = plugin
+
+        if conflicts:
+            conflict_details = []
+            for format_name, plugin_names in sorted(conflicts.items()):
+                plugins_str = ", ".join(sorted(plugin_names))
+                conflict_details.append(
+                    f"'{format_name}' name or alias used by plugins: {plugins_str}"
+                )
+
+            raise PluginError(
+                f"Multiple plugins cannot use the same name or alias:"
+                f"{dashlist(conflict_details)}\n"
+            )
+
+        return mapping
 
     def get_environment_specifier_by_name(
         self,
@@ -660,7 +711,9 @@ class CondaPluginManager(pluggy.PluginManager):
         :param source: full path to the environment spec file or source
         :returns: an environment specifier plugin that can handle the provided file
         """
-        hooks = self.get_environment_specifiers(supports_detection=True)
+        hooks = self.get_environment_specifiers(
+            supports_detection=True, with_aliases=False
+        )
         basename = os.path.basename(source)
 
         # Filename detection
@@ -732,33 +785,12 @@ class CondaPluginManager(pluggy.PluginManager):
         :return: Dict mapping format name to CondaEnvironmentExporter
         :raises PluginError: If multiple exporters use the same format name or alias
         """
-        mapping = {}
-        conflicts = {}  # format_name -> set of plugin names
-
-        for plugin in self.get_environment_exporters():
-            for format_name in (plugin.name, *plugin.aliases):
-                if format_name in mapping:
-                    if format_name not in conflicts:
-                        conflicts[format_name] = {mapping[format_name].name}
-                    conflicts[format_name].add(plugin.name)
-                else:
-                    mapping[format_name] = plugin
-
-        if conflicts:
-            conflict_details = []
-            for format_name, plugin_names in sorted(conflicts.items()):
-                plugins_str = ", ".join(sorted(plugin_names))
-                conflict_details.append(
-                    f"'{format_name}' used by plugins: {plugins_str}"
-                )
-
+        try:
+            return self._get_name_and_alias_mapping(self.get_environment_exporters())
+        except PluginError as err:
             raise PluginError(
-                f"Format name conflicts detected in environment exporters:"
-                f"{dashlist(conflict_details)}\n"
-                f"Multiple plugins cannot use the same format name or alias."
+                f"Format name conflicts detected in environment exporters.\n{err}"
             )
-
-        return mapping
 
     def detect_environment_exporter(self, filename: str) -> CondaEnvironmentExporter:
         """
--- a/conda/plugins/environment_exporters/environment_yml.py
+++ b/conda/plugins/environment_exporters/environment_yml.py
@@ -99,7 +99,7 @@ def conda_environment_exporters():
     """Register the built-in YAML and JSON environment exporters."""
     yield CondaEnvironmentExporter(
         name=ENVIRONMENT_YAML_FORMAT,
-        aliases=("yaml", "yml"),
+        aliases=("yaml", "yml", "env.yml"),
         default_filenames=("environment.yaml", "environment.yml"),
         export=export_yaml,
     )
--- a/conda/plugins/environment_specifiers/cep_24.py
+++ b/conda/plugins/environment_specifiers/cep_24.py
@@ -15,6 +15,7 @@ def conda_environment_specifiers():
 
     yield CondaEnvironmentSpecifier(
         name="cep-24",
+        aliases=("environment-yaml", "env.yml"),
         environment_spec=Cep24YamlFileSpec,
         default_filenames=("environment.yml", "environment.yaml"),
     )
--- a/conda/plugins/environment_specifiers/requirements_txt.py
+++ b/conda/plugins/environment_specifiers/requirements_txt.py
@@ -15,6 +15,7 @@ def conda_environment_specifiers():
 
     yield CondaEnvironmentSpecifier(
         name="requirements.txt",
+        aliases=("requirements", "reqs"),
         environment_spec=RequirementsSpec,
         default_filenames=("requirements.txt", "spec.txt"),
     )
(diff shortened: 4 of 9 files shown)


conda/plugins/manager.py
```diff
@@ -566,18 +567,71 @@ def get_environment_specifier_by_name(
...
+        :param basename: basename of the source file
+        :param hooks: mapping of environment specifier plugins
+        :returns: list of matching plugins, or None if no filename matches
+        """
+        candidates = [
+            hook
+            for hook_name, hook in hooks.items()
+            if hook.default_filenames
+            and any(
+                fnmatch.fnmatch(basename, pattern) for pattern in hook.default_filenames
```
dholth, line 649: Looks like the benefit here is that it provides [case insensitive match.](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch)

conda/plugins/manager.py
```diff
@@ -718,8 +815,11 @@ def detect_environment_exporter(self, filename: str) -> CondaEnvironmentExporter
 
         matches = []
         for exporter_config in self.get_environment_exporters():
-            # Check if basename exactly matches any of the default filenames
-            if basename in exporter_config.default_filenames:
+            # Check if basename matches any of the default filename patterns
+            if any(
+                fnmatch.fnmatch(basename, pattern)
+                for pattern in exporter_config.default_filenames
+            ):
```
kenodegard, line 815: By using `fnmatch.fnmatch` we are enforcing that all specs' filenames are case-insensitive, is that an assumption we can make or should we use `fnmatch.fnmatchcase`?

conda/plugins/manager.py
```diff
@@ -566,18 +567,71 @@ def get_environment_specifier_by_name(
                     f"Requested plugin '{name}' is unable to handle environment spec '{source}'"
                 )
 
-    def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier:
-        """Detect the environment specifier plugin for a given spec source
+    def _detect_filename_env_spec(
```
kenodegard, line 632: This duplicates code from `_detect_content_env_spec`

Would be better if `_detect_filename_env_spec` returns a collection of candidates which we then pass to `_detect_content_env_spec` for the `can_handle` checks

conda/plugins/manager.py
```diff
@@ -566,18 +567,71 @@ def get_environment_specifier_by_name(
...
+                    log.debug("EnvironmentSpec hook: %s can be %s", source, hook.name)
+                    found.append(hook)
+            except Exception as e:
+                log.error(
+                    "EnvironmentSpec hook: an error occurred when handling '%s' with plugin '%s'. %s",
+                    source,
+                    hook.name,
+                    e,
+                )
+                log.debug("%r", e, exc_info=e)
```
soapy1:
```suggestion
               raise PluginError(
                    dals(
                        f"""
                        An error occured when handling '{source}' with plugin '{hook.name}'.

                        {type(e).__name__}: {e}
                        """
                    )
                )
```

I think this should be raising a plugin error to be more consistent with the rest of the detection process. Currently if you provide an environment file with invaid contents you'll get back a stack trace. For example, consider the `environment.yml`

```
# looks like this is actually the form of a requirements.txt file!
numpy
```

Results in:
```
$ conda env create -p /tmp/e --file environment.yml
ERROR conda.plugins.manager:_detect_filename_env_spec(611): EnvironmentSpec hook: an error occurred when handling 'environment.yml' with plugin 'cep-24'. string indices must be integers
ERROR conda.plugins.manager:_detect_content_env_spec(653): EnvironmentSpec hook: an error occurred when handling 'environment.yml' with plugin 'cep-24'. string indices must be integers

      File "/home/sophia/projects/conda/conda/exception_handler.py", line 30, in __call__
        return func(*args, **kwargs)
      File "/home/sophia/projects/conda/conda/cli/main.py", line 62, in main_subshell
        exit_code = do_call(args, parser)
      File "/home/sophia/projects/conda/conda/cli/conda_argparse.py", line 205, in do_call
        result = getattr(module, func_name)(args, parser)
      File "/home/sophia/projects/conda/conda/notices/core.py", line 141, in wrapper
        return func(*args, **kwargs)
      File "/home/sophia/projects/conda/conda/cli/main_env_create.py", line 119, in execute
        env = spec.env
      File "/home/sophia/projects/conda/conda/env/specs/yaml_file.py", line 85, in env
        self._environment = env.from_file(self.filename)
      File "/home/sophia/projects/conda/conda/env/env.py", line 337, in from_file
        return from_yaml(yamlstr, filename=filename)
      File "/home/sophia/projects/conda/conda/env/env.py", line 286, in from_yaml
        errors = get_schema_errors(data)
      File "/home/sophia/projects/conda/conda/env/env.py", line 163, in get_schema_errors
        validator(data[key])
    TypeError: string indices must be integers
```

With the proposed change this becomes a formatted error message:
```
$ conda env create -p /tmp/e --file environment.yml

PluginError: An error occured when handling 'environment.yml' with plugin 'cep-24'.

TypeError: string indices must be integers
```

conda/plugins/environment_specifiers/environment_yml.py
```diff
@@ -16,4 +16,5 @@ def conda_environment_specifiers():
     yield CondaEnvironmentSpecifier(
         name="environment.yml",
         environment_spec=YamlFileSpec,
+        default_filenames=("environment.yml", "environment.yaml"),
```
soapy1, line 19: This plugin has autodetection disabled and eventually we'll want to remove it in favopur of the `cep-24` plugin. So I don't think this is needed. Also doesn't hurt to add tho!

conda/plugins/manager.py
```diff
@@ -566,18 +567,71 @@ def get_environment_specifier_by_name(
...
+        """Phase 1: Detect environment specifier by filename pattern matching.
 
-        Raises PluginError if more than one environment_spec plugin is found to be able to handle the file.
-        Raises EnvironmentSpecPluginNotDetected if no plugins were found.
+        :param source: full path to the environment spec file or source
+        :param basename: basename of the source file
+        :param hooks: mapping of environment specifier plugins
+        :returns: list of matching plugins, or None if no filename matches
+        """
+        candidates = [
```
soapy1: Is it ok to have multiple plugins have overlapping `default_filenames`? This is not allowed in the [export plugin](https://github.com/conda/conda/blob/main/conda/plugins/manager.py#L731), so maybe also shouldn't be allowed here.

conda/plugins/manager.py
travishathaway: I personally think a case insensitive match is okay.

conda/plugins/manager.py
travishathaway: @soapy1,

In theory this is nice, but it breaks the following test where we make sure that when detection for one env spec fails we don't halt the program execution 😬:

- tests/plugins/test_env_specs.py::test_naught_plugin_does_not_cause_unhandled_errors_during_detection

conda/plugins/manager.py
travishathaway: I think the way these two methods loop through and try to find applicable plugin hooks is different enough to warrant their own methods. I tried to understand what you were suggesting today but was unable to.

I apologize if there's something obvious here I'm missing.

conda/plugins/manager.py
travishathaway: If there's more than one in that list, we catch this later and raise exception. So, I think this would appear properly as an error and the user would be presented with remediation steps (e.g. removing one of the plugins).

conda/plugins/environment_specifiers/environment_yml.py
travishathaway: So, is it okay to leave in? 😅

conda/plugins/manager.py
soapy1: True, I see thie behaviour after testing it out with some dummy plugins!

Can I ask for 2 changes:
1. could you add a few tests that ensure plugins with overlapping `default_filenames` will produce the right error
2. Right now the error message that appears for colliding default filenames is the same as for colliding plugins. I think it would be easier for conda developers and plugins authors to understand what is going wrong if these error messages were more specifically informative for each error case. For example, filename-based filtering collisions could say something like `Multiple plugins are configured to read {filename}... <here are the plugins>... try using '--env-spec=<your desired plugin>'....`

conda/plugins/manager.py
soapy1: oh my, sorry for the bad piece of suggested code, thanks for trying it out! 

But I think this is a bug. Could we fix this and add a test before merging?

**Reproduction steps:**
1, Create a new `environment.yaml` file with invalid file contents
```
numpy
```

2. Run conda env create against the file and notice conda produces an error report
```
$ conda env create -p /tmp/e --file environment.yaml 
. . .
. . .

consider using the --no-plugins option to turn off plugins.
. . .
```

**What should happen**
If the provided file has an invalid format. conda should not produce an error report, instead it should provide a helpful error message.

conda/plugins/manager.py
travishathaway: @kenodegard,

Thought about this some more and I think the best way forward is to update the `get_environment_specifiers` method on the `CondaPluginManager` to be:

```python
    def get_environment_specifiers(self, supports_detection: bool | None = None) -> dict[str, CondaEnvironmentSpecifier]:
        """
        Returns a mapping from environment specifier name to environment specifier.
        """
        if supports_detection is None:
            return {
                hook.name: hook for hook in self.get_hook_results("environment_specifiers")
            }
        else:
            return {
                hook.name: hook
                for hook in self.get_hook_results("environment_specifiers")
                if hook.environment_spec.detection_supported == supports_detection
            }
```

This will allow us to get rid of the exception logic later down the line and hopefully make those methods easier to reason about.

conda/plugins/manager.py
```diff
@@ -566,82 +578,121 @@ def get_environment_specifier_by_name(
...
+        :param source: full path to the environment spec file or source
+        :param hooks: mapping of environment specifier plugins
+        :returns: tuple of (found plugins, autodetect disabled plugin names)
+        """
+        found = []
+
+        for hook_name, hook in hooks.items():
+            try:
+                if hook.environment_spec(source).can_handle():
+                    found.append(hook)
```
soapy1, line 695: It would be nice to keep these log lines. They have been helpful for debugging.
```suggestion
                    log.debug(
                        "EnvironmentSpec hook: %s can be %s",
                        source,
                        hook_name,
                    )
                    found.append(hook)
```

conda/plugins/manager.py
```diff
@@ -566,82 +578,121 @@ def get_environment_specifier_by_name(
...
+    ) -> list[CondaEnvironmentSpecifier]:
+        """Detect environment specifier by content-based autodetection.
+
+        :param source: full path to the environment spec file or source
+        :param hooks: mapping of environment specifier plugins
+        :returns: tuple of (found plugins, autodetect disabled plugin names)
+        """
+        found = []
+
+        for hook_name, hook in hooks.items():
```
soapy1, line 686:
```suggestion
        for hook_name, hook in hooks.items():
            log.debug("EnvironmentSpec hook: checking %s", hook_name)
```

conda/plugins/manager.py
```diff
@@ -566,82 +578,121 @@ def get_environment_specifier_by_name(
...
+                f"{dashlist(self.get_environment_specifiers())}"
+            )
+
+        if len(found) == 1:
+            try:
+                if found[0].environment_spec(source).can_handle():
+                    return found
+            except Exception as e:
+                raise PluginError(
+                    f"Failed to parse environment specification from file: {e}"
```
soapy1, line 668: It would be nice if this error provided a bit more context to the user + some information about how to resolve the issue.

```suggestion
                    f"Failed to parse environment specification from file using plugin '{found[0].name}': {e}\n\n"
                    "Try using --env-spec=<spec-name> to more exactly specify the environment spec\n"
                    "parser you want to use.\n\n"
                    "Available env specs:\n"
                    f"{dashlist(self.get_environment_specifiers())}"
```

conda/plugins/manager.py
```diff
@@ -566,82 +578,121 @@ def get_environment_specifier_by_name(
...
+            )
+        ]
+
+        if len(found) > 1:
+            raise PluginError(
+                f"Too many plugins found that can handle the environment file '{source}'.\n\n"
+                "Try using --env-spec=<spec-name> to more exactly specify the environment spec\n"
+                "parser you want to use.\n\n"
+                "Available env specs:\n"
+                f"{dashlist(self.get_environment_specifiers())}"
```
soapy1, line 659: These lines are repeated a few times. Might be a good opportunity to refactor into new Exception class that includes this in the formatting. Like,

```
class EnvSpecPluginNotFound(PluginError):
    def __init__(self, msg: str, plugins: list[str], *args, **kwargs):
        msg += "Try using --env-spec=<spec-name> to more exactly specify the environment spec\n"
                "parser you want to use.\n\n"
                "Available env specs:\n"
                f"{dashlist(plugins)}"
        super().__init__(msg, *args, **kwargs)
```

and then this can become `raise PluginError( f"Too many plugins found that can handle the environment file '{source}'.\n\n").

There is also this [EnvironmentSpecPluginNotDetected](https://github.com/conda/conda/blob/main/conda/exceptions.py#L1344) error that can be used.
Let's maybe get this in a follow up pr.

conda/plugins/manager.py
travishathaway: I'm going to skip including this because at this point that error message doesn't make sense for the user. We've successfully narrowed down the selection of an environment specifier to a single env spec, so this error is there to surface any parsing errors that have occurred while running `can_handle`. It's the responsibility of the parser error to instruct the user on what they can fix.

Hope that makes sense! I know this code is very complex and confusing. It's our fault for overloading that `can_handle` method to have multiple responsibilities 😬.

conda/plugins/manager.py
travishathaway: Luckily we have this as a follow up issue:

- https://github.com/conda/conda/issues/15718

That will be the perfect place to address this. I will link to your comment there.

travishathaway:
@soapy1,

I had to resolve some merge conflicts with a recent pull request you authored:

- https://github.com/conda/conda/pull/15701

I'm pretty sure I did the correct things here, but I just want to take one more look over it to make sure it's okay.

conda/plugins/manager.py
```diff
@@ -515,9 +516,40 @@ def get_config(self, data) -> PluginConfig:
...
+
+        if not with_aliases:
+            return {hook.name: hook for hook in env_spec_hooks}
+
+        try:
+            return self._get_name_and_alias_mapping(env_spec_hooks)
+        except PluginError as err:
+            raise PluginError(
+                f"Plugin name conflicts detected in environment specifiers.\n{err}"
+            )
```
travishathaway, line 548: @soapy1,

This was the biggest change I made. I added a new `with_aliases` keyword argument so that we can ignore the aliases when we want to. For example, when detect environment specifiers by filename or file content, we don't actually need the alias, and being able to not include them here saves us from extra deduplication logic elsewhere in the code.

travishathaway:
@soapy1,

Oops! Forgot auto-merge was on 😬. I think these changes should be fine though. The more important thing is that you are aware of them because you are also working quite often with this same code.

conda/plugins/manager.py
soapy1: Thanks for the heads up. This looks good to me. Will do a bit of QA on it this afternoon as well.

PR #15731, opened 2026-02-24
Demo: Environment spec plugin api refactor

soapy1:
### Description

This demo is an alternative approach to https://github.com/conda/conda/pull/15490.

It brings the form of the environment spec plugins closer to export plugins. In that, the `CondaEnvironmentSpecifier` dataclass becomes:

```python
@dataclass
class CondaEnvironmentSpecifier
    name: str
    validate: Callable[[PathType, str], bool]
    env: Callable[[str], Environment]
    detection_supported: bool
```

I think this implies a clearer way to extend functionality and puts the style of this plugin more in line with other existing plugins.

This demo also applies the suggested api changes from https://github.com/conda/conda/pull/15490. That is, passing the contents of the environment spec file to plugins. And caching the results of reading from the environment spec file.

Things that are missing from this demo:
* if this approach is acccepted, I think its probably a good option to deprecate the `conda.env.specs` module
* will require a change in conda-lockfiles

Additional possible changes:
* Stop accepting filename in the validation function. Plugins should only be validating on file contents. The function signature can becom `validate: Callable[[str], bool]`
* raise errors instead of returning `false` in validate functions. This allows for passing back more information to conda when validation for a plugin fails. Conda can use this information to provide more feedback in case of errors.
 
### Checklist - did you ...

### DIFF
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -628,7 +628,7 @@ class CondaEnvironmentSpecifier(CondaPlugin):
     name: str
     validate: Callable[[PathType, str], bool]
     env: Callable[[str], Environment]
-    detection_supported: bool
+    detection_supported: bool = True
 
 
 @dataclass
--- a/conda/plugins/environment_specifiers/environment_yml.py
+++ b/conda/plugins/environment_specifiers/environment_yml.py
@@ -57,6 +57,8 @@ def validate(filename: PathType, data: str, schema_errors: bool) -> bool:
     try:
         yaml_data = yaml.loads(data)
         if schema_errors:
+            if yaml_data is None:
+                return False
             errors = env.get_schema_errors(yaml_data)
             if errors:
                 return False
--- a/tests/cli/test_env.py
+++ b/tests/cli/test_env.py
@@ -18,6 +18,7 @@ from conda.core.prefix_data import PrefixData
 from conda.exceptions import (
     CondaEnvException,
     DryRunExit,
+    EnvironmentFileEmpty,
     EnvironmentFileNotFound,
     EnvironmentLocationNotFound,
     PackagesNotFoundError,
@@ -274,7 +275,7 @@ def test_conda_env_create_empty_file(
     tmp_file = path_factory(suffix=".yml")
     tmp_file.touch()
 
-    with pytest.raises(SpecNotFound):
+    with pytest.raises(EnvironmentFileEmpty):
         conda_cli("env", "create", f"--file={tmp_file}")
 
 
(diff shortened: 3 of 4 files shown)


conda/plugins/environment_specifiers/cep_24.py
```diff
@@ -1,19 +0,0 @@
...
-
-
-@hookimpl(tryfirst=True)
-def conda_environment_specifiers():
-    from ...env.specs.cep_24 import Cep24YamlFileSpec
-
-    yield CondaEnvironmentSpecifier(
-        name="cep-24",
-        environment_spec=Cep24YamlFileSpec,
-    )
```
jezdez, line 19: we'd probably want to keep that around? Since we don't have an experiments policy, the comment there is not binding

conda/plugins/environment_specifiers/explicit.py
```diff
@@ -5,15 +5,64 @@
...
+
+    return Environment(
+        platform=context.subdir,
+        explicit_packages=explicit_packages,
+    )
+
 
 @hookimpl
 def conda_environment_specifiers():
-    from ...env.specs.explicit import ExplicitSpec
```
jezdez, line 14: What happens with `ExplicitSpec`?

conda/plugins/types.py
```diff
@@ -592,22 +592,43 @@ def env(self) -> Environment:
...
+    Return type to use when defining a conda environment exporter plugin hook supporting a single platform.
 
-    :param name: name of the spec (e.g., ``environment_yaml``)
-    :param environment_spec: EnvironmentSpecBase subclass handler
+    :param name: name of the plugin (e.g., ``environment-yaml``)
+    :param validate: callable that determines if the plugin can handle a given file based on its path and/or content
+    :param env: callable that parses the file and returns an Environment object
+    :param detection_supported: boolean that determines if the plugin should be included in the set of available plugins
+        checked during environment_spec plugin detection. If False, the only way to use the plugin will be through explicitly
+        requesting it as a cli argument or setting in .condarc. By default, autodetection is enabled.
```
jezdez, line 625: It's a big API change FWIW, even with the "experimental" node, we can't really do this, I'd suggest to introduce a new plugin class

jezdez (commented):
I like this better since it flattens the plugin API, but I have concerns how to do this without breaking our promise on API stability

danyeaw:
Depends on https://github.com/conda/conda/issues/15757.

soapy1:
Closing this in favour of https://github.com/conda/conda/pull/15758. These merge conflicts were too difficult to all get right. #15758 also has the update suggested by @jezdez to introduce a new plugin class opposed to modifying the existing one.

PR #15732, opened 2026-02-24
Remove boltons dependency, vendor setutils, timeutils

dholth:
### Description

The 'boltons' library is designed so that its [individual modules can be vendored](https://boltons.readthedocs.io/en/latest/architecture.html#arch-integration).

We only use `IndexedSet` and two time functions. Vendor these to remove a 482kB dependency.

For this to provide a size benefit, conda-libmamba-solver would also need to remove [its boltons dependency](https://github.com/search?q=repo%3Aconda%2Fconda-libmamba-solver%20IndexedSet&type=code); but conda has already been updated to use duck typing for IndexedSet instead of a strict `isinstance()`

### Checklist - did you ...

### DIFF
--- a/conda/plan.py
+++ b/conda/plan.py
@@ -14,8 +14,7 @@ import sys
 from collections import defaultdict
 from logging import getLogger
 
-from _vendor.setutils import IndexedSet
-
+from ._vendor.setutils import IndexedSet
 from .base.constants import DEFAULTS_CHANNEL_NAME, UNKNOWN_CHANNEL
 from .base.context import context, reset_context
 from .common.constants import TRACE


conda/_vendor/setutils.py
```diff
@@ -0,0 +1,976 @@
+# Copyright (C) 2012 Anaconda, Inc
```
dholth, line 1: Should turn our own auto-prepend-copyright off for these files.

conda/_vendor/timeutils.py
```diff
@@ -0,0 +1,581 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2013, Mahmoud Hashemi
```
jezdez, line 3: The license would need to be added to LICENSE

jezdez (commented):
When we maintained the `_vendor` dir, we used the [vendoring](https://github.com/pradyunsg/vendoring) tool by Pradyun that was written for pip, to make sure we didn't need to maintain these by hand.

For the `boltons` case, I feel it's warranted to not reintroduce the `_vendor` dir since it would provide a slippery slope to go back to that anti-pattern. 

Let's copy setutils and timeutils to a different place like `conda.common` if it's so important to warrant using `boltons`. Is there a stdlib way to not use boltons?

dholth:
I also think it's a little odd that we require `IndexedSet` in the solver, but that would be a different refactor.
`timeutils` is an implementation detail which replaced a different dependency.
To me vendoring is a reliable pattern that is sometimes hard to keep going.

jaimergp:
I'd rather remove the odd reliance on `IndexedSet`. We don't really need it. We just need to de-dup the list while maintaining order via `list(dict.fromkeys(iterable))`.

dholth:
@jaimergp https://github.com/conda/conda/blob/main/conda/core/solve.py#L1420-L1424 requires that "final_precs must support set difference". Could be worked around.

https://github.com/conda/conda/blob/main/conda/core/solve.py#L256-L298 is the function that returns IndexedSet in the conda-libmamba-solver implementation, but the docstring claims it returns `tuple[PackageRef]` aka `PackageRecord`

jaimergp:
Thanks for the links. Exactly, is an arbitrary requirement. Whatever iterable comes in can be promoted to the set if needed, and then sorted again. IOW, I don't see as a hard requirement, more like a nice-to-have thing that ended up imposing a 3rd party requirement.

jezdez:
I'm glad we're finding this out, y'all!

PR #15739, opened 2026-02-26
Remove file validations and make requirements plugin validation more strict

soapy1:
### Description

This PR aims to remove the remaining file validation checks in all the builtin environment spec plugins. 

### Key changes:
* removing file name extension check from `can_handle`
* update the requirements.txt environment spec plugin to check that all lines are valid matchspecs

#### Notes: more restrictive requirements.txt plugin

Some plugins are very permissive in the scheme they support. For example, requirements environment files basically don't have a schema. So, when the file extension checking is removed we end up with collisions between plugins. 

For example, consider the environment.yaml
```
name: test
dependencies:
  - numpy
```

Is a valid environment.yaml file. And in the implementation of the requirements plugin, is a valid requirements file, because its just text. So, when  we run a create command (without file extension checking) we get the result:
```
$ conda env create -p /tmp/test --file environment.yaml

PluginError: Too many plugins found that can handle the environment file 'something.yaml':

cep-24, requirements.txt

Please make sure that you don't have any overlapping plugins installed.
```

This aims to improve the situation by checking that each line item in the requirements is a valid MatchSpec. So with this change there is less chance of collisions. And no valid environment.yaml files will be detected as a requirements file.
```
$ conda env create -p /tmp/test --file something.yaml                                                                                                                       
Channels:
 - defaults
Platform: linux-64
Solving environment: done
#
#
#     $ conda activate /tmp/test
#
#
#     $ conda deactivate
```

### Checklist - did you ...

### DIFF
--- a/tests/test_priority.py
+++ b/tests/test_priority.py
@@ -33,7 +33,7 @@ def test_reorder_channel_priority(
 ):
     # use "cheap" packages with no dependencies
     package2 = "zlib"
-    package1 = "ncurses"
+    package1 = "tzdata"
 
     # set pinned package
     if pinned_package:


conda/env/specs/requirements.py
```diff
@@ -90,20 +85,23 @@ def can_handle(self) -> bool:
...
             return False
-
-        # Extract the file extension (e.g., '.txt' or '' if no extension)
-        _, file_ext = os.path.splitext(self.filename)
-
-        # Check if the file has a supported extension
-        if not any(spec_ext == file_ext for spec_ext in self.extensions):
-            self.msg = f"File {self.filename} does not have a supported extension: {', '.join(self.extensions)}"
+        
+        if is_url(self.filename):
```
soapy1, line 89: I admit, this is a bit of an odd change. We check here if the file is a url because we'll say that the requirements plugin does not know how to read an environment from a remote location. This is a stop gap solution to fix test.

The whole caching/reading file should be dealt with in core conda instead of in conda plugins. https://github.com/conda/conda/pull/15490 or https://github.com/conda/conda/pull/15731 propose the api change that will move the responsibility of dealing with io to core conda. Once that change is made, this line diff will no longer be required.

conda/env/specs/yaml_file.py
```diff
@@ -41,25 +38,19 @@ def can_handle(self):
...
-
-        # Check if the file has a supported extension and exists
-        if not any(spec_ext == file_ext for spec_ext in YamlFileSpec.extensions):
-            return False
-
         try:
             yamlstr = env.load_file(self.filename)
             data = yaml.loads(yamlstr)
-            if data is None:
+            # We check for dict in order to avoid loading flat files as YAML.
```
ryanskeith, line 51: requirements.txt does look like a flat file.

conda/plugins/hookspec.py
ryanskeith, line 1: This is just a change is for future devs trying to make a new plugin so they don't go down the road of file checking.

conda/env/specs/requirements.py
travishathaway: That makes sense. These values should ideally just be "basenames", right?

travishathaway: Typically in CLI applications, this is the type of validation that would occur at the command line argument/option parsing layer. Click has always handled this type of thing very well:

- https://click.palletsprojects.com/en/stable/handling-files/#file-path-arguments

Issue #15742, opened 2026-02-27
Remove `IndexedSet`. Use `list(dict.fromkeys(iterable))` instead.


dholth:
I'd rather remove the odd reliance on `IndexedSet`. We don't really need it. We just need to de-dup the list while maintaining order via `list(dict.fromkeys(iterable))`.

_Originally posted by @jaimergp in https://github.com/conda/conda/issues/15732#issuecomment-3972035772_

kenodegard:
Or use `conda.common.iterators.unique`

jaimergp:
I wonder if we have to through the whole deprecation process for this change, given that technically we are changing the return type.

kenodegard:
It would still quack like a duck? So I think we'd be fine without

jaimergp:
If we return the dedup'd tuple it wouldn't support set operations, only the indexing bits. And maybe someone somewhere was relying on that. That number is virtually zero, though, I hope.

dholth:
See also https://github.com/conda/conda/pull/15732#issuecomment-3973829274

Seems like the important work happens in conda, not in the solver's return function. It would make sense to expect a generic sequence type from the solver and figure out all the set differences in that one conda function.

The main problem may the set math not deduplication.

jaimergp:
The main usage I'm seen now is in `PrefixGraph`, where they use set operations while expecting certain bits to maintain order too. But I think there are ways around it.

dholth:
It looks like this is closed.

PR #15744, opened 2026-02-27
Remove `boltons.setutils.IndexedSet`

jaimergp:
### Description

Part of https://github.com/conda/conda/issues/15742

`IndexedSet` was used for two different outcomes:

- Order-preserving set operations (mostly in `PrefixGraph`)
- Order-preserving deduplication of iterables

Deduplication can be easily replicated with `conda.common.iterators.unique()` followed by a `list()` or `tuple()`, depending on the use case, as long as all the elements are already collected.

In `for` loops where the deduplication was done as part of a more involved logic, I replaced the `IndexedSet` accumulator with a `dict[Any, None]`, because dictionaries are order-preserving for a while now. However this approach does not allow `set` operations like differences or intersections.

In `PrefixGraph`, the full power of the `IndexedSet` was actually used. I worked around those by replacing them with `dict[Any, None]` too, but then I also had to replace some APIs with their `dict` counterparts. For example:

- `set.add()` is sort of like `dict.setdefault()`.
- `set.remove()`, `set.discard()` and `set.pop()` can be modeled with `dict.pop(..., None)`.
- `set.update()` can be done with `dict.update()` as long as the input is also a dict, so I transformed the input iterable into a dict with `dict.fromkeys()`
- The one set difference I found, I had to simply iterate over the collection and `pop` elements one by one. Maybe not as efficient, but it's only one case.

### Checklist - did you ...

### DIFF
--- a/conda/cli/helpers.py
+++ b/conda/cli/helpers.py
@@ -13,7 +13,6 @@ from argparse import (
     _AppendAction,
     _HelpAction,
     _StoreAction,
-    _StoreTrueAction,
 )
 from typing import TYPE_CHECKING
 
@@ -107,7 +106,7 @@ def add_parser_create_install_update(p, prefix_required=False):
     # Add the file kwarg. We don't use {action="store", nargs='*'} as we don't
     # want to gobble up all arguments after --file.
     p.add_argument(
-        # "-f",  # FUTURE: 26.3: Enable this after deprecating alias in --force
+        "-f",
         "--file",
         default=[],
         action="append",
@@ -508,22 +507,9 @@ def add_parser_networking(p: ArgumentParser) -> _ArgumentGroup:
 
 def add_parser_package_install_options(p: ArgumentParser) -> _ArgumentGroup:
     from ..common.constants import NULL
-    from ..deprecations import deprecated
 
     package_install_options = p.add_argument_group(
         "Package Linking and Install-time Options"
-    )
-    package_install_options.add_argument(
-        "-f",
-        dest="force",
-        action=deprecated.action(
-            "25.9",
-            "26.3",
-            _StoreTrueAction,
-            addendum="Use `--force` instead.",
-        ),
-        default=NULL,
-        help=SUPPRESS,
     )
     package_install_options.add_argument(
         "--force",
--- a/conda/core/subdir_data.py
+++ b/conda/core/subdir_data.py
@@ -17,9 +17,7 @@ from typing import TYPE_CHECKING
 from ..auxlib.ish import dals
 from ..base.constants import CONDA_PACKAGE_EXTENSION_V1, REPODATA_FN
 from ..base.context import context
-from ..common.io import DummyExecutor, ThreadLimitedThreadPoolExecutor, dashlist
-from ..common.iterators import groupby_to_dict as groupby
-from ..common.iterators import unique
+from ..common.io import DummyExecutor, ThreadLimitedThreadPoolExecutor
 from ..common.path import url_to_path
 from ..common.serialize import json
 from ..common.url import join_url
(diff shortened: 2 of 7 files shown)


conda/core/solve.py
```diff
@@ -261,7 +260,7 @@ def solve_final_state(
         ignore_pinned=NULL,
         force_remove=NULL,
         should_retry_solve=False,
-    ):
+    ) -> tuple[PackageRecord, ...]:
```
dholth, line 263:
```suggestion
    ) -> Sequence[PackageRecord]:
```

I like generic types here. `tuple` implements `Sequence` but list implements `MutableSequence`

jaimergp: Yes, but we are returning tuples. Wasn't the general rule to be forgiving on inputs (and there we'd use Sequence), but explicit on outputs (where we say exactly what we are returning?).

dholth: If the abstract class says `-> Sequence[PackageRecord]`, then `conda-libmamba-solver` already complies with the type annotation by returning `IndexedSet` and the type checker gives us a squiggly line if we try to use any non-Sequence operations inside `conda`. This is just the way I like it.

jaimergp: Aren't we removing `IndexedSet` there too? https://github.com/conda/conda-libmamba-solver/pull/879

I don't particularly care, but the current approach we have followed in `conda` so far seems to be more about being as concrete as possible. A quick search in the repo:

- `-> tuple`: 81 results
- `-> list`: 60 results
- `-> Sequence`: 5 results
- `-> Iterable`: 63 results (but these are mostly `yield` functions)

jaimergp: @dholth, about [this comment](https://github.com/conda/conda/pull/15768#issuecomment-4031625952), this is not an abstract class. It's actually the classic Solver, which we subclass for libmamba, but it does return concrete types.

If we want an abstract class (and maybe we should), that's a concern for a separate PR.

jaimergp: Given the lack of response and Jannis' thumbs up to my counterpoint, I'll assume that we'll stick to `tuple[]` and hence resolve this comment.

dholth: I was in Peru most of that time.

I even think that list would be a more appropriate type for this result since it has to be built incrementally.

kenodegard: I'm strongly in favor of concrete return types as they indicate more clearly the intended _usage_

E.g., returning a `list` indicates the sequence is intended to be mutated further, returning a `tuple` indicates it should not be mutated

In this particular case we are returning the solved state, do we expect callers of `Solver.solve_final_state` to modify the state further? No → `tuple`, Yes → `list`

conda/cli/install.py
```diff
@@ -486,7 +484,7 @@ def revert_actions(prefix, revision=-1, index: Index | None = None):
     if not_found_in_index_specs:
         raise PackagesNotFoundError(not_found_in_index_specs)
 
-    final_precs = IndexedSet(PrefixGraph(link_precs).graph)  # toposort
+    final_precs = tuple(PrefixGraph(link_precs).graph)  # toposort
```
dholth, line 487: under which circumstances are there duplicates

conda/core/solve.py
```diff
@@ -122,7 +121,7 @@ def solve_for_transaction(
         force_remove=NULL,
         force_reinstall=NULL,
         should_retry_solve=False,
-    ):
+    ) -> UnlinkLinkTransaction:
         """Gives an UnlinkLinkTransaction instance that can be used to execute the solution
```
dholth:
```suggestion
        """Return an UnlinkLinkTransaction instance that can be used to execute the solution
```

conda/core/solve.py
dholth: I will approve this PR and not bikeshed on types.
My gut objection to tuple is that it implies building a copy of a list; and that tuple here makes IndexedSet on any subclass "a compile error" if we were using a strict language, until we update the other solver.
From the Python documentation, tuples are not conceptually "an immutable list". 

> Though tuples may seem similar to lists, they are often used in different situations and for different purposes. Tuples are [immutable](https://docs.python.org/3/glossary.html#term-immutable), and usually contain a heterogeneous sequence of elements that are accessed via unpacking (see later in this section) or indexing (or even by attribute in the case of [namedtuples](https://docs.python.org/3/library/collections.html#collections.namedtuple)). Lists are [mutable](https://docs.python.org/3/glossary.html#term-mutable), and their elements are usually homogeneous and are accessed by iterating over the list.

If we kept a copy of the tuple and returned it, so that mutating the structure would confuse another part of the code; or if we needed it to be hashable; then immutability would be necessary. Otherwise a list that we don't think you'll care to mutate would work as well.
In this case we have a homogenous list of items stored in an IndexedSet or a tuple.

PR #15753, opened 2026-03-02
Always export dependencies, even if they are empty

soapy1:
### Description

[cep-24](https://github.com/conda/ceps/blob/main/cep-0024.md#dependencies) says that `dependencies` is a required field of the environment.yaml spec. When exporting these environments, conda must include this field, but it may be an empty list.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/environment_exporters/environment_yml.py
+++ b/conda/plugins/environment_exporters/environment_yml.py
@@ -45,7 +45,6 @@ def to_dict(env: Environment) -> dict[str, Any]:
     # Convert requested_packages to dependencies list
     dependencies = []
     if env.requested_packages:
-        dependencies = []
         for item in env.requested_packages:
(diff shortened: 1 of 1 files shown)


conda/plugins/environment_exporters/environment_yml.py
```diff
@@ -43,6 +43,7 @@ def to_dict(env: Environment) -> dict[str, Any]:
         env_dict["channels"] = list(env.config.channels)
 
     # Convert requested_packages to dependencies list
+    dependencies = []
     if env.requested_packages:
         dependencies = []
         for item in env.requested_packages:
```
danyeaw, line 48:
```suggestion
    dependencies = []
    if env.requested_packages:
        for item in env.requested_packages:
```
I think we can remove the extra dependencies definition here.

Issue #15757, opened 2026-03-04
Discovery: py-rattler CondaEnvironmentSpecifier plugin


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

In https://github.com/conda/conda/pull/15731, @soapy1 proposed a change to the `CondaEnvironmentSpecifier` plugin to improve the API and performance. To make sure that this API is usable by future plugins, create a proof-of-concept plugin using py-rattler.

### Why is this needed?

To make sure we have the right API for EnvSpec plugins before settling on a design.

### What should happen?

Build a proof-of-concept plugin using py-rattler in a separate repo.

soapy1:
Demo implementation + notes here https://github.com/soapy1/conda-rattler-lockfile

From building this demo I think that the proposal in https://github.com/conda/conda/pull/15758 feels pretty good.  A few extra thoughts:
* there are still some rough edges to smooth out around maintining the 2 apis until the old api can be fully removed
* it might be nice to keep passing the full path to the environment file in case the plugin needs it for context. For example, if the environment format want to allow relative paths to packages, etc.

ryanskeith:
Comments about demo:
- This looks great.  I was amazed out how little code was necessary ~170 lines of code to make a plugin.
- I do appreciate the simplicity and also how this is echoing what we are doing with exporting.
- Though this has been done in another demo, I do like the content base approach.  These files should never be too large so storing the contents in memory should be cheap.  I also like the idea how it frees plugin developers.

I want to push back on the thought if we really need a transition period for the 2 API's.  I think the Environment Spec family is rather small (if any at all not outside our sphere of influence) and also advertised as experimental.  There is unnecessary complexity having to balance the two.

danyeaw:
+1 on the nice demo!

If we do change the API, we at least break https://github.com/jaimergp/conda-declarative/blob/main/conda_declarative/plugin.py, and any other non-public users of the plugin. It does say experimental in the docstring, but I'm not sure if that is sufficient to change a very public plugin hook API.

soapy1:
I agree that it's not ideal to maintain both plugin interfaces. But I do think we should. Also brought up in this comment https://github.com/conda/conda/pull/15731#discussion_r2865497870.

Is there anything that we are missing/need to discuss before making a decision about this direction?

Maybe @jezdez, do you have some time to weight in here (since you have also been involved in providing helpful feedback to previous iterations of this work)?

[...]

PR #15758, opened 2026-03-04
Demo: Environment spec plugin api refactor

soapy1:
### Description

This demo is an alternative approach to https://github.com/conda/conda/pull/15490.

It brings the form of the environment spec plugins closer to export plugins. In that, the `CondaEnvironmentSpecifier` dataclass becomes:

```python
@dataclass
class CondaEnvironmentSpecifier2
    name: str
    validate: Callable[[str], bool]
    env: Callable[[str], Environment]
    detection_supported: bool
```

I think this implies a clearer way to extend functionality and puts the style of this plugin more in line with other existing plugins.

This demo also applies the suggested api changes from https://github.com/conda/conda/pull/15490. That is, passing the contents of the environment spec file to plugins. And caching the results of reading from the environment spec file.

Additional key changes:
* Stop accepting filename in the validation function. Plugins should only be validating on file contents. The function signature can become `validate: Callable[[str], bool]`
* raise errors instead of returning `false` in validate functions. This allows for passing back more information to conda when validation for a plugin fails. Conda can use this information to provide more feedback in case of errors.

Things that are missing from this demo:
* if this approach is acccepted, I think its probably a good option to deprecate the `conda.env.specs` module
* will require a change in conda-lockfiles
 
### Checklist - did you ...

### DIFF
--- a/conda/cli/main_env_create.py
+++ b/conda/cli/main_env_create.py
@@ -116,16 +116,15 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     validate_file_exists(args.file)
 
     # detect the file format and get the env representation
-    spec_hook = context.plugin_manager.get_environment_specifier(
+    spec_hook = context.plugin_manager.get_environment_specifier2(
         source=args.file,
         name=context.environment_specifier,
     )
-    spec = spec_hook.environment_spec(args.file)
-    if context.subdir not in spec.available_platforms:
+    if context.subdir not in spec_hook.available_platforms:
         raise PlatformMismatchError(
-            [(args.file, spec.available_platforms)], context.subdir
+            [(args.file, spec_hook.available_platforms)], context.subdir
         )
-    env = spec.env_for(context.subdir)
+    env = spec_hook.env_for(context.subdir)
 
     # FIXME conda code currently requires args to have a name or prefix
     # don't overwrite name if it's given. gh-254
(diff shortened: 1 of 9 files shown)


soapy1:
Depends on https://github.com/conda/conda/issues/15757.

conda/plugins/manager.py
```diff
@@ -772,6 +774,236 @@ def get_environment_specifier(
...
+        :param filename: full path to the environment spec file/source
+        :param name: name of the environment plugin to load
+        :returns: an environment specifier plugin that matches the provided plugin name, or can handle the provided file
+        """
+        if not name:
+            return self.detect_environment_specifier2(source)
+        else:
+            return self.get_environment_specifier2_by_name(source=source, name=name)
+
+    def get_environment(
```
soapy1, line 1064: Not sure if this is giving too much responsibility to the plugin manager

conda/cli/main_env_create.py
```diff
@@ -111,12 +111,16 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     validate_file_exists(args.file)
 
     # detect the file format and get the env representation
-    spec_hook = context.plugin_manager.get_environment_specifier(
-        source=args.file,
-        name=context.environment_specifier,
+    env = context.plugin_manager.get_environment(
```
soapy1: Might want to rethink a better way to adopt the new api. We'll stick with this for now to finish proving out the demo.

PR #15761, opened 2026-03-05
Add first-party `--exclude-newer` policy

### LINKED ISSUE #15759: Add `--exclude-newer` support across the conda ecosystem
Add `--exclude-newer` to conda so packages newer than a configured cutoff are excluded from solving/search results. uv, pip, pixi, and rattler already have this. The name matches theirs.

Background: [Woodruff (Trail of Bits)](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns), [Nesbitt on cooldowns](https://nesbitt.io/2026/03/04/package-managers-need-to-cool-down.html), [Nesbitt on AI agents](https://nesbitt.io/2026/04/09/package-security-defenses-for-ai-agents.html).

```bash
conda install numpy --exclude-newer 7d
```

```yaml
# .condarc
exclude_newer: 7d

channel_settings:
  - channel: conda-forge
    exclude_newer: 3d

exclude_newer_package:
  openssl: false        # exempt
  ca-certificates: false
  numpy: 30d            # different cutoff
```

Accepted values include compact durations (`7d`, `3d12h`, `1w`), ISO 8601 durations (`P7D`), RFC 3339 timestamps, date-only values, plain seconds, and zero (`0`, `0d`) for no delay. Date-only values use the start of the next UTC day. Package overrides take precedence over channel overrides, and channel overrides take precedence over the global cutoff.

### Tracking

Cross-repo effort, roughly in dependency order:

- [x] **CEP 47**: specify the server-controlled `indexed_timestamp` field in package records -- [conda/ceps#154](https://github.com/conda/ceps/pull/154) (accepted and merged)
- [ ] **conda-index**: assign, persist, and emit `indexed_timestamp` in classic, sharded, and v3 repodata -- [conda/conda-index#341](https://github.com/conda/conda-index/issues/341), implemented by [conda/conda-index#342](https://github.com/conda/conda-index/pull/342) (ready for review)
- [ ] **rattler indexing and records**: add `indexed_timestamp` to records, indexing, solving, and Python bindings -- [conda/rattler#2502](https://github.com/conda/rattler/pull/2502) (draft, conflicting, changes requested)
- [x] **conda**: first-party `ExcludeNewerPolicy`, config and CLI, all policy scopes, classic solver and search support, docs for `26.9.0`, and fail-closed solver capability flags -- [#15761](https://github.com/conda/conda/pull/15761) (merged, targeted for conda 26.9.0, not yet released)
- [x] **mamba/libmambapy global primitive**: add `exclude_newer_timestamp` to `Database::Settings` -- [mamba-org/mamba#4228](https://github.com/mamba-org/mamba/pull/4228) (merged and released in [mamba 2.9.0](https://github.com/mamba-org/mamba/releases/tag/2.9.0))
- [ ] **conda-libmamba-solver**: consume conda's resolved policy for all scopes -- [conda/conda-libmamba-solver#905](https://github.com/conda/conda-libmamba-solver/pull/905) (rebased onto current main with the merged conda API fix, CI running, review required, depends on unreleased conda 26.9.0)
- [x] **rattler native policy**: add channel and package cutoffs in Rust -- [conda/rattler#2279](https://github.com/conda/rattler/pull/2279), [conda/rattler#2296](https://github.com/conda/rattler/pull/2296) (merged, not yet exposed as non-global py-rattler solve arguments)
- [ ] **conda-rattler-solver**: consume conda's resolved policy for all scopes -- [conda/conda-rattler-solver#49](https://github.com/conda/conda-rattler-solver/pull/49) (reviewed, changes required for trusted timestamps, cached packages, v3 records, and conda API compatibility, currently conflicting)
- [ ] **mamba CLI policy**: parse global and package policies in micromamba -- [mamba-org/mamba#4344](https://github.com/mamba-org/mamba/pull/4344) (open, review required, one CI job failing during CMake setup, no channel override support)

### How it works

Conda owns policy parsing and resolves relative cutoffs once per command. Package records prefer `indexed_timestamp` from CEP 47 and fall back to the builder-controlled `timestamp` when needed. Records without either timestamp are kept for compatibility.

Channel-side delivery of `indexed_timestamp` is still in progress in conda-index [#341](https://github.com/conda/conda-index/issues/341) through [PR #342](https://github.com/conda/conda-index/pull/342), and in rattler #2502. Until channels emit the field, the fallback does not provide CEP 47's server-controlled security property.

The v1 conda implementation does not add a generic repodata-filter plugin hook. Filtering happens where conda can enforce it consistently:

- **Classic solver**: filters reduced-index candidates and validates solved `link_precs` before transaction creation.
- **`conda search`**: filters conda-owned query results directly.
- **External solver backends**: must opt in with explicit global, channel, and package capability flags. Unsupported active policy scopes fail closed instead of being ignored.

The conda-libmamba-solver PR implements all three scopes. A pure global policy uses libmamba's native cutoff, while channel and package overrides use Python-side repodata filtering. The conda-rattler-solver PR is still under review and needs changes before it provides equivalent enforcement.

Originally named `cooldown` / `--cooldown`, renamed for ecosystem consistency.

Related: #6947

jezdez:
### Description

Adds first-party `--exclude-newer` support for conda as part of #15759.

`exclude_newer` delays eligibility for newly published packages by applying an absolute cutoff to package timestamps. The policy prefers `indexed_timestamp` when present. [CEP 47](https://github.com/conda/ceps/pull/154) is merged, and [conda-index#342](https://github.com/conda/conda-index/pull/342) tracks publisher support. Until channels emit `indexed_timestamp`, conda falls back to the builder-controlled `timestamp`. Records without either timestamp are kept for compatibility.

This intentionally does **not** add a generic `conda_repodata_filters` plugin hook. The hook was removed from this v1 because it would only cover `SubdirData`-based paths, would not apply to libmamba or rattler backends, and would make cache and performance semantics harder to reason about.

#### User-facing behavior

```yaml
# .condarc
exclude_newer: 7d

# per-channel overrides
channel_settings:
  - channel: conda-forge
    exclude_newer: 3d

# per-package overrides: exempt or use a different cutoff
exclude_newer_package:
  openssl: false
  ca-certificates: false
  numpy: 30d
```

CLI examples:

```bash
conda install numpy --exclude-newer 7d
conda create -n env python --exclude-newer 2026-04-01
conda update --all --exclude-newer P7D
```

Accepted values include compact durations (`7d`, `3d12h`, `1w`), ISO 8601 durations (`P7D`, `PT24H`), RFC 3339 timestamps, date-only values, plain seconds, and zero values (`0`, `0d`) for no delay, using the current command time as the cutoff. Date-only values match Pixi and Rattler semantics. `YYYY-MM-DD` means the start of the following day in UTC. Leave `exclude_newer` empty or unset to disable the global policy.

Compared with Pixi, this v1 intentionally supports the documented compact, ISO, and plain-seconds grammar instead of full `humantime` duration parity. For example, Pixi accepts spellings such as `7days`. Conda v1 keeps the accepted duration forms narrower for now.

#### Solver backend compatibility

This conda PR supports the classic solver and `conda search` directly.

- **Classic solver**: supported for global cutoffs, channel overrides, and per-package overrides. Filtering happens during reduced-index candidate retrieval, and solved packages are validated before transaction creation.
- **`conda search`**: supported independent of the selected solver backend because search applies the conda-owned policy directly to query results.
- **`conda-libmamba-solver`**: pending [conda-libmamba-solver#905](https://github.com/conda/conda-libmamba-solver/pull/905), which is non-draft and awaiting review. It supports all policy scopes. A global-only policy uses libmambapy's native `exclude_newer_timestamp`, added by [mamba#4228](https://github.com/mamba-org/mamba/pull/4228) and released in mamba 2.9.0. Channel and package overrides use Python-side filtering before repodata is loaded into libmamba.
- **`conda-rattler-solver`**: pending [conda-rattler-solver#49](https://github.com/conda/conda-rattler-solver/pull/49), which is non-draft, rebased onto current `main`, and awaiting review. It supports all policy scopes. A global-only policy uses py-rattler's native absolute cutoff. Channel and package overrides use Python-side repodata filtering.

Backends that do not opt in fail during solver initialization when the active policy would otherwise be ignored. A backend can declare global, channel, and package support independently so conda still fails closed for unsupported scopes.

#### Implementation details

- Adds a conda-owned `ExcludeNewerPolicy` that resolves config once per command using a single `now`.
- Moves neutral duration, date, and timestamp parsing into `conda.common.datetime` and keeps exclude-newer-specific policy in `conda.core.exclude_newer`.
- Keeps raw repodata cached and applies the policy at candidate or query time, avoiding stale filtered pickles for relative cutoffs.
- Applies the policy to `conda search` results and classic reduced-index candidate retrieval.
- Supports `.condarc` channel cutoffs through `channel_settings` and package overrides through `exclude_newer_package`.
- Validates solved `link_precs` before transaction creation so package-cache or backend bypasses fail closed.
- Preserves `indexed_timestamp` on `PackageRecord` so materialized records can enforce the preferred timestamp.
- Preserves backward-compatible aliases for old `cooldown` and `cooldown_exclude` settings.
- Uses explicit global, channel, and package solver capability flags. Unsupported backends fail closed when an active policy would otherwise be ignored.
- Documents the new user-facing behavior for the September `26.9.0` release.

#### Pixi comparison and follow-up work

A fresh pass over `prefix-dev/pixi` confirmed that this branch matches the core conda-package behavior needed for v1: date-only next-day UTC semantics, zero-duration cutoff behavior, relative cutoffs resolved at solve time, inclusion of unknown timestamps, per-package overrides, channel overrides, solver-time filtering, and final safety validation.

Deferred from this v1:

- Review, merge, and release the downstream solver integrations.
- Emit trusted `indexed_timestamp` values from channel indexes, tracked in conda-index#342.
- Replace Python-side filtering when public solver bindings expose native non-global policy scopes.
- Full Pixi and `humantime` duration grammar parity.
- PyPI exclude-newer behavior, which is out of scope for conda core.

Part of #15759

### Checklist - did you ...


conda/core/subdir_data.py
```diff
@@ -688,6 +709,12 @@ def _process_raw_repodata(
                     )
                     continue
 
+                if any(
+                    not record_filter.filter(fn, info)
+                    for record_filter in repodata_filters
+                ):
+                    continue
```
jaimergp: This is not marked as ready for review yet but let me comment on this part. By implementing the cooldown filter at the SubdirData level you will only have an effect on the classic solver and the `search` subcommand.

For the classic solver, it's actually more interesting to implement this in their own "index reduction" routines, because they are already processing repodata in there.

For rattler, their [solver](https://rattler.prefix.dev/py-rattler/solver/#rattler.solver.solver.solve) actually has a `exclude_newer` option already, so we only need to pass it there.

For libmamba, we never hit this bit, because we only use SubdirData to download the full JSON to cache and then we pass it without opening it. The sharded code is also using a different code path altogether. The 

So, at any rate, it feels to me that this is better implemented as part of the `Solver` API.

jezdez:
After reviewing the feedback (especially @jaimergp's comment about the SubdirData-level filter not reaching libmamba or rattler), here is the plan going forward.

**Rename to `--exclude-newer`**

Every other tool uses this name (uv, pip, pixi, rattler). `--cooldown` and `--cooldown-days` go away. The config key becomes `exclude_newer`. The filter field changes from `upload_timestamp` to `indexed_timestamp` to match the CEP rename (conda/ceps#154).

**Per-package overrides**

`cooldown_exclude` (flat list) becomes `exclude_newer_package` (dict), following uv's `exclude-newer-package` pattern. Each entry maps a package name to a duration, timestamp, or `false` (exempt entirely):

```yaml
exclude_newer: 7d
exclude_newer_package:
  openssl: false
  ca-certificates: false
  numpy: 30d
```

**Solver backend architecture**

The `conda_repodata_filters` hook stays in `SubdirData._process_raw_repodata()` for the classic solver and `conda search`. But each solver backend should handle `exclude_newer` natively instead of relying on `SubdirData` filtering:

- **rattler**: py-rattler's `solve_with_sparse_repodata()` already accepts `exclude_newer`. conda-rattler-solver just needs to pass `context.exclude_newer` through.
- **libmamba**: needs upstream support in `libmambapy`. The natural integration point is `Database::Settings`, which already exists for `matchspec_parser`. Adding `exclude_newer_timestamp` there covers both the monolithic JSON path (`set_repo_solvables_impl`) and the sharded path (`add_repo_from_packages_impl_loop`). I will open a PR on mamba-org/mamba for this.

The hookspec docstring claiming "regardless of which solver backend" will be fixed to reflect the actual scope.

Tracking issue updated: #15759

jezdez:
pre-commit.ci autofix

conda/core/subdir_data.py
dholth: I was going to say the same thing.

Bad to have plugins visiting all the repodata, could bog things down.

The sharded code only visits "possible dependencies". Since it's visiting less data could afford to filter by timestamp; this could reduce the total number of shards if filtered packages have dependencies that non-filtered packages lack. Shards code treats all repodata as shards even regular repodata.json; current implementation chooses to bypass shards code if there are no real shards.

In SubdirData I would want to start with unfiltered repodata, and only filter when lazily creating PackageRecord.

dholth:
After the solve, a second check in conda "are we installing 'newer' packages" should happen.

conda/core/subdir_data.py
jezdez: Thanks, this is the direction the branch now takes.

- The generic `conda_repodata_filters` hook and SubdirData-level filtering are gone.
- Classic applies `ExcludeNewerPolicy` during ReducedIndex candidate retrieval, instead of while processing raw repodata.
- `conda search` filters query results without mutating cached repodata.
- Solver support is explicit and fail-closed, so unsupported backends reject an active `--exclude-newer` policy instead of silently ignoring it.
- Rattler/libmamba support is left to native backend integration, where the cutoff can be passed through their solver APIs.

I also added the post-solve `link_precs` guard so cached records or backend bugs cannot bypass the policy.

dholth: https://github.com/conda/conda/blob/main/conda/_private/shards/shards.py#L354 and the other yield statement in this function would be a place to put a date filter, or it could be done during shards traversal which would also potentially limit the number of shards fetched.

jezdez:
@conda/conda-maintainers [CEP 47](https://github.com/conda/ceps/blob/main/cep-0047.md) has been minted, so I think we're ready to look at this PR

conda/core/subdir_data.py
jezdez: @dholth I'm not sure if I understand your comment. Are you suggesting changing the implementation?

conda/common/datetime.py
jaimergp, line 1: Cool module! I'd swear I've seen related utility functions through our codebases (but perhaps not in `conda/conda` per se, could be `conda-build`), so it's worth checking whether this is duplicating something that is already defined elsewhere. `MAX_SECONDS_TIMESTAMP` comes to mind if I think of timestamp logic (ms vs secs  handling), perhaps in conda-index business? 🤔

jezdez: Good catch. The seconds/ms threshold exists in `TimestampField` and in conda-index’s private `_make_seconds`, but those are field/indexing internals and do not cover the duration/RFC3339 cutoff parsing needed here. I kept this as a small common helper so `exclude_newer` can use it without depending on model serialization behavior or conda-index internals.

conda/core/subdir_data.py
jezdez: Resolving this as outdated. The PR no longer filters in SubdirData. Classic filters in reduced index/search, solver backends declare support, and link_precs are checked after solve.

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+    """Parse plain, compact, or ISO 8601 duration strings."""
+    value = value.strip()
+    try:
+        return timedelta(seconds=int(value))
+    except ValueError:
+        pass
+
+    match = _ISO8601_DURATION_RE.match(value)
+    if match:
+        if not any(group is not None for group in match.groups()):
```
kenodegard:
```suggestion
        if all(group is None for group in match.groups()):
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+    """Behavior for date-only strings passed to ``parse_datetime_to_timestamp``."""
+
+    REJECT = "reject"
+    NEXT_UTC_DAY = "next-utc-day"
+
+
+_COMPACT_DURATION_RE = re.compile(r"(\d+)\s*([wdhms])", re.IGNORECASE)
+_ISO8601_DURATION_RE = re.compile(
+    r"^P(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$",
+    re.IGNORECASE,
```
kenodegard: Let's annotate complex regex so they're easier to read:


```suggestion
    r"""
    ^P             # period/duration designator
    (?:(\d+)W)?    # number of week
    (?:(\d+)D)?    # number of days
    (?:T           # time designator
      (?:(\d+)H)?  # number of hours
      (?:(\d+)M)?  # number of minutes
      (?:(\d+)S)?  # number of seconds
    )?
    $
    """,
    re.IGNORECASE | re.VERBOSE,
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+"""Date, time, duration, and timestamp helpers."""
+
+from __future__ import annotations
+
+import re
+from datetime import date, datetime, timedelta, timezone
+from enum import Enum
+from types import MappingProxyType
+
+COMPACT_DURATION_UNITS = MappingProxyType(
```
kenodegard: Does this mean we should we stop using `frozendict`?

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+            seconds=(
+                weeks * COMPACT_DURATION_UNITS["w"]
+                + days * COMPACT_DURATION_UNITS["d"]
+                + hours * COMPACT_DURATION_UNITS["h"]
+                + minutes * COMPACT_DURATION_UNITS["m"]
+                + seconds
+            )
+        )
+
+    pairs = _COMPACT_DURATION_RE.findall(value)
```
kenodegard, line 74:
```suggestion
    # find all compact duration patterns (e.g., 2w, 3d, etc.)
    pairs = _COMPACT_DURATION_RE.findall(value)
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+    if not pairs:
+        return None
+
+    consumed = _COMPACT_DURATION_RE.sub("", value).strip()
```
kenodegard:
```suggestion
    # error if value contains invalid compact durations
    consumed = _COMPACT_DURATION_RE.sub("", value).strip()
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+
+def parse_duration(value: str) -> timedelta | None:
+    """Parse plain, compact, or ISO 8601 duration strings."""
+    value = value.strip()
+    try:
+        return timedelta(seconds=int(value))
+    except ValueError:
+        pass
+
+    match = _ISO8601_DURATION_RE.match(value)
```
kenodegard, line 55:
```suggestion
    # match ISO 8601 duration (e.g., P2W, P3D, etc.)
    match = _ISO8601_DURATION_RE.match(value)
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+    r"^P(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$",
+    re.IGNORECASE,
+)
+_DATE_ONLY_RE = re.compile(r"^\d{4}-\d{2}-\d{2}$")
+_COMPACT_DATE_ONLY_RE = re.compile(r"^\d{8}$")
+
+
+def parse_duration(value: str) -> timedelta | None:
+    """Parse plain, compact, or ISO 8601 duration strings."""
+    value = value.strip()
```
kenodegard, line 47:
```suggestion
    # try value as duration in seconds
    value = value.strip()
```

conda/common/datetime.py
```diff
@@ -0,0 +1,122 @@
...
+            int(amount) * COMPACT_DURATION_UNITS[unit.lower()] for amount, unit in pairs
+        )
+    )
+
+
+def parse_datetime_to_timestamp(
+    value: str,
+    *,
+    date_only: DateOnlyBehavior = DateOnlyBehavior.REJECT,
+) -> float | None:
```
kenodegard: This has only one call site and is always

```python
parse_datetime_to_timestamp(..., date_only=DateOnlyBehavior.NEXT_UTC_DAY)
```

So let's just drop `DateOnlyBehavior.REJECT`?

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+@dataclass(frozen=True)
+class _CutoffParser:
+    """Resolve user-provided cutoff values into absolute POSIX timestamps."""
+
+    now: float
+
+    @staticmethod
+    def is_false_override(value: object) -> bool:
```
kenodegard, line 35: Should we use `conda.auxlib.type_coercion.boolify`?

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,406 @@
...
+                return None
+
+        return cls(ChannelSelector.from_value(channel), cutoff)
+
+    def matches(self, record: PackageRecord | Mapping[str, Any]) -> bool:
+        return self.selector.matches(record)
+
+
+@dataclass(frozen=True)
+class ExcludeNewerPolicy:
```
kenodegard, line 240: This is a **very** different design from how we've approached much of our context values

It also looks like we only ever invoke it as `ExcludeNewerPolicy.from_context()` so I think we should make it a cached property of context

jezdez: I think that's fair, I can do the change back. This is just me trying out new patterns

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/core/exclude_newer.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/core/exclude_newer.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/common/datetime.py
jezdez: Addressed in a8f69cb6e.

conda/base/context.py
```diff
@@ -1594,6 +1618,31 @@ def description_map(self) -> dict[str, str]:
...
+                exclude_newer entries in channel_settings, and per-package
+                overrides can be set with exclude_newer_package.
+                """
+            ),
+            exclude_newer_package=dals(
+                """
+                Per-package overrides for the exclude_newer policy. Maps package
+                names to a duration string (e.g. "30d"), a timestamp, or false
+                to exempt the package entirely. For example:
+                  exclude_newer_package:
```
kenodegard, line 1641: Not a fan of introducing two new config values, can we add a special `*` key instead and opt for a single `exclude_newer` mapping?

```yaml
exclude_newer:
  *: 7d
  openssl: false
  numpy: 30d
```

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+                f"Invalid exclude_newer value {value!r}; duration must not be negative"
+            )
+        return self.now - duration_seconds
+
+    @staticmethod
+    def invalid(value: str) -> CondaValueError:
+        return CondaValueError(
+            f"Invalid exclude_newer value {value!r}; use e.g. 7d, P7D, "
+            "2026-04-01, or 2026-04-01T12:00:00Z"
+        )
```
kenodegard, line 95: We should define a new exception instead

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+
+    global_cutoff: float | None = None
+    channel_cutoffs: tuple[ChannelCutoff, ...] = ()
+    package_cutoffs: dict[str, float | None] | None = None
+    now: float = 0.0
+
+    @classmethod
+    def disabled(cls) -> ExcludeNewerPolicy:
+        return cls()
```
kenodegard: Can we define this as a constant instead? Alternatively do we need it at all?

conda/core/solve.py
```diff
@@ -68,6 +69,19 @@ class BaseSolver:
     _index: ReducedIndex | None
     _r: Resolve | None
 
+    supports_exclude_newer_global: ClassVar[bool] = False
+    supports_exclude_newer_channel: ClassVar[bool] = False
+    supports_exclude_newer_package: ClassVar[bool] = False
+
+    def __init_subclass__(cls, **kwargs):
```
kenodegard, line 76: Document why we need this. AFAIK this is because both CLS and CRS inherit from `Solver` (they have not been updated to inherit from `BaseSolver` yet).

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+        return bool(self.channel_cutoffs) and self.active
+
+    def cutoff_for(self, package_name: str | None) -> float | None:
+        if (
+            package_name
+            and self.package_cutoffs
+            and package_name in self.package_cutoffs
+        ):
+            return self.package_cutoffs[package_name]
+        return self.global_cutoff
```
kenodegard: Unused, delete?

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+    @staticmethod
+    def invalid(value: str) -> CondaValueError:
+        return CondaValueError(
+            f"Invalid exclude_newer value {value!r}; use e.g. 7d, P7D, "
+            "2026-04-01, or 2026-04-01T12:00:00Z"
+        )
+
+
+@dataclass(frozen=True)
+class ChannelSelector:
```
kenodegard: This shouldn't be imported and used directly, let's make it private?


```suggestion
class _ChannelSelector:
```

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+    @staticmethod
+    def normalize(value: object) -> str | None:
+        if value is None:
+            return None
+        value = str(value).strip().rstrip("/")
+        return value or None
+
+
+@dataclass(frozen=True)
+class ChannelCutoff:
```
kenodegard: This shouldn't be imported and used directly, let's make it private?


```suggestion
class _ChannelCutoff:
```

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+    def cutoff_for(self, package_name: str | None) -> float | None:
+        if (
+            package_name
+            and self.package_cutoffs
+            and package_name in self.package_cutoffs
+        ):
+            return self.package_cutoffs[package_name]
+        return self.global_cutoff
+
+    def cutoff_for_record(
```
kenodegard: Only used internally to `ExcludeNewerPolicy`, make private?


```suggestion
    def _cutoff_for_record(
```

conda/core/exclude_newer.py
```diff
@@ -0,0 +1,395 @@
...
+            timestamp = normalize_timestamp_seconds(value)
+        except (TypeError, ValueError):
+            return None
+        return timestamp
+
+    @staticmethod
+    def record_value(record: PackageRecord | Mapping[str, Any], key: str) -> Any:
+        if hasattr(record, "get"):
+            return record.get(key)
+        return getattr(record, key, None)
```
kenodegard: These are retrieval/normalization helpers, better suited as private module-level helper functions:


```suggestion
def _get_timestamp(record: PackageRecord | Mapping[str, Any]) -> float | None:
    value = _get_value(record, "indexed_timestamp") or _get_value(record, "timestamp")
    if not value:
        return None

    try:
        return normalize_timestamp_seconds(value)
    except (TypeError, ValueError):
        return None


def _get_value(record: PackageRecord | Mapping[str, Any], key: str) -> Any:
    if hasattr(record, "get"):
        return record.get(key)
    return getattr(record, key, None)
```

conda/base/context.py
jezdez: I’d keep these separate. `--exclude-newer` writes a scalar to the `exclude_newer` context parameter, so changing that parameter to a map requires a separate CLI-only value or custom translation. It also changes `CONDA_EXCLUDE_NEWER` from a scalar, while the current split matches uv’s `exclude-newer` and `exclude-newer-package` settings. Also, `*` needs quoting in YAML.

conda/core/exclude_newer.py
jezdez: I’d keep `CondaValueError` here. Nothing catches exclude-newer parse failures separately, and conda already uses `CondaValueError` for invalid user values. A new exception would not provide different handling. I did replace the semicolon-separated messages with sentences.

conda/core/exclude_newer.py
jezdez: Removed `disabled()` and now construct `ExcludeNewerPolicy()` directly in f6c026e91. The dependent solver PRs will need the same replacement.

conda/core/solve.py
jezdez: Added a docstring in f6c026e91 explaining that external solvers inherit from `Solver` and must opt into each capability rather than inherit classic’s flags.

conda/core/exclude_newer.py
jezdez: Removed `cutoff_for()` in f6c026e91.

conda/core/exclude_newer.py
jezdez: Renamed it to `_ChannelSelector` in f6c026e91 and removed its unused `value` field.

conda/core/exclude_newer.py
jezdez: Renamed it to `_ChannelCutoff` in f6c026e91.

conda/core/exclude_newer.py
jezdez: Renamed it to `_cutoff_for_record()` in f6c026e91.

conda/core/exclude_newer.py
jezdez: Moved record lookup and timestamp normalization to module-level `_get_value()` and `_get_timestamp()` helpers in f6c026e91.

conda/core/exclude_newer.py
kenodegard: Fair on the handling. I was less worried about catching it separately and more that `invalid()` is a factory for one specific `CondaValueError`. That shape usually maps better to a subclass (e.g., `PlatformMismatchError`). Personal preference though, not blocking 🤷🏼‍♂️

kenodegard (approved):

Issue #15577, opened 2026-01-09
Add Format Metadata to Environment Plugins


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Add `description` and `is_lockfile` optional fields to `CondaEnvironmentSpecifier` and `CondaEnvironmentExporter` plugin types so help text can display format descriptions and group lockfiles separately.

### Why is this needed?

Currently plugins only expose `name`, `aliases`, and `default_filenames`. Help text can't explain what each format does or distinguish lockfiles from environment specs.

### What should happen?

1. Add optional `description: str` and `is_lockfile: bool` fields to both plugin dataclasses in `conda/plugins/types.py`
2. Update built-in plugins (environment.yml, explicit, requirements.txt) to provide descriptions
3. Make fields optional with defaults: infer `is_lockfile` from name containing "lock", use name as fallback for `description`

**Files:**
- `conda/plugins/types.py`
- `conda/plugins/environment_specifiers/*.py`
- `conda/plugins/environment_exporters/*.py`

### Additional Context

Part of help text improvements for lockfiles.

soapy1:
I think it's nice to be able to get some extra metadata into the plugins. Though, I think the `is_lockfile` part of the api doesn't feel great imo. No strong objections, but a few thoughts:

I think allowing a plugin author to say that their plugin is a lockfile plugin introduces some set of errors where the plugin author and core conda disagree. For example, to core conda a lockfile is one that just lists out explicit packages. It doesn't need the plugin author to say. If core conda and the plugin author disagree, how do we resolve that? Is core conda meant to be able to make decisions based on this piece of information? 

I see that it would be nice be able to tell the user that some environment spec/export plugin is a lockfile. Is it possible to suggest that plugin authors use the `description` field to add the extra context that their plugin is a lockfile type plugin? Or that plugin authors might want to name their plugins in a way that suggests that it's a lockfile? This way the information about the plugin being intended for lockfiles can still be visible by conda cli users.

danyeaw:
A few of us discussed this interface today around @soapy1's comment above. Could we instead make this interface a `format_category: str`, and then we can pass in `"lockfile"`? We could also define the expected values in a `StrEnum`. This would make it more extensible to add another format to the help text.

travishathaway:
@soapy1, @jezdez and @danyeaw,

Please see my latest changes here and let me know what you think:

[...]

PR #15763, opened 2026-03-06
Add new metadata fields to environment plugins (`environment_format` and `description`)

travishathaway:
### Description

fixes: #15577

For full details about the changes, see the issue above. This implementation followed what was prescribed in the issue verbatim.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -9,9 +9,9 @@ Each type corresponds to the plugin hook for which it is used.
 
 from __future__ import annotations
 
+import enum
 import os
 from abc import ABC, abstractmethod
-from collections.abc import Callable
 from contextlib import nullcontext
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING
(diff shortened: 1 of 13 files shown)


soapy1 (approved):
This looks good to me ❇️  I'm not sure what those codspeed tests reporting a performance degradation is about. I don't think this change should be making any performance changes to the update command.

conda/plugins/types.py
```diff
@@ -660,6 +679,17 @@ def __post_init__(self):
...
 
+        # Set default description to name if not provided
+        if self.description is None:
+            self.description = self.name
+
+        # Infer is_lockfile from name if not explicitly set to True
+        # Note: If user explicitly sets is_lockfile=False but name contains "lock",
+        # the inference will override it. This is acceptable as lockfiles should
+        # have "lock" in their name by convention.
+        if not self.is_lockfile and "lock" in self.name.lower():
```
jezdez: I don't think looking at the name like that is a good API for a plugin hook

travishathaway: @jezdez, should we just remove that check then? I'm curious how we go forward.

PR #15767, opened 2026-03-09
Clear SIGALRM in finally block in timeout() function

### LINKED ISSUE #15702: conda/common/io.py::timeout() can leave SIGALRM enabled if callable raises an unexpected exception
### Checklist

### What happened?

The `timeout()` function in conda/conda/common/io.py sets an alarm, then tries to run a function.
```
384         signal.signal(signal.SIGALRM, interrupt)
385         signal.alarm(timeout_secs)
386
387         try:
388             ret = func(*args, **kwargs)
389             signal.alarm(0)
390             return ret
391         except (TimeoutException, KeyboardInterrupt):  # pragma: no cover
392             return default_return
```
The alarm is only cleared if `func()` returns normally.

If `func()` exits by raising an exception, then the alarm stays set. If the exception is something _other_ than a `TimeoutException` or `KeyboardInterrupt`, the caller might catch it, handle it, and go on... which can result in a the alarm firing and a `TimeoutException` being raised while executing arbitrary code at some later time.

In my case, what I was seeing were functions raising SSLError / SSLEOFError, which left the alarm set and interrupted transaction rollback, leaving me with corrupted installations.

Clearning the alarm in a `finally` block would handle all possible cases (normal return, expected exception return, unhandled exception return):

```
diff --git a/conda/common/io.py b/conda/common/io.py
index 2e7e56da0..e8dedd183 100644
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -386,10 +386,11 @@ def timeout(timeout_secs, func, *args, default_return=None, **kwargs):

         try:
             ret = func(*args, **kwargs)
-            signal.alarm(0)
             return ret
         except (TimeoutException, KeyboardInterrupt):  # pragma: no cover
             return default_return
+        finally:
+            signal.alarm(0)


 # use this for debugging, because ProcessPoolExecutor isn't pdb/ipdb friendly
```

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

Br1an67:
Fixes #15702

### Description

The `timeout()` function in `conda/common/io.py` sets a SIGALRM signal before running a wrapped function, but only cleared the alarm if the function returned normally. If the function raised an unexpected exception (e.g., `SSLError`, `SSLEOFError`), the alarm remained set and could fire at an arbitrary later time, causing a `TimeoutException` to interrupt unrelated code.

This fix moves the `signal.alarm(0)` call to a `finally` block, ensuring the alarm is always cleared regardless of how the function exits (normal return, expected exception, or unexpected exception).

### Checklist - did you ...

- [x] Add a file to the `news` directory for the next release's release notes?

```
 conda/common/io.py                          | 3 ++-
 news/15702-fix-timeout-signal-alarm-cleanup | 3 +++
 2 files changed, 5 insertions(+), 1 deletion(-)
```

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -389,10 +389,11 @@ def timeout(timeout_secs, func, *args, default_return=None, **kwargs):
 
         try:
             ret = func(*args, **kwargs)
-            signal.alarm(0)
             return ret
         except (TimeoutException, KeyboardInterrupt):  # pragma: no cover
             return default_return
+        finally:
+            signal.alarm(0)
 
 
 # use this for debugging, because ProcessPoolExecutor isn't pdb/ipdb friendly
--- a/news/15702-fix-timeout-signal-alarm-cleanup
+++ b/news/15702-fix-timeout-signal-alarm-cleanup
@@ -1,0 +1,3 @@
+### Bug fixes
+
+* Clear SIGALRM in ``finally`` block in ``timeout()`` function to prevent the alarm from firing unexpectedly when the wrapped function raises an unexpected exception.


travishathaway:
Hi @Br1an67,

We appreciate your contribution. Unfortunately, we still haven't determined the root cause of the issue you are trying to close. If you notice carefully, the issue still has the "pending::feedback" label which means we're still in the process of collecting vital information from the person who initially reported. Because of this, we are unable to consider this pull request for merging at this time.

In the future, we ask that you instead look for issue labeled, "good-first-issue" or "help-wanted". These are currently not very well organized, but we have set a goal to better organize these in the coming months.

We thank you for understanding.

jezdez:
Superseded by #16437, which includes the same alarm cleanup, regression coverage, and restoration of the previous handler. Thank you @Br1an67 for the original fix.

PR #15768, opened 2026-03-09
Docs: Tips and hints for type hints

jaimergp:
### Description

Comes from conversation in https://github.com/conda/conda/pull/15744#discussion_r2866427044

### Checklist - did you ...

### DIFF
--- a/docs/source/dev-guide/typing.md
+++ b/docs/source/dev-guide/typing.md
@@ -1,24 +1,24 @@
 # Type hinting
 
-`conda` is an old project that was created when Python type hinting didn't exist yet. As a result, many parts of its codebase are not annotated yet. That said, we strive to progressively improve the type annotations coverage with each contribution. In this page we summarise the guiding principles that you should apply when adding type annotations in your contributions.
+`conda` is an old project that was created when Python type hinting didn't exist yet. As a result, many parts of its codebase are not annotated. That said, we strive to progressively improve the type annotation coverage with each contribution. On this page, we summarize the guiding principles that you should apply when adding type annotations in your contributions.
 
 ## Main principles
 
-We subscribe to the [Typing Best Practices](https://typing.python.org/en/latest/reference/best_practices.html) recommended by the official Python documentation. Some additional tips can be found in [Google's Python Guide: Type Annotations](https://google.github.io/styleguide/pyguide.html#s3.19-type-annotations) (ignore the formatting concerns, those are dealt automatically by the pre-commit hooks).
+We adhere to the [Typing Best Practices](https://typing.python.org/en/latest/reference/best_practices.html) recommended by the official Python documentation. Some additional tips can be found in [Google's Python Guide: Type Annotations](https://google.github.io/styleguide/pyguide.html#s3.19-type-annotations) (ignore the formatting concerns, those are dealt with automatically by our pre-commit hooks).
 
 From those tips, we emphasize:
 
-- Be abstract on inputs and concrete on outputs. Tip: explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html).
+- Be abstract on inputs and concrete on outputs (tip: explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html)).
 - Use `from __future__ import annotations` in your modules to opt-in for the most recent type hinting capabilities (e.g. deferred evaluation).
 - If there's no return value, that actually means we are returning `None`. Annotate it!
 
 ## Consider using multiple signatures
 
-If the type hints are getting too complex (too many unions, many different return type hints), it may be time to consider whether you want to split the types in different signatures with `typing.overload` decorator. `conda.plugins.manager.CondaPluginManager.get_hook_results` is a good example for this use case.
+If your type hints are getting too complex (too many unions, many different return type hints), it may be time to consider whether you want to split the types in different signatures with `typing.overload` decorator. `conda.plugins.manager.CondaPluginManager.get_hook_results` is a good example for this use case.
 
 ## Custom type hints
 
-`conda` ships with some custom types for type hinting. They are spread around the codebase so sometimes they are not easy to find. This is a summary:
+`conda` ships with some custom types for type hinting. They are spread around the codebase, so sometimes they are not easy to find. This is a summary:
 
 - `conda.common.path.PathType`: Use it for values that represent file paths.
 - `conda.common.path.PathTypes`: Use it for an iterable of `PathType` values.
@@ -33,4 +33,4 @@ We use `ruff` to lint and format our codebase. We have selected a few type hinti
 
 The `linux-typing` job in `tests.yml` runs MyPy and uploads the [reports to codecov under the label `MyPy`](https://app.codecov.io/github/conda/conda?flags%5B0%5D=MyPy). Its only purpose is to track type hinting coverage, not to report type hinting errors (yet).
 
-In this [codecov.io webapp report](https://app.codecov.io/github/conda/conda/tree/main/conda?flags%5B0%5D=MyPy) you can find the coverage for every Python file in the codebase.
+In this [codecov.io webapp report](https://app.codecov.io/github/conda/conda/tree/main/conda?flags%5B0%5D=MyPy), you can find the coverage for every Python file in the codebase.


docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,69 @@
...
+    return list(value)
+```
+
+Note how we the return type hint is still `list[str]`. Using `Iterable[str]` is not necessarily wrong, but we are being too forgiving there: we do know we always return a `list`, so why not provide that information.
+
+Do explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html) to avoid having to create unnecessarily restrictive hints. For example, if you want to accept lists but not tuples, you might be tempted to use simply `list[...]`, but it would be even better to accept `MutableSequence[...]` for any objects implementing the same interface.
+
+## Prefer modern typing options
+
+Basically, use `from __future__ import annotations` in your modules and follow the linter rules. If you notice outdated type hints (e.g. using `Union[str, int]` instead of `str | int`) update them!
```
travishathaway: Using `from __future__ import annotations` isn't strictly necessary to support patterns like `str | None` since Python 3.10, but include it still has subtle affects on how on how postponed evaluation of type hints occurs.

See this link for more information:
- https://stackoverflow.com/questions/61544854/from-future-import-annotations

Because conda is now `python>=3.10`, I think any mention of "modern" type annotations with mentions of using the `__future__` module should reference that example rather than the example you have mentioned.

jaimergp:
Yes, those references are very useful! Maybe we can link to those and highlight the most important bits.

docs/source/dev-guide/typing.md
jaimergp: Maybe `Self` for classmethods, which was only introduced in 3.11?

jaimergp:
@travishathaway ready for another look <3

dholth:
The type guide says "For return values, prefer concrete types (list, dict, etc.) for concrete implementations. The return values of protocols and abstract base classes must be judged on a case-by-case basis."

@jaimergp we had a discussion about the return value of the solver abstract base class, where I thought it was appropriate to return `Sequence`.

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
+# Type hinting
+
+`conda` is an old project that was created when Python type hinting didn't exist yet. As a result, many parts of its codebase are not annotated yet. That said, we strive to progressively improve the type annotations coverage with each contribution. In this page we summarise the guiding principles that you should apply when adding type annotations in your contributions.
```
travishathaway:
```suggestion
`conda` is an old project that was created when Python type hinting didn't exist yet. As a result, many parts of its codebase are not annotated. That said, we strive to progressively improve the type annotation coverage with each contribution. On this page, we summarize the guiding principles that you should apply when adding type annotations in your contributions.
```

I know it's not officially, but I'm fairly certain we use American English spelling too 🇺🇸.

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Main principles
+
+We subscribe to the [Typing Best Practices](https://typing.python.org/en/latest/reference/best_practices.html) recommended by the official Python documentation. Some additional tips can be found in [Google's Python Guide: Type Annotations](https://google.github.io/styleguide/pyguide.html#s3.19-type-annotations) (ignore the formatting concerns, those are dealt automatically by the pre-commit hooks).
```
travishathaway:
```suggestion
We adhere to the [Typing Best Practices](https://typing.python.org/en/latest/reference/best_practices.html) recommended by the official Python documentation. Some additional tips can be found in [Google's Python Guide: Type Annotations](https://google.github.io/styleguide/pyguide.html#s3.19-type-annotations) (ignore the formatting concerns, those are dealt with automatically by our pre-commit hooks).
```

I think "adhere" sounds slightly better here. Feel free to take it or leave it.

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+From those tips, we emphasize:
+
+- Be abstract on inputs and concrete on outputs. Tip: explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html).
```
travishathaway:
```suggestion
- Be abstract on inputs and concrete on outputs (tip: explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html)).
```

I think this looks slightly better when it's included as a sentence fragment in parenthesis.

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+- Use `from __future__ import annotations` in your modules to opt-in for the most recent type hinting capabilities (e.g. deferred evaluation).
+- If there's no return value, that actually means we are returning `None`. Annotate it!
+
+## Consider using multiple signatures
+
+If the type hints are getting too complex (too many unions, many different return type hints), it may be time to consider whether you want to split the types in different signatures with `typing.overload` decorator. `conda.plugins.manager.CondaPluginManager.get_hook_results` is a good example for this use case.
```
travishathaway:
```suggestion
If your type hints are getting too complex (too many unions, many different return type hints), it may be time to consider whether you want to split the types in different signatures with `typing.overload` decorator. `conda.plugins.manager.CondaPluginManager.get_hook_results` is a good example for this use case.
```

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Custom type hints
+
+`conda` ships with some custom types for type hinting. They are spread around the codebase so sometimes they are not easy to find. This is a summary:
```
travishathaway:
```suggestion
`conda` ships with some custom types for type hinting. They are spread around the codebase, so sometimes they are not easy to find. This is a summary:
```

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+- `conda.common.path.PathType`: Use it for values that represent file paths.
+- `conda.common.path.PathTypes`: Use it for an iterable of `PathType` values.
+
+## Local development
+
+### Linters
+
+We use `ruff` to lint and format our codebase. We have selected a few type hinting related rules, but we have not added [`ANN`](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) yet. That said, you can enable it temporarily in your `pyproject.toml` while developing locally.
```
travishathaway, line 30: But what about the changes below? 😅 Aren't we turning that on with this PR? Or do we also need to install a separate package to enable it?

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Type hinting coverage
+
+The `linux-typing` job in `tests.yml` runs MyPy and uploads the [reports to codecov under the label `MyPy`](https://app.codecov.io/github/conda/conda?flags%5B0%5D=MyPy). Its only purpose is to track type hinting coverage, not to report type hinting errors (yet).
+
+In this [codecov.io webapp report](https://app.codecov.io/github/conda/conda/tree/main/conda?flags%5B0%5D=MyPy) you can find the coverage for every Python file in the codebase.
```
travishathaway:
```suggestion
In this [codecov.io webapp report](https://app.codecov.io/github/conda/conda/tree/main/conda?flags%5B0%5D=MyPy), you can find the coverage for every Python file in the codebase.
```

docs/source/dev-guide/typing.md
jaimergp: It's commented out. Pre-filled so it's easier to enable. The other lines are only "ignore rule for these files", so they don't have an effect,

travishathaway:
@jaimergp, I'll let you merge this in case there are any last minute changes you still need to make.

jaimergp:
Thanks, Travis. Let's see if anyone else in @conda/conda-maintainers has something to say.

docs/source/dev-guide/typing.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Main principles
+
+We adhere to the [Typing Best Practices](https://typing.python.org/en/latest/reference/best_practices.html) recommended by the official Python documentation. Some additional tips can be found in [Google's Python Guide: Type Annotations](https://google.github.io/styleguide/pyguide.html#s3.19-type-annotations) (ignore the formatting concerns, those are dealt with automatically by our pre-commit hooks).
+
+From those tips, we emphasize:
+
+- Be abstract on inputs and concrete on outputs (tip: explore the different abstract classes in [`collections.abc`](https://docs.python.org/3/library/collections.abc.html)).
+- Use `from __future__ import annotations` in your modules to opt-in for the most recent type hinting capabilities (e.g. deferred evaluation).
+- If there's no return value, that actually means we are returning `None`. Annotate it!
```
jezdez, line 13: I think this is where we differed in the past and I want to object to that statement

jezdez:
I'm mostly not super stoked about this, since the Python community at large and the conda community specifically hyperfocus on typing (but can't agree on one way!!), while there are much simpler and less dogmatic ways to structure code. It's the same odd framing that I found problematic about https://github.com/conda/ceps/pull/143#issuecomment-4078921995

jezdez:
If there would be a review status of "Meh", the PR would get it

docs/source/dev-guide/typing.md
jaimergp: I think it has some merit. Otherwise an unhinted return type can be understood as "I didn't check" or "returns None", which is very ambiguous. For `__init__`, an exception is provided in the linked docs. Perhaps I could include it here:

```suggestion
- If there's no return value, that actually means we are returning `None`. Annotate it. The only exception are the `__init__` methods in classes, which always return `None` by design, so it's redundant.
```

jaimergp:
I take it as a -0 vote, or are there ways to improve the guidelines to make that a +1, @jezdez?

docs/source/dev-guide/typing.md
kenodegard: I remember the past conversation was over my annotations of `__init__ -> None`, so explicitly calling that out as redundant and undesired makes sense

What about for our test suite? Do we want `-> None` on all of the tests? We should call this out explicitly too in these docs.

jaimergp: Hm, yes we probably don't need to enforce it in `test_*` functions either

jezdez:
Okay, my grumpiness has lessened

Issue #15573, opened 2026-01-09
Add Environment Specifier Plugin Support to `conda create`


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Enable `conda create` to parse YAML environment files using the existing environment specifier plugin system. Currently, `conda create` only supports simple text files, while `conda env create` can parse YAML files, explicit lockfiles, and other formats via plugins like conda-lockfiles provides.

### Why is this needed?

Users are confused about which command to use. `conda create` can't parse `environment.yaml` files and conda-lockfiles currently only work with `conda env create --file` because it uses the plugin system.

### What should happen?

Modify `conda create` to use the environment specifier plugin system (same as `conda env create`) to detect and parse environment files. This should support:
- YAML environment files (local and remote URLs)
- Explicit lockfiles
- conda-lockfiles and conda-lock plugins
- Simple text files (maintain backward compatibility)

Files to modify:
1. `conda/models/environment.py` - Update `Environment.from_cli()` to use `context.plugin_manager.get_environment_specifier()` to detect file format
2. `conda/cli/main_create.py` - May need updates for environment name extraction

Acceptance criteria:
- `conda create --file environment.yml` works like `conda env create --file environment.yml`
- `conda create --file explicit.txt` works with `@EXPLICIT` lockfiles
- `conda create --file conda-lock.yml` and other lockfiles that are support by conda-lockfiles
- `conda create --file requirements.txt` still works
- Remote YAML files work: `conda create --file https://example.com/environment.yml`
- All existing `conda create` functionality unchanged

### Additional Context

Part of replacing and aliasing `conda env create` in #13015. The plugin system already exists in `conda env create` - we're porting it to `conda create`. This will enable conda-lockfiles support in `conda create` and make all plugin-based formats available.

soapy1:
A demo of this functionality exists here: https://github.com/conda/conda/pull/15547

PR #15770, opened 2026-03-09
Add environment specifier support to conda create --file (#15573)

ryanskeith:
## Description

conda create --file now uses the environment specifier plugin system to read environment files (YAML, requirements.txt, explicit, etc.), matching conda env create. Previously it only supported flat package lists (one package per line) like `requirements.txt`.

### Changes
**conda/cli/main_create.py**

- Validate file existence before processing
- Validate that all --file arguments use the same format
- When -n/-p are omitted, read name/prefix from files via `get_name_prefix_from_env_files`
  - Raise a clear error if files do not define name or prefix
- Run clone vs file mutual-exclusivity check before file validation (so --clone + --file yields TooManyArgumentsError instead of EnvironmentFileNotFound)

**conda/models/environment.py**

- Environment.from_cli_with_file_envs: new method that uses get_environment_specifier for each --file instead of the old package-list parsing; returns (Environment, dict[str, Environment]) mapping file path to Environment
- Environment.from_cli: thin wrapper around from_cli_with_file_envs that returns only the merged Environment, keeping previous API promise
- CLI overrides file for name/prefix, also it is necessary to give a name/prefix in the multiple file case
- Environment.merge: name/prefix use last-wins semantics instead of first so it is consistent for all types
  - Except for channels of course which is appending and prioritizes by prepending.
- Environment.merge: external_packages merged by concatenation per key (aligned with other package lists)

**conda/cli/install.py**
- Uses Environment.from_cli_with_file_envs where the file→env map is needed for external package installers

**conda/cli/common.py**

Added:
- get_name_prefix_from_env_files: read name/prefix from env spec files (last file wins) exit early
- validate_environment_files_consistency: ensure all files share the same format
- validate_file_exists: validate file existence before processing

**tests/models/test_environment.py**

- Updated tests as Environment.merge behavior changes
- Updated test as spec_from_url is no longer used

**tests/test_create.py**

- Added simple test to ensure yaml files could be used
- Updated tests against mix packages as the error message was changed
- Added tests for cli behaviors
- Added tests for basic environment spec usage

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_create.py
+++ b/conda/cli/main_create.py
@@ -96,7 +96,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..gateways.disk.delete import rm_rf
     from ..reporters import confirm_yn
     from .common import (
-        get_name_prefix_from_env_files,
+        get_name_prefix_from_env_file,
         print_activate,
         validate_environment_files_consistency,
         validate_file_exists,
@@ -104,13 +104,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     )
     from .install import install, install_clone
 
-    if args.clone and args.file:
-        raise TooManyArgumentsError(
-            0,
-            len(args.file),
-            list(args.file),
-            "`--file` and `--clone` arguments are mutually exclusive.",
-        )
+    # When `--clone` is present, `--file` and `packages` are not allowed
+    if args.clone:
+        if args.file:
+            raise TooManyArgumentsError(
+                0,
+                len(args.file),
+                list(args.file),
+                "`--file` and `--clone` arguments are mutually exclusive.",
+            )
+        if args.packages:
+            raise TooManyArgumentsError(
+                0,
+                len(args.packages),
+                list(args.packages),
+                "Did not expect any new packages or arguments for `--clone`.",
+            )
 
     for fpath in args.file:
         validate_file_exists(fpath)
@@ -122,8 +131,10 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
                 "Multiple environment files were specified but no name or prefix was provided. "
                 "Please provide -n/--name or -p/--prefix when using multiple --file arguments."
             )
+
         if args.file:
-            name, prefix = get_name_prefix_from_env_files(args.file)
+            # We know there's only a single file being passed in at this point
+            name, prefix = get_name_prefix_from_env_file(args.file[0])
             if name is not None:
                 args.name = name
             if prefix is not None and args.name is None:
@@ -144,13 +155,6 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
                 "one of the arguments -n/--name -p/--prefix is required"
             )
 
-    if args.clone and args.packages:
-        raise TooManyArgumentsError(
-            0,
-            len(args.packages),
-            list(args.packages),
-            "Did not expect any new packages or arguments for `--clone`.",
-        )
     prefix_data = PrefixData.from_context(validate=True)
 
     if prefix_data.is_environment():
--- a/docs/source/dev-guide/plugins/environment_specifiers.rst
+++ b/docs/source/dev-guide/plugins/environment_specifiers.rst
@@ -46,7 +46,7 @@ Be sure to be very specific when implementing the ``can_handle`` method. It shou
 return a ``True`` if the file can be parsed by the plugin. Making the ``can_handle``
 method too permissive in the types of files it handles may lead to conflicts with other
 plugins. If multiple installed plugins are able to ``can_handle`` the same file type,
-conda will return an error to the user.  Also, assure that the ``can_handle`` method
+conda will return an error to the user.  Also, ensure that the ``can_handle`` method
 does not depend on a file extension or specific file name as users may specify the
 plugin by name to read in files.
 
--- a/docs/source/user-guide/tasks/manage-environments.rst
+++ b/docs/source/user-guide/tasks/manage-environments.rst
@@ -100,21 +100,41 @@ use the ``--no-default-packages`` flag:
 Creating an environment from an environment.yml file
 ====================================================
 
-Use the terminal for the following steps:
+Before you start, you need to use a text editor to create an environment file that
+looks like the following:
+
+.. code:: yaml
+
+   # environment.yml
+   name: myenv
+   channels:
+     - conda-forge
+   dependencies:
+     - python
+     - numpy
+
+The file above defines an environment named, "myenv" that uses ``conda-forge`` as its
+channel and adds the dependencies ``python`` and ``numpy``.
+
+Now that the ``environment.yml`` has been created, you can use the terminal for
+the following steps:
 
 #. Create the environment from the ``environment.yml`` file:
 
    .. code::
 
-      conda create -f environment.yml
-
-   Alternatively, ``conda env create --file environment.yml`` is
-   another way to create an environment from an environment.yml file.
+      conda create --file environment.yml
 
    The first line of the ``yml`` file sets the new environment's
    name. For details see :ref:`Creating an environment file manually
    <create-env-file-manually>`.
 
+   .. hint::
+
+      Creating environments from YAML files was previously done with
+      ``conda env create --file <file>``. This still works but is not the preferred
+      way of creating environments. This older invocation will be kept around to
+      ensure backwards compatibility with existing scripts and automations.
 
 #. Activate the new environment: ``conda activate myenv``
 
--- a/conda/cli/common.py
+++ b/conda/cli/common.py
@@ -309,23 +309,23 @@ def print_activate(env_name_or_prefix):
         print(message)  # TODO: use logger
 
 
-def get_name_prefix_from_env_files(files: list[str]) -> tuple[str | None, str | None]:
-    """Get name and prefix from environment spec files. Last file with either wins.
-
-    Takes both name and prefix from the same file to avoid mixing. Returns
-    (None, None) if no plugin can read the files or no name/prefix is present.
+def get_name_prefix_from_env_file(file: str) -> tuple[str | None, str | None]:
     """
-    for fpath in reversed(files or []):
-        try:
-            spec_hook = context.plugin_manager.get_environment_specifier(
-                source=fpath,
-                name=context.environment_specifier,
-            )
-            env = spec_hook.environment_spec(fpath).env
-            if env.name is not None or env.prefix is not None:
-                return env.name, env.prefix
-        except Exception:
-            pass
+    Get name and prefix from an environment spec file.
+
+    Intentionally swallows all errors and logs them.
+    """
+    try:
+        spec_hook = context.plugin_manager.get_environment_specifier(
+            source=file,
+            name=context.environment_specifier,
+        )
+        env = spec_hook.environment_spec(file).env
+        if env.name is not None or env.prefix is not None:
+            return env.name, env.prefix
+    except Exception as exc:
+        log.info(exc, exc_info=True)
+
     return None, None
 
 
--- a/conda/env/env.py
+++ b/conda/env/env.py
@@ -439,7 +439,7 @@ class EnvironmentYaml:
         ]
 
         return EnvironmentModel(
-            prefix=self.prefix or context.target_prefix,
+            prefix=self.prefix,
             platform=context.subdir,
             name=self.name,
             config=config,
--- a/tests/cli/test_common.py
+++ b/tests/cli/test_common.py
@@ -10,6 +10,7 @@ from pytest_mock import MockerFixture
 from conda.base.context import context, reset_context
 from conda.cli.common import (
     check_non_admin,
+    get_name_prefix_from_env_file,
     is_active_prefix,
     print_activate,
     validate_file_exists,
@@ -149,3 +150,142 @@ def test_print_activate_no_output(capsys, monkeypatch, env_var):
 
     captured = capsys.readouterr()
     assert captured.out == ""
+
+
+def test_get_name_prefix_from_env_file_with_both(mocker: MockerFixture):
+    """Test get_name_prefix_from_env_file returns name and prefix when both are present."""
+    # Mock the environment and spec hook
+    mock_env = mocker.Mock()
+    mock_env.name = "test_env"
+    mock_env.prefix = "/path/to/env"
+
+    mock_env_spec = mocker.Mock()
+    mock_env_spec.env = mock_env
+
+    mock_spec_hook = mocker.Mock()
+    mock_spec_hook.environment_spec.return_value = mock_env_spec
+
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        return_value=mock_spec_hook,
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name == "test_env"
+    assert prefix == "/path/to/env"
+
+
+def test_get_name_prefix_from_env_file_with_only_name(mocker: MockerFixture):
+    """Test get_name_prefix_from_env_file returns name when only name is present."""
+    # Mock the environment and spec hook
+    mock_env = mocker.Mock()
+    mock_env.name = "test_env"
+    mock_env.prefix = None
+
+    mock_env_spec = mocker.Mock()
+    mock_env_spec.env = mock_env
+
+    mock_spec_hook = mocker.Mock()
+    mock_spec_hook.environment_spec.return_value = mock_env_spec
+
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        return_value=mock_spec_hook,
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name == "test_env"
+    assert prefix is None
+
+
+def test_get_name_prefix_from_env_file_with_only_prefix(mocker: MockerFixture):
+    """Test get_name_prefix_from_env_file returns prefix when only prefix is present."""
+    # Mock the environment and spec hook
+    mock_env = mocker.Mock()
+    mock_env.name = None
+    mock_env.prefix = "/path/to/env"
+
+    mock_env_spec = mocker.Mock()
+    mock_env_spec.env = mock_env
+
+    mock_spec_hook = mocker.Mock()
+    mock_spec_hook.environment_spec.return_value = mock_env_spec
+
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        return_value=mock_spec_hook,
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name is None
+    assert prefix == "/path/to/env"
+
+
+def test_get_name_prefix_from_env_file_both_none(mocker: MockerFixture):
+    """Test get_name_prefix_from_env_file returns None, None when both are None."""
+    # Mock the environment and spec hook
+    mock_env = mocker.Mock()
+    mock_env.name = None
+    mock_env.prefix = None
+
+    mock_env_spec = mocker.Mock()
+    mock_env_spec.env = mock_env
+
+    mock_spec_hook = mocker.Mock()
+    mock_spec_hook.environment_spec.return_value = mock_env_spec
+
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        return_value=mock_spec_hook,
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name is None
+    assert prefix is None
+
+
+def test_get_name_prefix_from_env_file_exception_in_get_specifier(
+    mocker: MockerFixture,
+):
+    """Test get_name_prefix_from_env_file handles exceptions from get_environment_specifier."""
+    # Mock plugin_manager to raise an exception
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        side_effect=ValueError("Test error in get_environment_specifier"),
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name is None
+    assert prefix is None
+
+
+def test_get_name_prefix_from_env_file_exception_in_environment_spec(
+    mocker: MockerFixture,
+):
+    """Test get_name_prefix_from_env_file handles exceptions from environment_spec call."""
+    # Mock spec_hook to raise an exception when environment_spec is called
+    mock_spec_hook = mocker.Mock()
+    mock_spec_hook.environment_spec.side_effect = RuntimeError(
+        "Test error in environment_spec"
+    )
+
+    mocker.patch.object(
+        context.plugin_manager,
+        "get_environment_specifier",
+        return_value=mock_spec_hook,
+    )
+
+    name, prefix = get_name_prefix_from_env_file("test.yaml")
+
+    assert name is None
+    assert prefix is None


news/15573-conda-create-env-spec-support
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add environment specifier plugin support to `conda create --file`. Environment files (YAML, requirements.txt, explicit, etc.) are now parsed via the plugin system, matching `conda env create` behavior. Name and prefix can be read from files when `-n`/`-p` are omitted. (#15573)
```
travishathaway:
```suggestion
* conda create --file <filename> now supports environment.yaml files as well as requirements.txt and explicit environment exports and also supports automatically detecting name and prefix when --name/--prefix are omitted (#15573)
```

Just trying to put it in simpler language that most users will understand 😄

docs/source/user-guide/tasks/creating-projects.rst
```diff
@@ -63,6 +63,7 @@ Now that we have written a basic ``environment.yml`` file, we can create and act
 from it. To do so, run the following commands::
 
     conda env create --file environment.yml
+    # or: conda create --file environment.yml
```
travishathaway: I think this should be the preferred method going forward, so it would be nice to flip these around.

docs/source/user-guide/tasks/manage-environments.rst
```diff
@@ -108,6 +108,10 @@ Use the terminal for the following steps:
 
       conda env create -f environment.yml
 
+   Alternatively, ``conda create --file environment.yml`` accepts the same
```
travishathaway: Same thing here. I think we should promote `conda create --file <file>` as the recommended way to do this going forward with a note that `conda env create --file <file>` is still possible to support backwards compatibility.

conda/cli/install.py
```diff
@@ -443,6 +444,35 @@ def install(args, parser, command="install"):
 
     handle_txn(unlink_link_transaction, prefix, args, newenv)
 
+    if (
+        newenv
```
soapy1: What is the motivation for including this `newenv` check here? I would expect `conda update` or `conda install` to deal with updates to the `pip` section of an environment file. For example, I would expect the following workflow to work:

1. create a new environment.yaml
```
# environment.yaml
name: pip
dependencies:
  - pip
  - numpy
  - pip:
    - scipy
```

2. create the environment with conda
```
$ conda create -p /tmp/test --file environment.yaml
...
# environment created + pip deps installed
```

3. update the environment.yaml
```
# environment.yaml
name: pip
dependencies:
  - pip
  - numpy
  - pip:
    - scipy
    -  pandas   # new dep
```

4. update the environment with conda
```
$ conda install -p /tmp/test --file environment.yaml
. . .
```
**I would expect** the outcome of step 4 to be  that the new pip packages (pandas) are installed. However the **actual behaviour** is that these changes are ignored.

conda/cli/install.py
```diff
@@ -443,6 +444,35 @@ def install(args, parser, command="install"):
...
+        and env.external_packages
+        and not context.dry_run
+        and not context.download_only
+    ):
+        from .. import CondaError
+        from ..env.installers.base import get_installer
+        from ..env.pip_util import get_pip_workdir
```
soapy1, line 450: This is a neat way to isolate the pip installer stuff! Hopefully, we'll get to replace this soon 🌈

conda/cli/install.py
```diff
@@ -443,6 +444,35 @@ def install(args, parser, command="install"):
...
+
+        for installer_type, pkg_specs in env.external_packages.items():
+            try:
+                installer = get_installer(installer_type)
+                if installer_type == "pip":
```
soapy1: There is certainly room for improvement in the environment model's `external_packages` bit. It can maybe be removed in the future as conda starts to adopt conda-pypi's native support for pypi packages.

In the mean time, there is this [`conda.models.environment.EXTERNAL_PACKAGES_PYPI_KEY`](https://github.com/conda/conda/blob/main/conda/models/environment.py#L45) if you choose to want to use it.

```suggestion
                if installer_type == EXTERNAL_PACKAGES_PYPI_KEY:
```

docs/source/user-guide/tasks/manage-environments.rst
ryanskeith: My brain prefers `conda env` but I will change these around.

news/15573-conda-create-env-spec-support
ryanskeith: I tend to be very technical.  Thanks for edit.

conda/cli/main_create.py
```diff
@@ -96,41 +96,56 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    validate_environment_files_consistency(args.file)
+
+    if not args.name and not args.prefix and args.file:
+        name, prefix = get_name_prefix_from_env_files(args.file)
+        if name is not None:
+            args.name = name
+        if prefix is not None and args.name is None:
+            args.prefix = prefix
+        if args.name is not None or args.prefix is not None:
+            context.__init__(argparse_args=args)
```
soapy1, line 144: Whats the motivation for re-intializing the context here?

conda/cli/main_create.py
```diff
@@ -96,41 +96,56 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+            list(args.file),
+            "`--file` and `--clone` arguments are mutually exclusive.",
+        )
+
+    for fpath in args.file:
+        validate_file_exists(fpath)
+    validate_environment_files_consistency(args.file)
+
+    if not args.name and not args.prefix and args.file:
+        name, prefix = get_name_prefix_from_env_files(args.file)
```
soapy1: What is the intended behaviour for getting the right name/prefix when multiple environment files with a name or prefix is set?

Currently, if I have 2 environment files conda will choose the name/prefix from the last file. So, for the 2 environments files
```
prefix: /tmp/sdg
dependencies:
  - pip
  - numpy
  - pip:
    - scipy
```

```
name: nlp
dependencies:
  - nltk
```

If I run `conda create  --file tests/env/support/with-pip.yml --file tests/env/support/simple.yml` or `conda create --file tests/env/support/simple.yml  --file tests/env/support/with-pip.yml`, I'll get a different install location. To me, that is kind of confusing. It might be less confusing if conda produces and error requesting that the user provides a `--name`/`--prefix` arg if more than one `--file` is provided?

Some further QA here:
1. Also to confirm that if I set `--name` or `--prefix` these arguments are respected and take precedence over the name/prefix in the environment files.
2. And if I use files that do not provide a name/prefix I get an error asking to specify one. For example
```
$ conda create  --file tests/requirements-benchmarks.txt  --file tests/requirements-ci.txt  

ArgumentError: The environment file(s) do not specify a name or prefix. Please provide one via -n/--name or -p/--prefix.
```

conda/env/installers/pip.py
```diff
@@ -27,33 +28,28 @@ def _pip_install_via_requirements(prefix, specs, args, *_, **kwargs):
...
-    url_scheme = args.file.split("://", 1)[0]
-    if url_scheme in CONDA_SESSION_SCHEMES:
-        pip_workdir = None
-    else:
-        try:
-            pip_workdir = op.dirname(op.abspath(args.file))
-            if not os.access(pip_workdir, os.W_OK):
-                pip_workdir = None
-        except AttributeError:
-            pip_workdir = None
```
soapy1, line 40: hmmm, I think this is a pretty significant change in expectation for this function. Previously, it would work out what the `pip_workdir` is. But this change puts that responsibility on the caller. This might cause unintended consequences to users of the `conda.env.installer.pip.install` function who have not been notified of this change.

Would you be open to including some default value here. Like:
```
if workdir is None:
  workdir = get_pip_workdir(args.file)
```

conda/env/pip_util.py
```diff
@@ -14,11 +14,35 @@
...
+    """
+    Derive the working directory for pip install when resolving relative paths
+    in requirements (e.g. -e ./local_pkg).
+
+    Returns None for URLs or when no usable path is provided.
+    """
+    if file_path is None:
+        return None
+    if isinstance(file_path, list):
+        file_path = file_path[0] if file_path else None
```
soapy1: Hmmm, I don't think it's sufficient to pull the first file path and make that the pip workdir. The pip section of the environment.yaml is meant to support relative paths. For example, https://github.com/conda/conda/blob/main/tests/env/support/advanced-pip/env_template.yml. 

So, if I run the command:
```
$ conda create -p /tmp/six --file tests/env/support/with-pip.yml  --file  tests/env/support/advanced-pip/env_template.yml
```

conda will run the pip install with the workdir for the first file `tests/env/support/with-pip.yml` and will end up with the error:
```
Pip subprocess error:
ERROR: ./module_to_install_in_editable_mode is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).

failed
```

This is also confusing because it takes the first file path. But in `get_name_prefix_from_env_files` we take the data from the last file path.

conda/models/environment.py
```diff
@@ -296,26 +295,24 @@ def merge(cls, *environments):
         prefixes = [env.prefix for env in environments if env.prefix]
 
         if names:
-            name = names[0]
+            name = names[-1]
```
soapy1, line 320: What's the motivation for chaning this to last taking precedence over first?

conda/models/environment.py
```diff
@@ -554,6 +560,11 @@ def from_cli(
             config=EnvironmentConfig.from_context(),
         )
 
+        if envs_from_file:
+            file_env = cls.merge(*envs_from_file)
+            return cls.merge(file_env, cli_env)
```
soapy1: hmm, I'm not quite sure about the merging of config that happens here. 
This currently does not respect the precedence outlined in https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html, instead it puts the environment file as the highest precedence source of configuration. But, it should probably be at least below command line parameters (in terms of precedence). Currently this is only effecting the `channels` config, since this is the only piece of config exposed in an environment file format.

Also note, we have a similar situation with the `conda env create` command. But in that case, users can't provide a `--channels` cli argument and channels are all appended to eachother so this doesn't become such an issue.

Consider the example setup:

**condarc**
```
channels:
  - defaults
```

**simple.yml**
```
name: ex
channels:
  - conda-forge
dependencies:
  - python
  - r
```

## case 1: run a normal conda install
**expect**: 
  1. to get python installed from the conda-forge channel
  2. to get r installed from the conda-forge channel

**actual**: 
  1. ✔️ I get python installed from conda-forge
  2.  ✔️ I get r installed from the conda-forge channel
```
$ conda create -p /tmp/eight --file simple.yml  --dry-run  
Channels:
 - conda-forge
 - defaults
 . . .
  python             conda-forge/linux-64::python-3.14.3-h32b2ec7_101_cp314 
  r                  conda-forge/noarch::r-4.5-r45hd8ed1ab_1009 
. . .
```

## case 2: run a conda install with a `--channel r`
**expect**: 
1. to get python installed from the conda-forge channel
2. to get r installed from the r channel
3.  have the r channel be the highest priority

**actual**: 
1. ✔️ I get python installed from conda-forge,
2. ❌ I get r installed from the conda-forge channel
3. ❌ I get conda-forge as the highest priority channel
```
$ conda create -p /tmp/eight --file simple.yml --channel r  --dry-run  
Channels:
 - conda-forge
 - r
 - defaults
 . . .
python             conda-forge/linux-64::python-3.14.3-h32b2ec7_101_cp314 
r                  conda-forge/noarch::r-4.5-r45hd8ed1ab_1009 ```
. . .
```

ref: https://github.com/conda/conda/pull/15199

conda/models/environment.py
```diff
@@ -577,7 +588,7 @@ def extrapolate(self, platform: str) -> Environment:
             with repodata_manager as repodata_fn:
                 solver = solver_backend(
                     prefix="/env/does/not/exist",
-                    channels=context.channels,
+                    channels=self.config.channels,
```
soapy1, line 646: I'm not too familar with this `extrapolate` function. What's the motivation for changing this? If this is a required change for some other bug can we do it in a seperate PR to keep this one focused on just adding env spec support to `conda create`.

conda/cli/main_create.py
```diff
@@ -96,41 +96,56 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+            len(args.file),
+            list(args.file),
+            "`--file` and `--clone` arguments are mutually exclusive.",
+        )
+
+    for fpath in args.file:
+        validate_file_exists(fpath)
+    validate_environment_files_consistency(args.file)
+
+    if not args.name and not args.prefix and args.file:
```
soapy1: Is this logic also now applicable to `conda install`?

For example, if I have an environment.yaml
```
prefix: /tmp/myenv
dependencies:
  - numpy
``` 

And I run a `conda install --file environment.yaml` should I expect it to create an environment `/tmp/myenv`? Or should it try to install the specified dependencies into the current environment?

soapy1 (dismissed):
Looking great so far! I have a few questions around what the intended behaviour of some edge cases should be.

It would also be nice to add a few more tests. It might be neat to parameterize the integration tests in https://github.com/conda/conda/blob/main/tests/env/test_create.py to run against `conda env create` and `conda create` (I'm just spitballing there).

conda/cli/install.py
ryanskeith: Ah, hyper focused on the `create` part of the ticket.  Totally correct on pointing it out for the other functionality.

conda/cli/install.py
ryanskeith: I didn't like how I needed to do this.  I felt that the better alternative would be a refactor of the env.installers workflow.  It would be interesting to see if we allowed the installer to know what to do.  It seemed like too much change to have it scale better.  Might be worth an exploratory demo.

conda/cli/main_create.py
ryanskeith: The motivation is the same as the --dry-run.  We need to reinit the context if the code changes -name or -prefix. In the dry-run case, a temp path is supplied. 

At this point the determine_target_prefix has been run in the main.py code.  If name or prefix is not set, the target prefix is the base environment.  If one of them is set it is the normal behavior at the command line the target_prefix has already been set.  This case captures the case where an environment file has that information and the users did not supply on the cli.

conda/env/installers/pip.py
ryanskeith: That's a great idea.  One of my worries was if a deprecation notice was needed.  I like this idea.  I will include it.

conda/models/environment.py
ryanskeith: Other parts of the merge code overwrote (configs and variables were last key wins).  I thought name and prefix should be the same as well for consistency.

conda/models/environment.py
ryanskeith: Ah, I was running some code locally and saw that that my local .condarc was affecting the tests.  After consulting another conda team member and determined it was a bug in the code I made this small change.  I do understand that a separate PR is another option BUT this was also after recently reading the HOW WE USE GITHUB guide.

https://github.com/conda/conda/blob/main/HOW_WE_USE_GITHUB.md#should-we-make-unrelated-changes-at-the-same-time

This states it is ok to make small changes to unrelated work in PR's.  So I went ahead and included it and moved on.

conda/models/environment.py
ryanskeith: You are totally right that the CLI should take precedence and that is the intention of the code.  I changed the code so that the last should take precedence.  It seems the intention of the EnvironmengtConfig class is the same. 

https://github.com/conda/conda/blob/3c53f03c80c3645f22e6ec69789a3902308b7396/conda/models/environment.py#L105-L115

I will look more into this to understand why it is not doing this.

conda/env/pip_util.py
ryanskeith: I was thinking more of the case where we have all the requirements.txt in the same directory (like conda).  

You are bringing up a case where we need to do multiple pip installs.  Let's say that I have multiple files in different locations and they all have relatives paths taking one of these files will not lead to a good install.  It seems the code needs to be smarter if we allow for multiple files.

conda/cli/install.py
ryanskeith: For now, I feel we can hold off on this change as we are thinking about how we want external packages to be handled.

conda/cli/main_create.py
ryanskeith: That is correct.  So you are saying that if one is supplied we are ok and just use it but if both are then ambiguity exists and we should ask the user for help.  As a power user, I do like the idea that I can make use priority to my advantage and save typing.  I do see your point that this is a new case and the user should be explicit here.

I am honestly torn on this issue.

conda/env/pip_util.py
ryanskeith: So after reading the CEP on requirements.txt, conda expects requirements.txt to have their paths based on the cwd.

Moving forward with CEP 24 and handling multiple pip sections, the requirement is different.  We will need to loop through these for now as the expected behavior is for pip to run based off the environment file.

conda/cli/main_create.py
ryanskeith: That's a great question. 

As a user, I think it would make the most sense to me that it would install in the current environment.  conda install is first and foremost a package command (not environment).  At least in my mind.

conda/cli/main_create.py
```diff
@@ -96,41 +96,61 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+            "`--file` and `--clone` arguments are mutually exclusive.",
+        )
+
+    for fpath in args.file:
+        validate_file_exists(fpath)
+    validate_environment_files_consistency(args.file)
+
     if not args.name and not args.prefix:
-        if context.dry_run or context.download_only:
+        if args.file and len(args.file) > 1:
```
ryanskeith, line 129: I got some feedback that for multiple files that we should always ask the user for a name or prefix to keep things simple.  For the single case, we will look for them.

conda/env/pip_util.py
ryanskeith: Resolved by removing list behavior altogether in this function.  Since we are looping by file now in `conda create` we will call this over and over again if need be.

conda/cli/main_create.py
soapy1: Looks like you went with asking the user to provide a -n/-p if there are multiple files that define an install location. That's great!

I can also confirm that this doesn't change the behviour for users running commands like
```
$ conda create -p /something --file requirements.txt --file requirements2.txt

$ conda install -p /something --file requirements.txt --file requirements2.txt
``` 

@ryanskeith could you also add a few tests for this new functionality :pray:

conda/env/pip_util.py
soapy1: Nice solution! I think not ideal to have to call pip multiple times. But when I was testing it out it didn't feel out of place or anything. So I think this works well.

conda/models/environment.py
soapy1: hmmm, I think this still isn't quite right now I get the follwoing behaviour:

## case 1: run a normal conda install
**expect**: 
  1. to get python installed from the conda-forge channel
  2. to get r installed from the conda-forge channel
  3. conda-forge is the highest priority channel

**actual**: 
  1.  ❌ I get python installed from defaults
  2.  ❌ I get r installed from the defaults channel
  3.  ❌ defaults is the highest priority channel

```
$ conda create -p /tmp/eight --file simple.yml  --dry-run  
Channels:
 - defaults
 - conda-forge
 . . .
  python             pkgs/main/linux-64::python-3.14.3-h490e9c7_101_cp314  
 r                  pkgs/r/linux-64::r-4.3.1-r43_0 . . .
```

## case 2: run a conda install with a `--channel r`
**expect**: 
1. to get python installed from the conda-forge channel
2. to get r installed from the r channel
3.  have the r channel be the highest priority

**actual**: 
1. ❌ I get python installed from defaults,
2. ✔️ I get r installed from the r channel
3. ✔️ I get r as the highest priority channel
```
$ conda create -p /tmp/eight --file simple.yml --channel r  --dry-run  
Channels:
 - r
 - defaults
 - conda-forge
 . . .
  python             pkgs/main/linux-64::python-3.14.3-h490e9c7_101_cp314 
  r                  r/linux-64::r-4.3.1-r43_0 
. . .
```

conda/cli/main_create.py
soapy1: That sounds sensible to me :+1:

conda/cli/install.py
```diff
@@ -443,6 +444,36 @@ def install(args, parser, command="install"):
...
+                    else:
+                        installer.install(prefix, list(pkg_specs), args, file_env)
+                except InvalidInstaller:
+                    raise CondaError(
+                        f"Unable to install package for {installer_type} from environment file {fpath}. "
+                        "Please ensure your dependencies file has the correct spelling."
+                    )
+
+        if env.variables:
+            PrefixData(prefix).set_environment_env_vars(env.variables)
```
soapy1:
```suggestion
    if env.variables :
        PrefixData(prefix).set_environment_env_vars(env.variables)
```

Looks like this is indented a tiny bit too much.  Currently env vars don't get exported. For example a env.yaml that looks like
```
variables:
  VARIABLE: value
  VARIABLE2: value2
dependencies: []
```

conda/models/environment.py
soapy1: Cool, this seems to work now! I also did some further testing with overriding channels.

Consider the environment file
```
# simple.yml
name: ex
dependencies:
  - python
  - r
```

## case 1 run a create

**expect**
* packages to come from the default channel
* only the default channel is used

**actual**
* :heavy_check_mark: packages from the default channel
* :heavy_check_mark:  only the default channel is used

```
 $ conda create -p /tmp/eight --file simple.yml  --dry-run

Channels:
 - defaults
Platform: linux-64
. . .
```

## case 2 use conda-forge channel
**expect**
* packages to come from the conda-forge channel primarily
* the conda-forge and default channels are used

**actual**
* :heavy_check_mark: packages from the conda-forge channel
* :heavy_check_mark:   the conda-forge and default channels are used

```
$ conda create -p /tmp/eight --file simple.yml  --dry-run   --channel conda-forge                    
Channels:
 - conda-forge
 - defaults
Platform: linux-64
 . . .
```


## case 3 override channels

**expect**
* packages to come from the conda-forge channel
* only the conda-forge channel is used

**actual**
* :heavy_check_mark: packages from the conda-forge channel
* :heavy_check_mark:  only the conda-forge channel is used

```
 $ conda create -p /tmp/eight --file simple.yml  --dry-run   --channel conda-forge --override-channels

Channels:
 - conda-forge
 Platform: linux-64
. . .
```

conda/models/environment.py
```diff
@@ -487,7 +507,7 @@ def from_cli(
         cls,
         args: Namespace,
         add_default_packages: bool = False,
-    ) -> Environment:
+    ) -> tuple[Environment, dict[str, Environment]]:
```
soapy1: I think this api change is pretty significant. How would you feel about keeping `from_cli` the same as it was and intorduce a new function that returns this tuple of `[Environment, dict[str, Environment]]`?

conda/models/environment.py
```diff
@@ -533,27 +553,55 @@ def from_cli(
...
 
+        if envs_from_file:
+            file_env = cls.merge(*envs_from_file)
+            merged = cls.merge(base_env, file_env, cli_env)
+
+        else:
+            merged = cls.merge(base_env, cli_env)
+
+        # Respect override_channels flag and only use channels from the CLI
+        if getattr(args, "override_channels", False):
```
soapy1, line 617: I'm not a super big fan of this special management of configured channels (or the environemnt config merging) situation in general. It would be better if all the logic for merging configurations from all sources happened in the context object as suggested in https://github.com/conda/conda-planning/issues/88. I think that's a pretty big piece of work and I wouldn't want to explode the scope of this PR. 
Would a reasonable compromise be to create a to do item/issue for addressing this?

conda/models/environment.py
ryanskeith: I felt that because the internal code only calls it once, it was safe to make this change. I will look at what it would look like to make a new function.

conda/models/environment.py
ryanskeith: I agree it would be very reasonable.  I will make todo note about channel merging happening in a central place.  Perhaps we could also find some time before Q3 planning to chart out what that would look like for next steps.

conda/models/environment.py
ryanskeith: After looking at it I realize there is a simple solution.  I will keep from_cli the same and it is a thin wrapper around a new function called from_cli_with_file_envs.

docs/source/dev-guide/plugins/environment_specifiers.rst
```diff
@@ -46,7 +46,9 @@ Be sure to be very specific when implementing the ``can_handle`` method. It shou
 return a ``True`` if the file can be parsed by the plugin. Making the ``can_handle``
 method too permissive in the types of files it handles may lead to conflicts with other
 plugins. If multiple installed plugins are able to ``can_handle`` the same file type,
-conda will return an error to the user.
+conda will return an error to the user.  Also, assure that the ``can_handle`` method
```
travishathaway:
```suggestion
conda will return an error to the user.  Also, ensure that the ``can_handle`` method
```

travishathaway:
@soapy1,

I think it would make sense for you to take another quick look at the pull request. You can specifically focus on the changes I introduced in my last commit. Below is an explanation of what/why I changed things.

### Fixing `prefix` defaults

While I was going through and QA testing this myself, I discovered that this did not produce the expected error:

> When -n/-p are omitted, read name/prefix from files via get_name_prefix_from_env_files 

I was testing it with the command:

```
conda create --file env.yaml
```

Instead, it was giving the following error:

```
CondaValueError: 'base' is a reserved environment name
```

This was happening because of this line:

https://github.com/conda/conda/blob/aa6e9eca06d392ef8e6408369aa7400d5b9a031e/conda/env/env.py#L442

So, to fix this, I have simply removed falling back to `context.target_prefix`. I think this is the right behavior because supplying that as a fall back should be done outside of the object's instantiation (e.g. be provided directly as the `prefix` argument to the `EnvironmentYaml` object itself).

### Command line parsing

The other changes I made were related to command line parsing. I also decided to change the `get_name_prefix_from_env_files` function to just work on single env files because in the `main_create.py` file in the `execute` function, we're not actually passing in multiple env files to determine name/prefix (we raise an error before getting to that point).

PR #15774, opened 2026-03-10
Improves `conda search` documentation

schuylermartin45:
- This came out of some frustrations I had recently. As a user, I could not easily figure out how to find a package at a specific package or python version.

### Checklist - did you ...

### DIFF
--- a/conda/testing/fixtures.py
+++ b/conda/testing/fixtures.py
@@ -589,8 +589,16 @@ class TmpChannelFixture:
         noarch = channel / "noarch"
         noarch.mkdir(parents=True)
 
-        noarch_packages: dict[str, PackageRecord] = {}
-        subdir_packages: dict[str, PackageRecord] = {}
+        noarch_packages: dict[str, dict[str, object]] = {
+            "info": {"subdir": "noarch"},
+            "packages": {},
+            "packages.conda": {},
+        }
+        subdir_packages: dict[str, dict[str, object]] = {
+            "info": {"subdir": subdir.name},
+            "packages": {},
+            "packages.conda": {},
+        }
         iter_specs = list(specs)
         seen: dict[str, set[str]] = {}
         while iter_specs:
@@ -619,7 +627,8 @@ class TmpChannelFixture:
                 copyfile(source, target / fname)
 
                 # add package to repodata
-                packages[fname] = PackageRecord(
+                key = "packages" if fname.endswith(".tar.bz2") else "packages.conda"
+                packages[key][fname] = PackageRecord(
                     **{
                         field: value
                         for field, value in package_record.dump().items()
@@ -629,12 +638,8 @@ class TmpChannelFixture:
 
                 iter_specs.extend(package_record.depends)
 
-        (subdir / "repodata.json").write_text(
-            json.dumps({"info": {}, "packages": subdir_packages})
-        )
-        (noarch / "repodata.json").write_text(
-            json.dumps({"info": {}, "packages": noarch_packages})
-        )
+        (subdir / "repodata.json").write_text(json.dumps(subdir_packages))
+        (noarch / "repodata.json").write_text(json.dumps(noarch_packages))
 
         # ensure all packages were copied to the channel
         for spec in chain.from_iterable(seen.values()):
--- a/conda/testing/helpers.py
+++ b/conda/testing/helpers.py
@@ -281,10 +281,12 @@ def _get_index_r_base(
     channels = []
     for subchannel, subchannel_pkgs in packages.items():
         repodata = {
+            "repodata_version": 2,
             "info": {
                 "subdir": subchannel,
                 "arch": context.arch_name,
                 "platform": context.platform,
+                "base_url": f"https://conda.anaconda.org/{channel_name}/{subchannel}",
             },
             "packages": subchannel_pkgs,
         }
@@ -459,6 +461,7 @@ def record(
     channel=None,
     **kwargs,
 ):
+    kwargs.setdefault("fn", f"{name}-{version}-{build}.conda")
     return PackageRecord(
         name=name,
         version=version,
--- a/conda/testing/solver_helpers.py
+++ b/conda/testing/solver_helpers.py
@@ -118,7 +118,7 @@ class SimpleEnvironment:
     def _package_data(self, record):
         """Turn record into data, to be written in the JSON environment/repo files."""
         data = {
-            key: value
+            key: (",".join(value) if key in ("features", "track_features") else value)
             for key, value in vars(record).items()
             if key in self.REPO_DATA_KEYS
         }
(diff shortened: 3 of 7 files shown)


schuylermartin45:
I may make a sister PR shortly to `conda-lock` with the same documentation. They appear to use the same doc string in that project.

schuylermartin45:
Equivalent PR for `conda-lock`: https://github.com/conda/conda-lock/pull/899

dholth:
Here are some different [MatchSpec query language](https://github.com/conda/ceps/blob/main/cep-0029.md) strings that might be preferable.
```bash
% conda search "numpy-base 2.4.2 py313*"
Loading channels: done
# Name                       Version           Build  Channel             
numpy-base                     2.4.2 py313h23175f9_0  pkgs/main           
% conda search "numpy-base==2.4.2 py313*"
Loading channels: done
# Name                       Version           Build  Channel             
numpy-base                     2.4.2 py313h23175f9_0  pkgs/main           
% conda search "numpy-base==2.4.2[build='py313*']"
Loading channels: done
# Name                       Version           Build  Channel             
numpy-base                     2.4.2 py313h23175f9_0  pkgs/main           
% conda search "numpy-base==2.4.2"                
Loading channels: done
# Name                       Version           Build  Channel             
numpy-base                     2.4.2 py311h23175f9_0  pkgs/main           
numpy-base                     2.4.2 py312h23175f9_0  pkgs/main           
numpy-base                     2.4.2 py313h23175f9_0  pkgs/main           
numpy-base                     2.4.2 py314hb4b2851_0  pkgs/main           
% conda search "numpy-base[build='py313*']==2.4.2"
Loading channels: done
# Name                       Version           Build  Channel             
numpy-base                     2.4.2 py313h23175f9_0  pkgs/main           
```

news/update-conda-search-docs
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* Expands `conda search` documentation to include more search examples.
```
jaimergp:
```suggestion
* Expands `conda search` documentation to include more search examples. (#15774)
```

conda/cli/main_search.py
```diff
@@ -68,7 +68,15 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
 
             conda search conda-forge::numpy
+
+        Search for a package on a specific channel and platform::
+
             conda search 'numpy[channel=conda-forge, subdir=osx-64]'
+
+        Search for a package with a specific software version and python version::
+
+            conda search 'numpy-base[version="==2.4.2",build=py313*]'
```
jaimergp:
```suggestion
            conda search 'numpy-base[version="2.4.2",build=py313*]'
```

Does this mean that the above is not equivalent? Or `numpy-base==2.4.2=py313*`?

schuylermartin45: I saw things like `numpy-base==2.4.2=py313*` get suggested in forums online but I did not get that to work. Maybe I used a `==` for an `=` somewhere along the way. I find your suggestion easier to read/more consistent with the existing docs, personally.

schuylermartin45: `numpy-base==2.4.2=py313*` does work, I just confirmed. I probably entered it in as `numpy-base=2.4.2=py313*` yesterday.

dholth: I agree that `(comparison operator)version=build` is the preferable form, and `[]` syntax is for other keys that don't fit in the older `name version build` form.

schuylermartin45:
> Here are some different [MatchSpec query language](https://github.com/conda/ceps/blob/main/cep-0029.md) strings that might be preferable.

Should we reference that CEP in the help menu docs? SEO pointed me to a bunch of old StackOverflow posts when I tried to find more advanced options for `MatchSpec`.

conda/cli/main_search.py
schuylermartin45: I suppose I can list both. Let me go do that now.

schuylermartin45: Added.

conda/cli/main_search.py
```diff
@@ -68,7 +68,16 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         Search for a package on a specific channel::
 
             conda search conda-forge::numpy
+
+        Search for a package on a specific channel and platform::
+
             conda search 'numpy[channel=conda-forge, subdir=osx-64]'
+
+        Search for a package with a specific software version and python version::
```
dholth, line 76:
```suggestion
        Search for a package with a specific software version and build string::
```

schuylermartin45: Before I take this change, can we make this more clear/searchable? I think "python version" is a more searchable/what I imagine what most folks would look for, even if it is less technically correct.

travishathaway: Maybe a good compromise would be to have one example with "build" spelled out and another with "python"?

Maybe as the other example, you could use an "r" package?

```
conda search 'r-shiny==1.12.0=r45*'
```

Is there any other ways we semantically overload that field? 😅

schuylermartin45: I like that idea, let me make a quick edit and see what you all think.

schuylermartin45: Done

conda/cli/main_search.py
```diff
@@ -68,7 +68,21 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+
+            conda search 'numpy-base==2.4.2=py313*'
```
kenodegard: The double equals is unnecessary:

```
conda search 'numpy-base=2.4.2=py313*'
conda search 'numpy-base=2.*=py313*'
```

schuylermartin45: Confirmed locally, changing now.

dholth: According to https://github.com/conda/ceps/blob/main/cep-0029.md#version-matching the single-equals is fuzzy equality, and the double-equals is exact equality. Most `str(MatchSpec(...))` examples render out including double-equals.
IIUC an `=1` matches any version where the first component is `1` e.g. `1.*`

kenodegard: Yes, single equals is fuzzy but this only applies to select patterns: https://github.com/conda/ceps/blob/main/cep-0029.md#version-expression-parsing

E.g. only the following patterns are implicitly fuzzy:

```
pkg=1.8
pkg =1.8
```

Once we add a build string we need to be explicit with our fuzziness:

```
pkg 1.8.*
pkg 1.8.* *
pkg=1.8.*
pkg=1.8.*=*
pkg =1.8.* *
pkg ==1.8.* *
pkg[version=1.8.*]
pkg[version="1.8.*"]
```

dholth: Such MatchSpec

schuylermartin45:
Are there any further objections on this change? I was hoping this would be a quick win for users. I apologize for any `MatchSpec` related headaches this has caused.

schuylermartin45:
@travishathaway Thanks for the assist!

PR #14661, opened 2025-03-11
Allow `custom_mutlichannels` to override `default_channels`

kenodegard:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #15776, opened 2026-03-12
Fix: custom_multichannel strict channel priority

### LINKED ISSUE #15750: Regression in channel priority behavior between v25.11.1 and v26.1.0
### Checklist

### What happened?

It appears that the behavior of strict channel priority has changed between conda v25.11.1 and v26.1.0.

In v25.11.1, strict channel priority seemed to apply only at the top level when selecting between channels and multichannels, while still allowing flexible dependency resolution within the channels that make up a given multichannel. This is the behavior that our team relied upon.

In conda v26.1.0, this behavior appears to have changed. Strict channel priority now seems to apply more rigidly, preventing fallback to compatible packages from other channels within a multichannel configuration. As a result, resolution outcomes differ and, in our case, fail.

Here is a test case I used to reproduce:

I ran this from a linux-64 machine with base conda 25.1.1 and git installed in base.

```
/opt/conda/bin/git clone --depth 1 https://github.com/AnacondaRecipes/aggregate.git
cd aggregate
/opt/conda/bin/git clone https://github.com/AnacondaRecipes/holidays-ext-feedstock.git

/opt/conda/bin/conda config --add custom_multichannels.src_multichannel conda-forge
/opt/conda/bin/conda config --add custom_multichannels.src_multichannel https://repo.anaconda.com/pkgs/main
/opt/conda/bin/conda config --add custom_multichannels.src_multichannel marekw
/opt/conda/bin/conda config --set channel_priority strict

/opt/conda/bin/conda create --prefix tmp/test_conda_25 conda==25.11.1 conda-build==25.11.1
/opt/conda/bin/conda run --prefix tmp/test_conda_25 conda-build holidays-ext-feedstock -c src_multichannel --override-channels
```

This passes :white_check_mark:

Then:

```
/opt/conda/bin/conda create --prefix tmp/test_conda_26 conda==26.1.0 conda-build==26.1.0
/opt/conda/bin/conda run --prefix tmp/test_conda_26 conda-build holidays-ext-feedstock -c pbp_src_multichannel --override-channels
```

This fails :x:

- holidays-ext-feedstock requires `holidays >=0.9.5` but marekw channel that is placed in front contains v0.48.
- conda-forge is needed to supply lunarcalendar-ext that is not available in main.

conda 25.11.1 seems to be able to pick holidays >=0.9.5 from the main channel that is placed 2nd in the multichannel, while conda 26.1.0 attempts to pick 0.48 from marekw channel that is placed 1st. As a result, the solver breaks. I think this is due to strict channel priority because it passes with flexible.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

kenodegard:
### Description

Closes https://github.com/conda/conda/issues/15750

### Checklist - did you ...


kenodegard:
Hm I'm expecting this new test to fail as seen when run manually (see https://github.com/conda/conda/issues/15750#issuecomment-4049589580)

But the test passes and works correctly?

**Edit:** found that `context.custom_multichannels` is a `memoizedproperty` which does not behave as expected, resorted to `context._override`

.github/workflows/tests.yml
```diff
@@ -32,6 +32,93 @@ concurrency:
...
+          "runs-on": "macos-latest",
+          "python-version": "3.14",
+          "default-channel": "defaults",
+          "test-type": "integration",
+          "test-group": 1
+        }
+      ]
+    }
+  # TIP: hardcode selectors for focused CI debugging, but remember to revert before merging to main
+  PYTEST_SUBSET: tests/core/test_solve.py::test_strict_custom_multichannel_allows_fallback_to_later_subchannel
```
kenodegard: The changes in this PR to `tests.yml` is only to allow us to selectively run the CI for only a subset of tests

kenodegard:
Confirmed by reverting that https://github.com/conda/conda/pull/14661 is the culprit

Unsure why

kenodegard:
The issue is caused by a reverse lookup collision since the `pkgs/main` URL is shared between `defaults` and the custom multichannel.

tests/core/test_solve.py
```diff
@@ -2840,6 +2841,42 @@ def test_channel_priority_churn_minimized(tmpdir):
...
+    found in the latter channel in the custom multichannel.
+    """
+    if context.solver == "classic":
+        pytest.skip(
+            "classic solver does not support flexible priority within custom_multichannels"
+        )
+    with tmp_channel() as (channel_path, channel_path_str):
+        multichannel = (channel_path_str, "https://repo.anaconda.com/pkgs/main")
+        with context._override("_custom_multichannels", {"custom": multichannel}):
+            reset_context()
```
jaimergp: I don't know if we need to reset here because; `_override` works at the instance attribute level directly.

tests/test_priority.py
```diff
@@ -42,7 +43,7 @@ def test_reorder_channel_priority(
 
     # set pinned package
     if pinned_package:
-        monkeypatch.setenv("CONDA_PINNED_PACKAGES", package1)
+        monkeypatch.setenv("CONDA_PINNED_PACKAGES", f"{package1}=1")
```
jaimergp, line 46: Note these two expression mean something different.

- A pin with `package-name` (no version) means locking the installed package and disallowing any change (e.g. if I have `numpy=1.2.3=h123abc` and I pin `numpy` it will always stay installed as `1.2.3=h123abc`)
- A pin with `package-name=version` will act as a constraint. For `numpy=1.2.3=h123abc`, a pin of `numpy=1.2` means that the solver can freely choose packages in that match that matchspec.

conda/base/context.py
```diff
@@ -965,14 +965,21 @@ def custom_multichannels(self) -> dict[str, tuple[Channel, ...]]:
         else:
             default_channels = list(self._default_channels)
 
+        # Ensure that when "defaults" is present in custom multichannels, it overrides
+        # normal defaults.
```
ForgottenProgramme: what is meant by "normal defaults"?

tests/test_priority.py
travishathaway: Thanks for the clarification!

The reason I made this change is that I was seeing some weird behavior with our local channel. If I just leave the pin the way it is, and there is no cache present, then when I request for `versioned=1`, I actually end up with `versioned=2` 🤯.

Do you think this is a bug worth filing for separately? This only happens with a completely clean cache and restricting the pin in that way solves the bug.

Here's the install plan from my debugging:

```
Solving environment: ...working... done

## Package Plan ##

  environment location: /private/var/folders/f0/45_t0scx2qn0dfmqyktxrs040000gn/T/pytest-of-travishathaway/pytest-353/test_reorder_channel_priority_0/ef8d720c2abb

  added / updated specs:
    - dependent=1
    - versioned=1


    dependency-1.0             |                0           3 KB  file:///Users/travishathaway/dev/conda/tests/data/test-recipes
    dependent-1.0              |                0           3 KB  file:///Users/travishathaway/dev/conda/tests/data/test-recipes
    versioned-2.0              |                0           4 KB  file:///Users/travishathaway/dev/conda/tests/data/test-recipes
                                           Total:           9 KB

  dependency         test-recipes/noarch::dependency-1.0-0
  dependent          test-recipes/noarch::dependent-1.0-0
  versioned          test-recipes/noarch::versioned-2.0-0
```

conda/base/context.py
travishathaway: "normal defaults" is the usual multichannel that is created when a user specifies "defaults" in their conda configuration. Anaconda has historically done this to combine both their "main" and "r" channels together.

Do you think this comment could be worded differently to be clearer? Do you have any suggestions yourself?

tests/core/test_solve.py
travishathaway: Will remove 🫡

conda/base/context.py
jezdez:
```suggestion
        # the built-in `default_channels` list (which combines channels like "main" and
        # "r" into a single "defaults" multichannel).
```

jezdez:
Regarding Jaime's comment on the pin semantics change -- I agree the workaround is fine for unblocking this PR, but the underlying issue Travis found (bare pin with clean cache resolves to the wrong version) seems worth filing as a separate bug.

jezdez (approved):

travishathaway:
@jezdez, I'll fill this as a separate bug with steps to reproduce.

travishathaway:
xref: https://github.com/conda/conda/issues/15788

hmaarrfk:
I think i might be hitting this issue and its causing some pretty weird solves. 

Excited to try it.

PR #15782, opened 2026-03-17
Use default python setting

soapy1:
### Description

The `./dev/start` script has a default python defined, as suggested in the help
```
(base) sophia:conda/ (default-python-dev-env) $ source ./dev/start --help                                                                                                                                              Usage: source ./dev/start [options]

Options:
  -p, --python    VERSION    Python version for the env to activate. (default: 3.10)
```

This is pulled from the `_DEFAULT_PYTHON` variable. We should fall back to this value instead of hardcoding it here.

### Checklist - did you ...

- ~[ ] Add a file to the `news` directory ([using the template]~(https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
### DIFF
--- a/dev/start
+++ b/dev/start
@@ -168,7 +168,7 @@ if [ -z "${_INSTALLER_TYPE}" ]; then
 fi
 
 # fallback to default values if not set
-_PYTHON="${_PYTHON:-3.10}"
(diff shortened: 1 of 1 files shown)


travishathaway:
Thanks!

PR #15792, opened 2026-03-18
Raise an error if plugin can not handle a file

soapy1:
### Description

Previously environment spec plugins would retrun `False` if they were unable to handle the provided file. This PR:
* updates the bulit in plugins to raise an error instead of returning `False`
* updates the error returned when selecting an environment specifier by name to be `EnvironmentSpecPluginSelectionError` - include information about what errors happened in the plugin + list all other available plugins.

### Test it out - requirements file

Consider a valid requirements.txt style environment file:
```
numpy
python=3.14
```

#### We can confirm env file detection and env creation works

```
 $ conda env create -p /tmp/test --file reqs --dry-run
Retrieving notices: done
Channels:
 - defaults
Platform: linux-64
Solving environment: done
. . .
```

#### Using the env.yml plugin does not work

```
$ conda env create -p /tmp/test --file reqs --env-spec cep-24

EnvironmentSpecPluginSelectionError: Could not parse 'reqs' as 'cep-24'. Check that the file contents match the expected format. Errors reported from 'cep-24':

    ->  TypeError: string indices must be integers

If this environment is a different format, try:
    conda create --file reqs --format <format>

Available formats: 
    - cep-24 (environment-yaml, env.yml)
    - environment.yml
    - explicit
    - requirements.txt (requirements, reqs)
```

### Test it out - environment.yaml file

Consider a valid environment.yaml style environment file:
```
name: nlp
dependencies:
  - nltk
```

#### We can confirm env file detection and env creation works

```
$ conda env create --file environment.yml --dry-run
Channels:
 - defaults
Platform: linux-64
Solving environment: done
```

#### Using the requirements/explicit plugin does not work

```
$ conda env create --file environment.yml --env-spec reqs

EnvironmentSpecPluginSelectionError: Could not parse 'environment.yml' as 'reqs'. Check that the file contents match the expected format. Errors reported from 'reqs':

    ->  PluginError: environment.yml contains an invalid matchspec 'name: nlp'

If this environment is a different format, try:
    conda create --file environment.yml --format <format>

Available formats: 
    - cep-24 (environment-yaml, env.yml)
    - environment.yml
    - explicit
    - requirements.txt (requirements, reqs)

$ conda env create --file environment.yml --env-spec explicit

EnvironmentSpecPluginSelectionError: Could not parse 'environment.yml' as 'explicit'. Check that the file contents match the expected format. Errors reported from 'explicit':

    ->  PluginError: environment.yml is not an explicit file. Missing the explicit marker.

If this environment is a different format, try:
    conda create --file environment.yml --format <format>

Available formats: 
    - cep-24 (environment-yaml, env.yml)
    - environment.yml
    - explicit
    - requirements.txt (requirements, reqs)
```

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1349,7 +1349,7 @@ class EnvironmentSpecPluginSelectionError(CondaError):
         **kwargs,
     ):
         plugin_names = [
-            f"{name}{' (' if plugin.aliases else ''}{', '.join(plugin.aliases)}{')' if plugin.aliases else ''}"
+            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
             for name, plugin in plugin_specs.items()
         ]
         msg += f"\nAvailable formats: {dashlist(plugin_names, 4)}"


conda/env/specs/cep_24.py
```diff
@@ -32,21 +33,21 @@ def can_handle(self):
...
-        except CondaError:
-            log.debug("Failed to load %s as a YAML.", self.filename, exc_info=True)
-            return False
+            raise CondaValueError("No filename provided")
+
+        yamlstr = env.load_file(self.filename)
+        data = yaml.loads(yamlstr)
+        errors = env.get_schema_errors(data)
+        if errors:
+            raise PluginError(
```
soapy1, line 46: This change makes it so that this plugin raises an error instead of returning a `False` value if the plugin can not handle the file. Is this an ok change to make without going thru the deprecation cycle?

conda/env/specs/cep_24.py
```diff
@@ -32,21 +33,21 @@ def can_handle(self):
...
+                "Provided environment file is invalid:"
+                f"{dashlist([err for err in errors], indent=4)}"
```
kenodegard: isn't errors already a list?

conda/env/specs/explicit.py
```diff
@@ -38,14 +38,16 @@ def can_handle(self) -> bool:
         """
         # Return early if no filename was provided
         if self.filename is None:
```
kenodegard, line 40: Is there a reason why this tests for `is None` but cep_24/yaml_file tests for falsy?

conda/env/specs/requirements.py
```diff
@@ -43,26 +43,28 @@ def can_handle(self) -> bool:
         """
         # Return early if no filename was provided
         if self.filename is None:
```
kenodegard, line 45: ditto

kenodegard (approved):

conda/env/specs/cep_24.py
soapy1: true!

conda/env/specs/explicit.py
soapy1: I've updated these plugins to be more consistent with this. 

I guess also good to note that it would be nice if the validation steps between the plugins were a bit more aligned. For example, the requirements plugin needs a local file, but the cep-24 plugin allows the file to be remote. Conda should be dealing with all of those details probably. I think that is for another chunck of work though.

soapy1:
pre-commit.ci autofix

conda/env/specs/yaml_file.py
```diff
@@ -43,18 +43,14 @@ def can_handle(self):
         :return: True or False
         """
         if not self.filename:
```
kenodegard:
```suggestion
        if self.filename is None:
```

soapy1:
pre-commit.ci autofix

conda/exceptions.py
```diff
@@ -1341,6 +1340,22 @@ def __init__(self, msg: str, *args, **kwargs):
...
+class EnvironmentSpecPluginSelectionError(CondaError):
+    def __init__(
+        self,
+        msg: str,
+        plugin_specs: dict[str, CondaEnvironmentSpecifier],
+        *args,
+        **kwargs,
+    ):
+        plugin_names = [
+            f"{name}{' (' if plugin.aliases else ''}{', '.join(plugin.aliases)}{')' if plugin.aliases else ''}"
```
kenodegard: The double ifs here has tripped me up multiple times now so lets simplify it like so:

```suggestion
            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
```

PR #15793, opened 2026-03-19
Improve env spec plugin detection error messages

soapy1:
### Description

This PR aims to improve the error messaging around environment spec plugin detection by removing jargon error messages, providing actionable steps and surfacing useful information about the state of the system.

There are 2 main ways that env spec plugin detection fails:
1. The environment file is not detected
2. Too many environment spec plugins can read the file

#### 1. The environment file is not detected

In this type of failure, conda has determined that there are no plugins that are able to read the environment file. Conda will produce an error like:

```bash
$ conda env create --file pyproject.toml

EnvironmentSpecPluginNotDetected: Unable to detect the environment format for the provided file.
Please add to your prior command: --env-spec <format>

Available formats
    - cep-24: CEP-24 compliant YAML environment specification
    - environment.yml: Standard YAML environment specification with dependencies
    - explicit: Explicit package URLs for fully reproducible environments
    - requirements.txt: Simple text file with package specifications
```

#### 2. Too many environment spec plugins can read the file

This type of failure is when conda has found too many plugins that can read the file. There are 2 subtypes of errors here. For the purposes of testing, the following example are created when enabling autodetection on the environment.yaml plugin with the following diff:

```diff
diff --git a/conda/env/specs/yaml_file.py b/conda/env/specs/yaml_file.py
index 1222992a1..33e556def 100644
--- a/conda/env/specs/yaml_file.py
+++ b/conda/env/specs/yaml_file.py
@@ -25,7 +25,7 @@ class YamlFileSpec(EnvironmentSpecBase):
     # Do not use this plugin for in the environment spec detection process.
     # Users must specify using `environment.yaml` with the `--environment-specifier`
     # option.
-    detection_supported = False
+    detection_supported = True
 
     _environment = None
 
```

##### 1. There are multiple plugins that have a `default_filename` pattern that match the file

Consider the environment.yml file:
```yaml
name: nlp
dependencies:
  - nltk
```

Both the cep-24 and environment.yml plugin use the default filenames `environment.yml`:
```bash
$ conda env create --file environment.yml 

AmbiguousEnvironmentSpecPlugin: File 'environment.yaml' matches the default filename pattern for multiple formats.
Please add to your prior command: --env-spec <format>

Matched formats
    - cep-24: CEP-24 compliant YAML environment specification
    - environment.yml: Standard YAML environment specification with dependencies
```

##### 2. No plugins match a default_filename pattern and there are multiple plugins that can handle the file

Consider the env.yaml file:
```yaml
name: nlp
dependencies:
  - nltk
```

Both the cep-24 and environment.yml plugins can handle  `env.yaml`:
```bash
 $ conda env create --file env.yaml

AmbiguousEnvironmentSpecPlugin: File 'env.yaml' can be handled by multiple formats.
Please add to your prior command: --env-spec <format>

Matched formats
    - cep-24: CEP-24 compliant YAML environment specification
    - environment.yml: Standard YAML environment specification with dependencies
```

#### 3. One more change: Update the error formatting when using the `--env-spec`

Further, this pr updates the explicit plugin request error message to be consistent with the changes above. For example, reusing th `env.yaml` from above

```
$ conda env create --file env.yml --env-spec reqs    

EnvironmentSpecPluginSelectionError: Could not parse 'env.yml' as 'reqs'. Check that the file contents match the expected format. Errors reported from 'reqs':

    ->  PluginError: env.yml contains an invalid matchspec 'dependencies:'

Available formats
    - cep-24: CEP-24 compliant YAML environment specification
    - environment.yml: Standard YAML environment specification with dependencies
    - explicit: Explicit package URLs for fully reproducible environments
    - requirements.txt: Simple text file with package specifications

```

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -1340,20 +1340,35 @@ class SpecNotFound(CondaError):
         super().__init__(msg, *args, **kwargs)
 
 
+def format_env_spec_available_plugins(
+    plugin_specs: list[CondaEnvironmentSpecifier],
+    formats_header: str = "Available formats",
+    indent: int = 4,
+    cmd_suggestion: bool = True,
+):
+    msg = (
+        "\nPlease add to your prior command: --env-spec <format>\n"
+        if cmd_suggestion
+        else ""
+    )
+    msg += f"\n{formats_header}:"
+    msg += dashlist(
+        [f"{plugin.name}: {plugin.description}" for plugin in plugin_specs],
+        indent=indent,
+    )
+    return msg
+
+
 class AmbiguousEnvironmentSpecPlugin(PluginError):
     def __init__(
         self,
         msg: str,
         plugins: list[CondaEnvironmentSpecifier],
-        source: str,
         *args,
         **kwargs,
     ):
-        msg += (
-            f" Try explicitly specifying the format via `--env-spec=FORMAT` or "
-            f"setting `environment_specifier: FORMAT` in your `~/.condarc`.\n\n"
-            f"Matched formats:"
-            f"{dashlist([plg.name for plg in plugins], 4)}"
+        msg += format_env_spec_available_plugins(
+            plugins, formats_header="Matched formats"
         )
         super().__init__(msg, *args, **kwargs)
 
@@ -1362,34 +1377,23 @@ class EnvironmentSpecPluginSelectionError(CondaError):
     def __init__(
         self,
         msg: str,
-        plugin_specs: dict[str, CondaEnvironmentSpecifier],
+        plugin_specs: list[CondaEnvironmentSpecifier],
         *args,
         **kwargs,
     ):
-        plugin_names = [
-            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
-            for name, plugin in plugin_specs.items()
-        ]
-        msg += f"\nAvailable formats: {dashlist(plugin_names, 4)}"
+        msg += format_env_spec_available_plugins(plugin_specs, cmd_suggestion=False)
         super().__init__(msg, *args, **kwargs)
 
 
 class EnvironmentSpecPluginNotDetected(SpecNotFound):
     def __init__(
         self,
-        plugin_specs: dict[str, CondaEnvironmentSpecifier],
+        plugin_specs: list[CondaEnvironmentSpecifier],
         *args,
         **kwargs,
     ):
-        plugin_names = [
-            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
-            for name, plugin in plugin_specs.items()
-        ]
-        msg = (
-            "We weren't able to detect what kind of format this environment is. "
-            "Please add to your prior command: --env-spec [SELECT FROM BELOW]\n"
-            f"{dashlist(plugin_names, 4)}"
-        )
+        msg = "Unable to detect the environment format for the provided file."
+        msg += format_env_spec_available_plugins(plugin_specs)
         super().__init__(msg, *args, **kwargs)
 
 
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -604,10 +604,13 @@ class CondaPluginManager(pluggy.PluginManager):
         try:
             plugin = plugins[name]
         except KeyError:
-            raise CondaValueError(
-                f"You have chosen an unrecognized environment"
-                f" specifier type ({name}). Choose one of: "
-                f"{dashlist(plugins)}"
+            raise EnvironmentSpecPluginSelectionError(
+                msg=(
+                    f"You have chosen an unrecognized environment "
+                    f"specifier type ({name}). Please choose one "
+                    "of the available formats."
+                ),
+                plugin_specs=self.get_hook_results("environment_specifiers"),
             )
         else:
             # Try to load the plugin and check if it can handle the environment spec
@@ -620,11 +623,9 @@ class CondaPluginManager(pluggy.PluginManager):
                         f"Could not parse '{source}' as '{name}'. Check "
                         "that the file contents match the expected format. "
                         f"Errors reported from '{name}':\n\n"
-                        f"    ->  {type(e).__name__}: {str(e)}\n\n"
-                        "If this environment is a different format, try:\n"
-                        f"    conda create --file {source} --format <format>\n"
+                        f"    ->  {type(e).__name__}: {str(e)}\n"
                     ),
-                    plugin_specs=self.get_environment_specifiers(with_aliases=False),
+                    plugin_specs=self.get_hook_results("environment_specifiers"),
                 )
             else:
                 # If the plugin was not able to handle the environment spec, raise an error
@@ -655,9 +656,8 @@ class CondaPluginManager(pluggy.PluginManager):
 
         if len(found) > 1:
             raise AmbiguousEnvironmentSpecPlugin(
-                msg=f"File '{source}' matches the default filename pattern for multiple plugins.",
+                msg=f"File '{source}' matches the default filename pattern for multiple formats.",
                 plugins=found,
-                source=source,
             )
 
         if len(found) == 1:
@@ -726,9 +726,8 @@ class CondaPluginManager(pluggy.PluginManager):
 
             if len(found) > 1:
                 raise AmbiguousEnvironmentSpecPlugin(
-                    msg=f"File '{source}' can be handled by multiple plugins.",
+                    msg=f"File '{source}' can be handled by multiple formats.",
                     plugins=found,
-                    source=source,
                 )
 
         if len(found) == 1:
@@ -749,7 +748,7 @@ class CondaPluginManager(pluggy.PluginManager):
         ) as exc:
             # raise error if no plugins found that can read the environment file
             raise EnvironmentSpecPluginNotDetected(
-                plugin_specs=self.get_environment_specifiers(with_aliases=False),
+                plugin_specs=self.get_hook_results("environment_specifiers")
             ) from exc
 
     def get_environment_specifier(
--- a/tests/plugins/test_env_specs.py
+++ b/tests/plugins/test_env_specs.py
@@ -10,7 +10,6 @@ from conda import plugins
 from conda.auxlib.ish import dals
 from conda.exceptions import (
     AmbiguousEnvironmentSpecPlugin,
-    CondaValueError,
     EnvironmentSpecPluginNotDetected,
     EnvironmentSpecPluginSelectionError,
     PluginError,
@@ -312,7 +311,7 @@ def test_raises_an_error_if_plugin_name_does_not_exist(dummy_random_spec_plugin)
     """
     Ensures that an error is raised if the user requests a plugin that doesn't exist
     """
-    with pytest.raises(CondaValueError):
+    with pytest.raises(EnvironmentSpecPluginSelectionError):
         dummy_random_spec_plugin.get_environment_specifier_by_name(
             name="uhoh", source="test.random"
         )
@@ -344,7 +343,7 @@ def test_raise_error_for_multiple_registered_installers(
     filename = "test.random"
     with pytest.raises(
         AmbiguousEnvironmentSpecPlugin,
-        match=r"File 'test\.random' can be handled by multiple plugins\.",
+        match=r"File 'test\.random' can be handled by multiple formats\.",
     ) as error:
         dummy_random_spec_plugin.get_environment_specifier(filename)
 
@@ -362,7 +361,7 @@ def test_raise_error_for_overlapping_default_filename(
     """
     with pytest.raises(
         AmbiguousEnvironmentSpecPlugin,
-        match=r"File '(.+)environment.xml' matches the default filename pattern for multiple plugins.",
+        match=r"File '(.+)environment.xml' matches the default filename pattern for multiple formats.",
     ) as error:
         plugin_manager_with_xml_spec_2.detect_environment_specifier(str(xml_env))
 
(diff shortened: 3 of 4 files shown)


conda/plugins/manager.py
```diff
@@ -651,12 +652,10 @@ def _detect_filename_env_spec(
         ]
 
         if len(found) > 1:
-            raise PluginError(
-                f"Too many plugins found that can handle the environment file '{source}'.\n\n"
-                "Try using --env-spec=<spec-name> to more exactly specify the environment spec\n"
-                "parser you want to use.\n\n"
-                "Available env specs:\n"
-                f"{dashlist(self.get_environment_specifiers())}"
+            raise AmbiguousEnvironmentSpecPlugin(
```
soapy1, line 658: Is changing the type of error that is raised here in line with conda's deprecation policy?

conda/plugins/manager.py
```diff
@@ -746,11 +743,7 @@ def detect_environment_specifier(self, source: str) -> CondaEnvironmentSpecifier
         except (PluginError, CondaValueError) as exc:
             # raise error if no plugins found that can read the environment file
             raise EnvironmentSpecPluginNotDetected(
-                name=source,
-                plugin_names=hooks,
-                autodetect_disabled_plugins=self.get_environment_specifiers(
-                    supports_detection=False
-                ),
+                plugin_specs=self.get_environment_specifiers(with_aliases=False),
```
soapy1: This `get_environment_specifiers` gets called a handful of times? Is it worth caching the result (maybe will require cache invalidation whenever `load_plugins` function is called)?

conda/exceptions.py
```diff
@@ -1341,32 +1341,40 @@ def __init__(self, msg: str, *args, **kwargs):
...
+        plugin_specs: dict[str, CondaEnvironmentSpecifier],
+        *args,
+        **kwargs,
+    ):
+        plugin_names = [
+            f"{name}{' (' if plugin.aliases else ''}{', '.join(plugin.aliases)}{')' if plugin.aliases else ''}"
+            for name, plugin in plugin_specs.items()
+        ]
+        msg = (
+            "We weren't able to detect what kind of format this environment is. "
```
soapy1: This message is based on option 2 in https://github.com/conda/conda/issues/15718#issuecomment-3973117186. If the user is hitting this error, they have a file that is not going to be able to be parsed by one of these plugins, probably. If the user chooses an env spec plugin to run against [they will get a more specific error message ](https://github.com/conda/conda/pull/15792) about why their file is uncompliant.

conda/exceptions.py
```diff
@@ -1341,32 +1341,40 @@ def __init__(self, msg: str, *args, **kwargs):
...
+            "Please add to your prior command: --env-spec [SELECT FROM BELOW]\n"
+            f"{dashlist(plugin_names, 4)}"
         )
-        if autodetect_disabled_plugins:
```
soapy1, line 1377: `autodetect_disabled_plugins` should be included in the provided `plugin_specs`. We'll no longer distinguish between plugins that have autodetection enabled/disabled.

conda/plugins/manager.py
travishathaway: @soapy1,

Would it be possible for you to have `AmbiguousEnvironmentSpecPlugin` subclass `PluginError` here. That would solve any concerns around existing clients not catching this error.

conda/plugins/manager.py
travishathaway: I don't it's worth caching at this point. This seems like something we could note as a "TODO" in the doc string though so we know to revisit later.

Ideally, I would like to see some empirical evidence (i.e. profiling) before doing something like that.

tests/plugins/test_env_specs.py
```diff
@@ -577,5 +586,8 @@ def test_alias_and_name_collision_detect(
     with pytest.raises(PluginError):
         plugin_manager.get_environment_specifier_by_name("something.random", "random")
 
-    with pytest.raises(PluginError):
+    with pytest.raises(
```
travishathaway, line 593: For backwards compatibility testing, you might want to have a test that asserts a `PluginError`.

travishathaway (commented):
@soapy1, if we can address the point I raised about the new exception, I'll be happy to give an approval on this.

conda/plugins/manager.py
soapy1: Good idea!

tests/plugins/test_env_specs.py
soapy1: Good idea :+1:

conda/exceptions.py
```diff
@@ -1359,29 +1377,19 @@ def __init__(
...
-        self.name = name
-        msg = dals(
-            f"""
-            Environment at {name} is not able to be detected by any installed environment specifier plugins.
-
-            Available plugins: {dashlist(plugin_names, 16)}
-
-            """
+        plugin_names = [
+            f"{name}{' (' if plugin.aliases else ''}{', '.join(plugin.aliases)}{')' if plugin.aliases else ''}"
```
kenodegard:
```suggestion
            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
```

tests/plugins/test_env_specs.py
```diff
@@ -582,7 +591,10 @@ def test_alias_and_name_collision_detect(
...
+        PluginError,
+        match="'something.random' can be handled by multiple plugins.",
```
kenodegard: `match` is always a regex string, let's make it an r-string so its slightly harder to forget

```suggestion
        match=r"'something\.random' can be handled by multiple plugins\.",
```

tests/plugins/test_env_specs.py
```diff
@@ -341,9 +342,17 @@ def test_raise_error_for_multiple_registered_installers(
     for the same section.
     """
     filename = "test.random"
-    with pytest.raises(PluginError):
+    with pytest.raises(
+        AmbiguousEnvironmentSpecPlugin,
+        match="File 'test.random' can be handled by multiple plugins.",
```
kenodegard:
```suggestion
        match=r"File 'test\.random' can be handled by multiple plugins\.",
```

conda/exceptions.py
```diff
@@ -1340,6 +1340,24 @@ def __init__(self, msg: str, *args, **kwargs):
...
+        plugins: list[CondaEnvironmentSpecifier],
+        source: str,
+        *args,
+        **kwargs,
+    ):
+        msg += (
+            " To more exactly specify the format try:\n\n"
+            f"    conda env create --file {source} --env-spec <format>"
+            "\n\nMatched formats:"
+            f"{dashlist([plg.name for plg in plugins], 4)}"
```
kenodegard: Hm can we write this in a way that isn't strictly referencing `conda env create`?


```suggestion
            f" Try explicitly specifying the format via `--env-spec=FORMAT` or "
            f"setting `environment_specifier: FORMAT` in your `~/.condarc`.\n\n"
            f"Matched formats:"
            f"{dashlist([plg.name for plg in plugins], 4)}"
```

travishathaway:
@soapy1 and @kenodegard,  I'm going. to be out for the rest of the week, so please coordinate amongst yourselves to get this merged 🏁

soapy1:
@dashagurova these error messages have changed a bit from what was originally decided on in https://github.com/conda/conda/issues/15718. Could you give this a quick look to make sure we still captured the improvements suggested in #15718?

The pr description shows all the updated error messages and conditions which will trigger them.

conda/exceptions.py
```diff
@@ -1359,29 +1377,19 @@ def __init__(
...
-
-            """
+        plugin_names = [
+            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
+            for name, plugin in plugin_specs.items()
+        ]
+        msg = (
+            "We weren't able to detect what kind of format this environment is. "
+            "Please add to your prior command: --env-spec [SELECT FROM BELOW]\n"
+            f"{dashlist(plugin_names, 4)}"
```
kenodegard: Aligning the wording with what we wrote in `AmbiguousEnvironmentSpecPlugin`:

```suggestion
            f"Unable to detect what kind of format this environment is. "
            f"Try explicitly specifying the format via `--env-spec=FORMAT` or "
            f"setting `environment_specifier: FORMAT` in your `~/.condarc`.\n\n"
            f"Available formats:"
            f"{dashlist(plugin_names, 4)}"
```

conda/exceptions.py
```diff
@@ -1340,6 +1340,24 @@ def __init__(self, msg: str, *args, **kwargs):
...
+        plugins: list[CondaEnvironmentSpecifier],
+        source: str,
+        *args,
+        **kwargs,
+    ):
+        msg += (
+            f" Try explicitly specifying the format via `--env-spec=FORMAT` or "
+            f"setting `environment_specifier: FORMAT` in your `~/.condarc`.\n\n"
+            f"Matched formats:"
+            f"{dashlist([plg.name for plg in plugins], 4)}"
```
kenodegard: Might as well show the aliases here too:


```suggestion
        plugin_names = [
            f"{plugin.name} ({', '.join(plugin.aliases)})" if plugin.aliases else plugin.name
            for plugin in plugins
        ]
        msg += (
            f" Try explicitly specifying the format via `--env-spec=FORMAT` or "
            f"setting `environment_specifier: FORMAT` in your `~/.condarc`.\n\n"
            f"Matched formats:"
            f"{dashlist(plugin_names, 4)}"
```

soapy1: that's a good point. Lets extract this formatting so it can be reused across these errors.

soapy1: I've added a new function `conda.exceptions.format_env_spec_available_plugins` that formats a list of `CondaEnvironmentSpecifiers` into a list of available plugins with their associated descriptions.

conda/exceptions.py
soapy1: @kenodegard how would you feel about making this say 
```
For example:
    conda <command> --file <name of file> --env-spec <format>
```

This gives the user a very specific and simple action to resolve the problem

conda/exceptions.py
```diff
@@ -1340,48 +1340,72 @@ def __init__(self, msg: str, *args, **kwargs):
...
+    available_plugins_table = "".join(
+        "\n"
+        + " " * indent
+        + f"{prefix} "
+        + x.name
+        + " " * (max_name_len - len(x.name) + 2)
+        + " "
+        + x.description
+        for x in plugin_specs
+    )
```
jezdez: There is `dashlist` and `pretty_list` functions for this

conda/exceptions.py
```diff
@@ -1340,48 +1340,72 @@ def __init__(self, msg: str, *args, **kwargs):
...
+class EnvironmentSpecPluginNotDetected(SpecNotFound):
+    def __init__(
+        self,
+        plugin_specs: list[CondaEnvironmentSpecifier],
+        *args,
+        **kwargs,
+    ):
+        msg = (
+            "We weren't able to detect what kind of format this environment is. "
+            "\nPlease add to your prior command: --env-spec [SELECT FROM BELOW]\n"
```
jezdez: The tone here ("We weren't able to detect...") is inconsistent with the rest of the exception messages in this file. Existing messages use a factual, impersonal style, e.g.:
- "Could not find conda environment: ..."
- "No environment exporter plugin found for filename '...'"
- "The package '...' is disallowed by configuration."

```suggestion
        msg = (
            "Unable to detect the environment format for the provided file."
            "\nPlease add to your prior command: --env-spec [SELECT FROM BELOW]\n"
            "\nFor example:"
            "\n    conda <command> --file <name of file> --env-spec <format>\n"
        )
```

conda/exceptions.py
```diff
@@ -1340,48 +1340,72 @@ def __init__(self, msg: str, *args, **kwargs):
...
     ):
-        plugin_names = [
-            f"{name} ({', '.join(plugin.aliases)})" if plugin.aliases else name
-            for name, plugin in plugin_specs.items()
-        ]
-        msg += f"\nAvailable formats: {dashlist(plugin_names, 4)}"
+        msg += (
+            "\nPlease add to your prior command: --env-spec [SELECT FROM BELOW]\n"
+            "\nFor example:"
+            "\n    conda <command> --file <name of file> --env-spec <format>\n"
```
jezdez: This instruction block is duplicated verbatim in `EnvironmentSpecPluginNotDetected` below. Consider extracting it to a shared constant or building it in a helper, to keep them in sync.

jezdez: Nit: "[SELECT FROM BELOW]" reads like a form placeholder rather than CLI output. The existing pattern in the codebase uses concrete examples, e.g. `CommandNotFoundError` shows `$ conda init <SHELL_NAME>`. Consider:

```suggestion
            "\nPlease specify the format explicitly: --env-spec <format>\n"
```

conda/exceptions.py
soapy1: This change to the formatting was an intentional decision to stylize the output to look a bit more pretty:
```
Available formats:
      cep-24             CEP-24 compliant YAML environment specification
      environment.yml    Standard YAML environment specification with dependencies
      explicit           Explicit package URLs for fully reproducible environments
      requirements.txt   Simple text file with package specifications
```

The rest of conda doesn't use this type of styling. So, I think it's reasonable to roll this back. It might be worth it to come back to this alterative styling when we start working on some design improvements. At that point we could make this type of formatting a bit more generic too.

Using `dashlist`  the output looks like:

```
Available formats
    - cep-24: CEP-24 compliant YAML environment specification
    - environment.yml: Standard YAML environment specification with dependencies
    - explicit: Explicit package URLs for fully reproducible environments
    - requirements.txt: Simple text file with package specifications
```

conda/exceptions.py
soapy1: This softening of the language was also an intentional change discussed (briefly) in the [source issue](https://github.com/conda/conda/issues/15718#issuecomment-4085364637) (q 2). 
Let's keep the error messages consistent with the rest of conda for now. We can revisit this new tone to the laguage later.

conda/exceptions.py
soapy1: Included this suggested change with the fix for https://github.com/conda/conda/pull/15793#discussion_r3008684520

conda/exceptions.py
jezdez: Ah, I'm sorry, I missed that part, I think a wholesale updating of the tone will be incredibly welcome and IIUC the next quarter will allow that type of work

PR #15813, opened 2026-03-25
(Experiment, DNR) `ng` CLI for `conda`

jaimergp:
### Description

This PR adds a CLI reimplementation of `conda` using `py-rattler` as its main driver. Consider it an experiment to see what we can do with `py-rattler` and `rich` today.

Try it from your dev env with one of:

```
CONDA_EXPERIMENTAL=ng python -m conda ...
python -m conda.ng ...
```

Some subcommands have been reimplemented, others are still forwarding to the original `conda` one. The reimplementations are not 1:1 replacements at all, I took many shortcuts. Again, just playing to see what we can achieve.

- [x] `conda activate`: equivalent to `spawn`, as it creates a new shell
- [x] `conda create`
- [ ] `conda compare`
- [ ] `conda config`
- [ ] `conda deactivate`: just run `exit` in the new shell
- [ ] `conda env`
- [ ] `conda export`
- [x] `conda info`
- [x] `conda install`
- [x] `conda list`
- [ ] `conda notices`
- [ ] `conda package`
- [x] `conda remove`
- [ ] `conda rename`
- [x] `conda run`
- [x] `conda search`
- [x] `conda update`

What I think it's attractive from this experiment is how we could actually develop on the side this without flipping the whole table:

- `conda._ng` is a private namespace where we are developing the CLI. Maybe it will also end up containing an `api` subpackage with small Python facades in front of different py-rattler components.
- The CLI bits are using the 100% SAME `argparse` definitions (our `configure_parser` functions). We just change what the `execute` function does in each module. The module names are the same.
- If we detect `ng` as part of `conda.experimental`, then we import the modules from `conda._ng.cli` instead of `conda.cli`, with a fallback when absent..
- We can still have the plugin system around: we make sure that the hook calls are placed in the right positions.

Other benefits:

- Reveals the application/library boundaries quite clearly.
- Since `py-rattler` has no notion of a `context`, this allows us to see where our global config is leaking: we need to pass it explicitly. Informative for refactors.

Cons:

- Difficult to test thoroughly. Maybe we are lucky with integration tests, but I anticipate some work to make it compatible.

### Checklist - did you ...

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -149,6 +149,12 @@ def do_not_notify_outdated_conda(monkeypatch):
     monkeypatch.setenv("CONDA_NOTIFY_OUTDATED_CONDA", "false")
 
 
+@pytest.fixture(autouse=True)
+def force_use_ng(monkeypatch):
+    """Opt-in to `conda-ng` to see what breaks"""
+    monkeypatch.setenv("CONDA_EXPERIMENTAL", "ng")
+
+
 @pytest.fixture
 def plugin_manager(mocker) -> CondaPluginManager:
     pm = CondaPluginManager()


jaimergp:
Some screenshots as of [ec4f066](https://github.com/conda/conda/pull/15813/commits/ec4f0669aacfbbe60e0ccb4c9db051ba2ae63b3a)

# `conda list py-rattler`

image


---


# `conda create -d python`

image
image

---

# `conda info`

image
image

jaimergp:
Some exception niceness. Still not very informative, but at least it's, you know, red.

image

Issue #15581, opened 2026-01-12
Warning on mixed conda/pip environments during export


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Display a Warning when exporting lockfiles from environments containing packages installed with pip, uv, or other 3rd party Python package managers.

This depends on https://github.com/conda/conda/pull/15530, https://github.com/conda-incubator/conda-pypi/issues/63, and https://github.com/conda-incubator/conda-pypi/issues/82.

### Why is this needed?

Conda cannot reliably lock 3rd party Python packages because they don't provide the same dependency resolution guarantees as conda packages. Currently, `conda export` silently includes pip packages in the exported specification, which may lead to:

- Inconsistent environment reproduction - 3rd party Python packages may resolve to different versions
- Broken dependency chains - conda cannot verify pip package dependencies are satisfied
- User confusion - Users expect exported lockfiles to be fully reproducible

This creates a false sense of security and undermines the purpose of lockfiles. Users should be guided to use conda-pypi for proper integration before exporting.

### What should happen?

When `conda export` detects 3rd party Python packages in an environment, it should:

1. Raise a clear warning explaining that 3rd party packages cannot be included in lockfiles
2. List the detected packages (e.g., "requests (pip), flask (pip)")
3. Provide actionable solutions:
   - Install packages using `conda-pypi` for proper integration
   - Run `conda doctor --fix` to migrate existing pip packages
4. Link to documentation on conda-pypi and pip interoperability

**Example error message:**

```
CondaExportWarning: The exported environment contains 3rd party Python packages

Your environment contains 2 packages installed via pip. Conda cannot reliably
lock these packages for reproducible environments.

Detected packages:
  - requests==2.28.1 (pip)
  - flask==2.3.0 (pip)

Or use conda doctor to help remove the packages:
  conda doctor --fix

Learn more: https://docs.conda.io/.../pip-interoperability.html
```

### Additional Context

**Implementation approach:**
- Check `env.external_packages` in `conda/cli/main_export.py::execute()`
- Detect using existing `PrefixData.get_python_packages()` method

**Key code locations:**
- `conda/cli/main_export.py` (lines 138-220)
- `conda/models/environment.py` (lines 371-456)
- `conda/core/prefix_data.py::get_python_packages()` (lines 589-604)

jezdez:
Thanks for working on this. Calling out pip-installed packages on export is a solid improvement.

[...]

Issue #14915, opened 2025-06-06
Clarify `--quiet` flag behavior and add dedicated progress bar control


kenodegard:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Currently, conda's `--quiet` flag controls unrelated behaviors (progress bars, output messages, notifications, etc.) and is independent of the `--verbose` flag, leading to confusing combinations like `conda create --quiet --verbose`. We propose:

- Add a dedicated `--no-progress` flag to control only progress bar display
- Restructure `--quiet` to work as a proper counterpart to `--verbose` (i.e., reduce overall verbosity level)

#### Current problems:

- `--quiet` and `--verbose` are orthogonal, allowing confusing combinations like `--quiet --verbose`
- `--quiet` controls different behaviors (e.g., no progress bars and less output)
- No clean way to disable only progress bars (needed for debugging with pdb, CI logs, etc.)
- The current behavior is inconsistent with user expectations and other tools

#### Specific use cases:

- Debugging: Need to disable progress bars when setting breakpoints to avoid spinner interference with pdb prompts
- CI environments: Many CI scripts use -q expecting it to just disable progress bars, but it affects other output too
- Interactive detection doesn't work properly in CI logs (progress bars still appear when they shouldn't)

#### Phase 1: Add `--no-progress` flag

- Add `--no-progress` flag that only controls progress bar display
- Follow pip's pattern: `--progress-bar <on|off>` (defaulting to on)
- Automatically disable progress bars when stderr is not a terminal

#### Phase 2: Restructure `--quiet` (breaking change)

- Make `--quiet` work as the inverse of `--verbose` (reduce verbosity level)
- Support multiple levels: `-q`, `-qq`, etc.
- Maintain backward compatibility where possible, but document breaking changes

#### Current `--quiet` behaviors to redistribute:

- Progress bars → `--no-progress`
- `conda activate` messages after creation → verbosity level
- Transaction summaries → verbosity level
- Conda update notifications → separate flag (?) or verbosity level
- Pip section verbosity in `environment.yml` → verbosity level
- Crash log upload prompts → separate flag (?) or verbosity level

#### Reference implementations:

`pip`: `--progress-bar <on|off|raw>`
`pixi`: `--no-progress` + `--quiet` as verbosity reducer

#### Current verbose levels:

`-v`: more details
`-vv`: log level info
`-vvv`: log level debug
`-vvvv`: log level trace

[...]

PR #15838, opened 2026-04-01
Warn when exporting an environment with packages installed via pip

soapy1:
### Description
This adds a warning whenever a user is exporting an environment that has pip dependencies. For example, consider an environment:

```
$ conda create -p /tmp/exp python
$ conda activate /tmp/exp
$ pip install numpy
```

Exporting this environment now produces the result:

```
$ conda export -p /tmp/exp --format yaml --file env.yaml                                                                                                     
CondaExportWarning: The exported environment contains 3rd party Python packages.

Your environment contains 5 packages installed via pip. Conda cannot reliably lock these packages for reproducible environments.

Detected packages:
  - annotated-types==0.7.0
  - conda-lockfiles==0.1.2.dev26+gffb5db85b
  - pydantic==2.12.5
  - pydantic-core==2.41.5
  - typing-inspection==0.4.2

Learn more: https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/pip-interoperability.html
```

Users may also disable warnings:
```
$ PYTHONWARNINGS=ignore conda export -p /tmp/exp --format yaml --file env.yaml    
# no warning
```

or using the `--quiet` flag

```
$ conda export -p /tmp/exp --quiet
# no warning
```

fixes: https://github.com/conda/conda/issues/15581

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_export.py
+++ b/conda/cli/main_export.py
@@ -210,16 +210,16 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
 
     pd = PrefixData(prefix, interoperability=True)
     pypi_packages = pd.get_python_packages()
-    warns = []
+    warning: str | None = None
     if pypi_packages:
-        warns.append(
+        warning = (
             "The exported environment contains 3rd party Python packages.\n\n"
             f"Your environment contains {len(pypi_packages)} package{'s' if len(pypi_packages) > 1 else ''} "
             "installed via pip. Conda cannot reliably lock these packages for reproducible "
             "environments.\n\n"
             "Detected packages:"
             f"{dashlist([package.to_simple_match_spec() for package in pypi_packages])}"
-            "\n\nLearn more: https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/pip-interoperability.html",
+            "\n\nLearn more: https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/pip-interoperability.html"
         )
 
     # Export using the appropriate method
@@ -246,14 +246,14 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
                     "success": True,
                     "file": args.file,
                     "format": target_format,
-                    "warnings": warns,
+                    "warnings": [warning] if warning else [],
                 }
             )
-        elif warns and not context.quiet:
-            warnings.warn("\n\n".join(warns), CondaExportWarning)
+        elif warning and not context.quiet:
+            warnings.warn(warning, CondaExportWarning)
     else:
-        if warns and not context.quiet:
-            warnings.warn("\n\n".join(warns), CondaExportWarning)
+        if warning and not context.quiet:
+            warnings.warn(warning, CondaExportWarning)
         print(exported_content, end="")
 
     return 0
--- a/tests/cli/test_main_export.py
+++ b/tests/cli/test_main_export.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
     from pathlib import Path
 
     from conda.plugins.manager import CondaPluginManager
-    from conda.testing.fixtures import CondaCLIFixture, PipCLIFixture
+    from conda.testing.fixtures import CondaCLIFixture, PipCLIFixture, TmpEnvFixture
 
 
 def test_export(conda_cli: CondaCLIFixture) -> None:


conda/cli/main_export.py
```diff
@@ -201,6 +203,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
+    pd = PrefixData(prefix, interoperability=True)
```
soapy1, line 211: Also considered pulling this information from the `env` created above. This has a map of `external_packages`. Though, that is populated by a plugin. We wouldn't want bugs in plugins to make this warning pop up, so I think its safer to look at prefix data directly here.

conda/cli/main_export.py
```diff
@@ -201,6 +203,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    pypi_packages = pd.get_python_packages()
+    if pypi_packages:
+        print(
```
soapy1: This will just print the output. What should the behaviour be if a user provides the `--json` or `quite` flags?

conda/cli/main_export.py
```diff
@@ -201,6 +208,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
+    pd = PrefixData(prefix, interoperability=True)
+    pypi_packages = pd.get_python_packages()
+    if pypi_packages:
```
jezdez, line 214: Skip this when --json or --quiet. A stderr warning next to JSON output is awkward for scripting, and quiet should stay quiet.

conda/cli/main_export.py
```diff
@@ -201,6 +208,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        warnings.warn(
+            "The exported environment contains 3rd party Python packages.\n\n"
+            f"Your environment contains {len(pypi_packages)} packages installed "
+            "via pip. Conda cannot reliably lock these packages for reproducible "
+            "environments.\n\n"
+            "Detected packages:"
+            f"{dashlist([package.to_simple_match_spec() for package in pypi_packages])}"
+            "\n\nUse conda doctor to help remove the packages:\n"
+            "    conda doctor --fix\n\n"
```
jezdez: conda doctor --fix only really helps here if you have something like conda-pypi installed. On a stock install it reads like core conda promises behavior we do not ship. I would drop this bit, qualify it for conda-pypi users only, or stick to the docs link and generic guidance.

conda/cli/main_export.py
```diff
@@ -201,6 +208,22 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         channels=context.channels,
     )
 
+    pd = PrefixData(prefix, interoperability=True)
+    pypi_packages = pd.get_python_packages()
+    if pypi_packages:
+        warnings.warn(
+            "The exported environment contains 3rd party Python packages.\n\n"
+            f"Your environment contains {len(pypi_packages)} packages installed "
```
jezdez: Small copy nit: use singular when count is 1, or reword so agreement is not wrong.

tests/cli/test_main_export.py
```diff
@@ -637,7 +638,7 @@ def test_export_pip_dependencies_handling(conda_cli, format_name, parser_func):
         elif isinstance(dep, str):
             conda_deps.append(dep)
 
-    # If pip dependencies exist, verify their format
+    # If pip dependencies exist, verify their format and that a warning is present in stderr
```
jezdez: The comment says we check stderr for the warning, but the test never looks at stderr or uses pytest.warns. Either assert it or shorten the comment.

tests/cli/test_main_export.py
```diff
@@ -688,36 +690,40 @@ def test_export_with_pip_dependencies_integration(
             [format_flag] if format_flag else []
         )
 
-        stdout, stderr, code = conda_cli(*export_args)
-        assert code == 0, f"{format_name} export failed: {stderr}"
+        with pytest.warns(
```
jezdez: After json/quiet gating, add a case where export --json or -q with pip packages does not emit CondaExportWarning (warnings.catch_warnings or similar).

conda/cli/main_export.py
soapy1: hmmm, I'm reviewing the meaning of the `--json` flag here. I think it's a bit confusing. If the `--json` flag appears it mostly has the same meaning as `--format json` (the `--format` flag will take precedence over the `--json` flag if both are present).

But in the case when the `--json` flag is combined with the `--file` flag, it does not impact the format at all. Instead it is making the output to stdout be in json format. 

So, I think it's appropriate to:
1. when the `--quite` flag is provided -> skip the warning
2. when the `--json` and `--file` flag are both provided -> include the warning in the output status json
3. when the `--json` flag is provided without the `--file` flag (modifying the format) -> output the warning in order to be consistent with `conda export --format json`
4. every other command line opts -> output the warning

A couple of examples:
#### base line
```
$ conda export 
CondaExportWarning: The exported environment contains 3rd party Python packages. . .

channels: 
. . .
```

#### 1. `--quiet` flag
```
$ conda export --quiet      
# no warning
channels:
. . .
```

#### 2. json output format
```
$ conda export --json --file environment.yaml
# no warning
{
  "success": true,
  "file": "environment.yaml",
  "format": "environment-yaml"
  "warnings": ["The exported environment contains 3rd party Python packages..."]
}
```

#### 3. json export format
```
$ conda export  --json
CondaExportWarning: The exported environment contains 3rd party Python packages. . .

{
  "channels": {  
. . .
```

This is equivalent to 

```
$ conda export  --format json
CondaExportWarning: The exported environment contains 3rd party Python packages. . .

{
  "channels": {  
. . .
```

#### 4.1 just `--file` arg
```
 $ conda export --file spec.txt           
CondaExportWarning: The exported environment contains 3rd party Python packages...
```

#### 4.2 a prefix without pip packages
```
$ conda export -p /idont/exist           
# no pip packages - no warning
prefix: /idont/exist
. . .
```



Does this seem reasonable?

conda/cli/main_export.py
```diff
@@ -220,8 +241,19 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-            stdout_json({"success": True, "file": args.file, "format": target_format})
+            stdout_json(
+                {
+                    "success": True,
+                    "file": args.file,
+                    "format": target_format,
+                    "warnings": warns,
+                }
+            )
+        elif warns and not context.quiet:
```
soapy1: It would maybe be neat if conda could supress warnings globally if `context.quiet` was enabled. Maybe somewhere in https://github.com/conda/conda/blob/main/conda/cli/main.py#L25. Related to https://github.com/conda/conda/issues/14915

conda/cli/main_export.py
jezdez: > hmmm, I'm reviewing the meaning of the `--json` flag here. I think it's a bit confusing. If the `--json` flag appears it mostly has the same meaning as `--format json` (the `--format` flag will take precedence over the `--json` flag if both are present).
> 
> But in the case when the `--json` flag is combined with the `--file` flag, it does not impact the format at all. Instead it is making the output to stdout be in json format.

IIRC this was the behavior before the export refactor and had to be retained

soapy1: +1 to keeping existing behaviour for backwards compatibility.

To clariy the path forward here, would it be better if any time the `--json` flag is passed the warning is suppressed?

This means:
* `conda export --json` does not produce the same output as `conda export --format json` (the first command will have no warning, the second command will)
* `conda export --json --file environment.yaml` will include the warning in the output json document

conda/cli/main_export.py
jezdez: Let's follow up on that after the PR?

conda/cli/main_export.py
jezdez: Hm, I'd think the JSON output should be valid to stdout atleast?

soapy1: So like, for `conda export --json` and `conda export --format json` there should be no warning output? That means if user are outputting to json they will never see the warning unless they are using the `--file` argument.

I guess one more thing to note here is that this is using [python warnings module](https://docs.python.org/3/library/warnings.html), so all this warning output is going to stderr. And they can be fully supressed by setting `PYTHONWARNINGS=ignore`. I can update the tests to make that more clear.

soapy1: @jezdez following up on this question.

conda/cli/main_export.py
```diff
@@ -201,6 +208,20 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+    if pypi_packages:
+        warns.append(
+            "The exported environment contains 3rd party Python packages.\n\n"
+            f"Your environment contains {len(pypi_packages)} package{'s' if len(pypi_packages) > 1 else ''} "
+            "installed via pip. Conda cannot reliably lock these packages for reproducible "
+            "environments.\n\n"
+            "Detected packages:"
+            f"{dashlist([package.to_simple_match_spec() for package in pypi_packages])}"
+            "\n\nLearn more: https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/pip-interoperability.html",
+        )
```
jezdez:
```suggestion
    warning: str | None = None
    if pypi_packages:
        warning = (
            "The exported environment contains 3rd party Python packages.\n\n"
            f"Your environment contains {len(pypi_packages)} package{'s' if len(pypi_packages) > 1 else ''} "
            "installed via pip. Conda cannot reliably lock these packages for reproducible "
            "environments.\n\n"
            "Detected packages:"
            f"{dashlist([package.to_simple_match_spec() for package in pypi_packages])}"
            "\n\nLearn more: https://docs.conda.io/projects/conda/en/stable/user-guide/configuration/pip-interoperability.html"
        )
```

one string, no `"\n\n".join(...)` at the call sites. see the output-block suggestion below for the cascade.

conda/cli/main_export.py
```diff
@@ -220,8 +241,19 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+                    "file": args.file,
+                    "format": target_format,
+                    "warnings": warns,
+                }
+            )
+        elif warns and not context.quiet:
+            warnings.warn("\n\n".join(warns), CondaExportWarning)
     else:
+        if warns and not context.quiet:
+            warnings.warn("\n\n".join(warns), CondaExportWarning)
```
jezdez: cascade for the string refactor above:

```suggestion
        if args.json:
            stdout_json(
                {
                    "success": True,
                    "file": args.file,
                    "format": target_format,
                    "warnings": [warning] if warning else [],
                }
            )
        elif warning and not context.quiet:
            warnings.warn(warning, CondaExportWarning)
    else:
        if warning and not context.quiet:
            warnings.warn(warning, CondaExportWarning)
```

`[warning] if warning else []` keeps `"warnings"` always present as a list in the JSON blob.

tests/cli/test_main_export.py
```diff
@@ -720,6 +712,52 @@ def test_export_with_pip_dependencies_integration(
...
+
+def test_export_non_pip_env_warnings(
+    conda_cli: CondaCLIFixture,
+):
+    """Test that conda does not warn for environments without pip dependencies"""
+    # Export the environment in the specified format
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always", category=CondaExportWarning)
+        conda_cli("export", "--prefix=/does/not/exist")
+        assert len(w) == 0
```
jezdez:
```suggestion
def test_export_non_pip_env_warnings(
    tmp_env: TmpEnvFixture,
    conda_cli: CondaCLIFixture,
):
    """Test that conda does not warn for environments without pip dependencies"""
    with tmp_env("python") as prefix:
        PrefixData._cache_.clear()
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always", category=CondaExportWarning)
            conda_cli("export", f"--prefix={prefix}")
            assert len(w) == 0
```

exercises a real pip-less env so the assertion stays meaningful if `export` ever stops tolerating missing prefixes. `TmpEnvFixture` is already used elsewhere in this file.

jezdez (approved):
LGTM with comments above

PR #15841, opened 2026-04-02
add external-packages check

### LINKED ISSUE #15582: Add `conda doctor --fix` for `external-packages`
### Checklist

### What is the idea?

Add an `external-packages` health check to `conda doctor` that detects packages installed via pip/uv and provides a `--fix` option to migrate them to conda-pypi.

### Why is this needed?

There are two main use cases for this:

- Pip packages can't be reliably locked in conda lockfiles.
- When you try to protect your base environment (see conda-incubator/conda-self#88), it clones base to a default environment. If base contains pip packages, the clone can no longer ensure your new environment matched what you had before.

 A health check would let them detect and fix the issue proactively.

### What should happen?

- Detection: Use `PrefixData.get_python_packages()` to find packages installed from PyPI. In verbose mode, list each package. User-facing copy should say "installed from PyPI" rather than "pip-installed".
- Fix: `conda doctor --fix` runs `conda install` using the standard channel configuration, pinning each package to its exact installed version. If a wheels channel is in the user's `channels` list, it participates in the solve like any other channel. The fix attempts a single solve for all packages at once. If any packages cannot be satisfied, they are identified and dropped, and the solve is retried for the remaining set. The user is shown which packages will be migrated and which will be skipped before any changes are made.
- `--fix` does not call PyPI directly. Packages that cannot be satisfied from the configured channels are skipped with a clear message.

Example UX (illustrative):

```text
$ conda doctor
✗ Found 3 packages installed from PyPI.
  Run 'conda doctor --fix' to migrate them using your configured channels.

$ conda doctor --verbose
✗ Found 3 packages installed from PyPI.
  Channels: my-wheels, conda-forge
  - requests 2.32.3
  - flask 3.1.0
  - custom-package 1.0.0

$ conda doctor --fix
Found 3 packages installed from PyPI. Recommend adding a wheel channel to conda: https://conda.github.io/conda-pypi/add-channel (this doesn't exist yet).
Migrate using channels [conda-pypi, conda-forge]? [y/N]: y
  ✓ requests
  ✓ flask
  ✗ custom-package: not available on configured channels

Summary: 2 migrated, 1 skipped
```

ForgottenProgramme:
### Description

resolves #15582 
TODO:

- [x] Add health check 
- [x] Add tests
- [x] Add fix
- [ ] add test for the fix

### Checklist - did you ...


conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -32,10 +36,96 @@ def print_external_packages(prefix: str, verbose: bool) -> None:
         print("")
 
 
+FORBIDDEN_LIST = {"msgpack", "ruamel-yaml", "pip", "setuptools"}
```
ForgottenProgramme: I ended up in a terribly broken environment when I ended up uninstalling these packages using pip. So I am randomly adding them to a "forbidden list". 
Maybe they don't belong here, but these packages once deleted using pip caused my conda to break

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -32,10 +36,96 @@ def print_external_packages(prefix: str, verbose: bool) -> None:
...
+
+import subprocess
+
+
+def conda_has_package(name: str) -> bool:
```
ForgottenProgramme, line 42: do a `conda search` to ensure packages are available on conda

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -0,0 +1,131 @@
...
+            print(f"Failed to uninstall {name} using `pip`.")
+            continue
+
+        successful_uninstalls.append(name)
+
+    if not successful_uninstalls:
+        print("No packages were successfully uninstalled. Aborting reinstall.")
+        return 0
+
+    return reinstall_packages(args, successful_uninstalls, force_reinstall=True)
```
ForgottenProgramme: I am not taking into account exact versions. Probably I should ensure the same versions are installed.

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -0,0 +1,131 @@
...
+import subprocess
+
+
+def conda_has_package(name: str) -> bool:
+    result = subprocess.run(
+        ["conda", "search", name, "--json"],
+        capture_output=True,
+        text=True,
+    )
+    return "error" not in result.stdout and result.returncode == 0
```
kenodegard, line 48: `SubdirData` allows you to do the same thing in-process:

```suggestion
    return bool(SubdirData.query_all(name, context.channels, context.subdirs))
```

conda/plugins/subcommands/doctor/health_checks/external_packages.py
kenodegard: did these break your `base` env or did you run `conda doctor` on a different env?

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -0,0 +1,131 @@
...
+        successful_uninstalls.append(name)
+
+    if not successful_uninstalls:
+        print("No packages were successfully uninstalled. Aborting reinstall.")
+        return 0
+
+    return reinstall_packages(args, successful_uninstalls, force_reinstall=True)
+
+
+def migrate_to_pypi(prefix: str, args: Namespace, confirm: ConfirmCallback) -> int:
```
kenodegard, line 67: running a pypi migration on the `base` env has a high probability of breaking conda so we should prevent that

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -0,0 +1,131 @@
...
+    else:
+        print(f"{X_MARK} These packages are not installed by conda:\n")
+        for package in external_packages:
+            print(package.name)
+        print("")
+
+
+FORBIDDEN_LIST = {"msgpack", "ruamel-yaml", "pip", "setuptools"}
+
+import subprocess
```
kenodegard: move to the top

conda/plugins/subcommands/doctor/health_checks/external_packages.py
```diff
@@ -0,0 +1,131 @@
...
+        # check if conda can install it
+        if conda_has_package(name):
+            safe.append(name)
+        else:
+            external_only.append(name)
+
+    return safe, external_only
+
+
+import sys
```
kenodegard: move to the top

conda/plugins/subcommands/doctor/health_checks/external_packages.py
jezdez: @ForgottenProgramme Let's make this happen in-process

conda/plugins/subcommands/doctor/health_checks/external_packages.py
kenodegard: since we now prevent migrations for the base env this `FORBIDDEN_LIST` should be removed

kenodegard (commented):
I think this health_check belongs in conda-pypi.

While this initial implementation does not utilize the features in conda-pypi I expect a future version of this health_check to replace pip installs with conda packages (when available) and use native wheels installs as a fallback.

ForgottenProgramme:
closing this in favor of https://github.com/conda/conda-pypi/pull/347

PR #15850, opened 2026-04-03
Add startup performance benchmarks for CodSpeed

jezdez:
Part of #15867 (A5b).

conda's existing CodSpeed benchmark suite covers solver and subcommand-level performance but does not track the cost of the startup path itself. This adds nine tests that cover the phases between process start and first useful work:

**Import benchmarks** (3 tests) measure the CPU cost of importing `conda.cli.main`, `conda.base.context`, and `conda.cli.conda_argparse` from a clean `sys.modules` state. They use `benchmark.pedantic` with a setup function that scrubs all non-stdlib, non-test-infrastructure modules between rounds so each iteration does real import work. The cleanup uses `sys.stdlib_module_names` rather than a hardcoded list, so it automatically adapts when dependencies change.

**Function benchmarks** (3 tests) measure repeatedly-callable functions on the startup path: `context.__init__()` (config loading), `generate_parser()` (full CLI parser build), and `main("--version")` (end-to-end through the fast path).

Both groups are marked `@pytest.mark.benchmark` and are collected exclusively by the `linux-benchmarks` CI job (`pytest --codspeed`). Regular unit and integration jobs now filter them out via `-m "not integration and not benchmark"`.

**Module-count guardrails** (5 parametrized tests) run a subprocess with a clean interpreter and assert that each startup phase stays within a module budget. These are _not_ marked `benchmark` — they run in every CI job as lightweight guardrails that catch accidental dependency additions in the startup path. The budgets should ratchet down as startup optimizations land (particularly `import_argparse` and `generate_parser` once #15868 merges).

### Checklist - did you ...

      This PR is entirely new tests.
- ~Add / update outdated documentation?~
      Not applicable — no user-facing changes.


dholth:
Good idea. It's tricky to measure startup times! but the user can feel it.

dholth (approved):
No reason not to merge this one. Thanks for looking into this hard to measure area.

PR #15866, opened 2026-04-07
Use stdlib json instead of requests.compat.json

jezdez:
### Description

Re-applies #15849, which was reverted in #15865 to allow the startup benchmarks (#15850) to establish a CodSpeed baseline first.

Replace `from requests.compat import json` with `import json` in `conda/common/serialize/json.py`.

The `requests.compat` module is a Python 2-era compatibility shim whose JSON handling does:

```python
try:
    import simplejson as json
except ImportError:
    import json
```

Since `simplejson` is never installed in conda environments, this always resolves to stdlib `json`. But importing `requests.compat` to get there also unconditionally loads `urllib3`, `chardet`/`charset_normalizer`, and a chain of ~120 other modules — all to end up with the same `json` module we could import directly.

```
requests (67) → urllib3 (29) → cryptography (30) → charset_normalizer (9) → email.* (15)
```

Measured impact (hyperfine, Python 3.13):

| Phase                    | Before | After | Saved  |
| ------------------------ | ------ | ----- | ------ |
| Context init time        | 109 ms | 52 ms | −57 ms |
| Modules at context ready | 389    | 209   | −180   |

### Checklist - did you ...

  No behavioral change — `simplejson` is not present in conda environments and `requests.compat.json` always resolves to stdlib `json`. All existing tests cover this path unchanged.
  Not applicable — internal import change with no user-facing API impact.

### DIFF
--- a/conda/common/serialize/json.py
+++ b/conda/common/serialize/json.py
@@ -4,14 +4,12 @@
 
 from __future__ import annotations
 
+import json  # noqa: TID251
 from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, overload
 
 from frozendict import frozendict
-
-# detect the best json library to use
-from requests.compat import json
 
 if TYPE_CHECKING:
     from io import IO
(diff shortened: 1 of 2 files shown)


dholth:
Fun fact, stdlib json (which we were using anyway) is surprisingly fast.

jezdez:
@kenodegard Could I please get another + after we reverted the previous attempt? Codspeed successfully recognized the change above ^

kenodegard (approved):

PR #15868, opened 2026-04-07
Lazy subcommand parser loading (A2/A3)

jezdez:
Part of #15867 (A2 + A3).

### Problem

`conda/cli/conda_argparse.py` used to import all 20 `main_*` subcommand modules at module-load time via 20 `from .main_X import configure_parser as configure_parser_X` statements. Since `conda_argparse` is loaded on every `conda` invocation, every subcommand module -- and its transitive dependency tree -- was imported regardless of which command was actually run.

### Solution

Introduces `_LazySubParsersAction` (subclass of `_GreedySubParsersAction`) and a `_LazyParserMap` for the `_name_parser_map`:

- Stub parsers (name + help text) are registered for all built-in subcommands at `generate_parser()` time, without importing their modules.
- The real `configure_parser()` call is deferred until the subcommand is actually invoked; at that point its module is imported and the stub is filled in.
- `_LazyParserMap.keys()` / `__iter__` / `__contains__` answer from the pending-loader table and alias map, so argparse's help formatting and error messages don't trigger mass loading.
- Plugin discovery is deferred until `--help` is rendered (`_get_subactions`) or a subcommand is dispatched (`__call__`) -- `conda commands`, invalid-command errors, and the built-in-override check all still work.

The deprecated `configure_parser_*` functions and `*_rc_path` strings remain importable from `conda.cli.conda_argparse` and are registered via `deprecated.constant(factory=...)` (the canonical API from #15925), so their target modules (`conda.cli.main_*`, `conda.base.context`) also stay off the cold-start path until the deprecated name is touched. First access emits `PendingDeprecationWarning` as expected.

### Measured impact

macOS ARM64, devenv Python 3.13, 30 runs, `hyperfine --warmup 3`.

**Isolated ceiling** -- pure parser construction, no subcommand dispatched:

```
python -c "from conda.cli.conda_argparse import generate_parser
generate_parser().parse_args(['--version'])"
```

| | `main` | this PR | Δ |
|---|---|---|---|
| wall-clock | 545.0 ± 12.4 ms | 63.5 ± 7.1 ms | **−481.5 ms (8.6×)** |
| `sys.modules` total | 926 | 125 | **−801** |
| `conda.cli.main_*` | 26 | 0 | −26 |
| `conda.*` total | 170 | 11 | −159 |

**Bare module import** (captures the +1.2 ms cost of the 24 module-scope `deprecated.constant(factory=...)` registrations):

```
python -c "import conda.cli.conda_argparse"
```

| | pre-refactor A2/A3 | this PR | Δ |
|---|---|---|---|
| wall-clock | 58.7 ± 6.2 ms | 59.9 ± 4.2 ms | +1.2 ms (within σ) |

**End-to-end commands** (`conda --version`, `conda <cmd> --help`): essentially flat vs. `main`. Once a subcommand is dispatched or top-level help is rendered, `_ensure_plugins_loaded()` runs and pulls in the plugin subcommand modules. The lazy parser still avoids the 19 *other* builtin subcommands, but on this machine their transitive overlap with plugin discovery hides most of that in wall-clock. The isolated ceiling above is the honest upper bound; real-command savings compound once A6 (plugin-skip on activate) and A11 (plugin-skip on `conda run`) land.

### Notes

- Earlier revisions of this PR measured −142 ms / −505 modules against a pre-Track-A baseline. Most of Track A has since merged (A1, A4, A5, A5b, A7, A12, A15, A17, A18, A20a, A20b, A21, A23); none of those merges targeted the eager-subcommand-parser path, so the isolated delta here is bigger now, not smaller. See #15867 note ¹⁵ for the methodology.
- The previous revision registered deprecated re-exports lazily from a custom `__getattr__`. That had a small but real bug -- the first access to each deprecated symbol was silent, because the `_ConstantDeprecationRegistry` only took over *after* the first call had already returned. Migrating to module-scope `deprecated.constant(factory=...)` closes that hole at a noise-level import-time cost (+1.2 ms).

### Checklist - did you ...

- ~Add / update outdated documentation?~ Not applicable.

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -404,11 +404,29 @@ class _LazyParserMap(dict):
         super().__init__()
         self._action = action
 
+    @property
+    def _ready(self):
+        """True once the owning action is initialized and not mid-build."""
+        action = self._action
+        return hasattr(action, "_lazy_loaders") and not getattr(
+            action, "_building", False
+        )
+
     def _resolve(self, key):
         """Ensure *key* is fully loaded (builtin or plugin)."""
+        if not self._ready:
+            return
         self._action._ensure_loaded(key)
         if not super().__contains__(key):
             self._action._ensure_plugins_loaded()
+
+    def _resolve_all(self):
+        """Load every lazy subcommand and all plugins."""
+        if not self._ready:
+            return
+        for name in list(self._action._lazy_loaders):
+            self._action._ensure_loaded(name)
+        self._action._ensure_plugins_loaded()
 
     def __getitem__(self, key):
         self._resolve(key)
@@ -421,6 +439,26 @@ class _LazyParserMap(dict):
     def __contains__(self, key):
         self._resolve(key)
         return super().__contains__(key)
+
+    def items(self):
+        self._resolve_all()
+        return super().items()
+
+    def values(self):
+        self._resolve_all()
+        return super().values()
+
+    def keys(self):
+        self._resolve_all()
+        return super().keys()
+
+    def __iter__(self):
+        self._resolve_all()
+        return super().__iter__()
+
+    def __len__(self):
+        self._resolve_all()
+        return super().__len__()
 
 
 class _LazySubParsersAction(_GreedySubParsersAction):
@@ -437,17 +475,21 @@ class _LazySubParsersAction(_GreedySubParsersAction):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._name_parser_map = _LazyParserMap(self)
-        # choices is the same dict object in argparse's _SubParsersAction
         self.choices = self._name_parser_map
         self._lazy_loaders: dict[str, tuple[str, dict]] = {}
         self._lazy_aliases: dict[str, str] = {}
         self._loading_name: str | None = None
         self._plugins_loaded = False
+        self._building = False
 
     def add_lazy_subcommand(self, name, module_path, help_text, **kwargs):
         """Register a lazy-loaded subcommand with a stub parser."""
         aliases = kwargs.get("aliases", ())
-        stub = super().add_parser(name, help=help_text, **kwargs)
+        self._building = True
+        try:
+            stub = super().add_parser(name, help=help_text, **kwargs)
+        finally:
+            self._building = False
         self._lazy_loaders[name] = (module_path, kwargs)
         for alias in aliases:
             self._lazy_aliases[alias] = name
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -236,6 +236,9 @@ BUILTIN_COMMANDS = {
 }
 """Names (and aliases) of built-in commands; these cannot be overridden by plugin subcommands."""
 
+_PLUGIN_FREE_BUILTIN_COMMANDS = frozenset({"activate", "deactivate", "run"})
+"""Built-in commands that intentionally skip plugin discovery."""
+
 _BUILTIN_SUBCOMMANDS_BY_NAME = {
     name: (module_path, help_text, extra_kwargs)
     for name, module_path, help_text, extra_kwargs in _BUILTIN_SUBCOMMANDS
@@ -355,12 +358,10 @@ class ArgumentParser(ArgumentParserBase):
             action.choices, dict
         ):
             # Unknown command: discover plugins before rejecting so that plugin
-            # subcommands appear in the "choose from" error message.  Must read
-            # from _name_parser_map (not choices) after loading because the
-            # sort-reassignment above disconnects choices from _name_parser_map.
+            # subcommands appear in the sorted "choose from" error message.
             if not isiterable(value) and value not in action._name_parser_map:
                 action._ensure_plugins_loaded()
-            action.choices = dict(sorted(action._name_parser_map.items()))
+                action.choices = dict(sorted(dict.items(action._name_parser_map)))
         elif isinstance(action, _GreedySubParsersAction) and isinstance(
             action.choices, dict
         ):
@@ -580,10 +581,11 @@ class _LazySubParsersAction(_GreedySubParsersAction):
     def __call__(self, parser, namespace, values, option_string=None):
         parser_name = values[0]
         self._ensure_loaded(parser_name)
-        # Always discover plugins when a subcommand is dispatched so that
-        # the builtin-override check in configure_parser_plugins() runs even
-        # when the invoked command is a builtin (e.g. `conda info --help`).
-        self._ensure_plugins_loaded()
+        if parser_name not in _PLUGIN_FREE_BUILTIN_COMMANDS:
+            # Discover plugins when a subcommand is dispatched so that the
+            # builtin-override check in configure_parser_plugins() runs even
+            # when the invoked command is a builtin (e.g. `conda info --help`).
+            self._ensure_plugins_loaded()
         super().__call__(parser, namespace, values, option_string)
 
     def _get_subactions(self):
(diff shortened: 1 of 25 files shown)


conda/cli/conda_argparse.py
```diff
@@ -43,57 +41,177 @@
...
+    # and all subsequent accesses go through the registry and emit the warning.
+    from ..deprecations import deprecated
+
+    if name in _DEPRECATED_CONFIGURE_PARSER_EXPORTS:
+        module_path = _DEPRECATED_CONFIGURE_PARSER_EXPORTS[name]
+        mod = import_module(module_path)
+        val = mod.configure_parser
+        deprecated.constant(
+            "26.5",
+            "27.3",
```
kenodegard:
```suggestion
            "26.9",
            "27.3",
```

conda/cli/conda_argparse.py
```diff
@@ -43,57 +41,177 @@
...
+        _rc_values = {
+            "user_rc_path": user_rc_path,
+            "sys_rc_path": sys_rc_path,
+            "escaped_user_rc_path": user_rc_path.replace("%", "%%"),
+            "escaped_sys_rc_path": sys_rc_path.replace("%", "%%"),
+        }
+        val = _rc_values[name]
+        deprecated.constant(
+            "26.5",
+            "27.3",
```
kenodegard:
```suggestion
            "26.9",
            "27.3",
```

jezdez:
@kenodegard I'd love to get a review from you about this

conda/cli/conda_argparse.py
```diff
@@ -45,82 +48,219 @@
...
+        "conda.cli.main_update",
+        "Update conda packages to the latest compatible version.",
+        {"aliases": ["upgrade"]},
+    ),
+]
+
+BUILTIN_COMMANDS = {name for name, *_ in _BUILTIN_SUBCOMMANDS}
+for _name, _module, _help, _kw in _BUILTIN_SUBCOMMANDS:
+    for _alias in _kw.get("aliases", ()):
+        BUILTIN_COMMANDS.add(_alias)
```
kenodegard:
```suggestion
BUILTIN_COMMANDS = {
    name
    for command, _, _, kwargs in _BUILTIN_SUBCOMMANDS
    for name in (command, *kwargs.get("aliases", ()))
}
```

conda/cli/conda_argparse.py
```diff
@@ -45,82 +48,219 @@
...
+    ("configure_parser_package", "conda.cli.main_package"),
+    ("configure_parser_remove", "conda.cli.main_remove"),
+    ("configure_parser_rename", "conda.cli.main_rename"),
+    ("configure_parser_run", "conda.cli.main_run"),
+    ("configure_parser_search", "conda.cli.main_search"),
+    ("configure_parser_update", "conda.cli.main_update"),
+):
+    deprecated.constant(
+        "26.5",
+        "27.3",
```
kenodegard:
```suggestion
        "27.3",
        "27.9",
```

conda/cli/conda_argparse.py
```diff
@@ -45,82 +48,219 @@
...
+    ),
+    (
+        "escaped_sys_rc_path",
+        _escaped_sys_rc_path,
+        "Use `from conda.base.context import sys_rc_path` and escape locally.",
+    ),
+):
+    deprecated.constant(
+        "26.5",
+        "27.3",
```
kenodegard:
```suggestion
        "27.3",
        "27.9",
```

news/15867-lazy-subcommand-parser
```diff
@@ -0,0 +1,8 @@
+### Enhancements
+
+* Defer loading of subcommand parser modules until the subcommand is actually invoked, reducing the number of modules imported at startup and improving `conda` startup latency.
+
+### Deprecations
+
+* `conda.cli.conda_argparse.configure_parser_*` functions are pending deprecation. Import `configure_parser` directly from the relevant `conda.cli.main_*` module instead.
+* `conda.cli.conda_argparse.user_rc_path`, `sys_rc_path`, `escaped_user_rc_path`, and `escaped_sys_rc_path` are pending deprecation. Use `conda.base.context.user_rc_path` / `sys_rc_path` instead.
```
jaimergp:
```suggestion
### Enhancements

* Defer loading of subcommand parser modules until the subcommand is actually invoked, reducing the number of modules imported at startup and improving `conda` startup latency. (#15868)

### Deprecations

* `conda.cli.conda_argparse.configure_parser_*` re-exported function names are pending deprecation. Import `configure_parser` directly from the relevant `conda.cli.main_*` module instead. (#15868)
* `conda.cli.conda_argparse.user_rc_path`, `sys_rc_path`, `escaped_user_rc_path`, and `escaped_sys_rc_path` are pending deprecation. Use `conda.base.context.user_rc_path` / `sys_rc_path` instead. (#15868)
```

jaimergp:
I can reproduce the timings in the description. Not much of a change in the commands themselves, though. Why do we need to resolve all commands when I'm only invoking e.g. `conda run`? Can't we try blindly first to resolve the builtin command and only if that fails try load all subcommand plugins? Most used subcommands must be the builtins anyway.

Edit: ok, it needs https://github.com/conda/conda/pull/15883 too

conda/cli/conda_argparse.py
```diff
@@ -45,82 +48,219 @@
...
+    ("configure_parser_rename", "conda.cli.main_rename"),
+    ("configure_parser_run", "conda.cli.main_run"),
+    ("configure_parser_search", "conda.cli.main_search"),
+    ("configure_parser_update", "conda.cli.main_update"),
+):
+    deprecated.constant(
+        "26.5",
+        "27.3",
+        _name,
+        factory=_configure_parser_factory(_module_path),
```
kenodegard: Would a lambda be cleaner?

```suggestion
        factory=lambda module_path=_module_path: import_module(module_path).configure_parser,
```

conda/cli/conda_argparse.py
```diff
@@ -45,82 +48,219 @@
...
+        "26.5",
+        "27.3",
+        _name,
+        factory=_factory,
+        addendum=_addendum,
+    )
+del _name, _factory, _addendum
+
+
+_BUILTIN_SUBCOMMANDS: list[tuple[str, str, str, dict]] = [
```
kenodegard: This is duplicating the help text defined in the individual `configure_parser`. The `configure_parser` should lookup their help from here.

conda/cli/conda_argparse.py
jezdez: Applied in e45fdd057.

conda/cli/conda_argparse.py
jezdez: Applied. These now use 27.3 and 27.9.

conda/cli/conda_argparse.py
jezdez: Applied. These now use 27.3 and 27.9.

news/15867-lazy-subcommand-parser
jezdez: Applied, including the PR references.

conda/cli/conda_argparse.py
jezdez: Applied. Switched to the lambda factory.

conda/cli/conda_argparse.py
jezdez: Applied. Each built-in `configure_parser` now reads the centralized help text.

jezdez:
@kenodegard @jaimergp The review feedback is applied and this is rebased onto current main at 4e6e6752b. Fresh CI is running. Could you take another look?

kenodegard (approved):

PR #15877, opened 2026-04-09
Skip plugin loading on shell activate path (A6)

jezdez:
### Description

Skip plugin pre/post command hooks in `_Activator.execute()` when the plugin manager has not been initialized yet.

On the shell activate path (`conda shell.posix activate`), nothing triggers plugin loading, so the hooks are safely skipped — avoiding ~430 module imports (~240 ms) on every terminal prompt. When the plugin manager was already initialized by another code path (e.g. `main_subshell`), hooks fire as before.

The check uses `get_plugin_manager.cache_info().currsize` to inspect the `functools.cache` without triggering initialization.

| Metric | Before | After | Saved |
|---|---|---|---|
| Modules loaded (`conda shell.posix activate`) | ~834 | ~405 | −429 |
| Import time (`-X importtime`) | ~370 ms | ~130 ms | −240 ms |

Part of #15867.

### Checklist - did you ...

- [ ] ~Add / update outdated documentation?~ (no docs changes needed)

### DIFF
--- a/news/15867-skip-plugin-activate
+++ b/news/15867-skip-plugin-activate
@@ -1,6 +1,9 @@
 ### Enhancements
 
-* Skip plugin loading on the shell activate path (`conda shell.* activate|deactivate|hook|reactivate`). Plugin pre/post command hooks are only invoked when the plugin manager was already initialized by another code path. This avoids importing ~430 modules (~240 ms) on every terminal prompt. (#15867)
+* Skip plugin loading on the shell activate fast path (`conda shell.* activate | deactivate | reactivate | hook`).
+  Plugin `pre_commands` / `post_commands` hooks are only invoked when the plugin manager was already initialized
+  by another code path. Saves ~430 imports and ~240 ms on every terminal prompt. Plugin authors should not rely
+  on these hooks firing on activation; see the note in `conda.plugins.hookspec`. (#15867)
 
 ### Bug fixes
 
--- a/news/15867-skip-plugin-activate
+++ b/news/15867-skip-plugin-activate
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Skip plugin loading on the shell activate path (`conda shell.* activate|deactivate|hook|reactivate`). Plugin pre/post command hooks are only invoked when the plugin manager was already initialized by another code path. This avoids importing ~430 modules (~240 ms) on every terminal prompt. (#15867)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/15867-skip-plugin-activate
+++ b/news/15867-skip-plugin-activate
@@ -1,9 +1,8 @@
 ### Enhancements
 
-* Skip plugin loading on the shell activate fast path (`conda shell.* activate | deactivate | reactivate | hook`).
-  Plugin `pre_commands` / `post_commands` hooks are only invoked when the plugin manager was already initialized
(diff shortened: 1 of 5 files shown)


news/15867-skip-plugin-activate
```diff
@@ -0,0 +1,22 @@
+### Enhancements
+
+* Skip plugin loading on the shell activate fast path (`conda shell.* activate | deactivate | reactivate | hook`).
+  Plugin `pre_commands` / `post_commands` hooks are only invoked when the plugin manager was already initialized
+  by another code path. Saves ~430 imports and ~240 ms on every terminal prompt. Plugin authors should not rely
+  on these hooks firing on activation; see the note in `conda.plugins.hookspec`. (#15867)
```
danyeaw: Hey @jezdez, I don't see changes to the `hookspec`, but I think this would be a great thing to document there.

tests/test_activate.py
```diff
@@ -2268,3 +2268,77 @@ def test_activate_default_env(activator_cls, monkeypatch, conda_cli, tmp_path):
...
+    def conda_post_commands(self):
+        yield CondaPostCommand(
+            name="tracking-post",
+            action=self._post_action,
+            run_for={"activate", "deactivate", "reactivate", "hook"},
+        )
+
+
+@pytest.fixture()
+def _clear_plugin_manager_cache():
```
danyeaw: Could we use the existing `clear_plugin_manager_cache` fixture instead of creating a local one here?

tests/test_activate.py
```diff
@@ -2268,3 +2268,77 @@ def test_activate_default_env(activator_cls, monkeypatch, conda_cli, tmp_path):
...
+    assert get_plugin_manager.cache_info().currsize == 0
+
+    activator = PosixActivator([command])
+    activator.execute()
+
+    assert get_plugin_manager.cache_info().currsize == 0
+
+
+@pytest.mark.parametrize("command", ["activate", "deactivate", "reactivate", "hook"])
+def test_plugin_hooks_invoked_when_manager_loaded(
```
danyeaw: I don't think this is testing anything extra that `test_pre_post_command_invoked` doesn't already test.

danyeaw (commented):
Hey @jezdez, thanks for helping to improve this - this will be a nice speedup!

I just had a couple of comments about the implementation. Do you think it would be worthwhile adding this to the `test_startup_benchmarks` as `hook` to ensure the max modules stays low in the future?

news/15867-skip-plugin-activate
jezdez: Done in 7ac493663: I added notes to both `conda_pre_commands` and `conda_post_commands` in `conda.plugins.hookspec`, then tightened the news fragment so the hook behavior is described directly there too.

tests/test_activate.py
jezdez: Done in 7ac493663: the test now uses the shared `clear_plugin_manager_cache` fixture instead of defining a local fixture.

tests/test_activate.py
jezdez: Done in 7ac493663: I removed the extra tracker plugin and redundant `test_plugin_hooks_invoked_when_manager_loaded` case. The existing `test_pre_post_command_invoked` continues to cover the already-loaded plugin-manager path.

jezdez:
Also addressed the review-summary suggestion about startup benchmarks in 7ac493663: `tests/cli/test_startup_benchmarks.py` now includes a `shell_hook` module-count budget that runs `main_sourced("shell.posix", "hook")`.

danyeaw (approved):
Thanks @jezdez!

PR #15878, opened 2026-04-09
Fast path for conda --version / -V (A7)

jezdez:
### Description

Short-circuit in `main()` when the first argument is `-V` or `--version`, printing the version and returning `0` immediately. This avoids loading the argument parser, context, and plugin system entirely.

| Metric | Before | After | Saved |
|---|---|---|---|
| Modules loaded | 990 | 103 | −887 |
| Wall clock (`hyperfine`, min) | 473 ms | 38 ms | −435 ms |

**Why this matters:** `conda --version` is one of the most frequently run commands — CI pipelines, shell prompts, tooling integrations, and users checking their install all hit it regularly. Taking ~470 ms for something that just prints a string is noticeable, especially when competing tools (`uv --version`, `pixi --version`) respond in ~12 ms. The change is five lines with no behavioral difference.

Part of #15867.

### Checklist - did you ...

- [ ] ~Add / update outdated documentation?~ (no docs changes needed)

### DIFF
--- a/conda/cli/main.py
+++ b/conda/cli/main.py
@@ -97,6 +97,13 @@ def main(*args, **kwargs):
     args = args or sys.argv[1:]  # drop executable/script
     args = tuple(ensure_text_type(s) for s in args)
 
+    # Fast path: print version without loading the parser or plugin system.
+    if args and args[0] in ("-V", "--version"):
+        from .. import __version__
+
+        print(f"conda {__version__}")
+        return 0
+
     if args and args[0].strip().startswith("shell."):
         main = main_sourced
     else:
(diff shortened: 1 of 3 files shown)


conda/cli/main.py
```diff
@@ -97,6 +97,13 @@ def main(*args, **kwargs):
     args = args or sys.argv[1:]  # drop executable/script
     args = tuple(ensure_text_type(s) for s in args)
 
+    # Fast path: print version without loading the parser or plugin system.
+    if args and args[0] in ("-V", "--version"):
```
dholth, line 101: I see that conda behaves like this today, I can type `conda -V other --help random` and only the version comes out. `grep` e.g. appears to do the same.

dholth (approved):

Issue #14500, opened 2025-01-17
"conda env list" major slowdown in version 24.11.3 vs 24.5.0


AGenchev:
### Checklist

- [x] I added a descriptive title
- [x] I searched for other speed issues and didn't find a duplicate

### What happened?

It is on a very powerful hardware, no load.

```
$ time conda env list
# conda environments:
#
base                   /home/gele/miniconda3
py310                  /home/gele/miniconda3/envs/py310

real    0m2.440s
user    0m0.322s
sys     0m1.561s

$ conda --version
conda 24.11.3
```
~ 2.5 sec waiting for env list vs (older, **downgraded** on the same server conda):
```
$time conda env list
# conda environments:
#
base                     /home/gele/miniconda3
py310                    /home/gele/miniconda3/envs/py310

real    0m0.198s
user    0m0.153s
sys     0m0.044s
$ conda --version
conda 24.5.0
```
downgraded takes ~ 0.2s. So the slowdown is 10x times.

### Debug

```shell
# conda environments:
#
base                   /home/gele/miniconda3
py310                  /home/gele/miniconda3/envs/py310
```

### Conda info

```shell
conda info

     active environment : None
            shell level : 0
       user config file : /home/gele/.condarc
 populated config files : /home/gele/miniconda3/.condarc
          conda version : 24.11.3
    conda-build version : 25.1.1
         python version : 3.12.8.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=zen4
                          __conda=24.11.3=0
                          __cuda=12.7=0
                          __glibc=2.40=0
                          __linux=6.12.9=0
                          __unix=0=0
       base environment : /home/gele/miniconda3  (writable)
      conda av data dir : /home/gele/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/gele/miniconda3/pkgs
                          /home/gele/.conda/pkgs
       envs directories : /home/gele/miniconda3/envs
                          /home/gele/.conda/envs
               platform : linux-64
[...]

PR #15879, opened 2026-04-09
Defer heavy imports in notices/core.py and main_env.py (A4)

jezdez:
### Description

Defer module-level imports in `conda/notices/core.py` and `conda/cli/main_env.py` so that importing these modules no longer pulls in the full notices dependency tree (`requests`, `concurrent.futures`, `models.channel`, etc.).

**`notices/core.py`**: moves `get_channel_objs`, `cache`, `fetch`, `views`, and `ChannelNoticeResultSet` from module-level into the functions that actually use them (`retrieve_notices`, `display_notices`, `is_channel_notices_cache_expired`, and the `notices` decorator error handler). The `@notices` decorator remains unchanged on all `main_*.py` files — it's now cheap to import because the heavy work is deferred to when notices are actually fetched.

**`main_env.py`**: moves the `from . import main_export` from module-level into `configure_parser()`, alongside the other `main_env_*` imports that were already deferred.

This is a follow-up to the lazy subcommand parser loading in #15868 (A2/A3). Once subcommands load individually, A4 cuts the per-subcommand module cost significantly:

| Module | Baseline | With A4 | Savings |
|---|---|---|---|
| `main_create` | +560 modules, 183 ms | +173 modules, 65 ms | −387 modules, −118 ms |
| `main_install` | +560 modules, 176 ms | +173 modules, 69 ms | −387 modules, −107 ms |
| `main_env` | +647 modules, 282 ms | +32 modules, 10 ms | −615 modules, −272 ms |

Without A2/A3 (current eager `generate_parser()` path), the savings are negligible (~4 modules) because `conda_argparse` pulls everything in regardless.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_env.py
+++ b/conda/cli/main_env.py
@@ -6,8 +6,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from . import main_export
-
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace, _SubParsersAction
 
@@ -19,6 +17,7 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         main_env_list,
         main_env_remove,
         main_env_update,
+        main_export,
     )
 
     p = sub_parsers.add_parser(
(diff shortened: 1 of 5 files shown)


dholth (approved):
lgtm

PR #15880, opened 2026-04-09
Defer heavy imports in exceptions.py to error paths (A8)

jezdez:
### Description

Move `models.channel.Channel`, `auxlib.logz.stringify`, `common.io.dashlist`, `common.iterators.groupby_to_dict`, and `serialize.json.dumps` from module-level to the methods that use them in `conda/exceptions.py`.

These imports are only needed when formatting specific error messages, not when defining exception classes. Deferring them removes 71 modules (~23 ms) from any code path that imports `conda.exceptions`.

The heaviest: `models.channel` (+49 modules, 16.5 ms) and `common.io` (+13 modules via `concurrent.futures`/`threading`). `Channel` type hints moved to `TYPE_CHECKING`.

Complementary to A2/A3 (#15868): the biggest win comes from lazy parser loading preventing `exceptions.py` from being imported during startup at all, but deferring within `exceptions.py` itself prevents the cascade when subcommands do import it.

### Checklist - did you ...


dholth:
I like this idea. A program-stopping error can take as much time as it likes, if it speeds up the rest.

tests/test_deferred_exceptions_imports.py
```diff
@@ -0,0 +1,42 @@
...
+
+import pytest
+
+
+@pytest.mark.parametrize(
+    "module",
+    [
+        "conda.models.channel",
+        "conda.auxlib.logz",
+        "conda.common.iterators",
```
danyeaw, line 23:
```suggestion
        "conda.models.channel",
        "conda.auxlib.logz",
        "conda.common.iterators",
        "conda.common.io",
        "conda.common.serialize.json",
```

Hey @jezdez, should we add the other two deferred imports here, or was the a reason to exclude them like test performance?

danyeaw (approved):
Hey @jezdez, this looks great! One minor nit, I noticed in the PR description that it mentions JSONDecodeError, but I don't think that is in scope of the PR currently (only json_dumps).

ryanskeith (approved):
Looks good to me, let's merge it.  

One point to think about going forward is if we want to keep the test long term.  Guardrails are great but is it worth the cost of time and upkeep?

PR #15881, opened 2026-04-09
Defer `concurrent.futures` in `common/io.py` (A9)

jezdez:
### Description

Use PEP 562 `__getattr__` to lazily define `DummyExecutor`, `ThreadLimitedThreadPoolExecutor`, and `as_completed` in `conda/common/io.py`.

`concurrent.futures` costs ~45 modules to import. This module is loaded whenever `common/io.py` is imported, even for lightweight utilities like `dashlist` or `time_recorder`. Deferring the executor classes to first access means only code that actually uses parallel I/O pays the cost.

### Checklist - did you ...

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -9,7 +9,6 @@ import os
 import signal
 import sys
 from collections import defaultdict
-from concurrent.futures import Executor, Future, ThreadPoolExecutor, as_completed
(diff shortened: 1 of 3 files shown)


ryanskeith (approved):
This looks like a good improvement.

PR #15882, opened 2026-04-09
Defer ruamel.yaml import in serialize/yaml.py (A10)

jezdez:
### Description

Move `import ruamel.yaml` and the `CondaYAMLRepresenter` class definition inside cached factory functions (`_yaml()` / `_representer()`) in `conda/common/serialize/yaml.py`. `YAMLError` becomes a lazy module attribute via `__getattr__` (PEP 562).

This removes 32 `ruamel.yaml` modules from any code path that imports `common/serialize/yaml` without actually parsing YAML. Wall-clock impact on warm-cache macOS is negligible, but the module-count reduction helps:

- Cold-cache scenarios (CI, containers, first invocation)
- Module-count budget compliance (CodSpeed guardrails from #15850)
- Reducing import-time side-effect surface area

`CondaYAMLRepresenter` was a module-level name. To avoid silently breaking external consumers that import it directly, it is registered via `deprecated.constant(..., factory=...)` (added in #15925) and remains resolvable as `conda.common.serialize.yaml.CondaYAMLRepresenter`. First access emits `PendingDeprecationWarning` (removal in 27.3). The class is built once by the cached `_representer()` factory, so identity stays stable across accesses. Callers should use `conda.common.serialize.yaml.write` / `read`, or subclass `ruamel.yaml.representer.RoundTripRepresenter` directly.

### Checklist - did you ...

- [x] ~Add / update outdated documentation?~ (no user-facing docs needed)

### DIFF
--- a/conda/common/serialize/yaml.py
+++ b/conda/common/serialize/yaml.py
@@ -9,15 +9,19 @@ from io import StringIO
 from pathlib import Path
 from typing import TYPE_CHECKING, overload
 
+from ...deprecations import deprecated
+
 if TYPE_CHECKING:
     from io import IO
     from typing import Any
+
+    import ruamel.yaml
 
     from ..path import PathType
 
 
 @cache
-def _yaml():
+def _representer() -> type[ruamel.yaml.representer.RoundTripRepresenter]:
     import ruamel.yaml
 
     class CondaYAMLRepresenter(ruamel.yaml.representer.RoundTripRepresenter):
@@ -41,9 +45,15 @@ def _yaml():
             )
 
     CondaYAMLRepresenter.add_representer(None, CondaYAMLRepresenter.default)
+    return CondaYAMLRepresenter
+
+
+@cache
+def _yaml() -> ruamel.yaml.YAML:
+    import ruamel.yaml
 
     parser = ruamel.yaml.YAML(typ="rt")
-    parser.Representer = CondaYAMLRepresenter
+    parser.Representer = _representer()
     parser.indent(mapping=2, offset=2, sequence=4)
     parser.default_flow_style = False
     parser.sort_base_mapping_type_on_output = False
@@ -123,9 +133,30 @@ def loads(s: str) -> Any:
     return read(text=s)
 
 
-def __getattr__(name: str):
+# ``YAMLError`` is re-exported lazily via PEP-562 so that merely importing
+# this module does not pull in ``ruamel.yaml``. ``deprecated.constant``
+# below installs its own module-level ``__getattr__`` and wires this one
+# up as its fallback, so both names resolve correctly.
+def __getattr__(name: str) -> Any:
     if name == "YAMLError":
         import ruamel.yaml
 
         return ruamel.yaml.YAMLError
+
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+# ``CondaYAMLRepresenter`` used to be a module-level class. Keep it
+# importable for external consumers but emit a deprecation warning on
+# access. Passing ``factory=`` defers ``_representer()`` (which imports
+# ``ruamel.yaml``) until the deprecated symbol is actually used.
+deprecated.constant(
+    "26.9",
+    "27.3",
+    "CondaYAMLRepresenter",
+    factory=_representer,
+    addendum=(
+        "Use `conda.common.serialize.yaml.write`/`read` or subclass "
+        "`ruamel.yaml.representer.RoundTripRepresenter` directly."
+    ),
+)
--- a/conda/common/serialize/yaml.py
+++ b/conda/common/serialize/yaml.py
@@ -4,13 +4,10 @@
 
 from __future__ import annotations
 
-from enum import Enum
 from functools import cache
 from io import StringIO
 from pathlib import Path
 from typing import TYPE_CHECKING, overload
-
-import ruamel.yaml
 
 if TYPE_CHECKING:
     from io import IO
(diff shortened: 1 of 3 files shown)


conda/common/serialize/yaml.py
```diff
@@ -4,45 +4,44 @@
...
-import ruamel.yaml
-
 if TYPE_CHECKING:
     from io import IO
     from typing import Any
 
     from ..path import PathType
 
 
-class CondaYAMLRepresenter(ruamel.yaml.representer.RoundTripRepresenter):
```
kenodegard, line 22: Do we need to deprecate this?

jezdez: Done in 9a71ae516. `CondaYAMLRepresenter` is still importable from this module; accessing it now emits a `PendingDeprecationWarning` with removal in 27.3. Class lives in a cached `_representer_cls()` so identity is stable across accesses. `ruamel.yaml` is still deferred.

conda/common/serialize/yaml.py
```diff
@@ -4,47 +4,56 @@
...
+                    return self.represent_data(method())
 
+            # mirror JSON behavior
+            raise TypeError(
+                f"Object of type {data.__class__.__name__} is not YAML serializable"
+            )
 
-CondaYAMLRepresenter.add_representer(None, CondaYAMLRepresenter.default)
+    CondaYAMLRepresenter.add_representer(None, CondaYAMLRepresenter.default)
+    return CondaYAMLRepresenter
```
kenodegard, line 48: It would be cleaner to move this into a private submodule that we can lazy import

kenodegard: but I see that this pattern already mirrors what we do in with caching `_yaml` etc. so lets just leave this as proposed

tests/common/test_yaml_deferred_imports.py
```diff
@@ -0,0 +1,97 @@
...
+    result = subprocess.run(
+        [sys.executable, "-c", snippet],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    assert result.returncode == 0, f"Probe failed: {result.stderr}"
+    assert result.stdout.strip() == "False", (
+        "ruamel.yaml was loaded at import time; it should be deferred to _yaml()"
+    )
```
kenodegard: Simplify to only rely on assert + returncode?

```suggestion
    snippet = textwrap.dedent("""\
        import sys
        import conda.common.serialize.yaml
        assert "ruamel.yaml" not in sys.modules
    """)
    result = subprocess.run(
        [sys.executable, "-c", snippet],
        timeout=30,
    )
    assert result.returncode == 0, f"Probe failed: {result.stderr}"
```

tests/common/test_yaml_deferred_imports.py
```diff
@@ -0,0 +1,97 @@
...
+        print("ok")
+    """)
+    result = subprocess.run(
+        [sys.executable, "-c", snippet],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    assert result.returncode == 0, f"Probe failed: {result.stderr}"
+    assert result.stdout.strip() == "ok"
```
kenodegard:
```suggestion
    snippet = textwrap.dedent("""\
        from conda.common.serialize.yaml import dumps, loads
        data = {"key": "value", "nested": [1, 2, 3]}
        text = dumps(data)
        assert "key" in text
        loaded = loads(text)
        assert loaded["key"] == "value"
        assert loaded["nested"] == [1, 2, 3]
    """)
    result = subprocess.run(
        [sys.executable, "-c", snippet],
        timeout=30,
    )
    assert result.returncode == 0, f"Probe failed: {result.stderr}"
```

tests/common/test_yaml_deferred_imports.py
```diff
@@ -0,0 +1,97 @@
...
```
kenodegard:
```suggestion
    snippet = textwrap.dedent("""\
        from conda.common.serialize import yaml
        assert yaml.YAMLError is not None
    """)
    result = subprocess.run(
        [sys.executable, "-c", snippet],
        timeout=30,
    )
    assert result.returncode == 0, f"Probe failed: {result.stderr}"
```

tests/common/test_yaml_deferred_imports.py
```diff
@@ -0,0 +1,97 @@
...
+        cls = yaml.CondaYAMLRepresenter
+
+    assert cls is not None
+    assert any(
+        issubclass(w.category, (DeprecationWarning, PendingDeprecationWarning))
+        and "CondaYAMLRepresenter" in str(w.message)
+        for w in caught
+    ), (
+        f"expected deprecation warning, got: {[(w.category, str(w.message)) for w in caught]}"
+    )
```
kenodegard:
```suggestion
    with warnings.deprecated_call():
        yaml.CondaYAMLRepresenter
```

tests/common/test_yaml_deferred_imports.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_conda_yaml_representer_stable_identity() -> None:
+    """Repeated accesses return the same class object (cached)."""
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        cls_a = yaml.CondaYAMLRepresenter
+        cls_b = yaml.CondaYAMLRepresenter
+
+    assert cls_a is cls_b
```
kenodegard:
```suggestion
    with warnings.deprecated_call():
        cls_a = yaml.CondaYAMLRepresenter
        cls_b = yaml.CondaYAMLRepresenter

    assert cls_a is cls_b
```

kenodegard (approved):
Minor suggestions for tests

tests/common/test_yaml_deferred_imports.py
jezdez: Addressed in https://github.com/conda/conda/commit/4b37fd94c43522b2244e7d0f1a0ae51f1dd044d2. I moved the check into the subprocess assert, but kept `capture_output=True`/`text=True` so failure output remains useful.

tests/common/test_yaml_deferred_imports.py
jezdez: Addressed in https://github.com/conda/conda/commit/4b37fd94c43522b2244e7d0f1a0ae51f1dd044d2. I moved the checks into the subprocess asserts, but kept `capture_output=True`/`text=True` so failure output remains useful.

tests/common/test_yaml_deferred_imports.py
jezdez: Addressed in https://github.com/conda/conda/commit/4b37fd94c43522b2244e7d0f1a0ae51f1dd044d2. I moved the lazy `YAMLError` check into the subprocess assert, while keeping stderr capture for failures.

tests/common/test_yaml_deferred_imports.py
jezdez: Addressed in https://github.com/conda/conda/commit/4b37fd94c43522b2244e7d0f1a0ae51f1dd044d2. This now uses `pytest.deprecated_call()`, which matches the existing conda test pattern.

tests/common/test_yaml_deferred_imports.py
jezdez: Addressed in https://github.com/conda/conda/commit/4b37fd94c43522b2244e7d0f1a0ae51f1dd044d2. This now uses `pytest.deprecated_call()` while preserving the identity assertion.

PR #15883, opened 2026-04-09
Skip plugin hooks in `do_call()` when manager not loaded (A11)

jezdez:
### Description

Apply the same `cache_info()` guard pattern from A6 (#15877) to `do_call()` in `conda_argparse.py`. When the plugin manager has not been loaded (i.e. `get_plugin_manager` was never called), skip `invoke_pre/post_commands` rather than triggering full plugin discovery.

This is effective when combined with lazy subcommand parser loading (A2/A3, #15868): once `generate_parser()` no longer triggers plugin discovery, commands like `conda run` reach `do_call()` without the plugin manager loaded. Skipping the hooks avoids ~430 modules and ~240 ms of unnecessary plugin loading.

PoC measurements showed `conda run` dropping from 831 modules (352 ms) to 249 modules (117 ms) — a 3.0x speedup.

**Depends on:** A2/A3 (#15868) for the guard to be meaningful. Without lazy parser loading, `generate_parser()` still loads the plugin manager before `do_call()` runs.

Part of #15867. See also #14993.

### Checklist - did you ...

### DIFF
--- a/conda/cli/conda_argparse.py
+++ b/conda/cli/conda_argparse.py
@@ -185,9 +185,23 @@ def do_call(args: argparse.Namespace, parser: ArgumentParser):
         module = import_module(module_name)
         command = module_name.split(".")[-1].replace("main_", "")
 
-        context.plugin_manager.invoke_pre_commands(command)
+        # Only invoke plugin hooks when the plugin manager has already been
+        # loaded.  With lazy subcommand parser loading (A2/A3), commands like
+        # `conda run` reach do_call() without triggering plugin discovery,
+        # so accessing context.plugin_manager here would load all plugins
+        # (~430 modules, ~240 ms) for commands that never use them.
+        from ..plugins.manager import get_plugin_manager
+
+        try:
+            plugins_loaded = get_plugin_manager.cache_info().currsize > 0
+        except (AttributeError, TypeError):
+            plugins_loaded = True
+
+        if plugins_loaded:
+            context.plugin_manager.invoke_pre_commands(command)
         result = getattr(module, func_name)(args, parser)
-        context.plugin_manager.invoke_post_commands(command)
+        if plugins_loaded:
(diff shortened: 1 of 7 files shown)


conda/cli/conda_argparse.py
```diff
@@ -203,9 +203,23 @@ def do_call(args: argparse.Namespace, parser: ArgumentParser):
...
-        context.plugin_manager.invoke_pre_commands(command)
+        # Only invoke plugin hooks when the plugin manager has already been
+        # loaded.  With lazy subcommand parser loading (A2/A3), commands like
+        # `conda run` reach do_call() without triggering plugin discovery,
+        # so accessing context.plugin_manager here would load all plugins
+        # (~430 modules, ~240 ms) for commands that never use them.
+        from ..plugins.manager import get_plugin_manager
+
+        try:
+            plugins_loaded = get_plugin_manager.cache_info().currsize > 0
```
jaimergp: Neat

jaimergp:
Two questions about the "no guarantees for pre/post-commands":

- Shall we explicitly ban some commands from the pre- and post-command hooks? `run`, `activate`, `deactivate`?
- Shall we extend the Subcommand API so plugins can declare themselves as eligible for pre and post commands?

I think it's not clear which commands do work with pre/post-commands now.

conda/cli/conda_argparse.py
jezdez: Thanks!

jezdez:
@jaimergp Made this explicit in 0805b48ac. Run and shell activation commands never invoke pre/post hooks. I do not think another Subcommand flag is needed. `run_for` already declares eligible commands. The parser-side discovery skip is in #15868.

jezdez:
@jaimergp This is rebased onto current main at a88c15b6c, with the hook contract documented as discussed. Fresh CI is running. Could you take another look?

PR #15886, opened 2026-04-09
Speed up `_expand_search_path` by ~5x (A13)

jezdez:
### Description

Cut the cost of `Configuration._expand_search_path()` (the filesystem walk that runs on every `reset_context()`) by ~5x. Part of #15867 (A13).

cProfile on a realistic 27-entry search path showed ~85% of the time was spent inside `custom_expandvars()`, not the filesystem walk. Each call built `{**mapping, **kwargs}`, and when `mapping is os.environ` the splat iterates every env var and round-trips its bytes value, even when the template has no `$` or `%`.

Changes in `conda/common/configuration.py`:

- `custom_expandvars()`: short-circuit when the template has no `$` or `%`; drop the `{**mapping, **kwargs}` splat and look `kwargs`/`mapping` up lazily per regex match.
- `_expand_search_path()`: one `path.stat()` + `stat.S_ISREG`/`stat.S_ISDIR` dispatch instead of `Path.is_file() + Path.is_dir()`; `os.scandir()` for directory children so `DirEntry.is_file()` reuses the `d_type` bits from `getdents64()` on Linux/macOS/NTFS.
- `_search_path` is stored as a `tuple` instead of a lazy generator so callers can iterate it more than once.

### Measured impact

Micro-benchmark on a 27-entry search path (2 condarc files + 2 `condarc.d/` dirs with 10 yaml files each + the default `SEARCH_PATH`), 5000 iters x 3 runs, 50-iter warmup, `devenv-3.13-miniforge`:

| version | avg `_expand_search_path` | delta |
| --- | --- | --- |
| `origin/main` | 1327 - 1383 us | - |
| this branch   | 263 - 269 us   | ~-1.09 ms / call (~5.1x) |

`conda info --json` (hyperfine, 10 warmup + 40 runs) is within noise (825.0 +/- 12.4 ms vs 828.1 +/- 10.5 ms on `main`) because `conda info` is dominated by work outside `reset_context()`. The win shows up on pure-startup paths (`conda --version`, `conda activate`) and compounds across the ~2 invocations of `_expand_search_path` per process. `custom_expandvars()` is also called from other code paths.

### Tests

`tests/common/test_configuration.py`:

- `_search_path` is a tuple, not a generator.
- `_set_search_path` reloads `raw_data` on repeat calls (covers the `Configuration.__init__` clears-then-reloads flow on every `reset_context()`).
- `_set_search_path` re-expands when called with a different path.
- Empty search path yields an empty tuple.
- `test_set_search_path_reflects_fs_changes` parametrized over `file-created` / `dir-child-added` / `file-removed`.
- Existing `test_custom_expandvars` continues to cover the lazy lookup against `os.path.expandvars`.

<details>
<summary>History</summary>

An earlier revision cached the expansion on the instance keyed on `(search_path, kwargs)`. That returned stale expansions whenever a condarc file appeared, disappeared, or moved between calls with identical arguments, breaking `test_conda_config_validate_sslverify_truststore` and `test_conda_config_show_includes_plugin_settings` when they wrote a condarc into `$CONDA_PREFIX` after `reset_context()` had already warmed the cache. Pinning the cache with a `stat()`-fingerprint is hard to get right across mtime granularity, atomic renames, and symlink swaps, so the cache was dropped and the per-call work made cheap enough not to need one.

</details>

### Checklist - did you ...


jezdez:
Audit-driven fix pushed (84a776b0f):

Two issues with the original implementation:

1. **Cache hit skipped `_set_raw_data`.** `Configuration.__init__` clears `raw_data = {}` before calling `_set_search_path`. On a cache hit the early `return self` meant `raw_data` stayed empty, so anything that relied on condarc values after `reset_context()` would silently read defaults.

2. **Single-slot cache never actually hit in the real flow.** `Context.__init__` calls `_set_search_path` twice per init with *different* args — first `()` from `Configuration.__init__`, then `SEARCH_PATH + CONDA_PREFIX=…` from `Context.__init__` itself. The two calls kept overwriting each other's single cached entry, so on the next `reset_context()` both keys missed again.

Fix:
- Cache only the `_expand_search_path` result, keyed by `(search_path, kwargs)` in a dict.
- Always run `_set_raw_data` + `_reset_cache`.
- The dict lives on the instance and survives `Configuration.__init__` (which only clears `raw_data` / `_cache_`).

Verified with a real `reset_context()` loop: initial 2 expansions, 1st reset 1, subsequent resets 0. Also added tests for:

- Cache hit correctly reloads `raw_data`.
- Cache survives `Configuration.__init__` re-invocation.
- Both distinct keys from `Context.__init__`'s two calls are cached independently.

tests/common/test_configuration.py
```diff
@@ -874,6 +874,127 @@ def test_expand_search_path(tmp_path):
...
 
 
+@pytest.fixture
+def minimal_config_class():
+    """Configuration subclass with a single ``channels`` parameter."""
+
+    class MinimalConfig(Configuration):
+        channels = ParameterLoader(SequenceParameter(PrimitiveParameter("")))
+
+    return MinimalConfig
```
kenodegard, line 884: Does this have to be a fixture?

kenodegard (approved):

jezdez (commented):
(removed; replied inline on Ken's comment)

tests/common/test_configuration.py
jezdez: Not really - it was a leftover habit from splitting the tests out. Promoted to a module-level MinimalConfig class in https://github.com/conda/conda/commit/0bf1159cf101d5301d33b0879dec8b65b9753d73, all five call sites updated.

PR #15888, opened 2026-04-09
Make category_map a class-level constant (A15)

jezdez:
### Description

`context.category_map` was a `@memoizedproperty` that assembled the full parameter-to-category mapping dict on first access and cached it. The dict is large (~80 keys across 9 categories) but entirely static — it does not depend on any instance state, is never overridden by subclasses (no subclasses of `Context` exist), and is never mutated after creation.

Per review feedback from @dholth: replaced the `@memoizedproperty` with a `ClassVar[dict[str, tuple[str, ...]]]` class attribute, making it a plain class-level constant. This avoids both the property overhead and the memoization machinery entirely.

**Measured impact:** eliminates repeated dict construction and property dispatch on every help / `--describe` invocation. Part of #15867.

### Checklist - did you ...


conda/base/context.py
```diff
@@ -1279,7 +1279,7 @@ def environment_settings(self) -> dict[str, Any]:
         """Returns a dict of environment related settings"""
         return {key: getattr(self, key) for key in self.environment_context_keys}
 
-    @property
+    @memoizedproperty
```
dholth: This module also imports https://docs.python.org/3/library/functools.html#functools.cached_property which was fixed in Python 3.12 ; is there another reason why we can't always use cached_property?

I'm a little surprised that this one isn't a static property not a function. Does it get mutated or overwritten by a subclass?

Example `category_map = dict()` instead of `def category_map(): return dict()`

dholth (approved):
Always found this part of the code a bit weird (you added an option, whoops, warning about putting it in a category) but this change is OK. Interesting `ClassVar` type.

news/15867-memoize-category-map
```diff
@@ -0,0 +1,5 @@
+### Enhancements
+
```
dholth, line 2: When we publish the release, we should sort all of the programmer-related items after all of the ones that the user might have to think about; and separate them with a double space.

PR #15889, opened 2026-04-09
Start ContextStack with one slot instead of three (A17)

jezdez:
### Description

`ContextStack.__init__()` pre-allocated three `ContextStackObject` instances at startup. In practice the common path (a single `conda` invocation) never pushes the stack past depth 1. Those two extra objects are allocated, initialised with `SEARCH_PATH`, and then immediately forgotten.

This change starts with a single slot. The existing doubling strategy (`extend` by `len(self._stack)`) already handles arbitrarily deep nesting correctly — no behaviour changes for callers.

**Measured impact:** eliminates two `ContextStackObject.__init__()` calls (each of which calls `reset_context()`) at module import time. Part of #15867.

### Checklist - did you ...

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -2111,16 +2111,15 @@ class ContextStackObject:
 
 class ContextStack:
     def __init__(self):
-        self._stack = [ContextStackObject() for _ in range(3)]
+        self._stack = [ContextStackObject()]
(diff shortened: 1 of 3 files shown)


dholth (approved):
LGTM

PR #15890, opened 2026-04-09
Pre-compile hot-path regexes in match_spec and activate (A18)

jezdez:
### Description

Several regexes in `MatchSpec._parse_spec_str()` and `_parse_version_plus_build()` were passed as raw strings to `re.match()` / `re.search()` / `re.finditer()` on every call. Python's `re` module caches compiled patterns internally, but that cache has a fixed size (512 entries by default in CPython) and the lookup still incurs hashing and dict-access overhead on every hit.

This change extracts the five constant patterns in `match_spec.py` and the two constant patterns in `activate.py` as module-level `re.compile()` objects. The patterns never change at runtime so there is no trade-off.

**What was changed:**
- `match_spec.py`: `_BRACKETS_RE`, `_BRACKETS_KV_RE`, `_PARENS_RE`, `_NAME_VERSION_RE`, `_VERSION_BUILD_RE`
- `activate.py`: `_MSYS2_CYGWIN_PREFIX_RE`, `_PATH_SEPARATOR_RE`

The three `re.sub(re.escape(current_prompt_modifier), ...)` calls in `activate.py` use a runtime value and are left as-is.

**Measured impact:** eliminates `re.compile()` calls from the `MatchSpec` parse hot-path, which is exercised on every package spec the solver reads. Part of #15867.

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -60,6 +60,9 @@ if TYPE_CHECKING:
 
 log = getLogger(__name__)
 
+_MSYS2_CYGWIN_PREFIX_RE = re.compile(r"^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*")
+_PATH_SEPARATOR_RE = re.compile(r"\\|/")
+
 
 BUILTIN_COMMANDS = {
     "activate": ActivateHelp(),
@@ -614,7 +617,7 @@ class _Activator(metaclass=abc.ABCMeta):
 
                 # MSYS2 /c/
                 # cygwin /cygdrive/c/
-                if re.match("^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*", prefix):
+                if _MSYS2_CYGWIN_PREFIX_RE.match(prefix):
                     path = unix_path_to_win(path, prefix)
 
                 if isdir(path):
@@ -846,7 +849,7 @@ class _Activator(metaclass=abc.ABCMeta):
     def _resolve_prefix(self, env_name_or_prefix: str) -> str:
         """Hook for shell-specific activation path validation."""
         # get environment prefix
-        if re.search(r"\\|/", env_name_or_prefix):
+        if _PATH_SEPARATOR_RE.search(env_name_or_prefix):
             prefix = expand(env_name_or_prefix)
             if not isdir(join(prefix, "conda-meta")):
                 raise EnvironmentLocationNotFound(prefix)
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -60,8 +60,9 @@ if TYPE_CHECKING:
 
 log = getLogger(__name__)
 
-_MSYS2_CYGWIN_PREFIX_RE = re.compile(r"^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*")
-_PATH_SEPARATOR_RE = re.compile(r"\\|/")
+_MSYS2_CYGWIN_PREFIX_RE: re.Pattern[str] = re.compile(
+    r"^(/[A-Za-z]/|/cygdrive/[A-Za-z]/).*"
+)
 
 
 BUILTIN_COMMANDS = {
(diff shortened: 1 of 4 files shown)


conda/models/match_spec.py
```diff
@@ -37,6 +37,14 @@
 
 log = getLogger(__name__)
 
+_BRACKETS_RE = re.compile(r".*(?:(\[.*\]))")
```
dholth: fyi these are `<class 're.Pattern'>`

dholth: Could these be changed to re.VERBOSE? "This flag allows you to write regular expressions that look nicer and are more readable by allowing you to visually separate logical sections of the pattern and add comments."

conda/activate.py
```diff
@@ -846,7 +849,7 @@ def _get_environment_env_vars(self, prefix):
     def _resolve_prefix(self, env_name_or_prefix: str) -> str:
         """Hook for shell-specific activation path validation."""
         # get environment prefix
-        if re.search(r"\\|/", env_name_or_prefix):
+        if _PATH_SEPARATOR_RE.search(env_name_or_prefix):
```
dholth: Could be "if slash in env_name_or_prefix or backslash in env_name_or_prefix" avoiding regex entirely

dholth: Are these in a hot path or does this change make it a tiny bit slower to import activate?

dholth (approved):
LGTM; I didn't carefully check that the constants were exactly the same as the inline regex's.

conda/models/match_spec.py
jezdez: Added `re.Pattern[str]` annotations to all the constants in ec89f3b.

jezdez: Done, applied `re.VERBOSE` with inline comments to the three more complex patterns (`_BRACKETS_KV_RE`, `_NAME_VERSION_RE`, `_VERSION_BUILD_RE`). Left the simple ones (`_BRACKETS_RE`, `_PARENS_RE`) as-is since they're short enough to read inline.

conda/activate.py
jezdez: Good call, replaced with plain string checks and removed `_PATH_SEPARATOR_RE` entirely.

conda/models/match_spec.py
```diff
@@ -37,6 +37,38 @@
...
 log = getLogger(__name__)
 
+_BRACKETS_RE: re.Pattern[str] = re.compile(r".*(?:(\[.*\]))")
+_BRACKETS_KV_RE: re.Pattern[str] = re.compile(
+    r"""
+    ([a-zA-Z0-9_-]+?)       # key
+    =                        # separator
+    (["\']?)                 # optional opening quote
+    ([^\'"]*?)               # value
+    (\2)                     # matching closing quote
```
dholth, line 47: I checked whether MatchSpec would parse mismatched opening and closing quotes, which regex (or this regex?) isn't able to detect. I suppose there are nested quotes in theory in the new "where" clause?
It will parse such a matchspec `package[subdir='foo"]` but it will ignore the subdir key.

dholth: Python 3.11 adds `+` which should be good for performance. Not necessarily useful here?
```
{m,n}+

    Causes the resulting RE to match from m to n repetitions of the preceding RE, attempting to match as many repetitions as possible without establishing any backtracking points. 
```

jezdez: Yeah, that's pre-existing behavior from the backreference `(\2)` requiring matching quotes. Mismatched quotes silently fail to match and the key gets dropped. Worth fixing separately but out of scope for this PR.

jezdez: Interesting, but we still support Python 3.10 so possessive quantifiers aren't an option yet. These patterns also match short strings so backtracking isn't really a concern here.

PR #15891, opened 2026-04-09
Enable ruff G004 rule; use lazy log formatting across codebase

jezdez:
### Description

Replace `log.debug(f"...")` / `log.warning(f"...")` calls with `%-style` lazy formatting (`log.debug("...", arg)`) across the codebase. With f-strings the string is always interpolated before calling the log method, even when the log level is disabled. With `%`-style arguments the interpolation is deferred until the logging framework decides to emit the record.

Enable the ruff [`G004`](https://docs.astral.sh/ruff/rules/logging-f-string/) rule to enforce this going forward.

Files changed: `cli/main_run.py`, `common/configuration.py`, `core/envs_manager.py`, `core/link.py`, `core/package_cache_data.py`, `env/installers/pip.py`, `exception_handler.py`, `gateways/connection/download.py`, `gateways/disk/create.py`, `gateways/disk/delete.py`, `gateways/disk/update.py`, `gateways/repodata/__init__.py`, `gateways/subprocess.py`, `models/version.py`, `notices/core.py`, `notices/fetch.py`, `plugins/manager.py`, `plugins/subcommands/doctor/__init__.py`, `plugins/subcommands/doctor/health_checks/altered_files.py`, `plugins/subcommands/doctor/health_checks/environment_txt.py`, `testing/fixtures.py`.

The `deepcopy` → dict comprehension change in `resolve.py` (the other half of A20) has been split into #15917.

Part of #15867.

### Checklist - did you ...

- [x] Add a file to the `news` directory (using the template) for the next release's release notes?
- ~~Add / update outdated documentation?~~ (not applicable)

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -347,7 +347,7 @@ def create_link(src, dst, link_type=LinkType.hardlink, force=False):
         if lexists(dst) and not isdir(dst):
             if not force:
                 maybe_raise(BasicClobberError(src, dst, context), context)
-            log.info(f"file exists, but clobbering for directory: {dst!r}")
+            log.info("file exists, but clobbering for directory: %r", dst)
             rm_rf(dst)
         mkdir_p(dst)
         return
@@ -361,7 +361,7 @@ def create_link(src, dst, link_type=LinkType.hardlink, force=False):
     if lexists(dst):
         if not force:
             maybe_raise(BasicClobberError(src, dst, context), context)
-        log.info(f"file exists, but clobbering: {dst!r}")
+        log.info("file exists, but clobbering: %r", dst)
         rm_rf(dst)
 
     if link_type == LinkType.hardlink:
--- a/conda/cli/main_run.py
+++ b/conda/cli/main_run.py
@@ -146,7 +146,8 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     if response.rc != 0:
         log = getLogger(__name__)
         log.error(
-            f"`conda run {' '.join(args.executable_call)}` failed. (See above for error)"
+            "`conda run %s` failed. (See above for error)",
+            " ".join(args.executable_call),
         )
 
     # remove script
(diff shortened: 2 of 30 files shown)


dholth:
I've used `json.loads(json.dumps(dict))` which is also (much) faster than `deepcopy()` but has questionable memory usage.

conda/gateways/disk/create.py
```diff
@@ -337,7 +337,7 @@ def create_link(src, dst, link_type=LinkType.hardlink, force=False):
         if lexists(dst) and not isdir(dst):
             if not force:
                 maybe_raise(BasicClobberError(src, dst, context), context)
-            log.info(f"file exists, but clobbering for directory: {dst!r}")
+            log.info("file exists, but clobbering for directory: %r", dst)
```
dholth, line 350: the wording is a little awkward

jezdez: The wording is pre-existing, not changing it in this PR to keep the diff focused on the formatting change.

dholth:
Your f-string removal reminds me that some Python projects do an `if <log level...>: log()` to avoid that call entirely on a hot path.

jezdez:
> Your f-string removal reminds me that some Python projects do an `if <log level...>: log()` to avoid that call entirely on a hot path.

Yeah, `logging` isn't known to be fast; we could even consider switching if we want to be bold and improve speed, like with [structlog](https://github.com/hynek/structlog).

dholth:
I like structlog but I remember spending a long time trying and failing to integrate it with std logging.

jezdez:
pre-commit.ci autofix

kenodegard (commented):
Let's enable https://docs.astral.sh/ruff/rules/logging-f-string/?

dholth:
Was `ruff --fix` able to enact these changes?

dholth (approved):
LGTM. I see that some of these will not be called more than once in a conda session; but may as well consistently apply string formatting rules for all `log`

PR #15893, opened 2026-04-09
Require explicit `conda.plugins.types` import (A22)

jezdez:
### Description

Stop `conda.plugins` from eagerly importing `conda.plugins.types`, and stop exposing `types` as an implicit attribute after only `import conda.plugins`. Plugin authors should use an explicit `import conda.plugins.types`; the deprecated `conda.plugins.CondaSolver`-style re-exports remain available for their deprecation window and now load `conda.plugins.types` lazily through `deprecated.constant(factory=...)`.

Part of the startup-latency work tracked in #15867 (A22), using the `factory=` kwarg added in #15925.

What changed in `conda/plugins/__init__.py`:

- Dropped the eager `from . import types`.
- Removed `types` from `__all__`, so `conda.plugins.types` requires `import conda.plugins.types`.
- Replaced the 16 eager `deprecated.constant("26.3", "26.9", name, types.X, …)` calls with a loop that registers each name using `deprecated.constant(factory=lambda n=name: getattr(_load_types(), n), …)`.

Performance check with `hyperfine --shell=none` and the repo devenv Python:

- `origin/main`: `import conda.plugins` eagerly imports `types`: 153.2 ms ± 9.2 ms.
- This PR: `import conda.plugins` keeps `types` unloaded: 29.8 ms ± 1.8 ms.
- The explicit-import-only version is statistically tied with the previous lazy-attribute version (`1.00 ± 0.09`).

Tests (`tests/plugins/test_lazy_types.py`):

- `conda.plugins.types` is not in `sys.modules` after `import conda.plugins` (subprocess-isolated check).
- `conda.plugins.types` raises `AttributeError` until `import conda.plugins.types` is used.
- All 16 legacy re-exports emit a `(Pending)DeprecationWarning` and resolve to the type from `conda.plugins.types` (parametrized).
- Unknown attributes still raise `AttributeError`.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/__init__.py
+++ b/conda/plugins/__init__.py
@@ -27,7 +27,6 @@ as well as conda's internal implementations of plugins.
 
 from __future__ import annotations
 
-import sys
 from typing import TYPE_CHECKING
 
 from ..deprecations import deprecated
@@ -36,31 +35,15 @@ from .hookspec import hookimpl
 if TYPE_CHECKING:
     from types import ModuleType
 
-__all__ = ["hookimpl", "types"]
+__all__ = ["hookimpl"]
 
 
 def _load_types() -> ModuleType:
-    # ``__import__`` avoids recursing through this module's own
-    # ``__getattr__`` via ``_handle_fromlist``.
     return __import__("conda.plugins.types", fromlist=["types"])
 
 
-def __getattr__(
-    name: str,
-    # Default-arg capture keeps sys.is_finalizing reachable during
-    # interpreter shutdown, when module globals get cleared.
-    _finalizing=sys.is_finalizing,
-) -> object:
-    if _finalizing():
-        raise AttributeError(name)
-    if name == "types":
-        return _load_types()
-    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
-
-
-# ``deprecated.constant`` installs its own registry as ``__getattr__`` and
-# chains to the one defined above as its fallback, so ``conda.plugins.types``
-# resolves lazily while the 16 deprecated re-exports still warn on access.
+# ``deprecated.constant`` installs its own registry as module ``__getattr__``,
+# so the 16 deprecated re-exports still warn on access.
 for name in (
     "CondaAuthHandler",
     "CondaEnvironmentSpecifier",
--- a/conda/plugins/__init__.py
+++ b/conda/plugins/__init__.py
@@ -25,121 +25,65 @@ as well as conda's internal implementations of plugins.
 
 """
 
+from __future__ import annotations
+
+import sys
+from typing import TYPE_CHECKING
+
 from ..deprecations import deprecated
-from . import types
 from .hookspec import hookimpl
+
+if TYPE_CHECKING:
+    from types import ModuleType
 
 __all__ = ["hookimpl", "types"]
 
-deprecated.constant(
-    "26.3",
-    "26.9",
+
+def _load_types() -> ModuleType:
+    # ``__import__`` avoids recursing through this module's own
+    # ``__getattr__`` via ``_handle_fromlist``.
+    return __import__("conda.plugins.types", fromlist=["types"])
+
+
+def __getattr__(
+    name: str,
+    # Default-arg capture keeps sys.is_finalizing reachable during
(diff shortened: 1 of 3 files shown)


conda/plugins/__init__.py
```diff
@@ -24,121 +24,47 @@
...
+        "CondaRequestHeader",
+        "CondaSetting",
+        "CondaSolver",
+        "CondaSubcommand",
+        "CondaVirtualPackage",
+    }
 )
+
+
+def __getattr__(name: str) -> object:
```
dholth: module-level getattr affecting everything across any module with deprecated globals is a bit terrifying

jezdez: true story :-/

jezdez: I took another approach via https://github.com/conda/conda/pull/15925 now

conda/plugins/__init__.py
```diff
@@ -24,121 +24,65 @@
 
 """
 
+from __future__ import annotations
+
+import sys
+from typing import TYPE_CHECKING
+
 from ..deprecations import deprecated
-from . import types
```
kenodegard, line 29: What about deprecating the reexport entirely? Require explicit import of `conda.plugins.types`?

jezdez: The 16 type names (`CondaSubcommand`, `CondaHealthCheck`, etc.) are already deprecated via `deprecated.constant()` with a 26.3 → 26.9 window -- this PR just switches them from eager `value=` to lazy `factory=` so we don't pay the import cost at registration time. That part is unchanged.

Are you asking whether we should *also* deprecate `conda.plugins.types` as an attribute of `conda.plugins` (i.e., warn on `conda.plugins.types` access after `import conda.plugins`, and require an explicit `import conda.plugins.types`)? Or are you suggesting we skip the deprecation window and drop the 16 re-exports now?

kenodegard: I'm suggesting that instead of eagerly importing `conda.plugins.types` and allowing implicit usage we expect explicit imports.

Today we can say:

```
import conda.plugins

conda.plugins.types  # ← success!
```

If/when we drop the re-export:

```
import conda.plugins

conda.plugins.types  # ← failure!

import conda.plugins.types

conda.plugins.types  # ← success!
```

jezdez: Done in 8cb79b412: `conda.plugins.types` is no longer exposed from `conda.plugins.__all__`, and `import conda.plugins; conda.plugins.types` now raises `AttributeError`. Explicit `import conda.plugins.types` still works, and the 16 deprecated `conda.plugins.Conda*` re-exports still lazy-load through `deprecated.constant(factory=...)` for their deprecation window.

I also updated the tests/news/PR description. Focused tests pass (`tests/plugins/test_lazy_types.py`), pre-commit passed on the touched files, and the `hyperfine --shell=none` check showed this explicit-import-only version tied with the previous lazy-attribute version (`1.00 ± 0.09`) while preserving the cold-import win over `origin/main`.

tests/plugins/test_lazy_types.py
```diff
@@ -0,0 +1,101 @@
...
+    import conda.plugins.types as types_mod
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        value = getattr(conda.plugins, name)
+
+    assert any(
+        issubclass(w.category, (DeprecationWarning, PendingDeprecationWarning))
+        for w in caught
+    ), f"expected a (Pending)DeprecationWarning for conda.plugins.{name}"
```
kenodegard:
```suggestion
    with pytest.deprecated_call():
        value = getattr(conda.plugins, name)

```

kenodegard (approved):

tests/plugins/test_lazy_types.py
```diff
@@ -0,0 +1,96 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+"""Verify ``conda.plugins.types`` loads lazily and its deprecated re-exports."""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+import textwrap
+import warnings
```
kenodegard:
```suggestion
```

PR #15909, opened 2026-04-14
Add reporting for missing packages to configured channels in Exceptio…

tedpatrick:
This pull request adds a new feature to automatically report missing packages to conda channels when a `PackagesNotFoundInChannelsError` is raised. The reporting mechanism sends a GET request to each relevant channel's `/missing` endpoint with the missing package specs, helping channel maintainers track demand. The implementation is robust, avoids disrupting the normal error path, and is thoroughly tested.

**New missing package reporting feature:**

* Added a `_report_missing_packages` method to `ExceptionHandler` in `conda/exception_handler.py` that sends a fire-and-forget GET request to each deduplicated base channel URL when a `PackagesNotFoundInChannelsError` occurs. This includes handling various spec formats, skipping in offline mode, and logging failures at DEBUG level.
* Modified `handle_exception` to call `_report_missing_packages` when handling `PackagesNotFoundInChannelsError`.

**Testing and validation:**

* Introduced comprehensive tests in `tests/test_exceptions.py` to verify correct GET request formation, handling of raw string specs, offline mode behavior, HTTP error swallowing, channel deduplication, user messaging, and integration with `handle_exception`.
* Added necessary imports and test setup for the new feature and exception class. [[1]](diffhunk://#diff-f46006e3f43ffb1dd5d6862005427f6620f4dcfb1fa2f883d8482550069eeeccR6-R12) [[2]](diffhunk://#diff-f46006e3f43ffb1dd5d6862005427f6620f4dcfb1fa2f883d8482550069eeeccR31-R39)

PR #15911, opened 2026-04-14
Add `conda_exception_observers` plugin hook

### LINKED ISSUE #15910: Add `conda_exception_handlers` plugin hook
### What is the idea?

Add a `conda_exception_handlers` plugin hook that allows plugins to register observational callbacks invoked when exceptions are handled. Modelled after CPython's `sys.excepthook`, `threading.excepthook`, and `sys.unraisablehook`.

Handlers are strictly observational - they cannot suppress, modify, or redirect exceptions. Return values are ignored. Any exception raised by a handler is caught at the `BaseException` level and logged. Handlers receive a frozen `CondaExceptionInfo` dataclass with the exception triple and optional conda runtime state.

The design is informed by CPython's exception hook system:

- Hooks are purely observational - they affect reporting only, not control flow.
- Structured args for forward compatibility - handlers receive a frozen dataclass (like `threading.ExceptHookArgs`), so new fields can be added without breaking existing plugins.
- `BaseException`-level catch - unlike `threading.excepthook` which only catches `Exception`, we catch `BaseException` so a buggy plugin can never kill conda's error reporting path.
- Reference lifecycle - traceback references are cleared after handlers run, and documentation warns plugins not to store `exc_value`/`exc_traceback` references beyond the callback lifetime (following `sys.unraisablehook` precedent).

The API:

```python
@dataclass(frozen=True)
class CondaExceptionInfo:
    exc_type: type[BaseException]
    exc_value: BaseException
    exc_traceback: TracebackType
    argv: tuple[str, ...] | None = None
    conda_version: str | None = None
    return_code: int | None = None
    active_prefix: str | None = None
    target_prefix: str | None = None
    channels: tuple[str, ...] | None = None
    subdir: str | None = None
    offline: bool | None = None
    dry_run: bool | None = None
    quiet: bool | None = None
    json: bool | None = None

@dataclass
class CondaExceptionHandler(CondaPlugin):
    name: str
    hook: Callable[[CondaExceptionInfo], None]
    run_for: set[str]
```

- `hook`: callable invoked with a `CondaExceptionInfo` instance. Named after CPython's `excepthook` convention.
- `run_for`: set of exception class names matched against the exception's full MRO. Scopes include `{"BaseException"}` (everything), `{"Exception"}` (standard exceptions), `{"CondaError"}` (all conda errors), `{"PackagesNotFoundError"}` (specific error and subclasses), `{"MemoryError"}`, `{"KeyboardInterrupt"}`, `{"SystemExit"}`, or combinations like `{"CondaError", "MemoryError"}`.

The exception triple is always populated. Conda runtime fields (`argv`, `conda_version`, `return_code`, etc.) are populated all-or-nothing: either the runtime was available and all fields are set, or it wasn't and they're all `None`. Check `conda_version is not None` to tell the two cases apart. `active_prefix` can be `None` even when the runtime is available (no active environment).

`invoke_exception_handlers` on `CondaPluginManager` does the dispatch: builds the info object with the exception triple, tries to populate runtime fields from context (falls back to bare triple on failure), then calls matching handlers. The entire method body is wrapped in `try/except BaseException` so failures can never break conda's error reporting. Follows the same pattern as `invoke_pre_commands` / `invoke_post_commands`. Called from `ExceptionHandler.handle_exception` at the top, before the `isinstance` chain, so it fires for all exception types.

### Why is this needed?

[PR #15909](https://github.com/conda/conda/pull/15909) demonstrates demand for hooking into the exception path, but hardcoding vendor-specific logic in `ExceptionHandler` is not the right approach. Channel providers, enterprise users, and plugin authors need an extensible mechanism for error reporting and telemetry without touching conda core.

- Plugin authors can register callbacks for specific error types (e.g., report missing packages, send telemetry on solver failures).
- Channel maintainers can build plugins that track demand for missing packages.
- Enterprise users can integrate conda error reporting with internal monitoring systems.
- conda's error path is never disrupted by plugin bugs - all handlers are fire-and-forget.

### What should happen?

Plugin authors register a handler via `conda_exception_handlers`:

```python
@plugins.hookimpl
def conda_exception_handlers():
    yield plugins.types.CondaExceptionHandler(
        name="missing-package-reporter",
        hook=report_missing,
        run_for={"PackagesNotFoundInChannelsError"},
    )
```

When an exception is handled, `ExceptionHandler.handle_exception` invokes matching handlers before the error is printed. Handlers are fire-and-forget - conda's error path is never disrupted by plugin bugs. For non-`CondaError` exceptions (`MemoryError`, `KeyboardInterrupt`, `SystemExit`), the conda runtime fields may be `None`.

### Additional Context

Implementation: #15911 (types, hookspec, manager, exception handler integration, tests, tutorial docs). Supersedes #15909. Target: May release cycle.

Key design decisions:
- `hook` not `action` - named after CPython's `excepthook` to emphasize observational nature, distinct from `action` used by pre/post command hooks
- `run_for` uses class name strings - avoids import coupling; MRO walk supports subclass matching
- Frozen info object - prevents plugins from mutating shared state; forward-compatible
- Context fields on info object - `offline`, `dry_run`, `target_prefix`, `channels`, `subdir`, `quiet`, `json` surfaced directly so plugins can make decisions without importing context
- Channels use canonical names - `channels` contains canonical names (e.g. `defaults`, `conda-forge`) via `Channel.canonical_name`, not raw URLs, so credentials and tokens are never exposed to plugins
- Traceback cleanup - `exc_info` and `exc_tb` references are deleted after handlers run to limit exposure of frame locals and break reference cycles promptly
- All exception types - fires for `CondaError`, `MemoryError`, `KeyboardInterrupt`, `SystemExit`, etc. Runtime fields are `None` when context isn't available
- All-or-nothing runtime fields - follows CPython's flat args pattern (`threading.ExceptHookArgs`, `sys.UnraisableHookArgs`); no sentinel needed since plugins can check `conda_version is not None`
- `exception_handlers` not `error_handlers` - conda uses exceptions for flow control (`DryRunExit`, `CondaExitZero`, `CondaSystemExit` are all `CondaError` subclasses with `return_code = 0`). A plugin registered for `{"CondaError"}` sees these too, so "error" would be misleading. "Exception" is the accurate umbrella term, and it matches the CPython naming (`sys.excepthook`)

jezdez:
### Description

Adds a `conda_exception_handlers` plugin hook that lets plugins register observational callbacks for any exception type. Modelled after CPython's `sys.excepthook` - handlers are fire-and-forget and cannot suppress, modify, or redirect exceptions.

The motivation is #15909, which hardcodes missing-package reporting into `ExceptionHandler`. This hook provides a generic, extensible alternative for error telemetry, demand tracking, or internal monitoring without touching conda core.

Closes #15910

**What plugins receive**

`CondaExceptionInfo` is a frozen dataclass with:
- The exception triple (`exc_type`, `exc_value`, `exc_traceback`) - always populated
- Conda runtime fields (`argv`, `conda_version`, `return_code`, `active_prefix`, `target_prefix`, `channels`, `subdir`, `offline`, `dry_run`, `quiet`, `json`) - populated all-or-nothing from context, `None` when unavailable (e.g. `MemoryError` during early startup)

Plugins can check `conda_version is not None` to know if the runtime was available. `active_prefix` can also be `None` when no environment is active.

**Scope control via `run_for`**

`CondaExceptionHandler.run_for` is a set of class name strings matched against the exception's full MRO:
- `{"BaseException"}` catches everything
- `{"CondaError"}` catches all conda errors
- `{"PackagesNotFoundError"}` catches that error and its subclasses
- `{"MemoryError"}`, `{"KeyboardInterrupt"}`, `{"SystemExit"}` for specific types
- Combine as needed: `{"CondaError", "MemoryError"}`

**Security**

- `channels` contains canonical names (e.g. `defaults`, `conda-forge`) via `Channel.canonical_name`, not raw URLs - credentials and tokens are never exposed
- Traceback references are cleared after handlers run to limit exposure of frame locals
- Any exception raised by a handler (including `SystemExit`) is caught at `BaseException` level and logged

**Implementation**

Dispatch lives in `invoke_exception_handlers` on `CondaPluginManager`, following the same pattern as `invoke_pre_commands`/`invoke_post_commands`. Called from `ExceptionHandler.handle_exception` at the top, before the `isinstance` chain, so it fires for all exception types.

**Docs and tests**

Tutorial with a missing-package-reporter example using `get_session()` (with `offline`/`dry_run` guards). 17 tests covering filtering, MRO matching, fault tolerance, frozen info, `ExceptionHandler` integration, and dispatch for `MemoryError`/`KeyboardInterrupt`/`SystemExit`.

### Checklist - did you ...


conda/exception_handler.py
```diff
@@ -71,6 +71,13 @@ def handle_exception(self, exc_val: BaseException, exc_tb: TracebackType) -> int
             NoSpaceLeftError,
         )
 
+        try:
+            from .base.context import context
+
+            context.plugin_manager.invoke_exception_handlers(exc_val, exc_tb)
+        except BaseException:
+            log.debug("Failed to invoke exception handler plugins", exc_info=True)
```
travishathaway, line 79: @jezdez,

What are you thoughts on adding this to `handle_unexpected_exception` exceptions too?

I'm working on this project right now as a proof-of-concept:

- https://github.com/travishathaway/conda-sentry

I'm imagining a scenario where a system administrator may want to configure conda to use sentry in order to better track and debug runtime errors , and getting insight into unhandled exceptions would be pretty valuable.

jezdez: Good question, but I think the current placement already covers this. `invoke_exception_handlers` is called at the top of `ExceptionHandler.handle_exception`, *before* the `isinstance` chain, and `handle_unexpected_exception` has exactly one caller — the fall-through at the end of `handle_exception`. So every exception routed to `handle_unexpected_exception` has already fired the hook a few lines earlier; plugins don't need a second entry point.

To make that guarantee explicit rather than implicit, I just pushed a commit that adds:

- `test_invoked_for_unexpected_exception_path` — drives a plain `RuntimeError` through `ExceptionHandler.handle_exception` and asserts the handler is invoked, covering the `handle_unexpected_exception` path end-to-end.
- A "Choosing a `run_for` scope" table in the tutorial that maps use cases (error tracking, conda-only telemetry, diagnostics) to the right scope, plus an explicit callout that dispatch happens *before* conda's isinstance chain so "unexpected" exceptions are observed too.

For `conda-sentry` specifically: if you want to capture the exceptions conda treats as "unexpected" (the `RuntimeError`/`KeyError`/etc. that aren't `CondaError` subclasses), flip your handler to `run_for={"Exception"}`. That covers every standard exception while still skipping `KeyboardInterrupt` / `SystemExit`, which Sentry trackers usually don't want anyway. `{"CondaError"}` is a perfectly reasonable default if you only care about conda-originated failures — it's just not the one that corresponds to "unhandled".

Also rebased onto latest `main` while I was in here.

conda/plugins/manager.py
```diff
@@ -472,6 +482,82 @@ def invoke_post_commands(self, command: str) -> None:
...
+                try:
+                    handler.hook(exc_info)
+                except BaseException:
+                    log.debug(
+                        "Exception handler plugin %r failed",
+                        handler.name,
+                        exc_info=True,
+                    )
+
+            del exc_info, exc_tb
```
kenodegard: ?

conda/plugins/types.py
```diff
@@ -801,3 +802,108 @@ class CondaPackageExtractor(CondaPlugin):
...
+                      ``{"SystemExit"}`` — specific non-conda exceptions.
+                    - ``{"CondaError", "MemoryError"}`` — combine scopes.
+
+                    For non-``CondaError`` exceptions the conda-specific fields
+                    on :class:`CondaExceptionInfo` may be ``None``.
+    """
+
+    name: str
+    hook: Callable[[CondaExceptionInfo], None]
+    run_for: set[str]
```
kenodegard: Why `str`? Wouldn't this be better as:


```suggestion
    run_for: set[BaseException]
```

conda/exception_handler.py
```diff
@@ -70,6 +70,13 @@ def handle_exception(self, exc_val: BaseException, exc_tb: TracebackType) -> int
...
```
kenodegard: What about allowing `conda_exception_handler` to exit early? If the hook returns a truthy value `handle_exception` exits? This would allow plugins to define passthrough exception handlers (e.g., sentry reporting) versus UX exception handlers (e.g., changing how exceptions are rendered). Similar approach to [Python context manager's `__exit__`](https://docs.python.org/3/reference/datamodel.html#object.__exit__).

conda/plugins/types.py
```diff
@@ -801,3 +802,108 @@ class CondaPackageExtractor(CondaPlugin):
...
+    conda_version: str | None = None
+    return_code: int | None = None
+    active_prefix: str | None = None
+    target_prefix: str | None = None
+    channels: tuple[str, ...] | None = None
+    subdir: str | None = None
+    offline: bool | None = None
+    dry_run: bool | None = None
+    quiet: bool | None = None
+    json: bool | None = None
```
kenodegard, line 864: Why only these? Wouldn't it be better to pass context to the hook?

```python
hook: Callable[[CondaExceptionInfo, Context | None], None]
```

conda/plugins/manager.py
jezdez: Breaks reference cycles between the traceback and frame locals so the GC doesn't keep the entire call stack alive until the next sweep. Added a comment on the line explaining this.

conda/plugins/types.py
jezdez: Strings avoid import coupling. Plugins don't need to import conda's exception classes to register a handler, and a handler can target an exception from a plugin that might not even be installed (the string just won't match, no `ImportError`). The MRO walk handles subclass matching, so `{"CondaError"}` still catches all subclasses.

conda/exception_handler.py
jezdez: Keeping it observational-only, same as `sys.excepthook`. If a plugin could suppress or redirect an exception, a buggy handler could silently swallow errors and that'd be really hard to debug. If there's demand for custom rendering later we can add a separate hook with explicit opt-in.

conda/plugins/types.py
jezdez: `Context` is mutable, so passing it to plugins in the error path is risky. A plugin could change `context.channels` or `context.target_prefix` and affect conda's own error reporting or cleanup. The frozen dataclass gives a read-only snapshot. Trade-off is maintaining the field list, but that's better than exposing a mutable singleton to third-party code.

conda/plugins/manager.py
jezdez: Done in 97c04eb.

tests/plugins/test_exception_handlers.py
```diff
@@ -0,0 +1,422 @@
...
+    """active_prefix is populated from context."""
+    exc = CondaError("boom")
+    try:
+        raise exc
+    except CondaError:
+        _, exc_val, exc_tb = sys.exc_info()
+        plugin_manager.invoke_exception_handlers(exc_val, exc_tb)
+
+    info = catch_all_plugin.calls[0]
+    assert isinstance(info.active_prefix, (str, type(None)))
```
kenodegard: Can ensure its the same as `context.active_prefix`?

```suggestion
    assert info.active_prefix == context.active_prefix
```

tests/plugins/test_exception_handlers.py
```diff
@@ -0,0 +1,422 @@
...
+    """SystemExit.code is used as the return_code."""
+    exc = SystemExit(42)
+    try:
+        raise exc
+    except SystemExit:
+        _, exc_val, exc_tb = sys.exc_info()
+        plugin_manager.invoke_exception_handlers(exc_val, exc_tb)
+
+    info = base_exception_plugin.calls[0]
+    assert info.return_code == 42
```
kenodegard: These can be parameterized:


```suggestion
@pytest.mark.parametrize(
    "exc,return_code",
    [
        (ValueError("oops"), 1),
        (MemoryError(), 1),
        (KeyboardInterrupt(), 1),
        (SystemExit(2), 2),
        (SystemExit(42), 42),
    ],
)
def test_base_exception_plugin_dispatches_exception_types(
    base_exception_plugin,
    plugin_manager,
    exc,
    return_code,
):
    """A handler with run_for={'BaseException'} fires for any exception type."""
    try:
        raise exc
    except BaseException:
        _, exc_val, exc_tb = sys.exc_info()
        plugin_manager.invoke_exception_handlers(exc_val, exc_tb)

    assert len(base_exception_plugin.calls) == 1
    info = base_exception_plugin.calls[0]
    assert info.exc_value is exc
    assert info.exc_type is type(exc)
    assert info.return_code == return_code
```

tests/plugins/test_exception_handlers.py
```diff
@@ -0,0 +1,422 @@
...
+        _, exc_val, exc_tb = sys.exc_info()
+        plugin_manager.invoke_exception_handlers(exc_val, exc_tb)
+
+    assert len(catch_all_plugin.calls) == 1
+    info = catch_all_plugin.calls[0]
+    assert info.exc_value is exc
+    assert info.exc_type is CondaError
+    assert info.exc_traceback is not None
+    assert isinstance(info.argv, tuple)
+    assert isinstance(info.conda_version, str)
```
kenodegard: We could test this against the actual version string?

```suggestion
    assert info.conda_version == CONDA_VERSION
```

docs/source/dev-guide/plugins/exception_handlers.rst
```diff
@@ -0,0 +1,311 @@
...
+   requires-python = ">=3.10"
+   dependencies = ["conda"]
+
+   [project.entry-points."conda"]
+   conda-missing-reporter = "conda_missing_reporter.plugin"
+
+Install the package (``pip install -e .``) and the hook is active
+the next time conda runs.
+
+.. _entry points: https://packaging.python.org/en/latest/specifications/entry-points/
```
kenodegard: The QuickStart in `docs/source/dev-guide/plugins/index.rst` already explains how to register plugins, so we can just link to that?

kenodegard (approved):

tests/plugins/test_exception_handlers.py
jezdez: Done in d92c22b.

tests/plugins/test_exception_handlers.py
jezdez: Done in d92c22b.

tests/plugins/test_exception_handlers.py
jezdez: Done in d92c22b.

docs/source/dev-guide/plugins/exception_handlers.rst
jezdez: Done in d92c22b. Replaced with a :doc: link to the plugins Quick start.

jezdez:
Renamed the API in 24e9e6b:

- `conda_exception_handlers` -> `conda_exception_observers`
- `CondaExceptionHandler` -> `CondaExceptionObserver`
- `CondaExceptionInfo` -> `CondaExceptionEvent`
- `invoke_exception_handlers` -> `invoke_exception_observers`
- `run_for` -> `watch_for` (to distinguish from pre/post command `run_for` which has different semantics)

The "observer" naming makes the non-mutating, fire-and-forget nature self-evident from the API surface rather than relying on documentation caveats. `watch_for` pairs naturally with "observer" while `run_for` stays on pre/post commands where "run" fits.

Sorry for the churn post-approval @kenodegard, but figured this was worth getting right before it ships.

news/15911-exception-observer-plugin-hook
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add `conda_exception_observers` plugin hook for observational callbacks on any exception conda encounters (filterable per observer via `run_for`). (#15910)
```
kenodegard:
```suggestion
* Add `conda_exception_observers` plugin hook for observational callbacks on any exception conda encounters (filterable per observer via `watch_for`). (#15910)
```

kenodegard (approved):

PR #15915, opened 2026-04-14
Extend `--no-plugins` for per-plugin runtime disabling

### LINKED ISSUE #15914: Extend `--no-plugins` for per-plugin runtime disabling
### Checklist

### What is the idea?

Extend the existing `--no-plugins` flag so `--no-plugins=<name>[,...]` disables only the listed plugins for that invocation. Bare `--no-plugins` continues to disable all external plugins.

This is the per-plugin debugging piece of #16354.

### Why is this needed?

Users debugging plugin-related issues have no way to isolate a specific plugin without disabling all external plugins via `--no-plugins`. Selecting individual plugins lets users narrow the problem down to a reporter backend, command hook, or other plugin without disabling everything.

The precedent is pytest's `-p no:name`, which solves the same problem with the same underlying framework (pluggy).

### What should happen?

```
conda install some-package --no-plugins=conda-myreporter
conda doctor --no-plugins=my-custom-healthcheck
```

`CondaPluginManager` records distribution and entry-point aliases while loading plugins. After CLI parsing, `disable_plugins()` resolves the selected aliases to registered plugin names and calls `set_blocked()` before command hooks run.

### Additional Context

A persistent `disabled_plugins` config key in `.condarc` was considered but has a significant edge case: if a user disables their configured solver (e.g. `solver: libmamba`), conda cannot run at all, including to re-enable it. The per-invocation flag avoids this. Persistent disabling can be a follow-up once the edge cases are better understood.

jezdez:
### Description

Extend the existing `--no-plugins` flag for selective runtime disabling. Bare `--no-plugins` continues to disable all external plugins, while `--no-plugins=<name>[,...]` disables only the listed plugins for that invocation.

Plugins can be referenced by distribution name (for example, `conda-self`), entry point name, or canonical registered name. Built-in conda plugins are protected from being disabled. Disabling happens after entry points are loaded and blocks all hooks from each selected plugin module.

`conda run` forwards command arguments unchanged, including a bare `--no-plugins` passed to the executable without an explicit `--` separator.

Closes #15914.
Part of #16354.

### Checklist - did you ...

- [x] ~If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?~


conda/plugins/manager.py
```diff
@@ -469,12 +480,35 @@ def invoke_post_commands(self, command: str) -> None:
                 hook.action(command)
 
     def disable_external_plugins(self) -> None:
+        """Disable all external (non-built-in) plugins."""
+        self.disable_plugins(
+            name
+            for name, _ in self.list_name_plugin()
+            if not name.startswith(BUILTIN_PLUGIN_PREFIX)
```
dholth, line 487: I was going to ask about whether this feature existed, since disabling all (technically  plugins) is different than disabling all (all!) plugins.

Is it a little odd that we have to enumerate all plugins and then disable them, rather than storing filters and filtering during the lookup phase? Or will we have always enumerated the plugins by reading package metadata before we have imported anything from those modules?

conda/base/context.py
```diff
@@ -540,6 +540,9 @@ class Context(Configuration):
     ####################################################
 
     no_plugins = ParameterLoader(PrimitiveParameter(NO_PLUGINS))
+    no_plugin = ParameterLoader(
```
travishathaway, line 543: What about calling this `disabled_plugins` instead? I think this would be grammatically a little better even it if it is more to type.

On the command line, we could write it as `--disable-plugin=plugin-a --disable-plugin=plugin-b`.

I'm not going to block on this. I just thought I would suggest what I believe to be a better sounding alternative.

jezdez: Yeah, I was looking to use the `--no-<something>` pattern that we use elsewhere. I did consider `--disable-plugin` and it came in a close second only because of the precedence with the other form.

kenodegard (commented):
I'd much prefer we reuse `--no-plugins` and add aliases

`--no-plugins` → disable all plugins
`--no-plugins=...` → disable those listed

jezdez:
> I'd much prefer we reuse `--no-plugins` and add aliases
> 
> `--no-plugins` → disable all plugins `--no-plugins=...` → disable those listed

The plugins as a comma-separated list?

kenodegard:
> > I'd much prefer we reuse `--no-plugins` and add aliases
> > `--no-plugins` → disable all plugins `--no-plugins=...` → disable those listed
> 
> The plugins as a comma-separated list?

Ideally we'd support both 🤷🏼‍♂️ but reusing old patterns is also good for consistency

conda/plugins/manager.py
jezdez: Yes. Today `get_plugin_manager()` enumerates distribution and entry-point metadata, imports and registers each entry point, and only then `main_subshell()` applies `--no-plugins`. This PR keeps that existing behavior and blocks selected registered hooks afterward.

Avoiding the module import would require moving the filter into `load_entrypoints()`. Distribution and entry-point aliases are available before import, but canonical plugin names are not. I think that should be a separate loader change rather than part of this runtime-disabling change.

conda/base/context.py
jezdez: This has been superseded by the later review direction: the implementation now reuses the existing `--no-plugins` flag. Bare `--no-plugins` disables all external plugins, while `--no-plugins=<name>[,...]` disables only the selected plugins.

jezdez:
Updated this per the review: the implementation now reuses `--no-plugins`, with bare `--no-plugins` disabling all external plugins and `--no-plugins=<name>[,...]` disabling selected plugins. I also updated the PR and tracker wording to match.

conda/cli/conda_argparse.py
```diff
@@ -223,6 +234,17 @@ def __init__(self, *args, add_help=True, **kwargs):
...
+    def _parse_optional(self, arg_string):
+        if (
+            arg_string == "--no-plugins"
+            and (action := self._option_string_actions.get(arg_string)) is not None
+            and action.dest == "disabled_plugins"
+        ):
+            # Bare --no-plugins must not consume the following command. Interpret
+            # it as an empty value without modifying arguments passed to a child.
+            arg_string += "="
+        return super()._parse_optional(arg_string)
```
kenodegard, line 246: Does this force users to do `--no-plugins=PLUGIN`? Or will `--no-plugins PLUGIN` still work?

kenodegard: Looks like this breaks `--no-plugins PLUGIN`

jezdez: Ken, yes, the current implementation requires `=`. After trying the whitespace form locally, I think we should revisit overloading `--no-plugins` rather than add custom argument parsing.

With ordinary argparse `nargs="?"`, `conda --no-plugins info` consumes `info` as the plugin name instead of the command. Preserving both interpretations needs extra disambiguation, including handling external commands and keeping `conda run` arguments untouched. I would rather avoid that machinery in the plugin-recovery path.

I would propose:

- Keep `--no-plugins` as the existing boolean that disables all external plugins.
- Use repeatable `--disable-plugins NAME[,NAME...]` for selective disabling.
- Use repeatable `--enable-plugins NAME[,NAME...]` for explicit enable overrides, including under `--no-plugins` or `CONDA_NO_PLUGINS=true`.

The named options would accept both whitespace and `=` through normal argparse. This follows the existing separation between `--override-channels` and repeatable `--channel NAME`, and between `--no-shortcuts` and `--shortcuts-only NAME`.

That keeps the generic allowlist you suggested on #16520, preserves existing bare `--no-plugins` usage, and avoids guessing whether a token names a plugin or a command.

conda/cli/conda_argparse.py
```diff
@@ -290,12 +291,33 @@ def generate_pre_parser(**kwargs) -> ArgumentParser:
...
+    )
+    pre_parser.add_argument(
+        "--enable-plugins",
+        dest="enabled_plugins",
+        action="extend",
+        default=[],
+        metavar="PLUGIN[,PLUGIN...]",
+        type=comma_separated_stripped,
+        help="Keep the listed plugins enabled, overriding --no-plugins and CONDA_NO_PLUGINS.",
+    )
```
kenodegard, line 320: Should we include `choices`?

conda/cli/main.py
```diff
@@ -38,9 +48,19 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
     if context.no_plugins:
-        context.plugin_manager.disable_external_plugins()
+        context.plugin_manager.disable_external_plugins(
+            except_plugins=pre_args.enabled_plugins,
+        )
+    elif pre_args.disabled_plugins or pre_args.enabled_plugins:
+        context.plugin_manager.disable_plugins(
+            pre_args.disabled_plugins,
+            except_plugins=pre_args.enabled_plugins,
+        )
```
kenodegard, line 58:
```python
    context.plugin_manager.disable_external_plugins(
        None if context.no_plugins else pre_args.disabled_plugins,
        except_plugins=pre_args.enabled_plugins,
    )
```

kenodegard: Oh I see `disable_external_plugins` vs `disable_plugins`, exploring more...

PR #15887, opened 2026-04-09
Memoize context.root_writable to avoid repeated file-open (A14)

jezdez:
### Description

`context.root_writable` was a plain `@property` that opened a file handle against `$CONDA_ROOT_PREFIX/conda-meta/.conda_lock` on every access. It is called from `envs_dirs` and several other properties, so a single `conda` invocation can trigger it multiple times.

Switching to `@memoizedproperty` caches the result on the `Context` instance after the first call. The result does not change during a single process lifetime, so this is always safe.

**Measured impact:** eliminates repeated `open()` + `close()` syscalls on every access. Part of #15867.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~ (not applicable)

PR #15916, opened 2026-04-14
Replace auxlib.Entity package records with slotted dataclasses (A19b)

jezdez:
### Description

Replace the `auxlib.Entity` record hierarchy in `conda/models/records.py` with slotted dataclasses to reduce the cost of constructing and serializing package records. This covers `PackageRecord`, `PackageCacheRecord`, `SolvedRecord`, and `PrefixRecord`.

This is Track A, A19b, for #15867. Following @kenodegard's review, the Channel metaclass cleanup was split into #15942 (A19a) so the two changes could land independently. #15942 has since merged.

#### Changes

- Use `@dataclass(slots=True, init=False, eq=False, repr=False)` with a custom constructor and assignment handling for field coercion and validation.
- Combine the earlier `COERCIONS`, `DERIVED`, and `resolve_*` mechanisms and record field descriptors in a `FIELD_RESOLVERS` `ClassVar` registry. Each resolver receives `(value, instance)` and returns the coerced or derived value. This includes handling `subdir=None` correctly.
- Keep serialization conversions in a separate `DUMP_TRANSFORMS` `ClassVar` registry. A `Dumpable` protocol supports nested serialization.
- Preserve `_pkey`-based identity, hashing, and equality across record classes for solver compatibility. Intern `name`, `version`, `build`, and `subdir` strings with `sys.intern()` to share repeated values.
- Retain constructor aliases from `build_string` to `build`, `schannel` to `channel`, and `filename` to `fn`. Provide explicit `dump()`, `from_objects()`, `from_json()`, and `load()` methods, plus `__getitem__` and `get()` for dict-like attribute lookup. Keep the `set()` factory for `metadata`.
- Keep `Link`, `PathData`, `PathDataV1`, and `PathsData` on `auxlib.Entity`. These hold detailed installation metadata nested in `PrefixRecord` and are outside the record construction and serialization operations measured below.
- Convert dictionaries assigned to `PrefixRecord.link` into `Link` objects during construction and later assignment, so callers such as `conda clean` can read `link.source`.
- Preserve record field values in `from_objects()`, including empty `track_features`, `features`, and `flags`, and zero `timestamp` and `indexed_timestamp` values omitted by `dump()`. An explicit empty channel `track_features` value therefore takes precedence over stale cached features.
- Preserve `PackageCacheRecord.matches_metadata()` from #16347 so same-device package-cache selection continues comparing available `size` and `md5` values with requested package metadata, including legacy `.tar.bz2` metadata.
- Keep `PrefixRecord.extracted_package_dir` defaulting to `""`, matching the old `StringField(required=False)` behavior, so `link.py` can call `basename()` on an unpopulated value.
- Guard Entity-specific `__fields__` access in `conda/testing/helpers.py` for records that no longer expose that attribute.
- Use standard library features for the replacement implementation, including `functools.cache`, `pathlib.Path`, `typing.Protocol`, and f-strings, without adding dependencies.

#### Deprecations

The record-specific Entity field descriptors remain importable from `conda.models.records` during the deprecation period. They are pending deprecation, become deprecated in 27.3, and are scheduled for removal in 27.9. The original eight descriptors are `SubdirField`, `ChannelField`, `FilenameField`, `NoarchField`, `TimestampField`, `PackageTypeField`, `_FeaturesField`, and `Md5Field`. The current implementation also includes `IndexedTimestampField`, bringing the total to nine.

A lazy `deprecated.constant(factory=partial(_legacy_field, name))` shim loads these descriptors from `conda/models/_legacy_record_fields.py` on access. Record definitions no longer import that module or its dependencies eagerly. Timestamp parsing from `date` can still load `boltons.timeutils` when needed.

`PackageCacheRecord.md5` continues to compute a missing checksum on attribute access during the same deprecation period. The old `Md5Field` called `_calculate_md5sum()` when unset. Its replacement reads the inherited `md5` slot, calls `calculate_md5sum()` when needed, and uses `deprecated.topic("27.3", "27.9", ...)` to select the warning category for the running release. Callers can use `calculate_md5sum()` explicitly. Explicitly supplied or assigned checksums do not trigger automatic calculation or its warning.

The `md5` property is attached after the class definition because `@dataclass(slots=True)` removes class attributes whose names match inherited fields while preparing slots. `_make_md5_auto_compute_property()` keeps the getter and setter local to the factory while using the inherited slot for storage.

Deprecation coverage includes nine descriptor parameter cases, the implicit `md5` calculation case, and two explicit-value cases. The `pkg_tarball` and `cache_record_factory` fixtures share the tarball setup.

Package-cache selection coverage observes the public `calculate_md5sum()` method and verifies that missing requested `md5` metadata does not trigger checksum calculation.

#### Compatibility considerations

The record classes retain their named construction, serialization, identity, and lookup methods, but the public API is not identical to `auxlib.Entity`. In particular, record instances no longer have `__fields__` or an instance `__dict__`, and there is no explicit `__contains__` implementation.

Callers cannot attach arbitrary attributes or replace methods on individual instances of these slotted record classes. Downstream code, including libmamba-solver or anaconda-client, needs review if it uses patterns such as `record.some_custom_attr = ...`. A subclass without its own `__slots__` normally gains an instance `__dict__`, so the same restriction does not automatically apply to every subclass.

<details>
<summary>Historical performance and memory measurements</summary>

These are the measurements reported in the original PR description. They are retained as historical data and do not establish performance for the current revision.

**Per operation**

50,000 iterations with realistic 14-field repodata input, median of 5 rounds, Python 3.13 on macOS ARM64:

| Operation | Entity | Dataclass | Speedup |
| --- | --- | --- | --- |
| Instantiation | 26.3 µs | 8.1 µs | 3.2x |
| `dump()` serialization | 29.9 µs | 5.3 µs | 5.6x |

**At scale**

| Scenario | Entity | Dataclass | Saved |
| --- | --- | --- | --- |
| 2,000 records (`conda list`) init | 53 ms | 16 ms | -37 ms |
| 2,000 records (`conda list`) dump | 60 ms | 11 ms | -49 ms |
| 50,000 records (solver/repodata) init | 1,317 ms | 404 ms | -913 ms |
| 50,000 records (solver/repodata) dump | 1,495 ms | 264 ms | -1,231 ms |

**Memory, 50,000 records**

| Scenario | Entity | Dataclass | Saving |
| --- | --- | --- | --- |
| All unique names | 664 bytes/rec | 493 bytes/rec | 26% less |
| Realistic duplication (~5k packages, 3 channels) | 794 bytes/rec | 483 bytes/rec | 39% less |
| Realistic total | 38.8 MiB | 23.6 MiB | saves 15 MiB |

The original description also reported a CodSpeed `test_update[libmamba-update]` result of -28%, accompanied by a "Different runtime environments" warning. It noted the same -28% result on #15887 (A14), which did not change records. Those observations limit the comparison, but do not establish that the result was noise. The local record measurements above cover different operations and do not resolve that CI result.

</details>

### Checklist - did you ...

- [x] Add a file to the `releases/news` directory ([using the template](https://github.com/conda/conda/blob/main/releases/news/TEMPLATE)) for the next release's release notes? (`releases/news/15867-dataclass-records`)
- [x] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))? (`releases/qa/15867-dataclass-records`)
- [x] Add / update outdated documentation? (No user-facing documentation changes.)

### DIFF
--- a/conda/models/_legacy_record_fields.py
+++ b/conda/models/_legacy_record_fields.py
@@ -1,0 +1,195 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Legacy auxlib.Entity-based field descriptors for ``conda.models.records``.
+
+Kept around one release cycle so that plugins and downstreams importing the
+old descriptor names from :mod:`conda.models.records` continue to resolve.
+The dataclass-based record classes in :mod:`conda.models.records` no longer
+use any of these; they only exist to preserve import-time compatibility for
+code that still builds its own :class:`conda.auxlib.entity.Entity` subclasses.
+
+This module is imported lazily via ``deprecated.constant(factory=...)``.
+"""
+
+from __future__ import annotations
+
+from boltons.timeutils import dt_to_timestamp, isoparse
+
+from ..auxlib.entity import (
+    ComposableField,
+    EnumField,
+    ListField,
+    NumberField,
+    StringField,
+)
+from ..base.context import context
+from ..common.compat import isiterable
+from ..exceptions import PathNotFoundError
+from .channel import Channel
+from .enums import NoarchType, PackageType
+
+
+class NoarchField(EnumField):
+    def box(self, instance, instance_type, val):
+        return super().box(instance, instance_type, NoarchType.coerce(val))
+
+
+class TimestampField(NumberField):
+    def __init__(self):
+        super().__init__(default=0, required=False, default_in_dump=False)
+
+    @staticmethod
+    def _make_seconds(val):
+        if val:
+            val = val
+            if val > 253402300799:  # 9999-12-31
+                val /= (
+                    1000  # convert milliseconds to seconds; see conda/conda-build#1988
+                )
+        return val
+
+    @staticmethod
+    def _make_milliseconds(val):
+        if val:
+            if val < 253402300799:  # 9999-12-31
+                val *= 1000
+            val = val
+        return val
+
+    def box(self, instance, instance_type, val):
+        return self._make_seconds(super().box(instance, instance_type, val))
+
+    def dump(self, instance, instance_type, val):
+        return int(self._make_milliseconds(super().dump(instance, instance_type, val)))
+
+    def __get__(self, instance, instance_type):
+        try:
+            return super().__get__(instance, instance_type)
+        except AttributeError:
+            try:
+                return int(dt_to_timestamp(isoparse(instance.date)))
+            except (AttributeError, ValueError):
+                return 0
+
+
+class _FeaturesField(ListField):
+    def __init__(self, **kwargs):
+        super().__init__(str, **kwargs)
+
+    def box(self, instance, instance_type, val):
+        if isinstance(val, str):
+            val = val.replace(" ", ",").split(",")
+        val = tuple(f for f in (ff.strip() for ff in val) if f)
+        return super().box(instance, instance_type, val)
+
+    def dump(self, instance, instance_type, val):
+        if isiterable(val):
+            return " ".join(val)
+        else:
+            return val or ()
+
+
+class ChannelField(ComposableField):
+    def __init__(self, aliases=()):
+        super().__init__(Channel, required=False, aliases=aliases)
+
+    def dump(self, instance, instance_type, val):
+        if val:
+            return str(val)
+        else:
+            val = instance.channel
+            return str(val)
+
+    def __get__(self, instance, instance_type):
+        try:
+            return super().__get__(instance, instance_type)
+        except AttributeError:
+            url = instance.url
+            return self.unbox(instance, instance_type, Channel(url))
+
+
+class SubdirField(StringField):
+    def __init__(self):
+        super().__init__(required=False)
+
+    def __get__(self, instance, instance_type):
+        try:
+            return super().__get__(instance, instance_type)
+        except AttributeError:
+            try:
+                url = instance.url
+            except AttributeError:
+                url = None
+            if url:
+                return self.unbox(instance, instance_type, Channel(url).subdir)
+
+            try:
+                platform, arch = instance.platform.name, instance.arch
+            except AttributeError:
+                platform, arch = None, None
+            if platform and not arch:
+                return self.unbox(instance, instance_type, "noarch")
+            elif platform:
+                if "x86" in arch:
+                    arch = "64" if "64" in arch else "32"
+                return self.unbox(instance, instance_type, f"{platform}-{arch}")
+            else:
+                return self.unbox(instance, instance_type, context.subdir)
+
+
+class FilenameField(StringField):
+    def __init__(self, aliases=()):
+        super().__init__(required=False, aliases=aliases)
+
+    def __get__(self, instance, instance_type):
+        try:
+            return super().__get__(instance, instance_type)
+        except AttributeError:
+            try:
+                url = instance.url
+                fn = Channel(url).package_filename
+                if not fn:
+                    raise AttributeError()
+            except AttributeError:
+                fn = f"{instance.name}-{instance.version}-{instance.build}"
+            if not fn:
+                raise ValueError("Filename cannot be empty.")
+            return self.unbox(instance, instance_type, fn)
+
+
+class PackageTypeField(EnumField):
+    def __init__(self):
+        super().__init__(
+            PackageType,
+            required=False,
+            nullable=True,
+            default=None,
+            default_in_dump=False,
+        )
+
+    def __get__(self, instance, instance_type):
+        val = super().__get__(instance, instance_type)
+        if val is None:
+            noarch_val = instance.noarch
+            if noarch_val:
+                type_map = {
+                    NoarchType.generic: PackageType.NOARCH_GENERIC,
+                    NoarchType.python: PackageType.NOARCH_PYTHON,
+                }
+                val = type_map[NoarchType.coerce(noarch_val)]
+                val = self.unbox(instance, instance_type, val)
+        return val
+
+
+class Md5Field(StringField):
+    def __init__(self):
+        super().__init__(required=False, nullable=True)
+
+    def __get__(self, instance, instance_type):
+        try:
+            return super().__get__(instance, instance_type)
+        except AttributeError as e:
+            try:
+                return instance._calculate_md5sum()
+            except PathNotFoundError:
+                raise e
--- a/conda/models/_legacy_record_fields.py
+++ b/conda/models/_legacy_record_fields.py
@@ -72,6 +72,14 @@ class TimestampField(NumberField):
                 return 0
 
 
+class IndexedTimestampField(TimestampField):
+    def __get__(self, instance, instance_type):
+        try:
+            return NumberField.__get__(self, instance, instance_type)
+        except AttributeError:
+            return 0
+
+
 class _FeaturesField(ListField):
     def __init__(self, **kwargs):
         super().__init__(str, **kwargs)
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -278,18 +278,24 @@ def resolve_features(val: Any, record: PackageRecord) -> tuple[str, ...]:
     return tuple(f for f in (ff.strip() for ff in val) if f)
 
 
-def resolve_as_tuple(val: Any, record: PackageRecord) -> tuple:
+def resolve_as_tuple(
+    val: Any,
+    record: PackageRecord,
+    field_name: str = "sequence",
+) -> tuple:
     if isinstance(val, tuple):
         return val
     if not val:
         return ()
+    if isinstance(val, str):
+        raise ValidationError(field_name, val)
     return tuple(val)
 
 
 def resolve_optional_tuple(val: Any, record: PackageRecord) -> tuple | None:
     if val is None:
         return None
-    return resolve_as_tuple(val, record)
+    return resolve_as_tuple(val, record, "_requested_specs")
 
 
 def resolve_noarch(val: Any, record: PackageRecord) -> NoarchType | None:
@@ -387,10 +393,10 @@ class PackageRecord:
         "timestamp": resolve_timestamp,
         "track_features": resolve_features,
         "features": resolve_features,
-        "depends": resolve_as_tuple,
-        "constrains": resolve_as_tuple,
-        "flags": resolve_as_tuple,
-        "files": resolve_as_tuple,
+        "depends": partial(resolve_as_tuple, field_name="depends"),
+        "constrains": partial(resolve_as_tuple, field_name="constrains"),
+        "flags": partial(resolve_as_tuple, field_name="flags"),
+        "files": partial(resolve_as_tuple, field_name="files"),
         "noarch": resolve_noarch,
         "package_type": resolve_package_type,
         "platform": resolve_platform,
@@ -500,9 +506,9 @@ class PackageRecord:
                 setattr_(self, fname, sys.intern(val))
 
     def __setattr__(self, name: str, value: Any) -> None:
-        value = coerce_and_validate(name, value)
         if resolve := self.FIELD_RESOLVERS.get(name):
             value = resolve(value, self)
+        value = coerce_and_validate(name, value)
         object.__setattr__(self, name, value)
         if name in PKEY_FIELDS and hasattr(self, "_pkey_cache"):
             self.invalidate_pkey()
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import enum
 import json  # noqa: TID251
 import sys
-from dataclasses import MISSING, dataclass, field, fields
+from dataclasses import MISSING, dataclass, field, fields, is_dataclass
 from functools import cache, partial
 from pathlib import Path
 from typing import TYPE_CHECKING, ClassVar, Protocol, runtime_checkable
@@ -32,6 +32,7 @@ from ..auxlib.entity import (
     ListField,
     StringField,
 )
+from ..auxlib.exceptions import ValidationError
 from ..deprecations import deprecated
 from .channel import Channel
 from .enums import FileMode, LinkType, NoarchType, PackageType, PathEnum, Platform
@@ -39,6 +40,8 @@ from .enums import FileMode, LinkType, NoarchType, PackageType, PathEnum, Platfo
 if TYPE_CHECKING:
     from collections.abc import Callable
     from typing import Any
+
+    from .match_spec import MatchSpec
 
 
 class LinkTypeField(EnumField):
@@ -93,6 +96,50 @@ SENTINEL = object()
 
 TS_BOUNDARY = 253402300799  # 9999-12-31 in epoch seconds
 
+REQUIRED_FIELDS = ("name", "version", "build", "build_number")
+
+STRING_FIELDS = frozenset(
+    {
+        "name",
+        "version",
+        "build",
+        "subdir",
+        "fn",
+        "md5",
+        "legacy_bz2_md5",
+        "url",
+        "sha256",
+        "arch",
+        "preferred_env",
+        "python_site_packages_path",
+        "license",
+        "license_family",
+        "date",
+        "package_tarball_full_path",
+        "extracted_package_dir",
+        "requested_spec",
+        "auth",
+    }
+)
+
+INTEGER_FIELDS = frozenset({"build_number", "legacy_bz2_size", "size"})
+
+SEQUENCE_FIELDS = frozenset(
+    {
+        "depends",
+        "constrains",
+        "flags",
+        "track_features",
+        "features",
+        "files",
+        "_requested_specs",
+    }
+)
+
+PKEY_FIELDS = frozenset(
+    {"channel", "subdir", "name", "version", "build_number", "build", "fn"}
+)
+
 CACHE_FIELDS = frozenset({"_pkey_cache", "_hash_cache", "_memoized_md5"})
 
 FIELDS_EXCLUDED_FROM_DUMP = frozenset({"metadata"})
@@ -135,10 +182,10 @@ class Dumpable(Protocol):
 
 @cache
 def get_field_info(cls: type) -> tuple[tuple[str, Any, Any], ...]:
-    """Return (name, default, default_factory) for each public field of *cls*."""
+    """Return initialization metadata for each constructor field of *cls*."""
     info: list[tuple[str, Any, Any]] = []
     for f in fields(cls):
-        if f.name in CACHE_FIELDS:
+        if not f.init:
             continue
         default = f.default if f.default is not MISSING else SENTINEL
         default_factory = (
@@ -163,8 +210,8 @@ def get_dump_specs(cls: type) -> tuple[tuple[str, bool, Any], ...]:
 
 @cache
 def get_known_fields(cls: type) -> frozenset[str]:
-    """Return the set of non-cache field names for *cls*."""
-    return frozenset(f.name for f in fields(cls) if f.name not in CACHE_FIELDS)
+    """Return constructor field names for *cls*."""
+    return frozenset(f.name for f in fields(cls) if f.init)
 
 
 def resolve_channel(val: Any, record: PackageRecord) -> Channel | None:
@@ -209,6 +256,13 @@ def resolve_fn(val: Any, record: PackageRecord) -> str:
 
 
 def resolve_timestamp(val: Any, record: PackageRecord) -> int | float:
+    if not val and record.date:
+        from boltons.timeutils import dt_to_timestamp, isoparse
+
+        try:
+            return int(dt_to_timestamp(isoparse(record.date)))
+        except (TypeError, ValueError):
+            return 0
     if val and val > TS_BOUNDARY:
         return val / 1000
     return val
@@ -230,6 +284,12 @@ def resolve_as_tuple(val: Any, record: PackageRecord) -> tuple:
     if not val:
         return ()
     return tuple(val)
+
+
+def resolve_optional_tuple(val: Any, record: PackageRecord) -> tuple | None:
+    if val is None:
+        return None
+    return resolve_as_tuple(val, record)
 
 
 def resolve_noarch(val: Any, record: PackageRecord) -> NoarchType | None:
@@ -285,9 +345,34 @@ def dump_nested(val: Any) -> Any:
     return val.dump() if isinstance(val, Dumpable) else val
 
 
+def coerce_and_validate(name: str, val: Any) -> Any:
+    """Preserve the scalar validation performed by the old Entity fields."""
+    if val is None:
+        if name in REQUIRED_FIELDS:
+            raise ValidationError(name)
+        return val
+    if name in STRING_FIELDS:
+        if isinstance(val, (int, float, complex)):
+            return str(val)
+        if not isinstance(val, str):
+            raise ValidationError(name, val)
+    elif name in INTEGER_FIELDS and not isinstance(val, int):
+        raise ValidationError(name, val)
+    elif name == "timestamp" and not isinstance(val, (int, float)):
+        raise ValidationError(name, val)
+    elif name in SEQUENCE_FIELDS and not isinstance(val, tuple):
+        raise ValidationError(name, val)
+    return val
+
+
 @dataclass(slots=True, init=False, eq=False, repr=False)
 class PackageRecord:
-    """Representation of a concrete package archive (tarball or .conda file)."""
+    """Representation of a concrete package archive (tarball or .conda file).
+
+    The subclasses add information for solved, cached, and installed packages,
+    but do not add fields to :attr:`_pkey`. Records with the same identity
+    fields therefore compare equal and produce the same hash across subclasses.
+    """
 
     ALIASES: ClassVar[dict[str, str]] = {
         "build_string": "build",
@@ -358,7 +443,6 @@ class PackageRecord:
     size: int | None = None
 
     metadata: set[str] = field(default_factory=set, repr=False)
-
     _pkey_cache: tuple | None = field(
         default=None, repr=False, compare=False, init=False
     )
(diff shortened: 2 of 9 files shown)


conda/models/channel.py
```diff
@@ -48,31 +49,7 @@
 log = getLogger(__name__)
 
 
-class ChannelType(type):
```
dholth, line 51: Finally

conda/models/channel.py
```diff
@@ -82,11 +59,20 @@ class Channel(metaclass=ChannelType):
...
     """
 
-    _cache_ = {}
-
     @staticmethod
     def _reset_state() -> None:
-        Channel._cache_ = {}
+        Channel.from_value.cache_clear()
+
+    def __new__(cls, *args, **kwargs):
```
dholth, line 66: I've looked into this but was unsure whether `__new__` was doing everything that the metaclass was expected to do; including when `pickle` or subclassing were in play.  If we can make this work, fantastic.

conda/models/channel.py
```diff
@@ -97,7 +83,10 @@ def __init__(
         name: str | None = None,
         platform: str | None = None,
         package_filename: str | None = None,
+        **_extra: Any,
     ):
+        if self.__dict__:
```
dholth, line 88: Does this break subclassing? (Does that matter?)

conda/models/channel.py
```diff
@@ -127,6 +116,7 @@ def from_channel_name(channel_name: str) -> Channel:
         return _get_channel_for_name(channel_name)
 
     @staticmethod
+    @cache
```
dholth, line 119: magic go fast decorator

conda/models/channel.py
```diff
@@ -424,20 +414,25 @@ def dump(self) -> dict[str, Any]:
...
-        if platform:
-            channels = (
-                Channel(**{**channel.dump(), "platform": platform})
-                for channel in channels
-            )
-        self._channels = tuple(channels)
+        if channels is not None:
+            if platform:
+                channels = (
+                    Channel(**{**channel.dump(), "platform": platform})
```
dholth, line 430: 🤩 `**{**` 🤩

A star programmer wrote this.

jezdez: LOL so much for artifical "intelligence"

dholth: Some of my code review comments are observations unrelated to code quality.

kenodegard (commented):
I think this is a large enough change that we should and can split this further, e.g., the Channel refactor can be done in a separate PR from the records refactor.

Also what are we doing regarding deprecations?

conda/models/channel.py
kenodegard: It depends on the ordering in the subclass, e.g., the following would break:

```python
class SubChannel(Channel):
        self.value = ...
```

In the case of `MultiChannel` we don't call `super().__init__` so currently it doesn't matter

jezdez:
Good call @kenodegard, split done.

#15942 (A19a): channel refactor only (`conda/models/channel.py` + `tests/models/test_channel.py`, ~90 LOC). Replaces the `ChannelType` metaclass with `Channel.__new__` plus `@functools.cache` on `Channel.from_value`, plus a `deprecated.constant` shim for `ChannelType` so `isinstance(x, ChannelType)` keeps working one cycle.

This PR (now A19b): records refactor only. Rebased on `main`, dropped the channel diff, description updated to reflect the narrower scope. The two PRs are independent and can merge in either order.

On deprecations, I used a mix of shim strategies depending on risk:

1. `ChannelType` metaclass itself (in #15942): `deprecated.constant("26.9", "27.3", "ChannelType", type(Channel), ...)`. Downstream code doing `isinstance(x, ChannelType)` keeps working for free; anyone subclassing via `class Foo(metaclass=ChannelType)` gets a `PendingDeprecationWarning` and can migrate to a plain class.

2. Record field descriptor classes (`SubdirField`, `ChannelField`, `FilenameField`, `NoarchField`, `TimestampField`, `PackageTypeField`, `_FeaturesField`, `Md5Field`): moved verbatim into a new `conda/models/_legacy_record_fields.py` and re-exported via `deprecated.constant(factory=partial(_legacy_field, name))` lazy shims. The new records don't pay the import cost (`boltons.timeutils` etc. only load on first deprecated access), and any plugin or downstream still doing `from conda.models.records import ChannelField` gets a working class with a `PendingDeprecationWarning`. Covered by a parametrised test over all eight names.

3. `PackageCacheRecord.md5` auto-compute-on-read (the one real behaviour change): preserved one cycle via a module-level `property` shim that reads the inherited slot, falls through to `calculate_md5sum()` when unset, and routes the warning through `deprecated.topic("26.9", "27.3", ...)` so the deprecations framework picks the appropriate category per release. The property is assigned after the class body because `@dataclass(slots=True)` strips any class-body attribute whose name matches an inherited field before allocating slots, which would otherwise silently drop a `md5` property defined inline. The dataclass-slots workaround is contained inside `_make_md5_auto_compute_property()` so no helper symbols leak into `conda.models.records`. Covered by two tests (one for the auto-compute path, one parametrised over the explicit-set paths to confirm they stay silent).

Also found and fixed a latent bug while wiring up the `md5` shim: `_memoized_md5` is in `CACHE_FIELDS` and so was never initialised by `PackageRecord.__init__`, which would have raised `AttributeError` on the first `calculate_md5sum()` call on a fresh `PackageCacheRecord`. Switched that slot read to `getattr(self, "_memoized_md5", None)`.

`news/15867-channel-metaclass` and `news/15867-dataclass-records` cover the enhancements and deprecations under the right headings.

conda/models/records.py
```diff
@@ -543,201 +656,197 @@ def virtual_package(
...
+    property; the getter falls through to :meth:`calculate_md5sum` and uses
+    the ``deprecated`` framework to nudge callers toward the explicit method.
 
-    requested_spec = StringField(required=False)
-    """
-    The :class:`MatchSpec` that the user requested or ``None``
-    if the package it was installed as a dependency.
-    """
-    _requested_specs = ListField(
-        str, required=False, nullable=True, aliases=("requested_specs",)
```
kenodegard: The alias here meant the old object could be defined as:

`SolvedRecord(requested_specs=...)`

conda/models/records.py
```diff
@@ -240,203 +85,513 @@ class PathDataV1(PathData):
...
-    # from info/paths.json
     paths_version = IntegerField()
     paths = ListField(PathDataV1)
 
 
-class PackageRecord(DictSafeMixin, Entity):
-    """Representation of a concrete package archive (tarball or .conda file).
+SENTINEL = object()
+
+TS_BOUNDARY = 253402300799  # 9999-12-31 in epoch seconds
```
dholth, line 97: magic

dholth: It's fortunate that this is earlier than 1980-01-01 the earliest DOS, zipfile timestamp.
```
>>> print(datetime.datetime.fromtimestamp(253402300799/1000))
1978-01-11 16:31:40.799000
```

jezdez: ✨

jaimergp:
> slots=True means callers can't attach arbitrary attributes to record instances

Probably a bad idea, but if this becomes a problem: can we leave an `extra` attribute for arbitrary writes?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+        return " ".join(val) if val else ()
+    return val or ()
+
+
+def dump_enum(val: Any) -> Any:
+    return val.value if isinstance(val, enum.Enum) else val
+
+
+def dump_nested(val: Any) -> Any:
+    return val.dump() if isinstance(val, Dumpable) else val
```
jaimergp, line 360: Do we need all of these as public? They seem to be mostly internal to the module.

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+            noarch_coerced = (
+                noarch_val
+                if isinstance(noarch_val, NoarchType)
+                else NoarchType.coerce(noarch_val)
+            )
+            return NOARCH_TO_PACKAGE_TYPE.get(noarch_coerced)
+        return None
+
+    @property
+    def _pkey(self) -> tuple:
```
jaimergp, line 562: This lost its docstring. Intentional?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+        ]
+        from ..base.context import context
+
+        if context.separate_format_cache:
+            pk.append(self.fn)
+        result = tuple(pk)
+        object.__setattr__(self, "_pkey_cache", result)
+        return result
+
+    def invalidate_pkey(self) -> None:
```
jaimergp, line 590: Odd that `_pkey` is private but its invalidator is public.

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
             "platform": self.subdir,
             "version": self.version,
         }
 
-    arch = StringField(required=False, nullable=True)  # so legacy
-    platform = EnumField(Platform, required=False, nullable=True)  # so legacy
-
-    depends = ListField(str, default=())
-    constrains = ListField(str, default=())
+    def to_match_spec(self):
```
jaimergp:
```suggestion
    def to_match_spec(self) -> MatchSpec:
```

(may need the TYPE_CHECKING import above)

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+            channel=self.channel,
+            subdir=self.subdir,
+            name=self.name,
+            version=self.version,
+            build=self.build,
+        )
 
-    track_features = _FeaturesField(required=False, default=(), default_in_dump=False)
-    features = _FeaturesField(required=False, default=(), default_in_dump=False)
+    def to_simple_match_spec(self):
```
jaimergp:
```suggestion
    def to_simple_match_spec(self) -> MatchSpec:
```

(may need the TYPE_CHECKING import above)

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+
+def dump_enum(val: Any) -> Any:
+    return val.value if isinstance(val, enum.Enum) else val
+
+
+def dump_nested(val: Any) -> Any:
+    return val.dump() if isinstance(val, Dumpable) else val
+
+
+@dataclass(slots=True, init=False, eq=False, repr=False)
```
jaimergp, line 385: Note: init, eq and repr are False because no generation is needed since the class provides its own.

conda/models/records.py
```diff
@@ -548,165 +667,137 @@ def virtual_package(
...
-
-    package_tarball_full_path = StringField()
-    """Full path to the local package file."""
-
-    extracted_package_dir = StringField()
-    """Full path to the local extracted package."""
-
-    md5 = Md5Field()
-    """The md5 checksum of the package.
+    """Representation of a package in the local package cache."""
```
jaimergp: Why trim the docstring so much?

conda/models/records.py
```diff
@@ -548,165 +667,137 @@ def virtual_package(
...
+        # path reachable on first call.
+        memoized = getattr(self, "_memoized_md5", None)
+        if memoized:
+            return memoized
+        if Path(self.package_tarball_full_path).is_file():
             from ..gateways.disk.read import compute_sum
 
             md5sum = compute_sum(self.package_tarball_full_path, "md5")
-            setattr(self, "_memoized_md5", md5sum)
+            object.__setattr__(self, "_memoized_md5", md5sum)
```
jaimergp, line 857: I don't understand the need for this change :thinking:

conda/models/records.py
```diff
@@ -548,165 +667,137 @@ def virtual_package(
...
+        spec = self.requested_spec
+        if spec:
             return (spec,)
         return ()
 
-    def dump(self):
-        """
-        We need to expose _requested_specs as its public counterpart,
-        and also expose single specs as a plural list for backwards compatibility.
-        """
```
jaimergp, line 673: Again, not following what happened to all these docstrings. Maybe they were not very informative, but isn't there any value in them being kept a comments?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+    python_site_packages_path: str | None = None
+    license: str | None = None
+    license_family: str | None = None
+    package_type: PackageType | None = None
+
+    timestamp: int | float = 0
+    date: str | None = None
+    size: int | None = None
+
+    metadata: set[str] = field(default_factory=set, repr=False)
```
jaimergp, line 465: PR description says:

> metadata defaults to None instead of a set() factory (avoids 50k empty-set allocations).

Yet this one is a field with a `set` factory. Which one is stale? Should this be a property backed by a private counterpart to instantiate the set on first access (e.g. if something is doing `record.metadata.add(...)`)?

conda/models/records.py
jaimergp: Also, should they be frozen too?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+                noarch_val
+                if isinstance(noarch_val, NoarchType)
+                else NoarchType.coerce(noarch_val)
+            )
+            return NOARCH_TO_PACKAGE_TYPE.get(noarch_coerced)
+        return None
+
+    @property
+    def _pkey(self) -> tuple:
+        cached = self._pkey_cache
```
jaimergp, line 569: We are assuming the fields will never be mutated. This can lead to a stale identity key if someone does `record.build_number += 1`. Invalidate cache via `__setattr__` or freeze/forbid assignment?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+                if resolved is not val:
+                    setattr_(self, name, resolved)
 
-        Part of the :attr:`_pkey`.
-        """
-        return getattr(self.channel, "canonical_name", None)
+        for fname in INTERN_FIELDS:
+            val = getattr(self, fname, None)
+            if isinstance(val, str):
+                setattr_(self, fname, sys.intern(val))
```
jaimergp, line 520: TIL, neat!

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+                    setattr_(self, name, default_factory())
+                else:
+                    setattr_(self, name, default)
+            else:
+                setattr_(self, name, val)
 
-    @property
-    def channel_name(self) -> str | None:
-        """str: The canonical name of the channel of this package.
+        for name, resolve in self.FIELD_RESOLVERS.items():
```
jaimergp, line 507: Do we care about order here? Should it be documented or tested?

conda/models/records.py
```diff
@@ -548,165 +667,137 @@ def virtual_package(
...
-
-        if isfile(self.package_tarball_full_path):
+    def tarball_basename(self) -> str:
+        return Path(self.package_tarball_full_path).name
+
+    def calculate_md5sum(self) -> str | None:
+        # ``_memoized_md5`` is a slot but is not populated by PackageRecord's
+        # __init__ (it's listed in CACHE_FIELDS which skips initialization),
+        # so on a freshly-constructed PackageCacheRecord the slot read would
+        # raise AttributeError. Use getattr with a default to keep the slow
```
jaimergp: Why not initialise it to None anyway?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+    features: tuple[str, ...] = ()
+
+    noarch: NoarchType | None = None
+    preferred_env: str | None = None
+    python_site_packages_path: str | None = None
+    license: str | None = None
+    license_family: str | None = None
+    package_type: PackageType | None = None
+
+    timestamp: int | float = 0
```
jaimergp, line 460: `timestamp` used to fallback to `date` if absent:

```
    def __get__(self, instance, instance_type):
        try:
            return super().__get__(instance, instance_type)
        except AttributeError:
            try:
                return int(dt_to_timestamp(isoparse(instance.date)))
            except (AttributeError, ValueError):
                return 0
```

This is not the case anymore and should be either corrected or at least documented.

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+                src = obj.dump()
+            elif hasattr(obj, "__dict__"):
+                src = obj.__dict__
+            elif hasattr(obj, "__slots__"):
+                src = {
+                    s: getattr(obj, s)
+                    for s in obj.__slots__
+                    if hasattr(obj, s) and not s.startswith("_")
+                }
+            else:
```
jaimergp: What about the docstring via `__doc__`? Should we try to preserve it?

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+        known = get_known_fields(cls)
+
+        for obj in objects:
+            if isinstance(obj, dict):
+                src = obj
+            elif isinstance(obj, Dumpable):
+                src = obj.dump()
+            elif hasattr(obj, "__dict__"):
+                src = obj.__dict__
+            elif hasattr(obj, "__slots__"):
```
jaimergp: There are a few warnings on accessing this directly on [the official docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass), which recommend using `fields()` instead. e.g. to make sure we are not missing parent class slots

conda/models/records.py
```diff
@@ -447,66 +600,27 @@ def combined_depends(self):
...
-    def to_simple_match_spec(self):
-        return MatchSpec(
-            name=self.name,
-            version=self.version,
-        )
-
     @property
-    def namekey(self):
-        return "global:" + self.name
+    def namekey(self) -> str:
```
jaimergp, line 742: TIL about `namekey`. No clue what this is for. It's only [used in `conda.core.link`](https://github.com/conda/conda/blob/4f8feabe660fa6967f28bb2b56a4ab4421defd18/conda/core/link.py#L1387) for some obscure sorting? I think it's worth raising an issue about this probably unnecessary property so it goes through a deprecation cycle.

conda/models/records.py
```diff
@@ -447,66 +600,27 @@ def combined_depends(self):
...
+        return f"global:{self.name}"
 
     @property
-    def spec(self):
-        """Return package spec: name=version=build"""
+    def spec(self) -> str:
         return f"{self.name}={self.version}={self.build}"
```
jaimergp, line 748: Hm, shouldn't this have a double equals for the version? (I know this is not a regression in this PR, just found it odd).

conda/models/records.py
```diff
@@ -240,205 +85,513 @@ class PathDataV1(PathData):
...
+
+    metadata: set[str] = field(default_factory=set, repr=False)
+
+    _pkey_cache: tuple | None = field(
+        default=None, repr=False, compare=False, init=False
     )
-    """The sha256 checksum of the package."""
+    _hash_cache: int | None = field(default=None, repr=False, compare=False, init=False)
+
+    def __init__(self, **kwargs: Any) -> None:
```
jaimergp, line 471: We are not catching any unknown kwargs. See in this PR:

```python
>>> from conda.models.records import PackageRecord
>>> PackageRecord(jaime=True)
PackageRecord(build_number=0, channel=Channel("<unknown>"), subdir='osx-arm64', fn='--', timestamp=0)
```

In main:

```python
>>> from conda.models.records import PackageRecord
>>> PackageRecord(jaime=True)
    PackageRecord(jaime=True)
    ~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 754, in __call__
    instance = super().__call__(*args, **kwargs)
  File "/Users/jrodriguez/devel/conda/conda/models/records.py", line 504, in __init__
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 779, in __init__
    raise ValidationError(
    ...<3 lines>...
    )
conda.auxlib.exceptions.ValidationError: PackageRecord requires a name field. Instantiated with {'jaime': True}
```

jaimergp: Doesn't seem to error out if `name` is not provided either:

```python
>>> from conda.models.records import PackageRecord
>>> PackageRecord()
PackageRecord(build_number=0, channel=Channel("<unknown>"), subdir='osx-arm64', fn='--', timestamp=0)
```

Main:

```python
>>> PackageRecord()
    PackageRecord()
    ~~~~~~~~~~~~~^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 754, in __call__
    instance = super().__call__(*args, **kwargs)
  File "/Users/jrodriguez/devel/conda/conda/models/records.py", line 504, in __init__
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 779, in __init__
    raise ValidationError(
    ...<3 lines>...
    )
conda.auxlib.exceptions.ValidationError: PackageRecord requires a name field. Instantiated with {}
```

jaimergp: This PR doesn't fail for wrong types:

```python
>>> PackageRecord(name="a", version="1", build="a4040", build_number="4040")
PackageRecord(name='a', version='1', build='a4040', build_number='4040', channel=Channel("<unknown>"), subdir='osx-arm64', fn='a-1-a4040', timestamp=0)
```

But main did:

```python
>>> PackageRecord(name="a", version="1", build="a4040", build_number="4040")
    PackageRecord(name="a", version="1", build="a4040", build_number="4040")
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 754, in __call__
    instance = super().__call__(*args, **kwargs)
  File "/Users/jrodriguez/devel/conda/conda/models/records.py", line 504, in __init__
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 770, in __init__
    setattr(self, key, kwargs[key])
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 437, in __set__
    instance.__dict__[self.name] = self.validate(
                                   ~~~~~~~~~~~~~^
        instance,
        ^^^^^^^^^
        self.box(instance, instance.__class__, val),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/jrodriguez/devel/conda/conda/auxlib/entity.py", line 482, in validate
    raise ValidationError(getattr(self, 'name', 'undefined name'), val)
conda.auxlib.exceptions.ValidationError: Invalid value 4040 for build_number
```

jaimergp (changes requested):
I have several comments but one of them is critical: the new `__init__` doesn't maintain feature parity with its predecessor.

- No validation on required fields being absent
- No validation on passed values having the wrong type

For those we should only raise `ValidationError` to preserve the exception API.

jaimergp:
Also, have you considered other data structure backends for this? This [blog post](https://www.crumpledpaper.tech/2026-06-21-python-struct-profiling/) went deep into it, and dataclasses are not that bad, but there are cheaper options. You wrote a lot of the magic for the dataclass, why not go all the way with a normal slotted class or a [manual record](https://www.crumpledpaper.tech/2026-06-21-python-struct-profiling/#manual-record)? If it's about accuracy, then those things can be caught in tests.

conda/models/records.py
jaimergp: Ah, is it to avoid going through `self.__setattr__`?

conda/models/records.py
jezdez: Kept public as part of the resolver API.

conda/models/records.py
jezdez: Applied in 7621b994f. Restored the `_pkey` docstring.

conda/models/records.py
jezdez: Agreed, this only needs to be internal. Renamed to `_invalidate_pkey()` in 722272f1f. Identity-field writes still invalidate both caches automatically.

conda/models/records.py
jezdez: Applied in 7621b994f.

conda/models/records.py
jezdez: Applied in 7621b994f.

conda/models/records.py
jezdez: Kept mutable for compatibility. Identity-field writes now invalidate both caches.

conda/models/records.py
jezdez: Applied. Restored the cache-record and property documentation.

conda/models/records.py
jezdez: Yes. This cache write intentionally bypasses `__setattr__`.

conda/models/records.py
jezdez: Applied. Restored the `SolvedRecord` and requested-spec documentation.

conda/models/records.py
jezdez: Kept the `set()` factory and corrected the stale PR description.

conda/models/records.py
jezdez: Applied. Identity-field writes now invalidate `_pkey` and hash caches.

conda/models/records.py
jezdez: Thanks!

conda/models/records.py
jezdez: No ordering dependency here. These field assignments are independent.

conda/models/records.py
jezdez: Applied. `_memoized_md5` is now initialized to `None`.

conda/models/records.py
jezdez: Applied. A missing timestamp now falls back to `date`, while an explicit zero stays zero.

conda/models/records.py
jezdez: `from_objects()` only copies record fields. Class docstrings are not record data.

conda/models/records.py
jezdez: Applied. Dataclass sources now use `dataclasses.fields()`, including inherited fields.

conda/models/records.py
jezdez: Filed #16431.

conda/models/records.py
jezdez: Unchanged here. This preserves the existing `name=version=build` syntax.

conda/models/records.py
jezdez: Applied in 7621b994f: required fields and scalar types are validated again. Unknown kwargs remain ignored when required fields are present, matching `main`.

jezdez:
@jaimergp Leaving `extra` out until a concrete caller needs it. It adds machinery without preserving arbitrary attribute access. I also checked the linked post. Its main benchmark is type-definition and startup cost, not A19b’s instance, dump, and memory path. Its appendix puts dataclass and native slots at the same interpreted construction ratio, with the manual record slower. Keeping the dataclass.

jezdez:
@jaimergp The compatibility fixes from your review are included in the rebased head a99943af6. Fresh CI is running. Could you take another look?

dholth:
I like it.

jezdez:
@dholth Yay! Any chance you could take a look at it?

conda/models/_legacy_record_fields.py
```diff
@@ -0,0 +1,203 @@
...
+
+class TimestampField(NumberField):
+    def __init__(self):
+        super().__init__(default=0, required=False, default_in_dump=False)
+
+    @staticmethod
+    def _make_seconds(val):
+        if val:
+            val = val
+            if val > 253402300799:  # 9999-12-31
```
dholth, line 45: excessively futureproof

jezdez: That cutoff comes from the existing timestamp field. I’ve kept it unchanged here to preserve its seconds/milliseconds conversion behavior.

PR #15917, opened 2026-04-15
Replace deepcopy with dict comprehension in solver (A20)

jezdez:
### Description

`Resolve._get_reduced_index()` called `copy.deepcopy(specs_by_name_seed)` inside a loop over every explicit spec. `MatchSpec` objects are effectively immutable (they freeze their state on construction), so the values in the lists do not need to be copied — only the container lists do. A slice copy (`v[:]`) of each list is sufficient and removes the `import copy` from the module.

Split out from #15891 to keep the solver change focused and separately reviewable.

Part of #15867.

**Measured impact:** `deepcopy` → shallow copy is 11.7× faster (0.6 ms saved per solve with 10 explicit packages).

### Checklist - did you ...

- [x] Add a file to the `news` directory (using the template) for the next release's release notes?
- ~~Add / update outdated documentation?~~ (not applicable)

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -7,7 +7,6 @@ See conda.core.solver.Solver for the high-level API.
 
 from __future__ import annotations
 
-import copy
 import itertools
 from collections import defaultdict, deque
 from functools import cache
@@ -839,7 +838,7 @@ class Resolve:
                 #    broadening check to apply across packages at the explicit level; only
                 #    at the level of deps below that explicit package.
                 seen_specs = set()
-                specs_by_name = copy.deepcopy(specs_by_name_seed)
+                specs_by_name = {k: list(v) for k, v in specs_by_name_seed.items()}
 
                 dep_specs = set(self.ms_depends(pkg))
                 for dep in dep_specs:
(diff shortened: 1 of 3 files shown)


conda/resolve.py
```diff
@@ -839,7 +838,7 @@ def filter_group(_specs):
                 #    broadening check to apply across packages at the explicit level; only
                 #    at the level of deps below that explicit package.
                 seen_specs = set()
-                specs_by_name = copy.deepcopy(specs_by_name_seed)
+                specs_by_name = {k: list(v) for k, v in specs_by_name_seed.items()}
```
dholth: `v[:]` is exactly the same in Python if v is a list?

tests/test_resolve.py
```diff
@@ -16,3 +17,16 @@ def test_Resolve_make_channel_priorities():
...
+def test_specs_by_name_copy_is_independent() -> None:
+    """Copying specs_by_name with a dict comprehension must yield independent lists."""
+    seed = {
+        "numpy": [MatchSpec("numpy>=1.20")],
+        "scipy": [MatchSpec("scipy")],
+    }
+    copy = {k: list(v) for k, v in seed.items()}
+
+    copy["numpy"].append(MatchSpec("numpy<2"))
+    assert len(seed["numpy"]) == 1, "seed must not be mutated by changes to the copy"
```
dholth, line 32: An "is" check would also confirm that the lists are separate objects.

dholth (approved):

PR #15918, opened 2026-04-15
Replace CodSpeed benchmarks with Bencher

jezdez:
### Description

Replace the CodSpeed benchmark CI path with Bencher.

- Keeps the existing `linux-benchmarks` check as the single benchmark job and removes the side-by-side `linux-bencher-benchmarks` job.
- Runs `@pytest.mark.benchmark` tests with `pytest-benchmark` and uploads `benchmark_results.json` through `bencherdev/bencher`.
- Installs `pytest-benchmark` only in the benchmark job, so normal unit and integration jobs do not auto-load the plugin.
- Excludes benchmark-marked tests from the regular unit and integration matrix; the Bencher job is responsible for those tests.
- Updates benchmark type hints and local run instructions away from `pytest-codspeed`.
- Includes a small argparse assertion fix for Python builds that still quote `argparse` choices.

Requires `BENCHER_API_TOKEN` and `BENCHER_PROJECT_SLUG` repository secrets.

Part of #15867.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory (using the template) for the next release's release notes?~ (CI-only change)
- [ ] ~Add / update outdated documentation?~ (CI infrastructure only)


kenodegard:
This is difficult to evaluate as a single PR, we probably just need to run CodSpeed and Bencher in parallel for a week or so?

jezdez:
@kenodegard Yeah, I was wondering the same. From the onboarding experience, etc., Bencher seems more complete, and it has a ton of config options to be more deliberate about how we run performance tests.

jezdez:
Good call, pivoted to that. The PR now *adds* `linux-bencher-benchmarks` next to the existing CodSpeed job instead of replacing it, so both services run on every push to `main` and every PR for the length of our evaluation. A few things worth noting:

- Tests aren't forked: `@pytest.mark.benchmark` bodies are wrapped in a `benchmark.pedantic(...)` closure that works with either plugin (CodSpeed replaces the fixture under `--codspeed`, pytest-benchmark handles it otherwise). Locally you can still run `pytest --codspeed tests/cli/test_startup_benchmarks.py` or drop the flag and get wall-clock numbers.
- Bencher PR comment had a `NO THRESHOLD` warning on the previous revision because the `--threshold-*` flags were only on the push path. Both paths now configure a t-test threshold, and the PR path adds `--ci-only-thresholds` so the comment stays quiet until a baseline exists.
- Both Bencher steps gate on `env.BENCHER_API_TOKEN != ''`, so the job is inert on forks / before the secrets are set.

Once we have ~a week or two of both running on `main`, we can pick the one with cleaner signal on our startup-latency work (#15867) and rip out the other. Happy to drop the loser in a follow-up.

.github/workflows/tests.yml
```diff
@@ -348,11 +348,18 @@ jobs:
...
     needs: changes
     if: needs.changes.outputs.code == 'true'
 
     runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      checks: write
+      pull-requests: write
+    env:
+      BENCHER_API_TOKEN: ${{ secrets.BENCHER_API_TOKEN }}
```
travishathaway: @jezdez,

I had a question for you on this approach. The Bencher docs recommend not directly including API Tokens for pull requests that can be created from forks (as is the case for conda):

- https://bencher.dev/docs/how-to/github-actions/#pull-requests-from-forks

I'm wondering how that above "permissions" block affects this though. Is it only going to run for users, like maintainers, who have write access? Or does this still leave the door open for pwn requests as mentioned in the Bencher documentation?

kenodegard: Yes, I agree with adopting the two workflow approach, one for the testing itself (which can checkout forked code) and one for the upload (which will have the upload secret)

jezdez: I've made the split

travishathaway (approved):
I'm feeling pretty good about this. We'll probably have to tweak things here and there until we've setup proper dashboards in Bencher and have been gathering data for a bit.

PR #15925, opened 2026-04-16
Add `factory=` kwarg to `deprecated.constant`

jezdez:
### Description

Add a keyword-only `factory=` argument to `conda.deprecations.deprecated.constant` as a mutually-exclusive alternative to positional `value`. When given, `factory` must be a zero-argument callable; it is invoked the first time the deprecated symbol is accessed and the result is cached for subsequent accesses so identity is preserved.

#### Motivation

Today `deprecated.constant(deprecate_in, remove_in, name, value, ...)` captures `value` eagerly at registration time. This is fine for cheap values (strings, ints, already-imported classes) but breaks down when the value requires a heavy import — typical for "this symbol was moved to another module" deprecations.

Concrete example from the conda startup work (#15867): `conda.common.serialize.yaml.CondaYAMLRepresenter` was previously a module-level class. Moving it inside `_yaml()` to defer `ruamel.yaml` (32 modules) is a clean win, but registering it via `deprecated.constant(..., CondaYAMLRepresenter)` would force `ruamel.yaml` to import at registration time, erasing the gain. Other in-flight conda PRs hit the same problem and had to either (a) delete `deprecated.constant` calls and hand-roll `warnings.warn` (#15893, A22) or (b) hand-roll a custom `__getattr__` that lazily calls `deprecated.constant` on first access and is silent on that first access due to call-timing (#15868, A2/A3).

`factory=` lets all of these stay on the canonical public API.

#### API

The shape mirrors `dataclasses.field(default=..., default_factory=...)` — two mutually-exclusive ways to provide the underlying value:

```python


def _make_heavy():
    import some_heavy_dependency
    return some_heavy_dependency.Thing


deprecated.constant(
    "26.9", "27.3",
    "HEAVY_THING",
    factory=_make_heavy,
    addendum="Use `lightweight_alternative` instead.",
)
```

```pycon
>>> import foo  # some_heavy_dependency is NOT imported
>>> foo.HEAVY_THING  # some_heavy_dependency imports now, and only now
<stdin>:1: PendingDeprecationWarning: foo.HEAVY_THING is pending deprecation and will be removed in 27.3. Use `lightweight_alternative` instead.
>>> foo.HEAVY_THING is foo.HEAVY_THING  # cached, identity preserved
True
```

- Passing neither `value` nor `factory=` is a `TypeError`.
- Passing both `value` and `factory=` is a `TypeError`.
- Passing `factory=` with a non-callable is a `TypeError`.
- All existing `deprecated.constant(...)` call sites keep working unchanged.

#### Why only `constant`?

I reviewed every `DeprecationHandler` API:

| API | "Value" captured? | Benefits from a factory? |
|---|---|---|
| `@deprecated(...)` | the decorated callable | ❌ the callable is defined in the module being loaded |
| `@deprecated.argument(...)` | no value | ❌ |
| `deprecated.action(...)` | an `Action` class | ❌ argparse is already imported at the callsite |
| `deprecated.module(...)` | no value, fires at import | ❌ |
| `deprecated.topic(...)` | no value | ❌ |
| `deprecated.constant(...)` | eager `value` | ✅ this PR |

`constant` is the only API where the captured value can trigger arbitrary imports. The other APIs either take no value, or take something that by construction is already imported.

#### Tests

- `test_constant_factory` — registration does not invoke the factory; first access invokes it exactly once and emits the warning; subsequent accesses return the cached object (identity preserved) and still warn.
- `test_constant_factory_rejects_non_callable` — registering `factory=42` raises `TypeError` with a clear message.
- `test_constant_requires_value_or_factory` — passing neither is a `TypeError`.
- `test_constant_rejects_value_and_factory_together` — passing both is a `TypeError`.

#### Follow-ups

Open PRs in the conda-startup track that rebase on this and drop their bespoke shims in favor of `deprecated.constant(..., factory=...)`:

- #15882 (A10, driving case) — `conda.common.serialize.yaml.CondaYAMLRepresenter`.
- #15868 (A2/A3) — removes a custom `__getattr__` + "register-on-first-access" trick in `conda/cli/conda_argparse.py` (the current trick also has a minor bug where the first access is silent; `factory=` fixes it).
- #15893 (A22) — restores the 16 canonical `deprecated.constant` calls in `conda/plugins/__init__.py` that had to be deleted because they eagerly pulled in `conda.plugins.types`.
- #15926 (A24) — canonicalizes the remaining open-coded shims in `conda/auxlib/logz.py` and `conda/common/serialize/__init__.py` that eagerly import `conda.common.serialize.json` today.

### Checklist - did you ...

- [x] Add / update outdated documentation? (`docs/source/dev-guide/deprecations.md`)

### DIFF
--- a/conda/deprecations.py
+++ b/conda/deprecations.py
@@ -31,6 +31,9 @@ class DeprecatedError(RuntimeError):
     pass
 
 
+_UNSET = object()
+
+
 # inspired by deprecation (https://deprecation.readthedocs.io/en/latest/) and
 # CPython's warnings._deprecated
 class DeprecationHandler:
@@ -289,8 +292,9 @@ class DeprecationHandler:
         deprecate_in: str,
         remove_in: str,
         constant: str,
-        value: Any,
+        value: Any = _UNSET,
         *,
+        factory: Callable[[], Any] | None = None,
         addendum: str | None = None,
         stack: int = 0,
         deprecation_type: type[Warning] = DeprecationWarning,
@@ -300,10 +304,26 @@ class DeprecationHandler:
         :param deprecate_in: Version in which code will be marked as deprecated.
         :param remove_in: Version in which code is expected to be removed.
         :param constant:
-        :param value:
+        :param value: The value to return when the deprecated constant is
+            accessed. Mutually exclusive with ``factory``.
+        :param factory: A zero-argument callable invoked the first time the
+            deprecated constant is accessed; the result is cached. Use this to
+            avoid paying registration-time costs (e.g., heavy imports) for a
+            symbol that may never be accessed. Mutually exclusive with ``value``.
         :param addendum: Optional additional messaging. Useful to indicate what to do instead.
         :param stack: Optional stacklevel increment.
         """
+        if (value is _UNSET) == (factory is None):
+            raise TypeError(
+                f"deprecated.constant({constant!r}, ...): pass exactly one of "
+                "`value` or `factory=`."
+            )
+        if factory is not None and not callable(factory):
+            raise TypeError(
+                f"deprecated.constant({constant!r}, factory=...): `factory` "
+                "must be a zero-argument callable."
+            )
+
         # detect calling module
         module, fullname = self._get_module(stack)
         # detect function name and generate message
@@ -329,7 +349,14 @@ class DeprecationHandler:
             deprecations = _ConstantDeprecationRegistry(fullname, fallback)
             module.__getattr__ = deprecations  # type: ignore[method-assign]
 
-        deprecations.register(constant, message, category, stack, value)
+        deprecations.register(
+            constant,
+            message,
+            category,
+            stack,
+            value if factory is None else factory,
+            factory=factory is not None,
+        )
 
     def topic(
         self: Self,
@@ -447,7 +474,7 @@ class _ConstantDeprecationRegistry:
     when registered constants are accessed.
     """
 
-    deprecations: dict[str, tuple[str, type[Warning], int, Any]] = field(
+    deprecations: dict[str, tuple[str, type[Warning], int, Any, bool]] = field(
         default_factory=dict,
         init=False,
         repr=False,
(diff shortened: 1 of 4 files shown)


dholth:
At first glance I like this pattern.

travishathaway (approved):
Seems like a great way to squeeze some startup time improvements out. The documentation additions look great and are easy to understand too.

kenodegard (commented):
One thing worth considering: since `lambda` can't do imports (without using `importlib`), users will need to define named factory helpers in their module scope, which pollutes the namespace. A simple workaround is to delete the helper after registration:

```python
def factory_import():
    import ...
    return ...

deprecated.constant(factory=factory_import)
del factory_import
```

We could also add an `import=` keyword as syntactic sugar for the simple cases (e.g., re-exports), but this isn't strictly necessary given the `del` pattern is possible.

jezdez:
Good call on the namespace question. In practice the four follow-ups don't actually hit that case cleanly:

- **A10** — `_representer()` is also used inside `_yaml()` (`parser.Representer = _representer()`), so it can't be `del`-ed after registration.
- **A24** — the module-level `_json()` helper is shared by 3 factory lambdas in `conda.auxlib.logz` and 2 in `conda.common.serialize`.
- **A22** — uses an in-loop `lambda n=name: getattr(_load_types(), n)`, and `_load_types` is also used by the module's own `__getattr__`.
- **A2/A3** — doesn't use `factory=` at all (registers lazily from inside `__getattr__` with a precomputed `val`).

So the helpers stay around in all cases, but they're underscore-prefixed module privates that get reused, which I think is fine.

I also looked at whether `import=` would help those sites: only ~4 of the ~20-odd `factory=` call sites across the track are pure `module:attr` shapes. A10 builds a `RoundTripRepresenter` subclass, A24's `_DUMPS`/`jsondumps` construct+bind a `CondaJSONEncoder(...).encode`, and A2/A3 is already in a `__getattr__`. So `import=` would be additive sugar over `factory=`, not a substitute.

Given `factory=` already covers every case and you flagged `import=` as not strictly necessary, I'd rather land this as-is and revisit `import=` as a non-breaking follow-up if we start seeing a lot of pure re-export call sites in practice.

PR #15928, opened 2026-04-17
Add multi-platform `available_platforms` and `env_for` to `EnvironmentSpecBase`

### LINKED ISSUE #15927: Extend `EnvironmentSpecBase` with multi-platform `available_platforms` and `env_for`
### Checklist

### What is the idea?

Extend `EnvironmentSpecBase` with two new default-implemented APIs for multi-platform introspection:

```python
class EnvironmentSpecBase(ABC):
    # existing
    @property
    @abstractmethod
    def env(self) -> Environment:
        ...

    # new — multi-platform
    @property
    def available_platforms(self) -> tuple[str, ...]:
        """Platforms this spec can produce an Environment for.

        Defaults to (context.subdir,). Multi-platform specs
        (conda-lock-v1, rattler-lock-v6) override to expose the full
        platform set declared in the input file.
        """
        return (context.subdir,)

    def env_for(self, platform: str) -> Environment:
        """Return the Environment for a specific platform.

        Defaults to self.env when platform == context.subdir, raises
        ValueError otherwise. Multi-platform specs override to build the Environment
        directly from the parsed input without constructing one per
        platform.
        """
        if platform not in self.available_platforms:
            raise ValueError(...)
        return self.env
```

Key design decisions:
- **Two orthogonal primitives**: `available_platforms` for discovery, `env_for` for lookup. No convenience iterator — callers just write `(spec.env_for(p) for p in spec.available_platforms)`.
- **Additive, not breaking** — default implementations mean every existing plugin keeps working with zero changes.
- **`env_for` avoids constructing unused Environments** — naïvely filtering an "all platforms" iterable forces every platform to be built. `env_for` lets callers build only what they need.

The plugin is still tagged `**EXPERIMENTAL**`, so additive API changes are low-friction.

### Why is this needed?

Today, `EnvironmentSpecBase.env` returns a single `Environment` for `context.subdir`. Multi-platform lockfiles (conda-lock.yml, pixi.lock) contain packages for multiple platforms, but the loader API only exposes one at a time.

This creates problems for consumers that need the full multi-platform view:

1. **API symmetry**: The exporter side already accepts `Iterable[Environment]` for multi-platform output. The loader side should provide targeted per-platform access for the inverse direction. Without this, the env-spec plugin API is incomplete for round-trip workflows.

2. **Actionable error messages**: _"Lockfile does not contain packages for linux-64. Available platforms: osx-arm64, win-64"_ — `env_for("linux-64")` is the natural place to raise this.

3. **Lockfile-to-lockfile transformations**: Tools that translate between lockfile formats without re-solving need per-platform access. Without a formal API, consumers must mutate `context.subdir` and reparse — fragile and incomplete.

Formalising this contract in conda core (rather than ad-hoc duck-typing on individual plugins) gives third-party consumers a stable, documented API.

### What should happen?

- Any `EnvironmentSpecBase` subclass can be queried via `.available_platforms` and `.env_for(platform)`.
- Subclasses not overriding them get sensible single-platform defaults.
- Multi-platform specifiers (like conda-lockfiles' `CondaLockV1Loader` and `RattlerLockV6Loader`) override both.
- Documentation in `environment_specifiers.rst` explains when/why to override.

### Additional Context

- Reference implementation: conda-incubator/conda-lockfiles#127 (`available_platforms` + `env_for` on both lockfile loaders)
- Exporter-side prior art: conda-incubator/conda-lockfiles#55 (multiplatform export)
- Parent epic: conda-incubator/conda-lockfiles#83 (Multi-platform Lockfiles)

jezdez:
### Description

`EnvironmentSpecBase.env` returns an `Environment` for `context.subdir`. This is fine for single-platform specs (`environment.yml`, `requirements.txt`, `explicit`, CEP-24) where the spec only covers one platform anyway, but multi-platform lockfiles (`conda-lock.yml`, `pixi.lock`) hold packages for several platforms at once and there is no way to target any but the current one without mutating `context.subdir` and reparsing the input.

This PR adds two default-implemented APIs on `EnvironmentSpecBase`:

```python
@property
def available_platforms(self) -> tuple[str, ...]:
    return (context.subdir,)

def env_for(self, platform: str) -> Environment:
    if platform not in self.available_platforms:
        raise ValueError(...)
    return self.env
```

`available_platforms` reports the platforms a spec can produce an `Environment` for, and `env_for(platform)` returns the `Environment` for one of them. The defaults cover the single-platform case so existing specifiers need no changes. Multi-platform specifiers override both and iterate with `(spec.env_for(p) for p in spec.available_platforms)`, so each `Environment` is only built when a caller asks for it.

`conda env create`, `conda env update` and `Environment.from_cli_with_file_envs` previously called `spec.env` directly. They now check `context.subdir in spec.available_platforms` first and call `spec.env_for(context.subdir)` instead. When the check fails they raise `PlatformMismatchError` (a new `CondaValueError` subclass) naming the platforms the file does cover and suggesting `--platform=<subdir>`. `Environment.from_cli` collects the check results for every `-f` / `--file` argument before building any `Environment` and reports them all in one error instead of stopping at the first mismatch. A dedicated exception keeps the wording consistent across all three call sites.

Note: using `--platform=<subdir>` to select a platform the spec does cover but that does not match the host is allowed today (it is how `CONDA_SUBDIR` stages prefixes for other machines) and this PR does not change that. Whether lockfile-based installs should reject cross-platform selections by default is tracked separately in #15936.

Reference implementation on the loader side: conda-incubator/conda-lockfiles#128. Aligning the exporter plugin API the same way is tracked in #15929.

Closes #15927.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_env_create.py
+++ b/conda/cli/main_env_create.py
@@ -104,7 +104,11 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..env.pip_util import get_pip_workdir
-    from ..exceptions import CondaEnvException, CondaValueError, InvalidInstaller
+    from ..exceptions import (
+        CondaEnvException,
+        InvalidInstaller,
+        PlatformMismatchError,
+    )
     from ..gateways.disk.delete import rm_rf
     from .common import validate_file_exists
 
@@ -118,10 +122,8 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     )
     spec = spec_hook.environment_spec(args.file)
     if context.subdir not in spec.available_platforms:
-        raise CondaValueError(
-            f"{args.file!r} does not include packages for {context.subdir}.\n"
-            f"Available platforms: {', '.join(spec.available_platforms)}\n"
-            f"Select one with --platform=<subdir>."
+        raise PlatformMismatchError(
+            [(args.file, spec.available_platforms)], context.subdir
         )
     env = spec.env_for(context.subdir)
 
--- a/conda/cli/main_env_create.py
+++ b/conda/cli/main_env_create.py
@@ -119,8 +119,9 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     spec = spec_hook.environment_spec(args.file)
     if context.subdir not in spec.available_platforms:
         raise CondaValueError(
-            f"{args.file!r} does not include packages for {context.subdir}. "
-            f"Available platforms: {', '.join(spec.available_platforms)}"
+            f"{args.file!r} does not include packages for {context.subdir}.\n"
+            f"Available platforms: {', '.join(spec.available_platforms)}\n"
+            f"Select one with --platform=<subdir>."
         )
     env = spec.env_for(context.subdir)
 
--- a/conda/cli/main_env_create.py
+++ b/conda/cli/main_env_create.py
@@ -104,7 +104,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..env.env import print_result
     from ..env.installers.base import get_installer
     from ..env.pip_util import get_pip_workdir
-    from ..exceptions import CondaEnvException, InvalidInstaller
+    from ..exceptions import CondaEnvException, CondaValueError, InvalidInstaller
     from ..gateways.disk.delete import rm_rf
     from .common import validate_file_exists
 
@@ -117,7 +117,12 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         name=context.environment_specifier,
     )
     spec = spec_hook.environment_spec(args.file)
-    env = spec.env
+    if context.subdir not in spec.available_platforms:
+        raise CondaValueError(
+            f"{args.file!r} does not include packages for {context.subdir}. "
+            f"Available platforms: {', '.join(spec.available_platforms)}"
+        )
+    env = spec.env_for(context.subdir)
 
     # FIXME conda code currently requires args to have a name or prefix
     # don't overwrite name if it's given. gh-254
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -15,7 +15,7 @@ from ..base.context import context, validate_channels
 from ..common.constants import NULL
 from ..common.iterators import groupby_to_dict as groupby
 from ..core.prefix_data import PrefixData
-from ..exceptions import CondaValueError
+from ..exceptions import CondaValueError, PlatformMismatchError
 from ..history import History
 from ..misc import get_package_records_from_explicit
 from .match_spec import MatchSpec
@@ -552,14 +552,7 @@ class Environment:
             if context.subdir not in spec.available_platforms:
                 incompatible.append((fpath, spec.available_platforms))
         if incompatible:
-            details = "\n".join(
-                f"  {fp!r}: {', '.join(plats)}" for fp, plats in incompatible
-            )
-            raise CondaValueError(
-                f"The following files do not include packages for {context.subdir}:\n"
-                f"{details}\n"
-                f"Select a supported platform with --platform=<subdir>."
-            )
+            raise PlatformMismatchError(incompatible, context.subdir)
         for fpath, spec in file_specs.items():
             env = spec.env_for(context.subdir)
             envs_from_file.append(env)
--- a/conda/models/environment.py
+++ b/conda/models/environment.py
@@ -546,8 +546,14 @@ class Environment:
                 name=context.environment_specifier,
             )
             spec = spec_hook.environment_spec(fpath)
-            envs_from_file.append(spec.env)
-            fpath_envs_map[fpath] = spec.env
+            if context.subdir not in spec.available_platforms:
+                raise CondaValueError(
+                    f"{fpath!r} does not include packages for {context.subdir}. "
+                    f"Available platforms: {', '.join(spec.available_platforms)}"
+                )
+            env = spec.env_for(context.subdir)
+            envs_from_file.append(env)
+            fpath_envs_map[fpath] = env
 
         # Add default packages if required. If the default package is already
         # present in the list of specs, don't add it (this will override any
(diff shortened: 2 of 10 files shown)


travishathaway (approved):
@jezdez,

I had the chance to look this over and it looks overall pretty solid. Seems like a good idea for a change to the API, and it's nice that we get better error messaging out of it.

conda/cli/main_env_create.py
```diff
@@ -117,7 +117,13 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         name=context.environment_specifier,
     )
     spec = spec_hook.environment_spec(args.file)
-    env = spec.env
+    if context.subdir not in spec.available_platforms:
+        raise CondaValueError(
+            f"{args.file!r} does not include packages for {context.subdir}.\n"
+            f"Available platforms: {', '.join(spec.available_platforms)}\n"
+            f"Select one with --platform=<subdir>."
```
soapy1, line 124:
```suggestion
            
```

In this situation, the user is tyring to use a lockfile that does not have a the information for the current `subdir`.  If they try to install the packages for a different platform, they'll probably end up with a broken environment. 

For example, consider my test environment:
```
$ conda create -p /tmp/simple numpy
```

I can export a multiplatform lockfile (for the purpose of this test I export to platforms not including my host platform (linux-64)
```
$ conda export --format conda-lock-v1   -p /tmp/simp --file cl-multi.yaml --platform osx-64 --platform win-64
```

When I try to install this environment on my linux-64 machine I get an error (as expected)
```
$ conda create --file cl-multi.yaml -p /tmp/test1  

CondaValueError: The following files do not include packages for linux-64:
  'cl-osx64.yaml': osx-64
Select a supported platform with --platform=<subdir>.
```

If I follow the advice in the error, I get an installed environment. But, it's got packages for osx-64 in it, so it doesn't work
```
$ conda create --file cl-multi.yaml -p /tmp/test1 --platform=osx-64
$ conda activate /tmp/test1                                 
(/tmp/test1) sophia:~/ $ python --version               
zsh: exec format error: python
```

conda/models/environment.py
```diff
@@ -540,14 +540,30 @@ def from_cli_with_file_envs(
...
-            envs_from_file.append(spec.env)
-            fpath_envs_map[fpath] = spec.env
+            file_specs[fpath] = spec
+            if context.subdir not in spec.available_platforms:
+                incompatible.append((fpath, spec.available_platforms))
+        if incompatible:
+            details = "\n".join(
+                f"  {fp!r}: {', '.join(plats)}" for fp, plats in incompatible
+            )
+            raise CondaValueError(
```
soapy1, line 558: Looks like this is the same error as the in `main_env_create.py` and `main_env_update.py`. Might be nice to create a new error class for this to keep the verbage consistent between all the ways a user can encounter this error.

conda/cli/main_env_create.py
soapy1: Should there also be a check here to make sure the user is installing files for their current host platform? 

When I try this same thing on the main branch of conda I get an error:
```
$ conda create --file cl-multi.yaml -p /tmp/test2 --platform osx-64 

CondaValueError: 

Unable to create environment: Platform linux-64 not found in lockfile. Available platforms: osx-64
```

soapy1 (changes requested):

conda/models/environment.py
jezdez: Done, added `PlatformMismatchError` (subclass of `CondaValueError`) in `conda.exceptions` and routed all three call sites through it so the message shape is shared. Covered with direct unit tests in `tests/test_exceptions.py` in addition to the existing `from_cli` test which parametrizes over single- and multi-file inputs.

conda/cli/main_env_create.py
jezdez: Thanks for the concrete reproducer. This is the staging behaviour conda supports today via `CONDA_SUBDIR` / `--platform`, a prefix built for a non-host platform is expected to be unusable on the host (it is meant to be packaged and shipped). The loader-level `Platform linux-64 not found in lockfile` error you see on `main` was a side effect of the loader always reading `context.subdir`, not a deliberate host-platform guard, and this PR intentionally makes the platform selection work instead of accidentally filtering it out.

Whether `conda env create` / `conda env update` / `conda create --file` should reject `context.subdir != host_subdir` by default (à la `conda-lock install --force-platform`) is a real question, but flipping that default changes existing conda behaviour beyond env specs, so I'd rather not bundle it into this PR. I filed #15936 to settle that separately, it references this discussion and covers the staging use case, the conda-lock precedent, and the opt-in shape.

jezdez: Covered in #15936 together with the other thread, same underlying question about default behaviour for cross-platform installs.

soapy1: Sounds good, thanks for creating a new issue to track that!

I also think this error message should be updated to not give the advice to select a different platform to install. If a user follows this direction, they will probably end up with another error. 

Another piece of advice to give users is to regenerate their lockfile with the correct platform.

Maybe something like:
```
PlatformMismatchError: 'cl-multi.yaml' does not include packages for linux-64.
Available platforms: osx-64, win-64

In order to install the environment on a linux-64 system, please regenerate the `cl-multi.yaml` with linux-64 as a configured platform. Try:

  $ conda export --file cl-multi.yaml --platform osx-64 --platform win-64 --platform linux-64
```

soapy1 (approved):
Nice one! Just a suggestion about the `PlatformMismatchError` error message.

conda/cli/main_env_create.py
jezdez: done in 824c1f84. swapped the `--platform=<subdir>` hint for a `conda export` line that includes the current platform, same shape as your example. the multi-file branch uses the same pattern so the advice stays consistent across `conda env create`, `conda env update`, and `Environment.from_cli`.

single-file case now renders as:

```
Environment file 'cl-multi.yaml' does not include packages for linux-64.
Available platforms: osx-64, win-64

To install on linux-64, regenerate the environment file with linux-64 as a configured platform, for example:

    conda export --file cl-multi.yaml --platform osx-64 --platform win-64 --platform linux-64
```

multi-file case:

```
The following environment files do not include packages for linux-64:
  'a.yml': osx-64
  'b.yml': win-64, linux-aarch64

Regenerate each file with linux-64 as a configured platform (e.g. via `conda export --file <path> --platform linux-64 ...`).
```

PR #15934, opened 2026-04-19
Use record subdir instead of native platform when mutating channel's `canonical_name` + compare channels by `canonical_name`

### LINKED ISSUE #15933: `dist_str()` includes wrong or doubled platform subdir when channel URL has no explicit subdir
### Checklist

### What happened?

When a package is installed from a channel URL that has no subdir, say, `http://localhost:8080/my-channel` instead of `http://localhost:8080/my-channel/linux-64`), the index-building logic fakes the channel's `canonical_name` using `channel.url()`. That method falls back to `context.subdirs`, which is the **native** platform, when no explicit subdir is set on the channel. The resulting `canonical_name` already contains a platform subdir, but `dist_str()` unconditionally appends `self.subdir` again, which produces either a doubled subdir (on Linux), or a mismatched one (on macOS when inspecting a `linux-64` prefix).

I found this as part of https://github.com/conda-incubator/pytest-conda-solvers/pull/19.

I'm in the process of putting together a PR for this.

### Conda Info

```shell
active environment : base
    active env location : /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge
            shell level : 1
       user config file : /Users/agriyakhetarpal/.condarc
 populated config files : /Users/agriyakhetarpal/.condarc
          conda version : 26.3.3.dev40+g388d94d4f.d20260419
         python version : 3.10.20.final.0
                 solver : libmamba (default)
                          __conda=26.3.3.dev40+g388d94d4f.d20260419=0
                          __osx=26.4=0
                          __unix=0=0
       base environment : /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge  (writable)
      conda av data dir : /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-arm64
          package cache : /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge/pkgs
                          /Users/agriyakhetarpal/.conda/pkgs
       envs directories : /Users/agriyakhetarpal/Desktop/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniforge/envs
                          /Users/agriyakhetarpal/.conda/envs
    temporary directory : /var/folders/b3/2bq1m1_50bs4c7305j8vxcqr0000gn/T
               platform : osx-arm64
             user-agent : conda/26.3.3.dev40+g388d94d4f.d20260419 requests/2.33.1 CPython/3.10.20 Darwin/25.4.0 OSX/26.4 solver/libmamba conda-libmamba-solver/26.3.0 libmambapy/2.5.0
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

AI disclosure: I noticed this as part of my debugging and asked GPT-5 mini (0x token usage promotion via VS Code Chat) to dive into the code and confirm these findings in the code.

agriyakhetarpal:
### Description

Closes #15933 and xref https://github.com/conda-incubator/pytest-conda-solvers/pull/19.

With this PR, I have attempted to fix two things:
- When merging prefix/repodata entries, compares channels by their `canonical_name` instead of direct `Channel` equality to avoid false mismatches. `Channel.__eq__` includes `platform` in its comparison, so a prefix record whose channel was created from a bare URL (`platform=None`) does not match the corresponding index record whose channel was created from a subdir URL (`platform="linux-64"`), triggering the fake-channel path spuriously. Comparing by `canonical_name` accurately identifies whether two records belong to the same channel, regardless of how the channel object was constructed. `pytest-conda-solvers` happens to trigger it consistently because `package_record_from_dist_str` reconstructs channels from bare URLs parsed out of dist-strings.
- (See explanation/walkthrough below): ensures `prefix_channel.__canonical_name` is set from `base_url` so `dist_str()` appends the `subdir` only once (which fixes doubled/wrong-platform subdirs as I noted in the issue).

For a bare URL like `http://localhost:8080/my-channel`, which has no `subdir`, the `canonical_name` computes:

https://github.com/conda/conda/blob/c706f6d7d8633223c80f89cbcba4731692768ce5/conda/models/channel.py#L248-L251

`http://localhost:8080/my-channel`, i.e., it gives us just the `scheme://location/name`, with no `platform` appended.

However, in `_supplement_index_dict_with_prefix` and `_update_from_prefix` here:

https://github.com/conda/conda/blob/f24fba4aca4fd5bac6bb26b7e5b2d7bdafc52072/conda/core/index.py#L307-L308

https://github.com/conda/conda/blob/f24fba4aca4fd5bac6bb26b7e5b2d7bdafc52072/conda/core/index.py#L381-L382

We can follow on and see that `channel.url()` checks `self.platform`:

https://github.com/conda/conda/blob/f24fba4aca4fd5bac6bb26b7e5b2d7bdafc52072/conda/models/channel.py#L319-L342

So if a channel is created from a bare URL, `channel.url()` notices that `platform` is `None`, and we fall into the `else` branch. This appends the **native** platform (e.g., `osx-arm64` on macOS, `linux-64` on Linux), returning `http://localhost:8080/my-channel/linux-64` and so on into the `canonical_name`.

Then `dist_str()` independently appends `PackageRecord.subdir`:

https://github.com/conda/conda/blob/f24fba4aca4fd5bac6bb26b7e5b2d7bdafc52072/conda/models/records.py#L385-L392

With a mutated `canonical_name` of `http://localhost:8080/my-channel/linux-64` and `subdir="linux-64"`, we get concatenated platform strings:

Linux: `http://localhost:8080/my-channel/linux-64/linux-64::foo-1.0-0` (doubles the subdir)
macOS: `http://localhost:8080/my-channel/osx-arm64/linux-64::foo-1.0-0` (adds the wrong platform)

So, we can use `base_url` and leave `dist_str()` as the sole place that appends the subdir.

### Checklist - did you ...

- [ ] [TODO] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     

jezdez:
Ran this through an instrumented `Index` on the base commit. The fake branch at L307 is only reachable in one scenario: the one this PR is fixing. The original 2019 trigger (`pkgs/main` vs `pkgs/free`, closing #8302) isn't reachable anymore because `context.custom_multichannels["defaults"]` only contains `pkgs/main` and `pkgs/r` now -- `pkgs/free` resolves to its URL-form `canonical_name`, pkeys diverge on the first element, and we go to the "not in repodata" branch before ever reaching the fake.

What I ran, summarized:

| scenario | in `_data` | `__eq__` | `canonical_name` match | fake fires |
| --- | :-: | :-: | :-: | :-: |
| `pkgs/main` vs `pkgs/free` | no | no | no | no |
| bare URL vs subdir URL, same channel | **yes** | no | **yes** | **yes** |
| different locations, same name | no | no | no | no |
| same channel, different subdirs | no | no | yes | no |

Then I applied Change 1 alone and reran. Zero fake-branch firings. Reason: `_pkey`'s first element is `self.channel.canonical_name` (records.py:362), so `prefix_record in self._data` implies canonical_names already match. With Change 1's condition being identical to the first component of the outer membership check, the `else` is structurally unreachable.

So Change 2 (`url()` -> `base_url`) is touching dead code under the logic Change 1 introduces. Still a strict improvement as insurance, but worth calling out: the actual fix for the pytest-conda-solvers case (and for anyone constructing `PackageRecord`s from bare URLs) is Change 1.

Probably worth a follow-up to just delete the fake branch outright -- ~8 lines across both call sites, trigger scenario no longer exists, Change 1 makes it unreachable. Can file that separately if you're keeping this PR scoped.

Provenance for context: fake-channel pattern introduced in 8b9dd1bf80 (July 2019, 4.7.8), duplicated into `_update_from_prefix` by #13880 (24.9.0). ~7 years in the tree.

conda/core/index.py
```diff
@@ -373,12 +381,12 @@ def _update_from_prefix(
...
             if prec:
-                if prec.channel == prefix_prec.channel:
+                if prec.channel.canonical_name == prefix_prec.channel.canonical_name:
                     link = prefix_prec.get("link") or EMPTY_LINK
                     prec = PrefixRecord.from_objects(prec, prefix_prec, link=link)
                 else:
                     prefix_channel = prefix_prec.channel
-                    prefix_channel._Channel__canonical_name = prefix_channel.url()
+                    prefix_channel._Channel__canonical_name = prefix_channel.base_url
                     del prefix_prec._PackageRecord__pkey
```
jezdez, line 390: Same explanation applies here.

```suggestion
                    prefix_channel = prefix_prec.channel
                    # channel.url() appends the native platform subdir when
                    # channel.platform is None (bare URL), which corrupts
                    # canonical_name and causes dist_str() to produce a doubled
                    # or wrong-platform subdir. base_url never includes a subdir,
                    # so dist_str() appends it exactly once as intended.
                    prefix_channel._Channel__canonical_name = prefix_channel.base_url
                    del prefix_prec._PackageRecord__pkey
```

jezdez:
Question for maintainers before this merges: should we just delete the fake branch?

The #8302 scenario it was added for (`pkgs/main` <-> `pkgs/free`) doesn't reproduce anymore -- `pkgs/free` isn't in `custom_multichannels["defaults"]`, so pkeys diverge and the outer `in self._data` check fails before we ever get there. And with Change 1 applied, the `else` is unreachable anyway since pkey's first element is literally `canonical_name`.

Happy to leave it as-is with the `base_url` fix if you'd rather not scope-creep this PR, or if there's a case I missed where it's still needed. Otherwise it'd be nice to drop those ~8 lines (and the `_Channel__canonical_name` mangling) in a follow-up.

jaimergp:
I didn't have time to check the full diff yet, but:

> When merging prefix/repodata entries, compares channels by their canonical_name instead of direct Channel equality to avoid false mismatches.

This reminded me of https://github.com/conda/conda/pull/13418. Worth taking a look there in case there are overlaps.

jezdez:
> I didn't have time to check the full diff yet, but:
> 
> > When merging prefix/repodata entries, compares channels by their canonical_name instead of direct Channel equality to avoid false mismatches.
> 
> This reminded me of #13418. Worth taking a look there in case there are overlaps.

Ooooh, you're right!

conda/core/index.py
agriyakhetarpal: Resolving as this can't be applied anymore!

agriyakhetarpal:
Hi! Coming back to this PR, and picking it up!

To reply to

> So Change 2 (url() -> base_url) is touching dead code under the logic Change 1 introduces. Still a strict improvement as insurance, but worth calling out: the actual fix for the pytest-conda-solvers case (and for anyone constructing PackageRecords from bare URLs) is Change 1.
>
> Probably worth a follow-up to just delete the fake branch outright -- ~8 lines across both call sites, trigger scenario no longer exists, Change 1 makes it unreachable. Can file that separately if you're keeping this PR scoped.

and to

> Question for maintainers before this merges: should we just delete the fake branch?
> 
> The #8302 scenario it was added for (`pkgs/main` <-> `pkgs/free`) doesn't reproduce anymore -- `pkgs/free` isn't in `custom_multichannels["defaults"]`, so pkeys diverge and the outer `in self._data` check fails before we ever get there. And with Change 1 applied, the `else` is unreachable anyway since pkey's first element is literally `canonical_name`.
> 
> Happy to leave it as-is with the `base_url` fix if you'd rather not scope-creep this PR, or if there's a case I missed where it's still needed. Otherwise it'd be nice to drop those ~8 lines (and the `_Channel__canonical_name` mangling) in a follow-up.

Thanks for the investigation, @jezdez! I chose to drop the fake branch here because it's dead code anyway at both call sites, to avoid the churn of another PR (but I can split it into two PRs if suggested, as I don't feel strongly about this). PTAL at 0ba659d9d9901fe6410c798762d94c0c0a2e3161, thanks!


> I didn't have time to check the full diff yet, but:
> 
> > When merging prefix/repodata entries, compares channels by their canonical_name instead of direct Channel equality to avoid false mismatches.
> 
> This reminded me of #13418. Worth taking a look there in case there are overlaps.

@jaimergp, I took a look. #13418 touches `match_spec.py` (`_parse_channel` and `ChannelMatch.match`/`__str__` stuff), so it's a different code path from this one, because we handle the `core/index.py` prefix and repodata merges here. I'd say the broader theme is the same of course, as we're talking about comparing channels in the right manner, but the diffs don't overlap, and neither should block the other. I can revisit it separately from this PR!

ryanskeith (commented):
This seems very reasonable.  I would suggest adding one additional test to target the _update_from_prefix changes.  As the merge logic is the same, it would be nice to ad da test with the same setup as `test_supplement_index_merges_bare_and_subdir_url_channels` but of course using the Index.__get_item__ call.

Not a show stopper to address the bug but a nice upkeep step.

ryanskeith (approved):
Thank you for adding the new tests.

jezdez (approved):
Looks good to me. The fake-channel branch feedback has been addressed by deleting it at both call sites, and the added regression coverage now exercises both `_supplement_index_dict_with_prefix` and the `Index.__getitem__` / `_update_from_prefix` path. I also rechecked the #13418 overlap note; that PR is confined to `conda/models/match_spec.py`, so there is no direct diff overlap here. Focused local tests passed for the two new regression tests, and the current head CI is green.

PR #15942, opened 2026-04-21
Drop ChannelType metaclass from conda.models.channel (A19a)

jezdez:
### Description

Replace the `ChannelType` metaclass on `conda.models.channel.Channel` with `Channel.__new__` plus `@functools.cache` on `Channel.from_value`. Split out of #15916 per @kenodegard's review comment so the records and channel refactors can be reviewed and merged independently.

The `ChannelType` metaclass only existed to do two things at construction time:

1. Intercept single-arg `Channel(value)` calls for caching and identity.
2. Dispatch on the `channels=` kwarg to `MultiChannel`.

Both behaviours now live on the class itself:

- `Channel.__new__` handles the single-arg identity/cache fast path and the `channels=` kwarg dispatch to `MultiChannel`, returning a regular instance otherwise.
- `Channel.__init__` short-circuits on re-entry (when `__new__` returned a cached instance) via an empty-`__dict__` check, so Python's normal `__init__` pass after a cached `__new__` does not re-assign attributes on the cached object.
- `Channel.from_value` is wrapped with `@functools.cache` for the per-value caching the old `Channel._cache_` dict provided; `_reset_state()` now calls `cache_clear()` and stays wired to `context.register_reset_callaback`.
- `MultiChannel` mirrors the `__init__` re-entry guard and accepts a `**_extra` sink so `dump()`-based round-trips (e.g. `Channel(**{**channel.dump(), "platform": platform})`) keep working.

Refs: #15867

#### Deprecations

Adds a `deprecated.constant` shim so `from conda.models.channel import ChannelType` keeps working for one release cycle (deprecate 26.9, remove 27.3). The shim points at `type(Channel)`, so `isinstance(x, ChannelType)` stays semantically equivalent. Anyone doing `class Foo(metaclass=ChannelType)` gets a `PendingDeprecationWarning` and can migrate to a plain class.

#### Caveats reviewers should look at

Pickle and `copy.deepcopy` round-trip correctly, checked locally and covered by the existing `tests/models/test_channel.py` suite (`test_channel_cache`, `test_ppc64le_vs_ppc64`, the multichannel tests).

Subclassing still works for the common case of overriding attributes via kwargs. @kenodegard flagged one ordering pattern on the original PR:

```python
class SubChannel(Channel):
        self.value = ...
```

`super().__init__` sees a non-empty `__dict__` and short-circuits, so attributes set before it would be preserved but attributes assigned by `Channel.__init__` would not. `MultiChannel` does not call `super().__init__()`, so it is unaffected. Downstream subclasses that rely on `Channel.__init__` setting their attributes should call `super().__init__()` first.

Single-arg subclass construction (`MyChannel("url")`) still returns a plain `Channel`, same as on `main`; the old `ChannelType.__call__` already hardcoded `Channel.from_value` in this path. Kwargs-based subclass construction is unchanged.

`@functools.cache` on `from_value` replaces the old module-level `Channel._cache_` dict. Same invalidation semantics via `context.register_reset_callaback(Channel._reset_state)`.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes? (`news/15867-channel-metaclass`)
- [x] Add / update necessary tests? (existing `tests/models/test_channel.py` cache tests updated to use `Channel.from_value.cache_info()` / `cache_clear()`; the `ChannelType` deprecation shim verified to emit `PendingDeprecationWarning` and resolve to `type(Channel)`)
- [ ] Add / update outdated documentation? (no user-facing docs affected; `Channel` keeps the same public API)

### DIFF
--- a/conda/models/channel.py
+++ b/conda/models/channel.py
@@ -12,6 +12,7 @@ Object inheritance:
 from __future__ import annotations
 
 from copy import copy
+from functools import cache
 from logging import getLogger
 from typing import TYPE_CHECKING
 
@@ -36,6 +37,7 @@ from ..common.url import (
     split_scheme_auth_token,
     urlparse,
 )
+from ..deprecations import deprecated
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
@@ -48,31 +50,7 @@ if TYPE_CHECKING:
 log = getLogger(__name__)
 
 
-class ChannelType(type):
-    """
-    This metaclass does basic caching and enables static constructor method usage with a
-    single arg.
-    """
-
-    def __call__(cls, *args, **kwargs):
-        if len(args) == 1 and not kwargs:
-            value = args[0]
-            if isinstance(value, Channel):
-                return value
-            elif value in Channel._cache_:
-                return Channel._cache_[value]
-            else:
-                c = Channel._cache_[value] = Channel.from_value(value)
-                return c
-        elif "channels" in kwargs:
-            # presence of 'channels' kwarg indicates MultiChannel
-            channels = tuple(cls(**_kwargs) for _kwargs in kwargs["channels"])
-            return MultiChannel(kwargs["name"], channels)
-        else:
-            return super().__call__(*args, **kwargs)
-
-
-class Channel(metaclass=ChannelType):
+class Channel:
     """
     Channel:
     scheme <> auth <> location <> token <> channel <> subchannel <> platform <> package_filename
@@ -82,11 +60,20 @@ class Channel(metaclass=ChannelType):
 
     """
 
-    _cache_ = {}
-
     @staticmethod
     def _reset_state() -> None:
-        Channel._cache_ = {}
+        Channel.from_value.cache_clear()
+
+    def __new__(cls, *args, **kwargs):
+        if len(args) == 1 and not kwargs:
+            value = args[0]
+            if isinstance(value, Channel):
+                return value
+            return cls.from_value(value)
+        if "channels" in kwargs:
+            channels = tuple(cls(**_kw) for _kw in kwargs["channels"])
+            return MultiChannel(kwargs["name"], channels)
+        return super().__new__(cls)
 
     def __init__(
         self,
@@ -98,6 +85,8 @@ class Channel(metaclass=ChannelType):
         platform: str | None = None,
         package_filename: str | None = None,
     ):
+        if self.__dict__:
+            return
         self.scheme = scheme
         self.auth = auth
         self.location = location
@@ -127,6 +116,7 @@ class Channel(metaclass=ChannelType):
         return _get_channel_for_name(channel_name)
 
     @staticmethod
+    @cache
     def from_value(value: str | None) -> Channel:
         """Construct a new :class:`Channel` from a single value.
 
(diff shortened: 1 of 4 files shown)


conda/models/channel.py
```diff
@@ -97,7 +84,10 @@ def __init__(
         name: str | None = None,
         platform: str | None = None,
         package_filename: str | None = None,
+        **_extra: Any,
```
kenodegard: I don't think we need this?


```suggestion
```

conda/models/channel.py
```diff
@@ -424,20 +415,25 @@ def dump(self) -> dict[str, Any]:
 class MultiChannel(Channel):
     def __init__(
         self,
-        name: str,
-        channels: Iterable[Channel],
+        name: str | None = None,
+        channels: Iterable[Channel] | None = None,
         platform: str | None = None,
+        **_extra: Any,
```
kenodegard:
```suggestion
```

conda/models/channel.py
```diff
@@ -424,20 +415,25 @@ def dump(self) -> dict[str, Any]:
 class MultiChannel(Channel):
     def __init__(
         self,
-        name: str,
-        channels: Iterable[Channel],
+        name: str | None = None,
+        channels: Iterable[Channel] | None = None,
```
kenodegard: Is this because of typing?

tests/models/test_channel.py
```diff
@@ -1287,7 +1287,7 @@ def test_multichannel_priority():
 
 
 def test_ppc64le_vs_ppc64():
-    Channel._cache_.clear()
+    Channel.from_value.cache_clear()
```
kenodegard:
```suggestion
    Channel._reset_state()
```

tests/models/test_channel.py
```diff
@@ -1302,8 +1302,8 @@ def test_ppc64le_vs_ppc64():
         ppc64le_channel.url(with_credentials=True)
         == "https://conda.anaconda.org/dummy-channel/linux-ppc64le"
     )
-    print(Channel._cache_)
-    Channel._cache_.clear()
+    print(Channel.from_value.cache_info())
+    Channel.from_value.cache_clear()
```
kenodegard:
```suggestion
    Channel._reset_state()
```

jezdez:
Thanks @kenodegard, applied four of the five as-is in `fd475e27f`:

1. `**_extra: Any` on `Channel.__init__`: gone. Audited the call sites (`Channel(**parts)` in `core/subdir_data.py` and `Channel(**{**channel.dump(), "platform": platform})` inside `MultiChannel.__init__`) and both produce only keys that match the declared signature, so the sink was unnecessary defensive noise.
2. `**_extra: Any` on `MultiChannel.__init__`: gone, same reasoning.
3. `Channel.from_value.cache_clear()` in `test_ppc64le_vs_ppc64`: both call sites now use `Channel._reset_state()` as suggested.

On the typing question for `channels: Iterable[Channel] | None = None`: it's not really about typing, it's a signature-widening for a Python post-`__new__` quirk. When `Channel("defaults")` dispatches through `cls.from_value(value)` and `from_value` returns a fully-constructed `MultiChannel`, CPython then still invokes `type(returned).__init__(returned, "defaults")` with the original args (that's the rule whenever `__new__` returns an instance of `cls` or a subclass). So `MultiChannel.__init__` gets called with only `name="defaults"` bound; the re-entry guard `if self.__dict__: return` short-circuits before anything looks at `channels`, but the signature has to accept the call or we get `TypeError: MultiChannel.__init__() missing 1 required positional argument: 'channels'`. `test_channel_cache` hits this on every `Channel("defaults")` / `Channel("conda-canary")`.

I kept `name` required and only left `channels` optional (+ an `if channels is None: self._channels = ()` branch to match), and added an inline comment explaining the re-entry dance so it doesn't look like accidental over-optionality.

kenodegard (approved):

PR #15958, opened 2026-04-24
Make activate/deactivate return code 0 

### LINKED ISSUE #15826: [QA] Conda Activate –help gives error message on Windows
### Checklist

### What happened?

Description
When users run conda activate --help, the help content is displayed successfully, but the command ends with an error message: ERROR: 'conda activate --help' exited with code 1. This is misleading because the command appears to execute normally before reporting a failure.

## Steps to Reproduce

- Open a terminal.
- Run: conda activate --help
- Observe the command output and exit behavior.

## Actual Result
The help output is shown, but the command ends with:
ERROR: 'conda activate --help' exited with code 1.

## Expected Result
The command should display the help output successfully and exit without showing any error message.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

soapy1:
### Description

With this change, when a user executes a command to get the help output from activate/deactivate, conda will raise an error and have a 0 return code. Previsouly conda would raise and error and produce a 1 return code.

For example:

on linux
```
$ conda activate -h  

ActivateHelp: usage: conda activate [-h] [--[no-]stack] [env_name_or_prefix]

Activate a conda environment.

Options:

positional arguments:
  env_name_or_prefix    The environment name or prefix to activate. If the
                        prefix is a relative path, it must start with './'
                        (or '.\' on Windows).

optional arguments:
  -h, --help            Show this help message and exit.
  --stack               Stack the environment being activated on top of the
                        previous active environment, rather replacing the
                        current active environment with a new one. Currently,
                        only the PATH environment variable is stacked. This
                        may be enabled implicitly by the 'auto_stack'
                        configuration variable.
  --no-stack            Do not stack the environment. Overrides 'auto_stack'
                        setting.


$ echo $?           
0
```

on windows (cmd)
```
> conda activate -h

ActivateHelp: usage: conda activate [-h] [--[no-]stack] [env_name_or_prefix]

Activate a conda environment.

Options:

positional arguments:
  env_name_or_prefix    The environment name or prefix to activate. If the
                        prefix is a relative path, it must start with './'
                        (or '.\' on Windows).

optional arguments:
  -h, --help            Show this help message and exit.
  --stack               Stack the environment being activated on top of the
                        previous active environment, rather replacing the
                        current active environment with a new one. Currently,
                        only the PATH environment variable is stacked. This
                        may be enabled implicitly by the 'auto_stack'
                        configuration variable.
  --no-stack            Do not stack the environment. Overrides 'auto_stack'
                        setting.

> echo %errorlevel%
0
```

fixes: https://github.com/conda/conda/issues/15826

### Checklist - did you ...

### DIFF
--- a/conda/exceptions.py
+++ b/conda/exceptions.py
@@ -87,7 +87,7 @@ class ArgumentError(CondaError):
 
 
 class Help(CondaError):
-    pass
+    return_code = 0
 
 
 class ActivateHelp(Help):
--- a/conda/shell/condabin/_conda_activate.bat
+++ b/conda/shell/condabin/_conda_activate.bat
@@ -36,6 +36,16 @@ IF NOT EXIST "%__conda_tmp%" (
     ENDLOCAL & EXIT /B 2
 )
 
+:: Check if -h or --help is passed. If so, no need to do activation steps, exit early
+FOR %%A IN (%*) DO (
+    IF "%%A"=="-h" (
+        ENDLOCAL & EXIT /B 0
+    )
+    IF "%%A"=="--help" (
+        ENDLOCAL & EXIT /B 0
+    )
+)
+
 :: Check if conda produced output
 FOR /F "delims=" %%T IN (%__conda_tmp%) DO (
     :: T = "%TEMP%\<uuid>.env"
--- a/news/15958-make-activate-help-command-return-0
+++ b/news/15958-make-activate-help-command-return-0
@@ -1,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Update the `conda activate`/`conda deactivate` commands return a 0 exit code when run with the `-h`/`--help` flag. (#15826 via #15958)
+* Make the windows (bat activator) recognize the `-h`/`--help` flags. (#15826 via #15958)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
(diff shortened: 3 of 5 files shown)


conda/exceptions.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, message: str, **kwargs):
 
 
 class Help(CondaError):
-    pass
+    return_code = 0
```
soapy1, line 90: Which this change I still get an error on windows:

```
Error: Activation file missing for `conda activate --help`
```

soapy1: This was happening because this error message is originating from the _conda_activate.bat functions.

dholth (approved):
Nice

conda/exceptions.py
soapy1, line 90: This is a bit of a scary change imo. However, this change is only affecting error messages that are related to the help text shown during activation.

jezdez: Why is it scary?

soapy1: I think it's scary to be changing return codes that have been around for a long time. That being said, I feel good about this change because it should only be affecting the output of bulitin commands (`activate`, `deactivate`, `hook`, `commands`, `reactivate`) run with `-h`/`--help`.

jezdez (approved):

PR #15962, opened 2026-04-24
Rewrite lockfile-aware help text for `create`, `export`, `env create`, `env export`

### LINKED ISSUE #15960: Improve help text for lockfile workflows
**Problem**
With the conda-lockfiles plugin installed, the help text across export and create commands doesn't make it easy for users to discover or understand lockfile support.

Changes Needed
`conda export`
- remove format examples at the top that mix lockfile and non-lockfile formats in a single long line 
- `--format`, clearly separate lockfile formats from environment spec formats with short descriptions of what each group is good for. Put aliases (yaml/yml, reqs/txt) in parentheses next to canonical names instead of as separate entries
- `--file`, explain that standard lockfile filenames (conda-lock.yml, pixi.lock) are auto-detected — that's the easiest path. Custom filenames need --format
-  `--platform`: explain that it can be used to export envs for different platforms or multiplatform lockfile 
- Examples should show the simplest successful path first:

Examples ideas:
```
  Export environment spec:
    conda export --from-history > environment.yml
  
 Export a lockfile for the same platform:
    conda export --file pixi.lock
    conda export --file conda-lock.yml

  Export a lockfile for multiple platforms:
    conda export --file conda-lock.yml --platform linux-64 --platform osx-arm64

```

`conda create`

- Add `--file` flag with the same lockfile vs environment spec grouping as export, with short descriptions (solve vs no solve)
- Document `--format` flag as the override when auto-detection doesn't work
- Add examples (the simplest successful path first): create from lockfile, create from environment.yml, create from package specs
- Keep terminology consistent with export — same format names, same grouping

cc: @danyeaw

jezdez:
### Description

Three commands (`conda create`, `conda export`, `conda env create`) accept environment files as input or output. Lockfile plugins like [conda-lockfiles][cl] register through the same hooks as the built-in YAML / requirements / explicit exporters, but the help text never caught up: `--format` showed one flat list, `--file` didn't mention lockfiles, and the examples only covered environment specs.

Closes #15960.

### What this PR does

Each parser now builds its description, examples, and "Available formats" section from whatever specifier and exporter plugins are registered.

Examples use hardcoded built-in filenames (`environment.yml`, `explicit.txt`) for conda's own formats. The format listing and multi-platform lockfile example are fully dynamic — they reflect whatever plugins are installed.

`conda export --help` with only built-in plugins:

```text
Examples:
  Export an environment spec:
    conda export --from-history > environment.yml

  Export a lockfile for the same platform:
    conda export --file explicit.txt

Available formats:
  Environment specs:
    - environment-json (aliases: json) : environment.json
    - environment-yaml (aliases: yaml, yml, env.yml) : environment.yaml, environment.yml
    - requirements (aliases: reqs, txt) : requirements.txt, spec.txt

  Lockfiles:
    - explicit : explicit.txt
```

The format listing shows aliases (what to pass to `--format`) and default filenames (what triggers auto-detection via `--file`). The multi-platform example only renders when a registered lockfile exporter declares `multiplatform_export` (built-in `explicit` is single-platform). With [conda-lockfiles 0.2.0][cl-130] installed, the multi-platform line renders with that plugin's filename, and the registered lockfile formats appear under `Lockfiles:`.

`conda create --help` and `conda env create --help` get the same treatment with per-command wording.

### How it's wired up

| Piece | Where |
| --- | --- |
| `EnvironmentFormat.label` | `conda/plugins/types.py` — section header for the rendered listing. |
| `get_environment_exporters_grouped()` / `get_environment_specifiers_grouped()` | `conda/plugins/manager.py` — `EnvironmentFormat` → plugin lists (grouped view; distinct from `get_exporter_format_mapping()` for `--format` name/alias lookup). |
| `get_available_environment_formats(...)` | `conda/cli/formats.py` — renders the grouped bullets (aliases + default filenames). |
| `get_multiplatform_lockfile(...)` | `conda/cli/formats.py` — picks `default_filenames[0]` from the first lockfile exporter with `multiplatform_export` and filenames; used only for the optional multi-platform example. |

Each CLI `epilog()` pulls the grouped mapping from `context.plugin_manager`, composes hard-coded example lines, optionally appends the multi-platform block (export only), then appends the formatted listing when any plugins are present:

```python
formats = context.plugin_manager.get_environment_exporters_grouped()
multiplatform_lockfile = get_multiplatform_lockfile(formats)
# Built-in examples are hardcoded strings (environment.yml, explicit.txt)
# Multi-platform example only when multiplatform_lockfile is not None
if formats:
    examples.append("Available output formats:")
    examples.append(get_available_environment_formats(formats, indent=2))
```

### Tests

- `tests/cli/test_main_export.py`, `tests/cli/test_main_create.py`, `tests/cli/test_main_env_create.py` — parametrized `epilog()` expectations (including multi-platform gating) and `--help` smoke checks.
- `tests/cli/test_formats.py` — `get_multiplatform_lockfile` edge cases.

### Out of scope

- Per-format descriptions in help text (`description` is populated, just not rendered) — follow-up.
- Deprecated `--environment-specifier` / `--env-spec` still renders argparse's flat choices list. The non-deprecated `--format` uses `metavar="FORMAT"` to suppress it. Intentional nudge.

### Checklist - did you ...

- [x] Add a file to the `news` directory? (`news/15960-lockfile-help-text`)
- [x] ~Add / update outdated documentation?~ (CLI reference is generated from argparse.)

[cl]: https://github.com/conda-incubator/conda-lockfiles
[cl-130]: https://github.com/conda-incubator/conda-lockfiles/pull/130


jezdez:
@dashagurova Can you take a look at this by any chance?

dashagurova:
1. Do we have to include --env-spec for deprecation until v27.x? It was there for a second, I'd assume no one will miss it 

2. not sure if I was supposed to try it locally or if you are still working on it :) but I tried it locally 
conda export/conda create --help **conda-lockfiles plugin installed:**

```
Examples:

  Export an environment spec:
    conda export --from-history > conda-lock.yml ------> this is supposed to be environment.yml

  Export a lockfile for the same platform:
    conda export --file explicit.txt

  Export a lockfile for multiple platforms:
    conda export --file explicit.txt --platform linux-64 --platform osx-arm64 -------> is multi-platform supported in explicit.txt???

Available formats:

Environment specs:
  - conda-lock-v1 --------> should be in Lockfiles section
  - environment-json (json)
  - environment-yaml (yaml, yml, env.yml)
  - rattler-lock-v6 (pixi-lock-v6)  -------> should be in Lockfiles section
  - requirements (reqs, txt)

Lockfiles:
  - explicit
```
conda export/conda create --help **conda-lockfiles plugin NOT installed:**
```
Examples:

  Export an environment spec:
    conda export --from-history > environment.json --------> environment.YML?

  Export a lockfile for the same platform:
    conda export --file explicit.txt

  Export a lockfile for multiple platforms:
    conda export --file explicit.txt --platform linux-64 --platform osx-arm64 ---------> same here, explicit is not multiplatform 

Available formats:

Environment specs:
  - environment-json (json)
  - environment-yaml (yaml, yml, env.yml)
  - requirements (reqs, txt) ---------> I feel like users will assume this is pip-style reqs.txt.  

Lockfiles: ----------> this section still appears even when the plugin is not installed 
  - explicit
```

jezdez:
thanks for the close read, @dashagurova! quick pass through your notes:

**1. `--env-spec` deprecation** — yeah, the flag shipped, so per [CEP 9](https://github.com/conda-incubator/ceps/blob/main/cep-0009.md) it has to age out (currently scheduled for 27.x). dropping it sooner would need a deprecation-policy change, not a PR change.

**2. spec example showing `conda-lock.yml` / `environment.json`** — real bug, fixed in e213f9a. root cause turned out to be `CondaPluginManager.get_hook_results` sorting alphabetically by name — `example_filename_for` was just picking the alphabetically-first plugin. now pinned via a new `prefer_filenames=("environment.yml", "environment.yaml")` arg, so the spec example always renders `environment.yml` regardless of which plugins are loaded.

**3. multi-platform example with `explicit.txt`** — also a real bug, also fixed in the same commit. `explicit` only registers `export=` (single-platform), so the multi-platform line was lying. now gated behind a separate `example_filename_for(..., require_multiplatform=True)` lookup — without a multi-platform-capable lockfile plugin the entire block is omitted.

**4. `conda-lock-v1` / `rattler-lock-v6` in the wrong section** — that's a stale `conda-lockfiles`. `EnvironmentFormat.lockfile` was added in conda-incubator/conda-lockfiles@47386ed, which is post-0.1.1 and ships in [conda-incubator/conda-lockfiles#135](https://github.com/conda-incubator/conda-lockfiles/pull/135) (0.2.0, going out today). a `pip install -e .` from `main` should sort it; otherwise the release will.

**5. `Lockfiles:` section without the plugin** — that's intentional: the built-in `explicit` exporter is itself classified as a lockfile (`environment_format=EnvironmentFormat.lockfile`), so the section is correct even with no third-party plugin. only renders when at least one lockfile-classified plugin is registered.

**6. `requirements (reqs, txt)` reading like pip-style `reqs.txt`** — fair concern, but the labels live on the plugin definition, not in this PR. happy to do a follow-up that surfaces `description` in the help text — `CondaEnvironmentExporter.description` is already populated, just not rendered. opening an issue.

PR description updated to reflect all of the above. ready for another look whenever you have a sec!

conda/plugins/manager.py
kenodegard, line 1: I don't think these help text helpers belong here in the plugin_manager since they're not plugin loaders.

conda/plugins/manager.py
```diff
@@ -856,6 +857,124 @@ def get_exporter_format_mapping(self) -> dict[str, CondaEnvironmentExporter]:
...
+        :meth:`get_environment_exporters` or ``get_hook_results(
+        "environment_specifiers")``.
+        """
+        groups: dict[
+            EnvironmentFormat,
+            list[CondaEnvironmentExporter | CondaEnvironmentSpecifier],
+        ] = {}
+        for plugin in plugins:
+            groups.setdefault(plugin.environment_format, []).append(plugin)
+        return groups
```
kenodegard: We already have `groupby_to_dict`:

```python
groupby_to_dict(lambda plugin: plugin.environment_format, plugins)
```

conda/plugins/manager.py
```diff
@@ -856,6 +857,124 @@ def get_exporter_format_mapping(self) -> dict[str, CondaEnvironmentExporter]:
...
+
+    @staticmethod
+    def _format_entries(
+        plugins: Iterable[CondaEnvironmentExporter | CondaEnvironmentSpecifier],
+    ) -> list[str]:
+        """Render one entry per plugin as ``name`` or ``name (alias, ...)``."""
+        return [
+            f"{p.name} ({', '.join(p.aliases)})" if p.aliases else p.name
+            for p in plugins
+        ]
```
kenodegard: This generic formatter would be good to group with `dals` and `dashlist`, or do we adopt a new submodule pattern (e.g., `*.formats` submodules are for formatting heleprs)?

conda/plugins/manager.py
jezdez: Agreed, moved them out. They now live in a `FormatSummary` class in `conda/plugins/formats.py` — groups plugins once on construction, then `describe()` and `example_filename()` do the rendering. The plugin manager stays focused on loading.

conda/plugins/manager.py
jezdez: Looked into it — `groupby_to_dict` uses `itertools.groupby` under the hood which needs pre-sorted input to group correctly. Since plugin registration order isn't sorted by format, it would silently produce duplicate keys. Went with `dict.setdefault` instead, which handles unsorted input naturally and lives inside `FormatSummary.__init__` so it only runs once.

conda/plugins/manager.py
jezdez: Went with a new `conda/plugins/formats.py` module since the logic is plugin-specific (it reads `environment_format`, `aliases`, `default_filenames` off plugin types). Putting it next to `dals`/`dashlist` would have created a dependency from `conda/common` back into `conda/plugins`. The `FormatSummary` class keeps it self-contained — it uses `dashlist` for rendering but owns the grouping and format awareness.

conda/plugins/manager.py
kenodegard: > `groupby_to_dict` uses `itertools.groupby` under the hood which needs pre-sorted input to group correctly

Yes, `groupby_to_dict` uses `itertools.groupby` but it extends the values so pre-sorting is not required.

> Went with `dict.setdefault` instead

We can update `groupby_to_dict` to use `dict.setdefault` if we dislike the `collections` import. 🤷🏻‍♂️

jezdez: You're right, I misread the implementation - the `.extend()` call handles unsorted input correctly. Happy to switch to `groupby_to_dict` in `FormatSummary.__init__` if you'd prefer that over the inline `setdefault` loop. Either works, your call.

conda/cli/main_create.py
```diff
@@ -29,30 +32,61 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
...
+    formats = FormatSummary(
+        context.plugin_manager.get_hook_results("environment_specifiers")
+    )
+    spec_example = formats.example_filename(
+        EnvironmentFormat.environment,
+        prefer_filenames=("environment.yml", "environment.yaml"),
+    )
+    lock_example = formats.example_filename(
+        EnvironmentFormat.lockfile,
+        prefer_filenames=("conda-lock.yml", "pixi.lock"),
```
kenodegard: I disagree with the `prefer_filenames` approach. It makes the help text deterministic at the cost of introducing a leaky abstraction. That is especially awkward for the lockfile example; conda has no notion of either `conda-lock.yml` or `pixi.lock`, yet we hardcode it to shape the epilog.

I would rather use format-agnostic examples and keep concrete filenames out of the examples.

conda/plugins/formats.py
```diff
@@ -0,0 +1,115 @@
...
+                f"{plugin.name} ({', '.join(plugin.aliases)})"
+                if plugin.aliases
+                else plugin.name
+                for plugin in plugins
+            ]
+            sections.append(f"{category.label}:{dashlist(entries)}")
+
+        return "\n\n".join(
+            [f"{heading}:", *sections] if heading and sections else sections
+        )
```
kenodegard: Can we extend the list to also include other important information like `detection_supported` (where applicable) and `default_filenames`?

tests/cli/test_main_create.py
```diff
@@ -0,0 +1,35 @@
...
+    ],
+)
+def test_create_help_shows_examples_and_available_formats(
+    conda_cli: CondaCLIFixture, line: str
+) -> None:
+    """The rewritten `conda create --help` renders the command's example
+    sections and a dynamic listing of available input formats grouped by
+    category."""
+    stdout, _, _ = conda_cli("create", "--help", raises=SystemExit)
+    assert line in stdout
```
kenodegard: Better to use subtests and avoid rerunning `conda create --help`

```suggestion
def test_create_help_shows_examples_and_available_formats(
    conda_cli: CondaCLIFixture, subtests
) -> None:
    """The rewritten `conda create --help` renders the command's example
    sections and a dynamic listing of available input formats grouped by
    category."""
    stdout, _, _ = conda_cli("create", "--help", raises=SystemExit)
    for line in [
        "Create from package specs:",
        "conda create -n myenv python=3.12 numpy",
        "Create from an environment spec (solved at install time):",
        "Create from a lockfile (no solve, exact reproduction):",
        "Clone an existing environment:",
        "conda create -n env2 --clone env1",
        "Available input formats:",
        "Environment specs:",
    ]:
        with subtests.test(line):
            assert line in stdout
```

tests/cli/test_main_export.py
```diff
@@ -988,3 +990,68 @@ def test_export_invalid_platform_from_condarc_fails_fast(
...
+        "Environment specs:",
+    ],
+)
+def test_export_help_shows_examples_and_available_formats(
+    conda_cli: CondaCLIFixture, line: str
+) -> None:
+    """The rewritten `conda export --help` renders the example sections and a
+    dynamic listing of available formats grouped by category."""
+    stdout, _, _ = conda_cli("export", "--help", raises=SystemExit)
+    assert line in stdout
```
kenodegard:
```suggestion
def test_export_help_shows_examples_and_available_formats(
    conda_cli: CondaCLIFixture, subtests
) -> None:
    """The rewritten `conda export --help` renders the example sections and a
    dynamic listing of available formats grouped by category."""
    stdout, _, _ = conda_cli("export", "--help", raises=SystemExit)
    for line in [
        "Export an environment spec:",
        "conda export --from-history",
        "Export a lockfile for the same platform:",
        "Available formats:",
        "Environment specs:",
    ]:
        with subtests.test(line):
            assert line in stdout
```

conda/cli/main_create.py
jezdez: I understand, it was a tradeoff I was willing to accept given the relative uselessness when using built-in formats alone. Could you provide an example how using agnostic example would look like compared to the full example above?

conda/plugins/formats.py
jezdez: Sure why not

jezdez:
@kenodegard @dashagurova Pushed a new commit addressing the open review threads:

**Dropped `prefer_filenames`** — the built-in example filenames (`environment.yml`, `explicit.txt`) are now hardcoded strings in each parser. No third-party filenames leak into conda core. The multi-platform lockfile example stays dynamic (it requires a plugin that declares `multiplatform_export`, so using that plugin's `default_filenames[0]` is fine).

**Format listing updated** — the `Available formats:` section now renders each entry as:

```
name (aliases: alias, ...): file, ...
```

This shows both what to pass to `--format` (the aliases) and what triggers auto-detection via `--file` (the default filenames). For example:

```
Environment specs:
  - environment-yaml (aliases: yaml, yml, env.yml): environment.yaml, environment.yml
  - requirements (aliases: reqs, txt): requirements.txt, spec.txt
```

Further enrichment (`detection_supported`, `description`) is deferred to a follow-up.

Net change: -84 lines. PR description updated.

kenodegard (approved):
Pushed commits that keep the **same user-facing behavior** we were aiming for but with a slightly different split of responsibilities:

- Help rendering lives in `conda.cli.formats` (`get_available_environment_formats`, `get_multiplatform_lockfile`) instead of introducing `conda.plugins.formats.FormatSummary`.
- `conda.plugins.manager.CondaPluginManager` exposes `get_environment_exporters_grouped()` / `get_environment_specifiers_grouped()` to map `EnvironmentFormat` to plugin lists. That retains the manager's focus on “discovery / structured view of plugins” while `conda.cli` does everything `argparse`/CLI related.
- Examples still use hard-coded `environment.yml` / `explicit.txt`; the multi-platform line and the bullet list stay plugin-driven.

PR #15968, opened 2026-04-26
Display plugin settings in conda config --show-sources

### LINKED ISSUE #15912: `conda config --show-sources` does not display plugin settings
### Checklist

### What happened?

Plugin settings registered via the `conda_settings` hook and configured under the `plugins:` key in `.condarc` are not shown by `conda config --show-sources`.

**Steps to reproduce:**

1. Install a plugin that registers a setting via `conda_settings` (e.g. `conda-self` with its `self_permanent_packages` setting)
2. Configure it in `.condarc`:
   ```yaml
   plugins:
     self_permanent_packages:
       - anaconda-anon-usage
   ```
3. Run `conda config --show-sources`

**Expected:** The `plugins.self_permanent_packages` setting appears in the output.

**Actual:** The setting is absent from the output, even though it is correctly saved in `.condarc` and works at runtime.

### Root cause

`conda config --show-sources` calls `context.collect_all()`, which iterates over `Context.raw_data`. Plugin settings live in a separate `PluginConfig` object accessed via `context.plugins`, which has its own `raw_data` property that extracts the `plugins:` section from the condarc. The `--show-sources` code path never calls `context.plugins.collect_all()`, so plugin settings are invisible in the output.

The same likely applies to `conda config --show`, plugin settings under `plugins:` are not surfaced there either.

### Additional Context

Discovered during review of https://github.com/conda-incubator/conda-self/pull/116, which adds a `self_permanent_packages` plugin setting. See https://github.com/conda-incubator/conda-self/pull/116#discussion_r3068998694.

prady0t:
### Description

Fixes #15912 

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_config.py
+++ b/conda/cli/main_config.py
@@ -452,9 +452,20 @@ def execute_config(args: Namespace, parser: ArgumentParser) -> int | None:
             )
         else:
             lines = []
+            plugin_raw = context.plugins.raw_data
             for source, reprs in context.collect_all().items():
                 lines.append(f"==> {source} <==")
                 lines.extend(format_dict(reprs))
+                lines.append("")
+            for source, plugins in plugin_raw.items():
+                lines.append(f"==> {source} <==")
+                formatted = {
+                    "plugins": {
+                        name: meta._raw_value
+                        for name, meta in plugins.items()
+                    }
+                }
+                lines.extend(format_dict(formatted))
                 lines.append("")
             stdout_write("\n".join(lines))
         return 0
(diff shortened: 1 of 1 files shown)


conda/cli/main_config.py
```diff
@@ -452,10 +452,21 @@ def execute_config(args: Namespace, parser: ArgumentParser) -> int | None:
             )
         else:
             lines = []
+            plugin_raw = context.plugins.raw_data
             for source, reprs in context.collect_all().items():
                 lines.append(f"==> {source} <==")
                 lines.extend(format_dict(reprs))
                 lines.append("")
+            for source, plugins in plugin_raw.items():
```
prady0t, line 460: `collect_all` here gives inaccurate output. For example, in the `.condarc` file:

```yml
plugins:
  anaconda_telemetry: true
```

Gives:

```
plugins: true
```

instead of

```
plugins:
  anaconda_telemetry: true
```
Hence, I extracted the info from the raw_data itself.

Issue #15969, opened 2026-04-27
Reduce transaction latency: Track B implementation plan


jezdez:
Part of the [**conda-tempo**](https://github.com/jezdez/conda-tempo) performance initiative · [Track A — startup latency](https://github.com/conda/conda/issues/15867) · **Track B — transaction latency** (this issue) · [Track C — Python 3.15 and speculative research](https://github.com/jezdez/conda-tempo/blob/main/track-c-future.md)

### What?

Reduce conda's post-solver transaction latency through targeted changes across four repositories (`conda`, `conda-libmamba-solver`, `conda-package-handling`, `conda-package-streaming`). Same constraints as Track A: Python 3.10+ only, one PR per suspect targeting <100 LOC each where practical, no architectural changes, and every PR quotes a before/after micro-benchmark number plus an end-to-end hyperfine result when available.

Canonical executive summary, workload table, shipping status, remaining headroom, and measurement links live in the research report: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#executive-summary

### Why?

Track A reduced conda's startup cost. Track B reduces what happens *after* the solver returns: the transaction pipeline that verifies, downloads, extracts, and links packages onto disk. Users perceive this as "the transaction status lines print one after another, and each one takes forever." The lines are a symptom. The underlying code is serial in spots, quadratic in others, and hashes files unconditionally in a few places.

Related issues: #14500 (fixed by #15320, closed. Post-solver machinery still pays the full cost on any install path), #14426, #13115, #11814.

### Measured impact

The April macOS/Linux stack measures at roughly **10-20% faster on typical installs** and **20-40x faster on commands against large existing prefixes**. Full methodology, workload tables, Windows checkpoint notes, and caveats live in the [research report](https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#executive-summary).

Windows x86_64 has dedicated checkpoint data as of July 2026. Those numbers validate individual follow-up PR direction but are not folded into the headline table because the July branch mix did not reproduce the April macOS/Linux full-stack behavior.

These headline numbers measure the April 2026 full experimental branch stack, not the current eight-merged-PR subset. That stack included B4, B14, and B20, which have not all shipped. The merged subset has not been remeasured end to end, so the totals should not be presented as current conda performance.

[...]

PR #15970, opened 2026-04-27
Linear-time sort in `diff_for_unlink_link_precs` (B1)

jezdez:
### Description

Replaces the quadratic order-restoration step in `diff_for_unlink_link_precs` (`conda/core/solve.py`) with a linear pass over the already-ordered record sequences.

The old code did:

- `sorted(unlink_precs, key=lambda x: previous_records.index(x))`
- `sorted(link_precs, key=lambda x: final_precs.index(x))`

Each `tuple.index` scans the ordered sequence, so restoring order for a 2,000-unlink/link transaction against a 50,000-record prefix does millions of repeated position scans.

This PR keeps `unlink_precs` / `link_precs` as sets, then restores order by filtering the source-of-truth ordered sequences directly:

- `previous_records` in reverse order for unlink records
- `final_precs` in forward order for link records

That preserves the same output order without sorting or per-record index lookups. No semantic change.

### Focused measurement

Synthetic fixture with N=50,000 prefix records and k=2,000 unlink/link records:

| | Repeated index scan | Linear order restoration | Speedup |
|---|---:|---:|---:|
| `diff_for_unlink_link_precs` | 12.46 s | 15.9 ms | 782x |

At N=10,000 and k=400, the same focused benchmark measured a 215x speedup. These figures measure this function only. The earlier full-stack timing was removed because it covered a superseded combination of Track B changes rather than this PR in isolation.

### Checklist - did you ...

- [ ] Add / update necessary tests? Existing solve tests cover the semantics; happy to add a regression benchmark if desired.

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1462,9 +1462,9 @@ def diff_for_unlink_link_precs(
         for prec in noarch_python_precs:
             _add_to_unlink_and_link(prec)
 
-    unlink_precs_tuple = (
+    unlink_precs_tuple = tuple(
         rec for rec in reversed(previous_records) if rec in unlink_precs
     )
-    link_precs_tuple = (rec for rec in final_precs if rec in link_precs)
+    link_precs_tuple = tuple(rec for rec in final_precs if rec in link_precs)
 
     return unlink_precs_tuple, link_precs_tuple
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -1462,21 +1462,9 @@ def diff_for_unlink_link_precs(
         for prec in noarch_python_precs:
             _add_to_unlink_and_link(prec)
 
-    unlink_precs = tuple(
-        reversed(sorted(unlink_precs, key=_prec_order_key(previous_records)))
+    unlink_precs_tuple = (
+        rec for rec in reversed(previous_records) if rec in unlink_precs
     )
-    link_precs = tuple(sorted(link_precs, key=_prec_order_key(final_precs)))
-    return unlink_precs, link_precs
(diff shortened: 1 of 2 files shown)


conda/core/solve.py
```diff
@@ -1463,7 +1463,20 @@ def _add_to_unlink_and_link(rec):
             _add_to_unlink_and_link(prec)
 
     unlink_precs = tuple(
-        reversed(sorted(unlink_precs, key=lambda x: previous_records.index(x)))
+        reversed(sorted(unlink_precs, key=_prec_order_key(previous_records)))
```
dholth: ouch

dholth:
I've committed an alternate implementation directly to your branch. Feel free to revert if you don't like it. All we have to do is iterate over the lists that determine the order, and yield those elements that are also in the unordered sets.

conda/core/solve.py
```diff
@@ -1462,8 +1462,9 @@ def _add_to_unlink_and_link(rec):
...
             _add_to_unlink_and_link(prec)
 
-    unlink_precs = tuple(
-        reversed(sorted(unlink_precs, key=lambda x: previous_records.index(x)))
+    unlink_precs_tuple = (
+        rec for rec in reversed(previous_records) if rec in unlink_precs
     )
-    link_precs = tuple(sorted(link_precs, key=lambda x: final_precs.index(x)))
-    return unlink_precs, link_precs
+    link_precs_tuple = (rec for rec in final_precs if rec in link_precs)
```
dholth: A forbidden `assert len(before) == len(after)`?

jaimergp (approved):

PR #15971, opened 2026-04-27
Index exact-name PrefixGraph dependencies (B2)

jezdez:
### Description

Adds a per-name candidate index to `PrefixGraph.__init__()` for exact-name dependencies. Non-exact names, including globs, retain the full scan to preserve `MatchSpec` semantics.

### Focused measurement

Synthetic DAG fixture:

| N | Repeated candidate scan | Name index | Speedup |
|---:|---:|---:|---:|
| 1,000 | 4.18 s | 78 ms | 53x |

The focused result measures the exact-name path. The earlier full-stack W3 figures were removed because they covered a superseded combination of Track B changes, rather than this PR in isolation.

### Checklist - did you ...

- [x] Add / update necessary tests? Added coverage for globbed names and multiple same-name candidates in `GeneralGraph`.

### DIFF
--- a/conda/models/prefix_graph.py
+++ b/conda/models/prefix_graph.py
@@ -52,12 +52,9 @@ class PrefixGraph:
         self.graph = graph = {}
         self.spec_matches: dict[PrefixRecord, dict[MatchSpec, None]]
         self.spec_matches = spec_matches = {}
-        # Index records by package name so the parent-match inner loop
-        # iterates only the name-matching candidates instead of every
-        # record. MatchSpec(dep).name is always set; any record whose
-        # name doesn't match would have been filtered out by the
-        # original MatchSpec.match() anyway, just at O(N) per dep
-        # rather than O(1). See #15971 for the before/after benchmark.
+        # Index records by package name so exact-name dependencies only
+        # consider name-matching candidates. Non-exact names still scan
+        # every record. See #15971 for the before/after benchmark.
         by_name: dict[str, list[PrefixRecord]] = {}
         for rec in records:
             by_name.setdefault(rec.name, []).append(rec)
@@ -65,7 +62,9 @@ class PrefixGraph:
             parent_nodes: dict[PrefixRecord, None] = {}
             for dep in node.depends:
                 spec = MatchSpec(dep)
-                for candidate in by_name.get(spec.name, ()):
+                name = spec.get_exact_value("name")
+                candidates = records if name is None else by_name.get(name, ())
+                for candidate in candidates:
                     if spec.match(candidate):
                         parent_nodes[candidate] = None
             graph[node] = parent_nodes
--- a/conda/models/prefix_graph.py
+++ b/conda/models/prefix_graph.py
@@ -52,13 +52,22 @@ class PrefixGraph:
         self.graph = graph = {}
         self.spec_matches: dict[PrefixRecord, dict[MatchSpec, None]]
         self.spec_matches = spec_matches = {}
+        # Index records by package name so the parent-match inner loop
+        # iterates only the name-matching candidates instead of every
+        # record. MatchSpec(dep).name is always set; any record whose
+        # name doesn't match would have been filtered out by the
+        # original MatchSpec.match() anyway, just at O(N) per dep
+        # rather than O(1). See #15971 for the before/after benchmark.
+        by_name: dict[str, list[PrefixRecord]] = {}
+        for rec in records:
+            by_name.setdefault(rec.name, []).append(rec)
         for node in records:
-            parent_match_specs = tuple(MatchSpec(d) for d in node.depends)
-            parent_nodes = {
-                rec: None
-                for rec in records
-                if any(m.match(rec) for m in parent_match_specs)
-            }
+            parent_nodes: dict[PrefixRecord, None] = {}
+            for dep in node.depends:
+                spec = MatchSpec(dep)
+                for candidate in by_name.get(spec.name, ()):
+                    if spec.match(candidate):
+                        parent_nodes[candidate] = None
             graph[node] = parent_nodes
             matching_specs = dict.fromkeys(s for s in specs if s.match(node))
             if matching_specs:
--- a/news/15971-track-b-b2-prefix-graph
+++ b/news/15971-track-b-b2-prefix-graph
@@ -1,0 +1,19 @@
+### Enhancements
+
+* Build a per-name candidate index in `PrefixGraph.__init__()` to replace the quadratic scan over all records. Speeds up transactions against large environments — ~50× at 1,000 records, far more at the scales where users notice it. (#15971)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/15971-track-b-b2-prefix-graph
+++ b/news/15971-track-b-b2-prefix-graph
@@ -1,6 +1,6 @@
 ### Enhancements
 
-* Build a per-name candidate index in `PrefixGraph.__init__()` to replace the quadratic scan over all records. Speeds up transactions against large environments — ~50× at 1,000 records, far more at the scales where users notice it. (#15971)
+* Build a per-name candidate index in `PrefixGraph.__init__()` to accelerate graph construction for large prefixes. (#15971)
 
 ### Bug fixes
 
--- a/tests/models/test_prefix_graph.py
+++ b/tests/models/test_prefix_graph.py
@@ -987,6 +987,18 @@ def test_general_graph_bfs_simple():
     assert backwards is None
 
 
+def test_prefix_graph_name_glob_dependency():
+    consumer = PackageRecord(
+        name="consumer", version="1", build="0", build_number=0, depends=["foo*"]
+    )
+    foo = PackageRecord(name="foo", version="1", build="0", build_number=0)
+    foobar = PackageRecord(name="foobar", version="1", build="0", build_number=0)
+
+    graph = PrefixGraph((consumer, foo, foobar))
+
+    assert graph.graph[consumer] == {foo: None, foobar: None}
+
+
 def test_general_graph_bfs_version():
     a = PackageRecord(
         name="a", version="1", build="0", build_number=0, depends=["b", "c", "d"]
@@ -1002,6 +1014,9 @@ def test_general_graph_bfs_version():
     g2 = PackageRecord(name="g", version="2", build="0", build_number=0)
     records = [a, b, c, d, e, f, g1, g2]
     graph = GeneralGraph(records)
+
+    assert graph.graph[c] == {g1: None}
+    assert graph.graph[d] == {f: None, g2: None}
 
     a_to_g1 = graph.breadth_first_search_by_name(MatchSpec("a"), MatchSpec("g=1"))
     assert a_to_g1 == [MatchSpec("a"), MatchSpec("c"), MatchSpec("g=1")]


jezdez:
@jaimergp I benchmarked trying to use py-rattler's MatchSpec as [S18 in the research report](https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#s18-py-rattler-matchspec-vs-condas-b2-review-follow-up). Post-B2 conda vs py-rattler on the same synthetic DAG fixture used in Phase-2:

**Parse cost** (`MatchSpec(s)` on N dep strings):

| N (total deps) | conda | rattler | speedup |
|---:|---:|---:|---:|
| 485 | 1.18 ms | 407 µs | **2.9×** |
| 4 985 | 12.3 ms | 3.26 ms | **3.8×** |
| 24 985 | 65.5 ms | 17.5 ms | **3.7×** |

**Match cost** (100 pre-parsed specs × 100 records = 10 k matches):

| | conda | rattler | speedup |
|---|---:|---:|---:|
| 10 000 matches | 7.4 ms | 1.4 ms | **5.3×** |

**PrefixGraph-equivalent** (this PR's `PrefixGraph(records).graph` vs `rattler.PackageRecord.sort_topologically(records)`):

| N records | post-B2 conda | rattler pure | rattler + conversion | pure | w/conversion |
|---:|---:|---:|---:|---:|---:|
| 100 | 2.71 ms | 184 µs | 712 µs | 15× | **3.8×** |
| 1 000 | 68.5 ms | 1.94 ms | 7.15 ms | 35× | **9.6×** |
| 5 000 | 1.28 s | 11.4 ms | 38.1 ms | 112× | **33×** |

"rattler + conversion" includes building fresh `rattler.PackageRecord` instances from conda `PrefixRecord` inputs per call. That's what any realistic drop-in swap would pay unless records are kept rattler-shaped throughout. Conversion alone is ~70 % of the hybrid-path time at N = 5 000.

**TL;DR**: rattler is faster across every axis: 4× parse, 5× match, 15-100× full graph construction. Even with full per-call conversion the hybrid path is still 3.8-33× faster than B2 alone.

**Design question** the numbers raise:

1. **Narrow swap**: replace just `PrefixGraph.__init__` internals with a call to `rattler.PackageRecord.sort_topologically`, converting records on entry. ~3-10× at typical scales, no API changes.
2. **Wider swap**: keep rattler records in `SolverInputState` / `UnlinkLinkTransaction` so conversion happens once per solve, not per PrefixGraph call. ~15-100×. Bigger PR.
3. **Record-type migration**: replace `conda.models.records` wholesale. Big architectural change, outside Track B scope.

I'd tentatively park it as a follow-up PR scoped just to option (1), so B2 can still ship as the pure-Python fix for environments where adding py-rattler as a conda dep isn't on the table yet (conda-forge packaging, bootstrap). Option (1)'s number (~10× at N=1 000 on top of B2's 53×) would still be a welcome win on large-prefix workloads.

Caveat on grammar coverage: rattler's MatchSpec parser is strict and rejects unknown bracket keys; conda's is permissive and silently ignores them. For example, rattler raises on `python[extras=['dev','test']]` and `python_abi[optional=true]`, while conda accepts both with partial interpretation. A real swap either direction would need a compatibility pass against real conda-forge / conda-channel specs first.

Bench script: [`bench/phase2/bench_s18_matchspec_rattler.py`](https://github.com/jezdez/conda-tempo/blob/main/bench/phase2/bench_s18_matchspec_rattler.py), raw pyperf output: [`data/phase2/s18_matchspec_rattler/`](https://github.com/jezdez/conda-tempo/tree/main/data/phase2/s18_matchspec_rattler).

jezdez:
Follow-up on the py-rattler idea. I wanted to see where the sensible integration seam actually is, so I tried two different levels of abstraction on throwaway branches to get concrete data before proposing a real PR. Writing up what I found so we can talk about which (if any) is worth taking further.

**Trial 1: microbench-only (S18).** Isolated `rattler.MatchSpec` and `rattler.PackageRecord.sort_topologically` against post-B2 conda on the same synthetic DAG fixture. 3-4× on parse, 5× on match, 15-112× on the PrefixGraph equivalent. Documented in the comment above. This is the upper bound, not realistic.

**Trial 2: MatchSpec facade.** Prototype at [`conda/conda:jezdez/experiment-matchspec-rattler-facade`](https://github.com/jezdez/conda/tree/jezdez/experiment-matchspec-rattler-facade). New `conda/models/_match_spec_rattler.py` helper plus ~25 LOC in `MatchSpec.__init__`/`.match()` that lazily builds a rattler spec, caches a `rattler.PackageRecord` view on each conda record, and delegates matches when rattler can parse the spec. Completely optional: no rattler -> no-op, rattler rejects the spec -> fall through to conda. 15 spot-checks across spec grammars match conda's reference behaviour.

What it does end-to-end on W3 @ 50 000, same session, 5 hyperfine runs each:

| Configuration | mean | stddev |
|---|---:|---:|
| Stack baseline, no facade | 15.05 s | ±1.65 s |
| Stack + MatchSpec facade | 17.35 s | ±0.13 s |

The facade regresses by ~2.3 s with tight variance. Root cause: in the solver hot paths most `MatchSpec` instances are one-shot (built, matched once or twice, discarded). The per-spec rattler-parse cost is paid on every one of them; the per-match speedup doesn't compensate because each spec matches too few records to amortise. The record-side conversion cache works correctly, but 50 000 distinct records each touched a handful of times don't amortise the per-record conversion either.

So the generic `MatchSpec` facade is the wrong seam: it sits in a place where most callers only match once.

**Trial 3 (not yet built, what I'd try next if it's worth doing):** delegate only inside `PrefixGraph.__init__` to `rattler.PackageRecord.sort_topologically` when rattler is available. Skips the per-spec facade entirely. PrefixGraph does one record-pool conversion per call, then runs the whole graph build in Rust. The conversion amortises over the hundreds-of-matches this call site actually does, so it matches the "rattler + conversion" column in S18 (3.8-33× depending on N). Keeps B2 as the pure-Python fallback when rattler isn't installed.

**Trial 4 (architectural):** hold `rattler.PackageRecord` in `SolverInputState` / `UnlinkLinkTransaction` so conversion happens once per solve. Full S18 speedup (15-112× on PrefixGraph). Not a small PR; crosses into conda-libmamba-solver which is already on B11.

Writeup with per-trial data and takeaways lives in [S18b of the research report](https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#s18b-matchspec-facade-prototype-the-naive-approach-regresses). Would be curious which seam feels right to you before I put more time into trial 3 or 4. My read is that trial 3 is cheap to prototype and enough to call, but you might have opinions about how deep any rattler integration should reach into the solver state.

jaimergp:
> Writeup with per-trial data and takeaways lives in [S18b of the research report](https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md?rgh-link-date=2026-04-27T12%3A46%3A15Z#s18b-matchspec-facade-prototype-the-naive-approach-regresses). Would be curious which seam feels right to you before I put more time into trial 3 or 4. My read is that trial 3 is cheap to prototype and enough to call, but you might have opinions about how deep any rattler integration should reach into the solver state.


What I had envisioned (without any actual prototyping yet), it's to come up with reduced API surface for all the critical models and core objects, which presents us with an opportunity to:

- Clean up the classes API in `conda`
- Delegate to either backend and/or implement fallback strategies

So, for `MatchSpec` I would do:

```python
@dataclass(kw_only=True)
class MatchSpec:
    name: str = "*"
    version: str | None = None
    build: str | None = None
    ... # and so on

	def __post_init__(self):
        self._inner = ... # either conda or rattler
    @classmethod
    def from_string(cls, value: str) -> Self:
        pass
    def match(self, other: Record) -> bool:
        pass
    def __str__(self) -> str:
        pass
```

jezdez:
I built the companion bench we talked about. Using a fresh bench_big at sizes from 1k to 50k records, I ran `conda install -c conda-forge -y --dry-run requests`. That install has to pull in about 15 transitive deps that aren't in bench_big, so the solver has to do real work, and the post-solve step where conda compares the current prefix against the proposed plan (`diff_for_unlink_link_precs`) does a full `PrefixGraph` pass over the prefix on the way out. That's the place where the PrefixGraph-rattler seam would actually fire.

| N      | B2 stack | + rattler PrefixGraph | speedup |
|-------:|---------:|----------------------:|--------:|
| 1 000  | 1.75 s   | 1.90 s                | 0.92x   |
| 5 000  | 4.71 s   | 2.95 s                | 1.6x    |
| 10 000 | 12.60 s  | 4.11 s                | 3.1x    |
| 50 000 | 628.8 s  | 21.3 s                | 29.5x   |

For small prefixes (around 1k records or fewer), the per-call cost of converting conda's `PrefixRecord` to rattler's `PackageRecord` eats the time the rattler path saves, and the rattler build comes out very slightly slower than the pure-Python B2 path. Somewhere between 3k and 5k that flips and the rattler build starts winning; by 50k it's saving the best part of ten minutes.

I wrote it up alongside the S19 microbench and the W5 end-to-end numbers in the research report: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#w5-conda-install-real-pkg-against-realistic-bench_big-end-to-end

For shipping this, my thinking is to merge the current PR as-is, so the pure-Python B2 is both the baseline improvement and the fallback, and then do a follow-up PR that adds an optional rattler fast path inside `PrefixGraph.__init__`. The follow-up would do a `try: import rattler` at module load and fall back to the B2 name-indexed loop when rattler isn't available, which keeps py-rattler as a soft dep of conda.

I went ahead and opened the follow-up as a draft PR so there's something concrete to look at: #15980. It's stacked on this PR and explicitly marked do-not-merge-before-#15971 in the body. Happy to flesh it out further once this one lands, or hold off if you'd rather first have a conversation about py-rattler as a hard dep of conda.

jezdez:
Sorry @jaimergp, I posted the W5 numbers and the draft PR link without engaging with this properly. What you're describing is a bigger move than what I built, and I think it's the right long-term direction.

A thin `@dataclass` facade with `_inner` backing either conda or rattler fixes two things at once: the accumulated cruft in the current MatchSpec / Record / VersionSpec APIs, and the fact that today there's no clean seam for plugging an alternate backend in at all. The API cleanup is probably the bigger win, actually.

Two things I'd want to figure out first. Whether it's just MatchSpec or MatchSpec + Record + VersionSpec + PackageName all at once, since rattler's types hold references to each other and mixing conda-side and rattler-side versions inside one solve gets messy. And how to handle the fact that rattler's parser is stricter than conda's (rejects `[extras=...]` and `[optional=...]` that conda silently ignores); the facade either rejects them upfront, or keeps a conda-backed fallback for the sake of legacy tolerance.

#15980 is still useful as a follow-up to B2 that ships the speedup today, but it's not in any hurry. Happy to close it or leave it in draft until the shape of the bigger refactor is clearer. I can also take a pass at auditing conda + known-plugin call sites of the current model classes if that would help scope the migration.

conda/models/prefix_graph.py
```diff
@@ -52,13 +52,22 @@ def __init__(
         self.graph = graph = {}
         self.spec_matches: dict[PrefixRecord, dict[MatchSpec, None]]
         self.spec_matches = spec_matches = {}
+        # Index records by package name so the parent-match inner loop
+        # iterates only the name-matching candidates instead of every
+        # record. MatchSpec(dep).name is always set; any record whose
```
jaimergp: Just to have this documented in the conversation, because I didn't see any mentions: we are assuming the specs do not contain glob in the names, right?

conda/models/prefix_graph.py
```diff
@@ -52,13 +52,22 @@ def __init__(
...
+        # name doesn't match would have been filtered out by the
+        # original MatchSpec.match() anyway, just at O(N) per dep
+        # rather than O(1). See #15971 for the before/after benchmark.
+        by_name: dict[str, list[PrefixRecord]] = {}
+        for rec in records:
+            by_name.setdefault(rec.name, []).append(rec)
```
jaimergp, line 60: We should never have more than one record with the same name, why do we need a list?

conda/models/prefix_graph.py
jezdez: No. MatchSpec accepts name globs, so non-exact names now use the full scan; only exact names use the index. Added coverage in 85b1998.

conda/models/prefix_graph.py
jezdez: GeneralGraph subclasses this and intentionally has multiple candidates with the same name. A plain mapping would lose version-specific matches, so the list is needed. Added a direct regression in 85b1998.

jaimergp (approved):

PR #15973, opened 2026-04-27
Verify transaction actions in parallel (B6)

jezdez:
### Description

Transaction verification already owns a shared executor sized by `verify_threads`, but B6 previously fanned out only by prefix and then created a second executor for the actions inside each prefix. A normal install usually has one target prefix, so that shape either stayed serial or created nested pools.

This PR now flattens all unverified transaction actions into one ordered stream and maps their `verify()` calls directly across the existing executor. Prefix-level checks reuse the same executor after individual verification completes.

Since B9c (#15975) now batches codesign calls during verification, each executor worker returns its queued signing paths to the parent, which flushes one batch before prefix-level verification. This keeps batching effective without sharing thread-local state.

The default remains one verification worker. Setting `verify_threads` or `CONDA_VERIFY_THREADS` above one enables parallel action verification while preserving deterministic result ordering.

This removes `_verify_individual_level` and avoids a nested executor.

### Checklist - did you ...

- [x] Add / update outdated documentation? The existing `verify_threads` setting now controls action verification directly.

### DIFF
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -64,7 +64,7 @@ from ..reporters import confirm_yn, get_spinner
 from ..resolve import MatchSpec
 from ..utils import get_comspec, human_bytes, wrap_subprocess_call
 from .package_cache_data import PackageCacheData
-from .portability import batch_codesign_calls
+from .portability import batch_codesign_calls, codesign_paths
 from .path_actions import (
     AggregateCompileMultiPycAction,
     CompileMultiPycAction,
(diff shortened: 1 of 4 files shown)


jaimergp (approved):

jezdez:
Closing B6 after re-running it against current `main` after B9c merged.

Randomized paired real-package runs consistently regressed the prepare/verify phase:

- macOS W1: K=2 **+22.7%**, K=4 **+28.0%**
- macOS W2: K=2 **+35.6%**, K=4 **+39.5%**
- Linux arm64 W1: K=2 **+14.1%**, K=4 **+14.4%**

B9c moved codesign out of each action. The current workloads are now dominated by thousands of roughly 24 us path checks, so mapping every action through the executor costs more than the work it overlaps. The existing test's 50 ms sleep proves concurrency but does not represent the shipping action mix.

The original optimization premise no longer holds. Any future attempt should batch actions or target only demonstrably expensive work, with fresh end-to-end evidence.

Full measurements and disposition: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#s6-current-main-remeasurement-and-disposition-2026-07-20

PR #15974, opened 2026-04-27
Extract packages with a process pool (B8)

jezdez:
### Description

This updates B8 from a thread-cap-only change to Daniel's suggested process-pool approach for package extraction.

The motivating S8 fresh-cache trial used conda-tempo's documented package set (python=3.13 pandas scikit-learn jupyter). Four spawn-based process workers reduced the extraction phase from a 2.395 s serial median to 1.534 s, a 1.56x speedup. The full benchmark is recorded in the Track B report linked below.

Implementation details:

- Add conda._private.extract.extract_conda_package_archive as the lean built-in archive extraction helper.
- Keep filesystem preparation and package-cache metadata updates in the parent process.
- Use a spawn-based ProcessPoolExecutor only when every pending action uses conda's built-in extractor and process extraction is supported.
- Keep custom extractors, unsupported archive types, and debug mode on the existing thread path.
- Drain and finalize already-submitted extraction work even when a later fetch fails, so successful packages are not left extracted without their package-cache metadata.

### Checklist - did you ...

### DIFF
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -1,0 +1,20 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Lean package extraction helpers for subprocess workers."""
+
+from __future__ import annotations
+
+import os
+
+
+def extract_conda_package_archive(
+    source_full_path: str | os.PathLike,
+    destination_directory: str | os.PathLike,
+) -> None:
+    """Extract a conda package archive without importing conda runtime state."""
+    import conda_package_handling.api
+
+    conda_package_handling.api.extract(
+        os.fspath(source_full_path),
+        dest_dir=os.fspath(destination_directory),
+    )
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -6,6 +6,10 @@ from __future__ import annotations
 
 import os
 
+# This module is imported in each spawned extraction worker. Keep imports here
+# limited to the standard library; importing conda runtime state defeats the
+# process pool's startup benefit.
+
 
 def extract_conda_package_archive(
     source_full_path: str | os.PathLike,
--- a/conda/core/package_cache_data.py
+++ b/conda/core/package_cache_data.py
@@ -4,9 +4,15 @@
 
 from __future__ import annotations
 
+import multiprocessing
 import os
 from collections import defaultdict
-from concurrent.futures import CancelledError, ThreadPoolExecutor, as_completed
+from concurrent.futures import (
+    CancelledError,
+    ProcessPoolExecutor,
+    ThreadPoolExecutor,
+    as_completed,
+)
 from errno import EACCES, ENOENT, EPERM, EROFS
 from functools import partial
 from itertools import chain
@@ -18,12 +24,14 @@ from tarfile import ReadError
 from typing import TYPE_CHECKING
 
 from .. import CondaError, CondaMultiError, conda_signal_handler
+from .._private.extract import extract_conda_package_archive
 from ..auxlib.collection import first
 from ..auxlib.decorators import memoizemethod
 from ..auxlib.entity import ValidationError
 from ..base.constants import (
     CONDA_PACKAGE_EXTENSION_V1,
     CONDA_PACKAGE_EXTENSION_V2,
+    CONDA_PACKAGE_EXTRACTOR_NAME,
     PACKAGE_CACHE_MAGIC_FILE,
 )
 from ..base.context import context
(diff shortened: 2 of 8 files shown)


dholth:
Extract threads run into the GIL because `zstd` compression is really fast, compared to `bz2` which benefits more from Python threading.
3-6 processes can unpack `.conda` productively on my M1 laptop [according to this subinterpreter experiment](https://github.com/dholth/conda/commit/857a0447baf6f1f8709d5ccf7dec159e391b9cf3).
We could consider doing the same thing in a `ProcessPoolExecutor`. To reduce startup time, it would be helpful to make sure the subprocesses or subinterpreters do not import the rest of conda, i.e. a path designed not to import context `import conda._doesnt_import_context.extract`

jezdez:
@dholth thanks again for the process-pool suggestion. I reworked this draft so it is no longer just lowering `EXTRACT_THREADS`. It now uses a spawn-based `ProcessPoolExecutor` for conda's built-in `.conda` / `.tar.bz2` extraction with a lean `conda._private.extract` worker. The parent process keeps package-cache metadata and progress updates, and custom extractors stay on the plugin/thread path.

I also updated the PR description with the S8 numbers. The selected spawn-4 path measured 1.534 s median vs 2.395 s serial and 2.541 s thread-4. Does this match what you had in mind?

conda/_private/extract.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Lean package extraction helpers for subprocess workers."""
```
jaimergp, line 3: Can we add a note here saying that expensive imports are forbidden because this is a hot path?

conda/core/package_cache_data.py
```diff
@@ -71,8 +79,15 @@
     from conda_package_handling.api import THREADSAFE_EXTRACT
 except ImportError:
     THREADSAFE_EXTRACT = False
-# On the machines we tested, extraction doesn't get any faster after 3 threads
-EXTRACT_THREADS = min(os.cpu_count() or 1, 3) if THREADSAFE_EXTRACT else 1
+# The fallback thread pool is used for non-standard package extractor plugins
+# and debug mode. Built-in .conda/.tar.bz2 extraction normally runs in a
+# process pool below to avoid zstd's GIL bottleneck. See #15974.
+EXTRACT_THREADS = 2 if THREADSAFE_EXTRACT else 1
```
jaimergp: You have change this constant from `min(os.cpu_count() or 1, 3)` to 2. Is that intentional?

conda/core/path_actions.py
```diff
@@ -1466,22 +1466,25 @@ def verify(self):
...
-        # The alternative is passing the the classes to ExtractPackageAction __init__
-        from .package_cache_data import PackageCacheData
+        self.prepare_extract()
+        context.plugin_manager.extract_package(
+            self.source_full_path,
+            self.target_full_path,
+        )
+        self.finish_extract()
 
+    def prepare_extract(self):
```
jaimergp: Do we want these as public methods?

conda/core/path_actions.py
```diff
@@ -1466,22 +1466,25 @@ def verify(self):
...
 
         if lexists(self.target_full_path):
             rm_rf(self.target_full_path)
 
-        # extract the package using the appropriate plugin
-        context.plugin_manager.extract_package(
-            self.source_full_path,
-            self.target_full_path,
-        )
+    def finish_extract(self):
```
jaimergp: Same here.

conda/_private/extract.py
jaimergp: And maybe a test to ensure a new process that imports this doesn't end up with a polluted `sys.modules`.

jezdez: Added a worker-startup import constraint and a clean-interpreter test. It permits only the minimal conda modules and verifies conda_package_handling remains deferred.

conda/core/package_cache_data.py
jezdez: No. Restored the CPU-aware cap at three; this fallback still serves debug mode and transactions that contain a custom or unsupported extractor.

conda/core/path_actions.py
jezdez: Renamed it _prepare_extract.

conda/core/path_actions.py
jezdez: Renamed it _finish_extract.

jaimergp (approved):

dholth:
Hopefully this won't run into the same sandboxing constraints as the recent CUDA fix? Or I should allow the sandbox to hold that kind of lock. ([source code for processpoolexecutor](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/process.py))

PR #15975, opened 2026-04-27
Batch codesign calls for osx-arm64 prefix rewrites (B9c)

jezdez:
### Description

On osx-arm64, `update_prefix` signs every rewritten binary with `/usr/bin/codesign -s - -f <path>`. Phase-1 profiling of a W2 macOS install found 186 `subprocess.communicate` calls totaling 9.47 s, from per-package `codesign` process startup rather than bytecode compilation.

This PR queues codesign paths while transaction actions are verified, then flushes the batch through `batch_codesign_calls()` as `codesign -s - -f <path1> <path2> ...`. Direct `update_prefix()` callers still sign immediately. Each rewritten binary still gets the same ad-hoc signature, with fewer subprocess launches. Linux and Windows do not enter this path.

Stack interaction: B6 (#15973) now maps individual action verification directly across the shared verification executor and removes `_verify_individual_level`. After B6 lands, this PR must move the codesign queue to parent-visible transaction state before rebasing; a thread-local batch context does not propagate into executor workers.

### Measured mechanism

The profiled W2 install had 186 rewritten binaries. This PR reduces that work from 186 `codesign` subprocess invocations to one batched invocation while preserving the per-binary signature operation.

The earlier end-to-end timing was removed because it was taken on a superseded stack before the B6 interaction above was resolved. This PR does not claim a current standalone install-time speedup until it is rebased and remeasured.

### Checklist - did you ...

### DIFF
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -140,7 +140,8 @@ def _run_codesign(paths: list[str]) -> None:
     for i in range(0, len(paths), chunk_size):
         subprocess.run(
             ["/usr/bin/codesign", "-s", "-", "-f", *paths[i : i + chunk_size]],
-            capture_output=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
         )
 
 
--- a/conda/core/portability.py
+++ b/conda/core/portability.py
@@ -9,8 +9,10 @@ import re
 import struct
 import subprocess
 import threading
+from contextlib import contextmanager
 from logging import getLogger
 from os.path import basename, realpath
+from typing import TYPE_CHECKING
 
 from ..auxlib.ish import dals
 from ..base.constants import PREFIX_PLACEHOLDER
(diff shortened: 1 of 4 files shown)


conda/core/portability.py
```diff
@@ -117,12 +123,63 @@ def _update_prefix(original_data):
...
+
+    # macOS Darwin has ARG_MAX ~ 1 MB; per-path cost is typically
+    # 100-200 bytes of fully-qualified realpath. 500-path chunks stay
+    # well under the cap.
+    chunk_size = 500
+    for i in range(0, len(paths), chunk_size):
         subprocess.run(
-            ["/usr/bin/codesign", "-s", "-", "-f", realpath(path)], capture_output=True
+            ["/usr/bin/codesign", "-s", "-", "-f", *paths[i : i + chunk_size]],
+            capture_output=True,
```
jaimergp: We are not doing anything with the output, right? Not sure how much `codesign` speaks, but if it gets verbose, there's no point in storing it. Can we pipe it to `/dev/null`?

jaimergp (approved):

conda/core/portability.py
jezdez: Implemented in 20a9b4957. Both the direct and batched `codesign` calls now send unused stdout and stderr to `/dev/null`, with the focused portability test asserting both forms.

PR #15976, opened 2026-04-27
Adding shim for `conda.ng` module and `experimental=ng` setting value

### LINKED ISSUE #15940: Add `conda._ng` layout, `python -m conda.ng` entry point, and `experimental=ng` flag
### Checklist

### What is the idea?

Introduce `conda._ng` as the private internal package for the next-generation conda backend, expose `python -m conda.ng` as a convenience entry point, and gate the ng code path behind the existing `experimental` setting (`CONDA_EXPERIMENTAL=ng`).

### Why is this needed?

This is a set of prework:
- A private `conda._ng` namespace
- A `python -m conda.ng` entry point lets developers and testers opt in with a single invocation.
- Experimental option allows us to opt in during testing

### What should happen?

- Add `conda/_ng/` as the home for next-generation backend code. Add `conda/ng.py` as a module-level shim so `python -m conda.ng` works; the shim injects `ng` into `CONDA_EXPERIMENTAL` before delegating to the standard `conda` entry point.
- Register `ng` as a recognized value in `context.experimental`. In `do_call()`, check `"ng" in context.experimental` and route `create` and `install` to `conda._ng` when the flag is active; all other subcommands fall back to the classic `conda.cli` path.
- First iteration can be non-operational

### Additional Context

- A proof-of-concept lives on the `ng-common-cli` branch.
- This issue covers initial scaffolding only.

travishathaway:
### Description

This pull requests lays the foundation for future `conda.ng` (next generation) work.

fixes: #15940

### Checklist - did you ...


conda/ng.py
```diff
@@ -0,0 +1,53 @@
...
+guaranteed to be active.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+
+
+def _inject_ng_experimental() -> None:
```
jaimergp: This is going to be a pain for testing. Can we make it a context manager so it self cleans?

conda/ng.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Ensure ``ng`` is present in ``CONDA_EXPERIMENTAL`` before we start.
+    """
+    current = os.environ.get("CONDA_EXPERIMENTAL", "")
+    values = [v.strip() for v in current.split(",") if v.strip()]
+    if "ng" not in values:
+        values.append("ng")
+    os.environ["CONDA_EXPERIMENTAL"] = ",".join(values)
```
jaimergp: Why mess with the environment if we already have a `context` by now?

jaimergp:
xref https://github.com/conda/conda/pull/15813 for the original experiment

conda/ng.py
travishathaway: I was going back and forth on this myself. This is meant to instantiated by `main_subshell` so this environment variable would get picked up and override everything else. But, we do repeatedly call `context.__init__` so doing it once more here wouldn't hurt.

travishathaway:
Thanks for the suggestions @jaimergp.

Please check out the updates and let me know if you approve.

conda/_ng/cli/main_create.py
```diff
@@ -0,0 +1,36 @@
...
+        **Non-operational stub.**  This function currently delegates to the
+        classic ``conda.cli.main_create.execute`` implementation.  It exists
+        solely to prove that the routing infrastructure works end-to-end.
+        Replace or extend this body in subsequent iterations.
+    """
+    log.debug("conda._ng.cli.main_create.execute called (stub — delegating to classic)")
+
+    from conda.cli.main_create import execute as _classic_execute
+
+    return _classic_execute(args, parser)
```
jezdez, line 36: I think we're not doing ourselves a favor with naming this "classic" execute. It's the normal, regular, standard execute, and this here is the outlier.

conda/ng.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+conda.ng — convenience entry-point shim for the next-generation backend.
```
jezdez, line 4: I think this is wrong. For one, it marks this module as non-experimental, making it fall under the deprecation policy. Another is that it's also not explaining what it's doing, which is an antipattern. What is "ng"? 

If you're looking to make this an "experiment loader" let's just call it `conda.experiments`.

conda/ng.py
```diff
@@ -0,0 +1,77 @@
...
+        # Restore the context to its original state.
+        context.__init__(argparse_args=Namespace(experimental=original_experimental))
+
+
+def main() -> int:
+    """Shim entry point: inject the ng flag and delegate to conda.cli.main."""
+    from conda.cli.main import main as _main
+
+    with _inject_ng_experimental():
+        return _main()
```
jezdez, line 73: Sorry but no, this is clever, but we don't need to hack conda to get this into, we can just make the ability to switch experiments on and off into the main CLI/config system.

tests/_ng/test_ng.py
```diff
@@ -0,0 +1,240 @@
...
+from __future__ import annotations
+
+from argparse import Namespace
+from unittest.mock import MagicMock, patch
+
+from conda.base.context import context, reset_context
+
+# ---------------------------------------------------------------------------
+# conda.ng shim
+# ---------------------------------------------------------------------------
```
jezdez, line 14: Don't use section comments, that's an AI-ism that is code smell.

tests/_ng/test_ng.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class TestNgShim:
```
jezdez, line 17: Please continue to use function based tests like elsewhere in the test suite.

conda/cli/conda_argparse.py
```diff
@@ -182,9 +182,20 @@ def do_call(args: argparse.Namespace, parser: ArgumentParser):
...
         # func_name should always be 'execute'
-        module = import_module(module_name)
         command = module_name.split(".")[-1].replace("main_", "")
 
+        # When the ``ng`` experimental flag is active, route ``create`` and
+        # ``install`` through the next-generation backend instead of the
+        # classic implementation.  All other commands fall back to the
+        # classic path unchanged.
+        _ng_commands = {"create", "install"}
+        if "ng" in context.experimental and command in _ng_commands:
```
jezdez, line 192:
```suggestion
        if "cli-refresh" in context.experimental and command in _experimental_commands:
```

jezdez (changes requested):
I don't agree with the `_ng` name, it's a misnomer that we shouldn't introduce. I've left comments below. In short, the name is hiding what it does. Let's just call it what it is, an experiment. We can build the ability to create experiments into conda without hiding it behind short names.

conda/ng.py
travishathaway: Okay. @jaimergp was that one that suggested a contextmanager solution here.

Can you think of a better approach?

jezdez:
See https://github.com/conda/conda/issues/15940 for an updated tracking ticket for this PR

travishathaway:
I'm closing this pull request because the specifications have changed considerably since opening this one, and I would like to have a fresh conversation for the next round of work 🌷 🌻

PR #15977, opened 2026-04-27
Define public, private symbols as affecting deprecation policy

dholth:
### Description

Define public, private symbols in Python. Define private symbols as not being a "feature" requiring deprecation. Add advice for library users.

### Checklist - did you ...


docs/source/dev-guide/deprecations.md
```diff
@@ -2,7 +2,43 @@
 
 # Deprecations
 
-Conda abides by the Deprecation Schedule defined in [CEP-9][cep9]. To help make deprecations as much of a no-brainer as possible we provide several helper decorators and functions to facilitate the correct deprecation process.
+Conda abides by the Deprecation Schedule defined in [CEP-9][cep9].
+
+## What must be deprecated
+
+**Public** symbols must be deprecated before being removed.
```
jaimergp, line 9: It would be practical to define which ones are safe to depend on as a 3rd party:

- They are included in the `__all__` list.
- If there's no `__all__` list, they don't start with underscore, or they are not a child of a symbol that starts with underscore. In other words, their fully qualified symbol path does not contain leading underscores.

It may be tricky to identify non-`__all__` symbols whose name is not underscore-trailed, though. We should be careful with those.

docs/source/dev-guide/deprecations.md
```diff
@@ -2,7 +2,43 @@
...
+**Private** symbols are no longer considered "features" eligible for
+deprecation.[^1]
+
+[^1]: Private symbols introduced before 2026-05 may require deprecation per the
+    discretion of the conda maintainers.
+
+A private symbol is any symbol that starts with a single `_` character; any
+symbol that is inside a module that starts with a single `_` character; or any
+symbol that is not in a module's `__all__` list if it has such a list. This
+includes underscore-prefixed symbols in class instances.
```
jaimergp: What's the resolution for a symbol that's in `__all__` but in a `_private` module? Imagine this:

```python
# path: conda/_internal/main.py

from .other import something

__all__ = ["something"]
```

Can I depend on `something`? I assume we can't because we are within `_internal`, right? We may want to have `__all__` here for DX purposes, but not as a "this is public API" marker.

docs/source/dev-guide/deprecations.md
```diff
@@ -2,7 +2,43 @@
...
+
+Imported symbols (`import <package>` statements) are considered private.
+
+Type checkers should be able to check these rules.
+
+## Using conda as a library
+
+Use conda's public API. If you need a private symbol, pin your conda dependency
+to that specific conda release `conda==26.9.0`, and file an issue so we can
+provide a suitable public API.
```
jaimergp:
```suggestion
consider whether your use case needs a public API.
```

docs/source/dev-guide/deprecations.md
```diff
@@ -2,7 +2,43 @@
...
+
+### Re-exporting a private symbol to the public API
+
+[typing.python.org](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols)
+defines conventions for *re-exporting* symbols to the public API. Type checkers
+understand these rules. This system helps conda to expose a designed set of APIs
+to the library user.
+
+## Marking code for deprecation
```
jaimergp, line 44: In the community call where we talked about this, we found consensus to reduce the deprecation windows. This was not [captured](https://github.com/conda/governance/blob/34b7e858cc76c333ac00b5799563f3a5f89927c8/meetings/archive/20260218-conda-ecosystem.md) but I recall folks would be okay with halving the current windows? Or am I misremembering?

jezdez: Hm, can't be a consensus if not all maintainers were there, though. Since the CEP is applied by the conda maintainers and the build tools teams to their specific projects, it's something to propose outside this change.

jaimergp: Consensus within the attendees in that call, yes, sorry. As in "there was no opposition to the idea of reducing the deprecation windows in that call".

docs/source/dev-guide/deprecations.md
dholth: The "inside a private module" rule would apply. We can check if there's a better way to word it.

docs/source/dev-guide/deprecations.md
dholth: These rules are meant to be things that the editors are pretty good about pointing out. If they aren't, then we may need to make sure we have the right py.typed files and type definitions.

docs/source/dev-guide/deprecations.md
dholth: We also had a person who noticed that A had been deprecated as B and B had been deprecated as C. When it was time for them to update their code, they had to skip from A to C.
I would be OK with a shorter window.

dholth:
Related to #14895

ryanskeith (approved):
These look sound to me.

news/deprecations-update
```diff
@@ -0,0 +1,21 @@
...
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Define private symbols, modules (i.e. starts-with-underscore `_`) as not
+  requiring deprecation. Library users should avoid these symbols. See
+  [deprecations.md](https://github.com/conda/conda/blob/main/docs/source/dev-guide/deprecations.md).
```
kenodegard: Link to rendered docs instead?

```suggestion
  [deprecations.md](https://docs.conda.io/projects/conda/en/26.5.x/dev-guide/deprecations.html). (#15977)
```

kenodegard (approved):

jezdez (approved):

PR #15980, opened 2026-04-27
Optional py-rattler fast path for `PrefixGraph.__init__`

jezdez:
### Description

Draft PR, logically stacked on top of #15971. **Do not merge before #15971 lands** unless we intentionally want this PR to carry the B2 name-indexed fallback as well.

This PR adds a rattler-backed fast path inside `PrefixGraph.__init__` for the expensive dependency parsing and parent-matching part of graph construction.

The dependency story has changed since this branch was opened: `conda-rattler-solver` depends on `py-rattler >=0.25.0,<0.26.0a0`, and conda runtimes that include the rattler solver are therefore expected to have the `rattler` Python module available. That makes this less of an optional-dependency experiment and more of a potential large-prefix fast path using bindings that conda is already carrying.

Current implementation:

- Convert each conda `PrefixRecord` to a `rattler.PackageRecord` once per call, caching the converted view on the conda record where possible.
- Parse dependency strings with `rattler.MatchSpec`, using a per-call dedup cache for repeated dependency strings.
- Match candidates with `rattler.PackageRecord.matches` after narrowing candidates through the same by-name index introduced in #15971.
- Return the conda-shaped adjacency dict and `spec_matches`, then call `PrefixGraph._toposort()` so conda keeps its existing stable topological ordering and special-case edge handling.

The fallback still matters. Minimal/PyPI-style installs may not have the solver plugin dependency set, and rattler can still reject specs or records that conda accepts. If `rattler` is missing, cannot parse a spec, cannot convert a record, or hits an edge case, the fast path returns `None` and the caller falls back to the pure-Python name-indexed loop from #15971.

Full research context and the S19 / W5 benchmark writeups live in the research report: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#w5-conda-install-real-pkg-against-realistic-bench_big-end-to-end

### Numbers

Measured on `conda install -c conda-forge -y --dry-run requests` against a realistic synthetic prefix (exponential dep fan-out, version-constrained dep lines, mixed subdirs) using the fixture from https://github.com/jezdez/conda-tempo/blob/main/bench/seed_big_prefix.py:

| Prefix size (N) | #15971 only | with rattler fast path | speedup |
|---:|---:|---:|---:|
| 1,000 | 1.75 s | 1.90 s | 0.92x |
| 5,000 | 4.71 s | 2.95 s | 1.6x |
| 10,000 | 12.60 s | 4.11 s | 3.1x |
| 50,000 | 628.8 s | 21.3 s | **29.5x** |

At N=50,000, installing anything into a long-lived prefix drops from roughly 10 minutes to 21 seconds in the original W5 experiment. The current branch has since been adjusted to preserve conda's `_toposort()` ordering after the rattler-backed adjacency build, so these numbers should be rerun before the PR leaves draft.

The small-prefix result matters more now that `rattler` is expected to be available in normal conda runtimes: at N=1,000 the fast path regressed slightly in W5, and CodSpeed reported regressions on small update benchmarks. The July 2026 review conclusion is that the current always-try-rattler shape should not leave draft. The next shippable variant should gate the rattler path before any record conversion, starting conservatively at `len(records) >= 10_000`; N=5,000 can be reconsidered only after fresh W5 and CodSpeed-style reruns show that the current `_toposort()`-preserving implementation is still safely positive there.

### Remaining work before ready for review

1. **Threshold / activation policy**. Implement a pre-conversion large-prefix gate, initially `len(records) >= 10_000`, so common small-prefix operations continue through the #15971 pure-Python path. Revisit a lower 5,000-record gate only with fresh benchmark evidence.
2. **Benchmark rerun**. Rerun W5 with the current implementation that preserves conda's `_toposort()` ordering, plus small-prefix/CodSpeed-style workloads that previously showed regressions.
3. **Grammar coverage**. rattler's `MatchSpec` parser is stricter than conda's; it rejects a handful of bracket-key syntaxes that conda silently accepts (for example, `python[extras=['dev','test']]`). The fallback handles this transparently, but a compatibility sweep across real conda-forge / conda-channel specs would build confidence.
4. **Test coverage**. Add CI tests that exercise both paths and assert equivalent graph output, plus fallback behavior when rattler is unavailable or rejects a spec.

Tracking epic: conda/conda#15969.

### Checklist - did you ...

- [ ] Add / update necessary tests? Manual correctness checks only at this stage; CI test pending, see remaining work item 4.
- [x] Add / update outdated documentation? No user-facing docs change; the feature is invisible to callers of `PrefixGraph`.

### DIFF
--- a/conda/models/_prefix_graph_rattler.py
+++ b/conda/models/_prefix_graph_rattler.py
@@ -1,0 +1,195 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Optional py-rattler fast path for PrefixGraph construction.
+
+Scope: this module is imported lazily from
+``conda.models.prefix_graph.PrefixGraph.__init__``. If py-rattler is
+not installed, ``try_build_graph`` returns ``None`` and the caller
+falls back to the pure-Python name-indexed path from #15971.
+
+Strategy:
+
+1. Convert the conda ``PrefixRecord`` inputs to ``rattler.PackageRecord``
+   once per call, caching the view on the conda record so repeat
+   PrefixGraph calls within one process skip the reconversion.
+2. Build adjacency using ``rattler.MatchSpec`` for dep parsing and
+   ``rattler.PackageRecord.matches`` for the match test. Both are
+   Rust-backed; parse is ~4x faster, match is ~5x faster than conda's
+   native implementations (measured in S18).
+3. Cache parsed ``rattler.MatchSpec`` by dep string within a single
+   call: real prefixes have significant dep-string repetition.
+4. Sort topologically with ``rattler.PackageRecord.sort_topologically``
+   on the rattler record list, then re-key the adjacency dict into
+   that order before returning.
+5. Spec matching (the ``specs`` argument of PrefixGraph) uses conda's
+   ``MatchSpec`` because callers hand us conda ``MatchSpec`` instances
+   built from sources we don't control, which may use conda grammar
+   rattler rejects.
+
+Returns ``(graph, spec_matches)`` in the same shape the pure-Python
+loop produces. Returns ``None`` for any error so conda's fallback
+keeps operating correctly; the fast path is strictly additive.
+"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+try:
+    import rattler as _rattler
+except ImportError:
+    _rattler = None
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from .match_spec import MatchSpec
+    from .records import PrefixRecord
+
+_RATTLER_VIEW_ATTR = "_conda_prefix_graph_rattler_view"
+
+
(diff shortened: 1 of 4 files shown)


jezdez:
Closing this as a research result rather than a merge candidate.

The W5/S19 numbers show that a rattler-backed `PrefixGraph` can be very fast for truly large prefixes, but this PR is the wrong shape to maintain in conda: it makes `rattler` a quasi-optional dependency, adds a conda-record to rattler-record mapping layer, needs fallback behavior for grammar mismatches, and would require dual-path coverage for ordering, graph shape, cycles, noarch/Windows edge cases, and activation policy.

A threshold could reduce the small-prefix regression risk, but it would not remove that maintenance burden. The Track B path should keep #15971 as the shippable `PrefixGraph` improvement and treat rattler-backed records/specs as a wider future architecture question, not a narrow optional shim inside `PrefixGraph.__init__`.

Tracking epic: conda/conda#15969.

PR #15981, opened 2026-04-28
Fix `Environment.extrapolate` virtual packages for target platform

### LINKED ISSUE #15919: Lockfiles fail to export when the Python version is pinned
### Checklist

### What happened?

With `conda-lockfiles` instead, exporting a multi-platform lockfile fails if the python version is pinned in the environment, but succeeds if it is not pinned.

For example, an environment created with this environment.yaml:

```
name: python-pinned
channels:
  - conda-forge
dependencies:
  - python=3.13
  - numpy
```

1. `conda create --file environment.yaml`
2. `conda activate python-pinned`
3. `conda export --file conda-lock.yml --platform win-64 --platform linux-64 --platform osx-arm64`
4. See `LibMambaUnsatisfiableError`

```
~/Projects $ conda export --file conda-lock.yml --platform win-64 --platform linux-64 --platform osx-arm64                 (python-pinned) 
Channels:
 - conda-forge
Platform: osx-arm64
Solving environment: done
Channels:
 - conda-forge
Platform: osx-arm64
Solving environment: failed

LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides __glibc >=2.17,<3.0.a0 needed by python-3.13.0-h6355ac2_0_cp313t

Could not solve for environment specs
The following package could not be installed
└─ python =3.13 * is not installable because it requires
   └─ __glibc >=2.17,<3.0.a0 *, which is missing on the system.
```

### Conda Info

```shell
active environment : python-pinned
    active env location : /Users/dyeaw/minicondaprotected/envs/python-pinned
            shell level : 2
       user config file : /Users/dyeaw/.condarc
 populated config files : /Users/dyeaw/minicondaprotected/.condarc
                          /Users/dyeaw/minicondaprotected/condarc.d/anaconda-auth.yml
                          /Users/dyeaw/.condarc
          conda version : 26.3.2
         python version : 3.13.11.final.0
                 solver : libmamba (default)
                          __conda=26.3.2=0
                          __osx=26.3=0
                          __unix=0=0
       base environment : /Users/dyeaw/minicondaprotected  (writable)
      conda av data dir : /Users/dyeaw/minicondaprotected/etc/conda
  conda av metadata url : None
           channel URLs : 
          package cache : /Users/dyeaw/minicondaprotected/pkgs
                          /Users/dyeaw/.conda/pkgs
       envs directories : /Users/dyeaw/minicondaprotected/envs
                          /Users/dyeaw/.conda/envs
    temporary directory : /var/folders/7v/l5vtzkrj32l6b_nxxh821_740000gp/T
               platform : osx-arm64
             user-agent : conda/26.3.2 requests/2.33.1 CPython/3.13.11 Darwin/25.3.0 OSX/26.3 solver/libmamba conda-libmamba-solver/26.3.0 libmambapy/2.3.2 aau/0.7.6 c/. s/. e/. a/. o/.
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> /Users/dyeaw/minicondaprotected/.condarc <==
default_activation_env: python-3.13
channels:
  - defaults

==> /Users/dyeaw/minicondaprotected/condarc.d/anaconda-auth.yml <==
channel_settings:
  - channel: https://repo.anaconda.cloud/*
    auth: anaconda-auth

==> /Users/dyeaw/.condarc <==
channels:
  - defaults
  - conda-forge
show_channel_urls: True
always_yes: True
solver: libmamba
```

### Conda list

```shell
# packages in environment at /Users/dyeaw/minicondaprotected/envs/python-pinned:
#
# Name                     Version          Build                Channel
_openmp_mutex              4.5              7_kmp_llvm           conda-forge
bzip2                      1.0.8            hd037594_9           conda-forge
ca-certificates            2026.2.25        hbd8a1cb_0           conda-forge
libblas                    3.11.0           6_h51639a9_openblas  conda-forge
libcblas                   3.11.0           6_hb0561ab_openblas  conda-forge
libcxx                     22.1.3           h55c6f16_0           conda-forge
libexpat                   2.7.5            hf6b4638_0           conda-forge
libffi                     3.5.2            hcf2aa1b_0           conda-forge
libgcc                     15.2.0           hcbb3090_18          conda-forge
libgfortran                15.2.0           h07b0088_18          conda-forge
libgfortran5               15.2.0           hdae7583_18          conda-forge
liblapack                  3.11.0           6_hd9741b5_openblas  conda-forge
liblzma                    5.8.3            h8088a28_0           conda-forge
libmpdec                   4.0.0            h84a0fba_1           conda-forge
libopenblas                0.3.32           openmp_he657e61_0    conda-forge
libsqlite                  3.53.0           h1b79a29_0           conda-forge
libzlib                    1.3.2            h8088a28_2           conda-forge
llvm-openmp                22.1.3           hc7d1edf_0           conda-forge
ncurses                    6.5              h5e97a16_3           conda-forge
numpy                      2.4.3            py313he4a34aa_0      conda-forge
openssl                    3.6.2            hd24854e_0           conda-forge
python                     3.13.13          h20e6be0_100_cp313   conda-forge
python_abi                 3.13             8_cp313              conda-forge
readline                   8.3              h46df422_0           conda-forge
tk                         8.6.13           h010d191_3           conda-forge
tzdata                     2025c            hc9c84f9_1           conda-forge
```

### Additional Context

_No response_

jezdez:
### Description

Fixes #15919.

Multi-platform `conda export` of a lockfile fails with `nothing provides __glibc >=2.17,<3.0.a0` when an environment pins `python=3.13` (or any other spec that transitively depends on a target-platform-only virtual package).

**Root cause:** `Environment.extrapolate(platform)` correctly passes `subdirs=(platform, "noarch")` to the solver but never updates `context.subdir`. The `conda_virtual_packages` plugins (`linux.py`, `osx.py`, `windows.py`) gate on `context.subdir.startswith("...")` and therefore don't emit target-platform virtual packages when running on a different host.

**Fix:** wrap the solve in `context._override("_subdir", platform)` so the virtual_packages plugins emit virtual packages for the target during the solve. `context.subdir` is restored on both normal and exceptional exits.

**Drive-by fix to `Context._override`:** the previous implementation used `setattr(self, key, value)` which permanently shadowed `ParameterLoader` (and other non-data descriptor) attributes in `__dict__`. After exit, `reset_context()` could no longer pick up `CONDA_*` env-var changes for the overridden parameter, breaking tests like `tests/plugins/test_virtual_packages.py::test_subdir_override` whenever `extrapolate` had run earlier in the session. Reimplemented to manipulate `__dict__` directly (sentinel-based capture/restore) so the descriptor is reinstated on exit. The helper's docstring already flagged this as suspect.

**Known follow-up (out of scope):** this PR resolves the "nothing provides `__glibc`" failure mode. A second-order issue remains: when extrapolating from a non-Linux host to `linux-*`, `linux.py` falls back to emitting `__glibc==0=0` because it can't introspect the target. Strict pins like `python=3.13` (which require `__glibc>=2.17`) still need a `CONDA_OVERRIDE_GLIBC` (the integration test sets one). Tracking separately — likely needs sensible default `CONDA_OVERRIDE_GLIBC`/`CONDA_OVERRIDE_OSX` for cross-platform extrapolation.

### Tests

- Unit (parametrized over `linux-64` / `osx-arm64` / `win-64`):
  - `test_extrapolate_uses_target_subdir`: verifies `context.subdir` is the target during the solve and the matching virtual package (`__glibc` / `__osx` / `__win`) is emitted, then restored on exit.
  - `test_extrapolate_restores_subdir_on_solver_error`: verifies exception safety of the override.
- Integration (parametrized; slow): `test_extrapolate_python_pinned` reproduces the original bug end-to-end with `python=3.13`, applying the documented `CONDA_OVERRIDE_*` per target.

### Checklist - did you ...

- [x] ~Add / update outdated documentation?~ (no user-facing API change)


tests/models/test_environment.py
```diff
@@ -858,6 +858,163 @@ def test_from_cli_environment_inject_default_packages_override_file(
...
+    cannot be introspected from a foreign host, so ``CONDA_OVERRIDE_*``
+    is the supported way to drive the solve. Without the fix, the solve
+    fails earlier with ``nothing provides __glibc`` because the linux
+    plugin never runs (``context.subdir`` stays on the host platform).
+    """
+    if target == context.subdir:
+        pytest.skip("target equals host; extrapolate short-circuits")
+    for var, value in _CROSS_PLATFORM_OVERRIDES[target].items():
+        monkeypatch.setenv(var, value)
+    with tmp_env("python=3.13") as prefix:
```
kenodegard: Do we have to use python here? Or can we create a mock package mirroring the virtual package requirements?

tests/models/test_environment.py
```diff
@@ -858,6 +858,163 @@ def test_from_cli_environment_inject_default_packages_override_file(
...
+    env = Environment(
+        prefix="/fake/prefix",
+        platform=original_subdir,
+        config=EnvironmentConfig.from_context(),
+    )
+
+    result = env.extrapolate(target)
+
+    assert result.platform == target
+    assert captured, "Solver was not invoked"
```
kenodegard: We should only expect one solve right?


```suggestion
    assert len(captured) == 1
```

tests/models/test_environment.py
jezdez: Good call, done in 9acda8518 and 720f6ef14.

I went with a new `virtualdep-package` in the existing `tests/data/test-recipes/{linux,osx,win}-fake` subdirs rather than modifying `arch-package`, because `test_create_env_different_platform` does a real `--dry-run` install against `arch-package` and would have started implicitly requiring CI runners to have `__glibc >=2.28`. Adding a new package next to `arch-package` keeps those tests untouched.

While in there I added a second parametrize arm (`version="2.0"`) with a versioned virtual dep (`__glibc >=2.28` / `__osx >=11.0`) and `CONDA_OVERRIDE_*` envvars. The unversioned arm catches the main failure mode this PR fixes (target virtual package not emitted at all); the versioned arm documents the override path users on pins like `python=3.13` still need until the target-introspection follow-up lands. Windows doesn't get a versioned variant because `windows.py` emits `__win` with no meaningful version.

One small gotcha worth flagging: patching only `conda.base.constants.KNOWN_SUBDIRS` (which is what `test_create_env_different_platform` does) doesn't actually change what the solver accepts, because `conda.base.context` imports `KNOWN_SUBDIRS` once at module load and keeps its own reference to the original value. The existing test gets away with this because it sets `CONDA_SUBDIR=<fake>`, which adds the fake subdir to `context.subdirs`, which `known_subdirs` then merges in. My test patches `conda.base.context.KNOWN_SUBDIRS` directly so the fake target subdir is accepted.

tests/models/test_environment.py
jezdez: Applied in ac5436808 — thanks for the direct fixup. Extrapolate now calls the solver exactly once per target, and the captured call list is asserted to contain a single entry.

dholth: What does it mean to extrapolate an environment? The docstring for extrapolate() only says "Given the current environment, extrapolate the environment for the given platform."

kenodegard: > What does it mean to extrapolate an environment? The docstring for extrapolate() only says "Given the current environment, extrapolate the environment for the given platform."

To solve for a comparable environment on a different platform.

kenodegard (approved):

PR #15982, opened 2026-04-28
Add utility functions for working with `TERM=dumb`, `NO_COLOR` and `FORCE_COLOR` environment variables

### LINKED ISSUE #15943: Support NO_COLOR, FORCE_COLOR, TERM=dumb, and centralize CLI output modes
### Checklist

### What is the idea?

When building the ng code path, ensure it fully supports conda's existing CLI output modes from the start: `NO_COLOR`, `FORCE_COLOR`, `TERM=dumb`, `--verbose`, `--quiet`, `--json`, and `--dry-run`, with all output-mode logic centralized in the reporter abstraction.

### Why is this needed?

- Conda does not currently read standard terminal environment variables (`NO_COLOR`, `FORCE_COLOR`, `TERM=dumb`), leaving users and CI systems with no reliable way to control output styling.
- TTY detection is duplicated across many locations in the reporter backend rather than handled in one place.
- Establishing consistent output-mode behavior before more ng subcommands are added avoids a larger audit and retrofit later.

### What should happen?

- Respect the standard terminal output environment variables: `NO_COLOR` (suppress all ANSI), `FORCE_COLOR` (enable color even in non-TTY contexts such as CI pipelines), and `TERM=dumb` (plain text only). Rich supports all three natively when the `Console` is constructed correctly.
- Centralize TTY detection. The existing reporter backend checks `sys.stdout.isatty()` in many scattered locations. This logic belongs in one place.
- Suppress spinner and progress bar animations automatically in non-TTY contexts, when `NO_COLOR` is set, and when `TERM=dumb`.
- Handle `quiet`, `json`, `verbose`, and `dry-run` through the reporter abstraction rather than with inline checks in each ng CLI module.

### Additional Context

- The reporter backend plugin system (`conda/plugins/reporter_backends/`) is the right place to centralize these mode checks.
- A proof-of-concept for the ng path lives on the `ng-common-cli` branch.

travishathaway:
### Description

fixes: #15943

### Checklist - did you ...


conda/_ng/cli/common.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    Serialize Rich renderables (or plain strings) to plain text.
+
+    Uses :func:`create_console` so that ``NO_COLOR`` / ``FORCE_COLOR`` /
+    ``TERM=dumb`` are all respected automatically.
+    """
+    console = create_console(record=True)
+    with console.capture() as capture:
+        console.print(*objects)
+    return capture.get()
```
jezdez: Could express this as a `Console` subclass — fewer hidden imports, downstream code can `isinstance()`-check it, and `__init__` is the natural place for the env-var dispatch. (Folds in the `soft_wrap` default and absorbs `export_rich_renderable` as a method, so the unused module-level helper goes away.)

```suggestion
class CondaConsole(Console):
    """A :class:`rich.console.Console` that honours ``NO_COLOR``,
    ``FORCE_COLOR`` and ``TERM=dumb`` / ``TERM=unknown``.

    The following variables are honoured (in precedence order):

    * ``TERM=dumb`` or ``TERM=unknown`` – Rich disables all ANSI escape codes
      natively when it reads ``TERM`` from the environment; we don't override.
    * ``NO_COLOR`` (https://no-color.org/) – disables color only; bold and
      underline may still appear.
    * ``FORCE_COLOR`` – enables Rich markup even when stdout is not a TTY.
    * TTY detection via ``sys.stdout.isatty()`` – the default Rich behaviour.

    In non-TTY contexts the ``width`` is set to a very large value so piped /
    CI output isn't line-wrapped.
    """

    def __init__(self, *args, soft_wrap: bool = True, **kwargs) -> None:
        from conda.common.io import force_color, is_tty, no_color, term_dumb

        if "force_terminal" not in kwargs:
            if term_dumb():
                pass  # Rich disables escape codes natively for TERM=dumb
            elif no_color():
                kwargs.setdefault("no_color", True)
            elif force_color():
                kwargs.setdefault("force_terminal", True)

        if not is_tty():
            kwargs.setdefault("width", 100_000)

        super().__init__(*args, soft_wrap=soft_wrap, **kwargs)

    def to_text(self, *objects) -> str:
        """Render Rich renderables (or plain strings) to a plain string."""
        with self.capture() as capture:
            self.print(*objects)
        return capture.get()
```

`Console.capture()` doesn't require `record=True`, so the existing `record=True` argument in the original helper can be dropped. Pick any name you prefer for the method (`to_text`, `render_text`, `as_string`, …) — and there are no callers of `export_rich_renderable` outside this PR, so the name change is free.

conda/common/io.py
```diff
@@ -40,6 +40,69 @@
...
+def term_dumb() -> bool:
+    """
+    Return True when the terminal is known to be incapable of ANSI escape codes.
+
+    ``TERM=dumb`` (and ``TERM=unknown``) signals that the terminal cannot
+    handle *any* escape sequences — not just colour.  This suppresses all
+    animations (spinners, progress bars) as well as colour output.
+    """
+    return os.environ.get("TERM") in ("dumb", "unknown")
+
```
jezdez: Calling `TERM=dumb` a "standard" (in the docstring narrative below) is a stretch — it's a long-standing terminfo convention (see `infocmp dumb`, GNU readline, git, less, …) but not a published standard the way `NO_COLOR` is. `TERM=unknown` is much rarer; happy to keep it as a defensive fallback, just don't lump them together.

```suggestion
def term_dumb() -> bool:
    """
    Return True when ``TERM`` indicates a non-capable terminal.

    ``TERM=dumb`` is the widely-followed terminfo convention for terminals
    that cannot render any ANSI escape sequences (color, bold, animations).
    ``TERM=unknown`` is treated the same way as a defensive fallback.
    """
    return os.environ.get("TERM") in ("dumb", "unknown")
```

conda/plugins/reporter_backends/console.py
```diff
@@ -240,18 +240,26 @@ def progress_bar(
...
+        Determines whether to return a Spinner or QuietSpinner.
+
+        Animations are suppressed under the same conditions as
+        :meth:`progress_bar`.
         """
-        if context.quiet:
+        if context.quiet or not is_tty() or term_dumb():
             return QuietSpinner(message, fail_message)
         else:
             return Spinner(message, fail_message)
```
jezdez, line 271: The `context.quiet or not is_tty() or term_dumb()` check is duplicated; pull it into a helper so both methods stay in sync.

```suggestion
        **kwargs,
    ) -> ProgressBarBase:
        """
        Determines whether to return a TQDMProgressBar or QuietProgressBar.

        Animations are suppressed when:
        * ``context.quiet`` is set,
        * the output is not a TTY (e.g. piped to a file), or
        * ``TERM=dumb`` or ``TERM=unknown`` is set.
        """
        if self.animations_disabled:
            return QuietProgressBar(description, **kwargs)
        else:
            return TQDMProgressBar(description, **kwargs)

    @property
    def animations_disabled(self) -> bool:
        """True when progress bars/spinners should be suppressed."""
        return context.quiet or not is_tty() or term_dumb()

    def spinner(self, message: str, fail_message: str = "failed\n") -> SpinnerBase:
        """
        Determines whether to return a Spinner or QuietSpinner.

        Animations are suppressed under the same conditions as
        :meth:`progress_bar`.
        """
        if self.animations_disabled:
            return QuietSpinner(message, fail_message)
        else:
            return Spinner(message, fail_message)
```

tests/common/test_io.py
```diff
@@ -123,9 +128,96 @@ def test_thread_limited_executor_handles_thread_limit(
...
+            else:
                 _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
-        else:
-            _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
+
+
+# ---------------------------------------------------------------------------
+# Output-mode helpers (NO_COLOR, FORCE_COLOR, TERM=dumb)
+# ---------------------------------------------------------------------------
+
```
jezdez: We don't use section dividers elsewhere in the test suite — please drop it. (The matching banner in `tests/plugins/reporter_backends/test_console.py` around L246–L248 should also go.)

```suggestion

```

tests/common/test_io.py
```diff
@@ -123,9 +128,96 @@ def test_thread_limited_executor_handles_thread_limit(
...
+        monkeypatch.delenv("NO_COLOR", raising=False)
+        monkeypatch.delenv("FORCE_COLOR", raising=False)
+        with patch("conda.common.io.is_tty", return_value=True):
+            assert should_use_color() is True
+
+    def test_non_tty_disables_color(self, monkeypatch):
+        monkeypatch.delenv("NO_COLOR", raising=False)
+        monkeypatch.delenv("FORCE_COLOR", raising=False)
+        with patch("conda.common.io.is_tty", return_value=False):
+            assert should_use_color() is False
```
jezdez: Let's keep these as module-level functions and parametrize the inputs — that's the convention in the rest of `tests/common/`. Also makes the `from unittest.mock import patch` import unnecessary (use `monkeypatch.setattr` instead).

```suggestion
@pytest.mark.parametrize(
    "term,expected",
    [("dumb", True), ("unknown", True), ("xterm-256color", False)],
)
def test_term_dumb_with_term(monkeypatch, term, expected):
    monkeypatch.setenv("TERM", term)
    assert term_dumb() is expected


def test_term_dumb_without_term(monkeypatch):
    monkeypatch.delenv("TERM", raising=False)
    assert term_dumb() is False


@pytest.mark.parametrize("value", ["", "1"])
def test_no_color_when_env_var_set(monkeypatch, value):
    monkeypatch.setenv("NO_COLOR", value)
    assert no_color() is True


def test_no_color_when_term_dumb(monkeypatch):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.setenv("TERM", "dumb")
    assert no_color() is True


def test_no_color_false_without_env_vars(monkeypatch):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.delenv("TERM", raising=False)
    assert no_color() is False


def test_no_color_false_with_non_dumb_term(monkeypatch):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.setenv("TERM", "xterm-256color")
    assert no_color() is False


@pytest.mark.parametrize("value", ["1", ""])
def test_force_color_when_env_var_set(monkeypatch, value):
    monkeypatch.setenv("FORCE_COLOR", value)
    assert force_color() is True


def test_force_color_false_without_env_var(monkeypatch):
    monkeypatch.delenv("FORCE_COLOR", raising=False)
    assert force_color() is False


def test_should_use_color_no_color_takes_precedence(monkeypatch):
    monkeypatch.setenv("NO_COLOR", "1")
    monkeypatch.setenv("FORCE_COLOR", "1")
    assert should_use_color() is False


@pytest.mark.parametrize(
    "is_tty_value,expected",
    [(False, True), (True, True)],
)
def test_should_use_color_force_color_or_tty(monkeypatch, is_tty_value, expected):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.setenv("FORCE_COLOR", "1")
    monkeypatch.setattr("conda.common.io.is_tty", lambda: is_tty_value)
    assert should_use_color() is expected


@pytest.mark.parametrize(
    "is_tty_value,expected",
    [(True, True), (False, False)],
)
def test_should_use_color_tty_only(monkeypatch, is_tty_value, expected):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.delenv("FORCE_COLOR", raising=False)
    monkeypatch.setattr("conda.common.io.is_tty", lambda: is_tty_value)
    assert should_use_color() is expected
```

tests/plugins/reporter_backends/test_console.py
```diff
@@ -10,16 +10,27 @@
...
     """
+    # Pretend we are in a TTY so that progress_bar returns TQDMProgressBar
+    mocker.patch(
+        "conda.plugins.reporter_backends.console.is_tty",
+        return_value=True,
+    )
+    mocker.patch(
+        "conda.plugins.reporter_backends.console.term_dumb",
+        return_value=False,
+    )
```
jezdez: We already get `monkeypatch` from pytest for free; no need to pull in `pytest-mock`'s `mocker` to stub two boolean returns. (Limited to the new patches added by this PR; the pre-existing `mocker.patch` calls elsewhere in this file are out of scope.)

```suggestion
def test_console_reporter_renderer(monkeypatch):
    """
    Tests the ``ConsoleReporterRenderer`` class
    """
    monkeypatch.setattr(
        "conda.plugins.reporter_backends.console.is_tty", lambda: True
    )
    monkeypatch.setattr(
        "conda.plugins.reporter_backends.console.term_dumb", lambda: False
    )
```

conda/common/io.py
```diff
@@ -40,6 +40,69 @@
 IS_INTERACTIVE = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
 
+def is_tty() -> bool:
```
jezdez: Should we move these helper functions into a new `conda.common.terminal` module or similar?

jezdez (changes requested):
I think this needs more work to fit in with the rest of the conda code base

conda/common/io.py
travishathaway: I like that idea. I'll move these over to a new module.

conda/_ng/cli/common.py
travishathaway: @jezdez, I've decided to just remove this addition from this pull request. I don't believe we actually need it right now.

travishathaway:
@jezdez,

Do you think this pull request would be a good opportunity to deprecate `IS_INTERACTIVE` in favor of `conda.common.terminal.is_tty`? Seems like the right call to make.

conda/common/terminal.py
kenodegard, line 1: Should these helpers be cached?

travishathaway: I did some rough profiling and here's what I found:

Without `@cache`:
```
conda.common.terminal — call counts during `conda create`
function              spy calls  cprofile calls  tot time (ms)  cum time (ms)
is_tty                        7               7         0.008          0.169
```

With `@cache`:
```
conda.common.terminal — call counts during `conda create`
function              spy calls  cprofile calls  tot time (ms)  cum time (ms)
is_tty                        7               1         0.002          0.003
```

So, I think that's a pretty compelling case to use the `@cache` decorator here.

conda/common/terminal.py
```diff
@@ -0,0 +1,77 @@
...
+
+@cache
+def is_tty() -> bool:
+    """Return True if both stdin and stdout are connected to a TTY."""
+    return (
+        hasattr(sys.stdout, "isatty")
+        and sys.stdout.isatty()
+        and hasattr(sys.stdin, "isatty")
+        and sys.stdin.isatty()
+    )
```
jezdez: Could we keep this helper stdout-focused? The call sites here are deciding whether output can render progress/spinners, and this also preserves the old `IS_INTERACTIVE` behavior. If we later need "both stdin and stdout are TTYs" for prompts, I think that can be a separate helper with a more specific name.

```suggestion
def is_tty() -> bool:
    """Return True if stdout is connected to a TTY."""
    return hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
```

tests/common/test_terminal.py
```diff
@@ -0,0 +1,110 @@
...
+    assert should_use_color() is False
+
+
+@pytest.mark.parametrize(
+    "is_tty_value,expected",
+    [(False, True), (True, True)],
+)
+def test_should_use_color_force_color_or_tty(monkeypatch, is_tty_value, expected):
+    monkeypatch.delenv("NO_COLOR", raising=False)
+    monkeypatch.setenv("FORCE_COLOR", "1")
```
jezdez: Tiny test isolation thing: in a `TERM=dumb` shell, `no_color()` wins before `FORCE_COLOR`, so this expectation depends on the parent environment unless we set `TERM` here.

```suggestion
    monkeypatch.setenv("TERM", "xterm-256color")
    monkeypatch.setenv("FORCE_COLOR", "1")
```

tests/common/test_terminal.py
```diff
@@ -0,0 +1,110 @@
...
+    assert should_use_color() is expected
+
+
+@pytest.mark.parametrize(
+    "is_tty_value,expected",
+    [(True, True), (False, False)],
+)
+def test_should_use_color_tty_only(monkeypatch, is_tty_value, expected):
+    monkeypatch.delenv("NO_COLOR", raising=False)
+    monkeypatch.delenv("FORCE_COLOR", raising=False)
```
jezdez: Same here: this test is about the TTY fallback, so it should pin `TERM` to a non-dumb value.

```suggestion
    monkeypatch.delenv("FORCE_COLOR", raising=False)
    monkeypatch.setenv("TERM", "xterm-256color")
```

tests/common/test_io.py
```diff
@@ -123,9 +123,9 @@ def test_thread_limited_executor_handles_thread_limit(
...
 
-    with thread_class() as executor:
-        if should_fail:
-            with pytest.raises(RuntimeError):
+        with thread_class() as executor:
+            if should_fail:
+                with pytest.raises(RuntimeError):
+                    _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
+            else:
                 _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
```
jezdez: Looks like this block got indented under `should_fail`, so the `ThreadLimitedThreadPoolExecutor, False` case returns without exercising the executor. Moving the context manager back out keeps the original coverage.

```suggestion
    with thread_class() as executor:
        if should_fail:
            with pytest.raises(RuntimeError):
                _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
        else:
            _ = [executor.submit(time.sleep, 0.1) for _ in range(jobs)]
```

jezdez (commented):
Thanks Travis, this direction makes sense to me. I left a few small inline suggestions, mostly to keep the behavior/test surface tight. The main thing I'd simplify is keeping the TTY helper focused on stdout for output decisions; the rest is just test isolation/coverage cleanup.

conda/common/terminal.py
```diff
@@ -0,0 +1,72 @@
...
+    1. ``NO_COLOR`` set or ``TERM=dumb``/``TERM=unknown`` → no color.
+    2. ``FORCE_COLOR`` set → color even in non-TTY.
+    3. stdout is a TTY → color.
+    4. Otherwise → no color.
+    """
+    if no_color():
+        return False
+    if force_color():
+        return True
+    return is_tty()
```
jezdez, line 66: I'm not sure caching all of these helpers is worth the hidden invalidation behavior. These functions read mutable process state (`os.environ`, `sys.stdout`), and the tests already need to clear every cache when monkeypatching env vars.

Given the measured cost is tiny, I'd lean toward keeping the env-derived helpers live. If we do keep caching, I think we should add one explicit public invalidation helper instead of having tests reach into each function's `.cache_clear()`.

```suggestion
import os
import sys


def is_tty() -> bool:
    """Return True if stdout is connected to a TTY."""
    return hasattr(sys.stdout, "isatty") and sys.stdout.isatty()


def term_dumb() -> bool:
    """
    Return True when ``TERM`` indicates a non-capable terminal.

    ``TERM=dumb`` is the widely-followed terminfo convention for terminals
    that cannot render any ANSI escape sequences (color, bold, animations).
    ``TERM=unknown`` is treated the same way as a defensive fallback.
    """
    return os.environ.get("TERM") in ("dumb", "unknown")


def no_color() -> bool:
    """
    Return True when *color* output should be suppressed.

    Respects the ``NO_COLOR`` standard (https://no-color.org/): color is
    suppressed when the ``NO_COLOR`` environment variable is set to any value.
    Non-color ANSI formatting (bold, underline, etc.) may still be used.
    """
    return "NO_COLOR" in os.environ


def force_color() -> bool:
    """
    Return True when color output should be forced even in non-TTY contexts.

    Respects the ``FORCE_COLOR`` environment variable when it is set to a
    non-empty value.
    """
    return bool(os.environ.get("FORCE_COLOR"))


def should_use_color() -> bool:
    """
    Determine whether ANSI color output should be produced.

    The precedence order (highest to lowest) is:

    1. ``NO_COLOR`` set -> no color.
    2. non-empty ``FORCE_COLOR`` set -> color even in non-TTY.
    3. ``TERM=dumb``/``TERM=unknown`` -> no color.
    4. stdout is a TTY -> color.
    5. Otherwise -> no color.
    """
    if no_color():
        return False
    if force_color():
        return True
    if term_dumb():
        return False
    return is_tty()
```

tests/common/test_terminal.py
```diff
@@ -0,0 +1,112 @@
...
+    monkeypatch.setenv("TERM", "xterm-256color")
+    monkeypatch.setattr("conda.common.terminal.is_tty", lambda: is_tty_value)
+    assert should_use_color() is expected
```
jezdez, line 90: If we keep the helpers uncached, the cache-clearing fixture can go away and the `FORCE_COLOR` / `TERM=dumb` expectations can live in the combined `should_use_color()` cases.

```suggestion
from conda.common.terminal import (
    force_color,
    no_color,
    should_use_color,
    term_dumb,
)


@pytest.mark.parametrize(
    "term,expected",
    [("dumb", True), ("unknown", True), ("xterm-256color", False)],
)
def test_term_dumb_with_term(monkeypatch, term, expected):
    monkeypatch.setenv("TERM", term)
    assert term_dumb() is expected


def test_term_dumb_without_term(monkeypatch):
    monkeypatch.delenv("TERM", raising=False)
    assert term_dumb() is False


@pytest.mark.parametrize("value", ["", "1"])
def test_no_color_when_env_var_set(monkeypatch, value):
    monkeypatch.setenv("NO_COLOR", value)
    assert no_color() is True


def test_no_color_false_without_env_var(monkeypatch):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.setenv("TERM", "dumb")
    assert no_color() is False


def test_no_color_false_with_non_dumb_term(monkeypatch):
    monkeypatch.delenv("NO_COLOR", raising=False)
    monkeypatch.setenv("TERM", "xterm-256color")
    assert no_color() is False


@pytest.mark.parametrize("value", ["1", "true"])
def test_force_color_when_env_var_set(monkeypatch, value):
    monkeypatch.setenv("FORCE_COLOR", value)
    assert force_color() is True


@pytest.mark.parametrize("value", [None, ""])
def test_force_color_false(monkeypatch, value):
    if value is None:
        monkeypatch.delenv("FORCE_COLOR", raising=False)
    else:
        monkeypatch.setenv("FORCE_COLOR", value)
    assert force_color() is False


@pytest.mark.parametrize(
    "term,no_color_value,force_color_value,is_tty_value,expected",
    [
        ("xterm-256color", "1", "1", True, False),
        ("xterm-256color", None, "1", False, True),
        ("dumb", None, "1", False, True),
        ("dumb", None, None, True, False),
        ("unknown", None, None, True, False),
        ("xterm-256color", None, "", False, False),
        ("xterm-256color", None, None, True, True),
        ("xterm-256color", None, None, False, False),
    ],
)
def test_should_use_color(
    monkeypatch, term, no_color_value, force_color_value, is_tty_value, expected
):
    monkeypatch.setenv("TERM", term)
    if no_color_value is None:
        monkeypatch.delenv("NO_COLOR", raising=False)
    else:
        monkeypatch.setenv("NO_COLOR", no_color_value)
    if force_color_value is None:
        monkeypatch.delenv("FORCE_COLOR", raising=False)
    else:
        monkeypatch.setenv("FORCE_COLOR", force_color_value)

    monkeypatch.setattr("conda.common.terminal.is_tty", lambda: is_tty_value)

    assert should_use_color() is expected
```

conda/core/package_cache_data.py
```diff
@@ -28,11 +28,12 @@
 )
 from ..base.context import context
 from ..common.constants import NULL, TRACE
-from ..common.io import IS_INTERACTIVE, time_recorder
+from ..common.io import time_recorder
 from ..common.iterators import groupby_to_dict as groupby
 from ..common.path import expand, strip_pkg_extension, url_to_path
 from ..common.serialize import json
 from ..common.signals import signal_handler
+from ..common.terminal import is_tty
```
jezdez: Since the reporter now treats `TERM=dumb` the same as a non-animated terminal, I think the package download/extract status path should use the same condition.

```suggestion
from ..common.terminal import is_tty, term_dumb
```

conda/core/package_cache_data.py
```diff
@@ -810,7 +811,7 @@ def execute(self):
             if not context.verbose and not context.quiet and not context.json:
                 print(
                     "\nDownloading and Extracting Packages:",
-                    end="\n" if IS_INTERACTIVE else " ...working...",
+                    end="\n" if is_tty() else " ...working...",
```
jezdez: This surrounding status message should use the same terminal capability decision as the progress bar. Otherwise a real TTY with `TERM=dumb` can take the TTY header path while the progress bar itself is quiet.

```suggestion
                    end="\n" if is_tty() and not term_dumb() else " ...working...",
```

jezdez (commented):
Thanks Travis, this is much closer to fitting into the existing reporter path. The TTY / `TERM=dumb` progress direction makes sense to me, and I like having this logic move out of scattered `sys.stdout.isatty()` checks.

I do see one scope issue before I'd be comfortable approving this as-is: this PR no longer includes the ng/Rich console path where `NO_COLOR` and `FORCE_COLOR` would actually be consumed. With the current diff, those helpers are defined and tested, but production code does not use them.

So I think either this should stop closing #15943 and become groundwork for terminal/progress behavior, or the PR should include the ng/Rich integration that consumes those helpers. I'm fine with the former if we change `fixes: #15943` to `refs: #15943`.

I also left a few smaller suggestions around cache semantics, `TERM=dumb` consistency, and test isolation.

conda/core/package_cache_data.py
```diff
@@ -914,7 +915,7 @@ def cancelled():
                 bar.close()
 
             if not context.verbose and not context.quiet and not context.json:
-                if IS_INTERACTIVE:
+                if is_tty():
```
jezdez: Same condition here so the non-animated path gets the matching `done` suffix.

```suggestion
                if is_tty() and not term_dumb():
```

conda/core/package_cache_data.py
```diff
@@ -951,7 +952,7 @@ def _progress_bar(
             context_manager=context_manager,
             position=position,
             leave=leave,
-            enabled=not context.verbose and not context.quiet and IS_INTERACTIVE,
+            enabled=not context.verbose and not context.quiet and is_tty(),
```
jezdez: And this keeps the lower-level progress-bar enablement in sync with the reporter's `TERM=dumb` behavior.

```suggestion
            enabled=not context.verbose and not context.quiet and is_tty() and not term_dumb(),
```

tests/test_reporters.py
```diff
@@ -49,10 +49,14 @@ def test_render(capsys: CaptureFixture):
...
     """
     Ensure basic coverage of the :func:`~conda.reporters.get_progress_bar~` function
     """
+    mocker.patch(
+        "conda.plugins.reporter_backends.console.is_tty",
+        return_value=True,
+    )
     progress_bar_manager = get_progress_bar("test")
 
     assert isinstance(progress_bar_manager, TQDMProgressBar)
```
jezdez: This test now also depends on `term_dumb()`. In shells where `TERM=dumb`, patching only `is_tty()` still returns a `QuietProgressBar`.

```suggestion
def test_get_progress_bar(monkeypatch):
    """
    Ensure basic coverage of the :func:`~conda.reporters.get_progress_bar~` function
    """
    monkeypatch.setattr(
        "conda.plugins.reporter_backends.console.is_tty", lambda: True
    )
    monkeypatch.setattr(
        "conda.plugins.reporter_backends.console.term_dumb", lambda: False
    )
    progress_bar_manager = get_progress_bar("test")

    assert isinstance(progress_bar_manager, TQDMProgressBar)
```

tests/plugins/reporter_backends/test_console.py
```diff
@@ -230,3 +238,115 @@ def test_prompt_error_reading_stdin(mocker):
...
+
+
+def test_progress_bar_non_tty_returns_quiet_bar(monkeypatch):
+    """Verify progress_bar() returns QuietProgressBar when not in a TTY."""
+    monkeypatch.setattr("conda.plugins.reporter_backends.console.is_tty", lambda: False)
+    monkeypatch.setattr(
+        "conda.plugins.reporter_backends.console.term_dumb", lambda: False
+    )
+    monkeypatch.setattr(
+        "conda.plugins.reporter_backends.console.context", SimpleNamespace(quiet=True)
```
jezdez: Small test isolation issue: this case says it is testing the non-TTY branch, but `quiet=True` already forces `QuietProgressBar`. Setting `quiet=False` makes the test prove the `not is_tty()` condition.

```suggestion
        "conda.plugins.reporter_backends.console.context", SimpleNamespace(quiet=False)
```

travishathaway:
@jezdez,

I think that this pull request fulfills the requirements in issue #15943. There currently are no "ng" modules in conda, and when we decide to begin work on the latter half of this work, we'll be able to easily add those.

I think it would be best for us to merge this, consider the original issue closed and begin scoping and defining the next round. Even if we added "ng" modules, this wouldn't be get used anyway.

jezdez (approved):

conda/common/terminal.py
kenodegard: Closing the loop on this thread

Caching was initially added in https://github.com/conda/conda/pull/15982/commits/71e88180e21c6a4a2f2aa735d196dd0aae36d8f7

It was reverted in https://github.com/conda/conda/pull/15982/commits/fd44bea844d9b4fa0ae2620da100d7aa3f36bc07 per subsequent code review https://github.com/conda/conda/pull/15982#discussion_r3347083927

PR #15986, opened 2026-04-28
Add lazy properties to argparse help, description, and epilog

kenodegard:
### Description

Lazy properties:
- `conda.cli.helpers.LazyAction.choices`
- `conda.cli.helpers.LazyAction.help`
- `conda.cli.conda_argparse.ArgumentParser.description`
- `conda.cli.conda_argparse.ArgumentParser.epilog`

Part 1/2 of https://github.com/conda/conda/issues/15960

### Checklist - did you ...


jezdez:
@kenodegard To clarify do we need this vis a vis https://github.com/conda/conda/pull/15962?

kenodegard:
> To clarify do we need this vis a vis #15962?

Not a hard requirement, but definitely something we should add performance wise

jezdez:
Nice, this is a clean generalization of `LazyChoicesAction`. Left two inline comments. Will conflict with #15868 (A2/A3) which still has `LazyChoicesAction` in helpers.py, so let's coordinate merge order.

conda/cli/conda_argparse.py
```diff
@@ -198,13 +201,69 @@ def find_builtin_commands(parser: ArgumentParserBase) -> tuple[str, ...]:
...
-    def __init__(self, *args, add_help=True, **kwargs):
-        kwargs.setdefault("formatter_class", RawDescriptionHelpFormatter)
-        super().__init__(*args, add_help=False, **kwargs)
+    _description: str | None
+    _description_factory: Callable[[], str | None] | None
+    _epilog: str | None
+    _epilog_factory: Callable[[], str | None] | None
+
+    def __init__(
+        self,
```
jezdez, line 228: The bare `*` makes everything keyword-only, but stdlib argparse accepts `prog` positionally. Anything calling `ArgumentParser("my_prog")` will break. Safer to keep the old `*args` forwarding.

conda/cli/actions.py
```diff
@@ -60,3 +68,99 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+            self._choices = maybecall(self._choices_factory)
+            return self._choices
+
+    @choices.setter
+    def choices(self, value: Iterable[Any] | None):
+        if value is not None or self._choices_factory is None:
+            self._choices = value
+
+    @property
+    def help(self) -> str | None:
```
jezdez, line 118: Subtle behavior change: the old setter was a complete no-op, now it accepts non-None values. Probably better, but worth a news mention.

conda/cli/conda_argparse.py
kenodegard: The awkward part is only intercepting some parameters, if we widen the signature to stay stdlib-shaped, e.g.

```python
def __init__(self, *args, description=..., description_factory=..., ...):
```

then callers can mix parent positionals with our extras:

```python
ArgumentParser(
    "prog-str",
    "usage-str",
    "description-str",
    description_factory=lambda: ...,
)
```

The third positional is still `description` on the base class, so you can end up with both a literal description and `description_factory`, without tripping the mutex check.

Would renaming our subclass to something like `CondaParser` help? Renaming would signal it isn’t a drop-in alias of `argparse.ArgumentParser` and makes stricter constructor rules easier to swallow or do we keep `ArgumentParser` and enforce mutex another way (e.g. validation after `super()` parses positionals)?

PR #15987, opened 2026-04-28
Add conda-pypi as a dependency

danyeaw:
### Description

As part of https://github.com/conda/conda-pypi/issues/297, adds conda-pypi as a dependency of conda to ensure it is installed by default.

### Checklist - did you ...


dholth:
This one is a little spicy IMO. A circular dependency is never much fun.

danyeaw:
Hi @dholth, ya, you are right - circular dependencies are not great 😬. What we are trying to achieve is that some of these mature plugins (conda-pypi, conda-rattler-solver, conda-self, conda-lockfiles) are installed if someone upgrades to conda 26.5.x. What if we did something like conda-libmamba-solver, which is to comment it out in the `pyproject.toml`, but keep it as a run requirement in the recipe?

travishathaway:
I don't think this repository is the appropriate place to make this decision. This is the decision of the people in charge of packaging conda (e.g. conda's feedstock on conda-forge or Anaconda's main).

With that in mind, I don't think it needs to be mentioned in the pyproject.toml's dependencies at all. Beyond development of the project, I'm not even sure what the point of listing things there is 🤷.

danyeaw:
Hey @travishathaway, thanks for the feedback. Aren't we in charge of what the default experience of conda looks like? We have decided to use a plugin approach to make conda more extensible, which is great, but shouldn't we determine which plugins are mandatory and which ones are optional?

Another approach we could take is when a plugin becomes part of the default experience, we could bring it back in to the conda repo. An enabler for that is the updates to our deprecation policy.

travishathaway:
@danyeaw,

It's a good thing we're also in charge of packaging! 😉 I'm just saying that the feedstocks are the appropriate place to declare these dependencies.

jezdez:
I'm not sure if I understand, why wouldn't we want to add conda-pypi to conda as a dependency? The current pip integration in conda's environment.yml parsing is what we intend to fix with this, making it safer for conda users to install PyPI packages. Do you have concerns around this as the default behavior, @travishathaway?

dbast:
there are multiple places where this could be added:
* the meta.yaml / pyproject.toml in this repo, BUT conceptionally conda does not depend on conda-pypi even though `run dependencies` try to tell that.
* the meta.yaml in various feedstocks to actually reflect the dependencies that conda defines in its pyproject.toml 
* the construct.yml file of miniconda/miniforge to deliver some packages by default even if nothing else depends on it 

so, does the conda package of conda stand for conda alone or for the intended user experience?

How is a plugin elevated to a default plugin that should be always there and be part of the default user experience?

danyeaw:
If we only update the feedstocks and not the recipe in the repo, I'm concerned that there could be issues that could have been found in the canary builds but we wait until a release to find them.

The problem with us handling it from the installers like Miniforge and Miniconda is that users won't see the added plugins unless they reinstall using the installer, unless they manually install the plugins themselves. I would expect that users update their base conda pretty regularly when we tell them it is out of date, but would reinstall from scratch not that often.

One way we could signal to users which plugins are in the default experience is to add a documentation page about these plugins.

jezdez:
> so, does the conda package of conda stand for conda alone or for the intended user experience?

The purpose of the plugin system was always to enable conda to get retrofits of functionality via separate code projects, maintained independently from the conda release cycle, to ensure more consistency. A large reason why, for example, pytest or WordPress don't ship with all existing plugins is because that would negate the maintenance upsides of the separation of concerns. On the other hand, a select few plugins may be adopted by the maintainers of the main project as providing essential functionality. conda-libmamba-solver is the precedent here, all but undisputed as essential and maintained for years as a separate package.

> How is a plugin elevated to a default plugin that should be always there and be part of the default user experience?

Graduating from incubation is a necessary bar, but what makes conda-pypi a strong candidate beyond that is the combination of factors: it's maintained by a dedicated team with large overlap with the conda maintainers, it was a focused effort over more than a year to improve conda/PyPI compatibility, and it addresses a real gap in the default experience that users currently have to work around manually. Why _shouldn't_ we make use of this?

Separately, on the question of where to declare this: I think the recipe in this repo is the right place. If we only update feedstocks, we lose canary coverage and don't find integration issues until release time. And if we push it to the installers, users only see the plugin when they do a fresh install, not when they update conda, which most users do more regularly.

travishathaway:
Thanks for mentioning the canary builds. I forgot about those. I'm okay with adding it here.

travishathaway (approved):

jezdez (approved):

PR #15994, opened 2026-04-29
Adapt conda-libmamba-solver sharded repodata code to live in conda

### LINKED ISSUE #15906: Host sharded repodata from conda-libmamba-solver
### Checklist

### What is the idea?

### Why is this needed?

### What should happen?

### Additional Context

_No response_

dholth:
### Description

Fix #15906 

- [ ] Sharded repodata flag is part of libmamba-solver plugin. Implement a conda flag?

### Checklist - did you ...


conda/cli/install.py
```diff
@@ -328,14 +328,29 @@ def install(args, parser, command="install"):
...
+                "subdirs": context.subdirs,
+                "specs_to_add": env.requested_packages,
+                "repodata_fn": repodata,
+                "command": args.cmd,
+            }
+
+            # Check if solver supports build_repodata_subset parameter
+            sig = inspect.signature(solver_backend.__init__)
+            if "build_repodata_subset" in sig.parameters:
+                solver_kwargs["build_repodata_subset"] = conda_build_repodata_subset
```
dholth, line 351: I considered adding the code that detects `build_repodata_subset to `context.plugin_manager.get_cached_solver_backend()` but it didn't feel right. Should we move this code there, or is there a better place for this helper? We could return it as a `functools.partial(...)` supplying the `build_repodata_subset` callable.

conda/gateways/shards/typing.py
```diff
@@ -0,0 +1,49 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+"""Minimum set of interfaces required to describe shards code to clients (solvers)"""
+
+import typing
+from collections.abc import Callable, Iterable, Iterator
+from typing import Literal
+
+
+class Shards(typing.Protocol):
+    base_url: str
```
dholth, line 11: It would be easy to include the list of package names here, as a `dict.keys()` view, since we require that list as part of shards traversal.

If we wanted to use this to implement "conda search", it could use the package names aggregated across all channels, then request a repodata subset with a new "don't follow dependencies" flag that would only fetch root packages.

Or we could expose a bit of the `RepodataSubset`() class and add methods to update an existing subset, which is the {channel: Shards()} dict.

We would need to disable or toggle build-repodata-subset's "return None if there are no shards" logic, which happens if no channel has shards.

pyproject.toml
dholth, line 34:
```suggestion
  "menuinst >=2",
  "msgpack >=1.1.1",
```

_conda/shards/cache.py
```diff
@@ -0,0 +1,215 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from .typing import ShardDict
+
+log = logging.getLogger(__name__)
+
+SHARD_CACHE_NAME = "repodata_shards.db"
+ZSTD_MAX_SHARD_SIZE = (
```
ryanskeith, line 26: This is also in shards.py.  No a big deal but there is duplication.

_conda/shards/cache.py
```diff
@@ -0,0 +1,215 @@
...
+        if "://" in package:
+            raise ValueError("package must not contain '://'")
+
+        self.url = url
+        self.package = package  # remove this field to avoid confusion?
+        self.compressed_shard = compressed_shard
+
+    url: str
+    package: str
+    compressed_shard: bytes
```
ryanskeith, line 46: We could simplify this a little to use the dataclasses default init.
```suggestion
    url: str
    package: str  # remove this field to avoid confusion?
    compressed_shard: bytes
    
    def __post_init__(self) -> None:
        # prevent easy mistake of swapping url, package
        if "://" not in url:
            raise ValueError("url must contain '://'")
        if "://" in package:
            raise ValueError("package must not contain '://'")
```

_conda/shards/cache.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def connect(dburi="cache.db"):
```
ryanskeith, line 49: Maybe a rename so the usage of connect ShardCache.connect are easier to read?

_conda/shards/cache.py
```diff
@@ -0,0 +1,215 @@
...
+            # propagate.
+            has_errorcode = hasattr(e, "sqlite_errorcode")
+            if retry and (
+                (not has_errorcode) or (e.sqlite_errorcode == sqlite3.SQLITE_NOTADB)
+            ):
+                log.warning("%s '%s'; remove and retry.", dburi, e)
+                try:
+                    self.remove_cache()
+                except OSError as e:
+                    # alternate filename if primary cannot be removed.
```
ryanskeith, line 134: What if this one can't be removed either?

_conda/shards/cache.py
dholth: I guess we would want a constants.py or import all from cache.py to avoid circular dependencies

_conda/shards/misc.py
```diff
@@ -0,0 +1,211 @@
...
+
+    legacy_packages = repodata.get("packages", {})
+    conda_packages = repodata.get("packages.conda", {})
+
+    return {
+        **repodata,
+        "packages": {
+            k: v
+            for k, v in legacy_packages.items()
+            if f"{k[:-_len_tar_bz2]}{_conda}" not in conda_packages
```
ryanskeith, line 163: I wonder if string replace would be faster here.  Or perhaps not using `.conda` directly like this.
```suggestion
            if f"{k[:-_len_tar_bz2]}.conda" not in conda_packages
```
The intent is clear though so I can appreciate why it is written as is.

_conda/shards/shards.py
```diff
@@ -0,0 +1,785 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
```
ryanskeith, line 2:
```suggestion
```

_conda/shards/misc.py
```diff
@@ -0,0 +1,211 @@
...
+
+    return result
+
+
+def _is_http_error_most_400_codes(status_code: str | int) -> bool:
+    """
+    Determine whether the `HTTPError` is an HTTP 400 error code (except for 416).
+    """
+    return (
+        isinstance(status_code, int) and 400 <= status_code < 500 and status_code != 416
```
ryanskeith, line 114: Curious, why do we not care about 416?

_conda/shards/shards.py
```diff
@@ -0,0 +1,785 @@
...
+        """
+        if not self._fetched:
+            if isinstance(self.shardbase, Shards):
+                self._shard = self._fetch_from_shards()
+            else:  # ShardLike
+                self._shard = self.shardbase.visit_package(self.package)
+            self._fetched = True
+        return self._shard
+
+    def _fetch_from_shards(self) -> ShardDict:
```
ryanskeith, line 107: Perhaps we could cut this as it is only used once?

_conda/shards/shards.py
```diff
@@ -0,0 +1,785 @@
...
+    ShardFetch.fetch_batch(wanted)
+
+
+def fetch_channels(url_to_channel: dict[str, Channel]) -> dict[str, ShardBase] | None:
+    """
+    Args:
+        url_to_channel: not modified, must already be expanded to subdirs.
+
+    Attempt to fetch the sharded index first and then fall back to retrieving a
+    monolithic `repodata.json` file.
```
ryanskeith, line 725: We may want to try falling back to current_repodata.json first before going to repodata.json to still give speed where available.

_conda/shards/subset.py
```diff
@@ -0,0 +1,768 @@
...
+        subset = RepodataSubset(
+            (*channel_data.values(),), spec_to_package_name=spec_to_package_name_func
+        )
+        subset.reachable(root_packages, strategy=algorithm)
+        log.debug("%d (channel, package) nodes discovered", len(subset._nodes))
+
+    return channel_data
+
+
+# region workers
```
ryanskeith, line 577: Perhaps workers can be another module???

conda/cli/install.py
```diff
@@ -328,14 +328,29 @@ def install(args, parser, command="install"):
...
+
+            solver = solver_backend(**solver_kwargs)
```
ryanskeith, line 353: This pattern is repeated in main_remove.py, env/installers/conda.py and others.  We should write a helper or two so we can change it in one place.

ryanskeith (commented):
Major change.  I did some cases where we could make some small code improvements.  By moving it into conda proper, I feel we will get more eyes on it and it will improve.  I would suggest we improve at least the first and last of the following three things:

First, let's assure the repodata.json chain will remain intact going from shards repodata.json -> current_repodata.json then the repodata.json.  This could be a future enhancement though if there are sharding issues, current_repodata.json is a nice thing to have.

Second, it would be nice to add a helper function for the pre steps for the solver backend.  The code repeats the same like steps in a few places.  This is a smaller issue though and could be added later.

Lastly, the code needs to add msgpack to pyproject.toml for shards.

I haven't gone over the tests yet.  I will do so tomorrow.  This is Early feedback for the non US west coasters.

_conda/shards/cache.py
dholth: Maybe it doesn't need to be a `@dataclass` at all. Supposedly they are a bit slower than plain classes.

_conda/shards/cache.py
dholth: So we wouldn't have too many repetitions of the word `connect` that mean different things?

_conda/shards/cache.py
dholth: There is a `retry=False` to bound recursion.

_conda/shards/misc.py
dholth: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/416 means a range request was not satisfiable. Range requests were important for jlap. All the other 4xx codes would cause us to assume this remote file doesn't exist.
We could get rid of this function since no range requests are used.

_conda/shards/misc.py
dholth: Yup it's a little odd. We want to be correct if we inexplicably get a "file.conda.xyz.conda".

_conda/shards/shards.py
dholth: I pushed a change that removes all these and replaces them with the pre-commit-added license.

_conda/shards/shards.py
dholth: Yes, sorry a bit slop.

_conda/shards/shards.py
dholth: `current_repodata.json` cannot be used by any solver except the discontinued `classic` one.

_conda/shards/subset.py
dholth: I would be interested in separating both subset algorithms from `RepodataSubset()` in an elegant way, but it's also convenient to have it together.

dholth (commented):
Thanks for the feedback!

conda/cli/install.py
ryanskeith: It is a tough one.  I totally agree.

_conda/shards/cache.py
ryanskeith: `@dataclass` gives us the init for free.  There is also using slots=True and other things to make it more performant.  As you say it could be worth a look to see if just a plain dict with validation is good enough and performant.

_conda/shards/cache.py
ryanskeith: My thought is that I would see something like `get_connection` in the ShardCache code and immediately understand the difference between the two.  Not a big deal but I did do a double take and thus pointed it out.

_conda/shards/misc.py
ryanskeith: That's a great point. We would need to chop it off from the end.

_conda/shards/misc.py
ryanskeith: I see.  Thank you so much for the explaination.

_conda/shards/shards.py
ryanskeith: No hay de que.

_conda/shards/shards.py
ryanskeith: WHAAATTT!  Mind blown.  I guess I am too used to the classic from my packaging days.

_conda/shards/cache.py
dholth: Not a plain `dict` but a plain `class`, since this class doesn't really need any of the `@dataclass` features.

conda/gateways/shards/typing.py
dholth: Oh, I have that list of packages in the implementation; will add it here.

dholth:
See #16507, #16508, #16509 which split this PR into smaller pieces.

PR #15998, opened 2026-04-30
(DNM) Move all requirements.txt files to a single environment.yml

jaimergp:
### Description

Just an experiment combining the upcoming conditional dependencies plus a possible extension of optional dependency groups.

### Checklist - did you ...


jezdez:
A few question on this:

- there is a naming overlap with https://github.com/conda/ceps/pull/165 that would solve this for a different purpose, how would we teach this without confusing people?
- how would the selection happen on the CLI? `conda create --extra tests --extra s3`  or similar?
- would some extras be able to include other extras? or would yaml anchors be able to be used?
- how would the pip: section relate to this?
- how does this relate to the category field in CEP24, https://github.com/conda/ceps/blob/main/cep-0024.md#optional-extensions
wouldn’t this require a new CEP?
- finally, how would this relate to https://github.com/conda-incubator/conda-workspaces/issues/52

A minor comment, I do worry this proposal would further entrench the environment.yml format, even though we know it’s not ideal in user ergonomics, e.g. I’m pushing for [conda-workspaces](https://github.com/conda-incubator/conda-workspaces) in the sense of the declarative environment work we we started last year, to make it easier for conda users to manager their environment and not having to think about locking etc

That said, conda-workspaces is in incubation for that reason, to challenge those assumptions if needed

jezdez:
Oh, a few more:

- what would be the default behavior be, install no extras?
- would we want to have per extra group channel differences?

jaimergp:
Thanks @jezdez, really appreciate your comments; replying inline here.

> A few question on this:
> 
> - there is a naming overlap with https://github.com/conda/ceps/pull/165 that would solve this for a different purpose, how would we teach this without confusing people?

There's no overlap, it's intentionally the same design as in recipes.

> - how would the selection happen on the CLI? `conda create --extra tests --extra s3`  or similar?

That's a good option! `--with` could also fly.
> - would some extras be able to include other extras? or would yaml anchors be able to be used?

You can depend on matchspecs that depend on extras, so maybe `.[extras=another]`? Edge case and I'm leaning to say no. If the YAML parser can use anchors, why not, but they are often not as useful because you can't "merge" lists.

> - how would the pip: section relate to this?

Same as in the top-level. Within a group, add `pip: list[str]`.

> - how does this relate to the category field in CEP24, https://github.com/conda/ceps/blob/main/cep-0024.md#optional-extensions

Slightly in concept, but they don't offer the same functionality.

> wouldn’t this require a new CEP?

If we want to make this work, yes. For now it's just a vision.

> - finally, how would this relate to https://github.com/conda-incubator/conda-workspaces/issues/52

Very similar in goals, different approach. I feel this one here needs way less adoption effort for existing users, and that's its main strength. It falls short on everything else.

> A minor comment, I do worry this proposal would further entrench the environment.yml format, even though we know it’s not ideal in user ergonomics, e.g. I’m pushing for [conda-workspaces](https://github.com/conda-incubator/conda-workspaces) in the sense of the declarative environment work we we started last year, to make it easier for conda users to manager their environment and not having to think about locking etc

Yes, I agree, but see my angle about adoption effort above. It requires fewer updates to the mental model, I'd say.
 
> - what would be the default behavior be, install no extras?

Yes, top-level stays king.

> - would we want to have per extra group channel differences? 

That's feasible but I didn't think that far. We could also discuss what happens with multi-document YAMLs (separated by `---`) and whether that's a good abstraction for dependency groups at the input level. By adding a naming convention to `name:` (e.g. the same string can be appended different `[extra=group]` tags?), one can craft multi-env YAMLs without violating the CEP... technically 😬 

See Gemini-generated prototype at https://gist.github.com/jaimergp/d1855c677947ff7300751b7e680a3a2a

PR #15999, opened 2026-04-30
Fix `TypeError` from non-RFC-9457 `detail` in HTTP error classes

agriyakhetarpal:
### Description

Spotted in https://github.com/conda-incubator/pytest-conda-solvers/pull/20. `CondaHTTPError.__init__` (and the same path in `UnavailableInvalidChannel`) reads the `detail` field from any JSON response body and assumes it is a string, following RFC 9457. However, the code does not check whether the response carries `Content-Type: application/problem+json`, which is the content type that RFC 9457 requires for responses. So any server that independently uses `detail` in its own JSON schema can produce a non-string value here. 

This happens with FastAPI, for instance, which uses `detail` as a list of structured validation error objects in its `application/json` 422 responses. This is FastAPI's own error format, not RFC 9457 – so I'd say that they haven't implemented the RFC let alone conformed with it. There's an unchecked item in https://github.com/fastapi/fastapi/issues/10370 about this. This causes a `TypeError: can only concatenate str (not "list") to str` error. This PR fixes both classes to only read `detail` when the response carries `Content-Type: application/problem+json`, as RFC 9457 requires. Non-RFC-9457 responses (such as FastAPI's `application/json` `422`s) are no longer affected.

That said, there are FastAPI plugins that implement the RFC: https://pypi.org/project/fastapi-problem/ and https://pypi.org/project/fastapi-problem-details/ (and one can write their own middleware or whatever).

~Maybe the long-term fix would be to gate the RFC 9457 parsing on `Content-Type: application/problem+json`, but that is a separate change that I want to keep scoped, so I am proposing this here first. Please LMK if you don't think we should go ahead with this; I am only suggesting it because FastAPI is common enough as a framework.~ We went ahead with gating it for RFC 9457 here. Thanks!

cc: @jezdez

### Checklist - did you ...


conda/exceptions.py
```diff
@@ -661,9 +672,20 @@ def __init__(
...
+            # content type. Other frameworks independently use 'detail' in their own
+            # schema. FastAPI, for example, sets 'detail' to a list of structured
+            # validation error objects in plain application/json 422 responses.
+            # Those are not RFC 9457 responses, but the field name collision means
+            # we may receive a non-string here. We can stringify in that case.
+            # See https://datatracker.ietf.org/doc/html/rfc9457#section-3.1.4
+            # See https://github.com/conda-incubator/pytest-conda-solvers/pull/20
+            detail = body.get("detail")
+            if detail is not None:
+                message = detail if isinstance(detail, str) else str(detail)
```
kenodegard: Since this is an exact duplicate and also includes a large detailed comment we should move this into a helper function

agriyakhetarpal: Done! 808f4a7d577f40c07a4b888baf490761b2c3ca67

jezdez (dismissed):
Thanks for the fix -- the underlying bug is real. A few suggestions on the approach:

**1. Gate on `Content-Type` instead of stringifying.** RFC 9457 scopes `detail` to `application/problem+json` responses ([Section 3](https://www.rfc-editor.org/rfc/rfc9457#section-3)). The current code (and this PR) reads `detail` from any JSON body without checking the content type, which is why FastAPI's `application/json` responses collide. The fix should be to only read `detail` when the response carries the correct content type. Non-string `detail` from other schemas should be ignored, not stringified -- `str([{"loc": ...}])` isn't useful to users anyway.

**2. Drop the private helper function.** `_extract_http_response_detail` is a private module-level function in `exceptions.py` for essentially one line of logic. Per our [conventions](https://github.com/conda/conda/blob/main/AGENTS.md), prefer inlining at the call site. If deduplication of the shared body-parsing block is wanted (the whole `try: response.json()` block is duplicated between both classes), that would be a good follow-up.

**3. Use `monkeypatch` instead of `unittest.mock.patch`** for new test code (per [AGENTS.md](https://github.com/conda/conda/blob/main/AGENTS.md#tests)). The existing `patch` import is pre-existing, but new tests shouldn't add to it.

See inline suggestions for the proposed change at both call sites.

conda/exceptions.py
```diff
@@ -524,6 +524,29 @@ class ChannelDenied(ChannelNotAllowed):
...
+    Those are not RFC 9457 responses, but the field name collision means
+    we may receive a non-string here. We can stringify in that case.
+
+    See https://datatracker.ietf.org/doc/html/rfc9457#section-3.1.4
+    See https://github.com/conda-incubator/pytest-conda-solvers/pull/20
+    """
+    detail = body.get("detail")
+    if detail is None:
+        return None
+    return detail if isinstance(detail, str) else str(detail)
```
jezdez: This function can be removed entirely. Instead of defensively stringifying non-string `detail` values, gate on the content type that RFC 9457 requires (see suggestions on the call sites below).

conda/exceptions.py
```diff
@@ -567,9 +590,8 @@ def __init__(
         else:
             reason = body.get("reason") or reason
             message = body.get("message") or message
-            # if RFC 9457 'detail' is present, it is preferred over 'message'
-            # See https://datatracker.ietf.org/doc/html/rfc9457
-            message = body.get("detail") or message
+            if (detail := _extract_http_response_detail(body)) is not None:
+                message = detail
```
jezdez: Replace the function call with a content-type check. RFC 9457 scopes `detail` to `application/problem+json` responses, so we should only read it when the content type matches. Non-RFC responses (like FastAPI's `application/json` 422s) will simply be ignored.

```suggestion
            content_type = getattr(response, "headers", {}).get("content-type", "")
            if "application/problem+json" in content_type:
                detail = body.get("detail")
                if isinstance(detail, str):
                    message = detail
```

conda/exceptions.py
```diff
@@ -661,9 +683,8 @@ def __init__(
...
```
jezdez: Same change here:

```suggestion
            content_type = getattr(response, "headers", {}).get("content-type", "")
            if "application/problem+json" in content_type:
                detail = body.get("detail")
                if isinstance(detail, str):
                    message = detail
```

jezdez (changes requested):
Thanks for the fix -- the underlying bug is real. A few suggestions on the approach:

**1. Gate on `Content-Type` instead of stringifying.** RFC 9457 scopes `detail` to `application/problem+json` responses ([Section 3](https://www.rfc-editor.org/rfc/rfc9457#section-3)). The current code (and this PR) reads `detail` from any JSON body without checking the content type, which is why FastAPI's `application/json` responses collide. The fix should be to only read `detail` when the response carries the correct content type. Non-string `detail` from other schemas should be ignored, not stringified -- `str([{"loc": ...}])` isn't useful to users anyway.

**2. Drop the private helper function.** `_extract_http_response_detail` is a private module-level function in `exceptions.py` for essentially one line of logic. Per our [conventions](https://github.com/conda/conda/blob/main/AGENTS.md), prefer inlining at the call site. If deduplication of the shared body-parsing block is wanted (the whole `try: response.json()` block is duplicated between both classes), that would be a good follow-up.

**3. Use `monkeypatch` instead of `unittest.mock.patch`** for new test code (per [AGENTS.md](https://github.com/conda/conda/blob/main/AGENTS.md#tests)). The existing `patch` import is pre-existing, but new tests shouldn't add to it.

See inline suggestions for the proposed change at both call sites.

agriyakhetarpal:
Thanks @jezdez for the review and the inline code comments. Looks like we settled on my suggestion to gate it for RFC 9457 responses 😄 I'll edit the PR description and title.

Edit: the test failures are unrelated.

kenodegard (approved):

conda/exceptions.py
```diff
@@ -568,9 +568,11 @@ def __init__(
         else:
             reason = body.get("reason") or reason
             message = body.get("message") or message
-            # if RFC 9457 'detail' is present, it is preferred over 'message'
-            # See https://datatracker.ietf.org/doc/html/rfc9457
-            message = body.get("detail") or message
+            content_type = getattr(response, "headers", {}).get("content-type", "")
```
kenodegard, line 574: hm might as well keep the old comment/link


```suggestion
            # RFC 9457 'detail' is preferred over 'message' for
            # application/problem+json responses.
            # See https://datatracker.ietf.org/doc/html/rfc9457
            content_type = getattr(response, "headers", {}).get("content-type", "")
```

conda/exceptions.py
```diff
@@ -664,9 +666,11 @@ def __init__(
...
```
kenodegard, line 675:
```suggestion
            # RFC 9457 'detail' is preferred over 'message' for
            # application/problem+json responses.
            # See https://datatracker.ietf.org/doc/html/rfc9457
            content_type = getattr(response, "headers", {}).get("content-type", "")
```

PR #16014, opened 2026-05-04
Adds new `conda._preview` module for upcoming features

### LINKED ISSUE #16011: Add `conda/_preview/` namespace with `register()` convention and `env-setup` stubs
### Checklist

### What is the idea?

Add the `conda/_preview/` directory convention with a `register(context)` callback pattern, and scaffold the first preview (`env-setup`) with non-operational stubs wired into CLI dispatch.

### Why is this needed?

- Preview code needs a standard home that mirrors conda's module structure and is clearly private.
- A `register(context)` convention lets each preview control its own setup, keeping heavy imports deferred until the preview is actually enabled.
- The `env-setup` stubs prove the wiring works end-to-end without shipping functional changes.

### What should happen?

- Add `conda/_preview/` as a private namespace. Each preview gets a subdirectory mirroring conda's module structure.
- Convention: each preview's `__init__.py` exposes a `register(context)` function, called after final `context.__init__()` in `main()` for each enabled preview.
- Add `conda/_preview/env_setup/` with:
  - `__init__.py` with `register(context)` entry point
  - `cli/main_create.py` and `cli/main_install.py` stubs (raise `NotImplementedError` or similar)
- In `do_call()` (`conda/cli/conda_argparse.py`), route `create` and `install` to the preview modules when `context.preview("env-setup")` is true.
- No `python -m conda.ng` or separate CLI entry point.
- Add tests for the registration flow and CLI routing.

### Additional Context

- Part of #15940. Depends on #16010 (`context.preview`).
- Graduation path: when a preview becomes the default, its code moves from `_preview/` into the main tree, the label gets a deprecation notice, and the subdirectory is cleaned up.

### LINKED ISSUE #16019: Document `conda._preview` in developer documentation
### Checklist

### What happened?

We've recently added a new module to conda called, `conda._preview`. This is a special module that is intended to give us a space to move fast, prototype and experiment with radical changes to conda CLI. I want to make sure that developers have a clear understanding of why it exists, so that I'm proposing creating a new section in the developer documentation dedicated to this module.

### What should happen?

- [ ] Create a single page in the developer documentation called, "Preview"
- [ ] On this page, describe how it works, including:
  - [ ] Why is the module organization a mirror of the top level conda modules?
  - [ ] How do I enable overriding conda's subcommands? (including a trivial example)
  - [ ] How is it enabled? (e.g. `context.preview`)
  - [ ] What are the rules surrounding deprecation in this module?

### Additional Context

Please see this issue for all the work done so far for this:

- https://github.com/conda/conda/issues/15940

travishathaway:
### Description

fixes: #16011 

### Checklist - did you ...


jezdez (commented):
Couple of small suggestions — looks good otherwise.

jezdez (commented):
Couple of small suggestions — looks good otherwise.

conda/cli/main.py
```diff
@@ -46,6 +47,16 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
 
+    # Call register(context) for each enabled preview feature (imports are lazy/deferred).
+    # Unrecognized labels are silently skipped so typos don't crash the CLI.
+    for _preview_label in context.preview:
+        _preview_pkg = _preview_label.replace("-", "_")
+        try:
+            _preview_mod = import_module(f"conda._preview.{_preview_pkg}")
+            _preview_mod.register(context)
+        except ModuleNotFoundError:
+            pass
```
jezdez: The `except` wraps both `import_module()` and `register()` — a missing import *inside* `register()` gets silently swallowed. Consistent with how pre/post command hooks propagate, split them:

```suggestion
    for label in context.preview:
        package = label.replace("-", "_")
        try:
            mod = import_module(f"conda._preview.{package}")
        except ModuleNotFoundError:
            continue
        mod.register(context)
```

jezdez (commented):
Couple of small suggestions — looks good otherwise.

conda/cli/conda_argparse.py
```diff
@@ -182,9 +183,29 @@ def do_call(args: argparse.Namespace, parser: ArgumentParser):
...
+                "conda.cli", f"conda._preview.{_pkg}.cli", 1
+            )
+            if (
+                _candidate != module_name
+                and importlib.util.find_spec(_candidate) is not None
+            ):
+                module = import_module(_candidate)
+                break
+        if module is None:
+            module = import_module(module_name)
```
jezdez: Nit: leading underscores on locals read as "private/unused" — plain names are more natural here:

```suggestion
        module = None
        for label in context.preview:
            package = label.replace("-", "_")
            candidate = module_name.replace(
                "conda.cli", f"conda._preview.{package}.cli", 1
            )
            if (
                candidate != module_name
                and importlib.util.find_spec(candidate) is not None
            ):
                module = import_module(candidate)
                break
        if module is None:
            module = import_module(module_name)
```

tests/conftest.py
```diff
@@ -296,6 +296,7 @@ def __init__(self, *args, **kwargs):
             self.plugin_manager = mocker.MagicMock()
             self.repodata_fns = ["repodata.json", "current_repodata.json"]
             self.subdir = mocker.MagicMock()
+            self.preview = ("env-setup",)
```
jezdez: This means every test using this mock context now has `env-setup` active — could trip `OperationNotAllowed` in existing `do_call()` tests. Safer as empty?

```suggestion
            self.preview = ()
```

jezdez (approved):
Looking great with comments above, there is a slight increase in scope again due to the generic subcommand handling in `conda_argparse.py` you added. I think it works well though!

travishathaway:
> Looking great with comments above, there is a slight increase in scope again due to the generic subcommand handling in `conda_argparse.py` you added. I think it works well though!

It felt like the right way to go. We have the appropriate module resolution right there in the `func` attribute, so why not just pull the ol' switcheroo on module resolution? #ThingsYouCanDoWithPython

conda/cli/conda_argparse.py
```diff
@@ -182,9 +183,29 @@ def do_call(args: argparse.Namespace, parser: ArgumentParser):
...
+            )
+            if (
+                candidate != module_name
+                and importlib.util.find_spec(candidate) is not None
+            ):
+                module = import_module(candidate)
+                break
+        if module is None:
+            module = import_module(module_name)
+
```
kenodegard: I don't think we should do this.

Why don't we require previews to use the plugin framework?

jezdez: Please provide more reasons why we shouldn't do this so we can find a resolution.

We're working backward from an unrefined design (see earlier descriptions in https://github.com/conda/conda/issues/15940, and the epic https://github.com/conda/conda/issues/15873) around injecting the "ng" project into conda that allows a parallel codebase to exist, overwriting key aspects without touching the rest of the infrastructure. Jaime had PoCs that felt risky to ship to regular users. It's basically a feature flagging system to override sections of the conda codebase.

Looking at the _other_ proposals, this approach here was designed to be less controversial and simpler (by using the import system), per my refinement in https://github.com/conda/conda/issues/15940#issuecomment-4369255358. I did consider using the plugin system (and built a small PoC to try it out), but thought it would be too intertwined with the rest of the code to enable the flexibility asked in the epic. I'd be interested in following through with it TBH, if it fits the intended goals, and _that's_ seems most of the problem.

jezdez (changes requested):
Marking as needing a design discussion before moving on

conda/cli/conda_argparse.py
kenodegard: I'm not opposed to the `conda._preview` idea or the overall design goal, having a sanctioned space for preview features that users can opt into makes sense. My issue is with introducing a new subcommand loading mechanism when the plugin framework already exists for exactly this purpose.

The main thing blocking plugins from doing this today is the denylist that prevents overriding built-in subcommands. If we relax that restriction (or make it opt-in via the same kind of condarc flag this PR already uses), the plugin system can load `conda._preview` implementations without needing a parallel dispatch path. The hook would look something like:

```python
from conda.plugins import hookimpl, types

@hookimpl
def conda_subcommands():
    if "env_setup" in context.preview:
        yield types.CondaSubcommand(
            name="create",
            summary="...",
            action=preview_create,
        )
```

That would keep the `conda._preview` namespace and the opt-in UX, just without the new loading machinery.

travishathaway: @jezdez and @kenodegard,

I'm okay with the approach Ken laid out. It's less code to maintain. Everything can just be kept in `conda.plugins` module as we're working on the new interface.

Let's make a decision on this soon.

jezdez: I'm in favor of this proposal, relaxing the restriction on plugin-based command overrides (but not removing it for non-preview use cases), via this new preview mechanism.

travishathaway: > I'm in favor of this proposal, relaxing the restriction on plugin-based command overrides (but not removing it for non-preview use cases), via this new preview mechanism.

@jezdez, what does this look like? Do we only allow overriding core commands from modules that are registered within conda (`conda.*`) itself?

conda/cli/conda_argparse.py
```diff
@@ -140,27 +169,20 @@ def generate_parser(**kwargs) -> ArgumentParser:
...
+        name
+        for name, plugin_subcommand in context.plugin_manager.get_subcommands().items()
+        if _is_preview_command(plugin_subcommand)
+    }
+
+    # We only register if it's not being overriden by something in the
+    # `conda._preview` module
+    for builtin, config_func in BUILTIN_COMMAND_PARSERS.items():
+        if builtin not in preview_plugin_commands:
+            config_func(sub_parsers)
```
travishathaway, line 178: @kenodegard and @jezdez,

Here's the meat and potatoes 🥩 🥔  of my changes. Unfortunately, there is no `remove_parser` method available to us when registering command parsers, so we have to interrupt the process here to see if we should register the built-in command or not.

conda/cli/conda_argparse.py
```diff
@@ -273,6 +295,15 @@ def _exec_unix(executable_args, env_vars):
     os.execvpe(executable_args[0], executable_args, env_vars)
 
 
+def _is_preview_command(plugin_subcommand: CondaSubcommand) -> bool:
+    """
+    Determine whether this plugin subcommand is a preview command.
+    """
+    module_name = getmodule(plugin_subcommand.action).__name__
+
+    return module_name.startswith("conda._preview")
```
travishathaway: @jezdez,

Thoughts on this approach? Sufficient or scary? 😅

travishathaway:
Made some changes today. Still not finished, but I would appreciate some feedback to see if I'm heading in the right direction.

travishathaway:
I added a new page to the documentation today to explain previews and this new module. Happy to hear feedback on it, especially the, "Guidelines for creating previews". Is there anything else that should be added there?

jezdez:
I opened a stacked PR with the concrete follow-up changes for this branch: https://github.com/travishathaway/conda/pull/5

Summary of the approach there:

- Route bundled preview subcommands through a dedicated `conda.plugins.previews` module, where the trusted preview override policy lives.
- Keep preview subcommands as ordinary `CondaSubcommand` hook results. The preview label grouping stays private to the preview aggregator, which gates bundled hook providers with `context.preview_enabled(label)`.
- Keep `CondaSubcommand` unchanged for external plugin authors; there is no preview-specific public field or generic label API in this follow-up.
- Keep `conda_argparse.py` focused on parser registration; it only asks `is_preview_subcommand(...)` when deciding whether a built-in command may be overridden.
- For trusted preview overrides of built-in commands, register the existing built-in parser, optionally let the preview subcommand extend that parser via `configure_parser`, and redirect execution to the preview subcommand. This keeps create/install parser definitions in one place while letting normal invocations reach the non-operational preview stubs before raising `OperationNotAllowed`.
- Preserve bundled preview subcommands under `--no-plugins`.
- Preserve `ModuleNotFoundError`s raised from inside preview packages instead of treating them as unknown preview labels.

The narrow design choice is intentional: this does not add a general plugin label/grouping API. The only new routing policy is private to bundled previews, and the public plugin type stays as-is until there is a broader concrete use case.

I ran the targeted preview tests and the full `tests/plugins/test_subcommands.py` module locally. I used `-W ignore::DeprecationWarning` for pytest because my local dev env has an unrelated external `conda_content_trust` plugin warning that is configured as an error by this repo.

conda/plugins/previews.py
```diff
@@ -0,0 +1,49 @@
...
+        return
+
+    for subcommand_hook in subcommand_hooks:
+        yield from subcommand_hook()
+
+
+def is_preview_subcommand(plugin_subcommand: CondaSubcommand) -> bool:
+    """
+    Determine whether this plugin subcommand is a bundled preview command.
+    """
```
kenodegard:
```suggestion
    """Return whether *plugin_subcommand* was yielded by ``conda.plugins.previews``.

    Checks ``plugin_subcommand.impl.plugin_name`` (hook provenance set by the plugin manager).
    """
```

conda/plugins/previews.py
```diff
@@ -0,0 +1,49 @@
...
+
+def _preview_subcommands(
+    label: str,
+    *subcommand_hooks: Callable[[], Iterable[CondaSubcommand]],
+) -> Iterable[CondaSubcommand]:
+    if not context.preview_enabled(label):
+        return
+
+    for subcommand_hook in subcommand_hooks:
+        yield from subcommand_hook()
```
kenodegard: This helper doesn't add much value:

```suggestion
    from .._preview.env_setup import PREVIEW_LABEL as ENV_SETUP_PREVIEW_LABEL

    if context.preview_enabled(ENV_SETUP_PREVIEW_LABEL):
        from .._preview.env_setup.cli import main_create, main_install

        yield from main_create.conda_subcommands()
        yield from main_install.conda_subcommands()
```

conda/plugins/subcommands/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+from .. import previews
 from . import doctor
 
-plugins = [doctor]
+plugins = [doctor, previews]
```
kenodegard: I don't think `previews` belongs in `conda.plugins.subcommands.plugins`. `previews` is separate infrastructure that aggregates opt-in preview hooks; it only implements `conda_subcommands` today, but the preview design isn't subcommand-specific.

Let us instead load it explicitly in `get_plugin_manager()`, same as solvers:

```
plugin_manager.load_plugins(
    solvers,
    previews,
    *subcommands.plugins,
    ...
)
```

and revert the change here:


```suggestion
from . import doctor

plugins = [doctor]
```

conda/cli/main.py
```diff
@@ -46,6 +47,20 @@ def main_subshell(*args, post_parse_hook=None, **kwargs):
...
+        module_name = f"conda._preview.{package}"
+        try:
+            mod = import_module(module_name)
+        except ModuleNotFoundError as err:
+            # Ignore unknown preview labels, but surface missing imports inside previews.
+            if err.name == module_name:
+                continue
+            raise
+        mod.register(context)
+
```
kenodegard: I think this is leftover scaffolding from before the `conda.plugins.previews` refactor? So this can be deleted along with the register placeholder in `conda._preview.env_setup`?

conda/cli/conda_argparse.py
```diff
@@ -286,6 +303,16 @@ def configure_parser_plugins(sub_parsers) -> None:
...
+            if (
+                is_preview_subcommand(plugin_subcommand)
+                and name in BUILTIN_COMMAND_PARSERS
+            ):
+                parser = BUILTIN_COMMAND_PARSERS[name](sub_parsers)
+                if plugin_subcommand.configure_parser:
+                    plugin_subcommand.configure_parser(parser)
+                parser.set_defaults(_plugin_subcommand=plugin_subcommand)
+                continue
+
```
kenodegard: This `continue` skips the shared plugin path below and duplicates parser setup. I suggest merging preview into that path: built-in parser first, then optional `configure_parser` as an extension.

```python
def configure_parser_plugins(sub_parsers) -> None:
    """
    For each of the provided plugin-based subcommands, we'll create
    a new subparser for an improved help printout and calling the
    :meth:`~conda.plugins.types.CondaSubcommand.configure_parser`
    with the newly created subcommand specific argument parser.
    """
    plugin_subcommands = context.plugin_manager.get_subcommands()
    for name, plugin_subcommand in plugin_subcommands.items():
        # if the name of the plugin-based subcommand overlaps a built-in
        # subcommand, we print an error
        preview = is_preview_subcommand(plugin_subcommand)
        if name in BUILTIN_COMMANDS and not preview:
            log.error(
                dals(
                    f"""
                    The plugin '{name}' is trying to override the built-in command
                    with the same name, which is not allowed.

                    Please uninstall the plugin to stop seeing this error message.
                    """
                )
            )
            continue

        # create the parser for the subcommand
        if preview and name in BUILTIN_COMMAND_PARSERS:
            parser = BUILTIN_COMMAND_PARSERS[name](sub_parsers)
        else:
            parser = sub_parsers.add_parser(
                name,
                description=plugin_subcommand.summary,
                help=plugin_subcommand.summary,
                add_help=False,  # defer to subcommand's help processing
            )

        # case 1: plugin extends the parser
        if plugin_subcommand.configure_parser:
            plugin_subcommand.configure_parser(parser)

            # attempt to add standard help processing, will fail if plugin defines their own
            try:
                add_parser_help(parser)
            except argparse.ArgumentError:
                pass

        # case 2: preview subcommand uses the built-in parser and are non-greedy
        elif preview:
            pass

        # case 3: plugin has their own parser, see _GreedySubParsersAction
        else:
            parser.greedy = True

        # underscore prefixed indicating this is not a normal argparse argument
        parser.set_defaults(_plugin_subcommand=plugin_subcommand)

```

docs/source/dev-guide/previews.md
kenodegard, line 1: Good docs

tests/_preview/env_setup/cli/test_main_create.py
```diff
@@ -0,0 +1,129 @@
...
+    """conda create with CONDA_PREVIEW=env-setup is routed to the stub."""
+    monkeypatch.setenv("CONDA_PREVIEW", "env-setup")
+
+    out, err, exc = conda_cli(
+        "create", "--name", "test-env", "python", raises=OperationNotAllowed
+    )
+
+    assert exc.value is not None
+    assert "env-setup" in str(exc.value)
+    assert "conda create" in str(exc.value)
```
kenodegard: When we're checking the error message looks a certain way `pytest.raises(match)` is usually much cleaner:

```suggestion
    with pytest.raises(OperationNotAllowed, match=r"'env-setup'.+'conda create'"):
        conda_cli("create", "--name", "test-env", "python")
```

kenodegard (changes requested):

conda/plugins/subcommands/__init__.py
jezdez: Oh, that's much better, you're right, this was from before when the PR was loading the specific preview create and install subcommands here. Doing it in `get_plugin_manager` is cleaner.

conda/cli/main.py
travishathaway: I think this is okay to remove. I'm not sure what we're going to use this register function for anyways, and if we need it later, we'll add it.

conda/cli/conda_argparse.py
travishathaway: Integrated this suggestion. Thanks!

jezdez:
pre-commit.ci autofix

jezdez (approved):

kenodegard (approved):

PR #16045, opened 2026-05-06
Adds more robust file handling for `--file` type options

### LINKED ISSUE #15898: `conda create --file <(process substitution)` silently creates empty environment in 26.3.x
### Checklist

### What happened?

After upgrading from conda 26.1.1 to 26.3.1, conda create --file silently ignores package specs when the file is provided via bash process substitution (<(...)). The environment is created with zero packages, exit code 0, and no error or warning.

The same input produces a fully populated environment on 26.1.1. This is a regression likely introduced by the environment specifier plugin support added in 26.3.0 (#15573 / #15770). The new plugin-based format detection may not handle /dev/fd/ paths (which have no file extension) correctly.

## Reproduction
```
# conda 26.3.1 — BROKEN: creates empty environment, exit 0, no error
conda create --name test-env --yes --file <(echo "python=3.13.12=hc97d973_100_cp313")
conda list -n test-env    # empty!
conda env remove -n test-env -y

# conda 26.1.1 — WORKS: installs python + all dependencies
conda create --name test-env --yes --file <(echo "python=3.13.12=hc97d973_100_cp313")
conda list -n test-env    # python + 23 packages
conda env remove -n test-env -y
```
Also reproduces with full conda list --export output:
```
conda create --name test-env --yes --file <(conda list python --full-name --export)
```

## Expected behavior
`conda create --file <(...)` should parse package specs from process substitution identically to a regular file, as it did in 26.1.1 and all prior versions.

## Real-world impact: breaks pyenv-virtualenv
[pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) uses exactly this pattern to create conda-based virtual environments ([source, line ~164](https://github.com/pyenv/pyenv-virtualenv/blob/v1.3.0/bin/pyenv-virtualenv)):
```
conda create $QUIET $VERBOSE --name "${VIRTUALENV_PATH##*/}" --yes \
  "${VIRTUALENV_OPTIONS[@]}" \
  --file <(conda list python --full-name --export)
```
On 26.3.x this silently creates a bare environment with no bin/ directory. pyenv-virtualenv then fails trying to symlink python-config into the missing bin/, its trap cleanup ERR handler fires and rm -rfs the entire environment. This completely breaks pyenv virtualenv for all conda/miniforge-based Python versions.

Any CI system or developer machine that runs conda update conda in a shared miniforge installation will break all subsequent pyenv virtualenv calls.

### Conda Info

```shell
active environment : None
            shell level : 0
       user config file : /home/jenkins/.condarc
 populated config files : /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest/.condarc
                          /home/jenkins/.condarc
          conda version : 26.3.1
         python version : 3.13.12.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=sapphirerapids
                          __conda=26.3.1=0
                          __glibc=2.34=0
                          __linux=6.1.163=0
                          __unix=0=0
       base environment : /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest  (writable)
      conda av data dir : /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest/etc/conda
  conda av metadata url : None
          package cache : /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest/pkgs
                          /home/jenkins/.conda/pkgs
       envs directories : /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest/envs
                          /home/jenkins/.conda/envs
    temporary directory : /tmp
               platform : linux-64
             user-agent : conda/26.3.1 requests/2.32.5 CPython/3.13.12 Linux/6.1.163-186.299.amzn2023.x86_64 amzn/2023.10.20260302 glibc/2.34 solver/libmamba conda-libmamba-solver/26.3.0 libmambapy/2.5.0
                UID:GID : 1002:1002
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> /home/jenkins/tachyon/tools/installed/pyenv/linux/x86-64/2.6.27/versions/miniforge3-latest/.condarc <==
channels:
  - conda-forge

==> /home/jenkins/.condarc <==
channels:
  - conda-forge
```

### Conda list

```shell

```

### Additional Context

_No response_

travishathaway:
### Description

fixes: #15898

This pull request implements a fix for the above issue, but it also attempts to add a more robust `argparse` type callable for parsing file type arguments. If it detects files coming from a pipe it will create a temporary file ending with `.txt` (default environment specification).

These checks are skipped for Windows because it can't handles these types of piped input.

### Checklist - did you ...


jezdez (commented):
Might this leave files around in case any of the file interactions fail? I thought we had a code path for this type of stuff?

travishathaway:
> Might this leave files around in case any of the file interactions fail? I thought we had a code path for this type of stuff?

I think the only possibilities for leaving files around is in situations where we send a SIGTERM (`kill`) to the running process. The atexit, according to the documentation, handles all Python exceptions and clean up after itself. This also includes "ctrl-c" because we intercept those and make sure the program exits in an organized manner.

If we wanted to be very conservative about cleaning up these files after receiving SIGTERM, we could use the `signals` module to register a callback, but I'm convinced this absolutely necessary.

Regarding existing ways we handle tempfiles in conda, I was only able to find [Utf8NamedTemporaryFile](https://github.com/conda/conda/blob/4a9cc34560bfa7af774703e74374bb41beb1a45f/conda/auxlib/compat.py#L24-L43), but I don't think it brings any advantages in this case, and there's noting in the documentation stating it should be preferred as the way to interact with temporary files.

PR #16057, opened 2026-05-08
shards A: Add implementation from conda-libmamba-solver

### LINKED ISSUE #15906: Host sharded repodata from conda-libmamba-solver

dholth:
### Description

Part of #15906. Ports the sharded repodata implementation from conda-libmamba-solver into conda core.

- Add `conda/_private/shards/` package with the core sharding implementation (ShardBase, ShardLike, Shards classes)
- Add SQLite-backed shard cache (`cache.py`) for individual per-package shards
- Add pipelined and BFS traversal algorithms (`subset.py`) for dependency-driven repodata subsetting
- Add `QueueCache` to serialize SQLite writes through the cache thread (avoids lock contention)
- Add type definitions (`typing.py`) for shard data structures
- Add HTTP fetching with ETag/Last-Modified caching for the shards index
- Add `repodata_shards.msgpack.zst` index fetching integrated with `RepodataCache`
- Add `msgpack` as a new runtime dependency
- Add initial test suite with local HTTP server fixtures

### Checklist - did you ...

### DIFF
--- a/conda/_private/shards/__init__.py
+++ b/conda/_private/shards/__init__.py
@@ -1,0 +1,20 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+from . import cache, subset
+from .shards import (
+    ShardBase,
+    ShardLike,
+    Shards,
+    fetch_channels,
+    fetch_shards_index,
+)
+
+__all__ = [
+    "cache",
+    "subset",
+    "ShardBase",
+    "ShardLike",
+    "Shards",
+    "fetch_shards_index",
+    "fetch_channels",
+]
--- a/conda/_private/shards/shards.py
+++ b/conda/_private/shards/shards.py
@@ -1,0 +1,827 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Models for sharded repodata, and to make monolithic repodata look like sharded
+repodata.
+"""
+
+from __future__ import annotations
+
+import abc
+import concurrent.futures
+import functools
+import logging
+from collections import defaultdict
+from typing import TYPE_CHECKING
+from urllib.parse import urljoin, urlparse, urlunparse, uses_relative
+
+import msgpack
+import zstandard
+from libmambapy.bindings import specs
+
+import conda.exceptions
+from conda.base.context import context
+from conda.common.serialize import json
+from conda.core.subdir_data import SubdirData
+from conda.gateways.connection.session import get_session
+from conda.gateways.repodata import (
+    CACHE_CONTROL_KEY,
+    ETAG_KEY,
+    LAST_MODIFIED_KEY,
+    URL_KEY,
+    RepodataIsEmpty,
+    UnavailableInvalidChannel,
+    _add_http_value_to_dict,
+    conda_http_errors,
+)
+from conda.models.channel import Channel
+
+from . import cache
+
+log = logging.getLogger(__name__)
+
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, KeysView
+
+    from requests import Response
+
+    from conda.gateways.repodata import RepodataCache
+
+    from .typing import PackageRecordDict, RepodataDict, ShardDict, ShardsIndexDict
+
+SHARDS_CONNECTIONS_DEFAULT = 10
+ZSTD_MAX_SHARD_SIZE = (
+    2**20 * 16
+)  # maximum size necessary when compressed data has no size header
+
+# Schemes that urljoin handles correctly (registered in urllib.parse.uses_relative)
+_URLJOIN_SAFE_SCHEMES = frozenset(uses_relative)
+
+
+def _safe_urljoin_with_slash(base_url: str, relative_url: str = "") -> str:
+    """
+    Join base_url with relative_url, ensuring proper handling of all URL schemes.
+
+    Python's urllib.parse.urljoin only handles schemes registered in
+    ``urllib.parse.uses_relative``. For unregistered schemes like ``s3://``,
+    it returns just ``"."`` instead of the resolved URL. This function falls
+    back to a scheme-swap workaround for those cases.
+
+    The result always ends with "/" to enable proper string concatenation with filenames.
+
+    See: https://github.com/conda/conda-libmamba-solver/issues/866
+    """
+    parsed = urlparse(base_url)
+
+    # For schemes that urljoin handles correctly, use the standard behavior
+    if parsed.scheme in _URLJOIN_SAFE_SCHEMES:
+        # Standard urljoin behavior: join with relative_url, then "." for trailing slash
+        result = urljoin(urljoin(base_url, relative_url), ".")
+        return result
+
+    # For unregistered schemes (e.g. s3://), urljoin drops the host.
+    # Work around that by temporarily swapping in https://, then restoring
+    # the original scheme on the result.
+    relative_parsed = urlparse(relative_url)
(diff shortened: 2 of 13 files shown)


_conda/shards/misc.py
jezdez, line 1: This clearly was supposed to be used by other modules, but all of the functions are also still in the other modules, either we remove `misc.py` or update the other files to import from here.

conda/_private/shards/subset.py
```diff
@@ -0,0 +1,806 @@
...
+    from .cache import ShardCache
+    from .shards import ShardBase
+    from .typing import ShardDict
+
+# Waiting for worker threads to shutdown cleanly, or raise error.
+THREAD_WAIT_TIMEOUT = 5  # seconds
+REACHABLE_PIPELINED_MAX_TIMEOUTS = (
+    10  # number of times we can timeout waiting for shards
+)
+QUEUE_TIMEOUT = 1
```
jezdez, line 101: This is 5 in CLS, why the reduction?

conda/_private/shards/shards.py
```diff
@@ -0,0 +1,824 @@
...
+import functools
+import json  # noqa
+import logging
+from collections import defaultdict
+from typing import TYPE_CHECKING
+from urllib.parse import urljoin, urlparse, urlunparse, uses_relative
+
+import msgpack
+import zstandard
+from libmambapy.bindings import specs
```
jezdez, line 20: That's a problem IMO since it would make conda directly rely on libmambapy instead of separate concerns.

conda/_private/shards/shards.py
```diff
@@ -0,0 +1,824 @@
...
+    in offline mode, raises RepodataIsEmpty to signal unavailability.
+    """
+    # In offline mode, return cached data if available, even if expired
+    if context.offline:
+        if cache.cache_path_shards.exists():
+            return cache.cache_path_shards.read_bytes()
+        else:
+            # In offline mode with no cache, signal that shards are not available.
+            # The caller (fetch_shards_index) catches RepodataIsEmpty and falls back to non-sharded repodata.
+            raise conda.gateways.repodata.RepodataIsEmpty(
```
jezdez: I'm not sure if we should use `RepodataIsEmpty` this way, we usually import exceptions directly instead of calling them via the module interface.

conda/_private/shards/shards.py
```diff
@@ -0,0 +1,824 @@
...
+Models for sharded repodata, and to make monolithic repodata look like sharded
+repodata.
+"""
+
+from __future__ import annotations
+
+import abc
+import concurrent.futures
+import functools
+import json  # noqa
```
jezdez: This should import from `conda.common.serialize.json`

conda/_private/shards/__init__.py
jezdez, line 1: I would prefer if we introduced the `_conda` namespace separately and not in a last-minute feature PR like this one. Do we _actually_ need this new module for the shards to exist?

The PR also seems to contradict the rules as defind in the README, heavily importing from conda. I would think this would be better located in the gateways package.

jezdez (commented):
A few things worth looking into it, but largely no surprises

conda/_private/shards/__init__.py
dholth: "Submodules of `_conda` must have the _option_ of running without importing anything from `conda`" is meant to allow `_conda.fast_subinterpreter/fast_subprocess`, but it is not a requirement that the submodules avoid `conda`.

_conda/shards/misc.py
dholth: May have been lost while updating with the conda-libmamba-solver upstream, will take a look. Lean towards removal.

conda/_private/shards/shards.py
dholth: Was concerned about overhead from `conda.common.serialize.json` which IIUC will [call its `default()` method](https://github.com/conda/conda/blob/main/conda/common/serialize/json.py#L22) for every value in the json to deal with `auxlib.entity` types and others; but we know we have none of those types and can use the fast default encoder.
It would matter if we were to `dumps()` an entire `repodata.json` but from what I can tell we don't do that in gateways/repodata.
The latest version of this module only uses `except JSONDecodeError` and delegates all loads/dumps to repodata gateway.

conda/_private/shards/shards.py
dholth: Yes, it needs improved fallbacks if the solver doesn't pass `spec_to_package_name_func: Callable[[str], str] = spec_to_package_name` to `build_repodata_subset()`.

conda/_private/shards/shards.py
dholth: Do you mean the difference between `from conda.gateways.repodata import RepodataIsEmpty` instead of `import conda.gateways.repodata`?

conda/_private/shards/shards.py
```diff
@@ -0,0 +1,824 @@
...
+    # will raise the exception.
+    # repo_cache.load_state(
+    #     binary=True
+    # )  # won't succeed when .msgpack.zst is missing as it wants to compare the timestamp (returns empty state)
+
+    # Load state ourselves to avoid clearing when binary cached data is missing.
+    # If we fall back to monolithic repodata.json, the standard fetch code will
+    # load the state again in text mode.
+    try:
+        with repo_cache.lock("r+") as state_file:
```
dholth, line 649: This `json.loads()` etc. is moved into gateways/repodata [in another PR](https://github.com/conda/conda/pull/16069/changes)

conda/_private/shards/subset.py
dholth: The code immediately sets the number of allowed timeouts to `context.remote_read_timeout_secs * (context.remote_max_retries + 1)`

conda/_private/shards/subset.py
```diff
@@ -0,0 +1,806 @@
...
+We treat all repodata as sharded, even if no actual sharded repodata has been
+found.
+
+## Example usage
+
+The following constructs several repodata (`noarch` and `linux-64`) from a
+single channel name and a list of root packages:
+
+``` from conda.models.channel import Channel from
+_conda.shards_subset import build_repodata_subset
```
ryanskeith, line 38: Just a quick note here to update this sample code when we solidify where all the code goes.

ryanskeith (commented):
This is more of a re-review.  I think it looks good to me as a start.  I am worried about the import from libmamba but we should be able to resovle that quickly.

conda/_private/shards/subset.py
dholth: Good idea. We could also port over the shards docs from c-l-s.

About the import from libmamba, conda happens to depend on libmamba already. For completeness we could write an implementation that uses Python MatchSpec or a regex that can extract the name from a matchspec, but expect the solver to provide its own spec-to-package-name function during all practical use.

jezdez: I went ahead and fixed this in E (7000ae21a). Replaced with conda's own MatchSpec, which benchmarks as equivalent when cached (~10ms for 5000 unique specs either way). Importing libmambapy directly is not okay since it's a transitive dependency via conda-libmamba-solver, not a direct conda dependency.

jezdez (approved):

conda/_private/shards/subset.py
dholth: I did my own benchmark with all of `conda-forge/linux-64`, it comes out to 

MatchSpec (not cached) 9.77s
libmamba (not cached) 6.91s
MatchSpec (cached) 0.58s
libmamba (cached) 0.49s

`@functools.cache` does the trick.

PR #16069, opened 2026-05-11
Add check_mtime flag to RepodataCache to avoid manual lock() in shards

dholth:
### Description

Add a `check_mtime` parameter to `RepodataCache.load()` so the shards code can read cached data without the mtime validation that would otherwise clear etag/last-modified headers.

- Add `check_mtime=True` parameter to `RepodataCache.load()` (default preserves existing behavior)
- When `check_mtime=False`, skip the stat/mtime validation (shards manage cache validity via etag, not filesystem mtime)
- When `state_only=True`, tolerate missing data files during stat (normal for shards where only one of binary/text exists)
- Remove manual `cache.lock("r+")` calls from `shards.py` in favor of `cache.load(binary=True, check_mtime=False)`
- Simplify `fetch_shards_index()` state loading to use `cache.load(state_only=True, check_mtime=False)`
- Fix flaky `test_pipelined_concurrent_stress` by giving each thread its own cache directory
- Add `test_repodata_state_check_mtime` covering the new parameter

We will revisit this with a different strategy, possibly just storing separate etag for separate URL's which would make more sense. I think it's useful to share the single info file, since all the URLs represent the same repodata conceptually.

### Checklist - did you ...


jezdez (approved):

tests/shards/test_shards_subset.py
```diff
@@ -951,13 +951,29 @@ def test_pipelined_concurrent_stress(http_server_shards, mocker, tmp_path, num_t
     Run pipelined algorithm from multiple threads concurrently. This can expose
     race conditions in shared state or thread coordination.
 
-    (Actually the concurrency issues happen in fetch_channels() which deals with
-    reading, writing repodata_shards.msgpack.zst to disk.)
```
dholth, line 1224: I think this test is how we found the missing `lock(state file)` issue which would not be found with this change. We should consider removing this test.

A more focused set of tests would be able to deadlock or measure sqlite3 contention; another test to simulate what happens when two conda's are using the same cache directory; another test that looks inside the algorithm to check whether its parts share anything or not.

dholth:
What if we put this change on hold since the shards cache works as-is. Instead, we should fix the etag problem by storing separate cache headers per format. This part of the code (RepodataCache()) is tricky, fragile especially because it must deal with locking.

Here is an example stat file `.info.json` with shards. (the url containing the full filename is nice, but possibly accidental due to the way Channel() works)
```
{
  "etag": "\"e312818c32200d241c2ba6a9d08567ec\"",
  "mod": "Thu, 07 May 2026 15:14:19 GMT",
  "url": "https://conda.anaconda.org/conda-forge/osx-arm64/repodata_shards.msgpack.zst",
  "cache_control": "public, max-age=1200",
  "has_shards": {
    "last_checked": "2026-05-07T15:49:34.552874Z",
    "value": true
  },
  "mtime_ns": 1778168974553596672,
  "size": 251586,
  "refresh_ns": 1778168974554279000
}
```

Here is one without shards
```
{
  "etag": "\"9f0a4face99a52fbb36d441f0e212db9\"",
  "mod": "Mon, 11 May 2026 21:05:32 GMT",
  "has_shards": {
    "last_checked": "2026-05-07T11:40:46.905318Z",
    "value": false
  },
  "refresh_ns": 1778535129994887000,
  "cache_control": "public, max-age=30",
  "size": 63329366,
  "url": "https://repo.anaconda.com/pkgs/main/osx-arm64",
  "mtime_ns": 1778535129834166614
}
```

jezdez:
This sadly needs a major rebase

dholth:
I will probably wind up deleting this and create an issue for "store separate cache headers per format"

PR #16077, opened 2026-05-12
shards: update network thread test

### LINKED ISSUE #15906: Host sharded repodata from conda-libmamba-solver

dholth:
### Description

Update network threads test.

The short timeout is useful in the first part of the test because we pull from the output queue until it's empty; and we don't want to wait 1 second. Alternatively we could submit the shards and a None so that this thread would terminate on its own.

The short timeout isn't useful in the second part of the test and was causing failure. Instead, use a 1 second timeout. We will always get an item from the queue instead of timing out if the test passes.

Added a third part of the test to generate `KeyError` in the thread.

Fix #15906 

### Checklist - did you ...

### DIFF
--- a/tests/shards/test_shards_subset.py
+++ b/tests/shards/test_shards_subset.py
@@ -491,7 +491,9 @@ def test_shards_network_thread(http_server_shards, shard_cache_with_data, monkey
     Test network retrieval thread, meant to be chained after the sqlite3 thread
     by having network_in_queue = sqlite3 thread's network_out_queue.
     """
-    monkeypatch.setattr(shards_subset, "QUEUE_TIMEOUT", 0.01)
+    # Timeout allows us to finish `with suppress(Empty)` loop more quickly, but
+    # a too-short timeout can cause failures.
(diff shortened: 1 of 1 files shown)


tests/shards/test_shards_subset.py
danyeaw:
```suggestion
    network_thread.join(1)
    assert not network_thread.is_alive()
```

What if we set this to a small timeout? I think setting it to zero means a complete no-op, since it would just immediately return None.

danyeaw (approved):
Hey @dholth, nice update, I think increasing the timeout will fix the two blocking `get()` calls. Also I like the extra assertions you added.

tests/shards/test_shards_subset.py
dholth: Oh, `As join() always returns None, you must call [is_alive()](https://docs.python.org/3/library/threading.html#threading.Thread.is_alive) after join() to decide whether a timeout happened – if the thread is still alive, the join() call timed out.`

So our existing `join(timeout)` calls are harmless but useless. We don't actually need to wait for the thread to exit because we get all results from the queue.

PR #16078, opened 2026-05-12
Add v3 repodata sharding

### LINKED ISSUE #16067: shards: v3 repodata shard support for conda-pypi channels
#### What is the idea?
Support v3 repodata groups (e.g. v3.whl) during sharded repodata graph traversal and assembly, enabling channels like conda-pypi-redirect to serve wheel packages via shards.

### Why is this needed?
The shards implementation in the PRs #16057, #16058, and #16059, has shards support at `_conda/shards/shards.py` and currently only processes `packages` and `packages.conda` groups. The conda-pypi beta channel will use the v3 schema that stores wheel-translated package records under a top-level `v3` key (e.g. `v3.whl`). Without this support, those packages are never discovered during dependency graph traversal and are excluded from the assembled repodata, making them invisible to the solver.

### What should happen?
`shard_mentioned_packages` should traverse depends entries in v3 group records so their transitive dependencies are discovered. `ShardBase.build_repodata()` should merge v3 group records into the output repodata dict alongside the standard `packages` and `packages.conda` groups.

### Additional Context
There is a hacky branch at https://github.com/conda/conda/tree/refactor-sharded2 with this working.

danyeaw:
### Description

Closes #16067 by extracting dependency names and package records found in v3 repodata with `shard_mentioned_packages` and `iter_records`.

In order to support sharding for extra_depends, we would need to track a new node on each shard. So right now they are skipped.

### Checklist - did you ...


conda/_private/shards/shards.py
```diff
@@ -221,22 +220,26 @@ def shard_mentioned_packages(
...
-            group.values() for group in shard.get("v3", {}).values()
-        ),
+    all_records = (
+        record
+        for records in (
+            shard["packages"].values(),
+            shard["packages.conda"].values(),
+            *(
+                group.values() for group in shard.get("v3", {}).values()
+            ),  # v3 repodata groups
```
dholth: this eagerly extracts the groups and lazily extracts the values from those groups?

danyeaw: Yup, that's right, since Python must consume the generator to build the tuple. However, `dict.values()` is O(1) and returns a lightweight view. We could do full laziness with `itertools.chain.from_iterable`, or we could use a nested generator without the star, but I think both of those would be less readable.

dholth: Works for me as long as we don't have to build the entire list of packages here.

tests/shards/test_shards.py
```diff
@@ -790,6 +925,44 @@ def test_shardlike():
...
+    shardlike = ShardLike(
+        {"packages": {}, "packages.conda": {}, "info": {"base_url": ""}}
+    )
+    shardlike.visit_shard(
+        "mypkg",
+        {
+            "packages": {},
+            "packages.conda": {"mypkg-1.0-0.conda": {"name": "mypkg"}},
+            "v3": {
+                "whl": {"mypkg-1.0-py312-none-any.whl": {"name": "mypkg"}},
```
dholth: "v3" doesn't use extension inside each format section

```
        "conda": {
            "zstandard-0.25.0-py314hb3cf4b8_0": {
                ...
            },
            "zstd-1.5.7-h817c040_0": {
                ...
            }
        },
        "tar.bz2": {},
        "whl": {}
```

`python -m conda_index --repodata-next <channel dir>`

dholth (commented):
We need a way to skip v3 if the solver doesn't support it.
Some of the v3 repodata in the test is incorrect.
Would this be a convenient place to remove the `build_repodata_subset(spec_to_package_name_func: Callable[[str], str] | None = None,)` parameter (a way for the solver to pass its own) because our default implementation is fast enough?

tests/shards/test_shards.py
danyeaw: Thanks, updated the examples!

conda/_private/shards/subset.py
```diff
@@ -563,15 +570,17 @@ def build_repodata_subset(
         algorithm: desired traversal algorithm ("bfs" or "pipelined")
         spec_to_package_name_func: callable to convert package specs to names.
                                    Defaults to the standard spec_to_package_name.
-
+        repodata_version: repodata format version (1 = classic, 3 = v3).
```
dholth, line 573: 2 is with-base-url. https://github.com/conda/ceps/blob/main/cep-0015.md not that we would bother to support "repodata but it can't have a base_url"

tests/shards/test_shards.py
```diff
@@ -506,6 +506,150 @@ def test_shard_mentioned_packages_2():
...
+    assert names.count("openssl") == 1
+
+
+def test_shard_mentioned_packages_v3_ensures_hex_hash(mocker):
+    # spy on the name as bound inside shards.py (imported by-name at module level)
+    spy = mocker.spy(shards, "ensure_hex_hash")
+    record = {"sha256": b"\xde\xad\xbe\xef" * 8, "depends": ["zlib >=1.2"]}
+    shard = _v3_shard({"whl": {"pkg-1.0-py3_none_any_0": record}})
+    list(shard_mentioned_packages(shard, repodata_version=3))
+    spy.assert_called()
```
dholth, line 573: 🕴

conda/_private/shards/shards.py
```diff
@@ -479,9 +493,13 @@ def shard_url(self, package: str) -> str:
 
         Raise KeyError if package is not in the index.
         """
+        if cached := self._shard_url_cache.get(package):
```
dholth, line 496: Is this necessary? I'll let it pass for now.

news/16078-add-v3-repodata-sharding
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Extract dependency names and package records from v3 repodata groups in `shard_mentioned_packages` and `iter_records`. (#16067 via #16078)
```
dholth:
```suggestion
* Sharded repodata: extract dependency names and package records from v3 repodata groups in `shard_mentioned_packages` and `iter_records` for compatible solvers. (#16067 via #16078)
```

dholth (approved):

PR #16079, opened 2026-05-12
Ensure that `build_repodata_subset` partial is only applied to a solver's `__init__` function

soapy1:
### Description

Currently using the main branch of conda, the conda-rattler-solver [tests fail](https://github.com/conda-incubator/conda-rattler-solver/actions/runs/25753870477/job/75637093475?pr=53) with 
```
FAILED tests/test_user_agent.py::test_user_agent_conda_info[rattler] - AssertionError: assert 'conda-rattler-solver/0.0.7.dev10+g638ac7e6f' in 'conda/26.3.3.dev81 requests/2.34.0 CPython/3.10.20 Linux/6.17.0-1010-azure ubuntu/24.04.4 glibc/2.39 solver/rattler'
```

Inspecting conda, we can confirm that the main branch does not include the conda-rattler-solver version in the user agent.

```
$ conda info --json | jq .user_agent                                                                                                              
"conda/26.3.3.dev81 requests/2.33.1 CPython/3.10.20 Linux/6.17.4-76061704-generic pop/22.04 glibc/2.35 solver/rattler"
```

Previously it did (from commit 5a2724b95)
```
$ conda info --json | jq .user_agent                                                                                                                                                 
"conda/26.3.3.dev81 requests/2.33.1 CPython/3.10.20 Linux/6.17.4-76061704-generic pop/22.04 glibc/2.35 solver/rattler conda-rattler-solver/0.0.7.dev9+g46dc8bb7d py-rattler/0.23.2"
```

The root cause is conda getting caught in this exception in [`context.solver_user_agent`](https://github.com/conda/conda/blob/main/conda/base/context.py#L1113):
```
AttributeError("'functools.partial' object has no attribute 'user_agent'")
```

### Checklist - did you ...

- ~[ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -444,8 +444,14 @@ class CondaPluginManager(pluggy.PluginManager):
         ):
             from ..gateways.shards import build_repodata_subset
 
-            return functools.partial(
-                solver_plugin.backend, build_repodata_subset=build_repodata_subset
+            new_init = functools.partialmethod(
+                solver_plugin.backend.__init__,
+                build_repodata_subset=build_repodata_subset,
+            )
+            return type(
+                f"Partial{solver_plugin.backend.__name__}",
+                (solver_plugin.backend,),
+                {"__init__": new_init},
             )
 
         return solver_plugin.backend
--- a/tests/plugins/test_manager.py
+++ b/tests/plugins/test_manager.py
@@ -14,6 +14,7 @@ import pytest
 from packaging.version import Version
 
 from conda import plugins
+from conda.base.context import reset_context
 from conda.common.url import urlparse
 from conda.core import solve
 from conda.exceptions import CondaValueError, PluginError
@@ -24,6 +25,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterator
     from typing import Any
 
+    from pytest import MonkeyPatch
     from pytest_mock import MockerFixture
 
     from conda.plugins.manager import CondaPluginManager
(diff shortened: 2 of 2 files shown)


conda/plugins/manager.py
```diff
@@ -444,8 +444,14 @@ def get_solver_backend(self, name: str | None = None) -> type[Solver]:
...
-            return functools.partial(
-                solver_plugin.backend, build_repodata_subset=build_repodata_subset
+            new_init = functools.partialmethod(
+                solver_plugin.backend.__init__,
+                build_repodata_subset=build_repodata_subset,
+            )
+            return type(
+                f"Partial{solver_plugin.backend.__name__}",
+                (solver_plugin.backend,),
+                {"__init__": new_init},
```
dholth, line 454: Interesting solution. In that case, it would also be possible to
```
class Solver(solver_plugin.backend):
    def __init__(*args, build_repodata_subset=build_repodata_subset, **kwargs):
        return super().__init__(*args, build_repodata_subset=build_repodata_subset, **kwargs)
```

soapy1: Yep, that's true! Would you prefer to go with your alternative approach?

tests/plugins/test_manager.py
```diff
@@ -170,6 +172,45 @@ def test_known_solver(plugin_manager: CondaPluginManager):
...
+        name="test-classic",
+        backend=TestSolver,
+    )
+
+    class TestSolverPlugin:
+        @plugins.hookimpl
+        def conda_solvers(*args):
+            yield TestCondaSolver
+
+    monkeypatch.setenv("CONDA_REPODATA_USE_SHARDS", use_shards)
```
danyeaw:
```suggestion
    monkeypatch.setenv("CONDA_REPODATA_USE_SHARDS", str(use_shards))
```

danyeaw (approved):
Thanks @soapy1 so much for the fix!

conda/plugins/manager.py
dholth: This is OK, we'll discuss whether we should have a separate "build solver" and "get solver class" API instead in the future.

PR #16082, opened 2026-05-13
recipe/meta.yaml: Remove conda-pypi dependency

jsmolic:
Recently added dependency on conda-pypi could cause breakages in CI pipelines using `pip install` due to the externally-managed file (xref https://github.com/conda/conda-build/pull/5975). Although this has since been changed to [display a warning](https://github.com/conda/conda-pypi/pull/352) instead of erroring out, there is no release of conda-pypi that includes this. Temporarily remove conda-pypi from recipe dependencies to avoid breakages in CI pipelines.

### Checklist - did you ...

### DIFF
--- a/news/16082-remove-conda-pypi-dependency
+++ b/news/16082-remove-conda-pypi-dependency
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Remove `conda-pypi` dependency from the conda recipe to prevent downstream CI breakages. (#16082)
+
+### Deprecations
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Thanks @jsmolic!

danyeaw:
@kenodegard, FYI about this one. We need to release conda 26.5.0 without this dependency since it will break anyone using pip. Then we need to release conda-pypi 0.9.0 that includes this change https://github.com/conda/conda-pypi/pull/352 and https://github.com/conda/conda-pypi/pull/359, and then we'll add the dependency back.

PR #16093, opened 2026-05-14
Add auxlib collection typing and docstring coverage

adminlip:
This adds type annotations to the small helpers in conda.auxlib.collection and documents the last() helper so the module has clearer reference information while preserving the existing behavior. The change is intentionally limited to one auxlib module for issue #13437.

### DIFF
--- a/conda/auxlib/collection.py
+++ b/conda/auxlib/collection.py
@@ -1,11 +1,20 @@
 """Common collection classes."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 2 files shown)


danyeaw:
Hi @adminlip, thanks for the contribution! We are in the process of trying to remove usage of auxlib so I don't think enhancing it is needed. If you would like to working on improving our typing and auto docs, I would recommend that you start somewhere else. Thanks!

Issue #16083, opened 2026-05-13
Follow-on work post shards merge


dholth:
Improve sharded repodata code post-merge.

## Canary build fix -- merge order

1. [x] conda-incubator/conda-rattler-solver#57 -- Fix Windows canary build test syntax
2. [x] #16097 -- CI: upgrade c-l-s in base before test deps install (workaround for miniforge shipping c-l-s 26.4.0 with sqlite locking bug)
3. [x] conda-incubator/conda-rattler-solver#58 -- CI: same c-l-s upgrade workaround for conda-rattler-solver (**skip if step 7 lands first**)
4. [x] #16098 -- Skip concurrent sqlite multi-writer test on macOS Intel (prevents segfault from killing pytest)
5. [x] #16070 -- Fix notices cache mtime rounding (flaky \`test_clear_cache_invalidates_notices_cache\`)
6. [x] #16095 -- Fix v3 MatchSpec parser regression on PyPI-style extras in constrains (e.g. \`dask[array]\`)
7. [x] conda-forge/miniforge#878 -- Bump miniforge to c-l-s 26.4.2 (permanent fix, then remove workarounds from steps 2-3)

Steps 1-6 unblock the canaries. Step 7 is the permanent upstream fix.

## Other follow-on work

 - [ ] Documentation
 - [ ] Refactoring
 - [ ] If a .tar.bz2 and equivalent .conda are in v3 repodata, do we remove the .tar.bz2 like we do for packages/packages.conda
 - [ ] Should we add a separate "build solver instance" function in the plugin manager that passes build_repodata_subset plus the user's options; and a "get solver class" function which is currently only [used to enrich the user agent?](https://github.com/conda/conda/blob/82563a099a2266953eb8c58a91f430773d515a4b/conda/base/context.py#L1107-L1119)
 - [ ] https://github.com/conda/conda/pull/16069 "improve cache API". We want to take a different approach than in the linked PR, by storing separate cache keys for each repodata format (.json.zst, shards)
 - [x] #16087 -- Validate \`MatchSpec\` package names against CEP-26 -- #16096
 - [ ] Add support for parsing \`extra_depends\` in v3 Repodata, which will require tracking an extra node on each shard to track extras.
[...]

PR #16095, opened 2026-05-14
Fix v3 MatchSpec parser regression on PyPI-style extras in constrains

jezdez:
### Description

Fixes the conda canary build failures that started on May 13 after #15443 landed.

#### How it was found

The canary builds were failing because the Tests workflow couldn't pass. Tracing through the [test logs](https://github.com/conda/conda/actions/runs/25861311278/job/75991692721), `InvalidSpec` crashes were originating from `MatchSpec("dask[array]")` in the classic solver's `ReducedIndex._derive_reduced_index()` (index.py:548).

`dask[array]` is not a conda package - it's PyPI extras notation. But five packages on conda-forge/noarch use it in their `constrains` field:

```
anndata-0.12.{1,2}:   constrains: dask[array] >=2023.5.1,!=2024.8.*,!=2024.9.*,<2025.2.0
squidpy-1.{4.1,5.0}:  constrains: napari[pyqt5] ==0.4.15
awswrangler-3.16.0:    constrains: pyiceberg[pyarrow] >=0.7.0,<1
```

The v2 parser handles these silently (strips the brackets, keeps the name). The v3 parser crashes because it finds no `key=value` pairs inside the brackets.

#### Root cause

Two bugs in the v3 MatchSpec parser introduced in #15443:

1. **`_BRACKETS_RE_V3` regex anchor** (`^.*?(\[.*\])(?:\(.+\))?$`): The `$` anchor requires brackets at end-of-string. When parsing a full constrains string like `dask[array] >=2023.5.1,...`, the regex doesn't match (version spec follows the brackets), so the brackets are left in the package name. Later, when the classic solver re-parses just the extracted name `dask[array]`, the regex *does* match - but finds no `key=value` pairs and raises `InvalidSpec`.

2. **`PYTEST_CURRENT_TEST` dispatcher override**: `_parse_spec_str_dispatcher` used `os.environ.get("PYTEST_CURRENT_TEST")` to force the v3 parser for **all** test runs, including classic solver tests. This meant the classic solver's `ReducedIndex` code path hit the v3 parser in CI, triggering bug #1. It also silently prevented the test suite from exercising the v2 parser at all.

The v2 regex (`.*(?:(\[.*\]))`) doesn't have the `$` anchor, so the classic solver never had this problem before.

#### Fix

- Drop the `$` anchor from `_BRACKETS_RE_V3` so brackets match even when version specs follow
- Remove `PYTEST_CURRENT_TEST` from the dispatcher - the v3 parser now only activates for `context.solver == "rattler"`, matching production behavior
- Add a `rattler_solver` pytest fixture that explicitly sets `CONDA_SOLVER=rattler` for v3-specific tests, so they properly opt into the v3 parser without affecting classic solver test coverage

Relates to #16083.

### Checklist - did you ...


jaimergp:
> But five packages on conda-forge/noarch use it in their constrains field

These need to be patched out in conda-forge; that syntax, while not a problem before, is not valid.

jezdez:
> > But five packages on conda-forge/noarch use it in their constrains field
> 
> These need to be patched out in conda-forge; that syntax, while not a problem before, is not valid.

Makes sense to me, what process would we need to follow?

conda/models/match_spec.py
```diff
@@ -63,7 +62,7 @@
     re.VERBOSE,
 )
 # These _V3 are for new matchspecs in repodata v3
-_BRACKETS_RE_V3: re.Pattern[str] = re.compile(r"^.*?(\[.*\])(?:\(.+\))?$")
+_BRACKETS_RE_V3: re.Pattern[str] = re.compile(r"^.*?(\[.*\])")
```
jaimergp, line 65: Note that this was added to cover your request in https://github.com/conda/conda/pull/15443#pullrequestreview-4253987312

news/16083-fix-matchspec-v3-parser-regression
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix v3 MatchSpec parser regression where PyPI-style extras in `constrains` (e.g. `dask[array]`) crash the classic solver due to a regex anchor bug and a test-only dispatcher override. (#16083)
```
jaimergp: I think you can just add the PR number to the previous news item.

jezdez: Done - moved the entry to `news/15443-conditionals` under Bug fixes with `(#15443 via #16095)`.

conda/models/match_spec.py
jezdez: Right, I remember requesting the `[a][b]` edge case fix. The `$` anchor was your solution to prevent greedy matching across bracket groups — but it introduced the `dask[array] >=2023.5.1` regression where version specs after brackets cause the regex to fail entirely.

Dropping `$` is safe because the `[a][b]` case is already caught downstream at line 1056: `if "][" in value: raise InvalidSpec("Multiple bracket sections are not allowed")`. Verified this still works — `package[build=0][version=2]` correctly raises `InvalidMatchSpec`.

The v2 regex `_BRACKETS_RE` (`.*(?:(\[.*\]))`) also has no `$` anchor and handles the same patterns fine.

tests/models/test_match_spec.py
```diff
@@ -21,6 +22,16 @@
...
 
 
+@pytest.fixture
+def rattler_solver(monkeypatch):
+    """Activate the v3 MatchSpec parser by setting the solver to rattler."""
+    monkeypatch.setenv("CONDA_SOLVER", "rattler")
+    reset_context()
+    _PARSE_CACHE.clear()
+    yield
+    _PARSE_CACHE.clear()
```
kenodegard, line 30: We already have `parametrized_solver_fixture` and `solver_classic`/`solver_libmamba` fixtures, we should add this there

kenodegard: or since these tests are not solver tests should we rename this fixture?

`match_spec_v3` or similar?

jezdez: Good call - renaming to `match_spec_v3` and using `_solver_helper` for consistency.

jezdez: Done

kenodegard (approved):

PR #16096, opened 2026-05-14
Validate MatchSpec package names against CEP-26

### LINKED ISSUE #16087: `MatchSpec` silently accepts invalid package names containing quote characters, violating CEP-26
### Checklist

### What happened?

[cep26]: https://conda.org/learn/ceps/cep-0026

When a `MatchSpec` string is constructed with a quoted package name, the behavior is inconsistent and violates [CEP-26][cep26]:

```python
>>> MatchSpec("'package'").get("name")
"'package'"        # quotes silently become part of the name

>>> MatchSpec("'package==3'")
# fails, but with a confusing error rather than "quotes not valid in name"
```

In the first case, `'package'` is accepted without error and the single quotes are retained as literal characters in the name. In the second case, the parser fails but for the wrong reason, the quoting causes the version constraint to be split incorrectly rather than the input being rejected outright as an invalid name.

#### What should have happened instead?

[CEP-26][cep26] specifies that distributable package names MUST match:

```
^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$
```

Quote characters are not in the allowed set. Per [CEP-26][cep26]'s RFC 2119 language, violations of a MUST rule MUST result in an error. Both inputs above should raise `InvalidMatchSpec` with a message indicating that the name contains characters not permitted by [CEP-26][cep26].

#### Why this matters now

This bug has become more visible with the addition of the `when` key to `MatchSpec` (conda/conda#15443). The conditional dependencies CEP (conda/ceps#164) specifies that the `when` value MUST be quoted. A user who accidentally quotes an inner spec name instead:

```
pkg[when="'__unix'"]
```

should get a clear, actionable error. Instead they currently get either silent misparse or a confusing `InvalidMatchSpec` rooted in this same quote-retention behavior. Note that `pkg[when='__unix[version=">=2"]']` is a valid form and must continue to work, the issue is specifically with quotes appearing in the name component.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

jezdez:
### Description

Validate `MatchSpec` package names against CEP-26 rules instead of only rejecting quote characters. Uses a layered approach:

1. Lowercase the name (conda normalizes case at match time)
2. Check for invalid characters with a specific error listing them (e.g. `"package name contains invalid characters ('@)"`)
3. Validate against the full CEP-26 regex for both distributable (`^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`) and virtual (`^__[a-z0-9]...`) package names
4. Skip validation for glob patterns (containing `*`)

Two previously-xfailing test cases (`"We are not banning bad names"` from #15443) now pass since the CEP-26 regex catches `[` as an invalid character.

Closes #16087

### Checklist - did you ...

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -102,6 +102,13 @@ _NAME_VERSION_RE: re.Pattern[str] = re.compile(
     """,
     re.VERBOSE,
 )
+# CEP-26 package name validation (distributable and virtual)
+_CEP26_NAME_RE: re.Pattern[str] = re.compile(
+    r"^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$"
+)
+_CEP26_VIRTUAL_NAME_RE: re.Pattern[str] = re.compile(
+    r"^__[a-z0-9][._-]?([a-z0-9]+(\.|-|_|$))*$"
+)
 _VERSION_BUILD_RE: re.Pattern[str] = re.compile(
     r"""
     ((?:.+?)[^><!,|]?)      # version (non-greedy, not ending in operator)
@@ -113,6 +120,27 @@ _VERSION_BUILD_RE: re.Pattern[str] = re.compile(
     """,
     re.VERBOSE,
 )
+
+
+def _validate_package_name(name: str, original_spec_str: str) -> None:
+    """Validate a package name against CEP-26 rules.
+
+    Glob patterns (containing *) are not validated.
+    Names are lowercased before checking since conda normalizes case.
+    """
+    if "*" in name:
+        return
+    normalized = name.lower()
+    if invalid := re.findall(r"[^a-z0-9._-]", normalized):
+        raise InvalidMatchSpec(
+            original_spec_str,
+            f"package name contains invalid characters ({''.join(dict.fromkeys(invalid))})",
+        )
+    if not (_CEP26_NAME_RE.match(normalized) or _CEP26_VIRTUAL_NAME_RE.match(normalized)):
+        raise InvalidMatchSpec(
+            original_spec_str,
+            "invalid package name",
+        )
 
 
 class MatchSpecType(type):
@@ -915,11 +943,7 @@ def _parse_spec_str(spec_str):
             raise InvalidMatchSpec(
                 original_spec_str, f"no package name found in '{spec_str}'"
             )
-        if "'" in name or '"' in name:
-            raise InvalidMatchSpec(
-                original_spec_str,
-                "package name contains invalid quote characters",
-            )
+        _validate_package_name(name, original_spec_str)
     else:
         raise InvalidMatchSpec(original_spec_str, "no package name found")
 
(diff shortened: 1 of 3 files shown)


conda/models/match_spec.py
```diff
@@ -915,6 +915,11 @@ def _parse_spec_str(spec_str):
             raise InvalidMatchSpec(
                 original_spec_str, f"no package name found in '{spec_str}'"
             )
+        if "'" in name or '"' in name:
+            raise InvalidMatchSpec(
+                original_spec_str,
+                "package name contains invalid quote characters",
+            )
```
kenodegard: And we don't want to just test with the CEP-26 regex?


```suggestion
        if not re.match(r"^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$", name):
            raise InvalidMatchSpec(
                original_spec_str,
                "package name contains invalid characters",
            )
```

kenodegard: Or we can layer it:

```suggestion
        if invalid := re.findall(r"[^a-z0-9_]", name):
            raise InvalidMatchSpec(
                original_spec_str,
                f"package name contains invalid characters ({''.join(invalid)})",
            )
        if not re.match(r"^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$", name):
            raise InvalidMatchSpec(
                original_spec_str,
                "invalid package name",
            )
```

jezdez: Adopted the layered approach from your second suggestion. The validation now:

1. Lowercases the name (conda normalizes case at match time anyway)
2. Checks for invalid characters with a specific error listing them
3. Validates against the full CEP-26 regex for both distributable and virtual package names
4. Skips validation for glob patterns (containing `*`)

The two `xfail("We are not banning bad names")` cases from #15443 now pass since the CEP-26 regex catches `[` as an invalid character.

jaimergp:
Finally :D This needs to go together with the same change in conda-build (and conda-index?) though, otherwise we can generate bad names that we can't match 😬

tests/models/test_match_spec.py
```diff
@@ -656,6 +656,48 @@ def test_invalid_match_spec():
         str(MatchSpec("!xyz 1.3"))
 
 
+@pytest.mark.parametrize(
+    "spec,match",
+    [
+        ("'package'", "invalid characters"),
+        ('"package"', "invalid characters"),
+        ("'package==3'", "invalid characters"),
```
ryanskeith: To save time on tests I think we can throw out some of the version provisioning here. The code at this point should only be passing the name to the validator.   

They are valid in the way that they shouldn't show up in the name but probably overkill as the code shouldn't name validate on the later have due to the _NAME_VERSION_RE regex pattern.

ryanskeith (commented):
Looks good. Though we could reduce some of the tests.

tests/models/test_match_spec.py
jezdez: Good call -- trimmed the version-bearing cases. The name validator only sees the name after `_NAME_VERSION_RE` splits off the version, so those were redundant.

kenodegard (approved):

ryanskeith (approved):
Ah, misread on the tests.  The code is testing the whole MatchSpec process so we should definitely keep those in.

ryanskeith:
> Finally :D This needs to go together with the same change in conda-build (and conda-index?) though, otherwise we can generate bad names that we can't match 😬

Issue created: https://github.com/conda/conda-build/issues/5976

PR #16099, opened 2026-05-14
Add iter_records_v3() to yield unique keys, return v3 repodata instead of existing Shards.iter_records() API

dholth:
### Description

When iterating over (key, package record) in an individual Shards representing one channel, it would be better if each channel had unique keys. Append extension (section name) to each key in "v3".

### Checklist - did you ...

### DIFF
--- a/conda/_private/shards/shards.py
+++ b/conda/_private/shards/shards.py
@@ -336,25 +336,27 @@ class ShardBase(abc.ABC):
 
     def iter_records_v3(self) -> Iterable[tuple[tuple[str, str], dict]]:
         """
-        Yield ((filename, section), record) tuples for all packages in visited shards.
+        Yield ((key, section), record) tuples for all packages in visited
+        shards.
 
-        Section can be:
-        - "packages" for .tar.bz2 packages
-        - "packages.conda" for .conda packages
-        - "v3.whl", "v3.conda", "v3.tar.bz2" for v3 packages
+        Section can be: "packages" for .tar.bz2 packages, "packages.conda"
+        for .conda packages, "v3.whl", "v3.conda", "v3.tar.bz2" for v3 packages.
+
+        key is the same as the filename for "packages", "packages.conda" but is
+        different from the filename for v3 packages.
         """
         for shard in self.visited.values():
             if shard is None:
                 continue
             # Classic packages
             for package_group in ("packages", "packages.conda"):
-                for filename, record in shard.get(package_group, {}).items():
-                    yield (filename, package_group), record
+                for key, record in shard.get(package_group, {}).items():
+                    yield (key, package_group), record
             # v3 packages (iter_records() method depends on these coming last)
             for section_name, group in shard.get("v3", {}).items():
                 v3_section = f"v3.{section_name}"
-                for package_key, record in group.items():
-                    yield (package_key, v3_section), record
+                for key, record in group.items():
+                    yield (key, v3_section), record
 
 
 class ShardLike(ShardBase):
--- a/conda/_private/shards/shards.py
+++ b/conda/_private/shards/shards.py
@@ -326,6 +326,8 @@ class ShardBase(abc.ABC):
     def iter_records(self) -> Iterable[tuple[str, dict]]:
         """
         Yield (filename, record) tuples for all packages in visited shards.
+        For v3 packages, appends the section name as an extension (e.g., '.whl')
+        to preserve the original package type information.
         """
         repodata = self.build_repodata()
         for package_group in ("packages", "packages.conda"):
@@ -333,8 +335,10 @@ class ShardBase(abc.ABC):
         for shard in self.visited.values():
             if shard is None:
                 continue
-            for group in shard.get("v3", {}).values():
-                yield from group.items()
+            for section_name, group in shard.get("v3", {}).items():
+                for package_key, record in group.items():
+                    # Append section name as extension to preserve package type
+                    yield f"{package_key}.{section_name}", record
 
 
 class ShardLike(ShardBase):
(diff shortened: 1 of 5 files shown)


kenodegard (approved):

conda/_private/shards/shards.py
```diff
@@ -326,15 +326,19 @@ def build_repodata(self) -> RepodataDict:
...
             yield from repodata.get(package_group, {}).items()
         for shard in self.visited.values():
             if shard is None:
                 continue
-            for group in shard.get("v3", {}).values():
-                yield from group.items()
+            for section_name, group in shard.get("v3", {}).items():
+                for package_key, record in group.items():
+                    # Append section name as extension to preserve package type
+                    yield f"{package_key}.{section_name}", record
```
jezdez: Do we need to use the `.` here? It creates the wrong impression; it's a file extension, but it's.. just a label. Would a dash `-` also work?

```suggestion
                    # Append section name as extension to preserve package type
                    yield f"{package_key}-{section_name}", record
```

tests/shards/test_shards.py
```diff
@@ -990,8 +990,8 @@ def test_iter_records_v3():
     shardlike.visited["ghost"] = None
     records = dict(shardlike.iter_records())
     assert "mypkg-1.0-0.conda" in records
-    assert "mypkg-1.0-py312_none_any_0" in records
-    assert records["mypkg-1.0-py312_none_any_0"]["name"] == "mypkg"
+    assert "mypkg-1.0-py312_none_any_0.whl" in records
+    assert records["mypkg-1.0-py312_none_any_0.whl"]["name"] == "mypkg"
```
jezdez:
```suggestion
    assert "mypkg-1.0-py312_none_any_0-whl" in records
    assert records["mypkg-1.0-py312_none_any_0-whl"]["name"] == "mypkg"
```

conda/gateways/shards/typing.py
```diff
@@ -21,6 +21,8 @@ def __contains__(self, package: str) -> bool:
     def iter_records(self) -> Iterator[tuple[str, dict]]:
         """
         Yield (filename, record) tuples for all packages in visited shards.
+        For v3 packages, the filename includes the section name as an extension
+        (e.g., '.whl') to preserve package type information.
```
jezdez:
```suggestion
        For v3 packages, the filename includes the section name as a suffix
        (e.g., '-whl') to preserve package type information.
```

conda/_private/shards/shards.py
```diff
@@ -326,15 +326,19 @@ def build_repodata(self) -> RepodataDict:
     def iter_records(self) -> Iterable[tuple[str, dict]]:
         """
         Yield (filename, record) tuples for all packages in visited shards.
+        For v3 packages, appends the section name as an extension (e.g., '.whl')
```
jezdez:
```suggestion
        For v3 packages, appends the section name as a suffix (e.g., '-whl')
```

jezdez (commented):
My main concern is basically to not use dots for this unless we really need it for something, I would not be surprised if we did _other_ conda-pypi-type plugins for other ecosystems in the future, and then the file extensions might not fit with the file type.

conda/_private/shards/shards.py
kenodegard: Depends on whether `iter_records` is supposed to return `(filename, record)` or if `(key, record)` is acceptable?

kenodegard: Also consider `{section_name}/{package_key}`

kenodegard: If it has to be a filename can we just do:

```suggestion
                    # Append section name as extension to preserve package type
                    yield record.fn, record
```

dholth:
I might prefer to start returning a tuple `(section, package identifier)` or `(package identifier, section)`. Avoids string appends and is still a nice dict key.

dholth:
n.b. the next conda release is the first release that contains this API.

jaimergp:
In https://github.com/conda-incubator/conda-rattler-solver/pull/53#issuecomment-4455720888 it was suggested to use `(key, section, record)`. For the old `packages` and `packages.conda`, `key` could be `None` (top-level). You can also refer to `key` as `revision`.

news/shards-v3-unique-keys
```diff
@@ -0,0 +1,21 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* In v3 sharded repodata, yield unique keys from `Shards.iter_records()`
+  by appending section name (package extension) to keys. (`iter_records()`
```
dholth:
```suggestion
  by yielding (package, section_name) keys from (`iter_records()`
```

conda/_private/shards/shards.py
dholth: New version yields tuple keys

dholth:
If we need to preserve the `(str, record)` instead of the `((str, str), record)` signature then we could copy it to `iter_records_v3()` and preserve the older signature with `iter_records()`
Thinking forward to search, we might or might not need to expose individual shards (everything for one package for one channel) instead of all visited shards.

conda/gateways/shards/typing.py
```diff
@@ -18,9 +18,14 @@ def package_names(self) -> KeysView[str]:
     def __contains__(self, package: str) -> bool:
         """Check if a package is available in this shard collection."""
 
-    def iter_records(self) -> Iterator[tuple[str, dict]]:
+    def iter_records(self) -> Iterator[tuple[tuple[str, str], dict]]:
```
soapy1: Since this is changing a public api that already exists in a released version of conda, does this need to follow some process for changing the function signature? For example, this will probably break `conda-rattler-solver`.

dholth: Yes

dholth: I've changed it so that `iter_records_v3()` yields the v3 repodata with `(key, section), record` as the yield type. `iter_records()` is implemented by filtering the `v3()` results.

soapy1 (approved):

jezdez:
pre-commit.ci autofix

jezdez (approved):

Issue #16073, opened 2026-05-11
Extend `MatchSpec` support: conditionals, extras, flags


jaimergp:
Syntax described in https://github.com/conda/ceps/pull/164, https://github.com/conda/ceps/pull/165 and https://github.com/conda/ceps/pull/166, respectively.

This needs changing in the parsing engine in `match_spec.py`, new record fields in `records.py`, solver extensions for `when` fields, new matching logic... Not trivial! This epic aggregates all the necessary PRs:

- [x] Parse `when`, `extras`, `flags`: https://github.com/conda/conda/pull/15443
- [x] Add `extras` and `flags` to `PackageRecord`
  - [x] Initial support at https://github.com/conda/conda/pull/15443
- [ ] Match `extras`, `flags`
- [ ] Respect `when` in solver
   - [ ] classic: not supported
   - [ ] libmamba: pending implementation upstream
   - [x] rattler: https://github.com/conda-incubator/conda-rattler-solver/pull/28

Related issues which will or may be fixed as we work on this:

- https://github.com/conda/conda/issues/15749
- https://github.com/conda/conda/issues/15988
- https://github.com/conda/conda/issues/15654
- https://github.com/conda/conda/issues/11053
- https://github.com/conda/conda/issues/6965
- https://github.com/conda/conda/issues/16087

Issue #16016, opened 2026-05-04
QA Test Plan `26.5.0`


danyeaw:
## 1. Feature/Package Overview

**Name:** conda
**Version:** 26.5.0

**Brief Description:**
This release delivers a substantial cold-start performance overhaul — including a ~5x faster condarc search-path expansion, a fast path for `conda --version` / `-V` that bypasses full parser and plugin loading, deferred heavy imports in `notices/core.py` and `main_env.py`, pre-compiled hot-path regexes in `match_spec.py` and `activate.py`, replacement of `deepcopy` with a dict comprehension in the solver's `_get_reduced_index`, and faster `PrefixData` loading via `read_bytes()` + `json.loads()` — alongside three milestone features: sharded repodata ported from `conda-libmamba-solver` into `conda.gateways.repodata.shards` (#15906), `extras` key support in MatchSpec bracket syntax for optional dependency resolution (#15749), and improved help text across `conda export` and `conda create` for lockfile workflows (#15960).

Additional user-facing changes include `conda create --clobber` parity with `conda install` (#15584), `conda activate`/`deactivate -h`/`--help` now exiting with code 0 (#15958), a warning when exporting an environment that contains pip- or uv-installed packages (#15838), multi-platform env spec preflight errors in `conda env create`/`update` (#15927/#15928), PowerShell reactivation after install/update/remove so package env vars apply in-session (#15643), and a more robust Windows notices cache (#15796).

The release also bumps minimum dependencies (`pluggy >= 1.6.0`, `conda-libmamba-solver >= 26.4.0`), adds `conda-pypi`, `conda-lockfiles`, `conda-self`, and `conda-spawn` as default-installed plugin dependencies in the recipe, and carries several pending deprecations: `ChannelType` metaclass (#15942), `--env-spec`/`--environment-specifier` → `--format` (#15794), `conda.instructions` module (#15874), and the `conda.stdout`/`conda.stderr` log handlers (#15876). A `RemoveError` regression is also fixed when a missing conda dependency was installed via pip (#14050).

**Areas affected:**

- **Startup / CLI** — Fast `conda --version`, deferred imports, activate/deactivate `--help`
- **Solver** — MatchSpec `extras`, deepcopy removal, sharded repodata API
- **Repodata Gateway** — New `shards` module
- **Environment Specs / Export** — Multi-platform preflight, pip-in-env warning, lockfile help text
- **Shell Integration** — PowerShell reactivation, activation `--help` exit code
[...]

PR #16104, opened 2026-05-15
bugfix: MatchSpec parsing error fix (`AttributeError`)

travishathaway:
### Description

This fixes an error encountered during [QA testing](https://github.com/conda/conda/issues/16016) for release 26.5.0 Scenario 8.

xref: #16073

### Checklist - did you ...

### DIFF
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -394,7 +394,10 @@ class MatchSpec(metaclass=MatchSpecType):
         return True
 
     def _match_individual(self, record, field_name, match_component):
-        val = getattr(record, field_name)
+        # Use getattr with a default of None so that MatchSpec fields that are
+        # not yet present on PackageRecord (e.g. "extras") do not raise
+        # AttributeError and instead evaluate as a non-match.  See GH-16016.
+        val = getattr(record, field_name, None)
         try:
             return match_component.match(val)
         except AttributeError:
--- a/conda/models/match_spec.py
+++ b/conda/models/match_spec.py
@@ -297,7 +297,7 @@ class MatchSpec(metaclass=MatchSpecType):
         "license_family",
         "fn",
         "when",  # str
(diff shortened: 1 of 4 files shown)


conda/models/records.py
```diff
@@ -410,6 +410,7 @@ def dist_fields_dump(self):
     constrains = ListField(str, default=())
 
     flags = ListField(str, default=(), required=False, default_in_dump=False)
+    extras = ListField(str, default=(), required=False, default_in_dump=False)
```
jaimergp: This is not a list, but a `dict[str, list[str]]`

conda/models/match_spec.py
```diff
@@ -394,7 +394,10 @@ def match(self, rec: PackageRecord | dict[str, Any]) -> bool:
         return True
 
     def _match_individual(self, record, field_name, match_component):
-        val = getattr(record, field_name)
+        # Use getattr with a default of None so that MatchSpec fields that are
+        # not yet present on PackageRecord (e.g. "extras") do not raise
+        # AttributeError and instead evaluate as a non-match.  See GH-16016.
+        val = getattr(record, field_name, None)
```
jaimergp, line 400: This changes behavior with silenced errors, not sure we want to do that as part of a bug fix.

jaimergp (changes requested):

travishathaway:
Hi @jaimergp,

This pull request can be closed, but I did submit a possible implementation that might be helpful to reference later for #16073. It address the `dict[str, list[str]]` data type. But, this is unfortunately not so easy to handle given the way things are currently laid out, hence all the new code.

jaimergp:
This is what I don't get, the code _was_ there: https://github.com/conda/conda/pull/15443/files#diff-7d79ad89016b706bf3a9c43993ce3a13e868436944976cafbbb5cb239868af17R413-R416

PR #16130, opened 2026-05-20
Remove issue_comment trigger from `update.yml`

jaimergp:
We have never used this trigger and it exposes additional maintenance surface. Maintainers can still achieve the same operations with `workflow_dispatch` from the Actions tab.


jezdez (approved):
Reviewed: single-file change removes the vulnerable issue_comment trigger. Diff verified clean across all 29 repos.

PR #16149, opened 2026-05-24
Search shards for packages

### LINKED ISSUE #16134: Add search support for sharded repodata
### What is the idea?
When `repodata_use_shards` is enabled and a channel provides sharded repodata, `conda search` should fetch only the relevant package shard(s) rather than downloading the full `repodata.json`.

### Why is this needed?
conda search currently always fetches the full monolithic repodata.json for every channel and subdir, even when sharded repodata is available and the user may have recently cached relevant shards from a solve. For large channels like conda-forge, this results in downloading megabytes of data for a simple name lookup. The sharded repodata infrastructure introduced in CEP-16 already supports targeted per-package fetching, but that optimization is only wired into the solver path today. Search should benefit from the same performance improvements.

This would also fix https://github.com/conda/conda-pypi/issues/373, which is a case where there is currently no `repodata.json` available at all due to server issues.

### What should happen?
- When `repodata_use_shards` is enabled and a channel exposes a shards index, `conda search` resolves the name component of the MatchSpec to the corresponding shard(s) and fetches only those (instead of a full tree of shards like it does currently with install operations).
- Shard cache hits (from the SQLite shard cache) should be reused, so a search immediately after a solve costs nothing for the packages already cached.
- When the shards index is unavailable or the spec cannot be mapped to specific shards (e.g., a wildcard/glob match that spans many packages), the command falls back to the current monolithic repodata.json path.
- The behavior is consistent whether the search is run standalone or inside a channel that uses ShardLike wrapping for monolithic repodata.

soapy1:
### Description

Fixes: https://github.com/conda/conda/issues/16134

This PR adds the ability to search packages from sharded channels. If no shards are available for the channel, conda will fallback to searching from monolithic repodata. For example, the `conda-pypi` channel only serves sharded repodata. So, on the current release of conda, searching this channel is not possible.

```
$ conda search cowsay --override-channels --channel conda-pypi        
Loading channels: done
No match found for: cowsay. Search: *cowsay*

PackagesNotFoundInChannelsError: The following packages are not available from current channels:

  - cowsay
```

With this change:
```
$ conda search cowsay --override-channels --channel conda-pypi
Loading channels: done
# Name                       Version           Build  Channel             
cowsay                           3.0 py2.py3_none_any_0  conda-pypi          
cowsay                           4.0 py2.py3_none_any_0  conda-pypi          
cowsay                           6.0 py2.py3_none_any_0  conda-pypi          
cowsay                           6.0  py3_none_any_0  conda-pypi          
cowsay                           6.1  py3_none_any_0  conda-pypi        
```

### Speed
This pr also enables a speed up to searching for packages. For example, with shards disabled (eg. `CONDA_REPODATA_USE_SHARDS=false`):
```
$ hyperfine -w 1 --runs 10 'conda search numpy\=1.14 --channel conda-forge --override-channels'
Benchmark 1: conda search numpy\=1.14 --channel conda-forge --override-channels
  Time (mean ± σ):     23.364 s ±  3.935 s    [User: 20.054 s, System: 2.604 s]
  Range (min … max):   20.058 s … 30.337 s    10 runs
```

With shards enabled (eg. `CONDA_REPODATA_USE_SHARDS=true`):
```
 $ hyperfine -w 2 --runs 10 'conda search numpy\=1.14 --channel conda-forge --override-channels'
Benchmark 1: conda search numpy\=1.14 --channel conda-forge --override-channels
  Time (mean ± σ):     788.1 ms ±  26.2 ms    [User: 656.6 ms, System: 81.6 ms]
  Range (min … max):   756.5 ms … 842.7 ms    10 runs
```

### Checklist - did you ...


conda/cli/main_search.py
```diff
@@ -252,20 +296,18 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
+        if context.repodata_use_shards:
+            matches = _search_package_via_shards(spec, channel_urls, subdirs)
+        else:
+            matches = _search_package(spec, channel_urls, subdirs)
     if not matches and not args.skip_flexible_search and spec.get_exact_value("name"):
         flex_spec = MatchSpec(spec, name=f"*{spec.name}*")
         if not context.json:
             print(f"No match found for: {spec}. Search: {flex_spec}")
-        matches = sorted(
-            SubdirData.query_all(flex_spec, channel_urls, subdirs),
```
dholth, line 272: Would it make sense to provide a new interface that only provides the `.query_all()` method, so that search didn't have to know the difference between shards/not shards

soapy1: I like the idea of making cli search agnostic to shards. I found that the shards code also uses the subdirdata module. I think it's worth a look to see if there is some path forward there 👀

dholth: The big changes I want to make are to ban "*" search and contains-no-package-name search "all packages with sha256=x"

soapy1: Right, the building repodata subset system requires having the full name of the package? Though imho, that is a pretty big change around the expectations of the capabilities of the `conda search` command. Also good to consider the `conda repoquery search` command.

soapy1:
pre-commit.ci autofix

tests/shards/test_shards_subset.py
```diff
@@ -233,6 +233,75 @@ def test_build_repodata_subset(prepare_shards_test: None, tmp_path):
     )
 
 
+@pytest.mark.parametrize(
+    "algorithm,depth,root_package,expected_result",
+    [
+        (
+            "bfs",
+            0,
```
dholth, line 241: Seeing that depth = 0 means it will fetch the root packages and nothing else?

conda/_private/shards/subset.py
```diff
@@ -295,7 +297,9 @@ def _reachable_bfs(self, root_packages, shard_cache: cache.ShardCache):
                 node.visited = True
 
                 for next_node, _ in self._outgoing(node):
-                    if not next_node.visited:  # pragma: no branch
+                    if (
+                        not next_node.visited and next_node.distance <= self.depth
+                    ):  # pragma: no branch
```
dholth, line 302:
```suggestion
                    ): 
```
Does this branch sometimes return False now

dholth:
When `~/.condarc` is empty, conda install x says
```% python -m conda install conda
Retrieving notices: done

NoChannelsConfiguredError: No channels are configured. Conda requires at least one channel to search for packages.
```
but python -m conda search x says `PackagesNotFoundInChannelsError: ...`

Is not a regression, same on main branch. Fix at https://github.com/conda/conda/issues/16180

soapy1:
pre-commit.ci autofix

dholth:
I am impressed with the time savings. On my machine, conda-forge+defaults takes 6.39s to search without this feature, but 0.74s to search with shards search enabled.
The implementation looks easy to refactor if necessary and doesn't saddle us with a complicated new API.
I noticed `conda search -c conda-forge "*"` produces `PackagesNotFoundInChannelsError` if there are shards, but `conda search -c main "*"` would list all packages.
Last time I tried it, `pixi` would download all shards in this situation, but I think we are better off disallowing the wildcard search.

soapy1:
>I noticed conda search -c conda-forge "*" produces PackagesNotFoundInChannelsError if there are shards, but conda search -c main "*" would list all packages.

mmm, this is an important idea to bring up, thank you! Currently this update to searching does not work with wildcards. You can observe that `conda search -c main "*"` works because the main channel is not sharded to conda is falling back to pulling down monolitich repodata. I will add a note in the description about this caveat.

> I think we are better off disallowing the wildcard search.

hmmm, that's a pretty big change in expectation of the `conda search` command imo. If we choose to go that route, it would be nice to have a message for users using wildcards suggesting that they do something else. How about I make a follow up ticket for the team to discuss what the behaviour for this should be?

dholth:
Another test case `% python -m conda search "conda*" -c conda-forge`. Returns no packages in this PR. Personally, I dislike bare "\*" but "conda*" doesn't bother me because the results are probably tractable. In shards, we do know how many package names either query would find though (since we have the list of package names without downloading all shards)
Another,
```
% python -m conda search "coanda" -c conda-forge
Loading channels: done
No match found for: coanda. Search: *coanda*
```

tests/shards/test_shards_subset.py
```diff
@@ -602,7 +653,10 @@ def test_shards_network_thread(http_server_shards, shard_cache_with_data, monkey
 
                 # Make sure this is either one of the two packages from above ("foo" or "bar")
                 assert set(shard.get("packages", {}).keys()).intersection(
-                    ("foo.tar.bz2", "bar.tar.bz2")
+                    (
+                        "foo.tar.bz2",
+                        "bar.tar.bz2",
```
dholth, line 659: Not part of the review but I was annoyed by the formatter splitting all of these; this trailing comma causes or caused the formatter to split structures across multiple lines, and something is also adding the trailing comma.

dholth (approved):
Thanks

jezdez (approved):

conda/core/subdir_data.py
```diff
@@ -730,3 +730,122 @@ def _get_base_url(self, repodata: dict, with_credentials: bool = True) -> str:
...
+    :param channels: An iterable of urls for channels or `Channel` objects.
+        If None, will fall back to `context.channels`.
+    :param subdirs: If None, will fall back to context.subdirs.
+    :param repodata_fn: The filename of the repodata.
+    :return: A tuple of `PackageRecord` objects.
+    """
+    if context.repodata_use_shards:
+        return _search_package_via_shards(match_spec, channels, subdirs)
+    else:
+        return _search_package(match_spec, channels, subdirs, repodata_fn)
```
jezdez, line 851: Should we make these two functions public?

dholth: No

dholth: These functions are very much an implementation detail.

PR #16160, opened 2026-05-29
Add structured user facing error model to CondaError

### LINKED ISSUE #16036: Add structured user-facing error model to CondaError
### What is the idea?

Attach a structured user-facing error model to `CondaError` so that solver failures, dependency conflicts, and network errors surface a concise summary, optional cause, and ordered next-step hints instead of raw exception strings.

### Why is this needed?

- When commands fail today, users see walls of text, solver noise, or tracebacks with no clear explanation of what went wrong or what to do next.
- `--json` output cannot emit a well-formed, machine-readable error object because there is no structured payload attached to exceptions.
- The two error categories users hit most — `PackagesNotFoundError` and dependency conflicts (`UnsatisfiableError`) — have no actionable hints today.
- Defining the model on `CondaError` itself ensures every command and runner benefits, not just one code path.

### What should happen?

- Add `UserFacingErrorDetails` (summary, cause, ordered hints with optional stable `hint_code`) as a `user_facing` keyword on `CondaError`.
- Render structured errors with a Rich panel in interactive TTY mode, plain text otherwise, and no ANSI sequences in `--json` output.
- MVP coverage: `PackagesNotFoundError`, `UnsatisfiableError`, and their direct specializations on `conda create` and `conda install`.
- `--json` error output includes `user_facing` object with `summary`, `cause`, `hints`, and `hint_codes` alongside existing legacy fields.

### Additional Context

- Part of epic #15873.
- A proof-of-concept exists on the `improved-errors` branch (referenced in #15845).
- Full requirements are in the TRD from #15845.

kenodegard:
### Description

Resolves https://github.com/conda/conda/issues/16036

### Checklist - did you ...


conda/core/link.py
```diff
@@ -776,7 +776,29 @@ def _verify_transaction_level(prefix_setups):
             # this should never be able to be skipped, even with --force
             yield RemoveError(
                 "This operation will remove conda without replacing it with\n"
-                "another version of conda."
+                "another version of conda.",
+                guidance={
+                    "summary": (
+                        "Conda cannot remove itself from this environment without "
+                        "installing another conda build to replace it."
```
kenodegard:
```suggestion
                        "Conda cannot remove itself."
```

conda/core/link.py
```diff
@@ -776,7 +776,29 @@ def _verify_transaction_level(prefix_setups):
...
+                    ),
+                    "cause": (
+                        "The planned transaction would uninstall the conda package "
+                        "without linking a replacement."
+                    ),
+                    "hints": [
+                        {
+                            "text": "Install a specific conda build first, for example: "
+                            "conda install conda=<version>",
+                            "hint_code": "install_conda_build",
```
kenodegard:
```suggestion
                            "text": (
                                "Install a specific conda version, for example: \n"
                                "    conda install conda=<version>"
                            ),
                            "hint_code": "install_conda_version",
```

conda/_private/exception_guidance.py
```diff
@@ -0,0 +1,124 @@
...
+
+    .. attribute:: cause
+       An explanation of what went wrong, distinct from the raw error text.
+
+    .. attribute:: hints
+       An ordered list of :class:`GuidanceHint` objects with actionable next
+       steps. Hints are rendered in order.
+    """
+
+    summary: str | None = None
```
danyeaw, line 49: If `ErrorGuidance` is being used, what do you think about making the summary at least required?

conda/_private/exception_guidance.py
```diff
@@ -0,0 +1,124 @@
...
+    cause: str | None = None
+    hints: tuple[GuidanceHint, ...] = ()
+    hint_codes: tuple[str, ...] = field(init=False, repr=False)
+
+    def __post_init__(self) -> None:
```
danyeaw, line 56: Rather than storing a copy of the hint codes on the object, what if we generate it while serializing to json? I think that would keep the dataclass a bit cleaner.

conda/_private/exception_guidance.py
```diff
@@ -0,0 +1,124 @@
...
+    if isinstance(value, dict):
+        if not value:
+            return None
+        return _guidance_from_dict(value)
+    raise TypeError(
+        f"guidance must be dict or ErrorGuidance, not {type(value).__name__}"
+    )
+
+
+def _guidance_from_dict(value: dict[str, Any]) -> ErrorGuidance:
```
danyeaw: Should we add some additional validation here to ensure that guidance attached to the CondaError isn't malformed?

danyeaw (commented):
Hey @kenodegard, I did a early review focusing on the core data model implementation. This looks really, I see you put a lot of effort in to making this cleanly separated! I like your naming of guidance rather than user_facing as well.

conda/_private/exception_guidance.py
kenodegard: I intentionally kept summary optional, it's an override for the exception's default message. In several places I found the summary and existing message would be nearly identical, so requiring it would force redundancy.

The real invariant is that at least one field (summary, cause, or hints) must be populated. I'll make sure we enforce that in `__post_init__`.

conda/_private/exception_guidance.py
kenodegard: I agree `hint_codes` is only needed for the JSON dump but making it only generated during serialization results in messier code, instead of using `dataclasses.asdict` we end up with:

```python
    @property
    def hint_codes(self) -> tuple[str, ...]:
        """Hint codes extracted from :attr:`hints` for JSON serialization."""
        return tuple(hint.hint_code for hint in self.hints)

    def dump(self) -> dict[str, object]:
        """Serialize for :meth:`conda.CondaError.dump_map`."""
        dump = {}
        if self.summary:
            dump["summary"] = self.summary
        if self.cause:
            dump["cause"] = self.cause
        if self.hints:
            dump["hints"] = tuple(asdict(hint) for hint in self.hints)
        if self.hint_codes:
            dump["hint_codes"] = self.hint_codes
        return dump
```

kenodegard: eh decided to go this route after all since we want to filter out falsy values

conda/exceptions.py
```diff
@@ -941,59 +990,57 @@ def __init__(
...
-Your installed version is: {ref}
-"""
+                Your installed version is: {ref}
+                """
             ),
         }
 
         msg = ""
         self.unsatisfiable = []
         if len(bad_deps) == 0:
```
jezdez, line 1030: This PR changes the constructor contract/tests to use the classified conflict map shape, but `Resolve.solve()` still has a legacy `chains=False` caller in `conda/resolve.py` that passes `[[d] for d in wrong_version_packages]`. Can we either update that caller to the new map shape with a test, or keep a tested compatibility path here? Right now that edge still fails before producing a useful `UnsatisfiableError`.

jezdez (commented):
Thanks for pushing the structured guidance model. I found a few user-visible edges that are outside the current test coverage.

conda/_private/exception_guidance.py
```diff
@@ -0,0 +1,125 @@
...
+            exception: ``str(exception)`` is used as the headline unless
+                ``summary`` is set.
+
+        Returns:
+            Formatted string including the exception class name, cause,
+            and actionable hints.
+        """
+        if not isinstance(exception, CondaError):
+            raise ValueError(f"expected CondaError, got {type(exception).__name__}")
+        headline = (self.summary if self.summary else exception.message).rstrip()
```
jezdez: When there is no guidance summary, terminal output should use the formatted exception string. Otherwise templated `CondaError` messages leak raw placeholders once guidance is attached.

```suggestion
        headline = (self.summary or str(exception)).rstrip()
```

conda/exceptions.py
```diff
@@ -1014,33 +1061,138 @@ def __init__(
...
+        if not had_empty:
+            for class_name in bad_deps:
+                if class_name == "python":
+                    has_python = True
+                elif class_name == "request_conflict_with_history":
+                    has_history = True
+                elif class_name == "virtual_package":
+                    has_virtual = True
+                elif class_name == "direct":
+                    has_direct = True
```
jezdez: `_classify_bad_deps()` initializes all four conflict-category keys, even when most values are empty. Checking key presence makes guidance report every cause category for a single populated category.

```suggestion
        if not had_empty:
            for class_name, dep_class in bad_deps.items():
                if not dep_class:
                    continue
                if class_name == "python":
                    has_python = True
                elif class_name == "request_conflict_with_history":
                    has_history = True
                elif class_name == "virtual_package":
                    has_virtual = True
                elif class_name == "direct":
                    has_direct = True
```

conda/exceptions.py
kenodegard: This PR doesn't change how `bad_deps` is used, this PR _corrects_ the type annotation. The `UnsatisfiableError([[d] for d in wrong_version_packages], chains=False)` is a bug that fails on `main` too:

```pycon
>>> from conda import __version__
>>> __version__
'26.5.3'
>>> from conda.exceptions import UnsatisfiableError
>>> from conda.models.match_spec import MatchSpec
>>> wrong_version_packages = {MatchSpec("a"), MatchSpec("b")}
>>> UnsatisfiableError([[d] for d in wrong_version_packages], chains=False)
  File "<python-input-5>", line 1, in <module>
    UnsatisfiableError([[d] for d in wrong_version_packages], chains=False)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kodegard/.conda/current/lib/python3.13/site-packages/conda/exceptions.py", line 988, in __init__
    for class_name, dep_class in bad_deps.items():
                                 ^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'items'
```

jezdez: Good call!

kenodegard: Fixed in https://github.com/conda/conda/pull/16160/commits/e1d5d87b810f400497990ea79947b871a606c679 + https://github.com/conda/conda/pull/16160/commits/0f6e8442b11ed57f47f4e49ef11dd6a9bd359a4e

tests/_private/test_exception_guidance.py
```diff
@@ -0,0 +1,213 @@
...
+        "CondaError: Failed.",
+        "",
+        "Cause: Because of X.",
+        "Next steps:",
+        "  - (try_y) Try Y.",
+        "  - (try_z) Try Z.",
+    ]
+
+
+def test_ErrorGuidance_hint_codes_property() -> None:
```
jezdez, line 129: Not a super fan of this mixed capitalizaiton bTW

jezdez (approved):

PR #15840, opened 2026-04-01
Remove legacy subcommand detection

kenodegard:
### Description

Followup to removal of legacy subcommand loading.

We were still "finding" the old `conda-*` executables but had removed the ability to load them, e.g., when `conda-smithy` (which only implements the `conda-*` executable, not the plugin framework) is installed `conda 26.3.0` crashes:

<details><summary><code>conda list</code></summary>

Screenshot 2026-04-01 at 16 44 28

</details>

<details><summary><code>conda smithy</code></summary>

Screenshot 2026-04-01 at 16 44 42

</details> 

<details><summary><code>conda missing</code></summary>

Screenshot 2026-04-01 at 16 45 16

</details> 

So we remove usage of `find_commands` allowing us to deprecate `conda.cli.find_commands`. As a bonus also found a way to record the canonical name of where a plugin was implemented so we can display that path/name in our error messages and in `conda info`:

| Before | After |
|---|---|
| Screenshot 2026-04-01 at 16 48 05 | Screenshot 2026-04-01 at 16 49 47 |

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #16166, opened 2026-06-01
Print friendly plugin names, and print both plugins' names for conflicting plugins

agriyakhetarpal:
### Description

In the CI for a project at wntrblm/nox#1095, I recently ran into the following error:

```
PluginError: Conflicting plugins found for `post_solves`:

  - CondaPostSolve(name='signature-verification', action=<conda.trust.signature_verification._SignatureVerification object at 0x10774b770>)

Multiple conda plugins are registered via the `conda_post_solves` hook. Please make sure that you don't have any incompatible plugins installed.
```

The reason for this is that the CI setup uses conda 25.11.1 with setup-miniconda@v3. That means there is an older version of conda that still implements the signature-verification post_solves hook, and a newer conda-content-trust that does the same. In this scenario, `run_constrained: conda-content-trust >=0.3.0` was added only with conda 26.3.

After talking to @kenodegard and @jaimergp, we can do two things, which this PR does:
1. Use `importlib.metadata.packages_distributions` to output friendly plugin names. We fall back to the raw pluggy plugin name in case this isn't possible, which may happen when plugin developers perform editable installations, and they should understand a module path just fine. However, this won't be the case for a majority of plugin users, of course.
2. Restructure the conflict detection for plugins from a flat list to one where we print out plugins grouped by name, so both sides of each conflict appear on the same line.

Therefore, the old output was `- CondaPostSolve(name='signature-verification', action=<...object...>)`, and the new output should be something along the lines of `- 'signature-verification' provided by:` split on one line per provider, like this (for example):

```
Conflicting plugins found for `<hook>`:
  - 'archspec' provided by:
    - conda 26.1.0
    - conda-content-trust 0.2.0
```

### Checklist - did you ...


agriyakhetarpal:
The test failures are unrelated; it looks like the tests are getting cancelled for some reason. Thanks!

conda/plugins/manager.py
```diff
@@ -98,6 +98,22 @@
 log = logging.getLogger(__name__)
 
 
+def _plugin_source(impl_plugin_name: str) -> str:
```
kenodegard: nice followup to the `impl.plugin_name` work from https://github.com/conda/conda/pull/15840

there are probably other places adding this distname+version would also be useful (e.g., `conda info`)

conda/plugins/manager.py
```diff
@@ -385,20 +401,24 @@ def get_hook_results(self, name, **kwargs):
...
+            for plugin_name, conflicting in groupby_to_dict(
+                lambda plugin: plugin.name, plugins
+            ).items()
+            if len(conflicting) > 1
+        }
+        if conflict_groups:
+            lines = []
+            for conflict_name, conflicting in sorted(conflict_groups.items()):
+                providers = dashlist(
+                    [_plugin_source(p.impl.plugin_name) for p in conflicting],
```
kenodegard: Let's also sort the list of conflicts:


```suggestion
                    [_plugin_source(p.impl.plugin_name) for p in sorted(conflicting)],
```

tests/plugins/test_manager.py
```diff
@@ -105,8 +105,12 @@ def conda_virtual_packages():
     plugin_manager.register(SecondArchspec)
     with pytest.raises(
         PluginError, match="Conflicting plugins found for `virtual_packages`"
-    ):
+    ) as exc_info:
         plugin_manager.get_hook_results(name)
+    # Both conflicting sources must be listed so users know what to uninstall
+    error_text = str(exc_info.value)
+    assert "'archspec' provided by:" in error_text
+    assert error_text.count("\n    - ") >= 2
```
kenodegard: Hm probably cleaner to just assert the whole error message instead?

```python
    with pytest.raises(
        PluginError,
        match=(
            r"Conflicting plugins found for `virtual_packages`:\n\n"
            r"  - 'archspec' provided by:\n"
            r"    - conda.plugins.virtual_packages.archspec\n"
            r"    - tests.plugins.test_manager.test_get_hook_results.<locals>.SecondArchspec\n\n"
        ),
    ):
```

conda/plugins/manager.py
kenodegard: Let's add a unittest for `_plugin_source`

conda/plugins/manager.py
agriyakhetarpal: Thanks! Noting here that `CondaPlugin` dataclasses don't have an `lt` dunder, so I had to do `sorted(conflicting, key=lambda p: p.impl.plugin_name)` for this, to sort by the string plugin name.

conda/plugins/manager.py
agriyakhetarpal: Done in cb8cf6aad5289fbc9cffe9a9337ead78da5ecd41 and 76473fc0ff7ca44de053ff7f525c02a6eb87f271. There are some other uses of plugin names, but these two are the only ones that are applicable in this context.

jezdez:
I opened https://github.com/agriyakhetarpal/conda/pull/5 as a concrete sketch for the plugin-manager-method version of this change.

conda/cli/main_info.py
```diff
@@ -517,11 +518,13 @@ def _system_component(self) -> str:
...
-        output.append(f"conda-build: {plugin_name or '(missing)'}")
+        impl_name = getattr(getattr(conda_build, "impl", None), "plugin_name", None)
+        plugin_source = _plugin_source(impl_name) if impl_name else None
+        output.append(f"conda-build: {plugin_source or '(missing)'}")
         for name, plugin in sorted(subcommands.items()):
-            plugin_name = getattr(getattr(plugin, "impl", None), "plugin_name", None)
-            output.append(f"conda-{name}: {plugin_name or '(unknown)'}")
+            impl_name = getattr(getattr(plugin, "impl", None), "plugin_name", None)
+            plugin_source = _plugin_source(impl_name) if impl_name else None
+            output.append(f"conda-{name}: {plugin_source or '(unknown)'}")
```
kenodegard: I think we should add a `_plugin_name` helper function in manager

and then refactor both to accept the plugin itself, e.g.:
```python
def _plugin_name(plugin: CondaPlugin) -> str or None:
    ...

def _plugin_source(plugin: CondaPlugin) -> str or None:
    ....
```

conda/plugins/manager.py
```diff
@@ -385,20 +402,27 @@ def get_hook_results(self, name, **kwargs):
...
+        if conflict_groups:
+            lines = []
+            for conflict_name, conflicting in sorted(conflict_groups.items()):
+                providers = dashlist(
+                    [
+                        _plugin_source(p.impl.plugin_name)
+                        for p in sorted(conflicting, key=lambda p: p.impl.plugin_name)
+                    ],
+                    indent=4,
+                )
```
kenodegard:
```suggestion
        conflict_groups = {
            plugin_name: sorted(_plugin_source(plugin) for plugin in conflicting)
            for plugin_name, conflicting in sorted(
                groupby_to_dict(lambda plugin: plugin.name, plugins).items()
            )
            if len(conflicting) > 1
        }
        if conflict_groups:
            lines = []
            for conflict_name, conflicting in conflict_groups.items():
                providers = dashlist(conflicting, indent=4)
```

agriyakhetarpal:
Thanks @jezdez! I merged your PR into my PR. TIL about pluggy's `DistFacade`; pretty neat!

agriyakhetarpal:
I think we might need to harden `get_canonical_name` a bit so that we can be more helpful:

```python
from conda.plugins.hookspec import hookimpl
from conda.plugins.types import CondaSolver
from conda.exceptions import PluginError

pm = context.plugin_manager

class IntruderSolver:
    @hookimpl
    def conda_solvers(self):
        yield CondaSolver(name='classic', backend=object)

pm.register(IntruderSolver())

try:
    pm.get_hook_results('solvers')
except PluginError as e:
    print(str(e))
```

which gets us

```
Conflicting plugins found for `solvers`:

  - 'classic' provided by:
    - <unknown_module>.IntruderSolver[4392487968]
    - conda.plugins.solvers

Multiple conda plugins are registered via the `conda_solvers` hook. Please make sure that you don't have any incompatible plugins installed.
```

conda/cli/main_info.py
agriyakhetarpal: Thanks!

jezdez:
I opened https://github.com/agriyakhetarpal/conda/pull/6 as a follow-up for the remaining review tweaks: hardening canonical names for registered plugin instances, tightening the conflict-message tests, and clarifying the entry point bookkeeping/news wording.

conda/cli/main_info.py
```diff
@@ -527,13 +527,18 @@ def _system_component(self) -> str:
...
         conda_build = subcommands.pop("build", None)
-        plugin_name = getattr(getattr(conda_build, "impl", None), "plugin_name", None)
-        output.append(f"conda-build: {plugin_name or '(missing)'}")
-        for name, plugin in sorted(subcommands.items()):
-            plugin_name = getattr(getattr(plugin, "impl", None), "plugin_name", None)
-            output.append(f"conda-{name}: {plugin_name or '(unknown)'}")
+        plugin = getattr(getattr(conda_build, "impl", None), "plugin", None)
+        output.append(
+            f"conda-build: {plugin_manager.get_plugin_source(plugin) or '(missing)'}"
+        )
```
kenodegard: I'd like us to modify get_plugin_source to accept the plugin itself:


```suggestion
        output.append(
            f"conda-build: {plugin_manager.get_plugin_source(conda_build) or '(missing)'}"
        )
```

conda/cli/main_info.py
```diff
@@ -527,13 +527,18 @@ def _system_component(self) -> str:
...
+        for name, subcommand in sorted(subcommands.items()):
+            plugin = getattr(getattr(subcommand, "impl", None), "plugin", None)
+            output.append(
+                f"conda-{name}: {plugin_manager.get_plugin_source(plugin) or '(unknown)'}"
```
kenodegard:
```suggestion
            output.append(
                f"conda-{name}: {plugin_manager.get_plugin_source(subcommand) or '(unknown)'}"
```

conda/plugins/manager.py
```diff
@@ -157,22 +158,46 @@ def __init__(self, *args, **kwargs):
...
+                f"{module_name(module, plugin_class.__module__)}."
+                f"{plugin_class.__qualname__}[{id(plugin)}]"
+            )
+
+    def get_plugin_source(self, plugin: object) -> str | None:
+        """Return a human-readable source for a registered plugin,
+        or ``None`` if ``plugin`` is ``None``.
+        """
+        if plugin is None:
+            return None
```
kenodegard:
```suggestion
    def get_plugin_source(self, conda_plugin: CondaPlugin) -> str | None:
        """Return a human-readable source for a registered plugin,
        or ``None`` if ``plugin`` is ``None``.
        """
        plugin = getattr(getattr(conda_plugin, "impl", None), "plugin", None)
        if plugin is None:
            return None
```

kenodegard (changes requested):

agriyakhetarpal:
Thanks @kenodegard, makes sense to me!

kenodegard (approved):

jezdez (approved):

PR #16174, opened 2026-06-02
fix return type inconsistency in health check fixers

### LINKED ISSUE #16170: Fix the "reinstall_packages" func to have the correct return type
### Checklist

### What happened?

this function is defined to return an int, but it returns the `install` function that returns None on sucess, and raises exception on failure. 

the `reinstall_packages` function is used as a "fixer" of some health checks. They might also need to be updated.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

ForgottenProgramme:
### Description

resolves #16170 

Add a return statement to `reinstall_packages` to ensure the returned value is an int on success.

### What changed

* Updated `reinstall_packages()` to return 0 at the end.

Failure handling remains unchanged. If `install()` raises an exception, it propagates to the doctor command, where it is caught and converted into a non-zero exit code:

```python
except Exception as err:
    log.warning("Error running fix: %s (%s)", name, err)
    exit_code = 1
```

As a result, fixers now correctly return `0` on success, while failures continue to be reported through exceptions and result in an exit code of `1`.

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -91,6 +91,10 @@ def reinstall_packages(args, specs: list[str], **kwargs) -> None:
     args.prune = kwargs.get("prune", False)
     args.freeze_installed = kwargs.get("freeze_installed", False)
     args.solver_retries = kwargs.get("solver_retries", 0)
+    args.use_local = kwargs.get("use_local", False)
+    args.file = kwargs.get("file", [])
+    args.repodata_fns = kwargs.get("repodata_fns", None)
+    args.update_modifier = kwargs.get("update_modifier", NULL)
 
     install(args, None)
 
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -72,15 +72,13 @@ deprecated.constant(
 )
 
 
-def reinstall_packages(args, specs: list[str], **kwargs) -> int:
+def reinstall_packages(args, specs: list[str], **kwargs) -> None:
     """Reinstall packages using conda install.
 
     Helper for health fixes that need to reinstall packages.
-
     :param args: Parsed arguments namespace
     :param specs: Package specs to reinstall
     :param kwargs: Override default install options (e.g., force_reinstall=True)
-    :return: Exit code from install
     """
     args.packages = specs
     args.channel = kwargs.get("channel", None)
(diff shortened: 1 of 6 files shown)


jezdez (changes requested):
Thanks for digging into this. I think your diagnosis is right: `install()` is not an exit-code-returning API in the normal install path. It succeeds by completing without raising, and failures are already reported by exceptions. So `return install(...)` in `reinstall_packages()` is misleading and can leak `None` into the doctor fixer contract.

I think the fix should sit one layer lower, though. `reinstall_packages()` is already the adapter between the install machinery and doctor fixers, and it is currently typed as returning an `int`. To keep that contract useful for all callers, I’d make it normalize successful completion to `0` directly:

```python
install(args, None)
return 0
```

Then the fixers can keep returning `reinstall_packages(...)`.

I also looked for an existing exception that represents a generic install/reinstall failure, and I don’t think we have one. That’s probably fine here: conda already raises more specific errors for most install failures (`PackagesNotFoundInChannelsError`, `PackagesNotFoundInPrefixError`, `UnsatisfiableError`, `SpecsConfigurationConflictError`, `EnvironmentIsFrozenError`, `LinkError`, etc.). Preserving those signals seems more useful than wrapping them in a generic reinstall failure. The doctor runner already turns exceptions from fixers into a non-zero exit code.

So I think the smallest, clearest version is:

- keep `reinstall_packages() -> int`
- call `install(args, None)`
- return `0` after it completes
- leave the fixers as `return reinstall_packages(...)`

ForgottenProgramme:
> Preserving those signals seems more useful than wrapping them in a generic reinstall failure.

@jezdez , what do you mean by this? 
How is your proposed version avoiding this?

jezdez:
Good question. I only meant that we probably should not add a generic wrapper exception around `install()` here.

Your version and the one I suggested both preserve the original exception, since neither catches `install()` inside `reinstall_packages()`.

The difference I was pointing at is just where we turn success into `0`:

```python
install(args, None)
return 0
```

I’d slightly prefer doing that inside `reinstall_packages()`, since that keeps it as the adapter between `install()` and the doctor fixer return-code contract. Then the fixers can keep returning `reinstall_packages(...)`.

ForgottenProgramme:
> I’d slightly prefer doing that inside `reinstall_packages()`, since that keeps it as the adapter between `install()` and the doctor fixer return-code contract. Then the fixers can keep returning `reinstall_packages(...)`.

@jezdez , agreed.

jezdez:
Do we want to add a test to verify the return value?

jezdez (approved):

ForgottenProgramme:
> Do we want to add a test to verify the return value?

It's already tested, Jannis! 
https://github.com/conda/conda/blob/main/tests/cli/test_cli_install.py#L238

PR #16188, opened 2026-06-03
Add plugin subcommand aliases

### LINKED ISSUE #16187: Support aliases for plugin-provided subcommands
### Checklist

### What is the idea?

Allow conda subcommand plugins to define optional aliases for the commands they provide. This would let plugin-provided commands offer alternate names in the same way conda's built-in commands already support familiar pairs such as `remove` / `uninstall` and `update` / `upgrade`.

### Why is this needed?

Plugin authors currently need to choose a single command name or register multiple separate subcommands for the same behavior. That makes aliases harder to keep consistent in help output, tab completion, and command discovery.

A first-class alias mechanism would make plugin commands easier to discover and use, especially when a plugin wants to support a long descriptive command name plus a shorter common form, or preserve a previous command name while moving users toward a preferred one.

### What should happen?

A plugin should be able to expose a primary subcommand name and one or more alternate names. Users should be able to invoke the command through any of those names and get the same behavior.

Aliases should be discoverable through conda's existing command discovery surfaces, and conda should prevent plugin aliases from shadowing built-in commands or other registered plugin subcommands.

### Additional Context

Related but not duplicate issues found while searching:

- #1043 asks for user-configurable command aliases, which is broader than plugin-declared subcommand aliases.
- #11633 tracks native `conda env` command consolidation and aliases, but does not cover the plugin subcommand hook.

jezdez:
### Description

Closes #16187.

Adds optional aliases to `CondaSubcommand` so plugin-provided CLI commands can register alternate command names through argparse subparser aliases. Alias values are normalized when the plugin type is created, and aliases are rejected during parser configuration when they overlap with built-in conda commands, another plugin subcommand name, or another plugin alias.

This keeps the plugin hook backward compatible because aliases default to an empty tuple.

### Checklist - did you ...


conda/plugins/types.py
```diff
@@ -150,11 +154,23 @@ def __init__(
...
         configure_parser: Callable[[ArgumentParser], None] | None = None,
+        aliases: tuple[str, ...] = (),
     ) -> None:
         super().__init__(name=name)
         self.summary = summary
         self.action = action
         self.configure_parser = configure_parser
+        self.aliases = ()
+        if isinstance(aliases, str):
+            raise PluginError(f"Invalid plugin aliases for {self!r}")
```
soapy1: Small suggestion, could be more specific in this case. Maybe something like

```suggestion
            raise PluginError(f"Invalid plugin aliases for {self!r}. Expected a tuple of strings, received a string.")
```

conda/plugins/types.py
```diff
@@ -150,11 +154,23 @@ def __init__(
...
+        try:
+            self.aliases = tuple(
+                dict.fromkeys(alias.lower().strip() for alias in aliases)
+            )
+        except (AttributeError, TypeError):
+            raise PluginError(f"Invalid plugin aliases for {self!r}")
+        if any(not alias or alias == self.name for alias in self.aliases):
+            raise PluginError(f"Invalid plugin aliases for {self!r}")
```
soapy1: Could also be more specific here. For example
```suggestion
            raise PluginError(f"Invalid plugin aliases for {self!r}. Expected a tuple of strings, received a string.")
```

soapy1: Or, could conda ignore (or strip from the set of aliases) empty aliases or aliases that are the same as the name of the subcommand?

soapy1 (approved):
Neato, nice tests! Just a couple thoughts around the errors for invalid aliases.

conda/cli/conda_argparse.py
```diff
@@ -316,12 +322,46 @@ def configure_parser_plugins(sub_parsers) -> None:
...
+            log.error(
+                dals(
+                    f"""
+                    The plugin '{name}' is trying to register aliases that overlap
+                    with existing conda commands: {", ".join(overlapping_aliases)}
+
+                    Please uninstall the plugin to stop seeing this error message.
+                    """
+                )
+            )
```
kenodegard, line 355: `argparse` will complain if we try to register identical subcommands, can we just intercept that instead of adding this extra logic?

conda/plugins/types.py
```diff
@@ -150,11 +154,34 @@ def __init__(
...
+            raise PluginError(
+                f"Invalid plugin aliases for {self!r}. "
+                "Expected a tuple of strings, received a string."
+            )
+        try:
+            self.aliases = tuple(
+                dict.fromkeys(alias.lower().strip() for alias in aliases)
+            )
+        except (AttributeError, TypeError):
+            raise PluginError(f"Invalid plugin aliases for {self!r}")
```
kenodegard, line 172: eh, in most cases there will only be one alias, why force us to always wrap it into a tuple/list?

```suggestion
        if isinstance(aliases, str):
            aliases = [aliases]
        try:
            self.aliases = tuple(
                dict.fromkeys(alias.lower().strip() for alias in aliases)
            )
        except (AttributeError, TypeError):
            raise PluginError(
                f"Invalid plugin aliases for {self!r}"
                f"Expected a tuple of strings, received a string."
            )
```

conda/cli/conda_argparse.py
jezdez: Good call. I checked this against the supported Python range, and this is Python-version-dependent: newer `argparse` does raise for conflicting subparsers/aliases, but Python 3.10 silently accepts/overwrites them. Since conda still supports Python >=3.10, we cannot rely on argparse directly here.

We could backport the newer argparse conflict check into `_GreedySubParsersAction.add_parser()`, but I think we would still need some preflight logic for plugin-vs-plugin alias conflicts. Otherwise behavior becomes registration-order dependent, effectively “first plugin wins,” instead of rejecting shared aliases or aliases that claim another plugin’s real command name before partial registration.

jezdez:
@kenodegard Any concerns about this?

kenodegard (approved):

PR #16191, opened 2026-06-04
Update conda to use stdlib or backports zstd

### LINKED ISSUE #15486: Use compression.zstd and backports.zstd instead of zstandard
### Checklist

### What happened?

With the inclusion of `compression.zstd`, we have a [better maintained](https://gregoryszorc.com/blog/2024/03/17/my-shifting-open-source-priorities/), built-in alternative to the `zstandard` extension.

Switch to `compression.zstd`, falling back to `backports.zstd` on Python < 3.14.

See also [conda-package-handling](https://github.com/conda/conda-package-handling/issues/305), [conda-package-streaming](https://github.com/conda/conda-package-streaming/issues/137), conda-build?, conda/conda-index#234, https://github.com/conda/conda-libmamba-solver/pull/812

ryanskeith:
### Description

Solves issue #15486.

**Dependency changes**:
- Drop the runtime dependency on third-party zstandard in favor of stdlib compression.zstd (Python ≥3.14) and backports.zstd (Python 3.10–3.13).
- Wire backports.zstd into pyproject.toml, recipe/meta.yaml, and requirements.txt for tests.  Note: Abusing the fact that backports.zstd conda package is available for 3.14.  This may break in the future (see discussion below).

**conda core change**:
- Add conda.common.compression_zstd as a version-gated import shim, exporting zstd, ZstdError, and capped_decompress 
- Repodata .json.zst downloads use ZstdFile + copyfileobj for streaming decompression

**conda._private changes**:
- shard code uses capped_decompress in place of zstandard.decompress (for output + window bounds for untrusted shard data)

### Checklist - did you ...


conda/_private/shards/cache.py
```diff
@@ -175,15 +175,11 @@ def retrieve_multiple(self, urls: list[str]) -> dict[str, ShardDict | None]:
...
-        # In one test reusing the context saves difference between .006s and .01s
-        # We could make this a threadlocal.
-        dctx = zstandard.ZstdDecompressor()
-
         query = f"SELECT url, shard FROM shards WHERE url IN ({','.join(('?',) * len(urls))}) ORDER BY url"
         with self.conn as c:
             result: dict[str, ShardDict | None] = {
                 row["url"]: msgpack.loads(
-                    dctx.decompress(row["shard"], max_output_size=ZSTD_MAX_SHARD_SIZE)
+                    capped_decompress(row["shard"], max_output_size=ZSTD_MAX_SHARD_SIZE)
```
dholth, line 181: `ZSTD_MAX_SHARD_SIZE` exists because the old compression library would flat out refuse to decompress zstd that didn't include the decompressed-size header. To prevent bad shards that are too big it's very cool that you figured out a way to preserve the functionality, but unlike the original, `compression.zstd` would work with or without `max_output_size`.

dev/start
```diff
@@ -334,6 +339,7 @@ if updating; then
         "--file=${_SRC}/tests/requirements.txt" \
         "--file=${_SRC}/tests/requirements-ci.txt" \
         "$([ "${_MACH}" = "Linux" ] && echo "--file=${_SRC}/tests/requirements-Linux.txt")" \
+        ${_PRE314} \
```
dholth: extra stuff?

pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
 ]
 dependencies = [
   "archspec >=0.2.3",
+  "backports.zstd >=1.0.0; python_version < '3.14'",
```
dholth: the other libraries -handling, -streaming probably need this lower bound?

dev/start
ryanskeith: This is to handle backports.zstd dependency for Python < 3.14.

ryanskeith:
pre-commit.ci autofix

pyproject.toml
ryanskeith: I made this call as I saw the different versions of the package on our main channel.  I didn't look if indeed they backports.zstd actually follows semantic versioning.  I did see they had a few 0.x.0 releases so I made the lower bound to be safe. I will look at that deeper though.  It may not be necessary.

ryanskeith: Confirmed breaking changes reported in version 0.5.0. https://github.com/Rogdham/backports.zstd/blob/master/CHANGELOG.md

conda/_private/shards/cache.py
ryanskeith: I did think that we were being protective of the user, so that is why I left it in.  If we don't think it is required, we could take it out and save some lines of code.

dholth: Keep it

recipe/meta.yaml
```diff
@@ -39,6 +39,7 @@ requirements:
     - tqdm >=4
   run:
     - archspec >=0.2.3
+    - backports.zstd >=1.0.0       # [py<314]
```
dholth:
```suggestion
    - backports.zstd >=1.3.0       # [py<314]
```

tests/requirements-pre314.txt
```diff
@@ -0,0 +1,2 @@
+# Python < 3.14 only (stdlib compression.zstd from 3.14 onward; see conda.common.compression_zstd).
+backports.zstd >=1.0.0
```
dholth:
```suggestion
backports.zstd >=1.3.0
```

pyproject.toml
dholth:
```suggestion
  "backports.zstd >=1.3.0; python_version < '3.14'",
```
Let's go with 1.3.0. It is available on main and conda-forge and it fixes [a namespace-package bug.](https://github.com/Rogdham/backports.zstd/blob/master/CHANGELOG.md#130---2025-12-29)

conda/_private/zstd.py
jezdez, line 1: Let's just call it `compression`, `compression_zstd` seems unnecessarily specific and the conda.common module is more generic in module naming.

conda/_private/zstd.py
```diff
@@ -0,0 +1,33 @@
...
+
+import math
+import sys
+
+if sys.version_info >= (3, 14):
+    from compression import zstd
+else:
+    from backports import zstd
+
+ZstdError = zstd.ZstdError  # single symbol for callers to catch
```
jezdez: This is smelly

jezdez:
I think we can simplify the test/dev dependency handling a lot.

`backports.zstd >=1.3.0` can be installed unconditionally in test/dev envs, including Python 3.14. Both recipes intentionally support this pattern:

- In `AnacondaRecipes/backports.zstd-feedstock`, the recipe says Python 3.14+ gets an empty placeholder package: https://github.com/AnacondaRecipes/backports.zstd-feedstock/blob/4a565c8e88d6632cd99b6e603684ca53dd98e4ed/recipe/meta.yaml#L12-L18
- The defaults recipe only runs the build script for `py<314`, and only adds compiler/`zstd`/test deps for `py<314`: https://github.com/AnacondaRecipes/backports.zstd-feedstock/blob/4a565c8e88d6632cd99b6e603684ca53dd98e4ed/recipe/meta.yaml#L20-L48
- conda-forge documents the same split: real builds for 3.10-3.13, empty installable placeholder for 3.14+: https://github.com/conda-forge/backports.zstd-feedstock/blob/dad12f66c2cf43f37d5f741f0609fa2c7fb0c6b4/recipe/recipe.yaml#L18-L25
- conda-forge skips the compiled output on `python >=3.14`: https://github.com/conda-forge/backports.zstd-feedstock/blob/dad12f66c2cf43f37d5f741f0609fa2c7fb0c6b4/recipe/recipe.yaml#L30-L33
- conda-forge then provides the 3.14+ placeholder output that depends only on `python >=3.14` and tests stdlib `compression.zstd`: https://github.com/conda-forge/backports.zstd-feedstock/blob/dad12f66c2cf43f37d5f741f0609fa2c7fb0c6b4/recipe/recipe.yaml#L54-L65

I also checked the published `pkgs/main` records for maintained platforms, and Python 3.14 placeholder builds exist for `linux-64`, `linux-aarch64`, `osx-arm64`, and `win-64`. They are ~9 KB and only depend on `python >=3.14,<3.15.0a0` plus `python_abi 3.14.* *_cp314`.

So instead of carrying `tests/requirements-pre314.txt` and conditionals in `tests.yml`, `dev/start`, `dev/start.bat`, and `.devcontainer/post_create.sh`, we should be able to add this directly to `tests/requirements.txt`:

```text
backports.zstd >=1.3.0
```

Then delete the pre-3.14 requirements file and all the conditional install plumbing. The runtime/package deps can stay conditional in `pyproject.toml` and `recipe/meta.yaml`; this is just for test/dev env setup.

ryanskeith:
Hey @jezdez , it is true that it would be made simpler.  I just have a philosophical disagreement with how that package is packaged.  I don't think it should have been done that way as backports.zstd isn't available for Python 3.14+ and to have it available as a vaporware dependency seems confusing to me and may indeed for other packagers.  

You are right though that maybe simple is better than complex in the test case.  We could make it simple now and then make it complex in the future if the packagers decide to drop support for Python 3.15+.  It will still feel weird when we are installing backports.zstd when Python 3.17 and 3.18 are deployed.

dholth:
Looking forward to repodata v3 and conditional-dependencies support in conda. Otherwise, I don't have a better idea to say "I need either compression.zstd or backports.zstd" other than depending on backports.zstd... without having to forget about noarch python packages.

jezdez:
> Hey @jezdez , it is true that it would be made simpler. I just have a philosophical disagreement with how that package is packaged. I don't think it should have been done that way as backports.zstd isn't available for Python 3.14+ and to have it available as a vaporware dependency seems confusing to me and may indeed for other packagers.

Yeah, I hear you, and I don't know how it was very similarly, but not identically(!), implemented in this odd way in conda-forge and Anaconda. Either way, it would simplify our setup now and that's okay?

> You are right though that maybe simple is better than complex in the test case. We could make it simple now and then make it complex in the future if the packagers decide to drop support for Python 3.15+. It will still feel weird when we are installing backports.zstd when Python 3.17 and 3.18 are deployed.

Hah, true, we should leave a big code comment about this TBH

conda/_private/zstd.py
```diff
@@ -0,0 +1,33 @@
...
+
+__all__ = ["ZstdError", "capped_decompress", "zstd"]
+
+
+def capped_decompress(data: bytes, max_output_size: int) -> bytes:
```
jezdez, line 33: Could we make this a subclass of ZstdDecompressor instead that implements a `capped_decompress` method? That would make it a more descriptive API and allow us to evolve the implementation easier, e.g. in case zstd gains this ability.

dholth: Maybe not because the cap has to be an option passed to the `__init__` method of `ZstdDecompressor()`

It does look like we can pass options [directly into the one-shot API.](https://docs.python.org/3/library/compression.zstd.html#compression.zstd.decompress)

conda/_private/zstd.py
dholth: If I had to shorten it I'd call it `zstd` and not `compression`. Would we expect to put other compression things that conda does like gzip, zipfile, bz2 in here?

dholth: Another change I would make is to duplicate all the APIs that conda calls in here, so that there is no from conda.compression import zstd. Just conda.common.compression.compress, decompress, limited decompress

conda/_private/zstd.py
dholth: Perhaps `from compression.zstd import (symbol, symbol2, symbol3, ZstdError)`

conda/_private/zstd.py
ryanskeith: @jezdez  and @dholth, I wanted zstd originally but that module name was already exists elsewhere in gateways/repodata. I like have unique modules names so I expanded the name. If everyone is ok with it though I can move it to zstd.  We would just have common.zstd and gateways.repodata.zstd.

@dholth I will work on that change.  The major part of the design is to take care of the standard library versus backports so we can be explicit.  I wanted to make it easy for future us to not have to change multiple files.  It will be less code in the modules where the functions are used though, which is a plus.

dholth: Right, I remember, that module deals with `repodata.json.zst`

ryanskeith:
pre-commit.ci autofix

dholth:
[Here are my suggestions](https://github.com/conda/conda/commit/zstd-stdlib) (pushed to a second branch to avoid stepping on your toes). Instead of re-exporting the entire zstd module, we expose everything that conda actually needs to use. Client code changes like so
```
-                from ..._private.zstd import zstd
+                from ..._private import zstd
```

jezdez:
I regret the move into the `conda._private` namespace here, since that means I can't use it downstream in conda plugins, zstandard's options are difficult to get right, so I had hoped we'd offer an opinionated interface for it.

To quote from the `conda._private` module init:

> If you are an API user, stop! This module has no stability guarantees. Symbols under _conda can be renamed or removed at any time. Instead, look for re-exported API under conda.*.

Re-reading it in the context of this PR, I don't think it's helping conda's plugin ecosystem given the large surface area of other symbols in the codebase, Hyrem's law still applies, etc. 

The re-export in a repodata module in [Daniel's branch](https://github.com/conda/conda/commit/16f39ba093cf815bbbb0a4c3b4c8b1c7716351fb) seems not ideal given zstd's broad applications.

Here's what I need to do instead today in conda-workspaces: https://github.com/conda-incubator/conda-workspaces/blob/2415d17e0c1f7964850408a67c2408abc66e9768/conda_workspaces/archive.py#L168-L186

I don't see how zstd is different to the [json](https://github.com/conda/conda/blob/8214671354eb097caa033c644c4f76e415bde070/conda/common/serialize/json.py) or [yaml](https://github.com/conda/conda/blob/8214671354eb097caa033c644c4f76e415bde070/conda/common/serialize/yaml.py) modules in `conda.common`.

My recommendation remains that a `conda.common.compression` is added and a ready-to-use `zstd` symbol provided there.

dholth:
Conda's `json` exists so that we can serialize objects from a library `auxlib` that we are trying to eliminate. It shouldn't exist and neither should `conda.compression.zstd`.

If we were doing this a year ago we would be discussing whether conda should re-export a module that is guaranteed to be https://pypi.org/project/zstandard/ and we would be stuck today. This why we need `_private`. Like plugins or other long-term refactors, `_private`'s potential only materializes over time as enough "stuff" accumulates to add real flexibility when developing conda. Otherwise when we decide to change `compression.zstd` (greater reliance on librattler?) we get bug reports from projects that we were not previously aware of because they are using conda as a compression library instead of for its intended use as an installer, like how Python projects used to use pip as a way to get requests. Should we spend our time making conda a useful installer, or instead of doing that, re-export a collection of the exact libraries that conda uses today to implement an installer?

conda-workspaces does not need help to do a try/except import on compression.zstd and it should not need to implement "degraded mode" when zstd is not available. I wrote degraded mode into conda-package-handling just because it is incredibly inconvenient when the installer doesn't work; but increased use of zstd for not-just-packages, and easier availability of compression.zstd make degraded mode obsolete. Instead of using symbols out of conda, conda-workspaces can consider using either the standard library, or one of the many extant archive libraries, designed only to be an archive library.

jezdez:
Thanks, I think there are two separate questions here.

For this PR, I’m okay keeping the zstd compatibility layer under `conda._private` and not treating it as a supported public API. The linked issue is about replacing `zstandard` inside conda, so I won’t block this PR on adding `conda.common.compression`.

That said, if this is private, I think the abstraction should stay narrow. Re-exporting the selected backend module as `zstd` gives conda internals the whole `compression.zstd` / `backports.zstd` surface, which makes future backend changes harder. Could we apply the shape from 16f39ba, or something equivalent, where `conda._private.zstd` exposes only the operations conda actually uses: `compress`, `decompress`, `ZstdFile`, `ZstdError`, and `capped_decompress`, while keeping the backend module private as `_zstd`?

Separately, I still think there is a valid ecosystem question around whether conda plugins should have access to a supported compression helper. I’ll treat that as follow-up API design rather than something this PR needs to solve accidentally.

conda/_private/zstd.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+
+if sys.version_info >= (3, 14):
+    from compression import zstd
+else:
+    from backports import zstd
+
+ZstdError = zstd.ZstdError  # single symbol for callers to catch
+
+__all__ = ["ZstdError", "capped_decompress", "zstd"]
```
jezdez: This keeps the selected backend private and exposes only the operations conda actually calls.

```suggestion
if sys.version_info >= (3, 14):
    import compression.zstd as _zstd
    from compression.zstd import ZstdError, ZstdFile, compress, decompress
else:
    import backports.zstd as _zstd
    from backports.zstd import ZstdError, ZstdFile, compress, decompress

__all__ = [
    "capped_decompress",
    "compress",
    "decompress",
    "ZstdError",
    "ZstdFile",
]
```

conda/_private/zstd.py
```diff
@@ -0,0 +1,40 @@
...
+def capped_decompress(data: bytes, max_output_size: int) -> bytes:
+    """One-shot decompress of untrusted data with output and window bounds.
+
+    Replaces ``zstandard.decompress(..., max_output_size=...)`` and
+    ``ZstdDecompressor(max_window_size=...)`` (window cap was only in subset.py).
+    """
+    window_log_max = max(10, math.ceil(math.log2(max(max_output_size, 1))))
+    dctx = zstd.ZstdDecompressor(
+        options={zstd.DecompressionParameter.window_log_max: window_log_max}
+    )
```
jezdez: After the backend module is kept private as `_zstd`, use that private backend for the implementation details.

```suggestion
    dctx = _zstd.ZstdDecompressor(
        options={_zstd.DecompressionParameter.window_log_max: window_log_max}
    )
```

conda/gateways/repodata/zstd.py
```diff
@@ -86,17 +86,18 @@ def download_repodata(
...
-                closefd=True,
-            )
-        else:
-            writer = dest_path.open("wb")
-        with writer as repodata:
-            for block in response.iter_content(chunk_size=1 << 14):
-                repodata.write(block)
+        with dest_path.open("wb") as repodata:
+            if is_zst:
+                from ..._private.zstd import zstd
```
jezdez: Import the private helper module rather than a re-exported backend object.

```suggestion
                from ..._private import zstd
```

tests/gateways/test_zstd.py
```diff
@@ -10,10 +10,10 @@
 
 import pytest
 import requests
-import zstandard
 from pytest import MonkeyPatch
 
 import conda.gateways.repodata
+from conda._private.zstd import zstd
```
jezdez: Mirror the new helper-module import in the tests.

```suggestion
from conda._private import zstd
```

jezdez (commented):
Inline suggestions for the narrowed private zstd API shape discussed above.

tests/shards/conftest.py
```diff
@@ -23,9 +23,9 @@
 
 import msgpack
 import pytest
-import zstandard
 
 from conda._private.shards import cache, shards, subset
+from conda._private.zstd import zstd
```
jezdez: Mirror the new helper-module import here too.

```suggestion
from conda._private import zstd
```

tests/shards/test_cache.py
```diff
@@ -8,9 +8,9 @@
 import threading
 
 import pytest
-import zstandard
 
 from conda._private.shards.cache import AnnotatedRawShard, ShardCache
+from conda._private.zstd import zstd
```
jezdez: Mirror the new helper-module import here too.

```suggestion
from conda._private import zstd
```

tests/shards/test_shards.py
```diff
@@ -41,6 +40,7 @@
     fetch_shards_index,
     shard_mentioned_packages,
 )
+from conda._private.zstd import zstd
```
jezdez: Mirror the new helper-module import here too.

```suggestion
from conda._private import zstd
```

jezdez (commented):
Follow-up import suggestions for the narrowed helper module.

jezdez (approved):

Issue #10803, opened 2021-07-20
conda solver ignores virtual package run constraints


chenghlee:
## Current Behavior

conda up to 4.10.2 seems to ignore run constraints that involve virtual packages (`__cuda`, `__glibc`, `__osx`, etc.). This leads to situations where packages are being incorrectly installed onto systems that do not provide the necessary runtime support; for example:

* conda/conda#9896
* conda/conda#9944
* conda/conda#10480

### Steps to Reproduce

These bugs are captured in the following tests in `tests/core/test_solve.py` (that are currently marked as "known broken"):

* [`test_cuda_constrain_sat` and `test_cuda_constrain_unsat`](https://github.com/conda/conda/blob/248741a84/tests/core/test_solve.py#L357-L386)
* [`test_cuda_glibc_unsat_depend` and `test_cuda_glibc_unsat_constrain`](https://github.com/conda/conda/blob/248741a84/tests/core/test_solve.py#L403-L429)

## Expected Behavior

conda's solver should respect run constrains that involve virtual packages, and the tests above should pass.

## Environment Information
<details open><summary><code>`conda info`</code></summary><p>

```
     active environment : base
    active env location : $HOME/Work/Anaconda/conda/env_linux-64
            shell level : 1
       user config file : $HOME/.condarc
 populated config files : 
          conda version : 4.10.3
    conda-build version : 3.21.4
         python version : 3.8.8.final.0
       virtual packages : __cuda=11.0=0
                          __linux=5.4.0=0
                          __glibc=2.27=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : $HOME/Work/Anaconda/conda/env_linux-64  (writable)
      conda av data dir : $HOME/Work/Anaconda/conda/env_linux-64/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : $HOME/Work/Anaconda/conda/env_linux-64/pkgs
                          $HOME/.conda/pkgs
       envs directories : $HOME/Work/Anaconda/conda/env_linux-64/envs
                          $HOME/.conda/envs
               platform : linux-64
[...]

PR #16208, opened 2026-06-08
Enforce constrains on virtual packages in classic solver (#10803)

carterbox:
### Description

A `constrains` referencing a virtual package (e.g. __cuda) was silently ignored: gen_clauses skipped every __-prefixed dep, verify_specs raised ResolvePackageNotFound for unmatchable virtual specs, and the conflict graph never walked virtual constrains. Virtual packages are always present, so such a constraint must bind. Skip a __-spec only when the virtual package is absent from the resolver's index (e.g. bad_installed's installed-only index), and un-skip the four cuda/glibc solver tests.

### Checklist - did you ...

Part of fix for #10803
Companion to https://github.com/conda/conda-libmamba-solver/pull/962

### DIFF
--- a/conda/resolve.py
+++ b/conda/resolve.py
@@ -334,7 +334,15 @@ class Resolve:
         bad_deps.extend(
             (spec,)
             for spec in non_tf_specs
-            if (not spec.optional and not self.find_matches(spec))
+            # Virtual packages (e.g. __cuda) model host state; an unmatchable
+            # virtual spec is a system-compatibility conflict, not a missing
+            # package. Let it flow into SAT/conflict analysis instead of raising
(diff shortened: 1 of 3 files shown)


travishathaway:
Hi @carterbox,

Thanks for submitting this pull request. I've already given it a brief look. I'll bring it up with the rest of the project maintainers to see what they think. Support for the "classic" solver has been languishing a bit since we've made libmamba the default, so that may be the only stumbling block here. Overall, this looks like a reasonable addition, we just need to make sure this doesn't inadvertently break the solver for those still using it.

carterbox:
I'm aware that the "classic" solver is no longer the default, but I wanted to start with the "classic" because the existing tests and the "classic" solver are all in one repository.

PR #16210, opened 2026-06-08
Make conda env create an alias of conda create

### LINKED ISSUE #15576: Deprecate `conda env create` and Make it an Alias
### Checklist

### What is the idea?

After achieving feature parity between `conda create` and `conda env create`, deprecate `conda env create` and make it call `conda create` internally with appropriate defaults. Eventually remove the command entirely after the deprecation period.

### Why is this needed?

Once `conda create` has all the features of `conda env create`, having two commands that do the same thing is confusing and adds maintenance burden. We should consolidate to a single command (`conda create`) and make `conda env create` a simple alias for backward compatibility during the deprecation period.

### What should happen?

1. Add deprecation warning to `conda env create` pointing users to `conda create`
2. Make `conda env create` call `conda create` internally with appropriate defaults
3. Update all documentation to recommend `conda create`
4. Update tests to use `conda create` where appropriate
5. After deprecation period, remove the command entirely

Files to modify:
1. `conda/cli/main_env.py` - Add deprecation warning and call `conda create` internally
2. Documentation files - Update to recommend `conda create`
3. Test files - Update to use `conda create`

Acceptance criteria:
- `conda env create` shows deprecation warning but still works
- `conda env create` behavior matches `conda create` exactly
- Documentation updated to recommend `conda create`
- Tests updated to use `conda create` where appropriate

### Additional Context

Last step of issue #13015.

soapy1:
### Description

Fixes: https://github.com/conda/conda/issues/15576
This pr updates `conda env create` an alias of `conda env`.  ~This should not impact any user facing features of `conda env create`.~

A couple of things this depends on:
* [x] currently conda-pypi post install hooks will try to read environment files. However this does not support remote env files. conda env allows for environment files to be remote.
  * fix: https://github.com/conda/conda-pypi/pull/408
  * will need a release of conda-pypi: https://github.com/conda/conda-pypi/issues/411
* [x] bug fix: pip related actions are included in the json summary, for example when running `conda create . . . --json`. Some possible solutions:
  * https://github.com/conda/conda/pull/16269
  * https://github.com/conda/conda/pull/16270

### A couple of user facing changes:

A few user facing changes are required to make `conda create` and `conda env create` consistent. Some of these changes are not strictly required. Alternative solutions will also be listed.

#### Make `conda create` and `conda env create` dry run output consistent. 

Now they both 
  * have a return code 0 - not a change in behaviour
  * raise a `DryRunExit` - previously `conda env create` would not output any error codes
  * output the same actions report - previously `conda env create` would return a yaml report with all the packages it would have installed.

*Rationale*: This makes `conda env create` more consistent with the behaviour of `conda create` and other existing conda commands. However, this is a breaking change.

*Alternatives*: `conda env create` could keep the existing behaviour.

#### `conda env create` enforces not creating environments inside already existing prefix's 

*Rationale*: This makes `conda env create` more consistent with the behaviour of `conda create`.

*Alternatives*: `conda env create` could keep the existing behaviour.

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -398,9 +398,12 @@ def install(args, parser, command="install"):
 
     defer_json = context.json and bool(env.external_packages)
 
-    conda_actions = handle_txn(
-        unlink_link_transaction, prefix, args, newenv, defer_json_success=defer_json
-    ) or {}
+    conda_actions = (
+        handle_txn(
+            unlink_link_transaction, prefix, args, newenv, defer_json_success=defer_json
+        )
+        or {}
+    )
 
     if env.external_packages and not context.dry_run and not context.download_only:
         from .. import CondaError
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -64,7 +64,6 @@ from .common import check_non_admin
 from .main_config import set_keys
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
     from typing import Any
 
 log = getLogger(__name__)
@@ -401,7 +400,7 @@ def install(args, parser, command="install"):
 
     conda_actions = handle_txn(
         unlink_link_transaction, prefix, args, newenv, defer_json_success=defer_json
-    )
+    ) or {}
 
     if env.external_packages and not context.dry_run and not context.download_only:
         from .. import CondaError
@@ -612,11 +611,6 @@ def handle_txn(
                 ("subdir", context.subdir),
                 path=Path(prefix, DEFAULT_CONDARC_FILENAME),
             )
-
-    follow_up_action_results = {}
-    for action in follow_up_actions:
-        result = action()
-        follow_up_action_results.update(result)
 
     if context.json:
         actions = unlink_link_transaction._make_legacy_action_groups()[0]
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -14,6 +14,7 @@ import os
 from logging import getLogger
 from os.path import abspath
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from ..base.constants import (
     REPODATA_FN,
@@ -61,6 +62,10 @@ from ..reporters import confirm_yn, get_spinner
 from . import common
 from .common import check_non_admin
 from .main_config import set_keys
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Any
 
 log = getLogger(__name__)
 deprecated.constant(
@@ -135,7 +140,7 @@ def get_revision(arg, json=False):
         raise CondaValueError(f"expected revision number, not: '{arg}'", json)
 
 
-def get_index_args(args) -> dict[str, any]:
+def get_index_args(args) -> dict[str, Any]:
     """Returns a dict of args required for fetching an index
 
     Args:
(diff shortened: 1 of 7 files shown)


conda/cli/install.py
```diff
@@ -518,7 +539,14 @@ def revert_actions(prefix, revision=-1, index: Index | None = None):
...
 
 
-def handle_txn(unlink_link_transaction, prefix, args, newenv, remove_op=False):
+def handle_txn(
+    unlink_link_transaction,
+    prefix,
+    args,
+    newenv,
+    remove_op=False,
+    follow_up_actions: list[Callable] = [],
```
soapy1: I'm not really a fan of this approach. The crux of the problem is that we want to be able to get a report of what happens with the `pip` section of the environment.yaml included in the set of actions that conda will output when the `--json` arg is provided.

As we move towards making improvements to (1) how conda is doing reporting and (2) pip interobability with conda-pypi, maybe there will be a more clear path for refactoring this.

ryanskeith: The more I look at it, the more I agree with you that this doesn't feel right with the approach.  The major change has occurred by the PR pointing the execution to `conda.cli.main_create.execute`. What's left is to move features (the json tracking you are adding) and any sort of tweaks to output that we care about (I kinda liked the full solve environment.yaml output myself on conda env create but could be just me OR it could break someone's workflow).

ryanskeith: Perhaps we let handle_txn stay what it is and allow the install to handle the reporting instead of having handle_txn do all of it???

ryanskeith (commented):
It think this is a good attempt and I also agree with @soapy1 that this may not be the best way forward.  The orginal issue stated that these two things commands needed be featured match before aliasing.  In this PR, it is doing the last major feature parity for the json reporting plus the aliasing.  

There are some UX changes that will happen and may break users.  

My suggestion is that we do this in two parts (order to me doesn't matter).  We can handle the aliasing of the command and the json tracking separately.

conda/cli/install.py
soapy1: >Perhaps we let handle_txn stay what it is and allow the install to handle the reporting instead of having handle_txn do all of it???

hmmm, I think the reporting should stay in `handle_txn`. `handle_txn` has a lot of reporting logic in it. It handles:
* what to do if the user is doing a dry run
* what to do if the user needs to confirm an action
* what to output if there are no actions to be done
* outputing to json or plain english 

Further, it is used by:
* all install functions including `conda install`, `conda install --clone`, `conda install --revision`
* `conda remove`

Here is a demo I put together for moving just the json reporting into the `install` function: https://github.com/conda/conda/compare/main...soapy1:conda:add-pip-actions-to-json-output?expand=1. I like this approach less since it breaks up reporting that was previously consolidated. 

What I would really like to see as an improvement to how this reporting works is to be able to record actions that have been executed to a central reporting system. Then the reporting system should consider all the setting that a user has provided (eg. combination of `--dry-run`, `--json`, `--yes`, etc.) and output an appropriate report.

soapy1:
Heya @ryanskeith thanks for your review here! I've added a few extra notes for clarity and broken out seperate prs for the `--json` output issue. 2 approaches for this: https://github.com/conda/conda/pull/16269, https://github.com/conda/conda/pull/16270. Please take a look, I'd love to hear your take on these options, or some other approach.

danyeaw:
pre-commit.ci autofix

ryanskeith (approved):
Aliasing achieved.  Thanks @soapy1 .

PR #16184, opened 2026-06-02
Fix stale negative shard cache rechecks

jezdez:
### Description

Closes #16183.
Refs conda/conda-pypi#397.

This updates shard format probing so cached `has_shards: false` results are only trusted when classic `repodata.json` is already cached as a fallback. If the classic cache is absent, conda now rechecks sharded repodata instead of sending shards-only channels down a missing `repodata.json` fallback path.

The cache decision now lives in `RepodataState.should_check_format("shards")`, so `fetch_shards_index()` consistently reuses that state-level rule.

The failure was exposed by `conda-pypi`, which serves `noarch/repodata_shards.msgpack.zst` but not classic `noarch/repodata.json`. Users with stale `has_shards: false` cache state could see `UnavailableInvalidChannel: HTTP 404` even after shard metadata was available.

The tests cover the shards-only stale-cache path, the classic-cache fallback path, and the mixed-channel case. They also factor the local shard repository writer into a shared helper so the new regression tests reuse the existing shard fixture setup.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] ~Add / update outdated documentation?~

PR #16212, opened 2026-06-09
Fix repodata cache timeout calculation

jezdez:
### Description

Fix the repodata cache timeout calculation and make the cache staleness test deterministic.

The Linux failure seen in #16184 was `tests/gateways/test_repodata_gateway.py::test_stale` missing a tight wall-clock assertion by a few milliseconds:

```
assert 30.103342532 < 30.1
```

That assertion was fragile because the test used real elapsed time between cache save/load operations. While tightening the test, this also exposed that `RepodataCache.timeout()` did not match its docstring: it says it returns seconds until cache timeout, but it added elapsed age to `max-age` instead of subtracting it.

This changes `timeout()` to return the remaining TTL, with `<= 0` for expired caches, and patches the repodata gateway clock in `test_stale` so the expectations are exact rather than scheduler-dependent.

### Checklist - did you ...

- [x] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release release notes?~ Internal cache helper/debug-log correctness and test stability only.


conda/gateways/repodata/__init__.py
```diff
@@ -714,7 +714,7 @@ def timeout(self):
         max_age *= 10**9  # nanoseconds
         now = time.time_ns()
         refresh = self.state.get("refresh_ns", 0)
-        return ((now - refresh) + max_age) / 1e9
+        return (max_age - (now - refresh)) / 1e9
```
jezdez, line 721: @dholth Hey so... this was a surprise to me, do I read this wrong?

dholth: You might be right. conda doesn't use `timeout()` except to print a message, so it's safe to change. I tried using the cache in IPython, but the investigation wasn't very helpful.

dholth: I was able to check this by running a `CONDA_DEBUG=1 conda install --dry-run ...` a few times, which will either say 'cache timed out' or 'timeout in ...', and the timeout was moving in the wrong direction. So you're correct that it should be changed.

jezdez:
Retargeted this PR from `26.5.x` to `main` since we are unlikely to do another `26.5.x` release and this fix should not be stranded on an unreleased branch.

I checked the retargeted diff before changing the base; it remains scoped to the same two files and still contains only the repodata timeout calculation fix plus the deterministic test update.

soapy1 (approved):

PR #16220, opened 2026-06-10
Add test dep pytest-benchmark

soapy1:
### Description

This dependency is required to run the benchmark tests, for example in https://github.com/conda/conda/blob/main/tests/shards/test_shards_subset.py#L471.

This should also fix the failing tests in conda-libmamba-solver ci. For example, https://github.com/conda/conda-libmamba-solver/actions/runs/27306293923/job/80665021404

### Checklist - did you ...

- ~[ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
### DIFF
--- a/conda/testing/fixtures.py
+++ b/conda/testing/fixtures.py
@@ -105,6 +105,22 @@ def suppress_resource_warning():
     warnings.filterwarnings("ignore", category=ResourceWarning)
 
 
+@pytest.fixture(autouse=True)
+def suppress_pytest_bencher_deprecation_warning():
+    """
+    Suppress `PendingDeprecationWarning: FileType is deprecated. Simply open files after parsing arguments.`
+
+    Raised from pytest-bencher.
+
+    xref: https://github.com/ionelmc/pytest-benchmark/issues/283
+    """
+    warnings.filterwarnings(
+        "ignore",
+        message="FileType is deprecated. Simply open files after parsing arguments.",
+        category=PendingDeprecationWarning,
+    )
+
+
 @pytest.fixture(scope="function")
 def tmpdir(tmpdir, request):
     tmpdir = TemporaryDirectory(dir=str(tmpdir))
--- a/tests/requirements-ci.txt
+++ b/tests/requirements-ci.txt
@@ -6,6 +6,7 @@ importlib_resources >= 5.10  # only necessary for Python < 3.12
 pexpect
 pip
 pytest
+pytest-benchmark
 pytest-cov
 pytest-mock
 pytest-rerunfailures


jezdez (approved):

ryanskeith (approved):
Nice find.

soapy1:
@jezdez @ryanskeith looks like there were some issues with pytest-bencher and python 3.14. Pushed up a change. Probably could use a re-review when you have a sec to take another peek.

conda/testing/fixtures.py
```diff
@@ -105,6 +105,22 @@ def suppress_resource_warning():
...
 
+@pytest.fixture(autouse=True)
+def suppress_pytest_bencher_deprecation_warning():
+    """
+    Suppress `PendingDeprecationWarning: FileType is deprecated. Simply open files after parsing arguments.`
+
+    Raised from pytest-bencher.
+
+    xref: https://github.com/ionelmc/pytest-benchmark/issues/283
+    """
```
jezdez, line 116: This is `pytest-benchmark`, not `pytest-bencher`. Bencher is the reporting service; `pytest-benchmark` is the pytest plugin providing the `benchmark` fixture and emitting this warning.

```suggestion
def suppress_pytest_benchmark_deprecation_warning():
    """
    Suppress `PendingDeprecationWarning: FileType is deprecated. Simply open files after parsing arguments.`

    Raised from pytest-benchmark.

    xref: https://github.com/ionelmc/pytest-benchmark/issues/283
    """
```

jezdez (commented):
Context from #15918: we intentionally did not add `pytest-benchmark` to the shared CI requirements when switching from CodSpeed to Bencher. The benchmark plugin was installed only in the dedicated `linux-benchmarks` job so normal unit/integration jobs would not auto-load the plugin, and conda's own matrix was updated to use `not integration and not benchmark` / `integration and not benchmark`.

This failure shows up because the conda-libmamba-solver workflow vendors/runs conda's upstream tests but still selects upstream integration tests with just `-m "integration"`, so the dual-marked `integration` + `benchmark` shard benchmark cases are selected without the plugin installed. Adding `pytest-benchmark` to `requirements-ci.txt` is a reasonable compatibility fix for downstream workflows that may lag conda's marker expression, but let's keep the package name/comment precise.

jezdez:
Follow-up: I think we can avoid the autouse fixture entirely and keep this in pytest config instead.

`pyproject.toml` already centralizes the warning policy under `[tool.pytest.ini_options].filterwarnings`, including turning deprecation warnings into errors. Since this warning comes from the pytest plugin layer, a config-level filter is a better fit than a fixture: it is applied independently of test fixture discovery and keeps the exception next to the broader warning policy.

Suggested change:

```diff
diff --git a/conda/testing/fixtures.py b/conda/testing/fixtures.py
@@
-@pytest.fixture(autouse=True)
-def suppress_pytest_bencher_deprecation_warning():
-    """
-    Suppress `PendingDeprecationWarning: FileType is deprecated. Simply open files after parsing arguments.`
-
-    Raised from pytest-bencher.
-
-    xref: https://github.com/ionelmc/pytest-benchmark/issues/283
-    """
-    warnings.filterwarnings(
-        "ignore",
-        message="FileType is deprecated. Simply open files after parsing arguments.",
-        category=PendingDeprecationWarning,
-    )
-
-
 diff --git a/pyproject.toml b/pyproject.toml
 @@
 filterwarnings = [
   # elevate all deprecation warnings to errors
   "error::PendingDeprecationWarning",
   "error::DeprecationWarning",
   "error::FutureWarning",
+  # pytest-benchmark uses argparse.FileType, deprecated in Python 3.14.
+  "ignore:FileType is deprecated\\. Simply open files after parsing arguments\\.:PendingDeprecationWarning",
 ]
```

I would leave off a module qualifier unless we confirm the emitted warning module from CI. The exact message plus `PendingDeprecationWarning` category is narrow enough, and conda itself does not appear to use `argparse.FileType`.

Issue #16218, opened 2026-06-10
Release `26.5.3`


danyeaw:
### Summary

Placeholder for `conda 26.5.3` release.

| Pilot | @danyeaw  |
|---|---|
| Co-pilot | @travishathaway |

### Tasks

[milestone]: https://github.com/conda/conda/milestone/83
[process]: https://github.com/conda/conda/blob/main/RELEASE.md
[releases]: https://github.com/conda/conda/releases
[main]: https://github.com/AnacondaRecipes/conda-feedstock
[conda-forge]: https://github.com/conda-forge/conda-feedstock
[ReadTheDocs]: https://readthedocs.com/projects/continuumio-conda/
[zulip]: https://conda.zulipchat.com/#narrow/channel/480811-releases

<details open>  
<summary><h4>Patch 26.5.3</h4></summary>

- [x] Create release PR (see [release process][process])
    - [X] https://github.com/conda/conda/pull/16240
- [x] Update Zulip thread on [#releases][zulip]
    - [x] Announce `26.5.3` in-progress [#releases > conda 26.5.x @ 💬](https://conda.zulipchat.com/#narrow/channel/480811-releases/topic/conda.2026.2E5.2Ex/near/599236965)
- [x] [Publish release][releases] https://github.com/conda/conda/releases/tag/26.5.3
- [x] Bump/update feedstocks
    - [x] [Anaconda, Inc.'s feedstock][main]
    - [x] [conda-forge feedstock][conda-forge]
- [x] Hand off to packaging team(s)
- [x] Announce release
    - [x] Post on Zulip thread [#releases > conda 26.5.x @ 💬](https://conda.zulipchat.com/#narrow/channel/480811-releases/topic/conda.2026.2E5.2Ex/near/599254786)
- [x] #16244 

</details>

PR #16240, opened 2026-06-16
Changelog 26.5.3

danyeaw:
✂️ snip snip ✂️ the making of a new release.

Xref #16218


.authors.yml
```diff
@@ -1,6 +1,6 @@
 - name: Michael Sarahan
   email: msarahan@continuum.io
-  num_commits: 665
+  num_commits: 1330
```
jezdez: Looks like this duplicated the counts?

dholth: I use a patched version of rever that counts better especially when someone has e-mail aliases. Which version was used here?

danyeaw: Fixed by using the latest rever from the main branch of its repo with pipx. I'll update the release process.

travishathaway (approved):

Issue #12451, opened 2023-03-06
Add plugin hook for shell interfaces


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### Summary

Shell interfaces are difficult to support as they differ widely, and the conda development team has limited experience with these exotic shells to offer meaningful support.

Let's rely on the plugin infrastructure to allow plugins that extend the supported shell interfaces. This means that most of the shells supported out of the box today will be moved into standalone plugins, and we will request community support for future development efforts.

### Linked Issues & PRs

- [ ] https://github.com/conda/conda/issues/12406
- [ ] https://github.com/conda/conda/issues/11116
- [ ] Convert fish support into a plugin
- [ ] Convert xonsh support into a plugin
- [ ] Convert tcsh/csh support into a plugin

NewtonChutney:
Hello @kenodegard, any update on plugins?
A tentative date of release?

PR #16241, opened 2026-06-16
Elvish support

Noofbiz:
### Description

Added support for the elvish shell to conda. I mostly just copied what was done in a [previous PR](https://github.com/zzamboni/conda) and brought it up-to-date.

### Checklist - did you ...


travishathaway:
Hi @Noofbiz,

Thanks for submitting this pull request. I just wanted to give you some feedback on this and let you know where the project maintainers stand.

We currently don't want to accept more supported shells into the core of conda (it's already enough work maintaining the current set we have!) Instead, we've long been talk about a plugin hook that would allow users to add new shells and also maintain them. You can see some of that work listed here (although it's not that detailed):
- https://github.com/conda/conda/issues/12451

I can't say for sure when this work will go forward, but I encourage you to follow our releases because we will announce it there. Seeing that "elvish" has now been requested twice, this could be a good trial run for such a new plugin hook system.

Noofbiz:
That sounds great; while I was working on this I thought maybe I could do it as a plugin. Thanks for the feedback, and I'll look into making this into a plugin! ^_^

PR #16269, opened 2026-06-22
Add pip actions to json output - approach 1

soapy1:
### Description

closes https://github.com/conda/conda/pull/16270

This pr adds a `pip` section to the json report when running a `conda create -file <path to file> --json`. For example, for the env.yaml
```
name: test
dependencies:
  - python
  - pip:
    - numpy 
```

Previously, the output would be:
```
$ conda create --file tests/env/support/with-pip.yml  --json
{
  "success": true,
  "actions": {
    "FETCH": [],
    "PREFIX": "/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/envs/pip",
    "LINK": [
. . .
    ]
  },
  "prefix": "/path/tp/test"
}
```

Now, the output would be:
```
$ conda create --file tests/env/support/with-pip.yml  --json
{
  "success": true,
  "actions": {
    "FETCH": [],
    "PREFIX": "/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-miniconda/envs/pip",
    "LINK": [
. . .
    ],
  "PIP": [
      "numpy-2.5.0"
    ]
  },
  "prefix": "/path/tp/test"
}
```

In this approach we optionally return the actions from `handle_txn`. When the `install` function gets back a response from `handle_txn`, it can append actions and report the full list of actions taken, including pip installs.

The downside of this approach includes optionally making a function calling `handle_txn` responsible for reporting information back to the user.

### Checklist - did you ...

### DIFF
--- a/conda/cli/install.py
+++ b/conda/cli/install.py
@@ -392,13 +392,18 @@ def install(args, parser, command="install"):
                     raise CondaImportError(str(e))
                 raise e
 
-    handle_txn(unlink_link_transaction, prefix, args, newenv)
+    defer_json = context.json and bool(env.external_packages)
+
+    conda_actions = handle_txn(
+        unlink_link_transaction, prefix, args, newenv, defer_json_success=defer_json
+    )
 
     if env.external_packages and not context.dry_run and not context.download_only:
         from .. import CondaError
         from ..env.installers.base import get_installer
         from ..env.pip_util import get_pip_workdir
 
+        installer_results = {}
         external_envs = [
             (fpath, file_env)
             for fpath, file_env in fpath_envs_map.items()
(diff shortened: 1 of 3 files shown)


jezdez:
Thanks for splitting this out. Between this and #16270, I prefer this direction.

The important boundary, IMO, is that `install()` already owns the environment-file context and the external installer orchestration, while `handle_txn()` owns conda transaction execution/reporting. I'd avoid making `handle_txn()` run arbitrary follow-up callbacks as in #16270; that makes transaction handling responsible for pip/env-file side effects it does not really understand.

That said, I think this PR should tighten the `handle_txn()` contract a bit before merging. `return_actions=True` currently makes `handle_txn()` sometimes return an action mapping and sometimes `None` on the existing no-op path for non-create operations. Since `install()` then does `conda_actions.update(...)`, that can fail when there are external packages but no conda transaction work.

What I'd suggest:

- Keep this approach as the base, not #16270.
- Rename/narrow the option to express the actual behavior, e.g. `defer_json_success=True` or `report_success=False`, rather than a generic `return_actions`.
- Make the deferred-reporting path always return an action mapping, using `{}` for no-op conda transactions.
- Keep pip installation and pip action merging in `install()`.
- Emit the single final JSON success from `install()` after pip/external package work has completed.

That preserves the existing responsibility split while fixing the real issue: final JSON is currently emitted before pip actions exist.

conda/cli/install.py
```diff
@@ -392,13 +392,20 @@ def install(args, parser, command="install"):
                     raise CondaImportError(str(e))
                 raise e
 
-    handle_txn(unlink_link_transaction, prefix, args, newenv)
+    defer_json = (
+        context.json and bool(env.external_packages) and not context.download_only
```
ryanskeith: This is interesting.  What was the thinking on adding the `not context.download_only` clause?

ryanskeith (approved):
Thank you for looking at this a different way.  I like how these changes are minimal.

conda/cli/install.py
soapy1: I think this is not strictly necessary. `handle_txn` will cause the program to exit early with a `CondaExitZero` error if `context.download_only` is set.

Can remove it, thanks for checking!

PR #16276, opened 2026-06-24
Convert `InvalidInstaller` into a `CondaError` subclass

kenodegard:
### Description

- Make `InvalidInstaller` a `CondaError` with structured `ErrorGuidance`, so missing external installers use the normal conda error path instead of being caught and re-wrapped as a generic `CondaError` in three CLI modules.
- Centralize user-facing messaging in `InvalidInstaller`: summary plus actionable hints (spelling check; optional `conda-env-{name}` plugin hint when no env file path is available).
- Extend `get_installer()` with an optional `file=` argument so `conda install` can include the environment file path in the error summary.
- Remove redundant `except InvalidInstaller: raise CondaError(...)` blocks from `install.py`, `main_env_create.py`, and `main_env_update.py`.
- Fix `conda env update` reporting a literal `{0}` placeholder when an external installer is unavailable (broken f-string in the old error message).

Terminal output for these failures now uses the structured guidance format (summary + “Next steps” hints), consistent with other guided errors. Content is preserved; `--json` output gains structured `guidance` with stable `hint_code`s.

Identified while reviewing https://github.com/conda/conda/pull/15880

### Checklist - did you ...


conda/exceptions.py
```diff
@@ -1746,10 +1747,36 @@ def _format_exc(
...
 
-class InvalidInstaller(Exception):
-    def __init__(self, name: str):
+class InvalidInstaller(CondaError):
+    def __init__(self, name: str, *, file: str | None = None):
         msg = f"Unable to load installer for {name}"
-        super().__init__(msg)
+        hints: list[GuidanceHintTypedDict] = [
+            {
+                "text": "Please ensure your dependencies file has the correct spelling.",
```
soapy1: I think an `InvalidInstaller` error should have more to do with the user trying to use an installer that doesn't exist more than dependencies. And I haven't seen 'dependencies file' phrase in other places. So I think this message could be improved some. How about something like

```suggestion
                "text": "Please ensure you are requesting an available installer.",
```

Additionally, it might be nice to list out available installers. Though since those are not registered thru the plugin manager, it might be clunky. Probably good to punt on that?

conda/cli/main_env_update.py
```diff
@@ -151,21 +149,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-                    Unable to install package for {0}.
-
-                    Please double check and ensure you dependencies file has
-                    the correct spelling.  You might also try installing the
-                    conda-env-{0} package to see if provides the required
-                    installer.
-                    """
-                )
-            )
+        installers[installer_type] = get_installer(installer_type)
```
soapy1:
```suggestion
        installers[installer_type] = get_installer(installer_type, file=args.file)
```

 I think it would be good for this to be consistent with the conda install/update/create commands

conda/cli/main_env_create.py
```diff
@@ -210,30 +208,14 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
...
-                        f"""
-                        Unable to install package for {installer_type}.
-
-                        Please double check and ensure your dependencies file has
-                        the correct spelling. You might also try installing the
-                        conda-env-{installer_type} package to see if provides
-                        the required installer.
-                        """
-                    )
+            installer = get_installer(installer_type)
```
soapy1:
```suggestion
            installer = get_installer(installer_type, file=args.file)
```

I think it would be good for this to be consistent with the conda install/update/create commands

conda/exceptions.py
```diff
@@ -1746,10 +1747,36 @@ def _format_exc(
...
+            summary = f"Unable to install package for {name}."
+            hints.append(
+                {
+                    "text": (
+                        f"You might also try installing the conda-env-{name} package "
+                        "to see if provides the required installer."
+                    ),
+                    "hint_code": "install_conda_env_plugin",
+                },
+            )
```
soapy1:
```suggestion
            
```

I don't think this advice will work in any case. There are no installer plugins. And the `get_installer` function [imports installers from the `conda.env` module directly](https://github.com/conda/conda/blob/main/conda/env/installers/base.py#L20)

soapy1 (commented):
Just a few suggestions to improve the error messages. Cool to see the error hints in action!

soapy1 (approved):

PR #16284, opened 2026-06-25
Update Miniconda installer links

### LINKED ISSUE #16281: Install links to Miniconda have poor user experience
### Checklist

### What happened?

The installer pages link out to anaconda.com/docs for Miniconda install instructions. It ends up being a convoluted experience for the user, going from conda.io to anaconda.com/docs to anaconda.com/download to anaconda.com/download/success. The Miniconda installer links should just go straight to anaconda.com/download/success, since that's what the Miniconda install button on anaconda.com/download does (scroll to the bottom of the page).

### Additional Context

_No response_

kathatherine:
### Description

Closes https://github.com/conda/conda/issues/16281.

I just updated the Miniconda links to also go to the anaconda.com site. We may decide we want to do something different here, but this was the most consistent way to link out without taking people around a million links to get to their destination.

### Checklist - did you ...

### DIFF
--- a/docs/source/user-guide/install/linux.rst
+++ b/docs/source/user-guide/install/linux.rst
@@ -4,11 +4,11 @@ Installing on Linux
 
 #. Download the installer:
 
-   * `Miniconda installer for Linux <https://docs.anaconda.com/miniconda/>`__.
+   * `Miniconda installer for Linux <https://www.anaconda.com/download/success>`__.
 
-   * `Anaconda Distribution installer for Linux <https://www.anaconda.com/download/>`__.
+   * `Anaconda Distribution installer for Linux <https://www.anaconda.com/download>`__.
 
-   * `Miniforge installer for Linux <https://conda-forge.org/download/>`_.
+   * `Miniforge installer for Linux <https://conda-forge.org/download>`_.
 
 #. :ref:`Verify your installer hashes <hash-verification>`.
 
--- a/docs/source/user-guide/install/linux.rst
+++ b/docs/source/user-guide/install/linux.rst
@@ -4,7 +4,7 @@ Installing on Linux
 
 #. Download the installer:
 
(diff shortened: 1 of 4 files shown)


docs/source/user-guide/install/linux.rst
```diff
@@ -4,11 +4,11 @@ Installing on Linux
 
 #. Download the installer:
 
-   * `Miniconda installer for Linux <https://docs.anaconda.com/miniconda/>`__.
+   * `Miniconda installer for Linux <https://www.anaconda.com/download/success>`__.
```
danyeaw: Hey @kathatherine, it looks like the success link shows both the Anaconda Distribution and Miniconda links. Could we combine the links to either go to `/download` if we want to offer registration, or to `/download/success` if we want to go directly to the download page?

kathatherine: I'm fine with linking to /download directly, but the way that page is currently structured, the Miniconda link is all the way down at the bottom of the page and no one can find it (there's no anchor link option for that header either). Linking to download/success bypasses registration entirely, which seems to be fine for Miniconda (if you look at the button for Miniconda at the very bottom of the page, it links to download/success anyway). The Anaconda docs team has been asked to not bypass registration with our links, but conda is open source, so ¯\_(ツ)_/¯

danyeaw: I don't think we should bypass it either, since Miniconda isn't any less if a product than Anaconda Distribution, they both use the same main channel.

Users can either register, click on the skip registration button, or as you mentioned scroll all the way to the bottom for a separate Miniconda link.

kathatherine: Fair enough! It's at least slightly better than having to go to the Anaconda docs first. I'll update that.

danyeaw (approved):

danyeaw:
Thanks @kathatherine!

PR #16315, opened 2026-06-30
Use os.path.splitext to get win destination trash file

### LINKED ISSUE #15760: AttributeError at runtime for .splitext method on string / pathlib.Path
### Checklist

### What happened?

```
while isfile(dest_fn):
    dest_fn = dest_fn.splitext[0] + f".conda_trash_{counter}"
```
https://github.com/conda/conda/blob/main/conda/gateways/disk/delete.py#L159

After a quick five-minute inspection, I'm not 100% sure what the most elegant fix would be, but it looks like splitext should have been imported from os.path, and the line inside the while loop should probably be `dest_fn = splitext(dest_fn)[0] + f".conda_trash_{counter}"`.

Current code tries to call .splitext as if it were a string (maybe pathlib.Path?) method, which doesn't exist → AttributeError at runtime.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

soapy1:
### Description

Python strings do not have a `.splitext` function. This funciton is available in the `os.path` module. This pattern is used elsewhere in conda code, for example https://github.com/conda/conda/blob/main/conda/core/subdir_data.py#L327.

fixes https://github.com/conda/conda/issues/15760

### Checklist - did you ...

### DIFF
--- a/news/16315-dont-use-splitext-on-string
+++ b/news/16315-dont-use-splitext-on-string
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix windows `AttributeError` at runtime for for `.splitext` method. Use `os.path.splitext` to get the filename from strings instead. (#15760 via #16315)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/16315-dont-use-splitext-on-string
+++ b/news/16315-dont-use-splitext-on-string
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Fix windows `AttributeError` at runtime for for `.splitext` method. Use `os.path.splitext` to get the filename from strings instead. (#15760 via #16315)
+* Fix Windows runtime `AttributeError` from calling `.splitext` on strings. Use `os.path.splitext` instead. (#15760 via #16315)
 
 ### Deprecations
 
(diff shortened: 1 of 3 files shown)


tests/gateways/disk/test_delete.py
```diff
@@ -167,3 +171,39 @@ def test_try_rmdir_all_empty_doesnt_exist():
         assert isdir(td)
         rm_rf(td)
         assert not isdir(td)
+
+
+def test_unlink_to_rename_to_trash_win_fallback(
```
soapy1, line 176: I'm not sure if we are reaching silly levels of mocking here. But in order to exercise the previously broken piece of code, we need to force some `OSError`'s.

news/16315-dont-use-splitext-on-string
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix windows `AttributeError` at runtime for for `.splitext` method. Use `os.path.splitext` to get the filename from strings instead. (#15760 via #16315)
```
jezdez: Small wording cleanup for the news fragment.

```suggestion
* Fix Windows runtime `AttributeError` from calling `.splitext` on strings. Use `os.path.splitext` instead. (#15760 via #16315)
```

jezdez (approved):
Looks good. I left one small wording suggestion for the news fragment.

tests/gateways/disk/test_delete.py
jezdez: Level "silly" is a perfectly valid engineering excellence level. It's better than "uh uh" and "worked on my machine".

PR #16311, opened 2026-06-30
Speed up `test_dont_remove_conda` tests

danyeaw:
### Description

This PR speeds up the current slowest test and closes #16299:

1. Does a single 15sec set up of the test, which is then reused for other parameterized tests
2. Moves from 625s to 147s total test time, with 125s being classic solver run time

I also removed conda-build from the test because it made up 459s of the original test. It is a very heavy dependency to run tests with.

Overall 4x increase in test speed.

### Checklist - did you ...

- [X] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #16316, opened 2026-07-01
Add conda dependency removal coverage with reverse dependents

jezdez:
### Description

Adds root-prefix CLI coverage for the conda dependency removal guard when a conda dependency also has installed reverse dependents.

#16311 removed the expensive `conda-build` install from `test_dont_remove_conda`. This test uses existing fixture packages from `tests/data/test-recipes` (`another_dependent -> dependent -> dependency`) and temporarily models `dependency` as a conda runtime dependency. It asserts that removing `dependency` raises `RemoveError` and leaves conda plus the dependent fixture packages installed.

This is not the same package graph as the old `conda-build` / `pycosat` setup; it covers the related root-prefix CLI behavior without reinstalling `conda-build`.

### Checklist - did you ...

- [x] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2358,7 +2358,6 @@ def test_dont_remove_conda_dependency_with_dependent_packages(
     conda_cli: CondaCLIFixture,
 ):
     """Removing a conda dependency is blocked even when dependents would be unlinked."""
-    assert test_recipes_channel.is_dir()
     conda_cli(
         "install",
         f"--prefix={root_prefix_with_conda}",
@@ -2393,6 +2392,7 @@ def test_dont_remove_conda_dependency_with_dependent_packages(
             )
 
         assert any(isinstance(e, RemoveError) for e in exc.value.errors)
+        assert package_is_installed(root_prefix_with_conda, "conda")
         assert package_is_installed(root_prefix_with_conda, "dependency")
         assert package_is_installed(root_prefix_with_conda, "dependent")
         assert package_is_installed(root_prefix_with_conda, "another_dependent")
@@ -2400,6 +2400,18 @@ def test_dont_remove_conda_dependency_with_dependent_packages(
         conda_record["depends"] = original_depends
         conda_meta.write_text(json.dumps(conda_record))
         PrefixData._cache_.clear()
+        cleanup_packages = [
+            package
+            for package in ("another_dependent", "dependent", "dependency")
+            if package_is_installed(root_prefix_with_conda, package)
+        ]
+        if cleanup_packages:
+            conda_cli(
+                "remove",
+                f"--prefix={root_prefix_with_conda}",
+                *cleanup_packages,
+                "--yes",
+            )
 
 
 def test_dont_remove_conda_3(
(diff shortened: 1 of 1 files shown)


travishathaway (approved):

tests/test_create.py
```diff
@@ -2352,6 +2352,56 @@ def test_dont_remove_conda(
     assert package_is_installed(root_prefix_with_conda, "pycosat")
 
 
+def test_dont_remove_conda_dependency_with_dependent_packages(
+    root_prefix_with_conda: Path,
+    test_recipes_channel: Path,
+    conda_cli: CondaCLIFixture,
+):
+    """Removing a conda dependency is blocked even when dependents would be unlinked."""
+    assert test_recipes_channel.is_dir()
```
danyeaw:
```suggestion
```

I don't think this assert is doing much here.

tests/test_create.py
```diff
@@ -2352,6 +2352,56 @@ def test_dont_remove_conda(
...
+        with pytest.raises(CondaMultiError) as exc:
+            conda_cli(
+                "remove",
+                f"--prefix={root_prefix_with_conda}",
+                "dependency",
+                "--yes",
+            )
+
+        assert any(isinstance(e, RemoveError) for e in exc.value.errors)
+        assert package_is_installed(root_prefix_with_conda, "dependency")
```
danyeaw, line 2398:
```suggestion
        assert any(isinstance(e, RemoveError) for e in exc.value.errors)
        assert package_is_installed(root_prefix_with_conda, "conda")
        assert package_is_installed(root_prefix_with_conda, "dependency")
```

Adding a check for conda here would be more explicit.

tests/test_create.py
```diff
@@ -2352,6 +2352,56 @@ def test_dont_remove_conda(
...
+        assert package_is_installed(root_prefix_with_conda, "dependent")
+        assert package_is_installed(root_prefix_with_conda, "another_dependent")
+    finally:
```
danyeaw, line 2401: I noticed we aren't cleaning up the packages we installed in this session scoped fixture. It might not be worth the time it takes since this fixture is only being used by the tests above, but might be worth a note in case someone tries to use the fixture later.

danyeaw (approved):
Hey @jezdez, thanks for this! I think it would be good to update the PR description and title since this reverse dependency test you added in this PR was never explicitly tested for in the original tests. I think this PR enhances the original tests, which is great.

tests/test_create.py
jezdez: I checked the nearby/shared fixture patterns and I think your instinct here is right. I don't see a package-level cleanup pattern for session_tmp_env; most tests either use function-scoped tmp_env and let the temp prefix go away, or create a fixed shared env and leave it alone. So I removed the cleanup transaction and kept the finally block focused on restoring the conda metadata/cache. I also added a short note that these lightweight packages are intentionally left installed because this session-scoped prefix is local to this test group.

PR #16317, opened 2026-07-01
Improve Bencher PR signal

jezdez:
### Description

Bencher PR comments currently provide little signal when the benchmark artifact exists but no useful boundary can be calculated. Two things contribute to that:

- benchmark tracking only ran after the entire `Tests` workflow succeeded, so valid benchmark artifacts from otherwise-failed test runs did not help build base-branch history;
- the PR tracking command defined thresholds directly on PR branches, which made `--ci-only-thresholds` post comments even when no threshold or benchmark history had been cloned from the base branch.

This updates the benchmark tracking workflow to consume benchmark artifacts from successful or failed `Tests` runs, while still letting artifact discovery decide whether benchmark data exists. PR tracking now relies on thresholds cloned from the base branch and only starts posting PR comments when Bencher raises an alert. The Bencher check and job summary remain available for every tracked run.

The `Tests` workflow code-change filter now includes `.github/workflows/benchmarks.yml`, so benchmark workflow edits exercise the benchmark-producing job in PR CI.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?~ (CI-only workflow change)
- [x] Add / update necessary tests? (validated workflow YAML and `benchmarks.yml` with `actionlint`)
- [ ] ~Add / update outdated documentation?~ (no user-facing docs)

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -17,10 +17,7 @@ jobs:
   track-base-branch:
     if: >-
       !github.event.repository.fork
-      && (
-        github.event.workflow_run.conclusion == 'success'
-        || github.event.workflow_run.conclusion == 'failure'
-      )
+      && contains(['success', 'failure'], github.event.workflow_run.conclusion)
       && github.event.workflow_run.event == 'push'
       && (
         github.event.workflow_run.head_branch == 'main'
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -91,10 +91,7 @@ jobs:
   track-pull-request:
     if: >-
       !github.event.repository.fork
-      && (
-        github.event.workflow_run.conclusion == 'success'
-        || github.event.workflow_run.conclusion == 'failure'
-      )
+      && contains(['success', 'failure'], github.event.workflow_run.conclusion)
       && github.event.workflow_run.event == 'pull_request'
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -17,7 +17,10 @@ jobs:
   track-base-branch:
     if: >-
       !github.event.repository.fork
-      && github.event.workflow_run.conclusion == 'success'
+      && (
+        github.event.workflow_run.conclusion == 'success'
+        || github.event.workflow_run.conclusion == 'failure'
+      )
```
jaimergp: This can be expressed more succinctly with [`contains`](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#contains) if you like it:

```suggestion
      && contains(['success', 'failure'], github.event.workflow_run.conclusion)
```

.github/workflows/benchmarks.yml
```diff
@@ -91,7 +91,10 @@ jobs:
   track-pull-request:
     if: >-
       !github.event.repository.fork
-      && github.event.workflow_run.conclusion == 'success'
+      && (
+        github.event.workflow_run.conclusion == 'success'
+        || github.event.workflow_run.conclusion == 'failure'
+      )
```
kenodegard:
```suggestion
      && contains(['success', 'failure'], github.event.workflow_run.conclusion)
```

kenodegard:
~Is this the benchmark.yml workflow one we can exclude from the test.yml workflow's changes check?~

Never mind I see its being explicitly added here

kenodegard (approved):

Issue #16185, opened 2026-06-03
Discovery: Move the classic solver to its own conda repo


danyeaw:
### What is the idea?
Move the classic solver out of the conda repository into its own standalone repository (e.g. conda/conda-classic-solver), registered as an external plugin via the existing conda_solvers hook — the same pattern already in use for conda-libmamba-solver.

### Why is this needed?
The classic solver accounts for roughly 4,200 lines of code, plus several thousand more lines of dedicated tests. Keeping it in-tree adds CI overhead for every change to conda core, even changes that are unrelated to solving. Extracting it reduces the surface area of the core repo, shortens CI times for both projects, and makes it easier to iterate on the solver independently.

### What should happen?

- A new conda-classic-solver repository is created under the conda GitHub organization.
- The classic solver modules (solve.py, resolve.py, logic.py, _logic.py) and their dedicated tests are moved there.
- The new package registers itself via the conda_solvers entry point, making it a drop-in external plugin.
- The conda package declares conda-classic-solver as an install-time dependency so existing users see no behavioral change.
- The classic solver name continues to work and conda core falls back to or requires the external package instead of an in-tree implementation.

### Additional Context
The plugin infrastructure is already in place. The conda_solvers hook in conda/plugins/hookspec.py and the CondaSolver type in conda/plugins/types.py define the contract. conda-libmamba-solver follows this exact pattern and can serve as a reference for repository structure, CI setup, and packaging. The rattler solver also appears to follow the same external plugin model. The main work is the extraction, packaging, and wiring up the dependency — no new plugin APIs are needed.

The idea for this was part of the 26.5.x release retro.

kenodegard:
### Findings

1. **`conda-classic-solver` already exists,** its just frozen in time (created 2 years ago)

1. **`remove-classic` is stale, not unusable.** Hundreds of commits behind `main`, but the real delta is small (~14 files / ~170 lines of intent). Rebase manually.

2. **Two codebases to bring current.** Conda prep/deprecation is one track; `conda-classic-solver` is another. CCS was cut from an old snapshot and breaks on modern conda (e.g. `Spinner` gone, stale index APIs).

5. **Testing: `pytest-conda-solvers` first.** Don’t prioritize moving classic tests out of conda, or cloning conda to run its suite the way conda-libmamba-solver / conda-rattler-solver do. Get **pytest-conda-solvers** up and running first; that shared harness should drive solver plugin testing going forward.

### Prep work extracted from `remove-classic` branch

[...]

PR #16320, opened 2026-07-01
Split off `BaseSolver` from `Solver`

kenodegard:
### Description

This is prep-work for removing the classic (pycosat) solver from conda (and relying instead on the conda/conda-classic-solver plugin if the classic/pycosat solver is desired).

Xref https://github.com/conda/conda/issues/16185
Xref https://github.com/conda/conda/pull/14170

### Checklist - did you ...


jezdez (approved):

jezdez:
@jaimergp Any concerns with this?

conda/core/solve.py
```diff
@@ -253,6 +253,72 @@ def solve_for_diff(
...
+
+            # only look for a newer conda in the channel conda is currently installed from
+            conda_newer_spec = MatchSpec(f"{channel_name}::conda>{CONDA_VERSION}")
+
+            if paths_equal(self.prefix, context.conda_prefix):
+                if any(conda_newer_spec.match(prec) for prec in link_precs):
+                    return
+
+            conda_newer_precs = sorted(
+                SubdirData.query_all(
```
jaimergp, line 285: Both CLS and CRS have their own way of querying the index without using SubdirData, not sure how useful this is as is. Maybe we can provide an non-implemented method that needs to provide the latest `conda` version for the given `(channels, subdirs, repodata_fn)`?

jaimergp:
LGTM but I do have some thoughts. Comparing to https://github.com/conda/conda/pull/14170:

- Do we want to isolate the `resolve` imports more?
- Not sure what to do about the conda update checks. It feels that the Solver itself is not the right place to do it. Maybe a plugin hook? Not a blocker here though.

kenodegard:
> 1. Do we want to isolate the `resolve` imports more?

Made the `resolve` import inline in https://github.com/conda/conda/pull/16320/commits/ed98f2e1eb96bde10fe50daafb060beffdcc9a4f.

> 2. Not sure what to do about the conda update checks. It feels that the Solver itself is not the right place to do it. Maybe a plugin hook? Not a blocker here though.

I agree `_notify_conda_outdated` doesn’t really belong on the solver but I think changing it here is out of scope.

PR #16328, opened 2026-07-01
Preserve relocatable build-prefix shebangs

### LINKED ISSUE #16327: Conda generates non-relocatable launchers when built with rattler-build
## Checklist

## What happened?

When the `conda` package is built with rattler-build, the generated `bin/conda` and `condabin/conda` launchers use this shebang:

```text
#!/usr/bin/env python
```

Because the build prefix is no longer present in these files, rattler-build cannot record them for prefix relocation. After installation, the launchers therefore use whichever Python is first on `PATH`, rather than the Python belonging to the environment where conda is installed.

This breaks conda after activating a child environment that has Python but does not have conda. The launcher selects the child environment's Python and fails with:

```text
```

This was reported in [conda-forge/conda-feedstock#304](https://github.com/conda-forge/conda-feedstock/issues/304) and [conda-forge/conda-feedstock#305](https://github.com/conda-forge/conda-feedstock/issues/305) after [conda-forge/conda-feedstock#300](https://github.com/conda-forge/conda-feedstock/pull/300) migrated the feedstock to rattler-build. [conda-forge/conda-feedstock#306](https://github.com/conda-forge/conda-feedstock/pull/306) adds a workaround in the feedstock, but the behavior we need to fix is in conda's launcher generation.

### Minimal reproducer

This reproduces on Linux using the affected conda-forge build:

```bash
rm -rf /tmp/conda-rattler-repro /tmp/conda-rattler-child

conda create --yes \
  --prefix /tmp/conda-rattler-repro \
  --override-channels \
  --channel conda-forge \
  "conda=26.5.3=py314h9e666f3_1"

eval "$(/tmp/conda-rattler-repro/bin/python -m conda shell.bash hook)"
conda activate /tmp/conda-rattler-repro

conda create --yes \
  --prefix /tmp/conda-rattler-child \
  --override-channels \
  --channel conda-forge \
  python

conda activate /tmp/conda-rattler-child
conda --version
```

The final command should print the conda version. Instead, it raises `ModuleNotFoundError`.

### Root cause

`conda init --install` calls `generate_shebang_for_entry_point()` with `with_usr_bin_env=True`. Conda normally converts a long Python path to `/usr/bin/env python`, but it preserves conda-build's intentionally long, relocatable shebang with this special case:

```python
if os.environ.get("CONDA_BUILD") == "1" and "/_h_env_placehold" in executable:
    return shebang
```

Rattler-build also sets `CONDA_BUILD=1`, but its host prefix is named `/host_env_placehold...` rather than `/_h_env_placehold...`. The condition does not match, so conda replaces the absolute shebang before rattler-build can detect and record the prefix.

One option would be to modify rattler-build to use the same prefix as conda-build, but that implies that all package-building tools would need to match a specific interface for their paths. Instead, we should generalize the behavior of `conda init --install` to avoid path conventions specific to conda-build.

The behavior could be made independent of a build tool's temporary directory naming convention by preserving the shebang when:

- `CONDA_BUILD=1`, and
- the executable is contained within the build's `$PREFIX`.

The containment check is important because `CONDA_BUILD=1` can be inherited by commands that create unrelated environments during a build. Long shebangs outside `$PREFIX` should continue to use conda's normal fallback behavior.

I tested this approach with both the conda-build and rattler-build prefix forms. I also built and tested the conda feedstock end to end with unmodified rattler-build. Both launchers were recorded as relocatable text files, and conda continued to work after activating a child environment with its own Python.

## Conda Info

Not specific to the reporting environment. The failure is present in the published Linux package `conda-26.5.3-py314h9e666f3_1.conda`.

## Conda Config

No custom configuration is required.

## Conda list

n/a

bdice:
### Description

Closes #16327.

When conda is built with rattler-build (used in conda-forge since https://github.com/conda-forge/conda-feedstock/pull/300), `generate_shebang_for_entry_point()` does not recognize rattler-build's `host_env_placehold...` host prefix. It replaces conda's absolute Python shebang with `#!/usr/bin/env python`, leaving the package without the prefix metadata needed to relocate its launchers at installation time.

During a build, conda now preserves the absolute shebang when its Python executable is contained within `$PREFIX` instead of assuming conda-build's `_h_env_placehold...` naming convention for the host prefix.

The tests cover both conda-build and rattler-build prefix forms, as well as unrelated and `..`-escaped paths outside the build prefix. I also tested the change end to end by building the conda feedstock with unmodified rattler-build. Both conda launchers were recorded as relocatable text files, and conda continued to work after activating a child environment with a different Python.

### Checklist - did you ...

- [x] ~Add / update outdated documentation?~ No documentation changes are needed for this internal build-time fix.


jezdez (commented):
I did not find a reportable security issue, but I found one backward-compatibility issue in the shebang preservation logic.

conda/core/portability.py
```diff
@@ -384,12 +384,19 @@ def generate_shebang_for_entry_point(
...
+        build_prefix = os.path.abspath(build_prefix)
+        try:
+            executable_in_prefix = (
+                os.path.commonpath((executable_path, build_prefix)) == build_prefix
+            )
+        except ValueError:
+            executable_in_prefix = False
+        if executable_in_prefix:
+            # The build tool will relocate this prefix at installation time.
+            return shebang
```
jezdez, line 464: This looks a bit too broad for backward compatibility. It preserves the rattler-build prefix case, but it also returns a raw shebang for any executable under `$PREFIX` whenever `CONDA_BUILD=1`, so long non-placeholder prefixes or prefixes containing spaces skip the fallback below.

This keeps the new rattler-build marker while preserving the old fallback for arbitrary build prefixes. Could you also add tests for a non-placeholder long `$PREFIX` and a non-placeholder `$PREFIX` containing spaces?

```suggestion
    build_prefix = os.environ.get("PREFIX")
    build_prefix_markers = ("_h_env_placehold", "host_env_placehold")
    executable_has_build_prefix = any(
        path_part.startswith(build_prefix_markers)
        for path_part in os.path.abspath(executable).split(os.sep)
    )
    if os.environ.get("CONDA_BUILD") == "1" and executable_has_build_prefix:
        executable_in_prefix = True
        if build_prefix:
            executable_path = os.path.abspath(executable)
            build_prefix = os.path.abspath(build_prefix)
            try:
                executable_in_prefix = (
                    os.path.commonpath((executable_path, build_prefix)) == build_prefix
                )
            except ValueError:
                executable_in_prefix = False
        if executable_in_prefix:
            # The build tool will relocate this prefix at installation time.
            return shebang
```

bdice:
Thanks @jezdez! I added the two tests you requested and resolved a merge conflict. Please take another peek when you're able.

jezdez (approved):

PR #16305, opened 2026-06-29
Add `conda_error_hints` plugin hook

jezdez:
### Description

Refs #16290.

Adds a `conda_error_hints` plugin hook so plugins can contribute structured next-step hints for expected `CondaError` failures.

Plugin hooks yield public `CondaErrorHint` objects, which conda translates into the existing `GuidanceHint` model, merges after core guidance hints, deduplicates by `hint_code`, and renders through the existing terminal and JSON guidance paths.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #16329, opened 2026-07-01
Add `CondaError.append_hint`

kenodegard:
### Description

Add convenience method to append new hints to an existing exception instance.

Alternative solution to https://github.com/conda/conda/pull/16305

### Checklist - did you ...

### DIFF
--- a/conda/__init__.py
+++ b/conda/__init__.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from collections.abc import Iterable
     from subprocess import Popen
-    from typing import Any
+    from typing import Any, Self
 
     from ._private.exception_guidance import ErrorGuidance, ErrorGuidanceTypedDict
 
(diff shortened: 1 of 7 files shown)


jezdez:
Thanks for putting this together. I still think the cleaner direction is the dedicated `conda_error_hints` hook in [#16305](https://github.com/conda/conda/pull/16305), rather than making `conda_exception_observers` responsible for changing rendered error output.

Both PRs are still drafts, so I’m mostly looking at the API contract here. The observer hook is already useful for logging, telemetry, and demand tracking. If we also make it the supported way to mutate `CondaError` before rendering, that boundary gets fuzzy pretty quickly: plugin authors then have to know which exception mutations are part of the contract and which ones are just implementation details.

With [#16305](https://github.com/conda/conda/pull/16305), plugins yield `CondaErrorHint` objects from a hook whose job is specifically to add hints. Conda can then own the ordering, deduplication, terminal rendering, and JSON output format in one place.

I pushed a follow-up there to make those rules explicit: core hints win on duplicate `hint_code`, plugin hints are ordered deterministically, hook wrappers are skipped so plugin failures stay isolated, and docs/tests now cover that behavior.

So I’d continue with [#16305](https://github.com/conda/conda/pull/16305) and close this draft once we agree on that direction.

PR #16331, opened 2026-07-02
Speed up `PrefixGraph` topological sorting (#16331)

jezdez:
### Description

Part of #15969 as Track B follow-up B21. This remains independent of the name-index change merged in #15971 and only changes PrefixGraph._toposort_raise_on_cycles().

This replaces the repeated full-graph parent scans in PrefixGraph._toposort_raise_on_cycles() with a single reverse child-adjacency pass and queue. The implementation preserves conda's deterministic per-level package-name ordering, and when cycles are allowed it keeps _topo_sort_handle_cycles() seeing the same remaining graph shape before breaking a cycle.

I checked stdlib graphlib.TopologicalSorter as an alternative. It is much faster than the original repeated-scan path, but slower than the custom queue in the focused topo benchmark, so this PR keeps the custom implementation.

Focused synthetic topo timings on Windows, using a chain-like graph where each node depends on the previous three nodes:

| Records | Original repeated scan | graphlib | This PR |
|---:|---:|---:|---:|
| 1k | 0.0581 s | 0.0034 s | 0.0015 s |
| 2k | 0.2366 s | 0.0075 s | 0.0032 s |
| 5k | 1.5360 s | 0.0178 s | 0.0072 s |
| 10k | skipped | 0.0374 s | 0.0155 s |
| 50k | skipped | 0.2211 s | 0.0897 s |

The end-to-end Windows W3 signal was measured on the dedicated win-64 host in the branch combination where this bottleneck is visible (#15971 plus conda-libmamba-solver#921):

| Probe | #15971 + conda-libmamba-solver#921 | + this PR |
|---|---:|---:|
| W3 realistic, 5k records | 9.25 s | 8.018 s |
| W3 realistic, 10k records | 27.64 s | 13.736 s |
| W3 realistic, 50k records | 574.61 s | 64.459 s |

The preceding profile, prototype artifact, and branch confirmations are recorded in the [Track B report](https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md#windows-x86_64-checkpoint-2026-07-01).

### Checklist - did you ...

### DIFF
--- a/conda/models/prefix_graph.py
+++ b/conda/models/prefix_graph.py
@@ -228,23 +228,41 @@ class PrefixGraph:
         if not graph:
             return
 
-        while True:
-            no_parent_nodes = dict.fromkeys(
-                sorted(
-                    (node for node, parents in graph.items() if len(parents) == 0),
-                    key=lambda x: x.name,
-                )
-            )
-            if not no_parent_nodes:
-                break
+        node_positions = {node: position for position, node in enumerate(graph)}
+        parents_by_node = {node: set(parents) for node, parents in graph.items()}
+        children_by_node = {node: [] for node in graph}
+        for node, parents in parents_by_node.items():
+            for parent in parents:
+                if parent in children_by_node:
+                    children_by_node[parent].append(node)
 
-            for node in no_parent_nodes:
+        def sort_key(node):
+            return node.name, node_positions[node]
+
+        no_parent_nodes = sorted(
+            (node for node, parents in parents_by_node.items() if not parents),
+            key=sort_key,
+        )
+        while no_parent_nodes:
+            nodes = no_parent_nodes
+            no_parent_nodes = []
+
+            for node in nodes:
                 yield node
-                graph.pop(node, None)
+                graph.pop(node)
+                parents_by_node.pop(node)
 
-            for parents in graph.values():
-                for node in no_parent_nodes:
-                    parents.pop(node, None)
+                for child in children_by_node[node]:
+                    child_parents = parents_by_node.get(child)
+                    if child_parents is None:
+                        continue
+
+                    child_parents.discard(node)
+                    graph[child].pop(node, None)
+                    if not child_parents:
+                        no_parent_nodes.append(child)
+
+            no_parent_nodes.sort(key=sort_key)
 
         if len(graph) != 0:
             raise CyclicalDependencyError(tuple(graph))
@@ -255,7 +273,7 @@ class PrefixGraph:
         for k, v in graph.items():
             v.pop(k, None)
 
-        # disconnected nodes go first
+        # Disconnected nodes go first. Remove them before the main sort.
         nodes_that_are_parents = {
             node for parents in graph.values() for node in parents
         }
@@ -268,6 +286,8 @@ class PrefixGraph:
             ),
             key=lambda x: x.name,
         )
+        for node in disconnected_nodes:
+            graph.pop(node, None)
         yield from disconnected_nodes
 
         t = cls._toposort_raise_on_cycles(graph)
--- a/tests/models/test_prefix_graph.py
+++ b/tests/models/test_prefix_graph.py
@@ -14,7 +14,12 @@ from conda.exceptions import CyclicalDependencyError
 from conda.models.match_spec import MatchSpec
 from conda.models.prefix_graph import GeneralGraph, PrefixGraph
 from conda.models.records import PackageRecord
-from conda.testing.helpers import add_subdir_to_iter, get_solver_4, get_solver_5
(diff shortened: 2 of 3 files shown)


tests/models/test_prefix_graph.py
```diff
@@ -74,6 +74,16 @@ def get_sqlite_cyclical_record_set(tmpdir):
     return final_state, frozenset(specs)
 
 
+def package_record(name, depends=()):
```
jaimergp: I swear I've seen a helper like this somewhere else.

conda/models/prefix_graph.py
```diff
@@ -260,6 +282,8 @@ def _topo_sort_handle_cycles(cls, graph):
             ),
             key=lambda x: x.name,
         )
+        for node in disconnected_nodes:
+            graph.pop(node, None)
```
jaimergp, line 290: This is new. Is this also part of the speedup or a welcome but unrelated fix worth noting in the news item?

conda/models/prefix_graph.py
```diff
@@ -220,23 +220,45 @@ def _toposort_raise_on_cycles(cls, graph):
...
+            (node for node, parents in parents_by_node.items() if not parents),
+            key=sort_key,
+        )
+        while no_parent_nodes:
+            nodes = no_parent_nodes
+            no_parent_nodes = []
+
+            for node in nodes:
+                parents = parents_by_node.get(node)
+                if parents is None or parents:
```
jaimergp: So we only stay in the loop step if `node` is in `parents_by_node` but is not populated yet? IOW, present but empty?

tests/models/test_prefix_graph.py
```diff
@@ -798,6 +808,46 @@ def test_sort_without_prep(tmpdir, mocker, monkeypatch: MonkeyPatch):
...
+
+    assert tuple(record.name for record in graph.records) == (
+        "base",
+        "leaf",
+        "cycle-a",
+        "cycle-b",
+        "cycle-child",
+    )
+
+
```
jaimergp, line 864: A conversation with Claude (Opus 4.8) suggested to test demonstrating "wave ordering vs global ordering" is preserved. The current implementation is correct in that sense already, so this is only to prevent regressions in future refactors. I guess we are also fine with the amount of solver tests that depend on this implicitly, but I guess it doesn't hurt to have an explicit test for this. This is what Claude wrote; feel free to adapt it or drop it:

```suggestion

def test_prefix_graph_keeps_dependency_levels_separate_not_global_priority():
    # Guard against a future switch to a single global priority queue. A child
    # exposed partway through level 0 must still be emitted in the NEXT level --
    # after every remaining level-0 node -- even when its name sorts earlier.
    #
    # Both roots have a child, so neither is treated as a "disconnected" node and
    # hoisted ahead by _topo_sort_handle_cycles; both enter the real sort at
    # level 0.
    #
    #   Level 0: "m-root", "z-root"
    #   "a-child" depends only on "m-root", so it becomes ready as soon as
    #   "m-root" is emitted -- partway through level 0. Its name sorts before
    #   the still-pending "z-root".
    #
    #   Wave ordering    -> m-root, z-root, a-child, z-leaf   (correct / current)
    #   Global priority  -> m-root, a-child, z-root, z-leaf   ("a-child" < "z-root")
    records = (
        package_record("m-root"),
        package_record("z-root"),
        package_record("a-child", depends=("m-root",)),
        package_record("z-leaf", depends=("z-root",)),
    )

    graph = PrefixGraph(records)

    assert tuple(record.name for record in graph.records) == (
        "m-root",
        "z-root",
        "a-child",
        "z-leaf",
    )
    

```

tests/models/test_prefix_graph.py
jezdez: Good memory. I replaced the local helper with the existing `conda.testing.helpers.record` helper.

conda/models/prefix_graph.py
jezdez: These disconnected nodes have already been yielded before the main sort. Removing them from the working graph prevents them from being processed and yielded again. This is part of the faster sorting path rather than a separate public behavior change, so I kept the enhancement news item as-is.

conda/models/prefix_graph.py
jezdez: Yes, that was the intent. The queue only contains nodes that are still present and have no parents. I removed the defensive branch and now pop the queued node directly so that invariant is clear.

tests/models/test_prefix_graph.py
jezdez: Added a concise version of this wave-ordering regression test. Thanks for the concrete example.

PR #16334, opened 2026-07-02
Deprecate conda run resolving conda from caller environment

### LINKED ISSUE #16333: Make "conda run -n <env> conda ..." execute the target environment's conda
### Checklist

### What is the idea?

Today, `conda run .. conda` executes the `conda` entry point from the invoking installation, rather than the target environment.

For example:

```console
$ conda --version
conda 24.11.3

$ conda create -y -n tmp-bld conda conda-build anaconda-anon-usage

$ conda list -n tmp-bld | rg conda
conda                     26.5.3
conda-build               26.5.0
...

$ conda run -n tmp-bld conda --version
conda 24.11.3
```

In contrast, ordinary executables are resolved from the target environment:

```console
$ conda run -n tmp-bld python --version
Python 3.14.x
```

Likewise,

```console
$ conda run -n tmp-bld python -m conda --version
conda 26.5.3
```

uses the `conda` installed in the target environment.

This also affects **conda plugins**.

While

```console
conda run -n <env> <plugin-executable> ...
```

runs the executable from the target environment, invoking a plugin through

```console
conda run -n <env> conda <plugin> ...
```

uses the invoking installation's `conda`, meaning plugin discovery and plugin versions come from the invoking installation rather than the target environment.

### Why is this needed?

The command explicitly requests execution inside another environment, so it is natural to expect executable resolution to happen within that environment.

Today `conda` behaves differently from other executables:

- `conda run -n <env> python` → target environment
- `conda run -n <env> python -m conda ...` → target environment
- `conda run -n <env> conda ...` → invoking installation
- `conda run -n <env> conda <plugin> ...` → invoking installation's plugin discovery and plugin versions

Besides being surprising, this makes it difficult to test newer versions of `conda` and its plugins from an isolated environment using `conda run`, since the command continues to execute the invoking installation instead.

### What should happen?

If the target environment contains a `conda` installation, then

```console
conda run -n <env> conda ...
```

should execute that installation's `conda`.

Likewise,

```console
conda run -n <env> conda <plugin> ...
```

should use the target environment's plugin discovery and execute the plugin versions installed there.

This would make `conda` consistent with how `conda run` resolves other executables and would allow users to reliably exercise the `conda` installation and plugins contained within the target environment.

### Additional Context

Current workarounds are:

```console
conda run -n <env> python -m conda ...
```

or invoking the target environment's `conda` executable by absolute path.

I searched for an existing issue but didn't find one describing this specific behavior.

I'm intentionally filing this as a feature request rather than a bug report because this behavior appears to be longstanding, but from a user perspective it feels inconsistent with how `conda run` treats other executables.

jezdez:
### Description

Starts the deprecation cycle for #16333.

On POSIX, `conda run` currently uses the shell hook from the invoking installation. Its `conda()` function routes nested `conda` calls through `CONDA_EXE`, so they use the invoking installation instead of the target environment.

This behavior remains the default during the deprecation cycle. It becomes deprecated in 27.3 and is scheduled to change in 27.9. At that point, an unqualified nested `conda` command will resolve from the target environment's `PATH`, like other executables.

Users can select either installation explicitly during the transition:

- `conda run -n target "$CONDA_EXE" ...` uses the invoking installation.
- `conda run -n target python -m conda ...` uses the target installation.

The target-resolution path remains scoped to `conda run` and disabled until the behavior change.

### Checklist - did you ...

### DIFF
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -298,18 +298,34 @@ def wrap_subprocess_call(
             fh.write(f"conda activate {dev_arg} {quote_for_shell(prefix)}\n")
             if debug_wrapper_scripts:
                 fh.write(">&2 echo '*** environment after ***'\n>&2 env\n")
+            if not dev_mode:
+                # The POSIX hook defines `conda` as a shell function backed by
+                # the caller's CONDA_EXE. Keep shell-state commands working, but
+                # resolve normal `conda` calls from the activated PATH.
+                fh.write(
+                    r"""conda() {
+    \local cmd="${1-__missing__}"
+    case "$cmd" in
+        activate|deactivate)
+            __conda_activate "$@"
+            ;;
+        install|update|upgrade|remove|uninstall)
+            command conda "$@" || \return
+            __conda_activate reactivate
+            ;;
+        *)
+            command conda "$@"
+            ;;
+    esac
+}
+"""
+                )
             if multiline:
                 # The ' '.join() is pointless since mutliline is only True when there's 1 arg
                 # still, if that were to change this would prevent breakage.
                 fh.write("{}\n".format(" ".join(arguments)))
             else:
-                if not dev_mode and arguments and arguments[0] == "conda":
-                    # The POSIX hook defines `conda` as a shell function. `command`
-                    # skips functions and resolves the executable from the activated PATH.
-                    command_args = ("command", *arguments)
-                else:
-                    command_args = arguments
-                fh.write(f"{quote_for_shell(*command_args)}\n")
+                fh.write(f"{quote_for_shell(*arguments)}\n")
             # Capture the return code of the user's command in a variable
             # before deactivating. We don't need to unset this per se, because
             # the shell process will terminate and clean it up afterwards.
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -179,6 +179,8 @@ def wrap_subprocess_call(
     dev_mode,
     debug_wrapper_scripts,
     arguments: Sequence[str],
+    *,
+    use_target_conda_executable: bool = False,
 ):
     # Ensure arguments is a tuple of strings
     if not isiterable(arguments):
@@ -298,14 +300,19 @@ def wrap_subprocess_call(
             fh.write(f"conda activate {dev_arg} {quote_for_shell(prefix)}\n")
             if debug_wrapper_scripts:
                 fh.write(">&2 echo '*** environment after ***'\n>&2 env\n")
-            if not dev_mode:
+            if use_target_conda_executable and not dev_mode:
+                from .auxlib.ish import dals
+
                 # Keep the hook's `conda()` function, but make its command
                 # runner resolve `conda` from the activated PATH.
                 fh.write(
-                    r"""__conda_exe() (
-    command conda "$@"
-)
-"""
+                    dals(
+                        r"""
+                        __conda_exe() (
+                            command conda "$@"
+                        )
+                        """
+                    )
                 )
             if multiline:
                 # The ' '.join() is pointless since mutliline is only True when there's 1 arg
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -303,7 +303,13 @@ def wrap_subprocess_call(
                 # still, if that were to change this would prevent breakage.
                 fh.write("{}\n".format(" ".join(arguments)))
             else:
-                fh.write(f"{quote_for_shell(*arguments)}\n")
+                if not dev_mode and arguments and arguments[0] == "conda":
+                    # The POSIX hook defines `conda` as a shell function. `command`
+                    # skips functions and resolves the executable from the activated PATH.
+                    command_args = ("command", *arguments)
+                else:
+                    command_args = arguments
+                fh.write(f"{quote_for_shell(*command_args)}\n")
             # Capture the return code of the user's command in a variable
             # before deactivating. We don't need to unset this per se, because
             # the shell process will terminate and clean it up afterwards.
--- a/conda/utils.py
+++ b/conda/utils.py
@@ -304,7 +304,10 @@ def wrap_subprocess_call(
                 from .auxlib.ish import dals
 
                 # Keep the hook's `conda()` function, but make its command
-                # runner resolve `conda` from the activated PATH.
+                # runner resolve `conda` from the activated PATH. Match the
+                # subshell body used by `__conda_exe()` in `conda.sh`:
+                # https://github.com/conda/conda/blob/main/conda/shell/etc/profile.d/conda.sh
+                # `()` isolates shell-state changes, while `{}` does not.
                 fh.write(
                     dals(
                         r"""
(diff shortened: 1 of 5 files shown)


conda/utils.py
```diff
@@ -303,7 +303,13 @@ def wrap_subprocess_call(
                 # still, if that were to change this would prevent breakage.
                 fh.write("{}\n".format(" ".join(arguments)))
             else:
-                fh.write(f"{quote_for_shell(*arguments)}\n")
+                if not dev_mode and arguments and arguments[0] == "conda":
```
JeanChristopheMorinPerso: If I'm not mistaken, this solves `conda run -n myenv conda` but not `conda run -n myenv bash /myscript.sh` where `myscript.sh` calls conda? If so, doesn't this create some inconsistencies?

JeanChristopheMorinPerso: Here's another one: `conda run -n myenv time conda` wouldn't work because of `arguments[0] == "conda"`.

jezdez: Good catch. I pushed 3f02c98ca to handle the wrapper cases too.

The wrapper now keeps the hook-generated POSIX `conda()` function and only overrides its `__conda_exe` command runner after activation in non-dev mode:

```sh
__conda_exe() (
    command conda "$@"
)
```

So `activate`/`deactivate` still use the hook machinery, while normal `conda` calls resolve from the activated target `PATH`.

I added a parameterized regression for both direct `conda` and `time conda`. I also checked the `bash /myscript.sh` case manually with a script that calls `conda`, and it reaches the target-prefix executable.

conda/utils.py
```diff
@@ -298,6 +298,15 @@ def wrap_subprocess_call(
...
             if debug_wrapper_scripts:
                 fh.write(">&2 echo '*** environment after ***'\n>&2 env\n")
+            if not dev_mode:
+                # Keep the hook's `conda()` function, but make its command
+                # runner resolve `conda` from the activated PATH.
+                fh.write(
+                    r"""__conda_exe() (
+    command conda "$@"
+)
+"""
```
jaimergp: Style choice but I prefer `dedent()` (or that `dals()` helper) or, for short strings, just inline the newlines:

```suggestion
                fh.write(r'__conda_exe() {\n  command conda "$@"\n}\n')
```

Regardless, you need `{ }` for the function body!

jaimergp: Single line version:

```suggestion
                fh.write(r'__conda_exe() { command conda "$@"; }\n')
```

conda/utils.py
```diff
@@ -298,6 +300,20 @@ def wrap_subprocess_call(
...
+                from .auxlib.ish import dals
+
+                # Keep the hook's `conda()` function, but make its command
+                # runner resolve `conda` from the activated PATH.
+                fh.write(
+                    dals(
+                        r"""
+                        __conda_exe() (
+                            command conda "$@"
+                        )
```
jaimergp, line 316: Are you sure we don't need curly braces?

```suggestion
                        __conda_exe() {
                            command conda "$@"
                        }
```

conda/utils.py
```diff
@@ -298,6 +300,20 @@ def wrap_subprocess_call(
...
+                        """
+                    )
+                )
             if multiline:
                 # The ' '.join() is pointless since mutliline is only True when there's 1 arg
```
jaimergp:
```suggestion
                # The ' '.join() is pointless since multiline is only True when there's 1 arg
```

jaimergp:
What about users that had been expecting the shell function? How can they use it?

Without this PR, users can still do `conda run $CONDA_PREFIX/bin/conda` and move on. With this PR, to use `base` they have to do something like `BASE_PREFIX=$(conda info --root) conda run $BASE_PREFIX/bin/conda`.

Should we document these differences explicitly?

conda/utils.py
jezdez: Both forms are valid POSIX function bodies. `(...)` executes the body in a subshell, so shell-state changes remain isolated. `{ ... }` executes the body in the current shell. This override only runs `command conda "$@"`, so both forms produce the same result here. I kept `()` to preserve the shape and isolation of [`__conda_exe()` in `conda.sh`](https://github.com/conda/conda/blob/0d64c8165b028cd33ca3fba616475b93d4c74a5a/conda/shell/etc/profile.d/conda.sh#L3-L10). The explanation and source link are included in [0d64c8165](https://github.com/conda/conda/commit/0d64c8165b028cd33ca3fba616475b93d4c74a5a).

conda/utils.py
jezdez: Fixed in [0d64c8165](https://github.com/conda/conda/commit/0d64c8165b028cd33ca3fba616475b93d4c74a5a). Thanks!

jezdez:
You are right that changing the meaning immediately would be backward incompatible. I moved this PR to a deprecation cycle in [0d64c8165](https://github.com/conda/conda/commit/0d64c8165b028cd33ca3fba616475b93d4c74a5a).

The invoking installation remains the default during the transition. The behavior becomes deprecated in 27.3 and switches in 27.9, when an unqualified `conda` will resolve from the target environment `PATH`.

The command documentation now describes both explicit forms:

- `conda run -n target "$CONDA_EXE" ...` uses the invoking installation.
- `conda run -n target python -m conda ...` uses the target installation.

After the switch, `$CONDA_EXE` remains the explicit way to request the invoking installation.

docs/source/commands/run.rst
```diff
@@ -8,3 +8,26 @@
    :path: run
    :nodefault:
    :nodefaultconst:
+
+Nested ``conda`` commands
+=========================
+
+On POSIX systems, an unqualified ``conda`` command currently resolves through
+the invoking installation's shell function:
```
jaimergp, line 16:
```suggestion
On ``conda init``'d POSIX systems, an unqualified ``conda`` command
currently resolves through the invoking installation's shell function:
```

news/16333-conda-run-target-conda
```diff
@@ -0,0 +1,11 @@
+### Enhancements
+
+### Bug fixes
+
+### Deprecations
+
+* Mark `conda run --prefix <prefix> conda ...` resolving `conda` from the invoking installation on POSIX as pending deprecation. In 27.9, it will resolve from the target environment's `PATH`. Use `$CONDA_EXE` to select the invoking installation or `python -m conda` to select the target installation. (#16333)
```
jaimergp, line 7:
```suggestion
* Mark `conda run --prefix <prefix> conda ...` resolving `conda` from the invoking installation on POSIX as pending deprecation. In 27.9, it will resolve from the target environment's `PATH`. Use `${CONDA_EXE:-$(which conda)}` to select the invoking installation or `python -m conda` to select the target installation. (#16333)
```

conda/utils.py
jaimergp: Nice, thanks for the explanation.

docs/source/commands/run.rst
```diff
@@ -8,3 +8,26 @@
...
+
+This behavior is scheduled to change. In conda 27.9, ``conda`` will instead
+resolve from the target environment's ``PATH``, like other executables.
+
+Use an explicit form to avoid relying on behavior that will change:
+
+.. code-block:: console
+
+   # Use the invoking installation
+   $ conda run -n target "$CONDA_EXE" --version
```
jaimergp, line 30:
```suggestion
   $ conda run -n target "${CONDA_EXE:-$(which conda)}" --version
```

PR #16347, opened 2026-07-07
Prefer same-device package cache entries (B22)

jezdez:
### Description

Part of conda/conda#15969 as a Track B follow-up case (B22).

`PackageCacheData.get_entry_to_link()` can find the same package in multiple caches. Previously, it selected the first extracted entry, which could be on a different filesystem from the target prefix. That can cause copying even when an equivalent same-device entry could be hardlinked.

This PR prefers extracted entries on the target device. Selection matches available size and MD5 values, including legacy archive metadata, and stops at the first suitable entry. If none exists, it preserves the previous fallback selection. Memoizing the device comparison by cache root and target prefix avoids repeating those device checks.

This complements the later link/copy work by choosing a hardlink-eligible source when the cache topology allows it.

### Focused measurement

Comparison of `main` at [`7d3e813`](https://github.com/conda/conda/commit/7d3e81335f3016229f8f5dc7e82f69217d97f6d3) with B22 at [`6137fcd`](https://github.com/conda/conda/commit/6137fcd5c5dde3015bdd4d6b5f65cbf0dcb5a3c9) on macOS 26.6.2 arm64, APFS, and Python 3.13.14, using medians of five repeats with alternating execution order.

The setup used two equivalent synthetic extracted package-cache entries: the first on a mounted APFS disk image, the second on the target-prefix device. The timing covers cache selection, link-type detection, and file creation in an existing empty target directory. Source data was cached by the OS, with no explicit disk flush. Hardlinks were enabled and softlinks disabled.

| Case | `main` | B22 | Selected cache | Speedup |
| --- | ---: | ---: | --- | ---: |
| 512 files, 64 KiB each | 165.76 ms | 97.41 ms | cross-device copy fallback -> same-device hardlinks | 1.70x |

This demonstrates the benefit of selecting an equivalent same-device cache entry for this workload. It does not measure a full installation or imply a universal install speedup.

### Checklist - did you ...

- [x] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))? Not applicable.


conda/core/package_cache_data.py
```diff
@@ -67,6 +67,18 @@
...
+
+@cache
+def _paths_on_same_device(left, right):
+    # Device ids are stable for the package-cache and prefix directories during
+    # a transaction. Caching avoids repeated stat calls when several matching
+    # cache entries are considered for many package records.
+    try:
+        return os.stat(left).st_dev == os.stat(right).st_dev
+    except OSError:
+        return False
```
jaimergp: This may belong better in `gateways.disk`?

conda/core/package_cache_data.py
```diff
@@ -258,10 +270,22 @@ def get_all_extracted_entries(cls):
...
+        if target_prefix is not None:
+            # Same-device cache entries preserve hardlink and APFS clone fast paths.
+            pc_entry = first(
+                extracted_entries,
+                key=lambda pcrec: _paths_on_same_device(
+                    pcrec.extracted_package_dir, target_prefix
+                ),
+            )
+            if pc_entry is not None:
+                return pc_entry
```
jaimergp: This is begging for walrus!

conda/core/package_cache_data.py
```diff
@@ -258,10 +270,22 @@ def get_all_extracted_entries(cls):
...
+
+        pc_entry = next(iter(extracted_entries), None)
         if pc_entry is not None:
             return pc_entry
```
jaimergp: And this one too.

conda/core/package_cache_data.py
jezdez: Agreed. I moved it to `conda.gateways.disk.test` as the shared `paths_on_same_device()` predicate, next to the other disk capability checks. Calls are now keyed by package-cache root and target prefix so memoization is reused across packages.

conda/core/package_cache_data.py
jezdez: Done. The same-device lookup now uses a walrus assignment.

conda/core/package_cache_data.py
jezdez: Done for the fallback lookup too.

travishathaway (approved):
This looks good. My suggestion would be annotations for the arguments to `paths_on_same_device` (maybe [os.PathLike](https://docs.python.org/3/library/os.html#os.PathLike), but I'm not really sure that's important here.

PR #16350, opened 2026-07-07
Run pyc compiler from private helper script (B23)

jezdez:
### Description

Part of #15969 as Track B follow-up B23.

Conda currently invokes the target environment's Python as a subprocess and runs the standard-library compileall module with a temporary path list. This PR moves the compiler program into a stdlib-only private helper file and invokes that helper with the target Python.

The design was informed by [uv's bytecode compiler](https://github.com/astral-sh/uv/blob/main/crates/uv-installer/src/compile.rs), which uses multiple persistent target-Python helper processes, streams files over stdin and stdout, and applies per-file timeouts. This PR establishes only the helper-file boundary. By itself it does not add process isolation beyond conda's existing target-Python subprocess, and it does not add parallel workers.

### Measured impact

Focused compile-phase measurements on macOS:

| Case | main | B23 | Result |
| --- | ---: | ---: | ---: |
| 500 Python files | 951.56 ms | 808.42 ms | 1.18x |
| 2,000 Python files | 1.03 s | 1.22 s | 0.84x |

The result is workload-dependent and regresses the larger fixture. B24 is now independent and passes its process count directly to the target interpreter's compileall -j implementation.

B23 should remain draft unless conda intends to implement the rest of the uv-style persistent-worker protocol. As a standalone helper extraction it does not establish a speed or correctness win.

### Checklist - did you ...

### DIFF
--- a/conda/_private/pyc_compiler.py
+++ b/conda/_private/pyc_compiler.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""Compile explicit py/pyc path pairs for install transactions.
+
+This script is executed by the target environment's Python, which may not have
+conda installed. Keep it stdlib-only and do not import from conda.
+"""
+
+import errno
+import json  # noqa: TID251 - the target Python may not be able to import conda
+import os
+import py_compile
+import sys
+import traceback
+
+
+def main():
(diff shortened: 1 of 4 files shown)


jezdez:
B24 now uses the target Python's compileall -j directly, which gives us process-level parallelism without maintaining a separate worker protocol. This helper-only step no longer has a standalone benefit, so closing.

PR #16352, opened 2026-07-07
Allow skipping install-time Python bytecode compilation (B24)

jezdez:
### Description

Part of #15969 as Track B follow-up B24.

Python bytecode compilation is an install-time latency versus first-import tradeoff. Conda currently compiles bytecode for noarch Python packages during installation, while uv lets users choose whether to pay that cost during installation.

This PR preserves conda's current default and adds a way to skip that work:

- `--no-compile-pyc` disables bytecode compilation for `create`, `install`, and `update` commands.
- `compile_pyc: false` disables it through configuration.

When compilation remains enabled, conda continues to invoke the target environment's Python as a subprocess. The `compileall` worker count follows `default_threads` when configured. Otherwise, conda passes `-j 0` and lets Python choose based on the available CPUs.

This implementation is independent of B23. The target interpreter owns the `compileall` worker processes. No generated command-line Python or custom compilation helper is added.

### Checklist - did you ...

- [x] Add / update outdated documentation? CLI help and the configuration description cover the new control.

### DIFF
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -461,6 +461,7 @@ class Context(Configuration):
         aliases=("error_upload_url",),
     )
     force = ParameterLoader(PrimitiveParameter(False))
+    compile_pyc = ParameterLoader(PrimitiveParameter(True))
     json = ParameterLoader(PrimitiveParameter(False))
     _console = ParameterLoader(
         PrimitiveParameter(DEFAULT_CONSOLE_REPORTER_BACKEND, element_type=str),
@@ -1357,6 +1358,7 @@ class Context(Configuration):
             "allow_softlinks",
             "always_copy",
             "always_softlink",
+            "compile_pyc",
             "path_conflict",
             "rollback_enabled",
             "safety_checks",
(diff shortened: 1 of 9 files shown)


conda/base/context.py
```diff
@@ -1698,6 +1708,19 @@ def description_map(self) -> dict[str, str]:
...
+                """
+                Compile Python bytecode files for noarch Python packages during installation.
+                Disable this when startup-time bytecode compilation is preferable to
+                install-time compilation.
+                """
+            ),
+            pyc_compile_threads=dals(
+                """
+                Processes to use when compiling Python bytecode files during installation.
+                When set to 0, Python chooses a process count based on the available CPUs.
```
jaimergp: This is not what happens; it goes first to `default_threads`.

conda/base/context.py
```diff
@@ -715,6 +719,10 @@ def execute_threads(self) -> int | None:
             threads = 1
         return threads
 
+    @property
+    def pyc_compile_threads(self) -> int | None:
+        return self._pyc_compile_threads or self.default_threads or 0
```
jaimergp: Why do we need this separate from other thread values? Is it worth the granularity?

conda/base/context.py
jezdez: Good catch. I removed the separate `pyc_compile_threads` setting, so this description is gone. `default_threads` now documents that it also controls bytecode compilation, and an unset value still maps to `compileall -j 0`.

conda/base/context.py
jezdez: Agreed, the separate setting was more granular than the evidence justifies. I removed it. `compileall -j` now uses `default_threads` when set and `0` otherwise.

PR #16353, opened 2026-07-07
Retry built-in Windows package extraction (B25)

jezdez:
### Description

Part of conda/conda#15969 as Track B follow-up B25.

Built-in package extraction can fail transiently on Windows when antivirus, indexing, or filesystem watchers briefly hold package-cache files open. A retry can recover once the handle is released instead of failing the installation.

This PR passes built-in `.conda` and `.tar.bz2` extraction through conda's existing `exp_backoff_fn`. The retry lives in the shared extraction helper used by both the process pool and the plugin fallback.

The existing default is used without a custom attempt count. Windows retries `EACCES` and `EPERM` failures up to seven attempts, for a maximum sleep window of roughly 6.3 to 6.9 seconds. Other platforms call the extractor once without sleeping.

### Measured impact

This is a reliability-focused Track B follow-up. A focused Python 3.10.20 simulation of the Windows retry branch, using the same positional and keyword arguments as package extraction, produced:

| Case | `main` | B25 |
| --- | ---: | ---: |
| No-failure call, per call | 0.05 us | 0.26 us |
| 4 transient `EACCES` failures before success | fails after 1 attempt | succeeds on attempt 5 |
| 6 transient `EACCES` failures before success | fails after 1 attempt | succeeds on attempt 7 |

The wrapper adds about 0.21 us to a successful extraction call. A seventh consecutive failure still stops after attempt 7 instead of extending the wait further.

A separate macOS spawn-process-pool benchmark used 12 tiny `.conda` packages, 3 workers, and 50 randomized runs to include worker startup and the new runtime import:

| Case | `main` median | B25 median |
| --- | ---: | ---: |
| Process-pool startup and 12 extractions | 145.6 ms | 145.0 ms |

This showed no measurable process-pool regression.

### Checklist - did you ...

### DIFF
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -6,19 +6,21 @@ from __future__ import annotations
 
 import os
 
-# This module is imported in each spawned extraction worker. Keep imports here
-# limited to the standard library; importing conda runtime state defeats the
-# process pool's startup benefit.
+# This module is imported in each spawned extraction worker. Keep top-level
+# imports limited to the standard library so loading the worker stays cheap.
 
 
 def extract_conda_package_archive(
     source_full_path: str | os.PathLike,
     destination_directory: str | os.PathLike,
 ) -> None:
-    """Extract a conda package archive without importing conda runtime state."""
+    """Extract a conda package archive with standard file-operation retries."""
     import conda_package_handling.api
 
-    conda_package_handling.api.extract(
+    from ..gateways.disk import exp_backoff_fn
+
+    exp_backoff_fn(
+        conda_package_handling.api.extract,
         os.fspath(source_full_path),
         dest_dir=os.fspath(destination_directory),
     )
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -7,9 +7,8 @@ from __future__ import annotations
 import os
 import pickle
 
-# This module is imported in each spawned extraction worker. Keep imports here
-# limited to the standard library; importing conda runtime state defeats the
-# process pool's startup benefit.
+# This module is imported in each spawned extraction worker. Keep top-level
+# imports limited to the standard library so loading the worker stays cheap.
 
 
 def extract_conda_package_archive(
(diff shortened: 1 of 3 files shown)


conda/core/package_cache_data.py
```diff
@@ -1000,7 +1002,12 @@ def do_extract_action(prec, extract_action, progress_bar):
     extract_action.verify()
     # currently unable to do updates on extract;
     # likely too fast to bother
-    extract_action.execute(None)
+    if on_win:
```
jaimergp: `exp_backoff_fn` already returns early on non Windows:

https://github.com/conda/conda/blob/4f8feabe660fa6967f28bb2b56a4ab4421defd18/conda/gateways/disk/__init__.py#L19-L23

conda/core/package_cache_data.py
```diff
@@ -1000,7 +1002,12 @@ def do_extract_action(prec, extract_action, progress_bar):
...
+        # Antivirus and search indexers can briefly hold handles while the
+        # extractor writes the package cache. Give Windows a longer retry window.
+        exp_backoff_fn(extract_action.execute, None, max_tries=9)
```
jaimergp: 9 attempts means like almost 30s of potential waiting. Why that number in particular? Should we pick a smaller one?

conda/core/package_cache_data.py
jezdez: You are right. I removed the `on_win` branch and now call `exp_backoff_fn` directly. After rebasing over B8, the retry now lives in the shared built-in extraction helper so it also covers the process-pool path.

conda/core/package_cache_data.py
jezdez: Agreed. I removed `max_tries=9` and use the existing default of seven attempts. That caps the sleep window at about 6.3 to 6.9 seconds. The refreshed simulation succeeds after four or six transient failures and stops on a seventh.

PR #16355, opened 2026-07-07
Add conda plugins list subcommand

jezdez:
### Description

Implements the first slice of conda/conda#16354 after moving the plugin-management epic from `conda-self` into conda. This supersedes conda/conda-self#130.

This adds a bundled `conda plugins` subcommand with an initial `conda plugins list` action. The list command reports installed entry-point plugins with their distribution name, version, active/disabled status, canonical plugin name, and registered hook groups, with `--json` support for machine-readable output.

The plugin discovery logic lives on `CondaPluginManager`, reusing conda's loaded plugin state, pluggy hook callers, and entry-point distribution metadata instead of rescanning entry points from the subcommand implementation.

Part of #16354.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/manager.py
+++ b/conda/plugins/manager.py
@@ -59,7 +59,7 @@ from .types import CondaErrorHint, CondaExceptionEvent
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
     from types import TracebackType
-    from typing import Any, Literal, TypeVar, cast
+    from typing import Any, Literal, TypedDict, TypeVar, cast
 
     from pluggy import HookImpl
     from requests.auth import AuthBase
@@ -98,6 +98,14 @@ if TYPE_CHECKING:
     )
 
     P = TypeVar("P", bound=CondaPluginWithAliases)
+
+    class PluginInfo(TypedDict):
+        name: str
+        version: str
+        canonical_name: str
+        status: str
+        hooks: list[str]
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


conda/plugins/subcommands/__init__.py
```diff
@@ -1,5 +1,7 @@
 # Copyright (C) 2012 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+from importlib import import_module
+
 from . import doctor
 
-plugins = [doctor]
+plugins = [doctor, import_module(f"{__name__}.plugins")]
```
soapy1: What's the motivation for having these two different approaches of importing the subcommand plugin module?

conda/plugins/manager.py
```diff
@@ -201,6 +209,35 @@ def get_plugin_source(self, conda_plugin: CondaPlugin) -> str | None:
...
+        for plugin, dist in self.list_plugin_distinfo():
+            canonical_name = self.get_name(plugin)
+            if canonical_name is None or canonical_name in installed:
+                continue
+
+            installed[canonical_name] = {
+                "name": dist.project_name,
+                "version": dist.version,
+                "canonical_name": canonical_name,
+                "status": "disabled" if self.is_blocked(canonical_name) else "active",
```
soapy1, line 241: What should the output of `status` be if `CONDA_NO_PLUGINS=true` is set? I expect conda to list all the plugins installed, but have their status be disabled. Instead I get:
```
$ CONDA_NO_PLUGINS=true conda plugins list
No plugins installed.
```

What should the behaviour be here? And one more follow up question, are users meant to be able to disable single plugins? If so, how?

soapy1 (commented):
rad! Just a few questions.

conda/plugins/subcommands/__init__.py
kenodegard: I think it's because of the name collision, normally we do:

```python
from . import doctor as _doctor
from . import plugins as _plugins

plugins = [_doctor, _plugins]
```

jezdez: Switched this to the normal aliased imports:

```python
from . import doctor as _doctor
from . import plugins as _plugins

plugins = [_doctor, _plugins]
```

conda/plugins/manager.py
jezdez: Agreed: `CONDA_NO_PLUGINS=true conda plugins list` should list installed entry-point plugins as disabled. Pluggy unregisters blocked plugins, which meant we lost the registration name and hook callers. The manager now recovers the canonical name and inspects the hook markers directly, and there is a regression test for this exact command.

Per-plugin disabling is covered by #15915 via `--no-plugins=<name>[,...]`.

soapy1 (approved):

jezdez:
Adding this to the September release since July is almost over and I think we should ship this together with the other PRs assigned to #16354

PR #16356, opened 2026-07-07
Add conda plugins info subcommand

jezdez:
### Description

Adds `conda plugins info <name>` as the second implementation for #16354, based on merged #16355.

This extends `CondaPluginManager` with an installed-plugin detail lookup that parses the distribution metadata once with `packaging.metadata.Metadata`, then exposes its status and hook implementations. The detail output preserves legacy homepage metadata and includes every labeled `Project-URL` entry in text and JSON output. This also raises the minimum supported `packaging` version to 26.3.

Part of #16354.
Depends on #16355.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for inclusion in the next release notes?


conda/plugins/manager.py
```diff
@@ -18,11 +18,13 @@
 from collections.abc import Iterable, Mapping
 from contextlib import suppress
 from dataclasses import dataclass
+from email.message import Message
```
soapy1: Hmmm, what's the situation with this import? I see that `Message` is included in `_get_plugin_distribution_metadata(dist: DistFacade) -> Message | None:`. If I set a breakpoint inside that function, like:
```
def _get_plugin_distribution_metadata(dist: DistFacade) -> Message | None:
        raw_dist = getattr(dist, "_dist", None)
        metadata = getattr(raw_dist, "metadata", None)
        import pdb; pdb.set_trace()
        if isinstance(metadata, Message):
            return metadata
```

I can inspect the type of `metadata` I find:
```
(Pdb) type(metadata)
<class 'importlib.metadata._adapters.Message'>
```

conda/plugins/manager.py
```diff
@@ -244,6 +251,75 @@ def get_installed_plugins(self) -> list[PluginInfo]:
...
+        if metadata is None:
+            return ""
+
+        homepage = metadata.get("Home-page", "") or ""
+        if homepage:
+            return homepage
+
+        for project_url in metadata.get_all("Project-URL") or ():
+            label, separator, url = project_url.partition(",")
+            if separator and label.strip().lower() in {"home-page", "homepage"}:
```
soapy1: hmm, I'm not sure that looking at labels is a scalable way to go here. It looks like package authors can name their [project-url's anything](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-project-url). For example, `conda-index` [uses `Home`](https://github.com/conda/conda-index/blob/main/pyproject.toml#L58), so if you do `conda plugins info conda_index` the homepage field is left empty.

An alternative approach might be to list all the project-urls in addition to the home-page?

conda/plugins/manager.py
jezdez: Heh, imagine [my](https://github.com/python/cpython/blob/66e313f471516bfa04c5c8966c159fafe6be082e/Lib/importlib/metadata/_adapters.py#L20) [surprise](https://github.com/python/importlib_metadata/blob/9757b400ee412ddb1d685f139ed3300f607c059a/importlib_metadata/_adapters.py#L20)!

conda/plugins/manager.py
jezdez: Yeah, why not, good idea.

conda/plugins/manager.py
jezdez: I'll change it to the concrete class.

jezdez: I guess if it walks like a duck, and quacks like a duck, it's obviously an email message :D

conda/plugins/manager.py
soapy1: I took a shot at implementing this change in https://github.com/conda/conda/compare/fix/plugins-info-subcommand...soapy1:conda:show-project-urls?expand=1.

jezdez: Thanks for the branch. I adapted it in c6b9be8ba. `get_installed_plugin_info()` now keeps the legacy `homepage` value and adds ordered `project_urls` records containing both the label and URL. Text output renders each labeled URL separately, and the tests cover `Home`, `Documentation`, and legacy `Home-page`. I credited you as co-author.

conda/plugins/manager.py
jezdez: This now uses `packaging.metadata.Metadata` to parse the distribution metadata in 342a5e9c0. The implementation no longer imports or checks `email.message.Message`, so it does not depend on the private `importlib.metadata` adapter type.

jezdez:
@soapy1 I updated this to parse the distribution metadata once with `packaging.metadata.Metadata`, removed the `email.message.Message` handling and the one-call private helper, preserved the legacy homepage fallback, and exposed all labeled `Project-URL` entries. The minimum `packaging` version is now 26.3, and the dependent #16357 has been restacked.

soapy1 (approved):
Looks great!

conda/plugins/manager.py
soapy1: neato! the labeled url's are nice, good idea

PR #16357, opened 2026-07-07
Add conda plugins package management subcommands

jezdez:
### Description

Adds the package-management implementation for #16354.

This adds `conda plugins install`, `conda plugins remove`, and `conda plugins update` under the bundled `conda plugins` namespace. The subcommands reuse the existing conda install, remove, and update executors instead of implementing separate transaction logic.

Plugin installation accepts repeated `--channel` options and `--override-channels` through the native parser. These sources apply only to the package transaction, so a plugin can be installed before a `.condarc` exists without saving a bootstrap channel or changing existing channel configuration.

The change also adds `CondaPluginManager.is_conda_plugin_package()` as the shared predicate for package-content validation. It inspects extracted package metadata or installed files recorded in the selected prefix for Python distribution `entry_points.txt` metadata and requires the `conda` entry point group. Installed file paths account for relocated `noarch: python` packages. After solving, `conda plugins install` fetches and validates only directly requested package records before showing the transaction plan or asking for confirmation. Dependencies are fetched only after validation and confirmation. Dry runs do not fetch packages.

Remove, named update, and `update --all` inspect conda package records in the selected target prefix. Named updates validate after CLI and environment-file inputs are merged. File-based updates reject external packages and environment-variable changes because the command only manages conda plugin packages. `conda plugins update --all` expands only the plugin package names installed in the selected prefix, so it does not use the environment-wide `conda update --all` behavior.

The plugin install and update parsers omit `--only-deps`, because these commands manage plugin packages themselves.

The release fragments are under `releases/news/`, with a manual package-management scenario under `releases/qa/`.

Part of #16354.

### Checklist - did you ...

- [x] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?


travishathaway (commented):
@jezdez,

I had a chance to poke around at this a little and just wanted to point out one weird UX issue I ran into.

1. `conda plugins install <not-conda-plugin-pacakge>`
2. Runs a full solve, ask for a confirmation (y/n) and downloads all packages 
3. During the extract phase it errors

I understand this is probably because we need to peak inside of the conda package before verifying it's a conda plugin, but this is pretty undesirable from a UX point of view. If we go forward with this, I think we need to begin immediately planning how to address it.

jezdez:
@travishathaway Good catch. I reproduced this and pushed 411e7ffbb. After solving, `conda plugins install` now fetches and validates only the directly requested package before printing the package plan or asking for confirmation. An invalid package no longer fetches its dependencies, and nothing is linked into the target prefix. The requested archive still has to be downloaded and extracted because repodata does not include Python entry-point groups. A true pre-download check would require a new package and repodata metadata field.

conda/plugins/subcommands/plugins/package_validation.py
```diff
@@ -0,0 +1,108 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from ....core.link import UnlinkLinkTransaction
+    from ....models.records import PackageCacheRecord
+
+
+def require_installed_plugin_specs(specs: Iterable[str], command: str) -> None:
+    installed_names = frozenset(
+        plugin["name"] for plugin in context.plugin_manager.get_installed_plugins()
```
danyeaw: This gets plugins from conda's runtime environment. However, `conda plugin` takes -n and -p arguments. Should we use the target_prefix instead?

releases/news/16354-plugins-package-subcommands
```diff
@@ -0,0 +1,19 @@
+### Enhancements
```
danyeaw, line 1: We need to rebase this to the new releases directory and add a qa snippet.

conda/plugins/subcommands/plugins/package_validation.py
jezdez: Good catch. Remove, named update, and `update --all` now inspect conda package records in the selected target prefix. Named updates validate after CLI and environment-file inputs are merged. The prefix scan also preserves the standard missing-environment error, and file-based updates reject external packages and environment variables.

releases/news/16354-plugins-package-subcommands
jezdez: Rebased onto current `main`, moved both plugin fragments from `news/` to `releases/news/`, and added a `releases/qa/` scenario for rejection, install, named update, `update --all`, and removal.

releases/qa/16354-plugins-package-subcommands
```diff
@@ -0,0 +1,96 @@
...
+
+   ```console
+   conda plugins install --yes --name qa-plugins-16354 --override-channels --channel conda-forge conda-tree=1.2.0
+   ```
+
+5. Confirm that the plugin is installed, its subcommand loads, and installation can be repeated.
+
+   ```console
+   conda list --name qa-plugins-16354 conda-tree
+   conda run --name qa-plugins-16354 conda tree --help
```
soapy1, line 48: This command does not work for me if I follow the directions listed above. 

```
$ conda run -p /tmp/two conda tree --help 
usage: conda [-h] [-v] [--no-plugins] [-V] COMMAND ...
conda: error: argument COMMAND: invalid choice: 'tree' (choose from 'activate', 'check', 'clean', 'commands', 'compare', 'config', 'create', 'deactivate', 'doctor', 'env', 'export', 'index', 'info', 'init', 'install', 'list', 'menuinst', 'notices', 'package', 'plugins', 'pypi', 'remove', 'rename', 'repoquery', 'run', 'search', 'self', 'uninstall', 'update', 'upgrade')
ERROR conda.cli.main_run:execute(148): `conda run conda tree --help` failed. (See above for error)
```

Trying some stuff out, it looks like if I install conda-tree into my base environment this works.  Further, if I activate the environment I intsalled conda-tree into, I can see that the plugin is installed. But it's not really available

```
$  conda activate /tmp/two

(/tmp/two) $ conda list | grep conda-tree    
conda-tree                 1.2.0            py313h06a4308_0

 (/tmp/two)  $ conda plugins list                   
Name                   Version  Status  Hooks
---------------------------------------------
conda_index            0.13.0  active  subcommands
conda-libmamba-solver  26.7.0  active  solvers, subcommands
conda-lockfiles        0.2.1   active  environment_exporters, environment_specifiers
conda-pypi             0.12.0  active  health_checks, package_extractors, post_transaction_actions, settings, subcommands
conda-rattler-solver   0.1.1   active  solvers
conda-self             0.2.1   active  health_checks, settings, subcommands
menuinst               2.5.2   active  subcommands

 (/tmp/two)  $ conda plugins info conda-tree  

CondaValueError: No installed conda plugin found matching 'conda-tree'.
```

Also, I found that `conda plugins list` does not recognize plugins that are installed in my new environment. For example, I have conda-pypi 0.12.0 installed in my base env and conda-pypi 0.11.0 in my new env:
```
(/tmp/two) $ conda list | grep conda-pypi    
conda-pypi                 0.11.0           py313h06a4308_0

(/tmp/two)  $ conda plugins list
Name                   Version  Status  Hooks
---------------------------------------------
. . .
conda-pypi             0.12.0  active  health_checks, package_extractors, post_transaction_actions, settings, subcommands
. . .

(/tmp/two) $ conda plugins info conda-pypi    
Name          : conda-pypi
Version       : 0.12.0
Status        : active
Canonical name: conda_pypi.plugin
Hooks         : health_checks, package_extractors, post_transaction_actions, settings, subcommands
Summary       : Better PyPI interoperability for the conda ecosystem.
License       : 
Homepage      : https://github.com/conda/conda-pypi
Project URLs  :
  - homepage: https://github.com/conda/conda-pypi
```

I think it's reasonable for different environments to have their own plugins. But it looks like only base environment plugins are executed.

PR #16358, opened 2026-07-07
Use APFS clonefile for copy-mode installs (B26)

jezdez:
### Description

Part of conda/conda#15969 as a new Track B follow-up case (B26).

APFS supports copy-on-write file cloning via `clonefile`. This PR applies that primitive to conda's existing copy-mode file creation path, allowing copy-style installs on compatible APFS filesystems to avoid eager byte copies.

The implementation keeps conda's existing package-cache hardlink behavior as the source of truth for standard cache-backed installs, preserving the refcount behavior used by `conda clean --packages`.

### Measured impact

The expected win is for copy-mode installs on APFS, where file creation can become metadata-heavy instead of byte-copy-heavy.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Disk utility functions for creating new files or directories."""
 
+import errno
 import os
 import sys
 import tempfile
 import warnings as _warnings
-from errno import EACCES, EPERM, EROFS
+from ctypes import CDLL, c_char_p, c_int, get_errno
+from errno import EACCES, ENOSYS, EPERM, EROFS, EXDEV
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, splitext
 from shutil import copyfileobj, copystat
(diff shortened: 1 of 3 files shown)


conda/core/link.py
```diff
@@ -96,8 +97,15 @@ def determine_link_type(extracted_package_dir, target_prefix):
...
         return LinkType.softlink
+    # APFS can clone whole eligible directories, which is faster than issuing
+    # one hardlink syscall per file. Keep conda's cheap hardlink path first on
+    # non-macOS platforms unless hardlinking is unavailable.
+    if on_mac and clone_file_supported(source_test_file, target_prefix):
+        return LinkType.clone
     if hardlink_supported(source_test_file, target_prefix):
         return LinkType.hardlink
+    if clone_file_supported(source_test_file, target_prefix):
+        return LinkType.clone
```
wolfv: This would change the conda spec ... for conda-meta records, right?

conda/gateways/disk/create.py
```diff
@@ -389,6 +502,9 @@ def create_link(src, dst, link_type=LinkType.hardlink, force=False):
             )
 
             copy(src, dst)
+            return
+    elif link_type == LinkType.clone:
+        clone_link_or_copy(src, dst)
```
wolfv: You may be missing out on the biggest optimization on mac which is to clone the whole directory if possible :)

conda/core/link.py
jezdez: Good catch, I narrowed this PR so it no longer adds LinkType.clone or changes conda-meta link semantics.

conda/gateways/disk/create.py
jezdez: Yep, agreed, I split that out. This PR is now just the safe copy-mode clonefile primitive.

jezdez:
Closing this one to reopen it from a conda/conda branch, so the Track B stack stays in the main repo.

PR #16367, opened 2026-07-07
Use FICLONE for Linux copy-mode installs (B28)

jezdez:
### Description

Part of conda/conda#15969 as a new Track B follow-up case (B28).

Linux filesystems such as btrfs and XFS can support copy-on-write file cloning through the `FICLONE` ioctl. This PR applies that native primitive to conda's existing copy-mode file creation path for files at or above 64 KiB, allowing compatible Linux installs to avoid eager byte copies when the package cache and target prefix live on a reflink-capable filesystem.

This PR is stacked on B27 (#16369) until that PR lands. The new Track B case here is the Linux reflink copy backend.

### Measured impact

Focused copy-mode measurements on Ubuntu 24.04 arm64 with btrfs on `/dev/sdb1` and Python 3.14.6 from conda-forge, using medians of 20 randomized repeats. The benchmark compares conda's buffered copy fallback, Python 3.14's `shutil.copyfile()`, and this PR's gated `FICLONE` path. Each path also applies `copystat`.

| Case | Conda buffered copy | Python 3.14 `shutil.copyfile()` | B28 gated `FICLONE` | B28 vs conda | B28 vs `shutil` |
| --- | ---: | ---: | ---: | ---: | ---: |
| 1 file, 64 MiB | 14.47 ms | 8.62 ms | 0.060 ms | 242x | 144x |
| 512 files, 64 KiB each | 22.41 ms | 23.55 ms | 14.58 ms | 1.54x | 1.61x |
| 2,048 files, 1 KiB each | 53.34 ms | 60.68 ms | 55.71 ms | 0.96x | 1.09x |

The conda-forge Python 3.14.6 build does not expose `os.copy_file_range`, so `shutil.copyfile()` uses `sendfile` instead. After syncing writes, btrfs extent accounting showed the 64 MiB source and `FICLONE` target as fully shared. The conda buffered-copy and `shutil.copyfile()` targets were fully exclusive. `filefrag` also reported the same physical extent and the `shared` flag only for the source and `FICLONE` target.

The tiny-file case does not take the raw ioctl path. Its 0.96x result captures the clone-eligibility check before falling back to normal copy.

A separate 512-file size sweep on the same btrfs mount shows why the PR uses a 64 KiB gate. Raw `FICLONE` loses at 1-16 KiB, starts winning at 32 KiB, and is clearly ahead from 64 KiB upward.

| Bytes per file | Raw `FICLONE` speedup |
| ---: | ---: |
| 1 KiB | 0.85x |
| 4 KiB | 0.87x |
| 8 KiB | 0.91x |
| 16 KiB | 0.97x |
| 32 KiB | 1.05x |
| 64 KiB | 1.25x |
| 128 KiB | 1.59x |
| 256 KiB | 2.17x |
| 1 MiB | 6.50x |

This is a large and medium-file copy-mode win on reflink-capable filesystems, not a universal small-file win. Unsupported filesystems are cached by source and target device pair so repeated copy attempts use the normal fallback path without retrying the ioctl for every eligible file.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -12,6 +12,7 @@ from errno import EACCES, EINVAL, ENOSYS, EPERM, EROFS, EXDEV
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, splitext
 from shutil import copyfileobj, copystat
+from stat import S_ISREG
 
 from ... import CondaError
 from ...auxlib.ish import dals
@@ -119,6 +120,7 @@ _FICLONE_UNSUPPORTED_ERRNOS = frozenset(
     if error is not None
 )
 _FICLONE = 0x40049409
+_FICLONE_MIN_FILE_SIZE = 64 * 1024
 _FICLONE_UNSUPPORTED_DEVICES: set[tuple[int, int]] = set()
 _FICLONE_UNAVAILABLE = object()
 _FICLONE_IOCTL = None
@@ -373,12 +375,12 @@ def _clone_file(src, dst):
     Returns:
         True when cloning succeeds, or False to use the normal copy fallback.
     """
+    if on_linux:
+        return _clone_file_linux(src, dst)
     if islink(src) or lexists(dst):
         return False
     if on_mac:
         return _clone_file_macos(src, dst)
-    if on_linux:
-        return _clone_file_linux(src, dst)
     return False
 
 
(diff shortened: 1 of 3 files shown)


conda/gateways/disk/create.py
```diff
@@ -392,6 +417,57 @@ def _clone_file(src, dst):
     return False
 
 
+def _clone_file_linux(src, dst):
```
jaimergp, line 355: On Python 3.14, `shutil.copyfile` may use COW already ([see docs](https://docs.python.org/3/library/shutil.html#platform-dependent-efficient-copy-operations)). Are we accounting for that?

jezdez: Good catch, thanks. I checked this directly on Python 3.14.6 from conda-forge and btrfs. That build does not expose `os.copy_file_range`, so `shutil.copyfile()` uses `sendfile` and did not create shared extents. After syncing writes, btrfs extent accounting showed the `shutil` target as fully exclusive, while the `FICLONE` target shared the source extent.

For 64 MiB, `shutil.copyfile()` measured 8.62 ms versus 0.060 ms for gated `FICLONE`. For 512 x 64 KiB, it measured 23.55 ms versus 14.58 ms. I updated the PR description with the Python 3.14 comparison and extent check.

PR #16368, opened 2026-07-07
Use APFS clonefile for copy-mode installs (B26)

jezdez:
### Description

Part of conda/conda#15969 as a new Track B follow-up case (B26).

APFS supports copy-on-write file cloning via `clonefile`. This PR applies that primitive to conda's existing copy-mode file creation path, allowing copy-style installs on compatible APFS filesystems to avoid eager byte copies.

The implementation keeps conda's existing package-cache hardlink behavior as the source of truth for standard cache-backed installs, preserving the refcount behavior used by `conda clean --packages`.

### Focused measurement

Focused copy-mode file-creation measurements on macOS 26.5.2 arm64 with APFS and Python 3.10.20, using medians from 12 randomized runs against current `main` (`0cdbbf64e`):

| Case | `main` | B26 | Speedup |
| --- | ---: | ---: | ---: |
| 1 file, 64 MiB | 19.43 ms | 0.17 ms | 111x |
| 512 files, 64 KiB each | 76.86 ms | 50.77 ms | 1.51x |

These are file-creation primitive measurements, not current whole-transaction results. The stronger transaction-level APFS result is measured separately in B30 (#16376), which plans eligible package subtrees rather than only cloning individual files.

### Checklist - did you ...

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Disk utility functions for creating new files or directories."""
 
+import errno
 import os
 import sys
 import tempfile
 import warnings as _warnings
-from errno import EACCES, EPERM, EROFS
+from ctypes import CDLL, c_char_p, c_int, get_errno
+from errno import EACCES, ENOSYS, EPERM, EROFS, EXDEV
 from logging import getLogger
 from os.path import dirname, isdir, isfile, join, splitext
 from shutil import copyfileobj, copystat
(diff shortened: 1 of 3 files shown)


conda/gateways/disk/create.py
```diff
@@ -322,9 +338,58 @@ def copy(src, dst):
...
+    if _clone_file(src, dst):
+        try:
+            copystat(src, dst)
+        except OSError as e:  # pragma: no cover
+            log.debug("%r", e)
+        return
     _do_copy(src, dst)
 
 
+def _clone_file(src, dst):
```
jaimergp, line 283: Such a pity there's no stdlib method for this yet (xref https://github.com/python/cpython/issues/81338).

For now, can we add a docstring here explaining the basics? This is pretty lowlevel. What's the return value? Success / failure?

jezdez: Added. The docstring now explains that `_clone_file()` creates a copy-on-write clone when supported and returns `True` on success or `False` so the caller uses the normal copy fallback.

PR #16369, opened 2026-07-07
Use native Windows copy APIs for copy-mode installs (B27)

jezdez:
### Description

Part of conda/conda#15969 as Track B follow-up B27.

Windows exposes `CopyFile2` through CPython `_winapi` on Python 3.12 and newer. This PR uses that wrapper for copy-mode file creation with the same compatibility flag as `shutil.copy2`. Python 3.10 and 3.11 retain a cached `CopyFileW` fallback. Both paths avoid bouncing file data through Python read and write loops.

This PR is stacked on B26 (#16368) until that PR lands. The new Track B case here is the Windows copy backend.

### Measured impact

Focused Windows NTFS copy-mode benchmark on Garak, a Windows 11 ARM64 VM running in Parallels on macOS, using Python 3.13.13 from the repository `dev/start.bat` base environment. The table reports medians from 20 balanced repeats of conda Python copy, `CopyFileW`, and `_winapi.CopyFile2`. Each path applies `copystat`.

| Case | Python copy | `CopyFileW` | `CopyFile2` | `CopyFile2` speedup |
|---|---:|---:|---:|---:|
| 1 copy-mode file, 64 MiB | 44.02 ms | 15.11 ms | 16.18 ms | 2.72x |
| 512 copy-mode files, 64 KiB each | 1.897 s | 187.67 ms | 186.35 ms | 10.18x |
| 2,048 copy-mode files, 1 KiB each | 8.311 s | 951.91 ms | 944.57 ms | 8.80x |

Follow-up 80-repeat probes on the same Parallels VM did not reproduce the 7.1% single-file difference. With one shared destination and API time isolated, the paired median put `CopyFile2` 0.5% faster than `CopyFileW`. Repeating the original destination and `copystat` shape put `CopyFile2` 0.9% slower. The flag variants were also equivalent. Treat the native APIs as performance parity within about 1% on this fixture. `CopyFile2` preserves the material win over the Python loop while using the CPython-provided wrapper where available.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for inclusion in the next release notes?

### DIFF
--- a/conda/gateways/disk/create.py
+++ b/conda/gateways/disk/create.py
@@ -437,24 +437,48 @@ def _win_copy_file(src, dst):
     if _WIN_COPYFILE is _WIN_COPYFILE_UNAVAILABLE:
         return False
     if _WIN_COPYFILE is None:
-        try:
-            from ctypes import windll, wintypes
-        except ImportError:
-            _WIN_COPYFILE = _WIN_COPYFILE_UNAVAILABLE
-            return False
-        copy_file = windll.kernel32.CopyFileW
-        copy_file.restype = wintypes.BOOL
-        copy_file.argtypes = (wintypes.LPCWSTR, wintypes.LPCWSTR, wintypes.BOOL)
-        _WIN_COPYFILE = copy_file
+        _WIN_COPYFILE = _load_win_copy_file()
+    if _WIN_COPYFILE is _WIN_COPYFILE_UNAVAILABLE:
+        return False
 
-    # CopyFileW keeps Windows copy-mode installs in kernel32 instead of bouncing
-    # every file through Python's read/write loop. The target should not exist.
     if _WIN_COPYFILE(src, dst, True):
         log.log(TRACE, "windows copying %s => %s", src, dst)
         return True
     if lexists(dst):
         rm_rf(dst)
     return False
+
+
+def _load_win_copy_file():
+    """Return the best available native Windows file-copy function."""
+    try:
+        from _winapi import (
+            COPY_FILE_ALLOW_DECRYPTED_DESTINATION,
+            COPY_FILE_FAIL_IF_EXISTS,
+            CopyFile2,
+        )
+    except ImportError:
+        try:
+            from ctypes import windll, wintypes
+        except ImportError:
+            return _WIN_COPYFILE_UNAVAILABLE
+        copy_file = windll.kernel32.CopyFileW
+        copy_file.restype = wintypes.BOOL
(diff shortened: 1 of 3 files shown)


jezdez:
@zooba Do you have any concerns with this approach in general?

zooba:
No major concerns, but you might consider checking for `_winapi.CopyFile2` first and using that instead - without ctypes in the way it should be even faster. (`CopyFile2` also does the `copystat`, at least according to the docstring, which I'm pretty sure I wrote, so you can drop that bit too.)

Strictly speaking it's a private function and could change, but the definition matches the Windows API (which won't change) and the API is already future-proofed.

Also, if you want major speedups, run this on a Dev Drive 😉 It'll do copy-on-write automatically if both the source and destination are on the same drive.

jezdez:
> No major concerns, but you might consider checking for `_winapi.CopyFile2` first and using that instead - without ctypes in the way it should be even faster. (`CopyFile2` also does the `copystat`, at least according to the docstring, which I'm pretty sure I wrote, so you can drop that bit too.)
> 
> Strictly speaking it's a private function and could change, but the definition matches the Windows API (which won't change) and the API is already future-proofed.

Thanks, this was quite useful @zooba! I've made the changes to the PR. I've updated the stats above, but I've only run the benchmarks in a local VM in Windows on ARM64

> Also, if you want major speedups, run this on a Dev Drive 😉 It'll do copy-on-write automatically if both the source and destination are on the same drive.

Oooh, interesting, but as you know this would be quite tricky for most of Conda's users on Windows. Maybe a good thing to document though?

zooba:
> Maybe a good thing to document though?

Yeah, it should just require configuring the package dir and the envs dir to be on the Dev Drive. CPython has functions to identify a Dev Drive, the idea being that a Python-based tool could easily find and offer the reconfiguration. But documentation under some kind of "optimising Conda" docs would be the easiest way to include the recommendation.

Getting off the OS partition is the #1 way to speed things up, even if you don't reach for the other speedups that a Dev Drive will do. (And FWIW, mounting the partition into a directory on the OS drive is a big step backwards, but probably I'm one of the very few people who would think to try that!)

> I've updated the stats above, but I've only run the benchmarks in a local VM in Windows on ARM64

If I'm reading it right, the native API and ctypes came out about the same? That's interesting, but not a huge surprise that the cost of actually copying the file massively outweighs the calling overhead.

PR #16370, opened 2026-07-07
Add `conda_clean_paths` plugin hook

### LINKED ISSUE #14253: Refactor `conda clean` into a plugin
### Checklist

### What is the idea?

Refactor `conda clean` into a `conda_subcommands` plugin and add a new plugin hook for registering new cleanup tasks (e.g., `conda_cleanup_tasks`).

### Why is this needed?

`conda clean` is a great way to remove old or unused files from your `conda` installation. Currently it only finds and removes package caches (both unpackaged and bundled (`.tar.bz2`, `.tar.bz2.part`, `.conda`, and `.conda.part`) packages), index caches, temporary files (`.c~` and `.trash` files), and logfiles.

With the plugin architecture other plugins are now able to interface more closely with conda and they bring their own special files into the picture. E.g., conda-content-trust has its own folder where it collects signatures, conda-libmamba-solver creates cached solve files, etc.

Providing a standard framework for allowing these other plugins to extend the functionality of `conda clean` much the same way we do for `conda doctor` will go a long way.

### What should happen?

1. introduce a new `conda_cleanup_tasks` plugin hook
2. refactor `conda.cli.main_clean` into `conda.plugins.subcommands.clean`

### Additional Context

_No response_

kenodegard:
### Description

Partial implementation of [#14253](https://github.com/conda/conda/issues/14253): add a `conda_clean_paths` plugin hook so components can register removable file or directory paths for `conda clean`, without refactoring existing built-in clean targets.

- Introduce `CondaCleanPath` plugin type with a `find(target_prefix) -> paths` callable; plugins discover paths only — conda handles listing, confirmation, dry-run, and removal via existing `rm_items()` / `rm_rf()`.
- Add `conda_clean_paths` hookspec, `get_clean_paths()` on the plugin manager (sorted by name), and `flag` / `dest` properties for CLI integration.
- Wire plugin targets into `conda clean`: each registered plugin gets a dedicated `--<name>` flag in the Removal Targets group (e.g. `--notices-cache`). Plugin targets are **not** included in `--all`.
- Register a built-in `notices-cache` clean path that removes cached channel notice files under the user notices cache directory.

Existing `conda clean` behavior for tarballs, packages, index cache, tempfiles, and logfiles is unchanged.

> [!WARNING]
> **Tradeoff: plugin flags prevent lazy loading.** Each clean-path plugin is registered as a dedicated `--<name>` flag at parser setup time (`configure_parser` calls `get_clean_paths()`). That means the plugin manager must be loaded and all `conda_clean_paths` hook implementations must be importable before `conda clean --help` (or any `conda clean` invocation) can run — we cannot defer loading clean-path plugins until a flag is actually selected. A `--clean-path NAME` indirection (like `conda doctor`'s positional names) would allow lazy loading but at the cost of discoverability in `--help`. This PR chooses per-plugin flags; revisit if startup cost becomes an issue.

Closes #14253 (partial — hook + first built-in target; full subcommand refactor deferred).

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -96,6 +96,15 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         action="store_true",
         help="Remove log files.",
     )
+    from ..base.context import context
+
+    for clean_path in context.plugin_manager.get_clean_paths().values():
+        removal_target_options.add_argument(
+            clean_path.flag,
+            dest=clean_path.dest,
+            action="store_true",
+            help=clean_path.summary or f"Remove {clean_path.name} paths.",
+        )
 
     add_output_and_prompt_options(p)
 
@@ -369,6 +378,13 @@ def _execute(args, parser):
         # package caches
         return json_result
 
+    clean_paths = context.plugin_manager.get_clean_paths()
+    selected_clean_paths = [
+        clean_path
+        for clean_path in clean_paths.values()
+        if getattr(args, clean_path.dest, False)
+    ]
+
     if not (
         args.all
         or args.tarballs
@@ -376,12 +392,21 @@ def _execute(args, parser):
         or args.packages
         or args.tempfiles
         or args.logfiles
+        or selected_clean_paths
     ):
         from ..exceptions import ArgumentError
 
         raise ArgumentError(
             "At least one removal target must be given. See 'conda clean --help'."
         )
+
+    if selected_clean_paths:
+        target_prefix = os.fspath(context.target_prefix)
+        json_result["clean_paths"] = {}
+        for clean_path in selected_clean_paths:
+            paths = list(clean_path.find(target_prefix))
+            json_result["clean_paths"][clean_path.name] = {"files": paths}
+            rm_items(paths, **kwargs, name=f"{clean_path.name} path(s)")
 
     if args.tarballs or args.all:
         json_result["tarballs"] = tars = find_tarballs()
--- a/conda/plugins/types.py
+++ b/conda/plugins/types.py
@@ -53,7 +53,7 @@ if TYPE_CHECKING:
     SinglePlatformEnvironmentExport = Callable[[Environment], str]
     MultiPlatformEnvironmentExport = Callable[[Iterable[Environment]], str]
 
-    CondaCleanPathCallable: TypeAlias = Callable[[str], Iterable[str]]
+    CondaCleanPathCallable: TypeAlias = Callable[[str], Iterable[PathType]]
 
     # Callback type for health check fixer confirmation prompts.
     # Raises CondaSystemExit if user declines, or DryRunExit in dry-run mode.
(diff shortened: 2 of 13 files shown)


conda/plugins/types.py
```diff
@@ -866,6 +868,40 @@ class CondaPackageExtractor(CondaPlugin):
...
+    find: CondaCleanPathCallable
+    summary: str | None = None
+
+    @property
+    def flag(self) -> str:
+        return f"--{self.name}"
+
+    @property
+    def dest(self) -> str:
+        return f"clean_{self.name.replace('-', '_')}"
```
jezdez, line 902: This `dest` is not unique for all valid plugin names: `foo-bar` and `foo_bar` both become `clean_foo_bar`. Since `_execute()` later selects plugins by this attribute, passing one flag can run both clean providers. Please either validate unique/reserved CLI names/dests before registering the options, or store the selected clean target explicitly (for example with `append_const`) so each flag selects exactly one provider.

conda/cli/main_clean.py
```diff
@@ -96,6 +96,15 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         action="store_true",
         help="Remove log files.",
     )
+    from ..base.context import context
+
+    for clean_path in context.plugin_manager.get_clean_paths().values():
+        removal_target_options.add_argument(
+            clean_path.flag,
```
jezdez, line 103: This also needs to reject names that collide with built-in clean options. A plugin named `all` registers `--all` and makes parser construction fail with an argparse conflict; because parser generation builds all built-in subparsers, that can break startup before the user even runs `clean`. Please reserve the existing clean flags or skip/report those conflicts before calling `add_argument()`.

jezdez (commented):
Two parser edge cases I found while reviewing clean-path target registration.

conda/cli/main_clean.py
```diff
@@ -96,6 +96,15 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         action="store_true",
         help="Remove log files.",
     )
+    from ..base.context import context
+
+    for clean_path in context.plugin_manager.get_clean_paths().values():
```
jezdez, line 101: This calls every clean-path hook while the top-level parser is built. A hook that raises can break unrelated commands such as `conda install`. Dynamic flags also change existing argparse abbreviations. For example, `--index` currently means `--index-cache`, but a plugin named `index` takes it over.

Could this use one stable, repeatable option such as `--clean-path NAME`, with `--list` for discovery? Provider lookup could then happen when `clean` runs instead of while every command parser is built.

conda/cli/main_clean.py
```diff
@@ -369,20 +378,36 @@ def _execute(args, parser):
         # package caches
         return json_result
 
+    clean_paths = context.plugin_manager.get_clean_paths()
```
jezdez, line 381: This code is never reached when `--force-pkgs-dirs` is also set because the function returns above. `conda clean --force-pkgs-dirs --notices-cache` reports success without running the notices provider. Please run the selected providers before returning, or make `--force-pkgs-dirs` mutually exclusive with other targets.

jezdez (changes requested):
Requesting changes because the per-plugin flag approach can change or break the whole CLI.

The two existing name and destination collision threads are still open. There is also a broader problem: `generate_parser()` configures `clean` for every normal command. A failing clean-path hook or a bad plugin name can therefore stop unrelated commands before their arguments are parsed.

A single stable option would avoid most of this. For example, `conda clean --clean-path NAME` could be repeatable, with `conda clean --list` showing available providers. Provider lookup could then happen only when `clean` runs. The built-in `--notices-cache` option could remain a normal built-in flag if we want to keep it easy to find.

PR #16371, opened 2026-07-07
Aggregate transaction hardlink actions (B29)

jezdez:
### Description

Part of #15969 as Track B follow-up B29.

Install transactions often execute long runs of plain package hardlinks. This PR replaces adjacent exact LinkPathAction objects with a BulkHardLinkPathAction after transaction preparation has created compile actions and prefix records.

The wrapped actions still own verification, rollback, cleanup, and conda-meta path data. The bulk action reuses open directory file descriptors on supported platforms and uses a direct Windows hardlink attempt before conda's normal per-file fallback.

This was informed by the package-tree linking work in uv and Rattler/Pixi, but it is a narrower implementation. [Rattler groups paths by directory and executes groups through Rayon](https://github.com/conda/rattler/blob/main/crates/rattler/src/install/mod.rs); [uv walks a complete cached wheel tree with one link-state machine](https://github.com/astral-sh/uv/blob/main/crates/uv-fs/src/link.rs). This PR remains serial and only reuses hardlink setup.

This draft is stacked on B28 (#16367).

Part of the [conda-tempo Track B](https://github.com/jezdez/conda-tempo) performance work.

### Measured impact

The macOS measurements were rerun on AC power with no thermal or performance warning. A balanced ABBA warm-cache W1 comparison used eight recorded runs per branch after warming each branch:

| Workload | B28 median | B29 median | Result |
| --- | ---: | ---: | ---: |
| W1, normal hardlink install | 5.915 s | 5.919 s | neutral |

Focused 2,048-file runs can show a small action-loop win, but the direction does not hold across larger fixtures and does not survive the full transaction. There is currently no demonstrated user-visible speedup, so this remains a research draft.

Tracking issue: #15969. Full research report: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md

### Checklist - did you ...

- [x] Add a file to the news directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation? No user-facing documentation change.

PR #16376, opened 2026-07-07
Clone eligible package subtrees on APFS (B30)

jezdez:
### Description

Part of #15969 as Track B follow-up B30.

B26 through B28 add copy-on-write primitives for individual copy-mode files. This PR completes the macOS path by preferring copy semantics on clone-capable filesystems and cloning the largest eligible missing package subtree in one recursive `clonefile` call.

The planner walks the package manifest and extracted tree once. It descends through prefix directories that already exist, clones the deepest complete missing subtrees, and leaves existing target paths on conda's normal per-file path. A subtree is only eligible when its source and target paths match, every file is a normal copy-mode package path, and the extracted tree contains no untracked entries. Noarch Python packages remain on the existing per-file path because their paths are remapped.

Prefix-replaced files can be included in a directory clone, then are removed and recreated through `PrefixReplaceLinkAction`. Cloned directory modes and mtimes are normalized to conda's normal `mkdir` behavior. The original path actions still own verification, prefix metadata, cleanup, and rollback.

Cloning remains an implementation of `LinkType.copy`; this PR does not add a new link type. `conda-meta` therefore records cloned packages as copies, keeping clones independent from package-cache hardlink accounting. This addresses the metadata concern raised on the first B26 draft.

The macOS `clonefile(2)` documentation discourages arbitrary directory hierarchies in favor of recursive `copyfile(3)`. This implementation deliberately limits recursive `clonefile` to manifest-complete package subtrees with nonexistent destinations and per-file fallback. That matches the directory-clone paths in [uv](https://github.com/astral-sh/uv/blob/main/crates/uv-fs/src/link.rs) and [Rattler](https://github.com/conda/rattler/blob/main/crates/rattler/src/install/mod.rs), which Pixi uses. On the 2,956-file pandas package tree, one recursive `clonefile` call measured 36.4 ms versus 366.9 ms for `copyfile(COPYFILE_CLONE | COPYFILE_RECURSIVE)`.

This PR is based directly on B28 (#16367). B29 (#16371) was closed rather than retained in the stack.

### Measured impact

The macOS measurements were run on AC power with no thermal or performance warning. The table pools both command orders against B28, using the same warm package cache and separate prefixes.

| Workload | Runs per branch | B28 mean | B30 mean | Change |
| --- | ---: | ---: | ---: | ---: |
| W1 fresh Python + requests | 10 | 6.022 s | 5.939 s | -1.4% |
| W2 fresh data-science environment | 10 | 17.356 s | 13.411 s | -22.7% |
| Add W2 packages to an existing Python + requests prefix | 6 | 11.850 s | 8.693 s | -26.6% |

W2 system CPU time drops from 8.755 s to 4.567 s. Existing-prefix system CPU time drops from 6.420 s to 3.276 s. The existing-prefix result confirms that descending through shared prefix directories retains the directory-clone benefit.

### Checklist - did you ...

- [x] Add / update outdated documentation? The install deep dive now describes APFS copy-on-write directory clones and their `conda-meta` semantics.

### DIFF
--- a/conda/core/path_actions.py
+++ b/conda/core/path_actions.py
@@ -8,9 +8,11 @@ import os
 import re
 import sys
 from abc import ABC, abstractmethod
+from collections import defaultdict
 from itertools import chain
 from logging import getLogger
 from os.path import basename, dirname, getsize, isdir, isfile, join, normpath
+from stat import S_IMODE
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
@@ -42,6 +44,7 @@ from ..exceptions import (
 )
 from ..gateways.connection.download import download
 from ..gateways.disk.create import (
+    clone_directory,
     compile_multiple_pyc,
     copy,
     create_hard_link_or_copy,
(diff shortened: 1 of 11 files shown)


jezdez:
pre-commit.ci run

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
         self._execute_successful = True
 
 
+def _path_parts(path):
+    return tuple(part for part in path.replace("\\", "/").split("/") if part)
```
jaimergp, line 631: Isn't this the same as `tuple(Path(path).parts)`? Why the extra function?

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+
+    def reverse(self):
+        for action in reversed(self._actions):
+            action.reverse()
+
+    def cleanup(self):
+        for action in self._actions:
+            action.cleanup()
+
+    def _build_clone_plan(self):
```
jaimergp, line 706: This is the meat of the PR, can you add a small docstring explaining the overall logic? There are a few data structures here with non-obvious types.

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+        if safe_directories:
+            mode_probe = join(self.target_prefix, f".tmp.clone-mode.{uuid4().hex}")
+            os.mkdir(mode_probe)
+            try:
+                directory_mode = S_IMODE(os.stat(mode_probe).st_mode)
+            finally:
+                os.rmdir(mode_probe)
+
+        cloned_actions = set()
+        for top_level in sorted(expected_directories[()]):
```
jaimergp, line 675: An empty tuple as the key is an odd choice, let's cover this in the docstring of `_build_clone_plan()`.

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+            if error := action.verify():
+                return error
+        self._verified = True
+
+    def execute(self):
+        expected_directories, safe_directories, actions_by_directory = (
+            self._build_clone_plan()
+        )
+        directory_mode = None
+        if safe_directories:
```
jaimergp, line 666: Can you add a comment here on why we need this?

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+
+        for action in self._file_actions:
+            source_parts = _path_parts(action.source_short_path)
+            if not source_parts:
+                continue
+            if source_parts in seen_paths:
+                duplicate_paths.add(source_parts)
+            seen_paths.add(source_parts)
+            for index in range(len(source_parts) - 1):
+                expected_directories[source_parts[:index]].add(source_parts[index])
```
jaimergp, line 722:
```suggestion
            for index, source_part in enumerate(source_parts):
                expected_directories[source_parts[:index]].add(source_part)
```

Clearer?

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+                if islink(join(root, directory)):
+                    actual_files[root_parts].add(directory)
+                else:
+                    actual_directories[root_parts].add(directory)
+                    if directory in expected_children:
+                        traversable.append(directory)
+            directories[:] = traversable
+            actual_files[root_parts].update(files)
+
+        all_directories = set(expected_directories) | set(expected_files)
```
jaimergp, line 753:
```suggestion
        all_directories = set(*expected_directories, *expected_files)
```

Create the one set directly?

conda/core/path_actions.py
```diff
@@ -625,6 +628,207 @@ def execute(self):
...
+                cloned_actions.update(clone_actions)
+                return
+
+        if lexists(target_dir) and not isdir(target_dir):
+            return
+        children = expected_directories[directory]
+        if children:
+            mkdir_p(target_dir)
+            for child in sorted(children):
+                self._clone_missing_subtrees(
```
jaimergp, line 821: Recursion! 🏄

conda/core/path_actions.py
jaimergp: Also, is this worth parallelizing?

PR #16386, opened 2026-07-09
Get prefix record filename from name, version, build of package

### LINKED ISSUE #16235: Fix conda-meta filename for wheel packages on lockfile restore
### What is the idea?
`PrefixData._get_json_fn()` should derive the conda-meta filename from the package's name, version, and build string directly, rather than stripping it from the wheel filename.

### Why is this needed?
During lockfile restore, the wheel filename comes from the URL basename (e.g. pyqplot-0.7.2-py3-none-any.whl) which uses hyphenated wheel tags. The build string in the prefix record is conda-style (py3_none_any_0). The mismatch causes _load_single_record to reject the file, and conda list then shows conda-pypi packages as pip packages. Partially fixes #16217.

The normal solver path is unaffected because the solver record provides the correct fn directly (for example pyqplot-0.7.2-py3_none_any_0.json). This is only for lockfiles that go through the `_get_json_fn()`.

### What should happen?
`PrefixData.insert()` writes `conda-meta/{name}-{version}-{build}.json` regardless of the wheel filename or URL, and reloads cleanly.

Additional Context
Companion fix in conda-pypi (extractor index.json build string) is also needed for a full fix, see https://github.com/conda/conda-pypi/issues/417.

soapy1:
### Description

fixes https://github.com/conda/conda/issues/16235

### Checklist - did you ...

### DIFF
--- a/conda/core/prefix_data.py
+++ b/conda/core/prefix_data.py
@@ -489,16 +489,10 @@ class PrefixData(metaclass=PrefixDataType):
         return self
 
     def _get_json_fn(self, prefix_record: PrefixRecord) -> str:
-        fn = prefix_record.fn
-        # Check package extensions (dynamic) or .dist-info (pip-installed)
-        stripped, ext = strip_pkg_extension(fn)
-        if not ext and fn.endswith(".dist-info"):
-            stripped = fn[: -len(".dist-info")]
-        elif not ext:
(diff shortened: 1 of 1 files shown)


soapy1:
closing in favour of https://github.com/conda/conda/pull/16387

PR #16387, opened 2026-07-09
Fix conda-meta paths for wheel package records

### LINKED ISSUE #16235: Fix conda-meta filename for wheel packages on lockfile restore

### LINKED ISSUE #16265: Fix wheel conda-meta paths during unlink and rollback
### What is the idea?
Unify how conda derives conda-meta JSON filenames for repodata v3 wheel packages. Link and remove already use prefix_record.fn (via _get_json_fn()), but unlink and rollback use basename(extracted_package_dir), which produces a different name for the same package.

### Why is this needed?
For wheel packages from channels like conda-pypi, these two paths diverge (underscore repodata key vs hyphen PEP 427 wheel name). On a failed transaction, rollback tries to restore the wrong file, metadata is lost, and the environment can become invalid. This was observed in conda 26.5.3 during investigation of https://github.com/conda/conda-pypi/issues/419 and is separate from the pkgs-cache fix needed in #15620.

### What should happen?
All link, unlink, and rollback paths for wheel packages should read, write, rename, and restore the same conda-meta JSON file. If a transaction fails mid-unlink, rollback should restore the original metadata and leave the environment usable.

Concretely, `make_unlink_actions()` should `use _get_json_fn(prefix_record)` instead of `basename(extracted_package_dir)`.

### Additional Context
Related: conda #15620 (pkgs-cache naming)
conda-pypi #419 (base env corruption after failed self-update)

jezdez:
### Description

Fixes #16235.
Fixes #16265.

Conda-meta JSON paths now come from the prefix record's canonical `name-version-build` filename instead of the package filename or extracted package directory. This keeps prefix-data insert/remove, link, unlink, rollback, and package-size accounting aligned for package records whose artifact filename or extracted cache directory differs from the conda-meta filename, including wheel records from repodata v3 channels.

The filename helper now lives on `PrefixRecord`, with `PrefixData` delegating to it so existing insert/remove paths use the same rule.

### Checklist - did you ...

### DIFF
--- a/conda/models/records.py
+++ b/conda/models/records.py
@@ -708,6 +708,9 @@ class PrefixRecord(SolvedRecord):
     auth = StringField(required=False, nullable=True)
     """Authentication information."""
 
+    def _get_json_fn(self) -> str:
+        return f"{self.name}-{self.version}-{self.build}.json"
+
     def package_size(self, prefix_path: Path) -> int:
         """
         Compute the installed size of this package within a prefix.
@@ -721,9 +724,7 @@ class PrefixRecord(SolvedRecord):
         """
         total_size = 0
 
-        meta_file = (
-            prefix_path / "conda-meta" / f"{self.name}-{self.version}-{self.build}.json"
-        )
+        meta_file = prefix_path / "conda-meta" / self._get_json_fn()
         try:
             total_size += meta_file.stat().st_size
         except OSError:
--- a/conda/core/link.py
+++ b/conda/core/link.py
@@ -12,7 +12,7 @@ import warnings
 from collections import defaultdict
 from itertools import chain
 from logging import getLogger
-from os.path import basename, dirname, isdir, join
+from os.path import dirname, isdir, join
 from pathlib import Path
 from textwrap import indent
 from traceback import format_exception_only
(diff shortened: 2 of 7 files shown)


conda/models/records.py
```diff
@@ -708,6 +708,9 @@ class PrefixRecord(SolvedRecord):
     auth = StringField(required=False, nullable=True)
     """Authentication information."""
 
+    def _get_json_fn(self) -> str:
```
soapy1, line 711: Is there a motivation to prefix this funciton with `_`. It looks like it's being used by other modules. Can probably be renamed to `get_json_fn` to not imply anything special about this?

soapy1 (commented):
Looks good! For consistency, it might also be good to update the pypi prefix data loader plugin

https://github.com/conda/conda/blob/main/conda/plugins/prefix_data_loaders/pypi/__init__.py#L116

to be:
```
prefix_rec_json_path = (
            prefix_path / "conda-meta" / prefix_rec._get_json_fn()
        )
```

conda/models/records.py
jezdez: I’d keep the underscore for now. This is still an internal filename helper, and it mirrors the existing `PrefixData._get_json_fn`; using it across conda internals shouldn’t make it public API.

jezdez:
Good catch on the PyPI prefix-data loader cleanup path. Updated it to use `prefix_rec._get_json_fn()` and added a focused regression for the clobbered-record cleanup target.

soapy1 (approved):

PR #16388, opened 2026-07-09
Skip CUDA detection on macOS ARM64; handle multiprocessing error

dholth:
### Description

While running `conda` in a sandbox, I noticed a `PermissionError` that prevents conda from running. Since cuda isn't supported on osx-arm64 at all, we can safely skip the cuda check on that platform. This should save time as well.

Would this be faster on non-osx-arm64 platforms if we moved the cuda detection subprocess into `conda/_private` so that there were fewer collateral imports?

It offends me that we need to return `auxlib.NULL` to indicate "no cuda".

### Checklist - did you ...

### DIFF
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -10,6 +10,7 @@ import itertools
 import multiprocessing
 import os
 import platform
+import warnings
 from contextlib import suppress
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


news/osx-arm64-cuda-shortcut
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Shortcut CUDA detection on unsupported osx-arm64 platform; warn and assume not
+  supported on `PermissionError()` as when operating in a sandbox.
```
dholth:
```suggestion
  supported on `PermissionError()` as when operating in a sandbox. (#16388)
```

conda/plugins/virtual_packages/cuda.py
```diff
@@ -35,11 +36,28 @@ def cuda_version():
     Returns: version string (e.g., '9.2') or None if CUDA is not found.
```
jaimergp, line 36:
```suggestion
    Returns: version string (e.g., '9.2') or NULL if CUDA is not found.
```

jaimergp (approved):

conda/plugins/virtual_packages/cuda.py
```diff
@@ -35,11 +36,28 @@ def cuda_version():
...
     """
 
+    # Shortcut: CUDA is not available on macOS ARM64 (Apple Silicon)
+    system = platform.system()
+    machine = platform.machine()
+    if system == "Darwin" and machine == "arm64":
```
jezdez, line 42: Should this reuse on_mac?

dholth: I don't have a strong opinion on that, you're suggesting if on_mac and arm64?

PR #16392, opened 2026-07-10
Improve `cuda_version()` performance

dholth:
### Description

I noticed that `_cuda_driver_version_detector_target()` is a good candidate for inclusion in `conda/_private/`. Since it is called in a subprocess, we can save time by putting it somewhere that does not import most of conda inside the subprocess.

On my Linux machine it takes `.104s` with the current method and `.028s` with the improved method.

### Checklist - did you ...

### DIFF
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -10,7 +10,9 @@ import platform
 import warnings
 from typing import TYPE_CHECKING
 
-from conda._private.cuda import cuda_driver_version_detector_target
+from conda._private.cuda import (
+    cuda_driver_version_detector_target as _cuda_driver_version_detector_target,
+)
 
 from ...auxlib import NULL
 from .. import hookimpl
@@ -58,7 +60,7 @@ def cuda_version():
 
     # Spawn a subprocess to detect the CUDA version
     detector = context.Process(
-        target=cuda_driver_version_detector_target,
+        target=_cuda_driver_version_detector_target,
         args=(queue,),
         name="CUDA driver version detector",
         daemon=True,
--- a/conda/plugins/virtual_packages/cuda.py
+++ b/conda/plugins/virtual_packages/cuda.py
@@ -4,15 +4,13 @@
 
 from __future__ import annotations
 
-import ctypes
 import functools
-import itertools
 import multiprocessing
-import os
 import platform
 import warnings
-from contextlib import suppress
 from typing import TYPE_CHECKING
+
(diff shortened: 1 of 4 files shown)


jezdez:
I'd think we should move this into the plugin by Nvidia

dholth:
I worked on this because CUDA detection wasn't friendly when sandboxed (on an OS that doesn't support CUDA), then noticed we could save time by moving the subprocess to a different import path.

This change has nothing to do with `nvidia_virtual_packages` that doesn't use a subprocess at all.

conda/plugins/virtual_packages/cuda.py
```diff
@@ -58,14 +56,14 @@ def cuda_version():
         )
         return NULL
 
+    # Spawn a subprocess to detect the CUDA version
+    detector = context.Process(
+        target=cuda_driver_version_detector_target,
+        args=(queue,),
+        name="CUDA driver version detector",
+        daemon=True,
+    )
```
jaimergp, line 75: What's the difference moving this out of the `try` block? To avoid `NameError` in `finally` in case we fail to instantiate?

conda/plugins/virtual_packages/cuda.py
```diff
@@ -95,88 +93,3 @@ def conda_virtual_packages() -> Iterable[CondaVirtualPackage]:
         override_entity="version",
         # empty_override=NULL,  # falsy override → skip __cuda
     )
-
-
-def _cuda_driver_version_detector_target(queue):
```
jaimergp, line 100: No need to deprecate this because it was already private, true?

conda/plugins/virtual_packages/cuda.py
dholth: Exactly

conda/plugins/virtual_packages/cuda.py
dholth: I didn't consider that. It would be possible to import the new name inside this function "just in case"

jaimergp: Either that or a `deprecated.constant()` alias, but we need to leave something just in case.

jaimergp:
> I'd think we should move this into the plugin by Nvidia

https://github.com/conda-incubator/nvidia-virtual-packages only offers `__cuda_arch` so far, and it's not distributed yet on conda-forge or defaults. I agree on the principle, but we can already improve the situation here before those external prerequisites are met.

dholth:
Done - imported moved symbol as previous `_cuda_driver_version_detector_target` name.

PR #16393, opened 2026-07-10
move cuda detection subprocess code into conda._private

dholth:
### Description

Separate PR to get CI benchmark results.

### Checklist - did you ...

### DIFF
--- a/conda/_private/cuda.py
+++ b/conda/_private/cuda.py
@@ -1,0 +1,100 @@
+# Copyright (C) 2012 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+Code to detect CUDA version without importing much of conda.
+
+Called as a subprocess by conda.plugins.virtual_packages.cuda.
+"""
+
+from __future__ import annotations
+
+import ctypes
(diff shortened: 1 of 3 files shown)


jezdez:
I'd rather move this into the new https://github.com/conda-incubator/nvidia-virtual-packages repo TBH instead of further refactoring it in conda. @carterbox how would you feel about this?

carterbox:
Would you imagine that conda would now *require* the nvidia-virtual-packages plugin?

Otherwise, it would be breaking to remove `__cuda` from the virtual packages that conda can detect by default, and we would have to figure out some migration strategy for users.

carterbox:
Implementation-wise, it would make sense to detect both __cuda and __cuda_arch from the same place.

dholth:
Merged into #16392

jezdez:
> Would you imagine that conda would now _require_ the nvidia-virtual-packages plugin?
> 
> Otherwise, it would be breaking to remove `__cuda` from the virtual packages that conda can detect by default, and we would have to figure out some migration strategy for users.

Yeah, conda distributions would ship with the plugin

PR #16406, opened 2026-07-14
Fix `conda.common.io.IS_INTERACTIVE` deprecation

### LINKED ISSUE #16401: IS_INTERACTIVE import broken when lazy __getattr__ overwrites deprecation shim
### Checklist

### What happened?

Current main has a regression that breaks imports of `IS_INTERACTIVE` from `conda.common.io`. This causes external plugins that depend on this symbol to fail at load time.

User impact: The most visible downstream breakage is the Anaconda ToS plugin https://github.com/anaconda/conda-anaconda-tos/issues/294, which causes the plugin to fail to load

Expected behavior: IS_INTERACTIVE should remain importable throughout its deprecation cycle (pending deprecation now, removal scheduled for 27.3).

Actual behavior: The symbol cannot be imported, triggering ImportError for any code that imports it.

### Why does this happen?
Two recent changes on main most likely interact to create this regression:

1. #15982 converted IS_INTERACTIVE from a direct module attribute to a deprecation shim
2. #15881 added lazy import handling to the same file

The second change inadvertently overwrites the deprecation mechanism, making IS_INTERACTIVE inaccessible.

### What should be fixed?
The two import mechanisms need to coexist in conda/common/io.py. The deprecation shim should chain to the lazy import handler rather than being replaced by it.

We should also add test coverage for the specific import pattern that downstream consumers use to prevent similar regressions.

### Reproduction

```
from conda.common.io import IS_INTERACTIVE
# ImportError: cannot import name 'IS_INTERACTIVE' from 'conda.common.io'
```

### Conda Info

```shell
Reproducible with canary 26.5.3.95 or later
```

### Conda Config

```shell
# N/A - affects all configurations
```

### Conda list

```shell
# Affects any environment with conda-anaconda-tos or similar plugins
```

### Additional Context

Workaround: Downstream consumers can temporarily migrate to conda.common.terminal.is_tty() (the recommended replacement), though IS_INTERACTIVE should remain available until its scheduled removal in 27.3.

Related:

Upstream: #15982 (deprecation), #15881 (lazy imports)
Downstream: https://github.com/anaconda/conda-anaconda-tos/issues/294

kenodegard:
### Description

Reorder `IS_INTERACTIVE` deprecation so it doesn't get overwritten by `conda.common.io.__getattr__` implementation.

Followup to https://github.com/conda/conda/pull/15982 and https://github.com/conda/conda/pull/15881
Resolves #16401

### Checklist - did you ...

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -36,14 +36,6 @@ from .path import expand
 
 log = getLogger(__name__)
 
-deprecated.constant(
-    "26.9",
-    "27.3",
-    "IS_INTERACTIVE",
-    hasattr(sys.stdout, "isatty") and sys.stdout.isatty(),
-    addendum="Use `conda.common.terminal.is_tty()` instead.",
-)
-
 
 class DeltaSecondsFormatter(Formatter):
     """
@@ -577,6 +569,15 @@ def __getattr__(name: str):
         _load_concurrency()
         return globals()[name]
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+deprecated.constant(
+    "26.9",
+    "27.3",
+    "IS_INTERACTIVE",
+    hasattr(sys.stdout, "isatty") and sys.stdout.isatty(),
+    addendum="Use `conda.common.terminal.is_tty()` instead.",
+)
 
 
 def get_instrumentation_record_file() -> str:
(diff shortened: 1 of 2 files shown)


jezdez (approved):

jaimergp (approved):

ryanskeith (commented):
This might be a good practice to follow to have a deprecation warning test when things get deprecated so we don't repeat the same mishap.

PR #16414, opened 2026-07-16
Suggest using `conda self` to update if available

### LINKED ISSUE #16319: Recommend conda self update in outdated-conda warning when conda-self is installed
### What is the idea?

Keep the outdated-conda check in solver code for performance, but update the recommended command to `conda self update` whenever `conda-self` is installed in base (not only when base is frozen).

### Why is this needed?

The warning recommends `conda update -n base -c {channel} conda`, which often no-ops ([#12519](https://github.com/conda/conda/issues/12519)). Since conda 26.5+ depends on conda-self, most users should see `conda self update` instead.

Moving the check to conda-self via `post_command` would sacrifice index reuse (while we don't have sharded repodata on all channels. So updating the message in place is a low-cost fix https://github.com/conda/conda/issues/16216#issuecomment-4847312971).

### What should happen?

- [x] **libmamba:** Extend existing logic in [`solver.py`](https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/solver.py#L1068) — today `conda self update` is recommended only when base is frozen. Now we can recommend it whenever `conda-self` is in the prefix.
- [x] **rattler:** Add the same `conda-self` check to [`utils.py`](https://github.com/conda/conda-rattler-solver/blob/main/conda_rattler_solver/utils.py#L306) (currently always recommends `conda update`).
- [x] **classic solver:** Add the same check to [`solve.py`](https://github.com/conda/conda/blob/main/conda/core/solve.py#L1272).

**Out of scope:** Moving the check to conda-self `post_command` — deferred until per-name sharding is available on all channels conda users install from (defaults, conda-forge, etc.). Without shards, a post-command lookup falls back to full repodata fetches; solver index reuse is strictly better until then.

### Additional Context

- Ship after or with Issue 1 so the recommended command works.
- libmamba already has the `prefix_data.get("conda-self")` pattern

soapy1:
### Description

With this change, the conda classic solver will suggest using `conda self` to update conda when is outdated (if conda self is installed). For example,

```
$ conda list | grep conda-self
conda-self                 0.2.0            py310h06a4308_0

$ CONDA_SOLVER=classic conda create -n test imagesize --dry-run  
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 1.2.3
  latest version: 26.5.3

Please update conda by running

    $ conda self update
```

part of https://github.com/conda/conda/issues/16319

### Checklist - did you ...

### DIFF
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -268,7 +268,8 @@ class BaseSolver:
     def _notify_conda_outdated(self, link_precs):
         if not context.notify_outdated_conda or context.quiet:
             return
-        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
+        conda_prefix_data = PrefixData(context.conda_prefix)
+        current_conda_prefix_rec = conda_prefix_data.get("conda", None)
         if current_conda_prefix_rec:
             channel_name = current_conda_prefix_rec.channel.canonical_name
             if channel_name == UNKNOWN_CHANNEL:
@@ -296,6 +297,14 @@ class BaseSolver:
                 latest_version = conda_newer_precs[-1].version
                 # If conda comes from defaults, ensure we're giving instructions to users
                 # that should resolve release timing issues between defaults and conda-forge.
+                if conda_prefix_data.get("conda-self", None):
+                    conda_update_message = "conda self update"
+                else:
+                    conda_update_message = (
+                        f"conda update -n base -c {channel_name} conda"
+                    )
+                    if conda_prefix_data.is_frozen():
+                        conda_update_message += " --override-frozen"
                 print(
                     dedent(
                         f"""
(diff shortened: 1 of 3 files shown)


conda/core/solve.py
```diff
@@ -296,6 +297,14 @@ def _notify_conda_outdated(self, link_precs):
                 latest_version = conda_newer_precs[-1].version
                 # If conda comes from defaults, ensure we're giving instructions to users
                 # that should resolve release timing issues between defaults and conda-forge.
```
danyeaw:
```suggestion
```

I think this comment is stale and can be removed.

danyeaw (approved):
Thanks @soapy1!

jezdez (approved):

PR #16425, opened 2026-07-19
Fix bare except clauses across multiple files

koteshyelamati:
Replace bare `except:` clauses with specific exception types to avoid accidentally catching `BaseException` subclasses like `KeyboardInterrupt` and `SystemExit`.

**Files changed:**
- `conda/gateways/disk/delete.py`: 3 fixes (`except Exception:`, `except OSError:`)
- `conda/cli/common.py`: 1 fix (`except Exception:`)
- `conda/core/portability.py`: 1 fix (`except Exception:`)
- `conda/core/initialize.py`: 3 fixes (`except Exception:`)
- `conda/core/link.py`: 1 fix (`except Exception:`)

This is a safe, non-breaking change — bare `except:` and `except Exception:` behave identically for all practical exception types; the difference is that bare `except:` also catches `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which is almost never the intent.

### DIFF
--- a/conda/core/initialize.py
+++ b/conda/core/initialize.py
@@ -1455,7 +1455,7 @@ def init_fish_user(
             rc_content = fh.read()
     except FileNotFoundError:
         rc_content = ""
-    except:
+    except Exception:
         raise
 
     rc_original_content = rc_content
@@ -1596,7 +1596,7 @@ def init_xonsh_user(
             rc_content = fh.read()
     except FileNotFoundError:
         rc_content = ""
-    except:
+    except Exception:
         raise
 
     rc_original_content = rc_content
@@ -1775,7 +1775,7 @@ def init_sh_user(
             rc_content = fh.read()
     except FileNotFoundError:
         rc_content = ""
-    except:
+    except Exception:
         raise
 
     rc_original_content = rc_content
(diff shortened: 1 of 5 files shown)


koteshyelamati:
Thank you for letting me know! I'll review and sign the Contributor License Agreement to proceed with this contribution.

conda/core/initialize.py
```diff
@@ -1455,7 +1455,7 @@ def init_fish_user(
             rc_content = fh.read()
     except FileNotFoundError:
         rc_content = ""
-    except:
+    except Exception:
```
danyeaw, line 1458: These bare except blocks in this module don't do anything. Without the second handler, any non-FileNotFoundError exception already leaves the try naturally. With except Exception: raise, the same exception is caught and re-raised unchanged. Could you please just remove them instead of changing them?

conda/gateways/disk/delete.py
```diff
@@ -259,7 +259,7 @@ def backoff_rmdir(dirpath, max_tries=MAX_TRIES):
 
     try:
         rmtree(dirpath)
-    except:
+    except OSError:
```
danyeaw, line 262: Doesn't this one narrow the caught exceptions to pass with?

danyeaw (changes requested):
Hey @koteshyelamati, thanks for the contribution!

Did you sign the CLA with your full name (Kotesh Kumar Yelamati)? I'm looking for the PDF of it in the system.

PR #16433, opened 2026-07-21
Show installer metadata in `conda info`

### LINKED ISSUE #16432: Expose installer metadata in conda info
### Checklist

### What is the idea?

Constructor records the installer name, version, platform, and type in `$PREFIX/.installer.info`. Have `conda info` read that file when present and expose the known fields as installer metadata.

### Why is this needed?

Users and tooling need a central way to identify which distribution installed the base prefix and which installer produced it. This is especially useful when an update or uninstall operation must defer to an external installer or package manager. `conda info` is already the diagnostic interface for the current installation, so this does not require a new subcommand or plugin hook.

### What should happen?

Normal `conda info` output should include one concise distribution line when valid metadata exists. `conda info --json` should include a nested `installer` object with the supported `name`, `version`, `platform`, and `type` fields.

Installations without the file should keep their current text output. An unreadable or malformed file should not make `conda info` fail.

### Additional Context

Constructor added `.installer.info` in [conda/constructor#1250](https://github.com/conda/constructor/pull/1250). This request only consumes that existing file format.

jezdez:
### Description

Read Constructor's `$ROOT_PREFIX/.installer.info` metadata when it is present.

Normal `conda info` output gains one concise distribution line. JSON output gains a nested `installer` object with the installer name, version, platform, and type. Missing or malformed metadata leaves existing output unchanged.

Closes #16432.

### Checklist - did you ...

### DIFF
--- a/conda/cli/main_info.py
+++ b/conda/cli/main_info.py
@@ -30,6 +30,8 @@ if TYPE_CHECKING:
 
 log = getLogger(__name__)
 
+INSTALLER_INFO_FIELDS = ("name", "version", "platform", "type")
+
 
 def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser:
     from ..common.constants import NULL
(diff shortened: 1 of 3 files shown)


danyeaw (approved):
Thanks @jezdez!

PR #16434, opened 2026-07-21
Document that conda does not verify shard hashes

### LINKED ISSUE #16213: Docs: Update developer docs to clarify that conda doesn't check the hash of shard files
### Checklist

### What happened?

At the conda community office hours on 2026-06-08, a couple of folks from JFrog called in to discuss their implementation of sharded repodata for Artifactory.

In order for their implementation to work with [virtual repositories](https://docs.jfrog.com/artifactory/docs/virtual-repositories) they wanted to clarify that we would not start checking shard filename hashes in the future, because it would break their implementation based on how the virtual repos get combined.

Currently, conda and rattler aren't checking that the sha256 hash in the filename of a shard matches what is downloaded. I don't think that mamba is either, but I haven't verified this. I understand that the hash check is avoided because it would have a large performance penalty to check the hash of every shard.

A few options:
1. Update CEP 16 to make it clear that not checking the hash is the standard
2. Add this to our developer docs so that we know the rationale why we aren't checking the hash and the use case that JFrog brought up with their implementation
3. Add a feature config with a default config of disabled to check the hash

In scope of this issue is to update the developer docs and investigate how large the change would be to CEP 16 to clarify this.

### Additional Context

_No response_

travishathaway:
### Description

fixes: #16213 

On top of adding how we intentionally do not verify the content of shards by their hash, I have also taken this as a chance to tidy things up a bit and update some of the "index" pages to make them more readable.

### Checklist - did you ...


docs/source/dev-guide/deep-dives/index.md
jaimergp, line 1: Why the `.md` to `.rst` conversion? MyST should make both equivalent 🤔

docs/source/dev-guide/index.rst
```diff
@@ -11,20 +11,13 @@ Getting started :octicon:`rocket;1em;sd-text-primary`
 
 New to conda development? Start here.
 
-.. grid:: 1 2 2 2
-    :gutter: 2
-
-    .. grid-item-card:: Contributing 101 :octicon:`people;1em;sd-text-primary`
-        :link: contributing
-        :link-type: doc
+.. glossary::
```
jaimergp: This directive exports the symbols to a global Glossary somewhere, doesn't it? Should we use it for formatting reasons only?

docs/source/dev-guide/releasing.md
```diff
@@ -1,28 +1,9 @@
 # Releasing
 
-Conda's releases may be performed via the [rever command](https://regro.github.io/rever-docs/).
-Rever is configured to perform the activities for a typical conda release.
-To cut a release, simply run `rever <X.Y.Z>` where `<X.Y.Z>` is the
-release number that you want bump to. For example, `rever 1.2.3`.
+The conda release process is documented in `RELEASE.md` at the root of the repository on GitHub.
```
jaimergp:
```suggestion
The `conda` release process is documented in `RELEASE.md` at the root of the repository on GitHub.
```

docs/source/dev-guide/specs/index.md
jaimergp, line 1: Same here, why?

docs/source/dev-guide/implementation-notes/sharded-repodata.md
jaimergp, line 1: Is this a "spec" as in a "specification"? Reads more like implementation notes. Mentioning this because specifications belong in conda.org, but I think this is not such a thing. The content is fine though.

travishathaway: Fair comment. I was a little unsure about where to place this. I felt this was the least-bad place for it.

Do you have any other suggestions? Should this instead just be integrated into the existing deep-dive article?

docs/source/dev-guide/deep-dives/index.md
```diff
@@ -16,3 +16,38 @@ solvers
 logging
 sharded
 ```
+
+```{eval-rst}
+.. glossary::
```
travishathaway: This eval is here because when we render glossary definitions in markdown, it looks slightly different than when we do it in rst. I felt this was the easiest way around this problem for now.

docs/source/dev-guide/index.rst
travishathaway: I totally agree that we're abusing the semantic use of glossary here, but the formatting looks great! 🙃

In terms of accessibility, this is definitely a minus because users will assume this is just a list of terms and not and menu. I'm not sure how important that is to address though.

The other thing is that I've already used this directive in other index pages, so at this point we have to update everything else.

One idea is that we make our own directive that copies this style verbatim but uses fully semantic tags (e.g. `<nav>`) so the HTML reflects the actually purpose of the links. I don't know if this is really worth the effort though.

docs/source/dev-guide/deep-dives/index.md
jaimergp: That's so interesting and weird but can we do 😂

docs/source/dev-guide/implementation-notes/sharded-repodata.md
jaimergp: I would honestly create a new section "Implementation notes" with `/impl/` as the directory or something, and then put this one underneath. Even the current "spec" in this section is a "technical specification" which in hindsight is also implementation notes. 

So what if we rename this section from specifications to implementation notes, and add a small redirect to cover `/specs/` -> `/implementation-notes/`.

docs/source/dev-guide/index.rst
jaimergp: Maybe we can simply define a custom RST directive that is nothing more than an alias to glossary so this doesn't look as weird?

docs/source/dev-guide/implementation-notes/sharded-repodata.md
travishathaway: Sounds like a good idea. I'll go ahead and do this.

jezdez: What's the difference to deep-dives? Let's not invent new sections if we end up with just one document under it

docs/source/dev-guide/index.rst
travishathaway: I added a new `nav-glossary` directive for us.

travishathaway:
@jaimergp, this is ready for another review.

kenodegard:
> What's the difference to deep-dives? Let's not invent new sections if we end up with just one document under it

I agree with @jezdez here. The new shard page repeats a lot of what we already have in deep-dives, and a parallel “implementation notes” section (née specs) for that plus solver-state feels redundant.

I'd rather fold the hash-validation notes into `deep-dives/sharded`, move `solver-state` under deep-dives, and drop the extra section (with redirects from the old `specs/` URLs).

Separately, I pushed a small fix so `{nav-glossary}` works in Markdown without `{eval-rst}`.

kenodegard:
Opened https://github.com/travishathaway/conda/pull/6 with proposed combining of implementation-details with deep-dives

travishathaway:
@kenodegard,

Thanks for the suggestions. I agree with all of them and have merged them into this repository. I think this pull request is ready to approve and merge now.

kenodegard (approved):

dholth:
I've realized that rattler stores shards in `rattler/cache/repodata/shards-v1` by hash but conda stores them by URL. If a proxy decides to re-use upstream hashes while changing the content, then that would break the cache.

PR #16437, opened 2026-07-22
Fix timeout() leaving SIGALRM armed on unexpected exceptions

### LINKED ISSUE #15702: conda/common/io.py::timeout() can leave SIGALRM enabled if callable raises an unexpected exception

Adelagric:
Closes #15702.

## Problem

`conda.common.io.timeout()` arms a `SIGALRM` alarm, then only cancels it (`signal.alarm(0)`) when the wrapped callable returns normally:

```python
signal.signal(signal.SIGALRM, interrupt)
signal.alarm(timeout_secs)
try:
    ret = func(*args, **kwargs)
    signal.alarm(0)          # only reached on success
    return ret
except (TimeoutException, KeyboardInterrupt):
    return default_return
```

If `func()` raises any *other* exception, the alarm stays armed and later fires during unrelated code — surfacing as a spurious `TimeoutException` far from its origin (the symptom reported in #15702). The original `SIGALRM` handler is also never restored.

## Fix

Move the cleanup into a `finally` block that always cancels the alarm and restores the previously installed handler.

## Tests

Added `test_io.py` tests: one for the normal return value, and one asserting that after the callable raises an unexpected exception the alarm is cancelled and the previous handler is restored (Unix-only, since `timeout()` uses `SIGALRM`). Verified the test fails against the current code and passes with the fix.

## Note

Verified in isolation and with `ruff`; I couldn't run the full suite locally (the test conftest pulls in extras like `flask`/`xprocess`), so I've relied on CI for the complete run.

### DIFF
--- a/conda/common/io.py
+++ b/conda/common/io.py
@@ -473,15 +473,19 @@ def timeout(
         def interrupt(signum, frame):
             raise TimeoutException()
 
-        signal.signal(signal.SIGALRM, interrupt)
+        previous_handler = signal.signal(signal.SIGALRM, interrupt)
         signal.alarm(timeout_secs)
 
         try:
-            ret = func(*args, **kwargs)
-            signal.alarm(0)
-            return ret
+            return func(*args, **kwargs)
         except (TimeoutException, KeyboardInterrupt):  # pragma: no cover
             return default_return
+        finally:
+            # Always cancel the alarm and restore the previous handler, even if
+            # func() raised an unexpected exception; otherwise the alarm stays
+            # armed and can later fire during unrelated code. See #15702.
+            signal.alarm(0)
+            signal.signal(signal.SIGALRM, previous_handler)
 
 
 def _load_concurrency() -> None:
--- a/news/15702-timeout-sigalrm-leak
+++ b/news/15702-timeout-sigalrm-leak
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix `conda.common.io.timeout()` leaving the `SIGALRM` alarm armed (and the previous signal handler unrestored) when the wrapped callable raised an unexpected exception, which could cause the alarm to fire later during unrelated code. The alarm is now always cancelled and the previous handler restored via a `finally` block. (#15702)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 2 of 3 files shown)


Adelagric:
Heads-up: I found this overlaps with #15767 by @Br1an67, which proposed the same `finally`-block fix for the alarm back in March — credit to them for spotting it first. This PR additionally restores the previous `SIGALRM` handler (it was being overwritten and never reset) and adds regression tests. Happy to defer to #15767 if preferred; I opened this since that one appears stalled on the CLA check. Either way, @Br1an67's original diagnosis stands.

jezdez (approved):
I reproduced this against current `main`. An unexpected exception leaves the alarm armed and the replacement handler installed. This change cancels the alarm and restores the previous handler on every exit path, and the regression coverage verifies both. The implementation looks good to me.

PR #16456, opened 2026-07-28
Add missing recipes and correct recipes README generation

kenodegard:
### Description

Recently added dummy packages did not include a `meta.yaml` recipe so the `generate_recipes_readme.py` failed to include it in the recipes README. Others didn't commit recipes so couldn't be easily rebuilt.

While fixing these issues also made the following changes:
- dropped `font-ttf-inconsolata` in favor of other lighter weight packages
- reworked versioned recipes into variant recipes
- added descriptions for all recipes
- ensured all recipes are valid YAML format

### Checklist - did you ...


danyeaw (approved):
Hi @kenodegard, these changes look great! I pushed a couple of small cleanup changes:
1. Update the gitignore files to have the .conda packages
2. Update the dev-guide to the correct directory
3. Updates to the generate packages script to add a comment to the top, a shebang, and run conda-index as the last step

Please let me know if those updates look good to you.

tests/data/recipes/track_feature/conda_build_config.yaml
```diff
@@ -0,0 +1,9 @@
+pkg_version:
+  - "1.0"
+  - "2.0"
+feature:
+  - track_feature1
+  - track_feature2
+zip_keys:
+  - - pkg_version
```
ryanskeith, line 8: Neat.  I didn't know `- -` could start lists.

tests/data/recipes/generate_recipes_readme.py
ryanskeith, line 1: This file contains a some one-time-use helper functions that could be inlined for easier reading. Not a big deal but it did take me more time to read.

ryanskeith (approved):
These look good to me.  I learned something new with the ` - -` syntax in yaml.   Just as a note, I did see that the `.conda` is a bigger format in case we do want to save disk space. This is very small though bytes or kilobytes but there are many just in case being as small as possible is the goal.

PR #16457, opened 2026-07-28
Fix typos

mwtoews:
This fixes typos identified and fixed with codespell. This check can be added as a pre-commit, if requested.

Please review for false positives.


conda/core/initialize.py
```diff
@@ -1225,7 +1225,7 @@ def install_anaconda_prompt(target_path, conda_prefix, reverse):
     if not context.dry_run:
         create_shortcut(
             "%windir%\\System32\\cmd.exe",
-            "Anconda Prompt",
+            "Anaconda Prompt",
```
mwtoews, line 1228: This change might have some implications....

PR #16467, opened 2026-07-30
Unwrap CondaMultiError for guidance

### LINKED ISSUE #16462: [QA] RemoveError doesn't surface structured guidance
### Checklist

### What happened?

While testing #16036 (structured user-facing error model), we tested 
`conda remove conda` expecting a clear, structured `RemoveError` — i.e. a 
summary, cause, and actionable hint, consistent with the new error model.

On Linux, no structured/clear error message appears up front. Instead, the 
command proceeds directly into the removal transaction, prompts for 
confirmation, and only fails at the very end with a bare `RemoveError` line 
and no summary, cause, or hint.

### What did you expect to happen?

A clear `RemoveError` shown with a summary, cause, and actionable hint — 
consistent with the structured user-facing error model introduced in #16036 
(similar to what `PackagesNotFoundError` now provides).

### Steps to reproduce

```bash
conda remove conda
```

### Actual output (Linux)
```bash
The following packages will be SUPERSEDED by a higher-priority channel:
  conda-content-tru~ pkgs/main::conda-content-trust-0.3.2-~ -> repo/main::conda-content-trust-0.2.0-py314h06a4308_1

Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: failed

RemoveError: This operation will remove conda without replacing it with another version of conda
```
No summary, cause, or hints are shown — just the bare `RemoveError` string, 
preceded by an unrelated package-supersession prompt that gives no indication 
up front that the operation is going to fail.

### Environment

- conda version: 26.7.0 (canary/dev)
- OS: Linux
- Related to: #16036 (Add structured user-facing error model to CondaError)

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

_No response_

kenodegard:
### Description

Resolves #16462

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release's release notes?<br>**Added directly to CHANGELOG.md since we are on the release branch**
     

conda/exceptions.py
```diff
@@ -1744,36 +1829,21 @@ def print_conda_exception(exc_val: CondaError, exc_tb: TracebackType | None = No
...
         (stdout if rc else stderr)(f"{exc_json}\n")
     else:
         from .gateways.streams import stderr
 
-        if guidance is not None:
-            stderr(f"\n{guidance.format(exc_val)}\n")
-        else:
-            stderr(f"\n{exc_val!r}\n")
+        stderr(
+            f"\n{_format_leaf_errors(_iter_leaf_errors(exc_val), get_error_hints)}\n"
```
ryanskeith: Should we denote that multiple errors were triggered?

conda/exceptions.py
```diff
@@ -1731,6 +1731,91 @@ def maybe_raise(error: BaseException, context: Context):
...
+
+def _format_leaf_errors(leaves: Iterable[BaseException], get_error_hints) -> str:
+    """Format leaf errors for human output; dedupe hints by ``hint_code``."""
+    from dataclasses import replace
+
+    from ._private.exception_guidance import ErrorGuidance
+
+    shown: ErrorGuidance | None = None
+    parts: list[str] = []
+    for error in leaves:
```
ryanskeith, line 1768: Flat is most likely fine here as I am not seeing to many instances of nesting errors.  I would worry more for the user if we had more instances of nested errors, in which case I think we would want to denote it so the user is lead to try and fix the root cause first.

I wonder if there is any sort of general guidance that we should give in a multi error case?  Should we say fix the first and try again.  Should we suggest fixing them all and trying again?  Does the liklihood of false positives go up after the first error?

kenodegard: Generally we use `CondaMultiError` to aggregate errors from a phase—often parallel work (verify/execute, fetch/extract), sometimes sequential validation (env/config). Parallel workers don’t spawn their own pools, so we don’t get deep nestings from “pools of pools.” Nesting can still show up where a worker returns a `CondaMultiError` and the caller wraps again (link execute), though that path already tries to flatten:

https://github.com/conda/conda/blob/69533289e557f503cb6b0b2bd5a500534f528ad1/conda/core/link.py#L1082-L1091

In practice I only remember seeing single-leaf multis (e.g. verify wrapping one `RemoveError`); I don’t recall a real case with several independent user-facing errors at once.

Let’s revisit hint dedupe and/or denoting nested errors if/when we hit a practical multi-leaf or nested example.

conda/exceptions.py
kenodegard: It's a little messy to read but it already renders the error name, e.g.:

```
RemoveError: Conda cannot remove itself.

Cause: The planned transaction would uninstall the conda package without linking a replacement.
Next steps:
  - (install_conda_version) Install a specific conda version first, for example:
      conda install conda=<version>
  - (use_non_base_env) Use a separate environment for other packages instead of changing the base environment where conda runs.

EnvironmentNotWritableError: The current user does not have write permissions to the target environment.
  environment location: /Users/kodegard/.conda/devenv/Darwin/arm64/envs/devenv-3.10-miniconda
  uid: 501
  gid: 20
```

danyeaw (approved):

PR #16471, opened 2026-07-31
Add `shards_only_hint` to `UnavailableInvalidChannel` error

### LINKED ISSUE #16453: Hint to enable sharded repodata when classic repodata.json is missing but a shard index exists
### Checklist

### What is the idea?

When sharded repodata is turned off and monolithic `noarch/repodata.json` is missing for a channel subdir, but `repodata_shards.msgpack.zst` is available, conda should fail with a clear hint to enable sharded repodata instead of only reporting that the channel is unavailable.

### Why is this needed?

Shards-only channels (e.g. `conda-pypi`) are unusable when monolithic repodata is required. Users who disable shards (or hit a path that only loads monolithic repodata) currently see a generic unavailable-channel error. An actionable message points them at `repodata_use_shards` / `--repodata-use-shards`.

This is separate from skipping monolithic fallback when shards are already enabled: different code path, can ship independently.

### What should happen?

- On monolithic `noarch/repodata.json` 404 while sharded repodata is disabled, probe for `repodata_shards.msgpack.zst` (HEAD, or GET).
- If a shard index is present, raise an error that says the channel appears shards-only and how to enable sharded repodata.
- If no shard index is present, keep today’s unavailable-channel behavior.
- Do not probe for a shard index before every monolithic fetch (only after a 404).
- Do not attempt to install from the channel via monolithic repodata when shards are required.

### Additional Context

- Related: #16452
- #16183 / #16184 
- conda/conda-pypi#397.

ForgottenProgramme:
### Description
resolves #16453 

Add `shards_only_hint` bool to `UnavailableInvalidChannel` error. When set to `True`, the error message hints that shards are turned off for them, and they can turn them on using `--repodata-use-shards`. 

### Checklist - did you ...


conda/exceptions.py
ryanskeith, line 581: It looks like the shards_only_hint may get overwritten here.  If a JSON reason/message is supplied. Would it be better to add it after this?  Or does it make sense that the response would contain the necessary info for the user?

ryanskeith (commented):
Looks good overall.  I am just wondering if we would ever have a case where the addition would be overwritten and if that is the correct thing to do.

conda/gateways/repodata/__init__.py
```diff
@@ -188,6 +188,23 @@ def _add_http_value_to_dict(resp, http_key, d, dict_key):
...
 
+def _shard_index_exists(subdir_url: str) -> bool:
+    try:
+        session = get_session(subdir_url)
+        url = join_url(subdir_url, REPODATA_SHARDS_FN)
+        timeout = (
+            context.remote_connect_timeout_secs,
+            context.remote_read_timeout_secs,
+        )
+        # Try HEAD first. If HEAD not supported, try GET.
```
danyeaw: I think in practice just doing a HEAD request is plenty with no fall back. Sorry, I know the issue mentioned both.

If we are going to do a GET request, we probably want to stream it and close it immediately, otherwise we'll download the full shard index which is pretty wasteful for just finding out if it exists or not.

ForgottenProgramme:
pre-commit.ci run

conda/exceptions.py
```diff
@@ -587,6 +588,14 @@ def __init__(
                 if isinstance(detail, str):
                     message = detail
 
+        # Hint user to enable shards when channel is shards only
+        if shards_only_hint:
+            message += dedent("""
```
danyeaw: Hey @ForgottenProgramme, did you consider using the new Structured Errors with the guidance dictionary here? The output would be a bit nicer. Currently it shows:

```
Artifact noarch/repodata.json not found
This channel appears to provide only sharded repodata.
Enable shards with `conda config --set repodata_use_shards true`
or pass `--repodata-use-shards`.
```

Instead it would look like:

```
Artifact noarch/repodata.json not found

Cause: This channel appears to provide only sharded repodata.
Next steps:
  - (enable_repodata_shards) Enable shards with `conda config --set repodata_use_shards true` or pass `--repodata-use-shards`.
```

news/16471-add-enable-shards-hint
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Display a "enable shards" hint to users when a channel is shards only and shards are disabled.(#16453)
```
danyeaw:
```suggestion
* Display a "enable shards" hint to users when a channel is shards only and shards are disabled. (#16453)
```

danyeaw (approved):
Hi @ForgottenProgramme, this looks really good! I tested it locally as well now that the changes to 26.7 have been merged back to main. Just one suggestion on using the new error capabilities in conda.

ForgottenProgramme:
Thanks @danyeaw !
Done. :)

danyeaw (approved):

PR #16330, opened 2026-07-02
Use conda-launchers entry point stubs

jezdez:
### Description

Use the published `conda-launchers >=24.7.1` package for Windows entry points. Select `cli-32.exe`, `cli-64.exe`, or `cli-arm64.exe` from `share/conda-launchers` according to the target Python architecture.

Remove the bundled 64-bit launcher. Retain the existing bundled 32-bit stub only when the launcher package does not provide one, preserving `win-32` support with defaults.

Resolve the launcher from its installed package record or an extracted package in the current transaction, and verify its SHA-256 immediately before copying. This supports fresh installations and launcher upgrades.

Refs #16293 and #15448. Related implementation: #16488.

### Checklist - did you ...

- [x] Add a file to the `releases/news` directory ([using the template](https://github.com/conda/conda/blob/main/releases/news/TEMPLATE)) for the next release's release notes?
- [x] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #16488, opened 2026-08-03
Use conda-launchers as a runtime dependency

### LINKED ISSUE #15448: Complete native Windows ARM64 support
### Checklist

- [x] I searched open issues and could not find a duplicate maintainer tracker

### What?

Use this maintainer-owned epic to track the work required for conda to run, test, and be distributed natively on Windows ARM64.

The original feature request and community discussion remain in #11472. Launcher design, signing, packaging, and integration are tracked in #16293. Native package CI is tracked in #16540.

### Why?

An x64 conda installation can run on Windows ARM64 through emulation, but native support requires hosted CI, native dependency packages, signed native entry-point launchers, updated conda and conda-build packages, and a working bootstrap path.

### User impact

- Conda runs and is tested with native `win-arm64` Python and packages.
- Python entry points use native signed ARM64 launchers.
- Creating environments for another supported Windows architecture continues to work with `--platform` and conda-build.
- Fresh installations and upgrades receive the required launcher package while retaining files needed by older conda during the transition.

### Goals

- [x] Run the conda test suite on hosted Windows ARM64 runners. This is currently x64 emulation. (#16466)
- [ ] Run the test suite with native `win-arm64` Python and dependency packages. (#16540)
- [x] Publish a conda-forge `noarch: generic` launcher package containing CLI and GUI files for win-32, win-64, and win-arm64. (#16293)
- [x] Publish a defaults `noarch: generic` launcher package containing CLI and GUI files for win-64 and win-arm64. This package does not provide win-32 files. (#16293)
- [x] Build and sign each final launcher once per signing identity, then reuse the exact signed files across package formats and recipe rebuilds. (#16293)
- [ ] Complete final signature, hash, and native ARM64 execution checks on the production files with Defender and SmartScreen enabled. Package verification and a fresh Windows 11 test are available, but the published report does not explicitly establish native ARM64 execution. (#16293)
- [ ] Make conda and conda-build consume the published launcher package, with compatibility files retained where released conda and win-32 support require them. (#16293)
- [ ] Verify the selected launcher against its package metadata immediately before conda or conda-build copies it. (#16293)
- [ ] Publish compatible conda and conda-build packages in conda-forge and defaults after the launcher packages. (#16293)
- [ ] Verify a supported fresh-install or bootstrap path and upgrades from released conda using the final packages.

### Tasks

- [x] #16466
- [ ] #16540
- [ ] #16293
- [ ] #11472

### This epic is blocked by:

- [ ] Remaining native dependency packages and source bootstrap work in #16540. Native libmamba/libmambapy are available in defaults, while conda-forge publication remains pending.
- [ ] Launcher integration, compatibility with released conda, and dependent package publication in #16293.
- [ ] Final native Windows ARM64 bootstrap and execution verification.

### This epic blocks:

- [ ] Closing the original Windows ARM64 support request in #11472
- [ ] Declaring native Windows ARM64 support complete

### LINKED ISSUE #16293: Use conda-launchers package for Windows entry point stubs
### What is the idea?
Stop bundling Windows entry-point stub executables in conda/shell/ and instead resolve them at runtime from the installed conda-launchers package (>= 26.7.0) via a small importable conda_launchers API. This centralizes build, signing, and updates in [conda/conda-launchers](https://github.com/conda/conda-launchers) and enables native win-arm64 launcher support without emulation.

### Why is this needed?
conda currently vendors older Anaconda-signed stubs under conda/shell/ and maps win-arm64 to the x64 stub. [conda-launchers](https://github.com/conda/conda-launchers) is now the canonical build and signing source (including NumFOCUS-signed release assets). A runtime dependency on conda-launchers matches the long-term direction in #15448 and keeps conda independent of which certificate signed the stubs in a given distribution.

### What should happen?

**conda-launchers**

- Merge [conda/conda-launchers#31](https://github.com/conda/conda-launchers/pull/31): add `conda_launchers` helpers for launcher names, package-relative paths, and prefix-qualified paths; release as `conda-launchers` 26.7.0.

**conda**

- Merge [conda/conda#16330](https://github.com/conda/conda/pull/16330):
  - Add `conda-launchers >=26.7.0` as a required dependency.
  - Add resolver in `conda.core.launchers` that uses `conda_launchers` and verifies the installed prefix record owns the expected path.
  - Update Windows `conda init` and noarch entry-point actions to use the resolver.
  - Remove bundled `conda/shell/cli-*.exe` stubs.
  - Mark `WINDOWS_LAUNCHER_STUB_PATH` in conda/base/constants.py` as pending-deprecated.
  - Update or relocate signature verification tests (currently in `tests/test_codesigned.py` since stubs will no longer live in the conda source tree.

**conda-build**

- Apply the same pattern: depend on `conda-launchers` and resolve launcher paths at runtime instead of vendoring stubs (similar to https://github.com/conda/conda-build/pull/5545.

**Packaging (per distribution)**

- Ship `conda-launchers` alongside `conda` in each distribution that adopts this change.
- **conda-forge:**
  1. Land https://github.com/conda-forge/staged-recipes/pull/34195 (creates the `conda-launchers` feedstock from NumFOCUS-signed conda-launchers release assets).
  2. Once that feedstock is approved and has published packages, update `conda-feedstock` to depend on `conda-launchers` (run dependency alongside `conda`). Supersedes https://github.com/conda-forge/conda-feedstock/pull/307.
- **Anaconda/defaults:** ship `conda-launchers` as a dependency. Signing strategy TBD — may use NumFOCUS-signed release assets directly (same as conda-forge) or optionally Anaconda could re-sign with Anaconda's certificate.

**Closure**

- Tracked by #15448.
- Supersedes #14401 and the vendoring plan originally proposed here.

### Additional Context

- Upstream signing and release work: https://github.com/conda/conda-launchers/issues/25
- Prior signing context: #13685
- The release assets were submitted to WDSI. Microsoft reported that SmartScreen reputation had not yet been established and would accumulate through use.

jaimergp:
### Description

Use the published `conda-launchers >=24.7.1` package as a Windows runtime dependency. Select the launcher for the target Python architecture from its fixed paths under `share/conda-launchers`, including native `win-arm64`.

Resolve the source from the installed launcher package or its extracted files in the current transaction. Verify its SHA-256 immediately before copying. Retain the existing bundled 32-bit stub when the launcher package does not provide one, preserving `win-32` support with defaults. Keep the legacy `conda/shell/cli-64.exe` path so upgrades started by released conda can finish creating entry points.

Supersedes #16330 and includes its launcher resolver. Refs #16293, #15448, and #14401. The feedstock transition is tracked in conda-forge/conda-feedstock#307 and AnacondaRecipes/conda-feedstock#94.

### Checklist - did you ...

- [x] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?


tests/requirements.txt
```diff
@@ -2,6 +2,7 @@ archspec >=0.2.3
 backports.zstd >=1.3.0
 boltons >=23.0.0
 charset-normalizer
+conda-launchers
```
jaimergp:
```suggestion
conda-forge::conda-launchers
```

news/16488-launchers
```diff
@@ -0,0 +1,38 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
```
jaimergp:
```suggestion
```

Oops copy paste artifact?

jezdez:
I think fixed package-relative paths are the simplest interface:

- `share/conda-launchers/cli-32.exe`
- `share/conda-launchers/cli-64.exe`
- `share/conda-launchers/cli-arm64.exe`

This avoids coupling Python helper changes to releases of the signed executables. Conda should require the first `conda-launchers` version that guarantees these paths.

The current implementation needs more than changing `WINDOWS_LAUNCHER_STUB_PATH`. The constant is still resolved relative to `CONDA_PACKAGE_ROOT`, which points inside `site-packages/conda` after installation. The [Windows CI failure](https://github.com/conda/conda/actions/runs/30886189875/job/91917893494) confirms that it looks for `conda/share/conda-launchers/cli-64.exe`, not the file under the prefix that contains `conda-launchers`.

I suggest keeping the resolver mechanics from #16330 while replacing its imported helper with the fixed mapping above. The resolver should:

- Find `conda-launchers` in the prefix that contains the running conda.
- Verify that the `conda-launchers` prefix record owns the expected file.
- Use the extracted `conda-launchers` package when it is being linked in the same transaction.

I do not think a PR that deletes the bundled launchers can ship only `cli-64.exe`. That would leave win-arm64 on x64 emulation and would map win-32 to a 64-bit executable. We should either wait until `cli-32.exe`, `cli-64.exe`, and `cli-arm64.exe` are available and switch together, or retain the bundled launchers during the transition.

If win-arm64 continues using `cli-64.exe` temporarily, this PR should not close #16293. Native support remains incomplete until conda selects and tests `cli-arm64.exe`.

For the final test, a clean Windows 11 ARM VM on Apple Silicon should be suitable. It should be restored to a clean snapshot before installing the production-signed package, with Defender and SmartScreen enabled.

jaimergp:
> The current implementation needs more than changing `WINDOWS_LAUNCHER_STUB_PATH`. The constant is still resolved relative to `CONDA_PACKAGE_ROOT`, which points inside `site-packages/conda` after installation. The [Windows CI failure](https://github.com/conda/conda/actions/runs/30886189875/job/91917893494) confirms that it looks for `conda/share/conda-launchers/cli-64.exe`, not the file under the prefix that contains `conda-launchers`.
> 
> I suggest keeping the resolver mechanics from #16330 while replacing its imported helper with the fixed mapping above. The resolver should:
> 
> * Find `conda-launchers` in the prefix that contains the running conda.
> * Verify that the `conda-launchers` prefix record owns the expected file.
> * Use the extracted `conda-launchers` package when it is being linked in the same transaction.

Do we need to do all of that? The smallest fix now is to replace `CONDA_PACKAGE_ROOT` with `context.conda_prefix`. The previous implementation didn't check that ownership nor tried to leverage a more recent `conda` in the transaction. I don't disagree that they are nice improvements, but they are not required for this migration, I'd say. Would it be ok to tackle that separately in upcoming PRs?

> I do not think a PR that deletes the bundled launchers can ship only cli-64.exe. That would leave win-arm64 on x64 emulation and would map win-32 to a 64-bit executable. We should either wait until cli-32.exe, cli-64.exe, and cli-arm64.exe are available and switch together, or retain the bundled launchers during the transition.
> 
> If win-arm64 continues using cli-64.exe temporarily, this PR should not close https://github.com/conda/conda/issues/16293. Native support remains incomplete until conda selects and tests cli-arm64.exe.

I'm fine leaving ~16293~ 15448 (you meant that one, right?) open, but I don't see the point of keeping unused executables in the repo if the code never points to them. Is this out of concern that this is "part of the API"?

jezdez:
@jaimergp I updated this PR to use the published `conda-launchers` packages, select the correct target architecture including ARM64, and verify hashes before copying. The changes handle upgrades and preserve 32-bit support with a bundled fallback. I also merged #16330 into this PR and closed it as superseded.

conda/core/launchers.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Prefer the extracted package during upgrades, before installed files are unlinked.
+    launcher_info = next(
+        (
+            info
+            for info in source_package_infos
+            if info.repodata_record.name == "conda-launchers"
+        ),
+        None,
+    )
```
jaimergp, line 59: If several versions or builds are available, which one are you selecting here? We should sort them by version and build number if possible. Also check that the subdir is the expected one (`noarch` I assume?).

conda/core/link.py
```diff
@@ -515,6 +515,7 @@ def _prepare(
                     target_prefix,
                     lt,
                     specs,
+                    packages_info_to_link,
```
jaimergp, line 518: Good think this function is private, otherwise this is an API breakage, isn't it?

conda/core/launchers.py
```diff
@@ -0,0 +1,106 @@
...
+                launcher := _find_launcher(prefix, record.paths_data, short_path)
+            ):
+                return launcher
+
+    # Retain the bundled stub until defaults publishes a 32-bit launcher.
+    if subdir == "win-32":
+        return (
+            join(CONDA_PACKAGE_ROOT, "shell", "cli-32.exe"),
+            "0170dda609519c088b1e4619a1e1d15a01701a6c514bb55f99a85fcbbd541631",
+        )
```
jaimergp, line 85: Should we check for existence of this file first?

conda/core/path_actions.py
```diff
@@ -364,14 +363,17 @@ def create_python_entry_point_windows_exe_action(
...
-        source_short_path = WINDOWS_LAUNCHER_STUB_PATH[context.subdir]
+        from .launchers import get_windows_launcher_stub
+
+        source_exe_path, sha256 = get_windows_launcher_stub(
+            target_prefix,
+            source_prefixes=(context.conda_prefix,),
+            source_package_infos=source_package_infos,
+        )
+        source_directory = dirname(source_exe_path)
+        source_short_path = basename(source_exe_path)
```
jaimergp, line 376: You have changed the meaning of these paths (short path used to be something like `etc/shell/win-32.exe`, now it's only `win-32.exe`), but I guess it doesn't matter because the relevant part is that they are joined into the final path eventually. Just flagging to check this is intended and not an oversight.

conda/core/path_actions.py
```diff
@@ -364,14 +363,17 @@ def create_python_entry_point_windows_exe_action(
         target_prefix,
         requested_link_type,
         entry_point_def,
+        source_package_infos=(),
```
jaimergp, line 366: Not relevant for this PR but we do need type hints in this module. It's so confusing.

conda/core/initialize.py
```diff
@@ -1180,25 +1178,23 @@ def make_entry_point(target_path, conda_prefix, module, func):
...
-        )
     exe_path = target_path
-    source_exe_path = join(
-        CONDA_PACKAGE_ROOT, WINDOWS_LAUNCHER_STUB_PATH[context.subdir]
+    source_exe_path, sha256 = get_windows_launcher_stub(
+        conda_prefix, source_prefixes=(conda_prefix, context.conda_prefix)
     )
     if isfile(exe_path):
-        if compute_sum(exe_path, "md5") == compute_sum(source_exe_path, "md5"):
+        if compute_sum(exe_path, "sha256") == sha256:
```
jaimergp, line 1189: Worth flagging in the news file, I'd say?

jaimergp (commented):
Only a couple comments for clarification. The main issue I see with the added changes is the fact that we may pick blindly any `conda-launchers` in the cache, which may be non deterministic unless we sort explicitly. I can't approve or reject my own PR so consider this a conditional approval provided my comments are discussed or solved.

jaimergp:
Ah, and to clarify, we are not removing the bundled binaries with this PR, even though we are not using `win-64.exe` right? We do want `win-32.exe` as a fallback but why keep -64 around?

conda/core/launchers.py
```diff
@@ -0,0 +1,106 @@
...
+            launcher_info.extracted_package_dir, launcher_info.paths_data, short_path
+        ):
+            return launcher
+    else:
+        for prefix in source_prefixes:
+            record = PrefixData(prefix).get("conda-launchers", None)
+            if record is not None and (
+                launcher := _find_launcher(prefix, record.paths_data, short_path)
+            ):
+                return launcher
```
jezdez, line 75: Following up on [your noarch check](https://github.com/conda/conda/pull/16488#discussion_r3966407746), here is the suggested change for both lookup paths.

```suggestion
    if launcher_info is not None:
        if launcher_info.repodata_record.subdir != "noarch":
            raise SafetyError("Expected conda-launchers from the noarch subdir.")
        if launcher := _find_launcher(
            launcher_info.extracted_package_dir, launcher_info.paths_data, short_path
        ):
            return launcher
    else:
        for prefix in source_prefixes:
            record = PrefixData(prefix).get("conda-launchers", None)
            if record is None:
                continue
            if record.subdir != "noarch":
                raise SafetyError("Expected conda-launchers from the noarch subdir.")
            if launcher := _find_launcher(prefix, record.paths_data, short_path):
                return launcher
```

releases/news/16293-conda-launchers-entrypoints
```diff
@@ -0,0 +1,11 @@
+### Enhancements
+
+* Use the published `conda-launchers` package for Windows entry points, including native ARM64 launchers, and verify their package hashes before copying. Retain bundled launchers for upgrades from released conda and for packages without a 32-bit launcher. (#16293)
```
jezdez: Following up on [your news comment](https://github.com/conda/conda/pull/16488#discussion_r3966529588), this makes the change in conda init explicit too.

```suggestion
* Use the published `conda-launchers` package for Windows entry points, including native ARM64 launchers, and verify their SHA-256 hashes before copying. Use SHA-256 instead of MD5 when deciding whether `conda init` needs to replace an existing launcher. Retain bundled launchers for upgrades from released conda and for packages without a 32-bit launcher. (#16293)
```

conda/core/launchers.py
jezdez: Thanks for checking this. Conda keeps one installed build per package name, and `PrefixData.insert()` rejects a second record with the same name. `source_package_infos` contains the packages already selected for this transaction, so there should be at most one `conda-launchers` version and build in that list.

During an upgrade, the old build can still be installed while the selected replacement is already extracted. Preferring the extracted package here is intentional, including for a downgrade. Several builds can also exist in the package cache, but this function does not search those for candidates.

Good point about `noarch`, though. I’d check that for both the extracted package and an already installed package. I put the [suggestion on those lines](https://github.com/conda/conda/pull/16488#discussion_r3966711745) so it can be applied directly.

conda/core/link.py
jezdez: Yep, thankfully this one is private. Its only caller in conda is updated here too. The extra argument passes the transaction’s selected packages through to entry-point creation.

conda/core/launchers.py
jezdez: Yes, good catch. Checking here would give a much clearer error if the bundled fallback is missing. We’d still verify its SHA-256 immediately before copying.

```suggestion
    # Retain the bundled stub until defaults publishes a 32-bit launcher.
    if subdir == "win-32":
        path = join(CONDA_PACKAGE_ROOT, "shell", "cli-32.exe")
        if not isfile(path):
            raise FileNotFoundError(f"Missing bundled Windows launcher {path!r}.")
        return (
            path,
            "0170dda609519c088b1e4619a1e1d15a01701a6c514bb55f99a85fcbbd541631",
        )
```

conda/core/path_actions.py
jezdez: Yep, this is intentional, but I can see why it looks odd. The resolver returns the full source path, which we split into its parent directory and filename. `CreateInPrefixPathAction.source_full_path` joins them again, so verification and copying still use the same resolved file. The destination remains `Scripts/{command}.exe`. Thanks for double-checking.

conda/core/path_actions.py
jezdez: Oh yes, following these arguments takes more effort than it should. Agreed that a separate typing cleanup would help.

conda/core/initialize.py
jezdez: Yes, let’s call that out explicitly. The existing entry mentions hash verification, but misses the change from MD5 to SHA-256 in `conda init`. I added the [suggested wording to the news entry](https://github.com/conda/conda/pull/16488#discussion_r3966712014).

jezdez:
@jaimergp Yep, the 64-bit copy stays for upgrades from released conda. I should have called that out.

An older conda process still runs its already-loaded entry-point code after the transaction replaces conda's package files. If it then needs to create a Windows entry point, that code still reads `conda/shell/cli-64.exe`. Removing the bundled file would break that step, even with `conda-launchers` installed. The new resolver cannot help the already-running process.

The [upgrade regression test](https://github.com/conda/conda/blob/7d173277505e2fabb6bc8b094fa42080d4a07784/tests/test_create.py#L2393) covers this with conda 26.7.2 and Pygments. Once the upgraded conda starts, it uses the packaged 64-bit launcher, but we still need the bundled copy to let older clients complete the upgrade.

jaimergp:
Got it, so we only need it for one release cycle but I assume we'll play it safe and leave it a bit longer?

jaimergp:
Consider this my approval once the open suggestion comments get committed.

jezdez:
@jaimergp I'd keep it for now, but it's not quite a one-release-cycle requirement. Users can skip releases and upgrade straight from conda 26.7.2 to a much later version. That upgrade would still run the old entry-point code and need `conda/shell/cli-64.exe`.

So before removing the bundled copy, we'd need to make sure those older clients can still upgrade safely, or have a way to ensure they upgrade through an intermediate release first. Keeping it for an extra cycle alone wouldn't address direct upgrades from older versions.

jaimergp:
Ah I hadn't thought about that, good point. That said, IIRC the old code path relied on copying `Scripts/conda.exe` instead of the unbranded copy under `shell/`. Can we double check? 🙏

jezdez:
@jaimergp Yep, you remember correctly! We switched from `Scripts/conda.exe` to the bundled stubs in [#15678](https://github.com/conda/conda/pull/15678), which shipped in [26.3.0](https://github.com/conda/conda/releases/tag/26.3.0).

I double-checked 26.7.2, which our upgrade test uses. Its [entry-point code](https://github.com/conda/conda/blob/68e257311038f6b6599c0241d929f95a3b1da03c/conda/core/path_actions.py#L373-L374) reads from `CONDA_PACKAGE_ROOT`, with [win-64 mapped to `shell/cli-64.exe`](https://github.com/conda/conda/blob/68e257311038f6b6599c0241d929f95a3b1da03c/conda/base/constants.py#L112-L115). So the bundled copy is still needed for upgrades from releases with that code, including 26.7.2.

jaimergp:
Thanks for checking. Let's add a markdown file next to the bundled files explaining their new purpose and (if any) removal plans. After that we should merge!

jezdez (approved):
Looks good to me. The launcher selection, hash checks, and upgrade compatibility concerns are addressed, and the README explains why the bundled files remain. Thanks, Jaime!

PR #16494, opened 2026-08-05
Skip monolithic repodata.json fallback for shards-only channels

### LINKED ISSUE #16452: Skip monolithic repodata.json fallback for shards-only channels
### Checklist

### Checklist

### What is the idea?

For shards-only channel subdirs, conda should not keep falling back to monolithic `repodata.json` after it already knows shards work and monolithic repodata is missing.

Learn shards-only in conda only: after a valid shard index exists for a subdir, a monolithic `repodata.json` **404** means “no classic for this subdir.” Later shard misses should use the shard-index cache (or fail clearly), not request monolithic again.

Dual-format channels (shards + monolithic, e.g. conda-forge) must keep the monolithic repodata as a recovery path until monolithic is proven missing.

### Why is this needed?

Channels such as `conda-pypi` publish `repodata_shards.msgpack.zst` and do not publish usable classic `repodata.json`. After shards are known, classic fallback 404s (or would be unusable at multi-GB size). However, treating every `has_shards` channel as shards-only would wrongly remove classic recovery on dual-format channels.

Related: #16183 / #16184 (stale negative shard cache). This covers the remaining “don’t classic-fallback once classic is known absent” case.

### What should happen?

- Successful shard fetch for a URL still uses shards for that URL (unchanged).
- On classic `repodata.json` 404 when the subdir already had shards: record that classic is unavailable, prefer the cached shard index, and do not keep probing classic.
- On later shard-index refresh failure for that subdir: use the shard cache and do not request `repodata.json`.
- On shard failure for dual-format subdirs (monolithic never 404’d): still allow monolithic fallback.

### Additional Context

- Inference signal: valid shards + classic 404 → `has_repodata_json: false` (or equivalent) in repodata cache state. Bare `has_shards: true` alone must not disable classic fallback.
- Also relevant: `RepodataCache.load_state()` should not clear state when only a shard-index cache file exists (no classic JSON cache).

ForgottenProgramme:
### Description
resolves #16452 

Ensure that whenever classic repodata.json fetching 404s, we check for the presence of shards. If shards are present we conclude that the channel is shards only and mark `has_format("repodata_json")` as False. 

<s>
#### Note on CI failures

Seem to be unrelated to this PR. 
They are related to a newer version of `packaging` .It is raising `InvalidVersion` instead of `TypeError`. Conda is only catching `TypeError` so the tests fail. 
Should be a simple fix in deprecations.py to also catch  `InvalidVersion` error. But its not in the scope of this PR. 
</s>

### Checklist - did you ...


conda/gateways/repodata/__init__.py
```diff
@@ -907,8 +911,26 @@ def fetch_latest(self) -> tuple[dict | str, RepodataState]:
...
                 if self.repodata_fn != REPODATA_FN:
                     raise  # is UnavailableInvalidChannel subclass
+
+                # when self.repodata_fn==repodata.json, and RepodataIsEmpty error is raised:
+                # if shards are available, we can assume that repodata.json is not supported
+                has_shards, checked = cache.state.has_format("shards")
+                if (
+                    checked is not None and has_shards
+                ) or cache.cache_path_shards.exists():
+                    cache.state.set_has_format("repodata_json", False)
```
danyeaw: A `shards_available()` helper here would be nice to reduce duplication with the same logic a few lines below.

conda/_private/shards/shards.py
danyeaw, line 714: One of the asks in the issue is if a shard-index refresh fails, we use the cached shard index and do not keep probing classic.

Today, when the index is stale and _repodata_shards fails, fetch_shards_index leaves shards_data is None and returns None even if cache_path_shards still exists. That pushes the URL onto the classic fallback path (or drops it once classic is gated off), instead of serving the last good index.

What if we added some logic after the above exception handlers, to check if shards_data is None and the shard-index file exists, reuse it when either:

- has_shards is still True (typical transient / non-4xx failure)
- classic is known absent

danyeaw, line 743: This is less likely if we implement the on-disk recovery above, but if we are inside the stale-refresh path, `fetch_shards_index` can still return None (which could result in 404s until the flag `set_has_format` is recorded).

What if we did something like if classic is known absent and cache_path_shards exists, load/parse that file and return Shards (and restore has_shards=True if needed).

conda/gateways/repodata/__init__.py
danyeaw, line 916:
```suggestion
)
# Shards-only: skip classic probe when recorded absent (all-None
# fallback path bypasses the fetch_channels gate).
if (
    self.repodata_fn == REPODATA_FN
    and not cache.state.should_check_format("repodata_json")
    and cache.cache_path_shards.exists()
):
    return "{}", cache.state

try:
```

When every fetch_shards_index returns None, fetch_channels returns None before the new should_check_format("repodata_json") check. Callers then use SubdirData → fetch_latest, which still always probes classic. What if we add a skip here before the try to avoid hitting the network again?

danyeaw (commented):
Hi @ForgottenProgramme, really nice job so far, this is a complex part of the code base to work in for sure. I think it would also be good getting @dholth as a 3rd reviewer once we are happy with the implementation to pull in some of his expertise in this part of the code base.

conda/gateways/repodata/__init__.py
ForgottenProgramme: done.

conda/_private/shards/shards.py
ForgottenProgramme: working on this.

conda/gateways/repodata/__init__.py
ForgottenProgramme: Makes sense!

ForgottenProgramme:
@danyeaw , thanks for your review on this PR. Your suggested changes make sense! 👍🏼 
I have now updated the PR and also added tests for the new logic added.

conda/_private/shards/shards.py
danyeaw: Hey @ForgottenProgramme, I think we may still get in to a stuck event here:

1. Expected on conda-pypi channel: classic 404 → has_repodata_json=False, and today also save("{}") to cache.
2. Bad event: later stale shard refresh gets a 4xx → has_shards=False, shard file left on disk.

Result is we are stuck ~7 days: Recovery only runs inside that if, so you return None and don’t use the on-disk index until the 7-day re-probe window opens. Maybe something like this with a test:

```
    # Shards-only: prefer on-disk index even when should_check_format("shards") is
    # False (e.g. negative has_shards within the re-check interval after a 4xx).
    if (
        repo_cache.cache_path_shards.exists()
        and not cache_state.should_check_format("repodata_json")
    ):
        shards_index_url = f"{sd.url_w_credentials}/{REPODATA_SHARDS_FN}"
        with repo_cache.lock("r+"):
            shards_data = repo_cache.cache_path_shards.read_bytes()
        has_shards, _ = cache_state.has_format("shards")
        if not has_shards:
            cache_state.set_has_format("shards", True)
            repo_cache.refresh()
        shards_index: ShardsIndexDict = msgpack.loads(
            capped_decompress(shards_data, max_output_size=ZSTD_MAX_SHARD_INDEX_SIZE)
        )  # type: ignore
        return Shards(shards_index, shards_index_url)

    return None
```

tests/gateways/test_repodata_gateway.py
```diff
@@ -457,3 +457,44 @@ def test_get_cache_control_max_age():
...
+        tmp_path / "cache",
+        channel,
+        REPODATA_FN,
+        repo_interface_cls=CondaRepoInterface,
+    )
+    fetch.repo_cache.state.set_has_format("shards", True)
+    fetch.repo_cache.save(b"shards")
+    fetch.fetch_latest()
+    assert fetch.repo_cache.load_state().has_format("repodata_json")[0] is False
+
```
danyeaw, line 481: It might be good to add a `assert not fetch.repo_cache.cache_path_json.exists()` check to ensure that we aren't blocking a shards force-recheck.

danyeaw (commented):
Wow, this is looking great! The main fixes you made look great and that closes the most common state transitions. I think there is still one path here that might get stuck.

Should we also have a test that combines a normal channel and a shards only channel, and tests the outcome if the shards fetch is 404 (fetch_latest_parsed is never called for the classic-absent URL)?

tests/gateways/test_repodata_gateway.py
ForgottenProgramme: Hey, Dan!
I tried that assert after the soft-404 test, but it fails. It's because soft 404 still does save("{}"), so `cache_path_json` exists afterward. [That’s the existing behavior](https://github.com/conda/conda/blob/c46ac95a7e9a58e5cc682558961b69c9c0aaa4eb/conda/gateways/repodata/__init__.py#L973-L975).

The hard noarch 404 path doesn’t write that file. 

I added your suggested assert statement before the `fetch_latest()` to check the setup. Adding the assert after the fetch fails.

conda/gateways/repodata/__init__.py
```diff
@@ -897,6 +905,14 @@ def fetch_latest(self) -> tuple[dict | str, RepodataState]:
                 self.cache_path_json,
             )
 
+        # avoid network calls if repodata_json is set to False and return "{}""
```
danyeaw:
```suggestion
        # avoid network calls if repodata_json is set to False and return "{}"
```

Nit

tests/gateways/test_repodata_gateway.py
danyeaw: Ok, I think that is fine to keep this out of this PR. This would only matter if somehow a user got in to a no shard file and a negative shard flag situation, which would only really happen if the channel was broken for a prolonged time. Thanks!

conda/_private/shards/shards.py
```diff
@@ -819,12 +844,17 @@ def fetch_channels(url_to_channel: dict[str, Channel]) -> dict[str, ShardBase] |
...
-            futures_non_sharded[
-                executor.submit(
-                    SubdirData(Channel(channel_url)).repo_fetch.fetch_latest_parsed
-                )
-            ] = channel_url
+            sd = subdir_data[channel_url]
+            cache = sd.repo_fetch.repo_cache
+            cache.load_state()
+
+            # skip classic when has_repodata_json was recorded False
```
danyeaw:
```suggestion
            # Skip classic when has_repodata_json is False until
            # CHECK_ALTERNATE_FORMAT_INTERVAL (should_check_format) expires
```
Small nit for accuracy.

danyeaw (approved):
Hi @ForgottenProgramme, this looks really good! The residual-edge recovery and test_cached_shards_when_shards_check_skipped look good, and the mixed-channel gate test covers the fetch_channels skip. That closes the stuck path I was worried about when should_check_format("shards") is False.

I think it is still worth us getting @dholth to give this a 2nd review to look for any major issues we might have missed.

conda/_private/shards/shards.py
```diff
@@ -627,6 +627,14 @@ def _repodata_shards(url, cache: RepodataCache) -> bytes:
 # shards as not supported; otherwise, we will check again next time.
 
 
+def _shards_from_bytes(shards_data: bytes, shards_index_url: str) -> Shards:
```
jezdez, line 630: This is already in a private package, no need to use private names

conda/_private/shards/shards.py
```diff
@@ -705,15 +713,32 @@ def fetch_shards_index(sd: SubdirData) -> Shards | None:
                     cache_state.set_has_format("shards", False)
                 repo_cache.refresh()
 
+        # Use cached on-disk shards data in case re-fetch fails above or classic repodata.json is missing
+        if shards_data is None and repo_cache.cache_path_shards.exists():
+            has_shards, checked = cache_state.has_format("shards")
+            shards_still_ok = checked is not None and has_shards
+            classic_absent = not cache_state.should_check_format("repodata_json")
```
dholth, line 720:
```suggestion
            classic_absent = not cache_state.should_check_format("json")
```

Since we call the other formats "shards", "zst" it would make sense to call it just "json". I notice we never bother to use a constant in the existing formats.

conda/_private/shards/shards.py
```diff
@@ -705,15 +713,32 @@ def fetch_shards_index(sd: SubdirData) -> Shards | None:
...
+            if shards_still_ok or classic_absent:
+                with repo_cache.lock("r+"):
+                    shards_data = repo_cache.cache_path_shards.read_bytes()
```
dholth, line 723: Can we use repo_cache.load(binary=True) in all cases instead of `with repo_cache.lock()... cache_path_shards.read_bytes()`?

`load()` may or may not have to be changed to support this case; `lock()` opens the state file, but `load()` additionally reads it as json. (Oh, this work is [in the following PR](https://github.com/conda/conda/pull/16069))

conda/_private/shards/shards.py
ForgottenProgramme: I used the existing pattern in that file. We can change it throughout the code in a later PR if this is considered valuable.

conda/_private/shards/shards.py
ForgottenProgramme: I used the existing pattern in that file. We can change it throughout the code in a later PR if this is considered valuable.

danyeaw (approved):

dholth:
https://github.com/conda/conda/commit/733fdd54bc4ab8539e45f0eac202cdbad652bceb is my suggestion but it's not straightforward to emit it as a PR against this branch.

The format really should be called "json" and not "repodata_json" for consistency. The linked commit changes them all to `FORMAT_<name>` variables.

Is it correct that the repodata exception caught here has already checked that repodata.json.zst and repodata.json were unavailable?

conda/_private/shards/shards.py
dholth, line 630:
```suggestion
def shards_from_bytes(shards_data: bytes, shards_index_url: str) -> Shards:
```

ForgottenProgramme:
@dholth , 
with the default `ZstdRepoInterface` (`repodata_use_zst` is on by default), yes, `request_url_zstd_state` tries `repodata.json.zst` and then falls back to `repodata.json` before we hit the `RepodataIsEmpty` / `UnavailableInvalidChannel` handlers where we set the flag. But with zst off, only `repodata.json` is fetched.

My change only records the negative classic flag after that fetch raises, it doesn’t add its own zst/json probing.

I am going to merge this PR as is. Could you please submit your commit as a PR against main and I will review and merge it.

conda/_private/shards/shards.py
ForgottenProgramme: Lets get this in the next PR also. I don't want the CI to run for hours for this.

ForgottenProgramme:
Nevermind, I created the PR myself: https://github.com/conda/conda/pull/16538

PR #16499, opened 2026-08-06
Add conda config --clear option

### LINKED ISSUE #7617: conda config clear option

codewithdaniel1:
### Description

Adds a `conda config --clear KEY` option that explicitly writes an empty list for sequence configuration parameters.

This is different from `--remove-key`, which removes the key entirely and may allow values from another configuration source or default values to take effect.

Closes #7617

### Checklist - did you ...

- [ ] ~~Add / update outdated documentation?~~

### Testing

- `python -m pytest tests/cli/test_main_config.py tests/cli/test_config.py -q`
- `python -m pre_commit run --files conda/cli/condarc.py conda/cli/main_config.py tests/cli/test_config.py tests/cli/test_main_config.py news/7617-config-clear`

### DIFF
--- a/conda/cli/condarc.py
+++ b/conda/cli/condarc.py
@@ -598,6 +598,67 @@ class ConfigurationFile:
         else:
             raise CondaKeyError(key, "invalid parameter")
 
+    def clear_key(self, key: str) -> None:
+        """
+        Clear all values from a sequence configuration parameter.
+
+        The key remains explicitly configured with an empty list. This differs
+        from ``remove_key()``, which removes the key entirely and may allow a
+        value from another configuration source or a default to take effect.
+
+        Args:
+            key: Sequence configuration key name.
+
+        Raises:
+            CondaKeyError: If the key is unknown or is not a sequence parameter.
+            CouldntParseError: If the configured value is not a sequence.
+        """
+        from ..exceptions import CondaKeyError, CouldntParseError
+
+        if not self.key_exists(key):
+            raise CondaKeyError(key, "unknown parameter")
+
+        if aliased := self.context.name_for_alias(key):
+            log.warning(
+                "Key %s is an alias of %s; clearing value with latter",
+                key,
+                aliased,
+            )
+            key = aliased
+
+        first, *rest = key.split(".")
+
+        if first == "plugins" and hasattr(self.context, "plugins"):
+            if not rest:
+                raise CondaKeyError(key, "invalid parameter")
+
+            base_context = self.context.plugins
+            base_config = self.content.setdefault("plugins", {})
+            parameter_name, *rest = rest
+        else:
+            base_context = self.context
+            base_config = self.content
+            parameter_name = first
+
+        try:
+            parameter_type = base_context.describe_parameter(parameter_name)[
+                "parameter_type"
+            ]
+        except KeyError:
+            raise CondaKeyError(key, "unknown parameter")
+
+        if parameter_type != "sequence" or rest:
+            raise CondaKeyError(key, "invalid parameter")
+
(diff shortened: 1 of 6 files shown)


conda/cli/main_config.py
```diff
@@ -211,6 +211,12 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         default=[],
         metavar=("KEY", "VALUE"),
     )
+    config_modifiers.add_argument(
+        "--clear",
+        action="append",
+        help="""Clear all values from a list key.""",
+        metavar="KEY",
```
danyeaw, line 219:
```suggestion
        help="""Clear all values from a list key.""",
        default=[],
        metavar="KEY",
```

This is a nit, but for consistency, the others pass in default.

conda/cli/condarc.py
```diff
@@ -598,6 +598,67 @@ def remove_item(self, key: str, item: Any) -> None:
...
+        if not self.key_exists(key):
+            raise CondaKeyError(key, "unknown parameter")
+
+        if aliased := self.context.name_for_alias(key):
+            log.warning(
+                "Key %s is an alias of %s; clearing value with latter",
+                key,
+                aliased,
+            )
+            key = aliased
```
danyeaw, line 629: This alias code is not clearing the original configuration for me, so you end up having the original config key and the cleared aliased one. This would also be worth a test.

conda/cli/condarc.py
```diff
@@ -598,6 +598,67 @@ def remove_item(self, key: str, item: Any) -> None:
...
+
+        first, *rest = key.split(".")
+
+        if first == "plugins" and hasattr(self.context, "plugins"):
+            if not rest:
+                raise CondaKeyError(key, "invalid parameter")
```
danyeaw: I think the `unknown parameter` error above already catches this, since `key_exists` requires the plugin's `rest` to already be greater than 0. So this is then a dead code branch.

conda/cli/main_config.py
danyeaw:
```suggestion
        default=[],
        metavar="KEY",
```
Nit, for consistency.

danyeaw (commented):
Hi @codewithdaniel1, thanks again for the contribution! This looks really nice and will be a good feature to have in our config subcommand.

One thing I thought about is we recently added an ErrorGuidance object to our CondaError to provide the user with more actionable feedback on errors. It doesn't look like we have it implemented for CondaParseErrors or CondaKeyErrors though, so I think this should be out of scope for this PR.

Just a few small comments, could we please also add a small test for `plugins.<sequence>` clearing? Thanks!

conda/cli/main_config.py
codewithdaniel1: Added default=[] for consistency

conda/cli/condarc.py
codewithdaniel1: Fixed this by removing the original alias entry when writing the cleared canonical key, and added a regression test for the alias case

conda/cli/condarc.py
codewithdaniel1: I removed the unreachable branch for this.

codewithdaniel1:
Thanks for the review @danyeaw ! I addressed the requested changes: added default=[], fixed alias clearing with a regression test, removed the unreachable plugin branch, and added coverage for clearing plugins<sequence> settings. I kept ErrorGuidance out of scope as suggested.

danyeaw (approved):
Thanks @codewithdaniel1!

PR #16510, opened 2026-08-10
Fix channel name and URL matching in `channel_settings`

travishathaway:
### Description

Fix `channel_settings` request matching for named channels, exact channel URLs, and same-scheme URL glob patterns. This supports the channel matching work in [conda-auth#74](https://github.com/conda-incubator/conda-auth/pull/74).

Exact URL entries are compared using the request and configured URL scheme, hostname, port, and channel path before selecting an authentication handler. The comparison keeps migrated source and destination URLs and encoded path separators distinct while continuing to ignore conda tokens, platform subdirectories, and package filenames. Named channel matching, trailing slash normalization, missing or `None` channel entries, and last-match-wins behavior remain supported.

URL glob entries remain on the raw same-scheme matcher. They are classified before conda URL normalization so recognized package and repodata suffixes inside a pattern cannot be stripped and widen authentication-handler selection.

### Checklist - did you ...

- [x] ~If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?~


jezdez:
I pushed b48d93826 directly to the branch, building on the concrete URL matching change in 9e4f0c607. Exact URL entries compare the request and configured URL components before selecting an authentication handler. URL glob entries now stay on the raw same-scheme `fnmatch` path.

Why this changed:

- `Channel.canonical_name` represents a logical channel name, so requests with different schemes, hosts, ports, or channel paths can resolve to the same value.
- Exact URL entries in `channel_settings` need to remain associated with the configured scheme, hostname, port, and channel path before conda selects an authentication handler.
- Comparing `Channel.base_url` is not sufficient because channel parsing rewrites migrated channel URLs and decodes encoded path separators before comparison.
- `split_conda_url_easy_parts` preserves the concrete URL components needed for exact matching while continuing to ignore conda tokens, platform subdirectories, and package filenames.
- URL glob patterns must be classified before conda URL normalization because recognized package and repodata suffixes inside a pattern can otherwise be stripped as filenames.
- If a URL glob enters the exact-match path, raw nonmatching requests can normalize to the same channel path and select an authentication handler outside the configured pattern.
- Keeping URL globs on the existing raw matcher preserves the identical-scheme requirement and the configured `fnmatch` behavior.
- The release tooling on current `main` now reads fragments from `releases/news`, so the fragment moved there.

The tests use real URL resolution and cover differing hosts, schemes, ports, channel paths, migrated URLs, encoded path separators, recognized-extension glob matches and rejections, exact IPv6 URLs, and last-match behavior. The inline comments in `get_session` explain why glob classification happens before `split_conda_url_easy_parts` and why the raw same-scheme matcher remains separate. The PR description and checklist now reflect the final behavior and current release directories.

jezdez (approved):

Issue #16395, opened 2026-07-10
Extract the classic solver into its own repo (conda-pycosat-solver)


kenodegard:
### Checklist

- [x] I added a descriptive title
- [x] I searched open issues and couldn't find a duplicate

### What?

Move the classic solver out of `conda/conda` and into its own standalone, externally installable plugin repository, following the same pattern already used by [`conda/conda-libmamba-solver`](https://github.com/conda/conda-libmamba-solver) and [`conda/conda-rattler-solver`](https://github.com/conda/conda-rattler-solver).

The repo conda-classic-solver already exists for this purpose but we need to rename the package to `conda-pycosat-solver` (it has never been released, so this needs no deprecation), keeping `classic` registered as an alias so `CONDA_SOLVER=classic` and existing configs keep working. The in-tree modules being removed from `conda/conda` (`solve.py`, `resolve.py`, `logic.py`, `_logic.py`) do go through a standard CEP-9 deprecation cycle, since they're already shipped as part of conda core.

This is really two separate tracks that need to converge: bringing `conda/conda` to a state ready to depend on the external package (deprecation, module addenda, dropping the in-tree registration), and separately bringing `conda-classic-solver` itself up to date, since it was cut from an old snapshot and breaks against modern conda (for example, `Spinner` no longer exists, and it references stale index APIs).

### Why?

The classic solver is roughly 4,200 lines of code plus several thousand lines of dedicated tests. Keeping it in-tree adds CI overhead to every conda core change, even changes unrelated to solving. The plugin infrastructure (`conda_solvers` hook, `CondaSolver` type) already supports this exact pattern, and `conda-libmamba-solver` and `conda-rattler-solver` already prove it out.

Discovery https://github.com/conda/conda/issues/16185 narrowed the plan:

- [`conda/conda-classic-solver`](https://github.com/conda/conda-classic-solver) already exists, but it's frozen in time (created about two years ago).
- [`conda/conda`'s `remove-classic` branch](https://github.com/conda/conda/tree/remove-classic) is stale but not unusable: despite being hundreds of commits behind `main`, the actual delta against core is small, about 14 files and 170 lines. Some of that prep work has already been extracted and landed as https://github.com/conda/conda/pull/16312 and https://github.com/conda/conda/pull/16320.
[...]

PR #16520, opened 2026-08-11
Move the `classic` solver to the `conda-classic-solver` plugin

### LINKED ISSUE #16569: Finish extracting conda-classic-solver
- https://github.com/conda/conda/pull/14167
- https://github.com/conda/conda/pull/14170
- https://github.com/conda/conda/pull/16520
- https://github.com/conda/conda-classic-solver/pull/29

kenodegard:
### Description

Reapplies #14170 to current `HEAD`. Supersedes #16519.

Removes the built-in `classic` solver from conda's own plugin registration (`get_plugin_manager()` no longer loads `conda.plugins.solvers`). `--solver=classic` (and its `pycosat` alias) continue to work unchanged — they're now provided by the standalone [`conda-classic-solver`](https://github.com/conda/conda-classic-solver) plugin, which is bundled with conda by default, so this is not a user-facing behavior change.

`conda.plugins.solvers`, `conda.resolve`, `conda.common.logic`, and `conda.common._logic` are marked pending deprecation (removed in 27.9) for anyone depending on the classic solver's internals directly — they should depend on `conda-classic-solver` instead. `conda.core.solve.Solver` is untouched, since it's shared base infrastructure that `conda-libmamba-solver` also builds on, not classic-specific.

Test suite updates account for classic no longer shipping by default (installing `conda-classic-solver` where needed, dropping hardcoded `--solver=classic` where it wasn't actually testing solver-specific behavior), plus doc notes pointing readers to the new plugin location.

Closes #14170
Closes #16519
Closes https://github.com/conda/conda/issues/16569

### Checklist - did you ...


travishathaway (approved):

conda/exports.py
```diff
@@ -57,19 +58,50 @@
...
+    "NoPackagesFound",
+    ResolvePackageNotFound,
+    addendum="Use `conda.exports.ResolvePackageNotFound` instead.",
+)
+deprecated.constant(
+    "27.3",
+    "27.9",
+    "NoPackagesFoundError",
+    ResolvePackageNotFound,
+    addendum="Use `conda.exports.ResolvePackageNotFound` instead.",
```
jezdez: Could these point directly to `conda.exceptions.ResolvePackageNotFound`? That gives all three names a supported replacement and avoids `ResolvePackageNotFound` pointing back to itself.

```suggestion
    addendum="Use `conda.exceptions.ResolvePackageNotFound` instead.",
)
deprecated.constant(
    "27.3",
    "27.9",
    "NoPackagesFound",
    ResolvePackageNotFound,
    addendum="Use `conda.exceptions.ResolvePackageNotFound` instead.",
)
deprecated.constant(
    "27.3",
    "27.9",
    "NoPackagesFoundError",
    ResolvePackageNotFound,
    addendum="Use `conda.exceptions.ResolvePackageNotFound` instead.",
```

conda/exports.py
```diff
@@ -264,6 +296,20 @@
...
+    UnsatisfiableError,
+    addendum="Use `conda.exports.UnsatisfiableError` instead.",
+)
+deprecated.constant(
+    "27.3",
+    "27.9",
+    "Unsatisfiable",
+    UnsatisfiableError,
+    addendum="Use `conda.exports.UnsatisfiableError` instead.",
+)
```
jezdez: Could we keep the existing `UnsatisfiableError` registration above and add just the alias here? The second registration replaces its existing warning, changing the advertised removal from 27.3 to 27.9 and pointing back to itself.

```suggestion
deprecated.constant(
    "27.3",
    "27.9",
    "Unsatisfiable",
    UnsatisfiableError,
    addendum="Use `conda.exceptions.UnsatisfiableError` instead.",
)
```

conda/exceptions.py
```diff
@@ -74,7 +74,20 @@ def __init__(self, bad_deps: Iterable[Iterable[MatchSpec]]):
...
+    "NoPackagesFound",
+    ResolvePackageNotFound,
+    addendum="Use `conda.exports.ResolvePackageNotFound` instead.",
+)
+deprecated.constant(
+    "27.3",
+    "27.9",
+    "NoPackagesFoundError",
+    ResolvePackageNotFound,
+    addendum="Use `conda.exports.ResolvePackageNotFound` instead.",
```
jezdez: Same replacement here, so following these warnings doesn't lead to another deprecated import.

```suggestion
    addendum="Use `conda.exceptions.ResolvePackageNotFound` instead.",
)
deprecated.constant(
    "27.3",
    "27.9",
    "NoPackagesFoundError",
    ResolvePackageNotFound,
    addendum="Use `conda.exceptions.ResolvePackageNotFound` instead.",
```

conda/plugins/manager.py
```diff
@@ -1435,7 +1434,6 @@ def get_plugin_manager() -> CondaPluginManager:
     plugin_manager = CondaPluginManager()
     plugin_manager.add_hookspecs(CondaSpecs)
     plugin_manager.load_plugins(
```
jezdez, line 1436: How would you like `--no-plugins` recovery to work after this move? `disable_external_plugins()` also disables the extracted solver, so both `--no-plugins --solver=classic` and `CONDA_NO_PLUGINS=true` leave no solver available. I checked the base and this head: the base retains `classic`, and this head has an empty solver registry after disabling plugins.

This seems worth deciding here, together with updating the recovery guidance. We can keep the implementation in its own repo.

jezdez (commented):
Ken, I've opened [conda-classic-solver#71](https://github.com/conda/conda-classic-solver/pull/71) to carry over these changes from conda:

- Global, channel, and package `--exclude-newer` policies, including keeping the policy captured when the solver is created: [#15761](https://github.com/conda/conda/pull/15761), merged 21 August 2026.
- Keeping supplied indexes lazy, preserving supplied records, and reducing the index again when specs change: [#16635](https://github.com/conda/conda/pull/16635), merged 8 September 2026.
- Reporting unavailable pinned versions as channel errors: [#15653](https://github.com/conda/conda/pull/15653), merged 31 March 2026. That PR added the distinction between packages missing from channels and packages missing from a prefix.

Those fixes should be in the solver package before we switch `classic` over here.

I've left a few small suggestions for the deprecation messages below, plus a question about recovery with plugins disabled.

We should also make sure the fixed solver package is available to the recipe's build channels. The tests currently fetch it from `conda-canary/label/dev`, while the recipe uses an unqualified dependency.

jezdez:
Ken, I've [added the naming and release proposal to #16395](https://github.com/conda/conda/issues/16395#issuecomment-5761790122), building on the `conda-pycosat-solver` rename you outlined there. I'd like to finish that before this merges, keeping both `classic` and `pycosat` supported and publishing the renamed package with [conda-classic-solver#71's fixes](https://github.com/conda/conda-classic-solver/pull/71).

conda/plugins/manager.py
kenodegard: @jezdez Agreed, and I'd generalize this rather than special-case solvers: many conda subcommands don't need a solver at all.

Besides even for solver subcommands there are actions that don't perform a solve:
- Explicit installs (`conda install URI`) already skip the solver entirely
- Force removes (`conda remove --force`) doesn't skip the solver today but there's no reason we couldn't support that in the future

From my perspective this reinforces wanting a more generic allowlist system rather than a solver-specific carve-out.

Your #15915 work already does the inverse of what we need. So just a matter of adding the ability to *re-enable* under a blanket `--no-plugins`, e.g. `--no-plugins --plugins conda-pycosat-solver`.

IOW sounds like something to address in #15915.

conda/exports.py
kenodegard: https://github.com/conda/conda/pull/16520/commits/2f766da9814d9062e9efaeb64a24f68d49a28164

conda/exports.py
kenodegard: https://github.com/conda/conda/pull/16520/commits/2f766da9814d9062e9efaeb64a24f68d49a28164

conda/exceptions.py
kenodegard: https://github.com/conda/conda/pull/16520/commits/2f766da9814d9062e9efaeb64a24f68d49a28164

PR #16523, opened 2026-08-12
Add Windows ARM64 CI coverage

### LINKED ISSUE #16466: Add Windows ARM64 to the CI test matrix
### What is the idea?

Add native Windows ARM64 (`win-arm64`) coverage to conda's GitHub Actions test matrix using the `windows-11-arm` hosted runner, alongside the existing `windows-latest` (`win-64`) jobs.

### Why is this needed?

conda already runs Linux aarch64 and macOS arm64 in CI, but Windows testing is still x86_64-only. Platform detection, activation, entry points, and package install behavior can diverge on Windows ARM64, and several of those paths recently changed or are still landing (for example native launcher support). Without CI on real ARM64 Windows hardware, regressions stay invisible until users hit them. GitHub now provides free `windows-11-arm` runners for public repositories, so the previous CI hardware gap is no longer a hard blocker.

### What should happen?

- Extend the Windows job in the tests workflow so it can run on `windows-11-arm` in addition to `windows-latest`.
- Start with a reduced matrix slice (similar to how Linux aarch64 and macOS arm64 are constrained today): one current Python, one default channel, and a limited unit/integration split—not a full Cartesian expansion of the win-64 matrix.
- Get the suite green on that slice: fix real `win-arm64` bugs, or skip/xfail only where the ecosystem is not ready yet (with a clear reason).
- Document any bootstrap constraints in the workflow or contributor docs if the initial setup cannot use a native win-arm64 Miniconda/Miniforge installer yet.
- Leave canary/package builds for `win-arm64` out of scope unless the reduced test slice is already stable.

### Additional Context

- Broader Windows on Arm tracking: https://github.com/conda/conda/issues/11472
- Native win-arm64 launchers / conda-launchers dependency work: https://github.com/conda/conda/issues/16293
- Runner label: `windows-11-arm` (GitHub-hosted Windows 11 ARM64)
- Ecosystem note: conda-forge's win-arm64 support is still experimental, and there is not yet a native Miniforge/Miniconda win-arm64 installer; initial CI may need a documented bootstrap path until that exists.
- Related packaging context: https://conda-forge.org/blog/2026/02/09/win-arm64/

jezdez:
### Description

Adds Windows ARM64 runner coverage for #16466.

The existing Windows test job now runs five jobs on `windows-11-arm`. They use x64 Miniconda and win-64 packages under Windows emulation, while keeping the existing dependency files and libmamba test configuration.

This does not test native win-arm64 Python or packages. The existing requirements cannot currently be solved for win-arm64 because conda, libmambapy, and several test dependencies are unavailable. Native testing remains follow-up work for #16466.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,7 +94,7 @@ jobs:
             default-channel: conda-forge
     env:
       ErrorActionPreference: Stop  # powershell exit immediately on error
-      TEMP_DRIVE: ${{ matrix.runs-on == 'windows-11-arm' && 'C' || 'D' }}
+      TEMP_DRIVE: C
       PYTEST_MARKER: ${{ matrix.test-type == 'unit' && 'not integration and not benchmark' || 'integration and not benchmark' }}
       PYTEST_SPLITS: ${{ matrix.test-type == 'unit' && '2' || '3' }}
       CONDA_TEST_SOLVERS: ${{ github.event_name == 'pull_request' && 'libmamba' || 'libmamba,classic' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,6 +94,7 @@ jobs:
             default-channel: conda-forge
     env:
       ErrorActionPreference: Stop  # powershell exit immediately on error
+      TEMP_DRIVE: ${{ matrix.runs-on == 'windows-11-arm' && 'C' || 'D' }}
       PYTEST_MARKER: ${{ matrix.test-type == 'unit' && 'not integration and not benchmark' || 'integration and not benchmark' }}
       PYTEST_SPLITS: ${{ matrix.test-type == 'unit' && '2' || '3' }}
       CONDA_TEST_SOLVERS: ${{ github.event_name == 'pull_request' && 'libmamba' || 'libmamba,classic' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -99,11 +104,12 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set TEMP to D:/Temp
+      - name: Set TEMP
         run: |
-          mkdir "D:\\Temp"
-          echo "TEMP=D:\\Temp" >> $env:GITHUB_ENV
-          echo "TMP=D:\\Temp"  >> $env:GITHUB_ENV
+          $TEMP_DIR = "${{ runner.arch == 'ARM64' && 'C' || 'D' }}:\\Temp"
```
kenodegard: Let's define `TEMP_DRIVE` at the job level?

jezdez: Done in bd7d1eaab. `TEMP_DRIVE` is now defined once and reused for TEMP, the package cache, and Miniconda. The job-level expression uses `matrix.runs-on` because `runner.arch` is unavailable in job-level `env`.

.github/workflows/tests.yml
```diff
@@ -87,8 +88,13 @@ jobs:
             python-version: '3.10'
           - test-type: unit
             test-group: 3
+          - runs-on: windows-11-arm
+            python-version: '3.10'
+          - runs-on: windows-11-arm
+            default-channel: conda-forge
     env:
       ErrorActionPreference: Stop  # powershell exit immediately on error
+      TEMP_DRIVE: C
```
kenodegard: If its just the `C:` drive should we just remove all of these custom paths? Just use `runner.temp` where we need a TEMP?

jezdez: Done in dfa211df2. I removed `TEMP_DRIVE`, the TEMP setup step, and the setup-miniconda path overrides. The package cache now uses setup-miniconda’s default `~/conda_pkgs_dir`, and pytest already uses `runner.temp` for its temporary directory.

kenodegard (approved):

PR #16538, opened 2026-08-13
Use constants for "json", "shards", "zst" format keys in cache state

ForgottenProgramme:
### Description

PR based off of Daniel's https://github.com/conda/conda/commit/733fdd54bc4ab8539e45f0eac202cdbad652bceb 

### Checklist - did you ...

### DIFF
--- a/tests/gateways/test_repodata_gateway.py
+++ b/tests/gateways/test_repodata_gateway.py
@@ -35,7 +35,6 @@ from conda.gateways.repodata import (
     ETAG_KEY,
     FORMAT_JSON,
     FORMAT_SHARDS,
-    FORMAT_ZST,
     CondaRepoInterface,
     RepodataCache,
     RepodataFetch,
--- a/tests/gateways/test_repodata_gateway.py
+++ b/tests/gateways/test_repodata_gateway.py
@@ -462,7 +462,7 @@ def test_get_cache_control_max_age():
 
 
 def test_classic_soft_404_records_no_repodata_json(tmp_path, mocker):
-    """Record has_repodata_json False when classic 404s and shards are known."""
+    """Record has_json False when classic 404s and shards are known."""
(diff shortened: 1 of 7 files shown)


ForgottenProgramme:
pre-commit.ci autofix

tests/gateways/test_repodata_gateway.py
```diff
@@ -473,11 +475,11 @@ def test_classic_soft_404_records_no_repodata_json(tmp_path, mocker):
         REPODATA_FN,
         repo_interface_cls=CondaRepoInterface,
     )
-    fetch.repo_cache.state.set_has_format("shards", True)
+    fetch.repo_cache.state.set_has_format(FORMAT_SHARDS, True)
     fetch.repo_cache.save(b"shards")
     assert not fetch.repo_cache.cache_path_json.exists()
     fetch.fetch_latest()
-    assert fetch.repo_cache.load_state().has_format("repodata_json")[0] is False
+    assert fetch.repo_cache.load_state().has_format(FORMAT_JSON)[0] is False
```
jezdez: Nice follow-up, Mahe. Could we load the state into a variable and also assert that `"has_json" in state` and `"has_repodata_json" not in state` here? Production and the current assertion both use `FORMAT_JSON`, so they would still agree if the constant itself had the wrong value. This is just a small test-hardening suggestion.

jezdez (approved):
Looks good to me. This fixes the cache-state key consistently. Thanks for following up so quickly, Mahe.

tests/gateways/test_repodata_gateway.py
ForgottenProgramme: Thanks Jannis. 
Since this itself is the PR where we define that constant, I don't think we need to worry about it being set to a wrong value. 
But I still added your suggested asserts because it doesn't hurt to have that.

Issue #16552, opened 2026-08-17
conda 26.7 process-pool extraction crashes under pytest `CliRunner` (`BrokenProcessPool`)


M-Waszkiewicz-Anaconda:
## Summary

conda 26.7.0 package extraction uses `ProcessPoolExecutor(..., mp_context="spawn")`. That pool is created successfully on a normal host, then dies when conda is invoked **in-process** from pytest via Click `CliRunner`.

This is a separate failure from #16512. #16513 falls back to threads only when **constructing** the process pool raises `OSError` / `NotImplementedError` (e.g. missing `/dev/shm` on AWS Lambda). It does not cover this case: construction succeeds, workers crash later.

## Expected result

`conda-build` render / package fetch+extract completes when driven from a Click command under pytest `CliRunner`, as it did with conda 26.5.x (thread pool).

## Actual result

```text
BrokenProcessPool: A process in the process pool was terminated abruptly
while the future was running or pending.
```

Call path:

```text
conda_build/render.py            execute_download_actions → pfe.execute()
conda/core/package_cache_data.py ProcessPoolExecutor(mp_context="spawn")
                                 extract_executor.submit(...)
concurrent.futures.process       BrokenProcessPool
```

## Why this happens

`click.testing.CliRunner.invoke()` runs the Click command in the **same Python process** as pytest. conda 26.7 then spawns extract workers from that process.

Those workers re-import the pytest `__main__` module and exit. The parent sees `BrokenProcessPool`. Recipes that do not download/extract packages during render often pass; recipes that do hit this path fail.

#16513 does not help here:

```python
try:
    extract_executor = ProcessPoolExecutor(
        max_workers=EXTRACT_PROCESSES,
        mp_context=multiprocessing.get_context("spawn"),
    )
except (OSError, NotImplementedError):
    use_process_pool = False
```

On a normal Linux/macOS host the constructor does not raise. `BrokenProcessPool` is raised later, when submitted work runs, and is not caught.

## Workaround

`CONDA_DEBUG=1` forces the existing thread-pool extract path (`context.debug` disables `use_process_pool`).

## Suggested fix

Treat worker-pool failure the same as construction failure: if process-pool extraction is unavailable **or** the pool breaks, fall back to `ThreadPoolExecutor`. Catching `BrokenProcessPool` around submit/result (or probing spawn in a throwaway worker before using the pool) would cover in-process test runners as well as Lambda.

[...]

PR #16555, opened 2026-08-17
Report package extraction errors from process workers

jezdez:
### Description

Addresses #16552.

Some package extraction exceptions can be serialized by a spawned worker but cannot be reconstructed in the parent process. `ProcessPoolExecutor` then reports `BrokenProcessPool`, hiding the original extraction error.

Check extraction exceptions inside the worker and replace only those that cannot cross the process boundary with a pickle-safe `RuntimeError` retaining the original qualified type and message. Direct and threaded extraction paths remain unchanged.

Also retain process-pool failure causes in debug logs so unrelated worker and result-channel failures remain diagnosable.

This is a defensive conda-side workaround. The root fix for the currently affected exception classes is conda/conda-package-handling#349.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/main/news/TEMPLATE)) for the next release?

### DIFF
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -18,7 +18,14 @@ def extract_conda_package_archive(
     *,
     ensure_picklable_errors: bool = False,
 ) -> None:
-    """Extract a conda package archive without importing conda runtime state."""
+    """Extract a conda package archive without importing conda runtime state.
+
+    Args:
+        source_full_path: Package archive to extract.
+        destination_directory: Directory to extract into.
+        ensure_picklable_errors: Replace exceptions that cannot survive a pickle
+            round trip with a plain ``RuntimeError``.
+    """
     import conda_package_handling.api
 
     try:
--- a/conda/_private/extract.py
+++ b/conda/_private/extract.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import os
+import pickle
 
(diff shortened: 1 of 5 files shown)


conda/_private/extract.py
```diff
@@ -14,11 +15,24 @@
 def extract_conda_package_archive(
     source_full_path: str | os.PathLike,
     destination_directory: str | os.PathLike,
+    *,
+    ensure_picklable_errors: bool = False,
```
ryanskeith, line 19: A quick comment in the docstring would be nice here to explain the purpose of the new variable.  I know right now it makes sense but I could see the intention taking future dev's a bit to understand.  Reading the exception clause below made me think for a bit.  Perhaps just a comment in right before the raise would also work well.

ryanskeith (approved):
LGTM, as a workaround.  If we feel this will be temporary, we should add a dev note.  Also, my suggestion for a docstring documentation for the the new keyword argument would be nice but not necessary.

conda/_private/extract.py
jezdez: Documented the keyword argument in f3806a5d6. I plan to keep this as a defensive process-boundary check even after conda-package-handling fixes the currently known exceptions, so I left out a temporary-removal note.

Issue #16559, opened 2026-08-18
Scheduled Test Failure Tracker


jezdez:
### Checklist

- [x] I added a descriptive title
- [x] I searched open issues and found no duplicate

### What?

Use this issue as the single inbox for failures from the scheduled Tests workflow. Each failed run attempt will be added as a comment.

### Why?

The current automation creates a dated issue for each failure day. Retries crossing midnight and unrelated failures produce duplicate reports that require manual cleanup.

### User impact

Maintainers get one durable place to review scheduled test failures and can create focused issues only for confirmed regressions.

### Goals

- [ ] Review new scheduled failure comments
- [ ] Open focused issues for reproducible conda regressions
- [ ] Keep transient infrastructure failures as comments rather than separate issues

### Tasks

- [ ] Update the Tests workflow to report failures here

PR #16560, opened 2026-08-18
Track scheduled test failures in one issue

jezdez:
### Description

Send every scheduled Tests workflow failure to the permanent tracker #16559 instead of creating a dated issue. Each comment links to the exact run attempt, so reruns remain distinguishable without multiplying issues.

### Checklist - did you ...

- [x] ~Add a file to the news directory for the next release notes?~
- [x] ~Add or update necessary tests?~
- [x] ~Add or update outdated documentation?~

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -850,30 +850,14 @@ jobs:
           && steps.alls-green.outputs.result == 'failure'
         env:
           GH_TOKEN: ${{ secrets.AUTO_REPORT_TEST_FAILURE }}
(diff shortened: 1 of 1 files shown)


kenodegard (approved):

kenodegard (changes requested):
This actually means the test failure issue should never be closed

We can remove the timestamp from the title and append all failures to the "current" issue but then allow said issue to be closed when improvements have been made

Ultimately if these reporting issues are too noisy then...we should just disable them entirely

PR #16565, opened 2026-08-19
Require confirmation to replace environments with `conda env create`

### LINKED ISSUE #15071: `conda env create` silently replaces an existing environment without explicit `--yes`
### Checklist

### What happened?

In versions of conda older than 25.5.0, `conda env create` would produce an error when users try to create a new environment in a prefix that already exists. This was consistent with the behaviour of `conda create`. Starting in conda 25.5.0, `conda env create` will not produce the error.

For example, consider a case where conda has an existing environement called "existing". It was created running `conda create -n existing python`.

**Correct behaviour in conda 25.3.1**

`conda create` asks for the user to confirm that they want to remove the environment
```
$ conda create -n existing python                                                                                                                                                              
WARNING: A conda environment already exists at '/home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/existing'

Remove existing environment?
This will remove ALL directories contained within this specified prefix directory, including any other conda environments.

 (y/[n])? 
```

`conda env create` errors
```
$ conda env create -n existing --file tests/env/support/env_with_dependencies.yml                                                                                                              

CondaValueError: prefix already exists: /home/sophia/projects/conda/devenv/Linux/x86_64/envs/devenv-3.10-c/envs/existing
```

**Broken behaviour in conda 25.5.0**

`conda env create` will will install the environment into the existing prefix
```
$ conda env create -n existing --file tests/env/support/env_with_dependencies.yml                                                                                                              
Channels:
 - conda-forge
 - defaults
 - 
Platform: linux-64
. . .
```

jezdez:
### Description

Require `conda env create` to reject an existing environment unless replacement is authorized by `--yes` or `always_yes`. This preserves the existing prefix instead of silently deleting and recreating it. When replacement is authorized, report that the existing environment is being removed.

Closes #15071.

### Checklist - did you ...

- [x] Add a file to the `news` directory for the next release's release notes?
- [x] ~~Add / update outdated documentation?~~ Not applicable.

### DIFF
--- a/conda/cli/main_create.py
+++ b/conda/cli/main_create.py
@@ -123,6 +123,7 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
     from ..core.prefix_data import PrefixData
     from ..exceptions import ArgumentError, CondaValueError, TooManyArgumentsError
     from ..gateways.disk.delete import rm_rf
+    from ..gateways.streams import stdoutlog
     from ..reporters import confirm_yn
     from .common import (
         get_name_prefix_from_env_file,
@@ -132,6 +133,8 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
         validate_subdir_config,
     )
     from .install import install, install_clone
+
+    env_create = getattr(args, "env_create", False)
 
     # When `--clone` is present, `--file` and `packages` are not allowed
     if args.clone:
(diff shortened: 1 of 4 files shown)


conda/cli/main_create.py
```diff
@@ -187,6 +189,14 @@ def execute(args: Namespace, parser: ArgumentParser) -> int:
             raise CondaValueError(
                 "Cannot `create --dry-run` with an existing conda environment"
             )
+        if env_create and not args.yes:
```
ryanskeith: I noticed that the code is ignoring the `always_yes` config option and even in testing the code making sure that explicit use is required.  

This seems misaligned in this case as conda has other operations that do accept always_yes as a valid --yes (the documentation also says as much). `conda create` accepts always_yes as valid which is very close to this operation.

I would suggest the code accepts always_yes as a vaild --yes for documentation

ryanskeith (commented):
The code changes are fine and well done. I would suggest using `aways_yes` as a valid `--yes` and also adding a message to the user that an environment was overriden.

conda/cli/main_create.py
jezdez: Updated the check to use `context.always_yes`, so both `--yes` and `always_yes` authorize the replacement. The regression coverage now checks rejection without confirmation, both authorization paths, and JSON output. I also added a normal-output message before the existing environment is removed.

ryanskeith (approved):
Thank you for the changes.

Issue #14142, opened 2024-08-09
Deprecate `--dev` from activation logic


kenodegard:
### Checklist

- [X] I added a descriptive title
- [X] I searched open issues and couldn't find a duplicate

### What?

A while back (2019, see https://github.com/conda/conda/pull/8435) `--dev` was added to both `conda activate` and `conda deactivate` (along with a number of other subcommands, see https://github.com/conda/conda/pull/8435/commits/711dacfe27b99efb87d731511830de2dc09fd237, https://github.com/conda/conda/pull/8435/commits/b82ec695c3d366201431306c02becd6fcebb0e34, https://github.com/conda/conda/pull/8435/commits/9a4589a1fabc66c122fb0c20fa1f3aa2c8b1758b, https://github.com/conda/conda/pull/8435/commits/dd6ad75076f177cfc51b33a06dcb275e033aecfd, and https://github.com/conda/conda/pull/8435/commits/40b48326da1a2a40127ef6846b5bbc26701eea3c) to correctly support testing the development version of conda (and not some miniconda installation used to bootstrap the testing framework).

I strongly believe this is something we should work to remove primarily since we shouldn't be shipping our testing infrastructure in a user accessible way. I also consider the fact that the `--dev` argument logic being necessary is rather a symptom of other issues with the activation framework which should be addressed instead.

### Why?

Several attempts to correct and refactor `conda.activate` have been undertaken over the years. Each time, confusion arises around the validity/purpose of the `--dev` logic, much of which prevents meaningful improvements (speaking from personal experience).

### User impact

Remove arguments from subcommands that end users shouldn't be using in the first place.

### Goals
- [ ] Simplified `conda.activate` code base
- [x] Improved tests for `conda.activate` & `conda.shell` (confidence we are testing the correct thing)
    - https://github.com/conda/conda/pull/14141
- [ ] Prepare for effort to breakup activation logic into plugins

### Tasks
- [x] stop using `--dev` in test suite
  - #16571
- [x] deprecate `--dev`/`context.dev` usage in `conda.activate`
  - #16571
- [x] deprecate `--dev`/`context.dev` usage in shell scripts (`conda.shell`)
  - #16571
- [x] deprecate `--dev`/`context.dev` from all subcommands
  - `create`/`install`/`remove` via #16571
  - `run` via #15641
- [x] deprecate `context.dev`
  - #16571

### This epic is blocked by:
n/a

### This epic blocks:
n/a

danyeaw:
[...]

PR #16571, opened 2026-08-19
Stop using `--dev` in tests and deprecate `--dev`/`context.dev`

### LINKED ISSUE #15696: Remove dev mode logic for activation scripts?
There's a lot of logic in the activation scripts about "dev" mode that I'm not sure why we want to maintain. It's mostly for debugging, but I don't think anyone has used it in a long time, and we can always patch things locally to investigate whatever issue. It creates extra complexity for little gain, so if someone wants to give it a try, I think it's good for removal.

_Originally posted by @jaimergp in https://github.com/conda/conda/issues/15641#issuecomment-3871849696_

kenodegard:
### Description

`--dev`, `context.dev`, and conda-exported `_CE_M` / `_CE_CONDA` are a 2019 test-harness protocol so a temp env can still run checkout conda as `python -m conda`. That is not how development works anymore: `dev/start` sets `PYTHONPATH` to the repo and uses the devenv `conda` launcher; pytest already has an autouse `PYTHONPATH` fixture; `conda_cli` no longer injects `--dev`. `conda run --dev` is already pending deprecation (#15641).

This PR switches tests and wrappers to that `PYTHONPATH` + real `CONDA_EXE` path, and pending-deprecates the remaining knobs. `--dev` behavior is unchanged for external callers during the comment period.

- Tests no longer pass `--dev` or rewrite `CONDA_EXE` to `python -m conda`. Shell init uses the real launcher and `PYTHONPATH`.
- `conda activate` / hook without `--dev` no longer force `context.dev = False`. That was rewriting `CONDA_EXE` to a pip stub on Windows during stacking/reactivate (#15696).
- Deprecate `conda activate --dev`, `create`/`install`/`remove`/`init --dev`, `context.dev`, and `wrap_subprocess_call(dev_mode)`.
- conda still *exports* `_CE_M=-m` / `_CE_CONDA=conda` only when `context.dev` is true. Shell expansion of `$_CE_M $_CE_CONDA` stays (empty is a no-op; conda-build still uses the slots).
- `pytest_report_header` now also asserts checkout `conda.sh` / `conda.bat` (not site-packages copies).

Xref #14142
Resolves #15696

### Checklist - did you ...

### DIFF
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -44,6 +44,7 @@ from .common.compat import on_win
 from .common.path import path_identity as _path_identity
 from .common.path import paths_equal, unix_path_to_win, win_path_to_unix
 from .common.serialize import json
+from .deprecations import deprecated
 from .exceptions import (
     ActivateHelp,
     ArgumentError,
@@ -289,10 +290,17 @@ class _Activator(metaclass=abc.ABCMeta):
             try:
                 dev_idx = remainder_args.index("--dev")
             except ValueError:
-                context.dev = False
+                pass
             else:
                 del remainder_args[dev_idx]
                 context.dev = True
+                deprecated.topic(
+                    "27.3",
+                    "27.9",
+                    topic=f"`conda {command} --dev`",
+                    addendum="Set `PYTHONPATH` to the conda source root instead.",
+                    deprecation_type=FutureWarning,
+                )
 
         if command == "activate":
             self.stack = context.auto_stack and context.shlvl <= context.auto_stack
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -293,7 +293,6 @@ class _Activator(metaclass=abc.ABCMeta):
                 pass
             else:
                 del remainder_args[dev_idx]
-                context.dev = True
                 deprecated.topic(
                     "27.3",
                     "27.9",
--- a/conda/activate.py
+++ b/conda/activate.py
@@ -293,6 +293,7 @@ class _Activator(metaclass=abc.ABCMeta):
                 pass
             else:
                 del remainder_args[dev_idx]
+                context.dev = True
                 deprecated.topic(
                     "27.3",
                     "27.9",
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -885,6 +885,12 @@ class Context(Configuration):
         None means unset it.
         """
         if context.dev:
+            deprecated.topic(
+                "27.3",
+                "27.9",
+                topic="`conda.base.context.Context.dev`",
+                addendum="Set `PYTHONPATH` to the conda source root instead.",
+            )
             if pythonpath := os.environ.get("PYTHONPATH", ""):
                 pythonpath = os.pathsep.join((CONDA_SOURCE_ROOT, pythonpath))
             else:
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -26,6 +26,7 @@ from typing import TYPE_CHECKING
 
 from frozendict import frozendict
 
+from .. import CONDA_SOURCE_ROOT
 from .. import __version__ as CONDA_VERSION
 from ..auxlib.decorators import memoizedproperty
 from ..auxlib.ish import dals
@@ -612,6 +613,11 @@ class Context(Configuration):
     def dev(self) -> bool:
         return self._dev
 
+    @dev.setter
+    def dev(self, value: bool) -> None:
+        self._dev = value
+        self._cache_.pop("_dev", None)
+
     @property
     @deprecated(
         "26.9",
@@ -892,6 +898,28 @@ class Context(Configuration):
         The vars can refer to each other if necessary since the dict is ordered.
         None means unset it.
         """
+        if self._dev:
+            deprecated.topic(
+                "27.3",
+                "27.9",
+                topic="`conda.base.context.Context.dev`",
+                addendum="Set `PYTHONPATH` to the conda source root instead.",
+            )
+            if pythonpath := os.environ.get("PYTHONPATH", ""):
+                pythonpath = os.pathsep.join((CONDA_SOURCE_ROOT, pythonpath))
+            else:
+                pythonpath = CONDA_SOURCE_ROOT
+            return {
+                "CONDA_EXE": sys.executable,
+                "_CONDA_EXE": sys.executable,
+                # do not confuse with os.path.join, we are joining paths with ; or : delimiters
+                "PYTHONPATH": pythonpath,
+                "_CE_M": "-m",
+                "_CE_CONDA": "conda",
+                "CONDA_PYTHON_EXE": sys.executable,
+                "_CONDA_ROOT": self.conda_prefix,
+            }
+
         exe = os.path.join(
             self.conda_prefix,
             BIN_DIRECTORY,
--- a/conda/base/context.py
+++ b/conda/base/context.py
@@ -26,7 +26,6 @@ from typing import TYPE_CHECKING
 
 from frozendict import frozendict
 
-from .. import CONDA_SOURCE_ROOT
 from .. import __version__ as CONDA_VERSION
 from ..auxlib.decorators import memoizedproperty
 from ..auxlib.ish import dals
@@ -454,7 +453,7 @@ class Context(Configuration):
     )
     _debug = ParameterLoader(PrimitiveParameter(False), aliases=["debug"])
     _trace = ParameterLoader(PrimitiveParameter(False), aliases=["trace"])
-    dev = ParameterLoader(PrimitiveParameter(False))
+    _dev = ParameterLoader(PrimitiveParameter(False), aliases=("dev",))
     dry_run = ParameterLoader(PrimitiveParameter(False))
     _error_upload_url = ParameterLoader(
         PrimitiveParameter("https://conda.io/conda-post/unexpected-error"),
@@ -603,6 +602,15 @@ class Context(Configuration):
         """
         self.plugin_manager.load_settings()
         return self.plugin_manager.get_config(self.raw_data)
+
+    @property
+    @deprecated(
+        "27.3",
+        "27.9",
+        addendum="Set `PYTHONPATH` to the conda source root instead.",
+    )
+    def dev(self) -> bool:
+        return self._dev
 
     @property
     @deprecated(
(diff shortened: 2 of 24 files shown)


conda/activate.py
```diff
@@ -289,10 +290,16 @@ def _parse_and_set_args(self) -> None:
...
             else:
                 del remainder_args[dev_idx]
-                context.dev = True
+                deprecated.topic(
+                    "27.3",
+                    "27.9",
+                    topic=f"`conda {command} --dev`",
+                    addendum="Set `PYTHONPATH` to the conda source root instead.",
+                    deprecation_type=FutureWarning,
+                )
```
jaimergp, line 303: Don't we need to retain the `context.dev` assignment logic until it's deprecated for good?

conda/base/context.py
```diff
@@ -604,6 +603,15 @@ def plugins(self) -> PluginConfig:
         self.plugin_manager.load_settings()
         return self.plugin_manager.get_config(self.raw_data)
 
+    @property
+    @deprecated(
+        "27.3",
+        "27.9",
+        addendum="Set `PYTHONPATH` to the conda source root instead.",
```
jaimergp, line 611: Would it help to print the source root path here? Can we even?

conda/base/context.py
```diff
@@ -884,35 +892,22 @@ def conda_exe_vars_dict(self) -> dict[str, str | None]:
...
+            "CONDA_EXE": exe,
+            "_CONDA_EXE": exe,
+            # Shell wrappers expand `"$CONDA_EXE" $_CE_M $_CE_CONDA` (`python -m conda`
+            # when set). None unsets leftovers; keep the keys while wrappers expand them.
+            # https://github.com/conda/conda/issues/14142
+            "_CE_M": None,
+            "_CE_CONDA": None,
+            "CONDA_PYTHON_EXE": sys.executable,
+            "_CONDA_ROOT": self.conda_prefix,
+        }
```
jaimergp, line 938: Same don't we need to retain this til removed?

conda/activate.py
kenodegard: Since `--dev` is a test suite setting production users shouldn't be using it so it is probably safe to simply remove but yes I've effectively skipped the deprecation process and we're simply telling users what to do instead. I can add it back and this PR becomes focused on removing `--dev` from the tests.

conda/base/context.py
kenodegard: We can, idk if we should

agriyakhetarpal (changes requested):
Most of the changes here are mechanical and make sense to me – I can't see any reason why this should be a problem and IMHO it is alright to go forward with both removing both `--dev` usages in tests and the deprecations in the same PR.

I ran this locally and found a few additional things:
- If you set the `CONDA_DEV` environment variable (or, by extension, put `dev: true` in `.condarc` – I doubt anyone ever does that, but theoretically it is a valid configuration value), no deprecation warning is emitted because they are currently only on the CLI flag and on using `context.dev` programmatically.
- I tested `conda run --dev`, which I previously worked on, and it now emits warnings twice, which I found to be a bit non-intuitive:
    ```bash
  python -W always::PendingDeprecationWarning -m conda run --dev echo hi
    ```
    returns the following:
    ```
	/Users/agriyakhetarpal/Desktop/conda/conda/cli/conda_argparse.py:255: PendingDeprecationWarning: `--dev` is pending deprecation and will be removed in 27.3. Set `PYTHONPATH` to the conda source root instead.
	  super().__call__(parser, namespace, values, option_string)
	/Users/agriyakhetarpal/Desktop/conda/conda/utils.py:189: PendingDeprecationWarning: `conda.utils.wrap_subprocess_call(dev_mode)` is pending deprecation and will be removed in 27.9. Set `PYTHONPATH` to the conda source root instead.
	  deprecated.topic(
	hi
    ```
- Lastly, what do we do about `conda init --dev`? It is deprecated now, but I still see two uses of it, one in the Dev Containers config (which I've used to some effect before), and I found another in Xonsh shell config: https://github.com/conda/conda/blob/c33dcb27cf10a0981d041ac6bfd102413fc3465c/conda/shell/conda.xsh#L15-L16

conda/activate.py
kenodegard: Restored in https://github.com/conda/conda/pull/16571/commits/37b4345dc7d94c4f469f4ac9e7d0076f657a9bb3

conda/base/context.py
kenodegard: Restored in https://github.com/conda/conda/pull/16571/commits/37b4345dc7d94c4f469f4ac9e7d0076f657a9bb3

kenodegard:
@agriyakhetarpal thanks for taking a look!

> - If you set the `CONDA_DEV` environment variable (or, by extension, put `dev: true` in `.condarc` – I doubt anyone ever does that, but theoretically it is a valid configuration value), no deprecation warning is emitted because they are currently only on the CLI flag and on using `context.dev` programmatically.

Making `--dev` / `CONDA_DEV` a complete no-op was me being overzealous, behavior restored. Fixed `conda_exe_vars_dict` so it properly warns when `--dev` / `CONDA_DEV` is set.

Once `context.dev` is removed, leftover `dev: true` in `.condarc` is just another unknown key and conda already ignores those. We are not adding a separate “you still have this config defined” warning.

> - I tested `conda run --dev`, which I previously worked on, and it now emits warnings twice, which I found to be a bit non-intuitive:
    ```bash
  python -W always::PendingDeprecationWarning -m conda run --dev echo hi
    ```
    returns the following:
    ```
	/Users/agriyakhetarpal/Desktop/conda/conda/cli/conda_argparse.py:255: PendingDeprecationWarning: `--dev` is pending deprecation and will be removed in 27.3. Set `PYTHONPATH` to the conda source root instead.
	  super().__call__(parser, namespace, values, option_string)
	/Users/agriyakhetarpal/Desktop/conda/conda/utils.py:189: PendingDeprecationWarning: `conda.utils.wrap_subprocess_call(dev_mode)` is pending deprecation and will be removed in 27.9. Set `PYTHONPATH` to the conda source root instead.
	  deprecated.topic(
	hi
    ```

The two `conda run --dev` warnings are intentional and are different signals. `--dev` is the user-facing flag (and gets a user-facing `FutureWarning` once no longer pending). `wrap_subprocess_call(dev_mode)` is the library API (and gets a dev-facing `DeprecationWarning` once no longer pending). Both are pending today, so they appear as duplicates.

> - Lastly, what do we do about `conda init --dev`? It is deprecated now, but I still see two uses of it, one in the Dev Containers config (which I've used to some effect before), and I found another in Xonsh shell config: https://github.com/conda/conda/blob/c33dcb27cf10a0981d041ac6bfd102413fc3465c/conda/shell/conda.xsh#L15-L16

This got caught up in a broader find/replace. Reverted the `conda init --dev` deprecation.

conda/base/context.py
```diff
@@ -604,6 +604,20 @@ def plugins(self) -> PluginConfig:
...
+        "27.9",
+        addendum="Set `PYTHONPATH` to the conda source root instead.",
+    )
+    def dev(self) -> bool:
+        return self._dev
+
+    @dev.setter
+    def dev(self, value: bool) -> None:
+        self._dev = value
+        self._cache_.pop("_dev", None)
```
jezdez: `ParameterLoader` only implements `__get__`, so assigning `self._dev` creates an instance attribute. `reset_context()` rebuilds `_cache_` but leaves that attribute behind. With `CONDA_DEV=0`, a `hook --dev` followed by `reset_context()` still leaves `_dev` true. The manual `context.__dict__.pop("_dev")` in the new test is masking this leak.

Store the temporary override in the cache that reset already clears. Please also assert that reset restores the configured value and remove the manual cleanup.

```suggestion
        self._cache_["_dev"] = value
```

tests/cli/test_conda_argparse.py
```diff
@@ -157,3 +157,23 @@ def test_sorted_commands_in_error(capsys: CaptureFixture):
...
+def test_dev_flag_absent_is_not_deprecated() -> None:
+    parser = generate_parser()
+    args = parser.parse_args(["create", "-n", "x"])
+    assert not args.dev
+
+
+def test_init_dev_is_not_deprecated() -> None:
+    parser = generate_parser()
+    args = parser.parse_args(["init", "--dev"])
+    assert args.dev
```
jezdez: This only proves that argparse itself does not warn. `main_subshell()` reinitializes `Context` from the namespace, and `_dev` aliases `dev`, so `init --dev` still enters the deprecated configuration path. On this head, `python -W always::PendingDeprecationWarning -m conda init --dev --dry-run bash` emits five `Context.dev` warnings. The base emits none. That contradicts the stated revert.

Please give the init-only option a destination that does not feed `_dev`, preserve the existing generated shell behavior, and make this an execution-level regression. For example, with `warnings` imported at module scope:

```suggestion
def test_init_dev_is_not_deprecated(conda_cli: CondaCLIFixture) -> None:
    with warnings.catch_warnings():
        warnings.simplefilter("error", PendingDeprecationWarning)
        _, stderr, rc = conda_cli("init", "--dev", "--dry-run", "bash")

    assert rc == 0, stderr
```

conda/base/context.py
```diff
@@ -884,7 +898,13 @@ def conda_exe_vars_dict(self) -> dict[str, str | None]:
...
         None means unset it.
         """
-        if context.dev:
+        if self._dev:
+            deprecated.topic(
+                "27.3",
+                "27.9",
+                topic="`conda.base.context.Context.dev`",
+                addendum="Set `PYTHONPATH` to the conda source root instead.",
+            )
```
jezdez, line 907: This path handles `CONDA_DEV=1` and `dev: true`, not only library access. `deprecated.topic()` defaults to `DeprecationWarning`. In 27.3, normal warning filters hide it, so these users receive no active-period warning before removal in 27.9. This differs from CLI actions and prior configuration deprecations, which use `FutureWarning`.

Once the retained init path is separated from this setting, make this user-facing. Please also pin a handler to 27.3 in the test and assert the exact category, since `pytest.deprecated_call()` accepts both.

```suggestion
                deprecation_type=FutureWarning,
            )
```

jezdez (changes requested):
Thanks for following up on the earlier review. I rechecked the current head against the base commit and the original reproductions. Three focused issues remain: the dev override survives `reset_context()`, the retained `init --dev` path is still indirectly deprecated, and the configuration warning is hidden during the active deprecation period.

I do not think the broader wrapper, Windows, warning-count, or real-shell changes are needed here. Non-blocking: the PR description still lists `init --dev` as deprecated, and the news claim about reactivation should be qualified as configuration-backed.

conda/base/context.py
kenodegard: https://github.com/conda/conda/pull/16571/commits/58cd16f01a6111d43d59f750366af4bbf6ab094f

tests/cli/test_conda_argparse.py
kenodegard: https://github.com/conda/conda/pull/16571/commits/58cd16f01a6111d43d59f750366af4bbf6ab094f

conda/base/context.py
kenodegard: https://github.com/conda/conda/pull/16571/commits/58cd16f01a6111d43d59f750366af4bbf6ab094f

jezdez (approved):

PR #16573, opened 2026-08-20
Align `fn` and `build_number` in solve test fixtures

### LINKED ISSUE #75: Copy-paste quirk in `test_determine_constricting_specs_conflicts` fixture record
The `mypkg` record in `test_determine_constricting_specs_conflicts` declares `version="0.1.0"`, but `fn="mypkg-0.1.1"` (with `build="pypi_0"`, yet `build_number=1`), see https://github.com/conda/conda/blob/03329e0f4a627c9b9aa92ef34f7f93b9aa83e438/tests/core/test_solve.py#L3500-L3514.

The mismatch is inert because `determine_constricting_specs` never reads `fn`, but I think the record would be cleaner with a consistent filename.

agriyakhetarpal:
### Description

This PR fixes two inconsistencies in the fixtures for the `determine_constricting_specs` and `get_constrained_packages` tests, in that (i) the records said `version="0.1.0"` but noted `fn` as `"mypkg-0.1.1"`, and (ii) similarly, `build="pypi_0"` with `build_number=1`. Both of these fields are inert in this case and are not read by the code being tested. I had spotted this while porting these tests to pytest-conda-solvers.

Closes conda-incubator/pytest-conda-solvers#75, which turned out to be a nothingburger

### Checklist - did you ...


tests/core/test_solve.py
```diff
@@ -3708,7 +3708,7 @@ def test_determine_constricting_specs_multi_conflicts(tmpdir):
             subdir="conda-test",
             fn="mypkgnot-1.1.1",
```
jezdez: Keep the filename aligned with the record name too.

```suggestion
            fn="notmypkg-1.1.1",
```

jezdez (approved):
The edited fixture metadata is consistent apart from the inline suggestion.

PR #16579, opened 2026-08-21
Fix always-true skip condition in `test_dont_remove_conda_3`

danyeaw:
### Description

I noticed that in #16116, we accidentally introduced an always True skip condition in `test_dont_remove_conda_3`. This is a classic De Morgan's law problem:

x != A or x != B (with A ≠ B) is a "tautology". A single value can't equal both "libmamba" and "classic" at once, so at least one of the two != comparisons is always true, making it always true. If it is always true, we always skip it.

### Checklist - did you ...

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2411,7 +2411,7 @@ def test_dont_remove_conda_3(
     upgrades a dependency) it could produce spurious RemoveError, blocking
     further use of conda.
     """
-    if context.solver != "libmamba" or context.solver != "classic":
+    if context.solver not in ("libmamba", "classic"):
         pytest.skip(
             "This test can only be run with solvers that come shipped with conda"
         )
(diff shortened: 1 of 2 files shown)


tests/test_create.py
```diff
@@ -2411,7 +2411,7 @@ def test_dont_remove_conda_3(
     upgrades a dependency) it could produce spurious RemoveError, blocking
     further use of conda.
     """
-    if context.solver != "libmamba" or context.solver != "classic":
+    if context.solver not in ("libmamba", "classic"):
```
soapy1: conda now ships with rattler solver. I think it might be appropriate to add that here now

```suggestion
    if context.solver not in ("libmamba", "classic", "rattler"):
```

kenodegard (approved):

jezdez (approved):

PR #16585, opened 2026-08-24
Use tmp_env and local wheel for slow test

ForgottenProgramme:
### Description
#16344 

use `tmp_env` and a local wheel instead of `click`. 
The improvement really is marginal. 

### Checklist - did you ...

### DIFF
--- a/tests/cli/test_env.py
+++ b/tests/cli/test_env.py
@@ -425,6 +425,7 @@ def test_update_env_only_pip_json_output(
     conda_cli: CondaCLIFixture,
     request: pytest.FixtureRequest,
     wheelhouse: Path,
+    monkeypatch: MonkeyPatch,
 ):
     """
     Update an environment by adding only a pip package.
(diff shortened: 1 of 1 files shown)


tests/cli/test_env.py
```diff
@@ -438,18 +439,35 @@ def test_update_env_only_pip_json_output(
             reason="Known issue: https://github.com/conda/conda-libmamba-solver/issues/320",
```
danyeaw, line 440:
```suggestion
            reason="Known issue: https://github.com/conda/conda-libmamba-solver/issues/320",
            run=False,
```

What if we set this not to run? Right now we are doing a full create, update, and pip cycling and then failing which is pretty wasteful.

danyeaw (approved):
Hey @ForgottenProgramme, these changes look great. We could also add a `monkeypatch.setenv("PIP_DISABLE_PIP_VERSION_CHECK", "1")` to avoid pip checking for a version update which would save at least one more network call.

ForgottenProgramme:
@danyeaw, good ideas! Updated the PR per your suggestions.

danyeaw (approved):
Thanks!

PR #16596, opened 2026-08-28
Calculate and set `sha256_in_prefix` after only codesign batch is over

### LINKED ISSUE #16590: Regression: `conda doctor` reports false-positive altered files for all binary prefix-replaced files
**Checklist**

- [x] I searched open reports and found no duplicate

**What happened?**

Someone at Anaconda reported that their base environment had altered-files when running `conda doctor`, and I was able to immediately reproduce this even thought I have a frozen base environment.

On macOS arm64 with conda ≥ 26.7.0, every package containing files that undergo *binary* prefix replacement is immediately reported as having altered files by `conda doctor`, straight after a fresh install. Nothing modified the files after installation.

Minimal reproducer:

```bash
conda create -y -n repro openssl
conda doctor -n repro
```

Expected:

```
✅ There are no packages with altered files.
```

Actual:

```
❌ Altered Files:

openssl-4.0.1-hd24854e_1:
  lib/libcrypto.4.dylib
```

(For openssl, `lib/libcrypto.4.dylib` is the only file with `file_mode: binary` + `prefix_placeholder` in `conda-meta/openssl-4.0.1-hd24854e_1.json` — the flagged files always correspond exactly to the set of binary prefix-replaced files. Larger installs flag krb5, libcurl, tk, libxml2, mamba, Python extension modules, etc. — always exclusively Mach-O binaries.)

Confirmation that the recorded hash is the pre-signing hash — emulate conda's own binary prefix replacement on the pristine package-cache file (this is exactly the byte content `PrefixReplaceLinkAction.verify()` hashes at `path_actions.py:607`):

```python
import hashlib, json, re
from pathlib import Path

prefix = Path("/tmp/repo")  # the repro env
meta = json.loads(next((prefix / "conda-meta").glob("openssl-*.json")).read_text())
entry = next(p for p in meta["paths_data"]["paths"] if p["_path"] == "lib/libcrypto.4.dylib")

cached = Path.home() / "miniforge3/pkgs/openssl-4.0.1-hd24854e_1/lib/libcrypto.4.dylib"
data = cached.read_bytes()  # adjust path to your pkgs cache / package version

placeholder = entry["prefix_placeholder"].encode()
new_prefix = str(prefix).encode()
pat = re.compile(re.escape(placeholder) + rb"(?:(?!\0).)*\0", re.DOTALL)
def repl(m):
    s = m.group()
    pad = (len(placeholder) - len(new_prefix)) * s.count(placeholder)
    return s.replace(placeholder, new_prefix) + b"\0" * pad

digest = hashlib.sha256(pat.sub(repl, data)).hexdigest()
print("recomputed:", digest)
print("recorded:  ", entry["sha256_in_prefix"])
print("match:", digest == entry["sha256_in_prefix"])
```

This shows that they match.

**Root cause**

#15975 batched osx-arm64 codesigning for performance. The batching changes the ordering of operations during transaction verification:

1. `PrefixReplaceLinkAction.verify()` rewrites prefixes in an intermediate copy and computes `sha256_in_prefix` over it (`conda/core/path_actions.py:593-607`).
2. With batching active (`batch_codesign_calls()` wraps all `verify()` calls in `conda/core/link.py:633`), `update_prefix()` only *enqueues* the codesign call (`conda/core/portability.py:131-132`, `154-159`); the actual `codesign -s - -f` runs when the batch context exits (`portability.py:183-187`) — i.e. **after** the hash was recorded.
3. The signed intermediate is then linked into the prefix (`path_actions.py:619-625`).

So `sha256_in_prefix` in `conda-meta` is the hash of the unsigned file, while the on-disk file is ad-hoc signed, and the `altered-files` health check (`conda/plugins/subcommands/doctor/health_checks/altered_files.py`) compares exactly these two values. Consequences:

- Every binary prefix-replaced file installed on osx-arm64 is a false positive.
- Reinstalling affected packages re-records the same wrong hash, so `conda doctor altered-files --fix` can never clear the report.
- The reports appear even in frozen/protected environments, since the mismatch is created at install time by conda itself.

Before #15975, signing happened synchronously inside `update_prefix()`, before the hash was computed, so the recorded hash was correct.

**Suggested fix**

Ensure `sha256_in_prefix` is computed over the final, signed content — e.g. sign the intermediate synchronously before hashing in `PrefixReplaceLinkAction.verify()`, or defer the hash computation until the pending codesign batch has been flushed for that intermediate.

**Conda info**

```shell
conda version : 26.7.1
platform : osx-arm64
channels : conda-forge
```

**Additional context**

ForgottenProgramme:
### Description

Fixes #16590.

PR #15975 batches `codesign` operations while package actions are verified. `PrefixReplaceLinkAction.verify()` recorded `sha256_in_prefix` before the queued signing operation ran, so the stored digest represented the unsigned intermediate file rather than the signed file linked into the prefix. `conda doctor` then reported affected binary prefix-replaced files as altered immediately after installation on `osx-arm64`.

Compute `sha256_in_prefix` in `PrefixReplaceLinkAction.execute()` after queued `codesign` operations have completed and before linking the intermediate file into the prefix. The stored digest now represents the installed file.

### Checklist - did you ...

- [ ] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?


ForgottenProgramme:
```
(base) Fri Aug 28 14:49:51 maheiramkhan 😼 ~/documents/github/conda >conda create -y -n repro2 openssl
...
#
#
#     $ conda activate repro2
#
#
#     $ conda deactivate

(base) Fri Aug 28 14:50:03 maheiramkhan 😼 ~/documents/github/conda >conda doctor -n repro2
Environment Health Report for: /Users/maheiramkhan/documents/github/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniconda/envs/repro2

✅ There are no packages with altered files.
...

✅ No pinned specs found in /Users/maheiramkhan/documents/github/conda/devenv/Darwin/arm64/envs/devenv-3.10-miniconda/envs/repro2/conda-meta/pinned.
```

releases/news/16590-doctor-altered-files-codesign
```diff
@@ -0,0 +1,19 @@
+### Enhancements
```
jezdez, line 1: Could you move this fragment to `releases/news/16590-doctor-altered-files-codesign`? `rever_overrides.py` now reads release fragments only from `releases/news/`, so files under the top-level `news/` directory do not make it into the generated changelog.

jezdez (changes requested):
Thanks for fixing this, Mahe. The code change and regression test look good, and I reproduced the OpenSSL case locally. Just one small file move before this is ready.

jezdez (approved):

jezdez:
@ForgottenProgramme FYI I've taken the liberty to update the PR description above to make sure we have a good context.

jdalbers:
Great that this will be fixed, I hit the same symptom on macOS arm64.

Environment:
- macOS arm64
- conda 26.7.2
- Miniconda installed at /opt/miniconda3

`conda doctor --verbose` reports:

openssl-3.5.8-ha0b305a_0:
  lib/libcrypto.3.dylib

The package cache hash matches the package metadata:

74a193fd0ddd24f405c89c7d48399b35a5aff07fc21e5594cd4c30c263694545

but the installed prefix-rewritten file is:

a13a159d1e0546f0002e7e675183d3f83a8d480aad97149f68a0831c7a634dca

The conda-meta entry contains:

"sha256_in_prefix": "c326694dd577fd037851d4b470e140be76a28ce1125f1bb7150e71502366ad43"

The installed binary contains the expected /opt/miniconda3 paths, so the mismatch appears to happen during prefix replacement.

PR #16597, opened 2026-08-28
Adding breaking changes guide

### LINKED ISSUE #16570: Docs: Dev guide for how breaking CLI changes to conda are made
### Checklist

### What is the idea?

Add developer-guide documentation that describes how conda plans, announces, and ships breaking changes.

CEP 9 and the existing deprecations guide cover marking and removing public APIs. We also make user-visible breaking changes that are not API removals: changing config defaults, changing the default solver, and eventually removing or changing CLI subcommands.

The docs should turn past practice into a reusable process maintainers can follow and contributors can find next to Deprecations and Previews.

### Why is this needed?

We have several high-impact changes in flight or ahead (for example changing the `add_pip_as_python_dependency` default in #16404, and changing the default solver). Without a written process, we are getting in to disagreements about what we must do to make the change, or falling back to trying to apply CEP 9.

Clear docs reduce surprise for users and automation, keep solvers and config behavior consistent during transitions, and give reviewers a shared bar for when things are ready.

### What should happen?

Add a short developer-guide page (and link it from the Contributing & quality section) that covers at least:

**Change classes**
- Config default changes (keep the setting, change only the implicit default) or config removals
- Default solver (or other high-impact default) changes
- CLI subcommand or flag removals and behavior changes
- How these relate to CEP 9 API/behavior deprecations (and when CEP 9 alone is not enough)

**Communication and rollout patterns we already use**
- Warn on the CLI when it is practical (and when not to warn: explicit user opt-in/opt-out)
- Announce in the *previous* release notes, and publish a conda.org blog post for major changes
- Prefer a dedicated special release on an even month for the highest-impact default flips (as with making libmamba the default)
- Provide clear migration and opt-out paths in docs and release notes
- For installer-coupled behavior (for example protected base), consider a beta Miniconda/Miniforge installer before the general rollout

The page should point at precedents from previous changes.

### Additional Context

Related existing material to link or extend, not replace:
- CEP 8 (releases) and CEP 9 (deprecation schedule)
- Developer guide: Deprecations, Previews, Releasing
- Changelog special announcements for the libmamba default change (23.9.0 / 23.10.0)
- #16404

travishathaway:
### Description

closes: #16570 

### Checklist - did you ...

- [ ] ~Add a file to the `releases/news` directory ([using the template](https://github.com/conda/conda/blob/main/releases/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?~

### DIFF
--- a/docs/source/dev-guide/breaking-changes.md
+++ b/docs/source/dev-guide/breaking-changes.md
@@ -1,0 +1,131 @@
+[cep8]: https://github.com/conda/ceps/blob/main/cep-0008.md
+[cep9]: https://github.com/conda/ceps/blob/main/cep-0009.md
+
+# Breaking changes
+
+While conda strives towards providing a highly stable software and package management experience,
+sometimes the introduction of breaking changes cannot be avoided.
+For example, we may need to change a configuration default, update the solver,
+or remove/change a CLI subcommand or flag. While these changes are usually few and far between,
+it's important that we define the appropriate procedure to follow for our maintainers.
+Additionally, [CEP 9][cep9] defines the deprecation schedule
+for public API and behavior *removals*, but it does not by itself say how to plan, announce,
+and ship a user-visible breaking change that isn't a symbol removal. This page describes the
+process we use for that, building on CEP 9's schedule and [CEP 8][cep8]'s release cadence.
+
+Read this together with {doc}`Deprecations <deprecations>` (the warning mechanics in code) and
+{doc}`Releasing <releasing>` (how a release actually gets cut).
+
+## Change categories
+
+**Config default changes and removals**
+: The setting itself stays supported; only its *implicit* default value changes (for example,
+  flipping `add_pip_as_python_dependency` from `true` to `false`), or the setting is removed
+  outright. Users who already set the value explicitly should see no behavior change.
+
+**Default solver (or other high-impact default) changes**
+: Swapping which plugin or implementation runs when the user hasn't chosen one, such as the
+  23.10.0 switch to `conda-libmamba-solver`. These tend to have the widest blast radius and are
+  the most likely candidates for a dedicated special release (see below).
+
+**CLI subcommand or command option removals and behavior changes**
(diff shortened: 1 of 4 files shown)


docs/source/dev-guide/breaking-changes.md
```diff
@@ -0,0 +1,131 @@
...
+  does — not just deprecating the Python API that backs it.
+
+Every category follows the same rollout tools in [Communication and rollout](#communication-and-rollout);
+they differ mainly in *how* conda warns users ahead of time.
+
+## Relation to CEP 9
+
+[CEP 9][cep9] and {doc}`Deprecations <deprecations>` give us the schedule (pending deprecation →
+deprecation → removal) and the tooling (`deprecated`, `deprecated.argument`, `deprecated.action`,
+`deprecated.constant`, `deprecated.module`, `deprecated.topic`) for deprecating a **symbol**.
```
kenodegard: Incase more are added in the future.

```suggestion
`deprecated.constant`, `deprecated.module`, `deprecated.topic`, etc.) for deprecating a **symbol**.
```

kenodegard (approved):

ryanskeith (approved):

ryanskeith:
The guide looks reasonable and straight forward.  One thing that may be out of scope but worth mention is how to propose a breaking change and the process in which it gets approved.  I could see it in this guide.  I could also see it put in its own guide of its own.

travishathaway:
> The guide looks reasonable and straight forward. One thing that may be out of scope but worth mention is how to propose a breaking change and the process in which it gets approved. I could see it in this guide. I could also see it put in its own guide of its own.

I think this is more-or-less covered by this pull request:

- https://github.com/conda/conda-maintainers-team-planning/pull/121

As far as the decision making process is concerned, I don't believe breaking changes deserve special treatment.

PR #16599, opened 2026-08-28
Clarify repodata retry message

### LINKED ISSUE #12008: Hide `failed with repodata from current_repodata.json, will retry with next repodata source.`
### Checklist

### What is the idea?

I often get this message when installing packages with conda:

This means that conda failed solving the constraints with the packages available in the `current_repotada` index. This index contains a subset of the whole index, including only the latest versions of the index and their dependencies. Using a smaller index makes conda solve faster, which is why conda attempts first the solve based on this smaller index. See more details about that here: https://docs.conda.io/projects/conda-build/en/latest/concepts/generating-index.html#trimming-to-current-repodata

As a user I think this message is difficult to understand and makes me feel like I'm doing something wrong.

I'd say that this is an implementation detail that should not be exposed to the users and that I would suggest only displaying the message in verbose mode.

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

codewithdaniel1:
### Description

Fixes #12008

This clarifies the classic solver's repodata fallback message.

Instead of reporting an "unsuccessful attempt" before retrying, Conda now reports that it is "retrying with next repodata source." This keeps the retry visible while avoiding wording that can make a normal fallback look like an error.

This follows the discussion in #12008 to retain the retry information while making the message clearer and less alarming.

A regression test was added for the classic solver retry message.

### Checklist - did you ...

- [x] ~~If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?~~

### DIFF
--- a/releases/news/12008-clarify-repodata-retry
+++ b/releases/news/12008-clarify-repodata-retry
@@ -1,0 +1,3 @@
+### Bug fixes
+
+* Clarify the repodata retry message so it no longer describes the fallback as an unsuccessful attempt. (#12008 via #16599)
--- a/releases/news/12008-clarify-repodata-retry
+++ b/releases/news/12008-clarify-repodata-retry
@@ -1,3 +1,3 @@
-### Bug fixes
+### Enhancements
 
 * Clarify the repodata retry message so it no longer describes the fallback as an unsuccessful attempt. (#12008 via #16599)
--- a/conda/core/solve.py
+++ b/conda/core/solve.py
@@ -508,10 +508,7 @@ class Solver(BaseSolver):
                 " with flexible solve.\n"
             )
         elif self._repodata_fn != REPODATA_FN:
-            fail_message = (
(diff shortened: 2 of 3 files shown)


releases/news/12008-clarify-repodata-retry
```diff
@@ -0,0 +1,3 @@
+### Bug fixes
```
travishathaway:
```suggestion
### Enhancements
```

I think this would be better listed under this section in the release notes because it's not really a bug fix.

travishathaway (dismissed):
Hi @codewithdaniel1,

Thanks for finding and submitting an issue for the open issue! I just had one suggestion for the release notes. Otherwise, I think this looks good.

jezdez (approved):

danyeaw (approved):

danyeaw:
Thanks @codewithdaniel1!

PR #16607, opened 2026-08-31
Add stdin, stderr, and interactive TTY helpers

### LINKED ISSUE #16594: IS_INTERACTIVE deprecation isn't a 1:1 replacement
I noticed that the deprecation warning introduced in #15982, gave a hint to use `is_tty()` as a replacement. However, the implementation changed the definition of `is_tty()` to check if stdout is interactive, but removed the check for stdin.

In at least one plugin, conda-anaconda-tos, knowing whether stdin is interactive is also important, so that commands like `conda tos accept < /dev/null` don't error out. See https://github.com/anaconda/conda-anaconda-tos/pull/295

We could make this change directly in the plugin to also check stdin, but I thought I would open this to discuss if the change of the `IS_INTERACTIVE` definition was intentional.

CC: @travishathaway

travishathaway:
### Description

Closes #16594.

Keep `conda.common.terminal.is_tty()` focused on stdout. Add `is_stdin_tty()`
for stdin, `is_stderr_tty()` for stderr, and `is_interactive_tty()` for commands
that should prompt only when both stdout and stdin are connected to a TTY.

Update the `conda.common.io.IS_INTERACTIVE` deprecation guidance so callers can
select the helper that matches the stream behavior they need.

### Checklist - did you ...

- [ ] If this change is QA-worthy, add a snippet under `releases/qa/` ([using the template](https://github.com/conda/conda/blob/main/releases/qa/TEMPLATE))?


tests/common/test_terminal.py
```diff
@@ -88,3 +97,79 @@ def test_should_use_color(
...
+
+
+def test_is_tty_rejects_positional_args():
+    with pytest.raises(TypeError):
+        is_tty(True)
+    with pytest.raises(TypeError):
+        is_tty(True, True)
+
+
+def test_is_tty_missing_isatty_attribute_is_not_a_tty(monkeypatch):
```
ForgottenProgramme: The name of this test can be simplified a bit maybe. 
`test_is_tty_returns_false_without_isatty_attr` ? 

Not a blocker.

ForgottenProgramme (approved):
Looks good to me. 👍🏼

tests/common/test_terminal.py
jezdez:
```suggestion
def test_is_tty_returns_false_without_isatty_attr(monkeypatch):
```

jezdez (approved):
Agree with Mahe

jezdez (changes requested):
Thanks, Travis. I approved this too quickly earlier. Looking at the public API again, this adds the modes I asked us not to put on `is_tty()` in [my earlier review of #15982](https://github.com/conda/conda/pull/15982#discussion_r3340522388).

Please keep `is_tty()` focused on stdout and add a separately named helper for the prompt/input case that needs stdin and stdout. If there is a current caller that needs stderr too, that behavior should be named explicitly rather than added as another mode of `is_tty()`.

travishathaway:
@jezdez,

This is ready for another review:

1. `is_tty` has been reverted to its original state.
2. `is_interactive_tty`, `is_stdin_tty` and `is_stderr_tty` have been added as new options.
3. Deprecation notice now mentions `is_stdin_tty` and `is_interactive_tty`.

jezdez (approved):

PR #16615, opened 2026-09-02
Protect cache packages referenced by environment symlinks during cleanup

### LINKED ISSUE #16614: `conda clean --packages` deletes packages used by `always_softlink` environments
### Checklist

### What happened?

`conda clean --packages`, which is also run by `conda clean --all`, considers an extracted package unused when none of its files has a hard-link count greater than one. Symbolic links do not increase that count. As a result, `conda clean` deletes package-cache files that environments created with `always_softlink: true` still require, leaving dangling links and broken environments.

This reproduces with conda 26.7.1 from the current Miniforge installer:

```bash
docker run --rm -it ubuntu:latest bash
apt-get update -qq
apt-get install -y -qq curl bzip2 ca-certificates

curl -fsSL -o /tmp/miniforge.sh \
  "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-$(uname -m).sh"
bash /tmp/miniforge.sh -b -p /opt/miniforge3

/opt/miniforge3/bin/conda config --set always_softlink true
source /opt/miniforge3/etc/profile.d/conda.sh
conda activate base
conda install --force-reinstall python -y

readlink /opt/miniforge3/lib/python*/encodings/__init__.py
/opt/miniforge3/bin/python3 -c 'print(1)'

conda clean --all --yes
/opt/miniforge3/bin/python3 -c 'print(1)'
```

Before cleanup, `encodings/__init__.py` points into the extracted Python package and Python runs:

```text
/opt/miniforge3/pkgs/python-3.14.7-hbec3b18_104_cp314/lib/python3.14/encodings/__init__.py
1
```

Cleanup removes that directory:

```text
Will remove 127 (137.9 MB) tarball(s).
Will remove 1 index cache(s).
Will remove 18 (285.6 MB) package(s).
There are no tempfile(s) to remove.
There are no logfile(s) to remove.
```

The Python process then exits with status 1:

```text
Could not find platform independent libraries <prefix>
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
LookupError: no codec search functions registered: can't find encoding
```

The same behavior reproduces on current `main` at f79624e4a80a6f2de7079cf44235ac0220eb742e. A package installed with `CONDA_ALWAYS_SOFTLINK=true` has an activation script linked into the extracted package cache. `conda clean --packages --yes` removes that cache directory and leaves the activation script link dangling.

`conda clean --packages --help` warns that it does not check for packages installed using symbolic links. The `--all` help text does not show that warning, and neither command presents a specific runtime warning before deleting these packages.

I expected `conda clean --packages` and the package-removal part of `conda clean --all` not to delete extracted package directories that known environments still reference.

Conda already records the extracted cache directory and requested link type in each prefix's `conda-meta/*.json`, and `list_all_known_prefixes()` finds registered environments plus environments under configured `envs_dirs`. Cleanup could use those records to preserve known soft-linked packages. If conda cannot determine that cleanup is safe, it should skip extracted-package removal with an actionable message. `--yes` should only answer prompts, not bypass this safety check.

### Conda Info

```shell
active environment : base
active env location : /opt/miniforge3
shell level : 1
populated config files : /opt/miniforge3/.condarc
                         /root/.condarc
conda version : 26.7.1
python version : 3.14.7.final.0
solver : libmamba (default)
platform : linux-aarch64
```

### Conda Config

```shell
==> /opt/miniforge3/.condarc <==
channels:
  - conda-forge

==> /root/.condarc <==
always_softlink: True
```

### Conda list

```shell
# packages in environment at /opt/miniforge3:
#
# Name    Version  Build               Channel
conda     26.7.1   py314h9dd9068_0     conda-forge
python    3.14.7   hbec3b18_104_cp314  conda-forge
```

### Additional Context

The original `conda clean` discussion already noted that the hard-link-count approach breaks packages installed through symbolic links: https://github.com/conda/conda/issues/68#issuecomment-37984861

A reverse cache-side hint list was proposed in #1191 and rejected because it did not work for multi-user installations or read-only package caches. This report does not require that design. Existing prefix metadata can protect known environments, while a runtime guard can handle cases conda cannot inspect.

Open #10999 concerns the expected cache-linking behavior, not cleanup.

The current implementation and available metadata are visible here:

- https://github.com/conda/conda/blob/f79624e4a80a6f2de7079cf44235ac0220eb742e/conda/cli/main_clean.py#L107-L124
- https://github.com/conda/conda/blob/f79624e4a80a6f2de7079cf44235ac0220eb742e/conda/core/path_actions.py#L984-L988
- https://github.com/conda/conda/blob/f79624e4a80a6f2de7079cf44235ac0220eb742e/conda/core/envs_manager.py#L118-L161

jezdez:
### Description

Fixes #16614.

`conda clean --packages` uses hard-link counts to identify unused extracted packages. Symbolic links do not change those counts, so cleanup can delete cache files that environments created with `always_softlink` or `allow_softlinks` still need.

When the hard-link check finds removal candidates, conda now reads package records for every environment it can discover and keeps cache directories that those records reference through symbolic links. The scan includes registered environments, configured `envs_dirs`, and the root, conda, active, and target prefixes. It compares normalized paths first, then uses filesystem identity when different paths point to the same directory.

Cleanup still removes unreferenced cache packages when either soft-link option is enabled. Conda cannot protect an environment it cannot discover, so the `--packages` help now calls that out.

`get_softlinked_package_dirs()` lives in `conda.core.package_cache_data`. Prefix discovery stays in `conda.core.envs_manager`.

`--force-pkgs-dirs` remains the explicit option for removing complete package caches.

#16616 tracks deprecating and eventually removing package-cache soft links for future environments.

### Checklist - did you ...

- [x] Add a file to `releases/news` for the next release?
- [x] ~~If this change is QA-worthy, add a snippet under `releases/qa/`?~~

### DIFF
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -182,23 +182,8 @@ def find_tarballs() -> dict[str, Any]:
 
 
 def find_pkgs() -> dict[str, Any]:
-    from ..base.context import context
-
     warnings: list[str] = []
     pkg_sizes: dict[str, dict[str, int]] = {}
-    if context.always_softlink or context.allow_softlinks:
-        warnings.append(
-            "WARNING: Skipping package cache cleanup because `always_softlink` or "
-            "`allow_softlinks` is enabled. Removing cache packages may break "
-            "environments that use symlinks."
-        )
-        return {
-            "warnings": warnings,
-            "pkg_sizes": pkg_sizes,
-            "pkgs_dirs": _get_pkgs_dirs(pkg_sizes),
-            "total_size": _get_total_size(pkg_sizes),
-        }
-
     for pkgs_dir in find_pkgs_dirs():
         # pkgs are directories in pkgs_dir
         _, pkgs, _ = next(os.walk(pkgs_dir))
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -11,6 +11,7 @@ import os
 import sys
 from logging import getLogger
 from os.path import isdir, join
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -62,8 +63,8 @@ def configure_parser(sub_parsers: _SubParsersAction, **kwargs) -> ArgumentParser
         "--packages",
         action="store_true",
         help="Remove unused packages from writable package caches. "
-        "WARNING: This does not check for packages installed using "
-        "symlinks back to the package cache.",
+        "WARNING: Environments not listed by `conda info --envs` are not checked "
+        "for symlinks back to the package cache.",
     )
     removal_target_options.add_argument(
         "-t",
@@ -181,8 +182,22 @@ def find_tarballs() -> dict[str, Any]:
 
 
 def find_pkgs() -> dict[str, Any]:
+    from ..base.context import context
+
     warnings: list[str] = []
     pkg_sizes: dict[str, dict[str, int]] = {}
+    if context.always_softlink:
+        warnings.append(
+            "WARNING: Skipping package cache cleanup because `always_softlink` is "
+            "enabled. Removing cache packages may break environments that use symlinks."
+        )
+        return {
+            "warnings": warnings,
+            "pkg_sizes": pkg_sizes,
+            "pkgs_dirs": _get_pkgs_dirs(pkg_sizes),
+            "total_size": _get_total_size(pkg_sizes),
+        }
+
     for pkgs_dir in find_pkgs_dirs():
         # pkgs are directories in pkgs_dir
         _, pkgs, _ = next(os.walk(pkgs_dir))
@@ -202,6 +217,29 @@ def find_pkgs() -> dict[str, Any]:
                 pass
             else:
                 pkg_sizes.setdefault(pkgs_dir, {})[pkg] = size
+
+    if pkg_sizes:
+        from ..core.package_cache_data import get_softlinked_package_dirs
+
+        softlinked_pkgs = get_softlinked_package_dirs()
+        softlinked_pkgs_by_name: dict[str, list[Path]] = {}
+        for path in softlinked_pkgs:
+            softlinked_pkgs_by_name.setdefault(path.name.casefold(), []).append(path)
+        for pkgs_dir, pkgs in pkg_sizes.items():
+            for pkg in tuple(pkgs):
+                package_dir = Path(pkgs_dir, pkg).resolve(strict=False)
+                if package_dir in softlinked_pkgs:
+                    del pkgs[pkg]
+                    continue
+
+                for softlinked_pkg in softlinked_pkgs_by_name.get(pkg.casefold(), ()):
+                    try:
+                        matches = package_dir.samefile(softlinked_pkg)
+                    except OSError:
+                        continue
+                    if matches:
+                        del pkgs[pkg]
+                        break
 
     return {
         "warnings": warnings,
--- a/conda/cli/main_clean.py
+++ b/conda/cli/main_clean.py
@@ -186,10 +186,11 @@ def find_pkgs() -> dict[str, Any]:
 
     warnings: list[str] = []
     pkg_sizes: dict[str, dict[str, int]] = {}
-    if context.always_softlink:
+    if context.always_softlink or context.allow_softlinks:
         warnings.append(
-            "WARNING: Skipping package cache cleanup because `always_softlink` is "
-            "enabled. Removing cache packages may break environments that use symlinks."
+            "WARNING: Skipping package cache cleanup because `always_softlink` or "
+            "`allow_softlinks` is enabled. Removing cache packages may break "
+            "environments that use symlinks."
         )
         return {
             "warnings": warnings,
--- a/conda/core/package_cache_data.py
+++ b/conda/core/package_cache_data.py
@@ -111,7 +111,10 @@ def get_softlinked_package_dirs() -> set[Path]:
 
     package_dirs = set()
     for prefix in prefixes:
-        for record in PrefixData(prefix, interoperability=False).iter_records():
+        prefix_data = PrefixData(prefix, interoperability=False)
+        if not prefix_data.is_environment():
+            continue
+        for record in prefix_data.iter_records():
             link = getattr(record, "link", None)
             # Older records may have a source without recording the link type.
             if link and getattr(link, "type", LinkType.softlink) == LinkType.softlink:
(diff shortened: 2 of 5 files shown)


jaimergp:
As a side note: I stumbled upon this issue a few days ago while trying to find information about `ltrace`, where Ray made the case to remove softlinking altogether from `conda`: https://github.com/conda-forge/cmake-feedstock/issues/21#issuecomment-256067803

jezdez:
A literal decade later!

conda/cli/main_clean.py
```diff
@@ -181,8 +182,22 @@ def find_tarballs() -> dict[str, Any]:
 
 
 def find_pkgs() -> dict[str, Any]:
+    from ..base.context import context
+
     warnings: list[str] = []
     pkg_sizes: dict[str, dict[str, int]] = {}
+    if context.always_softlink:
```
jaimergp: What if the environment was created with softlinks enabled in the past, but the option is no longer in use? Or the other way around, the option is now enabled, but no new environments were ever created, so there's no cleanup?

conda/core/package_cache_data.py
```diff
@@ -91,6 +91,34 @@
...
+    prefixes = set(list_all_known_prefixes())
+    prefixes.update(
+        prefix
+        for prefix in (
+            context.root_prefix,
+            context.conda_prefix,
+            context.active_prefix,
+            context.target_prefix,
+        )
+        if prefix
```
jaimergp, line 109: Is this trying to check whether the prefix exists? Or can some values be None? Maybe check both?

jezdez:
Thanks for digging this up. Ray's argument still holds for package-cache-to-prefix soft links. Conda no longer creates them by default, but both `always_softlink` and `allow_softlinks` can still do so. Existing environments also remain vulnerable even if we stop creating new ones.

I've extended this PR's early guard to cover `allow_softlinks` too, while keeping the prefix-record scan for existing environments. I opened #16616 to discuss deprecating and eventually removing cache-to-prefix soft-link creation. That proposal deliberately leaves package-internal symlinks and legacy `LinkType.softlink` record handling alone, and calls out cross-filesystem and CVMFS users so we can understand current use before deciding on the migration.

conda/cli/main_clean.py
jezdez: Those cases are now handled the same way. We inspect the prefix records for every environment conda can discover and keep cache sources referenced through soft links, regardless of the current setting values. Environments that `conda info --envs` cannot discover remain outside that scan, which the help text now calls out.

conda/core/package_cache_data.py
jezdez: The `if prefix` check filters values such as `active_prefix=None`. You're right that a non-`None` value can still point to a prefix that does not exist or is not a conda environment. f8b5b385e now calls `PrefixData.is_environment()` before reading records. That checks for `conda-meta/history` without loading an empty record set, and the test covers both `None` and a non-environment target path.

conda/cli/main_clean.py
```diff
@@ -181,8 +182,23 @@ def find_tarballs() -> dict[str, Any]:
...
+            "WARNING: Skipping package cache cleanup because `always_softlink` or "
+            "`allow_softlinks` is enabled. Removing cache packages may break "
+            "environments that use symlinks."
+        )
+        return {
+            "warnings": warnings,
+            "pkg_sizes": pkg_sizes,
+            "pkgs_dirs": _get_pkgs_dirs(pkg_sizes),
+            "total_size": _get_total_size(pkg_sizes),
+        }
```
beckermr: I do not think we need this bit of code. It does cover cases when these config settings are turned on and the user has environments not known by conda. However, if an environment is not known by the the conda installation, it is hard to argue that conda should be responsible for it.

beckermr:
Thanks for this PR @jezdez! I left one comment.

conda/cli/main_clean.py
jezdez: Yep, agreed. The known-prefix scan covers what conda should protect here, so I dropped this guard and adjusted the regression test. Good call, Matt!

conda/core/package_cache_data.py
```diff
@@ -91,6 +91,37 @@
...
+    package_dirs = set()
+    for prefix in prefixes:
+        prefix_data = PrefixData(prefix, interoperability=False)
+        if not prefix_data.is_environment():
+            continue
+        for record in prefix_data.iter_records():
+            link = getattr(record, "link", None)
+            # Older records may have a source without recording the link type.
+            if link and getattr(link, "type", LinkType.softlink) == LinkType.softlink:
+                package_dirs.add(Path(link.source).resolve(strict=False))
```
beckermr, line 121: When did conda/conda-build stop writing/producing records with no recorded link type? Can it still do that? When it did that, what kinds of links were these?

If these older records were hardlinks, then after this PR the following sequence of commands would, I think, have a different result for a package in `myenv` that has no recorded link type:

```bash
$ conda clean --all --yes
$ conda remove -n myenv --all --yes
```

Before this PR, the first conda clean command would remove the package in the cache. The second command will remove the package in `myenv`. Thus, if I understand hardlinks correctly, the disk space would be freed.

However, after this PR, the first command will not remove the package from the cache (since conda doesn't know the link type). The second command will continue to do so from `myenv`. Thus, a user would need a third `conda clean` invocation to finally reclaim the disk space.

That might be a fine/acceptable outcome of this PR, but I wanted to check my understanding before we merge.

One possible change, which I am unsure of, is to allow conda to remove older package caches even if they are referenced by an env (but that reference has no link type specified). Depending on when conda/conda-build started recording link types were recorded, this might be a way to preserve existing behavior while also adding the extra layer of safety.

beckermr (commented):
Sorry @jezdez! I have one more comment related to something I noticed in the code when taking a second pass at reading through it. 

I am not sure if it needs to be addressed or not, so any advice/thoughts are very welcome!

conda/core/package_cache_data.py
jezdez: No worries, Matt! I checked the history, and conda 2.0.0 already wrote both fields together. I couldn’t back up my comment about older records, so I’ll correct that.

One detail about your example: `clean` already skips packages with live hardlinks, so that sequence needs another `clean` after removing the environment today too.

Your concern does apply to copied packages whose records omit the link type. The model accepts those records, but we can’t tell from the source alone whether the files were copied or symlinked. I’d keep the conservative handling here and use a real `PrefixRecord` in the test to cover it.

beckermr: Sounds good and thank you!

beckermr (approved):

travishathaway (approved):

